PR #42, opened 2015-11-10
Run itests in travis

nhandler:
This Pull Request is to allow me to iterate on getting itests working in travis. Ignore it for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,17 @@
 language: python
 python:
   - "2.7"
+env:
+  - TOX_ENV=docs
+  - TOX_ENV=manpages
+  - TOX_ENV=general_itests
+  - TOX_ENV=paasta_itests
+sudo: required
+services:
+  - docker
 install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,6 +1,17 @@
 language: python
 python:
   - "2.7"
+env:
+  - TOX_ENV=docs
+  - TOX_ENV=manpages
```
solarkennedy: Hehe, I wouldn't bother with manpages

.travis.yml
```diff
@@ -1,6 +1,17 @@
 language: python
 python:
   - "2.7"
+env:
```
solarkennedy: Can you make an explicit unit test env? Otherwise will they end up running at all here?

nhandler: I have a `before_script` entry calling `tox -e py`. It should run before travis does the itests. This will allow users to catch syntax errors without having to wait for the long tests to run.

PR #43, opened 2015-11-11
Add coveralls support

nhandler:
Once again, using this pull request to test and iterate. Please disregard.


asottile:
I'd suggest a .coveragrc, here's one that might get you started: https://github.com/pre-commit/pre-commit/blob/master/.coveragerc

PR #44, opened 2015-11-11
Make setup_chronos_job handle the situation where the docker image isn't available

solarkennedy:
I don't really like how this code is structured, if anyone has better opinions on what we *should* be doing, I can redo it.

We currently catch all exceptional cases in main. This is another case that can happen, and without catching the exception we will crash. But of course it isn't "that" exceptional.

In marathon we kinda split it up, some exceptions are caught in main, some are caught closer to where the actual function is invoked.

### DIFF
--- a/paasta_tools/setup_chronos_job.py
+++ b/paasta_tools/setup_chronos_job.py
@@ -55,6 +55,7 @@ from paasta_tools.utils import decompose_job_id
 from paasta_tools.utils import InvalidJobNameError
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import NoDeploymentsAvailable
+from paasta_tools.utils import NoDockerImageError
 from paasta_tools.utils import SPACER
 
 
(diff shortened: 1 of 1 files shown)


paasta_tools/setup_chronos_job.py
```diff
@@ -195,8 +196,9 @@ def main():
             cluster=cluster,
             soa_dir=soa_dir,
         )
-    except NoDeploymentsAvailable:
-        error_msg = "No deployments found for %s in cluster %s" % (args.service_instance, cluster)
+    except (NoDeploymentsAvailable, NoDockerImageError):
+        error_msg = "No deployment found for %s in cluster %s. Has Jenkins run for it?" % (
```
Rob-Johnson: are these similar enough that both exceptions are covered with the same error_msg?

mrtyler:
shipit

paasta_tools/setup_chronos_job.py
solarkennedy: I think so. If you are deploying the service for the first time, or just deploying a new instance for the same time, The outcome is pretty much the same and the resolution is the same.

PR #45, opened 2015-11-11
Adds in the 'balance' parameter

oholiab:
To change loadbalancing method. Currently we just use the default which is least_conn which doesn't work for everything.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -440,6 +440,8 @@ def load_service_namespace_config(service, namespace, soa_dir=DEFAULT_SOA_DIR):
       e.g. [('region:dc6-prod', 'region:useast1-prod')]
     - extra_healthcheck_headers: a dict of HTTP headers that must
       be supplied when health checking. E.g. { 'Host': 'example.com' }
+    - balance: load balancing method e.g. 'leastconn' or 'roundrobin' (see
+      haproxy parameter of the same name for complete list)
 
     :param service: The service name
     :param namespace: The namespace to read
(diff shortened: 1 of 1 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -470,7 +472,8 @@ def load_service_namespace_config(service, namespace, soa_dir=DEFAULT_SOA_DIR):
         'mode',
         'discover',
         'advertise',
-        'extra_healthcheck_headers'
+        'extra_healthcheck_headers',
+        'balance'
```
mrtyler: trailing ,

mrtyler:
What is your test plan for this change?

There may be more docs to update, e.g. the docs on marathon-*.yaml will need to know about the balance key.

paasta_tools/marathon_tools.py
oholiab: Are you saying there should be one or there is one? Because it looks right to me (at least based on what it was before)

oholiab:
Cool, I'll get on that (docs and looking in to a test plan) tomorrow

paasta_tools/marathon_tools.py
mrtyler: There should be one. It was wrong before.

EvanKrall:
I don't think this change will do what you want it to; the HAProxy configs are created by synapse-tools, not by paasta directly.

oholiab:
I've got a branch where I'm doing the change in synapse-tools but it was my understanding that I had to make sure that paasta was expecting the extra configuration options - or will it just ignore any it doesn't know about (which would be ideal as it really doesn't need to know load balancing strategy)?

oholiab:
Looks like I was wrong, dropping the PR :)

PR #47, opened 2015-11-11
Flesh out docs for smartstack.yaml

EvanKrall:
This adds documentation for all of the keys listed in the `key_whitelist` in marathon_tools.py.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -265,9 +265,140 @@ Each job configuration MAY specify the following options:
 smartstack.yaml
 ---------------
 
-The yaml where nerve namespaces are defined and bound to ports.
+Configure service registration, discovery, and load balancing.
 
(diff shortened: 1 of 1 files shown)


mrtyler:
1. Is that markdown formatting correct?

2. I assume you copied this from some other (internal) doc? If so, do you have a plan to kill the internal-only one or otherwise prevent the copies from drifting from one another?

EvanKrall:
1. This is ReStructured Text, not Markdown. Code blocks are started by ending the previous line with `::`
2. This is new content, not copied from internally. Our internal docs are more substantial, but somewhat tied to our configuration. I'll try to pull that content into this.

solarkennedy:
Looks fine to me. Could be more RSTy but a well commented config file is just fine by me.

EvanKrall:
I've copied a bunch of info from our internal smartstack docs and RSTified it.

PR #51, opened 2015-11-13
Use the right srvname

mrtyler:
This was never caught by automated tests because this bug is only triggered in interactive mode
without --service-name


mittonk:
Ah, the joys of question-and-answer argument parsing.

Rob-Johnson:
ship!

PR #53, opened 2015-11-14
Mrtyler instance filter

mrtyler:
I wanted to try out some stuff with a new instance of the reminder service, but that service has a ton of instances so doing paasta status is slow and painful. Now we can filter instances like we can filter clusters:

```
(py)troscoe@dev2-devc:~/paasta_tools (mrtyler_instance_filter) $ ./paasta_tools/paasta_cli/paasta_cli.py status -s paasta_canary -c norcal-devc,qwer -i sensu_canary,asdf
Pipeline: https://jenkins.yelpcorp.com/view/services-paasta_canary

cluster: norcal-devc
  instance: sensu_canary
    Git sha:    b8517e76
    Config:     configbb7d0b2a
      Status:   Enabled, Started
      Last:     OK (2015-11-13T16:40, a minute ago)
      Schedule: R/2015-11-14T00:50:00.000Z/PT10M Epsilon: PT60S
      Command:  sensu-scheduled-canary.sh
      Mesos:    Not running

Warning: This service does not have any instance matching these filters:
asdf

Warning: This service does not have any cluster matching these filters:
qwer
```

### DIFF
--- a/paasta_tools/paasta_cli/cmds/status.py
+++ b/paasta_tools/paasta_cli/cmds/status.py
@@ -146,19 +146,17 @@ def report_status_for_cluster(service, cluster, deploy_pipeline, actual_deployme
 
         # Case: service deployed to cluster.instance
         if namespace in actual_deployments:
-            unformatted_instance = instance
-            instance = PaastaColors.blue(instance)
+            formatted_instance = PaastaColors.blue(instance)
             version = actual_deployments[namespace][:8]
             # TODO: Perform sanity checks once per cluster instead of for each namespace
-            status = execute_paasta_serviceinit_on_remote_master('status', cluster, service, unformatted_instance,
-                                                                 verbose=verbose)
+            status = execute_paasta_serviceinit_on_remote_master('status', cluster, service, instance, verbose=verbose)
         # Case: service NOT deployed to cluster.instance
         else:
-            instance = PaastaColors.red(instance)
+            formatted_instance = PaastaColors.red(instance)
             version = 'None'
             status = None
 
-        print '  instance: %s' % instance
+        print '  instance: %s' % formatted_instance
         print '    Git sha:    %s' % version
         if status is not None:
             for line in status.rstrip().split('\n'):
--- a/paasta_tools/paasta_cli/cmds/status.py
+++ b/paasta_tools/paasta_cli/cmds/status.py
@@ -171,16 +171,15 @@ def report_bogus_filters(filters, items, item_type):
     """Warns the user if the filter used is not even in the deployed
     list. Helps pick up typos"""
     return_string = ""
-    if filters is not None:
-        bogus_filters = []
-        for filt in filters:
-            if filt not in items:
-                bogus_filters.append(filt)
-        if len(bogus_filters) > 0:
-            return_string = (
-                "\n"
-                "Warning: This service does not have any %s matching these filters:\n%s"
-            ) % (item_type, ",".join(bogus_filters))
+    bogus_filters = []
+    for filt in filters:
+        if filt not in items:
+            bogus_filters.append(filt)
+    if len(bogus_filters) > 0:
+        return_string = (
+            "\n"
+            "Warning: This service does not have any %s matching these names:\n%s"
+        ) % (item_type, ",".join(bogus_filters))
     return return_string
 
 
@@ -190,7 +189,7 @@ def report_status(service, deploy_pipeline, actual_deployments, cluster_filter, 
 
     deployed_clusters = list_deployed_clusters(deploy_pipeline, actual_deployments)
     for cluster in deployed_clusters:
-        if cluster_filter is None or cluster in cluster_filter:
+        if not cluster_filter or cluster in cluster_filter:
             report_status_for_cluster(
                 service=service,
                 cluster=cluster,
@@ -212,11 +211,11 @@ def paasta_status(args):
     if args.clusters is not None:
         cluster_filter = args.clusters.split(",")
     else:
-        cluster_filter = None
+        cluster_filter = []
     if args.instances is not None:
         instance_filter = args.instances.split(",")
     else:
-        instance_filter = None
+        instance_filter = []
 
     if actual_deployments:
         deploy_pipeline = list(get_planned_deployments(deploy_info))
--- a/paasta_tools/paasta_cli/cmds/status.py
+++ b/paasta_tools/paasta_cli/cmds/status.py
@@ -164,24 +164,23 @@ def report_status_for_cluster(service, cluster, deploy_pipeline, actual_deployme
             for line in status.rstrip().split('\n'):
                 print '    %s' % line
 
-    print report_bogus_filters(instance_filter, seen_instances)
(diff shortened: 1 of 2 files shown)


paasta_tools/paasta_cli/cmds/status.py
```diff
@@ -200,9 +213,20 @@ def paasta_status(args):
         cluster_filter = args.clusters.split(",")
     else:
         cluster_filter = None
+    if args.instances is not None:
+        instance_filter = args.instances.split(",")
```
Rob-Johnson: IMO ``instance_filter`` is a misnomer - filters are generally functions rather than lists; I'd expect 'desired_instances' or something.

Saying that, I see that we use ``cluster_filter`` and ``bogus_filters`` elsewhere, so I guess you're just following prior art here.

paasta_tools/paasta_cli/cmds/status.py
```diff
@@ -125,69 +130,77 @@ def get_actual_deployments(service):
...
+        for filt in filters:
+            if filt not in items:
+                bogus_filters.append(filt)
+        if len(bogus_filters) > 0:
             return_string = (
                 "\n"
-                "Warning: The following clusters in the filter look bogus, this service\n"
-                "is not deployed to the following cluster(s):\n%s"
-            ) % ",".join(bogus_clusters)
+                "Warning: This service does not have any %s matching these filters:\n%s"
```
Rob-Johnson: I think it would make sense to s/filters/names here.

paasta_tools/paasta_cli/cmds/status.py
```diff
@@ -200,9 +213,20 @@ def paasta_status(args):
         cluster_filter = args.clusters.split(",")
     else:
         cluster_filter = None
+    if args.instances is not None:
+        instance_filter = args.instances.split(",")
+    else:
+        instance_filter = None
```
Rob-Johnson: can the default here be [] ? That way you don't have to test for None everywhere

paasta_tools/paasta_cli/cmds/status.py
```diff
@@ -125,69 +130,77 @@ def get_actual_deployments(service):
...
+    seen_instances = []
     for namespace in deploy_pipeline:
         cluster_in_pipeline, instance = namespace.split('.')
+        seen_instances.append(instance)
 
         if cluster_in_pipeline != cluster:
-            # This function only prints things that are relevant to cluster
-            # We skip anything not in this cluster
+            continue
+        if instance_filter and instance not in instance_filter:
```
Rob-Johnson: I've made a comment below about the default for ``instance_filter`` being ``[]`` so you don't need the first part of this conditional

paasta_tools/paasta_cli/cmds/status.py
```diff
@@ -125,69 +130,77 @@ def get_actual_deployments(service):
...
     """Warns the user if the filter used is not even in the deployed
     list. Helps pick up typos"""
     return_string = ""
-    if cluster_filter is not None:
-        bogus_clusters = []
-        for c in cluster_filter:
-            if c not in deployed_clusters:
-                bogus_clusters.append(c)
-        if len(bogus_clusters) > 0:
+    if filters is not None:
```
Rob-Johnson: see comment about replacing default for ``filter`` with ``[]``

paasta_tools/paasta_cli/cmds/status.py
mrtyler: So just ``if instance not in instance_filter:``?

Are you sure?

paasta_tools/paasta_cli/cmds/status.py
mrtyler: Yeah I didn't invent this nomenclature.

We could call them _whitelist ?

mrtyler: ok see this is dumb. we're still having this conversation, github. just because i pushed a commit doesn't mean we're done.

paasta_tools/paasta_cli/cmds/status.py
Rob-Johnson: ah, you got me. you still need the truthy test, otherwise nothing will be included when instance_filter is ``[]``.

paasta_tools/paasta_cli/cmds/status.py
Rob-Johnson: _whitelist is good.

mrtyler:
r2 - None -> [], filter -> whitelist

paasta_tools/paasta_cli/cmds/status.py
```diff
@@ -125,69 +130,76 @@ def get_actual_deployments(service):
...
             for line in status.rstrip().split('\n'):
                 print '    %s' % line
 
+    print report_bogus_whitelists(instance_whitelist, seen_instances, 'instance')
 
-def report_bogus_filters(cluster_filter, deployed_clusters):
-    """Warns the user if the filter used is not even in the deployed
+
+def report_bogus_whitelists(whitelists, items, item_type):
+    """Warns the user if the whitelist used is not even in the deployed
```
Rob-Johnson: this docstring doesn't make sense to me. """Warns the user if there are invalid values in the whitelist""" would be clearer.

paasta_tools/paasta_cli/cmds/status.py
```diff
@@ -125,69 +130,76 @@ def get_actual_deployments(service):
...
         if status is not None:
             for line in status.rstrip().split('\n'):
                 print '    %s' % line
 
+    print report_bogus_whitelists(instance_whitelist, seen_instances, 'instance')
 
-def report_bogus_filters(cluster_filter, deployed_clusters):
-    """Warns the user if the filter used is not even in the deployed
+
+def report_bogus_whitelists(whitelists, items, item_type):
```
Rob-Johnson: the ``whitelist`` parameter here is in a single whitelist, not multiple whitelists, right? If so, this function should be called ``report_invalid_whitelist_values`` or something similar (and the rest of the references to whitelists should be fixed where appropriate.

paasta_tools/paasta_cli/cmds/status.py
```diff
@@ -125,69 +130,76 @@ def get_actual_deployments(service):
...
-            if c not in deployed_clusters:
-                bogus_clusters.append(c)
-        if len(bogus_clusters) > 0:
-            return_string = (
-                "\n"
-                "Warning: The following clusters in the filter look bogus, this service\n"
-                "is not deployed to the following cluster(s):\n%s"
-            ) % ",".join(bogus_clusters)
+    bogus_whitelists = []
+    for whitelist in whitelists:
```
Rob-Johnson: shouldn't this be ``for value in whitelist``?

mrtyler: I'm using "entry" rather than "value" ("value" is super confusing in such close proximity with "items") but I generally agree with the demeanor of your suggestions.

mrtyler:
v3 - More renames

Rob-Johnson:
shipit

PR #54, opened 2015-11-14
replace the git references in requirements.txt

Rob-Johnson:
this fixes packaging issues in dh-virtualenv being seen


Rob-Johnson:
hm, actually, this isn't the right version of chronos-python. we need another release.

PR #55, opened 2015-11-15
upgrade marathon and chronos-python

Rob-Johnson:
- Use versions of chronos-python and marathon from pypi
- Fix a test broken by the new version of marathon

### DIFF
--- a/paasta_itests/paasta_metastatus.feature
+++ b/paasta_itests/paasta_metastatus.feature
@@ -14,7 +14,7 @@ Feature: paasta_metastatus describes the state of the paasta cluster
   Scenario: High memory usage
     Given a working paasta cluster
      When an app with id "memtest" using high memory is launched
-      And a task belonging to the app with id "memtest" is in the task list
+      And a task belonging to the app with id "/memtest" is in the task list
      Then paasta_metastatus exits with return code "2" and output "CRITICAL: Less than 10% memory available."
 
   # paasta_metastatus defines 'high' cpu usage as > 90% of the total cluster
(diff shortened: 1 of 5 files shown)


paasta_itests/paasta_metastatus.feature
```diff
@@ -14,7 +14,7 @@ Feature: paasta_metastatus describes the state of the paasta cluster
   Scenario: High memory usage
     Given a working paasta cluster
      When an app with id "memtest" using high memory is launched
-      And a task belonging to the app with id "memtest" is in the task list
+      And a task belonging to the app with id "/memtest" is in the task list
```
mrtyler: This is wrong imo; the id is not "/memtest" -- that would be confusing since we just in the previous step said the id was "memtest". If marathon adds the leading /, our step code should also add it.

mrtyler:
fix'n'ship but I have no insight into these version changes so I'm not sure what feedback you're looking for.

paasta_itests/paasta_metastatus.feature
Rob-Johnson: You're right to say this isn't the right place for the leading slash - though I'm not sure the step code is the right place either. We make an effort in marathon_tools[1] to hide the user from the leading slash, so I've added some logic there to be consistent.

[1] See [here](https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_tools.py#L759), [here](https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_tools.py#L883) for examples.

paasta_tools/marathon_tools.py
```diff
@@ -778,12 +778,15 @@ def app_has_tasks(client, app_id, expected_tasks, exact_matches_only=False):
...
     :param client: the marathon client
-    :param app_id: the app_id to which the tasks should belong
+    :param app_id: the app_id to which the tasks should belong. The app_id has a / prepended
+    to it in the case it does not start with one.
     :param expected_tasks: the number of tasks to check for
     :param exact_matches_only: a boolean indicating whether we require exactly expected_tasks to be running
     :returns: a boolean indicating whether there are atleast expected_tasks tasks with
         an app id matching app_id
     """
+    if not app_id.startswith("/"):
```
mrtyler: robj and I discussed. I don't like papering over this problem, so he's going to remove the conditional.

shipit once that happens (+ revert docstring change up there and new test down there)

PR #59, opened 2015-11-17
reject jobs with an interval more frequent than 60s

Rob-Johnson:
chronos has a default schedule_horizon of 60s, and you can cause problems
if you run jobs more frequently than the schedule_horizon (see [this bug](https://github.com/mesos/chronos/issues/508) ). more often
than 60s is pretty demanding on our infrastructure, so just disallow anything
more frequent for now

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -277,7 +277,13 @@ class ChronosJobConfig(InstanceConfig):
                                 'does not conform to the ISO 8601 format:\n%s' % (start_time, schedule, str(exc)))
 
             try:
-                isodate.parse_duration(interval)  # 'interval' and 'duration' are interchangeable terms
+                # 'interval' and 'duration' are interchangeable terms
+                parsed_interval = isodate.parse_duration(interval)
+                # until we make this configurable, throw an
+                # error if we have a schedule < 60 seconds (the default schedule_horizone for chronos)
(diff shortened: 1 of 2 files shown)


paasta_tools/chronos_tools.py
```diff
@@ -277,7 +277,13 @@ def check_schedule(self):
                                 'does not conform to the ISO 8601 format:\n%s' % (start_time, schedule, str(exc)))
 
             try:
-                isodate.parse_duration(interval)  # 'interval' and 'duration' are interchangeable terms
+                # 'interval' and 'duration' are interchangeable terms
+                parsed_interval = isodate.parse_duration(interval)
+                # until we make this configurable, throw an
```
mrtyler: Please move all of this stuff outside of the try block

paasta_tools/chronos_tools.py
```diff
@@ -277,7 +277,13 @@ def check_schedule(self):
...
+                # error if we have a schedule < 60 seconds (the default schedule_horizone for chronos)
+                # https://github.com/mesos/chronos/issues/508
+                if parsed_interval < datetime.timedelta(seconds=60):
+                    msgs.append('Unsupported interval "%s": jobs must be run at an interval of > 60seconds' % interval)
```
mrtyler: 60 seconds

mrtyler:
fix'n'ship

PR #62, opened 2015-11-18
add missing scribereader dependency when building internally

Rob-Johnson:
It was reported that ``paasta logs`` is broken internally because the ``scribereader`` dependency is missing. We knew it was broken externally, but the regeression seen internally was caused by [this commit](https://github.com/Yelp/paasta/commit/96a2a5733c47522140b922f323f5ba7330c5f7c3#diff-2eeaed663bd0d25b7e608891384b7298L52). 
 
We wrapped the import of ``scribereader`` in a ``try/except`` in [this commit](https://github.com/Yelp/paasta/commit/52cb946c895773f11488c8ee1c9b724bdd3e83a4) so the tests that hit it fail gracefully when scribereader is not available (as is the case when travis runs them). This meant we missed the error when the package was missing internally.

``scribereader`` is an internal dependency, so we have to do some magic to inject the dependency when running the ``make itest_*`` tasks used to build the deb we use internally.

The ``setup.py`` file that's been removed was an artifact from the first attempt to resolve this, and is unused.

The change on L:52 is needed because tox installs the public and private dependencies in two separate steps: without the added ``--allow-external scribereader``, it won't fallback to the private version installed on line 48 when it tries to install the dependencies in ``requirements.txt`` from public pypi.

This is all too much work to patch over the problem, so we should prioritise fixing this by removing the scribereader dependency.

@mittonk


yelp_package/Makefile
```diff
@@ -21,12 +21,16 @@ UID:=`id -u`
 GID:=`id -g`
 DOCKER_RUN_LUCID:=docker run -t -v  $(CURDIR)/../:/work:rw paastatools_lucid_container
 DOCKER_RUN_TRUSTY:=docker run -t -v  $(CURDIR)/../:/work:rw paastatools_trusty_container
+ADD_MISSING_DEPS:=-diff --unchanged-line-format= --old-line-format= --new-line-format='%L' ../requirements.txt ./extra_requirements.txt >> ../requirements.txt
```
Rob-Johnson: Heh, sure is. The leading dash tells make not to check the exit code - it would normally fail on non 0. We Have to do this because diff uses the exit code to indicate if there was a diff or not rather than if there was an error (TIL!).

nhandler:
Since you confirmed offline that the local tests worked as expected, go ahead and ship it :+1:

PR #65, opened 2015-11-19
Some refactoring related to the argument 'clusterinstance'

zeldinha:
With this refactor it should be easier to add future functionality that makes using this a lot more user friendly. Got rid of 'clusterinstance' and separated it into cluster and instance for clarity.

### DIFF
--- a/paasta_tools/paasta_cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/paasta_cli/cmds/mark_for_deployment.py
@@ -56,7 +56,8 @@ def add_subparser(subparsers):
 def build_command(
     upstream_git_url,
     upstream_git_commit,
-    clusterinstance,
+    cluster,
+    instance,
 ):
     """upstream_git_url is the Git URL where the service lives (e.g.
     git@git.yelpcorp.com:services/foo)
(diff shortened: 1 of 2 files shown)


solarkennedy:
I think this is good, and I'm happy to see "args" not be a function argument for more things.

paasta_tools/paasta_cli/cmds/mark_for_deployment.py
```diff
@@ -109,4 +113,21 @@ def paasta_mark_for_deployment(args):
...
             instance=instance,
         )
+    return returncode
+
+
+def paasta_mark_for_deployment(args):
+    """Wrapping mark_for_deployment"""
+    cluster, instance = args.clusterinstance.split('.')
+    service = args.service
+    if service and service.startswith('services-'):
```
Rob-Johnson: can you explain what this and the next line is doing?

solarkennedy: This that thing where jenkins prepends "services-" in as the service name arg

Rob-Johnson: ah ok, makes sense. could you add a comment to explain that? In the long term it'd be good to move this logic out of paasta and into jenkins (we could just sed the $upstream_job_name that Jenkins passes as an arg)

solarkennedy: I'll make @zeldinha handle that in the next pass

paasta_tools/paasta_cli/cmds/mark_for_deployment.py
```diff
@@ -64,42 +65,45 @@ def build_command(
     instancename is where you want to deploy. E.g. cluster1.canary indicates
     a Mesos cluster (cluster1) and an instance within that cluster (canary)
     """
-    cmd = 'git push -f %s %s:refs/heads/paasta-%s' % (
+    cmd = 'git push -f %s %s:refs/heads/paasta-%s.%s' % (
```
mrtyler: Hard-coded . is a mistake.

PR #66, opened 2015-11-19
fail gracefully if you try and run paasta logs without scribe existing

Rob-Johnson:
I'd like a better way of doing this - logic for dealing with missing scribereader is sprawled into a few places now. If anyone has a better suggestion, I'm open to discussion.

### DIFF
--- a/paasta_tools/paasta_cli/cmds/logs.py
+++ b/paasta_tools/paasta_cli/cmds/logs.py
@@ -26,6 +26,7 @@ import re
 import sys
 
 import isodate
+
 try:
     from scribereader import scribereader
     from scribereader.scribereader import StreamTailerSetupError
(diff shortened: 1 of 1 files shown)


mrtyler:
shipit

Where else is there special logic for this? I only see scribereader in logs.py and in associated tests (@pytest.mark.skipif goofiness), which is not too bad. I think adding the message where you did is ok. I agree that having it in one single canonical place would be better but I don't have brilliant ideas for how to do that while still swallowing ImportError in cases where we don't care about it.

Rob-Johnson:
well, there is already a try/catch ImportError/pass at the top of this file, 2 more try/catch ImportError/pass in test_cmd_logs.py and now this. I'll ship this as is - should only be temporary...right?

PR #67, opened 2015-11-20
execute docker pull before running container

Rob-Johnson:
We found an edge case where this failed because the image didn't already exist on the host travis was running the tests on - dockerpy doesn't pull an image if it doesn't already exist, so this addresses that.

### DIFF
--- a/general_itests/steps/paasta_execute_docker_command.py
+++ b/general_itests/steps/paasta_execute_docker_command.py
@@ -38,6 +38,7 @@ def create_docker_container(context, task_id):
         context.docker_client.remove_container(container_name, force=True)
     except APIError:
         pass
+    context.docker_client.pull('ubuntu:trusty')
     container = context.docker_client.create_container(
         name=container_name,
         image='ubuntu:trusty',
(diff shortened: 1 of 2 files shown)


general_itests/steps/paasta_execute_docker_command.py
```diff
@@ -38,6 +38,7 @@ def create_docker_container(context, task_id):
         context.docker_client.remove_container(container_name, force=True)
     except APIError:
         pass
+    context.docker_client.pull('ubuntu:trusty')
```
mrtyler: Uh, do you have to hardcode this name? This seems pretty bad.

solarkennedy:
ship

general_itests/steps/paasta_execute_docker_command.py
Rob-Johnson: hm, thats fair, I could change the feature so that the image name is parameterised.

Rob-Johnson:
r2: parameterised image name

mrtyler:
I didn't notice this was only in itest code but probably still a good idea :)

shipit

PR #68, opened 2015-11-20
Use dulwich for mark for deployment

solarkennedy:
This converts mark-for-deployment to use dulwich so you don't need a local checkout.
This will be necessary for non-jenkins use cases and will be handy for the "rollback" command.

Some extra code was necissary in the `remote_git` module to support this. I would like @EvanKrall to look extra carefully at that, it was my first time using dulwich.

Manual testing:
```
kwa@dev13-devc:~/Projects/devops (master) $ git branch -a -v
  remotes/origin/paasta-nova-devc.demo                1aaf828 devops demo 2

(py)kwa@dev13-devc:~/Projects/paasta (use_dulwich_for_mark_for_deployment) $ ./paasta_tools/paasta_cli/paasta_cli.py mark-for-deployment -u git@git.yelpcorp.com:services/devops -c 700bda5e32aae2345af581af29494ef095bd9666 -l nova-devc.demo -s devops
Marked 700bda5e32aae2345af581af29494ef095bd9666 in for deployment on demo in the nova-devc cluster

kwa@dev13-devc:~/Projects/devops (master) $ git branch -a -v  | grep nova-devc
  remotes/origin/paasta-nova-devc.demo                700bda5 Revert "Failing test to test email notifications"
```

### DIFF
--- a/paasta_tools/paasta_cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/paasta_cli/cmds/mark_for_deployment.py
@@ -22,7 +22,8 @@ import sys
 from paasta_tools.paasta_cli.utils import get_jenkins_build_output_url
 from paasta_tools.paasta_cli.utils import validate_service_name
 from paasta_tools.utils import _log
-from paasta_tools.utils import _run
+from paasta_tools.utils import get_paasta_branch
+from paasta_tools import remote_git
 
 
 def add_subparser(subparsers):
--- a/paasta_tools/paasta_cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/paasta_cli/cmds/mark_for_deployment.py
@@ -19,7 +19,6 @@ deployment to a cluster.instance.
 
 import sys
 
-from paasta_tools.paasta_cli.utils import get_jenkins_build_output_url
 from paasta_tools.paasta_cli.utils import validate_service_name
 from paasta_tools.utils import _log
 from paasta_tools.utils import get_paasta_branch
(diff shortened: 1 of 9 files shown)


paasta_tools/paasta_cli/cmds/mark_for_deployment.py
```diff
@@ -53,57 +53,23 @@ def add_subparser(subparsers):
...
-        cluster=cluster,
-        instance=instance
+    remote_branch = get_paasta_branch(cluster=cluster, instance=instance)
+    ref_mutator = remote_git.make_force_push_mutate_refs_func(
+        target_branches=[remote_branch],
+        sha=commit,
     )
+    try:
+        remote_git.create_remote_refs(git_url=git_url, ref_mutator=ref_mutator, force=True)
+        loglines = ["Marked %s in for deployment on %s in the %s cluster" % (commit, instance, cluster)]
```
EvanKrall: put this and the next line into an else block after the except block.

solarkennedy: Oh right of course

solarkennedy: Wait, why? If everything goes right then this path is executed, if anything goes wrong we get the other path

EvanKrall:
Looks good. Maybe consider putting the definition of `make_force_push_mutate_refs_func` closer to where you use it, since you're the only user of it. You could define it inside of `mark_for_deployment` and not need the make_ function at all, but that does make it harder to test.

paasta_tools/paasta_cli/cmds/mark_for_deployment.py
EvanKrall: Unless you get an exception on line 65 or 66, in which case you'll get a misleading message.

PR #69, opened 2015-11-20
Mrtyler rejigger status

mrtyler:
This addresses some concerns brought up by carroux way back in review:115197:
- paasta level "started/stopped" (i.e. the result of "paasta stop/start") is only accurate at the instance level. It is wrong and misleading to display it for each config
- "enabled/disabled" is not very clear

After some thinking and some consultation with mfield, here's what I did:
- paasta level "started/stopped" is reported only at the top level (parallel with git sha), and no longer as part of individual configs
- Change "enabled/disabled" to "scheduled/not scheduled"
-- mfield noticed that there are sort of three states here: enabled, disabled because the whole service is disabled, disabled because this is an old version that we're keeping around for history. We couldn't think of a great way to communicate those three states and differentiating between the latter two would require additional refactoring, so I just went with "scheduled/not scheduled".

Screenshot:

### DIFF
--- a/paasta_tools/chronos_serviceinit.py
+++ b/paasta_tools/chronos_serviceinit.py
@@ -79,11 +79,8 @@ def get_short_task_id(task_id):
 
 
 def _format_config_hash(job):
-    config_hash = PaastaColors.red("UNKNOWN")
-    job_id = job.get("name", None)
-    if job_id:
-        (_, __, ___, config_hash) = chronos_tools.decompose_job_id(job_id)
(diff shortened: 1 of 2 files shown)


mrtyler:
I have a small window left before I'm out of here, so I'm merging this early to build it and try it out tonight. Please still review and comment.

solarkennedy:
This is good, thank you for continuing to improve the output of this. Lots of people (eventually) will be running this command a lot I think.

PR #71, opened 2015-11-20
Added automatic changelog

solarkennedy:
This adds a generated changelog and a step in `make release` to generate it.

I only start looking at tags *after* we started pushing to github, as before that we don't really have metadata to scrape.


Rob-Johnson:
shipit

PR #73, opened 2015-11-20
Make local-run default to use bash if set to be interactive

solarkennedy:
I this a small step towards ad-hoc "batches", I think a large portion will of the use cases will want to just get an interactive shell, so this makes it a little bit easier as you will be able to just say "-I"


Rob-Johnson:
there is an assumption here that 'bash' is always available. can this default to 'sh' instead?

solarkennedy:
Eh... I would rather not. At Yelp we install bash on every base image, if people run this and it *isn't* available, it will just say bash is not available and they will have to figure out what shell they want to run.

Rob-Johnson:
k. shipit

PR #74, opened 2015-11-20
Add emergency scale feature

dichiarafrancesco:
With this change your service can be scaled up or down without the need to restart the other dockers running.

Motivation:

I think this feature is very important to have. Very often many services need more resources just for a short period of time without restarting the service. Also for those services that keep states while running is truly important adding more resources don't loosing the state. I know that this command does not permanently scale the service and the next time the service is updated (config change, deploy, bounce, etc.), those settings will override the emergency scale in fact I called it "emergency scale".


paasta_tools/marathon_serviceinit.py
```diff
@@ -71,6 +71,19 @@ def restart_marathon_job(service, instance, app_id, normal_instance_count, clien
     start_marathon_job(service, instance, app_id, normal_instance_count, client, cluster)
 
 
+def scale_marathon_job(service, instance, app_id, delta, client, cluster):
+    name = PaastaColors.cyan(compose_job_id(service, instance))
+    _log(
+        service=service,
+        line="EmergencyScale: Scaling %s %s to %d instances" % (name, 'down' if delta < 0 else 'up', abs(int(delta))),
```
tsheasha: In case the delta the delta is 5 and the current number of instances is 20, on a scale up this will read : 
"Scaling up to 5 instances", while it should read, "Scaling up *by* 5 instances".

dichiarafrancesco: right!

Rob-Johnson:
Can you add some unit and integration tests for your code please? There are examples in the ``test`` and ``paasta_itests`` directory.

paasta_tools/paasta_cli/cmds/emergency_scale.py
```diff
@@ -0,0 +1,80 @@
...
+    )
+    status_parser.add_argument(
+        '-y', '--yelpsoa-config-root',
+        default=DEFAULT_SOA_DIR,
+        required=False,
+        help="Path to root of yelpsoa-configs checkout",
+    )
+    status_parser.add_argument(
+        '--delta',
+        default=0,
```
Rob-Johnson: I don't think this should have a default - being required + having a default are conflicting.

paasta_tools/paasta_cli/cmds/emergency_scale.py
```diff
@@ -0,0 +1,80 @@
...
+        help="Instance of the service that you want to scale. Like 'main' or 'canary'.",
+        required=True,
+    ).completer = lazy_choices_completer(list_instances)
+    status_parser.add_argument(
+        '-c', '--cluster',
+        help="The PaaSTA cluster that has the service instance you want to scale. Like 'norcal-prod'.",
+        required=True,
+    ).completer = lazy_choices_completer(list_clusters)
+    status_parser.add_argument(
+        '-a', '--appid',
```
Rob-Johnson: I've seen that this is in emergency-stop too, so I understand why you've added this. I don't think it should be there though, we make an effort to hide this kind of detail from the user. @solarkennedy is this just an artifact of old code or did we keep it around just for convenience?https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_serviceinit.py#L316-L318 suggests we don't need it.

paasta_tools/paasta_cli/utils.py
```diff
@@ -436,12 +436,17 @@ def run_paasta_serviceinit(subcommand, master, service, instancename, cluster, *
         app_id_flag = "--appid %s " % kwargs['app_id']
     else:
         app_id_flag = ''
-    command = 'ssh -A -n %s sudo paasta_serviceinit %s%s%s %s' % (
+    if 'delta' in kwargs and kwargs['delta']:
+        delta = "--delta %s" % kwargs['delta']
+    else:
+        delta = ''
+    command = 'ssh -A -n %s sudo paasta_serviceinit %s%s%s %s %s' % (
```
Rob-Johnson: I'm not sure about this - ``emergency_scale`` is a marathon only tool, but ``paasta_serviceinit`` doesn't have any validation to prevent it being called on a Chronos job. This doesn't map to anything useful for chronos, so we need to add some validation somewhere for this. I'm not sure where is the best place right now though.

solarkennedy: It is true, but kinda an artifact of paasta_servceinit being the entrypoint for every type of emergency remote command. I think that is ok.

paasta_tools/paasta_cli/cmds/emergency_scale.py
solarkennedy: No, this is a necessary argument to allow people to kill *specific* apps (or now, scale?) in emergency situations. Specifically in the case of a bounce gone wrong, where bounces do not progress because they are stuck.

paasta_tools/paasta_cli/cmds/emergency_scale.py
dichiarafrancesco: right. :S

paasta_tools/paasta_cli/cmds/emergency_scale.py
dichiarafrancesco: so being this feature an emergency one, i'll leave the code as it is, ok?

Rob-Johnson: yeah, thats fine. :+1:

solarkennedy:
Make sure tests pass then ship.

PR #78, opened 2015-11-23
update dh-virtualenv location

Rob-Johnson:
the link to dh-virtualenv was 404ing (I guess they removed the old version when they released 0.10-0~81) and causing Jenkins + Travis to fail. This fixes that (Travis should pass tests now?).


zeldinha:
Is this not going to fail if they bump the version again?

solarkennedy:
Ah, that was lame of me to hard-code the raw url. Thank you for fixing this properly.

PR #80, opened 2015-11-23
don't use vars() to check for scribereader

Rob-Johnson:


paasta_tools/paasta_cli/cmds/logs.py
```diff
@@ -31,7 +31,7 @@
     from scribereader import scribereader
     from scribereader.scribereader import StreamTailerSetupError
 except ImportError:
-    pass
+    scribereader = None
```
mittonk: That's an improvement.

paasta_tools/paasta_cli/cmds/logs.py
```diff
@@ -510,7 +510,7 @@ def tail_paasta_logs(service, levels, components, clusters, raw_mode=False):
 def paasta_logs(args):
     """Print the logs for as Paasta service.
     :param args: argparse.Namespace obj created from sys.args by paasta_cli"""
-    if 'scribereader' not in vars():
+    if scribereader is None:
```
mittonk: Much less scary.

However, still lacking automated tests; might be time to start considering those for this functionality.

mittonk:
Better, but still not enough tests.
:shipit:

PR #88, opened 2015-11-24
Enabled local-run to work on docker images on registries instead of building locally

solarkennedy:
This is more of a RFC, this is not complete yet.

This adds new mutually exclusive args to `local-run`: `--build` and `--pull`. Eventually it will (try to) auto-detect to make it require fewer command line args for day-to-day use.

### DIFF
--- a/paasta_tools/paasta_cli/cmds/local_run.py
+++ b/paasta_tools/paasta_cli/cmds/local_run.py
@@ -40,6 +40,7 @@ from paasta_tools.paasta_cli.utils import list_instances
 from paasta_tools.paasta_cli.utils import list_services
 from paasta_tools.utils import get_default_cluster_for_service
 from paasta_tools.utils import get_docker_host
+from paasta_tools.utils import get_docker_url
 from paasta_tools.utils import get_username
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import load_system_paasta_config
@@ -255,6 +256,27 @@ def add_subparser(subparsers):
         dest='yelpsoa_config_root',
         help='A directory from which yelpsoa-configs should be read from',
         default=service_configuration_lib.DEFAULT_SOA_DIR
+    )
+    build_pull_group = list_parser.add_mutually_exclusive_group()
+    build_pull_group.add_argument(
+        '-b', '--build',
+        help=("Build the docker image to run from scratch using the local Makefile's ",
+              "'cook-image' target. Defaults to try to use the local Makefile if present. ",
+              "otherwise local-run will pull and run the Docker image that is marked for ",
+              "deployment in the Docker registry."),
+        required=False,
+        action='store_true',
+        default=None,
+    )
+    build_pull_group.add_argument(
+        '-p', '--pull',
+        help=("Pull the docker image marked for deployment from the Docker registry and ",
+              "use that for the local-run. This is the opposite of --build. Defaults to ",
+              "autodetect a Makefile, if present will not pull, and instead assume that ",
+              "a local build is desired."),
+        required=False,
+        action='store_true',
+        default=None,
     )
     list_parser.add_argument(
         '-C', '--cmd',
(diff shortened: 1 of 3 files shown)


solarkennedy:
This is ready for review.

paasta_tools/paasta_cli/cmds/local_run.py
```diff
@@ -574,40 +619,42 @@ def configure_and_run_docker_container(docker_client, docker_hash, service, args
...
-        )
-        sys.exit(1)
-    if not os.path.isfile(os.path.join(os.getcwd(), 'Dockerfile')):
-        sys.stderr.write(
-            'ERROR: No Dockerfile in the current directory.\n'
-            'Are you in the root folder of the service directory? Does a Dockerfile exist?\n'
-        )
-        sys.exit(1)
+def local_makefile_present():
+    return True
```
mfield: this seems like a strong assumption

solarkennedy: Heheh, whoops, ty

paasta_tools/paasta_cli/cmds/local_run.py
```diff
@@ -574,40 +619,42 @@ def configure_and_run_docker_container(docker_client, docker_hash, service, args
...
 
 
 def paasta_local_run(args):
-    validate_environment()
+    # Try to autodetect if we should build the image locally or not
```
mfield: I think you only need one boolean flag here. Having both --pull and --build is confusing. If I read correctly, the only thing --build changes is whether the "Local Makefile detected..." line is printed to stderr.

paasta_tools/paasta_cli/cmds/local_run.py
```diff
@@ -574,40 +619,42 @@ def configure_and_run_docker_container(docker_client, docker_hash, service, args
...
-
     docker_client = Client(base_url=base_docker_url)
 
-    default_tag = 'paasta-local-run-%s-%s' % (service, get_username())
-    tag = os.environ.get('DOCKER_TAG', default_tag)
-    os.environ['DOCKER_TAG'] = tag
-
-    paasta_cook_image(None, service=service, soa_dir=args.yelpsoa_config_root)
+    if build:
+        default_tag = 'paasta-local-run-%s-%s' % (service, get_username())
```
mfield: Should we enhance get_username to look for an environment variable like SUDO_USER so it's easier to find out who to bug if a container starts to do evil things to a machine?

paasta_tools/paasta_cli/cmds/local_run.py
solarkennedy: I disagree. I want the flags so you can always be explicit about the behavior you want to override the auto-detect. The --build flag will (should?) tell the rest of the code to build the image (it runs `paasta_cook_image`)

paasta_tools/paasta_cli/cmds/local_run.py
solarkennedy: Hmmm. Yes I think so.

paasta_tools/paasta_cli/cmds/local_run.py
mfield: This is my understanding of the current implementation:
- `local-run --pull` will always set `build=False`
- `local-run --build` will set `build=True` unless you also specify `--pull`, which has a higher priority
- `local-run` with neither argument will always set `build=True` and will print "Local Makefile detected..." if a Makefile is detected.

Having two flags makes more sense if the first clause reads more like this, otherwise `--build` is always the default unless `--pull` is specified.
```python
if args.build is None and args.pull is None:
    build = local_makefile_present()
```

solarkennedy: Ah, I agree that is a cleaner way to represent this.

paasta_tools/paasta_cli/cmds/local_run.py
solarkennedy: This is done now in the get_username function.

mfield: :+1:

mfield:
:sheep:

PR #90, opened 2015-11-25
Only allow the itest_% Makefile target to run at Yelp

nhandler:
This target will currently fail outside of Yelp. Use the machine's
FQDN to determine if it is a Yelp box. If it is, run the target.
If it is not, simply display a warning.

This should hopefully help avoid confusion as to why the target fails to work.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Set ENV to 'YELP' if FQDN ends in '.yelpcorp.com'
+# Otherwise, set ENV to the FQDN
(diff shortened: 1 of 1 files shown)


Rob-Johnson:
I'd like to be able to build on my laptop - can you add a way for me to set an env var and get round this hostname checking please?

nhandler:
@Rob-Johnson The new commit I pushed allows you run run 'PAASTA_ENV=YELP make itest_trusty' and have it work (for boxes that don't have .yelpcorp.com in the FQDN).

Rob-Johnson:
shipit

PR #92, opened 2015-11-26
bump scribereader version in line with aed1812b917daba17a5fd8f1a0fe9b…

Rob-Johnson:
this is why builds are broken atm


zeldinha:
:shipit: says 'shipit'

PR #93, opened 2015-11-26
remove hardcoded docker registry location

Rob-Johnson:
I made an unhelpful comment on another review of 'we should remove this'. So here it is.


solarkennedy:
ship

solarkennedy:
I'm going to merge this as I need it.

PR #99, opened 2015-11-30
Paasta rollback now accepts none, one or a list of instances.

zeldinha:
Paasta rollback now accepts none, one or a list of instances. If any are wrong it will give a warning and carry on with the valid ones.

Added a bunch of tests to verify the behaviour on each case.

### DIFF
--- a/paasta_tools/paasta_cli/cmds/rollback.py
+++ b/paasta_tools/paasta_cli/cmds/rollback.py
@@ -25,6 +25,7 @@ from paasta_tools.paasta_cli.utils import list_instances
 from paasta_tools.paasta_cli.cmds.mark_for_deployment import mark_for_deployment
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import list_all_instances_for_service
+from paasta_tools.utils import PaastaColors
 
 
 def add_subparser(subparsers):
@@ -34,35 +35,35 @@ def add_subparser(subparsers):
         help='Rollback a docker image to a previous deploy')
 
     list_parser.add_argument('-k', '--commit',
-                             help='Git sha to mark for rollback',
+                             help='Git SHA to mark for rollback',
                              required=True,
                              )
-    list_parser.add_argument('-i', '--instance',
-                             help='Mark the instance[s] we want to roll back (e.g. '
-                             'canary, main)',
+    list_parser.add_argument('-i', '--instances',
+                             help='Mark one or more instances to roll back (e.g. '
+                             '"canary", "canary,main"). If no instances specified,'
+                             ' all instances for that service are rolled back',
+                             default='',
                              required=False,
                              ).completer = lazy_choices_completer(list_instances)
     list_parser.add_argument('-c', '--cluster',
-                             help='Mark the cluster we want to rollback (e.g. '
-                             'cluster1, cluster2)',
+                             help='Mark the cluster to rollback (e.g. '
+                             'cluster1)',
                              required=True,
                              ).completer = lazy_choices_completer(list_clusters)
     list_parser.add_argument('-s', '--service',
-                             help='Name of service you wish to rollback'
+                             help='Name of the service to rollback (e.g. '
+                             'service1)'
                              ).completer = lazy_choices_completer(list_services)
 
     list_parser.set_defaults(command=paasta_rollback)
 
 
-def validate_given_instances(service, args_instances):
+def validate_given_instances(service_instances, args_instances):
     """Trying to find which instances are actually valid for the given service"""
-    service_instances = list_all_instances_for_service(service)
-    invalid_instances = None
-
-    if not args_instances:
-        valid_instances = service_instances
+    if len(args_instances) is 0:
+        valid_instances = set(service_instances)
+        invalid_instances = set([])
     else:
-        args_instances = args_instances.split(",")
         valid_instances = set(args_instances).intersection(service_instances)
         invalid_instances = set(args_instances).difference(service_instances)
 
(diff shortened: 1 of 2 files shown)


paasta_tools/paasta_cli/cmds/rollback.py
```diff
@@ -53,22 +54,47 @@ def add_subparser(subparsers):
     list_parser.set_defaults(command=paasta_rollback)
 
 
+def validate_given_instances(service, args_instances):
+    """Trying to find which instances are actually valid for the given service"""
+    service_instances = list_all_instances_for_service(service)
+    invalid_instances = None
```
Rob-Johnson: use an empty list rather than None here

paasta_tools/paasta_cli/cmds/rollback.py
```diff
@@ -53,22 +54,47 @@ def add_subparser(subparsers):
...
+
+    if not args_instances:
```
Rob-Johnson: when is this condition true? isn't args_instances going either be 'main' or some other user input?

paasta_tools/paasta_cli/cmds/rollback.py
```diff
@@ -53,22 +54,47 @@ def add_subparser(subparsers):
     list_parser.set_defaults(command=paasta_rollback)
 
 
+def validate_given_instances(service, args_instances):
+    """Trying to find which instances are actually valid for the given service"""
+    service_instances = list_all_instances_for_service(service)
```
Rob-Johnson: I think you should move this logic to the caller; it'd be better if this was just a wrapper function around the set intersection/difference.

paasta_tools/paasta_cli/cmds/rollback.py
zeldinha: The argument is not mandatory anymore, so if empty it would pick all the instances of the service.

paasta_tools/paasta_cli/cmds/rollback.py
```diff
@@ -53,22 +54,47 @@ def add_subparser(subparsers):
...
-            git_url=git_url,
-            cluster=cluster,
-            instance=instance,
-            service=service,
-            commit=commit
-        )
+        instances, invalid = validate_given_instances(service, args.instance)
+
+        if invalid:
+            print "WARNING: These instances are not valid and will not be deployed: %s.\n" % (",").join(invalid)
```
Rob-Johnson: can you wrap this in PaastaColors.yellow ?

paasta_tools/paasta_cli/cmds/rollback.py
Rob-Johnson: but you default to 'main' in the argparser, right?

Rob-Johnson: woops, I misread your argparse code. ignore :)

paasta_tools/paasta_cli/cmds/rollback.py
```diff
@@ -37,9 +38,9 @@ def add_subparser(subparsers):
                              required=True,
                              )
     list_parser.add_argument('-i', '--instance',
```
solarkennedy: can you make this "instances"? In python you can abbreviate it too, so you can just say --instance and it will still work.

paasta_tools/paasta_cli/cmds/rollback.py
```diff
@@ -37,9 +38,9 @@ def add_subparser(subparsers):
...
-                             help='Mark the instance we want to roll back (e.g. '
-                             'canary, .main)',
-                             required=True,
+                             help='Mark the instance[s] we want to roll back (e.g. '
+                             'canary, main)',
```
solarkennedy: Can you remove the space to make it more obvious it is a csv? Also can you add more words to describe that you can describe a single instance or mulitple instance in csv form. Also can you describe the default behavior? Helpstrings are cheap and good.

paasta_tools/paasta_cli/cmds/rollback.py
```diff
@@ -53,22 +54,47 @@ def add_subparser(subparsers):
...
-            commit=commit
-        )
+        instances, invalid = validate_given_instances(service, args.instance)
+
+        if invalid:
+            print "WARNING: These instances are not valid and will not be deployed: %s.\n" % (",").join(invalid)
+
+        if instances:
+            for instance in instances:
+                print "Deploying %s to %s.%s.\n" % (service, cluster, instance)
```
solarkennedy: This print line is kinda misleading. Can you remove it and let the normal stdout of the mark_for_deployment function show through?

paasta_tools/paasta_cli/cmds/rollback.py
```diff
@@ -53,22 +54,47 @@ def add_subparser(subparsers):
...
+                returncode = mark_for_deployment(
+                    git_url=git_url,
+                    cluster=cluster,
+                    instance=instance,
+                    service=service,
+                    commit=commit
```
solarkennedy: trailing comma please

zeldinha:
I've done most of the suggested changes, thanks for the feedback! Please let me know if there's anything else that could be improved :)

paasta_tools/paasta_cli/cmds/rollback.py
```diff
@@ -33,44 +35,70 @@ def add_subparser(subparsers):
...
-                             help='Name of service you wish to rollback'
+                             help='Name of the service to rollback (e.g. '
+                             'service1)'
                              ).completer = lazy_choices_completer(list_services)
 
     list_parser.set_defaults(command=paasta_rollback)
 
 
+def validate_given_instances(service_instances, args_instances):
+    """Trying to find which instances are actually valid for the given service"""
```
Rob-Johnson: Try and be more specific in your docstrings. Here's an example:

Given two lists of instances, return the intersection and difference of those two lists.

:param service_instances: the list of instances belonging to a services
:param args_instances: the desired instances
:returns: a tuple of (common, difference), indicating instances common to the two lists, and those only in args_instances

Rob-Johnson:
fix and :ship:

solarkennedy:
ship and iterate

PR #101, opened 2015-12-01
Bump max_consecutive_failures to 30 (5 minutes)

solarkennedy:
After our recent outage, I think 1 minute is too aggressive. This bumps it to 5 minutes. This reduces our overall churn rate and gives more breathing room to the distributed system to recover from network issues.


Rob-Johnson:
shipit

solarkennedy:
Ping @EvanKrall  for a second opinion. This is a small change but it drastically changes the rate of flux of services when they are healthy, and "slows" down the reaction time to marathon when things fail.

PR #102, opened 2015-12-01
Make backoff_seconds work as a function of the instance count.

solarkennedy:
The docstring should be explanatory enough.


Rob-Johnson:
shipit

solarkennedy:
Ping @EvanKrall for a second opinion on this one too. This change is a "noop" for services that have 10 instances now, but will significantly effect the exponential backoff behavior of small and large instance count services, but hopefully in a good way.

PR #104, opened 2015-12-02
Make local-run stream the docker-pull with /dev/null as stdin

solarkennedy:
This is to improve the new adhoc batch stuff.

Without devnull as stdin, it just hangs expecting you to type in a username+password. I'm going to make the call that we never want to do this in paasta?

Also I added a small flag to allow streaming the output, so you *see* the pull so you get better feedback as it is being pulled. (instead of waiting with no output)


solarkennedy:
This is ready for review. It should not impact our other _run invocations that need interactive input.

Rob-Johnson:
shipit

solarkennedy:
Hmm... apparently we are trying to log to scribe in our _run invocation in local-run in the general itest.....

PR #105, opened 2015-12-02
Autodetect the instance in local-run

solarkennedy:
Adds new guess_cluster and guess_instance functions for local-run.

Defaulting to 'main' was an ok start, but not everything will have main. (scheduled-task only services)


Rob-Johnson:
we already have ``guess_service`` in [paasta_cli.utils](https://github.com/Yelp/paasta/blob/master/paasta_tools/paasta_cli/utils.py#L281) - could you keep all these together?

solarkennedy:
can-do

Rob-Johnson:
fix n ship

PR #107, opened 2015-12-03
Add 'paasta validate' command

nhandler:
Here is a first pass at a `paasta validate` command. It currently only does jsonschema validation against the marathon and chronos files and reports whether they are valid or not.


paasta_tools/paasta_cli/cmds/validate.py
```diff
@@ -0,0 +1,99 @@
...
+        for file_type in ['chronos', 'marathon']:
+            if basename.startswith(file_type):
+                validate_schema(basename, file_type)
+
+
+def add_subparser(subparsers):
+    validate_parser = subparsers.add_parser(
+        'validate',
+        description="Execute 'paasta validate' from service repo root",
+        help="Validate that all paasta config files in pwd are correct")
```
solarkennedy: Is this really correct? (pwd)
The way I read it, it looks like it looks in the hard-coded location of '/nail/etc/services' ?

paasta_tools/paasta_cli/cmds/validate.py
```diff
@@ -0,0 +1,99 @@
...
+    validate_parser.set_defaults(command=paasta_validate)
+
+
+def paasta_validate(args):
+    """Analyze the service in the PWD to determine if conf files are all valid
+    :param args: argparse.Namespace obj created from sys.args by paasta_cli"""
+    service = guess_service_name()
+    service_path = os.path.join('/nail/etc/services', service)
```
solarkennedy: I'm down with incremental value, and I'm if you ship this as a very first pass, but this isn't very useful as-is. 

It sounds like the first functionality you are implementing is a normal cli tool  to validate in an interactive way, but by hard-coding a path, it requires users to literally commit changes, waiting, and then running validate *after*. 

Can you use a -soaconfigs arg like many other of our cli tools to make it a bit more useful off the bat? Also can you not hard-code the /nail/etc/services string? (use the constant instead, s_c_l something)

paasta_tools/paasta_cli/utils.py
```diff
@@ -540,3 +540,45 @@ def get_jenkins_build_output_url():
...
+        "Failed to validate schema. More info:", "http://y/yelpsoa-configs")
+
+
+class PaastaValidateSchemas:
+
+    """Collection of jsonschemas used by 'paasta validate'.
+    Helpful as it avoids cumbersome maintenance of the unit tests.
+    """
+
+    MARATHON_SCHEMA = {
```
solarkennedy: I understand the desire to have them in python, but please do just normal json for these json schemas. I think that will be least surprising and makes it easier to interact with other jsonschema tools.

One of the things that helped me write the first schema was using an online interactive json schema tool, and this would be a bit harder to do if it wasn't *actually* json.

paasta_tools/paasta_cli/utils.py
```diff
@@ -540,3 +540,45 @@ def get_jenkins_build_output_url():
...
+
+class PaastaValidateMessages:
+
+    """Collection of messages printed out by 'paasta validate'.
+    Helpful as it avoids cumbersome maintenance of the unit tests.
+    """
+
+    SCHEMA_VALID = success("Successfully validated schema")
+    SCHEMA_INVALID = failure(
+        "Failed to validate schema. More info:", "http://y/yelpsoa-configs")
```
solarkennedy: No new y/links in new code please

paasta_tools/paasta_cli/cmds/validate.py
```diff
@@ -0,0 +1,99 @@
...
+from jsonschema import Draft4Validator
+from jsonschema import FormatChecker
+from jsonschema import ValidationError
+from paasta_tools.paasta_cli.utils import guess_service_name
+from paasta_tools.paasta_cli.utils import PaastaValidateMessages
+from paasta_tools.paasta_cli.utils import PaastaValidateSchemas
+
+
+def get_file_contents(path):
+    """Open a file for reading
```
Rob-Johnson: docstring format https://www.python.org/dev/peps/pep-0008/#documentation-strings

paasta_tools/paasta_cli/cmds/validate.py
```diff
@@ -0,0 +1,99 @@
...
+
+    :param path: path of file to read"""
+    return open(path).read()
+
+
+def get_schema(file_type):
+    """Get the correct schema to use for validation
+
+    :param file_type: what schema type should we validate against"""
```
Rob-Johnson: again https://www.python.org/dev/peps/pep-0008/#documentation-strings

paasta_tools/paasta_cli/cmds/validate.py
```diff
@@ -0,0 +1,99 @@
...
+    schema_name = '%s_SCHEMA' % file_type.upper()
+    return getattr(PaastaValidateSchemas, schema_name)
+
+
+def validate_schema(file_path, file_type):
+    """Check if the specified config file has a valid schema
+
+    :param file_path: path to file to validate
+    :param file_type: what schema type should we validate against"""
```
Rob-Johnson: again https://www.python.org/dev/peps/pep-0008/#documentation-strings

paasta_tools/paasta_cli/utils.py
```diff
@@ -540,3 +540,45 @@ def get_jenkins_build_output_url():
     if build_output:
         build_output = build_output + 'console'
     return build_output
+
+
+class PaastaValidateMessages:
```
Rob-Johnson: why are these in utils? they are only ever going to be used by paasta validate.

Rob-Johnson: also, why are these in a class? these can just be constants for the validate module.

paasta_tools/paasta_cli/utils.py
Rob-Johnson: +1

paasta_tools/paasta_cli/utils.py
```diff
@@ -540,3 +540,45 @@ def get_jenkins_build_output_url():
...
+    """Collection of messages printed out by 'paasta validate'.
+    Helpful as it avoids cumbersome maintenance of the unit tests.
+    """
+
+    SCHEMA_VALID = success("Successfully validated schema")
+    SCHEMA_INVALID = failure(
+        "Failed to validate schema. More info:", "http://y/yelpsoa-configs")
+
+
+class PaastaValidateSchemas:
```
Rob-Johnson: again, these schemas are very specific to the validate module, so shouldn't be in utils.py + don't really belong in a class, either.

paasta_tools/paasta_cli/cmds/validate.py
```diff
@@ -0,0 +1,99 @@
...
+from paasta_tools.paasta_cli.utils import guess_service_name
+from paasta_tools.paasta_cli.utils import PaastaValidateMessages
+from paasta_tools.paasta_cli.utils import PaastaValidateSchemas
+
+
+def get_file_contents(path):
+    """Open a file for reading
+
+    :param path: path of file to read"""
+    return open(path).read()
```
Rob-Johnson:
```
with open(path) as p:
   return p.read()
```

otherwise you don't close the file.

Rob-Johnson: can you also add some validation for a missing file?

tests/paasta_cli/test_cmds_validate.py
```diff
@@ -0,0 +1,130 @@
...
+    # Ensure each check in 'paasta_validate' is called
+
+    mock_guess_service_name.return_value = 'servicedocs'
+    paasta_validate(None)
+
+    assert mock_validate_all_schemas.called
+
+
+def test_get_schema_marathon_found():
+    get_schema('marathon')
```
Rob-Johnson: you're making no assertions here other than that it doesn't throw an exception - can you verify that the thing returned is what you expect?

tests/paasta_cli/test_cmds_validate.py
```diff
@@ -0,0 +1,130 @@
...
+  cmd: virtualenv_run/bin/python adindexer/adindex_worker.py
+  healthcheck_mode: cmd
+main_http:
+  cpus: 0.1
+  instances: 2
+  mem: 250
+"""
+    mock_get_file_contents.return_value = marathon_content
+
+    validate_schema('unused_service_path.yaml', 'marathon')
```
Rob-Johnson: can you make an assertion about this?

paasta_tools/paasta_cli/cmds/validate.py
nhandler: I can probably just do away with the path completely. If we expect that people will run `paasta validate` from within the same directory as their service, that should be fine. This would make it easier to validate code without having to commit changes and wait.

paasta_tools/paasta_cli/cmds/validate.py
nhandler: Sure. As an FYI, this function was from paasta_tools/paasta_cli/cmds/check.py . I guess I should probably extract it so we don't have 2 copies (and then both commands will properly close the file).

paasta_tools/paasta_cli/cmds/validate.py
```diff
@@ -0,0 +1,120 @@
...
+        print '  Validation Message: %s' % e.message
+    else:
+        print '%s: %s' % (SCHEMA_VALID, file_path)
+
+
+def validate_all_schemas(service_path):
+    """Finds all recognized config files in service directory,
+    and validates their schema.
+
+    :param service_path: path to location of configuration files"""
```
Rob-Johnson: docstring style - """ in wrong place

paasta_tools/paasta_cli/cmds/validate.py
```diff
@@ -0,0 +1,120 @@
...
+    validate_parser = subparsers.add_parser(
+        'validate',
+        description="Execute 'paasta validate' from service repo root",
+        help="Validate that all paasta config files in pwd are correct")
+    validate_parser.set_defaults(command=paasta_validate)
+
+
+def paasta_validate(args):
+    """Analyze the service in the PWD to determine if conf files are all valid
+    :param args: argparse.Namespace obj created from sys.args by paasta_cli"""
```
Rob-Johnson: docstring

paasta_tools/paasta_cli/schemas/chronos_schema.json
```diff
@@ -0,0 +1,11 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "description": "http://paasta.readthedocs.org/en/latest/yelpsoa_configs.html#chronos-clustername-yaml",
+    "type": "object",
+    "minProperties": 1,
+    "additionalProperties": {
+        "type": "object",
+        "properties": {},
```
Rob-Johnson: is this correct? we're not actually doing any validation on the properties?

paasta_tools/paasta_cli/schemas/marathon_schema.json
```diff
@@ -0,0 +1,11 @@
+{
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "description": "http://paasta.readthedocs.org/en/latest/yelpsoa_configs.html#marathon-clustername-yaml",
+    "type": "object",
+    "minProperties": 1,
+    "additionalProperties": {
+        "type": "object",
+        "properties": {},
```
Rob-Johnson: same question about properties validation?

paasta_tools/paasta_cli/cmds/validate.py
```diff
@@ -0,0 +1,120 @@
...
+    service_path = os.getcwd()
```
Rob-Johnson: I think kwa's first comment is still valid - you should accept a --soa-dir param that this uses (though it might be ok to default to pwd?)

paasta_tools/paasta_cli/schemas/chronos_schema.json
nhandler: I stole the schemas from our existing Gitolite checks. I am fine modifying them, but I would like to do that as a separate issue after this one is resolved.

paasta_tools/paasta_cli/cmds/validate.py
nhandler: I still fail to see the use case for this. The workflow I envisioned was a user is editing their service's soa configs and wants to make sure they are valid. So they run 'paasta validate', make any necessary changes, and repeat. This workflow does not require anything to be committed or any waiting.

The only scripts in paasta_tools/paasta_cli/cmds that actually support specifying yelpsoa-config-root are cook_image, emergency_scale, fsm, and local_run. Notice that `paasta check`, which is quite similar to `paasta validate` is not in this list.

Since it is pretty clear that the rest of the team feels this change would be useful, I will make it. However, I would love it if someone could describe in a bit more detail how they see the --soa-dir argument being used in practice and why this differs from the `paasta check` workflow.

nhandler:
Hold off on reviewing this last revision. Just realized I forgot to pass properly pass the soa_dir through all of the helper methods.

nhandler:
The previous commit should make the soa_dir work properly. It is also worth noting that I have another pull request I will make once this one is merged to add validation for the chronos schedule.

paasta_tools/paasta_cli/cmds/validate.py
```diff
@@ -0,0 +1,165 @@
...
+
+    :param args: argparse.Namespace obj created from sys.args by paasta_cli
+    """
+
+    service = args.service
+    soa_dir = args.yelpsoa_config_root
+    service_path = get_service_path(service, soa_dir)
+
+    if service_path is None:
+        return 1
```
solarkennedy: can you confirm this results in an actual exit code of 1 in the shell?

paasta_tools/paasta_cli/schemas/chronos_schema.json
solarkennedy: Yea yea, do it on a different pass. Here I did kind of the bare minimum with the json schema because I knew it was in flux. It has stabilized by now though.

solarkennedy:
Shipit. This command is going to be a little weird till we figure out exactly how we *want* to use it, and how it will work inside a git commit hook too.

Rob-Johnson:
are you planning on adding any integration tests for this?

nhandler:
@Rob-Johnson I can add some. Let me fix the issue @solarkennedy reported and ship this. I have a change that is about ready for review to add chronos validation. I will bundle the itests with that.

PR #111, opened 2015-12-04
Remove docker-dev.yelpcorp.com check

solarkennedy:
I think we need to just remove this check. Obviously it has no value outside of yelp, and I don't really want to add the complexity of imposing some sort of custom rule framework.

We should still encourage our local users to use our base images of course.


bobtfish:
I HATE this, as I have apps that are *waaay* better than the yelp default, that get flagged as bad due to this. Hate me :P

PR #113, opened 2015-12-04
Added docs explaining the difference between difference service modes

solarkennedy:
We don't really explain this anywhere, so I figured I would add it.

### DIFF
--- a/docs/source/workflow.rst
+++ b/docs/source/workflow.rst
@@ -1,6 +1,44 @@
 Cooking: How it all comes together
 ==================================
 
+Ways That PaaSTA Can Run Services
+---------------------------------
+
+Long Running Services
+^^^^^^^^^^^^^^^^^^^^^
+
+Long running services are are processes that are expected to run continuously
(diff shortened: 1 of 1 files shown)


docs/source/workflow.rst
```diff
@@ -1,6 +1,44 @@
...
+Alternative names: workers, daemons, batch daemons, servers
+
+Scheduled Jobs
+^^^^^^^^^^^^^^
+
+Scheduled tasks are those tasks that are periodically run, and are not expected
+to run continously. Due to their ephemeral nature, they often do not expose a TCP port.
+
+PaaSTA uses `Chronos <yelpsoa_configs.html#chronos-clustername-yaml>_` to define
+how these scheduled jobs should be executed, and how often they should run.
```
Rob-Johnson: s/how these scheduled jobs should be executed/what command these scheduled jobs should execute/

Rob-Johnson:
fix n ship

PR #117, opened 2015-12-09
Remove extra context from replication check output for performance reasons

solarkennedy:
The extra output is cool, but has a bad effect where we hammer the marathon api when things start to go wrong, causing a cascading effect.

I think we can't afford to run paasta status -v this aggressively.


solarkennedy:
@EvanKrall I'm going to go ahead and ship this

EvanKrall:
Sounds good to me.

PR #120, opened 2015-12-09
Begin testing against marathon 11

solarkennedy:
AFAIK this has the right marathon libraries to do this. Let's see if Travis agrees!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ importlib==1.0.3
 isodate==0.5.1
 kazoo==2.2
 lockfile==0.9.1
-marathon==0.7.4
+marathon==0.7.5
 mccabe==0.3.1
 mesos.cli==0.1.3
 mesos.interface==0.22.1.2
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
         'httplib2 >= 0.9, <= 1.0',
         'isodate >= 0.5.0',
         'kazoo >= 2.0.0',
-        'marathon >= 0.7.4',
+        'marathon >= 0.7.5',
         'mesos.cli == 0.1.3',
         'ordereddict >= 1.1',
         'path.py >= 8.1',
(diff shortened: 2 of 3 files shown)


solarkennedy:
For my own notes:
```
crossover bounce in progress on bounce.test1. New marathon app bounce.test1.newapp.confighash exists. 0 new tasks to bring up. 2 old tasks receiving traffic. 0 old tasks draining. 1 old apps.
ERROR:marathon:Got HTTP 422: {"message":"Bean is not valid","errors":[{"attribute":"instances","error":"must be greater than or equal to 0"}]}
Exception raised during deploy of service bounce:
  File "/work/paasta_tools/setup_marathon_job.py", line 412, in deploy_service
    soa_dir=soa_dir,
  File "/work/paasta_tools/setup_marathon_job.py", line 246, in do_bounce
    client.kill_task(task.app_id, task.id, scale=True)
  File "/tmp/local/lib/python2.7/site-packages/marathon/client.py", line 420, in kill_task
    .format(app_id=app_id, task_id=task_id), params)
  File "/tmp/local/lib/python2.7/site-packages/marathon/client.py", line 75, in _do_request
    raise MarathonHttpError(response)
MarathonHttpError: MarathonHttpError: HTTP 422 returned with message, "Bean is not valid"
```

Something about the [DELETE /v2/apps/{appId}/tasks/{taskId}](https://mesosphere.github.io/marathon/docs/rest-api.html#delete-v2-apps-appid) endpoint doesn't like how we are calling it.

solarkennedy:
@Rob-Johnson a retry and it goes green like you said. Sounds like this is a flake. I wonder if it was actually flakey before, we just ignored it because it had a more benign error message?

Dare me to ship? I don't really want to commit a flakey test, at the same time I don't want to hold up marathon 11 too long.

Rob-Johnson:
I don't care so much about the flakey test, so much as I care about a bug in marathon that's going to cause problems with our bounces. I'm trying to debug marathon atm, so hang tight and If I don't get anywhere soon then we can think about our next move.

solarkennedy:
I'm going to ship this, see how flakey it really is, and if necessary adjust our code to be more tolerant of it.

EvanKrall:
Our bounces should be relatively tolerant of any marathon weirdness; if it returns a strange error code, that iteration of the bounce will error out, but 15 seconds later it would try again.

PR #121, opened 2015-12-10
Nix paasta_cli => to just cli

solarkennedy:
We are already in the paasta_tools module, I don't think we have to call the cli component "paasta_cli". I got tired of typing "paasta" a lot.

Dare me to ship this?


Rob-Johnson:
go for it!

debian/changelog
```diff
@@ -2126,7 +2126,7 @@ paasta-tools (0.7.19-yelp1) lucid; urgency=low
 
 paasta-tools (0.7.18-yelp1) lucid; urgency=low
 
-  * Create servicedocs stubs for paasta_cli
+  * Create servicedocs stubs for cli
```
EvanKrall: I would leave changelog alone

PR #122, opened 2015-12-10
Dependent jobs field

Rob-Johnson:
This is the first part: all I want here is for the parents to be pulled in from the config, and for the validation to fail if parents are specified.


paasta_tools/chronos_tools.py
```diff
@@ -248,6 +251,12 @@ def check_retries(self):
                 return False, 'The specified retries value "%s" is not a valid int.' % retries
         return True, ''
 
+    def check_parents(self):
+        parents = self.get_parents()
+        if parents is not None:
+            return False, 'Parents are not yet supported'
```
solarkennedy: I wouldn't bother with this, just fake it till we make it.

However, maybe here is where you might check that schedule is also not defined?

PR #134, opened 2015-12-15
Validate chronos

### LINKED ISSUE #127: Make 'paasta validate' validate chronos files
chronos_tools.py already has validation logic. We should utilize this in `paasta validate`.

nhandler:
Add some chronos validation to the `paasta validate` command. This fixes #127

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -18,7 +18,6 @@ import os
 import pkgutil
 import yaml
 
-from collections import OrderedDict
 from glob import glob
 from jsonschema import Draft4Validator
 from jsonschema import FormatChecker
@@ -182,7 +181,7 @@ def validate_chronos(service_path):
             checks_passed, check_msgs = cjc.validate()
 
             # Remove duplicate check_msgs
-            unique_check_msgs = list(OrderedDict.fromkeys(check_msgs))
+            unique_check_msgs = list(set(check_msgs))
 
             if not checks_passed:
                 print invalid_chronos_instance(cluster, instance, "\n  ".join(unique_check_msgs))
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -176,7 +176,9 @@ def validate_chronos(service_path):
     instance_type = 'chronos'
 
     for cluster in list_clusters(service, soa_dir, instance_type):
-        for instance in list_all_instances_for_service(service, instance_type, soa_dir):
+        for instance in list_all_instances_for_service(
+                service=service, clusters=[cluster], instance_type=instance_type,
+                soa_dir=soa_dir):
             cjc = load_chronos_job_config(service, instance, cluster, False, soa_dir)
             checks_passed, check_msgs = cjc.validate()
 
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -167,7 +167,6 @@ def get_service_path(service, soa_dir):
 def path_to_soa_dir_service(service_path):
     soa_dir = os.path.dirname(service_path)
     service = os.path.basename(service_path)
-
     return soa_dir, service
 
 
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -18,16 +18,21 @@ import os
 import pkgutil
 import yaml
 
+from collections import OrderedDict
 from glob import glob
 from jsonschema import Draft4Validator
 from jsonschema import FormatChecker
 from jsonschema import ValidationError
+
+from paasta_tools.chronos_tools import load_chronos_job_config
 from paasta_tools.cli.utils import failure
 from paasta_tools.cli.utils import get_file_contents
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_services
 from paasta_tools.cli.utils import PaastaColors
 from paasta_tools.cli.utils import success
+from paasta_tools.utils import list_all_instances_for_service
+from paasta_tools.utils import list_clusters
 
 
 SCHEMA_VALID = success("Successfully validated schema")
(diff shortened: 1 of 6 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -149,6 +165,31 @@ def get_service_path(service, soa_dir):
...
+
+
+def validate_chronos(service_path):
+    soa_dir, service = path_to_soa_dir_service(service_path)
+    instance_type = 'chronos'
+
+    for cluster in list_clusters(service, soa_dir, instance_type):
+        for instance in list_all_instances_for_service(service, instance_type, soa_dir):
+            cjc = load_chronos_job_config(service, instance, cluster, False, soa_dir)
+            (checks_passed, check_msgs) = cjc.validate()
```
Rob-Johnson: you don't need to wrap these in brackets.

paasta_tools/cli/cmds/validate.py
```diff
@@ -149,6 +165,31 @@ def get_service_path(service, soa_dir):
...
+
+            # Remove duplicate check_msgs
+            unique_check_msgs = list(OrderedDict.fromkeys(check_msgs))
```
Rob-Johnson: - why do you need this? What duplicated are there going to be? check_msgs is an array of strings afaict, so those strings are always going to be unique?
- we don't care about ordering, so list(set(check_msgs)) would be fine here.

paasta_tools/cli/utils.py
```diff
@@ -305,7 +305,7 @@ def guess_instance(service, cluster, args):
     else:
         try:
             instances = list_all_instances_for_service(
-                service=service, cluster=cluster, instance_type=None, soa_dir=args.yelpsoa_config_root)
```
Rob-Johnson: my guess is that ``local_run`` probably cares about this, so I'd be surprised if you can remove this without any side effects.

If you are going to remove it, then please remove the cluster parameter from the function definition, given it's not used anywhere else.

paasta_tools/utils.py
```diff
@@ -804,7 +804,7 @@ def get_default_cluster_for_service(service):
     return cluster
 
 
-def list_clusters(service=None, soa_dir=DEFAULT_SOA_DIR):
+def list_clusters(service=None, soa_dir=DEFAULT_SOA_DIR, instance_type=None):
```
Rob-Johnson: I'd prefer it if you left this function as it is, and either did the filtering wherever you call it, or if you think that the filtering will be done in more than one place, create a wrapper function to do that filtering (like list_chronos_clusters(...) and list_marathon_clusters(...))

paasta_tools/utils.py
```diff
@@ -818,14 +818,21 @@ def list_clusters(service=None, soa_dir=DEFAULT_SOA_DIR):
...
     for yaml_file in glob.glob('%s/*.yaml' % srv_path):
-        cluster_re_match = re.search('/.*/(marathon|chronos)-([0-9a-z-]*).yaml$', yaml_file)
+        cluster_re_match = re.search(search_re, yaml_file)
         if cluster_re_match is not None:
             clusters.add(cluster_re_match.group(2))
     return sorted(clusters)
 
 
-def list_all_instances_for_service(service, cluster=None, instance_type=None, soa_dir=DEFAULT_SOA_DIR):
+def list_all_instances_for_service(service, instance_type=None, soa_dir=DEFAULT_SOA_DIR):
```
Rob-Johnson: ah ok, this makes sense. I see why you're changing this in ``cli/utils.py`` now.

paasta_tools/cli/cmds/validate.py
```diff
@@ -149,6 +165,31 @@ def get_service_path(service, soa_dir):
...
+    service = os.path.basename(service_path)
+
+    return soa_dir, service
+
+
+def validate_chronos(service_path):
+    soa_dir, service = path_to_soa_dir_service(service_path)
+    instance_type = 'chronos'
+
+    for cluster in list_clusters(service, soa_dir, instance_type):
```
Rob-Johnson: is this correct with the changes you've made? ``list_all_instances_for_service`` iterates over every cluster, doesn't it?

paasta_tools/cli/cmds/validate.py
nhandler: You would think that this wouldn't be needed. However, without it, you wind up with duplicate messages for errors with certain fields like cpu and mem. I believe this is at least partially due to these lines in chronos_tools.py.validate

```
# Use InstanceConfig to validate shared config keys like cpus and mem
error_msgs.extend(super(ChronosJobConfig, self).validate())
```

I can create a new issue to look into this, but for the time being, the de-duplication is necessary. I will switch to using `list(set(check_msgs))` though.

paasta_tools/cli/utils.py
nhandler: Oops. Thought I had removed that as well. Will do.

paasta_tools/cli/cmds/validate.py
nhandler: Yes, it does. I need to iterate over all of the clusters for `cjc = load_chronos_job_config(service, instance, cluster, False, soa_dir)`. I can try and clean up the code a bit to make this more clear.

paasta_tools/cli/cmds/validate.py
```diff
@@ -149,6 +164,31 @@ def get_service_path(service, soa_dir):
...
+        for instance in list_all_instances_for_service(service, instance_type, soa_dir):
```
solarkennedy: You have to pass in the cluster here, otherwise you will list instances for clusters you are not interested in. Please use kwargs

paasta_tools/cli/utils.py
solarkennedy: I'm missing it. The reason `local-run` does need it to help enhance the guess about which instance to use if you have specified the cluster.

nhandler:
@solarkennedy Added the clusters argument back to list_all_instances_for_service. I think that addresses the last outstanding issue needing fixing.

paasta_tools/cli/utils.py
solarkennedy: This isn't the best. The reasoning behind this guess function is to save people the trouble of typing in an instance name. By removing the cluster input the guess is not as good. For example, if you ran something like:

    paasta local-run -c cluster1

and you had an instance called "main" in cluster1 and "backup" in cluster2, this new change would make the guess function suggested "backup" instead of "main" :(

nhandler: So to clarify (for my sake as well as anyone else reading this), since I think these line comments and lack of context might be causing some confusion.

Previously, list_all_instances_for_service was not actually using its cluster argument (due to a scoping issue). As a result, the cluster argument in guess_instance was not actually doing anything. As part of some cleanup, I removed the argument. Per @solarkennedy's feedback, I added it back to list_all_instances_for_service, but I guess I forgot to update guess_instance. I have now fixed that.

paasta_tools/cli/cmds/validate.py
```diff
@@ -149,6 +164,33 @@ def get_service_path(service, soa_dir):
             return None
 
 
+def path_to_soa_dir_service(service_path):
+    soa_dir = os.path.dirname(service_path)
+    service = os.path.basename(service_path)
+
```
Rob-Johnson: remove this whitespace

nhandler: Gone

nhandler:
Got a verbal ship-it, so I'm going to merge this.

PR #142, opened 2016-01-07
Consolidate/tighten up filtering of marathon apps so status and bounc…

EvanKrall:
…e code don't look at/work on apps for different services or instances. PAASTA-1897


nhandler:
Ship It

PR #143, opened 2016-01-08
Added a script to drain and kill a marathon app as gracefully as possible

mjksmith:
Added a script to drain and kill a marathon app.

### DIFF
--- a/paasta_tools/graceful_app_drain.py
+++ b/paasta_tools/graceful_app_drain.py
@@ -1,7 +1,4 @@
 #!/usr/bin/env python
-"""
-Docstring
-"""
 
 import argparse
 import sys
--- a/paasta_tools/graceful_app_drain.py
+++ b/paasta_tools/graceful_app_drain.py
@@ -1,10 +1,8 @@
 #!/usr/bin/env python
 
 import argparse
-import sys
 
 import service_configuration_lib
-from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools import marathon_tools
 from paasta_tools import drain_lib
 from paasta_tools import bounce_lib
@@ -21,7 +19,7 @@ def parse_args():
     parser.add_argument(
         'appname',
         help="the app that will be drained",
-    )#.completer = lazy_choices_completer()
+    )
     parser.add_argument(
         '-d', '--soa-dir',
         dest="soa_dir",
@@ -43,7 +41,7 @@ def main():
         marathon_config.get_password(),
     )
 
-    if not is_app_id_running(app_id=full_appid, client=client):
+    if not marathon_tools.is_app_id_running(app_id=full_appid, client=client):
         print("Couldn't find an app named {0}".format(full_appid))
         quit()
 
@@ -69,7 +67,7 @@ def main():
 
     bounce_func = bounce_lib.get_bounce_method_func('down')
 
-    while is_app_id_running(app_id=full_appid, client=client):
+    while marathon_tools.is_app_id_running(app_id=full_appid, client=client):
         app_to_kill = client.get_app(full_appid)
         try:
             old_app_live_tasks, old_app_draining_tasks = get_old_live_draining_tasks([app_to_kill], drain_method)
--- a/paasta_tools/graceful_app_drain.py
+++ b/paasta_tools/graceful_app_drain.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python
+"""
+Docstring
+"""
+
+import argparse
+import sys
+
+import service_configuration_lib
+from paasta_tools.cli.utils import lazy_choices_completer
+from paasta_tools import marathon_tools
+from paasta_tools import drain_lib
+from paasta_tools import bounce_lib
+from paasta_tools.utils import decompose_job_id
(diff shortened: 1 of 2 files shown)


paasta_tools/graceful_app_drain.py
```diff
@@ -0,0 +1,98 @@
...
+    marathon_config = marathon_tools.load_marathon_config()
+    client = marathon_tools.get_marathon_client(
+        marathon_config.get_url(),
+        marathon_config.get_username(),
+        marathon_config.get_password(),
+    )
+
+    if not marathon_tools.is_app_id_running(app_id=full_appid, client=client):
+        print("Couldn't find an app named {0}".format(full_appid))
+        quit()
```
Rob-Johnson: use sys.exit here.

paasta_tools/graceful_app_drain.py
```diff
@@ -0,0 +1,98 @@
...
+        metavar="SOA_DIR",
+        default=service_configuration_lib.DEFAULT_SOA_DIR,
+        help="define a different soa config directory",
+    )
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    full_appid = args.appname
```
Rob-Johnson: We're expecting the user to have looked up the 'proper' app id in the marathon interface here, so the usage is ``graceful_app_drain /example--service.mesosstage--canary.gitfb0c7ac5.configbd205682``. IMO it'd be better if you could use ``-s {service} -i {instance}`` (I know you're only copying the pb, which uses the 'full' app id).

paasta_tools/graceful_app_drain.py
```diff
@@ -0,0 +1,98 @@
...
+    if not marathon_tools.is_app_id_running(app_id=full_appid, client=client):
+        print("Couldn't find an app named {0}".format(full_appid))
+        quit()
+
+    service, instance, _, __ = (s.replace('--', '_') for s in decompose_job_id(full_appid))
+    short_appid = marathon_tools.format_job_id(service, instance)
+    complete_config = marathon_tools.create_complete_config(service, instance, marathon_config)
+    cluster = load_system_paasta_config().get_cluster()
+    service_instance_config = marathon_tools.load_marathon_service_config(
+        service,
```
Rob-Johnson: could you kwarg all these please?

Rob-Johnson:
fine first pass, some tests would be good though.

paasta_tools/graceful_app_drain.py
solarkennedy: This particular script is designed to be used by admins in the situation where our bounce code can't proceed for whatever reason, and we need to gracefully kill a *specific app* to get it going. For this reason, the appid should not be calculated, it needs to be provided by an admin. This script is not meant to be used very often, and not by normal users.

paasta_tools/chronos_tools.py
```diff
@@ -394,6 +395,15 @@ def get_healthcheck_mode(self, _):
         # Healthchecks are not supported yet in chronos
         return None
 
+    def get_desired_state_human(self):
```
solarkennedy: Is this.. just a bad rebase/merge? This should be in master already.

paasta_tools/graceful_app_drain.py
solarkennedy: yes please

paasta_tools/graceful_app_drain.py
```diff
@@ -0,0 +1,101 @@
...
+        service_instance_config.get_drain_method(service_namespace_config),
+        service=service,
+        instance=instance,
+        nerve_ns=nerve_ns,
+        drain_method_params=service_instance_config.get_drain_method_params(service_namespace_config),
+    )
+
+    bounce_func = bounce_lib.get_bounce_method_func('down')
+
+    while marathon_tools.is_app_id_running(app_id=full_appid, client=client):
```
solarkennedy: This looks like it would iterate really fast, can you make it a bit more relaxed and iterate every 10s like the normal bounce? (but print that it is sleeping to let the user know why it "isn't doing anything")

paasta_tools/graceful_app_drain.py
```diff
@@ -0,0 +1,101 @@
...
+        app_to_kill = client.get_app(full_appid)
+        try:
+            old_app_live_tasks, old_app_draining_tasks = get_old_live_draining_tasks([app_to_kill], drain_method)
+            with bounce_lib.bounce_lock_zookeeper(short_appid):
```
solarkennedy: I would... actually not lock here. This is an admin script, and if we need this to happen, it needs to happen.

paasta_tools/graceful_app_drain.py
```diff
@@ -0,0 +1,101 @@
...
+                    serviceinstance="{0}.{1}".format(service, instance),
+                    bounce_method='down',
+                    service=service,
+                    cluster=cluster,
+                    instance=instance,
+                    marathon_jobid=full_appid,
+                    client=client,
+                    soa_dir=soa_dir,
+                )
+        except MarathonHttpError:
```
solarkennedy: Do you still need this?

solarkennedy:
double check the tests and ship

PR #145, opened 2016-01-08
Fixed a bug where non-smartstack services would fail to send alerts when they start failing

mjksmith:
Fixed a bug where non-smartstack services would fail to send alerts when they start failing.

### DIFF
--- a/tests/test_check_marathon_services_replication.py
+++ b/tests/test_check_marathon_services_replication.py
@@ -577,7 +577,10 @@ def test_get_healthy_marathon_instances_for_short_app_id_considers_new_tasks_not
     for i in range(0, 4):
         fake_task = mock.Mock()
         fake_task.app_id = '/service.instance.foo%s.bar%s' % (i, i)
-        fake_task.started_at = datetime.now() - one_minute*i
+
+        # when i == 0, produces a task that has just started (not healthy yet)
+        # otherwise produces a task that was started over a minute ago (healthy)
+        fake_task.started_at = datetime.now() - one_minute * i
         mock_result = mock.Mock()
         mock_result.alive = True
         fake_task.health_check_results = [mock_result]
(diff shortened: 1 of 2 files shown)


Rob-Johnson:
I know we have this internally, but could you add a comment describing the bug you're fixing please? Also, can you add some tests so we can verify your change works?

tests/test_check_marathon_services_replication.py
```diff
@@ -568,6 +570,26 @@ def test_get_healthy_marathon_instances_for_short_app_id():
...
 
 
+def test_get_healthy_marathon_instances_for_short_app_id_considers_new_tasks_not_healthy_yet():
+    fake_client = mock.Mock()
+    fakes = []
+    one_minute = timedelta(minutes=1)
+    for i in range(0, 4):
+        fake_task = mock.Mock()
+        fake_task.app_id = '/service.instance.foo%s.bar%s' % (i, i)
+        fake_task.started_at = datetime.now() - one_minute*i
```
solarkennedy: I'm kinda surprised flake let you get away with `one_minute*i` without a space?

tests/test_check_marathon_services_replication.py
```diff
@@ -568,6 +570,26 @@ def test_get_healthy_marathon_instances_for_short_app_id():
...
+        mock_result = mock.Mock()
+        mock_result.alive = True
+        fake_task.health_check_results = [mock_result]
+        fakes.append(fake_task)
+    fake_client.list_tasks.return_value = fakes
+    actual = check_marathon_services_replication.get_healthy_marathon_instances_for_short_app_id(
+        fake_client,
+        'service.instance',
+    )
+    assert actual == 3
```
solarkennedy: I think I'm missing something... why is this `3` and not... 0?

solarkennedy:
ship

Issue #147, opened 2016-01-08
Figure out how to launch both a 'scheduled' job and 'dependent' job at the same time


Rob-Johnson:
In the case that a user adds both a 'scheduled' job and a 'dependent' job to their soa config at the same time, ``setup_chronos_job.py`` may not work if it tries to create the 'dependent' job before it's parent, the scheduled job.

We should try and ensure that any scheduled jobs are created *before* the dependent jobs. We currently [shuffle all the jobs before they are created](https://github.com/Yelp/paasta/blob/master/paasta_tools/deploy_chronos_jobs#L3)

Issue #146, opened 2016-01-08
setup_chronos_job does not clean up old jobs


Rob-Johnson:
We currently have an itest that checks that ``setup_chronos_job`` will also remove the oldest 5 Chronos job. See the feature [here](https://github.com/Yelp/paasta/blob/master/paasta_itests/setup_chronos_job.feature#L35-L44).

The jobs chosen for deletion are the 5 oldest found by ``chronos_tools.lookup_chronos_jobs`` (see [code](https://github.com/Yelp/paasta/blob/master/paasta_tools/setup_chronos_job.py#L149-L153)).

By default, ``chronos_tools.lookup_chronos_jobs`` **ignores** disabled jobs. Given that when we bounce jobs, the disabled field is set to ``True``, old jobs will never be bounced.

The integration test first mentioned only passes because it creates 10 dummy jobs with the disabled field set to ``False``. See the template for the job [here](https://github.com/Yelp/paasta/blob/master/paasta_itests/steps/setup_chronos_job_steps.py#L58).

I've fixed this as part of #150 , this is just here for reference.

solarkennedy:
Whoa crap. I thought for sure I totally fixed that. :(
I certainly fixed it "live" on a box when I was troubleshooting chronos bounces, but I guess I got distracted and never pushed.

I have that commit sitting on my laptop :(

Issue #141, opened 2016-01-06
write integration tests for launching a dependent job


Rob-Johnson:
As an isolated task, this isn't too bad, but there is some code that could do with refactoring in the chronos integration tests. This ticket is just here for tracking.

To properly test this, the scenario will look like:

- launch a regular, 'scheduled' job
- launch a second, 'dependent' job that references the scheduled job as it's parent

PR #150, opened 2016-01-08
First support for creating dependent jobs - Fixes #140, #141, #146

### LINKED ISSUE #140: lookup the latest full job for a 'service.instance' parent
We hide the 'full' name of a chronos job from the user. When a user adds a dependent job, they need to specify a parent as 'service.instance'. When this job is created, we need to do some work to find the 'full' job name.

The acceptance criteria for this is to be able to launch a dependent job where the parent is specified in the service.instance format.

Rob-Johnson:
This is a first pass at adding support for dependent jobs: a user can now add a 'parents' key to the yelpsoa-configs for their job and have a dependent job created.

A user specifies the parent for a dependent job in the format `service.instance`. This value is an array, and must be specified as so in the soa config.

When the job object is formatted into the JSON to be POSTed to Chronos, the 'service.instance' is translated into the 'full' job name by finding the most recent job associated with that service instance.

Caveats:

- The 'parent' job must already be launched in chronos before the dependent job is launched #147
- Validation for a matching job is done at launch time, rather than at 'validate' time.
- parents is an array, and setting the ``parent`` value as a string rather than array will blow up

There is probably more, but these are the ones I can think of immediately.

I've also done some refactoring of the integration tests for chronos as part of this. The key reasons for this were:

- We can now operate on multiple jobs at the same time using the new `jobs` field I added to the context. Before ``context.chronos_job`` was the only key available, which makes it more difficult when I need to operate on multiple jobs at the same time.
- The integration tests for ``setup_chronos_job`` were operating on ChronosJobConfig objects, and the chronos client rather than going through the full workflow the user would go through. I've modified this, so now the features more closely follow the users pattern. E.G 'I have yelpsoa-configs, I run setup_chronos_job, I modify the yelpsoa-configs, I run setup_chronos_job' rather than 'I have a chronosjobconfig, I run setup_chronos_job.setup_job(myconfig), I edit a field in the config etc. I think this is a good thing - the only caveat here is there are exceptions in the itest output from errors connecting to Sensu. Hopefully I can fix that by adding something that acts like a sensu server to listen a discard messages.

I found a bug (#146) as a result of doing this, so I think it was worthwhile. The fix for #146 is also in this PR.

### DIFF
--- a/paasta_itests/environment.py
+++ b/paasta_itests/environment.py
@@ -64,6 +64,8 @@ def _clean_up_chronos_jobs(context):
                 print "after_scenario: Job %s is present in chronos. Deleting." % job['name']
                 context.chronos_client.delete(job['name'])
             time.sleep(1)
+    if hasattr(context, 'jobs'):
+        context.jobs = {}
 
 
 def _clean_up_mesos_cli_config(context):
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -33,6 +33,7 @@ from paasta_tools.utils import get_code_sha_from_dockerurl
 from paasta_tools.utils import get_config_hash
 from paasta_tools.utils import get_paasta_branch
 from paasta_tools.utils import get_docker_url
+from paasta_tools.utils import get_service_instance_list
 from paasta_tools.utils import InstanceConfig
 from paasta_tools.utils import InvalidJobNameError
 from paasta_tools.utils import load_deployments_json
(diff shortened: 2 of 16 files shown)


paasta_itests/environment.py
```diff
@@ -64,6 +64,8 @@ def _clean_up_chronos_jobs(context):
                 print "after_scenario: Job %s is present in chronos. Deleting." % job['name']
                 context.chronos_client.delete(job['name'])
             time.sleep(1)
+    if hasattr(context, 'jobs'):
```
solarkennedy: Is this correct, did you want `! hasattr`? If not, what is the purpose of this code?

paasta_tools/chronos_tools.py
```diff
@@ -674,3 +681,23 @@ def delete_job(client, job):
 def create_job(client, job):
     log.debug("Creating job: %s" % job)
     client.add(job)
+
+
+def find_matching_parent_jobs(job_name):
```
solarkennedy: Can you make this singular if this only returns one job?

paasta_itests/environment.py
Rob-Johnson: hm, this was to make sure that the jobs were sanitised between each scenario.

the 'setup' part of this is done in [here](https://github.com/Yelp/paasta/blob/88defcdedbbd82d4be659d6047cf07fbb5dd88e1/paasta_itests/steps/setup_steps.py#L116). Given it's in the 'given a working paasta cluster', which is done for each scenario (though that's not a formal requirement), it renders the code you've referenced pretty useless. Maybe I should leave this here just to be defensive (in which case, I should move the 'setup' part of it somewhere else.

Rob-Johnson:
huh, I didn't know Travis was smart enough to run tests on the merge of this branch into master, rather than just this branch. looks like there is something in cli/validate that is requiring the ``schedule`` field to exist.

solarkennedy:
I think this is going to be one of those things where the change is so significant that it will be hard to really see the forest through the trees.

Thank you for the itest refactor, it was a long time coming.

We can worry about #147 next, it is the next logical road block, but at least with this code you can manually begin experimenting.

ship!

PR #153, opened 2016-01-11
remove assumptions from paasta validate tests

Rob-Johnson:
as it stands, we rely on assumptions on the behaviour of ChronosJobConfig.validate for the paasta validate unit tests. This removes those assumptions using Mocks so that we don't have to update these unit tests every time the behaviour of ChronosJobConfig.validate changes.


Rob-Johnson:
@solarkennedy "Bean is not Valid" again https://travis-ci.org/Yelp/paasta/jobs/101630292

solarkennedy:
ship

PR #155, opened 2016-01-12
Added an extra verbosity level to paasta_metastatus

mjksmith:
Uses Krall's mesos slave analysis script to display free CPU and RAM on each host. Called by using paasta metastatus -vv (or --verbose --verbose).

### DIFF
--- a/paasta_itests/steps/paasta_metastatus_steps.py
+++ b/paasta_itests/steps/paasta_metastatus_steps.py
@@ -20,7 +20,6 @@ from behave import when, then
 sys.path.append('../')
 from paasta_tools.utils import _run
 from paasta_tools.utils import remove_ansi_escape_sequences
-from paasta_itests.itest_utils import get_docker_compose_id_from_name
 from paasta_tools import marathon_tools
 from marathon import MarathonApp
 
@@ -84,5 +83,5 @@ def check_metastatus_contains_hostname(context, flags, expected_return_code):
         context,
         flags,
         expected_return_code,
-        get_docker_compose_id_from_name('mesosslave')
+        'mesosslave.test_hostname',
     )
--- a/paasta_itests/steps/paasta_metastatus_steps.py
+++ b/paasta_itests/steps/paasta_metastatus_steps.py
@@ -57,12 +57,12 @@ def marathon_task_is_ready(context, app_id):
     marathon_tools.wait_for_app_to_launch_tasks(context.marathon_client, app_id, 1)
 
 
-@then(u'paasta_metastatus {flags} exits with return code "{expected_return_code}" and output "{expected_output}"')
-def check_metastatus_return_code(context, flags, expected_return_code, expected_output):
+@then(u'paasta_metastatus{flags} exits with return code "{expected_return_code}" and output "{expected_output}"')
+def check_metastatus_return_code_with_flags(context, flags, expected_return_code, expected_output):
     # We don't want to invoke the "paasta metastatus" wrapper because by
     # default it will check every cluster. This is also the way sensu invokes
     # this check.
-    cmd = '../paasta_tools/paasta_metastatus.py %s' % flags
+    cmd = '../paasta_tools/paasta_metastatus.py%s' % flags
     env = dict(os.environ)
     env['MESOS_CLI_CONFIG'] = context.mesos_cli_config_filename
     print 'Running cmd %s with MESOS_CLI_CONFIG=%s' % (cmd, env['MESOS_CLI_CONFIG'])
@@ -77,11 +77,21 @@ def check_metastatus_return_code(context, flags, expected_return_code, expected_
     assert expected_output in escaped_output
 
 
-@then(u'paasta_metastatus {flags} exits with return code "{expected_return_code}" and outputs the slave\'s hostname')
-def check_metastatus_contains_hostname(context, flags, expected_return_code):
-    check_metastatus_return_code(
-        context,
-        flags,
-        expected_return_code,
-        'mesosslave.test_hostname',
+@then(u'paasta_metastatus exits with return code "{expected_return_code}" and output "{expected_output}"')
+def check_metastatus_return_code_no_flags(context, expected_return_code, expected_output):
+    check_metastatus_return_code_with_flags(
+        context=context,
+        flags='',
+        expected_return_code=expected_return_code,
+        expected_output=expected_output,
+    )
+
+
+@then(u'paasta_metastatus -vv exits with return code "{expected_return_code}" and outputs the slave\'s hostname')
+def check_metastatus_contains_hostname(context, expected_return_code):
+    check_metastatus_return_code_with_flags(
+        context=context,
+        flags=' -vv',
+        expected_return_code=expected_return_code,
+        expected_output='mesosslave.test_hostname',
     )
--- a/paasta_itests/steps/paasta_metastatus_steps.py
+++ b/paasta_itests/steps/paasta_metastatus_steps.py
@@ -20,6 +20,7 @@ from behave import when, then
 sys.path.append('../')
 from paasta_tools.utils import _run
 from paasta_tools.utils import remove_ansi_escape_sequences
+from paasta_itests.itest_utils import get_docker_compose_id_from_name
 from paasta_tools import marathon_tools
 from marathon import MarathonApp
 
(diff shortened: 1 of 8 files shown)


paasta_tools/cli/cmds/metastatus.py
```diff
@@ -38,9 +38,9 @@ def add_subparser(subparsers):
     )
     status_parser.add_argument(
         '-v', '--verbose',
-        action='store_true',
+        action='count',
         dest="verbose",
-        default=False,
+        default=0,
         help="Print out more output regarding the state of the cluster",
```
solarkennedy: Can you adjust the help output to hint that there can be more?

paasta_tools/cli/cmds/metastatus.py
```diff
@@ -38,10 +38,11 @@ def add_subparser(subparsers):
     )
     status_parser.add_argument(
         '-v', '--verbose',
-        action='store_true',
+        action='count',
         dest="verbose",
-        default=False,
-        help="Print out more output regarding the state of the cluster",
+        default=0,
```
solarkennedy: Can you add this to the itest to ensure we are invoking this new code path on a real box?

paasta_tools/paasta_metastatus.py
```diff
@@ -286,6 +286,26 @@ def print_results_for_healthchecks(summary, ok, results, verbose):
             print_with_indent(line, 2)
 
 
+def print_extra_mesos_data():
+    mesos_state = get_mesos_state_from_leader()
```
solarkennedy: Do we have to get the state again? I thought we had it already.

If not can you inject this data?

paasta_tools/paasta_metastatus.py
```diff
@@ -286,6 +286,26 @@ def print_results_for_healthchecks(summary, ok, results, verbose):
             print_with_indent(line, 2)
 
 
+def print_extra_mesos_data():
```
solarkennedy: Add at least one test for this, and make the function just return a string and let the caller be in charge of printing.

Input an example mesos state and assert that some of the expected strings are there.

This state file is going to change as we upgrade mesos, it will be easier to be sure what breaks if we can inject it at the unit test level

paasta_tools/cli/utils.py
```diff
@@ -512,9 +512,9 @@ def execute_paasta_serviceinit_on_remote_master(subcommand, cluster, service, in
     return run_paasta_serviceinit(subcommand, master, service, instancename, cluster, **kwargs)
 
 
-def run_paasta_metastatus(master, verbose=False):
-    if verbose:
-        verbose_flag = " -v"
+def run_paasta_metastatus(master, verbose=0):
+    if verbose is not 0:
+        verbose_flag = " -" + 'v'*verbose
```
Rob-Johnson: don't use + to concatenate strings. " -%" % 'v'*verbose

paasta_tools/cli/utils.py
```diff
@@ -512,9 +512,9 @@ def execute_paasta_serviceinit_on_remote_master(subcommand, cluster, service, in
     return run_paasta_serviceinit(subcommand, master, service, instancename, cluster, **kwargs)
 
 
-def run_paasta_metastatus(master, verbose=False):
-    if verbose:
-        verbose_flag = " -v"
+def run_paasta_metastatus(master, verbose=0):
+    if verbose is not 0:
```
Rob-Johnson: don't use 'is/is not' for comparison to an int. ``if verbose > 0:``

solarkennedy:
Just for record keeping, this PR is waiting for itest + review feedback fixes.

Rob-Johnson:
I don't really like the logic for 'extra' checks being printed in ``main()``, here's my suggestion for how to implement what you need:

Add a boolean parameter like ``verbose_checks`` to ``get_mesos_status``. Decide if this should be set to True or False on the value  of``verbose`` in ``main``.

Add a check named ``assert_slave_stats`` to the list of checks performed by ``get_mesos_state``.

```
def assert_slave_stats(mesos_state):
    header_string = "%12s %36s %9s' % ('Hostname', 'CPU free', 'RAM free')"
    slave_outputs = ['%40s %8.2f %9.2f' % (slave['hostname'], slave['free_resources']['cpus'], slave['free_resources'],slave['mem']) 
                     for slave in get_extra_mesos_slave_data(mesos_state)]
    check_output = [header_string].extend(slave_outputs)
    return ("\n".join(check_output),True)
```

N.B the second value of the tuple returned by the ``assert_*`` functions indicates if this check is 'healthy'. If there is anything that this check shows that is worth paging on (I don't think there is), then we can add logic to decide if we should return True/False accordingly.

modify ``get_mesos_state`` to perform this check based on the ``verbose_checks`` parameter.

``get_mesos_state`` will end up looking something like:

```
def get_mesos_status(mesos_state, run_verbose_checks=False):
    """Gathers information about the mesos cluster.
       :return: tuple of a string containing the status and a bool representing if it is ok or not
    """

    cluster_results = run_healthchecks_with_param(mesos_state, [assert_quorum_size, assert_no_duplicate_frameworks])

    metrics = get_mesos_stats()
    check_functions =  [
        assert_cpu_health,
        assert_memory_health,
        assert_tasks_running,
        assert_slave_health
    ]

    if run_verbose_checks:
       check_functions.append(assert_slave_stats)
    metrics_results = run_healthchecks_with_param(metrics, check_functions)
  
    return cluster_results + metrics_results
```
By doing this, we move the logic for 'whats verbose' away from ``main``, and the 'verbose' checks become regular healthchecks

I'm happy to pair with you on this if you like, too.

paasta_tools/paasta_metastatus.py
```diff
@@ -286,11 +285,36 @@ def print_results_for_healthchecks(summary, ok, results, verbose):
             print_with_indent(line, 2)
 
 
+def get_extra_mesos_slave_data(mesos_state):
+    slaves = dict((slave['id'], slave) for slave in mesos_state['slaves'])
```
Rob-Johnson: why do you need the whole slave dict as the value here? how about:

```
dict((slave['id'], {}) for slave in mesos_state['slaves'])
```

that way, in the below logic, you're won't be modifying the slave dict.

paasta_tools/paasta_metastatus.py
```diff
@@ -286,11 +285,36 @@ def print_results_for_healthchecks(summary, ok, results, verbose):
...
+
+    for slave in slaves.values():
```
Rob-Johnson: ```for slave in mesos_state['slaves']```

paasta_tools/paasta_metastatus.py
```diff
@@ -286,11 +285,36 @@ def print_results_for_healthchecks(summary, ok, results, verbose):
...
+        slave['free_resources'] = dict(slave['resources'].items())
```
Rob-Johnson:
```
slave_state = slaves[slave['id]]
slave_state['free_resources'] = dict(slave['resources'].items())
```

paasta_itests/steps/paasta_metastatus_steps.py
```diff
@@ -57,12 +57,12 @@ def marathon_task_is_ready(context, app_id):
     marathon_tools.wait_for_app_to_launch_tasks(context.marathon_client, app_id, 1)
 
 
-@then(u'paasta_metastatus exits with return code "{expected_return_code}" and output "{expected_output}"')
-def check_metastatus_return_code(context, expected_return_code, expected_output):
+@then(u'paasta_metastatus {flags} exits with return code "{expected_return_code}" and output "{expected_output}"')
+def check_metastatus_return_code(context, flags, expected_return_code, expected_output):
```
Rob-Johnson: can you check what happens when ``flags`` is empty?

```Then paasta_metastatus exits with..```

Will ``flags`` just be an empty string?

File: paasta_tools/paasta_metastatus.py [hunk 11]
```diff
@@ -162,20 +178,34 @@ def assert_quorum_size(state):
...
     metrics_results = run_healthchecks_with_param(metrics, [
         assert_cpu_health,
         assert_memory_health,
+        assert_tasks_running,
         assert_slave_health,
-        assert_tasks_running])
+    ])
+
+    if verbosity >= 2:
+        metrics_results.extend(run_healthchecks_with_param(mesos_state, [assert_extra_slave_data]))
```
Rob-Johnson: I think it'd be better if you went with what I suggested originally: move the array of checks into a variable, append to it if the verbose flag is set >=2, and then pass that array to ``run_healthchecks_with_param``. It's not a huge deal, just means ``run_healthchecks_with_param`` is only called once.

Rob-Johnson:
fix n ship - great stuff :+1:

PR #156, opened 2016-01-12
Extend schemas

### LINKED ISSUE #129: Update 'paasta validate' schemas to be more complete/accurate
The schemas we are using for validation are quite minimal. Our schemas have also stabilized since then. We should update the schemas to reflect this.

nhandler:
This change attempts to bring out schemas (used for validation) in line with http://paasta.readthedocs.org/en/latest/yelpsoa_configs.html . I manually tested it against all of our existing yelpsoa configs with no unexpected results. While doing this, I discovered a small bug with the validate command that prevented it from validating all yaml files (and I fixed it).

Closes #129


solarkennedy:
ship

PR #157, opened 2016-01-14
Update itest to use chronos 2.4.0-0.1.20151007110204.ubuntu1404

nhandler:


solarkennedy:
ship if it passes

Rob-Johnson:
merging - just this stupid 'Invalid Bean' message that's causing it to fail.

PR #160, opened 2016-01-14
Make deploy_chronos_services shuffle

### LINKED ISSUE #154: deploy_chronos_services isn't shuffling
`deploy_chronos_services` is always running setup for services in the same order, because `list_chronos_jobs` outputs 1 space-delimited line, and `shuf` shuffles lines.

adding `| tr ' ' '\n'` in there seems to help.

nhandler:
The shuf command shuffles lines, but list_chronos_jobs outputs 1
space-delimited line. This fixes #154

### DIFF
--- a/paasta_tools/deploy_chronos_jobs
+++ b/paasta_tools/deploy_chronos_jobs
@@ -1,4 +1,4 @@
 #!/bin/bash
 if am_i_mesos_leader >/dev/null; then
-  list_chronos_jobs | shuf | xargs -n 1 -r -P 5 setup_chronos_job
+  list_chronos_jobs | tr ' ' '\n' | shuf | xargs -n 1 -r -P 5 setup_chronos_job
 fi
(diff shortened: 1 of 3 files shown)


paasta_tools/deploy_chronos_jobs
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 if am_i_mesos_leader >/dev/null; then
-  list_chronos_jobs | shuf | xargs -n 1 -r -P 5 setup_chronos_job
+  list_chronos_jobs | tr ' ' '\n' | shuf | xargs -n 1 -r -P 5 setup_chronos_job
```
solarkennedy: This is... fine I think. What if list_chronos_jobs (and list_marathon_jobs) did one per line anyway?

nhandler: I thought about that, but I got worried that people might have scripts parsing that output (similar to how we do that with deploy_chronos_jobs) which would break after such a change. This approach seemed like the easiest and safest solution to the problem.

If we are fine possibly breaking some scripts, I can modify list_chronos_jobs.

solarkennedy:
ship

PR #161, opened 2016-01-14
Make paasta check call validate (Closes #128)

### LINKED ISSUE #128: Make 'paasta check' call 'paasta validate'
It would be nice if `paasta check` was the go-to command for knowing if your service was good to go. We should look into making it call `paasta validate` (in addition to its current checks). This will probably require adding some verbosity arguments to `paasta validate` so the output is a bit more manageable (possibly only showing issues).

nhandler:
This will allow `paasta check` to notify users of additional errors with their service. It will catch things such as invalid yelpsoa config syntax (based on the jsonschemas) and invalid chronos schedules.

Closes #128

### DIFF
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -21,6 +21,7 @@ import urllib2
 from service_configuration_lib import DEFAULT_SOA_DIR
 from service_configuration_lib import read_service_configuration
 
+from paasta_tools.cli.cmds.validate import paasta_validate
 from paasta_tools.cli.utils import figure_out_service_name
 from paasta_tools.cli.utils import get_file_contents
 from paasta_tools.cli.utils import is_file_in_dir
@@ -349,6 +350,7 @@ def paasta_check(args):
     deployments_check(service, soa_dir)
     sensu_check(service, service_path)
     smartstack_check(service, service_path)
+    paasta_validate(None, service_path)
 
 
 def read_dockerfile_lines(path):
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -21,7 +21,6 @@ import urllib2
 from service_configuration_lib import DEFAULT_SOA_DIR
 from service_configuration_lib import read_service_configuration
 
-from paasta_tools.cli.cmds.validate import paasta_validate
 from paasta_tools.cli.cmds.validate import paasta_validate_soa_configs
 from paasta_tools.cli.utils import figure_out_service_name
 from paasta_tools.cli.utils import get_file_contents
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/check.py
```diff
@@ -308,6 +309,7 @@ def paasta_check(args):
     marathon_deployments_check(service_path)
     sensu_check(service, service_path)
     smartstack_check(service, service_path)
+    paasta_validate(None, service_path)
```
solarkennedy: I'm always confused when I see this kind of "None" thing in an arg. None what?
Can you kwargs this?

paasta_tools/cli/cmds/validate.py
```diff
@@ -209,14 +209,17 @@ def validate_chronos(service_path):
...
-def paasta_validate(args):
+def paasta_validate(args, service_path=None):
     """Analyze the service in the PWD to determine if conf files are all valid
 
     :param args: argparse.Namespace obj created from sys.args by cli
     """
 
-    service = args.service
-    soa_dir = args.yelpsoa_config_root
+    if service_path:
```
solarkennedy: I don't really like this kind of... thing.

Can you just make paasta_validate take in args and process them..

And then make a `paasta_validate_soa_configs` that *just takes a soa_configs* input?

Then paasta_validate_soa_configs becomes much more like what you might expect, it takes a service to validate and a soa_configs path to look at where to validate.

nhandler:
Did a bit of refactoring to make `paasta_validate` only take `args` and then call `paasta_validate_soa_configs`. This required adding/modifying a number of tests.

paasta_tools/cli/cmds/validate.py
```diff
@@ -156,10 +156,27 @@ def add_subparser(subparsers):
...
 
+def check_service_path(service_path):
+    """Check that the specified path exists and has yaml files
+
+    :param service_path: Path to directory that should contain yaml files
+    """
+    if not service_path or not os.path.isdir(service_path):
+        print failure("%s is not a directory" % service_path,
+                      "http://paasta.readthedocs.org/en/latest/yelpsoa_configs.html")
+        return 1
```
Rob-Johnson: I know this follows the pattern set in the rest of the file, but I'd really like to see these 0s and 1s changed to True or False and turned to the right exit code at exit time

nhandler: That is fine. The reason I went for the numeric approach is to allow us to easily extend this command to have multipole failure return values that we can handle in different ways. But I am fine changing it to the cleaner True/False for now.

solarkennedy:
afaict this looks fine. This is one of those things that will need a manual test to be sure, then ship.

PR #162, opened 2016-01-14
Consider chronos*.yaml for deploy checks (Closes: #110)

### LINKED ISSUE #110: `paasta check` doesn't consider chronos-*.yaml for deploy.yaml checks

nhandler:
This will avoid `paasta check` message such as:

```
✗ There are some instance(s) in deploy.yaml that are not referenced
  by any marathon instance:
  cluster-name.batch-name, cluster-name2.batch-name
  You should probably delete these deploy.yaml entries if they are unused.
```

when there is a chronos-*.yaml file referencing the instance.

This change also will not show:

```
✗ No marathon.yaml exists, so your service cannot be deployed.
  Push a marathon-[superregion].yaml and run `paasta generate-pipeline`.
  More info: http://y/yelpsoa-configs
```

if a chronos.yaml file exists (and will show a more accurate message if neither yaml file exists).

Closes #110

### DIFF
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -177,17 +177,34 @@ def git_repo_check(service):
         print PaastaCheckMessages.git_repo_missing(git_url)
 
 
-def marathon_check(service_path):
-    """Check whether a marathon yaml file exists in service directory, and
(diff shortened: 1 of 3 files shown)


solarkennedy:
ship

PR #164, opened 2016-01-15
Allow users to set accepted_resource_roles in marathon yamls.

EvanKrall:
internal ticket PAASTA-2455.

This passes the `accepted_resource_roles` parameter through marathon.yaml into the marathon config, allowing for apps to be deployed to dedicated hardware.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -312,6 +312,11 @@ class MarathonServiceConfig(InstanceConfig):
             'cmd': self.get_cmd(),
             'args': self.get_args(),
         }
+
+        accepted_resource_roles = self.get_accepted_resource_roles()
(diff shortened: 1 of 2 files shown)


nhandler:
You will also need to update `paasta_tools/cli/schemas/marathon_schema.json`

solarkennedy:
Can you update the docs as well?

This is going to cause a big bounce too right? :(

EvanKrall:
@solarkennedy Yeah, this would cause a large bounce.

We could avoid that by using the `--default_accepted_resource_roles` command line parameter for Marathon, and omit `accepted_resource_roles` from the formatted marathon dict if it's not specified.

Should I try to do that?

EvanKrall:
I've made it so that the config dict doesn't contain `accepted_resource_roles` if it's not specified in your marathon.yaml

PR #165, opened 2016-01-15
Add --yelpsoa-config-root argument to 'paasta check'

nhandler:
This change adds support for a --yelpsoa-config-root argument in `paasta check`. I believe I have updated all of the checks it does to properly handle this directory. All of the PaaSTA tests are also passing locally.

### DIFF
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -37,6 +37,7 @@ from paasta_tools.utils import get_service_instance_list
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import _run
+from service_configuration_lib import DEFAULT_SOA_DIR
 from service_configuration_lib import read_service_configuration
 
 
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/check.py
```diff
@@ -322,21 +333,22 @@ def smartstack_check(service, service_path):
 def paasta_check(args):
     """Analyze the service in the PWD to determine if it is paasta ready
     :param args: argparse.Namespace obj created from sys.args by cli"""
-    service = figure_out_service_name(args)
-    service_path = os.path.join('/nail/etc/services', service)
+    service_root = args.yelpsoa_config_root
+    service_name = figure_out_service_name(args, service_root)
```
solarkennedy: Sorry, I'm going to veto service_name.

For better or worse throughout our codebase we just call this "service", please make it consistent.

paasta_tools/cli/cmds/check.py
```diff
@@ -288,15 +299,15 @@ def sensu_check(service, service_path):
         print PaastaCheckMessages.SENSU_MONITORING_MISSING
 
 
-def service_dir_check(service):
+def service_dir_check(service_name, service_root=DEFAULT_SOA_DIR):
```
solarkennedy: Now that we have the big default in the args, can you remove the kwarg default here to prevent someone from making the mistake of *not* passing service_root, causing a regression?

paasta_tools/cli/cmds/check.py
```diff
@@ -322,21 +333,22 @@ def smartstack_check(service, service_path):
 def paasta_check(args):
     """Analyze the service in the PWD to determine if it is paasta ready
     :param args: argparse.Namespace obj created from sys.args by cli"""
-    service = figure_out_service_name(args)
-    service_path = os.path.join('/nail/etc/services', service)
+    service_root = args.yelpsoa_config_root
```
solarkennedy: Instead of calling this "service_root" everywhere, can you call it just call soa_dir?

solarkennedy:
fix+ship

PR #167, opened 2016-01-16
made InstanceConfig take service, cluster and instance as args

mjksmith:
InstanceConfig and it's children MesosServiceConfig and ChronosJobConfig now directly store service name, cluster and instance. This should enable some cleaner code in the future.

### DIFF
--- a/general_itests/steps/deployments_json_steps.py
+++ b/general_itests/steps/deployments_json_steps.py
@@ -73,9 +73,12 @@ def step_impl_when(context):
                    return_value=context.test_git_repo_dir),
         mock.patch('paasta_tools.generate_deployments_for_service.parse_args',
                    autospec=True, return_value=fake_args),
+        mock.patch('paasta_tools.generate_deployments_for_service.get_branches_for_service', autospec=True,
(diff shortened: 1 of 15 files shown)


solarkennedy:
I believe this is a good thing, and will make our code simpler in general because we will need to just pass this object around and the metadata will come with it.

It also enables the possibility for us to set environment variables in each app with this metadata, which is kinda something I've wanted for a while.

I would like a second opinion from someone who knows more about OOP in python. @EvanKrall  or @Rob-Johnson ?

EvanKrall:
Looking.

mjksmith:
I'll fix up the general_itests right now

Rob-Johnson:
I think this is fine - my only groan is about ``job_name`` & ``instance`` being used interchangeably in different parts of the codebase. I don't care which it is, just that it's consistent.

solarkennedy:
My opinion is: Call it `instance`, and only at the last second when we must interact with the clientlib/api call it by whatever it needs to be called (job, job_name, job_id, etc.)

PR #168, opened 2016-01-19
Fixed formatting for very verbose paasta metastatus

mjksmith:
made the formatting for paasta_metastatus -vvv more generous when making space for the hostname

It should support hostnames up to length 46 now


mjksmith:
I'm changing this to be done with the table formatting functions

PR #169, opened 2016-01-20
Pick an ephemeral port better

asottile:
To avoid these:

```
15:39:54 time="2016-01-19T15:39:54-08:00" level=fatal msg="Error response from daemon: Cannot start container 4ab7d672f26436db000e87d5c0939a681dff7364547b348990ee2154f8e83778: Error starting userland proxy: listen tcp 0.0.0.0:33421: bind: address already in use" 
```

We already use this strategy for other utilities (see `ephemeral-port-reserve` (internal) for where the code comes from via @bukzor).


solarkennedy:
Failing test is the bean on. 

Sounds legit to me, especially if we are already using this pattern in IRL. Thanks!

PR #173, opened 2016-01-20
Fixed formatting for very verbose paasta metastatus

mjksmith:
Used the table formatting tools in paasta_tools.utils to clean up very verbose metastatus formatting for long hostnames.


solarkennedy:
ship

PR #174, opened 2016-01-21
fixed paasta_metastatus -vv when there are no mesos slaves

mjksmith:
Running paasta metastatus -vv on a cluster with no mesos slaves will no longer throw an exception (hopefully)

This also takes into account running paasta metastatus on a mesos setup with no frameworks, or no tasks in a framework.

### DIFF
--- a/paasta_tools/paasta_metastatus.py
+++ b/paasta_tools/paasta_metastatus.py
@@ -66,10 +66,10 @@ def get_extra_mesos_slave_data(mesos_state):
     slaves = dict((slave['id'], {
         'free_resources': slave['resources'],
         'hostname': slave['hostname'],
-    }) for slave in mesos_state['slaves'])
+    }) for slave in mesos_state.get('slaves', []))
 
-    for framework in mesos_state['frameworks']:
-        for task in framework['tasks']:
+    for framework in mesos_state.get('frameworks', []):
+        for task in framework.get('tasks', []):
             resource_dict = slaves[task['slave_id']]['free_resources']
             for resource_type, value in task['resources'].items():
                 if resource_type in ['cpus', 'disk', 'mem']:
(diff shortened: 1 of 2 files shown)


paasta_tools/paasta_metastatus.py
```diff
@@ -180,14 +180,19 @@ def assert_quorum_size(state):
...
+        rows = [('Hostname', 'CPU free', 'RAM free')]
+        for slave in extra_slave_data:
+            rows.append((
+                slave['hostname'],
+                '%.2f' % slave['free_resources']['cpus'],
+                '%.2f' % slave['free_resources']['mem'],
+            ))
+        output = '\n'.join(('    %s' % row for row in format_table(rows)))[2:]
+    else:
+        output = '  No mesos slaves on this cluster'
```
solarkennedy: This needs to return False, as this is a very bad condition

solarkennedy: Also say it with an exclamation mark!!!

Be more specific like: "No mesos slaves registered in this cluster!!!"

paasta_tools/paasta_metastatus.py
```diff
@@ -66,10 +66,10 @@ def get_extra_mesos_slave_data(mesos_state):
     slaves = dict((slave['id'], {
         'free_resources': slave['resources'],
         'hostname': slave['hostname'],
-    }) for slave in mesos_state['slaves'])
+    }) for slave in mesos_state.get('slaves', []))
```
solarkennedy: I'm... going to make the call that I think this is wrong.

If we are literally missing the 'slaves' key, that is some sort of other error that we need to raise an exception for.

At this point of the code, it should be considered safe that the mesos_state is correctly formatted. (pretty sure)

Rob-Johnson:
can you paste you running this locally here? ship if it works :+1:

solarkennedy:
ship

PR #177, opened 2016-01-22
Enabled deploy_groups

mjksmith:
Placing multiple instances in a deploy group makes it so they are all deployed with the same docker container and are deployed at the same time (as specified in deploy.yaml). However, they can be controlled independently by flagging deploy_state and force_bounce in git refs.

### DIFF
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -19,6 +19,10 @@ import re
 import urllib2
 
 from paasta_tools.cli.cmds.validate import paasta_validate_soa_configs
+from service_configuration_lib import DEFAULT_SOA_DIR
+from service_configuration_lib import read_service_configuration
+
+from paasta_tools.chronos_tools import load_chronos_job_config
 from paasta_tools.cli.utils import figure_out_service_name
 from paasta_tools.cli.utils import get_file_contents
 from paasta_tools.cli.utils import is_file_in_dir
@@ -30,6 +34,7 @@ from paasta_tools.cli.utils import success
 from paasta_tools.cli.utils import validate_service_name
 from paasta_tools.cli.utils import x_mark
 from paasta_tools.marathon_tools import get_all_namespaces_for_service
+from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.monitoring_tools import get_team
 from paasta_tools.utils import _run
 from paasta_tools.utils import DEPLOY_PIPELINE_NON_DEPLOY_STEPS
@@ -37,11 +42,6 @@ from paasta_tools.utils import get_git_url
 from paasta_tools.utils import get_service_instance_list
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import PaastaColors
-from paasta_tools.utils import _run
-from service_configuration_lib import DEFAULT_SOA_DIR
-from service_configuration_lib import read_service_configuration
-from paasta_tools.marathon_tools import load_marathon_service_config
-from paasta_tools.chronos_tools import load_chronos_job_config
 
 
 def get_pipeline_config(service, soa_dir):
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -223,6 +223,7 @@ def get_chronos_steps(service, soa_dir):
                 instance=instance,
                 cluster=cluster,
                 soa_dir=soa_dir,
+                load_deployments=False,
             )
             steps.append(config.get_deploy_group())
     return steps
@@ -243,6 +244,7 @@ def get_marathon_steps(service, soa_dir):
                 instance=instance,
                 cluster=cluster,
                 soa_dir=soa_dir,
+                load_deployments=False,
             )
             steps.append(config.get_deploy_group())
     return steps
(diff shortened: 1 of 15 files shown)


solarkennedy:
Sorry, lots of widespread changes. Can you rebase?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -50,8 +50,8 @@ def add_subparser(subparsers):
         required=True,
     )
     list_parser.add_argument(
-        '-l', '--clusterinstance',
-        help='Mark the service ready for deployment in this clusterinstance (e.g. '
+        '-l', '--deploy_group',
```
solarkennedy: You are going to have to make this backwards compatible. All of our existing jenkins jobs are doing --clusterinstance right now

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -48,8 +48,8 @@ def add_subparser(subparsers):
         required=True,
     )
     list_parser.add_argument(
-        '-l', '--clusterinstance',
-        help='Mark the service ready for deployment in this clusterinstance (e.g. '
+        '-l', '--deploy_group', '--clusterinstance',
+        help='Mark the service ready for deployment in this deploy group (e.g. '
              'cluster1.canary, cluster2.main)',
```
EvanKrall: Make a note here that `--clusterinstance` is deprecated in favor of `--deploy_group`; I'm in favor of backwards compatibility here, but we should clean up the old option name at some point.

paasta_tools/utils.py
```diff
@@ -971,8 +984,12 @@ def get_branch_dict(self, service, branch):
         return self.get(full_branch, {})
 
 
+def get_paasta_branch_from_identifier(identifier):
```
EvanKrall: I feel like we should just call it `deploy_group` instead of `identifier`, and thus `get_paasta_branch_from_deploy_group`.

paasta_tools/cli/cmds/check.py
```diff
@@ -209,10 +211,16 @@ def get_chronos_steps(service, soa_dir):
     with what deploy.yaml has for linting."""
     steps = []
     for cluster in list_clusters(service, soa_dir):
-        for instance in get_service_instance_list(service=service, cluster=cluster,
-                                                  instance_type='chronos',
-                                                  soa_dir=soa_dir):
-            steps.append("%s.%s" % (cluster, instance[1]))
+        for _, instance in get_service_instance_list(service=service, cluster=cluster,
```
Rob-Johnson: I don't know if pre-commit will pick this up, but can you align these?

paasta_tools/cli/cmds/rollback.py
```diff
@@ -65,57 +60,53 @@ def add_subparser(subparsers):
...
-    cluster, instance and sha. These arguments will be verified and passed onto
+    deploy groups and sha. These arguments will be verified and passed onto
     mark_for_deployment.
     """
     service = figure_out_service_name(args)
-    cluster = args.cluster
     git_url = get_git_url(service)
     commit = args.commit
-    given_instances = args.instances.split(",")
+    given_deploy_groups = filter(bool, args.deploy_groups.split(","))
```
Rob-Johnson: This is...confusing. What are you trying to achieve here?

paasta_tools/cli/cmds/rollback.py
```diff
@@ -65,57 +60,53 @@ def add_subparser(subparsers):
...
-            returncode = mark_for_deployment(
-                git_url=git_url,
-                cluster=cluster,
-                instance=instance,
-                service=service,
-                commit=commit,
-            )
-    else:
-        print PaastaColors.red("ERROR: The service %s is not deployed into cluster %s.\n" % (service, cluster))
+    if len(deploy_groups) is 0:
```
Rob-Johnson: use ``==`` rather than ``is`` here

Rob-Johnson:
First pass, I'm gonna read this again.

paasta_tools/cli/cmds/rollback.py
mjksmith: The `instances` parameter to paasta rollback is supposed to be optional; if it's not included rollback should rollback every instance.

This code removes empty strings from `given_instances`. Before this if the instances argument wasn't passed to rollback it would default to `''`, so given_instances would be `['']`. This causes rollback not to work properly as the rest of the code checks to see if `len(given_instances) == 0`.

I chose to make this change rather than changing the logic for determining if `given_instances` is empty because this also deals with some typos (correcting multiple commas like `'instance1,,instance2'`) and also removing trailing commas (before `'instance_name,'` would resolve to `['instance_name', '']`).

TL;DR this feature was always broken and filtering with the `bool` function removes empty strings from the result of `split`.

Rob-Johnson: I see. This still fails if you do ``instance1, ,instance2`` though:

```python
> bool(' ')
True
> filter(bool, "instance1, ,instance2".split(","))
['instance1', ' ', 'instance2']
```
How about:

```python
# filter the empty strings
[deploy_group for deploy_group in args.deploy_groups.split(",") if not deploy_group]
```
?

Yeah, it's probably slower, but the cost of that vs the non obvious behaviour of this code is worth it imo. What do you think?

mjksmith: I'm not actually sure how argparse will handle spaces in an argument.

I still like your list comprehension more though, I'll make that change.

PR #178, opened 2016-01-22
Fix the broken link

ronin13:
Fixes the broken link here: http://paasta.readthedocs.org/en/latest/about/paasta_principles.html


solarkennedy:
Thanks!

PR #179, opened 2016-01-22
Have crossover bounce prefer killing unhealthy tasks.

EvanKrall:
This changes the bounce function interface so that the bounce code knows which old tasks are healthy and which are not. Specifically, I split the `old_app_live_tasks` into two groups: happy and unhappy. These are determined by `get_happy_tasks`.

This allows the bounce functions (specifically, crossover) to make decisions about which old tasks to kill, instead of treating them identically.


solarkennedy:
Seams legit. Ship.

PR #181, opened 2016-01-23
chronos now gives sensu the full job id of a failed task

mjksmith:
This should allow us to catch some bugs that are eluding us


solarkennedy:
I think this is ok. ship if the tests pass

PR #183, opened 2016-01-24
Install deps during manpages

asottile:
@solarkennedy this should fix it I would think


Rob-Johnson:
this works for me - I'm going to merge. thanks @asottile

PR #186, opened 2016-01-26
Make tox faster

EvanKrall:
Make local runs of tox faster by moving pre-commit to testenv:general_itest, moving coverage into its own testenv, and speeding up a unit test that was taking 1 second.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 env:
   - TOXENV=py27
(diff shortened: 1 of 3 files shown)


solarkennedy:
ship

asottile:
@EvanKrall you probably still want `pre-commit install -f` (trivially fast) in testenv (if that's a common target) so people'll get the hooks installed by default

PR #187, opened 2016-01-26
WIP: Implement 'pool': An attribute & constraint that lets you split a cluster

EvanKrall:
...into multiple parts, without having to use Mesos's roles.

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -218,6 +218,18 @@ class InstanceConfig(dict):
         <https://mesosphere.github.io/marathon/docs/native-docker.html>`_"""
         return self.config_dict.get('extra_volumes', [])
 
+    def get_pool(self):
+        """Which pool of nodes this job should run on. This can be used to mitigate noisy neighbors, by putting
+        particularly noisy or noise-sensitive jobs into different pools.
+
+        This is implemented with an attribute "pool" on each mesos slave and by adding a constraint to Marathon/Chronos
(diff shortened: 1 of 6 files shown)


EvanKrall:
I've marked this as WIP because I want to be very careful about merging and rolling this out. As is, this will cause a large bounce.

paasta_tools/utils.py
```diff
@@ -218,6 +218,18 @@ def get_extra_volumes(self):
         <https://mesosphere.github.io/marathon/docs/native-docker.html>`_"""
         return self.config_dict.get('extra_volumes', [])
 
+    def get_pool(self):
+        """Which pool of nodes this job should run on. This can be used to mitigate noisy neighbors, by putting
```
solarkennedy: Can you move these docs, or some of them to the actual docs location?

solarkennedy:
Are you sure you don't want to do the thing where if pool is not None, add the constraint?

Also of course we'll need to add the attributes to our slaves now

EvanKrall:
I'm not sure what you mean in your first sentence. My default behavior is pretty clearly specified: a constraint of `['pool', 'LIKE', 'default']`.

I've got an internal review for adding the attribute to our slaves.

solarkennedy:
Gotcha, it sound like you are going to deploy the new attribute first, and then optionally add this constraint to a bunch of apps, then push this code?

EvanKrall:
Yeah, that's the plan.

PR #189, opened 2016-01-26
Pick a smartstack port randomly, instead of always picking highest existing + 1.

EvanKrall:


paasta_tools/cli/fsm/autosuggest.py
```diff
@@ -37,17 +38,16 @@ def _get_smartstack_proxy_port_from_file(root, file):
     return int(port)
 
 
-def suggest_smartstack_proxy_port(yelpsoa_config_root):
-    """Pick the next highest smartstack proxy port from the 20000-21000 block"""
-    max_proxy_port = 0
+def suggest_smartstack_proxy_port(yelpsoa_config_root, range_min=20000, range_max=21000):
+    """Pick a random available port in the 20000-21000 block"""
+    available_proxy_ports = set(range(range_min, range_max + 1))
```
mjksmith: Won't this make the port return something in (range_min, range_max) instead of [range_min, range_max]?

EvanKrall: No, this would return something in [range_min, range_max].

```
In [1]: range_min = 1

In [2]: range_max = 3

In [3]: range(range_min, range_max + 1)
Out[3]: [1, 2, 3]
```

nhandler:
It might also be nice to add some more logic to catch the situation where we have no free ports in the specified range.

EvanKrall:
Possibly; we'd get an IndexError.

But sure, I'll wrap it and raise a more descriptive error.

paasta_tools/cli/fsm/autosuggest.py
mjksmith: Sorry, I got my set notation mixed up.

nhandler:
Ship It!

PR #190, opened 2016-01-27
Mark for deployment tags

nhandler:
In an effort to migrate us from branches to tags, this first pull request updates `mark_for_deployment` to tag as well as create a branch. Since branches are still being created like normal and no other code has been updated to switch from branches to tags, this should be a safe change.

The change to CHAGNELOG.md was to make the pre-commit hook happy.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -22,7 +22,7 @@ from paasta_tools.cli.utils import validate_service_name
 from paasta_tools.utils import _log
 from paasta_tools.utils import format_tag
 from paasta_tools.utils import get_paasta_branch_from_deploy_group
-from paasta_tools.utils import get_paasta_tag
+from paasta_tools.utils import get_paasta_tag_from_deploy_group
 
 
 def add_subparser(subparsers):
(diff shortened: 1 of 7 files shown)


Rob-Johnson:
lg2m. ship

paasta_tools/utils.py
```diff
@@ -975,6 +975,10 @@ def get_paasta_branch(cluster, instance):
     return 'paasta-%s.%s' % (cluster, instance)
 
 
+def get_paasta_tag(cluster, instance):
+    return get_paasta_branch(cluster, instance)
```
solarkennedy: I think this is a good start, long term I think we need the type of tag that start/stop does with a timestamp. Otherwise we are moving this same tag over and over which is kinda gross.

EvanKrall: +1 do this before you ship this branch

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -65,9 +67,11 @@ def add_subparser(subparsers):
 
 def mark_for_deployment(git_url, cluster, instance, service, commit):
```
solarkennedy: Can you... take this code and use it in the itest code that "mocks" it?

(general_itests/steps/deployments_json_steps.py:56)

solarkennedy:
You are going to have to rebase on top of @mjksmith's deploy_group changes carefully.

solarkennedy:
ship

PR #191, opened 2016-01-27
made cleanup_chronos_jobs send OK on sensu

mjksmith:
When instances are removed from a `chronos-*.yaml` `cleanup_chronos_jobs` now sends an OK status to sensu.

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -715,19 +715,6 @@ def find_matching_parent_job(job_name):
         return None
 
 
-def compose_check_name_for_job(service, instance):
+def compose_check_name_for_service_instance(check_name, service, instance):
     """Compose a sensu check name for a given job"""
-    return 'check-chronos-jobs.%s%s%s' % (service, INTERNAL_SPACER, instance)
-
-
-def send_event_to_sensu(service, instance, monitoring_overrides, soa_dir, status_code, message):
-    check_name = compose_check_name_for_job(service, instance)
-
-    monitoring_tools.send_event(
-        service=service,
-        check_name=check_name,
-        overrides=monitoring_overrides,
-        status=status_code,
-        output=message,
-        soa_dir=soa_dir,
-    )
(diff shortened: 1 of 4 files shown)


paasta_tools/cleanup_chronos_jobs.py
```diff
@@ -28,9 +28,11 @@
 import argparse
 import sys
 
+import pysensu_yelp
 import service_configuration_lib
 
 from paasta_tools import chronos_tools
+from paasta_tools.check_chronos_jobs import send_event_to_sensu
```
Rob-Johnson: Instead of doing this, can you move ``paasta_tools.check_chronos_jobs.send_event_to_sensu`` to ``paasta_tools.chronos_tools`` or even ``paasta_tools.utils`` if it is generic enough?

paasta_tools/chronos_tools.py
```diff
@@ -720,3 +720,21 @@ def find_matching_parent_job(job_name):
         return sorted_jobs[0]['name']
     else:
         return None
+
+
+def compose_check_name_for_job(service, instance):
+    """Compose a sensu check name for a given job"""
+    return 'check-chronos-jobs.%s%s%s' % (service, INTERNAL_SPACER, instance)
```
Rob-Johnson: ah, this isn't the correct place for this.

Here's what I would do:

in ``chronos_tools.py`` and ``marathon_tools.py``:
```python
def compose_check_name_for_service_instance(check_name, service, instance):
      return '%s.%s%s%s' % (check_name, service, INTERNAL_SPACER, instance)
```

in ``check_chronos_jobs.py``

```python
def send_event(service, instance):
     """check_chronos_jobs specific wrapper for monitoring_tools.send_event_to_sensu"""
     check_name = chronos_tools.compose_check_name_for_job('check_chronos_jobs', service, instance)
     return monitoring_tools.send_event(check_name, service, instance....)
```

You can then follow this pattern of copying the ``check_chronos_tools.send_event`` function in other scripts (like in #198)

solarkennedy:
fix+ship

PR #194, opened 2016-01-27
Add healthcheck_mode to service config

fede1024:
The healthcheck_mode option was added to nerve-tools, but it is not actually loaded from the config file since the 'healthcheck_mode' key is not in the whitelist.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -461,6 +461,7 @@ def load_service_namespace_config(service, namespace, soa_dir=DEFAULT_SOA_DIR):
     Retrevies the following keys:
 
     - proxy_port: the proxy port defined for the given namespace
+    - healthcheck_mode: the mode for the healthcheck (http or tcp)
     - healthcheck_port: An alternate port to use for health checking
     - healthcheck_uri: URI target for healthchecking
     - healthcheck_timeout_s: healthcheck timeout in seconds
@@ -495,6 +496,7 @@ def load_service_namespace_config(service, namespace, soa_dir=DEFAULT_SOA_DIR):
     # and there's other things that appear in the smartstack section in
     # several cases.
     key_whitelist = set([
+        'healthcheck_mode',
         'healthcheck_uri',
         'healthcheck_port',
         'healthcheck_timeout_s',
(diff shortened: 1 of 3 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -495,6 +496,7 @@ def load_service_namespace_config(service, namespace, soa_dir=DEFAULT_SOA_DIR):
     # and there's other things that appear in the smartstack section in
     # several cases.
     key_whitelist = set([
+        'healthcheck_mode',
```
solarkennedy: I'm surprised this is it? Don't we need something like line 522?

mjksmith: It doesn't took like that's necessary because line 515 adds everything in the whitelist to the configuration.

fede1024: configure_nerve in nerve tools will use this config, and it will default to the mode if it's not present.

solarkennedy:
@fede1024 I'm ok with the small change, but in paasta if a service specifies this mode, seems like we should take advantage of it and adjust our healthchecks accordingly?

Rob-Johnson:
I think this is ok to ship.

@solarkennedy [here](https://github.com/Yelp/nerve-tools/commit/09d43a765a4c2dd712739c497e3fb5ddf0658b15) is the corresponding commit for nerve-tools.

This is a consequence of ``configure_nerve`` using paasta as a library, particularly ``marathon_tools.get_services_runnning_here_for_nerve``. That in turn uses this ``key_whitelist`` to filter the list of keys returned from the ``s_c_l``, so @fede1024 needed to add ``healthcheck_mode`` to that whitelist so nerve can start using it.

> in paasta if a service specifies this mode, seems like we should take advantage of it and adjust our 
healthchecks accordingly?

I think we're actually okay here. We [already use](https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_tools.py#L413-L419) the ``healthcheck_mode`` param used, it was nerve that didn't have that capability. This actually brings nerve healthchecking and marathon healthchecking in line with eachother.

solarkennedy:
@fede1024 can you merge/rebase master and we can see how the tests look?

fede1024:
It's now rebased on master. Unit tests and integration tests look fine on my box.

solarkennedy:
Merged.

PR #196, opened 2016-01-27
Inject PAASTA_* variables into every task

solarkennedy:
This is something I've wanted for a while, but haven't pushed because it involves a big bounce to every app.


Rob-Johnson:
nice. :ship:

drolando:
+1

PR #197, opened 2016-01-27
Actually install schemas

nhandler:
This fixes an issue where the schemas used by `paasta validate` were not actually getting installed (causing a failure message to appear in validate and check).


solarkennedy:
ship

PR #199, opened 2016-01-28
skip dependent jobs if they are created before their parent

giuliano108:
Address #147 by catching `InvalidParentError` in `setup_chronos_job` and returning a 0 exit code.
Alternatively, we could add logic to `list_chronos_jobs` to shuffle the jobs while maintaining parent-dependent relationships. That's not currently possible because the `xargs -P 5` in 
 [deploy_chronos_jobs](https://github.com/Yelp/paasta/blob/42ae5c252fd80e08dea2885e09b3c11949748311/paasta_tools/deploy_chronos_jobs#L3)) is likely to de-serialize the jobs again.


Rob-Johnson:
shipit - :ship:

solarkennedy:
ship

PR #200, opened 2016-01-28
Show details of parents in output of ``paasta status``

### LINKED ISSUE #96: Show the status of a job's parents in paasta status
If a job has parent(s), then we should show the status of those jobs in the output of `paasta status`.

Rob-Johnson:
screen shot 2016-01-28 at 16 17 39

This is slightly wasteful, in that we're doing the lookups for the parent jobs multiple times. I'm ok with that for now, but probably a case where we can optimize if we get to the point where it's impacting on performance.

Closes #96

### DIFF
--- a/paasta_tools/chronos_serviceinit.py
+++ b/paasta_tools/chronos_serviceinit.py
@@ -132,7 +132,7 @@ def _format_last_result(job):
 
 def _format_schedule(job):
     if job.get('parents') is not None:
-        schedule = PaastaColors.yellow("None (Dependent Job). Run with -v to see parent details.")
+        schedule = PaastaColors.yellow("None (Dependent Job).")
     else:
         schedule = job.get("schedule", PaastaColors.red("UNKNOWN"))
     epsilon = job.get("epsilon", PaastaColors.red("UNKNOWN"))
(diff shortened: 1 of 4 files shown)


paasta_tools/chronos_serviceinit.py
```diff
@@ -167,6 +190,7 @@ def format_chronos_job_status(job, running_tasks, verbose):
     disabled_state = _format_disabled_status(job)
     (last_result, formatted_time) = _format_last_result(job)
     schedule = _format_schedule(job)
+    parents = _format_parents(job)
```
solarkennedy: Although probably not the simplest to do with how the code is now. Can you make this only show up in verbose mode?

As we use chronos more, paasta status is going to start getting really long. Most of the time people won't be interested in this data, and if they *are* inspecting a particular instance, *then* I think this detailed parent status will be useful.

solarkennedy:
Can you also merge master to simplify and get test fixes?

Rob-Johnson:
Ok, I've updated. If anyone has any better ideas for the wording on non-verbose dependent jobs, then I'm open to suggestions. 

updated screenshots:

non-verbose dependent job:
screen shot 2016-02-01 at 16 21 25

verbose dependent job:
screen shot 2016-02-01 at 16 25 01

scheduled job:
screen shot 2016-02-01 at 16 26 14

solarkennedy:
How about making the non-verbose option a bit more non-verbose:

```
Instance: examplejob.foo
  Status:  Scheduled
  Last:    OK (last run time)
  Parents: foo, bar
  Command:
  Mesos: 
```

solarkennedy:
In other words, go ahead and say the parents if you know there are some, and just have it replace the schedule part. (hopefully that is easy)

Rob-Johnson:
ok, final run:

screen shot 2016-02-03 at 14 44 52
screen shot 2016-02-03 at 14 44 59
screen shot 2016-02-03 at 14 45 05

paasta_tools/chronos_tools.py
```diff
@@ -177,14 +182,6 @@ def __init__(self, service, instance, cluster, config_dict, branch_dict):
             branch_dict=branch_dict,
         )
 
-    def __eq__(self, other):
```
giuliano108: Is this officially gone?

Rob-Johnson: ah, great catch.

Rob-Johnson:
@giuliano108 corrected

paasta_tools/chronos_serviceinit.py
```diff
@@ -131,12 +131,80 @@ def _format_last_result(job):
...
+    return dispatch_formatter
+
+
+def _get_schedule_field_for_job_type(job_type):
+    if job_type == chronos_tools.JobType.Dependent:
+        return 'Parents'
+    elif job_type == chronos_tools.JobType.Scheduled:
+        return 'Schedule'
+    else:
+        return 'Unknown'
```
giuliano108: I was expecting this `raise`...

solarkennedy:
ship

PR #201, opened 2016-01-28
made cleanup_chronos_jobs remove all non-paasta jobs

mjksmith:
This should make deploying robj's change easier and also prevent random chronos jobs from mixing in with paasta jobs in the future.

This also makes `cleanup_chronos_jobs` behave more like `cleanup_marathon_jobs`.


solarkennedy:
ship

PR #202, opened 2016-01-28
metastatus now prints an error when marathon is configured & unreachable

mjksmith:


paasta_tools/paasta_metastatus.py
```diff
@@ -349,8 +350,12 @@ def main():
         chronos_results = [('chronos is not configured to run here', True)]
 
     if marathon_config:
-        marathon_client = get_marathon_client(marathon_config)
-        marathon_results = get_marathon_status(marathon_client)
+        try:
+            marathon_client = get_marathon_client(marathon_config)
+            marathon_results = get_marathon_status(marathon_client)
+        except MarathonError:
```
solarkennedy: Can you make sure the error text is still printed?

solarkennedy:
fix +ship

paasta_tools/paasta_metastatus.py
```diff
@@ -349,8 +350,12 @@ def main():
         chronos_results = [('chronos is not configured to run here', True)]
 
     if marathon_config:
-        marathon_client = get_marathon_client(marathon_config)
-        marathon_results = get_marathon_status(marathon_client)
+        try:
+            marathon_client = get_marathon_client(marathon_config)
+            marathon_results = get_marathon_status(marathon_client)
```
Rob-Johnson: only wrap ``marathon_results = get_marathon_status(marathon_client)`` in the try/except here - creating the client doesn't throw an exception.

Rob-Johnson:
whilst you're here, could you do the same for chronos?

PR #203, opened 2016-01-28
Use acutal running tasks for marathon_services_running_here_works itest

solarkennedy:
Primary: @EvanKrall 

Seems so obvious amirite?

I've kept the get_local_slave_state but only if the assertion fails, I need it to debug why this didn't work.


solarkennedy:
@mjksmith this now includes a fix that came up as a mix between the new chronos-job sensu cleanup + the new cleanup_chronos_jobs non-paasta deletion behavior.

EvanKrall:
lgtm

mjksmith:
lgtm

PR #204, opened 2016-01-28
make release no longer breaks PEP8

mjksmith:
Currently running `make release` leaves setup.py in a state that doesn't conform to PEP8. This makes it so we have to run pre-commit twice when pushing a new release.

That sucks so I fixed it.


solarkennedy:
ship

PR #206, opened 2016-01-29
made paasta start/stop deploy-group aware

mjksmith:
This fixes a bug where paasta start and stop wouldn't work on tasks using deploy groups.

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -20,21 +20,12 @@ from service_configuration_lib import DEFAULT_SOA_DIR
 
 from paasta_tools import remote_git
 from paasta_tools import utils
+from paasta_tools.chronos_tools import load_chronos_job_config
 from paasta_tools.cli.utils import figure_out_service_name
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_instances
 from paasta_tools.cli.utils import list_services
-from paasta_tools.generate_deployments_for_service import get_instance_config_for_service
-
-
-def get_branches(service, soa_dir):
-    paasta_control_branches = set(('paasta-%s' % config.get_branch()
-                                   for config in get_instance_config_for_service(soa_dir, service)))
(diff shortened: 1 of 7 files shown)


paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -139,19 +131,48 @@ def paasta_start_or_stop(args, desired_state):
...
+            service=service,
+            instance=instance,
+            cluster=cluster,
+            soa_dir=soa_dir,
+            load_deployments=False,
+        )
+    else:
+        print "A service config other than type 'marathon' or 'chronos' was loaded!"
+        print "'%s' is not currently supported by paasta %s" % (instance_type, desired_state)
+        sys.exit(3)
```
solarkennedy: Yuck. We already have almost literally this same code in 
paasta_tools/cli/cmds/local_run.py:550
Can you move that to utils and share that instead?

paasta_tools/generate_deployments_for_service.py
```diff
@@ -139,6 +139,7 @@ def get_deploy_group_mappings(soa_dir, service, old_mappings):
     remote_refs = remote_git.list_remote_refs(git_url)
 
     for control_branch, deploy_group in deploy_group_branch_mappings.items():
+        print control_branch, deploy_group
```
solarkennedy: probably don't need this right?

solarkennedy:
We have the itest for this now, and it passes

https://travis-ci.org/Yelp/paasta/jobs/105775872#L208-L224

So that gives credit that this does work as intended.

I would like @EvanKrall to look at this as well. My main concern is the whole `paasta-` thing again, but if the itests pass and generate deployments can read it back out, then I say go for it.

Can you add a deploy_group aware assertion in the deployments.json itest?

mjksmith:
AFAIK the deployments.json itest already has an instance that uses deploy groups.

solarkennedy:
afaict this is good to go. The only think I'm worried about is the whole `paasta-` thing still.

Lets ship this carefully together. Merge when you are ready.

PR #207, opened 2016-01-29
added deploy_group to valid schemas for paasta check

mjksmith:
Config files that use deploy groups will no longer fail paasta check.


nhandler:
Ship It!

PR #209, opened 2016-01-30
Gracefully handle killing a marathon task that is already dead. PAASTA-1219

solarkennedy:
This has become pretty easy now that we are wrapping kill_task.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -971,5 +971,8 @@ def kill_task(client, app_id, task_id, scale):
         if e.error_message == 'Bean is not valid' and e.status_code == 422:
             log.debug("Probably tried to kill a task id that didn't exist. Continuing.")
             return []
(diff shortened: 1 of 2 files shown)


Rob-Johnson:
I'd argue this needs to be a level up (i.e. in the bounce code, rather than our kill_task wrapper). I don't agree with the assumption that this is *always* going to be ok to ignore.

solarkennedy:
Um. I could see the argument for that, at the same time, in *paasta* I can't think of a situation where we care that much. (I agree I wouldn't, say, put this in `marathon-python`. @EvanKrall thoughts?

EvanKrall:
We only call `kill_task` from one spot (aside from tests): in setup_marathon_job.py, so it doesn't particularly matter currently.

And since the docstring for marathon_tools.kill_task says `"""Wrapper to the official kill_task method that is tolerant of errors"""`, I think it's fine to put this error tolerance into our kill_task wrapper.

PR #210, opened 2016-01-30
Catch the race between inspecting a marathon app that isn't running during a bounce. PAASTA-1721

solarkennedy:
During a bounce you may not have a marathon app running, or it may have just went away.

We guard if the app_id is not running in the first place on non-verbose mode, so here is me doing it for verbose mode.


Rob-Johnson:
shipit

PR #211, opened 2016-01-30
Fix regression from https://github.com/Yelp/paasta/pull/179 where we …

EvanKrall:
…would kill apps with unhealthy tasks before letting them drain properly.

Since we were ignoring `old_app_live_unhappy_tasks` in calculating `apps_to_kill`, we would kill apps if all of their tasks were unhappy, even though we just started draining those tasks.

This patch makes it so we wait until the tasks are properly drained before killing them.


solarkennedy:
ship

PR #215, opened 2016-02-01
Use full GPG fingerprints

### LINKED ISSUE #212: Stop Using Short GPG Key IDs
There are a few places where we add a GPG key for apt to use to verify repositories. We are currently using short key IDs to do this. To avoid http://www.asheesh.org/note/debian/short-key-ids-are-bad-news.html , we should switch to using long key ids.

nhandler:
Use the full GPG fingerprint rather than just the short id. This avoids
accidentally adding the wrong key per
http://www.asheesh.org/note/debian/short-key-ids-are-bad-news.html
Closes #212


solarkennedy:
ship if pass

Issue #217, opened 2016-02-02
update chronos_schema.json to support dependent jobs


giuliano108:
```
gcioffi@devbox:~/git/yelpsoa-configs/example_service (master) $ paasta check
[..]
✗ Failed to validate schema. More info: http://paasta.readthedocs.org/en/latest/yelpsoa_configs.html: /nail/etc/services/example_service/chronos-mesosstage.yaml
  Validation Message: Additional properties are not allowed ('parents' was unexpected)
[..]
```

nhandler:
https://gist.github.com/zakgrant/5900025 has some comments about how you might handle doing this.

Rob-Johnson:
@giuliano108 if we want to be really hot on this, it'd be good to verify that the parent, if a string, is in the form 'service.instance' (can we also verify this for each parent in a list)?

PR #219, opened 2016-02-02
allow a string or array value for the 'parents' key of a dependent job

### LINKED ISSUE #149: allow a string or array value for the 'parents' key of a dependent job
Currently we only allow an array, but it'd be nice if the user could specify it as a string in the case that there is only one parent

giuliano108:
Allow `parents` jobs to be passed in the config as an array or a string, see #149.
Note that the `parents` key isn't currently schema-validated (will be fixed in #217).

Closes #149

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -176,6 +176,7 @@ class ChronosJobConfig(InstanceConfig):
             config_dict=config_dict,
             branch_dict=branch_dict,
         )
+        self.__normalize()
 
     def __eq__(self, other):
         return (
@@ -184,6 +185,13 @@ class ChronosJobConfig(InstanceConfig):
             (self.config_dict == other.config_dict) and
             (self.branch_dict == other.branch_dict)
         )
+
+    def __normalize(self):
+        """When the config supports alternative syntaxes, convert them
+        to a normal form."""
+        parents = self.get_parents()
+        if parents is not None and not isinstance(parents, list):
+            self.config_dict['parents'] = [parents]
 
     def get_service(self):
         return self.service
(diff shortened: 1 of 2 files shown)


paasta_tools/chronos_tools.py
```diff
@@ -185,6 +186,13 @@ def __eq__(self, other):
             (self.branch_dict == other.branch_dict)
         )
 
+    def __normalize(self):
+        """When the config supports alternative syntaxes, convert them
+        to a normal form."""
+        parents = self.get_parents()
+        if parents is not None and not isinstance(parents, list):
+            self.config_dict['parents'] = [parents]
```
solarkennedy: It seems a little strange to me that it mutates the config_dict.

What if you moved this code to the `get_parents` getter?

giuliano108: > What if you moved this code to the get_parents getter?

I thought about that, but the [\_\_eq\_\_](https://github.com/Yelp/paasta/blob/c57563ea7a6c6ab1d80bc03c5bb169581e3e8fc9/paasta_tools/chronos_tools.py#L181) method does something like "`self.config_dict == other.config_dict`" when testing `ChronosJobConfig` objects for equality. If I don't "normalize" the config, alternative syntaxes would fail the `__eq__` test even when they express the same thing in different ways (`parents: 'blah'` vs. `parents: ['blah']`).

Happy to patch the getter, if I'm overthinking. :)

solarkennedy: For now, junk the `eq` method. We don't use it and we don't normalize all the other things. Stick with a get_parents getter like most of our other code so that get_parents() always returns something consistent.

Rob-Johnson:
shipit!

PR #220, opened 2016-02-02
Stopped printing dashboards with metastatus

solarkennedy:
I think we are ready to stop doing this. The whole prod dashboards thing is misleading anyway, people are typing in ldap creds for the marathon RW dashboards, and it is all yelpy.


Rob-Johnson:
shipit

solarkennedy:
discarding in favor of a config file approach.

PR #225, opened 2016-02-03
Added generate-pipeline warning

solarkennedy:
There is lots of expectations and confusion about what this tool does, and we have had a couple of times where people have run it, not-knowing that it overwrites customization.

I would like to add this warning to make that more explicit.

### DIFF
--- a/paasta_tools/cli/cmds/generate_pipeline.py
+++ b/paasta_tools/cli/cmds/generate_pipeline.py
@@ -80,6 +80,20 @@ def get_git_repo_for_fab_repo(service):
     return repo
 
 
+def print_warning():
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/generate_pipeline.py
```diff
@@ -80,6 +80,20 @@ def get_git_repo_for_fab_repo(service):
     return repo
 
 
+def print_warning():
+    print "Warning: paasta generate-pipeline is a Yelp-specific tool and will"
+    print "not work outside Yelp."
+    print ""
+    print "paasta generate-pipeline is designed only to give a starting jenkings"
```
asottile: jenkings

solarkennedy:
I merged this manually.

Issue #223, opened 2016-02-03
Control tags besides start/stop are all interpreted as stop.


EvanKrall:
The change in #190 created tags like `paasta-norcal-prod.main-20160203T005046-deploy`. This results in a `desired_state` of `deploy` in deployments.json for the service in question. Then [marathon_tools interprets `deploy` as `!= "start"`](https://github.com/Yelp/paasta/blob/7be0ed698bc26b8eb840348ad001718e95539b18/paasta_tools/marathon_tools.py#L197), so the service in question is stopped.

We should potentially:
 - whitelist desired_state to start or stop in generate_deployments_for_service.py
 - blacklist `...-deploy` tags in generate_deployments_for_service.py
 - have the reimplementation of #190 create a different format of tag that doesn't get interpreted as a state change.

nhandler:
My vote would be for the whitelist. It feels like the safest solution. We also aren't going to be adding new types of tags that will require updating this list too often. I am fine actioning this when I fix up #190

solarkennedy:
How about the code that is responsible for looking up the desired state, only match the regex it is expecting? `.*(start|stop)$`

That way we can add more tags at will, and the right code can look out for them? (docker image code can look for the deploy tag, desired_state can look for start/stop, some future certification code can look for the certify tag, etc.

PR #227, opened 2016-02-04
Add a new itest to assert new services have a desired_state of start

nhandler:
This new itest should avoid us adding code that modifies deployments.json in such a way that new services no longer have a desired_state of 'start'. This should help avoid incidents like #223 from happening again.

### DIFF
--- a/general_itests/deployments_json.feature
+++ b/general_itests/deployments_json.feature
@@ -5,3 +5,8 @@ Feature: Per-Service Deployments.json can be written and read back
 	When paasta stop is run against the repo
 	 And we generate deployments.json for that service
         Then that deployments.json can be read back correctly
+
+    Scenario: New services have a desired_state of "start"
+	Given a test git repo is setup with commits
+	 When we generate deployments.json for that service
+	Then that deployments.json has a desired_state of "start"
--- a/general_itests/steps/deployments_json_steps.py
+++ b/general_itests/steps/deployments_json_steps.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 import contextlib
 import os
-import shutil
 import tempfile
 from datetime import datetime
 from time import time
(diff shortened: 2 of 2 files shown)


general_itests/deployments_json.feature
```diff
@@ -5,3 +5,8 @@ Feature: Per-Service Deployments.json can be written and read back
 	When paasta stop is run against the repo
 	 And we generate deployments.json for that service
         Then that deployments.json can be read back correctly
+
+    Scenario: New services have a desired_state of "start"
+	Given a test git repo is setup with commits
```
solarkennedy: Does this work with the existing git-repo in place? In the previous scenario we run paasta "stop", so I guess we .. somehow clear out the git repo between scenarios?

nhandler: My understanding is that the context is cleared between the scenarios (and the tests appear to confirm that). This is also why I removed the `shutil.rmtree` call in the corresponding step (it would cause issues if you try and chain multiple 'then' steps).

solarkennedy: Cool.

general_itests/steps/deployments_json_steps.py
```diff
@@ -124,4 +123,11 @@ def step_impl_then(context):
         },
     }
     assert expected_deployments == deployments, "actual: %s\nexpected:%s" % (deployments, expected_deployments)
-    shutil.rmtree(context.test_git_repo_dir)
+
+
+@then(u'that deployments.json has a desired_state of "{state}"')
```
solarkennedy: can you call this `expected_state` instead of just `state` here?
I think it reads better when you do the assertion

solarkennedy:
Ship. When you do tdd on your tags thing, this test *should* fail. (will read back "deploy" as the desired state) then write some safeguard code and it should pass.

PR #230, opened 2016-02-04
Upgrade mesos.interface to 0.23.1

nhandler:
Upgrade mesos.interface to 0.23.1 per step 3 on http://paasta.readthedocs.org/en/latest/upgrading_mesos.html


solarkennedy:
Can you confirm we are in .23 everywhere? This should be the last step so we don't use .23 protobufs on .22 clusters.

nhandler:
Yep, I have confirmed that all of our PaaSTA boxes are on 0.23.1.

solarkennedy:
ship

PR #231, opened 2016-02-04
Mark for deployment tags take2

### LINKED ISSUE #223: Control tags besides start/stop are all interpreted as stop.

nhandler:
Fix issues where a tag ending in '-deploy' was interpreted as stop

 * Fix an issue with the recently added itest to make it actually call mark-for-deployments before testing the desired_state
 * Add a whitelist that will prevent states other than 'start' and 'stop' from getting written to deployments.json (we will add 'deploy' later)

Without the `valid_states` code, the 'New services have a desired_state of "start"' itest fails as expected. With the new code, it passes.

Closes #223

### DIFF
--- a/paasta_tools/generate_deployments_for_service.py
+++ b/paasta_tools/generate_deployments_for_service.py
@@ -182,9 +182,7 @@ def get_desired_state(service, branch, remote_refs, deploy_group):
     """
     # (?:paasta-){1,2} supports a previous mistake where some tags would be called
     # paasta-paasta-cluster.instance
-    tag_pattern = r'^refs/tags/(?:paasta-){1,2}%s-(?P<force_bounce>[^-]+)-(?P<state>.*)$' % branch
-
-    valid_states = ['start', 'stop']
+    tag_pattern = r'^refs/tags/(?:paasta-){1,2}%s-(?P<force_bounce>[^-]+)-(?P<state>(start|stop))$' % branch
 
     states = []
     head_sha = remote_refs['refs/heads/paasta-%s' % deploy_group]
@@ -194,8 +192,7 @@ def get_desired_state(service, branch, remote_refs, deploy_group):
             match = re.match(tag_pattern, ref_name)
             if match:
                 gd = match.groupdict()
-                if gd['state'] in valid_states:
-                    states.append((gd['state'], gd['force_bounce']))
+                states.append((gd['state'], gd['force_bounce']))
 
     if states:
         # there may be more than one that matches, so take the one that sorts
--- a/paasta_tools/generate_deployments_for_service.py
+++ b/paasta_tools/generate_deployments_for_service.py
@@ -150,7 +150,6 @@ def get_deploy_group_mappings(soa_dir, service, old_mappings):
             mapping['docker_image'] = docker_image
 
             desired_state, force_bounce = get_desired_state(
-                service=service,
                 branch=control_branch,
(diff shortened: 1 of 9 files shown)


paasta_tools/generate_deployments_for_service.py
```diff
@@ -184,6 +184,8 @@ def get_desired_state(service, branch, remote_refs, deploy_group):
     # paasta-paasta-cluster.instance
     tag_pattern = r'^refs/tags/(?:paasta-){1,2}%s-(?P<force_bounce>[^-]+)-(?P<state>.*)$' % branch
 
+    valid_states = ['start', 'stop']
```
solarkennedy: I think this is more efficient if you put this in the regex in line 185 `(start|stop)`

That way you are only even looking at refs you care about

paasta_tools/generate_deployments_for_service.py
```diff
@@ -184,6 +184,8 @@ def get_desired_state(service, branch, remote_refs, deploy_group):
     # paasta-paasta-cluster.instance
```
solarkennedy: It looks like we have 0 coverage on this (important) function. Can you write at least one test for this?
```
def test_get_disired_state_understands_tags():
   remote_refs = (get some legit remote refs with all sorts of crazy stuff)
  service = fake_service
  branch = ?
  deploy_group = ?
  expected_desired_state = 'stopped'
  actual = ....


def get_desired_state_fails_gracefully_with_start():
  ....
  actual = ...
  assert actual == ("start", None)

```

paasta_tools/utils.py
```diff
@@ -1016,6 +1016,26 @@ def get_paasta_branch(cluster, instance):
     return get_paasta_branch_from_deploy_group('%s.%s' % (cluster, instance))
 
 
+def format_timestamp(dt=None):
+    if dt is None:
+        dt = datetime.datetime.utcnow()
+    return dt.strftime('%Y%m%dT%H%M%S')
+
+
+def get_paasta_tag_from_deploy_group(identifier, desired_state):
```
solarkennedy: These two functions probably need to become one eventually.

paasta_tools/generate_deployments_for_service.py
nhandler: I added these tests.

paasta_tools/utils.py
nhandler: There is quite a bit of refactoring that should get done. I was going to hold off until I finish the second half of this change before doing it.

solarkennedy:
You can ship, you can also do a manual verification of this by activating this code, running paasta mark-for-deployment on a service, which will make a special tag, then run generate-deployments, all in your local checkout, and verify that deployments.json is correct on a real service.

PR #235, opened 2016-02-05
paasta status - show the value of schedule_time_zone (PAASTA-2389)

giuliano108:
When no TZ is specified:
screen shot 2016-02-05 at 5 34 53 pm

PST:
screen shot 2016-02-05 at 5 35 28 pm

### DIFF
--- a/paasta_tools/chronos_serviceinit.py
+++ b/paasta_tools/chronos_serviceinit.py
@@ -138,8 +138,8 @@ def _format_schedule(job):
     epsilon = job.get("epsilon", PaastaColors.red("UNKNOWN"))
     schedule_time_zone = job.get("scheduleTimeZone", "null")
     if schedule_time_zone == "null":  # This is what Chronos returns.
-        schedule_time_zone = PaastaColors.grey("unspecified")
-    formatted_schedule = "%s Epsilon: %s TZ: %s" % (schedule, epsilon, schedule_time_zone)
+        schedule_time_zone = "UTC"
+    formatted_schedule = "%s (%s) Epsilon: %s" % (schedule, schedule_time_zone, epsilon)
     return formatted_schedule
 
 
--- a/paasta_tools/chronos_serviceinit.py
+++ b/paasta_tools/chronos_serviceinit.py
@@ -136,7 +136,10 @@ def _format_schedule(job):
     else:
         schedule = job.get("schedule", PaastaColors.red("UNKNOWN"))
     epsilon = job.get("epsilon", PaastaColors.red("UNKNOWN"))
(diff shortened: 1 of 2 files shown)


solarkennedy:
This is just a small cosmetic change, but can you make it say:

```
Schedule:  R/..../P2M (UTC)  Epsilon: 60S
or
Schedule: R/...P2M (PST) Epsilong: bla
```

That way the tz is right after the time, which is how it usually is displayed. That and we are more explicit about utc instead of unspecified.

giuliano108:
> That way the tz is right after the time, which is how it usually is displayed. That and we are more explicit about utc instead of unspecified.

Fair points, here's another attempt:

screen shot 2016-02-05 at 7 13 09 pm

screen shot 2016-02-05 at 7 14 31 pm

paasta_tools/chronos_serviceinit.py
```diff
@@ -136,7 +136,10 @@ def _format_schedule(job):
     else:
         schedule = job.get("schedule", PaastaColors.red("UNKNOWN"))
     epsilon = job.get("epsilon", PaastaColors.red("UNKNOWN"))
-    formatted_schedule = "%s Epsilon: %s" % (schedule, epsilon)
+    schedule_time_zone = job.get("scheduleTimeZone", "null")
+    if schedule_time_zone == "null":  # This is what Chronos returns.
```
solarkennedy: Yuck. Literally the string? :(

paasta_tools/chronos_serviceinit.py
```diff
@@ -136,7 +136,10 @@ def _format_schedule(job):
     else:
         schedule = job.get("schedule", PaastaColors.red("UNKNOWN"))
     epsilon = job.get("epsilon", PaastaColors.red("UNKNOWN"))
-    formatted_schedule = "%s Epsilon: %s" % (schedule, epsilon)
+    schedule_time_zone = job.get("scheduleTimeZone", "null")
```
solarkennedy: What if you just do job.get("sched", "UTC") ?

giuliano108: I meant to defensively cater for a case where `job.get("scheduleTimeZone")` returns `None` (which might not present itself, agreed). The code treats `"null"` and `None` in the same way, defaulting to `UTC`...

giuliano108: > Yuck. Literally the string? :(

Yep:

```
ipdb> job.get('scheduleTimeZone',None)
u'null'
ipdb>
```

Rob-Johnson:
@giuliano108 can we get some unit tests?

giuliano108:
> can we get some unit tests?
Done.

Rob-Johnson:
shipit

PR #236, opened 2016-02-05
Add extra_constraints, which doesn't override default constraints.

EvanKrall:
Also clarify docs on constraints.


tests/test_marathon_tools.py
```diff
@@ -1030,7 +1030,7 @@ def test_get_constraints_in_config(self):
             service='fake_name',
```
solarkennedy: can you rename line 1027 to "test_get_constraints_in_config_overrides_other_constraints" or something like that?

solarkennedy:
fix+ship

PR #238, opened 2016-02-05
Added experimental manual mesos task reconciliation script (orphan killer)

solarkennedy:
This is an experimental script that looks at the *slave* state and kills orphaned containers.

With the new mesos upgrade, it left behind a bunch. Some of the bug reports suggest that when you do upgrades to blow away the metadata..... In the future after we do an upgrade we can run this script in dry-run mode to see?

Comments welcome, but I want to ship this before the weekend so our boxes don't continue to have really full disks, bogus orphaned tasks running everywhere, and clusters out of resources.

Longer term maybe we can run this as a monitoring check, returning 0 if there is no bogus containers, and 1 if there is some.

### DIFF
--- a/paasta_tools/contrib/kill_orphaned_docker_containers.py
+++ b/paasta_tools/contrib/kill_orphaned_docker_containers.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env python
+import argparse
+
+from docker import Client
+
+from paasta_tools import mesos_tools
+from paasta_tools.utils import get_docker_host
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description=(
+            'Cross references running containers with task ids from the mesos slave',
+            ' and optionally kills them.'
+        )
+    )
+    parser.add_argument('-f', '--force', help="Actually kill the containers. (defaults to dry-run)")
+    args = parser.parse_args()
+    return args
+
+
+def get_running_task_ids_from_mesos_slave():
+    state = mesos_tools.get_local_slave_state()
+    frameworks = state.get('frameworks')
+    executors = [ex for fw in frameworks for ex in fw.get('executors', [])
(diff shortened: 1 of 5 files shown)


paasta_tools/contrib/kill_orphaned_docker_containers.py
```diff
@@ -0,0 +1,61 @@
...
+
+    for container in running_mesos_docker_containers:
+        mesos_task_id = mesos_tools.get_mesos_id_from_container(container=container, client=docker_client)
+        print mesos_task_id
+        if mesos_task_id not in running_mesos_task_ids:
+            if args.force:
+                print "Killing %s. (%s)" % (container["Names"][0], mesos_task_id)
+                docker_client.kill(container)
+            else:
+                print "Would kill %s. (%s)" % (container["Names"][0], mesos_task_id)
```
nhandler: Nitpick: Could you hint at using the --force option in this message?

nhandler:
+1 for adding the monitoring check. That would have been very helpful in catching/fixing this issue.

nhandler:
Ship It!

paasta_tools/contrib/kill_orphaned_docker_containers.py
```diff
@@ -0,0 +1,61 @@
...
+    args = parser.parse_args()
+    return args
+
+
+def get_running_task_ids_from_mesos_slave():
+    state = mesos_tools.get_local_slave_state()
+    frameworks = state.get('frameworks')
+    executors = [ex for fw in frameworks for ex in fw.get('executors', [])
+                 if u'TASK_RUNNING' in [t[u'state'] for t in ex.get('tasks', [])]]
+    return [e["id"] for e in executors]
```
mjksmith: return a set here to make this run in linear time instead of quadratic

PR #239, opened 2016-02-05
reduced kill_orphaned_docker_containers runtime

mjksmith:
reduced runtime  by casting running container list to a set

We check for the membership of every running docker container id against the list of running containers that are in mesos. Since checking membership in a list is an O(n) operation, this runs in quadratic time. Changing to a set reduces the runtime to linear since sets have constant-time lookup.


solarkennedy:
Nice. Ship.

PR #241, opened 2016-02-08
Add support for disk resource isolation

nhandler:
This is a first pass at adding this functionality. I went through trying to find all of the places we deal with cpu/mem and adjusted the code to also deal with disk space. I'm sure I've missed a spot or two in the code, but all of the tests are currently passing. I wanted to get this pull request tossed up so that I can get some feedback.

### DIFF
--- a/paasta_tools/check_mesos_resource_utilization.py
+++ b/paasta_tools/check_mesos_resource_utilization.py
@@ -70,9 +70,11 @@ def check_thresholds(percent):
     output = ""
     current_mem = stats['master/mem_percent']
     current_cpu = stats['master/cpus_percent']
+    current_disk = stats['master/disk_percent']
     percent = int(percent)
     cpu_print_tuple = (percent, current_cpu)
     mem_print_tuple = (percent, current_mem)
+    disk_print_tuple = (percent, current_disk)
     if current_mem >= percent:
         output += "CRITICAL: Memory usage is over %d%%! Currently at %f%%!\n" % mem_print_tuple
         over_threshold = True
@@ -83,6 +85,11 @@ def check_thresholds(percent):
         over_threshold = True
     else:
         output += "OK: CPU usage is under %d%%. (Currently at %f%%)\n" % cpu_print_tuple
+    if current_disk >= percent:
+        output += "CRITICAL: Disk usage is over %d%%! Currently at %f%%!\n" % disk_print_tuple
+        over_threshold = True
+    else:
+        output += "OK: Disk usage is under %d%%. (Currently at %f%%)\n" % disk_print_tuple
     if over_threshold is True:
         status = 2
     else:
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -107,6 +107,15 @@ class InstanceConfig(dict):
         cpus = self.config_dict.get('cpus')
         return cpus if cpus else .25
 
+    def get_disk(self):
+        """Gets the  amount of disk space required from the service's configuration.
+
+        Defaults to 1024 (1G) if no value is specified in the config.
+
+        :returns: The amount of disk space specified by the config, 1024 if not specified"""
+        disk = self.config_dict.get('disk')
+        return disk if disk else 1024
+
     def get_cmd(self):
         """Get the docker cmd specified in the service's configuration.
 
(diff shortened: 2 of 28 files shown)


paasta_tools/check_mesos_resource_utilization.py
```diff
@@ -83,6 +85,11 @@ def check_thresholds(percent):
         over_threshold = True
     else:
         output += "OK: CPU usage is under %d%%. (Currently at %f%%)\n" % cpu_print_tuple
+    if current_disk >= percent:
+        output += "CRITICAL: Disk usage is over %d%%! Currently at %f%%!\n" % disk_print_tuple
```
Rob-Johnson: can you add the equivalent of https://github.com/Yelp/paasta/pull/237here please?

paasta_tools/utils.py
```diff
@@ -201,6 +210,13 @@ def check_mem(self):
                 return False, 'The specified mem value "%s" is not a valid float.' % mem
         return True, ''
 
+    def check_disk(self):
+        disk = self.get_disk()
+        if disk is not None:
+            if not isinstance(disk, float) and not isinstance(disk, int):
+                return False, 'The specified disk value "%s" is not a valid float.' % disk
```
Rob-Johnson: this isn't valid - the value can be a float or an int.

paasta_tools/utils.py
```diff
@@ -107,6 +107,15 @@ def get_cpus(self):
...
         return cpus if cpus else .25
 
+    def get_disk(self):
+        """Gets the  amount of disk space required from the service's configuration.
+
+        Defaults to 1024 (1G) if no value is specified in the config.
+
+        :returns: The amount of disk space specified by the config, 1024 if not specified"""
+        disk = self.config_dict.get('disk')
+        return disk if disk else 1024
```
Rob-Johnson: this seems...small. can be make it 10G?

Rob-Johnson:
can you also update the docs for yelpsoa-configs?

paasta_tools/utils.py
solarkennedy: Only a small minority of services are that big. (run `docker images` on a box)

solarkennedy:
This looks really comprehensive! Fix docs and ship i say.
This *will* cause a big bounce, but with lucy temporarily suspended, I'm not afraid.

paasta_tools/utils.py
nhandler: The conditional allows for floats and ints. It is only the error message that is specific to floats. This code is based on the check above for memory. I can fix up both of the messages to be a bit more clear.

paasta_tools/utils.py
mjksmith: Can you use the default value of the `dict.get` method to provide the default value? i.e. `self.config_dict.get('disk', 1024)`

paasta_tools/utils.py
```diff
@@ -191,14 +200,21 @@ def check_cpus(self):
         cpus = self.get_cpus()
         if cpus is not None:
             if not isinstance(cpus, float) and not isinstance(cpus, int):
```
mjksmith: isinstance can be passed a tuple for its second argument so this could just be `isinstance(cpus, (float, int))`

Ditto for line 209

paasta_tools/utils.py
```diff
@@ -191,14 +200,21 @@ def check_cpus(self):
...
             if not isinstance(mem, float) and not isinstance(mem, int):
-                return False, 'The specified mem value "%s" is not a valid float.' % mem
+                return False, 'The specified mem value "%s" is not a valid float or int.' % mem
+        return True, ''
+
+    def check_disk(self):
+        disk = self.get_disk()
+        if disk is not None:
+            if not isinstance(disk, float) and not isinstance(disk, int):
+                return False, 'The specified disk value "%s" is not a valid float or int.' % disk
```
mjksmith: isinstance can be passed a tuple for its second argument so this could just be `isinstance(cpus, (float, int))`

Ditto for line 202 and 209

paasta_tools/check_mesos_resource_utilization.py
```diff
@@ -83,6 +85,11 @@ def check_thresholds(percent):
         over_threshold = True
     else:
         output += "OK: CPU usage is under %d%%. (Currently at %f%%)\n" % cpu_print_tuple
+    if current_disk >= percent:
```
EvanKrall: can you add blank lines between each if/else in here? These blocks kind of run into each other.

PR #242, opened 2016-02-08
re-ordered default deploy order in fsm

mjksmith:
This changes the default deploy order to stage -> prod canary -> prod non-canary & dev

@EvanKrall @asottile you can argue for or against this here

### DIFF
--- a/general_itests/steps/fsm_wizard_steps.py
+++ b/general_itests/steps/fsm_wizard_steps.py
@@ -76,6 +76,34 @@ def step_impl_when_fsm_auto(context):
                     'norcal-prod': 'PROD',
                     'nova-prod': 'PROD'
                 },
+                'fsm_deploy_pipeline': [
+                    {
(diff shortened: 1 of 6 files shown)


asottile:
nitpick: breaking all four of stages with a bad staging deploy isn't usually the nicest

could we do 1 stage by default and rest-of-stages after?

EvanKrall:
Pull this into a system config file plz

mjksmith:
@asottile this probably won't be implemented now because either we pick a single stage to be the default for every new service, or we make users pick a default stage when they run `fsm`. I'm hesitant to include stage-picking logic in `fsm` since it won't apply to all non-Yelp infrastructures.

mjksmith:
@EvanKrall I forgot to mention this in standup (I thought it was done) but this is ready for more review/to ship

solarkennedy:
This looks fine, just moving it to reading an external config file

PR #244, opened 2016-02-09
Use verbose flags instead of debug in paasta logs

solarkennedy:
Seems like we should be consistent about using verbose everywhere


nhandler:
Ship It!

Rob-Johnson:
excellent ship, @nhandler 

shipit, kwa.

image

PR #245, opened 2016-02-10
validate dependent jobs' parents - issue 148

### LINKED ISSUE #148: we should validate that a given service.instance parent exists before trying to deploy
We currently only check for the 'parent' key of a job being useful when we attempt to launch the job. Instead, this should (also?) be done as part of a config's validation.

giuliano108:
This should fix #148 .

Catches two types of config errors:
- Parents whose name is invalid (!= `service_name.instance_name`). Similar to what the JSON schema also enforces.
- Parents that are referenced in a dependent job, but are not actually configured.

Note that the second validation won't let a "`ServiceA` child" depend on a "`ServiceB` parent". That's the best we can do without parsing the entire `yelpsoa-configs` tree.

In my opinion `paasta_tools/cli/cmds/validate.py`, in order to validate the Chronos config, needs to know too many details about `ChronosJobConfig`. If need be, I'm happy to move as much of that logic into `paasta_tools/chronos_tools.py` as possible, so that the latter is in charge of the validation and the former just prints out the outcome.

screen shot 2016-02-10 at 6 08 14 pm

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -203,7 +203,7 @@ def validate_chronos(service_path):
 
     returncode = True
     for cluster in list_clusters(service, soa_dir, instance_type):
-        services_in_cluster = get_services_for_cluster(cluster=cluster, instance_type='chronos')
+        services_in_cluster = get_services_for_cluster(cluster=cluster, instance_type='chronos', soa_dir=soa_dir)
         valid_services = set([name + '.' + instance for name, instance in services_in_cluster])
         for instance in list_all_instances_for_service(
                 service=service, clusters=[cluster], instance_type=instance_type,
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -16,7 +16,6 @@ import json
 import os
 import pkgutil
 import sys
-from collections import OrderedDict
 from glob import glob
 
 import yaml
@@ -24,6 +23,7 @@ from jsonschema import Draft4Validator
 from jsonschema import FormatChecker
 from jsonschema import ValidationError
 
+from paasta_tools.chronos_tools import check_parent_format
 from paasta_tools.chronos_tools import load_chronos_job_config
 from paasta_tools.cli.utils import failure
 from paasta_tools.cli.utils import get_file_contents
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -205,16 +207,27 @@ def validate_chronos(service_path):
                 service=service, clusters=[cluster], instance_type=instance_type,
                 soa_dir=soa_dir):
             cjc = load_chronos_job_config(service, instance, cluster, False, soa_dir)
-            checks_passed, check_msgs = cjc.validate()
+            cjcs[service + '.' + cjc.get_job_name()] = cjc
```
solarkennedy: This looks a little more complicated than I think it needs to be. While not super efficient, can it just be:
```
for parent in get_parents()
    if parent not in get_services_for_cluster(cluster=cluster, instance_type=chronos):
        # something
```

giuliano108:
> This looks a little more complicated than I think it needs to be.

Yes. Using `get_services_for_cluster` also allows parent/dependent validation to work across services in the same cluster. In this new revision, I'm also making sure validation fails when a dependent job specifies itself as a parent.

screen shot 2016-02-15 at 6 42 53 pm

paasta_tools/cli/cmds/validate.py
```diff
@@ -201,12 +203,25 @@ def validate_chronos(service_path):
 
     returncode = True
     for cluster in list_clusters(service, soa_dir, instance_type):
+        services_in_cluster = get_services_for_cluster(cluster=cluster, instance_type='chronos')
+        valid_services = set([name + '.' + instance for name, instance in services_in_cluster])
```
Rob-Johnson: "%s%s%s" % (service, ".", instance)

paasta_tools/cli/cmds/validate.py
```diff
@@ -201,12 +203,25 @@ def validate_chronos(service_path):
...
                 service=service, clusters=[cluster], instance_type=instance_type,
                 soa_dir=soa_dir):
             cjc = load_chronos_job_config(service, instance, cluster, False, soa_dir)
+            parents = cjc.get_parents() or []
             checks_passed, check_msgs = cjc.validate()
 
+            for parent in parents:
+                if not check_parent_format(parent):
+                    continue
+                if service + '.' + instance == parent:
```
Rob-Johnson: "%s%s%s" % (service, ".", instance)

solarkennedy:
lgtm

Rob-Johnson:
fix n ship

PR #249, opened 2016-02-11
Added experimental script to kill old marathon deployments

solarkennedy:
I don't know "why" this has started happening, and I don't consider this to be a normal condition, but somehow we get these marathon deployments that are left behind for a long period of time, and they seem to harm bounces.

This seems like a good start at remediation, just killing deployments that have been around for a long time?

Almost all the time what is left behind are the remnants of the "kill and scale" procedures we do, so I "think" this is safe? I'll start by manually running this in a few clusters to kick them, and then maybe automate this.

### DIFF
--- a/paasta_tools/contrib/delete_old_marathon_deployments.py
+++ b/paasta_tools/contrib/delete_old_marathon_deployments.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+# Copyright 2015 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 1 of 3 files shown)


paasta_tools/contrib/delete_old_marathon_deployments.py
```diff
@@ -0,0 +1,90 @@
...
+    """Return the :class:`datetime.datetime.DateTime` for a time in the past.
+    :param value: a string containing a time format supported by :mod:`pytimeparse`
+    """
+    if value is None:
+        return None
+    return datetime_seconds_ago(timeparse.timeparse(value))
+
+
+def datetime_seconds_ago(seconds):
+    now = datetime.datetime.now(tz.tzutc())
```
nhandler: Use your now() function

nhandler:
Fix and Ship It!
Fix and Ship It!_at_the_San_Diego_Bay_04.jpg)

PR #254, opened 2016-02-11
Update paasta check sensu_check and smartstack_check to pass soa_dir

nhandler:
It was pointed out that when `paasta check` tests for a team and proxy_port, it displays a failure when a custom soa_dir is specified. This PR properly passes the soa_dir argument to fix that issue.


solarkennedy:
ship if the test pass

EvanKrall:
🐑it

PR #255, opened 2016-02-12
enabled scaling without bounces when editing soa_configs

mjksmith:
for some reason scaling up using `scale_app` doesn't work unless I set `force=True`. IDK enough to determine if this is bad

### DIFF
--- a/paasta_itests/steps/setup_marathon_job_steps.py
+++ b/paasta_itests/steps/setup_marathon_job_steps.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import contextlib
+import time
 
 import mock
 from behave import then
@@ -51,14 +52,14 @@ fake_service_config = {
             {'hostPath': u'/nail/srv', 'containerPath': '/nail/srv', 'mode': 'RO'},
             {'hostPath': u'/etc/boto_cfg', 'containerPath': '/etc/boto_cfg', 'mode': 'RO'}]
     },
-    'instances': 1,
+    'instances': 3,
     'mem': 300,
     'args': [],
     'backoff_factor': 2,
     'cpus': 0.25,
     'uris': ['file:///root/.dockercfg'],
     'backoff_seconds': 1,
-    'constraints': None
+    'constraints': None,
 }
 
 
@@ -69,10 +70,12 @@ def create_complete_app(context):
         mock.patch('paasta_tools.marathon_tools.load_marathon_config', return_value=context.marathon_config),
         mock.patch('paasta_tools.marathon_tools.load_system_paasta_config', return_value=context.system_paasta_config),
         mock.patch('paasta_tools.bounce_lib.load_system_paasta_config', return_value=context.system_paasta_config),
+        mock.patch('paasta_tools.setup_marathon_job.send_sensu_bounce_keepalive', autospec=True),
         mock.patch('paasta_tools.setup_marathon_job.load_system_paasta_config',
                    return_value=context.system_paasta_config),
     ) as (
         mock_create_complete_config,
+        _,
         _,
         _,
         _,
@@ -93,6 +96,18 @@ def create_complete_app(context):
         assert 'deployed' in return_tuple[1]
 
 
+@when(u'we change the number of instances to {number}')
+def change_number_of_context_instances(context, number):
+    fake_service_config['instances'] = int(number)
+
+
+@when(u'we run setup_marathon_job for {number}s')
+def run_setup_marathon_job(context, number):
+    for _ in xrange(int(number)):
+        create_complete_app(context)
+        time.sleep(1)
+
+
 @then(u'we should see it in the list of apps')
 def see_it_in_list(context):
     assert fake_appid in marathon_tools.list_all_marathon_app_ids(context.marathon_client)
@@ -101,3 +116,8 @@ def see_it_in_list(context):
 @then(u'we can run get_app on it')
 def can_run_get_app(context):
     assert context.marathon_client.get_app(fake_appid)
+
+
+@then(u'we should see the number of instances become {number}')
+def assert_instances_equals(context, number):
+    assert context.marathon_client.get_app(fake_appid).instances == int(number)
--- a/paasta_itests/steps/setup_marathon_job_steps.py
+++ b/paasta_itests/steps/setup_marathon_job_steps.py
@@ -52,7 +52,7 @@ fake_service_config = {
             {'hostPath': u'/nail/srv', 'containerPath': '/nail/srv', 'mode': 'RO'},
             {'hostPath': u'/etc/boto_cfg', 'containerPath': '/etc/boto_cfg', 'mode': 'RO'}]
     },
-    'instances': 3,
+    'instances': 1,
     'mem': 300,
     'args': [],
     'backoff_factor': 2,
@@ -101,9 +101,9 @@ def change_number_of_context_instances(context, number):
     fake_service_config['instances'] = int(number)
 
 
-@when(u'we run setup_marathon_job for {number}s')
+@when(u'we run setup_marathon_job until the instance count is {number}')
 def run_setup_marathon_job(context, number):
-    for _ in xrange(int(number)):
+    while context.marathon_client.get_app(fake_appid).instances != int(number):
         create_complete_app(context)
         time.sleep(1)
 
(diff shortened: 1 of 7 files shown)


paasta_itests/steps/setup_marathon_job_steps.py
```diff
@@ -93,6 +96,18 @@ def create_complete_app(context):
...
 
+@when(u'we change the number of instances to {number}')
+def change_number_of_context_instances(context, number):
+    fake_service_config['instances'] = int(number)
+
+
+@when(u'we run setup_marathon_job for {number}s')
+def run_setup_marathon_job(context, number):
+    for _ in xrange(int(number)):
+        create_complete_app(context)
```
solarkennedy: This is weird to me. Is the amount of sleep time so important that it has to be a parameter? Can you instead use the "wait for x tasks" or something?

solarkennedy: This isn't obvious to me how many instances we are starting off with. can you make *that* the parameter? "we run setup_marathon_job on a service with an instance count of X"

paasta_itests/steps/setup_marathon_job_steps.py
```diff
@@ -51,14 +52,14 @@
             {'hostPath': u'/nail/srv', 'containerPath': '/nail/srv', 'mode': 'RO'},
             {'hostPath': u'/etc/boto_cfg', 'containerPath': '/etc/boto_cfg', 'mode': 'RO'}]
     },
-    'instances': 1,
+    'instances': 3,
```
solarkennedy: I'm surprised this doesn't screw up other tests. If possible let this stay at 1 and for your needs adjust it as necissary

paasta_tools/graceful_app_drain.py
```diff
@@ -72,14 +72,24 @@ def main():
...
+        )
         do_bounce(
             bounce_func=bounce_func,
             drain_method=drain_method,
             config=complete_config,
             new_app_running='',
             happy_new_tasks=[],
-            old_app_live_tasks=old_app_live_tasks,
+            old_app_live_happy_tasks=old_app_live_happy_tasks,
+            old_app_live_unhappy_tasks=old_app_live_unhappy_tasks,
```
solarkennedy: I'm surprised to see this, I thought this was in master? Are you based off master?

mjksmith: Yes, this was never changed in master since there are no itests for `graceful_app_drain`

mjksmith:
@solarkennedy I fixed the comments you made about itests

solarkennedy:
@EvanKrall  can you review this?

paasta_tools/setup_marathon_job.py
```diff
@@ -400,17 +408,43 @@ def log_deploy_error(errormsg, level='event'):
...
+            )
+            scaling_app_happy_tasks = list(task_dict['happy'])
+            scaling_app_unhappy_tasks = list(task_dict['unhappy'])
+            scaling_app_draining_tasks = list(task_dict['draining'])
+            tasks_to_move_draining = max(min(len(scaling_app_draining_tasks), num_tasks_to_scale), 0)
+            num_tasks_to_scale = num_tasks_to_scale - tasks_to_move_draining
+            tasks_to_move_unhappy = max(min(len(scaling_app_unhappy_tasks), num_tasks_to_scale), 0)
+            num_tasks_to_scale = num_tasks_to_scale - tasks_to_move_unhappy
+            tasks_to_move_happy = max(min(len(scaling_app_happy_tasks), num_tasks_to_scale), 0)
+            protected_draining_tasks.union(scaling_app_draining_tasks[:tasks_to_move_draining])
```
EvanKrall: I think this is a noop -- `foo.union()` returns a new set, but doesn't modify foo.

EvanKrall: make sure to write a test that would catch this

paasta_itests/steps/setup_marathon_job_steps.py
```diff
@@ -93,6 +96,29 @@ def create_complete_app(context):
...
+    fake_service_config['instances'] = int(number)
+
+
+@when(u'we run setup_marathon_job until the instance count is {number}')
+def run_setup_marathon_job(context, number):
+    app_instances = -1
+    while app_instances != int(number):
+        try:
+            app_instances = context.marathon_client.get_app(fake_appid).instances
+        except MarathonHttpError:
```
solarkennedy: This function seems kinda funny to me. We already have functions to wait for a certain number of tasks, can you use those instead?

paasta_itests/steps/setup_marathon_job_steps.py
```diff
@@ -93,6 +96,29 @@ def create_complete_app(context):
...
+def change_number_of_context_instances(context, number):
+    fake_service_config['instances'] = int(number)
+
+
+@when(u'we run setup_marathon_job until the instance count is {number}')
+def run_setup_marathon_job(context, number):
+    app_instances = -1
+    while app_instances != int(number):
+        try:
+            app_instances = context.marathon_client.get_app(fake_appid).instances
```
solarkennedy: I'm surprised that this is called "instances" here, I thought it was called "tasks"?

paasta_tools/setup_marathon_job.py
```diff
@@ -400,17 +408,43 @@ def log_deploy_error(errormsg, level='event'):
         bounce_health_params
     )
 
+    if new_app_running:
+        protected_draining_tasks = set()
+        if new_app.instances < config['instances']:
+            client.scale_app(app_id=new_app.id, instances=config['instances'], force=True)
+        elif new_app.instances > config['instances']:
+            num_tasks_to_scale = len(new_app.tasks) - config['instances']
```
EvanKrall: probably easiest to put your `max(..., 0)` here instead of duplicating it 3 times later

paasta_tools/setup_marathon_job.py
```diff
@@ -400,17 +408,43 @@ def log_deploy_error(errormsg, level='event'):
...
+            scaling_app_draining_tasks = list(task_dict['draining'])
+            tasks_to_move_draining = max(min(len(scaling_app_draining_tasks), num_tasks_to_scale), 0)
+            num_tasks_to_scale = num_tasks_to_scale - tasks_to_move_draining
+            tasks_to_move_unhappy = max(min(len(scaling_app_unhappy_tasks), num_tasks_to_scale), 0)
+            num_tasks_to_scale = num_tasks_to_scale - tasks_to_move_unhappy
+            tasks_to_move_happy = max(min(len(scaling_app_happy_tasks), num_tasks_to_scale), 0)
+            protected_draining_tasks.union(scaling_app_draining_tasks[:tasks_to_move_draining])
+            old_app_draining_tasks[new_app.id] = set(scaling_app_draining_tasks[:tasks_to_move_draining])
+            old_app_live_unhappy_tasks[new_app.id] = set(scaling_app_unhappy_tasks[:tasks_to_move_unhappy])
+            old_app_live_happy_tasks[new_app.id] = set(scaling_app_happy_tasks[:tasks_to_move_happy])
```
EvanKrall: I'd move this line and the line above it up to just after you define `tasks_to_move_unhappy` and `tasks_to_move_happy`. This would make the structure of this section a bit more apparent.

Some vertical whitespace would help, too.

paasta_itests/steps/setup_marathon_job_steps.py
mjksmith: in marathon it's `app.instances` so I kept with that naming

mjksmith:
@solarkennedy I fixed the itests

paasta_tools/marathon_tools.py
```diff
@@ -879,6 +879,8 @@ def wait_for_app_to_launch_tasks(client, app_id, expected_tasks, exact_matches_o
 
 def create_complete_config(service, instance, marathon_config, soa_dir=DEFAULT_SOA_DIR):
     """Generates a complete dictionary to be POST'ed to create an app on Marathon"""
+    CONFIG_HASH_BLACKLIST = set(['instances'])
```
EvanKrall: Probably good to write a comment about what this is for / what kinds of things are eligible to be in it, e.g. "A set of config attributes that don't get included in the hash of the config. These should be things that Paasta+Marathon knows how to change without requiring a bounce."

PR #256, opened 2016-02-12
Ignore command healthchecks when judging how happy tasks are

solarkennedy:
Primary: @EvanKrall 

I am not super stoked about this one, but I can confirm that the command healthcheck results are just not available to marathon to use. Maybe because they happen on the mesos slave? (but we still see them in the marathon logs? idk.

Closes internal ticket PAASTA-3110.

### DIFF
--- a/paasta_itests/steps/bounces_steps.py
+++ b/paasta_itests/steps/bounces_steps.py
@@ -96,7 +96,7 @@ def when_there_are_num_which_tasks(context, num, which, state):
                 return
         time.sleep(0.5)
     raise Exception("timed out waiting for %d %s tasks on %s; there are %d" %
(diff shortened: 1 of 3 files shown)


mfield:
This is kind of scary. I'd really rather the bounce code didn't just assume my daemon tasks are starting successfully. Do you think it's worth putting a little web server in my daemon task's docker image so I can use HTTP healthchecks to work around this?

EvanKrall:
I believe the underlying issue is https://issues.apache.org/jira/browse/MESOS-3738. I think the correct fix here is to apply the patches from that ticket to our Mesos builds, until we get up to 0.26.0

solarkennedy:
Yea I'm going to discard this as it is just terrible and hopefully I can fix the core bug.

PR #259, opened 2016-02-16
added string interpolation to deploy groups

mjksmith:
@EvanKrall review -- the config interpolation we talked about

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -73,6 +73,18 @@ class InstanceConfig(dict):
         self.cluster = cluster
         self.instance = instance
         self.service = service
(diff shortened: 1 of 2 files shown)


mjksmith:
@asottile I think this will fix (or at least reduce the severity of) the internal ticket you made - by using interpolation in your deploy groups you could have one stage file symlinked to each of your stage clusters with the deploy groups "{cluster}" which would product deploy groups norcal-stageb, norcal-stagef, .... etc.

Can you let me know if this is a good solution for you?

EvanKrall:
Looks good to me. If desired, we can add more per-cluster facts later via system configs.

PR #265, opened 2016-02-24
made bounce kill apps only when they have instances == 0

mjksmith:
@solarkennedy @EvanKrall right now scaling tasks down kills them since `do_bounce` doesn't know about the other healthy tasks we want to keep up. This fixes that.

This blocks testing scaling & unpinning paasta-tools so I'd like this looked over ASAP please.

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -250,15 +250,10 @@ def do_bounce(
             log_bounce_action(line='%s bounce killing drained task %s' % (bounce_method, task.id))
             marathon_tools.kill_task(client=client, app_id=task.app_id, task_id=task.id, scale=True)
 
-    apps_to_kill = []
-    for app in old_app_live_happy_tasks.keys():
-        live_happy_tasks = old_app_live_happy_tasks[app]
-        live_unhappy_tasks = old_app_live_unhappy_tasks[app]
-        draining_tasks = old_app_draining_tasks[app]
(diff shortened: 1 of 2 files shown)


EvanKrall:
looking

paasta_tools/setup_marathon_job.py
```diff
@@ -250,15 +250,10 @@ def log_bounce_action(line, level='debug'):
...
-        live_happy_tasks = old_app_live_happy_tasks[app]
-        live_unhappy_tasks = old_app_live_unhappy_tasks[app]
-        draining_tasks = old_app_draining_tasks[app]
-
-        if 0 == len((live_happy_tasks | live_unhappy_tasks | draining_tasks) - killed_tasks):
-            apps_to_kill.append(app)
+    apps = set(old_app_live_happy_tasks.keys()) | set(old_app_live_unhappy_tasks.keys()) \
+        | set(old_app_draining_tasks.keys())
 
+    apps_to_kill = [app for app in apps if client.get_app(app).instances == 0]
```
EvanKrall: This is going to cause setup_marathon_job to take longer / put more load on the marathon server.

Don't the other healthy tasks that we want to keep alive end up in live_happy_tasks?

EvanKrall: oh jk, live_happy_tasks only includes "old" tasks.

You could skip killing the app if `app == marathon_jobid`

mjksmith:
I implemented the skip for when `app == marathon_jobid`

Thanks for getting to this so quickly

PR #266, opened 2016-02-25
Pass --env_vars_prefix to marathon and remove local_run workaround

### LINKED ISSUE #152: PORT exported by paasta makes it difficult to do dockerfiles according to best practices
From https://docs.docker.com/v1.8/articles/dockerfile_best-practices/#cmd you should use the json syntax
From https://github.com/Yelp/dumb-init it should look like `CMD ["dumb-init", "...", "..."]`

Paasta runs `--env=PORT=This_service_is_listening_on_the_PORT_variable__You_must_use_8888__see_y/paasta_deploy`

Even if you do it correctly with a dockerfile that looks like this:

```dockerfile
...
WORKDIR /code

ENV     PATH=/code/virtualenv_run/bin:$PATH
ENV     BASEPATH=/code
ENV     PORT=8888

CMD     ["dumb-init", "python", "./serviceinit.d/www_pages start-no-daemon"]
```

The `--env` line clobbers my Dockerfile's *correct* setting.

A workaround (which kinda sucks): 
`["dumb-init" "sh" "-c" "export PORT=8888; exec python serviceinit.d/www_pages start-no-daemon"]`

Suggestion: remove `--env=PORT=...` for two reasons:
1. It's yelpy
2. It breaks best practices

nhandler:
Pass the `--env_vars_prefix` argument to Marathon. By using this argument, we can avoid having it hog the `PORT` environment variable.

I'm removing the warnings in the docs. The contract still notes the requirement to use port 8888. I have also removed the hack in `local_run` to explicitly set the `PORT` environment variable to match production.

Fixes #152

### DIFF
--- a/docs/source/workflow.rst
+++ b/docs/source/workflow.rst
@@ -86,8 +86,6 @@ configuration for the running image:
   ``PAASTA_`` environment variables will also be injected, see the `related docs <yelpsoa_configs.html#marathon-clustername-yaml>`_
   for more information.
 
-  * **WARNING**: A PORT variable is provided to the docker image, but it represents the EXTERNAL port, not the internal one. The internal service MUST listen on 8888, so this PORT variable confuses some service stacks that are listening for this variable. Such services MUST overwrite this environment variable to function. (``PORT=8888 ./uwisgi.py```)
-
 * ``--publish``: Mesos picks a random port on the host that maps to and exposes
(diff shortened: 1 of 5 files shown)


solarkennedy:
Ship. I'm happy to see this go.

Rob-Johnson:
shipit

PR #268, opened 2016-02-25
Update emergency-start semantics for chronos jobs

keshavdv:
Update chronos_serviceinit so that it has the following semantics:

emergency-start with previously stopped jobs: Trigger manual run immediately
emergency-start with 'disabled: False' in config: Trigger manual run immediately
emergency-start with 'disabled: True' in config: noop
emergency-stop: Stop in flight job and mark as disabled in Chronos


solarkennedy:
I would like to look at this carefully. Things are a bit different now that we don't bounce in the same way with chronos that we used to.

keshavdv:
This definitely turned into a lot more than I expected so a close look at the changes would be great

keshavdv:
This ends up using the undocumented csv endpoint to get job status so that the itests actually assert something. I don't think this is a great way to do it, but until there's a real API in Chronos to get this info I'm not sure what else we could use.

paasta_tools/chronos_serviceinit.py
```diff
@@ -290,11 +313,26 @@ def status_chronos_jobs(jobs, job_config, verbose):
...
+        start_chronos_job(
+            service,
+            instance,
+            job_id,
+            client,
+            cluster,
+            job_config,
+            complete_job_config,
+            emergency=True,
+        )
```
mjksmith: Can you kwargs this? As well as `stop_chronos_job` and `restart_chronos_job` while you're at it.

paasta_tools/chronos_tools.py
```diff
@@ -597,6 +596,27 @@ def get_key(job):
...
+    node,example_job,success,queued
+    node,parent_job,success,idle
+    node,child_job,success,idle
+    link,parent_job,child_job
+
+    :param client: Chronos client object
+    :param service: service name
+    :param instance: instance name
+    """
+    resp = client._call("/scheduler/graph/csv", "GET")
```
solarkennedy: I'm glad we have this, it provides nice data and I though we would need it eventually.

I'm a bit concerned about speed. In your testing doesn't this take quite a number of seconds to GET?

keshavdv: The endpoint returns immediately, but the actual state it returns occasionally takes a second to be updated (killing a job that was "queued" is really the only the case where this happens) which is why the tests call this repeatedly.

tests/test_chronos_serviceinit.py
```diff
@@ -28,21 +29,33 @@ def test_start_chronos_job():
...
     with contextlib.nested(
         mock.patch('paasta_tools.chronos_serviceinit.chronos_tools.chronos.ChronosClient', autospec=True),
+        mock.patch('sys.stdout', new_callable=StringIO),
     ) as (
         mock_client,
+        mock_stdout,
     ):
         chronos_serviceinit.start_chronos_job(service,
                                               instance,
                                               job_id,
```
solarkennedy: Can you go ahead and kwargs the rest of these

tests/test_chronos_serviceinit.py
```diff
@@ -51,21 +64,33 @@ def test_start_chronos_job_does_not_run_disabled_job():
...
                                               mock_client,
                                               cluster,
-                                              job_config)
+                                              job_soa_config,
+                                              job_config,
+                                              emergency=True)
```
solarkennedy: more kwargs

paasta_tools/chronos_serviceinit.py
```diff
@@ -30,23 +30,36 @@
...
+    running the job now regardless of its 'schedule'. The job's "schedule" is unmodified. If a job is disabled,
+    this function does not do anything.
+    """
     name = PaastaColors.cyan(job_id)
+    # The job should be run immediately as long as the job is not disabled via
+    # the 'disabled' key in soa-configs. Since we clobber the 'disabled' key in
+    # complete_job_config (the dict that gets passed to Chronos), we need to look
+    # at the service's soa-config directly to check if it is disabled.
+    should_run_now = not job_config.get_disabled()
     log_reason = PaastaColors.red("EmergencyStart") if emergency else "Brutal bounce"
```
solarkennedy: I don't understand why we do this here. 

In fact this all seems really over complicated. Why can't this just be:


```python
if job_config.get_disabled():
    print "You can't emergency start a disabled job. Run `paasta start` first."
else:
    client.run(job_id)
```

All this other stuff is unnecessary (now) imo.

paasta_tools/chronos_tools.py
```diff
@@ -478,10 +479,8 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
     complete_config['description'] = get_config_hash(complete_config)
 
     # If the job was previously stopped, we should stop the new job as well
-    # NOTE this clobbers the 'disabled' param specified in the config file!
-    if desired_state == 'start':
-        complete_config['disabled'] = False
-    elif desired_state == 'stop':
+    # NOTE: this clobbers the 'disabled' param specified in the config file!
+    if desired_state == 'stop':
```
solarkennedy: I don't think we have to do this anymore.

What if we had the `get_disabled` for chronos jobs be just like `get_instances` in marathon? (where get_instances returns 0 when something is stopped)

paasta_tools/chronos_tools.py
Rob-Johnson: can you move this to the client so we can just do ``client.scheduler_graph`` ? we have commit access, so can merge you PR quickly

paasta_itests/paasta_serviceinit.feature
```diff
@@ -54,18 +56,29 @@ Feature: paasta_serviceinit
       And we wait for the chronos job stored as "myjob" to appear in the job list
       And we paasta_serviceinit emergency-start the service_instance "testservice.testinstance"
      Then the job stored as "myjob" is enabled in chronos
-      And the job has running tasks
+      And the job stored as "myjob" has running tasks
 
-  Scenario: paasta_serviceinit can run emergency-start on a disabled chronos job
+  Scenario: paasta_serviceinit cannot run emergency-start on a disabled chronos job
```
Rob-Johnson: this isn't really correct - we can run emergency-start on a disabled chronos job, it's just noop

paasta_tools/chronos_tools.py
keshavdv: If `get_disabled` looks at desired_state instead of being clobbered after the fact, then we can no longer differentiate a job a stopped job (paasta stop) with a 'disabled: True' key vs a stopped job without that key. This makes it impossible for emergency-stop to noop only for disabled jobs but not stopped ones.

keshavdv:
Updated to match #293. Calling emergency-start on a chronos job will now always trigger a manual run regardless of desired_state.

paasta_tools/chronos_tools.py
keshavdv: Done at  https://github.com/asher/chronos-python/pull/23.

paasta_tools/chronos_serviceinit.py
```diff
@@ -30,24 +30,24 @@
...
+    running the job now regardless of its 'schedule'. The job's "schedule" is unmodified.
+    """
     name = PaastaColors.cyan(job_id)
     log_reason = PaastaColors.red("EmergencyStart") if emergency else "Brutal bounce"
-    log_immediate_run = " and running it immediately" if not job_config["disabled"] else ""
+
     _log(
         service=service,
-        line="%s: Sending job %s to Chronos%s" % (log_reason, name, log_immediate_run),
+        line="%s: Sending job %s to Chronos" % (log_reason, name),
```
solarkennedy: This English sounds strange.

How about "%s: Kicking off manual run of %s in Chronos"

paasta_tools/chronos_tools.py
solarkennedy: Great job.

Rob-Johnson:
ah man, sorry, I reverted #293 yesterday, so I guess this branch needs rebasing again.

I don't want to block you anymore - #283 is still an issue with your change (which #293 was trying to fix), so lgtm as soon as possible and I can base any work off that.

nhandler:
In order to finally get this change merged, I have gone ahead and rebased for @keshavdv. However, since I am unable to push to branches on his forked repository, I am unable to update this Pull Request. I have created #370 instead. To avoid confusion, I am going to close this pull request.

PR #272, opened 2016-02-25
Add extra_constraints and pool to schemas for paasta validate

nhandler:
Currently, `paasta validate` complains about marathon/chronos configs that contain these keys. This change fixes that so they show up as valid.

I am also wondering if it is worth adding some more logic around the constraints to make `paasta validate` aware of invalid constraints.


solarkennedy:
Can you maybe rebase? I don't understand how these test failures are related.

nhandler:
That did it. Tests now pass.

solarkennedy:
ship

PR #273, opened 2016-02-27
Use returncode instead of sys.exit in the paasta cli

solarkennedy:
We use sys.exit in many many places. It seems easier if we just return a number and let that go back to the cli entrypoint and let it sys.exit in a central place.

It simplifies a lot of our testing and stuff. We still sys.exit early in other places when we panic, but that is ok I think.


nhandler:
Ship It!

PR #275, opened 2016-02-27
Allow marathon instance counts to be read from zookeeper

mjksmith:
Also provides an endpoint for autoscaling to update instance counts in zookeeper


paasta_tools/marathon_serviceinit.py
```diff
@@ -329,13 +329,14 @@ def perform_command(command, service, instance, cluster, verbose, soa_dir, app_i
...
-    if not app_id:
-        try:
-            app_id = marathon_tools.create_complete_config(service, instance, marathon_config, soa_dir=soa_dir)['id']
-        except NoDockerImageError:
-            job_id = compose_job_id(service, instance)
-            print "Docker image for %s not in deployments.json. Exiting. Has Jenkins deployed it?" % job_id
-            return 1
+    try:
+        complete_config = marathon_tools.create_complete_config(service, instance, marathon_config, soa_dir=soa_dir)
+        if not app_id:
```
solarkennedy: app_id isn't anything at this point, right?

mjksmith: `app_id` can be passed in as an arg but defaults to None

paasta_tools/marathon_tools.py
```diff
@@ -194,8 +207,22 @@ def get_instances(self):
...
+            instances = self.get_min_instances()
+            max_instances = self.get_max_instances()
+            if max_instances is not None:
+                try:
+                    zk_instances = get_instances_from_zookeeper(
+                        service=self.service,
+                        instance=self.instance,
+                    )
+                except NoNodeError:
+                    pass
```
solarkennedy: Can you add a comment to explain what this signifies?

paasta_tools/marathon_tools.py
```diff
@@ -185,8 +188,18 @@ def __repr__(self):
     def copy(self):
         return self.__class__(self.service, self.instance, dict(self.config_dict), dict(self.branch_dict))
 
+    def close_zookeeper_client(self):
+        self.zookeeper_client.stop()
+        self.zookeeper_client.close()
+
+    def get_min_instances(self):
+        return self.config_dict.get('instances', 1)
```
solarkennedy: I'm concerned about potential confusion here. What if this was literally `min_instances` in the config, to make it very obvious which things are autoscaling and which are static?

paasta_tools/marathon_tools.py
```diff
@@ -185,8 +187,14 @@ def __repr__(self):
     def copy(self):
         return self.__class__(self.service, self.instance, dict(self.config_dict), dict(self.branch_dict))
 
+    def get_min_instances(self):
```
mjksmith: @solarkennedy said: I'm concerned about potential confusion here. What if this was literally `min_instances` in the config, to make it very obvious which things are autoscaling and which are static?

but it got buried by a commit.

paasta_tools/setup_marathon_job.py
```diff
@@ -512,7 +512,7 @@ def setup_service(service, instance, client, marathon_config,
 
     log.info("Setting up instance %s for service %s", instance, service)
     try:
-        complete_config = marathon_tools.create_complete_config(service, instance, marathon_config, soa_dir)
+        complete_config = marathon_tools.create_complete_config(service, instance, soa_dir)
```
solarkennedy: Wow, way for us to fail at dependency injection amirite?

Someone who knows more than me would probably say that we should go the other direction and *actually* pass in the config object instead of having it re-look it up. (and then junk the service/instance inputs because they are in the ojbect now thanks to you)

paasta_tools/marathon_tools.py
```diff
@@ -888,7 +910,7 @@ def create_complete_config(service, instance, marathon_config, soa_dir=DEFAULT_S
     instance_config = load_marathon_service_config(
         service=service,
         instance=instance,
-        cluster=load_system_paasta_config().get_cluster(),
+        cluster=system_paasta_config.get_cluster(),
```
solarkennedy: Oh we are so silly.

paasta_tools/marathon_tools.py
```diff
@@ -194,8 +202,22 @@ def get_instances(self):
...
+            max_instances = self.get_max_instances()
+            if max_instances is not None:
+                try:
+                    zk_instances = get_instances_from_zookeeper(
+                        service=self.service,
+                        instance=self.instance,
+                    )
+                except NoNodeError:  # zookeeper doesn't have instance data for this app
+                    pass
+                else:
```
solarkennedy: can you confirm this indentation? Looks too indented

paasta_tools/marathon_tools.py
```diff
@@ -993,3 +1015,47 @@ def kill_task(client, app_id, task_id, scale):
...
+    with ZookeeperPool() as zookeeper_client:
+        instances = min(
+            marathon_config.get_max_instances(),
+            max(marathon_config.get_min_instances(), marathon_config.get_instances() + delta))
+        zookeeper_client.ensure_path(zookeeper_path)
+        zookeeper_client.set(zookeeper_path, instances)
+        return instances
+
+
+class ZookeeperPool(object):
```
solarkennedy: This... could use a docstring

paasta_tools/marathon_tools.py
```diff
@@ -993,3 +1015,47 @@ def kill_task(client, app_id, task_id, scale):
...
+    counter = 0
+    zk = None
+
+    @classmethod
+    def __enter__(cls):
+        if cls.zk is None:
+            cls.zk = KazooClient(hosts=load_system_paasta_config().get_zk_hosts(), read_only=True)
```
solarkennedy: I'm concerned here because this returns literally the zk hosts, and not the namespace we want.

I can't tell from the docs/source, but we should be able to pass in the correct "chroot" with our hosts  string.

https://kazoo.readthedocs.org/en/latest/_modules/kazoo/client.html#KazooClient.set_hosts

This is important long term as we don't want to make the assumption that we have the zk server all to ourselves and we pollute the root namespace.

paasta_tools/marathon_tools.py
mjksmith: @solarkennedy yes the `else` is part of the `try...except` block

paasta_tools/marathon_tools.py
```diff
@@ -993,3 +1015,47 @@ def kill_task(client, app_id, task_id, scale):
...
+def update_instances_for_marathon_service(service, instance, delta):
+    zookeeper_path = '/autoscaling/%s/%s/instances' % (service, instance)
+    marathon_config = load_marathon_service_config(
+        service=service,
+        instance=instance,
+        cluster=load_system_paasta_config().get_cluster(),
+        load_deployments=False,
+    )
+    with ZookeeperPool() as zookeeper_client:
+        instances = min(
```
solarkennedy: You might be jumping the gun a little bit on this function, but I understand why you wrote it.

It might be simpler if you had the function "just" take a number, and then make another function that does delta math. Don't go too nuts on this though, still early.

paasta_tools/marathon_tools.py
mjksmith: @solarkennedy AFAICT you just add the chroot to the end of the host list

For example `zk://10.40.5.5:22181,10.40.5.6:22181,10.40.1.17:22181/mesos-mesosstage`

paasta_tools/marathon_tools.py
mjksmith: @solarkennedy yeah this was really just for a quick demo of programatic control.

paasta_tools/marathon_tools.py
mjksmith: @solarkennedy looking further in `/etc/paasta/cluster.json` we specify a chroot as `mesos-CLUSTERNAME` (you can see it in the example I gave, it was a readout from `get_zk_hosts()` on `mesos3@mesosstage`).

We should document this but at Yelp we won't pollute the root namespace.

solarkennedy: Ah gotcha. I incorrectly assumed that get_zk_hosts returned *just* the hosts for the zk connection string as opposed to the whole thing including the chroot.

Carry on.

mjksmith:
This is ready for review now

paasta_itests/steps/setup_steps.py
```diff
@@ -183,7 +183,7 @@ def write_soa_dir_marathon_job(context, job_id):
         f.write(yaml.dump({
             "%s" % instance: {
                 'cpus': 0.1,
-                'ram': 100,
+                'mem': 100,
```
solarkennedy: Heh, lolz

paasta_tools/bounce_lib.py
```diff
@@ -26,12 +26,12 @@
 from kazoo.client import KazooClient
 from kazoo.exceptions import LockTimeout
 from marathon.models import MarathonApp
-from utils import compose_job_id
-from utils import load_system_paasta_config
```
solarkennedy: Wow. How did this work before?

solarkennedy:
Ship. Please get one other shipit and then pin+ship. (though this change should be pretty safe as we are not using max_instances yet)

paasta_itests/steps/setup_marathon_job_steps.py
```diff
@@ -58,14 +52,53 @@ def create_complete_app(context):
             pass
 
 
+@when(u'we set up an app to use zookeeper scaling with {number:d} max instances')
+def setup_zookeeper(context, number):
+    client = KazooClient(hosts='%s/mesos-testcluster' % get_service_connection_string('zookeeper'), read_only=True)
+    client.start()
+    try:
+        client.delete('/autoscaling', recursive=True)
```
Rob-Johnson: I take it this is here to cleanup between runs? If so, can you move this to ``environment.py`` where we have the other cleanup code (like removing runnning marathon apps, chronos jobs)

paasta_tools/marathon_tools.py
```diff
@@ -993,3 +1015,45 @@ def kill_task(client, app_id, task_id, scale):
             return []
         else:
             raise
+
+
+def get_instances_from_zookeeper(service, instance):
+    with ZookeeperPool() as zookeeper_client:
+        (instances, _) = zookeeper_client.get('/autoscaling/%s/%s/instances' % (service, instance))
```
Rob-Johnson: instances, _ = ...

ditch the surrounding brackets.

paasta_tools/marathon_tools.py
```diff
@@ -993,3 +1015,45 @@ def kill_task(client, app_id, task_id, scale):
...
+        if instances is None or instances == 'None':
+            raise NoNodeError
```
Rob-Johnson: it seems odd to do this - if kazoo has decided to not raise exceptions at this point, why're we modifying that? how about we return None here, and change your try/catch to check for None?

paasta_tools/marathon_tools.py
```diff
@@ -185,8 +187,14 @@ def __repr__(self):
...
 
+    def get_min_instances(self):
+        return self.config_dict.get('instances', 1)
+
+    def get_max_instances(self):
+        return self.config_dict.get('max_instances', None)
+
     def get_instances(self):
-        """Get the number of instances specified in the service's marathon configuration.
+        """Get the number of instances specified in zookeeper or the service's marathon configuration.
```
Rob-Johnson: can you be clearer here that it takes the higher value if both are found?

mjksmith:
@Rob-Johnson done. I found out that that extra exception handling for Zookeeper was unnecessary.

Rob-Johnson:
shipit!

PR #276, opened 2016-02-29
Updated /state endpoint for mesos 0.25.x upgrade

### LINKED ISSUE #232: Rename /state.json endpoint in conjunction with mesos 0.24.x to 0.25.x upgrade
In the mesos notes on upgrading from 0.24.x to 0.25.x, they note that the `/state.json` endpoint is becoming `/state`. We use this endpoint in at least one place: https://github.com/Yelp/paasta/blob/67793d2fe99a7f503add185fb8ad92b1a039e436/paasta_tools/mesos_tools.py#L276

This will need to be fixed when we do the mesos upgrade.

Note: A few other endpoints are being renamed, but I was not able to find references to them from a quick grep through our code.

ameya-pandilwar:
Fixes #232

### DIFF
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -276,6 +276,9 @@ def get_local_slave_state():
     stats_uri = 'http://%s:%s/state' % (hostname, MESOS_SLAVE_PORT)
     try:
         response = requests.get(stats_uri, timeout=10)
+        if response.status_code == 404:
+            fallback_stats_uri = 'http://%s:%s/state.json' % (hostname, MESOS_SLAVE_PORT)
+            response = requests.get(fallback_stats_uri, timeout=10)
     except requests.ConnectionError as e:
         raise MesosSlaveConnectionError(
             'Could not connect to the mesos slave to see which services are running\n'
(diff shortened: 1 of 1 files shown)


solarkennedy:
Whoa! Thanks!
However this does cause this test to fail:
https://travis-ci.org/Yelp/paasta/jobs/112490863#L4943

keshavdv:
It would be great if this could fall back to /state.json if /state returns a 404 so that the code remains backward compatible with older versions of Mesos.

Rob-Johnson:
thanks for this @ameya-pandilwar 

the /state endpoint wasn't introduced until mesos 0.25.x, and we run 0.23.x in our tests (and in production) - so won't be able to use it just yet.

It'd be great if you could check for the version and change the endpoint used accordingly? or, as @keshavdv suggested, just fallback to it if /state fails?

ameya-pandilwar:
Fixed the broken test, @solarkennedy !

Introduced a fallback to the backward compatible `/state.json` as @keshavdv suggested. FYI, @Rob-Johnson

solarkennedy:
Super. Thanks!

PR #280, opened 2016-03-01
add a note about the parents field in yelpsoa-configs

Rob-Johnson:
there will be more docs to follow up with, but this is a start


nhandler:
Just noting this here so it doesn't get forgotten, but you will also need to update the schemas in https://github.com/Yelp/paasta/tree/master/paasta_tools/cli/schemas

Rob-Johnson:
@nhandler: we've already done that https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/schemas/chronos_schema.json#L89-L100

nhandler:
Silly me. Forgot about that one :)

Issue #64, opened 2015-11-19
create alternative backend for logging


Rob-Johnson:
PaaSTA currently makes lots of assumptions about logging infrastructure; namely that it is Yelp's scribe infrastructure. The side effect of this is that logging in PaaSTA doesn't really work outside of Yelp. We should support alternative logging infrastructure.

PR #281, opened 2016-03-01
Pluggable log systems

EvanKrall:
This branch makes log writers and log readers pluggable. This is the first part of the solution for #64.

It does not currently implement any new log writers or readers besides Scribe / Scribereader (and some null/testing ones).


general_itests/fake_etc_paasta/volumes.json
```diff
@@ -0,0 +1,9 @@
+{
+    "volumes": [
+        {
+          "hostPath": "/nail/etc/services",
```
solarkennedy: For itests can you use something less yelpy to make it more obvious that we are not accidentally depending on real resources?

paasta_itests/fake_etc_paasta/volumes.json
```diff
@@ -0,0 +1 @@
+{"volumes": [{"hostPath": "/nail/etc/beep", "containerPath": "/nail/etc/beep", "mode": "RO"}, {"hostPath": "/nail/etc/bop", "containerPath": "/nail/etc/bop", "mode": "RO"}, {"hostPath": "/nail/etc/boop", "containerPath": "/nail/etc/boop", "mode": "RO"}]}
```
solarkennedy: Same here, can you not use 'nail/etc'?

paasta_tools/cli/cmds/logs.py
```diff
@@ -356,171 +293,268 @@ def prettify_log_line(line, requested_levels):
...
 
-    NOTE: We try pretty hard to supress KeyboardInterrupts to prevent big
-    useless stack traces, but it turns out to be non-trivial and we fail ~10%
-    of the time. We decided we could live with it and we're shipping this to
-    see how it fares in real world testing.
+def register_log_reader(name):
+    """Returns a decorator that registers that bounce function at a given name
+    so get_log_reader_classes can find it."""
+    def outer(bounce_func):
+        _log_reader_classes[name] = bounce_func
```
solarkennedy: This confuses me. Is this bounce-specific? Could this decorator not be applied to any function? (I don't understand what the bounce has to do with it)

EvanKrall: Oops. I copied this registration code from bounce_lib, but didn't fix the docstring.

paasta_itests/fake_etc_paasta/volumes.json
EvanKrall: pretty sure beep bop boop don't exist in our /nail/etc

paasta_tools/utils.py
```diff
@@ -465,9 +465,46 @@ class NoSuchLogLevel(Exception):
     pass
 
 
+# The active log writer.
+_log_writer = None
+# The map of name -> LogWriter subclasses, used by configure_log.
+_log_writer_classes = {}
+
+
+def register_log_writer(name):
```
solarkennedy: This is very similar to register_log_reader in paasta-cli logs, what do you think about a catchall logging.py to hold all the logging related stuff?

tox.ini
```diff
@@ -83,6 +83,8 @@ commands =
     python -m behave {posargs}
 
 [testenv:general_itests]
+setenv =
+    PAASTA_SYSTEM_CONFIG_DIR = {toxinidir}/general_itests/fake_etc_paasta
```
solarkennedy: I guess you don't need this in the paasta_itests too?

yelp_package/dockerfiles/itest/fake_sensu/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM ubuntu:trusty
+RUN apt-get update && apt-get install -yq socat
+CMD socat tcp-listen:3030,fork open:/dev/null,ignoreeof
```
solarkennedy: While I understand why you did this, it does add yet another docker image to build and orchistrate.

What if the sensu_host could be None and then we could just skip this under test?
And it would allow people to deploy paasta incrementally, without sensu.

paasta_tools/utils.py
```diff
@@ -659,6 +715,40 @@ def get_fsm_deploy_pipeline(self):
...
 
+    def get_log_writer(self):
+        """Get the log_writer configuration out of global paasta config
+
+        :returns: The log_writer dictionary.
+        """
+        try:
+            return self['log_writer']
+        except KeyError:
+            raise PaastaNotConfiguredError('Could not find log_writer in configuration directory: %s' % self.directory)
```
solarkennedy: I remember you mentioning this before, but are you really tied to it?

Seems totally legit to me that this would default to say the "stderr" writer, and then you won't have to patch all the things under test?

paasta_tools/utils.py
```diff
@@ -659,6 +715,40 @@ def get_fsm_deploy_pipeline(self):
...
+
+    def get_log_reader(self):
+        """Get the log_reader configuration out of global paasta config
+
+        :returns: the log_reader dictionary.
+        """
+        try:
+            return self['log_reader']
+        except KeyError:
+            raise PaastaNotConfiguredError('Could not find log_reader in configuration directory: %s' % self.directory)
```
solarkennedy: This one I understand, I can't think of a possible default for this

paasta_tools/paasta_serviceinit.py
```diff
@@ -64,6 +65,8 @@ def main():
     else:
         log.setLevel(logging.WARNING)
 
+    configure_log()
```
solarkennedy: you mentioned now that this happens automatically, if so can we consistently *not* call this in main to tidy up the code a bit?

tox.ini
EvanKrall: in paasta_itests, the system config is actually at /etc/paasta, the default.

paasta_tools/utils.py
EvanKrall: If you don't patch things properly, the tests would read your dev box's /etc/paasta, which is undesirable.

You'd at least need to patch `load_system_paasta_config` or `os.environ['PAASTA_SYSTEM_CONFIG_DIR']` to avoid that, and if you're going to patch out something, you might as well be explicit about its value in the test.

EvanKrall: Also for actual deployments, I think I'd like log_writer to be required. There's no sane default: if we default to `null` or a hypothetical `stderr` writer, then `paasta logs` would not work, which is the whole point of this. There's a separate subsystem for writing to stderr: python's logging.

If someone wants to deploy paasta without paasta logs working, they can set up the `null` log_writer.

yelp_package/dockerfiles/itest/fake_sensu/Dockerfile
EvanKrall: Sounds good, I'll work on this.

EvanKrall:
This is ready for further code review.

EvanKrall:
I should probably write docs for the system configuration options.

paasta_itests/steps/setup_chronos_job_steps.py
```diff
@@ -30,12 +30,21 @@ def job_exists(context, service, instance):
...
-@when(u'we run setup_chronos_job for service_instance "{service_instance}"')
-def run_setup_chronos_job(context, service_instance):
+@when(u'we run setup_chronos_job for service_instance "{service_instance}", expecting exit code {expected_exit_code}')
+def run_setup_chronos_job(context, service_instance, expected_exit_code):
     cmd = "../paasta_tools/setup_chronos_job.py %s -d %s" % (service_instance, context.soa_dir)
     exit_code, output = _run(cmd)
     context.exit_code, context.output = exit_code, output
 
+    try:
+        assert context.exit_code == int(expected_exit_code), \
```
Rob-Johnson: you can tell behave the type rather than having to cast. something like {expected_exit_code:d}

Rob-Johnson: I'm not convinced we should be doing assertions in a '@when' - they should be in a '@then' 

In other places we have the execution of a script as a '@then', because we don't need to do anything after. See [paasta metastatus](https://github.com/Yelp/paasta/blob/3a7a5043b537c2f24c954d9d4b2dd2e061a97fb7/paasta_itests/steps/paasta_metastatus_steps.py#L63-L80) for example. But this isn't the case for this - we often do setup_chronos_job then check for some other thing in a '@then'.

Rob-Johnson:
this looks pretty good to me

solarkennedy:
rebase and ship

PR #282, opened 2016-03-01
Fix an output bug in paasta check

mjksmith:
Right now if there are chronos jobs that aren't listed in `deploy.yaml`, `paasta check` prints out a list of marathon instances instead.

Example:
```
✗ There are some instance(s) you have asked to run in chronos that
  do not have a corresponding entry in deploy.yaml:

  You should probably add entries to deploy.yaml for them so they
  are deployed to those clusters.
```


nhandler:
Ship It!

PR #285, opened 2016-03-02
Set MARATHON_PORT env variable in local-run

nhandler:
We now use Marathon's `--env_vars_prefix` option. This change makes sure that services run with `local-run` have `MARATHON_PORT` set similarly to how it would be set once deployed.


solarkennedy:
ship

PR #288, opened 2016-03-03
Set real disabled value

### LINKED ISSUE #283: setting 'disable: true' in a chronos job is a noop unless you also 'paasta stop' the job
The offending code is here:

https://github.com/Yelp/paasta/blob/master/paasta_tools/chronos_tools.py#L480-L485

```  
# NOTE this clobbers the 'disabled' param specified in the config file!
if desired_state == 'start':
    complete_config['disabled'] = False
elif desired_state == 'stop':
    complete_config['disabled'] = True
```

This means that setting ``disabled`` to ``True`` is noop, unless you ``paasta stop`` your chronos job first.

Given that we **only** actually update the job when the config hash is different, and the config hash is computed *before* the code above clobbers the disabled value, you *have* to set the ``disabled: True`` **after** you paasta stop to get the desired effect.

Rob-Johnson:
Closes #283 

This change fixes the confusing/buggy behaviour in disabling a chronos job

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -471,21 +471,54 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
     )
 
     complete_config['name'] = compose_job_id(service, job_name)
+
+    # resolve conflicts between the 'desired_state' and soa_configs disabled
+    # flag.
     desired_state = chronos_job_config.get_desired_state()
+    soa_disabled_state = complete_config['disabled']
+
+    resolved_disabled_state = determine_disabled_state(desired_state,
+                                                       soa_disabled_state)
+
+    if resolved_disabled_state != soa_disabled_state:
+        log.info(("disabled state in deployments.json conflicts with state in"
+                  "soa_configs. setting disabled state to %r" %
+                  resolved_disabled_state))
+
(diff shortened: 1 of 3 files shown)


giuliano108:
Looks good to me, ShipIt!

paasta_tools/chronos_tools.py
```diff
@@ -471,23 +471,56 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
...
     return complete_config
 
 
+def determine_disabled_state(branch_state, soa_disabled_state):
+    """ Determine the *real* disabled state. There are two sources of truth for
+    the disabled/enabled state of a job: the 'disabled' flag in soa-configs,
+    and the state set in deployments.json by `paasta stop` or `paasta start`.
+    If the two conflict, then this determines the 'real' state for it.
+
+    We take a defensive stance here: if either source has disabled as True,
```
solarkennedy: Just throwing this out here.... what if we simply dropped `disabled: ` from soa configs entirely?

Then we have no more need for this function, and the there is only "one" way to enabled/disable a chronos job?

Rob-Johnson: Seems a good idea to me. We'd be more consistent with marathon apps then too.

Rob-Johnson:
closing in favour of #293

PR #289, opened 2016-03-04
Mesos 0.24.1

nhandler:
This change actually performs the mesos 0.24.1 upgrade. I split it out of my previous PR so that we could be certain the mesos.cli upgrade would work properly.


solarkennedy:
ship

Rob-Johnson:
ship

PR #290, opened 2016-03-04
Added an autoscaling_method field

mjksmith:


paasta_tools/marathon_tools.py
```diff
@@ -188,7 +187,7 @@ def copy(self):
         return self.__class__(self.service, self.instance, dict(self.config_dict), dict(self.branch_dict))
 
     def get_min_instances(self):
-        return self.config_dict.get('instances', 1)
+        return self.config_dict.get('min_instances', 1)
```
solarkennedy: ty. This is closer in line with AWS autoscaling words which is kinda the "gold standard"

paasta_tools/marathon_tools.py
```diff
@@ -204,33 +203,41 @@ def get_instances(self):
         :returns: The number of instances specified in the config, 0 if not
                   specified or if desired_state is not 'start'."""
         if self.get_desired_state() == 'start':
-            instances = self.get_min_instances()
             max_instances = self.get_max_instances()
             if max_instances is not None:
+                instances = self.get_min_instances()
```
solarkennedy: can you call this variable `min_instances` to make it more obvious where it came from?

paasta_tools/marathon_tools.py
```diff
@@ -204,33 +203,41 @@ def get_instances(self):
...
                         service=self.service,
                         instance=self.instance,
                     )
                 except NoNodeError:  # zookeeper doesn't have instance data for this app
-                    pass
+                    return instances
                 else:
-                    instances = max(
+                    return max(
                         min(zk_instances, max_instances),
```
solarkennedy: but won't zk_instances be not set by the time we get here if we get a NoNodeError ?

paasta_tools/marathon_tools.py
```diff
@@ -204,33 +203,41 @@ def get_instances(self):
...
-            return instances
+            else:
+                return self.config_dict.get('instances', 1)
         else:
             return 0
 
-    def get_backoff_seconds(self):
+    def get_autoscaling_params(self):
+        default_params = {
+            'method': 'default',
```
solarkennedy: I think this is good. Gives us room to grow.

paasta_tools/marathon_tools.py
```diff
@@ -326,8 +333,21 @@ def format_marathon_app_dict(self, app_id, docker_url, docker_volumes, service_n
...
+        docker_url = get_docker_url(system_paasta_config.get_docker_registry(), self.get_docker_image())
+        service_namespace_config = load_service_namespace_config(
+            service=self.service,
+            namespace=self.get_nerve_namespace(),
+        )
+        docker_volumes = system_paasta_config.get_volumes() + self.get_extra_volumes()
+
         complete_config = {
-            'id': app_id,
+            'id': partial_id,
```
solarkennedy: Just junk this and the whole concept of a partial id

I know it will cause a bounce, this is fine

paasta_tools/marathon_tools.py
```diff
@@ -204,33 +203,41 @@ def get_instances(self):
...
+    def get_backoff_seconds(self, instances=None):
         """backoff_seconds represents how many seconds a penalization factor for failing tasks.
         Every time a task fails, Marathon adds this value multiplied by a backoff_factor.
         In PaaSTA we know how many instances a service has, so we adjust the backoff_seconds
         to account for this, which prevents services with large number of instances from
         being penalized more than services with small instance counts. (for example, a service
         with 30 instances will get backed off 10 times faster than a service with 3 instances)."""
-        instances = self.get_instances()
+        if instances is None:
+            instances = self.get_instances()
```
solarkennedy: I understand why you are doing this, but another pattern you should consider is the "memoize" pattern if you haven't heard of it.

I think in paasta this pattern should be safe considering how we run things

tests/test_autoscaling_lib.py
```diff
@@ -0,0 +1,173 @@
...
+            mock.patch('paasta_tools.autoscaling_lib.load_system_paasta_config', autospec=True),
+    ) as (
+        mock_zk_client,
+        _,
+    ):
+        mock_zk_client.return_value = mock.Mock(get=mock.Mock(side_effect=NoNodeError))
+        assert fake_marathon_config.get_instances() == 5
+
+
+def test_get_zookeeper_instances_defaults_to_config_out_of_boumds():
```
solarkennedy: bounds

solarkennedy:
ship

PR #292, opened 2016-03-04
send OK event when a chronos job has been disabled

### LINKED ISSUE #284: ``check_chronos_job`` alerts when a job is disabled
following on from #283, I disabled a test job.

Once I had disabled it, I started receiving ``check_chronos_jobs`` alerts

``
Warning: example_service.mesosstage_robj_child isn't in chronos at all, which means it may not be deployed yet
``

This is incorrect: the job **is** in chronos, it's just disabled (and there have been no previous runs).

Here is the [offending code](https://github.com/Yelp/paasta/blob/master/paasta_tools/check_chronos_jobs.py#L132-L137)

Rob-Johnson:
closes #284

### DIFF
--- a/paasta_tools/check_chronos_jobs.py
+++ b/paasta_tools/check_chronos_jobs.py
@@ -104,10 +104,7 @@ def build_service_job_mapping(client, configured_jobs):
             instance=job[1],
             client=client,
         )
-        # filter the enabled jobs
-        enabled = chronos_tools.filter_enabled_jobs(matching_jobs)
-        # get the last run state for the job
-        with_states = last_run_state_for_jobs(enabled)
+        with_states = last_run_state_for_jobs(matching_jobs)
         service_job_mapping[job] = with_states
     return service_job_mapping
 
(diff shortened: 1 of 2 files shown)


paasta_tools/check_chronos_jobs.py
```diff
@@ -138,9 +135,13 @@ def sensu_message_status_for_jobs(service, instance, job_state_pairs):
         )
     else:
         full_job_id = job_state_pairs[0][0].get('name', '[Couldn\'t fetch job name]')
-        state = job_state_pairs[0][1]
-        sensu_status = sensu_event_for_last_run_state(state)
-        output = message_for_status(sensu_status, service, instance, full_job_id)
+        if job_state_pairs[0][0].get('disabled') is True:
```
solarkennedy: Why does this inspect the job_state_pairs? Can we ask the actual source of truth instead? (the instance_config.get_status()

That way we don't have to wait for a bounce?

paasta_tools/check_chronos_jobs.py
```diff
@@ -138,9 +135,13 @@ def sensu_message_status_for_jobs(service, instance, job_state_pairs):
...
+            sensu_status = pysensu_yelp.Status.OK
+            output = "Job %s%s%s is disabled - ignoring status." % (service, utils.SPACER, instance)
+        else:
+            state = job_state_pairs[0][1]
```
solarkennedy: I don't really understand why the job_state_pairs is 0,0 on the first one and 0,1 on the second one?

paasta_tools/check_chronos_jobs.py
Rob-Johnson: because the job at [0][0] is just a dict pulled from the api, not a ChronosJobConfig

paasta_tools/check_chronos_jobs.py
Rob-Johnson: job_state_pairs is [(job, state)] - so [0][1] gets the state, [0][0] gets the job itself. I should probably make these named tuples to make it a bit clearer.

solarkennedy:
ship

PR #293, opened 2016-03-04
Remove disabled field from chronos soa-configs

Rob-Johnson:


paasta_tools/chronos_tools.py
```diff
@@ -477,8 +473,8 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
     # this makes it trivial to compare configs and know when to bounce.
     complete_config['description'] = get_config_hash(complete_config)
 
-    # If the job was previously stopped, we should stop the new job as well
-    # NOTE this clobbers the 'disabled' param specified in the config file!
+    # use the 'description' field to determine the disabled state
+    # of the job
     if desired_state == 'start':
```
solarkennedy: This is.. not quiet right now.

The description field on line 474 needs to be calculated after this desired state. (after 481)

I guess this is a different bug.

paasta_tools/chronos_tools.py
```diff
@@ -477,8 +473,8 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
     # this makes it trivial to compare configs and know when to bounce.
     complete_config['description'] = get_config_hash(complete_config)
 
-    # If the job was previously stopped, we should stop the new job as well
-    # NOTE this clobbers the 'disabled' param specified in the config file!
+    # use the 'description' field to determine the disabled state
```
solarkennedy: This comment isn't correct anymore, I would just junk it. And then maybe move line 470 to line 477 to bring the related code closer together

paasta_tools/chronos_tools.py
Rob-Johnson: yeah, this was an existing bug - I fixed it in #288 but missed it here. fixed

solarkennedy:
ship if tests pass! Thanks!

PR #295, opened 2016-03-04
Order tasks in verbose paasta status

drolando:
`paasta status -v` returns the list of Non-Running Tasks in a random
order. Also it only shows the last 10, so you might see old runs but not
the more recent ones.

This change orders the not_running_tasks list by start time. Since
they're reversed in status_mesos_tasks_verbose they'll be printed from
the latest one to the oldest one.

AFAICT this script is run directly on the mesos master, so I wasn't able to really try it.

### DIFF
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -146,8 +146,6 @@ def get_running_tasks_from_active_frameworks(job_id):
 def get_non_running_tasks_from_active_frameworks(job_id):
     active_framework_tasks = get_current_tasks(job_id)
     not_running_tasks = filter_not_running_tasks(active_framework_tasks)
-    # Order the tasks by timestamp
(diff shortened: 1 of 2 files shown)


solarkennedy:
lgtm. Ship if tests pass.

paasta_tools/mesos_tools.py
```diff
@@ -127,6 +127,8 @@ def get_running_tasks_from_active_frameworks(job_id):
 def get_non_running_tasks_from_active_frameworks(job_id):
     active_framework_tasks = get_current_tasks(job_id)
     not_running_tasks = filter_not_running_tasks(active_framework_tasks)
+    # Order the tasks by timestamp
+    not_running_tasks.sort(key=lambda task: get_first_status_timestamp(task))
```
Rob-Johnson: I'd rather you moved this to the caller (paasta/tools:mesos_tools.py:251)

Rob-Johnson:
thanks @drolando

PR #298, opened 2016-03-04
refactored create_complete_config

mjksmith:


paasta_tools/marathon_tools.py
```diff
@@ -223,14 +223,15 @@ def get_instances(self):
...
-    def get_backoff_seconds(self):
+    def get_backoff_seconds(self, instances=None):
         """backoff_seconds represents how many seconds a penalization factor for failing tasks.
         Every time a task fails, Marathon adds this value multiplied by a backoff_factor.
         In PaaSTA we know how many instances a service has, so we adjust the backoff_seconds
         to account for this, which prevents services with large number of instances from
         being penalized more than services with small instance counts. (for example, a service
         with 30 instances will get backed off 10 times faster than a service with 3 instances)."""
-        instances = self.get_instances()
+        if instances is None:
```
solarkennedy: Why is it necessary to pass in instances like this?

paasta_tools/marathon_tools.py
```diff
@@ -326,8 +327,21 @@ def format_marathon_app_dict(self, app_id, docker_url, docker_volumes, service_n
...
+        docker_url = get_docker_url(system_paasta_config.get_docker_registry(), self.get_docker_image())
+        service_namespace_config = load_service_namespace_config(
+            service=self.service,
+            namespace=self.get_nerve_namespace(),
+        )
+        docker_volumes = system_paasta_config.get_volumes() + self.get_extra_volumes()
+
         complete_config = {
-            'id': app_id,
+            'id': partial_id,
```
solarkennedy: Why do we even need this partial id. Can we just get away with line 384?

It will cause a bounce I guess?

paasta_tools/setup_marathon_job.py
```diff
@@ -512,7 +512,7 @@ def setup_service(service, instance, client, marathon_config,
 
     log.info("Setting up instance %s for service %s", instance, service)
     try:
-        complete_config = marathon_tools.create_complete_config(service, instance, soa_dir)
+        complete_config = service_marathon_config.format_marathon_app_dict()
```
solarkennedy: What do you think about calling this variable marathon_app_dict?

paasta_tools/marathon_tools.py
mjksmith: @solarkennedy calling `format_marathon_app_dict` calls both `get_instances` and `get_backoff_seconds`. Since `instances` might read data from zookeeper this allows us to only query zookeeper

mjksmith: @solarkennedy another alternative would just be to use `min_instances` instead of `instances` here.

paasta_tools/marathon_tools.py
mjksmith: @solarkennedy yeah, I just did this to avoid a big bounce

paasta_tools/marathon_tools.py
solarkennedy: Yea. If this number is going to be dynamically changing a lot, then we can't use it in the hash computation. We have to use something more static. max_instances is actually what we really want

paasta_tools/marathon_tools.py
solarkennedy: Na, just do it

paasta_tools/marathon_tools.py
mjksmith: @solarkennedy this number is actually excluded from the hash calculation. We still want users to be able to change `min_instances` and `max_instances` without causing a bounce anyways

mjksmith: @solarkennedy fixed in #290

keshavdv: Can you also fix the docstring ("backoff_seconds represents how many seconds a penalization factor for failing tasks").

paasta_tools/marathon_tools.py
```diff
@@ -223,14 +223,15 @@ def get_instances(self):
         else:
             return 0
 
-    def get_backoff_seconds(self):
+    def get_backoff_seconds(self, instances=None):
         """backoff_seconds represents how many seconds a penalization factor for failing tasks.
```
Rob-Johnson: this sentence doesn't make sense. could you adjust it whilst you're here?

paasta_tools/marathon_tools.py
```diff
@@ -905,37 +925,12 @@ def create_complete_config(service, instance, soa_dir=DEFAULT_SOA_DIR):
     """Generates a complete dictionary to be POST'ed to create an app on Marathon"""
     # A set of config attributes that don't get included in the hash of the config.
```
Rob-Johnson: I think this comment is in the wrong place now?

paasta_tools/marathon_tools.py
```diff
@@ -361,6 +373,14 @@ def format_marathon_app_dict(self, app_id, docker_url, docker_volumes, service_n
         if accepted_resource_roles is not None:
             complete_config['accepted_resource_roles'] = accepted_resource_roles
 
+        code_sha = get_code_sha_from_dockerurl(docker_url)
+
+        config_hash = get_config_hash(
+            {key: value for key, value in complete_config.items() if key not in CONFIG_HASH_BLACKLIST},
+            force_bounce=self.get_force_bounce(),
```
Rob-Johnson: why is force_bounce a special case here instead of part of complete_config?

mjksmith: force_bounce is an argument passed to get_config_hash

paasta_tools/marathon_tools.py
mjksmith: @keshavdv fixed in #290

paasta_tools/marathon_tools.py
```diff
@@ -326,8 +327,21 @@ def format_marathon_app_dict(self, app_id, docker_url, docker_volumes, service_n
...
         :returns: A dict containing all of the keys listed above"""
+        instances = self.get_instances()
+
+        # A set of config attributes that don't get included in the hash of the config.
+        # These should be things that PaaSTA/Marathon knows how to change without requiring a bounce.
+        CONFIG_HASH_BLACKLIST = set(['instances', 'backoff_seconds'])
+
+        system_paasta_config = load_system_paasta_config()
+        docker_url = get_docker_url(system_paasta_config.get_docker_registry(), self.get_docker_image())
+        service_namespace_config = load_service_namespace_config(
```
solarkennedy: We are losing something here, and that is a soa_dir arg. :(

I think if you are going to do this kind of change where the object itself is responsible for reading in more config files, you are going to have to make the init take in the soa_dir arg to let it know "where it came from"

solarkennedy: Otherwise you would run one of our commands for with `soa_dir` pointed one place, but this function right here might blow up or read the wrong config file

mjksmith: @solarkennedy the only time `soa_dir` was used in `create_complete_config` was to load the correct `MarathonServiceConfig`.

solarkennedy: Yea but, that doesn't mean it was *right*, right?

solarkennedy:
ship

PR #299, opened 2016-03-04
Added 'framework' to the ServiceConfig objects

solarkennedy:
Right now we end up loading a config object and then storing how we ended up loading it (marathon vs chronos)

There are a couple places in our code where it would be easier for us to "just asked" the config object what framework it is for, so we can adjust say, or local-run behavior or the message we print for help in paasta stop, etc.

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -181,6 +181,7 @@ class ChronosJobConfig(InstanceConfig):
             config_dict=config_dict,
             branch_dict=branch_dict,
         )
+        self.framework = 'chronos'
 
     def get_service(self):
         return self.service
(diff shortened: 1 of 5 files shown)


mjksmith:
Looks good to me

Rob-Johnson:
doesn't the class itself contain this information? I don't know what we really get from this.
```
if config.framework == 'marathon':
    ....
elif config.framework == 'chronos':
    ....
```

rather than:

```
if isinstance(config, MarathonConfig):
  ...
elif isinstance(config, ChronosJobConfig):
  ....
```

can you give a more specific use case?

solarkennedy:
Nope, I'm mostly just an OOP noob :)

PR #300, opened 2016-03-06
Allow paasta to use docker http api

### LINKED ISSUE #287: Can't run paasta itests when using docker http api
When using OSX/boot2docker, you communicate with docker via the http api, rather than an unix socket.

Docker forces that you communicate with the docker daemon via https, but currently there is nothing in paasta that sets the correct locations for certs etc, so fails.

This ticket is to track being able to use paasta/run the tests outside of a ``/var/run/docker.sock`` setup.

keshavdv:
Fixes #287 and also enables us to use TLS in the future to secure Docker's api

### DIFF
--- a/general_itests/environment.py
+++ b/general_itests/environment.py
@@ -13,16 +13,13 @@
 # limitations under the License.
 import shutil
 
-from docker import Client
-
-from paasta_tools.utils import get_docker_host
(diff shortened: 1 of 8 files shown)


Rob-Johnson:
<3 thanks for this!

solarkennedy:
ship Thanks!

PR #301, opened 2016-03-07
Add stdout/err tail snippet to paasta status -vv (PAASTA-3268)

giuliano108:
`paasta status -vv` displays, for each Mesos task in the output (Marathon, Chronos, running and non-running) 10 lines from the tail of both `stdout` and `stderr`.

Happy path. `stdout`/`stderr` are purposely not indented (if they were, long lines wrapping would arguably make the output more confusing).
screen shot 2016-03-06 at 10 46 16 pm

Exceptions come from `mesos.cli`, they are trapped and displayed (`paasta status` still returns zero in that case).
screen shot 2016-03-06 at 10 54 15 pm

### DIFF
--- a/paasta_itests/steps/paasta_serviceinit_steps.py
+++ b/paasta_itests/steps/paasta_serviceinit_steps.py
@@ -110,6 +110,22 @@ def chronos_status_verbose_returns_healthy(context, service_instance):
     assert "Running Tasks:" in output
 
 
+@then((u'paasta_serviceinit status -vv for the service_instance "{service_instance}"'
+       ' exits with return code 0 and the correct output'))
+def paasta_serviceinit_tail_stdstreams(context, service_instance):
+    cmd = "../paasta_tools/paasta_serviceinit.py --soa-dir %s %s status -vv" % (context.soa_dir, service_instance)
+    print 'Running cmd %s' % cmd
+    (exit_code, output) = _run(cmd)
+    print 'Got exitcode %s with output:\n%s' % (exit_code, output)
+    print  # sacrificial line for behave to eat instead of our output
+
+    assert exit_code == 0
+    # The container we run doesn't really have a stdout/stderr. The message below
+    # comes from mesos.cli, proving that paasta_serviceinit tried to read stdout/sdterr,
+    # caught the Exception and presented the right information to the user.
+    assert "No such task has the requested file or directory" in output
+
+
 @when(u'we paasta_serviceinit emergency-stop the service_instance "{service_instance}"')
 def chronos_emergency_stop_job(context, service_instance):
     cmd = '../paasta_tools/paasta_serviceinit.py --soa-dir %s %s stop' % (context.soa_dir, service_instance)
--- a/paasta_itests/steps/paasta_serviceinit_steps.py
+++ b/paasta_itests/steps/paasta_serviceinit_steps.py
@@ -88,7 +88,7 @@ def marathon_restart_gets_new_task_ids(context, job_id):
 def chronos_status_returns_healthy(context, service_instance):
     cmd = '../paasta_tools/paasta_serviceinit.py --soa-dir %s %s status' % (context.soa_dir, service_instance)
     print 'Running cmd %s' % cmd
-    (exit_code, output) = _run(cmd)
+    exit_code, output = _run(cmd)
     print 'Got exitcode %s with output:\n%s' % (exit_code, output)
     print  # sacrificial line for behave to eat instead of our output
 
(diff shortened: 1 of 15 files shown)


paasta_itests/steps/paasta_serviceinit_steps.py
```diff
@@ -110,6 +110,22 @@ def chronos_status_verbose_returns_healthy(context, service_instance):
     assert "Running Tasks:" in output
 
 
+@then((u'paasta_serviceinit status -vv for the service_instance "{service_instance}"'
+       ' exits with return code 0 and the correct output'))
+def paasta_serviceinit_tail_stdstreams(context, service_instance):
+    cmd = "../paasta_tools/paasta_serviceinit.py --soa-dir %s %s status -vv" % (context.soa_dir, service_instance)
+    print 'Running cmd %s' % cmd
+    (exit_code, output) = _run(cmd)
```
Rob-Johnson: remove the surrounding brackets from (exit_code, output)

paasta_tools/chronos_serviceinit.py
```diff
@@ -245,8 +245,9 @@ def format_chronos_job_status(job, running_tasks, verbose):
 
     command = _format_command(job)
     mesos_status = _format_mesos_status(job, running_tasks)
-    if verbose:
-        mesos_status_verbose = status_mesos_tasks_verbose(job["name"], get_short_task_id)
+    if verbose > 0:
```
Rob-Johnson: can you give verbose a default parameter value?

paasta_tools/mesos_tools.py
```diff
@@ -226,13 +245,69 @@ def format_non_running_mesos_task_row(task, get_short_task_id):
...
+        output.append(error_message % (get_short_task_id(task['id']), 'timeout'))
+    return output
+
+
+def zip_tasks_verbose_output(table, stdstreams):
+    """Zip a list of strings (table) with a list of lists (stdstreams)
+    :param table: a formatted list of tasks
+    :param stdstreams: for each task, a list of lines from stdout/stderr tail
+    """
+    assert len(table) == len(stdstreams)
```
Rob-Johnson: instead of just throwing an assertion error, can you throw a ```ValueError('here's an explanation")```

paasta_tools/mesos_tools.py
```diff
@@ -226,13 +245,69 @@ def format_non_running_mesos_task_row(task, get_short_task_id):
...
+            for _ in xrange(nlines):
+                line = next(reversed_file, None)
+                if line is None:
+                    break
+                tail.append(line)
+            # reverse the tail, so that EOF is at the bottom again
+            if tail:
+                output.extend(tail[::-1])
+            output.append(PaastaColors.blue("      %s EOF" % fobj[0].path))
+    except (MasterNotAvailableException,
```
Rob-Johnson: can you be more specific about where these exceptions are thrown? as in, don't wrap this whole code block in try/catch - just the parts that will actually throw

tests/test_mesos_tools.py
```diff
@@ -422,3 +432,103 @@ def test_get_paasta_execute_docker_healthcheck_when_not_found():
...
+@mark.parametrize('test_case', [
+    [['taska', 'taskb'],  # test_case0 - OK
+     [['outlna1', 'outlna2', 'errlna1'],
+      ['outlnb1', 'errlnb1', 'errlnb2']],
+     ['taska', 'outlna1', 'outlna2', 'errlna1', 'taskb', 'outlnb1', 'errlnb1', 'errlnb2'],
+     False],
+    [['a'],  # test_case1 - can't zip different length lists
+     [1, 2],
+     None,
+     True]])
```
Rob-Johnson: This is kind of hard to read - can you split these test cases out into individual tests?

solarkennedy:
I think it looks good! ship

paasta_tools/mesos_tools.py
giuliano108: > can you be more specific about where these exceptions are thrown? as in, don't wrap this whole code block in try/catch - just the parts that will actually throw

Can I please _not_ do that? :)

The code can throw in `mesos.cli.cluster.files()` (line 503) and `reversed_file = reversed(fobj[0])` (line 271). 3 out of 5 Exceptions can be thrown at both locations. Two `except` blocks would mean building the same error message in two places, unless I add some logic to handle an "ok" flag to be checked right before returning from the function. And in my opinion readability would suffer.

Rob-Johnson: fair enough.

paasta_itests/steps/paasta_serviceinit_steps.py
giuliano108: > remove the surrounding brackets from (exit_code, output)

Removed the parens there, and anywhere else we were doing the same thing in the itests.

paasta_tools/chronos_serviceinit.py
giuliano108: > can you give verbose a default parameter value?

Good catch, I was meaning to do that.

tests/test_mesos_tools.py
giuliano108: > This is kind of hard to read - can you split these test cases out into individual tests?

As discussed, I'm dropping this.

Rob-Johnson:
shipit!

PR #303, opened 2016-03-07
Added a default autoscaling method

mjksmith:


paasta_tools/autoscaling_lib.py
```diff
@@ -55,17 +59,90 @@ def bespoke_autoscaling_method(*args, **kwargs):
 
 
 @register_autoscaling_method('default')
-def default_autoscaling_method(marathon_service_config):
-    # not implemented yet
-    return 0
+def default_autoscaling_method(marathon_service_config, delay=600, **kwargs):
+    PID_SETPOINT = 0.8
+    Kp = 0.2
```
Rob-Johnson: can you make these more descriptive please?

paasta_tools/autoscaling_lib.py
```diff
@@ -55,17 +59,90 @@ def bespoke_autoscaling_method(*args, **kwargs):
...
+    cpu_seconds_zk_node = '%s/cpu_seconds' % zookeeper_root
+
+    current_time = int(datetime.datetime.now().strftime('%s'))
+    error = 0.0
+    iterm = 0.0
+    pid_output = 0
+    average_cpu_seconds = 0.0
+
+    def clamp_value(number):
+        return min(max(number, -1), 1)
```
Rob-Johnson: a docstring on what this is doing would be good

mjksmith:
These last few pushes made autoscaling ignore marathon tasks that haven't had their healthchecks run yet (ignoring young services) and to sample CPU usage over the task's entire delay period without using sleep.

mjksmith:
Autoscaling is pretty slow right now, and I worry that it won't scale well when we start adding new services.

Right now `cProfile` indicates that over 95% of autoscale runtime is spend fetching data from somewhere (`{method 'recv' of '_socket.socket' objects}`) so I'm going to see if smart batching can fix anything.

paasta_tools/autoscaling_lib.py
```diff
@@ -31,78 +42,157 @@ def get_autoscaling_method(function_name):
...
+        return 0
+
+    job_id = marathon_service_config.format_marathon_app_dict()['id']
+
+    healthy_ids = {task.id for task in marathon_client.list_tasks(app_id=job_id) if task.health_check_results}
+
+    if not healthy_ids:
+        return 0
+
+    mesos_tasks = [task for task in get_running_tasks_from_active_frameworks(job_id) if task['id'] in healthy_ids]
```
solarkennedy: This is probably where all the time is spent.

For multi-service we are probably going to have to get this state once and iterate over all the services for speed :(

paasta_tools/autoscaling_lib.py
```diff
@@ -31,78 +42,157 @@ def get_autoscaling_method(function_name):
...
+    current_time = get_current_time()
+    time_delta = current_time - last_time
+
+    mesos_cpu_data = {task['id']: float(task.stats.get('cpus_system_time_secs', 0.0) + task.stats.get(
```
mjksmith: @solarkennedy `cProfile` indicates this is where most of the time is spent, specifically calling `task.stats`.

```
170602 function calls (168851 primitive calls) in 47.544 seconds
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
     1    0.000    0.000   33.837   33.837 autoscaling_lib.py:116(<dictcomp>)
 .....
   108    0.001    0.000   41.888    0.388 task.py:66(stats)
```

For some reason calling `task.mem_limit` at line 128 also takes a while (~0.25s / call totaling 8s overall) which would indicate `task.stats` isn't caching its result (`task.mem_limit` calls `task.stats.__getitem__` internally) but other methods that have that behaviour such as `task.rss` and `task.cpu_limit` complete very quickly (about 1ms per call).

paasta_tools/autoscaling_lib.py
```diff
@@ -31,78 +42,157 @@ def get_autoscaling_method(function_name):
...
+        except NoNodeError:
+            iterm = 0.0
+            last_error = 0.0
+            last_time = 0.0
+            last_cpu_data = []
+
+    if get_current_time() - last_time < delay:
+        return 0
+
+    job_id = marathon_service_config.format_marathon_app_dict()['id']
```
mjksmith: @solarkennedy this also takes a long time for services that don't have constraints set - running `marathon_tools.get_constraints` without any constraints set calls `mesos_tools.get_mesos_slaves_grouped_by_attribute`, which can take up to 15 seconds.

I guess this can be avoided just by filtering by service name and instance name, and either not autoscaling tasks that have multiple versions up (bouncing) or by ignoring the version difference and treating them all as one task for the purposes of autoscaling.

paasta_tools/autoscaling_lib.py
mjksmith: @solarkennedy I agree.

docs/source/autoscaling.rst
```diff
@@ -12,21 +12,45 @@ Enabling autoscaling
...
 
-Autoscaling parameters are stored in an ``autoscaling`` attribute of your instances as a dictionary. Within the ``autoscaling`` attribute, specifying a ``method`` will allow you to specify the logic that determines when to scale up and scale down your instance. If an autoscaling method isn't provided, the ``"default"`` autoscaling method will be used.
+Autoscaling parameters are stored in an ``autoscaling`` attribute of your instances as a dictionary. Within the ``autoscaling`` attribute, setting an ``ingester`` will allow you to specify a method that determines the utilization of your service. If an ingester isn't provided, the ``"mesos_cpu_ram"`` ingester will be used. Within the ``autoscaling`` attribute, setting a ``decider`` will allow you to specify the logic that determines when to autoscale your service. If a decider isn't provided, the ``"pid"`` decider will be used. Deciders and ingesters have their own optional keyword arguments that may be placed into the ``autoscaling`` dictionary as well.
```
Rob-Johnson: seems like we could do better with our nomenclature here - 'decider' doesn't seem great. I can't think of anything better right now though.

docs/source/autoscaling.rst
```diff
@@ -12,21 +12,45 @@ Enabling autoscaling
...
 
-Autoscaling methods
--------------------
+Autoscaling components
+----------------------
 
-The current autoscaling methods are:
+The currently available ingesters are:
```
Rob-Johnson: Make this a 'Ingesters' header (same with 'Deciders')

paasta_tools/autoscaling_lib.py
```diff
@@ -12,97 +12,304 @@
...
+
+_autoscaling_ingesters = {}
+_autoscaling_deciders = {}
 
-_autoscaling_methods = {}
 
+def register_autoscaling_component(name, method_type):
+    if method_type == 'ingester':
+        component_dict = _autoscaling_ingesters
+    elif method_type == 'decider':
```
Rob-Johnson: can you make these magic strings constants?

paasta_tools/autoscaling_lib.py
```diff
@@ -12,97 +12,304 @@
...
+class IngesterNoDataError(ValueError):
+    pass
 
 
-def get_instances_from_zookeeper(service, instance):
-    with ZookeeperPool() as zookeeper_client:
-        (instances, _) = zookeeper_client.get('%s/instances' % compose_autoscaling_zookeeper_root(service, instance))
-        return int(instances)
+@register_autoscaling_component('threshold', 'decider')
+def threshold_decider(marathon_service_config, ingester_method, marathon_tasks, mesos_tasks,
```
Rob-Johnson: can you add a docstring to this function, and change the function to return True/False rather than 0/-1 ?

paasta_tools/autoscaling_lib.py
```diff
@@ -12,97 +12,304 @@
...
+    elif abs(error) > threshold:
+        return -1
+    else:
+        return 0
+
+
+@register_autoscaling_component('pid', 'decider')
+def pid_decider(marathon_service_config, ingester_method, marathon_tasks, mesos_tasks,
+                delay=600, setpoint=0.8, **kwargs):
+    Kp = 0.2
```
Rob-Johnson: how about renaming these variables to something readable?

paasta_tools/autoscaling_lib.py
```diff
@@ -12,97 +12,304 @@
...
+
+    autoscaling_root = compose_autoscaling_zookeeper_root(
+        service=marathon_service_config.service,
+        instance=marathon_service_config.instance,
+    )
+    zk_iterm_path = '%s/pid_iterm' % autoscaling_root
+    zk_last_error_path = '%s/pid_last_error' % autoscaling_root
+    zk_last_time_path = '%s/pid_last_time' % autoscaling_root
+
+    def clamp_value(number):
```
keymone: what's the purpose of nested def here? it'll create a closure every time you call clamp_value and it doesn't seem like you need it.

docs/source/autoscaling.rst
```diff
@@ -12,21 +12,45 @@ Enabling autoscaling
...
+  :setpoint: the target utilization the controller aims for. Defaults to 0.8 (80%).
+  :delay: the number of seconds the decider must wait before fetching new data. Defaults to 600.
+:threshold: 
+  Autoscales when a service's utilization exceeds beyond a certain threshold.
+  
+  Autoscaling parameters:
+
+  :setpoint: the target utilization the controller aims for. Defaults to 0.8 (80%).
+  :threshold: the amount by which the setpoint must be exceeded in either direction before autoscaling is triggered. Defaults to 0.1 (10%).
+  :delay: the number of seconds the decider must wait before fetching new data. Defaults to 600.
```
Rob-Johnson: Could this delay be per-action? I'm thinking that you might want different delays for removing instances to adding instances - it might take significantly longer to bring a new instance up than to shut it down.

paasta_tools/autoscaling_lib.py
```diff
@@ -12,97 +12,304 @@
...
+    if error > threshold:
+        return 1
+    elif abs(error) > threshold:
+        return -1
+    else:
+        return 0
+
+
+@register_autoscaling_component('pid', 'decider')
+def pid_decider(marathon_service_config, ingester_method, marathon_tasks, mesos_tasks,
```
keymone: throw in a link to https://en.wikipedia.org/wiki/PID_controller in heredoc?

docs/source/autoscaling.rst
keymone: utilization_provider and decision_policy?

paasta_tools/autoscaling_lib.py
mjksmith: I can add a docstring, but I need three distinct return values to represent scale up, scale down and don't scale.

paasta_tools/autoscaling_lib.py
mjksmith: Kp, Ki, and Kd are the canonical PID constant names. I could do proportional_constant, integral_constant and derivative_constant? I think a docstring with a link to wikipedia's pid page would be more helpful.

docs/source/autoscaling.rst
mjksmith: I'm fine with changing 'decider' and 'ingester' to something else. utilization_provider and decision_policy sound good though I'd like to hear some other opinions/more agreement

Rob-Johnson: well, it doesn't really have to be that specific (it could be a metric other than utilizaton that you act on). I think something like metrics_provider would be better? decision_policy is good to me.

paasta_tools/autoscaling_lib.py
```diff
@@ -12,97 +12,354 @@
...
-def get_autoscaling_method(function_name):
-    return _autoscaling_methods[function_name]
+def get_autoscaling_ingester(name):
+    return _autoscaling_ingesters[name]
 
 
-def compose_autoscaling_zookeeper_root(service, instance):
-    return '/autoscaling/%s/%s' % (service, instance)
+def get_autoscaling_decider(name):
+    """
```
Rob-Johnson: is this docstring? this seems to return a decider?

Rob-Johnson:
shipit! This is really great work @mjksmith, thank you!

PR #306, opened 2016-03-08
ensure disabled flag is considered when computing config hash for chr…

### LINKED ISSUE #283: setting 'disable: true' in a chronos job is a noop unless you also 'paasta stop' the job

Rob-Johnson:
…onos job

we *should* be considering the disabled field when deciding if a job should be enabled, otherwise ``paasta stop`` doesn't work. This fixes #283

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -473,16 +473,16 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
     complete_config['name'] = compose_job_id(service, job_name)
     desired_state = chronos_job_config.get_desired_state()
 
-    # we use the undocumented description field to store a hash of the chronos config.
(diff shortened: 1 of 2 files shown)


giuliano108:
ShipIt.

PR #308, opened 2016-03-08
fixed the default autoscaling config args

mjksmith:
`self.config_dict.get('autoscaling', default_params)` doesn't work if `'autoscaling'` is only a partial dict -- e.g. there are two default params and `'autoscaling'` only contains one of them.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -227,7 +227,7 @@ class MarathonServiceConfig(InstanceConfig):
         default_params = {
             'method': 'default',
         }
-        return self.config_dict.get('autoscaling', default_params)
+        return dict(default_params, **self.config_dict.get('autoscaling', {}))
 
     def get_backoff_seconds(self):
         """backoff_seconds represents a penalization factor for relaunching failing tasks.


solarkennedy:
So.. is this a merge behavior you are after?

mjksmith:
@solarkennedy yes, if we have two default parameters and the user specifies one, we should only take the default for the parameter they didn't specify

paasta_tools/marathon_tools.py
```diff
@@ -227,7 +227,7 @@ def get_autoscaling_params(self):
         default_params = {
             'method': 'default',
         }
-        return self.config_dict.get('autoscaling', default_params)
```
solarkennedy: Seems ok. I've never seen this kwargs based merge thing. Is there a more pythonic way to do this operations? (dict.update?)

mjksmith: @solarkennedy not really since most python mutators don't return the result

The best you could do would be
```
result = {
    'method': 'default',
}
result.update(self.config_dict.get('autoscaling', {}))
return result
```

solarkennedy: meh. shipit

solarkennedy: I mean, ship the kwargs method as long as it works as intended.

Maybe be sure to update the docs?

PR #312, opened 2016-03-09
Do the mesos upgrade again

nhandler:
The mesos.cli upgrade appears to be working fine now. This PR is in preparation of trying the mesos 0.24.1 upgrade again.


solarkennedy:
ship

PR #313, opened 2016-03-09
Make FileLogWriter, which writes json-formatted log lines to a specified file.

### LINKED ISSUE #64: create alternative backend for logging

EvanKrall:
This depends on #281 and fixes #64.

This branch makes another LogWriter implementation: `file`, which writes to a specified file. I expect most people to want the file opened in append mode, so I've set that as the default.

It also allows you to format the path based on service / component / level / cluster / instance, in case you want different files per service or something.

cc @hashbrowncipher for sanity-checking my usage of flock and io.

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -16,9 +16,11 @@ from __future__ import print_function
 import contextlib
 import datetime
 import errno
+import fcntl
 import glob
 import hashlib
 import importlib
+import io
 import json
 import logging
 import os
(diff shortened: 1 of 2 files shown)


hashbrowncipher:
shipit. either with or without flock.

paasta_tools/utils.py
```diff
@@ -578,6 +580,50 @@ def log(self, service, line, component, level=DEFAULT_LOGLEVEL, cluster=ANY_CLUS
...
+    def log(self, service, line, component, level=DEFAULT_LOGLEVEL, cluster=ANY_CLUSTER, instance=ANY_INSTANCE):
+        path = self.format_path(service, component, level, cluster, instance)
+
+        # We use io.FileIO here because it guarantees that write() is implemented with a single write syscall,
+        # and on Linux, writes to O_APPEND files with a single write syscall are atomic.
+        #
+        # https://docs.python.org/2/library/io.html#io.FileIO
+        # http://article.gmane.org/gmane.linux.kernel/43445
+
+        to_write = format_log_line(level, cluster, service, instance, component, line) + self.line_delimeter
```
Rob-Johnson: "%s%s" % (format_log_line(level, cluster, service, instance, component, line), self.line_delimeter)

solarkennedy:
ship

PR #314, opened 2016-03-10
add some docs for paasta start/stop behaviour

Rob-Johnson:
I'm not sure this is the right place for these docs, but I think they're worth having.

### DIFF
--- a/docs/source/workflow.rst
+++ b/docs/source/workflow.rst
@@ -193,3 +193,51 @@ monitoring checks for free:
   instances are deployed on a cluster. If the service is registered in Smartstack
   it will look in Smartstack to count the available instances. Otherwise it
   counts the number of healthy tasks in Mesos.
+
+
+The PaaSTA command line
+------------------------
+
+The PaaSTA command line interface, ``paasta``, gives users of PaaSTA the
+ability to inspect the state of services, as well as stop and start existing
+services. See the man pages for a description and detail of options for any
+individual paasta command.  Some of the most frequently used commands are
+listed below:
+
+   * ``paasta start`` - sets the desired state of the service instance to
+  'started'. In the case of long-running services, this will mean ensuring that
+  the number of instances of your application matches that set in your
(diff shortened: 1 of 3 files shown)


docs/source/workflow.rst
```diff
@@ -193,3 +193,51 @@ monitoring checks for free:
...
+The PaaSTA command line
+------------------------
+
+The PaaSTA command line interface, ``paasta``, gives users of PaaSTA the
+ability to inspect the state of services, as well as stop and start existing
+services. See the man pages for a description and detail of options for any
+individual paasta command.  Some of the most frequently used commands are
+listed below:
+
+   * ``paasta start`` - sets the desired state of the service instance to
```
solarkennedy: This is good, but I wish we didn't have to duplicate it.

This "could" be the docstring of start/stop, and then the help text could just read the help string as part of the argparser, which would in turn get sucked into the manpage.
```
   add_parser(
     epilog=help(paasta_start)

def paasta_start()
"""Sets the desired state of the service....."""
```


Then in sphinx could we do
http://www.sphinx-doc.org/en/stable/domains.html#cross-referencing-python-objects

And do like...
```
See the man pages for fequlently bla:

  * ``paasta start`` - :py:meth:help(paasta_tools.cli.paasta_start)
```

This is all mostly a wishlist here. 

I won't block a shipit on being fancy with sphinx, but I will ask that you copy these docs into the help text of the commands themselves so they show up with --help (and end up in the manpage)

Rob-Johnson: I can't figure out how to do the sphinx magic - I got to the last hurdle, but can't get the ``:py:meth:help`` working.

>  I will ask that you copy these docs into the help text of the commands themselves so they show up with --help

Done for emergency-start and emergency-stop.

It's difficult for start and stop because we auto generate the contents of the description field. See [here](https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/start_stop_restart.py#L33-L47)

PR #316, opened 2016-03-10
Use authentication with mesos in itests

nhandler:
This change updates mesos/marathon/chronos to use [authentication](http://mesos.apache.org/documentation/latest/authentication/) to communicate. This more closely matches production, and it should allow us to spot issues (such as those caused by the mesos 0.24.1 upgrade) faster.

This PR also changes the itests to use mesos 0.23.1 again for the time being.


EvanKrall:
🐑it

PR #317, opened 2016-03-11
Mesos 0.24.1 Passes Tests!

nhandler:
I finally managed to get mesos 0.24.1 to pass the tests and have marathon properly authenticate. We were getting bitten by https://issues.apache.org/jira/browse/MESOS-3560

This issue is fixed in 0.26.0 via a breaking change.


EvanKrall:
Ship, though the more exciting change is to our puppet repo.

solarkennedy:
OMG why :(

How hard do you think it would be to go right to freaking .26?

nhandler:
Not sure. This 0.24.1 upgrade wasn't meant to be that bad, but look at all of the trouble it caused. I'll keep pushing to get us up-to-date, but I'm going to do it one at a time :)

PR #319, opened 2016-03-11
Upgrade to mesos 0.25.0

nhandler:
Now that I think I've sorted out all of the issues with mesos 0.24.1, it is time to upgrade to 0.25.0. I encountered one [issue](https://www.mail-archive.com/user@mesos.apache.org/msg04937.html) with this upgrade which I fixed (I will make a similar change within Puppet).

I will obviously hold off on merging this change until we are sure we have 0.24.1 working everywhere.

### DIFF
--- a/paasta_itests/docker-compose.yml
+++ b/paasta_itests/docker-compose.yml
@@ -18,7 +18,7 @@ mesosslave:
     - zookeeper
   environment:
     -CLUSTER: testcluster
-  command: 'mesos-slave --master=zk://zookeeper:2181/mesos-testcluster --resources="cpus(*):10; mem(*):512; disk(*):100" --credential=/etc/mesos-slave-secret'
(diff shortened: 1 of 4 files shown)


Rob-Johnson:
I'm sceptical of what the `` --launcher=posix`` actually does - it changes how the mesos containerizer works, which we don't use.

solarkennedy:
The docs say it will default to launcher=linux if on linux, sounds like we don't need it?
Also sounds like it doesn't matter because we use the docker containerizer only.

nhandler:
Well, poking around, we don't actually specify a `--containerizers` argument when launching `mesos-slave` in `paasta_itests/docker-compose.yml`. Per http://mesos.apache.org/documentation/latest/configuration/ , this causes it to default to `mesos`.

solarkennedy:
Ship

PR #320, opened 2016-03-12
Make local-run bomb out if cook-image fails

solarkennedy:
cc @drolando


Rob-Johnson:
shipit

drolando:
+1

solarkennedy:
Looks like I need to fix a general itest. Will fix+ship

PR #321, opened 2016-03-12
Display more data when doing the healthcheck during local-run

solarkennedy:
This is papercut that should help people figure out what is wrong with their container during local-run.


Rob-Johnson:
shipit

PR #325, opened 2016-03-14
Fix `paasta status -vv`

giuliano108:
PR #295 taught `paasta status -v` to sort/truncate the list of non running Mesos tasks. It introduced a bug that broke `paasta status -vv`: the truncated list of tasks was not consistently used throughout the code, leading to an `Can only zip same-length lists` exception.

b974221d08c9a3bc01793b494ac04d98b51ef717 fixes said bug and updates the tests.

01df0eed16b30ac0ff2578a555bfb192905e1032 attempts to remove some duplication in `status_mesos_tasks_verbose`.

### DIFF
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -351,7 +351,7 @@ def status_mesos_tasks_verbose(job_id, get_short_task_id, tail_stdstreams=False)
         output.extend(tasks_table_non_running)
     else:
         tasks_stdstreams_non_running = []
-        for task in non_running_tasks:
+        for task in non_running_tasks_ordered:
             tasks_stdstreams_non_running.append(format_stdstreams_tail_for_task(task, get_short_task_id))
         output.append(tasks_table_non_running[0])  # header
         output.extend(zip_tasks_verbose_output(tasks_table_non_running[1:], tasks_stdstreams_non_running))
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -300,6 +300,46 @@ def zip_tasks_verbose_output(table, stdstreams):
     return output
 
 
+def format_task_list(tasks, list_title, table_header, get_short_task_id, format_task_row, grey, tail_stdstreams):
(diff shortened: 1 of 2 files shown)


paasta_tools/mesos_tools.py
```diff
@@ -300,6 +300,46 @@ def zip_tasks_verbose_output(table, stdstreams):
...
+    :param get_short_task_id: A function which given a
+                              task_id returns a short task_id suitable for
+                              printing.
+    :param format_task_row: Formatting function, works on a task and a get_short_task_id function.
+    :param tail_stdstreams: If True, also display the stdout/stderr tail,
+                            as obtained from the Mesos sandbox.
+    :param grey: If True, the list will be made less visually prominent.
+    :return output: Formatted output (list of output lines).
+    """
+    if not grey:
```
solarkennedy: What if the argument was just called `color` and it default to a noop (lambda?) and the called could pick? not a blocker just an alternative suggestion

paasta_tools/mesos_tools.py
```diff
@@ -312,49 +352,45 @@ def status_mesos_tasks_verbose(job_id, get_short_task_id, tail_stdstreams=False)
...
-    else:
-        tasks_stdstreams_running = []
-        for task in running_and_active_tasks:
-            tasks_stdstreams_running.append(format_stdstreams_tail_for_task(task, get_short_task_id))
-        output.append(tasks_table_running[0])  # header
-        output.extend(zip_tasks_verbose_output(tasks_table_running[1:], tasks_stdstreams_running))
+    ]
+    output.extend(format_task_list(
+        running_and_active_tasks,
+        list_title,
```
solarkennedy: Can you do kwargs on both calls on this to make it more obvious for what "False" refers to, etc?

solarkennedy:
ship

paasta_tools/mesos_tools.py
Rob-Johnson: +1. default could be PaastaColors.white, and then colorize becomes ``lambda color_func: color_func(x)``

Rob-Johnson:
shipit

PR #326, opened 2016-03-14
Use tags rather than branches for deployments

nhandler:
This change migrates us from using branches to using tags for deployments. Ever since #252 was merged, we have been creating -deploy tags in addition to the git branches. With this change, we will now actually use those tags and stop creating branches.

### DIFF
--- a/general_itests/steps/deployments_json_steps.py
+++ b/general_itests/steps/deployments_json_steps.py
@@ -32,6 +32,7 @@ from paasta_tools.cli.cmds.mark_for_deployment import paasta_mark_for_deployment
 from paasta_tools.cli.cmds.start_stop_restart import paasta_stop
 from paasta_tools.utils import format_tag
 from paasta_tools.utils import format_timestamp
+from paasta_tools.utils import get_paasta_branch_from_deploy_group
 from paasta_tools.utils import get_paasta_tag_from_deploy_group
 from paasta_tools.utils import load_deployments_json
 
@@ -58,7 +59,7 @@ def step_impl_given(context):
     object_store.add_object(tree)
     object_store.add_object(commit)
 
-    context.test_git_repo.refs['refs/heads/paasta-test_cluster.test_instance'] = commit.id
+    context.test_git_repo.refs['refs/heads/master'] = commit.id
     context.expected_commit = commit.id
 
 
@@ -172,3 +173,10 @@ def step_impl_then_correctly_tagged(context):
     expected_formatted_tag = format_tag(expected_tag)
     assert expected_formatted_tag in context.test_git_repo.refs
     assert context.test_git_repo.refs[expected_formatted_tag] == context.expected_commit
+
+
+@then(u'the repository should not have old style branches')
+def step_impl_then_no_old_style_branches(context):
+    old_style_branch = get_paasta_branch_from_deploy_group(identifier='test_cluster.test_instance')
+    formatted_old_style_branch = 'refs/heads/%s' % old_style_branch
+    assert formatted_old_style_branch not in context.test_git_repo.refs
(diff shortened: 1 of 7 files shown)


paasta_tools/generate_deployments_for_service.py
```diff
@@ -104,6 +104,31 @@ def get_instance_config_for_service(soa_dir, service):
...
+              tag (with the most recent timestamp)  and sha is the sha it points at
+    """
+    most_recent_dtime = None
+    most_recent_ref = None
+    most_recent_sha = None
+    pattern = re.compile("^refs/tags/paasta-%s-(\d{8}T\d{6})-deploy$" % deploy_group)
+
+    for ref_name, sha in refs.iteritems():
+        match = pattern.match(ref_name)
+        if match:
```
Rob-Johnson: where's the else to the if? do we just ignore it?

Rob-Johnson: ah, I guess we do. this comes from the git repo, so could have any amount of tags/branches. ignore me

Rob-Johnson:
lg2m. ship

solarkennedy:
This particular change has been very difficult to have confidence about because it only runs on one box so there is not a super great way to "test" it in real life.

Can you do this procedure:

```
ssh sysgit
git clone soa-configs && cd soa-configs
git checkout deployments_test
rm .gitignore
cp -a /nail/etc/services/* .
git add -A
git commit -m "Added baseline deployments.jsons"

paasta list | shuf | xargs -n 1 -r -P 4 ../../your_paasta_clone/generate_deployments_for_service -y . -s
git diff
```

In other words, run *this* version of generate_deployments_for_service and do a `git diff` to show the differences in the deployments.jsons' that it *would* generate compared to what is currently generated.
Make sense?

nhandler:
After running the new and old `generate_deployments_for_service` side-by-side, it looks like the change should work as expected.

EvanKrall:
lgtm

general_itests/steps/deployments_json_steps.py
```diff
@@ -172,3 +173,10 @@ def step_impl_then_correctly_tagged(context):
     expected_formatted_tag = format_tag(expected_tag)
     assert expected_formatted_tag in context.test_git_repo.refs
     assert context.test_git_repo.refs[expected_formatted_tag] == context.expected_commit
+
+
+@then(u'the repository should not have old style branches')
+def step_impl_then_no_old_style_branches(context):
```
mjksmith: I don't know how I feel about this 'making sure we no longer do things the old way' testing. In the future this test won't have any context outside of old versions of paasta.

paasta_tools/generate_deployments_for_service.py
```diff
@@ -104,6 +104,31 @@ def get_instance_config_for_service(soa_dir, service):
...
+    :param refs: A dictionary mapping git refs to shas
+    :param deploy_group: The deployment group to return a deploy tag for
+
+    :returns: A tuple of the form (ref, sha) where ref is the actual deployment
+              tag (with the most recent timestamp)  and sha is the sha it points at
+    """
+    most_recent_dtime = None
+    most_recent_ref = None
+    most_recent_sha = None
+    pattern = re.compile("^refs/tags/paasta-%s-(\d{8}T\d{6})-deploy$" % deploy_group)
```
mjksmith: AFAIK we have this regex somewhere else already. Can we reuse that?

nhandler: Do we? We have some similar functions that are used to create the tags (that generate the timestamp and support various desired_states). But I don't remember (and am not finding with grep) any other places with this particular regex. If you can find one, I would be happy to remove the duplication and create a new function that takes in a deploy_group as an argument and returns the compiled regex.

general_itests/steps/deployments_json_steps.py
nhandler: Fair. I think I would still prefer to have the test for now and ticket the cleanup of it down the road. It is only an additional `Then` step tacked on to the end of an existing scenario. So the extra time it adds should be negligible.

PR #328, opened 2016-03-14
Allow 'paasta re.start -c' to use a list of clusters

nosmo:
Currently paasta restart -c accepts only one cluster. There has been an internal request to allow -c to use a list of clusters. 

There is some inconsistency around how this is currently implemented, with emergency_scale, emergency_restart, emergency_stop, emergency_start, restart and local_run using ```--cluster```, and logs, status and metastatus supporting ```--clusters```.

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -202,11 +202,14 @@ def paasta_start_or_stop(args, desired_state):
                 desired_state=desired_state,
             )
 
+    return_val = 0
     if invalid_deploy_groups:
         print "No branches found for %s in %s." % \
             (", ".join(invalid_deploy_groups), remote_refs)
-        print "Has it been deployed there yet?"
-        return 1
+        print "Has %s been deployed there yet?" % service
+        return_val = 1
+
+    return return_val
 
 
 def paasta_start(args):
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -57,8 +57,15 @@ def add_subparser(subparsers):
         ).completer = lazy_choices_completer(list_instances)
         status_parser.add_argument(
             '-c', '--clusters',
+            choices=list_clusters(),
             help="A comma-separated list of clusters to view. Defaults to view all clusters.\n"
             "For example: --clusters norcal-prod,nova-prod"
+        ).completer = lazy_choices_completer(list_clusters)
+        status_parser.add_argument(
+            '--cluster',
+            choices=list_clusters(),
+            help="A single cluster to view.\n"
+            " Deprecated! For example: --cluster norcal-prod. Will supersede any use of --clusters."
         ).completer = lazy_choices_completer(list_clusters)
 
         status_parser.add_argument(
@@ -151,18 +158,16 @@ def paasta_start_or_stop(args, desired_state):
     clusters = args.clusters
     soa_dir = args.soa_dir
     service = figure_out_service_name(args=args, soa_dir=soa_dir)
+
     if args.clusters is not None:
         clusters = args.clusters.split(",")
     else:
         clusters = list_clusters(service)
 
-    service_config = get_instance_config(
-        service=service,
-        cluster=clusters[0],
-        instance=instance,
-        soa_dir=soa_dir,
-        load_deployments=False,
-    )
+    if args.cluster is not None:
+        print ("Warning - the use of --cluster is deprecated and will be removed in "
+               "favour of --clusters in future versions.")
+        clusters = [args.cluster]
 
     try:
         remote_refs = remote_git.list_remote_refs(utils.get_git_url(service))
@@ -177,12 +182,7 @@ def paasta_start_or_stop(args, desired_state):
         print msg
         return 1
 
-    if 'refs/heads/paasta-%s' % service_config.get_deploy_group() not in remote_refs:
-        print "No branches found for %s in %s." % \
-            (service_config.get_deploy_group(), remote_refs)
-        print "Has it been deployed there yet?"
-        return 1
-
+    invalid_deploy_groups = []
     for cluster in clusters:
         service_config = get_instance_config(
             service=service,
@@ -192,12 +192,21 @@ def paasta_start_or_stop(args, desired_state):
             load_deployments=False,
         )
 
-        force_bounce = utils.format_timestamp(datetime.datetime.utcnow())
-        issue_state_change_for_service(
-            service_config=service_config,
-            force_bounce=force_bounce,
-            desired_state=desired_state,
-        )
+        if 'refs/heads/paasta-%s' % service_config.get_deploy_group() not in remote_refs:
+            invalid_deploy_groups.append(service_config.get_deploy_group())
+        else:
+            force_bounce = utils.format_timestamp(datetime.datetime.utcnow())
+            issue_state_change_for_service(
+                service_config=service_config,
+                force_bounce=force_bounce,
+                desired_state=desired_state,
+            )
+
+    if invalid_deploy_groups:
+        print "No branches found for %s in %s." % \
+            (", ".join(invalid_deploy_groups), remote_refs)
+        print "Has it been deployed there yet?"
+        return 1
 
 
 def paasta_start(args):
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -95,8 +95,9 @@ def make_mutate_refs_func(service_config, force_bounce, desired_state):
 def log_event(service_config, desired_state):
     user = utils.get_username()
     host = socket.getfqdn()
-    line = "Issued request to change state of %s to '%s' by %s@%s" % (
-        service_config.get_instance(), desired_state, user, host)
+    line = "Issued request to change state of %s (an instance of %s) to '%s' by %s@%s" % (
+        service_config.get_instance(), service_config.get_service(),
+        desired_state, user, host)
     utils._log(
         service=service_config.get_service(),
         level='event',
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -55,10 +56,11 @@ def add_subparser(subparsers):
             required=True,
         ).completer = lazy_choices_completer(list_instances)
         status_parser.add_argument(
-            '-c', '--cluster',
-            help='The PaaSTA cluster that has the service you want to %s. Like norcal-prod' % lower,
-            required=True,
-        ).completer = lazy_choices_completer(utils.list_clusters)
+            '-c', '--clusters',
```
nhandler: There is an [`nargs`](https://docs.python.org/2/library/argparse.html#nargs) option you can use to get a list.

Also, since we know the list of possible clusters, it might nice to start using the [choices](https://docs.python.org/2/library/argparse.html#choices) option.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -145,43 +148,48 @@ def issue_state_change_for_service(service_config, force_bounce, desired_state):
 def paasta_start_or_stop(args, desired_state):
     """Requests a change of state to start or stop given branches of a service."""
     instance = args.instance
-    cluster = args.cluster
+    clusters = args.clusters
     soa_dir = args.soa_dir
     service = figure_out_service_name(args=args, soa_dir=soa_dir)
+    if args.clusters is not None:
```
nhandler: If you switch to `nargs` as noted above, you will be able to drop this.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -145,43 +148,48 @@ def issue_state_change_for_service(service_config, force_bounce, desired_state):
...
-        print "Has it been deployed there yet?"
-        return 1
-
-    force_bounce = utils.format_timestamp(datetime.datetime.utcnow())
-    issue_state_change_for_service(
-        service_config=service_config,
-        force_bounce=force_bounce,
-        desired_state=desired_state,
-    )
+        try:
```
nhandler: These refs do not change based on the cluster we are operating on. Move this out of the loop. This will also prevent the user from seeing multiple identical error messages in the event that they can't communicate with the git server.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -145,43 +148,48 @@ def issue_state_change_for_service(service_config, force_bounce, desired_state):
...
+                "Error talking to the git server: %s\n"
+                "This PaaSTA command requires access to the git server to operate.\n"
+                "The git server may be down or not reachable from here.\n"
+                "Try again from somewhere where the git server can be reached, "
+                "like your developer environment."
+            ) % str(e)
+            print msg
+            return 1
+
+        if 'refs/heads/paasta-%s' % service_config.get_deploy_group() not in remote_refs:
```
nhandler: Depending on the order that the PRs get merged, you might end up with a conflict for this section with #326. It will probably be easiest if we merge this PR first and then the other one.

tests/cli/test_cmds_start_stop_restart.py
```diff
@@ -132,6 +132,7 @@ def test_stop_or_start_handls_ls_remote_failures(
     mock_stdout,
 ):
     args = mock.Mock()
+    args.clusters = 'cluster1,cluster2'
```
nhandler: For the sake of avoiding a regression, it might be nice to keep a test that makes sure you can run this command with only one cluster specified (and maybe one making sure that no clusters defaults to all clusters).

paasta_tools/cli/cmds/start_stop_restart.py
nosmo: Using nargs would be nice but this would introduce a third difference between this command and others - the existing --clusters commands use comma-separated lists. I'm happy to change those also but it seems like a bit of a big shift in usage.

nhandler: Fair. It should probably happen regardless, but that can be done as a separate PR later.

nhandler:
Looks like you also have some test failures that need fixing.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -145,13 +148,17 @@ def issue_state_change_for_service(service_config, force_bounce, desired_state):
...
     service = figure_out_service_name(args=args, soa_dir=soa_dir)
+    if args.clusters is not None:
+        clusters = args.clusters.split(",")
+    else:
+        clusters = list_clusters(service)
 
     service_config = get_instance_config(
         service=service,
-        cluster=cluster,
+        cluster=clusters[0],
```
mjksmith: Doing this means you're only going to validate that a deployment exists for the service config for the first cluster in the clusters list. Can you validate every service config instead? Something like 
```
service_configs = [get_instance_config(...) for cluster in clusters]
invalid_deploy_groups = [service_config.get_deploy_group() for service_config in service_configs if service_config.get_deploy_groups() not in remote_refs]
if invalid_deploy_groups:
    print "No branches found for %s in %s." % \
             (' ,'.join(invalid_deploy_groups), remote_refs)
          print "Has %s been deployed there yet?" % service
          return 1
```
Around line 180

mjksmith:
BTW this change looks safe in terms of existing scripts that use `paasta restart` -- opengrok shows that everyone who has a paasta restart script uses `-c` instead of `--cluster``.

nhandler:
@mjksmith I noticed that. Although, now that I think about it, it might be nice to support `--cluster` in addition to `--clusters` for a release to ease the manual transition (ideally showing a warning that it is deprecated in favor of `--clusters`).

paasta_tools/cli/cmds/start_stop_restart.py
nosmo: Should a bad service config halt execution altogether or should the command continue with any successful ones?

nosmo: I've pushed some changes that will halt execution if the git repo is in accessible but the function won't prematurely halt if there are any valid deploy groups, just log once complete. Not sure if this is the right way to do things, open to correction

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -170,19 +182,32 @@ def paasta_start_or_stop(args, desired_state):
...
+                service_config=service_config,
+                force_bounce=force_bounce,
+                desired_state=desired_state,
+            )
+
+    if invalid_deploy_groups:
         print "No branches found for %s in %s." % \
-            (service_config.get_deploy_group(), remote_refs)
+            (", ".join(invalid_deploy_groups), remote_refs)
         print "Has it been deployed there yet?"
```
mjksmith: Now that we might output multiple deploy groups here it might make sense to use quantity-neutral language here.
I'm all for `print "Has %s been deployed there yet?" % service`

mjksmith: We should mention that even though these deploy groups were ignored, the others were still started/stopped.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -55,10 +56,18 @@ def add_subparser(subparsers):
...
+            '-c', '--clusters',
+            choices=list_clusters(),
+            help="A comma-separated list of clusters to view. Defaults to view all clusters.\n"
+            "For example: --clusters norcal-prod,nova-prod"
+        ).completer = lazy_choices_completer(list_clusters)
+        status_parser.add_argument(
+            '--cluster',
+            choices=list_clusters(),
+            help="A single cluster to view.\n"
+            " Deprecated! For example: --cluster norcal-prod. Will supersede any use of --clusters."
```
mjksmith: Isn't `--clusters` superseding `--cluster`?

nosmo: In a more immediate sense the use of paasta restart --cluster mycluster --clusters othercluster,othercluster2 will result in the command only operating on mycluster. I'll change the wording to be less ambiguous

paasta_tools/cli/cmds/start_stop_restart.py
nosmo: Users will still see ```Issued request to change state of main to 'start' by user@host``` before the output about invalid deploy groups, that will hopefully suffice.

paasta_tools/cli/cmds/start_stop_restart.py
mjksmith: Maybe `--clusters` should supersede `--cluster` in this sense then. Or just throw a error if both are specified.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -55,10 +57,18 @@ def add_subparser(subparsers):
...
+            '-c', '--clusters',
+            choices=list_clusters(),
+            help="A comma-separated list of clusters to view. Defaults to view all clusters.\n"
+            "For example: --clusters norcal-prod,nova-prod"
+        ).completer = lazy_choices_completer(list_clusters)
+        status_parser.add_argument(
+            '--cluster',
+            choices=list_clusters(),
+            help="A single cluster to view.\n"
+            " Deprecated! For example: --cluster norcal-prod."
```
solarkennedy: We don't need this.

With python args you can use a substring of the arg and it will work.

For example you can say `paasta restart --serv foo`

Just make a "--clusters" arg and it will automatically be reference-able as -- cluster.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -145,17 +156,23 @@ def issue_state_change_for_service(service_config, force_bounce, desired_state):
...
+    if args.clusters and args.cluster:
+        print "--cluster and --clusters cannot be used at the same time"
+        return 1
+
+    if args.clusters is not None:
+        clusters = args.clusters.split(",")
+    else:
+        clusters = list_clusters(service)
+
+    if args.cluster is not None:
```
solarkennedy: We don't need this, just stick with `--clusters` and --cluster will still work

mjksmith:
Looks good to me

nhandler:
Ship It!

PR #331, opened 2016-03-17
stopped per-instance configs from nuking global defaults

mjksmith:
Fixed a bug where your entire `monitoring.yaml` config changes would be blown away if an instance had any monitoring configs. Probably applies to non-monitoring config files as well.

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -1345,7 +1345,7 @@ def format_table(rows, min_spacing=2):
 
 def deep_merge_dictionaries(source, destination):
     """
-    Merges two dicitonaries.
+    Merges two dictionaries.
     """
     result = copy.deepcopy(destination)
     stack = [(source, result)]
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -14,6 +14,7 @@
 from __future__ import print_function
 
 import contextlib
+import copy
 import datetime
 import errno
 import fcntl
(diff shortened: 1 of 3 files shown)


EvanKrall:
Probably worth some tests on your deep_merge_dictionaries function

solarkennedy:
Hmm. Don't we do (very verbose) monitoring getters that look in both places? Maybe those are all going to be obsolete with this change. I can't think of other config places that we have dictionaries like that, autoscaling might be a new one though.

EvanKrall:
Context: https://irclogs.yelpcorp.com/view/paasta/2016-03-16.html#161/h161,163,164,165,166,167,183,184,185,186,187,188,189,190,198,199,201,204,205,222,223,224,225,226,227

paasta_tools/utils.py
```diff
@@ -1340,3 +1340,16 @@ def format_table(rows, min_spacing=2):
             expanded_rows.append(expanded_row)
 
     return [(' ' * min_spacing).join(r) for r in expanded_rows]
+
+
+def deep_merge_dictionaries(source, destination):
```
Rob-Johnson: As the comment mentions [here](https://stackoverflow.com/questions/20656135/python-deep-merge-dictionary-data) what happens if an element in A is a dict but in B is not?

mjksmith: @Rob-Johnson either an `AttributeError` or a `KeyError` will be raised. Most other implementations I've seen just check for this and raise an exception. I suppose in this case we'll want to default to the per-instance configs.

mjksmith: This is fixed in the latest commit. If a value and a dictionary share the same key, whatever is in the per-instance config will override the per-service config.

Rob-Johnson:
How would you feel about this creating a new dictionary,  rather than modifying the dict passed in?

```
foo = {foo: {bar: baz}}; 
blargh = {foo: { bar: otherbaz}}
merged = deep_merge_dictionaries(bar, foo)
assert foo == {foo: {bar: baz}}
```

mjksmith:
@Rob-Johnson I agree with this being better. Changed in dd9b421

mjksmith:
@solarkennedy looking at those monitoring getters in `paasta_tools.monitoring_tools`, they're only used to get specific keys, so they're a lot faster than merging entire dictionaries.

Rob-Johnson:
shipit

PR #344, opened 2016-03-21
added schedule_time_zone to chronos schema

### LINKED ISSUE #338: Include schedule_time_zone in schema
Our `paasta_tools/cli/schemas/chronos_schema.json` file does not include the `schedule_time_zone`. We now support this field and include it on http://paasta.readthedocs.org/en/latest/yelpsoa_configs.html#chronos-clustername-yaml. The schema should be updated so that `paasta validate` does not claim services that utilize it are invalid.

mjksmith:
A couple of services were failing `paasta validate` because of this.

Fixes #338


nhandler:
Ship It!

PR #345, opened 2016-03-21
Move configuration of replication monitoring to yelpsoa config

nosmo:
Being able to configure per-component monitoring would be useful for certain services. Currently replication percentages can only be checked across the board via one command line argument to `check_marathon_services_replication.py`, and this PR moves this configuration to an instance's yelpsoa configuration file in the `replication_crit` key. 

The default of 50 is retained if the `replication_crit` value is not specified. According to puppet and opengrok check_marathon_services_replication is not in use with an argument anywhere.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -179,6 +179,11 @@ instance MAY have:
     to determine the order in which to build & deploy deploy groups. Defaults to
     ``clustername.instancename``. See the deploy group doc_ for more information.
 
+  * ``replication_crit``: An integer representing the percentage of instances that
+    need to be available for monitoring purposes. If less than ``replication_crit``
+    percent instances of a service's backends are not available, the monitoring
+    scripts will send a CRITICAL alert.
+
 In addition, each instancename MAY configure additional Marathon healthcheck
 options:
 
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -179,8 +179,8 @@ instance MAY have:
     to determine the order in which to build & deploy deploy groups. Defaults to
     ``clustername.instancename``. See the deploy group doc_ for more information.
 
-  * ``replication_crit``: An integer representing the percentage of instances that
-    need to be available for monitoring purposes. If less than ``replication_crit``
+  * ``replication_threshold``: An integer representing the percentage of instances that
+    need to be available for monitoring purposes. If less than ``replication_threshold``
     percent instances of a service's backends are not available, the monitoring
     scripts will send a CRITICAL alert.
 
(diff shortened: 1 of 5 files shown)


paasta_tools/check_marathon_services_replication.py
```diff
@@ -198,12 +193,13 @@ def send_event_if_under_replication(
     service,
     instance,
     cluster,
-    crit_threshold,
     expected_count,
     num_available,
     soa_dir,
 ):
     full_name = compose_job_id(service, instance)
+    job_config = marathon_tools.load_marathon_service_config(service, instance, cluster)
```
EvanKrall: Seems like we should be looking this up in check_healthy_marathon_tasks_for_service_instance; ideally we'd avoid re-reading the marathon.yaml over and over.

docs/source/yelpsoa_configs.rst
```diff
@@ -179,6 +179,11 @@ instance MAY have:
     to determine the order in which to build & deploy deploy groups. Defaults to
     ``clustername.instancename``. See the deploy group doc_ for more information.
 
+  * ``replication_crit``: An integer representing the percentage of instances that
```
Rob-Johnson: I'd prefer 'replication_threshold' rather than 'replication_crit'

paasta_tools/check_marathon_services_replication.py
```diff
@@ -26,7 +26,7 @@
 
 After retrieving that information, a fraction of available instances is calculated
 (available/expected), and then compared against a threshold. The default threshold
-is .50, meaning if fewer than 50 of a service's backends are available, the script
+is 50, meaning if less than 50% of a service's backends are available, the script
```
Rob-Johnson: isn't 50% inaccurate now? this should be 'less than ``replication_crit`` of a services...

nosmo: The default is still 50%: https://github.com/Yelp/paasta/blob/move-replication-monitoring-to-config/paasta_tools/marathon_tools.py#L525-L526 

Should this be removed? I'll clarify the language above either way

Rob-Johnson: heh, sorry, the diff hid the context from me.

paasta_tools/check_marathon_services_replication.py
nosmo: We're already looking up job_config here. Moving it so there isn't a lookup every time seems like a good idea but this will involve making monitoring_blacklist a parameter for most related functions also.

nosmo: Honestly I think it might be a little too messy to add another parameter to the various functions here. It wouldn't be the end of the world but this check adds no overhead over what was there before

Rob-Johnson:
shipit! :ship:

mjksmith:
LGTM

PR #352, opened 2016-03-23
Clean up orphaned docker container script

nosmo:
🐄 Make the orphaned docker script a little more monitoring-friendly
🐄 Relocate some of the functions into utils/mesos_tools 
🐄 Add test for relocated code

### DIFF
--- a/paasta_tools/contrib/kill_orphaned_docker_containers.py
+++ b/paasta_tools/contrib/kill_orphaned_docker_containers.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 import argparse
+import sys
 
 from paasta_tools import mesos_tools
 from paasta_tools.utils import get_docker_client
@@ -42,9 +43,10 @@ def main():
         if args.force:
             for container_name, mesos_task_id in orphaned_containers:
                 docker_client.kill(container_name)
+        sys.exit(1)
     else:
         print "OK: All mesos task IDs accounted for"
-
+        sys.exit(0)
 
 if __name__ == "__main__":
     main()
--- a/paasta_tools/contrib/kill_orphaned_docker_containers.py
+++ b/paasta_tools/contrib/kill_orphaned_docker_containers.py
@@ -22,7 +22,9 @@ def parse_args():
 def main():
     args = parse_args()
     docker_client = get_docker_client()
-    running_mesos_task_ids = mesos_tools.get_running_task_ids_from_mesos_slave()
+
+    running_mesos_task_ids = [task["id"] for task in mesos_tools.filter_running_tasks(
+        mesos_tools.get_running_tasks_from_active_frameworks(''))]
     running_mesos_docker_containers = get_running_mesos_docker_containers()
 
     orphaned_containers = []
--- a/paasta_tools/contrib/kill_orphaned_docker_containers.py
+++ b/paasta_tools/contrib/kill_orphaned_docker_containers.py
@@ -3,6 +3,7 @@ import argparse
 
 from paasta_tools import mesos_tools
 from paasta_tools.utils import get_docker_client
+from paasta_tools.utils import get_running_mesos_docker_containers
 
 
 def parse_args():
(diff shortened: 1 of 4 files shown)


paasta_tools/contrib/kill_orphaned_docker_containers.py
```diff
@@ -12,42 +13,35 @@ def parse_args():
...
-
-
 def main():
     args = parse_args()
     docker_client = get_docker_client()
-    running_mesos_task_ids = get_running_task_ids_from_mesos_slave()
-    running_mesos_docker_containers = get_running_mesos_docker_containers(docker_client)
-    print running_mesos_task_ids
+    running_mesos_task_ids = mesos_tools.get_running_task_ids_from_mesos_slave()
+    running_mesos_docker_containers = get_running_mesos_docker_containers()
```
mjksmith: Can `get_running_mesos_docker_containers` share the docker client that's made here?

paasta_tools/mesos_tools.py
```diff
@@ -137,6 +137,17 @@ def filter_not_running_tasks(tasks):
     return [task for task in tasks if task['state'] != 'TASK_RUNNING']
 
 
+def get_running_task_ids_from_mesos_slave():
```
mjksmith: AFAICT this is identical to `[task['id'] for task in get_running_tasks_from_active_frameworks('')]`.

nosmo: Nice! I was surprised there wasn't already a way in place to do this. Would it be an idea to make job_id default to `''` on any function where it can be used in this way in that case?

paasta_tools/mesos_tools.py
```diff
@@ -137,13 +137,13 @@ def filter_not_running_tasks(tasks):
...
 
-def get_running_tasks_from_active_frameworks(job_id):
+def get_running_tasks_from_active_frameworks(job_id=''):
     active_framework_tasks = get_current_tasks(job_id)
     running_tasks = filter_running_tasks(active_framework_tasks)
     return running_tasks
 
 
-def get_non_running_tasks_from_active_frameworks(job_id):
+def get_non_running_tasks_from_active_frameworks(job_id=''):
```
EvanKrall: what does it mean to have an empty string as job id?

paasta_tools/utils.py
```diff
@@ -1132,6 +1132,12 @@ def get_docker_client():
         return Client(base_url=get_docker_host(), **client_opts)
 
 
+def get_running_mesos_docker_containers():
+    client = get_docker_client()
```
EvanKrall: Why did you make this not a parameter when you moved this method from kill_orphaned_docker_containers.py?

paasta_tools/mesos_tools.py
nosmo: In this context, get all tasks from all jobs: https://github.com/Yelp/paasta/pull/352#discussion_r57098525

 It could use a docstring, which I'm happy to add.

paasta_tools/utils.py
nosmo: Nothing else seems to pass in a docker client as an argument anywhere else in utils.py, the one other function that interacts with docker in utils.py creates a client within the function itself (but does not use `get_docker_client`). Happy to move this but I figured I'd follow existing convention even if it's a single convention.

paasta_tools/mesos_tools.py
mjksmith: @EvanKrall mesos cli gets jobs by checking if `job_id in task['id']`, so empty string matches all tasks.

It's pretty hacky and based off of `mesos-cli` (`mesos_tools.get_current_tasks` calls [this method](https://github.com/mesosphere/mesos-cli/blob/master/mesos/cli/master.py#L199)) which in the long term might become `dcos-cli` so idk about the longevity of this.

paasta_tools/contrib/kill_orphaned_docker_containers.py
```diff
@@ -12,42 +13,37 @@ def parse_args():
...
+            orphaned_containers.append((container["Names"][0], mesos_task_id))
+
+    if orphaned_containers:
+        print "CRIT: Docker containers are orphaned: %s%s" % (", ".join(
+            "%s (%s)" % (container_name, mesos_task_id)
+            for container_name, mesos_task_id in orphaned_containers
+        ), " and will be killed" if args.force else "")
+        if args.force:
+            for container_name, mesos_task_id in orphaned_containers:
+                docker_client.kill(container_name)
```
solarkennedy: You need to exit non-0 (do 1) for things to properly be CRIT

paasta_tools/contrib/kill_orphaned_docker_containers.py
```diff
@@ -12,42 +13,37 @@ def parse_args():
...
+    else:
+        print "OK: All mesos task IDs accounted for"
```
solarkennedy: you can exit 0 here explicit for the OK if you want.

You can also use pysensu-yelp exits if you want to be fancy somehow
https://github.com/Yelp/pysensu-yelp/blob/master/pysensu_yelp/__init__.py#L14

paasta_tools/contrib/kill_orphaned_docker_containers.py
nosmo: According to pysensu this should be 2 for CRIT - should I be printing WARN, exiting with 2 or does it not matter?

solarkennedy:
Can you try merging master to see if the tests pass?

nosmo:
@solarkennedy when I test merging to master the test and the itests pass locally.

solarkennedy:
Heh:
```
Broadcast message from root@testing-gce-7f27a321-4311-4e05-999a-262065488062
	(unknown) at 12:09 ...
The system is going down for power off NOW!
```
There's your problem!

I've clicked on the travis button to rerun them.

solarkennedy:
ship

PR #330, opened 2016-03-16
Create a ``chronos_rerun`` script for rerunning past jobs

Rob-Johnson:

PR #354, opened 2016-03-23
Ensure cleanup jobs understands tmp jobs

### LINKED ISSUE #324: make `cleanup_chronos_jobs` aware of temporary jobs created by `paasta rerun`
falling out of #216 

we're going to be creating temporary jobs with ``paasta rerun`` - chronos_cleanup_jobs needs to know when it's ok to delete these

Rob-Johnson:
closes #324 

This makes ``cleanup_chronos_jobs`` aware of temporary jobs. It will only clean any temporary jobs up when:

- the job has completed (either successfully or unsuccessfully)
- the job completed > 24 hours ago.

By keeping the job around for 24 hours, the job will continue to be visible to users via ``paasta status`` once the run has completed (so service authors can inspect the job to find the current status)

### DIFF
--- a/paasta_itests/cleanup_chronos_jobs.feature
+++ b/paasta_itests/cleanup_chronos_jobs.feature
@@ -10,3 +10,13 @@ Feature: cleanup_chronos_jobs removes chronos jobs no longer in the config
      And the non-paasta jobs are not in the job list
      And the configured chronos jobs are in the job list
      And the unconfigured chronos jobs are not in the job list
+
+ Scenario: Cleanup chronos jobs ignores temporary jobs
+   Given a working paasta cluster
+     And we have yelpsoa-configs for the service "testservice" with enabled scheduled chronos instance "testinstance"
+     And we have a deployments.json for the service "testservice" with enabled instance "testinstance"
+    When we run chronos_rerun for service_instance testservice testinstance
+     And we store the name of the job for the service testservice and instance testinstance as myjob
+     And we run cleanup_chronos_jobs
+    Then we should get exit code 0
+     And we should see a job for the service "testservice" and instance "testinstance" in the job list
--- a/paasta_itests/cleanup_chronos_jobs.feature
+++ b/paasta_itests/cleanup_chronos_jobs.feature
@@ -20,3 +20,4 @@ Feature: cleanup_chronos_jobs removes chronos jobs no longer in the config
      And we run cleanup_chronos_jobs
     Then we should get exit code 0
      And we should see a job for the service "testservice" and instance "testinstance" in the job list
+     And there exists a job named "tmp testservice testinstance" in chronos
--- a/paasta_itests/cleanup_chronos_jobs.feature
+++ b/paasta_itests/cleanup_chronos_jobs.feature
@@ -20,4 +20,4 @@ Feature: cleanup_chronos_jobs removes chronos jobs no longer in the config
      And we run cleanup_chronos_jobs
     Then we should get exit code 0
      And we should see a job for the service "testservice" and instance "testinstance" in the job list
-     And there exists a job named "tmp testservice testinstance" in chronos
+     And there is a temporary job for the service testservice and instance testinstance
--- a/paasta_itests/steps/cleanup_chronos_job_steps.py
+++ b/paasta_itests/steps/cleanup_chronos_job_steps.py
@@ -86,7 +86,6 @@ def check_cleanup_chronos_jobs_output(context, expected_return_code):
     assert "Successfully Removed Tasks (if any were running) for:" in output
     assert "Successfully Removed Jobs:" in output
     for job in context.unconfigured_job_names:
-        print job
         assert '  %s' % job in output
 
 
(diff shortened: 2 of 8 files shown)


paasta_tools/chronos_tools.py
```diff
@@ -739,3 +739,20 @@ def get_job_for_service_instance(service, instance, include_disabled=True):
...
+    """
+    temporary_jobs = []
+    all_jobs = lookup_chronos_jobs(
+        client=client,
+        service=service,
+        instance=instance,
+        include_disabled=True,
+    )
+    for job in all_jobs:
+        if job['name'].startswith(TMP_JOB_IDENTIFIER):
```
mjksmith: Does this mean if a service author creates a service whose name starts with 'tmp' (for whatever reason) it won't get cleaned up? Maybe we should add a check to `paasta validate`/document this somewhere just in case.

Rob-Johnson: yeah, good idea.

Rob-Johnson: done in cb7bbc2

paasta_itests/cleanup_chronos_jobs.feature
```diff
@@ -10,3 +10,13 @@ Feature: cleanup_chronos_jobs removes chronos jobs no longer in the config
...
+
+ Scenario: Cleanup chronos jobs ignores temporary jobs
+   Given a working paasta cluster
+     And we have yelpsoa-configs for the service "testservice" with enabled scheduled chronos instance "testinstance"
+     And we have a deployments.json for the service "testservice" with enabled instance "testinstance"
+    When we run chronos_rerun for service_instance testservice testinstance
+     And we store the name of the job for the service testservice and instance testinstance as myjob
+     And we run cleanup_chronos_jobs
+    Then we should get exit code 0
+     And we should see a job for the service "testservice" and instance "testinstance" in the job list
```
EvanKrall: Might be worth making this look explicitly for the tmp job.

paasta_itests/steps/cleanup_chronos_job_steps.py
```diff
@@ -87,9 +86,17 @@ def check_cleanup_chronos_jobs_output(context, expected_return_code):
     assert "Successfully Removed Tasks (if any were running) for:" in output
     assert "Successfully Removed Jobs:" in output
     for job in context.unconfigured_job_names:
+        print job
```
EvanKrall: remove?

paasta_tools/chronos_tools.py
EvanKrall: Is there a way we could make it so the chronos job name for temporary jobs could never clash with names produced by other things? Then we could avoid that check.

paasta_itests/cleanup_chronos_jobs.feature
Rob-Johnson: done

paasta_tools/chronos_tools.py
Rob-Johnson: this is done in #330 - a timestamp is now added to the jobname to avoid conflicts

paasta_tools/cleanup_chronos_jobs.py
```diff
@@ -110,13 +143,17 @@ def main():
...
+                             chronos_tools.get_chronos_jobs_for_cluster(soa_dir=args.soa_dir)]
+
+    all_tmp_jobs = filter_tmp_jobs(filter_paasta_jobs(running_jobs))
+    expired_tmp_jobs = filter_expired_tmp_jobs(client, all_tmp_jobs)
 
-    running_service_jobs = set(deployed_job_names(client))
+    valid_tmp_jobs = [job for job in all_tmp_jobs if job not in expired_tmp_jobs]
 
-    to_delete = running_service_jobs - expected_service_jobs
+    to_delete = [job for job in running_jobs if job not in expected_service_jobs and job not in valid_tmp_jobs]
```
giuliano108: If:


```
running_jobs == expected_service_jobs + all_tmp_jobs + something_else_that_needs_removal
```

and:

```
all_tmp_jobs == valid_tmp_jobs + expired_tmp_jobs
```

any reason for not making everything a set and doing something like (arguably more readable):

```
valid_tmp_jobs = all_tmp_jobs - expired_tmp_jobs
to_delete = running_jobs - expected_service_jobs - valid_tmp_jobs
```

What's left should be: `something_else_that_needs_removal` + `expired_tmp_jobs`.

paasta_tools/cli/cmds/validate.py
```diff
@@ -203,6 +204,11 @@ def validate_chronos(service_path):
     chronos_spacer = paasta_tools.chronos_tools.INTERNAL_SPACER
 
     returncode = True
+
+    if service == TMP_JOB_IDENTIFIER:
```
giuliano108: I was expecting us to not allow `.startswith(TMP_JOB_IDENTIFIER)` (as opposed to the `==` you've got here)?

paasta_tools/cleanup_chronos_jobs.py
Rob-Johnson: heh,  yeah, but I'd argue the list comprehension is more readable?

paasta_tools/cli/cmds/validate.py
Rob-Johnson: yeah, I missed this. Good catch.

giuliano108:
ShipIt!

Issue #353, opened 2016-03-23
kazoo exception from mesos.cli


ronin13:
Noticed this during a timeout:

```
  Traceback (most recent call last):
  File "/usr/bin/setup_marathon_job", line 594, in main
    service_instance_config, soa_dir)
  File "/usr/bin/setup_marathon_job", line 514, in setup_service
    marathon_app_dict = service_marathon_config.format_marathon_app_dict()
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/paasta_tools/marathon_tools.py", line 379, in format_marathon_app_dict
    'constraints': self.get_constraints(service_namespace_config),
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/paasta_tools/marathon_tools.py", line 296, in get_constraints
    constraints.extend(self.get_routing_constraints(service_namespace_config))
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/paasta_tools/marathon_tools.py", line 304, in get_routing_constraints
    attribute=discover_level, blacklist=self.get_deploy_blacklist())
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/paasta_tools/mesos_tools.py", line 473, in get_mesos_slaves_grouped_by_attribute
    mesos_state = get_mesos_state_from_leader()
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/paasta_tools/mesos_tools.py", line 428, in get_mesos_state_from_leader
    state = master.CURRENT.state
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/mesos/cli/util.py", line 57, in __get__
    value = self.fget(inst)
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/mesos/cli/master.py", line 153, in state
    return self.fetch("/master/state.json").json()
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/mesos/cli/log.py", line 44, in timer
    return fn(*args, **kwargs)
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/mesos/cli/master.py", line 67, in fetch
    urlparse.urljoin(self.host, url),
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/mesos/cli/util.py", line 57, in __get__
    value = self.fget(inst)
  File "/usr/share/python/paasta-tools/lib/python2.7/site-packages/mesos/cli/master.py", line 61, in host
[...]

PR #355, opened 2016-03-23
Pin kazoo to 2.0.0

nhandler:
This related to #353 

This change downgrades us from kazoo 2.2 to 2.0.0 to work around a limitation in the unsupported mesos.cli. Once we fork/patch mesos.cli or swtich to dcos-cli, we can upgrade kazoo again.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,7 +16,7 @@ humanize==0.5.1
 importlib==1.0.3
 isodate==0.5.1
 jsonschema==2.5.1
-kazoo==2.2
+kazoo==2.0.0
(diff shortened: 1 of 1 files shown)


Rob-Johnson:
This didn't work - we're still seeing the issues reported in #353 (see the stack trace in internal ticket PAASTA-5245)

```
robj@mesos1-uswest1cdevc:~  % sudo /usr/share/python/paasta-tools/bin/python
>>> import pkg_resources
>>> pkg_resources.require("kazoo")[0].version
'2.2'
```

PR #360, opened 2016-03-25
fixed a bug where smartstack.yaml was overriding marathon.yaml

mjksmith:
I had this swapped when I added `deep_merge_dictionaries`.

I'd like this to be pushed out quickly because atm @gstarnberger is getting spammed with alerts.


nhandler:
Ship It!
Although, it might be worth considering renaming those arguments to minimize future confusion.

EvanKrall:
+1 to renaming; ship for now.

mjksmith:
I was thinking 'defaults' and 'overrides' but I'm up to options. I'll ship for now

PR #361, opened 2016-03-25
stopped alerting for disabled jobs on chronos

mjksmith:


Rob-Johnson:
I think it'd be better for you to filter out the disabled the jobs in ``main()`` on line 162, rather than have to do another lookup to check the disabled state for each job.

So it'd look something like:

```python
configured_jobs = chronos_tools.get_chronos_jobs_for_cluster(soa_dir=soa_dir)
disabled = [job for job in configured_jobs if job.get_disabled()]
enabled = [job for job in configured_jobs if job not in disabled]
for job in disabled:
   send_event(
     ....
     state=OK
     msg="Job is disabled - ignoring state"
   )
 service_job_mapping = build_service_job_mapping(client, enabled)
```

Doing it this way means you avoid polluting ``sensu_message_status_for_jobs`` with looking up disabled state, and avoid a second lookup for jobs.

What do you think?

mjksmith:
@Rob-Johnson `chronos_tools.get_chronos_jobs_for_cluster` doesn't return `ChronosJobConfig` objects, it just returns a list of pairs of `(service, instance)`.

Rob-Johnson:
shipit!

PR #362, opened 2016-03-25
Fix bintray upload

EvanKrall:
This fixes a few issues: For some reason, travis(dpl) didn't like my includePattern; it didn't try uploading anything. I've changed the pattern to be more generous, so I've made it so we don't drop dh-virtualenv*.deb into dist/ when running itest_trusty.

I also noticed I wasn't properly setting the date in bintray.json, so I fixed that.


nhandler:
Ship It!

PR #366, opened 2016-03-25
Only have travis build on master (and pull requests).

EvanKrall:
This should lighten the load on Travis and thus get us feedback sooner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,6 @@
+branches:
+  only:
+    - master
 language: python
 env:
   - TOXENV=py27


EvanKrall:
(see that this branch is not included on https://travis-ci.org/Yelp/paasta/branches)

nhandler:
Based on the fact that travis is only trying to run the `continuous-integration/travi-ci/pr` check, it looks like your change is fine.

Ship It!

mjksmith:
This is pretty sweet. I've always wondered why we check both branches and prs since we really only care about tests passing after a change is merged (aka what travis/pr checks)

nhandler:
The additional travis check was initially added so that you can iterate on a branch and get some test feedback as you go. However, for most of our branches, we open a PR shortly after pushing the branch to Github, so building all of the branches isn't necessary.

PR #368, opened 2016-03-28
made local run create a fresh deployments.json

mjksmith:
Before:

```
matts@dev2-devc:~/paasta (master) $ sudo python paasta_tools/cli/cli.py local-run -s example_service -c mesosstage -i mesosstage_main -p -y /nail/home/matts/yelpsoa-configs/
  File "paasta_tools/cli/cli.py", line 99, in <module>
    main()
  File "paasta_tools/cli/cli.py", line 95, in main
  File "/nail/home/matts/paasta/paasta_tools/cli/cmds/local_run.py", line 681, in paasta_local_run
    pull_image=pull_image,
  File "/nail/home/matts/paasta/paasta_tools/cli/cmds/local_run.py", line 592, in configure_and_run_docker_container
    soa_dir=args.yelpsoa_config_root,
  File "/nail/home/matts/paasta/paasta_tools/cli/utils.py", line 618, in get_instance_config
    soa_dir=soa_dir
  File "/nail/home/matts/paasta/paasta_tools/marathon_tools.py", line 150, in load_marathon_service_config
    deployments_json = load_deployments_json(service, soa_dir=soa_dir)
  File "/nail/home/matts/paasta/paasta_tools/utils.py", line 1192, in load_deployments_json
    raise NoDeploymentsAvailable
paasta_tools.utils.NoDeploymentsAvailable
```

After:

```
matts@dev2-devc:~/paasta (local_run_generate_deployments) $ sudo python paasta_tools/cli/cli.py local-run -s example_service -c mesosstage -i mesosstage_main -p -y /nail/home/matts/yelpsoa-configs/
Could not find deployment information for example_service in /nail/home/matts/yelpsoa-configs/.
local-run will attempt to generate it now.
Please wait while the image (docker-paasta.yelpcorp.com:443/services-example_service:paasta-3eaa22dff8a4963716c4903ff02054917d22a460) is pulled...
```

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -45,7 +45,6 @@ from paasta_tools.utils import get_docker_url
 from paasta_tools.utils import get_username
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import load_system_paasta_config
-from paasta_tools.utils import NoDeploymentsAvailable
 from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import PaastaNotConfiguredError
 from paasta_tools.utils import SystemPaastaConfig
@@ -588,26 +587,16 @@ def configure_and_run_docker_container(docker_client, docker_hash, service, inst
     soa_dir = args.yelpsoa_config_root
 
     volumes = list()
-    try:
-        instance_config = get_instance_config(
-            service=service,
-            instance=instance,
-            cluster=cluster,
-            load_deployments=pull_image,
-            soa_dir=soa_dir,
-        )
-    except NoDeploymentsAvailable:
-        sys.stdout.write(PaastaColors.yellow("Could not find deployment information for %s in %s.\n"
-                                             "local-run will attempt to generate it now.\n" % (
-                                                 service, soa_dir)))
+    if pull_image:
         generate_deployments_for_service(service, soa_dir)
-        instance_config = get_instance_config(
-            service=service,
-            instance=instance,
-            cluster=cluster,
-            load_deployments=pull_image,
-            soa_dir=soa_dir,
-        )
+
+    instance_config = get_instance_config(
+        service=service,
+        instance=instance,
+        cluster=cluster,
+        load_deployments=pull_image,
+        soa_dir=soa_dir,
+    )
 
     if pull_image:
         docker_url = get_docker_url(
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -35,6 +35,7 @@ from paasta_tools.cli.utils import guess_instance
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_instances
 from paasta_tools.cli.utils import list_services
+from paasta_tools.generate_deployments_for_service import generate_deployments_for_service
 from paasta_tools.marathon_tools import CONTAINER_PORT
 from paasta_tools.marathon_tools import get_healthcheck_for_instance
 from paasta_tools.paasta_execute_docker_command import execute_in_container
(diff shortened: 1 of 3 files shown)


mjksmith:
I guess doing it like this will mean `deployments.json` will only be generated once and then stay stale until it's manually deleted. Other options are deleting `deployments.json` after the instance config is loaded (and it was non-existant before), or always running `generate_deployments_for_service` in `local-run`.

EvanKrall:
Another option would be to have the deployments.json location not respect `-y`; maybe have it look at a different thing that happens to usually be the same as SOA_CONFIGS_ROOT?

mjksmith:
Do we want to only run `generate_deployments_for_service` when a non-default `soa_dir` is specified? It seems like a Yelpey assumption that the default `soa_dir` will always be up-to-date on dev machines.

mjksmith:
From internal JIRA:
Also, I observed that if we add a new instance in the marathon files, we need to provide that instance info entry in deployments.json file for it to locate the docker image for that instance.
Currently, I just copy one of similar docker_image ids for other instances on a particular region but not sure if that's the right way to do it.

This change will make this impossible by regenerating `deployments.json` every time `local-run` is called. If a service author is testing a change with `--pull` that they haven't marked for deployment yet, should they get an error? If we want  to give service authors the ability to override `generate_deployments_for_service` by specifying their own `deployments.json` maybe I'll just do what @EvanKrall suggested.

IMO running `local-run --pull` before marking a deploy group for deployment should print a helpful error about pushing a docker image and running mark for deployment.

EvanKrall:
Yeah, maybe we should just be spitting out a helpful error message that instructs people to copy deployments.json and/or run `paasta generate_deployments_for_service -y whatever`, instead of doing it automatically?

mjksmith:
I think that `local-run --pull` shouldn't work if a service hasn't been marked for deployment yet, and allowing developers to specify their on `deployments.json` goes against that. This way `--pull` gives a really strong indication of if a service has been pushed and marked correctly.

solarkennedy:
I second @EvanKrall 's suggestion of catching the the error and telling the user what they can run to generate deployments. (whether that is in their local checkout or in /etc/)

nhandler:
Ship It!

paasta_tools/cli/cmds/local_run.py
```diff
@@ -583,21 +585,36 @@ def configure_and_run_docker_container(docker_client, docker_hash, service, inst
...
+            service=service,
+            instance=instance,
+            cluster=cluster,
+            load_deployments=pull_image,
+            soa_dir=soa_dir,
+        )
+    except NoDeploymentsAvailable:
+        sys.stderr.write(PaastaColors.red(
+            "Error: No deployments.json found in %s/%s.\n"
+            "You can generate this by running paasta generate_deployments_for_service.\n" % (
```
solarkennedy: How about help a bit with 

```
You can generate this by running:
   generate_deployments_for_service -d %s -s %s 
(soa_dir, service)
```

paasta_tools/cli/cmds/local_run.py
```diff
@@ -583,21 +585,38 @@ def configure_and_run_docker_container(docker_client, docker_hash, service, inst
...
+                'soa_dir': soa_dir, 'service': service}))
+        return
 
+    if pull_image:
+        try:
+            docker_url = get_docker_url(
+                system_paasta_config.get_docker_registry(), instance_config.get_docker_image())
+        except NoDockerImageError:
+            sys.stderr.write(PaastaColors.red(
+                "Error: The deploy group %s for the service %s has not been marked for deployment.\n"
```
solarkennedy: Error: No sha has been marked for deployment for the %s deploy group.
Please ensure this service has either run through a jenkins pipeline or `paasta mark-for-deployment` has been run for $s

solarkennedy:
ship

PR #371, opened 2016-03-29
Use behave-pytest for pytest-style asserts

nhandler:
This switches us to use https://github.com/ribozz/behave-pytest , which will provide us with some more information when an assertion fails (actual and expected values). Some of the output will look a bit off (due to how we have color/formatting in our output), but it should still help with debugging test failures.


solarkennedy:
Neat! Ship!

PR #372, opened 2016-03-30
Only deploy to pypi on py27 build

EvanKrall:
This prevents pypi deployments from messing with dist/ on the itest_trusty build.

In testing my original bintray uploading branch, the pypi deployment wasn't running (since it only runs on tags), so my testing worked just fine. Once I merged, however, the uploads stopped working. It appears that the pypi development does something to the dist/ directory (cleans it out?), so if the pypi and bintray deployers both run on the same build, bintray won't work.

This should also prevent us from uploading the same package to pypi 6 times in one go.


nhandler:
From https://docs.travis-ci.com/user/deployment/pypi#Releasing-build-artifacts , it looks like Travis is designed to clean up any additional files/changes after it runs the tests and before the release.

You can set `skip_cleanup: true` to make Travis not clean up these artifacts.

Avoiding uploading the same package to pypi 6 times is definitely a good thing though. So ship it :)

PR #375, opened 2016-03-30
Upgrade to marathon 0.13.1 in itests

nhandler:
Apparently, our old version of marathon is trying to access newly renamed mesos endpoints. This issue (among other things) should be fixed by upgrading.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -21,6 +21,7 @@ import logging
 import os
 import re
 import socket
+from math import ceil
 from time import sleep
 
 import service_configuration_lib
@@ -255,7 +256,7 @@ class MarathonServiceConfig(InstanceConfig):
         if instances == 0:
             return 1
         else:
-            return float(10.0 / instances)
+            return int(ceil(10.0 / instances))
 
     def get_bounce_method(self):
         """Get the bounce method specified in the service's marathon configuration.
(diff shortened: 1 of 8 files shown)


Rob-Johnson:
note that this is a backwards incompatible change. I thought we were blocked on doing this by the client library?

paasta_tools/marathon_tools.py
```diff
@@ -255,7 +256,7 @@ def get_backoff_seconds(self):
         if instances == 0:
             return 1
         else:
-            return float(10.0 / instances)
```
solarkennedy: I'm concerned this will cause a big bounce, do we have to make this an int?

yelp_package/dockerfiles/itest/marathon/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN echo "debconf shared/accepted-oracle-license-v1-1 select true" | debconf-set
 RUN echo "debconf shared/accepted-oracle-license-v1-1 seen true" | debconf-set-selections
 RUN apt-get update && apt-get -y install lsb-release oracle-java8-installer
 
-RUN apt-get -y install libsasl2-modules marathon=0.11.1-1.0.432.ubuntu1404
+RUN apt-get -y install libsasl2-modules marathon=0.13.1-1.0.456.ubuntu1404
```
solarkennedy: Should be fine, this library is tested up to .14
https://github.com/thefactory/marathon-python/blob/master/.travis.yml

paasta_tools/marathon_tools.py
nhandler: Tests were failing due to the API wanting an integer instead of a float: https://github.com/mesosphere/marathon/blob/cf504d13e0dac9f6886ed9bcac0513cb5c6b7390/docs/docs/rest-api/public/api/v2/schema/AppDefinition.json#L45

solarkennedy: Alright, as long as we are aware that this will bounce everything.

Also I'm assuming that the existing versions of marathon we have in production can handle ints (seems like they should, line 257 is an int)

solarkennedy:
ship

solarkennedy:
I'll volunteer to pair with you to get this upgrade out

PR #377, opened 2016-03-31
Bump mesos.interface to 0.25.0

nhandler:
This is the final step in the mesos 0.25.0 upgrade I performed.


solarkennedy:
easy. ship if tests pass.

PR #379, opened 2016-03-31
Make the marathon cleanup script resolve all related sensu events.

solarkennedy:
Right now there are potentially 3 sensu alerts that could be thrown on a marathon service, this cleanup resolves them all.

### DIFF
--- a/tests/test_cleanup_marathon_jobs.py
+++ b/tests/test_cleanup_marathon_jobs.py
@@ -135,6 +135,7 @@ class TestCleanupMarathonJobs:
                 cluster='fake_cluster',
                 line=expected_log_line,
             )
+            assert mock_send_sensu_event.call_count == 3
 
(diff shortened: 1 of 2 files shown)


tests/test_cleanup_marathon_jobs.py
```diff
@@ -135,6 +135,7 @@ def test_delete_app(self):
                 cluster='fake_cluster',
                 line=expected_log_line,
             )
+            assert mock_send_sensu_event.call_count == 3
```
mittonk: Checking that `setup_marathon_job.123` is one of the calls would give more useful test-failure messages and be more robust if we have to add a 4th thing... but this works.

mittonk:
Looks good to me, I don't feel qualified to give a ship-it.

mjksmith:
Looks good to me.

Rob-Johnson:
shipit

PR #380, opened 2016-03-31
Remove docs about adding marathon-python to internal pypi

### LINKED ISSUE #378: Fix Upgrading Marathon Docs now that we use upstream marathon-python
http://paasta.readthedocs.org/en/latest/upgrading_marathon.html says to add the new marathon-python module to our internal pypi. We use the upstream version of this now, so this step is no longer necessary. This document should be updated to reflect this.

nhandler:
Fixes #378


solarkennedy:
ship

PR #381, opened 2016-03-31
Set real disabled value

Rob-Johnson:
This is *still* an issue - I was incorrect to close this with  #293

This is a dupe of #288, rebased onto current master.

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -479,12 +479,6 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
 
     resolved_disabled_state = determine_disabled_state(desired_state,
                                                        soa_disabled_state)
-
-    if resolved_disabled_state != soa_disabled_state:
-        log.info(("disabled state in deployments.json conflicts with state in"
-                  "soa_configs. setting disabled state to %r" %
-                  resolved_disabled_state))
-
     complete_config['disabled'] = resolved_disabled_state
 
     # we use the undocumented description field to store a hash of the chronos config.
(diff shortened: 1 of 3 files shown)


mittonk:
Seems reasonable.  :shipit:

paasta_tools/chronos_tools.py
```diff
@@ -471,14 +471,21 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
...
+
+    # resolve conflicts between the 'desired_state' and soa_configs disabled
+    # flag.
     desired_state = chronos_job_config.get_desired_state()
+    soa_disabled_state = complete_config['disabled']
+
+    resolved_disabled_state = determine_disabled_state(desired_state,
+                                                       soa_disabled_state)
+
+    if resolved_disabled_state != soa_disabled_state:
```
solarkennedy: We don't have to log this all the time, it will just add noise

paasta_tools/chronos_tools.py
```diff
@@ -488,6 +495,35 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
...
+
+    :param branch_state: the desired state on the branch - either start or
+    stop.
+    :param soa_disabled_state: the value of the disabled state in soa-configs.
+    """
+    if branch_state not in ['start', 'stop']:
+        raise ValueError('Expected branch_state to be start or stop')
+
+    if branch_state == 'start' and soa_disabled_state is True:
+        log.info("Job has been set to start, but config has disabled:True. Disabling job.")
```
solarkennedy: all this logging is really not necessary

paasta_tools/chronos_tools.py
```diff
@@ -471,14 +471,21 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
     )
 
     complete_config['name'] = compose_job_id(service, job_name)
+
+    # resolve conflicts between the 'desired_state' and soa_configs disabled
+    # flag.
     desired_state = chronos_job_config.get_desired_state()
```
solarkennedy: Why not put this logic in `get_desired_state` so that every caller of it will get it?

Rob-Johnson: I'd prefer to separate the presentation from the model - if you clobber ``get_desired_state``, there is no way you can access the 'real' value. I recently undid logic like this in #330 to ``get_cmd``

solarkennedy:
ship

PR #382, opened 2016-03-31
added a command to autoscale all services

mjksmith:
Also made autoscale_all_services grab a lock in zookeeper


solarkennedy:
ship

PR #394, opened 2016-04-04
Removed variable delay and other params from autoscaling

mjksmith:
Also updated docs to have more info on PID controllers and autoscaling.

### DIFF
--- a/docs/source/autoscaling.rst
+++ b/docs/source/autoscaling.rst
@@ -29,7 +29,7 @@ Let's look at sample marathon config file:
        metrics_provider: mesos_cpu_ram
        setpoint: 0.5
 
-This makes the service ``main`` autoscale using the ``pid`` decision policy and the ``mesos_cpu_ram`` metrics provider. Paasta will aim to keep this service's utilization at 50%.
+This makes the instance ``main`` autoscale using the ``pid`` decision policy and the ``mesos_cpu_ram`` metrics provider. Paasta will aim to keep this service's utilization at 50%.
 
 Autoscaling components
 ----------------------
--- a/docs/source/autoscaling.rst
+++ b/docs/source/autoscaling.rst
@@ -29,7 +29,7 @@ Let's look at sample marathon config file:
        metrics_provider: mesos_cpu_ram
        setpoint: 0.5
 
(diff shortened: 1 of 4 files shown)


docs/source/autoscaling.rst
```diff
@@ -53,12 +55,15 @@ Decision policies
 The currently available decicion policies are:
 
 :pid:
-  Uses a PID controller to determine when to autoscale a service.
+  Uses a PID controller to determine when to autoscale a service. See https://en.wikipedia.org/wiki/PID_controller for more information on PIDs.
```
solarkennedy: just use the normal rst hyperlink syntax

docs/source/autoscaling.rst
```diff
@@ -53,12 +54,14 @@ Decision policies
...
 
   Autoscaling parameters:
 
   :setpoint: the target utilization the controller aims for. Defaults to 0.8 (80%).
-  :delay: the number of seconds the decision policy must wait before fetching new data. Defaults to 600.
+  :Kp: the proportional control constant. Defaults to 0.2.
+  :Ki: the integral control constant. Defaults to 0.2.
```
solarkennedy: Are you sure there is a need to expose these to users? They seem really low level and I doubt anyone will really understand how to use them properly.

nhandler: Tend to agree with @solarkennedy . These controls are not something most services will need. If we find a workflow that requires such fine grain control, we probably want to talk to them to better understand their usecase rather than just having them adjust these settings.

mjksmith:
@solarkennedy @nhandler done. Some service authors wanted to be able to tune the PID constraints but let's see if they actually want that once autoscaling is released.

mjksmith:
Updated the title to better reflect the changes

docs/source/autoscaling.rst
```diff
@@ -27,9 +27,10 @@ Let's look at sample marathon config file:
      autoscaling:
        decision_policy: pid
        metrics_provider: mesos_cpu_ram
-       delay: 300
        setpoint: 0.5
 
+This makes the instance ``main`` autoscale using the ``pid`` decision policy and the ``mesos_cpu_ram`` metrics provider. Paasta will aim to keep this service's utilization at 50%.
```
nhandler: I know we have a few other instances where we say `Paasta`, but it is probably worth trying to be consistent and use `PaaSTA`.

solarkennedy:
ship

PR #395, opened 2016-04-05
make docker config location configurable

### LINKED ISSUE #6: location of dockercfg file is hardcoded
Falling out of #4 

The location of dockercfg, used by marathon/chronos to pull docker images from private registries, is hardcoded to /root/.dockercfg. This should be configurable.

ysyoung:
Closes #6

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -374,7 +374,7 @@ class ChronosJobConfig(InstanceConfig):
         else:
             return False, 'Your Chronos config specifies "%s", an unsupported parameter.' % param
 
-    def format_chronos_job_dict(self, docker_url, docker_volumes, system_paasta_config):
+    def format_chronos_job_dict(self, docker_url, docker_volumes, dockerfile_location):
         valid, error_msgs = self.validate()
         if not valid:
             raise InvalidChronosConfigError("\n".join(error_msgs))
@@ -387,7 +387,7 @@ class ChronosJobConfig(InstanceConfig):
                 'type': 'DOCKER',
                 'volumes': docker_volumes
             },
-            'uris': [system_paasta_config.get_dockerfile_location(), ],
+            'uris': [dockerfile_location, ],
             'environmentVariables': self.get_env(),
             'mem': self.get_mem(),
             'cpus': self.get_cpus(),
@@ -473,7 +473,7 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
     complete_config = chronos_job_config.format_chronos_job_dict(
         docker_url,
         docker_volumes,
-        system_paasta_config,
+        system_paasta_config.get_dockerfile_location(),
     )
 
     complete_config['name'] = compose_job_id(service, job_name)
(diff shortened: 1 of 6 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -265,6 +265,14 @@ def get_bounce_method(self):
         :returns: The bounce method specified in the config, or 'crossover' if not specified"""
         return self.config_dict.get('bounce_method', 'crossover')
 
+    def get_dockerfile_location(self):
+        """Get the URI for the .dockerfile.
+
+        :param service_config: The service instance's configuration dictionary.
+        :returns: The URI specified, or file:///root/.dockerfile if unspecified."""
+
+        return self.config_dict.get('dockerfile_location', 'file:///root/.dockercfg')
```
nhandler: Do we want to pull this path out into a constant somewhere like `DEFAULT_SOA_DIR`?

paasta_tools/utils.py
```diff
@@ -800,6 +800,13 @@ def get_sensu_port(self):
         """
         return int(self.get('sensu_port', 3030))
 
+    def get_dockerfile_location(self):
```
nhandler: This is identical to the function in `marathon_tools.py`. We should only have this in one file and then import it as necessary everywhere else.

tests/test_chronos_tools.py
```diff
@@ -912,7 +917,7 @@ def test_format_chronos_job_dict_invalid_param(self):
             branch_dict={},
         )
         with raises(chronos_tools.InvalidChronosConfigError) as exc:
-            invalid_config.format_chronos_job_dict('', [])
+            invalid_config.format_chronos_job_dict('', [], {})
```
nhandler: Can you use kwargs here? I have no clue what all of these empty strings/lists/dicts correspond to offhand.

paasta_tools/chronos_tools.py
```diff
@@ -374,7 +374,7 @@ def check(self, param):
         else:
             return False, 'Your Chronos config specifies "%s", an unsupported parameter.' % param
 
-    def format_chronos_job_dict(self, docker_url, docker_volumes):
+    def format_chronos_job_dict(self, docker_url, docker_volumes, system_paasta_config):
```
Rob-Johnson: can you make this the dockerfile location, rather than the system paasta config

paasta_tools/chronos_tools.py
```diff
@@ -473,6 +473,7 @@ def create_complete_config(service, job_name, soa_dir=DEFAULT_SOA_DIR):
     complete_config = chronos_job_config.format_chronos_job_dict(
         docker_url,
         docker_volumes,
+        system_paasta_config,
```
Rob-Johnson: make this ``system_paasta_config.get_dockerfile_location()``

tests/test_chronos_tools.py
solarkennedy: yes please
http://paasta.readthedocs.org/en/latest/style_guide.html#positional-args-versus-keyword-args

please make them all kwargs while you are at it. I understand that changing the other args is "technically" not related to this change, but it isn't that big of of a deal to make them consistent while you are in here.

paasta_tools/utils.py
ysyoung: fixed

paasta_tools/chronos_tools.py
ysyoung: fixed

paasta_tools/chronos_tools.py
ysyoung: fixed

PR #400, opened 2016-04-05
Make the marathon upgrade docs match current practice

nhandler:
Our marathon upgrade procedure has changed a bit since the last time we upgraded. This change updates our docs.


solarkennedy:
ship

PR #401, opened 2016-04-05
Make is_app_id_running tolerate leading slashes

solarkennedy:
This function needs to tolerate the situation where we could potentially give it an appid without a leading slash (like from a generated appid or a user provided one) or from the API which might return leading slashes. (the api is not consistent when it provides a leading slash or not.)


Rob-Johnson:
ship

PR #402, opened 2016-04-05
Add support for whitelisting.

### LINKED ISSUE #251: Add deploy_whitelist
Would be nice if we could have deploy_whitelist in marathon.yaml, like deploy_blacklist. We could express this with extra_constraints, except the default constraints have e.g. `["region", "GROUP_BY", # of regions]`, which needs to be aware of the number of regions that are whitelisted.

ysyoung:
Closes #251

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -161,6 +161,14 @@ instance MAY have:
 
    would indicate that PaaSTA should not deploy the service to the ``uswest1-prod`` region. By default the ``monitoring_blacklist`` will use the ``deploy_blacklist`` if it exists.
 
+  * ``deploy_whitelist``: A list of lists indicating a set of locations where deployment is allowed.  For example:
+
+      ``deploy_whitelist: [['region', 'uswest1-prod"]]``
+
+    would indicate that PaaSTA can **only** deploy in ``uswest1-prod``.  If this list is empty (the default), then
+    deployment is allowed anywhere.  This is superseded by the blacklist; if a host is both whitelisted and blacklisted,
+    the blacklist will take precedence.
+
   * ``monitoring_blacklist``: A list of lists indicating a set of locations to
     *not* monitor for Smartstack replication. For example:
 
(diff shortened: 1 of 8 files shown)


paasta_tools/utils.py
```diff
@@ -1312,6 +1318,20 @@ def deploy_blacklist_to_constraints(deploy_blacklist):
     return constraints
 
 
+def deploy_whitelist_to_constraints(deploy_whitelist):
+    """Convers a whitelist of locations into marathon appropriate constraints
```
EvanKrall: Converts

EvanKrall:
Can you write some tests for `slave_passes_whitelist` and `deploy_whitelist_to_constraints`? Would be good to test both the empty-whitelist and non-empty-whitelist case for each of those.

docs/source/yelpsoa_configs.rst
```diff
@@ -161,6 +161,15 @@ instance MAY have:
 
    would indicate that PaaSTA should not deploy the service to the ``uswest1-prod`` region. By default the ``monitoring_blacklist`` will use the ``deploy_blacklist`` if it exists.
 
+  * ``deploy_whitelist``: A list of lists indicating a set of locations where deployment is allowed.  For example:
```
EvanKrall: I wouldn't call it a list of lists anymore

docs/source/yelpsoa_configs.rst
```diff
@@ -161,6 +161,15 @@ instance MAY have:
...
+
+      ``deploy_whitelist: ['region', ['uswest1-prod", 'uswest2-prod]]``
+
+    would indicate that PaaSTA can **only** deploy in ``uswest1-prod`` or ``uswest2-prod``.  If this list
+    is empty (the default), then deployment is allowed anywhere.  This is superseded by the blacklist; if
+    a host is both whitelisted and blacklisted, the blacklist will take precedence.  Only one region-level
```
EvanKrall: I think the term we've been using is "location type", instead of "region-level"

EvanKrall:
fix'n'ship

PR #403, opened 2016-04-05
Upgrade tests to use marathon 0.14.2

nhandler:
With the mesos 0.26.0 upgrade underway, we can begin testing marathon 0.14.2 (which was tested against mesos 0.26.0).


solarkennedy:
ship

PR #404, opened 2016-04-06
added host networking

mjksmith:


docs/source/about/contract.rst
```diff
@@ -44,7 +44,8 @@ HTTP/TCP services
 -----------------
 
 * MUST be discoverable by SmartStack
-* MUST bind to port 8888
+* MUST bind to port 8888 if using the BRIDE `networking mode <../yelpsoa_configs.html#networking_mode>`_
```
solarkennedy: Can you make this more obvious this is an OR thing?

docs/source/yelpsoa_configs.rst
```diff
@@ -51,6 +51,13 @@ instance MAY have:
     namespace. For example ``canary`` instances can have ``nerve_ns: main`` to route
     their traffic to the same pool as the other ``main`` instances.
 
+  .. _networking_mode:
+
+  * ``networking_mode``: Specify which kind of
```
solarkennedy: I would rather this be called "net" and use lower case words like "host" or "none" or "bridge", just like docker does.


Then inside our codebase, as necessary, we use the symbols marathon needs.

paasta_tools/cli/schemas/chronos_schema.json
```diff
@@ -86,6 +86,10 @@
                 },
                 "uniqueItems": true
             },
+            "networking_mode": {
+                "enum": [ "BRIDGE", "HOST" ],
```
solarkennedy: I don't want this to be restrictive, as other users of this software might want to do custom named networking and stuff. Just say it is a string

docs/source/yelpsoa_configs.rst
mjksmith: @solarkennedy done. `'host'` and `'bridge'` become `'HOST'` and `'BRIDGE'` just before they're passed to marathon/chronos. If a service author gives a net string that's not none, host or bridge it just passes that string straight to marathon/chronos.

docs/source/yelpsoa_configs.rst
```diff
@@ -51,6 +51,12 @@ instance MAY have:
     namespace. For example ``canary`` instances can have ``nerve_ns: main`` to route
     their traffic to the same pool as the other ``main`` instances.
 
+  .. _net:
+
+  * ``net``: Specify which kind of
```
solarkennedy: please add a link to reference this in chronos as well. I know we don't currently use this much in chronos, but someone will probably want networking in chronos eventually

paasta_tools/chronos_tools.py
```diff
@@ -377,11 +377,19 @@ def format_chronos_job_dict(self, docker_url, docker_volumes, dockerfile_locatio
         if not valid:
             raise InvalidChronosConfigError("\n".join(error_msgs))
 
+        net = self.get_net()
+        docker_chronos_net_mapping = {
```
solarkennedy: Please consolidate these mappings in mesos_utils, this is actually a mapper to the mesos protobuf:
https://github.com/apache/mesos/blob/master/include/mesos/mesos.proto#L1689

docs/source/about/contract.rst
```diff
@@ -44,7 +44,7 @@ HTTP/TCP services
 -----------------
 
 * MUST be discoverable by SmartStack
-* MUST bind to port 8888
+* MUST bind to port 8888 if using the ``'bridge'`` `networking mode <../yelpsoa_configs.html#net>`_ or to port $MARATHON_PORT if using the ``'host'`` networking mode
```
solarkennedy: This deserves and extra warning

* MUST NOT bind on any other ports than ones specified by `$MARATHON_PORT*` in host mode otherwise they will conflict with other services

solarkennedy:
fix + ship

PR #408, opened 2016-04-06
Upgrade mesos to 0.28.0-2.0.16.ubuntu1404 in itests

nhandler:
This is the most recent release of Mesos, so after this upgrade, we will finally be up-to-date.


solarkennedy:
ship

Issue #373, opened 2016-03-30
Use docker containerizer in itests


nhandler:
When running PaaSTA, we set the mesos containerizer to `docker`. However, we do not specify a containerizer currently for our itests. This causes it to default to `mesos`. To ensure we are testing a setup like the one we actually run, we should update the itests to use the `docker` containerizer. This will also likely allow us to stop specifying `--launcher=posix` when running mesos.

EvanKrall:
This is somewhat tricky, as the tests run in a container. We'd need to volume-mount the docker socket or run [docker-in-docker](https://github.com/jpetazzo/dind).

mjksmith:
@EvanKrall The docker-in-docker people don't recommend you use it for CI. http://jpetazzo.github.io/2015/09/03/do-not-use-docker-in-docker-for-ci/
It looks like bind-mounting the docker socket is the way to go

matthewbentley:
Can I take a stab at this? It looks like we'd need to bind mount the docker socket into the mesos slaves, and tell mesos to use docker (in `paasta_itests/docker-compose.yml`, I think).

matthewbentley:
So when I make the following change:
```
diff --git a/paasta_itests/docker-compose.yml b/paasta_itests/docker-compose.yml
index a4a6935..b7e99fb 100644
--- a/paasta_itests/docker-compose.yml
+++ b/paasta_itests/docker-compose.yml
@@ -18,8 +18,13 @@ mesosslave:
     - zookeeper
   environment:
     -CLUSTER: testcluster
-  command: 'mesos-slave --master=zk://zookeeper:2181/mesos-testcluster --resources="cpus(*):10; mem(*):512; disk(*):100" --credential=/etc/mesos-slave-secret --launcher=posix'
+  command: 'mesos-slave --master=zk://zookeeper:2181/mesos-testcluster --resources="cpus(*):10; mem(*):512; disk(*):100" --credential=/etc/mesos-slave-secret --containerizers=docker --docker=/usr/bin/docker'
   hostname: mesosslave.test_hostname
+  privileged: true
+  volumes:
+    - /var/run/docker.sock:/var/run/docker.sock
+    - /usr/bin/docker:/usr/bin/docker

 marathon:
   build: ../yelp_package/dockerfiles/itest/marathon/
diff --git a/paasta_itests/steps/bounces_steps.py b/paasta_itests/steps/bounces_steps.py
index f3b51b5..1e4a654 100644
[...]

PR #411, opened 2016-04-07
Add memory swap param

### LINKED ISSUE #410: set the "memory-swap" parameter in docker containers.
Falling out of internal ticket PAASTA-3595.

By default, Docker sets the the value of ``memory.memsw.limit_in_bytes`` to the memory limit *2. This is undocumented, but can be seen [here](https://github.com/docker/docker/blob/master/daemon/daemon_unix.go#L292-L294).

We should fix this, ensuring that the memory limit set in soa-configs is memory+swap, rather than just memory so that containers are killed when they exceed that memory limit rather than swapping. We can do this by setting the ``--memory-swap`` parameter passed to the ``docker run`` invocation.

To do this in marathon and chronos, we can use the ``parameters`` argument available to pass to containers:

- [marathon](https://mesosphere.github.io/marathon/docs/native-docker.html#privileged-mode-and-arbitrary-docker-options)
- [chronos - undocumented, but in the code](https://github.com/mesos/chronos/blob/fc2567dbfb0567fd35a496be09a39474c3d0a7c6/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/Containers.scala#L30)

Unfortunately, this will cause a big bounce of applications.

Rob-Johnson:
closes #410 

I'm not sure the itests tell me much because of #373 ? I'm testing on mesosstage.

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -394,7 +394,7 @@ class ChronosJobConfig(InstanceConfig):
                 'type': 'DOCKER',
                 'volumes': docker_volumes,
                 'parameters': {
-                    'memory-swap': self.get_mem(),
+                    'memory-swap': "%sm" % self.get_mem(),
                 }
             },
             'uris': [dockerfile_location, ],
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -392,7 +392,10 @@ class ChronosJobConfig(InstanceConfig):
                 'image': docker_url,
                 'network': net,
                 'type': 'DOCKER',
-                'volumes': docker_volumes
+                'volumes': docker_volumes,
+                'parameters': {
+                    'memory-swap': self.get_mem(),
+                }
             },
             'uris': [dockerfile_location, ],
             'environmentVariables': self.get_env(),
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -380,7 +380,7 @@ class MarathonServiceConfig(InstanceConfig):
                         },
                     ],
                     "parameters": [
-                        {"key": "memory-swap", "value": "%sm" % str(float(self.get_mem()))},
+                        {"key": "memory-swap", "value": "%sm" % str(self.get_mem())},
                     ]
                 },
                 'type': 'DOCKER',
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -379,6 +379,9 @@ class MarathonServiceConfig(InstanceConfig):
                             'protocol': 'tcp',
                         },
                     ],
+                    "parameters": [
+                        {"key": "memory-swap", "value": float(self.get_mem())},
+                    ]
                 },
                 'type': 'DOCKER',
                 'volumes': docker_volumes,
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -380,7 +380,7 @@ class MarathonServiceConfig(InstanceConfig):
                         },
                     ],
                     "parameters": [
-                        {"key": "memory-swap", "value": float(self.get_mem())},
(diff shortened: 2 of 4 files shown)


paasta_tools/chronos_tools.py
```diff
@@ -392,7 +392,10 @@ def format_chronos_job_dict(self, docker_url, docker_volumes, dockerfile_locatio
                 'image': docker_url,
                 'network': net,
                 'type': 'DOCKER',
-                'volumes': docker_volumes
+                'volumes': docker_volumes,
+                'parameters': {
+                    'memory-swap': "%sm" % self.get_mem(),
```
solarkennedy: The docker man page says this "always" must be larger than -m, can you confirm it actually works if this is the same?


Also can you update the docs about "how paasta runs docker images to say that paasta doesn't allow swap?

solarkennedy:
I'm not afraid of the big bounce, just ship it slowly a couple clusters at a time like normal.

solarkennedy:
ship

paasta_tools/chronos_tools.py
Rob-Johnson: confirmed launching a marathon task - it works fine.

Rob-Johnson:
Chronos doesn't seem to do the right thing here:

here's the test job that was posted (Ive removed the volume mounts)

```
(py27)bash-4.1# ./paasta_tools/setup_chronos_job.py example_service.mesosstage_robj_test
Job to update: {'retries': 2, 'shell': True, 'container': {'image': u'docker-paasta.yelpcorp.com:443/services-example_service:paasta-4bacf539ce5184bd9fa1862db5afb9aad50703cc', 'parameters': {'memory-swap': '100m'}, 'type': 'DOCKER', 'network': 'BRIDGE', 'volumes': []}, 'name': 'example_service mesosstage_robj_test', 'schedule': 'R/2015-08-14T20:35:04+00:00/PT1M', 'mem': 100, 'epsilon': 'PT60S', 'scheduleTimeZone': None, 'description': 'config5812fbc7', 'cpus': 0.1, 'environmentVariables': [{'name': 'PAASTA_INSTANCE', 'value': 'mesosstage_robj_test'}, {'name': 'PAASTA_SERVICE', 'value': 'example_service'}, {'name': 'PAASTA_CLUSTER', 'value': u'mesosstage'}], 'disabled': False, 'command': 'sleep 2m', 'arguments': None, 'owner': 'noop', 'async': False, 'disk': 1024, 'uris': ['file:///root/.dockercfg'], 'constraints': [['pool', 'EQUALS', 'default']]}
Updated Chronos job: example_service mesosstage_robj_test
````

Note the ``'parameters': {'memory-swap': '100m'}``.

But the parameters never make it to the executor from what I can tell (the docker container doesn't have the memory-swap param set as it should) and querying the job via the API shows Chronos knows nothing of the params:

```
$ curl  -H 'Accept: application/json' 'http://chronos.paasta-mesosstage.yelp/scheduler/jobs'
```
screen shot 2016-04-12 at 17 08 24 1

File: paasta_tools/chronos_tools.py (hunk 1 above)
Comment (Rob-Johnson) [inline paasta_tools/chronos_tools.py]: also confirmed this works when the kernel cmdline option isn't set

```

robj@srv5-r6-sfo2:~  % cat /proc/cmdline
BOOT_IMAGE=/boot/vmlinuz-3.13.0-65-generic root=UUID=6e4297c5-c9a3-4bb3-bddc-0ecec13a2349 ro console=ttyS0,115200 console=tty0 elevator=noop
robj@srv5-r6-sfo2:~  % sudo docker run -it --memory 100m --memory-swap 100m ubuntu bash
Unable to find image 'ubuntu:latest' locally
latest: Pulling from library/ubuntu
87192bdbe00f: Already exists
28e09fddaacb: Already exists
7e15ce58ccb2: Already exists
a3ed95caeb02: Already exists
Digest: sha256:f6e8757419147ba099af8cec4db365b3603472bd6182722b16fdae932c0bf3bf
Status: Downloaded newer image for ubuntu:latest
WARNING: Your kernel does not support swap limit capabilities, memory limited without swap.
```

PR #413, opened 2016-04-07
Use json for mesos authentication

nhandler:
https://issues.apache.org/jira/browse/MESOS-2281 deprecates using plain text credentials for authentication in Mesos 0.29.0. The bug that forced us to move off of json credentials has been fixed. This PR switches us back to using json.


solarkennedy:
ship. I'm surprised that is the only docker file that needs these secrets?

nhandler:
For marathon and chronos, we specify `--mesos_authentication_principal` and `--mesos_authentication_secret_file` when we run them. The first argument includes the principal name. The second is a path to a file that only contains the password (nothing else).

PR #415, opened 2016-04-07
Upgrade marathon to 0.15.3-1.0.463.ubuntu1404 in itests

nhandler:
Now that we are running mesos 0.28.0, we can try and upgrade to marathon 0.15.3. This is the most recent release of marathon.


EvanKrall:
shiiiiip

PR #416, opened 2016-04-07
Bump mesos.interface to 0.28.0

nhandler:
We've upgraded to mesos 0.28.0. This bumps up mesos.interface in requirements.txt.


solarkennedy:
ship

PR #418, opened 2016-04-08
Revert "Merge branch 'dont-require-start-time'"

Rob-Johnson:
This reverts commit c8eaa31533407ff3400776b1abf1c611e192515d, reversing
changes made to c0733d2541c6a4d963ba6a344f42ab2e12fbd876.

This was done for good reason, as documented in our soa-configs:

>  Although Chronos supports an empty start time to indicate that the job should start immediately, we do not allow this. In a situation such as restarting Chronos, all jobs with empty start times would start simultaneously, causing serious performance degradation and ignoring the fact that the job may have just run.


solarkennedy:
ship

PR #422, opened 2016-04-08
Make the argparser print the full help whenever an argparse issue is encountered

solarkennedy:
The python arparser isn't very helpful when you ask for "help" or typo or a command or something like that, it simply prints the "usage" which is the

```
check,cook-image,emergency-restart,emergency-scale,emergency-start,emergency-stop,fsm,generate-pipeline,get-latest-deployment,info,itest,list,list-clusters,local-run,logs,mark-for-deployment,metastatus,performance-check,push-to-registry,rollback,security-check,start,restart,stop,status,validate
```
message. Not that helpful. This makes it so we print the full help page and respond to the "help" command.

### DIFF
--- a/paasta_tools/cli/cli.py
+++ b/paasta_tools/cli/cli.py
@@ -25,6 +25,18 @@ from paasta_tools.cli.utils import load_method
 from paasta_tools.cli.utils import modules_in_pkg as paasta_commands_dir
 
 
+class ThrowingArgumentParser(argparse.ArgumentParser):
+    """Overriding the error method allows us to print the whole help page,
+    otherwise the python arg parser prints a not-so-useful usage message that
+    is way too terse"""
+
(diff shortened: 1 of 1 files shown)


EvanKrall:
There's really not a way to do this besides subclassing?

Also can you paste the output of `paasta help` with this change?

solarkennedy:
With this change, `paasta help` will return the same thing as `paasta --help`: http://pastebin.com/7JTdVHJR

Yes, I believe this is the only way with python-argparse to prevent it from spitting out the terse "usage" text on error:
http://stackoverflow.com/questions/14728376/i-want-python-argparse-to-throw-an-exception-rather-than-usage

solarkennedy:
ping @EvanKrall

EvanKrall:
Shipit

PR #425, opened 2016-04-09
ignore temporary jobs when checking chronos jobs

### LINKED ISSUE #421: temporary jobs created by chronos-rerun trigger a sensu alert for check_chronos_jobs
```
Unknown: somehow there was more than one enabled job for example_service.mesosstage_robjreruntest. Talk to the PaaSTA team as this indicates a bug
```

This is because we don't filter the temporary jobs from the list to check.

Rob-Johnson:
closes #421 

The issue was a bit more of an edge case than I realised - it occurred to me that this wasn't consistent, and we weren't always being told about > 1 job existing when we had Chronos rerun jobs around.

What I found was:

- we filter **enabled** jobs when looking for chronos jobs belonging to a service_instance in ``check_chronos_jobs``
- when running ``chronos_rerun`` we launch a job with ``disabled: False``. If ``disabled: True``, then Chronos refuses to run the job at all. Instead, we control the 'run once' functionality using the repeat value in the schedule field (``R1/start-time/interval``)
- Chronos **disables** the job once the job has been run, and adjusts the schedule to have ``R0``
- As a result, the extra job only triggers the issue of > 1 job showing up for a service_isntance in ``check_chronos_jobs`` if ``check_chronos_jobs`` is run whilst the rerun job is still running.

### DIFF
--- a/paasta_tools/check_chronos_jobs.py
+++ b/paasta_tools/check_chronos_jobs.py
@@ -101,7 +101,8 @@ def build_service_job_mapping(client, configured_jobs):
             instance=job[1],
             client=client,
         )
-        with_states = last_run_state_for_jobs(matching_jobs)
+        filtered = chronos_tools.filter_non_temporary_chronos_jobs(matching_jobs)
+        with_states = last_run_state_for_jobs(filtered)
         service_job_mapping[job] = with_states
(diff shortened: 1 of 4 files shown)


solarkennedy:
ship

PR #431, opened 2016-04-15
removed portMapping from non-bridge nets

mjksmith:
I fixed this a while ago but it looks like it got re-included in a merge. Specifying `portMappings` with `--net=host` causes launching the docker container to fail.

The TF2 server will not go to space today ):


EvanKrall:
ship

PR #433, opened 2016-04-15
Use marathon 1.1.1 in itest and upgrade marathon-python

nhandler:
Marathon 1.1.1 has been released. Let's upgrade to it.

This will fail in Travis until @solarkennedy's [PR](https://github.com/thefactory/marathon-python/pull/92) is merged and a new version of marathon-python is released.


solarkennedy:
ship

PR #435, opened 2016-04-15
Don't alert teams if a service hasn't been deployed yet

solarkennedy:
I'm not really so sure we have to alert/page all the peoples just because something hasn't been deployed yet. These lines have mostly caused a bunch of pages to people who don't need to be paged for this.

Pretty much nobody cares when they are setting up a new service, so I think it would be better if we just "move along" as if it wasn't a big deal and exit 0.


solarkennedy:
From the survey, 47% are unsurprising that they get alerts and 36% are surprised.

Now just because that 47% is unsurprising, doesn't necessarily mean we can just make this change, and then set their expectations in the documentation?

nhandler:
In hindsight, we probably should have worded that question a bit differently. Just because they are (un)surprised that they received an alert doesn't mean they necessarily want to receive said alert or that they find it useful.

My vote would be to ship this change. If you are worried about the people who like this extra safeguard, maybe turn it into a configurable boolean that FPM sets.

PR #437, opened 2016-04-18
always round --memory-swap param up to closest mb

Rob-Johnson:
the value of the arg for ``--memory-swap`` is required to be >= the
``mem`` param passed to docker - it also doesn't accept float values,
so we round whatever value is set in the mem field up to the closest MB. 
The only reason we haven't been bitten by this is because none of our 
soa-configs specify a float value for mem, but having tested it, it definitely causes apps 
to fail.


solarkennedy:
ship

Issue #376, opened 2016-03-30
Maintenance mode


EvanKrall:
In order to enable auto-scaling (or manual scaling, or destructive maintenance like docker upgrades) of Paasta clusters, we need for tasks to be proactively moved off of boxes that will go down soon.

Here's my proposal:

1. Use [Mesos's Maintenance primitives](http://mesos.apache.org/documentation/latest/maintenance/) to mark nodes as going down. Even though [Marathon doesn't support them](https://github.com/mesosphere/marathon/issues/2414), I think this is the "right way" to indicate this status, and maybe Marathon will support them in the future. For the purposes of this issue, I'm going to call boxes in draining mode "at-risk", to avoid confusion with Paasta's task draining.
2. Have setup_marathon_job inspect the existing applications and count the number of tasks running on at-risk boxes, and scale up the application by that number. Once new tasks are healthy, drain at-risk tasks.
    1. Marathon may spin up those new tasks on an at-risk box. Since you can't modify constraints of running Marathon applications without Marathon restarting everything, we can't get Marathon to not do this unless something else eats up those resources (see 3.) However, this isn't a huge deal - the next time setup_marathon_job runs, we'll spin up more; eventually, at-risk boxes would be full and all new instances would have to go somewhere else.
3. To help with 2i, setup_marathon_job (or some other cron job) could try to make [dynamic reservations](https://mesos.apache.org/documentation/latest/reservation/) for all remaining resources on an at-risk box, reserving them for a bogus [role](https://mesos.apache.org/documentation/latest/roles/). We'd also need to unreserve these resources once the box is no longer at-risk.
4. Write helper scripts that can:
    1. mark a box as draining
    2. wait until it's fully drained
    3. undrain the box.
    4. A wrapper that does all 3 of the above and runs a specified commands, for doing e.g. docker restarts.

EvanKrall:
Step #3 of the above would prevent chronos from running new tasks on draining boxes; existing tasks would not be killed. That seems like sort of the ideal for chronos.

nhandler:
I played around with this a bit. I was able to schedule a maintenance (which didn't really do anything since the frameworks don't support it yet). I was then able to start the maintenance. This not only stopped all tasks on the slave, it stopped the entire mesos-slave process. As a result, when I sent the REST request indicating maintenance was over and instructing it to bring the host up again, it didn't do anything. I had to manually restart the mesos-slave process to get things back to normal.

solarkennedy:
[...]

PR #438, opened 2016-04-18
First pass at 'paasta maintenance' command

nhandler:
This is a first pass at a `paasta maintenace` command. I need to think a bit more about how to properly do itests in a useful way, but there are some unit tests.

Connect to #376

### DIFF
--- a/paasta_tools/paasta_maintenance.py
+++ b/paasta_tools/paasta_maintenance.py
@@ -1,0 +1,422 @@
+#!/usr/bin/env python
+# Copyright 2015 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+import datetime
+import json
+from socket import getfqdn
+from socket import gethostbyname
+
+from dateutil import tz
+from pytimeparse import timeparse
+from requests import Request
+from requests import Session
+from requests.exceptions import HTTPError
+
+from paasta_tools.mesos_tools import get_mesos_leader
+
+
+def parse_args():
+    """Parses the command line arguments passed to this script"""
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '-d', '--duration',
+        type=parse_timedelta,
+        default='1h',
+        help="Duration of the maintenance window. Any pytimeparse unit is supported.",
+    )
+    parser.add_argument(
+        '-s', '--start',
+        default=now(),
+        help="Time to start the maintenance window. Defaults to now.",
+    )
+    parser.add_argument(
+        'action',
+        choices=['drain', 'undrain', 'down', 'up', 'status'],
+        help="Action to perform on the speicifed hosts",
+    )
+    parser.add_argument(
+        'hostname',
+        nargs='*',
+        default=[getfqdn()],
+        help='Hostname(s) of machine(s) to start draining. '
+        'You can specify <hostname>|<ip> to avoid querying DNS to determine the corresponding IP.',
+    )
+    return parser.parse_args()
+
+
+def base_api():
+    """Helper function for making all API requests
+
+    :returns: a function that can be callecd to make a request
+    """
+    leader = get_mesos_leader()
+
+    def execute_request(method, endpoint, **kwargs):
+        url = "%s%s" % (leader, endpoint)
+        timeout = 15
+        s = Session()
+        s.auth = load_credentials()
+        req = Request(method, url, **kwargs)
+        prepared = s.prepare_request(req)
+        try:
+            resp = s.send(
+                prepared,
+                timeout=timeout,
+            )
+            resp.raise_for_status()
+            return resp
+        except HTTPError as e:
+            e.msg = "Error executing API request calling %s. Got error: %s" % (url, e.msg)
+            raise
+    return execute_request
+
+
+def master_api():
+    """Helper function for making API requests to the /master API endpoints
+
+    :returns: a function that can be called to make a request to /master
+    """
+    def execute_master_api_request(method, endpoint, **kwargs):
+        base_api_client = base_api()
+        return base_api_client(method, "/master%s" % endpoint, **kwargs)
+    return execute_master_api_request
+
+
+def maintenance_api():
+    """Helper function for making API requests to the /master/maintenance API endpoints
+
+    :returns: a function that can be called to make a request to /master/maintenance
+    """
+    def execute_schedule_api_request(method, endpoint, **kwargs):
+        master_api_client = master_api()
+        return master_api_client(method, "/maintenance%s" % endpoint, **kwargs)
+    return execute_schedule_api_request
+
+
+def get_schedule_client():
+    """Helper function for making API requests to the /master/maintenance/schedule API endpoints
+
+    :returns: a function that can be called to make a request to /master/maintenance/schedule
+    """
+    def execute_schedule_api_request(method, endpoint, **kwargs):
+        maintenance_api_client = maintenance_api()
+        return maintenance_api_client(method, "/schedule%s" % endpoint, **kwargs)
+    return execute_schedule_api_request
+
+
+def get_maintenance_schedule():
+    """Makes a GET request to the /master/maintenance API endpoint
+
+    :returns: a function that can be called to make a request to /master/maintenance
+    """
+    client_fn = get_schedule_client()
+    return client_fn(method="GET", endpoint="")
+
+
+def get_maintenance_status():
+    """Makes a GET request to the /master/maintenance/schedule/status API endpoint
+
+    :returns: a requests.response object representing the current maintenance status
+    """
+    client_fn = get_schedule_client()
+    return client_fn(method="GET", endpoint="/status")
+
+
+def get_hosts_with_state(state):
+    """Helper function to check the maintenance status and return all hosts
+    listed as being in a current state
+
+    :param state: State we are interested in ('down_machines' or 'draining_machines')
+    :returns: A list of hostnames in the specified state or an empty list if no machines
+    """
+    status = get_maintenance_status()
+    if not status or state not in status:
+        return []
+    return [machine['id']['hostname'] for machine in status[state]]
+
+
+def get_draining_hosts():
+    """Returns a list of hostnames that are marked as draining
+
+    :returns: a list of strings representing hostnames
+    """
+    return get_hosts_with_state(state='draining_machines')
+
+
+def get_down_hosts():
+    """Returns a list of hostnames that are marked as down
+
+    :returns: a list of strings representing hostnames
+    """
+    return get_hosts_with_state(state='down_machines')
+
+
+def is_host_draining(hostname=getfqdn()):
+    """Checks if the specified hostname is marked as draining
+
+    :param hostname: Hostname we want to check if draining (defaults to current host)
+    :returns: a boolean representing whether or not the specified hostname is draining
+    """
+    return hostname in get_draining_hosts()
+
+
+def is_host_down(hostname=getfqdn()):
+    """Checks if the specified hostname is marked as down
+
+    :param hostname: Hostname we want to check if down (defaults to current host)
+    :returns: a boolean representing whether or not the specified hostname is down
+    """
+    return hostname in get_down_hosts()
+
+
+def parse_timedelta(value):
+    """Return the delta in nanoseconds.
+    :param value: a string containing a time format supported by :mod:`pytimeparse`
+    :returns: an integer (or float) representing the specified delta in nanoseconds
+    """
+    error_msg = "'%s' is not a valid time expression" % value
+    try:
+        seconds = timeparse.timeparse(value)
+    except TypeError:
+        raise argparse.ArgumentTypeError(error_msg)
+    if not seconds:
+        raise argparse.ArgumentTypeError(error_msg)
+    return seconds_to_nanoseconds(seconds)
+
+
+def datetime_seconds_from_now(seconds):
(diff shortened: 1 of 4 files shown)


solarkennedy:
We don't usually put this kind of script as a subcommand of the "paasta" command. Mostly because the paasta super-command is very user-centric, and this not something a user would ever run. It makes assumptions that it can make API calls, has auth? and can directly connect to the mesos api, which none of the other paasta commands do.

Please start his off as like, paasta_maintenance? Like was your expectation would be that you could run this from anywhere, or from the local slave? Or from the master?

nhandler:
Ok, I have gone ahead and moved this to a `paasta_maintenance` script.

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,244 @@
...
+        return
+
+    if action != 'status' and not hostnames:
+        print "You must specify one or more hostnames"
+        return
+
+    try:
+        leader = get_mesos_leader()
+    except MesosMasterConnectionError:
+        print "Run this script from a mesos master!"
```
Rob-Johnson: This is surprising - I expected the workflow would be 'log onto a slave, mark as in maintenance' rather than 'discover where the master is, ssh in, mark a slave as down'. I'm not saying this isn't the right way, just not the way I immediately expected.

Rob-Johnson:
There are a couple of things that stand out here:

- If this is executed on the master, how will you drain the box? Will there be another process on each slave that checks whether it is in drain mode, and acts accordingly?

- There doesn't seem to be any error handling - requests will throw descriptive exceptions, but there is no logic to handle them here.

- There is an awful lot of repetition here. I'd encourage you try and refactor the api client to avoid this. 

Here's a start on something you could use - note that the final request execution logic is handled in one place, rather than *every* function, and a "client" of sorts is built for every HTTP resource. The final endpoint (like ``/status``)  is then just executing the client for a given resource. (Note I've not tested any of this code, so it's probably buggy - this is just an idea of what It might look like)

```
def base_api():
    leader = get_mesos_leader()
    def execute_request(method, endpoint, **kwargs):
        endpoint = "http://leader:PORT/%s" % endpoint
        s = Session()
        req = Request(method, endpoint, **kwargs)
        prepped = s.prepare_request(req)
        try:
            resp = s.send(prepped,
               stream=stream,
               verify=verify,
               proxies=proxies,
               cert=cert,
               timeout=timeout
            )
            resp.raise_for_status()
         except Exception as e:
            print("Error executing api request calling %s . Got response code %d with error %s" %(endpoint, resp.status_code, e)
    return execute_request


def master_api_client():
    def execute_master_api_request(method, endpoint)
        base_api_client = base_api()
        return base_api_client(method, "master/%s" % endpoint, **kwargs)
    return execute_master_api_request


def maintainance_api_client():
    def execute_schedule_api_request(method, endpoint)
        master_api_client = maintenance_api_client()
        return base_api_client(method, "schedule/%s" % endpoint, **kwargs)
    return execute_schedule_api_request


def get_schedule_client():
    def execute_schedule_api_request(method, endpoint)
        master_api_client = maintenance_api_client()
        return master_api_client(method, "schedule/%s" % endpoint, **kwargs)
    return execute_schedule_api_request


def get_maintenance_schedule():
    client_fn = get_schedule_client()
    return client_fn("GET", "")


def get_maintenance_status():
   client_fn = get_schedule_client()
   return client_fn("GET", "status")
```

paasta_tools/paasta_maintenance.py
nhandler: Part of the reason for this is that you can mark a host as being up/down without the box itself being available. i.e. when imaging some new PaaSTA slaves, we've been marking them as down so that they do not immediately register when they come up (we want to wait until Puppet is done running). If we had to SSH to the slave to do this, it would not be possible.

Rob-Johnson: ok, then how about a  ``--hostname`` option that defaults to the current host, rather than constraining this to *only* run on a master. Another use case for this might be 'log into a box, mark as down, drain, restart docker, mark as up', which becomes more of a pain if you have to perform some commands on the master and some on the slave. Is there any technical constraint to *only* allow it on the master, rather than either?

nhandler:
@Rob-Johnson I've attempted to refactor the code a bit based on your feedback. It could still probably use more error handling (I'm working on adding it), but I wanted to get this change up for more review.

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+def build_start_maintenance_payload(hostnames):
+    return get_machine_ids(hostnames)
+
+
+def get_machine_ids(hostnames):
+    machine_ids = []
+    for hostname in hostnames:
+        machine_id = dict()
```
Rob-Johnson: make this a named tuple rather than a dictionary.

```
from collections import namedtuple
MachineId = namedtuple('MachineId', 'hostname, ip')

machine_id = MachineId(hostname='foo', ip='bar')
x = machine_id.hostname
y = machine_id.bar
```

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,276 @@
...
+        # This is to allow specifying a hostname as "hostname|ipaddr"a
```
Rob-Johnson: why do we need this? are you that concerned about the overhead of a dns request?

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,276 @@
...
+    client_fn = get_schedule_client()
+    return client_fn(method="GET", endpoint="")
+
+
+def get_maintenance_status():
+    client_fn = get_schedule_client()
+    return client_fn(method="GET", endpoint="status")
+
+
+def timedelta_type(value):
```
Rob-Johnson: this isn't really a type, is it?

paasta_tools/paasta_maintenance.py
nhandler: I'm not worried at all about the overhead. We ran into a problem a while ago where we updated DNS and I marked a box as being 'down' in preparation for it being imaged. However, there was some error with the DNS change that required us to change where the hostname pointed. Once the revised DNS change was pushed, I had no easy way to mark the old host/ip mapping as being up (since the maintenance script was resolving the hostname to a different IP).

The other case where this is useful is if you want to mark a box as 'down' prior to it even existing in DNS.

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,276 @@
...
+        undrain(hostnames, start, duration)
+    elif action == 'down':
+        down(hostnames)
+    elif action == 'up':
+        up(hostnames)
+    elif action == 'status':
+        status()
+        schedule()
+    else:
+        print "Unknown Action: %s" % action
```
Rob-Johnson: raise a ValueError here

paasta_tools/paasta_maintenance.py
nhandler: The dictionary was used because this gets converted to JSON further down.

paasta_tools/paasta_maintenance.py
nhandler: We used something similar in https://github.com/Yelp/paasta/blob/b89717a110d1bbf0025f7cb6e254507e7846eb4a/paasta_tools/contrib/delete_old_marathon_deployments.py#L42 . It isn't an official type. But per https://docs.python.org/3/library/argparse.html#type , we can use this function with argparse to handle type-checking and type conversion at the argparse level.

Rob-Johnson: ah right, ok. heh, TIL!

ok, in that case... can you be more explicit? call it ``parse_timedelta`` or something. Also, you should probably catch the exception that'll be thrown by ``timeparse.timeparse(value)`` when this isn't legit and throw the correct ``argparse.ArgumentTypeError`` so that the output is formatted correctly by argparse

paasta_tools/paasta_maintenance.py
Rob-Johnson: man, that sounds like a bit of a corner case you're optimising for. If you insist on it, at least add some docs to the help text to tell people about it.

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,276 @@
...
+            machine_id['hostname'] = host
+            machine_id['ip'] = ip
+        else:
+            machine_id['hostname'] = hostname
+            machine_id['ip'] = gethostbyname(hostname)
+        machine_ids.append(machine_id)
+    return machine_ids
+
+
+def build_maintenance_schedule_payload(hostnames, start, duration, drain=True):
```
Rob-Johnson: docstrings please

paasta_tools/paasta_maintenance.py
Rob-Johnson: bump

Rob-Johnson:
please add docstrings to your functions for readability, too

paasta_tools/paasta_maintenance.py
nhandler: I really don't think this situation will be that uncommon. I am totally up for other methods of allowing the user to specify hostname/ip pairs rather than the style I have now, but I think having this functionality is necessary. I will add a note to the argparse help for this argument.

paasta_tools/paasta_maintenance.py
nhandler: That could work. I'll need to flip from mesos_tools.py:get_mesos_leader()  to using mesos.cli, but that shouldn't be too bad.

nhandler:
I think I've addressed all of the feedback. Our general_itests currently fail due to importing `paasta_tools.mesos_tools` but never using it (we need the side-effect of it setting an environment variable for `mesos.cli`). I plan to fix #498, which should resolve this issue.

nhandler:
#498 has been merged. I've updated this PR to use `paasta_tools.mesos_tools.get_mesos_leader` instead of using `mesos.cli` directly.

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,367 @@
...
+
+    return payload
+
+
+def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
+    """Loads the mesos-slave credentials from the specified file. These credentials will be used for all
+    maintenance API requests.
+    :param mesos_secrets: optional argument specifying the path to the file containing the mesos-slave credentials
+    """
+    with open(mesos_secrets) as data_file:
```
Rob-Johnson: need some handling a missing file here?

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,367 @@
...
+        data = json.load(data_file)
+    username = data['principal']
```
Rob-Johnson: key errors?

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,367 @@
...
+
+def drain(hostnames, start, duration):
+    """Schedules a maintenance window for the specified hosts and marks them as draining.
+    :param hostnames: a list of hostnames
+    :param start: the time to start the maintenance, represented as number of nanoseconds since the epoch
+    :param duration: length of the maintenance window, represented as number of nanoseconds since the epoch
+    """
+    payload = build_maintenance_schedule_payload(hostnames, start, duration, drain=True)
+    client_fn = get_schedule_client()
+    print client_fn(method="POST", endpoint="", data=json.dumps(payload)).text
```
Rob-Johnson: given that you're not raising the exception in base_api (line 79), I think you might get caught out here if there is an exception in the request, the client_fn won't return anything.

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,367 @@
...
+        undrain(hostnames, start, duration)
+    elif action == 'down':
+        down(hostnames)
+    elif action == 'up':
+        up(hostnames)
+    elif action == 'status':
+        status()
+        schedule()
+    else:
+        raise ValueError("Unknown Action: %s" % action)
```
Rob-Johnson: will you actually get here? you'll return on line 342 won't you?

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,367 @@
...
+    )
+    parser.add_argument(
+        'hostname',
+        nargs='*',
+        default=[getfqdn()],
+        help='Hostname(s) of machine(s) to start draining. '
+        'You can specify <hostname>|<ip> to avoid querying DNS to determine the corresponding IP.',
+    )
+    options = parser.parse_args()
+    return options
```
Rob-Johnson: nitpick, but why not just ``return parser.parse_args()`` ?

paasta_tools/paasta_maintenance.py
```diff
@@ -0,0 +1,367 @@
...
+
+    def execute_request(method, endpoint, **kwargs):
+        url = "%s%s" % (leader, endpoint)
+        timeout = 15
+        s = Session()
+        s.auth = load_credentials()
+        req = Request(method, url, **kwargs)
+        prepared = s.prepare_request(req)
+        try:
+            resp = s.send(prepared,
```
Rob-Johnson: indentation looks a bit odd here

Rob-Johnson:
you're still missing a lot of ``:returns:`` in your docstrings

paasta_tools/paasta_maintenance.py
nhandler: I believe this is actually valid and works. I'll move `prepared` to its own line, which will allow me to make the overall indentation look a bit more as you were probably expecting.

paasta_tools/paasta_maintenance.py
nhandler: I think I borrowed this from another script. I'll change it.

paasta_tools/paasta_maintenance.py
nhandler: Correct. I added the check at 342 to fail a bit earlier. after this else statement was already here. I'll remove it.

paasta_tools/paasta_maintenance.py
nhandler: I guess I'll update `base_api()` to re-raise the `HTTPError` with a friendlier message.

Rob-Johnson:
fix and ship

Rob-Johnson:
oh man, what's up with the giant diff? have you rebased or something?

PR #440, opened 2016-04-19
Make DOCKER_TAG check less strict

### LINKED ISSUE #439: paasta check for DOCKER_TAG is too strict
```Makefile
export DOCKER_TAG ?= test-$(USER)
```

```
$ paasta check -s www_pages | grep DOCKER_TAG
✗ The Makefile contains no reference to DOCKER_TAG. Make sure you
specify a DOCKER_TAG and that your itest tags your docker image with $DOCKER_TAG. http://paasta.readthedocs.org/en/latest/about/contract.html
```

EvanKrall:
it's also valid to have "export DOCKER_TAG ?= ...", etc.

Fixes #439

### DIFF
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -139,7 +139,7 @@ def makefile_has_a_tab(makefile_path):
 
 def makefile_has_docker_tag(makefile_path):
     contents = get_file_contents(makefile_path)
-    return re.search(r"^DOCKER_TAG\s*\?=", contents, re.MULTILINE) is not None
+    return re.search(r"DOCKER_TAG\s*\?=", contents, re.MULTILINE) is not None
 
 
 def makefile_check():


asottile:
:ship: :it:

paasta_tools/cli/cmds/check.py
```diff
@@ -139,7 +139,7 @@ def makefile_has_a_tab(makefile_path):
 
 def makefile_has_docker_tag(makefile_path):
     contents = get_file_contents(makefile_path)
-    return re.search(r"^DOCKER_TAG\s*\?=", contents, re.MULTILINE) is not None
+    return re.search(r"DOCKER_TAG\s*\?=", contents, re.MULTILINE) is not None
```
nhandler: Definitely an improvement. Do you think we should make this even less strict and not require `?=` to be used?

nhandler:
Ship It!

paasta_tools/cli/cmds/check.py
EvanKrall: To be honest, we'll never get this 100% right, which is why `paasta check` is only advisory. I don't care too much to try and make the "perfect regex".

But actually, `?=` is pretty important - paasta needs the makefile to respect DOCKER_TAG if it's already set.

PR #444, opened 2016-04-20
Made almost every cli tool take a soa_dir

mjksmith:


nhandler:
Cool, Ship It!

We have a mix of `-d` and `-y` for the short argument (due to collisions with other arguments). Think it is worth trying to make this more uniform?

paasta_tools/cli/cmds/metastatus.py
```diff
@@ -51,6 +52,13 @@ def add_subparser(subparsers):
         '-c', '--clusters',
         help=clusters_help,
     ).completer = lazy_choices_completer(list_clusters)
+    status_parser.add_argument(
+        '-d', '--soa-dir',
+        dest="soa_dir",
+        metavar="SOA_DIR",
```
solarkennedy: why do you need the metavar on all of these?

solarkennedy:
I agree shipit and I agree the consistency thing. We have grown up now and I think we are ready to get rid of the "-y" and stick with "-d". (my terrible grep doesn't reveal much conflict?)

But I wouldn't block a shipit with the consistency thing, that can be done in a second pass if you want.

paasta_tools/cli/cmds/metastatus.py
mjksmith: @solarkennedy I copied it from another cli command.

PR #451, opened 2016-04-25
interpolate datestrings when using local-run with a chronos job

Rob-Johnson:
closes internal ticket PAASTA-4142 - paasta local-run doesn't interpolate tron-style datestrings when running chronos jobs.

### DIFF
--- a/paasta_tools/cli/cmds/cook_image.py
+++ b/paasta_tools/cli/cmds/cook_image.py
@@ -77,6 +77,7 @@ def paasta_cook_image(args, service=None, soa_dir=None):
             service=service,
             loglevel='debug'
         )
+        print output
         if returncode != 0:
             _log(
                 service=service,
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/cook_image.py
```diff
@@ -77,6 +77,7 @@ def paasta_cook_image(args, service=None, soa_dir=None):
             service=service,
             loglevel='debug'
         )
+        print output
```
solarkennedy: Do you still want this? a debug log-level should already go to stdout.

Although, we should not really _log if we are doing a local-run if you ask me.

solarkennedy:
ship

paasta_tools/cli/cmds/cook_image.py
Rob-Johnson: ah, no, this was leftover from testing

PR #454, opened 2016-04-25
Make synapse port configurable via SystemPaastaConfig

EvanKrall:
Like it says. Stop hard-coding synapse port, and make it configurable through the system paasta config. It defaults to the current value, so ideally even without updating our system paasta config, nothing should break.

### DIFF
--- a/paasta_tools/bounce_lib.py
+++ b/paasta_tools/bounce_lib.py
@@ -29,7 +29,6 @@ from marathon.models import MarathonApp
 
 from paasta_tools.monitoring.replication_utils import \
     get_registered_marathon_tasks
-from paasta_tools.smartstack_tools import DEFAULT_SYNAPSE_PORT
 from paasta_tools.utils import compose_job_id
 from paasta_tools.utils import load_system_paasta_config
 
@@ -214,7 +213,7 @@ def kill_old_ids(old_ids, client):
             continue
 
 
-def get_happy_tasks(app, service, nerve_ns, min_task_uptime=None, check_haproxy=False):
+def get_happy_tasks(app, service, nerve_ns, system_paasta_config, min_task_uptime=None, check_haproxy=False):
     """Given a MarathonApp object, return the subset of tasks which are considered healthy.
     With the default options, this returns tasks where at least one of the defined Marathon healthchecks passes.
     For it to do anything interesting, set min_task_uptime or check_haproxy.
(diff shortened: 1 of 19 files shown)


paasta_tools/bounce_lib.py
```diff
@@ -214,7 +213,7 @@ def kill_old_ids(old_ids, client):
             continue
 
 
-def get_happy_tasks(app, service, nerve_ns, min_task_uptime=None, check_haproxy=False):
+def get_happy_tasks(app, service, nerve_ns, system_paasta_config, min_task_uptime=None, check_haproxy=False):
```
solarkennedy: How would you feel if this function took synapse_port instead of system_paasta config?

paasta_tools/monitoring/check_classic_service_replication.py
```diff
@@ -181,7 +180,12 @@ def get_service_replication(self, all_services):
     def run(self):
         self.setup_logging()
         all_service_config = read_services_configuration()
-        service_replication = self.get_service_replication(all_service_config.keys())
+        system_config = load_system_paasta_config()
+        service_replication = self.get_service_replication(
+            all_service_config.keys(),
```
solarkennedy: can you kwargs this?

paasta_tools/utils.py
```diff
@@ -827,10 +829,29 @@ def get_sensu_port(self):
...
+        :returns: the haproxy-synapse status port."""
+        return int(self.get('synapse_port', 3212))
+
+    def get_default_synapse_host(self):
+        """Get the default host we should interrogate for haproxy-synapse state.
+
+        :returns: A hostname that is running haproxy-synapse."""
+        return self.get('synapse_host', 'localhost')
+
+    def get_synapse_haproxy_url_format(self):
```
solarkennedy: I don't see this function used?

solarkennedy:
ship

paasta_tools/bounce_lib.py
EvanKrall: I originally had that, but decided against it. `get_happy_tasks` already does a lot of things (looks at marathon, synapse, etc.), and conceivably could do more in the future. If it ever needs to look at some other system config parameter, I'd rather not add yet another argument.

The counterargument would be: What if the synapse port that `get_happy_tasks` looked at depended on context? (e.g. we ran multiple different synapses for different purposes or something). I'm not sure I ever see that happening.

I'm starting to think that our convention for system paasta configs would look like:

 - Main functions or other "entry points", like `deploy_service`, would call `load_system_paasta_config`.
 - Intermediate functions (like `get_happy_tasks`) would generally have `system_paasta_config` as an argument
 - Single-purpose utility functions (like `get_registered_marathon_tasks`) would not take `system_paasta_config`, and instead would take specific parameters, like `synapse_port`.

paasta_tools/utils.py
EvanKrall: I've made a bunch of changes so that this gets called properly.

Can you take a look again and make sure they look sane?

PR #467, opened 2016-04-27
Switch from readthedocs.org->readthedocs.io

nhandler:
Read the Docs just sent out an email informing users that they are moving project subdomains from readthedocs.org to readthedocs.io. They will automatically forward users from the old domain to the new one, but they are encouraging projects to update any links they might have.


solarkennedy:
ship

PR #470, opened 2016-04-28
Document which config entries invoke a bounce

solarkennedy:
cc @mittonk

internal ticket PAASTA-4400

I'm also trying out this autorun thing so we don't have to "remember" to keep docs in sync with code.


solarkennedy:
Fixes #474

Rob-Johnson:
the autorun thing is cool. ty for doing this, shipit!

nhandler:
Assuming the generated docs turned out as expected, Ship It!

mittonk:
Looks good to me in this view, but I saw a missing quote in a rendered view at some point in this process.

solarkennedy:
fixed

PR #471, opened 2016-04-28
Use cookiecutter instead of home-grown templates

solarkennedy:


solarkennedy:
Hmm. I didn't realize we had itests for this thing.....

Rob-Johnson:
shipit

PR #478, opened 2016-04-29
Greatly improve the output of check_marathon_services replication

solarkennedy:
This adds a bit more whitespace to the chronos checks to be kinda markdown-style.

Also adds lots more action items in the marathon checks, and tests!

### DIFF
--- a/paasta_tools/check_marathon_services_replication.py
+++ b/paasta_tools/check_marathon_services_replication.py
@@ -175,11 +175,13 @@ def check_smartstack_replication_for_instance(
                 "\n"
                 "Things you can do:\n"
                 "\n"
-                "  * Widen SmartStack discovery settings\n"
-                "  * Increase the instance count\n"
                 "  * Fix the cause of the unhealthy service. Try running:\n"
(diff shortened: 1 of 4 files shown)


paasta_tools/check_marathon_services_replication.py
```diff
@@ -158,6 +158,33 @@ def check_smartstack_replication_for_instance(
...
+                "\n"
+                "Reasons this might be happening:\n"
+                "\n"
+                "  The service may simply not have enough copies or it could simply be\n"
+                "  unhealthy in that location. There also may not be enough resources\n"
+                "  in the cluster to support the requested instance count.\n"
+                "\n"
+                "Things you can do:\n"
+                "\n"
+                "  * Widen SmartStack discovery settings\n"
```
Rob-Johnson: I'd move this and 'increase the instance count' to a 'if your service isn't broken' section - this isn't the first thing to do, I wouldn't want people reading this output as an order of operation.

Rob-Johnson:
ship

PR #480, opened 2016-04-29
PAASTA-4182 enable paasta_serviceinit status with multiple service instances

huadongliu:
This patch enables paasta_serviceinit status to process multiple instance in one ssh session. For each instance, "instance:instance_name" is added to the output so that the client can parse accordingly.

The options added are --service and --instances.

$ sudo .tox/py27/bin/paasta_serviceinit.py -s robj-test-service -i main,memo-test status

The argument service_instance is now optional to keep backward compatibility. It will be replaced eventually when the client code is updated. At that time, service and instances will become mandatory arguments.
 
(py27)$ .tox/py27/bin/paasta_serviceinit.py
usage: paasta_serviceinit.py [-h] [-v] [-D] [-d SOA_DIR] [-s SERVICE]
                             [-i INSTANCES] [-a APP_ID] [--delta DELTA]
                             [service_instance]
                             {start,stop,restart,status,scale}

The patch is manually tested with multiple services with robj-test-service and instances main/memo-test. In addition, an itest test case is added to cover both the good (being deployed) and bad (not being deployed) service instances.

### DIFF
--- a/paasta_itests/paasta_serviceinit.feature
+++ b/paasta_itests/paasta_serviceinit.feature
@@ -43,14 +43,7 @@ Feature: paasta_serviceinit
      When we run the marathon app "test-service.main"
       And we wait for it to be deployed
      Then paasta_serviceinit status -vv for the service_instance "test-service.main" exits with return code 0 and the correct output
-
-  Scenario: paasta_serviceinit can run status -s service -i instances
-    Given a working paasta cluster
-      And I have yelpsoa-configs for the marathon job "test-service.main"
-      And we have a deployments.json for the service "test-service" with enabled instance "main"
-     When we run the marathon app "test-service.main"
-      And we wait for it to be deployed
-     Then paasta_serviceinit status -s "test-service" -i "main,test" has the correct output for instance main and exits with non-zero return code for instance test
+      And paasta_serviceinit status -s "test-service" -i "main,test" has the correct output for instance main and exits with non-zero return code for instance test
 
   Scenario: paasta_serviceinit can run emergency-stop on an enabled chronos job
     Given a working paasta cluster
--- a/paasta_itests/paasta_serviceinit.feature
+++ b/paasta_itests/paasta_serviceinit.feature
@@ -43,6 +43,7 @@ Feature: paasta_serviceinit
      When we run the marathon app "test-service.main"
       And we wait for it to be deployed
      Then paasta_serviceinit status -vv for the service_instance "test-service.main" exits with return code 0 and the correct output
+      And paasta_serviceinit status -s "test-service" -i "main" exits with return code 0 and the correct output
       And paasta_serviceinit status -s "test-service" -i "main,test" has the correct output for instance main and exits with non-zero return code for instance test
 
   Scenario: paasta_serviceinit can run emergency-stop on an enabled chronos job
--- a/paasta_itests/paasta_serviceinit.feature
+++ b/paasta_itests/paasta_serviceinit.feature
@@ -44,6 +44,14 @@ Feature: paasta_serviceinit
       And we wait for it to be deployed
      Then paasta_serviceinit status -vv for the service_instance "test-service.main" exits with return code 0 and the correct output
 
+  Scenario: paasta_serviceinit can run status -s service -i instances
+    Given a working paasta cluster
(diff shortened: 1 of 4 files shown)


paasta_itests/paasta_serviceinit.feature
```diff
@@ -44,6 +44,14 @@ Feature: paasta_serviceinit
       And we wait for it to be deployed
      Then paasta_serviceinit status -vv for the service_instance "test-service.main" exits with return code 0 and the correct output
 
+  Scenario: paasta_serviceinit can run status -s service -i instances
+    Given a working paasta cluster
+      And I have yelpsoa-configs for the marathon job "test-service.main"
+      And we have a deployments.json for the service "test-service" with enabled instance "main"
+     When we run the marathon app "test-service.main"
+      And we wait for it to be deployed
+     Then paasta_serviceinit status -s "test-service" -i "main,test" has the correct output for instance main and exits with non-zero return code for instance test
```
solarkennedy: How about just put this on line 46? It looks like we don't need the whole scenario

paasta_tools/paasta_serviceinit.py
```diff
@@ -12,7 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Usage: ./paasta_servceinit.py [-v] <servicename> <stop|start|restart|status|scale>
+"""Usage: ./paasta_servceinit.py [-v] [service_instance] [-s service] [-i instances] <stop|start|restart|status|scale>
```
solarkennedy: just junk this, no need to keep this up to date, it is what -h is for

paasta_tools/paasta_serviceinit.py
```diff
@@ -64,39 +72,52 @@ def main():
...
+        instances = args.instances.split(',')
     else:
-        log.error("I calculated an instance_type of %s for %s which I don't know how to handle. Exiting."
-                  % (instance_type, compose_job_id(service, instance)))
+        log.error("The name of service or the name of instance to inspect is missing. Exiting.")
         sys.exit(1)
 
+    cluster = load_system_paasta_config().get_cluster()
+    for instance in instances:
+        print("instance:%s" % instance)
```
solarkennedy: How about 


    log.debug("Looking at service.instance")

paasta_tools/paasta_serviceinit.py
```diff
@@ -64,39 +72,52 @@ def main():
...
+                cluster=cluster,
+                verbose=args.verbose,
+                soa_dir=args.soa_dir,
+            )
+        else:
+            log.error("I calculated an instance_type of %s for %s which I don't know how to handle. Exiting."
+                      % (instance_type, compose_job_id(service, instance)))
+            return_code = 1
+
+        if return_code:
```
solarkennedy: This is a bit odd because we are just overwriting `return_code` in the loop.

Can you return the max() of the return codes? I'm pretty sure there are other places in our code where we do this...

paasta_tools/paasta_serviceinit.py
```diff
@@ -64,39 +72,52 @@ def main():
...
+            sys.exit(return_code)
```
solarkennedy: I don't think we should bomb out so early in the loop.

paasta_itests/paasta_serviceinit.feature
```diff
@@ -43,6 +43,7 @@ Feature: paasta_serviceinit
      When we run the marathon app "test-service.main"
       And we wait for it to be deployed
      Then paasta_serviceinit status -vv for the service_instance "test-service.main" exits with return code 0 and the correct output
+      And paasta_serviceinit status -s "test-service" -i "main,test" has the correct output for instance main and exits with non-zero return code for instance test
```
solarkennedy: Can you add a test above to match the "good" case? Cause a python stack trace will also cause a non-zero exit code:(

    And paasta_serviceinit status -s "test-service" -i "main" has the correct output for instance main and exits with 0 and the correct output

solarkennedy:
fix + ship

solarkennedy:
shipit

PR #481, opened 2016-05-02
Default to using the max instance count when autoscaling without decision data

solarkennedy:
cc @mjksmith @heathvini

Defaulting to max_instances is probably a safer setting than min for production services with no data at all in zk. This is also true if we have a catastrophic failure of zk.

Originally I thought I might need to edit the PID logic, but I think this achieves the desired effect?


mjksmith:
Looks good to me. You don't have to edit the PID

heathvini:
Looks good, hopefully this doesn't just delay the issue to when the PID logic first does its computation with non-missing data.

PR #482, opened 2016-05-02
Added a cli entrypoint to autoscale_cluster

solarkennedy:
This exposes the autoscale_cluster code to a cli command to make it easier to iterate on, also logging.

### DIFF
--- a/paasta_tools/autoscale_cluster.py
+++ b/paasta_tools/autoscale_cluster.py
@@ -1,0 +1,44 @@
+#!/usr/bin/env python
+# Copyright 2015 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+import logging
+
+from paasta_tools.autoscaling_lib import autoscale_local_cluster
+
+
+log = logging.getLogger('__main__')
+logging.basicConfig()
+
+
+def parse_args():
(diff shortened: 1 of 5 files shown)


paasta_tools/autoscale_cluster.py
```diff
@@ -0,0 +1,44 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+import logging
+
+from paasta_tools.autoscaling_lib import autoscale_local_cluster
+
+
+log = logging.getLogger('__main__')
```
Rob-Johnson: Hm, I don't think this is the 'right' way to do logging.

I think you should have:
```
```

Then in ``autoscaling_lib.py`` have:

```
log = logging.getLogger(__name__).addHandler(logging.NullHandler())
```

That way, the logging in ``autoscaling_lib.py`` will respect the logging configuration of whatever is using the library (the use of the NullHandler means that if logging hasn't been setup, then it won't spew unwanted output, and won't complain about missing handlers), and you'll see the correct source of the log line in the output

solarkennedy: Hmm. We don't do this Null trick with other included libraries (marathon_tools), should we?

But when I do add the null handler, when I try to log from inside the library, I get this:

    AttributeError: 'NoneType' object has no attribute 'debug'

Rob-Johnson: ah, yeah, it's because the ``addHandler`` function returns None.  try this:

```

log.debug('foo')
```
I think we probably should do it everywhere that is a library, to protect against the caller not configuring logging - [the docs](https://docs.python.org/2/howto/logging.html#configuring-logging-for-a-library) mention that a message will be printed to stderr if they haven't and the library tries to log.

solarkennedy: Gotcha. It still shows up as "main", I guess because of the way I included the library?

Let me ship what I've got and I'll make a second pr for changing getlogger for all the things.

Rob-Johnson:
even when you swap the '__main__' for __name__ ?

solarkennedy:
Oh, duh of course, but when I do *that* then the log level is not set so I don't see any of the debug logs for any lines in the lib. :(

PR #483, opened 2016-05-02
Make the FQDN format configurable; previously hard-coded as paasta-%s.yelp

EvanKrall:
This makes it so you don't have to use the fake ".yelp" TLD for discovering where paasta masters are.


solarkennedy:
ship

PR #485, opened 2016-05-03
Remove the 'driver' keyword from log_writer options before passing it as kwargs to the log writer class, in case the log writer class doesn't take driver or **kwargs

EvanKrall:


paasta_tools/utils.py
```diff
@@ -526,6 +526,7 @@ def configure_log():
     log_writer_options = load_system_paasta_config().get_log_writer()
     global _log_writer
     LogWriterClass = get_log_writer_class(log_writer_options['driver'])
+    del log_writer_options['driver']  # In case the log writer class doesn't take **kwargs or a 'driver' parameter.
```
solarkennedy: I don't understand , what the class *does* need the driver?

Maybe the configuration would be easier to use if it was like this?

```
log_writer:
  driver: scribe
  options: 
    foo: bar
``

Then this code might be more natural:

```
log_writer = load.get_log_writer()
LWClass  = log_writer['driver']
lw_options = log_writer['options']

_log_writer = LWClass[**lw_options]
```

And make it part of the contract that these plugins must take kwargs and and it is up to the admin to make sure they only add options the writer can accept?

Comment (EvanKrall) [inline paasta_tools/utils.py]: That's probably better, it's just a breaking change to config parsing.

Comment (EvanKrall) [inline paasta_tools/utils.py]: I'm also thinking about allowing multiple log writers, so your config would look something like:

```
log_writer:
  scribe:
    port: 3535
  file:
    path_format: /var/log/paasta.log
  stdout: {}
```

Or maybe:

```
log_writer:
  - scribe:
      port: 3535
  - file:
      path_format: /var/log/paasta.log
  - file:
      path_format: /var/log/paasta/{service}.{instance}.log
```
or

```
log_writer:
  - driver: scribe
    options:
      port: 3535
  - driver: file
    options:
      path_format: /var/log/paasta.log
```

solarkennedy: I'm down for whatever, ship.

PR #489, opened 2016-05-04
PAASTA-4412 monkey patch MesosMaster.state to increase caching TTL

huadongliu:
With this patch, mesos master state fetches are consistently under a small fraction of a second except for the 1st fetch. requests_cache code is not touched because requests_cache alone produced fetch latencies above one second.


Rob-Johnson:
shipit

solarkennedy:
ship

PR #491, opened 2016-05-04
Be more consistent about how we log to stderr

### LINKED ISSUE #490: Unify logging behavior
From #482, we setup logging in `main` situations, set log levels, etc.

In our lib classes we also setup logging (sometimes) and use the same `__main__` logger, so it gets the same log level and such.

@Rob-Johnson recommends adding a null handler in the library classes so they can be used as-is, and allowing them to use `__name__` so that the log entry shows up with the "correct" class name.

@Rob-Johnson I'm trying to make it so the class "inherits" the same log level that was set in `main`, but it isn't working.

```
# Lib class
mainlog = logging.getLogger('__main__')
# Prints 30 (warn) all the time, because I think that is what it is set at import time
print mainlog.getEffectiveLevel()
log.setLevel(mainlog.getEffectiveLevel())
```

This seems a bit harder than it "should" be?

solarkennedy:
Fixes #490

primary: @Rob-Johnson 

This changes how we log to stdout/err. Notably this makes it so log level (DEBUG) is global to the thing, so *everything* will start outputting, like kazoo, marathon, etc. This is probably a good thing.

Also it corrects the logging in our library files to just use their own __name__ and not to do any basicConfig themselves.

### DIFF
--- a/paasta_tools/autoscale_cluster.py
+++ b/paasta_tools/autoscale_cluster.py
@@ -18,8 +18,7 @@ import logging
 from paasta_tools.autoscaling_lib import autoscale_local_cluster
 
 
-log = logging.getLogger('__main__')
-logging.basicConfig()
+log = logging.getLogger(__name__)
 
 
 def parse_args():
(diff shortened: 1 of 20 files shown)


Rob-Johnson:
shipit! thanks for doing this!

PR #492, opened 2016-05-04
Delete get_scribe_map, since this functionality is now an option on the scribe log_reader driver.

EvanKrall:


solarkennedy:
ship

PR #497, opened 2016-05-05
paasta local-run --dry-run

kentwills:
This will allow users to check the docker command paasta will run for local-run without actually doing a build, pull, or run.  The output is on stdout in json format.


kentwills:
@asottile @solarkennedy, This helps for internal tasks at Yelp, ping me and I can send you the ticket.

PR #499, opened 2016-05-06
Make marathon and chronos constraint logic a bit more consistent

### LINKED ISSUE #456: support extra_constraints in chronos configs
We added an `extra_constraints` option to `MarathonServiceConfig`. We should add it to `ChronosJobConfig` as well.

solarkennedy:
Primary: @Rob-Johnson 

This came out of some review feedback.

I guess the tests are passing kinda "artificially" because of inheritance. Should I move or duplicate some tests out of marathon_tools and into utils?

I'm also adding deploy_whitelist but not deploy_blacklist support, and documenting them.

Also updating the schema.

Closes #456

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -372,7 +372,21 @@ Each job configuration MAY specify the following options:
     <https://mesos.github.io/chronos/docs/api.html#constraints>`_ for more
     information.
 
+  * ``extra_constraints``: Adds to the default placement constraints for
+    services. This acts the same as ``constraints``, but adds to the default
+    constraints instead of replacing them. See ``constraints`` for details on
+    format and the default constraints.
+
+    *Note*: While this parameter is the same as ``extra_constraints`` in ``marathon-$cluster.yaml``,
+    the Marathon constrain language isn't exactly like the Marathon constraint language.
+    Be sure to read the constraint documentation for Chronos referenced in the ``constraints``
+    section.
+
   * ``pool``: See the `marathon-[clustername].yaml`_ section for details
+
+  * ``deploy_whitelist``: See the `marathon-[clustername].yaml`_ section for details
+
+  * ``deploy_blacklist``: *Not currently supported*.
 
   * ``deploy_group``: Same as ``deploy_group`` for marathon-*.yaml.
 
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -225,11 +225,12 @@ class ChronosJobConfig(InstanceConfig):
         original_env = super(ChronosJobConfig, self).get_env()
         return [{"name": key, "value": value} for key, value in original_env.iteritems()]
 
-    def get_constraints(self):
-        if 'constraints' in self.config_dict:
-            constraints = self.config_dict.get('constraints')
+    def get_calculated_constraints(self):
+        constraints = self.get_constraints()
+        if constraints is not None:
+            return constraints
         else:
-            constraints = self.config_dict.get('extra_constraints', [])
+            constraints = self.get_extra_constraints()
             constraints.extend(self.get_deploy_constraints())
             constraints.extend(self.get_pool_constraints())
         assert isinstance(constraints, list)
@@ -411,7 +412,7 @@ class ChronosJobConfig(InstanceConfig):
             'mem': self.get_mem(),
             'cpus': self.get_cpus(),
             'disk': self.get_disk(),
-            'constraints': self.get_constraints(),
+            'constraints': self.get_calculated_constraints(),
             'command': parse_time_variables(self.get_cmd()) if self.get_cmd() else self.get_cmd(),
             'arguments': self.get_args(),
             'epsilon': self.get_epsilon(),
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -29,6 +29,7 @@ from tron import command_context
 
 from paasta_tools.mesos_tools import get_mesos_network_for_net
 from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import deploy_whitelist_to_constraints
 from paasta_tools.utils import get_config_hash
 from paasta_tools.utils import get_docker_url
 from paasta_tools.utils import get_paasta_branch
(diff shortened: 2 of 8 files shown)


paasta_tools/chronos_tools.py
```diff
@@ -225,7 +226,17 @@ def get_env(self):
...
     def get_constraints(self):
-        return self.config_dict.get('constraints', [['pool', 'EQUALS', self.get_pool()]])
+        if 'constraints' in self.config_dict:
+            constraints = self.config_dict.get('constraints')
+        else:
+            constraints = self.config_dict.get('extra_constraints', [])
+            constraints.extend(self.get_deploy_constraints())
+            constraints.extend(self.get_pool_constraints())
+        assert isinstance(constraints, list)
+        return [[str(val) for val in constraint] for constraint in constraints]
```
Rob-Johnson: I'd rather you moved all this logic to the caller, and have separate ``get_constraints`` and ``get_extra_constraints`` functions.

paasta_tools/chronos_tools.py
```diff
@@ -225,7 +226,17 @@ def get_env(self):
...
 
     def get_constraints(self):
-        return self.config_dict.get('constraints', [['pool', 'EQUALS', self.get_pool()]])
+        if 'constraints' in self.config_dict:
+            constraints = self.config_dict.get('constraints')
+        else:
+            constraints = self.config_dict.get('extra_constraints', [])
+            constraints.extend(self.get_deploy_constraints())
+            constraints.extend(self.get_pool_constraints())
+        assert isinstance(constraints, list)
```
Rob-Johnson: isn't this done in the corresponding check_* function?

paasta_tools/utils.py
```diff
@@ -294,6 +294,10 @@ def get_pool(self):
         :returns: the "pool" attribute in your config dict, or the string "default" if not specified."""
         return self.config_dict.get('pool', 'default')
 
+    def get_pool_constraints(self):
+        pool = self.get_pool()
+        return [["pool", "LIKE", pool]]
```
Rob-Johnson: why've you made this LIKE rather than EQUALS?

Rob-Johnson:
| Should I move or duplicate some tests out of marathon_tools and into utils?

yeah, I guess those that the test for ``get_pool_constraints`` should be done for the ``InstanceConfig``

paasta_tools/utils.py
EvanKrall: Does LIKE work on both chronos and marathon?

EvanKrall: I'm not sure this is a good invocation of DRY -- marathon needs LIKE or CLUSTER (maybe?), chronos can use LIKE or EQUALS. It's somewhat coincidental that they both have similar constraint languages, and future frameworks might not have overlapping constraint languages.

It's probably good to unify on LIKE for consistency, though.

docs/source/yelpsoa_configs.rst
```diff
@@ -372,8 +372,22 @@ Each job configuration MAY specify the following options:
...
+    *Note*: While this parameter is the same as ``extra_constraints`` in ``marathon-$cluster.yaml``,
+    the Marathon constrain language isn't exactly like the Marathon constraint language.
+    Be sure to read the constraint documentation for Chronos referenced in the ``constraints``
+    section.
+
   * ``pool``: See the `marathon-[clustername].yaml`_ section for details
 
+  * ``deploy_whitelist``: See the `marathon-[clustername].yaml`_ section for details
+
+  * ``deploy_blacklist``: *Not currently supported*.
```
EvanKrall: Is this just because https://github.com/mesos/chronos/pull/592 only was merged recently?

paasta_tools/chronos_tools.py
EvanKrall: This logic needs to be somewhere, and the caller is `format_chronos_job_dict` in this case (and `format_marathon_app_dict` for the corresponding function in MarathonServiceConfig)

I think it's better to have this logic in its own function - `format_chronos_job_dict` is already pretty hairy, and I don't want it getting any longer. Whether that's called `get_constraints` or `combine_constraints` or something is a different argument, but this isn't the only place where we have a `get_something` method that actually looks at other attributes of the config -- e.g. `MarathonServiceConfig.get_instances` is fairly complex.

docs/source/yelpsoa_configs.rst
Rob-Johnson: @EvanKrall yeah, we're waiting on a release from them (or PAASTA-4133)

paasta_tools/chronos_tools.py
Rob-Johnson: > but this isn't the only place where we have a get_something method that actually looks at other attributes of the config

yeah, and I don't like that at all. I've been bitten quite a few times where I've not been able to access the data I actually want because the corresponding ``get_*`` function does some formatting/processing before returning it.

Though I also agree that ``format_chronos_job_dict`` is pretty complex, so I think ``combine_constraints`` is a fine solution.

solarkennedy: I've named it `get_calculated_constraints` and moved the common constraint functions into utils

paasta_tools/chronos_tools.py
solarkennedy: Welll... 

I forgot which particular bug I was traking tracking down, but this helped get it.

Do we have a policy with sprinkling asserts around where we *know* what better be returned by a function?

paasta_tools/utils.py
solarkennedy: I was aware of this, and I adjusted the words to match. I agree it is kinda a tossup and went with LIKE for consistency.

solarkennedy:
Ok next pass up for review. I understand this is kinda a messy change, if requested I can make separate PRs for unifying the constraint handling before adding the new deploy_* features to chronos.

paasta_tools/marathon_tools.py
```diff
@@ -298,7 +298,7 @@ def get_drain_method_params(self, service_namespace_config):
             default = {'delay': 30}
         return self.config_dict.get('drain_method_params', default)
 
-    def get_constraints(self, service_namespace_config):
+    def get_calculated_constraints(self, service_namespace_config):
         """Gets the constraints specified in the service's marathon configuration.
 
         These are Marathon app constraints. See
```
Rob-Johnson: please update this docstring - the notes about defaults now belong with ``get_constraints`` - this docstring should describe the different sets of constraints it merges

Rob-Johnson:
fix n ship - lg2m.

I'll also note this closes #456 to see if waffle does it's thing?

solarkennedy:
Fixes #456

I don't know what the rules are, but "closes" in the middle of that comment didn't do it. Let's see if this does.

solarkennedy:
Ah, title or description: https://github.com/waffleio/waffle.io/wiki/FAQs#prs-closing-keywords

PR #501, opened 2016-05-06
Graceful container drain

Rob-Johnson:
I've used this enough times now that it should be part of the codebase. @nhandler will probably use this atleast as a basis for draining hosts in #376


solarkennedy:
Heh, I didn't expect this to drain *all* the containers on a box! :)

Still ship though, I think this is fine for a contrib script.

PR #502, opened 2016-05-06
Validate that env settings are always strings

solarkennedy:
This adds assertions that our env settings are strings.


nhandler:
Ship It!

PR #505, opened 2016-05-07
Use mesos.cli for get_mesos_leader

### LINKED ISSUE #498: Make get_mesos_leader use mesos.cli
It is possible to do:

```
from mesos.cli.master import CURRENT
CURRENT.host
```

to get a URL representing the Mesos master. This comes from Zookeeper, rather than relying on the HTTP redirects. We should update `get_mesos_leader()` in mesos_tools.py to use this method. Hopefully, this will also result in a speedup for any tasks that use this function.

NOTE: `CURRENT.host` returns something like `http://123.456.789.101:5050`, while `get_mesos_leader()` currently returns something like `srv1-somewhere.something.com`.

nhandler:
Our current approach to determine the mesos leader relies on following the redirects. mesos.cli uses zookeeper. Let's use that.

I will work on adding some itests to properly test this change (especially the scenarios where we are unable to connect to zookeeper for some reason).

Fixes #498

### DIFF
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -121,7 +121,16 @@ def get_mesos_leader(hostname=MY_HOSTNAME):
     :returns: The current mesos-master hostname"""
     try:
         url = master.CURRENT.host
-    except MesosMasterConnectionError as e:
+        hostname = urlparse(url).hostname
+        # This check is necessary, as if we parse a value such as 'localhost:5050',
+        # it won't have a hostname attribute
+        if hostname:
+            host = socket.gethostbyaddr(hostname)[0]
+            fqdn = socket.getfqdn(host)
+            return fqdn
+        else:
+            raise Exception()
+    except:
         redirect_url = 'http://%s:%s/redirect' % (hostname, MESOS_MASTER_PORT)
         try:
             r = requests.get(redirect_url, timeout=10)
@@ -131,18 +140,6 @@ def get_mesos_leader(hostname=MY_HOSTNAME):
             raise MesosMasterConnectionError(repr(e))
         r.raise_for_status()
         return urlparse(r.url).hostname
-    hostname = urlparse(url).hostname
-    # This check is necessary, as if we parse a value such as 'localhost:5050',
-    # it won't have a hostname attribute
-    if hostname:
-        try:
-            host = socket.gethostbyaddr(hostname)[0]
-            fqdn = socket.getfqdn(host)
-            return fqdn
-        except (socket.error, socket.herror, socket.gaierror, socket.timeout) as e:
-            raise MesosMasterConnectionError(repr(e))
-    else:
-        raise MesosMasterConnectionError()
 
 
 def is_mesos_leader(hostname=MY_HOSTNAME):
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -22,6 +22,7 @@ import requests
 from kazoo.client import KazooClient
 from mesos.cli import util
 from mesos.cli.exceptions import SlaveDoesNotExist
+from urlparse import urlparse
 
 from paasta_tools.utils import format_table
 from paasta_tools.utils import PaastaColors
@@ -111,17 +112,35 @@ class MesosSlaveConnectionError(Exception):
     pass
 
 
-def get_mesos_leader():
+def get_mesos_leader(hostname=MY_HOSTNAME):
     """Get the current mesos-master leader's hostname.
+    Attempts to determine this by using mesos.cli to query ZooKeeper.
+    If this fails, we fallback to following the HTTP redirect
 
+    :param hostname: The hostname to query mesos-master on (only used if we can't use ZK)
     :returns: The current mesos-master hostname"""
-    url = master.CURRENT.host
-    m = re.search('^https?://(.*?):\d+$', url)
-    if m:
-        ip = m.group(1)
-        host = socket.gethostbyaddr(ip)[0]
-        fqdn = socket.getfqdn(host)
-        return fqdn
+    try:
+        url = master.CURRENT.host
+    except MesosMasterConnectionError as e:
+        redirect_url = 'http://%s:%s/redirect' % (hostname, MESOS_MASTER_PORT)
+        try:
+            r = requests.get(redirect_url, timeout=10)
+        except requests.exceptions.ConnectionError as e:
+            # Repackage the exception so upstream code can handle this case without
+            # knowing our implementation details.
+            raise MesosMasterConnectionError(repr(e))
+        r.raise_for_status()
+        return urlparse(r.url).hostname
+    hostname = urlparse(url).hostname
+    # This check is necessary, as if we parse a value such as 'localhost:5050',
+    # it won't have a hostname attribute
+    if hostname:
+        try:
+            host = socket.gethostbyaddr(hostname)[0]
+            fqdn = socket.getfqdn(host)
+            return fqdn
+        except (socket.error, socket.herror, socket.gaierror, socket.timeout) as e:
+            raise MesosMasterConnectionError(repr(e))
     else:
         raise MesosMasterConnectionError()
 
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 import datetime
 import json
+import logging
 import os
 import re
 import socket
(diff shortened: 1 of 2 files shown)


paasta_tools/mesos_tools.py
```diff
@@ -22,6 +22,7 @@
 from kazoo.client import KazooClient
 from mesos.cli import util
 from mesos.cli.exceptions import SlaveDoesNotExist
+from mesos.cli.master import CURRENT
```
huadongliu: mesos_tools imports mesos.cli master at line 79 and then uses master.CURRENT.xyz afterwards. You may want to do the same.

 77 MESOS_MASTER_PORT = 5050
 78 MESOS_SLAVE_PORT = '5051'
 79 from mesos.cli import master  # noqa

huadongliu: Please ignore my comment. I missed patch two earlier.

paasta_tools/mesos_tools.py
```diff
@@ -111,29 +111,29 @@ class MesosSlaveConnectionError(Exception):
...
-    try:
-        r = requests.get(redirect_url, timeout=10)
-    except requests.exceptions.ConnectionError as e:
-        # Repackage the exception so upstream code can handle this case without
-        # knowing our implementation details.
-        raise MesosMasterConnectionError(repr(e))
-    r.raise_for_status()
-    return re.search('(?<=http://)[0-9a-zA-Z\.\-]+', r.url).group(0)
+    url = master.CURRENT.host
+    print "URL: %s" % url
```
Rob-Johnson: can you get rid of these prints, or using logging.debug and a more useful message please

paasta_tools/mesos_tools.py
```diff
@@ -111,29 +111,29 @@ class MesosSlaveConnectionError(Exception):
...
+        raise MesosMasterConnectionError()
 
 
 def is_mesos_leader(hostname=MY_HOSTNAME):
     """Check if a hostname is the current mesos leader.
 
     :param hostname: The hostname to query mesos-master on
     :returns: True if hostname is the mesos-master leader, False otherwise"""
-    return hostname in get_mesos_leader(hostname)
+    return hostname in get_mesos_leader()
```
Rob-Johnson: I know that this was here before you, but shouldn't we be exact matching here?

suppose ``get_mesos_leader`` returns ``paasta6-blah.yelp.com`` - ``is_mesos_leader(hostname='blah')`` will return True, won't it?

paasta_tools/mesos_tools.py
```diff
@@ -111,29 +111,29 @@ class MesosSlaveConnectionError(Exception):
...
+    print "URL: %s" % url
+    print "CFG: %s" % os.environ['MESOS_CLI_CONFIG']
+    m = re.search('^https?://(.*?):\d+$', url)
+    if m:
+        ip = m.group(1)
+        host = socket.gethostbyaddr(ip)[0]
+        fqdn = socket.getfqdn(host)
+        return fqdn
+    else:
+        raise MesosMasterConnectionError()
```
Rob-Johnson: this is the wrong Exception to throw - this is raised when a regex doesn't match, not when you can't connect to the master.

paasta_tools/mesos_tools.py
```diff
@@ -111,29 +111,29 @@ class MesosSlaveConnectionError(Exception):
...
-        raise MesosMasterConnectionError(repr(e))
-    r.raise_for_status()
-    return re.search('(?<=http://)[0-9a-zA-Z\.\-]+', r.url).group(0)
+    url = master.CURRENT.host
+    print "URL: %s" % url
+    print "CFG: %s" % os.environ['MESOS_CLI_CONFIG']
+    m = re.search('^https?://(.*?):\d+$', url)
+    if m:
+        ip = m.group(1)
+        host = socket.gethostbyaddr(ip)[0]
```
Rob-Johnson: please add some error handling to these two socket calls

paasta_tools/mesos_tools.py
```diff
@@ -111,29 +111,29 @@ class MesosSlaveConnectionError(Exception):
...
-    except requests.exceptions.ConnectionError as e:
-        # Repackage the exception so upstream code can handle this case without
-        # knowing our implementation details.
-        raise MesosMasterConnectionError(repr(e))
-    r.raise_for_status()
-    return re.search('(?<=http://)[0-9a-zA-Z\.\-]+', r.url).group(0)
+    url = master.CURRENT.host
+    print "URL: %s" % url
+    print "CFG: %s" % os.environ['MESOS_CLI_CONFIG']
+    m = re.search('^https?://(.*?):\d+$', url)
```
Rob-Johnson: please use [urlparse](https://docs.python.org/2/library/urlparse.html) rather than writing regexps (I know we had a regexp before, but it's an easy win whilst you're here)

paasta_tools/mesos_tools.py
```diff
@@ -111,29 +111,29 @@ class MesosSlaveConnectionError(Exception):
...
-    redirect_url = 'http://%s:%s/redirect' % (hostname, MESOS_MASTER_PORT)
-    try:
-        r = requests.get(redirect_url, timeout=10)
-    except requests.exceptions.ConnectionError as e:
-        # Repackage the exception so upstream code can handle this case without
-        # knowing our implementation details.
-        raise MesosMasterConnectionError(repr(e))
-    r.raise_for_status()
-    return re.search('(?<=http://)[0-9a-zA-Z\.\-]+', r.url).group(0)
+    url = master.CURRENT.host
```
Rob-Johnson: should we have some error handling here to catch failures connecting to ZK? atm we monkey patch log.fatal so we throw actual exceptions, please catch/deal with them accordingly

Rob-Johnson:
please add some more tests to deal with connection failures

nhandler:
I've refactored the code to fallback to our old approach of following the redirect in the event of an exception. I also (re-)added some tests for the failing situations.

paasta_tools/mesos_tools.py
```diff
@@ -112,28 +113,41 @@ class MesosSlaveConnectionError(Exception):
...
-        # knowing our implementation details.
-        raise MesosMasterConnectionError(repr(e))
-    r.raise_for_status()
-    return re.search('(?<=http://)[0-9a-zA-Z\.\-]+', r.url).group(0)
+        url = master.CURRENT.host
+        hostname = urlparse(url).hostname
+        # This check is necessary, as if we parse a value such as 'localhost:5050',
+        # it won't have a hostname attribute
+        if hostname:
+            host = socket.gethostbyaddr(hostname)[0]
```
solarkennedy: Yuck. I don't really like how this depends on the dns being correct for both the forward and reverse dns. I guess it isn't that much better or worse than url parsing and assuming the hostname was set correct on the box in the first place

paasta_tools/mesos_tools.py
```diff
@@ -112,28 +113,41 @@ class MesosSlaveConnectionError(Exception):
...
+            fqdn = socket.getfqdn(host)
+            return fqdn
+        else:
+            raise Exception()
+    except:
```
solarkennedy: Can you add plenty of log.debugs here? if we are falling through to this backup method, I would like to see it during debug, and it will help in all these cases "what does mesos *think* the leader is?"

solarkennedy: I'm a little worried about this complexity, do we need it?

If zk discovery doesn't work, then I seriously doubt the mesos master http thing is going to work either. Can we only have one method to do this and only add a fallback method if we find that it is unreliable?

nhandler: There are other instances where it is possible for the new method to fail, but the old one to succeed. For example, if there are issues with the mesos-cli config file, the redirect approach would still work, while the new zk lookup won't.

But I'm fine removing it. Do you want to just raise an exception in this case?

solarkennedy: Yes please. If we can't work with zk or the mesos-cli isn't setup, then we have other issues.

Let it raise the normal exception it would raise. Screw html redirect parsing.

nhandler:
Removed the fallback code and added some more logging.

paasta_tools/mesos_tools.py
```diff
@@ -29,6 +31,9 @@
 from paasta_tools.utils import TimeoutError
 
 
+log = logging.getLogger(__name__)
+
```
solarkennedy: Also 

solarkennedy:
ship

paasta_tools/mesos_tools.py
```diff
@@ -111,29 +117,42 @@ class MesosSlaveConnectionError(Exception):
     pass
 
 
-def get_mesos_leader(hostname=MY_HOSTNAME):
-    """Get the current mesos-master leader's hostname. Raise
-    MesosMasterConnectionError if we can't connect.
+def get_mesos_leader():
+    """Get the current mesos-master leader's hostname.
+    Attempts to determine this by using mesos.cli to query ZooKeeper.
+    If this fails, we fallback to following the HTTP redirect
```
Rob-Johnson: please remove this and the :param: now that we're no longer doing this

Rob-Johnson:
fix and ship

PR #506, opened 2016-05-09
Make paasta status use the multi-instance endpoint

huadongliu:
Multiple instances of a service is passed to the endpoint with the -i option in a comma-separated list of instances. The result of each instance is streamed back the client and prefixed with spaces accordingly, 2 for the instance line and 4 for the rest.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -154,11 +154,11 @@ def report_status_for_cluster(service, cluster, deploy_pipeline, actual_deployme
         # Case: service NOT deployed to cluster.instance
         else:
             print '  instance: %s' % PaastaColors.red(instance)
-            print '    Git sha:    None'
+            print '    Git sha:    None (not deployed yet)'
 
-    if len(deployed_instances):
+    if len(deployed_instances) > 0:
         status = execute_paasta_serviceinit_on_remote_master('status', cluster, service, ','.join(deployed_instances),
-                                                             system_paasta_config, verbose=verbose, stream=True)
+                                                             system_paasta_config, stream=True, verbose=verbose)
         # Status results are streamed. This print is for possible error messages.
         if status is not None:
             for line in status.rstrip().split('\n'):
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -136,6 +136,8 @@ def report_status_for_cluster(service, cluster, deploy_pipeline, actual_deployme
     print
     print "cluster: %s" % cluster
     seen_instances = []
+    deployed_instances = []
+
     for namespace in deploy_pipeline:
         cluster_in_pipeline, instance = namespace.split('.')
         seen_instances.append(instance)
@@ -147,19 +149,17 @@ def report_status_for_cluster(service, cluster, deploy_pipeline, actual_deployme
 
         # Case: service deployed to cluster.instance
         if namespace in actual_deployments:
-            formatted_instance = PaastaColors.blue(instance)
-            version = actual_deployments[namespace][:8]
-            # TODO: Perform sanity checks once per cluster instead of for each namespace
-            status = execute_paasta_serviceinit_on_remote_master('status', cluster, service, instance,
-                                                                 system_paasta_config, verbose=verbose)
+            deployed_instances.append(instance)
+
         # Case: service NOT deployed to cluster.instance
         else:
-            formatted_instance = PaastaColors.red(instance)
-            version = 'None'
-            status = None
+            print '  instance: %s' % PaastaColors.red(instance)
+            print '    Git sha:    None'
 
-        print '  instance: %s' % formatted_instance
-        print '    Git sha:    %s' % version
+    if len(deployed_instances):
+        status = execute_paasta_serviceinit_on_remote_master('status', cluster, service, ','.join(deployed_instances),
+                                                             system_paasta_config, verbose=verbose, stream=True)
+        # Status results are streamed. This print is for possible error messages.
         if status is not None:
             for line in status.rstrip().split('\n'):
                 print '    %s' % line
--- a/paasta_tools/cli/utils.py
+++ b/paasta_tools/cli/utils.py
@@ -474,7 +474,7 @@ def check_ssh_and_sudo_on_master(master, timeout=10):
     return (False, output)
 
 
-def run_paasta_serviceinit(subcommand, master, service, instancenames, cluster, **kwargs):
+def run_paasta_serviceinit(subcommand, master, service, instances, cluster, stream, **kwargs):
     """Run 'paasta_serviceinit <subcommand>'. Return the output from running it."""
     if 'verbose' in kwargs and kwargs['verbose'] > 0:
         verbose_flag = '-v ' * kwargs['verbose']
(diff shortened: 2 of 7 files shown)


paasta_tools/cli/utils.py
```diff
@@ -490,16 +490,21 @@ def run_paasta_serviceinit(subcommand, master, service, instancename, cluster, *
         delta = "--delta %s" % kwargs['delta']
     else:
         delta = ''
-    command = 'ssh -A -n %s sudo paasta_serviceinit %s%s%s %s %s' % (
+    if 'stream' in kwargs and kwargs['stream']:
```
Rob-Johnson: any reason to hide the ``stream`` parameter in kwargs? What's special about it that it's not just a regular named parameter that's defaulting to ``False``?

paasta_tools/cli/utils.py
```diff
@@ -490,16 +490,21 @@ def run_paasta_serviceinit(subcommand, master, service, instancename, cluster, *
...
+        stream = kwargs['stream']
+    else:
+        stream = False
+    command = 'ssh -A -n %s sudo paasta_serviceinit %s%s-s %s -i %s %s %s' % (
```
Rob-Johnson: eh, this is getting a bit difficult to read. How about using the dict formatting?

```
'ssh -A -n %(master) sudo paasta_serviceinit %(verbose_flag) %(app_id_flag) -s %(service) -i %(instancenames) %(subcommand) %(delta)' % {
    'master': master,
    'verbose_flag': verbose_flag,
    ...
}
```

paasta_tools/cli/cmds/status.py
```diff
@@ -147,19 +149,17 @@ def report_status_for_cluster(service, cluster, deploy_pipeline, actual_deployme
...
         else:
-            formatted_instance = PaastaColors.red(instance)
-            version = 'None'
-            status = None
+            print '  instance: %s' % PaastaColors.red(instance)
+            print '    Git sha:    None'
 
-        print '  instance: %s' % formatted_instance
-        print '    Git sha:    %s' % version
+    if len(deployed_instances):
```
Rob-Johnson: can you be more explicit about what you're checking here please
```
if len(deployed_instances) > 0:
```

paasta_tools/cli/utils.py
Rob-Johnson: I see that this pattern came from the same for verbose flag - can you fix that whilst you're here please

paasta_tools/utils.py
```diff
@@ -874,7 +874,12 @@ def _run(command, env=os.environ, timeout=None, log=False, stream=False, stdin=N
             proctimer = threading.Timer(timeout, _timeout, (process,))
             proctimer.start()
         for line in iter(process.stdout.readline, ''):
-            if stream:
+            if stream and (' paasta_serviceinit ' in command and ' status ' in command):
```
Rob-Johnson: this seems pretty fragile. Can you instead move this into a function and have unit tests?

paasta_tools/cli/cmds/status.py
```diff
@@ -147,19 +149,17 @@ def report_status_for_cluster(service, cluster, deploy_pipeline, actual_deployme
...
-            status = execute_paasta_serviceinit_on_remote_master('status', cluster, service, instance,
-                                                                 system_paasta_config, verbose=verbose)
+            deployed_instances.append(instance)
+
         # Case: service NOT deployed to cluster.instance
         else:
-            formatted_instance = PaastaColors.red(instance)
-            version = 'None'
-            status = None
+            print '  instance: %s' % PaastaColors.red(instance)
```
solarkennedy: It is hard to read from the diff, but does this mean it prints all the instance names and git shas, and *then* prints that status for all of them?

I would rather it be interweaved. (where the server component does the printing)

paasta_tools/cli/utils.py
```diff
@@ -490,16 +490,21 @@ def run_paasta_serviceinit(subcommand, master, service, instancename, cluster, *
...
         master,
         verbose_flag,
         app_id_flag,
-        compose_job_id(service, instancename),
+        service,
+        instancenames,
```
solarkennedy: can you call this variable: `instances` please

paasta_tools/paasta_serviceinit.py
```diff
@@ -85,7 +87,16 @@ def main():
         sys.exit(1)
 
     cluster = load_system_paasta_config().get_cluster()
+    actual_deployments = get_actual_deployments(service, args.soa_dir)
+
     for instance in instances:
+        # For an instance, there might be multiple versions running, e.g. in crossover bouncing.
+        # In addition, mesos master does not have information of a chronos service's git hash.
+        # The git sha in deployment.json is simply used here.
+        version = actual_deployments['.'.join((cluster, instance))][:8]
```
solarkennedy: This is a bit odd, we don't normally report the git sha in this way.

I'm "ok" with this as a temporary measure, but longer term I would rather this be explicit about the desired versus actual here, because this reflects the "desired" git sha

paasta_tools/cli/cmds/status.py
solarkennedy: On a second pass I think I understand what this is for.

Can you make it more obvious with

    print ' Git sha:       None (not deployed yet)'

paasta_tools/cli/cmds/status.py
```diff
@@ -147,19 +149,17 @@ def report_status_for_cluster(service, cluster, deploy_pipeline, actual_deployme
...
+            print '    Git sha:    None'
 
-        print '  instance: %s' % formatted_instance
-        print '    Git sha:    %s' % version
+    if len(deployed_instances):
+        status = execute_paasta_serviceinit_on_remote_master('status', cluster, service, ','.join(deployed_instances),
+                                                             system_paasta_config, verbose=verbose, stream=True)
+        # Status results are streamed. This print is for possible error messages.
         if status is not None:
```
solarkennedy: What types of errors will end up on here no, stdrr?
Can you make it more obvious if that is the case? 

    print "Unexpected errors: "

or something

solarkennedy:
Can you pastebin an output of what this looks like after changes?

paasta_tools/cli/utils.py
huadongliu: execute_paasta_serviceinit_on_remote_master is used by a number of commands. Each of these commands has its set of arguments. Arguments that are not shared by these commands are made into kwargs. For example, app_id and delta are "paasta scale" only; verbose and stream are "paasta status" only. You may want to leave the verbose flag and the stream flag in kwargs.

huadongliu: Correction: stream is common to all commands, while verbose is "paasta status" only. I will leave the verbose flag unchanged while make stream a default argument if not explicitly specified.

huadongliu:
I found some issues with ssh streaming. Please wait for the next patch to review.

huadongliu:
Per kwa's request, I hard coded the CLI to point to the multi-instance endpoint at mesos1-uswest1cdevc. The output is kept at https://pb.yelpcorp.com/198865. Please ignore the ssh command print in the output.

solarkennedy:
lets do another pin + ship on this one

Rob-Johnson:
ship

PR #507, opened 2016-05-09
fix execute_paasta_serviceinit_call for emergency-start

Rob-Johnson:
closes internal PAASTA-4430

/cc @drolando


solarkennedy:
ship. We need to junk these commands....

PR #509, opened 2016-05-10
Log lots more things when doing cluster autoscaling

solarkennedy:
This is one of those things where I want to be sure of the "Why" when this thing calculates decisions.

This change makes it at the normal log level it answers the first why, and with -v it answers the second why, and -vv answers the deepest whys (you see boto xml responses and junk)


solarkennedy:
I'm going to merge this in because it is helping me debug this code right now and it is easier to have it in master, feel free to give any other code review and I'll make changes as well.

PR #510, opened 2016-05-10
Make paasta_autoscale_cluster and use console_scripts for it

solarkennedy:
I would like to start a new trend where we prepend our "plumbing" scripts with `paasta_` to make it more obvious what the command is for and where it came from.

`autoscale_cluster` is kinda unadorned. Is it for scaling a kubernetes cluster? A mysql one? Who knows.

This is actually an easy migration because in the links file you can specify multiple symlinks and deprecate the old ones. I didn't do that with this PR, but I will in the next one if shipit.


Rob-Johnson:
shipit

solarkennedy:
Hmm. For some reason itest_trusty is failing because we are not building with cyaml. How did that ever work?

PR #511, opened 2016-05-10
Load paasta configs from /etc/paasta recursively

solarkennedy:
I would like to start recursively reading files from /etc/paasta.

Mostly to make it easier for autoscaling where we'll have multiple json files per aws resource and I would like to have a dedicated folder for it.

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -33,6 +33,7 @@ import signal
 import sys
 import tempfile
 import threading
+from fnmatch import fnmatch
 from functools import wraps
 from subprocess import PIPE
 from subprocess import Popen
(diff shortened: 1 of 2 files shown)


paasta_tools/utils.py
```diff
@@ -677,20 +678,22 @@ class NoConfigurationForServiceError(Exception):
     pass
 
 
-def get_readable_files_in_glob(input_glob):
+def get_readable_files_in_glob(glob, path):
     """
-    Returns lexicographically-sorted list of files that are readable in an input glob
+    Returns a sorted list of files that are readable in an input glob by recusivly searching a path
```
EvanKrall: recursively

paasta_tools/utils.py
```diff
@@ -677,20 +678,22 @@ class NoConfigurationForServiceError(Exception):
...
+        for f in files:
+            fn = os.path.join(root, f)
+            if os.path.isfile(fn) and os.access(fn, os.R_OK) and fnmatch(fn, glob):
+                globbed_files.append(fn)
+    return sorted(globbed_files)
 
 
 def load_system_paasta_config(path=PATH_TO_SYSTEM_PAASTA_CONFIG_DIR):
     """
-    Reads Paasta configs in specified directory in lexicographical order and merges duplicated keys (last file wins)
```
EvanKrall: don't believe in specifying that it's lexicographical order?

solarkennedy: I didn't think that was really necessary

EvanKrall: Well, if we're going to say which file wins (`(last file wins)`), we should specify what we mean by "last".

EvanKrall:
Ship, seems reasonable.

PR #512, opened 2016-05-10
Update copyright for 2016

nhandler:
All of our copyright statements still say 'Copyright 2015 Yelp Inc.'

We should update them now that it is 2016.


EvanKrall:
ship

PR #513, opened 2016-05-10
Make setup_marathon_job capable of handling a list of service.instance

huadongliu:
This patch enables setup_marathon_job.py handle a list of service.instance to take advantage of mesos master state caching.  
{code}
$ .tox/py27/bin/setup_marathon_job.py --help
usage: setup_marathon_job.py [-h] [-d SOA_DIR] [-v]
                             SERVICE.INSTANCE [SERVICE.INSTANCE ...]

Creates marathon jobs.
{code}

We have a 60 seconds caching TTL at this point, and we have 244 service instances on prod. I am breaking into 4 chunks, hoping each chunk can finish within 60s. In addition, I removed service list shuffling. IMO, deploying instances of the same service in the same chunk process has better locality.
{code}
list_marathon_service_instances  | xargs -n 64 -r -P 4 .tox/py27/bin/setup_marathon_job.py
{code}

### DIFF
--- a/paasta_tools/deploy_marathon_services
+++ b/paasta_tools/deploy_marathon_services
@@ -1,5 +1,5 @@
 #!/bin/bash
 
 if am_i_mesos_leader >/dev/null; then
-  list_marathon_service_instances | xargs -n 64 -r -P 4 setup_marathon_job
+    list_marathon_service_instances | shuf | xargs --max-args=64 --no-run-if-empty --max-procs=4 setup_marathon_job
 fi
--- a/paasta_tools/deploy_marathon_services
+++ b/paasta_tools/deploy_marathon_services
@@ -1,5 +1,5 @@
 #!/bin/bash
 
 if am_i_mesos_leader >/dev/null; then
-  list_marathon_service_instances | shuf | xargs -n 1 -r -P 5 setup_marathon_job
+  list_marathon_service_instances | xargs -n 64 -r -P 4 setup_marathon_job
 fi
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -73,7 +73,7 @@ log = logging.getLogger(__name__)
 def parse_args():
     parser = argparse.ArgumentParser(description='Creates marathon jobs.')
     parser.add_argument('service_instance_list', nargs='+',
-                        help="The marathon instance of the service to create or update",
+                        help="The list of marathon service instances to create or update",
                         metavar="SERVICE%sINSTANCE" % SPACER)
     parser.add_argument('-d', '--soa-dir', dest="soa_dir", metavar="SOA_DIR",
                         default=marathon_tools.DEFAULT_SOA_DIR,
@@ -581,8 +581,8 @@ def main():
             if deploy_marathon_service(service, instance, client, soa_dir, marathon_config):
                 num_failed_deployments = num_failed_deployments + 1
 
-    print("%d out of %d service.instances are successfully deployed" %
-          (len(args.service_instance_list) - num_failed_deployments, len(args.service_instance_list)))
+    log.debug("%d out of %d service.instances failed to deploy." %
+              (num_failed_deployments, len(args.service_instance_list)))
 
     sys.exit(1 if num_failed_deployments else 0)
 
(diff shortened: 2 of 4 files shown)


tests/test_setup_marathon_job.py
```diff
@@ -54,7 +54,7 @@ class TestSetupMarathonJob:
         'password': 'admin_pass',
     }, '/fake/fake_file.json')
     fake_args = mock.MagicMock(
-        service_instance='what_is_love.bby_dont_hurt_me',
+        service_instance_list=['what_is_love.bby_dont_hurt_me'],
```
solarkennedy: args can't really be an array, they will be a string, you are going to have to csv this or do multiple args

paasta_tools/setup_marathon_job.py
```diff
@@ -72,7 +72,7 @@
 
 def parse_args():
     parser = argparse.ArgumentParser(description='Creates marathon jobs.')
-    parser.add_argument('service_instance',
+    parser.add_argument('service_instance_list', nargs='+',
```
solarkennedy: Oh. Hmmm..... TIL

paasta_tools/deploy_marathon_services
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
 
 if am_i_mesos_leader >/dev/null; then
-  list_marathon_service_instances | shuf | xargs -n 1 -r -P 5 setup_marathon_job
+  list_marathon_service_instances | xargs -n 64 -r -P 4 setup_marathon_job
```
solarkennedy: Can you still leave in the shuf?
The shuf protects us in a different way, so that if we do have a service that gets stuck and hurts us, at least on the next run there is a chance for randomness to help give others a chance.

paasta_tools/setup_marathon_job.py
```diff
@@ -72,7 +72,7 @@
...
                         help="The marathon instance of the service to create or update",
```
solarkennedy: can you update the help string?

paasta_tools/deploy_marathon_services
solarkennedy: Also, for best practice can you use the long-form of these xargs args to make it so I don't have to remember or lookup what they mean?

tests/test_setup_marathon_job.py
huadongliu: From https://docs.python.org/2/library/argparse.html#nargs:

'+'. Just like '*', all command-line args present are gathered into a list. Additionally, an error message will be generated if there wasn’t at least one command-line argument present.

paasta_tools/setup_marathon_job.py
```diff
@@ -582,29 +596,27 @@ def main():
             soa_dir=soa_dir,
         )
     except NoDeploymentsAvailable:
-        log.debug("No deployments found for %s in cluster %s. Skipping." % (args.service_instance,
-                                                                            load_system_paasta_config().get_cluster()))
-        sys.exit(0)
+        log.debug("No deployments found for %s.%s in cluster %s. Skipping." %
+                  (service, instance, load_system_paasta_config().get_cluster()))
+        return 1
```
solarkennedy: This should return 0, it is not an error condition.

paasta_tools/setup_marathon_job.py
```diff
@@ -547,33 +547,47 @@ def setup_service(service, instance, client, marathon_config,
...
+        try:
+            service, instance, _, __ = decompose_job_id(service_instance)
+        except InvalidJobNameError:
+            log.error("Invalid service instance specified. Format is service%sinstance." % SPACER)
+            num_failed_deployments = num_failed_deployments + 1
+        else:
+            if deploy_marathon_service(service, instance, client, soa_dir, marathon_config):
+                num_failed_deployments = num_failed_deployments + 1
+
+    print("%d out of %d service.instances are successfully deployed" %
```
solarkennedy: I would say log.debug this. This will be running very often and will clutter the logs with many of these messages.

For the most part this command should be "no news is good news" in the normal mode. (non verbose)

solarkennedy:
The "locality" (re: shuf) is a bit overrated here. There is no real locality going on here, it is python code man :)

tests/test_setup_marathon_job.py
solarkennedy: Nice!

Rob-Johnson:
shipit - this lg2m. Is there anything to be worried about with such a long cache time?

huadongliu:
robj, this patch does not change the 60 seconds mesos master state TTL, which was bumped up from 5 seconds one week ago.

solarkennedy:
ship

PR #517, opened 2016-05-12
Move http request cache install up to paasta serviceinit

huadongliu:
Current http request cache is installed at either marathon or chronos serviceinit per service.instance. Move it up at paasta serviceinit to serve all service instances.

### DIFF
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -22,6 +22,7 @@ from urlparse import urlparse
 import humanize
 import requests
 from kazoo.client import KazooClient
+from mesos.cli import util
 from mesos.cli.exceptions import SlaveDoesNotExist
 
 from paasta_tools.utils import format_table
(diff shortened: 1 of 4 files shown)


solarkennedy:
I would like to see some confirmation that this works for real. Can you time a multi-instance invocation of `paasta_serviceinit` with and without this change?

huadongliu:
Unfortunately I don't see a clear difference with and without the change. I added 50 instances to the robj-test-service to get the numbers.

huadongliu:
I debugged into request_cache and found that install_cache is session based. Thus, per instance install_cache essentially creates many sessions and there is no subsequent http request to make use of the cache. We see improved "paasta status" numbers because of the mesos.cli ttl=60 monkey patch.

I removed the mesos.cli ttl=60 monkey patch and then applied this change. I am seeing pretty much the same result as 0.19.4(master).

I will update the patch with mesos.cli ttl monkey patch removal.

paasta_tools/mesos_tools.py
```diff
@@ -86,12 +85,6 @@ def raise_cli_exception(msg):
 import mesos.cli.cluster  # noqa
 
 
-# monkey patch MesosMaster.state to use a larger ttl
-@util.CachedProperty(ttl=60)
```
solarkennedy: I would say leave this here for now. This is helping speed up *other* things that are not serviceinit. (like deploys, check scripts, etc)

I agree that we *can* remove this if you want to deploy the requests cache in all the places, but I'm ok if this PR just has serviceinit stuff first.

solarkennedy:
fix + ship

PR #518, opened 2016-05-12
just pass on non-paasta rather than skipping in graceful_container_drain.py

Rob-Johnson:
primary @giuliano108 

this just skips the non-paasta containers, rather than bailing out


solarkennedy:
ship

PR #519, opened 2016-05-12
Try to be more consistent about running 'master' scripts on the current leader

solarkennedy:
We seem to be not very consistent with ensuring these scripts run on the current leader.

I found that my autoscale_cluster was running 3 copies at once, causing issues. It and these other script are only designed to run "one at a time" on the current leader.

### DIFF
--- a/paasta_tools/autoscale_all_services.py
+++ b/paasta_tools/autoscale_all_services.py
@@ -16,6 +16,7 @@ import argparse
 
 from paasta_tools.autoscaling_lib import autoscale_services
 from paasta_tools.marathon_tools import DEFAULT_SOA_DIR
+from paasta_tools.mesos_tools import is_mesos_leader
 
 
 def parse_args():
@@ -33,5 +34,5 @@ def main():
     autoscale_services(soa_dir)
 
 
-if __name__ == '__main__':
+if __name__ == "__main__" and is_mesos_leader():
     main()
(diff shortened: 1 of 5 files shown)


EvanKrall:
If it's an actual issue to run more than one concurrently, we should be grabbing a ZK lock.

paasta_tools/autoscale_all_services.py
```diff
@@ -33,5 +34,5 @@ def main():
     autoscale_services(soa_dir)
 
 
-if __name__ == '__main__':
+if __name__ == "__main__" and is_mesos_leader():
```
nhandler: I've not seen this style used before. Any reason for doing this instead of adding the check inside of the various `main()` methods?

EvanKrall:
Do we really need to enforce this at this level, preventing an operator from running these scripts anywhere except the mesos leader? A lot of these scripts can be run from anywhere, so long as they have a connection+creds to mesos/marathon.

paasta_tools/autoscale_all_services.py
solarkennedy: no particular reason

solarkennedy:
It is true, if you want I can move this logic to the cron level? Then these commands will "always" run if run manually by an admin

solarkennedy:
I've made an internal CR for pushing this logic to the crons instead of having it in the python.

EvanKrall:
Cool, +1 for doing `am_i_mesos_leader &&` in cron jobs for informal locking. If there's actual issues that arise from running multiple copies of the script in parallel though, we really should be grabbing ZK locks.

solarkennedy:
I agree. Discarding for now.

PR #520, opened 2016-05-12
Migrate chronos scripts to use console_scripts

solarkennedy:
I want to do this carefully and in batches.

I'm moving chronos stuff first and prefixing them with `paasta_`. I'm leaving backwards compatible links for now so existing stuff keeps working. (crons, etc)

Notice I didn't change the itests (ubuntu.sh) at all, they should continue to work as is and prove that things wont break?


EvanKrall:
Ship carefully.

solarkennedy:
Awesome, the itest caught my `clenaup` typo.

PR #521, opened 2016-05-12
Make setup_marathon_job aware of hosts marked as draining

nhandler:
This is still a work in progress. But I decided to open up a PR to get a little more visibility.

This change attempts to make setup_marathon_job aware of hosts that are marked as draining (at-risk). If it notices that an application has tasks running on such hosts, it will scale up the application by that number.

I still need to add some paasta_itests for this change.

Connects to #376

### DIFF
--- a/paasta_itests/setup_marathon_job.feature
+++ b/paasta_itests/setup_marathon_job.feature
@@ -23,13 +23,13 @@ Feature: setup_marathon_job can create a "complete" app
       And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
       And an old app to be destroyed with constraints [["hostname", "UNIQUE"]]
     When there are 2 old healthy tasks
-     And we mark the host "mesosslave" as at-risk
-     And setup_service is initiated
+      And we mark the host "mesosslave" as at-risk
+      And setup_service is initiated
     When there are 2 new healthy tasks
-     And setup_service is initiated
-     And we wait a bit for the old app to disappear
+      And setup_service is initiated
+      And we wait a bit for the old app to disappear
     Then the old app should be gone
-     And the tasks on the host "mesosslave" should be drained
+      And the tasks on the host "mesosslave" should be drained
 
   Scenario: marathon apps can be scaled down
     Given a working paasta cluster
--- a/paasta_itests/setup_marathon_job.feature
+++ b/paasta_itests/setup_marathon_job.feature
@@ -29,7 +29,7 @@ Feature: setup_marathon_job can create a "complete" app
       And setup_service is initiated
       And we wait a bit for the old app to disappear
     Then the old app should be gone
-      And the tasks on the host "mesosslave" should be drained
+      And there should be 0 tasks on the host "mesosslave"
 
   Scenario: marathon apps can be scaled down
     Given a working paasta cluster
--- a/paasta_itests/setup_marathon_job.feature
+++ b/paasta_itests/setup_marathon_job.feature
@@ -18,6 +18,19 @@ Feature: setup_marathon_job can create a "complete" app
       And we run setup_marathon_job until it has 5 task(s)
      Then we should see the number of instances become 5
 
+  Scenario: marathon apps can be scaled up with at-risk hosts
+    Given a working paasta cluster
+      And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
+      And an old app to be destroyed with constraints [["hostname", "UNIQUE"]]
+    When there are 2 old healthy tasks
+     And we mark the host "mesosslave" as at-risk
+     And setup_service is initiated
+    When there are 2 new healthy tasks
+     And setup_service is initiated
+     And we wait a bit for the old app to disappear
+    Then the old app should be gone
+     And the tasks on the host "mesosslave" should be drained
+
   Scenario: marathon apps can be scaled down
     Given a working paasta cluster
       And I have yelpsoa-configs for the marathon job "test-service.main"
--- a/paasta_itests/setup_marathon_job.feature
+++ b/paasta_itests/setup_marathon_job.feature
@@ -20,16 +20,29 @@ Feature: setup_marathon_job can create a "complete" app
 
   Scenario: marathon apps can be scaled up with at-risk hosts
     Given a working paasta cluster
-      And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
+     And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
       And an old app to be destroyed with constraints [["hostname", "UNIQUE"]]
     When there are 2 old healthy tasks
-      And we mark the host "mesosslave" as at-risk
-      And setup_service is initiated
-    When there are 2 new healthy tasks
-      And setup_service is initiated
+    When setup_service is initiated
+      And there are 2 new healthy tasks
+    When we mark a host it is running on as at-risk
+    When setup_service is initiated
+      And there are 3 new healthy tasks
       And we wait a bit for the old app to disappear
     Then the old app should be gone
-      And there should be 0 tasks on the host "mesosslave"
+      And there should be 0 tasks on that at-risk host
+
+  Scenario: marathon apps can be deployed with at-risk hosts
+    Given a working paasta cluster
+      And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
+    When setup_service is initiated
+      And there are 2 new healthy tasks
+    When we mark a host it is running on as at-risk
+    When setup_service is initiated
+      And there are 3 new healthy tasks
+    When setup_service is initiated
+      And there are 2 new healthy tasks
+    Then there should be 0 tasks on that at-risk host
 
   Scenario: marathon apps can be scaled down
     Given a working paasta cluster
--- a/paasta_itests/setup_marathon_job.feature
+++ b/paasta_itests/setup_marathon_job.feature
@@ -1,5 +1,5 @@
 Feature: setup_marathon_job can create a "complete" app
-    
+
   Scenario: complete apps can be deployed
     Given a working paasta cluster
       And I have yelpsoa-configs for the marathon job "test-service.main"
@@ -20,29 +20,29 @@ Feature: setup_marathon_job can create a "complete" app
 
   Scenario: marathon apps can be scaled up with at-risk hosts
     Given a working paasta cluster
-     And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
+      And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
       And an old app to be destroyed with constraints [["hostname", "UNIQUE"]]
-    When there are 2 old healthy tasks
-    When setup_service is initiated
+     When there are 2 old healthy tasks
+     When setup_service is initiated
       And there are 2 new healthy tasks
-    When we mark a host it is running on as at-risk
-    When setup_service is initiated
+     When we mark a host it is running on as at-risk
+     When setup_service is initiated
       And there are 3 new healthy tasks
       And we wait a bit for the old app to disappear
-    Then the old app should be gone
+     Then the old app should be gone
       And there should be 0 tasks on that at-risk host
 
   Scenario: marathon apps can be deployed with at-risk hosts
     Given a working paasta cluster
       And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
-    When setup_service is initiated
+     When setup_service is initiated
       And there are 2 new healthy tasks
-    When we mark a host it is running on as at-risk
-    When setup_service is initiated
+     When we mark a host it is running on as at-risk
+     When setup_service is initiated
       And there are 3 new healthy tasks
-    When setup_service is initiated
+     When setup_service is initiated
       And there are 2 new healthy tasks
-    Then there should be 0 tasks on that at-risk host
+     Then there should be 0 tasks on that at-risk host
 
   Scenario: marathon apps can be scaled down
     Given a working paasta cluster
--- a/paasta_itests/steps/bounces_steps.py
+++ b/paasta_itests/steps/bounces_steps.py
@@ -36,6 +36,15 @@ def which_id(context, which):
 
 @given(u'a new {state} app to be deployed, with bounce strategy "{bounce_method}" and drain method "{drain_method}"')
 def given_a_new_app_to_be_deployed(context, state, bounce_method, drain_method):
+    given_a_new_app_to_be_deployed_constraints(context, state, bounce_method, drain_method, str(None))
+
+
+@given(u'a new {state} app to be deployed, ' +
+       u'with bounce strategy "{bounce_method}" ' +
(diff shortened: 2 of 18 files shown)


paasta_tools/setup_marathon_job.py
```diff
@@ -417,9 +418,16 @@ def log_deploy_error(errormsg, level='event'):
...
+        num_at_risk_tasks = 0
+        for host in hosts_tasks_running_on:
+            if host in draining_hosts:
+                num_at_risk_tasks += 1
+        log.debug("%s has %d tasks running on at-risk hosts." % (new_app.id, num_at_risk_tasks))
         protected_draining_tasks = set()
-        if new_app.instances < config['instances']:
-            client.scale_app(app_id=new_app.id, instances=config['instances'], force=True)
+        if new_app.instances < config['instances'] + num_at_risk_tasks:
+            client.scale_app(app_id=new_app.id, instances=config['instances'] + num_at_risk_tasks, force=True)
```
solarkennedy: Can you add a `_log` so when users are wondering why this happened it can be explained?

paasta_tools/setup_marathon_job.py
```diff
@@ -417,9 +418,16 @@ def log_deploy_error(errormsg, level='event'):
     )
 
     if new_app_running:
+        hosts_tasks_running_on = [task.host for task in new_app.tasks]
```
solarkennedy: 421:426 are pretty functional, can you make a smaller function to contain it?

solarkennedy:
This looks really simple, which I think is great!

Obviously it can't be "perfect", because there is the race between this script and a host dying of course. We'll need @huadongliu's bounce speed improvements to have a chance that this scaling can happen in a sane amount of time. (especially if we only get 2m on spot fleet.)

paasta_tools/setup_marathon_job.py
```diff
@@ -417,9 +418,16 @@ def log_deploy_error(errormsg, level='event'):
...
+        for host in hosts_tasks_running_on:
+            if host in draining_hosts:
+                num_at_risk_tasks += 1
+        log.debug("%s has %d tasks running on at-risk hosts." % (new_app.id, num_at_risk_tasks))
         protected_draining_tasks = set()
-        if new_app.instances < config['instances']:
-            client.scale_app(app_id=new_app.id, instances=config['instances'], force=True)
+        if new_app.instances < config['instances'] + num_at_risk_tasks:
+            client.scale_app(app_id=new_app.id, instances=config['instances'] + num_at_risk_tasks, force=True)
         elif new_app.instances > config['instances']:
```
EvanKrall: I thought it was a mistake at first that you hadn't made this `elif new_app.instances > config['instances'] + num_at_risk_tasks`, but then I realized that this is correct: once we scale up by enough, we want to start killing some instances to get off the at-risk hosts.

Can you add a comment to that effect?

paasta_tools/setup_marathon_job.py
```diff
@@ -417,9 +418,16 @@ def log_deploy_error(errormsg, level='event'):
...
             num_tasks_to_scale = max(min(len(new_app.tasks), new_app.instances) - config['instances'], 0)
             task_dict = get_old_happy_unhappy_draining_tasks_for_app(
```
EvanKrall: I think you need to add some code in the next few blocks of lines (433-456). This is the code that decides which of the tasks on the new app are going to be sent to the bounce function as "old" (e.g. eligible for killing).

We need to prefer tasks that are on at-risk hosts before other healthy tasks. Currently the order is draining > unhappy > happy; I think it needs to be draining > unhappy > at-risk > happy. Put them in `old_app_live_happy_tasks[new_app.id]`.

EvanKrall: Would be worth having an integration test for this:

```
Given a working cluster with 3 slaves
  And an app with 2 tasks and hostname unique constraint

 When we mark one of the slaves that our app is running on as at-risk
  And deploy_service with bounce strategy "crossover" and drain method "noop" is initiated

 Then the task on the at-risk node should be drained
```

EvanKrall:
Looks good so far. I think if you make the addition I suggested, this would be ready to go.

paasta_tools/setup_marathon_job.py
nhandler: I have a `log.debug` right above this snippet that outputs the number of tasks running on at-risk hosts. I will add another `log.debug`. within the `if` statement to note that we are scaling the application.

paasta_tools/setup_marathon_job.py
nhandler: Can do. I still need to add logic to drain the at-risk tasks (rather than letting it choose a potentially healthy/not at-risk task to drain).

nhandler:
I think I've addressed all of the feedback.

I do want to ultimately switch this code to use `docker-compose scale`, but I need to spend some more time to make all of the tests happy without the presence of the hostname on the mesosslave container.

paasta_tools/setup_marathon_job.py
```diff
@@ -294,49 +304,74 @@ def log_bounce_action(line, level='debug'):
...
         old_app_live_happy_tasks[app.id] = tasks_by_state['happy']
         old_app_live_unhappy_tasks[app.id] = tasks_by_state['unhappy']
         old_app_draining_tasks[app.id] = tasks_by_state['draining']
+        old_app_at_risk_tasks[app.id] = tasks_by_state['at_risk']
+
+    return old_app_live_happy_tasks, old_app_live_unhappy_tasks, old_app_draining_tasks, old_app_at_risk_tasks
 
-    return old_app_live_happy_tasks, old_app_live_unhappy_tasks, old_app_draining_tasks
+
+def get_num_at_risk_tasks(app):
```
Rob-Johnson: how about moving this to ``marathon_tools``?

Rob-Johnson:
just a thought - we can't guarantee that when we scale up, Marathon won't schedule the new instances on those hosts in maintenance mode can we? We could get in the situation where say:

- slave A with 3 instances of a task running on it
- I mark slave A as 'draining'
- setup marathon jobs scales instance count up by 3 (to 6)
- marathon schedules 3 new nodes on slave A
- setup marathon job scales instance count up by 6 (to 9)
...

I mean it *shouldn't* happen, but something worth considering

nhandler:
@Rob-Johnson That is item 2i in #376, which should be mitigated by item 3. I am fine holding off on shipping this change until I get item 3 completed, but I would like to get this PR in a 'Ship It' state so I can branch the item 3 branch off of this one.

EvanKrall:
@Rob-Johnson That state is fine -- eventually, the at-risk boxes would run out of space, and then we'd manage to launch 3 instances on not-at-risk boxes. Then when the at-risk box goes away, we'll be down to the correct replication. The only real downside is that draining a box will take longer, and it might be a bit confusing to mark a box as at-risk and end up with *more* tasks on that box.

The dynamic reservations thing will help avoid that, which will help draining boxes go faster.

paasta_itests/setup_marathon_job.feature
```diff
@@ -18,6 +18,19 @@ Feature: setup_marathon_job can create a "complete" app
       And we run setup_marathon_job until it has 5 task(s)
      Then we should see the number of instances become 5
 
+  Scenario: marathon apps can be scaled up with at-risk hosts
+    Given a working paasta cluster
+      And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
+      And an old app to be destroyed with constraints [["hostname", "UNIQUE"]]
+    When there are 2 old healthy tasks
```
EvanKrall: can you match indentation from above?

paasta_itests/steps/bounces_steps.py
```diff
@@ -36,6 +36,15 @@ def which_id(context, which):
 
 @given(u'a new {state} app to be deployed, with bounce strategy "{bounce_method}" and drain method "{drain_method}"')
 def given_a_new_app_to_be_deployed(context, state, bounce_method, drain_method):
+    given_a_new_app_to_be_deployed_constraints(context, state, bounce_method, drain_method, str(None))
```
EvanKrall: str(None)?

paasta_itests/setup_marathon_job.feature
```diff
@@ -18,6 +18,19 @@ Feature: setup_marathon_job can create a "complete" app
...
+     And we mark the host "mesosslave" as at-risk
+     And setup_service is initiated
+    When there are 2 new healthy tasks
+     And setup_service is initiated
+     And we wait a bit for the old app to disappear
+    Then the old app should be gone
+     And the tasks on the host "mesosslave" should be drained
```
EvanKrall: I'd like to see another integration test that checks that this all works with only a new app.

paasta_itests/setup_marathon_job.feature
```diff
@@ -18,6 +18,19 @@ Feature: setup_marathon_job can create a "complete" app
       And we run setup_marathon_job until it has 5 task(s)
      Then we should see the number of instances become 5
 
+  Scenario: marathon apps can be scaled up with at-risk hosts
+    Given a working paasta cluster
+      And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
+      And an old app to be destroyed with constraints [["hostname", "UNIQUE"]]
+    When there are 2 old healthy tasks
+     And we mark the host "mesosslave" as at-risk
```
EvanKrall: There's no guarantee that there's gonna be a task on "mesosslave", since there are 2 instances and 3 slaves, right?

Can we have this step pick one of the hosts where this app is actually running?

paasta_itests/steps/setup_marathon_job_steps.py
```diff
@@ -110,3 +123,33 @@ def can_run_get_app(context):
...
+        mock.patch.object(mesos.cli.master, 'CFG', config),
+    ) as (
+        mock_load_credentials,
+        _,
+    ):
+        mock_load_credentials.side_effect = paasta_maintenance.load_credentials(mesos_secrets='/etc/mesos-slave-secret')
+        paasta_maintenance.drain([host], start, duration)
+
+
+@then(u'the tasks on the host "{host}" should be drained')
```
EvanKrall: Draining tasks has another connotation; can we instead call this `there should be no tasks on the host "{host}"`?

paasta_itests/setup_marathon_job.feature
EvanKrall: I'm not convinced this test should pass consistently. This step is checking that there are no tasks on the `mesosslave` host, which doesn't match my understanding of what setup_marathon_job should do, at least not during the first run. If it's passing, I bet it's either because:
 - we're running setup_service twice after mesosslave is marked at-risk (first one scales up, second one scales down), and the steps in between are taking long enough that the scale up/down is finishing.
 - or because none of the original tasks were on mesosslave to begin with.

Let's do my suggestion where we pick one of the boxes that's at-risk, then after the next setup_service call, add `And there are 3 new healthy tasks`. Then, once we've run setup_service again, this last step asserting that mesosslave (or whatever we marked at-risk) is empty should pass.

paasta_itests/steps/setup_marathon_job_steps.py
nhandler: Good point. How about we make it a bit more flexible and go with, `there should be {number:d} tasks on the host "{host}"`?

paasta_tools/graceful_app_drain.py
```diff
@@ -73,7 +73,8 @@ def main():
         (old_app_live_happy_tasks,
          old_app_live_unhappy_tasks,
          old_app_draining_tasks,
-         ) = get_old_happy_unhappy_draining_tasks(
+         old_app_at_risk_tasks,
```
Rob-Johnson: is this right? I think the diff is saying that both line 75 and 76 are now the same?

nhandler: No, it is saying that line 75 is `old_app_draining_tasks` and line 76 is `old_app_at_risk_tasks` (note the `draining` vs `at_risk`). Subtle difference.

paasta_tools/setup_marathon_job.py
Rob-Johnson: bump

nhandler: Yep, still working on this. It is a bit annoying due to having to mock out a bunch of the paasta_maintenance stuff (due to the dependency on mesos.cli's config). This is what led me to discover that we can't safely run our itests individually (due to not properly setting up and cleaning up), and I got a bit distracted trying to sort that out. But thanks for the reminder :)

paasta_tools/graceful_app_drain.py
Rob-Johnson: ha, I shoudn't bother when I'm tired :)

paasta_itests/steps/bounces_steps.py
nhandler: Forgot about this bit. This value gets eval'ed further down. The reason for this is that behave doesn't support an array type parameter in a step definition. So we read in the array as a string and eval it back to the proper type. The `str(None)` is really `eval(str(None))`. `eval()` gets upset if you simply pass it an empty string.

solarkennedy:
lgtm

paasta_tools/paasta_maintenance.py
```diff
@@ -131,18 +135,26 @@ def get_maintenance_status():
...
+    client_fn = maintenance_api()
     return client_fn(method="GET", endpoint="/status")
 
 
+def schedule():
+    """Get the Mesos maintenance schedule. This contains hostname/ip mappings and their maintenance window.
+    :returns: None
+    """
+    schedule = get_maintenance_schedule()
+    print "%s:%s" % (schedule, schedule.text)
```
Rob-Johnson: do you need to do some error handling here to check the response is reasonable?

paasta_tools/paasta_maintenance.py
```diff
@@ -131,18 +135,26 @@ def get_maintenance_status():
...
+
 def get_hosts_with_state(state):
     """Helper function to check the maintenance status and return all hosts
     listed as being in a current state
 
     :param state: State we are interested in ('down_machines' or 'draining_machines')
     :returns: A list of hostnames in the specified state or an empty list if no machines
     """
-    status = get_maintenance_status()
+    status = get_maintenance_status().json()
```
Rob-Johnson: again - maybe need to do some error handling? the api client throws an ``HTTPError`` when things go wrong doesn't it?

Rob-Johnson:
fix and ship!

paasta_itests/setup_marathon_job.feature
```diff
@@ -18,6 +18,32 @@ Feature: setup_marathon_job can create a "complete" app
       And we run setup_marathon_job until it has 5 task(s)
      Then we should see the number of instances become 5
 
+  Scenario: marathon apps can be scaled up with at-risk hosts
+    Given a working paasta cluster
+     And a new healthy app to be deployed, with bounce strategy "crossover" and drain method "noop" and constraints [["hostname", "UNIQUE"]]
+      And an old app to be destroyed with constraints [["hostname", "UNIQUE"]]
+    When there are 2 old healthy tasks
```
EvanKrall: I think it's odd that this test involves an old application - I'd prefer we had a test with only a new application. However, I understand the itest steps weren't cooperating, so I don't consider this a ship blocker.

EvanKrall: Actually, it looks like by just removing the lines mentioning the old app, this scenario still passes. Will push a commit, one sec.

EvanKrall: Oh, I'm looking at the wrong itest.... The other one doesn't mention old tasks.

Ship!

solarkennedy:
Oh the itests :(

solarkennedy:
Well I got a green build... do I dare merge...

EvanKrall:
do it!

Rob-Johnson:
shipit...monday?

solarkennedy:
I'll ship monday with a pin.

PR #522, opened 2016-05-12
Dont try to do mesos leader detection in any script

solarkennedy:
This will move the complexity of our adhoc leader-locking to the caller (cron job).

It has a nice side effect of not requiring an operator to run these scripts on the current leader.


Rob-Johnson:
shipit

EvanKrall:
Ship carefully.

PR #523, opened 2016-05-13
make the command the last arg to serviceinit

Rob-Johnson:
The commit https://github.com/Yelp/paasta/commit/71280ede5558a75dba28b6cdbc572495294e5045#diff-7d4e89c64360b719f36a1ca16d58607cL505 changed the ordering of the args passed to paasta serviceinit. Internally, some of the sudoers rules we have in place broke because they relied on this ordering. This reverts that change, moving back to the previous layout.

This change also changes the way in which the serviceinit command is created so that whitespace arrangement is sorted in one place.

@giuliano108 primary - this fixes internal ticket PAASTA-4457


giuliano108:
Looks good, ShipIt.

PR #526, opened 2016-05-17
Hardcode a python2.7 dep to be compatible with all ubuntu

### LINKED ISSUE #477: Deb packages don't list python2.7 dependency
Although we don't require much in the way of system python packages, we still require a python2.7 interpreter to be installed. Our debs should probably list python2.7 as a dependency.

solarkennedy:
Fixes #477 

The dh-virtualenv docs recommend ${Python:Depends}, but this is not going to work out great for all ubuntu's, so I chose to hard-code it instead.


EvanKrall:
🐑

PR #527, opened 2016-05-17
Use named tuples in paasta metastatus

Rob-Johnson:
There isn't really a ticket associated with this - it's just some cleanup ahead of the work I'm doing for internal PAASTA-4188

- Changes the tuples we were previously using to a real named tuple 'HealthCheckResult' so that there are less magic array indexes dotted around the code
- Move the logic for formatting the output from 'unhealthy' health checks to a single formatting function
- Normalise the capitalisation of some of the output that's printed.


Rob-Johnson:
argh, fixing tests

solarkennedy:
shipit. I think this is cool and I like how you are able to consolidate the color logic.

PR #532, opened 2016-05-18
Specify schema/port in paasta_maintenance

nhandler:
get_mesos_leader only returns the hostname. We need an actual URL for paasta_maintenance.

It looks like this got dropped during one of the refactors. Without this change, paasta_maintenance throws an error.


solarkennedy:
ship

PR #534, opened 2016-05-20
Allow crossover to make progress in the face of failures

oktopuz:
internal ticket: PAASTA-3304

tox tests/test_bounce_lib.py succeeded

### DIFF
--- a/paasta_tools/bounce_lib.py
+++ b/paasta_tools/bounce_lib.py
@@ -350,29 +350,25 @@ def crossover_bounce(
     See the docstring for brutal_bounce() for parameters and return value.
     """
 
-    assert margin_factor >= 1
+    assert margin_factor > 0
+    assert margin_factor <= 1
 
-    if not new_app_running:
-        return {
-            "create_app": True,
-            "tasks_to_drain": set(),
-        }
-    else:
-        happy_count = int(len(happy_new_tasks) * margin_factor)
-        needed_count = max(new_config['instances'] - happy_count, 0)
+    needed_count = max(int(new_config['instances'] * margin_factor) -
+                       len(happy_new_tasks), 0)
 
-        old_tasks = []
-        for app, tasks in old_app_live_happy_tasks.items():
-            for task in tasks:
-                old_tasks.append(task)
+    old_tasks = []
+    for app, tasks in old_app_live_happy_tasks.items():
+        for task in tasks:
+            old_tasks.append(task)
 
-        for app, tasks in old_app_live_unhappy_tasks.items():
-            for task in tasks:
-                old_tasks.append(task)
-        return {
-            "create_app": False,
-            "tasks_to_drain": set(set(old_tasks[needed_count:])),
-        }
+    for app, tasks in old_app_live_unhappy_tasks.items():
+        for task in tasks:
+            old_tasks.append(task)
+
+    return {
+        "create_app": not new_app_running,
+        "tasks_to_drain": set(old_tasks[needed_count:]),
+    }
 
 
 @register_bounce_method('downthenup')
--- a/paasta_tools/bounce_lib.py
+++ b/paasta_tools/bounce_lib.py
@@ -339,8 +339,8 @@ def crossover_bounce(
     happy_new_tasks,
     old_app_live_happy_tasks,
     old_app_live_unhappy_tasks,
-    margin_factor=1,
-    minimum_health_capacity=0,
+    margin_factor=1.05,
+    minimum_health_capacity=5,
 ):
     """Starts a new app if necessary; slowly kills old apps as instances of the new app become happy.
 
--- a/paasta_tools/bounce_lib.py
+++ b/paasta_tools/bounce_lib.py
@@ -340,18 +340,15 @@ def crossover_bounce(
     old_app_live_happy_tasks,
     old_app_live_unhappy_tasks,
     margin_factor=1.05,
-    minimum_health_capacity=5,
 ):
     """Starts a new app if necessary; slowly kills old apps as instances of the new app become happy.
 
     :param margin_factor: the multiplication factor used to calculate the number of instances to be drained when minimum_health_capacity is reached.
-    :param minimum_health_capacity: the number of happy_new_tasks that allows applying margin_factor
 
     See the docstring for brutal_bounce() for the rest of parameters and return value.
     """
 
     assert margin_factor >= 1
-    assert minimum_health_capacity >= 0
 
     if not new_app_running:
         return {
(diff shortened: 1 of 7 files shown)


paasta_tools/bounce_lib.py
```diff
@@ -339,19 +339,29 @@ def crossover_bounce(
     happy_new_tasks,
     old_app_live_happy_tasks,
     old_app_live_unhappy_tasks,
+    margin_factor=1,
```
solarkennedy: These seem like strange defaults.

How about 1.05 and 1 ?

paasta_tools/bounce_lib.py
```diff
@@ -339,19 +339,29 @@ def crossover_bounce(
...
+    assert margin_factor >= 1
+    assert minimum_health_capacity >= 0
+
     if not new_app_running:
         return {
             "create_app": True,
             "tasks_to_drain": set(),
         }
     else:
         happy_count = len(happy_new_tasks)
```
solarkennedy: What if this was just

    floor(len(happy) * margin)

?

oktopuz: then we don't need minimum_health_capacity

paasta_tools/bounce_lib.py
oktopuz: replaced by 1.05

paasta_tools/bounce_lib.py
```diff
@@ -339,19 +339,24 @@ def crossover_bounce(
...
+
     if not new_app_running:
         return {
             "create_app": True,
             "tasks_to_drain": set(),
         }
     else:
-        happy_count = len(happy_new_tasks)
+        happy_count = int(len(happy_new_tasks) * margin_factor)
         needed_count = max(new_config['instances'] - happy_count, 0)
```
EvanKrall: I would apply the margin on new_config['instances'] when calculating needed_count:

```
        needed_count = max((new_config['instances'] * margin_factor) - happy_count, 0)
```

and make margin_factor something between 0 and 1.

paasta_tools/bounce_lib.py
```diff
@@ -339,19 +339,24 @@ def crossover_bounce(
     happy_new_tasks,
     old_app_live_happy_tasks,
     old_app_live_unhappy_tasks,
+    margin_factor=1.05,
```
EvanKrall: I know it increases the scope of the patch, but this really needs to be configurable via MarathonServiceConfig.

paasta_tools/bounce_lib.py
oktopuz: Applying it to new_config['instances'] looks too aggressive.
When happy_count = 0 we gonna drain 5% of instances anyway.

Moreover if new_app is broken, this will eventually drain the number of old_app instances to 0.

paasta_tools/bounce_lib.py
solarkennedy: @oktopuz can you make this a param? I would be happy to pair with you to help hint at how to do that.

oktopuz:
Made it configurable via MarathonServiceConfig.

solarkennedy:
This is looking like a really comprehensive PR! Thanks also for writing the test case.
The last thing is can you add some docs to the marathon section of the soa-configs doc page to explain what this input does for humans?

I give a shipit, I would like @EvanKrall to be primary as he is the local resident on the marathon bounce code and he asked to be reminded to think of the math parts. (pretty artificially inflating the happy_count achieves the desired effect)

paasta_tools/bounce_lib.py
EvanKrall: > When happy_count = 0 we gonna drain 5% of instances anyway.

I think that's okay, and desirable in some cases, such as if the cluster is very full.

> Moreover if new_app is broken, this will eventually drain the number of old_app instances to 0.

I don't think it would. `new_config['instances']` is based on the `instances` key in marathon.yaml or what the autoscaler decides, and isn't based on the number of running instances.

paasta_tools/bounce_lib.py
```diff
@@ -375,6 +378,7 @@ def downthenup_bounce(
     happy_new_tasks,
     old_app_live_happy_tasks,
     old_app_live_unhappy_tasks,
+    margin_factor=1,
```
EvanKrall: set the default to 1.0, just to make it obvious that it's intended to be a float

tests/test_bounce_lib.py
```diff
@@ -601,15 +601,15 @@ def test_crossover_bounce_no_existing_apps(self):
             "tasks_to_drain": set(),
         }
 
-    def test_crossover_bounce_old_but_no_new(self):
+    def test_crossover_bounce_old_app_but_no_new_app(self):
         """When marathon only has old apps for this service, crossover bounce should start the new one, but not kill any
         old tasks yet."""
 
-        new_config = {'id': 'foo.bar.12345', 'instances': 5}
+        new_config = {'id': 'foo.bar.12345', 'instances': 100}
```
EvanKrall: Why make this change?

paasta_tools/bounce_lib.py
oktopuz: replaced '1' by '1.0'

tests/test_bounce_lib.py
```diff
@@ -835,6 +861,29 @@ def test_crossover_bounce_mid_bounce_no_happy_old_lots_of_unhappy_old(self):
...
+        actual = bounce_lib.crossover_bounce(
+            new_config=new_config,
+            new_app_running=True,
+            happy_new_tasks=happy_tasks,
+            old_app_live_happy_tasks=old_app_live_happy_tasks,
+            old_app_live_unhappy_tasks=old_app_live_unhappy_tasks,
+            margin_factor=0.95,
+        )
+        assert actual['create_app'] is False
+        assert len(actual['tasks_to_drain']) == 125
```
EvanKrall: Hm. I don't think this should drain anything, since there are only 400 healthy tasks and we've asked for 500 total. It should only drain anything if the number of healthy tasks is more than 475 = 0.95*500.

tests/test_bounce_lib.py
oktopuz: reverted the change.
I wanted to modify the test, but then I created a new one and forgot to revert my changes here.

tests/test_bounce_lib.py
EvanKrall: Oh, I'm wrong, this is correct. This would drain 125 unhappy tasks, leaving 300 happy tasks and 75 unhappy tasks.

EvanKrall:
Looks good to me.

tests/test_bounce_lib.py
oktopuz: In the test we have 400 healthy tasks + 200 unhealthy tasks (i.e. 600) while we requested 500 instances only.
i.e. 600 - 125 = 475

However I realize that the number of happy+unhappy tasks can't be bigger than the number of instances.
I've set 100 unhappy tasks, and now it drains 25.

PR #535, opened 2016-05-20
continue serviceinit when an exception occurs on one service.instance

huadongliu:
An exception happened on one of the lucy instances and "paasta status" terminated.

Patch was tested simply with ".tox/py27/bin/paasta_serviceinit.py status -s robj-test-service -i main,autoscale_to_heck,memo-test". The "permission denied" exception occurs three times instead of once without the patch.


solarkennedy:
ship

PR #537, opened 2016-05-21
Enabling checking of disabled chronos jobs

solarkennedy:
When `check_chronos_jobs` runs and iterates over all the jobs, it has logic to understand to not alert on disabled jobs.

However we filter out those disabled jobs in the first place, which means we get a kind of bogus alert:

    foo isn't in chronos at all, which means it may not be deployed yet

But it is in chronos, it is just disabled. This code makes it so check_chronos_jobs includes disabled jobs, so they can hit the "it is ok, it is disabled" logic later on.


Rob-Johnson:
shipit

PR #539, opened 2016-05-23
Document the system paasta config options

EvanKrall:
In this branch I've documented most currently-existing SystemPaastaConfig options. (I didn't include `cluster_autoscaling_resources` because I'm not sure what to say about that.)

I've also spruced up the getting started docs.

These are changes that I think would be useful for anybody setting up a Paasta cluster for the first time, guided by my recent experience setting up a cluster for Eat24.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -43,6 +43,7 @@ Installing The PaaSTA Infrastructure
 
    installation/getting_started
    soa_configs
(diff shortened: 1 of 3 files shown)


EvanKrall:
I suppose I should link this from somewhere.

solarkennedy:
ship

PR #540, opened 2016-05-24
disabling requests_cache when creating/deleting a marathon app

huadongliu:
Marathon client uses requests to list_apps. It should be get cached. Otherwise, create_marathon_app will timeout since it cannot confirm the new app has been created.


solarkennedy:
Cache is cool though, how about just breaking the cache instead for functions that need to bypass it?
http://requests-cache.readthedocs.io/en/latest/api.html#requests_cache.backends.base.BaseCache.delete_url

huadongliu:
http://requests-cache.readthedocs.io/en/latest/user_guide.html#usage

"Also, you can use requests_cache.disabled() context manager for temporary disabling caching:"

solarkennedy:
ship

PR #542, opened 2016-05-26
Use the force=True flag to kill marathon tasks

huadongliu:
Using this flag will resolve the HTTP 409 exception when killing marathon tasks. Detailed root cause analysis is in paasta-4156.

Unit test was to reboot robj-test-service.memo-test and run sudo .tox/py27/bin/setup_marathon_job.py robj-test-service.memo-test to make sure the change is exercised.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -1054,7 +1054,7 @@ def kill_task(client, app_id, task_id, scale):
 def kill_given_tasks(client, task_ids, scale):
     """Wrapper to the official kill_given_tasks method that is tolerant of errors"""
     try:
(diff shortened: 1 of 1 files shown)


Rob-Johnson:
shipit I guess, but I don't really have enough context to be confident of the side effects this might cause.

solarkennedy:
Yea, I am also unsure of the side effects and the marathon docs are unsure :(

Please pin and let this sit/bake, as it may take a few days for issues to appear.

huadongliu:
Unfortunately the HTTP 409 exception only happens in prod. I wasn't able to reproduce it elsewhere. For this exception to happen, apps to be deployed have to stack up, i.e. 2+ old apps.

PR #543, opened 2016-05-26
PAASTA-4188 per attribute alerting

Rob-Johnson:
closes internal PAASTA-4188
 
I've got this up for visibility, but I'm still not 100% happy with it so won't deploy just yet - I've got to move a bunch of stuff out of ``main`` and bump the test coverage.

The user facing changes are:

- added a ``-g`` flag, allowing you to pick arbitrary attributes to group on. We'll probably update our monitoring to do something like ``/usr/bin/paasta_metastatus -vvv -g superregion region habitat pool``.
What I'd really like to do next is support grouping by multiple attributes, so that you could do ``-g region,pool`` and get stats for default/region, lucy/region.

- added a ``-t`` flag, that allows the user to pick the level at which things are considered 'critical'. This will apply to each metric, so if any of disk/cpu/mem are above threshold for a given grouping, this will exit 1. Note that this does *not* apply to hosts, where we don't really care if they're at 100%.

- on increasing the verbosity level, the output is appended to, rather than modified. Currently if you chose `-vvv`, you get info about slaves, and lose the attribute level info you get in `-vv`. This changes that, so that `-vvv` just *appends* to the output shown in `-vv`.

Internally (as you can see by the diff) there is quite a lot of refactoring done here. Particularly, I've been keen to try and reduce the number of ``print_*`` functions and keep all the printing logic in ``main``.

Here's the output on mesosstage:
 
-vv
image

-vvv
screen shot 2016-05-26 at 10 44 36


paasta_tools/paasta_metastatus.py
```diff
@@ -271,106 +263,157 @@ def assert_quorum_size(state):
...
+
+    :param slaves: a list of slave objects as defined in mesos state.
+    <https://mesos.apache.org/documentation/latest/endpoints/master/state.json/>``.
+    :param attribute: the attribute to group slaves by. This attribute must be common to all slaves.
+    :returns: groupby object, with slaves grouped by the value of the attribute
+    """
+    def key_func(slave):
+        return slave['attributes'].get(attribute, 'unknown')
+    sorted_slaves = sorted(slaves, key=key_func)
+    return itertools.groupby(sorted_slaves, key=key_func)
```
solarkennedy: nice

paasta_tools/paasta_metastatus.py
```diff
@@ -271,106 +263,157 @@ def assert_quorum_size(state):
...
+        ResourceUtilization(metric=field, total=total[index], free=free[index])
+        for index, field in enumerate(ResourceInfo._fields)
+    ]
 
-    cluster_results = run_healthchecks_with_param(mesos_state, [assert_quorum_size, assert_no_duplicate_frameworks])
 
-    metrics = get_mesos_stats()
-    metrics_results = run_healthchecks_with_param(metrics, [
+def has_registered_slaves(mesos_state):
+    return 'slaves' in mesos_state and mesos_state['slaves']
```
solarkennedy: This reads odd to me. Just a bikeshed, but how about

    len(mesos_state.get('slaves', []) > 0

paasta_tools/paasta_metastatus.py
```diff
@@ -514,19 +628,114 @@ def main():
...
+                ))
+            table_rows = sorted(table_rows, key=lambda x: x[0])
+            all_rows.extend(table_rows)
+            for line in format_table(all_rows):
+                print_with_indent(line, 4)
+        print marathon_summary
+        print_results_for_healthchecks(marathon_ok, marathon_results, args.verbose)
+        print chronos_summary
+        print_results_for_healthchecks(chronos_ok, chronos_results, args.verbose)
+    else:
```
solarkennedy: This would be vvv right?

solarkennedy:
The example print output looks really super. Shipit. 

I like the refactoring bits, I agree the "main" is getting pretty long. I'm not *that* scared, if you think it would help make it more understandable, you could do it like this:

```
# Mesos stuff
print mesos_summary
if verbose == 1:
  print more mesos stuff
elif verbose == 2:
  print even more stuff
elif verbose >= 3:
  print even more

# marathon stuff
print marathon_summary
if verbose == 1:
  ...
...
```

It makes it longer, but might help make it more readable because of the grouping? Not a real issue, just throwing it out there.

paasta_tools/paasta_metastatus.py
Rob-Johnson: yeah, and anything above.

paasta_tools/autoscaling_lib.py
```diff
@@ -455,18 +455,18 @@ def slave_pid_to_ip(slave_pid):
...
+        lambda slave: slave['attributes']['pool'],
+        mesos_state
+    )[pool]
+
+    free_pool_resources = pool_utilization_dict['free']
+    total_pool_resources = pool_utilization_dict['total']
+    utilization = 1.0 - min([
+        float(free_pool_resources[resource]) / total_pool_resources[resource]
+        for resource in free_pool_resources
+    ])
```
Rob-Johnson:
```
utilization = 1.0 - min([float(pair[0]/pair[1]) for pair in zip(free_pool_resources, total_pool_resources)])
```

PR #548, opened 2016-05-27
Document config blacklist automatically

nhandler:
This is a new attempt at https://github.com/Yelp/paasta/pull/470 . We had to revert the change in https://github.com/Yelp/paasta/commit/ca027d7015cced3214df90a0950e3e8e41572508 due to not having the `sphinxcontrib.autorun` package available on readthedocs. It looks like we can add a `readthedocs.yml` file to configure our project. This change adds such a file and moves our sphinx dependencies into requirements-dev.txt, which will then get installed by pip on readthedocs.

Due to https://github.com/rtfd/readthedocs.org/issues/1112 , we can only specify one requirementst.txt file to be installed on readthedocs. To work around this, I adjusted our requirements-dev.txt file to include requirements.txt (and adjusted tox).


solarkennedy:
Neato! Ship!

nhandler:
For future reference, this built cleanly on readthedocs: http://readthedocs.org/projects/paasta/builds/4045797/

PR #549, opened 2016-05-27
Support passing posargs through to paasta_itests

nhandler:
This change makes it possible to tag paasta_itests with `@wip` and then run them with `tox -e paasta_itests -- --tags=wip`. Previously, we were not passing the `{posargs}` properly down the stack.


solarkennedy:
ship

PR #550, opened 2016-05-27
remove leading slash from docker container name

huadongliu:
(Pdb) p container["Names"][0]
u'/mesos-34e6f338-dfa2-48e2-a798-40b0b0d318d5-S14.1ef2bcd1-0f65-41ae-b703-c2a1002d675c'

The leading slash needs to be removed.

### DIFF
--- a/paasta_tools/contrib/kill_orphaned_docker_containers.py
+++ b/paasta_tools/contrib/kill_orphaned_docker_containers.py
@@ -33,7 +33,7 @@ def main():
         mesos_task_id = mesos_tools.get_mesos_id_from_container(
             container=container, client=docker_client)
         if mesos_task_id not in running_mesos_task_ids:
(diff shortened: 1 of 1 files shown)


solarkennedy:
Ship.

Seems so strange that when you ask the docker api for the name it comes with a leading slash, but when you ask to kill a container it won't accept the leading slash?

Makes me think we are using `["Names"][0]` wrong.

solarkennedy:
Ship.

Seems so strange that when you ask the docker api for the name it comes with a leading slash, but when you ask to kill a container it won't accept the leading slash?

Makes me think we are using `["Names"][0]` wrong.

huadongliu:
Maybe you are right. Unfortunately I have killed all the reported orphaned containers. I will keep an eye on this. If another sensu ticket shows up, I will try ["Names"].

PR #551, opened 2016-05-27
Get rid of "special" config files

### LINKED ISSUE #515: Get rid of "special" files in /etc/paasta
Our strategy for loading system configs is generally to read files from /etc/paasta in SystemPaastaConfig, and a particular config key can exist in any file in this directory.

However, a few places in the codebase, we read files like /etc/paasta/marathon.json or /etc/paasta/chronos.json, and look for specific keys in those files, like `user`, `password`, and `url`.

We should read the marathon / chronos cluster information through SystemPaastaConfig, and make these files less special. This would entail changing the file format from 

```
{
  "user": "admin",
  "password": "something",
  "url": [
    "box1:5052",
    "box2:5052"
  ]
}
```

to

```
{
  "marathon_server": {
    "user": "admin",
    "password": "something",
    "url": [
      "box1:5052",
      "box2:5052"
    ]
  }
}
```

matthewbentley:
chronos.json, marathon.json, performance-check.json

I suspect I may have missed something(s), but all the unit and integration tests pass.

This will have to be paired with config changes in puppet too.

I'm not sure if I need to change the changelog at all? Or if there's any documentation that needs to be updated?

Closes #515

### DIFF
--- a/paasta_tools/paasta_metastatus.py
+++ b/paasta_tools/paasta_metastatus.py
@@ -26,7 +26,6 @@ from marathon.exceptions import MarathonError
 
 from paasta_tools import chronos_tools
 from paasta_tools import marathon_tools
-from paasta_tools.chronos_tools import ChronosNotConfigured
 from paasta_tools.chronos_tools import get_chronos_client
 from paasta_tools.chronos_tools import load_chronos_config
 from paasta_tools.marathon_tools import MarathonNotConfigured
@@ -39,6 +38,7 @@ from paasta_tools.mesos_tools import get_zookeeper_config
 from paasta_tools.mesos_tools import MasterNotAvailableException
 from paasta_tools.utils import format_table
 from paasta_tools.utils import PaastaColors
+from paasta_tools.utils import PaastaNotConfiguredError
 from paasta_tools.utils import print_with_indent
 
 HealthCheckResult = namedtuple('HealthCheckResult', ['message', 'healthy'])
@@ -611,7 +611,7 @@ def main():
     # Check to see if Chronos should be running here by checking for config
     try:
         chronos_config = load_chronos_config()
-    except ChronosNotConfigured:
+    except PaastaNotConfiguredError:
         chronos_results = [HealthCheckResult(message='Chronos is not configured to run here', healthy=True)]
 
     if marathon_config:
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -923,7 +923,7 @@ class SystemPaastaConfig(dict):
             return self['performance_check']
         except KeyError:
             raise PaastaNotConfiguredError(
-                'Could not find performance_check in configureation directory: %s' % self.directory
+                'Could not find performance_check in configuration directory: %s' % self.directory
             )
 
     def get_marathon_config(self):
@@ -934,7 +934,7 @@ class SystemPaastaConfig(dict):
             return self['marathon_config']
         except KeyError:
             raise PaastaNotConfiguredError(
-                'Could not find marathon_config in configureation directory: %s' % self.directory
+                'Could not find marathon_config in configuration directory: %s' % self.directory
             )
 
 
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -911,9 +911,7 @@ class SystemPaastaConfig(dict):
         try:
             return self['chronos_config']
         except KeyError:
-            raise PaastaNotConfiguredError(
-                'Could not find chronos_config in configuration directory: %s' % self.directory
-            )
+            return {}
 
     def get_performance_check_config(self):
         """Get the performance check config
@@ -922,9 +920,7 @@ class SystemPaastaConfig(dict):
         try:
             return self['performance_check']
         except KeyError:
-            raise PaastaNotConfiguredError(
-                'Could not find performance_check in configuration directory: %s' % self.directory
-            )
+            return {}
 
     def get_marathon_config(self):
         """Get the marathon config
@@ -933,9 +929,7 @@ class SystemPaastaConfig(dict):
         try:
             return self['marathon_config']
         except KeyError:
-            raise PaastaNotConfiguredError(
-                'Could not find marathon_config in configuration directory: %s' % self.directory
-            )
+            return {}
 
 
 def _run(command, env=os.environ, timeout=None, log=False, stream=False, stdin=None, **kwargs):
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -926,6 +926,17 @@ class SystemPaastaConfig(dict):
                 'Could not find performance_check in configureation directory: %s' % self.directory
             )
 
+    def get_marathon_config(self):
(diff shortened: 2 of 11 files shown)


tests/test_marathon_tools.py
```diff
@@ -241,28 +241,28 @@ def test_read_service_config_and_deployments(self):
 
     def test_load_marathon_config(self):
         expected = {'foo': 'bar'}
+        from_file = {'marathon_config': {'foo': 'bar'}}
         file_mock = mock.MagicMock(spec=file)
         with contextlib.nested(
-            mock.patch('paasta_tools.marathon_tools.open', create=True, return_value=file_mock),
-            mock.patch('json.load', autospec=True, return_value=expected)
+            mock.patch('paasta_tools.utils.open', create=True, return_value=file_mock),
```
solarkennedy: In my opinion we don't need to have these tests anymore. We can let load_system_paasta_config worry about that

paasta_tools/paasta_metastatus.py
```diff
@@ -495,7 +495,7 @@ def main():
     # Check to see if Chronos should be running here by checking for config
     try:
         chronos_config = load_chronos_config()
-    except ChronosNotConfigured:
+    except PaastaNotConfiguredError:
```
solarkennedy: Is this right? This exception shouldn't change

paasta_tools/utils.py
```diff
@@ -857,6 +857,39 @@ def get_cluster_fqdn_format(self):
...
         return self.get('cluster_fqdn_format', 'paasta-{cluster:s}.yelp')
 
+    def get_chronos_config(self):
+        """Get the chronos config
+
+        :returns: The chronos config dictionary"""
+        try:
+            return self['chronos_config']
+        except KeyError:
+            raise PaastaNotConfiguredError(
```
solarkennedy: Oh. Well seems like... maybe this should just return {} and let the more specific getters raise their more specific exceptinos?

paasta_tools/utils.py
```diff
@@ -857,6 +857,39 @@ def get_cluster_fqdn_format(self):
...
+
+    def get_performance_check_config(self):
+        """Get the performance check config
+
+        :returns: The performance_check config dictionary"""
+        try:
+            return self['performance_check']
+        except KeyError:
+            raise PaastaNotConfiguredError(
+                'Could not find performance_check in configureation directory: %s' % self.directory
```
solarkennedy: configureation typo on here and line 890

paasta_tools/utils.py
```diff
@@ -857,6 +857,39 @@ def get_cluster_fqdn_format(self):
...
+                'Could not find chronos_config in configuration directory: %s' % self.directory
+            )
+
+    def get_performance_check_config(self):
+        """Get the performance check config
+
+        :returns: The performance_check config dictionary"""
+        try:
+            return self['performance_check']
+        except KeyError:
```
solarkennedy: I don't think it' is that crazy for these things to instead return "defaults", which may be just an empty dict, and when the code that needs to load a key and really needs it, let *it* complain?

paasta_tools/paasta_metastatus.py
matthewbentley: The exception raised by `load_chronos_config` changed.  `load_chronos_config` calls `get_chronos_config` in `SystemPaastaConfig`, which raises a `PaastaNotConfiguredError`.

`load_chronos_config` could catch the `PaastaNotConfiguredError` and re-raise a `ChronosNotConfiguredError`, which might make more sense.

paasta_tools/utils.py
matthewbentley: I was just following what other `get_x` functions were doing in `SystemPaastaConfig`.

nhandler:
Other than the outstanding issues, this looks pretty good to me.

tests/test_marathon_tools.py
matthewbentley: To be clear, are you suggesting removing `test_load_marathon_config` and `test_load_marathon_config_path_dne`, and similar with chronos? I just want to make sure I don't remove the wrong tests

And should I add something to test_utils to make up for it?

matthewbentley:
Fixed all the issues, except the potentially unneeded tests.

tests/test_marathon_tools.py
EvanKrall: If there isn't already a test for `SystemPaastaConfig.get_marathon_config`, it's probably worth adding one or two; similarly for the other things you've added. Test the case where it exists and where it doesn't exist.

tests/test_chronos_tools.py
```diff
@@ -93,50 +93,64 @@ def test_chronos_config_object_no_user(self):
...
         file_mock = mock.MagicMock(spec=file)
         with contextlib.nested(
-            mock.patch('paasta_tools.chronos_tools.open', create=True, return_value=file_mock),
-            mock.patch('json.load', autospec=True, return_value=expected)
+            mock.patch('paasta_tools.utils.open', create=True, return_value=file_mock),
+            mock.patch('json.load', autospec=True, return_value=from_file),
+            mock.patch('os.path.isdir', autospec=True, return_value=True),
+            mock.patch('os.access', autospec=True, return_value=True),
+            mock.patch('paasta_tools.utils.get_readable_files_in_glob', autospec=True,
+                       return_value=['/some/fake/dir/some_file.json']),
```
EvanKrall: I think I'd just mock out utils.load_system_paasta_config instead of all of these -- we test the behavior of load_system_paasta_config elsewhere, we can just trust it to work here.

tests/test_chronos_tools.py
```diff
@@ -93,50 +93,64 @@ def test_chronos_config_object_no_user(self):
...
-        fake_path = '/dne'
+        from_file = {}
+        expected = {}
+        file_mock = mock.MagicMock(spec=file)
         with contextlib.nested(
-            mock.patch('paasta_tools.chronos_tools.open', create=True, side_effect=IOError(2, 'a', 'b')),
+            mock.patch('paasta_tools.utils.open', create=True, return_value=file_mock),
+            mock.patch('json.load', autospec=True, return_value=from_file),
+            mock.patch('os.path.isdir', autospec=True, return_value=True),
+            mock.patch('os.access', autospec=True, return_value=True),
```
EvanKrall: same as above

tests/test_chronos_tools.py
EvanKrall: Or just delete, per Kyle's suggestion on the similar `test_load_marathon_config*`

EvanKrall:
Other than the comments on the tests, looks good.

EvanKrall:
ship

matthewbentley:
For existing marathon.json and chronos.json files, this will put the user, password, and url parameters in the top level configuration. Will that be safe?

EvanKrall:
Yeah, I think that's already happening, and shouldn't affect anything for now. Once this paasta change gets deployed we can clean up those parameters from the top level.

EvanKrall:
Ship

Issue #547, opened 2016-05-27
Fix Documentation Warnings


nhandler:
We appear to be ignoring a number of errors/warnings in our documentation generation. You can see some of them at https://travis-ci.org/Yelp/paasta/jobs/133444236#L370 . We should resolve these issues.

PR #556, opened 2016-05-31
Fix a bunch of doc warnings (but not all, there are still some consis…

matthewbentley:
…tancy warnings)
Connects to #547. I needed to take a break and do something easy.


nhandler:
Ship It!

Rob-Johnson:
shipit

PR #557, opened 2016-06-01
Upgrade to mesos 0.28.1

nhandler:
Mesos 0.28.1 has been released. This change updates our itests to test against it.


solarkennedy:
ship

PR #558, opened 2016-06-03
Use sphinxcontrib-programoutput instead of sphinxcontrib-autorun

nhandler:
`sphinxcontrib-autorun` shows the command being executed along with its output in a syntax highlighted textarea. We really just want plain text. This is what `sphinxcontrib-programoutput` produces.


solarkennedy:
Nice! Ship.

PR #559, opened 2016-06-03
Add support in `paasta logs` for non tailing modes

ammaraskar:
In preparation for adding non tailing modes to the scribe log reader

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -25,12 +25,15 @@ from Queue import Empty
 
 import dateutil
 import isodate
+import pytz
 
 try:
     from scribereader import scribereader
     from scribereader.scribereader import StreamTailerSetupError
 except ImportError:
     scribereader = None
+
+from pytimeparse.timeparse import timeparse
 
 from paasta_tools import chronos_tools
 from paasta_tools.marathon_tools import format_job_id
@@ -68,7 +71,7 @@ def add_subparser(subparsers):
     )
     status_parser.add_argument(
         '-s', '--service',
-        help='The name of the service you wish to inspect. Defaults to autodetect.'
+        help='The name of the service you wish to inspect. Defaults to autodetect.',
     ).completer = lazy_choices_completer(list_services)
     components_help = 'A comma separated list of the components you want logs for.'
     status_parser.add_argument(
@@ -81,7 +84,7 @@ def add_subparser(subparsers):
         help=cluster_help,
     ).completer = completer_clusters
     status_parser.add_argument(
-        '-f', '-F', '--tail', dest='tail', action='store_true', default=True,
+        '-f', '-F', '--tail', dest='tail', action='store_true', default=False,
         help='Stream the logs and follow it for more data',
     )
     status_parser.add_argument(
@@ -91,7 +94,7 @@ def add_subparser(subparsers):
     status_parser.add_argument(
         '-r', '--raw-mode', action='store_true',
         dest='raw_mode', default=False,
-        help="Don't pretty-print logs; emit them exactly as they are in scribe."
+        help="Don't pretty-print logs; emit them exactly as they are in scribe.",
     )
     status_parser.add_argument(
         '-d', '--soa-dir',
@@ -100,9 +103,38 @@ def add_subparser(subparsers):
         default=DEFAULT_SOA_DIR,
         help="define a different soa config directory",
     )
+
+    status_parser.add_argument(
+        '-a', '--from', '--after', dest='time_from',
+        help='The time to start gettings logs from. This can be an ISO-8601 timestamp or a human readable duration '
+             'parseable by pytimeparse such as "5m", "1d3h" etc. For example: --from "3m" would start retrieving logs '
+             'from 3 minutes ago',
+    )
+    status_parser.add_argument(
+        '-t', '--to', dest='time_to',
+        help='The time to get logs up to. This can be an ISO-8601 timestamp or a human readable duration'
+             'parseable by pytimeparse such as "5m", "1d3h" etc. Defaults to right now',
+    )
+    status_parser.add_argument(
+        '-l', '-n', '--lines', dest='line_count',
+        help='The number of lines to retrieve from the specified offset. May optionally be prefixed with a "+" or "-" '
+             'to specify which direction from the offset, defaults to "-100"',
+        type=int,
+    )
+    status_parser.add_argument(
+        '-o', '--line-offset', dest='line_offset',
+        help='The offset at which line to start grabbing logs from. For example 1 would be the first line. Paired with '
+             '--lines +100 would give you the first 100 lines of logs. Defaults to the latest line\'s offset',
+        type=int,
+    )
     default_component_string = ','.join(DEFAULT_COMPONENTS)
     component_descriptions = build_component_descriptions(LOG_COMPONENTS)
-    epilog = 'COMPONENTS\n' \
+    epilog = 'TIME/LINE PARAMETERS\n' \
+             'The args for time and line based offsetting are mutually exclusive, they cannot be used together. ' \
+             'Additionally, some logging backends may not support offsetting by time or offsetting by lines.' \
+             '\n' \
+             '\n' \
+             'COMPONENTS\n' \
              'There are many possible components of Paasta logs that you might be interested in:\n' \
              'Run --list-components to see all available log components.\n' \
              'If unset, the default components are:\n\t%s\n' \
@@ -328,16 +360,36 @@ def get_log_reader():
 
 
 class LogReader(object):
-    def __init__(self, **kwargs):
-        pass
+    # Tailing, i.e actively viewing logs as they come in
+    SUPPORTS_TAILING = False
+    # Getting the last n lines of logs
+    SUPPORTS_LINE_COUNT = False
+    # Getting the last/prev n lines of logs from line #34013 for example
+    SUPPORTS_LINE_OFFSET = False
+    # Getting the logs between two given times
+    SUPPORTS_TIME = False
+    # Supporting at least one of these log retrieval modes is required
 
-    def tail_logs(service, levels, components, clusters, raw_mode=False):
+    def tail_logs(self, service, levels, components, clusters, raw_mode=False):
         raise NotImplementedError("tail_logs is not implemented")
+
+    def print_logs_by_time(self, service, start_time, end_time, levels, components, clusters, raw_mode):
+        raise NotImplementedError("print_logs_by_time is not implemented")
+
+    def print_last_n_logs(self, service, line_count, levels, components, clusters, raw_mode):
+        raise NotImplementedError("print_last_n_logs is not implemented")
+
+    def print_logs_by_offset(self, service, line_count, offset, levels, components, clusters, raw_mode):
+        raise NotImplementedError("print_logs_by_offset is not implemented")
 
 
 @register_log_reader('scribereader')
 class ScribeLogReader(LogReader):
-    def __init__(self, cluster_map, **kwargs):
+    SUPPORTS_TAILING = True
+
+    def __init__(self, cluster_map):
+        super(ScribeLogReader, self).__init__()
+
         if scribereader is None:
             raise Exception("scribereader package must be available to use scribereader log reading backend")
         self.cluster_map = cluster_map
@@ -559,6 +611,94 @@ class ScribeLogReader(LogReader):
             return env
 
 
+def generate_start_end_time(from_string="30m", to_string=None):
+    """Parses the --from and --to command line arguments to create python
+    datetime objects representing the start and end times for log retrieval
+
+    :param from_string: The --from argument, defaults to 30 minutes
+    :param to_string: The --to argument, defaults to the time right now
+    :return: A tuple containing start_time, end_time, which specify the interval of log retrieval
+    """
+    if to_string is None:
+        end_time = datetime.datetime.utcnow()
+    else:
+        # Try parsing as a a natural time duration first, if that fails move on to
+        # parsing as an ISO-8601 timestamp
+        to_duration = timeparse(to_string)
+
+        if to_duration is not None:
+            end_time = datetime.datetime.utcnow() - datetime.timedelta(seconds=to_duration)
+        else:
+            end_time = isodate.parse_datetime(to_string)
+            if not end_time:
+                raise ValueError("--to argument not in ISO8601 format and not a valid pytimeparse duration")
+
+    from_duration = timeparse(from_string)
+    if from_duration is not None:
+        start_time = datetime.datetime.utcnow() - datetime.timedelta(seconds=from_duration)
+    else:
+        start_time = isodate.parse_datetime(from_string)
+
+        if not start_time:
+            raise ValueError("--from argument not in ISO8601 format and not a valid pytimeparse duration")
+
+    # Covert the timestamps to something timezone aware
+    start_time = pytz.utc.localize(start_time)
+    end_time = pytz.utc.localize(end_time)
+
+    if start_time > end_time:
+        raise ValueError("Start time bigger than end time")
+
+    return start_time, end_time
+
+
+def validate_filtering_args(args, log_reader):
+    if not log_reader.SUPPORTS_LINE_OFFSET and args.line_offset is not None:
+        print PaastaColors.red(log_reader.__class__.__name__ + " does not support line based offsets")
+        return False
+    if not log_reader.SUPPORTS_LINE_COUNT and args.line_count is not None:
+        print PaastaColors.red(log_reader.__class__.__name__ + " does not support line count based log retrieval")
+        return False
+    if not log_reader.SUPPORTS_TAILING and args.tail:
+        print PaastaColors.red(log_reader.__class__.__name__ + " does not support tailing")
+        return False
+    if not log_reader.SUPPORTS_TIME and (args.time_from is not None or args.time_to is not None):
+        print PaastaColors.red(log_reader.__class__.__name__ + " does not support time based offsets")
+        return False
+
+    if args.tail and (args.line_count is not None or args.time_from is not None or
+                      args.time_to is not None or args.line_offset is not None):
+        print PaastaColors.red("You cannot specify line/time based filtering parameters when tailing")
+        return False
+
+    # Can't have both
+    if args.line_count is not None and args.time_from is not None:
+        print PaastaColors.red("You cannot filter based on both line counts and time")
+        return False
+
+    return True
+
+
+def pick_default_log_mode(args, log_reader, service, levels, components, clusters):
+    if log_reader.SUPPORTS_LINE_COUNT:
+        sys.stderr.write(PaastaColors.cyan("No filtering specified, grabbing last 100 lines") + "\n")
+        log_reader.print_last_n_logs(service, 100, levels, components, clusters, raw_mode=args.raw_mode)
+        return 0
+
+    elif log_reader.SUPPORTS_TIME:
+        start_time, end_time = generate_start_end_time()
+
+        sys.stderr.write(PaastaColors.cyan("No filtering specified, grabbing last 30 minutes of logs") + "\n")
+        log_reader.print_logs_by_time(service, start_time, end_time, levels, components, clusters,
+                                      raw_mode=args.raw_mode)
+        return 0
+
+    elif log_reader.SUPPORTS_TAILING:
+        sys.stderr.write(PaastaColors.cyan("No filtering specified, tailing logs") + "\n")
+        log_reader.tail_logs(service, levels, components, clusters, raw_mode=args.raw_mode)
+        return 0
+
+
 def paasta_logs(args):
     """Print the logs for as Paasta service.
     :param args: argparse.Namespace obj created from sys.args by cli"""
@@ -586,7 +726,39 @@ def paasta_logs(args):
 
     log_reader = get_log_reader()
 
+    if not validate_filtering_args(args, log_reader):
+        return 1
+
+    # They haven't specified what kind of filtering they want, decide for them
+    if args.line_count is None and args.time_from is None and not args.tail:
+        return pick_default_log_mode(args, log_reader, service, levels, components, clusters)
+
     if args.tail:
+        sys.stderr.write(PaastaColors.cyan("Tailing logs") + "\n")
         log_reader.tail_logs(service, levels, components, clusters, raw_mode=args.raw_mode)
-    else:
-        print "Non-tailing actions are not yet supported"
+        return 0
+
+    # If the logger doesn't support offsetting the number of lines by a particular line number
+    # there is no point in distinguishing between a positive/negative number of lines since it
+    # can only get the last N lines
+    if not log_reader.SUPPORTS_LINE_OFFSET and args.line_count is not None:
+        args.line_count = abs(args.line_count)
+
+    # Handle line based filtering
+    if args.line_count is not None and args.line_offset is None:
+        log_reader.print_last_n_logs(service, args.line_count, levels, components, clusters, raw_mode=args.raw_mode)
+        return 0
+    elif args.line_count is not None and args.line_offset is not None:
+        log_reader.print_logs_by_offset(service, args.line_count, args.line_offset,
+                                        levels, components, clusters, raw_mode=args.raw_mode)
+        return 0
+
+    # Handle time based filtering
+    try:
+        start_time, end_time = generate_start_end_time(args.time_from, args.time_to)
+    except ValueError as e:
+        print PaastaColors.red(e.message)
+        return 1
+
+    log_reader.print_logs_by_time(service, start_time, end_time, levels, components, clusters,
+                                  raw_mode=args.raw_mode)
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -33,6 +33,7 @@ import signal
 import sys
 import tempfile
 import threading
+from collections import OrderedDict
 from fnmatch import fnmatch
 from functools import wraps
 from subprocess import PIPE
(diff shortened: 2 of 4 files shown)


paasta_tools/cli/cmds/logs.py
```diff
@@ -100,9 +103,37 @@ def add_subparser(subparsers):
         default=DEFAULT_SOA_DIR,
         help="define a different soa config directory",
     )
+
+    status_parser.add_argument(
+        '-a', '--ago', dest='time_ago',
+        help='How long ago to get logs from, uses friendly durations like 32m or 1w. Defaults to 30 minutes',
+    )
+    status_parser.add_argument(
+        '-t', '--time', dest='time',
```
Rob-Johnson: the ``time`` and ``ago`` args seem confusing. Can we instead to ``from`` and ``to``?

paasta_tools/cli/cmds/logs.py
```diff
@@ -100,9 +103,37 @@ def add_subparser(subparsers):
...
+        '-lo', '--line-offset', dest='line_offset',
+        help='The offset at which line to start grabbing logs from. For example 1 would be the first line. Paired with '
+             '--count +100 would give you the first 100 lines of logs. Defaults to the latest line\'s offset',
+        type=int
+    )
     default_component_string = ','.join(DEFAULT_COMPONENTS)
     component_descriptions = build_component_descriptions(LOG_COMPONENTS)
-    epilog = 'COMPONENTS\n' \
+    epilog = 'TIME/LINE PARAMETERS\n' \
+             'The args for time and line based offsetting are mutually exclusive, they cannot be used together. ' \
```
Rob-Johnson: Is this enforced anywhere? I'd expect a change somewhere that did something like ``if time and line: raise ValueError('Cannot have both time and line')``

paasta_tools/cli/cmds/logs.py
```diff
@@ -121,18 +152,31 @@ def completer_clusters(prefix, parsed_args, **kwargs):
         return list_clusters()
 
 
+def color_component(component, color_param):
+    # If the color field is iterable try to join
+    # all the colors together, otherwise just apply
+    # the single color
+    try:
+        for color in color_param:
```
Rob-Johnson: Can you explain the use case for this? If I'm reading this correctly, this will go through a list of functions, applying each of them to the same argument? Is this so you can have something like ``strip(blue("foo"))``?

paasta_tools/cli/cmds/logs.py
```diff
@@ -328,16 +372,37 @@ def get_log_reader():
 
 
 class LogReader(object):
-    def __init__(self, **kwargs):
-        pass
+    def __init__(self):
+        # Tailing, i.e actively viewing logs as they come in
+        self.SUPPORTS_TAILING = False
```
Rob-Johnson: please make these non-caps - the only place you really use CAPS is for pretend constant variables, which these are not.

paasta_tools/cli/cmds/logs.py
```diff
@@ -559,6 +624,65 @@ def cluster_to_scribe_env(self, cluster):
             return env
 
 
+def generate_start_end_time(time_ago="30m", time=datetime.datetime.utcnow()):
```
Rob-Johnson: the default arg for time is a bad idea. Check this out:

```
>>> import datetime
>>> datetime.datetime.utcnow()
datetime.datetime(2016, 6, 6, 13, 26, 37, 870869)
>>> def x(y=datetime.datetime.utcnow()):
...     print y
...
>>> x()
2016-06-06 13:27:06.622029
>>> x()
2016-06-06 13:27:06.622029
>>> x()
2016-06-06 13:27:06.622029
>>> x()
2016-06-06 13:27:06.622029
```

Notice that the value printed is never returned: that's because default args are evaluated once, when the function is defined.

Instead, you should do the following:

```
def generate_start_end_time(time_ago="30m", time=None):
    if time is None:
         time=datetime.datetime.utcnow()
```

paasta_tools/cli/cmds/logs.py
Rob-Johnson: ah, nm, just found it 👍

paasta_tools/cli/cmds/logs.py
```diff
@@ -559,6 +624,65 @@ def cluster_to_scribe_env(self, cluster):
...
+
+        sys.stderr.write(PaastaColors.cyan("No filtering specified, grabbing last 30 minutes of logs") + "\n")
+        log_reader.print_logs_by_time(service, start_time, end_time, levels, components, clusters,
+                                      raw_mode=args.raw_mode)
+        return 0
+
+    elif log_reader.SUPPORTS_TAILING:
+        sys.stderr.write(PaastaColors.cyan("No filtering specified, tailing logs") + "\n")
+        log_reader.tail_logs(service, levels, components, clusters, raw_mode=args.raw_mode)
+        return 0
```
Rob-Johnson: what are these return values for? also, why're you writing to stderr on each of these? aren't they all informational (so should go to stdout)?

paasta_tools/cli/cmds/logs.py
ammaraskar: Please see this [line](https://github.com/Yelp/paasta/pull/559/files#diff-d3b2beac0ebf41c3a27cbe71e22e935aR450) right here.

It's mostly so I can get this combination of yellow and bold in the help output which looks pretty nice I think:
paasta logs help

paasta_tools/cli/cmds/logs.py
ammaraskar: Isn't this a constant? A logreader's supported retrieval modes won't exactly change during runtime

paasta_tools/cli/cmds/logs.py
ammaraskar: Thank you, will fix.

paasta_tools/cli/cmds/logs.py
ammaraskar: The reason for printing them to stderr is in case someone wants to pipe their logs out to a file for example. If we printed it to stdout, this output would get mangled in along with it.

paasta_tools/cli/cmds/logs.py
ammaraskar: So the reason I decided not to have a from and to is because I feel like it is much more convenient for end users. I agree, it is a bit confusing at first glance but figuring out the timestamps for `from` and `to` yourself is so much of a hassle. 

In most cases people will be wanting logs relative to the current time, so for example they get a report that their service is misbehaving, the report says the problem was encountered like 5 minutes ago. Well then it's simple enough to do `paasta logs --ago 10m` and get all the logs from 10 minutes ago. However if you had to manually figure out the timestamps you'd have like `paasta logs --from 2016-06-06T16:25:30+00:00 --to 2016-06-06T16:35:30+00:00`

Rob-Johnson: There is also precedent for the ``--from`` and ``--to`` style args elsewhere: 
- our main users (those at Yelp) are used to ``--min-date`` and ``--max-date`` to scribereader
- externally, ``docker logs`` has a ``--since`` arg (equivalent to our from)

How about ``--from`` and ``--to`` supporting the human readable style you're proposing? so you can do ``paasta logs --from "10 minutes ago" --to "5 minutes ago"`` with `--to` defaulting to now?

paasta_tools/cli/cmds/logs.py
Rob-Johnson: ah, I see. In that case, I'd prefer a more general 'compose' function:

in ``utils.py``
```
def compose(func_one, func_two):
     def composed(x):
         return func_one(func_two(x))
     return composed
```

then you can do:

```
color_and_bold = compose(bold, yellow)
color_and_bold(line)
```

or even further:
```
italic_bold_color = compose(italic_string, compose(bold, color))
return italic_bold_color("hi!")
```

we can generalise this further to support n functions if you like, but this is a fine start. What do you think?

paasta_tools/cli/cmds/logs.py
Rob-Johnson: I mean, the very fact that you're setting them in the constructor means they're not. They're a class level attribute that's overriden for each subclass, right?

ammaraskar: Sure they're not constants in a strict sense that they're only assigned once but semantically I'd argue they're still constants and deserve to be named as such.

paasta_tools/cli/cmds/logs.py
ammaraskar: That makes sense, I'll change it to --to and --from with the option of both timestamps and human readable time deltas.

paasta_tools/cli/cmds/logs.py
ammaraskar: That makes sense and yeah that'd be useful for future use, I'll add that in.

paasta_tools/cli/cmds/logs.py
```diff
@@ -81,7 +84,7 @@ def add_subparser(subparsers):
         help=cluster_help,
     ).completer = completer_clusters
     status_parser.add_argument(
-        '-f', '-F', '--tail', dest='tail', action='store_true', default=True,
+        '-f', '-F', '--tail', dest='tail', action='store_true', default=False,
```
nhandler: Do we really want to couple this default change to the addition of support for non-tailing? I would think this would make a better follow-up change after we let people test out this new functionality a bit (and update all of our documentation).

paasta_tools/cli/cmds/logs.py
```diff
@@ -100,9 +103,37 @@ def add_subparser(subparsers):
         default=DEFAULT_SOA_DIR,
         help="define a different soa config directory",
     )
+
+    status_parser.add_argument(
+        '-fr', '--from', dest='time_from',
```
nhandler: Usually, single dash (-) arguments have single-character names. I'm not sure whether argparse supports having `-fr`, but even if it does, we should probably stick with the unix standard.

paasta_tools/cli/cmds/logs.py
```diff
@@ -100,9 +103,37 @@ def add_subparser(subparsers):
...
+             'differences. Defaults to right now.'
+    )
+    status_parser.add_argument(
+        '-l', '--lines', dest='line_count',
+        help='The number of lines to retrieve from the specified offset. May optionally be prefixed with a "+" or "-" '
+             'to specify which direction from the offset, defaults to "-100"',
+        type=int,
+    )
+    status_parser.add_argument(
+        '-lo', '--line-offset', dest='line_offset',
```
nhandler: As above, let's stick to the convention that single - arguments have single-character names.

paasta_tools/cli/cmds/logs.py
nhandler: I think what @Rob-Johnson is saying is that if you want to treat it like a constant, move it out of the constructor into the top level scope. If you are going to have it in the constructor where it will be set for each instance of the class, rename it to be in lowercase.

ammaraskar: aah okay, that makes sense. LogReader is a singleton right now which is why I had it ass instance level attributes. Fixing.

paasta_tools/cli/cmds/logs.py
ammaraskar: kwa seemed to think so, also, just with this change it will default to tailing since that's currently the only thing ScribeLogReader is set to support

paasta_tools/cli/cmds/logs.py
ammaraskar: Any ideas for a single letter for this? `-f` conflicts with tail

paasta_tools/cli/cmds/logs.py
```diff
@@ -328,16 +359,37 @@ def get_log_reader():
...
+        raise NotImplementedError("print_last_n_logs is not implemented")
+
+    def print_logs_by_offset(self, service, line_count, offset, levels, components, clusters, raw_mode):
+        raise NotImplementedError("print_logs_by_offset is not implemented")
+
 
 @register_log_reader('scribereader')
 class ScribeLogReader(LogReader):
-    def __init__(self, cluster_map, **kwargs):
+    def __init__(self, cluster_map):
```
nhandler: You are dropping the `**kwargs` here. Since the tests pass, I'm assuming this means that we weren't actually passing in any additional arguments (or that you've refactored that out).

tests/cli/test_cmds_logs.py
```diff
@@ -755,3 +772,135 @@ def test_get_log_reader():
...
+    start_time, end_time = logs.generate_start_end_time()
+
+    # Default for no args, test that there's a 30 minute difference
+    time_delta = end_time - start_time
+
+    # Do a loose comparison to make this test less time sensitive
+    actual_time = time_delta.total_seconds()
+    ideal_time = 30 * 60
+
+    assert abs(actual_time - ideal_time) < 0.1
```
nhandler: Could you add a comment explaining why this needs to be `< 0.1` rather than `== 0`.

tests/cli/test_cmds_logs.py
```diff
@@ -755,3 +772,135 @@ def test_get_log_reader():
...
+
+
+def test_pick_default_log_mode():
+    with mock.patch('paasta_tools.cli.cmds.logs.LogReader.tail_logs') as tail_logs:
+        args, _ = parse_args(["logs"])
+
+        fake_reader = logs.LogReader()
+        fake_reader.SUPPORTS_TAILING = True
+
+        logs.pick_default_log_mode(args, fake_reader, None, None, None, None)
```
nhandler: Can you use kwargs to make it more obvious what each of the `None`s refer to?

tests/cli/test_cmds_logs.py
ammaraskar: There's a comment right above it:

```python
# Do a loose comparison to make this test less time sensitive
```

If it's == 0 it will fail on slower build platforms like travis.ci where lines of code aren't executed as instantly as you'd expect

tests/cli/test_cmds_logs.py
ammaraskar: Done

paasta_tools/cli/cmds/logs.py
```diff
@@ -100,9 +103,37 @@ def add_subparser(subparsers):
...
+    status_parser.add_argument(
+        '-l', '--lines', dest='line_count',
+        help='The number of lines to retrieve from the specified offset. May optionally be prefixed with a "+" or "-" '
+             'to specify which direction from the offset, defaults to "-100"',
+        type=int,
+    )
+    status_parser.add_argument(
+        '-o', '--line-offset', dest='line_offset',
+        help='The offset at which line to start grabbing logs from. For example 1 would be the first line. Paired with '
+             '--count +100 would give you the first 100 lines of logs. Defaults to the latest line\'s offset',
```
EvanKrall: what is `--count`?

ammaraskar: whoops, supposed to be `--lines`

EvanKrall: Seems like you're sorta trying to match `tail`'s semantics with the +100, but you're not quite there: `tail -n +100` starts from the 100th line (offset 100) and goes until EOF; `tail -n 100` gives you the 100 lines at the end of the file.

paasta_tools/cli/cmds/logs.py
```diff
@@ -559,6 +610,86 @@ def cluster_to_scribe_env(self, cluster):
...
+    if not log_reader.SUPPORTS_TAILING and args.tail:
+        print PaastaColors.red(log_reader.__class__.__name__ + " does not support tailing")
+        return False
+    if not log_reader.SUPPORTS_TIME and (args.time_from is not None or args.time_to is not None):
+        print PaastaColors.red(log_reader.__class__.__name__ + " does not support time based offsets")
+        return False
+
+    if args.tail and (args.line_count is not None or args.time_from is not None or
+                      args.time_to is not None or args.line_offset is not None):
+        print PaastaColors.red("You cannot specify line/time based filtering parameters when tailing")
```
EvanKrall: Why not?

Both the `tail` command and Kafka consumers are capable of starting from N lines/messages since the end of the stream/file, printing those, then continuing on to follow.

Scribereader is the outlier.

paasta_tools/cli/cmds/logs.py
ammaraskar: Kinda, I don't really like that particular part about tail's syntax. I much prefer this way of doing it where you specify which offset and then the line count in either direction.

paasta_tools/cli/cmds/logs.py
```diff
@@ -100,9 +103,37 @@ def add_subparser(subparsers):
...
+        help='The ISO-8601 timestamp to get the logs from. Alternatively, for convenience, time differences can be '
+             'specified in a human readable way, such as --from 5m will grab logs from 5 minutes ago',
+    )
+    status_parser.add_argument(
+        '-t', '--to', dest='time_to',
+        help='The ISO-8601 timestamp to get logs up to. Follows same convention as --from for human readable '
+             'differences. Defaults to right now.'
+    )
+    status_parser.add_argument(
+        '-l', '--lines', dest='line_count',
```
EvanKrall: how about -n to match `tail`/`head`?

paasta_tools/cli/cmds/logs.py
ammaraskar: Simplifies the API, how often do you actually need to print the last 100 lines and then also follow the logs? I 

I can change it if you think it's worth it.

paasta_tools/cli/cmds/logs.py
ammaraskar: Done.

paasta_tools/cli/cmds/logs.py
EvanKrall: I'm okay with leaving it as is for now, and once we actually have a KafkaLogReader we can revisit.

Though one could argue that which line we start reading from and which line we stop reading at should be independently controllable, much like `docker logs`' `--since` and `--until`.

Don't let me bikeshed you too much here. We can always change the interface in the future, particularly to go from stricter to less strict.

EvanKrall:
An idea that could make --help easier to read: What if we checked which log reader driver was specified in SystemPaastaConfig, and then only added the command line flags like `--after`, `--lines`, etc. if that driver supports those features?

Might be a bit of a refactor of what you have now; I don't know whether SystemPaastaConfig is easily available at that point in the code.

This seems fine to ship and iterate on.

ammaraskar:
Hmm I don't think it's available at the point where add_subparser is called, so it would require some additional complexity to do that.

paasta_tools/cli/cmds/logs.py
```diff
@@ -100,9 +103,37 @@ def add_subparser(subparsers):
         default=DEFAULT_SOA_DIR,
         help="define a different soa config directory",
     )
+
+    status_parser.add_argument(
+        '-a', '--from', '--after', dest='time_from',
```
Rob-Johnson: how about we just copy docker and have `'-s', '--since` ?

paasta_tools/cli/cmds/logs.py
```diff
@@ -100,9 +103,37 @@ def add_subparser(subparsers):
...
+        help='The ISO-8601 timestamp to get the logs from. Alternatively, for convenience, time differences can be '
```
Rob-Johnson: can you be more specific about what is allowed? 
> The time to get logs from. This can be expressed as an ISO-8601 datestring or any human-readable duration parseable by pytimeparse, such as 5m'

paasta_tools/cli/cmds/logs.py
```diff
@@ -559,6 +610,86 @@ def cluster_to_scribe_env(self, cluster):
             return env
 
 
+def generate_start_end_time(from_string="30m", to_string=None):
+    """
```
Rob-Johnson: sorry to nitpick, but can you fix the format of your docstrings please https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings and make sure you add ``:param:`` and ``:returns:`` docs

paasta_tools/cli/cmds/logs.py
```diff
@@ -559,6 +610,86 @@ def cluster_to_scribe_env(self, cluster):
...
+        # Try parsing as a a natural time duration first, if that fails move on to
+        # parsing as an ISO-8601 timestamp
+        to_duration = timeparse(to_string)
+
+        if to_duration is not None:
+            end_time = datetime.datetime.utcnow() - datetime.timedelta(seconds=to_duration)
+        else:
+            end_time = isodate.parse_datetime(to_string)
+
+    from_duration = timeparse(from_string)
```
Rob-Johnson: is this indented correctly? why do you parse the ``to_string`` inside the else block but not the from string?

paasta_tools/cli/cmds/logs.py
```diff
@@ -586,7 +717,37 @@ def paasta_logs(args):
...
     log_reader = get_log_reader()
 
+    if not validate_filtering_args(args, log_reader):
+        return 1
+
+    # They haven't specified what kind of filtering they want, decide for them
+    if args.line_count is None and args.time_from is None and not args.tail:
+        return pick_default_log_mode(args, log_reader, service, levels, components, clusters)
+
+    # Handle tailing
```
Rob-Johnson: please ditch this comment

paasta_tools/cli/cmds/logs.py
```diff
@@ -586,7 +717,37 @@ def paasta_logs(args):
...
     if args.tail:
+        # Print a little message before we start tailing to stderr so people don't panic
```
Rob-Johnson: and this one

paasta_tools/utils.py
```diff
@@ -325,6 +326,12 @@ def validate_service_instance(service, instance, cluster, soa_dir):
         sys.exit(3)
 
 
+def compose(func_one, func_two):
```
Rob-Johnson: I don't see a test for this anywhere?

paasta_tools/cli/cmds/logs.py
ammaraskar: It is, it's because the default argument for to_string is None, an alternative way to handle it would be to make the default to_string be something like "0s" but that just seems hacky.

paasta_tools/utils.py
ammaraskar: hmm, there's no explicit test for it but it was being covered because the tests do `--help` for each command. I'll add a test specifically for this though.

paasta_tools/cli/cmds/logs.py
ammaraskar: I kinda like `--to` and `--from`

`--since` and `--to` don't go together nearly as nicely.

paasta_tools/cli/cmds/logs.py
```diff
@@ -559,6 +610,86 @@ def cluster_to_scribe_env(self, cluster):
...
+
+
+def pick_default_log_mode(args, log_reader, service, levels, components, clusters):
+    if log_reader.SUPPORTS_LINE_COUNT:
+        sys.stderr.write(PaastaColors.cyan("No filtering specified, grabbing last 100 lines") + "\n")
+        log_reader.print_last_n_logs(service, 100, levels, components, clusters, raw_mode=args.raw_mode)
+        return 0
+
+    elif log_reader.SUPPORTS_TIME:
+        start_time, end_time = generate_start_end_time()
```
Rob-Johnson: you need to catch the ``ValueError`` that could be thrown by ``generate_start_end_time``

paasta_tools/cli/cmds/logs.py
Rob-Johnson: yeah, I like ``to`` and ``from``, but ``-a`` for ``--from`` isn't intuitive.

paasta_tools/cli/cmds/logs.py
```diff
@@ -559,6 +610,86 @@ def cluster_to_scribe_env(self, cluster):
...
+        end_time = datetime.datetime.utcnow()
+    else:
+        # Try parsing as a a natural time duration first, if that fails move on to
+        # parsing as an ISO-8601 timestamp
+        to_duration = timeparse(to_string)
+
+        if to_duration is not None:
+            end_time = datetime.datetime.utcnow() - datetime.timedelta(seconds=to_duration)
+        else:
+            end_time = isodate.parse_datetime(to_string)
```
Rob-Johnson: can you handle exceptions for invalid format here please

paasta_tools/cli/cmds/logs.py
```diff
@@ -559,6 +610,86 @@ def cluster_to_scribe_env(self, cluster):
...
+
+    from_duration = timeparse(from_string)
+    if from_duration is not None:
+        start_time = datetime.datetime.utcnow() - datetime.timedelta(seconds=from_duration)
+    else:
+        start_time = isodate.parse_datetime(from_string)
```
Rob-Johnson: error handling please

paasta_tools/cli/cmds/logs.py
```diff
@@ -586,7 +717,37 @@ def paasta_logs(args):
...
+    if args.line_count is not None and args.line_offset is None:
+        log_reader.print_last_n_logs(service, args.line_count, levels, components, clusters, raw_mode=args.raw_mode)
+        return 0
+    elif args.line_count is not None and args.line_offset is not None:
+        log_reader.print_logs_by_offset(service, args.line_count, args.line_offset,
+                                        levels, components, clusters, raw_mode=args.raw_mode)
+        return 0
+
+    # Handle time based filtering
+    start_time, end_time = generate_start_end_time(args.time_from, args.time_to)
```
Rob-Johnson: you need error handling here for ``ValueError`` too

paasta_tools/cli/cmds/logs.py
ammaraskar: Oh right, we can't have `-s` because `-s` is for service name

paasta_tools/cli/cmds/logs.py
ammaraskar: Not here, since this is just from the default arguments so it should never fail. Will add one for main though

nhandler:
Ship It!

PR #560, opened 2016-06-06
Add chronos status to paasta status output

keshavdv:
Adds chronos task status to the output of paasta status.

screen shot 2016-06-06 at 4 27 03 pm
screen shot 2016-06-06 at 4 28 20 pm


Rob-Johnson:
shipit

solarkennedy:
shipit.

I'm a little worried about speed, but we *should* be doing transparent request caching and this all comes from the single csv endpoint.

Except python-chronos uses httplib directly :(

keshavdv:
Merged into master manually

PR #561, opened 2016-06-07
Fix command quoting issues

### LINKED ISSUE #555: local-run has some issues with cmd quoting
The following job config doesn't work as expected as the command is incorrectly quoted when passed to `sh -c`:

```
example_chronos_job:
  cmd: "echo 'Hello' && sleep 60"
  schedule: 'R/2016-04-15T06:00:00Z/PT24H'
  schedule_time_zone: 'America/Los_Angeles'
```

This turns into:
`sh -c 'echo foo '"'"'&&'"'"' sleep 60'`

koikonom:
The command the container is supposed to run is being passed to
pipes.quote() once in get_docker_run_cmd() and once again in
run_docker_container().

If the string contains anything that pipes.quote thinks should be
quoted then we end up with some very weird results. For example
_"echo 'Hello' && sleep 60_" turns to "_echo foo '"'"'&&'"'"' sleep 60_".

The fix is to pass the whole command through pipes.quote() only once.

Fixes #555

### DIFF
--- a/general_itests/steps/local_run_steps.py
+++ b/general_itests/steps/local_run_steps.py
@@ -43,7 +43,7 @@ def non_interactive_local_run(context, var, val):
                         "--service fake_simple_service "
                         "--cluster test-cluster "
                         "--build "
-                        "--cmd '/bin/sh -c \"echo \"%s=$%s\" && sleep 2s && exit 42\"'" % (var, val))
+                        '''--cmd '/bin/sh -c "echo \\"%s=$%s\\" && sleep 2s && exit 42"' ''' % (var, val))
         context.local_run_return_code, context.local_run_output = _run(command=localrun_cmd, timeout=30)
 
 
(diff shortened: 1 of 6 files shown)


koikonom:
This fixes issue #555.

Rob-Johnson:
could you add an integration test please?

koikonom:
OK this turned out much weirder than expected. I had to change the quoting logic so we only quote the command the container runs and not the whole thing. 

Both unit and functional tests pass (had to change one to match the changes I made) and I also added a new functional test to cover the reported issue.

general_itests/steps/local_run_steps.py
```diff
@@ -76,3 +76,15 @@ def local_run_on_chronos_job(context):
                          "--build "
                          "--cmd '/bin/sh -c \"sleep 2s && exit 42\"'")
         context.local_run_return_code, context.local_run_output = _run(command=local_run_cmd, timeout=30)
+
+
+@when(u'we run paasta local-run in non-interactive mode on a chronos job with cmd set to \'echo hello && sleep 5\'')
```
nhandler: What if we merged this step with the previous one by making `non-interactive` and the command [step parameters](https://pythonhosted.org/behave/tutorial.html#step-parameters)?

koikonom: I thought about it. My rationale was that one test gets the command as a parameter from the command line and the other reads it from the yaml file, so we end up testing two separate things.

nhandler: I guess that makes sense. I would still probably make the command for your new step a parameter.

paasta_tools/cli/cmds/local_run.py
```diff
@@ -381,6 +381,37 @@ def get_container_name():
     return 'paasta_local_run_%s_%s' % (get_username(), randint(1, 999999))
 
 
+def check_for_sh(command):
```
nhandler: This feels rather hackish and error-prone (i.e. what if they use a different shell).

koikonom: You're right... How about I pull the list of shells out of /etc/shells?

keshavdv: Is there a reason you have to even check for this? Why can't we just quote whatever they provide and still always pass it in to /bin/sh (i.e. `sh -c "sh -c 'echo foo'"`)?

EvanKrall: Agreed, please don't attempt to parse command lines; let a shell do it.

solarkennedy:
ship if it works!

PR #562, opened 2016-06-07
disable service_configuration_lib yaml cache when setting up marathon jobs

huadongliu:
This is an alternative for https://github.com/Yelp/service_configuration_lib/pull/11 in case there are concerns for fixing service_configuration_lib.


EvanKrall:
This seems like a reasonable workaround. Ship.

PR #563, opened 2016-06-07
Upgrade to mesos 0.28.2

keshavdv:
Bumps version for itests


nhandler:
Ship it if paata_itests pass

PR #564, opened 2016-06-07
Add cpu-period and cpu-quota to marathon logic

### LINKED ISSUE #504: Support cfs_quota_us / cfs_period_us
Recently, Mesos added an option:

> `--[no]-cgroups_enable_cfs`  Cgroups feature flag to enable hard limits on CPU resources via the CFS bandwidth limiting subfeature. (default: false)

This is really intriguing, since it provides a way to hard-cap containers. We're not sure whether Mesos would set the appropriate options on Docker, but Marathon allows us to pass arbitrary command line flags to `docker run` [via the `container.docker.parameters` field](https://mesosphere.github.io/marathon/docs/rest-api.html#post-v2-apps), and docker has command line flags for this:

```
  --cpu-period                    Limit CPU CFS (Completely Fair Scheduler) period
  --cpu-quota                     Limit CPU CFS (Completely Fair Scheduler) quota
```

@solarkennedy and I played around with this, and it does appear to work:

```
krall@dev5-devc:~/ $ docker run -d --cpu-period 100000 --cpu-quota 20000 busybox sh -c 'yes >/dev/null'
481e13359f87800a5979d09f6c0caa3f8b54754bdfe2d6103cd16563c6000366
krall@dev5-devc:~/ $ docker stats --no-stream 481e13359f87800a5979d09f6c0caa3f8b54754bdfe2d6103cd16563c6000366
CONTAINER                                                          CPU %               MEM USAGE / LIMIT     MEM %               NET I/O             BLOCK I/O
481e13359f87800a5979d09f6c0caa3f8b54754bdfe2d6103cd16563c6000366   20.12%              409.6 kB / 135.2 GB   0.00%               648 B / 648 B       319.5 kB / 0 B
krall@dev5-devc:~/ $
```

We were thinking it would be useful to add some parameters to marathon.yaml:

 - `cfs_period_us`: would set `--cpu-period` pretty much directly, but would default to 100000 (100ms), which is the default in the kernel.
 - `cpu_burst_pct`: Would set --cpu-quota through some math, but would default to 100, allowing services to use up to 2x their CPU allocation before getting throttled. (The CPU shares setting that Mesos/docker is already setting would still apply).

Math: `cfs_quota_us = cpus * cfs_period_us * (100 + cpu_burst_pct) / 100`.

I think as described here, this would be a pretty straightforward change but could provide some big wins in preventing services from impacting each other, particularly during startup.

@Rob-Johnson You're interested in isolation things. Does this seem reasonable to you? Do you want to take this?

matthewbentley:
I think the existing unit tests should cover this new functionality, but it probably also needs integration tests.
Closed #504

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -69,6 +69,9 @@ PUPPET_SERVICE_DIR = '/etc/nerve/puppet_services.d'
 # These should be things that PaaSTA/Marathon knows how to change without requiring a bounce.
 CONFIG_HASH_BLACKLIST = set(['instances', 'backoff_seconds', 'min_instances', 'max_instances'])
 
+DEFAULT_CPU_PERIOD = 100000
+DEFAULT_CPU_BURST_PCT = 100
+
 log = logging.getLogger(__name__)
 logging.getLogger('marathon').setLevel(logging.WARNING)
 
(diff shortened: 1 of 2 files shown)


EvanKrall:
Can you add a test for `get_cpu_quota` sanity-checking the math, e.g. with cpu_burst_pct of 0 and 100?

matthewbentley:
Done

EvanKrall:
Integration tests are going to be difficult without #373. Let's deploy this carefully through our testing environment.

PR #566, opened 2016-06-08
Allow ScribeLogReader to retrieve logs by time and lines

ammaraskar:
Follow up from #559 

The first commit is a very naive approach with lots of code repetition, the second commit then comes in and refactors a bunch to make everything a little neater.

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -19,6 +19,8 @@ import json
 import logging
 import re
 import sys
+from collections import namedtuple
+from contextlib import contextmanager
 from multiprocessing import Process
 from multiprocessing import Queue
 from Queue import Empty
@@ -167,7 +169,29 @@ def prefix(input_string, component):
     return "%s: %s" % (LOG_COMPONENTS[component]['color'](component), input_string)
 
 
-def paasta_log_line_passes_filter(line, levels, service, components, clusters):
+# The reason this returns true if start_time or end_time are None is because
+# filtering by time is optional, and it allows us to simply do
+# if not check_timestamp_in_range(...): return False
+# The default arguments for start_time and end_time are None when we aren't
+# filtering by time
+def check_timestamp_in_range(timestamp, start_time, end_time):
+    """A convenience function to check if a datetime.datetime timestamp is within the given start and end times,
+    returns true if start_time or end_time is None
+
+    :param timestamp: The timestamp to check
+    :param start_time: The start of the interval
+    :param end_time: The end of the interval
+    :return: True if timestamp is within start_time and end_time range, False otherwise
+    """
+    if timestamp is not None and start_time is not None and end_time is not None:
+        if timestamp.tzinfo is None:
+            timestamp = pytz.utc.localize(timestamp)
+        return start_time < timestamp < end_time
+    else:
+        return True
+
+
+def paasta_log_line_passes_filter(line, levels, service, components, clusters, start_time=None, end_time=None):
     """Given a (JSON-formatted) log line, return True if the line should be
     displayed given the provided levels, components, and clusters; return False
     otherwise.
@@ -177,8 +201,30 @@ def paasta_log_line_passes_filter(line, levels, service, components, clusters):
     except ValueError:
         log.debug('Trouble parsing line as json. Skipping. Line: %r' % line)
         return False
+
+    timestamp = isodate.parse_datetime(parsed_line.get('timestamp'))
+    if not check_timestamp_in_range(timestamp, start_time, end_time):
+        return False
     return (
         parsed_line.get('level') in levels and
+        parsed_line.get('component') in components and (
+            parsed_line.get('cluster') in clusters or
+            parsed_line.get('cluster') == ANY_CLUSTER
+        )
+    )
+
+
+def paasta_app_output_passes_filter(line, levels, service, components, clusters, start_time=None, end_time=None):
+    try:
+        parsed_line = json.loads(line)
+    except ValueError:
+        log.debug('Trouble parsing line as json. Skipping. Line: %r' % line)
+        return False
+
+    timestamp = isodate.parse_datetime(parsed_line.get('timestamp'))
+    if not check_timestamp_in_range(timestamp, start_time, end_time):
+        return False
+    return (
         parsed_line.get('component') in components and (
             parsed_line.get('cluster') in clusters or
             parsed_line.get('cluster') == ANY_CLUSTER
@@ -239,7 +285,7 @@ def parse_chronos_log_line(line, clusters, service):
         )
 
 
-def marathon_log_line_passes_filter(line, levels, service, components, clusters):
+def marathon_log_line_passes_filter(line, levels, service, components, clusters, start_time=None, end_time=None):
     """Given a (JSON-formatted) log line where the message is a Marathon log line,
     return True if the line should be displayed given the provided service; return False
     otherwise."""
@@ -248,10 +294,14 @@ def marathon_log_line_passes_filter(line, levels, service, components, clusters)
     except ValueError:
         log.debug('Trouble parsing line as json. Skipping. Line: %r' % line)
         return False
+
+    timestamp = isodate.parse_datetime(parsed_line.get('timestamp'))
+    if not check_timestamp_in_range(timestamp, start_time, end_time):
+        return False
     return format_job_id(service, '') in parsed_line.get('message', '')
 
 
-def chronos_log_line_passes_filter(line, levels, service, components, clusters):
+def chronos_log_line_passes_filter(line, levels, service, components, clusters, start_time=None, end_time=None):
     """Given a (JSON-formatted) log line where the message is a Marathon log line,
     return True if the line should be displayed given the provided service; return False
     otherwise."""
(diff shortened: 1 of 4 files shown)


ammaraskar:
Works fine manually, but I'll see if I can find a better way to unit test this.

paasta_tools/cli/cmds/logs.py
```diff
@@ -167,7 +168,24 @@ def prefix(input_string, component):
     return "%s: %s" % (LOG_COMPONENTS[component]['color'](component), input_string)
 
 
-def paasta_log_line_passes_filter(line, levels, service, components, clusters):
+def check_timestamp_in_range(timestamp, start_time, end_time):
+    """A convenience function to check if a timestamp is within the given start and end times,
+    returns true if start_time or end_time is None
+
+    :param timestamp: The timestamp to check
```
EvanKrall: are these datetime objects or numbers representing unix time?

ammaraskar: datettime objects, I'll amend the docstring.

paasta_tools/cli/cmds/logs.py
```diff
@@ -167,7 +168,24 @@ def prefix(input_string, component):
     return "%s: %s" % (LOG_COMPONENTS[component]['color'](component), input_string)
 
 
-def paasta_log_line_passes_filter(line, levels, service, components, clusters):
+def check_timestamp_in_range(timestamp, start_time, end_time):
+    """A convenience function to check if a timestamp is within the given start and end times,
+    returns true if start_time or end_time is None
```
EvanKrall: interesting default. Why?

ammaraskar: Because `start_time` and `end_time` are sent as kwargs to all the `filter_*` functions, this way I can pass them straight into this function without having if statements everywhere.

paasta_tools/cli/cmds/logs.py
```diff
@@ -422,65 +490,94 @@ def tail_logs(self, service, levels, components, clusters, raw_mode=False):
...
+                'service': service,
+                'levels': levels,
+                'components': components,
+                'clusters': clusters,
+                'queue': queue,
+                'filter_fn': paasta_log_line_passes_filter,
+            }
+            process = Process(target=self.scribe_tail, kwargs=kw)
+            spawned_processes.append(process)
+            process.start()
```
EvanKrall: All of these seem the same except for `kw` and whether they take a `cluster` param. How about this refactor:

 - Make a single function that takes parameters `component, scribe_env, cluster` and looks up `kw` based on component
 - make run_code_over_scribe_envs call that function for each `component, scribe_env, cluster`

paasta_tools/cli/cmds/logs.py
```diff
@@ -549,6 +646,167 @@ def tail_logs(self, service, levels, components, clusters, raw_mode=False):
                 log.warn('Terminating.')
                 break
 
+    def print_logs_by_time(self, service, start_time, end_time, levels, components, clusters, raw_mode):
+        aggregated_logs = []
+
+        def default_fn(scribe_env):
+            stream_name = get_log_name_for_service(service)
```
EvanKrall: Make a function for calculating the stream name based on service/cluster/etc; this and refactor suggested above would make `default_fn`/`marathon_fn`/etc identical, plus you could reuse the stream name calculation logic between `tail_logs`, `print_logs_by_time`, and `print_last_n_logs`

paasta_tools/cli/cmds/logs.py
EvanKrall: I don't know if this really makes sense for the API you have, but maybe if start_time is None you could treat it as `float('-Infinity')` and if end_time is None you could treat it as `float('Infinity')`. If both were undefined, you'd have the same behavior - all timestamps would pass the check; but if e.g. only end_time were undefined you'd have a more sensible behavior: anything after end_time would be filtered out, anything before it would pass.

EvanKrall: (Feel free to ignore my suggestion.)

ammaraskar: That makes sense, and it'd be a pretty easy change. Plus there's convenient `datetime.datetime.min/max` constants

ammaraskar: On second thought, I prefer `None` instead of -Inf -> Inf since the API explicitly only handles either line based OR time based filtering. Mixing both of them is not allowed because it makes no sense, right now in the API if anywhere you encounter a start/end_time you can assume it will be None if we aren't filtering by it

paasta_tools/cli/cmds/logs.py
```diff
@@ -422,65 +527,38 @@ def tail_logs(self, service, levels, components, clusters, raw_mode=False):
...
+        def callback(component, stream_info, scribe_env, cluster):
+            kw = {
+                'scribe_env': scribe_env,
+                'service': service,
+                'levels': levels,
+                'components': components,
+                'clusters': clusters,
+                'queue': queue,
+                'filter_fn': stream_info.filter_fn,
+
```
solarkennedy: delete empty line

paasta_tools/cli/cmds/logs.py
```diff
@@ -422,65 +527,37 @@ def tail_logs(self, service, levels, components, clusters, raw_mode=False):
...
-                        'clusters': [cluster],
-                        'queue': queue,
-                        'parse_fn': parse_chronos_log_line,
-                        'filter_fn': chronos_log_line_passes_filter,
-                    }
-                    process = Process(target=self.scribe_tail, kwargs=kw)
-                    spawned_processes.append(process)
-                    process.start()
+
+        def callback(component, stream_info, scribe_env, cluster):
```
solarkennedy: can you... give this function a better name? I don't have any great suggestions, how about `start_scribe_tail`?

paasta_tools/cli/cmds/logs.py
```diff
@@ -422,65 +527,37 @@ def tail_logs(self, service, levels, components, clusters, raw_mode=False):
...
+            process = Process(target=self.scribe_tail, kwargs=kw)
+            spawned_processes.append(process)
+            process.start()
+
+        # Scribe specific detail: all messages in app_output get logged at debug level, so we need to add that to
+        # the levels if its one of the components we're grabbing
+        if ('stdout' in components) or ('stderr' in components):
+            levels.append('debug')
+
+        self.run_code_over_scribe_envs(clusters, components, callback)
```
solarkennedy: can you kwargs this?

paasta_tools/cli/cmds/logs.py
solarkennedy: Oh I see, disregard

solarkennedy:
shipit

paasta_tools/cli/cmds/logs.py
ammaraskar: just for clarity?

paasta_tools/cli/cmds/logs.py
```diff
@@ -383,9 +419,42 @@ def print_logs_by_offset(self, service, line_count, offset, levels, components,
...
 class ScribeLogReader(LogReader):
     SUPPORTS_TAILING = True
+    SUPPORTS_LINE_COUNT = True
+    SUPPORTS_TIME = True
+
+    COMPONENT_STREAM_INFO = {
+        'default': ScribeComponentStreamInfo(
+            per_cluster=False,
+            stream_name_fn=get_log_name_for_service,
+            filter_fn=paasta_log_line_passes_filter, parse_fn=None
```
EvanKrall: I generally prefer to either do one param per line or break at 120 characters and not mix. Can you move parse_fn to the next line?

paasta_tools/cli/cmds/logs.py
```diff
@@ -549,6 +626,112 @@ def tail_logs(self, service, levels, components, clusters, raw_mode=False):
...
+            print_log(line, levels, raw_mode)
+
+    def filter_and_aggregate_scribe_logs(self, scribe_reader_ctx, scribe_env, stream_name,
+                                         levels, service, components, clusters,
+                                         aggregated_logs, parser_fn=None, filter_fn=None,
+                                         start_time=None, end_time=None):
+        # Scribe specific detail: all messages in app_output get logged at debug level, so we need to add that to
+        # the levels if its one of the components we're grabbing
+        if ('stdout' in components) or ('stderr' in components):
+            levels.append('debug')
```
EvanKrall: Does this mean that if we add the app_output component, we'll see all debug messages? Any way around that?

paasta_tools/cli/cmds/logs.py
```diff
@@ -549,6 +626,112 @@ def tail_logs(self, service, levels, components, clusters, raw_mode=False):
...
+        if start_time.date() == today or end_time.date() == today:
+            # The reason we need a fake context here is because scribereader is a bit incosistent in its
+            # returns. get_stream_reader returns a context that needs to be acquired for cleanup code but
+            # get_stream_tailer simply returns an object that can be iterated over. We'd still like to have
+            # the cleanup code for get_stream_reader to be executed by this function's caller and this is
+            # one of the simpler ways to achieve it without having 2 if statements everywhere that calls
+            # this method
+            @contextmanager
+            def fake_context():
+                yield scribereader.get_stream_tailer(stream_name, host, port, True, -1)
```
EvanKrall: I don't think this is correct - my understanding is that a stream tailer will only give you logs from now onwards, and if you want logs from earlier today, you'd want get_stream_reader.

e.g. the command `scribereader -e sfo2 -f stream_paasta_yelp-main` would give you logs starting now, but `scribereader -e sfo2 stream_paasta_yelp-main` would give you logs starting today at midnight and ending now.

paasta_tools/cli/cmds/logs.py
ammaraskar: That's a good point. Yeah there is a very simple way around it, stderr/stdout will have their own filter_fn that don't filter out debug levels.

EvanKrall:
Ship

ammaraskar:
https://github.com/Yelp/paasta/commit/70e1b8e919f5d1c35e81dbeb89f3bf408f6e7029

PR #567, opened 2016-06-08
Generalize cpu quota

matthewbentley:
Adds the cpu quota options to chronos and local-run

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -71,6 +71,9 @@ no_escape = re.compile('\x1B\[[0-9;]*[mK]')
 
 DEFAULT_SYNAPSE_HAPROXY_URL_FORMAT = "http://{host:s}:{port:d}/;csv;norefresh"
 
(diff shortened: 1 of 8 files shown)


paasta_tools/utils.py
```diff
@@ -143,6 +146,18 @@ def get_cpus(self):
...
         return cpus
 
+    def get_cpu_period(self):
+        return self.config_dict.get('cfs_period_us', DEFAULT_CPU_PERIOD)
+
+    def get_cpu_quota(self):
+        cpu_burst_pct = self.config_dict.get('cpu_burst_pct', DEFAULT_CPU_BURST_PCT)
+        return self.get_cpus() * self.get_cpu_period() * (100 + cpu_burst_pct) / 100
+
+    def get_docker_parameters(self):
```
EvanKrall: Actually, can we call this "format_docker_parameters", since this isn't a true getter, and add a docstring explaining what it's doing and what format it's in?

EvanKrall:
Fix and ship.

Issue #524, opened 2016-05-13
paasta status now shows 'connection closed' output


Rob-Johnson:
```
robj@mesos1-uswest1cdevc:~  % paasta status -s paasta_canary -i mesosstage_main -c mesosstage
Pipeline: https://jenkins.yelpcorp.com/view/services-paasta_canary

cluster: mesosstage
instance: mesosstage_main
Git sha:    c19f9731 (desired)
State:      Running - Desired state: Started
Marathon:   Healthy - up with (5/5) instances. Status: Running.
Mesos:      Healthy - (5/5) tasks in the TASK_RUNNING state.
Smartstack:
    uswest1-devc - Healthy - in haproxy with (5/5) total backends UP in this namespace.
Connection to 10.40.1.49 closed.
````

That last line is new (I think). Let's get rid of it either way

PR #569, opened 2016-06-08
Make paasta_serviceinit exit more suitably when a command fails.

ammaraskar:
# Exit Code on control + c #525
2 parts to this fix, first, don't blindly assume a run_paasta_serviceinit command succeeds, actually check the return code to make sure. 

Secondly within paasta_serviceinit, we're catching Exception, unfortunately KeyboardInterrupt is derived from BaseException so the catch doesn't get triggered.

# No more connection messages #524
These messages are produced by the ssh command because we run with the `-t` flag. However, we don't really need informative messages out of ssh in the status command, so I've added the ability to ignore the output of ssh itself when executing paasta serviceinit remotely.

This allows critical commands like emergency scale, stop etc to still have output but allows for selective suppression.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import sys
+from subprocess import CalledProcessError
 
 from service_configuration_lib import read_deploy
 
@@ -158,7 +159,8 @@ def report_status_for_cluster(service, cluster, deploy_pipeline, actual_deployme
 
     if len(deployed_instances) > 0:
         status = execute_paasta_serviceinit_on_remote_master('status', cluster, service, ','.join(deployed_instances),
-                                                             system_paasta_config, stream=True, verbose=verbose)
+                                                             system_paasta_config, stream=True, verbose=verbose,
+                                                             ignore_ssh_output=True)
         # Status results are streamed. This print is for possible error messages.
         if status is not None:
             for line in status.rstrip().split('\n'):
@@ -224,14 +226,19 @@ def paasta_status(args):
 
     if actual_deployments:
         deploy_pipeline = list(get_planned_deployments(service, soa_dir))
-        report_status(
-            service=service,
-            deploy_pipeline=deploy_pipeline,
-            actual_deployments=actual_deployments,
-            cluster_whitelist=cluster_whitelist,
-            instance_whitelist=instance_whitelist,
-            system_paasta_config=system_paasta_config,
-            verbose=args.verbose,
-        )
+        try:
+            report_status(
+                service=service,
+                deploy_pipeline=deploy_pipeline,
+                actual_deployments=actual_deployments,
+                cluster_whitelist=cluster_whitelist,
+                instance_whitelist=instance_whitelist,
+                system_paasta_config=system_paasta_config,
+                verbose=args.verbose,
+            )
+        except CalledProcessError as e:
+            print PaastaColors.grey(PaastaColors.bold(e.cmd + " exited with non-zero return code."))
+            print PaastaColors.grey(e.output)
+            return e.returncode
     else:
         print missing_deployments_message(service)
--- a/paasta_tools/cli/utils.py
+++ b/paasta_tools/cli/utils.py
@@ -475,7 +475,7 @@ def check_ssh_and_sudo_on_master(master, timeout=10):
     return (False, output)
 
 
-def run_paasta_serviceinit(subcommand, master, service, instances, cluster, stream, **kwargs):
+def run_paasta_serviceinit(subcommand, master, service, instances, cluster, stream, ssh_flags='', **kwargs):
     """Run 'paasta_serviceinit <subcommand>'. Return the output from running it."""
     if 'verbose' in kwargs and kwargs['verbose'] > 0:
         verbose_flag = ' '.join(['-v' for i in range(kwargs['verbose'])])
@@ -494,10 +494,11 @@ def run_paasta_serviceinit(subcommand, master, service, instances, cluster, stre
     else:
         delta_flag = ''
 
-    ssh_flag = '-t' if stream else '-n'
+    ssh_flags += ' -t' if stream else ' -n'
+    ssh_flags = ssh_flags.strip()
 
     command_parts = [
-        "ssh -A %s %s sudo paasta_serviceinit" % (ssh_flag, master),
+        "ssh -A %s %s sudo paasta_serviceinit" % (ssh_flags, master),
         "-s %s" % service,
         "-i %s" % instances,
         verbose_flag,
@@ -515,7 +516,7 @@ def run_paasta_serviceinit(subcommand, master, service, instances, cluster, stre
 
 
 def execute_paasta_serviceinit_on_remote_master(subcommand, cluster, service, instances, system_paasta_config,
-                                                stream=False, **kwargs):
+                                                stream=False, ignore_ssh_output=False, **kwargs):
     """Returns a string containing an error message if an error occurred.
     Otherwise returns the output of run_paasta_serviceinit_status().
     """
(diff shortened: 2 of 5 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -158,7 +158,8 @@ def report_status_for_cluster(service, cluster, deploy_pipeline, actual_deployme
 
     if len(deployed_instances) > 0:
         status = execute_paasta_serviceinit_on_remote_master('status', cluster, service, ','.join(deployed_instances),
-                                                             system_paasta_config, stream=True, verbose=verbose)
+                                                             system_paasta_config, stream=True, verbose=verbose,
+                                                             ignore_ssh_output=True)
```
nhandler: What is the point of this new `ignore_ssh_output`? I don't see it getting used here.

ammaraskar: It's a kwarg that gets used here: https://github.com/Yelp/paasta/pull/569/files/72466fe2a29c8a20deade5663dea95d2478812ea#diff-7d4e89c64360b719f36a1ca16d58607cR530

paasta_tools/cli/utils.py
```diff
@@ -507,12 +508,14 @@ def run_paasta_serviceinit(subcommand, master, service, instances, cluster, stre
     command_without_empty_strings = [part for part in command_parts if part != '']
     command = ' '.join(command_without_empty_strings)
     log.debug("Running Command: %s" % command)
-    _, output = _run(command, timeout=timeout, stream=stream)
+    return_code, output = _run(command, timeout=timeout, stream=stream)
+    if return_code != 0:
```
huadongliu: _run already catches exceptions and the timeout error. Why do we need another level of error handling here?

ammaraskar: `_run` is, from what I understand, just a general purpose function to run commands. It doesn't by itself raise any Exceptions if the command it ran doesn't exit with return code 0. As a little contrived example if you wanted to call `grep` with `_run`, you wouldn't want it to raise an Exception if grep returned a non 0 code. However, for critical things like running ssh, we should inspect the return code and react accordingly.

ammaraskar:
For anyone else reviewing this, it's probably easier to view the diff individually by the 2 commits rather than the combined diff.

paasta_tools/cli/utils.py
```diff
@@ -493,10 +494,10 @@ def run_paasta_serviceinit(subcommand, master, service, instances, cluster, stre
     else:
         delta_flag = ''
 
-    ssh_flag = '-t' if stream else '-n'
+    ssh_flags += '-t' if stream else '-n'
```
nhandler: I see that you are explicitly adding trailing whitespace when you pass in some `ssh_flags`. That seems like a very easy thing for someone to overlook in the future. What if we added the whitespace here instead?

paasta_tools/cli/cmds/status.py
keshavdv: I'm wary of what happens when SSH fails to connect. I'm assuming this change also hides that output (e.g "Connection refused..") and I'm not sure we have other messaging in place for these situations. Can you test this out to see what happens?

nhandler:
I would fix the whitespace issue, but it looks good to me.

paasta_tools/cli/cmds/status.py
ammaraskar: So that's one of the reasons why I added this flag here, for stuff like `status`, it isn't critical if an ssh connection fails, and even if it does this will toss up an exception because the return code won't be 0

paasta_tools/cli/utils.py
ammaraskar: That makes sense, I'll do that

ammaraskar:
@keshavdv @huadongliu Any more concerns?

huadongliu:
For the 1st commit, we do not assume run_paasta_serviceinit command always succeeds on the master. Exceptions and other error information about the command are caught at the master and sent back to the client. returncode on the client side is the execution status, not command status on the master. The original complaint was "paasta status" exits with code 0 on KeyboardInterrupt. I think the fix is to set returncode when KeyboardInterrupt is caught. 

For the 2nd commit, kwargs is used differently from the way it was used in current code. There might be confusion although it is right.

ammaraskar:
>For the 1st commit, we do not assume run_paasta_serviceinit command always succeeds on the master. Exceptions and other error information about the command are caught at the master and sent back to the client. returncode on the client side is the execution status, not command status on the master. The original complaint was "paasta status" exits with code 0 on KeyboardInterrupt. I think the fix is to set returncode when KeyboardInterrupt is caught.

I'm gonna try to explain the whole process:

1. You call `paasta status`
2. `paasta status` then goes and calls `ssh ...`
3. You do a control+c
4. This interrupts the `ssh` command which returns an error code
5. Because run_paasta_serviceinit ignores the return code, it doesn't realize an error occurred and returns normally.
6. The whole `paasta status` command has no indication anything went wrong and exits with code 0

Now with this commit the process goes like this:

1. You call `paasta status`
2. `paasta status` then goes and calls `ssh ...`
3. You do a control+c
4. This interrupts the `ssh` command which returns an error code
5. run_paasta_serviceinit notices that ssh didn't return 0 so it raises an Exception
6. The Exception goes to the top level handler and causes an error code to be returned

paasta_tools/cli/utils.py
```diff
@@ -507,7 +508,9 @@ def run_paasta_serviceinit(subcommand, master, service, instances, cluster, stre
     command_without_empty_strings = [part for part in command_parts if part != '']
     command = ' '.join(command_without_empty_strings)
     log.debug("Running Command: %s" % command)
-    _, output = _run(command, timeout=timeout, stream=stream)
+    return_code, output = _run(command, timeout=timeout, stream=stream)
+    if return_code != 0:
+        raise CalledProcessError(return_code, command, output)
```
huadongliu: In case of a master side exception, all information about the error (stack trace etc) will be sent over ssh. I am not sure we need to raise an exception on the client side (with a not quite related stack trace) or to simply exit with the error code.

ammaraskar: >I am not sure we need to raise an exception on the client side (with a not quite related stack trace) or to simply exit with the error code.

I think a good compromise is to just catch this exception in the `paasta status` handler and just print out a small message at the end informing the user that `ssh ...` returned with a non zero return code.

ammaraskar:
screenshot from 2016-06-14 14-10-31

huadongliu:
Shipit

paasta_tools/cli/cmds/status.py
```diff
@@ -224,14 +226,19 @@ def paasta_status(args):
...
+                service=service,
+                deploy_pipeline=deploy_pipeline,
+                actual_deployments=actual_deployments,
+                cluster_whitelist=cluster_whitelist,
+                instance_whitelist=instance_whitelist,
+                system_paasta_config=system_paasta_config,
+                verbose=args.verbose,
+            )
+        except CalledProcessError as e:
+            print PaastaColors.grey(PaastaColors.bold(e.cmd + " exited with non-zero return code."))
```
solarkennedy: can you print the return code?

paasta_tools/cli/utils.py
```diff
@@ -524,7 +528,11 @@ def execute_paasta_serviceinit_on_remote_master(subcommand, cluster, service, in
         return (
             'ERROR: could not find connectable master in cluster %s\nOutput: %s' % (cluster, output)
         )
-    return run_paasta_serviceinit(subcommand, master, service, instances, cluster, stream, **kwargs)
+    if ignore_ssh_output:
```
solarkennedy: ignore_ssh_output is kinda a misnomer right? It isn't really ignoring the ssh output as much as the... ssh logging?

Do you know why INFO was too noisy to create that "disconnection" message before? It didn't used to do that. Any particular reason not to just do QUIET all the time?

ammaraskar: It's because of the streaming mode, it sends a `-t` flag to ssh which makes it print out those extra messages. Yeah I guess the name is kinda misleading.

>Any particular reason not to just do QUIET all the time?

I feel like having all of SSH's output is useful for commands like an emergency stop, just in case. (they all share this same code path)

PR #571, opened 2016-06-09
Implement HTTP drain method.

EvanKrall:


paasta_tools/drain_lib.py
```diff
@@ -196,3 +196,89 @@ def is_safe_to_kill(self, task):
...
+            'GET': requests.get,
+            'POST': requests.post,
+            'PUT': requests.put,
+            'PATCH': requests.patch,
+            'DELETE': requests.delete,
+            'OPTIONS': requests.options,
+            'HEAD': requests.head,
+        }[method]
+
+        resp = requests_func(url)
```
keshavdv: What happens when this throws other exceptions (timeout, ConnectionError, etc)?

paasta_tools/drain_lib.py
```diff
@@ -196,3 +196,89 @@ def is_safe_to_kill(self, task):
             return False
         else:
             return info.get("since", 0) < (time.time() - self.delay)
+
+
+class StatusCodeNotAcceptableException(Exception):
```
keshavdv: Not a big deal but I think pep8 says "Error" rather than "Exception"

paasta_tools/drain_lib.py
EvanKrall: They would bubble up, and setup_marathon_job would likely crash.

keshavdv: That means we would end up getting notified as opposed to the service author and I'm not sure if that's the right thing to do. Assuming the general use case is for a container to notify paasta when it can be terminated, if the endpoint doesn't work as expected I don't think we can do much about it.

EvanKrall: Yeah, I guess we'll have to make sure the bounce code can handle failures of the drain method. Currently, failures of the HacheckDrainMethod don't happen often / are not typically persistent, so crashing setup_marathon_job is fine.

If calling `drain()` fails (due to connection failure, timeout, bad status code, etc.), we could:

 - Ignore, which would mean we'd retry the next time. If the URL is bad or something, we'd probably not make progress on the bounce, but if it's just a temporary issue we'd eventually continue.
 - Kill the container immediately, which would let the bounce proceed in the face of persistent errors.

If calling `stop_draining()` fails, we could:
 - ignore; we might end up having a bunch of tasks stuck in draining.
 - pass the task into the bounce method as a draining task, which would mean we might kill it, depending on the state of the bounce.

If `is_draining` fails, we could:
 - assume the task is or is not draining.

If `is_safe_to_kill` fails, we could:
 - assume the task is or is not safe to kill

EvanKrall:
@keshavdv I've added error handling; can you take another look?

paasta_tools/setup_marathon_job.py
```diff
@@ -341,6 +368,18 @@ def get_old_happy_unhappy_draining_tasks(other_apps, drain_method, service, nerv
     return old_app_live_happy_tasks, old_app_live_unhappy_tasks, old_app_draining_tasks
 
 
+def undrain_tasks(to_undrain, leave_draining, drain_method, log_deploy_error):
+    # If any tasks on the new app happen to be draining (e.g. someone reverts to an older version with
+    # `paasta mark-for-deployment`), then we should undrain them.
+    for task in to_undrain:
+        if task not in leave_draining:
+            # TODO: don't take actions in deploy_service.
```
keshavdv: Is this still a todo?

keshavdv:
shipit!

paasta_tools/setup_marathon_job.py
EvanKrall: Yeah, I'd still like to avoid this, but it would require a more extensive refactor -- passing the list of things to drain/undrain into do_bounce. For the sake of expediency I'm going to avoid that for now.

I am, however, going to move this comment to deploy_service where we call undrain_tasks.

PR #572, opened 2016-06-09
set utilization to 0 on zero division error

Rob-Johnson:
if a slave has 0 available and 0 usage, then catch the ZeroDivisionError
and set usage to 0


giuliano108:
ShipIt.

PR #573, opened 2016-06-10
Add kill threshold

mattmb:
* Makes cleanup script more defensive. If it is going to delete more than half the marathon services it will now exit by default. It can be forced manually.
* Add integration tests for the cleanup script.

### DIFF
--- a/paasta_tools/cleanup_marathon_jobs.py
+++ b/paasta_tools/cleanup_marathon_jobs.py
@@ -27,9 +27,13 @@ Command line options:
 
 - -d <SOA_DIR>, --soa-dir <SOA_DIR>: Specify a SOA config dir to read from
 - -v, --verbose: Verbose output
+- -t <KILL_THRESHOLD>, --kill-threshold: The decimal fraction of apps we think
+    is sane to kill when this job runs
+- -f, --force: Force the killing of apps if we breach the threshold
 """
 import argparse
 import logging
+import sys
 import traceback
 
 import pysensu_yelp
@@ -47,13 +51,25 @@ from paasta_tools.utils import load_system_paasta_config
 log = logging.getLogger(__name__)
 
 
+class DontKillEverythingError(Exception):
+    pass
+
+
 def parse_args():
     parser = argparse.ArgumentParser(description='Cleans up stale marathon jobs.')
     parser.add_argument('-d', '--soa-dir', dest="soa_dir", metavar="SOA_DIR",
                         default=DEFAULT_SOA_DIR,
                         help="define a different soa config directory")
+    parser.add_argument('-t', '--kill-threshold', dest="kill_threshold",
+                        default=0.5,
+                        help="The decimal fraction of apps we think is "
+                             "sane to kill when this job runs")
     parser.add_argument('-v', '--verbose', action='store_true',
                         dest="verbose", default=False)
+    parser.add_argument('-f', '--force', action='store_true',
+                        dest="force", default=False,
+                        help="Force the cleanup if we are above the "
+                             "kill_threshold")
     args = parser.parse_args()
     return args
 
@@ -118,12 +134,15 @@ def delete_app(app_id, client, soa_dir):
         raise
 
 
-def cleanup_apps(soa_dir):
+def cleanup_apps(soa_dir, kill_threshold=0.5, force=False):
     """Clean up old or invalid jobs/apps from marathon. Retrieves
     both a list of apps currently in marathon and a list of valid
     app ids in order to determine what to kill.
 
-    :param soa_dir: The SOA config directory to read from"""
+    :param soa_dir: The SOA config directory to read from
+    :param kill_threshold: The decimal fraction of apps we think is
+        sane to kill when this job runs.
+    :param force: Force the cleanup if we are above the kill_threshold"""
     log.info("Loading marathon configuration")
     marathon_config = marathon_tools.load_marathon_config()
     log.info("Connecting to marathon")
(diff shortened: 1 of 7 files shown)


paasta_tools/utils.py
```diff
@@ -1548,3 +1548,7 @@ def __exit__(cls, *args, **kwargs):
             cls.zk.stop()
             cls.zk.close()
             cls.zk = None
+
+
+class DontKillEverything(Exception):
```
Rob-Johnson: you can put this back in ``cleanup_marathon_jobs``, it's not really generalizable. Further, can you name it properly ``DontKillEverythingError`` please?

paasta_tools/cleanup_marathon_jobs.py
```diff
@@ -118,45 +131,67 @@ def delete_app(app_id, client, soa_dir):
...
     if args.verbose:
         logging.basicConfig(level=logging.DEBUG)
     else:
         logging.basicConfig(level=logging.WARNING)
-
-    cleanup_apps(soa_dir)
+    try:
+        cleanup_apps(soa_dir, kill_threshold=kill_threshold, force=force)
+    except DontKillEverything:
+        sys.exit(1)
```
Rob-Johnson: does it even reach this? doesn't log.critical exit out?

paasta_tools/cleanup_marathon_jobs.py
```diff
@@ -118,45 +131,67 @@ def delete_app(app_id, client, soa_dir):
...
-            delete_app(
-                app_id=app_id,
-                client=client,
-                soa_dir=soa_dir,
-            )
+        running_apps.append(app_id)
+    apps_to_kill = [running_app for running_app in running_apps
+                    if running_app[0:2] not in valid_apps]
+
+    log.debug("Running apps: %s" % running_apps)
```
Rob-Johnson: could you do something like
```
log.debug([app.name for app in running_apps])
```

so that the output is more readable please?

paasta_tools/cleanup_marathon_jobs.py
```diff
@@ -118,45 +131,67 @@ def delete_app(app_id, client, soa_dir):
...
+    log.debug("Terminating: %s" % apps_to_kill)
+    if running_apps:
+        above_kill_threshold = float(len(apps_to_kill)) / float(len(running_apps)) > float(kill_threshold)
+        if above_kill_threshold and not force:
+            log.critical("Paasta was about to kill more than %s of the running services, this "
+                         "is probably a BAD mistake!, run again with --force if you "
+                         "really need to destroy everything" % kill_threshold)
+            raise DontKillEverything
+    for running_app in apps_to_kill:
+        app_id = ".".join(running_app)
```
Rob-Johnson: can you use ``marathon_tools.compose_app_id`` please

paasta_tools/cleanup_marathon_jobs.py
```diff
@@ -118,45 +131,67 @@ def delete_app(app_id, client, soa_dir):
...
             continue
-        if (service, instance) not in valid_services:
-            delete_app(
-                app_id=app_id,
-                client=client,
-                soa_dir=soa_dir,
-            )
+        running_apps.append(app_id)
+    apps_to_kill = [running_app for running_app in running_apps
+                    if running_app[0:2] not in valid_apps]
```
Rob-Johnson: the list comprehension is neater, but the `[0:2]` isn't very readable. how about
```
[(service,instance) for service, instance in running_apps if (service, instance) not in valid_apps]
```

paasta_tools/cleanup_marathon_jobs.py
Rob-Johnson: nm, that's log.fatal

paasta_tools/utils.py
mattmb: Sure sounds sensible. Have we thought about an errors module?

paasta_tools/cleanup_marathon_jobs.py
mattmb: Good shout, will do

paasta_tools/cleanup_marathon_jobs.py
mattmb: I also felt that the `[0:2]` was a bit nasty, that version is much more readable 👍

mattmb:
Thanks @Rob-Johnson I'll fix those things! Do we generally use fixup commits and then squashing down before we merge?

nhandler:
There really is no formal policy. I will occasionally squash my fixup commits before merging, but I know other people really like having the history available so they can say, "I started out this way, but I made changes A and B based on feedback from so-and-so". Do whatever you think would be the most useful.

paasta_tools/cleanup_marathon_jobs.py
mattmb: I think I'm going to leave this, I think it could be useful in debug mode to see the whole `app_id` tuple so you can see exactly what's happening.

mattmb:
@nhandler Thanks, yeah I like the "fixup then squash" pattern for relatively small changes like this. When you're changing something big then I would definitely advocate a "proper" commit on top explaining what and why.

mattmb:
@Rob-Johnson let me know if you're happy with those fixes and I'll squash down before we merge 😄

paasta_tools/cleanup_marathon_jobs.py
```diff
@@ -118,45 +134,68 @@ def delete_app(app_id, client, soa_dir):
...
+        sane to kill when this job runs.
+    :param force: Force the cleanup if we are above the kill_threshold"""
     log.info("Loading marathon configuration")
     marathon_config = marathon_tools.load_marathon_config()
     log.info("Connecting to marathon")
     client = marathon_tools.get_marathon_client(marathon_config.get_url(), marathon_config.get_username(),
                                                 marathon_config.get_password())
 
-    valid_services = get_services_for_cluster(instance_type='marathon', soa_dir=soa_dir)
+    valid_apps = get_services_for_cluster(instance_type='marathon', soa_dir=soa_dir)
```
Rob-Johnson: can you make this 'valid_services' again here please - it makes it easier to understand that this is the services declared in soa-configs vs the 'apps' that are running in marathon

Rob-Johnson:
fix and ship!

paasta_tools/cleanup_marathon_jobs.py
mattmb: :-P so I'd renamed them because it was confusing me that we were comparing running_apps and valid_services. Does marathon not call them services too?

Rob-Johnson: no, marathon has apps. we try and stick with service/instance as late as possible (right up to the api client call). I don't feel too strongly about it though, so go ahead and shipit if you do 👍

mattmb: So it does. Think you're right, it'd be better to stay consistent across the codebase, I'll switch it back 😄

EvanKrall: It's longer, but maybe `valid_service_instances`?

PR #574, opened 2016-06-10
Correct healthcheck_cmd docs

mattmb:
Correct mistake in docs. We no longer default to /bin/true:
97091a889f1315318533bfd1033b96e95d66674e

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -81,63 +81,6 @@ def get_schema(file_type):
     return json.loads(schema)
 
 
-def validate_healthcheck_cmd_has_cmd(config_file_object):
-    """Check that healthcheck_cmd is included if healthcheck_mode
-    is set to cmd
-
-    :param config_file_object: the loaded config object
-    """
-    for config in config_file_object.values():
-        if 'healthcheck_mode' in config:
-            if config['healthcheck_mode'] == 'cmd':
-                assert 'healthcheck_cmd' in config,\
-                       "healthcheck_cmd is required if healthcheck_mode"\
-                       " == cmd"
-
-
-def validate_schema_dependencies(file_path, file_type):
-    """Check for conditional dependencies in our schemas, this is
-    difficult and sometimes not possible with json schema
-
-    :param file_path: path to file to validate
-    :param file_type: what schema type should we validate against
-    """
-    basename = os.path.basename(file_path)
-    extension = os.path.splitext(basename)[1]
-    config_file_object = get_config_object(file_path, extension)
-    try:
-        if file_type == 'marathon':
-            validate_healthcheck_cmd_has_cmd(config_file_object)
-    except AssertionError as e:
-        print '%s: %s' % (SCHEMA_INVALID, file_path)
-        print '  Validation Message: %s' % e.message
-        return False
-    else:
-        print '%s: %s' % (SCHEMA_VALID, basename)
-        return True
-
-
-def get_config_object(file_path, extension):
-    """Open and attempt to deserialise a config file
-    returns the deserialised object
-
-    :param file_path: path to file to validate
-    :param extension: file extension
-    """
-    try:
-        config_file = get_file_contents(file_path)
-    except IOError:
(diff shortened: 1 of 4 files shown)


Rob-Johnson:
shipit.

I think the 'right' fix here is to just default to '/bin/true' and just consider the service healthy so long as the process is alive, but this will do for now.

keshavdv:
What do service authors expect to happen when they set mode to 'cmd' and don't specify an actual command (I'm not sure if tron did something by default)? Is there a reason they don't just omit the healthcheck entirely in that case (the reason we stopped defaulting to /bin/true was to try and avoid a bunch of unnecessary docker execs) or we just pretend they haven't specified any healthchecks?

solarkennedy:
I agree with @keshavdv, I don't think there can really be a good default for healthcheck_cmd, /bin/true is just wasteful.

solarkennedy:
ship

Rob-Johnson:
that's fine with me, but if we're not providing a default, let's add some validation to make sure this fails the pre-commit hook?

mattmb:
Agree with all, I did think about adding the default back but I think it's sensible to make it required. Haven't looked at the git hooks yet but I will before we merge this 👍

Rob-Johnson:
I doubt you can express that kind of thing in json schema, so It'll probably live in ``paasta_tools/cli/cmds/validate.py``

mattmb:
Yeah, I found the json schemas but haven't tried adding that validation yet. Haven't looked at the cli `validate.py`. Does that run on commit too?

Rob-Johnson:
yeah, iirc it's ``validate.py`` that converts YAML ->  JSON and checks the schema?

paasta_tools/cli/cmds/validate.py
```diff
@@ -81,19 +81,49 @@ def get_schema(file_type):
...
+def validate_healthcheck_cmd_has_cmd(config_file_object):
+    """Check that healthcheck_cmd is included if healthcheck_mode
+    is set to cmd
+
+    :param config_file_object: the loaded config object
+    """
+    for config in config_file_object.values():
+        if 'healthcheck_mode' in config:
+            if config['healthcheck_mode'] == 'cmd':
+                assert 'healthcheck_cmd' in config,\
```
solarkennedy: Does this work instead?
http://stackoverflow.com/a/11652478

tests/cli/test_cmds_validate.py
```diff
@@ -172,6 +173,52 @@ def test_marathon_validate_schema_list_hashes_good(
...
+main_worker:
+  cpus: 0.1
+  instances: 2
+  mem: 250
+  disk: 512
+  cmd: virtualenv_run/bin/python adindexer/adindex_worker.py
+  healthcheck_mode: cmd
+"""
+    mock_get_file_contents.return_value = marathon_content
+
```
solarkennedy: This looks a little strange with so many newlines inbetween the code, is this intentional?

paasta_tools/cli/cmds/validate.py
mattmb: That example actually looks better than the ones I found on SOF. I need to check it works with the healthcheck_mode itself being optional too. I'll give it another go, I'd definitely prefer if it was possible within the schema!

tests/cli/test_cmds_validate.py
mattmb: Definitely not, I guess I'll blame incorrect blood caffeine levels ;-)

paasta_tools/cli/schemas/marathon_schema.json
```diff
@@ -7,6 +7,20 @@
...
+                "properties": {
+                    "healthcheck_mode": {"enum": ["tcp", "http"]}
+                }
+            },
+            {
+                "properties": {
+                    "healthcheck_mode": {"enum": ["cmd"]},
+                    "healthcheck_cmd": {"type": "string"}
+                },
+                "required": ["healthcheck_cmd"]
```
mattmb: @solarkennedy got it working with json schema. There's a few non obvious changes between v3 and v4 e.g. you have to specify required keys like this now.

Rob-Johnson:
shipit

PR #575, opened 2016-06-10
Reduce logging verbosity during tests

### LINKED ISSUE #503: Make tests less verbose
As side-effects of our tests, lots of log entries get printed to the terminal. Since we test failure situations, lots of this text is red and mentions errors or other problems. This can make it a bit confusing when looking through the test output to try and understand why they failed.

It would be nice to clean this up a bit so the output is less verbose (or possibly allow setting the verbosity with an ENV variable).

keshavdv:
Fixes #503. Since we pass through arguments anyway, you can get full output with (stdout for a failing test is still displayed):

`tox -- -s`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,7 @@ deps =
     --requirement={toxinidir}/requirements-dev.txt
     --editable={toxinidir}
 commands =
-    py.test -s {posargs:tests}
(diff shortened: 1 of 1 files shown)


Rob-Johnson:
shipit

Issue #576, opened 2016-06-10
Use "real" configs for itests


matthewbentley:
eg, `paasta_itsests/steps/bounces_steps.py:49` hard codes a config.  We should have itests that test the config parsing/generation, either by modifying existing itests or creating new ones to do so.  Modifying the existing itests would also mean additions to the configs (eg #564, #567) wouldn't need to also change the itest setups.

Relates to #373, as it will be a lot easier to use real configs if we use docker to run the tests.

matthewbentley:
This would also help prevent diff like [this](https://github.com/Yelp/paasta/files/309553/docker.txt).

EvanKrall:
I'm attempting a refactor of setup_marathon_job and I'm incidentally doing this.

PR #577, opened 2016-06-10
Use docker containerizer to run itests.

### LINKED ISSUE #373: Use docker containerizer in itests

matthewbentley:
TL;DR: this tells the mesosslave to use docker to run itests, and bind mounts /var/run/docker.sock and /usr/bin/docker into the mesosslave(s) to use the host's docker.
It also updates a bunch of tests to actually work in docker
Closes #373
Should be followed up by something to deal with #576

### DIFF
--- a/paasta_itests/docker-compose.yml
+++ b/paasta_itests/docker-compose.yml
@@ -22,7 +22,6 @@ mesosslave:
   hostname: mesosslave.test_hostname
   volumes:
     - /var/run/docker.sock:/var/run/docker.sock
-    - /usr/bin/docker:/usr/bin/docker
 
 marathon:
   build: ../yelp_package/dockerfiles/itest/marathon/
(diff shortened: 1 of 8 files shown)


matthewbentley:
`make itest` works, but I'm not sure how this will interact with travis

Rob-Johnson:
this is great.

shipit!

matthewbentley:
OK, pretend I didn't try to pin the right docker version. If this travis build passes, we're probably good to go.

matthewbentley:
Well, it failed. That was unexpected. I guess we have tests for a reason

EvanKrall:
Ship!

PR #578, opened 2016-06-14
Paasta start/stop/restart multiple instances

mattmb:
This adds support to the paasta cli for specifying multiple instances when starting or stopping. This should make it easier to bounce all the instances for a service in one go or to specify a specific set of instances.

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -57,8 +57,9 @@ def add_subparser(subparsers):
         ).completer = lazy_choices_completer(list_instances)
         status_parser.add_argument(
             '-c', '--clusters',
-            help="A comma-separated list of clusters to view. Defaults to view all clusters.\n"
-            "For example: --clusters norcal-prod,nova-prod"
+            help="A comma-separated list of clusters to view. "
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -159,6 +159,11 @@ def paasta_start_or_stop(args, desired_state):
     else:
         clusters = list_clusters(service)
```
mattmb: Haven't changed this here but this line actually makes me a little nervous. Defaulting to all the instances for a service seems safe enough. But defaulting to all clusters makes me worried I might start/stop things in prod just by forgetting to specify a cluster? Anyone else think we should change this?

Rob-Johnson: hm, yeah, I kind of agree. I don't know what a sane default might be though, so we should probably require a cluster?

Rob-Johnson:
fix and ship

paasta_tools/cli/cmds/start_stop_restart.py
mattmb: Yeah, that'd be my preference, set it as required by argparse.

PR #580, opened 2016-06-15
Make autoscaling wait for tasks to be healthy

### LINKED ISSUE #472: Service autoscaling doesn't wait for healthy tasks
Service autoscaling only waits for a service to have healthcheck results, but it doesn't make sure it's healthy before it considers the service for autoscaling

mattmb:
This fixes #472. Previously we were only checking for a health check
result and not the actual content of the results (as the documentation
already states). Now we check that the a task is healthy before we
consider it for autoscaling.

### DIFF
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -31,6 +31,7 @@ from paasta_tools.bounce_lib import ZK_LOCK_CONNECT_TIMEOUT_S
 from paasta_tools.marathon_tools import compose_autoscaling_zookeeper_root
 from paasta_tools.marathon_tools import format_job_id
 from paasta_tools.marathon_tools import get_marathon_client
+from paasta_tools.marathon_tools import is_task_healthy
 from paasta_tools.marathon_tools import load_marathon_config
 from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.marathon_tools import MESOS_TASK_SPACER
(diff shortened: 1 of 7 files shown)


paasta_tools/autoscaling_lib.py
```diff
@@ -344,6 +344,16 @@ def humanize_error(error):
         return 'utilization within thresholds'
 
 
+def is_task_healthy(task):
+    """
+    Check a marathon task has healthcheck results
+    and check that they are all healthy.
+    """
+    if task.health_check_results:
+        return all([hcr.alive for hcr in task.health_check_results])
```
solarkennedy: This is kinda duplicating paasta_tools/bounce_lib.py:264, also paasta_tools/marathon_serviceinit.py:147

Can you combine these together in a marathon_tools.py function?

mattmb: Good shout @solarkennedy I'll consolidate that down tomorrow 👍 

Also spotted https://github.com/Yelp/paasta/blob/369273ba8944d88b80517eca3ef2f2f3f80e1339/paasta_tools/check_marathon_services_replication.py#L196 now that I've grep'd the repo. Although that seems to be ignoring young tasks, presumably to prevent firing alerts on brand new tasks that are starting? One minute seems kinda arbitrary though...

solarkennedy: Yea if there are no healthchecks then we have to use *something* to make a heuristic about a task being healthy. 1m is better than nothing, basically we want to make sure that constantly churning tasks are not considered "healthy"

paasta_tools/marathon_tools.py
```diff
@@ -1076,3 +1076,23 @@ def get_instances_from_zookeeper(service, instance):
     with ZookeeperPool() as zookeeper_client:
         (instances, _) = zookeeper_client.get('%s/instances' % compose_autoscaling_zookeeper_root(service, instance))
         return int(instances)
+
+
+def is_task_healthy(task, require_all=True, default_healthy=False):
+    """
+    Check that a marathon task is healthy
```
Rob-Johnson: Fix the formatting here please - this should be on the same line as the `"""`. Could you also add the `:returns` please

Rob-Johnson:
fix and 
image

PR #583, opened 2016-06-16
Add app_output to default paasta logs componenets, some minor cleanup…

ammaraskar:
… in logs

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -56,7 +56,7 @@ from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import get_log_name_for_service
 
 
-DEFAULT_COMPONENTS = ['build', 'deploy', 'monitoring']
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/logs.py
```diff
@@ -698,7 +686,7 @@ def filter_and_aggregate_scribe_logs(self, scribe_reader_ctx, scribe_env, stream
                     if filter_fn:
                         if filter_fn(line, levels, service, components, clusters,
                                      start_time=start_time, end_time=end_time):
-                            aggregated_logs.append(line)
+                            print_log(line, levels, raw_mode)
```
solarkennedy: but if you print right away without aggregating, will the logs be in the right order?

ammaraskar: This PR won't change the order, since it was just being appended to an array before. The order seemed fine before so I don't think it'll be a problem.

solarkennedy:
ship

ammaraskar:
https://github.com/Yelp/paasta/commit/20f8036b0152323f79a74742580d81fdc3136ef0

PR #586, opened 2016-06-17
ceil() when using marging_factor

oktopuz:
to work well when the number of instances is small


EvanKrall:
Looks good to me, will merge once travis finishes.

PR #591, opened 2016-06-22
add the -g and -H to the cli entrypoint for metastatus

Rob-Johnson:
internal PAASTA-4949

### DIFF
--- a/paasta_tools/cli/cmds/metastatus.py
+++ b/paasta_tools/cli/cmds/metastatus.py
@@ -69,7 +69,7 @@ def add_subparser(subparsers):
             'Note: This is only effective with -vv'
         )
     )
-    status_parser.add_argument('-H', '--humanize', action='store_true', dest="humanize", default=False,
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/metastatus.py
```diff
@@ -59,15 +59,33 @@ def add_subparser(subparsers):
...
+        '-g',
+        '--groupings',
+        nargs='+',
+        default=['region'],
+        help=(
+            'Group resource information of slaves grouped by attribute.'
+            'Note: This is only effective with -vv'
+        )
+    )
+    status_parser.add_argument('-H', '--humanize', action='store_true', dest="humanize", default=False,
```
solarkennedy: Is there any reason *not* to make it human readable by default and make our computers remember to pass in the false when they know they don't want human readable units?

solarkennedy:
ship

paasta_tools/cli/cmds/metastatus.py
Rob-Johnson: hm, no. I'll make the default in the cli entrypoint True

PR #592, opened 2016-06-22
Tolerate autoscaling for tasks with no healthcheck

mattmb:
This changes the logic in the service autoscaling lib to autoscale
services that don't have any healthcheck defined. If a service does have
a healthcheck defined the logic should be unchanged, i.e. we wait for
the healthcheck to pass before we autoscale.

### DIFF
--- a/docs/source/autoscaling.rst
+++ b/docs/source/autoscaling.rst
@@ -27,11 +27,12 @@ a situation where there is a loss of autoscaling state, a service up is the
 safest course of action until the autoscaler can scale back down to a "normal"
 instance count.
 
-Autoscaling only considers marathon tasks that have passing health checks. This
-is so new tasks that have higher-than-average load when starting up are ignored
-until their first health check passes. PaaSTA will not autoscale a service
-without any health checks, though these services can be autoscaled using the
-``bespoke`` decision policy.
+If you use autoscaling it is highly recommended that you also configure a
+healthcheck. This ensures that PaaSTA will not autoscale marathon tasks that are
+just starting up or unhealthy. If you have no healthcheck defined autoscaling will
+run as soon as the service is deployed. If you have defined a healthcheck,
+autoscaling will run as soon as the healthcheck passes. This is so new tasks that
+have higher-than-average load when starting up are ignored.
 
 Autoscaling parameters are stored in an ``autoscaling`` attribute of your
 instances as a dictionary. Within the ``autoscaling`` attribute, setting a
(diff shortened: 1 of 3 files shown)


mattmb:
We should be mindful that when we ship this we will probably start scaling some services that haven't been scaled before.

solarkennedy:
ship

docs/source/autoscaling.rst
```diff
@@ -27,11 +27,12 @@ a situation where there is a loss of autoscaling state, a service up is the
 safest course of action until the autoscaler can scale back down to a "normal"
 instance count.
 
-Autoscaling only considers marathon tasks that have passing health checks. This
-is so new tasks that have higher-than-average load when starting up are ignored
-until their first health check passes. PaaSTA will not autoscale a service
-without any health checks, though these services can be autoscaled using the
-``bespoke`` decision policy.
+If you use autoscaling it is highly recommended that you also configure a
```
Rob-Johnson: I'd like us to be clearer about what's going on here: specifically, this is a work around for some bug/behaviour in Marathon that we don't understand the cause of that's triggered when we use cmd healthchecks, **not**, as is hinted at here, a result of apps being launched without healthchecks.

For whatever reason, when using cmd healthchecks, Marathon loses track of the health of the task. Whether this is a fault in Marathon or Mesos is unknown. Marathon's behaviour in these circumstances is to exclude the ``healthchecks`` field from it's response to the ``/v2/apps/{{app_id}}`` endpoint.  We've seen this in a few places, internally, (PAASTA-4565, the chat I had with svens about the push-notifications-worker yesterday). This isn't actually indicative of a missing healthcheck: **all** Marathon apps created by PaaSTA have a healthcheck associated with them (whether the values of that healthcheck are defaults or specified in soa-configs is besides the point). It's just a bug somewhere else that manifests as a 'missing' healthcheck.

So, It'd be good if we can clarify this here. Rather than mentioning about 'missing' healthchecks, let's clarify that this is actually when we can't figure out if the task is healthy or not according to the custom healthcheck, in which case, we'll fall back to the task being 'up' as evidence of it's health. 

> The autoscaler will only modify services where all of the currently running tasks are up. In the case that the health of a task is unknown according to the healthcheck (occasionally seen in the case of apps that use a ``cmd`` healthcheck), but is still running, the autoscaler considers it to be healthy. It may be that you want to define a custom healthcheck for your service so that it is not considered 'healthy' until it is fully in service (say, if there is an associated warm up period for your worker) to prevent the autoscaler considering the higher-than-average load at startup as an indicator that it needs to increase the number of workers available.

Rob-Johnson:
shipit!

docs/source/autoscaling.rst
mattmb: Yup, interesting background for me anyway. Yesterday I actually tried creating without an app without a healthcheck just to try this out.

PR #594, opened 2016-06-23
Make itests use "real" configs by having them call setup_service instead of deploy_service

### LINKED ISSUE #576: Use "real" configs for itests

EvanKrall:
Fixes #576.

I also found that some of the tests were calling `/bin/sleep infinity` inside a container, which makes the metastatus tests a race condition -- sometimes the container dies and the resource allocation gets cleaned up before metastatus is run.

This change also paves the way for me to refactor setup_service/deploy_service/do_bounce.

### DIFF
--- a/yelp_package/dockerfiles/itest/mesos/Dockerfile
+++ b/yelp_package/dockerfiles/itest/mesos/Dockerfile
@@ -17,7 +17,7 @@ RUN echo "deb http://repos.mesosphere.com/ubuntu trusty main" > /etc/apt/sources
 RUN echo "deb http://apt.dockerproject.org/repo ubuntu-trusty main" > /etc/apt/sources.list.d/docker.list
 RUN apt-key adv --keyserver keyserver.ubuntu.com --recv 58118E89F3A912897C070ADBF76221572C52609D
 RUN apt-key adv --keyserver keyserver.ubuntu.com --recv 81026D0004C44CF7EF55ADF8DF7D54CBE56151BF
-RUN apt-get update && apt-get -y install libsasl2-modules mesos=0.28.2-2.0.27.ubuntu1404 docker-engine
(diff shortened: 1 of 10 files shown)


Rob-Johnson:
shipit

yelp_package/dockerfiles/itest/mesos/Dockerfile
```diff
@@ -14,8 +14,11 @@
 
 FROM ubuntu:trusty
 RUN echo "deb http://repos.mesosphere.com/ubuntu trusty main" > /etc/apt/sources.list.d/mesosphere.list
+RUN echo "deb http://apt.dockerproject.org/repo ubuntu-trusty main" > /etc/apt/sources.list.d/docker.list
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv 58118E89F3A912897C070ADBF76221572C52609D
 RUN apt-key adv --keyserver keyserver.ubuntu.com --recv 81026D0004C44CF7EF55ADF8DF7D54CBE56151BF
-RUN apt-get update && apt-get -y install libsasl2-modules mesos=0.28.2-2.0.27.ubuntu1404 docker.io
+RUN apt-get update && apt-get -y install libsasl2-modules mesos=0.28.2-2.0.27.ubuntu1404 docker-engine
```
EvanKrall: I should probably pin this.

solarkennedy:
I think this is fine, ship after fixing tests

PR #595, opened 2016-06-23
PAASTA-4980 fix regression in -v option for metastatus

Rob-Johnson:
I guess for whatever reason ``-v -v -v`` doesn't work anymore, and you have to have ``-vvv`` instead in metastatus. Probably happened in the recent rewrite.


solarkennedy:
ship

PR #598, opened 2016-06-24
Changes to cluster autoscaling

mattmb:
This is a WIP (needs testing, tests and docs etc.)

Just putting it up for early comments on the direction.

Also needs https://github.com/Yelp/paasta/pull/521/ to make the draining work and (I think) further changes to the maintenance scripts to prevent allocation of tasks to slaves in maintenance mode.

### DIFF
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -551,7 +551,7 @@ def pick_slaves_to_kill(mesos_state, number, pool='default'):
         return []
 
 
-def wait_and_terminate(instances_to_kill):
+def wait_and_terminate(instances_to_kill, dry_run):
     """Kills instances after waiting for slave to be running 0 mesos tasks
 
     :param instances_to_kill: dict of instances to kill {<slave_id>: <instance_id>}"""
@@ -567,7 +567,7 @@ def wait_and_terminate(instances_to_kill):
                     if is_ready_to_kill:
                         log.debug("Instance {0}: ready to kill")
                         log.info("TERMINATING: {0}".format(instance_id))
-                        ec2_client.termintate_instances(InstanceIds=[instance_id])
+                        ec2_client.termintate_instances(InstanceIds=[instance_id], DryRun=dry_run)
                         instances_to_kill.pop(slave_id)
                     log.debug("Instance {0}: NOT ready to kill")
                 log.debug("Waiting 5 seconds and then checking again")
@@ -575,11 +575,11 @@ def wait_and_terminate(instances_to_kill):
     except TimeoutError:
         log.error("Timed out after {0} waiting to drain {1}, now terminating anyway".format(CLUSTER_DRAIN_TIMEOUT,
                                                                                             instances_to_kill))
-        ec2_client.termintate_instances(InstanceIds=instances_to_kill)
+        ec2_client.termintate_instances(InstanceIds=instances_to_kill, DryRun=dry_run)
 
 
 @register_autoscaling_component('aws_spot_fleet_request', SCALER_KEY)
-def scale_aws_spot_fleet_request(resource, delta, target_capacity, slaves_to_kill):
+def scale_aws_spot_fleet_request(resource, delta, target_capacity, slaves_to_kill, dry_run):
     """Scales a spot fleet request by delta to reach target capacity
     If scaling up we just set target capacity and let AWS take care of the rest
     If scaling down we pick the slaves we'd prefer to kill, put them in maintenance
@@ -596,7 +596,8 @@ def scale_aws_spot_fleet_request(resource, delta, target_capacity, slaves_to_kil
         return
     elif delta > 0:
         return ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=target_capacity,
-                                                    ExcessCapacityTerminationPolicy='noTermination')
+                                                    ExcessCapacityTerminationPolicy='noTermination',
+                                                    DryRun=dry_run)
     elif delta < 0:
         slaves_by_ip = {slave: slave_pid_to_ip(slave) for slave in slaves_to_kill}
         instances_to_kill = {slave_id: get_instance_id_from_ip(ip) for slave_id, ip in slaves_by_ip.items()}
@@ -607,9 +608,10 @@ def scale_aws_spot_fleet_request(resource, delta, target_capacity, slaves_to_kil
         # mesos should put the slave back into the pool
         duration = 600
         drain(slaves_by_ip.values(), start, duration)
-        wait_and_terminate(instances_to_kill)
-        return ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=target_capacity,
-                                                    ExcessCapacityTerminationPolicy='noTermination')
+        ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=target_capacity,
+                                             ExcessCapacityTerminationPolicy='noTermination',
+                                             DryRun=dry_run)
+        wait_and_terminate(instances_to_kill, dry_run)
 
 
 class ClusterAutoscalingError(Exception):
@@ -632,7 +634,7 @@ def get_instance_id_from_ip(ip):
     return instance_id
 
 
-def autoscale_local_cluster():
+def autoscale_local_cluster(dry_run=False):
 
     system_config = load_system_paasta_config()
     autoscaling_resources = system_config.get_cluster_autoscaling_resources()
@@ -648,6 +650,6 @@ def autoscale_local_cluster():
                 slaves_to_kill = pick_slaves_to_kill(mesos_state, number_to_kill, pool=resource['pool'])
             else:
                 slaves_to_kill = []
-            resource_scaler(resource, delta, target, slaves_to_kill)
+            resource_scaler(resource, delta, target, slaves_to_kill, dry_run)
         except ClusterAutoscalingError as e:
             log.error('%s: %s' % (identifier, e))
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -13,7 +13,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-import re
 import time
 from collections import defaultdict
 from contextlib import contextmanager
@@ -23,6 +22,7 @@ from math import floor
 
 import boto3
 import requests
+from botocore.exceptions import ClientError
 from kazoo.client import KazooClient
 from kazoo.exceptions import NoNodeError
 
@@ -40,6 +40,7 @@ from paasta_tools.marathon_tools import set_instances_for_marathon_service
 from paasta_tools.mesos_tools import get_mesos_state_from_leader
 from paasta_tools.mesos_tools import get_mesos_task_count_by_slave
 from paasta_tools.mesos_tools import get_running_tasks_from_active_frameworks
+from paasta_tools.mesos_tools import slave_pid_to_ip
 from paasta_tools.paasta_maintenance import drain
 from paasta_tools.paasta_metastatus import get_resource_utilization_by_grouping
 from paasta_tools.utils import _log
@@ -437,18 +438,43 @@ def create_autoscaling_lock():
         zk.stop()
 
 
-@register_autoscaling_component('aws_spot_fleet_request', CLUSTER_METRICS_PROVIDER_KEY)
-def spotfleet_metrics_provider(spotfleet_request_id, mesos_state, resource):
+def get_spot_fleet_instances(spotfleet_request_id):
     ec2_client = boto3.client('ec2')
     spot_fleet_instances = ec2_client.describe_spot_fleet_instances(
         SpotFleetRequestId=spotfleet_request_id)['ActiveInstances']
-    desired_instances = len(spot_fleet_instances)
+    return spot_fleet_instances
+
+
+def get_sfr_instance_ips(spotfleet_request_id):
+    ec2_client = boto3.client('ec2')
+    spot_fleet_instances = get_spot_fleet_instances(spotfleet_request_id)
     instance_ips = {
         instance['PrivateIpAddress']
         for reservation in ec2_client.describe_instances(
             InstanceIds=[instance['InstanceId'] for instance in spot_fleet_instances])['Reservations']
         for instance in reservation['Instances']
     }
+    return instance_ips
+
+
+def get_sfr_state(spotfleet_request_id):
+    ec2_client = boto3.client('ec2')
+    try:
+        sfrs = ec2_client.describe_spot_fleet_requests(SpotFleetRequestIds=[spotfleet_request_id])
+    except ClientError as e:
+        if e.response['Error']['Code'] == 'InvalidSpotFleetRequestId.NotFound':
+            log.warn('Cannot find SFR {0}'.format(spotfleet_request_id))
+            return None
+        else:
+            raise
+    ret = sfrs['SpotFleetRequestConfigs'][0]['SpotFleetRequestState']
+    return ret
+
+
+@register_autoscaling_component('aws_spot_fleet_request', CLUSTER_METRICS_PROVIDER_KEY)
+def spotfleet_metrics_provider(spotfleet_request_id, mesos_state, resource):
+    desired_instances = len(get_spot_fleet_instances(spotfleet_request_id))
+    instance_ips = get_sfr_instance_ips(spotfleet_request_id)
     slaves = {
         slave['id']: slave for slave in mesos_state.get('slaves', [])
         if slave_pid_to_ip(slave['pid']) in instance_ips and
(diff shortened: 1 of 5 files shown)


mattmb:
Should be possible to support ASGs too. We'd just have to detach the instance before terminating: http://docs.aws.amazon.com/cli/latest/reference/autoscaling/detach-instances.html

paasta_tools/autoscaling_lib.py
```diff
@@ -508,41 +512,130 @@ def spotfleet_scaler(resource, error):
...
+def slave_pid_to_ip(slave_pid):
+    """Convert slave_pid to IP
+
+    :param: slave pid e.g. slave(1)@10.40.31.172:5051
+    :returns: ip address"""
+    regex = re.compile(r'.+?@([\d\.]+):\d+')
+    return regex.match(slave_pid).group(1)
+
+
+def pick_slaves_to_kill(mesos_state, number, pool='default'):
```
solarkennedy: I think longer term this function has to have the actual list from the the SFR/ASG in order to work.

Otherwise it won't take into account instances that are not even registered in mesos (duds, broken puppet, etc)

paasta_tools/autoscaling_lib.py
```diff
@@ -508,41 +512,130 @@ def spotfleet_scaler(resource, error):
...
     pass
 
 
+def get_instance_id_from_ip(ip):
+    """Get AWS instance ID from private IP
+
+    :param ip: private IP of AWS instance
+    :returns: Instance ID"""
+    ec2_client = boto3.client('ec2')
+    instances = ec2_client.describe_instances(Filters=[{'Name': 'private-ip-address', 'Values': [ip]}])
```
solarkennedy: Hmmmm.

I wonder if this only counts as "1" api call....

I'm pretty sure we are going to do something about that soon :(

paasta_tools/mesos_tools.py
```diff
@@ -615,3 +615,36 @@ def get_mesos_network_for_net(net):
         'host': 'HOST',
     }
     return docker_mesos_net_mapping.get(net, net)
+
+
+def get_mesos_task_count_by_slave(mesos_state, pool='default'):
+    """Get counts of running tasks per mesos slave. Also include separate count of chronos tasks
```
solarkennedy: why do we actually need a separate chronos_count?

are you intending to eventually weight them differently?

paasta_tools/autoscaling_lib.py
mattmb: Hmm, this is a good point we definitely want to clear up anything that's not registered in mesos!

I was trying to keep this function agnostic of the autoscaling provider but maybe I can add something somewhere else to achieve this

paasta_tools/autoscaling_lib.py
mattmb: You're worried we might hit some rate limits? I was also thinking that this was a bit convoluted, is there some way we can store the instance ID in mesos when we bootstrap the slaves?

paasta_tools/mesos_tools.py
mattmb: So my thinking was that batches are more difficult to "drain". In the sense that the best you can do is wait for the current run to finish. So I'm trying to sort by `chronos_count` and then the total number of tasks.

paasta_tools/autoscaling_lib.py
```diff
@@ -508,41 +512,130 @@ def spotfleet_scaler(resource, error):
...
+    :param instances_to_kill: dict of instances to kill {<slave_id>: <instance_id>}"""
+    ec2_client = boto3.client('ec2')
+    try:
+        with Timeout(seconds=CLUSTER_DRAIN_TIMEOUT):
+            while instances_to_kill:
+                mesos_state = get_mesos_state_from_leader()
+                slaves_count = get_mesos_task_count_by_slave(mesos_state, pool=None)
+                for slave_id, instance_id in instances_to_kill.items():
+                    num_tasks = slaves_count[slave_id]['count']
+                    log.debug("Instance: {0}, still running {1} tasks".format(instance_id, num_tasks))
```
Rob-Johnson: drop the 'still' here, and can you add a debug line for when you terminate please

paasta_tools/autoscaling_lib.py
```diff
@@ -37,18 +38,24 @@
 from paasta_tools.marathon_tools import MESOS_TASK_SPACER
 from paasta_tools.marathon_tools import set_instances_for_marathon_service
 from paasta_tools.mesos_tools import get_mesos_state_from_leader
+from paasta_tools.mesos_tools import get_mesos_task_count_by_slave
 from paasta_tools.mesos_tools import get_running_tasks_from_active_frameworks
+from paasta_tools.paasta_maintenance import drain
```
Rob-Johnson: this needs moving out into a more generic 'maintenance' library. ``paasta_maintenance`` is a script, and we shouldn't be importing from it here.

paasta_tools/autoscaling_lib.py
```diff
@@ -508,41 +512,130 @@ def spotfleet_scaler(resource, error):
...
+    sfr_id = resource['id']
+    ec2_client = boto3.client('ec2')
+    if delta == 0:
+        return
+    elif delta > 0:
+        return ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=target_capacity,
+                                                    ExcessCapacityTerminationPolicy='noTermination')
+    elif delta < 0:
+        slaves_by_ip = {slave: slave_pid_to_ip(slave) for slave in slaves_to_kill}
+        instances_to_kill = {slave_id: get_instance_id_from_ip(ip) for slave_id, ip in slaves_by_ip.items()}
```
Rob-Johnson: I don't see you catching the IndexError that this function throws

paasta_tools/autoscaling_lib.py
```diff
@@ -508,41 +512,130 @@ def spotfleet_scaler(resource, error):
...
+        start = time.time()
+        # This ensures we are in maintenance mode longer than the time before
+        # we give up trying to drain and just terminate
+        duration = CLUSTER_DRAIN_TIMEOUT + 300
+        drain(slaves_by_ip.values(), start, duration)
```
Rob-Johnson: please modify this function to return the http response and let the caller decide what to do, rather than relying on it's side-effecty behaviour

paasta_tools/mesos_tools.py
```diff
@@ -615,3 +615,36 @@ def get_mesos_network_for_net(net):
...
+
+    :param mesos_state: mesos state dict
+    :param pool: pool of slaves to return (None means all)
+    :returns: dict of {<slave_id>:{'count': <count>, 'slave': <slave_object>, 'chronos_count':<chronos_count>}}
+    """
+    all_mesos_tasks = get_running_tasks_from_active_frameworks('')  # empty string matches all app ids
+    if pool:
```
Rob-Johnson: so we're defaulting pool to 'default' here - is the only way to trigger this by setting pool=None, in which case we'll be given the default pool anyway? Can you clarify what case this is covering?

paasta_tools/mesos_tools.py
```diff
@@ -615,3 +615,36 @@ def get_mesos_network_for_net(net):
...
+        slaves = {
+            slave['id']: {'count': 0, 'slave': slave, 'chronos_count': 0} for slave in mesos_state.get('slaves', [])
+            if slave['attributes'].get('pool', 'default') == pool
+        }
+    else:
+        slaves = {
+            slave['id']: {'count': 0, 'slave': slave, 'chronos_count': 0} for slave in mesos_state.get('slaves', [])
```
Rob-Johnson: can you create a NamedTuple to represent this dict please, to make the structure of it clearer?

paasta_tools/autoscaling_lib.py
```diff
@@ -508,41 +512,130 @@ def spotfleet_scaler(resource, error):
...
+            while instances_to_kill:
+                mesos_state = get_mesos_state_from_leader()
+                slaves_count = get_mesos_task_count_by_slave(mesos_state, pool=None)
+                for slave_id, instance_id in instances_to_kill.items():
+                    num_tasks = slaves_count[slave_id]['count']
+                    log.debug("Instance: {0}, still running {1} tasks".format(instance_id, num_tasks))
+                    if num_tasks == 0:
+                        ec2_client.termintate_instances(InstanceIds=[instance_id])
+                        instances_to_kill.pop(slave_id)
+                time.sleep(5)
```
Rob-Johnson: please add a log note here to warn people that you're sleeping

paasta_tools/autoscaling_lib.py
```diff
@@ -508,41 +512,130 @@ def spotfleet_scaler(resource, error):
 
     if new_capacity != current_capacity:
         print "Scaling SFR %s from %d to %d!" % (resource['id'], current_capacity, new_capacity)
-        scale_aws_spot_fleet_request(sfr_id=resource['id'], target_capacity=new_capacity, ec2_client=ec2_client)
+        return new_capacity - current_capacity, new_capacity
+    else:
+        return 0
+
+
+def slave_pid_to_ip(slave_pid):
```
Rob-Johnson: can you move this to ``mesos_tools`` please

paasta_tools/mesos_tools.py
```diff
@@ -615,3 +615,36 @@ def get_mesos_network_for_net(net):
...
+        }
+    for task in all_mesos_tasks:
+        if task.slave['id'] not in slaves:
+            log.error("Task associated with %s but slave not in masters list of slaves" % task.slave['id'])
+            continue
+        else:
+            slaves[task.slave['id']]['count'] += 1
+            log.debug("Task framework: {0}".format(task.framework.name))
+            if task.framework.name == 'chronos':
```
Rob-Johnson: Can you verify this is correct? I thought chronos registers as ``chronos-2.4.0`` in mesos.

Rob-Johnson:
Please add some tests.

paasta_tools/mesos_tools.py
mattmb: Ugh, you're right! As discussed, I'm going to tackle this with a puppet change and a PR against Chronos to remove the version from this property rather than make this any more hacky here.

paasta_tools/mesos_tools.py
mattmb: I put this in so that there's a way to get *all* the pools. I'm not using it for the autoscaler but I think I'll leave it as it could be useful elsewhere.

paasta_tools/mesos_tools.py
mattmb: Don't think I can use a tuple because I need to update the counts below?

mattmb:
Ok, pretty happy with this now. Can someone take another look? Probably should write some documentation on the cluster level autoscaling but I'd suggest we get this deployed and start playing with it first.

paasta_tools/autoscaling_lib.py
```diff
@@ -514,41 +547,165 @@ def spotfleet_scaler(resource, error):
...
+    ec2_client = boto3.client('ec2')
+    try:
+        # This loop should always finish because the maintenance window should trigger is_ready_to_kill
+        # being true. Just in case though we set a timeout and terminate anyway
+        with Timeout(seconds=CLUSTER_DRAIN_TIMEOUT + 300):
+            while instances_to_kill:
+                for slave_pid, slave_data in instances_to_kill.items():
+                    instance_id = slave_data['instance_id']
+                    if not instance_id:
+                        # if we failed to find instance id then skip
```
solarkennedy: Just log.debug instead of a comment here

paasta_tools/autoscaling_lib.py
```diff
@@ -514,41 +547,165 @@ def spotfleet_scaler(resource, error):
...
+                        # shouldn't happen but just incase
+                        continue
+                    is_ready_to_kill = True
```
solarkennedy: I would like to see an optimization here eventually where

    is_ready_to_kill or has_no_tasks or is_not_in_mesos

paasta_tools/autoscaling_lib.py
```diff
@@ -514,41 +547,165 @@ def spotfleet_scaler(resource, error):
...
+        start = int(time.time() + CLUSTER_DRAIN_TIMEOUT)
+        # Set the duration to an hour, if we haven't cleaned up and termintated by then
+        # mesos should put the slave back into the pool
+        duration = 600
+        log.debug("Draining {0}".format(instances_to_kill))
+        log.debug("Decreasing spot fleet capacity to: {0}".format(target_capacity))
+        if not dry_run:
+            # sort to make testing easier
+            drain([instance['ip'] for instance in sorted(instances_to_kill.values())], start, duration)
+            ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=target_capacity,
```
solarkennedy: I'm a little concerned about the method here.

What happens if something goes wrong with the `wait_and_terminate` step? I don't think the autoscaler will be able to recover because those instances will be "lost".

I think you have to do it "atomically" where scale + kill at the same time on each instance, otherwise you risk instances falling through the cracks

paasta_tools/mesos_tools.py
Rob-Johnson: right, but what I mean is the default value doesn't make sense as it currently stands. If I wanted to get **all** the pools, then I'd have to do ``get_mesos_task_count_by_slave(mesos_state, pool=None)``, which is odd.


Two suggestions here:

- Set the default parameter to be None, so that you get **all** pools unless you specify otherwise
```
def get_mesos_task_count_by_slave(mesos_state, pool=None)
```


- rename the parameter to ``pool_filter``, and have it as a function that returned ``True`` when you matched a pool that you're interested in, and ``False`` otherwise. The caller could then do something like:
```
get_mesos_task_count_by_slave(mesos_state, lambda pool: pool == 'default')
get_mesos_task_count_by_slave(mesos_state, lambda pool: pool == 'lucy')
get_mesos_task_count_by_slave(mesos_state, lambda pool: True) #match any
```

I kind of like the second one for it's flexibility, but I'll leave it with you to decide.

paasta_tools/autoscaling_lib.py
mattmb: Agreed. So my initial plan is for the call to paasta maintenance to check the slave is running no tasks OR it has reached the maintenance window (which means it failed to drain in a timely fasion).

I'm thinking I might handle the `is_not_in_mesos` earlier on because by this point we're only looking at slaves that we found.

paasta_tools/mesos_tools.py
Rob-Johnson: just make the return value a namedtuple then - it's a real pain to have to figure out what keys might be in the dict returned everyime you want to use a function. a NamedTuple puts a contract around that.

paasta_tools/autoscaling_lib.py
```diff
@@ -514,41 +547,165 @@ def spotfleet_scaler(resource, error):
...
+        if not dry_run:
+            ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=target_capacity,
+                                                 ExcessCapacityTerminationPolicy='noTermination')
+            return
+    elif delta < 0:
+        number_to_kill = delta * -1
+        sfr_ips = get_sfr_instance_ips(sfr_id)
+        log.debug("IPs in SFR: {0}".format(sfr_ips))
+        sfr_sorted_slaves = [slave for slave in sorted_slaves if slave_pid_to_ip(slave['pid']) in sfr_ips]
+        log.debug("SFR slave kill preference: {0}".format(sfr_sorted_slaves))
```
Rob-Johnson: can you just print the slave ids here, rather than the entire object? this log line is going to be very hard to parse otherwise

paasta_tools/autoscaling_lib.py
```diff
@@ -514,41 +547,165 @@ def spotfleet_scaler(resource, error):
...
+        if number_to_kill > len(sfr_sorted_slaves):
+            log.error("Didn't find enough candidates to kill. This shouldn't happen so let's not kill anything!")
+            return
+        else:
```
Rob-Johnson: you don't need the else clause here.

paasta_tools/autoscaling_lib.py
```diff
@@ -514,41 +547,165 @@ def spotfleet_scaler(resource, error):
...
+            slaves_to_kill = sfr_sorted_slaves[0:number_to_kill]
+        log.debug("Set to kill: {0}".format([slave['pid'] for slave in slaves_to_kill]))
+        instances_to_kill = {}
+        for slave in slaves_to_kill:
+            ip = slave_pid_to_ip(slave['pid'])
+            instances_to_kill[slave['pid']] = {'ip': ip}
```
Rob-Johnson: can you reduce this into one line:

```
instances_to_kill[slave['pid']] = {
    'ip': ip,
    'instance_id': get_instance_id_from_ip(ip)
}
```

paasta_tools/autoscaling_lib.py
mattmb: That's a good point. I guess I could reduce target capacity by 1, then wait_and_terminate and repeat until we reach the target.

I suppose the only downside is if a particular instance takes a while to drain then we make the whole scaling event take longer? But I guess we don't want two runs to overlap anyway and we can be aggressive with the timeout on draining if need be.

paasta_tools/mesos_tools.py
mattmb: Fair point, I guess it makes it easier to consume. I'll fixup :-)

paasta_tools/mesos_tools.py
mattmb: Right, I see what you mean now 👍 

In the interests of keeping it simple I'm going to do option 1.

mattmb:
After discussion, I agree with @solarkennedy that we need to be more atomic when we scale down (especially so we handle the case when we fail to clean up an instance). We also need to add some logic to the scale down to take account of instance weightings.

My plan is to ship this and get it working in "dry_run" mode. Then I'll tackle the remaining points in a fresh branch.

@Rob-Johnson @solarkennedy are you happy with the other fixups? If so I'll autosquash and merge.

solarkennedy:
Yep, I think this is one of those things where we pretty much won't get it right the first time, or heck even close, and we'll have to run it in noop mode and iterate.

Ship!

Rob-Johnson:
ship!

PR #601, opened 2016-06-28
Added a --dry-run to the autoscale_cluster script

solarkennedy:
This particular section of the code sure could use a place to put tests, but right now the functions do to much :(

I'm going to wait for @mattmb to either incrementally ship his WIP or pair up with him to help refactor this code so I can more easily add tests.


mattmb:
@solarkennedy I'm at a point where this option would help me out testing things too. If you're happy I'll make a similar addition on my branch.

We should chat about tests too, have you thought about integration tests? Even if we called boto with dry run I would expect them to be tricky...

solarkennedy:
Yea I think full itests for this thing would be pretty freaking complicated.

solarkennedy:
Closing as @mattmb  has this in his branch

PR #602, opened 2016-06-28
Fix autoscaler metrics provider

mattmb:
The provider was getting a namedtuple from metastatus but treating it
like a dictionary and causing a TypeError. I guess metastatus changed at
some point introducing this bug.

I know this should have tests etc. but I'd rather add those on my other branch where I'm refactoring this code quite a bit.

### DIFF
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -467,11 +467,12 @@ def spotfleet_metrics_provider(spotfleet_request_id, mesos_state, pool):
         mesos_state
     )[pool]
 
+    log.debug(pool_utilization_dict)
     free_pool_resources = pool_utilization_dict['free']
     total_pool_resources = pool_utilization_dict['total']
     utilization = 1.0 - min([
-        float(free_pool_resources[resource]) / total_pool_resources[resource]
-        for resource in free_pool_resources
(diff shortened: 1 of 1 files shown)


paasta_tools/autoscaling_lib.py
```diff
@@ -467,11 +467,12 @@ def slave_pid_to_ip(slave_pid):
...
     )[pool]
 
+    log.debug(pool_utilization_dict)
     free_pool_resources = pool_utilization_dict['free']
     total_pool_resources = pool_utilization_dict['total']
     utilization = 1.0 - min([
-        float(free_pool_resources[resource]) / total_pool_resources[resource]
-        for resource in free_pool_resources
+        float(free_pool_resources.__getattribute__(resource)) / total_pool_resources.__getattribute__(resource)
+        for resource in free_pool_resources._fields
```
Rob-Johnson:
```
utilization = 1.0 - min([float(pair[0]/pair[1]) for pair in zip(free_pool_resources, total_pool_resources)])
```

solarkennedy:
Ship, but yea we need tests to make sure this code does what we think it does :(

Rob-Johnson:
really this is a big smell that some of this needs moving into another module

PR #603, opened 2016-06-28
Various logging improvements:

ammaraskar:
1. Warn when trying to grab logs from shared streams like chronos and marathon
2. Fix deploy component not going to devc to get logs
3. Use ujson for much faster processing of logs

Internal tickets: PAASTA-5033, PAASTA-5038

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -15,7 +15,6 @@
 """PaaSTA log reader for humans"""
 import argparse
 import datetime
-import json
 import logging
 import re
 import sys
@@ -28,6 +27,7 @@ from Queue import Empty
 import dateutil
 import isodate
 import pytz
+import ujson as json
 
 try:
     from scribereader import scribereader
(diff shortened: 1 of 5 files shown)


Rob-Johnson:
can you write some tests please?

ammaraskar:
For what part specifically? 

I don't see how testing the printing of the warning would be useful since its literally just one if condition.
The mapping is yelp specific and needs to be handled on a case by case basis.
The json library drop in replacement is covered by the various parse and filter tests that already process json.

paasta_tools/cli/cmds/logs.py
```diff
@@ -801,6 +806,8 @@ def determine_scribereader_envs(self, components, cluster):
             # If a component has a 'source_env', we use that
             # otherwise we lookup what scribe env is associated with a given cluster
             env = LOG_COMPONENTS[component].get('source_env', self.cluster_to_scribe_env(cluster))
+            if 'additional_source_envs' in LOG_COMPONENTS[component]:
```
solarkennedy: I guess here would be the place for some sort of test. `test_determine_scribereader_envs_handles_additional_envs`

requirements.txt
```diff
@@ -43,6 +43,7 @@ simplejson==3.6.5
 six==1.9.0
 thriftpy==0.1.15
 tron==0.6.1.1
+ujson==1.35
```
solarkennedy: Looking in the travis file it doesn't look like this needs any extra library help? (like yaml => libyaml-dev)

Also needs to be in setup.py

ammaraskar: yup, doesn't look like it has any dependencies https://github.com/esnme/ultrajson/blob/master/setup.py

solarkennedy:
fix/rerun tests? and ship

PR #611, opened 2016-07-06
Deep merge config files

mattmb:
This ensures we perform a deep merge of each config file rather
than just a dict update. The motivation is for cluster autocaler
config files which are generated by terraform and currently only
the last one is active.

Here's an example where we want to have one file for each SFR but merge together in `cluster_autoscaling_resources`:
```
{
  "cluster_autoscaling_resources": {
    "aws_spot_fleet_request.${cluster_name}-slaves": {
      "type": "aws_spot_fleet_request",
      "id": "${sfr_id}",
      "pool": "${pool}",
      "min_instances": ${min_instances},
      "max_instances": ${max_instances}
    }
  }
}
```


solarkennedy:
ship

PR #612, opened 2016-07-07
Fix paasta_metastatus: SystemPaastaConfig doesn't raise MarathonNotConfigured or ChronosNotConfigured if marathon_config/chronos_config don't exist

EvanKrall:


EvanKrall:
@matthewbentley can you take a look?

matthewbentley:
lgtm, assuming the tests pass

PR #613, opened 2016-07-07
Change logging for autoscaler

mattmb:
Log more at info so we can disable debug logging which is way too
verbose.


solarkennedy:
ship

PR #614, opened 2016-07-07
Fix bug handling task counts in autoscaler

mattmb:
The task counts are no longer a dictionary. This fixes the resulting TypeError.


solarkennedy:
ship?

PR #615, opened 2016-07-08
Fix delayed and suspended status when no deployments are in progress

keshavdv:
When no deployments are in progress, the previous logic incorrectly believes that the task is running


solarkennedy:
ship

PR #617, opened 2016-07-11
Autoscaler weights

mattmb:
* Adds support for weights to the spot autoscaler
* Makes multiple quick modifications to target capacity possible
* Handles an error when an instance is terminated but still reported by describe instances

### DIFF
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -25,6 +25,7 @@ import requests
 from botocore.exceptions import ClientError
 from kazoo.client import KazooClient
 from kazoo.exceptions import NoNodeError
+from requests.exceptions import HTTPError
 
 from paasta_tools.bounce_lib import LockHeldException
 from paasta_tools.bounce_lib import LockTimeout
@@ -683,11 +684,16 @@ def set_spot_fleet_request_capacity(sfr_id, capacity, dry_run):
                 time.sleep(5)
         except TimeoutError:
             log.error("Spot fleet {0} not in active state so we can't modify it.".format(sfr_id))
-            return
+            return False
     if dry_run:
-        return
-    return ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=capacity,
-                                                ExcessCapacityTerminationPolicy='noTermination')
+        return True
+    try:
+        ret = ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=capacity,
+                                                   ExcessCapacityTerminationPolicy='noTermination')
+    except ClientError as e:
+        log.error("Error modifying spot fleet request: {0}".format(e))
+        return False
+    return ret
 
 
 @register_autoscaling_component('aws_spot_fleet_request', SCALER_KEY)
@@ -743,16 +749,24 @@ def scale_aws_spot_fleet_request(resource, current_capacity, target_capacity, so
             duration = 600
             log.info("Draining {0}".format(slave_to_kill['pid']))
             if not dry_run:
-                drain([slave_to_kill['ip']], start, duration)
+                try:
+                    drain([slave_to_kill['ip']], start, duration)
+                except HTTPError as e:
+                    log.error("Failed to trigger drain on {0}: {1}\n Trying next host".format(slave_to_kill['ip'], e))
+                    continue
             log.info("Decreasing spot fleet capacity from {0} to: {1}".format(current_capacity, new_capacity))
-            set_spot_fleet_request_capacity(sfr_id, new_capacity, dry_run)
+            if not set_spot_fleet_request_capacity(sfr_id, new_capacity, dry_run):
+                log.error("Couldn't update spot fleet, stopping autoscaler")
+                break
             log.info("Waiting for instance to drain before we terminate")
             try:
                 wait_and_terminate(slave_to_kill, dry_run)
             except ClientError as e:
                 log.error("Failure when terminating: {0}: {1}".format(slave['pid'], e))
                 log.error("Setting spot fleet capacity back to {0}".format(current_capacity))
-                set_spot_fleet_request_capacity(sfr_id, current_capacity, dry_run)
+                if not set_spot_fleet_request_capacity(sfr_id, current_capacity, dry_run):
+                    log.error("Couldn't update spot fleet, stopping autoscaler")
+                    break
                 continue
             current_capacity = new_capacity
 
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -61,6 +61,7 @@ DECISION_POLICY_KEY = 'decision_policy'
 SCALER_KEY = 'scaler'
 
 AUTOSCALING_DELAY = 300
+AWS_SPOT_MODIFY_TIMEOUT = 30
 MISSING_SLAVE_PANIC_THRESHOLD = .3
 MAX_CLUSTER_DELTA = .2
 
@@ -662,6 +663,30 @@ def filter_sfr_slaves(sorted_slaves, sfr_id):
     return ret
 
 
+def set_spot_fleet_request_capacity(sfr_id, capacity, dry_run):
+    """ AWS won't modify a request that is already modifying. This
+    function ensures we wait a few seconds in case we've just modified
+    a SFR"""
+    ec2_client = boto3.client('ec2')
+    with Timeout(seconds=AWS_SPOT_MODIFY_TIMEOUT):
+        try:
+            state = None
+            while True:
+                state = get_sfr(sfr_id)['SpotFleetRequestState']
+                if state == 'active':
+                    break
+                log.debug("SFR {0} in state {1}, waiting for state: active".format(sfr_id, state))
+                log.debug("Sleep 5 seconds")
+                time.sleep(5)
+        except TimeoutError:
+            log.error("Spot fleet {0} not in active state so we can't modify it.".format(sfr_id))
+            return
+    if dry_run:
+        return
+    return ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=capacity,
+                                                ExcessCapacityTerminationPolicy='noTermination')
+
+
 @register_autoscaling_component('aws_spot_fleet_request', SCALER_KEY)
 def scale_aws_spot_fleet_request(resource, current_capacity, target_capacity, sorted_slaves, dry_run):
     """Scales a spot fleet request by delta to reach target capacity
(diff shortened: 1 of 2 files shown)


mattmb:
Some things I learned testing this:
* We only handle target capacity. AWS may over provision when we increase the target capacity. If this happens we currently still assume that the capacity is at the target when it may actually be over the target.
* We make a lot of `ec2_client.describe_instances` calls (1 per slave) to translate a slave pid (`slave(1)@10.1.1.1:5051`) into an instance ID. Would be better to find the instance ID some other way (maybe a mesos attribute).

paasta_tools/autoscaling_lib.py
```diff
@@ -569,104 +586,166 @@ def sort_slaves_to_kill(mesos_state, pool='default'):
...
-            ec2_client.terminate_instances(InstanceIds=instances_to_kill, DryRun=dry_run)
+            ec2_client.terminate_instances(InstanceIds=instance_id, DryRun=dry_run)
         except ClientError as e:
             if e.response['Error'].get('Code') == 'DryRunOperation':
                 pass
             else:
                 raise
 
 
+def get_instance_type_weights(sfr_id):
```
solarkennedy: Instead of passing the sfr_id object around, requiring lots of API calls, can you lookup  the SFR once and then pass that SFR object around?

Then this function devolves to just the array getter stuff. Also we won't have to lookup the sfr more than once

mattmb: Yeah, I think this is a good shout. I may still need to call multiple times to get the state when I check in `set_spot_fleet_request_capacity` but I could minimize it elsewhere. I'll take a look 👍

paasta_tools/autoscaling_lib.py
```diff
@@ -569,104 +586,166 @@ def sort_slaves_to_kill(mesos_state, pool='default'):
...
+        return
     elif delta < 0:
-        number_to_kill = delta * -1
-        sfr_ips = get_sfr_instance_ips(sfr_id)
-        log.debug("IPs in SFR: {0}".format(sfr_ips))
-        sfr_sorted_slaves = [slave for slave in sorted_slaves if slave_pid_to_ip(slave['pid']) in sfr_ips]
+        sfr_sorted_slaves = filter_sfr_slaves(sorted_slaves, sfr_id)
         log.info("SFR slave kill preference: {0}".format([slave['pid'] for slave in sfr_sorted_slaves]))
-        if number_to_kill > len(sfr_sorted_slaves):
+        kill_capacity = sum([slave['instance_weight'] for slave in sfr_sorted_slaves])
```
solarkennedy: Would this be better called "killable_capacity" ?

solarkennedy:
I think this looks good, `target_capacity` maps better against what AWS calls it, which hopefully will reduce other kinds of mismatch bugs. In ASGS it just will so happen that the units of `target_capacity` are actually instances?


The ec2 api has a bulk endpoint for describing ec2 instances, and boto3 has this generator thing?
http://boto3.readthedocs.io/en/latest/guide/collections.html#chainability
which is an optimization you could use to prefetch all the instance metadata right after you lookup the SFR and then pass around the "sfr_instances" object or something?

paasta_tools/autoscaling_lib.py
mattmb: Naming is hard 😉 

Yeah, that is what it represents

mattmb:
@solarkennedy yep, that was my thinking, for ASGs it would be a 1:1 mapping so you could keep the name as capacity and just be clear in the docs.

Re the calls to the ec2 api, I have used the bulk call at one stage but the code to translate slave pid to instance ID does make 1 call per host. It's a bit hacky anyway. So I think I'd rather try and get rid of it and somehow lookup the instance ID another way. I'd say that's one for another PR.

paasta_tools/autoscaling_lib.py
```diff
@@ -445,19 +446,21 @@ def get_spot_fleet_instances(spotfleet_request_id):
...
+    spot_fleet_instances = sfr['ActiveInstances']
+    instance_descriptions = describe_instances([instance['InstanceId'] for instance in spot_fleet_instances])
+    instance_ips = []
+    for instance in instance_descriptions:
+        try:
+            instance_ips.append(instance['PrivateIpAddress'])
+        except KeyError:
+            log.warning("Instance {0} does not have an IP. This normally means it has been"
+                        " terminated".format(instance['InstanceId']))
+            continue
```
Rob-Johnson: ditch the ``continue``

paasta_tools/autoscaling_lib.py
```diff
@@ -467,25 +470,42 @@ def get_sfr_state(spotfleet_request_id):
             return None
         else:
             raise
-    ret = sfrs['SpotFleetRequestConfigs'][0]['SpotFleetRequestState']
+    ret = sfrs['SpotFleetRequestConfigs'][0]
+    return ret
+
+
+def describe_instances(instance_ids):
```
Rob-Johnson: can you add a docstring here to explain what this does?

paasta_tools/autoscaling_lib.py
```diff
@@ -569,104 +589,165 @@ def sort_slaves_to_kill(mesos_state, pool='default'):
...
+                    break
+                log.debug("SFR {0} in state {1}, waiting for state: active".format(sfr_id, state))
+                log.debug("Sleep 5 seconds")
+                time.sleep(5)
+        except TimeoutError:
+            log.error("Spot fleet {0} not in active state so we can't modify it.".format(sfr_id))
+            return
+    if dry_run:
+        return
+    ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=capacity,
```
Rob-Johnson: can you return the response object that this returns, so the caller can do something if it fails?

paasta_tools/autoscaling_lib.py
```diff
@@ -569,104 +589,165 @@ def sort_slaves_to_kill(mesos_state, pool='default'):
...
+    :param target_capacity: target SFR capacity
     :param sorted_slaves: list of slaves by order to kill
     :param dry_run: Don't drain or make changes to spot fleet if True"""
+    target_capacity = int(target_capacity)
+    current_capacity = int(current_capacity)
+    delta = target_capacity - current_capacity
     sfr_id = resource['id']
-    ec2_client = boto3.client('ec2')
     if delta == 0:
         return
```
Rob-Johnson: want to add a ``log.info("no change required for sfr")`` or something?

paasta_tools/autoscaling_lib.py
```diff
@@ -569,104 +589,165 @@ def sort_slaves_to_kill(mesos_state, pool='default'):
...
-            drain([instance['ip'] for instance in sorted(instances_to_kill.values())], start, duration)
-            ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=target_capacity,
-                                                 ExcessCapacityTerminationPolicy='noTermination')
-        log.info("Waiting for instances to drain before we terminate")
-        wait_and_terminate(instances_to_kill, dry_run)
+        sfr_sorted_slaves.reverse()
+        while True:
+            if len(sfr_sorted_slaves) == 0:
+                break
+            slave_to_kill = sfr_sorted_slaves.pop()
```
Rob-Johnson: doing this means that if *any* of the logic below here fails, then we won't get to retry this host (because the ``sfr_sorted_slaves`` list is modified by pop().

Instead, can you do:
```
slave_to_kill = sfr_sorted_slaves[0]
```

then if we're successful, you can do:

```
sfr_sorted_slaves = sfr_sorted_slavesx[1:]
```

but, if you fail at any point, add the slave to the back of the queue so you try again?

paasta_tools/autoscaling_lib.py
```diff
@@ -569,104 +589,165 @@ def sort_slaves_to_kill(mesos_state, pool='default'):
...
+                drain([slave_to_kill['ip']], start, duration)
+            log.info("Decreasing spot fleet capacity from {0} to: {1}".format(current_capacity, new_capacity))
+            set_spot_fleet_request_capacity(sfr_id, new_capacity, dry_run)
+            log.info("Waiting for instance to drain before we terminate")
+            try:
+                wait_and_terminate(slave_to_kill, dry_run)
+            except ClientError as e:
+                log.error("Failure when terminating: {0}: {1}".format(slave['pid'], e))
+                log.error("Setting spot fleet capacity back to {0}".format(current_capacity))
+                set_spot_fleet_request_capacity(sfr_id, current_capacity, dry_run)
```
Rob-Johnson: if you reset it to the current capacity, then we'll loop through every remaining slave and fail with the
```
Terminating instance {0} with weight {1} would take us below our target of {2}....
```

message, right? Why not just break out if you're done here?

But, at the same time, why're we resetting here? Don't we just want to try again with the next slave?

mattmb: So I'm setting the capacity back because if we haven't killed that slave then we haven't actually changed the capacity. Also the next slave might have a different weight and would change the capacity by a different amount. Note that `current_capacity` gets updated at the end of the loop.

What I think this does: pick a slave to kill, work out its weight, reduce the SFR by that amount, try to terminate, if terminate fails: reset capacity, try again with next slave.

paasta_tools/autoscaling_lib.py
mattmb: Hmm, I see your point. I guess I thought that not retrying a failed slave was probably ok (it would still be tried again next time the scaler runs). If I make this change I'd have to add try/excepts around the drain calls and the set_spot_fleet calls?

mattmb:
@Rob-Johnson thanks for your comments 😄

I've added some fixups and some more comments, let me know what you think.

paasta_tools/autoscaling_lib.py
```diff
@@ -569,104 +595,166 @@ def sort_slaves_to_kill(mesos_state, pool='default'):
...
     if delta == 0:
+        log.info("Already at target capacity: {0}".format(target_capacity))
         return
     elif delta > 0:
         log.info("Increasing spot fleet capacity to: {0}".format(target_capacity))
-        if not dry_run:
-            ec2_client.modify_spot_fleet_request(SpotFleetRequestId=sfr_id, TargetCapacity=target_capacity,
-                                                 ExcessCapacityTerminationPolicy='noTermination')
-            return
+        set_spot_fleet_request_capacity(sfr_id, target_capacity, dry_run)
```
Rob-Johnson: you might want to check the return value here?

paasta_tools/autoscaling_lib.py
Rob-Johnson: that's not so bad is it?

paasta_tools/autoscaling_lib.py
Rob-Johnson: right, I see now. And do be sure, if we reduce the target capacity, the spotfleet won't automatically respond to that and shutdown an instance too, right? There also isn't any error catching around the ``set_spot_fleet_request_capacity`` calls - you should get the responses to those to make sure they were successful and retry if they're not

mattmb: Yep, exactly. We set the termination policy as `NoTermination`, so reducing the capacity doesn't trigger anything on the AWS side.

I'll try and add some better error handling in this loop, I think you're right that it could be a bit more robust at checking responses.

paasta_tools/autoscaling_lib.py
mattmb: Agreed 👍 

Ive added some more checks but kept the `pop()` method. I think that it makes more sense because if something goes wrong we shouldn't retry the slave or add it back to the list (we could end up in an infinite loop etc.) And the whole thing gets retried in the next run anyway.

mattmb:
@Rob-Johnson @solarkennedy happy if I ship this tomorrow?

Rob-Johnson:
shipit

PR #620, opened 2016-07-14
Reduce EC2 API calls from autoscaler

mattmb:
This consolidates the call to find slaves by IP into one EC2 describe
call. This prevents us hammering the EC2 api.

### DIFF
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -474,17 +474,21 @@ def get_sfr(spotfleet_request_id):
     return ret
 
 
-def describe_instances(instance_ids):
+def describe_instances(instance_ids, instance_filters=None):
     """This wraps ec2.describe_instances and catches instance not
     found errors. It returns a list of instance description
     dictionaries. It assumes one instance per reservation (which
-    seems to be the case for SFRs)
+    seems to be the case for SFRs) Optionally, a filter can be
+    passed through to the ec2 call
 
-    :param instance_ids: a list of instance ids
+    :param instance_ids: a list of instance ids, [] means all
+    :param instance_filters: a list of ec2 filters
     :returns: a list of instance description dictionaries"""
+    if not instance_filters:
+        instance_filters = []
     ec2_client = boto3.client('ec2')
     try:
-        instances = ec2_client.describe_instances(InstanceIds=instance_ids)
+        instances = ec2_client.describe_instances(InstanceIds=instance_ids, Filters=instance_filters)
     except ClientError as e:
         if e.response['Error']['Code'] == 'InvalidInstanceID.NotFound':
             log.warn('Cannot find one or more instance from IDs {0}'.format(instance_ids))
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling_lib.py
```diff
@@ -474,17 +474,19 @@ def get_sfr(spotfleet_request_id):
     return ret
 
 
-def describe_instances(instance_ids):
+def describe_instances(instance_ids, instance_filters=[]):
```
Rob-Johnson: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

mattmb: Ugh, python :-S

Good spot, will fixup, thanks!

solarkennedy: Ug, my least favorite gotcha :(

solarkennedy:
lgtm

paasta_tools/autoscaling_lib.py
```diff
@@ -775,19 +778,18 @@ class ClusterAutoscalingError(Exception):
     pass
 
 
-def get_instance_id_from_ip(ip):
+def get_instance_id_from_ip(ip, instance_descriptions):
```
Rob-Johnson: This whole function should be far less opinionated about what it's doing.

See the following:

```
def find_instance_for_ip(instance_descriptions, ip):
     return [instance for instance in instance_descriptions if instance['PrivateIpAddress']
```
Then the caller can decide what to do according to the return value:

```
instances = find_instance_for_ip(descriptions, ip)
if not instances:
   log.warning("Couldn't find instance ID for ip {0}".format(ip))
   //I can do something here now that there are none
if len(instances) > 1:
   log.warning("wtf how did this happen")
else:
    //only care about the id for now, but the whole instance object is here should
    //I need something else from it later
    instance_id = instances[0]['InstanceId']
```

See how the function now cares far less about what the result of the match 'means' (no matches, >1 match, etc), and what information in the return value is useful (only the ``instanceid`` before) and leaves it to the caller to care instead? It makes it more reusable.

mattmb: So definitely agree this should just be `get_instance_from_ip` and return the whole instance.

Not so sure on moving the check for the number of instances returned. Don't think this function should ever return more than one instance. It might return `None` if it can't find the IP you specify. I guess it would be sensible to check that we never get more than one instance (and maybe raise an error) but I'd still say that check should live in this function.

Rob-Johnson: >  Don't think this function should ever return more than one instance.

Sure, given the context, it makes sense that the length of the array returned is only 0  or 1. But,  if you do that, you're enforcing that opinion and the handling of the error on *everything* that calls the function. It might be in one place this only requires a ``log.warning``, but in another it requires it requires a ``ValueException`` or something else - you lose that flexibility if you put it in the function. 

>  It might return None if it can't find the IP you specify.

Nooo. This is just a wrapper around a ``filter``. Filter *always* just returns an empty list if none of the elements of the list passed to it 'match'. Please leave it that way.

If you also want to add a 'uniqueness' check, then do that separately and make it more generic (btw I totally see why you want this, just think the logic for it should be separate from the filtering). Maybe create a wrapper function if you think it'll be reused.

```
./paasta_tools/utils.py
def raise_unless_length(input_list, desired_length):
     if len(input_list) != desired_length:
         raise ValueError("Expected list of length %d, got %d" % (desired_length, len(input_list))
```

then just let this happen:

```
matching_instances = find_instance_for_ip(descriptions, ip)
raise_unless_length(matching_instances, 1)
```

I still think it's nasty because raise_unless_length is so side_effecty, but I guess it reduces the amount of code that's repeated in different places.

mattmb: Yeah, I can see why it would be nice to have the logic separately. Don't think the `raise_unless` function would help that much because the context changes depending on the thing you're checking. I'll try out your first suggestion and post it up 👍

mattmb: How's this: 78e5c82b604ad05d0607d35298bd64d61710ad2d ?

Rob-Johnson:
shipit!

PR #625, opened 2016-07-15
Make paasta stop/start/restart not print the same type of message rep…

ammaraskar:
…eatedly


Rob-Johnson:
Can we instead move the code we had above the loop on line 181 so that it only prints once for all instances? Seems the 'simpler' option - you're adding a lot of complexity here.

ammaraskar:
Sure, this was just meant to be expandable in case we add more commands or different components that can be started/stopped.

Rob-Johnson:
yeah, we can do that when we come to it - take the simple approach for now.

ammaraskar:
Replaced with much simpler implementation

Rob-Johnson:
ship

PR #626, opened 2016-07-18
Use packaged ephemeral-port-reserve

asottile:
This will also fix the socket timeout on kernel>=4.4 and python>=2.7 (fixed in 1.0.1 in ephemeral-port-reserve)


Rob-Johnson:
shipit

solarkennedy:
thanks!

PR #627, opened 2016-07-22
Paasta native mesos scheduler, initial implementation

EvanKrall:
This change allows Paasta to register as a Mesos framework (actually, one Mesos framework per service/instance) rather than using Marathon.

This is definitely a first pass. It's capable of running tasks, and I believe the bounce logic is implemented correctly, and a lot of the edge cases are tested: rollback, scale up, scale down. (I should probably make a test for when a new version is marked for deployment.)

Some of the things that are still missing:

 - It doesn't integrate at all with smartstack.
 - It doesn't do any sort of health checking. (We could implement them ourselves, probably stealing code from local-run, or we could wait until [Mesos implements http health checks](https://issues.apache.org/jira/browse/MESOS-2533))
 - It doesn't do any sort of placement constraints, so it won't try to spread instances across AZs, etc.
 - It's not passing the [docker parameters](https://github.com/Yelp/paasta/blob/5e4e4870952b6065af32c2dc968a4719f0b27a58/paasta_tools/utils.py#L168)
 - There's zero integration with `paasta status` (though this shouldn't be particularly hard, as pretty much all state is stored in Mesos, so we just need to extend the Mesos section a bit.)
 - It doesn't currently avoid using offers from boxes that have maintenance scheduled, though neither does Marathon. I think this would be fairly simple to add, though.
 - It also doesn't currently move tasks away from nodes that have maintenance scheduled, though I think this would also be straightforward to add - modify `need_more_tasks` and `make_healthiness_sorter`, and add an attribute to MesosTaskParameters to keep track of whether the task is on a draining box, and update it when we get inverse offers.

I'm also not a huge fan of the way I've done main() in native_mesos_scheduler.py. I'd perhaps like to integrate with systemd/upstart (or maybe Marathon, like DCOS wants) for supervising all of the individual scheduler processes. Alternatively, it may be fairly straightforward to rearchitect so that we register as a single framework, and keep track of which tasks belong to which service/instance ourselves.

However, all of these caveats are within native_mesos_scheduler.py and its tests. The refactors of marathon_tools (pulling common things into long_running_service_tools) and utils (making it easier to add more types of schedulers) have more impact on other parts of the codebase.


yelp_package/dockerfiles/trusty/Dockerfile
```diff
@@ -14,12 +14,18 @@
...
+
 RUN apt-get update && apt-get -y install dpkg-dev python-tox python-setuptools \
   python-dev debhelper python-yaml libyaml-dev python-pytest pyflakes \
-  git help2man zsh wget
+  git help2man zsh wget mesos=0.28.2-2.0.27.ubuntu1404
 
 RUN cd `mktemp -d` && wget http://mirrors.kernel.org/ubuntu/pool/universe/d/dh-virtualenv/dh-virtualenv_0.11-1_all.deb && dpkg -i dh-virtualenv_0.11-1_all.deb && apt-get -f install
 
+RUN apt-get install -yq zip
+RUN cd /usr/lib/python2.7/site-packages && zip -r /root/mesos.native-0.28.2-py27-none-any.whl mesos/native mesos.native-0.28.2.dist-info
```
solarkennedy: Why do you need this? This should be handled by the normal virtualenv constructs right?

tox.ini
```diff
@@ -76,6 +76,7 @@ deps =
     {[testenv]deps}
     behave==1.2.4
     behave-pytest==0.1.1
+    /root/mesos.native-0.28.2-py27-none-any.whl
```
solarkennedy: Oh right crap. Of course this package isn't on pypi.....

paasta_tools/utils.py
```diff
@@ -926,6 +926,12 @@ def get_marathon_config(self):
         except KeyError:
             return {}
 
+    def get_paasta_native_config(self):
+        try:
+            return self['paasta_native']
```
solarkennedy: Why not self.get('paasta_native', {}) ?

paasta_tools/drain_lib.py
```diff
@@ -100,7 +100,8 @@ class TestDrainMethod(DrainMethod):
     safe_to_kill_task_ids = set()
 
     def drain(self, task):
-        self.downed_task_ids.add(task.id)
+        if task.id not in self.safe_to_kill_task_ids:
```
solarkennedy: Is this a real bug in master?

EvanKrall: It's a bug, though it only manifests in a particular way:

 - a task is put in downed_task_ids by mark_arbitrary_task_as_safe_to_kill,
 - something (e.g. setup_marathon_job or native_mesos_scheduler) calls drain() again
 - then we call mark_arbitrary_task_as_safe_to_kill() again

In the existing itests, mark_arbitrary_task_as_safe_to_kill only gets called once, so this never happens.

tox.ini
EvanKrall: Yeah, this is a weird one. Fortunately, because we're building our .deb inside a docker container, we can do this one crazy trick (python purists hate him!)

I'll add some comments explaining it.

EvanKrall: I also wonder whether this is going to cause us to have a specific version dependency on the system mesos package...

EvanKrall:
Any other feedback?

solarkennedy:
shipit

PR #628, opened 2016-07-25
Check spotfleet status before we gather metrics

mattmb:
Make sure we check the state of the SFR as early as possible. Make sure
we check if it doesn't exist. Currently we don't check if it doesn't
exist and this can cause a boto ClientError later on.


solarkennedy:
ship!

PR #629, opened 2016-07-26
PAASTA-5190: Updating chronos to yelp release in itests

tub:
Adds our own release of chronos into the integration test suite.
Running `make itest` passes with this change in place.


solarkennedy:
ship

Rob-Johnson:
shipit!

PR #631, opened 2016-07-27
Make check_chronos_job only check the most recently run job

ammaraskar:


paasta_tools/check_chronos_jobs.py
```diff
@@ -90,6 +90,23 @@ def sensu_event_for_last_run_state(state):
         return pysensu_yelp.Status.OK
 
 
+def merge_duplicate_jobs(jobs):
```
solarkennedy: Is this really the right name for what this does?

Doesn't this function actually find_most_recent_jobs ?

paasta_tools/chronos_rerun.py
```diff
@@ -93,10 +93,11 @@ def set_tmp_naming_scheme(chronos_job):
     timestamp = timestamp.replace(':', '')
     timestamp = timestamp.replace('.', '')
 
-    chronos_job['name'] = '%s-%s%s%s' % (chronos_tools.TMP_JOB_IDENTIFIER,
-                                         timestamp,
-                                         chronos_tools.SPACER,
-                                         current_name)
+    chronos_job['name'] = '%s%s%s%s%s' % (chronos_tools.TMP_JOB_IDENTIFIER,
```
solarkennedy: I kinda expected the TMP_JOB_IDENTIFIER change to include the hyphen right?

tests/test_check_chronos_jobs.py
```diff
@@ -137,44 +137,68 @@ def test_sensu_event_for_last_run_state_invalid():
...
+    fake_configured_jobs = [('service1', 'chronos_job')]
+    fake_client = Mock(list=Mock(return_value=[('service1', 'chronos_job')]))
+
+    assert check_chronos_jobs.build_service_job_mapping(fake_client, fake_configured_jobs) == {
+        ('service1', 'chronos_job'): [(fake_job, chronos_tools.LastRunState.Fail)]
+    }
+
+    # simulate a re-run where we now pass
+    reran_job = {
+        'name': chronos_tools.TMP_JOB_IDENTIFIER + ' ' + fake_job['name'],
```
solarkennedy: Is it possible to use the exact same function that re_run uses to construct this job name?

ammaraskar:
Fixed your 3 concerns.

paasta_tools/check_chronos_jobs.py
```diff
@@ -90,6 +90,23 @@ def sensu_event_for_last_run_state(state):
         return pysensu_yelp.Status.OK
 
 
+def filter_to_most_recent_jobs(jobs):
+    merged = {}
```
solarkennedy: I don't like the name of this variable. It isn't that you are merging jobs or anything, it is just keeping the most recent "version" of the job output, right?

paasta_tools/check_chronos_jobs.py
```diff
@@ -90,6 +90,23 @@ def sensu_event_for_last_run_state(state):
...
+    for job in jobs:
+        service, instance = chronos_tools.decompose_job_id(job['name'])
+        if instance not in merged:
+            merged[instance] = job
```
solarkennedy: It sounds like this function makes the assumption that it will only receive jobs for a particular service, right?

That sounds a little dangerous and it wasn't totally obvious that it worked like that, can you at least add a docstring or change the function name?

solarkennedy:
I think functionally this will get the job done. @Rob-Johnson can you give a second opinion?

paasta_tools/check_chronos_jobs.py
```diff
@@ -90,6 +90,23 @@ def sensu_event_for_last_run_state(state):
         return pysensu_yelp.Status.OK
 
 
+def filter_most_recent_jobs_for_service(jobs):
+    most_recent = {}
+    for job in jobs:
+        service, instance = chronos_tools.decompose_job_id(job['name'])
+        if instance not in most_recent:
```
Rob-Johnson: The jobs you're passing in here have already been filtered (see the ``chronos_tools.lookup_chronos_jobs`` call on :122 includes ``service`` and ``instance`` filters so you don't need this) and we already have a function for sorting jobs.

Instead, I think the simple version is to remove this new function, and replace the line with ``filtered = chronos_tools.filter_non_temporary_chronos_jobs(matching_jobs)``  with:

```
most_recent_job = [chronos_tools.sort_jobs(matching_jobs)[0]]
```

Note that I'm having to ``make most_recent_job`` a list, because ``last_run_state_for_job`` and ``sensu_message_status_for_jobs`` expect lists rather than individual jobs. This is because we wait until creating the sensu message to check for the case that there is >1 chronos job for a given ``service.instance``, but now that we're sorting the jobs + getting the most recent, this needs to be refactored to be done elsewhere.

I won't block on that though. Ship this and we can sort that after.

ammaraskar: Got it, thanks for the feedback. I'll ship this and file a ticket for that refactor/cleanup for later.

solarkennedy:
ship!

tox.ini
```diff
@@ -76,7 +76,7 @@ deps =
     behave==1.2.4
     behave-pytest==0.1.1
 commands =
-    python -m behave {posargs}
+    behave {posargs}
```
Rob-Johnson: there was a reason that this was done (though I don't remember what)...howcome you're changing it?

ammaraskar: my bad, I was testing some itest changes earlier, and changed it just to see if it works like this. Will remove before committing.

Rob-Johnson:
shipit!

PR #632, opened 2016-07-27
Allow https healthcheck in ServiceNamespaceConfig, since hacheck now …

matthewbentley:
…supports it

We don't want to add it anywhere else, as marathon doesn't support https checks as far as I can tell


solarkennedy:
ship

PR #634, opened 2016-07-28
Ensure we treat instance weights as floats

mattmb:
@matthewbentley recently updated our instance weights for our SFRs. Turns out they don't have to be integers. And also there's an AWS limit on capacity so it makes sense to use smaller values!

Instance weights for SFRs don't have to be integers. This ensures that
we treat them as floats when deciding what to set as the target capacity
and how many hosts to kill.

### DIFF
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -549,32 +549,33 @@ def get_spot_fleet_delta(resource, error):
     if spot_fleet_request['SpotFleetRequestState'] != 'active':
         raise ClusterAutoscalingError('Can not scale non-active spot fleet requests. This one is "%s"' %
                                       spot_fleet_request['SpotFleetRequestState'])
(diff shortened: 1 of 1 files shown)


matthewbentley:
It seems that you may have gone a bit overboard with the `float(...)`s?

I'm pretty sure that python will return a float for `int + float`, so you should just need to change instances of `int(...)` to `float(...)` and change your format strings from `%d` to `%s`.

mattmb:
I have messed up some of the format strings, should be fixed now. I'm always a bit over cautious with the float int thing. Python doesn't always do what you'd guess. e.g. in python2:
```
>>> 4 / 6
0
```

ammaraskar:
If that's the problem you're trying to avoid you can just do
```python
from __future__ import division
```

https://mail.python.org/pipermail/tutor/2008-March/060886.html

mattmb:
@ammaraskar that's useful! IMO that should probably go at the top of every python module.

As discussed, I agree that I've been a bit paranoid with my floats but I don't think it does any harm to be explicit so I'm going to ship this as is.

PR #635, opened 2016-07-29
Add docker-compose example cluster

mattmb:
This is intended as an example of how PaaSta works and as a development
environment/playground. See example_cluster/README.md.

This shouldn't change the PaaSta codebase itself


setup.py
```diff
@@ -49,6 +49,7 @@
         'kazoo >= 2.0.0',
         'marathon >= 0.8.1',
         'mesos.cli == 0.1.5',
+        'mesos.interface == 0.28.0',
```
mattmb: I lied above. I did change one thing. This was already pinned in requirements.txt so I figure it's safe to add here.

example_cluster/docker-compose.yml
```diff
@@ -0,0 +1,96 @@
...
+  marathon:
+    build: ../yelp_package/dockerfiles/itest/marathon/
+    ports:
+      - '8080:8080'
+    links:
+      - zookeeper
+    environment:
+      CLUSTER: testcluster
+    command: 'marathon --zk zk://zookeeper:2181/marathon --master zk://zookeeper:2181/mesos-testcluster --no-logger --env_vars_prefix MARATHON_ --mesos_authentication_principal marathon --mesos_authentication_secret_file /etc/marathon_framework_secret'
+  trusty:
```
solarkennedy: why do you need this one?

solarkennedy:
Instead of calling all of these folders and tox envs "paastatools" with (with no hyphen) can you call all of them "playground"? (or example_cluster)

example_cluster/README.md
```diff
@@ -0,0 +1,36 @@
+## Example Cluster
```
nhandler: These are nice docs. Why not move them into the docs folder so they end up on readthedocs and are easily referenced?

mattmb: Thanks 😄 

I was thinking that might be nice, or at least a link. Do I need to translate to RST? Maybe add to the installation section or the contribution section. I'd like a symlink in here too so that people can find the docs from here.

EvanKrall:
I think it would be good to try and deduplicate the configuration in `example_cluster/docker-compose.yml` and `paasta_itests/docker-compose.yml` using some of the config extension features of docker-compose: https://docs.docker.com/compose/extends/, perhaps the [extends keyword](https://docs.docker.com/compose/extends/#/extending-services)

example_cluster/docker-compose.yml
mattmb: Hmm, who knows :-S I think you're right that it can go.

mattmb:
@EvanKrall nice 👍  hadn't seen that feature before! I'll give it a go and see if it comes out looking cleaner.

mattmb:
As mentioned in standup, should now be working but I had to migrate us to a newer version of docker compose. See b9b55b8 for detail.

EvanKrall:
Can we avoid checked-in RSA keys? I know they're only usable on the containers in this docker-compose setup, but if someone accidentally exposed their containers' SSH ports to the outside world, this could become a RCE vulnerability.

Perhaps you could run ssh-keygen at startup, and use volume mounts to share public keys between containers?

mattmb:
@EvanKrall So this turns out to be quite hard...

SSH keys need to have the right owner. Volume mounts are owned by the UID on the host. The only way round it is to create a user in the container with the same UID as the user on the host. That's further complicated if you use boot2docker etc.

It might be possible but I vote we leave it with the dummy keys committed here. It's not ideal but it'll work in more places.

EvanKrall:
Volumes don't have to be mapped outside of docker - you can just create a volume with e.g. the `VOLUME` directive in the Dockerfile, then mount them between containers with [`volumes_from`](https://docs.docker.com/compose/compose-file/#/volumes-from) in the docker-compose.yml

mattmb:
@EvanKrall interesting, so I ended up using the shared volumes definition in docker compose v2. Volumes from would have worked too but takes all the volumes from the container you specify.

Anyway ssh keys 🔫

EvanKrall:
ship!

docs/source/contributing.rst
```diff
@@ -19,6 +19,12 @@ Integration Tests
 Python 2.7, tox, and Docker are required to run the integration test suite.
 You can run ``make itest`` to execute them.
 
+Example Cluster
+^^^^^^^^^^^^^^^^^
+There is a docker compose configuration based on our itest containers that you
+can use to run the paasta code against a semi realistic cluster whilst you are
```
hashbrowncipher: I'd recommend 'semi-realistic'

docs/source/contributing.rst
```diff
@@ -19,6 +19,12 @@ Integration Tests
...
+developping. More instructions `here <./installation/example_cluster.html>`_
```
hashbrowncipher: developing?

docs/source/installation/example_cluster.rst
```diff
@@ -0,0 +1,67 @@
+Example Cluster
+---------------
+
+This folder contains a docker-compose setup for a semi-complete PaaSTA
+cluster. It is intended for showing how PaaSTA works and as a
+development environment for running the PaaSTA tools. It is **not** a
+good example of how to setup a PaaSTA cluster in production. At Yelp we
```
hashbrowncipher: 'setup' is a noun. 'set up' is a verb.

docs/source/installation/example_cluster.rst
```diff
@@ -0,0 +1,67 @@
...
+``docker-compose scale mesosslave=4``
+
+Some but not all the command line tools should work. Try:
+
+::
+
+    paasta metastatus
+    paasta status -s hello-world
+
+We don't have scribe so check ``/var/logs/paasta_logs`` and syslog on
```
hashbrowncipher: I think the reader would appreciate a comma before 'so'

docs/source/installation/example_cluster.rst
```diff
@@ -0,0 +1,67 @@
...
+the mesosmaster for the output from cron. Also note that all the slaves
+share the hosts docker daemon.
```
hashbrowncipher: host's

BenjaminDavison:
I tried the quick start on a mac (using newest version of Docker) and when running this command.

git clone root@git:dockercloud-hello-world

I was asked for a password, so I'm guessing the volumes for keys don't get mounted properly on Docker Mac, I will try on my Linux box on Monday.

mattmb:
@BenjaminDavison I was hoping that would "just work" but they have changed quite a bit with the volumes stuff in the new docker for mac. It should work fine with boot2docker or on linux.

mattmb:
Thanks for keeping me honest with the grammar @hashbrowncipher :-)

Will ship this now, when travis passes...

PR #636, opened 2016-07-29
paasta itest should stream output to stdout for easier debugging.

EvanKrall:


mattmb:
Would be nice for all the commands that jenkins runs (push-to-registry, mark-for-deployment)? When I was messing about with #635 all the output was going into the log file.

Other than that shipit!

PR #641, opened 2016-08-03
Make paasta status print lines proprotional to how verbose they requested

solarkennedy:
There are times I just wanted more than 10 lines, but didn't want to run paasta logs.

This makes it so -vvvvvv will work.

Also lots of kwargs to make sure I got all the args right in the right order.

### DIFF
--- a/paasta_tools/chronos_serviceinit.py
+++ b/paasta_tools/chronos_serviceinit.py
@@ -281,8 +281,12 @@ def format_chronos_job_status(client, job, running_tasks, verbose=0):
     command = _format_command(job)
     mesos_status = _format_mesos_status(job, running_tasks)
     if verbose > 0:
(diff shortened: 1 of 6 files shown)


EvanKrall:
so if I want 1000 lines I need to do `paasta status -vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv`?

I'd prefer either an exponential number of lines (so -vv is 10, -vvv is 100, -vvvv is 1000, etc.) or 10, 100, all.

Or, just a --lines argument.

solarkennedy:
Made it exponential.

EvanKrall:
lgtm

PR #642, opened 2016-08-04
Backwards compatible way to support multiple nerve namespaces

jolynch:
@EvanKrall I think that this hack is sufficient for PaaSTA to _support_ multiple namespaces being advertised from a single instance. I believe there is more cleanup to do in the drain and bounce code to fully get rid of ``nerve_ns``, but I'd like to take this iteratively since folks are blocking on  this.

What does this do:
* Adds a ``registration_namespaces`` option, maintaining backwards compatibility with ``nerve_ns``; also introduces the concept of registrations instead of tightly coupling to implementation details
* Refactors MarathonServiceConfig to provide ``get_registration_namespaces`` and powers ``get_nerve_ns`` off of that.

What is still to do in follow ups:
* Completely remove ``nerve_ns``
* Refactor the drain_lib, bounce_lib, and monitoring code to check more than just the primary namespace. This is probably the really hard part.
* DRY the registration logic between getting registrations for nerve and such, and things that use MarathonServiceConfig objects.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -531,7 +531,7 @@ class MarathonServiceConfig(InstanceConfig):
         registration_namespaces = self.config_dict.get('registration_namespaces', [])
         # Backwards compatbility with nerve_ns
         # FIXME(jlynch|2016-08-02, PAASTA-4964): DEPRECATE nerve_ns and remove it
-        if 'nerve_ns' in self.config_dict:
+        if not registration_namespaces and 'nerve_ns' in self.config_dict:
             registration_namespaces.append(self.config_dict.get('nerve_ns'))
 
         return registration_namespaces or [self.instance]
@@ -787,19 +787,12 @@ def read_all_namespaces_for_service_instance(name, instance, cluster=None, soa_d
     """
     if not cluster:
         cluster = load_system_paasta_config().get_cluster()
-    srv_info = service_configuration_lib.read_extra_service_information(
-        name,
-        "marathon-%s" % cluster,
-        soa_dir
-    )[instance]
 
-    registration_namespaces = srv_info.get('registration_namespaces', [])
-    # Backwards compatbility with nerve_ns
-    # FIXME(jlynch|2016-08-02, PAASTA-4964): DEPRECATE nerve_ns and remove it
-    if 'nerve_ns' in srv_info:
-        registration_namespaces.append(srv_info['nerve_ns'])
+    marathon_service_config = load_marathon_service_config(
+        name, instance, cluster, load_deployments=False, soa_dir=soa_dir
+    )
 
-    return registration_namespaces or [instance]
+    return marathon_service_config.get_registration_namespaces()
 
 
 def read_namespace_for_service_instance(name, instance, cluster=None, soa_dir=DEFAULT_SOA_DIR):
@@ -815,7 +808,8 @@ def read_namespace_for_service_instance(name, instance, cluster=None, soa_dir=DE
 
 
 def get_proxy_port_for_instance(name, instance, cluster=None, soa_dir=DEFAULT_SOA_DIR):
-    """Get the proxy_port defined in the namespace configuration for a service instance.
+    """Get the proxy_port defined in the first namespace configuration for a
+    service instance.
 
     This means that the namespace first has to be loaded from the service instance's
     configuration, and then the proxy_port has to loaded from the smartstack configuration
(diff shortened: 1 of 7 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -766,17 +776,42 @@ def deformat_job_id(job_id):
     return decompose_job_id(job_id)
 
 
-def read_namespace_for_service_instance(name, instance, cluster=None, soa_dir=DEFAULT_SOA_DIR):
-    """Retreive a service instance's nerve namespace from its configuration file.
-    If one is not defined in the config file, returns instance instead."""
+def read_all_namespaces_for_service_instance(name, instance, cluster=None, soa_dir=DEFAULT_SOA_DIR):
```
EvanKrall: This function is interesting. I wonder why we re-parsed the marathon.yaml instead of loading a MarathonServiceConfig object.

Seems like this could just be:

```
def read_all_namespaces_for_service_instance(name, instance, cluster=None, soa_dir=DEFAULT_SOA_DIR):
    if not cluster:
    marathon_service_config = load_marathon_service_config(service, instance, cluster, load_deployments=False, soa_dir=soa_dir)
    return marathon_service_config.get_registration_namespaces()
```

paasta_tools/marathon_tools.py
```diff
@@ -766,17 +776,42 @@ def deformat_job_id(job_id):
...
+
+    This is the namespace that clients ought talk to, as well as paasta
+    ought monitor. In this context "primary" just means the first one in the
+    list of registration namespaces.
+
+    If one is not defined in the config file, returns instance instead."""
+    return read_all_namespaces_for_service_instance(name, instance, cluster, soa_dir)[0]
 
 
 def get_proxy_port_for_instance(name, instance, cluster=None, soa_dir=DEFAULT_SOA_DIR):
```
EvanKrall: can you update the docstring for this: "Get the proxy_port defined in the _first_ namespace configuration for a service instance."

paasta_tools/marathon_tools.py
```diff
@@ -766,17 +776,42 @@ def deformat_job_id(job_id):
...
         "marathon-%s" % cluster,
         soa_dir
     )[instance]
-    return srv_info['nerve_ns'] if 'nerve_ns' in srv_info else instance
+
+    registration_namespaces = srv_info.get('registration_namespaces', [])
+    # Backwards compatbility with nerve_ns
+    # FIXME(jlynch|2016-08-02, PAASTA-4964): DEPRECATE nerve_ns and remove it
+    if 'nerve_ns' in srv_info:
+        registration_namespaces.append(srv_info['nerve_ns'])
```
EvanKrall: seems like we'd want to prepend, since `read_namespace_for_service_instance` is returning the first element of this list.

paasta_tools/marathon_tools.py
```diff
@@ -523,8 +523,18 @@ def get_healthcheck_timeout_seconds(self):
...
     def get_nerve_namespace(self):
-        return self.config_dict.get('nerve_ns', self.instance)
+        return self.get_registration_namespaces()[0]
+
+    def get_registration_namespaces(self):
+        registration_namespaces = self.config_dict.get('registration_namespaces', [])
+        # Backwards compatbility with nerve_ns
+        # FIXME(jlynch|2016-08-02, PAASTA-4964): DEPRECATE nerve_ns and remove it
+        if 'nerve_ns' in self.config_dict:
+            registration_namespaces.append(self.config_dict.get('nerve_ns'))
```
EvanKrall: Same as [my comment below](https://github.com/Yelp/paasta/pull/642/files#r73556137).

paasta_tools/marathon_tools.py
jolynch: Done

paasta_tools/marathon_tools.py
jolynch: Ack

paasta_tools/marathon_tools.py
jolynch: I'd prefer to just prefer ``registration_namespaces`` and caution heavily against having both at the same time. Docstrings updated to have this warning

solarkennedy:
This looks like a good first step. I think you will also need to edit the json schema?

I say ship!

EvanKrall:
This looks good to me.

jolynch:
@solarkennedy like that?

solarkennedy:
I think this is it. Ship!

PR #644, opened 2016-08-04
Change local-run to healthcheck via GET

ammaraskar:
This is what happens in real deploys, so this is what we should do.


nhandler:
Ship It if the tests pass.

solarkennedy:
ship

PR #645, opened 2016-08-05
Add validation of HTTP drain method and fix bounce errors

keshavdv:
This should allow apps with working HTTP drain endpoints to at least successfully bounce. This doesn't account for all the possible errors that could happen but we aren't sure what the desired behavior should be in those cases so we'll wait until they happen.


EvanKrall:
Ship

Issue #619, opened 2016-07-12
Cluster autoscaler drain/undrain


mattmb:
At the moment the cluster autoscaler just calls `drain` and then terminates the instance.

Todo:
- [x] Make the autoscaler check if the instance is ready to kill (either 0 tasks running or has reached its maintenance window)
- [ ] Call undrain after we terminate to remove the maintenance window
- [x] Think about what to call when something has gone wrong draining
- [x] Pass in hostname to the `drain` call instead of IP

mattmb:
@nhandler capturing what we just discussed re maintenance mode

PR #646, opened 2016-08-05
Add support for maintenance mode to autoscaler

mattmb:
This starts to tick off the items in #619 

* This makes the autoscaler check that a host has drained before it
terminates the instance.
* It also passes the hostname to the drain call, which is what the
maintenance library expects.
* Adjust the times we send to the maintenance library which expects
nanoseconds.

### DIFF
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -43,6 +43,7 @@ from paasta_tools.mesos_tools import get_mesos_task_count_by_slave
 from paasta_tools.mesos_tools import get_running_tasks_from_active_frameworks
 from paasta_tools.mesos_tools import slave_pid_to_ip
 from paasta_tools.paasta_maintenance import drain
+from paasta_tools.paasta_maintenance import is_safe_to_kill
 from paasta_tools.paasta_metastatus import get_resource_utilization_by_grouping
 from paasta_tools.utils import _log
 from paasta_tools.utils import DEFAULT_SOA_DIR
@@ -610,9 +611,8 @@ def wait_and_terminate(slave, dry_run):
                 if not instance_id:
                     log.warning("Didn't find instance ID for slave: {0}. Skipping terminating".format(slave['pid']))
                     continue
-                is_ready_to_kill = True
-                # is_ready_to_kill = paasta_maintenance.something(slave['pid'])
-                if is_ready_to_kill:
+                # Check if no tasks are running or we have reached the maintenance window
+                if is_safe_to_kill(slave['hostname']):
                     log.info("TERMINATING: {0}".format(instance_id))
                     try:
                         ec2_client.terminate_instances(InstanceIds=[instance_id], DryRun=dry_run)
@@ -662,6 +662,7 @@ def filter_sfr_slaves(sorted_slaves, sfr):
                       "This should never happen")
             continue
         sfr_sorted_slave_instances.append({'ip': ip,
+                                           'hostname': slave['hostname'],
                                            'pid': slave['pid'],
                                            'instance_id': instances[0]['InstanceId']})
     ret = []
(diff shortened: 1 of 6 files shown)


mattmb:
@nhandler you're probably best placed to review 😄

paasta_tools/autoscaling_lib.py
```diff
@@ -610,8 +611,8 @@ def wait_and_terminate(slave, dry_run):
                 if not instance_id:
                     log.warning("Didn't find instance ID for slave: {0}. Skipping terminating".format(slave['pid']))
                     continue
-                is_ready_to_kill = True
-                # is_ready_to_kill = paasta_maintenance.something(slave['pid'])
+                # Check if no tasks are running or we have reached the maintenance window
+                is_ready_to_kill = is_safe_to_kill(slave['hostname'])
                 if is_ready_to_kill:
```
nhandler: Why not just do `if is_safe_to_kill(slave['hostname'])` to avoid having the duplicate terminology (`is_ready_to_kill` and `is_safe_to_kill`)?

paasta_tools/autoscaling_lib.py
```diff
@@ -753,16 +755,17 @@ def scale_aws_spot_fleet_request(resource, current_capacity, target_capacity, so
                 break
             # The start time of the maintenance window is the point at which
             # we giveup waiting for the instance to drain and mark it for termination anyway
-            start = int(time.time() + CLUSTER_DRAIN_TIMEOUT)
+            start = int(time.time() + CLUSTER_DRAIN_TIMEOUT) * 1000000000  # nanoseconds
```
nhandler: There is `seconds_to_nanoseconds` in `paasta_tools.paasta_maintenance`, but I guess importing and doing the function call might be overkill for something this simple.

paasta_tools/autoscaling_lib.py
```diff
@@ -753,16 +755,17 @@ def scale_aws_spot_fleet_request(resource, current_capacity, target_capacity, so
...
             # Set the duration to an hour, if we haven't cleaned up and termintated by then
             # mesos should put the slave back into the pool
-            duration = 600
+            duration = 600 * 1000000000  # nanoseconds
             log.info("Draining {0}".format(slave_to_kill['pid']))
             if not dry_run:
                 try:
-                    drain([slave_to_kill['ip']], start, duration)
+                    drain([slave_to_kill['hostname']], start, duration)
```
nhandler: Draining actually requires specifying both a hostname and an IP address. As a convenience, I made paasta_maintenance attempt to query DNS to get the IP address, but I also left a way to bypass this. If you pass a hostname of the form `<hostname>|<ip>`, it will use the values you provide and skip the extra DNS lookup.

Since the autoscaling_lib is aware of both the IP and hostname it is trying to drain, it is probably best if it passes both to paasta_maintenance.

paasta_tools/paasta_maintenance.py
```diff
@@ -422,6 +424,43 @@ def status():
     print "%s" % status.text
 
 
+def is_safe_to_kill(hostname):
+    """Checks if a host has drained or reached its maintenance window
+    :param hostname: hostname to check
+    :returns: True or False
+    """
+    if is_host_drained(hostname) or hostname in get_hosts_in_maintenance_window():
```
nhandler: `return is_host_drained(hostname) or hostname in get_hosts_in_maintenance_window()`

paasta_tools/paasta_maintenance.py
```diff
@@ -422,6 +424,43 @@ def status():
...
+def is_host_drained(hostname):
+    """Checks if a host has drained successfully by confirming it is
+    draining and currently running 0 tasks
+    :param hostname: hostname to check
+    :returns: True or False
+    """
+    mesos_state = get_mesos_state_from_leader()
+    task_counts = get_mesos_task_count_by_slave(mesos_state)
+    slave_task_count = task_counts[hostname].count
+    if is_host_draining(hostname=hostname) and slave_task_count == 0:
```
nhandler: As above, drop the if statement and just `return is_host_draining(hostname=hostname) and slave_task_count == 0`

paasta_tools/paasta_maintenance.py
```diff
@@ -422,6 +424,43 @@ def status():
...
+        return True
+    return False
+
+
+def get_hosts_in_maintenance_window():
```
nhandler: This name is a bit deceptive. A maintenance window has a start time and a duration (which effectively gives it an end time). This function doesn't care if we are past the end time of the maintenance window.

paasta_tools/autoscaling_lib.py
mattmb: This is a good shout. I presume that weirdness comes from the mesos end? No idea why they need both. Fixup on its way.

paasta_tools/paasta_maintenance.py
mattmb: Yeah, I almost changed it to check the end time but this shouldn't actually be relevant to the autoscaler. I'll rename it to be clearer 👍

paasta_tools/autoscaling_lib.py
nhandler: Yeah, the requirement is from mesos. It is probably related to the fact that you can perform maintenance operations on a non-existent host. i.e. You can mark a newly provisioned host as being `down` until puppet finishes getting it setup.

mattmb:
@nhandler I'm happy to ship this if the tests pass now, feel free to merge if you get to it before I do

paasta_tools/autoscaling_lib.py
```diff
@@ -610,9 +611,8 @@ def wait_and_terminate(slave, dry_run):
                 if not instance_id:
                     log.warning("Didn't find instance ID for slave: {0}. Skipping terminating".format(slave['pid']))
                     continue
-                is_ready_to_kill = True
-                # is_ready_to_kill = paasta_maintenance.something(slave['pid'])
-                if is_ready_to_kill:
+                # Check if no tasks are running or we have reached the maintenance window
+                if is_safe_to_kill(slave['hostname']):
                     log.info("TERMINATING: {0}".format(instance_id))
```
nhandler: We probably want to log the hostname/ip as well (since going from an instance_id to a hostname/ip after the instance has been terminated can be challenging).

PR #649, opened 2016-08-08
Introduce a None check for task's started_at field in bounce_lib

ammaraskar:


paasta_tools/bounce_lib.py
```diff
@@ -252,7 +252,7 @@ def get_happy_tasks(app, service, nerve_ns, system_paasta_config, min_task_uptim
         tasks = tasks_in_smartstack
 
     for task in tasks:
-        if min_task_uptime is not None:
+        if min_task_uptime is not None and task.started_at is not None:
```
solarkennedy: I don't think this is quite right. I would actually prefer the started_at check to happen *after* this line. And then 

```
if started_at is None:
  # Hasn't started yet, so skip it as it can't be healthy
  continue
```

PR #650, opened 2016-08-08
make check_docker_image respect DOCKER_HOST, etc.

EvanKrall:


solarkennedy:
ship, looks like the only place where do this not in test?

PR #651, opened 2016-08-08
Only allow whitelisted volumes as a part of extra_volumes

keshavdv:
This prevents services from mounting arbitrary host volumes that may contain sensitive information. The whitelist is configured by placing a file in the system PaaSTA configuration directory (`/etc/paasta` by default) with the following format:

```
{
  'volumes_whitelist': {
    'myservice': ['/some/dir', '/my/otherdir'],
    'otherservice': ['/some/dir', '/my/otherdir'],
  }
}
```

### DIFF
--- a/paasta_itests/steps/setup_steps.py
+++ b/paasta_itests/steps/setup_steps.py
@@ -146,7 +146,8 @@ def working_paasta_cluster(context):
             {'hostPath': u'/nail/etc/beep', 'containerPath': '/nail/etc/beep', 'mode': 'RO'},
             {'hostPath': u'/nail/etc/bop', 'containerPath': '/nail/etc/bop', 'mode': 'RO'},
             {'hostPath': u'/nail/etc/boop', 'containerPath': '/nail/etc/boop', 'mode': 'RO'},
-        ]
+        ],
+        'volumes_whitelist': {}
     }, 'volumes.json')
 
 
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -653,9 +653,15 @@ def configure_and_run_docker_container(
         docker_pull_image(docker_url)
 
     # if only one volume specified, extra_volumes should be converted to a list
-    extra_volumes = instance_config.get_extra_volumes()
-    if type(extra_volumes) == dict:
-        extra_volumes = [extra_volumes]
+    extra_volumes = instance_config.get_extra_volumes(apply_whitelist=False)
+    extra_whitelisted_volumes = [v['hostPath'] for v in instance_config.get_extra_volumes(apply_whitelist=True)]
+
+    disallowed_volumes = list(set([v['hostPath'] for v in extra_volumes]) - set(extra_whitelisted_volumes))
+    if len(disallowed_volumes) > 0:
+        sys.stdout.write(PaastaColors.yellow(
+            "Warning: Your service is attempting to mount the following extra volumes "
+            "that are not whitelisted:\n\n\t%s\n\n"
+            "These mounts will fail when PaaSTA deploys your service.\n\n" % disallowed_volumes))
 
     for volume in system_paasta_config.get_volumes() + extra_volumes:
         volumes.append('%s:%s:%s' % (volume['hostPath'], volume['containerPath'], volume['mode'].lower()))
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -57,6 +57,7 @@ from paasta_tools.utils import SystemPaastaConfig
 from paasta_tools.utils import Timeout
 from paasta_tools.utils import TimeoutError
 from paasta_tools.utils import validate_service_instance
+from paasta_tools.utils import validate_whitelisted_volumes
 
 
 pick_random_port = functools.partial(ephemeral_port_reserve.reserve, '0.0.0.0')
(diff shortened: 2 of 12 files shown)


keshavdv:
The config would now look something like this:
```
{
  'volumes_whitelist': {
    'myservice': [{'hostPath': '/some/dir', 'mode': 'RW'}, {'hostPath': '/some/dir', 'mode': 'RO'}],
    'otherservice': [{'hostPath': '/otherservice/dir', 'mode': 'RW'}],
  }
}
```

For now, I assumed that the path defined in the config would be identical to the volume being mounted (i.e. specifying / as RW in the global config for a service wouldn't allow the service to mount /some_subdirectory)

paasta_tools/cli/cmds/local_run.py
```diff
@@ -652,10 +653,15 @@ def configure_and_run_docker_container(
         docker_hash = docker_url
         docker_pull_image(docker_url)
 
-    # if only one volume specified, extra_volumes should be converted to a list
+    # Warn if the service is attempting to mount disallowed volumes
```
solarkennedy: I don't like "disallowed", just all everything "non-whitelisted"

paasta_tools/setup_chronos_job.py
```diff
@@ -209,6 +210,23 @@ def main():
         log.warn("Skipping %s.%s: Parent job could not be found" % (service, instance))
         sys.exit(0)
 
+    # Check if the service is attempting to access un-whitelisted volumes
```
solarkennedy: Wait, why can this happen in setup_job? This means we have to  load the job config twice?

Don't we have the job config already from "complete_job_config" ?

keshavdv: complete_job_config is the JSON blob that we want to send to Chronos, which would contain the volumes that the service is permitted to mount. We need to load the instance config from the yaml files so we can see what the service actually wishes to mount to compare the two.

EvanKrall: Perhaps we can make create_complete_config take chronos_job_config as an argument (and system_paasta_config while we're at it), to avoid having to reload twice.

paasta_itests/steps/setup_steps.py
```diff
@@ -146,7 +146,8 @@ def working_paasta_cluster(context):
             {'hostPath': u'/nail/etc/beep', 'containerPath': '/nail/etc/beep', 'mode': 'RO'},
             {'hostPath': u'/nail/etc/bop', 'containerPath': '/nail/etc/bop', 'mode': 'RO'},
             {'hostPath': u'/nail/etc/boop', 'containerPath': '/nail/etc/boop', 'mode': 'RO'},
-        ]
+        ],
+        'volumes_whitelist': {}
```
EvanKrall: trailing comma

paasta_tools/cli/cmds/local_run.py
```diff
@@ -652,10 +653,15 @@ def configure_and_run_docker_container(
...
+    # Warn if the service is attempting to mount non-whitelisted volumes
     extra_volumes = instance_config.get_extra_volumes()
-    if type(extra_volumes) == dict:
-        extra_volumes = [extra_volumes]
+
+    non_whitelisted_volumes = validate_whitelisted_volumes(instance_config)
+    if len(non_whitelisted_volumes) > 0:
+        sys.stdout.write(PaastaColors.yellow(
+            "Warning: Your service is attempting to mount the following extra volumes "
+            "that are not whitelisted:\n\n\t%s\n\n"
```
mjksmith: AFAICT this will do `str(non_whitelisted_volumes)` and look like:

```
Warning: Your service is attempting to mount the following extra volumes that are not whitelisted:
    [{'foo': 'bar'}, {'foo': 'baz'}, {'foo': 'foobar'}]
```
Can we print the volumes line-by-line and in a way that's more user-friendly?

paasta_tools/setup_chronos_job.py
```diff
@@ -209,6 +210,23 @@ def main():
...
+    chronos_job_config = chronos_tools.load_chronos_job_config(
+        service=service,
+        instance=instance,
+        cluster=cluster,
+        soa_dir=soa_dir,
+    )
+    non_whitelisted_volumes = validate_whitelisted_volumes(chronos_job_config)
+    if len(non_whitelisted_volumes) > 0:
```
mjksmith: `if non_whitelisted_volumes:`

paasta_tools/setup_marathon_job.py
```diff
@@ -601,6 +601,12 @@ def setup_service(service, instance, client, service_marathon_config, soa_dir):
     full_id = marathon_app_dict['id']
     service_namespace_config = marathon_tools.load_service_namespace_config(service, instance)
 
+    # Check if the service is attempting to access non-whitelisted volumes
+    non_whitelisted_volumes = validate_whitelisted_volumes(service_marathon_config)
+    if len(non_whitelisted_volumes) > 0:
```
mjksmith: `if non_whitelisted_volumes`

paasta_tools/setup_marathon_job.py
```diff
@@ -601,6 +601,12 @@ def setup_service(service, instance, client, service_marathon_config, soa_dir):
...
+        output = "Attempting to mount unpermitted volumes: %s" % non_whitelisted_volumes
+        send_event(service, instance, soa_dir, pysensu_yelp.Status.CRITICAL, output)
```
mjksmith: kwargs

paasta_tools/utils.py
```diff
@@ -344,6 +364,19 @@ def get_net(self):
...
 
 
+def validate_whitelisted_volumes(instance_config):
+    all_extra_volumes = instance_config.get_extra_volumes(apply_whitelist=False)
+    extra_whitelisted_volumes = instance_config.get_extra_volumes(apply_whitelist=True)
+
+    non_whitelisted_volumes = []
+    for volume in all_extra_volumes:
+        v = {'hostPath': volume['hostPath'], 'mode': volume['mode']}
+        if v not in extra_whitelisted_volumes:
```
mjksmith: If only PEP 416 gave us frozendicts we could have used a set here :(

PR #653, opened 2016-08-09
Fix itest rsyslog race

mattmb:
Syslog takes a moment to startup. Occassionally chronos was beating it
and then crashing because it was unable to log.


solarkennedy:
ship

PR #654, opened 2016-08-09
Fix travis docker-engine

mattmb:
Either we need to do this or remove the apt-get steps and go with the latest version that Travis installs.


nhandler:
I think I would be in favor of keeping things pinned. It is one less thing that can change underneath us and cause failures. But we should also reach out to support@travis-ci.com (per https://blog.travis-ci.com/2016-08-09-outage-gce-images) to make sure they are aware of this issue.

Ship It!

PR #657, opened 2016-08-10
Added pyramid_swagger to paasta-api

huadongliu:
Main changes:
1. api input/output validation with swagger.json.
2. install swagger.json to paasta_tools_path/api/api_docs via package_data.

### DIFF
--- a/paasta_itests/steps/paasta_api_steps.py
+++ b/paasta_itests/steps/paasta_api_steps.py
@@ -36,10 +36,10 @@ def service_instance_status(context, app_count, job_id):
     (service, instance, _, __) = decompose_job_id(job_id)
 
     request = testing.DummyRequest()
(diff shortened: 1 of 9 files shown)


solarkennedy:
Fancy! I say ship.

EvanKrall:
Can we put swagger.json somewhere besides /etc? I think /etc/ is more for configuration, whereas swagger.json is data produced by the package. Probably belongs in /usr/share or something.


Plus, I think our puppet purges /etc/paasta, so we'd just end up deleting swagger.json anyway.

huadongliu:
The last updated addressed @EvanKrall comment to install swagger.json to paasta_tools_path/api/api_docs via package_data, similar to cli/schemas/*.json files.

EvanKrall:
Ship!

PR #658, opened 2016-08-10
Move version number to paasta_tool's __init__

ammaraskar:


solarkennedy:
ship

EvanKrall:
Why?

ammaraskar:
So it can be access programatically, we're making our web requests sent along a user-agent header so it would be nicer having something like "Paasta Tools v0.38.x" as the agent.

EvanKrall:
Cool, makes sense.

PR #661, opened 2016-08-10
Set user agent to PaaSTA Tools <version> when sending HTTP requests

ammaraskar:


solarkennedy:
This is cool, can you go a little extra and append the name of the script that is making the call?

paasta_tools/drain_lib.py
```diff
@@ -156,12 +158,13 @@ def post_spool(self, task, status):
...
                 'reason': 'Drained by Paasta',
             },
+            headers={'User-Agent': paasta_tools.USER_AGENT + " drain_lib.py"},
         )
         resp.raise_for_status()
 
     def get_spool(self, task):
         """Query hacheck for the state of a task, and parse the result into a dictionary."""
-        response = requests.get(self.spool_url(task))
+        response = requests.get(self.spool_url(task), headers={'User-Agent': paasta_tools.USER_AGENT + "drain_lib.py"})
```
solarkennedy: need a space before drain_lib.py

solarkennedy:
This isn't bad, and shippable as is, but instead can you make it so we can track down literally which basename script is making http calls from the header? http://stackoverflow.com/questions/4152963/get-the-name-of-current-script-with-python

paasta_tools/__init__.py
```diff
@@ -11,9 +11,21 @@
...
 #
 # Don't bump version manually. See `make release` docs in ./Makefile
+import os
+import sys
+
 __version__ = '0.35.0'
+USER_AGENT = "PaaSTA Tools %s" % __version__
+
+
+def get_user_agent():
```
solarkennedy: Seems way low level, why not in utils where most of our other generic functions go?

paasta_tools/__init__.py
```diff
@@ -11,9 +11,21 @@
...
+    if len(sys.argv) >= 1:
+        return USER_AGENT + os.path.basename(sys.argv[0])
```
solarkennedy: Won't this have no space at all between the version number and the script name?

requirements.txt
```diff
@@ -21,7 +21,7 @@ isodate==0.5.1
 jsonschema==2.5.1
 kazoo==2.2
 lockfile==0.9.1
-marathon==0.8.1
+marathon==0.8.5
```
solarkennedy: this is a tad risky. Can you not bump this quite yet?

solarkennedy:
fix + ship

PR #663, opened 2016-08-10
Stop making local-run guess --pull behavior. Still default to --build…

ammaraskar:
… but ask for --pull explicitly if cook-image target is not found

See internal ticket PAASTA-4994 for what prompted this change.

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -278,9 +278,7 @@ def add_subparser(subparsers):
         '-b', '--build',
         help=(
             "Build the docker image to run from scratch using the local Makefile's "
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -696,24 +692,28 @@ def local_makefile_present():
...
 
 
 def paasta_local_run(args):
-    if args.pull or args.dry_run:
-        build = False
-    elif args.build:
-        build = True
-    else:
-        build = local_makefile_present()
+    if (args.pull is not None) and (args.build is not None):
```
solarkennedy: Isn't this already taken care of ?
https://docs.python.org/2.7/library/argparse.html#mutual-exclusion

solarkennedy:
address + ship

PR #664, opened 2016-08-12
Revert "Upgrade mesos from 0.28.2 to 1.0.0 in itests"

nhandler:
We think the mesos 1.0.0 upgrade has caused our itests to become a bit more flakey. The issue was supposedly fixed in mesos 1.0.1. Reverting the upgrade should hopefully make the itests a bit more reliable, and we will then wait to upgrade until 1.0.1 is out (hopefully sometime next week).


solarkennedy:
Show us what you got travis!

nhandler:
paasta_itests failed on "zookeeper records can be used to scale services down" in the travis run to revert the mesos 1.0.0 upgrade: https://travis-ci.org/Yelp/paasta/jobs/151951830 :(

nhandler:
This change didn't help with our itest stability, so I am closing it.

PR #665, opened 2016-08-13
read extra nerve key-values from the classic service file

ydgholz:
With the intent they'll be read in https://github.com/Yelp/nerve-tools/blob/master/src/nerve_tools/configure_nerve.py#L133 and added as extra config settings.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -16,6 +16,7 @@ that interact with marathon. There's config parsers, url composers,
 and a number of other things used by other components in order to
 make the PaaSTA stack work.
 """
+import json
 import logging
 import os
 import re
@@ -885,6 +886,9 @@ def get_marathon_services_running_here_for_nerve(cluster, soa_dir):
     return nerve_list
 
 
+def _get_classic_service_puppet_file(service):
+    return os.path.join(PUPPET_SERVICE_DIR, service)
+
 def get_classic_services_that_run_here():
     # find all files in the PUPPET_SERVICE_DIR, but discard broken symlinks
     # this allows us to (de)register services on a machine by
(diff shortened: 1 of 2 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -911,6 +915,11 @@ def _namespaced_get_classic_service_information_for_nerve(name, namespace, soa_d
     port_file = os.path.join(soa_dir, name, 'port')
     nerve_dict['port'] = service_configuration_lib.read_port(port_file)
     nerve_name = compose_job_id(name, namespace)
+    try:
+        with open(_get_classic_service_puppet_file(name)) as extras:
+            nerve_dict.update(json.load(extras))
+    except (ValueError,IOError):
+        pass
```
mattmb: Might be worth logging the exception?

ydgholz: Yes, great idea. I'll look for docs related to the puppet_services.d directory too, to mention this new behaviour.

ydgholz:
I don't see a logical place to mention how the classic services file works; it is ok to leave it undocumented?

paasta_tools/marathon_tools.py
```diff
@@ -911,6 +916,14 @@ def _namespaced_get_classic_service_information_for_nerve(name, namespace, soa_d
...
     nerve_name = compose_job_id(name, namespace)
+    try:
+        with open(_get_classic_service_puppet_file(name)) as extras:
+            if os.fstat(extras.fileno()).st_size > 0:
+                nerve_dict.update(json.load(extras))
+    except IOError:
+        log.exception("Could not open {0} for reading".format(extras.name))
+    except ValueError:
+        log.exception("Could not interpret {0} as JSON".format(extras.name))
     return (nerve_name, nerve_dict)
```
solarkennedy: Is new line 933 going to actually allow you to achieve what you want here? As it forces a puppet/classic service to be announced under *every* smartstack namespace defined. (implying that you *must* define completely different services.

Is that ok with you?

ydgholz: Good question! I can use labels here to do what I want (explicitly making a label for tunnel_id & sharing them, instead of using the hostname). We only announce in one namespace at the moment, and I think our end-game is that the tunnel_id should be `service.namespace`.

solarkennedy:
@ydgholz If you still agree, per our conversation just hardcode the same string everywhere and forget all this complexity in paasta/smartstack to solve your problem.

ydgholz:
I agree

PR #666, opened 2016-08-13
increase wait time to alleviate itest run_until_number_tasks failures

huadongliu:
I don't know what has changed in travis and jenkins environments.  Since we have a good itest passing rate on dev boxes and travis/jenkins do not always fail, the failure does not seem to be a requests_cache related bug. Let's try give the test more time (maximum 20x0.5=10 to 30x1=30).


mattmb:
From what I was trying out locally on Friday, I don't *think* it's just timing out. That said, feel free to ship this, it can't hurt!

PR #667, opened 2016-08-15
Move requests_cache uninstall out of itest scenario cleanup

huadongliu:
This change restores commit 8eb42bbe37536bf744a72bec0cf3be4ad8a15111. requests_cache will get uninstalled where it is installed. It will avoid the risk of running setup_marathon_job.main() multiple times in itests.


solarkennedy:
Green checkmark? Ship?

PR #668, opened 2016-08-15
Autoscaler skip non deployed

solarkennedy:
I broke up the large autoscale_sevices function and made it skip services that haven't been deployed yet. (instead of raising an exception early.)

### DIFF
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -378,12 +378,14 @@ def get_configs_of_services_to_scale(cluster, soa_dir=DEFAULT_SOA_DIR):
                 cluster=cluster,
                 soa_dir=soa_dir,
             )
-            if service_config.get_max_instances() and service_config.get_desired_state() == 'start' \
-                    and service_config.get_autoscaling_params()['decision_policy'] != 'bespoke':
-                configs.append(service_config)
         except NoDeploymentsAvailable:
             log.debug("%s is not deployed yet, refusing to do autoscaling calculations for it" %
                       compose_job_id(service, instance))
+            continue
+
+        if service_config.get_max_instances() and service_config.get_desired_state() == 'start' \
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling_lib.py
```diff
@@ -361,27 +363,36 @@ def humanize_error(error):
...
-                    cluster=cluster,
-                    soa_dir=soa_dir,
-                )
-                if service_config.get_max_instances() and service_config.get_desired_state() == 'start' \
-                        and service_config.get_autoscaling_params()['decision_policy'] != 'bespoke':
-                    configs.append(service_config)
+            if service_config.get_max_instances() and service_config.get_desired_state() == 'start' \
+                    and service_config.get_autoscaling_params()['decision_policy'] != 'bespoke':
+                configs.append(service_config)
+        except NoDeploymentsAvailable:
```
Rob-Johnson: can you move this catch to line 381 to be clear it's ``load_marathon_service_config`` that'll throw the exception (and add a ``continue`` after the debug line to skip to the next iteration)

Rob-Johnson:
fix and ship

paasta_tools/autoscaling_lib.py
```diff
@@ -378,12 +378,14 @@ def get_configs_of_services_to_scale(cluster, soa_dir=DEFAULT_SOA_DIR):
                 cluster=cluster,
                 soa_dir=soa_dir,
             )
-            if service_config.get_max_instances() and service_config.get_desired_state() == 'start' \
-                    and service_config.get_autoscaling_params()['decision_policy'] != 'bespoke':
-                configs.append(service_config)
         except NoDeploymentsAvailable:
             log.debug("%s is not deployed yet, refusing to do autoscaling calculations for it" %
                       compose_job_id(service, instance))
+            continue
```
EvanKrall: This would be a good place to use try/except/else.

PR #671, opened 2016-08-16
Use the correct configs for api itests

huadongliu:
Hopefully this will fix recent itest instability.


solarkennedy:
Whoa! How was this working this whole time?

huadongliu:
itests are still failing with the same symptom.

PR #673, opened 2016-08-16
allow to pass --ulimit arguments to docker

gstarnberger:
Adds a new "ulimit" configuration option that allows to pass --ulimit arguments to Docker.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -39,6 +39,16 @@ instance MAY have:
     left on device' errors if they attempt to exceed these limits, and then be
     unable to write any more data to disk.
 
+  * ``ulimit``: Dictionary of ulimit values that are passed to Docker. Defaults
+    to empty dictionary. Each ulimit value is a dictionary with the soft limit
+    specified under the 'soft' key and the optional hard limit specified under
+    the 'hard' key. Ulimit values that are not set are inherited from the
+    default ulimits set on the Docker daemon. Example::
+
+      ulimit:
+        - nofile: {"soft": 1024, "hard": 2048}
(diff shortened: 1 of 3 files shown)


Rob-Johnson:
lg2m - shipit!

paasta_tools/utils.py
```diff
@@ -165,15 +165,28 @@ def get_cpu_quota(self):
         cpu_burst_pct = self.config_dict.get('cpu_burst_pct', DEFAULT_CPU_BURST_PCT)
         return self.get_cpus() * self.get_cpu_period() * (100 + cpu_burst_pct) / 100
 
+    def get_ulimit(self):
```
solarkennedy: Can you also add this in the soaconfigs.rst file for documentation?

docs/source/yelpsoa_configs.rst
```diff
@@ -39,6 +39,10 @@ instance MAY have:
     left on device' errors if they attempt to exceed these limits, and then be
     unable to write any more data to disk.
 
+  * ``ulimit``: Dictionary of ulimit options that are passed to Docker. Defaults
```
solarkennedy: Can you add an example? 

Also reading the upstream docs it looks like it gets taken in as an array?
https://docs.docker.com/engine/reference/commandline/run/#/set-ulimits-in-container-ulimit

gstarnberger: Example added.

The code generates zero or more --ulimit options that are passed to Docker. For example: `--ulimit nofile=1024:2048 --ulimit nice=20`.

I changed the config format in the most recent version of my branch to explicitly specify soft and hard limits (instead of implicitly specifying the hard limit in the limit value).

PR #675, opened 2016-08-17
Add back paasta-api itests

huadongliu:
itests without the paasta-api tests are failing in our jenkins environment. So paasta-api itests should be unrelated to the failures.


solarkennedy:
`DO NOT REVIEW` just makes me want to review it more!!!

PR #676, opened 2016-08-17
Renaming instancetype to step

tub:
yelpsoa-configs have been refactored to use 'step' rather than the slightly confusing term 'instancename' in the deploy pipline.


Rob-Johnson:
lg2m - shipit if tests pass

PR #677, opened 2016-08-17
code cleanup

dvska:


solarkennedy:
Nice. Are the test failures related to these changes though?

paasta_tools/api/views/instance.py
```diff
@@ -101,10 +101,8 @@ def instance_status(request):
     instance = request.swagger_data.get('instance')
     verbose = request.matchdict.get('verbose', False)
 
-    instance_status = {}
-    instance_status['service'] = service
-    instance_status['instance'] = instance
-
+    instance_status = {'service': service,
+                       'instance': instance}
```
EvanKrall: I tend to prefer single-line or line-per-key/value (or line-per-argument for function calls) over visual-indent.

I usually use visual-indent only when a function call is getting too long, and it has many short arguments that would make argument-per-line have too many lines, or when other nearby lines are single-line, like with our use of mocks in tests.

We should probably write this down in the style guide...

paasta_tools/cli/cmds/info.py
```diff
@@ -116,15 +116,11 @@ def get_service_info(service, soa_dir):
...
-    output.append('Monitored By: team %s' % get_team(service=service, overrides={}))
-    output.append('Runbook: %s' % PaastaColors.cyan(get_runbook(service=service, overrides={})))
-    output.append('Git Repo: %s' % git_url)
-    output.append('Jenkins Pipeline: %s' % pipeline_url)
-    output.append('Deployed to the following clusters:')
+    output = ['Service Name: %s' % service, 'Description: %s' % description,
+              'External Link: %s' % PaastaColors.cyan(external_link),
+              'Monitored By: team %s' % get_team(service=service, overrides={}),
+              'Runbook: %s' % PaastaColors.cyan(get_runbook(service=service, overrides={})), 'Git Repo: %s' % git_url,
+              'Jenkins Pipeline: %s' % pipeline_url, 'Deployed to the following clusters:']
```
EvanKrall: Can we do this section as line-per-value instead of visual indent? At the very least, let's put `Runbook:` and `Git Repo:` on separate lines.

paasta_tools/cli/cmds/local_run.py
```diff
@@ -74,20 +74,20 @@ def perform_http_healthcheck(url, timeout):
             try:
                 res = requests.get(url)
             except requests.ConnectionError:
-                return (False, "http request failed: connection failed")
+                return False, "http request failed: connection failed"
```
EvanKrall: I'm not sure how I feel about unwrapping these tuples. I don't think I feel strongly one way or another, but I'm slightly leaning towards leaving the parens on, because explicit is better than implicit.

I didn't see anything in pep8 about tuples/multiple return values and whether parens or no parens are preferred.

paasta_tools/cli/cmds/local_run.py
```diff
@@ -74,20 +74,20 @@ def perform_http_healthcheck(url, timeout):
...
-        return (False, "http request succeeded, code %d" % res.status_code)
+        return False, "http request succeeded, code %d" % res.status_code
     # check if response code is valid per https://mesosphere.github.io/marathon/docs/health-checks.html
-    elif res.status_code >= 200 and res.status_code < 400:
-        return (True, "http request succeeded, code %d" % res.status_code)
+    elif 200 <= res.status_code < 400:
+        return True, "http request succeeded, code %d" % res.status_code
     elif res.status_code >= 400:
-        return (False, "http request failed, code %d" % res.status_code)
+        return False, "http request failed, code %d" % res.status_code
```
EvanKrall: I just noticed a bug with this function: if the status code is <200, then control drops out the of the if/elif/elif and this function returns None.

https://github.com/Yelp/paasta/issues/678

paasta_tools/drain_lib.py
```diff
@@ -180,9 +180,8 @@ def get_spool(self, task):
         ])
         match = re.match(regex, response.text)
         groupdict = match.groupdict()
-        info = {}
-        info['service'] = groupdict['service']
-        info['state'] = groupdict['state']
+        info = {'service': groupdict['service'],
+                'state':   groupdict['state']}
```
EvanKrall: multi-line or single-line, please.

paasta_tools/marathon_tools.py
```diff
@@ -605,22 +605,9 @@ def load_service_namespace_config(service, namespace, soa_dir=DEFAULT_SOA_DIR):
...
-        'timeout_client_ms',
-        'retries',
-        'mode',
-        'discover',
-        'advertise',
-        'extra_healthcheck_headers'
-    ])
+    key_whitelist = {'healthcheck_mode', 'healthcheck_uri', 'healthcheck_port', 'healthcheck_timeout_s',
+                     'updown_timeout_s', 'proxy_port', 'timeout_connect_ms', 'timeout_server_ms', 'timeout_client_ms',
+                     'retries', 'mode', 'discover', 'advertise', 'extra_healthcheck_headers'}
```
EvanKrall: multi-line, please

tests/cli/fsm/autosuggest_test.py
```diff
@@ -61,7 +61,7 @@ def test_suggest_smartstack_proxy_port(self):
             55555,  # bogus out-of-range value
         ]
 
-        def get_smarstack_proxy_port_from_file_side_effect(*args):
+        def get_smarstack_proxy_port_from_file_side_effect():
```
EvanKrall: This is important -- this function is standing in for a function that takes arguments, and you've changed the signature. I think this is what's causing the unit test failures.

However, the way we're doing this seems kinda wonky anyway. I don't think we need this function at all - I believe we could just pass `side_effect=get_smartstack_proxy_port_from_file_returns` on line 66.

EvanKrall:
Thanks for submitting this! I think we can merge this once the tests pass. You can run the tests locally with `tox`, `tox -e general_itests` (which also includes pre-commit hooks, which appear to be failing), and `tox -e paasta_itests`.

solarkennedy:
I'm going to close this for now due to inactivity. Feel free to simply reopen after addressing the issues.

PR #679, opened 2016-08-18
check validity of clusters in start/stop/restart

Rob-Johnson:
rather than failing silently, exit early when the user specifies
an invalid cluster. An invalid cluster is one which the service
specified in the command is *not* deployed to, or, if the user
specifies an instance in the command, an invalid cluster is any
cluster which that service/instance is not deployed to.

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -156,16 +156,33 @@ def paasta_start_or_stop(args, desired_state):
     """Requests a change of state to start or stop given branches of a service."""
     soa_dir = args.soa_dir
     service = figure_out_service_name(args=args, soa_dir=soa_dir)
+    instances = args.instances.split(",") if args.instances else None
 
-    if args.clusters is not None:
+    # assert that each of the clusters that the user specifies are 'valid'
+    # for the instance list provided; that is, assert that at least one of the instances
+    # provided in the -i argument is deployed there.
+    # if there are no instances defined in the args, then assert that the service
+    # is deployed to that cluster.
+    # If args.clusters is falsey, then default to *all* clusters that a service is deployed to,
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -188,6 +205,7 @@ def paasta_start_or_stop(args, desired_state):
         # actually within this cluster.
         if instances is None:
             cluster_instances = list_all_instances_for_service(service, clusters=[cluster], soa_dir=soa_dir)
+            print("no instances specified; restarting all instances for service")
```
solarkennedy: Can you update the help text to make this obvious that it is the behavior? (which will translate to the man page)

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -156,16 +156,33 @@ def paasta_start_or_stop(args, desired_state):
...
+    # if there are no instances defined in the args, then assert that the service
+    # is deployed to that cluster.
+    # If args.clusters is falsey, then default to *all* clusters that a service is deployed to,
+    # and we figure out which ones are needed for each service later.
+    if instances:
+        instance_clusters = [list_clusters(service, soa_dir, instance) for instance in args.instances]
+        valid_clusters = sorted(list(set([cluster for cluster_list in instance_clusters for cluster in cluster_list])))
     else:
-        clusters = list_clusters(service)
+        print list_clusters(service, soa_dir)
```
solarkennedy: Did you mean to print this?

paasta_tools/cli/cmds/start_stop_restart.py
Rob-Johnson: done.

solarkennedy:
ship

PR #681, opened 2016-08-18
Add support for command- steps in check

### LINKED ISSUE #680: `paasta check` is broken when using command-* steps
The `paasta check` command is unaware that command-* steps are not deployments and hence thinks that the job is configured incorrectly.

tub:
Fixes `paasta check` for command- steps


solarkennedy:
ship

PR #682, opened 2016-08-18
remove mesos master state.json monkey patch

huadongliu:
mesos.cli has a default 5 seconds caching. Extra caching is provided by requests_cache EXPLICITLY in both serviceinit and setup_marathon_job. Clients (serviceinit, setup_marathon_job) should have the capability to enable/disable caching based on their requirements. Thus, revert this IMPLICIT caching in mesos_tools.


Rob-Johnson:
shipit

PR #683, opened 2016-08-18
Move paasta autoscaler config

mattmb:
This moves the AWS region to be set per autoscaling resources. This
means we don't have to assume that the paasta master only governs one
AWS region.

This moves the drain timeout and target utilization to be per pool
configurable in the paasta system config. This supports moving slow
draining containers to their own pool.


solarkennedy:
Seems fine. I would be worried about the ordered arguments as it it will be very easy to not place "region" in the third spot, or whatever. Optionally kwargs it up.

Ship.

PR #684, opened 2016-08-19
Revert "Upgrade mesos from 0.28.2 to 1.0.0 in itests"

Rob-Johnson:
This reverts commit a5e97eb7b2a85c0ca4e2c5661cc308314b4e341b.

We're not going to be rolling forward with this until Mesos releases 1.0.1 to address some bugs in the 1.0 release. At the same time, we're trying to understand why we're having flakiness in our integration tests. By reverting to a previous known 'good' version where we haven't seen such flakiness, we take one factor out of what might be causing it, and also bring our integration tests back in line with the version of Mesos we're actively developing against.


solarkennedy:
ship

PR #686, opened 2016-08-19
Make paasta emergency-stop kill an apps tasks harder

solarkennedy:
I guess just be more honest about what emergency stop does?


mattmb:
This doesn't change that much right? setup_marathon will still correct the state next time it runs?

Shipit

Rob-Johnson:
shipit.

solarkennedy:
Correct. That is why the goal here (I think) was to clarify what this thing does?)

solarkennedy:
Eh, I need to fix up the itest on this one.

solarkennedy:
I still can't get this to work :( :( :(

I'm going to try to extract out these changes into smaller PRs and maybe it will be come more clear?

PR #688, opened 2016-08-22
Make log message clearer when counting tasks

mattmb:
This is an expected error so decreasing the level and making the reason
clear.

### DIFF
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -634,7 +634,7 @@ def get_mesos_task_count_by_slave(mesos_state, pool=None):
         }
     for task in all_mesos_tasks:
         if task.slave['id'] not in slaves:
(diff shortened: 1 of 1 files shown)


paasta_tools/mesos_tools.py
```diff
@@ -634,7 +634,7 @@ def get_mesos_task_count_by_slave(mesos_state, pool=None):
         }
     for task in all_mesos_tasks:
         if task.slave['id'] not in slaves:
-            log.error("Task associated with %s but slave not in masters list of slaves" % task.slave['id'])
+            log.debug("Task asscoiated with slave {0} not found in {1} pool".format(task.slave['id'], pool))
```
tub: Asscoiated?

Rob-Johnson:
ship

PR #689, opened 2016-08-22
Add version to metastatus

mattmb:
I've wanted to find out the remote paasta version a few times so I've added it to metastatus here.

I've added a version endpoint in the API too for when we start using the API in our command line clients. @huadongliu can you check I haven't broken anything by moving the services endpoint?

### DIFF
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -20,7 +20,7 @@
     }
   ],
   "paths": {
-    "/{service}": {
+    "/services/{service}": {
       "get": {
         "responses": {
           "200": {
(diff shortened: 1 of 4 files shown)


paasta_tools/api/api_docs/swagger.json
```diff
@@ -50,7 +64,7 @@
         ]
       }
     },
-    "/{service}/{instance}/status": {
+    "/services/{service}/{instance}/status": {
```
solarkennedy: Ah, I figured we would have to namespace this eventually

solarkennedy:
I say ship. @huadongliu  should be primary though.

nhandler:
I would also love to have an easy way to view the version of mesos/marathon/chronos running on individual hosts, but that would probably make things a bit cluttered. I guess we can simply continue to use `mco` for that.

Ship It!

Rob-Johnson:
shipit

huadongliu:
shipit

PR #691, opened 2016-08-23
Added a paasta status option to use the api endpoint

huadongliu:
$ USE_API_ENDPOINT=true .tox/py27/bin/paasta status -s robj-test-service -i main
Pipeline: https://jenkins.yelpcorp.com/view/services-robj-test-service
cluster: mesosstage
instance: main
Git sha:    72ac8530 (desired)
State:      Bouncing (crossover) - Desired state: Started
Marathon:   Healthy - up with (2/2) instances. Status: Running

$ paasta status -s robj-test-service -i main
Pipeline: https://jenkins.yelpcorp.com/view/services-robj-test-service
cluster: mesosstage
instance: main
Git sha:    72ac8530 (desired)
State:      Bouncing (crossover) - Desired state: Started
Marathon:   Healthy - up with (2/2) instances. Status: Running
Mesos:      Healthy - (4/2) tasks in the TASK_RUNNING state.
Smartstack:
    norcal-devc - Critical - in haproxy with (0/2, 0%) total backends UP in this namespace.

### DIFF
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -11,12 +11,8 @@
   "produces": ["application/json"],
   "tags": [
     {
-      "name": "instances",
-      "description": "Instances of a service"
-    },
-    {
-      "name": "status",
-      "description": "Status of a serivce instance"
+      "name": "service",
+      "description": "information about a paasta serivce"
     }
   ],
   "paths": {
@@ -51,8 +47,8 @@
           }
         },
         "summary": "List instances of service_name",
-        "operationId": "listInstances",
-        "tags": ["instances"],
+        "operationId": "list_instances",
+        "tags": ["service"],
         "parameters": [
           {
             "in": "path",
@@ -81,8 +77,8 @@
           }
         },
         "summary": "Get status of service_name.instance_name",
-        "operationId": "statusInstance",
-        "tags": ["status"],
+        "operationId": "status_instance",
+        "tags": ["service"],
         "parameters": [
           {
             "in": "path",
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -137,6 +137,10 @@
           "format": "int32",
           "description": "The number of different running versions of the same service (0 for stopped, 1 for running and 1+ for bouncing)"
         },
+        "app_id": {
+          "type": "string",
+          "description": "ID of the desired version of a service instance"
+        },
         "bounce_method": {
           "type": "string",
           "description": "Method to transit between new and old versions of a service",
@@ -145,7 +149,12 @@
         "deploy_status": {
           "type": "string",
           "description": "Deploy status of a marathon service",
-          "enum": ["Running", "Deploying", "Stopped", "Delayed", "Waiting", "Not Running"]
+          "enum": ["Running", "Deploying", "Stopped", "Delayed", "Waiting", "NotRunning"]
+        },
+        "backoff_seconds": {
+          "type": "integer",
+          "format": "int32",
+          "description": "backoff in seconds before launching the next task"
         },
         "running_instance_count": {
           "type": "integer",
--- a/paasta_tools/api/client.py
+++ b/paasta_tools/api/client.py
@@ -1,0 +1,107 @@
+#!/usr/bin/env python
+# Copyright 2015-2016 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Client interface for the Paasta rest api.
+"""
+import logging
+
+import requests
+
+from paasta_tools.utils import get_user_agent
+from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.utils import PaastaNotConfiguredError
+
+
+log = logging.getLogger(__name__)
+
+
+class PaastaApiError(Exception):
+    pass
+
+
+class PaastaApiClient(object):
+    def __init__(self, cluster=None, system_paasta_config=None, timeout=30):
+        """Create a PaastaApiClient instance.
+        :param str cluster: name of the cluster of an api server
+        :param int timeout: Timeout (in seconds) for a request to an api endpoint
+        :param :class:`requests.Session` session: the session for request and response
+        """
+        self.timeout = timeout
+        self.session = requests.Session()
+        self.session.headers.update({'User-Agent': get_user_agent()})
+
+        try:
+            if not system_paasta_config:
+                system_paasta_config = load_system_paasta_config()
+            api_endpoints = system_paasta_config.get_api_endpoints()
+            if not cluster:
+                cluster = system_paasta_config.get_cluster()
+            self.server = api_endpoints[cluster]
+        except:
+            raise PaastaNotConfiguredError(
+                "Could not find cluster {0} in system paasta configuration directory".format(cluster))
+
+    def _do_request(self, method, path, params=None, data=None):
+        """Hit the api server."""
+        headers = {
+            'Content-Type': 'application/json', 'Accept': 'application/json'}
+        url = ''.join([self.server.rstrip('/'), path])
+        try:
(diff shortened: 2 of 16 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -689,19 +677,28 @@ class InvalidMarathonHealthcheckMode(Exception):
...
         for k, v in vars(cls).iteritems():
             if v == val:
                 return k
 
+    @classmethod
+    def fromstring(cls, str):
+        return getattr(cls, str, None)
+
+
+def get_marathon_app_deploy_status(app_id, client):
```
Rob-Johnson: can you flip the order of these arguments please.

paasta_tools/marathon_serviceinit.py
```diff
@@ -98,19 +95,38 @@ def get_bouncing_status(service, instance, client, job_config):
...
+
 def status_desired_state(service, instance, client, job_config):
     status = get_bouncing_status(service, instance, client, job_config)
-    desired_state = job_config.get_desired_state_human()
+    desired_state = get_desired_state(job_config.get_desired_state(), job_config.get_instances())
     return "State:      %s - Desired state: %s" % (status, desired_state)
 
 
-def status_marathon_job(service, instance, app_id, normal_instance_count, client):
+def status_marathon_job_human(service, instance, deploy_status, app_id=None,
```
Rob-Johnson: please remove the defaults from here. Make it the caller's responsibility to provide sane values for these - this function should really just encode a template to print.

paasta_tools/marathon_serviceinit.py
```diff
@@ -98,19 +95,38 @@ def get_bouncing_status(service, instance, client, job_config):
...
+    elif desired_state == 'start' and instances == 0:
+        return PaastaColors.bold('Stopped')
+    elif desired_state == 'stop':
+        return PaastaColors.red('Stopped')
+    else:
+        return PaastaColors.red('Unknown (desired_state: %s)' % desired_state)
+
+
+def get_desired_state(desired_state, instances):
+    return desired_state_human(desired_state, instances)
```
Rob-Johnson: if this just delegates to the ``desired_state_human`` function, then why have 2 of them at all?

paasta_tools/marathon_serviceinit.py
```diff
@@ -127,6 +143,35 @@ def status_marathon_job(service, instance, app_id, normal_instance_count, client
...
+        deploy_status = status_string
+
+    return deploy_status
+
+
+def status_marathon_job(service, instance, app_id, normal_instance_count, client):
+    status, delay = marathon_tools.get_marathon_app_deploy_status(app_id, client)
+    deploy_status_human = marathon_app_deploy_status_human(status, delay)
+    if status != marathon_tools.MarathonDeployStatus.NotRunning:
+        running_instances = client.get_app(app_id).tasks_running
```
Rob-Johnson: rather than doing this, please always set the ``running_instances`` count, and make it 0 when the status is ``marathon_tools.MarathonDeployStatus.NotRunning``:

```
running_instances = 0 if status == marathon_tools.MarathonDeployStatus.NotRunning else client.get_app(app_id).tasks_running 
```

paasta_tools/marathon_tools.py
Rob-Johnson: it seems odd to me that this returns 2 different types according to the running state of the app (an app, or a tuple of (``MarathonDeployStatus`` and an int). Given the function is called ``get_marathon_app_deploy_status``, I'd expect it to always return a value of ``MarathonDeployStatus``. Please also just make this return the DeployStatus, and make the caller decide if they want the delay, too.

paasta_tools/api/views/instance.py
```diff
@@ -47,17 +45,20 @@ def marathon_job_status(mstatus, client, job_config):
...
         return
 
-    if marathon_tools.is_app_id_running(app_id, client):
-        app = client.get_app(app_id)
-        deploy_status, _ = get_marathon_app_deploy_status(app, app_id, client)
-        mstatus['deploy_status'] = MarathonDeployStatus.tostring(deploy_status)
+    mstatus['app_id'] = app_id
+    mstatus['expected_instance_count'] = job_config.get_instances()
+
+    deploy_status, deploy_delay = marathon_tools.get_marathon_app_deploy_status(app_id, client)
```
Rob-Johnson: following on from the comment I left at the ``get_marathon_app_deploy_status`` function, just make this
```
deploy_status = marathon_tools.get_marathon_app_deploy_status(app_id, client)
```
you already have the code on line 60 to get the delay (so, in the code's current form, this is calculated twice?)

paasta_tools/cli/cmds/status.py
```diff
@@ -74,6 +82,12 @@ def add_subparser(subparsers):
         default=DEFAULT_SOA_DIR,
         help="define a different soa config directory",
     )
+    status_parser.add_argument(
```
Rob-Johnson: I don't feel too strongly about this, but a command line switch seems the wrong interface for this. how about a ``USE_PAASTA_API`` environment variable?

paasta_tools/cli/cmds/status.py
```diff
@@ -130,8 +144,45 @@ def get_actual_deployments(service, soa_dir):
     return actual_deployments
 
 
+def paasta_status_on_api_endpoint(cluster, service, instance, system_paasta_config, verbose):
+    session = requests.Session()
+    session.headers.update({'User-Agent': get_user_agent()})
+    api_client = PaastaApiClient(cluster=cluster,
```
Rob-Johnson: why isn't the session encapsulated in the client?

mattmb:
I'll make sure to take a proper look at this tomorrow.

Just to prompt discussion (not sure it's actually a good idea), have we thought about generating the client from the swagger definition. I googled it and one of the first results was: https://github.com/Yelp/bravado

I tried it like so:
```
>>> client = SwaggerClient.from_url('http://mesosmaster:5054/swagger.json')
>>> client.version.showVersion().result()
u'0.42.0'
>>> client.instances.listInstances(service='hello-world').result()  
{u'instances': [u'main', u'test_chronos']}
>>> 
```

huadongliu:
It is a good idea to use the swagger client instead. Let me make the change.

huadongliu:
Good point. Then we can use the SwaggerClient instead of PaastaApiClient.

mattmb:
Yep, like I said just an idea. Maybe we could use PaastaApiClient as a thin wrapper around SwaggerClient? It'd be nice not to have to call `result()` all the time and maybe there are some other things we can make nicer. What it would give us is one less place to define all the endpoints.

In general this looks great by the way 👍 . I say shipit. Up to you if you prefer to change the client first or merge and start a fresh branch.

huadongliu:
@mattmb Unfortunately bravado has a conflict requirement with tron on twisted, which is failing itests. I need to remove the bravado changes for now.

huadongliu:
A patch to remove tron dependency is going to be made in paasta-tools.

paasta_tools/api/client.py
```diff
@@ -0,0 +1,60 @@
...
+    if cluster not in api_endpoints:
+        log.error('Cluster %s not in paasta-api endpoints config', cluster)
+        return None
+
+    url = str(api_endpoints[cluster])
+    if not url.startswith("http://"):
+        log.error('Unsupported paasta-api url %s', url)
+        return None
+
+    api_server = url[7:]
```
mattmb: https://docs.python.org/2/library/urlparse.html would be nicer and probably make checking http above easier too.

paasta_tools/api/client.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Get swagger spec from file system instead of the api server
```
mattmb: I can't say I'm expert enough to know the right option here. Maybe you've thought it through or someone has an opinion? I guess this couples the client to a specific version of the API rather than whatever is running at the endpoint it hits. However, maybe that's actually safer? Happy either way unless anyone has any more detailed thoughts.

paasta_tools/api/api_docs/swagger.json
```diff
@@ -52,7 +48,7 @@
         },
         "summary": "List instances of service_name",
         "operationId": "listInstances",
```
mattmb: Now that these dictate the python methods can we make them pythonic i.e. `list_instances` rather than camelcase.

paasta_tools/api/client.py
huadongliu: Thanks for the pointer. This is a check before removing the "http://" part from the endpoint config.

paasta_tools/api/client.py
huadongliu: A swagger client can be made with either an url or local spec. Since the spec on the api server is the same as the spec from a local paasta installation, I chose the latter. We won't lose consistency and one hit to the api server is avoided.

mattmb:
Nice work @huadongliu 👏 

shipit!

PR #687, opened 2016-08-22
Move print statements in paasta_maintenance

mattmb:
Moving print statements so they don't show up in logs which use this as a library.

@nhandler I've just moved the print statements that were showing up in the autoscaler log. Would you be happy if I took this further and moved all the printing into `paasta_maintenance()`?

PR #693, opened 2016-08-23
Remove print statements

mattmb:
Move print statements out of function calls to prevent them showing up when these functions are imported elsewhere.

Follow up to #687

### DIFF
--- a/paasta_tools/paasta_maintenance.py
+++ b/paasta_tools/paasta_maintenance.py
@@ -33,7 +33,6 @@ from paasta_tools.mesos_tools import get_mesos_task_count_by_slave
 from paasta_tools.mesos_tools import MESOS_MASTER_PORT
 
 log = logging.getLogger(__name__)
-log.addHandler(logging.NullHandler())
 
 
 def parse_args():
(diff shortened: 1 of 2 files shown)


solarkennedy:
ship (assuming the tests are just flakes)

paasta_tools/paasta_maintenance.py
```diff
@@ -567,33 +576,31 @@ def paasta_maintenance():
...
     elif action == 'is_host_draining':
-        return is_host_draining(hostnames[0])
+        ret = is_host_draining(hostnames[0])
     elif action == 'is_host_past_maintenance_start':
-        return is_host_past_maintenance_start(hostnames[0])
+        ret = is_host_past_maintenance_start(hostnames[0])
     elif action == 'is_host_past_maintenance_end':
-        return is_host_past_maintenance_end(hostnames[0])
+        ret = is_host_past_maintenance_end(hostnames[0])
+    print ret
```
Rob-Johnson: do we really want to print an exit code without any context?

mattmb: fair point. I'll remove the printing for anything that's just an exit code before I ship

PR #699, opened 2016-08-23
 Added readiness_check boilerplate

solarkennedy:
This is the boilerplate for the paasta_readiness_check command, that will eventually be run by puppet as a fact to prevent it from doing potentially disruptive changes.

I mostly want to get this out of the way early so when I have business logic to review it is more clear.

### DIFF
--- a/paasta_tools/readiness_check.py
+++ b/paasta_tools/readiness_check.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python
+# Copyright 2015-2016 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+import logging
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description=(
+            'paasta_readiness_check is a tool to verfiy if changes are safe to '
+            'be made to a host.\n\n'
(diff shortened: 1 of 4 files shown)


paasta_tools/readiness_check.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description=(
+            'paasta_readiness_check is a tool to verfiy if changes are safe to '
+            'be made to a host.\n\n'
+            'It does this by inspecting the running tasks on localhost, and returns '
+            'non-0 if any of thoses tasks are "at-risk"\n\n'
+            'A task is at-risk if it is low in replication as determined by the replication '
```
nhandler: We were also using this terminology to refer to tasks running on a host marked as `draining` in `paasta_maintenance`. This isn't the first overloaded term we've used in PaaSTA. Maybe we should re-evaluate our terminology across the board and attempt to clear some of this up.

paasta_tools/readiness_check.py
```diff
@@ -0,0 +1,45 @@
...
+    return parser.parse_args()
+
+
+def main():
+    args = parse_args()
+    if args.verbose:
+        logging.basicConfig(level=logging.DEBUG)
+    else:
+        logging.basicConfig(level=logging.WARNING)
+    print "Everything is fine (not implemented yet"
```
nhandler: NP: Missing closing `)`

nhandler:
This really doesn't do anything at the present. If you are just laying the groundwork for future changes, maybe also create the unit/integration test file(s). If nothing else, they will attempt to `include` this script and confirm that this action is possible.

I guess go ahead and ship it if you want.

Rob-Johnson:
ship

solarkennedy:
@EvanKrall This is a good opportunity to pick a better name. `paasta_preflight_check`? `

solarkennedy:
Keep in mind this is a "systemy" thing and paasta users would not see it.

EvanKrall:
@solarkennedy how about `paasta maintenance is_safe_to_drain`?

solarkennedy:
I'm going to discard this in favor of @EvanKrall 's suggestion

PR #700, opened 2016-08-23
Fix local-run --dry-run

### LINKED ISSUE #698: paasta local-run --dry-run no longer works
Previously we were doing:
`/usr/bin/paasta local-run -s yelp-main -i main --dry-run`

This worked fine until recently, when we now get:
```
A local Makefile with a 'cook-image' target is required for --build
If you meant to pull the docker image from the registry, explicitly pass --pull
```

#663 changed this behavior.

Previously for a dry run, it set `build = False`:
```python
    if args.pull or args.dry_run:
        build = False
```

...but no longer.

I think this should still be the default. The test in that PR would have failed if it hadn't been modified to add `--pull`.

chriskuehl:
This fixes #698 and reverts the test change. The test should not have been changed as it would have caught the regression.

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -692,9 +692,13 @@ def paasta_local_run(args):
         if args.dry_run:
             sys.stderr.write('Running in dry mode. Assuming --pull\n')
             args.pull = True
-        else:
-            sys.stderr.write("Neither --pull or --build were passed. Assuming --build\n")
+        elif makefile_responds_to('cook-image'):
+            sys.stderr.write("Local Makefile with 'cook-image' target detected. Assuming --build\n")
             args.build = True
+        else:
+            sys.stderr.write("A local Makefile with a 'cook-image' target is required for --build\n")
+            sys.stderr.write("If you meant to pull the docker image from the registry, explicitly pass --pull\n")
+            return 1
 
     service = figure_out_service_name(args, soa_dir=args.yelpsoa_config_root)
     cluster = guess_cluster(service=service, args=args)
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -280,9 +280,9 @@ def add_subparser(subparsers):
             "Build the docker image to run from scratch using the local Makefile's "
             "'cook-image' target. Defaults to try to use the local Makefile if present."
         ),
-        required=False,
-        action='store_true',
-        default=None,
+        action='store_const',
+        const='build',
+        dest='action',
     )
     build_pull_group.add_argument(
         '-p', '--pull',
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -687,22 +687,14 @@ def configure_and_run_docker_container(
...
-        return False
-
-
 def paasta_local_run(args):
-    if (args.pull is None) and (args.build is None):
-        # Default to --build if neither is specified
-        if not local_makefile_present():
-            sys.stderr.write("A local Makefile with a 'cook-image' target is required for --build\n")
-            sys.stderr.write("If you meant to pull the docker image from the registry, explicitly pass --pull\n")
-            return 1
```
chriskuehl: I don't really understand the point of this special case. It seems better to just default to `--build` and let it fail below rather than add more tricky behavior.

Happy to revert this if you disagree.

solarkennedy: Yes please. I would rather the tool report this friendlier error message than letting it fail with a more cryptic error from downstream.

My reasoning is that it is *very* likely that a user meant to use --pull if they are running in a location with no makefile

chriskuehl: We should add a test in that case if this is behavior we care about.

chriskuehl:
@solarkennedy added it back

paasta_tools/cli/cmds/local_run.py
```diff
@@ -687,22 +687,18 @@ def configure_and_run_docker_container(
...
-        return True
-    else:
-        return False
-
-
 def paasta_local_run(args):
-    if (args.pull is None) and (args.build is None):
-        # Default to --build if neither is specified
-        if not local_makefile_present():
+    if args.pull is None and args.build is None:
```
asottile: I think what we actually want:

```python
if args.dry_run:
    args.pull = args.build = False
elif ...:
    # Do the guesswork
```

It strikes me as odd that local-run does gueswork at all.  If I'm in an unrelated repository running `local-run -s yelp-main` and that repository *happens* to have a makefile suddenly the command is responding differently than otherwise.

chriskuehl:
Agreed. I think this would be a lot less confusing if local-run didn't try to guess the intent.

So the behavior we really want is probably that `--dry-run`, `--pull`, and `--build` are all in a mutually-exclusive group since it doesn't make sense to use `--pull` or `--build` along with `--dry-run`.

And then we only try to guess if we're not dry-running.

asottile:
sg2m

chriskuehl:
How does this look? Now `--pull`, `--build`, and `--dry-run` are in the same mutex group and the default is `--build`.

This is really what was happening before, just not using argparse to set the default.

asottile:
I like this

solarkennedy:
Yea, this is better. Thank you @chriskuehl!

PR #703, opened 2016-08-24
pass soa_dir into guess_cluster/get_default_cluster_for_service.

### LINKED ISSUE #702: local-run ignores -y for guessing cluster
In [guess_cluster](https://github.com/Yelp/paasta/blob/bff8fe687c59075236a46b6ea479bf98f4af8f59/paasta_tools/cli/utils.py#L328) and [get_default_cluster_for_service](https://github.com/Yelp/paasta/blob/1f4b49a51fdc0f49dce2b38c970e502231b3df4a/paasta_tools/utils.py#L1192) we aren't passing around soa_dir, which means that the cluster-guessing functionality ignores -y/--yelpsoa-root

EvanKrall:
fixes #702

### DIFF
--- a/paasta_tools/cli/utils.py
+++ b/paasta_tools/cli/utils.py
@@ -31,7 +31,6 @@ from paasta_tools.utils import compose_job_id
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import get_default_cluster_for_service
 from paasta_tools.utils import list_all_instances_for_service
-from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import NoConfigurationForServiceError
 from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import validate_service_instance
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/utils.py
```diff
@@ -303,7 +303,7 @@ def guess_service_name():
     return os.path.basename(os.getcwd())
 
 
-def guess_instance(service, cluster, args):
+def guess_instance(service, cluster, args, soa_dir=DEFAULT_SOA_DIR):
```
nhandler: Just steal the `soa_dir` from `args.yelpsoa_config_root` like we do on line 314

paasta_tools/cli/utils.py
```diff
@@ -325,13 +325,13 @@ def guess_instance(service, cluster, args):
     return instance
 
 
-def guess_cluster(service, args):
+def guess_cluster(service, args, soa_dir=DEFAULT_SOA_DIR):
```
nhandler: As above, use `args.yelpsoa_config_root`

nhandler:
Fix and Ship

paasta_tools/cli/utils.py
EvanKrall: oh oops I didn't mean to change this line

EvanKrall: but I probably should go ahead and make guess_instance do the right thing too

PR #704, opened 2016-08-24
Remove yelpy links to jenkins pipeline

kkellyy:
Let's reimplement this in #710 in a more generic way.

### DIFF
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -40,6 +40,7 @@ from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import get_git_url
 from paasta_tools.utils import get_service_instance_list
 from paasta_tools.utils import is_deploy_step
+from paasta_tools.utils import JENKINS_HOST
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import PaastaColors
 
@@ -294,7 +295,7 @@ def deployments_check(service, soa_dir):
 
 
 def pipeline_check(service):
-    url = "https://jenkins.yelpcorp.com/view/services-%s/api/xml" % service
+    url = "%s/view/services-%s/api/xml" % (JENKINS_HOST, service)
     try:
         req_status = urllib2.urlopen(url).getcode()
         if req_status == 200:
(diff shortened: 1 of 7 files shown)


nhandler:
Looks good to me. Let me know when you are ready to actually ship this change.

EvanKrall:
We should make this configurable through SystemPaastaConfig (but not necessarily make @kkellyy do that work, so I'm okay merging this for now.)

solarkennedy:
Would you guys care if we removed this feature entirely?

nhandler:
I think until it becomes easier to discover these pipelines in the Jenkins UI, it is nice to have. But if anyone has strong feelings about removing it, I would be fine with that. Jenkins really isn't a strict dependency of PaaSTA, and this piece of code is a Yelpism.

solarkennedy:
Even at yelp with y-m jenkins isn't used like that. And with this change as it, it will be incorrect. I would rather not even report it, so my vote goes to simply remove it. I don't think you need to be reminded of a jenkins url every time you run status.

EvanKrall:
I'm okay with killing it, potentially replacing this with something similar to `dashboard_links`.

kkellyy:
I don't really mind moving it, but I'm curious if other ways to discover these pipelines exists. If they don't then it might be best to merge this now and eventually remove this yelpism later.

solarkennedy:
But if you merge now, won't all the links be "wrong"? I would rather just remove the link if it is going to be wrong.

kkellyy:
Only wrong if we deployed it immediately. This link is soon to be right.

tub:
Remember that this is also effects the `paasta check` command which compares the Jenkins pipeline to the soa configs.
I'm not against removing that check, but it should be part of a larger refactor that properly separates paasta and the soa-configs (or at least makes it pluggable).

solarkennedy:
@kkellyy we agreed to remove this yelpism from the code base in favor of something more generic.

Can you make this PR more "red" by removing all references to jenkins?

solarkennedy:
@kkellyy can you merge or rebase?

Also be sure to run tox and see flake8 issues.

solarkennedy:
I've rebased and merged.

PR #705, opened 2016-08-25
add ulimit and cap_add settings to json schema

gstarnberger:
I recently added two new options, but forgot to add them to the json schema for the marathon config.


Rob-Johnson:
lg2m, ty!

Issue #701, opened 2016-08-23
NoSlavesAvailable message should be more helpful with incorrect deploy_whitelist/blacklist


EvanKrall:
When you've got a deploy_whitelist with values that don't exist on your cluster (say, you typo an attribute value), `paasta status` will show something like:

```
$ paasta status -s mycoolservice -c norcal-prod
Pipeline: https://jenkins.yelpcorp.com/view/services-mycoolservice
 
cluster: norcal-prod
instance: coolinstance
Git sha:    916bba50 (desired)
ERROR:__main__:Exception raised while looking at service mycoolservice instance coolinstance:
ERROR:__main__:Traceback (most recent call last):
  File "/usr/bin/paasta_serviceinit", line 122, in main
    delta=args.delta,
  File "/usr/share/python/paasta-tools/local/lib/python2.7/site-packages/paasta_tools/marathon_serviceinit.py", line 351, in perform_command
    app_id = job_config.format_marathon_app_dict()['id']
  File "/usr/share/python/paasta-tools/local/lib/python2.7/site-packages/paasta_tools/marathon_tools.py", line 396, in format_marathon_app_dict
    'constraints': self.get_calculated_constraints(service_namespace_config),
  File "/usr/share/python/paasta-tools/local/lib/python2.7/site-packages/paasta_tools/marathon_tools.py", line 322, in get_calculated_constraints
    constraints.extend(self.get_routing_constraints(service_namespace_config))
  File "/usr/share/python/paasta-tools/local/lib/python2.7/site-packages/paasta_tools/marathon_tools.py", line 331, in get_routing_constraints
    whitelist=self.get_deploy_whitelist())
  File "/usr/share/python/paasta-tools/local/lib/python2.7/site-packages/paasta_tools/mesos_tools.py", line 523, in get_mesos_slaves_grouped_by_attribute
    raise NoSlavesAvailable("No mesos slaves were available to query. Try again later")
NoSlavesAvailable: No mesos slaves were available to query. Try again later

ssh -A -o LogLevel=QUIET -t 10.1.2.3 sudo paasta_serviceinit -s mycoolservice -i coolinstance status exited with non-zero return code.
```

The exception message should indicate that this can be caused by too-restrictive `deploy_whitelist` / `deploy_blacklist`.

PR #706, opened 2016-08-25
use the /slaves endpoint rather than /state where possible

Rob-Johnson:
``mesos_tools.get_mesos_slaves_grouped_by_attribute`` previously called the ``/state`` endpoint to gather info about slaves. switch to using the ``/slaves`` endpoint instead, and refactor a little to move some filtering to the callers.

I think this could be a fairly big win, given that it was called in the bounce lib's ``get_happy_tasks`` function

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -30,9 +30,12 @@ from marathon import MarathonClient
 from marathon import MarathonHttpError
 from marathon import NotFoundError
 
+from paasta_tools.mesos_tools import filter_mesos_slaves_by_blacklist
 from paasta_tools.mesos_tools import get_local_slave_state
 from paasta_tools.mesos_tools import get_mesos_network_for_net
 from paasta_tools.mesos_tools import get_mesos_slaves_grouped_by_attribute
+from paasta_tools.mesos_tools import get_slaves
+from paasta_tools.mesos_tools import NoSlavesAvailableError
 from paasta_tools.paasta_maintenance import get_draining_hosts
 from paasta_tools.utils import compose_job_id
 from paasta_tools.utils import decompose_job_id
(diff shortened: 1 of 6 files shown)


Rob-Johnson:
turns out that the mesos.cli implementation for getting the slaves doesn't use the ``/slaves`` endpoint, but uses ``/state`` and grabs the right parts, which is why I'm doing the _fetch

paasta_tools/marathon_tools.py
```diff
@@ -325,12 +328,33 @@ def get_calculated_constraints(self, service_namespace_config):
...
+        """
         discover_level = service_namespace_config.get_discover()
-        locations = get_mesos_slaves_grouped_by_attribute(
-            attribute=discover_level, blacklist=self.get_deploy_blacklist(),
-            whitelist=self.get_deploy_whitelist())
-
-        routing_constraints = [[discover_level, "GROUP_BY", str(len(locations))]]
+        slaves = get_slaves()
+        if not slaves:
+            raise NoSlavesAvailableError
```
Rob-Johnson: ah, yes. will fix.

Rob-Johnson:
cool, the itest caught some bugs. will fix tomorrow

paasta_tools/marathon_tools.py
EvanKrall: I like that idea.

paasta_tools/mesos_tools.py
```diff
@@ -520,30 +519,26 @@ def get_number_of_mesos_masters(host, path):
...
-                attr_map.setdefault(attr_val, []).append(slave['hostname'])
-        return attr_map
+    sorted_slaves = sorted(
+        slaves,
+        key=lambda slave: slave['attributes'].get(attribute, 'unknown')
+    )
+    return {key: list(group) for key, group in itertools.groupby(
+        sorted_slaves,
+        key=lambda slave: slave['attributes'].get(attribute)
+    )}
```
EvanKrall: I'm not sure this sorted/groupby chain is actually clearer than the old for-loop.

Subtle change of behavior though: previously, we'd ignore any slaves where the attribute was unset; now, I believe those end up under the key `None`.

paasta_tools/marathon_tools.py
```diff
@@ -324,12 +327,44 @@ def get_calculated_constraints(self, service_namespace_config):
...
+            raise NoSlavesAvailableError(
+                "No slaves could be found in the cluster."
+            )
+        filtered_slaves = filter_mesos_slaves_by_blacklist(
+            slaves=slaves,
+            blacklist=self.get_deploy_blacklist(),
+            whitelist=self.get_deploy_whitelist(),
+        )
+        if not filtered_slaves:
+            raise NoSlavesAvailableError(
```
mattmb: Is it definitely ok to let this bubble up? I'd guess it is okay but haven't checked.

mattmb:
This LGTM. I don't know this area of the code very well but it looks like a sensible change.

Have you checked the slaves endpoint returns attributes? Other than that I'd say shipit!

paasta_tools/marathon_tools.py
Rob-Johnson: nothing has changed here (see the stacktrace in #701), but this fixes #701 by providing more info.

Rob-Johnson:
@mattmb yes, I've double checked the slaves endpoint returns attributes: try it with
```
curl -s http://master:5050/master/slaves | jq ".slaves[].attributes"
```

PR #707, opened 2016-08-25
Removed zookeper scaling itests

solarkennedy:
Dare me?

The argument here is that these are providing relatively little value if we already have confidence that `get_instances()` works correctly, regardless of whether that value comes from ZK or from on disk.

We already have itests that setup_marathon_job can smoothly scale up and down in response to instance count.


nhandler:
I think the tests are nice to have, and will help us prevent regressions. However, I agree that the flake right now is hurting more than its helping. My vote would be to Ship this change and ticket further investigation into the itests with the goal of re-adding them in a non-flakey manner.

EvanKrall:
These tests are super painful. I agree with @nhandler - ship and ticket re-adding the tests.

mattmb:
Shipit!

PR #709, opened 2016-08-26
use the full state endpoint in metastatus again

### LINKED ISSUE #708: paasta metastatus is no longer accurate in -vv and -vvv
This is a result of the change in the 0.42 release (which changed the endpoint metastatus used).

Rob-Johnson:
this fixes #708 until I can fix this 'properly'. @solarkennedy I've run the itest locally and the output is correct again - please merge + release as we discussed? FYI, I couldn't just revert the previous commit because it included some other logic now required, including adding the ``get_mesos_state_summary_from_leader`` function. (me--)


solarkennedy:
There are unit tests and stuff? Can't I just do a full revert?

PR #713, opened 2016-08-26
Removed external tron dependency

huadongliu:
bravado swagger client is preferred by paasta-api. bravado requires 'twisted >= 14.0.0'. However, paasta-tools currently uses tron which requires 'Twisted >=10.0.0, <=12.3'. This patch is to copy the minimum required files from tron to paasta-tools. 

I didn't pull corresponding unit tests because command_context_test.py requires other tron modules. timeutils_test.py is clean but I want to treat it the same as command_context_test.py.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,6 +7,8 @@ omit =
     /usr/*
     */tmp*
     setup.py
+    paasta_tools/tron_command_context.py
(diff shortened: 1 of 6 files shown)


solarkennedy:
I say shipit if the tests pass. I'm pretty sure there is at least one test that includes this integration.

PR #714, opened 2016-08-26
Manpage Version of Generated Docs

nhandler:
We have a lot of useful information in our generated documentation. Sometimes, it is useful to be able to reference it right from the terminal without needing to open up a browser. This change will allow you to type `man paasta-tools` to do so.

I also fixed a few warnings that were appearing because of ioslation.rst

I was also toying with the idea of generating PDF or single-page HTML docs, but I felt that might be overkill.

### DIFF
--- a/build-manpages.sh
+++ b/build-manpages.sh
@@ -49,3 +49,7 @@ done
 
 # And then finally the "main" paasta command
 help2man --name='paasta' --version-string=$VERSION "paasta" > docs/man/paasta.1
+
+# Make our generated docs available at 'man paasta_tools"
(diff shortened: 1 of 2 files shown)


solarkennedy:
Oh wow, I've wanted this. Does it actually work though? Because the tox target for man page building is distinct.

nhandler:
`make itest_lucid` causes `docs/man/paasta-tools.1` to be created. This then gets included by `debian/rules` in the final deb file. I can view the manpage just fine, but it could probably do with being split up into multiple pages like we do on readthedocs.

solarkennedy:
ship

PR #717, opened 2016-08-30
Added skeleton for paasta_maint is_safe_to_drain

solarkennedy:
This is mostly more boilerplate, the hard part is filling in `are_local_tasks_in_danger()`

### DIFF
--- a/paasta_tools/paasta_maintenance.py
+++ b/paasta_tools/paasta_maintenance.py
@@ -18,8 +18,8 @@ import json
 import logging
 import sys
 from socket import getfqdn
-from socket import gethostname
 from socket import gethostbyname
+from socket import gethostname
 
 from dateutil import parser
 from pytimeparse import timeparse
@@ -490,11 +490,11 @@ def is_safe_to_drain(hostname):
             "on localhost."
         )
         return False
-    return are_local_tasks_in_danger()
+    return not are_local_tasks_in_danger()
 
 
 def are_local_tasks_in_danger():
-    return True
+    return False
 
 
 def is_host_drained(hostname):
@@ -604,7 +604,8 @@ def paasta_maintenance():
         return is_host_past_maintenance_start(hostnames[0])
     elif action == 'is_host_past_maintenance_end':
         return is_host_past_maintenance_end(hostnames[0])
-
+    else:
+        raise NotImplementedError("Action: '%s' is not implemented." % action)
 
 if __name__ == '__main__':
     if paasta_maintenance():
--- a/paasta_tools/paasta_maintenance.py
+++ b/paasta_tools/paasta_maintenance.py
@@ -18,6 +18,7 @@ import json
 import logging
 import sys
 from socket import getfqdn
+from socket import gethostname
 from socket import gethostbyname
 
 from dateutil import parser
@@ -60,6 +61,7 @@ def parse_args():
             'is_host_draining',
             'is_hosts_past_maintenance_end',
             'is_hosts_past_maintenance_start',
+            'is_safe_to_drain',
             'is_safe_to_kill',
             'schedule',
             'status',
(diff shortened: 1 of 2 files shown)


paasta_tools/paasta_maintenance.py
```diff
@@ -472,6 +474,29 @@ def is_safe_to_kill(hostname):
...
+        print (
+            "Due to the way is_safe_to_drain is implemented, it can only work "
+            "on localhost."
+        )
+        return False
+    return are_local_tasks_in_danger()
+
+
+def are_local_tasks_in_danger():
+    return True
```
EvanKrall: Might be worth having this return False for now, so that you can start hooking this into stuff before this is fully implemented.

solarkennedy: Are you sure? I need to to return true by default, otherwise all the preflight checks will fail ASAP.

paasta_tools/paasta_maintenance.py
```diff
@@ -472,6 +474,29 @@ def is_safe_to_kill(hostname):
...
+
+def is_safe_to_drain(hostname):
+    """Checks if a host has healthy tasks running locally that have low
+    replication in other places
+    :param hostname: hostname to check
+    :returns: True or False
+    """
+    if not is_hostname_local(hostname):
+        print (
+            "Due to the way is_safe_to_drain is implemented, it can only work "
```
EvanKrall: Well, how are you planning to implement this? I was assuming you'd take a look at the services running on that host, see whether any of them are already under-replicated, maybe look at cluster metastatus to see whether there's room for those services.

paasta_tools/paasta_maintenance.py
```diff
@@ -542,23 +567,6 @@ def paasta_maintenance():
...
-            'is_host_past_maintenance_start',
-            'is_safe_to_kill',
-            'schedule',
-            'status',
-            'undrain',
-            'up',
-    ]:
-        print "action must be 'drain', 'undrain', 'down', 'up', 'status', or 'schedule'"
-        return
-
```
EvanKrall: Since you're deleting this check, can you add an else at the end of the if-elif-elif-elif... chain below? Or maybe refactor this function so that you don't have to add subcommands in multiple places. (Maybe use argparse.add_subparser?)

solarkennedy: The arg-parser already has "choices", which should be enough
https://docs.python.org/3/library/argparse.html#choices

paasta_tools/paasta_maintenance.py
EvanKrall: The name of this function implies that all local tasks are assumed to be in danger until you create an implementation.

I'm assuming that you would not e.g. run puppet if local tasks are in danger, so as this is written, you'd be preventing anything with a pre-flight check from running.

paasta_tools/paasta_maintenance.py
solarkennedy: I was hoping to reuse `get_marathon_services_running_here_for_nerve()`, as they represent the ones that are going to be in smartstack in the first place, and cross-reference that with `check_synapse_replication`.

I don't want this call to be expensive, remember we are going to be calling it once per hour for every paasta slave.

paasta_tools/paasta_maintenance.py
solarkennedy: Oh right, I understand. You are right I need to correct the "not"s or change the wording.

paasta_tools/paasta_maintenance.py
solarkennedy: I also added an else

paasta_tools/paasta_maintenance.py
EvanKrall: Technically all that information is available over the network, though our code for retrieving it is all centered on the current host. I'm fine with this for now; we can modify later if we find a need for is_safe_to_drain remotely.

solarkennedy: Got it.

EvanKrall:
ship

PR #719, opened 2016-08-31
Make paasta logs verbose by default

solarkennedy:
Also print scribereader commands if using scribe


solarkennedy:
Note that this is "application verbose" as in it will show stdout output and junk.

Rob-Johnson:
shipit

PR #720, opened 2016-09-01
Fix itest_trusty libffi error

huadongliu:
This is to address jenkins itest_trusty failure.

03:45:57 dpkg-shlibdeps: error: couldn't find library libffi.so.5 needed by debian/paasta-tools/usr/share/python/paasta-tools/lib/python2.7/site-packages/_cffi_backend.so (ELF format: 'elf64-x86-64'; RPATH: '')

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -26,4 +26,4 @@ override_dh_auto_test:
 	true
 
 override_dh_virtualenv:
-	dh_virtualenv --extra-index-url 'https://pypi.yelpcorp.com/simple' --python=/usr/bin/python2.7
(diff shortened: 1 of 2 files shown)


solarkennedy:
I'm surprised the dev package doesn't also pull it in?

solarkennedy:
ship

huadongliu:
I was able to reproduce the failure on both jenkins and devbox with make itest_trusty. Adding libffi does NOT fix the problem. 

I am not sure why it succeeds on github boxes.

huadongliu:
The bravado commit will be reverted.

PR #721, opened 2016-09-02
Revert "Replace paasta-api client with swagger client"

huadongliu:
This reverts commit 33d9e781685355c68884a57637068c650bde8cb1.

### DIFF
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -11,8 +11,12 @@
   "produces": ["application/json"],
   "tags": [
     {
-      "name": "service",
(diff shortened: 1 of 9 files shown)


mattmb:
Oh, did something break? 😢

solarkennedy:
No, the root cause of this was the fact that when we build at yelp we have python-wheels available, but they are built against lucid libraries. In open-source land there are no wheels, so everything is built from source.

The final solution to this was to include the libffi dev packages and add "--no-use-wheels" to build from source in both jenkins/travis (yelp/opensource) to get it to work in both places.

The revert was just a temporary measure to fix the build till we figured the above out.

PR #723, opened 2016-09-02
move tron related tools to a sub package

Rob-Johnson:
The paasta_tools package is fairly disorganised - I'd like to start a pattern of creating sub-packages to improve the situation. I'll start here, by moving the tron related code to it's own subpackage.


solarkennedy:
Seems fine?

PR #724, opened 2016-09-02
Replace paasta-api client with swagger client

huadongliu:
This is to restore the previous revert by adding no-use-wheel option to pip.


solarkennedy:
ship

Rob-Johnson:
shipit

PR #725, opened 2016-09-02
Remove bounce_progress sensu ttl alert

solarkennedy:
I gave this alert a good-ole college try, but it was just *way* too noisy.
There are just too many ways for bounce progress to slow down.

Also in the lucy case the bounce is pretty much *always* in progress anyway.


Rob-Johnson:
shipit

PR #727, opened 2016-09-06
Add pytz to setup.py

mattmb:
I guess this used to be captured as a dependency of something...

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -56,6 +56,7 @@ setup(
         'pyramid == 1.7',
         'pysensu-yelp >= 0.2.2',
         'pytimeparse >= 1.1.0',
+        'pytz >= 2014.10',
         'python-dateutil >= 2.4.0',
         'requests == 2.6.2',
         'requests-cache >= 0.4.10,<= 0.5.0',


setup.py
```diff
@@ -56,6 +56,7 @@
         'pyramid == 1.7',
         'pysensu-yelp >= 0.2.2',
         'pytimeparse >= 1.1.0',
+        'pytz == 2014.10',
```
nhandler: Do we really need this specific version? I imagine any recent version would be sufficient. Can you switch this to a `>=` relationship? Timezone databases change all the time, so supporting more recent versions of the module would be preferred.

nhandler:
Fix and Ship!

solarkennedy:
ship

setup.py
EvanKrall: or at least do something from 2016, like 2016.6.1

mattmb:
I duped this from requirements.txt which is why it's pinned. I agree that there's probably no need, I'll unpin and then ship this if the tests pass.

PR #728, opened 2016-09-06
Autoscaler box decision

mattmb:
* Use state summary where possible in autoscaler
* Check which box is best to kill after each termination. This is useful because mesos will have changed the distribution of tasks when draining the first box it kills.
* Split out get_count_running_tasks_on_slave into mesos tools to keep paasta_maintenance cleaner.

### DIFF
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -39,6 +39,7 @@ from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.marathon_tools import MESOS_TASK_SPACER
 from paasta_tools.marathon_tools import set_instances_for_marathon_service
 from paasta_tools.mesos_tools import get_mesos_state_from_leader
+from paasta_tools.mesos_tools import get_mesos_state_summary_from_leader
 from paasta_tools.mesos_tools import get_mesos_task_count_by_slave
 from paasta_tools.mesos_tools import get_running_tasks_from_active_frameworks
 from paasta_tools.mesos_tools import slave_pid_to_ip
@@ -522,7 +523,8 @@ def describe_instances(instance_ids, region=None, instance_filters=None):
 
 
 @register_autoscaling_component('aws_spot_fleet_request', CLUSTER_METRICS_PROVIDER_KEY)
-def spotfleet_metrics_provider(spotfleet_request_id, mesos_state, resource, pool_settings):
+def spotfleet_metrics_provider(spotfleet_request_id, resource, pool_settings):
+    mesos_state = get_mesos_state_from_leader()
     sfr = get_sfr(spotfleet_request_id, region=resource['region'])
     if not sfr or not sfr['SpotFleetRequestState'] == 'active':
         log.error("Ignoring SFR {0} that does not exist or is not active.".format(spotfleet_request_id))
@@ -601,23 +603,6 @@ def get_spot_fleet_delta(resource, error):
     return current_capacity, new_capacity
 
 
-def sort_slaves_to_kill(mesos_state, pool='default'):
-    """Pick the best slaves to kill. This returns a list of slaves
-    after sorting in preference of which slaves we kill first.
-    It sorts first by number of chronos tasks, then by total number of tasks
-
-    :param mesos_state: mesos_state dict
-    :param pool: pool of slaves to consider
-    :returns: list of slaves"""
-    slaves = get_mesos_task_count_by_slave(mesos_state, pool=pool)
-    if slaves:
-        slaves_by_task_count = [slave.slave for slave in sorted(slaves.values(),
-                                                                key=lambda x: (x.chronos_count, x.count))]
-        return slaves_by_task_count
-    else:
-        return []
-
-
 def wait_and_terminate(slave, drain_timeout, dry_run, region=None):
     """Currently kills a slave, will wait for draining to complete soon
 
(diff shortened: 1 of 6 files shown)


mattmb:
We don't have good itest coverage here but I have tested this to some extent in mesosstage and locally.

paasta_tools/autoscaling_lib.py
```diff
@@ -664,23 +649,38 @@ def wait_and_terminate(slave, drain_timeout, dry_run, region=None):
...
 
 
+def sort_slaves_to_kill(slaves):
+    """Pick the best slaves to kill. This returns a list of slaves
+    after sorting in preference of which slaves we kill first.
+    It sorts first by number of chronos tasks, then by total number of tasks
+
+    :param slaves: list of slaves dict
+    :returns: list of slaves dicts"""
+    if slaves:
```
Rob-Johnson: if slaves is a list, then you can just return the list comprehension (because the comprehension over an empty list will return an empty list)

paasta_tools/autoscaling_lib.py
```diff
@@ -664,23 +649,38 @@ def wait_and_terminate(slave, drain_timeout, dry_run, region=None):
...
+        return []
+
+
 def get_instance_type_weights(sfr):
     launch_specifications = sfr['SpotFleetRequestConfig']['LaunchSpecifications']
     return {ls['InstanceType']: ls['WeightedCapacity'] for ls in launch_specifications}
 
 
-def filter_sfr_slaves(sorted_slaves, resource):
+def filter_sfr_slaves(slaves_list, resource):
```
Rob-Johnson: slave_list

paasta_tools/autoscaling_lib.py
```diff
@@ -664,23 +649,38 @@ def wait_and_terminate(slave, drain_timeout, dry_run, region=None):
...
+def sort_slaves_to_kill(slaves):
+    """Pick the best slaves to kill. This returns a list of slaves
+    after sorting in preference of which slaves we kill first.
+    It sorts first by number of chronos tasks, then by total number of tasks
+
+    :param slaves: list of slaves dict
+    :returns: list of slaves dicts"""
+    if slaves:
+        slaves_by_task_counts = [slave for slave in
+                                 sorted(slaves, key=lambda x: (x['task_counts'].chronos_count, x['task_counts'].count))]
```
Rob-Johnson: why don't you just ``return sorted(slaves, key=lambda x: (x['task_counts'].chronos_count, x['task_counts'].count))``?

paasta_tools/mesos_tools.py
```diff
@@ -653,14 +660,34 @@ def get_mesos_task_count_by_slave(mesos_state, pool=None):
             log.debug("Task framework: {0}".format(task.framework.name))
             if task.framework.name == 'chronos':
```
Rob-Johnson: make this a constant whilst you're here?

paasta_tools/mesos_tools.py
```diff
@@ -627,15 +627,22 @@ def get_mesos_network_for_net(net):
...
 
     :param mesos_state: mesos state dict
+    :param slaves_list: a list of slave dicts to count running tasks for.
     :param pool: pool of slaves to return (None means all)
-    :returns: dict of {<slave_id>:{'count': <count>, 'slave': <slave_object>, 'chronos_count':<chronos_count>}}
+    :returns: list of slave dicts {'task_count': SlaveTaskCount}
     """
     all_mesos_tasks = get_running_tasks_from_active_frameworks('')  # empty string matches all app ids
-    if pool:
+    if slaves_list:
```
Rob-Johnson: I think this function could be made slightly simpler:

lines 639-653 can be replaced with:
```
slaves = slave['id']: {'count': 0, 'slave': slave, 'chronos_count': 0} for slave in mesos_state.get('slaves', [])
if pool:
  slaves = [slave for slave in slaves if slave['attributes'].get('pool', 'default') == pool]
elif slaves:
  slave_hosts = [slave['hostname'] for slave in slaves_list]
  slaves = [slave for slave in slaves if slave['hostname'] in slave_hosts
```

Rob-Johnson:
shipit - I'd really like us to split this up though, it's getting difficult to read

paasta_tools/autoscaling_lib.py
mattmb: Yes indeed, that should work. It's only like this because it used to be actually getting a sub-object.

paasta_tools/mesos_tools.py
mattmb: Yeah, it's got a bit nasty hasn't it! I think there's one more case where we get slaves from all pools with no slaves_list. I like the idea of moving the dict comprehension out of the conditional though 👍 I'll try this out and send a fixup.

paasta_tools/autoscaling_lib.py
mattmb: I think the plural is ok here. It's a list of slaves rather than a list of things about a slave.

solarkennedy:
I also say ship, but `scale_aws_spot_fleet_request` is getting too complex. A hint of this is all the mocks required to test it.

mattmb:
Yep, I agree, the length of the unit test is also an indicator that it's doing too much!

I'll do the refactoring before/during my next ticket which is to make the autoscaler clean up cancelled SFRs.

PR #729, opened 2016-09-06
fix brokeness from get_mesos_slaves_grouped_by_attribute

Rob-Johnson:
we were mocking get_mesos_slaves_grouped_by_attribute in a lot of
places, so the api changes weren't noticed by the tests. this fixes
it's usage in more places

### DIFF
--- a/paasta_tools/check_marathon_services_replication.py
+++ b/paasta_tools/check_marathon_services_replication.py
@@ -361,12 +361,23 @@ def get_smartstack_replication_for_attribute(attribute, service, namespace, blac
               (the dictionary will contain keys for unique all attribute values)
     """
     replication_info = {}
-    unique_values = mesos_tools.get_mesos_slaves_grouped_by_attribute(attribute=attribute, blacklist=blacklist)
+    filtered_slaves = mesos_tools.get_all_slaves_for_blacklist_whitelist(
+        blacklist=blacklist,
+        whitelist=[],
+    )
+    if not filtered_slaves:
+        raise mesos_tools.NoSlavesAvailableError
+
+    attribute_slave_dict = mesos_tools.get_mesos_slaves_grouped_by_attribute(
+        slaves=filtered_slaves,
+        attribute=attribute
+    )
+
     full_name = compose_job_id(service, namespace)
 
-    for value, hosts in unique_values.iteritems():
+    for value, hosts in attribute_slave_dict.iteritems():
         # arbitrarily choose the first host with a given attribute to query for replication stats
-        synapse_host = hosts[0]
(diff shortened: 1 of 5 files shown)


paasta_tools/check_marathon_services_replication.py
```diff
@@ -361,12 +361,24 @@ def get_smartstack_replication_for_attribute(attribute, service, namespace, blac
...
+    )
+    if not filtered_slaves:
+        raise mesos_tools.NoSlavesAvailableError
+
+    attribute_slave_dict = mesos_tools.get_mesos_slaves_grouped_by_attribute(
+        slaves=filtered_slaves,
+        attribute=attribute
+    )
+
+    print attribute_slave_dict
```
mattmb: I'm guessing you didn't mean to print this?

paasta_tools/check_marathon_services_replication.py
```diff
@@ -361,12 +361,24 @@ def get_smartstack_replication_for_attribute(attribute, service, namespace, blac
...
+        blacklist=blacklist,
+        whitelist=[],
+    )
+    if not filtered_slaves:
+        raise mesos_tools.NoSlavesAvailableError
+
+    attribute_slave_dict = mesos_tools.get_mesos_slaves_grouped_by_attribute(
+        slaves=filtered_slaves,
+        attribute=attribute
+    )
```
mattmb: I guess you could move the grouping into a wrapper function too as we repeat the `get_all_slaves_for_blacklist_whitelist` and `get_mesos_slaves_grouped_by_attribute` twice.

Not massively fussed though

paasta_tools/marathon_tools.py
```diff
@@ -1009,6 +1010,8 @@ def is_app_id_running(app_id, client):
     :param client: A MarathonClient object"""
 
     all_app_ids = list_all_marathon_app_ids(client)
+    print app_id
+    print all_app_ids
```
mattmb: I guess these are leftover debug too

paasta_tools/check_marathon_services_replication.py
mattmb: Oh wait, that's sort of what you've changed...

Rob-Johnson: eh, yes, we could, but the problem was that the retrieval was tightly coupled with the filtering before - the thing I wanted to achieve was their separation.

Rob-Johnson:
@mattmb thanks for spotting the leftover prints - I've cleaned them up now

mattmb:
Shipit!

PR #730, opened 2016-09-06
Add percentages for attribute usage in metastatus

mattmb:
So I don't have to keep getting out a calculator, this adds a percentage
utilization for per-attribute groupings in paasta metastatus.


Rob-Johnson:
shipit

PR #731, opened 2016-09-06
Make marathon_serviceinit filter mesos tasks properly

solarkennedy:
Well, I don't like this and neither should you. I would like a different opinion on what to do here.

This function returns mesos tasks, but it uses the mesos-cli `fltr` stuff which is just "x in task", so `main` matches `main_sfo`.

One way to stop this is to add a terminator at the end, which is what I did. We could put this at a different layer?

### DIFF
--- a/paasta_tools/marathon_serviceinit.py
+++ b/paasta_tools/marathon_serviceinit.py
@@ -383,7 +383,10 @@ def get_short_task_id(task_id):
 
 def status_mesos_tasks(service, instance, normal_instance_count):
     job_id = marathon_tools.format_job_id(service, instance)
-    running_and_active_tasks = get_running_tasks_from_active_frameworks(job_id)
+    # We have to add a spacer at the end to make sure we only return
+    # things for service.main and not service.main_foo
+    filter_string = "%s%s" % (job_id, marathon_tools.MESOS_TASK_SPACER)
+    running_and_active_tasks = get_running_tasks_from_active_frameworks(filter_string)
     count = len(running_and_active_tasks)
     if count >= normal_instance_count:
         status = PaastaColors.green("Healthy")


Rob-Johnson:
but if you add a '.' at the end, won't this fail to match anything? Like if the ``job_id`` is ``paasta--canary.mesosstage--main.gitc90db969.configa9a3de58.f7996025-74e7-11e6-bacd-e6bdd18094a7``, and now you filter on that same string with a "." on the end (MESOS_TASK_SPACER), then it'll fail, right?

tbh, I'd just reimplement this in ``mesos_tools`` with exact matching (or adjust it to support a regexp match if you think we need it)

```
import mesos.cli.Task
...
def tasks(job_id, active_only=False):
   # return list(map(
   #          lambda x: task.Task(self, x),
   #          itertools.ifilter(
   #              lambda x: fltr in x["id"] or fnmatch.fnmatch(x["id"], fltr),
   #              self._task_list(active_only))))
   return [ 
      task.Task(x) for x
      in master.CURRENT._tasks(active_only)
      if job_id == x["id"]
   ]
        
```

solarkennedy:
@Rob-Johnson Nope, you are confusing our job id with the full task id in mesos.

Before this patch:
```
filter="service.main"
tasks:
   service.main.gitsha.config.skdjfskdljfajkdf
   service.main_sfo2.gitsha.config.sdfjksdjflkj
```
After this patch:
```
filter="service.main."
tasks:
   service.main.gitsha.config.skdjfskdljfajkdf
```

Rob-Johnson:
yes, sorry I missed that. I'd still rather see this in ``mesos_tools`` in a more generic function. I'm really surprised we haven't needed this before.

```python
def get_tasks_for_service_instance(service, instance, active_only=False):
    matching_tasks = (
        task for task in 
        master.CURRENT.tasks(active_only) 
        if (service, instance) ==  tuple(x["id"].split(MESOS_TASK_SPACER)[:2])
    )
    return [task.Task(x) for x in matching_tasks]    
```
but I'm not too fussed. Note that I swapped out the itertools.ifilter stuff here for a generator expression, too.

solarkennedy:
Hmm. I agree that function looks nicer.

solarkennedy:
Crap. Your suggestion just won't work very well. Look at these two example mesos task ids:
```
marketo--datapipeline.sfdc--account--contact--transform--spolt.git47e73cd1.config60295232.db31b679-75fe-11e6-b1c0-1e2b29fa0cb5
ct:1473364800000:0:paasta_canary sensu_canary:
```
What "spacer" should we use? What about underscore or hyphen transforms?

I think we need the get function to be generic and just take a string, and let the consumers of it (marathon_serviceinit, chronos_serviceinit) care about what string they are looking for exactly.

Rob-Johnson:
heh, right, which is what the previous version did. Back to your original PR then I guess. shipit!

PR #732, opened 2016-09-06
Disable service_configuration_lib caching of service configs in paasta-api

huadongliu:
service_configuration_lib caches files by default. This is a quick fix to unblock mmb by disabling service_configuration_lib caching completely. In the long term, pyinotify should offer a better solution since we don't expect service configs to change frequently. Unit test:

1. run paasta-api;
2. curl service/instance/status;
3. change services/service/marathon-cluster.yaml, e.g. mem;
4. curl service/instance/status again and verify config sha is different.

### DIFF
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -21,12 +21,14 @@ import os
 import sys
 
 import requests_cache
+import service_configuration_lib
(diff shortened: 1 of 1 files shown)


solarkennedy:
ship

PR #733, opened 2016-09-06
Make get_job_for_service_instance only return one chronos job ever

solarkennedy:
This function is kinda leftover from when we had multiple jobs per instance in chronos, where it was "ok" to see a bunch and discards all the duplicates and pick the most "recent" one.

This is now "wrong" in many cases and has lead to bugs where setup_chronos_job is picking the "first" one, which is a tmp job!

This change makes it so we don't even include disabled (tmp/rerun) jobs, and raises an exception if we ever are silently trying to discard duplicate jobs, which really should not happen anymore, right?

If I'm wrong, then at the very lease this thing should be called `get_jobs` and let the caller discard things I think? To me it is an exception if a caller asks to get the job for a service/instance and the code can't figure out which job is correct.

### DIFF
--- a/paasta_tools/chronos_serviceinit.py
+++ b/paasta_tools/chronos_serviceinit.py
@@ -170,7 +170,7 @@ def _format_parents_verbose(job):
     parent_service_instances = [tuple(chronos_tools.decompose_job_id(parent)) for parent in parents]
 
     # find matching parent jobs
-    parent_jobs = [chronos_tools.get_job_for_service_instance(*service_instance)
+    parent_jobs = [chronos_tools.get_job_for_service_instance(*service_instance, include_disabled=False)
                    for service_instance in parent_service_instances]
 
     # get the status of the last run of each parent job
--- a/paasta_tools/chronos_serviceinit.py
+++ b/paasta_tools/chronos_serviceinit.py
@@ -172,7 +172,7 @@ def _format_parents_verbose(job):
     # find matching parent jobs
     parent_jobs = [
         chronos_tools.get_jobs_for_service_instance(
-            *service_instance, include_disabled=False, include_temporary=False)[0]
+            *service_instance, include_disabled=True, include_temporary=False)[0]
         for service_instance in parent_service_instances
     ]
 
--- a/paasta_tools/chronos_serviceinit.py
+++ b/paasta_tools/chronos_serviceinit.py
@@ -170,8 +170,11 @@ def _format_parents_verbose(job):
     parent_service_instances = [tuple(chronos_tools.decompose_job_id(parent)) for parent in parents]
 
     # find matching parent jobs
-    parent_jobs = [chronos_tools.get_job_for_service_instance(*service_instance, include_disabled=False)
-                   for service_instance in parent_service_instances]
+    parent_jobs = [
+        chronos_tools.get_jobs_for_service_instance(
+            *service_instance, include_disabled=False, include_temporary=False)[0]
+        for service_instance in parent_service_instances
+    ]
 
     # get the status of the last run of each parent job
     parent_statuses = [(parent, _format_last_result(job)) for parent in parent_jobs]
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -154,6 +154,10 @@ class UnknownChronosJobError(Exception):
     pass
 
 
+class MultipleChronosJobs(Exception):
+    pass
+
+
 def read_chronos_jobs_for_service(service, cluster, soa_dir=DEFAULT_SOA_DIR):
     chronos_conf_file = 'chronos-%s' % cluster
     log.info("Reading Chronos configuration file: %s/%s/chronos-%s.yaml" % (soa_dir, service, cluster))
@@ -821,8 +825,8 @@ def get_job_for_service_instance(service, instance, include_disabled=True):
     :param service: the service to be queried
     :param instance: the instance to be queried
     :param include_disabled: include disabled jobs in the search. default: True
-    :returns: the most 'current' job for a given service instance; that is the most recent
-    job according to the result of sorting all the matching jobs.
+    :returns: the only job for a given service instance
+    An exception is raised if more than one job is returned.
     """
     chronos_config = load_chronos_config()
     matching_jobs = lookup_chronos_jobs(
@@ -832,10 +836,14 @@ def get_job_for_service_instance(service, instance, include_disabled=True):
         include_disabled=True
     )
     sorted_jobs = sort_jobs(matching_jobs)
-    if len(sorted_jobs) > 0:
+    if len(sorted_jobs) == 1:
         return sorted_jobs[0]
-    else:
+    elif len(sorted_jobs) == 0:
         return None
+    else:
+        raise MultipleChronosJobs(
+            "This function should never multiple jobs, but got: '%s'" % ",".join(map(str, sorted_jobs))
+        )
 
 
 def compose_check_name_for_service_instance(check_name, service, instance):
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -424,12 +424,9 @@ class ChronosJobConfig(InstanceConfig):
         if self.get_schedule() is not None:
             complete_config['schedule'] = self.get_schedule()
         else:
-            matching_parent_pairs = [(parent, get_job_for_service_instance(*parent.split(".")))
-                                     for parent in self.get_parents()]
-            for parent_pair in matching_parent_pairs:
-                if parent_pair[1] is None:
-                    raise InvalidParentError("%s has no matching jobs in Chronos" % parent_pair[0])
-            complete_config['parents'] = [parent_pair[1]['name'] for parent_pair in matching_parent_pairs]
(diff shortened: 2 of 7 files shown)


paasta_tools/chronos_serviceinit.py
```diff
@@ -170,7 +170,7 @@ def _format_parents_verbose(job):
     parent_service_instances = [tuple(chronos_tools.decompose_job_id(parent)) for parent in parents]
 
     # find matching parent jobs
-    parent_jobs = [chronos_tools.get_job_for_service_instance(*service_instance)
+    parent_jobs = [chronos_tools.get_job_for_service_instance(*service_instance, include_disabled=False)
```
huadongliu: This get_job_for_service_instance is used in the status_chronos_jobs code path only. The other get_job_for_service_instance is in https://github.com/Yelp/paasta/blob/8b83e67b7a8e4073010f35e7bd90c315c7570dfe/paasta_tools/chronos_tools.py#L388, which is used by chronos_rerun, chronos_serviceinit and setup_chronos_job.

Rob-Johnson:
> This change makes it so we don't even include disabled (tmp/rerun) jobs

You're right to say that we shouldn't be considering tmp jobs as parents, but I don't think that discounting *all* disabled jobs is the correct qualifier for that. That is, a disabled job doesn't always mean a tmp job. It's valid for a user to have non-temporary but disabled jobs (as in ``disabled: True`` in soa-configs) as it's parents.

> and raises an exception if we ever are silently trying to discard duplicate jobs, which really should not happen anymore, right?

It's totally valid for ``lookup_chronos_jobs`` to find > 1 job for a given service.instance, even with ``include_disabled=False``.  tmp jobs are only marked as disabled [once they finish](https://github.com/mesos/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/jobs/JobScheduler.scala#L286-L293).

I think the problem is that when doing the selection, the caller can't differentiate disabled from tmp jobs: having a >1 *non temp* disabled jobs is an error; the more general case of having >1 *disabled* job is not, as there may be temp jobs included in that list that haven't been cleaned up yet. 

So, here's my proposal:

- We should replace ``read_chronos_jobs_for_service`` with two new functions. In the one where we explicity *don't* want tmp jobs included, then raise an exception when we find >1.

```
def get_non_temporary_job_for_service_instance(..):
    # don't exclude disabled jobs in this lookup, because a regular job
    # can be disabled
    all_jobs_for_service = lookup_chronos_jobs(..., include_disabled=True)
    non_temporary_jobs = [job is job in all_jobs_for_service if not chronos_tools.is_temporary_job(job)]
    # now we know that if len(non_temporary_jobs) > 1, then we've hit an error
    if len(non_temporary_jobs) > 1:
       raise MultipleChronosJobsForServiceInstanceError("...")
    return non_temporary_jobs[1] if non_temporary_jobs else None

def get_temporary_jobs_for_service_instance(...):
   # tmp jobs are marked as disabled as soon as they finish, so
   # we need to include disabled. I don't know that there is a place 
   # where we use this (I can't think of a place where we *only* lookup tmp jobs,
   # so maybe YAGNI?
   all_jobs_for_service = lookup_chronos_jobs(..., include_disabled=True)
   # it's fine for there to be >1 of these, so no error validation
   return [job is job in all_jobs_for_service if chronos_tools.is_temporary_job(job)]

def lookup_chronos_jobs(include_disabled=True):
     #this interface stays the same, so that we can lookup *all* jobs
     #like we do in paasta status
```

Then, in all the places we *don't* want to consider tmp jobs (``format_chronos_job_dict``, where we build the list of parents for a job. ``_format_parents_verbose``, where we find a job object for a parent), we should use the ``get_non_temporary_job_for_service_instance``. 

But in places like [paasta status](https://github.com/Yelp/paasta/blob/917d730bb8c5de51fc57367f4a983438165c45dc/paasta_tools/chronos_serviceinit.py#L393-L406) where we want to find all jobs, we can use ``lookup_chronos_jobs`` to find them all.

solarkennedy:
@Rob-Johnson Gotcha. How about a new param instead of two whole new functions? (include_disabled/include_temporary) ?

solarkennedy:
Alright, sorry for pivoting here. There is more "red" and I think this is hopefully a better way to do it:

1. This change makes it so we no longer "lookup" parents in setup_chronos_job, we just set them. We don't nee the complexity of the lookup because we no longer have distinct jobs anymore.

2. I change the name of the function to `get_jobs`, to make it more clear that it might return more than one.

3. I added explicit options for including temporary or not, so the caller can make the choice.

paasta_tools/chronos_tools.py
```diff
@@ -154,6 +154,10 @@ class UnknownChronosJobError(Exception):
     pass
 
 
+class MultipleChronosJobs(Exception):
```
Rob-Johnson: can you make this ``MultipleChronosJobsError`` in line with pep

paasta_tools/chronos_serviceinit.py
```diff
@@ -170,8 +170,11 @@ def _format_parents_verbose(job):
     parent_service_instances = [tuple(chronos_tools.decompose_job_id(parent)) for parent in parents]
 
     # find matching parent jobs
-    parent_jobs = [chronos_tools.get_job_for_service_instance(*service_instance)
-                   for service_instance in parent_service_instances]
+    parent_jobs = [
+        chronos_tools.get_jobs_for_service_instance(
+            *service_instance, include_disabled=False, include_temporary=False)[0]
```
Rob-Johnson: ``include_disabled`` needs to be True here - It's fine for a parent job to be disabled, so we don't want to hide them in the output of status. It's just the temporary jobs we want to skip.

paasta_tools/chronos_tools.py
```diff
@@ -420,12 +424,9 @@ def format_chronos_job_dict(self, docker_url, docker_volumes, dockercfg_location
...
         else:
-            matching_parent_pairs = [(parent, get_job_for_service_instance(*parent.split(".")))
-                                     for parent in self.get_parents()]
-            for parent_pair in matching_parent_pairs:
-                if parent_pair[1] is None:
-                    raise InvalidParentError("%s has no matching jobs in Chronos" % parent_pair[0])
-            complete_config['parents'] = [parent_pair[1]['name'] for parent_pair in matching_parent_pairs]
+            # The input to parents is the normal paasta syntax, but for chronos we have to
+            # convert it to what chronos expects, which uses its own spacer
+            complete_config["parents"] = map(paasta_to_chronos_job_name, self.get_parents())
```
Rob-Johnson: the reason we did this pre-fetching was so that we could skip setting up a job if it's parents didn't yet exist in ``setup_chronos_job``. [here's the code where we caught this](https://github.com/Yelp/paasta/blob/master/paasta_tools/setup_chronos_job.py#L208-L210). without the validation, we fail later (when the request to Chronos fails) and less gracefully (we don't catch the right exception). this is the reason itests are failing atm. I don't mind failing later, but I think it's atleast worth catching this specific error in setup_chronos_job if it fails.

also, I'm surprised flake let you get away with the ``map`` - can you make it a list comprehension instead?

paasta_tools/chronos_tools.py
```diff
@@ -816,26 +827,22 @@ def update_job(client, job):
...
 
-def get_job_for_service_instance(service, instance, include_disabled=True):
+def get_jobs_for_service_instance(service, instance, include_disabled=True, include_temporary=False):
     """A wrapper for lookup_chronos_jobs that includes building the necessary chronos client.
     :param service: the service to be queried
     :param instance: the instance to be queried
     :param include_disabled: include disabled jobs in the search. default: True
-    :returns: the most 'current' job for a given service instance; that is the most recent
-    job according to the result of sorting all the matching jobs.
+    :returns: the jobs for a given service instance
```
Rob-Johnson: missed adding the ``include_temporary`` param

paasta_tools/chronos_tools.py
Rob-Johnson: oh - are we even using this anymore? I can't find a place you're raising it. Should this be in some caller - ``chronos_serviceinit`` maybe?

paasta_tools/chronos_tools.py
solarkennedy: Hah. Don't like map? :)

Yea I knew that this would make it so we "fail early" if the parents don't exist.
I thought this was "ok" as they would eventually get there and it wasn't quite worth the effort to kinda "look into the future."

Maybe as an optimization we can make `list_chronos_jobs` emit things parent-first to minimize early failures?

paasta_tools/chronos_serviceinit.py
solarkennedy: Ah right.

paasta_tools/chronos_tools.py
solarkennedy: Yea, I decided to stop trying to raise at this point and let the callers figure it out and let the function itself always return multiple jobs.

solarkennedy:
Still working on itests.

solarkennedy:
Ok. All tests pass now.

Rob-Johnson:
shipit - I think this looks good.

huadongliu:
Shipit.

PR #736, opened 2016-09-07
added a flag to list_marathon_service_instances to only show bouncing…

mjksmith:
… jobs

Sample output on internal JIRA ticket PAASTA-6418 (note this has been corrected to properly deformat job id)

This is about 185x faster than running setup_marathon_job on every service (maybe even faster now that we're not running multiple jobs per setup_marathon_job instance so we're not taking advantage of caching)

### DIFF
--- a/paasta_tools/list_marathon_service_instances.py
+++ b/paasta_tools/list_marathon_service_instances.py
@@ -52,7 +52,7 @@ def parse_args():
     parser.add_argument('-d', '--soa-dir', dest="soa_dir", metavar="SOA_DIR",
                         default=DEFAULT_SOA_DIR,
                         help="define a different soa config directory")
-    parser.add_argument('-m', '--minimal', dest='minimal', default=False,
+    parser.add_argument('-m', '--minimal', dest='minimal', action='store_true',
                         help="show only service instances that need bouncing")
     args = parser.parse_args()
     return args
--- a/paasta_tools/list_marathon_service_instances.py
+++ b/paasta_tools/list_marathon_service_instances.py
@@ -74,7 +74,7 @@ def get_current_apps():
     return {app.id.lstrip('/'): app for app in marathon_client.list_apps()}
 
 
-def get_desired_marathon_configs(cluster, soa_dir):
+def get_desired_marathon_configs(soa_dir):
     cluster = load_system_paasta_config().get_cluster()
     instances = get_services_for_cluster(
         instance_type='marathon',
--- a/paasta_tools/list_marathon_service_instances.py
+++ b/paasta_tools/list_marathon_service_instances.py
@@ -25,34 +25,106 @@ Command line options:
 
 - -d <SOA_DIR>, --soa-dir <SOA_DIR>: Specify a SOA config dir to read from
 - -c <CLUSTER>, --cluster <CLUSTER>: Specify which cluster of services to read
+- -m, --minimal: Only show service instances that need bouncing
 """
 import argparse
 import sys
 
 from paasta_tools.marathon_tools import DEFAULT_SOA_DIR
+from paasta_tools.marathon_tools import get_marathon_client
+from paasta_tools.marathon_tools import get_num_at_risk_tasks
+from paasta_tools.marathon_tools import load_marathon_config
+from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.utils import compose_job_id
+from paasta_tools.utils import decompose_job_id
 from paasta_tools.utils import get_services_for_cluster
+from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.utils import NoDockerImageError
+from paasta_tools.utils import use_requests_cache
 
 
 def parse_args():
-    parser = argparse.ArgumentParser(description='Lists marathon instances for a service.')
+    parser = argparse.ArgumentParser(
+        description='Lists marathon instances for a service.')
     parser.add_argument('-c', '--cluster', dest="cluster", metavar="CLUSTER",
                         default=None,
                         help="define a specific cluster to read from")
     parser.add_argument('-d', '--soa-dir', dest="soa_dir", metavar="SOA_DIR",
                         default=DEFAULT_SOA_DIR,
                         help="define a different soa config directory")
+    parser.add_argument('-m', '--minimal', dest='minimal', default=False,
+                        help="show only service instances that need bouncing")
     args = parser.parse_args()
     return args
 
 
+@use_requests_cache('list_marathon_services')
+def get_service_instances_that_need_bouncing(soa_dir):
+    def long_job_id_to_short_job_id(long_job_id):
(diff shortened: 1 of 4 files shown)


paasta_tools/list_marathon_service_instances.py
```diff
@@ -25,34 +25,114 @@
...
     parser.add_argument('-d', '--soa-dir', dest="soa_dir", metavar="SOA_DIR",
                         default=DEFAULT_SOA_DIR,
                         help="define a different soa config directory")
+    parser.add_argument('-m', '--minimal', dest='minimal', action='store_true',
+                        help="show only service instances that need bouncing")
     args = parser.parse_args()
     return args
 
 
+def long_job_id_to_short_job_id(long_job_id):
```
Rob-Johnson: this should probably be in marathon_tools

paasta_tools/list_marathon_service_instances.py
```diff
@@ -25,34 +25,114 @@
...
+    service, instance, _, __ = decompose_job_id(long_job_id)
+    return compose_job_id(service, instance)
+
+
+def get_current_apps():
+    marathon_config = load_marathon_config()
+    marathon_client = get_marathon_client(
```
Rob-Johnson: I'd suggest:

- moving this function marathon tools
- make it take a ``client`` as a parameter
- move the config loading + ``get_marathon_client`` call to ``main()`` as we do in other scripts like this ([here](https://github.com/Yelp/paasta/blob/master/paasta_tools/chronos_rerun.py#L158-L159), [here](https://github.com/Yelp/paasta/blob/master/paasta_tools/setup_marathon_job.py#L597-L598), [here](https://github.com/Yelp/paasta/blob/master/paasta_tools/setup_chronos_job.py#L174-L175) for some examples)

Rob-Johnson:
lg2m - fix and ship!

paasta_tools/list_marathon_service_instances.py
mjksmith: I'm not so sure. This method and the `get_desired_marathon_configs` method were originally part of `get_service_instances_that_need_bouncing` but I broke them apart to simplify testing. Plus, this method returns the apps in a format specific to this script (dictionary mapping id to app object) which might not be useful elsewhere, and passing around a marathon client from method to method to just use it once is kinda lame.

paasta_tools/list_marathon_service_instances.py
```diff
@@ -25,34 +25,114 @@
...
+    apps_that_need_bouncing = actual_ids.symmetric_difference(desired_ids)
+    apps_that_need_bouncing = {long_job_id_to_short_job_id(app) for app in apps_that_need_bouncing}
+
+    for app_id, app in current_apps.items():
+        short_app_id = long_job_id_to_short_job_id(app_id)
+        if short_app_id not in apps_that_need_bouncing:
+            if app.instances != desired_marathon_configs[app_id]['instances'] or get_num_at_risk_tasks(app) != 0:
+                apps_that_need_bouncing.add(short_app_id)
+
+    return (app_id.replace('--', '_') for app_id in apps_that_need_bouncing)
```
solarkennedy: This replace is risky, but I don't have any great suggestions. Sure we don't have something like this already?

tests/test_list_marathon_service_instances.py
```diff
@@ -0,0 +1,132 @@
...
+        mock.patch('paasta_tools.list_marathon_service_instances.get_desired_marathon_configs'),
+        mock.patch('paasta_tools.list_marathon_service_instances.get_current_apps'),
+        mock.patch('paasta_tools.list_marathon_service_instances.get_num_at_risk_tasks'),
+    ) as (
+        mock_get_desired_marathon_configs,
+        mock_get_current_apps,
+        mock_get_num_at_risk_tasks,
+    ):
+        mock_get_desired_marathon_configs.return_value = {'fake--service.fake--instance.sha.config': {'instances': 5}}
+        mock_get_current_apps.return_value = {'fake--service.fake--instance.sha.config': mock.MagicMock(instances=5)}
```
solarkennedy: Can you add a test specifically for when there is one desired app, but 2 different current apps returned?

I don't see such a test and could imagine that if we didn't get the symmetric difference logic correct, it might assume it doesn't need bouncing because the desired app is already in the set?

paasta_tools/list_marathon_service_instances.py
solarkennedy: You are duplicating `list_all_marathon_app_ids` no?

Passing the client around only "feels" lame because you only need the client a single time in your script.
Please pass the client around, even if you only end up using it one time for `list_all_marathon_app_ids()`

paasta_tools/list_marathon_service_instances.py
solarkennedy: This is "sorta" duplicating `get_matching_apps()`, which takes a service/instance and outputs the apps that match. (as opposed to looking at all the apps and reversing them)

Once nice approach to using `get_matching_apps` is that you only "look at" the ones that are legit.

How will this code tolerate the condition where a service was just deleted from soa-configs?

paasta_tools/list_marathon_service_instances.py
mjksmith: We use replace

paasta_tools/list_marathon_service_instances.py
mjksmith: It's not just `list_all_marathon_app_ids` because I also need a list of the current apps (to get the number of their instances). `list_all_marathon_app_ids` just returns their ids. I'll pass around the client though.

paasta_tools/list_marathon_service_instances.py
mjksmith: Services that are deleted from soa-configs will be marked as needing bouncing.

I didn't want to use `get_matching_apps` since I don't want the fuzzy matching. Also `get_matching_apps` and `get_matching_appids` go from short to long, which is the opposite of what I want.

tests/test_list_marathon_service_instances.py
mjksmith: It should see apps {'A.1'} vs {'A.1', 'A.2'} and return the symmetric difference {'A.2'} which indicates service 'A' needs bouncing. I'll add a test to confirm.

paasta_tools/list_marathon_service_instances.py
```diff
@@ -25,34 +25,109 @@
...
+
 def main():
     args = parse_args()
-    instances = get_services_for_cluster(cluster=args.cluster, instance_type='marathon', soa_dir=args.soa_dir)
-    composed = []
-    for name, instance in instances:
-        composed.append(compose_job_id(name, instance))
-    print '\n'.join(composed)
+    soa_dir = args.soa_dir
+    cluster = args.cluster
```
solarkennedy: We already grab the cluster here, but you look it up later.
I guess get_services_for_cluster understands `None`?

solarkennedy:
ship

PR #737, opened 2016-09-08
Add backoff_factor and max_launch_delay seconds to marathon schema

nhandler:
We added support for these options a while ago (and list them in our docs), but we never updated the schema file to support them. As a result, `paasta validate` claims they are invalid fields.


solarkennedy:
ship

PR #738, opened 2016-09-08
Disable caching on wait_for_app_to_launch_tasks

solarkennedy:
Ping @huadongliu, I'm not sure how we got by without this?

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -24,6 +24,7 @@ from math import ceil
 from time import sleep
 
 import requests
+import requests_cache
 import service_configuration_lib
 from kazoo.exceptions import NoNodeError
 from marathon import MarathonClient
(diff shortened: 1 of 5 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -1068,16 +1069,17 @@ def wait_for_app_to_launch_tasks(client, app_id, expected_tasks, exact_matches_o
...
-        try:
-            found = app_has_tasks(client, app_id, expected_tasks, exact_matches_only)
-        except NotFoundError:
-            pass
-        if found:
-            return
-        else:
-            print "waiting for app %s to have %d tasks. retrying" % (app_id, expected_tasks)
-            sleep(0.5)
+    with requests_cache.disabled():
```
huadongliu: I was hoping libraries like marathon_tools do not control caching behavior. Applications (setup_marathon_job, paasta_serviceinit etc) at a higher level are the ones that setup cache if needed, and disables cache temporarily. If the application does not setup cache, there is no need to disable cache at all. I would suggest move requests_cache.disabled up to the application that fails on caching.

solarkennedy: But it is so easy to forget :(

solarkennedy: What if I kept the explicit cache-disabled behavior here, but moved this function to itest utils so no normal (production) code ever tried to use it? (and forget to disable the cache for this one function)

huadongliu: Moving this function to itest would work.

FYI, itest run_setup_marathon_job calls setup_marathon_job.main(), where requests_cache starts to play.

solarkennedy:
@huadongliu  one more review?

huadongliu:
shipit

PR #740, opened 2016-09-08
[WIP] added api endpoint for interfacing with the autoscaler

mjksmith:
doggo


paasta_tools/api/api_docs/swagger.json
```diff
@@ -96,6 +96,87 @@
...
+            "description": "Instance name",
+            "name": "instance",
+            "required": true,
+            "type": "string"
+          }
+        ]
+      }
+    },
+    "/services/{service}/{instance}/autoscaler/{desired_instances}": {
+      "put": {
```
solarkennedy: This says put but the route says POST

paasta_tools/api/views/autoscaler.py
```diff
@@ -0,0 +1,38 @@
...
+    desired_instances = get_instances_from_zookeeper(service, instance)
+    return desired_instances
+
+
+@view_config(route_name='service.autoscaler.post', request_method='POST', renderer='json')
+def update_autoscaler_count(request):
+    service = request.swagger_data.get('service')
+    instance = request.swagger_data.get('instance')
+    desired_instances = request.swagger_data.get('desired_instances')
+    set_instances_for_marathon_service(service, instance, desired_instances)
```
solarkennedy: can you use kwargs everywhere?

paasta_tools/api/api.py
```diff
@@ -65,6 +65,8 @@ def make_app():
     config.include('pyramid_swagger')
     config.add_route('service.instance.status', '/v1/services/{service}/{instance}/status')
     config.add_route('service.list', '/v1/services/{service}')
+    config.add_route('service.autoscaler.get', '/v1/services/{service}/{instance}/autoscaler')
+    config.add_route('service.autoscaler.post', '/v1/services/{service}/{instance}/autoscaler/{desired_instances}')
```
Rob-Johnson: I think I'd expect this to consume the instance count in the body, rather than as a path in the URL?

so
```
curl -X POST -d '{"count": "1"}' paasta.api.com/v1/services/myservice/myinstance/autoscaler
```

(that might not be quite correct, but you get what I mean)

If I'm reading this correctly, you'll be POSTing to ``/services/myservice/myinstsance/autoscaler/3`` and no body?

solarkennedy:
Ship. Next steps include updating the docs after confirming it works correctly and stuff.

mjksmith:
@solarkennedy I'm not so sure about shipping because I haven't been able to successfully test this with the playground.

solarkennedy:
How about an itest then?

paasta_tools/api/api_docs/swagger.json
```diff
@@ -96,6 +96,89 @@
...
+          "404": {
+            "description": "Deployment key not found"
+          },
+          "500": {
+            "description": "Instance failure"
+          }
+        },
+        "summary": "Get status of service_name.instance_name",
+        "operationId": "get_autoscaler_count",
+        "tags": ["autoscaler"],
```
huadongliu: You may want to add the autoscaler tag to https://github.com/Yelp/paasta/pull/740/files#diff-529d3a317d236811b355ae5787b6657cL12.

PR #741, opened 2016-09-09
Fix potential paasta_metastatus DivizionByZero

nhandler:
Due to #730, in paasta metastatus -vvv, it is possible to get a DivizionByZero error if the host has all resources reserved, giving it a total of 0 resources.


mjksmith:
Ship it!

solarkennedy:
ship if the tests pass

PR #743, opened 2016-09-09
Small itest fixes for marathon itests

solarkennedy:
Hopefully making itests slightly faster.


nhandler:
Ship It if tests pass

Rob-Johnson:
shipit

solarkennedy:
Hah, it is so fast that these failing itests assume that by the time it shows up in marathon it is *healthy*!
I think I would rather make these tests relax their assertions and just by happy that a task is running in marathon at all, even if it isn't considered healthy yet?

solarkennedy:
This failing test is bogus and only ever worked because of time.
I needs to be fixed in a different way.

PR #744, opened 2016-09-09
Wait for deployment option

mattmb:
Still need to finish the unit tests and take a look at the itests.

However, this mostly seems to work so let me know if you think this is the right way to do this. I'm thinking that I'll change the internal code to generate the paasta pipelines once this has shipped.

I'm thinking that having `--wait-for-deployment` on should actually be the default but we can allow it to be disabled in the `deploy.yaml`. Something like:
```
- step: prod.main
  wait_for_deployment: False
```

And set the timeout, something like:
```
- step: prod.main
  wait_for_deployment: True
  deployment_timeout: 600
```

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -31,7 +31,7 @@ from paasta_tools.utils import get_paasta_tag_from_deploy_group
 from paasta_tools.utils import Timeout
 from paasta_tools.utils import TimeoutError
 
-DEFAULT_DEPLOYMENT_TIMEOUT = 1200  # seconds
+DEFAULT_DEPLOYMENT_TIMEOUT = 3600  # seconds
 
 
 def add_subparser(subparsers):
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -15,12 +15,20 @@
 """Contains methods used by the paasta client to mark a docker image for
 deployment to a cluster.instance.
 """
+import sys
+import time
+
 from paasta_tools import remote_git
+from paasta_tools.api import client
 from paasta_tools.cli.utils import validate_service_name
 from paasta_tools.utils import _log
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import format_tag
 from paasta_tools.utils import get_paasta_tag_from_deploy_group
+from paasta_tools.utils import Timeout
+from paasta_tools.utils import TimeoutError
+
+DEFAULT_DEPLOYMENT_TIMEOUT = 1200  # seconds
 
 
 def add_subparser(subparsers):
@@ -59,6 +67,21 @@ def add_subparser(subparsers):
         help='Name of the service which you wish to mark for deployment. Leading '
         '"services-" will be stripped.',
         required=True,
+    )
+    list_parser.add_argument(
+        '--wait-for-deployment',
+        help='Set to poll paasta and wait for the deployment to finish, '
+             'the default strategy is to mark for deployment and exit straightaway',
+        dest='block',
+        action='store_true',
+        default=False
+    )
+    list_parser.add_argument(
+        '-t', '--timeout',
+        dest="timeout",
+        type=int,
+        default=DEFAULT_DEPLOYMENT_TIMEOUT,
+        help="Time in seconds to wait for paasta to deploy the service. If the timeout is exceeded we return 1",
     )
     list_parser.add_argument(
         '-d', '--soa-dir',
@@ -107,9 +130,74 @@ def paasta_mark_for_deployment(args):
     if service and service.startswith('services-'):
         service = service.split('services-', 1)[1]
     validate_service_name(service, soa_dir=args.soa_dir)
-    return mark_for_deployment(
+    ret = mark_for_deployment(
         git_url=args.git_url,
         deploy_group=deploy_group,
         service=service,
         commit=args.commit,
     )
+    if args.block:
+        try:
+            _log(
+                service=service,
+                line="Waiting for deployment of {0} to {1} complete".format(args.commit, deploy_group),
+                component='deploy',
+                level='event'
+            )
+            wait_for_deployment(args.service, args.deploy_group,
+                                args.commit, args.soa_dir, args.timeout)
+            _log(
+                service=service,
+                line="Deployment of {0} to {1} complete".format(args.commit, deploy_group),
+                component='deploy',
+                level='event'
+            )
+        except TimeoutError:
+            sys.exit(1)
+    return ret
+
+
+def is_instance_deployed(service, instance, git_sha):
+    api = client.get_paasta_api_client()
+    status = api.service.status_instance(service=service, instance=instance).result()
+    # if it's a chronos service etc then skip waiting for it to deploy
+    if not status.marathon:
+        return True
+    return git_sha.startswith(status.git_sha) and \
+        status.marathon.app_count == 1 and \
+        status.marathon.deploy_status == 'Running' and \
+        status.marathon.expected_instance_count == status.marathon.running_instance_count
+
+
+def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
+    api = client.get_paasta_api_client()
+    instances = api.service.list_instances(service=service,
+                                           deploy_group=deploy_group).result()['instances']
+    if not instances:
+        _log(
+            service=service,
+            line="Couldn't find any instances for service {0} in deploy group {1}".format(service, deploy_group),
+            component='deploy',
+            level='event'
+        )
+        raise NoInstancesFound
+    try:
+        with Timeout(seconds=timeout):
+            while True:
+                if all([is_instance_deployed(service, instance, git_sha) for instance in instances]):
+                    break
+                time.sleep(10)
+    except TimeoutError:
+        _log(
+            service=service,
+            line="Timed out after {0} seconds, waiting for {1} in {2} to be deployed by PaaSTA. "
+                 "Try running 'paasta status -s {2} -v' to determine the cause. If the service is slow "
+                 "to start you may wish to increase the timeout".format(timeout, deploy_group, service),
+            component='deploy',
+            level='event'
+        )
+        raise
+
+
+class NoInstancesFound(Exception):
+    pass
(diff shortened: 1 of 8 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -15,12 +15,20 @@
...
+from paasta_tools.api import client
 from paasta_tools.cli.utils import validate_service_name
 from paasta_tools.utils import _log
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import format_tag
 from paasta_tools.utils import get_paasta_tag_from_deploy_group
+from paasta_tools.utils import Timeout
+from paasta_tools.utils import TimeoutError
+
+DEFAULT_DEPLOYMENT_TIMEOUT = 1200  # seconds
```
solarkennedy: This is too fast. We currently can't meet this SLA. How about an hour?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -107,9 +130,74 @@ def paasta_mark_for_deployment(args):
...
+                component='deploy',
+                level='event'
+            )
+        except TimeoutError:
+            sys.exit(1)
+    return ret
+
+
+def is_instance_deployed(service, instance, git_sha):
+    api = client.get_paasta_api_client()
```
solarkennedy: This doesn't seem to have a "cluster" associated with it. How will it know what cluster to contact?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -107,9 +130,74 @@ def paasta_mark_for_deployment(args):
...
+        return True
+    return git_sha.startswith(status.git_sha) and \
+        status.marathon.app_count == 1 and \
+        status.marathon.deploy_status == 'Running' and \
+        status.marathon.expected_instance_count == status.marathon.running_instance_count
+
+
+def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
+    api = client.get_paasta_api_client()
+    instances = api.service.list_instances(service=service,
```
solarkennedy: This needs to be cluster-specific somehow right? because "main" may not be deployed in all the different clusters.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -107,9 +130,74 @@ def paasta_mark_for_deployment(args):
...
     )
+    if args.block:
+        try:
+            _log(
+                service=service,
+                line="Waiting for deployment of {0} to {1} complete".format(args.commit, deploy_group),
+                component='deploy',
+                level='event'
+            )
+            wait_for_deployment(args.service, args.deploy_group,
```
solarkennedy: I'm concerned that this won't really reflect what authors really want. Think about yelp-main with potentially 1k instances across multiple clusters.

I think we are going to have to separate "what we deploy to" (deploy group) from "what we wait on" (clusters/instances?)

"Of course" people probably don't want to wait on batch jobs, do they only want to wait on the "main" stuff? Do they care about their deploy in testopia? (even though their deploy group is "everywhere")

paasta_tools/cli/cmds/mark_for_deployment.py
mattmb: D'oh, that's a very good point. I guess I'll need to pull the list of clusters first and then hit the API endpoint for each one...

paasta_tools/cli/cmds/mark_for_deployment.py
mattmb: Hmm. I think we should discus this one in a standup. To me it sounds like deploy groups actually already solve this (why would you want to separate)? For the yelp main example you could have:
```
- step: prod.main
  wait_for_deployment: True
- step: some_sort_of_notification
  command: make send_notification
- step: prod.everything_else
  wait_for_deployment: False
```

mattmb:
I think this is basically ready for more review. I couldn't quite work out how I'd do an itest for this. I guess I'd have to do a deployment of one version of a service, then deploy a new version and then run the `mark-for-deployment --wait-for-deployment` step? It'd all be a bit in the wrong order.

paasta_tools/cli/cmds/mark_for_deployment.py
solarkennedy: please kwargs all the things here

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -107,9 +133,103 @@ def paasta_mark_for_deployment(args):
...
+                                args.commit, args.soa_dir, args.timeout)
+            _log(
+                service=service,
+                line="Deployment of {0} to {1} complete".format(args.commit, deploy_group),
+                component='deploy',
+                level='event'
+            )
+        except TimeoutError:
```
solarkennedy: Before exiting 1 I would like to make this as friendly as possible. Something like:

```
mark-for-deployment timed out after %s (human_time).

This probably means that the deploy hasn't succeeded. The new service might not be healthy or a cluster is having issues. Here is current status of the deploy right now:

print are_instance_deployed(for each cluster)

Things you can try to debug this:

  paasta status -vv -s $service
  paasta logs -s $service

$sha is still marked to be the desired deployment. If you would like to roll back, you can run:

    paasta rollback --service $service --deploy-group $deploy-group

```

Basically spoonfeed as much as we can to them for figuring out what this means, besides exit 1. We really can provide a lot without doing much more work.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -107,9 +133,103 @@ def paasta_mark_for_deployment(args):
...
+    statuses = []
+    for instance in instances:
+        try:
+            statuses.append(api.service.status_instance(service=service, instance=instance).result())
+        except HTTPError as e:
+            if e.response.status_code == 404:
+                line = "Can't get status for instance {0}, service {1} in cluster {2}. "\
+                       "This is normally because it is a new service that hasn't been "\
+                       "deployed by PaaSTA yet".format(instance, service, cluster)
+                _log(
```
solarkennedy: This is going to be pretty noisy. We are talking quite a few lines every 10 seconds. I think it would be ok to just "print" this stuff

solarkennedy:
I say ship and iterate!

paasta_tools/cli/cmds/mark_for_deployment.py
mattmb: Nice 👍  I've actually mentioned something in when I catch the exception one level down but there's some good advice in their that I'll definitely add.

One thing I'm unsure about is the logging. At the moment I'm using `_log` which goes to paasta logs if I understand correctly. However, I guess it won't be shown in the Jenkins output. Ideally I'd like to see something in both.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -61,12 +76,34 @@ def add_subparser(subparsers):
...
+             'the default strategy is to mark for deployment and exit straightaway',
+        dest='block',
+        action='store_true',
+        default=False
+    )
+    list_parser.add_argument(
+        '-t', '--timeout',
+        dest="timeout",
+        type=int,
+        default=DEFAULT_DEPLOYMENT_TIMEOUT,
```
solarkennedy: You only really use this constant here, I think it would be ok to not have the indirection if we don't need it

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -102,14 +139,132 @@ def mark_for_deployment(git_url, deploy_group, service, commit):
...
+    ret = mark_for_deployment(
         git_url=args.git_url,
         deploy_group=deploy_group,
         service=service,
         commit=args.commit,
     )
+    if args.block:
+        try:
+            line = "Waiting for deployment of {0} to {1} complete".format(args.commit, deploy_group)
+            _log(
```
solarkennedy: I wouldn't "_log" this, especially not at the "event" level. It isn't and event.

I would just let log.info take care of it.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -102,14 +139,132 @@ def mark_for_deployment(git_url, deploy_group, service, commit):
...
+                                soa_dir=args.soa_dir,
+                                timeout=args.timeout)
+            line = "Deployment of {0} to {1} complete".format(args.commit, deploy_group)
+            _log(
+                service=service,
+                component='deploy',
+                line=line,
+                level='event'
+            )
+            log.info(line)
```
solarkennedy: This one you "dont" have to log.info, because _log does go to stdout by default, and it is an event worth logging to paasta logs

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -102,14 +139,132 @@ def mark_for_deployment(git_url, deploy_group, service, commit):
...
+                    break
+                time.sleep(10)
+    except TimeoutError:
+        human_status = ["{0}: {1}".format(cluster, data['deployed']) for cluster, data in cluster_map.items()]
+        line = "\nCurrent deployment status of {0} per cluster:\n".format(deploy_group) + "\n".join(human_status)
+        _log(
+            service=service,
+            component='deploy',
+            line=line,
+            level='event'
```
solarkennedy: I wouldn't _log this, too noisy to print this every 10 seconds to the scribe log

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -102,14 +139,132 @@ def mark_for_deployment(git_url, deploy_group, service, commit):
...
+    if not api:
+        # Assume not deployed if we can't reach API
+        return False
+    statuses = []
+    for instance in instances:
+        try:
+            statuses.append(api.service.status_instance(service=service, instance=instance).result())
+        except HTTPError as e:
+            if e.response.status_code == 404:
+                log.warning("Can't get status for instance {0}, service {1} in cluster {2}. "
```
solarkennedy: This log is probably going to be too noisy too. I think it would ok to make this log.debug and not show it by default.

solarkennedy:
Ship and try it out manually and see how it goes.

The logging stuff is important to me because we need to balance verbosity during the "normal" deploy that happens often, with providing enough context to make it obvious "what went wrong" during a failed deploy.

This should be a bit more obvious when you actually run the thing and see what it looks like on a real service that is deployed with lots of instances and lots of clusters.

paasta_tools/cli/cmds/mark_for_deployment.py
mattmb: Huh, I guess this is what I was missing. I was testing with the `FileLogWriter` class (which doesn't write to stdout) because I don't have scribe in my dev environment.

Issue #711, opened 2016-08-26
tests/test_autoscaling_lib.py::test_set_spot_fleet_request_capacity takes 1 second


EvanKrall:
Generally, nothing else in unit tests takes any appreciable amount of time. I'm not sure why it's taking a second -- theres' a time.sleep(5) in the function under test, but time.sleep is mocked out. Perhaps something is trying to make a TCP connection and that's timing out?

```
yelp-krall:~/eat24/paasta krall$ tox -- --durations=30
[...]

PR #746, opened 2016-09-12
Fix test pausing for 1 sec

mattmb:
@EvanKrall, I finally got round to fixing #711 because I was nearly about to reimplement the same slow test somewhere else!

This test was pausing because despite mocking time.sleep the timeout
excception was not raising until the 1 second had elapsed. This commit
makes time.sleep fire the timeout exception instead.

### DIFF
--- a/tests/test_autoscaling_lib.py
+++ b/tests/test_autoscaling_lib.py
@@ -25,6 +25,7 @@ from paasta_tools import marathon_tools
 from paasta_tools.mesos_tools import SlaveTaskCount
 from paasta_tools.paasta_metastatus import ResourceInfo
 from paasta_tools.utils import NoDeploymentsAvailable
+from paasta_tools.utils import TimeoutError
 
 
 def test_get_zookeeper_instances():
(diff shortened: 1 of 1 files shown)


tests/test_autoscaling_lib.py
```diff
@@ -922,10 +923,11 @@ def test_set_spot_fleet_request_capacity():
         mock.patch('paasta_tools.autoscaling_lib.AWS_SPOT_MODIFY_TIMEOUT', autospec=True)
```
solarkennedy: Can you rename this test to be something like "set_sfr_times_out" or something? to make it more obvious what this is testing

solarkennedy:
ship

EvanKrall:
Thanks :) ship

tests/test_autoscaling_lib.py
mattmb: It is testing a bit more than just the timeout. Also I'm a fan of the `test_<method_name>` format. Let me know if you feel strongly but otherwise I'll leave it as is.

PR #747, opened 2016-09-12
added bounce log parsing script

mjksmith:
What I used to get bounce log data for #736 

It's pretty nasty


paasta_tools/contrib/bounce_log_parse.py
```diff
@@ -0,0 +1,45 @@
+import json
```
solarkennedy: add #!/usr/bin/env python and be sure it is chmod +x

solarkennedy:
how about bounce_log_latency_parser.py ?

PR #748, opened 2016-09-13
import mesos_cli as a subpackage in paasta_tools

Rob-Johnson:
rather than forking, let's bring this into the paasta
project. This brings in everything that was needed for
the itests to pass - we'll save clearing out the bits we don't
need for the future.

Note that I've had to change [this](https://github.com/Yelp/paasta/compare/vendor-mesos-cli?expand=1#diff-b32f029cd959a6adcdcdb50bbdf326d2R43). I think I need to update the license at the top of the file to [make a note of what's changed?](https://tldrlegal.com/license/apache-license-2.0-(apache-2.0))

itests pass locally for me. I'll iterate from here to try and pick out the bits that we actually need (but follow up with separate PRs)

### DIFF
--- a/paasta_tools/mesos/zookeeper.py
+++ b/paasta_tools/mesos/zookeeper.py
@@ -1,0 +1,46 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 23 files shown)


Rob-Johnson:
the itests are failing because of all the lambdas. I guess I'll fix them

Rob-Johnson:
please pay attention to the [second commit](https://github.com/Yelp/paasta/pull/748/commits/64e4b21d24379b56af82127db41d49732c707991) when reviewing - I'd like reassurance that the functionality is the same.

paasta_tools/mesos/zookeeper.py
```diff
@@ -0,0 +1,46 @@
...
+# Helper for testing
+client_class = kazoo.client.KazooClient
+
+
+@contextlib.contextmanager
+def client(*args, **kwargs):
+    zk = client_class(*args, **kwargs)
+    try:
+        zk.start(timeout=TIMEOUT)
+    except kazoo.handlers.threading.KazooTimeoutError:
```
Rob-Johnson: Note that this is the fix for https://github.com/Yelp/paasta/issues/353

solarkennedy:
Ship! This isn't so bad.

PR #750, opened 2016-09-13
Add is_safe_to_drain logic to protect healthy instances

solarkennedy:
I would like to get this code out for early review.

No, there are not tests, it is mostly me experimenting with all the different smartstack replication-related stuff and putting them together in a "slave-centric" way.

I tried to reuse as many of the existing functions I could to extract the data I need.

### DIFF
--- a/paasta_tools/paasta_maintenance.py
+++ b/paasta_tools/paasta_maintenance.py
@@ -15,11 +15,21 @@
 import argparse
 import logging
 import sys
+import traceback
 from socket import getfqdn
+from socket import gethostbyname
 from socket import gethostname
 
 from paasta_tools import mesos_maintenance
-
+from paasta_tools import utils
+from paasta_tools.check_marathon_services_replication import load_smartstack_info_for_service
+from paasta_tools.marathon_tools import get_expected_instance_count_for_namespace
+from paasta_tools.marathon_tools import marathon_services_running_here
+from paasta_tools.marathon_tools import read_namespace_for_service_instance
+from paasta_tools.monitoring.replication_utils import backend_is_up
+from paasta_tools.monitoring.replication_utils import get_replication_for_services
+from paasta_tools.monitoring.replication_utils import ip_port_hostname_from_svname
+from paasta_tools.smartstack_tools import get_backends
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


paasta_tools/paasta_maintenance.py
```diff
@@ -104,10 +114,71 @@ def is_safe_to_drain(hostname):
...
+        for service, instance, port in local_services:
+            log.info("Inspecting %s.%s on %s" % (service, instance, port))
+            if is_healthy_in_synapse(port, local_backends) and \
+               synapse_replication_is_low(service, instance, system_paasta_config, local_backends=local_backends):
+                log.warning("%s.%s on port %s is healthy but the service is in danger!" % (
+                    service, instance, port
+                ))
+                return True
+        return False
+    except Exception:
```
Rob-Johnson: can you be more specific about what it is that you're try/catching?

paasta_tools/paasta_maintenance.py
```diff
@@ -104,10 +114,71 @@ def is_safe_to_drain(hostname):
     return not are_local_tasks_in_danger()
 
 
-def are_local_tasks_in_danger():
+def is_healthy_in_synapse(local_port, backends):
```
Rob-Johnson: how about calling this ``is_healthy_in_haproxy`` ? I was kind of expecting this to be reading the synapse json file rather than the haproxy csv given its name.

paasta_tools/paasta_maintenance.py
```diff
@@ -104,10 +114,71 @@ def is_safe_to_drain(hostname):
...
+                log.debug("Found a healthy local backend: %s" % backend)
+                return True
+            else:
+                log.debug("Found a unhealthy local backend: %s" % backend)
+                return False
+    log.debug("Couldn't find any haproxy backend listening on %s" % local_port)
     return False
 
 
+def synapse_replication_is_low(service, instance, system_paasta_config, local_backends):
```
Rob-Johnson: seems like this and ``check_smartstack_replication_for_instance`` in ``check_marathon_services_replication.py`` share quite a lot of logic. could you maybe DRY them up?

paasta_tools/paasta_maintenance.py
```diff
@@ -15,11 +15,21 @@
...
 import sys
+import traceback
 from socket import getfqdn
+from socket import gethostbyname
 from socket import gethostname
 
 from paasta_tools import mesos_maintenance
-
+from paasta_tools import utils
+from paasta_tools.check_marathon_services_replication import load_smartstack_info_for_service
```
solarkennedy: Eh, let me do that on a second pass. That function and get_smartstack_replication_for_attribute and pretty closely related, so probably a whole slew of functions need to be moved into smartstack tools together.

paasta_tools/paasta_maintenance.py
solarkennedy: It is purposely wide. The reasoning here is that if anything goes wrong we will fail broken, and for this script we need to fail safe and swallow the exception and let the show go on.

I can try to enumerate all the exceptions that this might have, but in the end if something is wrong and this scribe blows up, then it means the box is in such a state that it needs to be drained anyway, and any potential healthy tasks that might be on this box need to be discounted.

paasta_tools/paasta_maintenance.py
solarkennedy: Done. We still use the word synapse a lot here. I'm not really sure where we should stop calling it synapse and instead just call it haproxy.

paasta_tools/paasta_maintenance.py
solarkennedy: I was well aware, but I couldn't figure out how to really bring them together.

The one in monitoring tools does do very similar things, but focuses more on reporting on every "location" and stuff, where as this script only cares about what is local.

They do some of the initial information gathering, but the actions (checking) they perform is different.

PR #751, opened 2016-09-14
fix the string interpolation for missing slave error

Rob-Johnson:


solarkennedy:
ship

solarkennedy:
I usually don't make the assertions on these verbages so strict, I usually do like:

```
assert "0 slaves available" in e.error_message
assert "for test_service.instance" in error_message
```

PR #752, opened 2016-09-15
remove unused check_mesos_resource_utilization.py script

Rob-Johnson:
Was this was the first iteration of metastatus? This doesn't appear to have been worked on since we open sourced, so I propose binning it.


solarkennedy:
whoa? ship.

PR #753, opened 2016-09-15
Fix mark for deployment service name

mattmb:
We need to pass the shortened service name i.e. without 'services-'


Rob-Johnson:
shipit

PR #756, opened 2016-09-15
Fix image for local-run also support a dict json output

bchess:
Reports the image correctly when running local-run. Used to be 'None'.

Also add a flag --json-dict that, when used in conjunction with --dry-run, changes the json output to be a JSON dict that's easier to parse instead of an unstructured list of strings.


bchess:
we can eventually make --json-dict the default behavior. but I didn't want to break existing output format for now.

solarkennedy:
I say shipit.

PR #757, opened 2016-09-16
Make setup_marathon_job dynamically reserve resources

nhandler:
Due to the way we implemented maintenance mode's "draining", it is possible for tasks to get rescheduled onto the at-risk/draining host. To help avoid this and speed up the draining process, we decided to use dynamic reservations. Now, when a host is marked as 'draining', we will create a dynamic reservation for all free resources on the box (preventing anything new from being scheduled there). We will continue to dynamically reserve all free resources each time we kill a task running on the box. Eventually, all resources on the host will be dynamically reserved. The reservations will be removed once the box is undrained.

This change also updates `paasta_metastatus` so that dynamically reserved resources count as `used` resources. This should help us catch issues where we dynamically reserve too many resources in a cluster and no longer have sufficient capacity for new things to run.

I have also filed https://issues.apache.org/jira/browse/MESOS-6175 to try and get some dynamic reservation metrics into the `/metrics/snapshot` endpoint to allow simplifying/optimizing the metastatus code.

Connects to #376

### DIFF
--- a/paasta_tools/mesos_maintenance.py
+++ b/paasta_tools/mesos_maintenance.py
@@ -16,6 +16,7 @@ import argparse
 import datetime
 import json
 import logging
+from collections import namedtuple
 from socket import getfqdn
 from socket import gethostbyname
 
@@ -31,6 +32,9 @@ from paasta_tools.mesos_tools import get_mesos_task_count_by_slave
 from paasta_tools.mesos_tools import MESOS_MASTER_PORT
 
 log = logging.getLogger(__name__)
+Hostname = namedtuple('Hostname', ['host', 'ip'])
+Credentials = namedtuple('Credentials', ['file', 'principal', 'secret'])
+Resource = namedtuple('Resource', ['name', 'amount'])
 
 
 def base_api():
@@ -44,7 +48,7 @@ def base_api():
         url = "http://%s:%d%s" % (leader, MESOS_MASTER_PORT, endpoint)
         timeout = 15
         s = Session()
-        s.auth = load_credentials()
+        s.auth = (get_principal(), get_secret())
         req = Request(method, url, **kwargs)
         prepared = s.prepare_request(req)
         try:
@@ -245,25 +249,64 @@ def build_start_maintenance_payload(hostnames):
     return get_machine_ids(hostnames)
 
 
+def hostnames_to_componenets(hostnames, resolve=False):
+    """Converts a list of 'host[|ip]' entries into namedtuples containing 'host' and 'ip' attributes,
+    optionally performing a DNS lookup to resolve the hostname into an IP address
+    :param hostnames: a list of hostnames where each hostname can be of the form 'host[|ip]'
+    :param resolve: boolean representing whether to lookup the IP address corresponding to the hostname via DNS
+    :returns: a namedtuple containing the hostname and IP components
+    """
+
+    components = []
+    for hostname in hostnames:
+        # This is to allow specifying a hostname as "hostname|ipaddress"
+        # to avoid querying DNS for the IP.
+        if '|' in hostname:
+            (host, ip) = hostname.split('|')
+            components.append(Hostname(host=host, ip=ip))
+        else:
+            ip = gethostbyname(hostname) if resolve else None
+            components.append(Hostname(host=hostname, ip=ip))
+    return components
+
+
 def get_machine_ids(hostnames):
     """Helper function to convert a list of hostnames into a JSON list of hostname/ip pairs.
     :param hostnames: a list of hostnames
     :returns: a dictionary representing the list of machines to bring up/down for maintenance
     """
     machine_ids = []
-    for hostname in hostnames:
-        machine_id = dict()
-        # This is to allow specifying a hostname as "hostname|ipaddress"
-        # to avoid querying DNS for the IP.
-        if '|' in hostname:
-            (host, ip) = hostname.split('|')
-            machine_id['hostname'] = host
-            machine_id['ip'] = ip
-        else:
-            machine_id['hostname'] = hostname
-            machine_id['ip'] = gethostbyname(hostname)
+    components = hostnames_to_componenets(hostnames, resolve=True)
+    for component in components:
+        machine_id = {
+            'hostname': component.host,
+            'ip': component.ip,
+        }
         machine_ids.append(machine_id)
     return machine_ids
+
+
+def build_reservation_payload(resources):
+    """Creates the JSON payload needed to dynamically (un)reserve resources in mesos.
+    :param resources: list of Resource named tuples specifying the name and amount of the resource to (un)reserve
+    :returns: a dictionary that can be sent to Mesos to (un)reserve resources
+    """
+    payload = []
+    for resource in resources:
+        payload.append(
+            {
+                "name": resource.name,
+                "type": "SCALAR",
+                "scalar": {
+                    "value": resource.amount,
+                },
+                "role": "maintenance",
+                "reservation": {
+                    "principal": str(get_principal()),
+                },
+            }
+        )
+    return payload
 
 
 def build_maintenance_schedule_payload(hostnames, start=None, duration=None, drain=True):
@@ -331,7 +374,100 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
         log.error("%s does not contain Mesos slave credentials in the expected format. "
                   "See http://mesos.apache.org/documentation/latest/authentication/ for details" % mesos_secrets)
         raise
-    return username, password
+    return Credentials(file=mesos_secrets, principal=username, secret=password)
+
+
+def get_principal(mesos_secrets='/nail/etc/mesos-slave-secret'):
+    """Helper function to get the principal from the mesos-slave credentials
+    :param mesos_secrets: optional argument specifying the path to the file containing the mesos-slave credentials
+    :returns: a string containing the principal/username
+    """
+    return load_credentials(mesos_secrets).principal
+
+
+def get_secret(mesos_secrets='/nail/etc/mesos-slave-secret'):
+    """Helper function to get the secret from the mesos-slave credentials
+    :param mesos_secrets: optional argument specifying the path to the file containing the mesos-slave credentials
+    :returns: a string containing the secret/password
+    """
+    return load_credentials(mesos_secrets).secret
+
+
+def reserve(slave_id, resources):
+    """Dynamically reserve resources in marathon to prevent tasks from using them.
+    :param slave_id: the id of the mesos slave
+    :param resources: list of Resource named tuples specifying the name and amount of the resource to (un)reserve
+    :returns: boolean where 0 represents success and 1 is a failure
+    """
+    log.info("Dynamically reserving resoures on %s: %s" % (slave_id, resources))
+    payload = {
+        'slaveId': slave_id,
+        'resources': str(build_reservation_payload(resources)).replace("'", '"').replace('+', '%20')
+    }
+    client_fn = reserve_api()
+    try:
+        reserve_output = client_fn(method="POST", endpoint="", data=payload).text
+    except HTTPError as e:
+        e.msg = "Error adding dynamic reservation. Got error: %s" % e.msg
+        raise
+    return reserve_output
+
+
+def unreserve(slave_id, resources):
+    """Dynamically unreserve resources in marathon to allow tasks to using them.
+    :param slave_id: the id of the mesos slave
+    :param resources: list of Resource named tuples specifying the name and amount of the resource to (un)reserve
+    :returns: boolean where 0 represents success and 1 is a failure
+    """
+    log.info("Dynamically unreserving resoures on %s: %s" % (slave_id, resources))
+    payload = {
+        'slaveId': slave_id,
+        'resources': str(build_reservation_payload(resources)).replace("'", '"').replace('+', '%20')
+    }
+    client_fn = unreserve_api()
+    try:
+        unreserve_output = client_fn(method="POST", endpoint="", data=payload).text
+    except HTTPError as e:
+        e.msg = "Error adding dynamic unreservation. Got error: %s" % e.msg
+        raise
+    return unreserve_output
+
+
+def reserve_all_resources(hostnames):
+    mesos_state = get_mesos_state_summary_from_leader()
+    hosts = []
+    components = hostnames_to_componenets(hostnames)
+    for component in components:
+        hosts.append(component.host)
+    for slave in mesos_state['slaves']:
+        if slave['hostname'] in hosts:
+            log.info("Reserving all resources on %s" % slave['hostname'])
+            slave_id = slave['id']
+            resources = []
+            for resource in ['disk', 'mem', 'cpus']:
+                free_resource = slave['resources'][resource] - slave['used_resources'][resource]
+                for role in slave['reserved_resources']:
+                    free_resource -= slave['reserved_resources'][role][resource]
+                resources.append(Resource(name=resource, amount=free_resource))
+            reserve(slave_id=slave_id, resources=resources)
+
+
+def unreserve_all_resources(hostnames):
+    mesos_state = get_mesos_state_summary_from_leader()
+    hosts = []
+    components = hostnames_to_componenets(hostnames)
+    for component in components:
+        hosts.append(component.host)
+    for slave in mesos_state['slaves']:
+        if slave['hostname'] in hosts:
+            log.info("Unreserving all resources on %s" % slave['hostname'])
+            slave_id = slave['id']
+            resources = []
+            for role in slave['reserved_resources']:
+                for resource in ['disk', 'mem', 'cpus']:
+                    reserved_resource = slave['reserved_resources'][role][resource]
+                    resources.append(Resource(name=resource, amount=reserved_resource))
+            unreserve(slave_id=slave_id, resources=resources)
 
 
 def drain(hostnames, start, duration):
--- a/paasta_tools/mesos_maintenance.py
+++ b/paasta_tools/mesos_maintenance.py
@@ -271,7 +271,7 @@ def build_start_maintenance_payload(hostnames):
     return get_machine_ids(hostnames)
 
 
-def hostnames_to_componenets(hostnames, resolve=False):
+def hostnames_to_components(hostnames, resolve=False):
     """Converts a list of 'host[|ip]' entries into namedtuples containing 'host' and 'ip' attributes,
     optionally performing a DNS lookup to resolve the hostname into an IP address
     :param hostnames: a list of hostnames where each hostname can be of the form 'host[|ip]'
@@ -298,7 +298,7 @@ def get_machine_ids(hostnames):
     :returns: a dictionary representing the list of machines to bring up/down for maintenance
     """
     machine_ids = []
-    components = hostnames_to_componenets(hostnames, resolve=True)
+    components = hostnames_to_components(hostnames, resolve=True)
     for component in components:
         machine_id = {
             'hostname': component.host,
@@ -458,7 +458,7 @@ def unreserve(slave_id, resources):
 def reserve_all_resources(hostnames):
     mesos_state = get_mesos_master().state_summary()
     hosts = []
-    components = hostnames_to_componenets(hostnames)
+    components = hostnames_to_components(hostnames)
     for component in components:
         hosts.append(component.host)
     for slave in mesos_state['slaves']:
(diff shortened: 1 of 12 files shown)


paasta_tools/mesos_maintenance.py
```diff
@@ -334,6 +379,118 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
...
+    """Dynamically reserve resources in marathon to prevent tasks from using them.
+    :param slave_id: the id of the mesos slave
+    :param resource: the mesos resource type we want to reserve (i.e. cpus, mem, gpus, etc)
+    :param amount: integer representing the amount of the above resource to reserve
+    :returns: boolean where 0 represents success and 1 is a failure
+    """
+    log.info("Dynamically reserving %f %s on %s" % (amount, resource, slave_id))
+    payload = {
+        'slaveId': slave_id,
+        'resources': str(build_reservation_payload(slave_id, resource, amount)).replace("'", '"').replace('+', '%20')
```
mattmb: Do we have to do the string replacing? I guess mesos is expecting JSON in a application/x-www-form-urlencoded POST which admittedly seems pretty nasty of it.

What about `json.dumps` and `urllib.quote`. Also shouldn't `+` be `%2B`?

paasta_tools/mesos_maintenance.py
```diff
@@ -334,6 +379,118 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
...
+        raise
+    return unreserve_output
+
+
+def reserve_all_resources(hostnames):
+    mesos_state = get_mesos_state_summary_from_leader()
+    hosts = []
+    for hostname in hostnames:
+        if '|' in hostname:
+            (host, _) = hostname.split('|')
```
mattmb: We're repeating this a few times now. I'd suggest splitting it into its own function.

paasta_tools/paasta_metastatus.py
```diff
@@ -204,7 +218,7 @@ def assert_disk_health(metrics, threshold=10):
         )
 
 
-def assert_tasks_running(metrics):
+def assert_tasks_running((metrics, mesos_state)):
```
mattmb: Accidental search/replace?

paasta_tools/paasta_metastatus.py
```diff
@@ -246,7 +260,7 @@ def assert_no_duplicate_frameworks(state):
     )
 
 
-def assert_slave_health(metrics):
+def assert_slave_health((metrics, mesos_state)):
```
mattmb: As above?

paasta_tools/setup_marathon_job.py
```diff
@@ -242,6 +243,11 @@ def log_bounce_action(line, level='debug'):
 
     kill_given_tasks(client=client, task_ids=[task.id for task in tasks_to_kill], scale=True)
 
+    for task in bounce_lib.flatten_tasks(old_app_at_risk_tasks):
+        if task in tasks_to_kill:
+            hostname = task.host
+            reserve_all_resources([hostname])
```
mattmb: I'm glad this is only a small addition here 👍  I guess it could be pulled into `kill_given_tasks` which might mean the reservation happens sooner after a task is killed? However, I think we should try it like this first because it reads more cleanly.

paasta_tools/mesos_maintenance.py
```diff
@@ -334,6 +379,118 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
...
+
+            mem = slave['resources']['mem'] - slave['used_resources']['mem']
+            for role in slave['reserved_resources']:
+                mem -= slave['reserved_resources'][role]['mem']
+            reserve(slave_id=slave_id, resource='mem', amount=mem)
+
+            cpus = slave['resources']['cpus'] - slave['used_resources']['cpus']
+            for role in slave['reserved_resources']:
+                cpus -= slave['reserved_resources'][role]['cpus']
+            reserve(slave_id=slave_id, resource='cpus', amount=cpus)
```
mattmb: I guess this could be in a loop if you feel like cutting a few LOC.

paasta_tools/paasta_metastatus.py
Rob-Johnson: no, these are because the ``run_healthchecks_with_param`` expects a single parameter, and these functions now need both the ``mesos_state`` and ``metrics`` params, so @nhandler is wrapping them in a tuple to keep obeying the 'single param' requirement.

This is actually just hinting that the abstraction here is no longer really useful. Long term, we'll get rid of it.

@nhandler my suggestion here would be to drop the usage of ``run_healthchecks_with_param`` from ``get_mesos_metrics_health`` completely. Instead, replace it's implementation with:

```
def get_mesos_metrics_health(mesos_metrics, mesos_state):
    """Perform healthchecks against mesos metrics.
    :param mesos_metrics: a dict exposing the mesos metrics described in
    https://mesos.apache.org/documentation/latest/monitoring/
    :returns: a list of HealthCheckResult tuples
    """
   return [func(mesos_metrics, mesos_state) for func in [
        assert_cpu_health,
        assert_memory_health,
        assert_disk_health,
        assert_tasks_running,
        assert_slave_health,
    ]
```

mattmb: Oh, completely missed the extra bracket!

paasta_tools/mesos_maintenance.py
```diff
@@ -266,6 +288,29 @@ def get_machine_ids(hostnames):
...
+    payload = [
+        {
+            "name": resource,
+            "type": "SCALAR",
+            "scalar": {
+                "value": amount,
+            },
+            "role": "maintenance",
+            "reservation": {
+                "principal": str(get_principal()),
```
Rob-Johnson: why're you having to cast to a string here? you explicitly say in ``get_principal`` that it returns a string? might not need to worry about this?

paasta_tools/mesos_maintenance.py
Rob-Johnson: you don't need to do this, requests will urlencode the body for you. just make the value of the ``resources`` key the actual array that ``build_reservation_payload`` returns - requests will take take of the requests

paasta_tools/mesos_maintenance.py
Rob-Johnson: +1, please move this into a function

paasta_tools/mesos_maintenance.py
```diff
@@ -334,6 +379,118 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
...
+            hosts.append(hostname)
+    for slave in mesos_state['slaves']:
+        if slave['hostname'] in hosts:
+            log.info("Reserving all resources on %s" % slave['hostname'])
+            slave_id = slave['id']
+
+            disk = slave['resources']['disk'] - slave['used_resources']['disk']
+            for role in slave['reserved_resources']:
+                disk -= slave['reserved_resources'][role]['disk']
+            reserve(slave_id=slave_id, resource='disk', amount=disk)
```
Rob-Johnson: can't we bundle each of these up into one request? the endpoint takes an array of resources to reserve, so we can can do all this in one request

paasta_tools/paasta_metastatus.py
```diff
@@ -601,7 +619,7 @@ def main():
         mesos_state=mesos_state,
     )
     metrics = get_mesos_stats()
-    mesos_metrics_status = get_mesos_metrics_health(mesos_metrics=metrics)
+    mesos_metrics_status = get_mesos_metrics_health(mesos_metrics=metrics, mesos_state=mesos_state)
```
Rob-Johnson: We don't have to hit the master state endpoint - we could hit the state endpoint for each slave. It'll mean more work here, as we'll have to send the request to *every* slave in the cluster, and accumulate the results, but it does mean that we can avoid hitting the expensive state endpoint on the master. What do you think?

paasta_tools/mesos_maintenance.py
Rob-Johnson: further, don't you need to catch the exceptions that might be raised from the HTTP request? we do a ``raise_for_response``, and the docs mention a number of HTTP response codes that might be returned

paasta_tools/mesos_maintenance.py
```diff
@@ -334,6 +379,118 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
...
+    for slave in mesos_state['slaves']:
+        if slave['hostname'] in hosts:
+            log.info("Unreserving all resources on %s" % slave['hostname'])
+            slave_id = slave['id']
+            for role in slave['reserved_resources']:
+                disk = slave['reserved_resources'][role]['disk']
+                unreserve(slave_id=slave_id, resource='disk', amount=disk)
+
+                mem = slave['reserved_resources'][role]['mem']
+                unreserve(slave_id=slave_id, resource='mem', amount=mem)
```
Rob-Johnson: same comments here r.e. bundling into 1 request + error handling

paasta_tools/mesos_maintenance.py
nhandler: @Rob-Johnson That is how I initially had the code. Apparently, while Mesos wants the data submitted as `application/x-www-form-urlencoded`, it parses it like JSON. As a result, it complains if you have single quotes instead of double quotes. `requests` also encodes whitespace as `+`, while Mesos is expecting `%20`.

@mattmb I can try messing around with `urllib.quote`. I hate these `replace()` calls just as much as both of you.

paasta_tools/mesos_maintenance.py
nhandler: This was me fighting against `MagicMocks`. However, I made a number of changes to the itest since then, so let me double check if it still passes without this cast.

paasta_tools/paasta_metastatus.py
nhandler: Personally, I think https://issues.apache.org/jira/browse/MESOS-6175 is going to be our best long-term solution. I'm also not sure how well hitting every slave will scale unless we do it in parallel. I also have a feeling we will get hit by timeouts and connectivity issues a bit more frequently, which might slow down this tool. My vote would be to save move this discussion to a separate issue/PR.

paasta_tools/paasta_metastatus.py
solarkennedy: I second robj's suggestion

paasta_tools/setup_marathon_job.py
solarkennedy: What will un_reserve these resources?

nhandler: The `undrain()` function will call `unreserve_all_resources()`. We will need to make sure things like the autoscaler and other tools remember to call `undrain()`.

paasta_tools/mesos_maintenance.py
nhandler: Yeah, the tests get unhappy without this cast due to the MagicMocks. I've tried fighting them a bit without much luck. I would be more than happy to remove this cast if someone knows of a way to keep the tests passing without it.

paasta_tools/mesos_maintenance.py
nhandler: It is because we do a `raise_for_status` in `base_api()` that this becomes a bit more annoying. Since any exception will be caught in this function, which is used by all of the API calls in `mesos_maintenance.py`, we can't easily put error handling specific to dynamic reservations in here.

I think the best approach would be to remove `resp.raise_for_status()` from `base_api()` and move it up two layers into all functions calling `master_api()` (`down`, `up`, `reserve_api`, `unreserve_api`, `maintenance_api`). This will make it possible to adjust how we deal with exceptions for each of the API calls.

Rob-Johnson: @nhandler I don't quite get what you mean. we want to catch exceptions thrown by the ``reserve`` call here, so I was just expecting something like

```
try:
  reserve(slave_id=slave_id, resource='disk', amount=disk)
except HTTPError as e:
   log.debug('couldn't reserve resources; bailing')
   raise
```

(or whatever action you think is the *right* thing to do). [we rereaise the exception after it is caught in](https://github.com/Yelp/paasta/blob/master/paasta_tools/mesos_maintenance.py#L59) ``base_api`` and again in ``reserve`` which I think is fine; you need to catch the HTTPError thrown and deal with it appropriately at each call site.

> This will make it possible to adjust how we deal with exceptions for each of the API calls.

This isn't really useful - the answer is mostly going to be log.debug and raise - the handling of an exception will be unique at each call site, not at the api call

paasta_tools/paasta_metastatus.py
Rob-Johnson: sure, but we can't really rely on MESOS-6175. I absolutely agree that we should do this in parallel - in fact, I deleted an unused function from mesos-cli that did this. We can always bring that back to life. Let's see how this goes - I have a feeling we're going to suffer from these calls to the master state.json - but we can ship this version for now

paasta_tools/mesos_maintenance.py
nhandler: Eventually, I think we will want to add some retry logic in the event of a `409 Conflict` (as that will likely be do to race conditions), but for now, I'll simply log and raise.

nhandler:
I think I've addressed all of the feedback that I have received. However, there were some ugly rebases, so please double check this PR carefully to make sure I didn't do anything stupid by mistake.

paasta_tools/mesos_maintenance.py
```diff
@@ -331,7 +396,110 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
...
+def reserve(slave_id, resources):
+    """Dynamically reserve resources in marathon to prevent tasks from using them.
+    :param slave_id: the id of the mesos slave
+    :param resources: list of Resource named tuples specifying the name and amount of the resource to (un)reserve
+    :returns: boolean where 0 represents success and 1 is a failure
+    """
+    log.info("Dynamically reserving resoures on %s: %s" % (slave_id, resources))
+    payload = {
+        'slaveId': slave_id,
+        'resources': str(build_reservation_payload(resources)).replace("'", '"').replace('+', '%20')
```
solarkennedy: https://docs.python.org/2/library/urllib.html#urllib.quote ?

paasta_tools/mesos_maintenance.py
```diff
@@ -331,7 +396,110 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
...
+def unreserve(slave_id, resources):
+    """Dynamically unreserve resources in marathon to allow tasks to using them.
+    :param slave_id: the id of the mesos slave
+    :param resources: list of Resource named tuples specifying the name and amount of the resource to (un)reserve
+    :returns: boolean where 0 represents success and 1 is a failure
+    """
+    log.info("Dynamically unreserving resoures on %s: %s" % (slave_id, resources))
+    payload = {
+        'slaveId': slave_id,
+        'resources': str(build_reservation_payload(resources)).replace("'", '"').replace('+', '%20')
```
solarkennedy: https://docs.python.org/2/library/urllib.html#urllib.quote ?

paasta_tools/setup_marathon_job.py
solarkennedy: I just wish it wasn't the bounce code's responsibility to do this.

How does this reservation even work if the bounce code hasn't killed these tasks yet?

Also I have a hunch that this needs some try/excepts. The bounce has to go on in the face of slow api endpoints and broken reservation stuff, in the context of reserving away resource for at-risk tasks. Heck, if it was easy I would say make this asynchronous and best-effort so the bounce can move on at normal speed!

nhandler: If the bounce code hasn't killed these tasks yet, it will simply reserve whatever resources are available on the host at the time (which is still better than what it currently does). The next time a task is killed from the host, it will try to reserve all the free resources again. There are other places we can add calls to `reserve_all_resources` (i.e. cron job running on the leader that calls `reserve_all_resources` on any host marked as `draining`), but I'd like to hold off on that for now to avoid hammering the REST endpoints too much and to see how this works in practice.

paasta_tools/mesos_maintenance.py
nhandler: Nope. The single quotes mess up the parsing and cause an `HTTPError: 400 Client Error: Bad Request`. In my opinion, this is a bug in Mesos itself.

paasta_tools/mesos_maintenance.py
```diff
@@ -331,7 +399,110 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
...
+        unreserve_output = client_fn(method="POST", endpoint="", data=payload).text
+    except HTTPError:
+        log.debug("Error adding dynamic unreservation.")
+        raise
+    return unreserve_output
+
+
+def reserve_all_resources(hostnames):
+    mesos_state = get_mesos_master().state_summary()
+    hosts = []
```
Rob-Johnson: hosts = [component.host for component in hostnames_to_components(hostnames)]

paasta_tools/mesos_maintenance.py
```diff
@@ -331,7 +399,110 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
...
+                log.debug("Failed reserving all of the resources on %s (%s). Aborting." % (hostname, slave_id))
+                raise
+
+
+def unreserve_all_resources(hostnames):
+    mesos_state = get_mesos_master().state_summary()
+    hosts = []
+    components = hostnames_to_components(hostnames)
+    for component in components:
+        hosts.append(component.host)
```
Rob-Johnson: same suggestion as I made in ``reserve_all_resources`` to this logic. The fact that you do this twice hints you should move this logic into its own function with a test

Rob-Johnson: the logic here being duplicated from ``reserve_all_resources`` hints that you should move it into its own function with a unit test

paasta_tools/mesos_maintenance.py
```diff
@@ -331,7 +399,110 @@ def load_credentials(mesos_secrets='/nail/etc/mesos-slave-secret'):
...
+    return unreserve_output
+
+
+def reserve_all_resources(hostnames):
+    mesos_state = get_mesos_master().state_summary()
+    hosts = []
+    components = hostnames_to_components(hostnames)
+    for component in components:
+        hosts.append(component.host)
+    for slave in mesos_state['slaves']:
```
Rob-Johnson: how about removing the inner ``if`` here by filtering those that you're interested in.

```
known_slaves = [slave for slave in mesos_state['slaves'] if slave['hostname'] in hostnames]
for slave in known_slaves:
    log.info("Reserving all resources on %s" % hostname))
    ...
```

paasta_tools/mesos_maintenance.py
```diff
@@ -342,12 +513,13 @@ def drain(hostnames, start, duration):
...
     log.info("Draining: %s" % hostnames)
+    reserve_all_resources(hostnames)
     payload = build_maintenance_schedule_payload(hostnames, start, duration, drain=True)
     client_fn = get_schedule_client()
     try:
         drain_output = client_fn(method="POST", endpoint="", data=json.dumps(payload)).text
-    except HTTPError as e:
-        e.msg = "Error performing maintenance drain. Got error: %s" % e.msg
+    except HTTPError:
+        log.debug("Error performing maintenance drain.")
```
Rob-Johnson: why've you made this change? now we lose the extra information in the error message unless we have logging in debug mode?

tests/test_mesos_maintenance.py
```diff
@@ -41,13 +43,18 @@
...
+from paasta_tools.mesos_maintenance import reserve
 from paasta_tools.mesos_maintenance import schedule
 from paasta_tools.mesos_maintenance import seconds_to_nanoseconds
 from paasta_tools.mesos_maintenance import status
 from paasta_tools.mesos_maintenance import undrain
+from paasta_tools.mesos_maintenance import unreserve
 from paasta_tools.mesos_maintenance import up
 
 
+Resource = namedtuple('Resource', ['name', 'amount'])
```
Rob-Johnson: why're you redeclaring this?

paasta_tools/mesos_maintenance.py
nhandler: Most (if not all) of these blocks were resulting in their own exceptions being thrown due to `e.msg` not existing. I could probably throw a new `HTTPError` with a more descriptive message rather than simply calling `raise`.

tests/test_mesos_maintenance.py
nhandler: Because I forgot that I could simply import it :)

paasta_tools/mesos_maintenance.py
Rob-Johnson: ha, right. yes please!

PR #758, opened 2016-09-16
cleanup patching of mesos modules in mesos_tools

Rob-Johnson:
more cleanup. This time, I've targeted the monkey patching we were doing in ``mesos_tools``. 

- I've removed the ``log.fatal`` calls from the mesos-cli modules, instead replacing them with the relevant exceptions (we were monkey patching log.fatal to get this behaviour before).
- Replaced the implementation of ``mesos.Master.slaves`` with the version we had in mesos-tools, wherein we exact match on the id, rather than partial matching
- removed the ``get_mesos_state_from_current_leader`` function in ``mesos_tools``. When you create a new ``mesos.Master`` object, it calculates the current leader in it's constructor and redirects all requests to that leader, so we don't need this extra level of checking.
- removed the ``get_mesos_state_summary_from_current_leader`` function from ``mesos_tools``, replacing it with a ``state_summary`` function on the ``mesos.Master`` class.
- completely removed the zk timeout catching - before we were just rethrowing another exception, for no real benefit.

### DIFF
--- a/paasta_tools/autoscaling_lib.py
+++ b/paasta_tools/autoscaling_lib.py
@@ -40,8 +40,7 @@ from paasta_tools.marathon_tools import MESOS_TASK_SPACER
 from paasta_tools.marathon_tools import set_instances_for_marathon_service
 from paasta_tools.mesos_maintenance import drain
 from paasta_tools.mesos_maintenance import undrain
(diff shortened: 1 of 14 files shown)


mattmb:
Nice stuff 👏 

Shipit

PR #759, opened 2016-09-16
Don't wait for stopped instances

mattmb:
When we mark a service for deploy we shouldn't wait for stopped instances because they shouldn't be running.


solarkennedy:
Seems legit. Is this just an optimization? Seems like the logic should have worked anyway right? (because 0==0, and it is "running" from a marathon perspective?, just wasteful to wait for an app with 0 tasks to bounce?)

mattmb:
Not just an optimization. Our status endpoint returns "Stopped" if there are 0 expected and running instances: https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_tools.py#L743-L744

PR #760, opened 2016-09-16
Split cluster and service autoscaler

mattmb:
This is just refactoring. It should be a noop in terms of functionality.
The motivation is that the two autoscalers share very little code or
functionality.


solarkennedy:
ship

PR #761, opened 2016-09-16
update mesos master to retry zk connection PAASTA-5245

Rob-Johnson:
this will mean that we retry the ZK connection if a node is down.

the ``max_tries`` number is 10 here - any ideas for a better magic number? 5 doesn't make too much sense, because it'll pick a random host in the cluster (so we can't guarantee that 5 retries == 5 unique hosts)


solarkennedy:
I don't have any better ideas. Ship it!

PR #762, opened 2016-09-19
Refactor SFR cluster autoscaler

mattmb:
Splitting up the very large scale_aws_spot_fleet_request function in to
component parts. Not intending to change any logic here, just make the
code and tests easier to read.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -318,7 +318,7 @@ def set_spot_fleet_request_capacity(sfr_id, capacity, dry_run, region=None):
                 time.sleep(5)
         except TimeoutError:
             log.error("Spot fleet {0} not in active state so we can't modify it.".format(sfr_id))
-            return False
+            raise FailSetSpotCapacity
     if dry_run:
         return True
     try:
@@ -326,7 +326,7 @@ def set_spot_fleet_request_capacity(sfr_id, capacity, dry_run, region=None):
                                                    ExcessCapacityTerminationPolicy='noTermination')
     except ClientError as e:
         log.error("Error modifying spot fleet request: {0}".format(e))
-        return False
+        raise FailSetSpotCapacity
     return ret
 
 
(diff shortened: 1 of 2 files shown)


mattmb:
I'll make sure to test this somewhere before we ship because we don't have good itest coverage here and I've re-written enough of the unit tests that it's possible I've changed the functionality.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -364,69 +364,97 @@ def scale_aws_spot_fleet_request(resource, current_capacity, target_capacity, po
...
-            filtered_slaves = get_mesos_task_count_by_slave(mesos_state, slaves_list=filtered_sorted_slaves)
+        downscale_spot_fleet_request(resource=resource,
+                                     filtered_slaves=filtered_slaves,
+                                     current_capacity=current_capacity,
+                                     target_capacity=target_capacity,
+                                     pool_settings=pool_settings,
+                                     dry_run=dry_run)
+
+
+def gracefully_terminate_slave(resource, slave_to_kill, pool_settings, current_capacity, new_capacity, dry_run):
```
solarkennedy: Maybe on the next pass this function can be split up into two pieces, one to gracefully drain the slave, and the other to terminate + scale the sfr?

The first half of the code deals with draining and the slave_to_kill, the second part of this function deals with new_cap, current_cap, etc.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -364,69 +364,97 @@ def scale_aws_spot_fleet_request(resource, current_capacity, target_capacity, po
...
+                                       pool_settings=pool_settings,
+                                       current_capacity=current_capacity,
+                                       new_capacity=new_capacity,
+                                       dry_run=dry_run)
+        except HTTPError:
+            # Something wrong draining host so try next host
+            continue
+        except FailSetSpotCapacity:
+            break
+        current_capacity = new_capacity
```
solarkennedy: It seems strange that these are run at the end of the loop?

solarkennedy:
I've got a review out for adding sensu monitoring to the autoscaler it self to make it easier to know when we got it wrong?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
mattmb: Yeah, I think that's possibly a good shout. At least we can see the wood through the trees again now. I can even understand my own unit tests again which is a good sign!

paasta_tools/autoscaling/autoscaling_cluster_lib.py
mattmb: I suppose it could be set within the `gracefully_terminate_slave` function but it seemed easier to be outside and I put it at the end to make sure it only got updated if we definitely succeeded in changing the capacity.

mattmb:
Yeah, that sounds like a nice alert to have. It'd be nice to be paged if it hasn't run in a certain amount of time (both cluster and service). Do you want to tag me in it, I can't find it on a casual search?

PR #772, opened 2016-09-20
Test main methods in autoscaling scripts

mattmb:
Kind of a noddy test but would have picked up some ImportErrors that I
introduced recently so better than 0 coverage.

@nhandler this covers your comment in https://github.com/Yelp/paasta/pull/763 itests are pretty difficult with the autoscaler(s) but this would have at least caught the ImportError.


solarkennedy:
sgtm!

PR #775, opened 2016-09-20
I **think** this will fix the pyramid autoscaling api

mjksmith:
TIL `view_config` has to match the route you set up, it can't just have the same `route_name`.


solarkennedy:
ship?

PR #776, opened 2016-09-20
Merge in replication_utils into smartstack_tools

solarkennedy:
@Rob-Johnson requested that I move some of the shared functionality between paasta_maintenance and check_marathon_services_replication.

Not a big surprise, this was not possible without introducing circular imports, so I had to merge in replication_utils along with those related functions to get it to stop.


Rob-Johnson:
shipit - ty for doing this

PR #777, opened 2016-09-21
PaaSTA docker_exec, docker_stop, sysdig, docker_inspect

mattmb:
I've been playing with this on the sly since I had the idea at mesoscon but thought I'd post it up to see if anyone's interested.

It's a bit hacky, but the idea is: you run `paasta sysdig -s my-service -i main -c my-cluster` and you get jumped over SSH to a random host that's running the service and `csysdig` is launched with filters to show the mesos tasks running for that instance on that host. You can then jump into the containers or process etc.

If others like the idea I'll tidy it up and add some tests but I wanted to get some opinions before I spend the time.

Some issues:
* We have to ssh first to a box in the cluster to get the marathon credentials.
* The marathon credentials are embedded in the URL that then ends up in the command line (so anyone else on the paasta node running `ps` could see them)
* I've seen some issues when sysdig hits the mesos api for master/state.json which I think are because it times out after 5s: https://github.com/draios/sysdig/blob/a1af66ceb7bbd132a5f7154a8071fb4ac43aea15/userspace/libsinsp/mesos_http.h#L27 We could probably ask for that to be increased or configurable.

### DIFF
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -24,10 +24,10 @@ from pyramid.view import view_config
 from paasta_tools import marathon_tools
 from paasta_tools.api import settings
 from paasta_tools.cli.cmds.status import get_actual_deployments
+from paasta_tools.mesos_tools import get_running_tasks_from_active_frameworks
 from paasta_tools.paasta_serviceinit import get_deployment_version
 from paasta_tools.utils import NoDockerImageError
 from paasta_tools.utils import validate_service_instance
-from paasta_tools.mesos_tools import get_running_tasks_from_active_frameworks
 
 
 log = logging.getLogger(__name__)
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -27,6 +27,7 @@ from paasta_tools.cli.cmds.status import get_actual_deployments
 from paasta_tools.paasta_serviceinit import get_deployment_version
 from paasta_tools.utils import NoDockerImageError
 from paasta_tools.utils import validate_service_instance
+from paasta_tools.mesos_tools import get_running_tasks_from_active_frameworks
 
 
 log = logging.getLogger(__name__)
@@ -46,6 +47,7 @@ def marathon_job_status(mstatus, client, job_config):
         return
 
     mstatus['app_id'] = app_id
+    mstatus['slaves'] = list({task.slave['hostname'] for task in get_running_tasks_from_active_frameworks(app_id)})
     mstatus['expected_instance_count'] = job_config.get_instances()
 
     deploy_status = marathon_tools.get_marathon_app_deploy_status(client, app_id)
--- a/paasta_tools/cli/cmds/docker_exec.py
+++ b/paasta_tools/cli/cmds/docker_exec.py
@@ -14,10 +14,10 @@
 # limitations under the License.
 import subprocess
 
-from paasta_tools.mesos_tools import get_container_name
-from paasta_tools.cli.utils import get_subparser
 from paasta_tools.cli.utils import get_status_for_instance
+from paasta_tools.cli.utils import get_subparser
 from paasta_tools.cli.utils import pick_slave_from_status
+from paasta_tools.mesos_tools import get_container_name
 
 
 def add_subparser(subparsers):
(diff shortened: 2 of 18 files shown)


mattmb:
So, I added some commits to make this more generic. And to add some new helper commands, e.g. `paasta docker_exec -s my-service -i main -c testcluster /bin/bash`

Still needs some tests! Does anyone have any comments on the questions from above? I'm going to want to merge this sooner rather than later now that the branch is growing.

paasta_tools/cli/cmds/docker_exec.py
```diff
@@ -0,0 +1,43 @@
...
+
+def add_subparser(subparsers):
+    new_parser = get_subparser(description="'paasta docker_exec' works by picking a container running your service "
+                                           "at random. It then runs docker exec -ti <container_id> <commands> "
+                                           "where commands are those that you specify",
+                               help_text="Docker exec against a container running your service",
+                               command='docker_exec',
+                               function=paasta_docker_exec,
+                               subparsers=subparsers)
+    new_parser.add_argument('exec_command',
```
solarkennedy: Make this default to bash

solarkennedy:
Heck, I say ship! I don't mind that this ends up using ssh in the end, without something insane I think this is just fine.

What do you think about making python "exec" the ssh process so that it is the only thing running?

mattmb:
@solarkennedy so I did spend some time trying to `subprocess.Popen` which in theory is the non-blocking version of `subprocess.call`. However, I couldn't make the redirection of the output work nicely (for csysdig at least). I put this down to some weirdness of `Popen`. However, `call` seems to work nicely, even thought I thought it was just a thin wrapper on `Popen` 😕

paasta_tools/api/views/instance.py
```diff
@@ -46,6 +47,7 @@ def marathon_job_status(mstatus, client, job_config):
         return
 
     mstatus['app_id'] = app_id
+    mstatus['slaves'] = list(set([task.slave['hostname'] for task in get_running_tasks_from_active_frameworks(app_id)]))
```
mjksmith: Do a set comprehension instead of casting a list comprehension to a set

paasta_tools/cli/cmds/sysdig.py
```diff
@@ -0,0 +1,89 @@
...
+        help="Run the script here rather than SSHing to a PaaSTA master",
+        default=False,
+        action='store_true'
+    )
+
+
+def get_any_mesos_master(cluster):
+    system_paasta_config = load_system_paasta_config()
+    masters, output = calculate_remote_masters(cluster, system_paasta_config)
+    if masters == []:
```
mjksmith: I'd use `if not masters:` here, which won't break if `calculate_remote_masters ` ever changes to return a set, tuple, generator, etc.

(Also it's the PEP8 approved way of checking if sequences are empty).

paasta_tools/mesos_tools.py
```diff
@@ -653,3 +653,23 @@ def terminate_framework(framework_id):
     resp = requests.post('http://%s:%d/master/teardown' % (get_mesos_leader(), MESOS_MASTER_PORT),
                          data={"frameworkId": framework_id})
     resp.raise_for_status()
+
+
+def get_container_name(app_id, slave_hostname=None, task_id=None):
+    tasks = get_running_tasks_from_active_frameworks(app_id)
+    try:
+        if task_id:
+            task = [task for task in tasks if task['id'] == task_id][0]
```
mjksmith: `task = next(ask for task in tasks if task['id'] == task_id)` short-circuits :)

paasta_tools/mesos_tools.py
```diff
@@ -653,3 +653,23 @@ def terminate_framework(framework_id):
...
+        elif slave_hostname:
+            task = [task for task in tasks if slave_hostname in task.slave['hostname']][0]
```
mjksmith: `task = next(task for task in tasks if slave_hostname in task.slave['hostname'])`

paasta_tools/mesos_tools.py
```diff
@@ -653,3 +653,23 @@ def terminate_framework(framework_id):
...
+        else:
+            raise TaskNotFound("Couldn't find task")
+    except IndexError:
```
mjksmith: If you decide to use `next` you'll need to change this to `except StopIteration`.

paasta_tools/mesos_tools.py
mattmb: Nice, I like that, reads more cleanly than `[0]` 👍

mattmb:
Just the xenial build failing, which is already failing on master.

PR #779, opened 2016-09-21
Paasta local run itest timeout

mattmb:
This seems to be failing intermittently? I'm just guessing but maybe
this is now taking longer after #756. I've bumped up the timeouts to see
if it helps.


solarkennedy:
I haven't seen this myself, but I say ship

PR #780, opened 2016-09-21
Remove emergency-scale

solarkennedy:
emergency-scale was useful back when it required a "bounce" to scale up or down.

But now that the bounce code works so quickly, emergency-scale efforts are almost instantly thwarted.

I think this code has no place anymore, it would be faster for people to edit soa-configs directly to scale.


Rob-Johnson:
shipit. I'm enjoying how much code we've deleted recently.

PR #786, opened 2016-09-23
remove @CachedProperty decorator from master + slave state

Rob-Johnson:
For two reasons:
- it makes the interface to functions in the api
inconsistent (property based vs function based)
- it *forces* caching on the user of the api.


mattmb:
I guess we should ship this carefully. Are we very sure that the requests cache is in place to pick up any increased load on the state endpoint?

PR #789, opened 2016-09-23
Make paasta emergency-restart simply kill inflight tasks

solarkennedy:
IRL, the actual effect of these emergency commands is what normal users would call a "restart", because old tasks die and new tasks get launched by the bounce.

Therefore, that is what I've called the new command, as it "restarts" tasks ungracefully.

In practice, the implementation is to scale down the marathon app to 0 and let the bounce bring it back up.


Rob-Johnson:
fix the tests & shipit.

PR #790, opened 2016-09-23
Use proper version of mesos in trusty dockerfile

nhandler:
Looks like we forgot to bump this version during the mesos upgrade.


solarkennedy:
ship

PR #793, opened 2016-09-26
Configure the API for development mode

mattmb:
This allows us to run the API in a development mode with sensible
logging and auto reloading on code changes.

This shouldn't change how things are working if you call the `paasta-api` script but should allow us to use `pserve`

@mjksmith this solves our internal ticket PAASTA-6599


solarkennedy:
ship

PR #794, opened 2016-09-26
Fix cluster autoscaling termination loop

mattmb:
We were rounding the new target capacity in the main downscaling loop.
This leads to a situation where terminating an instance with a weight
less than 0.5 leaves the new target capacity at the same value. This
leads to terminating more boxes than intended!

Now we only round the value to an integer just before we send it to AWS
which is the only reason we have to round it anyway.


Rob-Johnson:
shipit

PR #796, opened 2016-09-27
Add some new functions to make it easier to monitor draining/down hosts

nhandler:
These new functions will make it easier to add monitoring checks to detect hosts that are forgotten in a draining/down state.

### DIFF
--- a/paasta_tools/mesos_maintenance.py
+++ b/paasta_tools/mesos_maintenance.py
@@ -200,6 +200,49 @@ def is_host_down(hostname=getfqdn()):
     return hostname in get_down_hosts()
 
 
+def get_hosts_forgotten_draining():
+    """Find hosts that are still marked as draining (rather than down) after the start
+    of their maintenance window. For each of these hosts, either mark them as 'down' for
+    maintenance, adjust the time of their maintenance window, or cancel the maintenance
+    window if it is no longer needed.
+    :returns: a list of hostnames of hosts forgotten draining
+    """
+    draining_hosts = get_draining_hosts()
+    hosts_past_maintenance_start = get_hosts_past_maintenance_start()
+    print("draining_hosts: %s" % draining_hosts)
+    print("hosts_past_maintenance_start: %s" % hosts_past_maintenance_start)
+    forgotten_draining = list(set(draining_hosts).intersection(hosts_past_maintenance_start))
+    return forgotten_draining
+
+
+def are_hosts_forgotten_draining():
+    """Quick way to test if there are any forgotten draining hosts.
+    :returns: a boolean that is True if there are any forgotten draining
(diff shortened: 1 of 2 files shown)


paasta_tools/mesos_maintenance.py
```diff
@@ -200,6 +200,49 @@ def is_host_down(hostname=getfqdn()):
...
+def get_hosts_forgotten_draining():
+    """Find hosts that are still marked as draining (rather than down) after the start
+    of their maintenance window. For each of these hosts, either mark them as 'down' for
+    maintenance, adjust the time of their maintenance window, or cancel the maintenance
+    window if it is no longer needed.
+    :returns: a list of hostnames of hosts forgotten draining
+    """
+    draining_hosts = get_draining_hosts()
+    hosts_past_maintenance_start = get_hosts_past_maintenance_start()
+    print("draining_hosts: %s" % draining_hosts)
```
mattmb: Can we make these `log.something` please

paasta_tools/mesos_maintenance.py
```diff
@@ -200,6 +200,49 @@ def is_host_down(hostname=getfqdn()):
     return hostname in get_down_hosts()
 
 
+def get_hosts_forgotten_draining():
+    """Find hosts that are still marked as draining (rather than down) after the start
+    of their maintenance window. For each of these hosts, either mark them as 'down' for
+    maintenance, adjust the time of their maintenance window, or cancel the maintenance
+    window if it is no longer needed.
+    :returns: a list of hostnames of hosts forgotten draining
```
Rob-Johnson: These docs aren't accurate (right?)- can you remove the ``For each of these hosts, either mark them as 'down' for maintenance, adjust the time of their maintenance window, or cancel the maintenance window if it is no longer needed.`` please?

paasta_tools/mesos_maintenance.py
nhandler: Heh, these prints should not have been included, but I can add some `log.debug` calls.

paasta_tools/mesos_maintenance.py
nhandler: The docs are accurate, but they would be better suited as a `Tip` on a `monitoring_check` than in the function definition. They describe actions a human operator can take to deal with hosts that are forgotten draining. I'll move them.

nhandler:
@mattmb Most of our maintenance will be managed by tools. I would also like to be able to use the start of the maintenance window as a sort of timeout on draining (hence these functions). In those situations, a forgotten draining host will likely be a sign that some tool is failing to timeout the draining process by transitioning the box to the `down` state or by undraining the box if we destroyed it and it no longer exists.

PR #800, opened 2016-09-28
Check connected frameworks

### LINKED ISSUE #798: paasta metastatus should consider the number of connected frameworks
Chronos behaves in unexpected ways sometimes - it can continue to operate even when the framework has died, so our check to see whether the api is available isn't really enough. Check on the number of connected/active frameworks in metastatus and page when we deviate from the expected count.

Here's a graph from the outage showing:

Active Frameworks:
image

Connected Frameworks:
image

Rob-Johnson:
closes #798 

- add healthcheck functions to metastatus for asserting the number of connected, disconnected, active, inactive frameworks
- update the ttl of the cached leader in MesosMaster to 5 seconds
- remove some duplication from metastatus
- update the MesosMaster.frameworks function to use the ``/master/frameworks`` endpoint, rather than ``state``

### DIFF
--- a/paasta_tools/mesos/master.py
+++ b/paasta_tools/mesos/master.py
@@ -59,7 +59,7 @@ class MesosMaster(object):
     def key(self):
         return self.config["master"]
 
-    @util.CachedProperty()
+    @util.CachedProperty(ttl=5)
     def host(self):
         return "{0}://{1}".format(self.config["scheme"], self.resolve(self.config["master"]))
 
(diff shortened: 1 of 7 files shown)


paasta_tools/mesos/master.py
```diff
@@ -199,13 +199,13 @@ def _framework_list(self, active_only=False):
         keys = ["frameworks"]
         if not active_only:
             keys.append("completed_frameworks")
-        return util.merge(self.state, *keys)
+        return util.merge(self.fetch("/master/frameworks").json(), *keys)
```
solarkennedy: Nice! On less state.json call!

paasta_tools/paasta_metastatus.py
```diff
@@ -660,20 +707,9 @@ def main():
...
-        print marathon_summary
-        print_results_for_healthchecks(marathon_ok, marathon_results, args.verbose)
-        print chronos_summary
-        print_results_for_healthchecks(chronos_ok, chronos_results, args.verbose)
-    else:
-        print mesos_summary
-        print_results_for_healthchecks(mesos_ok, all_mesos_results, args.verbose)
+    print "Master paasta_tools version: {0}".format(__version__)
+    print_results_for_healthchecks(mesos_summary, mesos_ok, all_mesos_results, args.verbose)
+    if args.verbose > 1:
```
solarkennedy: Huh. Sure looks like we are losing something here, but I guess by printing the summary in the healchecks stuff, we don't need such verbose if/else logic?

solarkennedy:
ship

PR #802, opened 2016-09-28
Performance-check uses a config file

kaisen:
Performance-check used to take in 'commit' and 'image' parameters and passed those explicitly to a performance check service endpoint defined in paasta configs.

This change makes it so that the service endpoint and general params are defined in a performance-check.yaml file in soa-configs. If no performance-check.yaml file is found then this command exits immediately.

### DIFF
--- a/paasta_tools/cli/cmds/performance_check.py
+++ b/paasta_tools/cli/cmds/performance_check.py
@@ -41,19 +41,20 @@ def add_subparser(subparsers):
 
 
 def load_performance_check_config(service, soa_dir):
-    performance_check_config = read_extra_service_information(
+    return read_extra_service_information(
         service_name=service,
         extra_info='performance-check',
         soa_dir=soa_dir,
     )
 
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/performance_check.py
```diff
@@ -12,70 +12,66 @@
...
-        sys.exit(0)
+def load_performance_check_config(service, soa_dir):
+    performance_check_config = read_extra_service_information(
+        service_name=service,
+        extra_info='performance-check',
+        soa_dir=soa_dir,
+    )
+
+    if not performance_check_config:
+        raise Exception("No performance check config found. Check your performance-check.yaml file.")
```
solarkennedy: Now, I understand where you are going with this, but because this will be the norm most of the time, I say "don't" raise.

Just let the submit function say:

```
print(No performance-check.yaml. Skipping.)
return
```

And then no need for all the scary messages.

PR #803, opened 2016-09-28
Start supporting xenial

nhandler:
This change makes Travis start to build paasta-tools for xenial. It does *not* change the version of Ubuntu we use in things like `paasta_itests` (that remains `trusty` for now). It also doesn't touch the bintray upload settings (which are set to only upload when `make itest_trusty` is called on a tagged release). So overall, this should be a pretty simple and low-risk change.

### DIFF
--- a/yelp_package/Makefile
+++ b/yelp_package/Makefile
@@ -21,6 +21,7 @@ UID:=`id -u`
 GID:=`id -g`
 DOCKER_RUN_LUCID:=docker run -t -v  $(CURDIR)/../:/work:rw paastatools_lucid_container
 DOCKER_RUN_TRUSTY:=docker run -t -v  $(CURDIR)/../:/work:rw paastatools_trusty_container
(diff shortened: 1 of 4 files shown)


yelp_package/Makefile
```diff
@@ -69,6 +70,23 @@ package_trusty: build_trusty_docker
 itest_trusty: package_trusty bintray.json
 	$(DOCKER_RUN_TRUSTY) /work/yelp_package/itest/ubuntu.sh paasta-tools_$(ACTUAL_PACKAGE_VERSION)_amd64.deb
 
+build_xenial_docker:
```
solarkennedy: At 3 times, it is a good excuse to dry this out if you can. I won't block a shipit though because make is scary amirite? :)

nhandler: Let me try and add in some wildcards like we've done in other projects.

nhandler:
The refactoring of the Makefile to DRY it up is a bit complicated. I'm going to merge this PR and create a new PR for the refactor.

PR #806, opened 2016-09-29
add info about queuing chronos tasks to metastatus

Rob-Johnson:
closes #806 
For now, this always returns OK, so won't ever trigger a page.


solarkennedy:
Ship, it is a nice to have

PR #807, opened 2016-09-29
DRY up the Makefile

nhandler:
This change attempts to DRY up the Makefile a bit. In order to do so, I also enabled bintray uploads for lucid and xenial (in addition to the already enabled trusty). I don't think that should cause any problems.


EvanKrall:
If everything passes, looks good to me.

Onward, to an only-6 month old OS!

PR #808, opened 2016-09-30
Fix metastatus

nhandler:
My dynamic reservation change broke our calculations of available resources in metastatus. I meant to treat a reserved resource as used, as if we reserve all the resources on a box, it has no free resources for tasks to run. But I put a `-` instead of a `+`.


mjksmith:
Ship

PR #809, opened 2016-09-30
Fix metastatus health calculations

nhandler:
We were previously converting from MB to GB before adding dynamic
reservations (MBs). This change adds the dynamic reservations before
doing the unit converion.


mjksmith:
SHIPIT

PR #810, opened 2016-09-30
Add cleanup_maintenance script

nhandler:
We've ended up with resources left reserved for maintenance on boxes that are marked as `up`. We have boxes marked as `draining` and `down` that are either non-existent or no longer preparing for a change that prevents them from being `up`.

This is a first pass at a cleanup script. It is designed to run in a cron job on the leader in each cluster.

Is this a bit extreme? Possibly. It relies on the assumption that the majority of our draining will be handled by tools with strictly enforced timeouts. In the event a human is manually performing maintenance, they will need to be very careful to set the maintenance window appropriately (but this is becoming more and more rare).

### DIFF
--- a/paasta_tools/cleanup_maintenance.py
+++ b/paasta_tools/cleanup_maintenance.py
@@ -26,6 +26,7 @@ from paasta_tools.mesos_maintenance import get_draining_hosts
 from paasta_tools.mesos_maintenance import get_hosts_forgotten_down
 from paasta_tools.mesos_maintenance import get_hosts_forgotten_draining
 from paasta_tools.mesos_maintenance import reserve_all_resources
+from paasta_tools.mesos_maintenance import seconds_to_nanoseconds
 from paasta_tools.mesos_maintenance import undrain
 from paasta_tools.mesos_maintenance import unreserve_all_resources
 from paasta_tools.mesos_maintenance import up
@@ -44,13 +45,15 @@ def parse_args():
 def cleanup_forgotten_draining():
     """Clean up hosts forgotten draining"""
     log.debug("Cleaning up hosts forgotten draining")
-    undrain(hostnames=get_hosts_forgotten_draining())
+    hosts_forgotten_draining = get_hosts_forgotten_draining(grace=seconds_to_nanoseconds(10 * 60))
+    undrain(hostnames=hosts_forgotten_draining)
 
 
 def cleanup_forgotten_down():
     """Clean up hosts forgotten down"""
     log.debug("Cleaning up hosts forgotten down")
-    up(hostnames=get_hosts_forgotten_down())
+    hosts_forgotten_down = get_hosts_forgotten_down(grace=seconds_to_nanoseconds(10 * 60))
+    up(hostnames=hosts_forgotten_down)
 
 
 def unreserve_all_resources_on_non_draining_hosts():
--- a/paasta_tools/cleanup_maintenance.py
+++ b/paasta_tools/cleanup_maintenance.py
@@ -72,7 +72,7 @@ def reserve_all_resources_on_draining_hosts():
     reserve_all_resources(hostnames=get_draining_hosts())
 
 
-def cleanup_maintenance():
+def main():
     log.debug("Cleaning up maintenance cruft")
     parse_args()
 
@@ -83,6 +83,6 @@ def cleanup_maintenance():
 
 
 if __name__ == "__main__":
-    if cleanup_maintenance():
+    if main():
         sys.exit(0)
     sys.exit(1)
--- a/paasta_tools/cleanup_maintenance.py
+++ b/paasta_tools/cleanup_maintenance.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python
+# Copyright 2016 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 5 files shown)


paasta_tools/cleanup_maintenance.py
```diff
@@ -0,0 +1,85 @@
...
+    """Clean up hosts forgotten down"""
+    log.debug("Cleaning up hosts forgotten down")
+    up(hostnames=get_hosts_forgotten_down())
+
+
+def unreserve_all_resources_on_non_draining_hosts():
+    """Unreserve all resources on non-draining hosts"""
+    log.debug("Unreserving all resources on non-draining hosts")
+    slaves = get_slaves()
+    hostnames = map((lambda slave: slave['hostname']), slaves)
```
mjksmith: hostnames = [slave['hostname'] for slave in slaves]

paasta_tools/cleanup_maintenance.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def cleanup_maintenance():
+    log.debug("Cleaning up maintenance cruft")
+    parse_args()
+
+    cleanup_forgotten_draining()
+    cleanup_forgotten_down()
+    unreserve_all_resources_on_non_draining_hosts()
+    reserve_all_resources_on_draining_hosts()
```
mattmb: Doesn't `setup_marathon_job` take care of this anyway?

paasta_tools/cleanup_maintenance.py
```diff
@@ -0,0 +1,85 @@
...
+    """Reserve all resources on draining hosts"""
+    log.debug("Reserving all resources on draining hosts")
+    reserve_all_resources(hostnames=get_draining_hosts())
+
+
+def cleanup_maintenance():
+    log.debug("Cleaning up maintenance cruft")
+    parse_args()
+
+    cleanup_forgotten_draining()
```
mattmb: This one still worries me a bit, for example:
* Autoscaler calls `drain(host)` with a maintenance window in 15 mins.
* After 15 mins the autoscaler might timeout and terminate the box (assuming the draining didn't complete)
* This termination isn't instantaneous
* In the window: after the 15 mins but before the autoscaler has a chance to kill the box this script could run and end up marking the box as undrained.

Maybe it's a small race condition? Maybe it could wait for `past_maintenance_start` + 10 mins though?

paasta_tools/cleanup_maintenance.py
```diff
@@ -0,0 +1,85 @@
...
+from paasta_tools.mesos_maintenance import up
+from paasta_tools.mesos_tools import get_slaves
+
+
+log = logging.getLogger(__name__)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Cleans up forgotten maintenance cruft.')
+    args = parser.parse_args()
```
mattmb: You might like to add parsing for `-v`s to set the log level.

paasta_tools/cleanup_maintenance.py
nhandler: I agree that is a valid potential race condition. I am not thrilled with hardcoding a certain number of minutes, as 10 minutes is a pretty long time when dealing with 2 minute timeouts for AWS boxes going away. What if, for the first pass, as a bit of a compromise, I include the `pasta_maintenance_start`+10 minutes, but let it bypass the 10 minutes if the slave is no longer registered in mesos (i.e. it was shutdown or terminated)?

paasta_tools/cleanup_maintenance.py
nhandler: Sort of. It doesn't do any of the cleanup for forgotten draining/down hosts. As for reserving/unreserving resources, there is a race condition. We only reserve resources when we kill a task on an `at-risk` host. If we try and reserve the resources before they are freed up, they will remain unreserved. This is particularly bad if we encounter the race condition on the last running task on an at-risk host. I looked into making `setup_marathon_job` try and do this cleanup, but we never actually get a list of all hosts, and I would prefer not to add that slowness/complexity into an already bloated script.

paasta_tools/cleanup_maintenance.py
nhandler: Adding the additional check to see if the host is registered in mesos is a bit more annoying due to how I have the code structured. It either involves placing it in a misleading location or adding a ton of extra API calls. I'll work on refactoring the code to make that happen, but to avoid blocking this PR too long, I've gone ahead and added the 10 minute grace period. While not ideal, I think it should still be better than what we currently have.

setup.py
```diff
@@ -74,6 +74,7 @@
         'paasta_tools/autoscale_all_services.py',
         'paasta_tools/check_marathon_services_replication.py',
         'paasta_tools/cleanup_marathon_jobs.py',
+        'paasta_tools/cleanup_maintenance.py',
```
solarkennedy: Can you use entrypoints instead? Can you can name the entrpoint `paasta_cleanup_maintenance`? I feel like we pollute the path every time we add a non-namespaced command like this.

nhandler: Done

paasta_tools/cleanup_maintenance.py
mattmb: Sounds good to me!

paasta_tools/cleanup_maintenance.py
mattmb: Sorry, I actually meant just `reserve_all_resources_on_draining_hosts`.

I'm happy for this to stay in, I guess it shouldn't do any harm.

Issue #811, opened 2016-10-03
paasta api with swagger client not able to handle errors


huadongliu:
paasta api uses [pyramid exception views to handle errors](https://github.com/Yelp/paasta/blob/master/paasta_tools/api/views/instance.py#L88). bravado throws exception when 404 is injected on the endpoint. 404 has been [defined in swagger.json](https://github.com/Yelp/paasta/blob/master/paasta_tools/api/api_docs/swagger.json#L72).

I am trying to use bravado config={'[also_return_response](http://bravado.readthedocs.io/en/latest/advanced.html)': True} to get http status code before interpreting results. The same exception is thrown.
 
$ USE_API_ENDPOINT=true paasta status -s robj-test-service -i main

cluster: mesosstage
Traceback (most recent call last):
  File "/nail/home/hliu/paasta/.tox/py27/bin/paasta", line 9, in <module>
    load_entry_point('paasta-tools==0.50.2', 'console_scripts', 'paasta')()
  File "/nail/home/hliu/paasta/paasta_tools/cli/cli.py", line 118, in main
    return_code = args.command(args)
  File "/nail/home/hliu/paasta/paasta_tools/cli/cmds/status.py", line 286, in paasta_status
    use_api_endpoint=use_api_endpoint
  File "/nail/home/hliu/paasta/paasta_tools/cli/cmds/status.py", line 248, in report_status
    use_api_endpoint=use_api_endpoint
  File "/nail/home/hliu/paasta/paasta_tools/cli/cmds/status.py", line 203, in report_status_for_cluster
    paasta_status_on_api_endpoint(cluster, service, instance, system_paasta_config, verbose=verbose)
  File "/nail/home/hliu/paasta/paasta_tools/cli/cmds/status.py", line 142, in paasta_status_on_api_endpoint
    status, h = client.service.status_instance(service=service, instance=instance).result()
  File "/nail/home/hliu/paasta/.tox/py27/local/lib/python2.7/site-packages/bravado/http_future.py", line 77, in result
    self.response_callbacks)
  File "/nail/home/hliu/paasta/.tox/py27/local/lib/python2.7/site-packages/bravado/http_future.py", line 126, in unmarshal_response
    raise_on_expected(incoming_response)
  File "/nail/home/hliu/paasta/.tox/py27/local/lib/python2.7/site-packages/bravado/http_future.py", line 149, in raise_on_expected
    swagger_result=http_response.swagger_result)
bravado.exception.HTTPError: 404 Not Found

[...]

PR #812, opened 2016-10-03
handle paasta api errors

huadongliu:
This is to fix [issue 811](https://github.com/Yelp/paasta/issues/811).

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -17,6 +17,7 @@ import sys
 from distutils.util import strtobool
 from subprocess import CalledProcessError
 
+from bravado.exception import HTTPError
 from service_configuration_lib import read_deploy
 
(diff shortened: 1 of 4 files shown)


mattmb:
You can see I had to do something similar in: https://github.com/Yelp/paasta/blob/1f5bf0b4d692f8a25c64d542338ddc97e25d6061/paasta_tools/cli/cmds/mark_for_deployment.py#L217

I wonder if we should add some code to our client that wraps the bravado call and raises our own errors? Or maybe it's okay to let the client user decide how to handle the HTTPError?

Anyway, this looks good so shipit!

huadongliu:
Yeah, I thought about wrapping the bravado call. I didn't go that route because bravado exceptions are [well defined](https://github.com/Yelp/bravado/blob/master/bravado/exception.py#L83) recently.

PR #815, opened 2016-10-04
Make mark for deployment more resilient

mattmb:
Haven't seen this happen but mark-for-deployment could incorrectly say
an instance is deployed if there is an error returned from the paasta
api. This seems undesirable so this widens the exception handling to
make it assume the instance isn't deployed if the paasta api returns
errors.


solarkennedy:
ship

PR #816, opened 2016-10-04
Add caching on frameworks endpoint in mesos-cli

mattmb:
After 3d07cbeacc4cfe0e9556500a1e51cbe4b49e5f4c we are getting the
frameworks from the proper endpoint. However, we aren't caching the
response as we were with the state endpoint. This means that anywhere
which calls the master.framework (which includes things like
task.framework) had to make a new HTTP call to /master/frameworks each
time.

This manifested in slow runs of the autoscaler where we happen to call
task.framework once per task in the cluster. See
get_mesos_task_count_by_slave in mesos_tools.

This adds the same caching as is used on the state.json endpoint.

### DIFF
--- a/tests/mesos/test_master.py
+++ b/tests/mesos/test_master.py
@@ -1,3 +1,4 @@
+from mock import Mock
 from mock import patch
 
 from paasta_tools.mesos import framework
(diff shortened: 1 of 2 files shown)


mattmb:
@Rob-Johnson 👍 definitely a good plan, tests added

tests/mesos/test_master.py
```diff
@@ -34,3 +35,30 @@ def test_framework_list_includes_completed_frameworks(mock_framework_list):
...
+    assert ret == mock_frameworks
+
+
+@patch.object(master.MesosMaster, '_frameworks', autospec=True)
+def test__framework_list(mock__frameworks):
+    mock_frameworks = Mock()
+    mock_completed = Mock()
+    mock_return = Mock(return_value={'frameworks': [mock_frameworks],
+                                     'completed_frameworks': [mock_completed]})
+    mock__frameworks.__get__ = mock_return
```
Rob-Johnson: mock_frameworks.return_value = mock_return?

PR #817, opened 2016-10-04
Validate paasta api update_autoscaler_count request

huadongliu:
We want to make sure the service.instance exists and the desired_instance count is between min and max instance counts.

> $ curl -X POST -d '{"desired_instances":4}' http://paasta-mesosstage.yelp:5055/v1/services/robj-test-service/autoscale_to_heck2/autoscaler
> ERROR: Unable to load service config for robj-test-service.autoscale_to_heck2
> 
> $ curl -X POST -d '{"desired_instances":3}' http://paasta-mesosstage.yelp:5055/v1/services/robj-test-service/autoscale_to_heck/autoscaler
> {"status":"SUCCESS","desired_instances":3}
> 
> $ paasta status -s robj-test-service -i autoscale_to_heck
> cluster: mesosstage
> instance: autoscale_to_heck
> Git sha:    ba88d9ce (desired)
> State:      Running - Desired state: Started
> Marathon:   Healthy - up with (4/3) instances. Status: Deploying
> Mesos:      Healthy - (4/3) tasks in the TASK_RUNNING state.
> 
> $ curl -X POST -d '{"desired_instances":1}' http://paasta-mesosstage.yelp:5055/v1/services/robj-test-service/autoscale_to_heck/autoscaler
> {"status":"WARNING desired_instances is less than min_instances 2","desired_instances":2}
> 
> $ paasta status -s robj-test-service -i autoscale_to_heck
> cluster: mesosstage
> instance: autoscale_to_heck
> Git sha:    ba88d9ce (desired)
> WARNING:paasta_tools.marathon_tools:Returning limited instance count 2. (zk had 1)
> State:      Running - Desired state: Started
> Marathon:   Healthy - up with (4/2) instances. Status: Deploying
> Mesos:      Healthy - (4/2) tasks in the TASK_RUNNING state.
> 
> $ curl -X POST -d '{"desired_instances":5}' http://paasta-mesosstage.yelp:5055/v1/services/robj-test-service/autoscale_to_heck/autoscaler
> {"status":"WARNING desired_instances is greater than max_instances 4","desired_instances":4}
> 
> $ paasta status -s robj-test-service -i autoscale_to_heck
> cluster: mesosstage
> instance: autoscale_to_heck
> Git sha:    ba88d9ce (desired)
> WARNING:paasta_tools.marathon_tools:Returning limited instance count 4. (zk had 5)
> State:      Running - Desired state: Started
> Marathon:   Warning - up with (3/4) instances. Status: Deploying
> Mesos:      Healthy - (4/4) tasks in the TASK_RUNNING state.
> 
> $ curl -X POST -d '{"desired_instances":4}' http://paasta-mesosstage.yelp:5055/v1/services/robj-test-service/main/autoscaler
> ERROR: Autoscaling is not enabled for robj-test-service.main

### DIFF
--- a/paasta_tools/api/views/autoscaler.py
+++ b/paasta_tools/api/views/autoscaler.py
@@ -19,6 +19,7 @@ from pyramid.response import Response
 from pyramid.view import view_config
 
 from paasta_tools.api import settings
+from paasta_tools.api.views.exception import ApiFailure
 from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.marathon_tools import set_instances_for_marathon_service
 
@@ -29,13 +30,18 @@ def get_autoscaler_count(request):
     instance = request.swagger_data.get('instance')
     cluster = settings.cluster
     soa_dir = settings.soa_dir
-    service_config = load_marathon_service_config(
-        service=service,
-        instance=instance,
-        cluster=cluster,
-        soa_dir=soa_dir,
-        load_deployments=False,
-    )
+    try:
+        service_config = load_marathon_service_config(
+            service=service,
+            instance=instance,
+            cluster=cluster,
+            soa_dir=soa_dir,
+            load_deployments=False,
+        )
+    except:
+        error_message = 'Unable to load service config for %s.%s' % (service, instance)
+        raise ApiFailure(error_message, 404)
+
     response_body = {'desired_instances': service_config.get_instances()}
     return Response(json_body=response_body, status_code=200)
 
@@ -45,6 +51,36 @@ def update_autoscaler_count(request):
     service = request.swagger_data.get('service')
     instance = request.swagger_data.get('instance')
     desired_instances = request.swagger_data.get('json_body')['desired_instances']
+
+    try:
+        service_config = load_marathon_service_config(
+            service=service,
+            instance=instance,
+            cluster=settings.cluster,
+            soa_dir=settings.soa_dir,
+            load_deployments=False,
+        )
+    except:
+        error_message = 'Unable to load service config for %s.%s' % (service, instance)
+        raise ApiFailure(error_message, 404)
+
+    max_instances = service_config.get_max_instances()
+    if max_instances is None:
+        error_message = 'Autoscaling is not enabled for %s.%s' % (service, instance)
+        raise ApiFailure(error_message, 404)
+
+    min_instances = service_config.get_min_instances()
+
+    # Dump whatever number from the client to zk. get_instances() will limit
+    # readings from zk to [min_instances, max_instances].
     set_instances_for_marathon_service(service=service, instance=instance, instance_count=desired_instances)
-    response_body = {'desired_instances': desired_instances}
+    status = 'SUCCESS'
+    if desired_instances > max_instances:
+        desired_instances = max_instances
+        status = 'WARNING desired_instances is greater than max_instances %d' % max_instances
+    elif desired_instances < min_instances:
+        desired_instances = min_instances
+        status = 'WARNING desired_instances is less than min_instances %d' % min_instances
+
+    response_body = {'desired_instances': desired_instances, 'status': status}
     return Response(json_body=response_body, status_code=202)
--- a/paasta_itests/steps/paasta_api_steps.py
+++ b/paasta_itests/steps/paasta_api_steps.py
@@ -15,8 +15,8 @@ from behave import then
 from pyramid import testing
 
 from paasta_tools.api import settings
+from paasta_tools.api.views.exception import ApiFailure
 from paasta_tools.api.views.instance import instance_status
-from paasta_tools.api.views.instance import InstanceFailure
 from paasta_tools.utils import decompose_job_id
 
 
@@ -44,7 +44,7 @@ def service_instance_status_error(context, error_code, job_id):
     response = None
     try:
         response = instance_status(request)
-    except InstanceFailure as exc:
+    except ApiFailure as exc:
         assert 'not found' in exc.msg
         assert exc.err == int(error_code)
     except:
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -101,7 +101,7 @@
       "get": {
         "responses": {
           "200": {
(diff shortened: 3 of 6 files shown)


paasta_tools/api/views/autoscaler.py
```diff
@@ -45,6 +46,25 @@ def update_autoscaler_count(request):
...
+            soa_dir=settings.soa_dir,
+            load_deployments=False,
+        )
+    except:
+        error_message = 'Unable to load service config for %s.%s' % (service, instance)
+        raise ApiFailure(error_message, 404)
+
+    max_instances = service_config.get_max_instances()
+    min_instances = service_config.get_min_instances()
+    if desired_instances > max_instances or desired_instances < min_instances:
```
mjksmith: `if min_instances < desired_instances < max_instances`

mjksmith: Also shouldn't this be using `<=` and not `<`?

paasta_tools/api/views/autoscaler.py
```diff
@@ -45,6 +46,25 @@ def update_autoscaler_count(request):
...
+        error_message = 'Incorrect desired_instances %d: min_instances %d max_instances %d' % (desired_instances, min_instances, max_instances)
+        raise ApiFailure(error_message, 400)
+
     set_instances_for_marathon_service(service=service, instance=instance, instance_count=desired_instances)
```
mjksmith: I've been thinking about having this use `instance_count=service_config.get_instances()` to verify to the user that the changes have been logged in zookeeper. For example, if a service isn't configured for autoscaling and someone posts to its autoscaling endpoint the API will happily return {'desired_instances': foo} even though nothing actually happened.

paasta_tools/api/views/autoscaler.py
```diff
@@ -45,6 +46,25 @@ def update_autoscaler_count(request):
...
+        )
+    except:
+        error_message = 'Unable to load service config for %s.%s' % (service, instance)
+        raise ApiFailure(error_message, 404)
+
+    max_instances = service_config.get_max_instances()
+    min_instances = service_config.get_min_instances()
+    if desired_instances > max_instances or desired_instances < min_instances:
+        error_message = 'Incorrect desired_instances %d: min_instances %d max_instances %d' % (desired_instances, min_instances, max_instances)
+        raise ApiFailure(error_message, 400)
```
mjksmith: I don't really like this because this requires anyone who implements the autoscaler to be aware of the limits in yelpsoa-configs as well as handle 400 errors. I like the idea of returning `instance_count=service_config.get_instances()`, which is limited to the service's min- and max-instances. That way if an autoscaler postal an out-of-bounds instance count, they still get an HTTP 200 that tells them authoritatively what the current desired instance count . That way if an autoscaler is going 'oh my god we need 500000 instances' PaaSTA will still scale up to max_instances and report back to the autoscaler the current instance count instead of just disregarding the request.

paasta_tools/api/views/autoscaler.py
huadongliu: To summarize, we have two choices here:
1. Extend get_autoscaler_count to include min/max so the client can have knowledge of the count range. Fail update_autoscaler_count EXPLICITLY when the desired count is beyond the range.
2. IMPLICITLY bound the desired count to [min,max] and return 200 to make client happy.

I prefer the explicit approach, but I am fine to implement the implicit one if the team thinks otherwise.

paasta_tools/api/views/autoscaler.py
```diff
@@ -45,6 +51,35 @@ def update_autoscaler_count(request):
...
+    max_instances = service_config.get_max_instances()
+    if max_instances is None:
+        error_message = 'Autoscaling is not enabled for %s.%s' % (service, instance)
+        raise ApiFailure(error_message, 404)
+
+    min_instances = service_config.get_min_instances()
+
+    if desired_instances > max_instances:
+        set_instances_for_marathon_service(service=service, instance=instance, instance_count=max_instances)
+        error_message = 'desired_instances is set to max_instances %d' % max_instances
```
mjksmith: It would be nice if these all returned in the same format to make it easier to parse

paasta_tools/api/views/autoscaler.py
```diff
@@ -45,6 +51,35 @@ def update_autoscaler_count(request):
...
+
+    max_instances = service_config.get_max_instances()
+    if max_instances is None:
+        error_message = 'Autoscaling is not enabled for %s.%s' % (service, instance)
+        raise ApiFailure(error_message, 404)
+
+    min_instances = service_config.get_min_instances()
+
+    if desired_instances > max_instances:
+        set_instances_for_marathon_service(service=service, instance=instance, instance_count=max_instances)
```
mjksmith: PaaSTA already handles banding to min_ and max_instances, so all we need is `set_instances_for_marathon_service(service=service, instance=instance, instance_count=desired_instances)` before this logic.

paasta_tools/api/views/autoscaler.py
```diff
@@ -45,6 +51,36 @@ def update_autoscaler_count(request):
...
+    status = 'SUCCESS'
+    if desired_instances > max_instances:
+        desired_instances = max_instances
+        status = 'WARNING desired_instances is greater than max_instances %d' % max_instances
+    elif desired_instances < min_instances:
+        desired_instances = min_instances
+        status = 'WARNING desired_instances is less than min_instances %d' % min_instances
+
+    response_body = {'desired_instances': desired_instances, 'status': status}
     return Response(json_body=response_body, status_code=202)
```
mjksmith: We can still return a 400-series status code when the autoscaler takes a best-effort approach in response to an out-of-bounds input. I just wanted the response body to look the same in both cases.

This status message works fine though.

huadongliu: 400 cannot be returned here because pyramid will take it as an exception and try to find the corresponding view handling. We either return 202 with warning or handle 400 error. Client would be surprised to see an error while the number of instance got changed to min_ or max_instances.

mjksmith: 202 is fine then.

PR #818, opened 2016-10-04
Make cleanup-maintenance check for empty lists of hosts and support -v

nhandler:
Looks like `mesos_maintenance` gets unhappy when passed empty lists of hostnames. Let's not do that. We should also support `-v`.


mjksmith:
LGTM.

PR #820, opened 2016-10-05
Change the words a bit on paasta status

solarkennedy:
It is a super common question when users say "why does it say running" when their service is "stopped".


huadongliu:
shipit

PR #821, opened 2016-10-05
Add code to cleanup cancelled SFRs

mattmb:
I'll want to try this out in mesosstage before I do much more work on it. Feel free to have a look and give me some feedback but you may want to hold off giving a thorough review until I'm confident this works!

Major changes:
* Use FulfilledCapacity rather than current TargetCapacity as base for calculation of new TargetCapacity
* Set a property on each resource marking it as cancelled and process these resources first.
* If scaling a cancelled resource: only remove instances if the cluster is otherwise over provisioned or at target.
* Gradually scale down cancelled resources

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -139,6 +139,7 @@ def spotfleet_metrics_provider(spotfleet_request_id, resource, pool_settings):
         return 0, 0
     current, target = get_spot_fleet_delta(resource, error)
     if sfr['SpotFleetRequestState'] == 'cancelled_running':
+        resource['min_capacity'] = 0
         if current - target < 0:
             log.info("Not scaling cancelled SFR {0} because we are under provisioned".format(spotfleet_request_id))
             return 0, 0
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -134,14 +134,22 @@ def spotfleet_metrics_provider(spotfleet_request_id, resource, pool_settings, co
         sfr['ActiveInstances'] = get_spot_fleet_instances(spotfleet_request_id, region=resource['region'])
         resource['sfr'] = sfr
         error = get_sfr_utilization_error(spotfleet_request_id, resource, pool_settings)
+    elif sfr['SpotFleetRequestState'] in ['submitted', 'modifying', 'cancelled_terminating']:
+        log.warning("Not scaling an SFR in state: {0} so {1}, skipping...".format(sfr['SpotFleetRequestState'],
+                                                                                  spotfleet_request_id))
+        return 0, 0
     else:
-        log.warning("Unexpected SFR state: {0} for {1}, skipping...".format(sfr['SpotFleetRequestState'],
-                                                                            spotfleet_request_id))
+        log.error("Unexpected SFR state: {0} for {1}".format(sfr['SpotFleetRequestState'],
+                                                             spotfleet_request_id))
+        raise ClusterAutoscalingError
+    if is_aws_launching_sfr_instances(sfr) and sfr['SpotFleetRequestState'] == 'active':
+        log.warning("AWS hasn't reached the TargetCapacity that is currently set. We won't make any "
+                    "changes this time as we should wait for AWS to launch more instances first.")
         return 0, 0
     current, target = get_spot_fleet_delta(resource, error)
     if sfr['SpotFleetRequestState'] == 'cancelled_running':
         resource['min_capacity'] = 0
-        if current - target < 0:
+        if current < target:
             log.info("Not scaling cancelled SFR {0} because we are under provisioned".format(spotfleet_request_id))
             return 0, 0
         else:
@@ -149,6 +157,12 @@ def spotfleet_metrics_provider(spotfleet_request_id, resource, pool_settings, co
             if target == 1:
                 target = 0
     return current, target
+
+
+def is_aws_launching_sfr_instances(sfr):
+    fulfilled_capacity = sfr['SpotFleetRequestConfig']['FulfilledCapacity']
+    target_capacity = sfr['SpotFleetRequestConfig']['TargetCapacity']
+    return target_capacity > fulfilled_capacity
 
 
 def cleanup_cancelled_sfr_config(spotfleet_request_id, config_folder, dry_run=False):
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -133,7 +133,15 @@ def spotfleet_metrics_provider(spotfleet_request_id, resource, pool_settings, co
     elif sfr['SpotFleetRequestState'] in ['cancelled_running', 'active']:
         sfr['ActiveInstances'] = get_spot_fleet_instances(spotfleet_request_id, region=resource['region'])
         resource['sfr'] = sfr
-        error = get_sfr_utilization_error(spotfleet_request_id, resource, pool_settings)
+        desired_instances = len(sfr['ActiveInstances'])
+        mesos_state = get_mesos_master().state
+        slaves = get_sfr_slaves(resource, mesos_state)
+        error = get_mesos_utilization_error(spotfleet_request_id,
+                                            resource=resource,
+                                            pool_settings=pool_settings,
+                                            slaves=slaves,
+                                            mesos_state=mesos_state,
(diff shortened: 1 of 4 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -162,19 +190,11 @@ def spotfleet_metrics_provider(spotfleet_request_id, resource, pool_settings):
...
 
 def get_spot_fleet_delta(resource, error):
-    ec2_client = boto3.client('ec2', region_name=resource['region'])
-    spot_fleet_request = ec2_client.describe_spot_fleet_requests(
-        SpotFleetRequestIds=[resource['id']])['SpotFleetRequestConfigs'][0]
-    if spot_fleet_request['SpotFleetRequestState'] != 'active':
-        raise ClusterAutoscalingError('Can not scale non-active spot fleet requests. This one is "%s"' %
-                                      spot_fleet_request['SpotFleetRequestState'])
-    current_capacity = int(spot_fleet_request['SpotFleetRequestConfig']['TargetCapacity'])
+    current_capacity = int(resource['sfr']['SpotFleetRequestConfig']['FulfilledCapacity'])
```
mattmb: This is actually quite a "big" change but I think a correct one.
Currently:
* We get the current target from AWS.
* Calculate the percentage error based on mesos stats on cpu/ram etc.
* Set a new target based on that error
* Kill instances and count down to the new target from the old target, stopping when we get close.

Now:
* Get current fulfilled by AWS
* Calculate the percentage error based on mesos stats on cpu/ram etc.
* Set a new target based on that error
* Kill instances and count down to the new target from the current fulfilled capacity, stopping when we get close.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -123,14 +124,45 @@ def describe_instances(instance_ids, region=None, instance_filters=None):
 
 
 @register_autoscaling_component('aws_spot_fleet_request', CLUSTER_METRICS_PROVIDER_KEY)
-def spotfleet_metrics_provider(spotfleet_request_id, resource, pool_settings):
-    mesos_state = get_mesos_master().state
+def spotfleet_metrics_provider(spotfleet_request_id, resource, pool_settings, config_folder, dry_run=False):
     sfr = get_sfr(spotfleet_request_id, region=resource['region'])
-    if not sfr or not sfr['SpotFleetRequestState'] == 'active':
-        log.error("Ignoring SFR {0} that does not exist or is not active.".format(spotfleet_request_id))
+    if not sfr or sfr['SpotFleetRequestState'] == 'cancelled':
```
solarkennedy: With the `if not sfr` here, are you worried at all about cleaning up the file just because `get_sfr` returns None for any silly reason? (bad code, api timeout, etc?)

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -123,14 +124,45 @@ def describe_instances(instance_ids, region=None, instance_filters=None):
...
+        log.error("SFR not found, removing config file.".format(spotfleet_request_id))
+        cleanup_cancelled_sfr_config(spotfleet_request_id, config_folder, dry_run=dry_run)
+        return 0, 0
+    elif sfr['SpotFleetRequestState'] in ['cancelled_running', 'active']:
+        sfr['ActiveInstances'] = get_spot_fleet_instances(spotfleet_request_id, region=resource['region'])
+        resource['sfr'] = sfr
+        error = get_sfr_utilization_error(spotfleet_request_id, resource, pool_settings)
+    else:
+        log.warning("Unexpected SFR state: {0} for {1}, skipping...".format(sfr['SpotFleetRequestState'],
```
solarkennedy: Sure you don't want to raise here? What else is there?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -123,14 +124,45 @@ def describe_instances(instance_ids, region=None, instance_filters=None):
...
+                                                                            spotfleet_request_id))
         return 0, 0
-    sfr['ActiveInstances'] = get_spot_fleet_instances(spotfleet_request_id, region=resource['region'])
-    resource['sfr'] = sfr
+    current, target = get_spot_fleet_delta(resource, error)
+    if sfr['SpotFleetRequestState'] == 'cancelled_running':
+        resource['min_capacity'] = 0
+        if current - target < 0:
```
solarkennedy: Could this just be

    if current > target:
?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
solarkennedy: This does sound correct, except when a SFR is just getting started, or during times of low EC2 capacity, the autoscaler is going to "overshoot" right? Because on scaleup it will only look at what is fulfilled *so-far*, and then scale up even more?

mattmb: Hmm, thanks for this comment. I was trying to remember why it had been the other way before!

There's some defence for the first case because if we don't find a percentage of registered mesos slaves vs. the number of instances AWS has spawned we will bail.

I think what would be nice is a check of AWS Target vs Fulfilled. They always aim to over provision, so if the AWS fulfilled is less than the AWS target then it means we are still waiting for AWS to spin up more instances. In that case there's not much point increasing the target until the next run.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
mattmb: I think this is ok. It should only be `None` on `InvalidSpotFleetRequestId.NotFound`, other exceptions are raised. Also, if the config still exists in S3 (i.e. we haven't killed it with terraform) then the autoscaler should put it back on the next run.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
mattmb: There is also `modifying` and `submitted`. In those cases doing nothing is a good choice. I'll change this to make it clearer though, and raise on anything not in the list of states amazons docs provides.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
mattmb: Haha, yes indeed 😊

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -313,6 +337,10 @@ def set_spot_fleet_request_capacity(sfr_id, capacity, dry_run, region=None):
                 state = get_sfr(sfr_id, region=region)['SpotFleetRequestState']
                 if state == 'active':
                     break
+                if state == 'cancelled_running':
+                    log.info("Not updating target capacity because this is a cancelled SFR, "
+                             "we are just draining and killing the instances")
+                    return
                 log.debug("SFR {0} in state {1}, waiting for state: active".format(sfr_id, state))
                 log.debug("Sleep 5 seconds")
                 time.sleep(5)
```
mattmb: I think it's okay? If we reach the timeout then we end up raising an exception before we get to L353. If it is cancelled_running then we return so we don't reach any set calls.

The reasoning you state is correct: you can't set a TargetCapacity on a cancelled SFR. And that's because AWS won't launch instances for cancelled SFRs so a target is meaningless.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -123,31 +124,112 @@ def describe_instances(instance_ids, region=None, instance_filters=None):
...
+        log.error("Unexpected SFR state: {0} for {1}".format(sfr['SpotFleetRequestState'],
+                                                             spotfleet_request_id))
+        raise ClusterAutoscalingError
+    if is_aws_launching_sfr_instances(sfr) and sfr['SpotFleetRequestState'] == 'active':
+        log.warning("AWS hasn't reached the TargetCapacity that is currently set. We won't make any "
+                    "changes this time as we should wait for AWS to launch more instances first.")
+        return 0, 0
+    current, target = get_spot_fleet_delta(resource, error)
+    if sfr['SpotFleetRequestState'] == 'cancelled_running':
+        resource['min_capacity'] = 0
```
solarkennedy: Shouldn't this be max_capacity?

Because we want the autoscaler to scale *down* the canceled sfr, and setting the max=0 will force that.

mattmb: I think the reason I didn't do this is that we get a more gradual removal by removing the min capacity and setting the error to -1. I might change this in the future though as it would read clearer just to set both to 0.

PR #822, opened 2016-10-06
PAASTA-6732 fix exception in cluster.py when trying to throw FileNotFoundForTaskException

Rob-Johnson:
closes PAASTA-6732

- fix the typo causing the original bug
- remove the globals in use
- remove the option to *not* make this throw an exception.
- add some tests to better understand its behaviour


solarkennedy:
How were we using this code for so long! :)

PR #823, opened 2016-10-06
Fix new cli commands

mattmb:
* Adds a new `<service>/<instance>/tasks` API endpoint and use this for `docker_*`
* Fix sysdig task filter
* Improve API dev environment

### DIFF
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -24,6 +24,7 @@ from paasta_tools.api import settings
 from paasta_tools.api.views.exception import ApiFailure
 from paasta_tools.cli.cmds.status import get_actual_deployments
 from paasta_tools.mesos_tools import get_running_tasks_from_active_frameworks
+from paasta_tools.mesos_tools import get_tasks_from_app_id
 from paasta_tools.paasta_serviceinit import get_deployment_version
 from paasta_tools.utils import NoDockerImageError
 from paasta_tools.utils import validate_service_instance
@@ -113,3 +114,20 @@ def instance_status(request):
         raise ApiFailure(error_message, 500)
 
     return instance_status
+
+
+@view_config(route_name='service.instance.tasks', request_method='GET', renderer='json')
+def instance_tasks(request):
+    status = instance_status(request)
+    task_id = request.swagger_data.get('task_id', None)
+    slave_hostname = request.swagger_data.get('slave_hostname', None)
+    try:
+        mstatus = status['marathon']
+    except KeyError:
+        raise InstanceFailure("Only marathon tasks supported", 400)
+    tasks = get_tasks_from_app_id(mstatus['app_id'], slave_hostname=slave_hostname, task_id=task_id)
+    tasks = [{'task_id': task['id'],
+              'slave_id': task.executor['tasks'][0]['slave_id'],
+              'slave_hostname': task.slave['hostname'],
+              'container_id': task.executor['container']} for task in tasks]
+    return tasks
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -24,7 +24,9 @@ from paasta_tools.api import settings
 from paasta_tools.api.views.exception import ApiFailure
 from paasta_tools.cli.cmds.status import get_actual_deployments
 from paasta_tools.mesos_tools import get_running_tasks_from_active_frameworks
+from paasta_tools.mesos_tools import get_task
 from paasta_tools.mesos_tools import get_tasks_from_app_id
+from paasta_tools.mesos_tools import TaskNotFound
 from paasta_tools.paasta_serviceinit import get_deployment_version
 from paasta_tools.utils import NoDockerImageError
 from paasta_tools.utils import validate_service_instance
@@ -116,18 +118,52 @@ def instance_status(request):
     return instance_status
 
 
-@view_config(route_name='service.instance.tasks', request_method='GET', renderer='json')
-def instance_tasks(request):
+@view_config(route_name='service.instance.tasks.task', request_method='GET', renderer='json')
+def instance_task(request):
     status = instance_status(request)
     task_id = request.swagger_data.get('task_id', None)
-    slave_hostname = request.swagger_data.get('slave_hostname', None)
+    verbose = request.swagger_data.get('verbose', False)
     try:
         mstatus = status['marathon']
     except KeyError:
-        raise InstanceFailure("Only marathon tasks supported", 400)
-    tasks = get_tasks_from_app_id(mstatus['app_id'], slave_hostname=slave_hostname, task_id=task_id)
-    tasks = [{'task_id': task['id'],
-              'slave_id': task.executor['tasks'][0]['slave_id'],
-              'slave_hostname': task.slave['hostname'],
-              'container_id': task.executor['container']} for task in tasks]
-    return tasks
+        raise ApiFailure("Only marathon tasks supported", 400)
+    try:
+        task = get_task(task_id, app_id=mstatus['app_id'])
+    except TaskNotFound:
+        raise ApiFailure("Task with id {0} not found".format(task_id), 404)
+    except:
+        error_message = traceback.format_exc()
+        raise ApiFailure(error_message, 500)
+    if verbose:
+        task = add_slave_info(task)
+        task = add_executor_info(task)
+    return task._Task__items
+
+
+@view_config(route_name='service.instance.tasks', request_method='GET', renderer='json')
+def instance_tasks(request):
+    status = instance_status(request)
+    slave_hostname = request.swagger_data.get('slave_hostname', None)
+    verbose = request.swagger_data.get('verbose', False)
+    try:
+        mstatus = status['marathon']
+    except KeyError:
+        raise ApiFailure("Only marathon tasks supported", 400)
+    tasks = get_tasks_from_app_id(mstatus['app_id'], slave_hostname=slave_hostname)
+    if verbose:
+        tasks = [add_executor_info(task) for task in tasks]
+        tasks = [add_slave_info(task) for task in tasks]
+    return [task._Task__items for task in tasks]
+
+
+def add_executor_info(task):
+    task._Task__items['executor'] = task.executor.copy()
+    task._Task__items['executor'].pop('tasks', None)
+    task._Task__items['executor'].pop('completed_tasks', None)
+    task._Task__items['executor'].pop('queued_tasks', None)
+    return task
+
+
+def add_slave_info(task):
+    task._Task__items['slave'] = task.slave._MesosSlave__items.copy()
+    return task
--- a/paasta_tools/cli/cmds/docker_inspect.py
+++ b/paasta_tools/cli/cmds/docker_inspect.py
@@ -13,12 +13,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import subprocess
+import sys
 
-from paasta_tools.cli.utils import get_status_for_instance
-from paasta_tools.cli.utils import get_subparser
-from paasta_tools.cli.utils import pick_slave_from_status
 from paasta_tools.cli.utils import get_container_name
+from paasta_tools.cli.utils import get_subparser
 from paasta_tools.cli.utils import get_task_from_instance
+from paasta_tools.cli.utils import PaastaTaskNotFound
 
 
 def add_subparser(subparsers):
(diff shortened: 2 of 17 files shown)


paasta_tools/api/views/instance.py
```diff
@@ -138,3 +139,20 @@ def instance_status(request):
...
+def instance_tasks(request):
+    status = instance_status(request)
+    task_id = request.swagger_data.get('task_id', None)
+    slave_hostname = request.swagger_data.get('slave_hostname', None)
+    try:
+        mstatus = status['marathon']
+    except KeyError:
+        raise InstanceFailure("Only marathon tasks supported", 400)
+    tasks = get_tasks_from_app_id(mstatus['app_id'], slave_hostname=slave_hostname, task_id=task_id)
+    tasks = [{'task_id': task['id'],
```
Rob-Johnson: sorry to be a pain, but I'd prefer:

```
/service/{service}/instance/{instance}/tasks
['task_1','task_2','task_3']

/service/{service}/instance/{instance}/tasks/{task_id}
{
   'slave_id': ...
   'slave_hostname': ....
   'container_id': ...
}
```

also, why deliver such a subset of information? Isn't there a bunch more info available from the respective marathon/mesos apis?

paasta_tools/cli/utils.py
```diff
@@ -742,3 +744,34 @@ def pick_slave_from_status(status, host=None):
     else:
         slaves = status.marathon.slaves
         return slaves[0]
+
+
+def get_task_from_instance(cluster, service, instance, slave_hostname=None, task_id=None):
+    api = client.get_paasta_api_client(cluster=cluster)
+    if not api:
+        sys.exit(1)
```
Rob-Johnson: provide a useful message here

paasta_tools/cli/utils.py
```diff
@@ -742,3 +744,34 @@ def pick_slave_from_status(status, host=None):
     else:
         slaves = status.marathon.slaves
         return slaves[0]
+
+
+def get_task_from_instance(cluster, service, instance, slave_hostname=None, task_id=None):
```
Rob-Johnson: It's not clear to me why we're not reusing the logic that ``paasta status`` uses to get task info? This seems like a really limited subset of the things you'd get if you just used the mesos api?

paasta_tools/mesos_tools.py
```diff
@@ -655,21 +655,10 @@ def terminate_framework(framework_id):
     resp.raise_for_status()
 
 
-def get_container_name(app_id, slave_hostname=None, task_id=None):
+def get_tasks_from_app_id(app_id, slave_hostname=None, task_id=None):
```
Rob-Johnson: I don't really understand why you've moved ``get_container_name`` to be in ``cli/utils.py``? Its usage is wider reaching that just the cli tools.

paasta_tools/mesos_tools.py
```diff
@@ -655,21 +655,10 @@ def terminate_framework(framework_id):
...
-    except StopIteration:
-        raise TaskNotFound("Couldn't find task")
-    container_name = "mesos-{0}.{1}".format(task.executor['tasks'][0]['slave_id'],
-                                            task.executor['container'])
-    return container_name
-
-
-class TaskNotFound(Exception):
-    pass
+    if task_id:
```
Rob-Johnson: this function serves 3 different purposes.

Please split them up into ``get_all_tasks_for_app``, ``filter_tasks_for_hostname``, ``get_task``

paasta_tools/cli/utils.py
```diff
@@ -742,3 +744,34 @@ def pick_slave_from_status(status, host=None):
...
+        sys.exit(1)
+    except HTTPError as e:
+        log.error("Problem with API call to find task details")
+        log.error(e.response.text)
+        sys.exit(1)
+    if not tasks:
+        log.error("Cannot find any tasks on host: {0} or with task_id: {1}".format(slave_hostname,
+                                                                                   task_id))
+        sys.exit(1)
+    task = tasks[0]
```
Rob-Johnson: You can't make guarantees about the ordering of this; is it really useful for the function just to return the first one? Do you have a usecase for 'given an appid, return one of its tasks (but I don't care which one)?'

paasta_tools/api/views/instance.py
mattmb: I think I actually tried to do this but pyramid (and or swagger) doesn't let us have optional params in the path. So it'd actually have to be something like: `/service/{service}/instance/{instance}/tasks` and a second endpoint: `/service/{service}/instance/{instance}/tasks/task/{task_id}`

Which seems worse? As for the filtering, just seemed easier to return what we need for now but I could be convinced either way.

paasta_tools/cli/utils.py
mattmb: Well, I don't actually care which task, I just need one. I could change this to return the list but then each cli client has to have the `[0]` or some other method for picking just one.

paasta_tools/mesos_tools.py
mattmb: Mostly because I call `sys.exit` from it so I don't want non cli use. I guess it would be better to raise an exception, then I could catch it in the cli clients and call the `sys.exit` there.

paasta_tools/cli/utils.py
mattmb: Do you know where about I could get the same information? Also doesn't `paasta status` just print strings?

mattmb:
I've also missed some tests that are failing

paasta_tools/api/views/instance.py
mattmb: This turns out to be a lie. It's totally possible, you just have to define a separate endpoint in the `swagger.json` and make the path variable required.

paasta_tools/cli/cmds/docker_stop.py
```diff
@@ -32,11 +33,15 @@ def add_subparser(subparsers):
...
-                                   host=args.host)
-    container = get_container_name(status.marathon.app_id, slave_hostname=slave, task_id=args.mesos_id)
+    try:
+        task = get_task_from_instance(cluster=args.cluster,
+                                      service=args.service,
+                                      instance=args.instance,
+                                      slave_hostname=args.host,
+                                      task_id=args.mesos_id)
+    except TaskNotFound:
+        sys.exit(1)
```
Rob-Johnson: log?

paasta_tools/cli/cmds/sysdig.py
```diff
@@ -85,5 +85,5 @@ def paasta_sysdig(args):
 
 def format_mesos_command(slave, app_id, mesos_url, marathon_url):
     sysdig_mesos = '{0},{1}'.format(mesos_url, marathon_url)
-    command = 'sudo csysdig -m {0} marathon.app.id="{1}" -v mesos_tasks'.format(sysdig_mesos, app_id)
+    command = 'sudo csysdig -m {0} marathon.app.id="/{1}" -v mesos_tasks'.format(sysdig_mesos, app_id)
```
Rob-Johnson: heh. marathon--

paasta_tools/cli/utils.py
```diff
@@ -742,3 +744,52 @@ def pick_slave_from_status(status, host=None):
     else:
         slaves = status.marathon.slaves
         return slaves[0]
+
+
+class TaskNotFound(Exception):
```
Rob-Johnson: probably a better place for this - mesos tools?

paasta_tools/mesos_tools.py
```diff
@@ -115,20 +115,32 @@ def get_current_tasks(job_id):
     return get_mesos_master().tasks(fltr=job_id, active_only=False)
 
 
+def is_task_running(task):
+    return task['state'] == 'TASK_RUNNING'
```
Rob-Johnson: how about making this an attribute on the class?

paasta_tools/mesos_tools.py
```diff
@@ -115,20 +115,32 @@ def get_current_tasks(job_id):
...
+
+
+def is_task_not_running(task):
+    return task['state'] != 'TASK_RUNNING'
```
Rob-Johnson: just use ``if not is_task_running``  - no need for this function

paasta_tools/mesos_tools.py
```diff
@@ -115,20 +115,32 @@ def get_current_tasks(job_id):
...
+
+
+def filter_tasks(tasks, filter_function, **kwargs):
+    return [task for task in tasks if filter_function(task, **kwargs)]
```
Rob-Johnson: this is ``filter`` but with the arguments reorganised and the nomenclature around tasks? you could rewrite this with s/tasks/things & s/task/thing, and just have the ``filter`` function. I'd ditch this if I were you.

paasta_tools/mesos_tools.py
```diff
@@ -655,21 +667,34 @@ def terminate_framework(framework_id):
...
-    container_name = "mesos-{0}.{1}".format(task.executor['tasks'][0]['slave_id'],
-                                            task.executor['container'])
-    return container_name
+    if slave_hostname:
+        tasks = filter_tasks(tasks, filter_task_by_hostname, hostname=slave_hostname)
+    return tasks
+
+
+def get_task(task_id, app_id=''):
+    tasks = get_running_tasks_from_active_frameworks(app_id)
```
Rob-Johnson: just pass the full task id in here and ditch the rest of the filtering

paasta_tools/mesos_tools.py
```diff
@@ -655,21 +667,34 @@ def terminate_framework(framework_id):
...
+        raise TooManyTasks("Found more than one task with id: {0}, this should not happen!".format(task_id))
+    return tasks[0]
+
+
+def filter_task_by_task_id(task, task_id):
+    return task['id'] == task_id
+
+
+def filter_task_by_hostname(task, hostname):
+    return hostname in task.slave['hostname']
```
Rob-Johnson: why ``in`` rather than ``==``?

paasta_tools/api/views/instance.py
```diff
@@ -113,3 +116,54 @@ def instance_status(request):
...
+
+
+@view_config(route_name='service.instance.tasks.task', request_method='GET', renderer='json')
+def instance_task(request):
+    status = instance_status(request)
+    task_id = request.swagger_data.get('task_id', None)
+    verbose = request.swagger_data.get('verbose', False)
+    try:
+        mstatus = status['marathon']
+    except KeyError:
```
mattmb: 👍 this sounds good. As discussed, I'm going to tackle this in a separate branch and add a basic `paasta status` implementation too

paasta_tools/cli/cmds/docker_inspect.py
```diff
@@ -30,11 +31,15 @@ def add_subparser(subparsers):
...
-                                   host=args.host)
-    container = get_container_name(status.marathon.app_id, slave_hostname=slave, task_id=args.mesos_id)
+    try:
+        task = get_task_from_instance(cluster=args.cluster,
+                                      service=args.service,
+                                      instance=args.instance,
+                                      slave_hostname=args.host,
+                                      task_id=args.mesos_id)
+    except TaskNotFound:
+        sys.exit(1)
```
mattmb: there's some logging before the exception is thrown, so not much more to say here.

paasta_tools/cli/utils.py
mattmb: I think I may have even declared this in two places, I'll check!

paasta_tools/mesos_tools.py
mattmb: need more ☕️

paasta_tools/mesos_tools.py
mattmb: Nice 👍 

I'll switch it out

paasta_tools/mesos_tools.py
mattmb: I think what I actually mean is `.startswith` (trying to support users typing short hostnames).

paasta_tools/mesos_tools.py
mattmb: I think I want to leave this because it has a desired side effect: if you pass the wrong task_id for the service and instance then this will return an empty list and you get a useful error message.

paasta_tools/cli/utils.py
mattmb: I'm renaming to `PaastaTaskNotFound` as technically the task is a different object to the one mesos-cli produces.

paasta_tools/mesos_tools.py
mattmb: I'm going to leave this for now because I don't want to make this change at the same time as everything else.

PR #824, opened 2016-10-06
Support multiple scribe environments per cluster

mattmb:
This allows for the case when you have multiple scribe environments per
cluster. The internal ticket is PAASTA-6744.

This change will need to be made in conjunction with a change to the
config files that define the cluster map.


mattmb:
It turns out we totally don't need this because we already aggregate elsewhere!

PR #825, opened 2016-10-06
added support for adhoc jobs

mjksmith:
I still have tests to write but this can start being reviewed

Example of running an adhoc job:

With soa configs adhoc-CLUSTERNAME.yaml:
```
do_all_the_things:
  cpus: 1
  mem: 1024
  disk: 1024
  cmd: echo "Hello world"
  deploy_group: prod.non_canary
```
```
(py27)matts@hostname:~/paasta (adhoc_jobs) $ python paasta_tools/cli/cli.py local-run -s example_happyhour -i do_all_the_things -y ~/soa-configs --pull -c CLUSTERNAME
Please wait while the image (docker:443/services-example_happyhour:paasta-DEADBEEF) is pulled...
--- CLIP ---
You're running a container in non-interactive mode.
This is how Mesos runs containers.
Note that some programs behave differently when running with no
tty attached (as your program is about to run).

Running docker command:
docker run --long --docker --command --here sh -c 'echo "Hello world"'

Your service is now running! Tailing stdout and stderr:
Hello world

Stopping and removing the old container 74a8174fc04b3f699b15e5893af491529eb19103668fe0047340e39cb8b43d79...
(Please wait or you may leave an orphaned container.)
...done
```

Example of using the 'interactive' instance:
```
(py27)matts@hostname:~/example_happyhour (master) $ paasta local-run
PaaSTA on this machine has not been configured with a default cluster.
In the future this behaviour will be depricated and you'll have to specify a cluster.
For now the default cluster is being set to 'testland'.
docker build -t paasta-local-run-example_happyhour-matts .
--- CLIP ---
Warning! You're running a container in interactive mode.
This is *NOT* how Mesos runs containers.
To run the container exactly as Mesos does, omit the -I flag.

Running docker command:
docker run --long --docker --command --here sh -c bash
root@dockerhost:/#
```


paasta_tools/cli/cmds/local_run.py
```diff
@@ -617,32 +625,47 @@ def configure_and_run_docker_container(
...
+                        'cpus': 1,
+                        'mem': 1024,
+                        'disk': 1024,
+                    },
+                    branch_dict={},
+                )
+                interactive = True
+            else:
+                sys.stderr.write(PaastaColors.red(
+                    "Error: you cannot use the default 'interactive' image with 'pull_image' specified.\n"))
```
solarkennedy: Aw, this was kinda the whole point of this :(

I guess this is "hard" because you don't know what deploy group or sha to use. 

Hmm....

paasta_tools/cli/cmds/local_run.py
```diff
@@ -712,9 +735,30 @@ def paasta_local_run(args):
...
+    if args.cluster:
+        cluster = args.cluster
+    else:
+        try:
+            cluster = local_run_config['default_cluster']
+        except KeyError:
+            sys.stderr.write(PaastaColors.yellow(
+                "PaaSTA on this machine has not been configured with a default cluster.\n"
+                "For now the default cluster is being set to 'testopia'.\n"))
+            cluster = 'testopia'
```
solarkennedy: No! Just, bomb out, its ok at this point

mjksmith: This was just here in case the puppet configs didn't ship before the paasta changes. I'll remove it.

paasta_tools/cli/cmds/local_run.py
solarkennedy: Ehhh crap. I guess I hadn't thought about how we would do this.

I guess we have to make a heuristic guesser  :(
Pick one of the deploy groups that any other instances are using that are defined in the same cluster.

paasta_tools/utils.py
```diff
@@ -1657,3 +1655,20 @@ def mean(iterable):
...
     return sum(iterable) / len(iterable)
+
+
+class NoTtyError(Exception):
+    pass
+
+
+def prompt_pick_one(sequence):
+    if not sys.stdin.isatty():
+        raise NoTtyError
```
solarkennedy: I understand the need to raise, but sometimes from a user perspective, a stacktrace isn't the right thing to see.

If there is nothing that expects to ever actually catch this, I would prefer it just return None when there is no tty to choose and then let calling code deal with None just like someone quit.

mjksmith: Right now everything that's using this method catches the `NoTtyError` and prints something helpful. Returning None is what happens when the user quits the prompt by typing 'Quit'.

paasta_tools/adhoc_tools.py
```diff
@@ -0,0 +1,110 @@
...
+            service=service,
+            instance='interactive',
+            cluster=cluster,
+            config_dict={},
+            branch_dict={},
+        )
+        deployments_json = load_v2_deployments_json(service, soa_dir=soa_dir)
+        try:
+            deploy_group = prompt_pick_one(deployments_json['deployments'].keys())
+        except NoTtyError:
```
mjksmith: I had a big response typed out but I realized you were right so I changed it.

paasta_tools/marathon_tools.py
```diff
@@ -959,7 +959,10 @@ def get_healthcheck_for_instance(service, instance, service_manifest, random_por
     or (None, None) if no healthcheck
     """
     smartstack_config = load_service_namespace_config(service, instance, soa_dir)
-    mode = service_manifest.get_healthcheck_mode(smartstack_config)
+    try:
+        mode = service_manifest.get_healthcheck_mode(smartstack_config)
```
solarkennedy: What do you think about having a `get_healthcheck_mode` func in the adhoc config that just returns None?

paasta_tools/cli/cmds/local_run.py
```diff
@@ -617,32 +624,33 @@ def configure_and_run_docker_container(
...
+            instance_config = get_instance_config(
+                service=service,
+                instance=instance,
+                cluster=cluster,
+                load_deployments=load_deployments,
+                soa_dir=soa_dir,
+            )
+    except NoConfigurationForServiceError as e:
+        sys.stderr.write(str(e) + '\n')
+        return
```
solarkennedy: return non-zero?

paasta_tools/utils.py
```diff
@@ -1402,6 +1390,16 @@ def get_branch_dict(self, service, branch):
         full_branch = '%s:paasta-%s' % (service, branch)
         return self.get(full_branch, {})
 
+    def get_branch_dict_v2(self, service, branch, deploy_group):
```
solarkennedy: I'm suprised to see this, not a rebase thing?

mjksmith: No I forgot to include this (since nothing was using the v2 deployments until now).

paasta_tools/cli/cmds/local_run.py
mjksmith: Nothing that calls this method is catching the return value and doing anything with it. I can change that?

paasta_tools/adhoc_tools.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+def load_adhoc_job_config(service, instance, cluster, load_deployments=True, soa_dir=DEFAULT_SOA_DIR):
+    general_config = service_configuration_lib.read_service_configuration(
+        service,
+        soa_dir=soa_dir
+    )
+    adhoc_conf_file = "adhoc-%s" % cluster
+    log.info("Reading adhoc configuration file: %s.yaml", adhoc_conf_file)
+    instance_configs = service_configuration_lib.read_extra_service_information(
```
solarkennedy: Singular right? It is just one config?

paasta_tools/adhoc_tools.py
```diff
@@ -0,0 +1,109 @@
...
+        service,
```
solarkennedy: kwargs just in case

paasta_tools/adhoc_tools.py
mjksmith: Plural since it loads every instance config in a file.

I copied this straight from `marathon_tools`.

paasta_tools/adhoc_tools.py
```diff
@@ -0,0 +1,109 @@
...
+            service=service,
+            config_dict=config_dict,
+            branch_dict=branch_dict,
+        )
+
+
+def get_default_interactive_config(service, cluster, soa_dir):
+    default_job_config = {
+        'cpus': 1,
+        'mem': 1024,
```
mjksmith: I.... don't even know if local-run enforces these limits

paasta_tools/cli/utils.py
```diff
@@ -796,6 +759,19 @@ def get_instance_configs_for_service(service, soa_dir):
                 soa_dir=soa_dir,
                 load_deployments=False,
             )
+        for _, instance in get_service_instance_list(
```
solarkennedy: Hopefully this doesn't provide an adhoc job to  something that isn't expecting it. (status? validate? check?)

paasta_tools/utils.py
```diff
@@ -1657,3 +1661,27 @@ def mean(iterable):
...
+
+    chooser = choice.Menu(choices, global_actions=global_actions)
+    chooser.title = 'Please pick a {choosing} from the choices below:'.format(choosing=choosing)
+    try:
+        result = chooser.ask()
+    except KeyboardInterrupt:
+        sys.stdout.write('\n')
+        sys.exit(1)
+
+    if isinstance(result, tuple) and result[1] == QUIT_ACTION:
```
solarkennedy: Are there situations where the results is a tuple but not Quit action?

I don't want to give callers tuples if they are expecting a singleton

mjksmith: Tuples result from passing a tuple in the sequence or by specifying additional `extra_actions`. If we're just passing lists of strings I don't expect the result to ever be a tuple except for when the user quits.

mjksmith:
I'll ship this on Monday and pin `paasta-tools` to `0.54.0` on stage and prod.

PR #827, opened 2016-10-07
add a check to local run to assert healthcheck body length

### LINKED ISSUE #826: Add a check to local run for some content in the response body.
We found a bug in Marathon yesterday that meant that it would never mark a task healthy if the response body was empty (even if the status line of the HTTP response indicated 200 OK).

Let's replicate this behaviour in local run, only returning healthy when the response body has some contents.

Rob-Johnson:
closes #826 

I'm going to hold off on this until I confirm this definitely was the problem

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -83,6 +83,9 @@ def perform_http_healthcheck(url, timeout):
             "Multiple content-type headers detected in response."
             " The Mesos healthcheck system will treat this as a failure!"))
         return (False, "http request succeeded, code %d" % res.status_code)
+    # check if there is response content in the body
+    if len(res.content) < 1:
+        return (False, "0 length response body detected.")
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -83,6 +83,9 @@ def perform_http_healthcheck(url, timeout):
             "Multiple content-type headers detected in response."
             " The Mesos healthcheck system will treat this as a failure!"))
         return (False, "http request succeeded, code %d" % res.status_code)
+    # check if there is response content in the body
+    if len(res.content) < 1:
```
mjksmith: `if not res.content` is the PEP 8-compliant way to check if a sequence is empty.

asottile:
Is a 200 with an empty body not a valid 200?  I've certainly implemented a trivial service which simply does `touch /web-root/status` when ready

Rob-Johnson:
I'm closing this - I was never able to prove that this was the cause of the issue. I'll reopen if we see a reoccurrence.

PR #829, opened 2016-10-08
added uwsgi autoscaler

mjksmith:


solarkennedy:
cc @bplotnick

tests/autoscaling/test_autoscaling_service_lib.py
```diff
@@ -315,13 +329,51 @@ def test_http_metrics_provider_no_data():
...
+    with mock.patch(
+        'paasta_tools.autoscaling.autoscaling_service_lib.get_json_body_from_service',
+        autospec=True,
+    ) as mock_get_json_body_from_service:
+        mock_get_json_body_from_service.return_value = {
+            'workers': [
+                {'status': 'idle'},
+                {'status': 'not-idle'},
+                {'status': 'not-idle'},
+                {'status': 'not-idle'},
```
bplotnick: I _believe_ the only two states are `busy` and `idle`. however in previous code, I have just used `busy` and `!= busy`. Either way, `not-idle` is not a valid state

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -168,17 +181,56 @@ def http_metrics_provider(marathon_service_config, marathon_tasks, mesos_tasks,
...
+    :param marathon_tasks: Marathon tasks to get data from
+    :param endpoint: The http endpoint to get the uwsgi stats from
+
+    :returns: the service's average utilization, from 0 to 1
+    """
+
+    def uwsgi_mapper(json):
+        workers = json['workers']
+        utilization = [1.0 if worker['status'] != 'idle' else 0.0 for worker in workers]
+        return average(utilization)
```
bplotnick: There are two subtle points here that you are missing:

The first is that during uwsgi startup (and i believe in the case where the _master_ reloads), all workers may not have spawned yet. In order to avoid a race condition, you may want to check that all workers are in the `accepting` state before you can say anything conclusively about the utilization

```python
for worker in workers:
    if worker['accepting'] == 0:
        ...do something to indicate that it isn't ready
```

The second subtlety is that uwsgi has the concept of async "[cores](http://uwsgi-docs.readthedocs.io/en/latest/Async.html)", which will not contribute to your utilization calculation... To be honest, I don't know if they should. I really don't fully understand them (I think you can treat them as threads), but I think it is safe to ignore them for now.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -77,6 +77,13 @@ class MetricsProviderNoDataError(ValueError):
     pass
 
 
+def average(iterable):
+    """
+    Returns the average value of an iterable
+    """
+    return sum(iterable) / len(iterable)
```
Rob-Johnson: wrap this in a try/except and check for ``ZeroDivisionError`` - could happen if you pass this to an empty list.

Also, can you move this to utils.py?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -168,17 +181,56 @@ def http_metrics_provider(marathon_service_config, marathon_tasks, mesos_tasks,
...
     for task in marathon_tasks:
         try:
-            endpoint_utilization = requests.get(
-                'http://%s:%s/%s' % (task.host, task.ports[0], endpoint),
-                headers={'User-Agent': get_user_agent()}
-            ).json()['utilization']
-            utilization.append(float(endpoint_utilization))
+            utilization.append(json_mapper(get_json_body_from_service(task.host, task.ports[0], endpoint)))
         except Exception:
             pass
```
Rob-Johnson: log.debug here when you fail maybe?

solarkennedy: Yes please. When debugging this thing we are just going to wonder "why" it isn't autoscaling up or down when "clearly" something is wrong, and the debug line will say "not valid json" or "couldn't connect", etc.


Also, if there is an exception, shouldn't it count against it against it?

For example, if all workers are flooded, the utilization *should* be 100%, but in practice we won't get any output from the endpoints and we'll get... 0 or [].

mjksmith: I don't have exceptions count towards the final average, so we just pretend like tasks that either don't return results or return invalid results. If no tasks return or product valid output, a `MetricsProviderNoDataError` is raised and logged.

solarkennedy: Right, but that is the problem:
* Expected: That when all workers are full the autoscaler scales things up
* Actual: When all workers are full the autoscaler raises and doesn't scale anything

docs/source/autoscaling.rst
```diff
@@ -75,21 +75,41 @@ Metrics providers
...
-  Autoscaling parameters:
+  Extra parameters:
 
-  :endpoint: the path to perform the HTTP request on (the requested URL will be ``http://$HOST:$PORT/endpoint``). Defaults to 'status'.
+  :endpoint:
+    the path to perform the HTTP request on (the requested URL will be
+    ``http://$HOST:$PORT/$endpoint``). Defaults to 'status'.
+
+:uwsgi:
+  Makes a request on a HTTP endpoint on your service. Extects a response with a
```
mxmul: s/Extects/Expects/

paasta_tools/autoscaling/autoscaling_service_lib.py
mjksmith: Ahh, I might have different logic for a timeout exception then.

bplotnick:
lg2m!

paasta_tools/autoscaling/autoscaling_service_lib.py
mjksmith: @solarkennedy how's this? I made a change to consider tasks that have requests that time out as being fully utilized

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -149,48 +150,99 @@ def clamp_value(number):
...
-            ).json()['utilization']
-            utilization.append(float(endpoint_utilization))
-        except Exception:
-            pass
+            utilization.append(json_mapper(get_json_body_from_service(task.host, task.ports[0], endpoint)))
+        except requests.exceptions.Timeout:
+            # If we time out querying an endpoint, assume the task is fully loaded
+            # This won't trigger in the event of DNS error or when a request is refused
+            # a requests.exception.ConnectionError is raised in those cases
+            utilization.append(1.0)
```
solarkennedy: Can you log.debug on this one as well? 
This is another example of "why is this thing not doing what I think it would do", and logging helps.

PR #831, opened 2016-10-10
Add get_num_deployments contrib script

nhandler:
This script simply outputs the number of marathon deployments. It is designed to be used by other scripts to avoid making disruptive changes while deployments are in progress. Ideally, the script would accept a cluster as an argument, but that would require quite a bit more help as we would need to actually SSH to a box in the specified cluster to access the proper marathon config.


solarkennedy:
Can you prefix this script with `paasta_` ?

Shipit

PR #832, opened 2016-10-10
added request caching for service autoscaler

mjksmith:
This speeds up the autoscaler in our largest cluster by 7x.


nhandler:
Wow, nice job. Ship It!

mattmb:
This is cool 👍 

That said can you make the commit message a bit better before we ship this? I'd like it to state what was making it slow i.e. where does the caching help us.

PR #833, opened 2016-10-11
Added tab-completion where we use deploy_groups

solarkennedy:
A number of things related to deploy groups / rollback / mark-for-deployment that I've addressed while working with the tools:
* Added tab completion wherever I could
* Made mark-for-deployment not require a git url like rollback
* Added a deploy_group column to rollback's suggestions (more to come on that front after this PR)
* `get_instance_config_for_service` => `get_instance_configs_for_service`
* Moved `get_instance_configs_for_service` to utils (can't move it to the big utils due to cross-imports)
* Tried to make the cli short-args more universal without breaking things

Let me know if this is too much and I can split up some things, but some of this has to be shipped together.

### DIFF
--- a/paasta_tools/cli/cmds/get_latest_deployment.py
+++ b/paasta_tools/cli/cmds/get_latest_deployment.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 from paasta_tools.cli.utils import lazy_choices_completer
+from paasta_tools.cli.utils import list_deploy_groups
 from paasta_tools.cli.utils import list_services
 from paasta_tools.cli.utils import PaastaColors
 from paasta_tools.cli.utils import validate_service_name
@@ -33,10 +34,10 @@ def add_subparser(subparsers):
         required=True,
     ).completer = lazy_choices_completer(list_services)
     list_parser.add_argument(
-        '-i', '--deploy-group',
+        '-l', '--deploy-group',
         help='Name of the deploy group which you want to get the latest deployment for.',
         required=True,
-    )
+    ).completer = lazy_choices_completer(list_deploy_groups)
     list_parser.add_argument(
         '-d', '--soa-dir',
         help='A directory from which soa-configs should be read from',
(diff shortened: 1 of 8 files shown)


paasta_tools/cli/utils.py
```diff
@@ -660,20 +661,29 @@ def extract_tags(paasta_tag):
     return regex_match.groupdict() if regex_match else {}
 
 
-def validate_given_deploy_groups(service_deploy_groups, args_deploy_groups):
+def list_deploy_groups(parsed_args=None, service=None, soa_dir=DEFAULT_SOA_DIR, **kwargs):
+    if service is None:
+        service = parsed_args.service or guess_service_name()
+    return set([config.get_deploy_group() for config in get_instance_configs_for_service(
```
mattmb: Optionally this could just be a set comprehension

paasta_tools/cli/utils.py
```diff
@@ -660,20 +661,29 @@ def extract_tags(paasta_tag):
     return regex_match.groupdict() if regex_match else {}
 
 
-def validate_given_deploy_groups(service_deploy_groups, args_deploy_groups):
+def list_deploy_groups(parsed_args=None, service=None, soa_dir=DEFAULT_SOA_DIR, **kwargs):
```
mattmb: Missing a unit test for this function?

solarkennedy:
Rebased and added test.

To answer your question... no, we don't really have itests for these particular tools.

mattmb:
Shipit 🚢 

I guess I'm not *too* worried about the lack of itests. Breaking these cli commands would be a pain but not catastrophic.

paasta_tools/cli/cmds/get_latest_deployment.py
```diff
@@ -33,10 +34,10 @@ def add_subparser(subparsers):
         required=True,
     ).completer = lazy_choices_completer(list_services)
     list_parser.add_argument(
-        '-i', '--deploy-group',
+        '-l', '--deploy-group',
```
solarkennedy: Crap :(

PR #843, opened 2016-10-13
Update mock patch checker to use the AST module

### LINKED ISSUE #840: paasta_tools/contrib/mock_patch_checker.py uses deprecated `compiler` module
This has been deprecated since python2.6.  You should switch to the `ast` module (does basically the same thing).

Note that the `compiler` module is gone in python3

mjksmith:
Fixes #840

### DIFF
--- a/paasta_tools/contrib/mock_patch_checker.py
+++ b/paasta_tools/contrib/mock_patch_checker.py
@@ -1,10 +1,9 @@
 #!/usr/bin/env python2.7
-import compiler
+import ast
 import sys
-from compiler.visitor import ASTVisitor
 
 
-class MockChecker(ASTVisitor):
+class MockChecker(ast.NodeVisitor):
     def __init__(self):
         self.errors = 0
         self.init_module_imports()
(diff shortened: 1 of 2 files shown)


solarkennedy:
Whoa, and it caught more non-autospec tests? Shipit.

mjksmith:
Yeah, I found a bug that can occur if a test library does both `import mock` and `from mock import patch`. I'll ship if tests pass.

paasta_tools/contrib/mock_patch_checker.py
```diff
@@ -20,32 +19,48 @@ def check_files(self, files):
...
+        except IOError:
+            print "Error opening filename: %s" % filename
             return
-        compiler.walk(ast, self)
         self.init_module_imports()
+        self.visit(file_ast)
 
-    def visitImport(self, node):
-        if (name for name in node.names if 'mock' in name):
+    def CallUsesMockPatch(self, node):
```
asottile: this should be `def call_uses_mock_patch(self, node):` or even `def _call_uses_mock_patch(...):`

paasta_tools/contrib/mock_patch_checker.py
```diff
@@ -20,32 +19,48 @@ def check_files(self, files):
...
-                        [arg for arg in node.args if isinstance(arg, compiler.ast.Keyword) and arg.name == 'autospec']):
+            if (self.imported_patch and self.CallUsesPatch(node)) or \
+                    (self.imported_mock and self.CallUsesMockPatch(node)):
+                if not any([keyword for keyword in node.keywords if keyword.arg == 'autospec']):
                     print "%s:%d: Found a mock without an autospec!" % (self.current_filename, node.lineno)
                     self.errors += 1
         except AttributeError:
             pass
-        for child in node.getChildNodes():
+        for child in ast.iter_child_nodes(node):
```
asottile: you want `self.generic_visit(node)`

paasta_tools/contrib/mock_patch_checker.py
```diff
@@ -20,32 +19,48 @@ def check_files(self, files):
...
+            return False
+
+    def CallUsesPatch(self, node):
+        try:
+            return node.func.id == 'patch'
+        except AttributeError:
+            return False
+
+    def visit_Import(self, node):
+        if (name for name in node.names if 'mock' == name.name):
```
asottile: this is trivially true:

```python
>>> bool(x for x in [])
True
```

mjksmith:
@asottile your changes are on master

PR #845, opened 2016-10-14
Try not scaling down before deleting marathon apps.

solarkennedy:
Let's give it a shot. Relates to https://github.com/mesosphere/marathon/issues/4581


mattmb:
I think this is a good idea, let's ship very carefully though!

PR #846, opened 2016-10-14
Added get_currently_deployed_sha and better rollback hints

solarkennedy:
I'm noticing that `rollback`, `mark-for-deployment`, `generate-deployments-for-service`, and `get-latest-deployment`, are all sharing or cross-importing. So I think I would like to start a `deployment_utils` to store this kind of thing.

Also, I'm enhancing `mark-for-deployment` to give very obvious rollback instructions, to help bring it up to parity with the usability of our existing yelp-main scripts.

Although this is "slow" because getting the latest sha currently uses git. @mjksmith is working on making deployments.json store this information directly.


solarkennedy:
I'm discarding this.

PR #847, opened 2016-10-15
Breaking: Removed multiple deploy-group options from rollback and mak…

solarkennedy:
…e it wait for the deployment

This is a large change, but makes it simpler for the code, and hopefully more robust.

It does mean that you can't "roll back everything" in a panic, and instead makes you specify the deploy group.

Hopefully this is helped by the fact that we have:
* tab completion
* It suggested deploy groups for you
* @mjksmith is going to add a "chooser" library so we can just ask the user, so they don't have to "remember" what their deploy groups are called

In return, it makes it easier for us to block after the rollback, which I think 99% of our users want and expect when they run a rollback command.

### DIFF
--- a/paasta_tools/cli/cmds/rollback.py
+++ b/paasta_tools/cli/cmds/rollback.py
@@ -12,23 +12,27 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import sys
+
 from humanize import naturaltime
 
 from paasta_tools.cli.cmds.mark_for_deployment import mark_for_deployment
+from paasta_tools.cli.cmds.mark_for_deployment import wait_for_deployment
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/rollback.py
```diff
@@ -79,54 +82,54 @@ def add_subparser(subparsers):
...
-        service=service,
-        soa_dir=soa_dir,
-    )
-    deploy_groups, _ = validate_given_deploy_groups(all_deploy_groups, deploy_groups)
     previously_deployed_shas = {}
     for ref, sha in list_remote_refs(git_url).items():
         regex_match = extract_tags(ref)
         try:
-            deploy_group = regex_match['deploy_group']
+            dg = regex_match['deploy_group']
```
mjksmith: Is there a more self-documenting name instead of `dg`?  Something like `ref_deploy_group`? It took me a while to realize what this code was doing.

solarkennedy:
I'm discarding this.

PR #848, opened 2016-10-15
store docker image by deploy group

mjksmith:
We're starting to run into problems from storing docker image by git branch instead of by deploy group (adhoc batches, paasta get_currently_deployed_sha).

As far as I remember we kept this in when we were switching to deploy groups because it allowed per-serviceinstance rollbacks and didn't involve changing `deployments.json` so it was easy to validate that deploy groups were working (just diff old and new `deployments.json` and expect no changes).

After this, a second change needs to be pushed that removes the v1 `deployments.json` and switches the rest of paasta over to v2.

### DIFF
--- a/paasta_tools/generate_deployments_for_service.py
+++ b/paasta_tools/generate_deployments_for_service.py
@@ -124,6 +124,10 @@ def get_deploy_group_mappings(soa_dir, service, old_mappings):
       the other properties of this app have not changed.
     """
     mappings = {}
+    v2_mappings = {
+        'deployments': {},
+        'controls': {},
+    }
 
     service_configs = get_instance_configs_for_service(
         soa_dir=soa_dir,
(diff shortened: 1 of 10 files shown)


paasta_tools/generate_deployments_for_service.py
```diff
@@ -158,7 +164,9 @@ def get_deploy_group_mappings(soa_dir, service, old_mappings):
             )
             mapping['desired_state'] = desired_state
             mapping['force_bounce'] = force_bounce
-    return mappings
+            v2_mappings['controls'].setdefault(control_branch_alias, {})['desired_state'] = desired_state
+            v2_mappings['controls'][control_branch_alias]['force_bounce'] = force_bounce
```
mjksmith: How about `service:cluster.instance`? Right now it's `service:paasta-cluster.instance`.

solarkennedy: How about... `cluster:service.instance`?

`service.instance` is much more consistent across the codebase at least.

solarkennedy: Ah crap. Na, disregard.

PR #851, opened 2016-10-18
dont jump the gun adopting v2 deployments

mjksmith:
oops


mjksmith:
Addendum: shouldn't itests have caught this?

solarkennedy:
We don't have any itests for this

solarkennedy:
ship. I think having the paasta- tag leader in the git tag is a "good thing".  It is just in our config files we don't really "need" it.

PR #852, opened 2016-10-18
made tox use venv_update

mjksmith:
Switching between branches sucks right now. I develop on one branch, switch to another, keep developing and run my tests. My tests fail because I don't have the correct dependancies, so I have to do `tox -r` for each of my environments, which takes forever. Then when I switch branches, I have to rebuild all of my environments again. Using venv_update makes everything faster, which makes me 😄 .

http://venv-update.readthedocs.io/en/stable/index.html

This currently increases the time required to run `tox` from ~7 seconds to ~8.8 seconds. Running `time tox` a bunch and averaging the result shows an increase by about 1.8 seconds. It also slightly increases the time it takes to run tests in travis (a couple of percents). However it reduces the time it takes to regenerate a venv (say, if you ran `tox` with a virtualenv activated) from about 2m45s to 35s and reduces the time it takes to switch between repos with different dependancies from about 2m45s to 10s (when only one dependency differs).

I'll look for ways to avoid slowing down vanilla `tox` runs.


tox.ini
```diff
@@ -67,33 +60,28 @@ commands =
...
-deps =
-    {[testenv]deps}
-    behave==1.2.4
-    behave-pytest==0.1.1
-    # This .whl is created in yelp_package/dockerfiles/trusty/Dockerfile
-    /root/mesos.executor-1.0.1-py27-none-any.whl
-    /root/mesos.scheduler-1.0.1-py27-none-any.whl
-    /root/mesos.native-1.0.1-py27-none-any.whl
 commands =
+    {toxinidir}/bin/venv-update venv= {envdir} install= -r {toxinidir}/requirements-dev.txt -e {toxinidir} \
```
mjksmith: I've realized this is a problem when running `make_trusty` and `make_xenial` so I'll have to figure out why this is the case (or hardcode it everywhere, which I don't really want to do).

PR #853, opened 2016-10-18
Pin pycparser

mattmb:
It's broken again:
https://github.com/eliben/pycparser/issues/151

2.1.4 apparently still works so pinning it to see if Travis will go green


Rob-Johnson:
ship if tests pass

mattmb:
Cancel this. The bug is fixed upstream and as it turns out pip + requirements.txt is useless for trying to pin a sub-dependency as you can see from the itest logs...

PR #837, opened 2016-10-13
Made mark-for-deployment way more verbose and have a progress bar

solarkennedy:
This adds a progress bar and *lots* more debug logging available with -v.

I dare say at yelp we should make -v the default in jenkins. It isn't like humans are looking at it all day, but when it *does* fail, we will want to see the debug logs I've added here. 

Here is what it looks like with verbose mode:
```
$ paasta mark-for-deployment --wait-for-deployment -c c19f9731c2665b4508d69f6263a8f502ad247489 -s paasta_canary -l everything -u git@git.yelpcorp.com:services/paasta_canary   -v
Marked c19f9731c2665b4508d69f6263a8f502ad247489 in for deployment in deploy group everything
Waiting for deployment of c19f9731c2665b4508d69f6263a8f502ad247489 for 'everything' complete...
  0% ( 0 of 18) |                                                                                                        | Elapsed Time: 0:00:00 ETA:  --:--:--DEBUG:paasta_tools.cli.cmds.mark_for_deployment:Inspecting the deployment status of paasta_canary.main on nova-prod
DEBUG:paasta_tools.cli.cmds.mark_for_deployment:Inspecting the deployment status of paasta_canary.sensu_canary on nova-prod
DEBUG:paasta_tools.cli.cmds.mark_for_deployment:paasta_canary.main on nova-prod is still bouncing, 2 versions running
DEBUG:paasta_tools.cli.cmds.mark_for_deployment:paasta_canary.sensu_canary in nova-prod is not a Marathon job. Marked as deployed.
Deploy to nova-prod complete! (main, sensu_canary)
 11% ( 2 of 18) |###########                                                                                               | Elapsed Time: 0:00:06 ETA: 0:00:52DEBUG:paasta_tools.cli.cmds.mark_for_deployment:Inspecting the deployment status of paasta_canary.main on everywhere-testopia
DEBUG:paasta_tools.cli.cmds.mark_for_deployment:paasta_canary.main on everywhere-testopia is still bouncing, 2 versions running
Deploy to everywhere-testopia complete! (main)
 16% ( 3 of 18) |#################                                                                                         | Elapsed Time: 0:00:09 ETA: 0:00:46DEBUG:paasta_tools.cli.cmds.mark_for_deployment:Inspecting the deployment status of paasta_canary.main on pnw-devc
DEBUG:paasta_tools.cli.cmds.mark_for_deployment:Inspecting the deployment status of paasta_canary.sensu_canary on pnw-devc
[...]

PR #854, opened 2016-10-19
Fix waiting on mark for deployment

mattmb:
@solarkennedy this was a subtle one! We missed how `with progressbar.ProgressBar()` works, it doesn't *require* `bar.finish()` to be called to exit. Unfortunately the tests didn't catch it either because they were exiting after the first `time.sleep` call.

---
We introduced a bug in https://github.com/Yelp/paasta/pull/837 where
we would exit from wait for deployment after the first check of all the
clusters and instances, regardless of status.

This adds back the while loop which is neccessary even with the `with
progressbar` and also improves the test case which would have caught
this regression.


solarkennedy:
ship

PR #860, opened 2016-10-21
Use marathon 1.4.0-snap18 in our itests

nhandler:
This is a test to see if marathon 1.4.0-snap18 behaves better than 1.3.0.

The bintray debs were built by our internal package building infrastructure and then manually uploaded. If we decide to stick with this unreleased version of marathon, I'll get automatic uploads to bintray setup.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,7 +20,7 @@ isodate==0.5.1
 jsonschema[format]==2.5.1
 kazoo==2.2
 lockfile==0.9.1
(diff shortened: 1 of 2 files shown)


solarkennedy:
Nice

solarkennedy:
I built a bespoke snap19 without doing the concat trick. :(
Giving it a shot.

solarkennedy:
I'll work on getting marathon-python updated.

PR #861, opened 2016-10-24
Use v2 deployments

mjksmith:
This removes the v1 deployments json and make all of `paasta_tools` use the new v2 format.

### DIFF
--- a/general_itests/steps/deployments_json_steps.py
+++ b/general_itests/steps/deployments_json_steps.py
@@ -36,7 +36,7 @@ from paasta_tools.cli.cmds.start_stop_restart import paasta_stop
 from paasta_tools.utils import format_tag
 from paasta_tools.utils import format_timestamp
 from paasta_tools.utils import get_paasta_tag_from_deploy_group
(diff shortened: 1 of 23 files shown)


mjksmith:
This broke since we reverted adhoc instances. I'll clean it up once that gets shipped again.

solarkennedy:
Rebase please :(

mjksmith:
@solarkennedy done

solarkennedy:
I really hope nothing is using v1, this will certainly expose that.

For a final test before shipping can you confirm that only the v1 dict goes away by doing a generate-deployments on yelp-main before and after activating a virtualenv?

If the only difference is the v1 dict, then ship.

mjksmith:
A code search seems to show nothing uses v1 deployments unless they access it in a very weird way. Exempli gratia:
```
deployments_file_name = 'deployments'
deployments_file = '%s.json' % deployments_file_name
with open(deployments_file, 'r'):
    # blah
```

mjksmith:
@solarkennedy confirmed the only difference is the absence of the v1 dict.

PR #862, opened 2016-10-24
Update `makefile_responds_to` to handle common recursive Makefile pattern

Uberi:
`paasta check` uses the `makefile_responds_to` function to check if the Makefile responds to a given target. However, the way that function works sometimes gives false negatives - the check fails sometimes when the Makefile actually does respond to a given target.

When a Makefile includes `$(MAKE) -C SOME_DIR $(MFLAGS)` and `SOME_DIR/Makefile` is a makefile, that line gets executed even when we specify the `-q` flag - see https://www.gnu.org/software/make/manual/html_node/MAKE-Variable.html#MAKE-Variable.

Here, the `-q` flag is passed into the inner MAKE invocation via `$(MFLAGS)`, which causes the inner invocation to exit with 0 or 1 if successful. If the inner MAKE exits with 1, the original MAKE gives an error.

In other words, with this pattern, `make test`  works fine, but `make -q test` fails with Error 1.

To get around this, we can use `make --dry-run` instead of `make -q`. This ensures the Makefile is checked more like it would be if we were actually running make normally.

### DIFF
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -116,19 +116,18 @@ def docker_check():
 
 
 def makefile_responds_to(target):
-    """Runs `make -q <target>` to detect if a makefile responds to the
+    """Runs `make --dry-run <target>` to detect if a makefile responds to the
     specified target."""
-    cmd = 'make -q %s' % target
+    # `make -q` is commonly used to detect whether a Makefile responds to a specified target.
+    # However, this doesn't work when the Makefile includes the common pattern `$(MAKE) -c SOMETHING $(MFLAGS)`,
+    # and `SOMETHING/Makefile` needs to be updated; the `$(MAKE)` means that line executes even in question mode,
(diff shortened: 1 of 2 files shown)


solarkennedy:
lgtm, I'm not sure why I did -q in the first place. Can you address the failing test?

paasta_tools/cli/cmds/check.py
```diff
@@ -116,19 +116,18 @@ def docker_check():
 
 
 def makefile_responds_to(target):
-    """Runs `make -q <target>` to detect if a makefile responds to the
+    """Runs `make --dry-run <target>` to detect if a makefile responds to the
     specified target."""
-    cmd = 'make -q %s' % target
+    # `make -q` is commonly used to detect whether a Makefile responds to a specified target.
```
solarkennedy: I wouldn't bother with all these comments debating why not use -q, I think --dry-run is obvious enough

solarkennedy:
Thanks!

PR #865, opened 2016-10-25
Skip unimplemented autoscaler resources

mattmb:
This makes it safe to put configs in place for autoscaling groups before
we have the code to scale them.


solarkennedy:
ship

PR #867, opened 2016-10-26
Sleep to throttle AWS API calls to SFR API

mattmb:
Obviously I don't like this as a "solution" but I am interested to see
if it makes a difference in practice.


mattmb:
Links to yelp internal OPSIMP-1375

PR #868, opened 2016-10-26
Fix downscaling cancelled SFRs

mattmb:
I thought I'd already shipped this but I guess not!

We are conservative when downscaling, sometimes we won't kill an
instance if it would take us below the target.

If the SFR is cancelled we need to make sure we kill at least one
instance so that we don't get stuck never killing an instance.


solarkennedy:
I think this is right.

mattmb:
I'm going to take that as a shipit ;-)

(I have also tested this in mesosstage)

PR #869, opened 2016-10-31
fix typo

Rob-Johnson:
@tub

### DIFF
--- a/paasta_tools/paasta_maintenance.py
+++ b/paasta_tools/paasta_maintenance.py
@@ -66,7 +66,7 @@ def parse_args():
             'undrain',
             'up',
         ],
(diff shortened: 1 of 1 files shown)


paasta_tools/paasta_maintenance.py
```diff
@@ -66,7 +66,7 @@ def parse_args():
             'undrain',
             'up',
         ],
-        help="Action to perform on the speicifed hosts",
+        help="Action to perform on the specifed hosts",
```
nhandler: This should be `specified` :)

tub: *sigh* - spelling is hard :(

PR #870, opened 2016-10-31
Install sphinxcontrib-programoutput from github

mattmb:
This module seems to have disappeared from pypi...

This installs it from github for now. To keep the build passing.


mattmb:
@nhandler it looks like we only use this in a couple of places in the docs, maybe we should change that given that it's not maintained?

PR #871, opened 2016-10-31
Make check_chronos_jobs ignore fresh jobs

solarkennedy:
We should not send OK for when a chronos job is "fresh". They are "fresh" all the time because we update the job all the time.

Instead this state means we just shouldn't send an event, because we don't have any information. We shouldn't report it as failing (we don't know that yet) nor should we send a resolve (as it hasn't actually run properly yet)

So I think the right thing to do is simply skip and move on until we have data.

There are a lot of ways to do this though, let me know if you don't like this approach.

Internal ticket PAASTA-6658

### DIFF
--- a/paasta_tools/check_chronos_jobs.py
+++ b/paasta_tools/check_chronos_jobs.py
@@ -73,19 +73,18 @@ def last_run_state_for_jobs(jobs):
 def sensu_event_for_last_run_state(state):
     """
     Given a LastRunState, return a corresponding sensu event type.
-    Raise a ValueError in the case that the state is not valid.
+    Will return None in the case that the job has not run yet,
+    indicating that no Sensu alert should be sent.
+    Raises ValueError in the case that the state is not valid.
     """
-    if state not in (
-        chronos_tools.LastRunState.Fail,
-        chronos_tools.LastRunState.Success,
-        chronos_tools.LastRunState.NotRun,
-    ):
-        raise ValueError('Expected valid LastRunState. Found %s' % state)
-
     if state is chronos_tools.LastRunState.Fail:
         return pysensu_yelp.Status.CRITICAL
-    else:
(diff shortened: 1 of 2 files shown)


paasta_tools/check_chronos_jobs.py
```diff
@@ -200,13 +199,17 @@ def main():
...
-            )
+            try:
+                sensu_output, sensu_status = sensu_message_status_for_jobs(
+                    chronos_job_config=chronos_job_config,
+                    service=service,
+                    instance=instance,
+                    cluster=cluster,
+                    job_state_pairs=job_state_pairs
+                )
+            except ValueError:
```
Rob-Johnson: This hints at the need for some refactoring. We're having to catch an exception that isn't even thrown in this function, but a function *it* depends on. I don't really like relying on that implementation detail for this logic.

maybe instead of this you could do:

```
fresh = job_state_pairs[0][1] == chronos_tools.LastRunState.NotRun if job_state_pairs else False
if not fresh:
   sensu_output, sensu_status = sensu_message_status_for_jobs(
                     chronos_job_config=chronos_job_config,
                     service=service,
                     instance=instance,
                     cluster=cluster,
                     job_state_pairs=job_state_pairs
                 )
      ...
```

I'm not completely happy with this here, but atleast now you're not tightly coupled to the implementation of ```sensu_message_status_for_jobs```

solarkennedy: Would you be less-un-happy if it used "None" as the signal for "Nothing to send to sensu" instead of an exception?

solarkennedy:
How about this?

Rob-Johnson:
eh, I guess so. This is where it's nice to have ``Maybe`` or ``Option`` types, so that the caller is forced to deal with the None case.

Can you add to the docstring that the function will return None in some cases?

PR #872, opened 2016-10-31
Globally catch KeyboardInterrupt and exit 1 on cli commands

solarkennedy:
I can't think of any reason not to do this at this layer.
If we do need to catch this in a more fine-grained way, we can still do that with an inner try/catch.


mjksmith:
LGTM. There are probably some places where we run something in a different process that might not close if we call sys.exit but it's better than spamming Ctrl-C and seeing a bunch of `KeyboardInterrupt` stack traces pop up.

PR #873, opened 2016-11-02
Revert "Revert "added support for adhoc jobs""

mjksmith:
We need to hold off shipping this until we fix all of the Makefiles and jenkins jobs that don't specify `-i`.


solarkennedy:
Ship. I thought you did this already?

PR #877, opened 2016-11-03
Fix stuck downscaling of SFRs

mattmb:
Because we use non integer weights for spots and because we are
conservative when scaling down it is possible for the autoscaler to get
stuck when scaling down SFRs

This applies the same fix applied for cancelled SFRs where we force the
autoscaler to kill at least one instance when it is scaling down.


solarkennedy:
Is there a danger that the autoscaler will always "kill at least one"?

mattmb:
we should be okay on that front because we only reach this function if we've decided that the we need to downscale

PR #879, opened 2016-11-04
Alert service authors when a service is stopped even when asked to stop

solarkennedy:


paasta_tools/marathon_tools.py
```diff
@@ -203,7 +203,7 @@ def get_min_instances(self):
     def get_max_instances(self):
         return self.config_dict.get('max_instances', None)
 
-    def get_instances(self):
+    def get_desired_instances(self):
```
mjksmith: `git grep` shows that a lot of things use `get_instances`. After this change these calls will get the current instance count in zk/soa_configs even if a service is stopped. Are we sure that all of these invocations don't want to be aware of if a services is started or stopped?

solarkennedy: From the ticket description it sounds like the only thing that should respect start/stop being 0 is the actual marathon json. Status, alerting, reporting, etc should be consistent and show this as "red" if we want this to be an "non-expected" condition.

mjksmith:
I don't really like the idea of alerting users that have stopped tasks.

The code itself looks fine though.

paasta_tools/marathon_tools.py
```diff
@@ -214,30 +214,35 @@ def get_instances(self):
...
-                                    limited_instances, zk_instances))
-                    return limited_instances
-            else:
-                instances = self.config_dict.get('instances', 1)
-                log.debug("Autoscaling not enabled, returning %d instances" % instances)
-                return instances
+            return self.get_instances()
         else:
             log.debug("Instance is set to stop. Returning '0' instances")
             return 0
```
EvanKrall: This function and `PaastaNativeServiceConfig.get_instances()` are now very similar -- can you put this function in `LongRunningServiceConfig`, make `PaastaNativeServiceConfig.get_instances()` just `return self.config_dict.get('instances')`, and change the two `get_instances` calls in native_mesos_scheduler to `get_desired_instances()`?

Rob-Johnson:
I can't say I'm totally convinced by this. Running ``paasta stop`` says that you expect the state of your app to be stopped - why is it a critical condition for PaaSTA to be in the state you asked it to be? PaaSTA is doing the right thing here. 

I kind of see it as a process failure if you're able to take down a service and not notice for a long time.
I'd expect to see tighter gitolite ACLs enforce that you can't just take down production by default, rather than enforcing the opinion of 'you've not done the right thing here' in PaaSTA. All of our other alerting is just enforcing 'actual state != desired state', rather than 'we don't think you should do this'.

Alternatively, how about another monitoring check that asserts there is no disparity between the state declared with ``paasta stop`` and the instances declared in soa-configs. We're kind of doing the latter here anyway, but having it as a separate alert means that it sits 'above' PaaSTA, and doesn't enforce the 'this is wrong' opinion in the codebase.

EvanKrall:
@Rob-Johnson the internal ticket explains my justifications:

> There are at least two ways to stop services in paasta:
1. `paasta stop`
2. setting `instances: 0` in marathon.yaml
>
> I expect `paasta stop` to be used when the service needs to be taken down in a hurry, like if it's causing site issues. Since `paasta stop` is overridden by a deploy, I don't consider it useful for when you want a service stopped for longer-term. In the longer-term case, you'd want to use `instances: 0` instead.
>
> However, since `paasta stop` takes effect until someone does `paasta start` or jenkins does `paasta mark-for-deployment`, it is possible for an errant `paasta stop` to cause a service to be down indefinitely.

> It's also possible for `paasta stop` to cause surprises later: if someone had intended for the service to be stopped long-term, and someone does a deploy, that deploy will override the stop and the service will start.

> Given all that, it seems to me that a service being `paasta stop`ped for longer than a few minutes or hours should be considered a mistake, and we should send an alert. If the service authors want the service to be down for a short period, they can ignore the alert, and if they want the service to be down indefinitely, they can use `instances: 0` instead.

solarkennedy:
@EvanKrall  can you re-review, I think I've accounted for your change request.

PR #880, opened 2016-11-04
Added get_currently_deployed_sha and better rollback hints

solarkennedy:
Now that @mjksmith has v2 deployments out, extracting the existing deployed git sha is trivial.

After this ships it will be very easy to add auto-rollback.

### DIFF
--- a/paasta_tools/cli/cmds/get_latest_deployment.py
+++ b/paasta_tools/cli/cmds/get_latest_deployment.py
@@ -17,10 +17,8 @@ from paasta_tools.cli.utils import list_deploy_groups
 from paasta_tools.cli.utils import list_services
 from paasta_tools.cli.utils import PaastaColors
 from paasta_tools.cli.utils import validate_service_name
-from paasta_tools.generate_deployments_for_service import get_latest_deployment_tag
-from paasta_tools.remote_git import list_remote_refs
+from paasta_tools.deployment_utils import get_currently_deployed_sha
 from paasta_tools.utils import DEFAULT_SOA_DIR
-from paasta_tools.utils import get_git_url
 
 
 def add_subparser(subparsers):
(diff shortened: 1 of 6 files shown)


paasta_tools/deployment_utils.py
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env python
```
mjksmith: Do you have future plans for this file? If not it seems kind of wasteful.

There's a lot of commands in utils (and elsewhere) that are only used in service deployment. I'm personally a fan of breaking up utils.

paasta_tools/cli/cmds/get_latest_deployment.py
```diff
@@ -53,13 +51,7 @@ def paasta_get_latest_deployment(args):
...
     validate_service_name(service, soa_dir)
 
-    git_url = get_git_url(
-        service=service,
-        soa_dir=soa_dir,
-    )
-    remote_refs = list_remote_refs(git_url)
-
-    _, git_sha = get_latest_deployment_tag(remote_refs, deploy_group)
+    git_sha = get_currently_deployed_sha(service=service, deploy_group=deploy_group, soa_dir=soa_dir)
```
mjksmith: This raises a `NoDeploymentsAvailableException` if the service isn't deployed yet. You should catch that instead of checking `not git_sha` later

solarkennedy:
This is ready for one more quick review pass

paasta_tools/deployment_utils.py
solarkennedy: I do have future plans for this file. I do intend to relocate things out of utils and fill up this one, but those moves are more "mechanical" and I didn't want to clutter up the review.

paasta_tools/cli/cmds/get_latest_deployment.py
solarkennedy: Addressed.

PR #881, opened 2016-11-07
Autoscaler ASG support

mattmb:
This is a fairly big refactor to add ASG support. Major things:

* Move to resource "setters" which set capacity for each type of resource but keep "scaler" the same for both
* Make list of instances a resource property so that it can be in the same place for SFRs and ASGs
* Add specific handling for ASG "delta" (we need this because we have to handle cleaning up ASGs differently)

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -25,8 +25,6 @@ import boto3
 from botocore.exceptions import ClientError
 from requests.exceptions import HTTPError
 
-from paasta_tools.autoscaling.utils import _autoscaling_components
-from paasta_tools.autoscaling.utils import register_autoscaling_component
 from paasta_tools.mesos_maintenance import drain
 from paasta_tools.mesos_maintenance import undrain
 from paasta_tools.mesos_tools import get_mesos_master
@@ -42,7 +40,6 @@ from paasta_tools.utils import TimeoutError
 CLUSTER_METRICS_PROVIDER_KEY = 'cluster_metrics_provider'
 DEFAULT_TARGET_UTILIZATION = 0.8  # decimal fraction
 DEFAULT_DRAIN_TIMEOUT = 600  # seconds
-SETTER_KEY = 'setter'
 
 AWS_SPOT_MODIFY_TIMEOUT = 30
 MISSING_SLAVE_PANIC_THRESHOLD = .3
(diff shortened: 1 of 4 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -178,20 +219,26 @@ def is_aws_launching_sfr_instances(sfr):
...
     if not dry_run:
-        os.remove(sfr_configs_to_delete[0])
-    log.info("Deleted SFR config {0}".format(sfr_configs_to_delete[0]))
+        os.remove(configs_to_delete[0])
+    log.info("Deleted Resource config {0}".format(configs_to_delete[0]))
 
 
-def get_sfr_slaves(resource, mesos_state):
-    instance_ips = get_sfr_instance_ips(resource['sfr'], region=resource['region'])
+def get_aws_slaves(resource, mesos_state):
```
Rob-Johnson: this seems kind of muddled: this isn't really getting aws slaves, but just matching slaves by the ips in a resource, right? so this won't return the complete set of aws slaves in the mesos_state; just the ones in this resource. Make this a bit more generic:

mesos_tools.py
```
def filter_slaves_by_ip(acceptable_ips, slaves):
   return filter(lambda slave: slave_pid_to_ip(slave['pid') in acceptable_ips)
```

autoscaling_lib.py
```
def get_slaves_in_resource(resource, slaves):
    instance_ips = get_instance_ips(resource['ActiveInstances'], region=resource['region'])
    return mesos_tools.filter_slaves_by_ip(instance_ips, slaves)
```

I also don't understand why this is creating a dict of id: slave - we already have the ids in the slave object, why do we need to make a more complicated data structure?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -333,48 +405,56 @@ def sort_slaves_to_kill(slaves):
...
+def get_instance_type_weights(resource):
+    if resource['type'] == 'aws_spot_fleet_request':
+        launch_specifications = resource['sfr']['SpotFleetRequestConfig']['LaunchSpecifications']
+        return {ls['InstanceType']: ls['WeightedCapacity'] for ls in launch_specifications}
+    else:
+        return StaticWeight()
+
+
+class StaticWeight:
+    def __getattr__(self, attr):
```
Rob-Johnson: this is a bit nasty...want to tighten it up a little?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -333,48 +405,56 @@ def sort_slaves_to_kill(slaves):
...
-    sfr_ips = get_sfr_instance_ips(sfr, region=resource['region'])
-    log.debug("IPs in SFR: {0}".format(sfr_ips))
-    sfr_slaves = [slave for slave in slaves_list if slave_pid_to_ip(slave['task_counts'].slave['pid']) in sfr_ips]
-    sfr_slave_ips = [slave_pid_to_ip(slave['task_counts'].slave['pid']) for slave in sfr_slaves]
-    sfr_instance_descriptions = describe_instances([], region=resource['region'],
-                                                   instance_filters=[{'Name': 'private-ip-address',
-                                                                      'Values': sfr_slave_ips}])
-    sfr_slave_instances = []
-    for slave in sfr_slaves:
+def filter_aws_slaves(slaves_list, resource):
```
Rob-Johnson: haven't you got another function called ``get_aws_slaves``?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -333,48 +405,56 @@ def sort_slaves_to_kill(slaves):
...
             log.error("Found more than one instance with the same private IP {0}. "
                       "This should never happen")
             continue
-        sfr_slave_instances.append({'ip': ip,
-                                    'task_counts': slave['task_counts'],
-                                    'hostname': slave['task_counts'].slave['hostname'],
-                                    'id': slave['task_counts'].slave['id'],
-                                    'pid': slave['task_counts'].slave['pid'],
-                                    'instance_id': instances[0]['InstanceId']})
+        slave_instances.append({'ip': ip,
```
Rob-Johnson: same comment r.e passing undocumented dicts around.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -396,20 +476,37 @@ def set_spot_fleet_request_capacity(sfr_id, capacity, dry_run, region=None):
...
+        return
+    asg_client = boto3.client('autoscaling', region_name=region)
+    try:
+        ret = asg_client.update_auto_scaling_group(AutoScalingGroupName=asg_name,
+                                                   DesiredCapacity=capacity)
+    except ClientError as e:
+        log.error("Error modifying ASG: {0}".format(e))
+        if e.response['Error']['Code'] == 'ValidationError':
+            log.warning("A validation error on an ASG update is usually caused by trying to update capacity "
+                        "outside min/max, which is ok if we're scaling down!")
```
solarkennedy: Is it? I didn't think you could set the desired count below min in an asg?

solarkennedy:
ship

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -49,518 +46,589 @@
...
+    def get_instance_type_weights(self):
+        return StaticWeight()
 
-    :param ip: private IP of AWS instance.
-    :param instance_descriptions: list of AWS instance description dicts.
-    :returns: list of instance description dicts"""
-    instances = [instance for instance in instance_descriptions if instance['PrivateIpAddress'] == ip]
-    return instances
 
+class StaticWeight:
```
Rob-Johnson: I still don't really like this, but add some docs to explain what it's for and when it'll go away.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -49,518 +46,589 @@
...
+        """Pick the best slaves to kill. This returns a list of slaves
+        after sorting in preference of which slaves we kill first.
+        It sorts first by number of chronos tasks, then by total number of tasks
+
+        :param slaves: list of slaves dict
+        :returns: list of slaves dicts"""
+        return sorted(slaves, key=lambda x: (x['task_counts'].chronos_count, x['task_counts'].count))
+
+    def scale_resource(self, current_capacity, target_capacity):
+        """Scales a spot fleet request by delta to reach target capacity
```
Rob-Johnson: this docs don't seem correct. Can you confirm that there is nothing spotfleet specific in here?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -49,518 +46,589 @@
...
+            amount_to_decrease = delta * -1
+            if amount_to_decrease > killable_capacity:
+                log.error("Didn't find enough candidates to kill. This shouldn't happen so let's not kill anything!")
+                return
+            self.downscale_aws_resource(
+                filtered_slaves=filtered_slaves,
+                current_capacity=current_capacity,
+                target_capacity=target_capacity)
+
+    def gracefully_terminate_slave(self, slave_to_kill, current_capacity, new_capacity):
```
Rob-Johnson: can you give a summary of how this works as docs?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -49,518 +46,589 @@
...
-        sfr_slave_instances.append({'ip': ip,
+            log.error("Failure when terminating: {0}: {1}".format(slave_to_kill['pid'], e))
+            log.error("Setting resource capacity back to {0}".format(current_capacity))
+            self.set_capacity(current_capacity)
+        finally:
+            log.info("Undraining {0}".format(slave_to_kill['pid']))
+            if not self.dry_run:
+                undrain([drain_host_string])
+
+    def filter_aws_slaves(self, slaves_list):
```
Rob-Johnson: why does this and ``get_aws_slaves`` exist?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -49,518 +46,589 @@
...
-            state = None
-            while True:
-                state = get_sfr(sfr_id, region=region)['SpotFleetRequestState']
-                if state == 'active':
+        ret = []
+        instance_type_weights = self.get_instance_type_weights()
+        for slave in slave_instances:
+            instance_description = [instance_description for instance_description in instance_descriptions
+                                    if instance_description['InstanceId'] == slave['instance_id']][0]
+            slave['instance_type'] = instance_description['InstanceType']
```
Rob-Johnson: As I mentioned before, can you define something that states what is actually in this? I still don't really like that there are other parts of the code base that are relying on this extra metadata being added without explicitly saying so.

```
class RichAWSInstace(object):
   """A regular AWS Instance object, with additional metadata."""

   def __init__(self, instance_type, instance_weight, instance):
        self.instance_type = instance_type
        self.instance_weight = instance_weight
        self.original_slave = slave
```

Then have this function return a list of of ``RichSlave`` objects.

Then everywhere that relies on this can say so with docs, variable names and type checking.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -49,518 +46,589 @@
...
 
-def get_spot_fleet_instances(spotfleet_request_id, region=None):
-    ec2_client = boto3.client('ec2', region_name=region)
-    spot_fleet_instances = ec2_client.describe_spot_fleet_instances(
-        SpotFleetRequestId=spotfleet_request_id)['ActiveInstances']
-    return spot_fleet_instances
+    def describe_instances(self, instance_ids, region=None, instance_filters=None):
+        """This wraps ec2.describe_instances and catches instance not
+        found errors. It returns a list of instance description
+        dictionaries. It assumes one instance per reservation (which
```
mattmb: Ah, I've actually fixed this because it doesn't hold for ASGs. In fact I could do with updating a few docstrings! 😢

paasta_tools/autoscaling/autoscaling_cluster_lib.py
mattmb: I like this idea so I've done roughly what you suggested.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
mattmb: As a bonus it removes the need for this hack

mattmb:
@Rob-Johnson I've added some commits to fixup some of the issues you raised.

Can you take a look at the new `PaastaAwsSlave` in particular.

I'm going to hold off on most of the other things you've suggested, a lot of the suggestions will make the code more readable (which I am very onboard with). I think it'll be easier to tackle them in a separate branch and ship'n'iterate

PR #883, opened 2016-11-07
Made paasta status and paasta metastatus return nonzero on ssh error

### LINKED ISSUE #718: paasta status exits 0 when it cannot ssh (should be nonzero)
```
$ env -i paasta status -c norcal-prod -s www_pages
Pipeline: https://jenkins.yelpcorp.com/view/services-www_pages

cluster: norcal-prod
    ERROR: could not find connectable master in cluster norcal-prod
    Output: ERROR cannot run check command ssh -A -n 10.44.6.103 sudo paasta_serviceinit -h
    Return code was 255 which is an unknown failure.
    HINT: Talk to #operations and pastebin this output
    Output from check command: Permission denied (publickey).
    ssh_exchange_identification: Connection closed by remote host


$ echo $?
0
```

mjksmith:
Fixes #718 

Also upgraded `mock` to `2.0.0` to fix a bug with `side_effect` that caused some unexpected behaviour.

### DIFF
--- a/paasta_tools/cli/cmds/metastatus.py
+++ b/paasta_tools/cli/cmds/metastatus.py
@@ -140,4 +140,4 @@ def paasta_metastatus(args):
         else:
             print "Cluster %s doesn't look like a valid cluster?" % args.clusters
             print "Try using tab completion to help complete the cluster name"
-    return all([return_code == 0 for return_code in return_codes])
+    return 0 if all([return_code == 0 for return_code in return_codes]) else 1
--- a/paasta_tools/cli/cmds/metastatus.py
+++ b/paasta_tools/cli/cmds/metastatus.py
@@ -77,16 +77,20 @@ def add_subparser(subparsers):
 def print_cluster_status(cluster, system_paasta_config, humanize, groupings, verbose=0):
     """With a given cluster and verboseness, returns the status of the cluster
     output is printed directly to provide dashbaords even if the cluster is unavailable"""
-    print "Cluster: %s" % cluster
-    print get_cluster_dashboards(cluster)
-    print execute_paasta_metastatus_on_remote_master(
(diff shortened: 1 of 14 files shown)


paasta_tools/cli/cmds/metastatus.py
```diff
@@ -121,15 +125,19 @@ def paasta_metastatus(args):
...
+                    system_paasta_config=system_paasta_config,
+                    humanize=args.humanize,
+                    groupings=args.groupings,
+                    verbose=args.verbose,
+                )
             )
         else:
             print "Cluster %s doesn't look like a valid cluster?" % args.clusters
             print "Try using tab completion to help complete the cluster name"
+    return all([return_code == 0 for return_code in return_codes])
```
mattmb: Is this doing what you expect? Won't it return True if everything returns 0 and then end up with `sys.exit(True)`?

mjksmith:
@mattmb the reason your fix wasn't caught by testing is because I didn't prefix my tests with 'test_' -_-

Always make sure your tests can fail, kids.

paasta_tools/cli/cmds/metastatus.py
```diff
@@ -121,15 +125,19 @@ def paasta_metastatus(args):
...
+                    system_paasta_config=system_paasta_config,
+                    humanize=args.humanize,
+                    groupings=args.groupings,
+                    verbose=args.verbose,
+                )
             )
         else:
             print "Cluster %s doesn't look like a valid cluster?" % args.clusters
             print "Try using tab completion to help complete the cluster name"
+    return 0 if all([return_code == 0 for return_code in return_codes]) else 1
```
solarkennedy: How about max() ?

solarkennedy:
ship

PR #884, opened 2016-11-08
Setup marathon speedup

mattmb:
Just scratching an itch with this PR. I've attempted to pin down some slow bits in `setup_marathon_job` and make them "better".

Not really sure how much time this might save. The getting of marathon tasks will only improve if we call this script with multiple services (which we currently don't). @mjksmith you're the expert on this right? Do you think this will help?

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -363,6 +363,7 @@ def deploy_service(
     marathon_jobid,
     config,
     client,
+    marathon_apps,
     bounce_method,
     drain_method_name,
     drain_method_params,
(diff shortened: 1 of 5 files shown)


mattmb:
Ug, looks like the itests have failed, I'll take a look and see if I've broken anything "real"

paasta_tools/setup_marathon_job.py
```diff
@@ -241,7 +241,8 @@ def log_bounce_action(line, level='debug'):
         at_risk_tasks=bounce_lib.flatten_tasks(old_app_at_risk_tasks),
     )
 
-    kill_given_tasks(client=client, task_ids=[task.id for task in tasks_to_kill], scale=True)
+    if tasks_to_kill:
+        kill_given_tasks(client=client, task_ids=[task.id for task in tasks_to_kill], scale=True)
```
mjksmith: It seems better that `kill_given_tasks` knows how to do 'the right thing' and doesn't call `client.kill_given_tasks` if `tasks_to_kill` is empty.

solarkennedy:
Ship!

paasta_tools/setup_marathon_job.py
mattmb: That's a good shout, I'll move this fix into the function itself.

mattmb:
@mjksmith I've only profiled it on my local docker-compose setup. It certainly makes some improvement but it's hard to measure in such a fake environment!

PR #887, opened 2016-11-10
Added kubernetes to paasta local run and validate.

matthewmcarlson:


paasta_tools/cli/cmds/local_run.py
```diff
@@ -354,6 +354,14 @@ def add_subparser(subparsers):
         required=False,
         default=False,
     )
+    list_parser.add_argument(
```
mjksmith: Is this for when we use kubectl?

PR #896, opened 2016-11-16
Removed the soa_dir option from generate-pipeline

### LINKED ISSUE #895: paasta generate-pipeline takes soa_dir but doesn't use it to find deploy.yaml

solarkennedy:
Resolves #895.

Realistically we are not going to make fab_repo soa-dir aware any time soon, so I would rather we remove this misleading option.

If we do make all the yelp repo tools parameterized or we change to jenkins job builder or something, we can easily add this option back in.

### DIFF
--- a/paasta_tools/cli/cmds/generate_pipeline.py
+++ b/paasta_tools/cli/cmds/generate_pipeline.py
@@ -43,13 +43,6 @@ def add_subparser(subparsers):
         '-s', '--service',
         help='Name of service for which you wish to generate a Jenkins pipeline',
     ).completer = lazy_choices_completer(list_services)
-    list_parser.add_argument(
-        '-d', '--soa-dir',
-        dest="soa_dir",
-        metavar="SOA_DIR",
-        default=DEFAULT_SOA_DIR,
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/generate_pipeline.py
```diff
@@ -43,28 +43,20 @@ def add_subparser(subparsers):
...
-    soa_dir = args.soa_dir
     try:
-        validate_service_name(service, soa_dir=soa_dir)
+        validate_service_name(service, soa_dir=DEFAULT_SOA_DIR)
     except NoSuchService as service_not_found:
         print service_not_found
         return 1
 
-    generate_pipeline(service=service, soa_dir=soa_dir)
+    generate_pipeline(service=service)
```
mjksmith: Why don't we leave `get_git_repo_for_fab_repo` and `generate_pipeline` (lines 73 and 97) soa-dir aware and just call `generate_pipeline(service=service, soa_dir=DEFAULT_SOA_DIR)`. That way we don't have to change a lot in case we ever decide to make changes to fab_repo.

PR #899, opened 2016-11-17
Local run build fix

mjksmith:
Right now `local-run --build` prompts for a deploy group even though it doesn't need to know the git sha


solarkennedy:
ship

PR #900, opened 2016-11-18
add a decorator for retrying zookeeper connections

Rob-Johnson:
we *should* be more resillient to single ZK boxes being
unavailable after this. PAASTA-6639

### DIFF
--- a/paasta_tools/mesos/master.py
+++ b/paasta_tools/mesos/master.py
@@ -88,7 +88,7 @@ class MesosMaster(object):
         path = "/" + path
 
         retry = KazooRetry(max_tries=10)
-        with zookeeper.client(hosts=hosts, read_only=True, connection_retry=retry) as zk:
+        with zookeeper.client(hosts=hosts, read_only=True, connection_retry=retry, command_retry=retry) as zk:
             def master_id(key):
                 return int(key.split("_")[-1])
 
--- a/paasta_tools/mesos/master.py
+++ b/paasta_tools/mesos/master.py
@@ -19,12 +19,14 @@ from __future__ import print_function
 import fnmatch
 import itertools
 import json
+import logging
 import os
 import re
 import urlparse
 
 import requests
 import requests.exceptions
+from kazoo.handlers.threading import KazooTimeoutError
 from kazoo.retry import KazooRetry
 
 from . import exceptions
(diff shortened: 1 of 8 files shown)


Rob-Johnson:
@ronin13 this is what I came up with

paasta_tools/utils.py
```diff
@@ -1700,3 +1701,45 @@ def prompt_pick_one(sequence, choosing):
         sys.exit(1)
     else:
         return result
+
+
+def retry(ExceptionToCheck, tries=4, delay=3, backoff=2, logger=None):
```
ronin13: As discussed, https://github.com/invl/retry / https://pypi.python.org/pypi/retry/ can be reused here.

paasta_tools/paasta_metastatus.py
```diff
@@ -39,6 +40,11 @@
 from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import print_with_indent
 
+
+logging.basicConfig()
+# kazoo can be really noisy - turn it down
+logging.getLogger("kazoo").setLevel(logging.CRITICAL)
```
mattmb: Was this info before? I agree it's annoyingly chatty, maybe ERROR would be okay though.

paasta_tools/mesos/master.py
```diff
@@ -76,12 +81,13 @@ def fetch(self, url, **kwargs):
...
         return self.resolve(open(cfg[6:], "r+").read().strip())
 
+    @retry(KazooTimeoutError, tries=5, delay=0.5, logger=logger)
     def _zookeeper_resolver(self, cfg):
         hosts, path = cfg[5:].split("/", 1)
         path = "/" + path
 
         retry = KazooRetry(max_tries=10)
-        with zookeeper.client(hosts=hosts, read_only=True, connection_retry=retry) as zk:
+        with zookeeper.client(hosts=hosts, read_only=True, connection_retry=retry, command_retry=retry) as zk:
```
mattmb: I don't really understand what these args are for. Isn't the decorator enough on its own?

Rob-Johnson: nope - the decorator deals with a non-zookeeper exeception(ConnectionTimeout) - connection_retry and command_retry deal with zookeeper specific errors.

PR #902, opened 2016-11-19
Fix logs trying to output unicode

mjksmith:
piping paasta logs or redirecting to a file causes an exception. This fixes it (I think).

paasta_itests fails and I don't know why.

Also I change all of the print statements to print functions because I can

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 from __future__ import print_function
 
+import collections
 import contextlib
 import copy
 import datetime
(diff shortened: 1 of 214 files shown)


solarkennedy:
Ship!

paasta_tools/utils.py
```diff
@@ -1700,3 +1701,21 @@ def prompt_pick_one(sequence, choosing):
...
+    elif isinstance(obj, collections.Mapping):
+        return type(obj)(map(encode_object_to_utf8, obj.iteritems()))
+    elif isinstance(obj, collections.Iterable):
+        return type(obj)(map(encode_object_to_utf8, obj))
+    else:
+        return obj
+
+
+def load_json(*args, **kwargs):
+    return encode_object_to_utf8(json.loads(*args, **kwargs))
```
asottile: this is the opposite of what you usually want to do.  The best practice is to encode / decode from bytes at application boundaries and deal with text inside your application.

With stdlib json at least you can do this with the `encoding=...` parameter in python2 (python3 enforces that you use text internally when dealing with the `json` interface).

This is probably not the right approach as written ;)

solarkennedy:
Rebase and ship? I don't mind a wrapper so much, but it isn't a silver bullet, because it probably won't solve `log()` invocations?

PR #905, opened 2016-11-23
Split metastatus into a lib and cli script

mattmb:
This takes the bulk of the metastatus code and moves it to
paasta_tools.metrics.metastatus

The motivation is to separate re-usable parts of metastatus from the
cli script. Specifically we had a problem because the autoscaler is
importing metastatus which then stomps on its logging options. With this
configuration it can import the library safely.


Rob-Johnson:
rename it metastatus_lib.py and shipit!

PR #906, opened 2016-11-24
Bounce log clean

mattmb:
Just a cosmetic change to the bounce logs

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -41,6 +41,7 @@ Command line options:
 - -v, --verbose: Verbose output
 """
 import argparse
+import json
 import logging
 import sys
 import traceback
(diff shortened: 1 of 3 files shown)


Rob-Johnson:
got an example diff?

Rob-Johnson:
bump^

paasta_tools/setup_marathon_job.py
```diff
@@ -200,19 +201,26 @@ def log_bounce_action(line, level='debug'):
...
-                '%s bounce in progress on %s.' % (bounce_method, serviceinstance),
-                'New marathon app %s %s.' % (marathon_jobid, ('exists' if new_app_running else 'not created yet')),
-                '%d new tasks to bring up.' % (config['instances'] - len(happy_new_tasks)),
-                '%d old tasks receiving traffic and happy.' % len(bounce_lib.flatten_tasks(old_app_live_happy_tasks)),
-                '%d old tasks unhappy.' % len(bounce_lib.flatten_tasks(old_app_live_unhappy_tasks)),
-                '%d old tasks draining.' % len(bounce_lib.flatten_tasks(old_app_draining_tasks)),
-                '%d old tasks at risk.' % len(bounce_lib.flatten_tasks(old_app_at_risk_tasks)),
-                '%d old apps.' % len(old_app_live_happy_tasks.keys()),
+                '%s bounce: ' % bounce_method,
+                json.dumps(log_dict)
```
solarkennedy: Are you sure json is more readable in this example? This means we are putting json within json in the raw log.

I don't see how this makes this log *more* readable for humans. cc @EvanKrall

mattmb:
Here's a before and after:

`crossover bounce in progress on hello-world.main. New marathon app hello-world.main.git9cdf957d.configbcc1cb47 not created yet. 5 new tasks to bring up. 0 old tasks receiving traffic and happy. 0 old tasks unhappy. 0 old tasks draining. 0 old tasks at risk. 0 old apps.`

`crossover bounce:  {"new_app": "hello-world.main.git9cdf957d.configa75c7adb", "tasks_to_launch": 5, "old_receiving": 2, "old_draining": 0, "new_app_running": "no", "old_at_risk": 0, "old_apps": 1}`

I agree that the json within json thing is a bit nasty but I actually find it easier to read than what we have before. Happy to be out voted if everyone prefers the current one though.

paasta_tools/setup_marathon_job.py
EvanKrall: This is gross, but I bet you could just pass a dict to _log.

Alternatively, what about having _log / LogWriter.log() / format_log_line() support an extra information parameter which would get serialized directly into the JSON?

solarkennedy:
I vote for before.

mattmb:
I don't feel super strongly about this so I'm going to hit close

PR #907, opened 2016-11-25
remove include_temporary from chronos_serviceinit

Rob-Johnson:
unintentionally included in d484d15714f84d7fc1908902520ab8afa2eb376c

### DIFF
--- a/paasta_tools/chronos_serviceinit.py
+++ b/paasta_tools/chronos_serviceinit.py
@@ -370,7 +370,6 @@ def perform_command(command, service, instance, cluster, verbose, soa_dir):
             instance=instance,
             client=client,
             include_disabled=True,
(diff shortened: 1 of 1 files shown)


mattmb:
You mean `include_temporary`?

Rob-Johnson:
Ha, I do.

mattmb:
Also, why don't we want to stop 'tmp' jobs? Is it because they are explicitly started by a user rather so paasta should leave them alone?

Rob-Johnson:
Eventually, we do (PAASTA-7038). But I need to finish writing tests and make a proper PR, rather than sneaking a change in with another PR.

mattmb:
Right, thought it sounded familiar. Sounds good, can you amend the commit message before you ship 😄

PR #913, opened 2016-12-01
Move requirements to build for example cluster

mattmb:
This moves the installation of the python requirements into the docker
build. This speeds up the startup a bunch and also means that you can
make sure you have the correct requirements by running `docker-compose
build` to rebuild all the containers.

@Rob-Johnson as I mentioned this is possible because of being able to customise the context in docker-compose. Previously I couldn't think of a way to do it.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,3 @@
+*
+!yelp_package/*
+!requirements*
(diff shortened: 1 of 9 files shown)


Rob-Johnson:
🚢

PR #914, opened 2016-12-01
Reduce the number of different soa_dir argument definitions

oktopuz:
Replace multiple soa_dir argument definitions by the one defined in paasta_tools.cli.argparse_helper


Rob-Johnson:
thanks for this @oktopuz - we need to think of a grander plan for refactoring the argparse bits, so I wouldn't worry about this for now. Feel free to close.

PR #915, opened 2016-12-01
paasta wait-for-deployment

oktopuz:
internal ticket: PAASTA-7113

* add paasta wait-for-deployment
* move all functions and tests related to wait-for-deployment to their own files
* make 'paasta wait-for-deployment' to return non-zero exit code for errors
* change all of the print statements in wait_for_deploymet.py to print functions


paasta_tools/cli/cmds/wait_for_deployment.py
```diff
@@ -0,0 +1,323 @@
...
+                               "used anywhere: %s.\n" %
+                               (",").join(invalid_deploy_groups)))
+        print(PaastaColors.red("You probably need one of these in-use deploy "
+                               "groups?:\n   %s" %
+                               (",").join(in_use_deploy_groups)))
+        return 1
+
+    old_git_sha = get_currently_deployed_sha(service=service,
+                                             deploy_group=args.deploy_group)
+    if old_git_sha == args.commit:
```
solarkennedy: I'm not sure this error message really makes in this context. It will pretty much *always* be the case that you are waiting on what is set to be deployed.

This needs to be only in mark_for_deployment

paasta_tools/cli/cmds/wait_for_deployment.py
```diff
@@ -0,0 +1,323 @@
...
+             level='event')
+
+        line = ("\n\nTimed out after {0} seconds, waiting for {1} in {2} to be "
+                "deployed by PaaSTA. \n\n"
+                "This probably means the deploy hasn't suceeded. The new "
+                "service might not be healthy or one or more clusters could be "
+                "having issues.\n\n"
+                "To debug: try running 'paasta status -s {2} -vv' or 'paasta "
+                "logs -s {2}' to determine the cause.\n\n"
+                "{3} is still *marked* for deployment. To rollback, you can "
```
mattmb: Same as my comment above, we should probably tweak this error message a bit because we haven't necessarily marked for deployment.

oktopuz: What do you think about getting latest deployment, comparing it with the provided git_sha, and exiting when they don't match?
This would help to avoid waiting for deployment that was not marked, and these messages would be valid without any tweak, I guess.

solarkennedy: @oktopuz that sounds like a good suggestion, go for it.

oktopuz:
I did some testing I realized that get-latest-deployment may not return the very latest deployment, if the previous one is still in flight. So, I just updated the error messages in question.

oktopuz:
rebased.

mattmb:
This looks good to me now 😄 I'm glad you checked on the get-latest-deployment there's so much async stuff in paasta that it's very easy for race conditions to creep in!

PR #916, opened 2016-12-02
fixed local run 'no tty message' not appearing with --build

mjksmith:
Sample error:
```
Warning! You're running a container in interactive mode.
This is *NOT* how Mesos runs containers.
To run the container exactly as Mesos does, omit the -I flag.

cannot enable tty mode on non tty input
make: *** [itest] Error 1
```

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -632,18 +632,25 @@ def configure_and_run_docker_container(
     Function prints the output of run command in stdout.
     """
 
+    if instance is None and args.healthcheck:
+        paasta_print(
+            "With --healthcheck, --instance must be provided!",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    if instance is None and not sys.stdin.isatty():
+        paasta_print(
+            "--instance and --cluster must be specified when using paasta local-run without a tty!",
+            file=sys.stderr,
+        )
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -632,18 +632,25 @@ def configure_and_run_docker_container(
...
     """
 
-    soa_dir = args.yelpsoa_config_root
+    if instance is None and args.healthcheck:
+        paasta_print(
+            "With --healthcheck, --instance must be provided!",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    if instance is None and not sys.stdin.isatty():
```
solarkennedy: How about if 

`interactive is Tru and not tty`

I think that would achieve the same effect and be clearer, but you would need to place it further down after we've figured out if we are interactive or not.

mjksmith: @solarkennedy it would be nice to have a different error message between no `--instance` (you need to specify `--instance` and `--cluster`) vs interactive being set with `-I` (don't use `-I` without a tty). We can't have that if we just check `interactive is True and not sys.stdin.isatty()`.

solarkennedy: Ok.

solarkennedy:
fix coverals and ship

PR #918, opened 2016-12-05
make python-crontab less noisy

Rob-Johnson:
It was logging messages whenever it couldn't parse a cron, which happens
every time we run setup_chronos_job with on a job with an ISO8601 schedule.
Turn down the logging.


solarkennedy:
Wouldn't this also happen during paasta validate, or paasta local-run, etc?

Rob-Johnson:
@solarkennedy you were right. updated.

solarkennedy:
ship

PR #919, opened 2016-12-05
made send_event not load deployments

mjksmith:
This causes `setup_chronos_job` to not fail when a chronos service has not been deployed yet


solarkennedy:
Ship. I'm guessing check_chronos and check_marathon would be the only places where this would show up.

PR #924, opened 2016-12-08
[WIP] convert services.yaml to bytes before dumping to yaml

mjksmith:
The fact that our tests didn't catch this is bad. I want to add more tests to this before shipping & unpinning.

If you absolutely need to unpin you can merge this (I tested it locally) and watch services.yaml for any weirdness as it deploys.


chriskuehl:
I added a test in #925.

..-UTcARg
```diff
@@ -0,0 +1,1255 @@
+---
```
chriskuehl: i guess you didn't mean to commit this?

mjksmith: 👍

solarkennedy:
@mjksmith just merge in #925 into this branch and see the tests pass.

PR #925, opened 2016-12-08
Add some useful comments at the top of the services.yaml

chriskuehl:
This adds a comment block at the top of the `services.yaml` like:

```yaml
# This file is automatically generated by paasta_tools.
# It was automatically generated at 2016-12-07T20:09:21.251099 on dev4-uswest1cdevc.dev.yelpcorp.com.
---
!!python/unicode 'someservice.main':
  !!python/unicode 'host': !!python/unicode '169.254.255.254'
  !!python/unicode 'port': 20974
!!python/unicode 'someservice2.main':
  !!python/unicode 'host': !!python/unicode '169.254.255.254'
  !!python/unicode 'port': 20000
```

This would have helped speed up the resolution of OPS-10738 a bit by making it clear where this file is generated, and also adds a test that would have failed (and is currently failing, since the bug still exists).


solarkennedy:
@chriskuehl I think this needs to be rebased on #924 when it is merged to make the tests pass, right?

chriskuehl:
@solarkennedy @mjksmith I rebased my branch on #924 and fixed a trivial conflict. Tests look good locally, should be fine to merge #924 and then this.

PR #926, opened 2016-12-08
PAASTA-7038 include rerun jobs in emergency-stop

Rob-Johnson:
before now, there was no way to interrupt in-flight tasks for a job
launched with 'paasta rerun'. This change includes the tasks launched by
paasta rerun in those killed by paasta emergency-stop


solarkennedy:
I'm surprised at all the test failures?

solarkennedy:
ship

PR #928, opened 2016-12-09
Use v2 deployments (again)

mjksmith:
Added a test this time (that is currently failing) that makes sure deploy group interpolation works


solarkennedy:
I'm not exactly sure how to think about this. Let me think a bit.

solarkennedy:
Well, I say "Shipit" because it does work, but can someone else give a second opinion? (@EvanKrall  or @Rob-Johnson  or @mattmb)

solarkennedy:
I'm going to make the call and discard this.

PR #929, opened 2016-12-09
PAASTA-7203 validate chronos cmds

Rob-Johnson:
Given that we don't validate, you won't get any errors until deploy
time. This adds validation to the chronos cmd so we pick it up with
paasta validate

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -307,7 +307,7 @@ class ChronosJobConfig(InstanceConfig):
                 parse_time_variables(command)
                 return True, ""
             except (ValueError, KeyError, TypeError):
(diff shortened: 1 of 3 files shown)


paasta_tools/chronos_tools.py
```diff
@@ -300,6 +300,17 @@ def check_parents(self):
...
         return True, ''
 
+    def check_cmd(self):
+        command = self.get_cmd()
+        if command:
+            try:
+                parse_time_variables(command)
+                return True, ""
+            except (ValueError, KeyError, TypeError):
+                return False, "Unparseable command. Hint: do you need to escape % chars?"
```
solarkennedy: Can you go ahead and spit out what the command was to the user that was unparseable?

solarkennedy:
ship

PR #930, opened 2016-12-10
Relax the message when a marathon app doesn't exist yet

solarkennedy:


Rob-Johnson:
I guess other monitoring will tell us if a given app can't be deployed?

PR #931, opened 2016-12-10
Fix get_healthcheck when nerve_ns is different

solarkennedy:
This is internal ticket PAASTA-7191.

This was another instance!=namespace mixup, but I think it only affect local-run.


solarkennedy:
Ug. @Rob-Johnson, I've "worked around" how we do this in local-run. trwtf is that we are calling a function from long_running_service_tools to get the nerve_ns, even if it is an an adhoc or chronos job.

You could argue I should move these functions even higher up into utils?

Rob-Johnson:
right, I see what you mean. Yeah, ``get_healthcheck_for_instance`` shouldn't live in ``long_running_service_tools``. Same with all the smartstack related bits in there - ``load_service_namespace_config``, ``ServiceNamespaceConfig``, ``InvalidSmartstackMode``, ``InvalidHealthcheckMode``. Isn't this what ``smartstack_tools`` was for?

solarkennedy:
Blarg. I'm going to punt.

PR #934, opened 2016-12-13
Check AWS instances register in mesos

mattmb:
This is deliberately a bit rough around the edges, it will be cleaner once I finish my other autoscaler branch. In the meantime I'd like to ship this out and set it up in sensu to see how noisy it will be.

--
This is a quick attempt to give us visibility of AWS resources that
don't end up registering in mesos.

This covers a lot of potential problems with the bootstrap process.

This could be better:
* maybe should check how long a resource has been hanging around for
* also check ASGs (I will update this when I merge my ASG branch)

### DIFF
--- a/paasta_tools/contrib/check_registered_slaves_aws.py
+++ b/paasta_tools/contrib/check_registered_slaves_aws.py
@@ -1,0 +1,49 @@
+#!/usr/bin/env python
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import argparse
+import sys
+
+from paasta_tools.autoscaling.autoscaling_cluster_lib import get_sfr
+from paasta_tools.autoscaling.autoscaling_cluster_lib import get_sfr_slaves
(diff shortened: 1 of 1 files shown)


solarkennedy:
Ship, let's try it out

paasta_tools/contrib/check_registered_slaves_aws.py
```diff
@@ -0,0 +1,51 @@
...
+import sys
+
+from paasta_tools.autoscaling.autoscaling_cluster_lib import get_sfr
+from paasta_tools.autoscaling.autoscaling_cluster_lib import get_sfr_slaves
+from paasta_tools.autoscaling.autoscaling_cluster_lib import get_spot_fleet_instances
+from paasta_tools.mesos_tools import get_mesos_master
+from paasta_tools.utils import load_system_paasta_config
+
+
+mesos_state = get_mesos_master().state
```
mattmb: Woops, good point. I'll fix and ship.

PR #935, opened 2016-12-13
support nerve namespaces in puppet service dir

somic:
My first paasta PR.

This is step 2 from internal PAASTA-7189.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -19,6 +19,7 @@ make the PaaSTA stack work.
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import json
 import logging
 import os
 import re
@@ -759,10 +760,16 @@ def get_classic_services_that_run_here():
             os.path.exists(os.path.join(PUPPET_SERVICE_DIR, i))
         }
 
+    puppet_service_dir_namespaces = {}
+    for service in puppet_service_dir_services:
+        with open(os.path.join(PUPPET_SERVICE_DIR, service)) as f:
+            puppet_service_dir_data = json.load(f)
+            puppet_service_dir_namespaces[service] = puppet_service_dir_data['namespaces']
+
     return sorted(
         service_configuration_lib.services_that_run_here() |
         puppet_service_dir_services
-    )
+    ), puppet_service_dir_namespaces
 
 
 def get_classic_service_information_for_nerve(name, soa_dir):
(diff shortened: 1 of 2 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -779,10 +786,16 @@ def _namespaced_get_classic_service_information_for_nerve(name, namespace, soa_d
...
-    for name in get_classic_services_that_run_here():
-        for namespace in get_all_namespaces_for_service(name, soa_dir, full_name=False):
-            classic_services.append(_namespaced_get_classic_service_information_for_nerve(
-                name, namespace[0], soa_dir))
+    classic_services_that_run_here, puppet_service_dir_namespaces = get_classic_services_that_run_here()
+    for service in classic_services_that_run_here:
+        if service in puppet_service_dir_namespaces:
+            for namespace in puppet_service_dir_namespaces[service]:
+                classic_services.append(
+                    _namespaced_get_classic_service_information_for_nerve(service, namespace, soa_dir))
```
solarkennedy: Can you use kwargs format here to make it more obvious what that second argument is?

paasta_tools/marathon_tools.py
```diff
@@ -759,10 +760,16 @@ def get_classic_services_that_run_here():
             os.path.exists(os.path.join(PUPPET_SERVICE_DIR, i))
         }
 
+    puppet_service_dir_namespaces = {}
+    for service in puppet_service_dir_services:
```
solarkennedy: I think the way this is organized is a little unintuitive.

I think it would be easier to understand if there were 3 functions that were concatenated:
* `get_classic_services_that_run_here`
* `get_puppet_services_that_run_here`
* `get_marathon_services_that_run_here`

I think if it was split up like this, then there wouldn't be to functions that sound really similar (get_classic_services_that_run_here, get_classic_services_running_here_for_nerve)
I think the confusion comes from the fast that you need both of these functions to correlate the puppet stuff together. I bet if the puppet parts were in their own thing, then you wouldn't have to do the whole "tuple" return thing.

Can you think about if this is better? If it does, please try it out, if not, I'm totally fine with working code.

paasta_tools/marathon_tools.py
```diff
@@ -779,10 +786,16 @@ def _namespaced_get_classic_service_information_for_nerve(name, namespace, soa_d
...
+    classic_services_that_run_here, puppet_service_dir_namespaces = get_classic_services_that_run_here()
+    for service in classic_services_that_run_here:
+        if service in puppet_service_dir_namespaces:
+            for namespace in puppet_service_dir_namespaces[service]:
+                classic_services.append(
+                    _namespaced_get_classic_service_information_for_nerve(service, namespace, soa_dir))
+        else:
+            for namespace in get_all_namespaces_for_service(service, soa_dir, full_name=False):
+                classic_services.append(
+                    _namespaced_get_classic_service_information_for_nerve(service, namespace[0], soa_dir))
```
solarkennedy: I'm confused, why on the puppet thing do we append the entire "namespace", but in this one we only do `namespace[0]` ?

PR #938, opened 2016-12-14
support nerve namespaces in puppet service dir

somic:
How about this instead?

This is a new PR that obsoletes https://github.com/Yelp/paasta/pull/935

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -19,6 +19,7 @@ make the PaaSTA stack work.
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import json
 import logging
 import os
 import re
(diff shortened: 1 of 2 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -779,10 +779,26 @@ def _namespaced_get_classic_service_information_for_nerve(name, namespace, soa_d
...
+    classic_services_here = service_configuration_lib.services_that_run_here()
+    puppet_services_here = get_puppet_services_that_run_here()
+    combined_services_here = sorted(
+        classic_services_here | set(puppet_services_here.keys())
+    )
+
+    for service in combined_services_here:
+        namespaces = []
+        if service in puppet_services_here:
+            namespaces += puppet_services_here[service]
```
solarkennedy: I was hoping that after this refactor `get_classic_services_running_here_for_nerve` wouldn't reference puppet at all, and that all puppet-related stuff could be in `get_puppet_services_that_run_here`.

But I see that it is a more efficient to kinda loop through the combined list here?

Again, my hope was that `get_services_running_here_for_nerve` would combine 3 sources: classic + puppet + marathon.

classic != puppet, so I was hoping to disconnect the two.


Here is an example of a bug that arises due to the current method: This loops through the puppet combined services, and *if* a service is in the puppet list, then we get the namespaces and add it to the return.

In other words, why do we get the classic list, get the puppet list, combine the two, and then loop over it again, just to ask "if service in puppet_services_here", we already got the puppet list?

Make two functions:
* `get_classic_services_running_here_for_nerve`: loops over `services_that_run_here` and attaches all_namespaces and returned the namespaced list
* `get_puppet_services_running_here_for_nerve`: loops over `puppet_services_here` and extracts the `namespaces` and attaches them and returns the combined list.

Modify `get_services_running_here_for_nerve` to combine all three together.

I think long term such a design is easier to understand, and easier to adjust the puppet parts, classic, parts, or marathon parts.

somic:
@solarkennedy this is now ready for another look

solarkennedy:
Looks great!

solarkennedy:
In the future, you don't have to make a new PR for each try, you can rebase and or force-push, or simply keep pushing commits, etc.

PR #940, opened 2016-12-15
fix the percentage for queued Chronos jobs

Rob-Johnson:
internal PAASTA-7231


paasta_tools/metrics/metastatus_lib.py
```diff
@@ -503,7 +503,7 @@ def assert_chronos_queued_jobs(client):
     num_jobs = len(chronos_tools.filter_enabled_jobs(client.list()))
 
     try:
-        perc_used = percent_used(all_jobs_queued, num_jobs)
+        perc_used = percent_used(num_jobs, all_jobs_queued)
```
solarkennedy: kwargs would make this a bit more obvious and may have helped catch the original bug

PR #942, opened 2016-12-16
Upgrade mesos to 1.1.0

nhandler:
We will need to do this upgrade before we can upgrade Marathon. Let's start testing it out.


mjksmith:
SHIPIT

mjksmith:
BE WARY FOR DATACENTRE GHOSTS

PR #944, opened 2016-12-27
wait-for-deployment improvements

oktopuz:
internal ticket: PAASTA-7205  

* wait-for-deployment should check status of all instances at least once even if timeout is small.
* wait-for_deployment should not re-check instances that are already deployed.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -335,8 +335,7 @@ def instances_deployed(cluster, service, instances, git_sha):
 
 
 def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
-    cluster_map = get_cluster_instance_map_for_service(soa_dir, service,
-                                                       deploy_group)
+    cluster_map = get_cluster_instance_map_for_service(soa_dir=soa_dir, service=service, deploy_group=deploy_group)
     if not cluster_map:
         _log(
             service=service,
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -260,125 +259,137 @@ def paasta_mark_for_deployment(args):
...
+                                 status.git_sha))
+            result.append(instance)
 
-    return sum(results)
+    return result
 
 
 def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
-    cluster_map = get_cluster_instance_map_for_service(soa_dir, service, deploy_group)
+    cluster_map = get_cluster_instance_map_for_service(soa_dir, service,
```
solarkennedy: please kwargs this. I'm surprised to see soa_dir as the first arg, not that it *matters* but I just hate arg order bugs.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -260,125 +259,137 @@ def paasta_mark_for_deployment(args):
...
-            level='event'
-        )
-        raise
-    return True
+        cluster['deployed'] = []
+    total_instances = sum([len(v["instances"]) for v in cluster_map.values()])
+    deadline = time.time() + timeout
+
+    with progressbar.ProgressBar(maxval=total_instances) as bar:
+        while time.time() < deadline:
```
solarkennedy: Does this have the "artifact" of having the progress bar finish at 100% even if the deadline is reached before actually finishing?

If so... is there any way around that? I think seeing the progress bar finish even if we raise the timeout is confusing.

paasta_tools/cli/cmds/mark_for_deployment.py
oktopuz: Done.

paasta_tools/cli/cmds/mark_for_deployment.py
oktopuz: This progress bar doesn't take time into account at all.
https://github.com/Yelp/paasta/pull/944/files#diff-c18610688c26664370ccca67b0151768R368

It calculates percentage of deployed instances, and it can only be at 100% if all instances are deployed.

PR #946, opened 2016-12-28
Always pass strings to the chooser

solarkennedy:
Always pass strings to this thing.

Internal ticket PAASTA-7266


nhandler:
Looks like there are still some test failures that need fixing.

solarkennedy:
Ready for review

PR #948, opened 2016-12-28
Make local-run healthchecking more strict about error codes. Fixes #678

### LINKED ISSUE #678: local-run behavior should be well-defined when the health check returns HTTP 1xx
perform_http_healthcheck has if/elif blocks for status codes 200-399 and 400+, but nothing for 0-199: https://github.com/Yelp/paasta/blob/01e0c901b19ab9a49e199f09fa5a400c0d3f9c49/paasta_tools/cli/cmds/local_run.py#L65

I suspect status code 1xx would cause a crash because this function would return None instead of a tuple.

solarkennedy:
Fixes #678

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -90,8 +90,8 @@ def perform_http_healthcheck(url, timeout):
     # check if response code is valid per https://mesosphere.github.io/marathon/docs/health-checks.html
     elif res.status_code >= 200 and res.status_code < 400:
         return (True, "http request succeeded, code %d" % res.status_code)
-    elif res.status_code >= 400:
-        return (False, "http request failed, code %d" % res.status_code)
+    else:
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -90,8 +90,8 @@ def perform_http_healthcheck(url, timeout):
     # check if response code is valid per https://mesosphere.github.io/marathon/docs/health-checks.html
     elif res.status_code >= 200 and res.status_code < 400:
         return (True, "http request succeeded, code %d" % res.status_code)
-    elif res.status_code >= 400:
-        return (False, "http request failed, code %d" % res.status_code)
+    else:
+        return (False, "http request failed, code %s" % str(res.status_code))
```
mattmb: Can we add a test case?

solarkennedy: I've added a specific test for 100 for @EvanKrall

PR #951, opened 2017-01-04
Safe load all the things

solarkennedy:
TRWTF is why safe_load isn't the default amirite?
Internal ticket is PAASTA-7208


hashbrowncipher:
👍 💯 🔐

PR #959, opened 2017-01-07
Make native mesos scheduler pass volumes, docker parameters, network mode, and choose a port.

EvanKrall:
Just a little Friday afternoon work to make the Paasta native Mesos scheduler a bit more feature-complete.

### DIFF
--- a/paasta_tools/native_mesos_scheduler.py
+++ b/paasta_tools/native_mesos_scheduler.py
@@ -465,6 +465,21 @@ class PaastaNativeServiceConfig(LongRunningServiceConfig):
         task.container.type = mesos_pb2.ContainerInfo.DOCKER
         task.container.docker.image = get_docker_url(system_paasta_config.get_docker_registry(),
                                                      self.get_docker_image())
+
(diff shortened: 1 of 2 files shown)


paasta_tools/native_mesos_scheduler.py
```diff
@@ -174,11 +181,22 @@ def start_task(self, driver, offer):
 
         while self.need_more_tasks(base_task.name) and \
                 remainingCpus >= task_cpus and \
-                remainingMem >= task_mem:
+                remainingMem >= task_mem and \
+                len(remainingPorts) >= 1:
+
+            task_port = random.choice(list(remainingPorts))
```
solarkennedy: Why not pick the next lowest? I think this is what marathon does.

EvanKrall: Marathon switched to random ports in mesosphere/marathon#1552. This makes it less likely that a container will come up on the same port as a recently-killed container for a different service and get traffic destined for the incorrect service. (This could happen if smartstack is delayed.)

PR #962, opened 2017-01-10
Fix multiple instances per reservation

mattmb:
SFRs only return one instance per reservation from the describe ec2
instances call. However ASGs do return several instances per
reservation. This makes it so we just build one list of instances which
is compatible for both.

Also suddenly today so do SFRs. I guess AWS changed something or we've just been "luckily" getting one instance per reservation in the past...

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -115,15 +115,16 @@ def describe_instances(instance_ids, region=None, instance_filters=None):
         instance_filters = []
     ec2_client = boto3.client('ec2', region_name=region)
     try:
-        instances = ec2_client.describe_instances(InstanceIds=instance_ids, Filters=instance_filters)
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -115,15 +115,16 @@ def describe_instances(instance_ids, region=None, instance_filters=None):
...
     except ClientError as e:
         if e.response['Error']['Code'] == 'InvalidInstanceID.NotFound':
             log.warn('Cannot find one or more instance from IDs {0}'.format(instance_ids))
             return None
         else:
             raise
-    ret = [reservation['Instances'][0] for reservation in instances['Reservations']]
-    return ret
+    instance_reservations = [reservation['Instances'] for reservation in instance_descriptions['Reservations']]
+    instances = [instance for reservation in instance_reservations for instance in reservation]
```
Rob-Johnson: I've always wondered why this flattening isn't in the stdlib.

PR #963, opened 2017-01-10
Bump local-run timeout from 60s->90s for travis

nhandler:
We are seeing `general_itests` fail due to local-run timing out (60s). This change tries bumping that timeout to 90s.


nhandler:
https://travis-ci.org/Yelp/paasta/jobs/190392006 is an example of a failing build

solarkennedy:
ship

PR #966, opened 2017-01-12
Correct task count objects

mattmb:
After the change in c266cf76d8e0e78b5f7889d61c9a4638f5fdd2d3 the
interaction with mesos_tools is incorrect. This corrects it so that we
pass the input mesos tools is expecting and then update the
PaastaAWSSlave object locally.

@Rob-Johnson can you sanity check this. I'd really like to do away with the weird dictionary that comes back from `get_mesos_task_count_by_slave`. But this feels like a slightly easier and safer change for now. This is just to merge it into my autoscaler-asg branch before I merge the lot to master and cut a release.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -218,7 +218,7 @@ class ClusterAutoscaler(object):
 
         :param slaves: list of slaves dict
         :returns: list of slaves dicts"""
-        return sorted(slaves, key=lambda x: (x['task_counts'].chronos_count, x['task_counts'].count))
+        return sorted(slaves, key=lambda x: (x.task_counts.chronos_count, x.task_counts.count))
 
     def scale_resource(self, current_capacity, target_capacity):
         """Scales an AWS resource based on current and target capacity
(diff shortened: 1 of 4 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -343,9 +343,12 @@ def downscale_aws_resource(self, filtered_slaves, current_capacity, target_capac
             current_capacity = new_capacity
             mesos_state = get_mesos_master().state_summary()
             if filtered_sorted_slaves:
-                filtered_slaves = get_mesos_task_count_by_slave(mesos_state, slaves_list=filtered_sorted_slaves)
-            else:
-                filtered_slaves = filtered_sorted_slaves
+                task_counts = get_mesos_task_count_by_slave(mesos_state,
+                                                            slaves_list=[{'task_counts': slave.task_counts}
```
Rob-Johnson: yeah, this is really nasty. You're kind of abusing the fact that {{get_mesos_tsak_count_by_slave}} doesn't do any type checking, plus that function is just nasty.

Rob-Johnson:
shipit, but let's please arrange a few hours to pair on refactoring this. It's nasty.

mattmb:
Yep, I definitely agree. I want to get out a fix for the task counting when marathon becomes unregistered. Once that's shipped I'll be happy to take a chainsaw to that function. Possibly later next week?

PR #967, opened 2017-01-13
Fix a script missed in refactor

mattmb:
This check was missed when I did the re-factor in
4fcbfb23d3e14ed94b8ba9cad9d9eb64e50d4c65

This is what's creating notification spam at the moment.


Rob-Johnson:
can you move this out of contrib if it's an important part of our monitoring?

PR #968, opened 2017-01-13
Fix/task count logic

mattmb:
This changes the way we select tasks in 2 places to include orphaned tasks:
* Where we count running tasks to decide if it is safe to kill a host
* Where we count resources used by running tasks to evaluate cluster utilization

@Rob-Johnson I think you are probably best placed to review this. But let's leave it until next week sometime. I'd like to have another read of what I've done before you go through it. I might even try running it in mesosstage if the itests pass.


paasta_tools/mesos/master.py
```diff
@@ -166,12 +166,17 @@ def slaves(self, fltr=""):
...
                 lambda x: fltr == x['id'], self.state['slaves'])))
 
-    def _task_list(self, active_only=False):
+    def _task_list(self, active_only=False, include_orphans=False):
         keys = ["tasks"]
         if not active_only:
             keys.append("completed_tasks")
+        frameworks = self._framework_list(active_only)
+        if include_orphans:
+            orphaned_tasks = self.state["orphaned_tasks"]
```
Rob-Johnson: this is ``orphan_tasks``

```
robj@robj-MacBookPro:~/configs  % curl -s http://paasta1-uswest1cdevc.dev.yelpcorp.com:5050/master/state | jq 'keys' |grep orphan
  "orphan_tasks",
```

paasta_tools/mesos/master.py
```diff
@@ -166,12 +166,17 @@ def slaves(self, fltr=""):
             itertools.ifilter(
                 lambda x: fltr == x['id'], self.state['slaves'])))
 
-    def _task_list(self, active_only=False):
+    def _task_list(self, active_only=False, include_orphans=False):
```
Rob-Johnson: I'm dubious if this is the right place for this. What do you think about making a new ``get_orphaned_tasks`` function and making the callers be explicit about wanting them?

paasta_tools/mesos/master.py
mattmb: Oh, thanks! I wonder if this will fix the itests...

paasta_tools/mesos/master.py
mattmb: Good question. Would be easy enough to move it "up a level". I did at least make the default to not include them.

I think there's an argument for keeping it here too. Along the lines of: "I'm looking to get all the tasks and don't know that orphans are a thing, I lookup the tasks function here and see the kwarg so I can now think about it and make a decision."

mattmb:
Still failing :-( I guess I'll try and recreate the failure...

paasta_tools/mesos_tools.py
```diff
@@ -139,8 +139,8 @@ def filter_not_running_tasks(tasks):
     return [task for task in tasks if not is_task_running(task)]
 
 
-def get_running_tasks_from_active_frameworks(job_id=''):
-    active_framework_tasks = get_current_tasks(job_id)
+def get_running_tasks_from_active_frameworks(job_id='', include_orphans=True):
+    active_framework_tasks = get_current_tasks(job_id, include_orphanse=include_orphans)
```
Rob-Johnson: here's your typo.

s/orphanse/orphans

mattmb: Yikes, thanks, I really shouldn't have rushed this so much...

I guess this implies poor unit tests too

mattmb:
@Rob-Johnson a07331c4c98bc25eb379e54476ad989b4a2692c0 should make this a less impactful change as discussed yesterday. It limits the inclusion of orphans to the areas I know to be broken without them.

mattmb:
@Rob-Johnson rebased and the tests are passing again :-) Can you take a look at the last commit?

paasta_tools/mesos_tools.py
```diff
@@ -139,13 +143,27 @@ def filter_not_running_tasks(tasks):
     return [task for task in tasks if not is_task_running(task)]
 
 
-def get_running_tasks_from_active_frameworks(job_id=''):
-    active_framework_tasks = get_current_tasks(job_id)
+def get_running_tasks_from_frameworks(job_id=''):
+    """ Will include active and completed frameworks but NOT orphaned
+    frameworks
```
Rob-Johnson: you mean orphaned tasks?

paasta_tools/mesos_tools.py
```diff
@@ -111,12 +111,16 @@ def is_mesos_leader(hostname=MY_HOSTNAME):
...
+def get_current_tasks(job_id, include_orphans=False):
     """ Returns a list of all the tasks with a given job id.
     :param job_id: the job id of the tasks.
     :return tasks: a list of mesos.cli.Task.
     """
-    return get_mesos_master().tasks(fltr=job_id, active_only=False)
+    mesos_master = get_mesos_master()
+    framework_tasks = mesos_master.tasks(fltr=job_id, active_only=False)
+    if include_orphans:
+        framework_tasks += mesos_master.orphan_tasks()
```
Rob-Johnson: but now the filtering doesn't do the correct thing? This will return all running tasks matching the filter, plus *all* the orphans if you set ``include_orphans=True``?

paasta_tools/mesos_tools.py
```diff
@@ -139,13 +143,27 @@ def filter_not_running_tasks(tasks):
...
+    """ Will include all running tasks, including orphans
+    """
+    framework_tasks = get_current_tasks('', include_orphans=True)
+    running_tasks = filter_running_tasks(framework_tasks)
+    return running_tasks
+
+
+def get_non_running_tasks_from_frameworks(job_id=''):
+    """ Will include active and completed frameworks but NOT orphaned
+    frameworks
```
Rob-Johnson: again, do you mean orphaned tasks?

paasta_tools/mesos_tools.py
```diff
@@ -425,12 +443,15 @@ def get_mesos_quorum():
...
-def get_all_tasks_from_state(mesos_state):
+def get_all_tasks_from_state(mesos_state, include_orphans=False):
     """Given a mesos state, find the tasks from all frameworks.
     :param mesos_state: the mesos_state
     :returns: a list of tasks
     """
-    return [task for framework in mesos_state.get('frameworks', []) for task in framework.get('tasks', [])]
+    tasks = [task for framework in mesos_state.get('frameworks', []) for task in framework.get('tasks', [])]
+    if include_orphans:
+        tasks += mesos_state.get('orphan_tasks', [])
```
Rob-Johnson: why don't you use your ``get_orphans`` function here?

paasta_tools/mesos_tools.py
mattmb: yeah I guess the filtering is ignored for orphans... I could add it in theory

I'm tempted to say we should ship this anyway just because this is a real bug that could cause an outage. That said I still haven't fully fixed the build...

paasta_tools/mesos_tools.py
mattmb: Mostly because this function already has the mesos state object so I don't want to make a call to mesos cli and risk adding another call to mesos api.

paasta_tools/mesos_tools.py
Rob-Johnson: sure, I get that, but I'd rather not ship bugs. the thing to do here is to move the ``mesos_master.orphan_tasks`` call to ``get_current_tasks``

paasta_tools/mesos_tools.py
Rob-Johnson: fair enough.

paasta_tools/mesos_tools.py
mattmb: Do you mean to `get_all_running_tasks` that would make sense because that function doesn't take a filter so you don't get any unexpected behaviour.

Rob-Johnson: yeah, I did, sorry.

mattmb: Ok, finally got round to sorting this! @Rob-Johnson can you take a look at the fixup. I'll squash it before I merge if you're happy.

PR #970, opened 2017-01-16
Revert "Use marathon-python compatible with marathon 1.4"

mattmb:
This reverts commit 392ff704c2926ece75abc850d73a8c9e949c216b.

Seeing some 422 errors when deploying marathon services. Yelp internal
ticket: PAASTA-7849

cc @Rob-Johnson

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,7 +22,7 @@ isodate==0.5.1
 jsonschema[format]==2.5.1
 kazoo==2.2
 lockfile==0.9.1
-marathon==0.8.9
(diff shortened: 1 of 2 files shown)


Rob-Johnson:
shipit

Rob-Johnson:
you should downgrade marathon to rc2 to get the itests passing too

mattmb:
So, this was totally a red-herring. The 422s were caused by bad config not an actual bug.

https://media.giphy.com/media/4pMX5rJ4PYAEM/giphy.gif

PR #972, opened 2017-01-19
async version of wait-for-deployment

oktopuz:
This change reduces wait-for-deployment running time (when git_sha is already deployed) from 4m50 to 1m12 for yelp-main prod.
It creates separate threads for each paasta cluster and up to 5 threads per cluster to query API.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -264,13 +264,19 @@ def paasta_mark_for_deployment(args):
 def instances_deployed(cluster, service, instances_in, instances_out, git_sha, green_light):
     """Create a thread pool to run _run_instance_worker()
 
-    Args:
-        green_light: an instance of threading.Event(). It is used to terminate
-            all running threads after KeyboardInterrupt.
-        instances_in (Queue): a thread-safe queue that contains all cluster
-            instances that need to be checked.
-        instances_out (Queue): a empty thread-safe queue. I will contain
-            instances that are not deployed yet.
+    :param cluster: the name of the cluster.
+    :param service: the name of the service.
+    :param instances_in: a thread-safe queue. Should contain all cluster
+                         instances that need to be checked.
+    :type instances_in: Queue
+    :param instances_out: a empty thread-safe queue. I will contain
+                          instances that are not deployed yet.
+    :type instances_out: Queue
+    :param git_sha: git sha marked for deployment.
+    :param green_light: an instance of threading.Event().
+                        It is supposed to be cleared when KeyboardInterrupt is
+                        received. All running threads should check it
+                        periodically and exit when it is cleared.
     """
     num_threads = min(5, instances_in.qsize())
 
@@ -292,6 +298,20 @@ def _run_instance_worker(cluster, service, instances_in,
 
     If an instance doesn't have the expected git_sha, add it to
     the instances_out queue to re-check it later.
+
+    :param cluster: the name of the cluster.
+    :param service: the name of the service.
+    :param instances_in: a thread-safe queue. Should contain all cluster
+                         instances that need to be checked.
+    :type instances_in: Queue
+    :param instances_out: a empty thread-safe queue. I will contain
+                          instances that are not deployed yet.
+    :type instances_out: Queue
+    :param git_sha: git sha marked for deployment.
+    :param green_light: an instance of threading.Event().
+                        It is supposed to be cleared when KeyboardInterrupt is
+                        received. All running threads should check it
+                        periodically and exit when it is cleared.
     """
 
     api = client.get_paasta_api_client(cluster=cluster)
@@ -380,9 +400,14 @@ def _run_instance_worker(cluster, service, instances_in,
 def _query_clusters(cluster_map, service, git_sha, green_light):
     """Run _run_cluster_worker() in a separate thread for each paasta cluster
 
-    Args:
-        green_light: an instance of threading.Event(). It is used to terminate
-            all running threads after KeyboardInterrupt.
+    :param cluster_map: see paasta_tools.generate_deployments_for_service.get_cluster_instance_map_for_service
+    :param cluster: the name of the cluster.
+    :param service: the name of the service.
+    :param git_sha: git sha marked for deployment.
+    :param green_light: an instance of threading.Event().
+                        It is supposed to be cleared when KeyboardInterrupt is
+                        received. All running threads should check it
+                        periodically and exit when it is cleared.
     """
     for cluster, cluster_data in cluster_map.items():
         if cluster_data['deployed'] != cluster_data['num_instances']:
@@ -398,12 +423,22 @@ def _query_clusters(cluster_map, service, git_sha, green_light):
                 time.sleep(.2)
         except (KeyboardInterrupt, SystemExit):
             green_light.clear()
-            paasta_print('KeyboardInterrut received. Terminating..')
+            paasta_print('KeyboardInterrupt received. Terminating..')
         cluster_data['worker'].join()
 
 
 def _run_cluster_worker(cluster, cluster_data, service, git_sha, green_light):
-    """Run instances_deployed() for a cluster"""
+    """Run instances_deployed() for a cluster
+
+    :param cluster: the name of the cluster.
+    :param cluster_data: a reference to cluster_map[cluster]
+    :param service: the name of the service.
+    :param git_sha: git sha marked for deployment.
+    :param green_light: an instance of threading.Event().
+                        It is supposed to be cleared when KeyboardInterrupt is
+                        received. All running threads should check it
+                        periodically and exit when it is cleared.
+    """
     instances_out = Queue()
     instances_deployed(cluster=cluster, service=service,
                        instances_in=cluster_data['queue'],
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -410,7 +410,7 @@ def _query_clusters(cluster_map, service, git_sha, green_light):
                         periodically and exit when it is cleared.
     """
     for cluster, cluster_data in cluster_map.items():
-        if cluster_data['deployed'] != cluster_data['num_instances']:
+        if cluster_data['deployed'] != len(cluster_data['instances']):
             worker = Thread(target=_run_cluster_worker,
                             args=(cluster, cluster_data, service, git_sha,
                                   green_light))
@@ -445,10 +445,10 @@ def _run_cluster_worker(cluster, cluster_data, service, git_sha, green_light):
                        instances_out=instances_out, git_sha=git_sha,
                        green_light=green_light)
     if cluster_data['queue'].empty():
-        cluster_data['deployed'] = (cluster_data['num_instances'] -
+        cluster_data['deployed'] = (len(cluster_data['instances']) -
                                     instances_out.qsize())
     cluster_data['queue'] = instances_out
-    if cluster_data['deployed'] == cluster_data['num_instances']:
+    if cluster_data['deployed'] == len(cluster_data['instances']):
         instance_csv = ", ".join(cluster_data['instances'])
         paasta_print("Deploy to {} complete! (instances: {})"
                      .format(cluster, instance_csv))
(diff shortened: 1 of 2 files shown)


oktopuz:
I'll fix Flake8 tests tomorrow.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -323,15 +366,57 @@ def instances_deployed(cluster, service, instances, git_sha):
...
                                  status.git_sha))
-            result.append(instance)
+            instances_in.task_done()
+
 
-    return result
+def _query_clusters(cluster_map, service, git_sha, green_light):
+    """Run _run_cluster_worker() in a separate thread for each paasta cluster
+
+    Args:
```
Rob-Johnson: also, please add the remaining args

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -323,15 +366,57 @@ def instances_deployed(cluster, service, instances, git_sha):
...
+        green_light: an instance of threading.Event(). It is used to terminate
+            all running threads after KeyboardInterrupt.
+    """
+    for cluster, cluster_data in cluster_map.items():
+        if cluster_data['deployed'] != cluster_data['num_instances']:
+            worker = Thread(target=_run_cluster_worker,
+                            args=(cluster, cluster_data, service, git_sha,
+                                  green_light))
+            worker.start()
+            cluster_data['worker'] = worker
```
Rob-Johnson: again, please don't mutate things like this. use a NamedTuple to represent closely coupled objects, rather than stuffing it all into a dict.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -323,15 +366,57 @@ def instances_deployed(cluster, service, instances, git_sha):
...
+
+    for cluster, cluster_data in cluster_map.items():
+        try:
+            while green_light.is_set() and cluster_data['worker'].isAlive():
+                time.sleep(.2)
+        except (KeyboardInterrupt, SystemExit):
+            green_light.clear()
+            paasta_print('KeyboardInterrut received. Terminating..')
```
Rob-Johnson: typo.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -258,16 +261,50 @@ def paasta_mark_for_deployment(args):
     return ret
 
 
-def instances_deployed(cluster, service, instances, git_sha):
-    result = []
+def instances_deployed(cluster, service, instances_in, instances_out, git_sha, green_light):
+    """Create a thread pool to run _run_instance_worker()
+
+    Args:
```
Rob-Johnson: please change the formatting here to match the rest of our codebase.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -347,30 +432,33 @@ def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
...
+    for cluster in cluster_map:
+        cluster_map[cluster]['deployed'] = 0
+        cluster_map[cluster]['queue'] = Queue()
+        cluster_map[cluster]['num_instances'] = len(cluster_map[cluster]['instances'])
+        total_instances += cluster_map[cluster]['num_instances']
+        for i in cluster_map[cluster]['instances']:
+            cluster_map[cluster]['queue'].put(i)
     deadline = time.time() + timeout
+    green_light = Event()
+    green_light.set()
```
Rob-Johnson: I'm not keen on how you're passing this event around. let me have a think about how to coordinate this better

paasta_tools/cli/cmds/mark_for_deployment.py
oktopuz: fixed.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -258,16 +261,50 @@ def paasta_mark_for_deployment(args):
...
+
+    Args:
+        green_light: an instance of threading.Event(). It is used to terminate
+            all running threads after KeyboardInterrupt.
+        instances_in (Queue): a thread-safe queue that contains all cluster
+            instances that need to be checked.
+        instances_out (Queue): a empty thread-safe queue. I will contain
+            instances that are not deployed yet.
+    """
+    num_threads = min(5, instances_in.qsize())
```
oktopuz: This limits the number of simultaneous requests to a cluster.
Otherwise if a service had 100 instances it would query all 100 at once, because they all would be in the instances_in queue.

paasta_tools/cli/cmds/mark_for_deployment.py
oktopuz: Done.

paasta_tools/cli/cmds/mark_for_deployment.py
oktopuz: I can't think of any reason to use NamedTuples here.
Could you explain this in more details please?

cluster_map[cluster]['worker'] as well as cluster_map[cluster]['queue'] is a reference.
I'm using them in an internal function that is not supposed to call by anyone outside of wait-for-deployment.
What is wrong with keeping them in the same dict as other attributes?

paasta_tools/cli/cmds/mark_for_deployment.py
oktopuz: Fixed.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -347,30 +432,33 @@ def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
...
     paasta_print("Waiting for deployment of {0} for '{1}' complete..."
                  .format(git_sha, deploy_group))
-    for cluster in cluster_map.values():
-        cluster['deployed'] = []
-    total_instances = sum([len(v["instances"]) for v in cluster_map.values()])
+
+    total_instances = 0
+    for cluster in cluster_map:
+        cluster_map[cluster]['deployed'] = 0
+        cluster_map[cluster]['queue'] = Queue()
```
oktopuz: again, could you explain why it is acceptable to add ['deployed'], but not ['queue']?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -347,30 +432,33 @@ def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
...
+        cluster_map[cluster]['num_instances'] = len(cluster_map[cluster]['instances'])
+        total_instances += cluster_map[cluster]['num_instances']
```
oktopuz: so that not to run the same loop on instances once again just to calculate one more value.

paasta_tools/cli/cmds/mark_for_deployment.py
oktopuz: I need to pass something to be able to exit from running threads after Ctrl+C.
I don't know any nice way of doing this in Python2.

On the other hand it is being passed the same way as 'service', 'git_sha' and etc.
Why not?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -347,30 +432,33 @@ def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
...
-                                   not in cluster_map[cluster]['deployed']],
-                        git_sha=git_sha)
-                    if len(cluster_map[cluster]['deployed']) == len(cluster_map[cluster]['instances']):
-                        instance_csv = ", ".join(cluster_map[cluster]['instances'])
-                        paasta_print("Deploy to %s complete! (instances: %s)" % (cluster, instance_csv))
-                bar.update(sum([len(v["deployed"]) for v in cluster_map.values()]))
-            if all([len(cluster['deployed']) == len(cluster["instances"]) for cluster in cluster_map.values()]):
+            _query_clusters(cluster_map, service, git_sha, green_light)
+            if not green_light.is_set():
+                raise KeyboardInterrupt
```
oktopuz: I'm afraid I disagree. This is exactly the correct exception.

This event isn't set only if KeyboardInterrupt was received by a running thread.
https://github.com/Yelp/paasta/pull/972/files#diff-c18610688c26664370ccca67b0151768R424
I'm just re-raising the same exception in the main thread to follow the same code path as it was before.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -347,30 +432,33 @@ def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
...
                  .format(git_sha, deploy_group))
-    for cluster in cluster_map.values():
-        cluster['deployed'] = []
-    total_instances = sum([len(v["instances"]) for v in cluster_map.values()])
+
+    total_instances = 0
+    for cluster in cluster_map:
+        cluster_map[cluster]['deployed'] = 0
+        cluster_map[cluster]['queue'] = Queue()
+        cluster_map[cluster]['num_instances'] = len(cluster_map[cluster]['instances'])
```
oktopuz: No difference indeed.
The idea was to create a short reference for 'len(cluster_map[cluster]['instances'])' but something went wrong.
I'll revert it.

paasta_tools/cli/cmds/mark_for_deployment.py
Rob-Johnson: @oktopuz Hm, you're right, maybe this is ok if you're only using the dict in the scope of this function. 
But here's why I'm generally against this:

Consider this:

```python

def foo(bar):
   bar['y'] = 'hi'

def baz(bar):
   print bar['y']

def qux(bar):
   print bar['x']
```

We use the common arg ``bar`` everywhere, but it means different things in different places. ``foo`` requires just a dict, ``baz`` requires a dict with the key ``y``, and ``qux`` requires a dict with the key ``x``. This doesn't really get encoded anywhere though; the caller has to know the implementation details of the function to know how to give it the 'right' thing. If you make NamedTuples for ``foo`` ``baz`` and ``qux``, you put a contract around what they have to look like.

In our case, I'd really like ``cluster_map`` and ``cluster_data`` dicts to be replaced with ``NamedTuple``. Then, rather than mutating it, you create new data structures that encode the coupling of ``cluster_map`` or ``cluster_data`` and some other thing (the worker, in this case). At the same time, I get that this wasn't your doing, so I don't expect you to refactor that. But, as a halfway house, I'd do this:

```python
workers = []
for cluster, cluster_data in cluster_map.items():
    if cluster_data['deployed'] != len(cluster_data['instances']):
             worker = Thread(target=_run_cluster_worker,
                             args=(cluster, cluster_data, service, git_sha,
                                   green_light))
             worker.start()
             workers.append(worker)   
            
for worker in workers:
    try:
        while green_light.is_set() and worker.isAlive():
            time.sleep(.2)
    except (KeyboardInterrupt, SystemExit):
        green_light.clear()
        paasta_print('KeyboardInterrupt received. Terminating..')
        cluster_data['worker'].join()
```

As far as I can tell, the second for loop doesn't the cluster_map for anything other than the worker, so why not just store each of the workers in a separate list?

paasta_tools/cli/cmds/mark_for_deployment.py
Rob-Johnson: eh, you're right, I think this is fine. Do we need to be graceful, or can we set ``daemon=True`` to just make these exit immediately when the main thread does.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -323,15 +386,72 @@ def instances_deployed(cluster, service, instances, git_sha):
...
+            cluster_data['worker'] = worker
+
+    for cluster, cluster_data in cluster_map.items():
+        try:
+            while green_light.is_set() and cluster_data['worker'].isAlive():
+                time.sleep(.2)
+        except (KeyboardInterrupt, SystemExit):
+            green_light.clear()
+            paasta_print('KeyboardInterrupt received. Terminating..')
+        cluster_data['worker'].join()
```
Rob-Johnson: do we need timeouts for this and the other ``join``, given they'll block indefinitely?

paasta_tools/cli/cmds/mark_for_deployment.py
oktopuz: This is refactored.

I couldn't use NamedTuple, because I needed to replace queues.
Therefore I used a class (ClusterData).

paasta_tools/cli/cmds/mark_for_deployment.py
oktopuz: I tried ```daemon=True``` and it didn't work well for KeyboardInterrupt because the interrupt can be received by any thread. Therefore we need an Event() anyway to terminate the main thread at least.

paasta_tools/cli/cmds/mark_for_deployment.py
oktopuz: It would block indefinitely if there was an infinite loop in the code.
However such loop would block tests as well. In other words it wouldn't come to master unnoticed.

Also if it blocks indefinitely, I would rather want someone to press Ctrl+C and exit via KeyboardInterrupt, or kill the process and let us know, instead of leaving the blocked thread to run indefinitely in the background.

oktopuz:
Done.

PR #973, opened 2017-01-20
Check marathon IDs are valid

mattmb:
This should validate the top level key in marathon-<cluster>.yaml which
becomes the marathon app ID. It compares it to an adapted regex that
marathon itself uses to validate the app ID.

https://github.com/Yelp/paasta/compare/validate-marathon-ids?expand=1#diff-0f83d210889c5e0c676f023e9b29c835R8 is the new line (lol structured data in a diff)

@solarkennedy this should do what you suggested


mattmb:
Note to self: just found one yelp instance with a '-' in it. So I guess that is valid too.

I'll update the regex before I ship this...

PR #974, opened 2017-01-24
Bump six package to fix build

mattmb:
sphinxcontrib-programoutput requires a newer version


mattmb:
https://github.com/Yelp/paasta/pull/976

PR #975, opened 2017-01-25
Remove use of the sphinxcontrib-programoutput module in docs

nhandler:
This module has caused us issues on a number of occasions. We are using it minimally, and nobody really seems to care about it. Let's just remove it and go back to static lists.


mattmb:
As you said @nhandler I think this was a red-herring! But man the pip error messages are terrible.

See https://github.com/Yelp/paasta/pull/976 for what I think is really up

mattmb:
I'd still say this might be worth shipping but it's also fine to just close if you feel we should keep this in.

PR #976, opened 2017-01-25
Fix the build

mattmb:
Think I've tracked this down to the fact that we have pinned six==1.9.0
and setuptools now needs a newer version.

Can't see any reason why it was pinned so let's see if the tests pass
with the new version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -50,7 +50,7 @@ retry==0.9.2
 sensu-plugin==0.1.0
 service-configuration-lib==0.10.1
 simplejson==3.6.5
-six==1.9.0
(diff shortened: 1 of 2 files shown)


mattmb:
Ok so what I think was up:
* setuptools new version
* this needed a later version of six which conflicted with our pin
* new setuptools breaks for certain versions of pip
* virtualenv installs who knows what version of pip when you run tox -e blah...

Upgrading pip + virtualenv seems to do the trick. We may also need to do this in the lucid image, xenial seems to "just work", probably because it has a later version of virtualenv/pip packaged.

PR #980, opened 2017-01-26
Bounce oracle change

mattmb:
This adds a new option to list_marathon_instances which lists all the
instances in one go but puts the instances expected to bounce first.
This way we can have one cron job (and so no race conditions) whilst
still speeding up the bounce by starting with the instances we expect to
bounce.


solarkennedy:
As discussed, this won't really help make the bounces faster due to the way that xargs works.

mattmb:
Agreed. This would actually be a sort of intermediate where the bounces would be prioritised but not as quickly as if they were in their own cron job. I think we can do something more clever long term but for now I'm closing this in favour of #991

PR #981, opened 2017-01-26
Add intermediate venv

mattmb:
I know this is kinda nuts but it does make the tests pass on a dev box.
I'm basically adding an intermediate venv which we install the latest
versions of pip + virtualenv. This way tox creates a virtualenv with
sufficiently new tools to work consistently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 build/
 dist/
 .tox
+.paasta
 .pip/
 *.py?
 .*.sw?
--- a/Makefile
+++ b/Makefile
@@ -26,8 +26,7 @@ docs:
 	tox -e docs
 
 test:
-	rm -rf .tox
-	tox
+	./test.sh
 
 itest: test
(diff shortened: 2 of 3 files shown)


mattmb:
This is to make the tests pass on Jenkins as well as travis

Rob-Johnson:
why don't we make this run in a container?

nhandler:
A container makes sense, although it will be unfortunate to have that extra overhead for unit tests. But can we merge this and iterate? Not being able to easily run tests locally makes development tougher.

Rob-Johnson:
yeah, shipit so we can do some work today. mmb is OOO, so we can merge it for him.

mattmb:
Thanks for shipping this :-) Looks like we've still got issues with the itests, I'll see if I can apply a similar fix...

I kinda agree that we shouldn't have this in a container. I know we need Docker for the itests but it'd be nice not to make it a dependency on running the unit tests too.

PR #989, opened 2017-01-31
try adding a logging mixin to the autoscaler

Rob-Johnson:
@mattmb is this really dumb? I'm hoping to encode a bit more information about 'what' is being autoscaled without having to add ``id: %s" % self.resource["id"]`` to every log line.


Rob-Johnson:
I checked this on mesosstage and it seemed to do the right thing.

PR #991, opened 2017-01-31
Move bounce lock to higher level

mattmb:
Currently we lock the `do_bounce` function on the service instance we are
bouncing. This moves the lock higher up to lock the `setup_service`
funcion. This should mean there can be no race conditions between two
instantiations of setup_marathon_job.

Note that this also makes failing to get the lock for a service a
non-fatal error. i.e. we just log and carry on. This is neccessary
because we sometimes expect two bounces to run at around the same time
so skipping one because it is in progress elsewhere isn't really a
problem.

The motivation for this fix is that currently two invocations of
`setup_marathon_job` can gather the same info about a service and both
decide they need to create a new marathon app. Although only one gets a
lock to do so the second may eventually get the lock and gets an error
from marathon when it tries to create an app that doesn't exist.

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -399,8 +399,6 @@ def deploy_service(
             instance=instance
         )
 
-    short_id = marathon_tools.format_job_id(service, instance)
-
     system_paasta_config = load_system_paasta_config()
     cluster = system_paasta_config.get_cluster()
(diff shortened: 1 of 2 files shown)


mattmb:
@solarkennedy I think this is the most sensible fix to our current situation (assuming the itests pass). What do you think?

Rob-Johnson:
lg2m

PR #992, opened 2017-01-31
Latest pip + venv on lucid too

mattmb:
As per 95ed63dface91540b32e6801674bf5b1ffc78c0f

As mentioned in https://github.com/Yelp/paasta/pull/976 I thought we
might need to bump the lucid versions too...

### DIFF
--- a/yelp_package/dockerfiles/lucid/Dockerfile
+++ b/yelp_package/dockerfiles/lucid/Dockerfile
@@ -20,6 +20,9 @@ MAINTAINER Kyle Anderson <kwa@yelp.com>
 # Which at this time is in this package
 RUN apt-get update && apt-get -y install dpkg-dev python-tox python-setuptools python-dev libssl-dev libffi-dev debhelper dh-virtualenv python-yaml libyaml-dev python-pytest pyflakes python2.7 python2.7-dev help2man zsh git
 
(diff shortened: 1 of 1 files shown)


yelp_package/dockerfiles/lucid/Dockerfile
```diff
@@ -20,6 +20,9 @@ MAINTAINER Kyle Anderson <kwa@yelp.com>
 # Which at this time is in this package
 RUN apt-get update && apt-get -y install dpkg-dev python-tox python-setuptools python-dev libssl-dev libffi-dev debhelper dh-virtualenv python-yaml libyaml-dev python-pytest pyflakes python2.7 python2.7-dev help2man zsh git
 
+RUN pip install -U pip
+RUN pip install -U virtualenv
```
solarkennedy: This may fix the build temporarily, but this kind of pip install is what causes breakage in the long term because it make all of our hard work of pinning stuff go to waste

mattmb: Yep, although we weren't really pinning before because we are installing these from apt anyway, so just relying on the docker cache to keep them somewhat the same.

I'd say that if the build becomes stable we can change these lines to `pip install pip==x.y.z` etc

Issue #993, opened 2017-01-31
test_mark_for_deployment occasionally hangs


Rob-Johnson:
@oktopuz can you help us out here please? I think this is something we discussed in the original PR https://github.com/Yelp/paasta/pull/972#discussion-diff-97113954R427 - we're seeing the tests occasionally hang indefinitely

oktopuz:
Sure. I'm on it.

oktopuz:
Fixed by #995

PR #995, opened 2017-02-01
Remove possible block while getting next instance from instances_queue

oktopuz:
#993 

Queue.get() is a blocking operation.
When there is only one instance in the queue, it is not empty, but only one thread will be able to read from it, while all other threads will be blocked.

block=False causes Queue.get() not to block and to raise an exception if nothing is available, and that will terminate the running thread.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -21,6 +21,7 @@ from __future__ import unicode_literals
 import logging
 import sys
 import time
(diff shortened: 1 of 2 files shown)


oktopuz:
ping @Rob-Johnson

Rob-Johnson:
will the user be exposed to the exception? I don't really want the output to be sprawled with exception messages.

oktopuz:
I thought it wouldn't send traceback to stdout, but it did in fact.
Merged.

PR #999, opened 2017-02-02
Relax service autoscaler task counting

mattmb:
This would need tests (and testing). Just posting for some feedback. Do we think this is sane? Maybe we should just hunt down these "healthcheckless" tasks and trigger a bounce for that app as @Rob-Johnson suggested?

Not included yet is converting this check to a threshold (which I still think is a good idea).

---
This makes two changes:
* Use the long marathon app_id. This means we should now only be
autoscaling the latest version of a paasta service. It also means we
can ignore any interaction with the bounce becase we are ignoring the
old service.
* Assume missing health checks are "normal" for tasks that haven't just
started. This is due to a bug in marathon where tasks can stop being
healthchecked but remain running and healthy.


Rob-Johnson:
lg2m

PR #1000, opened 2017-02-02
Limit downscaling by 30% of current_instances.

oktopuz:
internal ticket PAASTA-8027

* Limit how many instances to kill at a time.
* Additional log record when the number of instances changes.
* More tests for autoscale_marathon_instance()

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -240,13 +240,14 @@ def http_metrics_provider(marathon_service_config, marathon_tasks, endpoint='sta
 
 
 @register_autoscaling_component('mesos_cpu', SERVICE_METRICS_PROVIDER_KEY)
-def mesos_cpu_metrics_provider(marathon_service_config, marathon_tasks, mesos_tasks, **kwargs):
+def mesos_cpu_metrics_provider(marathon_service_config, marathon_tasks, mesos_tasks, log_utilization_data={}, **kwargs):
     """
     Gets the mean cpu utilization of a service across all of its tasks.
 
     :param marathon_service_config: the MarathonServiceConfig to get data from
     :param marathon_tasks: Marathon tasks to get data from
     :param mesos_tasks: Mesos tasks to get data from
+    :param log_utilization_data: A dict used to transfer utilization data to autoscale_marathon_instance()
 
     :returns: the service's mean utilization, from 0 to 1
     """
@@ -262,6 +263,7 @@ def mesos_cpu_metrics_provider(marathon_service_config, marathon_tasks, mesos_ta
         try:
             last_time, _ = zk.get(zk_last_time_path)
             last_cpu_data, _ = zk.get(zk_last_cpu_data)
+            log_utilization_data[last_time] = last_cpu_data
             last_time = float(last_time)
             last_cpu_data = (datum for datum in last_cpu_data.split(',') if datum)
         except NoNodeError:
@@ -279,6 +281,7 @@ def mesos_cpu_metrics_provider(marathon_service_config, marathon_tasks, mesos_ta
         raise MetricsProviderNoDataError("Couldn't get any cpu data from Mesos")
 
     cpu_data_csv = ','.join('%s:%s' % (cpu_seconds, task_id) for task_id, cpu_seconds in mesos_cpu_data.items())
+    log_utilization_data[str(current_time)] = cpu_data_csv
 
     with ZookeeperPool() as zk:
         zk.ensure_path(zk_last_cpu_data)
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -353,10 +355,26 @@ def autoscale_marathon_instance(marathon_service_config, marathon_tasks, mesos_t
...
     )
 
-    new_instance_count = marathon_service_config.limit_instance_count(current_instances + autoscaling_amount)
+    # Limit downscaling by 30% of current_instances until we find out what is
+    # going on in such situations
+    new_instance_count = max(current_instances + autoscaling_amount, int(current_instances * 0.7))
+
+    new_instance_count = marathon_service_config.limit_instance_count(new_instance_count)
     if new_instance_count != current_instances:
         write_to_log(
```
solarkennedy: This is going to be pretty noisy. Can you only log this debug if new_instance_count == int(current_instances * 0.7) ?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -353,10 +355,26 @@ def autoscale_marathon_instance(marathon_service_config, marathon_tasks, mesos_t
...
             config=marathon_service_config,
+            line=('decision_policy %s, metrics_provider %s, utilization %.2f,'
+                  ' error %.2f, autoscaling_amount %d, limited_downscaling %r'
```
solarkennedy: Even with this debug output, I don't think it will have the actual input data we need. We don't have the inputs to the utilization. (all the other stuff can more or less be calcualted after the fact

paasta_tools/autoscaling/autoscaling_service_lib.py
oktopuz: Done.

paasta_tools/autoscaling/autoscaling_service_lib.py
oktopuz: mesos_cpu_metrics_provider() doesn't know whether cpu time is too high or too low for the service.
Therefore I added a new param to transfer cpu_data from mesos_cpu_metrics_provider() to autoscale_marathon_instance() and write it to the log only when the autoscaler scales down too many instances.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -240,13 +240,14 @@ def utilization_mapper(json):
...
 @register_autoscaling_component('mesos_cpu', SERVICE_METRICS_PROVIDER_KEY)
-def mesos_cpu_metrics_provider(marathon_service_config, marathon_tasks, mesos_tasks, **kwargs):
+def mesos_cpu_metrics_provider(marathon_service_config, marathon_tasks, mesos_tasks, log_utilization_data={}, **kwargs):
     """
     Gets the mean cpu utilization of a service across all of its tasks.
 
     :param marathon_service_config: the MarathonServiceConfig to get data from
     :param marathon_tasks: Marathon tasks to get data from
     :param mesos_tasks: Mesos tasks to get data from
+    :patam log_utilization_data: A dict used to transfer utilization data to autoscale_marathon_instance()
```
mattmb: typo param

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -353,8 +359,20 @@ def autoscale_marathon_instance(marathon_service_config, marathon_tasks, mesos_t
...
+    # going on in such situations
+    safe_downscaling_threshold = int(current_instances * 0.7)
+    new_instance_count = max(current_instances + autoscaling_amount, safe_downscaling_threshold)
+
+    new_instance_count = marathon_service_config.limit_instance_count(new_instance_count)
     if new_instance_count != current_instances:
+        if new_instance_count == safe_downscaling_threshold:
+            write_to_log(
+                config=marathon_service_config,
+                line='Autoscaler clamped: %s' % str(log_utilization_data),
```
mattmb: Maybe I'm missing something? But I can't see where this is returned from the metric provider. Isn't it always going to be `{}`?

paasta_tools/autoscaling/autoscaling_service_lib.py
oktopuz: fixed. Thanks!

paasta_tools/autoscaling/autoscaling_service_lib.py
oktopuz: It isn't returned, it is supposed to be passed by reference from autoscale_marathon_instance(): https://github.com/Yelp/paasta/pull/1000/files#diff-699ef1f9441fd1dab1f660ff54347041R334

Each metrics_provider returns a number, and I wanted to keep it this way, otherwise I would have to adjust all of them.

mattmb: Right, gotcha. That seems okay to me

oktopuz:
merged.

PR #1001, opened 2017-02-02
Dedupe volumes when specified twice in a config

solarkennedy:
I thought for sure docker would tolerate this.

Internal ticket PAASTA-7920.

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -394,6 +394,11 @@ class InstanceConfig(dict):
         """
         return self.config_dict.get('net', 'bridge')
 
+    def get_volumes(self, system_volumes):
+        volumes = system_volumes + self.get_extra_volumes()
+        deduped = {v['containerPath'] + v['hostPath']: v for v in volumes}.values()
+        return sorted(deduped)
+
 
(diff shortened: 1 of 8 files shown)


paasta_tools/utils.py
```diff
@@ -394,6 +394,11 @@ def get_net(self):
         """
         return self.config_dict.get('net', 'bridge')
 
+    def get_volumes(self, system_volumes):
+        volumes = system_volumes + self.get_extra_volumes()
+        deduped = {v['containerPath'] + v['hostPath']: v for v in volumes}.values()
```
mattmb: This seems pretty good. Although it doesn't dictate which config "wins".

One possible problem: say system config has `/thing` mounted to `/thing` and the service config has`/another` mounted to `/thing/sub` then we know that definitely wont work right? Can we catch that case to?

Rob-Johnson: mmb: sure it does - extra_volumes will always have priority

solarkennedy:
Python 3 tests failing...

paasta_tools/utils.py
solarkennedy: I've added a test to assert this and updated the docs.

PR #1002, opened 2017-02-02
Make more tests pass in python3

asottile:
This gets ~some of the tests passing in python3 and gives us a way to whitelist in file-by-file as files are converted to passing (while preventing them from regressing)

### DIFF
--- a/tests/test_adhoc_tools.py
+++ b/tests/test_adhoc_tools.py
@@ -14,8 +14,6 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
-import contextlib
-
 import mock
 
 from paasta_tools import adhoc_tools
@@ -24,11 +22,9 @@ from paasta_tools.utils import NoConfigurationForServiceError
 
 
 def test_get_default_interactive_config():
-    with contextlib.nested(
-        mock.patch('paasta_tools.adhoc_tools.load_adhoc_job_config', autospec=True),
-    ) as (
-        mock_load_adhoc_job_config,
-    ):
+    with mock.patch(
+        'paasta_tools.adhoc_tools.load_adhoc_job_config', autospec=True,
+    ) as mock_load_adhoc_job_config:
         mock_load_adhoc_job_config.return_value = adhoc_tools.AdhocJobConfig(
             service='fake_service',
             instance='interactive',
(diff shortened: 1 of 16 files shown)


tests/test_adhoc_tools.py
```diff
@@ -48,15 +44,13 @@ def test_get_default_interactive_config():
...
-        mock.patch('paasta_tools.adhoc_tools.load_adhoc_job_config', autospec=True),
-        mock.patch('paasta_tools.adhoc_tools.load_v2_deployments_json', autospec=True),
-    ) as (
-        mock_prompt_pick_one,
-        mock_load_adhoc_job_config,
-        mock_load_deployments_json,
-    ):
+    with mock.patch(
+        'paasta_tools.adhoc_tools.prompt_pick_one', autospec=True,
+    ) as mock_prompt_pick_one, mock.patch(
```
EvanKrall: @asottile is this really the standard convention for this? This seems like a really awkward way of joining these together.

I'd really prefer a style where each `mock.patch(...) as foo` stays on one line, if possible.

EvanKrall: Seems like this parses, but flake8 is unhappy:

        with \
            mock.patch('paasta_tools.check_marathon_services_replication.send_event', autospec=True) \
                as mock_send_event, \
            mock.patch('paasta_tools.marathon_tools.read_registration_for_service_instance', autospec=True,
                       return_value=compose_job_id(service, instance)), \
            mock.patch('paasta_tools.check_marathon_services_replication.load_smartstack_info_for_service', autospec=True) \
                as mock_load_smartstack_info_for_service, \
            mock.patch('paasta_tools.marathon_tools.load_marathon_service_config', autospec=True) \
                as mock_load_marathon_service_config\
        :

asottile: So I'm not happy with this either, but the with statement kinda got screwed over in python3.  I commonly refer to this problem as "guido broke the with statement".

There's two alternatives that I also use but they also suck:

The first causes a blowup of indentation:

```python
with blahlbah(
    blahblah,
) as blah blah:
    with blahblahblah(
          blahblahblahblah,
    ):
        # etc.
```

The second uses numerous gross backslashes

```python
with \
    blahblahlbha(
        blahblahblah
    ) as blahblahblah, \
    blahblah(
        blahblahblah,
    ):
```


So I've been meaning to write something up on this subject for some time so I'll take an abbreviated approach here.

paasta has some major problems:
- the tests rely on way too much monkeypatch mocking
- a "fake" filesystem is often created in tests by mocking 5-10 separate functions in `os.path`
- the mocks are duplicated over and over and over and over

The first problem is a much harder one to solve, it involves changing the way problems are approached and the programming style taken.  Paasta has a tendancy to mix both functions and procedures into large twisting functions -- and then there's attempts to test them.  Testing procedures generally isn't a good idea as your test ends up being direct reimplementation of the code it is testing.  The tests are then super fragile and make it very difficult to refactor.  If the code was well factored to begin with, you'd unit test the pure functions and integration test the procedures and there'd be little to no monkeypatch mocks.

The second problem (dealing with filesystems) is perhaps the easiest one to fix (and I've done this in a few places while refactoring).  Patching `os.path` is terribly fragile and often difficult if many calls happen -- it also tightly couples the test implementation to the original function implementation again making it impossible to refactor.  One of the nice things about pytest is it gives you a built in fixture for creating a temporary filesystem that one can manipulate using pypath.  You can then (assuming your function takes in a path as an argument, if it doesn't, it probably should!) test your function by setting up this filesystem and passing it into your function under test.  This will give you a real assertion that your function actually works!

The third problem is repeated monkeypatch fixturing.  This is perhaps the easiest way to solve the with statement blowup that is a major problem in the paasta codebase.  pytest allows you to define your own custom fixtures (and yield fixtures) and reuse and reconfigure them throughout tests.  Here's a toy example:

before:
```python
def test_thing():
    with mock.patch.object(thing.mod, 'side_effecty_function', return_value=(5,)):
        with mock.patch.object(thing.mod, 'other_side_effecty_function', return_value=(5,)):
            assert thing.mod.thing(100) == 200
```

Quickly refactored, it becomes this:

```python
@pytest.yield_fixture
def mock_side_effecty_function():
    with mock.patch.object(thing.mod, 'side_effecty_function') as mck:
        yield mck

@pytest.yield_fixture
def mock_other_side_effecty_function():
    with mock.patch.object(thing.mod, 'other_side_effecty_function') as mck:
        yield mck

# pytest fixtures are magically injected through positional arguments
def test_thing(mock_side_effecty_function, mock_other_side_effecty_function):
    mock_side_effecty_function.return_value = (5,)
    mock_other_side_effecty_function.return_value = (5,)
    assert thing.mod.thing(100) == 200
```

Generally you write fixture per-file, if you need more general fixtures that get shared more generically, you'll collect them in `conftest.py` files that'll be available from the directory they're defined in downwards.

PR #1015, opened 2017-02-08
Make paasta_maintenance status human-friendly

nhandler:
We have received reports that `paasta_maintenance status` is hard to understand. This is because it spits out the raw json (you can pipe it to `jq` to make it pretty). This change makes `paasta_maintenance status` use a more human-friendly output format. The old output is available under a `paasta_maintenance cluster_status` command.

This is internal ticket PAASTA-6990

### DIFF
--- a/paasta_tools/mesos_maintenance.py
+++ b/paasta_tools/mesos_maintenance.py
@@ -634,16 +634,37 @@ def up(hostnames):
     return up_output
 
 
-def status():
+def raw_status():
     """Get the Mesos maintenance status. This contains hostname/ip mappings for hosts that are either marked as being
     down for maintenance or draining.
-    :returns: None
+    :returns: Response Object containing status
     """
     try:
         status = get_maintenance_status()
     except HTTPError:
         raise HTTPError("Error performing maintenance status.")
-    return status.text
+    return status
+
+
+def status():
+    """Get the Mesos maintenance status. This contains hostname/ip mappings for hosts that are either marked as being
+    down for maintenance or draining.
+    :returns: Text representation of the status
+    """
+    return raw_status().text
+
+
+def friendly_status():
+    """Display the Mesos maintenance status in a human-friendly way.
+    :returns: Text representation of the human-friendly status
+    """
+    status = raw_status().json()
+    ret = ""
+    for machine in status.get('draining_machines', []):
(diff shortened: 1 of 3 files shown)


paasta_tools/mesos_maintenance.py
```diff
@@ -634,16 +634,37 @@ def up(hostnames):
...
+def friendly_status():
+    """Display the Mesos maintenance status in a human-friendly way.
+    :returns: Text representation of the human-friendly status
+    """
+    status = raw_status().json()
+    ret = ""
+    for machine in status.get('draining_machines', []):
+        ret += "%s (%s): Draining\n" % (machine['id']['hostname'], machine['id']['ip'])
+    for machine in status.get('down_machines', []):
+        ret += "%s (%s): Down\n" % (machine['hostname'], machine['ip'])
```
Rob-Johnson: Have you tried running this? I'm fairly sure this is formatted incorrectly and you wont get new lines in the correct places. Can you show a pb?

paasta_tools/paasta_maintenance.py
```diff
@@ -221,6 +222,8 @@ def paasta_maintenance():
     elif action == 'up':
         mesos_maintenance.up(hostnames)
     elif action == 'status':
+        ret = "%s" % mesos_maintenance.friendly_status()
```
Rob-Johnson: why do you need to do this interpolation? isn't ``ret = mesos_maintenance.friendly_status()`` enough?

paasta_tools/mesos_maintenance.py
nhandler: I have tested this. Other than an extra newline at the end, it appears to work.

```
$ sudo paasta_maintenance status
mesos2-uswest1cdevc.dev.yelpcorp.com (10.40.1.50): Draining
mesos1-uswest1cdevc.dev.yelpcorp.com (10.40.1.49): Draining
```

```
$ sudo paasta_maintenance status
mesos1-uswest1cdevc.dev.yelpcorp.com (1.2.3.4): Draining
mesos2-uswest1cdevc.dev.yelpcorp.com (1.2.3.5): Down
```

paasta_tools/paasta_maintenance.py
nhandler: I think at one point, we were using it as a sort of type cast. I think we can drop it now that the functions being called are returning strings. Let me fix this.

PR #1018, opened 2017-02-08
Make realert_every be a function of the interval

oktopuz:
internal ticket: PAASTA-7870

* If realert_every is defined in chronos-<cluster>.yaml for the job, that value will be used.
* If the job doesn't have 'realert_every' in chronos-<cluster>.yaml, the value will be guessed based on the job interval.
* If the job doesn't have interval and has no 'realert_every' in chronos-<cluster>.yaml, paasta will read it from monitoring.yaml or will use the default value -1.

I'm not sure whether the guessed value should be more preferable than the one from monitoring.yaml or not.

On the one hand silently overriding values set in monitoring.yaml doesn't look good to me.
On the other hand if we don't override monitoring.yaml, people will have to remove 'realert_after' from monitoring.yaml and put it to marathon-<cluster>.yaml for every instance to be able to use this guessing logic for chronos jobs.

### DIFF
--- a/paasta_tools/check_chronos_jobs.py
+++ b/paasta_tools/check_chronos_jobs.py
@@ -38,6 +38,19 @@ def parse_args():
     return args
 
 
+def guess_realert_every(chronos_job_config):
+    """Returns the job interval (in minutes) or None if the job has no interval.
+
+    As we run check_chronos_jobs once a minute, we are going to skip
+    all re-alert attempts until the next scheduled run of the job.
+
+    This value is clamped by 1 day (1440 minutes) because we want a batch
+    that has a big interval to realert once per day anyway.
+    """
+    interval = chronos_job_config.get_schedule_interval_in_seconds()
+    return min(interval // 60, 24 * 60) if interval is not None else None
+
+
 def compose_monitoring_overrides_for_service(chronos_job_config, soa_dir):
     """ Compose a group of monitoring overrides """
     monitoring_overrides = chronos_job_config.get_monitoring()
@@ -46,8 +59,16 @@ def compose_monitoring_overrides_for_service(chronos_job_config, soa_dir):
     monitoring_overrides['check_every'] = '1m'
     monitoring_overrides['runbook'] = monitoring_tools.get_runbook(
         monitoring_overrides, chronos_job_config.service, soa_dir=soa_dir)
-    monitoring_overrides['realert_every'] = monitoring_tools.get_realert_every(
-        monitoring_overrides, chronos_job_config.service, soa_dir=soa_dir)
+
+    if 'realert_every' not in monitoring_overrides:
+        guessed_realert_every = guess_realert_every(chronos_job_config)
+        if guessed_realert_every is not None:
+            monitoring_overrides['realert_every'] = monitoring_tools.get_realert_every(
+                monitoring_overrides, chronos_job_config.service, soa_dir=soa_dir,
+                monitoring_defaults=lambda x: {'realert_every': guessed_realert_every}.get(x))
+        else:
+            monitoring_overrides['realert_every'] = monitoring_tools.get_realert_every(
+                monitoring_overrides, chronos_job_config.service, soa_dir=soa_dir)
     return monitoring_overrides
 
 
--- a/paasta_tools/check_chronos_jobs.py
+++ b/paasta_tools/check_chronos_jobs.py
@@ -72,6 +72,17 @@ def compose_monitoring_overrides_for_service(chronos_job_config, soa_dir):
     return monitoring_overrides
 
 
+def add_realert_status(sensu_output, realert_every_in_minutes):
+    if realert_every_in_minutes is None or realert_every_in_minutes == -1:
+        return "{}\n\nThis check will not realert.".format(sensu_output)
+    else:
+        hours = realert_every_in_minutes // 60
+        minutes = realert_every_in_minutes % 60
+        interval_string = (hours > 0) * ('%sh' % hours) + (minutes > 0) * ('%sm' % minutes)
+        return ("{}\n\nThis check realerts every {}."
+                .format(sensu_output, interval_string))
+
+
 def send_event(service, instance, monitoring_overrides, soa_dir, status_code, message):
     check_name = compose_check_name_for_service_instance('check_chronos_jobs', service, instance)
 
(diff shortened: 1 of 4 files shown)


paasta_tools/chronos_tools.py
```diff
@@ -257,6 +257,13 @@ def get_disabled(self):
     def get_schedule(self):
         return self.config_dict.get('schedule')
 
+    def get_schedule_interval_in_seconds(self):
+        try:
+            _, _, interval = self.get_schedule().split('/')
```
Rob-Johnson: I'd really like us to be able to do this for crontab too - can you find a library that'll let us calculate the next and last run, and use the delta as interval?

oktopuz: I feel like I'm missing some background information.
What does ISO8601-like schedule have to do with crontab?

Rob-Johnson: sorry, I should've been clearer. This isn't anything to do with the crontab program, but we support specifying schedules for jobs in both ISO8601 notation, and the same format as cron. This means we need to support the functionality you're adding when your schedule is in both formats.

oktopuz: I see. I overlooked the crontab format.

oktopuz: I've found https://pypi.python.org/pypi/croniter

oktopuz:
rebased to master to fix broken tests.

Rob-Johnson:
I would make this the default when no other ``realert_every`` key has been set in ``monitoring.yaml`` or ``chronos-*.yaml``. I don't think its a big deal to ask people to move the definition of the realert_every if they don't want that behavior to apply to chronos jobs.

oktopuz:
I'm looking for a way for override the global default value for chronos jobs only.

paasta_tools/check_chronos_jobs.py
```diff
@@ -34,6 +34,16 @@ def parse_args():
...
 
 
+def _guess_realert_every(chronos_job_config):
+    """Returns the job interval (in minutes) or None if the job has no interval.
+
+    As we run check_chronos_jobs once a minute, we are going to skip
+    all re-alert attempts until the next scheduled run of the job.
+    """
+    interval = chronos_job_config.get_schedule_interval_in_seconds()
+    return (interval // 60) if interval is not None else None
```
solarkennedy: is the `//` intentional?

paasta_tools/check_chronos_jobs.py
```diff
@@ -42,8 +52,14 @@ def compose_monitoring_overrides_for_service(chronos_job_config, soa_dir):
     monitoring_overrides['check_every'] = '1m'
     monitoring_overrides['runbook'] = monitoring_tools.get_runbook(
         monitoring_overrides, chronos_job_config.service, soa_dir=soa_dir)
+
+    if 'realert_every' not in monitoring_overrides:
+        guessed_realert_every = _guess_realert_every(chronos_job_config)
+        if guessed_realert_every is not None:
```
solarkennedy: Ok we may have one last problem with this approach: how will users expect this to interact with "alert_after" ?

I think it will be ok, but could be surprising to people who have long jobs (daily) and an alert_after with the default 2m.

Because on the first failure detection, it will be suppressed because it hasn't been 2m. Which means they will have to wait another 24h before the alert goes off.

I think for chronos jobs the alert_after should default to 0, or something else, I'm not 100% sure on what to do.

paasta_tools/check_chronos_jobs.py
solarkennedy: I think this needs to be clamped for safety. Like max(i, 24*60*60)

paasta_tools/check_chronos_jobs.py
```diff
@@ -42,8 +52,14 @@ def compose_monitoring_overrides_for_service(chronos_job_config, soa_dir):
...
+            monitoring_overrides['realert_every'] = guessed_realert_every
     monitoring_overrides['realert_every'] = monitoring_tools.get_realert_every(
         monitoring_overrides, chronos_job_config.service, soa_dir=soa_dir)
+
     return monitoring_overrides
```
solarkennedy: Can you clarify the alert text to let users know that the fact that they are getting the alert is not an indication that it tried to run, it is just a reminder that the last run status is failing. 

Bonus points if you just say in the alert text how often they should expect the reminder?

solarkennedy:
@oktopuz  can you rebase please?

oktopuz:
@solarkennedy, rebased, but I haven't addressed all issues yet.

paasta_tools/check_chronos_jobs.py
oktopuz: '//' is intentional, because the result of (interval // 60) is int in both python2.7 and python3.
While (interval / 60) is float in python3.

Just to be sure, do you want to clamp it as max(i, 24 * 60 * 60) or min(i, 24 * 60 * 60)?

solarkennedy: Max. The reasoning here is that we don't want a batch that runs once per month to only alert once per month. (too easy to forget)

oktopuz:
I implemented these features today:
* realert_every is a function of the interval when no other realert_every key is set in monitoring.yaml
* the function is clamped by 1 day to keep reminding about the alert
* the alert text contains information about how often it realerts.

I don't see a way to let users know that it is just a reminder that the last run status is failing, because the script doesn't know what happened last time. It just checks the current status.

Also I'm still not sure what to do with 'alert_after'.

paasta_tools/check_chronos_jobs.py
oktopuz: @solarkennedy: I'm not sure I understand why they will have to wait another 24h before the alert goes off?

check_chronos_jobs has check_every = 1m.
The realert_every logic is only applied to the 'create' event (https://github.com/Yelp/sensu_handlers/blob/master/files/base.rb#L241).

It looks like it should work as expected i.e. the alert will goes off as soon as check_chronos_jobs sends OK.

paasta_tools/check_chronos_jobs.py
```diff
@@ -46,11 +59,30 @@ def compose_monitoring_overrides_for_service(chronos_job_config, soa_dir):
...
+                monitoring_overrides, chronos_job_config.service, soa_dir=soa_dir,
+                monitoring_defaults=lambda x: {'realert_every': guessed_realert_every}.get(x))
+        else:
+            monitoring_overrides['realert_every'] = monitoring_tools.get_realert_every(
+                monitoring_overrides, chronos_job_config.service, soa_dir=soa_dir)
     return monitoring_overrides
 
 
+def add_realert_status(sensu_output, realert_every_in_minutes):
+    if realert_every_in_minutes is None or realert_every_in_minutes == -1:
```
solarkennedy: Don't bother with this case. If we can't predict when it will realert, then don't say anything.

solarkennedy:
fix +ship

PR #1019, opened 2017-02-10
Make service autoscaler lenient on running tasks

mattmb:
* Change to threshold of healthy tasks found in marathon rather than requiring match to number of instances we set
* Ensure we *always* scale up to min_instances when we're currently below regardless of what we find running

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -54,6 +54,7 @@ SERVICE_METRICS_PROVIDER_KEY = 'metrics_provider'
 DECISION_POLICY_KEY = 'decision_policy'
 
 AUTOSCALING_DELAY = 300
+MAX_TASK_DELTA = 0.3
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -323,9 +324,26 @@ def get_error_from_utilization(utilization, setpoint, current_instances):
 
 def autoscale_marathon_instance(marathon_service_config, marathon_tasks, mesos_tasks):
     current_instances = marathon_service_config.get_instances()
-    if len(marathon_tasks) != current_instances:
-        write_to_log(config=marathon_service_config,
-                     line='Delaying scaling as marathon is either waiting for resources or is delayed')
+    too_many_instances_running = len(marathon_tasks) > int((1 + MAX_TASK_DELTA) * current_instances)
+    too_few_instances_running = len(marathon_tasks) < int((1 - MAX_TASK_DELTA) * current_instances)
+    if too_many_instances_running or too_few_instances_running:
```
Rob-Johnson: doesn't this mean we'll only scale up to the ``min_instances`` if the delta between ``min_instances`` and current_instances is > 30%?

mattmb: This is actually catching a case that was missing before. We don't scale at all if we find more than 1.3* expected number of tasks or less than 0.7* expected. *But* we do want to scale to at least `min_instances` in this case (in case it was changed by a human).

The later call to `limit_instance_count` makes sure we do `min_instances` in the more general case when everything is within the new threshold.

PR #1021, opened 2017-02-13
Handle create app lock failure gracefully

mattmb:
We get these exceptions pretty frequently in our large clusters when we
are creating multiple apps at the same time. Currently the exception is
raised all the way up which causes setup_marathon_job to exit 1 and for
our deploy loop to start over. This catches the exception and continues
on. I think it should be safe because we will simply try again next time
to create the new marathon app.

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -525,6 +525,10 @@ def deploy_service(
             soa_dir=soa_dir,
             bounce_margin_factor=bounce_margin_factor,
         )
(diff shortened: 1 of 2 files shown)


mattmb:
I'm going to merge this. But as an open question (cc @nhandler @EvanKrall), do we know if we still really need this lock with marathon 1.4? Maybe it's got better at handling concurrent requests? I guess once we've got 1.4 rolled everywhere we could try disabling it and see if any weirdness ensues...

PR #1024, opened 2017-02-14
Ensure the cluster scaler jumps to min_capacity

### LINKED ISSUE #979: Go straight to min_capacity
https://github.com/Yelp/paasta/blob/master/paasta_tools/autoscaling/autoscaling_cluster_lib.py#L441

We should go straight to min_capacity as this is what an operator specifies, currently we limit to a max delta.

mattmb:
Fixes #979

We currently constrain to a `MAX_CLUSTER_DELTA` when scaling up or down.
This prevents sudden large changes when the autoscaler has made a
mistake. However, an operator increasing `min_capacity` on a resource
should trump this situation. This commit makes it so that we never set a
new capacity that is lower than the current `min_capacity`.

### DIFF
--- a/tests/autoscaling/test_autoscaling_cluster_lib.py
+++ b/tests/autoscaling/test_autoscaling_cluster_lib.py
@@ -244,6 +244,15 @@ class TestAsgAutoscaler(unittest.TestCase):
         ret = self.autoscaler.get_asg_delta(-0.5)
         assert ret == (20, int(floor(20 * (1.0 - autoscaling_cluster_lib.MAX_CLUSTER_DELTA))))
 
+        current_instances = int((10 * (1 - autoscaling_cluster_lib.MAX_CLUSTER_DELTA)) - 1)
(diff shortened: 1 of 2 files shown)


tests/autoscaling/test_autoscaling_cluster_lib.py
```diff
@@ -244,6 +244,14 @@ def test_get_asg_delta(self):
         ret = self.autoscaler.get_asg_delta(-0.5)
         assert ret == (20, int(floor(20 * (1.0 - autoscaling_cluster_lib.MAX_CLUSTER_DELTA))))
 
+        current_instances = int((10 * (1 - autoscaling_cluster_lib.MAX_CLUSTER_DELTA)) - 1)
+        self.autoscaler.resource = {
+            'min_capacity': 10,
+            'max_capacity': 40}
```
Rob-Johnson: put this brace on the next line.

PR #1026, opened 2017-02-16
Add meaningful docker hostnames when run with mesos-slave

bchess:
This introduces a wrapper intended to be invoked by mesos-slave in place of the real true /usr/bin/docker binary. It parses the CLI args intended for docker, extracting the MARATHON_HOST and MESOS_TASK_ID values. If available, it adds a '--hostname=' CLI argument, where the hostname is the first part of $MARATHON_HOST and lastpart of $MESOS_TASK_ID (dot-delimitted).

The intended use case is to set a meaningful hostname for our docker containers that can be used when debugging log files to track back which host and task the container was running on.

If MARATHON_HOST or MESOS_TASK_ID is unspecified, or if --hostname is already set, or if you're not doing a 'docker run', this wrapper will exec docker without any modification to the arguments.

I also upgraded pytest so I could use 'yield' inside fixtures.

I was pretty paranoid about my testing of argument parsing, though in the real world the arguments mesos-slave uses to run are pretty straightforward "-e KEY=VALUE" style.

Internal ticket PAASTA-7240


bchess:
i need to fix a couple pep8 things, but also there are some weird itest failures i think due to the pytest upgrade. Lemme investigate and I'll update this again later

asottile:
For what it's worth,  the behave plugin is incompatible with pytest 3. You can still use yield fixtures in pytest 2 by using `@pytest.yield_fixture`

bchess:
Ah, thanks @asottile. I'll use yield_fixture instead

paasta_tools/docker_hostname_wrapper.py
```diff
@@ -0,0 +1,87 @@
...
+        run_index = args.index('run')
+    except ValueError:
+        return
+
+    args.insert(run_index + 1, '--hostname=' + hostname)
+
+
+def main():
+    env_args = parse_env_args(sys.argv)
+    marathon_host = env_args.get('MARATHON_HOST')
```
solarkennedy: Why does this need to be marathon-specific at all? Why not just always feed the *true* hostname? (socket.getfqdn) is your intent to extract as much metadata as you can?

On chronos (or other framework) tasks, this will be None, we *could* inject a reduced from.

Then again why not just be more universal and concat the real hostname and the task id? (things we will always have)

bchess: Sure, I'm fine with using socket.getfqdn() instead.

tests/test_docker_hostname_wrapper.py
```diff
@@ -0,0 +1,186 @@
...
+
+    @pytest.mark.parametrize('args', [
+        ['docker', '--hostname', 'foo'],
+        ['docker', '--hostname=foo'],
+    ])
+    def test_long(self, args):
+        assert docker_hostname_wrapper.already_has_hostname(args) is True
+
+
+class TestGenerateHostname(object):
```
solarkennedy: Can you add some more tests with realistic mesos ids? I'm concerned about parsing this "user" input.

Here are two different classes:
```
ct:1487804100000:0:thirdparty_feeds thirdparty_feeds-cloudflare-all:
paasta--canary.main.git332d4a22.config458863b1.0126a188-f944-11e6-bdfb-12abac3adf8c
```

rpartition those! :)

bchess:
added an extra partition on spaces for chronos, plus tests for the examples you provided

bchess:
Whoops, I need to strip special characters like ':'

bchess:
Now stripping non-compliant characters from hostname.

I also dropped the max length from 64 to 63 because there seems to be some disagreement about it. Linux (and docker) seems to allow 64 but https://tools.ietf.org/html/rfc1034#section-3.1 says only 63 are allowed for domain names. Going with 63 to be safe.

paasta_tools/docker_hostname_wrapper.py
```diff
@@ -0,0 +1,92 @@
...
+                continue
+
+        in_env = False
+        if '=' not in arg:
+            continue
+
+        k, _, v = arg.partition('=')
+        result[k] = v
+
+    return result
```
asottile: One idea: instead of rolling your own argument parsing,  argparse + parse_known_args might be a good idea?

paasta_tools/docker_hostname_wrapper.py
```diff
@@ -0,0 +1,92 @@
...
+
+    hostname = host_hostname + '-' + task_id
+
+    # hostnames can only contain alphanumerics and dashes and must be no more
+    # than 63 characters
+    hostname = re.sub('[^a-zA-Z0-9-]+', '-', hostname)[:MAX_HOSTNAME_LENGTH]
+    return hostname
+
+
+def add_hostname(args, hostname):
```
asottile: Usually prefer apis which return new values instead of mutating arguments.  They're much easier to think about,  compose, and test

paasta_tools/docker_hostname_wrapper.py
```diff
@@ -0,0 +1,92 @@
...
+    try:
+        run_index = args.index('run')
+    except ValueError:
+        return
+
+    args.insert(run_index + 1, '--hostname=' + hostname)
+
+
+def main():
```
asottile: One pattern which would reduce the monkeypatching in tests would be:

```
def main(argv=None):
    args = argv if argv is not None else sys.argv[1:]
```

If you switch to argparse this gets even easier as argparse will look at sys.argv for you if you pass in None to parse_args


Sorry for the separate comments,  I'm on my phone and was bored and decided to look through PRs :D

paasta_tools/docker_hostname_wrapper.py
bchess: It appears that argparse parse_known_args() can't unpack combined short args :(
https://gist.github.com/bchess/326da50168729e82f5d17d913bef5cc3

paasta_tools/docker_hostname_wrapper.py
```diff
@@ -0,0 +1,99 @@
...
+
+    return args
+
+
+def main(argv=None):
+    argv = argv if argv is not None else sys.argv[1:]
+
+    env_args = parse_env_args(argv)
+    fqdn = socket.getfqdn()
+    mesos_task_id = env_args.get('MESOS_TASK_ID')
```
solarkennedy: I'm sorry to say this, but this is a marathon specific variable (I know, silly). AFAIK there is no mesos-provided var for this that is universal.

In chronos is is the same thing but *lowercase* `mesos_task_id` :(
https://github.com/yelp/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosTaskBuilder.scala#L57

It looks like upstream head they add both now. We can add that into our fork, but for now can you look for either?

bchess: done :dancer:

setup.py
```diff
@@ -109,6 +109,7 @@
             'paasta_setup_chronos_job=paasta_tools.setup_chronos_job:main',
             'paasta_chronos_rerun=paasta_tools.chronos_rerun:main',
             'paasta_cleanup_maintenance=paasta_tools.cleanup_maintenance:main',
+            'docker_hostname_wrapper=paasta_tools.docker_hostname_wrapper:main',
```
solarkennedy: Sorry for so much review, but long term we would like to use this wrapper for more things other than hostname.
also I prefer that new script have the prefix `paasta_` in them so that they can't possibly be mistaken for other tools on the filesystem.

Can you call it just `paasta_docker_wrapper` ?

solarkennedy:
Can you address the pre-commit pep8 stuff? (and be sure to use precommit if possible)
https://travis-ci.org/Yelp/paasta/jobs/204737041

bchess:
whoop, sorry about that fixed

PR #1027, opened 2017-02-17
detect stuck chronos jobs

oktopuz:
internal ticket PAASTA-8167.

### DIFF
--- a/paasta_tools/check_chronos_jobs.py
+++ b/paasta_tools/check_chronos_jobs.py
@@ -87,9 +87,9 @@ def build_service_job_mapping(client, configured_jobs):
     :param client: A Chronos client used for getting the list of running jobs
     :param configured_jobs: A list of jobs configured in Paasta, i.e. jobs we
         expect to be able to find
-    :returns: A dict of {(service, instance): [(chronos job, lastrunstate)]}
-        where the chronos job is any with a matching (service, instance) in its
-        name and disabled == False
+    :returns: A dict of {(service, instance): last_chronos_job}
+        where last_chronos_job is the latest job matching (service, instance)
+        or None if there is no such job
     """
     service_job_mapping = {}
     for job in configured_jobs:
@@ -101,11 +101,8 @@ def build_service_job_mapping(client, configured_jobs):
             include_disabled=True,
         )
         matching_jobs = chronos_tools.sort_jobs(matching_jobs)
-
-        if len(matching_jobs) > 0:
-            # Only consider the most recent one
-            status = chronos_tools.get_status_last_run(matching_jobs[0])
-            service_job_mapping[job] = [(matching_jobs[0], status[1])]
+        # Only consider the most recent one
+        service_job_mapping[job] = matching_jobs[0] if len(matching_jobs) > 0 else None
     return service_job_mapping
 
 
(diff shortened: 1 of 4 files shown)


paasta_tools/check_chronos_jobs.py
```diff
@@ -152,31 +144,65 @@ def message_for_status(status, service, instance, cluster):
...
+
+
+utc = TZ()
+
+
+def job_is_stuck(last_run_iso_time, interval_in_seconds):
+    if last_run_iso_time is None or interval_in_seconds is None:
+        return False
+    last_run_datatime = isodate.parse_datetime(last_run_iso_time)
+    return last_run_datatime + timedelta(seconds=interval_in_seconds) < datetime.now(utc)
```
Rob-Johnson: can you just use ``datetime.datetime.utcnow()``?

paasta_tools/check_chronos_jobs.py
```diff
@@ -152,31 +144,65 @@ def message_for_status(status, service, instance, cluster):
...
+
+
+def message_for_stuck_job(service, instance, cluster, last_run_iso_time, interval_in_seconds):
+    return ("Job %(service)s%(separator)s%(instance)s hasn't run since %(last_run)s,"
+            " and is configured to run every %(interval).1f minutes.\n\n"
```
Rob-Johnson: There isn't really any action here for the user - maybe suggest they speak to someone on the PaaSTA team?

paasta_tools/check_chronos_jobs.py
oktopuz: If I could, that wouldn't be python2.7.

utcnow() returns UTC date and time with with tzinfo=None and can't be compared with datetime objects that have a valid tzinfo.

```
Python 2.7.6 (default, Jun 28 2016, 17:53:28)
[GCC 4.6.3] on linux2
>>> import isodate
>>> from_iso = isodate.parse_datetime('2016-07-26T22:03:00+00:00')
>>> from_iso
datetime.datetime(2016, 7, 26, 22, 3, tzinfo=<FixedOffset '+00:00'>)
>>>
>>> from datetime import datetime
>>> time_now = datetime.utcnow()
>>> time_now
datetime.datetime(2017, 2, 21, 17, 42, 58, 580216)
>>>
>>> time_now > from_iso
  File "<stdin>", line 1, in <module>
TypeError: can't compare offset-naive and offset-aware datetimes
>>>
```

paasta_tools/check_chronos_jobs.py
oktopuz: 'the PaaSTA team' sounds like a yelpism.
I haven't found any references to it anywhere in the code.

I'm not sure what actions to recommend in that case except checking logs.
```
+            "You can view the logs for the job with:\n"
+            "\n"
+            "    paasta logs -s %(service)s -i %(instance)s -c %(cluster)s\n"
+            "\n
```

paasta_tools/check_chronos_jobs.py
```diff
@@ -152,31 +144,65 @@ def message_for_status(status, service, instance, cluster):
...
+class TZ(tzinfo):
+
+    def utcoffset(self, dt):
+        return timedelta(minutes=0)
+
+    def dst(self, dt):
+        return timedelta(minutes=0)
+
+
+utc = TZ()
```
solarkennedy: Having this instance here outside of any method looks funny to me.

Reading the later comment... yuck :(

But I understand now. Time zones man :)

paasta_tools/check_chronos_jobs.py
```diff
@@ -152,31 +144,65 @@ def message_for_status(status, service, instance, cluster):
...
+
+
+def job_is_stuck(last_run_iso_time, interval_in_seconds):
+    if last_run_iso_time is None or interval_in_seconds is None:
```
solarkennedy: This is dangerous because we can easily "lose" the last_run_time.

I think in theory we will get this early because of the "fresh" , but I don't see that?
We used to *not* send an event if the job was "fresh" and hence we had no data.

oktopuz: If the last_run_time is None, job_is_stuck() returns False.
That will cause the check to fallback to the old logic (line 203), i.e. to run sensu_event_for_last_run_state() as it was before.

solarkennedy:
Let's do it. Thank you for eliminating the "pairs" concept. In theory there should only ever be "1" job in chronos with a unique name, hopefully that is actually the case in practice.

PR #1029, opened 2017-02-17
Catch 409 conflict when bouncing marathon apps

mattmb:
In 0c950d0e43434222b598d8b8fc7da8c1383a666e we moved the lock in
`setup_marathon_job` to prevent race conditions between multiple bounce
loops which run at the same time. One motivation was seeing HTTP 409
conflicts from marathon when the bounce attempts to create the same app
twice. It turns out that this is still possible since we
`get_all_marathon_apps` outside the lock which can lead to two bounces
deciding to create the same app, resulting in HTTP 409 from marathon.

We could move this call into the lock but this would incurr a
performance hit as it would be called once per service instance. Instead
I am catching the error which I am pretty sure is safe. Essentially a
second bounce may decide it needs to create an app that already exists
but this is now a silent failure and the bounce will stop there.

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -51,6 +51,7 @@ from collections import defaultdict
 
 import pysensu_yelp
 import requests_cache
+from marathon.exceptions import MarathonHttpError
 from requests.exceptions import HTTPError
 
 from paasta_tools import bounce_lib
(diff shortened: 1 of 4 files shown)


paasta_tools/setup_marathon_job.py
```diff
@@ -234,7 +235,13 @@ def log_bounce_action(line, level='debug'):
             line='%s bounce creating new app with app_id %s' % (bounce_method, marathon_jobid),
         )
         with requests_cache.disabled():
-            bounce_lib.create_marathon_app(marathon_jobid, config, client)
+            try:
+                bounce_lib.create_marathon_app(marathon_jobid, config, client)
+            except MarathonHttpError as e:
+                if e.status_code == 409:
```
Rob-Johnson: I think you want an ``else`` here to reraise when its not 409.

mattmb: Whoops, you're right. Nice catch 👍

mattmb:
@Rob-Johnson ebb589f should fix the problem you spotted and 30689e7 is just to fix the tests

PR #1031, opened 2017-02-18
All the unit tests pass in python 3!

asottile:
Most of this was just straightforward contextlib switching.

There's one concering change in tests/autoscaling/test_autoscaling_service_lib.py (from  paasta_tools/autoscaling/autoscaling_service_lib.py).  The math here (copied from wikipedia) was incorrectly doing integer division (the default of `/` in python2) -- I switched it to floating point division (py3 default) via `from __future__ import division`.

I split the tedious nested changes into the first commit and the interesting bits into the second commit.

The version test can't pass until we fix version conflicts (I've submitted a patch to mesos to allow protobuf>=3 in mesos.interface to resolve the only remaining conflict).

### DIFF
--- a/tests/test_mesos_tools.py
+++ b/tests/test_mesos_tools.py
@@ -605,13 +605,13 @@ def test_get_mesos_task_count_by_slave():
         assert mock_get_all_running_tasks.called
         expected = [{'task_counts': mesos_tools.SlaveTaskCount(count=2, chronos_count=1, slave=mock_slave_1)},
                     {'task_counts': mesos_tools.SlaveTaskCount(count=2, chronos_count=0, slave=mock_slave_2)}]
-        assert len(ret) == len(expected) and sorted(ret) == sorted(expected)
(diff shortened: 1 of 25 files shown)


tests/test_mesos_tools.py
```diff
@@ -685,20 +668,18 @@ def test_get_tasks_from_app_id():
         ret = mesos_tools.get_tasks_from_app_id('app_id')
         mock_get_running_tasks_from_frameworks.assert_called_with('app_id')
         expected = [mock_task_1, mock_task_2, mock_task_3]
-        assert len(expected) == len(ret) and sorted(ret) == sorted(expected)
+        assert len(expected) == len(ret) and _ids(ret) == _ids(expected)
```
mattmb: What's the reason for this? Is the behaviour of `sorted` changed or have I just been doing this wrong?

asottile: mocks are not sortable

PR #1036, opened 2017-02-21
Revert "Use marathon 1.4.0 rc3 in itests"

mattmb:
This reverts commit 4e0efaf073835057eb6d180b5569c268d93eae54.

Reverting until we're actually using this anywhere. Found at least one
bug in marathon-python when using the new version.


mattmb:
As discussed I'm going to close this for now and create a failing test that demonstrates the brokenness (which is to do with marathon-python and the version of marathon)

In general though it would be cool to run the itests with different versions of marathon... I guess that might be tricky given the various install methods...

PR #1037, opened 2017-02-21
Send git refs as byte strings

mattmb:
After we changed to decoding git refs to unicode strings we also need to
make sure to encode back to bytes before we send to dulwich. Otherwise
we see: refname is not a bytestring: u'refs/tags...'

I've also converted the shas back to bytes although I'm not sure we
really needed to...

I'm going to try and add some itests that will catch bugs like this in
the future. But I'll do that seperately. cc @asottile

### DIFF
--- a/paasta_tools/remote_git.py
+++ b/paasta_tools/remote_git.py
@@ -81,5 +81,6 @@ def make_force_push_mutate_refs_func(targets, sha):
     def mutate_refs(refs):
         for target in targets:
             refs[target] = sha
+        refs = {k.encode('UTF-8'): v.encode('UTF-8') for k, v in refs.items()}
         return refs
     return mutate_refs
--- a/tests/test_remote_git.py
+++ b/tests/test_remote_git.py
@@ -65,7 +65,6 @@ def test_non_ascii_tags():
 
 def test_make_force_push_mutate_refs_func_overwrites_shas():
     targets = ['refs/heads/targeta', 'refs/tags/targetb']
-    newsha = 'newsha'
     input_refs = {
         'refs/heads/foo': '12345',
         'refs/heads/targeta': '12345',
@@ -74,19 +73,20 @@ def test_make_force_push_mutate_refs_func_overwrites_shas():
         'refs/tags/blah': '12345',
     }
     expected = {
-        'refs/heads/foo': '12345',
-        'refs/heads/targeta': newsha,
-        'refs/tags/targetb': newsha,
-        'refs/heads/ignored': '12345',
-        'refs/tags/blah': '12345',
+        b'refs/heads/foo': b'12345',
+        b'refs/heads/targeta': b'newsha',
+        b'refs/tags/targetb': b'newsha',
+        b'refs/heads/ignored': b'12345',
+        b'refs/tags/blah': b'12345',
     }
 
     mutate_refs_func = remote_git.make_force_push_mutate_refs_func(
         targets=targets,
-        sha=newsha,
+        sha='newsha',
     )
     actual = mutate_refs_func(input_refs)
     assert actual == expected
+    assert all([isinstance(k, bytes) for k in actual])
 
 
 @mock.patch('dulwich.client', autospec=True)


tests/test_remote_git.py
```diff
@@ -87,6 +87,7 @@ def test_make_force_push_mutate_refs_func_overwrites_shas():
     )
     actual = mutate_refs_func(input_refs)
     assert actual == expected
+    assert all([isinstance(k, bytes) for k in actual.keys()])
```
asottile: p3: iterating over a dictionary gives keys (`.keys()` isn't terribly useful)

paasta_tools/remote_git.py
```diff
@@ -81,5 +81,6 @@ def make_force_push_mutate_refs_func(targets, sha):
     def mutate_refs(refs):
         for target in targets:
             refs[target] = sha
+            refs = {k.encode('UTF-8'): v.encode('UTF-8') for k, v in refs.items()}
```
asottile: shouldn't this instead go in `_make_determine_wants_func`?

mattmb: If I'm reading it right `make_force_push_mutate_refs_func` is called by `_make_determine_wants_func` but in `create_remote_refs` if `force=True` we only end up calling `make_force_push_mutate_refs_func` so I think it has to go there?

tests/test_remote_git.py
mattmb: 👍

paasta_tools/remote_git.py
asottile: What a weird api :S

asottile:
:boat:

EvanKrall:
Seems to me like we shouldn't be converting from bytes to unicode in the first place; [git doesn't specify the encoding of refs](https://git-scm.com/docs/git-rev-parse):

> Note that any of the refs/* cases above may come either from the $GIT_DIR/refs directory or from the $GIT_DIR/packed-refs file. While the ref name encoding is unspecified, UTF-8 is preferred as some output processing may assume ref names in UTF-8.

asottile:
@EvanKrall paasta does significant string manipulation on git tags and makes assumptions that they are ASCII.  Performing these string operations on byte strings is nearly impossible in a correct manner (especially true in python 3).

The general approach that an application should take is to deal with text internally and convert from bytes at application boundaries.

tests/test_remote_git.py
```diff
@@ -65,7 +65,7 @@ def test_non_ascii_tags():
 
 def test_make_force_push_mutate_refs_func_overwrites_shas():
     targets = ['refs/heads/targeta', 'refs/tags/targetb']
-    newsha = 'newsha'
+    newsha = b'newsha'
```
asottile: this constant isn't terribly useful now (it's probably more readable to not have the constant)

mattmb: 👍

asottile:
:boat:

asottile:
oh actually, I have a diff to this (trying this branch out locally):

```diff
diff --git a/paasta_tools/remote_git.py b/paasta_tools/remote_git.py
index b611b03..ee20a00 100644
--- a/paasta_tools/remote_git.py
+++ b/paasta_tools/remote_git.py
@@ -80,7 +80,6 @@ def make_force_push_mutate_refs_func(targets, sha):
     :returns: A function to do the ref manipulation that a dulwich client can use"""
     def mutate_refs(refs):
         for target in targets:
-            refs[target] = sha
-        refs = {k.encode('UTF-8'): v.encode('UTF-8') for k, v in refs.items()}
+            refs[target.encode('UTF-8')] = sha.encode('UTF-8')
         return refs
     return mutate_refs
```

asottile:
Actually even that is insufficient, let me try a blanket patch based on my original suggestion

asottile:
This patch (on top of your branch) ensures that the correct types are used for both normal push and for force push.  I also moved the two functions to be near each other for readability.

```diff
diff --git a/paasta_tools/remote_git.py b/paasta_tools/remote_git.py
index b611b03..ae9969d 100644
--- a/paasta_tools/remote_git.py
+++ b/paasta_tools/remote_git.py
@@ -17,19 +17,37 @@ from __future__ import unicode_literals
 import dulwich.client
 import dulwich.errors
 
+from paasta_tools.utils import to_bytes
+
 
 def _make_determine_wants_func(ref_mutator):
     """Returns a safer version of ref_mutator, suitable for passing as the
     determine_wants argument to dulwich's send_pack method. The returned
     function will not delete or modify any existing refs."""
     def determine_wants(old_refs):
-        refs = dict(old_refs)
+        refs = {k.decode('UTF-8'): v.decode('UTF-8') for k, v in old_refs.items()}
         new_refs = ref_mutator(refs)
+        new_refs = {k.encode('UTF-8'): v.encode('UTF-8') for k, v in new_refs.items()}
         new_refs.update(old_refs)  # Make sure we don't delete/modify anything.
         return new_refs
     return determine_wants
 
 
+def make_force_push_mutate_refs_func(targets, sha):
+    """Create a 'force push' function that will inform send_pack that we want
+    to mark a certain list of target branches/tags to point to a particular
+    git_sha.
+
+    :param targets: List of branches/tags to point at the input sha
+    :param sha: The git sha to point the branches/tags at
+    :returns: A function to do the ref manipulation that a dulwich client can use"""
+    def mutate_refs(refs):
+        for target in targets:
+            refs[target.encode('UTF-8')] = sha.encode('UTF-8')
+        return refs
+    return mutate_refs
+
+
 def create_remote_refs(git_url, ref_mutator, force=False):
     """Creates refs (tags, branches) on a remote git repo.
 
@@ -68,19 +86,3 @@ def list_remote_refs(git_url):
         return {k.decode('UTF-8'): v.decode('UTF-8') for k, v in refs.items()}
     except dulwich.errors.HangupException as e:
         raise LSRemoteException("Unable to fetch remote refs: %s" % e)
-
-
-def make_force_push_mutate_refs_func(targets, sha):
-    """Create a 'force push' function that will inform send_pack that we want
-    to mark a certain list of target branches/tags to point to a particular
-    git_sha.
-
-    :param targets: List of branches/tags to point at the input sha
-    :param sha: The git sha to point the branches/tags at
-    :returns: A function to do the ref manipulation that a dulwich client can use"""
-    def mutate_refs(refs):
-        for target in targets:
-            refs[target] = sha
-        refs = {k.encode('UTF-8'): v.encode('UTF-8') for k, v in refs.items()}
-        return refs
-    return mutate_refs
```

mattmb:
Hmm I can't really see why we need to `decode` and then `encode` again in `_make_determine_wants_func` but it maybe I'm just missing something...

I think I'm going to merge this as is to unblock cutting a new release. Then we can follow up with a better fix in #1038 if you want.

asottile:
The reason for that is the non-force-push case is also broken

mattmb:
Oh man, paasta stop is also knackered:
https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/start_stop_restart.py#L120

It uses its own "ref_mutator" which we haven't patched...

mattmb:
Also, are you sure this would work @asottile:
`refs[target.encode('UTF-8')] = sha.encode('UTF-8')`

Will it actually update the key?

asottile:
Yes, see my branch where I've turned on the itests for this in python3 (so types are actually properly checked for these calls).

PR #1038, opened 2017-02-22
general_itests pass in py3

asottile:
@mattmb I took the git refs bytes change in this :) (#1037)

### DIFF
--- a/paasta_tools/remote_git.py
+++ b/paasta_tools/remote_git.py
@@ -23,11 +23,27 @@ def _make_determine_wants_func(ref_mutator):
     determine_wants argument to dulwich's send_pack method. The returned
     function will not delete or modify any existing refs."""
     def determine_wants(old_refs):
-        refs = dict(old_refs)
+        refs = {k.decode('UTF-8'): v.decode('UTF-8') for k, v in old_refs.items()}
         new_refs = ref_mutator(refs)
+        new_refs = {k.encode('UTF-8'): v.encode('UTF-8') for k, v in new_refs.items()}
         new_refs.update(old_refs)  # Make sure we don't delete/modify anything.
         return new_refs
     return determine_wants
+
+
+def make_force_push_mutate_refs_func(targets, sha):
+    """Create a 'force push' function that will inform send_pack that we want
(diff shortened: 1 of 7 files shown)


asottile:
Manual tests:
- I tested `paasta fsm` in both python2 and python3 (for python3 I had to manually upgrade protobuf as tox also does and modify the .dist-info of mesos.interface so I could use the cli).

paasta_tools/remote_git.py
```diff
@@ -23,13 +23,29 @@ def _make_determine_wants_func(ref_mutator):
...
+    :param sha: The git sha to point the branches/tags at
+    :returns: A function to do the ref manipulation that a dulwich client can use"""
+    def mutate_refs(refs):
+        for target in targets:
+            refs[target.encode('UTF-8')] = sha.encode('UTF-8')
+        return refs
+    return mutate_refs
+
+
 def create_remote_refs(git_url, ref_mutator, force=False):
```
mattmb: So I think the real problem is that anyone calling this can pass their own `ref_mutator`. And it turns out this is exactly what `paasta stop` does.

Could we wrap another function to do the encoding? i.e. like `_make_determine_wants_func` but for every call to `create_remote_refs` not just when `force is False`. That way we could just have the encoding in one place and it would catch the `paasta stop` usage too.

Also, if it's easy and you have time can you add a similar test for `paasta stop`. I haven't looked in detail but it's definitely broken on master so we shouldn't have a green build. I started work on some higher level itests that may or may not be a good idea (#1039) but it would be nice to have an itest like the one you added here for mark-for-deployment.

asottile: So when force is True, we're only calling it with the function above (which does handle encoding properly).  Should really just `assert is make_force_push_mutate_refs_func`?

There is a test which covers the stop case, but it only errored in python3 (because python2 allows **implicit** comparison between text and bytes) (which is why I enabled the general_itests for python3).

mattmb: Ah right, what I'd missed is that `paasta stop` has `force=False` so your change in `_make_determine_wants_func` does fix it. Thanks for fixing it! 🐑 away

PR #1039, opened 2017-02-22
Add example cluster test suite

mattmb:
Our itests are slightly limited. This aims to test a few cases that may
be missed by the itests. It is essentially a set of bash tests for
paasta that run basic paasta commands and expect a 0 error code.

For the first iteration we are:
* deploying a new service by running the same commands we run in Jenkins
pipelines
* running metastatus and status commands
* checking that the paasta api runs
* checking that the api shows that the hello world service is "Running"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,3 +44,7 @@ deploy:
       tags: true
       condition: $MAKE_TARGET == itest_trusty || $MAKE_TARGET == itest_xenial
       repo: Yelp/paasta
+matrix:
+  allow_failures:
+    - env: TOXENV=example_cluster
+  fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
   - TOXENV=general_itests
   - TOXENV=general_itests_py3
   - TOXENV=paasta_itests
+  - TOXENV=example_cluster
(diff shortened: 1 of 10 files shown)


Rob-Johnson:
isn't the right thing to do here to do the leg work to add the necessary parts to the itest infrastructure and encode your bash as behave tests?

mattmb:
I sort of agree with you @Rob-Johnson. But then again I think it'd be quite nice to have some tests that are executed the same way we actually run paasta. Also, this is kind of easier? It would definitely have caught the bad release I just pushed for example. i.e. paasta stop is broken but our itests pass.

Anyway, it's just an idea, happy to be downvoted. Also it doesn't seem to work on travis yet...

mattmb:
Crikey, this is actually working now

solarkennedy:
Can you add it as an `allowed_failures` entry and add `fast_finish` 
https://docs.travis-ci.com/user/customizing-the-build

and we can see how it goes without blocking stuff?

mattmb:
That's a good shout @solarkennedy I'll send a fixup to make that change 👍 

In general: I don't want this to block releases/merges if we've just broken some part of the build process but I do want to spot runtime errors we've introduced by changing the codebase itself.

PR #1040, opened 2017-02-22
paasta remote-run

keymone:
cli tool to allow spawning adhoc tasks on mesos cluster


paasta_tools/cli/cmds/remote_run.py
```diff
@@ -0,0 +1,314 @@
...
+    list_parser.add_argument(
+        '-d', '--dry-run',
+        help='Don\'t launch the task',
+        action='store_true',
+        required=False,
+        default=False,
+    )
+    list_parser.add_argument(
+        '-l', '--local',
+        help='Run mesos-related bits locally, rather than ssh\'ing to mesos-master',
```
Rob-Johnson: I think you'll eventually *have* to ssh to a mesos master to get the right credentials, but I can see why you're doing this to iterate.

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -0,0 +1,314 @@
...
+        reader_func=make_config_reader(instance_type)
+    )
+    scheduler = PaastaAdhocScheduler(
+        command=command,
+        service_config=service_config,
+        system_paasta_config=system_paasta_config,
+        dry_run=dry_run
+    )
+    driver = create_driver_with(
+        # TODO: randomize framework name to get one per task?
```
Rob-Johnson: yes please - add a timestamp or something to it to make it unique.

paasta_tools/cli/utils.py
```diff
@@ -554,16 +564,50 @@ def execute_chronos_rerun_on_remote_master(service, instancename, cluster, syste
...
-    return run_chronos_rerun(master, service, instancename, **kwargs)
+    try:
+        return run_chronos_rerun(
+            connectable_master(cluster, system_paasta_config),
+            service, instancename, **kwargs)
+    except NoMasterError as e:
+        return (-1, str(e))
+
+
+def run_on_master(cluster, system_paasta_config, cmd_parts,
```
Rob-Johnson: this is cool.

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -0,0 +1,314 @@
...
+            file=sys.stderr,
+        )
+        return 1
+
+    soa_dir = args.yelpsoa_config_root
+    dry_run = args.dry_run
+    instance = args.instance
+    command = shlex.split(args.cmd, posix=False) if args.cmd else None
+
+    if instance is None:
```
Rob-Johnson: I take it this is just for iterating on?

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -0,0 +1,314 @@
...
+        elif update.state == mesos_pb2.TASK_LOST or update.state == mesos_pb2.TASK_KILLED:
+            self.launched = False
+
+        if self.status != None:
+            paasta_print("Task %s is finished.", update.task_id.value)
+            for stream in mesos.cli.cluster.files(flist=['stdout','stderr'], fltr=update.task_id.value):
+                print "Printing %s for task %s" % (stream[0].path, update.task_id.value)
+                for line in stream[0].readlines():
+                    print line
+            driver.stop() # is this the right place to stop the driver?
```
Rob-Johnson: do you need to exit, too?

paasta_tools/cli/cmds/remote_run.py
keymone: seemed like sensible defaults. `adhoc-%s.yaml` has `interactive` section for local-run, let's have `remote` for `remote-run`?

paasta_tools/cli/cmds/remote_run.py
keymone: the ssh bit is already done: https://github.com/Yelp/paasta/pull/1040/files#diff-d4b928a56fb9cdd97ba9e24b6f836f76R307

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -0,0 +1,314 @@
...
+from paasta_tools.utils import SystemPaastaConfig
+from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import validate_service_instance
+
+from paasta_tools.native_mesos_scheduler import create_driver_with
+from paasta_tools.native_mesos_scheduler import compose_job_id
+from paasta_tools.native_mesos_scheduler import load_paasta_native_job_config
+
+
+class PaastaAdhocScheduler(Scheduler):
```
EvanKrall: I wonder whether it would be better to have this inherit from PaastaNativeScheduler. I think you'd basically just have to override `need_more_tasks` to return True until you start a task, and maybe `kill_tasks_if_necessary` to do nothing.

Rob-Johnson:
I've just had a thought: can we use the newer HTTP scheduler API rather than the Protobuf API? There will be a bit more work to do here given that there are fewer 'hello world' mesos frameworks to take inspiration from, but this is the 'future', so I think it could be worth doing.

https://mesos.apache.org/documentation/latest/scheduler-http-api/

paasta_tools/cli/cmds/remote_run.py
keymone: don't think so, from the docs i got the idea that `driver.run()` (line 256) blocks until driver is stopped. after `.run()` returns we might want to do more stuff depending on exit status, etc.

paasta_tools/cli/cmds/remote_run.py
keymone: i did consider that but amount of code and coupling to paatsa_native configuration stuff made me think it was an overkill so i only used the bit i was interested in - building docker tasks. i can iterate and couple to more paasta_native things later, it helps me to keep it simple for now.

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -76,34 +80,32 @@ def resourceOffers(self, driver, offers):
...
-            paasta_print("THIS IS NOT A DRILL!"
-                         "Would have launched: {task}".format(task=task))
-            self.status = 0
-            driver.stop()
+            driver.launchTasks(offer.id, [task])
 
 
     def statusUpdate(self, driver, update):
-        paasta_print("Mesos Scheduler: task %s is in state %d", update.task_id.value, update.state)
+        paasta_print("Mesos Scheduler: task %s is in state %d" % (update.task_id.value, update.state))
```
Rob-Johnson: there is a bunch more interesting info that comes from these messages. checkout https://github.com/Yelp/chronos/blob/master/src/main/scala/org/apache/mesos/chronos/scheduler/mesos/MesosJobFramework.scala#L205 for the stuff I get for chronos

paasta_tools/cli/cmds/remote_run.py
Rob-Johnson: right, I mean that you'll eventually want to remove this option, because we won't want users running the mesos framework locally.

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -0,0 +1,318 @@
...
+            paasta_print("Task %s is finished." % update.task_id.value)
+            # for stream in mesos.cli.cluster.files(flist=['stdout','stderr'], fltr=update.task_id.value):
+            #     print("Printing %s for task %s" % (stream[0].path, update.task_id.value))
+            #     for line in stream[0].readlines():
+            #         print line
+            driver.stop()
+
+
+    def new_task(self, offer):
+        task = self.service_config.base_task(self.system_paasta_config, portMappings=False)
```
Rob-Johnson: it would be nice if this wasn't bound to a class, but was just a function called ``build_base_task`` that took a ``system_paasta_config`` and a ``service_config`` and returned a new Mesos Task. That way you wouldn't have to depend on this class in the from the native scheduler code. Maybe a problem for another day, but there is definitely some refactoring to be done to make the building of new frameworks easier

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -0,0 +1,318 @@
...
+from paasta_tools.utils import SystemPaastaConfig
+from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import validate_service_instance
+
+from paasta_tools.native_mesos_scheduler import create_driver_with
+from paasta_tools.native_mesos_scheduler import compose_job_id
+from paasta_tools.native_mesos_scheduler import load_paasta_native_job_config
+
+
+class PaastaAdhocScheduler(Scheduler):
```
Rob-Johnson: can you move this class into ``paasta_tools/frameworks/adhoc.py`` (bonus points if you move the native framework too).

paasta_tools/cli/cmds/remote_run.py
keymone: actually i thought having that flag would help with debugging, but also right now the flow is:

    1. paasta remote-run ...
    2. `- ssh mesos-master ...
    3.    `- paasta remote-run ... -l
    4.       `- "curl" mesos-master ...

in absence of `-l` flag how/where is the decision made whether to run ssh (1) or to talk to mesos (2)? would ssh-invoking code would be a separate cli tool or is there some way to reliably say that `$0` is already running on mesos-master?

Rob-Johnson: the way we do this in other places is:

```
1. paasta remote-run ...
2. `- ssh mesos-master ...
3.    `- /usr/bin/paasta-remote-run ...`
```

so the cli part doesn't get executed on the master, but instead invokes another script.

paasta_tools/cli/cmds/remote_run.py
keymone: native framework seems to still be in quite a flux, i don't want to deal with more conflict merging than i have to. let's refactor after this ships.

paasta_tools/cli/cmds/remote_run.py
keymone: done

paasta_tools/cli/cmds/remote_run.py
keymone: as with moving native scheduler to separate module, i'd rather refactor this after PR ships.

paasta_tools/paasta_remote_run.py
```diff
@@ -0,0 +1,135 @@
...
+    command = args.cmd
+
+    if instance is None:
+        instance_type = 'adhoc'
+        instance = 'remote'
+    else:
+        instance_type = validate_service_instance(service, instance, cluster, soa_dir)
+
+    paasta_print('Scheduling a task on Mesos')
+    service_config = load_paasta_native_job_config(
```
solarkennedy: I don't quite understand, are we expecting these jobs to be "adhoc" jobs as in "adhoc-norcal-devc.yaml" ?

Because this looks like it *only* loads something from that class. 

I do see how this is convenient from a scheduler's perspective, but in the long term people are going to want to be able to run their existing "adhoc" and "chronos" jobs.

paasta_tools/frameworks/adhoc_scheduler.py
```diff
@@ -0,0 +1,103 @@
...
+    def registered(self, driver, framework_id, master_info):
+        paasta_print("Registered with framework id: {}".format(framework_id))
+
+
+    def resourceOffers(self, driver, offers):
+        if self.launched:
+            return
+
+        paasta_print("Recieved resource offers: {}".format([o.id.value for o in offers]))
+        offer = offers[0]
```
Rob-Johnson: don't you need to check if the task fits within the offer before you use it?

solarkennedy:
@EvanKrall this is your reminder that you wanted to check this out

EvanKrall:
I still think this has too much copy-paste from the native mesos scheduler. All of the mesos work, including comparing the offer size to the task requirements (as @Rob-Johnson mentioned), is already done for you if you just subclass PaastaNativeServiceConfig.

keymone:
i've moved stuff around and subclassed AdhocScheduler from NativeScheduler. we'll probably want to extract some common bits into BasicScheduler or something because all the LongRunningStuff that NativeScheduler is build around, but i can do that later. this is still wip as i'm trying to hunt down a bug that makes the task fail (so if you happen to notice what am i doing wrong - let me know).

cc @EvanKrall @solarkennedy @Rob-Johnson

paasta_tools/frameworks/adhoc_scheduler.py
```diff
@@ -0,0 +1,42 @@
...
+from paasta_tools.frameworks.native_scheduler import NativeScheduler
+
+class AdhocScheduler(NativeScheduler):
+    def __init__(self, *args, **kwargs):
+        self.dry_run = kwargs.pop('dry_run')
+        self.number_of_runs = 0
+        super(AdhocScheduler, self).__init__(*args, **kwargs)
+
+    def need_more_tasks(self, *args, **kwargs):
+        if self.number_of_runs == 0:
```
EvanKrall: [`NativeScheduler.start_task` checks `self.need_more_tasks` before checking if the resource offer is large enough for the service](https://github.com/Yelp/paasta/pull/1040/files#diff-e18a45d4446addac88f100b1f9874698R186), so if we get a small offer first, we might never launch a task.

I think this is fixable by changing this line to `if self.number_of_runs == 0 and len(self.tasks_with_flags) == 0:` May be worth adding a test for this.

paasta_tools/cli/utils.py
```diff
@@ -554,16 +564,51 @@ def execute_chronos_rerun_on_remote_master(service, instancename, cluster, syste
...
-            -1,
-            'ERROR: could not find connectable master in cluster %s\nOutput: %s' % (cluster, output)
-        )
-    return run_chronos_rerun(master, service, instancename, **kwargs)
+    try:
+        return run_chronos_rerun(
+            connectable_master(cluster, system_paasta_config),
+            service, instancename, **kwargs)
+    except NoMasterError as e:
+        return (-1, str(e))
```
EvanKrall: It's weird that this one returns -1 while the others return 255.

paasta_tools/frameworks/adhoc_scheduler.py
EvanKrall: er, maybe something like this:

```
if self.number_of_runs > 0:
    return False

if len(self.tasks_with_flags) == 0:
    return True
else:
    self.number_of_runs += 1
    return False
```

Or maybe just add the `number_of_runs` counter to the superclass's `start_task`.

keymone: fixed by making `need_more_tasks = len(tasks_with_flags) == 0` and having a flag that depends on outcome of `start_task`.

paasta_tools/cli/utils.py
keymone: yeah, i don't know the reason, just retained original functionality.

keymone:
@EvanKrall thanks for help, i reused native_scheduler behaviors for adhoc test. everything should be green now.

PR #1041, opened 2017-02-22
Don't stream already logged loglines

### LINKED ISSUE #1004: paasta itest prints everything twice
Was trying to debug a failing itest and was very very confused :)

solarkennedy:
There are still some use cases where we *don't* want to stream the output (where we capture it instead), but for itest and push I think it is ok to log and not stream. (as all logs go to stdout/err anyway).

Should I put a check to not allow a caller to stream and log at the same time to prevent someone from duplicating output again?

Fixes #1004

### DIFF
--- a/paasta_tools/cli/cmds/itest.py
+++ b/paasta_tools/cli/cmds/itest.py
@@ -87,7 +87,6 @@ def paasta_itest(args):
         component='build',
         service=service,
         loglevel='debug',
-        stream=True,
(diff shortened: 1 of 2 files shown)


mattmb:
What's actually causing the double printing? I picked a jenkins job at random and everything looks okay...

mattmb:
On a second look it definitely is duplicated.

PR #1044, opened 2017-02-27
Add basic support for NUMA cpu affinity settings. This enable the use…

matfra:
Control which physical CPU and memory node docker will be running on using
docker options "--cpuset-cpus" and "--cpuset-mems". To enable it, simply
pass as a docker environment variable PIN_TO_NUMA_NODE=n where n is the
id of a physical CPU (usually 0 or 1 on a dual CPU machine)

### DIFF
--- a/paasta_tools/docker_wrapper.py
+++ b/paasta_tools/docker_wrapper.py
@@ -8,6 +8,11 @@ underlying docker command.
 If the environment variables are unspecified, or if --hostname is already
 specified, this does not change any arguments and just directly calls docker
 as-is.
+
+Additionally this wrapper will look for the environment variable
+PIN_TO_NUMA_NODE which contains the physical CPU and memory to restrict the
+container to. If the system is NUMA enabled, docker will be called with the
+arguments cpuset-cpus and cpuset-mems.
 """
 from __future__ import absolute_import
 from __future__ import unicode_literals
@@ -84,31 +89,105 @@ def generate_hostname(fqdn, mesos_task_id):
     return hostname
 
 
-def add_hostname(args, hostname):
-    # Add --hostname argument immediately after 'run' command if it exists
+def add_argument(args, argument):
+    # Add an argument immediately after 'run' command if it exists
     args = list(args)
-
     try:
         run_index = args.index('run')
     except ValueError:
         pass
     else:
-        args.insert(run_index + 1, '--hostname=' + hostname)
-
+        args.insert(run_index + 1, argument)
     return args
+
+
+def get_cpumap():
+    # Return a dict containing the core numbers per physical CPU
+    core = 0
+    cpumap = {}
+    try:
+        with open('/proc/cpuinfo', 'r') as f:
+            for line in f:
+                m = re.match('physical\sid.*(\d)', line)
+                if m:
+                    cpuid = int(m.group(1))
+                    if cpuid not in cpumap:
+                        cpumap[cpuid] = []
+                    cpumap[cpuid].append(core)
+                    core += 1
+    except IOError:
+        pass
+    return cpumap
+
+
+def is_numa_enabled():
+    return os.path.exists('/proc/1/numa_maps')
+
+
+def arg_collision(new_args, current_args):
+    # Returns True if one of the new arguments is already in the
+    # current argument list.
+    cur_arg_keys = []
+    for c in current_args:
+        cur_arg_keys.append(c.split('=')[0])
+    return bool(set(new_args).intersection(set(cur_arg_keys)))
+
+
+def validate_int(i):
+    try:
+        return int(i)
+    except (ValueError, TypeError):
+        return None
+
+
+def validate_float(f):
+    # Ensure we return a float. If input is invalid we return 0.0
+    try:
+        return float(f)
+    except (ValueError, TypeError):
+        return 0.0
+
+
+def generate_cpuset_args(env_args, pinned_numa_node):
+    required_cores = validate_float(env_args.get('MARATHON_APP_RESOURCE_CPUS'))
+    cpumap = get_cpumap()
+    result = []
+
+    # No need to set up docker cpuset options if one of the above fails:
+    if (
+        # NUMA is supported by the system
+        is_numa_enabled() and
+        # Machine has multiple CPUs
+        len(cpumap) > 1 and
+        # Requested numa node exists
+        pinned_numa_node in cpumap and
+        # The numa node has more cores than the container requires
+        len(cpumap[pinned_numa_node]) >= required_cores
+    ):
+        result.append('--cpuset-cpus=' + ','.join(str(c) for c in cpumap[pinned_numa_node]))
+        result.append('--cpuset-mems=' + str(pinned_numa_node))
+
+    return result
 
 
 def main(argv=None):
     argv = argv if argv is not None else sys.argv
 
+    # Get Docker env variables
     env_args = parse_env_args(argv)
-    fqdn = socket.getfqdn()
 
     # Marathon sets MESOS_TASK_ID whereas Chronos sets mesos_task_id
     mesos_task_id = env_args.get('MESOS_TASK_ID') or env_args.get('mesos_task_id')
 
-    if mesos_task_id and can_add_hostname(argv):
-        hostname = generate_hostname(fqdn, mesos_task_id)
-        argv = add_hostname(argv, hostname)
+    # Check if we require this container to stick to a given NUMA node
+    pinned_numa_node = validate_int(env_args.get('PIN_TO_NUMA_NODE'))
+
+    if pinned_numa_node and not arg_collision(['--cpuset-cpus', '--cpuset-mems'], argv):
+        for c in generate_cpuset_args(env_args, pinned_numa_node):
+            argv = add_argument(argv, c)
+
+    if mesos_task_id and not already_has_hostname(argv):
+        hostname = generate_hostname(socket.getfqdn(), mesos_task_id)
+        argv = add_argument(argv, '--hostname=' + hostname)
 
     os.execlp('docker', 'docker', *argv[1:])
--- a/paasta_tools/docker_wrapper.py
+++ b/paasta_tools/docker_wrapper.py
@@ -17,11 +17,11 @@ arguments cpuset-cpus and cpuset-mems.
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import logging
 import os
 import re
 import socket
 import sys
-
 
 ENV_MATCH_RE = re.compile('^(-\w*e\w*|--env)(=(\S.*))?$')
 MAX_HOSTNAME_LENGTH = 63
@@ -120,8 +120,18 @@ def get_cpumap():
     return cpumap
 
 
-def is_numa_enabled():
-    return os.path.exists('/proc/1/numa_maps')
+def get_numa_memsize(nb_nodes):
+    # Return memory size in mB per NUMA node assuming memory is split evenly
+    # TODO: calculate and return real memory map
+    try:
+        with open('/proc/meminfo', 'r') as f:
+            for line in f:
+                m = re.match('MemTotal:\s*(\d)kb', line)
+                if m:
+                    return int(m) / 1024 / int(nb_nodes)
+    except IOError:
+        pass
+    return 0
 
 
 def arg_collision(new_args, current_args):
@@ -133,28 +143,41 @@ def arg_collision(new_args, current_args):
     return bool(set(new_args).intersection(set(cur_arg_keys)))
 
 
-def validate_int(i):
-    try:
-        return int(i)
-    except (ValueError, TypeError):
-        return None
-
-
-def validate_float(f):
+def extract_float(f):
     # Ensure we return a float. If input is invalid we return 0.0
     try:
         return float(f)
     except (ValueError, TypeError):
+        logging.warning('Could not read %s as a float' % str(f))
         return 0.0
 
 
-def generate_cpuset_args(env_args, pinned_numa_node):
-    required_cores = validate_float(env_args.get('MARATHON_APP_RESOURCE_CPUS'))
-    cpumap = get_cpumap()
-    result = []
+def is_numa_enabled():
+    if os.path.exists('/proc/1/numa_maps'):
+        return True
+    else:
+        logging.warning('The system does not support NUMA')
+        return False
 
-    # No need to set up docker cpuset options if one of the above fails:
+
+def append_cpuset_args(argv, env_args):
+    # Enable log messages to stderr
+    logging.basicConfig(stream=sys.stderr, level=logging.INFO)
+    try:
+        pinned_numa_node = int(env_args.get('PIN_TO_NUMA_NODE'))
+    except (ValueError, TypeError):
+        logging.warning('Could not read PIN_TO_NUMA_NODE value as an int.' +
+                        str(env_args.get('PIN_TO_NUMA_NODE')))
+        pinned_numa_node = None
+
+    cpumap = get_cpumap()
+
+    # We do nothing if any of these conditions are not met
     if (
+        # PIN_TO_NUMA_NODE has a correct value
+        pinned_numa_node and
+        # cpuset is not already manually set
+        not arg_collision(['--cpuset-cpus', '--cpuset-mems'], argv) and
         # NUMA is supported by the system
         is_numa_enabled() and
         # Machine has multiple CPUs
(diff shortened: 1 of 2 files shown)


paasta_tools/docker_wrapper.py
```diff
@@ -71,31 +76,65 @@ def generate_hostname(fqdn, mesos_task_id):
...
     argv = argv if argv is not None else sys.argv[1:]
 
     env_args = parse_env_args(argv)
-    fqdn = socket.getfqdn()
 
     # Marathon sets MESOS_TASK_ID whereas Chronos sets mesos_task_id
     mesos_task_id = env_args.get('MESOS_TASK_ID') or env_args.get('mesos_task_id')
 
+    # Invalid the variable if it has a bogus value
+    numa_cpuid=env_args.get('NUMA_CPU_AFFINITY')
```
solarkennedy: 1. This is a strange variable name. How about PIN_TO_NUMA_NODE ?
2. Are you anticipating that this is going to be provided by users? What would we actually set this to in production?

We can't just pin to one socket or another. Otherwise we'll get 2 tasks launched with the same pinning and then we'll have *really* slow performance.

This thing must be able to dynamically choose the "free" socket at runtime. The user has to set the variable to imply intent, but this wrapper has to make the decision about which socket to attach to at the last second.

paasta_tools/docker_wrapper.py
```diff
@@ -71,31 +76,65 @@ def generate_hostname(fqdn, mesos_task_id):
...
+    try:
+        int(numa_cpuid)
+    except:
+        numa_cpuid=None
+
+    if numa_cpuid and is_numa_enabled():
+        core_list=get_core_list(int(numa_cpuid))
+        if len(core_list) > 0:
+            argv = add_argument(argv, '--cpuset-cpus=' + ','.join(str(c) for c in core_list))
```
solarkennedy: what if the core_list is < the requested cpus?

matfra: That decision should be made by user/scheduler. Knowing the pool of machine that we have, we should not enable this option for services that require more core than what's on a single CPU on our smallest instance type. (This value would be 8 for most of common EC2 instances).
The good news is, even if we set this by mistake : it won't break anything and Docker will enforce the limit anyway so some thread will have to actively share a core.

paasta_tools/docker_wrapper.py
```diff
@@ -71,31 +76,64 @@ def generate_hostname(fqdn, mesos_task_id):
...
 
     env_args = parse_env_args(argv)
-    fqdn = socket.getfqdn()
 
     # Marathon sets MESOS_TASK_ID whereas Chronos sets mesos_task_id
     mesos_task_id = env_args.get('MESOS_TASK_ID') or env_args.get('mesos_task_id')
 
+    # Invalid the variable if it has a bogus value
+    try:
+        numa_cpuid = int(env_args.get('NUMA_CPU_AFFINITY'))
```
solarkennedy: This still needs some thought on *how* this will actually work in real life when mesos doesn't know that it is scheduling 2 tasks on the same host that have `NUMA_CPU_AFFINITY=0`.

paasta_tools/docker_wrapper.py
matfra: 1. Yeah PIN_TO_NUMA_NODE sounds more accurate!
2. Users should not play with this option. I am only planning to use this on stage and eventually give it a try on prod. If the performance increase is significant, this should definitely be  controlled by either Mesos or a much smarter version of this wrapper that would keep track of the local resources usage. ( I feel like Mesos is the right way to go though )

paasta_tools/docker_wrapper.py
matfra: I agree. For now the goal is to have this working on stage or in a small part of prod where we do not run multiple instances of one service on a single node. This is the very first step towards something much more smarter, or a Mesos patch...

paasta_tools/docker_wrapper.py
```diff
@@ -71,31 +76,64 @@ def generate_hostname(fqdn, mesos_task_id):
...
     # Marathon sets MESOS_TASK_ID whereas Chronos sets mesos_task_id
     mesos_task_id = env_args.get('MESOS_TASK_ID') or env_args.get('mesos_task_id')
 
+    # Invalid the variable if it has a bogus value
+    try:
+        numa_cpuid = int(env_args.get('PIN_TO_NUMA_NODE'))
+    except (ValueError, TypeError):
+        numa_cpuid = None
+
+    if numa_cpuid and is_numa_enabled():
```
bchess: Noting that there's no equivalent of already_has_hostname().. probably fine for these less common settings. But worth noting that it'll probably break the docker run if the arg is already supplied

matfra: This is a good point, should I convert already_has_hostname to a more generic already_has_argument and use re-use it for --cpuset-cpus and --cpuset-mems ?

bchess: Yeah I think that'd be good if you can.

solarkennedy:
@matfra can you take a look at my https://github.com/Yelp/paasta/compare/kwa_numa branch to take a look at what I think might be required to actually do what we want? (just a quick poc)

bchess:
i'm happy with the argparsing stuff

paasta_tools/docker_wrapper.py
```diff
@@ -71,31 +76,96 @@ def generate_hostname(fqdn, mesos_task_id):
...
-    argv = argv if argv is not None else sys.argv[1:]
+    argv = argv if argv is not None else sys.argv
 
     env_args = parse_env_args(argv)
-    fqdn = socket.getfqdn()
 
     # Marathon sets MESOS_TASK_ID whereas Chronos sets mesos_task_id
     mesos_task_id = env_args.get('MESOS_TASK_ID') or env_args.get('mesos_task_id')
 
+    pinned_numa_node = validate_int(env_args.get('PIN_TO_NUMA_NODE'))
```
solarkennedy: This is all way too much to be in the "main" function.

Please extract as much as possible into a different function so that the code path in main is minimal. (only branch into numa stuff if the env var exists)

paasta_tools/docker_wrapper.py
```diff
@@ -71,31 +76,105 @@ def generate_hostname(fqdn, mesos_task_id):
...
 
+    # Get Docker env variables
     env_args = parse_env_args(argv)
-    fqdn = socket.getfqdn()
 
     # Marathon sets MESOS_TASK_ID whereas Chronos sets mesos_task_id
     mesos_task_id = env_args.get('MESOS_TASK_ID') or env_args.get('mesos_task_id')
 
+    # Check if we require this container to stick to a given NUMA node
+    pinned_numa_node = validate_int(env_args.get('PIN_TO_NUMA_NODE'))
```
solarkennedy: The validate_int stuff is kinda puppety and not very pythony.

Please just return None and key off that and only turn to `int()` at the last second.

I want this whole block to be only 2 lines:

```
if env_get.get('Pin..):
    argv = append_cpuset_args(argv, env_get.get('Pin..))
```

For 2 reasons:
1. The more code out of main means it is under test
2. The the non-pinning code path "looks" even smaller and smaller" if you are stepping through the code with your eyes

paasta_tools/docker_wrapper.py
```diff
@@ -71,31 +76,105 @@ def generate_hostname(fqdn, mesos_task_id):
...
+def validate_float(f):
+    # Ensure we return a float. If input is invalid we return 0.0
+    try:
+        return float(f)
+    except (ValueError, TypeError):
+        return 0.0
+
+
+def generate_cpuset_args(env_args, pinned_numa_node):
+    required_cores = validate_float(env_args.get('MARATHON_APP_RESOURCE_CPUS'))
```
solarkennedy: please don't do the validate float stuff and instead be explicit about your errors.

```
if not required_cores:
   log.error('Could not extract requested cores from: "%s". Defaulting to 0.0')
   return []
```

paasta_tools/docker_wrapper.py
```diff
@@ -71,31 +76,105 @@ def generate_hostname(fqdn, mesos_task_id):
...
+    cpumap = get_cpumap()
+    result = []
+
+    # No need to set up docker cpuset options if one of the above fails:
+    if (
+        # NUMA is supported by the system
+        is_numa_enabled() and
```
solarkennedy: I would like the new `append_cpuset_args` to be in charge of doing this businessy logic, and I want this function to only be responsible for calculating the args.

Make it a real "function" only inputs and outputs.

solarkennedy: In the new `append_cpuset_args` function I would like better stderr lines to help us debug in the conditions where it isn't adding args:

```
if not numa_enabled():
    log.warning("NUMA optimization was requested but this is not a numa-enabled machine. No optimizations applied")
    return argv
if len(cpus) <= 1:
    log.warning(....
if......

# by now we know we need some args
append...
```
Again these will only show up to stderr when the user has requested numa.

And I dare say that we *should* log to stderr when we a user requests numa and we are going to do something about it.

```
print("NUMA optimization requested. Pinning to NUMA zone %d.", stderr)
```

paasta_tools/docker_wrapper.py
```diff
@@ -71,31 +76,105 @@ def generate_hostname(fqdn, mesos_task_id):
...
+        # Machine has multiple CPUs
+        len(cpumap) > 1 and
+        # Requested numa node exists
+        pinned_numa_node in cpumap and
+        # The numa node has more cores than the container requires
+        len(cpumap[pinned_numa_node]) >= required_cores
+    ):
+        result.append('--cpuset-cpus=' + ','.join(str(c) for c in cpumap[pinned_numa_node]))
+        result.append('--cpuset-mems=' + str(pinned_numa_node))
```
solarkennedy: I guess a TODO line for verifying that the mem requested is under the numa limit? Seems like an easy mistake to make.

tests/test_docker_wrapper.py
```diff
@@ -203,3 +204,273 @@ def test_not_run(self, mock_execlp):
...
+            is_numa_enabled=False,
+            cpu_info=[
+                'physical id    : 0',
+                'physical id    : 1',
+                'physical id    : 0',
+                'physical id    : 1',
+            ],
+        ):
+            docker_wrapper.main(argv)
+        argv = [
```
solarkennedy: This argv looks kinda out of nowhere, is it in the right place?

paasta_tools/docker_wrapper.py
matfra: This was not part of my initial idea of iterating over a simple version of the wrapper but I agree that it's a necessary precaution

paasta_tools/docker_wrapper.py
matfra: Making it even more minimalist as you suggest will oblige me to pass args and env_args (need to read MARATHON RESOURCES too) to the append_cpuset_args function, making it some kind of main "bis"

matfra: Or we extract all the docker env variables in the main and pass them to the function.

solarkennedy: Yep, you will have to pass in env_args and args.

solarkennedy:
Can you rebase to pick up new changes? Also `local-run` now will use the wrapper if it makes it easier to test.

matfra:
Doing this now!

matfra:
Memory size calculation for memory safeguard et pretty dumb for now : just dividing the total amount of RAM by the number of physical CPU. This should work in most of the cases but it will be better to get the real numbers (from /proc/zoneinfo for instance)

paasta_tools/docker_wrapper.py
```diff
@@ -84,31 +89,128 @@ def generate_hostname(fqdn, mesos_task_id):
...
+        logging.warning('Could not read PIN_TO_NUMA_NODE value as an int.' +
+                        str(env_args.get('PIN_TO_NUMA_NODE')))
+        pinned_numa_node = None
+
+    cpumap = get_cpumap()
+
+    # We do nothing if any of these conditions are not met
+    if (
+        # PIN_TO_NUMA_NODE has a correct value
+        pinned_numa_node and
```
solarkennedy: Don't need this "really" we already know that the user requested this

paasta_tools/docker_wrapper.py
```diff
@@ -84,31 +89,128 @@ def generate_hostname(fqdn, mesos_task_id):
...
+        # cpuset is not already manually set
+        not arg_collision(['--cpuset-cpus', '--cpuset-mems'], argv) and
```
solarkennedy: This if is way too complex and the comments don't add a lot of value.

Make it "self documenting" by testing each condition and outputting each thing and returning the right result.

```
if arg_collision():
    log.error('args already have the things')
    return argv
if not is_numa_enabled():
    log.warning('numa optimization requested, but not on a numa system. Doing nothing.')
    return argv
if ....
```

paasta_tools/docker_wrapper.py
```diff
@@ -84,31 +89,128 @@ def generate_hostname(fqdn, mesos_task_id):
...
+        return False
+
+
+def append_cpuset_args(argv, env_args):
+    # Enable log messages to stderr
+    logging.basicConfig(stream=sys.stderr, level=logging.INFO)
+    try:
+        pinned_numa_node = int(env_args.get('PIN_TO_NUMA_NODE'))
+    except (ValueError, TypeError):
+        logging.warning('Could not read PIN_TO_NUMA_NODE value as an int.' +
```
solarkennedy: + stuff is discouraged

Instead use the preferred format string stuff

https://pyformat.info/

matfra:
Applied @solarkennedy suggestions and corrections. Rebased on the latest master.

paasta_tools/docker_wrapper.py
```diff
@@ -84,31 +89,141 @@ def generate_hostname(fqdn, mesos_task_id):
...
+        logging.error('Could not read PIN_TO_NUMA_NODE value as an int: {}'.format(
+            env_args.get('PIN_TO_NUMA_NODE')))
+        return argv
+
+    cpumap = get_cpumap()
+
+    if len(cpumap) < 1:
+        logging.error('Less than 2 physical CPU detected')
+        return argv
+
```
solarkennedy: style-wise I think all this vertical whitespace is not needed between all the ifs. Please group all the sanity checks visually by removing the whitespace between the ifs.

paasta_tools/docker_wrapper.py
```diff
@@ -84,31 +89,141 @@ def generate_hostname(fqdn, mesos_task_id):
...
+
 def main(argv=None):
     argv = argv if argv is not None else sys.argv
 
+    # Get Docker env variables
     env_args = parse_env_args(argv)
-    fqdn = socket.getfqdn()
+
+    # Check if we require this container to stick to a given NUMA node
```
solarkennedy: I don't think this comment or the one on 215 add anything, the code is self-explanatory

paasta_tools/docker_wrapper.py
```diff
@@ -84,31 +89,141 @@ def generate_hostname(fqdn, mesos_task_id):
...
+    if arg_collision(['--cpuset-cpus', '--cpuset-mems'], argv):
+        logging.error('--cpuset options are already set. Not overriding')
+        return argv
+
+    if not is_numa_enabled():
+        logging.error('Could not detect NUMA on the system')
+        return argv
+
+    if len(cpumap[pinned_numa_node]) < extract_float(env_args.get(
+            'MARATHON_APP_RESOURCE_CPUS')):
```
solarkennedy: Can you extract this :196 and :201 into a extract_mem_requirement and extract_cpu_requirement and make that do the logic of extracting the right variable and float'ing it? My reasoning is that we need more than marathon support (Eventually) and having a function that encapsulates *how* we extract these values will make it tidy and testable.

In other words, I don't want this function `append_cpuset_args` to "know" about marathon at all.

paasta_tools/docker_wrapper.py
```diff
@@ -84,31 +89,141 @@ def generate_hostname(fqdn, mesos_task_id):
...
+        argv = append_cpuset_args(argv, env_args)
 
     # Marathon sets MESOS_TASK_ID whereas Chronos sets mesos_task_id
     mesos_task_id = env_args.get('MESOS_TASK_ID') or env_args.get('mesos_task_id')
 
     if mesos_task_id and can_add_hostname(argv):
-        hostname = generate_hostname(fqdn, mesos_task_id)
-        argv = add_hostname(argv, hostname)
+        hostname = generate_hostname(socket.getfqdn(), mesos_task_id)
+        argv = add_argument(argv, '--hostname=' + hostname)
```
solarkennedy: No `+`s. Do `'--hostname=%s' % hostname`

tests/test_docker_wrapper.py
```diff
@@ -212,3 +213,377 @@ def test_not_run(self, mock_execlp):
...
+            docker_wrapper.main(argv)
+
+        assert mock_execlp.mock_calls == [mock.call(
+            'docker',
+            'docker',
+            'run',
+            '--env=PIN_TO_NUMA_NODE=2',
+        )]
+
+    def test_numa_single_cpu(self, mock_execlp):
```
solarkennedy: test_numa_single_cpu_doesnt_bother_with_cpusets

tests/test_docker_wrapper.py
```diff
@@ -212,3 +213,377 @@ def test_not_run(self, mock_execlp):
...
+
+        assert mock_execlp.mock_calls == [mock.call(
+            'docker',
+            'docker',
+            'run',
+            '--env=PIN_TO_NUMA_NODE=1',
+            '--env=MARATHON_APP_RESOURCE_CPUS=3.0',
+        )]
+
+    def test_numa_enabled_no_marathon(self, mock_execlp):
```
solarkennedy: test_numa_enabled_unknown_cpu_requirement_skips_cpusets

tests/test_docker_wrapper.py
```diff
@@ -212,3 +213,377 @@ def test_not_run(self, mock_execlp):
...
+            'docker',
+            'docker',
+            'run',
+            '--cpuset-mems=1',
+            '--cpuset-cpus=1,3',
+            '--env=PIN_TO_NUMA_NODE=1',
+            '--env=MARATHON_APP_RESOURCE_CPUS=2',
+        )]
+
+    def test_numa_req_too_many_cores(self, mock_execlp):
```
solarkennedy: test_numa_req_too_many_cores_requested

matfra:
Ready to press the merge button

solarkennedy:
@matfra let's pin the next release "just in case" as we get confidence that this wrapper works IRL, then in the future changes to this thing will get included in the normal push.

I think we might currently be pinned due to a big change from krall and @nhandler, so sit tight a bit before trying to use this.

PR #1045, opened 2017-02-27
Do not fail wait-for-deployment when the git server is unreachable

oktopuz:
Do not fail wait-for-deployment when the git server is temporary unreachable.

internal ticket: PAASTA-8599

### DIFF
--- a/paasta_tools/cli/cmds/wait_for_deployment.py
+++ b/paasta_tools/cli/cmds/wait_for_deployment.py
@@ -30,6 +30,7 @@ from paasta_tools.cli.utils import validate_full_git_sha
 from paasta_tools.cli.utils import validate_given_deploy_groups
 from paasta_tools.cli.utils import validate_service_name
 from paasta_tools.remote_git import list_remote_refs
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/wait_for_deployment.py
```diff
@@ -137,9 +138,16 @@ def validate_git_sha(git_sha, git_url, deploy_group, service):
     Raise exception when the provided git_sha is not the latest
     marked for deployment in 'deploy_group' for 'service'.
```
solarkennedy: This is kinda weird. I remember why I added this in there: we want to protect people from "waiting forever" on a sha that isn't marked.

But in reality, this is kinda slow and if we are not going to block on it anyway, I think maybe we don't really need this verification.

oktopuz: It takes 2 second to validate this.
Looks like a reasonable tradeoff to me to avoid long waiting for something that wasn't even marked for deployment. Take a look at PAASTA-8570 for instance.

PR #1047, opened 2017-02-28
Add optional autoscaling info to metastatus

### LINKED ISSUE #984: Autoscaling info in metastatus
now that we're relying on this in prod, we should include some info about cluster autoscaling in ``paasta metastatus``. Particularly, I'd like to see that list of 'things' (asg, sfrs) that the autoscaler knows about, the min, max, current capacity. Maybe even 'am I scaling up or down right now'?

ping @mattmb since you'll be interested in this.

mattmb:
This adds optional info on the cluster autoscaler to metastatus. It
works by using the autoscaling library to retrieve details on the
resources it is responsible for and contacting AWS to find the current
capacities of the resources. The motivation is so an operator can
easily see the settings relavent to the autoscaler and the current
capacity. Fixes #984

It is behind a command line flag so should be safe to ship as is. At
yelp this will need a wrapper script on the master that ensures the
autoscaling configs are synced from S3 and that there are AWS
credentials in the environment.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -385,7 +385,7 @@ class SpotAutoscaler(ClusterAutoscaler):
 
     @property
     def exists(self):
-        return True if self.sfr else False
+        return False if not self.sfr or self.sfr['SpotFleetRequestState'] == 'cancelled' else True
 
     def get_sfr(self, spotfleet_request_id, region=None):
         ec2_client = boto3.client('ec2', region_name=region)
@@ -783,7 +783,7 @@ def autoscaling_info_for_resource(resource, pool_settings):
         dry_run=True
     )
     if not scaler.exists:
-        log.info("no scaler for resource {}. ignoring").format(resource['id'])
+        log.info("no scaler for resource {}. ignoring".format(resource['id']))
         return None
     try:
         current_capacity, target_capacity = scaler.metrics_provider()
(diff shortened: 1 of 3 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -734,3 +752,38 @@ def get_instances_from_ip(ip, instance_descriptions):
     :returns: list of instance description dicts"""
     instances = [instance for instance in instance_descriptions if instance['PrivateIpAddress'] == ip]
     return instances
+
+
+def get_autoscaling_info():
```
mattmb: Hmm, I initially had this function in `metastatus_lib` but thought it would live better here...

I see what you mean though. Maybe I should return a list of the `AutoscalingInfo` objects and then just create the `table` in `metastatus_lib`?

Rob-Johnson: sure, sg2m

mattmb: Oh, except that creates a circular import because the autoscaling lib already imports `metastatus_lib` :-(

I think I'll have to leave this as is

Rob-Johnson: okay - let me PR with a refactor against your branch before merging?

mattmb: Sure, feel free 👍

PR #1048, opened 2017-02-28
More links

solarkennedy:
@bchess we need this if we want the wrapper to be in the path (nice to have)


bchess:
shipit

solarkennedy:
Where is travis?

solarkennedy:
oh, same outage :(
https://www.traviscistatus.com/

PR #1056, opened 2017-03-03
Make local_run use the new docker wrapper

solarkennedy:
cc @bchess 

Do we support env vars with spaces in them?
Because when local run was passing `--env="MESOS_TASK_ID=foo"` to the wrapper it interpreted the var as `u'"MESOS_TASK_ID': foo"'}`

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -368,9 +368,9 @@ def get_container_name():
 
 def get_docker_run_cmd(memory, random_port, container_name, volumes, env, interactive,
                        docker_hash, command, net, docker_params):
-    cmd = ['docker', 'run']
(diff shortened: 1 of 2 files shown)


bchess:
That's a good point - i don't have a test case for that, it's probably true. `--env "MESOS_TASK_ID=foo"` may parse correctly however?

bchess:
@solarkennedy I'm less sure this is a valid case. Spaces, as long as their quoted, would ensure that it's still passed as a single argv. That works fine. But bash will absorb the quotes, so those don't get included in the argv, even if they're in the middle of the arg.

Only way I could re-create this on the CLI was if I did something like '--env="foo=bar bing"'. To be fair, `docker run` (or something lower) does appear to strip these quotes but this hardly seems standard.

PR #1060, opened 2017-03-06
Integrated test.sh in Makefile

matfra:
Now make test is a little bit smarter. It doesn't systematically delete the .paasta virtualenv
Non primary runs of make test now takes 8 seconds on my laptop.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,8 +36,7 @@ test: .paasta/bin/activate
 	touch .paasta/bin/activate
 
 itest: test .paasta/bin/activate
-	.paasta/bin/tox -e general_itests
-	.paasta/bin/tox -e paasta_itests
+	./itest.sh
 
 itest_%:
 	# See the makefile in yelp_package/Makefile for packaging stuff
--- a/Makefile
+++ b/Makefile
@@ -22,14 +22,22 @@ endif
 
 .PHONY: all docs test itest
 
-docs:
-	./docs.sh
+docs: .paasta/bin/activate
+	.paasta/bin/tox -e docs
 
-test:
-	./test.sh
+test: .paasta/bin/activate
+	.paasta/bin/tox
 
-itest: test
-	./itest.sh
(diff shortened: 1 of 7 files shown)


matfra:
Oops, python 3.7 is not gonna work ;-)

matfra:
Integrated itest.sh and docs.sh to Makefile too

Makefile
```diff
@@ -22,14 +22,22 @@ endif
 
 .PHONY: all docs test itest
 
-docs:
-	./docs.sh
+docs: .paasta/bin/activate
```
solarkennedy: Can you ship these makefile changes in a different PR so they can be bike-sheded independently?

solarkennedy: Wait sorry this *is* a different PR :)

Makefile
```diff
@@ -22,14 +22,22 @@ endif
...
-test:
-	./test.sh
+test: .paasta/bin/activate
+	.paasta/bin/tox
 
-itest: test
-	./itest.sh
+.paasta/bin/activate: requirements.txt requirements-dev.txt
+	test -d .paasta/bin/activate || virtualenv -p python2.7 .paasta
+	.paasta/bin/pip install -U pip
```
drolando: you should pin the pip virtualenv and tox versions

matfra:
Pinned version as recommended by @drolando . But is it really necessary?

mattmb:
I've been meaning to pin the versions for a while actually, could you do me a favour and have a grep for anywhere else we're installing pip/virtualenv etc. and pin them too. I'm pretty sure there's some stuff in the various Dockerfiles.

Thanks for cleaning this up, I made a bit of a mess when I spent a few days hitting it with a hammer until the builds went green!

matfra:
@mattmb I've pinned the version as you requested. I left itest.sh intact has make itest won't finish properly on my mac without it (this is because by default itest.sh is doesn't exit on failure and Makefile does)
Also I rebased on the latest master. I haven't tried to make release

PR #1061, opened 2017-03-06
Add service autoscaling info to status

mattmb:
Inspired by recent addition of cluster autoscaling to metastatus this
will give service autoscaling info in paasta status. Still need to fix
some tests and check it works properly. Just chucking up for visibility.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -372,8 +372,17 @@ def get_autoscaling_info(service, instance, cluster, soa_dir):
         cluster=cluster,
         soa_dir=soa_dir,
     )
+    marathon_config = load_marathon_config()
+    marathon_client = get_marathon_client(
+        url=marathon_config.get_url(),
+        user=marathon_config.get_username(),
+        passwd=marathon_config.get_password())
+    all_marathon_tasks, all_mesos_tasks = get_all_marathon_mesos_tasks(marathon_client)
     if service_config.get_max_instances() and service_config.get_desired_state() == 'start':
-        marathon_tasks, mesos_tasks = get_autoscaling_tasks(service_config)
+        marathon_tasks, mesos_tasks = filter_autoscaling_tasks(marathon_client,
+                                                               all_marathon_tasks,
+                                                               all_mesos_tasks,
+                                                               service_config)
 
         autoscaling_params = service_config.get_autoscaling_params()
         autoscaling_params.update({'noop': True})
@@ -393,7 +402,7 @@ def get_autoscaling_info(service, instance, cluster, soa_dir):
         return ServiceAutoscalingInfo(current_instances=str(service_config.get_instances()),
                                       max_instances=str(service_config.get_max_instances()),
                                       min_instances=str(service_config.get_min_instances()),
-                                      current_utilization="{0:.1f}%".format(utilization * 100),
+                                      current_utilization="{:.1f}%".format(utilization * 100),
                                       target_instances=str(new_instance_count))
     return None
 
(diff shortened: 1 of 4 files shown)


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -453,38 +519,43 @@ def autoscale_services(soa_dir=DEFAULT_SOA_DIR):
...
                             autoscale_marathon_instance(config, list(marathon_tasks.values()), mesos_tasks)
                         except Exception as e:
                             write_to_log(config=config, line='Caught Exception %s' % e)
     except LockHeldException:
         log.warning("Skipping autoscaling run for services because the lock is held")
 
 
+def get_autoscaling_tasks(config):
+    marathon_config = load_marathon_config()
+    marathon_client = get_marathon_client(
```
solarkennedy: I don't like how we make a new marathon client every single time this function is called. I don't know if that "hurts" a lot, but I bet the code would be faster if you passed the client object in, from status or from the other autoscaling tools. Not only would you not have to read creds of disk all the time, but maybe caching would be persistent?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -453,38 +519,43 @@ def autoscale_services(soa_dir=DEFAULT_SOA_DIR):
...
+        url=marathon_config.get_url(),
+        user=marathon_config.get_username(),
+        passwd=marathon_config.get_password())
+    all_marathon_tasks = marathon_client.list_tasks()
+    all_mesos_tasks = get_all_running_tasks()
```
solarkennedy: Before we called this one for the whole loop, now we are calling it once per marathon app, I don't think this will scale

mattmb: Hmm, good spot. They're not super expensive call iirc but there's no real reason to make this worse than it was before. I think I'll need to split this into two functions `get_all_tasks` and `filter_autoscaling_tasks`. I'll send a fixup when I get a min.

paasta_tools/autoscaling/autoscaling_service_lib.py
solarkennedy: If @oktopuz's change is merged, this might significantly slow down paasta status again because it will be a new client for every instance and not be cached.

mattmb: That's a good shout, I'll add a commit to pass in the client if it's not too hard...

paasta_tools/marathon_serviceinit.py
```diff
@@ -149,15 +151,22 @@ def status_marathon_job(service, instance, app_id, normal_instance_count, client
...
+def get_verbose_status_of_marathon_app(marathon_client, app, service, instance, cluster, soa_dir):
     """Takes a given marathon app object and returns the verbose details
     about the tasks, times, hosts, etc"""
     output = []
     create_datetime = datetime_from_utc_to_local(isodate.parse_datetime(app.version))
     output.append("  Marathon app ID: %s" % PaastaColors.bold(app.id))
     output.append("    App created: %s (%s)" % (str(create_datetime), humanize.naturaltime(create_datetime)))
-    output.append("    Tasks:")
 
+    autoscaling_info = get_autoscaling_info(marathon_client, service, instance, cluster, soa_dir)
```
solarkennedy: Does it not seem strange that we print this for every marathon app, but it should be "global" for an app/instance, right?

Also this is relatively expensive to run, and many don't have autoscaling info. (see line 375)

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -355,6 +365,80 @@ def get_error_from_utilization(utilization, setpoint, current_instances):
...
 
 
+def get_autoscaling_info(marathon_client, service, instance, cluster, soa_dir):
+    service_config = load_marathon_service_config(
+        service=service,
+        instance=instance,
+        cluster=cluster,
+        soa_dir=soa_dir,
+    )
+    all_marathon_tasks, all_mesos_tasks = get_all_marathon_mesos_tasks(marathon_client)
```
solarkennedy: This can be expensive, can you short-cirtuit this and only run it after line 376?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -355,6 +365,80 @@ def get_error_from_utilization(utilization, setpoint, current_instances):
...
+    if service_config.get_max_instances() and service_config.get_desired_state() == 'start':
+        marathon_tasks, mesos_tasks = filter_autoscaling_tasks(marathon_client,
+                                                               all_marathon_tasks,
+                                                               all_mesos_tasks,
+                                                               service_config)
+
+        autoscaling_params = service_config.get_autoscaling_params()
+        autoscaling_params.update({'noop': True})
```
solarkennedy: I don't really "like" that `get_autoscaling_params` does more than get, it *sets* them in ZK! 
Eh

mattmb: It doesn't does it? https://github.com/Yelp/paasta/blob/93d140eac5fcde50eb0b892494dc7031c7408c4a/paasta_tools/marathon_tools.py#L202-L208

Just fetching the config as far as I can see?

paasta_tools/autoscaling/autoscaling_service_lib.py
mattmb: Very good point!

paasta_tools/marathon_serviceinit.py
mattmb: Interesting question. I think in the case where we have two marathon apps the autoscaler will only act on the one that matches the current config+git sha. So really we should probably only print this info on the "new" app...

I think for now I'll leave it as it is, just so we don't block on shipping this. With the change on L375 it shouldn't be too expensive as it's just checking for the presence of min+max instances in cached yaml files.

PR #1062, opened 2017-03-06
Properly quote custom paasta local-run command

drolando:
We're double quoting the custom command received via `paasta local-run -C`. This is probably fine if the command is a single word, but it fails if it's something slightly more complex like "make test" since bash will now consider the entire string as the command name.

I'm not sure if this is the best way to check the command inside the docker run args list, but afaik the custom command must always be at the end of the docker run, so this should be fine.

With `pipes.quote`, the test failed like:
```
    def test_get_docker_run_cmd_quote_cmd():
        memory = 555
        random_port = 666
        container_name = 'Docker' * 6 + 'Doc'
        volumes = ['7_Brides_for_7_Brothers', '7-Up', '7-11']
        env = {}
        interactive = True
        docker_hash = '8' * 40
        command = ['make', 'test']
        net = 'host'
        docker_params = []
        actual = get_docker_run_cmd(memory, random_port, container_name, volumes, env,
                                    interactive, docker_hash, command, net, docker_params)

>       assert actual[-3:] == [u'sh', u'-c', u'make test']
E       assert ['sh', '-c', "'make test'"] == ['sh', '-c', 'make test']
E         At index 2 diff: u"'make test'" != u'make test'
E         Full diff:
E         - [u'sh', u'-c', u"'make test'"]
E         ?                 -           -
E         + [u'sh', u'-c', u'make test']

tests/cli/test_cmds_local_run.py:727: AssertionError
```

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -19,7 +19,6 @@ import datetime
 import functools
 import json
 import os
-import shlex
 import socket
 import sys
 import time
@@ -390,7 +389,7 @@ def get_docker_run_cmd(memory, random_port, container_name, volumes, env, intera
         cmd.extend((
             'sh', '-c'
         ))
-        cmd.append(' '.join(command))
+        cmd.append(command)
     return cmd
 
 
@@ -565,7 +564,7 @@ def run_docker_container(
     container_started = False
     container_id = None
     try:
-        (returncode, output) = _run(joined_docker_run_cmd)
+        (returncode, output) = _run(docker_run_cmd)
         if returncode != 0:
             paasta_print(
                 'Failure trying to start your container!'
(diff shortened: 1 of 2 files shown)


solarkennedy:
Woot!

lucagiovagnoli:
thanks @drolando! This is what I was mentioning last week @solarkennedy
+1

drolando:
I just run `python ../../packages/paasta/paasta_tools/cli/cli.py local-run -C 'make test-in-container'` on my devbox and it worked correctly

solarkennedy:
Can you run the general_itests (tox -e general_itests) and make sure they pass too?

drolando:
@solarkennedy @EvanKrall All tests and itests are now passing, but I'd like you to double check my change. I've followed krall's suggestion and stopped converting the command back and forth from string to array. It looks like Popen is able to properly quote stuff, so this now work both interactively and not.

The paasta_itests are flaky. The first time I got a timeout error, but then I rerun them and everything worked. They also passed on my devbox. Is this a known problem?

solarkennedy:
Yea it happens (flakey)

PR #1065, opened 2017-03-06
Stop querying slaves to figure out routing constraints or to determine smartstack healthiness of tasks

EvanKrall:
This provides us two benefits:

1. It helps address our current problem of CPU-overloaded mesos masters by reducing the calls against them; it also makes the bounce more likely to make progress in the face of overloaded mesos masters by reducing the number of things that can time out.
2. It allows us to control when routing constraints change, e.g. if we want to add or remove a region from a cluster, we can change the routing constraints before or afterwards. (Seems preferable to the current behavior of paasta automagically changing the routing constraints when we go from N>0 to N=0 or N=0 to N>0 slaves in a particular location.


paasta_tools/marathon_tools.py
```diff
@@ -269,21 +268,23 @@ def get_routing_constraints(self, service_namespace_config, system_paasta_config
...
-        if not slaves:
-            raise NoSlavesAvailableError(
-                "No slaves could be found in the cluster."
-            )
+
+        expected_slave_attributes = system_paasta_config.get_expected_slave_attributes()
+        if expected_slave_attributes is None:
+            return []
+
+        fake_slaves = [{"attributes": a} for a in expected_slave_attributes]
```
Rob-Johnson: this is odd. I guess this is just a shortcut so we can make ``filter_mesos_slaves_by_blacklist`` work for us?

paasta_tools/utils.py
```diff
@@ -1015,6 +1015,11 @@ def get_deploy_whitelist(self):
         """
         return self.get("deploy_whitelist", [])
 
+    def get_expected_slave_attributes(self):
+        """Return a list of dictionaries, representing the expected combinations of attributes in this cluster. Used for
```
Rob-Johnson: why's this a list of dictionaries, rather than just a dictionary of ``{ discover_type: possible_values}``?

EvanKrall: The list of dictionaries (each representing an example slave) helps us when we have constraints. E.g. if you `discover: habitat` but have `deploy_whitelist: [region, [uswest1-prod]]`, then we need to set constraints `['habitat', 'GROUP_BY', '2']` instead of `['habitat', 'GROUP_BY', '3']`

I could also figure some of this out by reading /nail/etc/services/location_mapping.json instead of a new configuration key, but that doesn't have any information about e.g. pools -- if you use a pool that only exists in one region, you've effectively added a `[region, LIKE, ...]` constraint

paasta_tools/marathon_tools.py
EvanKrall: Yep. And `filter_mesos_slaves_by_blacklist` is still used elsewhere, otherwise I would've refactored it to filter attributes.

EvanKrall:
Internal puppet CR for deploying the configs: https://reviewboard.yelpcorp.com/r/203311/

solarkennedy:
Please pin + ship

PR #1066, opened 2017-03-07
iterating on remote-run

keymone:
- make -d argument actually skip launching any tasks
- fix `paasta remote-run` cli tool ssh command
- make things more immutable inside NativeScheduler
  - `start_task` -> `tasks_for_offer` just returns the tasks that would be launched
  - `launch_tasks_for_offer` is the happy path of `resourceOffers`
  - `need_more_tasks` depends on existing tasks and tasks that are being scheduled, as it is repeatedly called within `tasks_for_offer`
- don't immediately remove finished tasks from `self.tasks_with_flags` to allow to react to dead tasks up the inheritance chain

### DIFF
--- a/paasta_tools/frameworks/native_scheduler.py
+++ b/paasta_tools/frameworks/native_scheduler.py
@@ -99,6 +99,7 @@ class NativeScheduler(mesos.interface.Scheduler):
 
         if service_config is not None:
             self.service_config = service_config
+            self.service_config.config_dict.update(service_config_overrides)
             self.recreate_drain_method()
         else:
             self.load_config()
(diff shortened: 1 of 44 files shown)


Rob-Johnson:
please leave things as ``snake_case`` and remove the ``camelCase``.

EvanKrall:
The mixed case is a little funky, but the mesos scheduler driver expects camelCase on certain methods. For the methods that aren't part of the Scheduler API, I went with snake_case for python consistency.

keymone:
@Rob-Johnson @EvanKrall 

from pep8 i read that snake_case is the preferred unless camelCase is already used. i really don't like mixing the two. i'll try to figure out a better way instead of moving everything over to camel.

keymone:
@EvanKrall yes, i thought about it but in the end decided that it's not worth the hassle, in the end it's only 2 or 3 methods that are camelCase.

i've updated this PR title and description as the scope has grown.

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -137,24 +158,37 @@ def task_fits(self, offer):
...
+                not(self.is_task_new(name, tid)) and
+                self.tasks_with_flags[tid].mesos_task_state in LIVE_TASK_STATES,
+            tasks))
 
     def is_task_new(self, name, tid):
         return tid.startswith("%s." % name)
 
-    def start_task(self, driver, offer):
+    def tasks_for_offer(self, driver, offer):
         """Starts a task using the offer, and subtracts any resources used from the offer."""
```
Rob-Johnson: can you update this comment? as I understand it, it doesn't start any tasks - just returns a list of MesosTasks to be run using the offer.

PR #1067, opened 2017-03-07
Cache expensive endpoints more

mattmb:
Just putting this up there as a potential easy thing that might
alleviate our performance issues in our biggest clusters.

Currently can take ~4s to get state.json so a 5s cache can't be helping
a lot. We could make this configurable if anyone feels strongly?
Otherwise I can't see it hurting that much.

cc @EvanKrall @solarkennedy not really a real fix but it is at least an easy one. I think we call `_frameworks` quite a bit because it seems to be used to get a list of tasks. i.e. `mesos.master.tasks`

### DIFF
--- a/paasta_tools/mesos/master.py
+++ b/paasta_tools/mesos/master.py
@@ -133,7 +133,7 @@ class MesosMaster(object):
         else:
             return cfg
 
-    @util.CachedProperty(ttl=5)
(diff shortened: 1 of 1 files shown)


solarkennedy:
cc @huadongliu 

I agree the defaults here are probably not helping. I say take it to 60s!

PR #1070, opened 2017-03-08
Set StrictHostKeyChecking=no

nhandler:
This is to work around some issues with ephemeral masters


solarkennedy:
Don't forget to use precommit

Issue #1074, opened 2017-03-09
override_dh_virtualenv forces paasta packaging to use internal pypi


nosmo:
When run outside of Yelp, `make yelp_package` fails due to the use of pypi.yelpcorp.com in debian/rules. This contradicts the behaviour used in Makefiles where the check `ifeq ($(PAASTA_ENV),YELP)` is used to allow external usage.

PR #1075, opened 2017-03-09
Only use internal pypi when requested

nosmo:
Addresses #1074.

### DIFF
--- a/yelp_package/Makefile
+++ b/yelp_package/Makefile
@@ -71,7 +71,7 @@ clean:
 VERSION = $(firstword $(subst -, ,$(RELEASE) ))
 # This sed isn't really right but at least it doesn't crash (and I couldn't get
 # \' or ''' or \\' to work :( )
-LAST_COMMIT_MSG = $(shell git log -1 --pretty=%B | sed -e "s/'/"/g")
+LAST_COMMIT_MSG = $(shell git log -1 --pretty=%B | sed -e 's/'\''/"/g')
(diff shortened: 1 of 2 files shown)


nhandler:
Can you resolve the conflicts?

solarkennedy:
@nosmo bump

yelp_package/Makefile
```diff
@@ -77,7 +77,7 @@ clean:
 VERSION = $(firstword $(subst -, ,$(RELEASE) ))
 # This sed isn't really right but at least it doesn't crash (and I couldn't get
 # \' or ''' or \\' to work :( )
```
solarkennedy: Is this comment still relevant?

yelp_package/Makefile
```diff
@@ -77,7 +77,7 @@ clean:
...
-LAST_COMMIT_MSG = $(shell git log -1 --pretty=%B | sed -e "s/'/"/g")
+LAST_COMMIT_MSG = $(shell git log -1 --pretty=%B | sed -e 's/'\''/"/g')
```
solarkennedy: I'm really surprised this works! How is this second single quote working?

nosmo: The fact that the string "ends" with the second single quote means that the string concatenation is handled by the shell itself. It works, but it's nasty. I'm going to replace this mess with \x27 which is less immediately readable but less ugly.

PR #1076, opened 2017-03-10
Fix/autoscaling info

mattmb:
Fixup a couple of things to make the cluster autoscaling info thing work.

### DIFF
--- a/paasta_tools/cli/cmds/metastatus.py
+++ b/paasta_tools/cli/cmds/metastatus.py
@@ -64,6 +64,13 @@ def add_subparser(subparsers):
         help="define a different soa config directory",
     )
     status_parser.add_argument(
+        '-a', '--autoscaling-info',
+        action='store_true',
+        default=False,
+        dest="autoscaling_info",
+        help="Show cluster autoscaling info, implies -vv"
+    )
+    status_parser.add_argument(
         '-g',
         '--groupings',
         nargs='+',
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/utils.py
```diff
@@ -517,25 +517,29 @@ def execute_paasta_serviceinit_on_remote_master(subcommand, cluster, service, in
...
         timeout = 20
+    autoscaling_flag = "-a" if autoscaling_info else ""
+    if autoscaling_flag and verbose < 2:
+        verbose_flag = '-vv'
     humanize_flag = "-H" if humanize else ''
     groupings_flag = "-g %s" % " ".join(groupings) if groupings else ''
-    cmd_args = " ".join(filter(None, [verbose_flag, humanize_flag, groupings_flag]))
-    command = ('ssh -A -n -o StrictHostKeyChecking=no %s sudo paasta_metastatus %s' % (
+    cmd_args = " ".join(filter(None, [verbose_flag, humanize_flag, groupings_flag, autoscaling_flag]))
+    command = ('ssh -A -n -o StrictHostKeyChecking=no %s paasta_metastatus %s' % (
```
Rob-Johnson: are you sure that we're ok without sudo?

mattmb: I thought so too. But it works when I run `/usr/bin/paasta_metastatus -vvv` without the sudo on a master so I guess it only needs credentials to make updates to mesos?

mattmb:
@Rob-Johnson I've added the sudo back for now, seeing as we've sorted our problem with the path another way. Let me know if you think this is good to go.

paasta_tools/cli/cmds/metastatus.py
```diff
@@ -64,6 +64,13 @@ def add_subparser(subparsers):
         help="define a different soa config directory",
     )
     status_parser.add_argument(
+        '-a', '--autoscaling-info',
```
Rob-Johnson: any reason to make this an optional thing? I don't know why we'd ever want to not see it.

mattmb: Yep, it's a little slow and it breaks if the credentials aren't found/are invalid. But we could catch that exception really. Let's see how it goes and maybe make it default later.

PR #1077, opened 2017-03-11
Adding rocket support for local runs.

mcarlson-zero:


solarkennedy:
@mcarlson-zero ( @matthewmcarlson ?) cool!
Now realistically we don't have plans to support rocket for real, so we won't merge this in as-is to avoid the maintenance overhead and copy/paste.

Do you intend to push this further or maybe just save it for when we have systemd and can support this for real?

solarkennedy:
per our conversation this isn't really "ready" right now, but when we get further along we can use this code for rkt.

PR #1079, opened 2017-03-13
Fix paasta local-run to support array commands

drolando:
@bukzor This should fix your issue. I tried running your itest command with this fixed and it worked.

The new test failed without the fixed since the code was generating a command like:
`[u'sh', u'-c', [u'sh', u'-c', u'make test']]`

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -389,7 +389,10 @@ def get_docker_run_cmd(memory, random_port, container_name, volumes, env, intera
         cmd.extend((
             'sh', '-c'
         ))
-        cmd.append(command)
+        if isinstance(command, list):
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -389,7 +389,10 @@ def get_docker_run_cmd(memory, random_port, container_name, volumes, env, intera
         cmd.extend((
             'sh', '-c'
         ))
-        cmd.append(command)
+        if isinstance(command, list):
+            cmd.extend(command)
+        else:
+            cmd.append(command)
```
bukzor: if command:
    if isinstance(command, six.string_types):
        cmd.extend(('sh', '-c', command))
    else:
        cmd.extend(command)

PR #1081, opened 2017-03-13
Change words for desired state in marathon_serviceinit

solarkennedy:
People were interpreting this with the left hand side being the "existing" state (configured) and the right hand side as the "desired" state (start/stopped) and thinking that a mistmatch meant that it was wrong.


nhandler:
@EvanKrall proposed just showing `Started` instead of `Configured`. I think adding some notes on ReadTheDocs that we can link people to for education would also help.

solarkennedy:
Well, we used to have it as "Started", but then people where more confused when they ran `paasta start` and saw the status as: `Started - Desired State: Stopped`

"Configured" is accurate in the sense that it is configured in marathon (even if that means started or stopped)

PR #1084, opened 2017-03-15
Bump simplejson to 3.10.0

oktopuz:


solarkennedy:
Shipit, but is there a reason why?

PR #1085, opened 2017-03-15
Don't specify path to Python 3 executable in tox.ini

sjaensch:
MacOS doesn't have `/usr/bin/python3`, let's just use `python3` instead. The only reason we specify an interpreter executable with a full path is because we don't want to get into trouble if a virtual environment is already active. Since the one activated for Paasta in the Makefile is a Python 2 one we shouldn't run into issues.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,7 +16,7 @@ commands =
     py.test {posargs:tests}
 
 [testenv:py3]
-basepython = /usr/bin/python3
+basepython = python3
 commands =
     # TODO: actually upgrade protobuf once this is resolved
     pip install protobuf==3.2.0
@@ -119,7 +119,7 @@ commands =
     behave {posargs}
 
 [testenv:general_itests_py3]
-basepython = /usr/bin/python3
+basepython = python3
 setenv = {[testenv:general_itests]setenv}
 changedir = {[testenv:general_itests]changedir}
 passenv = {[testenv:general_itests]passenv}


solarkennedy:
Thanks!

asottile:
This makes it so you cannot recreate the virtualenv while activated :(

sjaensch:
I think that's an OK trade-off, I'd consider being able to run the tests on
macOS more important. It's a tox-internal virtualenv in a hidden directory,
if you activate it manually it's not too much to ask to deactivate it
before recreating.

asottile:
This breaks _my_ workflow, couldn't you just symlink /usr/bin/python3 /usr/bin/python3.5 on osx so it satisfies PEP394?

keymone:
@asottile when i do that, i'm getting:

    ERROR: The executable /Users/xxx/dev/yelp/paasta/.tox/py3/bin/python3.6 is not functioning
    ERROR: It thinks sys.prefix is '/usr/local/Cellar/python3/3.6.0/Frameworks/Python.framework/Versions/3.6' (should be '/Users/xxx/dev/yelp/paasta/.tox/py3')
    ERROR: virtualenv is not compatible with this system or executable

do you have any ideas how to solve that? context: i've installed python3 via brew and it placed the binary at /usr/local/bin/python3, so i symlinked that into /usr/bin.

asottile:
osx virtualenv does not handle brew pythons (they're built in a somewhat-non-standard way due to the way brew works).  I'd suggest using the official installers instead: https://www.python.org/downloads/mac-osx/

PR #1086, opened 2017-03-15
cache base image on docker hub

keymone:
currently every itest rebuilds dockerfiles/{trusty,xenial} from scratch, this caches the image on dockerhub using sha256 of dockerfile as a tag.


keymone:
this brings total itests run time to under an hour, down from ~1h20m

mattmb:
@keymone nice 👍 , happy with anything that can speed the build up!

Can we setup a group dockerhub account? I seem to remember having one somewhere else I worked. Just thinking that we'll occasionally need to update this image so we all need to have access.

keymone:
@mattmb yeah, that was the idea, i only used personal for testing. and yes, the downside of this approach is that we'll have to bump FROM instruction every time we modify paasta-*-base images and push these images to the repo. i'll add some reviewers to see what people think.

Rob-Johnson:
sure, shipit.

yelp_package/Makefile
```diff
@@ -32,7 +32,16 @@ else
...
-build_%_docker:
+build_%_base:
+	set -x && \
+	cd dockerfiles/$*/ && \
+	export from=`grep ^FROM Dockerfile | cut -d' ' -f2` && \
+	[ -d ../../dockerfiles/paasta-$*-base/ ] && \
+	cd ../../dockerfiles/paasta-$*-base/ && \
+	docker pull $$from || docker build -t $$from .
+
+.SECONDEXPANSION:
```
EvanKrall: why is `.SECONDEXPANSION` needed here? Seems like this rule should work normally.

yelp_package/dockerfiles/trusty/Dockerfile
```diff
@@ -12,51 +12,7 @@
...
-# a system mesos package.
-RUN apt-get install -yq mesos=1.0.1-2.0.93.ubuntu1404 && \
-    cd /usr/lib/python2.7/site-packages && \
-	zip -r /root/mesos.native-1.0.1-cp27-none-linux_x86_64.whl mesos/native mesos.native-1.0.1.dist-info && \
-	zip -r /root/mesos.executor-1.0.1-cp27-none-linux_x86_64.whl mesos/executor mesos.executor-1.0.1.dist-info && \
-	zip -r /root/mesos.scheduler-1.0.1-cp27-none-linux_x86_64.whl mesos/scheduler mesos.scheduler-1.0.1.dist-info && \
-	apt-get remove -yq mesos
-
-RUN pip install --upgrade virtualenv==15.1.0
+FROM keymone/paasta-trusty-base:f291ab4b72d6a25ef6f3b60022bfe9575e99e0426040cf90e1311b773e78a44f
```
EvanKrall: we should probably make this under the Yelp namespace?

EvanKrall: ah, this was addressed above.

yelp_package/Makefile
keymone: my Makefile game is very weak, `build_%_docker: build_$$*_base ` <- this wasn't expanding without secondexpansion flag, i'd be happy to know what am i doing wrong there.

keymone:
@EvanKrall, you're right, that would be simpler, i'll try that approach

EvanKrall:
(looks like the failing tests aren't the ones using docker, so maybe a rebuild / merge of master is all that's needed?)

Issue #1087, opened 2017-03-15
paasta remote run needs to support constraints


Rob-Johnson:
we need to be able to add constraints to the tasks to be run by remote run - specifically, ``LIKE`` ``UNLIKE`` ``EQUALS``. Then we can support ``deploy_whitelist``  & ``deploy_blacklist`` which are needed.

EvanKrall:
Another option:

Instead of taking `deploy_whitelist` / `deploy_blacklist` and turning it into a Marathon/Chronos-like constraint language, and then parsing that, we could take an approach like in https://github.com/Yelp/paasta/pull/1069/files -- write functions `offer_matches_deploy_blacklist` and `offer_matches_deploy_whitelist`, and then check those in `start_task`.

Eventually, though, I think we'll want to implement MAX_PER constraints (at the very least, MAX_PER host, but we really should have MAX_PER habitat, etc.), and I don't know whether we'd want to describe those with the same language that Marathon uses, or add a configuration option for it.

keymone:
there is a problem with implementing MAX_PER for AdhocScheduler, as that only runs single task but somebody could potentially launch multiple adhoc jobs - we can't communicate that across frameworks.

keymone:
done in 1090

PR #1090, opened 2017-03-16
Remote run constraints support

keymone:
constraints for #1087 

this supports EQUALS, LIKE, UNLIKE, MAX_PER and UNIQUE

in CLI this adds repeatable adrgument to remote-run: `--cons <attr>,<OP>[,<val>]`, in paasta_remote_run, `--constraints-json`

all constraint-related code is in paasta_tools/frameworks/constraints.py, in native scheduler we have `self.constraints` read from configs or cli (cli has precedence), `self.constraint_state` and `self.constraint_state_lock` for stateful constraints (only max_per for now), state is swapped every time call to `driver.acceptOffers` succeeds. state is locked around processing each offer and in statusUpdate when processing dead tasks.

### DIFF
--- a/paasta_tools/cli/cmds/remote_run.py
+++ b/paasta_tools/cli/cmds/remote_run.py
@@ -109,20 +109,11 @@ def add_subparser(subparsers):
         default=False,
     )
     list_parser.add_argument(
-        '-E', '--cons-eq',
-        help='Mesos EQUALS constraint, --cons-eq <attr>,<value>',
-        required=False,
-        action='append'
-    )
-    list_parser.add_argument(
-        '-L', '--cons-like',
-        help='Mesos LIKE constraint, --cons-like <attr>,<value>',
-        required=False,
-        action='append'
-    )
-    list_parser.add_argument(
-        '-U', '--cons-unlike',
-        help='Mesos UNLIKE constraint, --cons-unlike <attr>,<value>',
+        '-X', '--cons',
+        help='Constraint option, format: <attr>,OP[,<value>], OP can be one of '
+        'the following: EQUALS matches attribute value exactly, LIKE and '
+        'UNLIKE match on regular expression, MAX_PER constrains number of '
+        'tasks per attribute value, UNIQUE is the same as MAX_PER,1',
         required=False,
         action='append'
     )
@@ -172,14 +163,9 @@ def paasta_remote_run(args):
         elif not isinstance(value, bool):
             cmd_parts.extend(['--%s' % arg_key, value])
 
-    constraints = []
-    for k, v in {'eq': 'EQUALS', 'like': 'LIKE', 'unlike': 'UNLIKE'}.items():
-        constraints.extend(
-            map(lambda e: e.replace(',', ',%s,' % v).split(',', 2),
-                args_vars['cons_%s' % k] or []))
-
+    constraints = map(lambda x: x.split(',', 2), args_vars['cons'])
     if len(constraints) > 0:
-        cmd_parts.extend(['--constraints', quote(json.dumps(constraints))])
+        cmd_parts.extend(['--constraints-json', quote(json.dumps(constraints))])
 
     paasta_print('Running on master: %s' % ' '.join(cmd_parts))
     return_code, status = run_on_master(
--- a/paasta_tools/cli/cmds/remote_run.py
+++ b/paasta_tools/cli/cmds/remote_run.py
@@ -163,7 +163,7 @@ def paasta_remote_run(args):
         elif not isinstance(value, bool):
             cmd_parts.extend(['--%s' % arg_key, value])
 
-    constraints = map(lambda x: x.split(',', 2), args_vars['cons'])
+    constraints = [x.split(',', 2) for x in args_vars['cons']]
     if len(constraints) > 0:
         cmd_parts.extend(['--constraints-json', quote(json.dumps(constraints))])
 
--- a/paasta_tools/cli/cmds/remote_run.py
+++ b/paasta_tools/cli/cmds/remote_run.py
@@ -109,7 +109,7 @@ def add_subparser(subparsers):
         default=False,
     )
     list_parser.add_argument(
-        '-X', '--cons',
+        '-X', '--constraint',
         help='Constraint option, format: <attr>,OP[,<value>], OP can be one of '
         'the following: EQUALS matches attribute value exactly, LIKE and '
         'UNLIKE match on regular expression, MAX_PER constrains number of '
@@ -163,7 +163,7 @@ def paasta_remote_run(args):
         elif not isinstance(value, bool):
             cmd_parts.extend(['--%s' % arg_key, value])
 
-    constraints = [x.split(',', 2) for x in args_vars['cons']]
+    constraints = [x.split(',', 2) for x in args_vars['constraint']]
     if len(constraints) > 0:
         cmd_parts.extend(['--constraints-json', quote(json.dumps(constraints))])
 
--- a/paasta_tools/cli/cmds/remote_run.py
+++ b/paasta_tools/cli/cmds/remote_run.py
@@ -15,7 +15,12 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import json
 import re
+try:
+    from shlex import quote
+except ImportError:
+    from pipes import quote
 
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_clusters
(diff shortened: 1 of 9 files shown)


keymone:
what do we set some limit on number of offers we can reject before terminating the framework? right now if none of the offers match constraints - it seems mesos will keep resending them indefinitely.

Rob-Johnson:
@keymone you can ask mesos not to send you the same offer again by providing ``filters`` to the ``declineOffers`` call. maybe use that + a timeout of a minute or something?

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -149,7 +163,11 @@ def paasta_remote_run(args):
         elif not isinstance(value, bool):
             cmd_parts.extend(['--%s' % arg_key, value])
 
-    paasta_print('Running on master: %s' % cmd_parts)
+    constraints = map(lambda x: x.split(',', 2), args_vars['cons'])
```
Rob-Johnson: sorry. ``[x.split(',',2) for x in args_vars['cons']]``

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -125,22 +140,35 @@ def launch_tasks_for_offers(self, driver, offers):
...
-
-                operation = mesos_pb2.Offer.Operation()
-                operation.type = mesos_pb2.Offer.Operation.LAUNCH
-                operation.launch.task_infos.extend(tasks)
-                driver.acceptOffers([offer.id], [operation])
-
-                launched_tasks.extend(tasks)
-            else:
-                driver.declineOffer(offer.id)
+            self.constraint_state_lock.acquire()
```
Rob-Johnson: im fairly sure you can use the lock as a context manager - you can forget about explicitly dropping the lock that way.

```python
with self.constraint_state_lock.acquire():
     ...
```

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -125,22 +140,35 @@ def launch_tasks_for_offers(self, driver, offers):
...
+                                health=None,
+                                mesos_task_state=TASK_STAGING,
+                                offer=offer))
+
+                    launched_tasks.extend(tasks)
+                    self.constraint_state = new_state
+                else:
+                    driver.declineOffer(offer.id)
+            except ConstraintFailAllTasksError:
+                paasta_print("Offer failed constraints for every task, rejecting 60s")
```
Rob-Johnson: :+1:

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -78,7 +88,7 @@ def __init__(self, service_name, instance_name, cluster,
                  system_paasta_config, soa_dir=DEFAULT_SOA_DIR,
                  service_config=None, reconcile_backoff=30,
                  instance_type='paasta_native', service_config_overrides={},
-                 reconcile_start_time=float('inf')):
+                 reconcile_start_time=float('inf'), constraints=[]):
```
Rob-Johnson: mutable default args :( 

```
>>> class X:
...    def __init__(self, foo=[]):
...        self.foo = foo
...        print foo
... 
>>> x1 = X()
[]
>>> x1.foo.append('y')
>>> x1.foo
['y']
>>> x2 = X()
['y']
>>> x2.foo.append('foo')
>>> x2.foo
['y', 'foo']
```

you'll have to set the default arg as ``None`` and do ``if not constraints: self.constraints = []``

keymone: crap, i just had a conversation about this couple days ago :) ruby habits will be dying slowly. thanks!

paasta_tools/frameworks/native_scheduler.py
keymone: nice, thanks

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -240,7 +278,15 @@ def tasks_for_offer(self, driver, offer):
             remainingMem -= task_mem
             remainingPorts -= set([task_port])
 
-        return tasks
+        # raise constraint error but only if no other tasks fit/fail the offer
+        if total > 0 and failed_constraints == total:
+            raise ConstraintFailAllTasksError
+
+        if len(tasks) > 0:
+            update_constraint_state(offer, self.constraints,
```
EvanKrall: Won't we need to update our constraint state (or a temporary copy) every time we add a task to `tasks`? As written, I think this method could end up creating multiple tasks on the same offer and violating a UNIQUE constraint.

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -125,22 +139,34 @@ def launch_tasks_for_offers(self, driver, offers):
...
-
-                operation = mesos_pb2.Offer.Operation()
-                operation.type = mesos_pb2.Offer.Operation.LAUNCH
-                operation.launch.task_infos.extend(tasks)
-                driver.acceptOffers([offer.id], [operation])
-
-                launched_tasks.extend(tasks)
-            else:
-                driver.declineOffer(offer.id)
+            with self.constraint_state_lock:
```
EvanKrall: Why do we need a lock here? I thought we only had one thread per scheduler.

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -400,6 +450,10 @@ def recreate_drain_method(self):
             **self.service_config.get_drain_method_params(self.service_config.service_namespace_config)
         )
 
+    def reload_constraints(self):
+        if len(self.constraints) == 0 and self.service_config.get_constraints():
+            self.constraints = self.service_config.get_constraints()
```
EvanKrall: Why copy the config's constraints into self?

Also, won't the `len(self.constraints) == 0` conditional prevent us from ever changing constraints?

paasta_tools/paasta_remote_run.py
```diff
@@ -87,7 +99,8 @@ def main(argv):
         soa_dir=soa_dir,
         reconcile_backoff=0,
         dry_run=dry_run,
-        service_config_overrides={'cmd': command}
+        service_config_overrides={'cmd': command},
+        constraints=constraints
```
EvanKrall: can we pass constraints in as part of `service_config_overrides`?

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -103,6 +108,15 @@ def add_subparser(subparsers):
         required=False,
         default=False,
     )
+    list_parser.add_argument(
+        '-X', '--cons',
```
keymone: cool, didn't know about that

paasta_tools/frameworks/constraints.py
```diff
@@ -0,0 +1,84 @@
...
+    returns False otherwise. Prints a error message and re-raises if an error
+    was thrown."""
+    for (attr, op, val) in constraints:
+        try:
+            offer_attr = next(
+                (x for x in offer.attributes if x.name == attr), None)
+            if offer_attr is None:
+                paasta_print("Attribute not found for a constraint: %s" % attr)
+                return False
+            elif not(CONS_OPS[op]([val, offer_attr.text.value,
```
keymone: argh, /me dumb. i thought lambda only supports single argument, probably because it only supports single statement. thanks

paasta_tools/frameworks/native_scheduler.py
keymone: i don't know threading properties of mesos driver, if we're 100% sure no multiple statusUpdate/resourceOffer calls can be invoked in parallel we can drop the lock, for now it's a case of "better safe than sorry". let me know if you're certain we can remove the lock.

keymone:
@EvanKrall the problem with using tasks_with_flags for constraint state is that we don't modify it until we actually launch tasks, but we need to be updating constraint state for each task that fits in an offer. i felt like this would get hairy very quickly and decided to keep the state separate and copy+mutate+swap for each new offer if it results in launched tasks.

also we i've added offer reference to each task in tasks_with_flags to be able to update constraint state when task goes away.

thanks for review, i think i've addressed all the points but the locking one. i'm totally fine with dropping locks if we're sure it won't bite us later on.

keymone:
@EvanKrall @Rob-Johnson is this ready for shiping?

paasta_tools/frameworks/constraints.py
```diff
@@ -0,0 +1,80 @@
...
+# Without this, the import of mesos.interface breaks because paasta_tools.mesos exists
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import re
+
+from paasta_tools.utils import paasta_print
+
+
+def max_per(cv, ov, at, st):
```
Rob-Johnson: can you write the names of these args out in full to make it a bit clearer?

PR #1093, opened 2017-03-16
RFC: Added service and instance docker labels to everything

solarkennedy:
RFC: I would like to provide cloudhealth with enough metadata that we can create curated projects and stuff. Using mesos's api to query this data makes sense, but I don't want them to even look at the mesos task id at all. Not only do we not directly control it, but often it will just be lies.

With this change, I add service and instance as **docker labels**, which can be read by cloudhealth via mesos, regardless of what framework we decide to use to launch containers.

Once this metadata is in here, we can consume it in cloudhealth without every developer having to add additional configuration in their services.

cc @bchess

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -216,9 +216,13 @@ class InstanceConfig(dict):
         Note: values must be strings
 
         :returns: A list of parameters to be added to docker run"""
-        parameters = [{"key": "memory-swap", "value": self.get_mem_swap()},
-                      {"key": "cpu-period", "value": "%s" % int(self.get_cpu_period())},
-                      {"key": "cpu-quota", "value": "%s" % int(self.get_cpu_quota())}]
+        parameters = [
+            {"key": "memory-swap", "value": self.get_mem_swap()},
+            {"key": "cpu-period", "value": "%s" % int(self.get_cpu_period())},
+            {"key": "cpu-quota", "value": "%s" % int(self.get_cpu_quota())},
+            {"key": "label", "value": "service=%s" % self.service()},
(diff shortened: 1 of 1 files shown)


solarkennedy:
And yes, this will cause a big bounce :(

paasta_tools/utils.py
```diff
@@ -216,9 +216,13 @@ def format_docker_parameters(self):
         Note: values must be strings
 
         :returns: A list of parameters to be added to docker run"""
-        parameters = [{"key": "memory-swap", "value": self.get_mem_swap()},
-                      {"key": "cpu-period", "value": "%s" % int(self.get_cpu_period())},
-                      {"key": "cpu-quota", "value": "%s" % int(self.get_cpu_quota())}]
+        parameters = [
+            {"key": "memory-swap", "value": self.get_mem_swap()},
+            {"key": "cpu-period", "value": "%s" % int(self.get_cpu_period())},
```
nhandler: Why not use `%d` instead of `int()`?

nhandler:
Fix the tests that are failing, but in principle, I am fine with adding some new labels for this information.

paasta_tools/utils.py
```diff
@@ -216,9 +216,13 @@ def format_docker_parameters(self):
...
+            {"key": "cpu-quota", "value": "%s" % int(self.get_cpu_quota())},
+            {"key": "label", "value": "service=%s" % self.service()},
+            {"key": "label", "value": "instance=%s" % self.service()},
```
solarkennedy: parameters.append(self.get_docker_params())

PR #1097, opened 2017-03-17
Revert "Don't specify path to Python 3 executable in tox.ini"

keymone:
Reverts Yelp/paasta#1085

brew is not recommended way to install python apparently


mattmb:
Out of interest what is the "recommended way"? I used brew and hit the same problem...

keymone:
@mattmb apparently it's to install official python binary from https://www.python.org/downloads/

PR #1098, opened 2017-03-17
OMG pyupgrade precommit

solarkennedy:
cc @asottile 

I'm concerned about the tests that were using `u'foo'` for force an explicit unicode literal. Should I revert that?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,3 +51,7 @@
         entry: except:$
         language: pcre
         files: \.py$
+-   repo: https://github.com/asottile/pyupgrade
+    sha: v1.0.0
+    hooks:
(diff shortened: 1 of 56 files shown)


asottile:
pyupgrade will only do safe changes, if a file does `from __future__ import unicode_literals` `u''` is identical to `''`

.pre-commit-config.yaml
```diff
@@ -51,3 +51,7 @@
         entry: except:$
         language: pcre
         files: \.py$
+-   repo: https://github.com/asottile/pyupgrade
+    sha: v1.0.0
```
asottile: I'd suggest v1.0.4, there were a few derps in the earlier versions

asottile:
Good work team!

PR #1102, opened 2017-03-20
gevent proof of concept for the service autoscaling

oktopuz:
A very quick optimization of the mesos cpu metrics provider to get stats asynchronously.
It shows 12% improvement in dev, which is a quite good result because dev services don't have enough instances to leverage gevent.

The next steps will be:
1. to support all provider
2. to query each mesos slave only once
3. will not query /slave(1)/state.json on every slave when collecting stats, because we probably don't need it.
3. (optional) to upgrade gevent

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,4 +1,5 @@
 [run]
+concurrency = gevent
 branch = True
 source = .
 omit =
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -22,7 +22,9 @@ from datetime import datetime
 from math import ceil
 from math import floor
 
+import gevent
 import requests
+from gevent import monkey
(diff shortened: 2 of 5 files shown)


EvanKrall:
@jolynch can you take a look?

mattmb:
This is cool, it hadn't really occurred to me that this would be the slow part when running the autoscaler. But it makes sense that it would be for services with a large number of tasks.

The next provider to do is here: https://github.com/Yelp/paasta/blob/a8dc3c3c21d222666e44d9e410108c9884edfca3/paasta_tools/autoscaling/autoscaling_service_lib.py#L185

That would speed up the uwsgi stats provider which we want to use for autoscaling yelp-main.

oktopuz:
It turns out coverage doesn't work well with gevent without 'concurrency = gevent'.
Also I have to tell coverage to skip a couple of wait-for-deployment tests because 'concurrency = gevent' is incompatible with threading.

PR #1103, opened 2017-03-22
[wip] Timeout on task staging

### LINKED ISSUE #1091: native framework/remote-run should support a TASK_STAGING timeout
the same issue as in https://github.com/Yelp/chronos/issues/63 - make sure we don't just leave tasks forever hanging because of docker having a bad time on agents.

Rob-Johnson:
closes #1091

### DIFF
--- a/paasta_tools/cli/cmds/remote_run.py
+++ b/paasta_tools/cli/cmds/remote_run.py
@@ -85,6 +85,13 @@ def add_remote_run_args(parser):
         required=False,
         default=False,
     )
+    parser.add_argument(
+        '-t', '--staging-timeout',
+        help='A timeout for the task to be launching before killed',
+        required=False,
+        default=60,
+        type=float,
+    )
 
 
 def add_subparser(subparsers):
@@ -115,7 +122,8 @@ def add_subparser(subparsers):
         'UNLIKE match on regular expression, MAX_PER constrains number of '
         'tasks per attribute value, UNIQUE is the same as MAX_PER,1',
         required=False,
-        action='append'
+        action='append',
+        default=[]
     )
     list_parser.set_defaults(command=paasta_remote_run)
 
@@ -143,7 +151,8 @@ def paasta_remote_run(args):
         'json_dict': False,
         'cmd': None,
         'verbose': True,
-        'dry_run': False
+        'dry_run': False,
+        'staging_timeout': None,
     }
     for key in args_vars:
         # skip args we don't know about
@@ -161,9 +170,9 @@ def paasta_remote_run(args):
         if isinstance(value, bool) and value:
             cmd_parts.append('--%s' % arg_key)
         elif not isinstance(value, bool):
-            cmd_parts.extend(['--%s' % arg_key, value])
+            cmd_parts.extend(['--%s' % arg_key, str(value)])
 
-    constraints = [x.split(',', 2) for x in args_vars['constraint']]
+    constraints = [x.split(',', 2) for x in args_vars.get('constraint', [])]
     if len(constraints) > 0:
         cmd_parts.extend(['--constraints-json', quote(json.dumps(constraints))])
 
--- a/paasta_tools/frameworks/native_scheduler.py
+++ b/paasta_tools/frameworks/native_scheduler.py
@@ -10,6 +10,7 @@ import random
 import threading
 import time
 import uuid
+from threading import Timer
 
 import mesos.interface
 import service_configuration_lib
@@ -72,6 +73,7 @@ class MesosTaskParameters(object):
         mesos_task_state=None,
         is_draining=False,
         is_healthy=False,
+        staging_timer=None,
         offer=None
     ):
         self.health = health
@@ -81,11 +83,12 @@ class MesosTaskParameters(object):
         self.is_healthy = is_healthy
         self.offer = offer
         self.marked_for_gc = False
+        self.staging_timer = staging_timer
 
 
 class NativeScheduler(mesos.interface.Scheduler):
     def __init__(self, service_name, instance_name, cluster,
-                 system_paasta_config, soa_dir=DEFAULT_SOA_DIR,
+                 system_paasta_config, staging_timeout, soa_dir=DEFAULT_SOA_DIR,
                  service_config=None, reconcile_backoff=30,
                  instance_type='paasta_native', service_config_overrides=None,
                  reconcile_start_time=float('inf')):
@@ -105,6 +108,9 @@ class NativeScheduler(mesos.interface.Scheduler):
 
         # wait this long after starting a reconcile before accepting offers.
         self.reconcile_backoff = reconcile_backoff
+
+        # wait this long for a task to launch.
+        self.staging_timeout = staging_timeout
 
         # Gets set when registered() is called
         self.framework_id = None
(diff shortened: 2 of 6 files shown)


paasta_tools/frameworks/native_scheduler.py
```diff
@@ -274,6 +296,10 @@ def statusUpdate(self, driver, update):
             if params.marked_for_gc:
                 self.tasks_with_flags.pop(task)
 
+        if task_params.mesos_task_state is not TASK_STAGING:
+            self.starting_tasks[task_id].cancel()
+            self.starting_tasks = {k: v for k, v in self.starting_tasks if k is not task_id}
```
keymone: isn't this just `del self.starting_tasks[task_id]` ?

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -190,6 +200,18 @@ def get_old_tasks(self, name, tasks):
     def is_task_new(self, name, tid):
         return tid.startswith("%s." % name)
 
+    def killer_for_task(self, driver, taskId):
+        """Returns a closure to kill a task
+        """
+        def kill_task():
+            kill_task(driver, taskId)
```
keymone: this confuses me because it looks like recursion. is there anything wrong with using lambda on line 213?

    return Timer(timeout_value, lambda: self.kill_task(driver, taskId))

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -160,7 +170,7 @@ def task_fits(self, offer):
 
         return True
 
-    def need_more_tasks(self, name, existingTasks={}, scheduledTasks=[]):
```
keymone: thanks, i missed those

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -138,6 +143,11 @@ def launch_tasks_for_offers(self, driver, offers):
                 operation.launch.task_infos.extend(tasks)
                 driver.acceptOffers([offer.id], [operation])
 
+                for task in tasks:
+                    startup_timer = self.startup_timer_for_task(30, driver, task.task_id.value)
+                    self.starting_tasks[task.task_id.value] = startup_timer
```
keymone: alternatively we could store staging_timer in MesosTaskParameters 7 lines above? would avoid having to maintain separate dict.

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -145,7 +143,7 @@ def launch_tasks_for_offers(self, driver, offers):
 
                 for task in tasks:
                     startup_timer = self.startup_timer_for_task(30, driver, task.task_id.value)
-                    self.starting_tasks[task.task_id.value] = startup_timer
+                    self.tasks_with_flags[task.task_id.value].startup_timer = startup_timer
```
keymone: it's `.start_timer` in MesosTaskParameters

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -297,8 +286,8 @@ def statusUpdate(self, driver, update):
                 self.tasks_with_flags.pop(task)
 
         if task_params.mesos_task_state is not TASK_STAGING:
-            self.starting_tasks[task_id].cancel()
-            self.starting_tasks = {k: v for k, v in self.starting_tasks if k is not task_id}
+            self.tasks_with_flags[task_id].startup_timer.cancel()
```
keymone: maybe worth checking that timer still exists? it gets cancelled on statusUpdate about STAGING -> STARTING but then will fail on statusUpdate about RUNNING?

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -274,6 +292,12 @@ def statusUpdate(self, driver, update):
             if params.marked_for_gc:
                 self.tasks_with_flags.pop(task)
 
+        if task_params.mesos_task_state is not TASK_STAGING:
+            if self.tasks_with_flags[task_id].start_timer:
+                print('cancelling timer')
```
keymone: drop this before shipping

keymone:
@EvanKrall in adhoc scheduler for instance we don't run `periodic()` at all, so that would be an issue. what is the general purpose of periodic? could we move it from being called from outside to be a configurable (optional) ticker with callback inside scheduler?

EvanKrall:
The purpose of `periodic` within the native long-running scheduler is to reload the configs, in case they change; we also consider whether any tasks need killing. (e.g. if we drained a task and now it is safe to kill).

The way `periodic` is called right now is pretty hacky, and I'd be happy to see it replaced with something better.

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -161,9 +169,9 @@ def paasta_remote_run(args):
         if isinstance(value, bool) and value:
             cmd_parts.append('--%s' % arg_key)
         elif not isinstance(value, bool):
-            cmd_parts.extend(['--%s' % arg_key, value])
+            cmd_parts.extend(['--%s' % arg_key, str(value)])
 
-    constraints = [x.split(',', 2) for x in args_vars['constraint']]
+    constraints = [x.split(',', 2) for x in args_vars.get('constraint', [])]
```
keymone: is this really needed if the default is specified in add_argument?

paasta_tools/frameworks/adhoc_scheduler.py
```diff
@@ -27,6 +27,8 @@ def __init__(self, *args, **kwargs):
         if kwargs.get('service_config_overrides') is None:
             kwargs['service_config_overrides'] = {}
         kwargs['service_config_overrides']['instances'] = 1
+        if kwargs.get('staging_timeout') is None:
+            kwargs['staging_timeout'] = 1
```
keymone: seems too low for a default? also there is another default in this PR - 60s, why are these values different?

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -150,12 +156,20 @@ def launch_tasks_for_offers(self, driver, offers):
                         driver.acceptOffers([offer.id], [operation])
 
                         for task in tasks:
+                            startup_timer = self.startup_timer_for_task(
```
keymone: it's a nitpick but naming inconsistencies annoy me, we currently have: staging_timeout, start_timer and startup_timer. can we settle on one name? staging_timer / staging_timeout? startup_timer / startup_timeout?

paasta_tools/native_mesos_scheduler.py
```diff
@@ -21,6 +21,7 @@ def parse_args(argv):
     parser.add_argument('-d', '--soa-dir', dest="soa_dir", metavar="SOA_DIR", default=DEFAULT_SOA_DIR)
     parser.add_argument('--stay-alive-seconds', dest="stay_alive_seconds", type=int, default=300)
     parser.add_argument('--periodic-interval', dest="periodic_interval", type=int, default=30)
+    parser.add_argument('--staging-timeout', dest="staging_timeout", type=int, default=3330)
```
keymone: why is default so high here?

paasta_tools/native_mesos_scheduler.py
```diff
@@ -37,6 +38,7 @@ def main(argv):
             service_name=service,
             instance_name=instance,
             cluster=cluster,
+            staging_timeout=args.staging_timeout,
```
keymone: either we need float() here or we don't need it at line 112?

Rob-Johnson: here we set the type in the argparser of ``native_mesos_scheduler``. we didn't do it for the ``remote_run`` parser - I should add it there though so I can remove the float cast.

paasta_tools/cli/cmds/remote_run.py
Rob-Johnson: eh, maybe not, but good to be safe?

Rob-Johnson:
@keymone addressed feedback

paasta_tools/cli/cmds/remote_run.py
keymone: yeah, ok

PR #1113, opened 2017-03-24
async version of uwsgi and http providers

oktopuz:
async version of uwsgi and http providers for the service autoscaler.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -25,6 +25,7 @@ from math import floor
 import gevent
 import requests
 from gevent import monkey
+from gevent import pool
 from kazoo.client import KazooClient
 from kazoo.exceptions import NoNodeError
 
@@ -159,10 +160,10 @@ def pid_decision_policy(zookeeper_path, current_instances, min_instances, max_in
     return int(round(clamp_value(Kp * error + iterm + Kd * (error - last_error) / time_delta)))
 
 
-def get_json_body_from_service(host, port, endpoint):
+def get_json_body_from_service(host, port, endpoint, timeout=2):
     return requests.get(
         'http://%s:%s/%s' % (host, port, endpoint),
-        headers={'User-Agent': get_user_agent()},
+        headers={'User-Agent': get_user_agent()}, timeout=timeout
     ).json()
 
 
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -182,19 +183,28 @@ def get_http_utilization_for_all_tasks(marathon_service_config, marathon_tasks,
...
             # This won't trigger in the event of DNS error or when a request is refused
             # a requests.exception.ConnectionError is raised in those cases
             utilization.append(1.0)
             log.debug("Recieved a timeout when querying %s on %s:%s. Assuming the service is at full utilization." % (
                 marathon_service_config.get_service(), task.host, task.ports[0]))
-        except Exception as e:
-            log.debug("Caught exception when querying %s on %s:%s : %s" % (
-                marathon_service_config.get_service(), task.host, task.ports[0], str(e)))
+        elif job.exception is not None:
+            log.debug("Caught exception when querying %s on %s:%s : %s" % (marathon_service_config.get_service(),
```
mattmb: I'd probably up this from debug, it is a real error even if it's not that critical.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -182,19 +183,28 @@ def get_http_utilization_for_all_tasks(marathon_service_config, marathon_tasks,
...
+            for task in marathon_tasks]
+    gevent.joinall(jobs)
+
+    for job in jobs:
+        if job.value is not None:
+            utilization.append(job.value)
+        elif isinstance(job.exception, requests.exceptions.Timeout):
             # If we time out querying an endpoint, assume the task is fully loaded
             # This won't trigger in the event of DNS error or when a request is refused
             # a requests.exception.ConnectionError is raised in those cases
```
mattmb: There's a subtle change here that means this comment isn't true now if I'm reading it right?

i.e. now if there's any exception querying a task we assume it to be "fully loaded". That might actually be reasonable although I think just dropping it from the list is also reasonable given that we are taking an average.

I guess I'm happy either way... but please update the comment if you want to keep the code like this.

oktopuz: Unless I missed something, I didn't change the way it worked.
A task should be considered as "fully loaded" when the job terminated with requests.exceptions.Timeout (tested with https://github.com/Yelp/paasta/blob/master/tests/autoscaling/test_autoscaling_service_lib.py#L312).
Any other exceptions should go to the line 203 (tested with https://github.com/Yelp/paasta/blob/master/tests/autoscaling/test_autoscaling_service_lib.py#L325).

Therefore the comment looks correct to me.

paasta_tools/autoscaling/autoscaling_service_lib.py
oktopuz: replaced with log.error.

paasta_tools/autoscaling/autoscaling_service_lib.py
mattmb: My bad, you're quite right! The behaviour is unchanged 👍

mattmb:
Some test failures to deal with but otherwise this LGTM 😄

oktopuz:
@mattmb, I had to redesign this a bit to keep the log messages in the same format.
Could you check it once again please?

mattmb:
LGTM 👍

PR #1115, opened 2017-03-28
Only update deployments.json if changed

mattmb:
This is a precursor to the work on paasta-deployd. It means that
deployments.json files will only be written to if they are being
updated. This makes it possible to watch for inotify updates to that
file and perform a bounce based on inotify events.

Note that this will probably change the behaviour of:
https://github.com/Yelp/paasta/blob/master/paasta_tools/generate_all_deployments#L9

In fact, we definitely need to do something or it will be deleting things that haven't changed in a while. Do we really need that cleanup? I guess if a folder removes all it's marathon-* files then we need to remove the deployents.json...

### DIFF
--- a/paasta_tools/generate_all_deployments
+++ b/paasta_tools/generate_all_deployments
@@ -1,19 +1,7 @@
 #!/bin/bash
 #
 # Generates all the per-service deployments.json files
-# and cleans up any leftovers
 #
 
-function delete_old_deployments {
-    # Delete any deployments.json file that is older than an hour
-    for service in $(comm -2 -3 <(find /nail/etc/services/ -maxdepth 1 -type d -printf %f\\n|tail -n +2|sort) <(paasta list|sort)); do
-        if [ -f /nail/etc/services/$service/deployments.json ]; then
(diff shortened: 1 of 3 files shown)


mattmb:
I've amended my commit to take a new approach to tidying up old `deployment.json` files.

paasta_tools/generate_all_deployments
```diff
@@ -6,18 +6,14 @@
 
 function delete_old_deployments {
     # Delete any deployments.json file that is older than an hour
-    find /nail/etc/services -iname deployments.json  -mmin +60 -delete
+    for service in $(comm -2 -3 <(find /nail/etc/services/ -maxdepth 1 -type d -printf %f\\n|tail -n +2|sort) <(paasta list|sort)); do
```
solarkennedy: I dare *don't* do this at all. These files don't hurt (afaik)
It isn't worth the risk of getting this bash wrong

mattmb: I think that's a pretty good shout. It was my first instinct too. I guess even if a `deployments.json` is left over then we won't do anything with the service unless it appears in paasta list anyway.

PR #1120, opened 2017-03-29
[WIP] Paasta deployd

mattmb:
This is a start on paasta-deployd

I'm trying to avoid touching the existing code base as much as possible but this does make a small change to `setup_marathon_job` job to make the poc "work".

I'll probably do a bit more work on this branch before we need to think about merging it in but it would be great to get some feedback on the high level architecture! Please don't waste your time digging too deep though as there's still likely to be quite a bit of change going on.

### DIFF
--- a/paasta_tools/deployd/master.py
+++ b/paasta_tools/deployd/master.py
@@ -5,7 +5,6 @@ from __future__ import unicode_literals
 import inspect
 import logging
 import socket
-import sys
 import time
 
 from six.moves.queue import Empty
@@ -72,28 +71,30 @@ class Inbox(PaastaThread):
                 self.bounce_q.put(service_instance)
 
 
-class DeployDaemon(object):
+class DeployDaemon(PaastaThread):
     def __init__(self):
-        self.log.info("paasta-deployd starting up...")
+        PaastaThread.__init__(self)
+        PaastaThread.daemon = True
+        self.started = False
         self.bounce_q = PaastaQueue("BounceQueue")
         self.inbox_q = PaastaQueue("InboxQueue")
         self.control = PaastaQueue("ControlQueue")
         self.inbox = Inbox(self.inbox_q, self.bounce_q)
         self.config = load_system_paasta_config()
+
+    def run(self):
+        self.log.info("paasta-deployd starting up...")
         with ZookeeperPool() as zk:
             self.log.info("Waiting to become leader")
             self.election = PaastaLeaderElection(zk, "/paasta-deployd-leader", socket.getfqdn(), control=self.control)
-            self.election.run(self.start)
-
-    @property
-    def log(self):
-        name = '.'.join([__name__, self.__class__.__name__])
-        return logging.getLogger(name)
+            self.is_leader = False
+            self.election.run(self.startup)
 
     def bounce(self, service_instance):
         self.inbox_q.put(service_instance)
 
-    def start(self):
+    def startup(self):
+        self.is_leader = True
         self.log.debug("This node is elected as leader {}".format(socket.getfqdn()))
         self.inbox.start()
         self.log.info("Starting all watcher threads")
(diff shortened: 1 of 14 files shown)


paasta_tools/deployd/master.py
```diff
@@ -0,0 +1,172 @@
...
+        self.inbox_q = PaastaQueue("InboxQueue")
+        self.control = PaastaQueue("ControlQueue")
+        self.inbox = Inbox(self.inbox_q, self.bounce_q)
+        self.config = load_system_paasta_config()
+
+    def run(self):
+        self.log.info("paasta-deployd starting up...")
+        with ZookeeperPool() as zk:
+            self.log.info("Waiting to become leader")
+            self.election = PaastaLeaderElection(zk, "/paasta-deployd-leader", socket.getfqdn(), control=self.control)
```
solarkennedy: This needs to have the cluster name in it, it isn't safe to assume you have the whole zk to yourself

paasta_tools/deployd/master.py
```diff
@@ -0,0 +1,172 @@
...
+            self.to_bounce[service_instance_key] = service_instance
+
+    def should_add_to_bounce(self, service_instance, service_instance_key):
+        if service_instance_key in self.to_bounce:
+            if service_instance.bounce_by > self.to_bounce[service_instance_key].bounce_by:
+                self.log.debug("{} already in bounce q with higher priority".format(service_instance))
+                return False
+        return True
+
+    def process_to_bounce(self):
```
solarkennedy: Why the extra function here? I don't understand why it doesn't just put things in the bouce_q directly?

paasta_tools/deployd/master.py
mattmb: 👍

paasta_tools/deployd/master.py
mattmb: So the plan is that this gives us a way to do the "priority by time" thing and staggering of bounces.

The idea is *everything* adds to `inbox_q` and then we only move things to the real `bounce_q` if they have reached the time we want them to bounce at. We can also add logic so instances get pushed real queue if it's empty i.e. there's nothing of higher priority stopping us doing so.

paasta_tools/deployd/master.py
```diff
@@ -0,0 +1,172 @@
...
+        for service_instance_key in self.to_bounce.keys():
+            if self.to_bounce[service_instance_key].bounce_by < int(time.time()):
+                service_instance = self.to_bounce.pop(service_instance_key)
+                self.bounce_q.put(service_instance)
+
+
+class DeployDaemon(PaastaThread):
+    def __init__(self):
+        PaastaThread.__init__(self)
```
Rob-Johnson: why does this deviate from the 'normal' way of constructing the super class?

I'd expect ``super(DeployDaemon, self).__init__()``?

paasta_tools/deployd/master.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def start_watchers(self):
+        """ should block until all threads happy"""
+        watcher_classes = [obj[1] for obj in inspect.getmembers(watchers) if inspect.isclass(obj[1]) and
+                           obj[1].__bases__[0] == watchers.PaastaWatcher]
+        self.watchers = [watcher(inbox_q=self.inbox_q,
+                                 cluster=self.config.get_cluster())
+                         for watcher in watcher_classes]
+        self.log.info("Starting the following watchers {}".format(self.watchers))
+        [watcher.start() for watcher in self.watchers]
```
Rob-Johnson: do:

```
for watcher in self.watchers:
    watcher.start()
```

since you're really just modelling a side effect, rather than any sort of transformation.

paasta_tools/deployd/master.py
```diff
@@ -0,0 +1,172 @@
...
+                                           splay_minutes=30,
+                                           watcher_name='daemon_start')
+        for service_instance in instances_to_add:
+            self.inbox_q.put(service_instance)
+
+    def start_watchers(self):
+        """ should block until all threads happy"""
+        watcher_classes = [obj[1] for obj in inspect.getmembers(watchers) if inspect.isclass(obj[1]) and
+                           obj[1].__bases__[0] == watchers.PaastaWatcher]
+        self.watchers = [watcher(inbox_q=self.inbox_q,
```
Rob-Johnson: why do you make this a class level property? are there other things in the class that depend on this variable?

paasta_tools/deployd/master.py
```diff
@@ -0,0 +1,172 @@
...
+                message = None
+            if message == "ABORT":
+                break
+            time.sleep(0.1)
+
+    def stop(self):
+        self.control.put("ABORT")
+
+    def start_workers(self):
+        for i in range(5):
```
Rob-Johnson: not critical, but we'll want to make this value configurable

paasta_tools/deployd/master.py
mattmb: So it turns out that python `Threads` are "old style" classes so don't work with super. Given that this inherits from an intermediate maybe it will work? I'll give it a try

paasta_tools/deployd/master.py
mattmb: Good question, I guess I was pre-empting some need for it in the future but there probably isn't one...

solarkennedy:
Keep going and ship as you see fit to see how this behaves in the wild.

PR #1130, opened 2017-04-05
make it possible to change the containerPort for the docker port mapping

EvanKrall:


solarkennedy:
Can you sprinkle in a bit of some doc changes and double check the tests?

solarkennedy:
The docs are a nice to have, here is where we reference 8888
http://paasta.readthedocs.io/en/latest/search.html?q=8888&check_keywords=yes&area=default

Ship!

solarkennedy:
rebase

PR #1134, opened 2017-04-11
Security check v1

transcedentalia:


paasta_tools/utils.py
```diff
@@ -1020,6 +1020,13 @@ def get_expected_slave_attributes(self):
         calculating the default routing constraints."""
         return self.get('expected_slave_attributes')
 
+    def get_security_check_command(self):
+        """Get the script to be executed during the security-check build step
+
+        :return: The name of the file
+        """
+        return self.get("security_check_command", {})
```
mattmb: I'd probably return `None` instead of an empty dict just for clarity

paasta_tools/cli/cmds/security_check.py
```diff
@@ -15,29 +15,38 @@
...
 
 def perform_security_check(args):
-    paasta_print('Not implemented yet')
+    security_check_command = load_system_paasta_config().get_security_check_command()
+    if not security_check_command:
+        paasta_print("Nothing to be executed during the security-check step")
+        return 0
+
+    ret_code, output = _run(security_check_command)
+    paasta_print("{0} exited with {1}: {2}", security_check_command, ret_code, output)
```
mattmb: This is okay like this. The following is just if you want to take a look, I'm happy for it to ship as is :-)

You might want to look at is the args for `_run`:
https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L1024

I'm not super familiar with it but it looks like there's some logic for streaming the output so you don't have to wait for the command to complete to see the results. You could use that and then just print the return code here. It also looks like it has a timeout option. That might be worth enabling because then we could be sure that this step won't run for more than x minutes which would mean we won't hold up pipelines in the case when something gets stuck.

paasta_tools/utils.py
transcedentalia: Done.

paasta_tools/cli/cmds/security_check.py
transcedentalia: It was a quick change, added all ^.

solarkennedy:
Check out the failing pre-commit:
https://travis-ci.org/Yelp/paasta/jobs/221066729

Activate precommit by committing with a virtualenv on or run the itests locally to repro.

PR #1135, opened 2017-04-12
handle interrupts, quote command properly on local side

keymone:
ssh with pseudo-tty and kill tasks on term/int/hup signals

### DIFF
--- a/paasta_tools/cli/utils.py
+++ b/paasta_tools/cli/utils.py
@@ -20,6 +20,7 @@ import logging
 import os
 import pkgutil
 import re
+import subprocess
 import sys
 from socket import gaierror
 from socket import gethostbyname_ex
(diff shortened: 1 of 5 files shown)


keymone:
this is PAASTA-10234

paasta_tools/cli/utils.py
```diff
@@ -601,7 +601,7 @@ def run_on_master(cluster, system_paasta_config, cmd_parts,
     except NoMasterError as e:
         return (err_code, str(e))
 
-    ssh_parts = ['ssh', '-A', '-n', master]
+    ssh_parts = ['ssh', '-t', '-t', '-A', '-n', master]
```
solarkennedy: two '-t' ? why do we need to force a tty?

paasta_tools/cli/utils.py
```diff
@@ -595,23 +596,39 @@ def run_on_master(cluster, system_paasta_config, cmd_parts,
...
     try:
         master = connectable_master(cluster, system_paasta_config)
     except NoMasterError as e:
         return (err_code, str(e))
 
-    ssh_parts = ['ssh', '-A', '-n', master]
-
-    if shell:
-        ssh_parts.extend('sh -c "%s"' % quote(' '.join(cmd_parts)))
+    import subprocess
```
Rob-Johnson: move this to the top of the file.

paasta_tools/cli/utils.py
keymone: i'm not entirely sure what's happening, but if i don't force tty stdout of cmd behind ssh stops flushing(?)

PR #1137, opened 2017-04-12
Don't throw an exception for unknown instance types

oktopuz:
internal ticket: PAASTA-10418

### DIFF
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -40,6 +40,11 @@ def chronos_instance_status(instance_status, service, instance, verbose):
     return cstatus
 
 
+def adhoc_instance_status(instance_status, service, instance, verbose):
+    cstatus = {}
+    return cstatus
+
+
 def marathon_job_status(mstatus, client, job_config):
     try:
         app_id = job_config.format_marathon_app_dict()['id']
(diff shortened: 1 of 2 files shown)


paasta_tools/api/views/instance.py
```diff
@@ -111,9 +111,6 @@ def instance_status(request):
             instance_status['marathon'] = marathon_instance_status(instance_status, service, instance, verbose)
         elif instance_type == 'chronos':
             instance_status['chronos'] = chronos_instance_status(instance_status, service, instance, verbose)
-        else:
```
solarkennedy: I think the "error" handling here is "correct". Instead of removing all errors, can you make a "stub" `adhoc_instance_status` function that returns an empty array like `chronos_instance_status` ? Then we have room to get more status later on (maybe), but we won't swallow a 404.

paasta_tools/paasta_serviceinit.py
```diff
@@ -122,15 +122,18 @@ def main():
...
-        version = get_deployment_version(actual_deployments, cluster, instance)
-        paasta_print('instance: %s' % PaastaColors.blue(instance))
-        paasta_print('Git sha:    %s (desired)' % version)
-
         try:
             instance_type = validate_service_instance(service, instance, cluster, args.soa_dir)
+            if instance_type == 'adhoc':
+                continue
+
+            # For an instance, there might be multiple versions running, e.g. in crossover bouncing.
```
solarkennedy: This comment isn't really adding any value, just junk it

oktopuz:
I'll merge this PR as soon as tests are green.

paasta_tools/api/views/instance.py
```diff
@@ -111,6 +116,8 @@ def instance_status(request):
             instance_status['marathon'] = marathon_instance_status(instance_status, service, instance, verbose)
         elif instance_type == 'chronos':
             instance_status['chronos'] = chronos_instance_status(instance_status, service, instance, verbose)
+        elif instance_type['adhoc'] == 'adhoc':
```
mittonk: (later): Uh, did you want `elif instance_type == 'adhoc'`?

PR #1138, opened 2017-04-12
Cache load_marathon_service_config() for 5 seconds

oktopuz:
Cache load_marathon_service_config() for 5 seconds to avoid memorizing it forever in paasta.api.


oktopuz:
@solarkennedy, this is green as well.

PR #1139, opened 2017-04-13
Watcher for deployd for autoscaling

mattmb:
First stab at watching zookeeper for changes to the number of instances.

I'll add some itests + unit tests before I merge this but it works for me in the example cluster.

### DIFF
--- a/paasta_tools/deployd/watchers.py
+++ b/paasta_tools/deployd/watchers.py
@@ -6,9 +6,13 @@ import time
 from functools import reduce
 
 import pyinotify
+from kazoo.protocol.states import EventType
+from kazoo.recipe.watchers import ChildrenWatch
+from kazoo.recipe.watchers import DataWatch
 
 from paasta_tools.deployd.common import PaastaThread
 from paasta_tools.deployd.common import ServiceInstance
+from paasta_tools.long_running_service_tools import AUTOSCALING_ZK_ROOT
 from paasta_tools.marathon_tools import DEFAULT_SOA_DIR
 from paasta_tools.utils import list_all_instances_for_service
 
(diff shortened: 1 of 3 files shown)


mattmb:
I'm going to work on the itests separately but I've added some unit tests so this is ready for review

paasta_tools/deployd/watchers.py
```diff
@@ -23,6 +26,59 @@ def __init__(self, inbox_q, cluster):
...
+    def process_folder_event(self, children, event):
+        self.log.debug("Folder change: {}".format(event))
+        if event and (event.type == EventType.CHILD):
+            fq_children = ["{}/{}".format(event.path, child) for child in children]
+            for child in fq_children:
+                if child not in self.watchers:
+                    self.watch_folder(child)
+
+    def run(self):
+        self.watchers['/autoscaling'] = self.watch_folder('/autoscaling')
```
solarkennedy: I'm worried that you are "hard-coding" these folders here, and I'm pretty sure they are wrong. (where is the cluster name?)

Please use (and refactor as needed) the function that returns this path
https://github.com/Yelp/paasta/blob/master/paasta_tools/long_running_service_tools.py#L322

Wow ok mabe the zk client itself is already chrooted inside the cluster-specific namespace

mattmb: Yep, exactly that. It's a good shout not to hardcode though. I'll send a fixup!

paasta_tools/deployd/watchers.py
```diff
@@ -23,6 +27,59 @@ def __init__(self, inbox_q, cluster):
...
+        self.log.debug("Adding watch on {}".format(path))
+        DataWatch(self.zk, path, func=self.process_node_event, send_event=True)
+
+    def process_node_event(self, data, stat, event):
+        self.log.debug("Node change: {}".format(event))
+        if event and (event.type == EventType.CREATED or event.type == EventType.CHANGED):
+            service, instance = event.path.split('/')[-3:-1]
+            self.log.info("Number of instances changed or autoscaling enabled for first time"
+                          " for {}.{}".format(service, instance))
+            service_instance = ServiceInstance(service=service,
```
solarkennedy: I think there is a high likelyhood of encountering a bug where the inputs here are malformed.

For example there could be some race where there is zk data but the service doesn't exist or something.

This may not be "user input" but I wouldn't trust this path.split stuff from ZK, there could literally be anything in there.

solarkennedy:
If this doesn't work, then an alternative approach would be for this thing to *add* the watchers itself whenever it detects new marathon files with min/max instances stuff. Basically you could lookup what watchers you want and watch them, instead of looking at all the zk nodes and making watchers in response.

paasta_tools/deployd/watchers.py
mattmb: Hmm, I guess this could go wrong. I would think the "path" might be okay though, surely ZK isn't going to return garbage as the name of a path? Anyway, I guess we'll find out!

mattmb:
Now that you mention it I suspect that might have been a better way to go. I guess it depends on whether the kazoo + zookeeper watchers are more/less reliable than the pyinotify watchers...

PR #1142, opened 2017-04-14
Update schema to require that 0 < bounce_margin_factor <= 1

nhandler:
From the docs (https://paasta.readthedocs.io/en/latest/yelpsoa_configs.html#marathon-clustername-yaml):

```
bounce_margin_factor: proportionally increase the number of old instances to be drained when the crossover bounce method is used. 0 < bounce_margin_factor <= 1. Defaults to 1 (no influence). This allows bounces to proceed in the face of a percentage of failures. It doesn’t affect any other bounce method but crossover.
```

### DIFF
--- a/paasta_tools/cli/schemas/marathon_schema.json
+++ b/paasta_tools/cli/schemas/marathon_schema.json
@@ -139,7 +139,11 @@
                 },
                 "bounce_margin_factor": {
                     "type": "number",
-                    "default": 1
+                    "default": 1,
+                    "minimum": 0,
+                    "maximum": 1,
+                    "exclusiveMinimum": true,
+                    "exclusiveMaximum": false
                 },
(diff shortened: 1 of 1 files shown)


nhandler:
Test failures appear unrelated to this change

EvanKrall:
Doesn't 0 make sense? (this would make crossover roughly equivalent to brutal)

nhandler:
I haven't gone through the code yet to see if there is a (good) reason for not allowing zero. This was just a quick and easy safety net to avoid someone making the same mistake I did. If we decide to allow zero in the future, we can adjust `exclusiveMinimum`

EvanKrall:
What is *your* reason for not allowing zero?

EvanKrall:
Oh, the assertion in crossover_bounce. Gross.

PR #1144, opened 2017-04-17
More caching for paasta status

oktopuz:
* cache get_current_tasks() for paasta status explicitly
 * cache cronos.client for paasta status
 * always cache get_service_instance_list for 5 seconds

'paasta status -s yelp-main -c norcal-prod' takes 20 seconds.
'paasta status -s yelp-main -c norcal-prod -v' takes 3m20s.

### DIFF
--- a/paasta_tools/chronos_serviceinit.py
+++ b/paasta_tools/chronos_serviceinit.py
@@ -21,7 +21,8 @@ import humanize
 import isodate
 
 from paasta_tools import chronos_tools
-from paasta_tools.mesos_tools import get_running_tasks_from_frameworks
+from paasta_tools.mesos_tools import get_cached_list_of_running_tasks_from_frameworks
+from paasta_tools.mesos_tools import select_tasks_by_id
 from paasta_tools.mesos_tools import status_mesos_tasks_verbose
 from paasta_tools.utils import _log
 from paasta_tools.utils import calculate_tail_lines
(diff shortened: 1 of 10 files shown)


tests/test_marathon_serviceinit.py
```diff
@@ -940,29 +940,30 @@ def test_get_short_task_id():
...
 def test_status_mesos_tasks_working():
     with mock.patch(
-            'paasta_tools.marathon_serviceinit.get_running_tasks_from_frameworks', autospec=True) as mock_tasks:
-        mock_tasks.return_value = [
-            {'id': 1}, {'id': 2}
-        ]
+            'paasta_tools.marathon_serviceinit.get_cached_list_of_running_tasks_from_frameworks',
+            autospec=True) as mock_tasks:
+        mock_tasks.return_value = [{'id': 'unused{0}unused{0}'.format(marathon_tools.MESOS_TASK_SPACER)}
+                                   for _ in xrange(2)]
```
solarkennedy: this is not py3 compat, check out the failing py3 tests

solarkennedy:
the deployd tests are failing again. 

Other than the python3 thing that needs to be fixed, I say ship. I can't think of a much better way to get cached functions without a bit of duplication, but I'm ok with it, and it means we can sprinkle in cached clients whenever.

I dare saw we should *always* used a cached client?

oktopuz:
I've fixed the py3 tests.

Let's merge this PR as it is, and I'll check if it is possible to use a cached mesos client by default everywhere and remove this duplication.

paasta_tools/chronos_serviceinit.py
```diff
@@ -330,22 +320,26 @@ def status_chronos_jobs(client, jobs, job_config, verbose):
         desired_state = job_config.get_desired_state_human()
         output.append("Desired:    %s" % desired_state)
         for job in jobs:
-            running_tasks = get_running_tasks_from_frameworks(job["name"])
-            output.append(format_chronos_job_status(client, job, running_tasks, verbose))
+            running_task_count = len(select_tasks_by_id(get_cached_list_of_running_tasks_from_frameworks(),
```
mattmb: Maybe something like this works:
```
time_cache(ttl=600)(get_running_tasks_from_frameworks(job["name"]))
```
or it might be?
```
time_cache(get_running_tasks_from_frameworks(job["name"]))(ttl=600)
```
I have a memory that something like that is possible, and it would avoid duplicating the functions just to turn on the caching?

mattmb: But yeah, don't block on shipping it!

PR #1145, opened 2017-04-18
Make setup_marathon_job return bounce_again var

mattmb:
This adds an extra return variable to setup_marathon_job with a time in
which the service instance it is deploying should be bounced again. To
start with this aims to be simple and naive. If it thinks everything is
steady state it returns -1, if it thinks more work needs to be done it
returns 60 (seconds). This can be further nuanced in the future.

This extra return variable is consumed by the paasta deployd workers
which can then re-add services to the bounce queue. For the existing
cron jobs the extra variable is just ignored.

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -241,7 +241,7 @@ def do_bounce(
                 if e.status_code == 409:
                     log.warning("Failed to create, app %s already exists. This means another bounce beat us to it."
                                 " Skipping the rest of the bounce for this run" % marathon_jobid)
-                    return
+                    return 60
                 raise
 
     tasks_to_kill = drain_tasks_and_find_tasks_to_kill(
@@ -295,7 +295,7 @@ def do_bounce(
         (apps_to_kill or tasks_to_kill),
         apps_to_kill == list(old_app_live_happy_tasks),
         tasks_to_kill == all_old_tasks,
-    ]):
+    ]) or (not all_old_tasks) and new_app_running:
         log_bounce_action(
             line='%s bounce on %s finishing. Now running %s' %
             (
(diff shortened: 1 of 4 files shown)


paasta_tools/setup_marathon_job.py
```diff
@@ -554,13 +557,13 @@ def log_deploy_error(errormsg, level='event'):
     except bounce_lib.LockHeldException:
         logline = 'Failed to get lock to create marathon app for %s.%s' % (service, instance)
         log_deploy_error(logline, level='debug')
-        return (0, "Couldn't get marathon lock, skipping until next time")
```
solarkennedy: Are you afraid of an infinite loop here? If we can't get a lock, then shouldn't we return -1 because another one beat us to it?

mattmb: Hmm, I think that would probably better. It also would then match the logic on L700

mattmb:
Looks like I've got some itests to fix up before I ship...

paasta_tools/setup_marathon_job.py
```diff
@@ -657,6 +660,19 @@ def main():
...
+    """deploy the service instance given and proccess return code
+    if there was an error we send a sensu alert.
+
+    :param service: The service name to setup
+    :param instance: The instance of the service to setup
+    :param client: A MarathonClient object
+    :param soa_dir: Path to yelpsoa configs
+    :param marathon_config: The service instance's configuration dict
+    :param marathon_apps: A list of all marathon app objects
+    :returns: A tuple of (status, bounce_in_seconds) to be used by paasta-deployd
```
Rob-Johnson: how about making this tuple a namedtuple?

mattmb: I did think I might at the time. We actually have two different tuples currently. I think I'll leave this be for now but I might "make this better" in my next stab at this file.

PR #1146, opened 2017-04-18
Increase timeout on deployd itests

mattmb:
Will this make things less flappy on travis?


tub:
I'm guessing you only need a shipit if travis gets less flappy? :)

solarkennedy:
I'll give you a shipit regardless

mattmb:
A failure but a different failure so I'm calling that a "success"...

PR #1148, opened 2017-04-18
Use wheels in non yelp env

jglukasik:
We've were getting some Travis build failures in compiling tornado. With this change we can use wheels for that when outside of the Yelp env.  This way we also don't need all the -dev packages

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -2,7 +2,7 @@
 # -*- makefile -*-
 
 ifeq ($(PAASTA_ENV),YELP)
-	EXTRA_PYPI="--extra-index-url 'https://pypi.yelpcorp.com/simple'"
+	EXTRA_PYPI="--extra-index-url 'https://pypi.yelpcorp.com/simple' --preinstall no-manylinux1"
 endif
 
 %:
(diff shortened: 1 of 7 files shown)


debian/rules
```diff
@@ -23,4 +23,4 @@ override_dh_auto_test:
 	true
 
 override_dh_virtualenv:
-	dh_virtualenv $EXTRA_PYPI --python=/usr/bin/python2.7 --preinstall no-manylinux1 --preinstall pip-custom-platform --pip-tool pip-custom-platform
+	dh_virtualenv $EXTRA_PYPI --python=/usr/bin/python2.7 --preinstall pip-custom-platform --pip-tool pip-custom-platform
```
asottile: you need `no-manylinux1`, manylinux wheels are not compatible with dh-virtualenv

this essentially defeats the purpose of this entire review :P

solarkennedy: Crap! why are manylinux wheels not compatible with dh-virtualenv?
You mean in some sort of technical sense, right? (not just yelp policy)

asottile: not just yelp policy: https://github.com/spotify/dh-virtualenv/issues/150

solarkennedy: Thank you for sharing the knowledge :)

PR #1152, opened 2017-04-19
--deploy-group for paasta start/stop/restart

oktopuz:
internal ticket: PAASTA-10461

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -21,10 +21,12 @@ import socket
 from paasta_tools import remote_git
 from paasta_tools import utils
 from paasta_tools.chronos_tools import ChronosJobConfig
+from paasta_tools.cli.cmds.status import paasta_args_mixer
 from paasta_tools.cli.utils import figure_out_service_name
 from paasta_tools.cli.utils import get_instance_config
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_all_instances_for_service
+from paasta_tools.cli.utils import list_deploy_groups
 from paasta_tools.cli.utils import list_instances
 from paasta_tools.cli.utils import list_services
 from paasta_tools.generate_deployments_for_service import get_latest_deployment_tag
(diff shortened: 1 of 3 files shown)


oktopuz:
```
Failing scenarios:
  paasta_deployd.feature:12  deployd will re-deploy an app if its marathon-*.yaml changes
  paasta_deployd.feature:27  deployd will re-deploy an app if its deployment.json is updated
  paasta_deployd.feature:41  deployd starts and only one leader
```

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -170,13 +175,13 @@ def paasta_start_or_stop(args, desired_state):
     # If args.clusters is falsey, then default to *all* clusters that a service is deployed to,
     # and we figure out which ones are needed for each service later.
     if instances:
-        instance_clusters = [list_clusters(service, soa_dir, instance) for instance in args.instances]
+        instance_clusters = [list_clusters(service, soa_dir, instance) for instance in instances]
```
mattmb: If instances is `None` (L169) then you'll get a `TypeError` here right?

oktopuz: If isntances is None or [], we are not going to be inside this block at all.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -160,8 +161,12 @@ def paasta_start_or_stop(args, desired_state):
     """Requests a change of state to start or stop given branches of a service."""
     soa_dir = args.soa_dir
     service = figure_out_service_name(args=args, soa_dir=soa_dir)
-    instances = args.instances.split(",") if args.instances else None
 
+    pargs = paasta_args_mixer(args, service)
```
solarkennedy: Don't you still need to add the "--deploy-group" option in the arg parser?

oktopuz: Indeed.
I added the options and updated tests to pass a real argparse tuple instead of mocking it.

This reveals that --cluster is a mandatory option for paasta start/stop/restart. That doesn't make lots of sense in combination with --deploy-group.
For now I left it as it was. But we might want to reconsider the way it works.

paasta_tools/cli/cmds/start_stop_restart.py
mattmb: That's quite true! Ship away.

PR #1154, opened 2017-04-20
Reduce possibility of noisy alerts due to delayed chronos jobs

oktopuz:
internal ticket: PAASTA-10506


oktopuz:
Failing itests scenarios:
  paasta_deployd.feature:12  deployd will re-deploy an app if its marathon-*.yaml changes
  paasta_deployd.feature:27  deployd will re-deploy an app if its deployment.json is updated
  paasta_deployd.feature:41  deployd starts and only one leader

PR #1155, opened 2017-04-20
A new version of pyOpenSSL requires this upgrade

mattmb:
There was a new release https://pypi.python.org/pypi/pyOpenSSL today
that breaks our build because we've pinned cryptography. This upgrades
it to see if we get a clean build.


mattmb:
That failure is a know flake so I'm gonna merge

PR #1156, opened 2017-04-20
Ensure we always scale up if task data missing

mattmb:
Even if there are not enough healthy tasks to get a completely accurate
calculation on utilization we should still scale up if the remaining
tasks are overutilized. This should prevent the autoscaler becoming
"wedged" if there are a few unhealthy tasks lingering in a small pool of
total tasks. This should preserve the current behaviour if the
utilization suggests we need to scale down. That is we still bail and do
nothing rather than scaling down because it could be risky to do so with
incomplete data.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -443,30 +443,15 @@ def get_utilization(marathon_service_config, autoscaling_params, log_utilization
     )
 
 
-def autoscale_marathon_instance(marathon_service_config, marathon_tasks, mesos_tasks):
-    current_instances = marathon_service_config.get_instances()
+def is_task_data_insufficient(marathon_service_config, marathon_tasks, current_instances):
     too_many_instances_running = len(marathon_tasks) > int((1 + MAX_TASK_DELTA) * current_instances)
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -492,6 +477,12 @@ def autoscale_marathon_instance(marathon_service_config, marathon_tasks, mesos_t
 
     safe_downscaling_threshold = int(current_instances * 0.7)
     if new_instance_count != current_instances:
+        if new_instance_count < current_instances and task_data_insufficient:
+            write_to_log(config=marathon_service_config,
+                         line='Delaying scaling *down* as we found too many or too few tasks running in marathon. '
```
solarkennedy: Delaying scaling *down* as we found too few healthy tasks running in marathon. '

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -492,6 +477,12 @@ def autoscale_marathon_instance(marathon_service_config, marathon_tasks, mesos_t
...
+                              'This can happen because tasks are delayed/waiting/unhealthy or because we are '
```
solarkennedy: "Will wait for all tasks to be healthy before making a decision to downscale"

PR #1157, opened 2017-04-21
Fix threaded timeout decorator

mattmb:
Currently the TimeoutError is thrown but the main thread will still
continue to wait for the function thread to finish before it exits
itself. Setting the daemon flag means that the child thread will be
killed if the main thread dies, which it does on receiving the timeout
exception.

Internally: PAASTA-10556

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -1829,6 +1829,7 @@ class _Timeout(object):
         self.func_thread = threading.Thread(target=self.run,
                                             args=args,
                                             kwargs=kwargs)
+        self.func_thread.daemon = True
         self.timeout = self.seconds + time.time()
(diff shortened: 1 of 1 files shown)


solarkennedy:
I'm really surprised this is it?
I thought for sure we have itest on this decorator

https://github.com/Yelp/paasta/blob/master/general_itests/timeout.feature

Do you know why it passed?

mattmb:
Hilariously @solarkennedy we seem to have a second timeout function in utils which this test is using. Interestingly, I'm not sure this test would have caught this bug anyway because the "sleep 1" would have finished after 1 second and then both threads would exit and behave as designed...

Really I should probably add an itest for this codepath but it feels like it could be time consuming to get right.

PR #1161, opened 2017-04-25
security-check returns the test results

transcedentalia:


paasta_tools/cli/cmds/security_check.py
```diff
@@ -49,4 +49,4 @@ def perform_security_check(args):
     ret_code, output = _run(security_check_command, timeout=300, stream=True)
     paasta_print("{} exited with code {}".format(security_check_command, ret_code))
     # The security-check should not block the pipeline in case of a failure
```
Rob-Johnson: can you remove this comment, if we are infact now blocking pipelines?

transcedentalia: Sure. The pipeline will still not block in case the security-check fails, but this is because of the Jenkins configuration.

PR #1162, opened 2017-04-25
Add metric emission to deployd

mattmb:
This adds some basic metric emission to deployd. Various timers are
emitted by the deployd workers. In addition, we check the length of the
queues every 20s and emit a gauge.

I'm using yelp_meteorite because it is super easy for this purpose.
Unfortunately it's not public so I've hacked the import to fall back to
a mock and that will make the metrics code "do nothing" if you are
outside yelp. I discussed making this a pluggable interface with robj
but I think it's overkill to do so but please shout up if you think we
should.

### DIFF
--- a/paasta_tools/deployd/master.py
+++ b/paasta_tools/deployd/master.py
@@ -14,6 +14,7 @@ from paasta_tools.deployd.common import PaastaQueue
 from paasta_tools.deployd.common import PaastaThread
 from paasta_tools.deployd.common import ServiceInstance
 from paasta_tools.deployd.leader import PaastaLeaderElection
+from paasta_tools.deployd.metrics import QueueMetrics
 from paasta_tools.deployd.workers import PaastaDeployWorker
 from paasta_tools.marathon_tools import DEFAULT_SOA_DIR
 from paasta_tools.utils import get_services_for_cluster
@@ -44,10 +45,10 @@ class Inbox(PaastaThread):
         except Empty:
             service_instance = None
         if service_instance:
-            self.log.debug("Processing {} to decide if we need to add it to bounce q".format(service_instance))
+            self.log.debug("Processing {}.{} to see if we need to add it to bounceq".format(service_instance.service,
+                                                                                            service_instance.instance))
             self.process_service_instance(service_instance)
         if self.inbox_q.empty() and self.to_bounce:
-            self.log.debug("Cleared inbox, checking if any services are ready to bounce")
             self.process_to_bounce()
         time.sleep(1)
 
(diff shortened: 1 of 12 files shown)


paasta_tools/deployd/metrics.py
```diff
@@ -0,0 +1,39 @@
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import time
+
+from paasta_tools.deployd.common import PaastaThread
+
+try:
+    import yelp_meteorite
```
solarkennedy: A more friendly way to do this is to "just" use python-statsd directly instead of the yelp_meterorite wrapper, except we don't get all the "free" dimensions :(

mattmb: I decided to follow @Rob-Johnson's advice in the end and turn it into an interface like our logging interfaces. That way we can still use yelp_meteorite but anyone that wanted to use statsd directly would have an easy way to add it in.

Rob-Johnson: <3

paasta_tools/deployd/master.py
```diff
@@ -44,10 +46,10 @@ def process_inbox(self):
         except Empty:
             service_instance = None
         if service_instance:
-            self.log.debug("Processing {} to decide if we need to add it to bounce q".format(service_instance))
+            self.log.debug("Processing {}.{} to see if we need to add it to bounceq".format(service_instance.service,
```
Rob-Johnson: is this a typo? if not, write out 'bounce queue' so people don't ticket that you've typo'd :)

paasta_tools/deployd/metrics.py
```diff
@@ -0,0 +1,79 @@
...
+    return outer
+
+
+def get_metrics_interface(name):
+    return _metricis_interfaces[name]()
+
+
+@register_metrics_interface(None)
+class NoMetrics(object):
+    def create_timer(self, *args, **kwargs):
```
Rob-Johnson: i'm not convinced about this pattern of using ``*args, **kwargs`` - it kind of implies to me that the metrics interface is inconsistent between implementations?

paasta_tools/utils.py
```diff
@@ -940,6 +940,16 @@ def get_log_reader(self):
         except KeyError:
             raise PaastaNotConfiguredError('Could not find log_reader in configuration directory: %s' % self.directory)
 
+    def get_deployd_metrics_provider(self):
+        """Get the metrics_provider configuration out of global paasta config
+
+        :returns: A string identifying the metrics_provider
+        """
+        try:
```
Rob-Johnson: this can just be ``return self.get('deployd_metrics_provider')``

paasta_tools/deployd/metrics.py
mattmb: Yeah, I guess. I was trying to use kwargs at least because it makes it easy to add dimensions as kwargs when you call this from the code...

PR #1163, opened 2017-04-26
U/jgl/paasta 10492 quiet logs

jglukasik:
Make itest logs less verbose by quieting the output of apt-get by redirecting to /dev/null

I attempted to do this the "clean" way, with the -qq argument, but still got a lot of extraneous output, from the forked dpkg process. I tried to pass along an option to dpkg to quiet its output, but that also did not work. [1]

[1] https://peteris.rocks/blog/quiet-and-unattended-installation-with-apt-get/


solarkennedy:
Looks better. I can see the all the logs in travis now without going raw.

Can you also add DEBIAN_FRONTEND=noninteractive ? This is considered best docker practice and will remove some of the debconf warnings that are going to stderr.

yelp_package/dockerfiles/itest/chronos/Dockerfile
```diff
@@ -14,20 +14,20 @@
 
 FROM ubuntu:trusty
 
-RUN apt-get update && apt-get -y install apt-transport-https
+RUN DEBIAN_FRONTEND=noninteractive apt-get update > /dev/null && DEBIAN_FRONTEND=noninteractive apt-get -y install apt-transport-https > /dev/null
```
solarkennedy: Just do

    ENV DEBIAN_FRONTEND=noninteractive

So you don't have to repeat yourself 8 times in the dockerfile

jglukasik: That was going to be my approach, but then I found this discussion: https://github.com/moby/moby/issues/4032#issuecomment-34597177

Seems like they discourage setting it with ENV, because then things will be weird when doing a `docker run -it <> bash`

solarkennedy: I think they are silly to actively discourage ENV variables like that. They are technically correct, but these are not docker images that people are pulling from around the world or anything, they are our images.

Please just do it to make it cleaner. I accept the caveat that anyone using these docker images in an interactive way will not get apt-get output. Fine :)

solarkennedy:
Ship

asottile:
Adding `ENV DEBIAN_FRONTEND=noninteractive` goes against docker best practices as you're exposing this environment variable to your runtime (and to all the commands) -- the best practice here is to *only* include it on the `apt-get install` command

asottile:
An aside: I've been debugging apt installation related stuff and it's very very difficult with all of them piped to `/dev/null`

PR #1167, opened 2017-04-26
[WIP] Task processing mesos scheduler

sagar8192:
Just wrote mesos scheduler and implemented the task execution interface per the CEP. Let me know what you think about it.

### DIFF
--- a/task_processing/task_processing/docker_task_configs.py
+++ b/task_processing/task_processing/docker_task_configs.py
@@ -2,6 +2,7 @@
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+
 class DockerTaskConfigs(object):
     def __init__(self, image, cmd, cpus, mem, disk, volumes, ports):
         self.image = image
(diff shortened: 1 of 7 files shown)


Rob-Johnson:
@sagar8192 most of the Mesos work is done for us in the [Native Scheduler](https://github.com/Yelp/paasta/blob/master/paasta_tools/frameworks/native_scheduler.py) already - we do need to do some refactoring to take the paasta specific parts like service/instance out of it though.

task_processing/task_processing/docker_task_configs.py
```diff
@@ -0,0 +1,13 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+class DockerTaskConfigs(object):
```
keymone: can this just be a named tuple?

keymone: dict is even better though might be less convenient (['blah'] instead of .blah), also namedtuple will freak out on typos while dict won't. i'm fine with both.

PR #1168, opened 2017-04-26
add service and instance labels to docker command line

somic:
Implements one of the approaches @solarkennedy and I discussed IRL. Looks clean enough to me to deserve a shot.


paasta_tools/utils.py
```diff
@@ -232,15 +232,23 @@ def get_cap_add(self):
...
-                      {"key": "cpu-period", "value": "%s" % int(self.get_cpu_period())},
-                      {"key": "cpu-quota", "value": "%s" % int(self.get_cpu_quota())}]
+        parameters = [
+            {"key": "memory-swap", "value": self.get_mem_swap()},
+            {"key": "cpu-period", "value": "%s" % int(self.get_cpu_period())},
+            {"key": "cpu-quota", "value": "%s" % int(self.get_cpu_quota())},
+        ]
+        if with_labels:
+            parameters.extend([
+                {"key": "label", "value": "service=%s" % self.service},
```
solarkennedy: can you prefix these all with "paasta"

`paasta_service=`
`paasta_instance=`
`paasta_team=`

This will line up with the env vars we set and will not collide with any future labels. (I think it is helpful to know where they come from)

Also can you add team?

solarkennedy:
To confirm here: the goal of this particular approach is to deploy new metadata without requiring a bounce, with the caveat that you will have to bounce a service to pick up that new metadata.

Team is the only thing that anyone would care about changing and have it reflect. I think it is ok for that to not take effect "instantly". I say ship.

paasta_tools/utils.py
somic: Done, renamed labels and added team. I am assuming config_dict will come to me with contents of monitoring.yaml already parsed and by default the team will be set to "none".

As discussed, can't use get_team from monitoring_tools here because monitoring_tools.py imports utils.

paasta_tools/utils.py
```diff
@@ -232,19 +231,35 @@ def get_cap_add(self):
...
+            parameters.extend([
+                {"key": "label", "value": "paasta_service=%s" % self.service},
+                {"key": "label", "value": "paasta_instance=%s" % self.instance},
+                {"key": "label", "value": "paasta_team=%s" % self.get_team()},
+            ])
         parameters.extend(self.get_ulimit())
         parameters.extend(self.get_cap_add())
         return parameters
 
+    def get_team(self):
```
solarkennedy: Unless you are going to go all the way and move this for real, just don't do this.

It doesn't do enough, it won't default to what is in monitoring.yaml, etc. 

Just skip team for now and get the MVP. We can add team later (and it won't cause any bounces)

PR #1170, opened 2017-04-27
Add some basic config options for deployd

mattmb:
This adds the bare bones obvious things we'd want to configure from
config files.


Rob-Johnson:
got any tests?

mattmb:
I've deliberately avoided writing any for deployd so far although I think it's probably time to do so...

I'll add tests here for the utils.py changes but I'll open a separate PR for the rest of the deployd tests.

PR #1172, opened 2017-04-27
remote-run start/stop/list, --detach for background mode, optional --run-id for naming your runs

keymone:
paasta remote-run gets subcommands: start / stop / list

paasta remote-run start:
- fix dry-run on master
- run in background with --detach
- optionally name your run with --run-id <blah>

paasta remote-run list: lists all active frameworks for specific cluster/service/instance with their respective run ids and framework ids

paasta remote-run stop: stop remote-run of cluster/service/instance by it's run id (-R) or framework id (-F)

### DIFF
--- a/paasta_tools/cli/cmds/remote_run.py
+++ b/paasta_tools/cli/cmds/remote_run.py
@@ -81,7 +81,7 @@ def add_common_args_to_parser(parser):
         help='Show Docker commands output',
         action='store_true',
         required=False,
-        default=True,
+        default=False,
     )
     parser.add_argument(
         '-R', '--run-id',
--- a/paasta_tools/cli/cmds/remote_run.py
+++ b/paasta_tools/cli/cmds/remote_run.py
@@ -35,7 +35,7 @@ from paasta_tools.utils import PaastaNotConfiguredError
 from paasta_tools.utils import SystemPaastaConfig
 
 
-def start_args(parser):
+def add_start_args_to_parser(parser):
     parser.add_argument(
         '-C', '--cmd',
         help=('Run Docker container with particular command, '
@@ -60,7 +60,7 @@ def start_args(parser):
     )
 
 
-def common_args(parser):
+def add_common_args_to_parser(parser):
     parser.add_argument(
         '-s', '--service',
         help='The name of the service you wish to inspect',
@@ -125,8 +125,8 @@ def add_subparser(subparsers):
         'start',
         help="Start task subcommand"
     )
-    start_args(start_parser)
-    common_args(start_parser)
+    add_start_args_to_parser(start_parser)
+    add_common_args_to_parser(start_parser)
     start_parser.add_argument(
         '-X', '--constraint',
         help='Constraint option, format: <attr>,OP[,<value>], OP can be one of '
@@ -142,7 +142,7 @@ def add_subparser(subparsers):
         'stop',
         help="Stop task subcommand"
     )
-    common_args(stop_parser)
+    add_common_args_to_parser(stop_parser)
     stop_parser.add_argument(
         '-F', '--framework-id',
         help='ID of framework to stop. Must belong to remote-run of selected service instance.',
@@ -154,7 +154,7 @@ def add_subparser(subparsers):
         'list',
         help="Stop task subcommand"
     )
-    common_args(list_parser)
+    add_common_args_to_parser(list_parser)
 
     main_parser.set_defaults(command=paasta_remote_run)
 
--- a/paasta_tools/cli/cmds/remote_run.py
+++ b/paasta_tools/cli/cmds/remote_run.py
@@ -209,9 +209,10 @@ def paasta_remote_run(args):
     if len(constraints) > 0:
         cmd_parts.extend(['--constraints-json', quote(json.dumps(constraints))])
 
+    graceful_exit = (args.action == 'start' and not args.detach)
     return_code, status = run_on_master(
         args.cluster, system_paasta_config, cmd_parts,
-        graceful_exit=args.action == 'start' and not args.detach)
+        graceful_exit=graceful_exit)
 
     # Status results are streamed. This print is for possible error messages.
     if status is not None:
--- a/paasta_tools/cli/cmds/remote_run.py
+++ b/paasta_tools/cli/cmds/remote_run.py
@@ -93,9 +93,9 @@ def add_common_args_to_parser(parser):
     parser.add_argument(
         '-d', '--dry-run',
         help='Don\'t launch the task',
-        action='count',
+        action='store_true',
         required=False,
-        default=0,
+        default=False,
     )
     parser.add_argument(
         '-i', '--instance',
@@ -181,6 +181,7 @@ def paasta_remote_run(args):
         'yelpsoa_config_root': DEFAULT_SOA_DIR,
         'cmd': None,
         'verbose': True,
+        'dry_run': False,
         'staging_timeout': None,
         'detach': False,
         'run_id': None,
@@ -208,13 +209,8 @@ def paasta_remote_run(args):
     if len(constraints) > 0:
         cmd_parts.extend(['--constraints-json', quote(json.dumps(constraints))])
 
-    if args.dry_run > 0:
-        cmd_parts.append('--dry-run')
-
-    paasta_print('Running on master: %s' % ' '.join(cmd_parts))
     return_code, status = run_on_master(
         args.cluster, system_paasta_config, cmd_parts,
-        dry=args.dry_run > 1,
         graceful_exit=args.action == 'start' and not args.detach)
 
     # Status results are streamed. This print is for possible error messages.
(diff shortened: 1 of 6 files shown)


paasta_tools/cli/cmds/remote_run.py
```diff
@@ -52,50 +77,36 @@ def add_remote_run_args(parser):
...
-        dest='dry_run_json_dict',
-    )
-    parser.add_argument(
-        '-C', '--cmd',
-        help=('Run Docker container with particular command, '
-              'for example: "bash". By default will use the command or args specified by the '
-              'soa-configs or what was specified in the Dockerfile'),
         required=False,
-        default=None,
+        default=True,
```
Rob-Johnson: set this to False.

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -35,7 +35,32 @@
 from paasta_tools.utils import SystemPaastaConfig
 
 
-def add_remote_run_args(parser):
+def start_args(parser):
```
Rob-Johnson: maybe ``apply_start_args_to_parser`` to make it clear that this function has side effects?

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -52,50 +77,36 @@ def add_remote_run_args(parser):
...
-        help='Show Docker commands output',
-        action='store_true',
-        required=False,
-        default=True,
     )
     parser.add_argument(
         '-d', '--dry-run',
         help='Don\'t launch the task',
-        action='store_true',
+        action='count',
```
Rob-Johnson: howcome you're changing this from a boolean?

paasta_tools/cli/utils.py
```diff
@@ -606,23 +606,18 @@ def run_on_master(cluster, system_paasta_config, cmd_parts,
...
         stdin_interrupt = True
         popen_kwargs = {'preexec_fn': os.setsid}
     else:
         stdin_interrupt = False
         popen_kwargs = {}
 
-    cmd_parts = ['ssh', '-t', '-t', '-A', master, "/bin/bash", "-c", quote(' '.join(cmd_parts))]
-    paasta_print(' '.join(cmd_parts))
+    cmd_parts = [
+        'ssh', '-q', '-o', 'StrictHostKeyChecking no', '-t', '-t', '-A',
```
Rob-Johnson: ``StrictHostKeyChecking no`` seems like a bad default?

paasta_tools/paasta_remote_run.py
```diff
@@ -136,5 +181,80 @@ def handle_interrupt(_signum, _frame):
...
+
+
+def remote_run_list(args):
+    paasta_config, service, cluster, soa_dir, instance, instance_type = extract_args(args)
+    # native_config = paasta_config.get_paasta_native_config()
+    # credential = mesos_pb2.Credential()
+    # credential.principal = native_config()['principal']
+    # credential.secret = native_config()['secret']
+    # mesos_uri = '%s:%d' % (mesos_tools.get_mesos_leader(), mesos_tools.MESOS_MASTER_PORT)
+    frameworks = requests.get('http://localhost:5050/frameworks').json()['frameworks']
```
Rob-Johnson: could you add these functions to the mesos api client instead if they're missing, or use that instead if not? code is [here](https://github.com/Yelp/paasta/blob/master/paasta_tools/mesos/master.py)

paasta_tools/cli/cmds/remote_run.py
keymone: we had -D to mean dry-run for paasta_tools/cli/cmds/remote_run.py and -d - dry-run for paasta_tools/paasta_remote_run.py

now -D means --detach and -d is repeatable option, when passed once it goes to paasta_remote_run on master, more than once - to cli/cmds/remote_run

paasta_tools/paasta_remote_run.py
keymone: good point, i didn't realize we had that class, will do.

paasta_tools/paasta_remote_run.py
```diff
@@ -37,18 +44,34 @@
...
         '-X', '--constraints-json',
         help=('Mesos constraints JSON'),
         required=False,
         default=None,
     )
+
+    stop_parser = subs.add_parser('stop', help='Stop task')
+    add_common_args_to_parser(stop_parser)
+    stop_parser.add_argument(
+        '-F', '--framework-id',
```
solarkennedy: I was kind hoping that we *didn't* need another thing that users have to know. 

Can we get away with just "instance" and have that dynamically calculated?

For example if a user launches "yelp-main.mycool-job"
I think they should be able to stop it with "paasta stop yelp-main.mycool-job", *not* have to run "paasta local-run stop my-cool-job -F frameworkid"

Is this possible?

I guess it is difficult because the date string and run_id are now part of the framework, we can't easily calculate it.

How about an alternative: I'm sure we have a function to list the frameworks, can we use the "chooser" function we have in paasta and allow people to interactively choose the framework id? (default to None, use the completer= function to get tab completion, etc)

paasta_tools/paasta_remote_run.py
```diff
@@ -101,6 +129,22 @@ def main(argv):
     if command:
         overrides_dict['cmd'] = command
 
+    run_id = args.run_id
+    if run_id is None:
+        run_id = ''.join(
+            random.choice(string.ascii_uppercase + string.digits) for _ in range(8))
+        paasta_print("Assigned random run-id: %s" % run_id)
+
+    if args.detach:
```
solarkennedy: I kinda hate things that background.
Won't this make it literally impossible to debug why it isn't working?
Is this implemented for completeness, do we actually have a use case where we want to hide this (more than cntrl-z?)

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -52,50 +77,36 @@ def add_remote_run_args(parser):
...
-              'for example: "bash". By default will use the command or args specified by the '
-              'soa-configs or what was specified in the Dockerfile'),
         required=False,
-        default=None,
+        default=False,
     )
     parser.add_argument(
-        '-i', '--instance',
-        help=("Simulate a docker run for a particular instance of the service, like 'main' or 'canary'"),
+        '-R', '--run-id',
```
solarkennedy: Can we re-use instance for this? ("make" an instance on the fly?)

Because if people are saying "--instance main", what does that really mean in the context of local-run? On paasta logs, status, and sfx things will show up under "main", *not* under this run-id var.

paasta_tools/paasta_remote_run.py
keymone: sort of for completeness. ctrl-z would still make remote-run rely on ssh connection flakiness. also considering how little useful output remote-run script generates in stdout (basically "TASK_RUNNING, TASK_FINISHED") i think detached runs will be the norm.

paasta_tools/paasta_remote_run.py
keymone: there are 2 ways to uniquely identify remote-run: compound identifier of (service, cluster, instance, run-id) and framework-id. i only decided to expose framework-id for completeness purpose, i don't expect people to use it very often.

we also can't just use instance. normally i don't expect users to even use instance, it will default to "remote" and is just a configuration preset. potentially people would have multiple instances for some wildly different batch jobs, but the problem i was trying to solve with run-id is to identify a remote-run among multiple with the same (service, cluster, instance) value.

we can't force users to just have different instances, because it kind of defeats the purpose of ad hoc runs.

we can implement a chooser for when user didn't provide neither run-id nor framework id, it would be useful.

paasta_tools/cli/cmds/remote_run.py
keymone: not easily, because people might run multiple remote-runs within same service/cluster/instance.

paasta_tools/mesos/master.py
```diff
@@ -69,15 +69,21 @@ def host(self):
...
         try:
-            return requests.get(
+            return method(
                 urljoin(self.host, url),
                 timeout=self.config["response_timeout"],
                 **kwargs)
         except requests.exceptions.ConnectionError:
             raise exceptions.MasterNotAvailableException(MISSING_MASTER.format(self.host))
 
+    def fetch(self, url, **kwargs):
```
Rob-Johnson: if you're doing this for api compatibility, then remove the default option from ``self.request`` and just set it here instead.

paasta_tools/mesos/master.py
```diff
@@ -69,15 +69,21 @@ def host(self):
         return "{}://{}".format(self.config["scheme"], self.resolve(self.config["master"]))
 
     @log.duration
-    def fetch(self, url, **kwargs):
+    def request(self, url, method=requests.get, **kwargs):
```
Rob-Johnson: I don't really like that you're forcing an implementation detail on the user (the implementation makes an assumption that this is a function from requests, but you can't really enforce this at the parameter level). Instead, why not make this arg a string ``http_method`` (so you'd use it like ``request(url="foo", method="POST") and do something like what's mentioned in the docs for a [prepared request](http://docs.python-requests.org/en/master/user/advanced/#prepared-requests). It keeps the interface cleaner.

Issue #1169, opened 2017-04-27
wait-for-deployment should be less picky


EvanKrall:
Today a deploy was held up because wait-for-deployment took longer than necessary to exit, due to two problems:

 - Turns out my conditional in https://github.com/Yelp/paasta/pull/1141/files was wrong -- I think we want >, not <, so when you have e.g. 11/10 instances, wait-for-deployment will continue blocking. This can happen if marathon does marathon things, or if the autoscaler adjusts the desired number of instances down.
 - wait-for-deployment will block as long as the Marathon status is anything other than Running, so "Deploying" and "Waiting" will block. I think this is unnecessary -- if we have enough instances, we have enough instances.

We could also make wait-for-deployment aware of bounce_margin_factor, but to be honest we might just be able to get rid of any checking of replication, and rely on the bounce to do the right thing. (At least for crossover bounce -- by the time there are no more of the old instances, there should be enough new ones. This is not true for brutal or downthenup.)

solarkennedy:
@somic this is a good optimization to the bounce if you are interested.

somic:
@solarkennedy my PR above addressed one of the aspects of this issue (2nd point), you addressed the first point in another PR and what's left is this:

"
We could also make wait-for-deployment aware of bounce_margin_factor, but to be honest we might just be able to get rid of any checking of replication, and rely on the bounce to do the right thing. (At least for crossover bounce -- by the time there are no more of the old instances, there should be enough new ones. This is not true for brutal or downthenup.)
"

Drop this for now or reopen this issue in order to get this done?

solarkennedy:
I'm ok with this being considered "done". You could argue that the margin factor is really for the bounce "in progress" and shouldn't be considered for "completion".

PR #1174, opened 2017-04-27
Fix wait_for_deployment with not enough instances

solarkennedy:
Whoops

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -394,7 +394,7 @@ def _run_instance_worker(cluster_data, instances_out, green_light):
                 cluster_data.instances_queue.task_done()
                 instances_out.put(instance)
                 continue
-            if (status.marathon.expected_instance_count <
+            if (status.marathon.expected_instance_count >
                     status.marathon.running_instance_count):
                 paasta_print("  {}.{} on {} isn't scaled up yet, "
                              "has {} out of {}"
--- a/tests/cli/test_cmds_wait_for_deployment.py
+++ b/tests/cli/test_cmds_wait_for_deployment.py
@@ -124,21 +124,21 @@ def test_instances_deployed(mock_get_paasta_api_client, mock__log):
     instances_out = Queue()
     f(cluster_data, instances_out, e)
     assert cluster_data.instances_queue.empty()
-    assert instances_out.get(block=True) == 'instance2'
+    assert instances_out.get(block=False) == 'instance2'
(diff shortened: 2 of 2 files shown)


solarkennedy:
Foiled by the tests that are somehow waiting forever...

solarkennedy:
@somic do you have time to investigate this? In theory this "one character" change will help improve the situation where users are needlessly waiting, but you will have to investigate how the tests are waiting.

solarkennedy:
Related to #1169

EvanKrall:
I can try to work on this later today, but if someone else has time first, please take it.

somic:
It appears to be timing out in instance3 block of test_instances_deployed

somic:
https://github.com/Yelp/paasta/blob/master/tests/cli/test_cmds_wait_for_deployment.py#L134

In instance3 block, instances_out queue is empty at the end so a blocking get() waits forever.

somic:
This fixes tests but I haven't wrapped my head around the code enough yet to say if it's the right fix:

```
diff --git i/tests/cli/test_cmds_wait_for_deployment.py w/tests/cli/test_cmds_wait_for_deployment.py
index 3f790f6..ff59a8f 100644
--- i/tests/cli/test_cmds_wait_for_deployment.py
+++ w/tests/cli/test_cmds_wait_for_deployment.py
@@ -131,7 +131,7 @@ def test_instances_deployed(mock_get_paasta_api_client, mock__log):
     instances_out = Queue()
     f(cluster_data, instances_out, e)
     assert cluster_data.instances_queue.empty()
-    assert instances_out.get(block=True) == 'instance3'
+    assert instances_out.empty()

     cluster_data.instances_queue = Queue()
     cluster_data.instances_queue.put('instance4')
```

somic:
@solarkennedy @EvanKrall ^^ - I want to make sure this doesn't just fix the tests but ensures correct behavior.

solarkennedy:
Hmm. I really don't know. @oktopuz you wrote these tests, can you help answer if why my change in the diff would require the change in the tests that @somic proposes?

oktopuz:
I suspect I used 'block=True' accidentally, it should be 'block=False' everywhere in tests.
@somic, given your change, you're right the queue should be empty in this case.

Could you replace 'block=False' with 'block=True' for all other cases in test_instances_deployed() as well?

somic:
I pushed my changes to this PR. @oktopuz please take a look

solarkennedy:
I approve

oktopuz:
lgtm

PR #1176, opened 2017-05-02
Kill unhealthy instances first

Rob-Johnson:
@mattmb I've got tests to fix and real testing to do - just thought I'd get the code up for review.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -26,6 +26,7 @@ import boto3
 from botocore.exceptions import ClientError
 from requests.exceptions import HTTPError
 
+from paasta_tools.autoscaling import ec2_fitness
 from paasta_tools.mesos_maintenance import drain
 from paasta_tools.mesos_maintenance import undrain
 from paasta_tools.mesos_tools import get_mesos_master
@@ -108,6 +109,32 @@ class ClusterAutoscaler(ResourceLogMixin):
         instance_reservations = [reservation['Instances'] for reservation in instance_descriptions['Reservations']]
         instances = [instance for reservation in instance_reservations for instance in reservation]
         return instances
+
+    def describe_instance_status(self, instance_ids, region=None, instance_filters=None):
+        """This wraps ec2.describe_instance_status and catches instance not
+        found errors. It returns a list of instance description
+        dictionaries.  Optionally, a filter can be passed through to
+        the ec2 call
+
+        :param instance_ids: a list of instance ids, [] means all
+        :param instance_filters: a list of ec2 filters
+        :param region to connect to ec2
+        :returns: a list of instance description dictionaries"""
+        if not instance_filters:
+            instance_filters = []
+        ec2_client = boto3.client('ec2', region_name=region)
+        try:
+            instance_descriptions = ec2_client.describe_instance_status(
+                InstanceIds=instance_ids,
+                Filters=instance_filters
+            )
+        except ClientError as e:
+            if e.response['Error']['Code'] == 'InvalidInstanceID.NotFound':
+                self.log.warn('Cannot find one or more instance from IDs {}'.format(instance_ids))
+                return None
+            else:
+                raise
+        return instance_descriptions
 
     def get_instance_ips(self, instances, region=None):
         instance_descriptions = self.describe_instances([instance['InstanceId'] for instance in instances],
@@ -234,15 +261,6 @@ class ClusterAutoscaler(ResourceLogMixin):
                 else:
                     raise
 
-    def sort_slaves_to_kill(self, slaves):
-        """Pick the best slaves to kill. This returns a list of slaves
-        after sorting in preference of which slaves we kill first.
-        It sorts first by number of chronos tasks, then by total number of tasks
-
-        :param slaves: list of slaves dict
-        :returns: list of slaves dicts"""
-        return sorted(slaves, key=lambda x: (x.task_counts.chronos_count, x.task_counts.count))
-
     def scale_resource(self, current_capacity, target_capacity):
         """Scales an AWS resource based on current and target capacity
         If scaling up we just set target capacity and let AWS take care of the rest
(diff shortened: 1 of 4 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -323,17 +341,63 @@ def filter_aws_slaves(self, slaves_list):
...
+                    'Name': 'private-ip-address',
+                    'Values': slave_ips
+                }
+            ]
+        )
         instance_type_weights = self.get_instance_type_weights()
-        slave_instances = [PaastaAwsSlave(slave, instance_descriptions, instance_type_weights) for slave in slaves]
-        return slave_instances
+        paasta_aws_slaves = []
+        for slave in slaves:
```
mattmb: This loop feels like it should be split into its own function, just to keep things small and easier to test.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -109,6 +110,32 @@ def describe_instances(self, instance_ids, region=None, instance_filters=None):
...
+
+        :param instance_ids: a list of instance ids, [] means all
+        :param instance_filters: a list of ec2 filters
+        :param region to connect to ec2
+        :returns: a list of instance description dictionaries"""
+        if not instance_filters:
+            instance_filters = []
+        ec2_client = boto3.client('ec2', region_name=region)
+        try:
+            instance_descriptions = ec2_client.describe_instance_status(
```
mattmb: We didn't need this call before right? But I guess it makes it possible to do the "filtering" outside of `PaastaAwsSlave`?

Rob-Johnson:
@mattmb can you help me out with this last test? I'm particularly stumped understanding what you're doing with [this side effect](https://github.com/Yelp/paasta/blob/master/tests/autoscaling/test_autoscaling_cluster_lib.py#L912-L913).

paasta_tools/autoscaling/autoscaling_cluster_lib.py
Rob-Johnson: yeah, we kind of do a join on the data we get from this API call and the describe_instances call

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -321,26 +340,76 @@ def gracefully_terminate_slave(self, slave_to_kill, current_capacity, new_capaci
...
+                if len(matching_descriptions) > 1:
+                    log.error("Found more than one instance with the same private IP {0}. "
+                              "This should never happen")
+                    raise ClusterAutoscalingError
+                description = matching_descriptions[0]
+                matching_status = [
+                    status for status in instance_statuses
+                    if status['InstanceId'] == description['InstanceId']
+                ]
+                assert len(status) == 1, "There should be only one InstanceStatus per instance"
```
solarkennedy: I kinda like the quick and easy assertions for those things that really are bugs and not real life error conditions, feel free to do the same for 377

paasta_tools/autoscaling/autoscaling_cluster_lib.py
solarkennedy: yes please. 
It kinda looks like there is one thing that correlates ips, one thing that correlates status's, and one thing that correlates descriptions?

And I think filter_aws_slaves is doing all three

mattmb:
@Rob-Johnson I think the idea of that side effect is it returns a list of instances but each time it returns the list with 1 fewer instance. i.e. it is simulating the fact that each time one will have been terminated and removed from the list. I'll have a quick poke at the failure and if it's easy I'll see if I can make it pass but we might be better to pick this up when we can pair on it?

mattmb:
Looks like you've changed that side effect in your branch @Rob-Johnson, undoing the change makes the tests pass:
```
diff --git a/tests/autoscaling/test_autoscaling_cluster_lib.py b/tests/autoscaling/test_autoscaling_cluster_lib.py
index 1bf3aec8..2b11cfa6 100644
--- a/tests/autoscaling/test_autoscaling_cluster_lib.py
+++ b/tests/autoscaling/test_autoscaling_cluster_lib.py
@@ -930,10 +930,8 @@ class TestClusterAutoscaler(unittest.TestCase):
             mock_get_mesos_task_count_by_slave.reset_mock()
             mock_sort_slaves_to_kill.reset_mock()
             mock_sfr_sorted_slaves = [mock_slave_1] * 10
-            mock_sort_slaves_to_kill.side_effect = [
-                [mock_sfr_sorted_slaves] + [mock_sfr_sorted_slaves[x:-1]]
-                for x in range(0, 10)
-            ]
+            mock_sort_slaves_to_kill.side_effect = [mock_sfr_sorted_slaves] + \
+                [mock_sfr_sorted_slaves[x:-1] for x in range(0, 10)]
             mock_get_mesos_task_count_by_slave.return_value = [
                 {'task_counts': mock_slave_1}
                 for x in range(0, 9)
```

Rob-Johnson:
ty @mattmb

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -321,26 +340,92 @@ def gracefully_terminate_slave(self, slave_to_kill, current_capacity, new_capaci
...
-                                                        instance_filters=[{'Name': 'private-ip-address',
-                                                                           'Values': slave_ips}])
         instance_type_weights = self.get_instance_type_weights()
-        slave_instances = [PaastaAwsSlave(slave, instance_descriptions, instance_type_weights) for slave in slaves]
-        return slave_instances
+        instance_statuses = self.instance_status_for_ips(slave_ips)
+        instance_descriptions = self.instance_descriptions_for_ips(slave_ips)
+
+        paasta_aws_slaves = []
+        for slave in slaves:
```
solarkennedy: Would this be any better if there was a function called

    get_ec2_description_for_mesos_slave

that was a memoized version of what you have here in lines 354-372

then the code would just need to iterate over each slave, but call the function in line 377 ?


I think that would simplify the code for readability sake, yet still allow it to be fast maybe?

Rob-Johnson: erm, maybe? though the function will only ever be called in one place, and I don't think it'll be any faster, given it's called once per run per mesos slave.

mattmb:
Just to capture what we said in meatspace:
* If we stop filtering use slave pid to IP then we need to have another way to filter our describe instances call, can't we do it by instance ID or tag?
* Then we can make the sorting prefer things that aren't registered at all and except cases when they need mesos data that isn't present.

We agree that this should be tackled in a separate PR

PR #1178, opened 2017-05-02
Adding some unit tests for deployd

mattmb:


paasta_tools/deployd/watchers.py
```diff
@@ -3,6 +3,7 @@
 
 import logging
 import time
+from functools import reduce
```
nhandler: Is this fixing some import error? If not, where is the new import being used?

mattmb: Yeah, they moved it in python3 :-(

This seems to make it work for both!

PR #1179, opened 2017-05-02
Make 'paasta push-to-registry' refuse to overwrite an existing image

oktopuz:
internal ticket: PAASTA-9910

### DIFF
--- a/paasta_tools/cli/cmds/push_to_registry.py
+++ b/paasta_tools/cli/cmds/push_to_registry.py
@@ -18,13 +18,23 @@ image to a registry.
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import base64
+import os
+
+import requests
+import simplejson as json
+from requests.exceptions import RequestException
+
 from paasta_tools.cli.utils import get_jenkins_build_output_url
 from paasta_tools.cli.utils import validate_full_git_sha
 from paasta_tools.cli.utils import validate_service_name
+from paasta_tools.generate_deployments_for_service import build_docker_image_name
 from paasta_tools.utils import _log
 from paasta_tools.utils import _run
 from paasta_tools.utils import build_docker_tag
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/push_to_registry.py
```diff
@@ -83,6 +99,18 @@ def paasta_push_to_registry(args):
...
+        try:
+            if is_docker_image_already_in_registry(service, args.commit):
+                paasta_print("The docker image is already in the PaaSTA docker registry. "
+                             "I'm NOT overriding the existing image. "
+                             "Add --force to override the image in the registry if you are sure what you are doing.")
+                return 0
+        except RequestException as e:
+            paasta_print("Can not connect to the PaaSTA docker registry to verify if this image exists.\n"
+                         "%s" % str(e))
+            return 1
```
solarkennedy: This is the only thing I would be worried about. We could have a lot broken pipelines if the registry can't talk back in 5 seconds.

oktopuz: I'm increasing it to 30 seconds.
It is the same timeout as docker uses https://github.com/moby/moby/blob/master/distribution/registry.go#L65

paasta_tools/cli/cmds/push_to_registry.py
```diff
@@ -108,3 +136,43 @@ def paasta_push_to_registry(args):
             level='event',
         )
     return returncode
+
+
+def read_docker_registy_creds(registry_uri):
+    dockercfg_path = os.path.expanduser('~/.dockercfg')
```
chriskuehl: fwiw new versions of docker use `~/.docker/config.json` instead (though maybe still support this old path)

oktopuz: The old path is still valid: https://github.com/moby/moby/blob/master/cli/config/config.go#L18

PR #1180, opened 2017-05-02
add historical stats when creating chronos job

Rob-Johnson:
doing this will mean that the statistics for when a job ran will
be persisted between config pushes. why didn't we do this sooner?

### DIFF
--- a/paasta_tools/setup_chronos_job.py
+++ b/paasta_tools/setup_chronos_job.py
@@ -42,6 +42,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import argparse
+import copy
 import logging
 import sys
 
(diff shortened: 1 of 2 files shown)


paasta_tools/setup_chronos_job.py
```diff
@@ -161,6 +161,44 @@ def setup_job(service, instance, complete_job_config, client, cluster):
     )
 
 
+def config_with_historical_stats(chronos_client, service, instance, job_config):
+    """
+    Given a complete_job_config, return a new chronos_job_config
```
keymone: if i understand correctly this isn't returning new config, but modifying the one passed to the function. i would very much prefer it to return a copy but changing the docstring is fine too.

Rob-Johnson:
tested in the example cluster:

{code}
robj@robj-MacBookPro:~/workspace/paasta/example_cluster fetch-chronos-stats-before-deploy % while true; do curl -s localhost:8081/scheduler/jobs | jq '.[] | "\(.description) \(.successCount) \(.errorCount) \(.lastSuccess) \(.lastError)"'; sleep 5; done
"config685d1b94 325 0 2017-05-03T16:08:03.550Z "
"config685d1b94 325 0 2017-05-03T16:08:03.550Z "
"config685d1b94 325 0 2017-05-03T16:08:03.550Z "
"config685d1b94 325 0 2017-05-03T16:08:03.550Z "
"configfdd84b2d 325 0 2017-05-03T16:08:03.550Z "
"configfdd84b2d 325 0 2017-05-03T16:08:03.550Z "
{code}

note that the config hash has changed on the last 2 log lines (after a push), and yet the success count, error, count, lastSuccess and lastError all hang around.

paasta_tools/setup_chronos_job.py
Rob-Johnson: agreed.

PR #1181, opened 2017-05-02
Optimize list_marathon_service_instances -m

oktopuz:
Optimize the time it takes to run 'list_marathon_service_instances -m' by calling get_draining_hosts() only once.

Internal ticket: PAASTA-7989

With this change it takes ~30 seconds to run it in norcal-prod.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -35,7 +35,6 @@ from paasta_tools.long_running_service_tools import InvalidHealthcheckMode
 from paasta_tools.long_running_service_tools import load_service_namespace_config
 from paasta_tools.long_running_service_tools import LongRunningServiceConfig
 from paasta_tools.mesos.exceptions import NoSlavesAvailableError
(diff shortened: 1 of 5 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -999,15 +999,16 @@ def is_old_task_missing_healthchecks(task, marathon_client):
     return False
 
 
-def get_num_at_risk_tasks(app):
+def get_num_at_risk_tasks(app, draining_hosts=None):
```
solarkennedy: Can you make this a required parameter?

I don't like it when functions kinda do things on their own when it "should" be the responsibility of the caller to inject the dependency.

Additionally it will stop us from making this performance mistake again!

oktopuz: get_num_at_risk_tasks is used by setup_marathon_job.
Let me see..

oktopuz: Done.

PR #1182, opened 2017-05-02
Add --port option to local-run to let user specify port to run on

jglukasik:


paasta_tools/cli/cmds/local_run.py
```diff
@@ -492,6 +500,24 @@ def get_local_run_environment_vars(instance_config, port0, framework):
     return env
 
 
+def check_if_port_free(port):
```
Rob-Johnson: can you make this function just return True or False, and move the exception handling outside this function? otherwise *every* caller of this function will have to sys.exit if the port isn't free, which isn't ideal.

paasta_tools/cli/cmds/local_run.py
```diff
@@ -492,6 +500,16 @@ def get_local_run_environment_vars(instance_config, port0, framework):
...
+    temp_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+    try:
+        temp_socket.bind(("127.0.0.1", port))
+    except socket.error:
+        return False
+    temp_socket.close()
```
Rob-Johnson: you might want to put this cleanup in a ``finally`` block just in case?

PR #1183, opened 2017-05-03
A new autoscaling policy that scales number of containers proportionally, and can be extended to incorporate load predictions.

EvanKrall:


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -168,6 +179,102 @@ def clamp_value(number):
...
+
+def save_historical_load(historical_load, zk_path_prefix):
+    with ZookeeperPool() as zk:
+        historical_load_bytes = serialize_historical_load(historical_load)
+        zk.ensure_path(zk_historical_load_path(zk_path_prefix))
+        zk.set(zk_historical_load_path(zk_path_prefix), historical_load_bytes)
+
+
+def serialize_historical_load(historical_load):
+    max_records = 1000000 // SIZE_PER_HISTORICAL_LOAD_RECORD
```
matfra: Would a size like that put a lot of pressure on zk ?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -387,13 +494,14 @@ def get_autoscaling_info(marathon_client, service, instance, cluster, soa_dir):
...
-                                               setpoint=autoscaling_params.pop('setpoint'),
+                                               setpoint=autoscaling_params['setpoint'],
                                                current_instances=service_config.get_instances())
             new_instance_count = get_new_instance_count(utilization=utilization,
                                                         error=error,
                                                         autoscaling_params=autoscaling_params,
                                                         current_instances=service_config.get_instances(),
-                                                        marathon_service_config=service_config)
+                                                        marathon_service_config=service_config,
+                                                        num_healthy_instances=len(marathon_tasks))
```
matfra: Is len(marathon_tasks) really is the number of healthy instances or is it the number of instances that we have planned to run (without any guarantee that they are all healthy) ?

EvanKrall: marathon_tasks comes from `filter_autoscaling_tasks`, which returns only the tasks that are healthy.

paasta_tools/autoscaling/autoscaling_service_lib.py
EvanKrall: If there's a lot of churn, yes, but I talked with @jolynch and he said this seemed fine.

Between Marathon (which has lots of data) and SmartStack (which does a lot of churn) we're already hitting our ZKs pretty hard anyway, and we've tuned them to snapshot less frequently so we don't fill up our disks.

EvanKrall:
Aw, man, how did I miss the unit test failure? That's embarassing.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -168,6 +179,102 @@ def clamp_value(number):
...
+                            near future.
+    """
+
+    forecast_policy_func = get_forecast_policy(forecast_policy)
+
+    current_load = (utilization - offset) * num_healthy_instances
+
+    historical_load = fetch_historical_load(zk_path_prefix=zookeeper_path)
+    historical_load.append((time.time(), current_load))
+    save_historical_load(historical_load, zk_path_prefix=zookeeper_path)
```
mattmb: I guess this writes to zookeeper? I suspect it might get called when paasta status is run now that we include the autoscaling info. Is that a problem?

EvanKrall: Yeah, I realized this when testing on stage. I think I might pull the saving of historical load into a separate cron job, so that we can control the resolution of data separately from the frequency of autoscaler runs.

PR #1184, opened 2017-05-03
Blacklist offers from slaves with stuck tasks

keymone:
- maintain set of blacklisted slaves
- add slave ids where tasks get stuck
- unblacklist slaves after one hour

### DIFF
--- a/paasta_tools/frameworks/native_scheduler.py
+++ b/paasta_tools/frameworks/native_scheduler.py
@@ -116,6 +116,10 @@ class NativeScheduler(mesos.interface.Scheduler):
         # Gets set when registered() is called
         self.framework_id = None
 
+        self.blacklisted_slaves = set()
+        self.blacklisted_slaves_lock = threading.Lock()
+        self.blacklist_timeout = 3600
+
         if service_config is not None:
             self.service_config = service_config
             self.service_config.config_dict.update(self.service_config_overrides)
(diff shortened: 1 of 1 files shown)


paasta_tools/frameworks/native_scheduler.py
```diff
@@ -150,6 +154,16 @@ def resourceOffers(self, driver, offers):
             for offer in offers:
                 driver.declineOffer(offer.id)
         else:
+            for idx, offer in enumerate(offers):
+                if offer.slave_id.value in self.blacklisted_slaves:
+                    log.critical("Ignoring offer %s from blacklisted slave %s" %
+                                 (offer.id.value, offer.slave_id.value))
+                    driver.declineOffer(offer.id)
+                    del offers[idx]
```
Rob-Johnson: can you even do this? Isn't offers a list? I'd do something like this:

```python
acceptable_offers = []
for offer in (offers):
      if offer.slave_id.value in self.blacklisted_slaves:
          log.critical("Ignoring offer %s from blacklisted slave %s" %
                          (offer.id.value, offer.slave_id.value))
          driver.declineOffer(offer.id)
      else:
          acceptable_offers.append(offer)

self.launch_tasks_for_offers(driver, acceptable_offers)
```

keymone: yep, del works on list just fine

Rob-Johnson: oh, right, of course. it's an index.

PR #1185, opened 2017-05-03
remote-run -j | --instances <number>

keymone:
run multiple copies of remote-run task when -j (--instances) flag passed. default to 1.


Rob-Johnson:
got a test?

keymone:
@Rob-Johnson working on it

PR #1186, opened 2017-05-04
Make logging better for deployd

mattmb:
Include timestamp and module name

### DIFF
--- a/paasta_tools/deployd/common.py
+++ b/paasta_tools/deployd/common.py
@@ -15,7 +15,7 @@ class PaastaThread(Thread):
 
     @property
     def log(self):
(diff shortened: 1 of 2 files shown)


solarkennedy:
do we really need a timestamp at log-output time? 
If you are just debugging in the foreground you can pipe to `ts`, but otherwise the timestamps should be in syslog?

mattmb:
That's a good point actually, I have a ticket to set it up with syslog...

Let me try that first and then I'll close this if it works.

mattmb:
Yup, I've added the syslog handler in the last commit and it works well.

PR #1188, opened 2017-05-04
Add maintenance + public config watchers

mattmb:
This should really be two commits but I screwed up. Tests are
incoming...

### DIFF
--- a/paasta_tools/deployd/watchers.py
+++ b/paasta_tools/deployd/watchers.py
@@ -11,10 +11,20 @@ from kazoo.recipe.watchers import ChildrenWatch
 from kazoo.recipe.watchers import DataWatch
 
 from paasta_tools.deployd.common import PaastaThread
+from paasta_tools.deployd.common import rate_limit_instances
 from paasta_tools.deployd.common import ServiceInstance
 from paasta_tools.long_running_service_tools import AUTOSCALING_ZK_ROOT
 from paasta_tools.marathon_tools import DEFAULT_SOA_DIR
+from paasta_tools.marathon_tools import deformat_job_id
+from paasta_tools.marathon_tools import get_all_marathon_apps
+from paasta_tools.marathon_tools import get_marathon_client
+from paasta_tools.marathon_tools import load_marathon_config
+from paasta_tools.marathon_tools import load_marathon_service_config
+from paasta_tools.mesos_maintenance import get_draining_hosts
+from paasta_tools.utils import get_services_for_cluster
 from paasta_tools.utils import list_all_instances_for_service
+from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.utils import PATH_TO_SYSTEM_PAASTA_CONFIG_DIR
 
 
 class PaastaWatcher(PaastaThread):
@@ -80,10 +90,10 @@ class AutoscalerWatcher(PaastaWatcher):
             time.sleep(0.1)
 
 
-class FileWatcher(PaastaWatcher):
+class SoaFileWatcher(PaastaWatcher):
 
     def __init__(self, inbox_q, cluster, **kwargs):
-        super(FileWatcher, self).__init__(inbox_q, cluster)
+        super(SoaFileWatcher, self).__init__(inbox_q, cluster)
         self.wm = pyinotify.WatchManager()
         self.wm.add_watch(DEFAULT_SOA_DIR, self.mask, rec=True)
         self.notifier = pyinotify.Notifier(watch_manager=self.wm,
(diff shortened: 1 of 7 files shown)


mattmb:
This now has unit tests and is ready to review

paasta_tools/deployd/watchers.py
```diff
@@ -45,6 +55,130 @@ def startup_checker(self, obj):
...
+                self.filewatcher.inbox_q.put(service_instance)
+
+    def get_service_instances_with_changed_id(self):
+        marathon_app_ids = [app.id.lstrip('/') for app in get_marathon_apps()]
+        instances = get_services_for_cluster(cluster=self.public_config.get_cluster(),
+                                             instance_type='marathon',
+                                             soa_dir=DEFAULT_SOA_DIR)
+        service_instances = []
+        for service, instance in instances:
+            config = load_marathon_service_config(service,
```
solarkennedy: kwargs please

paasta_tools/deployd/watchers.py
```diff
@@ -45,6 +55,130 @@ def startup_checker(self, obj):
...
+                return
+            service_instances = []
+            if new_config != self.public_config:
+                self.log.info("Public config has changed, now checking if it affects any services config shas")
+                self.public_config = new_config
+                service_instances = self.get_service_instances_with_changed_id()
+            if service_instances:
+                self.log.info("Found config change affecting {} service instances, "
+                              "now doing a staggered bounce".format(len(service_instances)))
+                service_instances = rate_limit_instances(service_instances,
```
solarkennedy: kwargs please

paasta_tools/deployd/watchers.py
```diff
@@ -45,6 +55,130 @@ def startup_checker(self, obj):
...
+
+    def filter_event(self, event):
+        if event.name.endswith('.json'):
+            return event
+
+    def watch_new_folder(self, event):
+        if event.maskname == 'IN_CREATE|IN_ISDIR':
+            self.filewatcher.wm.add_watch(event.pathname, self.filewatcher.mask, rec=True)
+
+    def process_default(self, event):
```
solarkennedy: can you name this something different? 

how about `process_event` ?

paasta_tools/deployd/watchers.py
```diff
@@ -45,6 +55,130 @@ def startup_checker(self, obj):
...
+            if service_instances:
+                self.log.info("Found config change affecting {} service instances, "
+                              "now doing a staggered bounce".format(len(service_instances)))
+                service_instances = rate_limit_instances(service_instances,
+                                                         self.public_config.get_deployd_big_bounce_rate(),
+                                                         self.__class__.__name__)
+            for service_instance in service_instances:
+                self.filewatcher.inbox_q.put(service_instance)
+
+    def get_service_instances_with_changed_id(self):
```
solarkennedy: Doesn't this duplicate what the "oracle" (list_marathon_service_instances -m) does?
Can you reuse or dedupe that code?

paasta_tools/deployd/watchers.py
```diff
@@ -45,6 +55,130 @@ def startup_checker(self, obj):
...
+
+    def get_at_risk_service_instances(self, draining_hosts):
+        marathon_apps = get_marathon_apps()
+        at_risk_tasks = [task for app in marathon_apps for task in app.tasks if task.host in draining_hosts]
+        self.log.info("At risk tasks: {}".format(at_risk_tasks))
+        service_instances = []
+        for task in at_risk_tasks:
+            app_id = task.app_id.strip('/')
+            service, instance, _, __ = deformat_job_id(app_id)
+            service_instances.append(ServiceInstance(service=service,
```
solarkennedy: Is there an issue of adding duplicates here? (like if there are 4 tasks at risk, all for the same app, we only need emit a bounce once right?)

paasta_tools/deployd/watchers.py
```diff
@@ -45,6 +55,130 @@ def startup_checker(self, obj):
...
+    def run(self):
+        self.notifier.loop(callback=self.startup_checker)
+
+    def startup_checker(self, obj):
+        self.is_ready = True
+
+
+def get_marathon_apps():
+    marathon_config = load_marathon_config()
+    marathon_client = get_marathon_client(marathon_config.get_url(), marathon_config.get_username(),
```
solarkennedy: won't this be happening a lot?
Can you do the cached marathon client trick we use in other places?

paasta_tools/deployd/watchers.py
mattmb: That's a good point. There is some de-duping in the `Inbox` logic which probably means we are okay...

It might be sane to de-dupe here though anyway. I think it would be clearer that way...

paasta_tools/deployd/watchers.py
mattmb: Hmm, I guess quite a bit: at most once per 20s but iff a host has been marked for draining since the last run. I'll think about whether we can live with slightly stale data here and if we can I'll try and import the cached client...

paasta_tools/deployd/watchers.py
mattmb: That's the name in the super class `pyinotify.ProcessEvent` unfortunately

paasta_tools/deployd/watchers.py
mattmb: *kwa*rgs ;-)

paasta_tools/deployd/watchers.py
mattmb: I hadn't thought of that. The `list_marathon_service_instances` code looks a bit different because it's checking for things that the other watchers cover (e.g. autoscaling)

Given that this will hopefully replace the oracle method I think it's probably not worth trying to refactor them to both share some code.

mattmb:
@Rob-Johnson I think it would be sane to split it as a submodule as long as it's easy to make the import code work in the main thread. I'll give that a go once I've merged all my open PRs (to avoid any more conflicts).

itests are definitely on my todo list. Currently they only cover the YelpSoa watcher and are flaking on travis (so skipped). I'm planning to work a separate branch to fix them and cover more of the watchers.

paasta_tools/deployd/watchers.py
mattmb: The ttl is actually also 20s https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_tools.py#L555

So unless we make it configurable it won't do much yet. I think I'm going to need to do quite a bit of work on caching soon so I think I'll leave this be for now.

paasta_tools/deployd/watchers.py
```diff
@@ -45,6 +55,135 @@ def startup_checker(self, obj):
...
+            for service_instance in service_instances:
+                self.filewatcher.inbox_q.put(service_instance)
+
+    def get_service_instances_with_changed_id(self):
+        marathon_app_ids = [app.id.lstrip('/') for app in get_marathon_apps()]
+        instances = get_services_for_cluster(cluster=self.public_config.get_cluster(),
+                                             instance_type='marathon',
+                                             soa_dir=DEFAULT_SOA_DIR)
+        service_instances = []
+        for service, instance in instances:
```
solarkennedy: This sorta overlaps with this
https://github.com/Yelp/paasta/blob/master/paasta_tools/list_marathon_service_instances.py#L91

this is a pretty expensive operation, luckily it will only happen on public config changes

PR #1193, opened 2017-05-04
wait-for-deployment will no longer block on Deploying or Waiting #1169

somic:
Fixes the second bullet point in #1169

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -386,7 +386,7 @@ def _run_instance_worker(cluster_data, instances_out, green_light):
                 cluster_data.instances_queue.task_done()
                 instances_out.put(instance)
                 continue
-            if status.marathon.deploy_status != 'Running':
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -386,7 +386,7 @@ def _run_instance_worker(cluster_data, instances_out, green_light):
                 cluster_data.instances_queue.task_done()
                 instances_out.put(instance)
                 continue
-            if status.marathon.deploy_status != 'Running':
+            if status.marathon.deploy_status not in ['Running', 'Deploying', 'Waiting']:
```
mattmb: I can see an argument for removing "Waiting".

I'm not 100% sure about deploying: https://mesosphere.github.io/marathon/docs/marathon-ui.html#deploying

I guess we still check the number of instances and that we only have the one correct marathon app running...

I guess technically an autoscaling event would put us in a "Deploying" state when really it's just scaling?

solarkennedy: Correct

PR #1194, opened 2017-05-05
Task proc scheduler

Rob-Johnson:
First part of porting remote-run to depend on the task processing module: take the parts of the long running scheduler into the task processing scheduler. From here we'll want to try and reduce the dependencies on the rest of the paasta codebase, and make the AdhocScheduler depend on the NativeScheduler in the task_processing module, rather than the paasta_tools one.


task_processing/task_processing/framework.py
```diff
@@ -1,225 +1,397 @@
...
+import time
+import uuid
+from threading import Timer
 
 import mesos.interface
-import mesos.native
+import service_configuration_lib
 from mesos.interface import mesos_pb2
 
+from paasta_tools.frameworks.constraints import check_offer_constraints
```
keymone: i'd really like this file to not import a single thing from paasta_*

Rob-Johnson: we'll get there - this is just a first pass. I'm not going to do everything in one PR

keymone: shipit!

PR #1196, opened 2017-05-05
docker wrapper can choose, flock, and apply random mac address

bchess:
Internal ticket PAASTA-10888, for CEP763

If PAASTA_FIREWALL environment variable is set to something non-empty and command is a 'docker run', the docker wrapper will assign a unique MAC address to the underlying `docker run` command.

The MAC address chosen is guaranteed unique on the node by holding an flock on a file in /var/run/paasta/mac-address. The first two bytes of the MAC are 02:52 so as to be local but not interfere with Docker's own [MAC address choosing-algorithm](https://github.com/docker/libnetwork/blob/75756e47120db1a2464d7b3efa693526955196f2/netutils/utils.go#L97). The rest of the bytes are random. We make 100 attempts to choose a random address before giving up.

If an error occurs (either /var/run/paasta/mac-address does not exist or is not writable, or a random mac address cannot be chosen), the wrapper prints a message to stderr before exec()-ing the underlying docker exe with no changes to its arguments.

On its own, this doesn't really do much - it just specifies a randomly chosen MAC address, whereas docker would also choose a random MAC address anyway. Future steps are to apply iptables rules prior to exec-ing docker.


paasta_tools/docker_wrapper.py
```diff
@@ -65,15 +73,41 @@ def can_add_hostname(args):
...
 
+    return False
+
+
+def is_run(args):
+    try:
+        list(args).index('run')
+        return True
+    except ValueError:
+        return False
```
chriskuehl: i think this probably isn't totally correct but it might be a microbug that we never see in practice?

e.g. if "run" happens to appear somewhere in the command line of another command, like `docker exec <container> some-command run`

tests/test_docker_wrapper.py
```diff
@@ -599,3 +623,81 @@ def _patch_docker_wrapper_dependencies(self, is_numa_enabled, cpu_info, node_mem
...
+        assert mock_execlp.mock_calls == [mock.call(
+            'docker',
+            'docker',
+            'run',
+            '--mac-address=12:34:56:78:90:ab',
+            '--env=PAASTA_FIREWALL=1',
+        )]
+
+    @mock.patch('sys.stderr', new_callable=six.StringIO, autospec=None)
+    def test_mac_address_no_lockdir(self, mock_stderr, mock_execlp, tmpdir):
```
chriskuehl: there's a `capsys` pytest fixture which might be slightly cleaner than mocking sys.stderr: https://docs.pytest.org/en/latest/capture.html

paasta_tools/docker_wrapper.py
bchess: Yeah its true, though it's also an existing problem. Our arg parsing is not sophisticated enough to differentiate between which args are for top-level docker args vs the ones that get passed to individual commands.

paasta_tools/mac_address.py
```diff
@@ -0,0 +1,53 @@
...
+
+    raise MacAddressException('Unable to pick unique MAC address')
+
+
+def obtain_lock(lock_filepath):
+    """ Open and obtain a flock on the parameter. Returns a file if successful, None if not
+    """
+    lock_file = open(lock_filepath, 'w')
+    try:
+        fcntl.flock(lock_file, fcntl.LOCK_EX | fcntl.LOCK_NB)
```
solarkennedy: Will this really lock anything if we simply "exec" docker  later?

Will anything clean these up?

bchess: This path is invoked if PAASTA_FIREWALL env variable is enabled (in docker_wrapper.py) so it doesn't affect everything automatically. If that's what you're asking?

Nothing sweeps up old lockfiles atm, we'll need a tmpwatch or cronjob or something.

bchess: The fd (and flock) will be held open through an exec, so unless the subsequent process explicitly closes the fd, it will remain held and locked.

PR #1197, opened 2017-05-05
connect to random master for remote-run

keymone:
distribute the load across available connectable masters


paasta_tools/cli/utils.py
```diff
@@ -419,11 +420,14 @@ class NoMasterError(Exception):
     pass
 
 
-def connectable_master(cluster, system_paasta_config):
+def connectable_master(cluster, system_paasta_config, random_master=False):
```
Rob-Johnson: is there any reason to parameterise other than being defensive?

keymone: nope

paasta_tools/cli/utils.py
```diff
@@ -419,11 +420,14 @@ class NoMasterError(Exception):
...
     masters, output = calculate_remote_masters(cluster, system_paasta_config)
     if masters == []:
         raise NoMasterError('ERROR: %s' % output)
 
+    if random_master and len(masters) > 1:
+        random.shuffle(masters)
```
solarkennedy: just do random.shuffle(), it works on things of length 1

paasta_tools/cli/utils.py
solarkennedy: don't bother parameterizing, just do it

paasta_tools/cli/utils.py
```diff
@@ -582,7 +585,8 @@ def execute_chronos_rerun_on_remote_master(service, instancename, cluster, syste
 
 
 def run_on_master(cluster, system_paasta_config, cmd_parts,
-                  timeout=None, err_code=-1, graceful_exit=False, stdin=None):
+                  timeout=None, err_code=-1, graceful_exit=False, stdin=None,
+                  random_master=False):
```
solarkennedy: don't bother with the random_master arg if we don't need it (I'm pretty sure we don't)

paasta_tools/cli/utils.py
```diff
@@ -438,7 +441,7 @@ def check_ssh_and_sudo_on_master(master, timeout=10):
     with sudo to verify that ssh and sudo work properly. Return a tuple of the
     success status (True or False) and any output from attempting the check.
     """
-    check_command = 'ssh -A -n -o StrictHostKeyChecking=no %s sudo paasta_serviceinit -h' % master
+    check_command = 'ssh -A -n -o StrictHostKeyChecking=no %s sudo /bin/true' % master
```
solarkennedy: At yelp we whitelist what commands some users can sudo, and we don't currently give "prod" users the ability to run sudo /bin/true,

Just do /bin/true (I guess)

keymone: makes sense, but the function name specifically says check_ssh_and_sudo so i assume it's only used in places where sudo is expected?

solarkennedy: Yea don't bother with sudo anymore, I don't think we really need it

PR #1198, opened 2017-05-06
[WIP] Task processing: fleshing out the interface

keymone:
(this is based off of Robj's branch)

it looks like a lot of changes, but really isn't. order in which to look at files: `task_processing/runner.py` -> `task_processing/task_processing/task_executor.py` -> `task_processing/task_processing/execution_framework.py`

this is a little step away from the declarative approach we've been taking up til now, where we'd specify what and how many times we want to run and let framework do the rest. i think we should decouple the declarative interface from imperative and expose the imperative first, so that people wanting to use framework asap are unblocked earlier (but also because there is no reason not to expose imperative interface).

## File renames:

- `docker_task_configs.py` -> `/dev/null`, replaced by namedtuple in task_executor.py
- `framework.py` -> `execution_framework.py` to align with terminology we used in CEP
- `mesos_task_execution.py` -> `task_executor.py` for the same reason (sort of)
- `runner.py` moved up a folder because it's not really a part of library

### TaskConfig

Is a blueprint for task to run, it is owned by library users and passed to TaskExecutor every time a task has to be launched (as opposed to passing to constructor).

### TaskExecutor

Is the main point of entry. library users will instantiate it once and feed it tasks to run. instantiation optionally would take arguments like parallelization factor, retry policy, etc and pass those on to instance of ExecutionFramework.

library users interact with TaskExecutor by calling these methods (explanations in the code): 

- run(task_config) blocks until task is no longer alive [src](https://github.com/Yelp/paasta/compare/task-proc-scheduler...task-proc-contd?expand=1#diff-23822790ac1ddcfc86284fc4f702aeadR33)
- run_promise(task_config) returns task_id and a promise(or is it a future?) [src](https://github.com/Yelp/paasta/compare/task-proc-scheduler...task-proc-contd?expand=1#diff-23822790ac1ddcfc86284fc4f702aeadR39)
- run_async(task_config, callbacks...) schedules a task for execution and optionally registers callbacks [src](https://github.com/Yelp/paasta/compare/task-proc-scheduler...task-proc-contd?expand=1#diff-23822790ac1ddcfc86284fc4f702aeadR44)
- wait_until_done self-explanatory

### ExecutionFramework

current code is full of paastaisms, they will go away soonish and hopefully we'll be left with some prototype code of actual framework that knows how to keep track of tasks and place them according to offers and constraints

library users shouldn't need to directly interact with execution framework though they can

TaskExecutor uses these methods:

- enqueue(task_config) to schedule task for execution [src](https://github.com/Yelp/paasta/compare/task-proc-scheduler...task-proc-contd?expand=1#diff-de0a2640ec27873366f44c0b09a08336R342)

- wait_for_task(task_id) blocking call to wait until task is no longer active [src](https://github.com/Yelp/paasta/compare/task-proc-scheduler...task-proc-contd?expand=1#diff-de0a2640ec27873366f44c0b09a08336R345)

- various on_xxx(task_id) methods to register callback for specific events [src](https://github.com/Yelp/paasta/compare/task-proc-scheduler...task-proc-contd?expand=1#diff-de0a2640ec27873366f44c0b09a08336R351)

## thoughts?

people who (unlike me) know python will probably have some valid criticisms over the listed functions or even approach in general - i'd love to hear them.

@Rob-Johnson @EvanKrall @jolynch @solarkennedy @sagar8192 @huadongliu


task_processing/task_processing/execution_framework.py
```diff
@@ -395,3 +338,222 @@ def unblacklist_slave(self, slave_id):
...
+                                 (offer.id.value, offer.slave_id.value))
+                    driver.declineOffer(offer.id)
+                    del offers[idx]
+
+            if len(offers) == 0:
+                return
+
+            self.launch_tasks_for_offers(driver, offers)
+
+    def statusUpdate(self, driver, update):
```
keymone: sure. i just didn't want to invest time into something that we might choose not to merge.

PR #1199, opened 2017-05-08
Fix tags being null when checking registry

mattmb:
If the image isn't in the registry docker returns something like:
``{"name":"hello-world","tags":null}``

This was causing an exception when `is_docker_image_already_in_registry`
should be returning False. This affects new services and blocks them
from being able to push-to-registry.

### DIFF
--- a/paasta_tools/cli/cmds/push_to_registry.py
+++ b/paasta_tools/cli/cmds/push_to_registry.py
@@ -172,7 +172,11 @@ def is_docker_image_already_in_registry(service, sha):
     with requests.Session() as s:
         r = s.get(url, timeout=30) if creds[0] is None else s.get(url, auth=creds, timeout=30)
         if r.status_code == 200:
-            return tag in r.json()['tags']
+            tags_resp = r.json()
+            if tags_resp['tags']:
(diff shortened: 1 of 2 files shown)


oktopuz:
lgtm.
Thank you for fixing this.

paasta_tools/cli/cmds/push_to_registry.py
```diff
@@ -172,7 +172,11 @@ def is_docker_image_already_in_registry(service, sha):
     with requests.Session() as s:
         r = s.get(url, timeout=30) if creds[0] is None else s.get(url, auth=creds, timeout=30)
         if r.status_code == 200:
-            return tag in r.json()['tags']
+            tags_resp = r.json()
```
solarkennedy: Isn't this just 

    return tag in r.json().get('tags', [])

oktopuz: Unfortunately it isn't.

```
>>> {'tags': None}.get('tags', []) is None
True
```

solarkennedy: Ohh

PR #1201, opened 2017-05-08
ensure adhoc scheduler service config has cmd

keymone:
lets fail fast, when trying to load config rather than after registering the framework and building a task

### DIFF
--- a/paasta_tools/frameworks/adhoc_scheduler.py
+++ b/paasta_tools/frameworks/adhoc_scheduler.py
@@ -17,6 +17,7 @@ from __future__ import unicode_literals
 
 from paasta_tools.frameworks.native_scheduler import LIVE_TASK_STATES
 from paasta_tools.frameworks.native_scheduler import NativeScheduler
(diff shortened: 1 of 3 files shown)


paasta_tools/frameworks/adhoc_scheduler.py
```diff
@@ -60,3 +61,7 @@ def tasks_and_state_for_offer(self, driver, offer, state):
 
     def kill_tasks_if_necessary(self, *args, **kwargs):
         return
+
+    def validate_config(self):
```
sagar8192: "validate_config" implies that we are validating the entire config. Since we are only checking if cmd exists or not, can we rename this to something like "check_if_cmd_exists"?

keymone: i don't know, the purpose of validate_config is to validate config, it just happens that we only have a single thing we want to validate atm. if we decide to extend it later - it will include more checks.

PR #1203, opened 2017-05-09
Create interface for subscribing to task events

Rob-Johnson:
first implementation of the event bus idea, that allows users to subscribe to events. Creates a ``TaskProcessingEvent`` abstraction around Mesos's ``StatusUpdate``, but exposes it via the ``_original_event`` var.


Rob-Johnson:
closed in favour of #1210

PR #1204, opened 2017-05-09
Switch to a human-readable serialization format for ZK load data

EvanKrall:


mattmb:
Did you catch my comment on #1183 @EvanKrall ?

I left if after it was merged unfortunately. I have a feeling we might update this data when someone runs paasta status because `get_autoscaling_info` uses the same decision policy code to work out what the autoscaler will do next...

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -261,51 +261,95 @@ def moving_average_forecast_policy(historical_load, moving_average_window_second
...
-        except NoNodeError:
-            return []
+        historical_load = []
+        for version in versions:
+            try:
+                historical_load_bytes, _ = zk.get(zk_historical_load_path(zk_path_prefix, version=version))
+                historical_load.extend(deserialize_historical_load(historical_load_bytes, version=version))
+            except NoNodeError:
+                pass
+        return sorted(historical_load)
```
solarkennedy: Are you sure it is ok for this list to have multiple versions inside of it? Can the consumers tolerate that?

EvanKrall: The deserialized data format is still the same.

EvanKrall:
Discarding, as this is a lot of complexity for a feature I don't care much about anymore; plus I've been using the repl to achieve this:

```
>>> from paasta_tools.autoscaling.autoscaling_service_lib import *
>>> load = fetch_historical_load('/autoscaling/yelp-main/main_useast1')
>>> print('\n'.join('%s %s' % x for x in load[-100:]))
```

PR #1205, opened 2017-05-09
sensu alert triggered when the security-check fails

transcedentalia:


paasta_tools/cli/cmds/security_check.py
```diff
@@ -47,6 +51,10 @@ def perform_security_check(args):
         return 0
 
     ret_code, output = _run(security_check_command, timeout=300, stream=True)
-    paasta_print("{} exited with code {}".format(security_check_command, ret_code))
+    paasta_print("{} exited with code {}:{}".format(security_check_command, ret_code, output))
+
+    sensu_status = pysensu_yelp.Status.CRITICAL if ret_code else pysensu_yelp.Status.OK
```
solarkennedy: Is this right? This looks backwards or maybe should be if `ret_code != 0` or something

paasta_tools/cli/cmds/security_check.py
```diff
@@ -47,6 +51,10 @@ def perform_security_check(args):
...
+    send_event(service=args.service, check_name='%s.security_check' % args.service,
+               overrides={'page': False, 'ticket': True}, status=sensu_status, output=output, soa_dir=DEFAULT_SOA_DIR)
```
solarkennedy: What do you think about dogfooding this first and sending all these tickets to the security team?

transcedentalia: I'm not very happy with this. I would prefer that every team will get a ticket created in its Jira project, go through the runbook and try to fix it. If they get stuck they could ask myself or the security team for help. I don't expect to many security-check failures though.

paasta_tools/cli/cmds/security_check.py
transcedentalia: I think if ret_code and if ret_code != 0 are equivalent (except for when ret_code is None), but I could use the latter for more clarity.

paasta_tools/cli/cmds/security_check.py
```diff
@@ -47,6 +51,10 @@ def perform_security_check(args):
         return 0
 
     ret_code, output = _run(security_check_command, timeout=300, stream=True)
-    paasta_print("{} exited with code {}".format(security_check_command, ret_code))
+    paasta_print("{} exited with code {}:{}".format(security_check_command, ret_code, output))
```
transcedentalia: Updated. Hope it's better now.

paasta_tools/cli/cmds/security_check.py
```diff
@@ -47,6 +51,11 @@ def perform_security_check(args):
         return 0
 
     ret_code, output = _run(security_check_command, timeout=300, stream=True)
-    paasta_print("{} exited with code {}".format(security_check_command, ret_code))
+    paasta_print("The security-check failed with {}. Please visit the security-check runbook "
```
mattmb: Technically it might not have failed right?

transcedentalia: true, updating.

PR #1210, opened 2017-05-10
Initial implementation of task-processor

Rob-Johnson:
@sagar8192 @jolynch want to take a look over this? the [example](https://github.com/Yelp/paasta/compare/runnable-task?expand=1#diff-dbe18ab3b0f1b695dcd0ac587d7dd8b8) gives a view to how the end user will use to interact with the library programatically.

Some of the important parts of the design are:

### Runner

We've created an abstraction, ``Runner``, which wraps around dispatching to the underlying task executor. This is the 'top level' object developers will use to interact with their tasks. Implementations of the runner provide different concurrency wrappers around the execution of the task. The initial implementations we've made are:

- `Async`: Executes user provided callbacks on different events types
- `Subscription`: Proxies all events from the executor back to the user via a parameterised queue.
- `SyncRunner`: Blocks completion of the `run` function until terminal event.
- `PromiseRunner` (TODO): Immediately returns a future object that the user can later block on or cancel

### TaskExecutor

A `TaskExecutor` are platform specific implementations for running tasks. Of course, we've only got an implementation for Mesos for now, but if we were to ever go in another direction, it'd mean us implementing another `TaskExecutor` for that platform. `Runner` objects take an `TaskExecutor` object in their construction.

### Event 

We provide wrapper classes for the underlying mesos `TaskStatus` updates. These are the events proxied out of the execution framework, which takes a `translator` function for converting native mesos updates into `EventXXX` objects.

As I said, we've left examples to show how this all ties together. It's a bit rough around the edges, but it works locally. We'll be working on unit tests + integration tests to prove it.


task_processing/executors/task_executor.py
```diff
@@ -0,0 +1,43 @@
...
+import six
+
+TaskConfig = namedtuple(
+    'TaskConfig',
+    ['image', 'cmd', 'cpus', 'mem', 'disk', 'volumes', 'ports', 'cap_add',
+     'ulimit', 'docker_parameters'],
+)
+
+
+def make_task_config(image="ubuntu:xenial", cmd="/bin/true", cpus=0.1,
```
jolynch: If possible I'd like to avoid doing what paasta has done with method -> type constructors. Can we re-use an existing solution like https://github.com/python-attrs/attrs or https://pypi.python.org/pypi/namedlist/1.4?

task_processing/executors/task_executor.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import abc
+from collections import namedtuple
+
+import six
+
+TaskConfig = namedtuple(
```
jolynch: This seems very coupled. Perhaps we can register these with the framework or something (plugins)

task_processing/executors/task_executor.py
```diff
@@ -0,0 +1,47 @@
...
+    'TaskConfig',
+    ['task_id', 'name', 'image', 'cmd', 'cpus', 'mem', 'disk', 'volumes',
+     'ports', 'cap_add', 'ulimit', 'docker_parameters'],
+)
+
+
+def make_task_config(task_id=None, name='foo', image="ubuntu:xenial",
```
jolynch: I'd like to avoid this anti-pattern I've seen in PaaSTA where we want defaults so we make methods and wrapper classes everywhere. What if we just used something off the shelf like attrs (https://github.com/python-attrs/attrs)?

task_processing/executors/mesos_executor.py
```diff
@@ -0,0 +1,62 @@
...
+            mesos_address,
+            False,
+        )
+
+        # start driver thread immediately
+        self.driver_thread = threading.Thread(target=self.driver.run, args=())
+        self.driver_thread.daemon = True
+        self.driver_thread.start()
+
+    def run(self, task_config):
```
jolynch: run taking task_config is leaking the interface IMO

it would be much cleaner IMO if we had a totally decoupled implementation, the MultiMesosExecutor, which defined it's own method add_task which took the arguments you need.

task_processing/executors/mesos_executor.py
```diff
@@ -0,0 +1,62 @@
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import threading
+
+import mesos.native
+from mesos.interface import mesos_pb2
+
+from task_processing.events.event import mesos_status_to_event
+from task_processing.executors.task_executor import TaskExecutor
```
jolynch: What's the difference between an ExecutionFramework and a TaskExecutor?

task_processing/executors/task_executor.py
```diff
@@ -0,0 +1,47 @@
...
+        ulimit = []
+    if docker_parameters is None:
+        docker_parameters = []
+
+    return TaskConfig(task_id, name, image, cmd, cpus, mem, disk, volumes,
+                      ports, cap_add, ulimit, docker_parameters)
+
+
+@six.add_metaclass(abc.ABCMeta)
+class TaskExecutor(object):
```
jolynch: Can we keep the interfaces in a separate module from the pieces that are plugins (like mesos), that way we can do extra_requires and folks can install the interface without installing mesos

task_processing/runners/runner.py
```diff
@@ -0,0 +1,18 @@
...
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import abc
+
+import six
+
+
+@six.add_metaclass(abc.ABCMeta)
+class Runner(object):
```
jolynch: I'm not sure I understand the full difference between runner and taskexecutor

Let's docstring it?

Also can we move this to a separate top level module?

e.g.

```
task_processing/
├── interface
│   ├── task_config.py
│   ├── task_executor.py
│   └── task_runner.py
└── plugins
    ├── mesos
    │   ├── mesos_executor.py
    │   └── multi_mesos_executor.py
    └── registry.py
```

This way we can use extra_requires to make it so folks can depend on the interface without depending on the mesos implementation details
```

File: task_processing/executors/task_executor.py (hunk 2 above)
Comment (keymone) [inline task_processing/executors/task_executor.py]: you mean coupled to specific containerization mechanism? task_config is the raw input into our runners, so we should definitely think hard about the shape of it, currently we're at the proof of concept stage.

File: task_processing/executors/mesos_executor.py (hunk 5 above)
Comment (keymone) [inline task_processing/executors/mesos_executor.py]: in this iteration TaskExecutor sets up the mesos driver and execution framework and wires them together, it holds the driver thread and is basically a gateway to anything platform specific. users instantiating a certain executor don't need to know the underlying interactions, also it would be a bit circular if execution framework owned the driver, because the driver sort of owns the execution framework.

File: task_processing/runners/runner.py (hunk 7 above)
Comment (keymone) [inline task_processing/runners/runner.py]: originally runner was called concurrency policy, they exist to give users choice of concurrency style they want to interact in. task executor is probably poorly named - it just wires together framework and the driver, and has less platform-specific interface (takes dicts instead of mesos_pb objects, etc).

on folder structure - i agree that keeping all platform-specific stuff in plugin folders looks better.

can you elaborate on multi_mesos_executor.py?

File: task_processing/runners/sync.py [hunk 8]
```diff
@@ -0,0 +1,22 @@
...
+
+from six.moves.queue import Queue
+
+from task_processing.runners.runner import Runner
+
+
+class Sync(Runner):
+    def __init__(self, executor):
+        self.executor = executor
+        self.queue = Queue.queue()
```
keymone: that's fixed already

task_processing/runners/sync.py
```diff
@@ -0,0 +1,22 @@
...
+
+    def run(self, task_config):
+        task_id = self.executor.run(task_config)
+        event_queue = self.executor.get_event_queue()
+
+        while True:
+            event = event_queue.get()
+            if event.task_id == task_id and event.is_finished():
+                return
```
keymone: that's fixed already

task_processing/executors/mesos_executor.py
```diff
@@ -0,0 +1,56 @@
...
+            "127.0.0.1:5050",
+            False,
+        )
+
+        # start driver thread immediately
+        self.driver_thread = threading.Thread(target=self.driver.run, args=())
+        self.driver_thread.daemon = True
+        self.driver_thread.start()
+
+    def run(self, task_config):
```
keymone: we went for task_id being part of task_config for now. task_config is still in flux so that might change later on.

task_processing/runners/async.py
```diff
@@ -0,0 +1,35 @@
...
+        callback_t.daemon = True
+        callback_t.start()
+
+    def run(self, task_config):
+        return self.executor.run(task_config)
+
+    def _callback_loop(self):
+        event_queue = self.executor.get_event_queue()
+        while True:
+            event = event_queue.get()
```
sagar8192: Since self.executor.get_event_queue() is gonna be shared between multiple objects Async/Sync/Subscription/Promise, the task update that we get in this queue may not be for this object. We need to check if the task update is for this object or not. If not, we must re-enqueue this update to self.executor.get_event_queue(). We also need locking to make sure that only one Async object is getting the lock. Maybe we don't need locking because of GIL? ¯\_(ツ)_/¯

PR #1211, opened 2017-05-10
Implement a forecaster that uses linear regression over a trailing window to estimate slope and level of the load.

EvanKrall:


solarkennedy:
Are you sure this is a good idea? The "inertia" and "statefulness" was one of the original reasons that we wanted to move away from the PID controller, and now this change makes it so that the past influences the future.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -243,22 +243,72 @@ def current_value_forecast_policy(historical_load, **kwargs):
...
 @register_autoscaling_component('moving_average', FORECAST_POLICY_KEY)
 def moving_average_forecast_policy(historical_load, moving_average_window_seconds, **kwargs):
     """Does a simple average of all historical load data points within the moving average window. Weights all data
     points within the window equally."""
 
-    window_end, _ = historical_load[-1]
-    window_begin = window_end - moving_average_window_seconds
+    windowed_data = trailing_window_historical_load(historical_load, moving_average_window_seconds)
+    windowed_values = [value for timestamp, value in windowed_data]
+    return sum(windowed_values) / len(windowed_values)
```
solarkennedy: Is there a chance for divide by zero here?

EvanKrall: no; historical_load always has the most recent load measurement, which would fall within the window.

EvanKrall:
The problem with the inertia with the PID controller was that it made the future depend on _how badly we'd done at autoscaling_ in the past, rather than past load; so if we were overscaled for a while, then the PID controller would underscale for a while to make up for it.

PR #1214, opened 2017-05-10
Increase timeout in mesos_cpu_metrics_provider

oktopuz:
and work properly when it times out.

Internal ticket: PAASTA-11068


oktopuz:
closing in favor of #1215

PR #1215, opened 2017-05-10
When data is missing for a task, ignore it, instead of having it skew utilization downwards.

EvanKrall:


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -450,13 +450,20 @@ def mesos_cpu_metrics_provider(marathon_service_config, marathon_tasks, mesos_ta
     monkey.patch_socket()
     jobs = [gevent.spawn(task.stats_callable) for task in mesos_tasks]
     gevent.joinall(jobs, timeout=10)
-    mesos_tasks = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs]))
+    mesos_tasks = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs if job.value is not None]))
```
oktopuz: If you exclude None here, zip() can mess everything:

```
>>> zip(['a', 'b', 'c', 'd'], [1, 2, 3, 4])
[('a', 1), ('b', 2), ('c', 3), ('d', 4)]
>>> zip(['a', 'b', 'c', 'd'], [3, 4])
[('a', 3), ('b', 4)]
```

I fixed it a bit differently and increased timeout as well: #1214

EvanKrall: Good catch.

PR #1217, opened 2017-05-11
Don't log when bounce "doesn't do anything"

jglukasik:
If the new app is running, and there are no more old tasks, no need to log anything

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -290,26 +290,26 @@ def do_bounce(
     all_old_tasks = set.union(all_old_tasks, *old_app_draining_tasks.values())
     all_old_tasks = set.union(all_old_tasks, *old_app_at_risk_tasks.values())
 
-    # log if we appear to be finished
-    if all([
-        (apps_to_kill or tasks_to_kill),
-        apps_to_kill == list(old_app_live_happy_tasks),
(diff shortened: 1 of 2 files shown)


paasta_tools/setup_marathon_job.py
```diff
@@ -306,6 +306,8 @@ def log_bounce_action(line, level='debug'):
             level='event',
         )
         return None
+    elif (not all_old_tasks) and new_app_running:
+        return None
     else:
         return 60
```
EvanKrall: I think this whole section might be better arranged as:

    if all_old_tasks or (not new_app_running):
        # still more work to do
        return 60

    if all([
        (apps_to_kill or tasks_to_kill),
        apps_to_kill == list(old_app_live_happy_tasks),
        tasks_to_kill == all_old_tasks,
    ]):
        log_bounce_action(
            line='%s bounce on %s finishing. Now running %s' %
            (
                bounce_method,
                serviceinstance,
                marathon_jobid
            ),
            level='event',
        )

    return None

Maybe with everything after `return 60` in an else block.

mattmb:
Thanks for looking at this!

paasta_tools/setup_marathon_job.py
mattmb: Yep, I think that would be sensible. Although the current diff will fix the issue so I'm happy if we just ship that too.

PR #1218, opened 2017-05-11
[taskproc] subscription runner and more

keymone:
- add all mesos event codes
- stop the runner in subscription example after all tasks terminated
- fix make_task_config caching task_id in def line
- lower blacklist timeout in mesos framework for testing purposes
- send TASK_LOST status update when detected stuck task so that we don't lose count of termination events
- re-queue task config in mesos framework if it didn't fit the current offer
- comment out some port shenanigans in framework
- make subscription runner stoppable (pull from event queue with timeout and check if we're stopping)


task_processing/events/event.py
```diff
@@ -38,3 +38,31 @@ class EventLost(EventTerminal):
 
 class EventStaging(EventBase):
     pass
+
```
jolynch: Let's add a todo to organize and explain these

task_processing/interfaces/task_executor.py
```diff
@@ -17,7 +17,16 @@
 
 def make_task_config(image="ubuntu:xenial", cmd="/bin/true", cpus=0.1,
```
jolynch: I'm concerned about this, let's add a TODO to use attrs or record classes or something?

task_processing/plugins/mesos/execution_framework.py
```diff
@@ -427,23 +437,23 @@ def create_new_docker_task(self, offer, task_config):
         docker.network = 2  # mesos_pb2.ContainerInfo.DockerInfo.Network.BRIDGE
         docker.force_pull_image = True
 
-        available_ports = []
-        for resource in offer.resources:
-            if resource.name == "ports":
-                available_ports = self.get_available_ports(resource)
+        # available_ports = []
```
jolynch: maybe todo this too?

task_processing/plugins/mesos/mesos_executor.py
```diff
@@ -55,8 +60,11 @@ def kill(self, task_id):
 
     def stop(self):
         self.execution_framework.stop()
+        log.debug("stopped the framework")
```
jolynch: nit: I think generally single quotes are preferred in general unless your string has single quotes in it

https://confluence.yelpcorp.com/display/ENG/Python+Style+Guidelines#PythonStyleGuidelines-Strings

keymone: ack

PR #1221, opened 2017-05-11
Package task processing

jolynch:
This sets us up as a separate package so we can publish separately, use separate tests, virtualenvs, etc ...

I'm not sure how to use different pre commit hooks, afaict we actually can't ... but I put it in there for the future.


keymone:
ship!

keymone:
still ship!

PR #1224, opened 2017-05-11
Move instance type to be a class function of our configs

solarkennedy:
The paasta codebase isn't very conducive to adding new instance types.

I would like slowly begin to help answer the question of "how would someone add a new instance_type to paasta"? be something like "implement this interface". One of the first interfaces to meet might be the "what kind of instance are you".

Right now we do a kinda awkward thing where we use this validate_service_instance function that happens to return the type. I think it will be more obvious if if we can just make that part of the class.

In the future I would like to move *more* things into the class object to make it easier for people to add new things for local-run, validate, status, etc.

### DIFF
--- a/paasta_tools/adhoc_tools.py
+++ b/paasta_tools/adhoc_tools.py
@@ -18,10 +18,10 @@ import logging
 
 import service_configuration_lib
 
-from paasta_tools.long_running_service_tools import LongRunningServiceConfig
 from paasta_tools.utils import deep_merge_dictionaries
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import get_paasta_branch
+from paasta_tools.utils import InstanceConfig
 from paasta_tools.utils import load_v2_deployments_json
 from paasta_tools.utils import NoConfigurationForServiceError
 from paasta_tools.utils import NoDeploymentsAvailable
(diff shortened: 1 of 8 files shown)


Rob-Johnson:
If we *have* to do this, can we just define constants instead of relying on magic strings?
``paasta_tools/utils.py``
```python
VALID_INSTANCE_TYPES=(CHRONOS_INSTANCE, MARATHON_INSTANCE, ADHOC_INSTANCE)
```

```paasta_tools/marathon_tools.py```
```python
MARATHON_TYPE='marathon'
class MarathonServiceConfig(): 
    ...
    def get_instance_type():
       return MARATHON_TYPE
```

```paasta_tools/elsewhere.py```
```python
def foo(config):
   instance_type = config.get_instance_type()
   assert type in VALID_INSTANCE_TYPES, "invalid instance type"
   if instance_type == MARATHON_TYPE:
       run_marathon_thing()
   elif...
````
On the other hand, why do we need to do this at all? Is doing

```python
instance_type = instance_config.get_instance_type()
if instance_type == "marathon":
   ...
```

really better than
```python

if isinstance(instance_config, MarathonServiceConfig):
   ....
```
? The information being encoded by the ```get_instance_type``` function is *already* encoded in the classname. We get nothing extra from what I can see?

solarkennedy:
Yea you are right, this in itself isn't very valuable.

PR #1225, opened 2017-05-11
Support adhoc.yaml in paasta check 

oktopuz:
internal ticket: PAASTA-7794

### DIFF
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -42,6 +42,7 @@ from paasta_tools.utils import _run
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import get_git_url
 from paasta_tools.utils import get_service_instance_list
+from paasta_tools.utils import INSTANCE_TYPES
 from paasta_tools.utils import is_deploy_step
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import paasta_print
@@ -222,6 +223,9 @@ def get_framework_steps(framework, service, soa_dir):
         'adhoc': load_adhoc_job_config,
     }
 
+    if framework not in load_config_func:
+        return []
+
     steps = []
     for cluster in list_clusters(service, soa_dir):
         for _, instance in get_service_instance_list(
@@ -248,10 +252,9 @@ def deployments_check(service, soa_dir):
     pipeline_steps = [step['step'] for step in pipeline_deployments]
     pipeline_steps = [step for step in pipeline_steps if is_deploy_step(step)]
 
-    frameworks = ['marathon', 'chronos', 'adhoc']
     framework_steps = {}
     in_deploy_not_frameworks = set(pipeline_steps)
-    for fr in frameworks:
+    for fr in INSTANCE_TYPES:
         framework_steps[fr] = get_framework_steps(fr, service, soa_dir)
         in_framework_not_deploy = set(framework_steps[fr]) - set(pipeline_steps)
         in_deploy_not_frameworks -= set(framework_steps[fr])
@@ -272,7 +275,7 @@ def deployments_check(service, soa_dir):
 
     if the_return is True:
         paasta_print(success("All entries in deploy.yaml correspond to a marathon, chronos or adhoc entry"))
-        for fr in frameworks:
+        for fr in INSTANCE_TYPES:
             if len(framework_steps[fr]) > 0:
                 paasta_print(success("All %s instances have a corresponding deploy.yaml entry" % fr))
     return the_return
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -185,7 +185,7 @@ def git_repo_check(service, soa_dir):
 
 
 def yaml_check(service_path):
-    """Check whether a marathon/chronos yaml file exists in service directory, and
+    """Check whether a marathon/chronos/adhoc yaml file exists in service directory, and
     print success/failure message(s).
 
     :param service_path: path to a directory containing the marathon/chronos yaml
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/check.py
```diff
@@ -256,37 +247,34 @@ def deployments_check(service, soa_dir):
...
-        paasta_print("%s There are some instance(s) you have asked to run in chronos that" % x_mark())
-        paasta_print("  do not have a corresponding entry in deploy.yaml:")
-        paasta_print("  %s" % PaastaColors.bold(", ".join(in_chronos_not_deploy)))
-        paasta_print("  You should probably configure these to use a 'deploy_group' or")
-        paasta_print("  add entries to deploy.yaml for them so they are deployed to those clusters.")
-        the_return = False
-    in_deploy_not_marathon_chronos = set(pipeline_steps) - set(marathon_steps) - set(chronos_steps)
-    if len(in_deploy_not_marathon_chronos) > 0:
+
+    frameworks = ['marathon', 'chronos', 'adhoc']
```
solarkennedy: Can you use utils.INSTANCE_TYPES ?

paasta_tools/cli/cmds/check.py
```diff
@@ -202,44 +203,34 @@ def yaml_check(service_path):
         paasta_print(PaastaCheckMessages.YAML_MISSING)
 
 
-def get_chronos_steps(service, soa_dir):
-    """This is a kind of funny function that gets all the chronos instances
-    for a service and massages it into a form that matches up with what
+def get_framework_steps(framework, service, soa_dir):
```
solarkennedy: How about

    get_deploy_groups_used_by_framework

paasta_tools/cli/cmds/check.py
```diff
@@ -202,44 +203,34 @@ def yaml_check(service_path):
...
     for cluster in list_clusters(service, soa_dir):
         for _, instance in get_service_instance_list(
                 service=service,
                 cluster=cluster,
-                instance_type='marathon',
+                instance_type=framework,
                 soa_dir=soa_dir,
         ):
-            config = load_marathon_service_config(
+            config = load_config_func[framework](
```
solarkennedy: Isn't this the `get_instance_config` function?

paasta_tools/cli/cmds/check.py
oktopuz: This will definitely make this function look better, however then we will subtract deploy_groups from steps in deployments_check().

paasta_tools/cli/cmds/check.py
oktopuz: It is, but I don't think that reusing it here makes a lot of sense. At this point we already know which instance type is that, while get_instance_config() reads configs one more time just to determine it (https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/utils.py#L666).

paasta_tools/cli/cmds/check.py
oktopuz: Done.

paasta_tools/cli/cmds/check.py
oktopuz: Renamed steps to deploy_groups.

paasta_tools/cli/cmds/check.py
```diff
@@ -202,44 +203,34 @@ def yaml_check(service_path):
         paasta_print(PaastaCheckMessages.YAML_MISSING)
```
solarkennedy: can you adjust this to account for adhoc-only services?

paasta_tools/cli/cmds/check.py
solarkennedy: It makes a lot of sense to not duplicate functionality in the codebase and to reduce the scope of this function though.

How about this as a compromise: make `get_instance_config` take an optional `instance_type` arg, and if none it will look it up. And here we pass it in.

I'm trying to reduce the future effort of adding new instance_types, having the logic about how to read them in in only one function will help.

paasta_tools/cli/cmds/check.py
oktopuz: done

oktopuz:
This is complete and ready for review.

paasta_tools/cli/cmds/check.py
```diff
@@ -198,95 +198,81 @@ def yaml_check(service_path):
...
+                    instance=instance,
+                    cluster=cluster,
+                    soa_dir=soa_dir,
+                    load_deployments=False,
+                    instance_type=instance_type,
+                )
+                steps.append(config.get_deploy_group())
+            except NotImplementedError:
+                pass
     return steps
```
solarkennedy: this is our chance to "get this right", please call these "deploy_groups", as that is what they are.

paasta_tools/cli/cmds/check.py
```diff
@@ -198,95 +198,81 @@ def yaml_check(service_path):
...
-        paasta_print("  add entries to deploy.yaml for them so they are deployed to those clusters.")
-        the_return = False
-    in_deploy_not_marathon_chronos = set(pipeline_steps) - set(marathon_steps) - set(chronos_steps)
-    if len(in_deploy_not_marathon_chronos) > 0:
+    pipeline_steps = [step['step'] for step in get_pipeline_config(service, soa_dir)]
+    pipeline_deploy_groups = [step for step in pipeline_steps if is_deploy_step(step)]
+
+    framework_deploy_groups = {}
+    in_deploy_not_frameworks = set(pipeline_deploy_groups)
+    for it in INSTANCE_TYPES:
```
solarkennedy: nice

paasta_tools/cli/utils.py
```diff
@@ -660,15 +662,17 @@ def get_jenkins_build_output_url():
...
+def get_instance_config(service, instance, cluster, soa_dir, load_deployments=False, instance_type=None):
     """ Returns the InstanceConfig object for whatever type of instance
     it is. (chronos or marathon) """
-    instance_type = validate_service_instance(
-        service=service,
-        instance=instance,
-        cluster=cluster,
-        soa_dir=soa_dir,
-    )
+    if instance_type is None:
```
solarkennedy: I promise this is a good thing.

Now we can actually speed things up by "injecting" what it is we are looking for instead of searching the filesystem each time.

PR #1234, opened 2017-05-16
Dependency & security support, and inotify-based synapse file watcher

bchess:
(I need to squash this before I actually merge)

This introduces support into paasta-tools InstanceConfig for the 'dependencies.yaml' file, and the 'security' dictionary within marathon-* and chronos-*.yaml files. 

Perhaps the biggest impacting change is to the chronos config loading. I make load_chronos_job_config to be more consistent with load_marathon_job_config - it now loads (and merges) the full general_config.

Using these changes I also introduce paasta_firewall_update, a CLI command that will watch the synapse services directory for changes to dependencies that occur on services that are running on this box. There are some TODOs in this file - all it does right now is log. Later we'll hook this up to something that updates iptable rules.

Internal ticket PAASTA-10896

### DIFF
--- a/paasta_itests/steps/setup_steps.py
+++ b/paasta_itests/steps/setup_steps.py
@@ -223,7 +223,7 @@ def write_soa_dir_marathon_job(context, job_id):
     if hasattr(context, "cmd"):
         soa[instance]['cmd'] = context.cmd
     with open(os.path.join(soa_dir, service, 'marathon-%s.yaml' % context.cluster), 'w') as f:
-        f.write(yaml.dump(soa))
+        f.write(yaml.safe_dump(soa))
     context.soa_dir = soa_dir
 
 
@@ -237,7 +237,7 @@ def write_soa_dir_native_service(context, job_id):
     if not os.path.exists(os.path.join(soa_dir, service)):
         os.makedirs(os.path.join(soa_dir, service))
     with open(os.path.join(soa_dir, service, 'paasta_native-%s.yaml' % context.cluster), 'w') as f:
-        f.write(yaml.dump({
+        f.write(yaml.safe_dump({
             "%s" % instance: {
                 'cpus': 0.1,
                 'mem': 100,
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -32,7 +32,9 @@ from six.moves.urllib_parse import urlsplit
 
 from paasta_tools import monitoring_tools
 from paasta_tools.mesos_tools import get_mesos_network_for_net
+from paasta_tools.mesos_tools import mesos_services_running_here
 from paasta_tools.tron import tron_command_context
+from paasta_tools.utils import deep_merge_dictionaries
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import get_config_hash
 from paasta_tools.utils import get_docker_url
@@ -182,7 +184,6 @@ class InvalidChronosConfigError(Exception):
 def read_chronos_jobs_for_service(service, cluster, soa_dir=DEFAULT_SOA_DIR):
     chronos_conf_file = 'chronos-%s' % cluster
     log.info("Reading Chronos configuration file: %s/%s/chronos-%s.yaml" % (soa_dir, service, cluster))
-
     return service_configuration_lib.read_extra_service_information(
         service,
         chronos_conf_file,
@@ -191,6 +192,12 @@ def read_chronos_jobs_for_service(service, cluster, soa_dir=DEFAULT_SOA_DIR):
 
 
 def load_chronos_job_config(service, instance, cluster, load_deployments=True, soa_dir=DEFAULT_SOA_DIR):
+    log.info("Reading general configuration file: service.yaml")
+    general_config = service_configuration_lib.read_service_configuration(
+        service,
+        soa_dir=soa_dir,
+    )
+
     service_chronos_jobs = read_chronos_jobs_for_service(service, cluster, soa_dir=soa_dir)
     if instance not in service_chronos_jobs:
         raise NoConfigurationForServiceError('No job named "%s" in config file chronos-%s.yaml' % (instance, cluster))
@@ -199,11 +206,14 @@ def load_chronos_job_config(service, instance, cluster, load_deployments=True, s
         deployments_json = load_deployments_json(service, soa_dir=soa_dir)
         branch = get_paasta_branch(cluster=cluster, instance=instance)
         branch_dict = deployments_json.get_branch_dict(service, branch)
+
+    general_config = deep_merge_dictionaries(overrides=service_chronos_jobs[instance], defaults=general_config)
+
     return ChronosJobConfig(
         service=service,
         cluster=cluster,
         instance=instance,
-        config_dict=service_chronos_jobs[instance],
+        config_dict=general_config,
         branch_dict=branch_dict,
     )
 
@@ -447,6 +457,8 @@ class ChronosJobConfig(InstanceConfig):
             'disk': self.check_disk,
             'schedule': self.check_schedule,
             'scheduleTimeZone': self.check_schedule_time_zone,
+            'security': self.check_security,
+            'dependencies_reference': self.check_dependencies_reference,
             'parents': self.check_parents,
             'cmd': self.check_cmd,
         }
@@ -505,7 +517,8 @@ class ChronosJobConfig(InstanceConfig):
         error_msgs.extend(super(ChronosJobConfig, self).validate())
 
         for param in ['epsilon', 'retries', 'cpus', 'mem', 'disk',
-                      'schedule', 'scheduleTimeZone', 'parents', 'cmd']:
+                      'schedule', 'scheduleTimeZone', 'parents', 'cmd',
+                      'security', 'dependencies_reference']:
             check_passed, check_msg = self.check(param)
             if not check_passed:
                 error_msgs.append(check_msg)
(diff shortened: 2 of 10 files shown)


bchess:
Oh service-configuration-lib 0.12.0 just landed on pypi so we may need to re-kick tests

tests/test_mac_address.py
```diff
@@ -66,7 +69,7 @@ def test_file_exists_exhaustion(tmpdir):
         with pytest.raises(mac_address.MacAddressException):
             mac_address.reserve_unique_mac_address(str(tmpdir))
     finally:
-        [p.kill() for p in flock_processes]
```
bchess: technically unrelated to this change, but I noticed that this test was leaving around tons of orphaned processes

paasta_itests/steps/setup_steps.py
```diff
@@ -223,7 +223,7 @@ def write_soa_dir_marathon_job(context, job_id):
     if hasattr(context, "cmd"):
         soa[instance]['cmd'] = context.cmd
     with open(os.path.join(soa_dir, service, 'marathon-%s.yaml' % context.cluster), 'w') as f:
-        f.write(yaml.dump(soa))
+        f.write(yaml.safe_dump(soa))
```
solarkennedy: Thank you. safe_* all the things!

paasta_tools/chronos_tools.py
```diff
@@ -191,6 +192,12 @@ def read_chronos_jobs_for_service(service, cluster, soa_dir=DEFAULT_SOA_DIR):
 
 
 def load_chronos_job_config(service, instance, cluster, load_deployments=True, soa_dir=DEFAULT_SOA_DIR):
+    log.info("Reading general configuration file: service.yaml")
```
solarkennedy: Is this really necessary?
You could put this at the debug level, but I say don't bother logging this

paasta_tools/utils.py
```diff
@@ -442,6 +477,36 @@ def get_volumes(self, system_volumes):
...
+    def get_dependencies(self):
+        """Get the contents of the dependencies_dict pointed to by the dependency_reference
+
+        Defaults to None if not specified in the config.
+
+        :returns: A list of dictionaries specified in the dependencies_dict, None if not specified"""
+        dependencies = self.config_dict.get('dependencies')
+        if not dependencies:
+            return None
+        return dependencies.get(self.get_dependencies_reference())
```
solarkennedy: Isn't this just

    self.config_dict.get('dependencies', None) ?

paasta_tools/utils.py
```diff
@@ -442,6 +477,36 @@ def get_volumes(self, system_volumes):
...
+    def get_outbound_firewall(self):
+        """Return 'block', 'monitor', or None as configured in security->outbound_firewall
+
+        Defaults to None if not specified in the config
+
+        :returns: A string specified in the config, None if not specified"""
+        security = self.config_dict.get('security')
+        if not security:
+            return None
+        return security.get('outbound_firewall')
```
solarkennedy: same

paasta_tools/firewall_update.py
```diff
@@ -0,0 +1,114 @@
...
+
+def setup_logging(verbose):
+    level = logging.DEBUG if verbose else logging.WARNING
+    logging.basicConfig(level=level)
+
+
+def run(args):
+    # Main loop waiting on inotify file events
+    inotify = Inotify(block_duration_s=1)  # event_gen blocks for 1 second
+    inotify.add_watch(args.synapse_service_dir.encode(), IN_MOVED_TO | IN_MODIFY)
```
solarkennedy: I think you are going to need to watch service.yaml too.

When it changes, you will need to *also* update firewall rules, but also you need to break the yaml cache for that (`service_configuration_lib._yaml_cache = {}`) otherwise it will read in a cached value.

solarkennedy: I don't know how you will update the service config object though

paasta_tools/chronos_tools.py
```diff
@@ -191,6 +192,12 @@ def read_chronos_jobs_for_service(service, cluster, soa_dir=DEFAULT_SOA_DIR):
...
+    general_config = service_configuration_lib.read_service_configuration(
```
solarkennedy: I wish we had a more universal place to put this and then dedupe code that does the same thing for chronos/marathon/adhoc/native/etc.

Suggestions welcome on better architecting this code so you won't have to copy paste this a bunch of times :)

paasta_tools/chronos_tools.py
```diff
@@ -956,3 +969,13 @@ def is_temporary_job(job):
...
     return job['name'].startswith(TMP_JOB_IDENTIFIER)
+
+
+def chronos_services_running_here():
+    """See what chronos services are being run by a mesos-slave on this host.
+    :returns: A list of triples of (service, instance, port)"""
+
+    return mesos_services_running_here(
+        framework_filter=lambda fw: fw['name'].startswith('chronos'),
+        parse_service_instance_from_executor_id=lambda task_id: decompose_job_id(task_id.split(MESOS_TASK_SPACER)[3])
```
solarkennedy: Heh. This seems really iffy. We don't really tightly control the implementation of how chronos task ids are generated :(

paasta_tools/firewall_update.py
```diff
@@ -0,0 +1,114 @@
...
+    for service, instance, port in marathon_services_running_here():
+        config = load_marathon_service_config(service, instance, cluster, load_deployments=False, soa_dir=soa_dir)
+        yield service, instance, config
+
+    for service, instance, port in chronos_services_running_here():
+        config = load_chronos_job_config(service, instance, cluster, load_deployments=False, soa_dir=soa_dir)
+        yield service, instance, config
+
+
+def smartstack_dependencies_of_running_firewalled_services(soa_dir=DEFAULT_SOA_DIR):
```
solarkennedy: This function is going to be expensive, as it has to iterate and load yamls for all these things, and it will run on every inotify event!

In the long term this is not very efficient. I don't have any great suggestions other than memoize/cache all the things, because we shouldn't have to re-read and re-parse these yamls all the time just because a synapse json file changed.

paasta_tools/chronos_tools.py
```diff
@@ -505,7 +517,8 @@ def validate(self):
         error_msgs.extend(super(ChronosJobConfig, self).validate())
 
         for param in ['epsilon', 'retries', 'cpus', 'mem', 'disk',
-                      'schedule', 'scheduleTimeZone', 'parents', 'cmd']:
+                      'schedule', 'scheduleTimeZone', 'parents', 'cmd',
```
solarkennedy: It kinda suck that this validate has to duplicate stuff. Can you put the security and deps in the InstanceConfig object so they are shared for all new types of things? (and not just chronos?)

paasta_tools/chronos_tools.py
bchess: ack. copy-pasted from load_marathon_job_config but yeah sounds fine

bchess: ack. copy-pasted from load_marathon_job_config but yeah sounds fine

paasta_tools/utils.py
bchess: not sure which part you're referring to... but this also handles the case of if config_dict['dependencies'] == None

paasta_tools/utils.py
bchess: same reply :)

paasta_tools/firewall_update.py
bchess: I thought we determined that service.yaml is part of the config hash, and thus changes to it result in bouncing the service?

paasta_tools/chronos_tools.py
bchess: true - isn't this the same for marathon too though? 

A completely different approach I could take would be to query the docker client for running containers, and then extract the service and instance name from env variables (sure would be nice to add labels). But marathon_services_running_here() was already the existing pattern so I stuck with that.

paasta_tools/firewall_update.py
bchess: It doesn't run on every inotify event. It runs at most every DEFAULT_UPDATE_SECS (5). We can adjust via cli option if it becomes a problem.

paasta_tools/chronos_tools.py
bchess: They are in the InstanceConfig. Oh, I just now noticed the "super(ChronosJobConfig, self).validate()" above. This should make this unnecessary. cpus and mem are also already there, so those should also be able to be removed.

paasta_tools/firewall_update.py
bchess: The time it takes is based on how many unique service-instances are running on the node. It has to query/load the mesos-slave json, then load the config files for each service-instance that's currently running. That doesn't seem too painful to me -- it doesn't seem inevitably bad - it doesn't get worse as we add more services overall or more traffic.

solarkennedy: Ok.

paasta_tools/chronos_tools.py
solarkennedy: It will be true for marathon, chronos, remote-run, spark, etc. 
In theory we have labels now! Keep doing this for now though, I guess just fyi about the limitations of this approach.

paasta_tools/firewall_update.py
solarkennedy: We only bounce services when the hash that json gets posted to marathon/chronos changes, it isn't a hash of the config object itself.

bchess:
squashing and rebasing onto master. if travis is green I plan on merging this

PR #1236, opened 2017-05-17
Add code for working with iptables

chriskuehl:
This adds code to create and manage iptables rules as part of the PaaSTA firewalls CEP.

The only public entrypoint right now is one function `general_update`, which could be run from a cronjob. It sets up standard rules (such as the `PAASTA` chain), then generates a list of desired rules, diffs them against existing rules, and makes any necessary adjustments. Currently it uses fake data for "list of service instances on this host" and "desired rules for the service".

In the future, we'll need to figure out the integration points between this code and other processes (e.g. the docker wrapper and the inotify watcher).

Some things that need to be done before merge:
* [x] tests
* [x] fix python3 tests
* [x] clean up `print` statements, add logging where appropriate
* [ ] (maybe) make it stop using fake data?

### Example usage

For now I'm just running the function manually and modifying the hard-coded test data.

#### Starting from a box with no iptables rules

It sets up the `PAASTA-INTERNET` chain, populates it, sets up an empty `PAASTA` chain, and sets up jumps to it in the appropriate locations. This is all a fancy noop.

```
$ sudo venv/bin/python -c 'from paasta_tools.firewall import general_update; general_update()'
creating chain: PAASTA-INTERNET
adding rule to PAASTA-INTERNET: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'0.0.0.0/0.0.0.0', target=u'ACCEPT', matches=())
adding rule to PAASTA-INTERNET: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'127.0.0.0/255.0.0.0', target=u'RETURN', matches=())
adding rule to PAASTA-INTERNET: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'10.0.0.0/255.0.0.0', target=u'RETURN', matches=())
adding rule to PAASTA-INTERNET: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'172.16.0.0/255.240.0.0', target=u'RETURN', matches=())
adding rule to PAASTA-INTERNET: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'192.168.0.0/255.255.0.0', target=u'RETURN', matches=())
adding rule to PAASTA-INTERNET: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'169.254.0.0/255.255.0.0', target=u'RETURN', matches=())
creating chain: PAASTA
adding rule to INPUT: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'0.0.0.0/0.0.0.0', target=u'PAASTA', matches=())
adding rule to FORWARD: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'0.0.0.0/0.0.0.0', target=u'PAASTA', matches=())
```

#### Launching a service instance

This creates a new chain for that service and populates it, then adds a jump on the `PAASTA` chain to that service's chain.

```
$ sudo venv/bin/python -c 'from paasta_tools.firewall import general_update; general_update()'
creating chain: PAASTA.cool-servi.003bae64d2
adding rule to PAASTA.cool-servi.003bae64d2: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'0.0.0.0/0.0.0.0', target=u'REJECT', matches=())
adding rule to PAASTA.cool-servi.003bae64d2: Rule(protocol=u'tcp', src=u'0.0.0.0/0.0.0.0', dst=u'169.254.255.254/255.255.255.255', target=u'ACCEPT', matches=((u'tcp', ((u'dport', u'20668'),)),))
adding rule to PAASTA: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'0.0.0.0/0.0.0.0', target=u'PAASTA.cool-servi.003bae64d2', matches=((u'mac', ((u'mac_source', u'02:42:A9:FE:00:02'),)),))
```

#### Adding another instance of that same service

It doesn't need to add another chain since it already has one for this service. Instead it just adds another jump for this new MAC address.

```
$ sudo venv/bin/python -c 'from paasta_tools.firewall import general_update; general_update()'
adding rule to PAASTA: Rule(protocol=u'ip', src=u'0.0.0.0/0.0.0.0', dst=u'0.0.0.0/0.0.0.0', target=u'PAASTA.cool-servi.003bae64d2', matches=((u'mac', ((u'mac_source', u'FE:A3:A3:DA:2D:50'),)),))
```

#### Other things

They pretty much work like you expect, so I didn't include example output.

## Future plans

* Add/improve integration points
* Locking of some kind around iptables updating
* Consider making some flows more efficient

(This PR started as https://github.com/chriskuehl/paasta/pull/1)

### DIFF
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -1,94 +1,117 @@
-#!/usr/bin/env python2.7
 from __future__ import absolute_import
 from __future__ import unicode_literals
 
+import collections
+import hashlib
 
-import re
-import shlex
-import subprocess
-
-
-class ChainDoesNotExist(Exception):
-    pass
+from paasta_tools import iptables
 
 
-def ensure_chain(chain, rules):
-    """Idempotently ensure a chain exists and has a set of rules.
+PRIVATE_IP_RANGES = (
+    '127.0.0.0/8',
+    '10.0.0.0/8',
+    '172.16.0.0/12',
+    '192.168.0.0/16',
+    '169.254.0.0/16',
+)
 
-    This function creates or updates an existing chain to match the rules
-    passed in.
 
-    This function will not reorder existing rules, but any new rules are always
-    inserted at the front of the chain.
+class ServiceGroup(collections.namedtuple('ServiceGroup', (
+    'service',
+    'dependency_group',
+    'mode',
+))):
+    """A service group.
+
+    :param service: the name of a service
+    :param dependency_group: the name of a dependency group
+    :param mode: either 'monitor' or 'block'
     """
-    try:
-        current_rules = list_chain(chain)
-    except ChainDoesNotExist:
-        create_chain(chain)
-        current_rules = set()
 
-    for rule in rules:
-        if rule not in current_rules:
-            print('adding rule: {}'.format(rule))
-            add_rule(chain, rule)
+    __slots__ = ()
 
-    for rule in current_rules:
-        if rule not in rules:
-            print('deleting rule: {}'.format(rule))
-            delete_rule(chain, rule)
+    @property
+    def chain_name(self):
+        """Return iptables chain name.
+
+        Chain names are limited to 28 characters, so we have to trim quite a
+        bit. To attempt to ensure we don't have collisions due to shortening,
+        we append a hash to the end.
+        """
+        chain = 'PAASTA.{}'.format(self.service[:10])
+        chain += '.' + hashlib.sha256(str(tuple(self))).hexdigest()[:10]
+        assert len(chain) <= 28, len(chain)
+        return chain
+
+    @property
+    def rules(self):
+        # TODO: actually read these from somewhere
+        return (
+            '-j REJECT --reject-with icmp-port-unreachable',
+            '-d 169.254.255.254/32 -p tcp -m tcp --dport 20666 -j ACCEPT',
+        )
+
+    def update_rules(self):
+        iptables.ensure_chain(self.chain_name, self.rules)
 
 
-def add_rule(chain, rule):
-    subprocess.check_call(['iptables', '-t', 'filter', '-I', chain] + shlex.split(rule))
+def active_service_groups():
+    """Return active service groups."""
+    # TODO: actually read these from somewhere
+    return {
+        ServiceGroup('cool-service', 'main', 'block'): {
+            '02:42:a9:fe:00:02',
+            'fe:a3:a3:da:2d:51',
+            'fe:a3:a3:da:2d:50',
+        },
+        ServiceGroup('cool-service', 'main', 'monitor'): {
+            'fe:a3:a3:da:2d:40',
+        },
+        ServiceGroup('dumb-service', 'other', 'block'): {
+            'fe:a3:a3:da:2d:30',
+            'fe:a3:a3:da:2d:31',
+        },
+    }
 
 
-def delete_rule(chain, rule):
-    subprocess.check_call(['iptables', '-t', 'filter', '-I', chain] + shlex.split(rule))
-
-
-def create_chain(chain):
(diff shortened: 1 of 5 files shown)


paasta_tools/firewall.py
```diff
@@ -0,0 +1,153 @@
...
+        },
+        ServiceGroup('dumb-service', 'other', 'block'): {
+            'fe:a3:a3:da:2d:30',
+            'fe:a3:a3:da:2d:31',
+        },
+    }
+
+
+def general_update():
+    """Update iptables to match the current PaaSTA state."""
```
solarkennedy: I think this function would be more readable if it did different steps. It would make it more self-documenting:

```
def general_update()
    self.ensure_internet_chain()
   active_chains = self.ensure_service_group_chains()
   self.ensure_dispatch_chains() # should this be first?
   self.delete_old_chains(active_chains=active_chains) # is this really a class method?
   ```

File: paasta_tools/firewall.py [hunk 2]
```diff
@@ -0,0 +1,153 @@
...
+        target='PAASTA',
+        matches=(),
+    )
+    iptables.ensure_rule('INPUT', jump_to_paasta)
+    iptables.ensure_rule('FORWARD', jump_to_paasta)
+
+    # Garbage collect any no-longer-needed service group chains.
+    paasta_chains = {
+        chain
+        for chain in iptables.all_chains()
```
solarkennedy: Is this safe? because won't it have all chains from *other* services too? (more than the ones for this class)

paasta_tools/firewall.py
chriskuehl: I think you're correct. I already started moving some out in order to make it easier to test; I'll continue that.

paasta_tools/firewall.py
chriskuehl: It's only considering chains that start with `PAASTA.` to potentially be garbage collected (the next line), so I think this is safe. We're assuming we "own" any chain that starts with `PAASTA.`.

chriskuehl:
@bchess now with tests!

Unfortunately I've had to do quite a bit of mocking in some of the tests (since obviously we don't want to be changing real iptables rules, and even doing `iptc.Table('filter')` raises an exception about requiring root, but overall I think this branch is in fairly good shape and ready for serious review.

Still need to fix the Python 3 tests and clean up print statements at least.

paasta_tools/iptables.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def ensure_rule(chain, rule):
+    rules = list_chain(chain)
+    if rule not in rules:
+        insert_rule(chain, rule)
+
+
+def insert_rule(chain_name, rule):
+    print('adding rule to {}: {}'.format(chain_name, rule))
```
bchess: use logging

tests/test_firewall.py
```diff
@@ -0,0 +1,170 @@
...
+    dst='0.0.0.0/0.0.0.0',
+    target=None,
+    matches=(),
+)
+
+
+@pytest.fixture
+def service_group():
+    return firewall.ServiceGroup(
+        service='my-cool-service',
```
bchess: i think paasta services can't have dashes

bchess: (does iptables support underscores ok?)

bchess:
overall looks good. fix n ship

tests/test_firewall.py
chriskuehl: good catch. yup, underscores are fine.

PR #1240, opened 2017-05-18
Make deployd check marathon ID before bouncing

mattmb:
As with the public config it doesn't make much sense to add things to
the bounce queue if their sha has not changed. This adds the same check
to changes to marathon-* files and deployments.json.

In addition I have disabled the yaml cache on the service configuration
lib since it was interfering with checking the sha. I'm pretty sure
we're not reading the configs frequently enough to need a cache.

### DIFF
--- a/paasta_tools/deployd/watchers.py
+++ b/paasta_tools/deployd/watchers.py
@@ -18,6 +18,7 @@ from paasta_tools.marathon_tools import DEFAULT_SOA_DIR
 from paasta_tools.marathon_tools import deformat_job_id
 from paasta_tools.marathon_tools import get_all_marathon_apps
 from paasta_tools.marathon_tools import get_marathon_client
+from paasta_tools.marathon_tools import list_all_marathon_app_ids
 from paasta_tools.marathon_tools import load_marathon_config
 from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.mesos_maintenance import get_draining_hosts
@@ -134,17 +135,18 @@ class PublicConfigFileWatcher(PaastaWatcher):
         self.is_ready = True
 
 
-def get_marathon_apps():
+def get_marathon_client_from_config():
     marathon_config = load_marathon_config()
     marathon_client = get_marathon_client(marathon_config.get_url(), marathon_config.get_username(),
                                           marathon_config.get_password())
-    return get_all_marathon_apps(marathon_client, embed_failures=True)
+    return marathon_client
 
 
 class MaintenanceWatcher(PaastaWatcher):
     def __init__(self, inbox_q, cluster, **kwargs):
         super(MaintenanceWatcher, self).__init__(inbox_q, cluster)
         self.draining = []
+        self.marathon_client = get_marathon_client_from_config()
 
     def run(self):
         self.is_ready = True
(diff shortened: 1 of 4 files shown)


paasta_tools/deployd/watchers.py
```diff
@@ -225,20 +229,20 @@ def process_default(self, event):
...
-                                                  instance=instance,
-                                                  cluster=self.public_config.get_cluster(),
-                                                  soa_dir=DEFAULT_SOA_DIR)
-            if config.format_marathon_app_dict()['id'] not in marathon_app_ids:
-                service_instances.append((service, instance))
-        return service_instances
+
+def get_service_instances_with_changed_id(instances, cluster):
+    marathon_app_ids = [app.id.lstrip('/') for app in get_marathon_apps()]
+    print(marathon_app_ids)
```
Rob-Johnson: stray print?

paasta_tools/deployd/watchers.py
```diff
@@ -225,20 +229,20 @@ def process_default(self, event):
...
+    service_instances = []
+    for service, instance in instances:
+        config = load_marathon_service_config(service=service,
+                                              instance=instance,
+                                              cluster=cluster,
+                                              soa_dir=DEFAULT_SOA_DIR)
+        print(config.format_marathon_app_dict()['id'])
```
Rob-Johnson: stray print?

paasta_tools/deployd/watchers.py
```diff
@@ -225,20 +229,20 @@ def process_default(self, event):
...
-            config = load_marathon_service_config(service=service,
-                                                  instance=instance,
-                                                  cluster=self.public_config.get_cluster(),
-                                                  soa_dir=DEFAULT_SOA_DIR)
-            if config.format_marathon_app_dict()['id'] not in marathon_app_ids:
-                service_instances.append((service, instance))
-        return service_instances
+
+def get_service_instances_with_changed_id(instances, cluster):
+    marathon_app_ids = [app.id.lstrip('/') for app in get_marathon_apps()]
```
Rob-Johnson: seems like there should be a helper function for this somewhere already?

paasta_tools/deployd/watchers.py
```diff
@@ -225,20 +229,20 @@ def process_default(self, event):
...
+    print(marathon_app_ids)
+    service_instances = []
+    for service, instance in instances:
+        config = load_marathon_service_config(service=service,
```
Rob-Johnson: I don't know if this is pep8 or not, but I much prefer breaking after the bracket
```python
config = load_marathon_service_config(
  service=service,
  instance=instance,
  cluster=cluster,
  soa_dir=DEFAULT_SOA_DIR
)
````

File: paasta_tools/deployd/watchers.py [hunk 5]
```diff
@@ -225,20 +229,20 @@ def process_default(self, event):
...
+                                              instance=instance,
+                                              cluster=cluster,
+                                              soa_dir=DEFAULT_SOA_DIR)
```
Rob-Johnson: should this be hard coded? I'm predicting a '#4000: paasta deployd ignores -y soa_dir argument'

paasta_tools/deployd/watchers.py
```diff
@@ -214,7 +214,11 @@ def process_default(self, event):
             if new_config != self.public_config:
                 self.log.info("Public config has changed, now checking if it affects any services config shas")
                 self.public_config = new_config
-                service_instances = self.get_service_instances_with_changed_id()
+                instances = get_services_for_cluster(cluster=self.public_config.get_cluster(),
```
Rob-Johnson: given that this returns a list of pairs, wasn't this better named ``service_instances``?

paasta_tools/deployd/watchers.py
mattmb: We actually do but there's some subtle difference regarding `embed_failures=True` iirc. I think I have an existing ticket to fix that up. So I guess I'll come back to this later...

paasta_tools/deployd/watchers.py
mattmb: Yeah, this is just laziness really. I think it should probably be a config variable rather than a cli variable... I'm going to leave this be because it's hardcoded in a few places but happy to fix it in a separate PR.

PR #1241, opened 2017-05-18
Remove quotes around environment variables in local-run

oktopuz:
They are redundant since we don't use any shell to execute 'docker run'.
They cause a problem with xenial images in new docker versions.

Internal ticket: OPS-11477


drolando:
can we have a regression test that really starts a docker container and check if the variables are correctly set?

oktopuz:
@drolando, I'm pretty sure if we have a test that runs 'docker run --env var=value', it will always work. But if someone uses a different format to specified environment variables in some other places, such regression test won't catch this. 

I don't feel like we should go farther than verifying the 'docker run' command.
If Docker doesn't work as expected when we run it with correct arguments, this should be tested in Docker, not in PaaSTA in my opinion.

PR #1242, opened 2017-05-18
test_mac_address no longer depends on /usr/bin/flock

### LINKED ISSUE #1239: Shelling out to flock in unit tests
Because of https://github.com/Yelp/paasta/blob/master/tests/test_mac_address.py#L37-L41 I'm not able to run the tests on OSX any more. At least I need to install flock since the test shells out to flock.

However, even with https://github.com/discoteq/flock I still get a failure. @bchess do we really need to run flock for these tests? In my mind the unit tests should just test that the lock file *would* have been created?

bchess:
_flock_process() doesn't shell out to flock - instead it forks and does it all in python itself. This fixes #1239

### DIFF
--- a/tests/test_mac_address.py
+++ b/tests/test_mac_address.py
@@ -3,8 +3,12 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 
 import contextlib
+import fcntl
 import itertools
-import subprocess
+import os
+import signal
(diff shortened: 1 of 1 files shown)


mattmb:
Awesome, thanks for doing this so quickly! 👍

PR #1244, opened 2017-05-19
Add cron option to paasta_firewall_update

chriskuehl:
This makes turns `paasta_firewall_update` have two modes:
* `paasta_firewall_update cron`
* `paasta_firewall_update daemon`

They share a few options (soa_dir, verbose), but the inotify one has some of its own options (like max update frequency) which don't apply to the cron command.

```
ckuehl@mesos1-uswest1cdevc:~/paasta$ .tox/py27/bin/paasta_firewall_update --help
usage: paasta_firewall_update [-h] [-d soa_dir] [-v] {daemon,cron} ...

Monitor synapse changes and update service firewall rules

positional arguments:
  {daemon,cron}         mode to run firewall update in

optional arguments:
  -h, --help            show this help message and exit
  -d soa_dir, --soa-dir soa_dir
                        define a different soa config directory (default
                        /nail/etc/services)
  -v, --verbose
```
```
ckuehl@mesos1-uswest1cdevc:~/paasta$ .tox/py27/bin/paasta_firewall_update cron --help
usage: paasta_firewall_update cron [-h]

Do a one-time update of iptables rules to match the current running services.

optional arguments:
  -h, --help  show this help message and exit
```
```
ckuehl@mesos1-uswest1cdevc:~/paasta$ .tox/py27/bin/paasta_firewall_update daemon --help
usage: paasta_firewall_update daemon [-h]
                                     [--synapse-service-dir SYNAPSE_SERVICE_DIR]
                                     [-u UPDATE_SECS]

Run a daemon which watches updates to synapse backends and updates iptables
rules.

optional arguments:
  -h, --help            show this help message and exit
  --synapse-service-dir SYNAPSE_SERVICE_DIR
                        Path to synapse service dir (default
                        /var/run/synapse/services)
  -u UPDATE_SECS, --update-secs UPDATE_SECS
                        Poll for new containers every N secs (default 5)
```

Currently specifying `--soa-dir` for `cron` does nothing (since the cronjob doesn't actually look at real data yet), but eventually it will thread that through.

### DIFF
--- a/paasta_tools/firewall_update.py
+++ b/paasta_tools/firewall_update.py
@@ -13,6 +13,7 @@ from inotify.adapters import Inotify
 from inotify.constants import IN_MODIFY
 from inotify.constants import IN_MOVED_TO
 
+from paasta_tools import firewall
 from paasta_tools.chronos_tools import chronos_services_running_here
 from paasta_tools.chronos_tools import load_chronos_job_config
 from paasta_tools.marathon_tools import load_marathon_service_config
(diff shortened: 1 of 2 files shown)


bchess:
shipit

PR #1247, opened 2017-05-22
Handle ConnectionError when getting list of tasks in smartstack

jglukasik:
Internally PAASTA-8357

### DIFF
--- a/paasta_tools/bounce_lib.py
+++ b/paasta_tools/bounce_lib.py
@@ -185,7 +185,7 @@ def is_task_in_smartstack(task, service, nerve_ns, system_paasta_config):
         )
         return task in registered_tasks
     except ConnectionError:
-        log.warning("Failed to connect to %s smartstack, assuming task is unhealthy" % task.host)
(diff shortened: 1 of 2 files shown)


paasta_tools/bounce_lib.py
```diff
@@ -174,13 +175,18 @@ def kill_old_ids(old_ids, client):
...
+    try:
+        registered_tasks = get_registered_marathon_tasks(
+            synapse_host=task.host,
+            synapse_port=system_paasta_config.get_synapse_port(),
+            synapse_haproxy_url_format=system_paasta_config.get_synapse_haproxy_url_format(),
+            service=compose_job_id(service, nerve_ns),
+            marathon_tasks=[task],
+        )
+        return task in registered_tasks
+    except ConnectionError:
```
mattmb: This is probably enough but maybe it's worth catching any `RequestException` https://github.com/kennethreitz/requests/blob/master/requests/exceptions.py#L12

PR #1250, opened 2017-05-23
Start to hook up iptables rules with real world state

chriskuehl:
This gets the cronjob correctly applying rules, with the exception of still using fake mac addresses. You can see it in action on `mesos1-uswest1cdevc`, where I'm running instances of `example_happyhour` and `robj-test-service` in both monitor and block mode. (Try e.g. `sudo iptables-save` to see the rules it's set.)

It supports the `internet` well-known name, allowing access to smartstack (but currently only via yocalhost), and monitor/block mode.

#### todo
* [x] use real MAC addresses for instances
* [x] fix/write tests

### DIFF
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -143,11 +143,8 @@ def active_service_groups(soa_dir):
     """Return active service groups."""
     from paasta_tools import firewall_update
     service_groups = collections.defaultdict(set)
-    for service, instance, framework in firewall_update.services_running_here(soa_dir):
-        service_groups[ServiceGroup(service, instance, framework, soa_dir)].add(
-            # TODO: how to get the mac address of this container?
-            '02:42:a9:fe:00:02',
-        )
+    for service, instance, framework, mac in firewall_update.services_running_here(soa_dir):
+        service_groups[ServiceGroup(service, instance, framework, soa_dir)].add(mac)
     return service_groups
 
 
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -12,6 +12,7 @@ import six
 from paasta_tools import iptables
 from paasta_tools.cli.utils import get_instance_config
 from paasta_tools.marathon_tools import get_all_namespaces_for_service
+from paasta_tools.utils import get_running_mesos_docker_containers
 from paasta_tools.utils import load_system_paasta_config
 
 
@@ -109,7 +110,7 @@ def _well_known_rules(conf):
             )
         else:
             # TODO: handle better
-            raise KeyError(resource)
+            raise AssertionError(resource)
 
 
 def _smartstack_rules(conf, soa_dir):
@@ -131,11 +132,28 @@ def _smartstack_rules(conf, soa_dir):
         )
 
 
+def services_running_here():
+    """Generator helper that yields (service, instance, mac address) of both
+    marathon and chronos tasks.
+    """
+    for container in get_running_mesos_docker_containers():
+        if container['HostConfig']['NetworkMode'] != 'bridge':
+            continue
+
+        service = container['Labels'].get('paasta_service')
+        instance = container['Labels'].get('paasta_instance')
+
+        if service is None or instance is None:
+            continue
+
+        mac = container['NetworkSettings']['Networks']['bridge']['MacAddress']
+        yield service, instance, mac
+
+
 def active_service_groups(soa_dir):
     """Return active service groups."""
-    from paasta_tools import firewall_update
     service_groups = collections.defaultdict(set)
-    for service, instance, mac in firewall_update.services_running_here():
+    for service, instance, mac in services_running_here():
         service_groups[ServiceGroup(service, instance, soa_dir)].add(mac)
     return service_groups
 
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -7,7 +7,13 @@ import hashlib
 import itertools
 import json
 
+import six
+
 from paasta_tools import iptables
+from paasta_tools.chronos_tools import load_chronos_job_config
+from paasta_tools.marathon_tools import get_all_namespaces_for_service
+from paasta_tools.marathon_tools import load_marathon_service_config
+from paasta_tools.utils import load_system_paasta_config
 
 
 PRIVATE_IP_RANGES = (
@@ -21,15 +27,16 @@ PRIVATE_IP_RANGES = (
 
 class ServiceGroup(collections.namedtuple('ServiceGroup', (
     'service',
-    'dependency_group',
-    'mode',
+    'instance',
+    'framework',
+    'soa_dir',
 ))):
     """A service group.
 
-    :param service: the name of a service
-    :param dependency_group: the name of a set of dependencies
-                             (the key in dependencies.yaml)
-    :param mode: either 'monitor' or 'block'
+    :param service: service name
+    :param instance: instance name
+    :param framework: Mesos framework (e.g. marathon or chronos)
+    :param soa_dir: path to yelpsoa-configs
     """
 
     __slots__ = ()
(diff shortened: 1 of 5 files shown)


paasta_tools/firewall.py
```diff
@@ -50,48 +57,98 @@ def chain_name(self):
...
     """Return active service groups."""
-    # TODO: actually read these from somewhere
-    return {
-        ServiceGroup('cool_service', 'main', 'block'): {
+    from paasta_tools import firewall_update
+    service_groups = collections.defaultdict(set)
+    for service, instance, framework in firewall_update.services_running_here(soa_dir):
+        service_groups[ServiceGroup(service, instance, framework, soa_dir)].add(
+            # TODO: how to get the mac address of this container?
             '02:42:a9:fe:00:02',
```
chriskuehl: @bchess I'm not really sure how we would go about getting the MAC address of the instance here. This loop iterates once per service container on the host, but I don't think it has any info that would help us get that container's mac address. Do you have any ideas?

It almost feels to me like we'd be better off iterating over the running docker containers on the host, rather than querying a remote API to get the list.

bchess: Yeah agreed - kwa and I sorta leaned in that direction in my earlier review (https://github.com/Yelp/paasta/pull/1234#discussion_r117272192). Also remember that there may be >1 running container with the same service and instance.

paasta_tools/firewall_update.py
```diff
@@ -95,16 +92,12 @@ def process_inotify_event(event, services_by_dependencies):
 
 
 def services_running_here(soa_dir):
-    # Generator helper that yields (service, instance, config) of both marathon and chronos tasks
-    cluster = load_system_paasta_config().get_cluster()
-
+    # Generator helper that yields (service, instance, framework) of both marathon and chronos tasks
```
bchess: If you change what this returns you'll need to fix up its existing callers, like smartstack_dependencies_of_running_firewalled_services()

paasta_tools/firewall.py
bchess: @solarkennedy can you point us in the right direction re: existing labels? On mesosstage the only label I see is MESOS_TASK_ID

bchess: Oh wait I see some with paasta_service and paasta_instance. format_docker_parameters() does it

paasta_tools/firewall_update.py
chriskuehl: good catch -- I think I'll probably change the caller (seems less yucky since it's straightforward to fix the other call-site). just for context, I made this change because from the perspective of the cronjob, all of the rules it sets up are per (service, instance) pair as opposed to per running container, and it was somewhat awkward to gather them together before constructing the chain.

paasta_tools/firewall.py
bchess: There's a fix for this in https://github.com/Yelp/paasta/pull/1253

bchess: @chriskuehl totally untested, but after #1253 I believe we can simply do this: https://gist.github.com/bchess/2900f9379ca4e4dac47dd103600e3c9c

Plus return the mac address

chriskuehl: That worked! The jump rules now look right, e.g.:

```
-A PAASTA -m mac --mac-source 02:42:A9:FE:00:08 -j PAASTA.robj-test-.11f80dd2dd
-A PAASTA -m mac --mac-source 02:42:A9:FE:00:07 -j PAASTA.example_ha.b9644fd192
-A PAASTA -m mac --mac-source 02:42:A9:FE:00:03 -j PAASTA.example_ha.ee6c7d4dcd
-A PAASTA -m mac --mac-source 02:42:A9:FE:00:06 -j PAASTA.example_ha.ee6c7d4dcd
-A PAASTA -m mac --mac-source 02:42:A9:FE:00:0A -j PAASTA.robj-test-.11f80dd2dd
-A PAASTA -m mac --mac-source 02:42:A9:FE:00:04 -j PAASTA.robj-test-.796beca72d
-A PAASTA -m mac --mac-source 02:42:A9:FE:00:09 -j PAASTA.example_ha.b9644fd192
-A PAASTA -m mac --mac-source 02:42:A9:FE:00:02 -j PAASTA.robj-test-.796beca72d
```

I've added a commit that gets the cronjob working. It's still pretty WIP-y, but I'll start getting this ready to ship.

paasta_tools/firewall_update.py
```diff
@@ -95,16 +91,23 @@ def process_inotify_event(event, services_by_dependencies):
...
+        instance = container['Labels'].get('paasta_instance')
+
+        if service is None or instance is None:
+            continue
 
-    for service, instance, port in marathon_services_running_here():
-        config = load_marathon_service_config(service, instance, cluster, load_deployments=False, soa_dir=soa_dir)
-        yield service, instance, config
+        # TODO: this sucks and is untested, maybe we should add a label for framework?
+        framework = 'marathon' if container['Ports'] else 'chronos'
```
solarkennedy: You could "not care" if you wanted to, but the normal `get_instance_config` thingy has to look through each yaml file looking for an instance. (we assume they are unique across frameworks)

chriskuehl: Ah, if we can "not care" then I think that's a good solution. I didn't realize we assumed they were unique, so I was defending against the case of "service.instance" having both a chronos and marathon config. thanks!

paasta_tools/firewall.py
```diff
@@ -50,48 +57,95 @@ def chain_name(self):
         return chain
 
     @property
+    def config(self):
+        load_fn = {
```
solarkennedy: This kinda duplicates
https://github.com/Yelp/paasta/blob/8eb6d8b406b10be540a825152341439a999e95cb/paasta_tools/cli/utils.py#L665

paasta_tools/firewall.py
```diff
@@ -50,48 +57,95 @@ def chain_name(self):
...
+        # TODO: log-prefix
+        return iptables.Rule(
+            protocol='ip',
+            src='0.0.0.0/0.0.0.0',
+            dst='0.0.0.0/0.0.0.0',
+            target='LOG',
+            matches=(),
+        )
+    else:
+        raise AssertionError(policy)
```
solarkennedy: This is a potentially dangerous decision, where a misconfiguration can cause firewall updates to crash everywhere, maybe leaving things inconsistent?

*maybe* just default to monitor? Or log and then literally do nothing if there is a typo?

paasta_tools/firewall.py
```diff
@@ -50,48 +57,95 @@ def chain_name(self):
...
-                src='0.0.0.0/0.0.0.0',
-                dst='169.254.255.254/255.255.255.255',
-                target='ACCEPT',
-                matches=(
-                    ('tcp', (('dport', '20668'),)),
-                )
-            ),
+            )
+        else:
+            # TODO: handle better
```
solarkennedy: yea, just log and do nothing. the show must go on here and user typos are not an excuse to crash the whole system.

I was hoping we could make a new log "component" for paasta logs (`security`?) and it could show up there.

paasta_tools/firewall.py
chriskuehl: fixed now, thanks for pointing that out!

paasta_tools/firewall.py
chriskuehl: agreed, before we start using this seriously in a stage, we need to do something better here.

i think for now it should be okay to ship this just to unblock all the work that depends on this; i'll file an internal ticket to improve error handling

chriskuehl:
@bchess I think this is pretty much good-to-go. I'm going to write a few more tests to cover some of the edge cases that we're missing coverage on, but the bulk of the work is here.

paasta_tools/firewall.py
```diff
@@ -50,48 +53,91 @@ def chain_name(self):
...
-            'fe:a3:a3:da:2d:40',
-        },
-        ServiceGroup('dumb_service', 'other', 'block'): {
-            'fe:a3:a3:da:2d:30',
-            'fe:a3:a3:da:2d:31',
-        },
-    }
+    from paasta_tools import firewall_update
+    service_groups = collections.defaultdict(set)
+    for service, instance, mac in firewall_update.services_running_here():
```
bchess: We should probably put services_running_here() inside firewall.py to avoid this nested circular import. Actually maybe even utils.py instead

tests/test_firewall_update.py
```diff
@@ -67,15 +67,66 @@ def test_parse_args_default_cron():
...
+@pytest.mark.usefixtures('mock_get_running_mesos_docker_containers')
+def test_services_running_here():
+    assert tuple(firewall_update.services_running_here()) == (
+        ('myservice', 'hassecurity', '02:42:a9:fe:00:0a'),
+        ('myservice', 'chronoswithsecurity', '02:42:a9:fe:00:0b'),
+    )
+
+
+@pytest.mark.usefixtures('mock_get_running_mesos_docker_containers')
+@mock.patch.object(firewall_update, 'load_system_paasta_config', return_value=mock.Mock(**{
```
bchess: autospec=True

PR #1251, opened 2017-05-23
Deployd actions for dying threads

mattmb:
This adds some sensible defences for workers and watchers dying due to
unexpected exceptions. In the case of a watcher dying we commit suicide
and simply let another leader take over. If a worker dies then we just
restart it, unless all workers die in which case we also commit suicide.

### DIFF
--- a/paasta_tools/deployd/master.py
+++ b/paasta_tools/deployd/master.py
@@ -7,6 +7,7 @@ import logging
 import logging.handlers
 import os
 import socket
+import sys
 import time
 
 import service_configuration_lib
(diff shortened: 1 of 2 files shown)


paasta_tools/deployd/master.py
```diff
@@ -136,15 +136,39 @@ def main_loop(self):
                 message = None
             if message == "ABORT":
                 break
+            if not self.all_watchers_running():
+                self.log.error("One or more watcher died, committing suicide!")
+                break
+            if self.all_workers_dead():
+                self.log.error("All workers have died, comitting suicide!")
+                break
```
solarkennedy: This won't exit non-zero though, is that ok?

mattmb: It'd be clearer if they did! It doesn't matter to the init system of course but I agree this should be `sys.exit(1)`

PR #1252, opened 2017-05-23
Fix deployd watcher NoDockerImage exception

mattmb:
Catch this exception when checking the marathon app id.

### DIFF
--- a/paasta_tools/deployd/watchers.py
+++ b/paasta_tools/deployd/watchers.py
@@ -25,6 +25,7 @@ from paasta_tools.mesos_maintenance import get_draining_hosts
 from paasta_tools.utils import get_services_for_cluster
 from paasta_tools.utils import list_all_instances_for_service
 from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.utils import NoDockerImageError
 from paasta_tools.utils import PATH_TO_SYSTEM_PAASTA_CONFIG_DIR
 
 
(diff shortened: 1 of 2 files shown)


paasta_tools/deployd/watchers.py
```diff
@@ -242,7 +243,11 @@ def get_service_instances_with_changed_id(marathon_client, instances, cluster):
                                               instance=instance,
                                               cluster=cluster,
                                               soa_dir=DEFAULT_SOA_DIR)
-        if config.format_marathon_app_dict()['id'] not in marathon_app_ids:
+        try:
+            config_app_id = config.format_marathon_app_dict()['id']
+        except NoDockerImageError:
+            config_app_id = None
+        if not config_app_id or config_app_id not in marathon_app_ids:
```
solarkennedy: Is this right?

Isn't it "if config_appid" (or config_app_id is not None)


Also this line here about `config_app_id not in marathon_ap_ids`, does that mean we would never give attention to draining apps and stuff? (where the appid is already in place)

mattmb: Hmm, I guess the brackets don't matter in this case because the unit test passes. It'd be clearer with them though so I'll send a fixup.

Regarding draining apps etc. we only call this code path from the config file watchers. So the watchers that are watching for autoscaling and maintenance don't check the app id and just put the service straight into the queue.

PR #1255, opened 2017-05-24
Make setup_marathon_job take care of slack instance slots on scaledown

solarkennedy:
This has happened a couple of times on yelp-main bounces. The steps to repro are:

1. Scale up an app to say instances: 200, don't let it scale up that far, let it get to 100 (get it into a Waiting state)
2. Do a code push (bounce)
3. Scale down an app to instances: 100

Expected: The bounce would "finish" because it would go from 200->100 and scale down that 100 of empty waiting slack
Actual: The bounce never scales down anything because there isn't anything to drain or kill, which is kinda the only way that the bounce knows how to scale down. It will wait to launch more just so it can drain+kill.

This fixes the bug by introducing the concept of "slack" and will remove slack on scaledown.
This could have been implemented as another arg for do_bounce. Let me know if you would like me to do that. If I did that I could move the scale_app call in :488 as well.

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -483,12 +483,12 @@ def deploy_service(
     if new_app_running:
         num_at_risk_tasks = get_num_at_risk_tasks(new_app, draining_hosts=get_draining_hosts())
         if new_app.instances < config['instances'] + num_at_risk_tasks:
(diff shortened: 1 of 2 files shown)


solarkennedy:
@EvanKrall  can you take a look at this one?

solarkennedy:
I'll ship this next week.

PR #1257, opened 2017-05-25
Make acceptable delay in check_chronos_jobs configurable

oktopuz:
internal tickets: PAASTA-11186, PAASTA-11277

### DIFF
--- a/paasta_tools/check_chronos_jobs.py
+++ b/paasta_tools/check_chronos_jobs.py
@@ -52,7 +52,7 @@ def compose_monitoring_overrides_for_service(chronos_job_config, soa_dir):
     """ Compose a group of monitoring overrides """
     monitoring_overrides = chronos_job_config.get_monitoring()
     if 'alert_after' not in monitoring_overrides:
-        monitoring_overrides['alert_after'] = '2m'
+        monitoring_overrides['alert_after'] = '15m'
     monitoring_overrides['check_every'] = '1m'
     monitoring_overrides['runbook'] = monitoring_tools.get_runbook(
         monitoring_overrides, chronos_job_config.service, soa_dir=soa_dir)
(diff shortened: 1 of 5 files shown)


paasta_tools/monitoring_tools.py
```diff
@@ -45,7 +45,8 @@ def monitoring_defaults(key):
                'https://paasta.readthedocs.io/en/latest/yelpsoa_configs.html#monitoring-yaml',
         'ticket': False,
         'project': None,
-        'realert_every': -1
+        'realert_every': -1,
+        'chronos_job_acceptable_delay': 15 * 60,  # 15 minutes
```
solarkennedy: I would rather this not be a first-class key on the monitoring dictionary, and instead by an input to a chronos config. (without a prefix of "chronos")

paasta_tools/check_chronos_jobs.py
```diff
@@ -224,7 +224,12 @@ def sensu_message_status_for_jobs(chronos_job_config, service, instance, cluster
             sensu_status = pysensu_yelp.Status.OK
             output = "Job %s%s%s is disabled - ignoring status." % (service, utils.SPACER, instance)
         else:
-            acceptable_delay = 15 * 60
+            monitoring_overrides = chronos_job_config.get_monitoring()
+            acceptable_delay = monitoring_tools.get_chronos_job_acceptable_delay(
```
solarkennedy: Is the use-case here that people want to give chronos more time before an alert, if so, why is `alert_after` not enough?

oktopuz:
* removed acceptable_delay
* increased alert_after for the chronos jobs check to 15m to compensate acceptable_delay removal
* refactored the check to make it more precise for crontab-like schedules with non-fixed interval (0 * * * 1-5, 0 0 */2 * * and etc)

PR #1258, opened 2017-05-25
filter non-proxy_port for firewall_update

bchess:
smartstack_dependencies_of_running_firewalled_services() only returns dependencies with no proxy_port

idk if this may conflict with your ongoing branch, we may need to make this a kwarg. But here's the code.

internal ticket PAASTA-11261

### DIFF
--- a/paasta_tools/firewall_update.py
+++ b/paasta_tools/firewall_update.py
@@ -12,6 +12,7 @@ from collections import defaultdict
 from inotify.adapters import Inotify
 from inotify.constants import IN_MODIFY
 from inotify.constants import IN_MOVED_TO
+from service_configuration_lib import read_service_configuration
 
 from paasta_tools import firewall
 from paasta_tools.chronos_tools import chronos_services_running_here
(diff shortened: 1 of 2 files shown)


paasta_tools/firewall_update.py
```diff
@@ -118,12 +119,21 @@ def smartstack_dependencies_of_running_firewalled_services(soa_dir=DEFAULT_SOA_D
 
         smartstack_dependencies = [d['smartstack'] for d in dependencies if d.get('smartstack')]
         for smartstack_dependency in smartstack_dependencies:
-            # TODO: filter down to only services that have no proxy_port
-            dependencies_to_services[smartstack_dependency].add((service, instance))
+            dependency_service, dependency_instance = smartstack_dependency.split('.', 1)
+
+            # filter down to only services that have no proxy_port
+            if service_proxy_port(dependency_service, soa_dir) is None:
+                dependencies_to_services[smartstack_dependency].add((service, instance))
```
solarkennedy: I kinda expected an "else" where the proxy is `not None` and and the corresponding firewall rule for yocalhost:proxy_port is in place?

bchess: the job of this script is to trigger the iptables updates necessary for direct access to specific backends. For services that are accessed purely through haproxy, we don't need to do any ongoing iptables rules as backends change -- we're just allow access to the yocalhost haproxy proxy_port.

bchess: (This script isn't involved in what happens in the docker_wrapper)

paasta_tools/firewall_update.py
```diff
@@ -12,6 +12,7 @@
 from inotify.adapters import Inotify
```
chriskuehl: All this code looks reasonable to me, but some thoughts about the approach of not allowing access to backends directly if a proxy_port exists:

1. As written, it's going to be pretty much impossible for a service to transition from accessing backends directly to using the proxy_port. As soon as they add a proxy_port, the firewall stops allowing access to the backends. They really need a transition period where both are allowed so that they can go update the clients to access the proxy_port before disabling direct access to the backends.

2.  @bobtfish mentioned in PAASTA-11261 that some services have a proxy_port which is used for the initial request, but afterwards they talk to the backends directly (he mentioned Vault and ES).

I think some kind of toggle for allowing direct access to the backends even when there's a proxy_port could fix this?

bchess: Yeah agreed. :/ Will think about this some more. Maybe we just don't do it for the time being.

PR #1260, opened 2017-05-26
Use private PyPI when available

chriskuehl:
internal ticket: PAASTA-11317

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,18 +15,20 @@
 # Set ENV to 'YELP' if FQDN ends in '.yelpcorp.com'
 # Otherwise, set ENV to the FQDN
 ifeq ($(findstring .yelpcorp.com,$(shell hostname -f)), .yelpcorp.com)
+	export PIP_INDEX_URL ?= https://pypi.yelpcorp.com/simple
 	PAASTA_ENV ?= YELP
 else
+	export PIP_INDEX_URL ?= https://pypi.python.org/simple
 	PAASTA_ENV ?= $(shell hostname -f)
 endif
 
 .PHONY: all docs test itest
 
 docs: .paasta/bin/activate
-	.paasta/bin/tox -e docs
+	.paasta/bin/tox -i $(PIP_INDEX_URL) -e docs
 
 test: .paasta/bin/activate
-	.paasta/bin/tox
+	.paasta/bin/tox -i $(PIP_INDEX_URL)
 
 .paasta/bin/activate: requirements.txt requirements-dev.txt
 	test -d .paasta/bin/activate || virtualenv -p python2.7 .paasta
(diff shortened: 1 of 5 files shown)


chriskuehl:
lol this is hopelessly broken

solarkennedy:
:(

What if we revert the tox extension and go back to bootstrappy stuff?

chriskuehl:
I don't think that actually works -- it was failing for me before (but not for some others). jlynch ended up installing a debian package which provided some build dependency which I *think* gets paasta building correctly on devboxes now.

What I think happened is that the people who regularly work on paasta had some wheel cached, and so didn't notice when a build dep got uninstalled (but did notice when I added tox-pip-extensions, which changed which wheel cache was used and thus caused failures). But I'm not totally sure.

In any case I'd still like to make this work at some point. If you're still having build trouble, I'm happy to revert the tox-pip-extensions commit for now, I'm just not sure it's a complete solution.

chriskuehl:
@macisamuele yep, looks like that did it, thanks! Those itests are annoying to test :P

chriskuehl:
@solarkennedy @mattmb @jolynch any thoughts on this? I think it's ready to ship.

The goal is to internally always use the internal registry without any extra config, and always use the public one externally.

It's a little hacky, but it makes local development _way_ faster (at least internally) and lets us avoid installing lots of build dependencies on our dev machines.

Makefile
```diff
@@ -15,18 +15,20 @@
 # Set ENV to 'YELP' if FQDN ends in '.yelpcorp.com'
 # Otherwise, set ENV to the FQDN
 ifeq ($(findstring .yelpcorp.com,$(shell hostname -f)), .yelpcorp.com)
```
mattmb: What about something like an if on whether you can `curl pypi.yelpcorp.com/simple` Would mean we depend on the user having curl installed... Maybe netcat is more ubiquitous `nc -v pypi.yelpcorp.com 443 -z`?

For example my laptop doesn't seem to have `hostname -f` with yelpcorp.com even when I'm on the Yelp network.

macisamuele: What about rely on an env variable to decide whatever use internal or public pypi?
I'm thinking to something similar to ``YELP=any-thing make ...`` to use internal pypi

It could reduce a little the hacks and will be fully operative on Linux and OS X

solarkennedy:
I'm ok with the increased complexity for the build speedup.

Makefile
chriskuehl: Sounds fine to me. I'll merge this for now since it doesn't change the status quo (laptops were using public PyPI before) but happy to have it also check an environment variable. Can you suggest a name? (Just `YELP`?)

PR #1263, opened 2017-05-31
Catch ValueErrors in processing {white,black}lists

jglukasik:
Currently, passing a bad {white,black}list will cause the autoscaler to crash. This PR will catch this kind of ValueError, assume the slave does NOT pass, logs, and continues.

Question: is this assumption sound? Or would it be better to assume it passes the lists?


solarkennedy:
lgtm, I think these are OK to do on malformed input.

PR #1267, opened 2017-06-02
Port the native scheduler to use pymesos

jolynch:
This gets rid of the knarly mesos.native dependency and paves the way to python3 :-D

### DIFF
--- a/paasta_itests/paasta_native.feature
+++ b/paasta_itests/paasta_native.feature
@@ -1,5 +1,4 @@
 Feature: Paasta native mesos framework
-  @wip
   Scenario: we can start a service
     Given a working paasta cluster, with docker registry docker.io
       And a new paasta_native config to be deployed, with 3 instances
--- a/paasta_itests/paasta_native.feature
+++ b/paasta_itests/paasta_native.feature
@@ -1,4 +1,5 @@
 Feature: Paasta native mesos framework
+  @wip
   Scenario: we can start a service
     Given a working paasta cluster, with docker registry docker.io
       And a new paasta_native config to be deployed, with 3 instances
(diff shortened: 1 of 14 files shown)


paasta_itests/paasta_native.feature
```diff
@@ -1,4 +1,5 @@
 Feature: Paasta native mesos framework
+  @wip
```
EvanKrall: remove

paasta_tools/frameworks/native_service_config.py
```diff
@@ -65,60 +70,88 @@ def task_name(self, base_task):
...
+        computed separately.
+        """
+        task = Dict()
+        docker_volumes = self.get_volumes(
+            system_volumes=system_paasta_config.get_volumes()
+        )
+        task = Dict(
+            container=Dict(
+                type='DOCKER',
+                docker=Dict(
```
EvanKrall: is there no way to get addict to create nested `Dict`s from nested `dict`s?

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -545,25 +534,27 @@ def create_driver(
...
-    credential = mesos_pb2.Credential()
-    credential.principal = system_paasta_config.get_paasta_native_config()['principal']
-    credential.secret = system_paasta_config.get_paasta_native_config()['secret']
+    master_uri = '{}:{}'.format(
+        mesos_tools.get_mesos_leader(), mesos_tools.MESOS_MASTER_PORT
+    )
 
-    framework.principal = system_paasta_config.get_paasta_native_config()['principal']
+    framework = Dict(
+        user=getpass.getuser(),
```
solarkennedy: this looks different. should it not be `None`? I don't think it is supposed to literally be the current user.

PR #1270, opened 2017-06-02
Catch IOErrors when writing to log files, so you can e.g. use paasta restart on boxes where you don't have write permission to the log file.

EvanKrall:


EvanKrall:
womp, I broke py3 https://stackoverflow.com/questions/29347790/difference-between-ioerror-and-oserror

PR #1271, opened 2017-06-02
let autoscaler measure capacity at region level instead of superregion

somic:
PAASTA-7692

Could it be as simple as this? 'region' vs 'datacenter' confusion is because slave attributes have datacenter=us-west-1, while autoscaler config (from S3) has region=us-west-1.

@solarkennedy @mattmb

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -196,14 +196,14 @@ class ClusterAutoscaler(ResourceLogMixin):
                     current_instances, expected_instances, MISSING_SLAVE_PANIC_THRESHOLD)
                 raise ClusterAutoscalingError(error_message)
 
-        pool_utilization_dict = get_resource_utilization_by_grouping(
-            lambda slave: slave['attributes']['pool'],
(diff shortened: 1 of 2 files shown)


Rob-Johnson:
Can we instead just fix the 'region' value in the autoscaler config? It'll probably be a bit harder to do (terraform code) but I think it's probably the 'right' thing to do.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -196,14 +196,14 @@ def get_mesos_utilization_error(self,
                     current_instances, expected_instances, MISSING_SLAVE_PANIC_THRESHOLD)
                 raise ClusterAutoscalingError(error_message)
 
-        pool_utilization_dict = get_resource_utilization_by_grouping(
-            lambda slave: slave['attributes']['pool'],
+        region_pool_utilization_dict = get_resource_utilization_by_grouping(
+            lambda slave: (slave['attributes']['pool'], slave['attributes']['datacenter'],),
```
solarkennedy: This 'datacenter' word really shouldn't be in here, but this is slightly better than what we do now.

I still think the long term solution is to "just" look at the slaves that are in the resource itself.

PR #1273, opened 2017-06-05
Add support for logging with prefix and limits

bchess:
This adds log-prefix and limit to monitor mode.

target_parameters ended up being a bit trickier than expected because adding a REJECT rule will cause iptables to automatically imply a "--reject-with icmp-port-unreachable" rule. This ends up causing comparison problems when loaded back in. To fix, I require that reject-with always be specified in the case of REJECT.

### DIFF
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -170,14 +170,17 @@ def services_running_here():
         if service is None or instance is None:
             continue
 
-        mac = container['NetworkSettings']['Networks']['bridge']['MacAddress']
-        yield service, instance, mac
+        network_info = container['NetworkSettings']['Networks']['bridge']
+
+        mac = network_info['MacAddress']
+        ip = network_info['IPAddress']
+        yield service, instance, mac, ip
 
 
 def active_service_groups(soa_dir):
     """Return active service groups."""
     service_groups = collections.defaultdict(set)
-    for service, instance, mac in services_running_here():
+    for service, instance, mac, ip in services_running_here():
         service_groups[ServiceGroup(service, instance, soa_dir)].add(mac)
     return service_groups
 
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -77,6 +77,13 @@ class ServiceGroup(collections.namedtuple('ServiceGroup', (
     def update_rules(self):
         iptables.ensure_chain(self.chain_name, self.rules)
 
+    @property
+    def log_prefix(self):
+        # log-prefix is limited to 29 characters total
+        # space at the end is necessary to separate it from the rest of the line
+        # no restrictions on any particular characters afaict
+        return 'paasta.{}'.format(self.service)[:28] + ' '
+
 
 def _default_rule(conf):
     policy = conf.get_outbound_firewall()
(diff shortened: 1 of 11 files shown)


bchess:
Now adding a syslog->paasta log adapter. (Probably scribe, but it uses whatever log is configured in utils). This is a UDP server - we'll need to set up a rule in syslog-ng to redirect kernel messages to this daemon.

bchess:
Scalability for the UDP server will need some work. Any messages that occur while it's processing will be dropped. Maybe easiest handled just by adding more copies of the daemon.

paasta_tools/firewall.py
```diff
@@ -154,14 +170,17 @@ def services_running_here():
         if service is None or instance is None:
             continue
 
-        mac = container['NetworkSettings']['Networks']['bridge']['MacAddress']
-        yield service, instance, mac
+        network_info = container['NetworkSettings']['Networks']['bridge']
+
+        mac = network_info['MacAddress']
+        ip = network_info['IPAddress']
+        yield service, instance, mac, ip
```
chriskuehl: github won't let me comment on it, but update docstring now that it returns ip?

paasta_tools/firewall_logging.py
```diff
@@ -0,0 +1,114 @@
...
+        data, socket = self.request
+        syslog_to_paasta_log(data, self.cluster)
+
+
+def syslog_to_paasta_log(data, cluster):
+    iptables_log = parse_syslog(data)
+    if iptables_log is None:
+        return
+
+    service, instance = lookup_service_instance_by_ip(iptables_log['SRC'])
```
chriskuehl: nice

paasta_tools/firewall_logging.py
```diff
@@ -0,0 +1,114 @@
...
+    if service is None or instance is None:
+        return
+
+    # prepend hostname
+    log_line = iptables_log['hostname'] + ': ' + iptables_log['message']
```
chriskuehl: agreed that it would be cool to eventually transform this into something more readable (maybe looking up smartstack ports), but this is a great start

paasta_tools/firewall.py
```diff
@@ -87,15 +94,22 @@ def _default_rule(conf):
             dst='0.0.0.0/0.0.0.0',
             target='REJECT',
             matches=(),
+            target_parameters=(
+                (('reject-with', ('icmp-port-unreachable',))),
+            ),
```
bchess: yeah I ran into that too. :P

bchess:
@solarkennedy looking for explicit feedback from you on this, as you were part of the earlier syslog2scribe CR

solarkennedy:
Nice job. This really does look a much better solution. You are right, it will be easier to iterate on this and make improve the user experience of reading these logs.

A common gotcha on the paasta code-base with daemons is dealing with the yaml cache, does that need to be address or have you already solved that?

Can you elaborate on your scaling strategy? I don't know how your normally scale a python udp servers, but when you say more copies, what do you mean exactly? Like one on 1516, 1517, etc?

bchess:
Updated with master, and fleshed out the scalability aspect. Admittedly this was a bit trickier than I thought, by default SocketServer uses SO_REUSEADDR when what we really want is SO_REUSEPORT. The latter will round-robin incoming packets across all processes listening on that port.

I added some additional code to handle the spawning of more workers. We _could_ instead just handle this in upstart or something - just spawn multiple copies of the process. child/parent doesn't matter, we just want to have multiple processes.

bchess:
On caching: I don't think this code relies on any config loading, aside from get_cluster(). services_running_here() queries docker client directly. So I don't think there's any caching happening.

bchess:
Yeah @chriskuehl, agreed, it'd be nice if we could find something else to manage a pool of identical workers. I'm just not sure what that thing is.. We can disable this behavior with "-w 1" if we find a good option.

bchess:
'numprocs' in supervsiord could work: http://supervisord.org/configuration.html?highlight=numprocs

paasta_tools/firewall_logging.py
```diff
@@ -0,0 +1,141 @@
...
+        level='debug',
+        cluster=cluster,
+        instance=instance,
+        line=log_line,
+    )
+
+
+def parse_syslog(data):
+    parsed_data = syslogmp.parse(data)
+    full_message = parsed_data.message.decode()
```
chriskuehl: we probably have quite a few syslog messages that don't decode as ascii or utf8 or whatever the locale will be, maybe decode as latin1 (so at least it never throws an exception) or just catch the exception and ignore the line (rather than crash the worker)?

bchess: ah good catch thx

PR #1277, opened 2017-06-06
Re-order replication check output to list CRITICAL before OK

matthewbentley:
Since CRITICAL sends pages, it makes sense to list the critical lines first.

I am not sure if this needs extra tests.

### DIFF
--- a/paasta_tools/check_marathon_services_replication.py
+++ b/paasta_tools/check_marathon_services_replication.py
@@ -158,6 +158,8 @@ def check_smartstack_replication_for_instance(
     else:
         expected_count_per_location = int(expected_count / len(smartstack_replication_info))
         output = ''
+        output_critical = ''
+        output_ok = ''
         under_replication_per_location = []
 
         for location, available_backends in sorted(smartstack_replication_info.items()):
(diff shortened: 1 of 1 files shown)


paasta_tools/check_marathon_services_replication.py
```diff
@@ -158,20 +158,25 @@ def check_smartstack_replication_for_instance(
...
-                output += '- Service %s has %d out of %d expected instances in %s (CRITICAL: %d%%)\n' % (
+                output_critical += '- Service %s has %d out of %d expected instances in %s (CRITICAL: %d%%)\n' % (
                     full_name, num_available_in_location, expected_count_per_location, location, ratio)
             else:
-                output += '- Service %s has %d out of %d expected instances in %s (OK: %d%%)\n' % (
+                output_ok += '- Service %s has %d out of %d expected instances in %s (OK: %d%%)\n' % (
                     full_name, num_available_in_location, expected_count_per_location, location, ratio)
             under_replication_per_location.append(under_replicated)
 
+        output += output_critical
```
solarkennedy: how about add a newline to separate the two (if there is critical output to separate)

PR #1288, opened 2017-06-09
Fail push-to-registry hard when image exists in the registry

oktopuz:
internal ticket: PAASTA-11415


solarkennedy:
Don't ship this until we have an answer for how we are going to not cause all pipeline to go read all the time.

oktopuz:
I'll re-open this (or re-submit) once we have a solution for jenkins.

PR #1290, opened 2017-06-09
only count non-terminal tasks in metastatus

Rob-Johnson:
we only want to calculate resource utilization of running tasks (and
thus ignore those that aren't actually using allocated resources).

internal PAASTA-11698


tests/metrics/test_metastatus_lib.py
```diff
@@ -689,6 +689,48 @@ def test_get_resource_utilization_per_slave():
...
+            },
+            'state': 'TASK_RUNNING'
+        },
+        {
+            'resources': {
+                'cpus': 10,
+                'mem': 10,
+                'disk': 10
+            },
+            'state': 'TASK_RUNNING'
```
mattmb: One of these should be a terminal state right? I guess we'll see if I'm right in a mo when travis finishes...

Rob-Johnson: we talked IRL about this - filtering happens before this function

PR #1294, opened 2017-06-12
Add unittest for run_chronos_rerun with run_all_related_jobs flag

macisamuele:
Added a small unittest to prevent regression caused by "wrong" ``chronos_rerun_graph`` generated SSH command.
This is a test only change :)


Rob-Johnson:
thanks!

PR #1300, opened 2017-06-12
Allow for per-service docker registry

jglukasik:
This PR allows service owners to specify their own docker registry to use.

Currently, the docker registry is set at a system-wide level in the system config.  This system-wide docker registry will still be the default, but service owners can optionally specify a different registry by setting the "docker_registry" key in their service configuration file.


solarkennedy:
docs plz

PR #1302, opened 2017-06-13
Enforce ordering on iptables rules

bchess:
This adds a function called reorder_chain() that ensures that all rules of a chain have REJECT being last, and LOG being second-last.

I have admittedly mixed feelings about it. Changing the existing behavior of ensure_chain() is pretty complex - it'd require a diff algo. So instead this is its own separate function. This way the code is reasonably simple to follow IMHO. I also wanted something that will fix existing chains, like if we introduce a bug with ordering at any point, or a failure results in inconsistent state. This is a straightforward way to fix any existing chain.

It doesn't cause any unnecessary reordering - it preserves order of existing ACCEPT rules. There's probably a way to do it with fewer operations, I'd need to break out my sorting algorithms. OTOH, it doesn't do any operations if things are already sorted, so in the overwhelming common case this incurs no additional iptables operations.

I also add LOG rules to 'block' situations (the second commit).

internal ticket PAASTA-11763

### DIFF
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -75,6 +75,7 @@ class ServiceGroup(collections.namedtuple('ServiceGroup', (
 
     def update_rules(self, soa_dir, synapse_service_dir):
         iptables.ensure_chain(self.chain_name, self.get_rules(soa_dir, synapse_service_dir))
+        iptables.reorder_chain(self.chain_name)
 
     @property
     def log_prefix(self):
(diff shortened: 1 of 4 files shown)


bchess:
Hang on, getting "can't commit: Invalid argument" when running this on mesosstage. will reopen when I'm ready

bchess:
Ok, this should be working now. python-iptables doesn't like when you re-use the same rule objects for replace_rule.

PR #1303, opened 2017-06-13
Remove mesos.interface now that we use pymesos.

EvanKrall:
@jglukasik is having dependency version conflicts with protobuf -- mesos.interface needs < 3, signalform needs >3.5.


EvanKrall:
I will probably discard this and let @jglukasik make this change on his own branch, I just want to see if the tests pass without this.

solarkennedy:
I don't see any reason not to go ahead and remove this if the tests pass, as it looks like we really are not using it.

PR #1306, opened 2017-06-14
Allow all services access to DNS

chriskuehl:
internal ticket PAASTA-11288

### DIFF
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -4,10 +4,12 @@ from __future__ import unicode_literals
 
 import collections
 import hashlib
+import io
 import itertools
 import json
 import logging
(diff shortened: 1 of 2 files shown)


paasta_tools/firewall.py
```diff
@@ -247,7 +263,56 @@ def active_service_groups():
     return service_groups
 
 
-def ensure_internet_chain():
+def _dns_servers():
+    with io.open(RESOLV_CONF) as f:
+        for line in f.read().splitlines():
```
bchess: You could also just do 'for line in f:' right?

chriskuehl: ah, yes indeed! fixed

PR #1307, opened 2017-06-14
Add timeout to HTTP draining requests

### LINKED ISSUE #1287: We need more aggressive timeouts on drain/stop_draining
I'm trying to bounce a service for which marathon has some tasks that appear to be on a box that went away. We apparently have a very long timeout on the HTTP call in stop_draining, so it takes foreeeeeeeeeeeever to run setup_marathon_job.

jvperrin:
Fixes #1287

@EvanKrall Is a timeout of 15 seconds enough time? Do we want to catch the timeout and proceed anyway when in `stop_draining`?

### DIFF
--- a/paasta_tools/drain_lib.py
+++ b/paasta_tools/drain_lib.py
@@ -284,7 +284,11 @@ class HTTPDrainMethod(DrainMethod):
             'HEAD': requests.head,
         }[method]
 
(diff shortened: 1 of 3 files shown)


EvanKrall:
setup_marathon_job already ignores exceptions in stop_draining, so I think it's okay to let the exception bubble up.

PR #1308, opened 2017-06-15
Allow all services to access scribe, metrics-relay, sensu

chriskuehl:
To make things hopefully a little more obvious, this introduces a "common" chain which is used by _all_ services. That includes these 3 things as well as a jump to the DNS chain (which we also provide for free, but is slightly more complicated so gets its own chain).

Tested on mesosstage; could hit all these things with netcat after but not before.

### DIFF
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -132,12 +132,12 @@ def _default_rules(conf, log_prefix):
 
 
 def _well_known_rules(conf):
-    # All services get DNS
+    # Allow access to certain resources for all services by default.
     yield iptables.Rule(
         protocol='ip',
         src='0.0.0.0/0.0.0.0',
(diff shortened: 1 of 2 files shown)


EvanKrall:
Yeah, how hard would it be to make this configurable in SystemPaastaConfig?

chriskuehl:
I think it'd be pretty easy. I'll look into that.

bchess:
One possible catch is that it may also be needed within the paasta_docker_wrapper. Not impossible, but may complicate things.

chriskuehl:
I'll ticket moving it to a config. I'd like to merge this now since it's blocking us actually starting to use this in real services on stagef, but will come back to this soon.

EvanKrall:
Sounds good.

PR #1314, opened 2017-06-16
Adding a timed flock to firewall operations

bchess:
Introduce a `timed_flock` contextmanager to utils, and use it in the paasta_firewall stuff:
- firewall_update cron will flock upon calling general_update(). Exceptions are not caught, causing the cron to fail.
- firewall_update inotify daemon will flock upon calling ensure_service_chains(). Exceptions will log an error but not propagate, allowing the daemon to continue running.
- docker_wrapper will flock upon calling prepare_new_container(). Exceptions will print to stderr but not propagate. The underlying docker command will still be executed.

This presumes that /var/run/paasta is writable. /var/run/paasta/mac-address is already required for use by the docker_wrapper, so this shouldn't really be anything new.

Internal ticket PAASTA-11247

### DIFF
--- a/paasta_tools/docker_wrapper.py
+++ b/paasta_tools/docker_wrapper.py
@@ -24,10 +24,13 @@ import re
 import socket
 import sys
 
+import pysensu_yelp
+
 from paasta_tools.firewall import DEFAULT_SYNAPSE_SERVICE_DIR
 from paasta_tools.firewall import firewall_flock
 from paasta_tools.firewall import prepare_new_container
 from paasta_tools.mac_address import reserve_unique_mac_address
+from paasta_tools.monitoring_tools import send_event
 from paasta_tools.utils import DEFAULT_SOA_DIR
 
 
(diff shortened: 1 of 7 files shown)


paasta_tools/docker_wrapper.py
```diff
@@ -296,11 +297,15 @@ def main(argv=None):
...
+            try:
+                with firewall_flock():
+                    prepare_new_container(
+                        DEFAULT_SOA_DIR,
+                        DEFAULT_SYNAPSE_SERVICE_DIR,
+                        env_args['PAASTA_SERVICE'],
+                        env_args['PAASTA_INSTANCE'],
+                        mac_address)
+            except Exception as e:
+                print('Unable to add firewall rules: {}'.format(e), file=sys.stderr)
```
chriskuehl: I'm sort of worried about this breaking and us never noticing. I think it's right to not block container startup (could lead to a nasty outage), but I wonder if there's any way we can know when this is happening?

Failing open like this isn't the kind of failure anyone would ever really notice, I think.

bchess: yeah it's a valid point. Maybe we can send a sensu alert..

bchess:
Fairly large diff in the new commit. I added calls to sensu send_event() in docker_wrapper. @solarkennedy can you look this part over? In particular, I'm not sure about check_name. I figure we want this to stay CRITICAL if it fires anywhere. It may be OK on some hosts but not on others, but that's still bad. Should I munge the hostname in check_event?

paasta_tools/docker_wrapper.py
```diff
@@ -272,6 +276,44 @@ def append_cpuset_args(argv, env_args):
...
+                    instance,
+                    mac_address)
+        except Exception as e:
+            output = 'Unable to add firewall rules: {}'.format(e)
+            sensu_status = pysensu_yelp.Status.CRITICAL
+
+    if output:
+        print(output, file=sys.stderr)
+
+    send_event(
```
solarkennedy: I appreciate the thought to include sensu, but I'm really hesitant to give a shipit on adding this to the wrapper.

This thing has to be bullet proof. 

Even if it sends alerts like this, are they *really* actionable to the end user?

I would rather this thing just log to stderr so it will show up in the logs and output of mesos.

bchess: I don't have a strong opinion. I agree with @chriskuehl that it's probably easy to miss those messages. But I also agree that adding sensu introduces a bit more risk. The whole thing _is_ wrapped in a try/except, fwiw. But if you still think it should be removed, no quibbles from me.

bchess: @solarkennedy just checking you still want to remove the sensu?

solarkennedy: Yes please

bchess:
removed the sensu stuff. @chriskuehl good to go?

chriskuehl:
yup, lgtm

PR #1315, opened 2017-06-19
security-check documentation updates

transcedentalia:


paasta_tools/cli/cmds/security_check.py
```diff
@@ -45,14 +46,22 @@ def add_subparser(subparsers):
 
 
 def perform_security_check(args):
+    """It runs a few security-checks including:
+    - shellshock: is the version of bash running inside the container vulnerable to shellshock?
+    - docker container: is the container running as root?
+    - packages: does the container has the latest packages installed? If not, the output will contain the name of the
+    package that needs to be updated and to which version.
```
mattmb: I would probably keep the docstring "vaguer" as to what it tests: "Runs the security check command configured in x, checks return code and prints output..."

paasta_tools/cli/cmds/security_check.py
```diff
@@ -27,8 +27,9 @@
 def add_subparser(subparsers):
     list_parser = subparsers.add_parser(
         'security-check',
-        description='Performs a security check (alpha)',
-        help='Performs a security check (alpha)',
+        description='Performs a security check consisting in a few tests like shellshock vulnerability, the user '
```
mattmb: Technically this is a bit yelp specific because the open source version could be doing anything. That said I don't think it's likely anyone outside yelp is using this script so I think it's fine to add this help here.

Also "consisting of" ;-)

paasta_tools/cli/cmds/security_check.py
mattmb: Having said that I guess you did this so that it appears nicely in readthedocs? Maybe it would be okay to include some sort of link to our internal wiki from the alert rather than readthedocs...

paasta_tools/cli/cmds/security_check.py
transcedentalia: Removed the yelp specific part though.

And thanks for the spell check. :)

paasta_tools/cli/cmds/security_check.py
transcedentalia: Yes, that's true. I made the docstrings verbose for the readthedocs.

But hope it's better now. :)

PR #1320, opened 2017-06-20
Catch any exception running setup_marathon_job

mattmb:
I don't really like catching all exceptions but in this case it seems
reasonable just because it's hard to know everything that could go wrong
in setup_marathon_job, especially as it has to talk to marathon's API
and could end up getting a weird response.

The motivation is to prevent a service not being bounced again when
there happens to be some transient exception thrown in the worker.

### DIFF
--- a/paasta_tools/deployd/workers.py
+++ b/paasta_tools/deployd/workers.py
@@ -58,12 +58,15 @@ class PaastaDeployWorker(PaastaThread):
             self.log.info("{} processing {}.{}".format(self.name, service_instance.service, service_instance.instance))
             marathon_apps = marathon_tools.get_all_marathon_apps(self.marathon_client, embed_failures=True)
             bounce_timers.setup_marathon.start()
-            return_code, bounce_again_in_seconds = deploy_marathon_service(service=service_instance.service,
-                                                                           instance=service_instance.instance,
-                                                                           client=self.marathon_client,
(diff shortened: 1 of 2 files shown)


mattmb:
cc @EvanKrall, as we discussed. I haven't bothered with the exponential backoff for now, I think it's okay to try without and just keep an eye on our graphs if anything weird happens.

paasta_tools/deployd/workers.py
```diff
@@ -58,12 +58,15 @@ def run(self):
...
-                                                                           marathon_apps=marathon_apps)
+            try:
+                return_code, bounce_again_in_seconds = deploy_marathon_service(service=service_instance.service,
+                                                                               instance=service_instance.instance,
+                                                                               client=self.marathon_client,
+                                                                               soa_dir=marathon_tools.DEFAULT_SOA_DIR,
+                                                                               marathon_config=self.marathon_config,
+                                                                               marathon_apps=marathon_apps)
+            except Exception:
+                return_code, bounce_again_in_seconds = -1, 60
```
solarkennedy: I wish there was a simple way to make this exponentially backoff. Is there?
My thinking here is that if we start overloading marathon/mesos and get exceptions, we will continue to hit it harder and harder because these deploys will continue to back up.

mattmb:
I'm going to merge this and add the backoff in a new PR just to speed things along @solarkennedy

PR #1325, opened 2017-06-22
Remote-run on taskproc

keymone:
- drop adhoc scheduler
- build mesos executor stack with sync runner
- allow passing get_disk default
- use retrying executor with default (3) retries
- return relevant status code

this depends on https://github.com/Yelp/task_processing/pull/29

### DIFF
--- a/paasta_tools/paasta_remote_run.py
+++ b/paasta_tools/paasta_remote_run.py
@@ -23,23 +23,31 @@ import re
 import signal
 import string
 import sys
-import threading
 from datetime import datetime
 
+from task_processing.plugins.mesos.mesos_executor import MesosExecutor
+from task_processing.runners.sync import Sync
+
+from paasta_tools import mesos_tools
 from paasta_tools.cli.cmds.remote_run import add_common_args_to_parser
 from paasta_tools.cli.cmds.remote_run import add_start_args_to_parser
 from paasta_tools.cli.utils import figure_out_service_name
-from paasta_tools.frameworks.adhoc_scheduler import AdhocScheduler
-from paasta_tools.frameworks.native_scheduler import create_driver
+from paasta_tools.frameworks.native_service_config import load_paasta_native_job_config
 from paasta_tools.mesos_tools import get_all_frameworks
 from paasta_tools.mesos_tools import get_mesos_master
 from paasta_tools.utils import compose_job_id
+from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import get_code_sha_from_dockerurl
+from paasta_tools.utils import get_config_hash
+from paasta_tools.utils import get_docker_url
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import paasta_print
 from paasta_tools.utils import PaastaColors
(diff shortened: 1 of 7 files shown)


keymone:
another dependency: https://github.com/Yelp/task_processing/pull/29

paasta_tools/paasta_remote_run.py
```diff
@@ -109,9 +118,83 @@ def extract_args(args):
...
+        cluster,
+        system_paasta_config,
+        instance_type='paasta_native',
+        soa_dir=DEFAULT_SOA_DIR,
+        config_overrides=None):
+    native_job_config = load_paasta_native_job_config(
+        service,
+        instance,
+        cluster,
+        soa_dir=soa_dir,
```
Rob-Johnson: I think you've got some dupes in here?

paasta_tools/paasta_remote_run.py
```diff
@@ -147,39 +232,55 @@ def remote_run_start(args):
...
+
+    mesos_address = '{}:{}'.format(
+        mesos_tools.get_mesos_leader(), mesos_tools.MESOS_MASTER_PORT
     )
-    driver = create_driver(
+
+    paasta_config = system_paasta_config.get_paasta_native_config()
+    # TODO: implement DryRunExecutor?
+    mesos_executor = MesosExecutor(
+        role='*',
```
Rob-Johnson: I think we want the taskproc role, don't we?

paasta_tools/utils.py
```diff
@@ -257,13 +257,13 @@ def format_docker_parameters(self, with_labels=True):
         parameters.extend(self.get_cap_add())
         return parameters
 
-    def get_disk(self):
+    def get_disk(self, default=1024):
```
Rob-Johnson: I don't know why you'd want to parameterize this? if you want a specific value, set it in your config?

keymone: i want to be able to provide a sensible default in the caller. right now the default is hardcoded in this method, it's simply a matter of convenience to expose it as optional arg.

paasta_tools/paasta_remote_run.py
keymone: eventually we will have that value in paasta native config or a separate paasta taskproc config, current value is for testing.

paasta_tools/paasta_remote_run.py
```diff
@@ -109,9 +117,107 @@ def extract_args(args):
...
+    )
+
+    taskproc_config = system_paasta_config.get('taskproc')
+    # TODO: implement DryRunExecutor?
+    mesos_executor = MesosExecutor(
+        role=taskproc_config.get('role', taskproc_config['principal']),
+        principal=taskproc_config['principal'],
+        secret=taskproc_config['secret'],
+        mesos_address=mesos_address,
+        framework_name="paasta-remote %s %s %s" % (
```
Rob-Johnson: rather than using a random string for the runid, why not just use millis in the date field?

keymone: could be problematic, as all the tronfig jobs will use same service and instance.

PR #1327, opened 2017-06-22
Restore old signal handler on exit

tiras-j:
Very minor change. I was reading through @bchess's PR #1314 and noticed the Timeout context manager is not restoring existing signal handlers should they exist. Most likely this is not and never will be an issue, but it's probably better safe than sorry.

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -1627,11 +1627,12 @@ class Timeout:
         raise TimeoutError(self.error_message)
 
     def __enter__(self):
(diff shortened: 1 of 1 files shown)


bchess:
good catch, thanks!

solarkennedy:
@tiras-j I'm not really sure why Travis can't install docker, even after me kicking it.

Idk, can you try rebasing on master?

tiras-j:
@solarkennedy sure thing!

PR #1329, opened 2017-06-22
Make mark and wait for deployment commands accept a short git sha

nhandler:
It is rather frustrating not being able to simply copy/paste a short git sha displayed in other tools into `paasta wait-for-deployment`. This PR attempts to fix that. It modifies `mark-for-deployment` and `wait-for-deployment` to accept a short git sha as an argument. It will then attempt to convert this to a full sha and pass that full sha around internally.

This is internal ticket PAASTA-10534

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -211,7 +211,9 @@ def paasta_mark_for_deployment(args):
         refs = remote_git.list_remote_refs(args.git_url)
         commits = short_to_full_git_sha(short=args.commit, refs=refs)
         if len(commits) != 1:
-            raise ValueError("%s matched %d git shas. Must match exactly 1." % (args.commit, len(commits)))
+            raise ValueError(
+                "%s matched %d git shas (with refs pointing at them). Must match exactly 1." %
+                (args.commit, len(commits)))
         args.commit = commits[0]
 
     old_git_sha = get_currently_deployed_sha(service=service, deploy_group=args.deploy_group)
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -21,6 +21,7 @@ from __future__ import unicode_literals
 import logging
 import sys
 import time
+from argparse import ArgumentTypeError
 from threading import Event
 from threading import Thread
 
@@ -35,9 +36,11 @@ from paasta_tools.api import client
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_deploy_groups
 from paasta_tools.cli.utils import list_services
+from paasta_tools.cli.utils import short_to_full_git_sha
 from paasta_tools.cli.utils import validate_full_git_sha
 from paasta_tools.cli.utils import validate_given_deploy_groups
 from paasta_tools.cli.utils import validate_service_name
+from paasta_tools.cli.utils import validate_short_git_sha
 from paasta_tools.deployment_utils import get_currently_deployed_sha
 from paasta_tools.generate_deployments_for_service import get_cluster_instance_map_for_service
 from paasta_tools.utils import _log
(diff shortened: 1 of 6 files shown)


paasta_tools/cli/utils.py
```diff
@@ -731,6 +731,27 @@ def validate_given_deploy_groups(all_deploy_groups, args_deploy_groups):
...
+    """Converts a short git sha to a full sha
+
+    :param short: A short git sha represented as a string
+    :param refs: A list of refs in the git repository
+    :return: The full git sha or None if one can't be found
+    """
+    for sha in list(set(refs.values())):
+        if sha.startswith(short):
+            return sha
+    return None
```
solarkennedy: Are you sure `None` is an OK return value here? 

I think it should raise, otherwise we will get confusing behavior when none is passed through

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -81,7 +84,7 @@ def add_subparser(subparsers):
         '-c', '-k', '--commit',
```
solarkennedy: It's too bad there isn't some sort of cli args mixing thingy?

paasta_tools/cli/utils.py
Rob-Johnson: I think it should return [] and the caller should raise.

Rob-Johnson:
```python
return [sha for sha in set(refs.values()) if sha.startswith(short)]
```

howcome you're doing list(set(... ? to remove dupes? if so, you can leave it as a set and just iterate through that.

nhandler: Unless we make this another interactive prompt (which I'm still not a huge fan of), returning a list seems like more trouble than it is worth. Or I guess another option would be to go with the list and make the caller raise an exception if the list doesn't have exactly 1 element.

The list(set( stuff was leftover from some earlier testing. I'll get rid of it.

paasta_tools/cli/utils.py
```diff
@@ -738,10 +738,7 @@ def short_to_full_git_sha(short, refs):
     :param refs: A list of refs in the git repository
     :return: The full git sha or None if one can't be found
     """
-    for sha in list(set(refs.values())):
-        if sha.startswith(short):
-            return sha
-    return None
+    return [sha for sha in set(refs.values()) if sha.startswith(short)]
```
solarkennedy: I would rather have the error handling happen once here instead of by the caller.

There is no room for having ambiguous git shas.

Don't even "raise", just exit 1 and print the friendly error message. This is in cli utils, it should only be used by user-interacting cli tools.

Rob-Johnson: I disagree - this is a library function, and library functions should be un opinionated about how return values should be interpreted. The function is 'just' pattern matching, there is no need to force 0 matches to mean exit(1).  If it shouldn't be in a library, move it back up into the script.

solarkennedy: I'll concede.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -202,6 +205,15 @@ def paasta_mark_for_deployment(args):
     if args.git_url is None:
         args.git_url = get_git_url(service=service, soa_dir=args.soa_dir)
 
+    try:
+        validate_full_git_sha(args.commit)
+    except ArgumentTypeError:
+        refs = remote_git.list_remote_refs(args.git_url)
+        commits = short_to_full_git_sha(short=args.commit, refs=refs)
+        if len(commits) != 1:
+            raise ValueError("%s matched %d git shas. Must match exactly 1." % (args.commit, len(commits)))
```
EvanKrall: There's a shortcoming of the overall SHA-expanding strategy, which is that we'll only be able to expand SHAs which have refs currently pointing to them. This is probably the vast majority of things we'd want to mark for deployment (master, other branches, and anything previously marked for deployment would be fine).

But occasionally, someone will want to mark for deployment something without a ref currently pointing at it, so it would be good to mention this shortcoming in the error specifically when we find 0 matches.

PR #1331, opened 2017-06-26
Fixing the security-check notification system

transcedentalia:


bobtfish:
👍

paasta_tools/monitoring_tools.py
```diff
@@ -169,32 +170,44 @@ def send_event(service, check_name, overrides, status, output, soa_dir, ttl=None
     :param output: The output to emit for this event
     :param soa_dir: The service directory to read monitoring information from
     """
+    system_paasta_config = load_system_paasta_config()
+    sensu_host = system_paasta_config.get_sensu_host()
```
Rob-Johnson: doesn't look like these are used anywhere?

paasta_tools/monitoring_tools.py
```diff
@@ -169,32 +170,44 @@ def send_event(service, check_name, overrides, status, output, soa_dir, ttl=None
...
+    sensu_port = system_paasta_config.get_sensu_port()
+
+    if not sensu_host:
+        return
```
Rob-Johnson: what's wrong with the default?

solarkennedy:
Instead of making a new function (`send_event` AND `send_event_to_sensu`) can you adjust send_event to meet your needs?

Even better, just default to 'localhost' ?

PR #1336, opened 2017-06-28
Add exponential back off deployd

mattmb:
If a worker fails to bounce a service due to an exception then it will
back off trying exponentially. This is especially to be kind to Marathon
incase it is overwhelmed and causing the bounces to fail.

cc @solarkennedy

### DIFF
--- a/paasta_tools/deployd/workers.py
+++ b/paasta_tools/deployd/workers.py
@@ -5,6 +5,7 @@ import time
 
 from paasta_tools import marathon_tools
 from paasta_tools.deployd.common import BounceTimers
+from paasta_tools.deployd.common import exponential_back_off
 from paasta_tools.deployd.common import PaastaThread
 from paasta_tools.deployd.common import ServiceInstance
 from paasta_tools.setup_marathon_job import deploy_marathon_service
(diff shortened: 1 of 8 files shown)


paasta_tools/deployd/workers.py
```diff
@@ -65,8 +67,13 @@ def run(self):
...
                                                                                marathon_config=self.marathon_config,
                                                                                marathon_apps=marathon_apps)
-            except Exception:
-                return_code, bounce_again_in_seconds = -1, 60
+            except Exception as e:
+                self.log.warning("deploy_marathon_service caused exception: {}".format(e))
+                return_code = -2
+            if return_code != 0:
+                failures += 1
+                bounce_again_in_seconds = exponential_back_off(failures, factor=60, base=2, max_time=6000)
```
Rob-Johnson: this seems quite high to me - can we make it configurable so we've got an easy way of experimenting to find the right values?

mattmb: +1 to it being configurable. I've just pushed a commit that should do so and I've set it to 30 so that after the first failure we wait the same amount of time as we would wait before reprocessing a non failed run.

PR #1337, opened 2017-06-29
Reduce travis builder time

EvanKrall:


EvanKrall:
First hypothesis was that by combining the ~3m30s builds of py27, py3, coverage, and docs, I could save some time. (In particular, because we need to install docker and tox and stuff). This doesn't seem to have helped much -- my first run of this combined build was ~11:40 (so a savings of 2:20).

I'm now trying to cache .tox.

EvanKrall:
Caching .tox seems to have helped a good amount: from ~12 minutes to ~3.7 for the combined py27/py3/coverage/docs builder.

Hopefully we'll see some additional savings for the other build steps.

@chriskuehl This should be safe to do, as we have `tox_pip_extensions_ext_venv_update = true`, right?

chriskuehl:
@EvanKrall yup, should be totally safe -- this is a great use-case for it.

.travis.yml
```diff
@@ -4,12 +4,19 @@ branches:
     - /^v[0-9.]+$/
     - /^maci-.*$/
 language: python
-cache: pip
+cache:
+  pip: true
+  directories:
+    - .tox/
+    - $HOME/.cache/pip-faster/
+    - $HOME/.pre-commit/
```
asottile: should probably also cache `$HOME/.cache/pip`

EvanKrall: `cache: pip: true` [does that.](https://docs.travis-ci.com/user/caching/#pip-cache)

asottile: ah yes, missed that :)

.travis.yml
```diff
@@ -4,13 +4,21 @@ branches:
...
+  # We don't currently use any secure environment variables, but just in case, and also to avoid caches that grow forever, remove log dirs.
+  - rm -rf .tox/*/log .tox/log .tox/*/local/log
+  # For some reason (because we build the manpages toxenv inside docker, maybe?), caching .tox/manpages leads to failing itest_% builds.
+  - rm -rf .tox/manpages
 python:
   - "3.6"
-cache: pip
 env:
-  - TOXENV=py36
-  - TOXENV=coverage
```
solarkennedy: I would be ok with dropping coverage counting in favor of faster builds

EvanKrall: Once you put it in the same travis build as py36, it only takes a few seconds.

asottile: imo there's little reason to have both py36 and coverage, just run coverage always (unlike old versions of coverage, the overhead is essentially zero)

PR #1338, opened 2017-06-29
Fix maintenance watcher

mattmb:
* Handle timeouts/connection errors speaking to mesos
* Handle the fact that a host can finish draining and then start draining again in the future!

### DIFF
--- a/paasta_tools/deployd/watchers.py
+++ b/paasta_tools/deployd/watchers.py
@@ -147,15 +147,23 @@ def get_marathon_client_from_config():
 class MaintenanceWatcher(PaastaWatcher):
     def __init__(self, inbox_q, cluster, **kwargs):
         super(MaintenanceWatcher, self).__init__(inbox_q, cluster)
-        self.draining = []
+        self.draining = set()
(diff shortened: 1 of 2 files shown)


paasta_tools/deployd/watchers.py
```diff
@@ -147,15 +148,27 @@ def get_marathon_client_from_config():
...
+            self.draining.remove(host)
+        return new_draining_hosts
+
     def run(self):
         self.is_ready = True
         while True:
-            draining_hosts = get_draining_hosts()
-            new_draining_hosts = [host for host in draining_hosts if host not in self.draining]
-            self.draining = draining_hosts
+            new_draining_hosts = self.get_new_draining_hosts()
```
solarkennedy: I didn't realize we are watching this every 20s. I'm pretty sure this endpoint is relatively expensive. We hit it too hard (and hence why we see errors from it often. Try cranking it down to 60s

mattmb: Hmm, yes I was thinking about this too. The downside to making it longer is making the draining for the cluster autoscaler slower...

I think we should at least make it a config variable.

PR #1340, opened 2017-06-29
Make deployd prioritise bouncing services on start

mattmb:
When deployd starts it must evaluate every service with a run of setup
marathon job to ensure it hasn't "missed" an event (e.g. autoscaling or
deployment). However, when we start we can prioritise services that we
know are deploying becuase their app ID has changed. This way if there
is a leader election or restart of deployd: bouncing/deploying services
will be at the front of the queue when it starts up.

Note the bulk of this is just moving stuff around. `prioritise_bouncing_services` is the new bit.

### DIFF
--- a/paasta_tools/deployd/common.py
+++ b/paasta_tools/deployd/common.py
@@ -8,6 +8,13 @@ from threading import Thread
 
 from six.moves.queue import Queue
 
+from paasta_tools.marathon_tools import DEFAULT_SOA_DIR
+from paasta_tools.marathon_tools import get_marathon_client
+from paasta_tools.marathon_tools import list_all_marathon_app_ids
+from paasta_tools.marathon_tools import load_marathon_config
+from paasta_tools.marathon_tools import load_marathon_service_config_no_cache
+from paasta_tools.utils import NoDockerImageError
+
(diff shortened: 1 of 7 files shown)


paasta_tools/deployd/common.py
```diff
@@ -62,3 +69,27 @@ def rate_limit_instances(instances, number_per_minute, watcher_name):
 def exponential_back_off(failures, factor, base, max_time):
     seconds = factor * base ** failures
     return seconds if seconds < max_time else max_time
+
+
+def get_service_instances_with_changed_id(marathon_client, instances, cluster):
```
solarkennedy: Don't we already have this logic? Isn't this the oracle?

mattmb: This is already in deployd, just moving it around. The oracle is slightly different because it is also looking at number of at risk tasks iirc. That said, this doesn't find apps which have both new and old apps in Marathon and I think it probably should...

I'll have a bit more of a think before we merge this

mattmb: Looking at the oracle code I think it's actually exactly what we want here. Whereas the config watchers just want to know if the newly configured app doesn't exist, here we want to know if there are any differences between configured and running apps which is what the oracle code does. I'm going to switch it to use that code.

PR #1345, opened 2017-07-04
Durable storage of information about tasks for native scheduler

EvanKrall:
This branch reworks how we store information about tasks in the native scheduler -- instead of a simple dictionary tasks_with_params, we now use a pluggable task store implementation. There are two implementations defined: DictTaskStore (for testing), and ZKTaskStore (for real-world use.)

Having durable storage allows us to remember useful things, such as:
- what box is this task running on?
- what port did we allocate for this task?

In this branch I've also improved our calls to drain methods, to include host/port. This should fix hacheck draining for the native scheduler.


paasta_tools/frameworks/task_store.py
```diff
@@ -0,0 +1,180 @@
...
+        """Return a merged MesosTaskParameters object, where attributes in other take precedence over self."""
+
+        new_dict = self.__dict__
+        new_dict.update(kwargs)
+
+        return MesosTaskParameters(**new_dict)
+
+    @classmethod
+    def deserialize(cls, serialized_params):
+        return cls(**json.loads(serialized_params))
```
solarkennedy: I guess loads doesn't need utf8 hints?

paasta_tools/frameworks/native_scheduler.py
```diff
@@ -422,7 +406,7 @@ def kill_tasks_if_necessary(self, driver):
         )
         new_tasks_to_kill = new_tasks_by_desirability[desired_instances:]
 
-        old_tasks = self.get_old_tasks(base_task.name, self.tasks_with_flags.keys())
+        old_tasks = self.get_old_tasks(base_task.name, self.task_store.get_all_tasks().keys())
```
keymone: now that this could be somewhere behind the network it's probably worth caching the first call to `get_all_tasks` and reusing here and on line 427?

paasta_tools/frameworks/task_store.py
```diff
@@ -0,0 +1,180 @@
...
+            self.overwrite_task(task_id, MesosTaskParameters(**kwargs))
+
+    def update_task(self, task_id, **kwargs):
+        existing_task = self.get_task(task_id)
+        if existing_task:
+            merged_params = existing_task.merge(**kwargs)
+        else:
+            merged_params = MesosTaskParameters(**kwargs)
+
+        self.overwrite_task(task_id, merged_params)
```
keymone: i would feel better if this was a CAS operation or update_task was wrapped in a lock, but if we're certain this won't ever get called from multiple threads - ship

EvanKrall: I've overridden update_task in ZKTaskStore so that it passes to set() the version received when reading the data, and retries if this fails. I've also better handled the case where someone else created the task between us reading it (and not seeing anything) and trying to create it.

paasta_tools/frameworks/native_scheduler.py
EvanKrall: Probably! I'll work on that.

paasta_tools/frameworks/task_store.py
EvanKrall: nope, [json docs](https://docs.python.org/3/library/json.html) say it can be "a str, bytes or bytearray instance containing a JSON document"

paasta_tools/frameworks/task_store.py
```diff
@@ -0,0 +1,206 @@
...
+        is_healthy=None,
+        offer=None,
+        resources=None,
+    ):
+        self.__dict__['health'] = health
+        self.__dict__['mesos_task_state'] = mesos_task_state
+        self.__dict__['is_draining'] = is_draining
+        self.__dict__['is_healthy'] = is_healthy
+        self.__dict__['offer'] = offer
+        self.__dict__['resources'] = resources
```
asottile: sorry to grave dig an old commit, but I thought you should know :)

assigning to `self.__dict__['x'] = ...` is better written as `self.x = ...`

I initially thought this was an attempt to side-step mypy, but mypy wasn't in play at this point...

asottile: oh I see the `__setattr__` / `__delattr__` below, tricky tricky

PR #1350, opened 2017-07-06
Fix deployd updating instance count

mattmb:


paasta_tools/deployd/common.py
```diff
@@ -71,19 +71,25 @@ def exponential_back_off(failures, factor, base, max_time):
...
     service_instances = []
     for service, instance in instances:
         config = load_marathon_service_config_no_cache(service=service,
                                                        instance=instance,
                                                        cluster=cluster,
                                                        soa_dir=DEFAULT_SOA_DIR)
         try:
-            config_app_id = config.format_marathon_app_dict()['id']
+            config_app = config.format_marathon_app_dict()
+            app_id = '/{}'.format(config_app['id'])
```
Rob-Johnson: howcome you're changing this too? did this not work at all previously?

mattmb: Yeah it was just being stripped in another function before the comparison. Now I need the marathon app objects too, not just their IDs.

PR #1351, opened 2017-07-07
Filling in the description field for security-check sensu email/ticket

transcedentalia:
In preparation for the sensu email/Jira notification that should eventually be fixed soon, I've realized the output from the _run call is empty so I replaced is with _ and passed a more relevant message.

### DIFF
--- a/paasta_tools/cli/cmds/security_check.py
+++ b/paasta_tools/cli/cmds/security_check.py
@@ -56,13 +56,15 @@ def perform_security_check(args):
         paasta_print("Nothing to be executed during the security-check step")
         return 0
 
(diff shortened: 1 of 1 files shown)


solarkennedy:
Can you instead "fix" the bug and have the `_run` wrapper add to the output variable, even in stream mode? Hint is in utils.py:1299

I always like to have as much context as possible in alerts sent to users to "save them a click"

solarkennedy:
@transcedentalia can you address the feedback and rebase?

solarkennedy:
@transcedentalia please rebase and address the feedback.

PR #1352, opened 2017-07-07
Allow for https healthchecks

jglukasik:
verify=False because I believe the certs here will be self-signed. Or, should I do this another way?


solarkennedy:
Are there any docs to update?

PR #1354, opened 2017-07-11
Ask marathon to embed tasks not failures

mattmb:
For some calls to /v2/apps we need to embed the tasks in the results. I
don't think we need the task failures and embedding them is deprecated
because it implies embedding everything else. See:
https://mesosphere.github.io/marathon/docs/rest-api.html#get-v2apps

Unfortunately embedding tasks still implies embedding deployments which
I think we also don't need. However, this is supposed to change at some
point in the future.

If the itests pass then I think this is safe and it should reduce the
load on marathon slightly.

### DIFF
--- a/paasta_itests/steps/bounces_steps.py
+++ b/paasta_itests/steps/bounces_steps.py
@@ -216,7 +216,7 @@ def when_setup_service_initiated(context):
         # 120 * 0.5 = 60 seconds
         for _ in range(120):
             try:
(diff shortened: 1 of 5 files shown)


mattmb:
@EvanKrall any thoughts before I merge this. I had a go curling the marathon API with `?embed=apps.tasks` and it seems to return the tasks. Do you think we've missed anything we need, the itests give me hope we will be okay.

PR #1355, opened 2017-07-11
Catch InvalidJobNameError in deployd

mattmb:
I'm not 100% on the root cause of this exception but the deploy daemons
file watcher certainly shouldn't die as a result. The safest thing is to
catch it and assume the service needs a bounce. Then it will be up to
the worker to have a go at deploying the service and back off it sees
the same error.


solarkennedy:
I'm sure it happens whenever someone deletes a job?

PR #1356, opened 2017-07-11
paasta status by owner

matthewbentley:
~~This works, despite the triply nested for loops.~~

~~But reviews are welcome~~

This is a rather large change.  The big piece is to remove `paasta_args_mixer`, and replace it with a function that takes the `args` and returns a dict of clusters:services:instances that match the args.  It does this by getting all the instances, and then filters them based on what is specified in the args.  The rest of this is fixing up tests and other cleanups.

This does remove the functionality for suggesting instances if none of them match.

A less destructive version of this is 2b8b54c7171fc2ebc209eda6c84a6c7da00c8b5c


matthewbentley:
It looks like the reason paasta status doesn't have any itests is paasta status requires ssh to be working.

matthewbentley:
@EvanKrall should I try to remove the paasta_args_mixer entirely?  Maybe by splitting it into a function to get the cluster whitelist and a function to get the instance whitelist?

solarkennedy:
@oktopuz can you comment on the paasta_args_mixer whitelist thingy? and review this?

oktopuz:
@EvanKrall, I have to admit I'm guilty of not using existing approach and introducing paasta_args_mixer. I'll try to explain why I did this.

The only items we actually need to get status are cluster, service and instances.
Neither marathon nor chronos does not know anything about deploy-groups or owners.
These two are interface options of 'paasta status' which we add to provide more suitability.

The problem I saw with whitelists was that to add a new option we have to repeat ourself many times.

For instance, to add --deploy-group to 'paasta status' we need to add deploy_group_whitelist as a parameter to paasta_status(), report_status(), report_status_for_cluster() and teach them how to work with it (https://github.com/Yelp/paasta/commit/c24b09c5e6029b7d25068c7715aca55355e05f95).

Next to add --owner to the status command, we need to update all these functions once again.

Then if we want to use these command line options in paasta start, stop or restart commands, we need to repeat almost the same code at paasta_start_or_stop() and write new tests to verify it. 

And so on.

That is a lot of repetition indeed. 

The idea of paasta_args_mixer was to get rid of all possible auxiliary options at the very first stage, so that we don't have to add new whitelists every time we add a new option.

It should get the argument list and return cluster, service and instances only, so that we can re-use it easily for any other command (see bba3222972f47d281d3b17d5a182ce973ca52fac).

@matthewbentley, if you're going to continue using the paasta_args_mixer patten, you should move  'if args.owner/else' from paasta_status() to paasta_args_mixer() and teach it how to work with more than 1 service.

I'd move figure_out_service_name() to paasta_args_mixer as well, so that it accepts 'args' only and returns a list of (service, clusters, instances) tuples.

Another options is to stop using paasta_args_mixer in paasta_status() as @EvanKrall suggests and follow https://github.com/Yelp/paasta/commit/c24b09c5e6029b7d25068c7715aca55355e05f95

EvanKrall:
Yeah, actually, I rescind my criticism of paasta_args_mixer; I agree that it's cleaner to simply pass a list of instances report_status_for_cluster instead of all the whitelist options.

Currently we're passing the same list of instances to report_status_for_cluster for every cluster. I don't see any cases where we're doing this in our current soa-configs, but if we had a service that looked like this:

```
$ cat marathon-prod.yaml
canary:
  deploy_group: canary_and_stage

main:
  deploy_group: prod.main
$ cat marathon-stage.yaml
main:
  deploy_group: canary_and_stage
```

I believe `paasta status --deploy-group canary_and_stage` would currently report prod's main instance, even though it isn't in the canary_and_stage deploy group. I think a similar bug could apply for owners -- if one team owned the `main` instance on one cluster, but another team owned the `main` instance on another cluster, we might accidentally report `main` in both clusters.

A fix would be to have paasta_args_mixer return data of the format cluster -> list of service/instance:
```
{
  "prod": [
    ("servicename", "canary"),
  ],
  "stage": [
    ("servicename", "main"),
  ],
}
```
or a list of tuples (cluster, service, instance).

paasta_tools/utils.py
```diff
@@ -1551,6 +1552,80 @@ def list_clusters(service=None, soa_dir=DEFAULT_SOA_DIR, instance_type=None):
...
+
+    :param clusters: The clusters to get configs for
+    :returns: A dict in the format {cluster_name: {service_name: {instance_name: {instance_conf}, ...}, ...}, ...}
+    """
+    confs = defaultdict(dict)
+    for t in INSTANCE_TYPES:
+        for c in clusters:
+            services = get_instance_configs(t, c, soa_dir)
+            for service, instance in services.items():
+                if service not in confs[c].keys():
```
matthewbentley: That doesn't seem to work, since `defaultdict` takes a type and `defaultdict(dict)` is not callable.  I think I can actually fix this though with `defaultdict(lambda: defaultdict(dict))`.  From here: https://stackoverflow.com/questions/9358983/python-dictionary-and-default-values/23933745#23933745

paasta_tools/utils.py
```diff
@@ -1551,6 +1552,80 @@ def list_clusters(service=None, soa_dir=DEFAULT_SOA_DIR, instance_type=None):
...
+
+
+def get_instances_by_owner(owners, clusters, soa_dir=DEFAULT_SOA_DIR):
+    """Get all service instances with onwer in a list
+
+    :param owners: A list of owners to match against
+    :param clusters: A list of clusters to look for services in
+    :returns: A dict of service_instances, in the format {service_name: [instance1, instance2, ...], ...}
+    """
+    all_instances = {s for c in clusters for s in get_services_for_cluster(cluster=c, soa_dir=soa_dir)}
```
matthewbentley: I think this is related to @EvanKrall's comment above: we seem to assume that, for many settings, there will be no difference between clusters.  I'll see if I can implement his suggested fix.

matthewbentley:
So I re-wrote `paasta_args_mixer` (to the point where it should probably be renamed), to (I hope) be a bit more flexible.  However, it seems to be a bit slow (maybe it's not?), and I need to re-add some of the error checking, and fix some tests.

But this will (well, should) allow us to do something like
```
paasta status -c norcal-devc,mesosstage -o paasta,operations -i main,canary
```
to see all the main and canary instances of any service on norcal-devc or mesosstage owned by paasta or operations.

matthewbentley:
TODO: add a filter for job type (ie marathon, chronos, adhoc, etc)

paasta_tools/utils.py
```diff
@@ -1567,6 +1571,79 @@ def list_clusters(service=None, soa_dir=DEFAULT_SOA_DIR, instance_type=None):
...
+    :returns: A dict of service_instances, in the format {cluster: {service_name: [instance1, instance2, ...], ...}}
+    """
+    all_instances = {c: get_services_for_cluster(cluster=c, soa_dir=soa_dir) for c in clusters}
+    service_confs = get_all_service_configs(soa_dir)
+    instance_confs = get_all_instance_configs(clusters, soa_dir)
+
+    cluster_services_instances = defaultdict(lambda: defaultdict(set))
+    for cluster, services_instances in all_instances.items():
+        for service, instance in services_instances:
+            service_team = service_confs.get(service, {}).get('monitoring', {}).get('team', None)
```
solarkennedy: this looks weird to me, are you not able to use the existing monitoring_tools.get_team function? (that does this logic already)

tests/cli/test_cmds_start_stop_restart.py
```diff
@@ -449,9 +450,11 @@ def test_cluster_throws_exception_no_matching_instance_clusters(
...
-    mock_figure_out_service_name.return_value = 'fake_service'
+    mock_apply_args_filters.return_value = {
+        'fake_cluster_2': {'fake_service': ['i1']},
+    }
 
     assert args.command(args) == 1
     output, _ = capfd.readouterr()
     assert "Invalid cluster name(s) specified: fake_cluster_3" in output
-    assert "Valid options: fake_cluster_1 fake_cluster_2" in output
+    # assert "Valid options: fake_cluster_1 fake_cluster_2" in output
```
solarkennedy: just remove this if it no longer true, or assert the opposite or whatever

solarkennedy:
I'm going to make the call: please make a new PR that is just the args mixer rewrite.

matthewbentley:
@solarkennedy Is the current design of the args mixer rewrite good to go?

matthewbentley:
closing in favor on #1474

PR #1358, opened 2017-07-12
Only write yaml services file when there is a difference in config content

thebostik:
This will prevent services with watchers from reloading this config on the cron schedule, which is expensive to parse. 
PERF-1673


ArnaudBrousseau:
This PR looks good to me 👍 

Deferring to @solarkennedy for final ship/merge!

PR #1359, opened 2017-07-12
paasta_oom_finder, initial commit

oktopuz:
internal ticket: PAASTA-567

This script looks for OOM records in /var/log/messages.
When it finds something, it tries to obtain service and instance name using 'docker inspect' and generates a record in the following format:

```
date/time hostname container_id service.instance
```

I'm going to run the script via cron every minute and stream its output to kafka using stdin2scribe.

### DIFF
--- a/paasta_tools/oom_finder.py
+++ b/paasta_tools/oom_finder.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python
+# Copyright 2015-2017 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from __future__ import absolute_import
+from __future__ import unicode_literals
+
+import os
+import re
+from tempfile import gettempdir
+
+from docker.errors import APIError
+
+from paasta_tools.utils import get_docker_client
+
+
+def position_file_name():
+    """Return an absolute path of the position file.
+
+    See tail_log_file() for details.
+    """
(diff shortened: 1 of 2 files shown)


bplotnick:
Could you do this directly in syslog-ng using filters?

oktopuz:
@bplotnick, I thought about syslog filters but decided to not use them.

syslog filtering works well for one pattern, but imaging if we need to add more information to this log record (for example, process name, used memory and etc).

```
Task in /docker/efd2cdf4f1848d2ff62a81848102c872a64f1d122d89a20b723fe6a824968264 killed as a result of limit of /docker/efd2cdf4f1848d2ff62a81848102c872a64f1d122d89a20b723fe6a824968264
Memory cgroup stats for /docker/efd2cdf4f1848d2ff62a81848102c872a64f1d122d89a20b723fe6a824968264: cache:168KB rss:102232KB rss_huge:2048KB mapped_file:4KB dirty:0KB writeback:0KB swap:0KB inactive_anon:20KB active_anon:100264KB inactive_file:0KB active_file:0KB unevictable:0KB
Memory cgroup out of memory: Kill process 118071 (python) score 1145 or sacrifice child
Killed process 118071 (python) total-vm:415788kB, anon-rss:100040kB, file-rss:16652kB
```

When I read /var/log/messages directly, it is relatively easy to find and combine all lines related to the same event. But I have no idea how to combine this information, when syslog-ng sends these lines one by one.

Also it may take some time to run docker inspect. Especially when docker is broken. I'm not sure how syslog deals with such issues.

ronin13:
Rather than maintaining cursor position in a separate file, and using docker inspect, isn't it better to use docker events --since 'Xm' --filter 'type=oom'  --format <json> (where X is the periodicity of the cron job).  This should also allow for simpler parsing. The only caveat is if docker daemon restarts (within that Xm window), then the history for last X minutes will be lost (which I guess is acceptable given that X is 1 here).

This should also avoid issues like:

> But I have no idea how to combine this information, when syslog-ng sends these lines one by one.

oktopuz:
@ronin13, if an event occurs at the same moment that cron runs 'docker events' there is a chance that the event will be missed or duplicated.

On the second thought I realized that adding additional information to the log is convenient but not necessary indeed, because we can always find what we need on our log servers.

I'm going to ship it as it is now to start collecting logs, and then change the script to work with syslog-ng filters as @bplotnick proposed.

PR #1363, opened 2017-07-13
Apollo 205 detector status

philipmulcahy:
New paasta verb: detector-status.
Intended use-case - automated deployment pipelines can use ```paasta --detector_status --detector_name=$MY_CANARY_DETECTOR``` in order to determine whether to proceed, rollback or stop.

Partial test in example cluster (doesn't demonstrate plumbing-in of SFX token via puppet).

```
root@088852dd120f:/work# export SFX_TOKEN="*************************"
root@088852dd120f:/work# paasta detector-status --detector_name=all | jq '.|length'
/usr/local/lib/python2.7/dist-packages/requests/packages/urllib3/util/ssl_.py:90: InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from configuring SSL appropriately and may cause certain SSL connections to fail. For more information, see https://urllib3.readthedocs.org/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning
349
root@088852dd120f:/work# 
```


philipmulcahy:
dang!  this code was prototyped in 3.6, but my timing s*cks!
I'll go away, lick my wounds and come back later.

PR #1364, opened 2017-07-13
Dedupe the bounce queue

mattmb:
If a service is in the backlog to be bounced and there is a change to
its marathon-.yaml or deployments.json the the deploy daemon will notice
the change and if the new app that is required is not yet created then
it will add a second instance to the bounce queue. This way the queue
keeps growing.

This makes the BounceQueue drop any duplicate service instances that are
already in the queue.

### DIFF
--- a/paasta_tools/deployd/master.py
+++ b/paasta_tools/deployd/master.py
@@ -29,6 +29,37 @@ from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import ZookeeperPool
 
 
+class DedupedQueue(PaastaQueue):
+    """This class extends the python Queue class so that the Queue is
+    deduplicated. i.e. there can be only one copy of each service instance
+    in the queue at any one time
+    """
+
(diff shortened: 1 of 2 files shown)


paasta_tools/deployd/master.py
```diff
@@ -29,6 +29,33 @@
 from paasta_tools.utils import ZookeeperPool
 
 
+class BounceQueue(PaastaQueue):
```
solarkennedy: What do you think about calling this "DedupedQueue"?

Also, is there any reason why this shouldn't be part of the main queue class and used for every queue? In what situation in this deployd use-case where we want duplicates anyway?

mattmb: 👍 to the better naming.

The inbox queue actually can have multiple copies of the same service instance with different times that they should bounce by. Then the inbox thread de-dupes when it reads off that queue and only stores service instances once and only overwrites when they are being bumped in priority so to speak.

PR #1366, opened 2017-07-13
Upgrade pre-commit and hooks

asottile:
congrats, this is now the largest project that I can (probably) commit to that has pre-commit -- hope you don't mind me using it as a test bed :)


asottile:
That's the next commit :D

PR #1367, opened 2017-07-14
Use asottile/add-trailing-comma

asottile:
This is the more-safe and battle-tested v0.3.0 -- I'll follow up with the (not yet completed) 1.x release once I get some testing with it done :)


solarkennedy:
Thanks!

PR #1368, opened 2017-07-14
Add haproxy support for mesos leader discovery

charleskwwan:
Previous request: https://github.com/Yelp/task_processing/pull/36
## Summary:
- Added get_mesos_proxy to mesos_tools
- Added accompanying tests
## Testing done:
- Manually called get_mesos_proxy to ensure it works with services.yaml
- make test
## Notes:
- Initially wanted to do a plugin for taskproc, but concluded that it would be too roundabout because in TaskProcessor, we would need to pass the class MesosExecutor and have the would-be plugin initialize it with the smartstack address. Paasta never loads the mesos plugin itself, which is confusing.
- I did not modify remote-run because we do not yet have a taskproc cluster in prod, so remote-run still uses Zookeeper via get_mesos_leader, also in mesos_tools. However, with the new function, we can test more easily without having to manually get the address ourselves.
- I looked around Paasta's smartstack_tools to find a way to access services in a more Paasta-like way, but couldn't find anything. If anyone has any idea better than looking at services.yaml, please comment!

### DIFF
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -24,6 +24,8 @@ from collections import namedtuple
 
 import humanize
 import requests
+import yaml
+from environment_tools import type_utils
 from kazoo.client import KazooClient
 from six.moves.urllib_parse import urlparse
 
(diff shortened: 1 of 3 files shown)


charleskwwan:
Travis tests are failing because the latest version of environment-tools 1.1.3 is not available yet on PyPI. Will talk to @EvanKrall to try to get this resolved.

paasta_tools/mesos_tools.py
```diff
@@ -47,6 +49,7 @@
 
 
 DEFAULT_MESOS_CLI_CONFIG_LOCATION = "/nail/etc/mesos-cli.json"
+DEFAULT_SERVICES_LOCATION = "/nail/etc/services/services.yaml"
```
sagar8192: Can we make this configurable? If someone wants to use it outside of yelp, then they should be able to configure it rather than changing the code.

paasta_tools/mesos_tools.py
```diff
@@ -104,6 +107,34 @@ def get_mesos_leader():
...
+                ltype,
+                'region',
+            )[0]
+        except KeyError:
+            pass
+
+    if not region:
+        raise KeyError("{} does not exist".format(location))
+
+    proxy_prefix = 'task_processing.mesos_leader_paasta_'
```
sagar8192: nit: Move this to top of the script with other constants.

paasta_tools/mesos_tools.py
```diff
@@ -104,6 +107,34 @@ def get_mesos_leader():
...
+        services = yaml.load(istream)
+
+    region = None
+    for ltype in type_utils.available_location_types():
+        try:
+            region = type_utils.convert_location_type(
+                location,
+                ltype,
+                'region',
+            )[0]
```
sagar8192: Add break after this line? If we have found the region already, then we don't have to iterate over rest of the regions?

paasta_tools/mesos_tools.py
charleskwwan: This isn't meant to be a plugin, just an import. If I made it to be the latter, paasta tools would have to be given the executor to initialize so that it could pass in the address. I'll make it configurable, but given that we are going paasta is going to initialize taskproc anyway, it will either have to have the info elsewhere, or allow the script to take options.

paasta_tools/mesos_tools.py
charleskwwan: Fair.

paasta_tools/mesos_tools.py
charleskwwan: Agreed.

paasta_tools/mesos_tools.py
sagar8192: I mean, this adds a dependency on paasta. If someone wants to use smartstack in taskproc, then they have to add paasta as a dependency. Also, paasta has its own utils to figure out master address. Does it make sense to keep this in paasta?

PR #1369, opened 2017-07-14
Scribereader was not py3 compatible until 0.2.6

jolynch:
Found by acoover, paasta logs doesn't work because scribereader had non absolute imports until 0.2.6 ...

I've upgraded to 0.2.6 and am testing paasta logs locally from the lucid package.


jolynch:
Local testing confirms that indeed paasta logs works now.

PR #1372, opened 2017-07-16
Resolve more autopep8 / add-trailing-comma conflicts

asottile:
Fixed a few more places where autopep8 and add-trailing-commas disagreed -- these weren't conflicting before the last branch as add-trailing-comma was missing some cases.


asottile:
woops! found another regression -- I'll fix that too :S

PR #1376, opened 2017-07-17
Python3 bytes

solarkennedy:
* fixes bytes handling from zk by explicitly decoding
* fixes a bug in paasta rollback that deals with tuple comparison instead of string comparison

### DIFF
--- a/tests/autoscaling/test_autoscaling_service_lib.py
+++ b/tests/autoscaling/test_autoscaling_service_lib.py
@@ -159,11 +159,10 @@ def test_pid_decision_policy():
     }
 
     with mock.patch(
-        'paasta_tools.utils.KazooClient', autospec=True,
-        return_value=mock.Mock(
+        'paasta_tools.utils.KazooClient', autospec=True, return_value=mock.Mock(
             get=mock.Mock(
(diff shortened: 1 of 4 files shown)


tests/autoscaling/test_autoscaling_service_lib.py
```diff
@@ -286,7 +286,7 @@ def test_mesos_cpu_metrics_provider():
     with mock.patch(
         'paasta_tools.utils.KazooClient', autospec=True,
         return_value=mock.Mock(get=mock.Mock(
-            side_effect=lambda x: (zookeeper_get_payload[x.split('/')[-1]], None),
+            side_effect=lambda x: (str(zookeeper_get_payload[x.split('/')[-1]]).encode('utf-8'), None),
```
asottile: strange that you need `str()` here

solarkennedy: the test inputs here are sometimes ints and stuff, I wanted to make sure the mock returned what zk has exactly (bytes of a floaty string)

asottile: maybe change the mapping type above to just have strings? (shrugs)

PR #1381, opened 2017-07-18
Fix worker exception handling

mattmb:
This turned into a larger diff than I expected...
I'll have another read through before someone reviews this.


mattmb:
Ok, I think this is ready to review. It basically splits out the processing of a service instance from the while loop that picks service instances out of the queue. It also adds a catch on any exceptions that happen in the processing function

PR #1382, opened 2017-07-18
paasta_oom_logger

oktopuz:
paasta_oom_logger replaces https://github.com/Yelp/paasta/pull/1359

It is supposed to be used as a syslog-ng destination that sends JSON-encoded messages to the Scribe stream 'tmp_paasta_oom_events'.

Sample output:
```
{"timestamp": 1500402947, "hostname": "some_host", "container_id": "12f2d3f2e93f", "service": "some_service", "instance": "some_instance"}
{"timestamp": 1500402955, "hostname": "some_host", "container_id": "12f2d3f2e93f", "service": "some_other_service", "instance": "some_other_instance"}
```

I will revert https://github.com/Yelp/paasta/pull/1359 later.


solarkennedy:
Why not use the normal paasta log function, which already emits json to the right scribe log and will show up in `paasta logs` ?

setup.py
```diff
@@ -116,6 +116,7 @@
             'paasta_firewall_update=paasta_tools.firewall_update:main',
             'paasta_firewall_logging=paasta_tools.firewall_logging:main',
             'paasta_oom_finder=paasta_tools.oom_finder:main',
```
solarkennedy: Can you remove this script as this one replaces it? (I don't want to slightly different scripts for the same thing)

paasta_tools/utils.py
```diff
@@ -707,6 +707,12 @@ def default(text):
             'help': 'Logs from security-related services such as firewall monitoring',
         },
     ),
+    (
+        'oom', {
```
solarkennedy: Can you add this to the list of DEFAULT_COMPONENTS so people will see them without any extra paasta logs cli args?

paasta_tools/oom_logger.py
```diff
@@ -0,0 +1,125 @@
...
+            docker_inspect = client.inspect_container(resource_id=container_id)
+        except (APIError):
+            continue
+        env_vars = get_container_env_as_dict(docker_inspect)
+        log_line = LogLine(
+            timestamp=timestamp,
+            hostname=hostname,
+            container_id=container_id,
+            cluster=cluster,
+            service=env_vars.get('PAASTA_SERVICE', ''),
```
solarkennedy: I guess default to "Unknown" for service/instance here instead of empty string.

setup.py
oktopuz: Done.

paasta_tools/utils.py
oktopuz: Done.

paasta_tools/oom_logger.py
oktopuz: Done. I hope we'll  never have a service called 'unknown' :)

PR #1386, opened 2017-07-20
Use bytes when zk.set and handle mesos read timeout exceptions

solarkennedy:
This solves a couple of things.

I still have this (yelp only): 
https://fluffy.yelpcorp.com/i/zDSLwRsxQRtp1t8kVN7vbCCgtTbxRcWV.html

Which really confuses me. Suggestions welcome or I can stare at it more.

### DIFF
--- a/paasta_itests/steps/setup_steps.py
+++ b/paasta_itests/steps/setup_steps.py
@@ -84,7 +84,7 @@ def setup_marathon_client():
 
 def setup_chronos_client():
     connection_string = get_service_connection_string('chronos')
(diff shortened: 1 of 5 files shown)


jolynch:
Pretty sure that None is the right version for chronos. When I put v1 I remember everything broke but None fixed it.

solarkennedy:
@jolynch you are right, I didn't read the chronos-python code closely enough.

solarkennedy:
Ok I'm not catching that exception and printing a friendly error message instead. This prevents 2 page long stack traces and allows paasta status to continue on in a degraded mode (no mesos task data).

PR #1388, opened 2017-07-20
use dynamodb persister for paasta remote-run

Rob-Johnson:


Rob-Johnson:
@keymone I need to go away and make the plugin interface understand persistence plugins as well as executors. Until then, I guess I can just import straight from the lib, rather than using the plugin interface.

requirements.txt
```diff
@@ -20,7 +20,7 @@ git+git://github.com/sensu/sensu-plugin-python@bf00ac33
 # so we pull from git
 git+git://github.com/thefactory/marathon-python@c1e0c341#egg=marathon
 # While task_processing is in rapid development mode, we install it from git
-git+git://github.com/Yelp/task_processing@95cc2f8b#egg=task_processing[mesos_executor]
+git+git://github.com/Yelp/task_processing#egg=task_processing[mesos_executor]
```
keymone: i'm hesitant to dropping specific git sha. shipping features unrelated to taskproc will always include latest changes on taskproc - that will trip up people.

paasta_tools/paasta_remote_run.py
```diff
@@ -220,8 +224,34 @@ def build_executor_stack(
...
     )
+
+    credentials_file = taskproc_config.get('boto_credential_file')
+    if credentials_file:
+        with open(credentials_file) as f:
+            credentials = json.loads(f.read())
+    else:
+        raise ValueError("Required aws credentials")
+
+    region = taskproc_config.get('aws_region')
```
keymone: i thought we chnaged that to dynamodb_region or something?

Rob-Johnson: nope, the last thing written out by puppet is aws region

requirements.txt
Rob-Johnson: ack.

paasta_tools/paasta_remote_run.py
keymone: ok, ship then 🚢

PR #1389, opened 2017-07-20
catch noslaveavailable on bad whitelist

jglukasik:
If the user gives a whitelist that PaaSTA can not fulfill, catch the NoSlavesAvailableException and return an empty list for the constraints, and then send an alert to the service author.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -25,6 +25,7 @@ import json
 import logging
 import os
 from math import ceil
+from sys import exit
 
 import requests
 import service_configuration_lib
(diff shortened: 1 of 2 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -368,6 +370,28 @@ def format_marathon_app_dict(self):
...
         net = get_mesos_network_for_net(self.get_net())
 
+        try:
+            constraints = self.get_calculated_constraints(
+                system_paasta_config=system_paasta_config,
+                service_namespace_config=service_namespace_config
+            )
+        except NoSlavesAvailableError as e:
+            cluster = system_paasta_config.get_cluster()
+            error_msg = (
```
solarkennedy: I appreciate the sentiment, but alerts are hard to do really well. Specifically this alert will never "resolve".

How about just `_log` instead?

paasta_tools/marathon_tools.py
```diff
@@ -368,6 +370,28 @@ def format_marathon_app_dict(self):
...
+            )
+            monitoring_tools.send_event(
+                service=self.service,
+                check_name="PaaSTA could not meet your specified constraints!",
+                overrides={},
+                status=pysensu_yelp.Status.CRITICAL,
+                output=error_msg,
+                soa_dir=self.soa_dir,
+            )
+            constraints = []
```
solarkennedy: This is too dangerous. If we can't calculate the constraints we need to bail. But this may be too deep to bail, I think you'll have to catch this further up and return a non-0 code.

The most important part here is that we don't crash deployd/smj entirely.

solarkennedy:
Can you rebase?

jglukasik:
Closing this in favor of https://github.com/Yelp/paasta/pull/1399

PR #1391, opened 2017-07-21
Hooks use python 3 defaults

asottile:


asottile:
Strange, these both failed in the same way -- any hints as to why this is failing?

asottile:
figured it out, there's an itest container that's still running python2.7

asottile:
I can't reproduce the failures in travis-ci locally with docker-compose

solarkennedy:
I've restarted the paasta_itests because it looks like it failed on apt-get dns stuff.

I'm ok with example_cluster tests being broken temporarily, in fact do you think your changes should be used instead of the ones in #1393?

asottile:
I'm still not entirely sure what's up with this branch -- I plan to debug it a bit more when I have some time (it'll need a rebase at this point now)

solarkennedy:
Take your time, there isn't a rush of course and we really appreciate your help.

EvanKrall:
Which container is running python2.7?

asottile:
I've been trying to tackle the last two here: https://github.com/Yelp/paasta/pull/1435

But I'm hitting a bit of a brick wall -- if you'd like to take over the branch and/or complete/debug it that would be awesome (since I'm mostly doing a bit here and there for fun when I get the chance).

solarkennedy:
@asottile ok to merge?

asottile:
@solarkennedy indeedy!  Thanks for the patience on this one :)

solarkennedy:
We really appreciate your efforts in this!

solarkennedy:
If I can get my proposal to name TVs after past employees, I will be sure to get you immortalized in uzbl config! :) (if that is OK)

asottile:
👍

PR #1392, opened 2017-07-22
Only get slave data from the paasta api on verbose mode

solarkennedy:
The paasta api needs to be "cheap" for wait for deployment.
We can't afford fetching state.json on the non-verbose mode of this, especially with the heavy load of wait-for-deployment polling constantly.

### DIFF
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -45,7 +45,7 @@ def adhoc_instance_status(instance_status, service, instance, verbose):
     return cstatus
 
 
-def marathon_job_status(mstatus, client, job_config):
+def marathon_job_status(mstatus, client, job_config, verbose):
     try:
         app_id = job_config.format_marathon_app_dict()['id']
     except NoDockerImageError:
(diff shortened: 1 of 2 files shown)


tests/api/test_instance.py
```diff
@@ -107,7 +107,7 @@ def test_instances_status_adhoc(
 
 @mock.patch('paasta_tools.api.views.instance.get_running_tasks_from_frameworks', autospec=True)
 @mock.patch('paasta_tools.api.views.instance.marathon_tools.is_app_id_running', autospec=True)
-def test_marathon_job_status(
+def test_marathon_job_status_berbose(
```
EvanKrall: s/berbose/verbose/

paasta_tools/api/views/instance.py
```diff
@@ -54,7 +54,8 @@ def marathon_job_status(mstatus, client, job_config):
         return
 
     mstatus['app_id'] = app_id
-    mstatus['slaves'] = list({task.slave['hostname'] for task in get_running_tasks_from_frameworks(app_id)})
+    if verbose is True:
```
nhandler: The `is True` part is not needed

solarkennedy: maybe, but my least favorite bug is `if "false": print ("hi")`

PR #1393, opened 2017-07-24
Fix example_cluster

mattmb:
The pip step was hanging for me. Seemed to be a setuptools bug of some
sort. This should upgrade and pin setuptools in the container.

### DIFF
--- a/yelp_package/dockerfiles/mesos-paasta/Dockerfile
+++ b/yelp_package/dockerfiles/mesos-paasta/Dockerfile
@@ -14,6 +14,7 @@ ADD ./requirements-dev.txt /paasta/requirements-dev.txt
 ADD ./requirements.txt /paasta/requirements.txt
 RUN pip install --upgrade pip==9.0.1
 RUN pip install --upgrade virtualenv==15.1.0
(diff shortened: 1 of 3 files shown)


solarkennedy:
Can you look at the example_cluster changes in #1391 to see if they are relevant?

asottile:
The only useful part of this patch (imo) is the `-r` (which was incorrect before and did nothing).  The hang isn't related to setuptools version but the `gevent` build itself (which is flakey).

mattmb:
I didn't do much to debug to be honest, this was just the first thing that I tried and it "worked". Just a coincidence I guess @asottile? Is the gevent flakiness "random"?

asottile:
Yeah it seemed random to me.  

Honestly a better fix would be factoring out gevent, it is unlikely to perform better than `multiprocessing` and is currently only used for two things:
- paasta api (could easily be replaced with a more-robust uwsgi solution)
- autoscaling lib stats retrieval (should really be using `multiprocessing.Pool`)

Plus, removing it will improve each virtualenv setup by nearly a minute

PR #1395, opened 2017-07-24
Fix resource api

matthewbentley:
The resource api is broken because python doesn't sort frozensets the way I expected.  This fixes it by sorting properly.


solarkennedy:
Why do you need a sorted datastructure to do what you need to do?

paasta_tools/metrics/metastatus_lib.py
```diff
@@ -322,6 +323,30 @@ def key_func(slave):
     return key_func
 
 
+@functools.total_ordering
+class MultiKey():
+    def __init__(self, attributes):
+        self.attributes = attributes
```
asottile: if you're writing a hashable datastructure it should be immutable and not dependent on order.  As written, `attributes` has no guarantee of immutability (and is a public mutable member), also `__str__` (and subsequently `__hash__`) aren't resilient to reordering (the mapping type passed in could yield different results for `.items()` dependent on payload).  As such, if inserted into a mapping type it may either:

1. not be retrievable
2. be duplicated by two "keys" that contain the same data but map to two things

matthewbentley:
The group-by functionality of paasta status starts with a list of slave-info dictionaries, sorts them by some attribute (ie pool), and returns a dict of attribute:[slave-info] ([here](https://github.com/Yelp/paasta/blob/8fda753c0ae7d2b9f4d0fab13a979836bdc6b28a/paasta_tools/metrics/metastatus_lib.py#L362); `key_func` just gets the attribute to be sorted by).

To group by multiple attributes, I need a data structure that retains the attribute names and values (so they can be output later), can be sorted properly, and used as the key in a dict (the current `frozenset` doesn't seem to sort properly).  As @asottile pointed out, what I wrote doesn't do a great job of that, so I'll see if I can find something else that does what I want.  One possibility is rather than using a key_func, write a comparison function that does what I need.

matthewbentley:
I found a (probably) better solution.  Sets don't sort, so this uses a different function to sort than to group.  The actual order of the sort doesn't matter, only the fact that after sorting slaves with equivalent attributes will be adjacent in the list.

PR #1396, opened 2017-07-24
Use python3.6 in api dockerfile

asottile:
cherry-pick from #1391 to try and isolate some changes


asottile:
ok so it is something about this patch that's causing my woes!

asottile:
I'll figure this out in my fork

PR #1397, opened 2017-07-24
Improve dockerfiles

asottile:


asottile:
The itest flaked, but here's a passing run of the same code: https://travis-ci.org/asottile/paasta/jobs/257072046

PR #1399, opened 2017-07-25
Handle NoSlavesException from list_marathon_service_instances

jglukasik:
Closing https://github.com/Yelp/paasta/pull/1389 in favor of this new approach.

Turns out this was a much simpler fix than I realized, and just requires adding NoSlavesAvailableError as an except in a try block in `get_desired_marathon_configs`. I added a test in `test_list_marathon_service_instance`, it fails without the NoSlaves catch, passes with it.

I started this by digging a lot into setup_marathon_job. I ended up writing a test to ensure `deploy_marathon_service` handled the NoSlavesAvailableError correctly. I was surprised to find the test pass, before I ever added code to `setup_marathon_job`. It was then that I took a closer look at the stack trace, and realized that `list_marathon_service_instances` was the real culprit. I think I'll leave the test to check against regressions though, if no one objects.

### DIFF
--- a/paasta_tools/list_marathon_service_instances.py
+++ b/paasta_tools/list_marathon_service_instances.py
@@ -40,6 +40,7 @@ from paasta_tools.marathon_tools import load_marathon_config
 from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.mesos.exceptions import NoSlavesAvailableError
 from paasta_tools.mesos_maintenance import get_draining_hosts
+from paasta_tools.utils import _log
 from paasta_tools.utils import compose_job_id
 from paasta_tools.utils import get_services_for_cluster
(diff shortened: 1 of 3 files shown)


paasta_tools/list_marathon_service_instances.py
```diff
@@ -89,7 +90,7 @@ def get_desired_marathon_configs(soa_dir):
                 soa_dir=soa_dir,
             ).format_marathon_app_dict()
             marathon_configs[marathon_config['id'].lstrip('/')] = marathon_config
-        except (NoDeploymentsAvailable, NoDockerImageError):
+        except (NoDeploymentsAvailable, NoDockerImageError, NoSlavesAvailableError):
```
solarkennedy: It is ok for NoDeployments and NoDocker to be silent.

This should not be silent and needs to `_log`

PR #1401, opened 2017-07-25
Make paasta_api docker image use python3.6

asottile:


asottile:
paasta_itests flaked, an identical run here passed: https://travis-ci.org/asottile/paasta/builds/257166364

PR #1402, opened 2017-07-25
Bug fixes to firewall_update related to py3 upgrade

bchess:
inotify filenames were bytestrings so the `suffix == '.json'` test was failing :(


bchess:
It's probably worth doing an itest for this sometime..

PR #1403, opened 2017-07-25
Bump task-proc and fix a unit test.

sagar8192:
Bumped task-proc to latest.

test_run_docker_container_with_user_specified_port test was failing on master because check_if_port_free is not mocked out. So, if port 1234 is taken on the box where these tests are running, then this unit test fails. Mocked it out to fix the test.

### DIFF
--- a/paasta_tools/paasta_remote_run.py
+++ b/paasta_tools/paasta_remote_run.py
@@ -47,6 +47,8 @@ from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import PaastaNotConfiguredError
 from paasta_tools.utils import SystemPaastaConfig
 from paasta_tools.utils import validate_service_instance
+from task_processing.runners.sync import Sync
+from task_processing.task_processor import TaskProcessor
 
 MESOS_TASK_SPACER = '.'
 
--- a/paasta_tools/paasta_remote_run.py
+++ b/paasta_tools/paasta_remote_run.py
@@ -25,8 +25,11 @@ import string
 import sys
 from datetime import datetime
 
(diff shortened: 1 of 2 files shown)


sagar8192:
reorder-imports pre-commit hook is not failing on my devbox. Anyone knows why it thinks task-processing stuff should be part of paasta import block?

asottile:
@sagar8192 on your box you must have `./task_processing` inside your paasta checkout?

sagar8192:
@asottile++ I had task_processing inside my paasta repo.

requirements.txt
```diff
@@ -20,7 +20,7 @@ git+git://github.com/sensu/sensu-plugin-python@bf00ac33
 # so we pull from git
 git+git://github.com/thefactory/marathon-python@c1e0c341#egg=marathon
 # While task_processing is in rapid development mode, we install it from git
-git+git://github.com/Yelp/task_processing@95cc2f8b#egg=task_processing[mesos_executor]
+git+git://github.com/Yelp/task_processing@29e4361e#egg=task_processing[mesos_executor]
```
Rob-Johnson: is there any reason to not have this set to HEAD all the time?

asottile: The same reason *anything* is pinned in requirements.txt -- so your build isn't broken all the time (reproducibility) -- this PR is a great example of *exactly why* this is a problem (the modified test would have been broken and all the paasta builds would have been broken -- likely taking **hours** of someone's time as they waited for travis-ci to rerun).

huadongliu:
shipit

requirements.txt
sagar8192: @Rob-Johnson Yeah, if we commit a bad change to task-proc, then it could block a paasta release.

PR #1405, opened 2017-07-26
Remove python2-specific requirements

asottile:
yay stdlib


asottile:
paasta_itests flaked, but identical code passed here: https://travis-ci.org/asottile/paasta/builds/257662163

PR #1409, opened 2017-07-28
Allow setting the shm-size

matthewbentley:
Mostly to run chromium/selenium in the container


Rob-Johnson:
could do with updating docs, but we can do that after hackathon

solarkennedy:
I think we need just an 'extra_docker_params' to allow room for weird 'only used once' extra args that some services need.

matthewbentley:
Or `allowed_docker_params`.  I'll look into it next week if I get some time.

PR #1410, opened 2017-07-28
Remove unused dockerfiles/itest/itest

asottile:


asottile:
this build has already passed here: https://www.travis-ci.org/asottile/paasta/builds/258642586

PR #1414, opened 2017-08-01
Fail wait-for-deployment if a cluster has no configured API endpoints

oktopuz:
internat ticket: PAASTA-10702

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -48,6 +48,7 @@ from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import format_tag
 from paasta_tools.utils import get_git_url
 from paasta_tools.utils import get_paasta_tag_from_deploy_group
+from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import paasta_print
 from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import TimeoutError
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -264,12 +276,10 @@ def paasta_mark_for_deployment(args):
...
                     )
             else:
-                paasta_print("Waiting for deployment aborted. PaaSTA will continue to try to deploy this code.")
-                paasta_print("If you wish to see the status, run:")
-                paasta_print()
-                paasta_print("    paasta status -s %s -v" % service)
-                paasta_print()
+                report_waiting_aborted(service, args.deploy_group)
             ret = 1
+        except NoSuchCluster:
```
solarkennedy: Can you print "why" it aborted? or it is already covered when the exception is raised?
Does it print which cluster it not defined?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -527,7 +537,14 @@ def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
 
     total_instances = 0
     clusters_data = []
+    api_endpoints = load_system_paasta_config().get_api_endpoints()
     for cluster in cluster_map:
+        if cluster not in api_endpoints:
+            paasta_print(PaastaColors.red(
+                'Cluster %s is NOT in paasta-api endpoints config.' %
```
solarkennedy: Ah, aparently yes

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -527,7 +537,14 @@ def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
 
     total_instances = 0
     clusters_data = []
+    api_endpoints = load_system_paasta_config().get_api_endpoints()
```
solarkennedy: I'm a bit confused at why we have to load the api_endpoints thing so early here.

Is there a function deeper down that actually looks up the api_endpoints and would be a more natural place to raise this exception?

oktopuz: In ~10 lines from this one there is a function which creates a number of threads to send async queries to clusters.

If I raise the exception in a child thread, it will only terminate that thread, but not the main process.
Therefore I will need to send a signal to other threads to terminate them.
And while such signal is being sent and processed all un-terminated threads still continue to print their output to stdout.

I'm certain that calling load_system_paasta_config().get_api_endpoints() here is the best way to fail fast, as PAASTA-10702 implies.

PR #1416, opened 2017-08-01
Use xenial for more dockerfiles

asottile:
I'll eventually follow up with more of these, this makes it easier to get all of the compose containers running python3.6


asottile:
Note that if this flakes, it has already passed here: https://travis-ci.org/asottile/paasta/builds/259912432

PR #1417, opened 2017-08-02
Always whitelist Google Public DNS

chriskuehl:
PAASTA-12338

### DIFF
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -30,6 +30,12 @@ PRIVATE_IP_RANGES = (
     '192.168.0.0/255.255.0.0',
     '169.254.0.0/255.255.0.0',
 )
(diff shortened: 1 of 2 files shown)


bobtfish:
-- to this change.

I don't think that allowing our containers to talk to google's public DNS is a good idea *at all*, as it puts us at the mercy of palo alto firewalls, means yelpcorp addresses won't resolve and other things that are likely to compromise reliability.

We should work out why we're configuring things with google's DNS servers (systemd)? and fix it!

bchess:
@bobtfish I'm not against specifying our own servers all the time. The 8.8.8.8 defaults come from the Docker project itself.

bobtfish:
Using 8.8.8.8 just feels like we're playing russian roulette, with a 3rd party we're not even paying loading the gun for us...

PR #1418, opened 2017-08-02
drop retrying executor from the stack

keymone:
tron isn't retrying so this would be unexpected behaviour, possibly undesirable. will re-add it carefully later.

### DIFF
--- a/paasta_tools/paasta_remote_run.py
+++ b/paasta_tools/paasta_remote_run.py
@@ -26,6 +26,9 @@ import sys
 from datetime import datetime
 
 from boto3.session import Session
+from task_processing.plugins.persistence.dynamodb_persistence import DynamoDBPersister
+from task_processing.runners.sync import Sync
+from task_processing.task_processor import TaskProcessor
 
 from paasta_tools import mesos_tools
 from paasta_tools.cli.cmds.remote_run import add_common_args_to_parser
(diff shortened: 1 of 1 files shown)


Rob-Johnson:
revert the changes to the import ordering that is causing pre-commit to complain and ship

asottile:
@Rob-Johnson @keymone I'm curious why it seems so many are developing with a checkout of ./task_processing *inside* the paasta checkout -- this is usually not the best idea as it doesn't represent how pip will install `task_processing` and should instead be done with `pip install -e ../task_processing` (or `pip install ../task_processing`) usually.

keymone:
@asottile i'm doing it because i didn't know any better, thanks for the hint! do i need to re-run `pip install ../task_processing` after i change it or does it just symlinks things?

asottile:
If you're using `-e` you don't need to re-run, otherwise yes (`-e` is an editable install)

sagar8192:
I think we should keep the retrying executor. I saw a couple of instances where remote-run got TASK_LOST, and it terminated. TASK_LOST is due to infra issues. Maybe the slave it was running on went away, mesos-slave process and/or docker died etc. I think we should retry at least for TASK_LOST. We can add a whitelist of status updates to retrying_exeutor. But, I agree that we should not retry for TASK_FAILED. This could be undesired for batches that are not idempotent. TASK_KILLED is a bit tricky.

keymone:
@sagar8192 it's a valid concern, but what happens to tron jobs when there's tron failure? i don't think they get restarted automatically. until we're sure all jobs can handle restarts and are idempotent we can't really force it on our users.

PR #1424, opened 2017-08-02
Add process name to the oom log

oktopuz:
This change should add name of the process invoked OOM killer to the log.

### DIFF
--- a/paasta_tools/oom_logger.py
+++ b/paasta_tools/oom_logger.py
@@ -24,7 +24,7 @@ destination paasta_oom_logger {
 };
 
 filter f_cgroup_oom {
-  match("killed as a result of limit of");
+  match(" killed as a result of limit of ") or match(" invoked oom-killer: ");
 };
 
 log {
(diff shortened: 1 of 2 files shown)


paasta_tools/oom_logger.py
```diff
@@ -75,35 +91,32 @@ def get_container_env_as_dict(docker_inspect):
...
+                log_line.process_name,
             ))
     logger.log_line('tmp_paasta_oom_events', line)
 
 
 def log_to_paasta(log_line):
     """Add the event to the standard PaaSTA logging backend."""
-    line = ('A process in the container %s on %s killed by OOM.'
-            % (log_line.container_id, log_line.hostname))
+    line = ('%s invoked oom-killer on %s (container_id = %s).'
```
solarkennedy: How about

    oom-killler killed a %s process on %s

paasta_tools/oom_logger.py
```diff
@@ -75,35 +91,32 @@ def get_container_env_as_dict(docker_inspect):
...
+            % (
+                log_line.process_name or 'A process',
```
solarkennedy: oh, I guess would need to change to "" or something, meh

PR #1425, opened 2017-08-02
Inject the PAASTA_DEPLOY_GROUP env variable

jglukasik:


jglukasik:
This PR has grown past its original scope.

There have been requests to add the PAASTA_TEAM variable into the env as well, and I thought I'd try and get that done in this same pass. Unfortunately, there is a circular dependency issue where we need a function from `monitoring_tools.py` in `utils.py`, but because the former is importing the latter, we end up in a dependency cycle. 

This PR has nearly finished the refactoring necessary to do this, but the integration tests are still failing. I hope to get back to this and fix it, but for now I'm going to close this PR and open another one with just the DEPLOY_GROUP stuff, so we can keep moving ahead on an internal project that needs that var.

PR #1426, opened 2017-08-03
Add type annotations and a mypy tox env

EvanKrall:
Even though I've only made a small dent into making mypy happy (452 more errors to go), if travis passes I'd like to get this merged so I don't have to fight merge conflicts forever.

This does actually fix a few bugs!

- graceful_app_drain was not calling some functions correctly
- check_synapse_replication wasn't passing a status code to fail() in one place
- kill_orphaned_docker_containers was passing a tuple of strings to argparse accidentally (we must have accidentally slipped a comma into our multi-line string literal.)
- InstanceConfig.get_deploy_whitelist was pretty wrong -- A deploy whitelist is of the form `["region", ["uswest1", "uswest2]]"`, or `[location_type, [location_val, ...]]`. It can only have two entries. We were concatenating the system config's deploy whitelist and the instance config's deploy whitelist, which only works if one or both of them is empty.


EvanKrall:
In particular, I'd like some eyes on the whitelist changes. Everything else should be more or less noop.

asottile:
You're probably butting heads against the same stuff I was in #1391 -- notably that there are *still* some itest containers running python2.7.  I've been slowly trying to move them to xenial (xenial as there's fewer vitualenv hax necessary) and python3.6

EvanKrall:
I think this is ready to merge. The example_cluster itest is failing, but it appears to be failing the same way master is.

solarkennedy:
Has a type assertion in the tests.

The change is large of course, but I'm in favor. Get the tests to pass, pin, and release.

tests/test_utils.py
```diff
@@ -1516,7 +1518,7 @@ def test_is_under_replicated_critical():
 
 
 def test_deploy_blacklist_to_constraints():
-    fake_deploy_blacklist = [["region", "useast1-prod"], ["habitat", "fake_habitat"]]
+    fake_deploy_blacklist = [("region", "useast1-prod"), ("habitat", "fake_habitat")]
```
solarkennedy: Are peoples yamls going to continue to work if they are using [["foo", "bar"]] ? If not, then why did these need to be changed to an array of tuples?

paasta_tools/marathon_tools.py
```diff
@@ -307,13 +308,17 @@ def get_routing_constraints(self, service_namespace_config, system_paasta_config
...
-            blacklist=self.get_deploy_blacklist(
-                system_deploy_blacklist=system_paasta_config.get_deploy_blacklist(),
-            ),
-            whitelist=self.get_deploy_whitelist(
-                system_deploy_whitelist=system_paasta_config.get_deploy_whitelist(),
-            ),
+            blacklist=self.get_deploy_blacklist(),
+            whitelist=self.get_deploy_whitelist(),
+        )
+        # A slave must be allowed by both the instance config's blacklist/whitelist and the system configs' blacklist/
```
solarkennedy: strange, was this a bug just now being catched?

tests/test_utils.py
EvanKrall: The function under test still works with 2-element arrays; mypy does nothing at runtime. This code needed to be changed because test_utils.py is one of the files we're checking with mypy.

paasta_tools/marathon_tools.py
EvanKrall: Yup. The whitelist code previously was taking the system whitelist and the service's whitelist and concatenating them together, then unpacking them in `slave_passes_whitelist` with  `(location_type, locations) = whitelist`. This would fail if we ever set both the system whitelist and the service whitelist. Fortunately, we don't set the system whitelist.

PR #1428, opened 2017-08-03
Make paasta without args compat with py3

solarkennedy:


asottile:
The *usual* fix for this is `subparsers.required = True` (at least in a 2+3 compatible codebase)

solarkennedy:
Second attempt with required=true

paasta_tools/cli/cli.py
```diff
@@ -84,6 +84,10 @@ def get_argparser():
     )
 
     subparsers = parser.add_subparsers(help="[-h, --help] for subcommand help")
+    subparsers.required = True
+    # https://stackoverflow.com/questions/18282403/argparse-with-required-subcommands/18283730#18283730
+    subparsers.dest = 'command'
```
asottile: This can be a keyword to `add_subparsers` (but `.required` cannot for some reason, maybe I'll make a PR to stdlib to enable this).  Here's an [example from pre-commit](https://github.com/pre-commit/pre-commit/blob/72efbb3950ef5943156c0171737b3d7edcfa00e7/pre_commit/main.py#L60) (note that pre-commit avoids needing `.required = true` due to some other hax that I wouldn't suggest (it implements a "default" command))

asottile: True to my word, here's the cpython PR: https://github.com/python/cpython/pull/3027

solarkennedy: Till we upgrade to a version of python with that PR, then I should ship what I have as is, correct?

asottile: You can move `dest='command'` as a kwarg to `add_subparsers`, but you won't be able to move `.required = ...` until then.

PR #1429, opened 2017-08-03
Use a version of chronos-python that doesn't warn on non v3

solarkennedy:


requirements.txt
```diff
@@ -5,16 +5,14 @@ blessings==1.6
...
 bravado==8.4.0
 choice==0.1
-chronos-python==1.2.0
 cookiecutter==1.4.0
 croniter==0.3.14
 docker-py==1.2.3
 dulwich==0.17.3
 ephemeral-port-reserve==1.0.1
 gevent==1.1.1
+git+git://github.com/asher/chronos-python@e05afa92bf#egg=chronos
```
asottile: fwiw, usually the `#egg=...` is not necessary in vcs urls with modern pip

Rob-Johnson: why aren't we doing a new release of chronos-python? we do have pip access

solarkennedy: I didn't think it was really necessary to make a pip release for every commit.

PR #1431, opened 2017-08-04
Convert marathon docker image to xenial

asottile:


asottile:
if it flakes, it passed here already: https://travis-ci.org/asottile/paasta/builds/260858943

PR #1433, opened 2017-08-04
Support multiple puppet service ports

jolynch:
configure_nerve just trusts the port, so I think this is all we need to support puppet services registering multiple service ports at once.

cc @matthewmcarlson


solarkennedy:
If you want, I'll take this and run with it and migrate all "port" files into the smartstack.yaml and deprecated port files so we only have the 1 way to do things?

jolynch:
@solarkennedy sure that sounds pretty reasonable, I think the old port file was just backwards compatibility for intlbs right?

solarkennedy:
No, we still use the contents of that file to inform puppet which port to have nerve advertise.

jolynch:
Yes, naturally. What I meant was that the only reason port was outside smartstack.yaml in the first place was to be backwards compatible with existing classic soa services setup through the intlbs afaik.

PR #1434, opened 2017-08-05
Make deployd skip deploys on broken instances

solarkennedy:
1. I'm deciding *not* to list service.instances that are not ready for some reason (broken deployments, image, etc)
2. I'm caching NoDeployments, as it was crashing deployd and causing it to start over.

### DIFF
--- a/tests/deployd/test_common.py
+++ b/tests/deployd/test_common.py
@@ -14,6 +14,7 @@ from paasta_tools.deployd.common import rate_limit_instances
 from paasta_tools.deployd.common import ServiceInstance
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import InvalidJobNameError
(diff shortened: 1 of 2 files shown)


tests/deployd/test_common.py
```diff
@@ -140,7 +140,7 @@ def test_get_service_instances_needing_update():
         ]
         mock_load_marathon_service_config.side_effect = mock_configs
         ret = get_service_instances_needing_update(mock.Mock(), mock_service_instances, 'westeros-prod')
-        assert ret == [('universe', 'c137'), ('universe', 'c138')]
+        assert ret == [('universe', 'c138')]
 
         mock_configs = [
             mock.Mock(format_marathon_app_dict=mock.Mock(side_effect=InvalidJobNameError)),
```
mattmb: Can you dupe this stanza with `NoDeploymentsAvailable` as the side_effect please. Just so we cover all the exceptions in the test.

PR #1435, opened 2017-08-05
Convert the playground and mesos dockerfiles to py36

asottile:
I hit a wall converting these to xenial so this will have to do :/


mattmb:
Thanks for looking at this @asottile!

I'd like to have a quick play before we merge, also have you spotted the travis failure (we made the example cluster non-blocking because it is a bit flakey)?

asottile:
hmmm could have sworn I got this to pass on my own travis-ci, I'll look into the failure there (though debugging these really is awful)

asottile:
This still breaks `example_cluster` @solarkennedy

PR #1438, opened 2017-08-08
check_chronos_jobs to take job runtime into account

oktopuz:
internal ticket: PAASTA-12181

### DIFF
--- a/paasta_tools/check_chronos_jobs.py
+++ b/paasta_tools/check_chronos_jobs.py
@@ -7,6 +7,7 @@ a CRITICAL event to sensu.
 - -d <SOA_DIR>, --soa-dir <SOA_DIR>: Specify a SOA config dir to read from
 """
 import argparse
+import logging
 import sys
 from datetime import datetime
 from datetime import timedelta
@@ -22,7 +23,9 @@ from paasta_tools import utils
 from paasta_tools.chronos_tools import compose_check_name_for_service_instance
 from paasta_tools.chronos_tools import DEFAULT_SOA_DIR
 from paasta_tools.chronos_tools import load_chronos_job_config
-from paasta_tools.utils import paasta_print
+
+
+log = logging.getLogger(__name__)
 
 
 def parse_args():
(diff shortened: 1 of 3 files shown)


paasta_tools/check_chronos_jobs.py
```diff
@@ -292,12 +298,19 @@ def main():
             except utils.NoDeploymentsAvailable:
                 paasta_print(utils.PaastaColors.cyan("Skipping %s because no deployments are available" % service))
                 continue
+            job_expected_runtime = 0
+            if chronos_job:
+                try:
+                    job_expected_runtime = int(client.job_stat(chronos_job['name'])['histogram']['99thPercentile'])
```
oktopuz: I just realized that this request is redundant when last_run_time + interval is more than datetime.now (i.e. the job hasn't started yet).
However this implementation is faster than the current one anyway.
I'd rather ship it as it is, and then refactor this part in a new PR.

paasta_tools/check_chronos_jobs.py
```diff
@@ -264,6 +277,19 @@ def sensu_message_status_for_jobs(chronos_job_config, service, instance, cluster
     return output, sensu_status
 
 
+def job_expected_runtime_callback(client, chronos_job):
+    """Returns a function that is supposed to be called from job_is_stuck() to get the job expected runtime."""
+    def f():
+        job_expected_runtime = 0
+        if chronos_job:
+            try:
+                job_expected_runtime = int(client.job_stat(chronos_job['name'])['histogram']['99thPercentile'])
```
solarkennedy: just `return int(...`

paasta_tools/check_chronos_jobs.py
```diff
@@ -264,6 +277,19 @@ def sensu_message_status_for_jobs(chronos_job_config, service, instance, cluster
...
+            except KeyError:
+                paasta_print(utils.PaastaColors.yellow("Can't get 99thPercentile for %s." % chronos_job['name']))
```
solarkennedy: just log.info() and then return the job interval

paasta_tools/check_chronos_jobs.py
```diff
@@ -264,6 +277,19 @@ def sensu_message_status_for_jobs(chronos_job_config, service, instance, cluster
     return output, sensu_status
 
 
+def job_expected_runtime_callback(client, chronos_job):
```
solarkennedy: Why does this have to be a callback? Why can this be a "normal" function?

oktopuz: We don't always need to know 99thPercentile of the job.
But the only way to know whether to make this request or not, is to verify if last_run_time + interval is bigger than datetime.now in job_is_stuck().

If I make this a "normal" function that is called from job_is_stuck() and returns int, then I'll have to pass 'client' and 'chronos_job' as arguments to sensu_message_status_for_jobs(), then to job_is_stuck().

With the current functional approach I only pass one argument.
In my opinion passing 'expected_runtime_callback' is more straightforward way than passing  'client' and 'chronos_job' and expect future developers to understand what are they and how/where do we use them.

paasta_tools/check_chronos_jobs.py
oktopuz: This script doesn't use logging at all.
Do you want to replace all paaasta_prints with log.info()?

solarkennedy: yea, good old normal logging for this one. paasta_prints are usually for user-facing output

oktopuz:
@solarkennedy, I removed some redundant arguments.
This is ready for review.

paasta_tools/check_chronos_jobs.py
```diff
@@ -170,32 +185,44 @@ def message_for_status(status, service, instance, cluster):
...
-    last_run_datatime = isodate.parse_datetime(last_run_iso_time)
-    return last_run_datatime + timedelta(seconds=interval_in_seconds) < datetime.now(pytz.utc)
+    dt_next_run = isodate.parse_datetime(last_run_iso_time) + timedelta(seconds=interval_in_seconds)
+    dt_now_utc = datetime.now(pytz.utc)
+    if dt_next_run >= dt_now_utc:
+        return False
+    try:
+        expected_runtime = int(client.job_stat(job_name)['histogram']['99thPercentile'])
+    except KeyError:
+        log.warn("Can't get 99thPercentile for %s." % job_name)
```
solarkennedy: This can happen a lot, no need to warn. I think debug is fine.

Also log "Assuming a runtime of $interval_in_seconds seconds")

PR #1439, opened 2017-08-08
Add b prefix to fix type errors

ddelnano:
When running `paasta status -s service-name -C chronos` I see the following stack trace

```
[36mFetching the last 100 lines and applying filters...[0m
  File "/usr/bin/paasta", line 11, in <module>
    load_entry_point('paasta-tools==0.66.19', 'console_scripts', 'paasta')()
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cli.py", line 122, in main
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cmds/logs.py", line 1106, in paasta_logs
    return pick_default_log_mode(args, log_reader, service, levels, components, clusters, instances)
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cmds/logs.py", line 1033, in pick_default_log_mode
    raw_mode=args.raw_mode,
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cmds/logs.py", line 771, in print_last_n_logs
    self.run_code_over_scribe_envs(clusters=clusters, components=components, callback=callback)
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cmds/logs.py", line 566, in run_code_over_scribe_envs
    callback(component, stream_info, scribe_env, cluster=cluster)
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cmds/logs.py", line 768, in callback
    parser_fn=stream_info.parse_fn,
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cmds/logs.py", line 786, in filter_and_aggregate_scribe_logs
    line = parser_fn(line, clusters, service)
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cmds/logs.py", line 304, in parse_chronos_log_line
    utc_timestamp = extract_utc_timestamp_from_log_line(line)
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cmds/logs.py", line 276, in extract_utc_timestamp_from_log_line
    tokens = re.match(iso_re, line)
  File "/opt/venvs/paasta-tools/lib/python3.6/re.py", line 172, in match
    return _compile(pattern, flags).match(string)
TypeError: cannot use a string pattern on a bytes-like object
```

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -269,17 +269,17 @@ def extract_utc_timestamp_from_log_line(line):
     or None if it could not parse the line
     """
     # Extract ISO 8601 date per http://www.pelagodesign.com/blog/2009/05/20/iso-8601-date-validation-that-doesnt-suck/
-    iso_re = r'^([\+-]?\d{4}(?!\d{2}\b))((-?)((0[1-9]|1[0-2])(\3([12]\d|0[1-9]|3[01]))?|W([0-4]\d|5[0-2])(-?[1-7])?|' \
-        r'(00[1-9]|0[1-9]\d|[12]\d{2}|3([0-5]\d|6[1-6])))([T\s]((([01]\d|2[0-3])((:?)[0-5]\d)?|24\:?00)([\.,]\d+' \
-        r'(?!:))?)?(\17[0-5]\d([\.,]\d+)?)?([zZ]|([\+-])([01]\d|2[0-3]):?([0-5]\d)?)?)?)? '
-
+    iso_re = br'^([\+-]?\d{4}(?!\d{2}\b))((-?)((0[1-9]|1[0-2])(\3([12]\d|0[1-9]|3[01]))?|W([0-4]\d|5[0-2])(-?[1-7])?|' \
+        br'(00[1-9]|0[1-9]\d|[12]\d{2}|3([0-5]\d|6[1-6])))([T\s]((([01]\d|2[0-3])((:?)[0-5]\d)?|24\:?00)([\.,]\d+' \
+        br'(?!:))?)?(\17[0-5]\d([\.,]\d+)?)?([zZ]|([\+-])([01]\d|2[0-3]):?([0-5]\d)?)?)?)? '
+    line = line if type(line) is bytes else line.encode() 
     tokens = re.match(iso_re, line)
 
     if not tokens:
         # Could not parse line
         return None
     timestamp = tokens.group(0).strip()
-    dt = isodate.parse_datetime(timestamp)
+    dt = isodate.parse_datetime(timestamp.decode())
     utc_timestamp = datetime_convert_timezone(dt, dt.tzinfo, dateutil.tz.tzutc())
     return utc_timestamp
 
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -272,7 +272,7 @@ def extract_utc_timestamp_from_log_line(line):
     iso_re = br'^([\+-]?\d{4}(?!\d{2}\b))((-?)((0[1-9]|1[0-2])(\3([12]\d|0[1-9]|3[01]))?|W([0-4]\d|5[0-2])(-?[1-7])?|' \
         br'(00[1-9]|0[1-9]\d|[12]\d{2}|3([0-5]\d|6[1-6])))([T\s]((([01]\d|2[0-3])((:?)[0-5]\d)?|24\:?00)([\.,]\d+' \
         br'(?!:))?)?(\17[0-5]\d([\.,]\d+)?)?([zZ]|([\+-])([01]\d|2[0-3]):?([0-5]\d)?)?)?)? '
-    line = line if type(line) is bytes else line.encode() 
+    line = line if type(line) is bytes else line.encode()
     tokens = re.match(iso_re, line)
 
     if not tokens:
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -73,8 +73,7 @@ DEPLOY_PIPELINE_NON_DEPLOY_STEPS = (
 ANY_CLUSTER = 'N/A'
 ANY_INSTANCE = 'N/A'
 DEFAULT_LOGLEVEL = 'event'
-no_escape = re.compile('\x1B\[[0-9;]*[mK]')
-no_escape_b = re.compile(b'\x1B\[[0-9;]*[mK]')
+no_escape = re.compile(b'\x1B\[[0-9;]*[mK]')
 
 DEFAULT_SYNAPSE_HAPROXY_URL_FORMAT = "http://{host:s}:{port:d}/;csv;norefresh"
 
@@ -842,7 +841,7 @@ def _now():
 
 def remove_ansi_escape_sequences(line):
     """Removes ansi escape sequences from the given line."""
-    return no_escape.sub('', line) if type(line) is str else no_escape_b.sub(b'', line)
+    return no_escape.sub(b'', line)
 
 
 def format_log_line(level, cluster, service, instance, component, line, timestamp=None):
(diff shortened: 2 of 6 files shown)


ddelnano:
@matthewbentley I am very much a python n00b but from seeing similar things on stackoverflow I _think_ this is the fix.

matthewbentley:
`SyntaxError: cannot mix bytes and nonbytes literals` when I tried to run this

```
  File "/usr/bin/paasta", line 11, in <module>
    load_entry_point('paasta-tools==0.66.19', 'console_scripts', 'paasta')()
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cli.py", line 117, in main
    args, parser = parse_args(argv)
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cli.py", line 103, in parse_args
    parser = get_argparser()
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cli.py", line 92, in get_argparser
    add_subparser(command, subparsers)
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cli.py", line 56, in add_subparser
    add_subparser_fn = load_method(module_name, 'add_subparser')
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/utils.py", line 63, in load_method
    module = __import__(module_name, fromlist=[method_name])
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/cli/cmds/logs.py", line 272
    iso_re = br'^([\+-]?\d{4}(?!\d{2}\b))((-?)((0[1-9]|1[0-2])(\3([12]\d|0[1-9]|3[01]))?|W([0-4]\d|5[0-2])(-?[1-7])?|' \
                ^
SyntaxError: cannot mix bytes and nonbytes literals
```

matthewbentley:
```
diff --git a/paasta_tools/cli/cmds/logs.py b/paasta_tools/cli/cmds/logs.py
index 131cf9d..c517c7f 100644
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -783,6 +783,7 @@ class ScribeLogReader(LogReader):
             try:
                 for line in scribe_reader:
                     if parser_fn:
+                        line = line.decode('UTF-8')
                         line = parser_fn(line, clusters, service)
                     if filter_fn:
                         if filter_fn(
```
should fix it, but there may be a better place to fix this

ddelnano:
ah would help if I tested my changes :).  My paasta repo on my dev box is not working at the moment.  But updated to what you suggested.

paasta_tools/cli/cmds/logs.py
```diff
@@ -785,6 +785,7 @@ def filter_and_aggregate_scribe_logs(
                     if parser_fn:
                         line = parser_fn(line, clusters, service)
                     if filter_fn:
+                        line = line.decode('UTF-8')
```
matthewbentley: Probably move this to after line 784.

ddelnano:
@matthewbentley finally able to verify I don't see that TypeError anymore.

I am still not seeing success but I'm thinking its related to my dev setup

```
  File "/usr/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/nail/home/ddelnano/pg/paasta/paasta_tools/cli/cli.py", line 129, in <module>
    main()
  File "/nail/home/ddelnano/pg/paasta/paasta_tools/cli/cli.py", line 122, in main
  File "/nail/home/ddelnano/pg/paasta/paasta_tools/cli/cmds/logs.py", line 1100, in paasta_logs
    log_reader = get_log_reader()
  File "/nail/home/ddelnano/pg/paasta/paasta_tools/cli/cmds/logs.py", line 462, in get_log_reader
    return log_reader_class(**log_reader_config.get('options', {}))
  File "/nail/home/ddelnano/pg/paasta/paasta_tools/cli/cmds/logs.py", line 535, in __init__
    raise Exception("scribereader package must be available to use scribereader log reading backend")
Exception: scribereader package must be available to use scribereader log reading backend
```

ddelnano:
I was able to get this working by `pip install scribereader -i https://pypi.yelpcorp.com/simple/`.  Why is this not in the requirements.txt?

Rob-Johnson:
@ddelnano because scribereader is a Yelpy thing, and we try to keep PaaSTA from depending on yelp-specific tools.

asottile:
The reason for the `TypeError` is bytes objects must be matched with bytes regexes (and text must be matched with text regexes).  In this case, I think *decoding* the incoming stream is probably not a good idea (the stream could contain non-utf8 data) and you should use a bytes regex to match the bytes objects.

`SyntaxError: cannot mix bytes and nonbytes literals` is failing because you have a multiline literal, each *part* of the literal must be prefixed:

```python
b = (
    br'foo'
    br'bar'
)
```

ddelnano:
@asottile thanks for the advice.  I updated the implementation based on your suggestion and also added a test to cover that case.  I also needed to update the `remove_ansi_escape_sequences` method to accept a byte object in order for the `paasta logs -s service-name -C chronos` to work.

ddelnano:
Not sure if these methods should *only* accept bytes objects or not but I updated everything to support both.

paasta_tools/utils.py
```diff
@@ -841,7 +842,7 @@ def _now():
 
 def remove_ansi_escape_sequences(line):
     """Removes ansi escape sequences from the given line."""
-    return no_escape.sub('', line)
+    return no_escape.sub('', line) if type(line) is str else no_escape_b.sub(b'', line)
```
solarkennedy: Can we get away with just the `b` version in this function? (do the tests still pass?)

ddelnano: It breaks ~30 tests.  I'm fine with updating them to all use bytes if that's what we want.  I just didn't know the context in which this is used so I opted for the most simple change until I got more feedback.

asottile: If the actual runtime only presents bytes, the tests are all currently irrelevant (and incorrect) -- this function should only support one sequence type (either `str` **or** `bytes`).

ddelnano: @asottile yea thats mainly why I wasn't sure if we need to keep the string support.  From having such little experience with this codebase I wasn't sure what we were expecting from the "real world".  So I would defer to whatever you guys think.

solarkennedy: I think in this case the ansi code regex should assume bytes and bytes only.

EvanKrall:
Yeah I think the easiest fix is probably the line.decode('utf-8') patch, so all LogReaders produce strings and not bytes.

asottile:
@EvanKrall there's no guarantee that services log properly encoded utf-8 text (heck, they could even be writing purely binary output) -- paasta shouldn't attempt to decode / reencode 3rd party streams at all imo.

EvanKrall:
@asottile fair point, I wasn't thinking about the app_output logs which could contain arbitrary bytes. The other logs are all produced by paasta or Marathon or Chronos, so we can be reasonably confident about their contents.

ddelnano:
I moved this to #1464.  After making `remove_ansi_escape_sequences` only accept bytes I ran into issues with getting the test suite passing and once I "fixed" the issues I still ended up with the bug I was trying to fix.  Many of the functions in the util module expect strings and so it created problems.  Instead I opted for decoding the stream as possible.  I only added this for the `parse_marathon_log_line` and `parse_marathon_log_line` functions since as @asottile mentioned we shouldn't be decoding bytes unless we are confident of the contents and as @EvanKrall said it should be safe to assume we can decode the marathon and chronos logs as utf-8.

PR #1440, opened 2017-08-08
Inject the PAASTA_DEPLOY_GROUP env variable

jglukasik:
This is the same as the start of https://github.com/Yelp/paasta/pull/1425 , but without all the refactoring mess necessary to inject the PAASTA_TEAM var.


tests/test_chronos_tools.py
```diff
@@ -402,6 +402,7 @@ def test_get_env(self):
             {"name": "PAASTA_CLUSTER", "value": "fake_cluster"},
             {"name": "PAASTA_SERVICE", "value": "fake_name"},
             {"name": "PAASTA_INSTANCE", "value": "fake_instance"},
+            {"name": "PAASTA_DEPLOY_GROUP", "value": "fake_cluster.fake_instance"},
```
solarkennedy: just say "fake_deploy_group"

PR #1441, opened 2017-08-08
Support FQDN firewall rules

chriskuehl:
This ended up being relatively straightforward since it's the same as the CIDR rules, but with a dns lookup first. I refactored the CIDR rule code a bit so they could share the generating of the iptables rules.

It's true that there will be some delay in these rules updating (currently 5 minutes, because that's how often the cronjob runs). I think for most use cases, this will be acceptable, but we may need to revisit if we find it's not sufficient. I think the benefit of this functionality outweighs that it's not perfectly correct, as there are some rules you can't otherwise express.

### DIFF
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -7,6 +7,7 @@ import json
 import logging
 import os.path
 import re
+import socket
 from contextlib import contextmanager
 
 import six
@@ -91,6 +92,7 @@ class ServiceGroup(collections.namedtuple(
         rules.extend(_well_known_rules(conf))
         rules.extend(_smartstack_rules(conf, soa_dir, synapse_service_dir))
         rules.extend(_cidr_rules(conf))
+        rules.extend(_fqdn_rules(conf))
         return tuple(rules)
 
     def update_rules(self, soa_dir, synapse_service_dir):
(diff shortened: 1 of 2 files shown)


bchess:
I'm not sure. It makes me nervous to just rely on the 5-minute cronjob. One example I can think of is S3 endpoints seem to have a TTL as low as _5_ seconds.

Would it be expensive to check the TTL as part of this, and log warnings for TTLs less than, say, 10 mins?

chriskuehl:
I think it might be hard to get the real TTL unless we're willing to do recursive resolution ourselves. Lots of recursive caching nameservers have a "countdown" TTL, e.g.

```bash
$ dig berkeley.edu @8.8.8.8 | grep '^b'
berkeley.edu.           226     IN      A       128.32.203.137
$ dig berkeley.edu @8.8.8.8 | grep '^b'
berkeley.edu.           225     IN      A       128.32.203.137
$ dig berkeley.edu @8.8.8.8 | grep '^b'
berkeley.edu.           224     IN      A       128.32.203.137

```

...but the real TTL is 300 if you query the authoritative nameserver:

```bash
$ dig berkeley.edu @adns1.berkeley.edu | grep '^b'
berkeley.edu.           300     IN      A       128.32.203.137
```

So even for things with large enough TTLs, we'd eventually see it as being less than 10 minutes.

For the use-case I'm thinking of (sauce-connect), they currently _never_ re-resolve the DNS, so this is an obvious improvement. But I can definitely see that this may not be something we want to make available for all services given that it's inherently somewhat behind the true DNS results.

bchess:
IMHO, I'd say "not merge" for now, and may revisit again in the future when we have more use cases

PR #1442, opened 2017-08-08
Add script to generate HAProxy map file

jvperrin:
It also sends commands via the HAProxy socket file to add and remove entries from the map without requiring HAProxy to reload. It'll just run via cron for now, but later on it would be better to use docker [pre and post hooks](https://github.com/moby/moby/issues/28837) for this instead of cron.

### DIFF
--- a/paasta_tools/contrib/get_containers_and_ips.py
+++ b/paasta_tools/contrib/get_containers_and_ips.py
@@ -3,9 +3,39 @@ import argparse
 import os
 import socket
 
+from paasta_tools.utils import atomic_file_write
 from paasta_tools.utils import get_docker_client
 
 HAPROXY_STATS_SOCKET = '/var/run/synapse/haproxy.sock'
+
+
+def get_prev_file_contents(filename):
+    if os.path.isfile(filename):
+        with open(filename, 'r') as fp:
+            prev_lines = [
+                # Remove any empty strings, since split could leave empty
+                # strings if there is any extra whitespace in the file
+                list(filter(None, line.strip().split(' ')))
+                for line
+                in fp.readlines()
+            ]
+            return {line[0]: line[1] for line in prev_lines}
+    return {}
+
+
+def extract_taskid_and_ip(docker_client):
+    service_ips_and_ids = []
+    for container in docker_client.containers():
+        networks = container['NetworkSettings']['Networks']
+        labels = container['Labels']
+
+        # Only add containers that are using bridged networking and are
+        # running as Mesos tasks
+        if 'bridge' in networks and 'MESOS_TASK_ID' in labels:
+            ip_addr = networks['bridge']['IPAddress']
+            task_id = labels['MESOS_TASK_ID']
+            service_ips_and_ids.append((ip_addr, task_id))
+    return service_ips_and_ids
 
 
 def send_to_haproxy(command):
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/get_containers_and_ips.py
```diff
@@ -0,0 +1,105 @@
...
+
+HAPROXY_STATS_SOCKET = '/var/run/synapse/haproxy.sock'
+
+
+def send_to_haproxy(command):
+    s = socket.socket(socket.AF_UNIX)
+    # 1 seconds should be more than enough of a timeout since HAProxy is local
+    s.settimeout(1)
+    s.connect(HAPROXY_STATS_SOCKET)
+    print("Sending {} to HAProxy stats socket".format(command))
```
zareenc: Maybe take out the print statements for the final version, but not a big deal

paasta_tools/contrib/get_containers_and_ips.py
```diff
@@ -0,0 +1,105 @@
...
+    s.send((command + '\n').encode())
+    file_handle = s.makefile()
+    response = file_handle.read().splitlines()
+    print("Response: {}".format(response))
```
zareenc: Same here

paasta_tools/contrib/get_containers_and_ips.py
```diff
@@ -0,0 +1,105 @@
...
+        mode = 'w'
+
+    with open(args.map_file, mode) as f:
+        if file_exists:
+            prev_lines = [line.strip().split(' ') for line in f.readlines()]
+            prev_ip_to_task_id = {line[0]: line[1] for line in prev_lines}
+        else:
+            prev_ip_to_task_id = {}
+
+        print("Previous file: {}".format(prev_ip_to_task_id))
```
zareenc: Same here

paasta_tools/contrib/get_containers_and_ips.py
```diff
@@ -0,0 +1,105 @@
...
+
+    file_exists = os.path.isfile(args.map_file)
+    if file_exists:
+        mode = 'r+'
+    else:
+        mode = 'w'
+
+    with open(args.map_file, mode) as f:
+        if file_exists:
+            prev_lines = [line.strip().split(' ') for line in f.readlines()]
```
zareenc: Since the map files allow for any arbitrary amount of whitespace in between the key and value, it may be good to make the parsing account for that in the future (e.g. in case the map is initialized before running the script with a different amount of whitespace). This could potentially mess up the assumption that the first two keys are the key and the value. But this should be fine for our purposes.

paasta_tools/contrib/get_containers_and_ips.py
jvperrin: Yeah, I thought they could be useful and shouldn't matter much since it's just being run via cron, but that's a good point, they shouldn't really be here still.

paasta_tools/contrib/get_containers_and_ips.py
```diff
@@ -0,0 +1,105 @@
...
+            prev_ip_to_task_id = {line[0]: line[1] for line in prev_lines}
+        else:
+            prev_ip_to_task_id = {}
+
+        print("Previous file: {}".format(prev_ip_to_task_id))
+
+        new_lines = []
+        ip_addrs = []
+        for container in client.containers():
```
solarkennedy: This function is pretty nested and a little hard to read.

As an alternative, what if this loop called a function like `extract_taskid_and_ip` that returned task_id and ip.

Then this script would return "just" the mapping.


Then have a different function that is concerned with taking the existing mapping from the file and comparing the two, and sending the differences off to a function that does the send_to_proxy stuff. (which is concerned with add vs set)

update_haproxy_mappings(new_mappings - old_mappings)
delete_haproxy_mappings(old_mappings - new_mappings)

paasta_tools/contrib/get_containers_and_ips.py
```diff
@@ -0,0 +1,105 @@
...
+        for ip_addr in prev_ip_to_task_id.keys():
+            # Remove any keys for containers that are no longer running
+            if ip_addr not in ip_addrs:
+                send_to_haproxy('del map {} {}'.format(
+                    args.map_file,
+                    ip_addr,
+                ))
+
+        # Replace the file contents with the new map
+        if file_exists:
```
solarkennedy: See if you can use the exiting `atomic_file_write` function we have for this

paasta_tools/contrib/get_containers_and_ips.py
```diff
@@ -0,0 +1,105 @@
...
+
+    client = get_docker_client()
+
+    file_exists = os.path.isfile(args.map_file)
+    if file_exists:
+        mode = 'r+'
+    else:
+        mode = 'w'
+
+    with open(args.map_file, mode) as f:
```
solarkennedy: This is way too much for a main function to do. Instead make a function to read in the contents of the file in this format and make it do the opening and have it return the data, returning and empty {} if the file doesn't exist.

paasta_tools/contrib/get_containers_and_ips.py
```diff
@@ -0,0 +1,105 @@
...
+                        method,
+                        args.map_file,
+                        ip_addr,
+                        task_id,
+                    ))
+
+                new_lines.append('{} {}'.format(ip_addr, task_id))
+
+        for ip_addr in prev_ip_to_task_id.keys():
+            # Remove any keys for containers that are no longer running
```
solarkennedy: This code will be more self-documenting with a function like `remove_stopped_container_entries()` that takes in a diff and updates haproxy and stuff.

jvperrin:
@solarkennedy I'd like to merge this now so it can be rolled out sooner rather than later, but you make some very good points. I think the organization of this file could be improved a lot to make it more modular and readable, so I'll create a ticket for that.

jvperrin:
Actually, never mind on that, I'll delay on this a bit and just use puppet for now to get it working for tomorrow and then come back and improve this before merging.

PR #1448, opened 2017-08-09
only consider maintenance role for utilization

Rob-Johnson:
closes JIRA PAASTA-12410

rather than counting *all* reserved resources as used, only count the
resources reserved by the maintenance mode

### DIFF
--- a/paasta_tools/metrics/metastatus_lib.py
+++ b/paasta_tools/metrics/metastatus_lib.py
@@ -39,6 +39,7 @@ ResourceUtilization = namedtuple('ResourceUtilization', ['metric', 'total', 'fre
 EXPECTED_HEALTHY_FRAMEWORKS = 2
 HIGH_QUEUE_GAUGE = 'org.apache.mesos.chronos.scheduler.jobs.TaskManager.highQueueSize'
 QUEUE_GAUGE = 'org.apache.mesos.chronos.scheduler.jobs.TaskManager.queueSize'
+MAINTENANCE_RESOURCE = 'maintenance'
 
 
 def get_num_masters():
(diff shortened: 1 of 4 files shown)


sagar8192:
I am not 100% certain about this change because I am not familiar with paasta code. Seems okay to me, though.

paasta_tools/metrics/metastatus_lib.py
```diff
@@ -39,6 +39,7 @@
 EXPECTED_HEALTHY_FRAMEWORKS = 2
 HIGH_QUEUE_GAUGE = 'org.apache.mesos.chronos.scheduler.jobs.TaskManager.highQueueSize'
 QUEUE_GAUGE = 'org.apache.mesos.chronos.scheduler.jobs.TaskManager.queueSize'
+MAINTENANCE_RESOURCE = 'maintenance'
```
nhandler: It would be nice to be able to import this value from somewhere and share that with `paasta_tools/mesos_maintenance.py` (which currently just hard codes `maintenance`

tests/metrics/test_metastatus_lib.py
```diff
@@ -1037,3 +1037,19 @@ def test_get_mesos_disk_status():
...
     actual = metastatus_lib.get_mesos_disk_status(metrics)
     assert actual == (100, 50, 50)
+
+
+def test_reserved_maintenence_resources_no_maintenenance():
+    actual = metastatus_lib.reserved_maintenence_resources({})
+    assert all([actual[x] == 0 for x in ['cpus', 'mem', 'disk']])
+
+
+def test_reserved_maintenence_resources():
```
nhandler: Can you also add a simple test that has resources reserved for the `maintenance` role plus some other bogus role?

PR #1449, opened 2017-08-09
Remove trailing slashes before de-duping volume mounts

jglukasik:
To fix the problem we saw with mounting a volume "/foo/bar", and an extra_volume mount of "/foo/bar/" causing a "Duplicate mount point" error


EvanKrall:
How is this not a problem for Marathon?

EvanKrall:
I can't seem to find anything in the Marathon source deduping things. Mesos [doesn't like duplicates](https://github.com/apache/mesos/blob/master/src/slave/containerizer/mesos/isolators/docker/volume/isolator.cpp#L249).

I think the only thing saving us from this in prod is that the one path we've seen this on so far is only listed in adhoc-*.yaml files, not any marathon-*.yaml files.

solarkennedy:
@jglukasik can you confirm that if we deployed this now, that it would not cause a big bounce?

EvanKrall:
How about the patch I proposed?

```
deduped = {v['containerPath'].rstrip('/'): v for v in volumes}.values()
```

This strips the / when deduping, but not for the actual value passed in. I believe nothing that we currently have deployed could be changed by this.

jglukasik:
@EvanKrall  I'm not sure how that would preserve the original value, since we do a ``return sort_dicts(deduped)`` at the end of ``get_volumes``, so we'd return the "deslashed" volume names, just like the original PR.

@solarkennedy I believe this _would_ cause a big bounce, based on the output of this command ``cat /etc/paasta/volumes.json  | grep '/"'``

EvanKrall:
The suggested code constructs a dictionary of the form `{deslashed_container_path: {"containerPath": original_container_path, "hostPath": original_host_path}}`:

```
>>> volumes = [{'containerPath': '/foo/', 'hostPath': '/bar'}]
>>> {v['containerPath'].rstrip('/'): v for v in volumes}
{'/foo': {'containerPath': '/foo/', 'hostPath': '/bar'}}
>>> {v['containerPath'].rstrip('/'): v for v in volumes}.values()
dict_values([{'containerPath': '/foo/', 'hostPath': '/bar'}])
```

So the key of this inner dictionary is deslashed, but the value is the original entry from `volumes`. Since we return `.values()` on this inner dictionary, the deslashed values are dropped, and the original containerPath value is retained.

PR #1452, opened 2017-08-10
split up metastatus checks

matthewbentley:
Full list of checks that can fail:

mesos:
- [x] assert\_quorum\_size: masters >= quorum
- [x] assert_no_duplicate\_frameworks
- [x] assert\_{cpu,memory,disk}\_health: cpu, mem, disk are below threshold
- [x] assert\_{connected,disconnected,active,inactive}\_framekworks: connected and active are as expected (2), disconnected and inactive are 0; should be changed to check for one connected, active marathon and one connected, active chronos

marathon:
- [x] assert\_marathon\_apps: marathon has apps running

chronos:
It looks like none of the checks can fail, but we should add:
- [x] assert_chronos_jobs: chronos has jobs


matthewbentley:
check_duplicate_frameworks tends to fail due to `    CRITICAL: Framework taskproc-default has 4 instances running--expected no more than 1.`

matthewbentley:
@Rob-Johnson can you look at this before I spend a bunch of time writing tests? Just to make sure I'm coding in the right direction

paasta_tools/metrics/metastatus_lib.py
```diff
@@ -215,7 +215,7 @@ def assert_tasks_running(metrics):
     )
 
 
-def assert_no_duplicate_frameworks(state):
+def assert_no_duplicate_frameworks(state, ignore=[]):
```
Rob-Johnson: I'd actually just make this ``assert_no_duplicate_frameworks(state, framework_name`` and call it for marathon + chronos, rather than the inverse

paasta_tools/metrics/metastatus_lib.py
```diff
@@ -550,12 +574,13 @@ def get_marathon_status(client):
     )
 
 
-def assert_chronos_scheduled_jobs(client):
+def assert_chronos_scheduled_jobs(client, can_fail=False):
```
Rob-Johnson: is there any reason why we shouldn't have 0 jobs? I say ditch the can_fail - if you don't want to check it, then dont call the function

paasta_tools/monitoring/check_mesos_active_frameworks.py
```diff
@@ -0,0 +1,54 @@
...
+        'Will fail if any of these are not found',
+    )
+
+
+def check_mesos_active_frameworks():
+    options = parse_args()
+    expected = options.expected.split(',')
+    master = get_mesos_master()
+    try:
+        state = master.state
```
Rob-Johnson: I guess this is the bad part about splitting these checks up - we'll be calling this often.

matthewbentley: It only seems to double the number of calls to state (rather than one for metastatus, one for each of the two checks that uses it).

PR #1453, opened 2017-08-10
install paasta dependencies in relevant containers

keymone:
install dependencies in mesosmaster and playground dockerfiles, otherwise we don't have things like task_processing lib in example cluster


EvanKrall:
The first line of requirements-dev.txt is `-r requirements.txt` so this shouldn't make a difference.

keymone:
@EvanKrall you're right. docker-compose was trolling me.

PR #1455, opened 2017-08-11
Add gpu resource for remote-run and cli (for hook validation)

charleskwwan:
## Summary
- add gpus resource to remote-run and basic instance config (in utils)
- add gpus functions for basic instance config
- add gpus to adhoc service schema (used by py-gitolite hooks to check yelpsoa-configs)
- add tests for gpus in utils; none for remote-run since no tests yet
- add gpus to paasta metastatus as another resource
- modify the ResourceInfo namedtuple to have gpus default to 0
## Testing
- tox
- manual testing for instance gpus
- manual testing for adhoc service schema - added a gpus instance in yelpsoa-config's taskproc-canary service and ran it against my schema using cli.cmds.validate
## Notes
- I will run an end-to-end test once I have updated yelpsoa-configs - validation hooks should use paasta's updated gpus schema
- Does anyone know if I need to update any version numbers or what the deployment path is to guarantee the hooks will use my changes?
- Not sure if metastatus for gpus is strictly necessary since we don't care about its threshold and we never reserve any

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
+import math
 import os
 import time
 from collections import namedtuple
@@ -213,10 +214,13 @@ class ClusterAutoscaler(ResourceLogMixin):
         self.log.debug(region_pool_utilization_dict)
         free_pool_resources = region_pool_utilization_dict['free']
         total_pool_resources = region_pool_utilization_dict['total']
-        utilization = 1.0 - min([
-            float(float(pair[0]) / float(pair[1]))
-            for pair in zip(free_pool_resources, total_pool_resources)
-        ])
+        free_percs = []
+        for pair in zip(free_pool_resources, total_pool_resources):
+            free, total = pair[0], pair[1]
+            if math.isclose(total, 0):
+                continue
+            free_percs.append(float(free) / float(total))
+        utilization = 1.0 - min(free_percs)
         target_utilization = self.pool_settings.get('target_utilization', DEFAULT_TARGET_UTILIZATION)
         return utilization - target_utilization
 
--- a/paasta_tools/cli/schemas/adhoc_schema.json
+++ b/paasta_tools/cli/schemas/adhoc_schema.json
@@ -27,7 +27,7 @@
                 "default": 1024
             },
             "gpus": {
-                "type": "number",
+                "type": "integer",
                 "minimum": 0,
                 "exclusiveMinimum": true,
                 "default": 0
--- a/paasta_tools/cli/schemas/adhoc_schema.json
+++ b/paasta_tools/cli/schemas/adhoc_schema.json
@@ -26,6 +26,12 @@
                 "exclusiveMinimum": true,
                 "default": 1024
             },
+            "gpus": {
+                "type": "number",
+                "minimum": 0,
+                "exclusiveMinimum": true,
+                "default": 0
+            },
             "cmd": {
                 "type": "string"
             },
--- a/paasta_tools/metrics/metastatus_lib.py
+++ b/paasta_tools/metrics/metastatus_lib.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 import copy
 import itertools
+import math
 from collections import Counter
 from collections import namedtuple
 from collections import OrderedDict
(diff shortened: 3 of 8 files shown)


paasta_tools/paasta_remote_run.py
```diff
@@ -171,6 +172,8 @@ def paasta_to_task_config_kwargs(
         'uris': [uris],
         'docker_parameters': docker_parameters,
     }
+    if gpus > 0:
+        kwargs['gpus'] = float(gpus)
```
solarkennedy: is it really a float?

paasta_tools/utils.py
```diff
@@ -410,6 +419,12 @@ def check_disk(self):
                 return False, 'The specified disk value "%s" is not a valid float or int.' % disk
         return True, ''
 
+    def check_gpus(self):
```
solarkennedy: I think this is overkill, can you instead add it to the json schema?

solarkennedy: Wait sorry, you did. :)

charleskwwan:
@nhandler told me that paasta metastatus also needs to be updated to include gpus, so I've pushed another change for that.

paasta_tools/paasta_remote_run.py
charleskwwan: doing this because task processing and mesos expects floats, even though gpus cannot be partially used.

paasta_tools/cli/schemas/adhoc_schema.json
```diff
@@ -26,6 +26,12 @@
                 "exclusiveMinimum": true,
                 "default": 1024
             },
+            "gpus": {
```
sagar8192: If gpus cannot be used partially, we probably want to enforce that here. This should be an int instead? Also, if we don't want to allow fractions for gpus, we should change taskproc to only accept ints.

paasta_tools/metrics/metastatus_lib.py
```diff
@@ -32,9 +32,10 @@
 from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import print_with_indent
 
-
+class ResourceInfo(namedtuple('ResourceInfo', ['cpus', 'mem', 'disk', 'gpus'])):
```
sagar8192: Instead of inheriting from namedtuple, can we just make cpus, mem, disk and gpus class members?

paasta_tools/metrics/metastatus_lib.py
```diff
@@ -78,8 +79,20 @@ def get_mesos_disk_status(metrics):
     return total, used, available
 
 
+def get_mesos_gpu_status(metrics):
+    """Takes in the mesos metrics and analyzes them, returning gpus status.
+
+    :param metrics: mesos metrics dictionary
+    :returns: Tuple of the output array and is_ok bool
+    """
+    total = metrics['master/gpus_total']
```
sagar8192: Are these available in master right now? I think mesos only adds it if we have slaves with gpus in the cluster. It won't have it otherwise.  Let's check if this keys is available or not.

paasta_tools/metrics/metastatus_lib.py
charleskwwan: I can't because there are parts of paasta that used namedtuple specific functions.

paasta_tools/metrics/metastatus_lib.py
charleskwwan: There are available as far as know. I checked mesosstage, which has never had GPUs and it has them: http://mesos.paasta-mesosstage.yelp:5050/metrics/snapshot

paasta_tools/cli/schemas/adhoc_schema.json
charleskwwan: Made change to taskproc, will follow through here.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -213,10 +213,13 @@ def get_mesos_utilization_error(
...
         total_pool_resources = region_pool_utilization_dict['total']
-        utilization = 1.0 - min([
-            float(float(pair[0]) / float(pair[1]))
-            for pair in zip(free_pool_resources, total_pool_resources)
-        ])
+        used_percs = []
+        for pair in zip(free_pool_resources, total_pool_resources):
+            try:
+                used_percs.append(float(float(pair[0]) / float(pair[1])))
+            except ZeroDivisionError:  # should occur for 0 gpus, which is valid
```
keymone: if you can easily check for error condition there's no reason to use exceptions:

    for used, total in zip(free_pool_resources, total_pool_resources):
        if math.isclose(total, 0):
            continue
        used_percs.append(float(used)/float(total))

charleskwwan: Good point! I'll make the change.

PR #1456, opened 2017-08-11
Remove six compatibility layer

asottile:


solarkennedy:
So great. My only reservation is that we still have this in py27:
https://github.com/Yelp/synapse-tools/blob/c0f8cd6cf7c8436d49030f638d4fd1c0075f8a85/src/synapse_tools/configure_synapse.py#L18

Given there are no six things in marathon_tools in this PR, it won't hurt us in the short term, right? (long term of course we should upgrade those packages to py3)

asottile:
Correct, it should be fine

PR #1458, opened 2017-08-14
Allow 'extra_docker_args'

### LINKED ISSUE #1411: Add a docker_opts_whitelist, to support somewhat arbitrary docker args
So we don't have to do this whenever someone needs a one-off arg: https://github.com/Yelp/paasta/pull/1409

matthewbentley:
Rather than adding a one-off rule for each one-time-use docker arg

Fixes #1411

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -254,11 +254,12 @@ class InstanceConfig(object):
                 {"key": "label", "value": "paasta_service=%s" % self.service},
                 {"key": "label", "value": "paasta_instance=%s" % self.instance},
             ])
-        shm = self.get_shm_size()
-        if shm:
-            parameters.extend([
-                {"key": "shm-size", "value": "%s" % shm},
-            ])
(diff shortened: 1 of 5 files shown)


paasta_tools/utils.py
```diff
@@ -81,6 +81,8 @@
 
 INSTANCE_TYPES = ('marathon', 'chronos', 'paasta_native', 'adhoc')
 
+DOCKER_ARGS_WHITELIST = ('shm-size')
```
solarkennedy: why do we really need a whitelist?

Is the attack vector someone editing their config using something like --privileged? (which is another requested argument)

I don't think we should be trying to protect against things like that at this layer in the codebase. If we are going to allow free-form docker arguments, just make it free-form.

matthewbentley: The intent here was to protect against adding --privileged (or similar).  But if we're going to add --privileged anyways, there's not much purpose to this.

PR #1462, opened 2017-08-15
Paasta args mixer rewrite

matthewbentley:
Currently loses the functionality where paasta status suggests what instances you might mean if no instances are returned, and where start_stop_restart lists valid clusters if you specify an invalid one.


tests/cli/test_cmds_start_stop_restart.py
```diff
@@ -410,11 +406,13 @@ def test_cluster_list_defaults_to_all():
     return True
 
 
-@mock.patch('paasta_tools.cli.cmds.start_stop_restart.figure_out_service_name', autospec=True)
+@pytest.mark.xfail(reason='This functionality will be moved')
+# re ^^: rewriting paasta args mixer removed the instance suggest feature, but it should be re-added
```
solarkennedy: :(

solarkennedy:
In general I approve of the centralization of the filtering in the mixer.
I'm sad about losing the suggestion stuff, that was a really big nice to have.

Asking @oktopuz to also review.

paasta_tools/cli/cmds/status.py
```diff
@@ -347,81 +324,72 @@ def verify_instances(args_instances, service, clusters):
...
+    if args.service:
+        filters.append(lambda conf: conf.get_service() in service.split(','))
+
+    if args.deploy_group:
+        filters.append(lambda conf: conf.get_deploy_group() in args.deploy_group.split(','))
+
+    return filters
+
+
+def apply_args_filters(args):
```
oktopuz: Could you add a docstring here?
We most certainly will reuse this function in other paasta subcommands, let's make it easier to reuse.

tests/cli/test_cmds_status.py
```diff
@@ -23,17 +23,25 @@
...
-from paasta_tools.cli.cmds.status import report_status
 from paasta_tools.cli.cmds.status import verify_instances
 from paasta_tools.cli.utils import NoSuchService
 from paasta_tools.cli.utils import PaastaCheckMessages
 from paasta_tools.cli.utils import PaastaColors
 
 
+def make_fake_instance_conf(cluster, service, instance, deploy_group=None):
+    conf = MagicMock()
+    conf.get_cluster = lambda: cluster
```
oktopuz: Out of curiosity, is this just another way to write 'conf.get_cluster.return_value = cluster'?
Or is there any difference?

matthewbentley: Yeah it looks like it's the same.  Should I do `.return_value` or use the lambda?

tests/cli/test_cmds_start_stop_restart.py
matthewbentley: I will attempt to re-add this once the other feedback has been addressed

paasta_tools/cli/cmds/status.py
matthewbentley: fixed

matthewbentley:
I'm a bit worried about the logic in `get_filters`.  `paasta {stop,start,restart}` only requires a cluster name, so as this is written now it will allow somebody to run `paasta stop -c norcal-devc`, which would match every instance in norcal-devc.

I think this could be fixed by calling `service = figure_out_service_name` in some situations, and requiring instances to match that, but I don't think we want that in all cases. For example, we may want to allow `paasta status -c norcal-devc --owner operations`, but we probably don't want to allow (without a warning) `paasta stop -c norcal-devc --owner operations` (when --owner is added in the future, that is).

This could alternatively be fixed by requiring some sort of `--yes-please-do-this-really` flag if more than a certain number of instances would be affected by start-stop-restart (or if more than one service would be affected?).

matthewbentley:
Something like this in start-stop-restart:
```
+    if len(affected_services) > 1:
+        paasta_print(PaastaColors.red("Warning: trying to start/stop/restart multiple services"))
+        if args.allow_multiple_services:
+            paasta_print(PaastaColors.red("    --allow-multiple-services specified, continuing anyways"))
+        else:
+            paasta_print("    please specify --allow-multiple-services for this command to take affect")
+            return 1
```

Rob-Johnson:
As we spoke about IRL, lets use the chooser to print all the service.instances that the command will impact when the number is >1 and ask the user to confirm

paasta_tools/cli/cmds/status.py
```diff
@@ -347,81 +324,72 @@ def verify_instances(args_instances, service, clusters):
...
-def paasta_args_mixer(args, service):
-    if args.deploy_group is None:
-        cluster_whitelist = args.clusters.split(",") if args.clusters is not None else []
-        if args.instances is not None:
-            instance_whitelist = verify_instances(args.instances, service, cluster_whitelist)
-        else:
-            instance_whitelist = []
+def get_filters(args):
+    if not any([args.clusters, args.instances, args.owner, args.deploy_group]):
+        service = figure_out_service_name(args, soa_dir=args.soa_dir)
```
oktopuz: I see.

When args.service is None, that line is not executed.
This leads to a bit different behavior than we have now when it tries to guess service name (https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/utils.py#L674).

```
$ cd ../yelpsoa-config/yelp-main
$ paasta status -c norcal-stagef
```

Expected result: it shows status of all yelp-main instances in norcal-stagef.
Actual result: paasta status iterates over all services.

oktopuz:
What about making either --service or --owner required, so that it would fail if none of them are specified?
I can't think of any cases when we actually needed to restart or to get status of ALL services.

matthewbentley:
We might want to fail (or warn) on start/stop/restart with --owner, mostly to keep this from being a foot-gun, which is why I added the check for > 1 services.

paasta_tools/cli/cmds/status.py
oktopuz: Except this issue ^^ everything else lgtm. Fix it & ship it.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -167,91 +174,77 @@ def print_chronos_message(desired_state):
...
-            cluster_instances = list_all_instances_for_service(service, clusters=[cluster], soa_dir=soa_dir)
-            paasta_print("no instances specified; restarting all instances for service")
+        for cluster, services_instances in pargs.items():
+            paasta_print("Cluster %s:" % cluster)
+            for service, instances in services_instances.items():
+                paasta_print("    Service %s:" % service)
+                paasta_print("        Instances %s" % ",".join(instances))
+
+        if args.allow_multiple_services:
+            paasta_print(PaastaColors.red("    --allow-multiple-services specified, continuing anyways"))
```
Rob-Johnson: I don't think that having both a flag and a the choice interaction is great. I'd just use the choice interaction instead, I think.

PR #1466, opened 2017-08-17
Make deployd bounce queue a priority queue

mattmb:
This changes the bounce queue to a python PriorityQueue. It adds
everything with priority 0 by default but lets you give a service
instance a higher or lower priority. Workers will fetch higher priority
instances first and fall back to FIFO if there is a tie.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -114,6 +114,13 @@ instance MAY have:
     This allows bounces to proceed in the face of a percentage of failures.
     It doesn’t affect any other bounce method but crossover.
 
+  * ``bounce_priority``: an integer priority that informs paasta-deployd which service
+    instances should take priority over each other. The default priority is 0 and higher numbers
+    are considered higher priority. For example: if there are three service instances that need
+    bouncing: the first with a ``bounce_prioirty`` -1, the second with no ``bounce_priority`` and the
+    third with ``bounce_priority`` 1. Then paasta-deployd will prioritise the bounce of the third
+    service instance, then the second service instance and finally the first service instance.
+
   * ``drain_method``: Controls the drain method; see `drain_lib
     <generated/paasta_tools.drain_lib.html>`_. Defaults to ``noop`` for
     instances that are not in Smartstack, or ``hacheck`` if they are.
--- a/paasta_tools/deployd/common.py
+++ b/paasta_tools/deployd/common.py
@@ -1,6 +1,7 @@
 import logging
 import time
 from collections import namedtuple
+from queue import PriorityQueue
 from queue import Queue
 from threading import Thread
 
@@ -8,13 +9,14 @@ from paasta_tools.marathon_tools import DEFAULT_SOA_DIR
 from paasta_tools.marathon_tools import get_all_marathon_apps
 from paasta_tools.marathon_tools import get_marathon_client
 from paasta_tools.marathon_tools import load_marathon_config
+from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.marathon_tools import load_marathon_service_config_no_cache
 from paasta_tools.utils import InvalidJobNameError
 from paasta_tools.utils import NoDeploymentsAvailable
 from paasta_tools.utils import NoDockerImageError
 
 BounceTimers = namedtuple('BounceTimers', ['processed_by_worker', 'setup_marathon', 'bounce_length'])
-ServiceInstance = namedtuple(
+BaseServiceInstance = namedtuple(
     'ServiceInstance', [
         'service',
         'instance',
@@ -22,15 +24,46 @@ ServiceInstance = namedtuple(
         'watcher',
         'bounce_timers',
         'failures',
+        'priority',
     ],
 )
+
+
+class ServiceInstance(BaseServiceInstance):
+    __slots__ = ()
+
+    def __new__(cls, service, instance, watcher, cluster, bounce_by, failures=0, bounce_timers=None, priority=None):
+        if priority is None:
+            priority = cls.get_priority(service, instance, cluster)
+        return super().__new__(
+            _cls=cls,
+            service=service,
+            instance=instance,
+            watcher=watcher,
+            bounce_by=bounce_by,
+            failures=failures,
+            bounce_timers=bounce_timers,
+            priority=priority,
+        )
+
+    def get_priority(service, instance, cluster):
+        try:
+            config = load_marathon_service_config(
+                service=service,
+                instance=instance,
+                cluster=cluster,
+                soa_dir=DEFAULT_SOA_DIR,
+            )
+        except (NoDockerImageError, InvalidJobNameError, NoDeploymentsAvailable) as e:
+            return 0
+        return config.get_bounce_priority()
 
 
 class PaastaThread(Thread):
 
     @property
     def log(self):
-        name = '.'.join([self.__class__.__module__, self.__class__.__name__])
+        name = '.'.join([type(self).__module__, type(self).__name__])
         return logging.getLogger(name)
 
 
@@ -42,7 +75,7 @@ class PaastaQueue(Queue):
 
     @property
     def log(self):
-        name = '.'.join([self.__class__.__module__, self.__class__.__name__])
+        name = '.'.join([type(self).__module__, type(self).__name__])
         return logging.getLogger(name)
 
     def put(self, item, *args, **kwargs):
@@ -50,7 +83,32 @@ class PaastaQueue(Queue):
         Queue.put(self, item, *args, **kwargs)
 
 
-def rate_limit_instances(instances, number_per_minute, watcher_name):
+class PaastaPriorityQueue(PriorityQueue):
+
+    def __init__(self, name, *args, **kwargs):
+        self.name = name
+        PriorityQueue.__init__(self, *args, **kwargs)
+        self.counter = 0
+
+    @property
+    def log(self):
+        name = '.'.join([type(self).__module__, type(self).__name__])
+        return logging.getLogger(name)
+
+    def put(self, priority, item, *args, **kwargs):
+        self.log.debug("Adding {} to {} queue with priority {}".format(item, self.name, priority))
+        # this counter is to preserve the FIFO nature of the queue, it increments on every put
+        # and the python PriorityQueue sorts based on the first item in the tuple (priority)
+        # and then the second item in the tuple (counter). This way all items with the same
+        # priority come out in the order they were entered.
+        self.counter += 1
+        PriorityQueue.put(self, (priority, self.counter, item), *args, **kwargs)
+
+    def get(self, *args, **kwargs):
+        return PriorityQueue.get(self, *args, **kwargs)[2]
+
+
+def rate_limit_instances(instances, cluster, number_per_minute, watcher_name):
     service_instances = []
     if not instances:
         return []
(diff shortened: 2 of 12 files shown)


mattmb:
The more I think about this the more I think it makes sense @solarkennedy 

This gives us a better mechanism for the big bounce too. i.e. for adding services at startup or when public config changes then we add them all with a v. low priority. This way they can never interfere with regular pushes etc.

mattmb:
Can't immediately work out what's up with the itests, they seem to be failing to build the containers...

mattmb:
The test failure was a typo in the new config option itests++

paasta_tools/deployd/common.py
```diff
@@ -1,31 +1,61 @@
...
 from paasta_tools.marathon_tools import load_marathon_config
+from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.marathon_tools import load_marathon_service_config_no_cache
 from paasta_tools.utils import InvalidJobNameError
 from paasta_tools.utils import NoDeploymentsAvailable
 from paasta_tools.utils import NoDockerImageError
 
 BounceTimers = namedtuple('BounceTimers', ['processed_by_worker', 'setup_marathon', 'bounce_length'])
-ServiceInstance = namedtuple(
+ServiceInstanceTuple = namedtuple(
```
Rob-Johnson: why? we don't normally encode the type in the class?

paasta_tools/deployd/common.py
```diff
@@ -1,31 +1,61 @@
...
         'watcher',
         'bounce_timers',
         'failures',
+        'priority',
     ],
 )
 
 
+def ServiceInstance(service, instance, watcher, cluster, bounce_by, failures=0, bounce_timers=None, priority=None):
+    if priority is None:
```
solarkennedy: do we really need this? why can't we get the priority when we call this instead of passing None?

paasta_tools/deployd/common.py
```diff
@@ -1,31 +1,61 @@
...
+        instance=instance,
+        watcher=watcher,
+        bounce_by=bounce_by,
+        failures=failures,
+        bounce_timers=bounce_timers,
+        priority=priority,
+    )
+
+
+def get_priority(service, instance, cluster):
```
solarkennedy: yuck, don't we already do this (read the marathon config)?

paasta_tools/deployd/common.py
```diff
@@ -50,7 +80,28 @@ def put(self, item, *args, **kwargs):
...
 
-def rate_limit_instances(instances, number_per_minute, watcher_name):
+class PaastaPriorityQueue(PriorityQueue):
+
+    def __init__(self, name, *args, **kwargs):
+        self.name = name
+        PriorityQueue.__init__(self, *args, **kwargs)
+        self.counter = 0
+
+    @property
```
solarkennedy: yay we need more properties

paasta_tools/deployd/common.py
```diff
@@ -50,7 +80,28 @@ def put(self, item, *args, **kwargs):
...
+        name = '.'.join([self.__class__.__module__, self.__class__.__name__])
+        return logging.getLogger(name)
+
+    def put(self, priority, item, *args, **kwargs):
+        self.log.debug("Adding {} to {} queue with priority {}".format(item, self.name, priority))
+        self.counter += 1
+        PriorityQueue.put(self, (priority, self.counter, item), *args, **kwargs)
+
+    def get(self, *args, **kwargs):
+        return PriorityQueue.get(self, *args, **kwargs)[2]
```
solarkennedy: Hmm, what is the 2 magic number here?

mattmb: It's the `item` from the queue, see the tuple we pass to `PriorityQueue.put`

paasta_tools/deployd/master.py
```diff
@@ -245,6 +247,7 @@ def prioritise_bouncing_services(self):
             self.inbox_q.put(ServiceInstance(
                 service=service,
                 instance=instance,
+                cluster=self.config.get_cluster(),
                 watcher=self.__class__.__name__,
                 bounce_by=int(time.time()),
                 bounce_timers=None,
```
solarkennedy: Wouldn't here be a good place to do priority=self.config.get_bounce_priority? 

Is the inbox_q not a priority queue?

paasta_tools/long_running_service_tools.py
```diff
@@ -107,6 +107,11 @@ def get_healthcheck_mode(self, service_namespace_config):
             raise InvalidHealthcheckMode("Unknown mode: %s" % mode)
         return mode
 
+    def get_bounce_priority(self):
+        """Gives a priority to each service instance which deployd will use to prioritise services
+        higher numbers are higher priority"""
+        return self.config_dict.get('bounce_priority', 0) * -1
```
solarkennedy: This -1 is going to cause confusion, but *not* giving -1 is going to cause even different confusion! :(

Add some docs in the RST file with examples to make this really clear

paasta_tools/deployd/common.py
Rob-Johnson: yeah, this pattern gets a -1 from me.

paasta_tools/deployd/common.py
solarkennedy: Also won't it be cached? And therefore anytime we change this it will never get updated in memory?

mattmb: I think I prefer it here. I was a bit worried that we will read the config every time we create a `ServiceInstance` but there is a 5s cache so I think it'll be okay? We do read the config in some cases when we create the `ServiceInstance` but even then we don't expose it cleanly. I feel like this might be the least messy way to do it.

paasta_tools/deployd/common.py
mattmb: See the comment below on why. Another thing this gives us is a place to override the config value from the code. e.g. if we want the startup/big bounce code to set a priority of -99 on all instances regardless of their configured priority.

paasta_tools/deployd/common.py
```diff
@@ -50,7 +80,28 @@ def put(self, item, *args, **kwargs):
         Queue.put(self, item, *args, **kwargs)
 
 
-def rate_limit_instances(instances, number_per_minute, watcher_name):
+class PaastaPriorityQueue(PriorityQueue):
+
+    def __init__(self, name, *args, **kwargs):
+        self.name = name
+        PriorityQueue.__init__(self, *args, **kwargs)
```
Rob-Johnson: don't you do ```super().__init__()``` here?

paasta_tools/deployd/master.py
mattmb: `self.config` here is the public config not the config of the service instance.

I've left the `InboxQueue` alone for now. We're basically getting two "priority mechanisms", the `bounce_by` which is used in the `InboxQueue` and the `bounce_priority` for the `BounceQueue`. Long term we might want to pull the concept of the InboxQueue, or limit it to just staggering repeated runs of `setup_marathon_job`

paasta_tools/deployd/common.py
```diff
@@ -50,7 +80,28 @@ def put(self, item, *args, **kwargs):
...
+        self.counter = 0
```
Rob-Johnson: whats this counter for? just the number of items in the queue?

paasta_tools/long_running_service_tools.py
mattmb: Yeah, it's somehow nasty both ways isn't it. Docs are a good shout though, I'll add something explicit before we ship.

paasta_tools/deployd/common.py
```diff
@@ -50,7 +80,28 @@ def put(self, item, *args, **kwargs):
...
+
+    @property
+    def log(self):
+        name = '.'.join([self.__class__.__module__, self.__class__.__name__])
+        return logging.getLogger(name)
+
+    def put(self, priority, item, *args, **kwargs):
+        self.log.debug("Adding {} to {} queue with priority {}".format(item, self.name, priority))
+        self.counter += 1
+        PriorityQueue.put(self, (priority, self.counter, item), *args, **kwargs)
```
Rob-Johnson: why do we put the counter on here?

paasta_tools/deployd/common.py
mattmb: It's an old type `Class` iirc so doesn't work with super.

paasta_tools/deployd/common.py
mattmb: The counter is to preserve the FIFO nature of the queue. I'll add a comment because it isn't intuitive actually. Essentially the `PriorityQueue` does a sort on the first item in the tuple and then on the second etc. So the we sort first by priority and then by this counter (which always  increments).

paasta_tools/deployd/common.py
mattmb: Naming is hard :-)

I wanted to keep the name `ServiceInstance` for the call from the other modules, how about `_ServiceInstance`? or `InternalServiceInstance` idk?

paasta_tools/deployd/common.py
asottile: the entire codebase is python3, there's no such thing as an "old type class"

paasta_tools/deployd/common.py
```diff
@@ -1,31 +1,61 @@
...
         'bounce_by',
         'watcher',
         'bounce_timers',
         'failures',
+        'priority',
     ],
 )
 
 
+def ServiceInstance(service, instance, watcher, cluster, bounce_by, failures=0, bounce_timers=None, priority=None):
```
asottile: to me it looks like you actually just want to define the initialization:

```python

class NT(collections.namedtuple('NT', ('foo', 'bar'))):
    # restore immutability that subclassing breaks
    __slots__ = ()

    def __new__(cls, foo, bar=5):
        # or with more logic
        return cls(foo, bar)
```

Note that (as with most immutable objects) you must override `__new__` instead of `__init__` (as `__init__` is too late)

paasta_tools/deployd/common.py
```diff
@@ -50,7 +80,28 @@ def put(self, item, *args, **kwargs):
...
+class PaastaPriorityQueue(PriorityQueue):
+
+    def __init__(self, name, *args, **kwargs):
+        self.name = name
+        PriorityQueue.__init__(self, *args, **kwargs)
+        self.counter = 0
+
+    @property
+    def log(self):
+        name = '.'.join([self.__class__.__module__, self.__class__.__name__])
```
asottile: generally prefer `type(self)` over `self.__class__` (python3 actually proposed removing `__class__` as it was leftover from python2 old style classes)

paasta_tools/deployd/common.py
mattmb: I didn't know that was something they'd sorted, I'll poke around and change anything old style to use `super()` 👍

paasta_tools/deployd/common.py
mattmb: My first instinct was for a sub class. I got a bit scared when I read the `namedtuple` source :-S

Let me take another look though, I agree that it would be the right way to do it

mattmb:
@Rob-Johnson @solarkennedy I've pushed some changes to address some of the feedback. Let me know what you think.

docs/source/yelpsoa_configs.rst
```diff
@@ -114,6 +114,13 @@ instance MAY have:
     This allows bounces to proceed in the face of a percentage of failures.
     It doesn’t affect any other bounce method but crossover.
 
+  * ``bounce_priority``: an integer priority that informs paasta-deployd which service
+    instances should take priority over each other. The default priority is 0 and higher numbers
+    are considered higher priority. For example: if there are three service instances that need
+    bouncing: the first with a ``bounce_prioirty`` -1, the second with no bounce_priority and the
```
Rob-Johnson: typo

PR #1467, opened 2017-08-18
Allow port ranges for CIDR rules

chriskuehl:
Tested this via example_happyhour on mesosstage, the rules look like this:

```yaml
- cidr: '169.229.226.0/24'
  port: '1:200'
- cidr: '169.229.226.0/24'
  port: '1024:65535'
```

We want this in order to use sauce-connect (it wants to whitelist ports 1024-65535).

### DIFF
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -249,10 +249,25 @@ def _smartstack_rules(conf, soa_dir, synapse_service_dir):
         yield _yocalhost_rule(port, 'proxy_port ' + namespace)
 
 
+def _ports_valid(ports):
+    for port in ports:
+        try:
+            port = int(port)
+        except ValueError:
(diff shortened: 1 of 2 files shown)


paasta_tools/firewall.py
```diff
@@ -263,20 +278,20 @@ def _cidr_rules(conf):
...
-            try:
-                port = int(port)
-            except ValueError:
-                log.exception(f'Unable to parse port: {port}')
+        if port_str is not None:
+            # port can be either a single port like "443" or a range like "1024:65535"
+            ports = str(port_str).split(':')
+
+            if len(ports) > 2:
+                log.error(f'"port" must be either a single value or a range like "1024:65535": {port}')
```
chriskuehl: oops, need to s/port/port_str/

bchess: yar

PR #1471, opened 2017-08-21
Upgrade pre-commit and hooks

asottile:


asottile:
Note that I also put the lists back to how I would format them -- one nice thing about the new pre-commit is it will make a best effort to not rewrite the yaml literals (finally! yay!) -- which is why I also upgraded pre-commit in this PR

solarkennedy:
I don't mind the []s

PR #1472, opened 2017-08-22
Enable deployd for the example cluster

mattmb:
* Use the new split out paasta-api but in development mode
* Remove deploy marathon cron job
* Start paasta-deployd in mesosmaster container

### DIFF
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -31,7 +31,6 @@ from paasta_tools import marathon_tools
 from paasta_tools.api import settings
 from paasta_tools.utils import load_system_paasta_config
 
-monkey.patch_all()
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


paasta_tools/api/api.py
```diff
@@ -102,6 +101,7 @@ def setup_paasta_api():
 
 
 def main(argv=None):
+    monkey.patch_all()
```
mattmb: @matthewbentley moving this is okay right? One of the problems I was having with the example cluster seemed to be related to the monkey patching. Something to do with using pserve to launch the API server which means we're not in the main thread...

Anyway, this should mean we wont have the monkey patching in the example cluster but we will still have it in production.

matthewbentley: This should work, although the docs recommend doing it as soon as possible (ie at the top).  The itests run the api server "for real", so if the itests pass this should be fine

mattmb:
This also makes the travis build of the example cluster work again:
https://travis-ci.org/Yelp/paasta/jobs/267235781

PR #1474, opened 2017-08-22
Status by owner, round 2

matthewbentley:


paasta_tools/cli/cmds/status.py
```diff
@@ -350,6 +356,18 @@ def get_filters(args):
...
+    if args.owner:
+        owners = args.owner.split(',')
+
+        filters.append(
+            # If the instance owner is None, check the service owner, else check the instance owner
+            lambda conf: get_team(
+                overrides={},
+                service=conf.get_service(),
+                soa_dir=args.soa_dir,
+            ) in owners if conf.get_instance_owner() is None else conf.get_instance_owner() in owners,
```
solarkennedy: Why is get_instance_owner not called "get_team" ?

matthewbentley: Probably because naming things is hard ™.  `get_team` would be a better name.

PR #1477, opened 2017-08-23
Make paasta status show the service name

nhandler:
Previously, you had to determine the service name from the app id. This causes `paasta status` to explicitly specify the service name.

This is PAASTA-12553


matthewbentley:
I could have sworn I added this, but I guess it slipped through the cracks when I re-wrote the args mixer re-write.

PR #1479, opened 2017-08-23
remove --memory-swap option from docker run

oktopuz:
remove --memory-swap option from docker run so that mesos can manage this value.

Internal ticket: PAASTA-12450


oktopuz:
Sure. I'll try.

oktopuz:
closing this for now in favor of https://github.com/Yelp/paasta/pull/1480

PR #1480, opened 2017-08-26
Add 64MB to memory-swap to work around PAASTA-12450

oktopuz:
Internal ticket: PAASTA-12450

Add 64MB to memory-swap to work around PAASTA-12450 and to be able to test mesos-slave with --cgroups_limit_swap.

If --cgroups_limit_swap works as intended mesos-slave should reduce memory.memsw.limit_in_bytes from x + 64MB to x + 32MB.

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -169,10 +169,11 @@ class InstanceConfig(object):
         """Gets the memory-swap value. This value is passed to the docker
         container to ensure that the total memory limit (memory + swap) is the
         same value as the 'mem' key in soa-configs. Note - this value *has* to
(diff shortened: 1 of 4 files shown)


solarkennedy:
This will cause a big bounce and needs to be carefully rolled out. (right?)

If so, please pin paasta tools so someone doesn't accidentally bounce all services on the next release.

oktopuz:
I pinned paasta-tools in norcal-prod, pnw-prod and nova-prod.
@Rob-Johnson, @mattmb, could you make release and roll it out during the UK day time, when the number of deployments is relatively low?

PR #1484, opened 2017-08-29
adjusts get_marathon_services_for_nerve to handle missing marathon files

fboxwala:
As per SMTSTK-192 - configure_nerve shouldn't crash when a marathon file is not found for a container still running. 

get_marathon_services_for_nerve is called [here](https://github.com/yelp/nerve-tools/blob/master/src/nerve_tools/configure_nerve.py#L289)

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -769,8 +769,9 @@ def get_marathon_services_running_here_for_nerve(cluster, soa_dir):
                     continue
                 nerve_dict['port'] = port
                 nerve_list.append((registration, nerve_dict))
-        except KeyError:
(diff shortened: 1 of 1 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -771,6 +771,10 @@ def get_marathon_services_running_here_for_nerve(cluster, soa_dir):
                 nerve_list.append((registration, nerve_dict))
         except KeyError:
             continue  # SOA configs got deleted for this app, it'll get cleaned up
+        except NoConfigurationForServiceError:
```
solarkennedy: Go ahead and throw this on line 772 along with the key error, no need to log or anything.

PR #1489, opened 2017-09-01
remote runs in paasta_serviceinit

keymone:
adhoc instances are now sort of adopted by remote-run cmd, except the "interactive" one because we don't care about it?

also only status command is handled for now.

### DIFF
--- a/paasta_tools/paasta_remote_run.py
+++ b/paasta_tools/paasta_remote_run.py
@@ -412,20 +412,22 @@ def remote_run_stop(args):
         paasta_print(teardown.text)
 
 
-frameworks = None
-def remote_run_list(args, reset_frameworks=False):
-    _, service, cluster, _, instance, _ = extract_args(args)
+def remote_run_frameworks():
+    return get_all_frameworks(active_only=True)
 
-    global frameworks
 
-    if reset_frameworks:
-        frameworks = None
-
+def remote_run_filter_frameworks(service, instance, frameworks=None):
     if frameworks is None:
-        frameworks = get_all_frameworks(active_only=True)
+        frameworks = remote_run_frameworks()
 
     prefix = "paasta-remote %s.%s" % (service, instance)
-    filtered = [f for f in frameworks if f.name.startswith(prefix)]
+    return [f for f in frameworks if f.name.startswith(prefix)]
+
+
+def remote_run_list(args, frameworks=None):
+    _, service, cluster, _, instance, _ = extract_args(args)
+    filtered = remote_run_frameworks_for(
+        service, instance, frameworks=frameworks)
     filtered.sort(key=lambda x: x.name)
     for f in filtered:
         launch_time, run_id = re.match(
--- a/paasta_tools/paasta_remote_run.py
+++ b/paasta_tools/paasta_remote_run.py
@@ -424,8 +424,7 @@ def remote_run_filter_frameworks(service, instance, frameworks=None):
     return [f for f in frameworks if f.name.startswith(prefix)]
 
 
-def remote_run_list(args, frameworks=None):
-    _, service, cluster, _, instance, _ = extract_args(args)
+def remote_run_list_report(service, instance, cluster, frameworks=None):
     filtered = remote_run_filter_frameworks(
         service, instance, frameworks=frameworks,
     )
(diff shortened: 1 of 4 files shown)


paasta_tools/paasta_remote_run.py
```diff
@@ -111,17 +111,36 @@ def extract_args(args):
...
+        if instance_type != 'adhoc':
+            paasta_print(
+                PaastaColors.red(
+                    (
+                        "Please use instance declared in adhoc.yaml for use "
+                        "with remote-run, {} is declared as {}"
+                    ).format(instance, instance_type)
+                )
+            )
+            os._exit(1)
```
Rob-Johnson: I think sys.exit(1) is the 'cleaner' way to exit. os._exit is pretty brutal

paasta_tools/paasta_remote_run.py
```diff
@@ -392,20 +411,35 @@ def remote_run_stop(args):
         paasta_print(teardown.text)
 
 
-def remote_run_list(args):
-    _, service, cluster, _, instance, _ = extract_args(args)
-    frameworks = get_all_frameworks(active_only=True)
+def remote_run_frameworks():
+    return get_all_frameworks(active_only=True)
```
Rob-Johnson: will ``active_only=True`` mean that you only see running frameworks? What if you want to see those that have finished?

paasta_tools/paasta_serviceinit.py
```diff
@@ -143,57 +145,107 @@ def main():
...
+                        instance=instance,
+                        cluster=cluster,
+                        verbose=args.verbose,
+                        soa_dir=args.soa_dir,
+                    )
+                elif instance_type == 'adhoc':
+                    if instance == 'interactive':
+                        continue
+                    if command != 'status':
+                        continue
```
Rob-Johnson: can we ``raise NotImplementedError`` here rather than failing silently?

paasta_tools/paasta_serviceinit.py
```diff
@@ -143,57 +145,107 @@ def main():
...
+
+                    paasta_remote_run.remote_run_list(
```
Rob-Johnson: I think this is fine for now, but we might eventually need to create the equivalent ``remote_run_serviceinit`` to encapsulate all the behaviour. I guess YAGNI for now.

paasta_tools/paasta_serviceinit.py
```diff
@@ -143,57 +145,107 @@ def main():
...
+                        Dict(
```
Rob-Johnson: I'm actually kind of against spreading addict any further than it needs to reach.

paasta_tools/paasta_remote_run.py
keymone: the next ticket will be about looking for tasks in persistence layer, so we probably don't care too much about this detail. if we do change it to show finished tasks, people will start asking questions when they don't find what they're looking for because mesos gc.

paasta_tools/paasta_serviceinit.py
keymone: i didn't want to deal with extending serviceinit argslist at this point, but eventually - yes, that will be the case.

keymone:
@Rob-Johnson thanks, i've addressed your points (i think).

@solarkennedy @EvanKrall is this a 🚢 ?

paasta_tools/paasta_serviceinit.py
```diff
@@ -143,57 +144,109 @@ def main():
...
+                        soa_dir=args.soa_dir,
+                    )
+                elif instance_type == 'adhoc':
+                    if instance == 'interactive':
+                        raise NotImplementedError
+                    if command != 'status':
+                        raise NotImplementedError
+
+                    paasta_remote_run.remote_run_list(
+                        argparse.Namespace(
```
solarkennedy: Can you not use the argparser?

keymone: i'd like to reuse the `remote_run_list` but it takes argparse.Namespace as an argument and i don't want to change it's signature. could you elaborate why i shouldn't use argparse here?

solarkennedy: It is nice to call our functions consistently using normal keyword arguments. Plus it helps with static validation (vim, pylint, mocking, etc) to make sure we are sending it valid keyword arguments. And finally just general consistency in the codebase.

keymone: done

PR #1492, opened 2017-09-05
Instance configs experiments

oktopuz:
TLDR; An idea of a new way to work with yelpsoa-configs, so that to only read them once and to not use any caches.

Right now to iterate over instances of MarathonServiceConfig, we usually read yelpsoa-configs once  to get the service-instance pairs (get_services_for_cluster), and then we read the same files many times to create MarathonServiceConfig objects (load_marathon_service_config).
Moreover we implemented cache https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_tools.py#L160 to solve some performance issues caused by this approach.

The idea is to get rid of all the complexity by introducing a new class (PaastaServiceConfig ) which is capable of creating instance configs without re-reading yaml files.
I adjusted check_marathon_services_replication to show how it would look like after switching to PaastaServiceConfig.

### DIFF
--- a/docs/source/generated/paasta_tools.paasta_service_config.rst
+++ b/docs/source/generated/paasta_tools.paasta_service_config.rst
@@ -1,0 +1,7 @@
+paasta_tools.paasta_service_config module
+=========================================
+
+.. automodule:: paasta_tools.paasta_service_config
+    :members:
+    :undoc-members:
+    :show-inheritance:
--- a/docs/source/generated/paasta_tools.rst
+++ b/docs/source/generated/paasta_tools.rst
@@ -65,6 +65,7 @@ Submodules
    paasta_tools.paasta_metastatus
    paasta_tools.paasta_native_serviceinit
    paasta_tools.paasta_remote_run
+   paasta_tools.paasta_service_config
    paasta_tools.paasta_serviceinit
    paasta_tools.remote_git
    paasta_tools.secret_tools
--- a/paasta_tools/adhoc_tools.py
+++ b/paasta_tools/adhoc_tools.py
@@ -76,6 +76,16 @@ class AdhocJobConfig(LongRunningServiceConfig):
             soa_dir=soa_dir,
         )
 
+    def __repr__(self):
+        return "AdhocJobConfig(%r, %r, %r, %r, %r, %r)" % (
+            self.service,
+            self.cluster,
+            self.instance,
+            self.config_dict,
+            self.branch_dict,
+            self.soa_dir,
+        )
+
 
 def get_default_interactive_config(service, cluster, soa_dir, load_deployments=False):
     default_job_config = {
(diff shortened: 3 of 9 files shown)


oktopuz:
This is getting too big, thus I stop at this point and try get your approval for merging.

There are a couple of safe cosmetic changes in adhoc_tools.py, chronos_tools.py and utils.py.
paasta_service_config.py is a new file. The only thing uses and has changes a lot is check_marathon_services_replication.py

PaastaServiceConfig provides the following interfaces to work with yelpsoa-config:

```
>>> from paasta_tools.paasta_service_config import PaastaServiceConfig
>>> from paasta_tools.utils import DEFAULT_SOA_DIR
>>>
>>> sc = PaastaServiceConfig(service='fake_service', soa_dir=DEFAULT_SOA_DIR)
>>>
>>> for instance in sc.instances(cluster='fake_cluster', instance_type='marathon'):
...     print(instance)
...
main
canary
>>>
>>> for instance_config in sc.instance_configs(cluster='fake_cluster', instance_type='marathon'):
...     print(instance_config.get_instance())
...
main
canary
>>>
>>> instance_config = sc.instance_config(cluster='fake_cluster', instance='main')
>>> print(instance_config.get_instance())
...
main
>>>
```

solarkennedy:
In general I'm in favor of this, however you should not try to work on this while @EvanKrall is working on the sharded stuff, as it will be merge-conflict hell.

If we do this, however, I would like to take the opportunity to give it a better name, like `Noodle`, to make it more obvious about what this object *is*. config, instance, instance_config, they kinda suck :(

oktopuz:
How about:

- class PaastaServiceConfig -> class NoddleMaker
- PaastaServiceConfig.instances -> NoddleMaker.instance_names (returns: generator[str])
- PaastaServiceConfig.instance_configs -> NoddleMaker.noodles (returns: generator[Noodle])
- PaastaServiceConfig.instance_config -> NoddleMaker.noodle (returns: Noodle)

where Noodle is a sub-class of InstanceConfig.

solarkennedy:
@EvanKrall can you share your opinion?

oktopuz:
I changed class and method names and rebased this branch.
Ping @EvanKrall

oktopuz:
@solarkennedy if you are only concerned about names, they can be changed later using sed if someone has a better idea.
This is mostly new code. It shouldn't break anything. 
It could be easier to come up with an opinion when we start using it.

EvanKrall:
I always imagined "Noodle" to replace "Instance", not "InstanceConfig", though if we're actually changing it, I think "Job" is a better name still.

So `InstanceConfig` -> `JobConfig`
`--instance` -> `--job`
`service_instance` -> `service_job`


Or if we're going to go with Noodle, then:

`InstanceConfig` -> `NoodleConfig`
`--instance` -> `--noodle`
`service_instance` -> `service_noodle`
etc.

But I don't think `Noodle` should refer *only* to the configuration of an instance. If we're going to call instances instances, then we should call their configs `InstanceConfig`.

oktopuz:
I removed PaastaServiceConfig -> NoodleMaker transformation and I would prefer to leave renaming out of the scope of this PR and consider it only as a way to quickly iterate instance names and instance configs without re-reading yelpsoa-configs.

PR #1493, opened 2017-09-06
Rename the bounce_queue to pending_bounces

solarkennedy:
I'm going to try renaming things one thing at a time.

Does "pending_bounces" better describe what the bounce_q currently is?


nhandler:
I don't really think this is any more or less obvious. Maybe something more in line with the descriptions in our documentation like `bounce_queue` -> `pending_setup_marathon_job_queue`, and `inbox_queue` -> `needs_to_bounce_eventually_queue`

I would also suggest proposing new names for all of the queues first (even if you ship the changes one at a time).

PR #1494, opened 2017-09-07
PAASTA-12576: Optimize check_marathon_service runtime.

oktopuz:
* Remove unused 'client' from check_service_replication
* Pass cluster to marathon_tools.get_expected_instance_count_for_namespace() to avoid calling load_system_paasta_config() at https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_tools.py#L927
* Cache mesos_tools.get_slaves for 10 seconds. This reduces the number of calls to mesos-master from 105 (76 seconds) to 9 (11 seconds) on norcal-prod.

### DIFF
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -572,7 +572,6 @@ def get_mesos_slaves_grouped_by_attribute(slaves, attribute):
     }
 
 
-@time_cache(ttl=10)
 def get_slaves():
     return get_mesos_master().fetch("/master/slaves").json()['slaves']
 
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -572,6 +572,7 @@ def get_mesos_slaves_grouped_by_attribute(slaves, attribute):
     }
 
 
+@time_cache(ttl=10)
 def get_slaves():
     return get_mesos_master().fetch("/master/slaves").json()['slaves']
 
--- a/paasta_tools/smartstack_tools.py
+++ b/paasta_tools/smartstack_tools.py
@@ -119,18 +119,13 @@ def get_multiple_backends(services, synapse_host, synapse_port, synapse_haproxy_
     return backends
 
 
-def load_smartstack_info_for_service(
-    service, namespace, blacklist, system_paasta_config,
-    soa_dir=DEFAULT_SOA_DIR, mesos_slaves=None,
-):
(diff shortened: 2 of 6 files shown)


paasta_tools/mesos_tools.py
```diff
@@ -570,6 +570,7 @@ def get_mesos_slaves_grouped_by_attribute(slaves, attribute):
     }
 
 
+@time_cache(ttl=10)
```
Rob-Johnson: im not so keen on this, since it isn't really 'opt in'. can you change the caller instead?

oktopuz: To change the caller, I'll need to rewrite [mesos_tools.get_all_slaves_for_blacklist_whitelist](https://github.com/Yelp/paasta/blob/master/paasta_tools/mesos_tools.py#L538), and then [smartstack_tools.get_smartstack_replication_for_attribute](https://github.com/Yelp/paasta/blob/master/paasta_tools/smartstack_tools.py#L140) and [marathon_serviceinit.status_smartstack_backends](https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_serviceinit.py#L284) because they use get_all_slaves_for_blacklist_whitelist and so on.

I'm happy to do this, because I don't feel we should have a function that silently query mesos master every time we call it, but could you confirm please that we are on the same page regarding changing the caller?

oktopuz: This PR is not complete yet, but I've already changed what you asked.

oktopuz:
This PR is ready for review.
It takes 19s to run the check on the norcal-prod cluster.

paasta_tools/smartstack_tools.py
```diff
@@ -314,3 +336,92 @@ def match_backends_and_tasks(backends, tasks):
...
+
+        :param location: A string that identifies a habitat, a region and etc.
+        :param hosts: A list of mesos slave hostnames in this location.
+        :param instance_config: An instance of MarathonServiceConfig.
+        :returns: A dict {"service.instance": number_of_instances}.
+        """
+        full_name = compose_job_id(instance_config.service, instance_config.instance)
+        if location not in self._cache:
+            self._cache[location] = get_replication_for_all_services_from_synapse_haproxy(
+                synapse_host=hosts[0],
```
Rob-Johnson: why bother passing a list of hosts if you're only ever going to use one?

Issue #1495, opened 2017-09-07
Speed up cluster autoscaler


matthewbentley:
In the worst case, it can take 10 minutes for each slave it wants to kill.

There are three things that could make this faster:
1. Make draining faster/more effecient
2. Make cluster autoscaler terminate multiple slaves at a time
3. Make cluster autoscaler act on multiple resources at a time

2 is probably the easiest, while 1 would probably have the best effect (while also being the hardest).  3 is also a bit difficult, as right now the result of scaling up/down one resource can affect what the autoscaler does to the next.

2 will get it down to at most ~10 minutes per resource, and 2+3 ~10 minutes total.

matthewbentley:
AFAICT, the way it currently works is:

For each resource:
1. Calculate a "mesos_utilization_error", which is the diff between target capacity and capacity in the pool+region cross-section (this is independant of the resource)
2. Calculate a target capacity, which is the current size of the resource scaled by the utilization_error (with minimums and maximums depending on a few factors) (this is dependent on the resource type)
3. Run `scale_resource` with the current and target capacity, which will just scale up if it needs to, or run the logic to carefully scale down
4. If scaling down, sort the hosts by how killable they are, and then drain and kill one at a time

The calculate_mesos_utilization_error can be moved outside the resource, so it gets run once and then passed to each resource.  This will allow us to scale all the resources at once.

Also, the scaling down can be done in parallel (see #1496).

drmorr0:
The cluster autoscaler is deprecated in favor of Clusterman.

PR #1496, opened 2017-09-07
Async scale down cluster resource (ie parallel draining+killing of slaves in an sfr/asg)

matthewbentley:
This changes the cluster autosclaler so that within one resource, it will attempt to kill multiple slaves at once.

Basically, it creates a coroutine per slave to kill, and then runs that until it hits `await asyncio.sleep(5)`.  At which point the execution will be suspended, and the `downscale_aws_resource` will resume, going on to the next slave.  At the end of `downscale_aws_resource`, we wait until all coroutines have completed (or been canceled) before returning.

The logic in `downscale_aws_resource` is a bit iffy, but I'm confident in the proper interactions between having multiple `gracefully_terminate_slave`s and `wait_and_terminate`s running at at time.

This is not actually multithreaded, as only one coroutine will run at once, but there is a bit of difficulty around setting the capacity correctly (and then resetting in case of failure).

Addresses (but not totally) #1495

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -282,7 +282,7 @@ class ClusterAutoscaler(ResourceLogMixin):
                 else:
                     raise
 
-    def scale_resource(self, current_capacity, target_capacity, event_loop):
+    async def scale_resource(self, current_capacity, target_capacity):
         """Scales an AWS resource based on current and target capacity
         If scaling up we just set target capacity and let AWS take care of the rest
         If scaling down we pick the slaves we'd prefer to kill, put them in maintenance
@@ -312,12 +312,10 @@ class ClusterAutoscaler(ResourceLogMixin):
                     "Didn't find enough candidates to kill. This shouldn't happen so let's not kill anything!",
                 )
                 return
-            event_loop.run_until_complete(
-                self.downscale_aws_resource(
-                    filtered_slaves=filtered_slaves,
-                    current_capacity=current_capacity,
-                    target_capacity=target_capacity,
-                ),
+            await self.downscale_aws_resource(
+                filtered_slaves=filtered_slaves,
+                current_capacity=current_capacity,
+                target_capacity=target_capacity,
             )
 
     async def gracefully_terminate_slave(self, slave_to_kill, capacity_diff):
@@ -931,11 +929,21 @@ def autoscale_local_cluster(config_folder, dry_run=False, log_level=None):
         time.sleep(3)
     sorted_autoscaling_scalers = sorted(autoscaling_scalers, key=lambda x: x.is_resource_cancelled(), reverse=True)
     event_loop = asyncio.get_event_loop()
-    for scaler in sorted_autoscaling_scalers:
-        autoscale_cluster_resource(scaler, event_loop)
-        log.debug("Sleep 3s to throttle AWS API calls")
-        time.sleep(3)
+    event_loop.run_until_complete(run_parallel_scalers(sorted_autoscaling_scalers))
     event_loop.close()
+
+
+async def run_parallel_scalers(sorted_autoscaling_scalers):
+    scaling_tasks = []
+    for scaler in sorted_autoscaling_scalers:
+        scaling_tasks.append(asyncio.ensure_future(autoscale_cluster_resource(scaler)))
+        log.debug("Sleep 3s to throttle AWS API calls")
+        await asyncio.sleep(3)
+    for task in scaling_tasks:
+        if task.cancelled() or task.done():
+            continue
+        await task
+    await asyncio.sleep(0.1)
 
 
 def get_scaler(scaler_type):
@@ -946,12 +954,12 @@ def get_scaler(scaler_type):
     return scalers[scaler_type]
 
 
-def autoscale_cluster_resource(scaler, event_loop):
+async def autoscale_cluster_resource(scaler):
     log.info("Autoscaling {} in pool, {}".format(scaler.resource['id'], scaler.resource['pool']))
     try:
         current, target = scaler.metrics_provider()
         log.info("Target capacity: {}, Capacity current: {}".format(target, current))
-        scaler.scale_resource(current, target, event_loop)
+        await scaler.scale_resource(current, target)
     except ClusterAutoscalingError as e:
         log.error('%s: %s' % (scaler.resource['id'], e))
 
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -450,6 +450,7 @@ class ClusterAutoscaler(ResourceLogMixin):
 
     async def downscale_aws_resource(self, filtered_slaves, current_capacity, target_capacity):
         killed_slaves = 0
+        terminate_tasks = {}
         while True:
             filtered_sorted_slaves = ec2_fitness.sort_by_ec2_fitness(filtered_slaves)[::-1]
             if len(filtered_sorted_slaves) == 0:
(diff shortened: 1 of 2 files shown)


matthewbentley:
There still needs to be some limit on how fast this can terminate instances, because we don't want everything to hit the 10 minute timeout within a couple minute window and end up killing a bunch of apps before they've properly bounced.

matthewbentley:
Hopefully this isn't _too_ scary.  I think I fixed up the tests, so if anyone wants to look at it it's here.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -234,53 +255,54 @@ def wait_and_terminate(self, slave, drain_timeout, dry_run, region=None, should_
...
-                        self.log.info("Instance {}: NOT ready to kill".format(instance_id))
-                    self.log.debug("Waiting 5 seconds and then checking again")
-                    time.sleep(5)
+            while True:
+                instance_id = slave.instance_id
+                if not instance_id:
+                    self.log.warning(
+                        "Didn't find instance ID for slave: {}. Skipping terminating".format(slave.pid),
+                    )
+                    continue
```
matthewbentley: This `continue` looks like it might be wrong.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -234,53 +255,54 @@ def wait_and_terminate(self, slave, drain_timeout, dry_run, region=None, should_
...
-                ec2_client.terminate_instances(InstanceIds=instance_id, DryRun=dry_run)
+                ec2_client.terminate_instances(InstanceIds=[instance_id], DryRun=dry_run)
             except ClientError as e:
                 if e.response['Error'].get('Code') == 'DryRunOperation':
                     pass
                 else:
                     raise
 
-    def scale_resource(self, current_capacity, target_capacity):
+    async def scale_resource(self, current_capacity, target_capacity):
```
matthewbentley: This function needs an `await` call for the branches other than `delta < 0`

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -504,18 +537,36 @@ def downscale_aws_resource(self, filtered_slaves, current_capacity, target_capac
...
-                continue
-            except FailSetResourceCapacity:
-                break
+                    capacity_diff=capacity_diff,
+                    timer=timer,
+                ),
+            )
+            killed_slaves += 1
+
+            for hostname, task in terminate_tasks.items():
```
matthewbentley: I want to re-evaluate the logic here, especially w/r/t failed tasks

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -504,18 +537,36 @@ def downscale_aws_resource(self, filtered_slaves, current_capacity, target_capac
...
+                    done_tasks.add(hostname)
+                    task.result()  # Raises an exception if the task raised an exception
+                except HTTPError:
+                    # Something wrong draining host so try next host
+                    continue
+                except FailSetResourceCapacity:
+                    break
 
             current_capacity = new_capacity
             mesos_state = get_mesos_master().state_summary()
```
matthewbentley: I'm still not entirely sure what the next few lines do...

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -904,10 +984,22 @@ def autoscale_local_cluster(config_folder, dry_run=False, log_level=None):
...
+async def run_parallel_scalers(sorted_autoscaling_scalers):
+    scaling_tasks = []
     for scaler in sorted_autoscaling_scalers:
-        autoscale_cluster_resource(scaler)
+        scaling_tasks.append(asyncio.ensure_future(autoscale_cluster_resource(scaler)))
         log.debug("Sleep 3s to throttle AWS API calls")
-        time.sleep(3)
+        await asyncio.sleep(3)
+    for task in scaling_tasks:
+        if task.cancelled() or task.done():
```
matthewbentley: I need to double check on the behavior here if a task raises an exception.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -207,80 +230,82 @@ def get_mesos_utilization_error(
...
-                continue
-            free_percs.append(float(free) / float(total))
-        utilization = 1.0 - min(free_percs)
-        target_utilization = self.pool_settings.get('target_utilization', DEFAULT_TARGET_UTILIZATION)
-        return utilization - target_utilization
+    def can_kill(self, hostname, should_drain, dry_run, timer):
+        if dry_run:
+            return True
+        if timer.ready():
+            self.log.warning("Timer expired before slave ready to kill...")
```
Rob-Johnson: can you hint at what this really means? is it a problem?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -504,21 +541,26 @@ def downscale_aws_resource(self, filtered_slaves, current_capacity, target_capac
...
+                    timer=timer,
+                ),
+            )
+            killed_slaves += 1
 
             current_capacity = new_capacity
             mesos_state = get_mesos_master().state_summary()
+            # It's not obvious what this loop is doing: it is updating the task counts for
+            #  for the slaves in filtered_sorted_slaves, so next time through the loop
+            #  we can re-sort
```
solarkennedy: Just delete this. It looks like it is leftover from when it used to be the fitness stuff

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -73,16 +91,26 @@ def log(self):
...
         self.pool_settings = pool_settings
         self.config_folder = config_folder
         self.dry_run = dry_run
+        self.utilization_error = utilization_error
         self.draining_enabled = draining_enabled
         if log_level is not None:
             self.log.setLevel(log_level)
 
-    def set_capacity(self, capacity):
+    def set_capacity(self, capacity, unrounded_capacity):
```
solarkennedy: Would much of this code be simpler if this *only* took and then relied on the implementation to do the rounding at the last minute?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -294,10 +319,12 @@ def scale_resource(self, current_capacity, target_capacity):
         delta = target_capacity - current_capacity
         if delta == 0:
             self.log.info("Already at target capacity: {}".format(target_capacity))
+            await asyncio.sleep(1)
             return
         elif delta > 0:
             self.log.info("Increasing resource capacity to: {}".format(target_capacity))
-            self.set_capacity(target_capacity)
+            self.set_capacity(target_capacity, target_capacity)
```
solarkennedy: very strange. If we have to do 2 args for this, can you use kwargs?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
solarkennedy: Why?

If it really does need that, then can you say so in the docstring?

matthewbentley: Python will throw a `RuntimeWarning` if an `async` function doesn't `await` on anything.  I was running into issues with this on testing, but it's probably worth adding the proper flag to not raise the RuntimeWarnings as Errors (and then I can remove all the extra `asyncio.sleep(0.1)`).

paasta_tools/autoscaling/autoscaling_cluster_lib.py
matthewbentley: Yep as mentioned on #1560 this can be removed in this PR

PR #1505, opened 2017-09-12
Fix type error, where whitelist gets checked against none in mesos_tools

matthewbentley:
Also fix type error when logging errors

### DIFF
--- a/paasta_tools/smartstack_tools.py
+++ b/paasta_tools/smartstack_tools.py
@@ -168,7 +168,7 @@ def get_smartstack_replication_for_attribute(attribute, service, namespace, blac
     replication_info = {}
     filtered_slaves = mesos_tools.get_all_slaves_for_blacklist_whitelist(
         blacklist=blacklist,
-        whitelist=[],
+        whitelist=None,
     )
     if not filtered_slaves:
         raise mesos_tools.NoSlavesAvailableError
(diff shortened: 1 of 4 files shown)


paasta_tools/smartstack_tools.py
```diff
@@ -168,7 +168,7 @@ def get_smartstack_replication_for_attribute(attribute, service, namespace, blac
     replication_info = {}
     filtered_slaves = mesos_tools.get_all_slaves_for_blacklist_whitelist(
         blacklist=blacklist,
-        whitelist=[],
+        whitelist=None,
```
Rob-Johnson: isn't it the code on the other side thats wrong? i.e. why aren't we passing [] here and the ``get_all_slaves_for_blacklist_whitelist`` function dealing with it correctly?

matthewbentley: `DeployWhitelist` is an `Optional`, so `None` is fine: https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L151

matthewbentley: actually, I think the type in utils.py is wrong

matthewbentley: ...or the type is right, but the docstring on `slave_passes_whitelist` is wrong

PR #1506, opened 2017-09-12
Chunk calls to aws describe instances to 199 at a time

### LINKED ISSUE #1501: Cluster autoscaler fails on resource with more than 200 instances
It tries to call describe aws resources with the list of IPs, which fails if the list is longer than allowed by the AWS API

matthewbentley:
describe instances has a limit of 200 at a time from aws, so we should call it with 199 ips at a time.

Fixes #1501

### DIFF
--- a/tests/autoscaling/test_autoscaling_cluster_lib.py
+++ b/tests/autoscaling/test_autoscaling_cluster_lib.py
@@ -1460,6 +1460,21 @@ class TestClusterAutoscaler(unittest.TestCase):
         ret = self.autoscaler.get_pool_slaves(mock_mesos_state)
         assert ret == {'id1': mock_mesos_state['slaves'][0]}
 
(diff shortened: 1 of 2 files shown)


tests/autoscaling/test_autoscaling_cluster_lib.py
```diff
@@ -1460,6 +1460,21 @@ def test_get_pool_slaves(self):
...
+            autospec=True,
+        ) as mock_describe_instances:
+            ips = list(range(567))
+
+            def mock_describe_instance(self, instance_ids, region, instance_filters):
+                return instance_filters[0]['Values']
+            mock_describe_instances.side_effect = mock_describe_instance
+
+            ret = self.autoscaler.instance_descriptions_for_ips(ips)
+            assert len(ret) == 567
```
mattmb: might be worth asserting `mock_describe_instances.call_count` too? but that's a nit-pick really.

PR #1508, opened 2017-09-12
Don't try to drain unhealthy instances when scaling the cluster down

matthewbentley:
Hopefully this will avoid some mesos bugs

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -359,7 +359,11 @@ class ClusterAutoscaler(ResourceLogMixin):
         self.log.info("Waiting for instance to drain before we terminate")
         try:
             self.wait_and_terminate(
-                slave_to_kill, drain_timeout, self.dry_run, region=self.resource['region'], should_drain=should_drain,
+                slave=slave_to_kill,
+                drain_timeout=drain_timeout,
+                dry_run=self.dry_run,
+                region=self.resource['region'],
+                should_drain=should_drain,
             )
         except ClientError as e:
             self.log.error("Failure when terminating: {}: {}".format(slave_to_kill.pid, e))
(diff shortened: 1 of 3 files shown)


mattmb:
I've just added a commit for PAASTA-12757.

Now I think though, maybe we should add a sleep step if draining is disabled?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -338,18 +353,20 @@ def gracefully_terminate_slave(self, slave_to_kill, current_capacity, new_capaci
...
             self.log.info("Undraining {}".format(slave_to_kill.pid))
-            if not self.dry_run:
+            if should_drain:
                 undrain([drain_host_string])
             raise
         self.log.info("Waiting for instance to drain before we terminate")
         try:
-            self.wait_and_terminate(slave_to_kill, drain_timeout, self.dry_run, region=self.resource['region'])
+            self.wait_and_terminate(
+                slave_to_kill, drain_timeout, self.dry_run, region=self.resource['region'], should_drain=should_drain,
```
Rob-Johnson: can you kwargs all of these please

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -245,7 +246,9 @@ def wait_and_terminate(self, slave, drain_timeout, dry_run, region=None):
                         )
                         continue
                     # Check if no tasks are running or we have reached the maintenance window
-                    if is_safe_to_kill(slave.hostname) or dry_run:
+                    if is_safe_to_kill(slave.hostname) or not should_drain:
```
mattmb: I think we want to avoid the `is_safe_to_kill` call completely if `should_drain` is `False`. That way we won't make any calls to the mesos maintenance endpoints at all.

PR #1513, opened 2017-09-19
Consider tasks from all versions of a service when autoscaling. PAASTA-12983

EvanKrall:


tests/test_autoscaling_service_lib.py
```diff
@@ -0,0 +1,48 @@
+import mock
```
solarkennedy: can you put this in tests/autoscaling/test_autoscaling_service_lib.py instead of starting a new file?

Or move the tests from that file into here? would like to keep the 1:1 mapping of python code and test code files to make it easier to tdd and discover

EvanKrall: oh, derp, I didn't see that file.

solarkennedy:
fix + ship

EvanKrall:
d'oh, the tests are failing because my `get_app` lambda never gets called (short circuit evaluation in `filter_autoscaling_tasks` means if `is_task_healthy` returns true, `marathon_client.get_app` and `is_old_task_missing_healthchecks` never get called.), and we have coverage set to fail if you have any unexercised code paths in tests/.

PR #1514, opened 2017-09-19
in local-run do not try to bind to a port if healthcheck is not defined

somic:
PAASTA-12578


somic:
In get_docker_run_cmd() method, we will not add --publish option to cmd if container_port is None. This logic drives this PR where we set container_port to None if healthcheck is not defined.

PR #1516, opened 2017-09-21
Make cleanup maintenance pass if unreserving fails

mattmb:
Currently we giveup undraining if we can't unreserve resources. I think
maybe this can happen for old hosts that have gone away or in other
circumstances. Either way we try again to unreserve later so it is not
essential that we fail early in this case. The downside of failing here
is the cleanup maint script may never finish.

### DIFF
--- a/paasta_tools/mesos_maintenance.py
+++ b/paasta_tools/mesos_maintenance.py
@@ -582,7 +582,10 @@ def drain(hostnames, start, duration):
     :returns: None
     """
     log.info("Draining: %s" % hostnames)
-    reserve_all_resources(hostnames)
+    try:
+        reserve_all_resources(hostnames)
+    except HTTPError as e:
+        log.warning("Failed to reserve resources, will continue to drain: %s" % e)
     payload = build_maintenance_schedule_payload(hostnames, start, duration, drain=True)
     client_fn = get_schedule_client()
     try:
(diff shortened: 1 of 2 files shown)


matthewbentley:
I think the cluster autoscaler will not try to kill a host if reserving the resources fails; so I'm not sure catching that error when reserving is the right thing to do here?

mattmb:
It does, you are right. I guess I'm not 100% sure if that's a bad thing or not...

The reserving is kind of best effort anyway since there's still a race condition that can mean a new task can end up on a draining host.

paasta_tools/mesos_maintenance.py
```diff
@@ -599,7 +602,10 @@ def undrain(hostnames):
     :returns: None
     """
     log.info("Undraining: %s" % hostnames)
-    unreserve_all_resources(hostnames)
+    try:
+        unreserve_all_resources(hostnames)
```
nhandler: This is the one potentially risky part. If we fail to unreserve the resources, but mark the host as being up, we will be paying for resources that can't be used. It probably makes sense to also update the maintenance cleanup scripts to check for dynamic resource reservations made for the purpose of maintenance that exist on boxes marked as being 'up' and then remove them.

mattmb: I think it already does right? i.e. when it gets to: https://github.com/Yelp/paasta/blob/master/paasta_tools/cleanup_maintenance.py#L103
the cleanup script should spot anything where we managed to undrain it but where we haven't unreserved the resources.

PR #1523, opened 2017-09-25
Added in config loading and client generation for sharded marathon configurations

solarkennedy:
I'm adding both soa-config knobs for choosing a shard and listing previous shards, as well as system functions to read marathon servers and previous marathon servers, as well as a client loader to get all those clients en-masse.

I didn't want to write tests or docs so early if this is not the approach we want. Let me know if you think this is the right approach to get us off on the right foot, idk maybe write a test exercising get_marathon_clients ? And docs.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -21,6 +21,7 @@ import datetime
 import json
 import logging
 import os
+from collections import namedtuple
 from math import ceil
 
 import requests
@@ -68,9 +69,19 @@ CONFIG_HASH_BLACKLIST = {'instances', 'backoff_seconds', 'min_instances', 'max_i
 log = logging.getLogger(__name__)
 logging.getLogger('marathon').setLevel(logging.WARNING)
 
+MarathonServers = namedtuple('MarathonServers', ['current', 'previous'])
+MarathonClients = namedtuple('MarathonClients', ['current', 'previous'])
+
 
 def load_marathon_config():
     return MarathonConfig(load_system_paasta_config().get_marathon_config())
+
+
+def get_marathon_servers():
+    system_config = load_system_paasta_config()
+    current = [MarathonConfig(x) for x in system_config.get_marathon_servers()]
+    previous = [MarathonConfig(x) for x in system_config.get_previous_marathon_servers()]
+    return MarathonServers(current=current, previous=previous)
 
 
 class MarathonNotConfigured(Exception):
@@ -536,9 +547,19 @@ class MarathonServiceConfig(LongRunningServiceConfig):
         '''Map this port on the host to your container's port 8888. Default is 0, which means Marathon picks a port.'''
         return self.config_dict.get('host_port', 0)
 
+    def get_marathon_shard(self):
+        """Returns the configued shard of Marathon to use.
+        Defaults to 0 currently as a fail-safe"""
+        return self.config_dict.get('marathon_shard', 0)
+
+    def get_previous_marathon_shards(self):
+        """Returns a list of Marathon shards a service might have been on previously.
+        Useful for graceful shard migrations. Defaults to an empty list"""
+        return self.config_dict.get('previous_marathon_shards', [])
+
 
 class MarathonDeployStatus:
-    """ An enum to represent marathon app deploy status.
+    """ An enum to represent Marathon app deploy status.
     Changing name of the keys will affect both the paasta CLI and API.
     """
     Running, Deploying, Stopped, Delayed, Waiting, NotRunning = range(0, 6)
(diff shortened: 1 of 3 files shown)


paasta_tools/utils.py
```diff
@@ -1315,6 +1315,8 @@ def get_readable_files_in_glob(glob: str, path: str) -> List[str]:
         'cluster_fqdn_format': str,
         'chronos_config': ChronosConfig,
         'marathon_config': MarathonConfigDict,
+        'marathon_servers': List[MarathonConfigDict],
+        'previous_marathon_servers': List[MarathonConfigDict],
```
EvanKrall: Might as well use MarathonServers here since you've defined it.

I think I probably would've just used List[MarathonConfigDict] everywhere though.

paasta_tools/marathon_tools.py
```diff
@@ -68,11 +69,21 @@
 log = logging.getLogger(__name__)
 logging.getLogger('marathon').setLevel(logging.WARNING)
 
+MarathonServers = namedtuple('MarathonServers', ['current', 'previous'])
+MarathonClients = namedtuple('MarathonClients', ['current', 'previous'])
```
EvanKrall: namedtuple is convenient for now, but actually I think we're going to want to define a class for this, since we'll want to define methods like `get_current_shard()` and `get_previous_shards()`

solarkennedy: but aren't those functions of a an instance?
I've already defined those functions in the MarathonInstanceConfig thingy

EvanKrall: Hm, yeah, that does seem like a better place to put that code.

paasta_tools/marathon_tools.py
```diff
@@ -605,6 +625,28 @@ def get_marathon_client(url, user, passwd, cached=False):
         return MarathonClient(url, user, passwd, timeout=30, session=session)
 
 
+def get_marathon_clients(MarathonServers, cached=False):
```
Rob-Johnson: ``marathon_servers``

paasta_tools/marathon_tools.py
```diff
@@ -536,9 +547,19 @@ def get_host_port(self):
         '''Map this port on the host to your container's port 8888. Default is 0, which means Marathon picks a port.'''
         return self.config_dict.get('host_port', 0)
 
+    def get_marathon_shard(self):
+        """Returns the configued shard of Marathon to use.
+        Defaults to 0 currently as a fail-safe"""
+        return self.config_dict.get('marathon_shard', 0)
```
Rob-Johnson: can we rename this ``preferred_marathon_shard`` ? are we going to block deploying this if their shard isn't available?

paasta_tools/marathon_tools.py
```diff
@@ -536,9 +547,19 @@ def get_host_port(self):
...
+
+    def get_previous_marathon_shards(self):
```
Rob-Johnson: I don't think this coupling is the right one - the shard isn't a part of the app - rather there should be another datastructure that encodes the pairing of a shard an an app owned by whatever is in charge of tying those things together,

paasta_tools/marathon_tools.py
solarkennedy: I can do that. Yea I think we kinda have to block if the destination shard is not available, otherwise we would be leaving old tasks behind. Also I don't want to implement hinted handoff.

paasta_tools/marathon_tools.py
solarkennedy: But it sure is convenient if we put the desired shard and the function that calculates the auto shard in here, as it can call self.get_marathon_shard() and the function for calculating the shard can inspect that and default to calculating based on self.service and self.instance. 

I'm not saying they could be separate functions not in this class, but we would have to inject service, instance into them in order to calculate the shard.

In other words, method 1:
```
shard = instanceconfig.get_marathon_shard(available_shards=shards)
app = marathon_tools.get_marathon_app(client=shards[shard], id=instanceconfig.generate_app_id()
```

Method 2:
```
prefered_shard = instanceconfig.get_preferred_marathon_shard()
if preferred_shard == None:
    prefrerred_shard = MarathonShardsThingy.calculate_shard_for_service(instanceconfig.service, instanceconfig.instance, available_shards=shards)
app = marathon_tools.get_marathon_app(client=shards[preferred_shard], id=instanceconfig.generate_app_id()
```

It isn't "that" much different. I would argue that the shard *is* kinda part of the app, because it is as much of a function of its name and config as it is a function of the shard topology.

Rob-Johnson: You're right that they dont appear that different, but I think that the presence of ``MarathonShardsThingy`` is a good one - we should be leaning towards the instance_config *just* being an expression of the data - not an expression of the data + a bunch of logic for calculating that data. It may seem small, but it significantly changes the overhead of 'what' instanceconfig is.

Saying that, the pattern is kind of established for instanceconfig doing this anyway (calculating the right value for its internal state, rather than something else being responsible for that,). This is just kind of kerbing that pattern where we can.

PR #1524, opened 2017-09-26
Apollo 205 detector status 36b

philipmulcahy:
Add new detector-status verb to paasta tools:
paasta-tools detector-status --detector_name='Business similarity search apollo client timings warning'
Enables CI tooling to call out to signalfx via paasta-tools after a new version is pushed to stage, canary or prod.
Integration tests have been intermittently failing for this change, but are passing at the moment.


solarkennedy:
@philipmulcahy I appreciate this work and am in favor of this kind of thing.

However, I think the paasta codebase is going to be a blocker for you. It is open source and in general stays away from "yelpy" things, and I think for this to be really good it is going to need to be better integrated with more yelp-specific stuff.

At the same time, there is the (small) need to import paasta tools so you can get the monitoring stuff.

Can you take a look at at internal ticket https://jira.yelpcorp.com/browse/PAASTA-12702. If that existed (I will volunteer to create it) I think it would be a better place for this sort of tool.

oktopuz:
I moved detector_status.py to yelp_paaasta_helpers.
paasta_tools/utils.py is the only thing that needs to be merged.

PR #1528, opened 2017-09-26
MesosMaster.orphan_tasks should return a list of Task objects

somic:
PAASTA-11595

I am relatively confident this or something like this in spirit is required to fix this ticket. It looks like orphan_tasks() method returns data straight out of state (which is a giant dict), instead of converting it to a list of Task objects.

This probably needs tests? And ideally I'd like to observe this in the wild to ensure data structures in mesos api look like what we expect? Mesos docs are lacking - http://mesos.apache.org/documentation/latest/endpoints/master/state/

cc @bobtfish as fyi

This is probably not yet ready to be merged.

### DIFF
--- a/paasta_tools/mesos/master.py
+++ b/paasta_tools/mesos/master.py
@@ -198,7 +198,11 @@ class MesosMaster(object):
         return lst[0]
 
     def orphan_tasks(self):
-        return self.state["orphan_tasks"]
+        return [
+            task.Task(self, x)
+            for x in self._task_list(False)
+            if x['id'] in self.state["orphan_tasks"]
+        ]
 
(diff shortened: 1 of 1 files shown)


somic:
In order to be able to verify correctness of this change, I am currently blocked on being able to obtain a sample output of /master/state api that has non-empty orphan_tasks.

Alternatively, to be able to merge this sooner, we could temporarily make orphan_tasks always return [] or not consider orphan_tasks in get_all_running_tasks()

somic:
Closing this for now, it will eventually be reopened. Related to https://github.com/Yelp/paasta/pull/1537

PR #1534, opened 2017-10-02
Added sharded marathon servers to the paasta_itests and example_cluster

solarkennedy:
This is not a shard-only config, this is still in hybrid mode, like our real infrastructure.

However, it allows itest-specific room so you could make assertions about an individual shard, like:
```
  marathon shard N should have appid foo
```
with the raw clients available in the paasta_itests context with this PR.

### DIFF
--- a/paasta_itests/steps/setup_steps.py
+++ b/paasta_itests/steps/setup_steps.py
@@ -30,8 +30,8 @@ from paasta_tools.utils import decompose_job_id
 from paasta_tools.utils import paasta_print
 
 
-def _get_marathon_connection_string():
-    return 'http://%s' % get_service_connection_string('marathon')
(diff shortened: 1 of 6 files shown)


paasta_itests/steps/setup_steps.py
```diff
@@ -59,6 +59,23 @@ def setup_system_paasta_config():
                 'user': None,
                 'password': None,
             },
+            'marathon_servers': [
+                {
+                    'url': _get_marathon_connection_string('marathon'),
```
EvanKrall: I guess I should probably update #1551 to allow a single string in `url`, rather than just a `List[str]`, since MarathonClient allows either, and we generally just pass `url` through unmodified.

PR #1537, opened 2017-10-03
temporarily ignore orphan_tasks in mesos_tools.get_all_running_tasks

somic:
PAASTA-11595

This is a temporary fix that will be eventually undone as part of https://github.com/Yelp/paasta/pull/1528.


paasta_tools/mesos_tools.py
```diff
@@ -154,8 +154,6 @@ def get_all_running_tasks():
     """ Will include all running tasks, including orphans
```
solarkennedy: I guess update the docstring?

PR #1538, opened 2017-10-03
local-run will detach only if healthcheck is needed

somic:
PAASTA-13120


solarkennedy:
Seems suspiciously simple!

PR #1541, opened 2017-10-04
Paasta 13123 ignore remote run frameworks in metastatus

oktopuz:
internal ticket: PAASTA-13123

before:
```
Mesos Status: CRITICAL
  Quorum: masters: 3 configured quorum: 2
  Frameworks:
    Framework: chronos count: 1
    Framework: marathon count: 1
  CPUs: 1.61 / 309 in use (0.52%)
  Memory: 8.63 / 548.10GB in use (1.58%)
  Disk: 16.00 / 3064.37GB in use (0.52%)
  No gpus found from mesos!
  Tasks: running: 17 staging: 0 starting: 0
  Slaves: active: 15 inactive: 0
  Connected Frameworks: expected: 2 actual: 4
  Disconnected Frameworks: expected: 0 actual: 0
  Inactive Frameworks: expected: 0 actual: 0
```

after:
```
Mesos Status: OK
  Quorum: masters: 3 configured quorum: 2
  Frameworks (configured 4, connected 4):
    Framework: chronos count: 1
    Framework: marathon count: 3
  CPUs: 1.61 / 309 in use (0.52%)
  Memory: 8.63 / 548.10GB in use (1.58%)
  Disk: 16.00 / 3064.37GB in use (0.52%)
  No gpus found from mesos!
  Tasks: running: 17 staging: 0 starting: 0
  Slaves: active: 15 inactive: 0
```

Also I added system_paasta_config as a mandatory parameter to get_marathon_servers().
load_system_paasta_config is a heavy operation. It reads, parses and merges many json files.
We should read it once and pass everywhere we need this information.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -77,10 +77,13 @@ def load_marathon_config():
     return MarathonConfig(load_system_paasta_config().get_marathon_config())
 
 
-def get_marathon_servers():
-    system_config = load_system_paasta_config()
-    current = [MarathonConfig(x) for x in system_config.get_marathon_servers()]
-    previous = [MarathonConfig(x) for x in system_config.get_previous_marathon_servers()]
+def get_marathon_servers(system_paasta_config):
+    """
+    :param system_paasta_config: A SystemPaastaConfig object representing the system
+                                 configuration.
+    """
+    current = [MarathonConfig(x) for x in system_paasta_config.get_marathon_servers()]
+    previous = [MarathonConfig(x) for x in system_paasta_config.get_previous_marathon_servers()]
     return MarathonServers(current=current, previous=previous)
 
 
@@ -92,6 +95,18 @@ class MarathonConfig(dict):
 
     def __init__(self, config):
         super(MarathonConfig, self).__init__(config)
+
+    @property
+    def url(self):
+        return self.get_url()
+
+    @property
+    def user(self):
+        return self.get_username()
+
+    @property
+    def passwd(self):
+        return self.get_password()
 
     def get_url(self):
         """Get the Marathon API url
--- a/paasta_tools/metrics/metastatus_lib.py
+++ b/paasta_tools/metrics/metastatus_lib.py
@@ -17,7 +17,6 @@ import itertools
 import math
 from collections import Counter
 from collections import namedtuple
-from collections import OrderedDict
 
 from humanize import naturalsize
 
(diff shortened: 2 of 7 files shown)


paasta_tools/metrics/metastatus_lib.py
```diff
@@ -259,37 +258,45 @@ def assert_tasks_running(metrics):
...
-def assert_no_duplicate_frameworks(state, frameworks_list):
-    """A function which asserts that there are no duplicate frameworks running, where
-    frameworks are identified by their name.
-
-    Note the extra spaces in the output strings: this is to account for the extra indentation
-    we add, so we can have:
-
-        frameworks:
-          framework: marathon count: 1
+def assert_framework_count(state, configured_framework_count):
```
Rob-Johnson: this seems more complicated than it needs to be. why're we making one 'mega' check rather than the function per check that we had before, allowing the caller to filter the frameworks checked like we already did in ``assert_no_duplicate_frameworks`` ? I think you could've just updated the counting code that was on line 227 to match whether the framework name startswith the same value in ``frameworks_list`` rather than the exact match?

paasta_tools/marathon_tools.py
```diff
@@ -93,6 +96,18 @@ class MarathonConfig(dict):
     def __init__(self, config):
         super(MarathonConfig, self).__init__(config)
 
+    @property
```
solarkennedy: yay properties!

paasta_tools/metrics/metastatus_lib.py
oktopuz: @Rob-Johnson, this might look like I'm making a mega-check, but I'm not in fact.
I'm just removing several unused or broken assertions.

assert_active_frameworks() - is not used at all.

assert_connected_frameworks, assert_disconnected_frameworks, assert_inactive_frameworks are called by get_framework_metrics_status which is, in its turn, called by paasta_metastatus.
There 3 assertions relies on mesos counters that only provide number, but don't specify which framework is connected, inactive or disconnected.

```
  "master/frameworks_active": 3,
  "master/frameworks_connected": 3,
  "master/frameworks_inactive": 1,
  "master/frameworks_disconnected": 1,
```

It was possible to check these number before we started using task processing frameworks because we knew how many frameworks we were supposed to have connected. But these days we might have 1 disconnected marathon framework and 1 connected task proc framework and assert_connected_frameworks wouldn't notice any problem. These numbers are not longer useful in my opinion.

As for frameworks_list we have in assert_no_duplicate_frameworks, it is hardcoded at [metastatus_lib.py](https://github.com/Yelp/paasta/blob/master/paasta_tools/metrics/metastatus_lib.py#L584).
The only place where the caller can actually set this value is [check_mesos_duplicate_frameworks.py](https://github.com/Yelp/paasta/blob/master/paasta_tools/monitoring/check_mesos_duplicate_frameworks.py#L44).
In other words if we need to add a new framework name to the list, we'll have to update it in puppet and update it one more time in paasta_tools.

I think I disagree that this is actually better than hardcoding it in the function where is it used indeed so that we need to update it in paasta_tools only.

It would be great if we had framework names in marathon.json. In this case we wouldn't need to maintain frameworks_list.

paasta_tools/marathon_tools.py
oktopuz: [get_marathon_clients](https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_tools.py#L628) doesn't work without them.

oktopuz:
@Rob-Johnson, please take a look at https://github.com/Yelp/paasta/pull/1550 as well.
Maybe you like that solution more than this one. I do.

oktopuz:
I'm closing this in favor of https://github.com/Yelp/paasta/pull/1550

PR #1543, opened 2017-10-04
Work around wait-for-deployment error

oktopuz:
swagger expects 'commad' to be a string, while get_cmd can return None.

https://github.com/Yelp/paasta/blob/master/paasta_tools/api/api_docs/swagger.json#L480


solarkennedy:
Can you make wait-for-deployment not inspect chronos things at all?

PR #1549, opened 2017-10-06
Use a uuid when doing local-run to give them unique ids

solarkennedy:


paasta_tools/cli/cmds/local_run.py
```diff
@@ -475,11 +476,12 @@ def get_local_run_environment_vars(instance_config, port0, framework):
         # In a local_run environment, the docker_image may not be available
         # so we can fall-back to the injected DOCKER_TAG per the paasta contract
         docker_image = os.environ['DOCKER_TAG']
+    fake_taskid = uuid.uuid1()
```
chriskuehl: can you use uuid4 instead? uuid1 is weird and depends on the computer's MAC address

uuid4 is just pure random (besides a few bits to indicate the uuid version). it's also faster, not that it matters so much here.

solarkennedy: Ah!

PR #1551, opened 2017-10-10
Add type annotations to marathon-related things

EvanKrall:
This adds type annotations to long_running_service_tools, marathon_tools, setup_marathon_job, and related tests. I did this work to help make sure the sharding change is correct.

#1552 has the actual meat of the sharding change -- this PR is separate so that #1552's diff is cleaner.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -23,22 +23,34 @@ import logging
 import os
 from collections import namedtuple
 from math import ceil
+from typing import Any
+from typing import Dict
+from typing import List
+from typing import Optional
+from typing import Tuple
 
 import requests
 import service_configuration_lib
 from marathon import MarathonClient
 from marathon import MarathonHttpError
 from marathon import NotFoundError
+from marathon.models.app import MarathonApp
+from marathon.models.app import MarathonTask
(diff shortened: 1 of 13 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -217,7 +352,7 @@ def __repr__(self):
             self.soa_dir,
         )
 
-    def copy(self):
+    def copy(self) -> "MarathonServiceConfig":
```
solarkennedy: Is `MarathonServiceConfig` supposed to be quoted here?

paasta_tools/setup_marathon_job.py
```diff
@@ -702,15 +757,20 @@ def main():
...
+    marathon_apps: Collection[marathon_tools.MarathonApp],
+) -> Tuple[int, float]:
     """deploy the service instance given and proccess return code
     if there was an error we send a sensu alert.
 
     :param service: The service name to setup
     :param instance: The instance of the service to setup
     :param client: A MarathonClient object
     :param soa_dir: Path to yelpsoa configs
-    :param marathon_config: The service instance's configuration dict
```
solarkennedy: wow, I guess we really were not using this?

PR #1552, opened 2017-10-10
Rendezvous hashing and sharded setup_marathon_job, cleanup_marathon_job, marathon_serviceinit, list_marathon_service_instances

EvanKrall:
This implements rendezvous hashing in MarathonClients, and then modifies setup_marathon_job to use MarathonClients instead of MarathonClient.

The basic idea is that we fetch marathon apps from all the relevant clients early in setup_marathon_job, and then pass around tuples of (app, client), (task, client), or (app_id, client).

This also changes the bounce methods' signature, so they take sets of task objects, rather than dicts of app name -> sets of tasks, since all the bounce methods immediately consolidate all the tasks into a single list/set anyway.

This is WIP as it breaks deployd, but I expect the changes to fix that will be in the deployd code/tests, so it will only add on to this patch, rather than replacing anything here.


EvanKrall:
Note that I've set the base of this PR to be #1551 to clear up some of the type annotation clutter from the diff.

EvanKrall:
I also have yet to incorporate #1534 into this branch.

paasta_tools/marathon_tools.py
```diff
@@ -89,7 +93,75 @@
 logging.getLogger('marathon').setLevel(logging.WARNING)
 
 MarathonServers = namedtuple('MarathonServers', ['current', 'previous'])
```
EvanKrall: Add type annotations for MarathonServer so mypy knows these are MarathonConfig objects.

paasta_tools/marathon_tools.py
```diff
@@ -89,7 +93,75 @@
...
+
+
+_RendezvousHashT = TypeVar('_RendezvousHashT')
+
+
+def rendezvous_hash(
+    choices: Sequence[_RendezvousHashT],
+    key: str,
+    salt: str='',
+    hash_func: Callable[[str], str]=get_config_hash,
```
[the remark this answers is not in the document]
EvanKrall: it's imported from utils.

paasta_tools/marathon_tools.py
```diff
@@ -89,7 +93,75 @@
...
+) -> _RendezvousHashT:
+    """
+
+    :param choices: A dictionary of string to arbitrary values. The string keys are used in the hash calculation, and
+                    the value with the "winning" key is returned."""
+    max_hash_value = None
+    max_hash_choice = None
+    for i, choice in enumerate(choices):
```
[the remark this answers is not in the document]
EvanKrall: Oh, oops, I was originally going to take a dictionary, but then changed to just a list of things, but forgot to update the docstring.

paasta_tools/marathon_tools.py
```diff
@@ -1182,16 +1254,28 @@ def get_matching_appids(servicename: str, instance: str, client: MarathonClient)
     Useful for fuzzy matching if you think there are marathon
     apps running but you don't know the full instance id"""
     marathon_apps = get_all_marathon_apps(client)
-    return [app.id for app in get_matching_apps(servicename, instance, marathon_apps)]
+    return [app.id for app in marathon_apps if does_app_id_match(servicename, instance, app.id)]
```
solarkennedy: In here and the next 20+ lines it references "servicename". A while ago I changed all the serivce/servicename/service_name/instance/instancename, etc to just be service/instance, can you make these consistent?

solarkennedy:
Is travis not running?

paasta_tools/setup_marathon_job.py
```diff
@@ -494,15 +518,29 @@ def log_deploy_error(errormsg: str, level: str='event') -> None:
...
-    other_apps = [a for a in existing_apps if a.id != '/%s' % config['id']]
+    existing_apps_with_clients = marathon_tools.get_matching_apps_with_clients(
+        service,
+        instance,
+        marathon_apps_with_clients,
+    )
+
+    new_client = clients.get_current_client_for_service(job_config)
+
+    new_apps_with_clients_list: List[Tuple[MarathonApp, MarathonClient]] = []
```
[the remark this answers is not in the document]
EvanKrall: Technically speaking, an app doesn't bounce -- it gets created, scaled up, scaled down, stopped. A service/instance bounces.

Maybe think of it as new == desired, old == undesired. Most of the time new/old is correct, like when bouncing because a new version was marked for deployment, but sometimes new isn't accurate, like when rolling back.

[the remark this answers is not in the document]
mattmb: Agreed, naming is hard! I like desired/undesired. If you wanted to rename them to that I think it'll be easier to understand for the next person reading this section.

solarkennedy:
I would be OK if you shipped deployd in a second PR

EvanKrall:
The deployd code is already in the branch, and is not too bad.

This code is not ready to ship yet -- there are no integration tests for it, and I need to make the existing itests pass.

EvanKrall:
I've managed to get all the existing tests passing, wooo.

But I'm writing a test to make sure that the bounce will migrate an app from one shard to another if you change the marathon_shard in the job config, and that's failing, so this is not quite ready to merge.

paasta_itests/itest_utils.py
```diff
@@ -29,19 +30,26 @@
...
+        'id', 'backoff_factor', 'backoff_seconds', 'max_instances', 'mem', 'cpus', 'instances',
+        'marathon_shard', 'previous_marathon_shards',
+    }
     with mock.patch.object(
         MarathonServiceConfig, 'get_min_instances', autospec=True, return_value=1,
     ), mock.patch.object(
         MarathonServiceConfig, 'get_max_instances', autospec=True,
     ) as mock_get_max_instances:
         mock_get_max_instances.return_value = context.max_instances if 'max_instances' in context else None
+        service_configuration_lib._yaml_cache = {}
```
EvanKrall: Why do we even use `service_configuration_lib.read_extra_service_information`? I feel like the magic yaml caching only ever bites us.

paasta_tools/list_marathon_service_instances.py
```diff
@@ -99,27 +104,31 @@ def get_desired_marathon_configs(soa_dir):
...
 
-    apps_that_need_bouncing = actual_ids.symmetric_difference(desired_ids)
-    apps_that_need_bouncing = {long_job_id_to_short_job_id(app_id) for app_id in apps_that_need_bouncing}
+    undesired_apps_and_clients = actual_ids_and_clients.symmetric_difference(desired_ids_and_clients)
+    apps_that_need_bouncing = {long_job_id_to_short_job_id(app_id) for app_id, client in undesired_apps_and_clients}
 
     draining_hosts = get_draining_hosts()
 
-    for app_id, app in current_apps.items():
+    for (app_id, client), app in current_apps_with_clients.items():
```
solarkennedy: I think there may be a subtle bug here if we get to the place where we have an appid on a old client, but we'll never get rid of it if the *actual* bounce code can't "find" it because it isn't in a previous client list.

I hope this doesn't happen, but if it does it will mean that the app is *constantly* being asked to bounce, but it will never converge. right?

paasta_tools/marathon_serviceinit.py
```diff
@@ -220,21 +220,20 @@ def status_marathon_job_verbose(service, instance, client, cluster, soa_dir):
...
+        tasks, output = get_verbose_status_of_marathon_app(
+            marathon_client=client,
+            app=app,
+            service=service,
+            instance=instance,
+            cluster=cluster,
+            soa_dir=soa_dir,
+        )
+        all_tasks.extend(tasks)
+        all_output.append(output)
```
solarkennedy: Can you keep the previous behavior with something like
```
if not `marathon_tools.get_matching_apps_with_clients`:
  append(Warning...)
```
but save the variable so you don't have to call the function twice

paasta_tools/marathon_serviceinit.py
```diff
@@ -220,21 +220,20 @@ def status_marathon_job_verbose(service, instance, client, cluster, soa_dir):
...
-                cluster=cluster,
-                soa_dir=soa_dir,
-            )
-            all_tasks.extend(tasks)
-            all_output.append(output)
-        else:
-            all_output.append("Warning: App %s is not running yet." % app_id)
+
+    relevant_clients = clients.get_all_clients_for_service(job_config)
+    marathon_apps_with_clients = marathon_tools.get_marathon_apps_with_clients(relevant_clients)
```
solarkennedy: I guess we were calling list_apps before in the old line 223? I'm concerned about the cost of fetching all apps across all relevant clients, but I guess in the happy case this is no worse than before

paasta_tools/marathon_serviceinit.py
```diff
@@ -446,20 +445,17 @@ def perform_command(command, service, instance, cluster, verbose, soa_dir, app_i
     normal_instance_count = job_config.get_instances()
     proxy_port = marathon_tools.get_proxy_port_for_instance(service, instance, cluster, soa_dir=soa_dir)
 
-    if client is None:
-        marathon_config = marathon_tools.load_marathon_config()
-        client = marathon_tools.get_marathon_client(
```
solarkennedy: I'm glad to see this get dry'd out accross the codebase

paasta_tools/list_marathon_service_instances.py
EvanKrall: I'm not sure I follow. You mean like the scenario tested in [this itest](https://github.com/Yelp/paasta/pull/1552/files/fc9641baaf795652f35a46fde3764f418c9609f7#diff-b66d150a775e1d9eb08819b944abeec5R72)?

paasta_tools/marathon_tools.py
```diff
@@ -94,7 +98,75 @@
...
+    hash_func: Callable[[str], str]=get_config_hash,
+) -> _RendezvousHashT:
+    """For each choice, calculate the hash of the index of that choice combined with the key, then return the choice
+    whose corresponding hash is highest.
+
+    :param choices: A sequence of arbitrary values. The "winning" value will be returned."""
+    max_hash_value = None
+    max_hash_choice = None
+    for i, choice in enumerate(choices):
+        str_to_hash = MESOS_TASK_SPACER.join([str(i), key, salt])
```
solarkennedy: It seems strange to me to use the MESOS spacer here. 
Does this matter? I expected this function to be very generic and not need anything paasta-related at all

paasta_tools/marathon_serviceinit.py
EvanKrall: yes, get_matching_appids calls get_all_marathon_apps which calls list_apps.

paasta_tools/marathon_tools.py
```diff
@@ -1259,22 +1330,45 @@ def get_matching_appids(servicename: str, instance: str, client: MarathonClient)
...
 
 def get_matching_apps(servicename: str, instance: str, marathon_apps: Collection[MarathonApp]) -> List[MarathonApp]:
     """Returns a list of appids given a service and instance.
     Useful for fuzzy matching if you think there are marathon
     apps running but you don't know the full instance id"""
+    return [app for app in marathon_apps if does_app_id_match(servicename, instance, app.id)]
+
+
+def get_matching_apps_with_clients(
+    servicename: str,
```
solarkennedy: still here, can you just say "service" ?

paasta_tools/paasta_serviceinit.py
```diff
@@ -86,37 +90,36 @@ def parse_args():
...
+    _chronos: Optional[ChronosClient]
 
-    def __init__(self, cached=False):
+    def __init__(self, cached: bool=False) -> None:
         self._cached = cached
         self._marathon = None
         self._chronos = None
 
-    def marathon(self):
+    def marathon(self) -> marathon_tools.MarathonClients:
```
solarkennedy: what do you think about `marathons` ?

paasta_tools/marathon_serviceinit.py
EvanKrall: So, I removed this because `get_matching_appids` calls `get_all_marathon_apps` and returns a subset of the results; then `is_app_id_running` makes sure the app id is in the results of `list_all_marathon_app_ids`, which calls `get_all_marathon_apps`. There's no real way for the else block to fire. It's effectively like writing:

```
for thing in things:
    if thing in things:
        ...
    else:
        raise Exception(...)
```

paasta_tools/marathon_tools.py
```diff
@@ -89,7 +93,75 @@
...
+
+
+class MarathonClients(object):
+    def __init__(self, current: List[MarathonClient], previous: List[MarathonClient]) -> None:
+        self.current = current
+        self.previous = previous
+
+    def get_current_client_for_service(self, job_config: 'MarathonServiceConfig') -> MarathonClient:
+        service_instance = compose_job_id(job_config.service, job_config.instance)
+        if job_config.get_marathon_shard() is not None:
```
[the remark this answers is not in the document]
EvanKrall: get_marathon_shard can return 0, which is semantically different than None.

paasta_tools/setup_marathon_job.py
```diff
@@ -496,15 +520,29 @@ def log_deploy_error(errormsg: str, level: str='event') -> None:
 
     system_paasta_config = load_system_paasta_config()
     cluster = system_paasta_config.get_cluster()
-    existing_apps = marathon_tools.get_matching_apps(service, instance, marathon_apps)
-    new_app_list = [a for a in existing_apps if a.id == '/%s' % config['id']]
-    other_apps = [a for a in existing_apps if a.id != '/%s' % config['id']]
+    existing_apps_with_clients = marathon_tools.get_matching_apps_with_clients(
+        service,
```
solarkennedy: kwargs please

paasta_tools/setup_marathon_job.py
```diff
@@ -577,19 +615,21 @@ def log_deploy_error(errormsg: str, level: str='event') -> None:
             scaling_app_at_risk_tasks = list(task_dict['at_risk'])
 
             tasks_to_move_draining = min(len(scaling_app_draining_tasks), num_tasks_to_scale)
-            old_app_draining_tasks[new_app.id] = set(scaling_app_draining_tasks[:tasks_to_move_draining])
+            old_app_draining_tasks[(new_app.id, new_client)] = set(scaling_app_draining_tasks[:tasks_to_move_draining])
```
solarkennedy: Isn't this code assuming that the old apps only exist on the new_client? Isn't that assumption incorrect?

paasta_tools/setup_marathon_job.py
```diff
@@ -733,12 +775,10 @@ def main() -> None:
...
-    marathon_config = get_main_marathon_config()
-    client = marathon_tools.get_marathon_client(
-        marathon_config.get_url(), marathon_config.get_username(),
-        marathon_config.get_password(),
-    )
-    marathon_apps = marathon_tools.get_all_marathon_apps(client, embed_tasks=True)
+    system_paasta_config = load_system_paasta_config()
+    clients = marathon_tools.get_marathon_clients(marathon_tools.get_marathon_servers(system_paasta_config))
+    unique_clients = clients.get_all_clients()
+    marathon_apps_with_clients = marathon_tools.get_marathon_apps_with_clients(unique_clients, embed_tasks=True)
```
solarkennedy: I'm kinda worried about this part of the code scaling up 10x

tox.ini
```diff
@@ -133,6 +133,7 @@ mypy_paths =
     paasta_tools/utils.py
     paasta_tools/marathon_tools.py
     paasta_tools/long_running_service_tools.py
+    paasta_tools/paasta_serviceinit.py
```
solarkennedy: woot

paasta_tools/marathon_tools.py
EvanKrall: I can make this any arbitrary format.

paasta_tools/setup_marathon_job.py
EvanKrall: No, it's assuming that the new app only exists on the new client.

This chunk of code is putting new tasks in the various categories for old tasks if we need to scale down the current app; that way, later code will drain and kill them.

paasta_tools/setup_marathon_job.py
EvanKrall: We'll see, I guess. If it's bad, we can improve it by having it only get the tasks from the relevant clients for the services in question. (We'd also want to make a similar change in `deployd.workers.PaastaDeployWorker.process_service_instance`.)

PR #1553, opened 2017-10-10
Use sensu-plugin-python from pypi

asottile:
now that https://github.com/sensu-plugins/sensu-plugin-python/issues/13 is resolved


asottile:
probably needs a rebuild of paasta_itests

PR #1554, opened 2017-10-11
add broadcast_log_all_services_running_here

somic:
PAASTA-13173

@solarkennedy - this does not exactly follow the ticket's description but I think it might be easier.

The end result is that we will be able to log to streams of all services running on this host from bash using something like this:

```
echo foo | /path/to/paasta/python -c 'import sys; from paasta_tools.marathon_tools import broadcast_log_all_services_running_here; broadcast_log_all_services_running_here(sys.stdin.read().strip())'
```

I think it's a bit less disruptive than rewriting entire script in python.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -21,6 +21,7 @@ import datetime
 import json
 import logging
 import os
+import sys
 from collections import defaultdict
 from collections import namedtuple
 from math import ceil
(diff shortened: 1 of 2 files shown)


EvanKrall:
Probably worth exposing this function as a script/endpoint rather than having graceful_mesos_shutdown use `python -c 'import ...'`

somic:
Added paasta_broadcast_log script

paasta_tools/marathon_tools.py
```diff
@@ -1112,3 +1114,27 @@ def get_num_at_risk_tasks(app, draining_hosts):
             num_at_risk_tasks += 1
     log.debug("%s has %d tasks running on at-risk hosts." % (app.id, num_at_risk_tasks))
     return num_at_risk_tasks
+
+
+def broadcast_log_all_services_running_here(line, component='monitoring'):
```
solarkennedy: This is a good start. Eventually I would like this to be more than just marathon. For example that other srv-configs io ticket would be easier to troubleshoot if we just saw in the logs that there was a shutdown right before. But that is a chronos job.

Could you use `mesos_services_running_here` instead?

somic: @solarkennedy take a look. We filter frameworks to allow both marathon* and chronos* - is this what you meant?

solarkennedy: Sure, but how about not filtering at all?

PR #1556, opened 2017-10-12
Allow metastatus to use cached mesos state and frameworks

keymone:
- new flag `--cache-enabled` to opt into use of cache server
- for now cache server is hardcoded to be original host + port 5055

### DIFF
--- a/paasta_tools/mesos/master.py
+++ b/paasta_tools/mesos/master.py
@@ -121,10 +121,10 @@ class MesosMaster(object):
 
         retry = KazooRetry(max_tries=10)
         with zookeeper.client(
-                hosts=hosts,
-                read_only=True,
-                connection_retry=retry,
-                command_retry=retry,
+            hosts=hosts,
+            read_only=True,
+            connection_retry=retry,
+            command_retry=retry,
         ) as zk:
             def master_id(key):
                 return int(key.split("_")[-1])
--- a/paasta_tools/mesos/master.py
+++ b/paasta_tools/mesos/master.py
@@ -73,8 +73,8 @@ class MesosMaster(object):
     @util.CachedProperty(ttl=5)
     def cache_host(self):
         host_url = urlparse(self.host)
-        host_url.port = 5055  # make this configurable
-        return str(host_url)
+        host_url._replace(netloc=host_url.hostname + ':5055')
+        return host_url.geturl()
 
     @log.duration
     def _request(self, url, method=requests.get, cached=False, **kwargs):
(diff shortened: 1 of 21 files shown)


paasta_tools/mesos/master.py
```diff
@@ -64,23 +65,44 @@ def key(self):
...
-        return "{}://{}".format(self.config["scheme"], self.resolve(self.config["master"]))
+        return "{}://{}".format(
+            self.config["scheme"],
+            self.resolve(self.config["master"]),
+        )
+
+    @util.CachedProperty(ttl=5)
+    def cache_host(self):
+        host_url = urlparse(self.host)
+        host_url._replace(netloc=host_url.hostname + ':5055')
```
Rob-Johnson: I'm happy to let this go as a first pass, but I really don't want this here long term. We need something more general - could just be a 'caching_mesos_master' config value in the mesos-cli.json, or maybe we could make the caching layer noop for everything but /state.json and /frameworks?

paasta_tools/mesos/master.py
```diff
@@ -98,7 +120,12 @@ def _zookeeper_resolver(self, cfg):
         path = "/" + path
 
         retry = KazooRetry(max_tries=10)
-        with zookeeper.client(hosts=hosts, read_only=True, connection_retry=retry, command_retry=retry) as zk:
+        with zookeeper.client(
+                hosts=hosts,
```
Rob-Johnson: this indenting seems too much?

paasta_tools/mesos/master.py
keymone: yeah, i agree, this is not ideal.

paasta_tools/mesos/master.py
```diff
@@ -148,7 +183,7 @@ def resolve(self, cfg):
 
     @util.CachedProperty(ttl=15)
     def state(self):
-        return self.fetch("/master/state.json").json()
+        return self.fetch("/master/state.json", cached=True).json()
```
solarkennedy: I appreciate that we are making the cache opt-out, but doesn't this give no way for our other uses of this function to opt out?

In other words, isn't this always using the cache, and the --use-cache option which is much further removed has no effect?

keymone: this arg is a way to tell _request "use caching if available". whether caching is available is defined by cache_enabled in MesosMaster.config: https://github.com/Yelp/paasta/pull/1556/files#diff-8a09a1f8585e7ecef8050b47ad0d189cR83

PR #1558, opened 2017-10-14
make paasta metastatus marathon-shard aware

oktopuz:


paasta_tools/monitoring/check_marathon_has_apps.py
```diff
@@ -12,28 +12,38 @@
...
+        marathon_clients.previous,
+    )]
+
+
 def check_marathon_apps():
-    config = load_marathon_config()
-    if not config:
+    clients = get_marathon_clients()
+    if not clients:
         paasta_print("UNKNOWN: Failed to load marathon config")
```
matthewbentley: "Failed to load marathon clients"

paasta_tools/monitoring/check_mesos_duplicate_frameworks.py
```diff
@@ -12,18 +12,28 @@
...
 from paasta_tools.mesos.exceptions import MasterNotAvailableException
 from paasta_tools.mesos_tools import get_mesos_master
 from paasta_tools.metrics.metastatus_lib import assert_framework_count
-from paasta_tools.metrics.metastatus_lib import get_marathon_framework_ids
+from paasta_tools.paasta_metastatus import verify_marathon_clients_and_return_framework_ids
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import paasta_print
 
 
+def get_marathon_clients():
```
matthewbentley: Move to marathon_tools? (This looks the same as in check_marathon_has_apps.py?)

paasta_tools/paasta_metastatus.py
```diff
@@ -76,59 +74,73 @@ def parse_args(argv):
...
 
 
+def verify_marathon_clients_and_return_framework_ids(marathon_clients):
+    ids = []
+    unreachable_clients = []
+    for client in marathon_clients:
+        try:
+            ids.append(client.get_info().framework_id)
+        except MarathonError:
+            ids.append(str(len(ids)))
```
matthewbentley: Why is this appending `str(len(ids))`?

paasta_tools/paasta_metastatus.py
```diff
@@ -76,59 +74,73 @@ def parse_args(argv):
     return parser.parse_args(argv)
 
 
+def verify_marathon_clients_and_return_framework_ids(marathon_clients):
```
matthewbentley: Can you add a docstring to this? I'm not understanding what it's doing, especially w/r/t the second to last line (`marathon_clients[:] = ...`).

oktopuz: I wanted to implement something more intelligent than failing when one marathon shard is down, but I think I abandon this idea this time, because I don't feel like I have a good solution.

paasta_tools/metrics/metastatus_lib.py
```diff
@@ -568,32 +568,41 @@ def run_healthchecks_with_param(param, healthcheck_functions, format_options={})
     return [healthcheck(param, **format_options) for healthcheck in healthcheck_functions]
 
 
-def assert_marathon_apps(client):
-    num_apps = len(client.list_apps())
-    if num_apps < 1:
+def assert_marathon_apps(clients):
+    num_apps = [len(c.list_apps()) for c in clients]
```
solarkennedy: Do you know if this is tolerant of a shard being down or not?
We could get all sorts of errors from clients, this kind assumes that they will always return a 0.

Can you add error handling... somewhere so that if a shard is down we can return a non-healthy result but at least move on?

oktopuz: It a shard is down, it should fail at line 83 of paasta_tools/paasta_metastatus.py

oktopuz: In short, this is tolerant of a shard being down because it fails and exits with 2 long before this assertion. This is similar to the way it works now: exit with 2 if either mesos or marathon is down. 

I couldn't come up with a good solution to acknowledge the failure of one or more shards and assert heathy ones.

PR #1559, opened 2017-10-17
Only remove tmp jobs if they finished longer than one schedule interv…

jglukasik:
…al ago

To fix the bug where: 1) a batch with a long schedule (like once a month) fails, 2) the developer manually runs `paasta rerun` to fix it, 3) the Sensu check clears at first, but then goes back to failing after the tmp job is removed after one day.

This will remove a tmp job only if it finished longer than one schedule interval ago.

### DIFF
--- a/paasta_tools/cleanup_chronos_jobs.py
+++ b/paasta_tools/cleanup_chronos_jobs.py
@@ -34,6 +34,7 @@ import pysensu_yelp
 
 from paasta_tools import chronos_tools
 from paasta_tools import monitoring_tools
+from paasta_tools import utils
 from paasta_tools.check_chronos_jobs import check_chronos_job_name
 from paasta_tools.utils import InvalidJobNameError
 from paasta_tools.utils import paasta_print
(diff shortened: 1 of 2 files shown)


paasta_tools/cleanup_chronos_jobs.py
```diff
@@ -133,10 +134,18 @@ def filter_expired_tmp_jobs(client, job_names):
...
+                instance=instance,
+                cluster=cluster,
+                soa_dir=soa_dir,
+            )
+            interval = chronos_job_config.get_schedule_interval_in_seconds()
             if last_run_state != chronos_tools.LastRunState.NotRun:
                 if ((datetime.datetime.now(dateutil.tz.tzutc()) -
                      dateutil.parser.parse(last_run_time)) >
-                        datetime.timedelta(days=1)):
+                        datetime.timedelta(seconds=interval)):
```
keymone: this means jobs with short interval are now removed much faster - do developers rely on having a day worth of history? maybe we want something like `max(1 day, job interval)` so we never go lower than 1 day?

PR #1560, opened 2017-10-17
Filter out slaves from the slaves list if not in mesos state

matthewbentley:
If an agent disapears from the mesos state while running the cluster
autoscaler, the cluster autoscaler needs to take into account the fact
that it no longer exists.  This updates the autoscaler to only look at
currently existing agents at each loop while scaling down

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -520,16 +520,23 @@ class ClusterAutoscaler(ResourceLogMixin):
             current_capacity = new_capacity
             mesos_state = get_mesos_master().state_summary()
             if filtered_sorted_slaves:
+                # Filter out slaves no longer in mesos state
+                updated_filtered_slaves = [
+                    slave for slave in filtered_sorted_slaves
+                    if slave['id'] in [s['id'] for s in mesos_state.get('slaves', [])]
+                ]
+                # Get updated task counts for slaves from state
                 task_counts = get_mesos_task_count_by_slave(
                     mesos_state,
                     slaves_list=[
                         {'task_counts': slave.task_counts}
-                        for slave in filtered_sorted_slaves
+                        for slave in updated_filtered_slaves
                     ],
                 )
-                for i, slave in enumerate(filtered_sorted_slaves):
+                # Update the task counts in the slaves list
+                for i, slave in enumerate(updated_filtered_slaves):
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -520,16 +520,25 @@ def downscale_aws_resource(self, filtered_slaves, current_capacity, target_capac
             current_capacity = new_capacity
             mesos_state = get_mesos_master().state_summary()
             if filtered_sorted_slaves:
+                # Filter out slaves no longer in mesos state
+                updated_filtered_slaves = [
+                    slave for slave in filtered_sorted_slaves
+                    if slave['id'] in [s['id'] for s in mesos_state.get('slaves', [])]
```
keymone: store the inner list comp into a variable?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -520,16 +520,26 @@ def downscale_aws_resource(self, filtered_slaves, current_capacity, target_capac
...
                     ],
                 )
-                for i, slave in enumerate(filtered_sorted_slaves):
+                # Update the task counts in the slaves list
+                for i, slave in enumerate(updated_filtered_slaves):
                     slave.task_counts = task_counts[i]['task_counts']
-            filtered_slaves = filtered_sorted_slaves
+            else:
+                updated_filtered_slaves = filtered_sorted_slaves
+            filtered_slaves = updated_filtered_slaves
```
solarkennedy: I don't really understand. What does this part of the code "do"?
It doesn't look like it is actually changing state, and the task_counts variables are not used later on?

matthewbentley: This part has confused me for a while, but I think I figured out what it does.  It basically gets a list of tasks from state_summary, and then updates the counts in the slave list, so next time through the loop when we call `ec2_fitness.sort_by_ec2_fitness`, it will correctly sort by the number of running tasks.  

This is useful here, since after stopping a slave its tasks may be moved elsewhere, and the fitness sort changes.  But now that I think of it it's almost useless on the async scaledown PR, since few or no slaves will actually be killed between iterations of the loop (in that branch).

matthewbentley:
I'm going to merge this now to fix the autoscaler staleness warnings, even though this section of code will be removed with #1496

PR #1562, opened 2017-10-17
Ignore images where RepoTags is None.

EvanKrall:
Sometimes, RepoTags is None; I'm not sure why. This patch will ignore those images.


EvanKrall:
nm, @bchess already committed an equivalent patch

PR #1563, opened 2017-10-18
Make chronos itest use xenial and upate gpg key for bintray

solarkennedy:
We were not using the webupd8 java anyway, we were using the java that came from openjdk9-headless as a zookeeper dep (pretty sure).


solarkennedy:
No travis?

solarkennedy:
502
https://travis-ci.org/Yelp/paasta/jobs/289691415#L1565

EvanKrall:
**I pushed a patch to this branch** that I think will make the itests pass (Solves `HOOK-ERROR in after_scenario: ChronosAPIError: No remaining Chronos servers to try`):

I discovered that the chronos itest container was failing on start because rsyslog isn't in lsb-release on xenial -- manually installing it fixes that problem. Maybe it would be better to switch to whatever systemd provides?

PR #1568, opened 2017-10-20
Bump up taskproc

huadongliu:
This includes the fix to not reconcile queued tasks.


keymone:
merging to make sure it makes it into the release

PR #1571, opened 2017-10-20
reduce hacheck timeout in drain_lib

somic:
OPS-12282

Current 15 second timeout for both connect and read operations seems too generous to me. Let's reduce it to 3s connect timeout and 1s read timeout? I think these numbers are good even though reduction from 15 to this level might be a bit too aggressive.

requests lib supports passing timeout as int/float or as tuple - https://github.com/requests/requests/blob/master/requests/api.py#L32

### DIFF
--- a/paasta_tools/drain_lib.py
+++ b/paasta_tools/drain_lib.py
@@ -19,7 +19,7 @@ import requests
 from paasta_tools.utils import get_user_agent
 
 _drain_methods = {}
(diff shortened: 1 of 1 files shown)


somic:
Here is another spot where we have 15s timeout for requests - https://github.com/Yelp/paasta/blob/master/paasta_tools/mesos_maintenance.py#L52

PR #1573, opened 2017-10-20
adds functionality to enable specific watchers in deployd

### LINKED ISSUE #1542: Add Toggles for deployd Watchers
It would be nice if we had the ability to enable/disable deployd watchers via PaaSTA configuration files. This would make it easier to test new watchers or disable watchers that are causing problems.

CC @mattmb

fboxwala:
Fixes #1542

### DIFF
--- a/paasta_tools/deployd/master.py
+++ b/paasta_tools/deployd/master.py
@@ -162,6 +162,16 @@ class DeployDaemon(PaastaThread):
     def bounce(self, service_instance):
         self.inbox_q.put(service_instance)
 
+    @property
+    def watcher_threads_enabled(self):
+        disabled_watchers = self.config.get_disabled_watchers()
+        watcher_classes = [
+            obj[1] for obj in inspect.getmembers(watchers) if inspect.isclass(obj[1]) and
+            obj[1].__bases__[0] == watchers.PaastaWatcher
+        ]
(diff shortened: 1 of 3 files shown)


paasta_tools/deployd/master.py
```diff
@@ -259,6 +259,7 @@ def prioritise_bouncing_services(self):
 
     def start_watchers(self):
         """ should block until all threads happy"""
+        watchers_enabled = self.config.get_watchers_enabled()
```
solarkennedy: Can you rename this to `disabled_watchers`?

It is much more likely we need to toggle off the watchers we don't want, I don't want people to have to enumerate all the watchers they want, then subtract.

Plus it makes the logic less weird on line 276

paasta_tools/deployd/master.py
```diff
@@ -272,6 +273,8 @@ def start_watchers(self):
             )
             for watcher in watcher_classes
         ]
+        self.watcher_threads = [x for x in self.watcher_threads if x.__class__.__name__ not in disabled_watchers]  # noqa
```
solarkennedy: I don't like that this is changing the watcher_threads right here.

Can you instead make `watcher_threads` a property so that it returns the correct value all the time, even before running the start_watchers function?

PR #1581, opened 2017-10-25
Set new terminate timeout based on the number of tasks on killed agent

matthewbentley:
When scaling down a cluster, we can set the timout for the next slave
based on the running tasks on the most recently killed slave.  If, on
one terminate, we hit the timeout, this will try to set a lower timeout
for the next terminate based on the running tasks on the just killed
slave.

note: I'm not sure about the 30s + 30s*(running_tasks).  30s + 60s*(running_tasks) might make more sense.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -67,8 +67,7 @@ log.addHandler(logging.NullHandler())
 
 class Timer(object):
     def __init__(self, timeout):
-        self.timeout = timedelta(seconds=timeout)
-        self.start()
+        self.reset_to(timeout)
 
     def start(self):
         self.last_start = datetime.now()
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -230,6 +233,19 @@ def check_expected_slaves(self, slaves, expected_instances):
...
                 raise ClusterAutoscalingError(error_message)
 
+    def get_new_timeout_after_terminate(self, slave_to_kill):
+        """Gets a new timeout to set based on the running tasks on a slave about to be killed.
+        The new timeout will be 30s + 30s*(running_tasks), up to 300 seconds.
+        If the slave cannot be found in the state summary, just return the default 300 seconds
+        """
+        state = get_mesos_master().state_summary()
+        timeout = 300
+        for slave in state.get('slaves', []):
```
solarkennedy: This is an expensive endpoint, can you use an injected version of this instead of fetching it for each slave?

matthewbentley: The reason I'm fetching it for each slave is that tasks might move around after terminating a slave, especially if this host was previously nearly unused.

matthewbentley:
I don't think this is really worth it, so I'm going to close it

PR #1584, opened 2017-10-26
[cluster autoscaler] expand logic around scaling cancelled resources

matthewbentley:
a new rule for scaling cancelled resources (asg/sfr):
1. if any resource (in a pool) is cancelled, and we're scaling down, ignore running resources
2. Else work as normal

Edit: this prevents a scenario where spot fleets are rolled, and once the new instances are available the cluster autoscaler immediately scales them to the minimum capacity, so in order for the old sfrs to be scaled down all the way it has to incrementally bump up the new ones and down the old ones, over the course of several runs.


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -934,27 +935,52 @@ def autoscale_local_cluster(config_folder, dry_run=False, log_level=None):
...
                 pool_settings=pool_settings,
                 config_folder=config_folder,
                 dry_run=dry_run,
                 log_level=log_level,
                 utilization_error=utilization_errors[(resource['region'], resource['pool'])],
                 draining_enabled=autoscaling_draining_enabled,
-            ))
+            )
+            autoscaling_scalers[resource['pool']].append(scaler)
+            any_scale_up_pool[resource['pool']] |= utilization_errors[(resource['region'], resource['pool'])] > 0
```
matthewbentley: This line is a bit silly now that I think of it...we can figure out this info just by looking at the utilization error for the pool.  So the first part of this pull request is redundant, since we won't try to do anything with cancelled resources anyways if the pool needs to scale up.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -934,27 +935,42 @@ def autoscale_local_cluster(config_folder, dry_run=False, log_level=None):
...
+    for (region, pool), scalers in autoscaling_scalers.items():
+        # if any resource in a pool is cancelled (and it's not scaling up), only look at cancelled ones
+        if any_cancelled_in_pool[(region, pool)] and utilization_errors[(region, pool)] < 0:
+            filtered_autoscaling_scalers += [s for s in scalers if s.is_resource_cancelled()]
+            skipped = [s for s in scalers if s.is_resource_cancelled()]
+            log.info("There are cancelled resources in pool %s, skipping active resources" % pool)
+            log.info("Skipped: %s" % skipped)
+        else:
+            filtered_autoscaling_scalers += scalers
+    sorted_autoscaling_scalers = sorted(
```
mattmb: Would it be easy to pull this out into a separate function that maybe returns `sorted_autoscaling_scalers`? If it's a real pain feel free to ignore me. I just worry about how big some of these methods have got already.

mattmb: It would probably make the tests a bit clearer too

PR #1587, opened 2017-10-27
Adding Cluster Boost feature (ready to merge)

matfra:
This enable the cluster autoscaler to boost its capacity for a specific duration.
The boosted capacity is based on a multiplier of the resource usage at the time we se the boost. We call that fake multiplied value the "expected_load". This value is calculated only once.

When the boost is active, the measure of resource utilization will always be the max of expected_load and the real resource usage so the cluster can keep scaling up normally if needed.

Zookeeper is used to keep the state

This is needs more tests before shipping the the base code should not change too much.


paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -0,0 +1,160 @@
+"""
+Provides functions to temporary boost cluster capacity.
+Useful during big service bounce or failovers.
+
+This works by setting a temporary multiplier on the cluster's measured load.
+This applies to any resource (cpu, memory, disk)
+If the usage increase overtime, the boost will still apply on top of that.
```
matfra: This comment is not up to date anymore

matfra:
The cli tool to call the set_boost function has not been written yet.

paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -0,0 +1,160 @@
...
+
+class BoostAlreadyActiveError(Exception):
+    pass
+
+
+def get_zk_boost_path(region: str, pool: str) -> str:
+    return '/paasta_cluster_autoscaler/{}/{}/boost'.format(region, pool)
+
+
+def get_boosted_load(region: str, pool: str, current_load: float) -> float:
```
solarkennedy: I don't see the `zk` injected here, where does it come from?

paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -0,0 +1,160 @@
...
+
+    try:
+        zk_boost_path = get_zk_boost_path(region, pool)
+        end_time = int(zk.get(zk_boost_path + '/end_time')[0].decode('utf-8'))
+        boost_factor = float(zk.get(zk_boost_path + '/factor')[0].decode('utf-8'))
+        expected_load = float(zk.get(zk_boost_path + '/expected_load')[0].decode('utf-8'))
+
+    except Exception as e:
+        log.info(e)
+        log.debug('No boost data in Zookeeper for pool {} in region {}.'.format(
```
solarkennedy: please except this on exactly the missing zoned, and then just return. But don't log, this will be happening all the time, and if there is no boost then there is no boost.

paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def set_boost_factor(
+    region: str,
+    pool: str,
+    factor=DEFAULT_BOOST_FACTOR,
+    duration_minutes=DEFAULT_BOOST_DURATION,
+    override=False,
+) -> bool:
+    if factor < MIN_BOOST_FACTOR:
```
solarkennedy: I appreciate the "defensiveness" of this code, but in real life the show must go on.

Instead of crashing, can you instead do min() or max() where appropriate?

paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -0,0 +1,160 @@
...
+            zk.ensure_path(zk_factor_path)
+            zk.ensure_path(zk_expected_load_path)
+            zk.set(zk_end_time_path, str(end_time).encode('utf-8'))
+            zk.set(zk_factor_path, str(factor).encode('utf-8'))
+            zk.set(zk_expected_load_path, '0'.encode('utf-8'))
+        except Exception:
+            log.error('Error setting the boost in Zookeeper')
+            raise
+
+        log.info('Cluster boost: Setting capacity boost factor {} for pool {} in region {} until {}'.format(
```
solarkennedy: Can you say "Set" instead of "setting" here, because by the time were are at this point the zk is set already.

paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -0,0 +1,160 @@
...
+            factor,
+            pool,
+            region,
+            datetime.fromtimestamp(end_time).strftime('%Y-%m-%d %H:%M:%S UTC'),
+        ))
+
+        # Let's check that this factor has been properly written to zk
+        return get_boost_values(region, pool, zk) == (end_time, factor, 0)
```
solarkennedy: This looks good, but there is quite a bit of room here for type errors. Are you worried about type errors?
Also can you do a namedTuple to give semantic meaning to the "0" at the end there?

paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def clear_boost(region: str, pool: str):
+    return set_boost_factor(region, pool, factor=1, duration_minutes=0, override=True)
```
solarkennedy: kwargs all the things instead of mixing

tests/autoscaling/test_cluster_boost.py
```diff
@@ -0,0 +1,109 @@
...
+@freeze_time(TEST_CURRENT_TIME)
+def test_set_boost_factor_with_defaults():
+    fake_region = 'westeros-1'
+    fake_pool = 'default'
+    base_path = cluster_boost.get_zk_boost_path(fake_region, fake_pool)
+
+    with patch_zk_client() as mock_zk_client:
+        cluster_boost.set_boost_factor(fake_region, fake_pool)
+
+    expected_end_time = int(TEST_CURRENT_TIME.timestamp()) + 60 * cluster_boost.DEFAULT_BOOST_DURATION
```
solarkennedy: I know this is blasphemy, but I would actually be OK if we didn't add a new dependency and simply did not test anything about the timestamp at all. Or change the code to take the timestamp as a param, and inject it for test purposes.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -1055,14 +1056,17 @@ def get_mesos_utilization_error(
...
     free_pool_resources = region_pool_utilization_dict['free']
     total_pool_resources = region_pool_utilization_dict['total']
-    free_percs = []
+    usage_percs = []
     for free, total in zip(free_pool_resources, total_pool_resources):
         if math.isclose(total, 0):
             continue
-        free_percs.append(float(free) / float(total))
+        current_load = total - free
+        boosted_load = cluster_boost.get_boosted_load(region, pool, current_load)
```
solarkennedy: kwargs please, it is too easy to get the arguments in the wrong order

paasta_tools/autoscaling/cluster_boost.py
matfra: I'm confident this should work:
>>> (0.000,1.5,2) == (0,1.5,2.00000)
True

matthewbentley:
Can you add a docstring to `get_mesos_utilization_error` to give an idea of what it returns?  Especially now that it doesn't necessarily return the true error.

paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -0,0 +1,195 @@
...
+
+MAX_BOOST_DURATION = 240
+
+
+BoostValues = namedtuple(
+    'BoostValues',
+    [
+        'end_time',
+        'boost_factor',
+        'expected_load',
```
mattmb: I'm trying to understand this value. It exists because you want the boost to just be `boost_factor` * `load_when_boost_first_runs`? Why not just make the boost always multiply by the current load?

paasta_tools/paasta_cluster_boost.py
```diff
@@ -0,0 +1,149 @@
...
+    else:
+        logging.basicConfig(level=logging.WARNING)
+
+    action = args.action
+    region = args.region
+    pool = args.pool
+
+    if action == 'set':
+        # Let's disallow people to set boost on a non existing pool
+        if not check_pool_exist(pool=pool, region=region):
```
matthewbentley: Is there a reason this is only checked on 'set'?

paasta_tools/autoscaling/cluster_boost.py
matfra: the object zk is initialized inside this function.

paasta_tools/autoscaling/cluster_boost.py
matfra: done

paasta_tools/autoscaling/autoscaling_cluster_lib.py
matfra: done

paasta_tools/autoscaling/cluster_boost.py
matfra: You understand right. We had a discussion about that and if we kept applying a boost multiplier to a boosted load and so on, we could end up scaling up way too much

paasta_tools/paasta_cluster_boost.py
matfra: Initially, because I was using mesos state, making a call to this function very expensive but @solarkennedy pointed me to system_paasta_config.get_expected_slave_attributes() which is  less expensive so I guess we could sysmatically check that.

matfra:
I'm ready to merge this

paasta_tools/autoscaling/autoscaling_cluster_lib.py
solarkennedy: it doesn't look done

PR #1595, opened 2017-10-31
skip unreachable tasks when draining

somic:
low hanging fruit for PAASTA-13381

By skipping unreachable tasks like this, please note that we will stop killing these checks as well because they will never be added to tasks_to_kill.

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -172,6 +172,8 @@ def drain_tasks_and_find_tasks_to_kill(
             all_draining_tasks.add((task, client))
 
     for task, client in all_draining_tasks:
+        if task.state == 'TASK_UNREACHABLE':
+            continue
         try:
             drain_method.drain(task)
         except Exception as e:
@@ -181,6 +183,8 @@ def drain_tasks_and_find_tasks_to_kill(
             tasks_to_kill.add((task, client))
 
     for task, client in all_draining_tasks:
+        if task.state == 'TASK_UNREACHABLE':
+            continue
         try:
             if drain_method.is_safe_to_kill(task):
                 tasks_to_kill.add((task, client))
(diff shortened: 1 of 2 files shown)


paasta_tools/setup_marathon_job.py
```diff
@@ -172,6 +172,8 @@ def drain_tasks_and_find_tasks_to_kill(
             all_draining_tasks.add((task, client))
 
     for task, client in all_draining_tasks:
```
solarkennedy: Whoa, isn't this loop really wasteful, especially in large task lists?

We are iterating over it twice on line 185?

Can you combine them so we only have to drain/check once?

Past that, is UNREACH the only state where we can do this? How instead we only do the draining if it is RUNNING?

somic:
@solarkennedy - another pass? (tests will fail, I will fix if this looks good)

Here we will now drain only tasks from tasks_to_drain, vs before when we drained all tasks from tasks_to_drain, already_draining_tasks and at_risk_tasks.

Also, we will kill all tasks that are not TASK_RUNNING or are safe to kill or is_safe_to_kill raises an exception.

paasta_tools/setup_marathon_job.py
```diff
@@ -170,21 +170,21 @@ def drain_tasks_and_find_tasks_to_kill(
...
-            )
-            tasks_to_kill.add((task, client))
-
-    for task, client in all_draining_tasks:
-        try:
-            if drain_method.is_safe_to_kill(task):
+            if task.state != 'TASK_RUNNING' or drain_method.is_safe_to_kill(task):
                 tasks_to_kill.add((task, client))
-                log_bounce_action(line='%s bounce killing drained task %s' % (bounce_method, task.id))
+                log_bounce_action(line='%s bounce killing not_running or drained task %s' % (bounce_method, task.id))
```
solarkennedy: Can you inject task.state directly so we can see it for debugging?

solarkennedy:
Can you address the failing tests?

PR #1596, opened 2017-10-31
check_mesos_outdated_tasks

oktopuz:
internal ticket: PAASTA-13357

This check looks for containers serving previous versions.

```
./check_mesos_outdated_tasks.py --bounce-time 2
CRITICAL - There are 2 tasks running in norcal-prod that are more than 2h older than their last bounce.
SERVICE.INSTANCE                COMMIT    CREATED               HOSTNAME                     CONTAINER
yelp-main.mobile_alerts_update  addafd00  2017-10-11 15:28:46   paasta23-r6-sfo2             23c11d92-4ad9-467c-98e2-4b5767d6e5af-S1.9e84d80f-64b6-49fb-b609-ff49dd0300c6
yelp-main.elites_csv_kew_worke  1e332c88  2017-10-20 14:21:15   paasta64-r1-sfo2             a3e48ca4-3bb6-463f-af46-7595829d04ff-S39.1ff04304-c7b4-415e-a49c-28d5f7096f47
```

### DIFF
--- a/paasta_tools/monitoring/check_mesos_outdated_tasks.py
+++ b/paasta_tools/monitoring/check_mesos_outdated_tasks.py
@@ -12,6 +12,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import argparse
 import datetime
 import sys
 import time
@@ -21,6 +22,23 @@ from paasta_tools.utils import load_system_paasta_config
 
 
 OUTPUT_FORMAT = "{:<30}  {:<8}  {:<20}  {:<27}  {}"
+FRAMEWORK_NAME = "marathon"
+MAX_BOUNCE_TIME_IN_HOURS = 4
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='Find all containers serving previous push versions.',
+    )
+    parser.add_argument(
+        '--bounce-time', dest="bounce_time", type=int,
+        default=MAX_BOUNCE_TIME_IN_HOURS,
+        help=(
+            "Ignore versions that were launched in the last BOUNCE_TIME hours "
+            "because they probably are still bouncing."
+        ),
+    )
+    return parser.parse_args()
 
 
 def get_mesos_state():
@@ -30,7 +48,7 @@ def get_mesos_state():
 
 def marathon_tasks(state):
     for framework in state.get('frameworks', []):
-        if framework['name'].lower().startswith('marathon'):
+        if framework['name'].lower().startswith(FRAMEWORK_NAME):
             for task in framework.get('tasks', []):
                 yield task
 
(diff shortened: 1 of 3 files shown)


paasta_tools/monitoring/check_mesos_outdated_tasks.py
```diff
@@ -0,0 +1,122 @@
...
+
+from paasta_tools.mesos_tools import get_mesos_master
+from paasta_tools.utils import load_system_paasta_config
+
+
+OUTPUT_FORMAT = "{:<30}  {:<8}  {:<20}  {:<27}  {}"
+
+
+def get_mesos_state():
+    state = get_mesos_master().state
```
Rob-Johnson: :( can we used the caching stuff that maksym wrote?

Further, can we just call the /frameworks endpoint, rather than getting the whole state?

paasta_tools/monitoring/check_mesos_outdated_tasks.py
```diff
@@ -0,0 +1,122 @@
...
+    return state
+
+
+def marathon_tasks(state):
+    for framework in state.get('frameworks', []):
+        if framework['name'].lower().startswith('marathon'):
```
Rob-Johnson: can you make 'marathon' a constant?

paasta_tools/monitoring/check_mesos_outdated_tasks.py
```diff
@@ -0,0 +1,122 @@
...
+    for slave in state['slaves']:
+        res[slave['id']] = slave['hostname']
+    return res
+
+
+def group_running_tasks_by_job_id_and_gitsha(state):
+    res = {}
+    for t in marathon_tasks(state):
+        if t['state'] == 'TASK_RUNNING':
+            job_id = t['name'][:t['name'].find('.', t['name'].find('.') + 1)]
```
Rob-Johnson: don't we already have functions in marathon_tools to do this?

paasta_tools/monitoring/check_mesos_outdated_tasks.py
oktopuz: @keymone could you explain please how does the caching works?
I tried to pass use_mesos_cache=True to get_mesos_master, but it failed because nothing was listening on 5055.
https://github.com/Yelp/paasta/blob/master/paasta_tools/mesos/master.py#L76

paasta_tools/monitoring/check_mesos_outdated_tasks.py
```diff
@@ -0,0 +1,122 @@
...
+        deploy_time = datetime.datetime.fromtimestamp(int(t['statuses'][0]['timestamp'])).strftime('%Y-%m-%d %H:%M:%S')
+        container_name = "{}.{}".format(
+            t['slave_id'],
+            t['statuses'][0]['container_status']['container_id']['value'],
+        )
+        hostname = slave_id2hostname[t['slave_id']]
+        hostname = hostname[:hostname.find('.')]
+        output.append(
+            OUTPUT_FORMAT.format(
+                job_id.replace('--', '_')[:30],
```
solarkennedy: can you call this task_id ?

paasta_tools/monitoring/check_mesos_outdated_tasks.py
```diff
@@ -0,0 +1,122 @@
...
+                slave_id2hostname,
+            ))
+    return output
+
+
+def main():
+    cluster = load_system_paasta_config().get_cluster()
+    output = check_mesos_tasks()
+    if output:
+        print("CRITICAL - There are {} outdated instances running in {}"
```
solarkennedy: Can you make the 4*3600 a argument so that you can get to it here and print it? (there are N tasks running that are T(time) older than their last bounce)

paasta_tools/monitoring/check_mesos_outdated_tasks.py
oktopuz: We have marathon_tools.parse_service_instance_from_executor_id that returns a tuple of srv_name, srv_instance. Then I still need to concatenate them and to get gitsha from the same task_id.
It looks more complex than the current implementation.

paasta_tools/monitoring/check_mesos_outdated_tasks.py
keymone: @oktopuz yeah, it's not deployed everywhere. passing `use_mesos_cache=True` is a way to opt-in if you know that caching is deployed, ideally we'd have that config option in mesos configs. also /slaves endpoint is not being cached.

oktopuz: @Rob-Johnson I'll configure this check to use caching endpoint when we have them everywhere.
For now I don't expect us to run this check more often than once an hour. That should be safe for masters.

oktopuz: I replaced it with
```
state = get_mesos_master(use_mesos_cache=True).state
```

PR #1601, opened 2017-11-03
Initial Support for a Sharded Marathon Dashboard

nhandler:
This is a first pass at adding a script to spit out information about the service instances in a cluster and their associated marathon shard. The API will ultimately get queries and displayed as a nice dashboard by an external service.

TODO:

* Add a real unit test
* Add type annotations

### DIFF
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -702,7 +702,7 @@
           "type": "string",
           "description": "Instance name"
         },
-        "shard": {
+        "shard_url": {
           "type": "string",
           "description": "Marathon Shard URL"
         }
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -678,7 +678,7 @@
     },
     "MarathonDashboard": {
       "type": "object",
-      "description": "A list of Marathon service, instance, and shard number for one or more clusters",
+      "description": "A list of Marathon service, instance, and shard url for one or more clusters",
       "additionalProperties": {
         "$ref": "#/definitions/MarathonDashboardCluster"
       }
@@ -692,7 +692,7 @@
     },
     "MarathonDashboardItem": {
       "type": "object",
-      "description": "Marathon service, instance, and shard number",
+      "description": "Marathon service, instance, and shard url",
       "properties": {
         "service": {
           "type": "string",
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -5,9 +5,15 @@
     "version": "1.0.0"
   },
   "basePath": "/v1",
-  "schemes": ["http"],
-  "consumes": ["application/x-www-form-urlencoded"],
-  "produces": ["application/json"],
+  "schemes": [
+    "http"
+  ],
+  "consumes": [
+    "application/x-www-form-urlencoded"
+  ],
+  "produces": [
+    "application/json"
+  ],
   "tags": [
     {
       "name": "service",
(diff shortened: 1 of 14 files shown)


paasta_tools/api/api_docs/swagger.json
```diff
@@ -546,14 +614,50 @@
...
+      "properties": {
+        "service": {
+          "type": "string",
+          "description": "Service name"
+        },
+        "instance": {
+          "type": "string",
+          "description": "Instance name"
+        },
+        "shard": {
```
solarkennedy: can you do shard_url instead of just 'shard' ?

paasta_tools/marathon_dashboard.py
```diff
@@ -0,0 +1,90 @@
...
+            soa_dir=soa_dir,
+        )
+        client = marathon_clients.get_current_client_for_service(job_config=service_config)
+        shard = client.servers[0]
+        service_info = {
+            'service': service,
+            'instance': instance,
+            'shard': shard,
+        }
+        dashboard[cluster].append(service_info)
```
solarkennedy: This "feels" like overkill. Would you able to achieve the same results if shard_url was an additional entry in the "status" api for a service/instance?

yelp_package/dockerfiles/itest/api/marathon.json
```diff
@@ -1 +1,30 @@
-{"marathon_config": {"url": "http://marathon:8080", "password": null, "user": null}}
+{
+  "marathon_config": {
+    "url": "http://marathon:8080",
```
solarkennedy: One serious downside to this: the urls generated by this tool won't be "actually" useful to developers :(

The "real" urls for developers are yelpy, and change from dev/prod, etc. Any ideas on how to provide "real" urls?

I was thinking if we could have a magic function that did that, then it could power this dashboard *and* enable a clickable url in paasta status.

nhandler: My original plan was to do that at the UI layer (which will be a separate CR) and have some sort of map from these urls to the "real" urls. But after thinking about this some more, I guess I could put the map in the system paasta configs and have these scripts use those URLs instead.

paasta_tools/marathon_dashboard.py
nhandler: It would require a lot more api requests to make the UI work. We can only get the status of a specific service instance with the `status` api. This new endpoint exposes information on all of the service instances in the cluster at once

paasta_tools/marathon_dashboard.py
```diff
@@ -0,0 +1,90 @@
...
+        instance = service_instance[1]
+        service_config = load_marathon_service_config(
+            service=service,
+            instance=instance,
+            cluster=cluster,
+            load_deployments=False,
+            soa_dir=soa_dir,
+        )
+        client = marathon_clients.get_current_client_for_service(job_config=service_config)
+        shard_url = client.servers[0]
```
EvanKrall: I know the dashboard service you're planning to make doesn't consume the output of this command, but this line raises a question for me:

Will the dashboard service proxy requests to the marathon dashboards? If not, we'd want to expose the dev-facing RO proxies instead of the raw marathon URL.

nhandler: I guess @solarkennedy's comment got buried. Currently, I'm working on adding logic to map this shard_url to the proper accessible url (which will be stored in system paasta configs).

nhandler:
Looks like this is finally passing tests again after I updated the dashboard to display usable links.

paasta_tools/api/api_docs/swagger.json
```diff
@@ -601,14 +669,50 @@
...
+    "MarathonDashboardCluster": {
+      "type": "array",
+      "description": "List of all the MarathonDashboardItems for a cluster",
+      "items": {
+        "$ref": "#/definitions/MarathonDashboardItem"
+      }
+    },
+    "MarathonDashboardItem": {
+      "type": "object",
+      "description": "Marathon service, instance, and shard number",
```
EvanKrall: shard url, not shard number

paasta_tools/marathon_tools.py
```diff
@@ -135,6 +135,9 @@ def __init__(self, current: List[MarathonClient], previous: List[MarathonClient]
         self.current = current
         self.previous = previous
 
+    def __repr__(self) -> str:
+        return 'Current: %s | Previous: %s' % (self.current, self.previous)
```
EvanKrall: I think usually it's good to have __repr__ return something that would more or less evaluate to the real thing, so I'd prefer `'MarathonClients(current=%r, previous=%r)' % (self.current, self.previous)`

nhandler:
I addressed @EvanKrall's feedback. Test failures are unrelated to this PR and I have prepared a fix in https://github.com/Yelp/paasta/pull/1646

PR #1605, opened 2017-11-05
Always fix the slack of old marathon apps when bouncing

solarkennedy:
I noticed this was happening during big bounces and lucy bounces. Here is what happens:

* paasta restart (or other signal) is initiated more than 1 time, like run paasta restart 100 times
* 100 apps are created, all waiting, 1 "new app" is the desired one
* The bounce never tries to scaled down those old apps, so they have to launch, then it drains, etc.

This same "take in the slack" technique was already being applied to the new_app on scaledown. Now it will also be applied to all the other old apps, because we are also scaling those down too.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -1484,3 +1484,11 @@ def broadcast_log_all_services_running_here(line: str, component: str='monitorin
 
 def broadcast_log_all_services_running_here_from_stdin(component: str='monitoring') -> None:
     broadcast_log_all_services_running_here(sys.stdin.read().strip())
+
+
+def take_up_slack(client: MarathonClient, app: MarathonApp) -> None:
+    slack = max(app.instances - len(app.tasks), 0)
+    if slack > 0:
+        log.info("Scaling %s down from %d to %d instances to remove slack." %
+                 (app.id, app.instances, app.instances - slack))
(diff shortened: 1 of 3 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -1484,3 +1484,11 @@ def broadcast_log_all_services_running_here(line: str, component: str='monitorin
...
 def broadcast_log_all_services_running_here_from_stdin(component: str='monitoring') -> None:
     broadcast_log_all_services_running_here(sys.stdin.read().strip())
+
+
+def take_up_slack(client: MarathonClient, app: MarathonApp) -> None:
+    slack = max(app.instances - len(app.tasks), 0)
+    if slack > 0:
+        log.info("Scaling %s down from %d to %d instances to remove slack." %
+                 (app.id, app.instances, app.instances - slack))
+        client.scale_app(app_id=app.id, instances=(app.instances - slack), force=True)
```
EvanKrall: Suggestion, feel free to ignore: The whole slack = max(instances - tasks, 0), instances = instances-slack thing is mildly confusing. Maybe:

    target_instances = min(app.instances, app.tasks)
    if target_instances < app.instances:
        log.info(...)
        client.scale_app(..., instances=target_instances, ...)

PR #1606, opened 2017-11-06
Cluster autoscaler typing

matthewbentley:
A few issues I ran into:
1. I'm not sure if `current_capacity` should be a float or an int
2. I'm not sure if we need the `TypedDict`s for `ClusterAutoscalingResource` and `ResourcePoolSetting`
3. A `MesosState` type should probably be defined somewhere


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -701,7 +811,7 @@ def set_capacity(self, capacity):
                             "Not updating target capacity because this is a cancelled SFR, "
                             "we are just draining and killing the instances",
                         )
-                        return
+                        return None
```
solarkennedy: kinda hints that we should be raising something?

I'm surprised that this and line 822 are not bugs. 

Still ship though, it looks like it 'normally' returns a request response dict

matthewbentley: The returns here aren't actually checked outside of tests; I left them in to keep the test logic the same.

PR #1607, opened 2017-11-06
Add bounce logic for json secret files

mattmb:
This adds the neccessary logic to the bounce code so that services will
be bounced when the related json files containing a secret are updated
*and* the signature of a secret used by that service changes.

* Make deployd aware of json files in secrets subdir and get it to
check if marathon app id has changed
* Check env vars in marathon config for SECRET(my-secret) format and
retreive signature for inclusion in config hash
* itest to check it works

### DIFF
--- a/paasta_tools/deployd/watchers.py
+++ b/paasta_tools/deployd/watchers.py
@@ -284,10 +284,19 @@ class YelpSoaEventHandler(pyinotify.ProcessEvent):
         name = '.'.join([__name__, type(self).__name__])
         return logging.getLogger(name)
 
-    def filter_event(self, event):
+    def get_service_name_from_event(self, event):
+        """Get service_name from the file inotify event,
+        returns None if it is not an event we're interested in"""
         starts_with = ['marathon-', 'deployments.json']
         if any([event.name.startswith(x) for x in starts_with]):
-            return event
+            service_name = event.path.split('/')[-1]
+        elif event.name.endswith('.json') and event.path.split('/')[-1] == 'secrets':
+            # this is needed because we put the secrets json files in a
+            # subdirectory so the service name would be "secrets" otherwise
+            service_name = event.path.split('/')[-2]
+        else:
+            service_name = None
+        return service_name
 
     def watch_new_folder(self, event):
         if event.maskname == 'IN_CREATE|IN_ISDIR' and '.~tmp~' not in event.pathname:
@@ -303,10 +312,9 @@ class YelpSoaEventHandler(pyinotify.ProcessEvent):
     def process_default(self, event):
         self.log.debug(event)
         self.watch_new_folder(event)
-        event = self.filter_event(event)
-        if event:
+        service_name = self.get_service_name_from_event(event)
+        if service_name:
             self.log.info("Change of {} in {}".format(event.name, event.path))
-            service_name = event.path.split('/')[-1]
             self.bounce_service(service_name)
 
     def bounce_service(self, service_name):
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -58,6 +58,8 @@ from paasta_tools.mesos_tools import filter_mesos_slaves_by_blacklist
 from paasta_tools.mesos_tools import get_mesos_network_for_net
 from paasta_tools.mesos_tools import get_mesos_slaves_grouped_by_attribute
 from paasta_tools.mesos_tools import mesos_services_running_here
+from paasta_tools.secret_tools import get_hmac_for_secret
+from paasta_tools.secret_tools import is_secret_ref
 from paasta_tools.utils import _log
 from paasta_tools.utils import BranchDict
 from paasta_tools.utils import compose_job_id
@@ -673,7 +675,7 @@ class MarathonServiceConfig(LongRunningServiceConfig):
         code_sha = get_code_sha_from_dockerurl(docker_url)
 
         config_hash = get_config_hash(
-            self.sanitize_for_config_hash(complete_config),
+            self.sanitize_for_config_hash(complete_config, system_paasta_config),
             force_bounce=self.get_force_bounce(),
         )
         complete_config['id'] = format_job_id(self.service, self.instance, code_sha, config_hash)
@@ -681,16 +683,44 @@ class MarathonServiceConfig(LongRunningServiceConfig):
         log.debug("Complete configuration for instance is: %s", complete_config)
         return complete_config
 
-    def sanitize_for_config_hash(self, config: FormattedMarathonAppDict) -> Dict[str, Any]:
-        """Removes some data from complete_config to make it suitable for
+    def sanitize_for_config_hash(
+        self,
+        config: FormattedMarathonAppDict,
+        system_paasta_config: SystemPaastaConfig,
+    ) -> Dict[str, Any]:
+        """Removes some data from config to make it suitable for
         calculation of config hash.
+
+        Also adds secret HMACs so that we bounce if secret data has changed.
+        We need this because the reference to the secret is all Marathon gets
+        and this will not change.
 
         :param config: complete_config hash to sanitize
         :returns: sanitized copy of complete_config hash
         """
         ahash = {key: copy.deepcopy(value) for key, value in config.items() if key not in CONFIG_HASH_BLACKLIST}
         ahash['container']['docker']['parameters'] = self.format_docker_parameters(with_labels=False)  # type: ignore
+        secret_hashes = self.get_secret_hashes(config['env'], system_paasta_config.get_vault_environment())
+        if secret_hashes:
+            ahash['paasta_secrets'] = secret_hashes
         return ahash
+
+    def get_secret_hashes(
+        self,
+        env: Dict[str, str],
+        vault_environment: str,
+    ) -> Dict[str, str]:
+
+        secret_hashes = {}
+        for _, env_var_val in env.items():
+            if is_secret_ref(env_var_val):
+                secret_hashes[env_var_val] = get_hmac_for_secret(
+                    env_var_val=env_var_val,
+                    service=self.service,
+                    soa_dir=self.soa_dir,
+                    vault_environment=vault_environment,
+                )
+        return secret_hashes
 
     def get_healthchecks(
         self,
--- a/paasta_tools/secret_tools.py
+++ b/paasta_tools/secret_tools.py
@@ -1,0 +1,56 @@
+# Copyright 2015-2017 Yelp Inc.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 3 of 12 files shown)


mattmb:
Damn, the itests are passing on my local machine but they seem to fail consistently on travis :-(

I guess I'll try and work out why...

paasta_tools/secret_tools.py
```diff
@@ -0,0 +1,45 @@
...
+import re
+
+from paasta_tools.utils import SystemPaastaConfig
+
+SECRET_REGEX = "^SECRET\([A-Za-z0-9_-]*\)$"
+
+
+def is_secret_ref(val: str) -> bool:
+    pattern = re.compile(SECRET_REGEX)
+    return True if pattern.match(val) else False
```
matthewbentley: `return pattern.match(val) is not None`?

paasta_tools/utils.py
```diff
@@ -1694,6 +1695,12 @@ def get_taskproc(self) -> Dict:
     def get_disabled_watchers(self) -> List:
         return self.config_dict.get('disabled_watchers', [])
 
+    def get_vault_environment(self) -> str:
+        """ Get the environment name for the vault cluster
+        This must match the environment keys in the secret json files
+        used by all services in this cluster"""
+        return self.config_dict.get('vault_environment')
```
matthewbentley: Do you want a default here?  Otherwise I think `.get` will return `None` (breaking the `str` type of the function)

paasta_tools/secret_tools.py
```diff
@@ -0,0 +1,45 @@
...
+) -> str:
+    secret_name = _get_secret_name_from_ref(val)
+    try:
+        secret_path = os.path.join(soa_dir, service, "secrets", "{}.json".format(secret_name))
+        with open(secret_path, 'r') as json_secret_file:
+            secret_file = json.load(json_secret_file)
+            return secret_file['environments'][system_paasta_config.get_vault_environment()]['signature']
+    except IOError as e:
+        print("Failed to load json secret at {}".format(secret_path))
+        return None
```
matthewbentley: If you're returning `None` the type of the function needs to be `Optional[str]`

paasta_tools/deployd/watchers.py
```diff
@@ -288,6 +288,8 @@ def filter_event(self, event):
         starts_with = ['marathon-', 'deployments.json']
         if any([event.name.startswith(x) for x in starts_with]):
             return event
+        elif '.json' in event.name and event.path.split('/')[-1] == 'secrets':
```
Rob-Johnson: can this be ``event.endswith('.json')`` ?

paasta_tools/deployd/watchers.py
```diff
@@ -288,6 +288,8 @@ def filter_event(self, event):
...
+            return event
```
Rob-Johnson: I think this function should return a boolean indicating whether it's interesting or not, instead of returning the event or nothing (I know that you've not changed that in this PR, but its worth fixing whilst we're here).

paasta_tools/deployd/watchers.py
```diff
@@ -305,8 +307,13 @@ def process_default(self, event):
         self.watch_new_folder(event)
         event = self.filter_event(event)
         if event:
+            if '.json' in event.name and event.path.split('/')[-1] == 'secrets':
```
Rob-Johnson: given that you've done this twice, it might be worth extracting a ``def service_for_event`` function to encapsulate this logic.

paasta_tools/marathon_tools.py
```diff
@@ -674,25 +676,53 @@ def format_marathon_app_dict(self) -> FormattedMarathonAppDict:
...
         log.debug("Complete configuration for instance is: %s", complete_config)
         return complete_config
 
-    def sanitize_for_config_hash(self, config: FormattedMarathonAppDict) -> Dict[str, Any]:
+    def sanitize_for_config_hash(
+        self,
+        config: FormattedMarathonAppDict,
+        system_paasta_config: SystemPaastaConfig,
+    ) -> Dict[str, Any]:
         """Removes some data from complete_config to make it suitable for
```
Rob-Johnson: typo 'removes data from complete_config' should be 'removed data from config'

paasta_tools/secret_tools.py
```diff
@@ -0,0 +1,45 @@
...
+SECRET_REGEX = "^SECRET\([A-Za-z0-9_-]*\)$"
+
+
+def is_secret_ref(val: str) -> bool:
+    pattern = re.compile(SECRET_REGEX)
+    return True if pattern.match(val) else False
+
+
+def get_hmac_for_secret(
+        val: str,
```
Rob-Johnson: fix indentation - needs to come in by 2 spaces.

paasta_tools/secret_tools.py
```diff
@@ -0,0 +1,45 @@
...
+        service: str,
+        soa_dir: str,
+        system_paasta_config: SystemPaastaConfig,
+) -> str:
+    secret_name = _get_secret_name_from_ref(val)
+    try:
+        secret_path = os.path.join(soa_dir, service, "secrets", "{}.json".format(secret_name))
```
Rob-Johnson: can you break this onto new lines.

Rob-Johnson: you can move this line outside of the try.

paasta_tools/secret_tools.py
```diff
@@ -0,0 +1,45 @@
...
+        with open(secret_path, 'r') as json_secret_file:
+            secret_file = json.load(json_secret_file)
+            return secret_file['environments'][system_paasta_config.get_vault_environment()]['signature']
```
Rob-Johnson: what happens if this dict isn't well formed? worth making these .gets() instead?

paasta_tools/secret_tools.py
```diff
@@ -0,0 +1,45 @@
...
+def is_secret_ref(val: str) -> bool:
+    pattern = re.compile(SECRET_REGEX)
+    return True if pattern.match(val) else False
+
+
+def get_hmac_for_secret(
+        val: str,
+        service: str,
+        soa_dir: str,
+        system_paasta_config: SystemPaastaConfig,
```
Rob-Johnson: if you only actually need ``system_paasta_config.get_vault_environment()`` can make this param ``vault_environment`` instead?

paasta_tools/marathon_tools.py
```diff
@@ -674,25 +676,53 @@ def format_marathon_app_dict(self) -> FormattedMarathonAppDict:
...
+        We need this because the reference to the secret is all Marathon gets
+        and this will not change.
+
         :param config: complete_config hash to sanitize
         :returns: sanitized copy of complete_config hash
         """
         ahash = {key: copy.deepcopy(value) for key, value in config.items() if key not in CONFIG_HASH_BLACKLIST}
         ahash['container']['docker']['parameters'] = self.format_docker_parameters(with_labels=False)  # type: ignore
+        secret_hashes = self.get_secret_hashes(config['env'], system_paasta_config)
+        if secret_hashes:
```
Rob-Johnson: this is the secret sauce thats going to avoid a big bounce right?

paasta_tools/secret_tools.py
mattmb: Interesting, I also don't have a test that exercises this, maybe it would through a type error if I did...

paasta_tools/utils.py
mattmb: I couldn't decide what was more dangerous tbh. I think None is probably best and I'll change it to Optional[str].

paasta_tools/deployd/watchers.py
mattmb: sgtm

paasta_tools/marathon_tools.py
mattmb: Right, that's the idea

paasta_tools/secret_tools.py
mattmb: It's a good question. Certainly the env key could be missing...

I think catching `KeyError` might be good idea...

paasta_tools/secret_tools.py
matthewbentley: I think you'll only get an error on `tox -e mypy`, but only once you add this file to that section of `tox.ini`

mattmb: Huh, weird I added it to the tox.ini and some tests but it still didn't complain. Although it also doesn't complain with Optional[str]. Anyway, I'm sure it should be optional!

paasta_tools/deployd/watchers.py
mattmb: I think this is a good thing to do and I'm going to make it contain the filtering logic too.

mattmb:
@Rob-Johnson @matthewbentley think I've covered all the feedback if you've got a mo for another look.

paasta_tools/deployd/watchers.py
```diff
@@ -284,10 +284,19 @@ def log(self):
...
         return logging.getLogger(name)
 
-    def filter_event(self, event):
+    def get_service_name_from_event(self, event):
+        """Get service_name from the file inotify event,
+        returns None if it is not an event we're interested in"""
         starts_with = ['marathon-', 'deployments.json']
         if any([event.name.startswith(x) for x in starts_with]):
-            return event
+            service_name = event.path.split('/')[-1]
```
solarkennedy: This is dangerous. What if the path doesn't have a '/' ?

You need to catch that

paasta_tools/deployd/watchers.py
```diff
@@ -303,10 +312,9 @@ def watch_new_folder(self, event):
     def process_default(self, event):
         self.log.debug(event)
         self.watch_new_folder(event)
-        event = self.filter_event(event)
-        if event:
+        service_name = self.get_service_name_from_event(event)
+        if service_name:
             self.log.info("Change of {} in {}".format(event.name, event.path))
```
solarkennedy: Can you add "scheduling a bounce for $service?

paasta_tools/secret_tools.py
```diff
@@ -0,0 +1,56 @@
...
+SECRET_REGEX = "^SECRET\([A-Za-z0-9_-]*\)$"
+
+
+def is_secret_ref(val: str) -> bool:
+    pattern = re.compile(SECRET_REGEX)
+    return pattern.match(val) is not None
+
+
+def get_hmac_for_secret(
+    val: str,
```
solarkennedy: Can you use a better variable than `val`? What "is" it?

paasta_tools/deployd/watchers.py
mattmb: I haven't changed this just moved it around. I'm pretty sure the paths are guaranteed to be absolute because we pass an absolute path to watch.

paasta_tools/deployd/watchers.py
mattmb: I think this log line is okay. Technically we haven't decided what service instances to bounce yet. That happens in `bounce_service`. So this just logs that we have detected a change in a relevant file.

PR #1608, opened 2017-11-06
Make paasta local-run lock on pulls

solarkennedy:
Although I hate to do it, it looks like we can't trust docker to pull 2 images simultaneously :(

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -52,6 +52,7 @@ from paasta_tools.utils import paasta_print
 from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import PaastaNotConfiguredError
 from paasta_tools.utils import SystemPaastaConfig
+from paasta_tools.utils import timed_flock
 from paasta_tools.utils import Timeout
 from paasta_tools.utils import TimeoutError
 from paasta_tools.utils import validate_service_instance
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -418,15 +419,17 @@ def docker_pull_image(docker_url):
...
     DEVNULL = open(os.devnull, 'wb')
-    ret, output = _run('docker pull %s' % docker_url, stream=True, stdin=DEVNULL)
-    if ret != 0:
-        paasta_print(
-            "\nPull failed. Are you authorized to run docker commands?",
-            file=sys.stderr,
-        )
-        sys.exit(ret)
+    with open('/tmp/paasta-local-run-pull.lock', 'w') as f:
+        with timed_flock(f, seconds=1800):
```
matthewbentley: Would it be useful to do something other than just throw a `TimeoutError` exception here?

paasta_tools/cli/cmds/local_run.py
```diff
@@ -418,15 +419,17 @@ def docker_pull_image(docker_url):
     bindings due to the docker auth/registry transition. Once we are past Docker 1.6
     we can use better credential management, but for now this function assumes the
     user running the command has already been authorized for the registry"""
-    paasta_print("Please wait while the image (%s) is pulled..." % docker_url, file=sys.stderr)
+    paasta_print("Please wait while the image (%s) is pulled (times out after 30m)..." % docker_url, file=sys.stderr)
```
nhandler: Can you move the timeout to a variable that you reference here and in the call to `timed_flock`? That way, they stay in sync even when we update the timeout in the future.

paasta_tools/cli/cmds/local_run.py
nhandler: A retry might make sense. Especially in the case of a Jenkins local-run for an itest, that would prevent the healthcheck from failing intermittently.

solarkennedy: Why? If we are going to retry, why not just wait longer?

If we would retry 10times, then I'll just make the timeout be 18000 instead

solarkennedy:
I agree ideally I would fix docker, but in practice I'm not. 

You say this might break non-interactive scripts, but why doesn't them being non-interactive make a difference? The thing is they (the non-interactive use-cases) are breaking *now* because of this issue, the PR is to fix them!

paasta_tools/cli/cmds/local_run.py
nhandler: You are right

PR #1609, opened 2017-11-06
Adjust check_mesos_duplicate_frameworks for sharding

oktopuz:
Internal ticket: PAASTA-13358

After this change check_mesos_duplicate_frameworks will no longer alert when the number of marathon frameworks registered in Mesos is bigger than the number of different frameworks configured in /etc/paasta/marathon.json. It will only check that all marathon framework configured in marathon.json are connected.


matthewbentley:
Shouldn't this be done as part of `check_mesos_active_frameworks`?  AFAICT, that check is that every configured framework is active (ie the configured frameworks are a subset of the active frameworks), while this one should be that there are no extra frameworks connected.

oktopuz:
I need help with this PR to figure out what we want to monitor.

As @matthewbentley mentioned there is `check_mesos_active_frameworks`. However there only verification it does is whether mesos has connected frameworks named 'marathon' and 'chronos'.
It certainly isn't sufficient for sharding.

On the other hand if `check_mesos_duplicate_frameworks` should ensure that there are no extra frameworks connected, then we probably should get rid of it, because there is an extra marathon framework connected to mesos in mesosstage, and that configuration is correct.

matthewbentley:
In that case, I'd say rename this check to `check_mesos_active_frameworks` and scrap the other one (so this one checks that at least the expected frameworks are active).  And maybe emit a warning if there are active frameworks that are not configured?

oktopuz:
I made check_mesos_duplicate_frameworks verify that there are no extra frameworks connected, and am going to update check_mesos_active_frameworks in a new PR.

This is ready for review.

PR #1612, opened 2017-11-06
increase read timeout for maintenance api calls

somic:
PAASTA-13433

### DIFF
--- a/paasta_tools/mesos_maintenance.py
+++ b/paasta_tools/mesos_maintenance.py
@@ -47,9 +47,8 @@ def base_api():
     """
     leader = get_mesos_leader()
 
-    def execute_request(method, endpoint, **kwargs):
+    def execute_request(method, endpoint, timeout=(3, 1), **kwargs):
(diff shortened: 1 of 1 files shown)


paasta_tools/mesos_maintenance.py
```diff
@@ -106,7 +105,7 @@ def maintenance_api():
     """
     def execute_schedule_api_request(method, endpoint, **kwargs):
         master_api_client = master_api()
-        return master_api_client(method, "/maintenance%s" % endpoint, **kwargs)
+        return master_api_client(method, "/maintenance%s" % endpoint, timeout=(3, 3), **kwargs)
```
solarkennedy: This is is still a bit too optimistic. Loading the dashboard in my browser I can see that it takes ~2.3 second in the normal case, sometimes it is better, but in loaded times (like yesterday) it is going to be worse.

How about 10s?

somic:
Increased read timeout to 10s

PR #1613, opened 2017-11-07
make graceful-app-drain using marathon sharded code

oktopuz:
internal ticket: PAASTA-13420

### DIFF
--- a/paasta_tools/graceful_app_drain.py
+++ b/paasta_tools/graceful_app_drain.py
@@ -33,23 +33,8 @@ def parse_args():
     return parser.parse_args()
 
 
-def main():
-    args = parse_args()
-    full_appid = args.appname.lstrip('/')
-    soa_dir = args.soa_dir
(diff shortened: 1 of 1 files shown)


paasta_tools/graceful_app_drain.py
```diff
@@ -109,5 +94,27 @@ def main():
...
+
+    for client in clients:
+        if marathon_tools.is_app_id_running(app_id=full_appid, client=client):
+            kill_marathon_app(
+                full_appid=full_appid,
+                cluster=cluster,
+                client=client,
+                soa_dir=args.soa_dir,
+            )
+            return 0
```
solarkennedy: This code assumes that it only exists on one shard. 

Instead can you move the return out of the loop and instead count how many apps were killed?
then return non-0 if apps_killed < 1

oktopuz: Done.

PR #1614, opened 2017-11-07
Emit metrics from cluster autoscaler

### LINKED ISSUE #1600: RFC: Emit metrics from cluster autoscaler

matthewbentley:
fixes #1600


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -50,6 +50,11 @@
 from paasta_tools.utils import Timeout
 from paasta_tools.utils import TimeoutError
 
+try:
```
Rob-Johnson: can you unify this with the work mmb did in deployd to make a more general metrics_provider available so that we don't have to do this for every library? i.e. take the metrics_provider config out of deployd and make it a 'global' setting?

matthewbentley: I've create #1621 to do that

PR #1616, opened 2017-11-08
RFC: Add cluster autoscaler docs

matthewbentley:
I did a brain dump on the cluster autoscaler.

missing: anything about the async/await.  That should probably be mentioned.  Also will need to be updated for cluster boost.

It can probably be a bit more clear about what I mean by region✕pool, and why it works like that.

### DIFF
--- a/docs/source/cluster_autoscaling.rst
+++ b/docs/source/cluster_autoscaling.rst
@@ -83,7 +83,7 @@ Current utilization
 
 PaaSTA will first add up all cpu, disk, and memory used by tasks in the
 region✕pool, and the total cpu, disk, and memory available in the same, and uses
-this to calculate the percent utilization of cpus, disk, and mem.  THe current
+this to calculate the percent utilization of cpus, disk, and mem.  The current
 utilization is the maximum of those three.
 
 The result of this is the cluster autoscaler will scale so the most used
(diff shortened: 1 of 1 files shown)


docs/source/cluster_autoscaling.rst
```diff
@@ -0,0 +1,154 @@
...
+"utilization error".  In short, this is the difference between the current
+utilization and the target utilization (either defined in /etc/paasta or the
+default of 0.8).
+
+Current utilization
+^^^^^^^^^^^^^^^^^^^
+
+PaaSTA will first add up all cpu, disk, and memory used by tasks in the
+region✕pool, and the total cpu, disk, and memory available in the same, and uses
+this to calculate the percent utilization of cpus, disk, and mem.  THe current
```
solarkennedy: The current

docs/source/cluster_autoscaling.rst
```diff
@@ -0,0 +1,154 @@
...
+utilization error is positive, and the region✕pool must scale up.  For this,
+the cluster autoscaler will loop through each resource (ASG or SFR) in this
+region✕pool, and set the target capacity to
+``ceil(current_capacity * (1+utilization_error))``.
+
+Scaling down
+------------
+
+Scaling down happens when the error is negative.  It takes more work as we
+cannot just terminate enough instances to get to the target capaicty without the
```
solarkennedy: capacity, use :set spell or some sort of spell check

PR #1617, opened 2017-11-08
Make PaaSTA API work with sharded marathon

oktopuz:
internal ticket: PAASTA-13419

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -18,7 +18,6 @@ make the PaaSTA stack work.
 """
 import copy
 import datetime
-import itertools
 import json
 import logging
 import os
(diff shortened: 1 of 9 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -897,11 +896,7 @@ def get_list_of_marathon_clients(
     if system_paasta_config is None:
         system_paasta_config = load_system_paasta_config()
     marathon_servers = get_marathon_servers(system_paasta_config)
-    marathon_clients = get_marathon_clients(marathon_servers, cached=cached)
-    return [c for c in itertools.chain(
-        marathon_clients.current,
-        marathon_clients.previous,
-    )]
+    return get_marathon_clients(marathon_servers, cached=cached).get_all_clients()
```
solarkennedy: Thank you for unifying this

solarkennedy:
I'm OK with this API not querying all clients all the time. 99% of the time there will only be one client to query.

PR #1621, opened 2017-11-09
Move deployd metrics to be generic

matthewbentley:
Also add typing for metrics

This is made to address @Rob-Johnson's comment on #1614 

It could probably use some more tests


paasta_tools/metrics/metrics_lib.py
```diff
@@ -0,0 +1,102 @@
...
+        self.name = name
+
+    def start(self) -> None:
+        log.debug("timer {} start at {}".format(self.name, time.time()))
+
+    def stop(self) -> None:
+        log.debug("timer {} stop at {}".format(self.name, time.time()))
+
+
+class Gauge:
```
Rob-Johnson: can you move this and the Timer into a noop_metrics.py module?

paasta_tools/metrics/metrics_lib.py
```diff
@@ -0,0 +1,102 @@
...
+    def stop(self) -> None:
+        ...
+
+
+class GaugeProtocol(Protocol):
+    def set(self, value: Union[int, float]) -> None:
+        ...
+
+
+class BaseMetrics(Protocol):
```
Rob-Johnson: can you make this a real abstract base class? https://docs.python.org/3/library/abc.html

paasta_tools/metrics/metrics_lib.py
```diff
@@ -0,0 +1,74 @@
...
+
+_metrics_interfaces: Dict[str, Type['BaseMetrics']] = {}
+
+
+class TimerProtocol(Protocol):
+    def start(self) -> None:
+        ...
+
+    def stop(self) -> None:
+        ...
```
Rob-Johnson: change these to raise a NotImplementedError

PR #1622, opened 2017-11-10
Add feature to manually pause the autoscaler

fboxwala:


fboxwala:
I still need to add a test but I wanted to force a travis run for reasons

fboxwala:
wafflebot I will fight you

Rob-Johnson:
rather than using the ssh rpc mechanism - can you make this happen with the HTTP API, and set the minutes in the body of a json string that you POST?

fboxwala:
So I'm not sure where we landed on this because of call quality. Do we think that pausing the autoscaler is a safe enough task for the API? @EvanKrall

Rob-Johnson:
@EvanKrall is right - we don't have any auth mechanisms on the API, so anyone could potentially pause the autoscaler. I don't think this is /that/ scary, and I'd be willing to take the risk in return for reducing the amount of code debt we're adding with more ssh-rpc.
 
We do also have precedent for this (unauthenticated RW)- there is a [RW endpoint](https://github.com/Yelp/paasta/blob/master/paasta_tools/api/views/autoscaler.py#L49-L86) allowing you to modify the desired instances for a service already.

paasta_tools/api/api.py
```diff
@@ -76,6 +76,8 @@ def make_app(global_config=None):
     config.add_route('service.list', '/v1/services/{service}')
     config.add_route('service.autoscaler.get', '/v1/services/{service}/{instance}/autoscaler', request_method="GET")
     config.add_route('service.autoscaler.post', '/v1/services/{service}/{instance}/autoscaler', request_method="POST")
+    config.add_route('pause_autoscaler.post', '/v1/pauseAutoscaler', request_method="POST")
```
matthewbentley: Can we somehow specify that this is the service autoscaler?

paasta_tools/api/api_docs/swagger.json
```diff
@@ -29,6 +29,49 @@
         "operationId": "showVersion"
       }
     },
+    "/pauseAutoscaler": {
+      "get": {
+        "responses": {
+          "200": {
+            "description": "The time until which the autoscaler is paused for",
```
matthewbentley: service autoscaler

paasta_tools/api/api.py
```diff
@@ -76,6 +76,8 @@ def make_app(global_config=None):
     config.add_route('service.list', '/v1/services/{service}')
     config.add_route('service.autoscaler.get', '/v1/services/{service}/{instance}/autoscaler', request_method="GET")
     config.add_route('service.autoscaler.post', '/v1/services/{service}/{instance}/autoscaler', request_method="POST")
+    config.add_route('pause_service_autoscaler.post', '/v1/pauseServiceAutoscaler', request_method="POST")
+    config.add_route('pause_service_autoscaler.get', '/v1/pauseServiceAutoscaler', request_method="GET")
```
Rob-Johnson: eh, can this be ``/service-autoscaler/pause`` ? sorry for nitpicking

paasta_tools/cli/utils.py
```diff
@@ -596,6 +596,29 @@ def execute_paasta_metastatus_on_remote_master(
     )
 
 
+def get_service_autoscale_pause_time(cluster):
```
Rob-Johnson: I think ``paasta_tools/autoscaling/autoscaling_service_lib.py`` would be a better home for these functions

fboxwala: It would take me a decent amount of time to make this change and I'm not 100% convinced the autoscaling service lib is the place right place for them. I would rather keep these functions in this file since they're only being used by the command line tool, and I'm not sure where else we would need to use them again anyway. Currently I believe the autoscaling service lib file only contains functions used by the autoscaler which these are not.

paasta_tools/api/client.py
```diff
@@ -59,4 +59,10 @@ def get_paasta_api_client(cluster=None, system_paasta_config=None):
         spec_dict = json.load(f)
     # replace localhost in swagger.json with actual api server
     spec_dict['host'] = api_server
-    return SwaggerClient.from_spec(spec_dict=spec_dict)
+
+    # sometimes we want the status code
+    if http_res:
+        config = {'also_return_response': True}
```
solarkennedy: Why not just do this all the time?

paasta_tools/api/views/pause_autoscaler.py
```diff
@@ -0,0 +1,55 @@
...
+            pause_until = zk.get(zk_pause_autoscale_path)[0].decode('utf8')
+        except (NoNodeError, ValueError):
+            pause_until = '0'
+        except Exception as e:
+            raise ApiFailure(e, 500)
+
+    return pause_until
+
+
+@view_config(route_name='pause_service_autoscaler.post', request_method='POST', renderer='json')
```
solarkennedy: Can you add a DELETE method to unpause? I think that would be very resty and less weird than posting 0 minutes.

paasta_tools/cli/utils.py
solarkennedy: Or how about just pause_service_autoscaler.py ?

paasta_tools/cli/cmds/pause_service_autoscaler.py
```diff
@@ -0,0 +1,108 @@
...
+        minutes = '0'
+
+    if args.info:
+        retval = get_service_autoscale_pause_time(args.cluster)
+    else:
+        retval = update_service_autoscale_pause_time(args.cluster, minutes)
+
+    if retval == 1:
+        paasta_print('Could not connect to paasta api. Maybe you misspelled the cluster?')
+        return 1
```
solarkennedy: I don't like how these return codes are "hard coded". 
What if it returned whatever it returned, but made the get/update functions print the exact error message? Instead of having the cli having to interpret the return code (and possibly get out of sync)

paasta_tools/api/client.py
fboxwala: If I do this all the time I have to update any call to the client that's already being made to accept a two-ple. This seemed easier, but I can go through and do that.

paasta_tools/pause_service_autoscaler.py
```diff
@@ -0,0 +1,54 @@
+import time
```
Rob-Johnson: can you move this into the autoscaling ``paasta_tools/autoscaling`` directory?

paasta_tools/pause_service_autoscaler.py
```diff
@@ -0,0 +1,54 @@
...
+    pause_time, http = api.service_autoscaler.get_service_autoscaler_pause().result()
+    if http.status_code == 500:
+        paasta_print('Could not connect to zookeeper server')
+        return 2
+
+    pause_time = int(pause_time)
+    if pause_time < time.time():
+        paasta_print('Service autoscaler is not paused')
+    else:
+        paused_readable = datetime.fromtimestamp(pause_time).strftime('%H:%M:%S %Y-%m-%d')
```
Rob-Johnson: can you make this iso8601 format with a timezone attached to it please?

paasta_tools/api/api.py
```diff
@@ -76,6 +76,9 @@ def make_app(global_config=None):
     config.add_route('service.list', '/v1/services/{service}')
     config.add_route('service.autoscaler.get', '/v1/services/{service}/{instance}/autoscaler', request_method="GET")
     config.add_route('service.autoscaler.post', '/v1/services/{service}/{instance}/autoscaler', request_method="POST")
+    config.add_route('service_autoscaler.pause.post', '/v1/service_austoscaler/pause', request_method="POST")
+    config.add_route('service_autoscaler.pause.delete', '/v1/service_austoscaler/pause', request_method="DELETE")
```
Rob-Johnson: typo's here in service_autoscaler

paasta_tools/api/views/pause_autoscaler.py
```diff
@@ -0,0 +1,67 @@
...
+            zk.set(zk_pause_autoscale_path, str(expiry_time).encode('utf-8'))
+        except Exception as e:
+            raise ApiFailure(e, 500)
+
+    return
+
+
+@view_config(route_name='service_autoscaler.pause.delete', request_method='DELETE', renderer='json')
+def delete_service_autoscaler_pause(request):
+    zk_pause_autoscale_path = '{}/paused'.format(AUTOSCALING_ZK_ROOT)
```
Rob-Johnson: can you create another constant for ZK_PAUSE_AUTOSCALE_PATH

paasta_tools/api/views/pause_autoscaler.py
```diff
@@ -0,0 +1,64 @@
...
+
+    return
+
+
+@view_config(route_name='service_autoscaler.pause.delete', request_method='DELETE', renderer='json')
+def delete_service_autoscaler_pause(request):
+    with ZookeeperPool() as zk:
+        try:
+            zk.ensure_path(ZK_PAUSE_AUTOSCALE_PATH)
+            zk.set(ZK_PAUSE_AUTOSCALE_PATH, str(time.time()))
```
solarkennedy: Why not just delete?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -695,8 +696,26 @@ def get_configs_of_services_to_scale(cluster, soa_dir=DEFAULT_SOA_DIR):
...
+    with ZookeeperPool() as zk:
+        try:
+            pause_until = zk.get(ZK_PAUSE_AUTOSCALE_PATH)[0].decode('utf8')
+            pause_until = int(pause_until)
+        except (NoNodeError, ValueError) as e:
+            pause_until = 0
+
+    if time.time() < pause_until:
+        return False
+    else:
```
solarkennedy: can you log.debug("Autoscaling is paused for another N minutes")

PR #1623, opened 2017-11-10
Always consider num_at_risk_tasks when calculating instance_counts du…

solarkennedy:
…ring a bounce


mattmb:
Can you explain this change? Doesn't this mean we might scale down tasks that have been added when we're scaled up because a host is maint'd?

PR #1627, opened 2017-11-14
Tweak mypy settings to allow for more granular type checking

EvanKrall:
This checks every file in paasta_tools, but relaxes the requirement that every function pass type checking: only functions with parameter/return type annotations will have their internals checked.

I also went ahead and added `disallow_untyped_defs = True` for generate_deployments_for_service, since it was almost there anyway.


paasta_tools/generate_deployments_for_service.py
```diff
@@ -274,40 +280,22 @@ def get_desired_state(branch: str, remote_refs, deploy_group: str) -> Tuple[str,
...
         with open(os.path.join(soa_dir, service, TARGET_FILE), 'r') as oldf:
             old_deployments_dict = json.load(oldf)
-            old_mappings = get_deploy_group_mappings_from_deployments_dict(old_deployments_dict)
     except (IOError, ValueError) as e:
-        old_mappings = {}
         old_deployments_dict = {}
     mappings, v2_mappings = get_deploy_group_mappings(
         soa_dir=soa_dir,
         service=service,
-        old_mappings=old_mappings,
```
solarkennedy: With no mappings, we will be touching this file all the time, right?

EvanKrall: This variable is entirely unused in `get_deploy_group_mappings`.

EvanKrall: We have a check immediately after to see whether the old file is the same as the new file:

```
    if deployments_dict != old_deployments_dict:
```

PR #1629, opened 2017-11-16
don't undrain unhappy tasks

somic:
In logs I came across a situation when we first get an exception during is_draining and mark task 'unhappy':

```Ignoring exception during is_draining of task MarathonTask... Treating task as 'unhappy'```

But then we try to undrain it in undrain_tasks:

```Ignoring exception during stop_draining of task MarathonTask```

This fixes it.

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -593,6 +593,7 @@ def deploy_service(
     )
 
     num_at_risk_tasks = 0
+    scaling_app_unhappy_tasks = None
     if new_app_running:
         num_at_risk_tasks = get_num_at_risk_tasks(new_app, draining_hosts=draining_hosts)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: paasta_tools/setup_marathon_job.py [hunk 1]
```diff
@@ -647,8 +648,12 @@ def log_deploy_error(errormsg: str, level: str='event') -> None:
                 new_client.scale_app(app_id=new_app.id, instances=(new_app.instances - slack), force=True)
 
         # TODO: don't take actions in deploy_service.
+        if scaling_app_unhappy_tasks is None:
```
solarkennedy: I don't think we should do this. 
A task can be unhappy because of the infrastructure *or* the app, if we need to undrain we should undrain.

somic: OK, discarding

PR #1634, opened 2017-11-22
fix bug in autoscaler for sharded marathon

Rob-Johnson:
dupe of https://github.com/Yelp/paasta/commit/7873bdffad718f12425f840aca94c418bf1a8e5e
in another place.


solarkennedy:
I'm worried that we are going to continue making this mistake.

matthewbentley:
Isn't this the kind of thing type checking should catch?

PR #1636, opened 2017-11-23
Show marathon dashboard links in paasta status

oktopuz:
internal ticket: PAASTA-13069


paasta_tools/marathon_tools.py
```diff
@@ -1524,3 +1524,19 @@ def take_up_slack(client: MarathonClient, app: MarathonApp) -> None:
...
+    try:
+        links = system_paasta_config.get_dashboard_links().get(cluster).get('Marathon RO')
+    except KeyError:
+        links = ''
+    if isinstance(links, list):
+        pinned_shard = job_config.get_marathon_shard()
+        if pinned_shard:
+            return links[pinned_shard]
+        else:
+            service_instance = compose_job_id(job_config.service, job_config.instance)
```
solarkennedy: I'm really concerned we have to do this in a second place. 
I think ideally we would only ever have the key= setting for rendezvous_hash in a single place ever.

I expected this code to look like this:

shard = lookup_shard(job_config)
return links[shard]

def lookup_shard(job_config):
... I guess this needs some sort of "choises".

...
```

I guess what I don't like is the possibility that this shard lookup can be "out of sync" with that the actual shard is.

This is not a blocker.

File: paasta_tools/marathon_tools.py [hunk 2]
```diff
@@ -1524,3 +1524,19 @@ def take_up_slack(client: MarathonClient, app: MarathonApp) -> None:
...
+def marathon_dashboard_base_url(job_config, system_paasta_config):
+    cluster = system_paasta_config.get_cluster()
+    try:
+        links = system_paasta_config.get_dashboard_links().get(cluster).get('Marathon RO')
+    except KeyError:
+        links = ''
+    if isinstance(links, list):
+        pinned_shard = job_config.get_marathon_shard()
+        if pinned_shard:
+            return links[pinned_shard]
```
solarkennedy: This is dangerous. Can you protect the code against errors where we haven't updated the links and there isn't a corresponding entry for it?

paasta_tools/marathon_serviceinit.py
```diff
@@ -165,12 +165,16 @@ def status_marathon_job(service, instance, app_id, normal_instance_count, client
...
 
 
-def get_verbose_status_of_marathon_app(marathon_client, app, service, instance, cluster, soa_dir):
+def get_verbose_status_of_marathon_app(marathon_client, app, job_config, system_paasta_config):
     """Takes a given marathon app object and returns the verbose details
     about the tasks, times, hosts, etc"""
     output = []
     create_datetime = datetime_from_utc_to_local(isodate.parse_datetime(app.version))
-    output.append("  Marathon app ID: %s" % PaastaColors.bold(app.id))
+    url = ("%s/ui/#/apps/%2F%s" % (
```
solarkennedy: Can you split this into a function, something like compose_dashboard_url(base, appid)

paasta_tools/marathon_serviceinit.py
```diff
@@ -422,6 +432,18 @@ def status_mesos_tasks(service, instance, normal_instance_count):
...
 
 
+def get_marathon_dashboard_links(marathon_clients, system_paasta_config):
+    """Return a dict of marathon clients and their corresponding dashboard URLs"""
+    cluster = system_paasta_config.get_cluster()
+    try:
+        links = system_paasta_config.get_dashboard_links().get(cluster).get('Marathon RO')
+    except KeyError:
+        pass
+    if isinstance(links, list) and len(links) >= len(marathon_clients.current):
```
solarkennedy: Can this tolerate the situation where there is a smaller number of clients (2 shards) yet 3 dashboards defined?

paasta_tools/marathon_serviceinit.py
```diff
@@ -165,12 +165,21 @@ def status_marathon_job(service, instance, app_id, normal_instance_count, client
...
 
 
-def get_verbose_status_of_marathon_app(marathon_client, app, service, instance, cluster, soa_dir):
+def get_marathon_dashboard(client, dashboards, app_id):
+    if dashboards is not None:
+        base_url = dashboards.get(client)
+        if base_url:
+            url = "{}/ui/#/apps/%2F{}".format(base_url.rstrip('/'), app_id.lstrip('/'))
+            return "  Marathon dashboard: %s" % PaastaColors.blue(url)
+    return PaastaColors.red("Please update Marathon dashboard URLs in /etc/paasta/dashboard_links.json")
```
solarkennedy: please make this less cary and just print the appid instead

paasta_tools/marathon_serviceinit.py
oktopuz: Yes, it can. [This assertion](https://github.com/Yelp/paasta/pull/1636/files#diff-f141232240ec690b401ca7d7aec03136R1180) tests such situation.

PR #1637, opened 2017-11-23
RFC: create spot_bid_creator script

matthewbentley:
This takes a csv from https://ec2instances.info, and returns a dict in then format:
```
{
  "instance_data": [instance-data],
}
```

where instance-data is in the format:
```
{
    "ebs_size": 60,
    "price": 12.8,
    "type": "x1e.32xlarge",
    "weight": 1.28
}
```

This is useful for creating instance data definitions for aws spot fleet.

It needs to be organized a bit better, and maybe a better name.

This also hard-codes some assumptions about bids, like we should bid 2 times the on demand price (unless the instance is too expensive, then we bid the price per core as set on the command line).

TODO: add a region flag, check whether instances are available on spot in that region.

also adds a script to create graphs like this: 
us-west-1-spot to analyze the spot market.

Also TODO: don't use the ec2instances.info info, instead get it from the api, and then combine the functionality of the 2 scripts (so we can analyze what vcpu/mem/ecu combinations would look like over the past n days).

### DIFF
--- a/paasta_tools/contrib/spot_bid_creator.py
+++ b/paasta_tools/contrib/spot_bid_creator.py
@@ -1,0 +1,193 @@
+#!/usr/bin/env python
+import argparse
+import csv
+import json
+
+
+def parse_ecu_per_vcpu(s):
+    s = s.strip(' units')
+    try:
+        s = float(s)
+    except ValueError:
+        s = 0
+    return s
+
+
+def parse_instance_storage(s):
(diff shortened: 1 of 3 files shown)


sarguru:
spot_bid_graph.py is not meaningful anymore with recent changes from amazon. We can cut it out.

Rob-Johnson:
RIP

PR #1638, opened 2017-11-28
Don't docker-compose pull in example_cluster; see bc3029fb3f01eadb931105eea480c9ec21c18ea8

EvanKrall:


EvanKrall:
Ok, well, example_cluster fails for a different reason now. I'm going to merge this.

PR #1640, opened 2017-12-01
Fix paused autoscaler logic bug

solarkennedy:
I think the logic was just backwards?


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -701,15 +701,15 @@ def autoscaling_is_paused():
...
-        except (NoNodeError, ValueError) as e:
+        except (NoNodeError, ValueError, AttributeError) as e:
             pause_until = 0
 
     remaining = pause_until - time.time()
     if remaining >= 0:
-        return False
-    else:
-        log.debug("Autoscaling is paused for {} minutes".format(str(remaining)))
+        log.debug("Autoscaling is paused for {} more minutes".format(str(remaining)))
```
nhandler: I think this should be `seconds`

PR #1641, opened 2017-12-01
pytest.fixture for SystemPaastaConfig 

oktopuz:


tests/cli/test_cmds_local_run.py
```diff
@@ -1615,16 +1596,7 @@ def test_missing_volumes_skipped(mock_exists):
...
-                {
-                    'volumes': [{
-                        "hostPath": "/hostPath",
-                        "containerPath": "/containerPath",
-                        "mode": "RO",
-                    }],
-                },
-                '/etc/paasta',
-            ),
+            system_paasta_config=system_paasta_config,
```
EvanKrall: I think most of the other changes are fine (replacing e.g. `SystemPaastaConfig({}, '/fake/config')` with the fixture), but this test cares specifically about the contents of SystemPaastaConfig, so I think it doesn't make sense to use the fixture here.

tests/cli/test_cmds_local_run.py
```diff
@@ -1563,16 +1553,7 @@ def test_volumes_are_deduped(mock_exists):
...
```
EvanKrall: see below

tests/cli/test_cmds_local_run.py
oktopuz: I reverted both changes.

PR #1642, opened 2017-12-01
Make metastatus able to group by multiple values properly

EvanKrall:
This fixes something that's been bothering me for a while, which is that there's no easy way to see resource utilization for a pool broken down by region. You currently either get region or pool, but can't distinguish default/uswest1-prod and default/sfo12-prod.

Example output: https://fluffy.yelpcorp.com/i/tdFCpRCrnWRHpF0p1HQLtmXhXthj5VSg.html

One thing I expect to be somewhat controversial, and am not very tied to: This adds the groupings (region and pool by default) before the hostname in the `-vvv` output. I think this will be useful, as currently with `-vvv` it's impossible to know what pool a host is in.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -1267,7 +1267,7 @@ def get_mesos_utilization_error(
         log.info("Failed to find utilization for region %s, pool %s, returning 0 error")
         return 0
 
-    log.debug(region_pool_utilization_dict)
+    log.debug(repr(region_pool_utilization_dict))
     free_pool_resources = region_pool_utilization_dict['free']
     total_pool_resources = region_pool_utilization_dict['total']
     free_percs = []
--- a/paasta_tools/metrics/metastatus_lib.py
+++ b/paasta_tools/metrics/metastatus_lib.py
@@ -18,10 +18,16 @@ import math
 from collections import Counter
 from collections import namedtuple
 from collections import OrderedDict
+from typing import Callable
+from typing import Collection
 from typing import Dict
 from typing import List
+from typing import Tuple
+from typing import TypeVar
 
 from humanize import naturalsize
+from mypy_extensions import TypedDict
+from typing_extensions import Counter as _Counter  # noqa
 
 from paasta_tools import chronos_tools
 from paasta_tools.mesos_maintenance import MAINTENANCE_ROLE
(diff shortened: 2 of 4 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -1267,7 +1267,7 @@ def get_mesos_utilization_error(
         log.info("Failed to find utilization for region %s, pool %s, returning 0 error")
         return 0
 
-    log.debug(region_pool_utilization_dict)
+    log.debug(repr(region_pool_utilization_dict))
```
EvanKrall: this was just to make mypy happy - mypy thinks log.debug wants a string, though it's actually happy with anything repr'able.

paasta_tools/metrics/metastatus_lib.py
```diff
@@ -347,15 +353,26 @@ def key_func(slave):
...
+) -> Callable[
+    ...,
+    Tuple[Tuple[str, str], ...],
+]:
     """ Return a closure that given a slave, will return the value of a list of
-    attributes, compiled into a hashable frozenset
+    attributes, compiled into a hashable tuple
 
-    :param attributes: the attribues to inspect in the slave
+    :param attributes: the attribues to inspect in the slaves
```
EvanKrall: oops

paasta_tools/metrics/metastatus_lib.py
```diff
@@ -718,7 +758,7 @@ def format_row_for_resource_utilization_healthchecks(healthcheck_utilization_pai
     ]
 
 
-def get_table_rows_for_resource_info_dict(attribute_value, healthcheck_utilization_pairs, humanize):
+def get_table_rows_for_resource_info_dict(attribute_values: List[str], healthcheck_utilization_pairs, humanize):
```
EvanKrall: I guess I should either finish annotating this or undo this.

paasta_tools/paasta_metastatus.py
```diff
@@ -128,7 +184,7 @@ def main(argv=None):
     except MasterNotAvailableException as e:
         # if we can't connect to master at all,
         # then bomb out early
-        paasta_print(PaastaColors.red("CRITICAL:  %s" % e.message))
+        paasta_print(PaastaColors.red("CRITICAL:  %s" % '\n'.join(e.args)))
```
EvanKrall: also to make mypy happy -- BaseException doesn't have a .message attribute in py3: https://stackoverflow.com/a/13063924

PR #1649, opened 2017-12-06
Perfscript

fboxwala:


debian/paasta-tools.links
```diff
@@ -48,4 +48,5 @@ opt/venvs/paasta-tools/bin/paasta_setup_chronos_job usr/bin/setup_chronos_job
 opt/venvs/paasta-tools/bin/paasta_tabcomplete.sh etc/bash_completion.d/paasta.bash
 opt/venvs/paasta-tools/bin/paasta usr/bin/paasta
 opt/venvs/paasta-tools/bin/setup_marathon_job.py usr/bin/setup_marathon_job
+opt/venvs/paasta-tools/bin/update_soa_memcpu.py usr/bin/update_soa_memcpu
```
solarkennedy: can you put paasta_ in the name somewhere to make it not clutter the system path. Do it in setup.py as well


Also please keep this sorted

paasta_tools/contrib/update_soa_memcpu.py
```diff
@@ -0,0 +1,164 @@
...
+def in_tempdir():
+    with tempdir() as tmp:
+        with cwd(tmp):
+            yield
+
+
+def clone(branch_name):
+    print('cloning')
+    remote = 'git@sysgit.yelpcorp.com:yelpsoa-configs'
+    subprocess.check_call(('chronic', 'git', 'clone', remote, '.'))
```
solarkennedy: I would prefer you not use chronic. For this kind of tool I think it is ok to just say what it is doing

solarkennedy:
Needs a setup.py change

debian/paasta-tools.links
fboxwala: Since there is already this line: https://github.com/Yelp/paasta/blob/master/setup.py#L103

Do I still need to add a line to `console_scripts` ?

solarkennedy:
Hmm. Scripts with the glob is ok.

Issue #335, opened 2016-03-17
Apply a default constraint that prevents all instances from being deployed to one box


EvanKrall:
In the mean time, for small services (where # instances <<  # hosts), we should apply a `["hostname", "UNIQUE"]` constraint. For larger services we should consider `['hostname', "GROUP_BY", 5]` or something, which would prevent more than ~20% of instances from ending up on one box (albeit with the issues listed in mesosphere/marathon#3515).

If https://github.com/mesosphere/marathon/issues/3515 gets implemented, we should switch to that, based on a percentage of the instances or something.

Rob-Johnson:
I had a think about this today and I'd like to get started  - here are some thoughts/questions:

- We'd have to take other constraints into account when calculating the number of slaves available -  attributes are made available in mesos' ``state.json``, but it's worth remembering.
- Currently we **don't** bounce when the instance count in soa-configs changes, but this wouldn't be the case if the instance count crossed the threshold from < number of slaves to >= number of slaves. Is this OK?
- If we add the ``['hostname', "GROUP_BY", n]`` constraint, where n is ``instances``/``slaves`` then we'd bounce every time we lost a slave. This is probably too much churn, and we need to come up with a way for it to be less sensitive.

mittonk:
I had 40% of my instances end up on one box, which then slowed down.  There's 10+ available hosts, better distribution would be helpful here.

solarkennedy:
I don't like the idea of applying this constraint automatically, what if we added it to the defaults in the fsm template? (assuming I ship my comment'd fsm template idea) Like this:

```
# marathon-PROD.yaml
main:
  # This is a starting instance count, as your service gets more traffic, please
  # check out $autoscaling docs to see how to scale the instances with load
  instances: 3
  # This extra constraint ensures the service instances don't all get scheduled to the
  # the same box. It is in general a good idea, but needs to be removed the service
  # has more than say, 3 instances. For example if the service is scaled up to 100
  # instance and we don't have 100 unique hostnames, it won't work.
  extra_constraints: ["hostname", "UNIQUE"]
```

Rob-Johnson:
@solarkennedy yeah, it's good enough until mesosphere/marathon#3515 makes progress - my only objection is that 3 is a bit of a magic number - I'd like to give users a way of figuring out what the actual limit of this is.

solarkennedy:
There is no firm actual determined limit of course, it is a function of how many slaves there are and the precise distribution of tasks at that second.

[...]

PR #1652, opened 2017-12-08
RFC: Automatically add ['hostname', 'UNIQUE'] to small services

nhandler:
We briefly discussed the idea of automatically adding a `['hostname', 'UNIQUE']` constraint to small services running in production. This would avoid situations where all of their tasks end up running on the same agent, which might then go away and cause replication issues.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -549,9 +549,7 @@ class MarathonServiceConfig(LongRunningServiceConfig):
         """
         if 'prod' in self.cluster:
             auto_hostname_unique_size = system_paasta_config.get_auto_hostname_unique_size()
-            app_size = self.get_max_instances()
-            if app_size is None:
-                app_size = self.get_desired_instances()
+            app_size = self.get_max_instances() or self.get_desired_instances() 
             if app_size <= auto_hostname_unique_size:
                 return [["hostname", "UNIQUE"]]
         return []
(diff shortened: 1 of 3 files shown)


solarkennedy:
Much of the comments are already documented in #335

paasta_tools/marathon_tools.py
```diff
@@ -531,8 +531,31 @@ def get_calculated_constraints(
...
+        in a production cluster automatically receive a hostname UNIQUE constraint to reduce
+        the risk of all tasks getting launched on the same agent, which might then be lost.
+
+        :param system_paasta_config: A SystemPaastaConfig object representing the system
+                                 configuration.
+        :returns: a set of constraints for marathon
+        """
+        if 'prod' in self.cluster:
+            auto_hostname_unique_size = system_paasta_config.get_auto_hostname_unique_size()
+            app_size = self.get_max_instances()
```
solarkennedy: How about 
```
app_size = self.get_max_instances() or self.get_desired_instances() 
```
?

paasta_tools/utils.py
```diff
@@ -1466,6 +1467,16 @@ def get_cluster(self) -> str:
...
 
+    def get_auto_hostname_unique_size(self) -> int:
+        """
+        We automatically add a ["hostname", "UNIQUE"] constraint to "small" services running in production clusters.
+        If there are less than or equal to this number of instances, we consider it small.
+        We fail safe and return 0 to avoid adding the ['hostname', 'UNIQUE'] constraint if this value is not defined
+
+        :returns: The integer size of a small service
+        """
+        return self.config_dict.get('auto_hostname_unique_size', 0)
```
solarkennedy: This needs to be -1 actually, because otherwise when someone "disables" a service it has to bounce to get a new constraint. I would rather it just scale from 0-1 without a bounce.

PR #1653, opened 2017-12-08
Convert drain_lib to asyncio so we can drain in parallel.

EvanKrall:


paasta_itests/environment.py
```diff
@@ -11,13 +11,13 @@
...
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import asyncio
 import os
 import shutil
 import time
 
 import mock
 import requests
-from behave_pytest.hook import install_pytest_asserts
```
EvanKrall: pytest-asyncio needed an upgrade of pytest to >=3.x; pytest 3.x does assertion rewriting differently than 2.x, which breaks behave-pytest: https://github.com/ribozz/behave-pytest/issues/2

paasta_tools/setup_marathon_job.py
```diff
@@ -471,15 +491,22 @@ def undrain_tasks(
...
-                drain_method.stop_draining(task)
+                await drain_method.stop_draining(task)
             except Exception as e:
                 log_deploy_error("Ignoring exception during stop_draining of task %s: %s." % (task, e))
 
+    undrain_task_futures = []
+    for task in to_undrain:
+        undrain_task_futures.append(asyncio.ensure_future(undrain_task(task)))
+    if undrain_task_futures:
+        asyncio.get_event_loop().run_until_complete(asyncio.wait(undrain_task_futures))
```
EvanKrall: I've written some variation of these five lines several times now, and I feel like there must be a cleaner way to write this.

setup.py
```diff
@@ -31,6 +31,7 @@
     include_package_data=True,
     install_requires=[
         'argcomplete >= 0.8.1',
+        'aiohttp',
```
EvanKrall: todo: pin this

paasta_tools/setup_marathon_job.py
henryzhangsta: Unfortunately when you are mixing async and non-async code you have to do this. You could use a map here

```
asyncio.get_event_loop().run_until_complete(asyncio.wait([
    asyncio.ensure_future(undrain_task(task)) for task in to_undrain
]))
```

If you don't care about the return value you could run this event loop in another thread

paasta_tools/setup_marathon_job.py
```diff
@@ -168,21 +169,27 @@ def drain_tasks_and_find_tasks_to_kill(
...
                 log_bounce_action(
                     line=("%s bounce killing task %s due to exception when draining: %s" % (bounce_method, task.id, e)),
                 )
                 tasks_to_kill.add((task, client))
 
-    for task, client in all_draining_tasks:
+        if tasks_to_drain:
+            asyncio.get_event_loop().run_until_complete(asyncio.wait([
+                asyncio.ensure_future(drain_and_kill_if_draining_fails(task, client)) for task, client in tasks_to_drain
+            ]))
```
smallredbean: [nit] Totally understand why doing this on one line, but this is a loooong line, impairing code readability (same to the code segment on lines 205-208).

EvanKrall: Yeah, it's unfortunate. I've never found a way I like of splitting list comprehensions onto multiple lines. In a previous commit, I had done it by appending to an empty list, and I think this way is slightly more readable.

PR #1655, opened 2017-12-11
Make setup_marathon_job ignore instances starting with _

### LINKED ISSUE #1226: Make it possible to have sections of marathon/chronos/adhoc yaml files that aren't treated as instances
YAML has the neat anchor/alias features, which allows for composition of dictionaries together.

However, every top-level entry in a marathon/chronos/adhoc.yaml is treated as a service instance, which means that there's no way to create an instance "template" that isn't used directly -- we can only reuse configs of other instances.

This can sometimes lead to surprising results. As an example, the following is a pretty standard use of anchors to reduce config duplication:

```yaml
main: &main_template
  cpus: 2
  mem: 5000
  instances: 30
  deploy_group: prod.main
  registrations: ['my-cool-service.main']
  env:
    ...

canary:
  <<: *main_template
  instances: 2
  deploy_group: prod.canary
```

If we now want to autoscale the main instance, we'd replace `instances` with `min_instances` and `max_instances` and maybe some `autoscaling` params:

```yaml
main: &main_template
  cpus: 2
  mem: 5000
  min_instances: 5
  max_instances: 40
  deploy_group: prod.main
  env:
    ...

canary:
  <<: *main_template
  instances: 2
  deploy_group: prod.canary
```

Except, oops! We just enabled autoscaling on canary, because canary does not override min_instances/max_instances.

We could flip it around, so that canary becomes the template, and main inherits from it, and that would solve this particular problem. However, we have some services that use `bounce_method: brutal` on canary to speed up deployments. Brutal is fine for canary, which represents a small fraction of capacity, but it would not be okay for main. We could solve this by overriding `bounce_method` on main, but we'd have to remember to override this for every instance that inherits from the same reference.

# Proposal 1:

We could have a dedicated top-level key which we totally ignore, under which you could construct whatever templates you like:

```yaml
__templates__:
  main_template: &main_template
    cpus: 2
    mem: 5000
    registrations: ['my-cool-service.main']
    env:
      ...

main:
  <<: *main_template
  instances: 30
  deploy_group: prod.main

canary:
  <<: *main_template
  instances: 2
  deploy_group: prod.canary
```

# Proposal 2:

We could ignore any top-level key that matches some pattern, like if it starts with an underscore:

```yaml
_main_template: &main_template
  cpus: 2
  mem: 5000
  registrations: ['my-cool-service.main']
  env:
    ...

main:
  <<: *main_template
  instances: 30
  deploy_group: prod.main

canary:
  <<: *main_template
  instances: 2
  deploy_group: prod.canary
```

nhandler:
This is to make it easier to define a template instance without having a marathon application get created for it.

Fixes #1226

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -74,6 +74,7 @@ from paasta_tools.utils import get_config_hash
 from paasta_tools.utils import get_paasta_branch
 from paasta_tools.utils import get_service_instance_list
 from paasta_tools.utils import get_user_agent
+from paasta_tools.utils import InvalidJobNameError
 from paasta_tools.utils import load_deployments_json
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import MarathonConfigDict
@@ -83,6 +84,7 @@ from paasta_tools.utils import PaastaNotConfiguredError
 from paasta_tools.utils import SystemPaastaConfig
 from paasta_tools.utils import time_cache
 
+# Marathon creates Mesos tasks with an id composed of the app's full name, a
 # Marathon creates Mesos tasks with an id composed of the app's full name, a
 # spacer, and a UUID. This variable is that spacer. Note that we don't control
 # this spacer, i.e. you can't change it here and expect the world to change
@@ -364,6 +366,10 @@ def load_marathon_service_config_no_cache(
         soa_dir=soa_dir,
     )
 
+    if instance.startswith('_'):
+        raise InvalidJobNameError(
+            "Unable to load marathon job config for %s.%s as instance name starts with '_'" % (service, instance),
+        )
     if instance not in instance_configs:
         raise NoConfigurationForServiceError(
             "%s not found in config file %s/%s/%s.yaml." % (instance, soa_dir, service, marathon_conf_file),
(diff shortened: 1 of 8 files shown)


solarkennedy:
I agree with krall, make it ignore at a more fundamental level, not setup_marathon_job

paasta_tools/marathon_tools.py
```diff
@@ -1474,12 +1480,12 @@ def is_old_task_missing_healthchecks(task: MarathonTask, app: MarathonApp) -> bo
...
-    at-risk (Mesos Maintenance Draining) hosts.
+    at - risk(Mesos Maintenance Draining) hosts.
 
-    :param app: A marathon application
-    :param draining_hosts: A list of hostnames that are marked as draining.
+    : param app: A marathon application
+    : param draining_hosts: A list of hostnames that are marked as draining.
                            See paasta_tools.mesos_maintenance.get_draining_hosts
-    :returns: An integer representing the number of tasks running on at-risk hosts
+    : returns: An integer representing the number of tasks running on at - risk hosts
```
EvanKrall: what's up with these changes? overzealous find and replace?

nhandler:
I believe I have modified the code to match up with @EvanKrall's recommendation.

paasta_tools/utils.py
```diff
@@ -2059,7 +2059,10 @@ def get_service_instance_list_no_cache(
             soa_dir=soa_dir,
         )
         for instance in instances:
-            instance_list.append((service, instance))
+            if instance.startswith('_'):
+                log.info("Ignoring %s.%s as instance name begins with '_'." % (service, instance))
```
EvanKrall: This will get annoying fast; it'll probably make most CLI tools print this on startup. Maybe do log.debug instead, or just drop the log line.

paasta_tools/marathon_tools.py
nhandler: No idea. I'll kill it.

paasta_tools/utils.py
nhandler: I'll flip it to `log.debug` for now. I think it might prove helpful in debugging some potential issues when we start rolling this out and services start adopting it. We can kill it off later if necessary.

PR #1656, opened 2017-12-11
Reexec paasta local-run --pull with sudo when the effective uid is not 0

oktopuz:
Internal ticket: PAASTA-13582

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -862,6 +862,9 @@ def configure_and_run_docker_container(
 
 
 def paasta_local_run(args):
+    if args.action == 'pull' and os.geteuid() != 0:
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -861,7 +861,21 @@ def configure_and_run_docker_container(
     )
 
 
+def reexecute_pull_with_sudo():
+    paasta_print("Re-executing paasta local-run --pull with sudo..")
```
solarkennedy: This is good, and does the same thing as the wrapper.

However, can you do even better by adding in the `sudo -H` *exactly* where we need it, which is wherever we shell out to an actual docker command.

oktopuz: As we discussed I added "-H", made it simpler and fixed tests.

PR #1662, opened 2017-12-13
Fix tail discovery, date handling and upgrade yelp_clog

fede1024:
- Fix log tailer discovery
- Fix date handling for range logs (timezone issues + fetching today's logs)
- Upgrade to new yelp_clog version with gzip decompression working in python3

I run different tests and they all seem to return the correct results.

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -812,7 +812,7 @@ class ScribeLogReader(LogReader):
 
     def scribe_get_from_time(self, scribe_env, stream_name, start_time, end_time):
         # Scribe connection details
-        host_and_port = scribereader.get_env_scribe_host(scribe_env, False)
+        host_and_port = scribereader.get_env_scribe_host(scribe_env, tail=False)
         host = host_and_port['host']
         port = host_and_port['port']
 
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/logs.py
```diff
@@ -812,30 +812,23 @@ def filter_and_aggregate_scribe_logs(
 
     def scribe_get_from_time(self, scribe_env, stream_name, start_time, end_time):
         # Scribe connection details
-        host_and_port = scribereader.get_env_scribe_host(scribe_env, True)
+        host_and_port = scribereader.get_env_scribe_host(scribe_env, False)
```
solarkennedy: can you kwargs this to make it more obvious what the "false" is for?

paasta_tools/cli/cmds/logs.py
```diff
@@ -812,30 +812,23 @@ def filter_and_aggregate_scribe_logs(
...
-            return fake_context()
-
-        log.info("Running the equivalent of 'scribereader -e %s %s" % (scribe_env, stream_name))
-        return scribereader.get_stream_reader(stream_name, host, port, start_time, end_time)
+        # Recent logs might not be archived yet. Log warning message.
+        warning_end_time = datetime.datetime.utcnow().replace(tzinfo=pytz.utc) - datetime.timedelta(hours=4)
+        if end_time > warning_end_time:
+            log.warn("Recent logs might be incomplete. Consider tailing instead.")
+
+        # scribeerader, sadly, is not based on UTC timestamps. It uses YST
```
solarkennedy: thank you for the comment, can you fix the scribeerader typo?

paasta_tools/cli/cmds/logs.py
```diff
@@ -812,30 +812,23 @@ def filter_and_aggregate_scribe_logs(
...
+        # dates instead.
+        start_date_yst = (start_time - datetime.timedelta(hours=8)).date()
```
solarkennedy: Is this always 8 though? Doesn't it change based on yst?

paasta_tools/cli/cmds/logs.py
```diff
@@ -812,30 +812,23 @@ def filter_and_aggregate_scribe_logs(
...
+        end_date_yst = (end_time - datetime.timedelta(hours=8)).date()
+
+        log.info("Running the equivalent of 'scribereader -e %s %s --min-date %s --max-date %s" \
```
solarkennedy: isn't -e deprecated?

paasta_tools/cli/cmds/logs.py
```diff
@@ -812,30 +812,23 @@ def filter_and_aggregate_scribe_logs(
...
+                 % (scribe_env, stream_name, start_date_yst, end_date_yst))
+        return scribereader.get_stream_reader(stream_name, host, port, start_date_yst, end_date_yst)
```
solarkennedy: can you kwargs this?

fede1024: Are you referring to the log line (using format instead of %) or get_stream_reader?

paasta_tools/cli/cmds/logs.py
fede1024: It is, however paasta is using an old version of scribereader which still relies on environments. Ideally the code should be refactored and ported to a recent scribereader version.

paasta_tools/cli/cmds/logs.py
fede1024: Does YST include day light saving? I'm not sure what the best conversion here would be.

paasta_tools/cli/cmds/logs.py
solarkennedy: get_stream_reader args

paasta_tools/cli/cmds/logs.py
solarkennedy: Yea, yst is just normal PST.

PR #1663, opened 2017-12-14
Perfscript

fboxwala:
I worked with sarth on perf to get this script to do the JIRA work that was being done in the splunk saved_search before.

I have put up an example of what the script does, to see it go to 
r/265677
and
PERF-2647

### DIFF
--- a/paasta_tools/contrib/paasta_update_soa_memcpu.py
+++ b/paasta_tools/contrib/paasta_update_soa_memcpu.py
@@ -5,6 +5,7 @@ import json
 import os
 import subprocess
 import tempfile
+import time
 
 import requests
 import ruamel.yaml as yaml
@@ -28,9 +29,17 @@ def parse_args(argv):
         required=True,
     )
     parser.add_argument(
+        '-n', '--no-tick',
+        help='Do not create a JIRA ticket',
+        action='store_true',
+        dest='no_tick',
+        default=False,
+    )
+    parser.add_argument(
         '-f', '--file-splunk',
         help='Splunk csv from which to pull data. Defaults to paasta_overprovision_alerts_fired.csv',
         dest="file_splunk",
+        default='paasta_overprovision_alerts_fired.csv',
     )
     return parser.parse_args(argv)
 
@@ -90,18 +99,22 @@ def get_perf_data(creds, filename):
         serv['project'] = d['result']['project']
         services_to_update.append(serv)
 
-    return services_to_update
+    return services_to_update[0:1]
 
 
 def clone(branch_name):
-    print('cloning')
     remote = 'git@sysgit.yelpcorp.com:yelpsoa-configs'
     subprocess.check_call(('git', 'clone', remote, '.'))
     subprocess.check_call(('git', 'checkout', '-b', branch_name))
 
 
-def commit(filename, over):
-    message = 'Updating {} for {}provisioned cpu'.format(filename, over)
+def commit(filename, serv):
+    message = 'Updating {} for {}provisioned cpu from {} to {} cpus'.format(
+        filename,
+        serv['state'],
+        serv['old_cpus'],
+        serv['cpus'],
+    )
     subprocess.check_call(('git', 'add', filename))
     subprocess.check_call(('git', 'commit', '-n', '-m', message))
 
@@ -116,18 +129,11 @@ def get_reviewers(filename):
     return authors[:3]
 
 
-def review(filename, description, over):
+def review(filename, description, provisioned_state):
     reviewers = ' '.join(get_reviewers(filename))
     subprocess.check_call((
-        'git',
-        'push',
-        '--force',
-        'origin',
-        'HEAD',
-    ))
-    subprocess.check_call((
         'review-branch',
-        '--summary=automatically updating {} for {}provisioned cpu'.format(filename, over),
+        '--summary=automatically updating {} for {}provisioned cpu'.format(filename, provisioned_state),
         '--description="{}"'.format(description),
         '--reviewers', reviewers,
         '--server', 'https://reviewboard.yelpcorp.com',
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -98,66 +116,109 @@ def get_reviewers(filename):
...
+def review(filename, description, over):
     reviewers = ' '.join(get_reviewers(filename))
-    description = (
-        'This change was made automatically by the perf_update_soaconfigs script. If not reviewed'
-        ' in a week, it will be merged by the perf or paasta teams. For more context on why these'
-        ' changes are being made see: PERF-2439'
-    )
     subprocess.check_call((
-        'review-branch',
+        'git',
```
solarkennedy: This is way too dangerous. 

Does it need to push at all?

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -98,66 +116,109 @@ def get_reviewers(filename):
...
-        clone(ticket)
-        edit_soa_configs(filename, instance, mem, cpu)
-        commit(filename, memcpu)
-        review(filename)
+    services_to_update = get_perf_data(args.splunk_creds, args.file_splunk)
+
+    for serv in services_to_update:
+        filename = '{}/{}.yaml'.format(serv['service'], serv['cluster'])
+        cpus = float(serv['cpus'])
+        over = 'over'
```
solarkennedy: this is a strange variable. how about calling it "provisioned_state" ?

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -98,66 +116,109 @@ def get_reviewers(filename):
...
-
-    with in_tempdir():
-        clone(ticket)
-        edit_soa_configs(filename, instance, mem, cpu)
-        commit(filename, memcpu)
-        review(filename)
+    services_to_update = get_perf_data(args.splunk_creds, args.file_splunk)
+
+    for serv in services_to_update:
+        filename = '{}/{}.yaml'.format(serv['service'], serv['cluster'])
```
solarkennedy: is `marathon` not in here?

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -98,66 +116,109 @@ def get_reviewers(filename):
...
+            c=serv['cluster'],
+            i=serv['instance'],
+            o=over,
+            d=serv['date'],
+            n=serv['owner'],
+            m=serv['money'],
+        )
+        ticket = create_jira_ticket(serv, args.jira_creds, ticket_desc)
+        with in_tempdir():
+            clone(ticket)
```
solarkennedy: Is it necessary to clone on its own? 

How about this script work on an existing repo, and just assume the repo is ready, either from jenkins or a human?

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -98,66 +116,109 @@ def get_reviewers(filename):
...
+        over = 'over'
+        if cpus > float(serv['old_cpus']):
+            over = 'under'
+
+        serv['over'] = over
+        ticket_desc = (
+            "We suspect that {s} {c} {i} may be {o}-provisioned"
+            " as of {d}.\n- Dashboard: y/{o}provisioned\n- Service"
+            " owner: {n}\n- Estimated monthly excess cost: {m}"
+            "\n- Runbook: y/rb-provisioning-alert"
```
solarkennedy: This needs to have maximum context. Can you add the cpu data too? 

    it originally had N cpus, but we think it only needs to have M cpus. Here is a graph to the dashboard:

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -98,66 +116,109 @@ def get_reviewers(filename):
...
+        ).format(
+            s=serv['service'],
+            c=serv['cluster'],
+            i=serv['instance'],
+            o=over,
+            d=serv['date'],
+            n=serv['owner'],
+            m=serv['money'],
+        )
+        ticket = create_jira_ticket(serv, args.jira_creds, ticket_desc)
```
solarkennedy: can you add an option to not make a ticket?

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -98,66 +116,109 @@ def get_reviewers(filename):
...
 
 
+def create_jira_ticket(serv, creds, description):
+    creds = creds.split(':')
+    options = {'server': 'https://jira.yelpcorp.com'}
+    jira_cli = JIRA(options=options, basic_auth=(creds[0], creds[1]))
+    jira_ticket = {}
+    # Sometimes the project name doesn't match a JIRA project
+    try:
+        jira_ticket = {
```
solarkennedy: Can you put this on line 156, and then only in the exception case just change the project name and try again?

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -98,66 +116,109 @@ def get_reviewers(filename):
...
+            'project': {'key': serv['project']},
+            'issuetype': {'name': 'Improvement'},
```
solarkennedy: not every project has "Improvement", do you need to specify this at all?

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -75,17 +59,51 @@ def in_tempdir():
...
     remote = 'git@sysgit.yelpcorp.com:yelpsoa-configs'
     subprocess.check_call(('git', 'clone', remote, '.'))
-    subprocess.check_call(('git', 'checkout', 'origin/master', '-b', branch_name))
+    subprocess.check_call(('git', 'checkout', '-b', branch_name))
 
 
-def commit(filename, memcpu):
-    message = 'Updating {} for under/overprovisioned {}'.format(filename, memcpu)
+def commit(filename, over):
+    message = 'Updating {} for {}provisioned cpu'.format(filename, over)
```
solarkennedy: Can you make this more descriptive? Heck have the whole message that was put into the reviewboard description and stick it in here! you can never not have enough context.

paasta_tools/contrib/paasta_update_soa_memcpu.py
fboxwala: The `cluster` string provided by the splunk search is of the form `marathon-{cluster}`

paasta_tools/contrib/paasta_update_soa_memcpu.py
fboxwala: The entire reviewboard description would be many lines for a commit message. I don't want to make the git history too hard to read. I'll add a little more detail though.

paasta_tools/contrib/paasta_update_soa_memcpu.py
fboxwala: I think I would rather clone and delete the repo. With jenkins I would have to add logic to clone it anyway, somewhere, wouldn't I?

paasta_tools/contrib/paasta_update_soa_memcpu.py
fboxwala: The JIRA client fails when this field is not defines

paasta_tools/contrib/paasta_update_soa_memcpu.py
fboxwala: (no)

matfra:
OK to merge but let's not run the tool yet

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -1,52 +1,44 @@
...
-        default='',
-        dest="cpu",
-        help='New value for cpus of service',
-    )
-    parser.add_argument(
-        '-t', '--ticket',
-        dest="ticket",
-        help='JIRA ticket tracking changes',
+        '-f', '--file-splunk',
+        help='Splunk csv from which to pull data. Defaults to paasta_overprovision_alerts_fired.csv',
```
thebostik: I don't see where we defined this default. The jenkins job at least it will be always be provided, I'm not sure a default is needed (except to help humans).

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -98,66 +128,100 @@ def get_reviewers(filename):
...
+        cpus = float(serv['cpus'])
+        provisioned_state = 'over'
+        if cpus > float(serv['old_cpus']):
+            provisioned_state = 'under'
+
+        serv['state'] = provisioned_state
+        ticket_desc = (
+            "We suspect that {s} {c} {i} may be {o}-provisioned"
+            " as of {d}. It initially had {x} cpus, but we think"
+            " it only needs {y} cpus."
```
thebostik: Maybe the word "only" could be removed in case this is an underprovisioned service.

PR #1664, opened 2017-12-14
Avoid using default event loop on secondary threads.

EvanKrall:
Deployd broke after #1653 was merged, because of errors like `There is no current event loop in thread 'Worker6'.`

This should fix that, by creating event loops where necessary. It also cleans up some boilerplate using the https://github.com/notion/a_sync library.

### DIFF
--- a/paasta_tools/frameworks/constraints.py
+++ b/paasta_tools/frameworks/constraints.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# Without this, the import of mesos.interface breaks because paasta_tools.mesos exists
 import re
 from typing import Any
 from typing import Callable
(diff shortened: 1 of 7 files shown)


EvanKrall:
Yeah, idk, I had some reservations about that too. I think with pinning it'll be okay -- the code is pretty straightforward: https://github.com/notion/a_sync/blob/master/a_sync/a_sync.py

I was considering writing my own `block`, but figured it would be better to build on the work of others.

mattmb:
I'm late to the party but I wonder if we could add an itest. I thought deployd did some "bouncing" in its itests but maybe it doesn't exercise exactly this code path. Can you take a look, I mean feel free to say no if it looks like a pain?

EvanKrall:
Yeah, I haven't dug much into why the deployd integration tests didn't break. The deployd itest does start up deployd and make sure something bounces.

PR #1667, opened 2017-12-15
adds a try/except to update_soa_memcpu

fboxwala:


thebostik:
Thanks again for an awesome internship, really nice work on this right sizing script!

PR #1668, opened 2017-12-18
Add jira dependency that was missed in #1663

ddelnano:
I was trying to use the paasta_update_soa_memcpu script for some Nowait services and I received the following error when running the script

```
(.paasta)ddelnano@dev7-uswest1cdevc:~/pg/paasta (master) $ paasta_tools/contrib/paasta_update_soa_memcpu.py
  File "paasta_tools/contrib/paasta_update_soa_memcpu.py", line 12, in <module>
    from jira.client import JIRA
ModuleNotFoundError: No module named 'jira'
```

Looks like the dependency was not added to the requirements.txt when the import was added in #1663.  Since there are no tests this was missed.  I opted to use jira 0.50 because 1.0.10 requires updating the requests version to >= 2.10.0

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,6 +18,7 @@ httplib2==0.9.2
 humanize==0.5.1
 inotify==0.2.8
 isodate==0.5.1
+jira==0.50
 jsonschema[format]==2.5.1
 kazoo==2.2
 marathon==0.9.3


ddelnano:
Since I still am learning the pythonic way of things I would defer to you / other's on what we should do.  From my experience with other package manger's I just assumed we would have updated one of the requirementxts*.txt files in the project root.

How would that requirements file get noticed by pip in the contrib subdirectory?  Or is it more a means of documenting the dependencies?

nhandler:
The `contrib` scripts aren't really part of paasta in the same sense as stuff under `paasta_tools`. This folder typically contains small helper scripts that individuals have prepared and found useful. It is a simple way to share them. However, we don't ever import/use these scripts from within paasta proper, they do not have tests, and they generally go through a much lighter review process.

The new requirements.txt files would need to be manually sourced inside of a virtual environment. They would be completely ignored by tox and the parent requirements.txt files.

ddelnano:
Ok well in that case maybe its not worth adding a new requirements file give the expectation of what is contained in the `contrib` directory.  I'm going to close for now.

PR #1669, opened 2017-12-19
Sanity-check mesos_cpu data.

EvanKrall:
@matfra

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -387,7 +387,7 @@ def mesos_cpu_metrics_provider(
 
     mesos_cpu_data = {}
     for task_id, stats in mesos_tasks_first_run.items():
-        stats2 = mesos_tasks_second_run[task_id]
+        stats2 = mesos_tasks_second_run.get(task_id)
         if stats is not None and stats2 is not None:
             try:
                 utime = min(float(stats['cpus_user_time_secs']), float(stats2['cpus_user_time_secs']))
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -374,45 +374,27 @@ def mesos_cpu_metrics_provider(
     monkey.patch_socket()
     jobs = [gevent.spawn(task.stats_callable) for task in mesos_tasks]
     gevent.joinall(jobs, timeout=60)
-    mesos_task_stats_1 = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs]))
+    mesos_tasks = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs]))
 
     current_time = int(datetime.now().strftime('%s'))
     time_delta = current_time - last_time
 
-    # sometimes, the mesos slave returns crazy numbers for cpus_system_time_secs and cpus_user_time_secs.
-    # we fetch twice, and use the lower of the two.
+    # Mesos slave statistics endpoint returns a crazy CPU value when a container is
+    # in the process of being killed. So we fetch twice, and use the lower of the two.
     jobs = [gevent.spawn(task.stats_callable) for task in mesos_tasks]
     gevent.joinall(jobs, timeout=60)
-    mesos_task_stats_2 = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs]))
+    mesos_tasks_second_run = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs]))
 
     mesos_cpu_data = {}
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -374,21 +374,46 @@ def mesos_cpu_metrics_provider(
...
     mesos_cpu_data = {}
-    for task_id, stats in mesos_tasks.items():
-        if stats is not None:
+    for task_id, stats in mesos_task_stats_1.items():
+        stats2 = mesos_task_stats_2[task_id]
+        if stats is not None and stats2 is not None:
+            class MesosGaveUsGarbageCpuError(Exception):
+                pass
+
+            def sanity_check(key):
```
matfra: Is it okay to define a function inside a for loop ?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -374,21 +374,46 @@ def mesos_cpu_metrics_provider(
...
-    mesos_tasks = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs]))
+    mesos_task_stats_1 = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs]))
 
     current_time = int(datetime.now().strftime('%s'))
     time_delta = current_time - last_time
 
+    # sometimes, the mesos slave returns crazy numbers for cpus_system_time_secs and cpus_user_time_secs.
+    # we fetch twice, and use the lower of the two.
+    jobs = [gevent.spawn(task.stats_callable) for task in mesos_tasks]
+    gevent.joinall(jobs, timeout=60)
```
matfra: This would potentially make the service autoscaler last 1 more minute.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -374,21 +374,46 @@ def mesos_cpu_metrics_provider(
...
+                    raise MesosGaveUsGarbageCpuError
+                return min(val1, val2)
+
             try:
-                utime = float(stats['cpus_user_time_secs'])
-                stime = float(stats['cpus_system_time_secs'])
-                limit = float(stats['cpus_limit']) - .1
+                utime = sanity_check('cpus_user_time_secs')
+                stime = sanity_check('cpus_system_time_secs')
+                limit = sanity_check('cpus_limit') - .1
```
matfra: no need to sanity check the limit

EvanKrall:
https://support.mesosphere.com/hc/en-us/requests/9230

matfra:
Reworked @EvanKrall 's patch and added checks

EvanKrall:
I'm happy with this. Let's get review from someone else before we ship.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -374,17 +374,24 @@ def mesos_cpu_metrics_provider(
...
     jobs = [gevent.spawn(task.stats_callable) for task in mesos_tasks]
     gevent.joinall(jobs, timeout=60)
-    mesos_tasks = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs]))
+    mesos_tasks_first_run = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs]))
 
     current_time = int(datetime.now().strftime('%s'))
     time_delta = current_time - last_time
 
+    # Mesos slave statistics endpoint returns a crazy CPU value when a container is
+    # in the process of being killed. So we fetch twice, and use the lower of the two.
```
nhandler: Is it possible for us to make the second request too quickly and get the crazy CPU value again? Do we need any sort of configurable delay in the middle?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -374,17 +374,24 @@ def mesos_cpu_metrics_provider(
...
+    jobs = [gevent.spawn(task.stats_callable) for task in mesos_tasks]
+    gevent.joinall(jobs, timeout=60)
+    mesos_tasks_second_run = dict(zip([task['id'] for task in mesos_tasks], [job.value for job in jobs]))
+
     mesos_cpu_data = {}
-    for task_id, stats in mesos_tasks.items():
-        if stats is not None:
+    for task_id, stats in mesos_tasks_first_run.items():
+        stats2 = mesos_tasks_second_run[task_id]
```
nhandler: Isn't there a potential issue where the tasks change between the first and second run causing this dictionary lookup to fail? Maybe use `mesos_tasks_second_run.get('task_id')` instead.

EvanKrall: good catch

PR #1671, opened 2017-12-20
Added a bespoke-autoscale cli helper

solarkennedy:
I feel a bit bad about not making tests.

But what tests should I *actually* write? I would have to mock out all the things? Maybe it could catch some autospec'y bugs?

### DIFF
--- a/paasta_tools/api/views/autoscaler.py
+++ b/paasta_tools/api/views/autoscaler.py
@@ -51,6 +51,11 @@ def update_autoscaler_count(request):
     service = request.swagger_data.get('service')
     instance = request.swagger_data.get('instance')
     desired_instances = request.swagger_data.get('json_body')['desired_instances']
+    if not isinstance(desired_instances, int):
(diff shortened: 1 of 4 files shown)


paasta_tools/api/views/autoscaler.py
```diff
@@ -51,6 +51,11 @@ def update_autoscaler_count(request):
     service = request.swagger_data.get('service')
     instance = request.swagger_data.get('instance')
     desired_instances = request.swagger_data.get('json_body')['desired_instances']
+    if not isinstance(desired_instances, int):
+        error_message = 'The provided body does not have an integer value for "desired_instances": {}'.format(
```
EvanKrall: Shouldn't this be covered by the schema on parameters?

solarkennedy: Hmm. It currently isn't and somehow is allowed to be None when I was testing this.
It allowed me to POST nothing and it became None in zk.

Let me see if I can enforce this at the schema level for real.

PR #1672, opened 2017-12-20
Allow the most recent tag to take precedence over a deploy tag.

solarkennedy:
Honestly I'm surprised at this. I'm kinda skeptical this works. We don't even need to run `get_latest_deployment_tag`? Is it really OK to look at all the tags and just sort by date?

This is a seriously iffy change. Obv I would run generate_deployments manually and see a diff.

Amazingly, a diff shows nothing on yelp-main. I should probably do a manual stop + deploy and test this for real?


EvanKrall:
This would require very careful deployment: it will likely cause a big bounce (as force_bounce will go from `None` to the date of last start for any services that have ever been been restarted) and will cause some services to stop (those that haven't been started since the last time they were stopped, and are running because a deploy happened since the last stop.)

solarkennedy:
Let me do a for-each and get an analysis of how many things would *actually* bounce.

PR #1674, opened 2017-12-20
Added a script to emit additional paasta metrics

solarkennedy:
Although ideally this number would be available "for free", mts's are expensive, and we don't need to emit this "constant" for every docker container at yelp.

To that end, I made this script that emits the metric directly. It doesn't change that much, so I'm think of running every 5m or so on masters.

### DIFF
--- a/paasta_tools/contrib/emit_allocated_cpu_metrics.py
+++ b/paasta_tools/contrib/emit_allocated_cpu_metrics.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+import logging
+
+import yelp_meteorite
+
+from paasta_tools.cli.utils import get_instance_config
+from paasta_tools.utils import get_services_for_cluster
+from paasta_tools.utils import load_system_paasta_config
+
+log = logging.getLogger(__name__)
+
+
+def emit_metrics_for_type(instance_type):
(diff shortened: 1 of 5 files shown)


paasta_tools/contrib/emit_allocated_cpu_metrics.py
```diff
@@ -0,0 +1,47 @@
...
+        dimensions = {
+            'service_name': service_instance_config.service,
+            'paasta_cluster': service_instance_config.cluster,
+            'instance_name': service_instance_config.instance,
+        }
+
+        log.info("Emitting paasta.service.cpus = {} with dimensions {}".format(
+            cpus, dimensions,
+        ))
+        gauge = yelp_meteorite.create_gauge('paasta.service.cpus', dimensions)
```
EvanKrall: While we're here, can we add .mem, .disk, and maybe .instance_count? (I think we're already doing instance count from autoscaled services, but it would feel more complete to have them all.)

Having mem and disk and such would allow us to make better dashboards along the lines of the will-yelp-main-fit dashboard. Instance count would allow us to calculate total cost of services.

EvanKrall: ...maybe also min_instances and max_instances?

PR #1677, opened 2017-12-21
Better cluster boost

matfra:
* Apply the cluster boost only on CPU resource
* Fix the call from a paasta boost
* Figure out the region(s) automatically
* Replaced get by status to remove confusion:
  - get is for the autoscaler
  - status is for a human to check

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -1112,7 +1112,7 @@ def autoscale_local_cluster(
     utilization_errors = get_all_utilization_errors(
         autoscaling_resources=autoscaling_resources,
         all_pool_settings=all_pool_settings,
-        mesos_state=MesosState,
+        mesos_state=mesos_state,
         system_config=system_config,
     )
     autoscaling_scalers: Dict[Tuple[str, str], List[ClusterAutoscaler]] = defaultdict(list)
@@ -1224,7 +1224,7 @@ def get_autoscaling_info_for_all_resources(mesos_state: MesosState) -> List[Auto
     utilization_errors = get_all_utilization_errors(
         autoscaling_resources=autoscaling_resources,
         all_pool_settings=all_pool_settings,
-        mesos_state=MesosState,
+        mesos_state=mesos_state,
         system_config=system_config,
     )
     vals = [
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -1267,6 +1267,7 @@ def get_mesos_utilization_error(
     When the boost feature is enabled, the current_load will be artifically increased
     and stored into boosted_load. If the boost is disabled, boosted_load = current_load
     """
+    system_config = load_system_paasta_config()
     try:
         region_pool_utilization_dict = get_resource_utilization_by_grouping(
             lambda slave: (slave['attributes']['pool'], slave['attributes']['datacenter'],),
@@ -1291,7 +1292,7 @@ def get_mesos_utilization_error(
         current_load = total - free
 
         # We apply the boost only on the cpu resource.
-        if resource == 'cpus':
+        if resource == 'cpus'and system_config.get_cluster_boost_enabled():
             boosted_load = cluster_boost.get_boosted_load(region=region, pool=pool, current_load=current_load)
         else:
             boosted_load = current_load
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -46,6 +46,7 @@ from paasta_tools.mesos_tools import get_mesos_task_count_by_slave
 from paasta_tools.mesos_tools import slave_pid_to_ip
 from paasta_tools.mesos_tools import SlaveTaskCount
 from paasta_tools.metrics.metastatus_lib import get_resource_utilization_by_grouping
+from paasta_tools.metrics.metastatus_lib import ResourceInfo
 from paasta_tools.metrics.metrics_lib import get_metrics_interface
 from paasta_tools.paasta_maintenance import is_safe_to_kill
 from paasta_tools.utils import load_system_paasta_config
@@ -1279,19 +1280,28 @@ def get_mesos_utilization_error(
         return 0
 
     log.debug(repr(region_pool_utilization_dict))
-    free_pool_resources = region_pool_utilization_dict['free']
-    total_pool_resources = region_pool_utilization_dict['total']
     usage_percs = []
-    for free, total in zip(free_pool_resources, total_pool_resources):
+    for resource in ResourceInfo._fields:
+        free = getattr(region_pool_utilization_dict['free'], resource)
+        total = getattr(region_pool_utilization_dict['total'], resource)
+
         if math.isclose(total, 0):
             continue
+
         current_load = total - free
-        boosted_load = cluster_boost.get_boosted_load(region=region, pool=pool, current_load=current_load)
+
+        # We apply the boost only on the cpu resource.
+        if resource == 'cpus':
+            boosted_load = cluster_boost.get_boosted_load(region=region, pool=pool, current_load=current_load)
+        else:
+            boosted_load = current_load
 
         usage_percs.append(boosted_load / float(total))
 
     if len(usage_percs) == 0:  # If all resource totals are close to 0 for some reason
         return 0
 
+    # We only look at the percentage of utilization. Whichever is the highest (closer or past the setpoint)
+    # will be used to determine how much we need to scale
     utilization = max(usage_percs)
     return utilization - target_utilization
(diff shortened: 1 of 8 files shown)


matfra:
I added a paasta config system toggle as suggested by @nhandler

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -1279,19 +1281,28 @@ def get_mesos_utilization_error(
         return 0
 
     log.debug(repr(region_pool_utilization_dict))
-    free_pool_resources = region_pool_utilization_dict['free']
-    total_pool_resources = region_pool_utilization_dict['total']
     usage_percs = []
-    for free, total in zip(free_pool_resources, total_pool_resources):
```
matfra: This is the only change that needs close attention @nhandler

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -1279,19 +1281,28 @@ def get_mesos_utilization_error(
...
+        free = getattr(region_pool_utilization_dict['free'], resource)
+        total = getattr(region_pool_utilization_dict['total'], resource)
+
         if math.isclose(total, 0):
             continue
+
         current_load = total - free
-        boosted_load = cluster_boost.get_boosted_load(region=region, pool=pool, current_load=current_load)
+
+        # We apply the boost only on the cpu resource.
```
nhandler: I imagine this will end up causing confusion. Our other autoscalers operate on all of the resources. I can easily see us running into a situation where we are out of memory/disk for some reason and someone tries running the cluster boost tool as a quick fix.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -1279,19 +1281,28 @@ def get_mesos_utilization_error(
...
+        if resource == 'cpus'and system_config.get_cluster_boost_enabled():
+            boosted_load = cluster_boost.get_boosted_load(region=region, pool=pool, current_load=current_load)
+        else:
+            boosted_load = current_load
 
         usage_percs.append(boosted_load / float(total))
 
     if len(usage_percs) == 0:  # If all resource totals are close to 0 for some reason
```
nhandler: Will this statement ever be true? Looks like it will only be 0 if we never run the for loop.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -1266,6 +1267,7 @@ def get_mesos_utilization_error(
     When the boost feature is enabled, the current_load will be artifically increased
     and stored into boosted_load. If the boost is disabled, boosted_load = current_load
     """
+    system_config = load_system_paasta_config()
```
oktopuz: Can you pass system_paastsa_config from autoscale_local_cluster() instead of reading and parsing the same files?

paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -89,6 +89,19 @@ def get_boosted_load(region: str, pool: str, current_load: float) -> float:
         return current_load
 
 
+def print_boost_value(region: str, pool: str) -> float:
+    """THis function prints the boost value if a boost is active
```
nhandler: NP: Fix capitalization

nhandler: It also isn't `print`ing the value, it is returning it.

paasta_tools/cli/cmds/boost.py
```diff
@@ -46,8 +46,9 @@ def add_subparser(subparsers):
     boost_parser.add_argument(
         '-c', '--cluster',
         type=str,
+        action='append',
         required=True,
-        help="Cluster to boost. ex: nova-prod. You can find a list of paasta clusters with `paasta list-clusters'",
+        help="Cluster to boost. ex: nova-prod .This option can be specified multiple times.",
```
nhandler: Drop the example cluster or put some dummy not yelpy one to match what we do elsewhere.

nhandler: Can you also make this argument behave similar to -c in other commands like paasta status? We take a comma separated list and have completion support.

paasta_tools/cli/cmds/boost.py
```diff
@@ -102,21 +97,23 @@ def paasta_boost(args):
...
-            override=args.override if args.action == 'set' else None,
-            boost=args.boost if args.action == 'set' else None,
-            verbose=args.verbose,
-        )
-    else:
-        paasta_print("Cluster %s doesn't look like a valid cluster" % args.cluster)
-        paasta_print("Here is a list of valid paasta clusters:\n{}".format(all_clusters))
-        return_code = 1
+    for cluster in args.cluster:
+        if cluster not in all_clusters:
```
nhandler: I thought we had a function to handle validating a cluster, but I'm struggling to find it right now. Can you either find it or make one (somewhere more generic than in this file)?

paasta_tools/cli/utils.py
```diff
@@ -636,34 +634,44 @@ def run_paasta_cluster_boost(
...
-        boost=boost,
-        verbose=verbose,
-    )
+    aggregated_code = 0
+    aggregated_stdout = ""
+    for cluster in output:
+        code = output[cluster][0]
+        stdout = output[cluster][1]
+        if not code == 0:
+            aggregated_code = code
```
nhandler: You append to aggregated_stdout, but you will only get the aggregated_code of the last failure. This might cause confusion when trying to interpret the stdout. Also, not sure if you also want to save stderr to be safe.

paasta_tools/paasta_cluster_boost.py
```diff
@@ -109,40 +101,46 @@ def paasta_cluster_boost():
...
-    elif action == 'clear':
-        if not cluster_boost.clear_boost(pool=pool, region=region):
-            paasta_print('Failed! Check the logs.')
-            return False
+    regions = get_regions(pool)
 
-    else:
-        raise NotImplementedError("Action: '%s' is not implemented." % action)
+    if len(regions) == 0:
+        paasta_print('Error, no slaves found in pool {}'.format(pool))
```
nhandler: Can you be consistent and do `ERROR:` like above?

paasta_tools/utils.py
```diff
@@ -1590,6 +1591,13 @@ def get_cluster_autoscaling_draining_enabled(self) -> bool:
         :returns A bool"""
         return self.config_dict.get('cluster_autoscaling_draining_enabled', True)
 
+    def get_cluster_boost_enabled(self) -> bool:
+        """ Enable the cluster boost. Note boost that applies only to the resource cpus anyway.
```
nhandler: "Note that the boost only applies to the CPUs"

paasta_tools/utils.py
```diff
@@ -1590,6 +1591,13 @@ def get_cluster_autoscaling_draining_enabled(self) -> bool:
...
+        If the boost is toggle on here but not configured, it will be transparent.
```
nhandler: "toggled"

paasta_tools/utils.py
```diff
@@ -1590,6 +1591,13 @@ def get_cluster_autoscaling_draining_enabled(self) -> bool:
...
+
+        :returns A bool"""
```
nhandler: Can you specify what this bool means (what is true/false)?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
matfra: Good suggestion.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
matfra: Keep in mind our main and only use case for now. I'm happy to re-visit the feature when a second use case shows up

paasta_tools/autoscaling/autoscaling_cluster_lib.py
matfra: It's been explicitely added in the commit 5c3a683b so I'm sure we can find more context about it. I agree to look into it but not part of this PR as I'm trying to make as few changes as possible to this lib

paasta_tools/autoscaling/cluster_boost.py
matfra: Good remark, I've renamed that function

paasta_tools/cli/cmds/boost.py
matfra: done

paasta_tools/cli/utils.py
matfra: If two cluster have different non-zero return code, what code should be returned?

paasta_tools/cli/cmds/boost.py
matfra: Are you thinking about this one ? https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/metastatus.py#L120

paasta_tools/cli/utils.py
oktopuz: @matfra usually we return max()
https://github.com/Yelp/paasta/blob/master/paasta_tools/paasta_serviceinit.py#L248

paasta_tools/utils.py
matfra: done

PR #1679, opened 2017-12-21
Perform paasta status ssh in parallel

henryzhangsta:
This makes paasta status take max(latencies) instead of sum(latencies), with respect to paasta_serviceinit time to complete.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import difflib
+import concurrent.futures
 import os
 import sys
 from collections import defaultdict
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -431,10 +431,16 @@ def paasta_status(args):
                     system_paasta_config=load_system_paasta_config(),
```
oktopuz: Could you use this opportunity and move load_system_paasta_config() higher, so that we don't reload it for every cluster?

oktopuz:
lgtm

tests/cli/test_cmds_status.py
```diff
@@ -119,19 +123,19 @@ def test_report_status_for_cluster_displays_deployed_service(
...
-    status.report_status_for_cluster(
+    _, output = status.report_status_for_cluster(
         service=service,
         cluster='fake_cluster',
         deploy_pipeline=planned_deployments,
         actual_deployments=actual_deployments,
         instance_whitelist=instance_whitelist,
         system_paasta_config=system_paasta_config,
     )
-    output, _ = capfd.readouterr()
```
EvanKrall: Does this test (and subsequent ones) still need the capfd fixture after you remove this?

EvanKrall:
Test failure seems to just be an unrelated itest flake. example_cluster is having trouble building with that stupid `E: Couldn't find any package by regex 'python3.6-dev'` error.

PR #1680, opened 2017-12-21
Enabled paasta_remote_run to connect to diffferent mesos clusters

huadongliu:
This is for TASKPROC-166 to enable paasta_remote_run schedule tasks on more than one mesos clusters. Note that "role" is added to adhoc schema so that resources with non-taskproc role can be used. This also helps TASKPROC-169 where "paasta remote_run" was stuck because resources with "*" role do not match the default taskproc role.

The change is tested on devc:
tronplayground-uswest1adevc: $ sudo .tox/py36/bin/paasta_remote_run.py start --service taskproc_canary --instance remote --cluster norcal-stageg --verbose
tronplayground-uswest1adevc: $ sudo .tox/py36/bin/paasta_remote_run.py start --service taskproc_canary --instance remote --cluster norcal-devc --verbose

### DIFF
--- a/paasta_tools/paasta_remote_run.py
+++ b/paasta_tools/paasta_remote_run.py
@@ -232,6 +232,7 @@ def build_executor_stack(
     service,
     instance,
     cluster,
+    role,
     pool,
     # TODO: move run_id into task identifier?
     run_id,
@@ -239,8 +240,10 @@ def build_executor_stack(
     framework_staging_timeout,
 ):
 
+    cluster_fqdn = system_paasta_config.get_cluster_fqdn_format().format(cluster=cluster)
     mesos_address = '{}:{}'.format(
-        mesos_tools.get_mesos_leader(), mesos_tools.MESOS_MASTER_PORT,
+        mesos_tools.find_mesos_leader(cluster_fqdn),
+        mesos_tools.MESOS_MASTER_PORT,
     )
 
     # TODO: implement DryRunExecutor?
@@ -248,7 +251,7 @@ def build_executor_stack(
 
     MesosExecutor = processor.executor_cls('mesos')
     mesos_executor = MesosExecutor(
-        role=taskproc_config.get('role', taskproc_config.get('principal')),
+        role=role,
         pool=pool,
         principal=taskproc_config.get('principal'),
         secret=taskproc_config.get('secret'),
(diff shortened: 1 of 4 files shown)


paasta_tools/utils.py
```diff
@@ -624,6 +625,11 @@ def get_extra_volumes(self) -> List[DockerVolume]:
         <https://mesosphere.github.io/marathon/docs/native-docker.html>`_"""
         return self.config_dict.get('extra_volumes', [])
 
+    def get_role(self, default: str) -> str:
```
solarkennedy: Why does this function need the default provided to it?
Isn't the purpose of this function exactly to return the default role if unset?

paasta_tools/paasta_remote_run.py
```diff
@@ -415,6 +418,7 @@ def handle_interrupt(_signum, _frame):
             processor=processor,
             service=service,
             instance=instance,
+            role=native_job_config.get_role('taskproc'),
```
solarkennedy: This should be 

job_config.get_role() or 'taskproc'

I think

paasta_tools/utils.py
huadongliu: I was trying to provide a default for get_role and 'taskproc' is not always a good one. After reading your comments, I think None is better.

paasta_tools/paasta_remote_run.py
```diff
@@ -415,6 +418,7 @@ def handle_interrupt(_signum, _frame):
             processor=processor,
             service=service,
             instance=instance,
+            role=native_job_config.get_role() or 'taskproc',
```
Rob-Johnson: can you make 'taskproc' a system_paasta_config entry for 'default_remote_run_role'?

paasta_tools/paasta_remote_run.py
```diff
@@ -232,23 +232,26 @@ def build_executor_stack(
...
     # TODO: move run_id into task identifier?
     run_id,
     system_paasta_config,
     framework_staging_timeout,
 ):
 
+    cluster_fqdn = system_paasta_config.get_cluster_fqdn_format().format(cluster=cluster)
     mesos_address = '{}:{}'.format(
-        mesos_tools.get_mesos_leader(), mesos_tools.MESOS_MASTER_PORT,
+        mesos_tools.find_mesos_leader("mesos." + cluster_fqdn),
```
Rob-Johnson: you don't need the mesos prefix - you'll get there just with paasta-{cluster}.yelp:MESOS_MASTER_PORT

solarkennedy:
@huadongliu can you update this?

huadongliu:
Updated. @Rob-Johnson Please take a look.

PR #1681, opened 2017-12-22
Ignore all .cache directories

henryzhangsta:


EvanKrall:
I think it's fine to merge as is, but if the intention was to ignore all .caches, then I think we want `**/.cache`

nhandler:
I actually think `.cache/` is what we want. It will only match directories named `.cache` and their contents. It will search for the directory recursively. There are some nice examples on https://www.atlassian.com/git/tutorials/gitignore#git-ignore-patterns (search for `build/logs/latest/debug.log`).

PR #1688, opened 2018-01-02
Fixed the output when asking for cluster boost status

matfra:


paasta_tools/paasta_cluster_boost.py
```diff
@@ -126,7 +126,7 @@ def paasta_cluster_boost():
                 return False
 
         elif action == 'status':
-            continue
+            pass
```
nhandler: Since all of the other if/elif blocks `return`, why not just move the `paasta_print` call into this `elif action == 'status'` to make it more obvious what is going on?

matfra: This is exactly where it was originally but turns out that a very human thing to do is to check the status right away after you set, or clear a boost. Therefore, you end up checking the status on all these actions.

nhandler: Oh, my bad. I completely glossed over the nested if statements within each elif and thought that every block would always return, causing the status to only show with an action of 'status'.

PR #1689, opened 2018-01-03
Use marathon 1.4.10 in itests

nhandler:
This is PAASTA-13542


solarkennedy:
Can you make a PR on marathon-python too?

nhandler:
I filed https://github.com/thefactory/marathon-python/pull/236

PR #1690, opened 2018-01-05
Revert the Mesos CPU sanity check and added Filter mesos cpu value

matfra:
* Revert the Mesos CPU sanity check

* Added filter for bogus mesos CPU values

The filter make sure the value reported by Mesos doesn't exceed the theorical maximum allowed CPU usage for this service. The feature is toggled by a system config. It's disabled by default so it
should be safe to roll out.
A warning message will be written in the logs everytime such a value is
encountered. Additionally with debug verbosity: the entire CPU stats will be printed.


matfra:
I highly recommend looking at the diffs one commit at a time. cc @EvanKrall

PR #1696, opened 2018-01-10
Enable yaml-cache in deployd

mattmb:
This re-enables the yaml cache in deployd whilst strategically clearing
the cache when deployd needs to be sure it is reading the latest data
from disk.

I'm pretty sure I caught all the places where I need to clear this
cache. Specifically, I've done anywhere in a watcher where the watcher
is using those yaml files to make a decision. I've also cleared it in
the worker before the worker runs SMJ for that service.

### DIFF
--- a/paasta_tools/deployd/common.py
+++ b/paasta_tools/deployd/common.py
@@ -10,6 +10,8 @@ from typing import List
 from typing import Optional
 from typing import Tuple
 
+import service_configuration_lib
+
 from paasta_tools.marathon_tools import DEFAULT_SOA_DIR
 from paasta_tools.marathon_tools import get_all_marathon_apps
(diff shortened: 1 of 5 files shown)


solarkennedy:
````
paasta_tools/deployd/common.py:205: error: Function is missing a type annotation
````

mattmb:
I forgot we were living in the future!

@solarkennedy what do you think in general? Is this a good idea or will it just cause some tricky to find bug somewhere down the road?

solarkennedy:
I say don't ship unless you can get some profiling data to indicate that this is a large percentage of time.

mattmb:
sgtm, I'll close this for now and we can always resurrect the idea if we find out its hurting us down the road.

PR #1697, opened 2018-01-10
Close connections to Zookeeper properly

keymone:
We seem to be leaking file descriptors of ZK clients. Calling `zk.stop()` is not enough so add a bunch of `zk.close()` calls where appropriate.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -756,3 +756,4 @@ def create_autoscaling_lock():
         lock.release()
     finally:
         zk.stop()
+        zk.close()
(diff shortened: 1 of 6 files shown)


mattmb:
https://github.com/python-zk/kazoo/issues/137 is what made us think about this.

The docs also state:
"This method should be called on a stopped client before it is discarded. Not doing so may result in filehandles being leaked."

https://access.redhat.com/solutions/488623 makes me think that anytime we peak over 1024 file descriptors we get exceptions from our kazoo code (including in bounce locks).

mattmb:
Also check this out:
https://github.com/python-zk/kazoo/pull/448

I think we just need to upgrade kazoo since that is in 2.4 and we're on 2.2!

PR #1699, opened 2018-01-11
Fix randomly closing ZK connections in paasta-api

mattmb:
We've been getting exceptions from paasta-api about closed connections.
The problem occurs when there are multiple requests made to the
instances endpoint at the same time. For autoscaled services we use
ZookeeperPool to pool our connections to zookeeper so that we can get
the correct number of instances. However when the API handles multiple
requests it will call `.stop()` and `.close()` on the zookeeper client
when it is possilbly still being used by another request. This leads to
exceptions like `kazoo.exceptions.ConnectionClosedError`.

I've "fixed" this by adding the `ZookeeperPool` to the api main func so
that we don't tear down the client until the whole api server quits.

### DIFF
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -30,6 +30,7 @@ import paasta_tools.api
 from paasta_tools import marathon_tools
 from paasta_tools.api import settings
 from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.utils import ZookeeperPool
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


keymone:
probably worth leaving a TODO comment in the code to investigate this further?

mattmb:
Yeah, the fix is a bit hacky isn't it. But it does work, I tried it out with the example-cluster.

Good idea to add a comment though, as otherwise it's not exactly obvious why I'm setting up a client and not passing it anywhere.

PR #1700, opened 2018-01-11
Check marathon replication on non-default reg

mattmb:
This is a subtle change to ensure that we atleast check Marathon
replication when we cannot check smartstack replication due to multiple
namespaces. This should mean that services which do not advertise into
the same namespace as their service_instance name will atleast have
their replication checked against Marathon.


paasta_tools/check_marathon_services_replication.py
```diff
@@ -314,18 +306,26 @@ def check_service_replication(
         cluster=instance_config.cluster,
         soa_dir=instance_config.soa_dir,
     )
+
     if proxy_port is not None:
```
solarkennedy: I don't like how this if block is structured. How about:

```
if proxy_port is None and primary_registration == jobid :
   check_smartstack
else
   check_marathon
```

PR #1701, opened 2018-01-12
Tweaks to paasta rightsizer script.

thebostik:
Few minor fixes and clarify ticket description after handling the same questions and seeing a few bugs after using this the last 2 weeks.

Update dashboard shortlink to prefill filters.

Don't drop the first result, preserve ordering of latest 3 authors, update ticket desc, dedent to fix no-tick option.

### DIFF
--- a/paasta_tools/contrib/paasta_update_soa_memcpu.py
+++ b/paasta_tools/contrib/paasta_update_soa_memcpu.py
@@ -107,7 +107,7 @@ def get_perf_data(creds, filename):
         serv['project'] = d['result']['project']
         services_to_update.append(serv)
 
-    return services_to_update[1:]
+    return services_to_update
 
 
 def clone(branch_name):
@@ -128,13 +128,17 @@ def commit(filename, serv):
 
 
 def get_reviewers(filename):
+    recent_authors = set()
     authors = subprocess.check_output((
         'git', 'log', '--format=%ae', '--', filename,
     )).decode('UTF-8').splitlines()
 
-    authors = list(set(authors))
     authors = [x.split('@')[0] for x in authors]
-    return authors[:3]
+    for author in authors:
+        recent_authors.add(author)
+        if len(recent_authors) >= 3:
+            break
+    return recent_authors
 
 
 def review(filename, description, provisioned_state, manual_rb):
@@ -201,6 +205,11 @@ def create_jira_ticket(serv, creds, description):
     return tick.key
 
 
+def _get_dashboard_qs_param(param, value):
+    # Some dashboards may ask for query string params like param=value, but not this provider.
+    return 'variables%5B%5D={}%3D{}:{}'.format(param, param, value)
+
+
 def main(argv=None):
     args = parse_args(argv)
     services_to_update = get_perf_data(args.splunk_creds, args.file_splunk)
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -214,11 +223,14 @@ def main(argv=None):
...
+            " during the 1 week prior to {d}. It initially had {x} cpus, but we think"
             " it needs {y} cpus."
-            "\n- Dashboard: https://y.yelpcorp.com/{o}provisioned\n- Service"
-            " owner: {n}\n- Estimated monthly excess cost: ${m}"
+            "\n- Dashboard: https://y.yelpcorp.com/{o}provisioned?{superregion_param}&{service_param}&{instance_param}"
+            "\n- Service owner: {n}"
+            "\n- Estimated monthly excess cost: ${m}"
+            "\n\nFor more information:"
             "\n- Runbook: https://y.yelpcorp.com/rb-provisioning-alert"
             "\n- Alert owner: team-perf@yelp.com"
```
solarkennedy: Can you add even more? Can you add who is supposed to give a shipit, who is responsible for merging changes in, what happens after the change is merged?

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -231,25 +243,29 @@ def main(argv=None):
             m=serv['money'],
             x=serv['old_cpus'],
             y=serv['cpus'],
+            superregion_param=_get_dashboard_qs_param('superregion', serv['cluster'].replace('marathon-', '')),
```
solarkennedy: Can you call this 'cluster' throughout?

thebostik: it has to be superregion at least where it will be passed to sfx as a query string variable, but i could change my format string name

paasta_tools/contrib/paasta_update_soa_memcpu.py
thebostik: Hmm, yea. I thought adding the auto-generated line would be helpful but this is definitely still a new process. I can see why more info is better. Do you think this needs 2 people to respond (1 to shipit, 1 to merge) or can we clarify here that we expect the person who reviews the number/dashboard can both shipit and merge the jenkins branch?

Was there something more you were suggesting re: the who part?

I can see the recommended steps in-line being helpful here. More minor clarifications off the beaten path we can always add to the y/rb-provisioning-alert doc anytime.

Would an announcement in EPS that we're now automatically ticketing and CRing over-provisioned services to reduce costs, and quickly explain that merge process, and explain why each team has to do their part even though they are individually just drops in a bucket help increase traction?

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -214,11 +223,14 @@ def main(argv=None):
 
         serv['state'] = provisioned_state
         ticket_desc = (
-            "We suspect that {s}.{i} in {c} may be {o}-provisioned"
-            " as of {d}. It initially had {x} cpus, but we think"
+            "This ticket and CR have been auto-generated, and only requires action from your team to ship."
```
solarkennedy: What action? Just say "All you need to do is give a shipit if you agree, then X"
Where X is... what actually happens. Will it get merged immediately ? Does a perf team member merge it in in a week? Just say what happens to give developers more faith in what they are shipping.

thebostik: I have been accused of talking diagonally before... thanks for the second round of feedback.

I'm hoping that everyone will merge this themselves after our and their review. I don't want to make anyone feel like we're executive decisioning their service configuration.

New diff is up with another version of this ticket description.

PR #1711, opened 2018-01-24
[WIP] Integrate vault with local-run

mattmb:
This still needs some tests but is basically working. I'm putting it up for early feedback before I polish it up.

* Create secret_provider framework
* Create a vault secret_provider which uses our internal vault_tools
package to decrypt secrets
* Update local-run to check for secrets in env vars for service and pass
environment to secret_provider to be decrypted

cc @filipposc5 you may be interested in this too!


paasta_tools/cli/cmds/local_run.py
```diff
@@ -519,6 +538,26 @@ def check_if_port_free(port):
...
+):
+    ecosystem = secret_provider_kwargs['vault_cluster_config'].get(cluster_name, 'devc')
+    secret_env_vars = {k: v for k, v in environment.items() if is_secret_ref(v)}
+    if secret_env_vars:
+        secret_provider = get_secret_provider(secret_provider_name, soa_dir, service_name, ecosystem)
+        secret_environment = secret_provider.decrypt_environment(
+            secret_env_vars,
+            **secret_provider_kwargs,
+        )
+    environment.update(secret_environment)
```
solarkennedy: Can you instead just return variables and make the caller update?

paasta_tools/secret_providers/vault.py
```diff
@@ -0,0 +1,45 @@
+import os
+
+from vault_tools.client.jsonsecret import get_plaintext
```
solarkennedy: Don't you have to gate this some how, as this i

paasta_tools/cli/cmds/local_run.py
```diff
@@ -519,6 +538,26 @@ def check_if_port_free(port):
...
+    environment,
+    soa_dir,
+    service_name,
+    cluster_name,
+    secret_provider_kwargs,
+):
+    ecosystem = secret_provider_kwargs['vault_cluster_config'].get(cluster_name, 'devc')
+    secret_env_vars = {k: v for k, v in environment.items() if is_secret_ref(v)}
+    if secret_env_vars:
+        secret_provider = get_secret_provider(secret_provider_name, soa_dir, service_name, ecosystem)
```
solarkennedy: kwargs. It is too easy to mess up the order of arguments here

paasta_tools/cli/cmds/local_run.py
```diff
@@ -519,6 +538,26 @@ def check_if_port_free(port):
...
 
+def decrypt_secret_environment_variables(
+    secret_provider_name,
+    environment,
+    soa_dir,
+    service_name,
+    cluster_name,
+    secret_provider_kwargs,
+):
+    ecosystem = secret_provider_kwargs['vault_cluster_config'].get(cluster_name, 'devc')
```
solarkennedy: Hmm. Yuck :(
Why does this need to be here, can't the lower-level secret provider "know" where it is?

paasta_tools/cli/cmds/local_run.py
mattmb: That sounds, like a good shout, let me look at this again

paasta_tools/secret_providers/vault.py
mattmb: I could add a `try` `except`. We actually import this by calling `__import__` directly and the name is configurable. Maybe I should set the default to the base provider then if anyone tries to use secrets they'll get a NotImplemented exception. And at yelp we can set the provider to vault explicitly?

paasta_tools/cli/cmds/local_run.py
mattmb: Not ideal is it :-(

The reason this exists is we need to map between paasta cluster and vault cluster. There are some paasta clusters that don't have a local vault cluster and we need to know which one to talk to in those cases (and which secret ciphertext to read from yelpsoa-configs.

Are you thinking that this map should be a config file that belongs to vault_tools instead of part of PaaSTAs config? I think maybe that would be better, it's really just here because it was an "easy" place to put it for now.

mattmb:
I've added tests and covered off the feedback. If travis passes this is ready for review

mattmb:
@solarkennedy can you give this another read through? I think I've fixed everything from the previous review.

PR #1713, opened 2018-01-24
(DO NOT REIVEW) Make spark-run's sudo behavior consistent with local-run's

huadongliu:
Unlike local-run's re-execute with "sudo -H" for pulling an image, spark-run does "sudo -H docker pull" in a child process. This patch makes spark-run do the same.


huadongliu:
Discard this PR since spark-run needs user credentials for aws access.

PR #1714, opened 2018-01-30
Opinions welcome: Convert service autoscaler to use asyncio

EvanKrall:
This changes the interface of several autoscaling_service_lib functions to be coroutines instead.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,5 +1,5 @@
 [run]
-concurrency = gevent
+concurrency = thread
 branch = True
 source = .
 omit =
--- a/paasta_itests/steps/autoscaling_service_steps.py
+++ b/paasta_itests/steps/autoscaling_service_steps.py
@@ -1,3 +1,4 @@
+import a_sync
 from behave import given
 from behave import then
 from behave import when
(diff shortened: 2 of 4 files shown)


EvanKrall:
I'm honestly a little conflicted about whether I want to merge this. I definitely think asyncio is the future of python, but I'm not convinced it's the present of python. Using `a_sync` to do zookeeper operations on another thread feels really hacky. The other option, though, is to use [aiozk](https://github.com/tipsi/aiozk), which seems [incomplete and based on an incomplete port of Kazoo to Tornado](https://github.com/tipsi/aiozk/issues/16#issuecomment-350686127).

The other places I'm using `a_sync.run`, `a_sync.Parallel`, or `a_sync.to_async` to convert sync to async:
- in `autoscale_marathon_instance()`, on `marathon_service_config.get_instances()`
  - this is fundamentally ZK code so it could be converted to aiozk probably
- in `mesos_cpu_metrics_provider`, around `task.stats_callable`, which is code originally from mesos.cli that we now own -- under the hood it uses requests; hypothetically we could convert it to use aiohttp and have an async interface.

Or maybe I just drop this, and run all the as-currently-written (sync + gevented task.stats_callable) `autoscale_marathon_instance`  calls under gevent?

EvanKrall:
re: aiozk's incompleteness -- I think only the recipes are incomplete; we use three recipes: Election, ChildrenWatch, and DataWatch. The latter two are tested, so presumably work. Election is untested, so probably doesn't work, but maybe we could just fix that if we wanted to move to aiozk.

mattmb:
I share your reservations around python+async in general. Although I think my prejudice is personal because I just find it tricky to form a mental model of async code.

Can you break down the motivation? I guess it's to "make it faster". What is really slow at the moment? I guess any http calls would be better async. That said, don't the mesos ones end up hitting `state.json`? I suspect we have some caching in the `mesos.cli` code but even so that might change things? I would have hoped the ZK calls were pretty quick but maybe I'm wrong?

EvanKrall:
While trying to benchmark this version of the autoscaler, I've found that it seems to deadlock; I'll need to deal with that before we can merge this.

solarkennedy:
@EvanKrall I vote close if you are not going to maintain this PR

PR #1715, opened 2018-01-30
Run jupyter notebook as non-root, added spark driver options

huadongliu:
Tested in norcal-devc.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -112,6 +112,24 @@ def add_subparser(subparsers):
         default=4,
     )
 
+    list_parser.add_argument(
+        '--driver-max-result-size',
+        type=int,
+        help='Limit of total size in GB of serialized results of all partitions',
+    )
+
+    list_parser.add_argument(
+        '--driver-memory',
+        type=int,
+        help='Size of Spark driver memory in GB',
+    )
+
+    list_parser.add_argument(
+        '--driver-cores',
+        type=int,
+        help='Number of CPU cores for the Spark driver',
+    )
+
     list_parser.set_defaults(command=paasta_spark_run)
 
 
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -165,6 +184,13 @@ def get_spark_configuration(
     spark_conf['SPARK_EXECUTOR_CORES'] = args.executor_cores
     spark_conf['SPARK_EXECUTOR_MEMORY'] = '%dg' % args.executor_memory
 
+    if args.driver_max_result_size:
+        spark_conf['SPARK_DRIVER_MAX_RESULT_SIZE'] = '%dg' % args.driver_max_result_size
```
sagar8192: Is the 'g' in here for GB? I yes, can we mention the unit in the help message?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -227,7 +255,7 @@ def configure_and_run_docker_container(
         return 1
     # Changes at docker ENTRYPOINT or CMD does not work.
     elif docker_cmd == 'jupyter':
-        docker_cmd = 'jupyter notebook -y --ip=%s --notebook-dir=%s --allow-root' % (
+        docker_cmd = 'jupyter notebook -y --ip=%s --notebook-dir=%s' % (
```
sagar8192: Just out of curiosity, why did we need it before?

huadongliu: Previously we didn't use --user for docker run. Root was the default user in the container to run jupyter.

paasta_tools/cli/cmds/spark_run.py
huadongliu: Yeah, the help message has it: 'Limit of total size in GB of serialized results of all partitions'

PR #1716, opened 2018-01-31
paasta deployd watches for new watches when autoscaling is enabled first time

chlgit:
When the watcher is built for the first time, the event would be set as None according to kazoo document. (http://kazoo.readthedocs.io/en/latest/api/recipe/watchers.html) 

Test this change at mesosstage, and the log shows deployd createed service instance after a new node had been added into zookeeper successfully:

2018-01-31 13:37:10,706:mesos1-uswest1cdevc:DEBUG:paasta_tools.deployd.watchers.AutoscalerWatcher:Node change: data = b'2', stat = ZnodeStat(czxid=253550824220, mzxid=253550824221, ctime=1517434630666, mtime=1517434630669, version=1, cversion=0, aversion=0, ephemeralOwner=0, dataLength=1, numChildren=0, pzxid=253550824220), event = None
2018-01-31 13:37:10,707:mesos1-uswest1cdevc:INFO:paasta_tools.deployd.watchers.AutoscalerWatcher:Number of instances changed or autoscaling enabled for first time for robj-test-service.main_1
2018-01-31 13:37:10,761:mesos1-uswest1cdevc:DEBUG:paasta_tools.deployd.common.PaastaQueue:Adding ServiceInstance(service='robj-test-service', instance='main_1', bounce_by=1517434630, watcher='AutoscalerWatcher', bounce_timers=None, failures=0, priority=0) to InboxQueue queue

### DIFF
--- a/paasta_tools/deployd/watchers.py
+++ b/paasta_tools/deployd/watchers.py
@@ -51,7 +51,7 @@ class AutoscalerWatcher(PaastaWatcher):
         """recursive nonsense"""
         if "autoscaling.lock" in path:
             return
-        self.log.debug("Adding watch on {}".format(path))
+        self.log.debug("Adding folder watch on {}".format(path))
         watcher = ChildrenWatch(self.zk, path, func=self.process_folder_event, send_event=True)
         self.watchers[path] = watcher
         children = watcher._prior_children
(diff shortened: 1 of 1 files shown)


paasta_tools/deployd/watchers.py
```diff
@@ -61,27 +61,31 @@ def watch_folder(self, path):
             for child in children:
                 self.watch_folder("{}/{}".format(path, child))
 
+    def _make_service_instance(self, path):
+        service, instance = path.split('/')[-3:-1]
```
solarkennedy: This sounds

paasta_tools/deployd/watchers.py
```diff
@@ -61,27 +61,31 @@ def watch_folder(self, path):
...
+            watcher=type(self).__name__,
+            failures=0,
+        )
+        self.inbox_q.put(service_instance)
+
     def watch_node(self, path):
-        self.log.debug("Adding watch on {}".format(path))
+        self.log.debug("Adding node watch on {}".format(path))
         DataWatch(self.zk, path, func=self.process_node_event, send_event=True)
+        self._make_service_instance(path)
```
solarkennedy: This sounds OK, but is won't this pick up on all random node watches, like changes to files that are not relevant?

Specifically, don't we only want to call this when the nodepath ends in "instances" ?

mattmb:
Thanks for taking a look at this, it's really appreciated!

When you were testing on mesosstage did you find that it *never* works without this patch? My understanding was that in general the code as it currently exists normally will add a watch on the newly created autoscaling nodes. But that sometimes something goes wrong and it "misses" creating the watch?

Another question: does the watcher get created but just not actioned on its first run (because of the event=None)? If so I think your fix explains it. But if you let the autoscaler create an `instances` key and change it and it still doesn't work then I'm surprised...

paasta_tools/deployd/watchers.py
mattmb: Should be okay because we only call this from a block that does an if on "instances"

paasta_tools/deployd/watchers.py
```diff
@@ -61,27 +61,31 @@ def watch_folder(self, path):
             for child in children:
                 self.watch_folder("{}/{}".format(path, child))
 
+    def _make_service_instance(self, path):
```
mattmb: Can you please call this `_enqueue_service_instance`

paasta_tools/deployd/watchers.py
chlgit: I think the check has been done at line 58:  " if children and ('instances' in children) "

paasta_tools/deployd/watchers.py
chlgit: Sure, will do

chlgit: I saw the watcher being built without this patch, and it called process_node_event. But since event=None, it doesn't pick up when autoscaling is enabled first time. For instances been changed case, autoscaler works fine even without this patch.

mattmb: Gotcha, I think this makes a lot of sense, it is for sure a bug!

I wonder if there's any other bugs that have a similar effect. Let's hope not!

Thanks again for taking a look :-)

PR #1717, opened 2018-02-01
Make paasta status display information about all marathon apps, even without -v.

EvanKrall:
I'm tired of having to do paasta status -v to see information relevant to the bounce.
Most of the time I don't actually care about the individual tasks, so I don't need to see that; I just need the numbers and statuses.


EvanKrall:
## During a bounce

### Old:
image

### New:
image


## Steady-state
### Old:
image

### New:
image

solarkennedy:
Just need to update the itests, they call some of these functions directly

PR #1722, opened 2018-02-06
Do a huge refactor to make get_expected_instance_count_for_service a lot faster

EvanKrall:
I recommend reviewing this per-commit, going backwards (last commit first).

- The raison d'être of this branch is the [third commit](https://github.com/Yelp/paasta/pull/1722/commits/d25962855ca6fa8adf5ca90c906d639d8c792ba2). Currently, `get_expected_instance_count_for_namespace` calls `load_marathon_service_config` once per instance. This ends up calling into service_configuration_lib.read_extra_service_information each time, which then does a copy.deepcopy on the cached parsed yaml. (At least we aren't loading the yaml N times.) This amounts to an O(N^2) running time based on the number of instances defined in your marathon.yaml: for each instance, copy a dict including each of the instances. This commit should change that into an O(N) algorithm, by using PaastaServiceConfigLoader, which produces many MarathonServiceConfigs from one copy of that dict. For yelp-main in prod, this currently takes about 4-5 seconds; this branch takes that to ~80ms.
- The [second commit](https://github.com/Yelp/paasta/pull/1722/commits/d28ecdfb8473f33a0c332abe6541ddbabdc5e529) renames PaastaServiceConfig to PaastaServiceConfigLoader, which I think is a better name -- the object loads configs, it is not itself a config.
- The [first commit](https://github.com/Yelp/paasta/pull/1722/commits/3b7b0705e7a95e98f759b6b4be0ca9cf45278732) makes several refactors (I probably should have done these in separate commits, but did not, and teasing that out after the fact would likely be very difficult. Sorry.):
  1. It changes the PaastaServiceConfig(Loader) code so that the user passes an InstanceConfig subclass (e.g. MarathonServiceConfig); this way, paasta_service_config(_loader).py does not need to import marathon_tools, and therefore marathon_tools can import paasta_service_config(_loader). So that PaastaServiceConfig can know what filename to look up for the particular config type specified, this commit makes the filename prefix a class variable.
  2. Because the AdhocJobConfig was using the v2 version of branch_dict, and PaastaServiceConfig no longer knows which version of branch_dict is appropriate, this commit changes all the other InstanceConfig types to also use the v2 style. (It's been ~15 months, it's time to finish this change.)
  3. Because a DeploymentsJson object either contained v1 or v2 information (never both), depending on whether `load_deployments_json` or `load_v2_deployments_json` was called, this commit also splits those into two different types.
  4. It makes all the keys of DeploymentsJsonV2 / BranchDictV2 required, so mypy will yell at you if you try to use an incomplete BranchDictV2. As part of this, the InstanceConfig interface changes, so that e.g. `load_marathon_service_config(..., load_deployments=False)` passes `branch_dict=None` instead of  `branch_dict={}`. This part of the change probably touched the most lines, particularly in tests.
  5. It makes DeploymentsJsonV{1,2} not subclass dict.

### DIFF
--- a/paasta_tools/check_marathon_services_replication.py
+++ b/paasta_tools/check_marathon_services_replication.py
@@ -350,7 +350,10 @@ def main():
 
     for service in list_services(soa_dir=args.soa_dir):
         service_config = PaastaServiceConfig(service=service, soa_dir=args.soa_dir)
-        for instance_config in service_config.instance_configs(cluster=cluster, instance_type='marathon'):
+        for instance_config in service_config.instance_configs(
+            cluster=cluster,
+            instance_type_class=marathon_tools.MarathonServiceConfig,
+        ):
             if instance_config.get_docker_image():
                 check_service_replication(
                     instance_config=instance_config,
--- a/paasta_tools/check_marathon_services_replication.py
+++ b/paasta_tools/check_marathon_services_replication.py
@@ -42,7 +42,7 @@ from paasta_tools import marathon_tools
 from paasta_tools import monitoring_tools
 from paasta_tools.marathon_tools import format_job_id
 from paasta_tools.mesos_tools import get_slaves
-from paasta_tools.paasta_service_config import PaastaServiceConfig
+from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.smartstack_tools import SmartstackReplicationChecker
 from paasta_tools.utils import _log
 from paasta_tools.utils import datetime_from_utc_to_local
@@ -349,7 +349,7 @@ def main():
     smartstack_replication_checker = SmartstackReplicationChecker(mesos_slaves, system_paasta_config)
 
     for service in list_services(soa_dir=args.soa_dir):
-        service_config = PaastaServiceConfig(service=service, soa_dir=args.soa_dir)
+        service_config = PaastaServiceConfigLoader(service=service, soa_dir=args.soa_dir)
         for instance_config in service_config.instance_configs(
             cluster=cluster,
             instance_type_class=marathon_tools.MarathonServiceConfig,
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -58,6 +58,7 @@ from paasta_tools.mesos_tools import filter_mesos_slaves_by_blacklist
 from paasta_tools.mesos_tools import get_mesos_network_for_net
 from paasta_tools.mesos_tools import get_mesos_slaves_grouped_by_attribute
 from paasta_tools.mesos_tools import mesos_services_running_here
+from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.secret_tools import get_hmac_for_secret
 from paasta_tools.secret_tools import is_secret_ref
 from paasta_tools.utils import _log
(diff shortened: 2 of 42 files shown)


tests/test_paasta_service_config.py
```diff
@@ -37,28 +37,38 @@ def create_test_service():
...
+    #     '%s:paasta-%s.example_child_job' % (TEST_SERVICE_NAME, TEST_CLUSTER_NAME): {
+    #         'docker_image': 'some_image', 'desired_state': 'start', 'force_bounce': None,
+    #     },
+    #     '%s:paasta-%s.sample_batch' % (TEST_SERVICE_NAME, TEST_CLUSTER_NAME): {
+    #         'docker_image': 'some_image', 'desired_state': 'start', 'force_bounce': None,
+    #     },
+    #     '%s:paasta-%s.interactive' % (TEST_SERVICE_NAME, TEST_CLUSTER_NAME): {
+    #         'docker_image': 'some_image', 'desired_state': 'start', 'force_bounce': None,
+    #     },
+    # })
```
EvanKrall: Oops, clean this up.

paasta_tools/marathon_tools.py
```diff
@@ -1370,16 +1370,15 @@ def get_expected_instance_count_for_namespace(
...
         soa_dir=soa_dir,
-    ):
-        srv_config = load_marathon_service_config(name, instance, cluster, soa_dir=soa_dir)
-        instance_ns = srv_config.get_nerve_namespace()
-        if namespace == instance_ns:
-            total_expected += srv_config.get_instances()
+        load_deployments=False,
+    )
+    for job_config in pscl.instance_configs(cluster=cluster, instance_type_class=MarathonServiceConfig):
+        if f"{service}.{namespace}" in job_config.get_registrations():
```
solarkennedy: This is great, and I think fixes a kind of long-standing bug of not counting canaries in the expected count!

paasta_tools/paasta_service_config.py
```diff
@@ -12,29 +12,32 @@
...
 from paasta_tools.utils import list_clusters
-from paasta_tools.utils import load_deployments_json
 from paasta_tools.utils import load_v2_deployments_json
 
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
 
 
+_InstanceConfig_T = TypeVar('_InstanceConfig_T', bound=InstanceConfig)
```
solarkennedy: I've never seen this before. Why do we have to do this kind of thing here and nowhere else?

solarkennedy:
* Re: deter raisin is great.
* PaastaServiceConfigLoader is a better name I agree
* 
i. I'm glad you were able to get this to work, the circular imports thing was a stumbling block for me before. Its just previously @Rob-Johnson objected to this, probably because it doesn't feel "pythonic"? (when you could just ask what kind of class it was in the first place instead?)  I'm ok with whatever, but the big benefit of not asking what kind of class you are, is that you don't have to import that class :)
ii/iii. This worries me the most. The reason we didn't move forward with the v1=>v2 change was that it broke start/stop. With your mypy safeguards it might be OK to do this again, but this is certainly the most dangerous aspect of this refactor. I don't know how better to "test" this than for you to run a generate-deployments-for-service on your local copy and look at it.
iv. this is fine
v. fine

paasta_tools/check_marathon_services_replication.py
```diff
@@ -349,8 +349,11 @@ def main():
     smartstack_replication_checker = SmartstackReplicationChecker(mesos_slaves, system_paasta_config)
 
     for service in list_services(soa_dir=args.soa_dir):
-        service_config = PaastaServiceConfig(service=service, soa_dir=args.soa_dir)
-        for instance_config in service_config.instance_configs(cluster=cluster, instance_type='marathon'):
+        service_config = PaastaServiceConfigLoader(service=service, soa_dir=args.soa_dir)
```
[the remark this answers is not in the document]
EvanKrall: Oops, missed this comment before merging. I made this change in #1732.

PR #1724, opened 2018-02-08
Rounding capacity to terminate too

matfra:
This prevents situations like that:

`2018-02-07 23:07:45,907:INFO:paasta_tools.autoscaling.autoscaling_cluster_lib.SpotAutoscaler.sfr-******************:Found 100.00% slaves registered in mesos for this resource (31/31)
2018-02-07 23:07:48,384:ERROR:paasta_tools.autoscaling.autoscaling_cluster_lib.SpotAutoscaler.sfr-******************:Didn't find enough candidates to kill. This shouldn't happen so let's not kill anything! Amount wanted to decrease: 2.4105, killable capacity: 2.41`


mattmb:
That other defence I was thinking of is later actually: https://github.com/Yelp/paasta/blob/64d72c2a0b51bc6edbe98192150a98783b5115c0/paasta_tools/autoscaling/autoscaling_cluster_lib.py#L679-L682

So this fix is definitely real. Weird how it hasn't come up before?

PR #1726, opened 2018-02-09
Use mesos 1.5.0 in itests

nhandler:
This is PAASTA-13910


nhandler:
Tests are unhappy (https://travis-ci.org/Yelp/paasta/jobs/339663613). I'll manually spin up the new version of mesos to try and debug.

PR #1730, opened 2018-02-13
Make sure PaastaServiceConfig, load_marathon_service_config, and load_{chronos,adhoc,paasta_native}_job_config correctly interpolate facts into deploy_group before loading branch dict.

EvanKrall:


solarkennedy:
I feel like we are doing it wrong, but I don't know how to do it better.

We need to parse the stuff to figure out the deploy_group, because we can't come up with the `branch_dict` object without the deploy_group. Maybe branch_dict shouldn't be in the init of the class and should be something that adds to the object afterwards? I don't know.

EvanKrall:
I think this will be a lot better once we unify onto PaastaServiceConfigLoader -- all this duplication can go away.

PR #1732, opened 2018-02-14
Have PaastaServiceConfigLoader.get_instances() skip instances which have no version marked for deployment.

EvanKrall:
Another breakage from #1722: anything that calls `PaastaServiceConfigLoader.instance_configs()` (including `check_marathon_services_replication`) now gets a `NoDeploymentsAvailable` exception if there's any instances for the service/cluster in question that do not have a corresponding entry in deployments.json. The v1 deployments json object (which we used to use in PaastaServiceConfigLoader for marathon configs) returns an empty branch dict instead of raising an error.

This has `instance_configs` simply skip over any instances where `NoDeploymentsAvailable` gets raised. This does mean that the set of instances returned will be different depending on whether `PaastaServiceConfigLoader` is initialized with `load_deployments=True` or `False`.

### DIFF
--- a/paasta_tools/paasta_service_config_loader.py
+++ b/paasta_tools/paasta_service_config_loader.py
@@ -28,6 +28,7 @@ from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import InstanceConfig
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import load_v2_deployments_json
+from paasta_tools.utils import NoDeploymentsAvailable
 
 
 log = logging.getLogger(__name__)
@@ -115,7 +116,10 @@ class PaastaServiceConfigLoader():
         if (cluster, instance_type_class) not in self._framework_configs:
             self._refresh_framework_config(cluster, instance_type_class)
         for instance, config in self._framework_configs.get((cluster, instance_type_class), {}).items():
-            yield self._create_service_config(cluster, instance, config, instance_type_class)
+            try:
+                yield self._create_service_config(cluster, instance, config, instance_type_class)
+            except NoDeploymentsAvailable:
+                pass
 
     def _framework_config_filename(self, cluster: str, instance_type_class: Type[_InstanceConfig_T]):
         return "%s-%s" % (instance_type_class.config_filename_prefix, cluster)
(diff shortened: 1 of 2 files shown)


paasta_tools/paasta_service_config_loader.py
```diff
@@ -115,7 +116,10 @@ def instance_configs(
         if (cluster, instance_type_class) not in self._framework_configs:
             self._refresh_framework_config(cluster, instance_type_class)
         for instance, config in self._framework_configs.get((cluster, instance_type_class), {}).items():
-            yield self._create_service_config(cluster, instance, config, instance_type_class)
+            try:
+                yield self._create_service_config(cluster, instance, config, instance_type_class)
+            except NoDeploymentsAvailable:
```
solarkennedy: This is kinda a departure from how we used to do it, where all the callers *must* catch NoDeployments and skip. 
I like the idea of never even bothering to show them this (pass). It is very rare that we are scratching our heads wondering why something is not happening because it is not deployed. (but it happened earlier when paasta was in its infancy)

EvanKrall: I'd prefer to keep the exception handling in the caller, but there's no clean way to do it with a generator like this. If you raise an exception, then any instances that would come afterwards will never get yielded.

I _could_ do something like yielding functions, which either return the instance config or raise NoDeploymentsAvailable, but I think that'd be a weird interface.

paasta_tools/check_marathon_services_replication.py
```diff
@@ -349,22 +350,30 @@ def main():
...
+                instance_config=instance_config,
+                all_tasks=all_tasks,
+                smartstack_replication_checker=smartstack_replication_checker,
+            )
+
+        all_instances = {compose_job_id(service, i) for i in service_config_loader.instances(
+            cluster=cluster,
+            instance_type_class=marathon_tools.MarathonServiceConfig,
+        )}
+        undeployed_instances = all_instances - deployed_instances
```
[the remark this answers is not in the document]
EvanKrall: K, I'll remove that. It turns out this pretty much goes nowhere anyway, as we don't show debug logging by default.

PR #1733, opened 2018-02-15
Slightly better "paasta status" output for a bad service name

chriskuehl:
This confused someone a bit earlier.

### Before:

```
$ paasta status -c norcal-prod -s yelp_main -i my_important_kew_worker
yelp_main doesn't have any instances matching my_important_kew_worker on norcal-prod.
```

(This makes it look like you misspelled the instance rather than the service.)

### After:

```
$ paasta status -c norcal-prod -s yelp_main -i my_important_kew_worker
The service "yelp_main" does not exist.
Did you mean any of these?
  yelp-main
  [another suggestion]
  [another suggestion]
  [another suggestion]
  [another suggestion]
```

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -385,8 +385,19 @@ def apply_args_filters(args):
 
     filters = get_filters(args)
 
+    all_services = list_services(soa_dir=args.soa_dir)
+
+    if args.service and args.service not in all_services:
+        paasta_print(PaastaColors.red(f'The service "{args.service}" does not exist.'))
+        suggestions = difflib.get_close_matches(args.service, all_services, n=5, cutoff=0.5)
+        if suggestions:
+            paasta_print(PaastaColors.red(f'Did you mean any of these?'))
+            for suggestion in suggestions:
+                paasta_print(PaastaColors.red(f'  {suggestion}'))
+        return {}
+
     i_count = 0
-    for service in list_services(soa_dir=args.soa_dir):
+    for service in all_services:
         if args.service and service != args.service:
             continue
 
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -385,8 +385,19 @@ def apply_args_filters(args):
...
+        suggestions = difflib.get_close_matches(args.service, all_services, n=5, cutoff=0.5)
+        if suggestions:
+            paasta_print(PaastaColors.red(f'Did you mean any of these?'))
+            for suggestion in suggestions:
+                paasta_print(PaastaColors.red(f'  {suggestion}'))
+        return {}
+
     i_count = 0
-    for service in list_services(soa_dir=args.soa_dir):
+    for service in all_services:
```
solarkennedy: It seems weird to me that this thing is so flexible enough to allow multiple services like this.

Do we still need to loop like this?

paasta_tools/cli/cmds/status.py
```diff
@@ -385,8 +385,19 @@ def apply_args_filters(args):
...
         if args.service and service != args.service:
```
solarkennedy: Likewise this, is it now irrelevant?

paasta_tools/cli/cmds/status.py
chriskuehl: It looks like we actually always only have one service (`args.service` is always set by the code above, this is probably a remnant from some time when it wasn't). I'll try to clean this up.

chriskuehl: actually we don't always have a service, e.g. if you do `paasta status -c norcal-prod -i main --owner webcore`, so it probably really is necessary to loop like this.

paasta_tools/cli/cmds/status.py
chriskuehl: I don't think so, for the same reason as above (you can specify an owner instead of a service to see the status of multiple services at once). I didn't even realize this was possible...

paasta_tools/cli/cmds/status.py
solarkennedy: Ah

PR #1734, opened 2018-02-15
Replaced spark-run docker envs with command line options

huadongliu:
This change simplifies the Spark service Dockerfile.

~/paasta/.tox/py36/bin/paasta spark-run --build --cluster norcal-devc --cmd pyspark
~/paasta/.tox/py36/bin/paasta spark-run --build --cluster norcal-devc --cmd jupyter
~/paasta/.tox/py36/bin/paasta spark-run --build --cluster norcal-devc --cmd "spark-submit /spark_client/examples/pi.py"

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -65,10 +65,16 @@ def add_subparser(subparsers):
     list_parser.add_argument(
         '-c', '--cluster',
         help=(
-            "The name of the cluster you wish to run Spark on. "
+            "The name of the cluster you wish to run Spark on."
         ),
-        choices=['norcal-devc', 'pnw-devc'],
+        choices=['norcal-devc', 'pnw-devc', 'mesosstage'],
         required=True,
+    )
+
+    list_parser.add_argument(
+        '-p', '--pool',
+        help="Name of the resource pool to run the Spark job.",
+        default='default',
     )
 
     list_parser.add_argument(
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -154,47 +160,76 @@ def get_docker_run_cmd(
...
+    cmd,
+    spark_conf,
+):
+    spark_env = {}
+
+    creds = Session().get_credentials()
+    spark_env['AWS_ACCESS_KEY_ID'] = creds.access_key
+    spark_env['AWS_SECRET_ACCESS_KEY'] = creds.secret_key
+
+    if cmd == 'jupyter':
```
solarkennedy: You probably need

    if 'jupyter' in cmd:

if you want this to be tolerant of many forms of running a notebook, no?

Why not just set these regardless though?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -154,47 +160,76 @@ def get_docker_run_cmd(
...
+    # spark_conf.append('--conf spark.mesos.principal=')
+
+    # derby.system.home property defaulting to '.',
+    # which requires directory permission changes.
+    spark_conf.append('--conf spark.driver.extraJavaOptions=-Dderby.system.home=/tmp/derby')
 
-    return spark_conf
+    spark_conf.append('--conf spark.mesos.constraints=pool:%s' % args.pool)
+
+    spark_conf.append('--conf spark.mesos.executor.docker.volumes=%s' % ','.join(volumes))
```
solarkennedy: Does this work? I would be surprised this works without complex quoting

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -154,47 +160,76 @@ def get_docker_run_cmd(
...
+    if not args.build:
+        spark_conf.append('--conf spark.mesos.uris=file:///root/.dockercfg')
+
+    # TODO PAASTA-13815
+    # spark_conf.append('--conf spark.mesos.secret=')
+    # spark_conf.append('--conf spark.mesos.principal=')
+
+    # derby.system.home property defaulting to '.',
+    # which requires directory permission changes.
+    spark_conf.append('--conf spark.driver.extraJavaOptions=-Dderby.system.home=/tmp/derby')
```
solarkennedy: What was this set to before? Does it need /tmp/derby to exist first?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -241,6 +276,20 @@ def configure_and_run_docker_container(
                     "Warning: Path %s does not exist on this host. Skipping this binding." % volume['hostPath'],
                 ),
             )
+
+    spark_ui_port = pick_random_port(args.service)
+    container_name = 'paasta_spark_run_%s_%s' % (get_username(), spark_ui_port)
+
+    spark_conf_str = get_spark_conf_str(
```
solarkennedy: use kwargs

paasta_tools/cli/cmds/spark_run.py
huadongliu: Previously this was not set. As a result, we did "RUN chmod -R o+w /work" in the Dockerfile to resolve the permission issue because we run spark-run as non-root. "RUN chmod -R o+w /work" will be removed with this paasta_tools release.

paasta_tools/cli/cmds/spark_run.py
huadongliu: Yeah, spark.mesos.executor.docker.volumes work fine without complex quoting, https://pb.yelpcorp.com/311715. If I don't set spark.mesos.executor.docker.volumes, this example will fail with ServiceKafkaDiscovery exceptions since the executor looks for KafkaDiscovery files.

paasta_tools/cli/cmds/spark_run.py
huadongliu: At this point, we launch the jupyter notebook server with the options that we pick (e.g. directories etc). There will be conflicts if we let users start jupyter with arbitrary options.

paasta_tools/cli/cmds/spark_run.py
huadongliu: Fixed.

PR #1736, opened 2018-02-21
Humanize floats in metastatus for cpu/gpu info

### LINKED ISSUE #1685: Round Metastatus Stats
There is no reason to show something like `0.09999999999999787/15.0 (0.67%)` in paasta metastatus. We should round/truncate that value.

matthewbentley:
fixes #1685

### DIFF
--- a/paasta_tools/metrics/metastatus_lib.py
+++ b/paasta_tools/metrics/metastatus_lib.py
@@ -27,6 +27,7 @@ from typing import TypeVar
 
 from humanize import naturalsize
 from mypy_extensions import TypedDict
+from typing_extensions import Counter as _Counter  # noqa
 
 from paasta_tools import chronos_tools
 from paasta_tools.mesos_maintenance import MAINTENANCE_ROLE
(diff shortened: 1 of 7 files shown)


paasta_tools/metrics/metastatus_lib.py
```diff
@@ -736,6 +735,12 @@ def format_table_column_for_healthcheck_resource_utilization_pair(healthcheck_ut
             naturalsize(healthcheck_utilization_pair[1].total * 1024 * 1024, gnu=True),
             utilization_perc,
         ))
+    elif humanize and healthcheck_utilization_pair[1].metric in ['cpus', 'gpus']:
```
solarkennedy: Do we ever not humanize?

Let's just humanize all the time and remove the complexity?

paasta_tools/paasta_metastatus.py
```diff
@@ -73,8 +73,8 @@ def parse_args(argv):
         help="Print out more output regarding the state of the cluster",
     )
     parser.add_argument(
-        '-H', '--humanize', action='store_true', dest="humanize", default=False,
-        help="Print human-readable sizes",
+        '-H', '--humanize', action='store_true', dest="humanize", default=True,
```
solarkennedy: Just remove it

PR #1738, opened 2018-02-21
Run spark and mesos framework as root

huadongliu:
.tox/py36/bin/paasta spark-run --cluster norcal-devc --cmd pyspark


nhandler:
Can we get some test coverage around this to make sure we don't break this functionality with future upgrades?

huadongliu:
@nhandler I have created a ticket to write an integration for spark-run. I am going to make a release to roll this out first since spark users are now being blocked.

PR #1742, opened 2018-02-23
Use the operator api for maintenance, draining, and up/down

matthewbentley:
Possibly needs more tests, and maybe can remove old code

### DIFF
--- a/paasta_tools/mesos_maintenance.py
+++ b/paasta_tools/mesos_maintenance.py
@@ -83,7 +83,8 @@ def operator_api():
             kwargs['headers']['Content-Type'] = 'application/json'
         else:
             kwargs['headers'] = {'Content-Type': 'application/json'}
-        return base_api_client("POST", "/api/v1", **kwargs)
+        data = kwargs.pop('data')
+        return base_api_client("POST", "/api/v1", data=json.dumps(data), **kwargs)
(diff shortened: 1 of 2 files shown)


paasta_tools/mesos_maintenance.py
```diff
@@ -76,6 +76,18 @@ def execute_master_api_request(method, endpoint, **kwargs):
     return execute_master_api_request
 
 
+def operator_api():
+    def execute_operator_api_request(**kwargs):
+        base_api_client = base_api()
+        if 'headers' in kwargs.keys():
```
nhandler: Is the `.keys()` necessary for some reason? Can't you just do `if 'headers' in kwargs`?

paasta_tools/mesos_maintenance.py
```diff
@@ -319,12 +332,17 @@ def datetime_to_nanoseconds(dt):
     return seconds_to_nanoseconds(int(dt.strftime("%s")))
 
 
-def build_start_maintenance_payload(hostnames):
+def build_start_maintenance_payload(hostnames, maint_type):
```
nhandler: Can we rename this function? It is a bit confusing calling `build_start_maintenance_payload` with a `main_type` of `stop_maintenance`.

PR #1744, opened 2018-02-27
Moved SPARK_OPTS out from jupyter

huadongliu:
yelp-dm has a different way of launching a jupyter notebook server. It becomes a general use case when spark-run is used on a service other than the default services/spark. This change is to make SPARK_OPTS available for non-default jupyter commands.


solarkennedy:
This is great, I'm glad to see more non-jupyter spark applications.

PR #1746, opened 2018-03-01
Added an option to mount a different spark driver work dir

huadongliu:
spark-run mounts $PWD:/spark_client:rw at the moment. There are cases where users want to mount with a different local_path:container_path. This change makes $PWD:/spark_driver:rw as default, and allows users to mount a different directory with --work-dir.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -1,3 +1,4 @@
+import argparse
 import json
 import os
 import socket
@@ -26,7 +27,7 @@ from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import PaastaNotConfiguredError
 from paasta_tools.utils import SystemPaastaConfig
 
-DEFAULT_SPARK_WORK_DIR = '/spark_client'
+DEFAULT_SPARK_WORK_DIR = '/spark_driver'
 DEFAULT_SPARK_DOCKER_IMAGE_PREFIX = 'paasta-spark-run'
 DEFAULT_SPARK_DOCKER_REGISTRY = 'docker-dev.yelpcorp.com'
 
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -386,7 +398,23 @@ def build_and_push_docker_image(args):
     return docker_url
 
 
+def validate_work_dir(s):
+    dirs = s.split(':')
+    if len(dirs) != 2:
+        msg = "--work-dir requires format local_abs_dir:container_abs_dir"
```
solarkennedy: This can be better. Please add the actual (s) that was provided, to hint at the user "what" needs to change exactly.

The raising, argparse.ArgumentTypeError is a very pythony thing to do, but isn't very user friendly. Does it show a stacktrace to the user?

If so, then please do something else and just exit. We don't need to show a whole scary stacktrace on a "normal" error.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -386,7 +398,23 @@ def build_and_push_docker_image(args):
...
+        raise argparse.ArgumentTypeError(msg)
+
+    for d in dirs:
+        msg = "--work-dir requires absolute local and container paths"
```
solarkennedy: Same here.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -82,6 +83,16 @@ def add_subparser(subparsers):
         default='default',
     )
 
+    list_parser.add_argument(
+        '-w', '--work-dir',
+        default='%s:%s' % (os.getcwd(), DEFAULT_SPARK_WORK_DIR),
+        help=(
+            "The read-write volume to mount in format "
+            "local_abs_dir:container_abs_dir, defaults "
+            "to $PWD:/spark_drvier."
```
solarkennedy: Do we already use a default printing argparse?
https://stackoverflow.com/questions/12151306/argparse-way-to-include-default-values-in-help

If not, this would be a fancier way to do this

PR #1748, opened 2018-03-02
Fix docker_ commands container name

mattmb:
The format mesos uses for the container name they pass to docker seems to
have changed in a recent version of mesos. This fixes our docker_
commands which rely on using the container name.

### DIFF
--- a/paasta_tools/cli/utils.py
+++ b/paasta_tools/cli/utils.py
@@ -1054,7 +1054,7 @@ def get_task_from_instance(cluster, service, instance, slave_hostname=None, task
 
 
 def get_container_name(task):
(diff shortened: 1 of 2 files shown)


mattmb:
https://github.com/apache/mesos/blob/a4805eff5dbef7146f34b239b834a688954991f9/src/slave/containerizer/docker.cpp#L128-L136

Happily no explanation on why *sigh*

matthewbentley:
It looks like this requires mesos >= 1.4 everywhere?

mattmb:
Yup, it's true. I guess this will break things in the opposite set of clusters to where it's broken now.

I'd rather not add code to diff on version or do anything more tricky though. I think we're nearly on 1.4+ everywhere...

PR #1751, opened 2018-03-05
Always return http response with bravado client, bump bravado

solarkennedy:
Newer versions of bravado hint that they understand these "extra" arguments and should remove the warning.

### DIFF
--- a/paasta_tools/api/client.py
+++ b/paasta_tools/api/client.py
@@ -29,7 +29,7 @@ from paasta_tools.utils import load_system_paasta_config
 log = logging.getLogger(__name__)
 
 
-def get_paasta_api_client(cluster=None, system_paasta_config=None, http_res=False):
+def get_paasta_api_client(cluster=None, system_paasta_config=None):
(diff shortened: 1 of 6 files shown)


solarkennedy:
Itests still failing 

```
Running cmd ../paasta_tools/monitoring/check_capacity.py mem --crit 99 --warn 90
  File "../paasta_tools/monitoring/check_capacity.py", line 173, in <module>
    run_capacity_check()
  File "../paasta_tools/monitoring/check_capacity.py", line 146, in run_capacity_check
    check = get_check_from_overrides(overrides, default_check, usage_value['groupings'])
TypeError: list indices must be integers or slices, not str
```

huadongliu:
Sorry I missed context on the ticket previously. I don't think upgrading bravado will resolve the ticket. I added a comment in the ticket. Let's see what bravado folks say. We may have to live with the warning.

The itest failure is on a case where also_return_response is not needed.

huadongliu:
This can be closed in favor of PR 1761.

PR #1753, opened 2018-03-05
Added options for mesos principal and secret for spark-run

huadongliu:


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -30,6 +30,7 @@
 DEFAULT_SPARK_WORK_DIR = '/spark_driver'
 DEFAULT_SPARK_DOCKER_IMAGE_PREFIX = 'paasta-spark-run'
 DEFAULT_SPARK_DOCKER_REGISTRY = 'docker-dev.yelpcorp.com'
+DEFAULT_SPARK_MESOS_SECRET_DIR = '/nail/etc/paasta_spark_secret'
```
solarkennedy: Why do you call this `DIR`? Isn't it just a normal file?

PR #1754, opened 2018-03-06
Add paasta secret cli module

mattmb:
* Extend secret provider so that it can write secrets to your local
checkout of yelpsoa-configs
* Extend secret provider so it can decrypt secrets (just useful to have)
* Add cli module to call these secret provider methods

the motivation is to deprecate the paasta_secret cli which is in the
internal vault_tools package in favour of this which is generic and
imports the vault_tools package for the vault provider (which is
currently the only one implemented).

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -1,0 +1,175 @@
+#!/usr/bin/env python
+# Copyright 2015-2018 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import os
+import sys
+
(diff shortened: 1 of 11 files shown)


paasta_tools/cli/cmds/secret.py
```diff
@@ -0,0 +1,172 @@
...
+            "You must run this tool from the root of your local yelpsoa checkout\n"
+            "The tool modifies files in yelpsoa-configs that you must then commit\n"
+            "and push back to git.",
+        )
+        sys.exit(1)
+    system_paasta_config = load_system_paasta_config()
+    secret_provider_kwargs = {
+        'vault_cluster_config': system_paasta_config.get_vault_cluster_config(),
+    }
+    clusters = cluster_names.split(',') if cluster_names else list_clusters(service_name, os.getcwd())
```
solarkennedy: kwargs please, way easy to get the order wrong on list_clusters

paasta_tools/cli/cmds/secret.py
```diff
@@ -0,0 +1,172 @@
...
+
+def paasta_secret(args):
+    secret_provider = _get_secret_provider_for_service(args.service, cluster_names=args.clusters)
+    if args.action in ["add", "update"]:
+        secret_provider.write_secret(
+            action=args.action,
+            secret_name=args.secret_name,
+            plaintext=get_plaintext_input(args),
+        )
+        secret_path = os.path.join(os.getcwd(), args.service, "secrets", "{}.json".format(args.secret_name))
```
solarkennedy: Is there any place where this path construction is already defined and can be reused?

mattmb: Not entirely but it can be simplified a bit 👍

PR #1755, opened 2018-03-08
Add PAASTA_MONITORING_TEAM environment variable

billyevans:
It gets team from get_team() method. Otherwise if that wasn’t set in the configuration the environment variable won’t be set.


mattmb:
Heya, the change itself is okay. Just wanted to give you a heads up that this is an example of something that's a little bit tricky to roll out. The reason being that when you push this change paasta will have to restart every container it manages to add this environment variable. We have ways to manage this but it requires a bit of planning. If I haven't scared you off then hit me up on IRC and I can tell you a bit more.

solarkennedy:
This looks fine. We do need to be prepared when we ship this as it will restart (bounce) all containers over time to get this new variable.

PR #1759, opened 2018-03-09
Don't add .git to default yelp git urls. Fixes #1756

### LINKED ISSUE #1756: Generate-pipeline passes repo suffix'd with .git to fab_repo
fab_repo expects that the repo is passed without the `.git` suffix but the url passed to fab_repo by paasta includes the suffix => [get_git_repo_for_fab_repo](https://github.com/Yelp/paasta/search?utf8=%E2%9C%93&q=get_git_repo_for_fab_repo&type=)

The fix is to remove the last 4 characters of this string (.git)

solarkennedy:
cc @acoover


acoover:
@solarkennedy yup, this should resolve it. Thanks for making this change

PR #1761, opened 2018-03-12
Pin all dependencies, introduce requirements-tools

sjaensch:
As you can see from the PR, there are an awful lot of dependencies not pinned at all. In addition to that, a bunch of direct dependencies were not listed in setup.py's `install_requires` section. With this PR paasta gets reproducible builds: all Python packages are pinned to its exact version. At the same time, just installing it (with `pip install`) should now yield a working installation. I've added requirements-tools in order to make sure we continue to pin all dependencies. Unfortunately, this requires duplicating the `install_requires` section in the `requirements-minimal.txt` file.

If we're fine with saying that people should create their own virtualenv from requirements.txt anyway then we can remove the section from setup.py. I assume nobody is installing paasta-tools in that way anyway. (n.b. the version on PyPI is out of date)

Note: this change also bumps the bravado version in order do fix a warning when used with newer versions of bravado-core (which up to this PR wasn't pinned).


solarkennedy:
@sjaensch the only other place that imports paasta-tools is https://github.com/Yelp/synapse-tools and https://github.com/Yelp/nerve-tools. 

I think this will increase the burden of those tools. This is "probably" ok?

sjaensch:
@solarkennedy how does it increase the burden on these tools? Things should continue to work as-is for them. They use the "pip install" way of using paasta-tools, so all that changes for them is that they'll install a few more packages (like boto3, croniter, task-processing etc.). But to be fair, those are actual dependencies of paasta-tools; synapse-tools and nerve-tools just happen to not use those parts of the code (or they happen install them due to other dependencies).

sjaensch:
But yes, it means we probably shouldn't remove the `install_requires` part of setup.py.

PR #1765, opened 2018-03-13
Always mock out time.sleep in all tests

solarkennedy:


solarkennedy:
This is good and speeds up the test suite by a few seconds.

PR #1766, opened 2018-03-19
[RFC] paasta api authentication

mattmb:
This is quite rough around the edges (or in other words I'm not super
happy about the implementation in its current form). Lacking tests etc.
I'm just opening this to get some opinions and see if anyone thinks it
is worth tidying up/working on for real.

The rough idea is to include a vault token in requests to the paasta API
as a header. The server then validates that the token is legit with
Vault and checks what policies are attached to the token. These vault
policies are then mappped to permissions on each api view.

I've also wrapped the client to generate + send the correct token in
response to 403s.

### DIFF
--- a/example_cluster/paasta/api.json
+++ b/example_cluster/paasta/api.json
@@ -1,5 +1,5 @@
 {
   "api_endpoints": {
-    "testcluster": "http://mesosmaster:5054"
+    "testcluster": "http://paasta_api:5054"
   }
 }
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -23,12 +23,19 @@ import requests_cache
 import service_configuration_lib
 from gevent import monkey
 from gevent.wsgi import WSGIServer
+from pyramid.authorization import ACLAuthorizationPolicy
 from pyramid.config import Configurator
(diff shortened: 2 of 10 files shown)


solarkennedy:
Seems like a lot of code. I don't suppose we can put this stuff in nginx instead? :)

mattmb:
Were you thinking it would be cool to have it in nginx because we could make something generic to all services? I was wondering if something like that would be possible/worthwhile.

I have a feeling it would end up as more code doing it in nginx though tbh. Here we are getting quite a bit for "free" from pyramid. In terms of a structure for the ACLs, endpoint permissions and so on.

I think it's the client side stuff that's the tricky bit. What I've done is wrap the swagger client in such a way that it prompts the CLI user if the auth fails. But making that robust and suitable for automated users too will be a bit fiddly. Some of the bravado related code is just boiler plate though, so it's a lot of LOC but not much cognitive overhead (maybe).

solarkennedy:
@mattmb would you like to close this given our recent discussions? (in favor of a new branch)

mattmb:
👍 sounds good. I'll leave the branch around because there's some stuff there we may go back to.

PR #1768, opened 2018-03-19
Fix a spacing issue

ealter:
Previous output:

`This service will soon be gracefully started/restarted, replacing old instances accordingto the bounce method chosen in soa-configs.`


Rob-Johnson:
thanks!

PR #1773, opened 2018-03-22
Only use sudo with paasta local-run when there is no docker config available

solarkennedy:
I don't want people blindly using sudo all the time, so this check only makes it use sudo when we "have to".

Technically you could use paasta without a registry that requires auth.

Hopefully this change will be an ok "compromise" between magic and non-surprising behavior.

But really I don't want auto-sudo so Tron can run this command *and* have the permission to kill it later.

cc @qui cc @bobtfish

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -923,8 +923,16 @@ def configure_and_run_docker_container(
     )
 
 
+def docker_config_available():
+    home = os.path.expanduser('~')
+    oldconfig = os.path.join(home, ".dockercfg")
+    newconfig = os.path.join(home, ".docker", "config.json")
+    return (os.path.isfile(oldconfig) and os.access(oldconfig, os.R_OK)) or \
+        (os.path.isfile(newconfig) and os.access(newconfig, os.R_OK))
+
+
 def paasta_local_run(args):
-    if args.action == 'pull' and os.geteuid() != 0:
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -20,6 +20,7 @@
 import time
 import uuid
 from os import execlp
+from pathlib import Path
```
nhandler: Would [os.path.expanduser](https://docs.python.org/3/library/os.path.html#os.path.expanduser) not be sufficient? i.e. `os.path.expanduser('~/.dockercfg')`

paasta_tools/cli/cmds/local_run.py
```diff
@@ -923,8 +924,15 @@ def configure_and_run_docker_container(
     )
 
 
+def docker_config_available():
+    home = Path.home()
+    oldconfig = os.path.join(home, ".dockercfg")
+    newconfig = os.path.join(home, ".docker", "config.json")
+    return os.path.isfile(oldconfig) or os.path.isfile(newconfig)
```
nhandler: To be safe, it might be better to check if the file is readable rather than whether it merely exists.

solarkennedy: Got it

paasta_tools/cli/cmds/local_run.py
solarkennedy: Yea, looks like the same. I'll use that method to save an import

PR #1784, opened 2018-03-28
Speed up paasta_serviceinit status for many instances.

### LINKED ISSUE #1783: `paasta_serviceinit status` is CPU-bound
It takes a very long time to run `paasta status` on services with many marathon instances defined. While there are ways to filter the instances you look at, this is a bit silly:

```
krall@10-64-190-53-uswest1bprod:~$ sudo time paasta_serviceinit -s yelp-main -i `( uq /nail/etc/services/yelp-main/marathon-norcal-prod.yaml ) | jq -r 'keys[]' | tr '\n' ',' | sed 's/,$//' ` status
...
207.69user 4.01system 3:34.62elapsed 98%CPU (0avgtext+0avgdata 263496maxresident)k
```

98% cpu!!! Three and a half minutes!!!

I think we had assumed for a long time that `status` was slow because it did too many HTTP calls. In reality, we're actually doing a really good job of caching the HTTP calls, but we're doing a bad job of caching things like `load_system_paasta_config` and `load_marathon_service_config`, and even though we're caching the JSON responses from e.g. Marathon, for each instance we're re-creating the marathon-python objects for every other instance. This amounts to an O(N^2) algorithm, and N has gotten pretty large for us.

EvanKrall:
Fixes #1783.

This reduces the time required to do an unfiltered `paasta status` by reducing the amount of CPU work necessary.

- We were already caching the JSON that Marathon returns for list_apps, but marathon_python was re-creating all the MarathonApp and other objects for each instance; now we cache those objects.
- We were calling load_system_paasta_config for each instance, now we cache that.
- We were calling load_service_namespace_config twice for every instance; now we pass around service_namespace_config
- We were calling load_marathon_service_config for each instance, which means we have to re-parse marathon-{cluster}.yaml for each instance; now paasta_serviceinit uses PaastaServiceConfigLoader to parse once and passes in the MarathonServiceConfig objects.
- We weren't filtering the call to marathon /apps, so marathon returned info about all services. Now we filter to the service, which is hopefully a good balance between many API calls and parsing too much JSON.


paasta_tools/marathon_serviceinit.py
```diff
@@ -528,7 +526,7 @@ def perform_command(
             return 1
 
     normal_instance_count = job_config.get_instances()
-    proxy_port = marathon_tools.get_proxy_port_for_instance(service, instance, cluster, soa_dir=soa_dir)
+    # proxy_port = marathon_tools.get_proxy_port_for_instance(service, instance, cluster, soa_dir=soa_dir)
```
solarkennedy: junk?

paasta_tools/utils.py
```diff
@@ -1426,7 +1426,12 @@ def get_readable_files_in_glob(glob: str, path: str) -> List[str]:
 )
 
 
+@time_cache(ttl=60)
```
solarkennedy: Do you think this will mess up deployd? Because it has a inotify watcher on this thing, so when it gets and event and.. calls this function nothing will be different maybe?

EvanKrall: That's a good point, I'll have to think about how to work around that.

EvanKrall:
@solarkennedy Can you take a look at the e88eb62 commit when you get a chance?

PR #1798, opened 2018-04-12
Added timeout for spark-run Jupyter kernels

huadongliu:
With this change, Spark frameworks without a Jupyter notebook connection will be terminated if the Jupyter kernel has been idle for more than one hour. Frameworks with a Jupyter notebook connection but using 32+ CPUs will also get terminated.


huadongliu:
Yeah, 32 is just an arbitrary number. Let's see how it works.

PR #1799, opened 2018-04-13
Add script to monitor for containers sharing an IP

qui:
This was causing some requests to get routed to the incorrect service. Planning to add a monitoring check with this.


paasta_tools/contrib/shared_ip_check.py
```diff
@@ -0,0 +1,40 @@
...
+from collections import defaultdict
+
+from paasta_tools.utils import get_docker_client
+
+
+def main():
+    docker_client = get_docker_client()
+    ip_to_containers = defaultdict(list)
+
+    for container in docker_client.containers():
```
solarkennedy: The docs for this version of docker-py is 404 now, does this return only running containers?

qui: Good point. Just tested it on a machine with stopped containers, looks like it does only return running containers.

solarkennedy:
Keep in mind this script will be in the "bin" folder of the venv, but it won't be in the global path. I think that is fine (especially given the name of this script), but we will have to execute it via its full path on disk.

PR #1805, opened 2018-04-18
don't abort downscaling an SFR/ASG if there are unregistered instance…

stug:
…s, the second part

This prevents metrics_provider() from bailing out when there are more unregistered instances than it wants to see if we're scaling down -- makes sure we get to the logic in https://github.com/Yelp/paasta/pull/1792


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -326,8 +326,6 @@ def check_expected_slaves(
             if float(current_instances) / expected_instances < (1.00 - MISSING_SLAVE_PANIC_THRESHOLD):
```
huadongliu: I think the autoscaler only refuses to scale when (current/expected < 1-missing_slave_panic_threshold) in scaling up. It can continue in all other cases, including (current/expected < 1) and is_scaling_down. 

When the autoscaler detects (current<expected), it is an opportunity to emit this data to signalfx.

sagar8192: @stug It would nice if you can add this to paasta tools. We can have paasta find-orphans or something. This way, we(operators) can use it and see the instances that not part of the cluster.  You just need to lookup instances with Paasta IAM role to find the aws instances query mesos-master /slaves API to find the agents registered, and print the delta between the two. We can also have an option to emit metrics to Signalfx using '--emit-metrics' flag.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -336,7 +334,18 @@ def check_expected_slaves(
                         error_message + "But this is an sfr in cancelled_running state, so continuing anyways",
                     )
                     return None
-                error_message += "(refusing to go past %.2f%% missing instances)" % MISSING_SLAVE_PANIC_THRESHOLD
+                elif self.utilization_error < 0:
+                    log.warn(
+                        error_message + "Continuing because we are scaling down and we don't want the new "
```
mattmb: Just nitpicking on the error message but maybe:
"Continuing because we are scaling down and so we don't need any unregistered instances"

stug: Oh, yep that's better, will change.

paasta_tools/autoscaling/autoscaling_cluster_lib.py
stug: @sagar8192 I feel like that's a little out of scope for this work...let's talk offline about whether that's work we want to prioritize or not.

Will add code to emit to signalfx for now.

PR #1807, opened 2018-04-23
adds timeout cli option for spark-run jupyter

a-johnston:
It seems that `MappingKernelManager.cull_idle_timeout` requires the front end and not just the server to be connected to recognize that a framework is still active. The default timeout has resulted in frameworks being lost for example when the underlying file was automatically reloaded or when the machine running the UI is asleep.


a-johnston:
Effectively the same logic is added with https://github.com/Yelp/paasta/pull/1809

PR #1809, opened 2018-04-24
Make spark-run jupyter kernel culling configurable

huadongliu:
Tested with "paasta spark-run -c norcal-devc --max-cores 40 --cmd jupyter".

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -194,6 +194,28 @@ def add_subparser(subparsers):
         default='default',
     )
 
+    jupyter_group = list_parser.add_argument_group(
+        title='Jupyter kernel culling options',
+        description='Idle kernels will be culled by default. Idle '
+        'kernels with connections can be overridden not to be culled.',
+    )
+
+    jupyter_group.add_argument(
+        '--cull-idle-timeout',
+        type=int,
+        default=7200,
+        help='Timeout (in seconds) after which a kernel is considered idle and '
+        'ready to be culled.',
+    )
+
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -455,10 +478,8 @@ def get_docker_cmd(args, instance_config, spark_conf_str):
 
     # Default cli options to start the jupyter notebook server.
     if original_docker_cmd == 'jupyter':
-        # Shutdown idle kernels that are not connected after one hour.
-        # Shutdown connected kernels if they use more than 32 cores.
-        cull_opts = '--MappingKernelManager.cull_idle_timeout=3600 '
-        if args.max_cores > 32:
+        cull_opts = '--MappingKernelManager.cull_idle_timeout=%s ' % args.cull_idle_timeout
+        if args.cull_connected_max_cores > 0 and args.max_cores > args.cull_connected_max_cores:
```
stug: It feels a little weird to me that cull_connected_max_cores is configurable by the user -- in what circumstance would we expect to see someone run with, say, `--max-cores=100 --cull-connected-max-cores=50`?  Maybe we should just expose a `--cull-connected` option, and if it's not specified we only turn it on if `max-cores` is over 32?

a-johnston: That was my first impression too. Potentially useful if the user is using some default configuration on top of just this CLI. For example an alias or script which accepts a variable `max_cores` but wants a higher default for this cull option.

Related, would an option accepting a profile from a yaml/json config as opposed to recording working cli argument combinations for a given job be useful?

PR #1812, opened 2018-05-01
Deploy scripts for Tron configs

qui:
Plan is to run `deploy_tron_jobs` as a cron job on each Tron master. For each namespace, it will generate the config, and update that namespace if the config is different than the current config. Eventually we'll have a namespace per service.

Here's a sample translation: https://fluffy.yelpcorp.com/i/9W65lKLX4Bkdq65R2F7hTw9299hp1d4s.html

The client was kinda cobbled together - at first, I was going to import the client from Tron, but it was annoying to deal with all of Tron's dependencies (like bsddb). Let me know if you have suggestions on improving it.

Todo:
- [x] tests
- [x] add a tron config to system paasta configs
- [x] better error handling
- [x] docstrings, etc. Are there guidelines on type hints?
- [ ] validation of fields in TronActionConfig, TronJobConfig (maybe later?)


paasta_tools/tron_tools.py
```diff
@@ -0,0 +1,379 @@
...
+        return self.action
+
+    def get_deploy_group(self):
+        return self.config_dict.get('deploy_group', '')
+
+    def get_cmd(self):
+        return self.config_dict.get('command')
+
+    def get_executor(self):
+        executor = self.config_dict.get('executor', 'ssh')
```
Rob-Johnson: this function will break if we ever introduce another executor, right? how about

```
      return 'mesos' if executor == 'paasta' else executor
```

paasta_tools/tron_tools.py
```diff
@@ -0,0 +1,379 @@
...
+        return response.json()
+
+    def _get(self, url, data=None):
+        return self._request('GET', url, data)
+
+    def _post(self, url, data=None):
+        return self._request('POST', url, data)
+
+    def update_namespace(self, namespace, new_config):
+        current_config = self._get('/api/config', {'name': namespace})
```
Rob-Johnson: seems like this validation should be moved up a layer (what if I want to update it in spite of it being the same?).

paasta_tools/list_tron_namespaces.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    args = parse_args()
+    namespaces = tron_tools.get_tron_namespaces_for_cluster(cluster=args.tron_cluster, soa_dir=args.soa_dir)
+    paasta_print('\n'.join(namespaces))
+    sys.exit(0)
```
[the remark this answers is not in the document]
qui: Probably not, just copied it out of list_chronos_jobs. Will take it out!

paasta_tools/tron_tools.py
qui: Yep, that looks better.

qui:
I moved the client out. I tried to do more but couldn't come up with a good organization without circular imports, since I thought some things should still be in a utils module. :/

qui:
I think this is ready for final review, @Rob-Johnson. I might add some itests or type hints if I get a chance, but no big changes.

I'm planning on adding all the check_* methods in TronActionConfig/TronJobConfig later, if that's ok.

paasta_tools/tron_tools.py
```diff
@@ -0,0 +1,371 @@
...
+                self.get_deploy_constraints(
+                    blacklist=self.get_deploy_blacklist(),
+                    whitelist=self.get_deploy_whitelist(),
+                    # Don't have configs for the paasta cluster
+                    system_deploy_blacklist=[],
+                    system_deploy_whitelist=None,
+                ),
+            )
+            constraints.extend(self.get_pool_constraints())
+            assert isinstance(constraints, list)
```
Rob-Johnson: whats this for?

paasta_tools/tron_tools.py
```diff
@@ -0,0 +1,371 @@
...
+            return constraints
+
+    def format_tron_action_dict(self, cluster_fqdn_format):
```
Rob-Johnson: what would you think of moving this outside of the class, and making it a function of the type ActionConfig -> String -> Dict? It kind of separates the object from concerns of whats done with that object. Same question for the TronJobConfig

paasta_tools/tron_tools.py
```diff
@@ -0,0 +1,371 @@
...
+    def __eq__(self, other):
+        if isinstance(other, type(self)):
+            return self.config_dict == other.config_dict
+        return False
+
+
+def load_tron_service_config(service, tron_cluster, soa_dir=DEFAULT_SOA_DIR):
+    """Load all configured jobs for a service, and any additional config values."""
+    tron_conf_file = os.path.join(
+        os.path.abspath(soa_dir), 'tron', tron_cluster, service + '.yaml',
```
Rob-Johnson: hm, I thought the tron-{cluster}.yaml file would be in each service directory?

If I'm reading this correctly, it looks in /nail/etc/services/tron/prod/myservice.yaml - but I was expecting us to do /nail/etc/services/myservice/tron-prod.yaml ?

qui: You're reading it correctly, it's looking for /nail/etc/services/tron/prod/myservice.yaml right now. It should be changed to the 2nd thing after we migrate the tronfig files to service directories, but I want to be able to use this script before we finish that migration.

paasta_tools/tron_tools.py
qui: dunno, another thing I copied from chronos tools :P

paasta_tools/tron_tools.py
qui: Sure, that makes sense. Done.

paasta_tools/cleanup_tron_namespaces.py
```diff
@@ -0,0 +1,76 @@
...
+    args = parser.parse_args()
+    return args
+
+
+def main():
+    args = parse_args()
+
+    client = tron_tools.get_tron_client()
+    namespaces = client.list_namespaces()
+    expected_namespaces = tron_tools.get_tron_namespaces_for_cluster(soa_dir=args.soa_dir)
```
solarkennedy: This script looks risky to me.
Can you make a note to add an option to dry run so we can learn to trust it? Otherwise the first time we run this script it will be very scary.

paasta_tools/paasta_deploy_tron_jobs
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+list_tron_namespaces | xargs -n 1 -r -P 1 setup_tron_namespace
```
solarkennedy: Can you use the ` paasta_`  versions here?

PR #1817, opened 2018-05-02
Use soa_dir in paasta info. Fixes #1813

### LINKED ISSUE #1813: paasta info does not show runbook for target soa_dir
When I run `paasta info -s <service> -d <custom soa dir>`, the 'Runbook' attribute output is not the value in custom soa dir. This is because `paasta info` doesn't set the soa_dir when calling `get_runbook`:
https://github.com/Yelp/paasta/blob/ccfe46b1e38cbfe0b48706f927d49d3a3f515599/paasta_tools/cli/cmds/info.py#L128

The definition for `get_runbook` accepts a `soa_dir` param:
https://github.com/Yelp/paasta/blob/ccfe46b1e38cbfe0b48706f927d49d3a3f515599/paasta_tools/monitoring_tools.py#L55

solarkennedy:


solarkennedy:
Yay tests! Helped me find another situation where we were not using soa_dir.

PR #1819, opened 2018-05-04
Added spark_args option to accommodate arbitrary Spark configuration

huadongliu:


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -36,6 +36,30 @@
...
+    'max-cores': 'spark.cores.max',
+    'executor-cores': 'spark.executor.cores',
+    'executor-memory': 'spark.executor.memory',
+    'driver-max-result-size': 'spark.driver.maxResultSize',
+    'driver-cores': 'spark.driver.cores',
+    'driver-memory': 'spark.driver.memory',
+}
+
+
+class DeprecatedAction(argparse.Action):
```
stug: Neat, I did not know about this

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -36,6 +36,30 @@
 DEFAULT_SPARK_MESOS_SECRET_FILE = '/nail/etc/paasta_spark_secret'
 
 
+deprecated_opts = {
+    '--jars': 'spark.jars',
```
stug: Pretty sure you don't want the `--` prefix on this...and might also want `j` here?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -130,43 +149,51 @@ def add_subparser(subparsers):
         help="Mesos secret (password) to run a framework on Mesos.",
     )
 
+    list_parser.add_argument(
+        '--spark-args',
+        help="Spark configurations documented in https://spark.apache.org/docs/latest/configuration.html.",
+    )
+
+    list_parser.add_argument(
+        '--jars',
```
stug: Probably want `-j`, too, since that was originally supported?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -326,6 +353,54 @@ def get_spark_conf_str(
...
 ):
     spark_conf = list()
+
+    base_args = {
+        'spark.cores.max': '4',
+        'spark.executor.cores': '2',
+        'spark.executor.memory': '4g',
+    }
+
+    if args.spark_args:
```
stug: nitpick, but it might be nice to put this into a method called something like `parse_spark_args`?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -326,6 +353,54 @@ def get_spark_conf_str(
...
+                ),
+            ),
+            file=sys.stderr,
+        )
+        sys.exit(1)
+
+    for opt, val in base_args.items():
+        spark_conf.append('--conf %s=%s' % (opt, val))
+
     spark_conf.append('--conf spark.app.name=%s' % container_name)
```
stug: why not include stuff like this in `base_args`?  Also what happens if the user tries to specify `spark.app.name` themselves?  Will the fact that we're appending this afterwards override their value?

PR #1820, opened 2018-05-04
Add a flag to disable reservation on maint calls

mattmb:
We are trying a new approach with mesos to not send offers from hosts in
maintenance mode. If this works we should not need to reserve the
resources on an agent to prevent tasks getting launched there. This
commit adds a config flag to disable the reservation in both the bounce
logic and the cluster autoscaler. Doing this behind a config flag to
just enable in the places where we have patched mesos first for testing.
If everything works out okay I can double back and remove the
reservation code.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -130,6 +130,7 @@ class ClusterAutoscaler(object):
         log_level: str=None,
         draining_enabled: bool=True,
         enable_metrics: bool=False,
+        enable_maintenance_reservation: bool=True,
     ) -> None:
         self.resource = resource
         self.pool_settings = pool_settings
@@ -143,6 +144,7 @@ class ClusterAutoscaler(object):
         self.instances: List[Dict] = []
         self.sfr: Optional[Dict[str, Any]] = None
         self.enable_metrics = enable_metrics
+        self.enable_maintenance_reservation = enable_maintenance_reservation
 
         self.setup_metrics()
 
@@ -532,7 +534,12 @@ class ClusterAutoscaler(object):
         if should_drain:
             try:
                 drain_host_string = "{}|{}".format(slave_to_kill.hostname, slave_to_kill.ip)
-                drain([drain_host_string], start, duration)
+                drain(
+                    hostnames=[drain_host_string],
+                    start=start,
+                    duration=duration,
+                    reserve_resources=self.enable_maintenance_reservation,
+                )
             except HTTPError as e:
                 self.log.error("Failed to start drain "
                                "on {}: {}\n Trying next host".format(slave_to_kill.hostname, e))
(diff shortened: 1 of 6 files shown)


paasta_tools/setup_marathon_job.py
```diff
@@ -334,7 +335,8 @@ def log_bounce_action(line: str, level: str='debug') -> None:
         if task in tasks_to_kill:
             hostname = task.host
             try:
-                reserve_all_resources([hostname])
+                if enable_maintenance_reservation:
```
solarkennedy: It looks like the whole purpose of this loop is to reserve. It would be the slightest bit faster if you put the if block before the loop at all

paasta_tools/mesos_maintenance.py
```diff
@@ -614,7 +614,7 @@ def unreserve_all_resources(hostnames):
                 raise HTTPError("Failed unreserving all of the resources on %s (%s). Aborting." % (hostname, slave_id))
 
 
-def drain(hostnames, start, duration):
+def drain(hostnames, start, duration, reserve_resources=True):
     """Schedules a maintenance window for the specified hosts and marks them as draining.
     :param hostnames: a list of hostnames
     :param start: the time to start the maintenance, represented as number of nanoseconds since the epoch
```
solarkennedy: Doesn't an if statement need to be here on line 625?

paasta_tools/mesos_maintenance.py
```diff
@@ -635,7 +635,7 @@ def drain(hostnames, start, duration):
     return drain_output
 
 
-def undrain(hostnames):
+def undrain(hostnames, unreserve_resources=True):
     """Unschedules the maintenance window for the specified hosts and unmarks them as draining. They are ready for
     regular use.
     :param hostnames: a list of hostnames
```
solarkennedy: Also here on line 644?

paasta_tools/mesos_maintenance.py
mattmb: Oh duh 🤦‍♂️  thanks for spotting!

paasta_tools/setup_marathon_job.py
mattmb: 👍

paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -532,7 +534,7 @@ def should_drain(self, slave_to_kill: 'PaastaAwsSlave') -> bool:
         if should_drain:
             try:
                 drain_host_string = "{}|{}".format(slave_to_kill.hostname, slave_to_kill.ip)
-                drain([drain_host_string], start, duration)
+                drain([drain_host_string], start, duration, reserve_resources=self.enable_maintenance_reservation)
```
Rob-Johnson: can you convert all these to kwargs please

paasta_tools/mesos_maintenance.py
```diff
@@ -635,17 +636,18 @@ def drain(hostnames, start, duration):
     return drain_output
 
 
-def undrain(hostnames):
```
Rob-Johnson: can you add a docstring?

paasta_tools/mesos_maintenance.py
```diff
@@ -614,18 +614,19 @@ def unreserve_all_resources(hostnames):
                 raise HTTPError("Failed unreserving all of the resources on %s (%s). Aborting." % (hostname, slave_id))
 
 
-def drain(hostnames, start, duration):
```
Rob-Johnson: can you add a docstring for the new param?

mattmb:
👍 on all those comments @Rob-Johnson I've fixup'd and travis is green so I'm going to rebase and merge

PR #1826, opened 2018-05-09
default to using the system paasta config for the remote-run cluster

Rob-Johnson:
@ealter I've raised this as an alternative to your PR - lmk what you think


ealter:
Looks great @Rob-Johnson

PR #1828, opened 2018-05-09
don't set utilization_error to -1 for cancelled_running SFR unless we…

stug:
…'re scaling down

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -146,6 +146,7 @@ class ClusterAutoscaler(object):
         self.enable_metrics = enable_metrics
         self.enable_maintenance_reservation = enable_maintenance_reservation
 
+        self.log.info('Initialized with utilization error %s' % self.utilization_error)
         self.setup_metrics()
 
     @property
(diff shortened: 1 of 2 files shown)


solarkennedy:
Also @mattmb you might be a good person to review this.

mattmb:
Hmm, does this really change what happens or are you just trying to make it clearer?

Before when we didn't return `0, 0` from the metrics provider presumably we try to scale up the cancelled_running SFR and skip it later?

huadongliu:
My read is that the only change in this PR is to check utilization_error < 0 in cancelled_running before scaling down. Rest of the PR is make the logic a bit clearer. Shiptit.

stug:
@mattmb this does change the behavior.  After https://github.com/Yelp/paasta/pull/1805, we'd never have `self.utilization_error > 0` for a `cancelled_running` SFR because `SpotAutoscaler.__init__()` sets `self.utilization_error = -1` if the SFR is in `cancelled_running`.  So that check in `metrics_provider()` where we return `0, 0` will never be triggered.  Moving the check into `__init__()` restores the behavior that existed prior to https://github.com/Yelp/paasta/pull/1805 (don't try to scale the cancelled_running SFR down unless we were scaling down anyway).

Sorry if that's confusing -- lemme know if you want to chat real fast in person.

PR #1832, opened 2018-05-11
Ping Slack with authors

solarkennedy:
Based on #1829, this connects the two and makes it so we will ping authors when their deploy is ready.

@chriskuehl do you think this approach is "better" (comparing before/after sha) or do I need to go all the way and do exactly what the existing thing does, which is always compare against the "production_deploy_group" ?

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -205,14 +205,37 @@ def report_waiting_aborted(service, deploy_group):
 
 
 class SlackDeployNotifier(object):
-    def __init__(self, service, deploy_info, deploy_group, commit, old_commit):
+    def __init__(self, service, deploy_info, deploy_group, commit, old_commit, git_url):
         self.sc = get_slack_client()
         self.service = service
         self.deploy_info = deploy_info
         self.deploy_group = deploy_group
         self.channels = deploy_info.get('slack_channels', [])
         self.commit = commit
-        self.old_commit = old_commit
+        self.old_commit = self.lookup_production_deploy_group_sha() or old_commit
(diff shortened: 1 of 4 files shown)


chriskuehl:
@solarkennedy I think this approach (comparing the two SHAs rather than against the `production_deploy_group`) might not work so well in some cases.

For example, imagine nobody has run a push through stageg in 3 days (they've all used stagef) so it's significantly behind master. Now somebody deploys through it and dozens of people get pinged who aren't involved in the current push at all. A similar thing can happen if e.g. somebody tries out a test branch on stage.

solarkennedy:
Hmmm.

How about I do `production_deploy_group` if available otherwise compare against the old commit?

chriskuehl:
Sounds good to me. In practice I bet this is really only a problem for yelp-main anyway.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -223,19 +246,22 @@ def deploy_group_is_set_to_notify(self):
...
             )
-            self.post(channels=self.channels, message=message)
+            self.post(channels=self.channels, message=mes)
             if self.old_commit is not None and self.commit != self.old_commit:
                 mes = (
                     "Roll it back at any time with:\n"
-                    f"```paasta rollback --service {self.service} --deploy-group {self.deploy_group} "
-                    f"--commit {self.old_commit}```"
+                    f"`paasta rollback --service {self.service} --deploy-group {self.deploy_group} "
+                    f"--commit {self.old_commit}`"
```
chriskuehl: wonder if this message will be confusing for y-m where that's _not_ the right way to initiate a rollback

PR #1836, opened 2018-05-15
read tron config files from service directories

chlgit:
I added tron_migration_status/tron-devc.yaml, and it's able to create namespace 'tron_migration_status' at tron-devc.


chlgit:
I was planning to keep existing deployment for Tron dir, and to use this script for service directories. But now the script can handle both.

paasta_tools/tron_tools.py
```diff
@@ -327,15 +327,31 @@ def format_tron_job_dict(job_config, cluster_fqdn_format, default_paasta_cluster
...
+    )
+    _update_config(config, tron_conf_path1)
+
+    tron_conf_path2 = os.path.join(
         os.path.abspath(soa_dir), 'tron', tron_cluster, service + '.yaml',
     )
-    config = service_configuration_lib._read_yaml_file(tron_conf_file)
-    if not config:
-        raise NoConfigurationForServiceError('No configuration found for Tron at %s' % tron_conf_file)
+    _update_config(config, tron_conf_path2)
```
qui: Does this mean we might end up with configuration for the same namespace in two different places? I don't think that we should let that happen. We should just check the two files in order, and return the first config that exists.

paasta_tools/tron_tools.py
```diff
@@ -327,15 +327,31 @@ def format_tron_job_dict(job_config, cluster_fqdn_format, default_paasta_cluster
     # Only pass non-None values, so Tron will use defaults for others
     return {key: val for key, val in result.items() if val is not None}
 
+def _update_config(config, tron_conf_path):
+    tron_config_file = service_configuration_lib._read_yaml_file(tron_conf_path)
```
qui: I used `_read_yaml_file` as a hack because the filepath was different. But for the service directory, I think we should use `read_extra_service_information` like Chronos/Marathon do: https://github.com/Yelp/paasta/blob/master/paasta_tools/chronos_tools.py#L198

paasta_tools/tron_tools.py
```diff
@@ -383,3 +400,13 @@ def get_tron_namespaces_for_cluster(cluster=None, soa_dir=DEFAULT_SOA_DIR):
...
     return namespaces
+
+def get_tron_namespaces_for_cluster(cluster=None, soa_dir=DEFAULT_SOA_DIR):
+    """Get all the namespaces that are configured in a particular Tron cluster."""
+    if not cluster:
+        cluster = load_tron_config().get_cluster_name()
+   
+    namespaces1 = _get_tron_namespaces_from_service_dir(cluster, soa_dir)
+    namespaces2 = _get_tron_namespaces_from_tron_dir(cluster, soa_dir)
+    namespaces= list(set(namespaces1) - set(namespaces2)) + namespaces2
```
qui: why not set(namespaces1 + namespaces2)?

paasta_tools/tron_tools.py
chlgit: If so, it will require we migrate the entire namespace against each cluster once. I presume it's the common case, but it won't hurt to keep some flexibility. Unless you have a strong opinion on that?

paasta_tools/tron_tools.py
chlgit: I don't know what I was thinking then :(

paasta_tools/tron_tools.py
chlgit: Good call. qui++

paasta_tools/tron_tools.py
qui: I think it will be confusing to have 2 different locations where you can make changes to your tronfig for the same namespace.

For big namespaces like MASTER, the service name (yelp-main, etc) won't match the old namespace (MASTER) so you can still move things out of MASTER incrementally.

tests/test_tron_tools.py
```diff
@@ -429,19 +460,25 @@ def test_load_tron_service_config(self, mock_job_config, mock_read_file):
             mock.call(job_1, soa_dir),
             mock.call(job_2, soa_dir),
         ]
-        expected_filename = '/other/services/tron/dev/foo.yaml'
-        mock_read_file.assert_called_once_with(expected_filename)
+        assert mock_read_service_info.call_count == 1
+        assert mock_read_file.call_count == 1
+        mock_read_service_info.assert_has_calls([mock.call('foo', 'tron-dev', '/other/services')])
+        mock_read_file.assert_has_calls([])
```
qui: Why is _read_yaml_file called with no arguments?

PR #1838, opened 2018-05-17
Print out a jenkins url for mark-for-deployments jobs

solarkennedy:


qui:
I'm not sure that $JENKINS_URL is the right env variable, docs say it is: "Full URL of Jenkins, like http://server:port/jenkins/". Maybe we want BUILD_URL instead?

It would be nice if we tested this on Jenkins servers first.

solarkennedy:
Thanks @qui, you are totally right.
It looks like I already have this in a different part of the code with BUILD_URL on failure notifications? I'm just going to close for now.

PR #1839, opened 2018-05-18
Make SpotAutoscaler cleanup cancelled_running SFR with no instances

stug:
More or less treating a cancelled_running SFR with no instances the same way we treat a cancelled SFR.  Let me know if you think that's too aggressive.

Also we've seen SFRs sit in this state more or less indefinitely (until we manually cancel them), and this may make us less likely to notice that.  Is that at all problematic?

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -814,7 +814,15 @@ class SpotAutoscaler(ClusterAutoscaler):
 
     def metrics_provider(self, mesos_state: MesosState) -> Tuple[float, int]:
         if not self.sfr or self.sfr['SpotFleetRequestState'] == 'cancelled':
-            self.log.error("SFR not found, removing config file.".format(self.resource['id']))
+            self.log.error("SFR not found or cancelled, removing config file.")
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -814,7 +814,15 @@ def get_spot_fleet_instances(
 
     def metrics_provider(self, mesos_state: MesosState) -> Tuple[float, int]:
         if not self.sfr or self.sfr['SpotFleetRequestState'] == 'cancelled':
-            self.log.error("SFR not found, removing config file.".format(self.resource['id']))
+            self.log.error("SFR not found or cancelled, removing config file.")
+            self.cleanup_cancelled_config(self.resource['id'], self.config_folder, dry_run=self.dry_run)
+            return 0, 0
+        if self.sfr['SpotFleetRequestState'] == 'cancelled_running' and len(self.instances) == 0:
```
solarkennedy: can you make this an elif instead? In practice it doesn't matter in this case, but I don't think you ever want both of these if blocks to ever accidentally execute

stug: sure, changed.

PR #1840, opened 2018-05-18
Don't send slack noifications if the old and new commits are the same

solarkennedy:
Since this situation is idempotent, and we do it every day with daily builds, I think we want to suppress these notifications.

@chriskuehl


qui:
Does this mean that the first deployment (old_commit is None) won't have messages?
Anyway, merging this.

PR #1843, opened 2018-05-25
[tron tools] error on namespace conflict

keymone:
raise an error if conflicting namespaces are detected during in `get_tron_namespaces_for_cluster`

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -406,11 +406,11 @@ def get_tron_namespaces_for_cluster(cluster=None, soa_dir=DEFAULT_SOA_DIR):
     namespaces1 = set(_get_tron_namespaces_from_service_dir(cluster, soa_dir))
     namespaces2 = set(_get_tron_namespaces_from_tron_dir(cluster, soa_dir))
 
-    if namespaces1 & namespaces2:
+    if namespaces1.intersection(namespaces2):
         raise ConflictingNamespacesError(
             "namespaces found in both service/*/tron and service/tron/*: {}".
-            format(namespaces1 & namespaces2)
+            format(namespaces1.intersection(namespaces2))
         )
 
-    namespaces = list(namespaces1 + namespaces2)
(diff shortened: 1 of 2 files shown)


paasta_tools/tron_tools.py
```diff
@@ -395,13 +395,22 @@ def _get_tron_namespaces_from_tron_dir(cluster, soa_dir):
...
     if not cluster:
         cluster = load_tron_config().get_cluster_name()
 
-    namespaces1 = _get_tron_namespaces_from_service_dir(cluster, soa_dir)
-    namespaces2 = _get_tron_namespaces_from_tron_dir(cluster, soa_dir)
-    namespaces = list(set(namespaces1 + namespaces2))
+    namespaces1 = set(_get_tron_namespaces_from_service_dir(cluster, soa_dir))
+    namespaces2 = set(_get_tron_namespaces_from_tron_dir(cluster, soa_dir))
+
+    if namespaces1.intersection(namespaces2):
```
Rob-Johnson: I'm not sure on the context, but I kind of think this error checking logic belongs in the caller?

keymone: result of this function is list of unique namespaces, caller can't check if there were any duplicates.

this function knows that namespaces can come from multiple sources - i think we should raise an error as fast and close to the source as possible.

tests/test_tron_tools.py
```diff
@@ -559,3 +559,22 @@ def test_get_tron_namespaces_for_cluster_default(self, mock_system_tron_config,
...
+        mock_ls.return_value = ['cool.yaml']
+        soa_dir = '/my_soa_dir'
+
+        try:
+            tron_tools.get_tron_namespaces_for_cluster(
+                cluster=cluster_name,
+                soa_dir=soa_dir,
+            )
+            assert False, "no exception raised on conflicting namespaces"
+        except tron_tools.ConflictingNamespacesError as e:
```
qui: `pytest.raises` can do this assertion for you. Should fix the test coverage issue in pre-commit.

PR #1846, opened 2018-05-29
[tron tools] update multiple namespaces in setup_tron_namespace

keymone:
- allow accepting multiple services to update in cli args
- add optional `--all` argument to update all namespaces

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,4 +1,3 @@
-repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     sha: v1.3.0
     hooks:
--- a/paasta_tools/setup_tron_namespace.py
+++ b/paasta_tools/setup_tron_namespace.py
@@ -65,13 +65,14 @@ def parse_args():
 
 def main():
     args = parse_args()
-
-    if args.verbose:
-        logging.basicConfig(level=logging.DEBUG)
-    else:
-        logging.basicConfig(level=logging.WARNING)
+    log_level = logging.DEBUG if args.verbose else logging.WARNING
+    logging.basicConfig(level=log_level)
 
     if args.all_namespaces:
+        if args.services:
+            log.error('Do not pass service names with --all flag')
+            sys.exit(1)
+
         try:
             services = tron_tools.get_tron_namespaces_for_cluster()
         except Exception as e:
--- a/paasta_tools/setup_tron_namespace.py
+++ b/paasta_tools/setup_tron_namespace.py
@@ -65,7 +65,7 @@ def parse_args():
 
 def main():
     args = parse_args()
-    log_level = logging.DEBUG if args.verbose else logging.WARNING
+    log_level = logging.DEBUG if args.verbose else logging.INFO
     logging.basicConfig(level=log_level)
 
     if args.all_namespaces:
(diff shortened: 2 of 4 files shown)


paasta_tools/setup_tron_namespace.py
```diff
@@ -62,22 +75,49 @@ def main():
...
-    try:
-        new_config = tron_tools.create_complete_config(
-            service=args.service,
-            soa_dir=args.soa_dir,
-        )
+        if args.all_namepsaces:
+            tron_config = tron_tools.load_tron_config()
+            try:
+                services = tron_tools.get_tron_namespaces_for_cluster(
+                    cluster=tron_config.get_tron_cluster())
```
qui: `get_tron_namespaces_for_cluster` already defaults to the cluster in the tron config, so you don't have to load the tron config.

paasta_tools/setup_tron_namespace.py
```diff
@@ -62,22 +75,49 @@ def main():
...
+            except ConflictingNamespacesError as e:
+                log.error('Failed to list tron namespaces: {error}'.format(
+                    error=str(e)))
+                sys.exit(1)
+        else:
+            services = [args.service]
```
qui: Isn't `args.service` already a list because of `nargs`? Otherwise, what happens if multiple services are passed in?

paasta_tools/setup_tron_namespace.py
```diff
@@ -62,22 +75,49 @@ def main():
...
+        for service in services:
+            try:
+                new_config = tron_tools.create_complete_config(
+                    service=args.service,
+                    soa_dir=args.soa_dir,
+                )
+                client.update_namespace(args.service, new_config)
+                updated.append(service)
+            except (
+                    InvalidTronConfig,
```
chlgit: Are they all the errors that might happen here? If not, maybe we can catch a general error here. I can't come up with the scenario that we should raise the exception rather than continue to work on other services.

keymone: i had the same question. in the end i just moved that catch block as is. if we swallow all exceptions here we might miss a real issue that needs to be addressed. i'll leave it for now.

paasta_tools/setup_tron_namespace.py
```diff
@@ -62,22 +71,48 @@ def main():
...
         logging.basicConfig(level=logging.WARNING)
 
-    try:
-        new_config = tron_tools.create_complete_config(
-            service=args.service,
-            soa_dir=args.soa_dir,
-        )
+        if args.all_namepsaces:
+            try:
+                services = tron_tools.get_tron_namespaces_for_cluster()
```
Rob-Johnson: do we need mutual exclusion between args.services and args.all_namespaces? or a log line to tell the user whats happening?

.pre-commit-config.yaml
```diff
@@ -1,4 +1,3 @@
-repos:
```
asottile: why did you revert this to the legacy format?

requirements-dev.txt
```diff
@@ -20,7 +19,7 @@ path.py==8.1
 pbr==3.1.1
 pep8==1.5.7
 pluggy==0.6.0
-pre-commit==1.10.1
+pre-commit==0.18.1
```
asottile: this version is really really old -- you've gone back in time to 2017

PR #1853, opened 2018-06-06
Added safer passing of credentials to spark

### LINKED ISSUE #1851: Spark-run leaks AWS credentials into process title
The spark run command will pass AWS keys via ``--env AWS_SECRET_ACCESS_KEY=....``. This becomes visible to other users on the machine via ps.

It should be possible to set this env variable in the paasta process, pass it via ``--env AWS_SECRET_ACCESS_KEY`` without the value, and then unset it in the paasta process.

Qmando:
Pass aws credentials via the environment rather than as ``--env key=value``. This prevents you from being able to see the keys using ``ps``

Fixes #1851 

To verify this code works:
```
>>> os.execlpe('docker', *['docker', 'run', '-it', '--env', 'foo', 'ubuntu', {'foo': 'bar'}])
root@dd9e6ceafaee:/# echo $foo
bar
```


huadongliu:
Shipit!

PR #1854, opened 2018-06-06
Just assign to self instead of side-stepping through __dict__

asottile:


asottile:
was testing out a new linter and couldn't help myself when I saw this :)

hope you all are doing well!

paasta_tools/frameworks/task_store.py
```diff
@@ -40,12 +40,12 @@ def __init__(
         offer=None,
```
asottile: A note here that `mypy` would actually complain about this `__init__` if it were being typechecked -- the annotations for `mesos_task_state`, `is_draining`, and `is_healthy` are non-nullable, but these parameters are `Optional`.  fortunately (?) mypy doesn't typecheck functions without annotations so this isn't _currently_ failing (but it might be a bug / surprise in the future!)

asottile:
oh nevermind, I see the `__setattr__` / `__delattr__` dance -- tricky tricky

PR #1855, opened 2018-06-06
pyupgrade learns how to f-string

asottile:


asottile:
rebased again after conflicts -- would love a review! CC @EvanKrall @chriskuehl

paasta_itests/steps/setup_chronos_job_steps.py
```diff
@@ -57,7 +57,7 @@ def old_jobs_leftover(context, job_count, job_name):
     for i in range(job_count):
         job_definition = copy.deepcopy(context.jobs[job_name])
         # modify the name by replacing the last character in the config hash
-        modified_name = "%s%s" % (job_definition['name'][:-1], i)
+        modified_name = "{}{}".format(job_definition['name'][:-1], i)
```
EvanKrall: I'm a big fan of f-strings but much less of a fan of str.format -- printf-style formatting is a well-entrenched standard, and the new {} style is longer: in particular, `%d` becomes `{:d}`, which encourages lazy people to just use `{}` for ints, which is less explicit.

Is there an option for disabling this kind of conversion?

paasta_tools/adhoc_tools.py
```diff
@@ -95,7 +95,7 @@ def __init__(
         )
 
     def __repr__(self) -> str:
-        return "AdhocJobConfig(%r, %r, %r, %r, %r, %r)" % (
+        return "AdhocJobConfig({!r}, {!r}, {!r}, {!r}, {!r}, {!r})".format(
```
EvanKrall: why is this one using .format instead of an f string?

paasta_itests/steps/setup_chronos_job_steps.py
asottile: % formatting is the "old style" formatting, has a bunch of edge cases, and in general is less friendly to work with.  You can read more about it [here](https://pyformat.info/).  In fact, % formatting was almost removed from python3 and was only kept for legacy reasons.

% formatting is filled with quite a few [spiders](https://twitter.com/codewithanthony/status/970182992885817344) and has quite a few ways to shoot yourself in the foot (static analysis of whether `foo` is a tuple or not for instance: `'%s' % foo`)

> which encourages lazy people to just use `{}` for ints, which is less explicit.

You should just use `{}` for ints.  Anything else is an unnecessary conversion -- in fact `{:d}` isn't even correct either (the `:d` only adds an `isinstance()` check and then it goes straight to `str(...)`).

Note that many forms of `%` formatting also _don't have an equivalent f-string_ and won't be changed.

There is no current option to disable going from % formatting to .format(), you really should let go of the past :D

paasta_tools/adhoc_tools.py
asottile: see the note here: https://github.com/asottile/pyupgrade#f-strings

PR #1856, opened 2018-06-07
[tron tools] correctly skip deployment of unchanged tron namespaces

keymone:


paasta_tools/tron/client.py
```diff
@@ -88,9 +89,11 @@ def update_namespace(self, namespace, new_config, skip_if_unchanged=True):
                 'no_header': 1,
             },
         )
-        if skip_if_unchanged and new_config == current_config['config']:
-            log.info('No change in config, skipping update.')
-            return
+
+        if skip_if_unchanged:
+            if yaml.load(new_config) == yaml.load(current_config['config']):
```
qui: This is because of ordering, right? Like, the Tron configs are currently in non-alphabetical order, whereas this script sends them in alphabetical order.

I am pretty sure that after we run this script once, the configs will remain alphabetical, so it's a one time problem. I don't mind this change if you think it's better this way, though.

keymone: pretty much, but also stuff that comes back from tron has yaml comments which also screws up the comparison. eventually we should change it on tron's side to just send us the result of yaml.dump, but for now this is our simplest option.

PR #1857, opened 2018-06-11
Bump boto requirements to fix trusty build

keymone:


requirements.txt
```diff
@@ -1,3 +1,4 @@
+--extra-index-url=https://pypi.yelpcorp.com/simple
```
sjaensch: Why do we need this here? Isn't it handled in the Makefile?

requirements.txt
```diff
@@ -7,8 +8,8 @@ arrow==0.12.1
 async-timeout==3.0.0
 binaryornot==0.4.4
 boto==2.48.0
-boto3==1.3.0
-botocore==1.4.93
+boto3>=1.3.0
+botocore>=1.4.93
```
sjaensch: Typically, a requirements.txt is used for reproducible builds; you should pin every package to an exact version. setup.py (or requirements-minimal.txt - not sure why paasta has both) has loose minimum requirements.

I think paasta needs to decide if it's an application or a library, and then modify things accordingly. It might also be a bit of both, in which case things might be fine as-is (or it needs two separate requirements specifications). See https://github.com/Yelp/requirements-tools/blob/master/README.md for further information.

mattmb: It is kinda both :-( 

Generally we treat it as an application and discourage people from importing it. But there's at least 1 place I know we do end up importing it.

requirements.txt
```diff
@@ -87,13 +91,18 @@ tzlocal==1.2
...
 webcolors==1.7
 WebOb==1.7.4
 websocket-client==0.44.0
 whichcraft==0.4.0
 wsgicors==0.7.0
 yarl==1.1.1
+yelp-cgeom==1.3.1
 yelp-clog==2.12.1
+yelp-logging==1.0.37
+yelp_meteorite
```
mattmb: This isn't opensource right, so won't this break on travis? We have another requirements file for the yelpy stuff

requirements.txt
mattmb: @keymone I think @sjaensch is right in general though. Can we not just pin to a higher version or does that break another build?

requirements.txt
```diff
@@ -49,6 +50,7 @@ ply==3.4
 poyo==0.4.0
 progressbar2==3.10.1
 py==1.5.2
+pyinotify
```
sjaensch: Please pin this one as well.

requirements.txt
sjaensch: Yep, looks like this is handled by https://github.com/Yelp/paasta/blob/master/yelp_package/extra_requirements_yelp.txt, which already has these requirements.

keymone:
@sjaensch shippable now?

PR #1860, opened 2018-06-12
[tron] Fix fetching tron configs when jobs is None

keymone:


paasta_tools/tron_tools.py
```diff
@@ -346,9 +346,7 @@ def load_tron_service_config(service, tron_cluster, soa_dir=DEFAULT_SOA_DIR):
         raise NoConfigurationForServiceError('No Tron configuration found for service %s' % service)
 
     extra_config = {key: value for key, value in config.items() if key != 'jobs'}
-    job_configs = []
-    for job in config.get('jobs', []):
-        job_configs.append(TronJobConfig(job, soa_dir))
+    job_configs = [TronJobConfig(job, soa_dir) for job in config.get('jobs') or []]
```
Rob-Johnson: ``[TronJobConfig(job, soa_dir) for job in config.get('jobs', [])]``

keymone: that's exactly what i'm changing in this PR, it will return None is that's the value behind `jobs` key:

    {}.get('jobs', [])             # => []
    {'jobs': None}.get('jobs', []) # => None - bad

PR #1861, opened 2018-06-12
Update constraint format for Tron

qui:
I changed this when I added Mesos runs to Tron, forgot to update it here.

Tested by running setup_tron_namespace from yelpsoa-configs against a local trond and it works! Task runs on mesosstage.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -289,7 +289,13 @@ def format_tron_action_dict(action_config, cluster_fqdn_format):
                 'value': param['value'],
             } for param in action_config.format_docker_parameters()
         ]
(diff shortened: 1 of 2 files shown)


paasta_tools/tron_tools.py
```diff
@@ -289,7 +289,13 @@ def format_tron_action_dict(action_config, cluster_fqdn_format):
                 'value': param['value'],
             } for param in action_config.format_docker_parameters()
         ]
-        result['constraints'] = action_config.get_calculated_constraints()
+        result['constraints'] = [
+            {
+                'attribute': constraint[0],
```
keymone: could just be `dict(zip(['attribute', 'operator', 'value'], constraint))`

PR #1863, opened 2018-06-14
Fix more async problems

EvanKrall:
- Switch to gunicorn for the API, so that we don't have to use gevent's monkey-patching to get concurrency, since that breaks asyncio.
- Fix some call sites I missed when changing the mesos module to async

### DIFF
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -18,11 +18,10 @@ Responds to paasta service and instance requests.
 import argparse
 import logging
 import os
+import sys
 
 import requests_cache
 import service_configuration_lib
-from gevent import monkey
-from gevent.wsgi import WSGIServer
 from pyramid.config import Configurator
 from wsgicors import CORS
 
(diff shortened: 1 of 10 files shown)


paasta_tools/api/api.py
```diff
@@ -110,27 +124,21 @@ def setup_paasta_api():
...
+        os.environ["PAASTA_API_DEBUG"] = "1"
 
     if args.soa_dir:
-        settings.soa_dir = args.soa_dir
-
-    server = WSGIServer(('', int(args.port)), make_app())
-    log.info("paasta-api started on port %d with soa_dir %s" % (args.port, settings.soa_dir))
-
-    try:
-        # We create the Zookeeper pool here to prevent the context manager
```
mattmb: How confident are you that we don't need to do this somewhere else? I guess each worker will instantiate its own ZookeeperPool object but can each worker serve multiple requests in parallel?

EvanKrall: Fairly confident. We're running gunicorn with multiple workers but each worker only has one thread, and isn't using gevent or asyncio or anything like that to be able to do more than one request at a time.

PR #1866, opened 2018-06-19
ensure cmd is not None before interpolating

Rob-Johnson:
fixes an issue where we'd attempt to interpolate the command with
contextvaribales even if the command is None:

### DIFF
--- a/paasta_tools/chronos_rerun.py
+++ b/paasta_tools/chronos_rerun.py
@@ -63,7 +63,7 @@ def parse_args():
     return args
 
 
-def modify_command_for_date(chronos_job, date):
+def modify_command_for_date(chronos_job, date, verbose):
     """
     Given a chronos job config, return a cloned job config where the command
     has been modified to reflect what it would have run as on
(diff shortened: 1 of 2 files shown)


paasta_tools/chronos_rerun.py
```diff
@@ -255,6 +256,7 @@ def main():
             soa_dir=args.soa_dir,
         )
         original_command = chronos_job_config.get_cmd()
+        print(original_command)
```
macisamuele: Is this print needed?

paasta_tools/chronos_rerun.py
```diff
@@ -76,7 +76,8 @@ def modify_command_for_date(chronos_job, date):
         interpolate in the context of the date provided.
     """
     current_command = chronos_job['command']
-    chronos_job['command'] = chronos_tools.parse_time_variables(current_command, date)
+    if current_command is not None:
```
macisamuele: I would expect that all the jobs have a command defined.
Checking the docs seems that `cmd`  is optional as the docker container could have an entry point defined.
Could we print something if command is None? I see that `paasta rerun` as `--verbose` option, we might write it there

keymone: i agree we should warn user that something went wrong. personally i'd prefer to just fail here - how can we allow things to continue without command? something has gone terribly wrong.

Rob-Johnson: not necessarily @keymone  - command is optional (it falls back to using the default in the docker image).

sure, I can log an extra line if you think its helpful

PR #1871, opened 2018-06-22
Added spark-run command to start a spark history server

huadongliu:
Tests:
paasta spark-run --spark-args 'spark.eventLog.enabled=true spark.eventLog.dir=s3a://{BUCKET}' --cmd 'spark-submit /work/examples/s3.py'
paasta spark-run --spark-args 'spark.history.fs.logDirectory=s3a://{BUCKET}' --cmd history-server

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -128,7 +128,7 @@ def add_subparser(subparsers):
 
     list_parser.add_argument(
         '-C', '--cmd',
-        help="Run Spark with the spark-shell, pyspark, spark-submit or jupyter command.",
+        help="Run the spark-shell, pyspark, spark-submit, jupyter, or history-server command.",
         default='pyspark',
     )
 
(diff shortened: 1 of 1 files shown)


solarkennedy:
How about just `history-server` ? I don't think we need the .sh or the "start" and it makes it feel like the other sub-commands.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -539,10 +555,14 @@ def configure_and_run_docker_container(
         get_spark_env(
             args,
             spark_conf_str,
+            spark_ui_port,
         ),
     )
 
-    paasta_print('\nSpark Monitoring URL http://%s:%d\n' % (socket.getfqdn(), spark_ui_port))
+    if 'start-history-server.sh' in docker_cmd:
```
solarkennedy: Can you be more strict? Can you just say `if docker_cmd == history-server` ?

huadongliu:
Previously the thought was to match spark on paasta commands as much as the vanilla spark commands. So that what folks learn on spark can be applied to spp, and vice versa. Using history-server as a sub-command works too, since we added a jupyter sub-command anyway.

PR #1872, opened 2018-06-25
Add support for kubernetes services

mattmb:
Naturally this is a WIP and there is a lot to do.

This aims to:
* Create a k8s deployment dict with minimal features from a kubernetes-<cluster>.yaml
* Have a cron job that can assert the deployment is up to date and update it if not
* Add `get_kubernetes_services_running_here_for_nerve` so that nerve can list and register k8s services via their pod IP
* Add enough settings to the k8s deployment to closely mimic a crossover bounce

I'm going to write some unit tests next before this ships. itests we can do separately on another branch. That said, reviews/suggestions welcome at this point.


paasta_tools/kubernetes_tools.py
```diff
@@ -0,0 +1,583 @@
...
+    def get_sanitised_volume_name(self, volume_name) -> str:
+        """I know but we really aren't allowed many characters..."""
+        volume_name = volume_name.rstrip('/')
+        sanitised = volume_name.replace('/', 'slash-')
+        return sanitised.replace('_', '--')
+
+    def get_sidecar_containers(self, system_paasta_config) -> List:
+        registrations = " ".join(self.get_registrations())
+        hacheck_sidecar = {
+            "image": "docker-paasta.yelpcorp.com:443/mmb-hacheck-test",
```
mattmb: todo: make this a config option

mattmb:
Reviewers should definitely check that ad1bf19c04470999eaf5f7fb5e5080d7d93be38b is safe. It's the only thing I think I've added that affects the current implementation

mattmb:
Thanks, I think I generally agree with your list. The sidecar is a low effort way to get started but I'm not really sure it's the best longterm plan. I think if we like the k8s deployment primitives we will either need a sidecar or a binary that we can run from the service containers that talks to hacheck on the host (again assuming that we want the same smartstack integration we have at the moment).

`is_pod_healthy_in_smartstack.py` is also necessary if we want to match the existing bounce logic with smartstack integration. For marathon we do something similar we just do it centrally: https://github.com/Yelp/paasta/blob/master/paasta_tools/bounce_lib.py#L220

paasta_tools/kubernetes_tools.py
```diff
@@ -0,0 +1,581 @@
...
+            cluster=self.cluster,
+            config_dict=dict(self.config_dict),
+            branch_dict=dict(self.branch_dict) if self.branch_dict is not None else None,
+            soa_dir=self.soa_dir,
+        )
+
+    def get_bounce_method(self) -> str:
+        """Get the bounce method specified in the service's kubernetes configuration."""
+        # map existing bounce methods to k8s equivalents.
+        kube_deploy_stategy_map = {'crossover': 'RollingUpdate', 'downthenup': 'Recreate'}
```
Rob-Johnson: nit: can you make these constants somewhere please

paasta_tools/kubernetes_tools.py
```diff
@@ -0,0 +1,581 @@
...
+        return kube_deploy_stategy_map[self.config_dict.get('bounce_method', 'crossover')]
```
Rob-Johnson: worth making this ``kube_deploy_strategy_map.get(...)`` instead, incase a bad value creeps into the config dict?

paasta_tools/kubernetes_tools.py
```diff
@@ -0,0 +1,581 @@
...
+            "env": self.get_kubernetes_environment(),
+            "ports": [
+                {
+                    "containerPort": 6666,
+                },
+            ],
+            "volumeMounts": [
+                {
+                    "mountPath": "/var/spool/hacheck",
+                    "name": self.get_sanitised_volume_name("/var/spool/hacheck"),
```
Rob-Johnson: why'd you need to mount this? because hacheck doesn't create it itself?

paasta_tools/kubernetes_tools.py
```diff
@@ -0,0 +1,581 @@
...
+        soa_dir=soa_dir,
+    )
+    return kubernetes_service_config.get_registrations()
+
+
+def get_kubernetes_services_running_here() -> List[KubeService]:
+    services = []
+    pods = requests.get('http://127.0.0.1:10255/pods').json()
+    for pod in pods['items']:
+        if pod['status']['phase'] != 'Running' or pod['metadata']['namespace'] != 'paasta':
```
Rob-Johnson: I didn't see the namespace being set anywhere in format_kubernetes_app_dict?

paasta_tools/setup_kubernetes_job.py
```diff
@@ -0,0 +1,179 @@
...
+        soa_dir=soa_dir,
+    )
+    sys.exit(1 if num_failed_deployments else 0)
+
+
+def setup_kube_deployments(
+    kube_client: KubeClient,
+    service_instances: List[str],
+    soa_dir: str=DEFAULT_SOA_DIR,
+) -> int:
```
Rob-Johnson: if we're not doing anything with the number here, then can we return a boolean instead?

paasta_tools/kubernetes_tools.py
mattmb: huh, I can't remember why I thought this would be needed, I think you are right that it isn't needed.

paasta_tools/kubernetes_tools.py
mattmb: It seems to be a separate param to the API so we set it in the `create_` call later

paasta_tools/kubernetes_tools.py
mattmb: We talked a bit about this IRL. I might try and do some fancy exception handling but I might also just ship this as is and let the KeyError bubble up if it does happen.

paasta_tools/kubernetes_tools.py
```diff
@@ -204,8 +206,8 @@ def get_bounce_method(self) -> str:
         # map existing bounce methods to k8s equivalents.
         return KUBE_DEPLOY_STATEGY_MAP[self.config_dict.get('bounce_method', 'crossover')]
 
-    def get_deployment_strategy_config(self) -> Dict:
-        strategy_dict = {}
+    def get_deployment_strategy_config(self) -> Dict[str, Any]:
+        strategy_dict: Dict[str, Any] = {}
```
vkhromov: JFYI, you can use `TypedDict` from `mypy_extensions` to annotate those dicts.

PR #1874, opened 2018-06-28
Fix check registered slaves aws

davent:
`check_registered_slaves_aws.py` is producing the following error:
```
  File "/opt/venvs/paasta-tools/bin/check_registered_slaves_aws.py", line 58, in <module>
    main()
  File "/opt/venvs/paasta-tools/bin/check_registered_slaves_aws.py", line 52, in main
    if check_registration(threshold):
  File "/opt/venvs/paasta-tools/bin/check_registered_slaves_aws.py", line 22, in check_registration
    dry_run=True,
  File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/autoscaling/autoscaling_cluster_lib.py", line 959, in __init__
    super(AsgAutoscaler, self).__init__(*args, **kwargs)
TypeError: __init__() missing 3 required positional arguments: 'utilization_error', 'max_increase', and 'max_decrease'
```

This appears to be due to the ClusterAutoscaler class requiring 3 parameters which are not being passed through from check_registered_slaves_aws.py#16
* utilization_error
* max_increase
* max_decrease

I have set these all to `0.0` in the branch to honor the required for a value.

Also, the call to `get_mesos_master().state` on line 11 is to an async function in mesos/master.py#201. Therefore we need to block on this call until it returns.

### DIFF
--- a/paasta_tools/contrib/check_registered_slaves_aws.py
+++ b/paasta_tools/contrib/check_registered_slaves_aws.py
@@ -3,9 +3,10 @@ import argparse
 import sys
 
 from a_sync import block
+
 from paasta_tools.autoscaling.autoscaling_cluster_lib import get_scaler
-from paasta_tools.mesos_tools import get_mesos_master
 from paasta_tools.mesos.exceptions import MasterNotAvailableException
+from paasta_tools.mesos_tools import get_mesos_master
 from paasta_tools.utils import load_system_paasta_config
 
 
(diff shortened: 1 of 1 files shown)


solarkennedy:
For the purposes of this script, those values (0.0) should be fine

davent:
Thanks, @solarkennedy.

PR #1878, opened 2018-06-28
Edit paasta add secret instructions

KevinHock:
To metnion the yelp_servlib client library and necessary prefix.

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -46,7 +46,9 @@ def add_subparser(subparsers):
         help="The name of the secret to create/update, "
              "this is the name you will reference in your "
              "services marathon/chronos yaml files and should "
-             "be unique per service.",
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/secret.py
```diff
@@ -90,12 +92,13 @@ def print_paasta_helper(secret_path, secret_name):
           "main:\n"
           "  cpus: 1\n"
           "  env:\n"
-          "    MY_SECRET_VAR: SECRET({})\n\n"
+          "    PAASTA_SECRET_MY_SECRET_VAR: SECRET({})\n\n"
```
mattmb: Thanks for doing this :-)

Do you think we should go a little further and nudge people to use the same name?

In other words change it to something like:
`PAASTA_SECRET_{}: SECRET({})\n\n".format(secret_name.upper(), secret_name)`

PR #1881, opened 2018-06-29
choose a host in the same pool that the service runs in to query for …

stug:
…smartstack backends

### DIFF
--- a/tests/test_smartstack_tools.py
+++ b/tests/test_smartstack_tools.py
@@ -255,11 +255,12 @@ def test_get_replication_for_instance(
     system_paasta_config,
 ):
     mock_mesos_slaves = [
-        {'hostname': 'host1', 'attributes': {'region': 'fake_region1'}},
-        {'hostname': 'host2', 'attributes': {'region': 'fake_region1'}},
+        {'hostname': 'host1', 'attributes': {'region': 'fake_region1', 'pool': 'default'}},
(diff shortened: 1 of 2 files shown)


tests/test_smartstack_tools.py
```diff
@@ -255,8 +255,8 @@ def test_get_replication_for_instance(
     system_paasta_config,
 ):
     mock_mesos_slaves = [
-        {'hostname': 'host1', 'attributes': {'region': 'fake_region1'}},
-        {'hostname': 'host2', 'attributes': {'region': 'fake_region1'}},
+        {'hostname': 'host1', 'attributes': {'region': 'fake_region1', 'pool': 'default'}},
+        {'hostname': 'host2', 'attributes': {'region': 'fake_region1', 'pool': 'default'}},
```
mattmb: Can you expand the test a bit to test the new logic?

I think we could assert that `get_replication_for_all_services` is called with the correct host?

mattmb: But ship anyway if it is a real pain to do so for some reason :-)

mattmb:
Would you also mind amending the commit message to say "why"

tests/test_smartstack_tools.py
stug: Yup good call, went too fast.

EvanKrall:
Oops, was looking at an out of date diff, you already added the test. Ship it!

PR #1883, opened 2018-06-30
Fix broken link to soa_configs.html

jackchi:
Here is the real link:
http://paasta.readthedocs.io/en/latest/soa_configs.html


solarkennedy:
Thanks!

PR #1885, opened 2018-07-03
Fix `docker_wrapper.py` doesn't see environment variables

vkhromov:
Since PaaSTA passes environment variable values to a Docker container
via `VAR=val docker --env VAR`, fixing `docker_wrapper.py` to read those
values not only from the command line arguments but also from the
environment.

### DIFF
--- a/paasta_tools/docker_wrapper.py
+++ b/paasta_tools/docker_wrapper.py
@@ -33,7 +33,7 @@ MAX_HOSTNAME_LENGTH = 63
 
 
 def parse_env_args(args):
-    result = {}
+    result = {k: v for k, v in os.environ.items()}
     in_env = False
     in_file = False
     for arg in args:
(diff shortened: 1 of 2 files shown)


tests/test_docker_wrapper.py
```diff
@@ -7,7 +7,24 @@
 from paasta_tools import docker_wrapper
 
 
+@contextmanager
+def patch_environ(envs):
```
solarkennedy: cool!

paasta_tools/docker_wrapper.py
```diff
@@ -33,7 +33,7 @@
 
 
 def parse_env_args(args):
-    result = {}
+    result = {k: v for k, v in os.environ.items()}
```
solarkennedy: This is just `dict(os.environ.items())` right?

vkhromov: Oops, sorry, missed this comment.
Yes, it is just `dict(os.environ.items())`.
Unfortunately, I've already merged it.

PR #1886, opened 2018-07-03
Add kubernetes as an instance type

mattmb:
This is needed to get generate_deployments working for kubernetes- yaml
files. I've deliberately left paasta status not implemented for now.
This also adds the list_kubernetes_service_instances script so we can
start with the list|xargs setup_kubernetes_job like pattern before we
add support for the k8s instances to deployd.

NB: I think this should be safe, I've read some code and checked things still work in the example cluster. I guess the itests should flag if I've broken anything else.

### DIFF
--- a/paasta_tools/cli/utils.py
+++ b/paasta_tools/cli/utils.py
@@ -34,6 +34,7 @@ from paasta_tools import remote_git
 from paasta_tools.adhoc_tools import load_adhoc_job_config
 from paasta_tools.api import client
 from paasta_tools.chronos_tools import load_chronos_job_config
+from paasta_tools.kubernetes_tools import load_kubernetes_service_config
 from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.monitoring_tools import _load_sensu_team_data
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -95,6 +95,7 @@
         'paasta_tools/generate_services_yaml.py',
         'paasta_tools/get_mesos_leader.py',
         'paasta_tools/list_marathon_service_instances.py',
+        'paasta_tools/list_kubernetes_service_instances.py',
```
solarkennedy: Can you do the entry_points method instead and prefix it with ` paasta_`?

paasta_tools/cli/utils.py
```diff
@@ -1013,6 +1014,20 @@ def get_instance_configs_for_service(service, soa_dir, type_filter=None):
                     soa_dir=soa_dir,
                     load_deployments=False,
                 )
+        if 'kubernetes' in type_filter:
```
Rob-Johnson: feels like this whole bit of code could be DRY'd up a little. but nbd, we can come back to it

PR #1887, opened 2018-07-03
Make travis fail if we break example cluster

mattmb:
I have three justifications but opinions welcome:
* It seems to be about as stable as the itests give or take
* It occassionally has caught bugs that we don't cover in the itests and
have ended up pushing to prod
* I use it quite a bit for development and it's a pita if it's been
broken by some minor thing that can be hard to work out if it changed
some time in the past.


solarkennedy:
I agree it has been about as stable as the itests.

PR #1888, opened 2018-07-04
Parallelize get_happy_tasks

EvanKrall:


EvanKrall:
Never mind, I remember why we query the individual host the task is on: we need to be sure to query a box where the task should be discoverable. Boxes where other tasks are running, even from the same app, are not necessarily sufficient if your `discover`/`advertise` levels are smaller than the cluster.

This concurrency will solve the immediate problem.

EvanKrall:
deployd uses its own threads, and it looks like asyncio's thread pool is a per-event-loop (so, per base thread) singleton. This could end up launching a lot of threads, but other than GIL contention I think there shouldn't be any interaction between workers.

EvanKrall:
Closed in favor of https://github.com/Yelp/paasta/pull/1924

PR #1893, opened 2018-07-05
Starting point for validation of Tron configs

qui:
Added a JSON schema for Tron configs, and `validate` methods for TronActionConfig and TronJobConfig. `paasta validate` will check the schema and call validate on the configs. Eventually, the goal is to have all of the checks that are done with our pre-commit script and `tronfig` to be included here instead. If `paasta validate` passes, we should be confident that Tron will accept the config.

There's some extra logic in the CLI to support something like `paasta validate -s tron/prod`, so that we can use this in a pre-commit hook for those files, too. Not the cleanest, but I figure it's better than maintaining multiple types of pre-commit checks.

This shows the general approach I'm going for. If this looks reasonable, I'll add tests. And then we can follow up with more check methods.

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -33,6 +33,9 @@ from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_services
 from paasta_tools.cli.utils import PaastaColors
 from paasta_tools.cli.utils import success
+from paasta_tools.tron_tools import list_tron_clusters
+from paasta_tools.tron_tools import load_tron_service_config
+from paasta_tools.tron_tools import MASTER_NAMESPACE
 from paasta_tools.utils import get_services_for_cluster
 from paasta_tools.utils import list_all_instances_for_service
 from paasta_tools.utils import list_clusters
@@ -72,6 +75,18 @@ def invalid_chronos_instance(cluster, instance, output):
 
 def valid_chronos_instance(cluster, instance):
     return success(f'chronos-{cluster}.yaml has a valid instance: {instance}.')
+
+
+def invalid_tron_job(cluster, job, output, filename):
+    return failure(
+        '%s has an invalid job: %s.\n  %s\n  '
+        'More info:' % (filename, job, output),
+        "http://tron.readthedocs.io/en/latest/jobs.html",
+    )
+
+
+def valid_tron_job(cluster, job, filename=None):
+    return success(f'{filename} has a valid job: {job}.')
 
 
 def get_schema(file_type):
@@ -137,7 +152,7 @@ def validate_all_schemas(service_path):
         if os.path.islink(file_name):
             continue
         basename = os.path.basename(file_name)
-        for file_type in ['chronos', 'marathon', 'adhoc']:
+        for file_type in ['chronos', 'marathon', 'adhoc', 'tron']:
             if basename.startswith(file_type):
                 if not validate_schema(file_name, file_type):
                     returncode = False
@@ -209,6 +224,60 @@ def path_to_soa_dir_service(service_path):
     return soa_dir, service
 
 
+def validate_tron(service_path):
+    soa_dir, service = path_to_soa_dir_service(service_path)
+    returncode = True
+
+    if soa_dir.endswith('/tron'):
+        # Makes it possible to validate files in tron/ rather than service directories
+        # TODO: Clean up after migration to services is complete
+        cluster = service
+        soa_dir = soa_dir[:-5]
+        filenames = [filename for filename in os.listdir(service_path) if filename.endswith('.yaml')]
+        for filename in filenames:
+            namespace = os.path.splitext(filename)[0]
+            file_path = os.path.join(service_path, filename)
+            if not validate_schema(file_path, 'tron'):
+                returncode = False
+            if not validate_tron_namespace(namespace, cluster, soa_dir, tron_dir=True):
+                returncode = False
+    else:
+        # Normal service directory
+        for cluster in list_tron_clusters(service, soa_dir):
+            if not validate_tron_namespace(service, cluster, soa_dir):
+                returncode = False
+
+    return returncode
+
+
+def validate_tron_namespace(service, cluster, soa_dir, tron_dir=False):
+    returncode = True
+    job_configs, other_config = load_tron_service_config(service, cluster, soa_dir)
+
+    if tron_dir:
+        display_name = f'{cluster}/{service}.yaml'
+    else:
+        display_name = f'tron-{cluster}.yaml'
+
+    if service != MASTER_NAMESPACE and other_config:
+        returncode = False
+        other_keys = list(other_config.keys())
+        paasta_print(failure(
+            f"{display_name}: Non-{MASTER_NAMESPACE} namespace cannot have other config values, found {other_keys}",
+            "http://tron.readthedocs.io/en/latest/jobs.html",
+        ))
+
+    for job_config in job_configs:
+        check_msgs = job_config.validate()
+        if check_msgs:
+            returncode = False
+            paasta_print(invalid_tron_job(cluster, job_config.get_name(), "\n  ".join(check_msgs), display_name))
+        else:
+            if not tron_dir:
+                paasta_print(valid_tron_job(cluster, job_config.get_name(), display_name))
+    return returncode
+
+
 def validate_chronos(service_path):
     """Check that any chronos configurations are valid"""
     soa_dir, service = path_to_soa_dir_service(service_path)
@@ -269,6 +338,9 @@ def paasta_validate_soa_configs(service_path):
         returncode = False
 
     if not validate_chronos(service_path):
+        returncode = False
+
+    if not validate_tron(service_path):
         returncode = False
 
     return returncode
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -34,8 +34,7 @@ from paasta_tools.cli.utils import list_services
 from paasta_tools.cli.utils import PaastaColors
 from paasta_tools.cli.utils import success
 from paasta_tools.tron_tools import list_tron_clusters
-from paasta_tools.tron_tools import load_tron_service_config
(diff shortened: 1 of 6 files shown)


paasta_tools/tron/tron_timeutils.py
```diff
@@ -28,6 +33,21 @@ def delta_total_seconds(td):
     return (microseconds + (seconds + days * 24 * 3600) * 10**6) / 10**6
 
 
+def check_timedelta(value: Any, param_name: str) -> Tuple[bool, str]:
```
qui: This is stricter than Tron's validation. This only allows d, h, m, s (for days, hours, minutes, seconds), while Tron allows any of ['m', 'min', 'mins', 'minute', 'minutes'] for minutes, and similar things for the other units. 

I was trying to avoid copying too much complex validation code from Tron. Was thinking we could do a similar thing (i.e. being less flexible than Tron) for fields like `schedule`.

paasta_tools/cli/cmds/validate.py
```diff
@@ -209,6 +224,60 @@ def path_to_soa_dir_service(service_path):
...
 
+def validate_tron(service_path):
+    soa_dir, service = path_to_soa_dir_service(service_path)
+    returncode = True
+
+    if soa_dir.endswith('/tron'):
+        # Makes it possible to validate files in tron/ rather than service directories
+        # TODO: Clean up after migration to services is complete
+        cluster = service
+        soa_dir = soa_dir[:-5]
```
solarkennedy: These 5 characters are the /tron right? 
How about just .strip('/tron')?

paasta_tools/cli/cmds/validate.py
```diff
@@ -209,6 +224,60 @@ def path_to_soa_dir_service(service_path):
...
+            file_path = os.path.join(service_path, filename)
+            if not validate_schema(file_path, 'tron'):
+                returncode = False
+            if not validate_tron_namespace(namespace, cluster, soa_dir, tron_dir=True):
+                returncode = False
+    else:
+        # Normal service directory
+        for cluster in list_tron_clusters(service, soa_dir):
+            if not validate_tron_namespace(service, cluster, soa_dir):
+                returncode = False
```
solarkennedy: Can we validate the schema here too?

paasta_tools/tron/tron_timeutils.py
solarkennedy: This sounds good to me. I think tron's `schedule` field is unnecessarily flexible, but in a different way.

Are you planning on preemptively fixing the MASTER files that we have already?

paasta_tools/tron_tools.py
```diff
@@ -255,6 +280,47 @@ def get_cleanup_action(self, default_paasta_cluster):
...
+    def check_max_runtime(self) -> Tuple[bool, str]:
+        max_runtime = self.get_max_runtime()
+        return check_timedelta(max_runtime, 'max_runtime')
+
+    def check_expected_runtime(self) -> Tuple[bool, str]:
+        expected_runtime = self.get_expected_runtime()
+        return check_timedelta(expected_runtime, 'expected_runtime')
+
+    def check_actions(self) -> Tuple[bool, List[str]]:
+        actions = self.get_actions('dummy-default')
```
solarkennedy: Why is it that we need a dummy-default here? How is it that the object itself doesn't have enough information to "know" what the config is in the first place?

Should the cluster we loaded he config from be part of the object? (self.cluster?) or can we get away with None?

solarkennedy:
The general approach looks fine to me. I like your end goal of making it so that if `paasta validate` passes, then we can be confident it can work.

But, is it really so bad that we use a combination of tronfig and paasta validate? Seeing the date validation code, there may be some things that are "just easier" in the tron codebase to do, like command context variable interpolation, etc. What if this thing "just" also ran tronfig -v for validating "everything else" ?

paasta_tools/tron_tools.py
qui: Ah, it totally can be None. Fixed.

paasta_tools/cli/cmds/validate.py
qui: Yeah, that happens in https://github.com/Yelp/paasta/blob/28b51a7321127102cf2c0a684408986485e9bc7d/paasta_tools/cli/cmds/validate.py#L155 for regular service directories. But since files in tron/* don't start with `tron-`, we have to add the schema validation here.

paasta_tools/cli/cmds/validate.py
qui: strip removes any combination of the characters, so it'll remove other stuff if the service name ends in one of those.

paasta_tools/tron/tron_timeutils.py
qui: Yeah, I would fix anything that doesn't pass validation when I changed the pre-commit hook.

paasta_tools/cli/cmds/validate.py
```diff
@@ -209,6 +224,60 @@ def path_to_soa_dir_service(service_path):
...
+            "http://tron.readthedocs.io/en/latest/jobs.html",
+        ))
+
+    for job_config in job_configs:
+        check_msgs = job_config.validate()
+        if check_msgs:
+            returncode = False
+            paasta_print(invalid_tron_job(cluster, job_config.get_name(), "\n  ".join(check_msgs), display_name))
+        else:
+            if not tron_dir:
```
chlgit: What will happen if tron_dir?

qui: It only prints messages if something is invalid, not the successful validation messages. It was too noisy / easy to miss the invalid messages for tron prod otherwise.

qui:
Hmm, yeah, it would be nice to let Tron do as much validation for us as possible. We could do something like just validate PaasTA-specific values in the instance configs, then generate the complete config for Tron, then run tronfig on that. I'm going to update this review.

keymone:
the only thing that worries me is duplication of validation logic, but i don't see obvious way around it. adding tron to paasta dependencies is a no-go. json schema is great but we'd need to use it in tron for validation as well and keep both in sync (not a bad idea tbh). it seems for things that go beyond types and form we're only left with `tronfig -c`, which isn't great when multiple namespaces have been touched - each one could be fine, but together they could fail.

but we should definitely ship this. :ship:

qui:
@keymone @solarkennedy Updated, take another look if you'd like.

Significant changes:
* Moved most logic to `validate_complete_config` in tron_tools, which calls both the `validate` method on job configs and shells out to `tronfig`
* took out the expected_runtime/max_runtime validation (`tronfig` can validate that instead). `validate` doesn't do much anymore, but leaves room for additional stuff later.
* tests!

qui:
@keymone I agree `tronfig` doesn't catch everything right now, but I think we can update it so that it does in the future.

PS. manual testing: `paasta validate` passes on all the tron directories we have now, plus our test service with Mesos jobs added.

paasta_tools/tron_tools.py
```diff
@@ -453,6 +419,47 @@ def create_complete_config(service, soa_dir=DEFAULT_SOA_DIR):
...
+        ['tronfig', '-', '-V', '-n', service],
+        input=complete_config,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        encoding='utf-8',
+    )
+    if proc.returncode != 0:
+        process_errors = proc.stderr.strip()
+        if process_errors:  # There was an error in validation
+            print(proc.stderr)
```
chlgit: Using paasta_print here? Or could we combine proc.stderr and proc.stdout and return them together?

qui: I'll change it to paasta_print. The reason I don't return stderr is because I don't want validation to fail if the `tronfig` command fails to run for some reason (e.g. it doesn't exist). If `tronfig` runs, whether or not validation passed, all the messages will be in stdout.

paasta_tools/cli/schemas/tron_schema.json
```diff
@@ -0,0 +1,136 @@
...
+                "realert_every": {
+                    "type": "integer",
+                    "minimum": 1,
+                    "exclusiveMinimum": false
+                },
+                "dependencies": {
+                    "type": "array",
+                    "items": {"type": "string"}
+                },
+                "irc_channels": {
```
solarkennedy: Can you add slack_channels while you are at it? I guess I need to add that to our other schemas

paasta_tools/tron_tools.py
```diff
@@ -148,14 +156,23 @@ def get_calculated_constraints(self):
             constraints.extend(self.get_pool_constraints())
             return constraints
 
+    def validate(self) -> List[str]:
+        # Use InstanceConfig to validate shared config keys like cpus and mem
+        error_msgs = super(TronActionConfig, self).validate()
```
solarkennedy: I don't totally understand this. If we are already inheriting from InstanceConfig, is the inherited method not enough?
Is the purpose here just to make the return errors prettier with the instance name?
If so I think it would be OK to put that in the InstanceConfig

qui: It used to call other methods that were on TronActionConfig (e.g. check_expected_runtime). I guess I left this here to leave room for that in the future.

paasta_tools/cli/schemas/tron_schema.json
qui: Sure, done.

vkhromov:
If we use a JSON schema to validate configs and then an ad hoc Python code to actually read the config and represent it in the application, then there could be differences between declared schema in JSON schema and de facto schema assumed by the Python code.
Can we use some kind of code generator to generate a Python code to read the config from the JSON schema or from another schema description (Thrift, protobuf, etc)?

qui:
@vkhromov I think the *Config classes are meant to validate more complex rules that we can't represent in a JSON schema. For example, checking that the deploy_group exists. The Python code also adds defaults (which might be possible in other types of schema descriptions, I'm not too familiar). 

I agree that it's brittle that we need to update both the JSON schema and *Config classes when we add/modify fields, though. I'm going to merge this for now, but if there's a better way to represent these configs, let's think about how to do it for all instance types!

PR #1898, opened 2018-07-11
Add --user flag to `paasta docker_exec`

chriskuehl:
This lets you do `paasta docker_exec ... --user=root`, which is something I often want to do when exec-ing into a container (the `nobody` user isn't always so useful for troubleshooting).

Currently my workflow is `paasta docker_exec`, let it find me a container to exec into, then SSH to the host again and `docker exec --user=root` into it. This argument will let me skip those manual steps.


solarkennedy:
shlex eh? :)

PR #1901, opened 2018-07-18
Don't load deployments for Tron validation

qui:
Adds a setting to TronJobConfig so it won't try to load deployments, since deployments.json isn't always available in soaconfigs for validation.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -138,11 +138,6 @@ class TronActionConfig(InstanceConfig):
     def get_expected_runtime(self):
         return self.config_dict.get('expected_runtime')
 
-    def get_docker_url(self):
-        if self.branch_dict:
-            return super(TronActionConfig, self).get_docker_url()
-        return ''
-
     def get_calculated_constraints(self):
         """Combine all configured Mesos constraints."""
         constraints = self.get_constraints()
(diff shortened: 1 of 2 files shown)


qui:
It's generated here: https://github.com/Yelp/paasta/blob/master/paasta_tools/generate_deployments_for_service.py#L145

It's the mapping between deploy group names (like "everything.prod") to the docker image/git SHA that's currently deployed for that group. We generate them for /nail/etc/services but don't commit it into git.

paasta_tools/tron_tools.py
```diff
@@ -138,6 +138,11 @@ def get_requires(self):
     def get_expected_runtime(self):
         return self.config_dict.get('expected_runtime')
 
+    def get_docker_url(self):
+        if self.branch_dict:
+            return super(TronActionConfig, self).get_docker_url()
+        return ''
```
solarkennedy: Do you need to call get_docker_url in the situation where we don't have deployments?
This pattern looks like it could cause bugs, the other `get_docker_url` will raise if it doesn't have something to work with. Why does tron need to have different behavior?

qui: Hm, I've pushed a commit to only call get_docker_url in `format_tron_action_dict` if there's a docker_image (meaning the deployments were loaded). That way, other callers of get_docker_url will get the regular behavior (where it raises an error). Does that look better @solarkennedy?

We still need the empty string fallback in format_tron_action_dict because we use tronfig to validate, which requires generating the full config. For other instance types, we only generate the full config when we're deploying, and we know deployments.json will be there.

solarkennedy: Hmm. I can't think of any better way to make this cleaner without plumbing a "load_deployments" bool all the way in or something. 

Your explanation for why this is a problem for tron only make sense. shipit.

PR #1902, opened 2018-07-18
Local run: Don't ask me to pick a deploy group when there's only one

mjksmith:
```
mjksmith@mydevbox:~/paasta (master) $ paasta local-run --pull -s my_cool_service -c cluster -I
Please pick a deploy group from the choices below (or "quit" to quit):
 0: everything

    quit

Enter number or name; return for next page
? 
```

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2756,11 +2756,11 @@ def prompt_pick_one(sequence: Collection[str], choosing: str) -> str:
         )
         sys.exit(1)
 
-    if len(sequence) == 1:
-        return sequence[0]
-
     global_actions = [str('quit')]
-    choices = [(str(item), str(item)) for item in sequence]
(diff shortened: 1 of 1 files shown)


paasta_tools/utils.py
```diff
@@ -2757,7 +2757,10 @@ def prompt_pick_one(sequence: Collection[str], choosing: str) -> str:
         sys.exit(1)
 
     global_actions = [str('quit')]
-    choices = [(str(item), str(item)) for item in sequence]
```
solarkennedy: Any idea why we `str()`'d this originally? Does the choice library tolerate "Nones" and other nonsense?

mjksmith: I'm not sure, but now that we're using mypy we know sequence is only going to contain strings so it's fine.

Also @solarkennedy I don't have write access any more so can you merge this and/or re-add me? Thanks :)

PR #1903, opened 2018-07-18
Only pass secret environment variables to execlpe

huadongliu:
Before [the execlpe change](https://github.com/Yelp/paasta/pull/1875), local-run printed:
paasta_docker_wrapper run --env ENV_VAR=VAL1 --env SECRET_VAR=VAL2

After the execlpe change, local-run prints
paasta_docker_wrapper run --env ENV_VAR --env SECRET_VAR

With this change, local-run will print
paasta_docker_wrapper run --env ENV_VAR=VAL1 --env SECRET_VAR

Testing:
$ .tox/py36/bin/paasta local-run -s example_happyhour -i main --pull -C /bin/bash
...
Running docker command:
paasta_docker_wrapper run --env PAASTA_SERVICE=example_happyhour --env PAASTA_INSTANCE=main --env PAASTA_CLUSTER=norcal-devc --env PAASTA_DEPLOY_GROUP=dev.everything ... 
**--env MY_SECRET_VAR** --memory=600m --memory-swap=664m --cpu-period=100000 ...

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -402,10 +402,13 @@ def get_container_name():
 
 def get_docker_run_cmd(
     memory, chosen_port, container_port, container_name, volumes, env, interactive,
-    docker_hash, command, net, docker_params, detach,
+    docker_hash, command, net, docker_params, detach, secret_env={},
 ):
     cmd = ['paasta_docker_wrapper', 'run']
-    for k in env.keys():
+    for k, v in env.items():
+        cmd.append('--env')
+        cmd.append(f'{k}={v}')
+    for k in secret_env.keys():
         cmd.append('--env')
         cmd.append(f'{k}')
     cmd.append('--memory=%dm' % memory)
@@ -617,6 +620,7 @@ def run_docker_container(
             sys.exit(1)
     else:
         chosen_port = pick_random_port(service)
+    secret_environment = {}
     environment = instance_config.get_env_dictionary()
     if not skip_secrets:
         secret_environment = decrypt_secret_environment_variables(
@@ -627,7 +631,9 @@ def run_docker_container(
             cluster_name=instance_config.cluster,
             secret_provider_kwargs=secret_provider_kwargs,
         )
-        environment.update(secret_environment)
+        for k in secret_environment.keys():
+            del environment[k]
+
     local_run_environment = get_local_run_environment_vars(
         instance_config=instance_config,
         port0=chosen_port,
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -627,7 +631,9 @@ def run_docker_container(
             cluster_name=instance_config.cluster,
             secret_provider_kwargs=secret_provider_kwargs,
         )
-        environment.update(secret_environment)
+        for k in secret_environment.keys():
```
solarkennedy: This seems strange to me. before this line updated the environment, now we are deleting it. Why do we have to delete it if it wasn't in there in the first place?

huadongliu: If you don't delete secret envs, you will see something like this:

--env MY_SECRET_VAR=SECRET(test-secret) ... --env MY_SECRET_VAR

The first part comes from environment and the second part comes from secret_environment.

mattmb: It's even worse isn't it because `SECRET(test-secret)` will have been decrypted at that point. So the one thing we want to hide is actually still being shown!

paasta_tools/cli/cmds/local_run.py
```diff
@@ -681,7 +688,7 @@ def run_docker_container(
     else:
         paasta_print('Running docker command:\n%s' % PaastaColors.grey(joined_docker_run_cmd))
 
-    merged_env = {**os.environ, **environment}
+    merged_env = {**os.environ, **secret_environment}
```
mattmb: This doesn't seem quite right to me? Surely we need both environment and secret environment passed to the exec?

paasta_tools/cli/cmds/local_run.py
huadongliu: No, "--env MY_SECRET_VAR=SECRET(test-secret)" won't be shown. Only "--env MY_SECRET_VAR" is shown. That is why I delete secret_environment from the super set environment dict.

paasta_tools/cli/cmds/local_run.py
huadongliu: Non-secret environments are passed with explicit docker --env ENV_VAR=ENV_VAL while secret environments are passed with execlpe.

mattmb: Ah I see, I think we should consolidate to have it all passed by execlpe.

That's the case for Mesos now I think so we want to keep local-run as consistent as we can. Naturally it shouldn't make much difference, but you never know.

PR #1904, opened 2018-07-19
validate valid team name

chlgit:
This is the first pass for TRON-369, which validate team name at pre-commit.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -34,6 +34,7 @@ from paasta_tools.utils import NoConfigurationForServiceError
 from paasta_tools.utils import NoDeploymentsAvailable
 from paasta_tools.utils import paasta_print
 
+from paasta_tools.cli.utils import list_teams
 
 MASTER_NAMESPACE = 'MASTER'
 SPACER = '.'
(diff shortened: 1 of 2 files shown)


solarkennedy:
As discussed, try to pull in the code from our existing check for this and unify things (we we can check-all-the-teams).

paasta_tools/tron_tools.py
```diff
@@ -270,6 +271,20 @@ def get_cleanup_action(self, default_paasta_cluster):
...
+    def check_monitoring(self) -> Tuple[bool, List[str]]:
+        msgs: List[str] = []
+        monitoring = self.get_monitoring()
+        checks_passed = True
+        if monitoring is not None:
+            team_name = monitoring.get('team', None)
+            if team_name is None:
+                checks_passed = False
+                msgs.append('Team name is required for monitoring')
+            elif team_name not in list_teams():
```
qui: In our typical team list, is 'noop' a valid team? Or does it need to be a special case (as it is in the pre-receive hook)? 

Also, maybe we should provide suggestions if there's a valid team name that is similar.

chlgit: noop is a valid team name. suggestion is added.

paasta_tools/tron_tools.py
```diff
@@ -270,6 +271,20 @@ def get_cleanup_action(self, default_paasta_cluster):
         action_dict['name'] = 'cleanup'
         return self._get_action_config(action_dict, default_paasta_cluster)
 
+    def check_monitoring(self) -> Tuple[bool, List[str]]:
```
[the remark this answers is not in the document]
chlgit: Ok, got it. Fixed.

chlgit:
Yeah, I thought about that. And I think it makes sense if we refactor the code and centralize the logic when we move the monitoring part from Tron config to monitoring.yaml.

qui:
I see. Either way, I think we'll have to allow users to override monitoring.yaml settings in their job configs (like we do with marathon), so it might be worth thinking about now. If it turns out to be a big change, then we can decide to put it off.

chlgit:
I did a quick check but couldn't see an easy change for that. I will say let's do that in the separated PR.

PR #1905, opened 2018-07-19
Catch exceptions on format_marathon_app_dict

mattmb:
We currently only catch the ways we know it can go wrong. There's a lot
of user input and our validation with commit hooks can only be so good.
We need deployd to continue if only one marathon app dict is malformed
so this catches all Exceptions.


solarkennedy:
I think this is the right thing to do

paasta_tools/list_marathon_service_instances.py
```diff
@@ -102,7 +100,9 @@ def get_desired_marathon_configs(soa_dir):
                 cluster=cluster,
                 instance=instance,
             )
-        except (NoDeploymentsAvailable, NoDockerImageError):
+        # Not ideal but we rely on a lot of user input to create the app dict
+        # and we really can't afford to bail if just one app definition is malformed
+        except Exception:
```
solarkennedy: Hmm, how about just log like above?

PR #1906, opened 2018-07-19
Add cpu + mem constraints to k8s

mattmb:
Easier that I thought, we use the same values we use for marathon,
limits are "hard" requirements are "soft" but used by the scheduler.

I've tested that the code works and creates a deployment with the limits
on the container but haven't tried to actually push a container to see
what happens when you hit the limit.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -46,6 +46,7 @@ from kubernetes.client import V1ObjectMeta
 from kubernetes.client import V1PodSpec
 from kubernetes.client import V1PodTemplateSpec
 from kubernetes.client import V1Probe
+from kubernetes.client import V1ResourceRequirements
 from kubernetes.client import V1RollingUpdateDeployment
 from kubernetes.client import V1Volume
 from kubernetes.client import V1VolumeMount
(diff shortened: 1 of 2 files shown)


tests/test_kubernetes_tools.py
```diff
@@ -275,6 +276,28 @@ def test_get_kubernetes_environment(self):
...
+            'paasta_tools.kubernetes_tools.KubernetesDeploymentConfig.get_cpu_burst_pct', autospec=True,
+            return_value=200,
+        ), mock.patch(
+            'paasta_tools.kubernetes_tools.KubernetesDeploymentConfig.get_mem', autospec=True,
+            return_value=2048,
+        ):
+            assert self.deployment.get_resource_requirements() == V1ResourceRequirements(
+                limits={
+                    'cpu': 0.6,
+                    'memory': 2048,
```
vkhromov: Please see the comment above, this one and the one below probably should be `str(2048*1024)`.

paasta_tools/kubernetes_tools.py
```diff
@@ -347,6 +348,18 @@ def get_kubernetes_environment(self) -> List[V1EnvVar]:
         ]
         return kubernetes_env
 
+    def get_resource_requirements(self) -> V1ResourceRequirements:
+        return V1ResourceRequirements(
+            limits={
+                'cpu': self.get_cpus() * self.get_cpu_burst_pct() / 100,
+                'memory': self.get_mem(),
```
vkhromov: From [the doc](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-memory) it seems that Kubernetes expects `memory` as a `str` with either a suffixed numeric value or a number of bytes, while [get_mem()](https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L315) returns `float` with Mb.  If that's true then there should be some conversion.
The same for `memory` below.

While here, could you please also double check that `cpu` value is OK here and below.

mattmb: Good catch 👍 I don't think many of our services will run with 1024th of the memory they ask for! Fixup inbound.

I'm pretty sure the CPU stuff is okay, I at least read that bit of the docs pretty closely.

PR #1908, opened 2018-07-19
RFC: smartstack errors meteorite emitter

solarkennedy:
cc @somic @bplotnick @jnb 

I want to have metrics from our service mesh available "for free" to our service owners. I'm thinking it will be useful to them for the following use-cases:
* free dashboards  (y/paasta-status)
* easy to integrate with the slo framework
* easy to retrieve for auto-rollback / push on green tooling

Additionally, when we move to envoy, we'll be able to update this script and live in "dual" mode, and these other tools won't have to change maybe? Maybe we'll need to translate the metric name.

I'm sure we'll have to update the metric whitelist over time, and maybe munge them to be friendlier.

Also the hack to just assume the backend name is service.instance isn't super great long term.

### DIFF
--- a/paasta_tools/contrib/emit_synapse_services_metrics.py
+++ b/paasta_tools/contrib/emit_synapse_services_metrics.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python3
+import yelp_meteorite
+
+from paasta_tools import utils
+from paasta_tools.smartstack_tools import retrieve_haproxy_csv
+
+
+GUAGES = ['ereq', 'rate']
+COUNTERS = ['hrsp_2xx', 'hrsp_3xx', 'hrsp_4xx', 'hrsp_5xx']
+
+
+def parse_haproxy_backend_name(backend):
+    # Currently a hack, the backend doesn't always match service.instance,
+    # but at least the data will be reported.
+    return backend.split('.')[0], backend.split('.')[1]
+
+
+def report_metric_to_meteorite(backend, metric, value, paasta_cluster):
+    try:
+        paasta_service, paasta_instance = parse_haproxy_backend_name(backend)
+    except IndexError:
+        return
+
+    meteorite_dims = {
+        'service_name': paasta_service,
+        'paasta_cluster': paasta_cluster,
+        'instance_name': paasta_instance,
+    }
+    path = f'paasta.service.requests.{metric}'
+    if metric in GUAGES:
+        guage = yelp_meteorite.create_gauge(path, meteorite_dims)
+        guage.set(value)
+    elif metric in COUNTERS:
+        counter = yelp_meteorite.create_counter(path, meteorite_dims)
+        counter.count(value)
+    else:
+        raise ValueError(f"{metric} hasn't been configured as a guage or counter")
+    print(f"Sent {path}: {value} to meteorite")
+
+
+def report_all_metrics_to_meteorite(csv, paasta_cluster):
+    for row in csv:
+        if row['svname'] == 'BACKEND':
+            for metric in GUAGES + COUNTERS:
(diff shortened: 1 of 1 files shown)


bplotnick:
Overall, this seems fine, though I'm wondering the tradeoff of this vs emitting meteorite metrics from syslog2scribe, which already processes the access logs - maybe it's complexity?

Additionally, this "pull-based" mechanism seems like it would be better suited as a [fullerite](https://github.com/Yelp/fullerite) collector rather than using meteorite, though I'd defer to the metrics folks to tell me why.

Considering you want something for this _now_, rather than waiting for our Envoy migration to complete, this isn't that useful, but i'll lay out my idea for our approach for our service mesh metrics anyway:

Envoy natively supports statsd for push metrics and prometheus for pull. Either way, at Yelp, we don't use vanilla statsd (we use our extension, meteorite) and we don't use prometheus (we use fullerite). We'll have to do some work either way (i'm actually not sure - we might be able to consume statsd metrics, but we want tagging, which is not supported in the native statsd protocol).

My plan was to go for push metrics by implementing an Envoy "StatSink" for the meteorite protocol (internally, this is COREBACK-5812). Since meteorite appears to be similar to the DataDog statsd protocol, I think it shouldn't be too hard.

Alternatively, we could go for pull-based, though I think the approach here would be to parse prometheus metrics in fullerite rather than modifying Envoy. The advantage here is that Envoy pre-aggregates metrics and even supports emitting histograms.

solarkennedy:
@bplotnick 

* syslog2scribe can only see errors, it can't really see rates of stuff, which is something I want
* fullerite is good, but it would put dimensions I don't want (hostname) and not be able to easily apply the metrics I do want (paasta service). Additionally I need these counters to be aggregated per region, which meteorite (statsite) can do, but fullerite wont
* Envoy may have native support, but we need/want to have all the good variables as dimensions, including what version of the code served the request (some day)

bplotnick:
> syslog2scribe can only see errors, it can't really see rates of stuff, which is something I want

It gets every access log message, not just errors. The only-error thing is just for that one stream. In fact it emits a zipkin span for every single request (with a few small exceptions) from the Zipkin Firehose project. So (if we had a meteorite library), we potentially could emit metrics.

> fullerite is good, but it would put dimensions I don't want (hostname) and not be able to easily apply the metrics I do want (paasta service). Additionally I need these counters to be aggregated per region, which meteorite (statsite) can do, but fullerite wont

Oh i didn't realize that it was that restrictive on the dimensions. The aggregation per-region is usually done in something like SignalFX. This way it is much more flexible on what you aggregate.

> Envoy may have native support, but we need/want to have all the good variables as dimensions, including what version of the code served the request (some day)

I was going under the assumption that you could arbitrarily inject tags from the request. Our plan anyway was to tag endpoints with the version so that we could do metadata-based routing (e.g the "personal canary" that we mentioned in the all-hands)

However, i'm now looking at the [stats documentation](https://www.envoyproxy.io/docs/envoy/latest/api-v2/config/metrics/v2/stats.proto) on Envoy and I realize that you can't inject the endpoint metadata as tags (this is usually called `UPSTREAM_METADATA` in other parts of the Envoy code). I think it would be really easy to add, but just another thing that we'd need to consider

solarkennedy:
While I could do this in c++ in syslog 2 scribe, I would have to re-implement bucketing and counting and all those things. It does allow for fancier things in the future though than these "naive" counters from the csv. Let's start with this as a v1 though.

I agree that we'll have to wait for envoy to do anything much more fancy than this, but we can get error rates in "now" and add more fancy things later, we'll have a consistent metric namespace and dimensions.

bplotnick:
sg2m. I appreciate you indulging me 😃

PR #1910, opened 2018-07-19
check valid nodes

chlgit:
The second pass for TRON-369, which validate nodes at pre-commit.


chlgit:
Yeah, I thought tronfig too. What I was really thinking is TRON-308. I am hoping that we can render the command in paasta validate in the near future, so it can catch more errors. To achieve that, we need master configure. And since we need master config anyway, so I just validate nodes here.

We can also choose to put more responsibility on tronfig. But if I recall it correctly, from our last discussion, we might want to retire tronfig validation in the future. Or anything change since then?

PR #1911, opened 2018-07-20
Add --ask param to start/stop/restart

matthewbentley:
This is something I was surprised to not find.

This was only a few minutes of my time and I'm not super attached to it, so feel free to not include it.

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -57,6 +57,12 @@ def add_subparser(subparsers):
             help="define a different soa config directory",
         )
         status_parser.set_defaults(command=cmd_func)
+        status_parser.add_argument(
+            '--ask',
+            dest='ask',
+            action='store_true',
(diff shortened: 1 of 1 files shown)


solarkennedy:
I'm going to not include. I think it is OK for a cli tool to do exactly what you ask, but then if it tries to be smart (do more than one thing) it will ask. This is just like apt-get, where if you apt-get install one thing and it only needs to install one thing, it won't ask. But if it needs to install more than one thing to do what you requested, it will ask.

matthewbentley:
The original reason for this is that it won't ask if you try to stop (for example) all the instances of a service.  So doing `paasta stop -c pnw-prod -s yelp_res` will stop all instances without asking.  I'm not sure if that should fall under doing more than one thing.

solarkennedy:
Oh right.

How about this then, can you make a PR that asks if the affected *instances* is > 1?

matthewbentley:
👍 IIRC the original reason for doing if services > 1 was to keep the original behavior (in case it is used in a script somewhere).

PR #1913, opened 2018-07-24
Unified __repr__ across paasta instance objects and send cluster into all TronActionConfigs

solarkennedy:
I think this is a change that we wanted anyway.

While working on `paasta local-run` for tron, I was chasing down there `None` was coming from when trying to pull docker images. In reaction to that, I added some "defensive" code to protect against `None` in our InstanceConfigs and unified the `__repr__` so Tron could use it.

While doing that though, I needed to get cluster in everywhere. This change allows us to set `cluster` at the Tron Job level instead of each action.

Note: I'm a little concerned that I don't fully understand the intent of the parameterized test on :186 on `test_tron_tools.py`.

### DIFF
--- a/tests/test_tron_tools.py
+++ b/tests/test_tron_tools.py
@@ -53,6 +53,7 @@ class TestTronActionConfig:
         action_config = tron_tools.TronActionConfig(
             service='my_service',
             instance=tron_tools.compose_instance('cool_job', 'print'),
+            cluster="fake-cluster",
             config_dict=action_dict,
             branch_dict={},
         )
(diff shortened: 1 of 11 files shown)


tests/test_tron_tools.py
```diff
@@ -178,15 +183,10 @@ def test_get_action_config(
         assert action_config == tron_tools.TronActionConfig(
             service=expected_service,
             instance=tron_tools.compose_instance('my_job', 'normal'),
-            config_dict={
-                'name': 'normal',
-                'command': 'echo first',
-                'cluster': expected_cluster,
```
qui: I think this would pass if you just took out 'cluster' from this dict.
I guess this works, but only because action_dict is mutable. It gets changed in the course of format_tron_action_dict (the service and deploy group get added in some cases) so I think it's better to assert on a separate dict.

solarkennedy: Got it.

solarkennedy:
Yea I was working "action up", but if I updated the docs then I should update the jobconfig (not just action) too :)

PR #1914, opened 2018-07-25
Added a remote_run option to override dynamodb aws_region config

huadongliu:
When tron runs a remote-run job on pnw-prod, the dynamodb name is taskproc-events-pnw-prod. However, the aws_region is us-west-1, which only has taskproc-events-norcal-prod.

An alternative fix is to infer aws region from paasta cluster name. For example, region will be us-west-2 if the paasta cluster is pnw-prod. Since the problem is only seen with tron (which is not an usual paasta remote-run configuration), I choose to add an option to override the incorrect aws_region config on tron servers. 

Tested on mesosstage (us-west-1) against pnw-devc (us-west-2) cluster.

### DIFF
--- a/paasta_tools/paasta_remote_run.py
+++ b/paasta_tools/paasta_remote_run.py
@@ -242,6 +242,7 @@ def build_executor_stack(
     run_id,
     system_paasta_config,
     framework_staging_timeout,
+    region,
 ):
 
     cluster_fqdn = system_paasta_config.get_cluster_fqdn_format().format(cluster=cluster)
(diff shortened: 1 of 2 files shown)


paasta_tools/paasta_remote_run.py
```diff
@@ -242,6 +242,7 @@ def build_executor_stack(
     run_id,
     system_paasta_config,
     framework_staging_timeout,
+    region,
```
keymone: needs to  default to `None`?

huadongliu: Added default for args.aws_region.

PR #1915, opened 2018-07-27
Allow chronos rerun to skip services that have not been deployed

solarkennedy:
Fixes internal ticket PAASTA-14651

### DIFF
--- a/paasta_tools/chronos_tools.py
+++ b/paasta_tools/chronos_tools.py
@@ -43,6 +43,7 @@ from paasta_tools.utils import InvalidJobNameError
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import load_v2_deployments_json
 from paasta_tools.utils import NoConfigurationForServiceError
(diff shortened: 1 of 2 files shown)


qui:
We could silently miss Chronos job configs for jobs that are really related. There might be a weird case where the deploy group for a related job is changed to something incorrect?

Do you think that's worth worrying about?

Also, since you mentioned more informative error messages - one idea I thought of was adding the service name to the DeploymentsJsonV2 constructor, since load_v2_deployments_json has the service name.

solarkennedy:
missed: You are right, it is possible. I guess we could do two passes, on the first pass we find related jobs with load_deployments=False, *then* reload only the related ones with load_deployments=True once we know which ones we want. Let me think about it.

error messages: Yes. I have a stash with what you are describing which is half in progress. I'll follow through on that, but via a second PR.

solarkennedy:
@qui I thought about it, and I think the likelihood of a job depending on another job that is not deployed (and silently skipping it) is rare enough to not worry about. @macisamuele do you agree?

PR #1917, opened 2018-07-30
Fix k8s ebs bounce

mattmb:
* pin k8s service instances with EBS to downthenup bounce method
* pin k8s service instances with EBS to instances=1

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -500,6 +500,12 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
     def get_sanitised_instance_name(self) -> str:
         return self.get_instance().replace('_', '--')
 
+    def get_instances(self) -> int:
+        """ For now if we have an EBS instance it means we can only have 1 instance
+        since we can't attach to multiple instances. In the future we might support
+        statefulsets which are clever enough to manage EBS for you"""
+        return 1 if self.get_aws_ebs_volumes() else super().get_instances()
+
     def format_kubernetes_app(self) -> V1Deployment:
         """Create the configuration that will be passed to the Kubernetes REST API."""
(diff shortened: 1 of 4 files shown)


vkhromov:
TBH I'm not sure that this method is OK, since we'll implicitly override users' configs setting, so they probably would be surprised by i.e. discovering a single running instance after specifying something like `instances: 3` without any notifications.
Probably it's better to have pre-commit hook and to explicitly explain to users that they need to fix in their configs.

Rob-Johnson:
I think I agree - we should refuse to deploy (and hopefully catch the problem in pre-commit) if we had unmatchable constraints, rather than changing the settings without warning.

paasta_tools/kubernetes_tools.py
```diff
@@ -246,6 +246,10 @@ def copy(self) -> "KubernetesDeploymentConfig":
     def get_bounce_method(self) -> str:
         """Get the bounce method specified in the service's kubernetes configuration."""
         # map existing bounce methods to k8s equivalents.
+        # but if there's an EBS volume we must downthenup to free up the volume.
+        # in the future we may support stateful sets to dynamically create the volumes
+        if self.get_aws_ebs_volumes():
+            return 'Recreate'
```
keymone: KUBE_DEPLOY_STATEGY_MAP['downthenup']

just to be consistent with the line below

or even:

    if self.get_aws_ebs_volumes():
        strategy = 'downthenup'
    else:
        strategy = self.config_dict.get('bounce_method', 'crossover')
    return KUBE_DEPLOY_STATEGY_MAP[strategy]

mattmb:
Ok, makes sense to me, we should raise an error or something. Maybe I'll revisit the error handling for `format_kubernetes_deployment` in general.

PR #1919, opened 2018-07-31
Set k8s deployment spec version

mattmb:
You can omit but k8s then chooses the latest version which seems to be a
beta spec. So I think we ought to hard code the version of the schema
here.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -479,6 +479,8 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
 
         code_sha = get_code_sha_from_dockerurl(docker_url)
         complete_config = V1Deployment(
+            api_version='apps/v1',
+            kind='Deployment',
             metadata=V1ObjectMeta(
                 name="{service}-{instance}".format(
                     service=self.get_sanitised_service_name(),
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -479,6 +479,8 @@ def format_kubernetes_app(self) -> V1Deployment:
 
         code_sha = get_code_sha_from_dockerurl(docker_url)
         complete_config = V1Deployment(
+            api_version='apps/v1',
+            kind='Deployment',
```
vkhromov: Can you please double check that we need to set the `kind` as well?
I expect k8s to able to deduce the kind from `V1Deployment`.

tests/test_kubernetes_tools.py
```diff
@@ -493,6 +493,8 @@ def test_format_kubernetes_app_dict(self):
                 force_bounce=mock_get_force_bounce.return_value,
             )
             expected = V1Deployment(
+                api_version='apps/v1',
+                kind='Deployment',
```
vkhromov: Please see the comment above.

paasta_tools/kubernetes_tools.py
vkhromov: Actually, looking at `V1` part in the `V1Deployment`, shouldn't k8s deduce the version as well?

PR #1920, opened 2018-07-31
Add flag to disable reservation cleanup

mattmb:
This shouldn't be needed on our patched version of k8s that doesn't send
offers to maint'd hosts. This adds a flag so I can disable it in the
cron job that cleans up maint'd hosts.


solarkennedy:
If you are confident that mesos (not k8s) doesn't need this, I would rather we keep this in place for now, and then just remove this code all together.

mattmb:
Really I want the flag so that we can try it out in one cluster to confirm everything is okay. My plan is to come back and remove the reservation code completely once we patch all the mesos clusters to ignore maint'd offers.

PR #1923, opened 2018-08-01
paasta local-run support for tron actions

solarkennedy:


paasta_tools/utils.py
```diff
@@ -771,14 +772,14 @@ def stringify_constraints(uscs: Optional[List[UnstringifiedConstraint]]) -> List
 @time_cache(ttl=60)
 def validate_service_instance(service: str, instance: str, cluster: str, soa_dir: str) -> str:
     for instance_type in INSTANCE_TYPES:
-        services = get_services_for_cluster(cluster=cluster, instance_type=instance_type, soa_dir=soa_dir)
-        if (service, instance) in services:
+        sis = get_services_for_cluster(cluster=cluster, instance_type=instance_type, soa_dir=soa_dir)
+        if (service, instance) in sis:
             return instance_type
     else:
+        suggestions = suggest_possibilities(word=instance, possibilities=[si[1] for si in sis])
```
qui: These will only be for the last instance type.

paasta_tools/utils.py
```diff
@@ -2255,16 +2256,27 @@ def get_service_instance_list_no_cache(
...
-            else:
-                instance_list.append((service, instance))
+        if srv_instance_type == 'tron':
+            tron_config_content = service_configuration_lib.read_extra_service_information(
+                service,
+                conf_file,
+                soa_dir=soa_dir,
+            )
+            for job in tron_config_content.get('jobs', []) or []:
+                for action in job['actions']:
```
qui: Hm, this doesn't include cleanup actions... Same with load_tron_instance_config.
They're annoying to deal with, maybe worth a refactor of how that works later.

paasta_tools/utils.py
solarkennedy: Hah. Hmm. Who would ever need anything other than suggestions for tron? :)

paasta_tools/cli/cmds/local_run.py
```diff
@@ -866,17 +872,25 @@ def configure_and_run_docker_container(
...
-                PaastaColors.red(
-                    "Error: No sha has been marked for deployment for the %s deploy group.\n"
-                    "Please ensure this service has either run through a jenkins pipeline "
-                    "or paasta mark-for-deployment has been run for %s\n" % (
-                        instance_config.get_deploy_group(), service,
+            if instance_config.get_deploy_group() is None:
+                paasta_print(
+                    PaastaColors.red(
+                        f"Error: {service}.{instance} has no 'deploy_group' set. Please set one so "
+                        "the proper version of the code can be downloaded for use.",
```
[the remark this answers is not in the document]
solarkennedy: yes

solarkennedy:
Ok, I think this is ready for real review @qui. 

The things I'm most concerned about are:
* service=None in the tronjob object. This seems "correct" in the sense that there is no "service" for things like the MASTER.yaml, but it feels like a place where bugs could come in.
* deploy_group = None. Before it was `''`, but I like `None` better than an empty string, but then again it shows up in the `env` as a the string None, which I always find hilarious. I'm not sure, sometimes there really is no deploy_group?
* `paasta_setup_tron_namespace` breaking. I manually ran it to try to catch things. Would you like me to add a `--dry-run` option and make it spit out the yaml it *would* POST and make an itest for it?

paasta_tools/tron_tools.py
```diff
@@ -423,6 +430,22 @@ def format_tron_job_dict(job_config, cluster_fqdn_format, default_paasta_cluster
     return {key: val for key, val in result.items() if val is not None}
 
 
+def load_tron_instance_config(service, instance, cluster, load_deployments=True, soa_dir=DEFAULT_SOA_DIR):
+    jobs, _ = load_tron_service_config(
+        service=service,
+        tron_cluster=cluster,
```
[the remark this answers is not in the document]
solarkennedy: Well. It "can", if we want it to.

When using `local-run`, I'm expecting people to type in `-c prod`. 

I may not exactly understand your concern though.  It is true that when users say `-c prod`, they will get actions that have `cluster: other`. At this point I think that is "ok".

paasta_tools/tron_tools.py
```diff
@@ -423,22 +435,51 @@ def format_tron_job_dict(job_config, cluster_fqdn_format, default_paasta_cluster
...
+        load_deployments=load_deployments,
+        soa_dir=soa_dir,
+    )
+    requested_job, requested_action = instance.split('.')
+    for job in jobs:
+        if job.get_name() == requested_job:
+            for action in job.get_actions(default_paasta_cluster=cluster):
```
qui: My concern about this is that if `cluster` is a Tron cluster, e.g. `prod`, and the action doesn't have a cluster configured, Tron will think that it needs to schedule the task on `paasta-prod`, instead of something like `paasta-norcal-prod` which is correct.

It doesn't matter for local-run, since the task isn't scheduled on Mesos, but it does matter if we want to use load_tron_instance_config for anything that affects Tron configuration.

solarkennedy: Hmm. You are right, this is a bit concerning and may have some lasting oddities if we go with this approach.

solarkennedy: I think this is a very valid concern and could have repercussions to our design in the future. Or just plain bugs where `cluster != tron_cluster`, and user-confusion, etc. Let's talk about it IRL and see what we can do short and long term.

solarkennedy:
> Why did you add service to TronJobConfig in the first place? I don't mind having it, but I don't see it being used anywhere.

It is used 
https://github.com/Yelp/paasta/pull/1923/files#diff-184161efc7cf921cdbcadbdef891bb50R230
so that we can load from a service dir without having an explicit `service: foo` in the job description.

It also makes it more natural here:
https://github.com/Yelp/paasta/pull/1923/files#diff-184161efc7cf921cdbcadbdef891bb50R482

Where the *caller* is specifying the service to load, seemed like the natural thing (injecting the service name into the loader) to do.

> Happy with deploy_group as either None or empty string.
Cool.

> I think --dry-run is a good idea for paasta_setup_tron_namespace, I've done that by commenting out/printing many times when testing various configuration changes.

I'll do that in a second PR.

> I see that cleanup actions still aren't included - if you decided against that, that's fine, but I would like to know the decision either way.
Sorry, I didn't mean to ignore this question. Yea cleanup_action is a job-level thing, I think it is OK to skip them in the context of local-run where individual actions are executed.

solarkennedy:
I've split this up into smaller PRs.
Also we might re-think how we do 'tron clusters' in general.

PR #1927, opened 2018-08-03
Use paasta service configs to create k8s healthchecks

keymone:


Rob-Johnson:
lg2m - does it work?

Rob-Johnson:
can you maybe write a test? I'm aware we're creating some test debt - even if this is 'poc' work

paasta_tools/kubernetes_tools.py
```diff
@@ -350,11 +351,48 @@ def get_resource_requirements(self) -> V1ResourceRequirements:
             },
         )
 
+    def get_liveness_probe(
```
mattmb: Can we replicate the logic here a little more closely? I'd like it to return `None` if it's not a smartstack service for example. https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_tools.py#L756

Happy to ship without it but if you feel like it we should be able to support the `cmd` and `tcp` modes too: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes

+1 on unit tests

PR #1929, opened 2018-08-07
Basic `paasta status` for k8s

vkhromov:


vkhromov:
```
$ USE_API_ENDPOINT=1 paasta_tools/cli/cli.py status -s robj-test-service -i main_k8s_stateful
instance: main_k8s_stateful
Git sha:    2e4d373f (desired)
State:      Configured - Desired state: Started
Kubernetes:   Healthy - up with (1/1) instances. Status: Running

service: robj-test-service
cluster: kubestage

$ USE_API_ENDPOINT=1 paasta_tools/cli/cli.py status -s robj-test-service -i main_k8s
instance: main_k8s
Git sha:    2e4d373f (desired)
State:      Configured - Desired state: Stopped
Kubernetes:   Healthy - up with (10/10) instances. Status: Running

service: robj-test-service
cluster: kubestage
```

paasta_tools/api/api_docs/swagger.json
```diff
@@ -602,6 +606,74 @@
         "bounce_method"
       ]
     },
+    "InstanceStatusKubernetes": {
+      "type": "object",
+      "properties": {
+        "desired_state": {
+          "type": "string",
+          "description": "Desired state of a service, for both Marathon and Chronos",
```
solarkennedy: copy/paste

paasta_tools/cli/cmds/status.py
```diff
@@ -447,14 +609,16 @@ def apply_args_filters(args):
...
+    args,
+) -> int:
     """Print the status of a Yelp service running on PaaSTA.
     :param args: argparse.Namespace obj created from sys.args by cli"""
     soa_dir = args.soa_dir
     system_paasta_config = load_system_paasta_config()
 
     if 'USE_API_ENDPOINT' in os.environ:
-        use_api_endpoint = strtobool(os.environ.get('USE_API_ENDPOINT'))
+        use_api_endpoint = strtobool(os.environ['USE_API_ENDPOINT'])
```
solarkennedy: I would be OK if you "forced" paasta status to use the API for k8s, no questions asked.

paasta_tools/api/api.py
```diff
@@ -118,6 +119,8 @@ def setup_paasta_api():
         cached=False,
     )
 
+    settings.kubernetes_clients = kubernetes_tools.get_kubernetes_clients()
```
mattmb: Can this fail? Maybe add a try except as we can't be sure the config etc exists?

paasta_tools/kubernetes_tools.py
```diff
@@ -649,13 +650,22 @@ def get_kubernetes_services_running_here_for_nerve(
...
-class KubeClient():
+class KubeClient(object):
     def __init__(self) -> None:
         kube_config.load_kube_config(config_file='/etc/kubernetes/admin.conf')
         self.deployments = kube_client.AppsV1Api()
         self.core = kube_client.CoreV1Api()
 
 
+class KubeClients(object):
+    def get_current_client_for_service(self, job_config: KubernetesDeploymentConfig) -> KubeClient:
```
mattmb: I think it's better to skip doing this. For Marathon we have logic to get multiple clients because we shard by service instance across multiple marathons. Hopefully we won't be doing the same with k8s anytime soon...

paasta_tools/cli/cmds/status.py
```diff
@@ -221,10 +268,110 @@ def paasta_status_on_api_endpoint(cluster, service, instance, system_paasta_conf
...
+    )
+    desired_state = desired_state_human(
+        kubernetes_status.desired_state,
+        kubernetes_status.expected_instance_count,
+    )
+    paasta_print(f"State:      {bouncing_status} - Desired state: {desired_state}")
+
+    status = KubernetesDeployStatus.fromstring(kubernetes_status.deploy_status)
+    if status != KubernetesDeployStatus.NotRunning:
+        if status == KubernetesDeployStatus.Delayed:
```
mattmb: This `Delayed` state isn't defined for k8s right? I wonder if we can find a way to map it to the equivalent thing in k8s...

I'd be tempted to remove the logic otherwise. Anyway, don't consider this a blocker we can leave it in if you don't think it'll do any harm and iterate later.

paasta_tools/cli/cmds/status.py
mattmb: +1, that was my thought too. It won't work any other way so no need to make the user specify it.

vkhromov: Please note that this diff uses the single `paasta status` command for both Marathon/Mesos and k8s setups.  For Marathon/Mesos `paasta status` can use either ssh or REST API depending on `USE_API_ENDPOINT` environment variable value, for k8s setup `paasta status ` can use only REST API (when `USE_API_ENDPOINT` is true) and just doesn't work otherwise.
We can detect the current setup only after receiving a response from the REST API.
Said so, we cannot force `paasta status` to use REST API for k8s without forcing it for Marathon/Mesos as well.
Also note, that the REST API is inferior comparing to ssh, as it i.e. doesn't support `-v`.
So, either we use RESP API by default for both Marathon/Mesos and k8s setups (probably after extending REST API to support `-v` and other missing thing), or we introduce something like `paasta status_k8s` specifically for k8s.
What do you think?

paasta_tools/cli/cmds/status.py
vkhromov: This is actually the second thing I wanted to discuss (beside fixing `-v` for both Marathon/Mesos and k8s setups).
How many states do we need for the k8s setup?
Probably it's possible to deduce `Waiting`/`Delayed`/`NotRunning`/`Stopped`/etc from pods statuses and/or error messages, the question is: do we actually need them?

vkhromov:
```
$ USE_API_ENDPOINT=1 paasta_tools/cli/cli.py status -s dsc-test-service -i main_k8s_stateful
instance: main_k8s_stateful
Git sha:    1f8cfbc2 (desired)
State:      Configured - Desired state: Started
Kubernetes:   Healthy - up with (1/1) instances. Status: Running

service: dsc-test-service
cluster: kubestage

$ USE_API_ENDPOINT=1 paasta_tools/cli/cli.py status -s dsc-test-service -i main_k8s
instance: main_k8s
Git sha:    1f8cfbc2 (desired)
State:      Configured - Desired state: Started
Kubernetes:   Healthy - up with (10/10) instances. Status: Running

service: dsc-test-service
cluster: kubestage

$ USE_API_ENDPOINT=1 paasta_tools/cli/cli.py status -c kubestage -s dsc-test-service
instance: main_k8s
Git sha:    1f8cfbc2 (desired)
State:      Configured - Desired state: Started
Kubernetes:   Healthy - up with (10/10) instances. Status: Running
instance: main_k8s_stateful
Git sha:    1f8cfbc2 (desired)
State:      Configured - Desired state: Started
Kubernetes:   Healthy - up with (1/1) instances. Status: Running

service: dsc-test-service
cluster: kubestage
```

PR #1930, opened 2018-08-07
Update check_registered_slaves_aws.py ignore recently created asgs and sfrs

ddelnano:
See OPS-13784 for details.

## Todo
- [x] Build a deb and deploy to stagef/g and verified I did not introduce any regressions with the script https://fluffy.yelpcorp.com/i/QZDhq9XbSL4jZ11z724GhVbwNDC2Rv97.html.  Works as long as I update /etc/paasta/monitoring.json with a check_registered_slave_threshold key.
- [x] Verified that `aws autoscaling describe-auto-scaling-groups --group-name` gives me output unit tests expect https://fluffy.yelpcorp.com/i/Xb526BdnKq03q5TRGlfv40Sd5bCVT6Cm.html
- [x] Verified that  `aws ec2 describe-spot-fleet-requests` gives me the output my unit tests expect  https://fluffy.yelpcorp.com/i/GHnrsHK3n3DjxcNFcPJ4zvKB27Tp4QLn.html. 

I noticed that check_registered_slaves_aws.py looks at /etc/paasta/cluster_autoscaling/{s3_bucket}/{cluster}.json files.  However, I was unable to find any spot fleets in any of those files.  Is that intended?

Unfortunately I can't prove this fixes OPS-13784 unless it stays on stage for a while.  My original plan for testing was going to be build a deb manually, install it on the hosts I care, and disable puppet.  However, cep1070 is going to explicitly ignore hosts with puppet disabled.  I could put my manually built package on the apt repo so that I can install it through puppet but would rather have a jenkins job do that for me rather than me copying the deb.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_cluster_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_cluster_lib.py
@@ -97,9 +97,6 @@ DEFAULT_DRAIN_TIMEOUT = 600  # seconds
 
 AWS_SPOT_MODIFY_TIMEOUT = 30
 MISSING_SLAVE_PANIC_THRESHOLD = .3
-# Age threshold in seconds that should be met before an asg or sfr should
-# exceed before being checked for slave registration.
-CHECK_REGISTERED_SLAVE_THRESHOLD = 3600
 
 log = logging.getLogger(__name__)
 log.addHandler(logging.NullHandler())
@@ -153,7 +150,9 @@ class ClusterAutoscaler(object):
         self.enable_maintenance_reservation = enable_maintenance_reservation
 
         self.log.info('Initialized with utilization error %s' % self.utilization_error)
-        self.setup_metrics()
+        config = load_system_paasta_config()
+        self.slave_newness_threshold = config.get_monitoring_config().get('check_registered_slave_threshold')
+        self.setup_metrics(config)
 
     @property
     def log(self) -> logging.Logger:
@@ -161,10 +160,9 @@ class ClusterAutoscaler(object):
         name = '.'.join([__name__, self.__class__.__name__, resource_id])
         return logging.getLogger(name)
 
-    def setup_metrics(self) -> None:
+    def setup_metrics(self, config: SystemPaastaConfig) -> None:
         if not self.enable_metrics:
             return None
-        config = load_system_paasta_config()
         dims = {
             'paasta_cluster': config.get_cluster(),
             'region': self.resource.get('region', 'unknown'),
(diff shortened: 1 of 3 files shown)


paasta_tools/autoscaling/autoscaling_cluster_lib.py
```diff
@@ -96,6 +97,9 @@
 
 AWS_SPOT_MODIFY_TIMEOUT = 30
 MISSING_SLAVE_PANIC_THRESHOLD = .3
+# Age threshold in seconds that should be met before an asg or sfr should
+# exceed before being checked for slave registration.
+CHECK_REGISTERED_SLAVE_THRESHOLD = 3600
```
EvanKrall: This should probably be defined as a getter on SystemPaastaConfig instead of a constant, so that when we want to tweak this we don't need to push a new version of Paasta.

ddelnano: Good call.  I thought about this problem but didn't know where the best place to make it configurable was.

mattmb:
Hmm, we're actually moving away from those S3 config files as we move to clusterman. Now that AWS supports tagging of SFRs (or more correctly passing a tag specification to the instances). So clusterman is just using the AWS API to fetch all the SFRs and then filtering them based on tags.

ddelnano:
@mattmb so with the move to clusterman this will be deprecated?  Also not sure if either of you say this CR but can you look at https://reviewboard.yelpcorp.com/r/327638/ as well?

mattmb:
Well I think we'll have to refactor it to work the same way without the S3 files. But for now, this will work so I say ship.

ddelnano:
@EvanKrall I started on refactoring the threshold to exist in /etc/paasta/monitoring.json.

As for 
> I wonder if we shouldn't be looking at each box individually, instead of the ASG/SFR? This check would still fire if we drastically scaled up an existing ASG/SFR, or if the SFR took a long time to fulfill the capacity request (i.e. if we're near our AWS instance limits)

I think that's outside the scope of my refactor for the ticket I'm working on.  Sound fair?

paasta_tools/autoscaling/autoscaling_cluster_lib.py
ddelnano: My newest commit uses a new key in /etc/paasta/monitoring.json so we can deploy a tweak via puppet.  See https://reviewboard.yelpcorp.com/r/329219 and my latest commit.

solarkennedy:
@ddelnano do you need any more help on this? Would you like me to press the green button?

PR #1932, opened 2018-08-08
Added dry-run support to paasta_setup_tron_namespace

solarkennedy:
Sorry this is so large. I kinda need both of these changes in place to be able to run `paasta_setup_tron_namespace` in itest mode.

Specifically:
1. I inject more things into `create_complete_config` instead of it loading more stuff. It still loads some things, but in general I think we should inject more and "tell don't ask"
2. I do more `cluster` => `tron_cluster` versus `default_paasta_cluster` stuff. I kinda which I did this change first before I did local-run, because I understand the nuances better.
3. Dry'd out the behave stuff a bit
4. Added the behave test to run this thing on fake soa-configs

I can split this up if you want, just say the word. Also I can postpone on the `tron_cluster` stuff if we think we are going to change that concept.

### DIFF
--- a/general_itests/fake_soa_configs_tron/fake_simple_service/tron-test-cluster.yaml
+++ b/general_itests/fake_soa_configs_tron/fake_simple_service/tron-test-cluster.yaml
@@ -1,0 +1,7 @@
+jobs:
+- name: sample_tron_job
+  actions:
+    - name: action1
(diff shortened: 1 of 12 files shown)


solarkennedy:
I've split this up to make it easier to review.

PR #1934, opened 2018-08-09
Inject tron_cluser and default_paasta_cluser into generate_complete_config for tron

solarkennedy:
This does more dependency injection and makes it easier for dry-run.


solarkennedy:
This is obsoleted by #1938 which removes the {tron_cluster} concept.

PR #1936, opened 2018-08-09
Make suggestions when an instance doesn't exist

solarkennedy:


paasta_tools/utils.py
```diff
@@ -770,15 +771,16 @@ def stringify_constraints(uscs: Optional[List[UnstringifiedConstraint]]) -> List
 
 @time_cache(ttl=60)
 def validate_service_instance(service: str, instance: str, cluster: str, soa_dir: str) -> str:
+    suggestions: List[str] = []
     for instance_type in INSTANCE_TYPES:
-        services = get_services_for_cluster(cluster=cluster, instance_type=instance_type, soa_dir=soa_dir)
-        if (service, instance) in services:
+        sis = get_services_for_cluster(cluster=cluster, instance_type=instance_type, soa_dir=soa_dir)
+        if (service, instance) in sis:
```
mattmb: `s/sis/service_instances/` just because it took me a few seconds to parse what the variable name meant.

solarkennedy: can do

PR #1937, opened 2018-08-09
Handle a None deploy_group properly

solarkennedy:
This is prerequisite to doing local-run tron.

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -866,17 +866,27 @@ def configure_and_run_docker_container(
         try:
             docker_url = instance_config.get_docker_url()
         except NoDockerImageError:
-            paasta_print(
-                PaastaColors.red(
-                    "Error: No sha has been marked for deployment for the %s deploy group.\n"
-                    "Please ensure this service has either run through a jenkins pipeline "
-                    "or paasta mark-for-deployment has been run for %s\n" % (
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -866,17 +866,27 @@ def configure_and_run_docker_container(
...
-                PaastaColors.red(
-                    "Error: No sha has been marked for deployment for the %s deploy group.\n"
-                    "Please ensure this service has either run through a jenkins pipeline "
-                    "or paasta mark-for-deployment has been run for %s\n" % (
-                        instance_config.get_deploy_group(), service,
+            if instance_config.get_deploy_group() is None:
+                paasta_print(
+                    PaastaColors.red(
+                        f"Error: {service}.{instance} has no 'deploy_group' set. Please set one so "
+                        "the proper version of the code can be downloaded for use.",
```
mattmb: Wording sounds a bit weird to me. Maybe "Please set one so PaaSTA knows which image to run for this service" is better?

mattmb:
Is it worth improving the errors anywhere else we have that Exception?

paasta_tools/cli/cmds/local_run.py
solarkennedy: fixed

solarkennedy:
chronos-rerun is the only other "user-facing" command where this could happen, but it is much less likely. Meh.

PR #1938, opened 2018-08-09
Added the tron instance config loader, removed the concept of a tron_cluster

solarkennedy:


paasta_tools/utils.py
```diff
@@ -2232,6 +2232,35 @@ def list_all_instances_for_service(
     return instances
 
 
+def get_tron_instance_list_from_yaml(service: str, conf_file: str, soa_dir: str) -> Collection[Tuple[str, str]]:
+    instance_list = []
+    tron_config_content = service_configuration_lib.read_extra_service_information(
```
qui: Hm, so this only works for Tron files that are in service directories?

solarkennedy: Yes! I did that intentionally. I think it is "OK" to require users to have their files in place in the "paasta format" (services/*) in order for paasta commands to understand them.

PR #1942, opened 2018-08-10
using tronfig validate with master

chlgit:
With master config file, we would be able to validate more stuff, such as valid_node, correctly rendered command, and so on.


paasta_tools/tron_tools.py
```diff
@@ -511,13 +511,15 @@ def validate_complete_config(service: str, cluster: str, soa_dir: str=DEFAULT_SO
     ]
     complete_config = yaml.dump(other_config, Dumper=Dumper)
 
+    master_config_path = os.path.join(os.path.abspath(soa_dir), 'tron', cluster, 'MASTER.yaml')
```
qui: Might as well use the MASTER_NAMESPACE variable here.

chlgit:
I can confirm it works with MASTER file as well.

PR #1943, opened 2018-08-13
Attempt to fix build

mattmb:
tox-pip-extensions is complaining and there is a new version of tox 3
days ago. Pinning back to see if it fixes the build.


keymone:
🚢

PR #1944, opened 2018-08-17
make spark-run send spark job resource requirements via clusterman me…

stug:
…trics

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -3,7 +3,10 @@ import json
 import os
 import socket
 import sys
+import time
 
+import staticconf
+from boto3.exceptions import Boto3Error
 from botocore.session import Session
 from ruamel.yaml import YAML
 
@@ -27,6 +30,14 @@ from paasta_tools.utils import paasta_print
 from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import PaastaNotConfiguredError
 from paasta_tools.utils import SystemPaastaConfig
+try:
+    import clusterman_metrics
+    CLUSTERMAN_YAML_FILE_PATH = '/nail/srv/configs/clusterman.yaml'
+    staticconf.YamlConfiguration('/nail/srv/configs/clusterman_metrics.yaml', namespace='clusterman_metrics')
+except ImportError:
+    # our cluster autoscaler is not currently open source, sorry!
+    clusterman_metrics = None
+    CLUSTERMAN_YAML_FILE_PATH = None
 
 AWS_CREDENTIALS_DIR = '/etc/boto_cfg/'
 DEFAULT_SERVICE = 'spark'
@@ -162,6 +173,14 @@ def add_subparser(subparsers):
         'For example, --spark-args "spark.mesos.constraints=pool:default\;instance_type:m4.10xlarge '
         'spark.executor.cores=4".',
     )
+
+    if clusterman_metrics:
+        list_parser.add_argument(
+            '--suppress-clusterman-metrics-errors',
+            help='Continue even if sending resource requirements to Clusterman fails. This may result in the job '
+            'failing to acquire resources.',
+            action='store_true',
+        )
 
     list_parser.add_argument(
         '-j', '--jars',
@@ -376,7 +395,7 @@ def load_aws_credentials_from_yaml(yaml_file_path):
         )
 
 
-def get_spark_conf_str(
+def get_spark_config(
     args,
     container_name,
     spark_ui_port,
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -480,10 +490,46 @@ def get_spark_conf_str(
...
+            writer.send((metric_key, int(time.time()), desired_quantity))
+
+
+def get_aws_region_for_paasta_cluster(paasta_cluster):
+    with open(CLUSTERMAN_YAML_FILE_PATH, 'r') as clusterman_yaml_file:
+        clusterman_yaml = YAML().load(clusterman_yaml_file.read())
+        return clusterman_yaml['mesos_clusters'][paasta_cluster]['aws_region']
+
+
+def spark_memory_to_megabytes(spark_memory_string):
```
huadongliu: in https://spark.apache.org/docs/latest/configuration.html, spark.executor.memory is
"Amount of memory to use per executor process, in MiB unless otherwise specified. (e.g. 2g, 8g)."

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -480,10 +490,46 @@ def get_spark_conf_str(
...
+
+
+def emit_resource_requirements(spark_config_dict, paasta_cluster):
+    num_executors = int(spark_config_dict['spark.cores.max']) / int(spark_config_dict['spark.executor.cores'])
+    memory_per_executor = spark_memory_to_megabytes(spark_config_dict['spark.executor.memory'])
+
+    desired_resources = {
+        'cpus': int(spark_config_dict['spark.cores.max']),
+        'mem': memory_per_executor * num_executors,
+        'disk': memory_per_executor * num_executors,  # rough guess since spark does not collect this information
```
drmorr0: what if you just set this to None instead of making a guess?  I worry that this might make bin-packing harder by adding more constraints, although maybe that's a good thing?  Also maybe it doesn't matter since we're not really disk-constrained... but on the other hand most of our services underestimate the amount of disk space they use... but on the gripping hand, we can't put hard limits on the amount of disk a service/job uses anyways, so....... *shrug*

Your call what to do here.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -480,10 +490,46 @@ def get_spark_conf_str(
...
+    }
+    dimensions = {'spark_framework_name': spark_config_dict['spark.app.name']}
+
+    paasta_print('Sending resource request metrics to Clusterman')
+    aws_region = get_aws_region_for_paasta_cluster(paasta_cluster)
+    metrics_client = clusterman_metrics.ClustermanMetricsBotoClient(region_name=aws_region)
+
+    with metrics_client.get_writer(clusterman_metrics.APP_METRICS) as writer:
```
drmorr0: There's a possibility this call might change a little bit; since metrics get duplicated to SFX, and SFX doesn't like timeseries with unbounded dimension, we either need to filter metrics emissions that have unbounded dimension or reduce the length of time we keep this data in SFX.  I want to chat with the metrics team before making a decision here one way or the other.

yelp_package/extra_requirements_yelp.txt
```diff
@@ -1,4 +1,5 @@
 --extra-index-url=https://pypi.yelpcorp.com/simple
+clusterman_metrics==1.2.7
```
drmorr0: this will need to bump to the new version of clusterman_metrics once I release it

paasta_tools/cli/cmds/spark_run.py
stug: ok, I think this and the fact that I'll need to bump clusterman_metrics means I should wait to merge this

paasta_tools/cli/cmds/spark_run.py
stug: We talked offline, and we constrain spark.executor.memory to the format "\dg" so this is fine

paasta_tools/cli/cmds/spark_run.py
stug: Yeah I'm not super sure here.  To add a fourth hand (I'm goro?), in some cases that I don't fully understand, Spark will spill a bunch of data to disk, and I _think_ we've seen some jobs eat all the disk on a box before.  Anyway, it seems like it's nice to annotate that these jobs can use some disk, so I guess I'm going to leave it for now?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -480,10 +490,46 @@ def get_spark_conf_str(
...
+
+
+def create_spark_config_str(spark_config_dict):
+    spark_config_entries = list()
+    for opt, val in spark_config_dict.items():
+        spark_config_entries.append(f'--conf {opt}={val}')
+    return ' '.join(spark_config_entries)
+
+
+def emit_resource_requirements(spark_config_dict, paasta_cluster):
```
huadongliu: Do we need to deal with exceptions here, e.g. writing to aws failure? I think spark-run can proceed with a message that cluster may not scale up?

stug: I'm a little unsure what the right move is.  In cases where it's a user using spark-run interactively, displaying a message and continuing may be fine -- maybe the user decides to cancel the run, maybe they let it hobble along.  

If it's being run as a schedule batch, is it still ok?  The batch may _never_ be able to get the resources it needs.  It might be better in that case to just fail and let tron (or whatever) retry if desired.  It also seems preferable to take the same approach in both cases (since otherwise we'd need some way to determine whether we're being run interactively or what).

I think is situation is a little trickier because of the way we're currently thinking about the Spark signal.  We aren't planning on running a "current resources" signal, so if a spark run fails to notify about its resources, there's no mechanism trying to make sure the "orphaned" run can keep its resources, let alone get more resources.

Seems like the options are:
1. Fail and let the user/scheduler retry if desired
2. Warn and go ahead, keeping the current signal implementation.  The job may never be able to scale up all the way (seems like it will, at best, be able to use the cluster
3. Warn and go ahead, but use a current resources signal and change the spark signal so that its job is only to report additional resources needed by spark jobs as they scale up
4. Warn and go ahead, and modify the spark signal so that it looks for orphaned spark jobs (jobs whose framework name looks like a spark job but have no resource request metrics) and reports their resources, more or less making it act as a current resources signal for _only_ oprhaned spark jobs.

I think 1 is the best for now and that we may want to consider moving to 3 if it becomes problematic.  Maybe @drmorr0 has thoughts?

PR #1948, opened 2018-08-20
Add paasta deploy events using yelp_meteorite.

rahulravindran0108:
**Summary**
This mimics the internal functionality for event_detector which was
recently decommissioned. Emits an event name-spaced `deploy.paasta`
and along with this sends cluster, instance and service as a dimension.
Open to instrument more dimensions if need be.

### DIFF
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -408,7 +408,7 @@ def do_bounce(
                         'cluster': cluster,
                         'instance': instance,
                         'service': service,
-                    }
+                    },
                 )
         return None
 
--- a/paasta_tools/setup_marathon_job.py
+++ b/paasta_tools/setup_marathon_job.py
@@ -401,13 +401,12 @@ def do_bounce(
             )
 
             if yelp_meteorite:
-                # Emit deploy.paasta event
                 yelp_meteorite.events.emit_event(
                     'deploy.paasta',
                     dimensions={
-                        'cluster': cluster,
-                        'instance': instance,
-                        'service': service,
+                        'paasta_cluster': cluster,
+                        'paasta_instance': instance,
+                        'paasta_service': service,
                     },
                 )
         return None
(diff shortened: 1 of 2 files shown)


paasta_tools/setup_marathon_job.py
```diff
@@ -392,6 +399,17 @@ def log_bounce_action(line: str, level: str='debug') -> None:
                 ),
                 level='event',
             )
+
+            if yelp_meteorite:
+                # Emit deploy.paasta event
+                yelp_meteorite.events.emit_event(
+                    'deploy.paasta',
+                    dimensions={
+                        'cluster': cluster,
```
solarkennedy: Here is some bad news: we haven't been consistent in our dimensions with our metrics.
https://github.com/Yelp/paasta/blob/master/paasta_tools/contrib/emit_allocated_cpu_metrics.py#L26

What do you think about getting us off onto the "right" foot and moving us twoards a consistent `paasta_cluster`, `paasta_service` and `paasta_instance` ?

paasta_tools/setup_marathon_job.py
```diff
@@ -392,6 +399,17 @@ def log_bounce_action(line: str, level: str='debug') -> None:
                 ),
                 level='event',
             )
+
+            if yelp_meteorite:
+                # Emit deploy.paasta event
```
solarkennedy: This comment doesn't really say anything that the next line doesn't say

paasta_tools/setup_marathon_job.py
rahulravindran0108: I'll go ahead and rename these dimensions to be consistent with others.

paasta_tools/setup_marathon_job.py
rahulravindran0108: Agreed. Will remove this

rahulravindran0108:
@solarkennedy addressed inconsistencies with dimensions. This is ready for review.

PR #1952, opened 2018-08-23
Add support for StatefulSets

mattmb:
This gives the k8s logic the ability to manage StatefulSets. We manage
them the same way as Deployments. The distinguishing factor is that they
support dynamic EBS volumes that follow the Pods. The new yelpsoa syntax
looks like:
```
persistent_volumes:
  - container_path: /mnt/whatever
    size: 10
    mode: RW
```

Note this is deliberately a bit MVP. The StorageClass has to be created
by hand and we don't enforce constraints on AZ yet so you must be
careful to create everything in 1 AZ.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -21,6 +21,7 @@ from typing import NamedTuple
 from typing import Optional
 from typing import Sequence
 from typing import Tuple
+from typing import Union
 
 import requests
 import service_configuration_lib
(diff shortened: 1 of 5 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -529,53 +575,106 @@ def format_kubernetes_app(self) -> V1Deployment:
...
-                metadata=V1ObjectMeta(
-                    name="{service}-{instance}".format(
-                        service=self.get_sanitised_service_name(),
-                        instance=self.get_sanitised_instance_name(),
+            if self.get_persistent_volumes():
+                complete_config = V1StatefulSet(
+                    api_version='apps/v1',
+                    kind='StatefulSet',
+                    metadata=V1ObjectMeta(
+                        name="{service}-{instance}".format(
```
keymone: can you reduce duplication a bit? `name`, `service_name`,  `labels`, `containers`, `volumes`, `replicas` are (as far as i can tell) identical between these two branches.

mattmb: 👍 good shout, let me take a crack

paasta_tools/kubernetes_tools.py
```diff
@@ -515,67 +535,94 @@ def get_desired_instances(self) -> int:
...
+        return [
+            V1PersistentVolumeClaim(
+                metadata={'name': self.get_persistent_volume_name(volume)},
+                spec=V1PersistentVolumeClaimSpec(
+                    # must be ReadWriteOnce for EBS
+                    access_modes=["ReadWriteOnce"],
+                    storage_class_name=self.get_storage_class_name(),
+                    resources=V1ResourceRequirements(
+                        requests={
+                            'storage': f"{volume['size']}Gi",
```
vkhromov: How is resource type `storage` going to work?
Also probably it's better to name it `storage-space` as planned at https://github.com/kubernetes/community/blob/master/contributors/design-proposals/scheduling/resources.md#future-storage-space

PR #1958, opened 2018-08-29
Autofill `kind` and `api_version` attributes

vkhromov:
Autofill `kind` and `api_version` attributes for k8s model classes based
on their names.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -13,7 +13,10 @@
 """
 """
 import copy
+import inspect
 import logging
+import re
+import sys
 from typing import Any
 from typing import Dict
 from typing import List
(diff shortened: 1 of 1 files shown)


vkhromov:
@mattmb,

Just to clarify
 * we're patching only classes we've imported into `kubernetes_tools` module
 * we're filling `kind` and `api_version` attributes according to class' name, it seems it's possible to deduce these attributes from the name.
 
Answering to your question:

>Will that regex always work if they come up with some new naming scheme

`TestKubernetesDeploymentConfig::test_format_kubernetes_app_dict` test particularly verifies `api_version` and `kind`, if k8s changes naming scheme, the test will be broken as a result.

> We have to be careful about changing anything that affects our config sha because it could trigger a "big bounce" where we restart every service. You can imagine we might update the k8s client lib and they change from v1beta to v1 or something and suddenly we re-create every app.

`k8s` cannot do that, since as far as we don't change imports from `V1Something` to `V${another_version}Something` there will be no change, even if `k8s` introduces `V1beta1Something` or `V2Something`.

On the other hand, we probably should ignore content of `api_version` and `kind` while calculating hash to prevent such scenario.  What do you think?

> I'd be okay with making the `api_version` configurable as a `system_paasta_config` thing. But I don't know how to do the kind. Is it really not available without a regex?

Why do you need two places to configure `api_version`?  We've already chosen which `api_version` to use by using corresponding `V${api_version}Something` classes, plus I'm not sure that a server accepts `V1Something` with `api_version` equals to `2` or vice versa.

vkhromov:
At the moment we have only two places with `api_version`, so we're probably fine without this change for now.
Let not merge it now.

PR #1956, opened 2018-08-28
Enable/Disable types of Slack notifications

jordanleex13:
https://jira.yelpcorp.com/browse/PAASTA-14607

Can specify the following types in a deploy.yaml file:
- notify_after_mark
- notify_after_good_deploy
- notify_after_auto_rollback
- notify_after_abort

PR #1959, opened 2018-08-29
Make Slack notifications more concise

chriskuehl:
The new Slack notifications are great at providing tons of information, but they can be pretty noisy as the typical notification takes up between 5 and 9 lines of chat (plus another 5-8 lines in many cases, when we have a mark-for-deployment followed immediately by a successful deployment notification):

I was thinking we could use Slack attachments or something to hide most of this information away in a collapsible summary, but I was surprised to not be able to find any way to do this.

Instead, what if we had a one-line summary in the main channel, and posted all the additional details in a thread followup? Here's how those same two messages look with this change:

Clicking to expand the thread has the rest of the information:

I haven't updated tests or anything yet, mostly just looking for feedback on the message format. Is this an improvement? Is there a better solution?

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -220,7 +220,7 @@ class SlackDeployNotifier(object):
     def get_url_message(self):
         build_url = os.environ.get('BUILD_URL')
         if build_url is not None:
-            message = f"(Jenkins: {build_url}/consoleFull)"
+            message = f"<{build_url}/consoleFull|Jenkins Job>"
         else:
             message = f"(Run by {getpass.getuser()} on {socket.getfqdn()})"
         return message
@@ -241,8 +241,8 @@ class SlackDeployNotifier(object):
                 return ""
             else:
                 slacky_authors = ", ".join([f"<@{a}>" for a in authors.split()])
-                log.debug(f"Pinging authors: {slacky_authors}")
-                return f"Pinging authors: {slacky_authors}"
+                log.debug(f"Authors: {slacky_authors}")
+                return f"Authors: {slacky_authors}"
         else:
             return f"(Could not get authors: {authors})"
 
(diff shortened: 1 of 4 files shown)


fhats:
I sort of like having all this detail up-front without having to dig through a thread, but I also agree that the current verbosity is a bummer.

A few ideas:

"Will start deploying soon!" seems like it just wastes a line. I reckon this was added to reduce confusion or be more "technically correct," as "marked for deployment" is not necessarily the same as "deploying"... but I think it's overkill. Let's just rm it.

That jenkins link is pretty long. Formatting that with display text that's human-friendly (and unlikely to wrap), perhaps like "Jenkins Build 454", should be possible using slack's full formatting mode, or by making the message an attachment and including it as a field, or the title URL.

I don't like the verbiage "Pinging authors". Just "Authors:" would be fine. It doesn't help your mission, just a personal beef I have :)

Lastly, we could remove the Jenkins URL from the deploy finished message? Since these messages reference SHAs, maybe we could let people rely on the search bar to find the start-of-deploy message if they really don't want to navigate to the jenkins index themselves?

Perhaps we could combine the above with this change to get the best of both worlds? Something like

**error_pages** - Marked **01f0deadbeef** for deployment on **stagef**
Authors: @fhats
[Jenkins Build 454](https://famous.dog)
(1 Reply)
]    Roll it back with this command etc.

chriskuehl:
@fhats thanks for the comments! Those are some good ideas, I can definitely make some of those changes.

I do think we should try to keep the initial message to a single line (for the standard, no-error message), or at least provide a compact mode that teams can opt into via soaconfigs. I find that even 3-4 lines dropped into the middle of a conversation (especially if it happens twice, like it often does now) can be pretty disruptive.

chriskuehl:
Another thought is that I suspect most teams don't want the mark-for-deployment notification, only the notification when it finishes (I know my team has expressed this sentiment). A follow-up might be to make that configurable by soaconfigs.

fhats:
Good points, yeah! +1 for configurability as long as the defaults are sane :)

I know you're a fan of a single line, but what if the jenkins link was moved "below the fold," leaving just the sha/cluster and author? I think 2 lines really isn't too bad and that way you'd still get the author notification, which I suspect is widely appreciated

chriskuehl:
Yeah, two lines is probably not so bad, that's what it was on IRC after all. I'm happy with that. @solarkennedy any thoughts before I start updating this PR?

chriskuehl:
One reason it's kind of nice to hide all the authors is that for one of our large monolithic services, we have 20+ people pushing at once, so the authors line is actually 4-5 lines itself. Probably not a huge deal though.

EvanKrall:
> Another thought is that I suspect most teams don't want the mark-for-deployment notification, only the notification when it finishes (I know my team has expressed this sentiment). A follow-up might be to make that configurable by soaconfigs.

This is done: https://github.com/Yelp/paasta/pull/1956

solarkennedy:
I think this looks great! I disagree with the config option. I think we can get a balance with a great default for everyone.

The only thing I can think of is splurging for the rollback command, but I think by now everyone knows that you can roll back (or click the reply to see the command)

I say ship!

chriskuehl:
Fixed tests and rebased, I think this is ready for another review / merge?

I changed the Jenkins URL to be clickable but kept the authors "under the fold" for now -- let me know if anyone has strong thoughts about changing this.

PR #1963, opened 2018-08-31
Ignore `kind` and `version` when calculating config hash

vkhromov:
We're using `kind` and `api_version` to calculate the config hash, any of these
attributes could be changed during migration to a new version of `k8s` client
library, so we'll restart all instances as a result of such change.

At the same time, these attributes are transport-only and are used only to
inform the `k8s` server endpoint about a class and a version of the class used
to generate JSON.  A change in these attributes without a change in other
attributes is no-op.

Thus, we better ignore changes in these attributes while calculating the config
hash.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -83,7 +83,7 @@ from paasta_tools.utils import VolumeWithMode
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


keymone:
would be great if you described why this change is needed for posterity.

vkhromov:
@keymore, please see the updated description for the pull request.

PR #1964, opened 2018-08-31
Remove cluster address from Tron actions

qui:
`mesos_address` is a MASTER-level setting and can't be set in actions anymore, so the tronfig we generate is invalid. (https://github.com/Yelp/Tron/pull/509)

If we take mesos_address out, we don't need the cluster_fqdn_format and default cluster anymore. The cluster name just comes from the filename (i.e. tron-norcal-prod) now.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -65,13 +65,6 @@ class TronConfig(dict):
         except KeyError:
             raise TronNotConfigured('Could not find name of Tron cluster in system Tron config')
 
-    def get_default_paasta_cluster(self):
-        """:returns The PaaSTA cluster to run actions on by default"""
-        try:
-            return self['default_paasta_cluster']
-        except KeyError:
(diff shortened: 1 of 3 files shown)


paasta_tools/tron_tools.py
```diff
@@ -268,32 +262,30 @@ def _get_action_config(self, action_dict, default_paasta_cluster):
         else:
             branch_dict = None
 
-        cluster = action_dict.get('cluster') or self.get_cluster() or default_paasta_cluster
```
solarkennedy: Hehe, I'm kinda glad to see this go.

paasta_tools/tron_tools.py
```diff
@@ -308,9 +300,8 @@ def check_monitoring(self) -> Tuple[bool, str]:
         return True, ''
 
     def check_actions(self) -> Tuple[bool, List[str]]:
-        cluster = "fake-cluster-for-validation"
```
solarkennedy: woot, glad to see this go too.

PR #1965, opened 2018-09-03
Upgrade pre-commit hooks

asottile:
Automated via:
- `pre-commit autoupdate`
- `pre-commit run --all-files`
- I then had to revert changing the `detect-secrets` version since it broke for some reason? didn't dig into this

Relevant changes:
- [`pyupgrade` learned how to rewrite `super()`](https://github.com/asottile/pyupgrade#super-calls)


solarkennedy:
Thanks!

PR #1966, opened 2018-09-03
add poddisruptionbudgets

Rob-Johnson:
manage poddisruptionbudgets alongside deployments.
see the commit messages for a more thorough explanation.

the other 'standout' change here is that I've added [hypothesis](https://hypothesis.readthedocs.io/) as a dev requirement to aid with tests (see test_max_unavailable for an example use). lmk if you're against this.


paasta_tools/paastakube/api_client.py
```diff
@@ -0,0 +1,684 @@
+"""
```
mattmb: Can you leave a comment around the bit you've fixed and a link to the GH issue(s)

Also worth adding the same comment to this line I think: `self.deployments = kube_client.AppsV1Api(ApiClient())`

paasta_tools/setup_kubernetes_job.py
```diff
@@ -162,19 +168,94 @@ def reconcile_kubernetes_deployment(
             kube_client=kube_client,
             application=formatted_application,
         )
+        ensure_pod_disruption_budget(
```
mattmb: What do you think about moving this into its own loop? I'm worried `reconcile_kubernetes_deployment` is already pretty big. But we could move `load_kubernetes_service_config_no_cache` up a level and have a `reconcile_pod_disruption_budgets` maybe?

vkhromov: Copy-paste, all these 3 `ensure_pod_disruption_budget` can be moved out of the `if` statement.

paasta_tools/paastakube/api_client.py
```diff
@@ -0,0 +1,684 @@
...
+from kubernetes.client.rest import ApiException
+from kubernetes.client.rest import RESTClientObject
+from six import integer_types
+from six import iteritems
+from six import text_type
+from six.moves.urllib.parse import quote
+# python 2 and python 3 compatibility library
+
+
+def set_disrupted_pods(self, disrupted_pods):
```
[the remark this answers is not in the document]
Rob-Johnson: hmmm, I'll look again. I was getting a a bit blinded figuring this out yesterday, you could be right.

paasta_tools/paastakube/api_client.py
```diff
@@ -0,0 +1,684 @@
...
+    self._disrupted_pods = disrupted_pods
+
+
+models.V1beta1PodDisruptionBudgetStatus.disrupted_pods = property(
+    fget=models.V1beta1PodDisruptionBudgetStatus.disrupted_pods,
```
vkhromov: Also, since `models.V1beta1PodDisruptionBudgetStatus.disrupted_pods` is a `property`, it can't be used as a getter here (that's why that "TypeError: 'property' object is not callable" error).  You can either introduce `get_disrupted_pods` function and use it here as a getter, or try to extract something callable fro the property.

paasta_tools/setup_kubernetes_job.py
vkhromov: `ensure_pod_disruption_budget` calls are still there.  Can you please either do something with them or left a comment explaining why this code need to be left unchanged?

Rob-Johnson: I've updated so there is only 1 - you were too fast :)

mattmb:
lgtm 👍 
looks like you have some mypy errors to fix btw

PR #1969, opened 2018-09-05
Deployd leader problems

mattmb:
* Fix a bug where the startup stalls and process doesn't exit on abdication
* Add some logging to help catch other leadership bugs

### DIFF
--- a/paasta_tools/deployd/leader.py
+++ b/paasta_tools/deployd/leader.py
@@ -1,4 +1,5 @@
 import logging
+import threading
 import time
 
 from kazoo.client import KazooState
@@ -40,8 +41,7 @@ class PaastaLeaderElection(Election):
                 reconnection_checker.start()
         elif state == KazooState.LOST:
             self.log.error("Leadership lost, quitting!")
-            self.control.put("ABORT")
-            self.client.stop()
+            self.suicide()
 
     def reconnection_listener(self):
         attempts = 0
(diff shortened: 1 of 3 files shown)


paasta_tools/deployd/leader.py
```diff
@@ -54,5 +54,14 @@ def reconnection_listener(self):
             time.sleep(5)
             attempts += 1
         self.log.error("Connection did not recover, abdicating!")
+        self.suicide()
+
+    def suicide(self):
```
Rob-Johnson: can we name this something else? ``def terminate`` should be ok (maybe even ``_terminate`` to indicate this isn't really a public api)

paasta_tools/deployd/master.py
```diff
@@ -286,9 +288,21 @@ def start_watchers(self):
...
+            if all([watcher.is_ready for watcher in self.watcher_threads]):
+                return
+            self.log.info("Sleeping and waiting for watchers to all start")
+            self.log.info([
+                {
+                    'name': watcher.__class__.__name__,
+                    'ready': watcher.is_ready,
+                } for watcher in self.watcher_threads
+            ])
+            time.sleep(5)
```
Rob-Johnson: this seems a long time? I'm kind of expecting these workers to start quickly - what could lead to them taking so long?

mattmb: Yeah, I wasn't quite sure what was reasonable. I think we have to get a watch on all the ZK nodes but I guess that is quick. Any reason not to be conservative though?

paasta_tools/deployd/master.py
```diff
@@ -286,9 +288,21 @@ def start_watchers(self):
...
-            self.log.debug("Sleeping and waiting for watchers to all start")
+        attempts = 0
+        while attempts < 120:
+            if all([watcher.is_ready for watcher in self.watcher_threads]):
+                return
+            self.log.info("Sleeping and waiting for watchers to all start")
+            self.log.info([
+                {
+                    'name': watcher.__class__.__name__,
+                    'ready': watcher.is_ready,
```
keymone: maybe lets only log watchers that aren't ready?

PR #1971, opened 2018-09-05
Make paasta-related dimensions consistent

solarkennedy:
This makes the dimensions we emit consistently named. This must be shipped and release with the other changes in order to minimize dashboard breakage.


thebostik:
I mentioned this to Kyle offline, my only concern is doing the "rip the bandaid" approach instead of a period of supporting both on the emission side while we update everything. But I'll be available Friday to help move any manual dashboards, etc. that aren't covered by the other CRs.

PR #1975, opened 2018-09-07
Change cpu_burst_pct to cpu_burst_add 

kawaiwanyelp:
### Description
- Currently, we enforce a 10x burst on cpu usage via the cpu_burst_pct parameter in PaaSTA configs, but this has led to jobs using too many cpus.
- I changed cpu_burst_pct (i.e. multiplicative) to cpu_burst_add (i.e. additive), which provides a lower and much safer cpu limit.
- Instead of a 10x default, we default to 1 additional cpu.
- I also changed related code in kubernetes and chronos tools, as well as the autoscaling lib (i.e. marathon) to use the new parameter instead.
- Relevant tests were changed as well.

### Testing done
- make test

### Note to reviewers
- We initially considered changing from 10x to 2x, but even 2x is too much for jobs that request more cpus. A +1 is more than enough for smaller jobs, and will not adversely affect larger jobs. If they need more, they can supply their own cpu_burst_add.
- [UPDATE] @EvanKrall asked why I changed some test numbers; the answer is: "There are checks to make sure boxes don't use more cpus than the limit (cpus + burst) allowed. Essentially, the tests we do explicitly use numbers that were good for multiplicative burst, but too high for additive burst. As such, if I left the tests as they were, they would break because the tests would see that more cpus are being used than the limit allows. So, I changed the tests to maintain their intent - testing the limit check."

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -422,7 +422,7 @@ def mesos_cpu_metrics_provider(
             if system_paasta_config.get_filter_bogus_mesos_cputime_enabled():
                 # It is unlikely that the cputime consumed by a task is greater than the CPU limits
                 # that we enforce. This is a bug in Mesos (tracked in PAASTA-13510)
-                max_cpu_allowed = 1 + marathon_service_config.get_cpu_burst_add()
+                max_cpu_allowed = 1 + marathon_service_config.get_cpu_burst_add() / marathon_service_config.get_cpus()
                 task_cpu_usage = cputime_delta / time_delta
 
                 if task_cpu_usage > (max_cpu_allowed * 1.1):
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -422,8 +422,9 @@ def mesos_cpu_metrics_provider(
             if system_paasta_config.get_filter_bogus_mesos_cputime_enabled():
                 # It is unlikely that the cputime consumed by a task is greater than the CPU limits
                 # that we enforce. This is a bug in Mesos (tracked in PAASTA-13510)
-                max_cpu_allowed = (100 + marathon_service_config.get_cpu_burst_pct()) / 100
+                max_cpu_allowed = 1 + marathon_service_config.get_cpu_burst_add()
                 task_cpu_usage = cputime_delta / time_delta
+
                 if task_cpu_usage > (max_cpu_allowed * 1.1):
                     log.warning(
                         'Ignoring potentially bogus cpu usage {} for task {}'.format(
--- a/paasta_tools/cli/schemas/adhoc_schema.json
+++ b/paasta_tools/cli/schemas/adhoc_schema.json
@@ -86,9 +86,9 @@
                 "maximum": "1000000",
                 "exclusiveMinimum": false
             },
-            "cpu_burst_pct": {
-                "type": "integer",
-                "minimum": 0,
+            "cpu_burst_add": {
+                "type": "float",
+                "minimum": 0.0,
                 "exclusiveMinimum": false
             },
             "extra_docker_args": {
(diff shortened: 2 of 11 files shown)


paasta_tools/cli/schemas/adhoc_schema.json
```diff
@@ -86,7 +86,7 @@
                 "maximum": "1000000",
                 "exclusiveMinimum": false
             },
-            "cpu_burst_pct": {
+            "cpu_burst_add": {
```
EvanKrall: As soon as this version of paasta-tools gets installed on the git server for yelpsoa-configs, any pushes that modify services that still have cpu_burst_pct will start failing.

I think that's fine - just make sure you have a yelpsoa-configs CR ready to merge that changes all our uses of cpu_burst_pct, and when you ship this change make sure to run puppet manually on the git server, then merge the change.

paasta_tools/utils.py
```diff
@@ -97,7 +97,7 @@
 DEFAULT_SYNAPSE_HAPROXY_URL_FORMAT = "http://{host:s}:{port:d}/;csv;norefresh"
 
 DEFAULT_CPU_PERIOD = 100000
-DEFAULT_CPU_BURST_PCT = 900
+DEFAULT_CPU_BURST_ADD = 1
```
EvanKrall: This will cause a big bounce - you'll probably want to work with someone in Europe to ship this off-hours.

tests/autoscaling/test_autoscaling_service_lib.py
```diff
@@ -320,10 +320,10 @@ def test_mesos_cpu_metrics_provider_filter_bogus_values_big_cpu_limit():
     |  inst  | prev cputime | elasped_time | cputime | limit |  norm_cputime  |      utilization      |
     +--------+--------------+--------------+---------+-------+----------------+-----------------------+
     | 1      |          126 |          600 |     480 |   2.1 | 480/(2.1-0.1)  | (240-126)/600 => 0.19 |
-    | 2      |            0 |          600 |    9600 |   2.1 | 9600/(2.1-0.1) | 4800/600 => 8.0       |
+    | 2      |            0 |          600 |    2400 |   2.1 | 2400/(2.1-0.1) | 1200/600 => 2.0       |
     | 3(bug) |            0 |          600 |  987654 |   2.1 | 9600/(2.1-0.1) | 493827/600 => 823     |
     | -      |            - |            - |       - |     - | -              | -                     |
-    | avg    |              |              |         |       |                | 4.095                 |
+    | avg    |              |              |         |       |                | 1.095                 |
```
EvanKrall: I'm not sure I understand why this (and the following tests) needed to change -- afaict all this logic should still work exactly as it did before.

kawaiwanyelp: There are checks to make sure boxes don't use more cpus than the limit (cpus + burst) allowed. Essentially, the tests we do explicitly use numbers that were good for multiplicative burst, but too high for additive burst. As such, if I left the tests as they were, they would break because the tests would see that more cpus are being used than the limit allows. So, I changed the tests to maintain their intent - testing the limit check.

paasta_tools/utils.py
kawaiwanyelp: Roger.

paasta_tools/cli/schemas/adhoc_schema.json
kawaiwanyelp: Cool.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -422,8 +422,9 @@ def results_or_None(fut):
             if system_paasta_config.get_filter_bogus_mesos_cputime_enabled():
                 # It is unlikely that the cputime consumed by a task is greater than the CPU limits
                 # that we enforce. This is a bug in Mesos (tracked in PAASTA-13510)
-                max_cpu_allowed = (100 + marathon_service_config.get_cpu_burst_pct()) / 100
+                max_cpu_allowed = 1 + marathon_service_config.get_cpu_burst_add()
```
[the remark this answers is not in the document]
kawaiwanyelp: Hm, perhaps it should be cpus + cpu_burst_add instead then? e.g. if cpus = 2 and cpu_burst_add = 3, then 1 + cpu_burst_add/cpus = 1 + 3/2 = 2.5, when the limit should be 2 + 3 = 5.

tests/autoscaling/test_autoscaling_service_lib.py
EvanKrall: Ah, gotcha.

PR #1976, opened 2018-09-10
pyupgrade learns python3 classes

asottile:


EvanKrall:
@asottile Thanks for continuing to automate making our code better!

PR #1979, opened 2018-09-13
adding clusterman metrics for paasta boost

drmorr0:
This is all we need in the PaaSTA codebase for paasta boost to work.  This will duplicate boost requests to zookeeper and dynamodb for now.

### DIFF
--- a/paasta_tools/autoscaling/cluster_boost.py
+++ b/paasta_tools/autoscaling/cluster_boost.py
@@ -15,6 +15,14 @@ from time import time as get_time
 
 from kazoo.client import KazooClient
 from kazoo.exceptions import NoNodeError
+try:
+    import clusterman_metrics
+    CLUSTERMAN_YAML_FILE_PATH = '/nail/srv/configs/clusterman.yaml'
+    staticconf.YamlConfiguration('/nail/srv/configs/clusterman_metrics.yaml', namespace='clusterman_metrics')
+except ImportError:
+    # our cluster autoscaler is not currently open source, sorry!
+    clusterman_metrics = None
+    CLUSTERMAN_YAML_FILE_PATH = None
 
 from paasta_tools.utils import ZookeeperPool
 
@@ -161,6 +169,13 @@ def set_boost_factor(
     zk_boost_path = get_zk_boost_path(region, pool)
     current_time = get_time()
     end_time = current_time + 60 * duration_minutes
+
+    if clusterman_metrics:
+        metrics_client = clusterman_metrics.ClustermanMetricsBotoClient(region_name=region, app_identifier='default')
+        with metrics_client.get_writer(clusterman_metrics.APP_METRICS) as writer:
+            writer.send(('boost_factor', current_time, factor))
+            if duration_minutes > 0:
+                writer.send(('boost_factor', end_time, 1.0))
 
     zk_end_time_path = zk_boost_path + '/end_time'
     zk_factor_path = zk_boost_path + '/factor'


paasta_tools/paasta_cluster_boost.py
```diff
@@ -26,6 +26,11 @@
 def parse_args():
     """Parses the command line arguments passed to this script"""
     parser = argparse.ArgumentParser()
+    parser.add_argument(
```
solarkennedy: Do we need this to be an arg at all?
This command is running on master, just lookup the cluster you are a member of. We'll never be calling this command from some remote place.

paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -13,8 +13,17 @@
 from datetime import datetime
 from time import time as get_time
 
+import staticconf
```
solarkennedy: I'm surprised this works. I don't see it in requirements.txt

Can we get away without it? We don't need all the fanciness that staticconf provides in this situation.

paasta_tools/cli/utils.py
```diff
@@ -626,6 +627,7 @@ def run_paasta_cluster_boost(
     else:
         verbose_flag = None
 
+    cluster_flag = f'--cluster {cluster}'
```
solarkennedy: I think we can get away without this

paasta_tools/autoscaling/cluster_boost.py
drmorr0: clusterman_metrics reads all of its configuration from srv-configs via staticconf, so I don't think we can get away without it; it's included as a dependency of clusterman_metrics.  (It's also already in requirements.txt, `PyStaticConfiguration==0.10.3`)

paasta_tools/paasta_cluster_boost.py
drmorr0: Yea, good call.  I'll take this out.

paasta_tools/autoscaling/cluster_boost.py
```diff
@@ -13,8 +13,17 @@
...
 from kazoo.client import KazooClient
 from kazoo.exceptions import NoNodeError
+try:
+    import clusterman_metrics
+    CLUSTERMAN_YAML_FILE_PATH = '/nail/srv/configs/clusterman.yaml'
```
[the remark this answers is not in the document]
drmorr0: This is basically required for this to work; I factored this out into a separate file so at least it's in one place (and we're not duplicating code between here and spark_run)

paasta_tools/cli/utils.py
drmorr0: Gone.

paasta_tools/clusterman.py
```diff
@@ -0,0 +1,17 @@
+import staticconf
```
solarkennedy: I'm still surprised this works, as we are not including it in any requirements.txt. It must be some transitive dependency.

paasta_tools/clusterman.py
```diff
@@ -0,0 +1,17 @@
...
+    try:
+        import clusterman_metrics
+        clusterman_yaml = CLUSTERMAN_YAML_FILE_PATH
+        staticconf.YamlConfiguration(CLUSTERMAN_METRICS_FILE_PATH, namespace='clusterman_metrics')
+    except ImportError:
+        # our cluster autoscaler is not currently open source, sorry!
+        clusterman_metrics = None
+        clusterman_yaml = None
+
+    return clusterman_metrics, clusterman_yaml
```
solarkennedy: Can we get away without returning the YAML at all? That would simplify this function.

drmorr0: @stug needs the yaml for the spark_run stuff, idk if there's a better way to do it.

stug: I need it in spark-run, but we could possibly put logic in spark-run so that it checks for the existence of clusterman_metrics and then uses the YAML file path (imported from this module) if it can use it (i.e. if clusterman_metrics was imported)?

PR #1980, opened 2018-09-14
`paasta metastatus` via HTTP API

vkhromov:
Add HTTP API to executes the code used to generate output of `paasta metastatus` on the HTTP API server side, and returns the output and the exit code for that.
Use that HTTP API to generate result for `paasta metastatus` instead of using `ssh ... /usr/bin/paasta_metastatus` if `USE_API_ENDPOINT` environment variable is `true`.
```
$ paasta metastatus -c norcal-stagef -vv -a >/tmp/paasta-metastatus-ORIG
$ USE_API_ENDPOINT=1 paasta metastatus -c norcal-stagef -vv -a >/tmp/paasta-metastatus-NEW 
$ diff -u /tmp/paasta-metastatus-ORIG /tmp/paasta-metastatus-NEW
--- /tmp/paasta-metastatus-ORIG 2018-09-14 04:15:49.874849796 -0700
+++ /tmp/paasta-metastatus-NEW  2018-09-14 04:15:37.787245547 -0700
@@ -27,8 +27,8 @@
     Resource id                Pool      State   Current  Target  Min capacity  Max capacity  Instances
     spam_paasta_stagef_xenial  spam      active  1        1       1             1             1
     yr_paasta_stagef           yelp_res  active  5        5       3             20            5
-    paasta_stagef              default   active  0        0       0             0             0
     paasta_kew_stagef          kew       active  3        3       3             3             3
+    paasta_stagef              default   active  0        0       0             0             0
 Marathon Status: OK
   marathon apps:       1696
   marathon tasks:      3580
```

### DIFF
--- a/paasta_tools/cli/cmds/metastatus.py
+++ b/paasta_tools/cli/cmds/metastatus.py
@@ -140,8 +140,7 @@ def paasta_metastatus_on_api_endpoint(
         res = client.metastatus.metastatus(cmd_args=[str(arg) for arg in cmd_args]).result()
         output, exit_code = res.output, res.exit_code
     except HTTPError as exc:
-        paasta_print(exc.response.text)
-        return exc.status_code
+        output, exit_code = exc.response.text, exc.status_code
 
     return exit_code, output
 
@@ -158,23 +157,18 @@ def print_cluster_status(
     """With a given cluster and verboseness, returns the status of the cluster
     output is printed directly to provide dashboards even if the cluster is unavailable"""
     if use_api_endpoint:
-        return_code, output = paasta_metastatus_on_api_endpoint(
-            cluster=cluster,
-            system_paasta_config=system_paasta_config,
-            groupings=groupings,
-            verbose=verbose,
-            autoscaling_info=autoscaling_info,
-            use_mesos_cache=use_mesos_cache,
-        )
+        metastatus_func = paasta_metastatus_on_api_endpoint
     else:
-        return_code, output = execute_paasta_metastatus_on_remote_master(
-            cluster=cluster,
-            system_paasta_config=system_paasta_config,
-            groupings=groupings,
-            verbose=verbose,
-            autoscaling_info=autoscaling_info,
-            use_mesos_cache=use_mesos_cache,
-        )
+        metastatus_func = execute_paasta_metastatus_on_remote_master
+
+    return_code, output = metastatus_func(
+        cluster=cluster,
+        system_paasta_config=system_paasta_config,
+        groupings=groupings,
+        verbose=verbose,
+        autoscaling_info=autoscaling_info,
+        use_mesos_cache=use_mesos_cache,
+    )
 
     paasta_print("Cluster: %s" % cluster)
     paasta_print(get_cluster_dashboards(cluster))
--- a/paasta_tools/cli/cmds/metastatus.py
+++ b/paasta_tools/cli/cmds/metastatus.py
@@ -12,9 +12,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import os
+from distutils.util import strtobool
 from typing import Sequence
+from typing import Tuple
 
+from bravado.exception import HTTPError
+
+from paasta_tools.api.client import get_paasta_api_client
 from paasta_tools.cli.utils import execute_paasta_metastatus_on_remote_master
+from paasta_tools.cli.utils import get_paasta_metastatus_cmd_args
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_services
 from paasta_tools.utils import DEFAULT_SOA_DIR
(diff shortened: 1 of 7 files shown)


paasta_tools/paasta_metastatus.py
```diff
@@ -338,9 +349,50 @@ def main(argv: Optional[List[str]]=None) -> None:
...
+        if 'file' in kwargs:
+            del kwargs['file']
+        return print(*args, **kwargs, file=output)
+
+    exit_code = 1
+    PAASTA_PRINT_PATCH_LOCK.acquire()
+    global paasta_print
+    orig_paasta_print = paasta_print
+    try:
+        # Monkey-patching `paasta_print` everywhere it's used to output results of `paasta metastatus` cmd.
```
keymone: i'm not a fan of such patching, eventually we will shoot ourselves in the foot. capturing output is tricky and there's definitely more than one way we print stuff in paasta.

ideally we would rewrite `print_output` to simply do `paasta_print(data for data in get_output())` and make a `get_output` function  that returns list of messages (or a generator) instead of printing them.

i realize that this may be a lot of changes, but i think it's important.

paasta_tools/cli/cmds/metastatus.py
```diff
@@ -110,24 +117,64 @@ def add_subparser(
...
     output is printed directly to provide dashboards even if the cluster is unavailable"""
-    return_code, output = execute_paasta_metastatus_on_remote_master(
-        cluster=cluster,
-        system_paasta_config=system_paasta_config,
-        groupings=groupings,
-        verbose=verbose,
-        autoscaling_info=autoscaling_info,
-        use_mesos_cache=use_mesos_cache,
-    )
+    if use_api_endpoint:
```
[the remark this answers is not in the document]
vkhromov: done

paasta_tools/cli/cmds/metastatus.py
```diff
@@ -110,24 +117,64 @@ def add_subparser(
...
+            groupings=groupings,
+            verbose=verbose,
+            autoscaling_info=autoscaling_info,
+            use_mesos_cache=use_mesos_cache,
+        )
+        res = client.metastatus.metastatus(cmd_args=[str(arg) for arg in cmd_args]).result()
+        output, exit_code = res.output, res.exit_code
+    except HTTPError as exc:
+        paasta_print(exc.response.text)
+        return exc.status_code
```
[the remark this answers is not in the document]
vkhromov: done

paasta_tools/cli/cmds/metastatus.py
```diff
@@ -176,6 +223,12 @@ def paasta_metastatus(
     """Print the status of a PaaSTA clusters"""
     soa_dir = args.soa_dir
     system_paasta_config = load_system_paasta_config()
+
+    if 'USE_API_ENDPOINT' in os.environ:
```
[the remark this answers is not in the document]
vkhromov: We're already using `USE_API_ENDPOINT` for `paasta status` via HTTP API, so it's used there for consistency.

paasta_tools/paasta_metastatus.py
vkhromov: The current design is a compromise to make "`paasta metastatus` via HTTP" change as simple as possible.

Ideally we probably should export all `metastatus_lib` functions we use to generate `paasta metastatus` output via HTTP API and use them on the client.  I've tried to start with that, but it seems that it's a way too much work for the result.

What do you think about introducing a context manager which sets output destination for `paasta_print`, and change `paasta_print` to output everything into that destination if it's set?  This way we will not need any monkey patching?
I.e.
```
with set_paasta_print_file(output):
    print_output(argv)
```

keymone: context looks better, though it's still going to try to capture output behind the scenes. i don't want to block progress, so ship away, but in future we need to rethink this.

vkhromov: Replaced with contextmanager-based code.

solarkennedy:
The cmd_args csv approach is interesting, but a bit non-traditional. It does make it easier to iterate and migrate to.

Unlike `status`, can `metastatus` be migrated to use the api with all of its features intact? If so, do you plan to remove the ssh version of metastatus after you have confirmed that this all works? I would rather only have one way of doing things.

solarkennedy:
ship!

paasta_tools/utils.py
```diff
@@ -2854,13 +2854,33 @@ def to_bytes(obj: Any) -> bytes:
...
+    buf = getattr(f, 'buffer', None)
+    # Here we're assuming that the file object works with strings and its
+    # `buffer` works with bytes. So, if the file object doesn't have `buffer`,
+    # we output via the file object itself using strings.
+    obj_to_arg: Callable[[Any], Any]
+    if buf is not None:
+        f = buf
+        obj_to_arg = to_bytes
+    else:
+        def obj_to_arg(o: Any) -> str: return to_bytes(o).decode('UTF-8', errors="ignore")
```
Rob-Johnson: the `errors="ignore"` seems dangerous - can you explain?

paasta_tools/utils.py
```diff
@@ -2854,13 +2854,33 @@ def to_bytes(obj: Any) -> bytes:
         return str(obj).encode('UTF-8')
 
 
+TLS = threading.local()
```
Rob-Johnson: howcome we need this to be thread local?

vkhromov: If `paasta_print()` is called from different threads, then it could have different `paasta_print_file` object sets by context managers in different threads. So, `paasta_print_file` needs to be a per-thread variable.

paasta_tools/utils.py
vkhromov: `to_bytes()` converts its string and object args to UTF-8-encoded byte sequences.  At the same time, it doesn't convert byte sequence args at all, so there is no guaranty that a byte sequences returned from `to_bytes()` can be safely converted back to a string using UTF-8 encoding. So, there is a possibility that we can have `UnicodeEncodeError` exception if we try to decode `to_bytes()`'s return value to a string using UTF-8 encoding.

At the same time, `print_paasta()` was introduced partially to prevent `UnicodeEncodeError: 'ascii' codec can't encode character` errors while printing.  So, to satisfy that, we need to make sure, that decoding `to_bytes()`'s return value is always successful.  To do that, I've added `error="ignore"` flag.

paasta_tools/utils.py
vkhromov: I.e. let imagine that `paasta_print()` is simultaneously called from two threads serving two `/metastatus` HTTP API requests from two clients.  Is this case, we have two separate `io.StringIO` objects containing responses to corresponding clients, one in every thread, and we need `print_paasta()` to write it's output into a corresponding `io.String` output object.

PR #1985, opened 2018-09-18
Drop all capabilities by default in docker containers.

solarkennedy:
I got lazy and copy pasted in the tests.


mattmb:
Any chance of somehow making it a SystemPaastaConfig so that you can roll it out with a config change? It'll be another big bounce right which will mean pinning otherwise.

mattmb:
++ on the general idea btw

I also wouldn't be *that* surprised if some service ends up needing one from the list. Maybe it's also worth making it configurable per service?

solarkennedy:
I think we can take another big bounce. Or I can batch it.

I was really hoping to not add yet another configuration option.
I *would* be surprised if we have a service that needs one of those capabilities actually.

solarkennedy:
The tests are failing because of some weird coverage stdout io nonsense?

solarkennedy:
Tests are passing now.

PR #1987, opened 2018-09-19
Update paasta status to use http api when needed

mattmb:
This makes it so paasta status will force use of the HTTP api for k8s
instances whilst still defaulting to the SSH api for other instance
types. As we add features to the HTTP api e.g. for marathon app status
we can migrate it over to using the HTTP by adding to
HTTP_ONLY_INSTANCE_CONFIG.

I think I've fixed all the tests but lets see what travis says...

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -20,13 +20,14 @@ from collections import defaultdict
 from distutils.util import strtobool
 from typing import Callable
 from typing import DefaultDict
+from typing import Dict
 from typing import Iterable
 from typing import List
 from typing import Mapping
-from typing import Optional
 from typing import Sequence
 from typing import Set
 from typing import Tuple
+from typing import Type
 
 from bravado.exception import HTTPError
 from service_configuration_lib import read_deploy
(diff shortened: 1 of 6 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -385,38 +400,42 @@ def report_status_for_cluster(
...
-        else:
-            return_code, status = execute_paasta_serviceinit_on_remote_master(
-                'status', cluster, service, ','.join(deployed_instances),
+        if not use_api_endpoint and ssh_instances or not http_only_instances:
+            ssh_return_code, status = execute_paasta_serviceinit_on_remote_master(
+                'status', cluster, service, ','.join(ssh_instances),
                 system_paasta_config, stream=False, verbose=verbose,
                 ignore_ssh_output=True,
             )
+            return_code = ssh_return_code + return_code
```
Rob-Johnson: is this right? so if we get exit 1 and 1, the exit code will show as 2?

mattmb: Hmm, yeah let me try something else

PR #1988, opened 2018-09-20
Remove coverage accounting

solarkennedy:
idk, do we need it?


solarkennedy:
It was at 82% apparently.

PR #1991, opened 2018-09-24
Add docs for cpu_burst_add

kawaiwanyelp:
By @solarkennedy's request, I have added some documentation for the `cpu_burst_add` parameter under the cpus bullet point, just so people know it exists and specifies a number of cpus (no longer percentage based).

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -17,10 +17,10 @@ specify the following options:
     available free CPU, but is guaranteed to get the CPU shares specified.  For
     a more detailed read on how this works in practice, see the docs on `isolation <isolation.html>`_.
 
-    * ``cpu_burst_add``: Number of additional CPUs an instance may use while
-      bursting; if unspecified, PaaSTA defaults to 1. For example, if a service
-      specifies that it needs 2 CPUs normally and 1 for burst, the service may
-      go up to 3 CPUs, if needed.
+  * ``cpu_burst_add``: Maximum number of additional CPUs an instance may use
+    while bursting; if unspecified, PaaSTA defaults to 1. For example, if a
+    service specifies that it needs 2 CPUs normally and 1 for burst, the service
+    may go up to 3 CPUs, if needed.
 
   * ``mem``: Memory (in MB) an instance needs. Defaults to 1024 (1GB). In Mesos
     memory is constrained to the specified limit, and tasks will reach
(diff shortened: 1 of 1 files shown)


docs/source/yelpsoa_configs.rst
```diff
@@ -17,6 +17,9 @@ specify the following options:
     available free CPU, but is guaranteed to get the CPU shares specified.  For
     a more detailed read on how this works in practice, see the docs on `isolation <isolation.html>`_.
 
+    * ``cpu_burst_add``: Number of additional CPUs an instance may use while
```
solarkennedy: The indentation looks off here. Can you double check?

docs/source/yelpsoa_configs.rst
```diff
@@ -17,6 +17,9 @@ specify the following options:
...
+      bursting; if unspecified, PaaSTA defaults to 1.
```
solarkennedy: Say a bit more, `which means if a serivce ask for `cpus: 1`, it can actually burst to 2 full cpus if needed.

docs/source/yelpsoa_configs.rst
kawaiwanyelp: The indentation is one level deeper than the cpus entry above, same as the PaaSTA variables listed underneath the env entry below.

docs/source/yelpsoa_configs.rst
```diff
@@ -17,6 +17,11 @@ specify the following options:
     available free CPU, but is guaranteed to get the CPU shares specified.  For
     a more detailed read on how this works in practice, see the docs on `isolation <isolation.html>`_.
 
+    * ``cpu_burst_add``: Number of additional CPUs an instance may use while
```
qui: "Maximum number", to make it clear this is a hard limit?

docs/source/yelpsoa_configs.rst
solarkennedy: Please don't. This is a top level key and not some sort of sub-key of `cpus`. Env is different because that is just a list of stuff, those are not configuration options for the file.

kawaiwanyelp: Fair.

docs/source/yelpsoa_configs.rst
kawaiwanyelp: Okay, I'll make the change.

PR #1992, opened 2018-09-25
Add missing tron action fields

keymone:
Fields related to external job dependencies.


qui:
I think you also need to update https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/schemas/tron_schema.json#L14 for paasta validate to work. (It runs on pre-commit and pre-receive if the tron file is in a service directory right now.)

PR #1993, opened 2018-09-26
Expanded the shared_ip check to also check iptables rules

solarkennedy:
I feel a tad dirty overloading this check. It is certainly within the same spirit of what the check does, but it checks for two ports using the same ip instead of two containers using the same ip?

The output on the failing host looks like this:
```
OK - No Docker containers sharing an IP on this host.
CRITICAL - This is the second time we've seen a rule with the same target_parameters!
Rule(protocol='tcp', src='0.0.0.0/0.0.0.0', dst='0.0.0.0/0.0.0.0', target='DNAT', matches=(('tcp', (('dport', ('31537',)),)),), target_parameters=(('to-destination', ('169.254.0.24:8888',)),))
The other rule is:
Rule(protocol='tcp', src='0.0.0.0/0.0.0.0', dst='0.0.0.0/0.0.0.0', target='DNAT', matches=(('tcp', (('dport', ('31880',)),)),), target_parameters=(('to-destination', ('169.254.0.24:8888',)),))
```

I'm not too attached, I can make a separate monitoring check and stuff if you want, it is just a tad more work.

@bplotnick

### DIFF
--- a/paasta_tools/contrib/shared_ip_check.py
+++ b/paasta_tools/contrib/shared_ip_check.py
@@ -1,7 +1,19 @@
+#!/usr/bin/env python3.6
 import sys
 from collections import defaultdict
 
+import iptc
+
+from paasta_tools import iptables
 from paasta_tools.utils import get_docker_client
+
+
+def list_docker_nat_rules():
+    chain_name = 'DOCKER'
+    table = iptc.Table(iptc.Table.NAT)
+    chain = iptc.Chain(table, chain_name)
+    for rule in chain.rules:
+        yield iptables.Rule.from_iptc(rule)
 
 
 def main():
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/shared_ip_check.py
```diff
@@ -33,7 +50,19 @@ def main():
         return 2
     else:
         print('OK - No Docker containers sharing an IP on this host.')
-        return 0
+
+    seen = []
```
qui: This could be a dict of target_parameters to rules, then checking if it contains target_parameters would be faster and you don't need find_rule.

solarkennedy: Ah! good idea

solarkennedy:
@qui can you review this again please?

paasta_tools/contrib/shared_ip_check.py
```diff
@@ -33,7 +45,19 @@ def main():
...
+    for rule in list_docker_nat_rules():
+        target = rule.target_parameters
+        if target not in targets_seen:
+            targets_seen[target] = rule
+        else:
+            print("CRITICAL - This is the second time we've seen a rule with the same target_parameters!")
+            print(rule)
+            print("The other rule with that target is:")
+            print(targets_seen[target])
+            return 2
```
qui: Do you want to complete the loop in case there is more than one repeated rule?

PR #1995, opened 2018-09-27
Replace unicode errors from user-inputted stuff

solarkennedy:
This was the only other place in our codebase where we were doing unicode decoding with stuff that comes from the user (paasta_print already did it).

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -1971,24 +1971,25 @@ def _run(
         if timeout:
             proctimer = threading.Timer(timeout, _timeout, [process])
             proctimer.start()
-        for linestr in iter(process.stdout.readline, b''):
-            line = linestr.decode('utf-8')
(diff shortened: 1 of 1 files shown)


solarkennedy:
But we also do this thing where we return an `output` object, even if we don't log.
I don't know if it is such great idea to just put bytes in there.

solarkennedy:
@EvanKrall no, I also can't paasta_print -> print if we still want the behavior of stripping trailing newlines if they exist (because that means I would have to decode anyway)

EvanKrall:
```
>>> b'hi hello this ends with a newline\n'.strip(b'\n')
b'hi hello this ends with a newline'
```

The output thing may be a blocker / make this turn into a much bigger branch, though I do feel pretty strongly that we should be able to handle arbitrary bytes when forwarding the output of services. (What if a program spits out binary output, like gzip or png?)

PR #1996, opened 2018-09-28
Add {}-type formatting for Tron commands

qui:
Now that Tron commands use {} instead of %, parse_time_variables needs to handle those differently for local-run.

Chronos commands still use % now... The 2nd commit provides a migration path forward:
1. Release this version of paasta tools
2. change all commands to {} in soaconfigs, with use_percent_format: False
3. Change parse time variables to always use {} instead of percent
4. Remove use_percent_format from soaconfigs
5. Remove use_percent_format from paasta tools

Alternatively, we could not bother with migrating and leave use_percent hardcoded to True for Chronos. Then we wouldn't need the 2nd commit at all.

### DIFF
--- a/paasta_tools/chronos_rerun.py
+++ b/paasta_tools/chronos_rerun.py
@@ -79,8 +79,8 @@ def modify_command_for_date(chronos_job, date, verbose, use_percent):
     current_command = chronos_job['command']
     if current_command is not None:
         chronos_job['command'] = chronos_tools.parse_time_variables(
-            current_command,
-            date,
+            command=current_command,
+            parse_time=date,
             use_percent=use_percent,
         )
     else:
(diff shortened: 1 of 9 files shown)


solarkennedy:
This looks great! And I like your plan. I would even volunteer to help! (I don't want to leave things in mixed-mode)

paasta_tools/chronos_rerun.py
```diff
@@ -72,12 +72,17 @@ def modify_command_for_date(chronos_job, date, verbose):
...
     :param date: a ``datetime.datetime`` object.
+    :param use_percent: if time variables use percent formatting
     :returns chronos_job: a chronos_job dict with the command modified to
         interpolate in the context of the date provided.
     """
     current_command = chronos_job['command']
     if current_command is not None:
-        chronos_job['command'] = chronos_tools.parse_time_variables(current_command, date)
+        chronos_job['command'] = chronos_tools.parse_time_variables(
+            current_command,
```
solarkennedy: can you kwargs all the things please

solarkennedy:
Oh yea. Script it up!
It might be a little harder because this requires the `use_percent_format: false` to be added, which means yaml manipulation.

PR #1997, opened 2018-09-28
Fix suggestions on bad instance names

solarkennedy:
This removes our hilarious suggestions of:
```
Error: bumpy.main doesn't look like it has been configured to run on the norcal-devc cluster.
Did you mean one of: main, main, main?
Did you mean one of: main, main?
```
Because we are appending the entire suggestion string lolz.

Also I'm not sure why this function is so inefficient, it iterates over *all* services, even though we know the service name?

### DIFF
--- a/paasta_tools/firewall_update.py
+++ b/paasta_tools/firewall_update.py
@@ -123,8 +123,9 @@ def smartstack_dependencies_of_running_firewalled_services(soa_dir=DEFAULT_SOA_D
     dependencies_to_services = defaultdict(set)
     for service, instance, _, _ in firewall.services_running_here():
         config = get_instance_config(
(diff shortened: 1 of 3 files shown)


tests/cli/test_utils.py
```diff
@@ -513,6 +513,17 @@ def test_list_instances_no_service(
     assert actual == expected
 
 
+def test_validate_service_instance_gives_suggestions():
```
qui: Should be deleted in favor of the tests you added in tests/test_utils.py, right? It's broken anyway.

solarkennedy: Whoa whoops! 
Nothing to see here....

PR #1998, opened 2018-10-01
Refactor check_marathon_services_replication

mattmb:
This should be a noop in terms of functionality

Mostly moving things from check_marathon_services_replication to
monitoring_tools so that I will be able to re-use some of these
methods to create check_kubernetes_services_replication.

I also changed get_proxy_port_for_instance to take an InstanceConfig so
that I can reuse for k8s too. And removed the
read_all_registrations_for_service_instance and
read_registration_for_service_instance which I feel are not really
necessary to be separate functions.

### DIFF
--- a/paasta_tools/check_marathon_services_replication.py
+++ b/paasta_tools/check_marathon_services_replication.py
@@ -32,7 +32,6 @@ instance then it will be used instead.
 """
 import argparse
 import logging
-import os
 from datetime import datetime
 from datetime import timedelta
 
@@ -41,57 +40,19 @@ import pysensu_yelp
 
 from paasta_tools import marathon_tools
 from paasta_tools import monitoring_tools
+from paasta_tools.long_running_service_tools import get_proxy_port_for_instance
 from paasta_tools.marathon_tools import format_job_id
 from paasta_tools.mesos_tools import get_slaves
 from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.smartstack_tools import SmartstackReplicationChecker
-from paasta_tools.utils import _log
 from paasta_tools.utils import datetime_from_utc_to_local
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import is_under_replicated
+from paasta_tools.utils import list_services
 from paasta_tools.utils import load_system_paasta_config
 
 
 log = logging.getLogger(__name__)
-
-
-def send_event(instance_config, status, output):
-    """Send an event to sensu via pysensu_yelp with the given information.
-
-    :param instance_config: an instance of MarathonServiceConfig
-    :param status: The status to emit for this event
-    :param output: The output to emit for this event"""
-    # This function assumes the input is a string like "mumble.main"
-    monitoring_overrides = instance_config.get_monitoring()
-    if 'alert_after' not in monitoring_overrides:
-        monitoring_overrides['alert_after'] = '2m'
-    monitoring_overrides['check_every'] = '1m'
-    monitoring_overrides['runbook'] = monitoring_tools.get_runbook(
-        monitoring_overrides,
-        instance_config.service, soa_dir=instance_config.soa_dir,
-    )
-
-    check_name = (
-        'check_marathon_services_replication.%s' %
-        instance_config.job_id
-    )
-    monitoring_tools.send_event(
-        service=instance_config.service,
-        check_name=check_name,
-        overrides=monitoring_overrides,
-        status=status,
-        output=output,
-        soa_dir=instance_config.soa_dir,
-        cluster=instance_config.cluster,
-    )
-    _log(
-        service=instance_config.service,
-        line='Replication: %s' % output,
-        component='monitoring',
-        level='debug',
-        cluster=instance_config.cluster,
-        instance=instance_config.instance,
-    )
 
 
 def parse_args():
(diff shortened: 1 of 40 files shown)


mattmb:
I think the tests will pass but this should have careful review just because I've touched to code that sends alerts and which generates nerve configs.

mattmb:
I've also just added a commit to `SmartstackReplicationChecker` to make it generic and add `Mesos SmartstackReplicationChecker`

That commit could be reviewed separately (the tests pass on both commits). But I'll leave them as one PR I think.

paasta_tools/kubernetes_tools.py
```diff
@@ -784,10 +756,14 @@ def get_kubernetes_services_running_here_for_nerve(
     nerve_list = []
     for kubernetes_service in kubernetes_services:
         try:
-            registrations = read_all_registrations_for_service_instance(
-                kubernetes_service.name, kubernetes_service.instance, cluster, soa_dir,
+            kubernetes_service_config = load_kubernetes_service_config(
```
EvanKrall: If `load_kubernetes_service_config` is implemented like `load_marathon_service_config`, then this loop will re-parse kubernetes.yaml for every instance, which can cause significant slowdowns for large yaml files. Can you switch to PaastaServiceConfigLoader instead?

paasta_tools/native_mesos_scheduler.py
```diff
@@ -109,10 +109,14 @@ def get_paasta_native_services_running_here_for_nerve(cluster, soa_dir, hostname
...
-                name, instance, cluster, soa_dir,
+            job_config = load_paasta_native_job_config(
+                service=name,
+                instance=instance,
+                cluster=cluster,
+                load_deployments=False,
+                soa_dir=soa_dir,
             )
-            for registration in registrations:
+            for registration in job_config.get_registrations():
```
EvanKrall: Could also do PaastaServiceConfigLoader here, if it's not too much work.

paasta_tools/smartstack_tools.py
```diff
@@ -412,42 +414,33 @@ def match_backends_and_tasks(
...
-    >>> mesos_slaves = get_slaves()
-    >>> system_paasta_config = load_system_paasta_config()
-    >>> instance_config = load_marathon_service_config(service='fake_service',
-    ...                       instance='fake_instance', cluster='norcal-stagef')
-    >>>
-    >>> c = SmartstackReplicationChecker(mesos_slaves, system_paasta_config)
-    >>> c.get_replication_for_instance(instance_config)
-    {'uswest1-stagef': {'fake_service.fake_instance': 2}}
-    >>>
+    _get_allowed_locations_and_hosts must be implemented in super class
```
EvanKrall: s/super class/subclass/

Maybe use `abc.ABC` and decorate `_get_allowed_locations_and_hosts` with `@abc.abstractmethod`

paasta_tools/smartstack_tools.py
```diff
@@ -412,42 +414,33 @@ def match_backends_and_tasks(
...
 
 
+class SmartstackHost(NamedTuple):
+    hostname: str
+    pool: str
+
+
 class SmartstackReplicationChecker:
-    """Retrieves the number of registered instances in each discoverable location.
+    """Base class for checking smartstack replication. Extendable for different frameworks
```
EvanKrall: period at end of sentence

paasta_tools/check_marathon_services_replication.py
```diff
@@ -301,18 +164,13 @@ def check_service_replication(
     """
     expected_count = instance_config.get_instances()
     log.info("Expecting %d total tasks for %s" % (expected_count, instance_config.job_id))
-    proxy_port = marathon_tools.get_proxy_port_for_instance(
-        name=instance_config.service,
-        instance=instance_config.instance,
-        cluster=instance_config.cluster,
-        soa_dir=instance_config.soa_dir,
-    )
+    proxy_port = get_proxy_port_for_instance(instance_config, soa_dir=instance_config.soa_dir)
```
solarkennedy: Is soa_dir an attribute of our instance configs? Cause that is cool.
If so, how about just always extract that from instance_config inside this function and simplify the calling args?
We'll never need to read an instance_config from one place and use a different soa_dir to lookup the port elsewhere,

paasta_tools/long_running_service_tools.py
```diff
@@ -389,3 +389,27 @@ def set_instances_for_marathon_service(
...
+
+    :param name: The service name
+    :param instance: The instance of the service
+    :param cluster: The cluster to read the configuration for
+    :param soa_dir: The SOA config directory to read from
+    :returns: The proxy_port for the service instance, or None if not defined"""
+    registration = service_config.get_registrations()[0]
+    service, namespace, _, __ = decompose_job_id(registration)
+    nerve_dict = load_service_namespace_config(
+        service=service, namespace=namespace, soa_dir=soa_dir,
```
solarkennedy: yea, just `instance_config.soa_dir` if that is a thing.

paasta_tools/check_marathon_services_replication.py
mattmb: oh good point, I don't see why not!

paasta_tools/kubernetes_tools.py
mattmb: I guess it will be a bit of repeated work for multiple instances of the same service. At least it's just those running on the particular host so I don't think it's that bad. I'm probably going to leave it be for now because I'm already changing a lot here and don't want to make it any riskier.

paasta_tools/smartstack_tools.py
mattmb: I think you mean "full stop" 😛

PR #2001, opened 2018-10-02
Add k8s cluster

mattmb:
This adds a kubernetes example cluster setup that roughly re-uses our
mesos based example cluster.

I'm planning to work back from this to create a proper itest suite but
for now this:
* creates a non-smartstack k8s service
* checks that paasta status works
* checks that paasta-api shows the k8s service as running

There's two minor code changes to make these new tests pass, they're on separate commits so you can check them individually. NB, I'm also setting this as a new travis check but allowing it to fail rather than block the build for now.


example_cluster/dind-cluster-v1.11.sh
```diff
@@ -0,0 +1,2049 @@
+#!/bin/bash
```
solarkennedy: A 2049 line bash script? Sure why not!!! :)

paasta_tools/cli/cmds/status.py
```diff
@@ -416,7 +416,7 @@ def report_status_for_cluster(
                 for deployed_instance in deployed_instances
                 if (deployed_instance in http_only_instances or use_api_endpoint)
             ]
-            if any(return_code != 200 for return_code in return_codes):
+            if any(return_codes):
```
solarkennedy: wait, why not filter out 200's?

example_cluster/docker-compose.k8s.yaml
```diff
@@ -0,0 +1,84 @@
+version: '3'
+networks:
+  kubeadm-dind-net:
+    external: True
+volumes:
+  ssh-data: {}
+  nail-etc:
+services:
+  itest_trusty:
+    build: ../yelp_package/dockerfiles/trusty/
```
vkhromov: Can we use Xenial?

example_cluster/dind-cluster-v1.11.sh
mattmb: :-P at least I'm just vendoring it: https://github.com/kubernetes-sigs/kubeadm-dind-cluster

paasta_tools/cli/cmds/status.py
mattmb: Let me double check, I think the function can't ever return the 200 at the moment

PR #2002, opened 2018-10-03
Script for migrating Chronos commands to {}-formatting or back

qui:
We can use `--include-flag` first, then run it without once we're safely on the next version where {}-format is default. In case we need to go back on paasta tools for whatever reason, we can also use `--downgrade` to go back to %.

I wanted to keep the diffs as small as possible, so this processes the files line-by-line to preserve formatting as much as possible. Ugly, but it works! 😬

### DIFF
--- a/paasta_tools/contrib/migrate_chronos_percent_format.py
+++ b/paasta_tools/contrib/migrate_chronos_percent_format.py
@@ -1,0 +1,197 @@
+"""Script for migrating Chronos command string formats.
+
+Used to upgrade from v0.80.11 to v0.81.0 or downgrade to a version <= v0.80.11.
+"""
+import argparse
+import os
+import re
+import shutil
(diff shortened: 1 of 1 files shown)


solarkennedy:
Thanks for going the extra mile to minimize diffs to get us through this migration. 
As far as adhoc contrib scripts go, this looks really solid and may come in handy for future stuff!

paasta_tools/contrib/migrate_chronos_percent_format.py
```diff
@@ -0,0 +1,197 @@
+"""Script for migrating Chronos command string formats.
```
solarkennedy: But a shebang here for `#!/usr/bin/env python3.6` make it executable and dh-virtualenv should replace it with the right shebang to make it slightly easier to run.

PR #2003, opened 2018-10-03
Fix paasta api memory leak

vkhromov:
Add `cleanup_self` to `@async_ttl_cache` and use it for all methods

`@async_ttl_cache` decorator caches a function call result using a key
composed from function's args values.  If it's used for a method, it
will use `self` as a part of the key, creating a strong reference to the
class instances in the internal cache as a result.  After some time,
references to the class instances are removed, so the class instance
reference in the cache becomes the only existing reference to that class
instance, and since the method is never called with that class instance
as a `self` arg, the corresponding record in the internal cache is never
expired, creating the memory leak as a result.

Introducing `cleanup_self` arg to `@async_ttl_cache` decorator, so while
it's `True` the decorator creates a weak reference to `self`, and
autocleans corresponding cache record on the object finalisation.  Using
it for all methods, so there will be no more corresponding memory leaks
in `paasta-api`.

While here, also replaced annoying `print()` with logging.

### DIFF
--- a/paasta_tools/api/views/marathon_dashboard.py
+++ b/paasta_tools/api/views/marathon_dashboard.py
@@ -15,15 +15,19 @@
 """
 Marathon Dashboard
 """
+import logging
+
 from pyramid.view import view_config
 
 from paasta_tools.api import settings
 from paasta_tools.marathon_dashboard import create_marathon_dashboard
 
+log = logging.getLogger(__name__)
+
 
 @view_config(route_name='marathon_dashboard', request_method='GET', renderer='json')
 def marathon_dashboard(request):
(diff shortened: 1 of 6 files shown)


solarkennedy:
!

mattmb:
I can't really think of a better solution. I wonder how much the cache is really helping given that the MesosMaster object probably changes all the time. i.e. I wonder how many cache hits we actually get? Anyway this hasn't changed that so I say ship (other than my comment above).

vkhromov:
@mattmb, I kind of merged them.  I could try to extract some common code from `inner`'s functions as well, but that seems to be not trivial.

PR #2005, opened 2018-10-04
Run all tron actions in paasta with the docker '--init' option

solarkennedy:
I know I just made a screencast explaining how to properly use dumb-init with complex shell commands, but we can solve this for every user simultaneously with this option, and eliminate an entire class of user-error by not making them have another set of quotes and stuff with `/bin/sh -c 'something'`

Tron on paasta is greenfield, so I'm not afraid of pushing this change for tron things only.

@chriskuehl

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -423,7 +423,7 @@ def get_docker_run_cmd(
         cmd.append(f'{k}')
     cmd.append('--memory=%dm' % memory)
     for i in docker_params:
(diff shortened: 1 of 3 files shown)


solarkennedy:
Gosh. Hurray for itests. Let me address them.

qui:
Just curious, what happens if you have --init and run a command that also contains 'exec dumb-init'?

solarkennedy:
As long as we continue to run tron commands with `/bin/sh` (which we do), then it will continue to work. It will just have `tini` -> `dumb-init` -> `python` instead of just dumb-init. Signals will get forwarded like before.

PR #2007, opened 2018-10-09
APOLLO-565: Add a flag on updating haproxy on the map updates

ronin13:
APOLLO-565: Add a flag on updating haproxy on the map updates.

This commit adds a flag (--update-haproxy) which when provide updates
the haproxy with updated map information. The map file contains
the internal bridge IP address of a service and mesos task information
of the service.

This changes the present default behavior which updates haproxy on every
map update. This is not scalable (often fails with `Errno 11`)
and not needed since with synapse haproxy is restarted quite often.
The map_file is part of haproxy.cfg which means we don't need to
actively update haproxy, instead rely on these restarts for haproxy
to pick up updated info.

Note that we are only changing the default, hence, to revert to old behavior,
the flag (--update-haproxy) needs to be provided wherever it is invoked
(puppet).

Tested on a stage box.

### DIFF
--- a/paasta_tools/contrib/get_containers_and_ips.py
+++ b/paasta_tools/contrib/get_containers_and_ips.py
@@ -96,6 +96,12 @@ def main():
         ),
     )
     parser.add_argument(
+        '--update-haproxy',
+        '-U',
+        action='store_true',
+        help='Whether to update haproxy for map updates',
+    )
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


tub:
@ronin13 Can you add a bit of background about why you might want to use the flag?

ronin13:
@tub  Thanks. Added.

PR #2008, opened 2018-10-09
Expose calculated (pre-limited) number of instance by autoscaler

sarguru:
This change
  * adds the function get_instances_without_limiting() which gets number of instance required as calculated by autoscaler without limiting.
 * adds response parameter calculated_instance which exposes this in the API endpoint v1/services/{service}/{instance}/autoscaling.

This change is to find out the acutal desired instances, since min & max
instance values could be adjusted by the operators/users in emergency
scenarious to pre-scale the services.

Example response now
```
{
  "desired_instances": 3,
  "calculated_instances": 2
}
```

### DIFF
--- a/paasta_tools/long_running_service_tools.py
+++ b/paasta_tools/long_running_service_tools.py
@@ -185,7 +185,7 @@ class LongRunningServiceConfig(InstanceConfig):
         """
         return self.config_dict.get('bounce_priority', 0) * -1
 
-    def get_instances(self) -> int:
+    def get_instances(self, with_limit: bool=True) -> int:
         """Gets the number of instances for a service, ignoring whether the user has requested
         the service to be started or stopped"""
         if self.get_max_instances() is not None:
@@ -199,7 +199,7 @@ class LongRunningServiceConfig(InstanceConfig):
                 log.debug("No zookeeper data, returning max_instances (%d)" % self.get_max_instances())
                 return self.get_max_instances()
             else:
-                limited_instances = self.limit_instance_count(zk_instances)
+                limited_instances = self.limit_instance_count(zk_instances) if with_limit else zk_instances
                 return limited_instances
         else:
             instances = self.config_dict.get('instances', 1)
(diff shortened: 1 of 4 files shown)


paasta_tools/long_running_service_tools.py
```diff
@@ -206,6 +206,26 @@ def get_instances(self) -> int:
             log.debug("Autoscaling not enabled, returning %d instances" % instances)
             return instances
 
+    def get_instances_without_limit(self) -> int:
```
mattmb: How about just making this a param on `get_instances`. Something like `get_instances(self, with_limit=True)` so that it defaults to the current behaviour.

sarguru: Sounds good. Making the change.

paasta_tools/long_running_service_tools.py
```diff
@@ -185,7 +185,7 @@ def get_bounce_priority(self) -> int:
         """
         return self.config_dict.get('bounce_priority', 0) * -1
 
-    def get_instances(self) -> int:
+    def get_instances(self, with_limit=True) -> int:
```
mattmb: Sorry to nitpick but can you add a type too (we're trying to add them whenever we add new stuff)
The syntax is like: `get_instances(self, with_limit: bool=True) -> int:`

tests/api/test_autoscaler.py
```diff
@@ -28,9 +28,11 @@ def test_get_autoscaler_count():
         'paasta_tools.api.views.autoscaler.load_marathon_service_config',
         autospec=True,
     ) as mock_load_marathon_service_config:
-        mock_load_marathon_service_config.return_value = mock.MagicMock(get_instances=mock.MagicMock(return_value=123))
+        mock_load_marathon_service_config.return_value = mock.MagicMock(get_instances=mock.MagicMock(return_value=123),
+                get_instances_without_limit=mock.MagicMock(return_value=122))
```
mattmb: this no longer matches the code right?

PR #2009, opened 2018-10-09
Add check k8s replication

mattmb:
* move black/whitelist checks to `long_running_service_tools`
* move `send_replication_event_if_under_replication` to `monitoring_tools`
* add check_kubernetes_services_replication script and necessary additions to `kubernetes_tools` to make it work

### DIFF
--- a/paasta_tools/check_kubernetes_services_replication.py
+++ b/paasta_tools/check_kubernetes_services_replication.py
@@ -1,0 +1,164 @@
+#!/usr/bin/env python
+# Copyright 2015-2016 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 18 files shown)


paasta_tools/check_kubernetes_services_replication.py
```diff
@@ -0,0 +1,164 @@
...
+    :param instance_config: an instance of KubernetesDeploymentConfig
+    :param smartstack_replication_checker: an instance of KubeSmartstackReplicationChecker
+    """
+    expected_count = instance_config.get_instances()
+    log.info("Expecting %d total tasks for %s" % (expected_count, instance_config.job_id))
+    proxy_port = get_proxy_port_for_instance(instance_config)
+
+    registrations = instance_config.get_registrations()
+    # if the primary registration does not match the service_instance name then
+    # the best we can do is check marathon for replication (for now).
```
Rob-Johnson: s/marathon/kubernetes

Rob-Johnson: ps: can we do better than this and figure out what to do with namespaces with multiple instances advertising to it?

mattmb: yeah, I still think we should do that but I guess it's best left to a separate ticket, it is tricky because we can have two instances advertising into the same namespace etc...

PR #2012, opened 2018-10-10
Ignore log lines whose timestamps are half microsecond to 60 seconds

huadongliu:
This is to workaround [the bug in isotime](https://github.com/gweis/isodate/blob/master/src/isodate/isotime.py#L127).

>>> isodate.parse_datetime('2018-09-22T16:14:59.9999994')
datetime.datetime(2018, 9, 22, 16, 14, 59, 999999)

>>> isodate.parse_datetime('2018-09-22T16:14:59.9999995')
 File "<stdin>", line 1, in <module>
 File "/nail/home/hliu/paasta7/.tox/py36/lib/python3.6/site-packages/isodate/isodatetime.py", line 56, in parse_datetime
 tmptime = parse_time(timestring)
 File "/nail/home/hliu/paasta7/.tox/py36/lib/python3.6/site-packages/isodate/isotime.py", line 133, in parse_time
 tzinfo)
ValueError: second must be in 0..59

>>> print(Decimal('59.9999995').quantize(Decimal('.000001')))
60.000000
>>> print(Decimal('59.9999994').quantize(Decimal('.000001')))
59.999999


paasta_tools/cli/cmds/logs.py
```diff
@@ -251,8 +251,10 @@ def paasta_app_output_passes_filter(
     except ValueError:
         log.debug('Trouble parsing line as json. Skipping. Line: %r' % line)
         return False
-
-    timestamp = isodate.parse_datetime(parsed_line.get('timestamp'))
+    try:
+        timestamp = isodate.parse_datetime(parsed_line.get('timestamp'))
+    except ValueError:
```
mattmb: Can you please leave a comment with the link to the github issue.
Also, should it be False? Isn't the timestamp technically in range?

huadongliu: I linked the issue. You are right that the timestamp IS in range. However, because the timestamp is not correctly parsed (due to the bug), it cannot proceed to component and cluster checks to give a definitive True. I think false positive is better than missing a log line. So changed to return True here.

PR #2013, opened 2018-10-11
convert cmd to a list for kubernetes deployments

Rob-Johnson:
kubernetes requires cmd to be a list of strings, rather than a string.
support splitting a string into a list, or even supplying a list as the
cmd

### DIFF
--- a/paasta_tools/frameworks/native_service_config.py
+++ b/paasta_tools/frameworks/native_service_config.py
@@ -243,7 +243,7 @@ class NativeServiceConfig(LongRunningServiceConfig):
                 ],
             },
             'command': {
-                'value': self.get_cmd(),
+                'value': str(self.get_cmd()),
                 'uris': [
                     {
                         'value': system_paasta_config.get_dockercfg_location(),
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -263,7 +263,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         cmd = super(LongRunningServiceConfig, self).get_cmd()
         if cmd:
             if isinstance(cmd, str):
-                return cmd.split(' ')
+                return ['sh', '-c', cmd]
             elif isinstance(cmd, list):
                 return cmd
             else:
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -259,6 +259,18 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             soa_dir=self.soa_dir,
         )
 
+    def get_cmd(self) -> Optional[List[str]]:
+        cmd = super(LongRunningServiceConfig, self).get_cmd()
+        if cmd:
+            if isinstance(cmd, str):
+                return cmd.split(' ')
+            elif isinstance(cmd, list):
+                return cmd
+            else:
+                raise ValueError("cmd should be str or list")
+        else:
+            return None
+
     def get_bounce_method(self) -> str:
         """Get the bounce method specified in the service's kubernetes configuration."""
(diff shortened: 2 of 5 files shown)


solarkennedy:
Wait wait wait, at the last minute we are converting to a string as we post, I don't think this is going to do what you think.
https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell

That array is like an `exec` array, it isn't in a shell. I don't think this is going to do what users expect.

```
["echo foo"]
```
will not work, because `echo foo` is not a binary.

Rob-Johnson:
we're not - we're converting to a list as we post.

`cmd: /bin/echo hi` in soa-configs becomes `['/bin/echo', 'hi;]` see https://github.com/Yelp/paasta/pull/2013/files#diff-8099076c2ecaf9f1a063335425abd3c1R262

it is only for marathon that we're going list -> string (the type of the function means that it now accepts `Union[str, list]` so I have to deal with the case, even if we don't make it publically available

paasta_tools/kubernetes_tools.py
```diff
@@ -259,6 +260,18 @@ def copy(self) -> "KubernetesDeploymentConfig":
             soa_dir=self.soa_dir,
         )
 
+    def get_cmd(self) -> Optional[List[str]]:
+        cmd = super(LongRunningServiceConfig, self).get_cmd()
+        if cmd:
+            if isinstance(cmd, str):
```
solarkennedy: I think i'm going to vote:

```
str?:
   return ['/bin/sh', '-c', cmd]
list?:
    return cmd
else:
    raise
```

paasta_tools/frameworks/native_service_config.py
```diff
@@ -243,7 +243,7 @@ def base_task(self, system_paasta_config: SystemPaastaConfig, portMappings=True)
                 ],
             },
             'command': {
-                'value': self.get_cmd(),
+                'value': str(self.get_cmd()),
```
solarkennedy: Can't this be:

```
'value': self.format_cmd()
```

And if the native thing needs to always be a string, then it's format_cmd method can do its own `str`?

EvanKrall:
I have a pretty strong distaste for `shlex`. It doesn't work 100% the same way as a shell does, so there's going to be situations that confuse users. (i.e. what happens when someone tries to use `$ENV_VAR`?)

I generally prefer:

1. using arrays all the way from config file to execve syscall, without any munging, or
2. if we _must_ accept a string and pass an array, wrapping the string in `['sh', '-c', cmd]`.

Rob-Johnson:
I think we can support both, but I think it'd be pretty rough on the user if their only option was to use yaml arrays in soa-configs. I'll add the `/bin/bash -c` and be done with it.

paasta_tools/kubernetes_tools.py
```diff
@@ -259,6 +259,18 @@ def copy(self) -> "KubernetesDeploymentConfig":
             soa_dir=self.soa_dir,
         )
 
+    def get_cmd(self) -> Optional[List[str]]:
+        cmd = super(LongRunningServiceConfig, self).get_cmd()
+        if cmd:
```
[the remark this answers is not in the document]
Rob-Johnson: if you don't set cmd, it'll come in as None rather than an empty string (unless you do `cmd: ''`

paasta_tools/utils.py
```diff
@@ -464,7 +464,7 @@ def get_gpus(self, default: float=0) -> float:
         gpus = self.config_dict.get('gpus', default)
         return gpus
 
-    def get_cmd(self) -> Optional[str]:
+    def get_cmd(self) -> Optional[Union[str, list]]:
```
EvanKrall:
```suggestion
    def get_cmd(self) -> Optional[Union[str, List[str]]]:
```

PR #2014, opened 2018-10-11
use healthcheck_mode before mode

Rob-Johnson:
right now we default to looking at 'mode' in smartstack.yaml for
what mode should be used for healthchecks. there is nothing in place
that one must include mode for a service, and healthcheck mode 'should'
be enough. ensure that we look at 'healthcheck_mode' before 'mode' when
figuring out the healthcheck for a service.

also fixes a bug that meant that healthcheck_cmd was ignored if it was
specified in smartstack.yaml rather than the instance config

Edit: I guess this raises the question of what having 2 keys is even for? outside of healthchecks, what do we care about the mode for?

### DIFF
--- a/paasta_tools/long_running_service_tools.py
+++ b/paasta_tools/long_running_service_tools.py
@@ -57,6 +57,17 @@ BounceMethodConfigDict = TypedDict('BounceMethodConfigDict', {"instances": int})
 
 class ServiceNamespaceConfig(dict):
 
+    def get_healthcheck_mode(self) -> str:
+        """Get the healthcheck mode for the service. In most cases, this will match the mode
+        of the service, but we do provide the opportunity for users to specify both. Default to the mode
+        if no healthcheck_mode is specified.
+        """
+        healthcheck_mode = self.get('healthcheck_mode', None)
+        if not healthcheck_mode:
+            return self.get_mode()
+        else:
+            return healthcheck_mode
+
     def get_mode(self) -> str:
         """Get the mode that the service runs in and check that we support it.
         If the mode is not specified, we check whether the service uses smartstack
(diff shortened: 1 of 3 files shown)


paasta_tools/long_running_service_tools.py
```diff
@@ -87,6 +98,9 @@ def is_in_smartstack(self) -> bool:
         else:
             return False
 
+    def get_healthcheck_cmd(self) -> Optional[str]:
```
solarkennedy: This isn't right. We don't allow people to specify a healthcheck_cmd in their smartstack config file.

I mean, I guess we could?

solarkennedy:
The `mode` is the smartstack mode really, and should *only* be used as a sane default if nothing else is specified. 

You are right though, it is a bug if we are ignoring healthcheck_mode in an instance config.

paasta_tools/long_running_service_tools.py
Rob-Johnson: oh, you're right, my mistake

paasta_tools/long_running_service_tools.py
```diff
@@ -57,6 +57,17 @@ class LongRunningServiceConfigDict(InstanceConfigDict, total=False):
 
 class ServiceNamespaceConfig(dict):
 
+    def get_healthcheck_mode(self) -> str:
```
solarkennedy: I'm really surprised this is in the ServiceNamespaceConfig. I thought `healthcheck_mode` was only something we let people set in their marathon config?

Rob-Johnson: it's in https://paasta.readthedocs.io/en/latest/yelpsoa_configs.html#smartstack-yaml ?

EvanKrall: Yeah, it's possible to have a separate healthcheck mode and mode for smartstack -- we can do just TCP checking for an HTTP smartstack service, or do HTTP checking for a TCP smartstack service.

PR #2018, opened 2018-10-15
Remove use_percent_format from Chronos schema

qui:
Final step in removing % format. No one has to know this ever existed! :)


solarkennedy:
Can you remove `paasta_tools/contrib/migrate_chronos_percent_format.py` if you want to scorch the sky?

PR #2020, opened 2018-10-16
Make all chronos jobs use --init instead of requiring dumb-init

solarkennedy:
I did this for tron, I think I want to do it for chronos. 

This won't require a big bounce or anything, but once this is deployed out 100%, we can remove the `exec dumb-init` from every chronos invocation.


Rob-Johnson:
are you sure it won't cause a big bounce of chronos jobs?

solarkennedy:
"Bouncing" chronos jobs just changes their next run configuration.

PR #2021, opened 2018-10-17
Fix `is_pod_ready()` always returns `True`

vkhromov:
`V1PodCondition.status` is a string, see
https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1PodCondition.md


vkhromov:
Actually, it's kind of ternary value as it can also be `"Unknown"` :).

PR #2029, opened 2018-10-22
Re-factor paasta cluster boost

mattmb:
I'd like to re-use the cluster boost logic to boost the service
autoscalers on a per-service level. Soon we will only need the
clusterman metrics bit for clusters anyway so I intend to separate that
block completely soon. This is just the refactor and I will add the
changes to the service autoscaler and boost cli in a second commit.

This:
* renames the module to load_boost.
* makes all the functions take a zk_boost_path which is {region}/{pool}
for cluster boost but can be like {cluster}/{service}/{instance} for the
service autoscaler
* adds some tests and tidy up

### DIFF
--- a/paasta_tools/paasta_cluster_boost.py
+++ b/paasta_tools/paasta_cluster_boost.py
@@ -16,7 +16,7 @@ import argparse
 import logging
 import sys
 
-from paasta_tools.autoscaling import cluster_boost
+from paasta_tools.autoscaling import load_boost
 from paasta_tools.utils import load_system_paasta_config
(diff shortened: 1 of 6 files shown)


paasta_tools/paasta_cluster_boost.py
```diff
@@ -114,13 +109,16 @@ def paasta_cluster_boost():
         return False
 
     for region in regions:
+        zk_boost_path = load_boost.get_zk_cluster_boost_path(region, pool)
```
solarkennedy: kwargs, it would be super easy to reverse these args

PR #2036, opened 2018-10-25
Service autoscaler performance stuff

mattmb:
Probably best to review the commits independently
* The first is a change to caching which I think is safe? Hopefully nothing long-running is going to break?
* The second makes it possible to run multiple autoscalers (e.g. using xargs). By default it will stick with the current behaviour of iterating through all the services sequentially

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -21,6 +21,8 @@ from contextlib import contextmanager
 from datetime import datetime
 from math import ceil
 from math import floor
+from typing import Optional
+from typing import Sequence
 
 import a_sync
 import gevent
@@ -45,18 +47,18 @@ from paasta_tools.marathon_tools import get_marathon_clients
 from paasta_tools.marathon_tools import get_marathon_servers
 from paasta_tools.marathon_tools import is_old_task_missing_healthchecks
 from paasta_tools.marathon_tools import is_task_healthy
-from paasta_tools.marathon_tools import load_marathon_service_config
+from paasta_tools.marathon_tools import MarathonServiceConfig
 from paasta_tools.marathon_tools import MESOS_TASK_SPACER
 from paasta_tools.mesos_tools import get_all_running_tasks
 from paasta_tools.mesos_tools import get_cached_list_of_running_tasks_from_frameworks
+from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.utils import _log
-from paasta_tools.utils import compose_job_id
 from paasta_tools.utils import DEFAULT_SOA_DIR
-from paasta_tools.utils import get_services_for_cluster
 from paasta_tools.utils import get_user_agent
+from paasta_tools.utils import list_services
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import mean
-from paasta_tools.utils import NoDeploymentsAvailable
+from paasta_tools.utils import SystemPaastaConfig
 from paasta_tools.utils import use_requests_cache
 from paasta_tools.utils import ZookeeperPool
 try:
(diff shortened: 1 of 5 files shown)


mattmb:
@EvanKrall I also had a crack at using both `ThreadPoolExecutor` and just adding async/await for only `autoscale_marathon_instance` (which I think would be a similar speedup but hopefully share the cache)

However, I couldn't make it work and ended up with `"Cannot run the event loop while another loop is running"` etc. I'm not quite sure what we're doing wrong but I'm too much of a python async noob to work it out!

paasta_tools/mesos/slave.py
```diff
@@ -36,6 +36,9 @@ def __getitem__(self, name):
     def __str__(self):
         return self.key()
 
+    def __hash__(self):
```
vkhromov: This is a noop for `@async_ttl_cache(..., cleanup_self=True)`, as we don't use hash for `self` in that case.

mattmb: We just discussed IRL and I think @vkhromov is right. It might be safe for me to disable `cleanup_self` for the slaves but I want to do a bit of testing to make sure...

vkhromov: As for `@async_ttl_cache(cleanup_self=False)`, it's not enough to add `__hash__()`, you also need to add `__eq__()` method to `MesosSlave`, otherwise there will be a memory leak again:
```
$ cat 3.py
#! /usr/bin/env python3.6

import functools

class WithoutEq(object):
    def __init__(self, q):
        self.q = q
    def __hash__(self):
        return 1

class WithEq(object):
    def __init__(self, q):
        self.q = q
    def __hash__(self):
        return 2
    def __eq__(self, other):
        return True

def k(self):
    return functools._make_key((self,), {}, typed=False)

def q(cls):
    d = dict()
    a1 = cls(1)
    a2 = cls(2)
    d[k(a1)] = 1
    d[k(a2)] = 2
    print(f'{d}')

q(WithoutEq)
q(WithEq)

$ ./3.py
{[<__main__.WithoutEq object at 0x7f605db61e48>]: 1, [<__main__.WithoutEq object at 0x7f605db61e80>]: 2}
{[<__main__.WithEq object at 0x7f605db61e80>]: 2}
```

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -565,80 +564,91 @@ def is_task_data_insufficient(marathon_service_config, marathon_tasks, current_i
 
 
 def autoscale_marathon_instance(marathon_service_config, system_paasta_config, marathon_tasks, mesos_tasks):
```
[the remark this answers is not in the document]
solarkennedy: How about in a second review?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -650,29 +660,24 @@ def humanize_error(error):
...
 
 
-def get_configs_of_services_to_scale(cluster, soa_dir=DEFAULT_SOA_DIR):
-    services = get_services_for_cluster(
-        cluster=cluster,
-        instance_type='marathon',
-        soa_dir=soa_dir,
-    )
+def get_configs_of_services_to_scale(cluster, soa_dir=DEFAULT_SOA_DIR, services=None):
+    if not services:
```
solarkennedy: I would prefer that this part (fetching the list of services) be handled higher up (heck, main?) and not have this function do it

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -694,39 +699,35 @@ def autoscaling_is_paused():
...
-    except LockHeldException:
-        log.warning("Skipping autoscaling run for services because the lock is held")
+    system_paasta_config = load_system_paasta_config()
+    cluster = system_paasta_config.get_cluster()
+    configs = get_configs_of_services_to_scale(cluster=cluster, soa_dir=soa_dir, services=services)
+
+    marathon_clients = get_marathon_clients(get_marathon_servers(system_paasta_config))
+    apps_with_clients = get_marathon_apps_with_clients(marathon_clients.get_all_clients(), embed_tasks=True)
+    all_mesos_tasks = a_sync.block(get_all_running_tasks)
+    if configs:
```
solarkennedy: Do we need to nest here? Can't configs just be `[]` if there isn't anything to autoscale?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -565,80 +564,91 @@ def is_task_data_insufficient(marathon_service_config, marathon_tasks, current_i
...
-                config=marathon_service_config,
-                line='Delaying scaling *down* as we found too few healthy tasks running in marathon. '
-                     'This can happen because tasks are delayed/waiting/unhealthy or because we are '
-                     'waiting for tasks to be killed. Will wait for sufficient healthy tasks before '
-                     'we make a decision to scale down.',
+    try:
+        with create_autoscaling_lock(marathon_service_config.service, marathon_service_config.instance):
+            current_instances = marathon_service_config.get_instances()
+            task_data_insufficient = is_task_data_insufficient(
+                marathon_service_config,
```
Rob-Johnson: kwargs?

paasta_tools/mesos/slave.py
mattmb: This is all correct! But I'm going to remove the `__hash__` method because I get the performance increase without it!

I worked out why:
* We call `task_stats` which calls `self.slave()` but this calls `self.master.slave(self["slave_id"])` and that call is also cached. So we only get one `MesosSlave` in the cache each time anyway.
* This leads me to conclude I don't need to add the `__hash__` or `__eq__`, I can just bump the TTL on the statistics cache.

paasta_tools/autoscaling/autoscaling_service_lib.py
mattmb: I guess it stops us creating a ZK connection if we have nothing to do. But that is cheap so I guess I'll remove it

paasta_tools/autoscaling/autoscaling_service_lib.py
mattmb: Yeah, I see what you mean, let me have a crack at that

paasta_tools/autoscaling/autoscaling_service_lib.py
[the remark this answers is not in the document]
vkhromov: Or just another one commit in the this review?

EvanKrall:
> @EvanKrall I also had a crack at using both `ThreadPoolExecutor` and just adding async/await for only `autoscale_marathon_instance` (which I think would be a similar speedup but hopefully share the cache)
> 
> However, I couldn't make it work and ended up with `"Cannot run the event loop while another loop is running"` etc. I'm not quite sure what we're doing wrong but I'm too much of a python async noob to work it out!

This happens when you try to run something that's decorated with `@a_sync.to_blocking` / `@a_sync.block` from within an async coroutine. To get around it, you need to `@a_sync.to_async` it again, or get at the underlying async coroutine.

```
import asyncio

import a_sync

@a_sync.to_blocking
async def a():
    await asyncio.sleep(1)
    return 5


def b():
    a()


async def c():
    a()

print("running a():")
a()
print("running b():")
b()
print("running c():")
a_sync.to_blocking(c)()
```
```
(py36) krall@dev20-uswest1cdevc:~/pg/paasta (master) $ python ./blurp.py
running a():
running b():
running c():
  File "./blurp.py", line 23, in <module>
    a_sync.to_blocking(c)()
  File "/nail/home/krall/pg/paasta/.tox/py36/lib/python3.6/site-packages/a_sync/a_sync.py", line 156, in blocking
    return loop.run_until_complete(async_func(*args, **kwargs))
  File "/usr/lib/python3.6/asyncio/base_events.py", line 466, in run_until_complete
    return future.result()
  File "./blurp.py", line 16, in c
    a()
  File "/nail/home/krall/pg/paasta/.tox/py36/lib/python3.6/site-packages/a_sync/a_sync.py", line 156, in blocking
    return loop.run_until_complete(async_func(*args, **kwargs))
  File "/usr/lib/python3.6/asyncio/base_events.py", line 454, in run_until_complete
    self.run_forever()
  File "/usr/lib/python3.6/asyncio/base_events.py", line 411, in run_forever
    'Cannot run the event loop while another loop is running')
RuntimeError: Cannot run the event loop while another loop is running
sys:1: RuntimeWarning: coroutine 'a' was never awaited
```

mattmb:
I see @EvanKrall. I'm going to merge this and see if I can get a happy balance between caching/parallelising with the dumb xargs option. If not I'll double back and try to async and un-async allthethings...

PR #2040, opened 2018-10-29
Ignore failure to get IP for maintenance code

mattmb:
I think it is okay to swallow this failure and just log. I am assuming
mesos will just skip/ignore any schedules with null IPs. Ideally we
should bubble this up and raise an error. I have a stashed diff which
would do that but it ends up being a big change that I'd prefer to
avoid...

What do you reckon @drmorr0 ? I guess the downside with this method is that it could fail without us noticing for real hosts (if say we had a real DNS issue).


paasta_tools/mesos_maintenance.py
```diff
@@ -377,7 +377,11 @@ def hostnames_to_components(hostnames, resolve=False):
             (host, ip) = hostname.split('|')
             components.append(Hostname(host=host, ip=ip))
         else:
-            ip = gethostbyname(hostname) if resolve else None
+            try:
+                ip = gethostbyname(hostname) if resolve else None
+            except Exception:
+                ip = None
+                log.error(f"Failed to resolve IP for {hostname}, continuing regardless")
```
solarkennedy: Wait wait, are you sure you want to append still though?

mattmb: You're right, no need to actually append. Let me fix this up!

PR #2042, opened 2018-10-31
Add the Mesos containerizer to Marathon

huadongliu:
I ran setup_marathon_job.py dsc-test-service.gpu manually with below patch on a mesos master. The task can grab 1 gpu and run in a mesos containerizer.

V2: 
*Removed containerizer type from config. Instead, it is inferred depending on if gpu is used.
*Changed gpus type from float to int.

### DIFF
--- a/paasta_tools/cli/schemas/marathon_schema.json
+++ b/paasta_tools/cli/schemas/marathon_schema.json
@@ -80,6 +80,11 @@
                     "exclusiveMinimum": true,
                     "default": 1024
                 },
+                "gpus": {
+                    "type": "integer",
+                    "minimum": 0,
+                    "exclusiveMinimum": false
+                },
                 "instances": {
                     "type": "integer",
                     "minimum": 0,
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -277,6 +277,7 @@ class FormattedMarathonAppDict(BounceMethodConfigDict, total=False):
     env: Dict[str, str]
     mem: float
     cpus: float
+    gpus: int
     disk: float
     constraints: List[Constraint]
     cmd: str
@@ -658,8 +659,8 @@ class MarathonServiceConfig(LongRunningServiceConfig):
 
         net = get_mesos_network_for_net(self.get_net())
 
-        complete_config: FormattedMarathonAppDict = {
-            'container': {
+        if self.get_container_type() == 'DOCKER':
+            container_dict: MarathonContainerInfo = {
                 'docker': {
                     'image': docker_url,
                     'network': net,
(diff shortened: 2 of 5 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -657,17 +658,27 @@ def format_marathon_app_dict(self) -> FormattedMarathonAppDict:
...
+                'network': net,
+                "parameters": self.format_docker_parameters(),
+            },
+            'type': 'DOCKER',
+            'volumes': docker_volumes,
+        }
+        # Only image and forcePullImage are supported: "Mesos Containerizer and Universal Container Runtime"
+        # in http://mesosphere.github.io/marathon/1.4/docs/native-docker.html
+        if self.get_containerizer_type() == 'MESOS':
+            container_dict = {  # type: ignore
```
solarkennedy: Can you change this to do a normal if/then with the container_type instead of mutating `container_dict` ?

paasta_tools/marathon_tools.py
```diff
@@ -679,6 +690,7 @@ def format_marathon_app_dict(self) -> FormattedMarathonAppDict:
             'env': self.get_env(),
             'mem': float(self.get_mem()),
             'cpus': float(self.get_cpus()),
+            'gpus': float(self.get_gpus()),
```
solarkennedy: This will cause a big bounce, right?

huadongliu: @solarkennedy Good catch! Also, marathon needs to be restarted to have the --enable_features gpu_resources option. I will make this conditional.

paasta_tools/cli/schemas/marathon_schema.json
```diff
@@ -214,6 +219,10 @@
                 "container_port": {
                     "type": "number"
                 },
+                "containerizer_type": {
```
Rob-Johnson: nit: I think 'containerizer' is enough

paasta_tools/marathon_tools.py
```diff
@@ -277,6 +277,7 @@ class FormattedMarathonAppDict(BounceMethodConfigDict, total=False):
     env: Dict[str, str]
     mem: float
     cpus: float
+    gpus: float
```
Rob-Johnson: can you really have a float? does that even work?

paasta_tools/cli/schemas/marathon_schema.json
huadongliu: yeah, it worked, look for dsc-test-service.gpu in pnw-devc. I am using float here hoping gpus can be shared in the near future, i.e. a task can have 0.5 gpus.

paasta_tools/marathon_tools.py
huadongliu: Yeah, it works. Look for dsc-test-service.gpu in pnw-devc. I am using float here hoping gpus can be shared in the near future, i.e. a task can have 0.5 gpus.

paasta_tools/marathon_tools.py
```diff
@@ -689,7 +702,11 @@ def format_marathon_app_dict(self) -> FormattedMarathonAppDict:
             'args': self.get_args(),
         }
 
-        if net == 'BRIDGE':
+        if complete_config['container']['type'] == 'MESOS' and self.get_gpus() is not None:
```
keymone: we should probably show a warning to the user when containerizer is DOCKER and gpus is set? and the same with containerizer MESOS  and net==BRIDGE?

paasta_tools/marathon_tools.py
```diff
@@ -658,16 +659,28 @@ def format_marathon_app_dict(self) -> FormattedMarathonAppDict:
 
         net = get_mesos_network_for_net(self.get_net())
 
-        complete_config: FormattedMarathonAppDict = {
-            'container': {
+        if self.get_gpus() is None:
```
solarkennedy: Even though this is one line, can you extract this out into a `get_container_type()` function?
I promise it will be useful in the future. No need for a config at this point, just make that function do the gpu logic at at first.

PR #2043, opened 2018-10-31
confirm affected cluster and instances before really start or stop

chlgit:
Before the instances are actually started/stopped, the command line would ask for the confirmation. An example is [here](https://fluffy.yelpcorp.com/i/dcmP9ThnSmN4CdXM6kDKwW6dsdNvxj4L.html).

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -140,6 +140,19 @@ def print_chronos_message(desired_state):
         )
 
 
+def confirm_to_continue(cluster_service_instances, desired_state):
+    paasta_print(f'You are about to {desired_state} the following instances:')
+    i_count = 0
+    for cluster, services_instances in cluster_service_instances:
+        for service, instances in services_instances.items():
+            for instance in instances.keys():
+                paasta_print(f'cluster = {cluster}, instance = {service}')
+                i_count += 1
+    if sys.stdin.isatty():
+        return choice.Binary(f'Are you sure you want to {desired_state} these {i_count} instances?', False).ask()
+    return False
+
+
 def paasta_start_or_stop(args, desired_state):
     """Requests a change of state to start or stop given branches of a service."""
     soa_dir = args.soa_dir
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -140,6 +140,19 @@ def print_chronos_message(desired_state):
...
 
+def confirm_to_continue(cluster_service_instances, desired_state):
+    paasta_print(f'You are about to {desired_state} the following instances:')
+    i_count = 0
+    for cluster, services_instances in cluster_service_instances:
+        for service, instances in services_instances.items():
+            for instance in instances.keys():
+                paasta_print(f'cluster = {cluster}, instance = {service}')
+                i_count += 1
+    if sys.stdin.isatty():
```
solarkennedy: The implicit `else` here kinda concerns me.
I think we need to default to true if we can't ask the tty. Or have a --yes option.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -140,6 +140,19 @@ def print_chronos_message(desired_state):
         )
 
 
+def confirm_to_continue(cluster_service_instances, desired_state):
+    paasta_print(f'You are about to {desired_state} the following instances:')
+    i_count = 0
+    for cluster, services_instances in cluster_service_instances:
+        for service, instances in services_instances.items():
+            for instance in instances.keys():
+                paasta_print(f'cluster = {cluster}, instance = {service}')
```
qui: I think you mean {instance}, not {service} (look at the example output again)

chlgit: lol

chlgit:
Only ask for confirmation if cluster or instance is None.

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -169,6 +182,13 @@ def paasta_start_or_stop(args, desired_state):
 
     invalid_deploy_groups = []
     marathon_message_printed, chronos_message_printed = False, False
+
+    if args.clusters is None or args.instances is None:
```
solarkennedy: Thank you. Not only will this help make the tests pass, but it will be less annoying to users and scripts

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -140,6 +140,19 @@ def print_chronos_message(desired_state):
         )
 
 
+def confirm_to_continue(cluster_service_instances, desired_state):
+    paasta_print(f'You are about to {desired_state} the following instances:')
```
solarkennedy: Can you also add
```
print("Either --instances or --clusters not specified. Asking for confirmation.")
```
To educate the user and explain *why* we are asking?

PR #2044, opened 2018-11-06
[tron] Add support for trigger_timeout attribute

keymone:


solarkennedy:
don't forget the json schema?

solarkennedy:
We need to be careful to wait for a tron upgrade before merging this + paasta release, right?

PR #2045, opened 2018-11-06
RFC: Added tron support in the paasta api

solarkennedy:
This is an experimental branch to see what it would take to get tron stuff in `paasta status`.

I think I want `paasta status` to be able see tron stuff for a couple reasons:
* Gets users to "feel" like they are interacting with paasta?
* Understands the `--cluster` argument instead of requiring users to "be" in the right place

But there are some serious downsides:
* The concept of looking at a whole "job" doesn't match up with the status api (which looks at individual actions
* The concept of a run_id doesn't fit anywhere in the paasta api?
* Idk, I just feel like it would take a lot of work to get feature parity with tronview.

Is this the direction you think we should go? (getting paasta status with feature parity with tronview?)

### DIFF
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -563,15 +563,19 @@
         },
         "marathon": {
           "$ref": "#/definitions/InstanceStatusMarathon",
-          "description": "Marathon specific instance status"
+          "description": "Marathon instance status"
         },
         "kubernetes": {
           "$ref": "#/definitions/InstanceStatusKubernetes",
-          "description": "Kubernetes specific instance status"
+          "description": "Kubernetes instance status"
         },
         "chronos": {
           "$ref": "#/definitions/InstanceStatusChronos",
-          "description": "Chronos specifid instance status"
+          "description": "Chronos instance status"
+        },
+        "tron": {
+          "$ref": "#/definitions/InstanceStatusTron",
+          "description": "Tron instance status"
         }
       }
     },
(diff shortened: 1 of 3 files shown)


qui:
Agreed, my main goal is for users to feel like they are interacting with paasta. I don't want them to forget that their jobs exist 🙂If Tron jobs are in the service directory, they should show up in paasta status.

I don't think we need to get to feature parity with tronview, just something similar to Chronos. For each action, we can show what job it is part of, the most recent state of the action, and the last job run ID for that state. Maybe the job state/schedule too?

Then we can print a tronview command or tronweb link if they want to see the whole job.

qui:
How about something like this? This shows one job with two actions. There's some duplication but I don't think it's too bad.
```
service: redshift_to_parquet_spark
cluster: norcal-devc
    instance: daily_dev_smoke_test.ping_cluster_to_run
    Git sha:    de5339e4 (desired)
    Tron job: redshift_to_parquet_spark.daily_dev_smoke_test
      Status: Enabled    Last: running (2018-11-20 05:30:00)
      Schedule: daily 05:30:00
      Dashboard: http://tron1-uswest1cdevc.dev.yelpcorp.com:8089/web/#job/redshift_to_parquet_spark.daily_dev_smoke_test/148
    Action: ping_cluster_to_run
      Status:   running
      Start time: 2018-11-20 05:30:00
      Bare Command: log_done {shortdate}; sleep 1200
      Command: log_done 2018-11-20; sleep 1200

    instance: daily_dev_smoke_test.queue_job_to_control_plane
    Git sha:    de5339e4 (desired)
    Tron job: redshift_to_parquet_spark.daily_dev_smoke_test
      Status: Enabled    Last: running (2018-11-20 05:30:00)
      Schedule: daily 05:30:00
      Dashboard: http://tron1-uswest1cdevc.dev.yelpcorp.com:8089/web/#job/redshift_to_parquet_spark.daily_dev_smoke_test/148
    Action:     queue_job_to_control_plane
      Status:   queued
      Start time: None
      Bare Command: echo done
      Command:
```

Maybe we could go even further, when verbose isn't set:
```
service: redshift_to_parquet_spark
cluster: norcal-devc
    instance: daily_dev_smoke_test.ping_cluster_to_run
    Tron job: redshift_to_parquet_spark.daily_dev_smoke_test
      Run: http://tron1-uswest1cdevc.dev.yelpcorp.com:8089/web/#job/redshift_to_parquet_spark.daily_dev_smoke_test/148
    Action: ping_cluster_to_run
      Status:   running (started 2018-11-20 05:30:00)
      Bare Command: log_done {shortdate}; sleep 1200

    instance: daily_dev_smoke_test.queue_job_to_control_plane
    Tron job: redshift_to_parquet_spark.daily_dev_smoke_test
      Run: http://tron1-uswest1cdevc.dev.yelpcorp.com:8089/web/#job/redshift_to_parquet_spark.daily_dev_smoke_test/148
    Action:     queue_job_to_control_plane
      Status:   queued
      Bare Command: echo done
```

solarkennedy:
Thank you @qui, this is good stuff :)

I'll try that out. For this first pass, in order to get that data in I'm going to data this API returns include things like a dashboard link and the job/action object, to make it easier for the "frontend" (different PR) to display what you have there.

PR #2046, opened 2018-11-07
RFC: Address the identity=0 race condition by managing IP addresses ourselves

danielpops:
Sending this out as a request for comments on how good/bad/okay of an idea this is.

The idea is that we do our own IP address management and tell docker to use the IP address we choose, so that we can have a chance to update the mapfile before actually launching the container. If we wanted to go this way, we'd probably want to stop generating the mapfile out of band. But, we'd also need something to manage "garbage collection" of old entries in the mapfile over time, which could be done but introduces more locking/synchronization semantics between more things.

There are definitely a few things to think through here (for e.g. will we need changes to our docker networking infra to enable this?), but this seems like it could do the trick. Thoughts?

### DIFF
--- a/paasta_tools/docker_wrapper.py
+++ b/paasta_tools/docker_wrapper.py
@@ -174,7 +174,7 @@ def get_ip_for_container():
     return new_ip
 
 
-def add_ip_address(argv, service, instance):
+def add_ip_address(argv)
     ip_addr = get_ip_for_container()
     # Only add the --ip argument if we actually got an IP
     if ip_addr:
--- a/paasta_tools/docker_wrapper.py
+++ b/paasta_tools/docker_wrapper.py
@@ -26,6 +26,7 @@ from ipaddress import ip_network
 from paasta_tools.firewall import DEFAULT_SYNAPSE_SERVICE_DIR
 from paasta_tools.firewall import firewall_flock
 from paasta_tools.firewall import prepare_new_container
+from paasta_tools.ip import ip_flock
 from paasta_tools.mac_address import reserve_unique_mac_address
 from paasta_tools.utils import DEFAULT_SOA_DIR
 
@@ -160,11 +161,20 @@ def get_ip_for_container():
     subnet = ip_network(u'169.254.1.1/20', False)
     hosts = list(subnet.hosts())
     new_ip = random.choice(hosts)
-    while new_ip in current_ip_addressess:
+    while new_ip in current_ip_addresses:
         # TODO: Infinite loop?
         new_ip = random.choice(hosts)
 
     return new_ip
+
+
+def add_ip_address(argv, service, instance):
+    ip_addr = get_ip_for_container()
+    # Only add the --ip argument if we actually got an IP
+    if ip_addr:
+        argv = add_argument(argv, f'--ip={ip_addr}')
+    }
+    return argv
 
 
 def generate_hostname(fqdn, mesos_task_id):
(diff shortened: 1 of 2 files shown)


paasta_tools/docker_wrapper.py
```diff
@@ -330,4 +363,27 @@ def main(argv=None):
...
+        try:
+            # Need cross-process locking semantics to avoid race condition where
+            # multiple paasta invocations try to acquire the same IP address
+
+            # get_lock()
+
+            ip_addr = get_ip_for_container()
+            # Only add the --ip argument if we actually got an IP
+            if ip_addr:
+                argv = add_argument(argv, f'--ip={ip_addr}')
```
ronin13: Does this work with mesos? From docker docs, the --ip is supported only for user-defined networks.

There is a similar precedent in paasta code where --mac-addres is added  in paasta firewall: https://github.com/Yelp/paasta/blob/master/paasta_tools/docker_wrapper.py#L289

But I don't think the IP address can be pinned unless mesos runs a user-defined network.

danielpops: Hrmmm I didn’t consider any meals implications. Does it matter? We are completely controlling the `docker run` Invocation and the docker daemon, and media currently doesn’t have any influence on what IP address is chosen (it’s randomly assigned by docker run)

We would need to update our puppet docker daemon to use a custom bridge network, which I think should be fine, but I don’t really have any way of knowing how reasonable that is

paasta_tools/docker_wrapper.py
```diff
@@ -134,6 +138,51 @@ def can_add_mac_address(args):
...
+        if arg.startswith('--ip'):
+            return False
+
+    return True
+
+
+def get_ip_for_container():
+    # This would look at the currently outstanding IP addresses
+    # Something like but not necessarily =>
+    current_ip_addresses = subprocess.check_output((
```
danielpops: I feel like this is certainly better to be done in python, but I had this handy and I don't know if the 3 year old, 2 major versions behind of `docker-py` the repo pins to has this functionality.

paasta_tools/docker_wrapper.py
```diff
@@ -134,6 +139,51 @@ def can_add_mac_address(args):
...
+            "sudo docker network inspect $i; "
+        "done "
+            "| jq -r '.[].Containers | to_entries | .[].value.IPv4Address' "
+            "| sed 's/\/.*$//g'"
+    ), shell=True).split()
+    # Then try to get one in the subnet of the docker
+    # Ideally get this from config etc
+    subnet = ip_network(u'169.254.16.0/24', False)
+    hosts = list(subnet.hosts())
+    new_ip = random.choice(hosts)
```
solarkennedy: You can just use sets and subtract: `hosts - current_ip_addresses` and then random choice from the leftovers

solarkennedy:
Are you sure there is no way to re-use the existing paasta-firewalld thing to achieve what you want?
Granted, it picks and allocates *mac* addresses, but it could also do ips. We already have that inotify-based daemon that write iptables rules, but it could also write down the map file?

As discussed in our meeting, I would like you to think about doing this at the client-library level using our existing puppet SSL pki as an alternative, because it solves lots of other "holes" in this approach (at the cost of making it more complex for clients to verify the caller).

solarkennedy:
@danielpops are you going to continue with this or discard?

solarkennedy:
@danielpops I'm going to close this PR due to inactivity. Please rebase and reopen if you want to continue with this approach.

PR #2047, opened 2018-11-13
Enable docker --init for all paasta things by default

solarkennedy:
@chriskuehl 
This is so we can enable this feature, but also opt out without causing a big bounce.


solarkennedy:
I'm discarding this in favor of the `ENTRYPOINT` approach.

solarkennedy:
Re-opening again. The Entrypoint approach isn't ideal because:
1. Quite a few of our services override ENTRYPOINT anyway
2. Marathon overrides the ENTRYPOINT to /bin/sh if you use `cmd` (!)
3. I don't want to make users have to convert their cmd -> args in array format, I would rather they just "remember" to use dumb-init over that.

PR #2051, opened 2018-11-20
Read Marathon and Chronos logs as strings not bytes

kawaiwanyelp:
### Description
- Sometimes logs are read in as byte strings, but we attempt to decode them as strs, which can result in errors.
- This change checks if log lines are bytes; if they are, we convert them to strs first.

### Testing done
- manual testing

### Reviewers
- According @solarkennedy, @EvanKrall may have strong opinions about this. However, given that some people need this, I suggest making this change, at least temporarily, until we can come to an agreement about how to do this better going forward.

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -788,7 +788,7 @@ class ScribeLogReader(LogReader):
                 for line in scribe_reader:
                     # temporary until all log lines are strings not byte strings
                     if isinstance(line, bytes):
-                        line = line.decode()
+                        line = line.decode('utf-8')
                     if parser_fn:
                         line = parser_fn(line, clusters, service)
                     if filter_fn:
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/logs.py
```diff
@@ -786,6 +786,9 @@ def filter_and_aggregate_scribe_logs(
         with scribe_reader_ctx as scribe_reader:
             try:
                 for line in scribe_reader:
+                    # temporary until all log lines are strings not byte strings
+                    if isinstance(line, bytes):
+                        line = line.decode()
```
solarkennedy: `decode('utf8')` ?

kawaiwanyelp: It's 'utf-8' by default, but I guess an explicit arg is better. I'll make the change.

PR #2053, opened 2018-11-21
Allow underscores in instance configs for marathon/chronos/adhoc

solarkennedy:
This has been undocumented and not allowed via the json schema for a while.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -1,9 +1,35 @@
 Preparation: paasta_tools and yelpsoa-configs
 =========================================================
 
-
 paasta_tools reads configuration about services from several YAML
 files in `soa-configs <soa_configs.html>`_:
+
+Each object inside of these YAML files is called an "instance" of a PaaSTA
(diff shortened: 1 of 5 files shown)


docs/source/yelpsoa_configs.rst
```diff
@@ -1,10 +1,35 @@
...
-
 paasta_tools reads configuration about services from several YAML
 files in `soa-configs <soa_configs.html>`_:
 
+Each object inside of these YAML files is called an "instance" of a PaaSTA
+service. It describes a unique way to run the service, with a unique command,
+cpu and ram requirements, etc.
+
+Duplication can be reduced by using YAML `anchors and merges<https://gist.github.com/bowsersenior/979804>`_.
+Any object in the YAML prefixed with ``_`` will be ignored.
```
mattmb: I'd be more specific, something like: "PaaSTA will not attempt to run any instances in the YAML prefixed with ``_`` but you can still reference them as templates with YAML anchors"

solarkennedy: Done, I like the active voice.

PR #2054, opened 2018-11-26
[WIP] add some api endpoints that return k8s objects

mattmb:
This is a bit speculative but mostly to demonstrate how this would work if we think it's helpful. The idea is that our k8s operator could just post a paastay looking instance config and get ready to use k8s json back.

I've added a few endpoints as an example that allow you to get a `Deployment` a `PodTemplateSpec` or a list of `V1Volume`s

Here's what it looks like in practice:
```
$ cat thing.json 
{
  "cpus": 1,
  "deploy_group": "testcluster.everything",
  "disk": 40,
  "instances": 2,
  "mem": 101,
  "container_image": "something...",
  "registrations": ["hello-world.main"]
}
$ curl -s -XPOST -d@thing.json localhost:5054/v1/services/hello-world/thing/kube_config | jq .
{
  "api_version": "apps/v1",
  "kind": "Deployment",
  "metadata": {
    "annotations": null,
...
```

### DIFF
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -71,6 +71,21 @@ def make_app(global_config=None):
     config.include('pyramid_swagger')
     config.add_route('resources.utilization', '/v1/resources/utilization')
     config.add_route('service.instance.status', '/v1/services/{service}/{instance}/status')
+    config.add_route(
+        'service.instance.kube_config',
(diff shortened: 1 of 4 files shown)


mattmb:
Don't think we're going to go with this now. I'll leave the branch around just in case but closing the PR.

PR #2057, opened 2018-11-27
Add hacheck sidecar Dockerfile

mattmb:
This is an example to use as a sidecar in k8s. See code here:
https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes_tools.py#L347-L364

cc @Rob-Johnson @vkhromov


vkhromov:
👍

PR #2063, opened 2018-11-29
Bump vault-tools to get newlines at end of secrets files

EvanKrall:
pre-commit end-of-file newline fixer keeps wanting to modify paasta secrets. @Qmando made vault-tools add a newline to the end of the secrets files back in June, but we're running a version from March. This bumps to the latest version.


KevinHock:
I pinged @mattmb outside of GitHub, but we might want to revert this until `paasta secret add` works.

KevinHock:
(Nevermind with respect to urgency, sort of, because there is a workaround.)

PR #2065, opened 2018-11-29
Add setup custom resource script

mattmb:
Pushing for early feedback. This is a script that allows us to specify a
file prefix like flinkcluster- or cassandra- etc. and map it to a
kubernetes custom object like FlinkCluster. For now it just adds the
neccessary metadata and submits the instances YAML config as the spec for
the k8s object. Like other paasta things we sha and check the sha to see
if we need to update.

Unit tests to follow if we think this is the way to go.

### DIFF
--- a/paasta_tools/setup_crd.py
+++ b/paasta_tools/setup_crd.py
@@ -31,9 +31,10 @@ from typing import Sequence
 import service_configuration_lib
 
 from paasta_tools.kubernetes_tools import create_custom_resource
-from paasta_tools.kubernetes_tools import ensure_paasta_namespace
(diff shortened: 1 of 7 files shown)


paasta_tools/setup_crd.py
```diff
@@ -0,0 +1,250 @@
...
+    for service, config in config_dicts.items():
+        if not reconcile_kubernetes_resource(
+            kube_client=kube_client,
+            service=service,
+            instance_configs=config,
+            kind=kind,
+            custom_resources=crs,
+            version=version,
+        ):
+            succeded = False
```
vkhromov: Either `break` for `False` or collect results in a list and then `return all(...) ...`

PR #2067, opened 2018-12-01
Constrain the maximum cpu a Spark executor can use

huadongliu:
Tested with a spark run with --spark-args "spark.cores.max=10 spark.executor.cores=5" and make sure the docker run command on the mesos agent has "--cpus=5".


solarkennedy:
Crazy that this doesn't happen automatically!

PR #2078, opened 2018-12-12
Bump vault_tools

mattmb:
Fixes token error on `paasta secret add` caused by incompatible return
type from the previous version of vault_tools...


filipposc5:
:shipit:

PR #2079, opened 2018-12-17
Reorganize remote-run args and make paasta_remote_run respect `--dry-run`

kawaiwanyelp:
### Description
- Currently, `remote-run` args are sort of wonky:
    - Some args do not and cannot work (e.g. `-y`, which cannot work since `paasta remote-run` SSHs to a master to run a task, and so cannot take local soa-configs with it)
    - Some args do not work but should (e.g. `--dry-run`)
    - Some args are internal, but are exposed via `paasta remote-run` (e.g. `--aws-region`, which determines which dynamodb table we persist to; this isn't a user concern, however)
- In this PR, I have:
    - Reorganized arguments so that actions only have the options specific to them instead of sharing some irrelevant ones
    - Moved 'internal' options to `paasta_remote_run`, which is 'called' from `paasta remote-run`, but isn't meant to be publicly used (like `-y`, `--debug`, etc.)
   - Make `--dry-run` work

### Tests
- manual testing (since no unit and integration tests currently exist - that's a separate issue)

### DIFF
--- a/paasta_tools/cli/cmds/remote_run.py
+++ b/paasta_tools/cli/cmds/remote_run.py
@@ -15,7 +15,6 @@
 import json
 import re
 from functools import lru_cache
-from shlex import quote
 
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_clusters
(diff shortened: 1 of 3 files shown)


kawaiwanyelp:
I'm closing this PR in favor of #2083

PR #2082, opened 2018-12-18
increase timeout when talking to git

chlgit:
It is taking longer than 20s in yelp-main, since the repo is so big. Increasing timeout should be helpful.
More details are [here](https://jira.yelpcorp.com/browse/PAASTA-15089).


ssk2:
:shipit:

PR #2083, opened 2018-12-18
Reorganize remote-run args and make paasta_remote_run respect `--dry-run`

kawaiwanyelp:
### Description
- Currently, `remote-run` args are sort of wonky:
    - Some args do not and cannot work (e.g. `-y`, which cannot work since `paasta remote-run` SSHs to a master to run a task, and so cannot take local soa-configs with it)
    - Some args do not work but should (e.g. `--dry-run`)
    - Some args are internal, but are exposed via `paasta remote-run` (e.g. `--aws-region`, which determines which dynamodb table we persist to; this isn't a user concern, however)
- In this PR, I have:
    - Reorganized arguments so that actions only have the options specific to them instead of sharing some irrelevant ones
    - Moved 'internal' options to `paasta_remote_run`, which is 'called' from `paasta remote-run`, but isn't meant to be publicly used (like `-y`, `--debug`, etc.)
   - Make `--dry-run` work

### Tests
- manual testing (since no unit and integration tests currently exist - that's a separate issue)
- This is duplicate of #2079, minus the changes that made `make itest_trusty` and `make itest_xenial` fail

### DIFF
--- a/paasta_tools/paasta_remote_run.py
+++ b/paasta_tools/paasta_remote_run.py
@@ -34,21 +34,22 @@ from task_processing.runners.sync import Sync
 from task_processing.task_processor import TaskProcessor
 
 from paasta_tools import mesos_tools
-from paasta_tools.cli.cmds.remote_run import add_common_args_to_parser
-from paasta_tools.cli.cmds.remote_run import add_start_args_to_parser
+from paasta_tools.cli.cmds.remote_run import add_list_parser
+from paasta_tools.cli.cmds.remote_run import add_start_parser
+from paasta_tools.cli.cmds.remote_run import add_stop_parser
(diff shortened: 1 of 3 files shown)


paasta_tools/paasta_remote_run.py
```diff
@@ -230,6 +232,15 @@ def paasta_to_task_config_kwargs(
     return kwargs
 
 
+def task_config_to_dict(task_config):
```
solarkennedy: Maybe this could be the `__repr__` of a TASK_CONFIG_INTERFACE or something?

kawaiwanyelp: I tried that but because `task_config` is a PRecord, it doesn't give me a dict that I can pretty print.

PR #2087, opened 2018-12-27
make sure spark app names are unique even when the default is overrid…

stug:
…den by the user

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -391,15 +391,14 @@ def load_aws_credentials_from_yaml(yaml_file_path):
 
 def get_spark_config(
     args,
-    container_name,
     spark_ui_port,
     docker_img,
     system_paasta_config,
     volumes,
 ):
-    # User configurable Spark options
+    # Defaults for user configurable Spark options
     user_args = {
-        'spark.app.name': container_name,
+        'spark.app.name': 'paasta_spark_run_{}'.format(get_username()),
         'spark.cores.max': '4',
         'spark.executor.cores': '2',
         'spark.executor.memory': '4g',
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -391,15 +391,14 @@ def load_aws_credentials_from_yaml(yaml_file_path):
...
     spark_ui_port,
     docker_img,
     system_paasta_config,
     volumes,
 ):
-    # User configurable Spark options
+    # Defaults for user configurable Spark options
     user_args = {
-        'spark.app.name': container_name,
+        'spark.app.name': 'paasta_spark_run_{}'.format(get_username()),
```
huadongliu: If an user does not set app name, and has multiple spark-runs, these spark frameworks end up with the same name, right? e.g. in tron, paasta_spark_run_batch.

stug: See below at line 501 -- I update this later by adding back the ui_port and timestamp.

More or less, I want to add the ui_port and timestamp to all app names.  By only setting `paasta_spark_run_$user` here, that's the only part of the app name that can be overridden by user args.

Maybe it'd be more clear to set `spark.add.name` to `None` here and then change the bit below to
```
user_args['spark.app.name'] = '{}_{}_{}'.format(
    user_args['spark.app.name'] or 'paasta_spark_run_{}'.format(get_username()),
    spark_ui_port,
    int(time.time()),
)
```
so that all the logic happens in one place?  In fact, I like that better, will change it.

huadongliu: Yeah, I like the change to have all the logic in one place. Is appending spark_ui_port good enough (i.e. time is not needed)? spark_ui_port is seeded with service name + pid.

stug:
Closing this in favor of the changes in #2089 (using webui_url instead of framework name to identify spark jobs), which make these unnecessary.

PR #2088, opened 2018-12-27
remove mesos principal and mesos secret arguments from spark-run -- u…

stug:
…sers should not be able to set these


huadongliu:
Please fix the test.

stug:
@huadongliu test is fixed

PR #2090, opened 2019-01-02
Move authors back above the thread in mark-for-deployment's Slack notifications

chriskuehl:
(Follow-up to #1959)

Quite a few people have expressed that having the "Authors" ping inside a thread can be annoying. I agree -- while it keeps the notifications in the main channel concise, it makes it hard to dismiss the notifications (you have to actually open the thread in Slack).

Let's try moving the authors list back into the original message, while keeping the rollback instructions, Jenkins links, etc. in the thread?

It looks like this:

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -278,9 +278,11 @@ class SlackDeployNotifier:
                 self._notify_with_thread(
                     'notify_after_mark',
                     self.channels,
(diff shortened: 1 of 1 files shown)


chriskuehl:
Tested and confirmed that you only get one notification (for the initial ping), and not another immediately afterwards when the bot posts the followup in the thread. It doesn't appear that the people pinged will automatically "follow" the thread (which is desired in this case to avoid multiple pings).

solarkennedy:
can you fix the flake8 stuff?

chriskuehl:
@solarkennedy should be fixed now

PR #2091, opened 2019-01-03
Don't update the contents of services yaml file when the time changes

solarkennedy:
We don't really need to do this, and it causes staticconf to constantly look at this file. Internal ticket is PERF-1672.

In practice this won't really get us anything because of min_interval, but it is slightly better.

cc @thebostik


solarkennedy:
Huh. You are right!

PR #2093, opened 2019-01-04
Make vault key settable

mattmb:
This makes it possible to override the vault transit key used for
encyption/decryption in service.yaml. This will allow services to
specify their own key and then customise the policies to determine who
has access to decrypt/encrypt.


solarkennedy:
Where was transit key "hardcoded" to `paasta` before?

mattmb:
In vault_tools which is the Yelpy bit that isn't open source. You can see I'm bumping the version of that so that we have the latest version which doesn't hard code the key.

PR #2094, opened 2019-01-04
Document delay injection in smartstack.yaml

yannlandry:
This adds documentation for the `fixed_delay` field in `smartstack.yaml`.

Please see the (internal, see branch name) related ticket for more information.

`make docs` passes with this as result:
screen shot 2019-01-04 at 11 02 12 am

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -766,6 +766,36 @@ Routing and Reliability
  * ``timeout_client_ms``: HAProxy `client inactivity timeout <http://cbonte.github.io/haproxy-dconv/configuration-1.5.html#4.2-timeout%20client>`_
    in milliseconds, defaults to 1000.
 
+Fault Injection
+```````````````
+
+These keys are meant to control the built-in fault and delay injection features
+of Envoy, the new proxy we are using to replace HAProxy, and thus won't work
+with the latter.
+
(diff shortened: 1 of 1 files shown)


docs/source/yelpsoa_configs.rst
```diff
@@ -766,6 +766,36 @@ Routing and Reliability
...
+          ecosystem:stagec:
+            duration_ms: 3000
+            percent: 20
+          superregion:uswest1-prod:
+            duration_ms: 1000
+            percent: 40
+
+   - ``duration_ms``: The duration of the delay in milliseconds.
+   - ``percent``: Percentage of requests (0-100) to be randomly affected by this
+     delay. Note that due to Envoy's current definition of _percentage_ as an
```
bplotnick: I wonder why this isn't italicized?

bplotnick:
lg2m! I filed #2096 since I think it should be more obvious in the documentation that a feature like this only works for Envoy (and is a noop elsewhere)

docs/source/yelpsoa_configs.rst
yannlandry: Other italicized text uses `*text*` and no underscores. Maybe it's not supported. I fixed it now.

solarkennedy:
In the long run, I would be OK if there was another place where these docs lived, and adjusted the paasta docs to point to them.

PR #2098, opened 2019-01-08
Add a warning if the secret is empty

Qmando:
Sometimes you could accidentally encrypt an empty string and since that generates a non-empty ciphertext, you might not notice.


paasta_tools/cli/cmds/secret.py
```diff
@@ -175,10 +175,13 @@ def paasta_secret(args):
         service = args.service
     secret_provider = _get_secret_provider_for_service(service, cluster_names=args.clusters)
     if args.action in ["add", "update"]:
+        plaintext = get_plaintext_input(args)
+        if not plaintext:
+            print("Warning: Given plaintext is an empty string.")
```
mattmb: I'd be okay if you want to `sys.exit` here. Can't think why anyone would want to continue with an empty string

PR #2104, opened 2019-01-17
Bump versions of pip, tox and virtualenv in Makefile, make tests pass on macos

keymone:
Any particular reason we need that pin?

The reason i made this PR is `make test` was failing worse than it usually does on macos.


solarkennedy:
Pinning protects us against unexpected breakage. Can you unpin, make it work, pip freeze, and then pin again?

PR #2106, opened 2019-01-18
Allow templates prefixed with underscores in tronfigs

kawaiwanyelp:
### Description
Roses are red,
Violet are blue,
If marathon and chronos configs allow underscore-prefixed templates,
then tronfigs can too!

### Testing
make test
manual testing on soa-configs files

### Notes to reviewers
- I ran into an edge case where if we put templates into a MASTER file, `tronfig` (as part of validation) will fail on any file that depends on it. But since we are moving tronfigs to service directories and jobs will no longer live in MASTER files, this will not be a factor much longer so I am removing it.

### DIFF
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -157,5 +157,11 @@
         ".+": {"$ref": "#definitions/job"}
       }
     }
+  },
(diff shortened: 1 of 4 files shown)


solarkennedy:
Looks easy! Yea, no templates in MASTER.

PR #2110, opened 2019-01-19
Remove some dead code

asottile:
dead code found using https://github.com/asottile/dead

### DIFF
--- a/paasta_tools/monitoring_tools.py
+++ b/paasta_tools/monitoring_tools.py
@@ -70,10 +70,6 @@ def get_page(overrides, service, soa_dir=DEFAULT_SOA_DIR):
     return __get_monitoring_config_value('page', overrides, service, soa_dir)
 
 
-def get_alert_after(overrides, service, soa_dir=DEFAULT_SOA_DIR):
-    return __get_monitoring_config_value('alert_after', overrides, service, soa_dir)
-
-
 def get_realert_every(
     overrides, service, soa_dir=DEFAULT_SOA_DIR,
     monitoring_defaults=monitoring_defaults,
@@ -85,10 +81,6 @@ def get_realert_every(
         soa_dir=soa_dir,
         monitoring_defaults=monitoring_defaults,
     )
-
-
-def get_check_every(overrides, service, soa_dir=DEFAULT_SOA_DIR):
-    return __get_monitoring_config_value('check_every', overrides, service, soa_dir)
 
 
 def get_irc_channels(overrides, service, soa_dir=DEFAULT_SOA_DIR):
--- a/paasta_tools/tron/tron_command_context.py
+++ b/paasta_tools/tron/tron_command_context.py
@@ -2,33 +2,9 @@
 have variables that need to be rendered.
 This is a COPY of https://github.com/Yelp/Tron/blob/master/tron/command_context.py.
 """
-import functools
 import operator
 
 from paasta_tools.tron import tron_timeutils
-
-
-def build_context(object, parent):
-    """Construct a CommandContext for object. object must have a property
-    'context_class'.
-    """
-    return CommandContext(object.context_class(object), parent)
-
-
-def build_filled_context(*context_objects):
-    """Create a CommandContext chain from context_objects, using a Filler
-    object to pass to each CommandContext. Can be used to validate a format
-    string.
-    """
-    if not context_objects:
-        return CommandContext()
-
-    filler = Filler()
-
-    def build(current, next):
-        return CommandContext(next(filler), current)
-
-    return functools.reduce(build, context_objects, None)
 
 
 class CommandContext:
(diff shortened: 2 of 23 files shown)


asottile:
(as I don't have access to the closed source bits you'll want to cross reference the deleted functions with library usage)

paasta_tools/monitoring_tools.py
```diff
@@ -70,10 +70,6 @@ def get_page(overrides, service, soa_dir=DEFAULT_SOA_DIR):
     return __get_monitoring_config_value('page', overrides, service, soa_dir)
 
 
-def get_alert_after(overrides, service, soa_dir=DEFAULT_SOA_DIR):
```
solarkennedy: Gosh, this is kinda embarrassing!

I guess it is better to just be "honest" about what we support, and apparently don't support alert_after or realert_every somehow?

Can you remove these from the docs, or I can in a second pass

asottile: heh then I have to wait for CI, I can do that though :)

paasta_tools/tron/tron_command_context.py
```diff
@@ -75,58 +51,11 @@ def __ne__(self, other):
         return not self == other
 
 
-class JobContext:
```
solarkennedy: I was surprised we had this in here, but I can't figure out how we were/are using this. @chlgit or @qui ?

solarkennedy: I don't see it in any private validation hooks

asottile: It was added in this commit but never referenced: 5231c6e3c420265b2027c715de1aa22691e5a786

paasta_tools/utils.py
```diff
@@ -2473,11 +2463,6 @@ class DeploymentsJsonV2Dict(TypedDict):
     controls: Dict[BranchName, _DeploymentsJsonV2ControlsDict]
 
 
-class DeploymentsJsonDict(TypedDict):
```
solarkennedy: @EvanKrall this was probably created with the intention of using mypy with it somehow? But It doesn't seem to be actually used anywhere. OK to delete?

paasta_tools/monitoring_tools.py
asottile: actually, I'd be more comfortable letting a second pass pick this up -- as I don't know what internal docs reference this

paasta_tools/utils.py
asottile: for context, it was added in this commit but never referenced: 3b7b0705e7a95e98f759b6b4be0ca9cf45278732

man I should really make a tool which spits out these "add in / last referenced in" blurbs :laughing:

paasta_tools/monitoring_tools.py
jvperrin: I think the docs are correct on the support actually, since `alert_after` does appear to be configurable, but just doesn't use this function to do overrides: https://github.com/Yelp/paasta/blob/4fba3ff97da4163db622d4f869a52fb1f48f3ca8/paasta_tools/monitoring_tools.py#L250-L251

Similarly, `realert_every` is set to a default but appears configurable? https://github.com/Yelp/paasta/blob/4fba3ff97da4163db622d4f869a52fb1f48f3ca8/paasta_tools/monitoring_tools.py#L206

`check_every` is not though, and makes it clear in the docs: https://github.com/Yelp/paasta/blob/4fba3ff97da4163db622d4f869a52fb1f48f3ca8/docs/source/yelpsoa_configs.rst#L907-L909

It's also unconditionally overridden at https://github.com/Yelp/paasta/blob/4fba3ff97da4163db622d4f869a52fb1f48f3ca8/paasta_tools/monitoring_tools.py#L252

Does this look right, or are these not actually available as options to services? As far as I can tell it is pulling in the monitoring options from services correctly since `monitoring_overrides` ultimately comes from https://github.com/Yelp/paasta/blob/4fba3ff97da4163db622d4f869a52fb1f48f3ca8/paasta_tools/utils.py#L533-L535

paasta_tools/tron/tron_command_context.py
qui: I think we copied all of https://github.com/Yelp/Tron/blob/master/tron/command_context.py. This was to support Tron-like command context variables in Chronos, before Tron on PaaSTA was a thing.

solarkennedy:
Well uh. I guess we should have faith in the tests and the automation :)

asottile:
@solarkennedy anything left for me to do here?

solarkennedy:
Nope. Just waiting for us all to be brave enough over here to click the button.

If there were closed source things or other things that used `paasta` as a library, how would we "annotate" them to be immune to the deadcode culling?

vkhromov:
> > If there were closed source things or other things that used paasta as a library, how would we "annotate" them to be immune to the deadcode culling?
> 
> 🤔 good question. probably code comments? Or explicitly have a "public" api (or forbid using `paasta` as a library and require consuming tools to use the cli?)

Can we mark to-be-removed objects as [`@deprecated`](https://github.com/tantale/deprecated) and actually remove them after a cool down timeout (e.g. a couple of months)?

asottile:
probably overkill since you own both sides of the integration and can release at will :man_shrugging: -- also `DeprecationWarning`s are hidden by default in python so it probably won't even help (you also don't need to use `deprecated` -- there's `warnings.warn` which is standardlib)

solarkennedy:
Sorry, we just don't have the time to go through this and make sure it won't break things.

solarkennedy:
@asottile when you originally did this, did you actually remove the dead code by hand after `dead` reported it? or was there sed magic?

asottile:
I did hand removals and repeated invocations because sometimes killing a dead function would make more dead

PR #2111, opened 2019-01-23
Bump pysensu-yelp, add more monitoring options

jvperrin:
This is mostly for the `priority` option, but having `tags`, `component`, and `description` can be useful too.

I'm also planning on refactoring this soon to just pass through any supported options to `pysensu-yelp` (except a few that can't be configured, like `check_every` for example). I'm hoping this should make the monitoring overrides a bit simpler anyway since a new function and dictionary key doesn't have to be added for every new option in `pysensu-yelp`. Thoughts?

### DIFF
--- a/paasta_tools/monitoring_tools.py
+++ b/paasta_tools/monitoring_tools.py
@@ -46,6 +46,7 @@ def monitoring_defaults(key):
         'ticket': False,
         'project': None,
         'realert_every': -1,
+        'tags': [],
     }
     return defaults.get(key, None)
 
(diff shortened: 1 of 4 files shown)


paasta_tools/monitoring_tools.py
```diff
@@ -186,36 +203,41 @@ def send_event(service, check_name, overrides, status, output, soa_dir, ttl=None
...
         'source': 'paasta-%s' % cluster,
+        'tags': get_tags(overrides, service, soa_dir),
         'ttl': ttl,
+        'sensu_host': system_paasta_config.get_sensu_host(),
+        'sensu_port': system_paasta_config.get_sensu_port(),
+        'component': get_component(overrides, service, soa_dir),
+        'description': get_description(overrides, service, soa_dir),
     }
 
-    sensu_host = system_paasta_config.get_sensu_host()
```
solarkennedy: This is a change in behavior, because we'll could be sending "null" into pysensu-yelp, which means in turn it will try to 'connect' to it, right?

jvperrin: I do still check below with `if result_dict.get('sensu_host'):` if this value is set to `None` (or a non-empty value anyway), do you think there should be some more controls or protections on it?

PR #2112, opened 2019-01-23
Setup/cleanup kubernetes CRDs for cluster/service

keymone:


paasta_tools/setup_kubernetes_crd.py
```diff
@@ -0,0 +1,140 @@
...
+
+        existing_crd = None
+        for crd in existing_crds.items:
+            if crd.metadata.name == desired_crd.metadata['name']:
+                existing_crd = crd
+                break
+
+        try:
+            if existing_crd:
+                desired_crd.metadata['resourceVersion'] = existing_crd.metadata.resource_version
```
Rob-Johnson: though this might be the right field for us to use, I'm not sure I necessarily want users to have to specify the resource version for the CRD. Can we instead take a hash of the CRD to determine the resource version, and use that?

paasta_tools/kubernetes_tools.py
```diff
@@ -828,14 +829,20 @@ def get_kubernetes_services_running_here_for_nerve(
 
 class KubeClient:
     def __init__(self) -> None:
-        kube_config.load_kube_config(config_file=KUBE_CONFIG_PATH)
+        if 'KUBECONFIG' in os.environ:
```
vkhromov:
```
kube_config_path = os.environ.get('KUBECONFIG', KUBE_CONFIG_PATH)
```
?

paasta_tools/kubernetes_tools.py
```diff
@@ -828,14 +829,20 @@ def get_kubernetes_services_running_here_for_nerve(
...
+            kube_config_path = os.environ.get('KUBECONFIG')
+        else:
+            kube_config_path = KUBE_CONFIG_PATH
+
+        kube_config.load_kube_config(config_file=kube_config_path)
```
vkhromov:
```
kube_config.load_kube_config(config_file=os.environ.get('KUBECONFIG', KUBE_CONFIG_PATH))
```
?

paasta_tools/setup_kubernetes_crd.py
keymone: as i understand resourceVersion is opaque value maintained by the server. it's used for concurrency control and clients aren't supposed to modify it. https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#concurrency-control-and-consistency

PR #2114, opened 2019-01-30
Expose tron monitoring info to paasta

kawaiwanyelp:
### Description
- As in title
- For now, we do not want to turn on tron monitoring in paasta if a job does not have monitoring set. However, we do want to expose the job's team no matter what.

### Testing
- manual testing

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -38,6 +38,7 @@ from paasta_tools.utils import NoConfigurationForServiceError
 from paasta_tools.utils import NoDeploymentsAvailable
 from paasta_tools.utils import paasta_print
 
+from paasta_tools import monitoring_tools
 from paasta_tools.monitoring_tools import list_teams
 from typing import Optional
 from typing import Dict
@@ -256,7 +257,13 @@ class TronJobConfig:
         return self.config_dict.get('schedule')
 
     def get_monitoring(self):
-        return self.config_dict.get('monitoring')
+        srv_monitoring = dict(monitoring_tools.read_monitoring_config(
+            self.service,
+            soa_dir=self.soa_dir,
+        ))
+        tron_monitoring = self.config_dict.get('monitoring', {})
+        srv_monitoring.update(tron_monitoring)
+        return srv_monitoring
 
     def get_queueing(self):
         return self.config_dict.get('queueing')
@@ -316,6 +323,7 @@ class TronJobConfig:
                 )
         else:
             branch_dict = None
+        action_dict['monitoring'] = self.get_monitoring()
 
         return TronActionConfig(
             service=action_service,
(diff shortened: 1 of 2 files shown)


paasta_tools/tron_tools.py
```diff
@@ -256,7 +257,16 @@ def get_schedule(self):
...
 
     def get_monitoring(self):
-        return self.config_dict.get('monitoring')
+        monitoring_conf = self.config_dict.get('monitoring', {})
+        if not monitoring_conf:
+            monitoring_conf = monitoring_tools.read_monitoring_config(
+                self.service,
+                soa_dir=self.soa_dir,
+            )
+            # PAASTA-15158: For now, we do not expose more than the 'team' from
```
solarkennedy: I don't think this is worth doing and I think it will cause more confusion than you think.

This means that we'll get a team, and start doing stuff, but we *won't* be using any of the other settings in monitoring.yaml, so emails will go to a different place than they have configured, etc.

kawaiwanyelp: Alright, I'll switch monitoring on for all tron jobs.

paasta_tools/tron_tools.py
```diff
@@ -256,7 +257,15 @@ def get_schedule(self):
         return self.config_dict.get('schedule')
 
     def get_monitoring(self):
-        return self.config_dict.get('monitoring')
+        monitoring_conf = self.config_dict.get('monitoring', {})
```
solarkennedy: This isn't quite right.

I think this should use the existing `get_monitoring` function we have for instance_configs where we need it.
It isn't like we should use one or the other, it should "merge" the results of the two all the time. Setting page: true for job should be all you need (not have to also specify the team, which is already in monitoring.yaml)

I don't think this function of this object "should" read yet another yaml file though.

kawaiwanyelp: I've modified `get_monitoring` to merge the tron monitoring config dict into the dict we get from `monitoring.yaml`.

qui:
lg2m, pending review from @solarkennedy, since this still reads the yaml file.

Also, once this is released, you should inform developers that this turns monitoring on for all Tron jobs.

paasta_tools/tron_tools.py
```diff
@@ -316,6 +323,7 @@ def _get_action_config(self, action_name, action_dict):
                 )
         else:
             branch_dict = None
+        action_dict['monitoring'] = self.get_monitoring()
```
solarkennedy: This line feels out of place. I understand why we "need" it, but I feel like this should actually override the `get_monitoring` method in `TronActionConfig`.

Looking at the code though, it looks like this "is" how we are injecting monitoring config in other InstanceConfigs. idk, it looks like how we handle monitoring data in our instance configs isn't super tight. They may break in the future, but for now I say ship

PR #2116, opened 2019-01-31
K8S nerve updates

mattmb:
Probably easiest to review commits separately:
* Prefix paasta k8s labels with paasta_ 
* Enable registration of non-paasta k8s services

I haven't been careful to make this backwards compatible. Right now it takes destroying the exisiting resources and causes de-registration of running things in nerve. I think I'm okay with that at the moment because we don't have too much running. I have tested this on kubestage and it will require a bump of nerve-tools too.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -630,9 +630,9 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
                 instance=self.get_sanitised_instance_name(),
             ),
             labels={
-                "service": self.get_service(),
-                "instance": self.get_instance(),
-                "git_sha": code_sha,
+                "yelp.com/paasta_service": self.get_service(),
+                "yelp.com/paasta_instance": self.get_instance(),
+                "yelp.com/paasta_git_sha": code_sha,
             },
         )
 
@@ -664,8 +664,8 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
                         replicas=self.get_desired_instances(),
                         selector=V1LabelSelector(
                             match_labels={
-                                "service": self.get_service(),
-                                "instance": self.get_instance(),
+                                "yelp.com/paasta_service": self.get_service(),
+                                "yelp.com/paasta_instance": self.get_instance(),
                             },
                         ),
                         template=self.get_pod_template_spec(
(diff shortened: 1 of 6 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -775,7 +780,7 @@ def get_kubernetes_services_running_here() -> Sequence[KubeService]:
     services = []
     pods = requests.get('http://127.0.0.1:10255/pods').json()
     for pod in pods['items']:
-        if pod['status']['phase'] != 'Running' or pod['metadata']['namespace'] != 'paasta':
+        if pod['status']['phase'] != 'Running' or 'smartstack_registrations' not in pod['metadata']['annotations']:
```
vkhromov: 'annotations' is optional in `pod['metadata']`, the code probably should be more defensive here trying not to trigger `KeyError`.

vkhromov: `annotations` is optional in `pod['metadata']`, the code probably should be more defensive here trying not to trigger `KeyError`.

paasta_tools/kubernetes_tools.py
```diff
@@ -815,14 +821,23 @@ def get_kubernetes_services_running_here_for_nerve(
...
-                load_deployments=False,
-                soa_dir=soa_dir,
-            )
-            for registration in kubernetes_service_config.get_registrations():
+            # first try getting the list of registrations from
+            # the pods annotations. this is helpful because it doesn't
+            # require a paasta config like kuberentes-<cluster>.yaml
+            # if that fails we fall back to looking up registrations
+            # in the config file
+            if kubernetes_service.registrations:
```
vkhromov: Wondering, maybe we should use the merge of them...

mattmb: Hmm, maybe I should go further and just ditch the old method...

paasta_tools/kubernetes_tools.py
```diff
@@ -815,25 +822,23 @@ def get_kubernetes_services_running_here_for_nerve(
...
                 reg_service, reg_namespace, _, __ = decompose_job_id(registration)
-                nerve_dict = load_service_namespace_config(
-                    service=reg_service, namespace=reg_namespace, soa_dir=soa_dir,
-                )
+                try:
+                    nerve_dict = load_service_namespace_config(
+                        service=reg_service, namespace=reg_namespace, soa_dir=soa_dir,
+                    )
+                except Exception:
+                    log.warning(f"Could not get smartstack config for {reg_service}.{reg_namespace}, skipping")
```
vkhromov: Please print the exception info in the warning, so we won't wonder what exactly happened here.

PR #2121, opened 2019-02-04
Add config for kubernetes custom resources

mattmb:
This makes it possible to set the CustomResources that we will sync in
kubernetes as a paasta config dictionary. Means we can add more custom
resources without releasing paasta_tools.

### DIFF
--- a/tests/test_kubernetes_tools.py
+++ b/tests/test_kubernetes_tools.py
@@ -1193,7 +1193,13 @@ def test_update_deployment():
 def test_create_custom_resource():
     mock_client = mock.Mock()
     formatted_resource = mock.Mock()
-    create_custom_resource(mock_client, formatted_resource, version='v1', kind=mock.Mock(plural='someclusters'))
+    create_custom_resource(
+        kube_client=mock_client,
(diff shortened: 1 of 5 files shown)


tests/test_kubernetes_tools.py
```diff
@@ -1193,7 +1193,13 @@ def test_update_deployment():
 def test_create_custom_resource():
     mock_client = mock.Mock()
     formatted_resource = mock.Mock()
-    create_custom_resource(mock_client, formatted_resource, version='v1', kind=mock.Mock(plural='someclusters'))
+    create_custom_resource(
+        kube_client=mock_client,
+        formatted_resource=formatted_resource,
+        version='v1',
+        kind=mock.Mock(plural='someclusters'),
+        group='yelp.com',
```
vkhromov: just grumbling, but since you've changed group name from the hardcoded "yelp.com" to non-hardcoded values, it probably better to use some nonsense instead of previous hardcoded value for the group name in tests, to test that the hardcoded name haven't leaked out from somewhere and the tests isn't catching it as the result.

PR #2123, opened 2019-02-06
Add disk to tron validation

kawaiwanyelp:
### Description
- In order to complete adding disk support to Tron, we need to make sure tronfigs can be validated by paasta, which includes the new disk field

### Testing
make test

### DIFF
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -32,12 +32,17 @@
         "executor": {"enum": ["ssh", "paasta"]},
         "cpus": {
           "type": "number",
-          "minumum": 0,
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/schemas/tron_schema.json
```diff
@@ -32,12 +32,17 @@
         "executor": {"enum": ["ssh", "paasta"]},
         "cpus": {
           "type": "number",
-          "minumum": 0,
+          "minimum": 0,
```
EvanKrall: maybe we should get jsonschema validation for our json schemas 😆

kawaiwanyelp: as well as our json schema validation for that validation

PR #2133, opened 2019-02-14
Pass X-Nerve-Check-IP to configure_nerve

mattmb:
This is so that we send the IP of Kubernetes Pods to Nerve in this
header. That way a system level hacheck knows what IP to check against.
Note that I am also making us send the header in the case of marathon
tasks. This is so we can use the same hacheck for both marathon and
kubernetes tasks.


mattmb:
Probably see this first: https://github.com/Yelp/hacheck/pull/24

solarkennedy:
@mattmb  do you still need this?

mattmb:
Oh man, I messed up that conflict resolution! Let's have another go...

PR #2135, opened 2019-02-19
Added a script to kill bad containers that share ips or ports

solarkennedy:
I think we are ready to auto-remediate this situation instead of just alerting on it.

cc @qui @bplotnick

### DIFF
--- a/paasta_tools/contrib/kill_bad_containers.py
+++ b/paasta_tools/contrib/kill_bad_containers.py
@@ -1,0 +1,73 @@
+#!/opt/venvs/paasta-tools/bin/python3
+import sys
+
+import iptc
+
+from paasta_tools import iptables
+from paasta_tools.utils import get_docker_client
+
+
+def list_docker_nat_rules():
+    chain_name = 'DOCKER'
+    table = iptc.Table(iptc.Table.NAT)
+    chain = iptc.Chain(table, chain_name)
+    for rule in chain.rules:
+        yield iptables.Rule.from_iptc(rule)
+
+
+def get_container_from_dport(dport, docker_client):
+    for container in docker_client.containers():
+        try:
+            ports = container['Ports']
+            for port in ports:
+                if "PublicPort" in port:
+                    if port["PublicPort"] == int(dport):
+                        return container
+        except KeyError:
+            print(ports)
+            pass
+
+
+def target_rule_to_dport(rule):
+    try:
+        # (('tcp', (('dport', ('31493',)),)),)
+        return rule.matches[0][1][0][1][0]
+    except IndexError:
+        return None
+
+
+def kill_containers_with_duplicate_iptables_rules(docker_client):
+    targets_seen = {}
+    for rule in list_docker_nat_rules():
+        target = rule.target_parameters
+        if target not in targets_seen:
+            targets_seen[target] = rule
+        else:
+            print("This is the second time we've seen a rule with the same target_parameters!")
+            print(rule)
+            dport = target_rule_to_dport(rule)
+            container1 = get_container_from_dport(dport, docker_client)
+            print("The other rule with that target is:")
+            print(targets_seen[target])
+            dport = target_rule_to_dport(rule)
+            container2 = get_container_from_dport(dport, docker_client)
+            if container1["Id"] == container2["Id"]:
+                print("The same container is getting traffic for both ports!")
+                print(container1)
+                print("Killing the container")
+                docker_client.kill(container1["Id"])
+            else:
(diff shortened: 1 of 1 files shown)


bplotnick:
Do we have some way to track numbers of containers killed by this? something something statsd?

I'm concerned that this is going to mask the problem and potentially cause us to miss something big.

solarkennedy:
You are right that the purpose of this PR is to mask the problem.
This will log and show up in splunk, is that enough?

bplotnick:
That's fine. Let's just make sure we're watching it!

qui:
Do you need to do anything else to remove the IP tables rules?

solarkennedy:
Gosh, I sure wish I didn't! 
But you have a good point, this will just keep happening if port (ip?) are used again.
I'll add a rule deletion line.

solarkennedy:
Ok, I've added the rule deletion.
This feels pretty iffy, so I'll definitely turn this on slowly and watch its behavior before turning it on in prod.

paasta_tools/contrib/kill_bad_containers.py
```diff
@@ -62,6 +62,9 @@ def kill_containers_with_duplicate_iptables_rules(docker_client):
                 print(container2)
                 print("Not sure which to kill, picking the first one")
                 docker_client.kill(container1["Id"])
+            print("Deleting the iptables rules")
+            chain.delete_rule(iptables_rule)
+            chain.delete_rule(raw_rules[target])
```
qui: If it's two different containers, and we only kill one of them, don't we need to keep one of the rules?

solarkennedy: Hmm right. Crap.

paasta_tools/contrib/kill_bad_containers.py
```diff
@@ -0,0 +1,78 @@
...
+            targets_seen[target] = rule
+            raw_rules_seen[target] = iptables_rule
+        else:
+            print("This is the second time we've seen a rule with the same target_parameters!")
+            print(rule)
+            dport = target_rule_to_dport(rule)
+            container1 = get_container_from_dport(dport, docker_client)
+            print("The other rule with that target is:")
+            print(targets_seen[target])
+            dport = target_rule_to_dport(rule)
```
qui: Should this be targets_seen[rule] instead of rule? otherwise it's the same as L48

solarkennedy: ! You are totally right

paasta_tools/contrib/kill_bad_containers.py
```diff
@@ -0,0 +1,78 @@
...
+def get_container_from_dport(dport, docker_client):
+    for container in docker_client.containers():
+        try:
+            ports = container['Ports']
+            for port in ports:
+                if "PublicPort" in port:
+                    if port["PublicPort"] == int(dport):
+                        return container
+        except KeyError:
+            print(ports)
```
keymone: this feels too defensive or maybe i misunderstand something, if KeyError is thrown, it means `Ports` is not found in `container` and `ports` variable will be `None`, what information do you  expect to get out of that `print` call?

how about:

    for container in docker_client.containers():
      for ports in container.get('Ports', []):
        for port in ports:
          if port.get('PublicPort') == int(dport):
            return container

paasta_tools/contrib/kill_bad_containers.py
```diff
@@ -0,0 +1,78 @@
...
+            pass
+
+
+def target_rule_to_dport(rule):
+    try:
+        # (('tcp', (('dport', ('31493',)),)),)
```
keymone: fun fact: this comment is actually almost valid python destructuring code:

    (_tcp,((_dport,(value,),),),), = rule.matches
    return int(value)

not that i suggest to use it

paasta_tools/contrib/kill_bad_containers.py
solarkennedy: Hah. It is not too defensive. The world is a zoo and I found all sorts of weird things.

paasta_tools/contrib/kill_bad_containers.py
solarkennedy: Interesting, I hadn't thought about going in that direction (Because I copy/pasted it out of a repl).

I mean, why *not* do it like that? (still with a try/except though)

keymone: i think it doesn't really make the structure of right-hand side obvious, which is what i would want from pattern-matching/destructuring assignments

PR #2136, opened 2019-02-21
Filter non-tron monitoring keys

kawaiwanyelp:
### Description
- Filter non-tron monitoring keys, since if we pass them through, tron will freak out

### Testing
make test

### Notes
- I wanted to use `paasta_tools.cli.cmds.validate.get_schema`, but that would be a circular dependency because `validate` depends on `tron_tools`. I still wanted to use the tron schema as the source of truth, so I decided to "copy" the `get_schema` functionality over to `tron_tools`. It is a little ugly :(

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -13,7 +13,9 @@
 import datetime
 import difflib
 import glob
+import json
 import os
+import pkgutil
 import re
 import subprocess
 from string import Formatter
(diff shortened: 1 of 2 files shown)


paasta_tools/tron_tools.py
```diff
@@ -46,6 +48,11 @@
 
 MASTER_NAMESPACE = 'MASTER'
 SPACER = '.'
+MONITORING_KEYS = {
+    k for k in json.loads(
```
solarkennedy: Why do you need the k for k in and the dictioniary? Isn't `.keys()` enough?

paasta_tools/tron_tools.py
```diff
@@ -46,6 +48,11 @@
 
 MASTER_NAMESPACE = 'MASTER'
 SPACER = '.'
+MONITORING_KEYS = {
```
solarkennedy: How about `VALID_MONITORING_KEYS` ?

PR #2138, opened 2019-02-22
validate deploy groups for Tron in paasta validate

chlgit:
If the deploy_group doesn't exist in deploy.yaml, the error message looks like:

✗ tron-mesosstage.yaml is invalid:
  deploy_group everything1 is not in deploy.yaml
  More info: http://tron.readthedocs.io/en/latest/jobs.html

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -40,6 +40,8 @@ from paasta_tools.utils import paasta_print
 
 from paasta_tools import monitoring_tools
 from paasta_tools.monitoring_tools import list_teams
+from paasta_tools.utils import get_pipeline_config
+from paasta_tools.utils import is_deploy_step
 from typing import Optional
 from typing import Dict
 from typing import Any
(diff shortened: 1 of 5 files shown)


qui:
Since it's also possible to set a service/deploy group in an action, can you validate that too?

paasta_tools/tron_tools.py
```diff
@@ -387,11 +389,25 @@ def check_actions(self) -> Tuple[bool, List[str]]:
                 msgs.extend(action_msgs)
         return checks_passed, msgs
 
+    def check_deploy_group(self) -> Tuple[bool, str]:
+        deploy_groups = [action.get_deploy_group() for action in self.get_actions()]
+        deploy_groups.append(self.get_deploy_group())
+        check_deploy_groups = [deploy_group for deploy_group in deploy_groups if deploy_group is not None]
+        if len(check_deploy_groups) > 0:
+            pipeline_steps = [step['step'] for step in get_pipeline_config(self.service, self.soa_dir)]
```
qui: the service can also vary by action, so I think you might want to do this verification in the action instead

PR #2144, opened 2019-02-27
Add docs for monitoring defaults

kawaiwanyelp:
### Description
- Updated the docs to reflect the new tron monitoring behavior, where paasta now defaults to `monitoring.yaml`.

### Notes
- Related to: https://github.com/Yelp/Tron/pull/621

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -279,9 +279,7 @@ instance MAY have:
     Parsing the Marathon config file will fail if both args and cmd are
     specified [#note]_.
 
-  * ``monitoring``: A dictionary of values that configure overrides for
-    monitoring parameters that will take precedence over what is in
-    `monitoring.yaml`_. These are things like ``team``, ``page``, etc.
+  * ``monitoring``: See the `monitoring.yaml`_ section for details.
 
   * ``autoscaling``: See the `autoscaling docs <autoscaling.html>`_ for valid options and how they work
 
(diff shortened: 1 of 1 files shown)


docs/source/yelpsoa_configs.rst
```diff
@@ -544,6 +544,29 @@ Example Job
             cpus: .5
             mem: 100
 
+Default Monitoring
+^^^^^^^^^^^^^^^^^^
+
+For any monitoring fields not set in a Tron job's configuration, PaaSTA will
```
solarkennedy: You don't need to duplicate this. 

There is already

```monitoring: A dictionary of values that configure overrides for monitoring parameters that will take precedence over what is in [[monitoring.yaml]]. These are things like team, page, etc.```

Put this into the "common options" section and I think that would be OK

Comment (kawaiwanyelp) [reply docs/source/yelpsoa_configs.rst]: I've moved stuff to the `monitoring.yaml` section.

File: docs/source/yelpsoa_configs.rst [hunk 2]
```diff
@@ -976,6 +976,17 @@ A marathon service that overrides options on different instances (canary)::
...
 
+A tron job that pages for a specific job in a service::
+
+  # monitoring.yaml
+  team: midend
+  page: false
+
+  # tron-prod.yaml
+  jobs:
+  - name: "foo"
```
solarkennedy: Please use the dictionary format instead of the array format

PR #2150, opened 2019-03-04
Upgrade hooks, fix flake8

asottile:


solarkennedy:
Thanks as always @asottile for gardening are codebase.

PR #2154, opened 2019-03-06
set the desired state of the custom resource

Rob-Johnson:
we want the initial state to be running: this'll have the same semantics
as paasta start/stop, where any changes to the desired state are
forgotten with each deployment


paasta_tools/setup_kubernetes_cr.py
```diff
@@ -203,6 +203,10 @@ def format_custom_resource(
                 'yelp.com/paasta_service': service,
                 'yelp.com/paasta_instance': instance,
                 'yelp.com/paasta_cluster': cluster,
+                'yelp.com/desired_state': 'running',
```
keymone: drop?

PR #2156, opened 2019-03-06
Added a PAASTA_LAUNCHED_BY variable

solarkennedy:


paasta_tools/cli/cmds/local_run.py
```diff
@@ -563,6 +564,7 @@ def get_local_run_environment_vars(instance_config, port0, framework):
         'MESOS_CONTAINER_NAME': 'localrun-%s' % fake_taskid,
         'MESOS_TASK_ID': str(fake_taskid),
         'PAASTA_DOCKER_IMAGE': docker_image,
+        'PAASTA_LAUNCHED_BY': get_possible_launched_by_user_variable_from_env(),
```
thebostik: Looks like the docstring says this may not be present, but from reading this it looks like it would always be present but have the value of None?
I think latter makes sense, some default value so we don't have to be as safe when trying to consume the environment variable as compared to if it's optional.

solarkennedy: No, if it isn't present, then the variable literally will not be there.

I'm not planning on injecting this variable into things like marathon where it doesn't make as much sense.

thebostik: ah got it, I see what you mean by optional in the docs now.

PR #2159, opened 2019-03-08
Fix `paasta status` for Flink tasks without deployments

vkhromov:


paasta_tools/cli/cmds/status.py
```diff
@@ -740,8 +740,13 @@ def paasta_status(
     clusters_services_instances = apply_args_filters(args)
     for cluster, service_instances in clusters_services_instances.items():
         for service, instances in service_instances.items():
-            actual_deployments = get_actual_deployments(service, soa_dir)
-            if actual_deployments:
+            all_flink = all(i == FlinkClusterConfig for i in instances.values())
```
keymone: isinstance?

vkhromov: `isinstance` checks that an object is an instance of a class, here we are checking that `i` is the given class.

PR #2162, opened 2019-03-09
Convert mark-for-deployment to a state machine managed by pytransitions

EvanKrall:


paasta_tools/automatic_rollbacks.py
```diff
@@ -110,3 +118,81 @@ def watch_for_slack_webhooks():
...
+            model=self,
+            states=list(self.states()),
+            transitions=list(self.valid_transitions()),
+            initial=self.start_state(),
+            after_state_change=self.after_state_change,
+            queued=True,
+        )
+
+        for state in list(self.states()):
+            # Call e.g. on_enter_start(self.enter_start)
```
solarkennedy: Why do we need this? The docs say this is automatic?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -423,60 +427,21 @@ def paasta_mark_for_deployment(args):
...
-    slack_notifier = SlackDeployNotifier(
-        deploy_info=deploy_info, service=service,
-        deploy_group=deploy_group, commit=commit, old_commit=old_git_sha, git_url=args.git_url,
-        auto_rollback=args.auto_rollback,
-    )
 
-    ret = mark_for_deployment(
-        git_url=args.git_url,
-        deploy_group=deploy_group,
+    deploy_process = MarkForDeploymentProcess(
```
solarkennedy: Please gate this on `if args.auto_rollback`.

I know it is cleaner, but we have no itests, and this code can stand to wait until it has proven itself at least a little bit before we put it on the critical path of deploying all services at yelp.

paasta_tools/automatic_rollbacks.py
EvanKrall: Oh, I missed the bit where `on_enter_*` on your model gets automatically called; I only saw the part where you can pass a function/method name to `on_enter_*` on your Machine.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -783,7 +999,8 @@ def wait_for_deployment(service, deploy_group, git_sha, soa_dir, timeout):
                  .format(git_sha, deploy_group))
 
     deadline = time.time() + timeout
-    green_light = Event()
+    if green_light is None:
+        green_light = Event()
```
solarkennedy: Is this safe to do even if there are not threads going on?

EvanKrall: Yep, pretty sure.

keymone:
revert #2168 to apply this again

PR #2165, opened 2019-03-11
Added tron support in the paasta api

EmanekaT:
I picked up kwa's work. #2045
But instead of adding it to frontend, I added get requests directly to paasta client. This requires me to add tron api endpoints in /etc/paasta to work. I have updated puppet and will post a review for that later.

Test output (updated):
https://fluffy.yelpcorp.com/i/mXRZNMNg5GjMxKzNt1kFkPQN31F7k1Nd.html

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -17,9 +17,7 @@ import difflib
 import os
 import sys
 from collections import defaultdict
-from datetime import datetime
 from distutils.util import strtobool
-from itertools import groupby
 from typing import Callable
 from typing import DefaultDict
 from typing import Dict
@@ -31,22 +29,17 @@ from typing import Set
 from typing import Tuple
 from typing import Type
 
-import humanize
 from bravado.exception import HTTPError
 from service_configuration_lib import read_deploy
 
 from paasta_tools import kubernetes_tools
-from paasta_tools.adhoc_tools import AdhocJobConfig
+from paasta_tools import tron_tools
 from paasta_tools.api.client import get_paasta_api_client
-from paasta_tools.chronos_tools import ChronosJobConfig
 from paasta_tools.cli.utils import execute_paasta_serviceinit_on_remote_master
 from paasta_tools.cli.utils import figure_out_service_name
 from paasta_tools.cli.utils import get_instance_configs_for_service
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_deploy_groups
-from paasta_tools.flinkcluster_tools import FLINK_INGRESS_PORT
-from paasta_tools.flinkcluster_tools import FlinkClusterConfig
-from paasta_tools.flinkcluster_tools import sanitised_name
 from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
 from paasta_tools.kubernetes_tools import KubernetesDeployStatus
 from paasta_tools.marathon_serviceinit import bouncing_status_human
@@ -57,7 +50,6 @@ from paasta_tools.marathon_tools import MarathonDeployStatus
 from paasta_tools.monitoring_tools import get_team
 from paasta_tools.monitoring_tools import list_teams
 from paasta_tools.utils import compose_job_id
-from paasta_tools.utils import datetime_from_utc_to_local
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import get_soa_cluster_deploy_files
 from paasta_tools.utils import InstanceConfig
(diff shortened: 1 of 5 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -448,6 +449,44 @@ def print_kubernetes_status(
...
+
+    job_content = client.get_job_content(job=job)
+    latest_run_id = client.get_latest_job_run_id(job_content=job_content)
+    action_run = client.get_action_run(job=job, action=action, run_id=latest_run_id)
+
+    job_content['url'] = client.master_url + f'/api/jobs/{job}/{latest_run_id}'
+    return job_content, action_run
+
+
+def parse_job(job_content: dict, action_run: dict, output: Sequence[str], verbose: bool) -> None:
```
keymone: rename to `write_job_status`

paasta_tools/cli/cmds/status.py
```diff
@@ -534,6 +573,13 @@ def report_status_for_cluster(
                 for line in status.rstrip().split('\n'):
                     output.append('    %s' % line)
 
+    for name, instance in instance_whitelist.items():
+        if str(instance) == "<class 'paasta_tools.tron_tools.TronActionConfig'>":
```
keymone: `isinstance(instance, tron_tools.TronActionConfig)`

paasta_tools/cli/cmds/status.py
```diff
@@ -534,6 +573,13 @@ def report_status_for_cluster(
                 for line in status.rstrip().split('\n'):
                     output.append('    %s' % line)
 
+    for name, instance in instance_whitelist.items():
```
keymone: this will display status for every action in every job for a given service. we should instead have single status item per job consisting of some info about every action of that job preferably in topological order - tron api provides all the data necessary for that.

paasta_tools/tron_tools.py
```diff
@@ -125,6 +125,17 @@ def get_value(self, key, args, kwds):
                 return Formatter.get_value(key, args, kwds)
 
 
+def get_tron_api_endpoints(cluster: str):
+    endpoints = load_system_paasta_config().config_dict.get('tron_api_endpoints', {})
+    if cluster not in endpoints.keys():
```
keymone: nitpick: `item in dict.keys()` is equivalent  to `item in dict`

paasta_tools/tron_tools.py
```diff
@@ -125,6 +125,17 @@ def get_value(self, key, args, kwds):
...
+        raise Exception(f"tron api endpoint is not defined for cluster {cluster}")
+    return endpoints[cluster]
+
+
+def get_tron_client_for_cluster(cluster: str):
+    return TronClient(get_tron_api_endpoints(cluster))
```
keymone: the way this function is used, you're going to create a client `num_clusters * num_services * num_instances` times. since client is only different per-cluster - you're doing the same work a lot. you could do memoization of `get_tron_client_for_cluster` but that's not great for long-lived processes (when configs change some of client objects may become stale). i would suggest to create these client objects much earlier in the call stack and pass as argument to functions that need them, such as `tron_instance_status`.

this will be tricky as the best candidate, `report_status_for_cluster`, is already invoked `num_clusters * num_services` times, but it is a cli function, so you don't have to worry about cache invalidation there - you can make a global dict `TRON_CLIENTS[cluster]` and only create a client object for a given cluster if it's not yet in the dict.

paasta_tools/utils.py
```diff
@@ -2097,7 +2097,7 @@ def get_slack_token(self) -> str:
         return self.config_dict.get('slack', {}).get('token', None)
 
     def get_tron_config(self) -> dict:
-        return self.config_dict.get('tron', {})
+        return self.config_dict.get('tron_api_endpoints', {})
```
keymone: why do we need to change this?

paasta_tools/cli/utils.py
```diff
@@ -771,9 +771,9 @@ def run_on_master(
         # signals don't travel over ssh, kill process when anything lands on stdin instead
         cmd_parts.append(
             # send process to background and capture it's pid
-            '& p=$!; ' +
+            '& p=$!; '
             # wait for stdin with timeout in a loop, exit when original process finished
-            'while ! read -t1; do ! kill -0 $p 2>/dev/null && kill $$; done; ' +
+            'while ! read -t1; do ! kill -0 $p 2>/dev/null && kill $$; done; '
             # kill original process if loop finished (something on stdin)
```
EmanekaT: This is modified by pre-commit hook ....

paasta_tools/utils.py
EmanekaT: Since I changed the schema, I just think this makes more sense.

paasta_tools/cli/cmds/status.py
EmanekaT: I tried but this doesn't work unfortunately.

EmanekaT:
I'm closing the ticket as I realized that paasta clients on dev boxes cannot access tron masters on prod, not even with jumpboxes.

PR #2178, opened 2019-03-19
Make paasta status command show tron jobs

EmanekaT:
I added an argument cluster for paasta-api command.
I tested by running paasta-api locally.
https://fluffy.yelpcorp.com/i/3N0wlHV39FQPhL8BG8tStWw7KNV8JF9M.html
There is another ticket for adding tron-api endpoints to /etc/paasta.
And there will be another ticket for paasta client changes.


paasta_tools/api/api.py
```diff
@@ -148,6 +156,9 @@ def main(argv=None):
     if args.soa_dir:
         os.environ["PAASTA_API_SOA_DIR"] = args.soa_dir
 
+    if args.cluster:
```
solarkennedy: Do you need all three? Can you just do the CLI arg instead of having the cli/config/and env ?

paasta_tools/tron/client.py
```diff
@@ -109,3 +109,17 @@ def list_namespaces(self):
         """Gets the namespaces that are currently configured."""
         response = self._get('/api')
         return response.get('namespaces', [])
+
+    def get_job_content(self, job):
+        return self._get(f"/api/jobs/{job}/")
+
+    def get_latest_job_run_id(self, job_content):
```
solarkennedy: This isn't a super efficient way to do this.

The only reason we do this in check_tron_jobs is because we need to collect all the runs for context.

For here, if you only want the most recent run, just do `num_runs=1`

solarkennedy: `?num_runs=1` on the `get_job` function

paasta_tools/utils.py
```diff
@@ -2098,7 +2098,7 @@ def get_slack_token(self) -> str:
         return self.config_dict.get('slack', {}).get('token', None)
 
     def get_tron_config(self) -> dict:
-        return self.config_dict.get('tron', {})
+        return self.config_dict.get('tron_api_endpoints', {})
```
solarkennedy: Are you going to make a review in puppet to deploy this new config file?

You are going to have to provide both old and new

paasta_tools/tron_tools.py
```diff
@@ -723,3 +723,14 @@ def list_tron_clusters(service: str, soa_dir: str = DEFAULT_SOA_DIR) -> List[str
         if cluster_re_match is not None:
             clusters.append(cluster_re_match.group(1))
     return clusters
+
+
+def get_tron_api_endpoints(cluster: str):
```
solarkennedy: I think this should just be `get_tron_api_endpoint` (not plural) if it only returns one

paasta_tools/api/views/instance.py
```diff
@@ -268,7 +302,7 @@ def instance_status(request):
         error_message = traceback.format_exc()
         raise ApiFailure(error_message, 500)
 
-    if instance_type != 'flinkcluster':
+    if instance_type != 'flinkcluster' and instance_type != 'tron':
```
solarkennedy: Why do we need this?

paasta_tools/tron/client.py
solarkennedy: Eh disregard. My approach will fetch "scheduled" jobs, and really we want the most latest "completed" job.

paasta_tools/api/views/instance.py
EmanekaT: version = get_deployment_version(actual_deployments, settings.cluster, instance) returns None.

paasta_tools/utils.py
EmanekaT: Is the old one used anywhere? I didn't find tron.json in /etc/paasta.

paasta_tools/api/api.py
EmanekaT: settings.cluster is the one that needs to be here. I put env here because I don't want to pass "args" to the setup_paasta_api function or make it global. Also I want to be more consistent so it's easier to read.

paasta_tools/tron/client.py
```diff
@@ -109,3 +109,17 @@ def list_namespaces(self):
         """Gets the namespaces that are currently configured."""
         response = self._get('/api')
         return response.get('namespaces', [])
+
+    def get_job_content(self, job):
```
vkhromov: Can you please add type annotations for args and the return value here and in other new/updated functions?

EmanekaT: Sure

paasta_tools/api/api.py
```diff
@@ -53,6 +53,11 @@ def parse_paasta_api_args():
         dest="soa_dir",
         help="define a different soa config directory",
     )
+    parser.add_argument(
```
qui: Why do we need to be able to run a PaaSTA API server for a different cluster?

paasta_tools/api/api_docs/swagger.json
```diff
@@ -812,6 +816,60 @@
         "last_status"
       ]
     },
+    "InstanceStatusTron": {
+      "type": "object",
+      "required": [ "job_name", "job_url"],
+      "properties": {
+        "job_name": {
+          "type": "string",
+          "description": "State of a Tron action or job"
```
qui: ? All of these descriptions need to be updated.

paasta_tools/api/api_docs/swagger.json
```diff
@@ -812,6 +816,60 @@
...
+        },
+        "action_command": {
+          "type": "string",
+          "description": "The raw command for the action"
+        },
+        "action_start_time": {
+          "type": "string",
+          "description": "The raw command for the action"
+        },
+        "stdout": {
```
qui: Why both stdout and action_stdout?

paasta_tools/tron/client.py
```diff
@@ -109,3 +109,17 @@ def list_namespaces(self):
...
         response = self._get('/api')
         return response.get('namespaces', [])
+
+    def get_job_content(self, job: str) -> dict:
+        return self._get(f"/api/jobs/{job}/")
+
+    def get_latest_job_run_id(self, job_content: dict) -> str:
+        job_runs = sorted(
+            job_content.get('runs', []),
+            key=lambda k: (k['end_time'] is None, k['end_time'], k['run_time']),
```
qui: I think it will make more sense to just return the last run by run number that is not scheduled.

paasta_tools/api/views/instance.py
```diff
@@ -115,6 +116,45 @@ def chronos_instance_status(
...
+
+    job_content = client.get_job_content(job=job)
+    latest_run_id = client.get_latest_job_run_id(job_content=job_content)
+    action_run = client.get_action_run(job=job, action=action, run_id=latest_run_id)
+
+    # job info
+    status['job_name'] = short_job
+    status['job_status'] = job_content['status']
+    status['job_schedule'] = '{} {}'.format(job_content['scheduler']['type'], job_content['scheduler']['value'])
+    status['job_url'] = client.master_url + f'/api/jobs/{job}/{latest_run_id}'
```
qui: The master URL won't work for prod. You need the URLs that go through the jumpbox (e.g. tronweb-norcal-prod). I'd suggest you add a separate dashboard link for Tron, with the tronweb addresses.

paasta_tools/api/views/instance.py
```diff
@@ -115,6 +116,45 @@ def chronos_instance_status(
     return cstatus
 
 
+def tron_instance_status(
+    instance_status: Mapping[str, Any],
+    service: str,
+    instance: str,
+    verbose: bool,
```
qui: Are you planning to make it less verbose later?

paasta_tools/api/api.py
EmanekaT: This makes it easier to test. kwa suggested to add this.

paasta_tools/api/views/instance.py
EmanekaT: No.

qui: We should not be showing stdout and stderr if the user does not increase the level of verbosity (with -v). It should be consistent with how paasta status looks with `paasta status -v`, `paasta status -vv`, and `paasta status -vvv` for Marathon and Chronos. 

Sorry if I was unclear. I meant that you can follow up with the changes for increasing levels of verbosity. But for the first pass, `paasta status` with no v's should return less information.

EmanekaT: Ah, I see. But I think this verbose variable is not included in the request from the client side. So I moved verbose to the client side.

paasta_tools/utils.py
```diff
@@ -1649,7 +1649,7 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     register_marathon_services: bool
     register_native_services: bool
     nerve_readiness_check_script: str
-    tron: Dict
+    tron: str
```
qui: why is this a string?

paasta_tools/tron_tools.py
```diff
@@ -723,3 +723,10 @@ def list_tron_clusters(service: str, soa_dir: str = DEFAULT_SOA_DIR) -> List[str
         if cluster_re_match is not None:
             clusters.append(cluster_re_match.group(1))
     return clusters
+
+
+def get_tron_dashboard_for_cluster(cluster: str):
+    dashboards = load_system_paasta_config().config_dict['dashboard_links'][cluster]
```
qui: nit: use get_dashboard_links

paasta_tools/api/views/instance.py
```diff
@@ -115,6 +116,45 @@ def chronos_instance_status(
...
+        status['action_name'] = action
+    if action_run['state']:
+        status['action_state'] = action_run['state']
+    if action_run['start_time']:
+        status['action_start_time'] = action_run['start_time']
+    if action_run['raw_command']:
+        status['action_raw_command'] = action_run['raw_command']
+    if action_run['stdout']:
+        status['action_stdout'] = action_run['stdout']
+    if action_run['stdout']:
```
qui: stderr

tests/api/test_instance.py
```diff
@@ -327,3 +327,41 @@ def test_add_slave_info():
...
+        'command': 'fake_command',
+        'stdout': 'fake_stdout',
+        'stderr': 'fake_stderr',
+    }
+    mock_get_tron_dashboard_for_cluster.return_value = 'http://fake_url/'
+
+    request = testing.DummyRequest()
+    request.swagger_data = {'service': 'fake_service', 'instance': 'fake_job.fake_action'}
+    response = instance.instance_status(request)
+    assert response['tron']['action_command'] == 'fake_command'
```
qui: How about asserting on the entire expected response, instead of only checking a couple values? You've already put a bunch of test values in the client response, might as well check they're all there.

For example, this would have caught the "stderr" typo earlier.

paasta_tools/utils.py
```diff
@@ -2097,7 +2097,7 @@ def get_slack_token(self) -> str:
         none available """
         return self.config_dict.get('slack', {}).get('token', None)
 
-    def get_tron_config(self) -> dict:
+    def get_tron_config(self) -> Union[str, Any]:
```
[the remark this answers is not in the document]
EmanekaT: No. mypy complains if it's a dict

PR #2180, opened 2019-03-20
Added tron jobs to paasta status. paasta client part

EmanekaT:
I tested by changing paasta endpoint to localhost, ran the updated paasta api, and then ran this updated client. 
Test result: the first part is message received from api.
https://fluffy.yelpcorp.com/i/tLMB4NlNxgP880cLbZ0QXqjqb8hSHxqK.html

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -478,7 +478,7 @@ def print_tron_status(
     if verbose:
         output.append(f"      Status: {tron_status.job_status}")
         output.append(f"      Schedule: {tron_status.job_schedule}")
-    output.append("      DashBoard: {}".format(PaastaColors.blue(tron_status.job_url)))
+    output.append("      Dashboard: {}".format(PaastaColors.blue(tron_status.job_url)))
 
     output.append(f"    Action: {tron_status.action_name}")
     output.append(f"      Status: {tron_status.action_state}")
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -55,6 +55,7 @@ from paasta_tools.marathon_serviceinit import status_marathon_job_human
 from paasta_tools.marathon_tools import MarathonDeployStatus
 from paasta_tools.monitoring_tools import get_team
 from paasta_tools.monitoring_tools import list_teams
+from paasta_tools.tron_tools import TronActionConfig
 from paasta_tools.utils import compose_job_id
 from paasta_tools.utils import datetime_from_utc_to_local
 from paasta_tools.utils import DEFAULT_SOA_DIR
@@ -68,6 +69,7 @@ from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import paasta_print
 from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import SystemPaastaConfig
+
 HTTP_ONLY_INSTANCE_CONFIG: Sequence[Type[InstanceConfig]] = [
     FlinkClusterConfig,
     KubernetesDeploymentConfig,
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -464,6 +467,33 @@ def print_kubernetes_status(
...
+    instance: str,
+    output: List[str],
+    tron_status,
+    verbose: int = 0,
+) -> int:
+    output.append(f"    Tron job: {tron_status.job_name}")
+    if verbose:
+        output.append(f"      Status: {tron_status.job_status}")
+        output.append(f"      Schedule: {tron_status.job_schedule}")
+    output.append("      DashBoard: {}".format(PaastaColors.blue(tron_status.job_url)))
```
qui: Dashboard

paasta_tools/cli/cmds/status.py
```diff
@@ -464,6 +467,33 @@ def print_kubernetes_status(
...
+    output.append(f"    Action: {tron_status.action_name}")
+    output.append(f"      Status: {tron_status.action_state}")
+    if verbose:
+        output.append(f"      Start time: {tron_status.action_start_time}")
+    output.append(f"      Command: {tron_status.action_command}")
+    if verbose:
+        output.append(f"      Raw Command: {tron_status.action_raw_command}")
+        output.append(f"      Stdout: {tron_status.action_stdout}")
+        output.append(f"      Stderr: {tron_status.action_stderr}")
+    output.append('\n')
```
qui: I don't think you need this extra line.

paasta_tools/cli/cmds/status.py
```diff
@@ -464,6 +467,33 @@ def print_kubernetes_status(
...
+    output.append("      DashBoard: {}".format(PaastaColors.blue(tron_status.job_url)))
+
+    output.append(f"    Action: {tron_status.action_name}")
+    output.append(f"      Status: {tron_status.action_state}")
+    if verbose:
+        output.append(f"      Start time: {tron_status.action_start_time}")
+    output.append(f"      Command: {tron_status.action_command}")
+    if verbose:
+        output.append(f"      Raw Command: {tron_status.action_raw_command}")
+        output.append(f"      Stdout: {tron_status.action_stdout}")
```
qui: I think this should only show on two v's. Also put a newline before the stdout because it is several lines.

paasta_tools/cli/cmds/status.py
```diff
@@ -489,6 +519,11 @@ def report_status_for_cluster(
         in SSH_ONLY_INSTANCE_CONFIG
     ]
 
+    tron_jobs = [
```
qui: Why does this need to be done separately? Can you add Tron to HTTP_ONLY_INSTANCE_CONFIG?

EmanekaT: tron is not in deployed_instances/actual_deployments. It's less misleading in my opnion to add tron sperately than adding tron to deployed instances, as tron is not really deployed. Or I could rewrite the code as it's kinda dirty right now but that is too time consuming.

paasta_tools/cli/cmds/status.py
EmanekaT: ?

EmanekaT:
I didn't add any test that checks print_tron_status as I think that's unnecessary and no existing tests check other print_* functions.

paasta_tools/cli/cmds/status.py
qui: It should not have a capital "B" (should be "b").

paasta_tools/cli/cmds/status.py
qui: Raw command should be with verbose == 1.  Also, can you also add a new line after the ":" so it looks like the following?
```
Stdout:
logging.info ...
logging.debug ...

Stderr:
Exception!
```

EmanekaT:
> Can you test the output one more time and paste the result here?

https://fluffy.yelpcorp.com/i/FJwlKlnRZGC2qx8WXwnKpg4mKGNdlDMW.html

PR #2181, opened 2019-03-20
add spark-run option for mrjobs

stug:
This is roughly ready but I'm still doing some full tests of using this to run mrjobs on an actual cluster.

### DIFF
--- a/tests/cli/test_cmds_spark_run.py
+++ b/tests/cli/test_cmds_spark_run.py
@@ -85,7 +85,7 @@ def test_get_spark_config(
     )
 
     assert spark_conf['spark.master'] == 'mesos://fake_leader:5050'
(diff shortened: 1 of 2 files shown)


tests/cli/test_cmds_spark_run.py
```diff
@@ -123,11 +122,17 @@ class TestConfigureAndRunDockerContainer:
         'fake_dir',
     )
 
+    @pytest.fixture
+    def mock_create_spark_config_str(self):
+        with mock.patch(
+            'paasta_tools.cli.cmds.spark_run.create_spark_config_str'
```
lydian: add `autospec=True`

stug: yep, good call.  TravisCI is telling me this, too :-/

PR #2184, opened 2019-03-21
Rename the deployd inbox_q to 'instances_that_need_to_be_bounced_in_the_future'

solarkennedy:
I think this is an easier to understand name, in the code and on the graphs.

I'll make one PR per rename for my own sanity and so we can bikeshed things one at a time.

### DIFF
--- a/paasta_tools/deployd/master.py
+++ b/paasta_tools/deployd/master.py
@@ -58,11 +58,11 @@ class DedupedPriorityQueue(PaastaPriorityQueue):
 
 
 class Inbox(PaastaThread):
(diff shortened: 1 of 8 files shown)


qui:
Hm, this naming still confuses me. How about... inbox_q -> instances_to_process?

solarkennedy:
Hehe. But I don't like "process", does that mean they need to be bounced? Or that they just need to be looked at to see if they need to be bounced?

solarkennedy:
Correct. It is the second.
The names of these queue make most sense when you look at all three at a time maybe.

Let me do this one, and then the other ones, and then make it will make more sense.

PR #2185, opened 2019-03-21
WIP: Convert from pyinotify to inotify

solarkennedy:


paasta_tools/deployd/watchers.py
```diff
@@ -110,55 +108,145 @@ class SoaFileWatcher(PaastaWatcher):
...
     @property
     def mask(self):
+        mask = inotify.constants.IN_ALL_EVENTS
         boring_flags = ['IN_CLOSE_NOWRITE', 'IN_OPEN', 'IN_ACCESS', 'IN_ATTRIB']
-        return reduce(
-            lambda x, y: x | y,
-            [v for k, v in pyinotify.EventsCodes.OP_FLAGS.items() if k not in boring_flags],
-        )
+        for flag in boring_flags:
+            mask = mask - inotify.constants.boring_flag
```
keymone: `mask = mask - getattr(inotify.constants, flag)`

keymone: also probably cache the value or even pre-calculate it into a property during loading/initialization, doesn't seem like it changes ever

paasta_tools/deployd/watchers.py
```diff
@@ -110,55 +108,145 @@ class SoaFileWatcher(PaastaWatcher):
...
```
keymone: same comments as above

paasta_tools/deployd/watchers.py
```diff
@@ -110,55 +108,145 @@ class SoaFileWatcher(PaastaWatcher):
...
+        return mask
 
     def run(self):
-        self.notifier.loop(callback=self.startup_checker)
-
-    def startup_checker(self, obj):
-        self.is_ready = True
```
mattmb: I think this is part of the "contract" and I don't see it moved else where.

Essentially once the watcher is confident that it will not miss any new events it marks itself ready so that the main thread can move on.

paasta_tools/deployd/watchers.py
mattmb: +1, doesn't look quite right

paasta_tools/deployd/watchers.py
```diff
@@ -110,55 +108,145 @@ class SoaFileWatcher(PaastaWatcher):
...
+        return mask
 
     def run(self):
-        self.notifier.loop(callback=self.startup_checker)
+        i = inotify.adapters.InotifyTree(path=PATH_TO_SYSTEM_PAASTA_CONFIG_DIR, mask=self.mask)
+        for event in i.event_gen():
+            self.process_event(event)
 
-    def startup_checker(self, obj):
-        self.is_ready = True
```
mattmb: as above

PR #2188, opened 2019-03-27
Remove characters that are not allowed in env vars

qui:
Secret names with dashes are commonly used, like `my-secret`. But dashes aren't allowed in bash variables.

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import os
+import re
 import sys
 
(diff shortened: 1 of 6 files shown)


solarkennedy:
What if this was enforced at the schema level instead?

qui:
How about both? I'll add it to the schema, but it would be nice if the suggestions were correct, too.

qui:
The message isn't super helpful, but it works:
```
✗ Failed to validate schema. More info: http://paasta.readthedocs.io/en/latest/yelpsoa_configs.html: /nail/home/qui/configs/yelpsoa-configs/compute-infra-test-service/chronos-mesosstage.yaml
  Validation Message: Additional properties are not allowed ('MMB-SECRET' was unexpected)
```

PR #2191, opened 2019-04-03
Notify authors via slack with auto-rollback

solarkennedy:


solarkennedy:
I figured it would just be easier to always mock this out.

tests/cli/test_cmds_mark_for_deployment.py
```diff
@@ -36,6 +37,12 @@ class fake_args:
     timeout = 10.0
 
 
+@fixture(autouse=True, scope='session')
```
EvanKrall: `autouse=True, scope='session' ` makes me feel dirty:

 - these four lines at the top of the file magically apply to tests at the bottom of the file (300+ lines away).
- doesn't this make it impossible to test `get_authors_to_be_notified`?
- `scope='session'` will mean that the mock will never be reset - so if someone modifies the return value it won't be undone for the next test.

In general, I prefer tests to be independent more than I like DRY test code. Specifically, I like to be able to know exactly what a test is doing just by looking at that test; not having to scan the whole file (or any plugins that get installed) to know what might be mocked out.

paasta_tools/slack.py
```diff
@@ -72,3 +73,18 @@ def get_slack_client():
     if token is None:
         token = optionally_load_system_paasta_config().get_slack_token()
     return PaastaSlackClient(token=token)
+
+
+def get_authors_to_be_notified(git_url, from_sha, to_sha):
```
EvanKrall: I'm not convinced this is better off here than in mark_for_deployment.py.

tests/cli/test_cmds_mark_for_deployment.py
solarkennedy: Ok

paasta_tools/slack.py
solarkennedy: Ok

solarkennedy:
@EvanKrall  is the this the one you would rather me not ship? Would you rather me duplicate the authors method over instead of turning into a regular function?

PR #2195, opened 2019-04-08
Make it possible to test paasta api locally with local client

EmanekaT:
I tested with local client and it works.


example_cluster/setup_fake_config.py
```diff
@@ -0,0 +1,57 @@
+import json
+import os
+import shutil
+import socket
+import subprocess
+import sys
+from distutils.dir_util import copy_tree
+
+
+def find_open_ports():
```
solarkennedy: `paasta_tools.cli.utils.pick_random_port`

example_cluster/setup_fake_config.py
```diff
@@ -0,0 +1,57 @@
...
+
+def main():
+
+    print('-------------------------------------------------------')
+    print('Please run echo $$PAASTA_SYSTEM_CONFIG_DIR to continue')
+    print("Please set environment variable PAASTA_TEST_CLUSTER to the cluster you want to use.")
+    print("This is necessary for tron jobs")
+    print('-------------------------------------------------------')
+    cluster = os.environ.get('PAASTA_TEST_CLUSTER', 'norcal-devc')
+    config_path = 'fake_config'
```
solarkennedy: call this `etc_paasta_for_development`, and gitignore it

example_cluster/setup_fake_config.py
```diff
@@ -0,0 +1,57 @@
...
+        # execute paasta api command
+        p = subprocess.Popen(
+            ['paasta-api', '-D', '-c', cluster, str(port)],
+            stderr=subprocess.STDOUT, stdout=subprocess.PIPE,
+        )
+        for line in iter(p.stdout.readline, b''):
+            sys.stdout.write(line.decode(sys.stdout.encoding))
+    finally:
+        # clean up
+        os.environ.pop('PAASTA_TEST_CLUSTER', None)
```
solarkennedy: You don't need to do this, it isn't like it is going to persist past this pid

example_cluster/setup_fake_config.py
```diff
@@ -0,0 +1,57 @@
...
+        shutil.rmtree(config_path)
```
solarkennedy: I don't think you have to do this either. Just leave it behind

example_cluster/setup_fake_config.py
```diff
@@ -0,0 +1,57 @@
...
+    api_endpoints = {'api_endpoints': {cluster: f'http://localhost:{port}'}}
+    os.remove(config_path + '/api_endpoints.json')
+    with open(config_path + '/api_endpoints.json', 'w') as f:
+        json.dump(api_endpoints, f)
+
+    try:
+        # export config path
+        os.environ['PAASTA_SYSTEM_CONFIG_DIR'] = config_path
+        # execute paasta api command
+        p = subprocess.Popen(
```
solarkennedy: How about just `exec`?

example_cluster/setup_fake_config.py
```diff
@@ -0,0 +1,57 @@
...
+    copy_tree('/etc/paasta', config_path)
+    # Generate tron.json
+    tron_config = {'tron': {'url': f'http://tron-{cluster}:8089'}}
+    with open(config_path + '/tron.json', 'w') as f:
+        json.dump(tron_config, f)
+    # find unused port
+    port = find_open_ports()
+    # Generate api endpoints
+    api_endpoints = {'api_endpoints': {cluster: f'http://localhost:{port}'}}
+    os.remove(config_path + '/api_endpoints.json')
```
solarkennedy: You can just overwrite the file, do you have to bother removing it first?

example_cluster/setup_fake_config.py
```diff
@@ -0,0 +1,57 @@
+import json
```
solarkennedy: how about call this `run-paasta-api-in-dev-mode` ?

Makefile
```diff
@@ -30,6 +30,9 @@ docs: .paasta/bin/activate
 test: .paasta/bin/activate
 	.paasta/bin/tox -i $(PIP_INDEX_URL)
 
+api: .paasta/bin/activate
```
solarkennedy: how about `dev-api` ? It will make room for other `dev-` targets

Makefile
```diff
@@ -30,6 +30,9 @@ docs: .paasta/bin/activate
...
+	python3.6 ./example_cluster/setup_fake_config.py
```
solarkennedy: This assumes the virtualenv is active.
Instead call `.paasta/bin/python ./example...`

Makefile
```diff
@@ -30,6 +30,9 @@ docs: .paasta/bin/activate
...
+
 .paasta/bin/activate: requirements.txt requirements-dev.txt
```
solarkennedy: Although I don't really know why we have this in the first place. This is what tox "does" ?

example_cluster/setup_fake_config.py
EmanekaT: Isn't exec used to execute python code? I'll need to manually parse arguments then.

solarkennedy: That is `eval`.  Exec makes it so a new process "takes over"

https://docs.python.org/3/library/os.html#process-management

Makefile
EmanekaT: which tox? tox in .paasta/bin/activate?

solarkennedy: You can discard this comment. If we wanted to, we could use tox to create this virtualenv (or just use the tox's py36 venv) instead of manually calling pip here.

example_cluster/setup_fake_config.py
EmanekaT: Since I copied the file from etc/paasta, I'll need to chmod if I don't remove.

example_cluster/setup_fake_config.py
EmanekaT: ephemeral_port_reserve  is not found even though it's in requirements.txt... It works if I run the script with .tox/py36-linux/bin/activate && python -m ...

solarkennedy: We can pair together to fix this if you want, but this represents a stale venv. You can `rm -rf .paasta` and make `make` recreate it to get an updated venv.

example_cluster/setup_fake_config.py
EmanekaT: I'm not sure if I can do it. `I tried os.execl('.tox/py36-linux/bin/python', 'python', '-m', 'paasta_tools.api.api', *['-D', '-c', cluster, str(port)])` which again does not have the installed dependency.

Makefile
```diff
@@ -30,6 +30,13 @@ docs: .paasta/bin/activate
 test: .paasta/bin/activate
 	.paasta/bin/tox -i $(PIP_INDEX_URL)
 
+dev-api: .paasta/bin/activate
+	@if [ ! -d .tox/py36-linux ] ; \
```
solarkennedy: I think this is "fine", but isn't really in the spirit of `make`.

Instead, use make's ability to understand dependencies and conditionally run them, just like `.paasta/bin/activate`:

```
.tox/py36-linux:
    .paasta/bin/tox...

dev-api:  .tox/py36-linux
   .tox/py36....
```

PR #2196, opened 2019-04-09
Have the wait-for-deployment state machine also wait for rollback.

EvanKrall:
This doesn't yet introduce the `abandon`/`complete` states, but it does let you roll back while waiting for deployment, and roll forward when waiting for rollback.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -783,7 +783,6 @@ class MarkForDeploymentProcess(automatic_rollbacks.DeploymentProcess):
         self.update_slack_status(f"Marking `{self.commit[:8]}` for deployment for {self.deploy_group} failed. Please see Jenkins for more output.")  # noqa E501
 
     def on_enter_deploying(self):
-        self.update_slack()
         # if self.block is False, then deploying is a terminal state so we will promptly exit.
         # Don't bother starting the background thread in this case.
         if self.block:
@@ -809,7 +808,6 @@ class MarkForDeploymentProcess(automatic_rollbacks.DeploymentProcess):
             self.trigger('mfd_succeeded')
 
     def on_enter_rolling_back(self):
-        self.update_slack()
         if self.block:
             thread = Thread(target=self.do_wait_for_deployment, args=(self.old_git_sha,), daemon=True)
             thread.start()
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -29,6 +29,7 @@ from threading import Event
 from threading import Thread
 from typing import Collection
 from typing import Iterator
+from typing import List
 from typing import Mapping
 from typing import Optional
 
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -715,68 +748,106 @@ def status_code_by_state(self) -> Mapping[str, int]:
...
+
+    def get_available_buttons(self) -> List[str]:
+        buttons = []
+
+        if self.is_terminal_state(self.state):
+            # If we're about to exit, always clear the buttons, since once we exit the buttons will stop working.
+            return []
+
+        if self.old_git_sha != self.commit:
+            if self.state in self.rollback_states:
```
solarkennedy: I was kinda hoping we could have both buttons consistently on or off, even if the other button doesn't "do" anything, for the sake of UI stability and not making buttons disappear out from underneath you. 

What do you think? Would it be less confusing to have a button that doesn't do anything, or would it be less confusing to only have a button that can do something?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -715,68 +748,106 @@ def status_code_by_state(self) -> Mapping[str, int]:
...
 
-    def on_enter_deploy_aborted(self):
+        if self.mark_for_deployment_return_code != 0:
+            self.trigger('mfd_failed')
+        else:
+            self.update_slack_status(f"Marked `{self.old_git_sha[:8]}` for {self.deploy_group}.")
+            self.trigger('mfd_succeeded')
+
+    def on_enter_rolling_back(self):
+        self.update_slack()
```
solarkennedy: do you need this since you have the mega update_state_change?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -715,68 +748,106 @@ def status_code_by_state(self) -> Mapping[str, int]:
...
+        if self.block:
+            thread = Thread(target=self.do_wait_for_deployment, args=(self.old_git_sha,), daemon=True)
+            thread.start()
+
+    def on_exit_rolling_back(self):
+        self.wait_for_deployment_green_light.clear()
```
solarkennedy: Is there a race condition between having the light cleared, switching from back or forward, and then having the light uncleared?

In other words, can we switch states fast enough for either of the wait_for_deployment threads from noticing? And does it matter?

EvanKrall: Yeah I was wondering about this. I couldn't get it to happen in practice, but that probably doesn't mean it's impossible. I guess there are two clean ways to deal with this:

1. use separate "green lights" for deployment and rollback
2. use a lock to make sure only one wait_for_deployment is happening at a time

I think I'm leaning towards option 1

solarkennedy: Hehe, if only we could kill threads!

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: I think it would be confusing to have buttons that do nothing. Especially since they wouldn't cause us to re-mark-for-deployment or anything that might help us recover from weird situations like "someone ran `paasta rollback` in a different terminal".

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: Can you take another look? I went with option 1.

PR #2197, opened 2019-04-11
Add complete/abandon states

EvanKrall:
Add complete and abandon states; make deployed/rolled_back not terminal.
This makes the user click a button before we exit, giving them the chance
to roll back even once the deploy is at 100%.

### DIFF
--- a/paasta_tools/automatic_rollbacks.py
+++ b/paasta_tools/automatic_rollbacks.py
@@ -9,6 +9,7 @@ from typing import Collection
 from typing import Iterator
 from typing import List
 from typing import Mapping
+from typing import Optional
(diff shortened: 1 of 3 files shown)


EvanKrall:
(Note that this doesn't yet automatically move forward from `deployed` to `complete`, or from `rolled_back` to `abandon`, so it will block until someone clicks a button in slack.

solarkennedy:
I don't want this in place unless it happens automatically after a timeout. Otherwise people have two buttons to press to move things along, and it isn't "obvious" that they forgot to press this one.

EvanKrall:
@solarkennedy Timeouts added; please re-review.

PR #2198, opened 2019-04-11
Make kill_bad_containers handle the edge case where a second container is missing

solarkennedy:
It looks like this happens every once in a while in the past 30 days. 
I don't know why, so I've made the script print out what it does know and exit non-zero so we'll get a ticket. (instead of stack tracing)


test
```diff
@@ -0,0 +1,161 @@
+2019-04-08T19:31:01-07:00 <user.notice> 10-69-204-110-uswest2cprod cron_kill_bad_containers:     sys.exit(main())
```
chlgit: Do you really want to commit this file?

PR #2203, opened 2019-04-15
Add script for notifying teams about long-running Spark jobs

qui:
Planning on running this as a cron job on masters and sending the output of the script to our team. The output will allow us to follow up with teams in case the jobs continue to run or some jobs don't correspond to valid services.

### DIFF
--- a/paasta_tools/check_spark_jobs.py
+++ b/paasta_tools/check_spark_jobs.py
@@ -42,8 +42,10 @@ def get_time_running(framework):
 
 def get_spark_properties(framework):
     webui_url = framework.get('webui_url')
-    env_endpoint = f'{webui_url}/api/v1/applications/{framework.id}/environment'
+    if not webui_url:
+        return None
 
+    env_endpoint = f'{webui_url}/api/v1/applications/{framework.id}/environment'
     try:
         response = requests.get(env_endpoint, timeout=5)
     except (requests.exceptions.Timeout, requests.exceptions.ConnectionError) as e:
@@ -56,8 +58,8 @@ def get_spark_properties(framework):
 
     try:
         return response.json()['sparkProperties']
-    except (ValueError, KeyError) as e:
-        logger.warning(f'Unable to get sparkProperties for {framework.id}: {e!r}')
+    except (ValueError, KeyError):
+        logger.warning(f'Unable to get sparkProperties for {framework.id}: got response {response.text}')
         return None
 
 
@@ -149,15 +151,20 @@ def report_spark_jobs(min_hours, no_notify):
     )
     messages_by_service = get_messages_by_service(results)
 
-    print('\n\n'.join(messages_by_service.values()))
-
-    if not no_notify:
-        valid_services = set(list_services())
-        for service, message in messages_by_service.items():
-            if service in valid_services:
+    valid_services = set(list_services())
+    messages_for_unknown_services = []
+    for service, message in messages_by_service.items():
+        if service in valid_services:
+            print(f'{message}\n')
+            if not no_notify:
                 notify_framework_owners(service, message)
-            else:
-                print(f'Did not notify for invalid service: {service}')
+        else:
+            messages_for_unknown_services.append(message)
+
+    print('\nINVALID SERVICES')
+    print('----------------')
+    print('The following frameworks are associated with services that are not configured in PaaSTA.\n')
+    print('\n\n'.join(messages_for_unknown_services))
 
 
 def main():
--- a/paasta_tools/check_spark_jobs.py
+++ b/paasta_tools/check_spark_jobs.py
@@ -1,0 +1,170 @@
+import argparse
+import datetime
+import logging
+from collections import defaultdict
+
+import requests
+
+from paasta_tools import mesos_tools
+from paasta_tools.monitoring_tools import send_event
+from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import list_services
+
+
+logger = logging.getLogger(__name__)
+
+
+JUPYTER_PREFIX = 'jupyterhub_'
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description='Reports long-running Spark frameworks.',
+    )
+    parser.add_argument(
+        '--min-hours',
+        type=float,
+        help='Report frameworks that have been registered for more than this duration',
+        default=0,
+    )
+    parser.add_argument(
+        '--no-notify',
+        action='store_true',
+        help='Skip notifying the teams that own each framework',
+    )
+    return parser.parse_args()
+
+
+def get_time_running(framework):
+    registered_time = datetime.datetime.fromtimestamp(framework['registered_time'])
+    return datetime.datetime.now() - registered_time
+
+
+def get_spark_properties(framework):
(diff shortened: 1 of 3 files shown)


paasta_tools/check_spark_jobs.py
```diff
@@ -0,0 +1,170 @@
...
+    else:
+        return service
+
+
+def get_matching_framework_info(min_hours):
+    frameworks = mesos_tools.get_all_frameworks(active_only=True)
+    matching_info = []
+    min_timedelta = datetime.timedelta(hours=min_hours)
+    for framework in frameworks:
+        if not framework.active:
```
solarkennedy: I think this is fine, but with `active_only`, do we need this?

paasta_tools/check_spark_jobs.py
```diff
@@ -0,0 +1,170 @@
...
+            continue
+        if framework.get('principal') != 'spark':
+            continue
+        time_running = get_time_running(framework)
+        if time_running >= min_timedelta:
+            info = {
```
solarkennedy: This function is good. I think it might be easier to understand and test if it *only* filtered long running spark frameworks from an input list of frameworks and literally returned the framework objects, and let some other function worry about extracting the pieces of data you need (service, url, etc).

But it does mean you would have to `get_time_running` more than once, which isn't super efficient. 

But in general, this function does 3 thing, fetches, filters, and creates a custom return object. Again I think this is just fine for this script, but was just a thought I had as I was reading the code.

paasta_tools/check_spark_jobs.py
```diff
@@ -0,0 +1,170 @@
...
+def notify_framework_owners(service, output):
+    overrides = {
+        'page': False,
+        'alert_after': 0,
+        'tip': 'Ask the user to check the job UI and terminate the job if appropriate.',
+        'runbook': 'http://y/spark-debug',
+    }
+    send_event(
+        service=service,
+        check_name=f'long_running_spark_jobs.{service}',
```
solarkennedy: This choice of check name is interesting.
On the plus side, we don't have to worry about some sort of "primary key" on the spark job (we will have plenty of adhoc things), so using service means we don't have to worry about lots of different unique sensu events.
On the minus side, if we have multiple long-running jobs per service we'll have grouping, and the check won't resolve until *all* the jobs are killed. They may cause confusion for users.

All this to say: I support your choice of check name

paasta_tools/check_spark_jobs.py
```diff
@@ -0,0 +1,170 @@
...
+        return None
+
+    if response.status_code != 200:
+        logger.warning(f'Bad response from {env_endpoint}: {response.status_code}')
+        return None
+
+    try:
+        return response.json()['sparkProperties']
+    except (ValueError, KeyError) as e:
+        logger.warning(f'Unable to get sparkProperties for {framework.id}: {e!r}')
```
solarkennedy: It might be helpful to print the response (not json) to debug to help figure out "why" we are getting a Error here.

paasta_tools/check_spark_jobs.py
stug: I think the `active_only` kwarg makes it not return the `completed_frameworks`, but the `frameworks` can still still have `"active": False`

paasta_tools/check_spark_jobs.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def get_time_running(framework):
+    registered_time = datetime.datetime.fromtimestamp(framework['registered_time'])
+    return datetime.datetime.now() - registered_time
+
+
+def get_spark_properties(framework):
+    webui_url = framework.get('webui_url')
+    env_endpoint = f'{webui_url}/api/v1/applications/{framework.id}/environment'
```
stug: Given that the code is not sure that the framework _has_ a `webui_url`, maybe this should bail if it doesn't?

I'm pretty sure all spark frameworks should have one, but since we're using `.get()` there's a chance we get back `None`

paasta_tools/check_spark_jobs.py
```diff
@@ -0,0 +1,170 @@
...
+        min_hours=min_hours,
+    )
+    messages_by_service = get_messages_by_service(results)
+
+    print('\n\n'.join(messages_by_service.values()))
+
+    if not no_notify:
+        valid_services = set(list_services())
+        for service, message in messages_by_service.items():
+            if service in valid_services:
```
stug: nit: given that `guess_service` can return `None`, maybe `get_matching_framework_info` or `get_messages_by_service` should filter those out before we get here.   Currently I think the way this'll go is that we'll get some entries in `messages_by_service` for a `None` service and then say that we didn't notify for service: `None`.

It'd maybe be nicer to handle that case explicitly: notice that we couldn't figure out the service for a framework and report that it couldn't figure out the service for the framework with name x, webui_url y.

Unsure how likely this is to happen, though.

paasta_tools/check_spark_jobs.py
qui: ^ Yep, that's what I found.

paasta_tools/check_spark_jobs.py
qui: Thanks for the support :) I think it's ok for the check not to resolve until everything is killed, since the text says the alert is for a particular service and lists all the frameworks.

That reminds me, do I need to send OK events for all services that **don't** have long-running jobs? As far as I can tell, if I don't, these alerts won't resolve.

paasta_tools/check_spark_jobs.py
qui: I think it will be relatively common at first. One case is where the service is None, If we're unable to get the value of PAASTA_SERVICE through the Spark URL. It can also be invalid if PAASTA_SERVICE has a malformed value, which is pretty common in our jupyter setups. For example, the cookiecutter jupyter configs add the user name in some cases, or use dashes instead of underscores for the service name.

I separated these out into a separate section in the report output, so it's more clear that we need to take action on them ourselves. It includes the name and url. All None messages would show up there too, under one heading for "service None". Does that address your concern?

Here's what some test output looks like:
```
Found the following long-running Spark frameworks associated with service service1.
Please check why they are still running and terminate if appropriate.

spark1 (running for 7:00:00)
  user: test_user
  job UI: url1

Found the following long-running Spark frameworks associated with service service2.
Please check why they are still running and terminate if appropriate.

spark2 (running for 7:00:00)
  user: test_user
  job UI: url2
spark3 (running for 7:00:00)
  user: test_user
  job UI: url3


INVALID SERVICES
----------------
The following frameworks are associated with services that are not configured in PaaSTA.

Found the following long-running Spark frameworks associated with service service_dne.
Please check why they are still running and terminate if appropriate.

spark3 (running for 7:00:00)
  user: test_user
  job UI: url3
```

paasta_tools/check_spark_jobs.py
qui: Yeah, I like the slight efficiency for now, but if the return output gets too much more complicated in the future, I'll separate it.

paasta_tools/check_spark_jobs.py
solarkennedy: Haha! Yes, you need to send OK events. (I forgot too)

PR #2205, opened 2019-04-18
move rollback command earlier

drmorr0:
Print the command to roll back at the same time that we send the slack message.


solarkennedy:
I'm not super sure when the "best" time to print this command is.
At the beginning, sure, seems like you are "prepared" with a rollback command if anything goes wrong.
At the end, seems sensible, because after a deploy if you want to roll back, it will be the "last" thing you see.

I think it should be printed twice.

PR #2206, opened 2019-04-18
Renamed 'inbox' gauge to 'instances_that_need_to_be_checked_up_on'

solarkennedy:
`instances_that_need_to_be_checked_up_on` is the best string I can think of that represents what the `inbox.to_bounce` list "is".

### DIFF
--- a/paasta_tools/deployd/master.py
+++ b/paasta_tools/deployd/master.py
@@ -58,12 +58,12 @@ class DedupedPriorityQueue(PaastaPriorityQueue):
 
 
 class Inbox(PaastaThread):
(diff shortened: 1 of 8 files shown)


solarkennedy:
You are right. That is more clear now that I've gotten to the last part of the rename.

solarkennedy:
Well, we didn't know they were instances before. Maybe they were services?

EvanKrall:
How about `instances_to_bounce_now` and `instances_to_bounce_later`?

solarkennedy:
updated to krall's suggestions

I think all of these are better than "inbox"

PR #2214, opened 2019-04-26
[poc] Add possibility to dump Kubernetes objects instead of creating/updating them in setup_kubernetes_cr

keymone:


paasta_tools/setup_kubernetes_cr.py
```diff
@@ -85,21 +134,27 @@ def main() -> None:
...
     )
     sys.exit(0 if setup_kube_succeeded else 1)
 
 
 def setup_all_custom_resources(
     kube_client: KubeClient,
     soa_dir: str,
-    system_paasta_config: SystemPaastaConfig,
+    cluster: str,
+    custom_resources: str,
```
mattmb: Not a string right?

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -85,21 +134,27 @@ def main() -> None:
         logging.basicConfig(level=logging.INFO)
 
     kube_client = KubeClient()
+    if args.output_to_std:
```
mattmb: might as well be an `if` `else` I guess?

keymone: in that  case `KubeClient()` instantiation would be duplicated

PR #2217, opened 2019-05-02
Bump the drain timeout to 10min

solarkennedy:
I think it isn't so bad to have a long drain timeout.

When things go bad, it is nice that a large bounce can still make progress, even under heavy load.

Otherwise, we can get into a situation where we can *never* make progress!

### DIFF
--- a/paasta_tools/drain_lib.py
+++ b/paasta_tools/drain_lib.py
@@ -165,7 +165,7 @@ class HacheckDrainMethod(DrainMethod):
         service: str,
         instance: str,
         registrations: List[str],
-        delay: float = 240,
+        delay: float = 600,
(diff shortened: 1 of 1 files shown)


EvanKrall:
you probably want to change https://github.com/Yelp/paasta/pull/2217/files#diff-ab112c5181a013143e3f3dadd2152d27R176 instead?

solarkennedy:
Wait, so we have a ttl of 40 minutes? That sounds like it should be long enough.

EvanKrall:
Yeah, probably fine as is, though I'm not sure we really need the expiration at all - setup_marathon_job knows how to un-drain tasks if necessary.

PR #2223, opened 2019-05-07
Make kill_bad_containers kill both containers for good measure when dupes are detected

solarkennedy:


qui:
I think we should probably also take more action in this case: https://github.com/Yelp/paasta/blob/c5e994d7560dac45bf7d54f3e51fa8ec6393951d/paasta_tools/contrib/kill_bad_containers.py#L58

But we can see what comes up once we start paging on failures and figure out what to do in that case.

PR #2230, opened 2019-05-09
Kubernetes secrets

mattmb:
This script takes a service, iterates through the paasta secrets on disk
and "syncs" them into kubernetes secrets. These are then native k8s
secrets that can be referenced as env vars/create tmpfs volumes.

I'm also saving the secret signature in a ConfigMap. This is so whatever
is watching for changes in secrets doesn't need to watch the secret
itself (and hence have access) to the secret to know if it needs to
bounce the service in question.


paasta_tools/kubernetes_scripts/paasta_secrets_sync.py
```diff
@@ -0,0 +1,186 @@
+#!/usr/bin/env python
```
Rob-Johnson: paasta_tools/kubernetes/bin ?

paasta_tools/secret_providers/vault.py
```diff
@@ -140,3 +142,36 @@ def decrypt_secret(self, secret_name: str) -> str:
...
             context=self.service_name,
         ).decode('utf-8')
+
+    def decrypt_secret_raw(self, secret_name: str) -> bytes:
+        ecosystem = self.get_vault_ecosystems_for_clusters()[0]
+        secret_path = os.path.join(
+            self.secret_dir,
+            f"{secret_name}.json",
+        )
+        if not hasattr(self, 'client'):
```
vkhromov: Is there a reason why you're doing this `hasattr()` check and not usual `self.client = None` in `__init__()` and then `if self.client is not None:` here?

paasta_tools/kubernetes_scripts/paasta_secrets_sync.py
vkhromov: but why @robj?

paasta_tools/secret_providers/vault.py
mattmb: ug, yeah I wasn't super happy with this either. I think it was something like when decrypting we only connect to 1 cluster but encrypting we talk to multiple. Anyway, let me re-visit and see if I can come up with something less ugly.

paasta_tools/kubernetes_scripts/paasta_secrets_sync.py
mattmb: yeah, I got myself in a mess here. initially I went with `kubernetes/` and was hoping I could move some other stuff in a follow-up PR but it conflicts with the imported `kubernetes` package. I guess just put this in the top-level dir with everything else?

Rob-Johnson: just because that top level dir has *everything* in it - it'd be nice if we had a paasta_tools.kubernetes module in paasta-tools, with it's own bin dir

paasta_tools/kubernetes_tools.py
```diff
@@ -1246,3 +1329,109 @@ def fromstring(cls, _str: str) -> int:
...
+def update_secret(
+    kube_client: KubeClient,
+    secret: str,
+    service: str,
+    secret_provider: BaseSecretProvider,
+) -> None:
+    service = sanitise_service_name(service)
+    kube_client.core.replace_namespaced_secret(
+        name=f"paasta-secret-{service}-{secret}",
+        namespace="paasta",
```
Rob-Johnson: are we going to have to do some rbac bits to allow operators in other ns to get this?

paasta_tools/kubernetes_tools.py
```diff
@@ -1246,3 +1329,109 @@ def fromstring(cls, _str: str) -> int:
...
+            metadata=V1ObjectMeta(
+                name=f"paasta-secret-{service}-{secret}-signature",
+                labels={"yelp.com/paasta_service": service},
+            ),
+            data={'signature': secret_signature},
+        ),
+    )
+
+
+def sanitise_service_name(
```
Rob-Johnson: surely we have this somewhere else already

paasta_tools/kubernetes_tools.py
mattmb: That's a good question. I just found `They can only be referenced by pods in that same namespace` as a snippet in the docs so I guess initially we can't with this code...

I guess we'll have to parametrise this and sync to multiple namespaces...

Maybe we could even have a map of `k8s namespace` =>`[]services` and `paasta` => `*` but other more specific namespaces would only sync a few services secrets?

I'll probably tackle this as a follow up ticket if you don't object

paasta_tools/kubernetes_tools.py
mattmb: not exactly unless I've missed something. we have https://github.com/Yelp/paasta/blob/c44d689242df366d765cc734dfa5b083b4481a28/paasta_tools/marathon_tools.py#L974 for example

I think it's okay to live here, it is kinda k8s specific even if by coincidence we do something similar with marathon app names.

paasta_tools/secret_providers/vault.py
mattmb: ok, just pushed a commit to clean this up a little!

PR #2236, opened 2019-05-11
Script for getting resource allocation for each task

qui:
I was thinking we could run this in a cron job, pipe the output of this into a scribe log, and ingest that into splunk.

The contents of this log should be enough to track how long each task runs and how many resources it uses. We can also join this with the DockerStats logs, for utilization data, through the mesos_container_id - the DockerStats include container_name, which is just `mesos-{mesos_container_id}`. 

Does this look reasonable?

### DIFF
--- a/paasta_tools/contrib/get_container_allocation.py
+++ b/paasta_tools/contrib/get_container_allocation.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python
+import a_sync
+
+from paasta_tools import mesos_tools
+from paasta_tools.utils import load_system_paasta_config
+
+
+def get_mesos_container_id_from_task(task):
+    for status in task['statuses']:
+        container_id = status.get('container_status', {}).get('container_id', {}).get('value')
+        if container_id:
+            return container_id
+
+
+def get_paasta_service_instance_from_task(task):
+    docker_params = task['container'].get('docker', {}).get('parameters', [])
+    service, instance = None, None
+    for param in docker_params:
+        if param['key'] == 'label':
+            label = param['value']
+            if label.startswith('paasta_service='):
+                service = label.split('=')[1]
+            if label.startswith('paasta_instance='):
+                instance = label.split('=')[1]
+    return service, instance
+
+
+@a_sync.to_blocking
+async def get_task_allocation_info():
+    tasks = await mesos_tools.get_cached_list_of_running_tasks_from_frameworks()
+    info_list = []
+    for task in tasks:
+        info = {'resources': task['resources']}
+        info['mesos_container_id'] = get_mesos_container_id_from_task(task)
+        service, instance = get_paasta_service_instance_from_task(task)
+        info['service'] = service
+        info['instance'] = instance
+        info_list.append(info)
+
+
+def main():
+    cluster = load_system_paasta_config().get_cluster()
+    info_list = get_task_allocation_info()
+    for info in info_list:
+        info['cluster'] = cluster
+        print(info)
+
+
+if __name__ == '__main__':
+    main()
--- a/paasta_tools/contrib/get_running_task_allocation.py
+++ b/paasta_tools/contrib/get_running_task_allocation.py
@@ -1,15 +1,22 @@
 #!/usr/bin/env python
+import time
+
 import a_sync
+import simplejson as json
 
 from paasta_tools import mesos_tools
 from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.utils import paasta_print
 
 
-def get_mesos_container_id_from_task(task):
+def get_container_info_from_task(task):
     for status in task['statuses']:
+        if status['state'] != 'TASK_RUNNING':
+            continue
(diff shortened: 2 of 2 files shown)


paasta_tools/contrib/get_container_allocation.py
```diff
@@ -0,0 +1,50 @@
...
+async def get_task_allocation_info():
+    tasks = await mesos_tools.get_cached_list_of_running_tasks_from_frameworks()
+    info_list = []
+    for task in tasks:
+        info = {'resources': task['resources']}
+        info['mesos_container_id'] = get_mesos_container_id_from_task(task)
+        service, instance = get_paasta_service_instance_from_task(task)
+        info['service'] = service
+        info['instance'] = instance
+        info_list.append(info)
```
thebostik: based on the convo we had, is there more metadata we could grab from the task and stick into info, such as task launch time? it'd be useful, not sure if it's available.

paasta_tools/contrib/get_container_allocation.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def main():
+    cluster = load_system_paasta_config().get_cluster()
+    info_list = get_task_allocation_info()
+    for info in info_list:
+        info['cluster'] = cluster
```
thebostik: same naming idea, can we do `paasta_cluster` here?

paasta_tools/contrib/get_container_allocation.py
```diff
@@ -0,0 +1,50 @@
...
+
+@a_sync.to_blocking
+async def get_task_allocation_info():
+    tasks = await mesos_tools.get_cached_list_of_running_tasks_from_frameworks()
+    info_list = []
+    for task in tasks:
+        info = {'resources': task['resources']}
+        info['mesos_container_id'] = get_mesos_container_id_from_task(task)
+        service, instance = get_paasta_service_instance_from_task(task)
+        info['service'] = service
```
thebostik: just a naming suggestion, since we mostly changed to `paasta_service` / `paasta_instance` can we call it that here too?

paasta_tools/contrib/get_container_allocation.py
thebostik: in terms of other metadata to add on to `info`, i think we might also want a `time`, and maybe this machine's `hostname`? is there any other useful metadata we might want?

if we do ingest this output into splunk we could output it as a list, and just print the metadata once (not that the size of the output will be that big to worry about)

paasta_tools/contrib/get_container_allocation.py
```diff
@@ -0,0 +1,50 @@
...
+        info['instance'] = instance
+        info_list.append(info)
+
+
+def main():
+    cluster = load_system_paasta_config().get_cluster()
+    info_list = get_task_allocation_info()
+    for info in info_list:
+        info['cluster'] = cluster
+        print(info)
```
thebostik: not sure if it's absolutely required since i'm not familiar with all the return types of the info that we're grabbing, but it would be safest to do `json.dumps` on the `info` here, otherwise we might not be able to extract it all cleanly in splunk

paasta_tools/contrib/get_container_allocation.py
qui: @thebostik What do you mean by printing the metadata once?

thebostik: so related to the metadata that we need to tie to the data we print, such as the current time.

if we are iterating over the info_list, we'll be printing out each task separately.

instead, i was suggesting we could do a single print each time this script runs, and output json for later splunk ingestion in a format like:
```
{"time": 1557857743,
"paasta_master": "my_hostname",
"paasta_cluster": "pnw-prod",
... other metadata ...
"tasks": [
  {"paasta_service": "foo", "paasta_instance": "bar", "resources": ...},
  {"paasta_service": "baz", "paasta_instance": "bar", "resources": ...},
...
]}
```

paasta_tools/contrib/get_container_allocation.py
qui: Added task launch time and the paasta pool.

paasta_tools/contrib/get_container_allocation.py
qui: Hm, I don't think hostname will be that useful if we already know it's running on the paasta master for that cluster.

Is there an advantage to doing a single print, besides not repeating the metadata? There can be thousands of tasks, so I was thinking it would be nicer to have separate Splunk events for each one instead of one huge event.

thebostik: not much of an advantage otherwise that i can think of offhand. maybe a guarantee that the list was complete?
maybe separate is best! still i think we'll want to `json.dumps` anything we print as output, whether it's a single task or a list of them

qui: OK, I'll leave it this way for now. I think I added the json.dumps (at the end of main), unless you meant something else?

PR #2237, opened 2019-05-13
Support SSL for paasta-api client

mattmb:
This is so we can support SSL from the swagger client. This is only
enabled once we configure https endpoints for our API. It also supports
presenting client certificates. I tested this all with the
example-cluster and an nginx proxy and it seems to work fine.

### DIFF
--- a/paasta_tools/api/client.py
+++ b/paasta_tools/api/client.py
@@ -22,6 +22,7 @@ from typing import Any
 from urllib.parse import urlparse
 
 from bravado.client import SwaggerClient
+from bravado.requests_client import RequestsClient
 
 import paasta_tools.api
 from paasta_tools.utils import load_system_paasta_config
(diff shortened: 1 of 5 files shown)


paasta_tools/api/client.py
```diff
@@ -21,9 +21,9 @@
 from typing import Any
 from urllib.parse import urlparse
 
-from bravado.client import SwaggerClient
-
 import paasta_tools.api
+from bravado.client import SwaggerClient
```
vkhromov: 100500% nit: I was under impression that we keep different kind of imports in separate sections, that's way `bravado.client` was in it's own section: because it isn't a part of platform like`typing` and `urllib`, and also not a part of current project like `paasta_tools`, but a third-party library.  If that's true, we probably better keep that separation here.

paasta_tools/api/client.py
```diff
@@ -65,10 +65,39 @@ def get_paasta_api_client(
...
+    def __init__(
+        self,
+        scheme: str,
+        cluster: str,
+        system_paasta_config: SystemPaastaConfig,
+    ):
+        if scheme == 'https':
+            if system_paasta_config.get_enable_client_cert_auth():
+                home = os.environ.get('HOME', '')
+                ssl_cert = (f'{home}/paasta.crt', f'{home}/paasta.key')
```
vkhromov: cert/key should be different for different clusters, right?
Also, we probably better put them into a subdirectory inside `$HOME`, something like `$HOME/.paasta/pki/mesosstage.crt`?

paasta_tools/utils.py
```diff
@@ -1786,6 +1789,22 @@ def get_auto_hostname_unique_size(self) -> int:
...
+
+    def get_default_api_ca_certificate(self) -> str:
+        """ A default CA certificate to fall back to if there is not a cluster mapped in
+        get_api_ca_certificates
+        """
+        return self.config_dict.get('default_api_ca_certificate', '/etc/paasta/pki/ca.crt')
+
+    def get_enable_client_cert_auth(self) -> bool:
+        """
+        If enabled present a client certificate from $HOME/.paasta.crt nad $HOME/.paasta.key
```
vkhromov: `s/\<nad\>/and/`

vkhromov: filenames are different from the ones which are used in the code above.

paasta_tools/api/client.py
mattmb: 👍good point. I nearly made this configurable and was just being lazy because I didn't want to solve the $HOME substitution side. Let me take another look.

paasta_tools/utils.py
mattmb: Good spot!

paasta_tools/api/client.py
vkhromov: You can use something like
```
dir = os.path.expanduser('~/.paasta/pki')
ssl_cert = (f'{dir}/{cluster}.crt', f'{dir}/{cluster}.key')
```

mattmb: Nice, don't think I knew about `expanduser`!

paasta_tools/api/client.py
mattmb: Sure, I guess we kinda care but it's tricky to enforce in pre-commit!

paasta_tools/utils.py
```diff
@@ -1607,6 +1607,9 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     cluster: str
     dashboard_links: Dict[str, Dict[str, str]]
     api_endpoints: Dict[str, str]
+    api_ca_certificates: Dict[str, str]
+    default_api_ca_certificate: str
```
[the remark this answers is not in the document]
mattmb: Yeah, I think you're right. I'll leave it for now though seeing as its not doing much harm and it might be handy during testing/rollout.

PR #2239, opened 2019-05-14
move source of truth of valid clusters to config

Rob-Johnson:
rather than dynamically generating the list out of a regex, use the list
of clusters that is defined in system paasta config. This will stop files
distributed with yelpsoa-configs but not managed by paasta-tools from
causing issues when they have unexpected naming conventions

internal ticket PAASTA-15450 - accompanying puppet review /r/404806 which needs to be shipped first

### DIFF
--- a/example_cluster/paasta/clusters.json
+++ b/example_cluster/paasta/clusters.json
@@ -1,0 +1,1 @@
+{"clusters": ["testcluster"]}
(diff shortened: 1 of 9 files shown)


Rob-Johnson:
test breaks are real - I'm fixing

PR #2240, opened 2019-05-15
Watch SLOs for services and message the slack thread when they start failing.

EvanKrall:


paasta_tools/automatic_rollbacks/signalfx.py
```diff
@@ -0,0 +1,51 @@
...
+    return float(ts) / 1000
+
+
+def tail_signalfx(query: str, lookback_seconds: float, callback: Callable) -> None:
+    if lookback_seconds > 0:
+        start_timestamp_milliseconds = (time.time() - lookback_seconds) * 1000
+    else:
+        start_timestamp_milliseconds = None
+
+    with open('/nail/home/krall/.signalfx_token') as f:
```
solarkennedy: er. I'm pretty sure we have this as a system pasta config

paasta_tools/automatic_rollbacks/signalfx.py
```diff
@@ -0,0 +1,51 @@
...
+                # in SignalFX, query looks like ... .publish('<stream_label>')
+                stream_label: str = props.get('sf_streamLabel', tsid)
+
+                # sf_key seems to have the list of dimensions with a couple extra stuff
+                # The SignalFX query should minimally specify the 'service' dimension.
+                # .sum(by=['dim1','dim2'])
+                sf_key: List[str] = props.get('sf_key', [])
+
+                dimensions: Dict[str, str] = {dim: props[dim] for dim in sf_key}
+                tsid_metadata_map[tsid] = {'stream_label': stream_label, 'dimensions': dimensions}
```
solarkennedy: Is it guaranteed that we get the Metadata first? Because we use those variables in the elif

paasta_tools/automatic_rollbacks/slack.py
```diff
@@ -69,6 +56,33 @@ def get_slack_blocks_for_deployment(
             },
         },
     ]
+    if slo_watchers is not None and len(slo_watchers) > 0:
+        num_failing = len([w for w in slo_watchers if w.failing])
```
solarkennedy: eventually I would like this to be a function where we can make more elaborate messages and stuff

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -979,6 +985,23 @@ def after_state_change(self):
         self.update_slack()
         super().after_state_change()
 
+    def individual_slo_callback(self, label, bad):
+        if bad:
+            message = f"SLO started failing: {label}"
+        else:
+            message = f"SLO stopped failing: {label}"
```
solarkennedy: Can you rephrase this so it isn't a double-negative?
How about

    The {label} SLO is now OK.

yelp_package/extra_requirements_yelp.txt
```diff
@@ -1,7 +1,20 @@
 --extra-index-url=https://pypi.yelpcorp.com/simple
+cached-property==1.3.1
 clusterman_metrics==2.2.0
 crypto_lib==4.0.0
+gitdb2==2.0.3
```
solarkennedy: Are all these other things intentional?

yelp_package/extra_requirements_yelp.txt
```diff
@@ -1,7 +1,20 @@
...
+gitpython==2.1.9
+inflection==0.3.1
+markdown==2.4
+paasta-tools==0.84.3
```
solarkennedy: probably not this

solarkennedy:
This looks great as a first pass.
I'm a tad worried about failures with sfx, exceptions, etc.
How do you feel about making those threads "bulletproof" so that some sort of exception in them cannot possibly make mfd crash?

paasta_tools/automatic_rollbacks/signalfx.py
EvanKrall: oh, derp

paasta_tools/automatic_rollbacks/signalfx.py
EvanKrall: I'm not sure. Seems like in practice it happens that way. signalfx-python also collects metadata in the same fashion, and doesn't do any sort of error handling for missing metadata: https://github.com/signalfx/signalfx-python/blob/master/signalfx/signalflow/computation.py#L74

If it's a problem, we could queue the data messages until the metadata shows up.

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: > This looks great as a first pass.
> I'm a tad worried about failures with sfx, exceptions, etc.
> How do you feel about making those threads "bulletproof" so that some sort of exception in them cannot possibly make mfd crash?

Exceptions in threads don't bubble up to the main thread, so we should be okay there.

EvanKrall:
> This looks great as a first pass.
> I'm a tad worried about failures with sfx, exceptions, etc.
> How do you feel about making those threads "bulletproof" so that some sort of exception in them cannot possibly make mfd crash?

yelp_package/extra_requirements_yelp.txt
EvanKrall: oops, probably not that.

yelp_package/extra_requirements_yelp.txt
EvanKrall: slo-transcoder depends on gitpython which depends on gitdb2.

The intent here was to pin all the dependencies of slo-transcoder. Maybe we can get away without doing that?

PR #2243, opened 2019-05-17
validate unique instance names per cluster per service

stug:


paasta_tools/cli/cmds/validate.py
```diff
@@ -93,6 +95,19 @@ def valid_tron_namespace(cluster, filename):
     return success(f'{filename} is valid.')
 
 
+def duplicate_instance_names_message(service, cluster, instance_names):
+    instance_name_list = "\n\t".join(instance_names)
+    message = (
+        f"Service {service} has duplicate instance names for cluster {cluster}: "
```
qui: Maybe something like `uses the following instance names for more than one config in cluster {cluster}` would make the problem clearer?

stug: Yeah that's clearer.  I think I have a failing test, so I'll change it.

PR #2245, opened 2019-05-20
Add instance type as env variable

qui:
For spark-run we need to manually override the value, because it loads other instance configs (e.g. adhoc, marathon, etc).


solarkennedy:
This will trigger a big bounce, so please pin and ask EU to unpin for us.

PR #2246, opened 2019-05-22
Try to make a better reviewboard title for the paasta rightsizer.

solarkennedy:
Before:

    automatically updating myservice/foo-mycluster.yaml for over-provisioned cpu

After:

    Rightsizing myservice.main in mycluster to make it not have over-provisioned cpu

### DIFF
--- a/paasta_tools/contrib/paasta_update_soa_memcpu.py
+++ b/paasta_tools/contrib/paasta_update_soa_memcpu.py
@@ -37,7 +37,7 @@ def parse_args(argv):
     )
     parser.add_argument(
         '-m', '--manual',
-        help='Do not automatically publish the review',
+        help='Do not publish a review',
         action='store_true',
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -266,6 +266,7 @@ def main(argv=None):
             service_param=_get_dashboard_qs_param('paasta_service', serv['service']),
             instance_param=_get_dashboard_qs_param('paasta_instance', serv['instance']),
         )
+        summary = f"Rightsizing {serv['service']}.{serv['instance']} in {serv['cluster']} to make it not have {provisioned_state}"
```
thebostik: provisioned_state is just `over` or `under` so i think you will still need to jam `provisioned cpu` on at the end of this string

PR #2247, opened 2019-05-23
set paasta_config_sha as annotation on crs

Rob-Johnson:
we currently set them as a label, rather than an annotation. however,
labelselectors used by deployments and other k8s objects  tend to be
immutable, and as such can't easily be modified to track changes to
paasta_config_sha. change to an annotation so we still have the metadata


Rob-Johnson:
closing - this isn't as simple as I anticipated

PR #2248, opened 2019-05-23
Add puppet service type option to paasta_dump_locally_running_services

kaisen:
The original paasta_dump_locally_running_services script only dumped information for marathon services. In order to capture information for locally running TCP services (like cassandra) I added the option to dump puppet service info.


solarkennedy:
You will forever be playing catchup with the state of paasta this way.
Maybe this is in the wrong repo? What if you made this part of nerve-tools, so that you could use *exactly* the same function that nerve did when dumping this output?

https://github.com/Yelp/nerve-tools/blob/master/src/nerve_tools/configure_nerve.py#L376

kaisen:
> You will forever be playing catchup with the state of paasta this way.
> Maybe this is in the wrong repo? What if you made this part of nerve-tools, so that you could use _exactly_ the same function that nerve did when dumping this output?
> 
> https://github.com/Yelp/nerve-tools/blob/master/src/nerve_tools/configure_nerve.py#L376

That would work and would be easier to make changes to, but would that go against https://github.com/Yelp/nerve-tools/issues/11?

solarkennedy:
meh?

bplotnick:
@solarkennedy how strongly do you feel about this? "forever playing catchup" kind of feels YAGNI

Also, we could easily modify nerve_tools to call some common function that this thing calls. And then we only need to "play catchup" in one place rather than in both nerve_tools and here.

solarkennedy:
That is a good idea (the common function).

But yea, as soon as you ship this, are you going to add support for k8s?

paasta_tools/dump_locally_running_services.py
```diff
@@ -13,25 +13,61 @@
...
+    )
+    return parser.parse_args(argv)
+
+
+def main(argv=None):
+    args = parse_args(argv)
+    service_type = args.service_type
+    soa_dir = args.soa_dir
+
+    if service_type == 'marathon':
```
solarkennedy: Don't you need to do both? Otherwise are you going to do two cron jobs?

solarkennedy:
@kaisen it has been a while. Do you need me to click the green button, are you ready to go?

kaisen:
@solarkennedy this is blocked on some other work right now. I do want to combine both TCP and HTTP services in the same output per your suggestion. I'll get to that this week and then this branch is ready to go.

kaisen:
@solarkennedy This is good to go. IMO if/when we add k8s support in this script (or make other changes), we can deal with the refactor then. Merging this now will unblock https://github.com/Yelp/nerve-tools/pull/56

PR #2249, opened 2019-05-24
Only support paasta tron jobs in the non-MASTER namespace, optional jobs key

solarkennedy:


solarkennedy:
This feels a bit tricky to me.

I think it is worth it to make it so the thing that loads jobs doesn't even try to read in the "tron" folder.
I think it is also "OK" to not support paasta jobs in the MASTER.yaml file, and just post the `jobs:` as is (but it will have to be a `[]`?)

paasta_tools/tron_tools.py
```diff
@@ -527,25 +527,27 @@ def load_tron_instance_config(
...
-    if not config:
-        config = service_configuration_lib._read_yaml_file(os.path.join(tronfig_folder, f"{service}.yaml"))
     if not config:
         raise NoConfigurationForServiceError('No Tron configuration found for service %s' % service)
     return config
 
 
+def extract_jobs_from_tron_yaml(config):
+    config = {key: value for key, value in config.items() if not key.startswith('_')}  # filter templates
+    if 'jobs' in config and config.get('jobs') is None:
```
keymone: i  would make this path more explicitly "deprecated":

    if 'jobs' in config:
      log.warn("jobs key in per-service tronfig is deprecated, please de-dent")
      return config.get('jobs', {})
    else:
      return config

paasta_tools/setup_tron_namespace.py
```diff
@@ -107,6 +108,22 @@ def main():
...
+    master_config = tron_tools.create_complete_master_config(
+        cluster=args.cluster,
+        soa_dir=args.soa_dir,
+    )
+    if args.dry_run:
+        log.info(f"Would update {MASTER_NAMESPACE} to:")
+        log.info(f"{master_config}")
+        updated.append(MASTER_NAMESPACE)
+    else:
+        if client.update_namespace(MASTER_NAMESPACE, master_config):
```
keymone: this is new code, not mentioned in PR title or description, did you mean to include it?

solarkennedy: I mentioned it in the title sorta.
The theory here is that we treat the master namespace as special case *only* in this script, and nowhere else. That reduces the complexity anywhere else.
I think it is OK for paasta not to support having jobs in the `MASTER` namespace anyway.

paasta_tools/tron_tools.py
solarkennedy: I would rather not. 
I see those deprecation warnings for those users who we can't control.

I control all of the configs at yelp, so after this change I'll remove the `jobs` key everywhere. No need for anyone else to see it. Plus, nobody would see this anyway, it ends up on the tron server.

keymone: you're right, users won't see this. i was thinking about somebody naming their job "jobs" - we  should somehow warn this will not work correctly.

solarkennedy: Haha. That is a funny edge case, but it will soon be not important.

This code is only for the transition period. I plan to:
* ship this
* de-indent all the things and remove the jobs word in soa-configs
* update this code to just assume everything in soa-configs are jobs

keymone: Ok, ship away!

paasta_tools/setup_tron_namespace.py
```diff
@@ -107,6 +108,22 @@ def main():
...
+            updated.append(MASTER_NAMESPACE)
+            log.debug(f'Updated {MASTER_NAMESPACE}')
+        else:
+            skipped.append(MASTER_NAMESPACE)
+            log.debug(f'Skipped {MASTER_NAMESPACE}')
+
```
qui: Should still catch and report errors that happen here

paasta_tools/tron_tools.py
```diff
@@ -654,26 +656,11 @@ def _get_tron_namespaces_from_tron_dir(cluster, soa_dir):
...
-    pass
-
-
 def get_tron_namespaces_for_cluster(cluster=None, soa_dir=DEFAULT_SOA_DIR):
     """Get all the namespaces that are configured in a particular Tron cluster."""
     if not cluster:
         cluster = load_tron_config().get_cluster_name()
-
-    namespaces1 = set(_get_tron_namespaces_from_service_dir(cluster, soa_dir))
-    namespaces2 = set(_get_tron_namespaces_from_tron_dir(cluster, soa_dir))
```
qui: Want to delete this function?

tests/test_tron_tools.py
```diff
@@ -988,54 +977,23 @@ def test_validate_complete_config_passes(
...
     def test_get_tron_namespaces_for_cluster(self, mock_ls, mock_walk):
         cluster_name = 'stage'
-        expected_namespaces = ['app', 'foo', 'cool']
+        expected_namespaces = ['app', 'foo']
         mock_walk.return_value = [
             ('/my_soa_dir/foo', [], ['tron-stage.yaml']),
             ('/my_soa_dir/app', [], ['tron-stage.yaml']),
             ('my_soa_dir/woo', [], ['something-else.yaml']),
         ]
-        mock_ls.return_value = ['cool.yaml']
```
qui: I think you can take out the mock for ls now

qui:
If you're worried about the change, should be enough to test by running `setup_tron_namespaces -a --dry-run`

paasta_tools/cli/schemas/tron_schema.json
```diff
@@ -161,7 +161,6 @@
       }
     }
   },
-  "required": ["jobs"],
```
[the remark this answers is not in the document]
solarkennedy: will do. I'll try to make this transition happen quickly

paasta_tools/setup_tron_namespace.py
solarkennedy: What if we didn't?

There is a kind of "race" between updating the master namespace (setting pools, etc), where it really has to happen first, before we try to update other stuff. 

What if we made updating the master configs a failing condition, and we "fail fast" here instead of moving on to update other stuff?

solarkennedy:
This PR has gotten pretty big.

If you want, I can do differerent change sets:
1. Remove support for jobs in the MASTER namespace
2. Remove support for jobs in non-service dirs
3. Add support for the jobs as top-level keys

Would you like me to break it up?

paasta_tools/setup_tron_namespace.py
qui: Oh, it wasn't clear to me that `client.update_namespace` would raise an exception if the request was unsuccessful, but I see that it does. Failing fast sounds good to me.

chlgit:
Looking at the code again, I think it's a good idea to break it up. If there's any step going wrong, it would be easier to roll back.

PR #2252, opened 2019-05-28
Updated deployd logging a bit

solarkennedy:
I'm trying my best to really understand how deployd works.

My queue renaming helps, but there is still the `to_bounce` queue, which isn't really the bounce_now or bounce_later queues.

I have a hunch that we might have a bottleneck on the single inbox worker. I also think we could get rid of the bounce_now queue entirely and just have a bounce_now and bounce_later queue.

For now this is just more logging and stuff.

### DIFF
--- a/paasta_tools/deployd/workers.py
+++ b/paasta_tools/deployd/workers.py
@@ -1,5 +1,8 @@
+import datetime
 import time
 from collections import namedtuple
+
+import humanize
 
 from paasta_tools import marathon_tools
 from paasta_tools.deployd.common import BounceTimers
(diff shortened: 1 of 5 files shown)


paasta_tools/deployd/workers.py
```diff
@@ -68,14 +71,18 @@ def setup_timers(self, service_instance):
...
         while True:
             service_instance = self.instances_to_bounce_now.get()
             try:
                 bounce_again_in_seconds, return_code, bounce_timers = self.process_service_instance(service_instance)
             except Exception as e:
-                self.log.error("Worker failed to process service instance and will retry. "
-                               "Caused by exception: {}".format(e))
+                self.log.error(
+                    f"{self.name} Worker failed to process service instance and will retry. "
+                    f"Caused by exception: {format(e)}",
```
mattmb: is `format(e)` what you want? I don't actually know what it does...

solarkennedy: It is. Otherwise we get pages and pages of stacktrace

PR #2253, opened 2019-05-29
Make SLO failures trigger rollbacks automatically.

EvanKrall:


paasta_tools/automatic_rollbacks/slack.py
```diff
@@ -59,29 +59,37 @@ def get_slack_blocks_for_deployment(
...
         else:
-            blocks.append({
-                "type": "section",
-                "text": {
-                    "type": "mrkdwn",
-                    "text": f":ok_hand: All {len(slo_watchers)} SLOs are currently passing.",
-                },
-            })
+
+            num_unknown = len([w for w in slo_watchers if w.bad_before_mark is None or w.bad_after_mark is None])
```
solarkennedy: This block of code is complex enough to justify its own function.
It would be a little easier to test I think if that was the case

paasta_tools/automatic_rollbacks/state_machine.py
```diff
@@ -96,20 +106,51 @@ def start_timer(self, timeout, trigger, message_verb):
...
+            self.timer_running = False
 
         def schedule_callback():
             """Unfortunately, call_at is not threadsafe, and there's no call_at_threadsafe, so we need to schedule the
             call to call_at with call_soon_threadsafe."""
             self.timer_handle = self.event_loop.call_later(timeout, times_up)
+            self.timer_trigger = trigger  # This allows cancel_timer to selectively cancel.
+            self.timer_timeout = timeout  # saved for restart_timer
+            self.timer_message_verb = message_verb  # saved for restart_timer
+            # self.timer_started()
```
solarkennedy: Something else actually starts the timer on the method, right?

solarkennedy:
I'm worried that the complexity of this code is a bit beyond what I can keep in my head, due to the combinations of callbacks, more states, and class inheritance. I'm not really sure what to do about that though, I think this is correctly encapsulating the complexity of the problem?

paasta_tools/automatic_rollbacks/state_machine.py
EvanKrall: oh, oops, `timer_started` was meant to deal with an issue that I papered over with the `timer_running` boolean. I'll clean this up.

PR #2254, opened 2019-05-29
Pass POD_NAME to kubernetes pods

jedirandy:
- As it's needed by operator-sdk for some functionalities like leader election, see https://github.com/operator-framework/operator-sdk/blob/a9ee17ecae3d1de4e1231ed46fb4d6bc69769054/pkg/k8sutil/k8sutil.go#L94

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -473,6 +473,16 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
                     ),
                 ),
             ),
+            V1EnvVar(
+                # this is used by some functions of operator-sdk
+                # it uses this environment variable to get the pods
+                name='POD_NAME',
+                value_from=V1EnvVarSource(
+                    field_ref=V1ObjectFieldSelector(
+                        field_path='metadata.name',
+                    ),
+                ),
+            ),
         ]
         return kubernetes_env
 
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -473,6 +473,14 @@ def get_kubernetes_environment(self) -> Sequence[V1EnvVar]:
                     ),
                 ),
             ),
+            V1EnvVar(
+                name='POD_NAME',
```
ronin13: Would be good to add a comment on why this is required:  operator specification requires POD_NAME to be passed during deployment. etc., so that environment variable is not renamed in future (which will break lot of things).

tests/test_kubernetes_tools.py
```diff
@@ -379,6 +379,7 @@ def test_get_container_env(self):
     def test_get_kubernetes_environment(self):
         ret = self.deployment.get_kubernetes_environment()
         assert 'PAASTA_POD_IP' in [env.name for env in ret]
+        assert 'POD_NAME' in [env.name for env in ret]
```
ronin13: Is there an integration test of sorts as well for this?

paasta_tools/kubernetes_tools.py
jedirandy: right, I'll comment on that.

tests/test_kubernetes_tools.py
mattmb: the paasta itests are pretty slim for kubernetes. we do start a cluster and run a service I think so to some extent this is covered. We don't block on travis on it but it did work for your PR: https://travis-ci.org/Yelp/paasta/jobs/538753020

basically, don't worry about itests for now :-)

PR #2255, opened 2019-05-29
Fix deployd bounce_length timer

mattmb:
We currently include bounces where we just run setup_marathon_job and
find that the state is already good. In our dashboards we mostly exclude
this by just excluding stuff less than 60s which is the default sleep
between reprocessing. That's not always the case though if the queues
are busy or we ever change that sleep. This should ditch the metric for
things that aren't processed more than once i.e. they aren't really
bounced.

### DIFF
--- a/paasta_tools/deployd/common.py
+++ b/paasta_tools/deployd/common.py
@@ -33,6 +33,7 @@ BaseServiceInstance = namedtuple(
         'bounce_timers',
         'failures',
         'priority',
+        'processed_count',
     ],
 )
 
@@ -50,6 +51,7 @@ class ServiceInstance(BaseServiceInstance):
         failures: int = 0,
         bounce_timers: Optional[BounceTimers] = None,
         priority: Optional[int] = None,
+        processed_count: int = 0,
     ) -> 'ServiceInstance':
         if priority is None:
             priority = get_priority(service, instance, cluster)
@@ -62,6 +64,7 @@ class ServiceInstance(BaseServiceInstance):
             failures=failures,
             bounce_timers=bounce_timers,
             priority=priority,
+            processed_count=processed_count,
         )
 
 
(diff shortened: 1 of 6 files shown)


paasta_tools/deployd/common.py
```diff
@@ -50,6 +51,7 @@ def __new__(
         failures: int = 0,
         bounce_timers: Optional[BounceTimers] = None,
         priority: Optional[int] = None,
+        reprocessed_count: int = 0,
```
solarkennedy: What if this was called "processed_count" and started at 0 still, but was incremented on workers.py:115 instead? (where it actually gets processed I think)

paasta_tools/deployd/workers.py
```diff
@@ -129,9 +130,10 @@ def process_service_instance(self, service_instance):
                               bounce_again_in_seconds,
                           ))
         else:
-            bounce_timers.bounce_length.stop()
             self.log.info("{}.{} in steady state".format(
                 service_instance.service,
                 service_instance.instance,
             ))
+        if service_instance.reprocessed_count > 0:
+            bounce_timers.bounce_length.stop()
```
solarkennedy: I don't quite get this. 

Maybe we should be measuring the time to run s_m_j *and* the total bounce time?

mattmb: We do measure the time for both already. The only thing I'm trying to change here is not measuring total bounce time for service instances that are only processed once by a worker to check that they are in the correct state. I want the averages of total bounce time to only include things that are really being bounced (not just being checked by deployd).

mattmb: Oh I do have an indentation error though I think...

paasta_tools/deployd/common.py
mattmb: I think it's okay to modify where it is now because the `process_service_instance` doesn't actually return a `service_instance`.

But I do agree that `processed_count` is clearer let me update

PR #2257, opened 2019-05-29
Fix spark-run history-server class path

huadongliu:
make test
.tox/py36-linux/bin/paasta spark-run --spark-args 'spark.history.fs.logDirectory=s3a://TEST_PATH ' --cmd history-server


huadongliu:
Yep, this fixes the error we saw earlier.

PR #2259, opened 2019-05-30
Actually validate the instance configs when we run paasta validate

solarkennedy:
This is the new #1 mistake we make at yelp, now `paasta validate` will actually run the `validate` functions and validate things.

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -30,6 +30,7 @@ from paasta_tools.chronos_tools import load_chronos_job_config
 from paasta_tools.chronos_tools import TMP_JOB_IDENTIFIER
 from paasta_tools.cli.utils import failure
 from paasta_tools.cli.utils import get_file_contents
+from paasta_tools.cli.utils import get_instance_config
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import PaastaColors
 from paasta_tools.cli.utils import success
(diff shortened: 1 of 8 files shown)


paasta_tools/utils.py
```diff
@@ -715,7 +715,7 @@ def check_deploy_group(self) -> Tuple[bool, str]:
         if deploy_group is not None:
             pipeline_deploy_groups = get_pipeline_deploy_groups(service=self.service, soa_dir=self.soa_dir)
             if deploy_group not in pipeline_deploy_groups:
-                return False, f'deploy_group {deploy_group} is not in service {self.service} deploy.yaml'
+                return False, f'{self.service}.{self.instance} uses deploy_group {deploy_group}, but it is not deploy.yaml'  # noqa: E501
         return True, ''
```
kawaiwanyelp: Does this need to change so that we always have a deploy_group? (Thinking about farrell's question)

paasta_tools/cli/cmds/validate.py
```diff
@@ -308,35 +309,61 @@ def validate_chronos(service_path):
...
-        services_in_cluster = get_services_for_cluster(cluster=cluster, instance_type='chronos', soa_dir=soa_dir)
+        services_in_cluster = get_services_for_cluster(cluster=cluster, instance_type=instance_type, soa_dir=soa_dir)
         valid_services = {f"{name}{chronos_spacer}{instance}" for name, instance in services_in_cluster}
         for instance in list_all_instances_for_service(
                 service=service, clusters=[cluster], instance_type=instance_type,
                 soa_dir=soa_dir,
         ):
             cjc = load_chronos_job_config(service, instance, cluster, False, soa_dir)
             parents = cjc.get_parents() or []
-            checks_passed, check_msgs = cjc.validate()
```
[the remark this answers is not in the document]
solarkennedy: Hmm. I've renamed this function to `validate_chronos_dependencies`, because that kind of validation can only happen at the service level, and can't really happen on a validate() method on a single job.

paasta_tools/utils.py
solarkennedy: I could have a special case for when the deploy group is `""`

PR #2260, opened 2019-05-30
Begin logging service autoscaler state in a json format

thebostik:
This will provide us with detailed information about steady state operation of the autoscaler, most common failure modes and provide the data we'll use to set thresholds for automatic alerting about dangerous utilization before moving ahead with that.

I considered writing to clog directly here but it's inside a loop for all the master's autoscaling decisions, so stdout getting piped over seems preferable almost? Downside is that a service's (not the central) log will be harder to read without grepping for `event`. Open to ideas here.

Make test green. Make itest green.

This is tracked by PAASTA-15443.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -598,25 +598,41 @@ def autoscale_marathon_instance(
                 num_healthy_instances=num_healthy_instances,
             )
             safe_downscaling_threshold = int(current_instances * 0.7)
-            if new_instance_count > current_instances or \
-               new_instance_count < current_instances and not task_data_insufficient:
-                set_instances_for_marathon_service(
-                    service=marathon_service_config.service,
-                    instance=marathon_service_config.instance,
-                    instance_count=new_instance_count,
-                )
             _record_autoscaling_decision(
-                marathon_service_config,
-                autoscaling_params,
-                utilization,
-                log_utilization_data,
-                error,
-                current_instances,
-                num_healthy_instances,
-                new_instance_count,
-                task_data_insufficient,
-                safe_downscaling_threshold,
+                marathon_service_config=marathon_service_config,
+                autoscaling_params=autoscaling_params,
+                utilization=utilization,
+                log_utilization_data=log_utilization_data,
+                error=error,
+                current_instances=current_instances,
+                num_healthy_instances=num_healthy_instances,
+                new_instance_count=new_instance_count,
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -585,70 +588,123 @@ def autoscale_marathon_instance(
...
+                current_instances=current_instances,
+                num_healthy_instances=num_healthy_instances,
+                new_instance_count=new_instance_count,
+                task_data_insufficient=task_data_insufficient,
+                safe_downscaling_threshold=safe_downscaling_threshold,
+            ),
+        ),
+        level='debug',
+    )
+    if new_instance_count != current_instances:
```
EvanKrall: I think duplicating this logic here and on line 601 is going to lead to these getting out of sync, which will mean log messages that say we do something different than we actually do.

Can we put these log lines back next to the actions they're describing, but keep the JSON log and the meteorite stuff?

thebostik: I hear what you're saying but it's super long so I pulled it over wholesale thinking it would be nice to keep all of the logging side effects together. Agree there is duplication on 601/602, I could try to drop that into a helper.

The way I was thinking of it is that the decision was already made in the helpers autoscale_marathon_instance calls, this is just grouping together some basic logic about the numbers to interpret what happened for a human (if people do read these?).

I actually feel a little bad about dumping json into what will also end up in the service's `paasta logs -C deploy`, not sure if you had a better pattern I could refer to for going straight to clog?

thebostik: I guess I'm wondering who those debug level messages were actually for? Now that we're printing the entire json, how does it sound to just drop them (new diff)?

Anyway it's a kind of transitionary state, I think that logging helper will still possibly run / initiate more math happening on the numbers to find, e.g., critical alerting levels. Not sure how you feel about this approach over vs an SLO oriented one, but wanted to get this out so we can talk more about it with some code to refer to.

EvanKrall: Yeah, if we think the new JSON line is at least as informative as the current debug messages (i.e. if we can deduce which code path would have been taken), I'm not tied to them. I like easily parsed logs.

In that interest, maybe add `task_data_insufficient`?

thebostik: face palm so hard, I actually removed the variable that initiated doing this whole thing.

Ok, new diff uploaded and fixed a type issue.

PR #2263, opened 2019-06-05
Spark set intelligent defaults shuffle partitions

talalryz:
set intelligent defaults for spark.sql.shuffle.partitions.

Spark sets a default value of 200. 
We set it to default to 2*spark.cores.max, and print a warning message that ideally the user
should be setting this themselves.

Also, added tests that verify that we default to twice the number of max cores only when the 
user does not provide a value for spark.sql.shuffle.partitions.

make test green.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -483,16 +483,16 @@ def get_spark_config(
             user_args[fields[0]] = fields[1]
 
     if 'spark.sql.shuffle.partitions' not in user_args:
-        user_args['spark.sql.shuffle.partitions'] = str(
+        num_partitions = str(
             2 * int(user_args['spark.cores.max']),
         )
+        user_args['spark.sql.shuffle.partitions'] = num_partitions
         paasta_print(
             PaastaColors.yellow(
-                'Warning: spark.sql.shuffle.partitions has been set to {num_partitions} '
-                'to be equal to twice the number of requested cores, but you should '
-                'consider setting a higher value if necessary.'.format(
-                    num_partitions=user_args['spark.sql.shuffle.partitions'],
-                ),
+                f'Warning: spark.sql.shuffle.partitions has been set to'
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -482,6 +482,20 @@ def get_spark_config(
             # Update default configuration
             user_args[fields[0]] = fields[1]
 
+    if 'spark.sql.shuffle.partitions' not in user_args:
+        user_args['spark.sql.shuffle.partitions'] = str(
+            2 * int(user_args['spark.cores.max']),
+        )
+        paasta_print(
```
solarkennedy: Why do we need to print a warning?
Why can't this just be a "sane default" for users?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -482,6 +482,20 @@ def get_spark_config(
...
+            PaastaColors.yellow(
+                'Warning: spark.sql.shuffle.partitions has been set to {num_partitions} '
+                'to be equal to twice the number of requested cores, but you should '
+                'consider setting a higher value if necessary.'.format(
```
solarkennedy: please just use fstrings

paasta_tools/cli/cmds/spark_run.py
talalryz: 1. I believe there is value in warning that we are setting a default for this parameter that is not the spark default. 
2. Too many partitions can cause a sizeable overhead e.g., a job with 800 cores can emit 1600 partitions!! The default value that we're setting is a better guess than what spark provides but may still be bad. 
3. There is quite a bit of information setting the right value of spark.sql.shuffle.partitions, depending on the amount of data as well; something that we're not doing.

paasta_tools/cli/cmds/spark_run.py
talalryz: switched to using fstrings.

PR #2266, opened 2019-06-05
Add cleanup_kubernetes_cr script

vkhromov:


paasta_tools/cleanup_kubernetes_cr.py
```diff
@@ -0,0 +1,169 @@
...
+from paasta_tools.kubernetes_tools import KubeKind
+from paasta_tools.kubernetes_tools import list_custom_resources
+from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.utils import SystemPaastaConfig
+
+log = logging.getLogger(__name__)
+
+
+class CustomResource(NamedTuple):
```
Rob-Johnson: isn't this already somewhere else, maybe `kubernetes_tools.py`?

paasta_tools/cleanup_kubernetes_cr.py
```diff
@@ -0,0 +1,169 @@
...
+    file_prefix: str
+    version: str
+    kube_kind: KubeKind
+    group: str
+
+
+def load_custom_resources(system_paasta_config: SystemPaastaConfig) -> Sequence[CustomResource]:
```
Rob-Johnson: same - this should probably be in a place it can be reused

vkhromov: done

paasta_tools/cleanup_kubernetes_cr.py
vkhromov: done

paasta_tools/cleanup_kubernetes_cr.py
```diff
@@ -0,0 +1,142 @@
...
+    soa_dir: str,
+    cluster: str,
+    custom_resources: Sequence[CustomResource],
+) -> bool:
+    cluster_crds = {
+        crd.spec.names.kind
+        for crd in
+        kube_client.apiextensions.list_custom_resource_definition(
+            label_selector="yelp.com/paasta_service",
+        ).items
```
Rob-Johnson: doesn't this need to be `items()` ?

paasta_tools/kubernetes_tools.py
```diff
@@ -1445,3 +1475,13 @@ def sanitise_service_name(
     service: str,
 ) -> str:
     return service.replace('_', '--')
+
+
+def load_custom_resources(
```
Rob-Johnson: this is actually `load_custom_resource_definitions` isn't it?

paasta_tools/kubernetes_tools.py
```diff
@@ -146,6 +149,13 @@ class KubeService(NamedTuple):
     registrations: Sequence[str]
 
 
+class CustomResource(NamedTuple):
```
Rob-Johnson: is this actually `CustomResourceDefinition`?

paasta_tools/cleanup_kubernetes_cr.py
```diff
@@ -0,0 +1,142 @@
...
+    }
+    log.debug(f"CRDs found: {cluster_crds}")
+    results = []
+    for custom_resource in custom_resources:
```
Rob-Johnson: this is kind of confusing - aren't we actually iterating over resource definitions here?

paasta_tools/cleanup_kubernetes_cr.py
```diff
@@ -0,0 +1,142 @@
...
+        if custom_resource.kube_kind.singular not in cluster_crds:
+            # TODO: kube_kind.singular seems to correspond to `crd.names.kind`
+            # and not `crd.names.singular`
+            log.warning(
+                f"CRD {custom_resource.kube_kind.singular} "
+                f"not found in {cluster}",
+            )
+            continue
+        config_dicts = load_all_configs(
```
Rob-Johnson: i'm surprised that we've added another way of doing this.

can we just use the existing `get_service_instance_list` from `paasta_tools.utils`? I think it's slightly different in that it only returns the instance list, rather than the body of the yaml, but I think that's ok - we only seem to use that anyway?

I think you'll have to add `CassandraCluster` to `paasta_tools.utils.INSTANCE_TYPES` to use it though

mattmb: I think this was a deliberate choice so that it wasn't coupled to paasta knowing about all the different instance types...

I'm not sure what's better to be honest. But imagine you have an operator and don't want anything to do with paasta but just want something to sync CRs.

Rob-Johnson: but this thing only queries custom resources with the `yelp.com/paasta_service` label as a qualifier, so presumably they're being looked after by paasta anyway in that case?

yeah, it's kind of odd to me in a place where we're only doing half the job - I can't imagine that case.

paasta_tools/cleanup_kubernetes_cr.py
vkhromov: it is `items` since we need that to work both with a list and with a mock:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

kube_client = <Mock id='139930041401472'>, soa_dir = '/nail/soa', cluster = <Mock id='139930041401640'>
custom_resource_definitions = [<Mock id='139930035160736'>, <Mock id='139930035162304'>], service = None, instance = None

    def setup_all_custom_resources(
        kube_client: KubeClient,
        soa_dir: str,
        cluster: str,
        custom_resource_definitions: Sequence[CustomResourceDefinition],
        service: str = None,
        instance: str = None,
    ) -> bool:
        cluster_crds = {
            crd.spec.names.kind
            for crd in
            kube_client.apiextensions.list_custom_resource_definition(
>               label_selector="yelp.com/paasta_service",
            )
        }
E       TypeError: 'Mock' object is not iterable

paasta_tools/setup_kubernetes_cr.py:155: TypeError
```

paasta_tools/kubernetes_tools.py
vkhromov: done

paasta_tools/kubernetes_tools.py
vkhromov: done

paasta_tools/cleanup_kubernetes_cr.py
vkhromov: done

paasta_tools/cleanup_kubernetes_cr.py
vkhromov: Can we do that as a part of another CR please?
This one is already becomes too big for a just additional of simple script.

Rob-Johnson: yeah I guess so

paasta_tools/cleanup_kubernetes_cr.py
Rob-Johnson: ohhh, sorry, `items` is actually a property on that thing, not just the regular `items()` func on a dict. ignore me.

PR #2268, opened 2019-06-05
Add orphan checker to find orphan registrations

bplotnick:
This adds an orphan checker to find orphan znode registrations. It works by 1) scraping the local zk for registrations and 2) pulling nerve configs via xinetd and 3) comparing the two

Some shortcomings here:
* There are transient false-positives due to race conditions. For example, if a service goes down between the time that we scrape zk and get the nerve config, then we'll get a false positive. We can probably handle this by having some history
* There are some services that I've blacklisted for various reasons, mainly k8s stuff. I checked in the blacklist here, but I will probably just stick this in puppet instead. Orphans in the blacklist will not be detected (though I don't know if this is an issue for k8s)


paasta_tools/contrib/check_orphans.py
```diff
@@ -0,0 +1,213 @@
...
+def get_instance_data(service_blacklist_path: str) -> Tuple[Set[InstanceTuple], Set[InstanceTuple]]:
+    blacklisted_services = set(yaml.load(service_blacklist_path))
+    # Dump ZK
+    zk_data = get_zk_data(blacklisted_services)
+    zk_instance_data = read_from_zk_data(zk_data)
+
+    hosts = {x[0] for x in zk_instance_data}
+
+    # Dump Nerve configs from each host via xinetd
+    coro = gather_files(hosts)
```
solarkennedy: What is coro?

paasta_tools/contrib/check_orphans.py
```diff
@@ -0,0 +1,213 @@
...
+    collisions: List[str] = []
+    for zk_inst in zk_instance_data:
+        nerve_services = instance_by_addr[(zk_inst.host, zk_inst.port)]
+        if len(nerve_services) >= 1 and zk_inst.service not in nerve_services:
+            collisions.append(
+                f'[{zk_inst.host}:{zk_inst.port}] {zk_inst.service} collides with {nerve_services}',
+            )
+
+    if collisions:
+        logging.warning('Collisions found! Traffic is being misrouted!')
```
solarkennedy: Gotta set the return value, to 2 here.
Then maybe filter down the return value and do `sys.exit(main()` or something like that

paasta_tools/contrib/check_orphans.py
```diff
@@ -0,0 +1,213 @@
...
+    for orphan in orphans:
+        orphans_by_host[orphan.host].append((orphan.port, orphan.service))
+    with open(os.path.join(output_dir, 'orphans_by_host.yaml'), 'w') as f:
+        yaml.dump(dict(orphans_by_host), f)
+
+    # groupby service
+    orphans_by_service: DefaultDict[str, List[Tuple[str, int]]] = defaultdict(list)
+    for orphan in orphans:
+        orphans_by_service[orphan.service].append((orphan.host, orphan.port))
+    with open(os.path.join(output_dir, 'orphans_by_service.yaml'), 'w') as f:
```
solarkennedy: why do we need a yaml file?, here and line 187?

paasta_tools/contrib/check_orphans.py
```diff
@@ -0,0 +1,213 @@
...
+
+    collisions: List[str] = []
+    for zk_inst in zk_instance_data:
+        nerve_services = instance_by_addr[(zk_inst.host, zk_inst.port)]
+        if len(nerve_services) >= 1 and zk_inst.service not in nerve_services:
+            collisions.append(
+                f'[{zk_inst.host}:{zk_inst.port}] {zk_inst.service} collides with {nerve_services}',
+            )
+
+    if collisions:
```
solarkennedy: Although normally no news is good news, if there are no collisions can you assure the user?
```
print(f"OK: No collisions found in the {len(services)} announced by nerve")
```

paasta_tools/contrib/check_orphans.py
```diff
@@ -0,0 +1,213 @@
...
+    logging.basicConfig(level=logging.INFO)
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--service-blacklist-path',
+        default=DEFAULT_SERVICE_BLACKLIST_PATH,
+        help="Path to a file that contains services to ignore",
+    )
+    parser.add_argument(
+        '--output-dir',
+        default=DEFAULT_OUTPUT_DIR,
```
solarkennedy: I don't think we really need this?

paasta_tools/contrib/check_orphans.py
```diff
@@ -0,0 +1,213 @@
...
+        f.write('\n'.join(collisions))
+
+    logging.info('Wrote out orphan data')
+
+
+def main():
+    logging.basicConfig(level=logging.INFO)
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        '--service-blacklist-path',
```
solarkennedy: How many are we talking about here? Does it have to be a yaml? can it just be a csv cli param?

paasta_tools/contrib/check_orphans.py
bplotnick: This was originally for my own benefit, but I think is good for whatever checker script to output the orphans. To address your previous comment as well, I wrote this so that it isn't meant to act as a check itself, rather it would be used by a check.

solarkennedy: Why though?

Even still, you could just run the command and `>` the output and make the check look for a string in the output?

But I say just run the thing directly.

paasta_tools/contrib/check_orphans.py
bplotnick: I was seeing around 30 at the very least. But actually in this iteration of this script, we might not need the blacklist, but it will incur a 1 second connection timeout penalty for each instance when the script tries to get the nerve file

paasta_tools/contrib/check_orphans.py
bplotnick: `coroutine`. it can be combined with the next line, which i'll do now

paasta_tools/contrib/check_orphans.py
bplotnick: I strongly dislike doing string matching on stdout as a signal. I tend not to prefer the unix-y exit-code-as-a-signal approach, but I think I feel much less strongly about it as you do, so i'll go ahead and change it to have the exit code signal the existence of the orphans

solarkennedy: Thank you

paasta_tools/contrib/check_orphans.py
bplotnick: We want the check to fail when there are orphans (timebombs), not just collisions (detonated timebombs). How about I return `3` (to avoid [reserved](https://www.tldp.org/LDP/abs/html/exitcodes.html#EXITCODESREF) exit codes) for collisions and 4 for orphans?

solarkennedy: Heh, I wouldn't worry about the reserved return codes and instead follow the standard nagios check api
https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/pluginapi.html

0 (ok) for nothing
1 (warning) for timebombs
2 (critical) for collisions

bplotnick:
@solarkennedy I've switched it to a check. PTAL

paasta_tools/contrib/check_orphans.py
```diff
@@ -0,0 +1,211 @@
...
+    )
+    args = parser.parse_args()
+
+    zk_instance_data, nerve_instance_data = get_instance_data(args.service_blacklist_path)
+
+    return check_orphans(zk_instance_data, nerve_instance_data)
+
+
+if __name__ == '__main__':
+    sys.exit(main().value)
```
solarkennedy: what is up with the `.value` here? It isn't already just a number?

bplotnick: nope. it's an `Enum`. https://docs.python.org/3/library/enum.html#programmatic-access-to-enumeration-members-and-their-attributes

PR #2269, opened 2019-06-05
emit metrics when deployd starts

drmorr0:


drmorr0:
The travis mypy check looks like just a flake, merging.

PR #2270, opened 2019-06-06
Prevent deployd from blocking on stderr in itests

EvanKrall:
I think this is the reason itests are failing so persistently in Jenkins.

### DIFF
--- a/paasta_itests/steps/paasta_deployd_steps.py
+++ b/paasta_itests/steps/paasta_deployd_steps.py
@@ -2,6 +2,7 @@ import errno
 import fcntl
 import json
 import os
+import threading
 import time
 from subprocess import PIPE
 from subprocess import Popen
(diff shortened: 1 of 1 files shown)


EvanKrall:
Huh, I didn't notice that. Maybe that has a similar effect?

from [man(7) pipe](http://man7.org/linux/man-pages/man7/pipe.7.html):
```
 Pipe capacity
       A pipe has a limited capacity.  If the pipe is full, then a write(2)
       will block or fail, depending on whether the O_NONBLOCK flag is set
       (see below).  Different implementations have different limits for the
       pipe capacity.  Applications should not rely on a particular
       capacity: an application should be designed so that a reading process
       consumes data as soon as it is available, so that a writing process
       does not remain blocked.
```

Seems like that would make writes to stderr (logging) throw an exception instead of block?

PR #2271, opened 2019-06-07
Only support configuring tron jobs from service directories

solarkennedy:
This isn't "that" much simpler than the other PR I guess.

But fundamentally this "simplifies" the code in two ways:
* we remove code that existed to support reading namespaces in non-service dirs
* We remove the "other_config" concept, and explicitly load the MASTER namespace in its own way, instead of treating MASTER namespaces and service namespaces like they are the same.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -511,7 +511,7 @@ def load_tron_instance_config(
     load_deployments: bool = True,
     soa_dir: str = DEFAULT_SOA_DIR,
 ) -> TronActionConfig:
-    jobs, _ = load_tron_service_config(
+    jobs = load_tron_service_config(
         service=service,
         cluster=cluster,
         load_deployments=load_deployments,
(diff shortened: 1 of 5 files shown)


paasta_tools/tron_tools.py
```diff
@@ -556,68 +552,68 @@ def load_tron_service_config(service, cluster, load_deployments=True, soa_dir=DE
...
-
-    if service == MASTER_NAMESPACE:
-        other_config = format_master_config(
-            other_config,
-            system_paasta_config.get_volumes(),
-            system_paasta_config.get_dockercfg_location(),
-        )
-
-    other_config['jobs'] = {
+    preproccessed_config = {}
```
chlgit: nit: don't want to combine this with next line?

paasta_tools/tron_tools.py
```diff
@@ -556,68 +552,68 @@ def load_tron_service_config(service, cluster, load_deployments=True, soa_dir=DE
...
-
     # PaaSTA-specific validation
     for job_config in job_configs:
         check_msgs = job_config.validate()
         if check_msgs:
             return check_msgs
 
+    master_config_path = os.path.join(os.path.abspath(soa_dir), 'tron', cluster, MASTER_NAMESPACE + '.yaml')
+
+    preproccessed_config = {}
```
chlgit: nit: same here. I guess it's more personal preference (shrug).

paasta_tools/tron_tools.py
solarkennedy: meh

PR #2273, opened 2019-06-07
Only expose rollback buttons if there is a sha to roll back to

solarkennedy:


solarkennedy:
Huh. I don't know.
I think the code is probably easier to think about when the state machine states are "static".

PR #2274, opened 2019-06-08
Removed the non-state machine code for mark-for-deployment

solarkennedy:
I'm sure I've got lots of dead code still in here.

### DIFF
--- a/paasta_tools/automatic_rollbacks/slack.py
+++ b/paasta_tools/automatic_rollbacks/slack.py
@@ -259,8 +259,11 @@ class SlackDeploymentProcess(DeploymentProcess, abc.ABC):
         return buttons
 
     def update_slack_thread(self, message, color=None):
-        print(f"Updating slack thread with: {message}")
-
+        if self.slack_client is None:
+            print(f"Would update the slack thread with: {message}")
+            return
+        else:
+            print(f"Updating slack thread with: {message}")
         if color:
             resp = self.slack_client.api_call(
                 'chat.postMessage',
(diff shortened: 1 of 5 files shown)


EvanKrall:
Did we decide we should respect `slack_notify: true` in deploy.yaml before shipping this? Do you want to do that on this branch?

I think we can ignore the `notify_after_mark`/`notify_after_good_deploy`/`notify_after_auto_rollback` flags for now -- the threading we do now makes the extra messages much less intrusive.

Looks like the behavior should be to exit after marking if we don't specify --wait-for-deployment: https://github.com/Yelp/paasta/blob/a556160095279df542411d00a057a5f9844ac759/paasta_tools/cli/cmds/mark_for_deployment.py#L888

Also I'd suggest testing this manually: I use `.tox/py36-linux/bin/paasta mark-for-deployment --service katamari_test_service -k 9eb185e5c60b97c8e49560b7c866dac7ba4bfc05 --auto-rollback --wait-for-deployment --deploy dev.everything --verbose` -- you'd drop the --auto-rollback flag and also try without --wait-for-deployment.

EvanKrall:
Also, maybe we should exit when the deploy finishes, and not wait 600 seconds if --auto-rollbacks is not specified? That would be a matter of doing something like https://github.com/Yelp/paasta/blob/a556160095279df542411d00a057a5f9844ac759/paasta_tools/cli/cmds/mark_for_deployment.py#L888 but for the `deployed` state and `args.auto_rollback` / `self.auto_rollback`

paasta_tools/automatic_rollbacks/slack.py
```diff
@@ -259,8 +259,11 @@ def get_available_buttons(self) -> List[str]:
         return buttons
 
     def update_slack_thread(self, message, color=None):
-        print(f"Updating slack thread with: {message}")
-
+        if self.slack_client is None:
+            print(message)
```
EvanKrall: perhaps keep the print format the same, or be explicit that this isn't going to slack:

```
print(f"Would update slack thread with: {message}")
```

PR #2275, opened 2019-06-10
check_flink_services_replication

vkhromov:


paasta_tools/check_kubernetes_services_replication.py
```diff
@@ -89,9 +66,9 @@ def check_healthy_kubernetes_tasks_for_service_instance(
     )
 
 
-def check_service_replication(
+def check_kubernetes_service_replication(
```
Rob-Johnson: I think using the word service at all is kind of confusing (just because of the clash with the kubernetes service concept vs yelp's). how about `check_kubernetes_pod_replication` ?

paasta_tools/check_services_replication_tools.py
```diff
@@ -0,0 +1,122 @@
...
+from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import InstanceConfig_T
+from paasta_tools.utils import list_services
+from paasta_tools.utils import load_system_paasta_config
+
+
+log = logging.getLogger(__name__)
+
+
+def parse_args() -> argparse.Namespace:
```
Rob-Johnson: can you add an arg for `service.instance` and then default to checking all if it isn't provided? that'll make this easier to do the `list_kubernetes_service_instances|shuf|xargs` trick in the future if this starts taking too long

paasta_tools/monitoring_tools.py
```diff
@@ -400,14 +401,21 @@ def check_smartstack_replication_for_instance(
 
 def send_replication_event_if_under_replication(
     instance_config,
-    expected_count,
-    num_available,
+    expected_count: int,
+    num_available: int,
+    subinstance_type: Optional[str] = None,
```
Rob-Johnson: maybe `sub_component` could be a better word?

paasta_tools/check_services_replication_tools.py
vkhromov: done

paasta_tools/check_kubernetes_services_replication.py
vkhromov: done

paasta_tools/monitoring_tools.py
vkhromov: done

PR #2278, opened 2019-06-11
Fix loading jobs that are empty

solarkennedy:
Before, my logic would return `jobs: {}` when extracting jobs, because `{}` is false in python.

### DIFF
--- a/tests/test_tron_tools.py
+++ b/tests/test_tron_tools.py
@@ -816,6 +816,21 @@ class TestTronTools:
         )
 
     @mock.patch('paasta_tools.tron_tools.load_tron_yaml', autospec=True)
+    def test_load_tron_service_config_empty(self, mock_load_tron_yaml):
+        mock_load_tron_yaml.return_value = {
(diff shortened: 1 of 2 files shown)


tests/test_tron_tools.py
```diff
@@ -1004,3 +1019,38 @@ def test_list_tron_clusters(self, mock_glob):
...
+
+def test_extract_jobs_from_tron_yaml_defaults_to_jobs_if_available():
+    config = {
+        "_template": "foo",
+        "job0": "bar",
+        "jobs": {
+            "job1": "baz",
+        },
+    }
+    assert tron_tools.extract_jobs_from_tron_yaml(config) == {"job1": "baz:"}
```
qui: typo at the end (extra `:`)

solarkennedy: whoa I swear the tests passed on my machine...
must be a vim artifact..... :)

solarkennedy:
I rebased and merged this on the command line

PR #2280, opened 2019-06-13
Coreml 838 set aws region spark run

talalryz:
make test green
make itest green

This review changes paasta spark-run to set the aws_region in the environment, in 
addition the the aws creds which are already set in the environment.
We want aws_region to be set in the environment because we are removing aws creds management from spark_etl, and many flows in spark_etl require an aws region to be specified.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -924,9 +924,6 @@ def paasta_spark_run(args):
             )
             return 1
 
-    paasta_print(PaastaColors.red('lolololol'))
-    paasta_print(PaastaColors.red(os.environ['AWS_DEFAULT_REGION']))
-
     return configure_and_run_docker_container(
         args,
         docker_img=docker_url,
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -33,6 +33,7 @@ from paasta_tools.utils import PaastaNotConfiguredError
 from paasta_tools.utils import SystemPaastaConfig
 
 AWS_CREDENTIALS_DIR = '/etc/boto_cfg/'
+DEFAULT_AWS_REGION = 'us-west-2'
 DEFAULT_SERVICE = 'spark'
 DEFAULT_SPARK_WORK_DIR = '/spark_driver'
 DEFAULT_SPARK_DOCKER_IMAGE_PREFIX = 'paasta-spark-run'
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -909,6 +924,9 @@ def paasta_spark_run(args):
             )
             return 1
 
+    paasta_print(PaastaColors.red('lolololol'))
+    paasta_print(PaastaColors.red(os.environ['AWS_DEFAULT_REGION']))
+
```
lydian: maybe make a more descriptive output?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -404,6 +418,7 @@ def load_aws_credentials_from_yaml(yaml_file_path):
         return (
             credentials_yaml['aws_access_key_id'],
             credentials_yaml['aws_secret_access_key'],
+            credentials_yaml['region'],
```
lydian: Most of the yaml doesn't have region, maybe use `.get('region')` to ensure it won't break?

tests/cli/test_cmds_spark_run.py
```diff
@@ -433,15 +434,18 @@ def test_get_docker_cmd_mrjob():
 def test_load_aws_credentials_from_yaml(tmpdir):
```
lydian: maybe have another test to make sure missing region in yaml won't break everything.

paasta_tools/cli/cmds/spark_run.py
talalryz: okay, this needs to be changed!

paasta_tools/cli/cmds/spark_run.py
huadongliu: Looks like order of precedence is:
1. --aws-region
2. region in yaml files (if exists)
3. Session.region_name

Whenever returning from get_aws_credentials_and_region(), we should have a valid region instead of None.

talalryz: > Most of the yaml doesn't have region, maybe use `.get('region')` to ensure it won't break?

ack!

> Looks like order of precedence is:
> 
> 1. --aws-region
> 2. region in yaml files (if exists)
> 3. Session.region_name
> 
> Whenever returning from get_aws_credentials_and_region(), we should have a valid region instead of None.

we could have a variable  DEFAULT AWS REGION at the top of the file, and set it to be 'us-west-2'. Then, we can ensure that get_aws_credentials_and_region() always returns a region.

tests/cli/test_cmds_spark_run.py
talalryz: ack !

paasta_tools/cli/cmds/spark_run.py
talalryz: @lydian @huadongliu I think it makes sense not to look for the aws region in the credentials files, since the region really shouldn't be there. 

We can just look if --aws-region is supplied and if not, we can default to 'us-west-2'? This also keeps our logic simple.

PR #2281, opened 2019-06-14
Incrementally turn on the new mark-for-deployment state machine by activating on wait-for-deployment

solarkennedy:
This is kinda like a 50% roll out


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -463,7 +463,7 @@ def paasta_mark_for_deployment(args):
 
     deploy_info = get_deploy_info(service=service, soa_dir=args.soa_dir)
 
-    if args.auto_rollback:
+    if args.auto_rollback or args.wait_for_deployment:
```
EvanKrall: it's actually `args.block` - see line 130

PR #2282, opened 2019-06-17
Add certificate issue/renewal logic

mattmb:
Issues a certificate via vault if https is enabled for the API endpoint.
If the certificate is expired then we renew it and retry the request.

### DIFF
--- a/paasta_tools/api/auth_decorator.py
+++ b/paasta_tools/api/auth_decorator.py
@@ -1,0 +1,143 @@
+import functools
+
+from bravado.docstring_property import docstring_property
+from bravado.exception import HTTPBadRequest
+
+import paasta_tools.api
+from paasta_tools.utils import load_system_paasta_config
+
+
+class AuthFutureDecorator:
+
+    def __init__(self, future, cluster_name):
+        self.future = future
+        self.attempts = 0
+        self.cluster_name = cluster_name
+
+    def result(self, timeout=None):
+        """ Wraps a Future's result method with a timer. """
+        try:
+            result = self.future.result(timeout)
+        except HTTPBadRequest as e:
+            # Yes nginx responds with a 400 if the certificate
+            # is expired. So the best we can do is watch for that
+            # return code and check the body of the response
+            if 'SSL' in e.response.text:
+                if self.attempts >= 3:
+                    raise
+                self.attempts += 1
+                system_paasta_config = load_system_paasta_config()
+                paasta_tools.api.client.renew_issue_cert(
+                    system_paasta_config=system_paasta_config,
+                    cluster=self.cluster_name,
+                )
+                result = self.result(timeout)
+            else:
+                raise
+        return result
+
+
+class AuthResourceDecorator:
+
+    def __init__(self, resource, cluster_name):
+        self.resource = resource
+        self.cluster_name = cluster_name
+
+    def __getattr__(self, name):
+        return decorate_client(self.resource, self._with_auth_check, name)
+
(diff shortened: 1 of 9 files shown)


paasta_tools/api/auth_decorator.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+class AuthFutureDecorator:
+
+    def __init__(self, future, cluster_name):
+        self.future = future
+        self.attempts = 0
+        self.cluster_name = cluster_name
+
+    def result(self, *args, **kwargs):
```
vkhromov: `def result(self, timeout=None)`?

paasta_tools/api/auth_decorator.py
```diff
@@ -0,0 +1,145 @@
...
+        """ Wraps a Future's result method with a timer. """
+        try:
+            result = self.future.result(*args, **kwargs)
```
vkhromov: `result = self.future.result(timeout)`?

paasta_tools/api/auth_decorator.py
```diff
@@ -0,0 +1,145 @@
...
+            if 'SSL' in e.response.text:
+                if self.attempts >= 3:
+                    raise
+                self.attempts += 1
+                system_paasta_config = load_system_paasta_config()
+                paasta_tools.api.client.renew_issue_cert(
+                    system_paasta_config=system_paasta_config,
+                    cluster=self.cluster_name,
+                )
+                result = self.result(*args, **kwargs)
```
vkhromov: It seems that on the first recursion call you'll try to get `result` from the same `self.future` object, right?
If that's true then it isn't clear why there will be a second HTTP request/response with the new cert/key and not just the same exception raised as the result of the previous future execution result.

paasta_tools/api/client.py
```diff
@@ -89,16 +91,41 @@ def __init__(
         system_paasta_config: SystemPaastaConfig,
     ) -> None:
         if scheme == 'https':
+            ca_cert_path = None
             if system_paasta_config.get_enable_client_cert_auth():
+                ecosystem = system_paasta_config.get_vault_cluster_config()[cluster]
                 paasta_dir = os.path.expanduser('~/.paasta/pki')
-                ssl_cert = (f'{paasta_dir}/{cluster}.crt', f'{paasta_dir}/{cluster}.key')
+                if not os.path.isfile(f'{paasta_dir}/{ecosystem}.crt'):
```
vkhromov: probably better to check the existence of all three files: key/crt/ca?

paasta_tools/api/client.py
```diff
@@ -89,16 +91,41 @@ def __init__(
         system_paasta_config: SystemPaastaConfig,
     ) -> None:
         if scheme == 'https':
+            ca_cert_path = None
             if system_paasta_config.get_enable_client_cert_auth():
+                ecosystem = system_paasta_config.get_vault_cluster_config()[cluster]
```
vkhromov: This seems to be assume that we use the same cert/key/ca for all paasta clusters in the ecosystem, right?

paasta_tools/api/auth_decorator.py
mattmb: Yup, we discussed in real life. I think this is just bad naming, it definitely fires another request with the new cert.

paasta_tools/api/client.py
mattmb: Yeah, I think this is ok to start with. I was going back and forth between cluster/ecosystem. But as this is just authn I think ecosystem will be fine and per-cluster permission can be at the authz layer.

PR #2284, opened 2019-06-17
Send all slack notifications for m-f-d to some channel.

solarkennedy:


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -631,13 +631,16 @@ def get_slack_client(self) -> SlackClient:
...
         Currently only uses the first slack channel available, and doesn't support
         multi-channel notifications. """
-        try:
-            return self.deploy_info.get('slack_channels', ["#test"])[0]
-        except (IndexError, AttributeError):
-            return "#test"
+        if self.deploy_info.get('slack_notify', True) is True:
+            try:
+                return self.deploy_info.get('slack_channels', ["#deploy"])[0]
+            except (IndexError, AttributeError):
```
EvanKrall: seems like you should never get AttributeError, since you're specifying a default in the call to `.get()`.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -631,13 +631,16 @@ def get_slack_client(self) -> SlackClient:
...
     def get_slack_channel(self) -> str:
-        """ Safely get some slack channel to post to. Defaults to ``#test``.
+        """ Safely get some slack channel to post to. Defaults to ``#deploy``.
         Currently only uses the first slack channel available, and doesn't support
         multi-channel notifications. """
-        try:
-            return self.deploy_info.get('slack_channels', ["#test"])[0]
-        except (IndexError, AttributeError):
-            return "#test"
+        if self.deploy_info.get('slack_notify', True) is True:
```
EvanKrall: `is True` seems unnecessary. Schemas are a better way to be strict about this being a boolean in deploy.yaml.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -631,13 +631,16 @@ def get_slack_client(self) -> SlackClient:
...
+            try:
+                return self.deploy_info.get('slack_channels', ["#deploy"])[0]
+            except (IndexError, AttributeError):
+                return "#deploy"
+        else:
+            return '#deploy'
```
EvanKrall: let's make this a constant, since it's used 3 times.

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: (I would just drop the default argument on get, and rely on the `except` block.)

solarkennedy: ok

paasta_tools/cli/cmds/mark_for_deployment.py
solarkennedy: It is unnecessary, I just thought it read better

paasta_tools/cli/cmds/mark_for_deployment.py
solarkennedy: can do

PR #2286, opened 2019-06-18
No jobs keyword again

solarkennedy:
I was able to unify the `job` loading code, but not quite the action loading code.

For action names, we really just need the name, and not the other code for parsing other things about the action, so I thought it was an ok compromise.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -883,40 +883,6 @@ def test_load_tron_yaml_empty(mock_read_file, mock_read_service_info):
     mock_read_service_info.assert_has_calls([mock.call('foo', 'tron-dev', soa_dir)])
 
 
-def test_get_tron_instance_list_from_yaml_finds_actions_properly():
-    fake_tron_job_config = {
-        "ssh_option": "foo",
-        "jobs": [
-            {
-                "name": "job1",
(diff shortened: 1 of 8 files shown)


tests/test_utils.py
```diff
@@ -869,25 +869,37 @@ def test_get_service_instance_list_ignores_underscore():
...
+    soa_dir = '/other/services'
+
+    with pytest.raises(utils.NoConfigurationForServiceError):
+        utils.load_tron_yaml('foo', 'dev', soa_dir=soa_dir)
+
+    assert mock_read_service_info.call_count == 1
+    mock_read_service_info.assert_has_calls([mock.call('foo', 'tron-dev', soa_dir)])
+
+
+def test_get_tron_instance_list_from_yaml_with_dicts():
```
qui: Are you going to follow up with more changes so that `actions` is always a dict? Otherwise, I'd prefer to keep 2 test cases, one where actions is a list and one where it's not.

qui:
mypy and general_itests failures look real, too.

tests/test_utils.py
solarkennedy: I'll add the lists test

PR #2288, opened 2019-06-19
Add cassandracluster replication checker

mattmb:
Included is a bunch of boilerplate so that generate_deployments will
work for cassandracluster files.


mattmb:
I've got some tests to fix...

PR #2291, opened 2019-06-21
Make primary slack message for mark-for-deployment more compact by mo…

EvanKrall:
…ving detail into a reply in the thread.


solarkennedy:
This summary and emoji thing.. idk shipit

EvanKrall:
The emoji thing is to prevent the :alert: emojis from spinning forever after we roll back for an SLO violation.

PR #2298, opened 2019-06-26
Email Spark framework users if possible

qui:
Guess from the framework user and name. It won't work for Spark apps defined in code, which are often run by root and have a custom framework name, but those should be covered by the service notification.

Also return with the appropriate return code so we can control the frequency that our team gets the full check output.

### DIFF
--- a/paasta_tools/check_spark_jobs.py
+++ b/paasta_tools/check_spark_jobs.py
@@ -2,7 +2,11 @@
 import argparse
 import datetime
 import logging
+import smtplib
+import sys
 from collections import defaultdict
+from email.message import EmailMessage
+from socket import getfqdn
 
 import pysensu_yelp
 import requests
@@ -14,6 +18,7 @@ from paasta_tools.utils import list_services
 
 
 logger = logging.getLogger(__name__)
+email_from_address = f'paasta@{getfqdn()}'
 
 
 JUPYTER_PREFIX = 'jupyterhub_'
@@ -33,6 +38,11 @@ def parse_args():
         '--no-notify',
         action='store_true',
         help='Skip notifying the teams that own each framework',
+    )
+    parser.add_argument(
+        '--email-domain',
+        default=None,
+        help='Email domain for notifying users',
     )
     return parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


solarkennedy:
Hmm. It would be nice if we could "just" set the notification_email for sensu, but in practice this check could catch multiple different users at once.

paasta_tools/check_spark_jobs.py
```diff
@@ -34,6 +39,11 @@ def parse_args():
         action='store_true',
         help='Skip notifying the teams that own each framework',
     )
+    parser.add_argument(
+        '--email-domain',
+        default=None,
```
solarkennedy: I think you could get away with `socket.getfqdn().split('.', 1)[1]`

paasta_tools/check_spark_jobs.py
```diff
@@ -148,7 +158,32 @@ def update_check_status(service, output, status):
...
+    msg['Subject'] = f'Long-running Spark framework {framework_info["name"]}'
+    content = 'Please check why it is still running and terminate if appropriate.\n'
+    content += format_framework(framework_info)
+    msg.set_content(content)
+    with smtplib.SMTP('localhost') as s:
+        s.send_message(msg)
+
+
+def report_spark_jobs(min_hours, no_notify, email_domain=None):
     results = get_matching_framework_info(
```
solarkennedy: I think this variable `results` would be better named `frameworks`

paasta_tools/check_spark_jobs.py
```diff
@@ -173,13 +208,18 @@ def report_spark_jobs(min_hours, no_notify):
                 update_check_status(service, message, pysensu_yelp.Status.WARNING)
             else:
                 update_check_status(service, 'No long running spark jobs', pysensu_yelp.Status.OK)
+        if email_domain:
+            for result in results:
+                email_user(result, email_domain)
+
```
solarkennedy: I think it would be cool if there was an `else` here that would print who it would email and what services it would alert

qui:
> Hmm. It would be nice if we could "just" set the notification_email for sensu, but in practice this check could catch multiple different users at once.

Would it be able to use sensu if there's no valid service/team associated with the framework? I think we could use a list of notification emails, but getting a valid service is the hard part.

paasta_tools/check_spark_jobs.py
qui: `dev.yelpcorp...`? I tried this, Gmail did not like it.

paasta_tools/check_spark_jobs.py
qui: The guessed service is already part of messages_by_service we print above. Added a print statement to `email_user`, too.

PR #2309, opened 2019-07-05
enable spark event logging by default

lydian:


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -532,6 +534,21 @@ def get_spark_config(
     # Limit a container's cpu usage
     non_user_args['spark.mesos.executor.docker.parameters'] += ',cpus={}'.format(user_args['spark.executor.cores'])
 
+    if 'spark.eventLog.enabled' not in user_args:
```
huadongliu: I would leave spark.eventLog.enabled in user_args with default true so that users can disable event logging if they want to.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -532,6 +534,21 @@ def get_spark_config(
...
+        user_args['spark.eventLog.enabled'] = 'true'
+
+    if 'spark.eventLog.dir' not in user_args and user_args['spark.eventLog.enabled'] == 'true':
```
huadongliu: Line 494 "Update default configuration" overrides default user spark options with command line args.  Please move spark.eventLog.dir to user_args with defaults based on args.cluster.

lydian: Thanks for the suggestion, already changed to that one :)

PR #2313, opened 2019-07-09
Error on all methods in draining itest

qui:
An attempt to improve itests in response to DAR-634. We have tests to make sure the bounce makes progress when some draining methods are failing, might as well check all of them.

This would have caught the error in DAR-634 (fails at L135, "the old app should be running") if we took out the "Intentionally crashing for testing purposes" so that the Exception has no args and would have triggered the e.args[0] IndexError. But now we have a unit test for both the 0 arg and 1 arg cases, so keeping it seems reasonable?

### DIFF
--- a/paasta_tools/drain_lib.py
+++ b/paasta_tools/drain_lib.py
@@ -149,10 +149,13 @@ class CrashyDrainDrainMethod(NoopDrainMethod):
     async def drain(self, task: DrainTask) -> None:
         raise Exception("Intentionally crashing for testing purposes")
 
-
-@register_drain_method('crashy_is_safe_to_kill')
-class CrashySafeToKillDrainMethod(NoopDrainMethod):
     async def is_safe_to_kill(self, task: DrainTask) -> bool:
+        raise Exception("Intentionally crashing for testing purposes")
+
+    async def stop_draining(self, task: DrainTask) -> None:
+        raise Exception("Intentionally crashing for testing purposes")
+
+    async def is_draining(self, task: DrainTask) -> bool:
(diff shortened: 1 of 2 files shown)


paasta_tools/drain_lib.py
```diff
@@ -149,12 +149,15 @@ class CrashyDrainDrainMethod(NoopDrainMethod):
...
-class CrashySafeToKillDrainMethod(NoopDrainMethod):
     async def is_safe_to_kill(self, task: DrainTask) -> bool:
         raise Exception("Intentionally crashing for testing purposes")
 
+    async def stop_draining(self, task: DrainTask) -> None:
+        raise Exception("Intentionally crashing for testing purposes")
+
+    async def is_draining(self, task: DrainTask) -> bool:
+        raise Exception("Intentionally crashing for testing purposes")
+
```
EvanKrall: I think my reading of setup_marathon_job is that these four methods would all get called -- but I'm not totally sure. Is it worth counting calls and making sure?

qui: Is it possible to count calls with behave? I looked into counting calls in the unit tests, but there's no test case that calls all the methods.

EvanKrall: behave doesn't provide any functionality to do it AFAIK, but we could make the methods increment some class variables or something.

qui: Hm, I don't think it's worth it. The important behavior is that the bounce works even if the methods crash. I did verify that they're being called now, and that the test fails, if you break their exception handlers.

PR #2314, opened 2019-07-10
use endpoint for status cmd - verbose is not working as of now

felfilali:
This is working for non verbose. 
The method chronos_instance_status does not support verbose.

so this is a WIP until i figure out a work around for verbose.


paasta_tools/cli/cmds/status.py
```diff
@@ -840,12 +842,6 @@ def paasta_status(
     :param args: argparse.Namespace obj created from sys.args by cli"""
     soa_dir = args.soa_dir
     system_paasta_config = load_system_paasta_config()
-
-    if 'USE_API_ENDPOINT' in os.environ:
```
qui: I think we want to keep this as an option until we have merged @stug 's changes for Marathon and confirmed that everything works on the API as expected.

paasta_tools/cli/cmds/status.py
```diff
@@ -74,10 +72,9 @@
     FlinkDeploymentConfig,
     KubernetesDeploymentConfig,
     AdhocJobConfig,
-]
-SSH_ONLY_INSTANCE_CONFIG = [
     ChronosJobConfig,
 ]
+SSH_ONLY_INSTANCE_CONFIG = []
```
qui: On the other hand, I think you can remove this.

paasta_tools/chronos_serviceinit.py
```diff
@@ -314,18 +314,33 @@ def status_chronos_jobs(client, jobs, job_config, verbose):
...
+        soa_dir=soa_dir,
+    )
+    if sorted_matching_jobs == []:
         return "%s: chronos job is not set up yet" % PaastaColors.yellow("Warning")
     else:
         output = []
         desired_state = job_config.get_desired_state_human()
-        output.append("Desired:    %s" % desired_state)
-        for job in jobs:
+        output.append("    Desired:    %s" % desired_state)
```
qui: Instead of adding the spaces in front everywhere in chronos_serviceinit, you could copy https://github.com/Yelp/paasta/blob/f5f9f0f3d214b90ca0384a9d9ffe349531dd749d/paasta_tools/cli/cmds/status.py#L621-L622 into print_chronos_status.

tests/api/test_instance.py
```diff
@@ -66,14 +67,20 @@ def test_instances_status_marathon(
...
 @mock.patch('paasta_tools.api.views.instance.validate_service_instance', autospec=True)
 @mock.patch('paasta_tools.api.views.instance.get_actual_deployments', autospec=True)
-@mock.patch('paasta_tools.api.views.instance.select_tasks_by_id', autospec=True)
-@mock.patch('paasta_tools.api.views.instance.get_cached_list_of_running_tasks_from_frameworks', autospec=True)
+@mock.patch('paasta_tools.chronos_serviceinit.select_tasks_by_id', autospec=True)
+@mock.patch('paasta_tools.chronos_serviceinit.get_cached_list_of_running_tasks_from_frameworks', autospec=True)
+@mock.patch('paasta_tools.chronos_serviceinit.chronos_tools.lookup_chronos_jobs', autospec=True)
+@mock.patch('paasta_tools.chronos_serviceinit.chronos_tools.get_job_type', autospec=True)
+@mock.patch('paasta_tools.chronos_serviceinit.chronos_tools.get_chronos_status_for_job', autospec=True)
 def test_chronos_instance_status(
```
qui: You should just mock `paasta_tools.chronos_serviceinit.status_chronos_jobs`, assert that it's called correctly, and delete the other assertions here. You don't need to test the implementation of chronos_serviceinit (e.g. that it calls lookup_chronos_jobs and stuff) because the chronos_serviceinit tests should cover that already.

felfilali:
https://fluffy.yelpcorp.com/i/b6PMRdnZCTTw3tRXXSfqRPvwN6qWMCDB.html

paasta_tools/chronos_serviceinit.py
```diff
@@ -283,14 +282,14 @@ def format_chronos_job_status(client, job, running_task_count, verbose=0):
...
-        mesos_status = f"{mesos_status}\n{mesos_status_verbose}"
+        mesos_status = f"      {mesos_status}\n      {mesos_status_verbose}"
     return (
-        "Job:     %(job_name)s\n"
-        "  Status:   %(disabled_state)s (%(chronos_state)s)"
-        "  Last:     %(last_result)s (%(formatted_time)s)\n"
-        "  %(schedule_type)s: %(schedule_value)s\n"
-        "  Command:  %(command)s\n"
-        "  Mesos:    %(mesos_status)s" % {
+        "    Job:     %(job_name)s\n"
```
qui: Do you still need the extra whitespace here, after adding the whitespace in `print_chronos_status`?

paasta_tools/cli/cmds/status.py
```diff
@@ -70,12 +70,11 @@
...
 from paasta_tools.utils import SystemPaastaConfig
 
+
 HTTP_ONLY_INSTANCE_CONFIG: Sequence[Type[InstanceConfig]] = [
     FlinkDeploymentConfig,
     KubernetesDeploymentConfig,
     AdhocJobConfig,
-]
-SSH_ONLY_INSTANCE_CONFIG = [
     ChronosJobConfig,
```
qui: Chronos shouldn't be HTTP_ONLY, we want to be able to switch between SSH and HTTP with the use_api_endpoint option.

tests/api/test_instance.py
```diff
@@ -90,24 +85,12 @@ def test_chronos_instance_status(
...
-    )
 
     request = testing.DummyRequest()
     request.swagger_data = {'service': 'fake_service', 'instance': 'fake_instance'}
 
-    response = instance.instance_status(request)
-    assert response['chronos']['schedule']['schedule'] == 'always'
-    assert response['chronos']['schedule_type'] == 'schedule'
+    instance.instance_status(request)
+    assert mock_status_chronos_jobs.called
```
qui: You could assert `mock_status_chronos_jobs.call_args` are what you expect, to check the cluster, etc.

paasta_tools/cli/cmds/status.py
felfilali: ahh ok, I was under the impression that we wanted to rm ssh all together.

qui:
I think the SSH only logic is still wrong - we don't need SSH_ONLY_INSTANCES anymore because everything can go through the API, but if something is not in HTTP_ONLY, then it should use HTTP or SSH based on use_api_endpoint.

Now, it looks like if you request a Chronos instance with use_api_endpoint=True, we would request its status twice, both through the api and ssh.

paasta_tools/cli/cmds/status.py
```diff
@@ -70,12 +70,13 @@
...
 from paasta_tools.utils import SystemPaastaConfig
 
+
 HTTP_ONLY_INSTANCE_CONFIG: Sequence[Type[InstanceConfig]] = [
     FlinkDeploymentConfig,
     KubernetesDeploymentConfig,
     AdhocJobConfig,
 ]
-SSH_ONLY_INSTANCE_CONFIG = [ChronosJobConfig]
+SSH_ONLY_INSTANCE_CONFIG: Sequence[Type[InstanceConfig]] = [ChronosJobConfig]
```
qui: I thought this list should be empty? Since Chronos can go either way now.

felfilali: so when it's empty, with the original logic, we use ssh even with use_api_enpoint being true.

should I rm it from the list?

felfilali:
https://fluffy.yelpcorp.com/i/3PscPQRH1NVVlX8QqmVGV0QvsrVkdM2P.html

PR #2317, opened 2019-07-10
Marathon status api endpoint

stug:
Eek, this is kind of big 😬 

Mostly this is a rough copy of just the data fetching parts of marathon_serviceinit.

### DIFF
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -680,6 +680,7 @@
                         "Stopped",
                         "Delayed",
                         "Waiting",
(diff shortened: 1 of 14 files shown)


paasta_tools/api/api_docs/swagger.json
```diff
@@ -652,6 +653,25 @@
           "format": "int32",
           "description": "The number of desired instances of the service"
         },
+        "autoscaling_info": {
```
solarkennedy: This is potentially where you could have shipped this incrementally, by adding one additional piece of data at a time (auto, status, mesos, ss)

PR #2318, opened 2019-07-10
format output of updated marathon instance status api endpoint

stug:
This formats the output of the changes in https://github.com/Yelp/paasta/pull/2317

Now ready for review

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -417,37 +417,8 @@ def create_mesos_running_tasks_table(running_tasks):
     ]
     rows.append(table_header)
     for task in running_tasks:
-        if task.rss.value is not None and task.mem_limit.value is not None:
-            mem_percent = 100 * task.rss.value / task.mem_limit.value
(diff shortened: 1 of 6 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -299,33 +305,357 @@ def print_marathon_status(
...
+            [task.id, task.hostname, mem_string, cpu_string, deployed_at_string]
+        )
+        rows.extend(format_tail_lines_for_mesos_task(task.tail_lines, task.id))
+
+    return format_table(rows)
+
+
+def get_mesos_task_memory_string(task):
+    if task.rss.value is None or task.mem_limit.value is None:
+        return task.mem_limit.error_message
```
solarkennedy: Is this guaranteed to be set in this case if either are these are none?

stug: Good point -- I'll make it look at both `error_message` fields.  They should both be `None` if one of them is because they they both come from the task stats on the underlying Mesos task, but that's not necessarily something on this side of the API should know about.

PR #2321, opened 2019-07-15
Secret names can't have underscores for k8s

mattmb:


mattmb:
Sure, let me fix them up

PR #2324, opened 2019-07-16
Added test_is_old_task_missing_healthchecks_with_real_data test

solarkennedy:
I was trying to make sure `is_old_task_missing_healthchecks` was working exactly like we thought it did, but it is a bit complex because it depends on the interface to marathon python.

From what I can tell it is still correct.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -802,11 +802,15 @@ def filter_autoscaling_tasks(
     marathon_tasks = {}
     for app in marathon_apps:
         for task in app.tasks:
-            if task.id.startswith(job_id_prefix) and (
-                is_task_healthy(task) or not
-                app.health_checks or is_old_task_missing_healthchecks(task, app)
-            ):
-                marathon_tasks[task.id] = task
+            if task.id.startswith(job_id_prefix):
+                if not app.health_checks:
+                    # If we don't have healthchecks defined anyway, then we should just
+                    # let the task through the filter
+                    marathon_tasks[task.id] = task
+                elif is_task_healthy(task) or is_old_task_missing_healthchecks(task, app):
+                    # healthy is good, but also we need to consider those old tasks missing
+                    # health checks, because those are still valid candidates for autoscaling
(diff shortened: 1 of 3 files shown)


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -802,11 +802,15 @@ def filter_autoscaling_tasks(
...
-                marathon_tasks[task.id] = task
+            if task.id.startswith(job_id_prefix):
+                if not app.health_checks:
+                    # If we don't have healthchecks defined anyway, then we should just
+                    # let the task through the filter
+                    marathon_tasks[task.id] = task
+                elif is_task_healthy(task) or is_old_task_missing_healthchecks(task, app):
+                    # healthy is good, but also we need to consider those old tasks missing
+                    # health checks, because those are still valid candidates for autoscaling
+                    marathon_tasks[task.id] = task
```
EvanKrall: I'm not sure I like the repetition of the task body here. What about:

```
if task.id.startswith(job_id_prefix):
    if any([
    ]):

File: tests/example-marathon-app-with-missing-healthchecks.json [hunk 2]
```diff
@@ -0,0 +1,450 @@
+{
```
EvanKrall: There's a lot of stuff in this json that isn't relevant to the test. Can you pare it down to only the parts necessary?

EvanKrall:
If you think it works the same way as before, and this new test passes, what's the point of this change? Sounds like you haven't figured out what's wrong yet.

paasta_tools/autoscaling/autoscaling_service_lib.py
solarkennedy: I agree that is better.
I couldn't understand the previous format, and I wasn't super happy with the nested ifs.

tests/example-marathon-app-with-missing-healthchecks.json
solarkennedy: yes

solarkennedy:
It is true, the point of this change is simply to increase test coverage and confidence that this function works.

PR #2328, opened 2019-07-17
slos attribute disappeared from sink object between slo-transcoder 2.…

EvanKrall:
…4.0 and 2.5.6 - access it a different way


solarkennedy:
Autospec somehow? :(

EvanKrall:
> Autospec somehow? :(

Unfortunately since slo-transcoder is an internal library, it isn't installed in the py36-linux tox environment, so we can't autospec.

PR #2330, opened 2019-07-18
cap_add support in k8s

chlgit:
Add capabilities for containers. For example, yelp-main main_uswest1 instances needs these capabilities: "DAC_OVERRIDE", "CHOWN", "SETGID", "SETUID"]

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -40,6 +40,7 @@ from kubernetes.client import models
 from kubernetes.client import V1AWSElasticBlockStoreVolumeSource
 from kubernetes.client import V1beta1PodDisruptionBudget
 from kubernetes.client import V1beta1PodDisruptionBudgetSpec
+from kubernetes.client import V1Capabilities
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -553,6 +555,16 @@ def get_liveness_probe(
 
         return probe
 
+    def get_security_context(self):
```
vkhromov: please annotate the return type.

chlgit: kk, done.

paasta_tools/kubernetes_tools.py
```diff
@@ -553,6 +555,16 @@ def get_liveness_probe(
 
         return probe
 
+    def get_security_context(self) -> V1SecurityContext:
```
vkhromov: since the function returns `None` in some cases, the return type is actually `Optional[V1SecurityContext]`.

chlgit: Ah, you're right. Let me fix it in the separated PR.

PR #2331, opened 2019-07-18
Avoid updating slack messages if they are unchanged; print blocks whe…

EvanKrall:
…n we get an invalid_blocks error.

This is intended to:

1. Avoid the rate_limited errors we're getting when enough mark-for-deployments are running in parallel
2. Help track down the invalid_blocks error we're getting sometimes when initially posting the details message.

### DIFF
--- a/paasta_tools/automatic_rollbacks/slack.py
+++ b/paasta_tools/automatic_rollbacks/slack.py
@@ -94,6 +94,8 @@ class SlackDeploymentProcess(DeploymentProcess, abc.ABC):
         self.human_readable_status = "Initializing..."
         self.slack_client = self.get_slack_client()
         self.last_action = None
+        self.summary_blocks_str = ''
+        self.detail_blocks_str = ''
         self.send_initial_slack_message()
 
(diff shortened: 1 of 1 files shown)


paasta_tools/automatic_rollbacks/slack.py
```diff
@@ -295,28 +295,42 @@ def send_initial_slack_message(self):
...
-            channel=self.slack_channel_id,
-            blocks=detail_blocks,
-            ts=self.detail_slack_ts,
-        )
-        if resp["ok"] is not True:
-            log.error("Posting to slack failed: {}".format(resp["error"]))
+        summary_blocks_str = json.dumps(summary_blocks, sort_keys=True)
+        detail_blocks_str = json.dumps(detail_blocks, sort_keys=True)
+
+        if getattr(self, 'old_summary_blocks_str', '') != summary_blocks_str:
```
solarkennedy: this getattr things is kind weird, why not make this a normal attribute and default it to empty string?

EvanKrall: Yeah, that's probably better.

PR #2333, opened 2019-07-19
Include kubernete.application module in setup.py so it is installed 

EmanekaT:
application folder is not installed right now as kubernetes is a module. This should fix it. I built a package and tested on kubestage.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ setup(
     author_email='kwa@yelp.com',
     description='Tools for Yelps SOA infrastructure',
     packages=find_packages(exclude=("tests*", "scripts*")),
+    py_modules=['paasta_tools.kubernetes.application'],
     include_package_data=True,
     install_requires=get_install_requires(),
     scripts=[


solarkennedy:
Why do we need this? We don't need this for any other entry point?
Do we need the k8s stuff to be so nested? We don't deeply nest any other stuff in the codebase?

EmanekaT:
> Why do we need this? We don't need this for any other entry point?
> Do we need the k8s stuff to be so nested? We don't deeply nest any other stuff in the codebase?

@solarkennedy  If I have paasta_tools.kubernetes module, I would get https://travis-ci.org/Yelp/paasta/jobs/559178335 (I don't know why. python3 should use absolute import by default?) So I removed __init__.py from paasta_tools/kubernetes. Then paasta_tools/kubernetes/applications is not installed since paasta_tools/kubernetes is not a module anymore. 

I would like to use nested folders in the first place because I think it's more readable than having a bunch of files.

solarkennedy:
I think that "flat is better than nested" here. Just one folder for kubernetes stuff is probably just fine.

PR #2334, opened 2019-07-19
spark event logging decides bucket based on aws credentials account

lydian:
tested locally with get_default_event_log_dir output the expected result

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -35,15 +35,15 @@ from paasta_tools.utils import PaastaNotConfiguredError
 from paasta_tools.utils import SystemPaastaConfig
 
 
-AWS_CREDENTIALS_DIR = '/etc/boto_cfg/'
-DEFAULT_SPARK_RUN_CONFIG = '/nail/srv/configs/spark.yaml'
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -330,14 +330,35 @@ def get_docker_run_cmd(
...
+    with open(DEFAULT_SPARK_RUN_CONFIG) as fp:
+        spark_run_conf = YAML().load(fp.read())
+    try:
+        account_id = boto3.client(
+            'sts',
+            aws_access_key_id=access_key,
+            aws_secret_access_key=secret_key,
+        ).get_caller_identity().get('Account')
+    except Exception:
+        account_id = None
```
vkhromov: It worth to `log.warn` about the exception.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -330,14 +330,35 @@ def get_docker_run_cmd(
...
+
+    for conf in spark_run_conf['environments'].values():
+        if account_id == conf['account_id']:
```
vkhromov: Can this condition ever be true if `account_id` is `None` after the exception above?
If it can't then it's better to `return None` from the exception handler above instead of doing this loop.

PR #2335, opened 2019-07-22
Drop mem validation for cassandra

mattmb:
Just temporary whilst we fix a bug and move to using the same units as
everything else.


ronin13:
LGTM.

PR #2340, opened 2019-07-27
Use `black` Python code formatter

vkhromov:
Add call to `black` from the pre-commit hooks
Reformat the existing sources using `black`.
Please see https://github.com/psf/black for more details.


solarkennedy:
I, for one, welcome our strict auto-formatting overlords!
It looks like there are type issues in the test failures. Does black handle our type annotations?

keymone:
just curious - why black instead of yapf? as i recall, black was formatting things differently to our internal guidelines and yapf was configurable enough to adhere to those.

vkhromov:
> just curious - why black instead of yapf? as i recall, black was formatting things differently to our internal guidelines and yapf was configurable enough to adhere to those.

There are several reasons to choose `black` over `yapf`:
* `yapf` sometimes non-deterministic while running for the same code in the same environment, which means that you can have `yapf` changing files already changed in the previous `yapf` run, which is annoying for pre-commit hooks, as you'll have several failures in the row for `git ci` .
* `yapf` definitely non-deterministic between Python versions, e.g. see https://github.com/google/yapf/issues/696.
* `yapf` doesn't support some Python 3.x syntax which we're using here, like f-strings:
```
(py36) vkhromov@dev53-uswest1adevc paasta % pip install yapf
Collecting yapf
  Downloading https://files.pythonhosted.org/packages/79/22/d711c0803b6c3cc8c96eb54509f23fec1e3c078d5bfc6eb11094e762e7bc/yapf-0.28.0
-py2.py3-none-any.whl (180kB)
     |████████████████████████████████| 184kB 16.1MB/s
Installing collected packages: yapf
Successfully installed yapf-0.28.0
WARNING: You are using pip version 19.1.1, however version 19.2.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.

(py36) vkhromov@dev53-uswest1adevc paasta % yapf paasta_tools
yapf: directory specified without '--recursive' flag: paasta_tools
(py36) vkhromov@dev53-uswest1adevc paasta % yapf --recursive paasta_tools
  File "/nail/home/vkhromov/github.com/Yelp/paasta/.tox/py36/lib/python3.6/site-packages/yapf/yapflib/pytree_utils.py", line 115, in
 ParseCodeToTree
    tree = parser_driver.parse_string(code, debug=False)
  File "/usr/lib/python3.6/lib2to3/pgen2/driver.py", line 106, in parse_string
    return self.parse_tokens(tokens, debug)
  File "/usr/lib/python3.6/lib2to3/pgen2/driver.py", line 71, in parse_tokens
    if p.addtoken(type, value, (prefix, start)):
  File "/usr/lib/python3.6/lib2to3/pgen2/parse.py", line 159, in addtoken
    raise ParseError("bad input", type, value, context)
lib2to3.pgen2.parse.ParseError: bad input: type=3, value='"{instance_type_class.config_filename_prefix}-{cluster}"', context=('', (1
21, 16))
```
* A bit speculative, but the ultimate goal of `black` seems to be to become the default formatter for any Python code, so eventually `black` formatting convention will be universal.

keymone:
alright, i don't care much which formatter we use as long as we use one :)

vkhromov:
> It looks like there are type issues in the test failures. Does black handle our type annotations?

It does.

The issue here is that `mypy` expects `# type: ignore` comments at the same line as the typing error, not at the end of a statement with the typing error.  So when `black` splits a statement into several lines, it puts the comment at the end of the statement, which sometimes could be at a different line from the line with the typing error.
This is a bit annoying, I fixed that mostly by putting comment at the end of correct lines, but it's more preferable to reduce amount of such comments to the absolute minimum.

.pre-commit-config.yaml
```diff
@@ -36,21 +31,26 @@ repos:
...
-    -   id: add-trailing-comma
-        args: [--py36-plus]
+    -   id: flake8
+        exclude: ^docs/source/conf.py$
+-   repo: https://github.com/pre-commit/mirrors-autopep8
+    rev: v1.4.3
+    hooks:
+    -   id: autopep8
+-   repo: https://github.com/psf/black
+    rev: stable
```
asottile: fwiw you shouldn't use floating refs if you want determinism

tox.ini
```diff
@@ -173,6 +173,7 @@ commands =
 
 [flake8]
 max-line-length = 120
+ignore = E501,E203,W503
```
asottile: you want `extend-ignore=E501`

.pre-commit-config.yaml
vkhromov: unfortunately, it doesn't seem that `black` has tags for stable releases but only for alpha/beta ones: https://github.com/psf/black/releases

asottile: right, you want to use one of those, the `stable` tag is just a moving alias for the latest beta

PR #2342, opened 2019-07-30
Deprecate Ubuntu Trusty

vkhromov:


solarkennedy:
Keep in mind we still deploy our tools to some trusty boxes (`mco find -C paasta_tools::install -F lsbdistcodename=trusty`)

If you ship this, you will have to either make sure we don't try to deploy them there in the first place, or just not care, or whatever.

vkhromov:
> Keep in mind we still deploy our tools to some trusty boxes (`mco find -C paasta_tools::install -F lsbdistcodename=trusty`)
> 
> If you ship this, you will have to either make sure we don't try to deploy them there in the first place, or just not care, or whatever.

Added `itest_trusty` back, so we have trusty debs.
Everything else is run using Xenial.

solarkennedy:
> > Keep in mind we still deploy our tools to some trusty boxes (`mco find -C paasta_tools::install -F lsbdistcodename=trusty`)
> > If you ship this, you will have to either make sure we don't try to deploy them there in the first place, or just not care, or whatever.
> 
> Added `itest_trusty` back, so we have trusty debs.
> Everything else is run using Xenial.

That works. As long as itest_trusty still spits out a trusty deb, then we can still do this change to make everything else xenial-by-default.

PR #2343, opened 2019-07-30
Refactor deployd to use a single queue for simplicity.

EvanKrall:


paasta_tools/deployd/common.py
```diff
@@ -203,3 +129,46 @@ def get_marathon_clients_from_config() -> MarathonClients:
...
+        try:
+            while True:
+                wait_until, bounce_by, si = self.unavailable_service_instances.get_nowait()
+                if wait_until < now:
+                    self.available_service_instances.put((bounce_by, si))
+                else:
+                    self.unavailable_service_instances.put((wait_until, bounce_by, si))
+                    return
+        except Empty:
+            print('empty')
```
solarkennedy: nix

paasta_tools/deployd/metrics.py
```diff
@@ -26,12 +23,9 @@ def __init__(self, inbox, cluster, metrics_provider):
...
-                self.instances_to_bounce_later.qsize(),
-            )
-            self.instances_that_need_to_be_checked_up_on_gauge.set(
-                len(self.inbox.keys()),
+                self.queue.unavailable_service_instances.qsize(),
             )
             self.instances_to_bounce_now_gauge.set(
-                self.instances_to_bounce_now.qsize(),
+                self.queue.available_service_instances.qsize(),
             )
```
solarkennedy: Although not in this PR, with this change it would be neat to add additional metrics, like "oldest entry in the available queue" and "average bounce_by time" or something

paasta_tools/deployd/common.py
```diff
@@ -48,39 +44,24 @@ def __new__(
...
         cluster: str,
         bounce_by: float,
+        wait_until: float,
         failures: int = 0,
         bounce_timers: Optional[BounceTimers] = None,
-        priority: Optional[int] = None,
         processed_count: int = 0,
     ) -> 'ServiceInstance':
-        if priority is None:
-            priority = get_priority(service, instance, cluster)
```
solarkennedy: Don't forget to remove this configuration option as a second pass

PR #2345, opened 2019-07-30
apply min/max_instance clamps later

drmorr0:
This change

a) always calls `set_instances_for_marathon_service`
b) applies the min/max instance clamps at s-m-j time

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -706,6 +706,12 @@ def autoscale_marathon_instance(
                         ),
                         level="event",
                     )
+            else:
+                write_to_log(
+                    config=marathon_service_config,
+                    line="Staying at %d instances (%s)" % (current_instances, humanize_error(error)),
+                    level="debug",
+                )
     except LockHeldException:
         log.warning(
             "Skipping autoscaling run for {service}.{instance} because the lock is held".format(
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -142,8 +142,6 @@ def threshold_decision_policy(current_instances, error, **kwargs):
 def proportional_decision_policy(
     zookeeper_path,
     current_instances,
-    min_instances,
-    max_instances,
     setpoint,
     utilization,
     num_healthy_instances,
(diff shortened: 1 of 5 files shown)


paasta_tools/setup_marathon_job.py
```diff
@@ -696,24 +698,25 @@ def log_deploy_error(errormsg: str, level: str = "event") -> None:
             new_app, draining_hosts=draining_hosts
         )
         if new_app.instances < config["instances"] + num_at_risk_tasks:
+            new_instance_count = config["instances"] + num_at_risk_tasks
+            if max_instances and max_instances < new_instance_count:
```
vkhromov: If `max_instances` can be zero, then `if max_instances is not None and ...`

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -688,33 +679,28 @@ def autoscale_marathon_instance(
...
+                write_to_log(
+                    config=marathon_service_config,
+                    line="Delaying scaling *down* as we found too few healthy tasks running in marathon. "
+                    "This can happen because tasks are delayed/waiting/unhealthy or because we are "
+                    "waiting for tasks to be killed. Will wait for sufficient healthy tasks before "
+                    "we make a decision to scale down.",
+                    level="debug",
+                )
+                return
+            else:
```
solarkennedy: Won't this create a lot of excess noise when the old/new are the same?

paasta_tools/setup_marathon_job.py
drmorr0: Fixed.

paasta_tools/autoscaling/autoscaling_service_lib.py
drmorr0: Fixed.

drmorr0:
I restored the "Staying at <whatever>" messages that got erased in #2260; I'm not sure if they got removed intentionally or not but I think they are useful and helped us understand what happened here.  I also am reasonably convinced this solves the problem, so I'm ready to commit if @solarkennedy you agree :)

thebostik:
> I restored the "Staying at " messages that got erased in #2260; I'm not sure if they got removed intentionally or not but I think they are useful and helped us understand what happened here. I also am reasonably convinced this solves the problem, so I'm ready to commit if @solarkennedy you agree :)

That works for me!

solarkennedy:
@drmorr0 can you rebase?

drmorr0:
> @drmorr0 can you rebase?

Oops this has already been merged.

PR #2346, opened 2019-07-31
Internalize `yelp-clog`

vkhromov:


yelp_package/extra_requirements_yelp.txt
```diff
@@ -13,6 +13,7 @@ slo-transcoder==2.5.6
 smmap2==2.0.3
 vault-tools==0.7.34
 yelp-cgeom==1.3.1
+yelp-clog==3.3.1
```
solarkennedy: Please keep the version the same for now. Make this PR a "noop" change with regards to the behavior of the codebase.

Make sure that the jenkins build is green, etc.

PR #2347, opened 2019-07-31
Use yelp-clog 3.3.1

vkhromov:


solarkennedy:
Can you confirm that the tests + itests pass locally when using a Yelp environment?

vkhromov:
> Can you confirm that the tests + itests pass locally when using a Yelp environment?

Yep, both `make test` and `make itest` are fine when run from my devbox.

PR #2350, opened 2019-08-01
Use Python 3.7

vkhromov:
Replace most of `3\.?6` with corresponding `3.7`, bump incompatible
dependencies, and fix PEP 479 [1] related issues.
    
Also, removed `http_drain_method` test since it fails at github and
the failure isn't reproduceable locally.  Let's fix it later.
    
[1] https://www.python.org/dev/peps/pep-0479/


.pre-commit-config.yaml
```diff
@@ -44,8 +44,8 @@ repos:
     rev: stable
     hooks:
     -   id: black
-        language_version: python3.6
-        args: ['--target-version', 'py36']
+        language_version: python3.7
+        args: ['--target-version', 'py37']
```
keymone: we can probably skip this as it's in `pyproject.toml`?

requirements-unchecked.txt
```diff
@@ -0,0 +1,2 @@
+git+git://github.com/benoitc/http-parser.git@2306df28fbd7946036a2609776a77833f5dda81e#egg=http-parser
```
keymone: why do we need this file?

vkhromov: There is no release of `http-parser` which supports Python 3.7
See e.g. https://github.com/benoitc/http-parser/issues/80 .
So, we have to compile it from `master`.

drolando:
> lgtm, but do we have python 3.7 on all paasta hosts?
yeah both xenial and bionic have python3.7 and we don't have any paasta trusty host anymore.

requirements-unchecked.txt
drolando: Not saying that you need to do it in this PR, but that project seems unmaintained. https://github.com/MagicStack/httptools seems to be pretty similar and has recent commits, so you should probably switch over. Or fork / take ownership of http-parser

vkhromov: > Not saying that you need to do it in this PR, but that project seems unmaintained. https://github.com/MagicStack/httptools seems to be pretty similar and has recent commits, so you should probably switch over. Or fork / take ownership of http-parser

Please tell that to `pymesos` developers :)
We don't use `http-parser` directly, it's a transitive dependency.

.pre-commit-config.yaml
vkhromov: Let's do that in another PR, we already have had too much pain with `black` and Python version which it uses.

Jenkinsfile
```diff
@@ -6,7 +6,7 @@ yproperties() // Sets releng approved global properties (SCM polling, build log
 CHANNELS = ['paasta']
 GIT_SERVER = 'git@github.com'
 PACKAGE_NAME = 'mirrors/Yelp/paasta'
-DIST = ['trusty', 'xenial', 'bionic']
```
solarkennedy: We still need trusty debs. Things have not changed from https://github.com/Yelp/paasta/pull/2342

vkhromov: It seems that deadsnakes officially doesn't support Python 3.7 on Trusty: https://github.com/deadsnakes/issues/issues/63

EvanKrall: looks like some of the roles require paasta-tools because meshd depends on `paasta_dump_locally_running_services`. Maybe there's a way to break that dependency; maybe meshd doesn't always need it, or maybe we can publish `paasta_dump_locally_running_services` as a separate package that continues to run on py36 for trusty support.

vkhromov:
Still have 17 Trusty boxes running PaaSTA:
```
$ mco find -C paasta_tools::install -F lsbdistcodename=trusty | wc -l
17
```

nemacysts:
this should be unblocked now - the remaining trusty box is in the process of being decomm'd and puppet is already broken on it (and there are no plans to fix puppet on it)

vkhromov:
Hm, cannot reproduce `http_drain_method` test failure either on devbox or in a docernised Ubuntu 20.x.

PR #2352, opened 2019-08-02
Fix broken yelp-clog dependencies

vkhromov:
Install `monk` 0.7.0 to fix broken `yelp-clog` 3.3.1 dependencies.

For testing, built `deb` package, installed it locally, and checked that the issue was fixed.
Also run `make test` and `make itest`.

### DIFF
--- a/yelp_package/extra_requirements_yelp.txt
+++ b/yelp_package/extra_requirements_yelp.txt
@@ -6,6 +6,7 @@ gitdb2==2.0.3
 gitpython==2.1.9
 inflection==0.3.1
 markdown==2.4
+monk==0.7.0
 python-jsonschema-objects==0.3.1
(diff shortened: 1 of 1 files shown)


vkhromov:
> Usually with a new requirement someone does a `check-requirements` or something to make sure we've pinned in other stuff that monk comes with. Can you run that to make sure there shouldn't also be other stuff added?

AFAIK `check-requirements` is run during `make test`.
Anyway:
```
vkhromov@dev53-uswest1adevc paasta % . .tox/py36-linux/bin/activate
(py36-linux) vkhromov@dev53-uswest1adevc paasta % check-requirements
Checking requirements...
....                                                                                                                         [100%]
4 passed in 1.06 seconds
```

PR #2353, opened 2019-08-05
More test artefacts to ignore

vkhromov:


vkhromov:
local `make test` and `make itest` are fine as well.

PR #2354, opened 2019-08-06
Change implementation of DelayDeadlineQueue to avoid getting stuck.

EvanKrall:
The previous implementation would only put things in available_service_instances upon entering get() or put(). If nothing put()s for a long time, and all watchers are currently waiting on get(), ServiceInstances could be stuck in unavailable_service_instances.

### DIFF
--- a/paasta_tools/deployd/common.py
+++ b/paasta_tools/deployd/common.py
@@ -1,7 +1,8 @@
+import asyncio
+import concurrent
 import logging
 import time
 from collections import namedtuple
-from queue import Empty
 from queue import PriorityQueue
 from queue import Queue
 from threading import Thread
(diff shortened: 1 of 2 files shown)


solarkennedy:
Gosh. I'm not going to pretend I understand this :(

mattmb:
Hmm, I'm not sure about this. Feels a bit weird using asyncio and Threads

Can we not just call `process_unavailable_service_instances` somewhere else? Like in another Thread or even in the main loop: https://github.com/Yelp/paasta/blob/master/paasta_tools/deployd/master.py#L124

keymone:
i agree with mattmb, not a fan of adding asyncio to cover small gap in functionality. can we just have original implementation + separate thread to put things from unavailable to available instead of doing it in get/put?

    def __init__():
        ...
        self.process_unavailable_service_instances_loop()
    
    def process_unavailable_service_instances_loop(self) -> None:
        def inner() -> None:
            while True:
                self.process_unavailable_service_instances()
                time.sleep(1)
    
        Thread.new(target=inner, daemon=True)

EvanKrall:
@keymone Isn't this change "original implementation + separate thread to put things from unavailable to available instead of doing it in get/put"?

I don't like the while True/sleep 1 suggestion. That's just a poor man's event loop, and it wastes CPU cycles to produce less accurate results, plus it requires a bunch of complexity in `process_unavailable_service_instances` that I avoid by using asyncio.sleep. I could make it sleep the exact amount of time until the next entry (but in some way that's interruptable in case a new entry shows up; maybe by waiting on an Event with a timeout), but that's essentially what the asyncio event loop is doing; why not use that?

I also think my original choice of using a PriorityQueue for `unavailable_service_instances` was unfortunate, since there's no documented way to peek at the first item. (Though it looks like I could just use `unavailable_service_instances.queue[0]` since `.queue` is a list managed by [heapq](https://docs.python.org/3/library/heapq.html))

@mattmb I do think it's unfortunate that I have to launch another thread here to manage this; if deployd were entirely asyncio-based, this change would be less awkward. (But that's a PR for a different time)

EvanKrall:
Closing in favor of #2358

PR #2355, opened 2019-08-06
refactor setup_kubernetes_job to use controller_wrappers

EmanekaT:
I tested on kubestage. I installed this package, disabled it for all other than the three apps I used for the test, and ran `bash -c "paasta_list_kubernetes_service_instances | shuf |xargs setup_kubernetes_job -v"`
Change: https://reviewboard.yelpcorp.com/r/428545/
Result: https://fluffy.yelpcorp.com/i/TrKP15nGsZmRLP1WlTnR0TBHHwg9w74D.html

### DIFF
--- a/paasta_tools/setup_kubernetes_job.py
+++ b/paasta_tools/setup_kubernetes_job.py
@@ -23,30 +23,20 @@ Command line options:
 import argparse
 import logging
 import sys
-from typing import Optional
 from typing import Sequence
-from typing import Tuple
 from typing import Union
 
-from kubernetes.client import V1beta1PodDisruptionBudget
-from kubernetes.client import V1DeleteOptions
 from kubernetes.client import V1Deployment
 from kubernetes.client import V1StatefulSet
-from kubernetes.client.rest import ApiException
 
-from paasta_tools.kubernetes_tools import create_deployment
-from paasta_tools.kubernetes_tools import create_pod_disruption_budget
-from paasta_tools.kubernetes_tools import create_stateful_set
+from paasta_tools.kubernetes.application.controller_wrappers import Application
+from paasta_tools.kubernetes.application.controller_wrappers import DeploymentWrapper
+from paasta_tools.kubernetes.application.controller_wrappers import StatefulSetWrapper
 from paasta_tools.kubernetes_tools import ensure_namespace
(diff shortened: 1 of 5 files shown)


tests/test_setup_kubernetes_job.py
```diff
@@ -175,340 +147,237 @@ def test_ensure_pod_disruption_budget_replaces_outdated():
...
-        assert not mock_create_kubernetes_application.called
-        mock_deploy = (
-            mock_load_kubernetes_service_config_no_cache.return_value.format_kubernetes_app()
-        )
-        mock_update_kubernetes_application.assert_called_with(
-            kube_client=mock_kube_client, application=mock_deploy
-        )
+        mock_stateful_set_wrapper.assert_called_with(mock_deploy)
 
-        # changed configsha so update
```
mattmb: did you capture all these cases in the new test? I'm not sure if they're 100% necessary but I'd like to keep them if we can easily

paasta_tools/setup_kubernetes_job.py
```diff
@@ -82,6 +73,7 @@ def parse_args() -> argparse.Namespace:
 
 
 def main() -> None:
+    global setup_kube_succeeded
```
mattmb: how come the change to a global? I'd prefer to keep the return values if it's not too tricky

tests/test_setup_kubernetes_job.py
EmanekaT: Yes I did. Just the code is strutured differently, so I removed some that does not make sens e anymore.  But the testcases like 'changed configsha so update' are all there.

tests/test_setup_kubernetes_job.py
```diff
@@ -51,78 +45,45 @@ def test_main():
...
+        mock_service_instances = ["kuruptf_m"]
+        setup_kube_deployments(
+            kube_client=mock_client,
+            service_instances=mock_service_instances,
+            soa_dir="/nail/blah",
         )
+        assert mock_create_application_object.call_count == 0
 
 
 def test_ensure_pod_disruption_budget_create():
```
qui: Since the code has moved to controller_wrappers, the tests for pod disruption budgets should go there too.

PR #2356, opened 2019-08-07
Remove paasta generate-pipeline command

nemacysts:
It's probably time to rip the bandaid off completely now. At this point,
the only services that could be using this are probably:
    * using customized pipelines (which this doesn't support)
    * or are doing something non-standard (and probably not changing at all)
    * {insert another reason here}

In any case, it's probably time to remove this escape hatch and, if
necessary, make our current standard pipelines support any of the
esoteric usages of generate-pipline or move jobs to custom Jenkinsfiles.


solarkennedy:
Thanks! use `git grep generate-pipeline` to find other references and to fix the tests.

nemacysts:
ugh, my bad - `make itest` took long enough that I (for some reason) assumed that it did was running the package itests too

PR #2357, opened 2019-08-07
Adjust implementation of DelayDeadlineQueue so that put() calls made …

EvanKrall:
…while all ServiceInstances are in unavailable_service_instances will return as the wait_until times pass, instead of blocking until another thread enters get() or put().

This is my attempt at avoiding the need for another thread entirely. I think it's a little insane, and I am not as confident in its correctness as #2354.


EvanKrall:
closing in favor of #2358

PR #2358, opened 2019-08-07
A third implementation of DelayDeadlineQueue that uses a background thread but not asyncio

EvanKrall:
A simpler version of https://github.com/Yelp/paasta/pull/2357

### DIFF
--- a/paasta_tools/deployd/common.py
+++ b/paasta_tools/deployd/common.py
@@ -4,6 +4,8 @@ from collections import namedtuple
 from queue import Empty
 from queue import PriorityQueue
 from queue import Queue
+from threading import Condition
+from threading import Event
 from threading import Thread
 from typing import Any
 from typing import Collection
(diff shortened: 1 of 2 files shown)


paasta_tools/deployd/common.py
```diff
@@ -150,43 +152,46 @@ def __init__(self) -> None:
...
+                        wait_until, bounce_by, si = (
+                            self.unavailable_service_instances.get_nowait()
+                        )
+                        if wait_until < time.time():
+                            self.available_service_instances.put_nowait((bounce_by, si))
+                        else:
+                            self.unavailable_service_instances.put_nowait(
+                                (wait_until, bounce_by, si)
+                            )
+                            timeout = wait_until - time.time()
```
mattmb: Is it possible this could be negative and is it "okay" if it is?

EvanKrall: Yeah, Condition.wait() [internally checks for timeout <= 0](https://github.com/python/cpython/blob/3.7/Lib/threading.py#L299)

PR #2359, opened 2019-08-07
upgraded mypy to 0.711

EmanekaT:
I need to upgrade mypy so get method in typeddict could be recognized. https://github.com/python/mypy/issues/3843

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -297,7 +297,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         )
 
     def copy(self) -> "KubernetesDeploymentConfig":
-        return self.__class__(
+        return self.__class__(  # type: ignore
             service=self.service,
             instance=self.instance,
             cluster=self.cluster,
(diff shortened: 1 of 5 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1471,7 +1471,9 @@ def load_custom_resource_definitions(
     for custom_resource_dict in system_paasta_config.get_kubernetes_custom_resources():
         kube_kind = KubeKind(**custom_resource_dict.pop("kube_kind"))  # type: ignore
         custom_resources.append(
-            CustomResourceDefinition(kube_kind=kube_kind, **custom_resource_dict)
+            CustomResourceDefinition(  # type: ignore
```
EmanekaT: I didn't fix this because this was ignored already.

paasta_tools/paasta_metastatus.py
```diff
@@ -271,12 +271,17 @@ def fill_table_rows_with_service_instance_stats(
     limiting_factor = "Unknown"
     # service_instance_stats.keys() should be a subset of resource_free_dict
     for rsrc_name, rsrc_amt_wanted in service_instance_stats.items():
-        if rsrc_amt_wanted > 0:
+        if rsrc_amt_wanted > 0:  # type: ignore
```
EmanekaT: rsrc_amt_wanted could be either int or float, and i know how to "fix" this. This behavior itself is not possible if we use static types.

paasta_tools/tron_tools.py
```diff
@@ -29,7 +29,7 @@
 try:
     from yaml.cyaml import CSafeDumper as Dumper
 except ImportError:  # pragma: no cover (no libyaml-dev / pypy)
-    Dumper = yaml.SafeDumper
+    Dumper = yaml.SafeDumper  # type: ignore
```
EmanekaT: Same. Dumper should never be of two different types if it's statically typed.

PR #2360, opened 2019-08-08
K8s storage classes

keymone:
Support two simple options for storage class name attribute for persistent volumes: `ebs` and `ebs-slow`.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -695,10 +695,10 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         ]
 
     def get_storage_class_name(self, volume: PersistentVolume) -> str:
-        storage_class_name = volume.storage_class_name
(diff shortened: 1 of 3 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -694,12 +694,12 @@ def get_volume_claim_templates(self) -> Sequence[V1PersistentVolumeClaim]:
...
         ]
 
-    def get_storage_class_name(self) -> str:
-        # TODO: once we support node affinity this method should return the
-        # name of the storage class for a particular AZ. We should also enforce
-        # that a storage class exists with the correct name (like we do for the
-        # paasta namespace object)
-        return "ebs-gp2-us-west-1a"
+    def get_storage_class_name(self, volume: PersistentVolume) -> str:
+        storage_class_name = volume["storage_class_name"]
```
mattmb: can we default it to "ebs"? otherwise we force people to add more yaml

PR #2362, opened 2019-08-09
Add HPA for paasta services(k8s deployments)

EmanekaT:
Summary
* HPA is only created for deployments.
* Whenever setup_kubernetes_job runs, HPA is synced to the lastest autoscaling setting. It is created, updated, or deleted based on yelpsoa_config setting.
* existing number of replicas in deployment objects do not change when the service is updated if autoscaling is enabled. 
* I assume when instance keyword exists in yelpsoa_config, autoscaling is disabled.

Tested on kubestage.
https://fluffy.yelpcorp.com/i/rcMS6M9r6jVP8ncJ557z6lQb5RRFsDzq.html

### DIFF
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -13,6 +13,7 @@ from kubernetes.client import V2beta1CrossVersionObjectReference
 from kubernetes.client import V2beta1HorizontalPodAutoscaler
 from kubernetes.client import V2beta1HorizontalPodAutoscalerSpec
 from kubernetes.client import V2beta1MetricSpec
+from kubernetes.client import V2beta1PodsMetricSource
 from kubernetes.client import V2beta1ResourceMetricSource
 from kubernetes.client.rest import ApiException
 
@@ -37,6 +38,14 @@ class Application(ABC):
         item: Union[V1Deployment, V1StatefulSet],
         logging=logging.getLogger(__name__),
     ) -> None:
+        """
+        This Application wrapper is an interface for creating/deleting k8s deployments and statefulsets
+        soa_config is KubernetesDeploymentConfig. It is not loaded in init because it is not always required.
+        :param item: Kubernetes Object(V1Deployment/V1StatefulSet) that has already been filled up.
+        :param logging: where logs go
+        """
+        if not item.metadata.namespace:
+            item.metadata.namespace = "paasta"
         self.kube_deployment = KubeDeployment(
             service=item.metadata.labels["yelp.com/paasta_service"],
             instance=item.metadata.labels["yelp.com/paasta_instance"],
@@ -196,12 +205,21 @@ class DeploymentWrapper(Application):
         self.delete_pod_disruption_budget(kube_client)
         self.delete_horizontal_pod_autoscaler(kube_client)
 
+    def get_existing_app(self, kube_client: KubeClient):
+        return kube_client.deployments.read_namespaced_deployment(
+            name=self.item.metadata.name, namespace=self.item.metadata.namespace
+        )
+
     def create(self, kube_client: KubeClient) -> None:
         create_deployment(kube_client=kube_client, formatted_deployment=self.item)
         self.ensure_pod_disruption_budget(kube_client)
         self.sync_horizontal_pod_autoscaler(kube_client)
 
     def update(self, kube_client: KubeClient) -> None:
+        # If autoscaling is enabled, do not update replicas.
+        # In all other cases, replica is set to max(instances, min_instances)
+        if not self.get_soa_config().get("instances"):
+            self.item.spec.replicas = self.get_existing_app(kube_client).spec.replicas
         update_deployment(kube_client=kube_client, formatted_deployment=self.item)
         self.ensure_pod_disruption_budget(kube_client)
         self.sync_horizontal_pod_autoscaler(kube_client)
@@ -211,6 +229,9 @@ class DeploymentWrapper(Application):
         In order for autoscaling to work, there needs to be at least three configurations
         min_instnace, max_instance, and instance
         """
+        self.logging.info(
+            "Syncing HPA setting for {self.item.metadata.name}/name in {self.item.metadata.namespace}"
+        )
         hpa_exists = self.exists_hpa(kube_client)
         # NO autoscaling
         if self.get_soa_config().get("instances"):
@@ -256,8 +277,8 @@ class DeploymentWrapper(Application):
             metrics.append(
                 V2beta1MetricSpec(
                     type="Pods",
-                    resource=V2beta1ResourceMetricSource(
-                        name="http", target_average_value=target
+                    pods=V2beta1PodsMetricSource(
+                        metric_name="http", target_average_value=target
                     ),
                 )
             )
@@ -265,8 +286,8 @@ class DeploymentWrapper(Application):
             metrics.append(
                 V2beta1MetricSpec(
                     type="Pods",
-                    resource=V2beta1ResourceMetricSource(
-                        name="uwsgi", target_average_value=target
+                    pods=V2beta1PodsMetricSource(
+                        metric_name="uwsgi", target_average_value=target
                     ),
                 )
             )
@@ -288,11 +309,18 @@ class DeploymentWrapper(Application):
                 ),
             ),
         )
+        self.logging.debug(body)
         if not hpa_exists:
+            self.logging.info(
+                "Creating new HPA for {metrics_provider} {self.item.metadata.name}/name in {self.item.metadata.namespace}"
+            )
             kube_client.autoscaling.create_namespaced_horizontal_pod_autoscaler(
                 namespace=self.item.metadata.namespace, body=body, pretty=True
             )
         else:
+            self.logging.info(
+                "Updating new HPA for {metrics_provider} {self.item.metadata.name}/name in {self.item.metadata.namespace}/namespace"
+            )
             kube_client.autoscaling.patch_namespaced_horizontal_pod_autoscaler(
                 name=self.item.metadata.name,
                 namespace=self.item.metadata.namespace,
@@ -305,6 +333,12 @@ class DeploymentWrapper(Application):
             kube_client.autoscaling.read_namespaced_horizontal_pod_autoscaler(
                 name=self.item.metadata.name, namespace=self.item.metadata.namespace
             )
+        except ValueError as e:
+            self.logging.error(
+                "Error occurs for {self.item.metadata.name}/name in {self.item.metadata.namespace}/namespace.  \
+                This could a bug in k8s. For detail: https://github.com/kubernetes-client/python/issues/415"
+            )
+            self.logging.error(e)
         except ApiException as e:
             if e.status == 404:
                 return False
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -205,8 +205,8 @@ class DeploymentWrapper(Application):
 
     def sync_horizontal_pod_autoscaler(self, kube_client: KubeClient) -> None:
         """
-        In order for autoscaling to work, there needs to be at least three configurations
-        min_instnace, max_instance, and instance
+        In order for autoscaling to work, there needs to be at least two configurations
+        min_instnace, max_instance, and there cannot be instance.
         """
         self.logging.info(
             f"Syncing HPA setting for {self.item.metadata.name}/name in {self.item.metadata.namespace}"
@@ -221,6 +221,11 @@ class DeploymentWrapper(Application):
         body = self.soa_config.get_autoscaling_metric_spec(
             name=self.item.metadata.name, namespace=self.item.metadata.namespace
         )
+        if not body:
+            raise Exception(
+                f"CRIT: autoscaling misconfigured for {self.kube_deployment.service}.\
+                    {self.kube_deployment.instance}. Please correct the configuration and  update pre-commit hook."
+            )
         self.logging.debug(body)
         if not hpa_exists:
             self.logging.info(
@@ -263,7 +268,7 @@ class DeploymentWrapper(Application):
                 # Deployment does not exist, nothing to delete but
                 # we can consider this a success.
                 self.logging.debug(
-                    f"not deleting nonexistent HPA/{self.item.metadata.name} from namespace/{self.item.metadata.namespace}".format(
+                    "not deleting nonexistent HPA/{self.item.metadata.name} from namespace/{self.item.metadata.namespace}".format(
                         self.item.metadata.name, self.item.metadata.namespace
                     )
                 )
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -118,7 +118,7 @@ class Application(ABC):
                     )
                 )
             else:
-                raise e
+                raise
         else:
             self.logging.info(
                 "deleted pod disruption budget/{} from namespace/{}".format(
@@ -195,7 +195,7 @@ class DeploymentWrapper(Application):
                     )
                 )
             else:
-                raise e
+                raise
         else:
             self.logging.info(
                 "deleted deploy/{} from namespace/{}".format(
@@ -263,6 +263,10 @@ class DeploymentWrapper(Application):
             self.get_soa_config().get("autoscaling", {}).get("decision_policy", "")
             == "bespoke"
         ):
+            self.logging.error(
+                f"Sorry, bespoke is not implemented yet. Please use a different decision \
+                policy if possible for {self.item.metadata.name}/name in namespace{self.item.metadata.namespace}"
+            )
             return
         elif metrics_provider == "mesos_cpu":
             metrics.append(
@@ -292,7 +296,10 @@ class DeploymentWrapper(Application):
                 )
             )
         else:
-            self.logging.error("Wrong metrics specified")
+            self.logging.error(
+                f"Wrong metrics specified: {metrics_provider} for\
+                {self.item.metadata.name}/name in namespace{self.item.metadata.namespace}"
+            )
             return
 
         body = V2beta1HorizontalPodAutoscaler(
@@ -312,14 +319,14 @@ class DeploymentWrapper(Application):
         self.logging.debug(body)
         if not hpa_exists:
             self.logging.info(
-                "Creating new HPA for {metrics_provider} {self.item.metadata.name}/name in {self.item.metadata.namespace}"
+                f"Creating new HPA for {metrics_provider} {self.item.metadata.name}/name in {self.item.metadata.namespace}"
             )
             kube_client.autoscaling.create_namespaced_horizontal_pod_autoscaler(
                 namespace=self.item.metadata.namespace, body=body, pretty=True
             )
         else:
             self.logging.info(
-                "Updating new HPA for {metrics_provider} {self.item.metadata.name}/name in {self.item.metadata.namespace}/namespace"
+                f"Updating new HPA for {metrics_provider} {self.item.metadata.name}/name in {self.item.metadata.namespace}/namespace"
             )
             kube_client.autoscaling.patch_namespaced_horizontal_pod_autoscaler(
                 name=self.item.metadata.name,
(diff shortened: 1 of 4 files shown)


paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
 
     def create(self, kube_client: KubeClient) -> None:
         create_deployment(kube_client=kube_client, formatted_deployment=self.item)
         self.ensure_pod_disruption_budget(kube_client)
+        self.sync_horizontal_pod_autoscaler(kube_client)
 
     def update(self, kube_client: KubeClient) -> None:
+        # If autoscaling is enabled, do not update replicas.
+        # In all other cases, replica is set to max(instances, min_instances)
+        if not self.get_soa_config().get("instances"):
```
mattmb: I think this is a little confusing. How about if we override `get_instances` in `KubernetesDeploymentConfig` and maybe have an `initial_instances`? I don't 100% know how the HPA works but my guess is that you just set `replicas` to some initial value and then the HPA is used to get the real value?

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+        target = (
+            float(self.get_soa_config().get("autoscaling", {}).get("setpoint", "0.8"))
+            * 100
+        )
+        # TODO support bespoke
+        if (
+            self.get_soa_config().get("autoscaling", {}).get("decision_policy", "")
+            == "bespoke"
+        ):
+            return
```
mattmb: can we log something please

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+                metrics=metrics,
+                scale_target_ref=V2beta1CrossVersionObjectReference(
+                    kind="Deployment", name=self.item.metadata.name
+                ),
+            ),
+        )
+        self.logging.debug(body)
+        if not hpa_exists:
+            self.logging.info(
+                "Creating new HPA for {metrics_provider} {self.item.metadata.name}/name in {self.item.metadata.namespace}"
```
vkhromov: f-string?

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+            )
+            kube_client.autoscaling.create_namespaced_horizontal_pod_autoscaler(
+                namespace=self.item.metadata.namespace, body=body, pretty=True
+            )
+        else:
+            self.logging.info(
+                "Updating new HPA for {metrics_provider} {self.item.metadata.name}/name in {self.item.metadata.namespace}/namespace"
```
vkhromov: f-string?

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+            )
+
+    def exists_hpa(self, kube_client: KubeClient) -> bool:
+        try:
+            kube_client.autoscaling.read_namespaced_horizontal_pod_autoscaler(
+                name=self.item.metadata.name, namespace=self.item.metadata.namespace
+            )
+        except ValueError as e:
+            self.logging.error(
+                "Error occurs for {self.item.metadata.name}/name in {self.item.metadata.namespace}/namespace.  \
```
vkhromov: f-string?

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+                This could a bug in k8s. For detail: https://github.com/kubernetes-client/python/issues/415"
```
vkhromov: It seems that the pointed issue was closed in 2018.
Can you please double-check that it is still there for Kubernetes 1.15.2 and if that's true provide an example of the error?

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+            )
+            self.logging.error(e)
```
vkhromov: It's better to log an error with all corresponding context, in this case - output `e` within the previous message using the same `logging .error`.

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
         self.ensure_pod_disruption_budget(kube_client)
+        self.sync_horizontal_pod_autoscaler(kube_client)
+
+    def sync_horizontal_pod_autoscaler(self, kube_client: KubeClient) -> None:
+        """
+        In order for autoscaling to work, there needs to be at least three configurations
+        min_instnace, max_instance, and instance
+        """
+        self.logging.info(
+            "Syncing HPA setting for {self.item.metadata.name}/name in {self.item.metadata.namespace}"
```
vkhromov: f-string?

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+            # Remove HPA if autoscaling is disabled
+            if hpa_exists:
+                self.delete_horizontal_pod_autoscaler(kube_client)
+                return
+        # Get autoscaling configurations
+        min_replicas = self.get_soa_config().get("min_instances")
+        max_replicas = self.get_soa_config().get("max_instances")
+        if not min_replicas or not max_replicas:
+            self.logging.error(
+                "min/max_instances are not specified. Autoscaling is not enabled."
```
vkhromov: Please add more context for the error

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+            )
+            return
+        metrics_provider = (
+            self.get_soa_config()
+            .get("autoscaling", {})
+            .get("metrics_provider", "mesos_cpu")
+        )
+        # TODO support multiple metrics
```
vkhromov: These `TODO`s, are they going to be fixed in this CR, or they are for the following CRs?

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+        max_replicas = self.get_soa_config().get("max_instances")
+        if not min_replicas or not max_replicas:
+            self.logging.error(
+                "min/max_instances are not specified. Autoscaling is not enabled."
+            )
+            return
+        metrics_provider = (
+            self.get_soa_config()
+            .get("autoscaling", {})
+            .get("metrics_provider", "mesos_cpu")
```
vkhromov: Here and in several places below, why there is "mesos"?

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+            self.logging.error(
+                "Error occurs for {self.item.metadata.name}/name in {self.item.metadata.namespace}/namespace.  \
+                This could a bug in k8s. For detail: https://github.com/kubernetes-client/python/issues/415"
+            )
+            self.logging.error(e)
+        except ApiException as e:
+            if e.status == 404:
+                return False
+            else:
+                raise e
```
vkhromov: just `raise`?

paasta_tools/kubernetes_tools.py
```diff
@@ -795,16 +823,21 @@ def get_pod_template_spec(
         docker_volumes = self.get_volumes(
             system_volumes=system_paasta_config.get_volumes()
         )
+        annotations = {"smartstack_registrations": json.dumps(self.get_registrations())}
+        metrics_provider = self.config_dict.get("autoscaling", {}).get(
+            "metrics_provider", ""
+        )
+        if metrics_provider in {"http", "uwsgi"}:
+            annotations["autoscaling"] = metrics_provider  # type: ignore
```
vkhromov: Can you please add a proper type annotation to `annotations` var to avoid this `# type: ignore` workaround?

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+            kube_client.autoscaling.patch_namespaced_horizontal_pod_autoscaler(
+                name=self.item.metadata.name,
+                namespace=self.item.metadata.namespace,
+                body=body,
+                pretty=True,
+            )
+
+    def exists_hpa(self, kube_client: KubeClient) -> bool:
+        try:
+            kube_client.autoscaling.read_namespaced_horizontal_pod_autoscaler(
```
vkhromov: nit
Checking existence of something by using `read_*` and comparing HTTP error with 404 is quite fragile.
We could get 404 not only because the object doesn't exist, but also because of an API change, because of incorrectly configured LB, or million other reasons.
It's much better to use `list_*` with proper set of selector and compare amount of results with `1`.

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+            if e.status == 404:
+                # Deployment does not exist, nothing to delete but
+                # we can consider this a success.
+                self.logging.debug(
+                    "not deleting nonexistent HPA/{} from namespace/{}".format(
+                        self.item.metadata.name, self.item.metadata.namespace
+                    )
+                )
+            else:
+                raise e
```
vkhromov: just `raise`

paasta_tools/kubernetes_tools.py
```diff
@@ -670,7 +692,13 @@ def get_desired_instances(self) -> int:
         """ For now if we have an EBS instance it means we can only have 1 instance
         since we can't attach to multiple instances. In the future we might support
         statefulsets which are clever enough to manage EBS for you"""
-        instances = super().get_desired_instances()
+
+        if self.get_desired_state() == "start":
+            instances = self.config_dict.get("instances") or self.get_min_instances()
+        else:
+            instances = 0
+            log.debug("Instance is set to stop. Returning '0' instances")
```
vkhromov: Shouldn't the state be explicitly compared with "stop" instead of assuming that it's "stop" since it isn't "start"?

paasta_tools/kubernetes/application/controller_wrappers.py
EmanekaT: HPA updates the replicas parameter proportionally based on metrics. So we would not update replicas here. Ideally, we should just use patch_namespaced_deployment for update, but I'm not sure if that's a good idea here.  
And sure I can override get_instances in KubernetesDeploymentConfig, but KubernetesDeploymentConfig does not know if the formatted_application is used for update or create. So I either need to refactor the code to make it know, or send a request to k8s API to check if it exists or not. Refactoring the format_application is dirtier than what I do now; Sending requests to k8s wastes network resources.

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -188,14 +203,173 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+        )
+        hpa_exists = self.exists_hpa(kube_client)
+        # NO autoscaling
+        if self.get_soa_config().get("instances"):
+            # Remove HPA if autoscaling is disabled
+            if hpa_exists:
+                self.delete_horizontal_pod_autoscaler(kube_client)
+                return
+        # Get autoscaling configurations
+        min_replicas = self.get_soa_config().get("min_instances")
```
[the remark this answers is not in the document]
EmanekaT: Yes I noticed that. I just didn't use it since I think just get the key from dict is cleaner. (e.g., when others are reviewing the code, there is no need to go to another file).

paasta_tools/kubernetes/application/controller_wrappers.py
EmanekaT: Use patch instead of replace is probably the ideal solution here. What do you think?

paasta_tools/kubernetes/application/controller_wrappers.py
EmanekaT: Open issue: https://github.com/kubernetes-client/gen/issues/52
Example on Kubestage: https://fluffy.yelpcorp.com/i/QZtq9Gl99Lw0LrVqpGXh5CVmxMQD7CPJ.html

paasta_tools/kubernetes/application/controller_wrappers.py
EmanekaT: This is the autoscaling config we have in yelpsoa_configs right now. I don't have plans on changing right now.

paasta_tools/kubernetes/application/controller_wrappers.py
EmanekaT: Yes, they will be fixed once I finished implementing autoscaler.

paasta_tools/kubernetes/application/controller_wrappers.py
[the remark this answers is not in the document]
qui: If we use `self.soa_config.get_*()` consistently, we can change the default values for configs more easily.

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -180,22 +195,181 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+        max_replicas = self.get_soa_config().get("max_instances")
+        if not min_replicas or not max_replicas:
+            self.logging.error(
+                "Please specify min_instances and max_instances for autoscaling to work"
+            )
+            return
+        metrics_provider = (
```
qui: I think most of the logic that isn't making API calls - like deciding which V2beta1MetricSpec to use - should go in KubernetesDeploymentConfig in methods like `get_autoscaling_metric_spec`, not here in the controller.

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -115,8 +130,8 @@ def ensure_pod_disruption_budget(
         self, kube_client: KubeClient
     ) -> V1beta1PodDisruptionBudget:
         pdr = pod_disruption_budget_for_service_instance(
-            service=self.item.service,
-            instance=self.item.instance,
+            service=self.kube_deployment.service,
+            instance=self.kube_deployment.instance,
             min_instances=self.soa_config.get_desired_instances()
```
qui: Hm, I'm wondering how we want to handling autoscaling for pod disruption budgets. As is, we would be willing to go pretty low because get_desired_instances returns min_instances if autoscaling is enabled. But we also can't update the PDB all the time.

Can we use a percentage instead of min_instances? https://kubernetes.io/docs/tasks/run-application/configure-pdb/#specifying-a-poddisruptionbudget

paasta_tools/kubernetes/application/controller_wrappers.py
qui: I think you may still need to implement get_instances to reflect autoscaling because it's used in other places - for example, `check_kubernetes_services_replication.py`. But I don't have an opinion on whether you should use that or create a method to patch the deployment.

paasta_tools/kubernetes/application/controller_wrappers.py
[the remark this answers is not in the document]
EmanekaT: Make sense. I'll update that.

paasta_tools/kubernetes/application/controller_wrappers.py
EmanekaT: Perhaps, I just didn't want to add more methods to KubernetesDeploymentConfig. It's already super long. But I'll move the code there.

tests/kubernetes/application/test_controller_wrapper.py
```diff
@@ -1,79 +1,290 @@
...
+        app.get_soa_config.return_value = config_dict
+        app.soa_config.config_dict = config_dict
+        app.exists_hpa.return_value = exists_hpa
+
+    def mock_get_autoscaling_metric_spec(name, namespace):
+        return KubernetesDeploymentConfig.get_autoscaling_metric_spec(
+            self=mock_config, name=name, namespace=namespace
         )
+
+    def mock_max_instance():
```
qui: You should just let get_autoscaling_metric_spec be a mock and test the different config_dict cases in the config class (in test_kubernetes_tools).

tests/kubernetes/application/test_controller_wrapper.py
```diff
@@ -1,79 +1,290 @@
...
+def test_sync_horizontal_pod_autoscaler():
+    mock_client = mock.MagicMock(autospec=True)
+    app = mock.MagicMock(autospec=True)
+    app.item.metadata.name = "fake_name"
+    app.item.metadata.namespace = "faasta"
+    mock_config = mock.MagicMock(autospec=True)
+
+    # helper Functions for mocking
+    def setup_app(app, config_dict, exists_hpa):
+        app.reset_mock()
```
qui: My preference is to split up tests into different test cases instead of resetting mocks, but I see this resetting pattern already exists in paasta_tools, so *shrug*

tests/kubernetes/application/test_controller_wrapper.py
```diff
@@ -1,79 +1,290 @@
...
-        mock_req_pdr.spec.min_available = 10
+        mock_req_pdr.spec.max_unavailable = 10
         mock_pdr_for_service_instance.return_value = mock_req_pdr
 
-        mock_client = mock.Mock()
+        mock_client = mock.MagicMock()
 
         mock_pdr = mock.Mock()
-        mock_pdr.spec.min_available = 10
+        mock_pdr.spec.max_unavailable = 10
```
qui: This mock_pdr seems unused.

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -180,22 +174,107 @@ def deep_delete(self, kube_client: KubeClient) -> None:
...
+            f"Syncing HPA setting for {self.item.metadata.name}/name in {self.item.metadata.namespace}"
+        )
+        hpa_exists = self.exists_hpa(kube_client)
+        # NO autoscaling
+        if self.get_soa_config().get("instances"):
+            # Remove HPA if autoscaling is disabled
+            if hpa_exists:
+                self.delete_horizontal_pod_autoscaler(kube_client)
+            return
+        body = self.soa_config.get_autoscaling_metric_spec(
```
qui: This may be None, should handle that case too.

paasta_tools/kubernetes_tools.py
```diff
@@ -331,6 +359,81 @@ def get_bounce_method(self) -> str:
             )
         return KUBE_DEPLOY_STATEGY_MAP[bounce_method]
 
+    def get_autoscaling_metric_spec(
+        self, name: str, namespace: str = "paasta"
+    ) -> V2beta1HorizontalPodAutoscaler:
```
qui: Optional, if it can return None

paasta_tools/kubernetes_tools.py
```diff
@@ -331,6 +359,81 @@ def get_bounce_method(self) -> str:
...
+                V2beta1MetricSpec(
+                    type="Pods",
+                    pods=V2beta1PodsMetricSource(
+                        metric_name="uwsgi", target_average_value=target
+                    ),
+                )
+            )
+        else:
+            log.error(
+                f"Wrong metrics specified: {metrics_provider} for\
```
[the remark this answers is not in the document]
EmanekaT: Yes, I'll do it once autoscaling works. But let's leave it here.

tests/kubernetes/application/test_controller_wrapper.py
EmanekaT: Can I just use what I have now? It's painful to mock return values for get_autoscaling_metric_spec, and rewrite what I have now.

tests/kubernetes/application/test_controller_wrapper.py
EmanekaT: Same as ^^

tests/kubernetes/application/test_controller_wrapper.py
EmanekaT: My preference is to feed some real data and mock the kube_client...

paasta_tools/kubernetes/application/controller_wrappers.py
EmanekaT: I raise an error if this happens.

PR #2365, opened 2019-08-09
Upgrade pre-commit

chriskuehl:
Newer versions of pre-commit are able to run hooks in parallel, leading to faster hook runs, both at commit time and test time.

Before this, `pre-commit run --all-files` takes 100.79 seconds on my devbox; afterwards, it takes 27.53 seconds.

### DIFF
--- a/requirements-dev-minimal.txt
+++ b/requirements-dev-minimal.txt
@@ -7,7 +7,7 @@ hypothesis
 mock
 path.py
 pep8
-pre-commit==1.10.1
+pre-commit
 pylint
 pytest
 pytest-asyncio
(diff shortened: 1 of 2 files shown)


vkhromov:
Some of pre-commit hooks can change files being committed, in some cases several pre-commit hooks can change the same file, wondering how that works when they are run in parallel.

chriskuehl:
@vkhromov to avoid that problem, only one hook is run at a time, but that hook is run multiple times concurrently over different chunks of files.

In theory you could get an even greater speedup by running all of the hooks at once, but as you mentioned, it would lead to them stomping over each others' changes.

You can see the implementation details at https://github.com/pre-commit/pre-commit/pull/851

PR #2368, opened 2019-08-13
Add new metrics about deployd queues.

EvanKrall:
[Here](https://app.signalfx.com/#/dashboard/EB0FKzIAgAA?groupId=EB0FKkdAcAA&configId=EB0FL7GAgAA&startTimeUTC=1565659876000&endTimeUTC=1565662563000) is an example of where I'm going with this -- you can see the yelp-main cert bump from 18:41 to 18:48, then at 18:59 I restarted deployd, so from 18:59 to 19:11 you can see the startup behavior.

During the yelp-main bump, because so many instances are added to the queue at once, you can see `instances_with_past_deadline` jump, as well as `max_time_past_deadline` and `average time past deadline` (which is `max_time_past_deadline` / `instances_with_past_deadline`) on the second chart. You can also see how we stop processing the backlog when all these to-be-bounced-immediately instances show up.

The backlog = `instances_to_bounce_now` - `available_instances_with_deadline_in_next_60s` = all instances in `available_instances` with deadline more than 60s in the future.

### DIFF
--- a/paasta_tools/deployd/master.py
+++ b/paasta_tools/deployd/master.py
@@ -15,7 +15,7 @@ from paasta_tools.deployd.common import PaastaQueue
 from paasta_tools.deployd.common import PaastaThread
 from paasta_tools.deployd.common import ServiceInstance
 from paasta_tools.deployd.leader import PaastaLeaderElection
-from paasta_tools.deployd.metrics import QueueMetrics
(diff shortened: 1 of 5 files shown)


EvanKrall:
Yeah; the number of metrics would probably be fine but I've always found SignalFX a little clumsy when dealing with aggregations of time series that come and go.

PR #2372, opened 2019-08-14
PAASTA-15618 Added a check if a pod is a paasta service

azretkenzhaliev:
Added a check if a pod is a paasta service by checking if it has labels to eliminate an NoneType is not subscriptable exception.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1166,7 +1166,8 @@ def filter_pods_by_service_instance(
     return [
         pod
         for pod in pod_list
-        if pod.metadata.labels["yelp.com/paasta_service"] == service
+        if pod.metadata.labels is not None
+        and pod.metadata.labels["yelp.com/paasta_service"] == service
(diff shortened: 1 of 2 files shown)


azretkenzhaliev:
@mattmb please, review 👍

paasta_tools/kubernetes_tools.py
```diff
@@ -1163,12 +1163,16 @@ def get_all_pods(kube_client: KubeClient, namespace: str = "paasta") -> Sequence
...
     pod_list: Sequence[V1Pod], service: str, instance: str
 ) -> Sequence[V1Pod]:
-    return [
-        pod
-        for pod in pod_list
-        if pod.metadata.labels["yelp.com/paasta_service"] == service
-        and pod.metadata.labels["yelp.com/paasta_instance"] == instance
-    ]
+    pod_search_results = []
+    for pod in pod_list:
```
keymone: can we retain list comprehension form of this function? you can add `pod.metadata.labels is not None and ...` to the `if` condition.

azretkenzhaliev: We can. Added. 👍

PR #2374, opened 2019-08-16
RFC: Prevent scaling down services if the utilization > setpoint

matthewbentley:
The affects the moving_average forcast policy.  If the usage was low for
several minutes, this could affect the average enough to prompt scaling
down in more than one service autoscaler run.
This prevents that case by refusing to scale down if the current usage
is greater than the setpoint (or the high point of the good_enough_window).

Internal ticket PAASTA-15711

I could also add some logging of what this would do in different situations; eg, what it would have done before this change, or what it would do with an exponentially weighted average.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -194,9 +194,9 @@ def proportional_decision_policy(
     if desired_number_instances < current_instances:
         if good_enough_window:
             _, high = good_enough_window
-            if utilization - offset >= high:
+            if utilization >= high:
                 desired_number_instances = current_instances
-        elif utilization - offset >= setpoint:
+        elif utilization >= setpoint:
             desired_number_instances = current_instances
 
     if good_enough_window:
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -188,6 +188,17 @@ def proportional_decision_policy(
 
     desired_number_instances = int(round(predicted_load / (setpoint - offset)))
 
+    # Don't scale down if the current utilization >= the setpoint (or the high point of the good enough window)
+    # This prevents the case where the moving_average forcast_policy thinks the service needs to scale
+    #  down several times in a row due to under-utilization in the near past
+    if desired_number_instances < current_instances:
+        if good_enough_window:
+            _, high = good_enough_window
+            if utilization - offset >= high:
+                desired_number_instances = current_instances
+        elif utilization - offset >= setpoint:
+            desired_number_instances = current_instances
+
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -188,6 +188,17 @@ def proportional_decision_policy(
 
     desired_number_instances = int(round(predicted_load / (setpoint - offset)))
 
+    # Don't scale down if the current utilization >= the setpoint (or the high point of the good enough window)
+    # This prevents the case where the moving_average forcast_policy thinks the service needs to scale
+    #  down several times in a row due to under-utilization in the near past
+    if desired_number_instances < current_instances:
+        if good_enough_window:
+            _, high = good_enough_window
+            if utilization - offset >= high:
```
qui: Why are we subtracting offset (the baseline load) again?

For example, if the good_enough_window is [0.4, 0.6], and utilization is 0.65, even if the baseline load of a container is 0.1, don't we still want to prevent scaling down?

qui:
Also I would like to see a unit test for this if possible.

paasta_tools/autoscaling/autoscaling_service_lib.py
matthewbentley: Isn't the point of the offset to say "if utilization is reported as 0.65, but the offset is 0.1, then utilization is actually 0.55"?  At least that's what the rest of the code (and the docs IIRC) implies.

qui: I'm not exactly sure if the setpoint and good enough windows are supposed to include the offset or not, I'll defer to krall.

EvanKrall: We add the offset back in when comparing to the good enough window: https://github.com/Yelp/paasta/blob/739d9255ed2a2c3208b65bb2adb7d5527aec45e9/paasta_tools/autoscaling/autoscaling_service_lib.py#L193-L197

So I think we actually don't want to subtract it here.

matthewbentley: I'm still not sure what the "correct" action would be here, but I agree that not subtracting it is more consistent with the current code.

EvanKrall: The goal of the service autoscaler is to keep a particular metric near a target setpoint. The offset is just used to help figure out how much the metric will change by a particular scaling choice. (e.g. if you've got a service that has some baseline CPU per container that doesn't depend on traffic levels, we want to make sure we correct for that.)

We want the settings to match the graphs as much as possible -- if you've got a setpoint of 0.5, your graph should be close to 0.5 if the autoscaler is working correctly, regardless of your offset.

PR #2380, opened 2019-08-21
Added alphanumeric validation for paasta cli --secret-name argument

azretkenzhaliev:


tests/cli/test_cmds_secret.py
```diff
@@ -106,6 +107,53 @@ def test__get_secret_provider_for_service():
         )
 
 
+def test_check_secret_name():
+    assert secret.check_secret_name("Yelp-Yay-2019_20_08") == "Yelp-Yay-2019_20_08"
+    assert secret.check_secret_name("KGZ-2019_20_08") == "KGZ-2019_20_08"
+
+    try:
```
Rob-Johnson: you should use ``pytest.raises`` here instead

http://doc.pytest.org/en/latest/assert.html#assertions-about-expected-exceptions

azretkenzhaliev: Added.

PR #2381, opened 2019-08-22
add json schema for k8s and logic to call validate and merge with marath…

felfilali:
about tests:
I kind of just tagged along previous marathon tests. please let me know if I should add new ones or if I missed any.

about schema content:
I followed the doc that should have a comprehensive list of content, but I also added ones from current k8s yaml files we have in place. lmk if i need to rm or add some more.

about validation: 
Is there anything specific to k8s that we anticipate will require validation? i can add a method for that@ 

Thanks!

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -129,10 +129,6 @@
                     "minimum": 0,
                     "exclusiveMinimum": false
                 },
-                "backoff_factor": {
-                    "type": "integer",
-                    "default": 2
-                },
                 "registrations": {
                     "type": "array",
                     "items": {
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -318,25 +318,6 @@
                 "healthcheck_timeout_seconds": {
                     "type": "number",
                     "default": 10
-                },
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -0,0 +1,344 @@
...
+                "healthcheck_uri": {
+                    "type": "string",
+                    "default": "/status"
+                },
+                "cpu_burst_add": {
+                    "type": "number",
+                    "minimum": 0.0,
+                    "exclusiveMinimum": false
+                },
+                "extra_docker_args": {
```
Rob-Johnson: we can't have this in k8s

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -0,0 +1,345 @@
...
+                    "type": "integer",
+                    "minimum": 0,
+                    "exclusiveMinimum": false
+                },
+                "max_instances": {
+                    "type": "integer",
+                    "minimum": 0,
+                    "exclusiveMinimum": false
+                },
+                "backoff_factor": {
```
chlgit: I think we could remove `backoff_factor` for now. This won't work anyway. We can add it back if we make decision for https://jira.yelpcorp.com/browse/PAASTA-15611.

PR #2389, opened 2019-08-27
add DistributedQueue class for deployd that persists queue to ZK

stug:
Just a WIP but want to show folks what I've been up to.


paasta_tools/deployd/queue.py
```diff
@@ -0,0 +1,126 @@
...
+            self.client.ensure_path(path)
+
+        self.local_state_lock = Condition()
+        self.entry_nodes = []
+        self.locked_entry_nodes = set()
+        self.update_local_state(None)
+
+    def update_local_state(self, event):
+        with self.local_state_lock:
+            print("refreshing state")
```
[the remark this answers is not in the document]
stug: this is just a WIP -- I will clean this stuff up later

paasta_tools/deployd/queue.py
```diff
@@ -0,0 +1,126 @@
...
+    def qsize(self):
+        return len(self.entry_nodes)
+
+    @property
+    def queue(self):
+        return self.entry_nodes
+
+    # TODO: should priority just be a string?  can make it length 20 for two priorities?  or even make that configurable?
+    def put(self, value, priority):
+        self._check_put_arguments(value, priority)
```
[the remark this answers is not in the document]
stug: hmm good point.  probably -- i mostly stole this from the kazoo implementation

paasta_tools/deployd/queue.py
```diff
@@ -0,0 +1,126 @@
...
+                        break
+
+                self.local_state_lock.wait()
+
+        entry_data, entry_stat = entry
+
+        try:
+            yield entry_data
+        except Exception:
+            self._release(first_available_entry_node)
```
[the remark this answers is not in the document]
stug: yeah good point

paasta_tools/deployd/queue.py
```diff
@@ -0,0 +1,126 @@
...
+        except NoNodeError:
+            self.client.delete(lock_path)
+            return None
+
+    def _consume(self, entry_node):
+        # necessary in case we lose connection at some point
+        if not self._holds_lock(entry_node):
+            return  # TODO: log?
+        with self.client.transaction() as transaction:
+            transaction.delete(f"{self.locks_path}/{entry_node}")
```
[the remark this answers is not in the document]
stug: I don't _think_ so -- it's an atomic operation and ZK is totally ordered (see https://zookeeper.apache.org/doc/r3.4.13/zookeeperInternals.html).

EvanKrall:
@stug Github won't let me request your review on this PR, but can you take a look at what I've done?

paasta_tools/deployd/queue.py
```diff
@@ -19,54 +29,54 @@ def __init__(self, client, path):
...
+
+        formatted = f"{timestamp:014.3f}"
+        assert len(formatted) == 14
 
-    @property
-    def queue(self):
-        return self.entry_nodes
+        return formatted
 
     # TODO: should priority just be a string?  can make it length 20 for two priorities?  or even make that configurable?
```
stug: Assuming you didn't mean to leave this?

paasta_tools/deployd/queue.py
```diff
@@ -107,21 +148,43 @@ def _lock_and_get_entry(self, entry_node):
...
-    def _release(self, entry_node):
+    def _release(self, entry_node: str):
         if not self._holds_lock(entry_node):
             return
         self.client.delete(f"{self.locks_path}/{entry_node}")
+
+    def get_available_service_instances(
+        self
+    ) -> Iterable[Tuple[float, ServiceInstance]]:
+        self._update_local_state(None)
```
stug: Do you actually need to do this?  Since we have a watch, I think that these should just stay up to date, right?  You might just want to put this method in the condition context manager so that stuff doesn't change while you're looking at it.

paasta_tools/deployd/queue.py
```diff
@@ -107,21 +148,43 @@ def _lock_and_get_entry(self, entry_node):
...
+            if entry_node not in self.locked_entry_nodes:
+                deadline, wait_until = self._parse_entry_node(entry_node)
+                now = time.time()
+                if wait_until <= now:
+                    yield (deadline, entry_node)
+
+    def get_unavailable_service_instances(
+        self
+    ) -> Iterable[Tuple[float, float, ServiceInstance]]:
+        self._update_local_state(None)
```
stug: Same here

paasta_tools/deployd/queue.py
```diff
@@ -76,29 +86,60 @@ def get(self):  # TODO: timeout?
...
+        return next_upcoming_wait_until
+
+    def _lock_and_get_entry(self, entry_node: str):
         try:
             lock_path = f"{self.locks_path}/{entry_node}"
+            self.locked_entry_nodes.add(entry_node)
             self.client.create(lock_path, value=self.id, ephemeral=True)
         except NodeExistsError:
+            self.locked_entry_nodes.add(entry_node)
             return None
```
stug: Do we actually need the 2nd `self.locked_entry_nodes.add(entry_node)`?    Only `self.client.create()` can raise `NodeExistsError`, in which case we'll have already added the node to `self.locked_entry_nodes`.

Also, anyway, it seems like the watcher should run shortly after this in either case.  Either we got the lock in which case the watcher will start and wait to acquire the condition's lock, or we fail to get the lock, which probably means that the watcher is currently waiting to run.  Maybe we don't even need this?

paasta_tools/deployd/queue.py
```diff
@@ -19,54 +29,54 @@ def __init__(self, client, path):
...
-            raise TypeError(f"priority must be an int (got type {type(priority)})")
-        elif priority < 0 or priority > 9999999999:
-            raise ValueError(
-                f"priority must be between 0 and 9999999999 (got {priority})"
-            )
-
     @contextmanager
-    def get(self):  # TODO: timeout?
+    def get(self, block: bool = True, timeout: float = float("inf")):
+        self._update_local_state(None)
```
stug: I'm not sure this is necessary -- we have the watcher already, and this will mean that all deployd threads will end up blocking each other waiting for the condition lock

EvanKrall: Hm. I added this so that put() followed by an immediate nonblocking get() will always work, which is a situation that only really applies in testing. You're probably right about the extra blocking not being great; I guess I can just make the tests explicitly call _update_local_state and/or use small timeouts instead of true nonblocking.

paasta_tools/deployd/queue.py
EvanKrall: good catch

paasta_tools/deployd/queue.py
EvanKrall: Yeah, I should probably make a copy inside the context manager, good call.

Like for `get`, I added `_update_local_state` for the benefit of the tests; it's nice to be able to `put()` and then immediately `get_available_service_instances` and have that include what you put. Maybe I'll just make the tests explicitly call `_update_local_state`

tests/deployd/test_queue.py
```diff
@@ -0,0 +1,142 @@
...
+        assert times[1][1] == si2
+        assert 0.019 < times[2][0] < 0.031
+        assert times[2][1] == si3
+
+    def test_return_immediately_when_blocking_on_distant_wait_until_and_available_task_comes_in(
+        self, queue
+    ):
+        """Same as above, except there's a far-off unavailable item already."""
+        queue.put(make_si(wait_until=time.time() + 100, bounce_by=time.time() + 100))
+        self.test_return_immediately_when_blocking_on_empty_queue_and_available_task_comes_in()
```
stug: oh whoa!  I didn't know you could do this.

paasta_tools/deployd/common.py
```diff
@@ -30,11 +33,12 @@
     [
         "service",
         "instance",
+        "watcher",
+        "cluster",
```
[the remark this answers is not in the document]
EvanKrall: Good question. I added it because ServiceInstance takes cluster as a parameter, and I noticed it wasn't actually doing anything with it.

Looks like I can just drop the parameter -- turns out `ServiceInstance` took `cluster` so that it could look up the appropriate priority from soa-configs: https://github.com/Yelp/paasta/pull/2343/files#diff-0d5c7263f04c8928f403a53f529860cfL74

I think I may also just unify BaseServiceInstance / ServiceInstance; I think the only reason ServiceInstance existed was to have an __init__ that called `get_priority`.

paasta_tools/deployd/queue.py
EvanKrall: I forget which test I was working on when I added this, but I think it's reasonable: if we get a NodeExistsError here, we know that this entry_node is already locked.

And since `_lock_and_get_entry` is run while `self.local_state_condition` is held, I think there shouldn't be any weird race conditions here.

paasta_tools/deployd/metrics.py
```diff
@@ -71,20 +71,16 @@ def __init__(
...
-        )
-
         currently_available_instances = tuple(
-            self.queue.available_service_instances.queue
+            self.queue.get_available_service_instances()
         )
         currently_unavailable_instances = tuple(
-            self.queue.unavailable_service_instances.queue
+            self.queue.get_unavailable_service_instances()
         )
```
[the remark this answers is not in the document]
EvanKrall: get_available_service_instances also only returns (deadline, si) whereas get_unavailable... returns (wait_until, deadline, si).

But I've refactored the implementation of these two methods so they call a common one.

tests/deployd/test_queue.py
EvanKrall: Yeah, hopefully not too clever.

EvanKrall:
@stug I've made some pretty significant changes, particularly in b502088 and 14b5a8d, can you take another look?

paasta_tools/deployd/queue.py
```diff
@@ -107,14 +123,25 @@ def get(self, block: bool = True, timeout: float = float("inf")):
...
+            "watcher": "unknown",
+            "bounce_by": now,
+            "wait_until": now,
+            "enqueue_time": now,
+            "bounce_start_time": now,
+            "failures": 0,
+        }
+        si_dict = json.loads(entry_data.decode("utf-8"))
+        merged = {**defaults, **si_dict}
+        return ServiceInstance(**merged)  # type: ignore
```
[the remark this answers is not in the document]
EvanKrall: I was iterating on itests by keeping the docker-compose environment up and just repeatedly running itests; I think some znodes got left behind by one run and caused the next run to crash with missing arguments to ServiceInstance.

I figure we'll probably want to change the ServiceInstance interface at some point, so we may as well have code for handling missing values. Some values don't make sense to be missing (like service/instance).

paasta_tools/deployd/queue.py
```diff
@@ -146,47 +173,50 @@ def _lock_and_get_entry(self, entry_node: str):
...
+    ) -> Iterable[Tuple[float, float, ServiceInstance]]:
         results = []
         with self.local_state_condition:
             for entry_node in self.entry_nodes:
                 if entry_node not in self.locked_entry_nodes:
                     deadline, wait_until = self._parse_entry_node(entry_node)
-                    now = time.time()
-                    if wait_until <= now:
-                        results.append((deadline, entry_node))
+                    data, _ = self.client.get(f"{self.entries_path}/{entry_node}")
```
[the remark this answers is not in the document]
EvanKrall: Hm, you're right, we don't actually use the ServiceInstance right now in the metrics; we're just counting numbers.

Honestly I just did it to match what the in-memory queue was providing, and didn't think about what metrics.py was actually using. Maybe I can make a parameter for whether to actually include the ServiceInstance?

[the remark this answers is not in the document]
stug: Yeah I am not sure how much it matter, but I guess I've heard a couple things recently about issues with load on ZK so I figured I'd point it out.  Seems like it's possible that this would slow things down.

I can imagine that in the future we could build some nice monitoring tools that actually fetch the ServiceInstances, but maybe those things could take the list of queue nodes and do their own fetching?

PR #2393, opened 2019-08-29
Upgrade service-configuraiton-lib to get less cruft and the new inotify caching library

solarkennedy:


solarkennedy:
No, it is still in use by the paasta firewall code.

We had pynotify already anyway in extra requirements, just unpinned.

PR #2396, opened 2019-09-03
Add Flink metadata to paasta_tools API

Bronek:
This will expose metadata in API endpoint, hence allowing us to extend `paasta status` to show more useful information for Flink instances.

A separate pull request to use this metadata in `paasta status` yet to come.

### DIFF
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -161,7 +161,7 @@ def flink_instance_status(
     status: Optional[Mapping[str, Any]] = None
     client = settings.kubernetes_client
     if client is not None:
-        status = flink_tools.get_flink_config(
+        status = flink_tools.get_flink_status(
             kube_client=client, service=service, instance=instance
         )
     return status
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -167,6 +167,18 @@ def flink_instance_status(
     return status
 
 
+def flink_instance_metadata(
+    instance_status: Mapping[str, Any], service: str, instance: str, verbose: int
+) -> Optional[Mapping[str, Any]]:
+    metadata: Optional[Mapping[str, Any]] = None
+    client = settings.kubernetes_client
+    if client is not None:
+        metadata = flink_tools.get_flink_metadata(
+            kube_client=client, service=service, instance=instance
+        )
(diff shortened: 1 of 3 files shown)


paasta_tools/api/views/instance.py
```diff
@@ -726,8 +738,13 @@ def instance_status(request):
             )
         elif instance_type == "flink":
             status = flink_instance_status(instance_status, service, instance, verbose)
-            if status is not None:
+            metadata = flink_instance_metadata(instance_status, service, instance, verbose)
+            if status is not None and metadata is not None:
```
keymone: this can be simplified

    instance_status["flink"] = {}
    if status:
        instance_status["flink"]["status"] = status
    if metadata:
        instance_status["flink"]["metadata"] = metadata

paasta_tools/flink_tools.py
```diff
@@ -183,6 +183,22 @@ def get_flink_config(
...
+        co = kube_client.custom.get_namespaced_custom_object(
+            **flink_custom_object_id(service, instance)
+        )
+        metadata = co.get("metadata")
+        return metadata
+    except ApiException as e:
+        if e.status == 404:
+            return None
+        else:
+            raise
```
keymone: raise the original exception here, we don't want to discard the reason this failed

Bronek: but that's what `raise` with no argument does, inside `except` clause. Did I miss something?

keymone: you're right! TIL

paasta_tools/api/views/instance.py
```diff
@@ -726,10 +738,14 @@ def instance_status(request):
...
             status = flink_instance_status(instance_status, service, instance, verbose)
+            metadata = flink_instance_metadata(
+                instance_status, service, instance, verbose
+            )
+            instance_status["flink"] = {}
             if status is not None:
-                instance_status["flink"] = {"status": status}
-            else:
-                instance_status["flink"] = {}
+                instance_status["flink"]["status"] = status
```
Rob-Johnson: are you sure you want to leave this empty when status is None, rather than leaving the response as an empty dictionary?

Bronek: yes, that preserves the old behaviour, I think.

PR #2398, opened 2019-09-04
Make is_old_task_missing_healthchecks deal with tz-aware datetimes

EvanKrall:
A much shorter patch would have been to just change `datetime.datetime.now()` to `datetime.datetime.utcnow()`, but I think I'm of the opinion that naive datetimes are almost always an antipattern.


EvanKrall:
I think I'd also like to make a PR to marathon-python that makes it return aware datetime objects instead of naive ones, but that has a larger potential blast radius.

PR #2400, opened 2019-09-04
Bumped to the latest version of the k8s client

solarkennedy:
Not sure how to fully test this.
metastatus tests were trying to connect kubestage, so I mocked them out.
Upgrading for good measure, but also to remove a YAML loading warning.


vkhromov:
As for the testing, I usually install a manually built version of `paasta-tools` at `kubestage` and wait a couple of hours.

PR #2401, opened 2019-09-04
Change mesos api read timeout from 1s to 2s

kawaiwanyelp:
### Description
We are hammering the Mesos API and sometimes it times out, preventing things like maintenance clean ups from happening. To alleviate this problem, I have bumped the default read timeout for the Mesos API from 1s to 2s.

### Testing done
`make test`


kawaiwanyelp:
Rebased and pushed again to trigger a new Travis build.

PR #2404, opened 2019-09-05
Show config SHA from Flink metadata

Bronek:
Show config SHA from metadata, for Flink clusters only

It is conceivable that in the future we could extend this to show more data, or handle metadata for other types of clusters (if API is extended as well). At this moment we only have demand for config SHA in Flink,  though.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -246,8 +246,10 @@ def paasta_status_on_api_endpoint(
             cluster, service, instance, output, status.adhoc, verbose
         )
     elif status.flink is not None:
+        flink_status = status.flink.get("status")
+        flink_metadata = status.flink.get("metadata")
         return print_flink_status(
-            cluster, service, instance, output, status.flink.get("status"), verbose
+            cluster, service, instance, output, flink_status, flink_metadata, verbose
         )
     elif status.chronos is not None:
         return print_chronos_status(output, status.chronos.output)
@@ -713,12 +715,36 @@ def status_kubernetes_job_human(
         )
 
 
+def _strip_config_prefix(config_sha: str) -> str:
+    prefix = "config"
+    prefix_len = len(prefix)
+    if len(config_sha) > prefix_len and config_sha[:prefix_len] == prefix:
+        return config_sha[prefix_len:]
+    return config_sha
+
+
 def print_flink_status(
-    cluster: str, service: str, instance: str, output: List[str], status, verbose: int
+    cluster: str,
+    service: str,
+    instance: str,
+    output: List[str],
+    status,
+    metadata,
+    verbose: int,
 ) -> int:
     if status is None:
         output.append(PaastaColors.red("    Flink cluster is not available yet"))
         return 1
+
+    config_sha = None
+    if metadata is not None and metadata.labels is not None:
+        config_sha = metadata.labels.get("yelp.com/paasta_config_sha")
+        if config_sha is not None:
+            config_sha = _strip_config_prefix(config_sha)
+
+    # Since metadata should be available no matter the state, we show it before state
+    if config_sha is not None:
+        output.append(f"    Config SHA: {config_sha}")
 
     if status.state != "running":
         output.append(
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -22,6 +22,7 @@ from datetime import datetime
 from datetime import timedelta
 from distutils.util import strtobool
 from itertools import groupby
+from typing import Any
 from typing import Callable
 from typing import DefaultDict
 from typing import Dict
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -713,13 +715,37 @@ def status_kubernetes_job_human(
...
+    output: List[str],
+    status,
+    metadata,
+    verbose: int,
 ) -> int:
     if status is None:
         output.append(PaastaColors.red("    Flink cluster is not available yet"))
         return 1
 
+    config_sha = None
```
vkhromov:
```
if metadata and metadata.labels:
    config_sha = metadata.labels.get("yelp.com/paasta_config_sha")
    if config_sha:
        if config_sha.startswith("config"):
            config_sha = config_sha[5:]
        output.append(f"    Config SHA: {config_sha}")
```
?

paasta_tools/cli/cmds/status.py
```diff
@@ -246,8 +246,10 @@ def paasta_status_on_api_endpoint(
             cluster, service, instance, output, status.adhoc, verbose
         )
     elif status.flink is not None:
+        flink_status = status.flink.get("status")
+        flink_metadata = status.flink.get("metadata")
         return print_flink_status(
-            cluster, service, instance, output, status.flink.get("status"), verbose
```
vkhromov: mb just pass `status.flink` into `print_flink_status()`, since the status to print is now more than just `status.flink.status`?

paasta_tools/cli/cmds/status.py
Bronek: yes, that's one way to simplify this. I am also considering to always show config sha, and when not known then show "(unknown)"

paasta_tools/cli/cmds/status.py
Bronek: I like this suggestion, thank you

paasta_tools/cli/cmds/status.py
```diff
@@ -713,13 +715,37 @@ def status_kubernetes_job_human(
...
+    return config_sha
+
+
 def print_flink_status(
-    cluster: str, service: str, instance: str, output: List[str], status, verbose: int
+    cluster: str,
+    service: str,
+    instance: str,
+    output: List[str],
+    status,
```
Rob-Johnson: can you add the type signature on these two arguments too?

paasta_tools/cli/cmds/status.py
Rob-Johnson: yeah, I think I'd rather show 'unknown'. It's a situation that shouldn't really ever happen. If anything, it's worth outputting something like `config: ERROR - Could not determine config sha`.

Bronek: well, that error is not really contributing much. I will just show "(unknown)"

Rob-Johnson: It'll signal to the user that there is something wrong. It should be a rare case, but I'd rather do something to let the user know that this isn't an expected state. You could even just error out at this point

Bronek: The small issue I have with letting the error out is that you could hypothetically run `paasta status` against a cluster running an old version of paasta-tools, which does not yet set the metadata. Also, if the current version of paasta-tools is installed but there is a bug somewhere else (e.g. preventing the config_sha label from being set in the metadata), I think being able to show the rest of the state will aid troubleshooting.

Rob-Johnson: > you could hypothetically run paasta status against a cluster running an old version of paasta-tools, which does not yet set the metadata

I don't think you need to worry about this: we have control over the versions of paasta tools we run everywhere.

>  if the current version of paasta-tools is installed but there is a bug somewhere else (e.g. preventing the config_sha label from being set in the metadata), I think being able to show the rest of the state will aid troubleshooting.

If that label doesn't exist, then the cluster isn't going to be functional anyway (the operator depends on it). I think it'd be a pretty exceptional circumstance and we should treat is as such

Bronek: ok, I just removed the defensive checks so if this errors out, the user will get all the Python stack he needs to learn about the problem.

paasta_tools/cli/cmds/status.py
```diff
@@ -714,12 +715,27 @@ def status_kubernetes_job_human(
...
 ) -> int:
+    status = flink.get("status")
     if status is None:
         output.append(PaastaColors.red("    Flink cluster is not available yet"))
         return 1
 
+    # Since metadata should be available no matter the state, we show it first. If this errors out
+    # then we cannot really do much to recover, because cluster is not in usable state anyway
+    metadata = flink.get("metadata")
+    config_sha = metadata.labels.get("yelp.com/paasta_config_sha")
```
Rob-Johnson: I think you can still provide hints here, even if you're going to throw the exception:

```
 config_sha = metadata.labels.get("yelp.com/paasta_config_sha")
if config_sha is None:
    raise ValueError(f"expected config sha on Flink CR. ObjectMeta {metadata}")
if config_sha.startswith("config"):
    config_sha = config_sha[6:]
```

PR #2406, opened 2019-09-05
Make tests produce more output

vkhromov:
Travis fails a build/test run if there is no output for more than 10
minutes from it.  In the case of slow network and many packages to
process, `apt-get update/install` could easily take more than 10
minutes.  Since we redirect the output from `apt-get` in our test runs
into `/dev/null`, Travis could terminate out runs in such case.

On the other hand, having the output from `apt-get` run may be useful.

So, let stop redirecting the output from `apt-get` into `/dev/null`


solarkennedy:
you can discard this if you want

PR #2409, opened 2019-09-05
Use common docker images to greatly speed up itest/builds

solarkennedy:
I'm not sure if this is actually a good idea or not. Here is the big change:

* Uses docker hub automated builds instead of having travis push
* Makes our biggest images use the common images

This will mean we need to do a kind of "two phase" push if we change the base image. I think that is ok due to how rare that is.

It also means things are a bit indirected, because you have to look in the base image to see what is inside. Eh?

It also means that the docker pulls are longer at first, but hopefully shorter in the long term than having to build all these images over and over via apt-get.

The end goal is to never (or have very few) apt-gets in our builds at all.

### DIFF
--- a/yelp_package/dockerfiles/paastatools_xenial_container/Dockerfile
+++ b/yelp_package/dockerfiles/paastatools_xenial_container/Dockerfile
@@ -1,0 +1,73 @@
+# Copyright 2015-2016 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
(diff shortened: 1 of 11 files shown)


yelp_package/dockerfiles/paastatools_xenial_container/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+        sudo \
+        lsb-release \
+        docker-ce=17.03.2~ce-0~ubuntu-xenial \
+        mesos=1.7.2-2.0.1 \
+        chronos=2.5.0-yelp32-1.ubuntu1604 \
+        marathon=1.4.11-1.0.676.ubuntu1604 \
+        rsyslog \
+        python2.7-dev \
+        virtualenv \
+        zookeeper
```
Bronek: I think we can add here `&& rm -rf /var/lib/apt/lists/*` , since the only container which runs `apt-get install` is `...gitremote/Dockerfile`. We might just as well add openssh-server to this image and get rid of `apt-get install` in the other one entirely. I am not sure how much space does apt-get cache need but there is a chance that this will make this image smaller, despite having one more package more installed.

Bronek:
Given that we this PR we are about to start using yelp/pastatools_xenial_container  for CI, which we want to be reproducible, I'd rather make use of docker tags in all the Dockerfiles which pull this container.

solarkennedy:
I'm going to discard this for now, the tests are passing with the new local mirror without this complexity.

PR #2413, opened 2019-09-06
apply min/max instance clamps only when reading from zookeeper, not when writing

drmorr0:
The last time I pushed this we tried to clamp in s-m-j, but @EvanKrall pointed out that we already clamp when we read values from ZK, so there's no need to try to do this logic again in s-m-j.  So this time we're _only_ removing the code that clamps when writing to ZK, which should solve the problems where we sometimes bottom out to min_instances even if there's load on the cluster.

### DIFF
--- a/paasta_itests/paasta_deployd.feature
+++ b/paasta_itests/paasta_deployd.feature
@@ -74,7 +74,9 @@ Feature: paasta-deployd deploys apps
       And we set the instance count in zookeeper for service "test-service" instance "main" to 3
      Then we should see "test-service.main" listed in marathon after 45 seconds
      Then we should see the number of instances become 3
-     When we set the instance count in zookeeper for service "test-service" instance "main" to 4
+     # we are testing that even if the values written to ZK are outside the min/max bounds, the instances still get
+     # clamped to the right values.
+     When we set the instance count in zookeeper for service "test-service" instance "main" to 5
      Then we should see the number of instances become 4
 
   Scenario: deployd will scale down an app if the instance count changes in zk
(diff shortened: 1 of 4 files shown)


paasta_itests/paasta_deployd.feature
```diff
@@ -74,7 +74,7 @@ Feature: paasta-deployd deploys apps
       And we set the instance count in zookeeper for service "test-service" instance "main" to 3
      Then we should see "test-service.main" listed in marathon after 45 seconds
      Then we should see the number of instances become 3
-     When we set the instance count in zookeeper for service "test-service" instance "main" to 4
+     When we set the instance count in zookeeper for service "test-service" instance "main" to 5
```
EvanKrall: Can you add a comment explaining why it only goes to 4?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -725,6 +716,13 @@ def autoscale_marathon_instance(
                         ),
                         level="event",
                     )
+            else:
+                write_to_log(
```
solarkennedy: I still don't think we should log this.
We don't need to continuously log for every service/instance that the autoscaler isn't doing anything, even at the "debug" level for paasta logs

EvanKrall: I really don't think it's very costly, and would be helpful when reading logs to debug autoscaler issues.

paasta_itests/paasta_deployd.feature
drmorr0: Fixed

PR #2414, opened 2019-09-06
Make service autoscaler consider tasks healthy if they are in smartstack, even if Marathon thinks they are unhealthy.

EvanKrall:
Based on https://github.com/Yelp/paasta/pull/2398 which I plan to merge shortly, so ignore the changes to marathon_tools.py

I think the docstring I added to filter_autoscaling_tasks justifies this change pretty well.


EvanKrall:
@solarkennedy @mattmb @stug bump, anybody want to review this?

PR #2415, opened 2019-09-06
[WIP] list pods for verbose option of paasta status

chlgit:
I need to add test cases. The output looks like:

service: yelp-main
cluster: kubestage
    instance: main_uswest1
    Git sha:    1c59071f (desired)
    State:      Bouncing (crossover) - Desired state: Started
    Kubernetes:   Healthy - up with (3/3) instances. Status: Running
      App created: 2019-07-12 13:31:48 (a month ago)
      Pods:
        Pod ID                                    Host deployed to                            Deployed at what localtime       Health
        yelp-main-main--uswest1-6784fff8d8-k47kk  ip-10-40-10-254.us-west-1.compute.internal  2019-09-05T03:16 (a day ago)     Healthy
        yelp-main-main--uswest1-6bf68cdd54-td74g  ip-10-40-17-254.us-west-1.compute.internal  2019-09-04T07:02 (2 days ago)    Healthy
        yelp-main-main--uswest1-86958bcf6f-4ttfz  Unknown                                     2019-09-06T04:20 (11 hours ago)  Unhealthy
        yelp-main-main--uswest1-86958bcf6f-7fh96  ip-10-40-11-62.us-west-1.compute.internal   2019-09-06T04:20 (11 hours ago)  Healthy
        yelp-main-main--uswest1-86958bcf6f-q95xj  Unknown                                     2019-09-06T04:20 (11 hours ago)  Unhealthy
        yelp-main-main--uswest1-ddcd67644-vb249   ip-10-40-17-254.us-west-1.compute.internal  2019-09-05T04:32 (a day ago)     Healthy
        yelp-main-main--uswest1-dfb48d57d-pspd6   ip-10-40-17-91.us-west-1.compute.internal   2019-09-06T01:25 (14 hours ago)  Unhealthy


paasta_tools/api/views/instance.py
```diff
@@ -207,8 +207,18 @@ def kubernetes_job_status(
 ) -> None:
     app_id = job_config.get_sanitised_deployment_name()
     kstatus["app_id"] = app_id
+    kstatus["pods"] = []
     if verbose > 0:
-        kstatus["slaves"] = [pod.spec.node_name for pod in pod_list]
+        for pod in pod_list:
+            kstatus["pods"].append(
+                {
+                    "name": pod.metadata.name,
```
vkhromov: we probably need the namespace as well?

qui:
What about grouping the pods by replica set? replica sets roughly correspond to Marathon apps. That way, you can easily see if the newest replica set is coming up healthy when you make a code or config change.

Also, we should also eventually add Smartstack information for kubernetes.

chlgit:
Yeah, I expect there would be three PRs for this verbose option, one for pods, one for replica, and one for smartstack. My objective would be something like:

service: yelp-main
  [Replica Set Section]
  [Pods Section]
  [Smartstack Section]

paasta_tools/api/views/instance.py
chlgit: Right. I forgot that not everything is running under paasta namespace. Added it.

paasta_tools/cli/cmds/status.py
```diff
@@ -602,6 +602,36 @@ def format_marathon_task_table(tasks):
...
+    rows = [("Pod ID", "Host deployed to", "Deployed at what localtime", "Health")]
+    for pod in pods:
+        local_deployed_datetime = datetime_from_utc_to_local(
+            datetime.fromtimestamp(pod.deployed_timestamp)
+        )
+        hostname = f"{pod.host}" if pod.host is not None else "Unknown"
+
+        if pod.phase is None:
+            health_check_status = PaastaColors.grey("N/A")
+        elif pod.phase == "Running":
```
EmanekaT: I'm not sure if this is a good indication of healthness of pods. Running does not necessarily mean the pod is healthy . It could transition to failed state after running for a few seconds. And pending does not mean it's unhealthy. Perhaps we should use liveness probe instead.

paasta_tools/cli/cmds/status.py
```diff
@@ -831,6 +861,20 @@ def print_kubernetes_status(
             )
         )
     )
+    if kubernetes_status.create_timestamp:
```
EmanekaT: It seems you are assuming create_timestamp exists, so this if condition is unnecessary?  I'm not sure if create_timestamp is always available on k8s api. So I think it's safer to make it optional here:     kstatus["create_timestamp"] = app.metadata.creation_timestamp.timestamp()

EmanekaT:
Also, since k8s is different from marathon, we perhaps should reconsider what is the best paasta status api.

paasta_tools/api/api_docs/swagger.json
```diff
@@ -1113,9 +1113,25 @@
...
+                    "type": "array",
+                    "description": "Pods associated to this app",
+                    "items": {
+                        "$ref": "#/definitions/KubernetesPod"
+                    }
+                },
                 "error_message": {
                     "type": "string",
-                    "description": "Error message when a marathon job ID cannot be found"
+                    "description": "Error message when a kubernetes app ID cannot be found"
```
EmanekaT: app ID?

paasta_tools/api/api_docs/swagger.json
```diff
@@ -1124,6 +1140,24 @@
...
+                    "type": "number",
+                    "format": "float",
+                    "description": "Time at which the pod was deployed"
+                },
+                "phase": {
+                    "type": "string",
+                    "description": "The status of the pod"
+                }
+            }
+        },
```
EmanekaT: Add a container section with logs?

EmanekaT:
> What about grouping the pods by replica set? replica sets roughly correspond to Marathon apps. That way, you can easily see if the newest replica set is coming up healthy when you make a code or config change.
> 
> Also, we should also eventually add Smartstack information for kubernetes.

We probably want to do something with https://jira.yelpcorp.com/browse/PAASTA-15782 before that.

paasta_tools/cli/cmds/status.py
chlgit: The thing I worried about here is that if we make new paasta release, there's no guarantee that server side would be upgraded first. If the client side upgrades first without this condition, it would throw errors because of NoneType. At least from the my testing, create_timestamp is always available on k8s api. Should we make this optional? It would mask the errors imo.

paasta_tools/api/api_docs/swagger.json
chlgit: I just follow the name of the property 'app_id' at line 1075. Do you have better name?

paasta_tools/cli/cmds/status.py
EmanekaT: @chlgit good point. I guess you can make another PR to remove this line after it's release. Leave it as it is for now.

paasta_tools/api/api_docs/swagger.json
EmanekaT: What about "kubernetes object" ?

paasta_tools/cli/cmds/status.py
chlgit: My personally option is that pending should be considered as unhealthy. I think if liveness check fails, the phase would transit to failed state, so the phase is actually taking liveness probe into account. Besides, the liveness_probe doesn't contain the result of liveness check:
                          'liveness_probe': {'_exec': None,
                                             'failure_threshold': 30,
                                             'http_get': {'host': None,
                                                          'http_headers': None,
                                                          'path': '/status',
                                                          'port': 8888,
                                                          'scheme': 'HTTP'},
                                             'initial_delay_seconds': 60,
                                             'period_seconds': 10,
                                             'success_threshold': 1,
                                             'tcp_socket': None,
                                             'timeout_seconds': 10}, 

I think eventually the healthness should be determined by the smartstack (I plan to add smartstack section through another PR).

EmanekaT: @chlgit Why can't just make pending pending? And unhealthy if it's failed or is running but unhealthy? I would assume pending != unhealthy if I were the user.

EmanekaT: @chlgit  Or we can even have both phase and health_check_status. health_check_status is unknown unless we can really make sure we know it.

paasta_tools/api/api_docs/swagger.json
chlgit: Yeah, a container section should belong to `-vv` or `-vvv` option imo. I think we will definitely add it later. Just don't want to do everything in this PR. Otherwise, this gonna be very huge.

PR #2416, opened 2019-09-09
Try to emulate how mesos runs the paasta_docker_wrapper

solarkennedy:


solarkennedy:
cc @nemacysts

paasta_tools/cli/cmds/local_run.py
```diff
@@ -826,7 +827,11 @@ def run_docker_container(
     if interactive or not simulate_healthcheck:
         # NOTE: This immediately replaces us with the docker run cmd. Docker
         # run knows how to clean up the running container in this situation.
-        execlpe("paasta_docker_wrapper", *docker_run_cmd, merged_env)
+        wrapper_path = shutil.which("paasta_docker_wrapper")
+        # To properly simulate mesos, we pop the PATH, which is not available to
+        # The executor
+        merged_env.pop("PATH")
```
EvanKrall: Wouldn't the more correct thing be to just get rid of `merged_env` entirely and only use `environment`?

paasta_tools/cli/cmds/local_run.py
```diff
@@ -826,7 +827,11 @@ def run_docker_container(
     if interactive or not simulate_healthcheck:
         # NOTE: This immediately replaces us with the docker run cmd. Docker
         # run knows how to clean up the running container in this situation.
-        execlpe("paasta_docker_wrapper", *docker_run_cmd, merged_env)
+        wrapper_path = shutil.which("paasta_docker_wrapper")
```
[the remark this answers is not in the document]
solarkennedy: There are lots of cases like that, but Tron gives you a PATH

paasta_tools/cli/cmds/local_run.py
solarkennedy: I thought about that, but then remembers that we need some `environment` variables because that is how we do secrets. 

If you want I can make a function that strips out everything from the environment *except* PAASTA_ vars, and then use that instead? (instead of explicility popping PATH)

paasta_tools/docker_wrapper.py
```diff
@@ -20,6 +21,15 @@
 import socket
 import sys
 
+
+if "PATH" not in os.environ:
+    # This command is sometimes executed in a sanitized environment
```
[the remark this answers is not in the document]
solarkennedy: afaik no, not without a custom executor.

paasta_tools/cli/cmds/local_run.py
EvanKrall: Doesn't paasta local-run set the PAASTA_ variables?

solarkennedy: I think I better understand what you were getting at here. Let me retry.

PR #2417, opened 2019-09-10
ignore incomplete jobs

Bronek:
A cluster in startup phase may report incomplete job data. Trying to show these only creates problems, so let's just skip over.


vkhromov:
I started wondering, maybe we should update the operator instead?
Probably it should add only correct records into the status.

Bronek:
> I started wondering, maybe we should update the operator instead?
> Probably it should add only correct records into the status.

this reminds me that we guarantee in the operator that "jid" will be always set. Updated the last commit not to check it, so if it is null we will have justified error.

PR #2427, opened 2019-09-16
Store flink dashboard url in metadata:annotations

Bronek:
This change implies that we will perhaps want to apply similar design to other K8s resources, because of the addition of a k8s metadata label.

The idea is to store a `yelp.com/dashboard_url` in `metadata:labels` ; this is created inside `setup_kubernetes_cr.py` ; the actual URL string-building function is `get_dashboard_url` and is likely to change. At this moment the assumption is that we want to maintain sanitized name at its current location and also that we will not bother defining Flink dashboard url in configuration for **every** cluster, so this function is capable of building a "default" one, same as the old function 'flink_tools.get_dashboard_url' did.

Yet to update integration tests for this one.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -49,7 +49,6 @@ from paasta_tools.cli.utils import list_deploy_groups
 from paasta_tools.cli.utils import NoSuchService
 from paasta_tools.cli.utils import validate_service_name
 from paasta_tools.flink_tools import FlinkDeploymentConfig
-from paasta_tools.flink_tools import get_dashboard_url
 from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
 from paasta_tools.kubernetes_tools import KubernetesDeployStatus
 from paasta_tools.marathon_serviceinit import bouncing_status_human
@@ -785,9 +784,8 @@ def print_flink_status(
         output.append(f"    No other information available in non-running state")
         return 0
 
-    dashboard_url = get_dashboard_url(
-        cluster=cluster, service=service, instance=instance
-    )
+    dashboard_url = metadata.annotations.get("yelp.com/flink_dashboard_url")
+    dashboard_url += metadata.name
     if verbose:
         output.append(
             f"    Flink version: {status.config['flink-version']} {status.config['flink-revision']}"
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -784,7 +784,7 @@ def print_flink_status(
         output.append(f"    No other information available in non-running state")
         return 0
 
-    dashboard_url = metadata.annotations.get("yelp.com/flink_dashboard_url")
+    dashboard_url = metadata.annotations.get("yelp.com/dashboard_url")
     dashboard_url += metadata.name
     if verbose:
         output.append(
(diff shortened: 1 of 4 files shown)


paasta_tools/setup_kubernetes_cr.py
```diff
@@ -243,6 +263,9 @@ def format_custom_resource(
         },
         "spec": instance_config,
     }
+    dashboard_url = get_dashboard_url(kind, service, instance, cluster)
+    if dashboard_url:
+        resource["metadata"]["labels"]["yelp.com/dashboard_url"] = dashboard_url
```
Rob-Johnson: yeah, I'm not sure we should change the labels here: we run the risk of masking the 'real' labels, and the users won't be sure of which are from kubernetes and which are from paasta.

instead, I'd just append to the metadata dictionary

Bronek:
I am thinking to get rid of "sanitized name" suffix in `yelp.com/dashboard_url` and let `status.print_flink_status` instead concatenate this by itself, from `metadata:name`. Good or bad idea?

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -216,6 +219,22 @@ def setup_custom_resources(
     return succeded
 
 
+def get_dashboard_url(
+    kind: str, service: str, instance: str, cluster: str
+) -> Optional[str]:
+    if kind == "Flink":
```
vkhromov: is the kind actually capitalized?

vkhromov:
@Bronek, can you please describe why you need this change?

Bronek:
> @Bronek, can you please describe why you need this change?

The idea is to provide an alternative to hardcoded generation of URL, so we can access Flink dashboards behind a network proxy etc.

paasta_tools/setup_kubernetes_cr.py
Bronek: Yes, I checked at the source. However you do have a point - this should be case-insensitive, I will fix.

paasta_tools/setup_kubernetes_cr.py
vkhromov: you cannot add arbitrary info into metadata itself, it's just a structure defined inside k8s: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L110
Instead os that please use `Annotations` map which is there for arbitrary data: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go#L238

Bronek: Agree, I will store this in annotations. Thank you!

paasta_tools/cli/cmds/status.py
```diff
@@ -785,9 +784,8 @@ def print_flink_status(
         output.append(f"    No other information available in non-running state")
         return 0
 
-    dashboard_url = get_dashboard_url(
-        cluster=cluster, service=service, instance=instance
-    )
+    dashboard_url = metadata.annotations.get("yelp.com/dashboard_url")
```
Rob-Johnson: are you missing a `/` separator here?

Bronek: No - the dashboard_url is guaranteed to end with `/`, see `setup_kubernetes_cr.get_dashboard_url` where this field is set.

PR #2428, opened 2019-09-16
Fix more liberal healthchecks

EvanKrall:
#2414 introduced a bug:
https://github.com/Yelp/paasta/pull/2414/files#diff-699ef1f9441fd1dab1f660ff54347041R922

The previous iteration of this code checked `task.id.startswith(someprefix)` for every task in every app, but task ids always start with the app id so #2414 only looked at the app id.

However, I forgot that Marathon prefixes app ids with `/`, leading this code to always find zero tasks.

This change reintroduces the changes from #2414 (which were reverted in https://github.com/Yelp/paasta/commit/e49a45f33dc7e457ef1411a881ba9e29a9750500) and strips leading slashes from the app id before checking for the prefix.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -22,7 +22,9 @@ from contextlib import contextmanager
 from datetime import datetime
 from math import ceil
 from math import floor
+from typing import Dict
 from typing import Iterator
+from typing import List
 from typing import Mapping
 from typing import Optional
 from typing import Sequence
@@ -41,10 +43,12 @@ from marathon.models.app import MarathonTask
 from paasta_tools.autoscaling.forecasting import get_forecast_policy
 from paasta_tools.autoscaling.utils import get_autoscaling_component
 from paasta_tools.autoscaling.utils import register_autoscaling_component
+from paasta_tools.bounce_lib import filter_tasks_in_smartstack
 from paasta_tools.bounce_lib import LockHeldException
 from paasta_tools.bounce_lib import LockTimeout
 from paasta_tools.bounce_lib import ZK_LOCK_CONNECT_TIMEOUT_S
 from paasta_tools.long_running_service_tools import compose_autoscaling_zookeeper_root
+from paasta_tools.long_running_service_tools import load_service_namespace_config
 from paasta_tools.long_running_service_tools import set_instances_for_marathon_service
 from paasta_tools.long_running_service_tools import ZK_PAUSE_AUTOSCALE_PATH
 from paasta_tools.marathon_tools import AutoscalingParamsDict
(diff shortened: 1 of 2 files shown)


paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -887,32 +893,67 @@ def filter_autoscaling_tasks(
...
-                marathon_tasks[task.id] = task
+                healthy_marathon_tasks[task.id] = task
+
+    service_namespace_config = load_service_namespace_config(
+        service=config.service, namespace=config.get_nerve_namespace()
+    )
+    if service_namespace_config.is_in_smartstack():
+
+        for task in filter_tasks_in_smartstack(
+            [task for tasks in relevant_tasks_by_app.values() for task in tasks],
```
solarkennedy: If you want to put one more task in here, do the kwarg

```
    tasks=[tasks for ....
```

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -887,32 +893,67 @@ def filter_autoscaling_tasks(
...
+            service=config.service,
+            nerve_ns=config.get_nerve_namespace(),
+            system_paasta_config=system_paasta_config,
+            max_hosts_to_query=20,
+            haproxy_min_fraction_up=0.01,  # Be very liberal. See docstring above for rationale.
+        ):
+            healthy_marathon_tasks[task.id] = task
```
solarkennedy: Is this actually *filtering*?

We start off with `healthy_marathon_tasks` from marathon, but how does this smartstack filter really filter anything out?

won't `healthy_marathon_tasks` just still have whatever marathon had?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -887,32 +893,67 @@ def filter_autoscaling_tasks(
...
             ):
-                marathon_tasks[task.id] = task
+                healthy_marathon_tasks[task.id] = task
+
+    service_namespace_config = load_service_namespace_config(
+        service=config.service, namespace=config.get_nerve_namespace()
+    )
+    if service_namespace_config.is_in_smartstack():
+
+        for task in filter_tasks_in_smartstack(
```
solarkennedy: I'm worried about the performance impact of this. It is all async down there, but.. this could end up being a *lot* of csv and haproxy calls, right?

paasta_tools/autoscaling/autoscaling_service_lib.py
EvanKrall: Correct, this outer function is always adding things, that's the whole point of this PR.

But the inner function (filter_tasks_in_smartstack) takes a list of tasks and returns a subset of them, so the inner function is filtering.

paasta_tools/autoscaling/autoscaling_service_lib.py
EvanKrall: It should be 20 calls per service/instance, just like setup_marathon_job. And we reduced the CSV-parsing burden by orders of magnitude in #2371.

paasta_tools/autoscaling/autoscaling_service_lib.py
EvanKrall: smurfy!

PR #2432, opened 2019-09-19
Make it easier to understand http draining exceptions

solarkennedy:
I have two goals:
* want to make it easy to see what url we were trying to hit and make it really obvious what the code was
* I want paasta logs lines to always just be one line, I don't think we need the whole traceback in there (if you think we need the traceback, I think I would prefer to just `log` it.


solarkennedy:
@EvanKrall got it.

PR #2433, opened 2019-09-19
Small tweaks to rollback english

solarkennedy:


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -695,10 +695,16 @@ def valid_transitions(self) -> Iterator[state_machine.TransitionDefinition]:
     def disable_auto_rollbacks(self):
         self.cancel_auto_rollback_countdown()
         self.auto_rollback = False
+        self.update_slack_status(
+            f"Automatic rollback disabled for this deploy. To disable this permanently for this step, `edit deploy.yaml`"
```
EvanKrall: > To disable this permanently for this step, edit `deploy.yaml`
instead of 
> To disable this permanently for this step, `edit deploy.yaml`

Also maybe worth mentioning what config key to use?

paasta_tools/automatic_rollbacks/state_machine.py
```diff
@@ -114,10 +114,10 @@ def start_timer(self, timeout, trigger, message_verb):
 
         timer_start = time.time()
         timer_end = timer_start + timeout
-        formatted_time = datetime.datetime.fromtimestamp(timer_end)
+        formatted_time = datetime.datetime.fromtimestamp(timer_end).strftime("%X")
 
         self.notify_users(
-            f"Will {message_verb} in {timeout} seconds, (at {formatted_time})"
+            f"Will {message_verb} in {timeout} seconds, (at {formatted_time})! Press 'Disable auto rollbacks' to cancel this."
```
EvanKrall: `start_timer` is used for more than just the automatic rollback countdown -- it's also used for the auto-abandon countdown and the auto-complete countdown. Also, the text `Disable auto rollbacks` is defined by MarkForDeploymentProcess.

Probably best to add a parameter to `start_timer` called `cancel_button_text` or something, and use that. And since afaik we don't currently allow you to cancel an auto-certify or auto-abandon countdown, you'll either need to add buttons for those or omit this part of the message `if cancel_button_text is None`.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -695,10 +695,16 @@ def valid_transitions(self) -> Iterator[state_machine.TransitionDefinition]:
...
+        )
 
     def enable_auto_rollbacks(self):
         self.auto_rollback = True
         self.auto_rollbacks_ever_enabled = True
+        self.update_slack_status(
+            f"Automatic rollback enabled for this deploy. Will watch for failures and rollback when necessary"
```
EvanKrall: Maybe also mention setting `auto_rollbacks` in `deploy.yaml` in this message too.

paasta_tools/automatic_rollbacks/slo.py
```diff
@@ -357,6 +357,7 @@ def start_auto_rollback_countdown(self) -> None:
             self.get_auto_rollback_delay(),
             "rollback_slo_failure",
             "automatically roll back",
+            "Click 'Disable auto rollbacks' to cancel this!",
```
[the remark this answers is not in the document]
solarkennedy: It does all seem pretty intertwined

solarkennedy:
Ok I think I got what you wanted

Makefile
```diff
@@ -67,3 +67,11 @@ clean:
 	find . -name '__pycache__' -delete
 	rm -rf .tox
 	rm -rf .paasta
+
+yelpy: ## Installs the yelp-internal packages into the default tox environment
+	.tox/py36-linux/bin/pip-custom-platform install -i https://pypi.yelpcorp.com/simple -r yelp_package/extra_requirements_yelp.txt -r ./extra-linux-requirements.txt
```
EvanKrall: This may make more sense as a tox environment, but I like where you're going. This shouldn't just be something that I have to search for in bash history.

PR #2434, opened 2019-09-19
Make async ttl cache handle deleted entries

stug:
Make async_ttl_cache handle case where another coroutine deleted the cache entry it was waiting on


tests/test_async_utils.py
```diff
@@ -44,14 +44,17 @@
     assert await flaky_coroutine() is None
 
 
+class DataHolder:
```
EvanKrall: would it be simpler to just use a dict or single-element list?

stug: yeah I thought about that but it felt a little obscure to me I guess

PR #2435, opened 2019-09-20
catch a couple more bravado errors when getting paasta status from th…

stug:
…e api

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -35,6 +35,8 @@ from typing import Tuple
 from typing import Type
 
 import humanize
+from bravado.exception import BravadoConnectionError
+from bravado.exception import BravadoTimeoutError
 from bravado.exception import HTTPError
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -228,8 +230,13 @@ def paasta_status_on_api_endpoint(
...
         ).result()
     except HTTPError as exc:
-        paasta_print(exc.response.text)
+        output.append(PaastaColors.red(exc.response.text))
         return exc.status_code
+    except (BravadoConnectionError, BravadoTimeoutError) as exc:
+        output.append(
+            PaastaColors.red(f"Could not connect to API: {exc.__class__.__name__}")
+        )
+        return 1
```
solarkennedy: Can you add a big catchall `except` for everything else?
I never want a user's status to crash entirely if it can't get a status on a particular instance.

stug: Yep good call, done.

PR #2441, opened 2019-09-26
remove stop using gevent in autoscaling_service_lib and remove gevent…

stug:
… from paasta entirely


solarkennedy:
This look totally sane! Always great to see requirements.txt shrink. Please ask @EvanKrall to be your primary here.

PR #2442, opened 2019-09-26
RFC: Change how CLI commands are constructed to use the git methodology

solarkennedy:
This fundamentally changes how we do CLI commands in paasta, for a couple of reasons:

1. Paasta cli commands are slow. A long time ago we had a review to make it work more like the `bzr` command, where cli parsing and import/execution were distinct. This greatly sped up the paasta-cli, but we rejected it because of the complexity it had. This makes it at least possible to do this without the big complexity, because each cli tool will be completely self-contained. The `paasta-status` command is self-contained with this change, and we can actually separate out cli parsing from execution, and then maybe we can make tab completion great again.

2. the `paasta` command doesn't need to import all the things. This will allow it to scale as we add new capabilities. In practice this doesn't actually speed anything up yet though :(

3. It makes room for other cli plugins, like git. Similar to how `git yelp-list-repos` works, new cli args can be added via `paasta-$subcommand` and they will automatically become available. We can add private commands if we wanted to. We an add new commands in other languages, etc. Maybe that means we need to give up and have `paasta flink-status`, but at least the `paasta-status` command can detect this and defer to the `flink-status` command, regardless of the language it is written in.

### DIFF
--- a/paasta_tools/cli/cli.py
+++ b/paasta_tools/cli/cli.py
@@ -14,117 +14,13 @@
 # limitations under the License.
 # PYTHON_ARGCOMPLETE_OK
 """A command line tool for viewing information from the PaaSTA stack."""
-import argparse
-import logging
+import os
 import sys
 
-import argcomplete
-import pkg_resources
 
-from paasta_tools.cli import cmds
-from paasta_tools.cli.utils import load_method
-from paasta_tools.cli.utils import modules_in_pkg as paasta_commands_dir
-from paasta_tools.utils import paasta_print
-
-
-class PrintsHelpOnErrorArgumentParser(argparse.ArgumentParser):
-    """Overriding the error method allows us to print the whole help page,
-    otherwise the python arg parser prints a not-so-useful usage message that
-    is way too terse"""
-
-    def error(self, message):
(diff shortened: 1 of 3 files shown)


solarkennedy:
We can even implement this one subcommand at a time. We don't need to big-bang change all the things. Status is an example here, but we can start by "whitelisting" subcommands and starting with less critical ones.

solarkennedy:
Yea, this branch is not meant to be shipped as-is, and only works for status. In practice I would make a whitelist and opt in commands one at a time.

poros:
@solarkennedy I very much appreciate that you are moving towards a pluggable systems for paasta commands! Thanks a lot for the consideration :)

Rob-Johnson:
yep, I think this is an agreeable approach.

I think living in both worlds for a while is the right thing to do. You don't even need to whitelist -  we can do an if else (if we can figure out a way to see if we have a subparser for the command) and go to the subcommand if it doesn't exist. gives us an easy way of extending the list of subcommands without needing a whitelist.

solarkennedy:
Got it. I'll start us off with `fsm` for real in a real PR.

PR #2443, opened 2019-09-27
Rename check_flink_services and add single test

Bronek:
Note, this change will break flink monitoring because of the rename.


debian/paasta-tools.links
```diff
@@ -2,7 +2,7 @@ opt/venvs/paasta-tools/bin/am_i_mesos_leader.py usr/bin/am_i_mesos_leader
 opt/venvs/paasta-tools/bin/autoscale_all_services.py usr/bin/autoscale_all_services
 opt/venvs/paasta-tools/bin/check_capacity.py usr/bin/check_capacity
 opt/venvs/paasta-tools/bin/check_chronos_has_jobs.py usr/bin/check_chronos_has_jobs
-opt/venvs/paasta-tools/bin/check_flink_services_replication.py usr/bin/check_flink_services_replication
```
solarkennedy: Another way you could have done this is to use the (better) entrypoints methodology, which would allow you to define commands and which python function (main) is associated with them, and you could do both.

PR #2446, opened 2019-09-28
Added external paasta command plugin capabilities

solarkennedy:
This adds the capability to have `paasta-foo` commands, and get the main `paasta` entrypoint to understand them and exec them.

When one runs `paasta --help`, you see them in the list (but right now it can't see the help, future improvement)

```
$ paasta -h
...
                        and instance
    validate            Validate that all paasta config files in pwd are
                        correct
    wait-for-deployment
                        Wait a service to be deployed to deploy_group
    api                 (run 'paasta api -h' for usage)
    fsm                 (run 'paasta fsm -h' for usage)
    deployd             (run 'paasta deployd -h' for usage)
    bounce-lock         (run 'paasta bounce-lock -h' for usage)
$
```

There is a performance hit, because you have to load 2 python interpreters :(

Maybe when we hit the 50% mark, we can cut over and use paasta-tools-go as the entrypoint for speed, and make python commands be the "external" ones!

In the future we can move more things "out" of the big paasta command by `git mv`'ing files down a level and out of the `cmds` folder, like I've done with fsm.

*Then*, we can do something similar with status, using the same technique!
We can have `paasta-status-flink`, and the `paasta-status` command can intelligently exec to it if a user is running paasta status on a flink job, etc. (probably can't support mixed paasta status, eh)


solarkennedy:
@Rob-Johnson this is ready for review.
@EvanKrall you also might have opinions on this design.

paasta_tools/cli/cli.py
```diff
@@ -38,6 +40,33 @@ def error(self, message):
         sys.exit(1)
 
 
+def list_external_commands():
+    p = subprocess.check_output(["/bin/bash", "-c", "compgen -A command paasta-"])
```
EvanKrall: From my testing it seems like compgen will list functions, which can be exported. This would confuse os.execlp later. Maybe not a big deal.

Looks like the bash argument `-p` (privileged) would prevent that:

```
-p      Turn  on  privileged mode.  In this mode, the $ENV and $BASH_ENV files are not processed, shell
        functions are not inherited from the environment, and the SHELLOPTS, BASHOPTS, CDPATH, and GLO‐
        BIGNORE  variables,  if  they  appear in the environment, are ignored.  If the shell is started
        with the effective user (group) id not equal to the real user (group) id, and the -p option  is
        not supplied, these actions are taken and the effective user id is set to the real user id.  If
        the -p option is supplied at startup, the effective user id is not reset.  Turning this  option
        off causes the effective user and group ids to be set to the real user and group ids.
```

paasta_tools/cli/cli.py
```diff
@@ -38,6 +40,33 @@ def error(self, message):
...
+
+
+def external_commands_items():
+    for command in list_external_commands():
+        command_help = get_command_help(command)
+        yield command, command_help
+
+
+def exec_out_to(command, argv):
+    os.execlp(f"paasta-{command}", *argv[1:])
```
keymone: it's a nit, but passing first element of args *and* args itself seems redundant?

since this function isn't supposed to be generic command calling facility, what about renaming it to `exec_subcommand` and just pass single `args` argument?

paasta_tools/cli/cli.py
```diff
@@ -115,13 +147,22 @@ def main(argv=None):
     Ensure we kill any child pids before we quit
     """
     logging.basicConfig()
+
+    # if we are an external command, we need to exec out early, and not
+    # let the subparser see all the extra args that it may not know about
```
keymone: not sure i understand this comment. does it imply execing out early should be the last thing that happens in this function? because there is no `return` and parser will parse those args just few lines below.

paasta_tools/cli/cli.py
solarkennedy: 👍

paasta_tools/cli/cli.py
solarkennedy: I've reworded.

paasta_tools/cli/cli.py
solarkennedy: Got it

PR #2447, opened 2019-09-30
Verify number of taskmanagers reported by Flink dashboard when monitoring

Bronek:
This will add HTTP call to Flink dashboard inside the monitoring job, and in turn compare the number of taskmanagers (as reported by Flink) against the expected number. Since this call makes use of a significant part of the stack, it is a good end-to-end test.

Extended the unit test for the "happy path", yet need to extend unit tests to cover error cases (e.g. dashboard not reachable etc.)

Note, as a result of refactoring there is also an internal change in how dashboard URL is reported by `paasta status` because now annotation will contain the full URL rather than only "root" part.

### DIFF
--- a/paasta_tools/check_flink_services_health.py
+++ b/paasta_tools/check_flink_services_health.py
@@ -17,6 +17,7 @@ Usage: ./check_flink_services_health.py [options]
 """
 import datetime
 import logging
+from typing import Optional
 from typing import Sequence
 
 import pysensu_yelp
@@ -79,8 +80,8 @@ Things you can do:
 def send_event_if_not_enough_taskmanagers(
     instance_config: FlinkDeploymentConfig,
     expected_count: int,
-    num_reported: int,
-    strerror: str,
+    num_reported: Optional[int],
+    strerror: Optional[str],
 ) -> None:
     under_replicated = False
     if strerror is None:
--- a/paasta_tools/check_flink_services_health.py
+++ b/paasta_tools/check_flink_services_health.py
@@ -19,6 +19,8 @@ import datetime
 import logging
 from typing import Sequence
 
+import pysensu_yelp
+
 from paasta_tools import flink_tools
 from paasta_tools import monitoring_tools
 from paasta_tools.check_services_replication_tools import main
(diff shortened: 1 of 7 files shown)


paasta_tools/check_flink_services_health.py
```diff
@@ -53,6 +57,66 @@ def healthy_flink_containers_cnt(si_pods: Sequence[V1Pod], container_type: str)
...
+            "Reasons this might be happening:\n"
+            "\n"
+            "  The service may simply be unhealthy. There also may not be enough resources\n"
+            "  in the cluster to support the requested instance count.\n"
+            "\n"
+            "Things you can do:\n"
+            "\n"
+            "  * Fix the cause of the unhealthy service. Try running:\n"
+            "\n"
+            "      paasta status -s %(service)s -i %(instance)s -c %(cluster)s -vv\n"
```
vkhromov: this is probably copy-pasted from somewhere, but a multiline string would be much more readable here.

paasta_tools/check_flink_services_health.py
```diff
@@ -70,7 +134,22 @@ def check_flink_service_health(
...
+        )
+        reported_taskmanagers = overview.get("taskmanagers", 0)
+    except ValueError as e:
+        strerror = str(e)
+
+    send_event_if_not_enough_taskmanagers(
+        instance_config=instance_config,
+        expected_count=taskmanagers_expected_cnt,
+        num_reported=reported_taskmanagers,
+        strerror=strerror,
```
vkhromov: It's actually surprising that `mypy` says nothing here.
`strerror` and `reported_taskmanagers` can be `None`, but are passed into a function expecting `str` and `int` types accordingly.

paasta_tools/flink_tools.py
```diff
@@ -217,5 +220,29 @@ def set_flink_desired_state(
...
+    try:
+        response = _dashboard_get(service, instance, cluster, "overview")
+        return json.loads(response)
+    except requests.RequestException as e:
+        url = e.request.url
+        err = e.strerror
+        raise ValueError(f"failed HTTP request to Jobmanager dashboard {url}: {err}")
+    except json.JSONDecodeError as e:
+        err = str(e)
+        raise ValueError(f"JSON decoding error from Jobmanager dashboard: {err}")
```
vkhromov: why not just `{e}` instead of `{str(e)}`?

paasta_tools/check_flink_services_health.py
Bronek: Thank you, I added `Optional` here. Indeed it is surprising that mypy missed it.

PR #2451, opened 2019-10-01
Removed chronos commands

solarkennedy:
Part 2 is removing the command line stuff.


solarkennedy:
@Rob-Johnson this is ready to review again

PR #2453, opened 2019-10-02
Add paasta_pool dimension to paasta.service.* metrics

jangsutsr:


solarkennedy:
I'm a bit worried about the performance impact of getting instance configs for all the things in tight loops. It is expensive to load the yaml for all these things.

Can you measure the time before and after on the synapse metrics collection script?

jangsutsr:
> I'm a bit worried about the performance impact of getting instance configs for all the things in tight loops. It is expensive to load the yaml for all these things.
> 
> Can you measure the time before and after on the synapse metrics collection script?

@solarkennedy `get_instance_config` is also invoked within loops in `emit_allocated_cpu_metrics.py`, and `emit_allocated_cpu_metrics.py` works well under cron frequency of once a minute, so we should be good performance-wise.

paasta_tools/contrib/emit_synapse_services_metrics.py
```diff
@@ -2,6 +2,7 @@
 import yelp_meteorite
```
solarkennedy: Just delete this file

PR #2454, opened 2019-10-03
Paasta 15597 bounce brutally b8s

felfilali:
PLEASE IGNORE
this is super early stage and want to share with chl && mingqiz


solarkennedy:
@felfilali if you would like to make a PR that is not suitable for merging, just click the dropdown and make a "draft" PR
https://github.blog/2019-02-14-introducing-draft-pull-requests/

paasta_tools/setup_kubernetes_job.py
```diff
@@ -103,43 +103,78 @@ def validate_job_name(service_instance: str) -> bool:
...
-        if (
-            app
-            and (app.kube_deployment.service, app.kube_deployment.instance)
-            not in existing_apps
-        ):
-            app.create(kube_client)
-        elif app and app.kube_deployment not in existing_kube_deployments:
-            app.update(kube_client)
-        else:
-            log.debug(f"{app} is up to date, no action taken")
```
EmanekaT: Instead of changing this file, I recommend you add 
```
    if self.soa_configs. get_bounce_method() == "brutal":
        app.deep_delete(kube_client)
```
in DeploymentWrapper.  You don't need to change paasta_tools/setup_kubernetes_job.py  at all.

EmanekaT: I forgot to mention. I'm not sure if deep_delete actually wait for the deletion to finish or not. And I'm not sure if that's what you expect. If you want it to wait for deletion to finish, I suggest you to create a new kubeclient and use a seperate thread for this. So it does not block other actions.

felfilali: Deep_delete doesnt wait and if we dont wait , we cannot call create, Do you recommend a background (delete+create) so the client doenst have to wait... or (delete+ wait for delete to finish+create)
Also not exactly sure where to add that if statement in the DeploymentWrapper class?

EmanekaT: @felfilali I recommend you add another delete function that waits, and also run delete + create async so it doesn't block. So (delete+ wait for delete to finish+create) async.

You can just modify create method in DeploymentWrapper class. line 192.

PR #2459, opened 2019-10-03
show health check status for verbose option

chlgit:
This shows the smartstack status.
Sample output looks like: https://fluffy.yelpcorp.com/i/0hK83L3sWXcMGlrBDnX8J3dTj1dMTSmg.html

### DIFF
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -1121,6 +1121,10 @@
                     "format": "float",
                     "description": "Unix timestamp when this app was created"
                 },
+                "smartstack": {
+                    "$ref": "#/definitions/SmartstackStatus",
+                    "description": "Status of the service in smartstack"
+                },
                 "namespace": {
                     "type": "string",
                     "description": "The namespace this app is running"
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -337,8 +337,8 @@
                     },
                     {
                         "in": "query",
-                        "description": "Omit Mesos information",
-                        "name": "omit_mesos",
+                        "description": "Include Mesos information",
+                        "name": "include_mesos",
                         "required": false,
                         "type": "boolean"
                     }
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -330,8 +330,8 @@
                     },
                     {
                         "in": "query",
-                        "description": "Omit Smartstack information",
-                        "name": "omit_smartstack",
+                        "description": "Include Smartstack information",
+                        "name": "include_smartstack",
                         "required": false,
                         "type": "boolean"
                     },
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -56,6 +56,7 @@ from paasta_tools.autoscaling.autoscaling_service_lib import get_autoscaling_inf
 from paasta_tools.cli.cmds.status import get_actual_deployments
 from paasta_tools.flink_tools import set_flink_desired_state
 from paasta_tools.kubernetes_tools import KubeClient
+from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
 from paasta_tools.long_running_service_tools import ServiceNamespaceConfig
 from paasta_tools.marathon_serviceinit import get_marathon_dashboard_links
 from paasta_tools.marathon_serviceinit import get_short_task_id
@@ -519,7 +520,12 @@ def kubernetes_smartstack_status(
         job_config
     )
 
-    expected_smartstack_count = job_config.get_instances()
+    expected_smartstack_count = marathon_tools.get_expected_instance_count_for_namespace(
+        service=service,
+        namespace=instance,
+        cluster=settings.cluster,
+        instance_type_class=KubernetesDeploymentConfig,
+    )
     expected_count_per_location = int(
         expected_smartstack_count / len(node_hostname_by_location)
     )
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -188,7 +188,7 @@ def kubernetes_instance_status(
     service: str,
     instance: str,
     verbose: int,
-    omit_smartstack: bool,
+    include_smartstack: bool,
 ) -> Mapping[str, Any]:
     kstatus: Dict[str, Any] = {}
     job_config = kubernetes_tools.load_kubernetes_service_config(
@@ -216,7 +216,7 @@ def kubernetes_instance_status(
             pod_list=pod_list,
         )
 
-        if not omit_smartstack:
+        if include_smartstack:
             service_namespace_config = kubernetes_tools.load_service_namespace_config(
                 service=service,
                 namespace=job_config.get_nerve_namespace(),
@@ -276,7 +276,7 @@ def marathon_instance_status(
     service: str,
     instance: str,
     verbose: int,
-    omit_smartstack: bool,
+    include_smartstack: bool,
     omit_mesos: bool,
 ) -> Mapping[str, Any]:
     mstatus: Dict[str, Any] = {}
@@ -299,7 +299,7 @@ def marathon_instance_status(
         )
     )
 
-    if not omit_smartstack:
+    if include_smartstack:
         service_namespace_config = marathon_tools.load_service_namespace_config(
             service=service,
             namespace=job_config.get_nerve_namespace(),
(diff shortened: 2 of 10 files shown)


paasta_tools/api/views/instance.py
```diff
@@ -207,6 +214,22 @@ def kubernetes_instance_status(
             verbose=verbose,
             pod_list=pod_list,
         )
+
+        if not omit_smartstack:
```
solarkennedy: Instead of the double-negative, can you just make this a bool for `include_smartstack` ?

solarkennedy:
@EmanekaT can you help review this? I expect us to do a lot of changes to this part of the codebase as we ramp up k8s on stateless services. 

@stug you are a local expert in the API stuff + status, please also give this a pass.

paasta_tools/api/views/instance.py
chlgit: OK, change to `include_smartstack`. This probably breaks paasta status temporary because the swagger change.

paasta_tools/smartstack_tools.py
```diff
@@ -432,6 +433,41 @@ def match_backends_and_tasks(
...
+    backend_pod_pairs = []
+
+    for backend in backends:
+        ip, port, _ = ip_port_hostname_from_svname(backend["svname"])
+        backends_by_ip[ip].append(backend)
+
+    for pod in pods:
+        ip = pod.status.pod_ip
+        for backend in backends_by_ip.pop(ip, [None]):
+            backend_pod_pairs.append((backend, pod))
```
stug: I don't really know my kubernetes yet and will defer to @EmanekaT, but do we expect more than one backend per pod?

EmanekaT: Yes. One backend per pod since a pod only runs a single instance.

paasta_tools/api/api_docs/swagger.json
```diff
@@ -330,8 +330,8 @@
                     },
                     {
                         "in": "query",
-                        "description": "Omit Smartstack information",
-                        "name": "omit_smartstack",
+                        "description": "Include Smartstack information",
+                        "name": "include_smartstack",
                         "required": false,
                         "type": "boolean"
                     },
```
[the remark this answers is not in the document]
chlgit: Yeah, I can change `omit_mesos` too. Another option would be keeping it as the current status until we remove mesos related code. 
I agree this might break paasta status for a short period because of the different issue (Although I think it's for different reason, which client tries to get the params that's not provided by API server). Doing the two-phase change seems not worth the effort. I would like to ship this with @solarkennedy's change for removing chronos together, so we only need to break it once.

paasta_tools/api/views/instance.py
```diff
@@ -672,7 +784,7 @@ def instance_status(request):
     service = request.swagger_data.get("service")
     instance = request.swagger_data.get("instance")
     verbose = request.swagger_data.get("verbose") or 0
-    omit_smartstack = request.swagger_data.get("omit_smartstack") or False
+    include_smartstack = request.swagger_data.get("include_smartstack", True)
```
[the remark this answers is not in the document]
chlgit: Yeah, you're right. `dict.get()` doesn't work as I want. On the other hand, `request.swagger_data.get("include_smartstack") or True` doesn't do what we want either. The statement would be True all the time. I will add one condition to check.

paasta_tools/api/views/instance.py
```diff
@@ -475,13 +499,87 @@ def marathon_smartstack_status(
...
+) -> MutableMapping[str, Any]:
+    ...
+
+
+def build_smartstack_location_dict(
+    location: str,
+    matched_backends_and_tasks=List[
+        Tuple[Optional[HaproxyBackend], Optional[Union[MarathonTask, V1Pod]]]
+    ],
+    should_return_individual_backends: bool = False,
```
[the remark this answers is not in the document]
chlgit: Mypy complains it. It says non-overloaded version should have the type hint that covers all overloaded cases.

paasta_tools/api/views/instance.py
```diff
@@ -475,13 +499,87 @@ def marathon_smartstack_status(
...
+
+    smartstack_replication_checker = KubeSmartstackReplicationChecker(
+        nodes=kubernetes_tools.get_all_nodes(settings.kubernetes_client),
+        system_paasta_config=settings.system_paasta_config,
+    )
+    node_hostname_by_location = smartstack_replication_checker.get_allowed_locations_and_hosts(
+        job_config
+    )
+
+    expected_smartstack_count = job_config.get_instances()
```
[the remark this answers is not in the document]
chlgit: Yeah, I do. :)

paasta_tools/smartstack_tools.py
chlgit: It should be one backend per pod now, but things would change after [PAASTA-14818](https://jira.yelpcorp.com/browse/PAASTA-14818). At the end, I just followed what we do for marathon for now.

EmanekaT: @chlgit I'm not very familiar with this. But I think those non-smartstack pods should not be exposed to synapse so they are not supposed to be backends? I guess we had ip to backend map on marathon because in Marathon world, services use node ip. But now they use pod ip, which is unique per service.

chlgit: Yup, you're also right. Currently, it's per backed per now, but this might change in the future. I don't have strong opinion here. Let me know if you would like to change.

paasta_tools/api/views/instance.py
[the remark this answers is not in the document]
chlgit: On second thought, I think this should be the same as we do for marathon tasks. I will change it.

paasta_tools/smartstack_tools.py
stug: No this seems fine, I was more just making sure I wasn't missing anything

paasta_tools/api/views/instance.py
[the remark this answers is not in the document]
stug: Oh, right :)

chlgit:
After I tested this, I can confirm this is fine. swagger++

paasta_tools/api/views/instance.py
```diff
@@ -475,13 +500,92 @@ def marathon_smartstack_status(
...
+    should_return_individual_backends: bool = False,
+) -> Mapping[str, Any]:
+
+    registration = job_config.get_registrations()[0]
+    instance_pool = job_config.get_pool()
+
+    smartstack_replication_checker = KubeSmartstackReplicationChecker(
+        nodes=kubernetes_tools.get_all_nodes(settings.kubernetes_client),
+        system_paasta_config=settings.system_paasta_config,
+    )
```
stug: So this seems like it'll only find kubernetes backends, yeah?  I think that is probably fine, but it seems like neither this nor the equivalent marathon service code is dealing with the case where we have instances in both mesos and kubernetes that are advertising using the same smartstack namespace.  Is that a problem?  Might be worth at least filing a ticket about.

paasta_tools/cli/cmds/status.py
```diff
@@ -919,6 +919,15 @@ def print_kubernetes_status(
         output.append("      Pods:")
         pods_table = format_kubernetes_pod_table(kubernetes_status.pods)
         output.extend([f"        {line}" for line in pods_table])
+
+    if kubernetes_status.smartstack is not None:
+        smartstack_status_human = get_smartstack_status_human(
+            kubernetes_status.smartstack.registration,
+            kubernetes_status.smartstack.expected_backends_per_location,
+            kubernetes_status.smartstack.locations,
+            getattr(kubernetes_status.smartstack, "error_message", None),
```
stug: Pretty sure you don't need this `getattr` -- I just had that in there originally because I added the error message later and didn't want to cause errors during deployment if the server didn't return this field.

paasta_tools/marathon_tools.py
```diff
@@ -1362,14 +1363,19 @@ def create_complete_config(
 
 
 def get_expected_instance_count_for_namespace(
-    service: str, namespace: str, cluster: str = None, soa_dir: str = DEFAULT_SOA_DIR
+    service: str,
+    namespace: str,
+    cluster: str = None,
+    instance_type_class: Type[LongRunningServiceConfig] = MarathonServiceConfig,
+    soa_dir: str = DEFAULT_SOA_DIR,
```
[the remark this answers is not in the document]
chlgit: Yeah, I know. I spent dome time to find a good home for it but didn't come out any.

paasta_tools/api/views/instance.py
chlgit: I think we won't have this case for a long time, such as more than a couple of hours, when we really start migration. I need to do some investigation about the migration plan.

paasta_tools/cli/cmds/status.py
chlgit: Cool, I'll change that.

PR #2461, opened 2019-10-07
Fix memory leak, again

vkhromov:
We still shouldn't use `self` while making the key from args.

### DIFF
--- a/paasta_tools/async_utils.py
+++ b/paasta_tools/async_utils.py
@@ -27,8 +27,10 @@ def async_ttl_cache(
 ) -> Callable[
     [Callable[..., Awaitable[T]]], Callable[..., Awaitable[T]]  # wrapped  # inner
 ]:
-    async def call_or_get_from_cache(cache, async_func, args, kwargs):
-        key = functools._make_key(args, kwargs, typed=False)
+    async def call_or_get_from_cache(
+        cache, async_func, args_without_self, args, kwargs
+    ):
+        key = functools._make_key(args_without_self, kwargs, typed=False)
         try:
             future, last_update = cache[key]
             if ttl is not None and time.time() - last_update > ttl:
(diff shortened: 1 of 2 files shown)


paasta_tools/async_utils.py
```diff
@@ -27,8 +27,10 @@ def async_ttl_cache(
 ) -> Callable[
     [Callable[..., Awaitable[T]]], Callable[..., Awaitable[T]]  # wrapped  # inner
 ]:
-    async def call_or_get_from_cache(cache, async_func, args, kwargs):
-        key = functools._make_key(args, kwargs, typed=False)
+    async def call_or_get_from_cache(
+        cache, async_func, args_without_self, args, kwargs
+    ):
+        key = functools._make_key(args_without_self, kwargs, typed=False)
```
stug: it might be nice to leave a comment here explaining the risk of the memory leak?  although maybe it's fine since you wrote a test.

tests/test_async_utils.py
```diff
@@ -164,3 +165,38 @@ class TestException(Exception):
...
+
+    x = 42
+
+    class TestClass:
+        @async_ttl_cache(ttl=None, cleanup_self=True)
+        async def f(self):
+            return x
+
+    # `instance_caches` from `async_ttl_cache()` applied to `TestClass.f`.
+    instance_caches = TestClass.f.__closure__[1].cell_contents
```
stug: Suggestion: I added a `cache` kwarg to `async_ttl_cache` so that I wouldn't have to do this.  You could use that instead to inject the cache instead of having to pull it out of the closure.

vkhromov:
@stug, I'm going to merge this PR as is, as the leak is affecting us, and will address your comment in a new PR.

stug:
@vkhromov sure -- they were just suggestions anyway so no pressure.

vkhromov:
> @vkhromov sure -- they were just suggestions anyway so no pressure.

The followup PR is https://github.com/Yelp/paasta/pull/2464

PR #2462, opened 2019-10-07
make sure k8s schema supports everything in marathon schema

chlgit:
Add the configurations shown in `marathon_schema.json`.

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -114,6 +114,11 @@
                     "exclusiveMinimum": true,
                     "default": 1024
                 },
+                "gpus": {
+                    "type": "integer",
+                    "minimum": 0,
+                    "exclusiveMinimum": false
+                },
                 "instances": {
                     "type": "integer",
                     "minimum": 0,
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -334,6 +366,33 @@
...
+                },
+                "dependencies_reference": {
+                    "type": "string"
+                },
+                "extra_docker_args": {
+                    "type": "object",
+                    "additionalProperties": {
+                        "type": "string"
+                    }
+                },
```
vkhromov: k8s doesn't support extra docker args

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -254,6 +273,13 @@
                     },
                     "additionalProperties": true
                 },
+                "marathon_shard": {
+                    "type": "integer",
+                    "minimum": 0
+                },
+                "previous_marathon_shards": {
+                    "type": "array"
+                },
```
vkhromov: `k8s_shard`?

paasta_tools/cli/schemas/kubernetes_schema.json
chlgit: Yeah, the point for this change is to allow user (or us) using `git mv` for marathon-{cluster}.yaml to kubernetes-{cluster}.yaml and clean up afterward.

paasta_tools/cli/schemas/kubernetes_schema.json
chlgit: Right, we also need to modify this later.

solarkennedy:
👍 I is ok that we don't support these features yet. As we move stagef they will "magically" start working or just not apply, and we can clean up on a second pass. This change is just to make it easier rename files back and forth.

PR #2468, opened 2019-10-10
add logging for the smartstack host used during replication checks

drmorr0:
@chriskuehl added this in manually during DAR-698, I think we should just print this all the time.

### DIFF
--- a/paasta_tools/smartstack_tools.py
+++ b/paasta_tools/smartstack_tools.py
@@ -14,6 +14,7 @@
 import abc
 import collections
 import csv
+import logging
 import socket
 from typing import cast
 from typing import Collection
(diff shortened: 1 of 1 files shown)


paasta_tools/smartstack_tools.py
```diff
@@ -479,6 +481,7 @@ def get_replication_for_instance(
         instance_pool = instance_config.get_pool()
         for location, hosts in attribute_host_dict.items():
             hostname = self._get_first_host_in_pool(hosts, instance_pool)
+            log.debug("Getting replication info from host %s", hostname)
```
chriskuehl: nit: maybe include the pool (`from host %s for pool %s`) since it will print this message out several times (once per pool)

drmorr0:
It was merged manually.

PR #2469, opened 2019-10-11
Fix a bug when instance is 0 in setup_kubernetes_jobs

EmanekaT:
I forgot 0 means false in Python :(


paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -197,7 +197,7 @@ def create(self, kube_client: KubeClient) -> None:
     def update(self, kube_client: KubeClient) -> None:
         # If autoscaling is enabled, do not update replicas.
         # In all other cases, replica is set to max(instances, min_instances)
-        if not self.get_soa_config().get("instances"):
+        if not self.get_soa_config().get("instances") is None:
```
chlgit: `if self.get_soa_config().get("instances") is not None` is better.

PR #2470, opened 2019-10-11
use instance name as container name to avoid 63 character restriction

EmanekaT:
WIth this change together with replacing '_' with '-' in yelpsoa_configs, we can avoid majority  of errors like https://fluffy.yelpcorp.com/i/tvb8Mwj1txVq4QDGSxm8LkFDRx5VWKTH.html 

I'm not sure where else container name is used. Definitely let me know if you know. Just ignore this PR if you don't know.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -680,7 +680,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
                     _exec=V1ExecAction(command=["/bin/sh", "-c", "sleep 30"])
                 )
             ),
-            name=self.get_sanitised_deployment_name(),
+            name=self.get_sanitised_instance_name(),
             liveness_probe=self.get_liveness_probe(service_namespace_config),
             ports=[V1ContainerPort(container_port=self.get_container_port())],
             security_context=self.get_security_context(),
--- a/tests/test_kubernetes_tools.py
+++ b/tests/test_kubernetes_tools.py
@@ -495,7 +495,7 @@ class TestKubernetesDeploymentConfig(unittest.TestCase):
                         period_seconds=10,
                         timeout_seconds=10,
                     ),
-                    name="kurupt-fm",
+                    name="fm",
                     ports=[V1ContainerPort(container_port=8888)],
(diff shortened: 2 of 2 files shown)


solarkennedy:
This just changes the `Name` field, so this should just cause a big bounce, right?

EmanekaT:
> This just changes the `Name` field, so this should just cause a big bounce, right?

Yes. I added reviewers just to check if there might be unexpected errors, since there might be things that rely on this convention.

vkhromov:
This will break again on instance names like `salesforce_single_payment_program_with_rass_opportunity_id_upsert_worker`, right?
(you can run
```
$ sed -n 's|^\([a-z0-9_]*\):.*|\1|p' yelp-main/marathon-*.yaml | sort -u | awk '{print length($0), $0}' | sort -n
```
in `yelpsoa-configs` to find more of such monsters).

What I'd recommend to do is to truncate the combination of the service name and the instance name so the resulting string will be shorter than some number of characters (let's call that `k`).
Please note that `k` probably should be less than `63`, as `get_kubernetes_containers()` declares just a template for pods, which later is used by things like `Deployment` to create actual pods.  While doing that, `Deployment` adds a suffix to the templated pod name, so the resulting pod name looks like something like `yelp-main-main--uswest1-85c4f84997-jh95s`.

EmanekaT:
> What I'd recommend to do is to truncate the combination of the service name and the instance name so the resulting string will be shorter than some number of characters (let's call that `k`).

It seems there are only 5 instances with names longer than 63. I think we can rename them. I don't really want to make this too complicated. Instance names are used as labels, which are used as dimensions in signalfx. This 63-character limit also applies to labels. So I would like yelpsoa_configs, labels, and container_names to share the same instance name.

vkhromov:
Max pod template name length seems to be not `63` but `63 - len("-85c4f84997-jh95s") = 46`.
Plus `get_sanitised_instance_name()` converts `_`s into `--`s, reducing that even more.

EmanekaT:
> assuming that instance names in yelpsoa will be truncated and `s/_/-/g`ed, plus there will be a pre-commit hook to enforce these rules for new names.

Cool. I definitely need to replace _ with - during migration, and there will be a pre-commit hook.

PR #2472, opened 2019-10-14
Handle 'None' when printing via the API

solarkennedy:
This appears to be happening because we are calling paasta_print in this exception handling, but in the paasta api, and the fname to print is None at the time.


solarkennedy:
Internal traceback: https://fluffy.yelpcorp.com/i/DZqpnPzklJjJPrrG5tMx15RfsQDvBvPQ.html

cc @davidbauer99

paasta_tools/mesos/cluster.py
```diff
@@ -15,34 +15,30 @@
...
-
 
 async def get_files_for_tasks(task_list, file_list, max_workers):
     no_files_found = True
 
     async def process(task_fname):
         task, fname = task_fname
         try:
-            fobj = await task.file(fname)
+            fobj = await task.file(task_fname)
```
stug: I think this will break -- `task_fname` is a tuple of `(task, fname)`

paasta_tools/utils.py
```diff
@@ -3268,8 +3268,8 @@ def set_paasta_print_file(file: Any) -> Iterator[None]:
 
 
 def paasta_print(*args: Any, **kwargs: Any) -> None:
-    f = kwargs.pop("file", sys.stdout)
-    f = getattr(TLS, "paasta_print_file", f)
+    f = kwargs.pop("file", sys.stdout) or sys.stdout
+    f = getattr(TLS, "paasta_print_file", f) or sys.stdout
```
stug: I think we got this wrong -- should be `or f` instead of `or sys.stdout`

stug: To be clear (since github is showing more than just the line I highlighted) I mean the second of these -- the `getattr`

paasta_tools/mesos/cluster.py
solarkennedy: fixed

paasta_tools/utils.py
solarkennedy: ah right

paasta_tools/utils.py
```diff
@@ -3268,8 +3268,8 @@ def set_paasta_print_file(file: Any) -> Iterator[None]:
 
 
 def paasta_print(*args: Any, **kwargs: Any) -> None:
-    f = kwargs.pop("file", sys.stdout)
-    f = getattr(TLS, "paasta_print_file", f)
+    f = kwargs.pop("file", sys.stdout) or sys.stdout
```
vkhromov: why that `or ...` is necessary here and below?

solarkennedy: because the file kwarg here could be None, and when that is the case, this function stack traces because None doesn't make sense here.
See the stack trace here:
https://fluffy.yelpcorp.com/i/DZqpnPzklJjJPrrG5tMx15RfsQDvBvPQ.html

PR #2476, opened 2019-10-17
show replicaset for verbose option

chlgit:
It will look like https://fluffy.yelpcorp.com/i/7l1xdNT1Pt78DZRxqh35SZDnKx3sqpdB.html

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -64,6 +64,7 @@ from kubernetes.client import V1Pod
 from kubernetes.client import V1PodSpec
 from kubernetes.client import V1PodTemplateSpec
 from kubernetes.client import V1Probe
+from kubernetes.client import V1ReplicaSet
 from kubernetes.client import V1ResourceRequirements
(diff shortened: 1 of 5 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1299,6 +1300,15 @@ def list_matching_deployments(
     )
 
 
+def replicasets_for_service_instance(
+    service: str, instance: str, kube_client: KubeClient
+) -> Sequence[V1ReplicaSet]:
+    return kube_client.deployments.list_namespaced_replica_set(
+        namespace="paasta",
+        label_selector=f"yelp.com/paasta_service={service},yelp.com/paasta_instance={instance}",
+    ).items
```
EmanekaT: Should this be wrapped in try catch, and return an error code? I guess it's OK to not catch the error as well. It's up to you. Looks good overall.

chlgit: Yeah, I want to keep the consistent between this and other functions, such as `list_matching_deployments`.

PR #2481, opened 2019-10-18
Add IAM role annotation for k8s

mattmb:
This is so I can use kiam with a paasta deployment.


paasta_tools/kubernetes_tools.py
```diff
@@ -937,7 +937,10 @@ def get_pod_template_spec(
         docker_volumes = self.get_volumes(
             system_volumes=system_paasta_config.get_volumes()
         )
-        annotations = {"smartstack_registrations": json.dumps(self.get_registrations())}
+        annotations = {
+            "smartstack_registrations": json.dumps(self.get_registrations()),
+            "iam.amazonaws.com/role": self.get_iam_role(),
```
vkhromov: Can you please double-check that the role equals to `""` actually mean "no role"?
I'd expect that there should be no annotation with key `iam.amazonaws.com/role` in that case.

mattmb: kiam is okay with an empty string from what I can tell. In the future I expect to change the default to some determined value such that all paasta services get their own IAM role but they can override if need be.

PR #2484, opened 2019-10-21
Set disk quota for k8s

vkhromov:
Set `ephemeral-storage` resource and limit for pods according to the
configuration, setting the disk quota as a result.

Please note that that is a soft disk quota which doesn't prevent a pod
from using more disk space than requested.  Instead of that, Kubernetes
evicts the violating pod, which then either rescheduled to another node
or becomes `Evicted`.  Please see the corresponding KEP for more
information:
https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20180906-quotas-for-ephemeral-storage.md


solarkennedy:
Please next think about how we can propagate the "evicted" messages up to the user, just like we do OOM alerts.

I'm all for adding this constraint in k8s, but I want users to know right away *why* their thing isn't working when it runs out of disk.

PR #2485, opened 2019-10-21
K8s attribute naming scheme migration (phase 1)

keymone:
Duplicate `yelp.com/paasta_` into `paasta.yelp.com/` attributes in relevant places.


vkhromov:
```bash
$ git grep yelp.com/paasta_service paasta_tools | wc -l
25
$ git grep paasta.yelp.com/service paasta_tools | wc -l
11
```
I haven't looked at the code intently, but can you please double-check that you've actually changed all relevant places?

keymone:
@vkhromov the discrepancy is because i've changed all the places we write those attributes but not the places where we read them.

PR #2487, opened 2019-10-21
validate instance/service names no more than 63 characters

EmanekaT:
Check if service name is longer than 63 for all serivces.
Check if instance name is longer than 63 for k8s instances.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -14,7 +14,7 @@ so you are free to use them for YAML templates.
 
 **Note** that service names (the name of the folder where your config file is located) should be no more than 63 characters.
 For kubernetes services(config files with kubernetes as prefix), the instance names should be nore more than 63 characters as well.
-_ is counted as two characters.
+_ is counted as two character. We convert _  to -- because underscore is not allowed in kubernetes pod names.
 
 Example::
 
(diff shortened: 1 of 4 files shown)


docs/source/yelpsoa_configs.rst
```diff
@@ -12,6 +12,10 @@ Duplication can be reduced by using YAML `anchors and merges <https://gist.githu
 PaaSTA will **not** attempt to run any definition prefixed with ``_``,
 so you are free to use them for YAML templates.
 
+**Note** that service names (the name of the folder where your config file is located) should be no more than 63 characters.
+For kubernetes services(config files with kubernetes as prefix), the instance names should be nore more than 63 characters as well.
+_ is counted as two characters.
```
chlgit: "Should be `no` more"?
Also, could you add an example to explain why '_' is counted as two characters?

paasta_tools/cli/cmds/validate.py
```diff
@@ -108,6 +109,37 @@ def get_schema(file_type):
...
 
+def validate_instance_names(config_file_object, file_path):
+    return_code = True
+    for instance_name in config_file_object:
+        if (
+            not instance_name.startswith("_")
+            and len(sanitise_kubernetes_name(instance_name)) > 63
+        ):
+            return_code = False
+            paasta_print(
```
chlgit: I think you want to have a list to record the instances violating the restriction together, so that you can just `paasta_print` the error message once. Also, by doing that, you can do `return len(list) == 0`.

PR #2489, opened 2019-10-21
Disk quota for Docker

vkhromov:


paasta_tools/utils.py
```diff
@@ -113,6 +114,8 @@
 DEFAULT_CPU_PERIOD = 100000
 DEFAULT_CPU_BURST_ADD = 1
 
+DOCKER_DISK_QUOTA = strtobool(os.environ.get("DOCKER_DISK_QUOTA", "0"))
```
solarkennedy: I personally wouldn't bother with this. I vote just us a normal release and roll it out slowly.

paasta_tools/utils.py
```diff
@@ -460,6 +463,13 @@ def format_docker_parameters(
             {"key": "cpu-period", "value": "%s" % int(self.get_cpu_period())},
             {"key": "cpu-quota", "value": "%s" % int(self.get_cpu_quota())},
         ]
+        if self.use_docker_disk_quota():
+            parameters.append(
+                {
+                    "key": "storage-opt",
+                    "value": "size=%s" % int(self.get_disk() * 1024 * 1024),
```
solarkennedy: can you do fstrings instead?

solarkennedy:
This will do a big bounce, so pin and roll it out slowly.

paasta_tools/utils.py
vkhromov: We don't have XFS everywhere (think e.g. about Travis tests and MBP), so this one it to enable disk quota only where it's supported.

solarkennedy: But does it hurt to set these options when quotas are not supported on the underlying storage driver?

If it doesn't hurt, then I vote have them on all the time.

vkhromov: Not sure that I understand you, so let me explain this env var-base design.

We cannot just add `--storege-opt size=something` arg to every Docker call, Docker will fail to start with message like `Error response from daemon: --storage-opt is supported only for overlay over xfs with 'pquota' mount option` if it doesn't support disk quotas on the host.

On the other hand, to universally verify from `paasta-tools` code if Docker running on a host supports disks quota we need to do the following chain of not-so-trivial checks:
1. Get the path where Docker stores containers volumes (that could require reading Docker config)
2. Get the filesystem to which that path points (e.g. we use a chain of symlinks at some of our hosts)
3. Get the filesystem details like FS type and mounting options

So. IMO it's much easier to just set the env variable on a host which we already configured to have Docker volumes dir on a volume formatted to XFS filesystem and mounted with prjquota option.

solarkennedy: Got it.

Historically we used the `/etc/paasta/paasta.json` (SystemPaastaConfig) because we have the puppet facilities to edit that, and restart deployd when things change, etc.

I'm not going to block on using an env variable, but just keep in mind this is different to how we normally configure paasta stuff system-wide.

vkhromov: Please see the updated version which uses a flag from `/etc/paasta/` to control the disk quota enforcement.

PR #2502, opened 2019-10-24
I think not cleaning up even dead masters is good

azretkenzhaliev:
What do you think? @Rob-Johnson


Rob-Johnson:
add a unit test, but otherwise looks good :shipit:

PR #2504, opened 2019-10-24
set the paasta_instance docker label for spark-run

drmorr0:
@qui said:

> no, because that info isn’t readily available from the Mesos master, but the labels are
> i mean, the environment variables are used by the DockerStats collector
> both are necessary
> the paasta_instance_resource_utilization_summary report joins the tmp_paasta_allocation report (which uses labels) and the fullerite data (which uses env)

So this change makes the Docker labels match the environment variables, which should make this stuff show up in Splunk.


drmorr0:
(note that tests are failing but the same set of tests were failing on master and should be fixed by #2503)

PR #2505, opened 2019-10-24
Mount MESOS_SANDBOX in paasta local-run

huadongliu:
Service owners are unlikely (or should not) to be aware of MESOS_SANDBOX. I think MESOS_SANDBOX was set in local-run to mimic the behavior of running in Mesos. Thus, a common scratch directory like /tmp is more appropriate than the mesos specific /mnt/mesos/sandbox.

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -65,6 +65,9 @@ from paasta_tools.utils import TimeoutError
 from paasta_tools.utils import validate_service_instance
 
 
+MESOS_SANDBOX_DIR = "/mnt/mesos/sandbox"
+
+
 def parse_date(date_string):
     return datetime.datetime.strptime(date_string, "%Y-%m-%d")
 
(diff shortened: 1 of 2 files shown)


solarkennedy:
I'm confused. local-run mimics mesos-based tasks for good reason. The variable has MESOS in the name, and is set that way on mesos for real. 

Is there some other way to fix the core problem without "pretending" to not be here?

What if instead we still set the variable to be "correct", but we bind mount /tmp/mktmpdir to be there? (just like mesos makes a tmpdir?)

huadongliu:
Agree that mounting a temporary directory is better. I will update the PR.

paasta_tools/cli/cmds/local_run.py
```diff
@@ -953,7 +956,8 @@ def configure_and_run_docker_container(
         return 1
 
     soa_dir = args.yelpsoa_config_root
-    volumes = list()
+    (mktemp_rc, mktemp_output) = _run("mktemp -d")
```
EvanKrall: would it be better to use https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory instead?

huadongliu:
I added SPARK_LOCAL_DIRS=/tmp to the failing Spark tests.

PR #2506, opened 2019-10-25
kubernetes_tools: fix NoneType is not iterable in _is_ready check

keymone:


paasta_tools/kubernetes_tools.py
```diff
@@ -1387,7 +1387,7 @@ def filter_pods_by_service_instance(
 
 def _is_it_ready(it: Union[V1Pod, V1Node],) -> bool:
     ready_conditions = [
-        cond.status == "True" for cond in it.status.conditions if cond.type == "Ready"
+        cond.status == "True" for cond in it.status.conditions or [] if cond.type == "Ready"
```
vkhromov: I'm surprised that this one hasn't been catched by `mypy`.

Bronek: there is an unrelated `mypy` regression on master.

PR #2513, opened 2019-10-28
Basic support for paasta local-run for k8s

solarkennedy:
This doesn't do anything fancy about sidecars or whatever yet, but will prevent local-run from crashing when running it for something in stagef.

This code looks a little overkill? But maybe we need room for more stuff and other frameworks?

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -897,30 +897,16 @@ def run_docker_container(
     return returncode
 
 
-def command_function_for_framework(framework, date):
+def format_command_for_type(command, instance_type, date):
     """
-    Given a framework, return a function that appropriately formats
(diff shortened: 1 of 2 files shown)


solarkennedy:
> this function is needlessly verbose, lets reduce it down to
> 
> ```
> def command_function_for_framework(framework, date):
> """
> Given a framework, return a function that appropriately formats
> the command to be run.
> """
>     if framework in ["marathon", "adhoc", "kubernetes"]:
>         return lambda cmd: cmd
>     elsif framework == "tron":
>         return lambda cmd: parse_time_variables(cmd, date)
>     else:
>         raise ValueError(f"Invalid Framework for local_run to run: {framework}")
> ```
> 
> otherwise lgtm

Agreed. I've changed it to just be a function (instead of a function that returns a function) instead.

PR #2519, opened 2019-10-29
fix paasta status error on stagef

EmanekaT:
>   File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/mesos_tools.py", line 752, in get_mesos_slaves_grouped_by_attribute
>     sorted_slaves = sorted(slaves, key=lambda slave: slave["attributes"].get(attribute))
> TypeError: '<' not supported between instances of 'str' and 'NoneType'

I don't know why attribute would be none. But setting it to '' should fix the issue.

### DIFF
--- a/paasta_tools/mesos_tools.py
+++ b/paasta_tools/mesos_tools.py
@@ -750,7 +750,11 @@ def get_mesos_slaves_grouped_by_attribute(slaves, attribute):
               (response can contain multiple 'attribute_value)
     """
     sorted_slaves = sorted(
-        slaves, key=lambda slave: slave["attributes"].get(attribute, "")
+        slaves,
+        key=lambda slave: (
+            slave["attributes"].get(attribute) is None,
+            slave["attributes"].get(attribute),
+        ),
     )
     return {
         key: list(group)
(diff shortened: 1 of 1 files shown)


paasta_tools/mesos_tools.py
```diff
@@ -749,7 +749,9 @@ def get_mesos_slaves_grouped_by_attribute(slaves, attribute):
     :returns: a dictionary of the form {'<attribute_value>': [<list of hosts with attribute=attribute_value>]}
               (response can contain multiple 'attribute_value)
     """
-    sorted_slaves = sorted(slaves, key=lambda slave: slave["attributes"].get(attribute))
+    sorted_slaves = sorted(
+        slaves, key=lambda slave: slave["attributes"].get(attribute, "")
+    )
     return {
         key: list(group)
         for key, group in itertools.groupby(
```
solarkennedy: What about line 756? Looks... the same?

EmanekaT: groupby should be fine since it does not compare keys with </>.

EmanekaT: I tested that key could be None.

EmanekaT:
> I agree that this will fix the attribute error, but doesn't it also change the sorting behavior?
> 
> I guess this never worked with `None` before, but by making the default the `""`, will those get sorted to the top? Makes me think this change will sort broken slaves right to the top, which sounds bad.

Good point. I just updated this so that None attributes is sorted to the end.

PR #2520, opened 2019-10-30
RELENG-41130: Track PaaSTA rollback events

nemacysts:
We want to track rollback rates for services in EE Metrics, so we've instrumented both the `paasta rollback` command and the Slack/SLO rollback events to log to the same audit stream that we use for m-f-d.

I wanted to log this as part of the existing _log_audit calls in m-f-d, but it doesn't seem like there's any way of knowing at that point if we're rolling back and why.

**testing done**: 

- `make test itest`
- manual (i.e., clicking buttons and checking the audit stream)

**questions**:
- writing tests for the state machine stuff doesn't look particularly simple from what I saw - any ideas here?
- I added an enum for type safety, but not sure if it'd be preferable to just use simple named constants and type things as `str` - in the same vein, would we prefer TypedDicts for all new dicts (e.g., the one returned by `__build_rollback_audit_details`) or am I overthinking things?
- maybe it would make more sense to just build mock.call objects and compare those in the rollback tests?

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -29,6 +29,7 @@ from queue import Queue
 from threading import Event
 from threading import Thread
 from typing import Collection
+from typing import Dict
 from typing import Iterator
 from typing import Mapping
 from typing import Optional
@@ -66,6 +67,7 @@ from paasta_tools.utils import list_services
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import paasta_print
 from paasta_tools.utils import PaastaColors
+from paasta_tools.utils import RollbackTypes
 from paasta_tools.utils import TimeoutError
 
 
@@ -604,6 +606,7 @@ class MarkForDeploymentProcess(SLOSlackDeploymentProcess):
                 "source": self.rollforward_states,
                 "dest": "start_rollback",
                 "trigger": "rollback_button_clicked",
+                "before": self.log_user_rollback,
             }
             yield {
                 "source": self.rollback_states,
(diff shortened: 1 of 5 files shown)


paasta_tools/cli/cmds/rollback.py
```diff
@@ -206,14 +209,22 @@ def paasta_rollback(args):
...
-        returncode = max(
-            mark_for_deployment(
-                git_url=git_url,
-                service=service,
-                deploy_group=deploy_group,
-                commit=commit,
-            ),
-            returncode,
+        rolled_back_from = get_currently_deployed_sha(service, deploy_group)
+        returncode |= mark_for_deployment(
```
solarkennedy: Huh, til `|=`

solarkennedy:
> * writing tests for the state machine stuff doesn't look particularly simple from what I saw - any ideas here?
Yea it is hard. I wouldn't bother personally, I think your test coverage is ok.

> * I added an enum for type safety, but not sure if it'd be preferable to just use simple named constants and type things as `str` - in the same vein, would we prefer TypedDicts for all new dicts (e.g., the one returned by `__build_rollback_audit_details`) or am I overthinking things?
Hmm. I think the enum is nice, idk wfm.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -935,6 +939,34 @@ def deploy_group_is_set_to_notify(self, notify_type):
...
+    ) -> Dict[str, str]:
+        return {
+            "rolled_back_from": self.commit,
+            "rolled_back_to": self.old_git_sha,
+            "rollback_type": rollback_type.value,
+            "deploy_group": self.deploy_group,
+        }
+
+    def log_slo_rollback(self) -> None:
+        _log_audit(
```
solarkennedy: How bullet proof is this function call? Under no circumstances should a failure here block a deploy or crash a rollback or anything bad.

The user experience is particularly bad when things like this happen because the slack interface just "stops working"

nemacysts: afaik, it's pretty bullet-proof (it's using some pretty battle-tested logging code underneath) and I can't think of a way to make my helper function throw an exception (that wouldn't be caught by mypy before a release) - if we want to be extra-safe, we could guard this with a catch-all try-except.

PR #2523, opened 2019-10-30
Handle tron jobs when they haven't run yet in paasta status

solarkennedy:


paasta_tools/api/views/instance.py
```diff
@@ -105,7 +105,12 @@ def tron_instance_status(
 
     job_content = client.get_job_content(job=job)
     latest_run_id = client.get_latest_job_run_id(job_content=job_content)
```
keymone: sorry for late response, i think the bug is actually in this line :(

PR #2525, opened 2019-10-31
Report an error when tron fails to get job status

keymone:


paasta_tools/api/views/instance.py
```diff
@@ -102,15 +102,16 @@ def tron_instance_status(
...
     job_content = client.get_job_content(job=job)
-    latest_run_id = client.get_latest_job_run_id(job_content=job_content)
+
     try:
-        action_run = client.get_action_run(job=job, action=action, run_id=latest_run_id)
-    except IndexError:
-        action_run = {"state": f"Hasn't Run Yet (no job run id #{latest_run_id})"}
+        latest_run_id = client.get_latest_job_run_id(job_content=job_content)
+        if not latest_run_id:
+            action_run = {"state": f"Hasn't Run Yet (no job run id #{latest_run_id})"}
```
solarkennedy: Huh. I thought for sure I looked at the stacktrace and found that there was an ID, but it was crashing inside the get_action_run function.

If this get_latest_run really returns none (and not 0 or 1) then I don't think printing "None" here helps the user.

keymone: from stacktrace:

     File "/opt/venvs/paasta-tools/lib/python3.6/site-packages/paasta_tools/tron/client.py", line 113, in get_latest_job_run_id
       return job_runs[0]["run_num"]
    IndexError: list index out of range`

but you're right about `not latest_run_id`, as a rubyist at heart i will never accept that `0` is falsy

PR #2528, opened 2019-11-01
Fix bounce_method on k8s instances

keymone:
also drop REVMAP as it's no longer used anywhere


keymone:
merged it manually because /shrug

PR #2530, opened 2019-11-05
paasta status on kubernetes should use pod_ip in smartstack section

EmanekaT:


mattmb:
I have a similar fix in https://github.com/Yelp/paasta/pull/2518

The only thing I wasn't sure about is whether to keep the host name too. The reason it's useful is because we use AWS hostnames in k8s and so the k8s section doesn't tell you which AZ and the Pod IP doesn't give you that but the smartstack host does.

mattmb:
I've decided to merge my PR since it had some other stuff in it that I wanted to get out. I'll give you a shipit on this too though so if you feel strongly that the hack I did of putting host name and pod ip in the same field is bad then I'm quite happy for you to merge this to replace it

PR #2532, opened 2019-11-06
Always make sure we are using the safe c loader for yaml

solarkennedy:
I was doing a thing and wondering why the yaml loading was slow.

I would like to add wrappers so nobody can ever forget to use yaml in the right way (although they can forget to use the wrapper).

Modern version of yaml are now safe by default, but we need to still use the c loader where we can.


vkhromov:
To enforce using the wrapper, can we add a pre-commit hook which prevents committing code with `import yaml` everywhere besides `paasta_tools/yaml.py`?

PR #2534, opened 2019-11-07
Dump subcommands information into json for consumption by other commands

mks-m:
golang binary could use this for printing help messages or generating autocompletion, etc

### DIFF
--- a/paasta_tools/cli/cli.py
+++ b/paasta_tools/cli/cli.py
@@ -15,6 +15,7 @@
 # PYTHON_ARGCOMPLETE_OK
 """A command line tool for viewing information from the PaaSTA stack."""
 import argparse
+import json
 import logging
 import os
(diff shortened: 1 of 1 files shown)


solarkennedy:
This is interesting... does it work?

I don't consider this a "public" command, can you move it out of the main `paasta` part and into something like `paasta_dump_command_choices` or something? That way it isn't in the "critical" path of the main paasta command, doesn't slow it down, etc.

keymone:
it does work, i don't know if we really need it, maybe eventually, i'll just let it sit here for now

solarkennedy:
@keymone Are you still planning on taking this approach?

keymone:
yes, eventually, it's a bit low on priorities list.

mattmb:

PR #2536, opened 2019-11-07
Avoid cutting name of flink jobs running in cluster

AshishKhatkar:
Currently, it is assumed that job name length will be 32 characters. While checking paasta status we see that name of jobs is getting trimmed. If we avoid cutting the names, it will  be useful while using the flink-supervisor cli. Refer to FLINK-1830 for further details.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -852,7 +852,6 @@ def print_flink_status(
 
     output.append(f"    Jobs:")
     if verbose:
-        fmt = "      {job_name_var: <{max_column_length}} State       Job ID                           Started"
         output.append(
             f'      {"Job Name": <{allowed_max_job_name_length}} State       Job ID                           Started'
         )
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -838,13 +838,28 @@ def print_flink_status(
         f" {status.overview['slots-available']}/{status.overview['slots-total']} slots available"
     )
 
+    # Avoid cutting job name. As opposed to default hardcoded value of 32, we will use max length of job name
+    max_job_name_length = max(
+        [len(job["name"].split(".", 2)[2]) for job in status.jobs]
+    )
+    # Limiting longest allowed job name to be 120 chars. This is to ensure while printing on cli the formatting is maintained on normal screens
+    allowed_max_job_name_length = min(120, max_job_name_length)
+
     output.append(f"    Jobs:")
     if verbose:
+        fmt = "      {job_name_var: <{max_column_length}} State       Job ID                           Started"
         output.append(
-            f"      Job Name                         State       Job ID                           Started"
+            fmt.format(
+                job_name_var="Job Name", max_column_length=allowed_max_job_name_length,
+            )
         )
     else:
-        output.append(f"      Job Name                         State       Started")
+        fmt = "      {job_name_var: <{max_column_length}} State       Started"
+        output.append(
+            fmt.format(
+                job_name_var="Job Name", max_column_length=allowed_max_job_name_length,
+            )
+        )
 
     # Use only the most recent jobs
     unique_jobs = (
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -838,13 +838,28 @@ def print_flink_status(
         f" {status.overview['slots-available']}/{status.overview['slots-total']} slots available"
     )
 
+    # Avoid cutting job name. As opposed to default hardcoded value of 32, we will use max length of job name
+    max_job_name_length = max(
+        [len(job["name"].split(".", 2)[2]) for job in status.jobs]
```
poros: not blocking: `job["name"].split(".", 2)[2]` is repeated below, perhaps you could define an internal function or lambda to avoid drift

paasta_tools/cli/cmds/status.py
```diff
@@ -838,13 +838,28 @@ def print_flink_status(
...
+    )
+    # Limiting longest allowed job name to be 120 chars. This is to ensure while printing on cli the formatting is maintained on normal screens
+    allowed_max_job_name_length = min(120, max_job_name_length)
+
     output.append(f"    Jobs:")
     if verbose:
+        fmt = "      {job_name_var: <{max_column_length}} State       Job ID                           Started"
```
vkhromov:
```
f'      {"Job Name": <{allowed_max_job_name_length}} State       Started'
```
is shorter and clearer

paasta_tools/cli/cmds/status.py
```diff
@@ -838,13 +838,28 @@ def print_flink_status(
...
         output.append(
-            f"      Job Name                         State       Job ID                           Started"
+            fmt.format(
+                job_name_var="Job Name", max_column_length=allowed_max_job_name_length,
+            )
         )
     else:
-        output.append(f"      Job Name                         State       Started")
+        fmt = "      {job_name_var: <{max_column_length}} State       Started"
```
vkhromov: the same as above

paasta_tools/cli/cmds/status.py
```diff
@@ -838,13 +838,28 @@ def print_flink_status(
         f" {status.overview['slots-available']}/{status.overview['slots-total']} slots available"
     )
 
+    # Avoid cutting job name. As opposed to default hardcoded value of 32, we will use max length of job name
+    max_job_name_length = max(
+        [len(job["name"].split(".", 2)[2]) for job in status.jobs]
+    )
+    # Limiting longest allowed job name to be 120 chars. This is to ensure while printing on cli the formatting is maintained on normal screens
+    allowed_max_job_name_length = min(120, max_job_name_length)
```
vkhromov: Please use the real width of the terminal, something like:
```
allowed_max_job_name_length = min(shutil.get_terminal_size((120, 50)).columns, max_job_name_length)
```
Also, please make sure that the whole line fits into the terminal and not only the job name column.

paasta_tools/cli/cmds/status.py
AshishKhatkar: Created function instead of lambda. 
Reason : "E731 do not assign a lambda expression, use a def"

paasta_tools/cli/cmds/status.py
poros: that's actually a good idea +1

AshishKhatkar: Thanks, this makes more sense.

paasta_tools/cli/cmds/status.py
AshishKhatkar: Ack.

paasta_tools/cli/cmds/status.py
AshishKhatkar: Ack

paasta_tools/cli/cmds/status.py
```diff
@@ -838,13 +843,23 @@ def print_flink_status(
         f" {status.overview['slots-available']}/{status.overview['slots-total']} slots available"
     )
 
+    # Avoid cutting job name. As opposed to default hardcoded value of 32, we will use max length of job name
+    max_job_name_length = max([len(get_flink_job_name(job)) for job in status.jobs])
+    # Limiting longest allowed job name to be 120 chars. This is to ensure while printing on cli the formatting is maintained on normal screens
+    allowed_max_job_name_length = min(
```
vkhromov: Please see the last paragraph of the previous comment about this code.
`allowed_max_job_name_length` is width of the first column only, it cannot be equal to the width of the terminal, but should be less of it, otherwise  there will be no space for `State`, `Job ID`, and `Started` columns.
`allowed_max_job_name_length` need to be reduced so there will be space for these columns.

AshishKhatkar: Please check the calculations once again. Sorry for misunderstanding.

PR #2537, opened 2019-11-07
Make the paasta api not crash on tron jobs that haven't run yet

solarkennedy:


paasta_tools/api/views/instance.py
```diff
@@ -129,17 +129,17 @@ def tron_instance_status(
 
     if action:
         status["action_name"] = action
-    if action_run["state"]:
+    if action_run.get("state"):
```
keymone: lets go more meta here:

    for attr in ["state", "start_time", "raw_command", "stdout", "stderr", "command"]:
        attr_val = action_run.get(attr)
        if attr_val is not None:
            status[f"action_{attr}"] = attr_val

keymone: of course with exceptions for `stdout` and `stderr`, because irregularities are fun :(

solarkennedy: Um... what if we didn't go more meta?
I don't think the DRY is worth it here, there are 2 of one kind, 4 or so of another, and I kinda want better defaults for some of these maybe like "Hasn't run yet" for end time or something?

keymone: well.. i very much prefer getting rid of these `if ... elif` chains, they are a smell to me. when there are irregularities in handling of each option - i prefer a dict of functions even more. in fact i think we should be taking paasta in that direction and replacing all the `if` chains with modules implementing "interfaces" for total handling of instance types rather than spreading it out into `status`/`start_stop_restart`/`logs`/`etc`, that way we can easily support new instance types by requiring `paasta-new-instance-type` to implement said interface and invoking that binary with relevant options. if multiple different instance types have to be processed - multiple binaries would be invoked with a list of instances to process and json or whatever output would be collected by paasta and rendered into some output.

solarkennedy: We chatted in slack. We both agree that in the longer term want a better interface for defining `status` on stuff. 

In the shorter term, I'm just going to add these .get's to stop the stack traces.

PR #2538, opened 2019-11-08
Removed remaining code for ssh-based paasta status

solarkennedy:
Lots of dead code to remove here, but carefully.

Some functions were being imported from other places, so I moved that code from serviceinit to where it made more sense, along with the tests.

Then I removed the remaining serviceinit code and tests.


solarkennedy:
@stug this is such a large diff. I'll ship this monday.

PR #2540, opened 2019-11-08
Use all the fstrings

solarkennedy:
Consistency in the codebase is a big nice to have.

I used @asottile's flynt command to convert these.

What could possibly go wrong! :)

### DIFF
--- a/paasta_tools/adhoc_tools.py
+++ b/paasta_tools/adhoc_tools.py
@@ -35,7 +35,7 @@ def load_adhoc_job_config(
     general_config = service_configuration_lib.read_service_configuration(
         service, soa_dir=soa_dir
     )
-    adhoc_conf_file = "adhoc-%s" % cluster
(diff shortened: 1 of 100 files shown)


asottile:
I didn't write `flynt` and would suggest not using it, it performs unsafe changes

asottile:
(I assume you saw it due to me forking it, I'm currently trying to disassociate my software with it currently since users are confusing it with pyupgrade which performs the same function but with safety checks)

solarkennedy:
Whoops! sorry @asottile I didn't read things closely enough. 
We are running with py3-plus, but not py36-plus! (til).

I'll try this with pyupgrade first and see how it goes. Thanks!

I would send love to you again if there was such a thing cross companies. 💌

asottile:
I'm still in SF :) would be happy to grab lunch or something!

asottile:
and yeah the test failures are flynt doing unsafe things and breaking your code :P

PR #2546, opened 2019-11-12
add hpa_metrics_collector

EmanekaT:
I haven't tested this yet after modification and will test this manually after shipped a new change to HPA to add a label that enables sfx adapter.
 I plan to run this as a daemon that launches together with kubernetes pod (or paasta?). Cluster name and sfx token will be passed in as arguments specified in puppet.


EmanekaT:
I'll try to move this to fullerite instead.

PR #2547, opened 2019-11-12
limit the length of the volume name

chlgit:
The length of the docker volume can't be more than 63 characters. Also, persistent name can't be more than 253 characters.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -12,6 +12,7 @@
 # limitations under the License.
 import base64
 import copy
+import hashlib
 import itertools
 import json
 import logging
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -474,20 +475,26 @@ def get_deployment_strategy_config(self) -> V1DeploymentStrategy:
...
+            sanitised_name = (
+                sanitised_name[0 : length_limit - 4]
+                + hashlib.md5(sanitised_name.encode("ascii")).hexdigest()[:4]
+            )
+        return sanitised_name
 
     def get_docker_volume_name(self, docker_volume: DockerVolume) -> str:
         return self.get_sanitised_volume_name(
-            "host--{name}".format(name=docker_volume["hostPath"])
+            "host--{name}".format(name=docker_volume["hostPath"], length_limit=63)
```
keymone: (here and in pv name) seems you're passing `length_limit` to `.format()` instead of `get_sanitised_colume_name`?

paasta_tools/kubernetes_tools.py
```diff
@@ -474,20 +475,26 @@ def get_deployment_strategy_config(self) -> V1DeploymentStrategy:
...
         volume_name = volume_name.rstrip("/")
         sanitised = volume_name.replace("/", "slash-").replace(".", "dot-")
-        return sanitise_kubernetes_name(sanitised)
+        sanitised_name = sanitise_kubernetes_name(sanitised)
+        if length_limit and len(sanitised_name) > length_limit:
+            sanitised_name = (
+                sanitised_name[0 : length_limit - 4]
+                + hashlib.md5(sanitised_name.encode("ascii")).hexdigest()[:4]
+            )
+        return sanitised_name
```
vkhromov: Can you please add a separator between the truncated original name and the hash suffix?

chlgit: Good idea. I will add '--' between them.

paasta_tools/kubernetes_tools.py
chlgit: Nice catch, thanks.

PR #2549, opened 2019-11-12
Add install-hooks make and tox targets

drolando:
Adds a make target that lets you easily install the pre-commit hooks for this repo without having to run the entire test suite.

There doesn't seem to be a generic virtualenv (.paasta only contains tox, no other dependencies afaict), so I had to create a new one (.tox/install-hooks). But since the only dependency is pre-commit I don't think it's a big deal


drmorr0:
I wouldn't say no if you wanted to set up a "generic virtualenv" target, too, but that's up to you.  This is fine as well.

PR #2551, opened 2019-11-12
Fix hpa to work with sfx metrics adapter

EmanekaT:
* Added a metrics selector that allows sfx adapter to filter out cluster specific metrics
* Fix api version.
Tested on kubestage.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -383,7 +383,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         )
 
     def get_autoscaling_metric_spec(
-        self, name: str, namespace: str = "paasta"
+        self, name: str, cluster: str, namespace: str = "paasta"
     ) -> Optional[V2beta1HorizontalPodAutoscaler]:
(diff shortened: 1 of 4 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -398,7 +398,9 @@ def get_autoscaling_metric_spec(
         # TODO support multiple metrics
         metrics = []
         target = autoscaling_params["setpoint"] * 100
+        annotations: Dict[str, str] = {}
         # TODO support bespoke PAASTA-15680
+        selector = V1LabelSelector(match_labels={"kubernetes_cluster": cluster})
```
EvanKrall: This selector affects the signalfx query? It looks like a k8s selector.

EmanekaT: The sfx adapter reads the selector and converts it dimension
https://github.com/signalfx/signalfx-k8s-metrics-adapter/tree/a3e1f4e374bfe90676e28718a2df7f7fc4e46dd5#signalfx-smart-agent-dependency

HPA queries the metrics server (SFX adapter in this case), so it doesn't care the selector.

PR #2552, opened 2019-11-13
Add PAASTA_* environment variables

EmanekaT:
I didn't add MARATHON_APP_ID because it's marathon only.


EmanekaT:
It looks like I need to add PAASTA* to docker_wrapper as well.

paasta_tools/kubernetes_tools.py
```diff
@@ -621,6 +622,42 @@ def get_kubernetes_secret_env_vars(
...
+                ),
+            ),
+            V1EnvVar(
+                name="PAASTA_APP_RESOURCE_MEM",
+                value_from=V1EnvVarSource(
+                    resource_field_ref=V1ResourceFieldSelector(
+                        resource="requests.memory"
+                    )
+                ),
+            ),
```
EvanKrall: we should just set these in InstanceConfig.get_env()

paasta_tools/kubernetes_tools.py
```diff
@@ -621,6 +622,42 @@ def get_kubernetes_secret_env_vars(
             )
         return ret
 
+    def get_service_container_env(self) -> Sequence[V1EnvVar]:
+        kubernetes_env = [
+            V1EnvVar(name="PAASTA_APP_DOCKER_IMAGE", value=self.get_docker_image(),),
```
EvanKrall: should set this in instance_config.get_env()

EmanekaT:
I realized that I still need to modify docker_wrapper.py.

paasta_tools/cli/cmds/local_run.py
```diff
@@ -622,6 +622,14 @@ def get_local_run_environment_vars(instance_config, port0, framework):
         env["MARATHON_APP_LABELS"] = ""
         env["MARATHON_APP_ID"] = "/simulated_marathon_app_id"
         env["MARATHON_HOST"] = hostname
+        env["PAASTA_HOST"] = hostname
```
solarkennedy: I think this is the wrong place for these. The only reason we have these special variables is because they *come* from mararathon, so we have to simulate them.

The other variables are set further up the top.

But instead can you make this a function of the base instance class? That way they will automatically get applied, whether they happen in local run, or k8s, or spark, or tron, or whatever.

solarkennedy:
> I realized that I still need to modify docker_wrapper.py.

The docker_wrapper currently only applies to mesos tasks and paasta local-run, it won't (Currently does not) apply to k8s.

paasta_tools/cli/cmds/local_run.py
EmanekaT: I moved everything other than PAASTA_HOST to get_env. PAASTA_HOST is special so it's added sperately for k8s in kubernetes_tools, mesos in docker_wrapper, and in local_run.py.

EmanekaT:
Added:
MARATHON_APP_DOCKER_IMAGE
MARATHON_APP_RESOURCE_MEM
MARATHON_APP_RESOURCE_CPUS
MARATHON_APP_RESOURCE_GPUS
MARATHON_APP_RESOURCE_DISK
MARATHON_HOST
MARATHON_PORT 

Missing:
MARATHON_PORT0: duplicate of  MARATHON_PORT
MARATHON_APP_ID: marathon specific

paasta_tools/docker_wrapper.py
```diff
@@ -341,6 +343,10 @@ def main(argv=None):
 
     if mesos_task_id and can_add_hostname(argv):
         hostname = generate_hostname(socket.getfqdn(), mesos_task_id)
+        # Add PAASTA_HOST environment variable
```
solarkennedy: Due to the complexity of how this works, and the fact that it is implemented so that it works only with the wrapper (and won't work in k8s), I vote to defer on `PAASTA_HOST` for now.

I like the other variables though!

paasta_tools/utils.py
```diff
@@ -545,7 +547,13 @@ def get_env_dictionary(self) -> Dict[str, str]:
             "PAASTA_CLUSTER": self.cluster,
             "PAASTA_DEPLOY_GROUP": self.get_deploy_group(),
             "PAASTA_DOCKER_IMAGE": self.get_docker_image(),
+            "PAASTA_APP_RESOURCE_CPUS": str(self.get_cpus()),
```
solarkennedy: `APP` was a marathon-specific thing. Can you do just `PAASTA_RESOURCE_X` ?

paasta_tools/cli/cmds/local_run.py
```diff
@@ -622,6 +622,8 @@ def get_local_run_environment_vars(instance_config, port0, framework):
...
```
drolando: ++

paasta_tools/kubernetes_tools.py
```diff
@@ -637,6 +637,12 @@ def get_kubernetes_environment(self) -> Sequence[V1EnvVar]:
                     field_ref=V1ObjectFieldSelector(field_path="metadata.name")
                 ),
             ),
+            V1EnvVar(
```
solarkennedy: Whoa! I didn't think this would work!
Huh!

paasta_tools/docker_wrapper.py
solarkennedy: Disregard, I didn't see the k8s part.

paasta_tools/docker_wrapper.py
```diff
@@ -341,6 +343,10 @@ def main(argv=None):
...
+        if env_file_name is not None:
```
solarkennedy: I thought the env_file_name was only used in the cases where we have secrets. 

Instead of appending to this file, can you instead "just" add an additional env arg?

paasta_tools/cli/cmds/local_run.py
drolando: Actually, what's the content of PAASTA_HOST gonna be? The hostname of the physical machine?
Or the hostname that we use currently which is the concatenation of the machine hostname and some other sha?

EmanekaT: Nice catch. I thought the hostname here is the real one.

drolando: Yeah the one that I want is the real one. There's already the HOST variable that contains the concatenation.

EmanekaT: I added PAASTA_GIT_SHA, which parses docker url to get the git-sha using 
 def get_code_sha_from_dockerurl(docker_url: str) -> str:
```
     """We encode the sha of the code that built a docker image *in* the docker
     url. This function takes that url as input and outputs the partial sha
     """
     try:
         parts = docker_url.split("/")
         parts = parts[-1].split("-")
         return "git%s" % parts[-1][:8]
     except Exception:
         return "gitUNKNOWN"
```

It's hard to add container id since we cannot know it before creating the container with docker_run.

paasta_tools/utils.py
```diff
@@ -556,6 +566,9 @@ def get_env_dictionary(self) -> Dict[str, str]:
         env.update(user_env)
         return {str(k): str(v) for (k, v) in env.items()}
 
+    def get_container_port(self) -> int:
+        return self.config_dict.get("container_port", DEFAULT_CONTAINER_PORT)
+
```
EvanKrall: Why move this from LongRunningServiceConfig?

My understanding is that we only allocate ports for Marathon+K8s tasks (which both subclass LongRunningServiceConfig) -- e.g. Tron actions don't have ports.

paasta_tools/cli/cmds/local_run.py
drolando: > It's hard to add container id since we cannot know it before creating the container with docker_run.

Oh yeah good point, didn't think of that. Just ignore it then

paasta_tools/utils.py
EmanekaT: I wanted to use get_container_port in InstanceConfig, and it's not possible if it's in LongRunningServiceConfig. I didn't know that Tron doesn't use container_port though. Let me move it back.

EmanekaT: But we'll have two DEFAULT_CONTAINER_PORT in marathon and k8s. What about just leave it in get_env? I don't think Tron cares about having an extra envvar.

EmanekaT: hmm, let me make it class variable then.

paasta_tools/utils.py
```diff
@@ -551,8 +551,10 @@ def get_env_dictionary(self) -> Dict[str, str]:
         }
         if self.get_gpus() is not None:
             env["PAASTA_RESOURCE_GPUS"] = str(self.get_gpus())
-        if self.get_docker_image():
+        try:
             env["PAASTA_GIT_SHA"] = get_git_sha_from_dockerurl(self.get_docker_url())
+        except Exception:
```
EvanKrall: is there a more specific exception that you can catch?

EmanekaT: No. This is only to work around the failed travis test, where git_registry doesn't exist. Docker image could be missing as well in some test cases. 
This is not intended to catch any errors in practice. And we don't really need to catch the error if we can't get dockerurl. The service wouldn't even run.

PR #2553, opened 2019-11-14
Full hostname in oom kill log lines

thebostik:
Write out the value of PAASTA_HOST for an OOM-killed log's value of hostname if we have it.
Depends on https://github.com/Yelp/paasta/pull/2552.

### DIFF
--- a/tests/test_oom_logger.py
+++ b/tests/test_oom_logger.py
@@ -28,10 +28,10 @@ from paasta_tools.oom_logger import main
 def sys_stdin():
     return [
         "some random line1\n",
-        "1500316299 dev37-devc [30533610.306528] apache2 invoked oom-killer: "
+        "1500316299 dev42-uswest1adevc [30533610.306528] apache2 invoked oom-killer: "
         "gfp_mask=0x24000c0, order=0, oom_score_adj=0\n,"
         "some random line2\n",
-        "1500316300 dev37-devc [30533610.306529] Task in "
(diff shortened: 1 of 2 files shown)


tests/test_oom_logger.py
```diff
@@ -86,7 +86,11 @@ def sys_stdin_without_process_name():
 def docker_inspect():
     return {
         "Config": {
-            "Env": ["PAASTA_SERVICE=fake_service", "PAASTA_INSTANCE=fake_instance"]
+            "Env": [
+                "PAASTA_SERVICE=fake_service",
+                "PAASTA_INSTANCE=fake_instance",
+                "PAASTA_HOST=dev42-uswest1adevc-64e7207e-f738-4236-b0c5-8f4fc5e0adbd",
```
drolando: Are you sure about this example? I thought the new PAASTA_HOST variable would only have the hostname in it

EmanekaT: I upgraded PAASTA_HOST to use box's hostname

thebostik: Ok, oops that was a misunderstanding of that change, in that case. I am just going to grab `HOSTNAME` from inside the container, which is currently what gets logged in the logs we need to join to. hostname from outside the container is the fallback, which doesn't have the mesos task id information that is logged from the service (to identify which is the OOM'ing instance on a host).

solarkennedy:
@thebostik is this PR still needed?

thebostik:
Thanks for the poke Kyle, just discarding this.

PR #2559, opened 2019-11-15
Fix times in paasta status for k8s; fix timestamps in CLI for marathon

EvanKrall:
Basically, the paasta API was outputting timestamps for Marathon tasks/apps that were off by 7 or 8 hours; we had code in the CLI to correct this. The Paasta API was producing correct timestamps for k8s, but we copy-pasted the same correction code from the Marathon status CLI, which resulted in incorrect timestamps.

The first commit fixes the k8s bug; the second commit fixes the paasta API for marathon tasks and removes the correction; it does so by bumping marathon-python to a version that [produces TZ-aware datetime objects](https://github.com/thefactory/marathon-python/pull/267)

### DIFF
--- a/paasta_tools/check_marathon_services_replication.py
+++ b/paasta_tools/check_marathon_services_replication.py
@@ -34,6 +34,7 @@ import argparse
 import logging
 from datetime import datetime
 from datetime import timedelta
+from datetime import timezone
 
 import a_sync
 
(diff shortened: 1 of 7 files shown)


paasta_tools/check_marathon_services_replication.py
```diff
@@ -77,14 +77,14 @@ def filter_healthy_marathon_instances_for_short_app_id(all_tasks, app_id):
     tasks_for_app = [
         task for task in all_tasks if task.app_id.startswith("/%s" % app_id)
     ]
-    one_minute_ago = datetime.now() - timedelta(minutes=1)
+    one_minute_ago = datetime.now(timezone.utc) - timedelta(minutes=1)
```
solarkennedy: You could do utcnow if you want

EvanKrall: nope, utcnow returns a naive datetime. The [datetime docs](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) specifically recommend this construction instead:

> *Warning*: Because naive `datetime` objects are treated by many `datetime` methods as local times, it is preferred to use aware datetimes to represent times in UTC. As such, the recommended way to create an object representing the current time in UTC by calling `datetime.now(timezone.utc)`.

solarkennedy: Haha. Jeeze :(

PR #2564, opened 2019-11-21
Report evicted pods when running paasta status on flink resources

gvolco:
The Flink operator is already returning a pod_status list. 
This change incorporates that info into `paasta status`.

Here is how the status would look like:
```
service: xxxxx
cluster: xxxxx
    instance: main_k8s_copy
    Git sha:    97cd8b76 (desired)
    Config SHA: 2c215c54
    Flink version: 1.6.4
    URL: http://xxxxx/
    State: running
    Pods: 7 running, 0 evicted, 0 other
    Jobs: 1 running, 0 finished, 0 failed, 0 cancelled
    5 taskmanagers, 1/5 slots available
    Jobs:
      Job Name                     State       Started
      xxxxxxx                      RUNNING     2019-11-21 05:51:40 (4 hours ago)
```
Running in verbose mode would produce the following pod list:
```
    Pods:
      Pod Name      Host                         Phase
      xxxxxx        ip-x-x-x-x.compute.internal  Running
      xxxxxx        ip-x-x-x-x.compute.internal  Running
      xxxxxx        ip-x-x-x-x..compute.internal Running
```

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -33,6 +33,7 @@ from typing import Sequence
 from typing import Set
 from typing import Tuple
 from typing import Type
+from typing import Union
 
 import humanize
 from bravado.exception import BravadoConnectionError
@@ -838,6 +839,26 @@ def print_flink_status(
         output.append(f"    No other information available in non-running state")
         return 0
 
+    pod_running_count = pod_evicted_count = pod_other_count = 0
+    for pod in status.pod_status:
+        if pod["phase"] == "Running":
+            pod_running_count += 1
+        elif pod["phase"] == "Failed" and pod["reason"] == "Evicted":
+            pod_evicted_count += 1
+        else:
+            pod_other_count += 1
+        evicted = (
+            PaastaColors.red(f"{pod_evicted_count}")
+            if pod_evicted_count > 0
+            else f"{pod_evicted_count}"
+        )
+    output.append(
+        "    Pods:"
+        f" {pod_running_count} running,"
+        f" {evicted} evicted,"
+        f" {pod_other_count} other"
+    )
+
     output.append(
         "    Jobs:"
         f" {status.overview['jobs-running']} running,"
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -838,6 +839,21 @@ def print_flink_status(
...
+        if pod["phase"] == "Running":
+            pod_running_count += 1
+        elif pod["phase"] == "Failed" and pod["reason"] == "Evicted":
+            pod_evicted_count += 1
+        else:
+            pod_other_count += 1
+    output.append(
+        "    Pods:"
+        f" {pod_running_count} running,"
+        f" {pod_evicted_count} evicted,"
```
vkhromov: this one should probably be printed in the red color if the count of evicted pods is greater than 0?

gvolco: Yeah. I didn't add colors because `paasta status` for flink doesn't use any colors at all. But I guess it won't hurt to add some. Will update the PR in a moment!

AshishKhatkar: @gvolco paasta status for flink uses color when service is not running. I agree with @vkhromov using color will be a better idea.

paasta_tools/cli/cmds/status.py
```diff
@@ -838,6 +839,21 @@ def print_flink_status(
         output.append(f"    No other information available in non-running state")
         return 0
 
+    pod_running_count = pod_evicted_count = pod_other_count = 0
+    for pod in status.pod_status:
+        if pod["phase"] == "Running":
+            pod_running_count += 1
+        elif pod["phase"] == "Failed" and pod["reason"] == "Evicted":
```
AshishKhatkar: Any reason why we don't want to show failed and evicted separately. In verbose mode you are showing failed and evicted as separate. Might be misleading if we see evicted here and failed in verbose description.

gvolco: When a Pod is evicted, `phase` will be 'Failed', and `reason` will be 'Evicted'.
On verbose mode, the list will look like this:
```
    Pods:
      Pod Name      Host                         Phase
      xxxxxx        ip-x-x-x-x..compute.internal Running
      xxxxxx        ip-x-x-x-x.compute.internal  Running
      xxxxxx        ip-x-x-x-x.compute.internal  Failed
        Evicted: Disk quota exceeded
      xxxxxx        ip-x-x-x-x..compute.internal Running
      xxxxxx        ip-x-x-x-x..compute.internal Running
      xxxxxx        ip-x-x-x-x..compute.internal Running
```

PR #2569, opened 2019-11-26
add command for adhoc spark run

drmorr0:
I kinda want to add a couple tests to this to make sure it's doing the right thing for commands we expect, but I didn't get around to it today.  When I get back, i guess.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -117,7 +117,7 @@ def add_subparser(subparsers):
         "-i",
         "--instance",
         help=("Start a docker run for a particular instance of the service."),
-        default="client",
+        default="adhoc",
     ).completer = lazy_choices_completer(list_instances)
 
     try:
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -332,13 +332,14 @@ def get_docker_run_cmd(container_name, volumes, env, docker_img, docker_cmd, nvi
...
-def get_smart_paasta_instance_name(instance):
+def get_smart_paasta_instance_name(instance, cmd):
     if os.environ.get("TRON_JOB_NAMESPACE"):
         tron_job = os.environ.get("TRON_JOB_NAME")
         tron_action = os.environ.get("TRON_ACTION")
         return f"{tron_job}.{tron_action}"
     else:
-        return f"{instance}_{get_username()}"
+        how_submitted = cmd.split(" ")[0]
+        return f"{instance}_{get_username()}_{how_submitted}"
```
stug: The one case this doesn't handle is the mrjob spark runner.  See for instance https://sourcegraph.yelpcorp.com/sysgit/yelpsoa-configs@ad7081eeafa1119137db6def88b8097df87e9626/-/blob/ad_creative/tron-norcal-prod.yaml#L90 -- the command is just `python -m ...`

I think the move for that case is just to look at the `--mrjob` arg they're passing.

PR #2575, opened 2019-12-03
Make a test to detect big bounces in advance.

EvanKrall:
We've wanted this for a while: https://jira.yelpcorp.com/browse/PAASTA-14168

Since we just surprised ourselves with a big bounce, maybe now is a good time to introduce this.


solarkennedy:
Hmm. I swear we used to have this.

tests/test_marathon_tools.py
```diff
@@ -2340,6 +2340,33 @@ def test_get_accepted_resource_roles(self):
         )
         assert marathon_config.get_accepted_resource_roles() == ["ads"]
 
+    def test_warning_big_bounce(self):
+        marathon_config = marathon_tools.MarathonServiceConfig(
```
solarkennedy: can you do k8s too?

PR #2577, opened 2019-12-04
Added k8s soa-configs docs

solarkennedy:


docs/source/yelpsoa_configs.rst
```diff
@@ -157,6 +157,169 @@ Marathon, Chronos, Tron, or ``paasta remote-run``.
...
+
+  * ``cmd``: The command that is executed. If a string, will be wrapped in ``/bin/sh -c``.
+    If a list, will be executed directly as is with no shell parsing.
+
+  * ``args``: An array of docker args if you use the `"entrypoint"
+    <https://docs.docker.com/reference/builder/#entrypoint>`_ functionality.
+
+  * ``monitoring``: See the `monitoring.yaml`_ section for details.
+
+  * ``autoscaling``: See the `autoscaling docs <autoscaling.html>`_ for valid options and how they work
```
EmanekaT: This would be different for Kubernetes and Marathon. I'll write another one for Kubernetes later. I guess it's OK to just leave it here for now.

docs/source/yelpsoa_configs.rst
```diff
@@ -157,6 +157,169 @@ Marathon, Chronos, Tron, or ``paasta remote-run``.
...
+    * ``haproxy_min_fraction_up``: if ``check_haproxy`` is True, we check haproxy on up to 20 boxes to see whether a task is available.
+      This fraction of boxes must agree that the task is up for the bounce to treat a task as healthy.
+      Defaults to 1.0 -- haproxy on all queried boxes must agree that the task is up.
+
+  * ``bounce_margin_factor``: proportionally increase the number of old instances
+    to be drained when the crossover bounce method is used.
+    0 < bounce_margin_factor <= 1. Defaults to 1 (no influence).
+    This allows bounces to proceed in the face of a percentage of failures.
+    It doesn’t affect any other bounce method but crossover.
+    See `the bounce docs <bouncing.html>`_ for a more detailed description.
```
EmanekaT: I think this needs update as well. I deprecated upthedown but we need to update/rephrase here and there. For example, in "The bounce_margin_factor setting in marathon yaml files .",   we can use marathon/kubernetes instead. Also, we might want to remove chronos section? Leaving it here will make me feel that this doc is not trustworthy.

docs/source/yelpsoa_configs.rst
solarkennedy: I'm really hoping they can be the same

EmanekaT: @solarkennedy they will be the same for existing functions but I plan to add more functionalities that cannot really be supported by marathon autoscaler. Also, things like frequency of autoscaling is different and should be documented. Also, the architecture/algorithm is different and should be written down for those who are interested.

PR #2578, opened 2019-12-05
Use `KUBECONFIG` env var while checking availability of k8s

vkhromov:


vkhromov:
> i wonder if we should make a helper function for that, we already have the same bit of code in another place: https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes_tools.py#L1195

originally I planned to do
```
KUBE_CONFIG_PATH = os.environ.get("KUBECONFIG", "/etc/kubernetes/admin.conf")
```
and then just use `KUBE_CONFIG_PATH` everywhere, but Python allows you to change environment variables, so I discarded that idea.

PR #2579, opened 2019-12-06
Add additional authors as a CLI argument

dbgrigsby:
Not all git hosts are gitolite, and relying on the authors-of-changeset
gitolite script is not an option for repositories hosted there.  
Since most calls to this command are automated, Release
Engineering will be modifying those calls to pass in the git authors
directly. This change enables that.

For those manual calls by developers, we intend to pull out the username
of the developer from OS calls, which will be done in a future CR. 

This argument is an `append ` action, so authors will be passed in via `--author user1 --author user2` and so on.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -199,6 +199,13 @@ def add_subparser(subparsers):
         default=30,
         help="After noticing an SLO failure, wait this many seconds before automatically rolling back.",
     )
+    list_parser.add_argument(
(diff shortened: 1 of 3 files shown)


dbgrigsby:
I started writing a unit test for this change, but none of the tests in https://github.com/Yelp/paasta/blob/master/tests/cli/test_cmds_mark_for_deployment.py seem to be at that low of a level. Would appreciate opinions on if it's necessary or not.

dbgrigsby:
Discarded because this is a fork and not a branch on the repo, oops

PR #2580, opened 2019-12-06
Add additional authors as a CLI argument

dbgrigsby:
Not all git hosts are gitolite, and relying on the authors-of-changeset
gitolite script is not an option for repositories hosted there.
Since most calls to this command are automated, Release
Engineering will be modifying those calls to pass in the git authors
directly. This change enables that.

For those manual calls by developers, we intend to pull out the username
of the developer from OS calls, which will be done in a future CR.

This argument is an append action, so authors will be passed in via --author user1 --author user2 and so on.

I started writing a unit test for this change, but none of the tests in https://github.com/Yelp/paasta/blob/master/tests/cli/test_cmds_mark_for_deployment.py seem to be at that low of a level. Would appreciate opinions on if it's necessary or not.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -199,6 +199,13 @@ def add_subparser(subparsers):
         default=30,
         help="After noticing an SLO failure, wait this many seconds before automatically rolling back.",
     )
+    list_parser.add_argument(
+        "--author",
+        dest="additional_authors",
+        default=None,
+        action='append',
+        help="Additional author(s) of the deploy, who will be pinged in Slack"
+    )
 
     list_parser.set_defaults(command=paasta_mark_for_deployment)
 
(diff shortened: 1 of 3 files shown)


paasta_tools/remote_git.py
```diff
@@ -117,7 +117,7 @@ def get_authors(git_url, from_sha, to_sha):
             f"could not understand the git repo in {git_url} for authors detection",
         )
 
-    if "yelpcorp.com" in git_server:
+    if "git.yelpcorp.com" in git_server:
```
nemacysts: could this be an  `==` check now? if I'm reading the regex correctly, `git_server` will always be `{user}@{githost}`

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -253,21 +260,27 @@ def report_waiting_aborted(service, deploy_group):
...
         git_url=git_url, from_sha=from_sha, to_sha=to_sha
     )
-    if ret == 0:
-        if authors == "":
-            return ""
-        else:
-            slacky_authors = ", ".join({f"<@{a}>" for a in authors.split()})
-            log.debug(f"Authors: {slacky_authors}")
-            return f"^ {slacky_authors}"
+    if ret == 1:
```
nemacysts: we should probably add some comments here about why it's fine for this to fail/that we expect this to start failing and additionally create an internal ticket to clean up this logic once we no longer hit any git hosts that call `authors-of-changeset`

paasta_tools/remote_git.py
dbgrigsby: There's also sysgit, so this covers all gitolite hosts. We could explicitly match for those two urls though

paasta_tools/cli/cmds/mark_for_deployment.py
dbgrigsby: Added a comment!

solarkennedy: I think this is fine. What do you think about just making it so we are only using one *or* the other?

And only if author is not provided on the cli do we try to fetch it.

paasta_tools/remote_git.py
solarkennedy: I bet it has `git@ ` and other nonsense.

solarkennedy: We don't have any paasta services hosted on sysgit, shouldn't need to add that there.

paasta_tools/cli/cmds/mark_for_deployment.py
dbgrigsby: @solarkennedy I like that a lot!

Writing it that way feels more useful to a developer triggering the deployment manually. Allowing a dev manually deploying to override who is notified for a deployment is a nice quality of life. 

Additionally, this allows us to continue treating a non-zero return from `remote_git.get_authors` as a failure, instead of swallowing it in some cases. 

I did a quick [internal poll](https://yelp.slack.com/archives/CMSDQP0K1/p1575934938401500)on preferring override vs append, and the result was well in favor of append. 

I've updated the PR with that change.

solarkennedy: Seems pretty reasonable to me for a cli argument. "tell, don't ask" kinda thing. (we are telling it which authors to use, idk)

dbgrigsby:
Amended my previous commit to fix a linter check, now merging.

PR #2581, opened 2019-12-06
More consistent paasta prefixed attributes

keymone:
* rename `maybe_add_yelp_prefix` with `paasta_prefixed`
* fix `paasta_prefixed` for discovery attributes, that aren't under `paasta.yelp.com` namespace
* replace hardcoded namespacing with `paasta_prefixed` everywhere

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -136,6 +136,7 @@ KUBE_DEPLOY_STATEGY_MAP = {
 }
 HACHECK_POD_NAME = "hacheck"
 KUBERNETES_NAMESPACE = "paasta"
+DISCOVERY_ATTRIBUTES = {"region", "superregion", "ecosystem", "habitat"}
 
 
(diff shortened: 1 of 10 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1519,14 +1520,21 @@ def filter_nodes_by_blacklist(
     ]
 
 
-def maybe_add_yelp_prefix(attribute: str,) -> str:
-    return YELP_ATTRIBUTE_PREFIX + attribute if "/" not in attribute else attribute
+def paasta_prefixed(attribute: str,) -> str:
+    # discovery attributes are exempt for now
+    discovery_attributes = ["region", "superregion", "ecosystem", "habitat"]
```
vkhromov: `paasta_prefixed()` is called a lot, `discovery_attributes` is initialized on every call, and then scanned linearly.
Maybe convert the list into the set and make `discovery_attributes` a global variable?

PR #2582, opened 2019-12-06
validate new autoscaling configurations

EmanekaT:
Design a new autoscaling configuration before implementing it. Only validation part is added and fully tested now.
 **Purpose**
* Add support for external metrics based on any signalfx metrics, multiple metrics, and memory. 
* Simplify existing autoscaling configuration by removing unused parameters. 
* Make the new configuration as simple as possible and as powerful as possible, while restrict potential intended user mistakes. 

**Complex User cases**
*  Scale instance A based on instance B
*  Use the same signalfx metrics for multiple instances across different clusters. 
*  Set canary to a certain percent of non-canary by using the number of instances of non-canary as metrics. 

I believe all of above use cases could be supported by allowing users to customize the dimension of their metrics. They can either user http metrics collector and override its metrics, or sent their own metrics to signalfx. Adding dimension parameter to http/uwsgi is mostly for users who have already implemented http metrics to use advanced autoscaling without modifying existing code. For new users, it's probably better to define and implement their own collector instead. Writing their own collector is not harder than exposing http metrics anyway.

**Potential issues** 
*  Users might corrupt other metrics.
* SignalFX queries for custom metrics is not validated since I can't find appropriate library for this.
* It could be complex for users to configure complex autoscaling behavior. For use cases like canary percentage, perhaps it is worth adding another configuration. But that would increase the complexity of implementation and configuration.

### DIFF
--- a/general_itests/fake_soa_configs_validate/fake_invalid_service/kubernetes-test-cluster.yaml
+++ b/general_itests/fake_soa_configs_validate/fake_invalid_service/kubernetes-test-cluster.yaml
@@ -8,6 +8,8 @@ main2:
     FOO: BAR
   deploy_group: fake_deploy_group
   horizontal_autoscaling:
+      max_replicas: 3
+      max_replicas: 1
       cpu:
           target_average_value: 70
           signalflow_metrics_query: "data(\"fake-external\").publish"
--- a/general_itests/fake_soa_configs_validate/fake_invalid_service/kubernetes-test-cluster.yaml
+++ b/general_itests/fake_soa_configs_validate/fake_invalid_service/kubernetes-test-cluster.yaml
@@ -1,0 +1,16 @@
+---
+main2:
+  cpus: .1
+  mem: 100
+  disk: 200.0
+  instances: 1
+  env:
+    FOO: BAR
+  deploy_group: fake_deploy_group
+  horizontal_autoscaling:
+      cpu:
+          target_average_value: 0.7
+          signalflow_metrics_query: "data(\"fake-external\").publish"
+      fake-external:
+          target_average_value: 0.55
+          signalflow_metrics_query: "data(\"fake-external\").publish"
--- a/general_itests/fake_soa_configs_validate/fake_invalid_service/kubernetes-test-cluster.yaml
+++ b/general_itests/fake_soa_configs_validate/fake_invalid_service/kubernetes-test-cluster.yaml
@@ -9,8 +9,8 @@ main2:
   deploy_group: fake_deploy_group
   horizontal_autoscaling:
       cpu:
-          target_average_value: 0.7
+          target_average_value: 70
           signalflow_metrics_query: "data(\"fake-external\").publish"
       fake-external:
-          target_average_value: 0.55
+          target_average_value: 55
           signalflow_metrics_query: "data(\"fake-external\").publish"
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -194,9 +194,11 @@
                 },
                 "horizontal_autoscaling": {
                     "type": "object",
+                    "additionalProperties": false,
                     "properties": {
                         "cpu": {
                             "type": "object",
+                            "additionalProperties": false,
                             "properties": {
                                 "target_average_value": {
                                     "type": "number",
@@ -206,10 +208,14 @@
                                     "exclusiveMinimum": true,
                                     "exclusiveMaximum": false
                                 }
-                            }
+                            },
+                            "required": [
+                                "target_average_value"
+                            ]
                         },
                         "memory": {
                             "type": "object",
+                            "additionalProperties": false,
                             "properties": {
                                 "target_average_value": {
                                     "type": "number",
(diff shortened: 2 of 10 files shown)


EmanekaT:
Added kyle just in case if you have any opinions on the new design.

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -194,9 +194,11 @@
                 },
                 "horizontal_autoscaling": {
                     "type": "object",
+                    "additionalProperties": false,
```
solarkennedy: Can you confirm that every existing kubernetes.yaml file conforms to this new schema?

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -263,13 +277,16 @@
...
-                            }
+                            },
+                            "required": [
+                                "target_average_value"
+                            ]
                         }
                     },
                     "patternProperties": {
-                        "^[a-z]([-a-z0-9]*[a-z0-9])?$": {
+                        "(?!(^cpu|memory|http|uwsgi)$)(^[a-z]([-a-z0-9]*[a-z0-9])?$)": {
```
solarkennedy: This regext looks pretty complex.
Can you say in words what this (should) represents?

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -206,10 +208,14 @@
                                     "exclusiveMinimum": true,
                                     "exclusiveMaximum": false
                                 }
-                            }
+                            },
+                            "required": [
+                                "target_average_value"
```
solarkennedy: Is this a new requirement? 

What if instead of requiring a value we provided a sane default for users?

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -282,7 +299,10 @@
                                 "signalflow_metrics_query": {
                                     "type": "string"
                                 }
-                            }
+                            },
+                            "required": [
+                                "signalflow_metrics_query"
```
solarkennedy: Have you and @EvanKrall already come to an agreement that this is the expected user interface for defining "custom" autoscaling stuff?

paasta_tools/cli/schemas/kubernetes_schema.json
EmanekaT: Yes they are. And the old schema will still be supported until we move them to this new one.

paasta_tools/cli/schemas/kubernetes_schema.json
EmanekaT: The first part exclude the keywords from patterned properties. The second half is copied from kubernetes documentation. That is metrics name.

paasta_tools/cli/schemas/kubernetes_schema.json
EmanekaT: Sure I can remove it.

paasta_tools/cli/schemas/kubernetes_schema.json
EmanekaT: Yes. We reached an agreement

paasta_tools/cli/schemas/kubernetes_schema.json
EmanekaT: I'll add this to paasta docs once I finish. I can't add comments to this json (I'm happy about this somehow).

paasta_tools/cli/schemas/kubernetes_schema.json
EmanekaT: @solarkennedy after rethinking, I think perhaps we should add it back as required? It is not something people can easily forget since it's the only parameter for cpu/memory metrics. It only needs to be set once, but will make the configuration more readable. People need to check paasta doc everytime if it is hidden.

solarkennedy: It is certainly a tradeoff between making "hidden" defaults and making users specify all the things.

Please get a second opinion from @EvanKrall

EmanekaT:
I will test this later. Everything else should be fully implemented. HPAMetrics Fullerite collector will be updated after this is shipped.

general_itests/fake_soa_configs_validate/fake_invalid_service/kubernetes-test-cluster.yaml
```diff
@@ -0,0 +1,16 @@
...
+  mem: 100
+  disk: 200.0
+  instances: 1
+  env:
+    FOO: BAR
+  deploy_group: fake_deploy_group
+  horizontal_autoscaling:
+      cpu:
+          target_average_value: 0.7
+          signalflow_metrics_query: "data(\"fake-external\").publish"
```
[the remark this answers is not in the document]
EmanekaT: Yep, this should fail .

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -192,6 +192,128 @@
...
+                        },
+                        "cpu": {
+                            "type": "object",
+                            "additionalProperties": false,
+                            "properties": {
+                                "target_average_value": {
+                                    "type": "integer",
+                                    "default": 70,
+                                    "minimum": 0,
+                                    "maximum": 100,
```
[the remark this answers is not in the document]
EmanekaT: I changed it from decimal to integer, but forgot to update tests.

PR #2584, opened 2019-12-10
specify sidecar resource requirements

chlgit:
How sophisticated we need here? I am not sure. We have a plan to retire hacheck sidecar, and it's the only sidecar we use now.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -569,6 +569,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
                             )
                         )
                     ),
+                    resources=self.get_sidecar_resource_requirements(),
                     name=HACHECK_POD_NAME,
                     env=self.get_kubernetes_environment(),
                     ports=[V1ContainerPort(container_port=6666)],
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -678,6 +679,12 @@ def get_resource_requirements(self) -> V1ResourceRequirements:
             },
         )
 
+    def get_sidecar_resource_requirements(self) -> V1ResourceRequirements:
+        return V1ResourceRequirements(
+            limits={"cpu": 0.1, "memory": "1024Mi", "ephemeral-storage": "1024Mi"},
```
vkhromov: `ephemeral-storage` better should be set to a real value, as k8s will reserve these `1024Mi` at a node and will not schedule anything there if there will be not enough `ephemeral-storage` there.
E.g. if the sidecar doesn't use the local disk for temporary files then `ephemeral-storage` should be `0Mi`.

chlgit: Ah, I see.

qui: I think hacheck needs a little bit of local disk, for writing files to /var/spool during draining, but probably a very small amount.

chlgit: Let's start from 256Mi for the disk and revisit it if needed.

PR #2585, opened 2019-12-11
replace jupyter with jupyter-lab in spark-run

huadongliu:
This PR enables users to have the jupyterhub experience on their local boxes for whatever reason. It opens additional opportunities since /nail/home is mounted (required for aws creds). Users are expected to use images with the yelp_jupyterhub package, for example, the docker-dev.yelpcorp.com/core-ml-centralized-jupyterhub-gpu image.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -166,7 +166,7 @@ def add_subparser(subparsers):
     list_parser.add_argument(
         "-C",
         "--cmd",
-        help="Run the spark-shell, pyspark, spark-submit, jupyter, or history-server command.",
+        help="Run the spark-shell, pyspark, spark-submit, jupyter-lab, or history-server command.",
     )
 
     list_parser.add_argument(
@@ -306,7 +306,7 @@ def get_docker_run_cmd(container_name, volumes, env, docker_img, docker_cmd, nvi
 
     sensitive_env = {}
 
-    non_interactive_cmd = ["spark-submit", "jupyter", "history-server"]
+    non_interactive_cmd = ["spark-submit", "history-server"]
     if not any(c in docker_cmd for c in non_interactive_cmd):
         cmd.append("--interactive=true")
         if sys.stdout.isatty():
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -397,10 +397,10 @@ def get_spark_env(args, spark_conf, spark_ui_port, access_key, secret_key):
...
 
     # Default configs to start the jupyter notebook server
-    if args.cmd == "jupyter":
-        dirs = args.work_dir.split(":")
-        spark_env["JUPYTER_RUNTIME_DIR"] = dirs[1] + "/.jupyter"
-        spark_env["JUPYTER_DATA_DIR"] = dirs[1] + "/.jupyter"
+    if args.cmd == "jupyter-lab":
+        spark_env["JUPYTER_RUNTIME_DIR"] = "/source/.jupyter"
+        spark_env["JUPYTER_DATA_DIR"] = "/source/.jupyter"
+        spark_env["JUPYTER_CONFIG_DIR"] = "/source/.jupyter"
```
lydian: This will make jupyterlab use /source/.jupyter/jupyter_notebook_config.py instead of the one they have in their homedir.  Our settings for /source/.jupyter/jupyter_notebook_config.py will be using the s3 for notebooks instead of their local. Do you think it is what those user expected?

huadongliu: One of the goals is to have a close to jupyterhub experience. Yes, notebooks in S3 are expected. Users of this command are expected to have a jupyterhub ready repo by installing the yelp-jupyterhub package. It could be error prone for users to install things by themselves.

lydian: another issue I can think of is we need to ask user to install the kernel from their home dir, otherwise, the default one should be the one in docker images, and therefore the changes they made on their devbox still won't appear on the notebooks.  I am wondering if we should call the `<their venv>/bin/python -m ipykernel install --name application --display-name local-repo --prefix=/source/virtualenv_run_jupyter`  for those users, or we should having a session to ask them to run this or include this in their dockerfile.

huadongliu: I would leave it to the user since usage of the command is expected to be rare.

lydian: agree, as long as we don't advertise people to use this.  In general, i would still prefer people use centralized jupyterhub, and this is just a backup plan.

PR #2586, opened 2019-12-11
Bump slo transcoder dependency

matfra:
Lots of improvements to slo transcoder including schema changes that are required to validate SLOs for newly supported services


EvanKrall:
May be worth bumping https://github.com/Yelp/sticht/blob/master/setup.py as well

PR #2587, opened 2019-12-12
Check services replication metrics

stug:
first commit unifies the two scripts for monitoring replication in mesos/marathon and kubernetes

second commit adds logic to emit metrics about the replication of each service and also track replication in aggregate in a cluster so that we can alert when there are widespread problems with the service mesh (instead of waiting for folks to complain to us)

I have already run all of these scripts on the norcal-devc masters and they work and emit metrics.

### DIFF
--- a/paasta_tools/check_services_replication_tools.py
+++ b/paasta_tools/check_services_replication_tools.py
@@ -13,23 +13,37 @@
 # limitations under the License.
 import argparse
 import logging
+from typing import Any
 from typing import Callable
+from typing import List
 from typing import Sequence
+from typing import Tuple
 from typing import Type
+from typing import Union
 
+import a_sync
+from marathon import MarathonClient
+from marathon.models.task import MarathonTask
 from mypy_extensions import Arg
 
 from paasta_tools.kubernetes_tools import get_all_nodes
 from paasta_tools.kubernetes_tools import get_all_pods
 from paasta_tools.kubernetes_tools import KubeClient
+from paasta_tools.kubernetes_tools import V1Node
 from paasta_tools.kubernetes_tools import V1Pod
+from paasta_tools.marathon_tools import get_marathon_clients
+from paasta_tools.marathon_tools import get_marathon_servers
+from paasta_tools.mesos_tools import get_slaves
 from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.smartstack_tools import KubeSmartstackReplicationChecker
+from paasta_tools.smartstack_tools import MesosSmartstackReplicationChecker
+from paasta_tools.smartstack_tools import SmartstackReplicationChecker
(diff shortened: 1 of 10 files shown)


paasta_tools/check_services_replication_tools.py
```diff
@@ -54,6 +74,22 @@ def parse_args() -> argparse.Namespace:
...
+        "the script will return a warning status",
+    )
+    parser.add_argument(
+        "--crit",
+        dest="under_replicated_crit_pct",
+        type=float,
+        default=10,
+        help="The percentage of under replicated service instances past which "
+        "the script will return a critical status",
+    )
```
stug: I would be curious to hear opinions on sane defaults for these.

tests/test_check_kubernetes_services_replication.py
```diff
@@ -103,69 +103,14 @@ def test_check_service_replication_for_non_smartstack(instance_config):
...
-            all_pods=[],
+            all_tasks_or_pods=[],
             smartstack_replication_checker=None,
         )
         mock_check_healthy_kubernetes_tasks.assert_called_once_with(
             instance_config=instance_config, expected_count=100, all_pods=[]
         )
 
 
-def test_check_all_kubernetes_services_replication(instance_config):
```
stug: these tests moved to tests/test_check_services_replication_tools.py

paasta_tools/check_services_replication_tools.py
EvanKrall: idk, 30% / 50%? Probably depends on the service. Would it be annoying for it to be configurable?

stug: to clarify, this check is for _all_ instances on a cluster.  So it's a question of what percentage of service instances are failing their replication checks.

paasta_tools/check_services_replication_tools.py
```diff
@@ -97,32 +128,118 @@ def check_all_kubernetes_based_services_replication(
...
+    if mesos:
+        tasks_or_pods, slaves = get_mesos_tasks_and_slaves(system_paasta_config)
+        replication_checker = MesosSmartstackReplicationChecker(
+            mesos_slaves=slaves, system_paasta_config=system_paasta_config,
+        )
+    else:
+        tasks_or_pods, nodes = get_kubernetes_pods_and_nodes(namespace)
+        replication_checker = KubeSmartstackReplicationChecker(
+            nodes=nodes, system_paasta_config=system_paasta_config,
+        )
```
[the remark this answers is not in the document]
stug: Yes good point.  And in fact all that really needs to be passed in is the replication checker and the pods/tasks.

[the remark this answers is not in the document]
stug: We talked offline, but to summarize: this turned out to be more trouble than it was worth, unfortunately.

PR #2590, opened 2019-12-13
Instance support modules

keymone:
Paasta codebase is littered with long `if ... elif` chains switching on `instance_type` values. I'm trying to make instance type more of a first-class citizen by moving everything related to how certain instance types are handled into their own modules. Currently we can largely categorize paasta codebase in following big categories:

a) api handles request/response and how things are represented on the wire
b) cli handles commands, args parsing and rendering to terminal
c) various `_tools` and other modules dealing with concrete tech (mesos, kubernetes, tron)

In this PR I'm adding `paasta_tools.instance` layer that should sit in between [ab] and [c].

Eventually it's useful to have a more pluggable system where modules or other binaries can be registered in paasta configuration system to "handle" certain instance_types, but we need to figure out the interface between these components first.

### DIFF
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -31,9 +31,6 @@ from typing import Tuple
 
 import a_sync
 import isodate
-from kubernetes.client import V1Pod
-from kubernetes.client import V1ReplicaSet
-from kubernetes.client.rest import ApiException
 from marathon import MarathonClient
 from marathon.models.app import MarathonApp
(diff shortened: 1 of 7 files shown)


EmanekaT:
Looks great! I personally think it would be even better if we have an object for each instance type put the individual status methods there with the same name. So we can 
1. cache api calls like pods_for_service_instance. We shouldn't request this again if it was just requested 2s ago. 
2. c and paasta_tools.instance are in the same class/object so they can share the same code easily.

PR #2591, opened 2019-12-13
fix make itest

chlgit:


solarkennedy:
Is there something non-deterministic about this sha for k8s? What was the previous change that triggered this to change and start failing?

chlgit:
I merged this branch for @mattmb https://github.com/Yelp/paasta/pull/2481

PR #2593, opened 2019-12-13
Support spark args with spaces in them

brenns10:
This addresses PAASTA-16138. Spark arguments which contain spaces will get split improperly, and even if they were split correctly, they will be joined again improperly when set in `SPARK_OPTS`. This patch uses shlex to split and quote arguments correctly. It only works on Python 3.3 and above, due to the use of `shlex.quote()`. Not sure if that will be a problem here.

Tested this in dev by invoking a spark job with spark args in spaces. I checked in the spark UI and saw that the value was successfully passed in.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -2,6 +2,7 @@ import argparse
 import json
 import logging
 import os
+import shlex
 import socket
 import sys
 import time
@@ -519,7 +520,7 @@ def get_spark_config(
         non_user_args["spark.mesos.uris"] = "file:///root/.dockercfg"
 
     if args.spark_args:
-        spark_args = args.spark_args.split()
+        spark_args = shlex.split(args.spark_args)
         for spark_arg in spark_args:
             fields = spark_arg.split("=", 1)
             if len(fields) != 2:
@@ -608,7 +609,8 @@ def create_spark_config_str(spark_config_dict, is_mrjob):
         spark_config_entries.append(f"--spark-master={spark_master}")
 
     for opt, val in spark_config_dict.items():
-        spark_config_entries.append(f"{conf_option} {opt}={val}")
+        quoted_opt = shlex.quote(f"{opt}={val}")
+        spark_config_entries.append(f"{conf_option} {quoted_opt}")
     return " ".join(spark_config_entries)
 
 
--- a/tests/cli/test_cmds_spark_run.py
+++ b/tests/cli/test_cmds_spark_run.py
@@ -163,6 +163,8 @@ class TestGetSparkConfig:
             setattr(args, k, v)
         if "cluster" not in paasta_args:
             args.cluster = "fake_cluster"
+        if "spark_args" not in paasta_args:
+            args.spark_args = ""
         return get_spark_config(
             args=args,
             spark_app_name="fake_name",
@@ -256,6 +258,26 @@ class TestGetSparkConfig:
                 assert key not in spark_conf
             else:
                 assert spark_conf[key] == expected_value
+
+    def test_parse_args_with_spaces(self):
+        spark_args = 'spark.executor.extraJavaOptions="-Xss4M -XX:Option=blah"'
+        expected = {
+            "spark.executor.extraJavaOptions": "-Xss4M -XX:Option=blah",
+        }
+        conf = self.get_spark_config({"spark_args": spark_args})
+        for key, value in expected.items():
+            assert conf[key] == value
+
+    def test_join_args_with_spaces(self):
+        spark_config_dict = {
+            "option1": "nospace",
+            "option2": "with spaces",
+        }
+
+        string = create_spark_config_str(spark_config_dict, False)
+
+        assert "--conf option1=nospace" in string
+        assert "--conf 'option2=with spaces'" in string
 
 
 @mock.patch("paasta_tools.cli.cmds.spark_run.get_aws_credentials", autospec=True)


stug:
paasta-tools uses python3.6, so I think that part of it should be fine.

EvanKrall:
**Preface**: I don't expect you, as someone whose day job isn't maintaining Paasta / this file, to make this file perfect. I just want to write them down, and maybe you can figure out a subset that's reasonable for you to do.

All the manipulation of the spark args in this file, scares me. Technically your change doesn't make this situation any worse -- you're just changing the method by which we first go from str to List[str]. `shlex` just is a code smell to me that makes me look for things like what I've describe below:

0. The user's shell splits the command they have typed, `paasta spark-run ...` into an array `["paasta", "spark-run", ...]` and passes that as `argv` to [one of the `exec` family of syscalls](http://man7.org/linux/man-pages/man3/exec.3.html).
1. argparse grabs one entry of `argv` and puts it in `args.spark_args`
2. [We split the string `args.spark_args` into multiple strings with `shlex.split`](https://github.com/Yelp/paasta/blob/a5d784551af0cea3ec643bbe610d08940c4590f4/paasta_tools/cli/cmds/spark_run.py#L523)
3. [We split those strings on = to get key/value](https://github.com/Yelp/paasta/blob/a5d784551af0cea3ec643bbe610d08940c4590f4/paasta_tools/cli/cmds/spark_run.py#L525), and [put those in a dictionary](https://github.com/Yelp/paasta/blob/a5d784551af0cea3ec643bbe610d08940c4590f4/paasta_tools/cli/cmds/spark_run.py#L546)
4. That [dictionary gets returned by `get_spark_config`](https://github.com/Yelp/paasta/blob/a5d784551af0cea3ec643bbe610d08940c4590f4/paasta_tools/cli/cmds/spark_run.py#L588) and then [passed to `create_spark_config_str`](https://github.com/Yelp/paasta/blob/a5d784551af0cea3ec643bbe610d08940c4590f4/paasta_tools/cli/cmds/spark_run.py#L761)
5. We [join the key and value with an equals sign](https://github.com/Yelp/paasta/blob/a5d784551af0cea3ec643bbe610d08940c4590f4/paasta_tools/cli/cmds/spark_run.py#L612), then use shlex.quote to prevent the shell from splitting on any spaces in step 9.
6. Then we [join all those quoted key=value arguments together with spaces](https://github.com/Yelp/paasta/blob/a5d784551af0cea3ec643bbe610d08940c4590f4/paasta_tools/cli/cmds/spark_run.py#L613) to form a single string
7. In `get_docker_cmd` we either:
  a. [Stick some prefix on the front](https://github.com/Yelp/paasta/blob/a5d784551af0cea3ec643bbe610d08940c4590f4/paasta_tools/cli/cmds/spark_run.py#L821)
  b. [Insert that string into the middle of another string](https://github.com/Yelp/paasta/blob/a5d784551af0cea3ec643bbe610d08940c4590f4/paasta_tools/cli/cmds/spark_run.py#L840-L842)
8. Then in [`get_docker_run_cmd` we make an array](https://github.com/Yelp/paasta/blob/a5d784551af0cea3ec643bbe610d08940c4590f4/paasta_tools/cli/cmds/spark_run.py#L332) `[<a bunch of docker run stuff>, "sh", "-c", <giant string constructed in previous steps>]`
9. Finally, `/bin/sh` inside the container will parse the massive string back into an array of arguments and execute it.

So we've gone from `str` to `List[str]` to `Dict[str, str]` to `List[str]` to `str`, then done some terrifying string munging, then eventually back to a List[str].

My preference would be that we stick with List[str] the entire way:
1. Use the `nargs='+'` or `action='append'` options to `argparse`, so that the script can just accept a List[str] on the command line. This would be a breaking interface change, so the prudent way to do it would be to make a second argument that supports the list, and keep `--spark-args` taking a single string that we split on our own.
1. Have `get_spark_config` return a List[str] instead of a Dict[str, str].
  a. This would mean dropping `shlex.quote` and `" ".join` in `create_spark_config_str`, and calling `create_spark_config_str` on the non-user options from within `get_spark_config.
1. Do terrifying array munging instead of terrifying string munging in `get_docker_cmd`
1. Drop `sh -c` in `get_docker_run_cmd`, and just concatenate the two List[str]s together.

brenns10:
I did feel a bit dirty/hacky throwing shlex in there. I was using this implementation in dev to enable some option I was testing out which happened to contain spaces. So I did the minimum change, and made a PR just because it worked. However I do care about the code quality here, and since you wrote your comment I've devoted a few small blocks of time to seeing what I could do to improve the situation.

I feel that there's some complexity missing from your overview. In particular, it's Step 7, the `get_docker_cmd()` step.

First, there's the case where we're not running a "real" command (i.e. when we're running jupyter lab). In this case, we have to [stuff the spark options into an environment variable](https://github.com/Yelp/paasta/blob/sbrennan_shlex_the_spark_args/paasta_tools/cli/cmds/spark_run.py#L398), and so we'd need to join them back up to a string anyway, using shlex to ensure spaces are preserved.

Second, there's the case where we're running a "real" command and we need to insert the spark options into that command. In this case, the command is already a string rather than a List[str], since it too was passed into Paasta as a single entry in argv (`paasta spark-run --cmd 'spark-submit foo.py --bar'`). To do things your preferred way, we would need to `shlex.split()` the application command to turn it into a List[str], so that we could splice the lists of arguments together, and run them without the `sh -c <big string>`. But that would not be strictly backwards compatible, since shlex doesn't do expansion, subshells, conditionals, etc. (Hopefully there are few or no commands relying on those features out there!)

If Paasta's CLI had been designed to avoid invoking the shell within the docker container (e.g. `paasta spark-run [paasta args] -- [application command]` rather than `paasta spark-run [paasta args] --cmd 'application shell command'`) then we would be able to simplify the code a fair amount and be confident that everything is compatible. However in the current state of things, I'm not sure that there's any real winning options:

1. Keep things the way they are in this PR, lots of string mangling. We use shlex on the `--spark-args` option which is pretty gross.
2. Prefer List[str] all the way. To do this properly we encounter all of the issues I described above.
3. Some compromise in between, where we deprecate `--spark-args` and pass each arg individually, relying on shell escaping. This eliminates using `shlex` for parsing, but we'd still need to use it to join the args together into the large string mostrosities which we end up executing or sticking into environment variables.

I'm willing to implement 1 or 3. 2 seems like either pretty sizable migration or a risky backwards-incompatible change, both of which I'd like to avoid. I don't know whether 3 is enough of an improvement to justify the changes, but I'm totally willing to do it.

brenns10:
Closing this out - I may have been the only one who wanted this improvement, and if it becomes a problem later, the branch is always here.

PR #2594, opened 2019-12-16
PAASTA-15954 Add kubernetes_remove_evicted_pods script

gvolco:
Adds kubernetes_remove_evicted_pods script which removes Evicted pods and notifies service owners.

Sample invocation:
```
$ sudo ./kubernetes_remove_evicted_pods.py -d /nail/home/gabrielv/services
INFO:__main__:Pods in evicted state: ['gabrielv-deployment-5f4585b554-clzk8', 'compute-infra-test-service-main--k8s-5c7dd5589f-bsf4s']
INFO:__main__:Could not get service name for pod gabrielv-deployment-5f4585b554-clzk8
INFO:__main__:Notifying owners for service compute-infra-test-service
INFO:__main__:Removing pod compute-infra-test-service-main--k8s-5c7dd5589f-bsf4s
```

This produces the following Sensu alert:
```
The following pods have been evicted and will be removed from the cluster:
- compute-infra-test-service-main--k8s-5c7dd5589f-bsf4s: Pod ephemeral local storage usage exceeds the total limit of containers 1280Mi.
...
```

### DIFF
--- a/paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
+++ b/paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
@@ -1,0 +1,156 @@
+#!/usr/bin/env python
+# Copyright 2015-2019 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Usage: ./kubernetes_remove_evicted_pods.py [options]
+
+Removes Evicted pods and notifies service owners
+"""
+import argparse
+import logging
+from collections import defaultdict
+from collections import namedtuple
+from typing import Dict
+from typing import List
+from typing import Mapping
+from typing import Sequence
+
+from kubernetes.client import V1DeleteOptions
+from pysensu_yelp import Status
+
+from paasta_tools.kubernetes_tools import get_all_pods
+from paasta_tools.kubernetes_tools import KubeClient
+from paasta_tools.kubernetes_tools import V1Pod
+from paasta_tools.monitoring_tools import send_event
+from paasta_tools.utils import DEFAULT_SOA_DIR
+
+
+log = logging.getLogger(__name__)
+EvictedPod = namedtuple("EvictedPod", ["podname", "namespace", "eviction_msg"])
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description="Removes evicted pods and notifies service owners"
+    )
+    parser.add_argument(
+        "-d",
+        "--soa-dir",
+        dest="soa_dir",
+        default=DEFAULT_SOA_DIR,
+        help="define a different soa config directory",
+    )
+    parser.add_argument(
+        "-v", "--verbose", action="store_true", dest="verbose", default=False
+    )
+    parser.add_argument(
+        "-n", "--dry-run", action="store_true", dest="dry_run", default=False
+    )
+    args = parser.parse_args()
+    return args
+
+
+def get_evicted_pods(pods: Sequence[V1Pod]) -> Sequence[V1Pod]:
+    return [
+        pod
+        for pod in pods
+        if pod.status.phase == "Failed" and pod.status.reason == "Evicted"
+    ]
+
+
+def get_pod_service(pod: V1Pod) -> str:
+    return pod.metadata.labels.get("paasta.yelp.com/service")
+
+
+def notify_service_owners(
+    services: Mapping[str, Sequence[EvictedPod]], soa_dir: str, dry_run: bool,
+) -> None:
+    check_overrides = {
+        "page": False,
+        "alert_after": "0m",
+        "realert_every": 1,
+        "tip": "Pods can be Evicted if they go over the allowed quota for a given resource. Check the Eviction message to figure out which resource quota was breached",
+    }
+    for service in services.keys():
+        check_name = f"pod-eviction.{service}"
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
```diff
@@ -0,0 +1,155 @@
...
+                client.core.delete_namespaced_pod(
+                    pod.podname,
+                    pod.namespace,
+                    body=delete_options,
+                    grace_period_seconds=0,
+                )
+                log.info(f"Removing pod {pod.podname}")
+
+
+def evicted_pods_per_service(client: KubeClient) -> Mapping[str, List[EvictedPod]]:
```
vkhromov: nit: it's preferable to use `Sequence` or `Iterable` (depending on do you need indexable access or not) instead of `List` of the list is read-only.

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
```diff
@@ -0,0 +1,155 @@
...
+                Status.CRITICAL,
+                check_output,
+                soa_dir,
+            )
+
+
+def remove_pods(
+    client: KubeClient, services: Mapping[str, List[EvictedPod]], dry_run: bool
+) -> None:
+    delete_options = V1DeleteOptions()
```
vkhromov: nit
Deletions are synchronous by default, which means that the script will be blocked untime the pod will be successfully deleted.
May be use `Background` to speed up the script?

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
```diff
@@ -0,0 +1,155 @@
...
+                    grace_period_seconds=0,
+                )
+                log.info(f"Removing pod {pod.podname}")
+
+
+def evicted_pods_per_service(client: KubeClient) -> Mapping[str, List[EvictedPod]]:
+    all_pods = get_all_pods(kube_client=client, namespace="")
+    evicted_pods = get_evicted_pods(all_pods)
+    log.info(f"Pods in evicted state: {[pod.metadata.name for pod in evicted_pods]}")
+    evicted_pods_aggregated = {}
```
vkhromov: defaultdict

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
```diff
@@ -0,0 +1,155 @@
...
+    for pod in evicted_pods:
+        service = get_pod_service(pod)
+        if service:
+            evicted_pods_aggregated[service] = evicted_pods_aggregated.get(
+                service, []
+            ) + [
+                EvictedPod(
+                    pod.metadata.name, pod.metadata.namespace, pod.status.message
+                )
+            ]
```
vkhromov:
```
evicted_pods_aggregated[service].append(EvictedPod(
                    pod.metadata.name, pod.metadata.namespace, pod.status.message
                ))
```

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
gvolco: I noticed that **mypy** is throwing: `Need type annotation for 'evicted_pods_aggregated'`.

**evicted_pods_aggregated** is used on the return statement, so its type needs to match the type on the function signature (otherwise mypy fails).

So the best I can do is to switch from **List** to  **MutableSequence** (unless I'm missing something else here)

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
gvolco: Sure! I will add the Background propagation_policy and update the review in a moment

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
gvolco: As discussed IRL, I'm switching to **defaultdict** to simplify this snippet.

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
vkhromov: The type of `evicted_pods_aggregated` needs to be a subtype of the return type for `evicted_pods_per_service()` if the former is returned from the later.
So, the former could be e.g. `Dict[std, List[EvictedPod]]`, while the later could be `Mapping[str, Iterable[EvictedPod]]`.

vkhromov:
```python
#! /usr/bin/env python

from typing import *

def f() -> Mapping[str, Iterable[int]]:
    z : Dict[str, List[int]] = {}
    z["qqq"] = [1,2,3]
    z["qqq"].append(4)
    return z

print(f())
```
```bash
$ ./1.py
{'qqq': [1, 2, 3, 4]}

$ mypy --strict 1.py
Success: no issues found in 1 source file
```

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
vkhromov: It seems that you can also set `gracefull_period_seconds` to zero
See https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1DeleteOptions.md

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
gvolco: That makes a lot of sense, thanks for the pointer! I'm going to update the PR in a moment.
By the way, I'll be using Sequence instead of Iterable, as I need index access on one of the functions.

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
gvolco: yeah, that param is already set on `delete_namespaced_pod`.
Do you think that it needs to be set on V1DeleteOptions as well?

vkhromov: anything that works fast and correctly is fine for me :)

PR #2595, opened 2019-12-16
Report Envoy backend status

kaisen:
Adds a section to the output of `paasta status` that displays a service's backend status as seen by Envoy. Envoy's admin endpoint `/clusters` is queried to see the backend endpoint configuration.

A tricky edge case I had to handle was casper-proxied services. Casper endpoints are also listed in a service's endpoint configuration with no identifying information other than the fact that they are set at a higher priority than the service's actual backend endpoints. I ended up having to do a one-time scan for casper endpoints and then filtering them out.

Example command outputs:
https://fluffy.yelpcorp.com/i/49xzkDWrQPT9ZK8z38gZ821rw6SFB5m8.html
https://fluffy.yelpcorp.com/i/L5fbZX4HM2Rrr4bDrV8cnRtPGwFnQvbl.html


paasta_tools/envoy_tools.py
```diff
@@ -0,0 +1,201 @@
...
+    :param envoy_admin_port: The port that Envoy's admin interface is listening on
+    :returns backends: A list of dicts representing the backends of all
+                       services or the requested service
+    """
+    clusters_info = retrieve_envoy_clusters(
+        envoy_host=envoy_host, envoy_admin_port=envoy_admin_port,
+    )
+
+    casper_endpoints = set()
+    for cluster_status in clusters_info["cluster_statuses"]:
```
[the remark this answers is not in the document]
kaisen: Dug into this a bit. A service that is proxied through casper still defines its own "BACKEND" haproxy config which only lists the actual service endpoints, and this is what smartstack_tools.py reads. The part of the config that does the proxying through casper is defined in the "FRONTEND" part of the service's config, which is not read by smartstack_tools.py.

In the corresponding Envoy config, both casper and the actual service backends are defined in the service config, and they both get reported by the `/clusters` endpoint unfortunately.

paasta_tools/api/views/instance.py
```diff
@@ -544,6 +555,74 @@ def marathon_smartstack_status(
...
+    for location, hosts in slave_hostname_by_location.items():
+        envoy_host = hosts[0]
+        sorted_backends = sorted(
+            envoy_tools.get_backends(
+                registration,
+                envoy_host=envoy_host,
+                envoy_admin_port=envoy_tools.get_envoy_admin_port(),
+            ),
+            key=lambda backend: backend["eds_health_status"],
+        )
```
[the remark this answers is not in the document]
kaisen: I'm running into some mypy typing issues when I initially tried to refactor this. I could continue to refactor and pull out duplicate code into separate functions but IMO the code is starting to become unwieldy. If the end goal is to remove smartstack code, I'm wondering if it's worth having a bit of duplicate code for now which would make removing smartstack code easier in the future. Thoughts?

paasta_tools/envoy_tools.py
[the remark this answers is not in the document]
stug: Sorry for the late response here (I was sick all last week).  The issue is that we are simultaneously migrating from synapse/haproxy to Envoy and from mesos to kubernetes, which means that we risk ending up with 4 copies everything (e.g. `match_[pods|tasks]_with_[envoy|haproxy]_backends`).  I see you've pushed some more commits, so maybe you've already done this, but just wanted to make it clear I'm not trying to be a purist for its own sake.

paasta_tools/api/views/instance.py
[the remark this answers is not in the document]
stug: Sorry for the late response here (I was sick all last week).  The issue is that we are simultaneously migrating from synapse/haproxy to Envoy and from mesos to kubernetes, which means that we risk ending up with 4 copies everything (e.g. `match_[pods|tasks]_with_[envoy|haproxy]_backends`).  I'm not trying to be a purist or anything -- I just want to try to keep things organized.  Please take another crack if you haven't already (I'm still looking at your recent commits).

paasta_tools/api/views/instance.py
```diff
@@ -336,37 +356,86 @@ def marathon_smartstack_status(
...
         "registration": registration,
         "expected_backends_per_location": expected_count_per_location,
         "locations": [],
     }
 
     for location, hosts in slave_hostname_by_location.items():
-        synapse_host = hosts[0]
-        sorted_backends = sorted(
-            get_backends(
+        if service_mesh == ServiceMesh.SMARTSTACK:
```
stug: it really feels like the smartstack and envoy parts should just be sparate methods.  Is that where you were running into mypy issues?  If so please try a little more, and maybe I can take a crack at it if you can't make it work.

paasta_tools/api/views/instance.py
```diff
@@ -336,37 +356,86 @@ def marathon_smartstack_status(
...
+                    backend["port_value"],
+                ) in casper_proxied_backends:
+                    is_proxied_through_casper = True
+            location_dict = {
+                "name": location,
+                "running_backends_count": running_backends_count,
+                "backends": envoy_backends,
+                "is_proxied_through_casper": is_proxied_through_casper,
+            }
+            service_mesh_status["locations"].append(location_dict)
```
stug: can we make this a separate function, too?  I am concerned that this function is getting pretty complex.

paasta_tools/api/views/instance.py
```diff
@@ -521,6 +590,9 @@ def instance_status(request):
     include_smartstack = request.swagger_data.get("include_smartstack")
     if include_smartstack is None:
         include_smartstack = True
+    include_envoy = request.swagger_data.get("include_envoy")
+    if include_envoy is None:
+        include_envoy = True
```
vkhromov:
```
include_envoy = request.swagger_data.get("include_envoy", True)
```
?

paasta_tools/cli/cmds/status.py
```diff
@@ -722,6 +730,66 @@ def build_smartstack_backends_table(backends):
     return format_table(rows)
 
 
+def get_envoy_status_human(
+    registration, expected_backends_per_location, locations
```
vkhromov: Could you please add typing annotations for args.

paasta_tools/cli/cmds/status.py
```diff
@@ -722,6 +730,66 @@ def build_smartstack_backends_table(backends):
...
+                f"      Proxied through Casper: {is_proxied_through_casper_output}"
+            )
+
+            backends_table = build_envoy_backends_table(location.backends)
+            output.extend([f"      {line}" for line in backends_table])
+
+    return output
+
+
+def build_envoy_backends_table(backends):
```
vkhromov: Here and below, for all added/updated functions, could you please add typing annotation for the arg and the return value.

kaisen:
Did some refactoring of the dense smartstack/envoy code. I somehow can't test the `paasta status` command locally anymore without getting an error message with `print_marathon_status`. Going to look into it tomorrow.

paasta_tools/api/views/instance.py
kaisen: Unfortunately we can't do that because the key actually exists in the `swagger_data` dict with a `None` value.

```
>>> test = {'somekey': None}
>>> print(test.get('somekey'))
None
>>> print(test.get('somekey', True))
None
```

kaisen:
Pinging @vkhromov and @analogue for a final review.

paasta_tools/utils.py
```diff
@@ -2279,6 +2281,19 @@ def get_tron_config(self) -> dict:
...
 
+    def get_envoy_admin_endpoint_format(self) -> str:
+        """ Get the format string for Envoy's admin interface. """
+        return self.config_dict.get(
+            "envoy_admin_endpoint_format", "http://{host:s}:{port:d}/{endpoint:s}"
+        )
+
+    def get_envoy_admin_port(self) -> int:
+        """ Get the port that Envoy's admin interface is listening on
+        from /etc/hosts. """
```
analogue: nit: /etc/services

analogue:
Thanks for leaving the code base in a better state than you found it by adding type annotations :)

PR #2596, opened 2019-12-16
K8s itest

EmanekaT:
* I only tested this on devbox right now. https://fluffy.yelpcorp.com/i/ScL84JRzqMc3dPzVzjCbnBdqJnXZ55vK.html
* I expect it to work on anywhere Kind can be installed, including Travis. I'm not sure about Jenkins. 
* I still need to add a paasta api for it so that it is possible to test commands like paasta status. 
* I don't think it can be used to test everything. Its main purpose should be to test any kubernetes related code that doesn't have yelp dependencies on yelp infrastructures. For example, creating kubernetes objects from provided soa_configs. 
* I wanted to change this back to Draft but realized that github is a great company and they don't have the function. 
* I use kind because minikube requires root permission to install. But minikube is a more mature product and allow us to test more things. It can be installed on travis as well. I actually would like to install minikube on devbox and use that instead. This should be fairly simple. An issue is that minikube doesn't multiple nodes yet, that does not seem to be an issue for our purposes?


EmanekaT:
BTW this PR is not ready yet. Just open for suggestions.

k8s_itests/Makefile
```diff
@@ -0,0 +1,16 @@
+# You need kind installed to run this!
+
+CLUSTER  := paasta-k8s-test-cluster
```
keymone: kind clusters are global, use `$USER` to isolate

k8s_itests/Makefile
```diff
@@ -0,0 +1,16 @@
...
+export KIND_CLUSTER=$(CLUSTER)
+
+all: clean init itest
+
+init:
+	kind create cluster --name=$(KIND_CLUSTER) --config=./deployments/cluster.yaml
+
+itest:
+	export KUBECONFIG=`kind get kubeconfig-path --name="paasta-k8s-test-cluster"`; ../.paasta/bin/tox -e k8s_itests
```
keymone: - `s/paasta-k8s-test-cluster/$KIND_CLUSTER/`
- break up the line after semicolon please, use backslash so that it's still the same shell
- you can drop `export` and semicolon to pass the env var directly

k8s_itests/cmds.py
```diff
@@ -0,0 +1,12 @@
+import subprocess
```
keymone: don't know about others, but it really irks me that our `*_itests` folders all precede the main source folder `paasta_tools`, this makes bunch of things less efficient, like navigating search results, how about we rename all of it to `tests_integration_*`? cc @solarkennedy @Rob-Johnson @vkhromov @chlgit

k8s_itests/deployments/cluster.yaml
```diff
@@ -0,0 +1,22 @@
+---
+kind: Cluster
+apiVersion: kind.sigs.k8s.io/v1alpha3
+network: k8s_kind
```
keymone: this is not going to work in travis without more shenanigans, pre-process it to be absent by default and `k8s_kind` in yelp envs. the attribute is only supported by our fork of kind.

k8s_itests/scripts/generate_all_deployments
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+#
+# Generates all the per-service deployments.json files
+#
+
+set -eo pipefail
+
+# xargs will return 0 if everything went ok, but 12X if something else went wrong
+python -m paasta_tools.cli.cli list -y `printenv SOA_DIR` | shuf | xargs -n 1 -r -P 4 generate_deployments_for_service -d `printenv SOA_DIR` -s
```
keymone: please split into multiple lines

k8s_itests/test_autoscaling.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+# @pytest.fixture(scope='session', autouse=True)
+def test_cluster_info():
+    service_instances = cmd(
+        f'python -m paasta_tools.list_kubernetes_service_instances -d {os.environ["SOA_DIR"]}',
+        capture_output=True,
+    )
+    cmd(
+        f'python -m paasta_tools.setup_kubernetes_job {service_instances.stdout.strip()} -v -d {os.environ["SOA_DIR"]}',
```
keymone: please split into multiple lines

k8s_itests/Makefile
EmanekaT: Nice catch. I set the cluster to a fixed name because it's painful to dynamically change the cluster names in the assets.

keymone: the issue is that if two people on same machine run these itests their clusters will conflict, this is why i think we should use $USER

EmanekaT:
Finally it works. I'll clean up the code now.

k8s_itests/Makefile
```diff
@@ -0,0 +1,28 @@
...
+export SOA_DIR=$(PAASTA_CONFIG_DIR)/fake_soa_config
+export PAASTA_SYSTEM_CONFIG_DIR=$(PAASTA_CONFIG_DIR)/fake_etc_paasta
+export PAASTA_API_PORT= `ephemeral-port-reserve`
+export KUBECONFIG=$(shell kind get kubeconfig-path --name=$(KIND_CLUSTER))
+
+ifeq ($(findstring .yelpcorp.com,$(shell hostname -f)), .yelpcorp.com)
+	export KIND_CONFIG=deployments/kind/cluster-devbox.yaml
+else
+	export KIND_CONFIG=deployments/kind/cluster.yaml
+endif
```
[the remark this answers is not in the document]
EmanekaT: It's actually the network section that makes the difference. regular network port is occupied on devboxes.

k8s_itests/Makefile
```diff
@@ -0,0 +1,28 @@
+# You need kind installed to run this!
```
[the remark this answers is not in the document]
EmanekaT: But then you need to install go, kubectl, and more. Those require root permission. We already have kind on devboxes, and kind is installed on .travis.yml.

k8s_itests/scripts/render_template.py
```diff
@@ -0,0 +1,119 @@
...
+                os.mkdir(new_dst)
+            except OSError as e:
+                if e.errno != os.errno.EEXIST:
+                    raise
+            render(f.path, new_dst, values, exclude)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(
+        description="Replaces all <%%> in all files in src with values provided, and writes the results to dst folder. $() is reserved for environment variables. File/dir that starts with . are ignored"
```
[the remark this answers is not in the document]
EmanekaT: hmm, I didn't find one that satisfies my exact demand so I wrote one. It's not hard to such a short script anyway.

k8s_itests/deployments/kind/cluster.yaml
```diff
@@ -0,0 +1,22 @@
...
+        hostPath: /etc/paasta
+        readOnly: true
+  - role: worker
+    extraMounts:
+      - containerPath: /nail
+        hostPath: /nail
+        readOnly: true
+      - containerPath: /etc/paasta
+        hostPath: /etc/paasta
+        readOnly: true
```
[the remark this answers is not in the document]
EmanekaT: Nice catch. This is just copied from flink-operator. I just removed those mounts, since they are not needed or used.

k8s_itests/Makefile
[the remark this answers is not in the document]
EmanekaT: https://sourcegraph.yelpcorp.com/sysgit/puppet@00403a1e6a221a119c5381d62dab0661b13881de/-/blob/modules/role/manifests/devbox.pp#L338

k8s_itests/Makefile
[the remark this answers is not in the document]
EvanKrall: Relying on system dependencies means your code is likely to break if you run it on a different box. This is why we use tox to manage python virtualenvs, and docker to manage pretty much everything else.

On the python side, we require make and python3.6, since we [install tox in the Makefile](https://github.com/Yelp/paasta/blob/db7498cadaedbc7af8fab501fe1e74a533265e85/Makefile#L39), and use tox to install all other python dependencies.

For the mesos itests, we just require make, python3.6, and docker. Once we have docker, we install all our dependencies ourselves.

You only need root to install things if you're trying to install into root-owned directories like /usr/bin - if you set GOPATH and GOROOT to some directory within the paasta checkout, you won't need root. If you definitely need root, that's what docker is for.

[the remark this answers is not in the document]
EmanekaT: I guess there are three possible use cases. 1. devboxs. 2. travis CI. 3. others. 
1. On all devboxes, we already have the required dependencies https://sourcegraph.yelpcorp.com/sysgit/puppet@00403a1e6a221a119c5381d62dab0661b13881de/-/blob/modules/role/manifests/devbox.pp#L149-150
2. On travis CI, I installed all dependencies.
3. This is the case you talked about. And I think root is required to install go, or perhaps there is a way to workaround that I don't know? And sure I can write a script to install kubectl and kind, as on Travis. But the user still might need to configure cluster.yaml file themselves depending on their system environment. And what if the user uses Mac or Windows? What about podman? Perhaps I go too far but I initially only intend to support the first two use cases and that's why I removed the installation part.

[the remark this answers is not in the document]
EmanekaT: I guess the only solution to this problem is to containerize paasta but can we?

k8s_itests/README
```diff
@@ -0,0 +1,16 @@
...
+    2. All environment variables are set in the Makefile and are exported. They are exposed to tox as well.
+    3. All leftover from old tests are removed. A Kind cluster is then created.
+    4. Tox runs
+    5. Files in deployments/paasta folder are copied, dynamically rendered, and pasted to .tmp deployments/.tmp folder.
+    6. PyTests run.
+    7. Before real tests run, paasta soa configurations are set up and paasta api starts.
+    8. Real tests run.
+After test:
+    PaaSTA api server stops. But kind cluster is not destroied and all the assets are not cleaned up. This is because creating
+    kind is slow and it's a good idea to reuse it. You need to run `make clean`.
```
EvanKrall: This is probably fine for dev boxes and Travis, but I'm worried about Jenkins. Is this ever cleaned up on Jenkins?

EmanekaT: I don't know. I didn't add this test to jenkins yet because I don't know if jenkins have kubectl, or even Go. I don't clean them up mostly because it's easier for people to do manual testing. If we decide to run this test on Jenkins in the future, we can just add another step `make clean`.

PR #2601, opened 2019-12-19
Use the service configuration library to set up Spark on Tron

drmorr0:
This uses the changes in https://github.com/Yelp/service_configuration_lib/pull/39 to set up a spark environment for Tron.  Tested manually, but there's still going to be more work to do.

I also made changes so that we actually get a "real" development virtualenv for paasta, cuz I'm sick of typing `.tox/py36-linux/bin/pytbnon` all the time.

### DIFF
--- a/paasta_tools/setup_tron_namespace.py
+++ b/paasta_tools/setup_tron_namespace.py
@@ -127,7 +127,7 @@ def main():
                     skipped.append(service)
                     log.debug(f"Skipped {service}")
         except Exception as e:
-            log.exception(f"Update for {service} failed: {str(e)}")
+            log.error(f"Update for {service} failed: {str(e)}")
             log.debug(f"Exception while updating {service}", exc_info=1)
             failed.append(service)
 
--- a/paasta_tools/setup_tron_namespace.py
+++ b/paasta_tools/setup_tron_namespace.py
@@ -127,7 +127,7 @@ def main():
                     skipped.append(service)
                     log.debug(f"Skipped {service}")
         except Exception as e:
-            log.error(f"Update for {service} failed: {str(e)}")
+            log.exception(f"Update for {service} failed: {str(e)}")
             log.debug(f"Exception while updating {service}", exc_info=1)
             failed.append(service)
 
--- a/paasta_tools/spark_tools.py
+++ b/paasta_tools/spark_tools.py
@@ -1,0 +1,111 @@
+import logging
+import os
+import sys
+from typing import List
+from typing import Optional
+from typing import Tuple
+
+import boto3
+from botocore.session import Session
+from ruamel.yaml import YAML
+
+from paasta_tools.utils import paasta_print
+from paasta_tools.utils import PaastaColors
+
+AWS_CREDENTIALS_DIR = "/etc/boto_cfg/"
+DEFAULT_SPARK_RUN_CONFIG = "/nail/srv/configs/spark.yaml"
+DEFAULT_SPARK_SERVICE = "spark"
+EXTRA_SPARK_VOLUMES = ("/etc/passwd", "/etc/group")
+log = logging.getLogger(__name__)
+
+
+def _load_aws_credentials_from_yaml(yaml_file_path) -> Tuple[str, str]:
+    with open(yaml_file_path, "r") as yaml_file:
+        try:
+            credentials_yaml = YAML().load(yaml_file.read())
+        except Exception as e:
+            paasta_print(
+                PaastaColors.red(
+                    "Encountered %s when trying to parse AWS credentials yaml %s. "
+                    "Suppressing further output to avoid leaking credentials."
+                    % (type(e), yaml_file_path)
+                )
+            )
+            sys.exit(1)
+
+        return (
+            credentials_yaml["aws_access_key_id"],
+            credentials_yaml["aws_secret_access_key"],
+        )
+
+
+def get_aws_credentials(
+    service: str = DEFAULT_SPARK_SERVICE,
+    no_aws_credentials: bool = False,
+    aws_credentials_yaml: Optional[str] = None,
+    profile_name: Optional[str] = None,
+) -> Tuple[Optional[str], Optional[str]]:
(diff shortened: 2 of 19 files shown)


paasta_tools/spark_tools.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def get_formatted_spark_volumes(instance_config, system_volumes) -> List[str]:
+    volume_dicts = instance_config.get_volumes(system_volumes)
+    volumes = [
+        f"{v['hostPath']}:{v['containerPath']}:{v['mode'].lower()}"
+        for v in volume_dicts
+        if os.path.exists(v["hostPath"]) and v["hostPath"] not in EXTRA_SPARK_VOLUMES
+    ]
+    for v in EXTRA_SPARK_VOLUMES:
```
qui: I think you want these extra spark volumes on the driver container, and don't need them in the executors. Is that right @huadongliu?

tests/test_tron_tools.py
```diff
@@ -102,7 +102,7 @@ def test_get_executor_paasta(self):
             config_dict=action_dict,
             branch_dict={},
         )
-        assert action_config.get_executor() == "mesos"
+        assert action_config.get_executor() == "paasta"
```
qui: I'd like to see some test cases for the spark executor, too.

paasta_tools/tron_tools.py
```diff
@@ -189,7 +198,39 @@ def get_deploy_group(self) -> Optional[str]:
...
+        return command
+
+    def get_env(self):
+        env = super().get_env()
+        spark_env = {}
+        if self.get_executor() == "spark":
+            spark_env = get_mesos_spark_env(
+                spark_app_name="tron_spark_{self.get_service()}_{self.get_instance()}",
+                spark_ui_port=pick_random_port(
+                    f"{self.get_service()},{getuser()}".encode("utf8")
```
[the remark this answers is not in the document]
drmorr0: Ah, good point, it looks like `spark-run` uses "service + PID", I changed to that here.

paasta_tools/setup_tron_namespace.py
```diff
@@ -127,7 +127,7 @@ def main():
                     skipped.append(service)
                     log.debug(f"Skipped {service}")
         except Exception as e:
-            log.error(f"Update for {service} failed: {str(e)}")
+            log.exception(f"Update for {service} failed: {str(e)}")
```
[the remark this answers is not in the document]
drmorr0: drmorr's reading comprehension--

paasta_tools/spark_tools.py
drmorr0: Oh you're right, in `spark-run` it adds these _after_ it creates the spark executor env stuff.

drmorr0: Actually we don't need these at all for Tron, these are just for jupyter-lab.

paasta_tools/tron_tools.py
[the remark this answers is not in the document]
qui: Is the process for this PID setup_tron_namespace? Because then you'll get the same port for all spark jobs in the same service.

paasta_tools/spark_tools.py
```diff
@@ -0,0 +1,118 @@
...
+from typing_extensions import TypedDict
+
+from paasta_tools.utils import paasta_print
+from paasta_tools.utils import PaastaColors
+
+AWS_CREDENTIALS_DIR = "/etc/boto_cfg/"
+DEFAULT_SPARK_MESOS_SECRET_FILE = "/nail/etc/paasta_spark_secret"
+DEFAULT_SPARK_RUN_CONFIG = "/nail/srv/configs/spark.yaml"
+DEFAULT_SPARK_SERVICE = "spark"
+EXTRA_SPARK_DRIVER_VOLUMES = ("/etc/passwd", "/etc/group")
```
qui: I guess you can remove this now

paasta_tools/tron_tools.py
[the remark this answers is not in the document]
drmorr0: Bleh, you're right.  Thanks.

drmorr0:
I had to mess around with a bunch of travis nonsense to get the build to pass.  It seems like we're running into (some variant of) https://github.com/pypa/pip/issues/6275.  I used the `python setup.py egg_info` workaround which appears to fix things.

paasta_tools/tron_tools.py
```diff
@@ -189,7 +196,41 @@ def get_deploy_group(self) -> Optional[str]:
         return self.config_dict.get("deploy_group", None)
 
     def get_cmd(self):
-        return self.config_dict.get("command")
+        command = self.config_dict.get("command")
+        if self.get_executor() == "spark":
+            # Spark expects to be able to write to MESOS_SANDBOX but we don't
+            # use that environment variable so we have to unset it here
```
stug: I am still not sure I understand this.

paasta_tools/tron_tools.py
```diff
@@ -189,7 +196,41 @@ def get_deploy_group(self) -> Optional[str]:
...
+                mesos_secret=load_mesos_secret_for_spark(),
+                paasta_cluster=self.get_cluster(),
+                paasta_pool=self.get_pool(),
+                paasta_service=self.get_service(),
+                paasta_instance=self.get_instance(),
+                docker_img=self.get_docker_url(),
+                volumes=format_volumes(
+                    self.get_volumes(load_system_paasta_config().get_volumes())
+                ),
+                user_spark_opts=self.config_dict.get("spark"),
```
stug: does this mean that the key in the tronfig for this will just be `spark`?  If so maybe `spark_args` or something would be better?

drmorr0: yea that's probably true

drmorr0:
> Mostly looks good to me. Are we planning to do another pass and make spark-tools to use the code in service_configuration_lib, too?

@stug Yes, but we need to get all of the existing Tron jobs that use `paasta spark-run` using this first.

PR #2604, opened 2019-12-21
Move decryption out of paasta local-run cmd

jangsutsr:
A follow-up PR will make paasta spark-run to use the helper function.

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -39,10 +39,7 @@ from paasta_tools.cli.utils import pick_random_port
 from paasta_tools.generate_deployments_for_service import build_docker_image_name
 from paasta_tools.long_running_service_tools import get_healthcheck_for_instance
 from paasta_tools.paasta_execute_docker_command import execute_in_container
-from paasta_tools.secret_tools import get_secret_provider
-from paasta_tools.secret_tools import is_secret_ref
-from paasta_tools.secret_tools import is_shared_secret
-from paasta_tools.secret_tools import SHARED_SECRET_SERVICE
+from paasta_tools.secret_tools import decrypt_secret_environment_variables
 from paasta_tools.tron_tools import parse_time_variables
 from paasta_tools.utils import _run
 from paasta_tools.utils import DEFAULT_SOA_DIR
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -750,14 +679,20 @@ def run_docker_container(
...
+        try:
+            secret_environment = decrypt_secret_environment_variables(
+                secret_provider_name=secret_provider_name,
+                environment=environment,
+                soa_dir=soa_dir,
+                service_name=service,
+                cluster_name=instance_config.cluster,
+                secret_provider_kwargs=secret_provider_kwargs,
+            )
+        except Exception:
```
keymone: catch secret-specific error instead of `Exception` and print error details here instead of in util module

paasta_tools/secret_tools.py
```diff
@@ -103,3 +104,72 @@ def get_secret_hashes(
...
+                service_secret_env,
+                service_name,
+                secret_provider_name,
+                soa_dir,
+                cluster_name,
+                secret_provider_kwargs,
+            )
+        )
+        decrypted_secrets.update(
+            decrypt_secret_environment_for_service(
```
keymone: i know you're preserving original functionality, but wouldn't this call override secrets from previous `.update`? seems like the order should be switched, more specific secrets written last, overriding more general secrets.

paasta_tools/secret_tools.py
```diff
@@ -103,3 +104,72 @@ def get_secret_hashes(
...
+                shared_secret_env,
+                SHARED_SECRET_SERVICE,
+                secret_provider_name,
+                soa_dir,
+                cluster_name,
+                secret_provider_kwargs,
+            )
+        )
+    except Exception as e:
+        paasta_print(f"Failed to retrieve secrets with {e.__class__.__name__}: {e}")
```
keymone: re-raise secret-specific error instead of logging in utility module

paasta_tools/secret_tools.py
qui: Hm, I wrote the original functionality, and didn't think about this...  In practice I don't think these dicts will share any keys, because the keys come from the env dict in yelpsoa-configs, which will already be de-deduplicated. For example, if someone puts
```
env:
    SECRET1: SECRET(first)
    SECRET1: SHARED_SECRET(second)
```
the service_secret_env at this point will have one or the other value for SECRET1, depending on how the yaml parsing works.

I think that means the order of updates doesn't matter here, but switching them is fine if that makes it more readable.

keymone: good point, i missed that `service_secret_env` and `shared_secret_env` are built from the same dictionary so their keys can't overlap, resolving this.

PR #2605, opened 2019-12-23
Fix status for kubernetes instances

keymone:


keymone:
sorry @qui, i noticed your comment too late and already shipped this :(

> What if we centralized it into a class

we could, i don't think it's much different from just having a module with functions. having a class lowers the barrier to make your code stateful, which i'm generally trying to avoid, but ultimately these are the same:

    from paasta_tools.instance.kubernetes import KubernetesInstanceType
    KubernetesInstanceType.get_status()

    from paasta_tools.instance import kubernetes
    kubernetes.get_status()

i still plan to move more instance types into separate modules and when i do it i'll have a better sense of what each module's (class') signature should look like.

in the end i'd like us to end up with a kind of instance handler registry, which we populate during startup of any paasta script (hopefully lazily).

PR #2609, opened 2020-01-02
add api endpoint to return deploy queue

stug:
I already tested this locally -- I can pb you the output if you'd like.

### DIFF
--- a/paasta_tools/api/views/deploy_queue.py
+++ b/paasta_tools/api/views/deploy_queue.py
@@ -1,0 +1,50 @@
+#!/usr/bin/env python
+# Copyright 2015-2020 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


paasta_tools/api/views/deploy_queue.py
```diff
@@ -0,0 +1,70 @@
...
+        "available_service_instances": available_service_instance_dicts,
+        "unavailable_service_instances": unavailable_service_instance_dicts,
+    }
+
+
+def service_instance_to_dict(service_instance: ServiceInstance) -> Mapping[str, Any]:
+    service_instance_dict = service_instance._asdict()
+    return {
+        field: service_instance_dict[field]
+        for field in SERVICE_INSTANCE_FIELDS_TO_RETURN
```
EvanKrall: Since it looks like you're only using this constant once, it may be better to just unroll the loop, so that mypy can verify that the keys are correct.

Alternatively -- any reason why we're whitelisting these keys? Why not just return them all? This seems to be everything but `watcher`, which seems like it might come in handy when trying to debug what's happening.

stug: yeah you are right, i will just remove the whitelist

PR #2610, opened 2020-01-03
Read spark.mesos.secret from vault

jangsutsr:
A follow-up puppet change is needed to make sure paasta vault token file
is present in tron servers.

### DIFF
--- a/paasta_tools/spark_tools.py
+++ b/paasta_tools/spark_tools.py
@@ -2,6 +2,8 @@ import logging
 import os
 import socket
 import sys
+from typing import Any
+from typing import Dict
 from typing import Mapping
 from typing import Optional
 from typing import Tuple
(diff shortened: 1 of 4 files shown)


paasta_tools/tron_tools.py
```diff
@@ -205,24 +205,40 @@ def get_cmd(self):
             command = "unset MESOS_DIRECTORY MESOS_SANDBOX; " + command
         return command
 
+    def get_vault_token_file(self):
+        return self.config_dict.get(
+            "vault_token_file", "/var/spool/.paasta_vault_token"
```
drmorr0: nit: Can we make `/var/spool/.paasta_vault_token` a global constant?

paasta_tools/spark_tools.py
```diff
@@ -105,13 +108,18 @@ def get_default_event_log_dir(**kwargs) -> str:
...
+def load_mesos_secret_for_spark(
+    secret_provider_name: str,
+    soa_dir: str,
+    service_name: str,
+    cluster_name: str,
+    secret_provider_kwargs: Dict[str, Any],
+) -> str:
+    return decrypt_secret_environment_variables(
+        secret_provider_name=secret_provider_name,
+        environment={"SPARK_MESOS_SECRET": "SHARED_SECRET(SPARK_MESOS_SECRET)"},
```
drmorr0: Maybe worth having these as global constants?

paasta_tools/tron_tools.py
stug: yeah or maybe put it in the SystemPaastaConfig?

PR #2614, opened 2020-01-07
add cli command to show deploy queue

stug:


stug:
I had to add a second commit to fix some mypy issues in `tests/`.  In my first commit I added `tests/__init__.py`, and this caused a bunch of mypy issues -- I think we were never actually running mypy on these files because it couldn't import them previously..?

tests/test_kubernetes_tools.py
```diff
@@ -280,11 +283,11 @@ def test_get_cmd_list(self):
             service="kurupt",
             instance="fm",
             cluster="brentford",
-            config_dict={"cmd": ["/bin/echo", "hi"]},
+            config_dict={"cmd": "/bin/echo hi"},
             branch_dict=None,
             soa_dir="/nail/blah",
         )
-        assert deployment.get_cmd() == ["/bin/echo", "hi"]
+        assert deployment.get_cmd() == ["sh", "-c", "/bin/echo hi"]
```
EvanKrall: what's this doing here?

EvanKrall: Oh, I see your comment about mypy in tests/.

Most things in tests/ aren't really meant to be mypy-tested; I think sometimes mypy is a little too strict with tests, which are big on duck typing.

tests/test_secret_tools.py
```diff
@@ -36,7 +36,7 @@ def test_is_secret_ref():
     # if it is not a string it can't be a secret ref
     # so this checks that we are catching the TypeError
     assert not is_secret_ref(None)
-    assert not is_secret_ref(3)
+    assert not is_secret_ref("3")
```
EvanKrall: this assert is explicitly meant to check for non-string values; maybe do type: ignore instead

tests/test_kubernetes_tools.py
stug: I think you added some of these files to the list of what mypy should check :) https://github.com/Yelp/paasta/blame/master/tox.ini

tests/test_secret_tools.py
stug: Aha, ok, will change.

stug: Or, well, is this something that mypy should prevent from happening (in which case we wouldn't need to test it, really) ?

stug: We talked and I am going to fix it -- everything that calls this method has type anotations, and they all only pass in `str`s.  However, since the origin of the values being passed to this method is yelpsoa-configs YAMLs, we can't totally guarantee that it'll only get `str`s.  Maybe we can add more validation?  But in the meantime we'll keep this test.

PR #2621, opened 2020-01-10
Fix paasta autoscale error messaging

enochtangg:
Description:
The following code is used to catch bravado exception 404s when a user specifies an unknown service name or a service that does not have autoscaling configured. Instead of returning the stack trace, we relay an error message.

Testing Done:
* `make test`
* Example of service without autoscaling configured: `paasta autoscale -c norcal-prod -s billing -i sfn_flat_rate_subscription_bookkeeping_worker --set=30`

### DIFF
--- a/paasta_tools/cli/cmds/autoscale.py
+++ b/paasta_tools/cli/cmds/autoscale.py
@@ -66,34 +66,39 @@ def paasta_autoscale(args):
             "Could not connect to paasta api. Maybe you misspelled the cluster?"
         )
         return 1
-
-    if args.set is None:
-        log.debug("Getting the current autoscaler count...")
-        res, http = api.autoscaler.get_autoscaler_count(
-            service=service, instance=args.instance
-        ).result()
-    else:
-        log.debug(f"Setting desired instances to {args.set}.")
-        body = {"desired_instances": int(args.set)}
-        try:
+    try:
+        if args.set is None:
+            log.debug("Getting the current autoscaler count...")
+            res, http = api.autoscaler.get_autoscaler_count(
+                service=service, instance=args.instance
+            ).result()
+        else:
+            log.debug(f"Setting desired instances to {args.set}.")
+            body = {"desired_instances": int(args.set)}
             res, http = api.autoscaler.update_autoscaler_count(
                 service=service, instance=args.instance, json_body=body
             ).result()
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/autoscale.py
```diff
@@ -72,9 +75,17 @@ def paasta_autoscale(args):
...
-            service=service, instance=args.instance, json_body=body
-        ).result()
+        try:
+            res, http = api.autoscaler.update_autoscaler_count(
+                service=service, instance=args.instance, json_body=body
+            ).result()
+        except HTTPNotFound:
+            paasta_print(
+                PaastaColors.red(
+                    f"ERROR: No such service or service is not configured to autoscale."
```
solarkennedy: Can you make this more user friendly?

How about


    Error: {instance} is not configured to autoscale, so 'paasta autoscale' cannot scale it up on demand.
    If you want to be able to boost this service, please configure autoscaling for it in the config file by setting min and max instances. Example:
    
         {instance}:
            min_instances: 5
            max_instances: 50

paasta_tools/cli/cmds/autoscale.py
```diff
@@ -72,9 +75,17 @@ def paasta_autoscale(args):
     else:
         log.debug(f"Setting desired instances to {args.set}.")
         body = {"desired_instances": int(args.set)}
-        res, http = api.autoscaler.update_autoscaler_count(
-            service=service, instance=args.instance, json_body=body
-        ).result()
+        try:
```
solarkennedy: Can you encapsulate this so it covers the `set is None` case on 70 too?

paasta_tools/cli/cmds/autoscale.py
```diff
@@ -63,26 +66,39 @@ def paasta_autoscale(args):
...
-            instance=args.instance,
-            cluster=args.cluster,
+            _log_audit(
+                action="manual-scale",
+                action_details=body,
+                service=service,
+                instance=args.instance,
+                cluster=args.cluster,
+            )
+    except HTTPNotFound:
```
kawaiwanyelp: nit: I would add a test that if we failed somehow, the exception is caught. Feel free to ignore, it's a minor test.

PR #2623, opened 2020-01-14
Remove all references of check_synapse_replication

enochtangg:
Description:
This PR removes all old references of `check_synapse_replication` as it has been migrated to `check_marathon_services_replication`. Paasta documentation has also been updated and old references have been removed .

Testing Done:

* `make test`
* `make docs`


docs/source/generated/paasta_tools.spark_tools.rst
```diff
@@ -0,0 +1,7 @@
+paasta_tools.spark_tools module
+===============================
+
+.. automodule:: paasta_tools.spark_tools
+    :members:
+    :undoc-members:
+    :show-inheritance:
```
stug: what's the deal with this?

enochtangg: I think that was auto-generated from `make docs`.

kawaiwanyelp: I think it was generated when he ran `make docs`.

PR #2627, opened 2020-01-16
Add a script that subscribes to the Mesos master's event stream, and …

EvanKrall:
…distributes TASK_UPDATED messages to the services' logs, along with an interpretation. PAASTA-16165


paasta_tools/subscribe_mesos_master.py
```diff
@@ -0,0 +1,247 @@
+import argparse
```
solarkennedy: log_task_lifecycle_events.py ?

paasta_tools/subscribe_mesos_master.py
```diff
@@ -0,0 +1,247 @@
...
+            return marathon_tools.parse_service_instance_from_executor_id(
+                executor_id or task_id
+            )
+        elif framework_name.startswith("tron"):
+            return tron_tools.decompose_executor_id(executor_id)[:2]
+        elif framework_name.startswith("paasta-remote "):
+            # sorta gross, but it's the same format.
+            return marathon_tools.parse_service_instance_from_executor_id(executor_id)
+        else:
+            raise UnknownFrameworkError(
```
solarkennedy: I looks like this is fatal?

Can you catch this later on so that this script wont crash when a jolt run happens or spark or whatever?

solarkennedy: Oh, you have the skip thing in another function.

Hmm. Just kinda concerns me that we assume we know all the framework names. If like, chronos started randomly today this script would crash?

paasta_tools/subscribe_mesos_master.py
```diff
@@ -0,0 +1,247 @@
...
+            "interpretation": interpret_task_updated(task_updated),
+            "git_sha": git_sha,
+            "task_updated": task_updated,
+        }
+        utils._log(
+            service=service,
+            instance=instance,
+            component=LOG_COMPONENT,
+            cluster=self.cluster,
+            line=json.dumps(message),
```
solarkennedy: We don't normally put json in here, right? `_log` *already* puts json onto scribe, so this is json in json?

I guess you want that so that m-f-d or whatever can read this programmatically. Hmm...

EvanKrall: Yeah this is json-in-json. _log takes a `str` for line. In practice I could _probably_ just pass a dict for message, but that might break something downstream that assumes it's a string.

PR #2630, opened 2020-01-18
Use statefulset when aws_ebs_volume is configured

EmanekaT:
Right now, we use statefulset when persistent volume is configured. We should do the same with ebs volume as well. The only services using this configuration are test services run by us.
```
mingqiz@dev54-uswest1adevc:~/yelpsoa-configs (master) $ git grep aws_ebs_volumes
compute-infra-test-service/kubernetes-kubestage.yaml:  aws_ebs_volumes:
compute-infra-test-service/kubernetes-norcal-stagef.yaml:  aws_ebs_volumes:
kubernetes-service/kubernetes-kubestage.yaml:  aws_ebs_volumes:
```

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1036,7 +1036,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             docker_url = self.get_docker_url()
             code_sha = get_code_sha_from_dockerurl(docker_url)
             complete_config: Union[V1StatefulSet, V1Deployment]
-            if self.get_persistent_volumes():
+            if self.get_persistent_volumes() or self.get_aws_ebs_volumes():
                 complete_config = V1StatefulSet(
                     api_version="apps/v1",
                     kind="StatefulSet",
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1036,7 +1036,7 @@ def format_kubernetes_app(self) -> Union[V1Deployment, V1StatefulSet]:
             docker_url = self.get_docker_url()
             code_sha = get_code_sha_from_dockerurl(docker_url)
             complete_config: Union[V1StatefulSet, V1Deployment]
-            if self.get_persistent_volumes():
+            if self.get_persistent_volumes() or self.get_aws_ebs_volumes():
```
kawaiwanyelp: Please update the tests as well: https://github.com/Yelp/paasta/blob/master/tests/test_kubernetes_tools.py#L800

EmanekaT: Updated tests. I have fixed the two failed tests in another PR. 
The three itests are failed for unknown reason (broken dependencies?).

EmanekaT:
After rethinking, perhaps the best solution is to let users decide if they want to use statefulset or deployment. We make it impossible for users to add ebs volumes for deployments/persistent volumes in this way. Since it does not block sourcegraph migration, I'll close this PR.

PR #2631, opened 2020-01-21
Remove old labels `yelp.com/paasta_`

Bronek:
This is rather intrusive change, let's apply it while the number of impacted services is still relatively low.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1001,9 +1001,6 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
                 instance=self.get_sanitised_instance_name(),
             ),
             labels={
-                "yelp.com/paasta_service": self.get_service(),
-                "yelp.com/paasta_instance": self.get_instance(),
-                "yelp.com/paasta_git_sha": code_sha,
                 "paasta.yelp.com/service": self.get_service(),
                 "paasta.yelp.com/instance": self.get_instance(),
                 "paasta.yelp.com/git_sha": code_sha,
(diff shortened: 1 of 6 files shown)


Bronek:
> Seems good to me overall. I'm not a paasta repo expert, but looks good for Kafka :)
> 
>     * Is the `yelp.com/paasta_service` label in [paasta_tools/setup_kubernetes_crd.py](https://github.com/Yelp/paasta/blob/cd520ec556ee4161eed0d03cf1ca405c36634441/paasta_tools/setup_kubernetes_crd.py#L119) left here on purpose?
> 
>     * Ditto for [tests/test_cleanup_kubernetes_jobs.py](https://github.com/Yelp/paasta/blob/cd520ec556ee4161eed0d03cf1ca405c36634441/tests/test_cleanup_kubernetes_jobs.py#L33)

oops, I missed these - thank you!

ronin13:
Can we document the side-effects of this change in the PR.  Can this be applied in-place or does this require   re-creation of objects (statefulsets, deployments etc.). Thanks.

Bronek:
> Can we document the side-effects of this change in the PR. Can this be applied in-place or does this require re-creation of objects (statefulsets, deployments etc.). Thanks.

This will replace the definition of the k8s objects which are owned by Paasta, which typically does *not* include StatefulSets etc. since such objects are owned by their respective k8s operators, rather than Paasta. Since the only (production ready) k8s operator that we currently have is Flink, the impact will be mostly limited to Flink cluster.

mattmb:
Cleaning up and closing some very old PRs

PR #2632, opened 2020-01-21
Fix failed HPA tests

EmanekaT:
@Bronek pointed out a couple of failed HPA tests. I fixed them.

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -18,4 +18,4 @@ override_dh_auto_test:
 	true
 
 override_dh_virtualenv:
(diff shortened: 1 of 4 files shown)


vkhromov:
@kaisen @EmanekaT 
Why we need `requirements-bootstrap.txt`?
Where did you get packages versions for it?

kaisen:
@vkhromov See #2636 for the root cause (pip-custom-platform incompatibility with pip>20). I think what happened is that I followed the internal documentation on establishing a requirements-bootstrap.txt file but I did it on a devbox, and our internal pypi only has pip==9

PR #2633, opened 2020-01-22
prevent paasta status from showing traceback

chlgit:


paasta_tools/kubernetes_tools.py
```diff
@@ -1369,14 +1369,18 @@ def list_deployments(
...
+                        name=pod.metadata.name,
+                        namespace="paasta",
+                        container=container.name,
+                        tail_lines=num_tail_lines,
+                    )
                 )
-            )
+            except ApiException as e:
+                log.error(e)
+
```
stug: The problem with this is that `paasta status` just won't get anything back, so users won't know the difference between something going wrong and there just not being any stdout/stderr.  Ideally we'd follow the example of the mesos version: https://github.com/Yelp/paasta/blob/v0.93.2/paasta_tools/mesos_tools.py#L414

paasta_tools/kubernetes_tools.py
```diff
@@ -1369,14 +1369,20 @@ def list_deployments(
...
+                        container=container.name,
+                        tail_lines=num_tail_lines,
+                    )
                 )
-            )
+            except ApiException as e:
+                tail_lines_dict["stdout"].append(
+                    f"couldn't read stdout/stderr because {e.getResponseBody()}"
+                )
+
```
stug: This isn't really stdout.  Can we just follow the same pattern as we do for mesos and return a separate field for the error (https://github.com/Yelp/paasta/blob/v0.93.19/paasta_tools/mesos_tools.py#L456)?

That way the client can decide whether to, say, put the error in red or whatever.

PR #2636, opened 2020-01-23
Use requirements-bootstrap.txt to pin version of pip

kaisen:
The build is failing because of pip-custom-platform's incompatibility with pip>=20. See https://github.com/asottile/pip-custom-platform/issues/32

This pins the version of pip used in the itests via a requirements-bootstrap.txt file.


kaisen:
Merged in #2636

PR #2637, opened 2020-01-24
Showing pod status irrespective of state of flink cluster

AshishKhatkar:
Currently, if a flink cluster is in unrecoverable state, paasta status doesn't show pod status while pods might be running and some of them might be in crashloopbackoff if we check them using kubectl. The change will enable paasta status -v to show pod status if it is available irrespective of the state of the cluster.


AshishKhatkar:
@poros done.

PR #2638, opened 2020-01-24
Update requirements-bootstrap.txt to the state in v0.93.2

vkhromov:
Set requirements to the state they were in `0.93.2` version of
`paasta-tools` and pin `pip` to `18.1`.


keymone:
🚢

PR #2639, opened 2020-01-24
Add second docker wrapper

mattmb:
This is kinda ugly but it does mean that docker inspects will be fast
again. With this 100 docker inspects ~ 15s without ~ 105s

There might be a better fix but just putting this up so we can consider
it.

The motivation is that mesos times out on slow docker inspects and the
python wrapper can be very slow at importing all the python code.

It might be simpler to "fix" this in the python script. @Bronek has this branch: https://github.com/Yelp/paasta/compare/u/bronek/delay_firewall_import but it does break a few tests. It's a bit ugly to move the imports but maybe it's less ugly than a second wrapper?

### DIFF
--- a/debian/paasta-tools.links
+++ b/debian/paasta-tools.links
@@ -34,7 +34,8 @@ opt/venvs/paasta-tools/bin/paasta_cleanup_tron_namespaces usr/bin/paasta_cleanup
 opt/venvs/paasta-tools/bin/paasta_cluster_boost.py usr/bin/paasta_cluster_boost
 opt/venvs/paasta-tools/bin/paasta_deploy_tron_jobs usr/bin/paasta_deploy_tron_jobs
 opt/venvs/paasta-tools/bin/paasta-deployd usr/bin/paasta-deployd
-opt/venvs/paasta-tools/bin/paasta_docker_wrapper usr/bin/paasta_docker_wrapper
+opt/venvs/paasta-tools/bin/paasta_docker_wrapper usr/bin/paasta_docker_wrapper_python
+opt/venvs/paasta-tools/bin/paasta_docker_wrapper.sh usr/bin/paasta_docker_wrapper
 opt/venvs/paasta-tools/bin/paasta_dump_locally_running_services usr/bin/paasta_dump_locally_running_services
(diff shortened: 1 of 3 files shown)


Bronek:
I see a small issue with this approach - if the word "inspect" appears anywhere in the docker command, we will make an assumption that's a docker verb - but it could be something else, e.g. some option.

What if we do not change the paasta wrapper here, but instead stop using `paasta_docker_wrapper inspect` itself, and replace with either a new dedicated e.g. `paasta_docker_inspect` or plain `docker inspect` ?

mattmb:
Yep, I see what you mean @Bronek tricky thing is that it would be tricky to point Mesos to different files for inspect vs. other docker calls. Probably possible but we'd have to patch the mesos agent I suspect?

solarkennedy:
Instead of a second wrapper and patching mesos, what about fixing this a different way by using an older version of pip or something?

Bronek:
> Instead of a second wrapper and patching mesos, what about fixing this a different way by using an older version of pip or something?

That's already done and we are not in a dire need of a workaround, however I guess there are other benefits from cutting the "docker inspect" time significantly.

Personally I would vote to take this workaround, and just hope that we do not see "inspect" used outside of docker verb as long as we keep using Mesos.

keymone:
i would very much prefer not having multiple wrappers. delayed module imports + `if not inspect` logic directly in python code is all we need.

Rob-Johnson:
as we've spoken about IRL, I'd also say no-ship.
In the long term, I think all the logic that's in the wrapper should move elsewhere, and if we do need a thin layer here, then we should write it in something faster.

In the short term, I think we should just shuffle imports around and add the 'if not inspect' logic in the existing wrapper, as @keymone has suggested.

Bronek:
I fixed unit tests in the "lets delay imports" approach fix, and created github.com/Yelp/paasta/pull/2650 . I am quite pleased with the result :)

mattmb:
I'm happy with the result, much faster now that we've moved the imports in https://github.com/Yelp/paasta/pull/2650

We can discard this option now

PR #2649, opened 2020-01-30
Added a reminder in slack notifications to run paasta status if it takes too long

solarkennedy:
This time I'm not using the sticht timer mechanism.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -15,6 +15,7 @@
 """Contains methods used by the paasta client to mark a docker image for
 deployment to a cluster.instance.
 """
+import datetime
 import getpass
 import logging
 import math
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -799,6 +851,7 @@ def on_enter_deploying(self):
                 target=self.do_wait_for_deployment, args=(self.commit,), daemon=True
             )
             thread.start()
+            self.schedule_paasta_status_reminder()
 
     def on_exit_deploying(self):
         self.wait_for_deployment_green_lights[self.commit].clear()
```
EvanKrall: you could throw some code to cancel the reminder here

solarkennedy: In order to do that though, I would have to build up the same kind of `timer_handle` logic that stitch does, right?

EvanKrall: yeah, youd have to save the return value of call_later.

solarkennedy: ok. I'm just in general hesitant to turn this functionality into its own full state machine

solarkennedy: Ok. Cancelation support added.

PR #2651, opened 2020-01-30
Make the default tron executor and node be paasta

solarkennedy:
We made it!
There are no jobs at yelp that do not have an explicit executor/node.

I think implicit is better than explicit (sane defaults) when it comes to lots and lots of configs that humans have to interact with all the time, so I would like to make the default executor/node be paasta.

After shipping, I would like to simplify all our tron job configs by removing the things that are already set with this default.


keymone:
shall we also change the default on tron side? https://github.com/Yelp/Tron/blob/master/tron/config/config_parse.py#L364

maybe start phasing out ssh completely and work on adding kubernetes support?

PR #2652, opened 2020-01-31
Swap default service location from gitolite to GitHub

dbgrigsby:
All `services/` repositories are on github.yelpcorp.com now, and any
stragglers outside of that namespace have their git_url manually
specified in their config. New repositories are always created on
GitHub, so changing the default host to GitHub lets new services work
without requiring extra config changes

### DIFF
--- a/paasta_tools/contrib/mass-deploy-tag.sh
+++ b/paasta_tools/contrib/mass-deploy-tag.sh
@@ -26,7 +26,7 @@ for service in ${services} ; do
 	fi
 	# git_repo=$(paasta info -s ${service} | grep -oP 'Git Repo: \K.*$')
 	git_repo=$(script -qc "paasta info -s ${service}" | sed -r "s/\x1B\[([0-9]{1,2}(;[0-9]{1,2})?)?[m|K]//g" | grep 'Git Repo: ' | cut -d' ' -f3)
-	default_git_repo=git@git.yelpcorp.com:services/${service}.git
+	default_git_repo=git@github.yelpcorp.com:services/${service}.git
 	echo git clone ${git_repo-${default_git_repo}} ${service}
 	git clone ${git_repo} ${service}
 	unset git_repo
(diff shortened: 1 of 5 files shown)


vkhromov:
I just run
```shell
$ git grep -F git.yelpcorp | grep -v services
```
and found the following quite suspicious places:
* https://github.com/Yelp/paasta/blob/master/paasta_tools/remote_git.py#L120
* https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/mark_for_deployment.py#L219

Could you please double-check that all cases where we access the `services` repo are handled properly.

dbgrigsby:
@vkhromov Thanks for double checking

The first occurrence there is correct, there are some services outside of the `services/` namespace that remain on gitolite, and manually specify their `git_url`, ignoring this default change. They still rely on that gitolite command to check authors

The second usage there is currently necessary, because a simple service runs on sysgit right now that generates a deployment for services, regardless of git host. 

Looking at all the remaining uses of `git.yelpcorp`, each one make sense

```
% git grep -ni git.yelpcorp | grep -v changelog                                                                                                                                                                                                                   Jenkinsfile:36:                "git@git.yelpcorp.com:mirrors/Yelp/paasta.git",
docs/source/upgrading_marathon.rst:36:   #. Clone the repo where it is built. (``<git@git.yelpcorp.com:packages/marathon>``)
paasta_tools/cli/cmds/mark_for_deployment.py:219:    client.connect(("sysgit.yelpcorp.com", 5049))
paasta_tools/contrib/paasta_update_soa_memcpu.py:160:    remote = "git@sysgit.yelpcorp.com:yelpsoa-configs"
paasta_tools/remote_git.py:120:    if "git.yelpcorp.com" in git_server:
tests/test_remote_git.py:141:        "git@git.yelpcorp.com:yelp-main", "a", "b"
tests/test_remote_git.py:144:        command="ssh git@git.yelpcorp.com authors-of-changeset yelp-main a b",
```

result 1 refers to the internal paasta mirror, 2 to a package rather than a service, 3 and 5 are discussed above, 4 refers to an unmigrated sysgit repo, 6 and 7 are a test for a case that still exists

PR #2653, opened 2020-01-31
is_in_smartstack bug caused erroneous behaviour from CNI wrapper

azretkenzhaliev:
First draft.

### DIFF
--- a/paasta_tools/long_running_service_tools.py
+++ b/paasta_tools/long_running_service_tools.py
@@ -92,7 +92,7 @@ class ServiceNamespaceConfig(dict):
         return self.get("discover", "region")
 
     def is_in_smartstack(self) -> bool:
-        if self.get("proxy_port") is not None:
+        if self.get("proxy_port") is not None or self.get("registrations") is not None:
             return True
         else:
             return False


paasta_tools/long_running_service_tools.py
```diff
@@ -92,7 +92,7 @@ def get_discover(self) -> str:
         return self.get("discover", "region")
 
     def is_in_smartstack(self) -> bool:
-        if self.get("proxy_port") is not None:
+        if self.get("proxy_port") is not None or self.get("registrations") is not None:
```
solarkennedy: Can you use `get_registrations` instead of directly getting the entry in the dict?

The reasoning here is that `get_registrations` contains additional logic about setting sane defaults for registrations, even if the user didn't explicitly set them.

A service can be in smartstack, even if a user doesn't explicitly set a `registrations:`

solarkennedy: Also remember which object you are working on here.

`self` represents ServiceNamespaceConfig, which is *not* a paasta instance config.

azretkenzhaliev: Thanks Kyle, I was wrong with the existence of the bug. Apparently, everything worked after I fixed the bug in my CNI wrapper, and there was no bug in here. Closing PR.

PR #2654, opened 2020-01-31
Make the default tox env run the tests

solarkennedy:


solarkennedy:
I've asked @jangsutsr to take over from here to fix the sparky tests

PR #2655, opened 2020-02-04
Add validation & documentation to newly-added tron spark arguments

jangsutsr:


paasta_tools/cli/schemas/tron_schema.json
```diff
@@ -139,6 +140,15 @@
                 },
                 "trigger_timeout": {
                     "$ref": "#definitions/time_delta"
+                },
+                "spark_cluster": {
```
solarkennedy: Do you think this is better than `spark_paasta_cluster`? 
Same with `spark_pool`

docs/source/yelpsoa_configs.rst
```diff
@@ -609,6 +609,17 @@ Each Tron **action** of a job MAY specify the following:
     then the command should be something available in the docker container (it should NOT
     start with ``paasta local-run``).
 
+If a Tron **action** of a job is of executor type ``spark``, it MAY specify the following:
+
+  * ``spark_cluster``: The Paasta cluster on which to run spark jobs. Default to the same
+    cluster the tron job is running on.
+
+  * ``spark_pool``: The Paasta pool on which to run spark jobs. Default to ``default`` pool
```
solarkennedy: Can you make this default to run on the same pool as the job itself? It would be nice if users didn't have to specify this twice.

stug: Is this the pool/cluster that the driver will run on or that the executors will run on?  The text makes it a little unclear.

If it's the executors, we definitely don't want to default them to the `default` pool.  If it's the driver, I don't think we want to default it to whatever the executors are running on.  The executors should run on a `batch` or `stable_batch` pool, and unless I'm mistaken we haven't changed the metrics we send to clusterman to include the driver's resources.

drmorr0: This is the executors.  I am hesistant to have it default to run on the same pool as the job itself, because the job itself more than likely be running on a different cluster entirely (and probably running on a `batch` pool).  I agree the `default` pool probably isn't the right default, maybe just have it run on the `batch` pool by default?

docs/source/yelpsoa_configs.rst
```diff
@@ -609,6 +609,17 @@ Each Tron **action** of a job MAY specify the following:
...
+    if not specified.
+
+  * ``spark_args``: Dictionary of spark configurations documented in
```
drmorr0: note that, as it currently stands, we dropped the initial `spark.` from the option name, and the code will prepend that before it sends the options over.  In other words, users will write this:

```
spark_args:
    cores.max: 10
    driver.port: 123456
```

not this:

```
spark_args:
    spark.cores.max: 10
    spark.driver.port: 123456
```

It's not actually clear to me that is the _best_ way to do things, it might just cause more confusion in the long run.  I'd be OK changing the code so it doesn't do this if you think it's better.

paasta_tools/cli/schemas/tron_schema.json
drmorr0: I don't really have a problem with it, but we can change it if you want.

docs/source/yelpsoa_configs.rst
stug: My gut feeling is that we should keep the full names with the `spark.` prefix to prevent confusion.  Say, for instance, I am a spark n00b and I am looking at a Spark job and want to google the spark args -- better for them to be whole.

jangsutsr: Yeah, actually the code stripping `spark.` prefix has been removed so we are all good here.

PR #2656, opened 2020-02-04
Make node no longer a required element of tron schema

solarkennedy:


qui:
Node is still required in Tron configuration, so you'll either have to update Tron or add a `node: paasta` when it's not specified by the user. Also, since we still have some ssh executors - is it possible to automatically add `node: localhost` for them? If not, can you add a validation check that all ssh executors still have a node specified?

solarkennedy:
I changed the code in a previous paasta PR to inject 'node: paasta' so tron always has a node to work with. This should keep tron happy but still allow us to not have to specify a node all the time.

I don't know if I want to automatically add node: localhost for ssh executors, but I have a CR in reviewboard to add validation to ensure that if you do have ssh, it isn't 'node: paasta' (it must be explicitly something else).

PR #2657, opened 2020-02-04
Don't try to retrieve authors from GitHub repos with gitolite commands

dbgrigsby:
The old way of finding authors was to call a gitolite command that
finds the authors between two commits. For services on GitHub, we pass
in the authors when making a mark-for-deployment call, so there is no
benefit to trying to call this gitolite command on a GitHub repository.

This change was prompted by the noisiness of timed deploys for shas
already on master not having authors passed in, and paasta's slackbot
complaining about not being able to find any authors from GitHub as a 
fallback when there are no authors for gitolite either.

I updated all `git_url=None` in tests to have a reasonable value, since get_authors_to_be_notified  requires an actual string and adding a check for the case where `git_url` is None doesn't make a lot of sense, since it will always actually exist for real mark for deployment calls

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -279,9 +279,11 @@ def get_authors_to_be_notified(git_url, from_sha, to_sha, authors):
     if from_sha is None:
         return ""
 
+    authors_to_notify = []
+
     if authors:
         authors_to_notify = authors
-    else:
+    elif 'git.yelpcorp.com' in git_url:
         ret, git_authors = remote_git.get_authors(
             git_url=git_url, from_sha=from_sha, to_sha=to_sha
         )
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -279,16 +279,21 @@ def get_authors_to_be_notified(git_url, from_sha, to_sha, authors):
...
         ret, git_authors = remote_git.get_authors(
             git_url=git_url, from_sha=from_sha, to_sha=to_sha
         )
         if ret == 0:
             authors_to_notify = git_authors.split()
         else:
             return f"(Could not get authors: {git_authors})"
+    else:
+        # We have no way of getting authors on the fly if the repository is not on gitolite
+        pass
```
solarkennedy: Why not `return ""` here and not continue as if we have something to work with?

dbgrigsby: that's better

solarkennedy:
Be sure to have the hooks installed so that your code gets auto-linted.

PR #2658, opened 2020-02-04
Add check_flink_pods_running script

catlynk:
Testing done:
1) make test
2) manual test: https://fluffy.yelpcorp.com/i/40HFkZSmWF1zVqkF3S4lHbfXBzWSv6cB.html

I'm treating pods of different subcomponent in the same way and sending one sensu event as long as there's any pod in the service.instance that's not in running state. Don't see much benefit for splitting them up considering the actions to take when getting this alert is pretty much the same (describe the pod and see why it's in pending state). But I'm happy to split them if you'd prefer

### DIFF
--- a/paasta_tools/check_flink_pods_running.py
+++ b/paasta_tools/check_flink_pods_running.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python
+# Copyright 2015-2019 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Usage: ./check_flink_pods_running.py [options]
+"""
+import logging
+from typing import Sequence
+
+import pysensu_yelp
+
+from paasta_tools import flink_tools
+from paasta_tools import kubernetes_tools
+from paasta_tools.check_services_replication_tools import main
+from paasta_tools.flink_tools import FlinkDeploymentConfig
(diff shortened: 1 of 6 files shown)


paasta_tools/check_flink_pods_running.py
```diff
@@ -0,0 +1,93 @@
...
+    instance_config: FlinkDeploymentConfig,
+    all_tasks_or_pods: Sequence[V1Pod],
+    smartstack_replication_checker: KubeSmartstackReplicationChecker,
+) -> None:
+    si_pods = filter_pods_by_service_instance(
+        pod_list=all_tasks_or_pods,
+        service=instance_config.service,
+        instance=instance_config.instance,
+    )
+    taskmanager_expected_running_cnt = instance_config.config_dict.get(
```
JoeMalt: is the choice of 10 here arbitrary, or is this actually the default number of taskmanagers we run if it's not specified?

poros:
@catlynk would you mind holding this up one more day so that I get the chance to speak with some people? I know that data-streams-core wants to be notified for pending pods, too, so it might good to make it generic.

paasta_tools/check_flink_pods_running.py
catlynk: I'm keeping it consistent with the other check here https://sourcegraph.yelpcorp.com/mirrors/Yelp/paasta/-/blob/paasta_tools/check_flink_services_health.py#L120
I didn't find any default value for number of taskmanager instances but maybe I was looking at the wrong places

catlynk:
@poros Sure thing, I'll hold off on pushing this for now

JoeMalt:
@catlynk I talked with @poros about your questions, and I'm not 100% sure why we'd need to configure anything in yelpsoa-configs. For check_flink_services_health we configure it in Puppet here: https://sourcegraph.yelpcorp.com/sysgit/puppet@master/-/blob/modules/profile_kube/manifests/master.pp#L652, would that not work for this check? If somebody wants to use it for something else (e.g. Kafka), they can just duplicate that configuration.

As for services, I don't think this check actually needs to be service-aware at all, since it just needs to let stream-processing know if any Flink pod, from any service, is Pending for a while. 

I hope this is useful, if I've misunderstood your questions then please let me know.

mattmb:

PR #2660, opened 2020-02-05
revert vault changes and just pass in the shared secret env var

drmorr0:
We just pass in the `SHARED_SECRET` into the environment of each of the executors.  If I understand how this all works correctly, this is what we want?

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -35,7 +35,6 @@ except ImportError:  # pragma: no cover (no libyaml-dev / pypy)
     Dumper = yaml.SafeDumper  # type: ignore
 
 from paasta_tools.spark_tools import get_default_event_log_dir
-from paasta_tools.spark_tools import load_mesos_secret_for_spark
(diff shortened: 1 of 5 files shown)


paasta_tools/tron_tools.py
```diff
@@ -219,36 +219,18 @@ def get_spark_paasta_cluster(self):
...
-                    cluster_name=self.get_cluster(),
-                    secret_provider_kwargs={
-                        "vault_cluster_config": system_paasta_config.get_vault_cluster_config(),
-                        "vault_auth_method": "token",
-                        "vault_token_file": self.get_vault_token_file(
-                            system_paasta_config
-                        ),
-                    },
-                ),
+                mesos_secret=SPARK_MESOS_SECRET_KEY,
```
jangsutsr: Just curious how spark driver container (mesos slave) would know to decrypt this env variable?

drmorr0: the Mesos master scans the environment variables for things starting with `SECRET(` or `SHARED_SECRET(` and will decrypt them before starting up the executor.

PR #2661, opened 2020-02-06
Assign git_sha label to kubernetes custom resources

keymone:


paasta_tools/setup_kubernetes_cr.py
```diff
@@ -290,6 +300,11 @@ def reconcile_kubernetes_resource(
     instance: str = None,
 ) -> bool:
 
+    try:
+        deployments = load_v2_deployments_json(service=service)
+    except NoDeploymentsAvailable:
+        return True
```
mattmb: Maybe we should log something?

keymone: the way this script is structured, it goes over *every* service to see if there are any k8s custom instances, so logging here would be noisy. we should plan for some time to refactor.

keymone:
cc @chlgit @EmanekaT need your eyes on this too

PR #2662, opened 2020-02-07
fix docker_url for validation

drmorr0:
Duplicating logic here: https://github.com/Yelp/paasta/blob/master/paasta_tools/tron_tools.py#L574

I debating pulling this out into its own function but it didn't really seem worth it?  IDK, I could be convinced to do it better, though.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -203,6 +203,15 @@ class TronActionConfig(InstanceConfig):
     def get_deploy_group(self) -> Optional[str]:
         return self.config_dict.get("deploy_group", None)
 
+    def get_docker_url(self) -> str:
+        # It's okay for tronfig to contain things that aren't deployed yet - it's normal for developers to
+        # push tronfig well before the job is scheduled to run, and either they'll deploy the service before
+        # or get notified when the job fails.
+        #
+        # This logic ensures that we can still pass validation and run setup_tron_namespace even if
+        # there's nothing in deployments.json yet.
+        return "" if not self.get_docker_image() else super().get_docker_url()
+
     def get_cmd(self):
         command = self.config_dict.get("command")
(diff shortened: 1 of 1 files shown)


drmorr0:
Looks like travis was just a flake.

qui:
We decided in https://github.com/Yelp/paasta/pull/2302 that it's okay for tronfig to contain things that aren't deployed yet - it's normal for developers to push tronfig well before the job is scheduled to run, and either they'll deploy the service before or get notified when the job fails. Before this change, we were getting paged for tronfig deployment in that case.

But that reasoning is easy to forget - let's pull this logic into a `get_docker_url()` method on TronActionConfig and add a comment about why we're doing this.

paasta_tools/tron_tools.py
```diff
@@ -203,6 +203,14 @@ def get_action_name(self):
     def get_deploy_group(self) -> Optional[str]:
         return self.config_dict.get("deploy_group", None)
 
+    def get_docker_url(self) -> str:
+        # it's okay for tronfig to contain things that aren't deployed yet - it's normal for developers to push tronfig
+        # well before the job is scheduled to run, and either they'll deploy the service before or get notified when the
+        # job fails. Before this change, we were getting paged for tronfig deployment in that case.
+        #
+        # This logic ensures that we can still pass validation even if there's nothing in deployments.json yet
```
qui: nit: the comment above is about generating the config for Tron (format_tron_action_dict), which happens during setup_tron_namespace, not about validation. I guess this logic ensures that both pass even if there's nothing yet. We don't care about it when actually applying config, so it makes sense that validation wouldn't need the docker image either.

I think this comment would be fine if you just removed this sentence? But I have more context on it, so please reword in whatever way makes sense to you.

drmorr0: Updated

PR #2663, opened 2020-02-12
Stop reading datacenter (aws region) attribute from expected_slave_at…

EvanKrall:
…tributes, since yelp-autoconfigure-mesos doesn't set it. Use a setting in /etc/paasta instead. PAASTA-16304

Corresponding puppet review: https://reviewboard.yelpcorp.com/r/481051/


solarkennedy:
fix tests + ship

PR #2665, opened 2020-02-13
Restore Report Envoy backend status #2595

ymilki:
This reverts commit ced20ed9a12e3757eaf8c75d7c8b97717a12c663 which was created by #2640 which was a revert of #2595. This effectively restores #2595 . No new changes are introduced in this PR.

We've fixed the issues which were caused by missing envoy configuration.

### DIFF
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -351,6 +351,13 @@
                     },
                     {
                         "in": "query",
+                        "description": "Include Envoy information",
+                        "name": "include_envoy",
+                        "required": false,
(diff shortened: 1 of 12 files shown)


stug:
Just for posterity -- this actually restores #2595

ymilki:
> lg2m. You might consider releasing just on a stage to make sure everything is set depending on your confidence. Or you might be able to set up your local paasta API to think it's in stage (instead of the default norcal-devc)

I did test locally against norcal-devc and norcal-stagef to reproduce the issue and verify the envoy configuration fix.
We will definitely be watching this rollout.

PR #2668, opened 2020-02-18
[KAFKA-25246] Adds support for kafka cluster status in `paasta status`

rohangulati:
***Summarized Context***
These changes adds support for getting more details in the `paasta status` commands. I have also refactored the code that handles printing different output for different instance types like `flink`, `marathon` and make it generic so that adding new type is easier. To make it consistent I had to change the method signatures of all custom status printing commands and make is consistent with the following interface

```python
InstanceStatusWriter = Callable[
    [
        NamedArg(str, "cluster"),
        NamedArg(str, "service"),
        NamedArg(str, "instance"),
        NamedArg(List[str], "output"),
        NamedArg(Any, "status"),
        NamedArg(int, "verbose"),
    ],
    int,
]
```

***Testing Done***

```make test```

**Output**

```================================================================================== 1874 passed, 4 skipped in 19.50 seconds ===================================================================================
__________________________________________________________________________________________________ summary ___________________________________________________________________________________________________
  tests: commands succeeded
  congratulations :)```

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -39,6 +39,7 @@ import humanize
 from bravado.exception import BravadoConnectionError
 from bravado.exception import BravadoTimeoutError
 from bravado.exception import HTTPError
+from mypy_extensions import Arg
 from service_configuration_lib import read_deploy
 
 from paasta_tools import kubernetes_tools
@@ -78,7 +79,6 @@ from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import remove_ansi_escape_sequences
 from paasta_tools.utils import SystemPaastaConfig
 
-
 ALLOWED_INSTANCE_CONFIG: Sequence[Type[InstanceConfig]] = [
     FlinkDeploymentConfig,
     CassandraClusterDeploymentConfig,
@@ -87,6 +87,18 @@ ALLOWED_INSTANCE_CONFIG: Sequence[Type[InstanceConfig]] = [
     AdhocJobConfig,
     MarathonServiceConfig,
     TronActionConfig,
+]
+
+InstanceStatusWriter = Callable[
+    [
+        Arg(str, "cluster"),
+        Arg(str, "service"),
+        Arg(str, "instance"),
+        Arg(List[str], "output"),
+        Arg(Any),
+        Arg(int, "verbose"),
+    ],
+    int,
 ]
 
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: paasta_tools/cli/cmds/status.py [hunk 1]
```diff
@@ -1492,3 +1570,15 @@ def status_marathon_job_human(
         return "Marathon:   {} - {} (app {}) is not configured in Marathon yet (waiting for bounce)".format(
             status, name, desired_app_id
         )
+
+
+# Add other custom status writers here
+# See `print_tron_status` for reference
+INSTANCE_TYPE_WRITERS: Mapping[str, InstanceStatusWriter] = defaultdict(
```
mattmb: nit: can we push this to the top of the file?

paasta_tools/cli/cmds/status.py
```diff
@@ -249,19 +261,14 @@ def paasta_status_on_api_endpoint(
...
-    elif status.tron is not None:
-        return print_tron_status(service, instance, output, status.tron, verbose)
-    elif status.adhoc is not None:
-        return print_adhoc_status(
-            cluster, service, instance, output, status.adhoc, verbose
-        )
-    elif status.flink is not None:
-        return print_flink_status(
-            cluster, service, instance, output, status.flink, verbose
+    # get instance type
```
mattmb: I think you don't need this comment :-)

paasta_tools/cli/cmds/status.py
```diff
@@ -271,6 +278,17 @@ def paasta_status_on_api_endpoint(
         return 0
 
 
+# find_instance_type finds the instance type from the status api response
```
mattmb: style nit: can you make this a docstring rather than a comment

paasta_tools/cli/cmds/status.py
```diff
@@ -1049,6 +1082,51 @@ def print_tron_status(
     return 0
 
 
+def print_kafka_status(
```
mattmb: We should maybe add a test for this? See TestPrintKubernetesStatus for an example

paasta_tools/cli/cmds/status.py
rohangulati: @mattmb I tried pushing this to the top but python somehow is not able to resolve reference to these methods

mattmb: Ah, yeah, thanks for trying

python--

PR #2677, opened 2020-02-21
Refactor instance listing into a single function

qui:
We can simplify the tron logic because we don't have to worry about
jobs and actions being lists. They are always dicts now.

This is in preparation for changing the way we read instance configs
(PAASTA-16300). Ideally that would be a single function, e.g.
read_instance_configs, that all other tools can use.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -26,6 +26,7 @@ from typing import Tuple
 
 import yaml
 from service_configuration_lib import pick_random_port
+from service_configuration_lib import read_extra_service_information
 from service_configuration_lib import read_yaml_file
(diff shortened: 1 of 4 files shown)


qui:
I'm going to keep working off this branch for PAASTA-16300. I thought it would be easier to review in separate chunks, but if you're not sure why I want this change, you can wait for the second PR.

solarkennedy:
This is cool. Thank you for this cleanup. We've come a long way!

qui:
👍 for reference, the 2nd PR is https://github.com/Yelp/paasta/pull/2683, which refactors instance config loading into another function called load_service_instance_config. 

I just want there to be a single way to get instance configs. load_service_instance_config if you need the whole config, and read_service_instance_names if you only need the instance name.

PR #2678, opened 2020-02-24
PAASTA-15540 Fix duplicated log lines on `paasta logs -f`

gvolco:
`run_code_over_scribe_envs()` invokes the `callback()` function for every logging component. But callback() ignores the `component` param and uses `components` instead.  This gets passed down to the log filtering functions, which end up matching log lines more than once.

This bug affects print_last_n_logs, print_logs_by_time and tail_logs. But the first two do not expose the problem thanks to [paasta/pull/2322](https://github.com/Yelp/paasta/pull/2322). 
Note that we still need to keep the fix on pull/2322 as scribe is still duplicating log lines when pulling data in non-streaming mode.

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -25,9 +25,17 @@ from multiprocessing import Queue
 from queue import Empty
 from time import sleep
 from typing import Any
+from typing import Callable
+from typing import ContextManager
 from typing import Dict
+from typing import Iterable
 from typing import List
+from typing import Mapping
+from typing import MutableSequence
+from typing import Sequence
 from typing import Set
+from typing import Tuple
+from typing import Type
 
 import isodate
 import pytz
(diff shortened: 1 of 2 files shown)


vkhromov:
It seems that the logging-related code lacks typing annotation, which partially caused this bug.
Maybe it's time to add typing annotations in this code to make sure that we aren't missing the same kind of errors?

paasta_tools/cli/cmds/logs.py
```diff
@@ -226,15 +239,15 @@ def check_timestamp_in_range(timestamp, start_time, end_time):
...
-    components,
-    clusters,
-    instances,
-    start_time=None,
-    end_time=None,
-):
+    line: str,
+    levels: Sequence[str],
+    service: str,
+    components: Union[str, Iterable[str]],
```
vkhromov: Could you please double-check that `componens` is a `Union` and not just `Iterable` here?

We use the `componens`  variable below for
```
parsed_line.get("component") in components
```
Wondering, how all of that is supposed to work if `components` has the `str` type and isn't equal to `"default"`...

gvolco: This particular filtering function will always receive components='default'. Let me give this another pass and clean up this part of the code.

PR #2680, opened 2020-02-24
Fix external metrics (for uwsgi/http)

EmanekaT:
https://github.com/Yelp/fullerite/pull/434

### DIFF
--- a/docs/source/hpa.rst
+++ b/docs/source/hpa.rst
@@ -110,7 +110,7 @@ When ``dimensions`` is not provided, the average values of all HTTP metrics expo
 You can find your HTTP metrics and its dimensions on SignalFX.
 
 When ``dimensions`` is provided, the value retrieved from signalfx with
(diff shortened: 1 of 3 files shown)


EmanekaT:
> You mention you're getting multiple metrics per pod. Are those for sidecar containers, etc? Would it be better to do `.sum(by="kubernetes_pod_name")` instead of `.mean(by=...)`? With `mean`, the average will decrease with every new sidecar.

Sorry for the ambiguity. There is one time series per pod. Like, it doesn't take average of metrics from all pods.

EvanKrall:
wouldn't `.mean(by="kubernetes_pod_name")` also produce one time series per kubernetes_pod_name (presumably different per pod)?

EmanekaT:
> wouldn't `.mean(by="kubernetes_pod_name")` also produce one time series per kubernetes_pod_name (presumably different per pod)?

Hmm, you are right. I should do `.mean(by="paasta_yelp_com_instance")`. Sorry, it's confusing.

PR #2683, opened 2020-02-25
Load instance configs everywhere with utils function that includes automated config

qui:
The design in PAASTA-16256 specifies that configuration from automated sources can be stored in a new subdirectory in each service's yelpsoa-configs. For example,
```
foo_service/
  marathon-norcal-devc.yaml
  service.yaml
  deploy.yaml
  auto/
    marathon-norcal-devc.yaml
```

The regular instance config in `marathon-norcal-devc.yaml` is deep-merged with `auto/marathon-norcal-devc.yaml`, such that keys are only used from `auto` if they are not specified in the regular file.

For example, given
```
marathon-norcal-devc.yaml
----
foo:
  cpus: 1
  mem: 20
bar:
  cpus: 2

auto/marathon-norcal-devc.yaml
----
foo:
  mem: 30
bar:
  mem: 30
```
The result would be `mem: 20` for `foo` and `mem: 30` for `bar`.

I've implemented this in `utils.load_service_instance_config(s)` and replaced all the places that we load instance configs with it. I only anticipate enabling auto configs for stateless marathon/kubernetes services this quarter, but I would like to keep the way we load instance configs consistent everywhere.

The next step is helpers to write the `auto/` values, which will include validating them. No instance types will be enabled anywhere except in mesosstage/kubestage for now, which I think makes this safe to release even before validation.

### DIFF
--- a/paasta_tools/flink_tools.py
+++ b/paasta_tools/flink_tools.py
@@ -20,14 +20,13 @@ import requests
 import service_configuration_lib
 from mypy_extensions import TypedDict
 
-from paasta_tools.kubernetes_tools import InvalidJobNameError
-from paasta_tools.kubernetes_tools import NoConfigurationForServiceError
 from paasta_tools.kubernetes_tools import sanitised_cr_name
 from paasta_tools.long_running_service_tools import LongRunningServiceConfig
 from paasta_tools.long_running_service_tools import LongRunningServiceConfigDict
 from paasta_tools.utils import BranchDictV2
 from paasta_tools.utils import deep_merge_dictionaries
 from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import load_service_instance_config
 from paasta_tools.utils import load_v2_deployments_json
 
 
@@ -113,22 +112,11 @@ def load_flink_instance_config(
     general_config = service_configuration_lib.read_service_configuration(
         service, soa_dir=soa_dir
     )
-    flink_conf_file = "flink-%s" % cluster
-    instance_configs = service_configuration_lib.read_extra_service_information(
-        service, flink_conf_file, soa_dir=soa_dir
+    instance_config = load_service_instance_config(
+        service, instance, "flink", cluster, soa_dir=soa_dir
     )
-
-    if instance.startswith("_"):
-        raise InvalidJobNameError(
-            f"Unable to load kubernetes job config for {service}.{instance} as instance name starts with '_'"
-        )
-    if instance not in instance_configs:
-        raise NoConfigurationForServiceError(
-            f"{instance} not found in config file {soa_dir}/{service}/{flink_conf_file}.yaml."
-        )
-
     general_config = deep_merge_dictionaries(
-        overrides=instance_configs[instance], defaults=general_config
+        overrides=instance_config, defaults=general_config
     )
 
     branch_dict: Optional[BranchDictV2] = None
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -69,7 +69,6 @@ import choice
 import dateutil.tz
 import requests_cache
 import service_configuration_lib
-import yaml
 from docker import Client
 from docker.utils import kwargs_from_env
 from kazoo.client import KazooClient
@@ -88,6 +87,7 @@ PATH_TO_SYSTEM_PAASTA_CONFIG_DIR = os.environ.get(
     "PAASTA_SYSTEM_CONFIG_DIR", "/etc/paasta/"
 )
 DEFAULT_SOA_DIR = service_configuration_lib.DEFAULT_SOA_DIR
+AUTO_SOACONFIG_SUBDIR = "autotuned_defaults"
 DEFAULT_DOCKERCFG_LOCATION = "file:///root/.dockercfg"
 DEPLOY_PIPELINE_NON_DEPLOY_STEPS = (
     "itest",
@@ -1717,6 +1717,7 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     api_endpoints: Dict[str, str]
     auth_certificate_ttl: str
     auto_hostname_unique_size: int
+    auto_config_instance_types_enabled: Dict[str, bool]
     boost_regions: List[str]
     cluster: str
     cluster_autoscaler_max_decrease: float
@@ -1929,6 +1930,9 @@ class SystemPaastaConfig:
         :returns: The integer size of a small service
         """
         return self.config_dict.get("auto_hostname_unique_size", -1)
+
+    def get_auto_config_instance_types_enabled(self) -> Dict[str, bool]:
+        return self.config_dict.get("auto_config_instance_types_enabled", {})
 
     def get_api_endpoints(self) -> Mapping[str, str]:
         return self.config_dict["api_endpoints"]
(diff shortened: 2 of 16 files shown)


paasta_tools/utils.py
```diff
@@ -88,6 +87,7 @@
     "PAASTA_SYSTEM_CONFIG_DIR", "/etc/paasta/"
 )
 DEFAULT_SOA_DIR = service_configuration_lib.DEFAULT_SOA_DIR
+AUTO_SOACONFIG_SUBDIR = "auto"
```
solarkennedy: Has there been any discussion yet for making this a hidden dir, and also giving a more descriptive name, like `.automatically_applied_overrides`?

paasta_tools/utils.py
```diff
@@ -2759,8 +2763,63 @@ def get_services_for_cluster(
...
+        service, instance_type, cluster, soa_dir
+    ).get(instance, {})
+    return deep_merge_dictionaries(overrides=user_config, defaults=auto_config,)
+
+
+def load_service_instance_auto_configs(
+    service: str, instance_type: str, cluster: str, soa_dir: str = DEFAULT_SOA_DIR,
+) -> Dict[str, Dict[str, Any]]:
+    enabled_types = load_system_paasta_config().get_auto_config_instance_types_enabled()
+    conf_file = f"{instance_type}-{cluster}"
```
solarkennedy: It is too bad this *must* be a yaml file with this function. Do you think if our "api" here was a json file, people would be less inclined to think that it is a file they could edit by hand?

paasta_tools/utils.py
```diff
@@ -2759,8 +2763,63 @@ def get_services_for_cluster(
...
+    auto_config = load_service_instance_auto_configs(
+        service, instance_type, cluster, soa_dir
+    ).get(instance, {})
+    return deep_merge_dictionaries(overrides=user_config, defaults=auto_config,)
+
+
+def load_service_instance_auto_configs(
+    service: str, instance_type: str, cluster: str, soa_dir: str = DEFAULT_SOA_DIR,
+) -> Dict[str, Dict[str, Any]]:
+    enabled_types = load_system_paasta_config().get_auto_config_instance_types_enabled()
```
solarkennedy: As designed, this could be a little expensive, loading this file on every config. Can you check to make sure this won't be a big performance hit? (not sure if there is an easy benchmark, `paasta validate` maybe?)

paasta_tools/utils.py
matfra: I do like the idea of json for a file which is a machine to machine interface

paasta_tools/utils.py
matfra: I'm not against this being a .auto or .recommended_values or .reco
@solarkennedy they are not overrides because they only apply if a key is missing in the parent dir files. They aren't default either because a default value (which is a feature we still need to implement, typically for new services) is by definition not taylored/optimized for a given service.
Naming is hard

solarkennedy: `.automatically_tuned_settings`?

paasta_tools/utils.py
qui: I think if we get the name of the directory right, people won't be too inclined to edit it by hand. 

JSON does sound better for performance, though, I'll try that out.

paasta_tools/utils.py
qui: Sure, worst case it doubles the config loading time (if we keep it as YAML). It will be less in practice because the auto/ files will contain less config than the regular ones.

I simulated that by copying all the marathon* files in yelp-main into the auto/ directory. With this change, `paasta validate -s yelp-main` takes 37.744s, without 29.212s. 

I also looked at how long it takes to load all marathon instances. Again, I copied the marathon files into auto/ for every service to simulate the worst case:
```
service_configuration_lib.disable_yaml_cache()

for service in list_services(soadir):
    for cluster in clusters:
        instances = read_service_instance_names(service, "marathon", cluster, soadir)
        for _, instance in instances:
            config = load_service_instance_config(service, instance, "marathon", cluster, soadir)
```
This takes 1m52.380s. As expected, that's just about 2x the time as it takes to load everything before this PR. There are about 7000 service instances, so that's about 8ms per service config before, 16ms after. Even if it's 2x slower, a few extra milliseconds seems fine to me.

solarkennedy: ty for confirming.
JSON will help if we do that too.

I was more concerned about load_system_paasta_config, but looking at that, it is cached, which is great!

paasta_tools/utils.py
qui: I don't think we should make it a hidden directory. If it's possible that a developer would want to override something in them, I think they should be able to easily figure out their current settings. It's easy enough to find a setting like cpus from other sources, but for things like autoscaling, yelpsoa-configs would be the only place you could see that.

How about... `autotuned_default_configs`?

paasta_tools/utils.py
qui: I'll consider JSON as a follow-up (PAASTA-16348).

paasta_tools/utils.py
ddelnano: I agree that making it a hidden directory will hurt  a developer's ability to find the value as @qui mentioned.

As for the name I agree with @matfra's opinion that the human values are the overrides so I like the theme of "automatically tuned" and "defaults". So +1 for @qui's suggestion of `autotuned_default_configs` or other similar name like `autotuned_defaults` or `automatically_tuned_defaults`

qui: I went with `autotuned_defaults`, thanks @ddelnano!

paasta_tools/flink_tools.py
```diff
@@ -113,22 +112,11 @@ def load_flink_instance_config(
...
-        )
-    if instance not in instance_configs:
-        raise NoConfigurationForServiceError(
-            f"{instance} not found in config file {soa_dir}/{service}/{flink_conf_file}.yaml."
-        )
-
     general_config = deep_merge_dictionaries(
-        overrides=instance_configs[instance], defaults=general_config
+        overrides=instance_config, defaults=general_config
     )
```
EvanKrall: I'm seeing a lot of copies of these three lines -- does it make sense to move them into `load_service_instance_config` so they can be DRYed?

paasta_tools/utils.py
```diff
@@ -2767,8 +2771,63 @@ def get_services_for_cluster(
     return instance_list
 
 
-def parse_yaml_file(yaml_file: str) -> Any:
-    return yaml.safe_load(open(yaml_file))
+def load_service_instance_configs(
```
EvanKrall: now that this exists I wonder if we can deprecate PaastaServiceConfigLoader, which seems like an overly OOP version of basically this function.

paasta_tools/flink_tools.py
qui: Yeah, I tried this, and the main reason I didn't was that the PaastaServiceConfigLoader needed something different. When I look into removing it, I'll consider this too.

paasta_tools/utils.py
qui: I'll follow up on this (PAASTA-16358) - will take some work to make sure that the new function replaces its functionality everywhere it is used.

EvanKrall: yeah definitely low-priority

PR #2684, opened 2020-02-25
Add k8s deployment errors to paasta status (-vv or -vvv)

kawaiwanyelp:
### Description
I've made it so that when using -vv or -vvv with `paasta status` for a k8s service, k8s errors will be included, in addition to any stdout or stderr. It looks like [this] (https://gist.github.com/kawaiwanyelp/5bfd63bbfd1c7d440099af4ff838b1ac).

The errors and instance output are per container per pod. Pods themselves don't record any aggregated error states for its containers. So, even if a container is in a crash loop, the pod may still be "healthy" (we show this in `paasta status`, and I'd like to fix that in a separate PR).

### Testing done
`make test`
manual testing by making a local `paasta-api` process talk to a real dev k8s cluster

### DIFF
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -1292,9 +1292,25 @@
                     "type": "string",
                     "description": "The status of the pod"
                 },
+                "containers": {
+                    "type": "array",
+                    "items": {
+                        "$ref": "#/definitions/KubernetesContainer",
+                        "description": "List of containers in the pod"
+                    }
+                }
+            }
+        },
+        "KubernetesContainer": {
+            "type": "object",
+            "properties": {
+                "name": {
+                    "type": "string",
+                    "description": "Name of the container"
+                },
                 "tail_lines": {
                     "$ref": "#/definitions/TaskTailLines",
-                    "description": "Stdout and stderr tail of the task"
+                    "description": "Stdout and stderr tail of the container"
                 }
             }
         },
--- a/paasta_tools/instance/kubernetes.py
+++ b/paasta_tools/instance/kubernetes.py
@@ -16,7 +16,7 @@ from paasta_tools import kubernetes_tools
 from paasta_tools import marathon_tools
 from paasta_tools import smartstack_tools
 from paasta_tools.cli.utils import LONG_RUNNING_INSTANCE_TYPE_HANDLERS
-from paasta_tools.kubernetes_tools import get_tail_lines_for_kubernetes_pod
+from paasta_tools.kubernetes_tools import get_tail_lines_for_kubernetes_container
 from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
 from paasta_tools.long_running_service_tools import LongRunningServiceConfig
 from paasta_tools.long_running_service_tools import ServiceNamespaceConfig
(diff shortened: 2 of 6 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1378,29 +1379,64 @@ def list_deployments(
...
+                name=pod.metadata.name,
+                namespace=pod.metadata.namespace,
+                container=container.name,
+                tail_lines=num_tail_lines,
+            )
+            lines["stdout"].extend(log.split("\n"))
+        except ApiException as e:
+            # there is a potential race condition in which a pod's containers
+            # have not failed, but have when we get the container's logs. in this
+            # case, use the error from the exception, though it is less accurate.
```
chlgit: What's if we move the check for container.state above here? In this way, we won't have race condition here, right?

paasta_tools/api/api_docs/swagger.json
```diff
@@ -1293,8 +1293,38 @@
...
+                    "description": "Stdout and stderr tail for each container in the pod"
+                }
+            }
+        },
+        "PodTailLines": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "properties": {
+                    "container": {
```
chlgit: Could we make "container" as the properties for kubernetesPod? By doing that, we don't need to define PodTailLines.

paasta_tools/kubernetes_tools.py
kawaiwanyelp: It won't because the state check is on the current in-memory pod object, which can fall behind the actual pod object in Kubernetes. Getting the pod/container logs with `read_namespaced_pod_log` does not update the pod object we have. It's the same race condition as PaaSTA having a pod object for a deleted pod, meaning PaaSTA thinks it exists but k8s has already moved on.

In order to get an update, we need to make a call to the k8s API to retrieve, which I don't think is worth it here. Users can just call `paasta status` again to get the update.

chlgit: Ah, got it.

paasta_tools/api/api_docs/swagger.json
kawaiwanyelp: I've added a containers array to the `KubernetesPod` object in the swagger spec, which includes the `tail_lines`. I agree it's more intuitive.

paasta_tools/instance/kubernetes.py
```diff
@@ -91,20 +91,22 @@ def set_cr_desired_state(
         num_tail_lines = calculate_tail_lines(verbose)
 
         for pod in pod_list:
-            if num_tail_lines > 0:
-                tail_lines = await get_tail_lines_for_kubernetes_pod(
-                    client, pod, num_tail_lines
+            containers = [
```
keymone: Any reason this is not a dictionary?

kawaiwanyelp: It's because in the Swagger spec, I created KubernetesContainer object, which I think is nice for ensuring there is always a name associated with the rest of a container's details (for now its only the tail lines). The list of dictionaries is here just the server-side representation of that object. I don't mind changing it to a dictionary instead though I think it wouldn't be as nice.

paasta_tools/kubernetes_tools.py
```diff
@@ -1376,31 +1378,71 @@ def list_deployments(
...
+    return tail_lines
+
+
+def format_tail_lines_for_kubernetes_pod(
+    pod_containers: Sequence, pod_name: str,
+) -> List[str]:
+    rows: List[str] = []
+    for container in pod_containers:
+        error_message = container.tail_lines.error_message
+        if error_message is not None:
```
chlgit: nit: if container.tail_lines.error_message is not None and replace error_message later? I don't think error_message make it much clear here.

PR #2685, opened 2020-02-25
Fix spark_mesos_secret shared secret usage

jangsutsr:


paasta_tools/tron_tools.py
```diff
@@ -71,7 +71,8 @@
     )["definitions"]["job"]["properties"]["monitoring"]["properties"].keys()
 )
 MESOS_EXECUTOR_NAMES = ("paasta", "spark")
-SPARK_MESOS_SECRET_KEY = "SHARED_SECRET(SPARK_MESOS_SECRET)"
+SPARK_MESOS_SECRET_KEY = "PAASTA_SECRET_SPARK_MESOS_SECRET"
+SPARK_MESOS_SECRET_VAL = "SHARED_SECRET(SPARK_MESOS_SECRET)"
```
solarkennedy: Should be `$PAASTA_SECRET_SPARK_MESOS_SECRET` right? (because it is a shell expansion)

jangsutsr: Good news is we don't need all of these: they are all moving to `service_configuration_lib` https://github.com/Yelp/service_configuration_lib/pull/41/files

PR #2688, opened 2020-02-27
allow setting max_unavailable in the system_paasta_config

drmorr0:
this will enable us to have spotinst scale us down a bit more intelligently


drmorr0:
For some reason travis didn't pick up the new SHA.  But the tests pass locally now.

PR #2689, opened 2020-02-27
Set sane timeouts and a friendly error message when the Mesos leader is unavailable

solarkennedy:
The big win here is adding a timeout here. 
When running from a devbox it "only" fails in 15 seconds (3 * 5) instead of 15 minutes.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -18,6 +18,7 @@ from paasta_tools.cli.utils import list_instances
 from paasta_tools.cli.utils import pick_random_port
 from paasta_tools.clusterman import get_clusterman_metrics
 from paasta_tools.mesos_tools import find_mesos_leader
+from paasta_tools.mesos_tools import MesosLeaderUnavailable
 from paasta_tools.spark_tools import DEFAULT_SPARK_SERVICE
 from paasta_tools.spark_tools import get_aws_credentials
 from paasta_tools.spark_tools import get_default_event_log_dir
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -423,8 +424,15 @@ def get_spark_config(
         user_args["spark.eventLog.enabled"] = "true"
         user_args["spark.eventLog.dir"] = default_event_log_dir
 
+    try:
+        mesos_address = find_mesos_leader(args.cluster)
+    except MesosLeaderUnavailable:
+        print(
+            "Couldn't reach the Mesos leader from here. Are you running this command from the right environment?",
```
kawaiwanyelp: nit: "right environment" may not be obvious enough for developers. I'd put something like "(e.g. attempting to access prod from dev)".

paasta_tools/mesos_tools.py
```diff
@@ -165,7 +170,12 @@ def find_mesos_leader(cluster):
         raise ValueError("Mesos master is required to find leader")
 
     url = f"http://{master}:{MESOS_MASTER_PORT}/redirect"
-    response = requests.get(url)
+    try:
+        # Timeouts here are for connect, read
+        response = requests.get(url, timeout=(5, 30))
+    except Exception as e:
+        print(f"Error finding the mesos leader: {e}", file=sys.stderr)
+        raise MesosLeaderUnavailable(e)
```
kawaiwanyelp: Embedding the original exception inside `MesosLeaderUnavailable` will obscure the original exception in a print-out. I think it should be `raise MesosLeaderUnavailable() from e` instead.

solarkennedy: I don't really understand. My motivation here is that the original error (timeout, failed to connect, whatever it is) still gets printed to stderr in some way.

If I do `raise MLU from e` here, would you give a shipit if I printed the `e` in paasta spark run?

kawaiwanyelp: My motivation for chaining here is that I'd like to see the original exception class itself printed as well as well as the traceback to see where it happened. Often times, logging only the message is not quite enough for debugging, though this case it's less important.

kawaiwanyelp: Hm, given the output, I'm fine without the chaining. As long as the original exception is visible, even without a traceback, I'm happy with it.

PR #2694, opened 2020-03-03
Show real hostname for k8s pods in paasta status

kawaiwanyelp:
### Description
Currently, we just show pod hostnames in `paasta status` like `ip-00-00-00-00.us-west-1.compute.internal`, but this review makes it so that real hostnames are shown instead, making them directly copyable for SSH-ing.

### Testing
`make test`
Manual testing with a local paasta-api connected to a test k8s cluster

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2014,3 +2014,14 @@ def set_cr_desired_state(
     kube_client.custom.replace_namespaced_custom_object(**cr_id, body=cr)
     status = cr.get("status")
     return status
+
+
+def get_pod_hostname(kube_client: KubeClient, pod: V1Pod) -> str:
+    """Gets the hostname of a pod's node from labels"""
+    try:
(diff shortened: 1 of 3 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -2014,3 +2014,14 @@ def set_cr_desired_state(
     kube_client.custom.replace_namespaced_custom_object(**cr_id, body=cr)
     status = cr.get("status")
     return status
+
+
+def get_pod_hostname(kube_client: KubeClient, pod: V1Pod) -> str:
+    """Gets the hostname of a pod's node from labels"""
+    try:
+        node = kube_client.core.read_node(name=pod.spec.node_name)
```
solarkennedy: Do you know what the performance implications are of this, esp on larger services that have lots of pods? (hundreds)

I'm concerned about making this command fan-out with N api calls as it gets larger.

In other words, I want to make sure that our most important (large) services can run paasta status just as fast and error-proof as our smaller services.

PR #2695, opened 2020-03-04
use batch name as spark app name 

lydian:
Providing a more meaningful/distinguishable name so that we can better search the job on history server

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -2,6 +2,7 @@ import argparse
 import json
 import logging
 import os
+import re
 import socket
 import sys
 import time
@@ -301,6 +302,11 @@ def add_subparser(subparsers):
     list_parser.set_defaults(command=paasta_spark_run)
 
 
+def sanitize_container_name(container_name):
+    # container_name only allows [a-zA-Z0-9][a-zA-Z0-9_.-]
+    return re.sub("[^a-zA-Z0-9_.-]", "_", re.sub("^[^a-zA-Z0-9]+", "", container_name))
+
+
 def get_docker_run_cmd(container_name, volumes, env, docker_img, docker_cmd, nvidia):
     cmd = ["paasta_docker_wrapper", "run"]
     cmd.append("--rm")
@@ -315,7 +321,7 @@ def get_docker_run_cmd(container_name, volumes, env, docker_img, docker_cmd, nvi
             cmd.append("--tty=true")
 
     cmd.append("--user=%d:%d" % (os.geteuid(), os.getegid()))
-    cmd.append("--name=%s" % container_name)
+    cmd.append("--name=%s" % sanitize_container_name(container_name))
     for k, v in env.items():
         cmd.append("--env")
         if k in SENSITIVE_ENV:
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -594,6 +594,21 @@ def run_docker_container(
...
+def get_spark_app_name(original_docker_cmd):
+    # Use submitted batch name as default spark_run job name
+    if "spark-submit" in original_docker_cmd:
+        parser = argparse.ArgumentParser()
+        parser.add_argument("cmd", choices=["spark-submit"])
+        parser.add_argument("others", nargs="+")
+        args, _ = parser.parse_known_args(original_docker_cmd)
+        scripts = [arg for arg in args.others if arg.endswith(".py")]
+        if len(scripts) > 0:
+            batch_name = scripts[0].split("/")[-1].replace(".py", "")
```
solarkennedy: I'm a bit worried about other characters slipping through here.
You are assuming there is only a .py file. 

Do you know what the actual limitations of the spark_app_name are? If so, can you make a `sanitize_batch_name` function to do what happens on this line, but in a more robust way?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -617,13 +632,13 @@ def configure_and_run_docker_container(
...
             )
 
+    original_docker_cmd = args.cmd or instance_config.get_cmd()
     spark_ui_port = pick_random_port(args.service + str(os.getpid()))
-    spark_app_name = "paasta_spark_run_{}".format(get_username())
+    spark_app_name = get_spark_app_name(original_docker_cmd)
+
     container_name = spark_app_name + "_" + str(spark_ui_port)
-    original_docker_cmd = args.cmd or instance_config.get_cmd()
     if "jupyter" not in original_docker_cmd:
```
solarkennedy: Can you unify the logic so that you can also account for `jupyter` in the new function you made?

paasta_tools/cli/cmds/spark_run.py
lydian: Not really, the code will use the first xxx.py file, because spark-submit require the next argument to be the script we want to run, and then follow by the script's opts and args.  Therefore using the first one should be good enough. 

Also, the spark-submit doesn't recognize python module, and therefore we cannot do something like "spark-submit python -m module.path.to.script" to run the code. Which means we are always able to have a .py file in the argument.  The only exception is if people are running with script written in other language, and therefore is unable to find argument ends with .py, but In that case, we will simply return "paasta_spark_run" as what we have so far.

solarkennedy: It sounds like you covered the known knowns.
In general I'm always very defensive against user-inputed filenames. Eventually this code will have to handle emoji, non-ascii characters, and uncode left-to-right characters! (j/k maybe)

paasta_tools/cli/cmds/spark_run.py
lydian: haha, I actually found this "jupyter" is not really in use. because we rename the command to be "jupyter-lab". 

It also seems fine to append port to jupyter-lab, so I am removing this section. However, in order to identify the app from spark history server, I am also making all these type to have name starts as "paasta_jupyter"

paasta_tools/cli/cmds/spark_run.py
lydian: one thing I noticed is that the run_docker_container will use the spark app name as its `--name` , but docker container does  require name to be [a-zA-Z0-9][a-zA-Z0-9_.-]. Therefore I made a separate sanitize_container_name, so that it won't break the docker run.  Wondering if it is a better idea to do the sanitize in the spark app instead.  The spark doesn't seem to limit anything on the naming, and therefore it is not going to cause problem on the spark side.

PR #2696, opened 2020-03-04
Implement deploy_whitelist and deploy_blacklist for k8s pods as node affinities

kawaiwanyelp:
### Description
To support `deploy_blacklist` and `deploy_whitelist` for k8s instances, I've implemented logic to convert those configs to node affinities. I've also taken the liberty of converting the pool selector (previously implemented as a node selector) into a node affinity so that it's all unified.

### Testing
`make test`
Manual testing using `setup_kubernetes_job` to create test pods with the new node affinities

### Notes
While testing, I found out that `kubectl describe pod` _does not_  show affinities, only selectors. However, it will still show failures if the affinities result in no scheduleable nodes, which can be confusing. It is possible to see them using `kubectl get pod -o json`, which shows the raw pod configs, and I've documented this in the code. I haven't seen any reported issues so it doesn't seem like anyone is going to get around to "fixing" this issue.

Also, given that I changed the `test_warning_big_bounce` SHA, it appears that releasing this will cause a big bounce, at least for k8s instances.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -39,6 +39,7 @@ from humanfriendly import parse_size
 from kubernetes import client as kube_client
 from kubernetes import config as kube_config
 from kubernetes.client import models
+from kubernetes.client import V1Affinity
 from kubernetes.client import V1AWSElasticBlockStoreVolumeSource
 from kubernetes.client import V1beta1PodDisruptionBudget
 from kubernetes.client import V1beta1PodDisruptionBudgetSpec
@@ -61,6 +62,10 @@ from kubernetes.client import V1LabelSelector
 from kubernetes.client import V1Lifecycle
 from kubernetes.client import V1Namespace
 from kubernetes.client import V1Node
+from kubernetes.client import V1NodeAffinity
+from kubernetes.client import V1NodeSelector
+from kubernetes.client import V1NodeSelectorRequirement
+from kubernetes.client import V1NodeSelectorTerm
 from kubernetes.client import V1ObjectFieldSelector
 from kubernetes.client import V1ObjectMeta
 from kubernetes.client import V1PersistentVolumeClaim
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1155,7 +1160,7 @@ def get_pod_template_spec(
                     service_namespace_config=service_namespace_config,
                 ),
                 share_process_namespace=True,
-                node_selector=self.get_node_selector(),
+                affinity=V1Affinity(node_affinity=self.get_node_affinity(),),
```
chlgit: Could we use node_selector for pool lable and use affinity for others, such as whitelist, blacklist and constraint?

paasta_tools/kubernetes_tools.py
```diff
@@ -1164,8 +1169,42 @@ def get_pod_template_spec(
             ),
         )
 
-    def get_node_selector(self) -> Mapping[str, str]:
-        return {"yelp.com/pool": self.get_pool()}
+    def get_node_affinity(self) -> V1NodeAffinity:
+        """Converts deploy_whitelist and deploy_blacklist in node affinities.
+
+        note: At the time of writing, `kubectl describe` does not show affinities,
+        only selectors. To see affinities, use `kubectl get pod -o json` instead.
```
chlgit: This is another reason I would like to keep node selectors. At least, we can get more info for `kubectl describe`.

paasta_tools/kubernetes_tools.py
kawaiwanyelp: I'll move the pool affinity back to a selector. As a side note, constraints as affinities might not be a good fit, but we can discuss more in another PR.

tests/test_kubernetes_tools.py
```diff
@@ -1062,11 +1073,73 @@ def test_get_pod_template_spec_smartstack_service(self):
...
                     node_selector={"yelp.com/pool": "default"},
+                    affinity=V1Affinity(
+                        node_affinity=mock_get_node_affinity.return_value,
+                    ),
                     restart_policy="Always",
                     volumes=[],
                 ),
             )
 
+    @pytest.mark.parametrize(
```
chlgit: nit: I personally prefer three tests rather than using parametrize. This is a little hard to read for me. Also, by doing that, the extra bonus is you can use test name to point out what you really want to test, so no comments needed.

PR #2699, opened 2020-03-04
Tools for committing autotuned_default configs

qui:
config_utils.AutoConfigUpdater is my planned interface for all scripts that want to write to `autotuned_defaults`. The docstring has an example of general usage, and `paasta_tools/contrib/rightsizer_soaconfigs_update.py`is an implementation for the rightsizer searches we have so far. 

For validation, I would like to skip the regular yelpsoa-configs pre-receive validation with an `f/` branch, because the pre-receive validation hooks take too long when a commit affects many services. I'm counting on the validate() method to catch all issues before we push changes.

Tested with
```
$ git clone git@sysgit.yelpcorp.com:yelpsoa-configs ~/soa-remote --bare
$ python -m paasta_tools.contrib.rightsizer_soaconfigs_update --splunk-creds <redacted> --csv-report paasta_rightsizer_memory_recs.csv --git-remote ~/soa-remote --branch f/qui/test1 -v
```


qui:
The `u/qui/test-script` branch on yelpsoa-configs has a diff I generated from the memory recs, if you want to check it out.

paasta_tools/config_utils.py
```diff
@@ -0,0 +1,198 @@
...
+        else:
+            log.error(f"Push to {branch} failed with:\n {e.stdout}")
+            raise
+
+
+def validate_auto_config_file(filepath: str):
+    basename = os.path.basename(filepath)
+    for file_type in KNOWN_CONFIG_TYPES:
+        if basename.startswith(file_type):
+            return bool(validate_schema(filepath, f"auto/{file_type}"))
```
sergsyelp: Is this still 'auto'? Not 'autotuning_defaults'? Maybe we should put the name into a constant?

tests/test_config_utils.py
```diff
@@ -0,0 +1,190 @@
...
+    with open(expected_path) as f:
+        assert yaml.safe_load(f) == {"a": 2}
+
+
+@mock.patch("paasta_tools.config_utils.validate_schema", autospec=True)
+def test_validate_auto_config_file_config_types(mock_validate, tmpdir):
+    for config_type in config_utils.KNOWN_CONFIG_TYPES:
+        filepath = f"service/{config_type}-cluster.yaml"
+        config_utils.validate_auto_config_file(filepath)
+        mock_validate.assert_called_with(filepath, f"auto/{config_type}")
```
sergsyelp: Needs update?

paasta_tools/config_utils.py
qui: It's just the path for the schemas, so it doesn't matter too much - I didn't think we needed to use the full name for internal stuff. But I can definitely use the constant from paasta_tools.utils if you think that would be better.

sergsyelp: Yes, I'd use a constant just in case we change our mind and need to rename the directory and if I to revisit this code in a half of a year it would be easier to connect logically related parts if the naming is consistent. Thanks!

paasta_tools/contrib/rightsizer_soaconfigs_update.py
```diff
@@ -0,0 +1,122 @@
...
+    with AutoConfigUpdater(
+        source_identifier, args.git_remote, branch=args.branch, tmp_dir="/nail/tmp"
+    ) as updater:
+        for (service, extra_info), instance_recommendations in results.items():
+            existing_recommendations = updater.get_existing_configs(service, extra_info)
+            for instance_name, recommendation in instance_recommendations.items():
+                existing_recommendations.setdefault(instance_name, {})
+                existing_recommendations[instance_name].update(recommendation)
+            updater.write_configs(service, extra_info, existing_recommendations)
+        updater.commit_to_remote(extra_message=extra_message)
```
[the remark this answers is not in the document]
qui: I added a `--push-to-remote` option so the default is dry-run, to reduce the chance anyone manually running the script unintentionally pushes changes. I also added a `--local-dir` option.

The script will do what you want (no git ops, just write files locally) if you specify `--local-dir` and don't add `--push-to-remote`.

paasta_tools/cli/schemas/auto/kubernetes_schema.json
```diff
@@ -0,0 +1,26 @@
...
+    "minProperties": 1,
+    "patternProperties": {
+        "^([a-z0-9]|[a-z0-9][a-z0-9_-]*[a-z0-9])*$": {
+            "type": "object",
+            "additionalProperties": false,
+            "minProperties": 1,
+            "properties": {
+                "cpus": {
+                    "type": "number",
+                    "minimum": 0,
```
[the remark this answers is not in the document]
matfra: yep, note that the the current memory recs don't seem to have the minimum enforced so I'm happy to leave it at 128, or even 64 in the code and 0.01 for CPU seems reasonable too

qui:
> 1. Some way of avoiding frequent minor fluctuations of configs, e.g.  1000mb vs 1050mb. Maybe rounding to some reduced set of settings, or having the recommender generate a good-enough-window where it won't bother changing a setting.

The Splunk searches contain a good enough window of 10% for cpus and 5% for mem. We may have to tune that, and I'll think about how we could add that guardrail in paasta-tools as well. 

> 2. some variant of service style/type/category, for informing the recommendations. The right cpus recommendation for a python uWSGI service very different from the right recommendation for an asyncio service.

For now, the rightsizer distinguishes between uswgi and cpu based autoscaling. @matfra says: "If CPU is used as an auto scaling metric, then we aim to adjust the CPU of the container so that the usage will be as close as possible to the setpoint. If the service is using another signal to autoscale (like UWSGI workers usage), then look at the P80 of usage over time and we add 40% of margin."

PR #2700, opened 2020-03-05
allow min_instance to be 0

EmanekaT:
Right now, if min_instance is 0, HPA won't be created.


EmanekaT:
Actually I think we'd rather not change due to https://github.com/kubernetes/kubernetes/issues/69687

PR #2701, opened 2020-03-05
Make pods with at least one container down properly show up as unhealthy

kawaiwanyelp:
### Description
Currently, if a pod's phase is "Running", we report it is as "Healthy". However, the "Running" phase just means that the pod is scheduled, its containers have (previously) been created. However, this can be a bit misleading, especially if all it's pods are in the waiting state, perhaps in a CrashLoopBackoff. 

I've made it so that in this case, we report it as Unhealthy instead of Healthy by checking for error messages.

### Testing done
manual testing against a failing instance


paasta_tools/api/api_docs/swagger.json
```diff
@@ -1295,6 +1295,10 @@
                     "type": "string",
                     "description": "name of the pod in Kubernetes"
                 },
+                "host": {
```
chlgit: This is from another PR, right? rebase maybe?

paasta_tools/cli/cmds/status.py
```diff
@@ -655,6 +655,11 @@ def format_kubernetes_pod_table(pods):
             health_check_status = PaastaColors.grey("N/A")
         elif pod.phase == "Running":
             health_check_status = PaastaColors.green("Healthy")
+            for container in pod.containers:
+                # a container is waiting or terminated
+                if container.tail_lines.error_message:
+                    health_check_status = PaastaColors.red("Unhealthy")
+                    break
```
chlgit: I sort of worry this would cause the inconsistent view between paasta status and the replication check. Could you make [is_pod_ready](https://sourcegraph.yelpcorp.com/mirrors/Yelp/paasta/-/blob/paasta_tools/kubernetes_tools.py#L1623) as the only function to check if the pod is healthy.

paasta_tools/api/api_docs/swagger.json
kawaiwanyelp: No it isn't the swagger spec is out of date.

PR #2704, opened 2020-03-06
Make get_pod_hostname return NotScheduled if pod has not yet been scheduled

kawaiwanyelp:
### Description
In #2684, I made `paasta status` check the pod container statuses for error messages. And in #2694, I made `paasta status` get the real hostname for pods, instead of the node names. In both cases, if there were no containers yet (i.e. the pod hasn't been scheduled yet), it would cause an Exception because `pod.status.container_statuses` and `pod.spec.node_name` would be `None`. This PR fixes the issue by checking if they are `None` first before doing anything with them.

### Testing
`make test`
manual testing on a test cluster with a node the had to run on a non-existent pool (couldn't be scheduled)

### Notes
On a broader note, when inspecting k8s objects, it is important to check that whether or not they are `None` or not. Most properties are optional, according to the k8s documentation.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2005,6 +2005,8 @@ def set_cr_desired_state(
 
 def get_pod_hostname(kube_client: KubeClient, pod: V1Pod) -> str:
     """Gets the hostname of a pod's node from labels"""
+    if not pod.spec.node_name:  # can be none, if pod not yet scheduled
+        return "NotScheduled"
     try:
         node = kube_client.core.read_node(name=pod.spec.node_name)
     except ApiException:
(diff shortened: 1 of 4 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1300,7 +1300,8 @@ def get_kubernetes_services_running_here_for_nerve(
 class KubeClient:
     def __init__(self) -> None:
         kube_config.load_kube_config(
-            config_file=os.environ.get("KUBECONFIG", KUBE_CONFIG_PATH),
+            # config_file=os.environ.get("KUBECONFIG", KUBE_CONFIG_PATH),
```
solarkennedy: this was just for testing right?

solarkennedy:
I don't think you need to comment on every line, in my experience `foo = bar or []` is a pretty common idiom.

paasta_tools/kubernetes_tools.py
kawaiwanyelp: Yeah :P

PR #2705, opened 2020-03-09
Ensure a k8s affinity object is added to pods only if there are affinities to set

kawaiwanyelp:
### Description
I previously merged and reverted #2696 after discovering that having an empty affinity object on a pod resulted in all nodes being deselected from deployment. This PR reverts that revert and accounts for the "no affinities" case by only adding an affinity object to a pod spec if there are node affinities to add. I've also added a number of tests to account for that case.

### Testing
`make test`
Tested by scheduling a pod with no deploy whitelists or blacklists on a test cluster

### Notes
Since releasing #2696 caused an incident, I'd like to make a release and test it in kubestage before unpinning everywhere.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -39,6 +39,7 @@ from humanfriendly import parse_size
 from kubernetes import client as kube_client
 from kubernetes import config as kube_config
 from kubernetes.client import models
+from kubernetes.client import V1Affinity
 from kubernetes.client import V1AWSElasticBlockStoreVolumeSource
 from kubernetes.client import V1beta1PodDisruptionBudget
 from kubernetes.client import V1beta1PodDisruptionBudgetSpec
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1134,6 +1139,28 @@ def get_pod_template_spec(
...
+            volumes=self.get_pod_volumes(
+                docker_volumes=docker_volumes,
+                aws_ebs_volumes=self.get_aws_ebs_volumes(),
+            ),
+        )
+        # need to check if there are node selectors/affinities. if there are none
+        # and we create an empty affinity object, k8s will deselect all nodes.
+        node_affinity = self.get_node_affinity()
+        if node_affinity is not None:
+            pod_spec_kwargs["affinity"] = V1Affinity(node_affinity=node_affinity)
```
qui: What if the method was just `get_affinity` and returned an Optional[V1Affinity]? I think it would make the code simpler.

stug:
> Looks roughly good to me, but we should think about whether we can test this somehow before we deploy.

Oops looks like you already did this -- carry one :)

PR #2709, opened 2020-03-10
Script that sets all k8s deployments to max_instances

EvanKrall:


paasta_tools/long_running_service_tools.py
```diff
@@ -244,26 +244,35 @@ def get_healthcheck_mode(
...
+        except NoNodeError:
+            log.debug("No zookeeper data, returning None")
+            return None
+
+    def set_autoscaled_instances(self, instance_count: int) -> None:
+        """Set the number of instances in the same way that the autoscaler does."""
+        set_instances_for_marathon_service(
+            service=self.service, instance=self.instance, instance_count=instance_count,
+        )
+
```
EvanKrall: This is unused, for now  -- the idea being that @enochtangg can make the autoscaling API endpoint use this / the KubernetesDeploymentConfig version of it.

enochtangg: @EvanKrall Just curious, would it make sense to keep these methods in `MarathonServiceConfig()`, and replicate the same functionality in `KubernetesServiceConfig()`? Then just have the parent class `LongRunningServiceConfig()` as not implemented? Not sure how the convention was set up here, but do we want marathon/k8s specific methods in LongRunningServiceConfig anyways?

EvanKrall: Yeah, I think that makes sense, though the native scheduler uses this, and it looks less delete-able than I thought. Lemme do a quick refactor.

paasta_tools/autoscaling/max_all_k8s_services.py
```diff
@@ -0,0 +1,42 @@
...
+        ):
+            # Force min_instances = max_instances so that we scale up.
+            max_instances = instance_config.get_max_instances()
+            if max_instances is not None:
+                instance_config.config_dict["min_instances"] = max_instances
+                formatted_application = instance_config.format_kubernetes_app()
+                wrapper = get_application_wrapper(formatted_application)
+                wrapper.soa_config = instance_config
+                print(f"Scaling up {service}.{instance_config.instance}")
+                wrapper.update(kube_client)
```
EmanekaT: Since you change min_instances to max_instances, in get_autoscaling_metric_spec, HPA sets min_replicas to max_replicas. And with limit_instance_count function, replicas in formatted_application is also set to max_instances. 

But how do you make sure when setup_kubernetes_job is called, this won't get replaced?

EmanekaT: For example, here, https://github.com/Yelp/paasta/blob/df9c867fe1bcdf714ba85da84a8f2bb90a0105b6/paasta_tools/kubernetes_tools.py#L389 min_replicas is still min_replicas when run by setup_kubernetes_job

EmanekaT: Perhaps we should not update min_replicas and max_replicas in HPA when they are the same. And not allow users to set them as the same.

EvanKrall: Oh, good point, I hadn't thought about that.

paasta_tools/long_running_service_tools.py
EvanKrall: @enochtangg k I moved this stuff in d2e209dd6

paasta_tools/autoscaling/max_all_k8s_services.py
EvanKrall: @EmanekaT in fef79425 I set replicas on the formatted k8s object instead of changing min_instances. I think this should fix this problem.

PR #2718, opened 2020-03-12
Avoid connecting to mesos leader during paasta validate

jangsutsr:
This PR also enforces root access in driver container when
spark_cluster_manager == kubernetes. This is a temporary solution to
circumvent permission issue of reading k8s cluster SSL certificates of
host machines.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -191,6 +191,7 @@ class TronActionConfig(InstanceConfig):
         config_dict,
         branch_dict,
         soa_dir=DEFAULT_SOA_DIR,
+        for_validation=False,
     ):
         super().__init__(
             cluster=cluster,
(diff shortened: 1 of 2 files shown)


paasta_tools/tron_tools.py
```diff
@@ -204,13 +205,19 @@ def __init__(
         self.spark_ui_port = pick_random_port(
```
solarkennedy: can you also not do this under validation?

solarkennedy:
It is too bad that we make setup_tron_namespace do this at all? Can we not just pass the dns entry into spark and let it discover the leader "just in time" ?

jangsutsr:
@solarkennedy `find_mesos_leader` has been the default behavior so I think there's a reason behind? Also considering the scope of this PR, even using general dns entry should still be avoided in `paasta validate` anyways.

qui:
> @solarkennedy `find_mesos_leader` has been the default behavior so I think there's a reason behind? Also considering the scope of this PR, even using general dns entry should still be avoided in `paasta validate` anyways.

I think @solarkennedy is suggesting you just put the dns entry in here: https://github.com/Yelp/service_configuration_lib/blob/master/service_configuration_lib/spark_config.py#L87

Then `get_spark_config_dict` can just pass the dns entry string and not have to make any requests.

Worth trying to see if it works, I think. It seems cleaner - otherwise, we have to update the tronfig every time there's a leader election. I think setup_tron_namespace will handle this correctly, but still.

stug:
> It is too bad that we make setup_tron_namespace do this at all? Can we not just pass the dns entry into spark and let it discover the leader "just in time" ?

Another thing I was just saying on slack is that this means that if the master changes after we setup the job here, jobs will fail, which is not ideal.

paasta_tools/tron_tools.py
```diff
@@ -423,6 +432,11 @@ def format_docker_parameters(
         parameters = super().format_docker_parameters(with_labels=with_labels)
         if self.get_executor() == "spark":
             parameters.append({"key": "net", "value": "host"})
+            if self.get_spark_cluster_manager() == "kubernetes":
+                # TODO: this is a temprary solution to circumvent permission issue of reading
+                # SSL ceritficates of k8s clusters. We should remove this clause once a stable
+                # solution is available.
+                parameters.append({"key": "user", "value": "root"})
```
stug: is there a ticket you could link for this?

PR #2721, opened 2020-03-13
paasta validate for specific instance type

chlgit:
I have tested this against local yelpsoa-configs repo. And the pre-receive hook just calls "paasta validate".

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -354,25 +354,31 @@ def validate_unique_instance_names(service_path):
     check_passed = True
 
     for cluster in list_clusters(service, soa_dir):
-        service_instances = get_service_instance_list(
-            service=service, cluster=cluster, soa_dir=soa_dir
-        )
-        instance_names = [service_instance[1] for service_instance in service_instances]
-        instance_name_to_count = Counter(instance_names)
-        duplicate_instance_names = [
-            instance_name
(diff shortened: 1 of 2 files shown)


solarkennedy:
marathon/k8s are the majority of our instances.

How about just comment out the duplicate check while we are in this transition?

Certainly duplicate instance names are not "normal", but this is not a normal situation.

Another alternative is just to use a `f/` branch to bypass the validation entirely? 

The real question is: are there any other bad things that happen if we push duplicate instance names? Does anything else crash or fail?

chlgit:
Ah.....never know we have a f-branch for yelpsoa-configs before. I think I can try that at kubestage to see if anything bad happened.

qui:
As written, this basically removes the duplicate name check, right? Before, it was pulling all instances, regardless of instance type, and checking among them if there were any duplicates. So it would catch a marathon one that has the same name as an adhoc one, for example. Now, it's getting instance types one at a time - e.g. checking for duplicates within marathon instances. That is already essentially enforced by YAML, since it's a dictionary. 

Here's some context on why we added the check: https://jira.yelpcorp.com/browse/PAASTA-15470.

I think trying f/ branches is a good idea for this.

If you want, might be worth revisiting if we need the check anymore, but please don't ship this change - either remove entirely or leave it as is.

chlgit:
I have decided to use f-branch approach.

PR #2722, opened 2020-03-16
Implement k8s bespoke policies for paasta API

enochtangg:
The following code enables bespoke service autoscaling via `paasta autoscale` for k8s. The paasta api determines if the service's `instance_type` is marathon or kubernetes. Then, it routes to its respective ServiceConfig's `set_autoscaled_instances()` method which update its deployment replicas count.

Testing Done:
* 3 new unit tests
* make test
* manually tested with `paasta-api`

Note that this PR should not be released before https://github.com/Yelp/paasta/pull/2709.

### DIFF
--- a/paasta_tools/api/views/autoscaler.py
+++ b/paasta_tools/api/views/autoscaler.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 # Copyright 2015-2016 Yelp Inc.
 #
-#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -84,8 +83,6 @@ def get_autoscaler_count(request):
 
     instance_type = get_instance_type(service, instance, cluster, soa_dir)
     service_config = get_service_config(instance_type, service, instance, cluster, soa_dir)
-    print(service_config)
-    print(service_config.get_sanitised_deployment_name())
 
     response_body = {
         "desired_instances": service_config.get_instances(),
@@ -131,8 +128,6 @@ def update_autoscaler_count(request):
         )
 
     if instance_type == 'marathon':
-        # Dump whatever number from the client to zk. get_instances() will limit
-        # readings from zk to [min_instances, max_instances].
         set_instances_for_marathon_service(service=service, instance=instance, instance_count=desired_instances)
     elif instance_type == 'kubernetes':
         kube_client = KubeClient()
--- a/paasta_tools/api/views/autoscaler.py
+++ b/paasta_tools/api/views/autoscaler.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 # Copyright 2015-2016 Yelp Inc.
 #
+#
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -15,16 +16,20 @@
 """
 PaaSTA service list (instances) etc.
 """
+from kubernetes.client import V1Deployment
+from kubernetes.client import V1DeploymentSpec
+from kubernetes.client import V1LabelSelector
+
 from pyramid.response import Response
 from pyramid.view import view_config
 
 from paasta_tools.api import settings
 from paasta_tools.api.views.exception import ApiFailure
+from paasta_tools.kubernetes_tools import KubeClient
+from paasta_tools.kubernetes_tools import load_kubernetes_service_config
+from paasta_tools.kubernetes_tools import set_instances_for_kubernetes_service
 from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.marathon_tools import set_instances_for_marathon_service
-from paasta_tools.kubernetes_tools import load_kubernetes_service_config
-from paasta_tools.long_running_service_tools import set_instances_for_marathon_service
-from paasta_tools.marathon_tools import load_marathon_service_config
 from paasta_tools.utils import NoConfigurationForServiceError
 from paasta_tools.utils import validate_service_instance
 
@@ -50,7 +55,7 @@ def get_service_config(instance_type, service, instance, cluster, soa_dir):
                 instance=instance,
                 cluster=cluster,
                 soa_dir=soa_dir,
-                load_deployments=False,
+                load_deployments=True,
             )
         elif instance_type == 'kubernetes':
             service_config = load_kubernetes_service_config(
@@ -58,7 +63,7 @@ def get_service_config(instance_type, service, instance, cluster, soa_dir):
                 instance=instance,
                 cluster=cluster,
                 soa_dir=soa_dir,
-                load_deployments=False,
+                load_deployments=True,
             )
         else:
             error_message = f"Autoscaling is not supported for {service}.{instance} because instance type is neither " \
@@ -79,6 +84,8 @@ def get_autoscaler_count(request):
 
     instance_type = get_instance_type(service, instance, cluster, soa_dir)
     service_config = get_service_config(instance_type, service, instance, cluster, soa_dir)
+    print(service_config)
+    print(service_config.get_sanitised_deployment_name())
 
     response_body = {
         "desired_instances": service_config.get_instances(),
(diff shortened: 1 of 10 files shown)


paasta_tools/api/views/autoscaler.py
```diff
@@ -20,8 +20,51 @@
...
+                % ".".join([settings.cluster, instance])
+        )
+        raise ApiFailure(error_message, 404)
+    except Exception as e:
+        raise ApiFailure(e, 500)
+
+
+def get_service_config(instance_type, service, instance, cluster, soa_dir):
+    try:
+        if instance_type == 'marathon':
```
EmanekaT: Nit: use a map like https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes_tools.py#L140 
Map "kubernetes"/"marathon" to functions.

paasta_tools/api/views/autoscaler.py
```diff
@@ -30,17 +73,9 @@ def get_autoscaler_count(request):
...
-        )
-    except Exception:
-        error_message = f"Unable to load service config for {service}.{instance}"
-        raise ApiFailure(error_message, 404)
+
+    instance_type = get_instance_type(service, instance, cluster, soa_dir)
+    service_config = get_service_config(instance_type, service, instance, cluster, soa_dir)
 
     response_body = {
         "desired_instances": service_config.get_instances(),
```
EmanekaT: Note that this implementation only supports deployment. Perhaps you can update [it](https://github.com/Yelp/paasta/blob/371da7c1c02bcdabb8317502cd2e9f9aa83b8c9d/paasta_tools/kubernetes_tools.py#L928) to support statefulset ? Or return error if this service uses statefulset. You can tell if it is a stateful set or not like [this](https://github.com/Yelp/paasta/blob/371da7c1c02bcdabb8317502cd2e9f9aa83b8c9d/paasta_tools/kubernetes_tools.py#L1025)

enochtangg: added to `set_instances_for_kubernetes_service()`.

paasta_tools/api/views/autoscaler.py
```diff
@@ -98,5 +116,12 @@ def update_autoscaler_count(request):
             "WARNING desired_instances is less than min_instances %d" % min_instances
         )
 
+    if instance_type in SERVICE_CONFIG_MAP:
+        service_config.set_autoscaled_instances(instance_count=desired_instances)
+    else:
+        error_message = f"Autoscaling is not supported for {service}.{instance} because instance type is neither " \
+                        f"marathon or kubernetes."
+        raise ApiFailure(error_message, 404)
```
kawaiwanyelp: nit: Maybe 501 instead of 404?

paasta_tools/kubernetes_tools.py
```diff
@@ -1599,6 +1603,22 @@ def create_pod_disruption_budget(
...
+    formatted_application = service_config.format_kubernetes_app()
+    formatted_application.spec.replicas = instance_count
+    if service_config.get_persistent_volumes():
+        kube_client.deployments.patch_namespaced_stateful_set_scale(name=name,
+                                                                    namespace="paasta",
+                                                                    body=formatted_application)
+    else:
+        kube_client.deployments.patch_namespaced_deployment_scale(name=name,
+                                                                  namespace="paasta",
+                                                                  body=formatted_application)
```
kawaiwanyelp: Not a request for a fix, just a comment.

Was initially confused since it looks like this wouldn't play well with `setup_kubernetes_job`, since they would bump into each other. But @EvanKrall's PR #2709 fixes the issue using `get_autoscaled_instances` so it's fine.

tests/test_kubernetes_tools.py
```diff
@@ -1902,6 +1903,16 @@ def test_update_deployment():
...
 
 
+@mock.patch("paasta_tools.kubernetes_tools.KubernetesDeploymentConfig", autospec=True)
+def test_set_instances_for_kubernetes_service(mock_kube_deploy_config):
+    replicas = 5
+    mock_client = mock.Mock()
+    mock_kube_deploy_config.get_sanitised_deployment_name.return_value = 'fake_deployment'
+    mock_kube_deploy_config.format_kubernetes_app.return_value = mock.Mock()
+    set_instances_for_kubernetes_service(mock_client, mock_kube_deploy_config, replicas)
+    assert mock_client.deployments.patch_namespaced_deployment_scale.call_count == 1
```
kawaiwanyelp: nit: You should also check the args for the `patch_namespace_deployment_scale` to make sure the proper replica value is passed in.

tests/test_kubernetes_tools.py
```diff
@@ -1902,6 +1903,16 @@ def test_update_deployment():
     )
 
 
+@mock.patch("paasta_tools.kubernetes_tools.KubernetesDeploymentConfig", autospec=True)
+def test_set_instances_for_kubernetes_service(mock_kube_deploy_config):
```
kawaiwanyelp: You should also have a case for stateful sets.

paasta_tools/api/views/autoscaler.py
EvanKrall: Ohh, good point, get_autoscaled_instances doesn't handle statefulset. Do you mind fixing this? I could also try to fix it tomorrow.

enochtangg: Ah is probably the reason why I had trouble testing on statefulsets, I can add this.

paasta_tools/api/views/autoscaler.py
```diff
@@ -20,8 +20,51 @@
...
+        error_message = (
+                "Deployment key %s not found. Try to execute the corresponding pipeline if it's a fresh instance"
+                % ".".join([settings.cluster, instance])
+        )
+        raise ApiFailure(error_message, 404)
+    except Exception as e:
+        raise ApiFailure(e, 500)
+
+
+def get_service_config(instance_type, service, instance, cluster, soa_dir):
```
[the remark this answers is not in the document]
enochtangg: @solarkennedy this function only calls either calls `load_marathon_service_config()` and `load_kubernetes_service_config` which I think are the functions you are referring to? I don't see code that takes in an instances_type and returns its respective service_config though.

paasta_tools/kubernetes_tools.py
```diff
@@ -944,7 +944,11 @@ def get_autoscaled_instances(self) -> int:
             return None
 
     def set_autoscaled_instances(self, instance_count: int) -> None:
-        raise NotImplementedError()
+        """Set the number of instances in the same way that the autoscaler does."""
+        kube_client = KubeClient()
```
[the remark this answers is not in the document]
enochtangg: not that I'm aware of. We don't actually use the kube_client until we set the replica number. The current design pattern in `kubernetes_tools.py` is to instantiate functions that talk to kube_client outside of the `KubernetesDeploymentConfig` class and to pass into kube_client as a parameter

paasta_tools/api/views/autoscaler.py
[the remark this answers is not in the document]
solarkennedy: We have a couple. `PaastaServiceConfigLoader` is one. `get_instance_config` is another. @EvanKrall might know another?

paasta_tools/kubernetes_tools.py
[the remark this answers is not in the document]
solarkennedy: That is too bad. It would be nice not to have to instantiate a kube client upon every API request.

What is `instance_set_state` doing with `kube_client=settings.kubernetes_client` ?

If this works, can you make the code re-use the existing kube_client that the api already has?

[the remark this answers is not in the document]
enochtangg: @solarkennedy  Makes sense. This would be tricky because `set_autoscaled_instances` is an overridden method from `LongRunningServiceConfig()` which requires the same signature. I'm unsure if we want to decouple this? Also, `get_autoscaled_instances` uses a new instantiated kube client.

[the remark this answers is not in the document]
solarkennedy: Ok. Well I recommend functions not instantiate their own clients and instead get one passed in if available.

I think it is fine for each InstanceConfig to have their own implementation of set_autoscaled_instances. To me that is the whole point of having a class abstraction like this.

paasta_tools/api/views/autoscaler.py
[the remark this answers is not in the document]
qui: Please use this one: https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L2807
load_marathon_service_config and load_kubernetes_service_config call that. There should be a single function we use to load configs, so we can add things like autoconfig and every user of instance config gets that in the same way. 

We are planning to deprecate PaastaServiceConfigLoader: https://jira.yelpcorp.com/browse/PAASTA-16358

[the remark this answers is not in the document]
qui: Oh, I see you calling load_marathon_service_config above. I guess it's up to you if you need the extra functionality in those methods or not, sorry I missed that.

PR #2724, opened 2020-03-17
Made paasta_broadcast_log work with k8s and mesos

solarkennedy:
This script is useful to emit things like "This server going down for maintenance" .

This PR adds support to Tron tasks as well as k8s ones.

### DIFF
--- a/paasta_tools/broadcast_log_to_services.py
+++ b/paasta_tools/broadcast_log_to_services.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python
+# Copyright 2015-2016 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import sys
+
+from paasta_tools.kubernetes_tools import get_all_kubernetes_services_running_here
+from paasta_tools.marathon_tools import marathon_services_running_here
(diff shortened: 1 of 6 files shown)


paasta_tools/broadcast_log_to_services.py
```diff
@@ -0,0 +1,60 @@
...
+        tron_services = tron_jobs_running_here()
+    except MesosSlaveConnectionError:
+        marathon_services = []
+        tron_services = []
+    kubernetes_services = get_all_kubernetes_services_running_here()
+    return marathon_services + tron_services + kubernetes_services
+
+
+def main() -> None:
+    broadcast_log_all_services_running_here(sys.stdin.read().strip())
```
EvanKrall: I'm surprised this works well enough -- this will log one entry per service for the whole execution, regardless of how many lines of input.

paasta_tools/broadcast_log_to_services.py
```diff
@@ -0,0 +1,60 @@
...
+        )
+
+
+def get_all_services_running_here(cluster, soa_dir):
+    try:
+        marathon_services = marathon_services_running_here()
+        tron_services = tron_jobs_running_here()
+    except MesosSlaveConnectionError:
+        marathon_services = []
+        tron_services = []
```
EvanKrall: is it possible for `marathon_services_running_here` to succeed and `tron_jobs_running_here` to fail, or vice-versa? 

maybe split each of them into a separate try/catch.

paasta_tools/broadcast_log_to_services.py
```diff
@@ -0,0 +1,60 @@
...
+    kubernetes_services = get_all_kubernetes_services_running_here()
```
EvanKrall: is there any way this call can fail? may be worth throwing a try/catch around this too.

paasta_tools/broadcast_log_to_services.py
solarkennedy: good idea

paasta_tools/broadcast_log_to_services.py
solarkennedy: correct. the input line here is like "shutting down now". I don't know why it was designed to take stdin instead of jus ta cli arg

PR #2726, opened 2020-03-18
Bump alert_after from 2m to 5m for check_kubernetes_services_replication 

EmanekaT:
The default is [2m](https://github.com/Yelp/paasta/blob/master/paasta_tools/monitoring_tools.py#L277)

### DIFF
--- a/paasta_tools/check_kubernetes_services_replication.py
+++ b/paasta_tools/check_kubernetes_services_replication.py
@@ -74,7 +74,7 @@ def check_kubernetes_pod_replication(
     instance_config: KubernetesDeploymentConfig,
     all_tasks_or_pods: Sequence[V1Pod],
     smartstack_replication_checker: KubeSmartstackReplicationChecker,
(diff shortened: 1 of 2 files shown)


paasta_tools/check_kubernetes_services_replication.py
```diff
@@ -87,6 +89,13 @@ def check_kubernetes_pod_replication(
     proxy_port = get_proxy_port_for_instance(instance_config)
 
     registrations = instance_config.get_registrations()
+    if "monitoring" not in instance_config.config_dict:
+        instance_config.config_dict["monitoring"] = {}
+    instance_config.config_dict["monitoring"][
+        "alert_after"
+    ] = instance_config.config_dict["monitoring"].get(
+        "alert_after", default_alert_after
+    )
```
chlgit: nit: could we have a function to setup the default monitoring config? I think it would be easier to read.

PR #2728, opened 2020-03-18
Expose k8s node selectors and node affinities for Kubernetes apps

kawaiwanyelp:
### Description
In an effort to give devs more control over how their nodes are scheduled, as well as offer a limited equivalent for (Marathon) constraints to k8s tasks, I've exposed node selectors and affinities under a `node_selectors` config value in soa-configs. 

This is not intended to provide a functional replacement to Marathon constraints for  k8s apps. The Kubernetes scheduler already tries to spread pods out, much better than Marathon, so we shouldn't need as strong controls. However, if a dev wants to, say, only run their instances on specific instance types, then it is possible with this feature.

I've added comments and updated the soa-configs documentation to provide an explanation on how it's supposed to work.

### Testing
- `make test
- Manual testing on a test cluster
  - Status quo, no `node_selectors`
  - `node_selectors` test cases with string value (becomes a node selector), list of strings value (becomes an "in" node affinity), and each operator config value (becomes a node affinity)

### Notes
The k8s schema allows the `constraints` and `extra_constraints` keys, likely because it was initially copied from the Marathon one. I *have not* deprecated these because some existing configs for k8s service instances do set them, again likely because they were converted from Marathon ones, even though they aren't supported.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -173,16 +173,16 @@ These options are only applicable to tasks scheduled on Kubernetes.
             - operator: NotIn
               values: ["c3.8xlarge"]
 
-    * Requires that a node have the ``vip`` label::
+    * Requires that a node have the ``ssd`` label::
 
         node_selectors:
-          vip:
+          ssd:
             - operator: Exists
 
-    * Requires that a node not have the ``vip`` label::
+    * Requires that a node not have the ``ssd`` label::
 
         node_selectors:
-          vip:
+          ssd:
             - operator: DoesNotExist
 
     * Requires that a node have a label ``priority`` with a value greater than 1
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -170,35 +170,30 @@ These options are only applicable to tasks scheduled on Kubernetes.
 
         node_selectors:
           instance_type:
-            operator: NotIn
-            values: ["c3.8xlarge"]
+            - operator: NotIn
+              values: ["c3.8xlarge"]
 
     * Requires that a node have the ``vip`` label::
 
         node_selectors:
           vip:
-            operator: Exists
+            - operator: Exists
 
     * Requires that a node not have the ``vip`` label::
 
         node_selectors:
           vip:
-            operator: DoesNotExist
+            - operator: DoesNotExist
 
     * Requires that a node have a label ``priority`` with a value greater than 1
       and less than 5::
 
         node_selectors:
           priority:
-            operator: Gt
-            value: 3
-
-    * Requires that a node have a priority less than 3::
-
-        node_selectors:
-          priority:
-            operator: Lt
-            value: 3
+            - operator: Gt
+              value: 1
+            - operator: Lt
+              value: 5
 
     .. note::
 
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -106,11 +106,19 @@ specify the following options:
     validate that the bind mounts are "safe".
 
 
-``Placement Options (Constraints)``
------------------------------------
+Placement Options
+-----------------
 
-Constraint options control how Mesos schedules a task, whether it is scheduled by
-Marathon, Tron, or ``paasta remote-run``.
+Placement options provide control over how PaaSTA schedules a task, whether it
+is scheduled by Marathon (on Mesos), Kubernetes, Tron, or ``paasta remote-run``.
+Most commonly, it is used to restrict tasks to specific locations.
+
+.. _general-placement-options:
+
+General
+^^^^^^^
+
+These options are applicable to tasks scheduled through Mesos or Kubernetes.
 
   * ``deploy_blacklist``: A list of lists indicating a set of locations to *not* deploy to. For example:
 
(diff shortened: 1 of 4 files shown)


docs/source/yelpsoa_configs.rst
```diff
@@ -127,6 +133,78 @@ Marathon, Tron, or ``paasta remote-run``.
...
+            - operator: DoesNotExist
+
+    * Requires that a node have a priority greater than 1 and less than 5::
+
+        node_selectors:
+          priority:
+            - operator: Gt
+              value: 1
+            - operator: Lt
+              value: 5
```
chlgit: I am not sure I understand this part. How do we assign the priority to nodes? And what's the scenario you expect users would need this?

paasta_tools/kubernetes_tools.py
```diff
@@ -2049,3 +2099,25 @@ def get_pod_hostname(kube_client: KubeClient, pod: V1Pod) -> str:
...
+def to_node_label(label: str) -> str:
+    """k8s-ifies certain special node labels"""
+    if label in {"instance_type", "instance-type"}:
+        # kube_client is assumed to already be configured
+        version_info = kube_client.VersionApi().get_code()
+        if int(version_info.major) == 1 and int(version_info.minor) < 17:
+            # the beta instance-type label is deprecated as of k8s v1.17
+            return "beta.kubernetes.io/instance-type"
+        else:
+            return "node.kubernetes.io/instance-type"
```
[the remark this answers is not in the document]
kawaiwanyelp: Unfortunately no. If you check our nodes, they all have the `beta` version of the label set.

docs/source/yelpsoa_configs.rst
kawaiwanyelp: I'm just using priority as an example of a label that could have a numeric value. I don't envision any particular scenario where we will end up using the Gt or Lt operators, but they do exist in k8s, so I just added them. We can remove them if we don't want to support them.

chlgit: I am not sue if this would confuse users, but I think we can remove it anytime if we want.

paasta_tools/kubernetes_tools.py
```diff
@@ -2049,3 +2099,25 @@ def get_pod_hostname(kube_client: KubeClient, pod: V1Pod) -> str:
         return pod.spec.node_name
     # if label has disappeared (say we changed it), default to node name
     return node.metadata.labels.get("yelp.com/hostname", pod.spec.node_name)
+
+
+def to_node_label(label: str) -> str:
```
solarkennedy: @EvanKrall can you weigh in on this function? Unlike before with mesos we just injected constraints kinda "as is".
With this function we are adding a bit of abstraction but converting things for users.

paasta_tools/kubernetes_tools.py
```diff
@@ -2049,3 +2099,25 @@ def get_pod_hostname(kube_client: KubeClient, pod: V1Pod) -> str:
...
+    """k8s-ifies certain special node labels"""
+    if label in {"instance_type", "instance-type"}:
+        # kube_client is assumed to already be configured
+        version_info = kube_client.VersionApi().get_code()
```
solarkennedy: Where did `kube_client` come from? 
It looks like this line of code actually makes an api call and stuff. 

It is really too bad this function needs to do that?

paasta_tools/kubernetes_tools.py
```diff
@@ -2049,3 +2099,25 @@ def get_pod_hostname(kube_client: KubeClient, pod: V1Pod) -> str:
...
+        if int(version_info.major) == 1 and int(version_info.minor) < 17:
+            # the beta instance-type label is deprecated as of k8s v1.17
+            return "beta.kubernetes.io/instance-type"
+        else:
+            return "node.kubernetes.io/instance-type"
+    elif label in {
+        "datacenter",
```
qui: Are these also all configurable via deploy_whitelist/blacklist?

I also wonder if this function is necessary, especially if we need to make an api call to get the kubernetes version. What if we replaced all the old beta selectors with the new name when we do an upgrade?

paasta_tools/kubernetes_tools.py
kawaiwanyelp: It's an alias: `from kubernetes import client as kube_client`. We use it in the `KubeClient` class, instances of which we also confusingly name `kube_client`.

An API call is indeed unfortunate but since the instance_type label differs by version, I can't really see a better choice to provide some level of compatibility protection. We could just drop this and make all services specify the full label, but then we'd have to change them if we ever upgrade. We could also preemptively upgrade k8s to v1.17, but the Python k8s client only supports up to v1.15.

docs/source/yelpsoa_configs.rst
```diff
@@ -106,11 +106,17 @@ specify the following options:
...
-``Placement Options (Constraints)``
------------------------------------
+Placement Options
+-----------------
 
-Constraint options control how Mesos schedules a task, whether it is scheduled by
-Marathon, Tron, or ``paasta remote-run``.
+Placement options provide control over how PaaSTA schedules a task, whether it
+is scheduled by Marthon (on Mesos), Kubernetes, Tron, or ``paasta remote-run``.
+Most commonly, it is used to set topological restrictions on tasks.
```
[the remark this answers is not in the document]
kawaiwanyelp: "Topology" means how nodes are arranged, typically by (hiera) location. I can use location as a more intuitive, non-infra term.

paasta_tools/kubernetes_tools.py
kawaiwanyelp: I wanted to create a single function to serve as a source of truth for any sort of label name changing, so it includes the instance-type label fudging, as well as the location label stuff too. deploy_whitelist/blacklist stuff use this function to fudge the location labels.

deploy_whitelist/blacklist in k8s are just special cases of node_selectors, specially "in" and "notin". It's all based on labels anyway.

For the replacement idea, do you mean we just set the beta selector here and just change it when we upgrade? I'm not sure I like that since there is a period of incompatibility between when we upgrade k8s and when we upgrade everything else to match.

docs/source/yelpsoa_configs.rst
```diff
@@ -127,6 +135,83 @@ Marathon, Tron, or ``paasta remote-run``.
...
+    * Requires that a node have the ``vip`` label::
+
+        node_selectors:
+          vip:
+            - operator: Exists
+
+    * Requires that a node not have the ``vip`` label::
+
+        node_selectors:
+          vip:
```
matfra: These are great examples, thank you for implementing this feature and for the detailled documentation :-)
As you know at Yelp we actually have a pool called vip, so may I suggest that for this specific example we use something different than 'vip', just to avoid any possible confusion (in case someone reads the docs)
A typical real world use case, and one of my dream, is to be able to specify that you want to run on a node that has fast nvme SSD storage. While we could use a list of instance types as constraint (what we currently do), I think this is a great example to use a label 'ssd' or 'fast_storage' . Can the doc use this as an example?

paasta_tools/kubernetes_tools.py
enochtangg: Adjusted, and removed API call to k8s.

enochtangg:
We've removed the extra API call to check k8s version before creating node selectors. Right now, we've upgrade to k8s 1.17.4 everywhere so using the label `node.kubernetes.io/instance-type` should be safe.

PR #2729, opened 2020-03-18
FLINK-2303 status color

sofyat:
Show job information for all jobs only in verbose mode, otherwise show only the first few jobs. Use colors to highlight job status.


solarkennedy:
Neat! I'm a fan of color in general.
Can you run `make install-hooks` to get the hooks installed so `black` runs? (why the tests are failing)

sofyat:
It should be fixed now

PR #2730, opened 2020-03-18
U/tingyenl/coreml 1401 get a meaningful spark app name

lydian:
tested locally to make sure the same error is not happening again, and also update the tests and configure the input type to ensure we capture the error.


solarkennedy:
@lydian can you try rebasing on master?

PR #2731, opened 2020-03-18
pause HPA for kubernetes services using zk

enochtangg:
The following code is used to pause HPA when a user has specified `paasta pause_service_autoscaler`. Currently, if a user runs pause autoscaler, the paasta api will write to zk `time_until` in minutes and `resumed` which is a bool. 

On a cron job, `setup_kubernetes_job` will update the k8s deployment which updates HPA and makes min_instanes = current replicas. This restricts the service from scaling down, but still allows it to scale up. Note that the `update` is only ran on two conditions: if service is paused OR if service is not paused and has not been resumed yet. 

Testing Done:
* make test
* added units tests

### DIFF
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -12,6 +12,7 @@ from kubernetes.client import V1Deployment
 from kubernetes.client import V1StatefulSet
 from kubernetes.client.rest import ApiException
 
+from paasta_tools.autoscaling.autoscaling_service_lib import autoscaling_is_paused
 from paasta_tools.kubernetes_tools import create_deployment
 from paasta_tools.kubernetes_tools import create_pod_disruption_budget
 from paasta_tools.kubernetes_tools import create_stateful_set
@@ -279,12 +280,17 @@ class DeploymentWrapper(Application):
             f"Syncing HPA setting for {self.item.metadata.name}/name in {self.item.metadata.namespace}"
         )
         hpa_exists = self.exists_hpa(kube_client)
-        # NO autoscaling
-        if not self.should_have_hpa():
-            # Remove HPA if autoscaling is disabled
-            if hpa_exists:
+        if hpa_exists:
+            if not self.should_have_hpa():
+                # Remove HPA if autoscaling is disabled
                 self.delete_horizontal_pod_autoscaler(kube_client)
-            return
+                return
+            elif autoscaling_is_paused():
+                self.logging.info(
+                    f"Autoscaler is paused. Setting min instances to {self.item.spec.replicas}."
+                    f"HPA will not scale down service."
+                )
+                self.soa_config.set_min_instances(self.item.spec.replicas)
 
         body = self.soa_config.get_autoscaling_metric_spec(
             name=self.item.metadata.name,
@@ -292,10 +298,11 @@ class DeploymentWrapper(Application):
             namespace=self.item.metadata.namespace,
         )
         if not body:
-            raise Exception(
+            self.logging.info(
                 f"CRIT: autoscaling misconfigured for {self.kube_deployment.service}."
-                + f"{self.kube_deployment.instance}.Please correct the configuration and update pre-commit hook."
+                f"{self.kube_deployment.instance}.Please correct the configuration and update pre-commit hook."
             )
+            return
         self.logging.debug(body)
         if not hpa_exists:
             self.logging.info(
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -13,6 +13,8 @@ from kubernetes.client import V1StatefulSet
 from kubernetes.client.rest import ApiException
 
 from paasta_tools.autoscaling.autoscaling_service_lib import autoscaling_is_paused
+from paasta_tools.autoscaling.autoscaling_service_lib import is_autoscaling_resumed
+from paasta_tools.autoscaling.autoscaling_service_lib import write_autoscaling_paused
 from paasta_tools.kubernetes_tools import create_deployment
 from paasta_tools.kubernetes_tools import create_pod_disruption_budget
 from paasta_tools.kubernetes_tools import create_stateful_set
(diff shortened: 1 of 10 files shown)


paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -281,7 +282,7 @@ def sync_horizontal_pod_autoscaler(self, kube_client: KubeClient) -> None:
         )
         hpa_exists = self.exists_hpa(kube_client)
         # NO autoscaling
-        if not self.should_have_hpa():
+        if not self.should_have_hpa() or autoscaling_is_paused():
```
kawaiwanyelp: nit: Now that there are multiple reasons for deleting the HPA, I think its worth adding a log line here just to explain why the HPA is gone.

tests/kubernetes/application/test_controller_wrapper.py
```diff
@@ -167,74 +167,95 @@ def test_sync_horizontal_pod_autoscaler_no_autoscaling():
...
-    assert (
-        mock_client.autoscaling.replace_namespaced_horizontal_pod_autoscaler.call_count
-        == 0
-    )
-    assert app.delete_horizontal_pod_autoscaler.call_count == 0
+    with mock.patch(
+        "paasta_tools.kubernetes.application.controller_wrappers.autoscaling_is_paused",
+        autospec=True,
+    ) as mock_autoscaling_is_paused:
+        mock_autoscaling_is_paused.return_value = True
```
kawaiwanyelp: Should probably be `False` (not paused), so the test will only delete the HPA if there is actually no autoscaling configured.

also, nit: Maybe patch as a decorator instead of a context manager.

tests/kubernetes/application/test_controller_wrapper.py
```diff
@@ -167,74 +167,95 @@ def test_sync_horizontal_pod_autoscaler_no_autoscaling():
     # Do nothing
     config_dict = {"instances": 1}
     app = setup_app(config_dict, False)
-    app.sync_horizontal_pod_autoscaler(kube_client=mock_client)
```
kawaiwanyelp: You should write tests for creating and deleting the HPA based solely on whether or not the autoscaler is paused.

(This comment isn't related to this line, I just needed it to appear as a standalone comment)

EmanekaT:
I think this might not work. `setup_kubernetes_job` does not  run either create() or update() commands until the user changes yelpsoa_configs or updates git master.

Some suggestions:
* Instead of using ZK, what about using annotations? You can add an annotation to the paused deployements and refector setup_kubernetes_jobs to check  those annotations. Then setup_kubernetes_jobs will run update() for those services with annotations. 
* You can either delete HPA or change min_replicas to current_replicas. HPA can still scale up in this way.

enochtangg:
@EmanekaT I'm tempted to just change `setup_kubernetes_job` so that it simply run's `update()` when it sees that the autoscaler has been paused on zk. The reason or this is because:
1. Adding annotations requires adding labels to each individual deployments/pods in an entire cluster, then removing them after pausing has stopped. Rather than just a single entry on zk. 
2. Some clusters have a mix of marathon and k8s services which requires storing the pause time in both zk and annotation labels on deployments.

Realistly, we have three options here:
- Only use ZK.
- Use both ZK and annotations.
- Only use annotations. But deprecate `paasta pause_service_autoscale`  for marathon services.

Personally, I feel like doing the first is the most easiest and reasonable fix at the moment. But eventually, when we fully migrate to k8s, we can look at switching from zk to annotations. What are your thoughts?

EmanekaT:
> @EmanekaT I'm tempted to just change `setup_kubernetes_job` so that it simply run's `update()` when it sees that the autoscaler has been paused on zk. The reason or this is because:
> 
> 1. Adding annotations requires adding labels to each individual deployments/pods in an entire cluster, then removing them after pausing has stopped. Rather than just a single entry on zk.
> 2. Some clusters have a mix of marathon and k8s services which requires storing the pause time in both zk and annotation labels on deployments.
> 
> Realistly, we have three options here:
> 
> * Only use ZK.
> * Use both ZK and annotations.
> * Only use annotations. But deprecate `paasta pause_service_autoscale`  for marathon services.
> 
> Personally, I feel like doing the first is the most easiest and reasonable fix at the moment. But eventually, when we fully migrate to k8s, we can look at switching from zk to annotations. What are your thoughts?

Makes sense. Go for it :)

paasta_tools/setup_kubernetes_job.py
```diff
@@ -140,11 +144,21 @@ def setup_kube_deployments(
...
+            app.update(kube_client)
+        elif not autoscaling_is_paused() and not is_autoscaling_resumed():
+            resumed_autoscaler = True
             app.update(kube_client)
         else:
             log.debug(f"{app} is up to date, no action taken")
 
+    if resumed_autoscaler:
+        write_autoscaling_resumed()
+
```
EvanKrall, line 161: This elif is the reason for the existence of `is_autoscaling_resumed`, right? We're trying to avoid updating all k8s deployments whenever the autoscaler is paused?

I worry about the resumed key and the paused key getting out of sync somehow. Is there a way to tell the last time the k8s deployment was updated, and compare that to the pause expiration timestamp?

I also worry about `setup_kubernetes_job` getting called with some subset of services. (At the moment, we call SKJ with all service instances with `paasta_list_kubernetes_service_instances | shuf | xargs setup_kubernetes_job`, but for Marathon before we had deployd we parallelized `setup_marathon_job` with `xargs --max-procs`. I could imagine us doing this if SKJ starts taking too long to run.)

If called with a subset of services, we'd still update the resumed key in ZK and then any services that we excluded would never get their HPA reset.

I think comparing to the pause expiration timestamp would avoid this issue too.

enochtangg: @EvanKrall Hmm good point. One way we could fix this is adding an annotation to all deployments that have been paused. I think this might be better than checking the last time the k8s deployment was updated, because deployments could possibly be updated for multiple reasons outside of being paused.

Also by adding annotations to each in individual deployment, we could still support calling `setup_kubernetes_job` for a subset of services.

paasta_tools/setup_kubernetes_job.py
```diff
@@ -140,7 +142,15 @@ def setup_kube_deployments(
             not in existing_apps
         ):
             app.create(kube_client)
-        elif app and app.kube_deployment not in existing_kube_deployments:
+        elif (
+            app
+            and (app.kube_deployment not in existing_kube_deployments)
+            or autoscaling_is_paused()
+        ):
+            app.update(kube_client)
```
EvanKrall, line 150: I think you could potentially avoid some `app.update`s by using `elif autoscaling_is_paused() and is_autoscaling_resumed(app.item)` but I'm not sure.

enochtangg: Yeah I think we can put them all in one line, but I'm afraid that is confusing to read. Also, I've condense this further by wrapping the whole thing with `if app` as opposed to writing it on each line.

PR #2733, opened 2020-03-19
Add option in list_kubernetes_service_instances to show sanitised names

kawaiwanyelp:
### Description
list_kubernetes_service_instances currently shows service and instance names composed with a dot. I'm adding an option to show k8s sanitised names instead, i.e. the names deployments and statefulsets use when created by PaaSTA. I'm doing this so i can grep for the name of the paasta-contract-monitor's deployment from the source of truth (PaaSTA), and delete the deployment periodically so the PCM actually properly tests scheduling.

### Testing
`make test`
manual testing on a test cluster

### Notes
@drmorr0, I noticed that the Pd Disruption Budgets created by PaaSTA do not use any sanitised names. However, in the interest of not deleting the budgets that we have now, I'm not going to properly sanitise them in this review. I've added a TODO.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -995,10 +995,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
 
     def get_kubernetes_metadata(self, git_sha: str) -> V1ObjectMeta:
         return V1ObjectMeta(
-            name="{service}-{instance}".format(
-                service=self.get_sanitised_service_name(),
-                instance=self.get_sanitised_instance_name(),
-            ),
+            name=self.get_sanitised_deployment_name(),
             labels={
                 "yelp.com/paasta_service": self.get_service(),
                 "yelp.com/paasta_instance": self.get_instance(),
@@ -1010,10 +1007,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         )
 
     def get_sanitised_deployment_name(self) -> str:
-        return "{service}-{instance}".format(
-            service=self.get_sanitised_service_name(),
-            instance=self.get_sanitised_instance_name(),
-        )
+        return get_kubernetes_app_name(self.get_service(), self.get_instance())
 
     def get_min_task_uptime(self) -> int:
         return self.config_dict.get("bounce_health_params", {}).get(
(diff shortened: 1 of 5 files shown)


paasta_tools/list_kubernetes_service_instances.py
```diff
@@ -55,6 +56,15 @@ def parse_args():
         default=DEFAULT_SOA_DIR,
         help="define a different soa config directory",
     )
+    parser.add_argument(
+        "-s",
```
chlgit: Would that conflict with -s --service?

paasta_tools/kubernetes_tools.py
```diff
@@ -1578,6 +1569,7 @@ def pod_disruption_budget_for_service_instance(
     service: str, instance: str, max_unavailable: Union[str, int],
 ) -> V1beta1PodDisruptionBudget:
     return V1beta1PodDisruptionBudget(
+        # TODO: the PDB name needs to be sanitised, same as the app names.
```
chlgit: This is a potential DAR in my opinion. I would prefer we fix it as early as possible that we only have one Ocean now.

paasta_tools/list_kubernetes_service_instances.py
kawaiwanyelp: It wouldn't since -s is not used here. However, it might cause some confusion so I'll remove the short option.

paasta_tools/kubernetes_tools.py
kawaiwanyelp: I don't think it's a potential incident, just a potential source of confusion. What matters for PDBs are the pod selectors (using labels), which are correctly configured.

tests/test_kubernetes_tools.py
```diff
@@ -2073,6 +2067,17 @@ def test_parse_container_resources():
     ) == KubeContainerResources(2, 100, None)
 
 
+@mock.patch.object(kubernetes_tools, "sanitise_kubernetes_name", autospec=True)
```
solarkennedy: does this have to be mocked?

kawaiwanyelp: Nope, but I want to assert that it was called.

paasta_tools/kubernetes_tools.py
chlgit: I mean, what if one day we forgot the name didn't be sanitized and use '_' in the name. Will that cause PDB can't be setup so the cluster can't be scaled down?

PR #2736, opened 2020-03-21
Add ability for services to opt out of inbound network traffic

y-trobinso:
Adds a new soa configuration variable, `inbound_network_mode` that accepts the following attributes:

* `full`: the default. No change versus existing behavior.
* `restrict`: adds a single iptables rule to `REJECT` all traffic for the service's `dport` at the end of the chain.

Advice welcome on testing this to verify the iptables chain works as expected. Unit tests otherwise added and pass as expected.

Context: PAASTA-16455

### DIFF
--- a/paasta_tools/cli/schemas/marathon_schema.json
+++ b/paasta_tools/cli/schemas/marathon_schema.json
@@ -351,6 +351,12 @@
                 "security": {
                     "type": "object",
                     "properties": {
+                        "inbound_network_mode": {
+                            "enum": [
+                                "full",
+                                "restrict"
+                            ]
+                        },
                         "outbound_firewall": {
                             "enum": [
                                 "block",
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -327,9 +327,9 @@ def _dns_servers():
                 yield parts[1]
 
 
-def ensure_shared_chains(conf):
+def ensure_shared_chains(conf=None):
     _ensure_dns_chain()
-    if conf.get_network_mode() != "local":
+    if conf is None or conf.get_network_mode() != "local":
         _ensure_internet_chain()
         _ensure_common_chain()
 
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -327,10 +327,11 @@ def _dns_servers():
                 yield parts[1]
 
 
-def ensure_shared_chains():
+def ensure_shared_chains(conf):
     _ensure_dns_chain()
-    _ensure_internet_chain()
-    _ensure_common_chain()
+    if conf.get_network_mode() != "local":
+        _ensure_internet_chain()
+        _ensure_common_chain()
 
 
 def _ensure_common_chain():
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -181,6 +181,7 @@ def _yocalhost_rule(port, comment, protocol="tcp"):
         target_parameters=(),
     )
 
+
 def _reject_remaining_inbound_traffic_rule(port, protocol="tcp"):
     """Return an iptables rule denying all other traffic.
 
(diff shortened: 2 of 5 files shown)


solarkennedy:
You will need to add this to the marathon schema json file to let it in.

Keep in mind this will not work with k8s, which we are migrating to at full steam.

Also this will break any healthchecks we do on normal http or tcp services from external sources (hacheck, other synapse haproxies, etc), right?

chriskuehl:
The code looks right to me but for iptables changes I have very little confidence in my own ability to correctly understand what a rule is actually doing. My reading is that this will work because our PAASTA chain is invoked from INPUT, but I think this deserves some manual testing before merging. Unfortunately the tests don't really tell us anything about how the iptables rules actually behave.

Also some additional questions:

* Is it possible to have Marathon configure containers it launches to bind to localhost only? That seems a lot cleaner to me than this approach where we're effectively simulating that with the firewall, although I am not sure if it will require lots of changes to health-checking or anything like that.
* Is this going to cause the smartstack-based health-checking to fail for any services using this? My understanding is that smartstack health-checking (separate from the marathon health-checking) works by looking at the synapse healthcheck status from an arbitrary host in the same cluster. If the smartstack check is failing, like I think this will, does it cause problems like stuck paasta bounces, failed sensu alerting, etc?
* Is this even worth doing given the move to k8s?

y-trobinso:
> You will need to add this to the marathon schema json file to let it in.
+1. I'll add this to the PR (and fix the integration, too).

> Keep in mind this will not work with k8s, which we are migrating to at full steam.
We're aware. This is expected to be a stopgap that we can deploy for a small number of services at this writing, so that shouldn't be an issue.

> Also this will break any healthchecks we do on normal http or tcp services from external sources (hacheck, other synapse haproxies, etc), right?

Yes, it should. We'll be testing this with one of our own services in SEC first to determine impact, and have no plans to roll this out to other services at this time.

y-trobinso:
> The code looks right to me but for iptables changes I have very little confidence in my own ability to correctly understand what a rule is actually doing. My reading is that this will work because our PAASTA chain is invoked from INPUT, but I think this deserves some manual testing before merging. Unfortunately the tests don't really tell us anything about how the iptables rules actually behave.

I'm not 100% sure how to manually test this, but I'll look into doing so on my branch if possible.


> Also some additional questions:
> 
> * Is it possible to have Marathon configure containers it launches to bind to localhost only? That seems a lot cleaner to me than this approach where we're effectively simulating that with the firewall, although I am not sure if it will require lots of changes to health-checking or anything like that.
> * Is this going to cause the smartstack-based health-checking to fail for any services using this? My understanding is that smartstack health-checking (separate from the marathon health-checking) works by looking at the synapse healthcheck status from an arbitrary host in the same cluster. If the smartstack check is failing, like I think this will, does it cause problems like stuck paasta bounces, failed sensu alerting, etc?
> * Is this even worth doing given the move to k8s?

I think this is still worth doing given the move to k8s, but I'll ask this during my 1:1 this week. It's also expected this will cause health checks to fail, and we'll be testing this in our own cluster to determine what that looks like.

solarkennedy:
You can test this manually if your venv is activated and you use `paasta local-run`. Use the -y switch to point to a local copy which has the option enabled to block.

danielpops:
> Also this will break any healthchecks we do on normal http or tcp services from external sources (hacheck, other synapse haproxies, etc), right?

Whoa, I thought these checks were all done locally, so I'm surprised to hear that! @solarkennedy is this a non-standard setup? is there a sourcegraph or otherwise means of identifying the extent of these external health checks?

solarkennedy:
> > Also this will break any healthchecks we do on normal http or tcp services from external sources (hacheck, other synapse haproxies, etc), right?
> 
> Whoa, I thought these checks were all done locally, so I'm surprised to hear that! @solarkennedy is this a non-standard setup? is there a sourcegraph or otherwise means of identifying the extent of these external health checks?

It is very typical at Yelp to have things like synapse and hacheck healthcheck a service from outside a container. I can't think of a sog query, probably your best bet is to tcpdump or tail an access log to get the "real" source of healthchecks from "in the wild".

y-trobinso:
We did some testing on Friday, re, tcpdump and reading the access logs from the service we want to protect with this feature. As far as I can tell, healthcheck traffic is seen by the host as originating from 127.0.0.1 and 169.254.1.1 (read: smartstack), and nothing else at this writing.

I'm comfortable shipping this as-is and iterating, as well as opening up the smartstack port in iptables if the rules I kept for them don't work. I'll run this by the team on Monday to verify if this works for us.

Thanks again for the fast CRs and the substantive feedback.

PR #2737, opened 2020-03-23
Make k8s_itests run in docker containers

EmanekaT:
I didn't really want to do this since it's slow. But there are more refactoring than I thought to run it outside a container. I give up. 
I'll add a local docker registry later if I can. It's hard as kind runs in docker network, so it cannot talk to host network.

### DIFF
--- a/yelp_package/dockerfiles/itest/k8s/wait_paasta_api.sh
+++ b/yelp_package/dockerfiles/itest/k8s/wait_paasta_api.sh
@@ -9,7 +9,7 @@ shift
 cmd="$@"
 
 until [ $(curl -o /dev/null -s -w %{http_code} http://127.0.0.1:${PAASTA_API_PORT}/v1/version) = 200 ]; do
-      >&2 echo "PaaSTA APi is unavailable."
+      >&2 echo "PaaSTA API is unavailable."
         sleep 1
         count=$((${count}+1))
         if [ ${count} -gt 9 ]; then
--- a/yelp_package/dockerfiles/itest/k8s/wait_paasta_api.sh
+++ b/yelp_package/dockerfiles/itest/k8s/wait_paasta_api.sh
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -e
+
+count=1
+
+host="("
+shift
+cmd="$@"
+
+until [ $(curl -o /dev/null -s -w %{http_code} http://127.0.0.1:${PAASTA_API_PORT}/v1/version) = 200 ]; do
+      >&2 echo "PaaSTA APi is unavailable."
+        sleep 1
+        count=$((${count}+1))
+        if [ ${count} -gt 9 ]; then
+            echo "Timeout after trying ${count} times"
+            exit 1
+        fi
+    done
+
+    >&2 echo "PaaSTA API is up."
+    exec $cmd
+")"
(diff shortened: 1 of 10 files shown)


yelp_package/dockerfiles/itest/k8s/wait_paasta_api.sh
```diff
@@ -0,0 +1,23 @@
...
+        sleep 1
+        count=$((${count}+1))
+        if [ ${count} -gt 9 ]; then
+            echo "Timeout after trying ${count} times"
+            exit 1
+        fi
+    done
+
+    >&2 echo "PaaSTA API is up."
+    exec $cmd
```
EvanKrall: unindent these lines 4 spaces - `done` should match up with `until`.

EvanKrall:
```
exec "${cmd[@]}"
```

yelp_package/dockerfiles/itest/k8s/wait_paasta_api.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -e
+
+count=1
+
+host="("
+shift
+cmd="$@"
```
EvanKrall:
```
cmd=("$@")
```

yelp_package/dockerfiles/itest/k8s/wait_paasta_api.sh
```diff
@@ -0,0 +1,23 @@
...
+        count=$((${count}+1))
+        if [ ${count} -gt 9 ]; then
+            echo "Timeout after trying ${count} times"
+            exit 1
+        fi
+    done
+
+    >&2 echo "PaaSTA API is up."
+    exec $cmd
+")"
```
EvanKrall: what's up with this? won't this result in

```
-bash: ): command not found
```
?

(oh, the `exec` on the previous line prevents this line from ever running. Still, what's this doing here?)

yelp_package/dockerfiles/itest/k8s/wait_paasta_api.sh
```diff
@@ -0,0 +1,23 @@
...
+
+host="("
+shift
+cmd="$@"
+
+until [ $(curl -o /dev/null -s -w %{http_code} http://127.0.0.1:${PAASTA_API_PORT}/v1/version) = 200 ]; do
+      >&2 echo "PaaSTA API is unavailable."
+        sleep 1
+        count=$((${count}+1))
+        if [ ${count} -gt 9 ]; then
```
EvanKrall: I'd recommend running `shellcheck` on this script, which will probably tell you to put double quotes around the URL so that if PAASTA_API_PORT ends up with spaces in it, word splitting doesn't cause the URL to get split into multiple arguments for curl.

You probably also want double quotes around the whole `$(curl ...)` for similar reasons.

yelp_package/dockerfiles/itest/k8s/wait_paasta_api.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+set -e
+
+count=1
+
+host="("
```
EvanKrall: Doesn't look like $host is used anywhere.

yelp_package/dockerfiles/itest/k8s/wait_paasta_api.sh
EmanekaT: Oops. I just realized one of my vim plugins (I don't know which one) converted the script I copied to this werid thing. It worked well so I never notice.

PR #2738, opened 2020-03-23
Don't bother pinging authors after a mark-for-deployment

solarkennedy:


EvanKrall:
Actually, uh. Can you make sure we still ping in case we roll back or roll forward?

We really just want to avoid a second ping on startup.

solarkennedy:
@EvanKrall can you look at this one more time?

solarkennedy:
I'm going to close this for now.

PR #2739, opened 2020-03-24
Turn off pod eviction alerts for stateless services

EmanekaT:
I just commented it out for all services because I think this should not happen on stateful sets. They all use ebs volumes.


EmanekaT:
hmm, I don't know where all of these commits are from. I'll clean them up.

PR #2740, opened 2020-03-24
Turn off pod eviction alerts for stateless services

EmanekaT:
I just commented it out for all services because I think this should not happen on stateful sets. They all use ebs volumes.

### DIFF
--- a/paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
+++ b/paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
@@ -148,7 +148,7 @@ def main() -> None:
     kube_client = KubeClient()
 
     evicted_pods = evicted_pods_per_service(kube_client)
-    notify_service_owners(evicted_pods, args.soa_dir, args.dry_run)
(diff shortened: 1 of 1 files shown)


paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
```diff
@@ -148,7 +148,7 @@ def main() -> None:
     kube_client = KubeClient()
 
     evicted_pods = evicted_pods_per_service(kube_client)
-    notify_service_owners(evicted_pods, args.soa_dir, args.dry_run)
+    # notify_service_owners(evicted_pods, args.soa_dir, args.dry_run)
```
chlgit: Could you add a comment which links to ticket number so we won't forget why we do this.

Rob-Johnson: can you just delete the line rather than commenting it out?

qui:
I agree this isn't useful in some cases, but I think the notification matters for disk usage. The message can say something like, "Pod ephemeral local storage usage exceeds the total limit of containers 3256Mi", which means the pod actually went over the disk quota and the service owner might want to increase it.

PR #2743, opened 2020-03-24
FLINK-2303 further changes

JoeMalt:
- if not all jobs are being shown, print a warning to use -v to see them all
- -v option now shows all jobs, -vv needed to see pods, exceptions etc
- only color the status, rather than the whole line

Note: I planned to put Sofya on this PR too but she doesn't seem to be a member of the Yelp GH org, I'll send the PR to her for an unofficial review

Screenshots:
- normal usage (no -v), showing warning: https://yelp-shootie.appspot.com/shot/4582659305504768
- with -v, showing all jobs: https://yelp-shootie.appspot.com/shot/5166412535431168
- with -vv, showing all jobs and pods, exceptions etc: https://yelp-shootie.appspot.com/shot/4884504471666688
- with <3 jobs available, warning isn't shown: https://yelp-shootie.appspot.com/shot/4503976779710464

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1096,6 +1096,10 @@ def print_flink_status(
                     )
     if verbose > 1 and len(status.pod_status) > 0:
         append_pod_status(status.pod_status, output)
+    if verbose == 1:
+        output.append(
+            PaastaColors.yellow(f"    Use -vv to view pod information and exceptions")
+        )
     return 0
 
 
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1105,9 +1105,7 @@ def print_flink_status(
     if verbose and len(status.pod_status) > 0:
         append_pod_status(status.pod_status, output)
     if verbose == 1 and status.exceptions:
-        output.append(
-            PaastaColors.yellow(f"    Use -vv to view pod information and exceptions")
-        )
+        output.append(PaastaColors.yellow(f"    Use -vv to view exceptions"))
     return 0
 
 
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -916,7 +916,9 @@ def append_pod_status(pod_status, output: List[str]):
             PaastaColors.green
             if pod["phase"] == "Running" and pod["container_state"] == "Running"
             else PaastaColors.red
-            if pod["phase"] in ("Failed", "CrashLoopBackOff")
+            # pods can get stuck in phase: Running and state: CrashLoopBackOff, so check for that
+            if pod["phase"] == "Failed"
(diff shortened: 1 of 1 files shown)


AshishKhatkar:
Can we add a warning in case of -v to let users know they can see exceptions and pod information using -vv ? 
A suggestion maybe show pod information in case of -v as well.

JoeMalt:
@AshishKhatkar +1 for showing a warning with -v to use -vv to see exceptions and pod information, will add this.

I would prefer not to put pod info in -v, so that there's some way to keep the "display all jobs but no pods" mode that was the default before these changes. Happy to change if you feel strongly about it though.

poros:
> @AshishKhatkar +1 for showing a warning with -v to use -vv to see exceptions and pod information, will add this.
> 
> I would prefer not to put pod info in -v, so that there's some way to keep the "display all jobs but no pods" mode that was the default before these changes. Happy to change if you feel strongly about it though.

I feel pretty strongly about it :P
I'd like to see the pods information with {{-v}} because it similar to what regular {{paasta status}} is doing and because it is hard to find anything with all the clutter coming from the tracebacks when there are many jobs.

@JoeMalt would these two motives suffice?

JoeMalt:
@poros That makes a lot of sense, will add pods to -v.

poros:
@JoeMalt on line 909 we are adding the pod status info. Since you are at it, could you please change the color of the `phase` information to be green/red/yellow more or less like `paasta status` does (they do `health` but the concept is similar)?

I think that we might have some kind of corner case where a pod is running but in crashloopbackoff or something like that, so you might need to be a bit more clever there (https://github.com/Yelp/paasta/pull/2643).

JoeMalt:
@poros changes made.

New screenshot with -v: https://yelp-shootie.appspot.com/shot/5696753353883648

AshishKhatkar:
> @poros changes made.
> 
> New screenshot with -v: https://yelp-shootie.appspot.com/shot/5696753353883648

Looks great. Thanks @JoeMalt

paasta_tools/cli/cmds/status.py
```diff
@@ -912,11 +912,19 @@ def append_pod_status(pod_status, output: List[str]):
         ("Pod Name", "Host", "Phase", "Uptime")
     ]
     for pod in pod_status:
+        color_fn = (
+            PaastaColors.green
+            if pod["phase"] == "Running" and pod["container_state"] == "Running"
+            else PaastaColors.red
+            if pod["phase"] in ("Failed", "CrashLoopBackOff")
```
AshishKhatkar: Crashloopbackoff is `pod['container_state']`

JoeMalt: fixed in latest version

screenshot: https://yelp-shootie.appspot.com/shot/5689814631317504

also fixed headings for -v (removed Job ID): https://yelp-shootie.appspot.com/shot/5668874989600768

PR #2746, opened 2020-03-24
Expose HPA status and SFX metrics in paasta status

EmanekaT:
This branch is based on PAASTA-16313. 
This is the api part of implementation for HPA added. There will be anotehr PR for the client part. 
Tested with itest. More tests will be added in the client part. 
```
 $ curl http://0.0.0.0:33222/v1/services/compute-infra-test-service/autoscaling/status
{"service": "compute-infra-test-service", "instance": "autoscaling", "git_sha": "a0ec9b36", "kubernetes": {"app_count": 1, "desired_state": "start", "bounce_method": "crossover", "app_id": "compute-infra-test-service-autoscaling", "pods": [], "replicasets": [], "expected_instance_count": 1, "deploy_status": "Waiting", "running_instance_count": 0, "create_timestamp": 1584934018.0, "namespace": "paasta", "autoscaling_status": {"min_instances": 1, "max_instances": 10, "metrics": [{"name": "memory", "target_value": "7000"}, {"name": "http", "target_value": "70"}, {"name": "uwsgi", "target_value": "70"}, {"name": "external", "target_value": "20"}, {"name": "cpu", "target_value": "7000"}], "desired_replicas": 0, "last_scale_time": "N/A"}, "evicted_count": 0}}(py36-linux)
```

Note that the tests depend on #2737 . So I'll add them after that is merged.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -256,7 +256,7 @@ def paasta_status_on_api_endpoint(
         return 1
     except Exception as e:
         output.append(PaastaColors.red(f"Exception when talking to the API:"))
-        output.append(e)
+        output.append(str(e))
         return 1
 
     if status.git_sha != "":
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -16,7 +16,6 @@ import concurrent.futures
 import difflib
 import shutil
 import sys
-import traceback
 from collections import defaultdict
 from collections import OrderedDict
 from datetime import datetime
@@ -255,10 +254,9 @@ def paasta_status_on_api_endpoint(
             PaastaColors.red(f"Could not connect to API: {exc.__class__.__name__}")
         )
         return 1
-    except Exception:
-        tb = sys.exc_info()[2]
+    except Exception as e:
         output.append(PaastaColors.red(f"Exception when talking to the API:"))
-        output.extend(line.strip() for line in traceback.format_tb(tb))
+        output.append(e)
         return 1
 
     if status.git_sha != "":
--- a/paasta_tools/instance/hpa_metrics_parser.py
+++ b/paasta_tools/instance/hpa_metrics_parser.py
@@ -9,7 +9,6 @@ class HPAMetricsParser:
         Parse target metrics.
         """
         metric_spec = getattr(metric, metric.type.lower())
-        print(metric_spec)
         status = {}
         switchers = {
             "Pods": self.parse_pod_metric,
@@ -38,7 +37,6 @@ class HPAMetricsParser:
         )
         return status
 
-    # External metrics
     def parse_external_metric(self, metric_spec, status):
         status[self.NAME] = metric_spec.metric_name
         status[self.TARGET] = (
@@ -55,7 +53,6 @@ class HPAMetricsParser:
             else metric_spec.current_value
         )
 
-    # Pod metrics
     def parse_pod_metric(self, metric_spec, status):
         status[self.NAME] = metric_spec.metric_name
         status[self.TARGET] = metric_spec.target_average_value
@@ -64,7 +61,6 @@ class HPAMetricsParser:
         status[self.NAME] = metric_spec.metric_name
         status[self.CURRENT] = metric_spec.current_value
 
-    # Resource metrics
     def parse_resource_metric(self, metric_spec, status):
         status[self.NAME] = metric_spec.name
         status[self.TARGET] = (
--- a/paasta_tools/instance/hpa_metrics_parser.py
+++ b/paasta_tools/instance/hpa_metrics_parser.py
@@ -1,0 +1,82 @@
+class HPAMetricsParser:
+    def __init__(self, hpa):
+        self.NAME = "name"
+        self.TARGET = "target_value"
+        self.CURRENT = "current_value"
+
+    def parse_target(self, metric):
+        """
+        Parse target metrics.
+        """
+        metric_spec = getattr(metric, metric.type.lower())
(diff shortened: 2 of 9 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -414,7 +412,7 @@ def get_hpa_metric_spec(
         self, name: str, cluster: str, namespace: str = "paasta"
     ) -> Optional[V2beta1HorizontalPodAutoscaler]:
         hpa_config = self.config_dict["horizontal_autoscaling"]
-        min_replicas = hpa_config.get("min_replicas", 0)
+        min_replicas = hpa_config.get("min_replicas", 1)
```
EmanekaT: Default to 1 unless users want 0.

paasta_tools/cli/cmds/status.py
```diff
@@ -255,10 +254,9 @@ def paasta_status_on_api_endpoint(
             PaastaColors.red(f"Could not connect to API: {exc.__class__.__name__}")
         )
         return 1
-    except Exception:
-        tb = sys.exc_info()[2]
+    except Exception as e:
         output.append(PaastaColors.red(f"Exception when talking to the API:"))
-        output.extend(line.strip() for line in traceback.format_tb(tb))
+        output.append(e)
```
EmanekaT: The old code removed the most important part of error message, and wasted me a lot of time to pinpoint the simple error.

paasta_tools/instance/hpa_metrics_parser.py
```diff
@@ -0,0 +1,82 @@
...
+        self.NAME = "name"
+        self.TARGET = "target_value"
+        self.CURRENT = "current_value"
+
+    def parse_target(self, metric):
+        """
+        Parse target metrics.
+        """
+        metric_spec = getattr(metric, metric.type.lower())
+        print(metric_spec)
```
chlgit: do you want to remove this?

paasta_tools/instance/hpa_metrics_parser.py
```diff
@@ -0,0 +1,82 @@
...
+            "External": self.parse_external_metric_current,
+            "Resource": self.parse_resource_metric_current,
+        }
+        switchers[metric.type](metric_spec, status)
+        status[self.CURRENT] = (
+            str(status[self.CURRENT]) if status[self.CURRENT] else "N/A"
+        )
+        return status
+
+    # External metrics
```
chlgit: I don't think you need this comment.

paasta_tools/instance/hpa_metrics_parser.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def parse_external_metric_current(self, metric_spec, status):
+        status[self.NAME] = metric_spec.metric_name
+        status[self.CURRENT] = (
+            metric_spec.current_average_value
+            if getattr(metric_spec, "current_average_value")
+            else metric_spec.current_value
+        )
+
+    # Pod metrics
```
chlgit: And this. The function name talks for themselves.

paasta_tools/instance/hpa_metrics_parser.py
```diff
@@ -0,0 +1,82 @@
...
+    def parse_pod_metric(self, metric_spec, status):
+        status[self.NAME] = metric_spec.metric_name
+        status[self.TARGET] = metric_spec.target_average_value
+
+    def parse_pod_metric_current(self, metric_spec, status):
+        status[self.NAME] = metric_spec.metric_name
+        status[self.CURRENT] = metric_spec.current_value
+
+    # Resource metrics
```
chlgit: and this too.

paasta_tools/instance/kubernetes.py
```diff
@@ -73,6 +75,44 @@ def set_cr_desired_state(
...
+def autoscaling_status(
+    kube_client: kubernetes_tools.KubeClient,
+    job_config: LongRunningServiceConfig,
+    namespace: str,
+):
+    status = {}
+    try:
+        hpa = kube_client.autoscaling.read_namespaced_horizontal_pod_autoscaler(
+            name=job_config.get_sanitised_deployment_name(), namespace=namespace
+        )
```
chlgit: I think only this line might throw ApiException. Could you move the remaining out of the try block?

paasta_tools/instance/kubernetes.py
```diff
@@ -73,6 +75,44 @@ def set_cr_desired_state(
...
+        status["desired_replicas"] = hpa.status.desired_replicas
+        status["last_scale_time"] = (
+            hpa.status.last_scale_time.replace(tzinfo=pytz.UTC).isoformat()
+            if getattr(hpa.status, "last_scale_time")
+            else "N/A"
+        )
+    except ApiException as e:
+        if e.status == 404:
+            # This means that HPA doesn't exist. Return nothing.
+            return {}
```
chlgit: Out of curiosity, will this mask the error that HPA doesn't exist but it should?

EmanekaT: Good point.

paasta_tools/instance/kubernetes.py
EmanekaT: Instead of this, I caught all exceptions and throw a runtime exception. If anything goes wrong, we can pinpoint the cause easily.

chlgit: Is this a good idea? If the api call failed. I think we should make this function fail gracefully rather than throw the entire traceback. Similar to what we did for https://jira.yelpcorp.com/browse/PAASTA-16182

EmanekaT: Fixed.

PR #2747, opened 2020-03-25
Log mesos container id and memory limit with OOMkills

matfra:
For autotune, we want to be able to join the oomkill logs with tmp_paastaallocations logs.
However currently the oomkill log only contains the docker id while paasta alllocation only contains docker id (when running on k8s) or mesos_task_id (when running on mesos)
This PR adds mesos_task_id to the log. If running on kubernetes, it will have the value 'null'

Also adding the useful memory limit of the container at the time of the OOM

### DIFF
--- a/paasta_tools/oom_logger.py
+++ b/paasta_tools/oom_logger.py
@@ -65,6 +65,8 @@ LogLine = namedtuple(
         "service",
         "instance",
         "process_name",
+        "mesos_container_id",
+        "mem_limit",
     ],
 )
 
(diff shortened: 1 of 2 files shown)


paasta_tools/oom_logger.py
```diff
@@ -184,6 +189,8 @@ def main():
         env_vars = get_container_env_as_dict(docker_inspect)
         service = env_vars.get("PAASTA_SERVICE", "unknown")
         instance = env_vars.get("PAASTA_INSTANCE", "unknown")
+        mesos_container_id = env_vars.get("MESOS_CONTAINER_NAME", "mesos-null")
```
qui: Smart, then we can use the same regex and get "null" as the container id for k8s tasks :) Took me a minute to figure out why we were doing this.

matfra: yeah a bit hacky :-D

PR #2748, opened 2020-03-25
Emit metric for OOMs

qui:
I think this will be more useful for dashboards to see the impact of memory request changes. I don't think you can count events in sfx, unfortunately.


solarkennedy:
Cool.

ddelnano:
🚢

PR #2750, opened 2020-03-25
Increase default cpu/mem for tasks

qui:
With automatic rightsizing, the goal is for most services to not
need to set cpu/mem in their configs. When a service first starts,
we won't have any usage data to make recommendations on, so
the defaults should be generous. The rightsizer will scale them
down as appropriate after that. This maintains the 1:4 ratio
of cpus to mem, which matches the makeup of our clusters pretty
well.


qui:
I could limit this to long running instances (marathon/k8s) if you all would prefer that.

solarkennedy:
No I think it is OK.

matfra:
We need to make sure than sooner or later we will rightsize automatically all the tasks that can potentially inherit from this default. Is that the case for tron jobs as well ?

qui:
I think rightsizing Tron jobs is on our long-term roadmap, and in the meantime, I don't think having this default will hurt too much. Most people do choose cpus/mem now.

PR #2752, opened 2020-03-27
Remove flaky deployd tests

EmanekaT:


solarkennedy:
I vote yes. deployd is going away in favor of k8s, and I think this flakey test isn't worth the protection.

PR #2753, opened 2020-03-27
Double the default requested disk for services to 2g

solarkennedy:
I get it, we are trying to cram as much as we can on our servers.

But the thing is, our docker images for many of our services are over 1g, and it causes constant pages and evictions.

I think this is one place where we need to be a bit more honest about how big our docker images are.


matfra:
Will this create a massive big bounce ?

solarkennedy:
yes

matfra:
so let's ship carefully, first let's get some number on will this fit or not on the current hosts we have. if it does, cool. if it doesn't we'll have to add more EBS

PR #2754, opened 2020-03-30
Add --strip-headers option to paasta logs (PAASTA-16280)

JoeMalt:
This PR adds --strip-headers, which causes paasta logs to just print the timestamp and message, skipping cluster, instance, log level and component. This was requested by Connectors, because the long headers make Flink logs quite difficult to read.

**Examples**
without strip-headers: https://yelp-shootie.appspot.com/shot/4643532397084672
with strip-headers: https://yelp-shootie.appspot.com/shot/4551395315810304

I would quite like to remove the 

This was requested for Flink services, but works equally well for regular PaaSTA services. Here it is for Apollo:
without strip-headers: https://yelp-shootie.appspot.com/shot/4594127790931968
with strip-headers: https://yelp-shootie.appspot.com/shot/4594127790931968

I'd like to strip the hostname and commit too, but those are part of the `message` field in the JSON and not guaranteed to be present, so I'm leaving them for now.

@Rob-Johnson assigning to you as CI onpoint, feel free to reassign.

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -169,6 +169,13 @@ def add_subparser(subparsers) -> None:
         "--lines +100 would give you the first 100 lines of logs. Defaults to the latest line's offset",
         type=int,
     )
+    status_parser.add_argument(
+        "-sh",
+        "--strip-headers",
+        dest="strip_headers",
+        help="Print log lines without header information.",
+        action="store_true",
+    )
     default_component_string = ",".join(DEFAULT_COMPONENTS)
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/logs.py
```diff
@@ -169,6 +169,13 @@ def add_subparser(subparsers) -> None:
         "--lines +100 would give you the first 100 lines of logs. Defaults to the latest line's offset",
         type=int,
     )
+    status_parser.add_argument(
+        "-sh",
```
Bronek: please do not use two letters for a short option because that's very confusing. Choose one letter instead (capitals are fine, too)

PR #2757, opened 2020-03-30
Allow users to customize signalflow query for http/uwsgi based extern…

mzq592:
After thinking about it, I think we should just allow users to customize their own signalflow query for http/uwsgi.


solarkennedy:
@EvanKrall can you review this one? It is a pretty advanced feature that we currently don't have in mesos/paasta. YAGNI?

mattmb:

PR #2761, opened 2020-04-02
Provide better status updates from Kubernetes

drmorr0:
We read the Kubernetes event stream to find the status message.
Unfortunately there is no way to get the event stream for _all_
"subobjects" from a Deployment, so we just look at a whitelisted set of
Kubernetes objects to get events from.  We then aggregate all of the
event streams, sort them, and take the most recent one as the status
message.

I also fixed a minor bug where services that have been `paasta stop` ed
would still show up as "Waiting" instead of "Stopped".

You can see the results of this here: https://fluffy.yelpcorp.com/i/6Tv1p2vQ5Kmssb5LTJHnZZJvnFgTVfbT.html

Line 10 shows that the container is failing for some reason
Line 26 shows that the service is stopped (this used to say `Waiting`)
Line 58 shows that there are capacity issues in the cluster
Lines 74, 98, and 110 show "running" services -- I'm not sure if this is useful or if it's just confusing.

### DIFF
--- a/paasta_tools/instance/kubernetes.py
+++ b/paasta_tools/instance/kubernetes.py
@@ -1,5 +1,6 @@
 from typing import Any
 from typing import Dict
+from typing import List
 from typing import Mapping
 from typing import MutableMapping
 from typing import Sequence
@@ -9,6 +10,7 @@ import pytz
 from kubernetes.client import V1Pod
 from kubernetes.client import V1ReplicaSet
 from kubernetes.client.rest import ApiException
+from mypy_extensions import TypedDict
 
 from paasta_tools import cassandracluster_tools
 from paasta_tools import flink_tools
@@ -37,6 +39,14 @@ INSTANCE_TYPE_CR_ID = dict(
     cassandracluster=cassandracluster_tools.cr_id,
     kafkacluster=kafkacluster_tools.cr_id,
 )
+
+
+class KubernetesAutoscalingStatusDict(TypedDict):
+    min_instances: int
+    max_instances: int
+    metrics: List
+    desired_replicas: int
+    last_scale_time: str
 
 
 def cr_id(service: str, instance: str, instance_type: str) -> Mapping[str, str]:
@@ -79,13 +89,23 @@ def autoscaling_status(
     kube_client: kubernetes_tools.KubeClient,
     job_config: LongRunningServiceConfig,
     namespace: str,
-):
-    status = {}
-    hpa = kube_client.autoscaling.read_namespaced_horizontal_pod_autoscaler(
-        name=job_config.get_sanitised_deployment_name(), namespace=namespace
-    )
-    status["min_instances"] = hpa.spec.min_replicas
-    status["max_instances"] = hpa.spec.max_replicas
+) -> KubernetesAutoscalingStatusDict:
+    try:
+        hpa = kube_client.autoscaling.read_namespaced_horizontal_pod_autoscaler(
+            name=job_config.get_sanitised_deployment_name(), namespace=namespace
+        )
+    except ApiException as e:
+        if e.status == 404:
+            return KubernetesAutoscalingStatusDict(
+                min_instances=-1,
+                max_instances=-1,
+                metrics=[],
+                desired_replicas=-1,
+                last_scale_time="unknown (could not find HPA object)",
+            )
+        else:
+            raise
+
     # Parse metrics sources, based on
     # https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V2beta1ExternalMetricSource.md#v2beta1externalmetricsource
     metric_stats = []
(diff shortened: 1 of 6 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1865,13 +1868,90 @@ def update_stateful_set(
...
+
+    for ss in kube_client.deployments.list_namespaced_stateful_set(
+        app.metadata.namespace, label_selector=ls,
+    ).items:
+        events += get_events_for_object(kube_client, ss, "StatefulSet")
+
+    return sorted(
+        events,
+        key=lambda x: (
+            x.get("lastTimestamp") or x.get("eventTime") or x.get("firstTimestamp")
```
keymone, line 1928: the assumption is that at least one of these exists, what if they don't? i'd add something like `or 0` in the end or fail explicitly with relevant error message.

paasta_tools/kubernetes_tools.py
```diff
@@ -1865,13 +1868,90 @@ def update_stateful_set(
     )
 
 
+def get_events_for_object(
```
enochtangg, line 1871: Seems like this is the same thing as `list_namespaced_event()`?  https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#list_namespaced_event

paasta_tools/kubernetes_tools.py
```diff
@@ -1865,13 +1868,90 @@ def update_stateful_set(
...
+    kube_client: KubeClient,
+    obj: Union[V1Pod, V1Deployment, V1StatefulSet, V1ReplicaSet],
+    kind: str,  # for some reason, obj.kind isn't populated when this function is called so we pass it in by hand
+) -> List[V1Event]:
+    host = KubeConfiguration().host
+
+    # The python kubernetes client doesn't support the V1Events API
+    # yet, so we have to make the request by hand (we need the V1Events
+    # API so that we can query by the involvedObject.name/kind)
+    events = kube_client.request(
```
enochtangg, line 1881: Seems like this is the same thing as list_namespaced_event()? https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CoreV1Api.md#list_namespaced_event

drmorr0: As best as I can tell, the one under the CoreV1 API does not actually return the "correct" event list.  I did a bunch of experiments with this and I couldn't get any events back, despite events showing up in the event stream.

drmorr0: I updated the comment to this effect.

paasta_tools/kubernetes_tools.py
```diff
@@ -1865,13 +1868,90 @@ def update_stateful_set(
...
+        query_params={
+            "fieldSelector": f"involvedObject.name={obj.metadata.name},"
+            f"involvedObject.kind={kind}"
+        },
+    )
+    parsed_events = json.loads(events.data)
+    return parsed_events["items"]
+
+
+def get_all_events_for_service(
```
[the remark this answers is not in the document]
drmorr0: Fixed.

paasta_tools/kubernetes_tools.py
drmorr0: Fixed

paasta_tools/instance/kubernetes.py
```diff
@@ -81,9 +81,20 @@ def autoscaling_status(
...
     status = {}
-    hpa = kube_client.autoscaling.read_namespaced_horizontal_pod_autoscaler(
-        name=job_config.get_sanitised_deployment_name(), namespace=namespace
-    )
+    try:
+        hpa = kube_client.autoscaling.read_namespaced_horizontal_pod_autoscaler(
+            name=job_config.get_sanitised_deployment_name(), namespace=namespace
+        )
+    except ApiException as e:
+        if e.status == 404:
```
drmorr0, line 89: @EvanKrall Can you take a look at this section?  I think this is a better solution to the `paasta status` error than just adding something to `error_message`, because `error_message` prevents the rest of the status output from being displayed.

drmorr0:
I added a fix to prevent a massive traceback if an HPA object doesn't exist (see also https://github.com/Yelp/paasta/pull/2762)

drmorr0:
Also (hopefully) fixed the mypy errors.

paasta_tools/instance/kubernetes.py
EvanKrall: Is there some easy way we can have `status` output a combination of `error_message` and actual responses?

Otherwise, yeah, ship. Anything that makes `paasta status` degrade more gracefully in the face of its many failure modes is good.

drmorr0: I think what I'd ideally do here is "add a command-line option that displays errors at the bottom", but that sounds like more work than what I want to do in this PR.  I'll make a ticket, though.

PR #2773, opened 2020-04-13
Delete and reschedule paasta contract monitor deployments for better alerting

enochtangg:
### Summary
The following code is used to delete paasta-contract-monitor deployments and its respective pods in order for them to be automatically rescheduled via `setup_kubernetes_job`. By default, when a deployment has been changed or pods encounters a failure, kubernetes simply tries to restart these pods on the same node. However, the purpose of the pcm is to test the entire scheduling/container lifecycle to ensure that deployments working. Therefore, we want to delete and reschedule these deployments from scratch.

Flow:
`delete_paasta_contract_monitor_job.py`(cron) will delete pcm deployments -> `setup_kubernetes_job` will automatically detect missing deployment and reschedule them

### Testing Done
* Added units for new cron job
* `make test`
* Manually tested on kubestage master, deployment was successfully deleted and new pods were reschedule (checking creationTimestamp)


paasta_tools/delete_paasta_contract_monitor_job.py
```diff
@@ -0,0 +1,68 @@
...
+from subprocess import PIPE
+
+from paasta_tools.kubernetes_tools import delete_deployment
+from paasta_tools.kubernetes_tools import ensure_namespace
+from paasta_tools.kubernetes_tools import KubeClient
+
+log = logging.getLogger(__name__)
+
+
+def get_paasta_contract_monitor_deployment_names():
```
chlgit, line 34: Is get_sanitized_deployment_name not enough here? https://sourcegraph.yelpcorp.com/mirrors/Yelp/paasta@0dd9866f88d51bfe0229f09e2db61db404903301/-/blob/paasta_tools/kubernetes_tools.py#L1056:9

paasta_tools/delete_paasta_contract_monitor_job.py
```diff
@@ -0,0 +1,68 @@
...
+    p2 = subprocess.Popen(
+        ["grep", "paasta-contract-monitor"], stdin=p1.stdout, stdout=PIPE
+    )
+    p1.stdout.close()
+    output = p2.communicate()[0].decode("utf-8").split()
+    return output
+
+
+def main() -> None:
+    pcm_deployment_names = get_paasta_contract_monitor_deployment_names()
```
kawaiwanyelp, line 53: I think this script might be _too specific_ to the PCM, i.e. it specifically finds the PCM deployment to delete it. I think it goes against the spirit of PaaSTA as a service deployer, in that it shouldn't assume the existence of any one service. I'd suggest instead accepting a service and instance as an argument (just like any other paasta command), and finding the deployment name from that instead.

paasta_tools/delete_paasta_contract_monitor_job.py
```diff
@@ -0,0 +1,68 @@
...
+    log.debug(f"Deleting deployments: {pcm_deployment_names}")
+    kube_client = KubeClient()
+    ensure_namespace(kube_client=kube_client, namespace="paasta")
+
+    for deployment_name in pcm_deployment_names:
+        try:
+            delete_deployment(kube_client=kube_client, deployment_name=deployment_name)
+        except Exception as err:
+            log.error(f"Unable to delete {deployment_name}: {err}")
```
kawaiwanyelp, line 62: You're only logging the error here. I think you should also exit non-zero in this case. This way we can get alerted if something goes wrong.

paasta_tools/delete_paasta_contract_monitor_job.py
kawaiwanyelp: Seconded.

tests/test_delete_paasta_contract_monitor.py
```diff
@@ -0,0 +1,47 @@
+import mock
+from pytest import raises
+
+from paasta_tools.delete_paasta_contract_monitor_job import (
+    get_paasta_contract_monitor_deployment_names,
+)
+from paasta_tools.delete_paasta_contract_monitor_job import main
+
+
+def test_main():
```
kawaiwanyelp, line 10: You should also test the case where the deletion fails.

paasta_tools/delete_paasta_contract_monitor_job.py
```diff
@@ -0,0 +1,68 @@
...
+        [
+            "python",
+            "-m",
+            "paasta_tools.list_kubernetes_service_instances",
+            "--sanitise",
+        ],
+        stdout=PIPE,
+    )
+    p2 = subprocess.Popen(
+        ["grep", "paasta-contract-monitor"], stdin=p1.stdout, stdout=PIPE
```
keymone, line 45: `paasta-contract-monitor` is a service, it feels weird to me that paasta codebase should even know about it

paasta_tools/delete_paasta_contract_monitor_job.py
solarkennedy: I agree. Just make this a general "delete_deployments" command and it can be useful in many ways.
Heck we had like hundereds of deployments we wanted to delete yesterday!

paasta_tools/delete_paasta_contract_monitor_job.py
solarkennedy: Yea uh. Why do we need to subprocess here? Can't we just call our own functions?

paasta_tools/delete_paasta_contract_monitor_job.py
solarkennedy: yea append  to a list of bools and return `all()` or something

paasta_tools/delete_kubernetes_deployments.py
```diff
@@ -0,0 +1,85 @@
...
+log = logging.getLogger(__name__)
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description="Deletes list of deployments.")
+    parser.add_argument(
+        "service_instance_list",
+        nargs="+",
+        help="The list of service instances to delete",
+        metavar="SERVICE%sINSTANCE" % SPACER,
```
kawaiwanyelp, line 43: Since we're on python3, please use format strings.

paasta_tools/delete_kubernetes_deployments.py
```diff
@@ -0,0 +1,85 @@
...
+    app_names = []
+    for service_instance in service_instance_list:
+        try:
+            service, instance, _, __ = decompose_job_id(service_instance)
+            app_name = get_kubernetes_app_name(service, instance)
+            app_names.append(app_name)
+        except InvalidJobNameError:
+            log.error(
+                "Invalid service instance specified. Format is service%sinstance."
+                % SPACER
```
kawaiwanyelp, line 59: Since we're on python3, please use format strings.

paasta_tools/delete_kubernetes_deployments.py
```diff
@@ -0,0 +1,85 @@
...
+    service_instance_list = args.service_instance_list
+    deployment_names = get_deployment_names_from_list(service_instance_list)
+
+    log.debug(f"Deleting deployments: {deployment_names}")
+    kube_client = KubeClient()
+    ensure_namespace(kube_client=kube_client, namespace="paasta")
+
+    for deployment_name in deployment_names:
+        try:
+            delete_deployment(kube_client=kube_client, deployment_name=deployment_name)
```
kawaiwanyelp, line 76: nit: I'd prefer that you add a log line here specifying which deployment you're deleting, instead of one big log line saying ones you will delete (above). That way, if there are multiple and one fails, you know which ones succeeded in what order and which one did not.

Also, make sure that the log level is above debug since you don't have a verbose mode. Alternatively, set the log level to debug so everything shows up.

PR #2775, opened 2020-04-13
Added aio speedups packages

solarkennedy:
Although this didn't seem to improve our situation, the aiohttp docs recommend installing these "speedup" packages to get better performance when doing lots of aio http requests.

I'm particularly interested in aiodns, which should be used automatically when available (again, didn't help our particular issue)


solarkennedy:
I'm going to discard this for now. Seems like a safe nice to have, but I can't tell if it makes a noticeable improvement.

PR #2776, opened 2020-04-13
Support envoy in pod readiness probe

ymilki:
Provides a new pod readiness probe that supports checking both smartstack and envoy when considering pod readiness.

Envoy readiness means:
* an egress cluster exists with one of the backends matching the pod ip and container port

In the case that the service is full mesh enabled:
*  an ingress cluster exists with a single backend matching the pod ip and container port
* an egress cluster exists with one of the backends matching the host ip and an unknown port

This check has several gates:
* `readiness_check_script`: Location of the script in the hacheck container
* `bounce_health_params:check_envoy`: Toggle checking envoy in readiness probe (this is disruptive)
* `envoy_toggles.yaml`: Toggle envoy for a service
* `full_mesh_envoy_toggles.yaml`: Toggle full mesh envoy for a service

Note: This PR does not enable the readiness check because changing the readiness check is disruptive to the k8s cluster. A subsequent PR will enable this readiness script.


ymilki:
This PR was rebuilt in #2889

PR #2777, opened 2020-04-13
Stop using the deprecated dict instead of string for marathon healthchecks

solarkennedy:
The client lib expects just a string now and does this dict for you.
This removes a ton of deprecation warnings when you run certain paasta commands that interact with marathon


solarkennedy:
Removes 

    WARNING:marathon:Deprecated: Using command as dict instead of string is deprecated

PR #2779, opened 2020-04-14
Always run all asyncio requests in a single session instead of one session per request

solarkennedy:
Although this makes our auto scaler slower, because it starts to queue requests, it does eliminate the timeout errors.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -269,17 +269,16 @@ def deserialize_historical_load(historical_load_bytes):
     return historical_load
 
 
-async def get_json_body_from_service(host, port, endpoint, timeout=2):
-    async with aiohttp.ClientSession(
-        conn_timeout=timeout, read_timeout=timeout
-    ) as session:
-        async with session.get(
-            f"http://{host}:{port}/{endpoint}", headers={"User-Agent": get_user_agent()}
-        ) as response:
-            return await response.json()
+async def get_json_body_from_service(host, port, endpoint, session):
+    async with session.get(
+        f"http://{host}:{port}/{endpoint}", headers={"User-Agent": get_user_agent()}
+    ) as response:
+        return await response.json()
 
 
-async def get_http_utilization_for_a_task(task, service, endpoint, json_mapper):
+async def get_http_utilization_for_a_task(
+    task, service, instance, endpoint, json_mapper, session
+):
     """
     Gets the task utilization by fetching json from an http endpoint
     and applying a function that maps it to a utilization.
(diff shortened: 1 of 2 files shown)


solarkennedy:
Testing done:
* unit test
* manually tested on prod (hot patch) and `wc`d the number of timeouts. After this patch it always was 0.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -331,12 +332,15 @@ async def get_http_utilization_for_all_tasks(
     utilization = []
     service = marathon_service_config.get_service()
 
-    futures = [
-        asyncio.ensure_future(
-            get_http_utilization_for_a_task(task, service, endpoint, json_mapper)
-        )
-        for task in marathon_tasks
-    ]
+    async with aiohttp.ClientSession() as session:
```
Rob-Johnson, line 335: can you leave a comment behind of why we do it like this, rather than one session per request? I've got a feeling we'll end up coming back to this one day and pick it as an easy place to make a performance improvement

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -331,12 +332,15 @@ async def get_http_utilization_for_all_tasks(
...
+        futures = [
+            asyncio.ensure_future(
+                get_http_utilization_for_a_task(
+                    task, service, endpoint, json_mapper, session
```
Rob-Johnson, line 339: can you kwargs this?

paasta_tools/autoscaling/autoscaling_service_lib.py
Rob-Johnson, line 335: also, do we need a timeout here?
https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientTimeout

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -268,17 +268,16 @@ def deserialize_historical_load(historical_load_bytes):
...
+async def get_json_body_from_service(host, port, endpoint, session):
+    async with session.get(
+        f"http://{host}:{port}/{endpoint}", headers={"User-Agent": get_user_agent()}
+    ) as response:
+        return await response.json()
 
 
-async def get_http_utilization_for_a_task(task, service, endpoint, json_mapper):
+async def get_http_utilization_for_a_task(
+    task, service, endpoint, json_mapper, session
```
matfra, line 279: I will give you 5 cents if you add instance here and in the log at line 303

matfra: and line 308

paasta_tools/autoscaling/autoscaling_service_lib.py
solarkennedy: added

solarkennedy: added

paasta_tools/autoscaling/autoscaling_service_lib.py
solarkennedy: added

PR #2781, opened 2020-04-15
Remove annotations from metadata spec and improve pause HPA

enochtangg:
### Summary
The following code is responsible for fixing the update deployment feedback loop caused by the revision annotation being bump after each iteration of `setup_kubernetes_job`. This code removes the need for pause_service_autoscaler to depend on annotations in the metadata spec. Instead, in order to determine whether the deployment has been paused, it manually just queries the k8s API, and check the annotations on the object directly.

In addition, we've also added an extra check so deployments will not continually be updated when paused. Conditions for `setup_kubernetes_job` to run `update()`:
- If k8s deployment config sha is different than existing one.
- If HPA is paused AND the deployment has not been paused yet.
- If HPA is resumed AND the deployment is still paused.

### Testing Done
* make test
* manually tested on kubestage. This time, we ensured that deployments are only updated once after being paused.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -47,6 +47,7 @@ from paasta_tools.bounce_lib import filter_tasks_in_smartstack
 from paasta_tools.bounce_lib import LockHeldException
 from paasta_tools.bounce_lib import LockTimeout
 from paasta_tools.bounce_lib import ZK_LOCK_CONNECT_TIMEOUT_S
+from paasta_tools.kubernetes_tools import get_annotations_for_deployment
 from paasta_tools.kubernetes_tools import KubeClient
 from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
 from paasta_tools.kubernetes_tools import write_annotation_for_kubernetes_service
@@ -846,10 +847,12 @@ def autoscaling_is_paused():
         return False
 
 
-def is_autoscaling_resumed(formatted_application: Union[V1Deployment, V1StatefulSet]):
-    if formatted_application.metadata.annotations:
-        if "is_paused" in formatted_application.metadata.annotations:
-            return formatted_application.metadata.annotations["is_paused"] == "False"
+def is_autoscaling_resumed(service_config: KubernetesDeploymentConfig):
+    kube_client = KubeClient()
+    annotations = get_annotations_for_deployment(kube_client, service_config)
+    if annotations:
+        if "is_paused" in annotations:
+            return annotations["is_paused"] == "False"
     return True
 
 
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -847,16 +847,14 @@ def autoscaling_is_paused():
         return False
 
 
-def is_autoscaling_resumed(service_config: KubernetesDeploymentConfig):
-    kube_client = KubeClient()
+def is_deployment_marked_paused(
+    kube_client: KubeClient, service_config: KubernetesDeploymentConfig
+):
     annotations = get_annotations_for_deployment(kube_client, service_config)
-    if annotations:
-        if "is_paused" in annotations:
-            return annotations["is_paused"] == "False"
-    return True
+    return annotations.get("is_paused", "False") == "True"
 
 
-def write_autoscaling_paused(
+def mark_deployment_as_paused(
     kube_client: KubeClient,
     service_config: KubernetesDeploymentConfig,
     formatted_application: Union[V1Deployment, V1StatefulSet],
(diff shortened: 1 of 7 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1005,6 +1005,12 @@ def get_desired_instances(self) -> int:
             )
         return instances
 
+    def get_annotations(self, kube_client: KubeClient) -> Dict:
```
drmorr0, line 1008: I don't think this function is really necessary anymore, it's literally just wrapping `get_annotations_for_deployment`.

paasta_tools/setup_kubernetes_job.py
```diff
@@ -142,12 +142,13 @@ def setup_kube_deployments(
                 app.kube_deployment.instance,
             ) not in existing_apps:
                 app.create(kube_client)
-            elif (
-                app.kube_deployment not in existing_kube_deployments
-                or autoscaling_is_paused()
-            ):
+            elif app.kube_deployment not in existing_kube_deployments:
```
drmorr0, line 145: Can you add some comments explaining what the logic is doing here?  It's a bit complicated to understand what's happening just by looking at the code.

Secondary question: can any of the cases where we call `app.update` be condensed?  That might make it easier to understand?  Or it might make it harder/more complicated, not sure.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -846,10 +847,12 @@ def autoscaling_is_paused():
         return False
 
 
-def is_autoscaling_resumed(formatted_application: Union[V1Deployment, V1StatefulSet]):
-    if formatted_application.metadata.annotations:
-        if "is_paused" in formatted_application.metadata.annotations:
-            return formatted_application.metadata.annotations["is_paused"] == "False"
+def is_autoscaling_resumed(service_config: KubernetesDeploymentConfig):
+    kube_client = KubeClient()
```
solarkennedy, line 851: Can you have this injected instead of instantiating a new client object?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -846,10 +847,12 @@ def autoscaling_is_paused():
...
+    annotations = get_annotations_for_deployment(kube_client, service_config)
+    if annotations:
+        if "is_paused" in annotations:
```
solarkennedy, line 854: can you make this a tiny bit more verbose, how about `paasta_service_autoscaler_is_paused` ?

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -846,10 +847,12 @@ def autoscaling_is_paused():
...
 
 
-def is_autoscaling_resumed(formatted_application: Union[V1Deployment, V1StatefulSet]):
-    if formatted_application.metadata.annotations:
-        if "is_paused" in formatted_application.metadata.annotations:
-            return formatted_application.metadata.annotations["is_paused"] == "False"
+def is_autoscaling_resumed(service_config: KubernetesDeploymentConfig):
+    kube_client = KubeClient()
+    annotations = get_annotations_for_deployment(kube_client, service_config)
+    if annotations:
```
solarkennedy, line 853: This can just be 

    return annotations.get('paused', False) == False

right?

Also, do we have to have this "False" as a string here?

Can it just be the fact that this *has* the annotation or not?

enochtangg: Yeah, "False" is going to be string when reading from k8s annotations. Also, I thought about having the annotation itself as an identifier of whether not it is paused, but for the sake of my own testing, it was better for me to explicitly see that the deployment has been paused/resumed. Maybe other devs would like to see that too?

paasta_tools/autoscaling/autoscaling_service_lib.py
enochtangg: Good point, to avoid confusion with `autoscaling_is_paused()` which reads pause time from zk, I've changed this to `is_deployment_marked_paused()`

paasta_tools/setup_kubernetes_job.py
enochtangg: Hmm, I thought about different condensed version of this, but that seems to make things very confusing. Having it explicitly laid out seems like the easiest way to comprehend.

paasta_tools/autoscaling/autoscaling_service_lib.py
```diff
@@ -846,14 +847,14 @@ def autoscaling_is_paused():
...
-def is_autoscaling_resumed(formatted_application: Union[V1Deployment, V1StatefulSet]):
-    if formatted_application.metadata.annotations:
-        if "is_paused" in formatted_application.metadata.annotations:
-            return formatted_application.metadata.annotations["is_paused"] == "False"
-    return True
+def is_deployment_marked_paused(
+    kube_client: KubeClient, service_config: KubernetesDeploymentConfig
+):
+    annotations = get_annotations_for_deployment(kube_client, service_config)
+    return annotations.get("is_paused", False) == "True"
```
drmorr0, line 854: This line is confusing at first glance, since you're comparing a bare literal `False` to a stringified `"True"`.  I suggest making the default `False` a string, and leaving a comment that the annotations from k8s are strings.

PR #2782, opened 2020-04-15
Fix kwarg name in oom count metric and add tests

qui:
The main change is just L176 in oom_logger.py. Probably spent more time on the tests than I should have, but I wanted to make sure they would catch any similar typos next time. 

test_main was never being run after https://github.com/Yelp/paasta/pull/2346 because that change moved the import of ScribeLogger into main() itself, so `from paasta_tools.oom_logger import ScribeLogger` would fail even if clog was in the venv. Since we are mocking ScribeLogger anyway, we can run the tests regardless of whether it is in the venv or not.

After I made that change, I fixed the mesos container id mock and added tests for the sfx method.

### DIFF
--- a/paasta_tools/oom_logger.py
+++ b/paasta_tools/oom_logger.py
@@ -46,13 +46,18 @@ from paasta_tools.utils import get_docker_client
 from paasta_tools.utils import load_system_paasta_config
 
 
+# Sorry to any non-yelpers but this won't
+# do much as our metrics and logging libs
+# are not open source
 try:
     import yelp_meteorite
 except ImportError:
-    # Sorry to any non-yelpers but you won't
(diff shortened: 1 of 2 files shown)


paasta_tools/oom_logger.py
```diff
@@ -173,17 +178,16 @@ def send_sfx_event(service, instance, cluster):
             "paasta.service.oom_events", dimensions=dimensions,
         )
         counter = yelp_meteorite.create_counter(
-            "paasta.service.oom_count", dimensions=dimensions,
+            "paasta.service.oom_count", default_dimensions=dimensions,
```
solarkennedy, line 181: Ah! Gosh. God bless autospec I guess.

PR #2784, opened 2020-04-16
Paasta 16560 check oom events thresholds

qui:
With autotuned_defaults running services with smaller margins on resources, we expect occasional OOMs. We want to keep these OOM alerts actionable, and don't want service owners to disable autotuned_defaults every time an OOM occurs. Even if a service isn't using autotuned_defaults, they don't necessarily need to bump memory on every OOM.

We will still be showing every OOM in dashboards and logs, and the rightsizer will take them into account - we just want to give the rightsizer some time to adjust. 

At the same time, the alert should still show serious issues with the memory value. We can explore different settings, but I'm thinking we can start with something conservative like alert_threshold=2 and check_every=5, so it will alert if any 2 containers in the same service instance OOM within a 5 min period.

### DIFF
--- a/paasta_tools/check_oom_events.py
+++ b/paasta_tools/check_oom_events.py
@@ -65,6 +65,20 @@ def parse_args(args):
         help="Sensu 'realert_every' to use.",
     )
     parser.add_argument(
+        "--check-interval",
+        dest="check_interval",
+        type=int,
+        default=1,
+        help="How often this check runs, in minutes.",
+    )
+    parser.add_argument(
+        "--alert-threshold",
+        dest="alert_threshold",
+        type=int,
+        default=1,
+        help="Number of OOM kills required in the check interval to send an alert.",
+    )
+    parser.add_argument(
         "-s",
         "--superregion",
         dest="superregion",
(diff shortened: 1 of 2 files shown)


paasta_tools/check_oom_events.py
```diff
@@ -104,49 +115,38 @@ def latest_oom_events(cluster, superregion, interval=60):
...
+    instance_name = f"{instance.service}.{instance.instance}"
     if not is_check_enabled:
-        return (
-            Status.OK,
-            "This check is disabled for {}.{}.".format(
-                instance.service, instance.instance
-            ),
-        )
+        return (Status.OK, f"This check is disabled for {instance_name}.")
     if len(oom_events) == 0:
```
sergsyelp, line 134: If we send None instead of an empty list as oom_event into the function for some reason len(None) going to err.

paasta_tools/check_oom_events.py
```diff
@@ -104,49 +115,38 @@ def latest_oom_events(cluster, superregion, interval=60):
...
-            % (
-                len(oom_events),
-                ",".join(
-                    sorted({e.process_name for e in oom_events if e.process_name})
-                ),
-                instance.service,
-                instance.instance,
-            ),
+            f"The Out Of Memory killer killed processes for {instance_name} "
+            f"in the last {interval_string}.",
```
matfra, line 143: It would be great to add some more information here.
Ideally we'd know if the service was using autotune and we would add something like:
If using autotune:
"This service is using autotuned_defaults, its memory requirements will be adjusted automatically at <time of next run of autotune>
If not using autotune:
"Did you know you could let PaaSTA determine how much memory this service need ? Just delete the mem line in your yelpsoa-configs file"

Let's say we log version of docker image as well:
if oom kill alert triggers on version N but wasn't triggered on version N-1
It appears OOM kill started shortly after the release of this new version, maybe the new code has a larger memory footprint.

To remediate this issue, we advise you to manually set the value of the memory you require to a generous value.

qui: I would rather update the linked runbook (y/check-oom-events, or a new link for PAASTA-16475) with this information, and keep the message here as short as possible - it's faster to update a runbook than the code.

I don't think the check itself can know which version is newer, and modify the message based on that, but I agree it would be useful to make it easy for service owners to correlate OOMs with version changes and add that tip to the runbook. I'll make a follow-up ticket for that.

Also, the possibility that a service deployment causes OOMs seems like a good reason to still send the alerts to service owners, even if the service is on autotuned_defaults. That way, during the push, the service owner can revert or increase memory as needed, instead of waiting for us to respond.

In general, I don't think we should disable this alerting until we really do respond to OOMs in real time. Until that point, service owners might want to take action based on the OOMs, and I think they should know about them.

PR #2786, opened 2020-04-17
Do not attempt to update HPA configs of a stateful set

jangsutsr:
Currently our HPA-related logic only applies to k8s deployments, but in the context of setup_kubernetes_job, a paasta service instance can also be a k8s stateful set. Apply HPA config update against a stateful set can lead to issues.

For example, `is_deployment_marked_paused` method assumes a deployment with name 'foo' already exists and attempts to fetch annotations of it. In the case of a stateful set, a deployment with the same name does not exist and k8s API server will return 404, which crashes setup_kubernetes_job run.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -47,7 +47,7 @@ from paasta_tools.bounce_lib import filter_tasks_in_smartstack
 from paasta_tools.bounce_lib import LockHeldException
 from paasta_tools.bounce_lib import LockTimeout
 from paasta_tools.bounce_lib import ZK_LOCK_CONNECT_TIMEOUT_S
-from paasta_tools.kubernetes_tools import get_annotations_for_deployment
+from paasta_tools.kubernetes_tools import get_annotations_for_kubernetes_service
 from paasta_tools.kubernetes_tools import KubeClient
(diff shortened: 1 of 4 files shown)


paasta_tools/setup_kubernetes_job.py
```diff
@@ -146,13 +147,17 @@ def setup_kube_deployments(
             elif app.kube_deployment not in existing_kube_deployments:
                 log.debug(f"Updating {app} because configs have changed.")
                 app.update(kube_client)
-            elif autoscaling_is_paused() and not is_deployment_marked_paused(
-                kube_client, app.soa_config
+            elif (
```
drmorr0, line 150: I'm a little worried that this approach is going to be too brittle.  What if we add another call to `is_deployment_marked_paused` somewhere else and forget the `isinstance` gate?  I guess my preference would be to add a `try...catch` block to `is_deployment_marked_paused` itself so that if we try to call it on something that's not a deployment we can fail gracefully.  This isn't a strong preference, but just my two cents.

As an aside, another reason I suggest this is because I think this block of if/else's is already pretty cumbersome, and I'm wary of making it more-so.

jangsutsr: I ended up renaming `get_annotations_for_deployment` to `get_annotations_for_kubernetes_service`, and takes both stateful sets and deployments into consideration in its implementation.

PR #2789, opened 2020-04-21
Never flood the bounce queue upon deployd restart

solarkennedy:
Since our queue is persistent now, I don't think we need to flood the bounce queue every time we restart deployd.


EvanKrall:
We have this code because we don't know what we missed in the mean time. Sure, if deployd was only down for a few seconds its' probably fine, but what if deployd is down for an hour?

Maybe a good compromise would be to have a command line tool that adds things to the queue?

solarkennedy:
Good point. 
I'm going to take my "read only" request a bit more seriously and discard this PR.
If someone else thinks this is a good idea or wants to make a new cli tool, please steal the patch.

PR #2791, opened 2020-04-22
Pass aws credentials to spark executors as environment variables

88manpreet:


solarkennedy:
@huadongliu can you review this one? Doesn't "--aws-credentials-yaml" do this for users already?

lydian:
> @huadongliu can you review this one? Doesn't "--aws-credentials-yaml" do this for users already?

We only do it for spark driver, the change is for spark executor. Normally spark executor doesn't need this env.  This env is only helpful when people trying to access AWS resources inside the executor.  For native spark script, we don't recommend people access AWS resource inside executor,  because it is higher chance to hit the AWS api limit.  However, it looks like this is the quick fix for mrjob harness script, so maybe go with this fix first, and if people do see the s3 limit error, we can discuss with the team for the workaround.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -441,6 +441,8 @@ def get_spark_config(
         "spark.executorEnv.PAASTA_INSTANCE": paasta_instance,
         "spark.executorEnv.PAASTA_CLUSTER": args.cluster,
         "spark.executorEnv.PAASTA_INSTANCE_TYPE": "spark",
+        "spark.executorEnv.AWS_SECRET_ACCESS_KEY": secret_key,
```
lydian, line 444: non_user_args will failed if people configure this.  I don't think we want to fail people's job if they want to use a different creds.  Would you mind add another check and only set the value if it was not set?

88manpreet: Good catch. Updated.

mattmb:

PR #2794, opened 2020-04-23
Bump version for service configuration lib to 2.4.3

nisan1101:


solarkennedy:
Please only bump requirements.txt and let a new release come afterwards by a paasta team member.

paasta is not (just) a python package, so it has other release consideraitons.

nisan1101:
reverted the paasta version

PR #2797, opened 2020-04-24
Fix InstanceStatusFlink name collision in go-swagger generated client

keymone:
InstanceStatus/flink property gets a struct named InstanceStatusFlink, which
collides with InstanceStatus/flink/status property definition.


poros:
Thanks for fixing the naming clash, but I have no recollection of how this API is used... 😕 
@Bronek do you remember, by any chance?

PR #2798, opened 2020-04-29
Dont flood deployd queue on start

stug:
Only flood the deployd queue on deployd startup if we aren't using the ZK queue.

Also added a simple script to add an arbitrary instance to the queue (I tested this on a real master and it works).

### DIFF
--- a/paasta_tools/contrib/add_to_deploy_queue.py
+++ b/paasta_tools/contrib/add_to_deploy_queue.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python
+# Copyright 2015-2020 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+import sys
+import time
+
+from kazoo.client import KazooClient
+from service_configuration_lib import DEFAULT_SOA_DIR
+
+from paasta_tools.deployd.common import ServiceInstance
+from paasta_tools.deployd.queue import ZKDelayDeadlineQueue
+from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.utils import NoConfigurationForServiceError
+from paasta_tools.utils import paasta_print
+from paasta_tools.utils import PaastaColors
+from paasta_tools.utils import validate_service_instance
+
+
+def parse_args(default_bounce_by_delay_secs):
+    parser = argparse.ArgumentParser(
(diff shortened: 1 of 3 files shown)


paasta_tools/deployd/master.py
```diff
@@ -126,7 +126,10 @@ def startup(self) -> None:
         self.log.info(
             "All watchers started, now adding all services for initial bounce"
         )
-        self.add_all_services()
+
+        # Fill the queue if we are not using the persistent ZK queue
+        if not self.config.get_deployd_use_zk_queue():
```
solarkennedy, line 131: What do you think about removing the feature toggle complexity soon?

paasta_tools/deployd/master.py
```diff
@@ -126,7 +126,10 @@ def startup(self) -> None:
         self.log.info(
             "All watchers started, now adding all services for initial bounce"
         )
-        self.add_all_services()
+
+        # Fill the queue if we are not using the persistent ZK queue
```
solarkennedy, line 130: I don't think you need this comment

paasta_tools/contrib/add_to_deploy_queue.py
```diff
@@ -0,0 +1,85 @@
...
+    except NoConfigurationForServiceError as e:
+        paasta_print(PaastaColors.red(str(e)))
+        sys.exit(1)
+
+    service_instance = ServiceInstance(
+        service=service,
+        instance=instance,
+        bounce_by=time.time() + args.bounce_by_delay_secs,
+        wait_until=time.time(),
+        watcher="manually_added",
```
solarkennedy, line 71: This is cool. You could go further and do something like "manually_added_by_$user" ?

paasta_tools/contrib/add_to_deploy_queue.py
```diff
@@ -0,0 +1,85 @@
...
+            soa_dir=DEFAULT_SOA_DIR,
+        )
+    except NoConfigurationForServiceError as e:
+        paasta_print(PaastaColors.red(str(e)))
+        sys.exit(1)
+
+    service_instance = ServiceInstance(
+        service=service,
+        instance=instance,
+        bounce_by=time.time() + args.bounce_by_delay_secs,
```
solarkennedy, line 69: I feel like if something is manually added, it shouldn't have a delay?

paasta_tools/deployd/master.py
stug: I guess I was figuring that deployd will go away when Mesos goes away and we'd leave it till then?  I don't have a strong opinion though.

paasta_tools/contrib/add_to_deploy_queue.py
stug: Yeah I can do that

paasta_tools/contrib/add_to_deploy_queue.py
stug: Yeah you are right actually.  I was thinking about the case where we needed to fully repopulate the queue, but that's probably not the most common case.

EvanKrall: Yeah maybe this could be a parameter. I'd probably default to later and make the user specify a parameter if they want it now.

PR #2799, opened 2020-04-29
Remove ServiceInstances from the deploy queue if they fail too many t…

stug:
…imes.

Now that we have a persistent queue, we don't want to keep failing entries in there indefinitely.

### DIFF
--- a/paasta_tools/deployd/workers.py
+++ b/paasta_tools/deployd/workers.py
@@ -48,6 +48,9 @@ class PaastaDeployWorker(PaastaThread):
         self.marathon_clients = marathon_tools.get_marathon_clients(
             self.marathon_servers
         )
+        self.max_failures = (
+            system_paasta_config.get_deployd_max_service_instance_failures()
+        )
 
     def setup_timers(self, service_instance: ServiceInstance) -> BounceTimers:
         bounce_length_timer = self.metrics.create_timer(
@@ -103,19 +106,27 @@ class PaastaDeployWorker(PaastaThread):
                         max_time=6000,
                     )
                 if bounce_again_in_seconds:
-                    bounce_by = int(time.time()) + bounce_again_in_seconds
-                    service_instance = ServiceInstance(
-                        service=service_instance.service,
-                        instance=service_instance.instance,
-                        bounce_by=bounce_by,
-                        wait_until=bounce_by,
-                        watcher=self.name,
-                        failures=failures,
-                        processed_count=service_instance.processed_count + 1,
-                        bounce_start_time=service_instance.bounce_start_time,
-                        enqueue_time=time.time(),
-                    )
-                    self.instances_to_bounce.put(service_instance)
+                    if failures >= self.max_failures:
+                        self.log.info(
+                            f"{self.name} Worker removing "
+                            f"{service_instance.service}.{service_instance.instance} "
+                            f"from queue because it has failed {failures} times "
+                            f"(max is {self.max_failures})"
+                        )
+                    else:
+                        bounce_by = int(time.time()) + bounce_again_in_seconds
+                        service_instance = ServiceInstance(
+                            service=service_instance.service,
+                            instance=service_instance.instance,
+                            bounce_by=bounce_by,
+                            wait_until=bounce_by,
+                            watcher=self.name,
+                            failures=failures,
+                            processed_count=service_instance.processed_count + 1,
+                            bounce_start_time=service_instance.bounce_start_time,
+                            enqueue_time=time.time(),
+                        )
+                        self.instances_to_bounce.put(service_instance)
             self.busy = False
             time.sleep(0.1)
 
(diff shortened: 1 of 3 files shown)


paasta_tools/utils.py
```diff
@@ -2086,6 +2086,14 @@ def get_deployd_startup_oracle_enabled(self) -> bool:
         """
         return self.config_dict.get("deployd_startup_oracle_enabled", True)
 
+    def get_deployd_max_service_instance_failures(self) -> int:
+        """Determines how many times a service instance entry in deployd's queue
+        can fail before it will be removed from the queue.
+
+        :returns: An integer
+        """
+        return self.config_dict.get("deployd_max_service_instance_failures", 100)
```
solarkennedy, line 2095: In practice I never saw anything this high, it as always like 20 for some reason

paasta_tools/deployd/workers.py
```diff
@@ -102,7 +105,7 @@ def run(self) -> None:
                         base=2,
                         max_time=6000,
                     )
-                if bounce_again_in_seconds:
+                if bounce_again_in_seconds and failures < self.max_failures:
```
solarkennedy, line 108: I think this would be more helpful like:
```
if max_fail:
  _log("Failed N times in a row trying to deploy this service. Refusing to continue.")
elfi max_attempts:
  _log("Tried to deploy this service N times. It hasn't failed but isn't making any progress. Backing off and enquing an hour later)
elsif bounce_again:
...
```

paasta_tools/deployd/workers.py
```diff
@@ -102,7 +105,7 @@ def run(self) -> None:
...
                     bounce_by = int(time.time()) + bounce_again_in_seconds
```
solarkennedy, line 109: I think this is where we could use some backoff on attempts

Maybe

```
bounce_by = time.time() + bounce_again_in_seconds * attempts ** 2
```
But eventually we'll be past the magical.. 600 seconds or whatever the drain timeout is?

Idk I think attempts and max_failures are kinda the same, both should be purged from the queue eventually otherwise they will continuosly make deployd spin and waste resources

stug: Yeah I think I would rather do it based on failures than attempts.  Did you actually see entries with lots of attempts and not many failures..?

paasta_tools/deployd/workers.py
stug: Yeah I can log more

paasta_tools/utils.py
stug: Sure, I'll drop it to that

paasta_tools/deployd/workers.py
solarkennedy: Yes. Services in a crash loop don't count as hard failures, they will just get tried to bounce every 60 seconds.

solarkennedy:
The "Attempts" backoff can be done in a different review if you want.

paasta_tools/deployd/workers.py
```diff
@@ -103,19 +109,34 @@ def run(self) -> None:
...
+                            f"{service_instance.service}.{service_instance.instance} "
+                            f"from queue because it has failed more than the "
+                            f"maximum of {self.max_failures} times."
+                        )
+                    elif service_instance.processed_count >= self.max_attempts:
+                        self.log.info(
+                            f"{self.name} Worker removing "
+                            f"{service_instance.service}.{service_instance.instance} "
+                            f"from queue because it has been attempted more than "
+                            f"the maximum of {self.max_attempts} times."
```
solarkennedy, line 124: I feel like these should have a more explicit "continue" so that code later down execute? Except I guess it does need to execute. idk.

paasta_tools/deployd/workers.py
```diff
@@ -103,19 +109,34 @@ def run(self) -> None:
...
-                        processed_count=service_instance.processed_count + 1,
-                        bounce_start_time=service_instance.bounce_start_time,
-                        enqueue_time=time.time(),
-                    )
-                    self.instances_to_bounce.put(service_instance)
+                    if failures >= self.max_failures:
+                        self.log.info(
+                            f"{self.name} Worker removing "
+                            f"{service_instance.service}.{service_instance.instance} "
+                            f"from queue because it has failed more than the "
```
solarkennedy, line 116: because it has failed {failures} times

paasta_tools/deployd/workers.py
```diff
@@ -103,19 +109,34 @@ def run(self) -> None:
...
+                            f"maximum of {self.max_failures} times."
+                        )
+                    elif service_instance.processed_count >= self.max_attempts:
+                        self.log.info(
+                            f"{self.name} Worker removing "
+                            f"{service_instance.service}.{service_instance.instance} "
```
solarkennedy, line 122: Attempted to be bounced {service_instance.processed_count} times ...

PR #2800, opened 2020-04-29
Ensure inbound_network_mode propagates smartstack iptables dependencies

y-trobinso:
Currently, firewall_update.py only checks for outbound_firewall rules when determining whether to include smartstack mapping information (consumed downstream by iptables rules).

This change ensures that inbound_network_mode is also checked to ensure this information is made available.


y-trobinso:
> Can you add unit test coverage in `paasta/tests/test_firewall_update.py`?

Added.

PR #2802, opened 2020-05-05
Extend Kubernetes Custom Resources with general yelpsoa-configs

flavray:
This commit extends the configuration/custom resources submitted to Kubernetes for services deployed there (this is targetting resources managed by operators).

Re-using the logic used in [marathon_tools.py](https://github.com/Yelp/paasta/blob/master/paasta_tools/marathon_tools.py#L349), reading the "general" service configuration (`monitoring.yaml`, `smartstack.yaml`, etc...) and merging it with every instance configuration found in `kubernetes-{cluster}` files.

Shipping this change will cause the `paasta.yelp.com/config_sha` label to be updated even if the CRs themselves will not change. Not sure whether that is an issue for existing operators?

-----

* `make test` passes
* Tested on a local run of the `kafka-operator` (`setup_kubernetes_cr`)
  * New CR spec [here](https://fluffy.yelpcorp.com/i/DklZT5WCGCgqpbK2Rsw6pxCjKQbBFXmT.html#L48-90)
  * The CR specs do not change after applying with the "extended" resource (extra fields are ignored by the operator as they are not in the CRD). ([Before](https://fluffy.yelpcorp.com/i/rbMXjw5hc39W6Qpzk26KbNNkNCggkX6n.html) – [After](https://fluffy.yelpcorp.com/i/T0L8FKSdKF8l9KhVBH99btP5wLNDrm35.html))

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -73,6 +73,7 @@ from docker import Client
 from docker.utils import kwargs_from_env
 from kazoo.client import KazooClient
 from mypy_extensions import TypedDict
+from service_configuration_lib import read_extra_service_information
 from service_configuration_lib import read_service_configuration
 
 import paasta_tools.cli.fsm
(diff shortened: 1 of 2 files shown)


vkhromov:
Updating `paasta.yelp.com/config_sha` means that operators will restart pods and other resources they manage, which probably isn't good for some of them.

poros:
flink-operator should ignore `paasta.yelp.com/config_sha` by now, so this should not lead to any cluster restart afaik. Given that without a change in the CRD the spec won't even be updated, this should not be a problem either. (cc @Bronek in case I am grossly mistaken)

Out of curiosity, what is this change for? Are we preparing to add any of the other configurations into paasta-tools-go?

poros:
@Rob-Johnson I did not mean to re-request your review, I misclicked and I can't undo, sorry :/

Bronek:
@flavray , do you mind rebasing from master?

mattmb:

PR #2803, opened 2020-05-05
make pool constraint 'EQUALS' instead of 'LIKE'

drmorr0:
(COMPINFRA-333) We have a "stable" pool and a "stable_batch" pool, and
jobs that are supposed to run on the "stable" pool are getting scheduled
on the "stable_batch" pool because the constraint is too open.  This
change should make it so that stuff only gets scheduled on the pool they
specify.

Note that this will cause a big bounce, and will affect all Marathon and
Tron jobs, even though the issue we saw was only for Tron.

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -835,7 +835,7 @@ class InstanceConfig:
 
     def get_pool_constraints(self) -> List[Constraint]:
         pool = self.get_pool()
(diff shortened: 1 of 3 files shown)


drmorr0:
Marathon doesn't support EQUALS.  So we're going to fix this in task_processing instead.

PR #2808, opened 2020-05-13
Refactor inbound_firewall to not depend on smartstack dependencies

y-trobinso:
Previously, this was placed at the end of `_smartstack_rules`, placing inbound filters after rules established to allow smartstack traffic. When envoy is used, this doesn't make sense. It also doesn't make sense to require a dependencies file for this filtering behavior, either.

As such, this has been moved upwards one layer and refactored accordingly. I've also included an allow for traffic from localhost, should other rules not propagate it to this location.

Including Krall as CC, due to some chatter about healthcheck traffic today. This is an opt-in feature, so this should only impact services that include it.

### DIFF
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -218,7 +218,6 @@ def _inbound_traffic_rule(conf, service_name, instance_name, protocol="tcp"):
             )
 
 
-
 def _smartstack_rules(conf, soa_dir, synapse_service_dir):
     for dep in conf.get_dependencies() or ():
         namespace = dep.get("smartstack")
--- a/paasta_tools/firewall.py
+++ b/paasta_tools/firewall.py
@@ -77,13 +77,19 @@ class ServiceGroup(collections.namedtuple("ServiceGroup", ("service", "instance"
             # for several minutes after the directory disappears from soa-configs.
             return ()
 
-        if not conf.get_outbound_firewall():
+        if not conf.get_inbound_firewall() and not conf.get_outbound_firewall():
             return ()
 
-        rules = list(_default_rules(conf, self.log_prefix))
-        rules.extend(_well_known_rules(conf))
-        rules.extend(_smartstack_rules(conf, soa_dir, synapse_service_dir))
-        rules.extend(_cidr_rules(conf))
+        rules = list()
+
+        if conf.get_inbound_firewall():
+            rules.extend(_inbound_traffic_rule(conf, self.service, self.instance))
+
+        if conf.get_outbound_firewall():
+            rules.extend(_default_rules(conf, self.log_prefix))
+            rules.extend(_well_known_rules(conf))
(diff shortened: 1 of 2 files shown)


paasta_tools/firewall.py
```diff
@@ -182,19 +188,29 @@ def _yocalhost_rule(port, comment, protocol="tcp"):
...
+        port = conf.get_proxy_port(service_name, instance_name)
+        yield iptables.Rule(
+            protocol=protocol,
+            src="0.0.0.0/0.0.0.0",
+            dst="0.0.0.0/0.0.0.0",
+            target="REJECT",
+            matches=((protocol, (("dport", (str(port),)),)),),
+            target_parameters=((("reject-with", ("icmp-port-unreachable",))),),
+        )
+        yield iptables.Rule(
```
danielpops, line 206: do you need moar local IPs like 169.254.255.254 169.254.0.1 etc?

y-trobinso: That depends on whether we want to continue to support smartstack traffic (re, our conversation about using this with envoy). If so, we'll also want those two IPs.

I guess they can't hurt regardless. I'll add them.

paasta_tools/firewall.py
```diff
@@ -77,13 +81,19 @@ def get_rules(self, soa_dir, synapse_service_dir):
             # for several minutes after the directory disappears from soa-configs.
             return ()
 
-        if not conf.get_outbound_firewall():
+        if not conf.get_inbound_firewall() and not conf.get_outbound_firewall():
             return ()
```
kawaiwanyelp, line 85: nit: Are these lines necessary anymore? You're already checking for `get_inbound_firewall` and `get_outbound_firewall` below.

paasta_tools/firewall.py
```diff
@@ -182,19 +192,30 @@ def _yocalhost_rule(port, comment, protocol="tcp"):
...
+    If this is set to "reject", this is limited only to traffic from localhost"""
+    policy = conf.get_inbound_firewall()
+    if policy == "reject":
+        port = conf.get_proxy_port(service_name, instance_name)
+        yield iptables.Rule(
+            protocol=protocol,
+            src="0.0.0.0/0.0.0.0",
+            dst="0.0.0.0/0.0.0.0",
+            target="REJECT",
+            matches=((protocol, (("dport", (str(port),)),)),),
```
kawaiwanyelp, line 207: Not a review comment. Just an "oof"; why do these parentheses and commas remind me of Lisp...

paasta_tools/firewall.py
y-trobinso: Since `tuple(list())` should indeed be equivalent, removed. :)

y-trobinso:
I confirmed that the new Travis failures occurring in this branch were from upstream changes. Closed.

PR #2809, opened 2020-05-14
Add tween to log request info to CLog

kawaiwanyelp:
### Description
This PR adds the a tween to the PaaSTA API to make it log response times, status codes, and exception information for requests made to clog (if available). We aren't really logging this elsewhere; once we get this into scribe, we can start processing it to see how our API is performing.

### Testing
make test
Ran the API locally against the kubestage cluster with the tween set up. Had it print to the console instead of actually logging to clog.

### DIFF
--- a/paasta_tools/api/tweens/request_logger.py
+++ b/paasta_tools/api/tweens/request_logger.py
@@ -15,13 +15,11 @@
 Creates a tween that logs information about requests.
 """
 import json
-import logging
-import time
 import traceback
+from datetime import datetime
 
 import pyramid
 
-from paasta_tools import utils
 from paasta_tools.api import settings
 
 try:
@@ -52,15 +50,18 @@ class request_logger_tween_factory:
         )
 
     def _log(
-        self, endpoint, level=logging.INFO, additional_fields=None,
+        self, timestamp=None, level="INFO", additional_fields=None,
     ):
         if clog is not None:
+            # `settings` values are set by paasta_tools.api.api:setup_paasta_api
+            if timestamp is None:
+                timestamp = datetime.now()
             dct = {
-                "timestamp": time.time(),
-                "hostname": utils.get_hostname(),
-                "level": logging.getLevelName(level),
+                "human_timestamp": timestamp.strftime("%Y-%m-%dT%H:%M:%S"),
+                "unix_timestamp": timestamp.timestamp(),
+                "hostname": settings.hostname,
+                "level": level,
                 "cluster": settings.cluster,
-                "endpoint": endpoint,
             }
             if additional_fields is not None:
                 dct.update(additional_fields)
@@ -68,28 +69,28 @@ class request_logger_tween_factory:
             clog.log_line(self.log_name, line)
 
     def __call__(self, request):
-        start_time = time.time()  # start clock for response time
+        start_time = datetime.now()  # start clock for response time
         request_fields = {
             "path": request.path,
             "params": request.params.mixed(),
         }
         response_fields = {}
-        log_level = logging.INFO
+        log_level = "INFO"
 
         try:
             response = self.handler(request)
 
             response_fields["status_code"] = response.status_int
             if 300 <= response.status_int < 400:
-                log_level = logging.WARNING
+                log_level = "WARNING"
             elif 400 <= response.status_int < 600:
-                log_level = logging.ERROR
+                log_level = "ERROR"
                 response_fields["body"] = response.body.decode("utf-8")
 
             return response
 
         except Exception as e:
-            log_level = logging.ERROR
+            log_level = "ERROR"
             response_fields.update(
                 {
                     "status_code": 500,
@@ -100,11 +101,11 @@ class request_logger_tween_factory:
             raise
 
         finally:
-            response_time_ms = (time.time() - start_time) * 1000
+            response_time_ms = (datetime.now() - start_time).total_seconds() * 1000
             response_fields["response_time_ms"] = response_time_ms
 
             self._log(
-                endpoint=request.matched_route.name if request.matched_route else None,
+                timestamp=start_time,
                 level=log_level,
                 additional_fields={
                     "request": request_fields,
--- a/paasta_tools/api/tweens/request_logger.py
+++ b/paasta_tools/api/tweens/request_logger.py
@@ -1,0 +1,113 @@
+# Copyright 2015-2016 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+Creates a tween that logs information about requests.
+"""
+import json
+import logging
+import time
+import traceback
+
+import pyramid
+
+from paasta_tools import utils
+from paasta_tools.api import settings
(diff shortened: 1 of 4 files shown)


paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,113 @@
...
+        except Exception as e:
+            log_level = logging.ERROR
+            response_fields.update(
+                {
+                    "status_code": 500,
+                    "exc_type": type(e).__name__,
+                    "exc_info": traceback.format_exc(),
+                }
+            )
+            raise
```
keymone, line 100: why `raise` instead of `return` here if we've updated response fields?

paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,113 @@
...
+        )
+
+    def _log(
+        self, endpoint, level=logging.INFO, additional_fields=None,
+    ):
+        if clog is not None:
+            dct = {
+                "timestamp": time.time(),
+                "hostname": utils.get_hostname(),
+                "level": logging.getLevelName(level),
```
keymone, line 61: let's not do potentially expensive calculations every time we log a request. it's not like hostname or loglevel is going to chnage much during server runtime.

paasta_tools/api/tweens/request_logger.py
keymone: disregard that, i misread what response_fields are

paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,113 @@
...
+        self.log_name = registry.settings.get(
+            "request_log_name", DEFAULT_REQUEST_LOG_NAME,
+        )
+
+    def _log(
+        self, endpoint, level=logging.INFO, additional_fields=None,
+    ):
+        if clog is not None:
+            dct = {
+                "timestamp": time.time(),
```
Rob-Johnson, line 59: can you use the `start_time` field that you defined in `_call` here instead?

paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,113 @@
...
+                "hostname": utils.get_hostname(),
+                "level": logging.getLevelName(level),
+                "cluster": settings.cluster,
+                "endpoint": endpoint,
```
Rob-Johnson, line 63: why not just put this in the `request` body?

paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,113 @@
...
+            }
+            if additional_fields is not None:
+                dct.update(additional_fields)
+            line = json.dumps(dct, sort_keys=True)
+            clog.log_line(self.log_name, line)
+
+    def __call__(self, request):
+        start_time = time.time()  # start clock for response time
+        request_fields = {
```
Rob-Johnson, line 72: can you also put client_addr, user_agent, request headers, http method in here?

paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,113 @@
...
+                }
+            )
+            raise
+
+        finally:
+            response_time_ms = (time.time() - start_time) * 1000
+            response_fields["response_time_ms"] = response_time_ms
+
+            self._log(
+                endpoint=request.matched_route.name if request.matched_route else None,
```
Rob-Johnson, line 107: if we do this then we won't get the url for 404s right? don't we just want the 'raw' path? `request.path` maybe?

paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def _log(
+        self, endpoint, level=logging.INFO, additional_fields=None,
+    ):
+        if clog is not None:
+            dct = {
+                "timestamp": time.time(),
+                "hostname": utils.get_hostname(),
+                "level": logging.getLevelName(level),
+                "cluster": settings.cluster,
```
qui, line 62: Where does this get added?

kawaiwanyelp: I added a comment in the latest version of this PR explaining where, but [here](https://github.com/Yelp/paasta/blob/master/paasta_tools/api/api.py#L169) is where.

paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,113 @@
...
+from paasta_tools import utils
+from paasta_tools.api import settings
+
+try:
+    import clog
+except ImportError:
+    clog = None
+
+
+DEFAULT_REQUEST_LOG_NAME = "tmp_paasta_requests"
```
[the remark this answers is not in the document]
kawaiwanyelp: Done.

paasta_tools/api/tweens/request_logger.py
kawaiwanyelp: I just removed this since I was already recording `request.path` in the request dict.

paasta_tools/api/tweens/request_logger.py
kawaiwanyelp: Added all except `user_agent`, which is already in the headers.

paasta_tools/api/tweens/request_logger.py
kawaiwanyelp: I put removed the endpoint argument, but the request dict still includes the `request.path`.

paasta_tools/api/tweens/request_logger.py
kawaiwanyelp: Done.

paasta_tools/api/tweens/request_logger.py
kawaiwanyelp: I moved the `hostname` to api settings, so that it gets set once when the API starts up. I also removed the extra level processing, and now the level is passed in as a string directly.

paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,117 @@
...
+                {
+                    "status_code": 500,
+                    "exc_type": type(e).__name__,
+                    "exc_info": traceback.format_exc(),
+                }
+            )
+            raise
+
+        finally:
+            response_time_ms = (datetime.now() - start_time).total_seconds() * 1000
```
qui, line 107: Why use datetime and then convert to seconds later? Doesn't time.time() work?
Also, this is minor, but why not just keep it as seconds instead of multiplying by 1000? We can always do unit conversions while analyzing the data.

kawaiwanyelp: I switched to datetime to make it easier to add a human timestamp to the log lines. As for why not keep it as seconds, I kept it in line with most response time conventions and recorded it in milliseconds.

paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,117 @@
...
+
+    def _log(
+        self, timestamp=None, level="INFO", additional_fields=None,
+    ):
+        if clog is not None:
+            # `settings` values are set by paasta_tools.api.api:setup_paasta_api
+            if timestamp is None:
+                timestamp = datetime.now()
+            dct = {
+                "human_timestamp": timestamp.strftime("%Y-%m-%dT%H:%M:%S"),
```
Rob-Johnson, line 60: can you include the timezone in this too?

paasta_tools/api/tweens/request_logger.py
```diff
@@ -0,0 +1,117 @@
...
+        self.log_name = registry.settings.get(
+            "request_log_name", DEFAULT_REQUEST_LOG_NAME,
+        )
+
+    def _log(
+        self, timestamp=None, level="INFO", additional_fields=None,
+    ):
+        if clog is not None:
+            # `settings` values are set by paasta_tools.api.api:setup_paasta_api
+            if timestamp is None:
```
Rob-Johnson, line 57: `if not timestamp`

PR #2812, opened 2020-05-15
[wip] Add cluster blacklist to autotune rightsizer

kawaiwanyelp:
### Description
I'm adding an argument to prevent the rightsizer from working on specific clusters. 

### Testing
<still testing>

### DIFF
--- a/paasta_tools/contrib/rightsizer_soaconfigs_update.py
+++ b/paasta_tools/contrib/rightsizer_soaconfigs_update.py
@@ -74,6 +74,14 @@ def parse_args():
         dest="source_id",
     )
     parser.add_argument(
+        "--cluster-blacklist",
+        help="Space-separated list of clusters to not rightsize",
+        required=False,
+        nargs="+",
+        default=[],
+        dest="cluster_blacklist",
+    )
+    parser.add_argument(
         "-v",
         "--verbose",
         help="Logging verbosity",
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/rightsizer_soaconfigs_update.py
```diff
@@ -73,6 +73,14 @@ def parse_args():
         default=None,
         dest="source_id",
     )
+    parser.add_argument(
+        "--cluster-blacklist",
+        help="Space-separated list of clusters to not rightsize",
+        required=False,
+        nargs="+",
```
qui, line 80: Does the script work if you don't pass --cluster-blacklist? Just reading the docs on + vs * and want to double-check.

paasta_tools/contrib/rightsizer_soaconfigs_update.py
```diff
@@ -121,7 +133,9 @@ def main(args):
     extra_message = get_extra_message(report["search"])
     config_source = args.source_id or args.csv_report
 
-    results = get_recommendations_by_service_file(report["results"])
+    results = get_recommendations_by_service_file(
```
qui, line 136: I think it would be a more logical code structure if you did the filtering separately, before the get_recommendations_by_service_file transform.

kawaiwanyelp:
Going to pause on doing this since we don't need it for now.

PR #2814, opened 2020-05-18
Optimize generate_deployments_for_service, taking it from ~3 minutes …

EvanKrall:
…on yelp-main to ~20 seconds (most of which is waiting for git)

PAASTA-16665

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2722,7 +2722,7 @@ def read_service_instance_names(
 ) -> Collection[Tuple[str, str]]:
     instance_list = []
     conf_file = f"{instance_type}-{cluster}"
-    config = service_configuration_lib.read_extra_service_information(
+    config = __read_extra_service_information_cached(
         service, conf_file, soa_dir=soa_dir
     )
     config = filter_templates_from_config(config)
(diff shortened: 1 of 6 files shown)


paasta_tools/utils.py
```diff
@@ -2852,6 +2852,12 @@ def load_service_instance_configs(
     return merged
 
 
+# be extremely sure you do not modify the return value of this.
```
qui, line 2855: Hm, how much faster does this make things? service_configuration_lib already does some YAML caching:  https://github.com/Yelp/service_configuration_lib/blob/master/service_configuration_lib/__init__.py#L90

But it always deep-copies, which makes it a little slower. Just wondering if the speed is worth having to worry about accidental mutations here.

EvanKrall: Yeah, the deep copy is the second-slowest part. In my pyflame of stock `generate_deployments_for_service`, copy.deepcopy takes 15.5% of time, so roughly 25-30s. (`get_deploy_group_mappings` is the bulk of the rest of it)

keymone: seems safe to do pickle+unpickle? should be faster iirc

EvanKrall: really? it would surprise me if serializing/deserializing is faster than deepcopy.

PR #2816, opened 2020-05-20
Log service instance update list

jangsutsr:
This PR also changes the default log level from warning to info for more
verbosity.

### DIFF
--- a/paasta_tools/setup_kubernetes_job.py
+++ b/paasta_tools/setup_kubernetes_job.py
@@ -21,6 +21,7 @@ Command line options:
 - -v, --verbose: Verbose output
 """
 import argparse
+import json
 import logging
 import sys
 from typing import Optional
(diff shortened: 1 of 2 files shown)


paasta_tools/setup_kubernetes_job.py
```diff
@@ -135,8 +136,10 @@ def setup_kube_deployments(
         for service_instance in service_instances_with_valid_names
     ]
 
+    apps_updated = []
     for _, app in applications:
         if app:
+            apps_updated.append(str(app))
             if (
                 app.kube_deployment.service,
                 app.kube_deployment.instance,
```
drmorr0, line 145: Can we make all the "debug" messages (except for the "app is up to date" message) `info`-level instead of `debug`-level?  That will be really helpful for understanding why changes are getting made.

jangsutsr:
The unit tests should have shown the appearance of the added log output.

PR #2818, opened 2020-05-27
Notify sysgit deploymentsd watcher to create a new deployments.json o…

stug:
…n paasta start/stop/restart

Previously this was handled by a gitolite hook, but that went away when services moved off gitolite. This is the same approach used by mark-for-deployment; we just missed that we needed it on start/stop/restart as well.

See PAASTA-16512

---

I ran this locally and it triggered a restart of compute-infra-test-service essentially immediately.

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -28,6 +28,7 @@ from paasta_tools.api.client import get_paasta_api_client
 from paasta_tools.cli.cmds.status import add_instance_filter_arguments
 from paasta_tools.cli.cmds.status import apply_args_filters
 from paasta_tools.cli.utils import get_instance_config
+from paasta_tools.cli.utils import trigger_deploys
 from paasta_tools.flink_tools import FlinkDeploymentConfig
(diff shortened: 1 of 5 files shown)


paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -126,9 +127,10 @@ def issue_state_change_for_service(service_config, force_bounce, desired_state):
         force_bounce=force_bounce,
         desired_state=desired_state,
     )
-    remote_git.create_remote_refs(
-        utils.get_git_url(service_config.get_service()), ref_mutator
```
dbgrigsby, line 130: Is it enough for this to just be called on stop/stop/restart? We often make use of paasta mark-for-deployment alone in Jenkinsfiles -- is paasta restart always called by paasta itself when a sha is marked for deployment?

This PR might break `def test_mark_for_deployment_yelpy_repo`

stug: It is still called from mark-for-deployment -- it is just not defined in that module anymore (I think this is what you are asking?)

dbgrigsby: Ah I misread the diff, looks good to me then! Ship!

PR #2820, opened 2020-05-27
Set pod's dns policy to Default to use node's dns resolver

keymone:
Internal ticket: PAASTA-16641. This is going to cause a big bounce.


Rob-Johnson:
now that we've updated coredns on the that should be doable, but yeah, please test first

keymone:
this was reverted until we figure out coredns config

PR #2821, opened 2020-05-27
Speed up generate deployments for service again

EvanKrall:
Redo of https://github.com/Yelp/paasta/pull/2814 in which I forgot to check whether the tests passed before merging.

Now that CID has moved the yelp-main git repo over, we're actually down to ~5 seconds.


EvanKrall:
Looks like that failing itest was a flake; hopefully this branch didn't make it flaky.

PR #2822, opened 2020-05-28
PAASTA-16499 `paasta validate` now checks if vault secrets are defined

gvolco:
This change adds a new test to `paasta validate` so that operators will know if any of their service instances reference a secret that doesn't exist.

The reason for this change is that mesos would allow such situations, but kubernetes would fail to bring up the service.
So an early check will help to avoid these situations.

More info on PAASTA-16499

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -32,12 +32,16 @@ from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import PaastaColors
 from paasta_tools.cli.utils import success
 from paasta_tools.kubernetes_tools import sanitise_kubernetes_name
+from paasta_tools.secret_tools import get_secret_name_from_ref
+from paasta_tools.secret_tools import is_secret_ref
+from paasta_tools.secret_tools import is_shared_secret
 from paasta_tools.tron_tools import list_tron_clusters
 from paasta_tools.tron_tools import validate_complete_config
 from paasta_tools.utils import get_service_instance_list
 from paasta_tools.utils import list_all_instances_for_service
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import list_services
+from paasta_tools.utils import load_system_paasta_config
 
 
 SCHEMA_VALID = success("Successfully validated schema")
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -417,6 +421,58 @@ def validate_autoscaling_configs(service_path):
     return returncode
 
 
+def validate_secrets(service_path):
+    soa_dir, service = path_to_soa_dir_service(service_path)
+    system_paasta_config = load_system_paasta_config()
+    return_value = True
+    for cluster in list_clusters(service, soa_dir):
+        vault_env = system_paasta_config.get_vault_cluster_config().get(cluster)
```
keymone, line 429: no need to call `system_paasta_config.get_vault_cluster_config()` in every iteration

paasta_tools/cli/cmds/validate.py
```diff
@@ -417,6 +421,58 @@ def validate_autoscaling_configs(service_path):
...
+        ):
+            instance_config = get_instance_config(
+                service=service,
+                instance=instance,
+                cluster=cluster,
+                load_deployments=False,
+                soa_dir=soa_dir,
+            )
+            for env_value in instance_config.config_dict.get("env", {}).values():
+                if is_secret_ref(env_value):
```
keymone, line 446: probably worth taking this block out into separate function - easier to understand what's going on and easier to test

paasta_tools/cli/cmds/validate.py
```diff
@@ -417,6 +421,58 @@ def validate_autoscaling_configs(service_path):
...
+                                )
+                            )
+                            return_value = False
+                    else:
+                        paasta_print(
+                            failure(f"Secret file {secret_file_name} not defined", "")
+                        )
+                        return_value = False
+    if return_value:
+        paasta_print(success("All secrets are present"))
```
keymone, line 472: lets not print this message for services than don't have any secrets defined

paasta_tools/cli/cmds/validate.py
gvolco: you're right, will move this call outside the for loop.

paasta_tools/cli/cmds/validate.py
gvolco: sounds good. I'm going to extract some logic into a helper function.

paasta_tools/cli/cmds/validate.py
gvolco: I think I'm going to leave this, but rephrasing the message. The extra logic needed to print based on the existence of secrets would make the function a bit more convoluted.

Issue #1461, opened 2017-08-15
Why do we have our own print function?


EvanKrall:
Seriously, what is `paasta_print` for, other than causing [bugs](https://github.com/Yelp/paasta/pull/1383)?

I don't remember. @mjksmith added it back in his internship, I assume for a good reason, but I think we've lost the theory of it.

keymone:
+1, lets either drop or document reasons for having it.

asottile:
For one, python 3 **does not flush when printing by default** (perhaps ironically, the issue/patch you linked to is actually a strong reason *to* have a wrapper, though it wasn't the original intention).

The main reasons for introducing it were:
- python2 / python3 compatibility
- printing text data when `LANG` is not set or is `LANG=C`
    - by default, python attempts to *encode* output using the "automatically detected locale"
    - In many of the cases where paasta runs (cron, marathon, docker, etc.) these environment variables aren't present
    - paasta produces textual output that is not ASCII encodable
- printing bytes is awkward in python3, there are cases where paasta has output from logs / etc. where it should be written unmodified / without decoding to output

I've written [slides about demystifying io in python 3](https://anthonywritescode.github.io/episodes-py3/04-file-io/) and plan to record a video when I get some time :)

vkhromov:
+1 to get rid of it.

1. paasta is Python 3 only now, we don't need compatibility with Python 2.
2. We should never ever need to `print()` raw bytes, right?
3. It doesn't seem that the paasta code itself contains any non-ASCII characters, at least `find paasta_tools -name '*.py' | xargs grep --color='auto' -P -n "[\x80-\xFF]"` found nothing suspicious.

As for enforcing flushes and fixing `UnicodeEncodeError: 'ascii' codec can't encode character ...` errors, can we do something with `print()` function itself?

asottile:
paasta itself does not have non-ascii literals, but it runs applications which produce arbitrary output

in python3.7+ the default changes from ascii to utf-8 so you might have better mileage there

the one thing to consider is you'll want `print(..., flush=True)` in most cases which might make using it a bit unfortunate

vkhromov:
Migrating to python3.7+ indeed should help with printing unicode characters for undefined `LANG` or `LANG=C`:
```
$ cat enc.py
import sys
print(sys.version)
print(chr(246), chr(9786), chr(9787))

$ python3.6 enc.py; python3.7 enc.py
[...]

PR #2826, opened 2020-05-31
Remove paasta_print

keymone:
Python 2 and the mirgation nightmares are in the past, so let's finally get rid of this.


keymone:
previous discussion: #1461, i think we should ship and deal with fallout.

vkhromov:
From the discussion, it seems that we better migrate to python 3.7 first...

keymone:
see #2837, #2838 and #2839

PR #2827, opened 2020-06-01
Fix request logger clog setup

kawaiwanyelp:
### Description
I initially tried to release request logging for the PaaSTA API but it caused the API to crash when it tried to log since clog was not properly setup. I've added a function `setup_clog` that runs before the pyramid Configurator is set up, which resolves the issue.

### Testing
Ran paasta-api locally with clog enabled (I did not have it enabled before, so I didn't catch the issue)

### DIFF
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -30,6 +30,7 @@ import paasta_tools.api
 from paasta_tools import kubernetes_tools
 from paasta_tools import marathon_tools
 from paasta_tools.api import settings
+from paasta_tools.api.tweens import request_logger
 from paasta_tools.utils import load_system_paasta_config
 
 
@@ -83,6 +84,7 @@ def make_app(global_config=None):
     )
 
     config.include("pyramid_swagger")
+    config.include(request_logger)
     config.add_route("resources.utilization", "/v1/resources/utilization")
     config.add_route(
         "service.instance.status", "/v1/services/{service}/{instance}/status"
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -23,6 +23,7 @@ import sys
 import manhole
 import requests_cache
 import service_configuration_lib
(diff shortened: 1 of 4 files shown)


paasta_tools/api/api.py
```diff
@@ -199,6 +205,16 @@ def setup_paasta_api():
     requests_cache.install_cache("paasta-api", backend="memory", expire_after=5)
 
 
+def setup_clog(scribe_host="169.254.255.254", scribe_port=1463):
+    if clog:
+        clog.config.configure(
+            scribe_host=scribe_host,
+            scribe_port=scribe_port,
+            monk_disable=False,
```
Rob-Johnson, line 213: I suspect we only want one of these disabled, but I'm not sure what the usage is supposed to be (both are disabled by default). @flavray can you weigh in here?

paasta_tools/api/api.py
```diff
@@ -199,6 +205,16 @@ def setup_paasta_api():
     requests_cache.install_cache("paasta-api", backend="memory", expire_after=5)
 
 
+def setup_clog(scribe_host="169.254.255.254", scribe_port=1463):
+    if clog:
+        clog.config.configure(
```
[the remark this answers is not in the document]
kawaiwanyelp: I decided to load the config from `/nail/srv/configs/clog.yaml`, which is what most services use.

paasta_tools/api/api.py
kawaiwanyelp: These are bare bones configs from `/nail/srv/configs/clog.yaml` so I think they make sense at defaults. These settings are also used in the oom_logger.

flavray: It's fine to have them both enabled, only the newer versions of yelp_clog allow scribe to be disabled.
Would you mind also adding `preferred_backend="monk"`, otherwise it'll log things to scribe instead of Monk.

PR #2832, opened 2020-06-03
U/tingyenl/coreml 1401 get a meaningful spark app name

lydian:
clean up for  https://github.com/Yelp/paasta/pull/2730

make test pass locally 
also test running with adhoc spark batch seems to work, and have the correct app name being configured: http://history.spark.paasta-pnw-devc.yelp/history/fc21f809-33fc-48c7-a034-69c195279885-19660/jobs/ (The filename I tested is s3.py, and correctly set the app name to be paasta_s3_tingyenl_53434 )

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -602,6 +602,21 @@ def run_docker_container(
     return 0
 
 
+def get_spark_app_name(original_docker_cmd):
+    # Use submitted batch name as default spark_run job name
+    if "spark-submit" in original_docker_cmd:
+        parser = argparse.ArgumentParser()
+        parser.add_argument("cmd", choices=["spark-submit"])
+        parser.add_argument("others", nargs="+")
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -602,6 +613,37 @@ def run_docker_container(
...
+    # Use submitted batch name as default spark_run job name
+    spark_app_name = "paasta_spark_run"
+
+    docker_cmds = (
+        shlex.split(original_docker_cmd)
+        if isinstance(original_docker_cmd, str)
+        else original_docker_cmd
+    )
+
+    if "spark-submit" in original_docker_cmd:
```
huadongliu, line 628: Looks like commands like "python s3.py" also benefit from the change. Not many such Spark jobs though.

lydian: yup, that would work, too!

PR #2841, opened 2020-06-09
Fix setup_kubernetes_cr for template entries

keymone:
Skip template entries after loading


paasta_tools/setup_kubernetes_cr.py
```diff
@@ -170,6 +170,12 @@ def setup_all_custom_resources(
         config_dicts = load_all_configs(
             cluster=cluster, file_prefix=crd.file_prefix, soa_dir=soa_dir
         )
+        for svc, config in config_dicts.items():
+            for instance in config:
+                if instance[0] == "_":
```
gvolco, line 175: nit: we could add a comment explaining why we want to remove this entry

PR #2843, opened 2020-06-10
Refactor inbound firewall rules to mark traffic in PREFILTER, then act during later chains

y-trobinso:
After debugging my previous inbound filter changes, it was clear that getting traffic passed through to the `PAASTA` service chain was going to be cumbersome and buggy after the traffic was already routed to Docker.

So I tried a new strategy:
```
sudo iptables -t nat -I DOCKER 1 -p tcp ! -s 127.0.0.1 --dport $DPORT -j MARK --set-mark $DPORT
sudo iptables -I DOCKER-USER 1 -m mark --mark $DPORT -j REJECT
```
And this worked much better.

This change intercepts inbound traffic very early -- specifically, in the nat `PREFILTER` chain of iptables -- and marks it for later operations (in this change, log or reject). This has the benefit of identifying the source of traffic before it gets to the `MASQUERADE` actions in prefiltering. This in turn makes it substantially easier to identify the source of traffic _and_ to tidy up these rules in conveniently named `PAASTA` chains.

This change is long because, to do this, I needed to push the table definition down through paasta_tools/iptables.py. I also added two new chains (`PAASTA-PREFILTER` and `PAASTA-MARK`) to keep the new rules legible and easy to find in manual debugging.

Tests will be added before this ships. This should be considered a WIP until they've been added as noted in the comments below.

### DIFF
--- a/paasta_tools/cli/schemas/marathon_schema.json
+++ b/paasta_tools/cli/schemas/marathon_schema.json
@@ -354,6 +354,7 @@
                         "inbound_firewall": {
                             "enum": [
                                 "accept",
+                                "monitor",
                                 "reject"
                             ]
                         },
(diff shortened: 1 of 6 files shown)


y-trobinso:
Closing this to revisit later.

In particular, this will need another refactor of iptables rules before it can safely ship.

PR #2844, opened 2020-06-11
Optimize deployd workers

EvanKrall:
These changes drastically decrease the amount of CPU time needed by deployd's workers / setup_marathon_job.

The first commit, 546cc64, makes it so we only request information about apps that matter from Marathon.

The second and third commits, 30b7371 and 700784c, aim to reduce the number of times `load_system_paasta_config` is called by passing around SystemPaastaConfig objects to almost everything that needs it that ends up being called by deployd. (The one exception remaining in my pyflame charts is https://github.com/Yelp/paasta/blob/766c25dcf3bba640e51a4e9ba8091f63910b79bf/paasta_tools/utils.py#L2915 -- removing this would give a further 20% CPU time improvement). It turns out that even though `load_system_paasta_configs` tries very hard not to re-parse YAML, the code that determines whether the files have changed takes a nontrivial amount of time, especially if you call it dozens of times.

(I could be convinced to optimize load_system_paasta_config a different way -- honestly I think it might be a lot cleaner to lean on the caching / singleton-y behavior of load_system_paasta_config, but don't have it bother to check whether the files need reloading most of the time. Instead we'd have some method to force reloading.)

The last commit makes load_service_namespace_config faster by making it only read smartstack.yaml instead of reading [seven files](https://github.com/Yelp/service_configuration_lib/blob/02b8b297b53efbedbae30ad929765d1a6cbd5c18/service_configuration_lib/__init__.py#L139) and then immediately discarding most of that information.

It's a little hard to tell _exactly_ how much faster this makes deployd, since we don't have graphs of throughput numbers or CPU time consumed by workers, but I estimate that all these changes together seems to make it roughly 10x faster.

### DIFF
--- a/paasta_tools/marathon_tools.py
+++ b/paasta_tools/marathon_tools.py
@@ -1405,7 +1405,7 @@ def get_matching_appids(
     Useful for fuzzy matching if you think there are marathon
     apps running but you don't know the full instance id"""
     marathon_apps = get_all_marathon_apps(
-        client, service_name=service, embed_tasks=embed_tasks
+        client, service_name=service, instance_name=instance, embed_tasks=embed_tasks
     )
     return [
         app.id for app in marathon_apps if does_app_id_match(service, instance, app.id)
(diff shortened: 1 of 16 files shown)


paasta_tools/marathon_tools.py
```diff
@@ -406,6 +406,7 @@ def load_marathon_service_config(
     cluster: str,
     load_deployments: bool = True,
     soa_dir: str = DEFAULT_SOA_DIR,
+    # system_paasta_config: Optional[SystemPaastaConfig] = None
```
qui, line 409: Do you need this?

EvanKrall: good catch

PR #2850, opened 2020-06-24
Use paasta-tools-go/cmd/paasta as entrypoint

keymone:
The binary is shipped with paasta-tools deb to avoid having to deal with multiple packages providing same usr/bin/paasta link.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,8 +40,6 @@ before_install:
   - sudo apt-get update
   - sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-ce=18.03.1~ce-0~ubuntu devscripts --force-yes
   - eval "$(gimme 1.13.1)"
-  - which go
-  - go version
   - 'if [ "$TRAVIS_BRANCH" = "master" -a "$TRAVIS_PULL_REQUEST" = "false" -a "$DOCKER_USERNAME" != "" ]; then docker login -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"; fi'
 install:
   - pip install coveralls tox==3.2 tox-pip-extensions==1.3.0 ephemeral-port-reserve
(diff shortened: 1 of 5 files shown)


yelp_package/Makefile
```diff
@@ -47,6 +47,11 @@ itest_%: package_$$* bintray_$$*
 	$(DOCKER_RUN) /work/yelp_package/itest/ubuntu.sh paasta-tools_$(ACTUAL_PACKAGE_VERSION)_amd64.deb
 
 package_%: build_$$*_docker
+	rm -rf ../bin && \
+	mkdir -p ../bin && \
+	GO111MODULE=on GOBIN=`pwd`/../bin go get -v github.com/Yelp/paasta-tools-go/cmd/paasta@v0.0.3 && \
```
vkhromov, line 52: nit: `&& \` aren't necessary here

keymone:
> Will it still be possible to use this version of the `paasta` command within our local virtualenv? I guess in this world the `paasta` entrypoint is not strictly necessary, and we can just call `paasta-whatever` directly.

Python’s entrypoint isn’t going anywhere, so inside venv it will get priority because venv changes your PATH. Eventually (and slowly) we will get rid of it, but it’s a good point, maybe we should teach golang’s version to detect virtualenvs.

.travis.yml
```diff
@@ -37,11 +37,9 @@ before_install:
   - sudo bash -c 'echo "deb http://download.docker.com/linux/ubuntu xenial stable" > /etc/apt/sources.list.d/docker.list'
   - sudo apt-key adv --keyserver keyserver.ubuntu.com --recv 9DC858229FC7DD38854AE2D88D81803C0EBFCD88
 
-  # Go repos
-  - sudo add-apt-repository ppa:longsleep/golang-backports -y
-
   - sudo apt-get update
-  - sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-ce=18.03.1~ce-0~ubuntu devscripts golang-1.13 --force-yes
+  - sudo apt-get -o Dpkg::Options::="--force-confnew" install -y docker-ce=18.03.1~ce-0~ubuntu devscripts --force-yes
+  - eval "$(gimme 1.13)"
```
Rob-Johnson, line 42: 1.14?

PR #2853, opened 2020-06-25
Make HPA autoscaling target = setpoint + offset when creating HPA fro…

stug:
…m "autoscaling" (vs "horizontal_autoscaling") section of soa configs.

This is causing autoscaling to behave differently when we migrate services to kubernetes since we currently do not look at the offset.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -517,7 +517,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         autoscaling_params = self.get_autoscaling_params()
         metrics_provider = autoscaling_params["metrics_provider"]
         metrics = []
-        target = autoscaling_params["setpoint"]
+        target = autoscaling_params["setpoint"] + autoscaling_params.get("offset", 0)
         annotations: Dict[str, str] = {}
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -517,7 +517,7 @@ def get_autoscaling_metric_spec(
         autoscaling_params = self.get_autoscaling_params()
         metrics_provider = autoscaling_params["metrics_provider"]
         metrics = []
-        target = autoscaling_params["setpoint"]
+        target = autoscaling_params["setpoint"] + autoscaling_params.get("offset", 0)
```
EvanKrall, line 520: This isn't quite right. In the mesos autoscaler, regardless of offset, if measured load == setpoint, we leave the number of instances alone.

Basically the fraction we scale up or down is `(load - offset) / (setpoint - offset) - 1`. Regardless of the offset, if load = setpoint, then this is 0; we don't scale up or down. When we're not at the setpoint, the offset helps determine the slope of the line.

What we actually want is:
- subtract offset from the measured load, probably in the signalflow code.
- also subtract offset from the setpoint

stug:
Closing this in favor of https://github.com/Yelp/paasta/pull/2858 which solves this and adds the moving_average policy for http & uwsgi metrics

PR #2856, opened 2020-06-30
Always sync HPAs and Pod Disruption Budgets in setup_kubernetes_job

stug:
Previously we only synced these if we determined that the Deployment needed to be updated, but there are cases where users change autoscaling settings (which only affect the HPA) without changing anything that affects the Deployment.

See PAASTA-16675

---

Note that I am also getting rid of the `is_paused` annotation we were putting on deployments.  This really was just so that we would know we needed to call `app.update()` so that the HPA got updated when autoscaling transitions between paused and resumed.  However, because of what I'm describing above, we actually need to check these every time, so we don't need this extra bit of coordination.

I also considered just having us update the deployment every time, since it seems like that is fine -- if no fields have changed it just bumps the `resourceVersion` and `generation` fields, but the pods seem to be unaffected.  Let me know if you think that'd be better.

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -29,14 +29,11 @@ from typing import Mapping
 from typing import Optional
 from typing import Sequence
 from typing import Tuple
-from typing import Union
 
 import a_sync
 import aiohttp
 from kazoo.client import KazooClient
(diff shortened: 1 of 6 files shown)


drmorr0:
I'm pretty sure this is OK.  It's hard to know how what all the effects of these types of changes are, but this seems minor enough that we can ship it and just keep an eye on it.

PR #2857, opened 2020-07-01
Add support to emit task allocation metrics for all namespaces with matching given prefix

88manpreet:
Current state: [get_running_task_allocation](https://sourcegraph.yelpcorp.com/mirrors/Yelp/paasta@ace761b252ddc14ebce9b98f19c12cbf1a7f0305/-/blob/paasta_tools/contrib/get_running_task_allocation.py#L222:26) is currently hard-coded to fetch task allocation metrics from kubernetes pods only for [paasta](https://sourcegraph.yelpcorp.com/mirrors/Yelp/paasta@ace761b252ddc14ebce9b98f19c12cbf1a7f0305/-/blob/paasta_tools/kubernetes_tools.py#L1828:1) namespace.
Ideal state is to emit task allocation metrics for all namespaces prefixed as `paasta`. This will eventually be used in cost reports to report cost allocation for all kubernetes pods.

The code change is backwards compatible (i.e. no need to change to the [puppet](https://sourcegraph.yelpcorp.com/sysgit/puppet@45039da539c1efac674dff4dd075424b205fd1a9/-/blob/modules/profile_kube/manifests/master_monitoring.pp#L56:84) deployment to emit metrics to scribe topic).

By default, task allocation metrics are emitted for all namespaces prefixed as `paasta`. User could provide different prefix using `--namespace-prefix`.

### DIFF
--- a/paasta_tools/contrib/get_running_task_allocation.py
+++ b/paasta_tools/contrib/get_running_task_allocation.py
@@ -142,7 +142,9 @@ def get_container_type(container_name: str, instance_name: str) -> str:
     """
     To differentiate between main service containers and sidecars
     """
-    if instance_name and container_name == kubernetes_tools.sanitise_kubernetes_name(instance_name):
+    if instance_name and container_name == kubernetes_tools.sanitise_kubernetes_name(
+        instance_name
+    ):
         return MAIN_CONTAINER_TYPE
     else:
         return container_name
@@ -214,8 +216,8 @@ def parse_args() -> argparse.Namespace:
     )
     parser.add_argument(
         "--namespace-prefix",
-        help="prefix of the namespace to fetch the logs for" \
-	    "Used only when scheduler is kubernetes",
+        help="prefix of the namespace to fetch the logs for"
+        "Used only when scheduler is kubernetes",
         dest="namespace_prefix",
         default="paasta",
     )
@@ -229,7 +231,7 @@ def main(args: argparse.Namespace) -> None:
     else:
         client = KubeClient()
         all_namespaces = kubernetes_tools.get_all_namespaces(client)
-        matching_namespaces = [n for n in all_namespaces if args.namespace_prefix in n]
+        matching_namespaces = [n for n in all_namespaces if n.startswith(args.namespace_prefix)]
         for matching_namespace in matching_namespaces:
             display_task_allocation_info(cluster, args.scheduler, matching_namespace)
 
--- a/paasta_tools/contrib/get_running_task_allocation.py
+++ b/paasta_tools/contrib/get_running_task_allocation.py
@@ -101,9 +101,11 @@ async def get_mesos_task_allocation_info() -> Iterable[TaskAllocationInfo]:
     return info_list
 
 
-def get_all_running_kubernetes_pods(kube_client: KubeClient,) -> Iterable[V1Pod]:
+def get_all_running_kubernetes_pods(
+    kube_client: KubeClient, namespace: str
+) -> Iterable[V1Pod]:
     running = []
-    for pod in kubernetes_tools.get_all_pods(kube_client):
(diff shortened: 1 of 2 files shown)


88manpreet:
Test results:

[Default command](https://fluffy.yelpcorp.com/i/WmRmGWTrKfdSwJsgJFhT8gxQqVQhrf62.html#L1
): (default namespace: paasta)

Namespace: [ paasta-spark](https://fluffy.yelpcorp.com/i/VQKWM6bKHxckF67tXmBc9S6n6xgSj5v9.html)
https://fluffy.yelpcorp.com/i/VQKWM6bKHxckF67tXmBc9S6n6xgSj5v9.html

Namespace: [paasta](https://fluffy.yelpcorp.com/i/P6XxNXTsXFZVTNkTjKl1QSgnPn7Ncv56.html)

paasta_tools/kubernetes_tools.py
```diff
@@ -2283,6 +2288,8 @@ def create_kubernetes_secret_signature(
 
 
 def sanitise_kubernetes_name(service: str,) -> str:
+    if not service:
```
88manpreet, line 2291: Reason for adding this:
`paasta-elasticsearches` as service and instance as [null](https://fluffy.yelpcorp.com/i/wND4SG2TpflSqhh6jKgVN1qmngxkLshh.html
).

paasta_tools/contrib/get_running_task_allocation.py
```diff
@@ -208,12 +212,26 @@ def parse_args() -> argparse.Namespace:
...
     return parser.parse_args()
 
 
 def main(args: argparse.Namespace) -> None:
     cluster = load_system_paasta_config().get_cluster()
-    info_list = get_task_allocation_info(args.scheduler)
+    client = KubeClient()
+    all_namespaces = kubernetes_tools.get_all_namespaces(client)
+    matching_namespaces = [n for n in all_namespaces if args.namespace_prefix in n]
+    for matching_namespace in matching_namespaces:
```
qui, line 229: This seems like it will duplicate data for Mesos - e.g. if there are 3 matching kubernetes prefixes, we'll output data for Mesos 3 times. Can you apply the namespace logic only if the scheduler is Kubernetes?

paasta_tools/kubernetes_tools.py
qui: Instead of allowing this error, I think we handle this case in the script:  https://github.com/Yelp/paasta/blob/master/paasta_tools/contrib/get_running_task_allocation.py#L139

paasta_tools/contrib/get_running_task_allocation.py
88manpreet: Good catch. Addressed.

paasta_tools/kubernetes_tools.py
88manpreet: After moving the check to above line, test result: https://fluffy.yelpcorp.com/i/jnZTg0bXRr5l6dGmN3QvjB0hc5f6TJph.html#L60-62

paasta_tools/contrib/get_running_task_allocation.py
```diff
@@ -208,12 +212,30 @@ def parse_args() -> argparse.Namespace:
...
 
 def main(args: argparse.Namespace) -> None:
     cluster = load_system_paasta_config().get_cluster()
-    info_list = get_task_allocation_info(args.scheduler)
+    if args.scheduler == "mesos":
+        display_task_allocation_info(cluster, args.scheduler, args.namespace_prefix)
+    else:
+        client = KubeClient()
+        all_namespaces = kubernetes_tools.get_all_namespaces(client)
+        matching_namespaces = [n for n in all_namespaces if args.namespace_prefix in n]
```
qui, line 232: nit: `n.startswith(args.namespace_prefix)` instead of just checking if it's contained, to make it a real prefix match

PR #2866, opened 2020-07-16
SEC-12874: Add prototypical authz to mark-for-deployment

y-trobinso:
WIP.

The purpose of this review is to allow for separate authz on mark-for-deployment, while allowing for other paasta commands to be executed by unprivileged paasta users. This is needed because certain git hosts (ie, Github) are unable to protect tags or implement this behavior in githooks, preventing this behavior to be enforced by git itself.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -218,10 +218,11 @@ def add_subparser(subparsers):
 def mark_for_deployment(git_url, deploy_group, service, commit, allowed_groups=None):
     """Mark a docker image for deployment"""
     username = get_username()
-    if allowed_groups is not None and username not in ldap_user_search(allowed_groups):
-        logline = f"current user is not authorized to perform this action (should be in one of {allowed_groups})"
-        _log(service=service, line=logline, component="deploy", level="event")
-        return 1
+    if allowed_groups is not None:
+        if not any([username in ldap_user_search(group) for group in allowed_groups]):
+            logline = f"current user is not authorized to perform this action (should be in one of {allowed_groups})"
+            _log(service=service, line=logline, component="deploy", level="event")
+            return 1
 
     tag = get_paasta_tag_from_deploy_group(
         identifier=deploy_group, desired_state="deploy"
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -219,7 +219,7 @@ def mark_for_deployment(git_url, deploy_group, service, commit, allowed_groups=N
     """Mark a docker image for deployment"""
     username = get_username()
     if allowed_groups is not None and username not in ldap_user_search(allowed_groups):
-        logline = f"current user is not authorized to perform this action (must be in one of {allowed_groups})"
+        logline = f"current user is not authorized to perform this action (should be in one of {allowed_groups})"
         _log(service=service, line=logline, component="deploy", level="event")
         return 1
 
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -216,7 +216,9 @@ def add_subparser(subparsers):
     list_parser.set_defaults(command=paasta_mark_for_deployment)
 
 
-def mark_for_deployment(git_url, deploy_group, service, commit, allowed_groups=DEFAULT_ALLOWED_PUSH_GROUPS):
+def mark_for_deployment(
+    git_url, deploy_group, service, commit, allowed_groups=DEFAULT_ALLOWED_PUSH_GROUPS
+):
     """Mark a docker image for deployment"""
     username = get_username()
     if allowed_groups is not None:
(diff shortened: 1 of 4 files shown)


EvanKrall:
huh, you're right: tags can't be protected like branches can. https://github.community/t/feature-request-protected-tags/1742/25 / https://github.com/isaacs/github/issues/1091

Odd, since tags and branches are basically the same thing server-side.

EvanKrall:
I wouldn't rely on userspace authz for security -- someone could just clone the paasta repo, remove the authz, and run mark-for-deployment (Or just create the git tag manually).

Some alternatives:

1. We could switch back to branches as the source of truth for what a deploy group should be using. This would allow us to enable github's branch restrictions: https://docs.github.com/en/github/administering-a-repository/enabling-branch-restrictions. If we switched back to the branch format that we used to use before #326 / #231 / #252 (just `paasta-{deploy_group}`), we'd lose the history of what was marked for deployment when. (AFAIK we pretty much just use this for suggesting versions to roll back to in `paasta rollback`. Maybe if we switched back to branches, we'd keep creating the tags and just accept the risk that a non-pushmaster could manipulate the `paasta rollback` suggestions.)

2. We could store mark-for-deployment / start-stop data somewhere other than git tags. We chose git branches/tags originally because git already uses SSH, which provides easy authn at the command line, and gitolite made it possible to authz tag creation.
I'm not sure I have a great alternative suggestion, but we'd want, at a minimum:

- easy authn/authz of CLI users, jenkins.
- backed up
- a data model flexible enough to store both the current and historical values of deploy_group -> SHA.

zachm:
@EvanKrall just to add a little more context. This isn't really security or authz per se, but more of a compliance guardrail. Our rough strategy, looks like this, so let me know what you think:

* For in-scope services like YM, only a subset of users (in an AD group) should be able to MFD.
* Most (~85%) of MFD actions happen via Jenkins, which already does the right think in this way.
* The remainder happen via the PaaSTA CLI, which is why we're here talking to you.
* This PR is really intended as a footgun safety, so that folks don't unintentionally violate controls.
* We recognize that anyone could just push the tag manually to git, or edit the CLI code to remove the check.
* Because this PR doesn't provide full coverage, we are also running reporting on "paasta_audit_logs" to look for anything that is amiss, and would triage/remediate from there.

Let me know if that changes anything from your perspective. Thanks!

y-trobinso:
Happy to discuss this further as needed. The goal of this PR is to provide a stopgap in a world in which we still use git tags, though the alternatives you propose seem potentially viable.

I'll roll this over to internal discussion to address particulars.

danielpops:
> Because this PR doesn't provide full coverage, we are also running reporting on "paasta_audit_logs" to look for anything that is amiss, and would triage/remediate from there.

Given that fact, I think it's important that we frequently exercise the nefarious activity scenario and verify that our logging / monitoring is always working

paasta_tools/utils.py
```diff
@@ -113,6 +115,12 @@
 
 DEFAULT_CPU_PERIOD = 100000
 DEFAULT_CPU_BURST_ADD = 1
+LDAP_SEARCH_BASE = os.environ.get("PAASTA_SECRET_LDAP_SEARCH_BASE")
+LDAP_SEARCH_OU = os.environ.get("PAASTA_SECRET_LDAP_SEARCH_OU")
+LDAP_HOST = os.environ.get("PAASTA_SECRET_LDAP_HOST")
+LDAP_USERNAME = os.environ.get("PAASTA_SECRET_LDAP_USERNAME")
+LDAP_PASSWORD = os.environ.get("PAASTA_SECRET_LDAP_PASSWORD")
```
EvanKrall, line 122: I think PAASTA_SECRET_ variables are only ever set inside of containers that are run by paasta. `mark-for-deployment` doesn't run inside paasta containers (only dev boxes and Jenkins), so these won't be set.

Probably the best bet is to put this stuff in /etc/paasta, add getters to `SystemPaastaConfig`, add type definitions in `SystemPaastaConfigDict`, and then have . Though be aware that the LDAP creds will have to be world-readable for this to work.

You'd then call `load_system_paasta_config` in ldap_user_search or the calling site, or pass an existing SystemPaastaConfig object in, if available.

y-trobinso: I chatted with @ztm to confirm we won't be posting these new push groups as Linux groups, which prevents using `os.get_groups()`. As for having the LDAP creds be world readable, that seems to be no different than our existing setup, which we're fine with for now.

As such, I've implemented the change you proposed in the second comment above.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -77,6 +79,7 @@
 DEFAULT_DEPLOYMENT_TIMEOUT = 3600  # seconds
 DEFAULT_AUTO_CERTIFY_DELAY = 600  # seconds
 DEFAULT_SLACK_CHANNEL = "#deploy"
+DEFAULT_ALLOWED_PUSH_GROUPS = os.environ.get("PAASTA_SECRET_ALLOWED_PUSH_GROUPS", None)
```
EvanKrall, line 82: should this also be in SystemPaastaConfig?

y-trobinso: +1. Fixed.

PR #2867, opened 2020-07-20
MONK-1127 Adding support for prestop hook and termination grace period in paasta

rohangulati:
## Context
Adding support for defining prestop hook and termination grace period in PaaSTA. An example configuration for this would be as follow
```yaml
cpu: 2
lifecycle:
    termination_grace_period_s: 10
    pre_stop_command: 
        - /bin/bash
        - C
        - sleep 30
```

## Testing
`make test` passed and added new tests

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -242,6 +242,11 @@ class CustomResourceDefinition(NamedTuple):
     group: str
 
 
+class KubeLifecycleDict(TypedDict, total=False):
+    termination_grace_period_s: int
+    pre_stop_command: Union[str, List[str]]
+
+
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -957,6 +961,17 @@ def get_kubernetes_containers(
         )
         return containers
 
+    def get_kubernetes_container_termination_action(self) -> V1Handler:
+        command = self.config_dict.get("lifecycle", KubeLifecycleDict({})).get(
+            "pre_stop_command", []
+        )
+        # default pre stop hook for the container
+        if len(command) == 0:
```
Rob-Johnson, line 969: `if not command` ?

rohangulati: Fixed

PR #2878, opened 2020-08-05
Fix the check OOM event script if there aren't any OOMs

drmorr0:
If there haven't been any OOM events recorded in the last day, the
scribe stream to track the OOM events will be empty.  This will make the
OOM checker script crash.  This fix prevents a crash in this case.
(Unfortunately if you get the scribe stream name wrong it will _also_
return the same exception, I don't know any good way to get around
that).

### DIFF
--- a/paasta_tools/check_oom_events.py
+++ b/paasta_tools/check_oom_events.py
@@ -28,6 +28,7 @@ from paasta_tools.utils import load_system_paasta_config
 
 try:
     from scribereader import scribereader
+    from clog.readers import StreamTailerSetupError
 except ImportError:
     scribereader = None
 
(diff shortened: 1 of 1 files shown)


paasta_tools/check_oom_events.py
```diff
@@ -104,13 +105,19 @@ def read_oom_events_from_scribe(cluster, superregion, num_lines=1000):
...
+                j = json.loads(line)
+                if j.get("cluster", "") == cluster:
+                    yield j
+            except json.decoder.JSONDecodeError:
+                pass
+    except StreamTailerSetupError as e:
+        if "No data in stream" in str(e):
             pass
+        else:
+            raise e
```
kawaiwanyelp, line 120: `if "No data in stream" not in str(e):`
&ensp;&ensp;`raise e` ?

drmorr0: Eh, I guess I kinda feel like this way is a little easier to reason about?  Having the double-negative there is weird.

PR #2885, opened 2020-08-12
Support iam_role using AWS pod identity solution

vkhromov:
## Support iam_role using AWS pod identity solution

Added `iam_role_provider` parameter for services which can be set to
either `kiam` (default) or `aws`.  If it's set to `aws` then PaaSTA uses
the AWS pod identity solution [1] instead of `kiam` to support per-pod
IAM profiles.

[1] https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/

## Workaround for AWS pod identity and non-root programs

AWS pod identity creates a token with `0600` permissions and
`root:root`:
```
-rw------- 1 root root 1498 Aug 13 12:47 /run/secrets/eks.amazonaws.com/serviceaccount/..2020_08_13_12_47_11.793276204/token
```
This prevents programs running inside containers using a non-root
account to read the token.  See [1] for details.

This CR working around that by adding
```
  securityContext:
›         fsGroup: 65534
```
into the pod spec.
After that the token is owned by the given group and has `0640`
permissions:
```
-rw-r----- 1 root nobody 1498 Aug 13 14:20 /run/secrets/eks.amazonaws.com/serviceaccount/..2020_08_13_14_20_31.793276204/token      ```

The id of the group can be changed via the `fs_group` service parameter.

[1] https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -77,6 +77,7 @@ from kubernetes.client import V1PersistentVolumeClaimSpec
 from kubernetes.client import V1Pod
 from kubernetes.client import V1PodAffinityTerm
 from kubernetes.client import V1PodAntiAffinity
+from kubernetes.client import V1PodSecurityContext
 from kubernetes.client import V1PodSpec
 from kubernetes.client import V1PodTemplateSpec
 from kubernetes.client import V1Probe
@@ -1330,6 +1331,18 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
                 pod_spec_kwargs[
                     "service_account_name"
                 ] = create_or_find_service_account_name(iam_role)
+                # PAASTA-16919: remove everything related to fs_group when
+                # https://github.com/aws/amazon-eks-pod-identity-webhook/issues/8
+                # will be fixed.
+                fs_group = self.get_fs_group()
+                if fs_group is None:
+                    # We need some reasoable default for group id of a process
+                    # running inside the container.  Seems like most of such
+                    # programs run as `nobody`, let's use that as a default.
+                    fs_group = 65534
+                pod_spec_kwargs["security_context"] = V1PodSecurityContext(
+                    fs_group=fs_group
+                )
         else:
             annotations["iam.amazonaws.com/role"] = self.get_iam_role()
 
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -19,6 +19,7 @@ import json
 import logging
 import math
 import os
+import re
 from datetime import datetime
 from enum import Enum
 from pathlib import Path
@@ -85,6 +86,7 @@ from kubernetes.client import V1RollingUpdateDeployment
 from kubernetes.client import V1Secret
 from kubernetes.client import V1SecretKeySelector
 from kubernetes.client import V1SecurityContext
+from kubernetes.client import V1ServiceAccount
 from kubernetes.client import V1StatefulSet
 from kubernetes.client import V1StatefulSetSpec
 from kubernetes.client import V1TCPSocketAction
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by vkhromov** [comment]
```
python -c "
import boto3
print(boto3.client('sts').get_caller_identity().get('Arn'))
"
+ python -c '
import boto3
print(boto3.client('\''sts'\'').get_caller_identity().get('\''Arn'\''))
'
arn:aws:sts::528741615426:assumed-role/flink-kubernetes/botocore-session-1597331179
sleep infinity
+ sleep infinity
```

File: paasta_tools/kubernetes_tools.py [hunk 1]
```diff
@@ -1322,6 +1324,32 @@ def get_pod_template_spec(
                 "termination_grace_period_seconds"
             ] = termination_grace_period
 
+        if self.get_iam_role_provider() == "aws":
+            annotations["iam.amazonaws.com/role"] = ""
```
keymone, line 1328: is this defensive against something else somewhere setting this attribute? in this case i would instead fail very loudly with something like `raise RuntimeError("can't have iam.amazonaws.com/role with aws provider")`. silently disappearing attributes can be quite annoying to debug.

paasta_tools/kubernetes_tools.py
```diff
@@ -2568,3 +2596,33 @@ def to_node_label(label: str) -> str:
...
+
+
+_RE_NORMALIZE_IAM_ROLE = re.compile(r"[^0-9a-zA-Z]+")
+
+
+def create_or_find_service_account_name(
+    iam_role: str, namespace: str = "paasta"
+) -> str:
+    kube_client = KubeClient()
+    sa_name = "paasta--" + _RE_NORMALIZE_IAM_ROLE.sub("-", iam_role)
```
keymone, line 2614: if serviceaccount is a namespaced object, do we need `paasta--` prefix?

paasta_tools/long_running_service_tools.py
```diff
@@ -203,6 +205,12 @@ def get_replication_crit_percentage(self) -> int:
     def get_iam_role(self) -> str:
         return self.config_dict.get("iam_role", "")
 
+    def get_iam_role_provider(self) -> str:
+        return self.config_dict.get("iam_role_provider", "kiam")
+
+    def get_fs_group(self) -> Optional[int]:
```
keymone, line 211: maybe lets move the default `33` and explanation comment here?

paasta_tools/kubernetes_tools.py
vkhromov: `iam.amazonaws.com/role` is used by `kiam`.
Previously we set `iam.amazonaws.com/role` to `""` if `am_role` wasn't set.
So, I just want to separate aws changes from kiam changes, and don't change anything related to kiam, to make sure that `kiam` will behave exactly the same way as before.
We could remove that later.

paasta_tools/kubernetes_tools.py
vkhromov: you're right, let's remove it.

paasta_tools/long_running_service_tools.py
vkhromov: `fs_group` might be useful besides the AWS pod identity-related workaround, but only for the workaround it need to be set to `33` by default.

paasta_tools/kubernetes_tools.py
vkhromov: Actually
```
NAME                                                     SECRETS   AGE
cassandra-operator                                       1         406d
default                                                  1         407d
flink-operator                                           1         407d
kafka-operator                                           1         302d
nrtsearch-operator                                       1         93d
paasta--arn-aws-iam-528741615426-role-flink-kubernetes   1         5h32m
zzzzzzz                                                  1         3d4h
```
so, even it's namespaced, there could be other service accounts in the `paasta` namespace not related to paasta tools, so we probably better separate them somehow?

vkhromov:
After collecting more information about services that use `iam_role` I've changed the default `fs_group` value to `65534`.

paasta_tools/kubernetes_tools.py
vkhromov: ok, going to merge this PR, we can change it in a next PR.

PR #2888, opened 2020-08-14
Print instance name in lighter blue so it's easier to read

JoeMalt:
Trivial change for PAASTA-16917: the dark blue text is pretty difficult to read, cyan is easier.

Before:
Screenshot 2020-08-14 at 17 40 48

After:
Screenshot 2020-08-14 at 17 40 30

`make test` and `make acceptance` pass. Tested on Mesos and k8s services.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -237,7 +237,7 @@ def paasta_status_on_api_endpoint(
     system_paasta_config: SystemPaastaConfig,
     verbose: int,
 ) -> int:
-    output.append("    instance: %s" % PaastaColors.blue(instance))
+    output.append("    instance: %s" % PaastaColors.cyan(instance))
     client = get_paasta_api_client(cluster, system_paasta_config)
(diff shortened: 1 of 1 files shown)


vkhromov:
Not trying to block this PR, but "blue" looks fine for me:
image
The way a terminal represents standard ANSI colours is up to the terminal, and usually configurable, e.g. this is colour my iTerm uses to show anything "blue":
image
So, instead of fixing all possible terminal applications which use uncomfortable colours, maybe it's better to adjust your terminal? 🙂

keymone:
i agree that default blue is a bit too dark for black background, something brighter would be more readable and will cost us nothing.

JoeMalt:
Not sure why `make k8s_itests` failed on Travis (seems to be a pip issue), but they just passed locally so I'm confident nothing is broken.

@vkhromov: thanks for the pointer to iTerm configs, I didn't know you could do that. I'm still going to merge this PR to improve the out-of-the-box UX.

PR #2889, opened 2020-08-18
MESH-857 Pod readiness checks supporting Envoy / Smartstack

gvolco:
Adds a new pod readiness script (`/check_proxy_up.sh`) that supports both Envoy and Smartstack.

For now, this readiness script won't be enabled on any service, meaning that we won't have any bounces after merging.

To enable the new readiness check on a single service, we can add the following to yelpsoa-configs (`KubernetesDeploymentConfigDict`):
```
bounce_health_params:
  check_envoy
```
`check_haproxy` can be set as well to check smartstack (and can coexist with `check_envoy`).

To enable the new readiness check on every service, the following param can be added to `SystemPaastaConfigDict`
```
enable_envoy_readiness_check: true
```
`enable_nerve_readiness_check` can be set as well to check smartstack (and can coexist with `enable_envoy_readiness_check`)

### Tests:
- Multiple manual tests on a k8s node:
```
# PAASTA_HOST="XXXXXXXX" PAASTA_POD_IP=YYYYYY python paasta_tools/contrib/is_pod_healthy_in_proxy.py --enable-smartstack --enable-envoy 8888 [service].[instance1] [service].[instance2] [service].[instance3]
UserWarning: clog is unavailable
  warnings.warn("clog is unavailable")
# echo $?
0
``` 
- Built paasta_tools and rolled it out to kubestage.
  - No bounces detected
  - New deployments still use /check_smartstack_up.sh as expected 

Note: At some point after making a new paasta-tools release (and before using the new readiness check on any service), I'll have to update the hacheck-sidecar docker image to get the latest paasta-tools pkg and the new check script.

### DIFF
--- a/paasta_tools/contrib/is_pod_healthy_in_proxy.py
+++ b/paasta_tools/contrib/is_pod_healthy_in_proxy.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+# Copyright 2015-2018 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+import os
+import sys
+
+from paasta_tools.envoy_tools import are_services_up_in_pod as is_envoy_ready
+from paasta_tools.smartstack_tools import (
+    are_services_up_on_ip_port as is_smartstack_ready,
+)
+from paasta_tools.utils import load_system_paasta_config
+
+
+system_paasta_config = load_system_paasta_config()
+
+synapse_port = system_paasta_config.get_synapse_port()
+synapse_host = "169.254.255.254"
+synapse_haproxy_url_format = system_paasta_config.get_synapse_haproxy_url_format()
+
+envoy_host = os.environ["PAASTA_HOST"]
+envoy_admin_port = system_paasta_config.get_envoy_admin_port()
+envoy_admin_endpoint_format = system_paasta_config.get_envoy_admin_endpoint_format()
+pod_ip = os.environ["PAASTA_POD_IP"]
+
+
+def get_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser()
+
+    parser.add_argument(
+        "--enable-smartstack",
+        dest="smartstack_readiness_check_enabled",
+        action="store_true",
+        help="Check smartstack readiness",
+    )
+
+    parser.add_argument(
+        "--enable-envoy",
+        action="store_true",
+        dest="envoy_readiness_check_enabled",
+        help="Check envoy readiness",
+    )
+
+    parser.add_argument(
+        "pod_port", help="Pod Port", type=int,
+    )
+
+    parser.add_argument(
+        "services", nargs="+", help="List of service.instance names",
+    )
+
+    return parser
+
+
+def main() -> None:
+    args = get_parser().parse_args()
+
+    if args.smartstack_readiness_check_enabled:
+        smartstack_ready = is_smartstack_ready(
+            synapse_host=synapse_host,
+            synapse_port=synapse_port,
+            synapse_haproxy_url_format=synapse_haproxy_url_format,
+            services=args.services,
+            host_ip=pod_ip,
+            host_port=args.pod_port,
+        )
+    else:
+        smartstack_ready = True
+
+    if args.envoy_readiness_check_enabled:
+        envoy_ready = is_envoy_ready(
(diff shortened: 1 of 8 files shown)


paasta_tools/envoy_tools.py
```diff
@@ -49,10 +47,48 @@
...
+            envoy_admin_uri_format=envoy_admin_uri_format,
+        )
+
+        for be in backends:
+            if (
+                be[0]["eds_health_status"] == "HEALTHY"
+                and be[0]["address"] == pod_ip
+                and be[0]["port_value"] == pod_port
+            ):
+                services_with_atleast_one_backend_up[registration] = True
```
vkhromov, line 83: IMO `services_with_atleast_one_backend_up` makes the code unnecessary complicated. 🙂 
Maybe `return False` here, `return True` instead of `return all(services_with_atleast_one_backend_up.values())` below, and get rid of `services_with_atleast_one_backend_up`?

gvolco: You are right. I guess this approach (using `services_with_atleast_one_backend_up`) made sense on smartstack, as synapse [would often keep](https://github.com/Yelp/paasta/blob/master/paasta_tools/smartstack_tools.py#L393) old backends around for a while.
But I don't think this is needed on Envoy, so I'll go with your recommendation :)

gvolco:
I have just rebased from master to include changes on https://github.com/Yelp/paasta/pull/2887 and fixed merge conflicts/tests

paasta_tools/kubernetes_tools.py
```diff
@@ -724,6 +724,19 @@ def get_aws_ebs_volume_name(self, aws_ebs_volume: AwsEbsVolume) -> str:
...
         return d.get("mode", "RO") == "RO"
 
+    def get_readiness_check_script(
+        self, system_paasta_config: SystemPaastaConfig
+    ) -> str:
+        """Script to check if a service is up in smartstack / envoy"""
+        enable_envoy_check = self.get_enable_envoy_readiness_check(system_paasta_config)
+        enable_nerve_check = self.get_enable_nerve_readiness_check(system_paasta_config)
+        if enable_nerve_check and enable_envoy_check:
+            return "/check_proxy_up.sh --enable-smartstack --enable-envoy"
```
ymilki, line 734: The check scripts themselves could be configured in the conf (`system_paasta_config.get_nerve_readiness_check_script()`). Are we ok with hardcoding them here instead?

At least we are building the sidecar in the same repo so they shouldn't drift apart given testing.

paasta_tools/utils.py
```diff
@@ -2032,6 +2032,18 @@ def get_enable_client_cert_auth(self) -> bool:
         """
         return self.config_dict.get("enable_client_cert_auth", True)
 
+    def get_enable_nerve_readiness_check(self) -> bool:
+        """
+        If enabled perform rediness checks on nerve
```
ymilki, line 2037: nit: typo in comments here and below

yelp_package/dockerfiles/hacheck-sidecar/Dockerfile
```diff
@@ -3,9 +3,10 @@ FROM docker-dev.yelpcorp.com/xenial_yelp
 ARG PIP_INDEX_URL=https://pypi.yelpcorp.com/simple
 ENV PIP_INDEX_URL=$PIP_INDEX_URL
 
-RUN apt-get update && apt-get install -y hacheck python paasta-tools=0.81.14-yelp1
+RUN apt-get update && apt-get install -y hacheck python paasta-tools
```
ymilki, line 6: I'm not so sure about unpinning paasta-tools, though it depends on the deployment process for the sidecar. I just don't want any surprises if paasta-tools has to be rolled back and we forget the sidecar needs to be rebuilt. Since we only have the docker image is the only artifact, it's not clear what is installed in the sidecar at a glance.

This actually leads me to realize hacheck and python have the same problem.

paasta_tools/contrib/is_pod_healthy_in_proxy.py
```diff
@@ -0,0 +1,104 @@
...
+    else:
+        envoy_ready = True
+
+    if smartstack_ready and envoy_ready:
+        sys.exit(0)
+    else:
+        if not smartstack_ready:
+            print("Failed local smartstack check")
+        if not envoy_ready:
+            print("Failed local envoy check")
```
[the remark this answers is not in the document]
gvolco: Sure, I'll add more details to the err msg (although you would be describing a pod to get these errors, so the pod ip and node ip are already known).

paasta_tools/kubernetes_tools.py
gvolco: np, I'll add these scripts as params on system_paasta_config.

yelp_package/dockerfiles/hacheck-sidecar/Dockerfile
gvolco: Yeah I thought about this, but I don't know yet which paasta_tools version will contain this change. I could re-pin afterwards though.
Note that the hacheck sidecar container image needs to be [manually built and pushed](https://github.com/Yelp/paasta/blob/master/yelp_package/dockerfiles/hacheck-sidecar/Makefile#L6).

gvolco:
Implemented review feedback

PR #2891, opened 2020-08-19
Swagger to OpenAPI

keymone:
- oapi.yaml alongside swagger.json
- pre-commit hook to keep the two in sync
- makefile target to generate paasta_tools/paastaapi
- generated paastaapi
- client builder function in paasta_tools/api


Makefile
```diff
@@ -85,3 +85,15 @@ install-hooks:
 
 k8s_itests: .paasta/bin/activate
 	make -C k8s_itests all
+
+openapi-codegen:
```
vkhromov, line 89: Can we run it from a pre-commit hook, so the generated files always will be in sync with schemas?
We also better run it from a test inside Travis to make sure that no PR with the pre-commit hook skipped will pass.

mypy.ini
```diff
@@ -15,6 +15,9 @@ ignore_missing_imports = False
 # Only specific test files (defined in tox.ini) will be checked, so let's check everything in those.
 check_untyped_defs = True
 
+[mypy-paasta_tools.paastaapi.*]
+ignore_errors = True
```
vkhromov, line 19: It seems that the generated classes don't have typing annotation?
If that's correct, can they be added somehow?  They are really useful to catch errors early on a devbox instead of doing the same later at prod.

vkhromov (commented):
Looks good!
Some comments.

mypy.ini
keymone: official openapi generator isn't targeting very modern python features. it's the same as the one used to generate kubernetes api - do we have mypy support there? otherwise we'll have to wait (or contribute) for https://github.com/triaxtec/openapi-python-client to get more mature. i didn't spend enough time to figure out what exactly isn't supported in our oapi.yaml, maybe during crustacean.

vkhromov: TBH I'm not sure which kubernetes api -generated classes you're talking about.
I was looking at the [openapi-generated `EnvoyBackend` class](https://github.com/Yelp/paasta/pull/2891/files#diff-aed879f20100f44329e2acda282233d0R21) which looks like is going to replace the hand-made version from https://github.com/Yelp/paasta/blob/master/paasta_tools/envoy_tools.py#L37 .

Makefile
vkhromov: I agree that running it from a pre-commit hook is up for discussion, but the test at Travis is a must IMO, otherwise we'll always have a drifted schema.

mypy.ini
keymone: would it be a ok idea to wrap auto-generated class in a hand-made one and add the annotations manually? i think it's the best we can do for now. i promise to keep looking into python-native generator because it's got all the goodies - annotations, dataclasses, cleaner class hierarchy, more modern codebase, less python2-compatibility bloat, etc.

Makefile
keymone: travis is running all pre-commit hooks in general_itests target, i've added relevant checks to existing pre-commit hook, this should do the trick.

mypy.ini
vkhromov: We probably should be fine by directly using the auto-generated files for now since we don't have any alternatives anyway.
But please create a ticket about the necessity to eventually upgrade the openapi generator, so we don't forget about that.

Makefile
vkhromov: I'm also wondering if we need to prevent committing changes in `paastaapi/` without having changes in schema files...

paasta_tools/api/client.py
```diff
@@ -83,6 +84,30 @@ def get_paasta_api_client(
...
+    api_endpoints = system_paasta_config.get_api_endpoints()
+    if cluster not in api_endpoints:
+        log.error("Cluster %s not in paasta-api endpoints config", cluster)
+        return None
+
+    parsed = urlparse(api_endpoints[cluster])
+    server_variables = dict(scheme=parsed.scheme, host=parsed.netloc)
+
+    config = paastaapi.Configuration(server_variables=server_variables)
+    client = paastaapi.ApiClient(configuration=config)
```
Rob-Johnson, line 107: do we need to do something here to integrate the PaastaRequestsClient class below just like we have with the swagger client?

Rob-Johnson (approved):
very exciting - looks a good start. have you got a rollout plan for migrating to using it?

paasta_tools/api/client.py
keymone: as i understand we can provide/modify relevant `ssl_*` attributes in `paastaapi.ApiClient`, i didn't test that yet. which environment do we use https in? in api_endpoints i only see http.

vkhromov (approved):
:ship:

paasta_tools/api/client.py
mattmb: Yeah, we never rolled it out anywhere. In theory it's setup in mesosstage but may or may not work in terms of issuing the certs. The other "magic" was here https://github.com/Yelp/paasta/blob/master/paasta_tools/api/auth_decorator.py which makes the client re-issue the cert if it is expired. Anyway, I'm okay if the SSL stuff breaks with this migration, we can always make it work if and when we come back to rolling out the authn layer.

mattmb (approved):

PR #2897, opened 2020-08-24
Separate git server from primary server in git config

kawaiwanyelp:
### Description
I realized when trying to make a git config file that the server we trigger deploys on and the server clone soa-configs from will be different, but are the same right now. So, the two need different keys in the git config.

### Testing done
`make test`

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2493,6 +2493,7 @@ class SystemPaastaConfig:
                     "yelpsoa-configs": {
                         "repo_name": "yelpsoa-configs",
                         "git_server": DEFAULT_SOA_CONFIGS_GIT_URL,
+                        "deploy_server": DEFAULT_SOA_CONFIGS_GIT_URL,
                     },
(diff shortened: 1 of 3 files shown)


paasta_tools/utils.py
```diff
@@ -2493,6 +2493,7 @@ def get_git_config(self) -> Dict:
                     "yelpsoa-configs": {
                         "repo_name": "yelpsoa-configs",
                         "git_server": DEFAULT_SOA_CONFIGS_GIT_URL,
+                        "primary_server": DEFAULT_SOA_CONFIGS_GIT_URL,
```
stug, line 2496: nit: I am not sure the name `primary` here is as evocative as it could be (I have all the context and it took me a sec).  Might be better to call it something like `trigger_deploys_on_server` or something like that.

kawaiwanyelp: I changed it to `deploy_server` instead of `primary_server`.

PR #2898, opened 2020-08-25
Fix pip fails to retry after network errors

vkhromov:
Ubuntu `python-pip` package contains an ancient version of `pip`, which
fails to retry after network error with the following message:
```
  File
  "/venv/share/python-wheels/urllib3-1.13.1-py2.py3-none-any.whl/urllib3/util/retry.py",
  line 228, in increment
      total -= 1
      TypeError: unsupported operand type(s) for -=: 'Retry' and 'int'
```

As a result, most of our Travis jobs fail on initial runs and need several manual retries.

This PR fixes that by upgrading `pip` to a more recent version.

## Testing
Run Travis tests a couple of times with no failures, see https://travis-ci.org/github/Yelp/paasta/builds/720948239 and https://travis-ci.org/github/Yelp/paasta/builds/720955289

### DIFF
--- a/yelp_package/dockerfiles/bionic/Dockerfile
+++ b/yelp_package/dockerfiles/bionic/Dockerfile
@@ -46,6 +46,7 @@ RUN apt-get update > /dev/null && \
         zsh > /dev/null \
     && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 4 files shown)


Rob-Johnson (approved):

gvolco (commented):
lgtm

PR #2899, opened 2020-08-27
MESH-857 Fix hacheck-sidecar command path

gvolco:
`get_readiness_check_script()` and `get_envoy_nerve_readiness_check_script()` will return a string such as `/check_proxy_up.sh --enable-smartstack --enable-envoy` instead of returning a List of arguments (which is what V1ExecAction expects).

This triggers the following errors when enabling Envoy readiness checks:
```
  Warning  Unhealthy  60s (x29 over 5m40s)  kubelet, ip-10-114-45-230.us-west-1.compute.internal  Readiness probe failed: OCI runtime exec failed: exec failed: container_linux.go:346: starting container process caused "exec: \"/check_proxy_up.sh --enable-smartstack --enable-envoy\": stat /check_proxy_up.sh --enable-smartstack --enable-envoy: no such file or directory": unknown
```

So I'm amending the command passed to V1ExecAction, providing each argument as a list item.
Note that this isn't affecting any service as Envoy readiness checks are currently disabled.

### Tests:
- Manually built and installed a new paasta-tools version on kubestage
- No services where bounced as expected
- Enabling envoy checks don't produce the error mentioned above anymore, and the hacheck sidecar gets to the 'ready' status
- Tested this while enabling envoy checks, smartstack checks, and both checks at the same time on a single service.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -726,7 +726,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
 
     def get_readiness_check_script(
         self, system_paasta_config: SystemPaastaConfig
-    ) -> str:
+    ) -> List[str]:
         """Script to check if a service is up in smartstack / envoy"""
         enable_envoy_check = self.get_enable_envoy_readiness_check(system_paasta_config)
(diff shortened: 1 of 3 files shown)


vkhromov (approved):
:ship: assuming that all tests will be green

PR #2900, opened 2020-08-27
Fix check_orphans.py to handle k8s instances

chriskuehl:
k8s instances have their own IP address and always run on port 8888; we can't use the instance's IP to talk to the host's xinetd anymore to find its nerve registrations.

Instead, we use the `name` field from the nerve registration, which for PaaSTA instances is the hostname of the PaaSTA host which we can use to fetch the configured nerve registrations. This is the same as the `instance_id` in the nerve config.

After this change, the number of hosts which fail to fetch registrations drops from 1658 to 32 (all timeout errors). I spot-checked a bunch of those and confirmed that it is checking the right IP, but that those hosts aren't running xinetd because they are not PaaSTA hosts at all (just some other host type using nerve for service discovery).

Note that I had to introduce DNS resolution because we don't put the host IP in the zookeeper registration, only its name. This seems like the pragmatic approach for now as opposed to a larger change to service registration.

I tested this on prod, stagef, stageg, and devc. It discovered a few orphans on stageg which I confirmed were real problems and which aren't being caught with the current (master) version of the script.

### DIFF
--- a/paasta_tools/contrib/check_orphans.py
+++ b/paasta_tools/contrib/check_orphans.py
@@ -1,9 +1,12 @@
 #!/usr/bin/env python3
 import argparse
 import asyncio
+import functools
 import json
 import logging
 import os.path
+import re
+import socket
 import sys
 from collections import defaultdict
 from enum import Enum
@@ -71,6 +74,10 @@ def get_zk_data(ignored_services: Set[str]) -> SmartstackData:
 
 
 class InstanceTuple(NamedTuple):
+    # paasta_host may be different from the service's host if running on k8s.
+    # We need the actual PaaSTA host because the k8s pod does not listen for
+    # xinetd connections.
+    paasta_host: str
     host: str
     port: int
     service: str
@@ -78,10 +85,39 @@ class InstanceTuple(NamedTuple):
 
 def read_from_zk_data(registrations: SmartstackData) -> Set[InstanceTuple]:
     return {
-        InstanceTuple(instance_data["host"], instance_data["port"], service)
+        InstanceTuple(
+            host_to_ip(instance_data["name"], instance_data["host"]),
+            instance_data["host"],
+            instance_data["port"],
+            service,
+        )
         for service, instance in registrations.items()
         for instance_data in instance.values()
     }
+
+
+@functools.lru_cache()
+def host_to_ip(host: str, fallback: str) -> str:
+    """Try to resolve a host to an IP with a fallback.
+
+    Because DNS resolution is relatively slow and can't be easily performed
+    using asyncio, we cheat a little and use a regex for well-formed hostnames
+    to try to guess the IP without doing real resolution.
+
+    A fallback is needed because in some cases the nerve registration does not
+    match an actual hostname (e.g. "prod-db15" or "prod-splunk-master").
+    """
+    for match in (
+        re.match(r"^(\d+)-(\d+)-(\d+)-(\d+)-", host),
+        re.match(r"^ip-(\d+)-(\d+)-(\d+)-(\d+)", host),
+    ):
+        if match:
+            return ".".join(match.groups())
+    else:
+        try:
+            return socket.gethostbyname(host)
+        except socket.gaierror:
+            return fallback
 
 
 async def transfer_one_file(
@@ -93,8 +129,8 @@ async def transfer_one_file(
             asyncio.open_connection(host=host, port=port, limit=2 ** 32), timeout=1.0
         )
         resp = await asyncio.wait_for(reader.read(), timeout=1.0)
-    except (asyncio.TimeoutError, ConnectionRefusedError):
-        logger.warning(f"error getting file from {host}")
+    except (asyncio.TimeoutError, ConnectionRefusedError) as ex:
+        logger.warning(f"error getting file from {host}: {ex!r}")
         return (host, None)
 
     return (host, resp.decode())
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/check_orphans.py
```diff
@@ -71,19 +74,52 @@ def get_zk_data(blacklisted_services: Set[str]) -> SmartstackData:
...
+    to try to guess the IP without doing real resolution.
+
+    A fallback is needed because in some cases the nerve registration does not
+    match an actual hostname (e.g. "prod-db15" or "prod-splunk-master").
+    """
+    for match in (
+        re.match(r"^(\d+)-(\d+)-(\d+)-(\d+)-", host),
+        re.match(r"^ip-(\d+)-(\d+)-(\d+)-(\d+)$", host),
+    ):
+        if match:
```
vkhromov, line 114: Maybe replace it with something like
```
# somewhere above
_RE_HOSTNAMES = re.compile(r"^(?:ip-)?(\d+)-(\d+)-(\d+)-(\d+)")
...
# here
if _RE_HOSTNAMES.match(host):
```
?

paasta_tools/contrib/check_orphans.py
```diff
@@ -113,12 +149,15 @@ async def gather_files(hosts: Set[str]) -> Dict[str, str]:
 
 
 def read_one_nerve_file(nerve_config: str) -> Set[InstanceTuple]:
-    services = json.loads(nerve_config)["services"]
+    nerve_config = json.loads(nerve_config)
     return {
         InstanceTuple(
-            service["host"], service["port"], service["zk_path"][len(PREFIX) :]
+            host_to_ip(nerve_config["instance_id"], service["host"]),
```
vkhromov, line 155: Could you please update the PR description explaining why we use `nerve_config["instance_id"]` as a host name.

vkhromov (changes requested):
Looks good, please see comments.

paasta_tools/contrib/check_orphans.py
chriskuehl: It's this part:

> Instead, we use the name field from the nerve registration, which for PaaSTA instances is the hostname of the PaaSTA host which we can use to fetch the configured nerve registrations.

I'll make it more explicit that those are the same thing.

paasta_tools/contrib/check_orphans.py
chriskuehl: I can move it if you think it improves readability? I am not sure moving the regex far from the function that uses it does improve readability, and the performance should be the same anyway.

paasta_tools/contrib/check_orphans.py
```diff
@@ -71,19 +74,52 @@ def get_zk_data(blacklisted_services: Set[str]) -> SmartstackData:
...
+    Because DNS resolution is relatively slow and can't be easily performed
+    using asyncio, we cheat a little and use a regex for well-formed hostnames
+    to try to guess the IP without doing real resolution.
+
+    A fallback is needed because in some cases the nerve registration does not
+    match an actual hostname (e.g. "prod-db15" or "prod-splunk-master").
+    """
+    for match in (
+        re.match(r"^(\d+)-(\d+)-(\d+)-(\d+)-", host),
+        re.match(r"^ip-(\d+)-(\d+)-(\d+)-(\d+)$", host),
```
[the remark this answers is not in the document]
chriskuehl: Yes, the names in the nerve registrations are something like `ip-10-69-42-150` (not fully qualified and not always actually resolvable). I can relax the regex though in case this changes in the future.

paasta_tools/contrib/check_orphans.py
vkhromov: I especially dislike that there are two regexes when one is more than enough 🙂 
Performance will be better since here you're compiling regexes on every function call, while by using a pre-compiled regex you'll do that only once.

chriskuehl: I don't feel strongly about this but I feel like this is playing regex golf at the expense of readability; these are targeting two different hostname formats and I think using two regexes is appropriate. After your proposed change it's not as obvious what the two hostname formats are.

The pre-compiling shouldn't matter due to the internal regex cache: https://docs.python.org/3/library/re.html#re.compile

If you still think it is worth doing I will do it though.

paasta_tools/contrib/check_orphans.py
vkhromov: The code above uses `instance_data["name"]`, which is obviously the name field.
But here it's `nerve_config["instance_id"]`.  Is it the name field as well?  Why it's called `instance_id` then, if it's "name"? And etc. 🙂  
This is confusing.
Without an additional explanation I'm reading this code like we're trying to resolve the instance name as a host name.

paasta_tools/contrib/check_orphans.py
vkhromov: > The pre-compiling shouldn't matter due to the internal regex cache: https://docs.python.org/3/library/re.html#re.compile

good to know

ok, let's resolve it then.

paasta_tools/contrib/check_orphans.py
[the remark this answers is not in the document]
vkhromov: If it works then I'm fine with it.

paasta_tools/contrib/check_orphans.py
chriskuehl: I agree that the naming is confusing but that comes from what nerve does): https://github.com/yelp/nerve#configuration

I'll add a comment in the code in addition to the PR description to clarify.

vkhromov (approved):

PR #2901, opened 2020-08-28
Update Spark Run To Work with Session Token and Credential Process

smrojas:
This PR does the following:

- updates the version of boto3 and botocore to support when a credential_process is defined for a profile
- updates `get_aws_credentials` method and helper functions so that it no longer fails if a a profile configured with either a `credential_process` or a `aws_session_token`. For the former, the boto3 API returns the values as expected. In the case where a profile is configure with the session token inline, added support to handle loading and reading this key from a yaml file.

**Testing**
Automated testing
`make itest` passes

Added two new tests:

1. Check for `aws_session_token` when reading from a passed yaml file
2. Verify that we read the correct profile name when passing `profile_name` to `get_aws_credentials` rather than falling back to the default profile

I also had to fix the tests because the `args` getting passed to `get_aws_credentials` were not being respected.

Manual testing
_profile that has credential process_
`paasta spark-run --cmd /bin/bash --aws-profile dev-read-only` works when I have a credential process defined in my `~/.aws/credentials`

I get authed via the credential process as spark-run starts.

_manually specifying the aws yaml file (parsing the keys)_
`paasta spark-run --cmd /bin/bash --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml`

Keys are statically defined inline and the correct identity is used

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -413,6 +413,7 @@ def get_spark_config(
     volumes,
     access_key,
     secret_key,
+    session_token,
 ):
     # User configurable Spark options
     user_args = {
@@ -428,7 +429,9 @@ def get_spark_config(
     }
 
     default_event_log_dir = get_default_event_log_dir(
-        access_key=access_key, secret_key=secret_key
+        access_key=access_key,
+        secret_key=secret_key,
+        session_token=session_token,
     )
     if default_event_log_dir is not None:
         user_args["spark.eventLog.enabled"] = "true"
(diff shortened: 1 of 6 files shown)


vkhromov:
I have no idea what's happening here 🙂 , but there are `kiam` and AWS pod identity which could be used to assign AWS roles to PaaSTA k8s containers.

danielpops:
@vkhromov this is for okta federated roles, which have a trust relationship to okta and I don't think we want to involve kiam or scope the containers to the role like that

paasta_tools/spark_tools.py
```diff
@@ -95,13 +101,12 @@ def get_default_event_log_dir(**kwargs) -> str:
         spark_run_conf = {}
 
     try:
-        account_id = (
```
danielpops, line 98: This is referenced below, you shouldn't just return here

smrojas: yeah this mistake was caught in the tests and my next update will fix it

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -413,6 +413,7 @@ def get_spark_config(
     volumes,
     access_key,
     secret_key,
+    session_token,
```
lydian, line 416: should we make this optional? 
Otherwise, you will need to modify this line https://sourcegraph.yelpcorp.com/Yelp/paasta@31a946cfbf28b8b11382fc5c688bf64317fc05fc/-/blob/paasta_tools/cli/cmds/spark_run.py#L678

smrojas:
The plan is is to just support temp creds (since there is some manual parsing of config files rather than solely relying on boto3 library. We don't expect everyone to be using temp creds, and its not even supported everywhere yet. This is just to get the ball rolling, and moving over to temp creds will be a separate initiative

paasta_tools/cli/cmds/spark_run.py
smrojas: Hm, isn't that file just a mirror of this code? It's modified on line `688` below.

lydian: ah, you're right. some code are folded, and I misread the code.

PR #2902, opened 2020-08-31
Expand proxy_port range down to 19000. PAASTA-16947

EvanKrall:


paasta_tools/cli/fsm/autosuggest.py
```diff
@@ -57,9 +57,9 @@ def get_inuse_ports_from_etc_services():
 
 
 def suggest_smartstack_proxy_port(
-    yelpsoa_config_root, range_min=20000, range_max=21000
```
kawaiwanyelp, line 60: Do we need to do this? In our soa-configs repo, we call this function in our `Makefile`, and we can just set the range there manually instead of having to change PaaSTA.

EvanKrall: We also call this from [fsm_cmd.py](https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/fsm_cmd.py#L79), so I think this is the better place to do it.

PR #2906, opened 2020-09-02
Use default git remote in config for rightsizer_soaconfigs_update.py

qui:
The jenkins job currently hardcodes sysgit, after this change is released I can remove it so it uses the default from the config.


kawaiwanyelp (approved):
lgtm, thanks for catching this! I suppose the next step will be to create a PR to remove the arg from Jenkinsfiles?

qui:
yep, i'll update the Jenkinsfile after this is released.

PR #2908, opened 2020-09-04
Fix tron_tools to work with aws access token

lydian:
it looks like last time we modified the get_aws_credentials, we forget to update tron_tools. This changes fixed the issue. 

I've ensure: 
- make test pass
- `git grep get_aws_credentials` and ensured everything in the repo is expecting three output instead of two.


drmorr0 (approved):

smrojas (approved):
Thanks for this! Sorry I missed it initially

PR #2912, opened 2020-09-13
Use `extrapolation="last_value"` for custom signalflow used by HPA

stug:
Without this, missing data can cause the signalfx metrics adapter to request 0 instances


askedrelic (approved):

matthewbentley (approved):
🎉

PR #2914, opened 2020-09-14
Add option for gunicorn worker timeouts to the api

qui:
docs: https://docs.gunicorn.org/en/stable/settings.html#timeout


askedrelic:
shipit

PR #2915, opened 2020-09-15
Stop fetching events in paasta status because it is slowing down etcd

qui:
I didn't remove all the code, to leave open the possibility of adding this info back in once we understand what's going on and see if we can optimize the query. I'll add a ticket to the postmortem about making a decision on that.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2370,17 +2370,6 @@ async def get_kubernetes_app_deploy_status(
     kube_client: KubeClient,
     desired_instances: int,
 ) -> Tuple[int, str]:
-    # Try to get a real status message but we don't ever want to crash if this fails
-    try:
-        event_stream = await get_all_events_for_service(app, kube_client)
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -2398,7 +2387,7 @@ async def get_kubernetes_app_deploy_status(
         deploy_status = KubernetesDeployStatus.Stopped
     else:
         deploy_status = KubernetesDeployStatus.Running
-    return deploy_status, deploy_message
```
stug, line 2401: can we not just make it return one thing?  or is that too hard because the cli is expecting something...?

At the very least can we leave a TODO with an explanation?

stug (approved):
looks good to me

paasta_tools/kubernetes_tools.py
qui: Yeah, there are a few function calls in between - I just didn't want to edit them all if we are going to bring the deploy message back. Added a comment.

PR #2916, opened 2020-09-21
Skip creating the HPA if it is not required

qui:


matthewbentley (approved):

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -285,13 +285,14 @@ def sync_horizontal_pod_autoscaler(self, kube_client: KubeClient) -> None:
         In order for autoscaling to work, there needs to be at least two configurations
         min_instnace, max_instance, and there cannot be instance.
         """
-        self.logging.info(
-            f"Syncing HPA setting for {self.item.metadata.name}/name in {self.item.metadata.namespace}"
-        )
+        should_have_hpa = self.should_have_hpa()
```
stug, line 288: What if instead of this we just moved the logic in `should_have_hpa` into `get_autoscaling_metric_spec` and just had that method return `None` if there shouldn't be an hpa?  I think part of why this bug exists is that this logic is split across two modules, meaning that it's possible to call the canonical "create me an HPA" method for Deployments that shouldn't have HPAs and have it return something.

All the logic that makes up `should_have_hpa` is just method calls on `self.soa_config` anyway.

stug (commented):
I am open to argument on this, but I think we can clean up this module a little bit by moving some of the logic back into `kubernetes_tools` (which admittedly is kind of a monster, but I think the current division of responsibilities between them doesn't make sense and makes it hard to think about what's going on).

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -269,56 +269,43 @@ def update_related_api_objects(self, kube_client: KubeClient) -> None:
...
-        self.logging.info(
-            f"Syncing HPA setting for {self.item.metadata.name}/name in {self.item.metadata.namespace}"
-        )
-        hpa_exists = self.exists_hpa(kube_client)
-        if hpa_exists:
-            if not self.should_have_hpa():
-                # Remove HPA if autoscaling is disabled
-                self.delete_horizontal_pod_autoscaler(kube_client)
-                return
-            elif autoscaling_is_paused():
```
qui, line 297: Note I'm changing the behavior when autoscaling is paused.

Before we would set min_instances, so that when `self.soa_config.get_autoscaling_metric_spec` is called after that, the HPA would still be created, but not scale below the current instance count.

I think if we want to leave all the logic to the deploy config, we shouldn't have this interaction. Deleting the HPA will also have the effect of not changing the replica count. I think this is fine, based on this comment in the PR where pausing was implemented: https://github.com/Yelp/paasta/pull/2731#issuecomment-601401601

paasta_tools/kubernetes/application/controller_wrappers.py
qui: OK, I've tried to streamline the logic - let me know what you think.

paasta_tools/kubernetes/application/controller_wrappers.py
stug: Yeah I think that is cleaner -- it never really sat that well with me that we were editting the config class here.

paasta_tools/kubernetes_tools.py
```diff
@@ -570,18 +570,25 @@ def namespace_external_metric_name(self, metric_name: str) -> str:
...
+            return None
+
+        autoscaling_params = self.get_autoscaling_params()
+        if autoscaling_params["decision_policy"] == "bespoke":
+            return None
+
         # use new autoscaling configuration if it exists.
         if "horizontal_autoscaling" in self.config_dict:
             return self.get_hpa_metric_spec(name, cluster, namespace)
+
```
stug, line 586: Nit: can we change the order of these two blocks so that we check the decision policy after we check for the `horizontal_autoscaling` key?  The `decision_policy` only applies to the "legacy" `autoscaling` section, and this is meant to defer to the `horizontal_autoscaling` section if it exists.  So it doesn't really make sense to check a key in the legacy `autoscaling` section and then defer to `horizontal_autoscaling`.

paasta_tools/kubernetes_tools.py
```diff
@@ -570,18 +570,25 @@ def namespace_external_metric_name(self, metric_name: str) -> str:
...
         min_replicas = self.get_min_instances()
         max_replicas = self.get_max_instances()
-        if not min_replicas or not max_replicas:
-            log.error(
-                f"Please specify min_instances and max_instances for autoscaling to work: {min_replicas}, {max_replicas}"
```
stug, line 580: why remove this?  I guess just because no one's ever going to see it?

stug (approved):
Thank you for doing the small refactor -- that section reads a lot more cleanly now.

paasta_tools/kubernetes_tools.py
qui: We're calling this method on every update, so it's not really an error if a service just has `instances` instead of min/max.

paasta_tools/kubernetes_tools.py
qui: Sure, done.

PR #2920, opened 2020-09-23
Do not return hpa if min or max is 0

qui:
error from setup_kubernetes_job:

```
"message": "HorizontalPodAutoscaler.autoscaling \"ad--landscape-main\" is invalid: [spec.minReplicas: Invalid value: 0: must be greater than or equal to 1, spec.maxReplicas: Invalid value: 0: must be greater than 0, spec.metrics: Forbidden: must specify at least one Object or External metric to support scaling to zero replicas]",
```

I just changed this yesterday for no good reason, we can change it back.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -576,6 +576,9 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         if self.get_desired_state() == "stop":
             return None
 
+        if not self.is_autoscaling_enabled():
+            return None
+
         # use new autoscaling configuration if it exists.
         if "horizontal_autoscaling" in self.config_dict:
             return self.get_hpa_metric_spec(name, cluster, namespace)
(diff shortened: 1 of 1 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -586,7 +586,7 @@ def get_autoscaling_metric_spec(
 
         min_replicas = self.get_min_instances()
         max_replicas = self.get_max_instances()
-        if min_replicas is None or max_replicas is None:
+        if not min_replicas or not max_replicas:
```
stug, line 589: Can we either add a comment saying that we are handling both 0 and None here or (even better) make that explicit somehow in the code by checking for both None and 0?

My impulse is usually the same as yours -- when I see something relying on implicit conversion to boolean I switch it to `is not None`, but here we are (implicitly) dealing with 0s.

stug (approved):

paasta_tools/kubernetes_tools.py
qui: OK, I separated the checks.

get_min_instances returns `get('min_instances', 1)`, so it shouldn't be None. And `is_autoscaling_enabled` checks if max_instances is 0.

PR #2921, opened 2020-09-24
Support PaaSTA secrets as files

vkhromov:
Add `secret_volume` option into yelpsoa, which allows to mount an                                                      
existing secret as a file into an instance.                                                                            
                                                                                                                       
Example:                                                                                                               
```                                                                                                                    
main:                                                                                                                  
  secret_volumes:                                                                                                      
  - container_path: /home/nobody/.ssh                                                                                  
    secret_name: zzz_password                                                                                          
    default_mode: 0444                                                                                                 
    items:                                                                                                             
    - key: rsa-private                                                                                                 
      path: id_rsa                                                                                                     
    - key: rsa-public                                                                                                  
      path: id_rsa.pub                                                                                                 
      mode: 0666                                                                                                       
```                                                                                                                    
                                                                                                                       
As a result, the content of the `rsa-private` key from the
`zzz_password` secret will be accessible via the
`/home/nobody/.ssh/id_rsa` file with `0444` mode, and the content of
`rsa-public` key from the same secret will be accessible via
`/home/nobody/.ssh/id_rsa.pub` file with `0666` mode.

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -505,6 +505,50 @@
                     },
                     "uniqueItems": true
                 },
+                "secret_volumes": {
+                    "type": "array",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "container_path": {
+                                "type": "string"
+                            },
+                            "secret_name": {
+                                "type": "string"
+                            },
+                            "default_mode": {
+                                "type": "string"
+                            },
+                            "items": {
+                                "type": "array",
+                                "items": {
+                                    "type": "object",
+                                    "properties": {
+                                        "key": {
+                                            "type": "string"
+                                        },
+                                        "path": {
+                                            "type": "string"
+                                        },
+                                        "mode": {
+                                            "type": "string"
+                                        }
+                                    },
+                                    "required": [
+                                        "key",
+                                        "path"
+                                    ]
+                                },
+                                "uniqueItems": true
+                            }
+                        },
+                        "required": [
+                            "container_path",
+                            "secret_name"
+                        ]
+                    },
+                    "uniqueItems": true
+                },
                 "replication_threshold": {
                     "type": "integer",
                     "minimum": 0
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -228,6 +228,19 @@ class PersistentVolume(VolumeWithMode):
     storage_class_name: str
 
 
+class SecretVolumeItem(TypedDict, total=False):
+    key: str
+    path: str
+    mode: Union[str, int]
+
+
+class SecretVolume(TypedDict, total=False):
+    secret_name: str
+    container_path: str
+    default_mode: Union[str, int]
+    items: List[SecretVolumeItem]
+
+
 class InstanceConfigDict(TypedDict, total=False):
     deploy_group: str
     mem: float
@@ -247,6 +260,7 @@ class InstanceConfigDict(TypedDict, total=False):
     role: str
     extra_volumes: List[DockerVolume]
     aws_ebs_volumes: List[AwsEbsVolume]
+    secret_volumes: List[SecretVolume]
     security: SecurityConfigDict
     dependencies_reference: str
     dependencies: Dict[str, Dict]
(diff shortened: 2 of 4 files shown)


paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -505,6 +505,21 @@
                     },
                     "uniqueItems": true
                 },
+                "secret_volumes": {
+                    "type": "array",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "container_path": {
```
qui, line 513: The docs indicate this has to be a directory, so maybe container_dir would be better?

https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-files-from-a-pod

So the config would be something like:
```
  - container_dir: /home/nobody/.ssh/
    secret_name: id_rsa
```
And you'd have to make sure the secret name matches the filename you want it to be, I guess.

paasta_tools/utils.py
```diff
@@ -228,6 +228,11 @@ class PersistentVolume(VolumeWithMode):
     storage_class_name: str
 
 
+class SecretVolume(VolumeWithMode):
```
qui, line 231: nit: it looks like we don't use the mode and always set `read_only=True`, so this could just inherit from TypedDict

qui (commented):
Ideally, you wouldn't have to define a secret as an env variable to be able to use it as a file. But, reading the secret sync code, I don't see why that is a requirement. It seems like would work without the environment variable. Am I missing something?

paasta_tools/cli/schemas/kubernetes_schema.json
vkhromov: Fixed by introducing `items`

vkhromov:
> Ideally, you wouldn't have to define a secret as an env variable to be able to use it as a file. But, reading the secret sync code, I don't see why that is a requirement. It seems like it would work without the environment variable. Am I missing something?

Something still need to create `V1Secret` object in Kubernetes.
We already have a code which creates these `V1Secret` object for environment variables.
Here I'm just reusing them for secret volumes as well.

We can try to start with this approach and improve it later.

vkhromov:
## Test
with
```
  secret_volumes:
  - container_path: /home/nobody/.ssh
    secret_name: new-test-secret
    default_mode: 0444
    items:
        # The whole secret and the only key inside it are both named
        # `new-test-secret`, that's why that name is in two places in the
        # `secret_volumes` section.
    - key: new-test-secret
      path: secret_file
      mode: 0666
```
the key `new-test-secret` from the secret `new-test-secret` is available as `/home/nobody/.ssh/secret_file` file with corresponding permissions: 
```bash
www-data@compute-infra-test-service-main--k8s-74665dff4d-2jrgc:/$ ls -la /home/nobody/.ssh/
total 0
drwxrwxrwt 3 root root 100 Oct 14 13:47 .
drwxr-xr-x 3 root root  18 Oct 14 13:47 ..
drwxr-xr-x 2 root root  60 Oct 14 13:47 ..2020_10_14_13_47_30.973516100
lrwxrwxrwx 1 root root  31 Oct 14 13:47 ..data -> ..2020_10_14_13_47_30.973516100
lrwxrwxrwx 1 root root  18 Oct 14 13:47 secret_file -> ..data/secret_file


www-data@compute-infra-test-service-main--k8s-74665dff4d-2jrgc:/$ ls -la $(realpath /home/nobody/.ssh/secret_file)
-rw-rw-rw- 1 root root 9 Oct 14 13:47 /home/nobody/.ssh/..2020_10_14_13_47_30.973516100/secret_file


www-data@compute-infra-test-service-main--k8s-74665dff4d-2jrgc:/$ cat /home/nobody/.ssh/secret_file
FRIDAY!!!
```

qui:
@vkhromov Right, I think this is the code that creates the V1Secret objects, right? https://github.com/Yelp/paasta/blob/29415bb791b68e50676d5e4f125900533e3a16d0/paasta_tools/kubernetes/bin/paasta_secrets_sync.py#L160

That iterates through the secrets dir, so I don't think it depends on the secret being defined in an env variable.

qui (approved):

PR #2922, opened 2020-09-25
script to track and auto-apply Puppet-created kubernetes resources

drmorr0:
## Description

The goal here is to replace the cron'ed `kubectl apply` in puppet with something a little bit more intelligent.  This script creates a `.applied` directory inside `/etc/kubernetes/puppet_resources`, which tracks what's previously been applied to the cluster.  New files or changed files will get re-applied.  Things that have previously been deleted will thus exist in `.applied` but not in `puppet_resources`, and the script will run `kubectl delete` on such objects.

The script will only run on the kubernetes leader, and will rsync the `.applied` folder to the other masters.

All the operations in here are idempotent (e.g., if `kubectl apply` runs a second time, then it just won't make any changes), so this should be robust in case something goes wrong.  If the script _does_ find some errors it will exit 1, which I'll configure puppet to create a ticket for someone to manually investigate.

## Testing done

Ran manually on kubestage
Added some tests via `pyfakefs`

### DIFF
--- a/paasta_tools/apply_external_resources.py
+++ b/paasta_tools/apply_external_resources.py
@@ -1,0 +1,76 @@
+#!/opt/venvs/paasta-tools/bin/python
+import os
+import sys
+from filecmp import cmp
+from shutil import copy
+from subprocess import CalledProcessError
+from subprocess import run
+from traceback import print_exc
+
+APPLIED_DIRECTORY = ".applied"
+
+
+# This script expects the KUBECONFIG environment variable to be set correctly
+def main(puppet_resource_root: str) -> int:
+    exit_code = 0
+    applied_resource_root = os.path.join(puppet_resource_root, APPLIED_DIRECTORY)
+
+    # Loop through everything in the puppet resource path
+    for root, dirs, files in os.walk(puppet_resource_root):
+        # modifying the 'dirs' variable in-place will update the order that os.walk visits things
+        if APPLIED_DIRECTORY in dirs:
+            dirs.remove(APPLIED_DIRECTORY)
+        dirs.sort()  # Need to apply things in the correct order
+
+        # Check to see if there's a difference between what Puppet created and
+        # what's been previously applied
+        for filename in sorted([f for f in files if f.endswith(".yaml")]):
+            path = os.path.join(root, filename)
+            applied_path = os.path.join(
+                applied_resource_root, os.path.relpath(path, puppet_resource_root)
+            )
+            print(f"comparing {path} and {applied_path}")
+            if not os.path.exists(applied_path) or not cmp(
+                path, applied_path, shallow=False
+            ):
+                # This is idempotent; if something gets out of sync and a resource gets applied
+                # a second time, kubectl just won't make any changes
+                try:
+                    run(["kubectl", "apply", "-f", path], check=True)
+                    os.makedirs(os.path.dirname(applied_path), exist_ok=True)
+                    copy(path, applied_path)
+                except CalledProcessError:
+                    print(f"There was a problem applying {path}:\n")
+                    print_exc(
+                        file=sys.stdout
+                    )  # keep all messages on the same stream so they're in order
+                    exit_code = 1
+                    continue
+
+    # Loop through all the files that have been previously applied and see
+    # if Puppet has removed any of them
+    for root, dirs, files in os.walk(applied_resource_root):
+        dirs.sort(reverse=True)  # for deleting things, we need to go in reverse order
+        for filename in sorted([f for f in files if f.endswith(".yaml")], reverse=True):
+            path = os.path.join(root, filename)
+            puppet_path = os.path.join(
+                puppet_resource_root, os.path.relpath(path, applied_resource_root)
+            )
+            if not os.path.exists(puppet_path):
+                print(f"Deleting resource {path}...")
+                try:
+                    run(["kubectl", "delete", "-f", path], check=True)
+                    os.remove(path)
+                except CalledProcessError:
+                    print(f"There was a problem deleting {path}:\n")
+                    print_exc(
+                        file=sys.stdout
+                    )  # keep all messages on the same stream so they're in order
+                    exit_code = 1
+                    continue
+
+    return exit_code
+
+
+if __name__ == "__main__":
+    sys.exit(main(sys.argv[1]))
--- a/requirements-dev-minimal.txt
+++ b/requirements-dev-minimal.txt
@@ -8,6 +8,7 @@ mock
 path.py
 pep8
 pre-commit
+pyfakefs
 pylint
 pytest
 pytest-asyncio
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -22,6 +22,7 @@ pep8==1.5.7
 pluggy==0.6.0
 pre-commit==1.18.0
 pycodestyle==2.3.1
+pyfakefs==4.1.0
 pyflakes==1.6.0
 Pygments==2.2.0
 pylint==1.7.2
--- a/tests/test_apply_external_resources.py
+++ b/tests/test_apply_external_resources.py
@@ -1,0 +1,150 @@
+import os
+from subprocess import CalledProcessError
+
+import mock
+import pytest
+
+from paasta_tools.apply_external_resources import main
+
+
+@pytest.fixture
+def mock_run():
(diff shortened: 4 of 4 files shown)


paasta_tools/apply_puppet_resources.py
```diff
@@ -0,0 +1,70 @@
...
+            print(f"comparing {path} and {applied_path}")
+            if not os.path.exists(applied_path) or not cmp(
+                path, applied_path, shallow=False
+            ):
+                # This is idempotent; if something gets out of sync and a resource gets applied
+                # a second time, kubectl just won't make any changes
+                try:
+                    run(["kubectl", "apply", "-f", path], check=True)
+                    os.makedirs(applied_path[: -len(filename)], exist_ok=True)
+                    copy(path, applied_path)
```
stug, line 40: If the new version removes a field from what was previously there, I don't think this will catch it.

[this page](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#how-to-update-objects) implies that maybe `kubectl diff` will make that possible...?

stug (approved):
If fixing the removed fields issue is too tricky, feel free to ship this and leave PAASTA-16317 open

paasta_tools/apply_puppet_resources.py
keymone, line 40: puppet is purging files/folders in the resources directory: https://sourcegraph.yelpcorp.com/sysgit/puppet@4d099dd9af8c980491242e60735e9c973b4749a5/-/blob/modules/profile_kube/manifests/master_resources.pp#L9, you may want to keep it in `/etc/kubernetes/puppet_resources.applied` or something?

keymone (changes requested):
~~lgtm, but see comment~~

does this really belong in paasta repo? those are system resources applied by puppet and managed in puppet repo, so let's have this script in puppet repo?

vkhromov (changes requested):
Looks good, please see comments.

drmorr0:
> ~lgtm, but see comment~
> 
> does this really belong in paasta repo? those are system resources applied by puppet and managed in puppet repo, so let's have this script in puppet repo?

Discussed on slack; we decided to keep the script here, but rename it as `apply_external_resources.py` to make it less Puppet-specific.  This also lets us write some tests (which I will do now).

drmorr0:
The tests I added are... really more "integration" tests than "unit" tests if we care about that sort of thing... but it's a small/simple-enough script that I felt like refactoring to have discrete, easily-unit-testable-functions was more work than it was worth.

paasta_tools/apply_external_resources.py
```diff
@@ -0,0 +1,76 @@
...
+def main(puppet_resource_root: str) -> int:
+    exit_code = 0
+    applied_resource_root = os.path.join(puppet_resource_root, APPLIED_DIRECTORY)
+
+    # Loop through everything in the puppet resource path
+    for root, dirs, files in os.walk(puppet_resource_root):
+        # modifying the 'dirs' variable in-place will update the order that os.walk visits things
+        if APPLIED_DIRECTORY in dirs:
+            dirs.remove(APPLIED_DIRECTORY)
+        dirs.sort()  # Need to apply things in the correct order
```
vkhromov, line 23: Please correct me if I'm wrong, but it doesn't seem that `dirs` variable is used anywhere?

paasta_tools/apply_external_resources.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Check to see if there's a difference between what Puppet created and
+        # what's been previously applied
+        for filename in sorted([f for f in files if f.endswith(".yaml")]):
```
vkhromov, line 27: Here we sort files inside a single directory, but full paths are still in arbitrary order.
We need to sort full paths for all files from all directories.

paasta_tools/apply_external_resources.py
```diff
@@ -0,0 +1,76 @@
...
+                        file=sys.stdout
+                    )  # keep all messages on the same stream so they're in order
+                    exit_code = 1
+                    continue
+
+    # Loop through all the files that have been previously applied and see
+    # if Puppet has removed any of them
+    for root, dirs, files in os.walk(applied_resource_root):
+        dirs.sort(reverse=True)  # for deleting things, we need to go in reverse order
+        for filename in sorted([f for f in files if f.endswith(".yaml")], reverse=True):
```
vkhromov, line 54: The same as above: `dirs` doesn't seem to be used, full paths for files from all directories should be sorted.

vkhromov (changes requested):

paasta_tools/apply_external_resources.py
drmorr0: `dirs` is used by `os.walk` -- if you modify it in place, it will change the order in which directories are visited:

> When topdown is True, the caller can modify the dirnames list in-place (perhaps using del or slice assignment), and walk() will only recurse into the subdirectories whose names remain in dirnames; this can be used to prune the search, impose a specific order of visiting, or even to inform walk() about directories the caller creates or renames before it resumes walk() again.

(from the Python documentation: https://docs.python.org/3.8/library/os.html#os.walk)

paasta_tools/apply_external_resources.py
drmorr0: see above.

paasta_tools/apply_external_resources.py
drmorr0: see above.

paasta_tools/apply_external_resources.py
vkhromov: Good to know!
Resolving it.

paasta_tools/apply_external_resources.py
```diff
@@ -0,0 +1,76 @@
...
+from subprocess import run
+from traceback import print_exc
+
+APPLIED_DIRECTORY = ".applied"
+
+
+# This script expects the KUBECONFIG environment variable to be set correctly
+def main(puppet_resource_root: str) -> int:
+    exit_code = 0
+    applied_resource_root = os.path.join(puppet_resource_root, APPLIED_DIRECTORY)
```
stug, line 16: Did we fix the issue where puppet might purge the `.applied` directory?

paasta_tools/apply_external_resources.py
```diff
@@ -0,0 +1,76 @@
...
+            )
+            print(f"comparing {path} and {applied_path}")
+            if not os.path.exists(applied_path) or not cmp(
+                path, applied_path, shallow=False
+            ):
+                # This is idempotent; if something gets out of sync and a resource gets applied
+                # a second time, kubectl just won't make any changes
+                try:
+                    run(["kubectl", "apply", "-f", path], check=True)
+                    os.makedirs(applied_path[: -len(filename)], exist_ok=True)
```
stug, line 40: nit: might be better to use `os.path.dirname` for clarity

paasta_tools/apply_external_resources.py
```diff
@@ -0,0 +1,76 @@
...
+        if APPLIED_DIRECTORY in dirs:
+            dirs.remove(APPLIED_DIRECTORY)
+        dirs.sort()  # Need to apply things in the correct order
+
+        # Check to see if there's a difference between what Puppet created and
+        # what's been previously applied
+        for filename in sorted([f for f in files if f.endswith(".yaml")]):
+            path = os.path.join(root, filename)
+            applied_path = os.path.join(
+                applied_resource_root, path[len(puppet_resource_root) + 1 :]
```
stug, line 30: nit: might be better to use `os.path.relpath` for clarity

paasta_tools/apply_puppet_resources.py
drmorr0: As best as I can tell, `kubectl apply` handles fields that get removed.

stug (approved):
lg2m, just want to make sure that the puppet purging isn't going to be an issue.

Some nits about using the stuff in `os.path` for the path manipulations but eh.

paasta_tools/apply_external_resources.py
drmorr0: I'm going to have Puppet create and manage the `.applied` directory.

paasta_tools/apply_external_resources.py
drmorr0: Thanks, I forgot this existed

PR #2924, opened 2020-09-28
Unify error interface between bravado and oapi clients

keymone:
this is a copy of #2923, i merged that into wrong branch

* remove pkg_resources for version lookup because we have version constant in paasta_tools module
* add error type objects to oapi and swagger api clients
* remove explicit imports of bravado for error types


keymone:
diff is equivalent to #2923, will assume that same shipits apply

PR #2925, opened 2020-09-28
Updating annotation key of dashboard_url for Flink services

AshishKhatkar:
These changes are relevant for allowing large instance names for Flink services


paasta_tools/setup_kubernetes_cr.py
```diff
@@ -281,10 +277,10 @@ def format_custom_resource(
         "spec": instance_config,
     }
 
-    url = get_dashboard_url(kind, service, instance, cluster)
+    url = get_dashboard_base_url(kind, cluster)
     if url:
-        resource["metadata"]["annotations"]["yelp.com/dashboard_url"] = url
-        resource["metadata"]["annotations"][paasta_prefixed("dashboard_url")] = url
+        resource["metadata"]["annotations"]["yelp.com/dashboard_base_url"] = url
+        resource["metadata"]["annotations"][paasta_prefixed("dashboard_base_url")] = url
```
poros, line 283: Have you checked with #paasta that they are happy with you changing the name of this common annotation?

AshishKhatkar: This annotation is only used for Flink services. The code path that the changes touch is limited to Flink.

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -281,10 +277,10 @@ def format_custom_resource(
         "spec": instance_config,
     }
 
-    url = get_dashboard_url(kind, service, instance, cluster)
+    url = get_dashboard_base_url(kind, cluster)
     if url:
-        resource["metadata"]["annotations"]["yelp.com/dashboard_url"] = url
-        resource["metadata"]["annotations"][paasta_prefixed("dashboard_url")] = url
+        resource["metadata"]["annotations"]["yelp.com/dashboard_base_url"] = url
```
keymone, line 282: drop this one, yelp.com/attribute is deprecated scheme

paasta_tools/cli/cmds/status.py
```diff
@@ -974,7 +974,7 @@ def print_flink_status(
         )
     else:
         output.append(f"    Flink version: {status.config['flink-version']}")
-    dashboard_url = metadata.annotations.get(paasta_prefixed("dashboard_url"))
+    dashboard_url = metadata.annotations.get("flink.yelp.com/dashboard_url")
```
keymone, line 977: add a comment that this attribute is populated by flink operator

keymone (changes requested):
fix'n'ship

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -281,10 +277,9 @@ def format_custom_resource(
         "spec": instance_config,
     }
 
-    url = get_dashboard_url(kind, service, instance, cluster)
+    url = get_dashboard_base_url(kind, cluster)
     if url:
-        resource["metadata"]["annotations"]["yelp.com/dashboard_url"] = url
-        resource["metadata"]["annotations"][paasta_prefixed("dashboard_url")] = url
+        resource["metadata"]["annotations"][paasta_prefixed("dashboard_base_url")] = url
```
poros, line 282: why don't we prefix this one for flink like `dashboard_url` if we are the only ones using it?

tests/test_setup_kubernetes_cr.py
```diff
@@ -237,9 +237,9 @@ def test_format_custom_resource():
                 },
                 "annotations": {
                     "yelp.com/desired_state": "running",
-                    "yelp.com/dashboard_url": "http://flink.k8s.paasta-mycluster.yelp:31080/kurupt--fm-radio--station",
+                    "yelp.com/dashboard_base_url": "http://flink.k8s.paasta-mycluster.yelp:31080/",
```
poros, line 240: haven't you dropped this one?

paasta_tools/setup_kubernetes_cr.py
AshishKhatkar: My idea is that any label or annotation that comes from PaaSTA should be prefixed with PaaSTA as opposed to Flink. For the annotation operator is adding it should have Flink prefix because that annotation is added by operator not PaaSTA.

tests/test_setup_kubernetes_cr.py
AshishKhatkar: I will update. I missed this.

paasta_tools/setup_kubernetes_cr.py
poros: I believe that the guidance that @Rob-Johnson set is to group them by project, not by source: https://yelpwiki.yelpcorp.com/display/COMPUTE/Operator+development+guidelines#Operatordevelopmentguidelines-Labelsandannotations

AshishKhatkar: I thought this was limited to labels and annotations attached by operator. Why I would prefer it to be PaaSTA prefixed is because this annotation is added by PaaSTA, if we face any problem we can directly point to the source (in this case PaaSTA) which adds those annotations/labels. Also, how I look at PaaSTA annotations/labels is that operators will not update them, if operator needs to make an update they can add their own annotation with modification.

I am happy to update it if you think the reasoning is not justified.

poros: It's reasonable and I can live with it. Let's move ahead.

poros (approved):

keymone (approved):

PR #2928, opened 2020-09-30
Updating flink service check to point to correct dashboard url

AshishKhatkar:
Fixing the flink service health check based on changes released in #2925

Changes tested on infrastage : https://fluffy.yelpcorp.com/i/WHmQ80fvg45tZczQB1nQK755VG6tHkn6.html

### DIFF
--- a/paasta_tools/check_flink_services_health.py
+++ b/paasta_tools/check_flink_services_health.py
@@ -142,7 +142,7 @@ def check_flink_service_health(
     num_healthy_jobmanagers = healthy_flink_containers_cnt(si_pods, "jobmanager")
     num_healthy_taskmanagers = healthy_flink_containers_cnt(si_pods, "taskmanager")
 
-    servie_cr_name = get_cr_name(si_pods)
+    service_cr_name = get_cr_name(si_pods)
 
     results = [
         check_under_replication(
@@ -166,7 +166,7 @@ def check_flink_service_health(
         check_under_registered_taskmanagers(
             instance_config=instance_config,
             expected_count=taskmanagers_expected_cnt,
-            cr_name=servie_cr_name,
+            cr_name=service_cr_name,
         ),
     ]
     output = "\n########\n".join([r[1] for r in results])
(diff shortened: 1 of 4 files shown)


paasta_tools/check_flink_services_health.py
```diff
@@ -122,6 +142,8 @@ def check_flink_service_health(
     num_healthy_jobmanagers = healthy_flink_containers_cnt(si_pods, "jobmanager")
     num_healthy_taskmanagers = healthy_flink_containers_cnt(si_pods, "taskmanager")
 
+    servie_cr_name = get_cr_name(si_pods)
```
JoeMalt, line 145: nit: typo in `service` here

paasta_tools/check_flink_services_health.py
```diff
@@ -105,6 +108,23 @@ def check_under_registered_taskmanagers(
...
+        pod
+        for pod in si_pods
+        if pod.metadata.labels["flink.yelp.com/container-type"] == "jobmanager"
+        and is_pod_ready(pod)
+        and container_lifetime(pod).total_seconds() > 60
+    ]
+    if len(jobmanager_pod) == 1:
+        return jobmanager_pod[0].metadata.name.split("-jobmanager-")[0]
+    else:
+        return ""
```
JoeMalt, line 125: what do you think about erroring out here rather than returning an empty string?

AshishKhatkar: I handled this case in `check_under_registered_taskmanagers`.

paasta_tools/check_flink_services_health.py
AshishKhatkar: done

paasta_tools/check_flink_services_health.py
JoeMalt: Yeah I see it's handled there, I just wonder whether a more specific error message here is useful. Not a big deal though, don't block on this.

JoeMalt (approved):
lgtm

paasta_tools/check_flink_services_health.py
AshishKhatkar: Yeah, my understanding is if the jobmanager is not running we won't have running dashboard which will be pointed out by the output message handling. Also, in such cases jobmanager pod check will also fail.

gvolco (approved):

PR #2930, opened 2020-10-02
move the spark_run major logic to service_configuration_lib

lydian:
# Description
Update to use the latest `service_configuration_lib.spark_config`. The major logic in `spark_run.py`, `spark_tool.py` and `tron_tool.py` are already moved to the library.  

# Tests
- unit test pass
- manually test with paasta spark-run, I am able to run the spark jobs without issue. 
- manually test with `paasta_tools/setup_tron_namespace.py` ensure the output config is the same as the one generated from master branch.


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -364,19 +398,25 @@ def get_smart_paasta_instance_name(args):
...
     spark_env = {}
 
-    if access_key is not None:
+    access_key, secret_key, _ = aws_creds
+    if access_key:
         spark_env["AWS_ACCESS_KEY_ID"] = access_key
         spark_env["AWS_SECRET_ACCESS_KEY"] = secret_key
         spark_env["AWS_DEFAULT_REGION"] = args.aws_region
     spark_env["PAASTA_LAUNCHED_BY"] = get_possible_launched_by_user_variable_from_env()
-    spark_env["PAASTA_INSTANCE_TYPE"] = "spark"
```
jsleight, line 375: I take it the servic_configuration_lib sets this now?

jsleight (approved):
lgtm.  I only skimmed the tests since you mentioned doing e2e tests.

88manpreet (approved):

paasta_tools/cli/cmds/spark_run.py
lydian: oh, nice catch!  I found that this is for the driver environment, which is not configured in the service_configuration_lib.   
This env doesn't seems to impact the actual spark run. However, I've launched the paasta spark-run and did a comparison. The updated PR is now ensure we have exactly the same environment variable on driver.

88manpreet (approved):

jsleight (approved):

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -639,7 +522,7 @@ def get_spark_app_name(
     if spark_app_name is None:
         spark_app_name = "paasta_spark_run"
 
-    spark_app_name += "_{}_{}".format(get_username(), spark_ui_port)
+    spark_app_name += f"_{get_username()}"
```
qui, line 525: Will this break https://github.com/Yelp/paasta/blob/master/paasta_tools/check_spark_jobs.py#L169 ? If so, can you update that, too?

qui (approved):

paasta_tools/tron_tools.py
```diff
@@ -213,52 +213,46 @@ def __init__(
...
-                volumes=[
-                    f"{v['hostPath']}:{v['containerPath']}:{v['mode'].lower()}"
-                    for v in self.get_volumes(load_system_paasta_config().get_volumes())
-                ],
-                user_spark_opts=self.config_dict.get("spark_args", {}),
-                event_log_dir=get_default_event_log_dir(
-                    service=self.get_service(),
-                    aws_credentials_yaml=self.config_dict.get("aws_credentials_yaml"),
-                ),
+            mesos_leader = mesos_leader = (
```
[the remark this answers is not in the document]
lydian: oh nice catch!  cleaning this up!

paasta_tools/cli/cmds/spark_run.py
lydian: the spark_app_name will be passing to the service_configuration_lib and append ui_port and time
https://github.com/Yelp/service_configuration_lib/blob/master/service_configuration_lib/spark_config.py#L521

namely, the original code would generate 
`<spark_app_name>_<user>_<port>`
the new code is having the name 
`<spark_app_name>_<user>_<port>_<time>`
(appending `time` so that it matches the jupyterhub use case)  

Thanks for finding this out, I will fix that script!

lydian:
check_spark_jobs.py is also updated :)

qui (approved):

PR #2939, opened 2020-10-14
Limit the number of extrapolations on the service load timeseries in …

stug:
…legacy autoscaling signalflow.

Without this, when instances die we'll keep the service load timeseries (e.g. uwsgi) pegged at its last value for some time.

See PAASTA-17043 for details.

---

I chose 10 kind of arbitrarily, but it looks alright when I update the graph.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -177,7 +177,7 @@ moving_average_window = '{moving_average_window_seconds}s'
 filters = filter('paasta_service', '{paasta_service}') and filter('paasta_instance', '{paasta_instance}') and filter('paasta_cluster', '{paasta_cluster}')
 
 current_replicas = data('kube_hpa_status_current_replicas', filter=filters, extrapolation="last_value").sum(by=['paasta_cluster'])
-load_per_instance = data('{signalfx_metric_name}', filter=filters, extrapolation="last_value")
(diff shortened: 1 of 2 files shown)


matfra (approved):
Yes we've all got bitten by the combo last value and infinite extrapolation in the past. It should almost be forbidden !

paasta_tools/kubernetes_tools.py
```diff
@@ -177,7 +177,7 @@
 filters = filter('paasta_service', '{paasta_service}') and filter('paasta_instance', '{paasta_instance}') and filter('paasta_cluster', '{paasta_cluster}')
 
 current_replicas = data('kube_hpa_status_current_replicas', filter=filters, extrapolation="last_value").sum(by=['paasta_cluster'])
```
qui, line 179: We'd have to think through the implications, but I wonder if this should just be the count of load timeseries? Since we divide the summed load by this number, I think that might give us a more accurate average.

qui (approved):

paasta_tools/kubernetes_tools.py
stug: I think maybe the issue there would be that the `uwsgi` timeseries is only collected for pods that have passed their readiness check, but we need to provide a scaling target for the total number of pods for the Deployment, so it might lead to us overscaling?

PR #2940, opened 2020-10-20
make wait-for-deployment check the actually-running git SHAs, not just what's in deployments.json

EvanKrall:
PAASTA-17026

We've got a race condition currently with wait-for-deployment, which leads to it thinking deployments are done before they've really started.

If wait-for-deployment polls for status after the `yelpsoa-configs` cron updates deployments.json, but before the `setup_kubernetes_job` runs, and the old version is fully deployed, then:

- the app_count check on line 1247 will pass. (There's exactly one app -- the old app).
- the git_sha check on line 1260 will pass (since that only checks deployments.json)
- the deploy_status check on line 1272-1275 will pass (since the old version is still Running)
- the required_instance_count check on line 1297 will pass (since the old version is fully deployed).

This patch adds a new `active_shas` property to the status response, and checks this in wait-for-deployment.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -1309,6 +1309,33 @@ def _run_instance_worker(cluster_data, instances_out, green_light):
                 cluster_data.instances_queue.task_done()
                 instances_out.put(instance_config)
                 continue
+
+            # https://github.com/Yelp/paasta/pull/2940#discussion_r508334043
+            try:
+                active_shas = long_running_status.active_shas
+            except ApiAttributeError:
+                active_shas = None
+            # `is not None` is necessary here because I'm adding the new attribute to the API response at the same time as I'm adding this reference.
+            if active_shas is not None:
+                active_git_shas = {g for g, c in active_shas}
+
+                if cluster_data.git_sha not in active_git_shas:
+                    print(
+                        f"{cluster_data.service}.{instance} on {cluster_data.cluster} not yet running {cluster_data.git_sha}."
+                    )
+                    cluster_data.instances_queue.task_done()
+                    instances_out.put(instance_config)
+                    continue
+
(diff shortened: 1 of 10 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1309,6 +1309,28 @@ def _run_instance_worker(cluster_data, instances_out, green_light):
                 cluster_data.instances_queue.task_done()
                 instances_out.put(instance_config)
                 continue
+
+            # `is not None` is necessary here because I'm adding the new attribute to the API response at the same time as I'm adding this reference.
+            if long_running_status.active_shas is not None:
```
keymone, line 1314: due to the way generated code works atm, you'll need to do this:

    try:
        active_shas = long_running_status.active_shas
    except paasta_tools.paastaapi.ApiAttributeException:
        active_shas = None

i'm getting this fixed, already got PR merged into openapi generator, just didn't switch to new generator image in Makefile

paasta_tools/instance/kubernetes.py
```diff
@@ -441,11 +449,10 @@ def kubernetes_status(
     active_shas = kubernetes_tools.get_active_shas_for_service(
         [app, *pod_list, *replicaset_list]
     )
-    kstatus["app_count"] = max(
-        len(active_shas["config_sha"]), len(active_shas["git_sha"])
```
mattmb, line 445: Hmm, is this not important for bounces that just change config? I guess we don't wait-for-deployment in those cases?

mattmb (approved):
I haven't read the code in detail but the problem and description sound right if we really are checking the desired sha in the status call.

keymone (approved):
lgtm

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: Interesting. When I test my branch, I get None. I'll add this snippet in to be cautious.

PR #2942, opened 2020-10-20
Add pod_name and pod_ip to tmp_paasta_allocation

sergsyelp:
Add pod_name and pod_ip to tmp_paasta_allocation to create a direct mapping between service / instance and pod name / ip

[Example get_running_task_allocation output](https://fluffy.yelpcorp.com/i/QjhwT9X7W7MR8zntcx9QsPDP99HSzkmK.html)

### DIFF
--- a/paasta_tools/contrib/get_running_task_allocation.py
+++ b/paasta_tools/contrib/get_running_task_allocation.py
@@ -33,6 +33,8 @@ class TaskAllocationInfo(NamedTuple):
     resources: Mapping[str, float]
     start_time: float
     docker_id: str
+    pod_name: str
+    pod_ip: str
     mesos_container_id: str  # Because Mesos task info does not have docker id
 
 
@@ -95,6 +97,8 @@ async def get_mesos_task_allocation_info() -> Iterable[TaskAllocationInfo]:
                 resources=task["resources"],
                 start_time=start_time,
                 docker_id=None,
+                pod_name=None,
+                pod_ip=None,
                 mesos_container_id=mesos_container_id,
             )
         )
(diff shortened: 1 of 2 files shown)


matthewbentley (approved):

matfra:
Looks good, do you think there is a value in updating tests for the functions you've modified?
https://github.com/Yelp/paasta/blob/df399ead28319bfd7af2f33d45078ee76b0b780b/tests/test_kubernetes_tools.py#L1536

sergsyelp:
> Looks good, do you think there is a value in updating tests for the functions you've modified?
> https://github.com/Yelp/paasta/blob/df399ead28319bfd7af2f33d45078ee76b0b780b/tests/test_kubernetes_tools.py#L1536

sure on it

paasta_tools/contrib/get_running_task_allocation.py
```diff
@@ -129,13 +133,15 @@ def get_kubernetes_resource_request(
 
 def get_kubernetes_metadata(
     pod: V1Pod,
-) -> Tuple[Optional[str], Optional[str], Optional[str]]:
+) -> Tuple[Optional[str], Optional[str], Optional[str], Optional[str], Optional[str]]:
     labels = pod.metadata.labels or {}
     node_selectors = pod.spec.node_selector or {}
+    pod_name = pod.metadata.name
```
matfra, line 139: Would it be safer to use a get here ? I recon it's highly unlikely that a pod doesn't have an name but I'm not 100% sure a pod can not have an IP

qui (approved):
Looks good. This should probably graduate out of contrib into a real script so we can add more tests, but we can do that later.

paasta_tools/contrib/get_running_task_allocation.py
sergsyelp: 'get' won't work because pod metadata / status objects do not support this method. pod.metadata.name and pod.status.pod_ip are actually methods, in an unlikely situation pod doesn't have a name / IP assigned these methods just return 'None', which is going to be passed as 'null' to Splunk.

matfra (approved):

sergsyelp:
!integrate

PR #2944, opened 2020-10-21
Add host_ip to get_running_task_allocation output

sergsyelp:
Add host_ip to get_running_task_allocation script output to simplify mapping of a running pod to a host.

Change in api/client.py is caused by an automated import reordering.


jfongatyelp (approved):

matfra (approved):

matfra:
Before releasing a new version, you should be able to drop this file on a kube master in stage and test that it works as expected

PR #2946, opened 2020-10-21
Have status API return full-length git SHAs in active_shas for marath…

EvanKrall:
…on apps, to be consistent with kubernetes. PAASTA-17026

This should fix the `{service}.{instance} on {cluster} not yet running {full git SHA}` bug we saw earlier.


EvanKrall:
(For context, I introduced a bug in https://github.com/Yelp/paasta/pull/2940 where the API returns 8-char SHA prefixes for marathon apps in `active_shas` but wait-for-deployment expects full-length SHAs. For k8s we return full-length SHAs.)

drmorr0 (approved):

PR #2947, opened 2020-10-22
fix spark-run with mrjob

lydian:
# Description
- fixed the issue that triggers `spark.master` key not exist error when use with `--mrjob` 

The original spark job called `create_spark_conf_str` twice, and unfortunately the create_spark_conf_str will drop the `spark.master`if `--mrjob` is set.  Therefore the second time we called the function will raise the error. 
The fix updates:
1.  create_spark_conf_str is called only once
2. also update create_spark_conf_str doesn't modify the passing argument spark_conf dict to ensure the stability. 

# Test: 
- make test pass
- manually run with with `paasta spark-run --mrjob` works on ads devc batch` also works!

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -400,12 +400,12 @@ def get_smart_paasta_instance_name(args):
 
 def get_spark_env(
     args: argparse.Namespace,
-    spark_conf: Dict[str, str],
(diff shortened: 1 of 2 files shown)


qui (approved):

lydian:
@qui I merged the change, but given it is late, I'll let you decide when to make a release. Thanks!

PR #2950, opened 2020-10-24
[oapi] Discard unknown keys in oapi client

keymone:


EvanKrall (approved):
Thanks for this, this should make oapi a lot less painful.

PR #2951, opened 2020-10-26
Don't include from contrib, remove init, add splunk tools

keymone:


EvanKrall:
I think this will cause mypy to fail, which is why I added the __init__.py in the first place. Maybe we should throw a `# type: ignore` or something on this line? (Or just make mypy not type-check `tests/test_kubernetes_tools.py`?

```
tests/test_kubernetes_tools.py:62: error: Cannot find module named 'paasta_tools.contrib.get_running_task_allocation'
tests/test_kubernetes_tools.py:62: note: See https://mypy.readthedocs.io/en/latest/running_mypy.html#missing-imports
```

EvanKrall:
e.g. 
```
krall@dev54-uswest1adevc:~/pg/paasta (u/maksym/fix-package) $ git diff
diff --git a/tox.ini b/tox.ini
index 0b57d8e54..7514a0bfb 100644
--- a/tox.ini
+++ b/tox.ini
@@ -162,7 +162,7 @@ mypy_paths =
     tests/deployd/test_common.py
     tests/deployd/test_watchers.py
     tests/frameworks
-    tests/test_kubernetes_tools.py
+    ; tests/test_kubernetes_tools.py
     tests/test_long_running_service_tools.py
     tests/test_marathon_tools.py
     tests/test_setup_kubernetes_job.py
```

keymone:
@krall ugh.. i didn't notice   that.. i made this PR because creating __init.py__  under contrib made all the itest_<dist> fail, the error didn't have any details so i  decided to have splunk_tools.py instead. if you know why paasta-tools would fail to build a package with contrib - i'm fine with fixing that directly as well :)

keymone:
@EvanKrall  ugh.. i didn't notice that.. i made this PR because creating `__init.py__` under contrib made all the `itest_<distro>` fail, the error didn't have any details so i decided to have `splunk_tools.py` instead. if you know why paasta-tools would fail to build a package with contrib - i'm fine with fixing that directly as well :)

EvanKrall:
I wonder if it's because we're including `__init__.py` in the list of `scripts` in setup.py? This seems to make `make itest_xenial` happy:
```
diff --git a/setup.py b/setup.py
index 15f272f18..298441ead 100644
--- a/setup.py
+++ b/setup.py
@@ -87,7 +87,7 @@ setup(
         "paasta_tools/synapse_srv_namespaces_fact.py",
     ]
     + glob.glob("paasta_tools/contrib/*.sh")
-    + glob.glob("paasta_tools/contrib/*.py"),
+    + [s for s in glob.glob("paasta_tools/contrib/*.py") if s != "paasta_tools/contrib/__init__.py"],
     entry_points={
         "console_scripts": [
             "paasta=paasta_tools.cli.cli:main",
```

EvanKrall:
feels like there should be a cleaner way to exclude `__init__.py` than a list comprehension though

keymone:
`glob.glob("paasta_tools/contrib/[!_]*.py")` should do the trick

PR #2954, opened 2020-10-26
Check if config_sha / git_sha are None before trying to call .startswith

EvanKrall:
PAASTA-17026


EvanKrall:
closed in favor of https://github.com/Yelp/paasta/pull/2955

PR #2955, opened 2020-10-26
Return null config_sha when k8s objects don't have config_sha annotation

EvanKrall:
PAASTA-17080


ronin13 (approved):
Thanks for the fix! :)

PR #2965, opened 2020-10-30
Only exit non-zero if there is a critical failure for replication scr…

qui:
…ipts

The alert for this script is the cron staleness check, which will alert if the exit code of the command is anything other than 0.

I thought about adding `test $? -lt 2` to the end of the command so that would only alert if the exit code of the script were 2, but if the script encounters any Python exceptions, it will exit 1, and I think we should probably still alert in that case.

### DIFF
--- a/paasta_tools/check_services_replication_tools.py
+++ b/paasta_tools/check_services_replication_tools.py
@@ -75,14 +75,6 @@ def parse_args() -> argparse.Namespace:
         help="define a different soa config directory",
     )
     parser.add_argument(
-        "--warn",
-        dest="under_replicated_warn_pct",
-        type=float,
-        default=5,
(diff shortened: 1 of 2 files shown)


kawaiwanyelp (approved):
I think this is fine, but the staleness check configured on puppet warns and crits on the same percentage, so I'm not sure this will necessarily help? Or maybe the intention is here to clean up some tech debt? Also, if before this change, you need remove it from all usages of this script in puppet.

qui:
Yeah, I'll remove usages of `--warn` from puppet. In a previous change (https://github.com/Yelp/paasta/pull/2958), I added a threshold so that it only warns if there are not enough failing instances, but that didn't reduce pages because of the problem in this PR.

PR #2966, opened 2020-11-03
 Use .get with defaults instead of [] to avoid ApiAttributeError on missing timeStamp or message. PAASTA-17100

EvanKrall:


paasta_tools/kubernetes_tools.py
```diff
@@ -1912,7 +1913,11 @@ def format_pod_event_messages(
     rows: List[str] = list()
     rows.append(PaastaColors.blue(f"Pod Events for {pod_name}"))
     for message in pod_event_messages:
-        rows.append(f"   Event at {message['timeStamp']}: {message['message']}")
```
keymone, line 1915: no need to catch exception here, you should be able to use .get:

    rows.append(f"   Event at {message.get('timeStamp', 'unknown time')}: {message['message']}")

keymone (approved):
minor nit

paasta_tools/kubernetes_tools.py
EvanKrall: thanks, didn't know about that.

drmorr0 (approved):

PR #2967, opened 2020-11-04
Fix paasta status mesh data for kubernetes instances

keymone:
Use correct service name for mesh data in paasta status for kubernetes instances


Rob-Johnson (approved):
can we add a test (or an assertion to an existing test) to make sure we don't introduce this regression again?

PR #2971, opened 2020-11-11
Revert Yelp/paasta#2970

kawaiwanyelp:
### Description
Yelp/paasta#2970 seems to be causing issues with setup_kubernetes_job by making it fail for a bit whenever a new HPA for a service needs to be created, though it does eventually succeed those services (then it fails for another service :P). This in turn blocks service deployments. So, this PR just reverts the change. We'll revisit it again later.


EvanKrall:
title should be "Revert Yelp/paasta#2768" I think

EvanKrall (approved):

PR #2974, opened 2020-11-12
Improve metastatus performance by pulling all pods at once & caching

jfongatyelp:
Currently in prod, metastatus is timing out when requesting any resources with groupings (aka when verbosity>1)

This adds 2 improvements:
1. Cache any requests to ~~`kuberetes.client.get_all_nodes`~~ `kubernetes_tools.get_all_nodes` or ~~`kubernetes.client.get_all_pods`~~ `kubernetes_tools.get_all_pods`
2. Reduce `# pods` API requests to a single `get_all_pods` API request and filter in memory instead. 
  - With the original code, each `get_pods_by_node` query took 1-2 seconds themselves, and there are over 599 nodes in pnw-prod. This went well over the 120 timeout for the api.

With this change, running `paasta_metastatus -vvv` returns in 1m35s, where previously it was taking >30m (I never was able to get it to finish).  As a result, running the API, it was able to return successfully even with `--max-request-seconds 120`, where in prod we're currently just getting an empty status after 120s.

### DIFF
--- a/paasta_tools/metrics/metastatus_lib.py
+++ b/paasta_tools/metrics/metastatus_lib.py
@@ -56,6 +56,7 @@ from paasta_tools.mesos_tools import is_task_terminal
 from paasta_tools.mesos_tools import MesosResources
 from paasta_tools.mesos_tools import MesosTask
 from paasta_tools.utils import PaastaColors
+from paasta_tools.utils import time_cache
 from paasta_tools.utils import print_with_indent
 
 
@@ -452,7 +453,7 @@ def assert_mesos_tasks_running(metrics: MesosMetrics,) -> HealthCheckResult:
 
 
 def assert_kube_pods_running(kube_client: KubeClient,) -> HealthCheckResult:
-    statuses = [get_pod_status(pod) for pod in get_all_pods(kube_client)]
+    statuses = [get_pod_status(pod) for pod in get_kube_all_pods(kube_client)]
     running = statuses.count(PodStatus.RUNNING)
     pending = statuses.count(PodStatus.PENDING)
     failed = statuses.count(PodStatus.FAILED)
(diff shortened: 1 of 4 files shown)


paasta_tools/metrics/metastatus_lib.py
```diff
@@ -856,6 +857,15 @@ def get_resource_utilization_by_grouping(
         for attribute_value, slaves in slave_groupings.items()
     }
 
+@time_cache(ttl=300)
+def get_kube_all_nodes(kube_client: KubeClient) -> Sequence[V1Node]:
+    nodes: Sequence[V1Node] = get_all_nodes(kube_client)
+    return nodes
+
+@time_cache(ttl=300)
+def get_kube_all_pods(kube_client: KubeClient) -> Sequence[V1Pod]:
```
Rob-Johnson, line 866: minor nit: can you call these functions `get_all_[pods|nodes]_cached` instead? also, maybe they'd be better off living in `kubernetes_tools` instead?

Rob-Johnson (changes requested):
fix and ship

keymone:
1) are we definitely sure caching works? `time_cache` uses function args as cache key, so i imagine if we call the cached function twice with different `KubeClient` instances (even if for the same k8s cluster) we're going to create two cache entries

2) i'm worried we're going to beat our kubernetes api servers to death. it already takes ~20s to get prod pods and the kind of caching we want to use here is not shared by paasta api workers, so each worker will expire independently.

3) (almost offtopic) i'm sure kubernetes natively provides a ton of metrics, we're already collecting some into prometheus, can we query that instead? this should probably be a separate discussion, @robj thoughts?

we can ship this as a short term fix, but in the long term i'd rather cache `get_pods_by_node` calls and make them asynchronous. in longer term figure out how to share cache between paasta api workers.

jfongatyelp:
> 1. are we definitely sure caching works? `time_cache` uses function args as cache key, so i imagine if we call the cached function twice with different `KubeClient` instances (even if for the same k8s cluster) we're going to create two cache entries\

I haven't verified exactly, but In my manual runs, before I changed all instances of `get_all_pods` to the cached version the run took an extra 60s. But in `paasta_metastatus`, we only ever create `kube_client` once and pass it to relevant methods in this library, so `time_cache` should actually hit the same cache.

> we can ship this as a short term fix, but in the long term i'd rather cache `get_pods_by_node` calls and make them asynchronous. in longer term figure out how to share cache between paasta api workers.

Thanks, shared caching or pulling from prometheus sounds great. I think this change will at least get `paasta metastatus` functioning again, and I'll open a separate ticket to improve how we fetch this information

paasta_tools/metrics/metastatus_lib.py
jfongatyelp: That makes sense, thanks. I've moved these over.

qui (approved):

PR #2975, opened 2020-11-16
U/talal/fix spark paasta instance name

talalryz:
FIxing a regression caused by https://github.com/Yelp/paasta/pull/2930


lydian (approved):
Do you want to make the tron_tools change on the same PR?

paasta_tools/tron_tools.py
```diff
@@ -191,6 +191,13 @@ def pick_spark_ui_port(service, instance):
     return preferred_port
 
 
+def get_smart_paasta_instance_name():
+    if os.environ.get("TRON_JOB_NAMESPACE"):
+        tron_job = os.environ.get("TRON_JOB_NAME")
+        tron_action = os.environ.get("TRON_ACTION")
+        return f"{tron_job}.{tron_action}"
+
```
talalryz, line 199: From the tron guide https://github.com/Yelp/Tron/blob/master/docs/command_context.rst, 
these  ENV variables should be available here as well.

Since this pathway is tron only, I'm not checking for other args here

talalryz:
> Do you want to make the tron_tools change on the same PR?

Yes, i've updated for that change

paasta_tools/tron_tools.py
```diff
@@ -235,6 +242,7 @@ def get_spark_config_dict(self):
         aws_creds = get_aws_credentials(
             aws_credentials_yaml=self.config_dict.get("aws_credentials_yaml")
         )
+        paasta_instance = get_smart_paasta_instance_name() or self.get_instance()
```
kawaiwanyelp, line 245: This is unnecessary (meaning you also don't have to define a `get_smart_paasta_instance_name` in tron_tools. Whenever `TronActionConfig` is initialized, we already compose the job and action [here](https://github.com/Yelp/paasta/blob/983ada69507cd1727826ec4f1674e482d21aad89/paasta_tools/tron_tools.py#L571).

kawaiwanyelp (changes requested):
The only change necessary is to patch `spark_run` I think.

paasta_tools/tron_tools.py
talalryz: I've dropped the changes to tron_tools

kawaiwanyelp (approved):

PR #2977, opened 2020-11-17
Cache things in redis

mks-m:
`redcache.d` is a decorator that fetches function result from redis cache if one is configured, otherwise (or upon error or timeout) will call the cached function directly.

takes two arguments:

- `resolution` is the time resolution to look for cached fragments, eg for `resolution=60`, the cache key will be calculated as `timestamp - timestamp % 60`.
- `selector` is a function that receives cached function's arguments and returns a hashable subset that will be used to derive cache key, should be used for things like skipping stateful objects, like network connections etc.

the decorator must only be used for pure functions(no side-effects) that take long time to run, it's not practical to use for anything that runs in less than 100-200ms.

### DIFF
--- a/paasta_tools/cli/cmds/metastatus.py
+++ b/paasta_tools/cli/cmds/metastatus.py
@@ -28,6 +28,9 @@ from paasta_tools.utils import PaastaColors
 from paasta_tools.utils import SystemPaastaConfig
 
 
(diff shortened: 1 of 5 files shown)


vkhromov (approved):

mattmb:

PR #2978, opened 2020-11-20
[cli] Only load the required subcommand

keymone:
Due to inconssistencies in current paasta cli implementation, we're preloading all subcommand modules to find actual subcommands and populate the argument parser.

This PR (for the time being) hardcodes existing subcommands, uses stripped down arg parser to find the subcommand we want to run, loads that subcommand and prepares the arg parser for it.

Drops the total overhead from 2.5s to 0.2s, though the impact won't be as large for actual subcommands because most of load time is spent in `util` modules which is used by all of them.

### DIFF
--- a/paasta_tools/cli/cli.py
+++ b/paasta_tools/cli/cli.py
@@ -21,6 +21,9 @@ import pkgutil
 import subprocess
 import sys
 import warnings
+from typing import Any
+from typing import List
+from typing import Tuple
 
 import argcomplete
 
@@ -73,16 +76,6 @@ def calling_external_command():
         return sys.argv[1] in list_external_commands()
     else:
         return False
-
-
-def get_command_help(command):
-    return f"(run 'paasta {command} -h' for usage)"
-
-
-def external_commands_items():
-    for command in list_external_commands():
-        command_help = get_command_help(command)
-        yield command, command_help
 
 
 def exec_subcommand(argv):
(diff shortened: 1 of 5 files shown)


vkhromov (changes requested):
There is failed test which seems to be related to CLI, could you please fix that.

vkhromov (approved):
:ship: assuming that the failed tests are OK and that doesn't mean that something real is being broken.

Rob-Johnson (approved):

paasta_tools/cli/cli.py
```diff
@@ -84,7 +99,49 @@ def add_subparser(command, subparsers):
     add_subparser_fn(subparsers)
 
 
-def get_argparser():
+PAASTA_SUBCOMMANDS = {
```
Rob-Johnson, line 102: In the long term we should just standardise the naming for this rather than having to maintain a mapping. I think start-stop-restart needs to be pulled apart, but otherwise it's just s/_/- right?

keymone: yep, i wanted to reduce the diff and impact, so i didn't pull start/stop/restart apart and didn't rename pause_service_autoscaler command.

Rob-Johnson:
oh, another failure mode: can you verify that if you pass an incorrect subcommand (like `paasta foo`) you'll still get the right help output (invalid choice: pick between xxx)? reading this I think you might not get that, since you haven't loaded all the argparsers?

keymone:
> oh, another failure mode: can you verify that if you pass an incorrect subcommand (like `paasta foo`) you'll still get the right help output (invalid choice: pick between xxx)? reading this I think you might not get that, since you haven't loaded all the argparsers?

yep:

```
(py36-linux) dev ~/paasta % paasta foo                                                                                                                                         
Argument parse error: argument : invalid choice: 'foo' (choose from 'help', 'api', 'autoscale', 'boost', 'bounce-lock', 'check', 'cook-image', 'deployd', 'echo', 'fsm', 'get-docker-image', '
get-latest-deployment', 'info', 'itest', 'list', 'list-clusters', 'list-deploy-queue', 'local-run', 'logs', 'mark-for-deployment', 'metastatus', 'pause_service_autoscaler', 'performance-chec
k', 'push-to-registry', 'remote-run', 'restart', 'rollback', 'secret', 'security-check', 'spark-run', 'start', 'status', 'stop', 'sysdig', 'tools-hadown', 'true', 'validate', 'wait-for-deplo
yment')                                                                                                                                                                                       
The PaaSTA command line tool. The 'paasta' command is the entry point to  
...
```

PR #2979, opened 2020-11-21
Never use custom metrics interface for signalfx metrics adapter, PAASTA-17091

EvanKrall:
The first commit, ec85007, makes it so we don't ever use the custom metrics interface for the signalfx metrics adapter, only the external interface.

The second commit, e76906f, removes all code dealing with `horizontal_autoscaling`, since nothing uses this, and the first commit likely introduces a bug in this code w.r.t. uwsgi autoscaling. (It would generate signalflow code that would query _all_ uwsgi timeseries with no filtering.)


paasta_tools/kubernetes_tools.py
```diff
@@ -642,6 +535,7 @@ def get_autoscaling_metric_spec(
             if (
                 autoscaling_params.get("forecast_policy") == "moving_average"
                 or "offset" in autoscaling_params
+                or load_system_paasta_config().get_hpa_always_uses_external_for_signalfx()
             ):
                 hpa_metric_name = self.namespace_external_metric_name(metrics_provider)
                 signalflow = LEGACY_AUTOSCALING_SIGNALFLOW.format(
```
stug, line 541: I think this means that all instances using the `metrics_provider: uwsgi` or `http` will end up with an implicit `forecast_policy: moving_average` -- is that what we want?

stug: (oops meant to start a review -- not done!)

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -192,118 +192,6 @@
                 "autoscaling": {
                     "type": "object"
                 },
-                "horizontal_autoscaling": {
-                    "type": "object",
-                    "additionalProperties": false,
-                    "properties": {
```
stug, line 198: It looks like there's one instance of compute-infra-test-service in norcal-stagef still using this section, so you probably need to change/remove it before shipping this

paasta_tools/kubernetes_tools.py
stug: (There probably aren't too many instances using `uwsgi` or `http` without `moving_average` I bet)

stug (approved):
I think this looks roughly good but I want to confirm that we're ok with forcing everything to use `moving_average` (my guess is that's ok?)

I also didn't look super closely at the removed itest yelpsoaconfigs -- it doesn't really look like we're losing any coverage, but can you confirm that's the case?

PR #2980, opened 2020-11-23
First attempt at running tests on github actions.

EvanKrall:
travis-ci.org seems to be intentionally capacity-limited so as to force people to move to travis-ci.com before the end of the year, when travis-ci.org will stop working. The open source tier on travis-ci.com has severe limits (1000 build minutes per org per...month? ever? We'd use 1000 build minutes in about 10 or 11 paasta builds.)
https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

This ports us over to Github Actions, which are free for open source projects, with [generous limits](https://docs.github.com/en/free-pro-team@latest/actions/reference/usage-limits-billing-and-administration) (probably the first one we'll hit is the 20 concurrent builders, which is still ~4x better than travis was.)


Rob-Johnson (approved):
👍  thankyou for doing this

PR #2983, opened 2020-11-30
allow tagging secret as shared across environments

piax93:
This change allows to add a metadata string when creating or updating a secret which is meant to have the same value across multiple environments. This is done to allow developers to provide context about why a secret has been duplicated instead of following the best practice of using different secret value in different runtime environments. This ultimately allows to create catalogue of secrets which are meant to be replicated and detect when new duplicated secrets are added not following the convention.

The additional parameter has been implemented in the latest version of `vault-tools` (0.9.2). I recognise this is a pretty large bump for the dependency, so I'd ask reviewers to look at the library changelog for potentially breaking changes introduced in the meantime (cause I don't have enough knowledge of this codebase to catch possible edge cases).

Also, please point me to any bit of documentation which may need updating from this.

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -105,6 +105,16 @@ def add_subparser(subparsers):
         default=False,
         help="Optionally pass the plaintext from stdin",
     )
+    secret_parser.add_argument(
+        "--cross_environment",
+        required=False,
+        type=str,
+        help=(
+            "Provide motivation in case the same value is being duplicated "
+            "across multiple runtime environments when adding or updating a secret"
+        ),
+        metavar="MOTIVATION",
+    )
     secret_parser.set_defaults(command=paasta_secret)
 
 
@@ -209,7 +219,10 @@ def paasta_secret(args):
         if not plaintext:
             print("Warning: Given plaintext is an empty string.")
         secret_provider.write_secret(
-            action=args.action, secret_name=args.secret_name, plaintext=plaintext
+            action=args.action,
+            secret_name=args.secret_name,
+            plaintext=plaintext,
+            cross_environment_motivation=args.cross_environment,
         )
         secret_path = os.path.join(
             secret_provider.secret_dir, f"{args.secret_name}.json"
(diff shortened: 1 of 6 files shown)


mattmb:
I think our itests are probably pretty incomplete so we do need to be a bit careful. You should check `add` `update` `decrypt` work as expected with the latest version of vault-tools. On the mesos masters the decryption doesn't rely on paasta so that should be safe. The syncing of secrets to Kubernetes does rely on this code though (https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes/bin/paasta_secrets_sync.py) so we should probably validate that works with the version bump too (or roll out slowly to just confirm it's okay in stage). cc @vkhromov might have a better knowledge of what's changed in vault-tools

What's the longer term plan with this flag? How are you going to enforce that people fill it out?

piax93:
Yes (if they need to have the same secret value across different environments).
Of course they could also just edit the JSON output manually, but we thought it would be better if they were given the option to directly state their intentions when running the command to add/update a secret.

mattmb (approved):
Gotcha, sounds good. Please coordinate with onpoint when you're ready to roll out. I don't think there's anything breaking in the vault-tools bump but it would be sensible to pin paasta-tools in puppet and just validate that updating the secret for a kubernetes paasta service works cleanly in a stage before we unpin and roll it everywhere

paasta_tools/cli/cmds/secret.py
```diff
@@ -105,6 +105,16 @@ def add_subparser(subparsers):
         default=False,
         help="Optionally pass the plaintext from stdin",
     )
+    secret_parser.add_argument(
+        "--cross_environment",
```
keymone, line 109: no underscores please

paasta_tools/cli/cmds/secret.py
```diff
@@ -105,6 +105,16 @@ def add_subparser(subparsers):
...
+        required=False,
+        type=str,
+        help=(
+            "Provide motivation in case the same value is being duplicated "
```
keymone, line 113: argument name doesn't correspond to help message, maybe rename it to `--cross-env-motivation`?

keymone (changes requested):
fix-ship

paasta_tools/cli/cmds/secret.py
piax93: done

paasta_tools/cli/cmds/secret.py
piax93: done

keymone (approved):

PR #2984, opened 2020-12-01
Monkey-patch default HPA scaling policy into HPA spec

kawaiwanyelp:
### Description
This is a first attempt at support HPA scaling policies that were newly introduced in Kubernetes v1.18 in PaaSTA to control how quickly we scale down. Couple of things to note:

- I've had to monkey-patch the scaling policies configs into the JSON data the k8s client posts to get this to work, because the k8s client library only supports v1.17. 
- I've only created (what I think is) a sensible default so far, but the next step will be to make the defaults configurable in system paasta configs or on a per-instance basis from soa-configs.
- The sensible default is to at most per minute scale down by either 30% of current replicas or 5% of max replicas, whichever is highest. Essentially, this means we only scale down at most by 30% of max replicas, and the rate of scale down plateaus to 5% of max replicas as we scale down. 
- I have not touched the scale up policy. The default scale up policy is [here](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#scaling-policies).

### Testing
`make test`
Updating an HPA in kubestage with these changes yields the following alpha annotation (I pretty-formatted it for readability) visible in `kubectl get hpa`:
```
"autoscaling.alpha.kubernetes.io/behavior": "{
  \"ScaleUp\": {
    \"StabilizationWindowSeconds\": 0,
    \"SelectPolicy\": \"Max\",
    \"Policies\": [
      {
        \"Type\": \"Pods\",
        \"Value\": 4,
        \"PeriodSeconds\": 15
      },
      {
        \"Type\": \"Percent\",
        \"Value\": 100,
        \"PeriodSeconds\": 15
      }
    ]
  },
  \"ScaleDown\": {
    \"StabilizationWindowSeconds\": 300,
    \"SelectPolicy\": \"Max\",
    \"Policies\": [
      {
        \"Type\": \"Percent\",
        \"Value\": 30,
        \"PeriodSeconds\": 60
      },
      {
        \"Type\": \"Pods\",
        \"Value\": 1,
        \"PeriodSeconds\": 60
      }
    ]
  }
}"
```

### Notes
- I could make the policy a little more safe, by decreasing the percent scale down value, and/or increasing the `periodSeconds` parameter, which is the scale-down interval.
- Not sure I want to make this configurable on the first pass. I want to see if this helps before I go further. We should be able to see more controlled scale-downs.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -404,13 +404,22 @@ class KubeClient:
             ),
             fset=_set_disrupted_pods,
         )
+
         self.deployments = kube_client.AppsV1Api()
         self.core = kube_client.CoreV1Api()
         self.policy = kube_client.PolicyV1beta1Api()
         self.apiextensions = kube_client.ApiextensionsV1beta1Api()
         self.custom = kube_client.CustomObjectsApi()
         self.autoscaling = kube_client.AutoscalingV2beta2Api()
-        self.request = kube_client.ApiClient().request
+
+        self.api_client = kube_client.ApiClient()
+        self.request = self.api_client.request
+        # This function is used by the k8s client to serialize OpenAPI objects
+        # into JSON before posting to the api. The JSON output can be used
+        # in place of OpenAPI objects in client function calls. This allows us
+        # to monkey-patch the JSON data with configs the api supports, but the
+        # Python client lib may not yet.
+        self.jsonify = self.api_client.sanitize_for_serialization
 
 
 class KubernetesDeploymentConfig(LongRunningServiceConfig):
@@ -486,12 +495,44 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             defaults=default_params,
         )
 
+    # TODO: move the default scaling policy to system paasta configs
+    def get_autoscaling_scaling_policy(self, max_replicas: int) -> Dict:
+        """Returns the k8s HPA scaling policy in raw JSON. Requires k8s v1.18
+        to work.
+        """
+        # The HPA scaling algorithm is as follows. Every sync period (default:
+        # 15 seconds), the HPA will:
+        #   1. determine what the desired capacity is from metrics
+        #   2. apply min/max replica scaling limits
+        #   3. rate-limit the scaling magnitude (e.g. scale down by no more than
+        #      30% of current replicas)
+        #   4. constrain the scaling magnitude by the period seconds (e.g. scale
+        #      down by no more than 30% of current replicas per 60 seconds)
+        #   5. record the desired capacity, then pick the highest capacity from
+        #      the stabilization window (default: last 300 seconds) as the final
+        #      desired capacity.
+        #      - the idea is to stabilize scaling against (heavily) fluctuating
+        #        metrics
+        return {
+            "scaleDown": {
+                "stabilizationWindowSeconds": 300,
+                # the policy in a human-readable way: scale down every 60s by
+                # at most 30% of current replicas.
+                "selectPolicy": "Max",
+                "policies": [{"type": "Percent", "value": 30, "periodSeconds": 60}],
+            }
+        }
+
     def namespace_external_metric_name(self, metric_name: str) -> str:
         return f"{self.get_sanitised_deployment_name()}-{metric_name}"
 
     def get_autoscaling_metric_spec(
-        self, name: str, cluster: str, namespace: str = "paasta"
-    ) -> Optional[V2beta2HorizontalPodAutoscaler]:
+        self,
+        name: str,
+        cluster: str,
+        kube_client: KubeClient,
+        namespace: str = "paasta",
+    ) -> Optional[Union[V2beta2HorizontalPodAutoscaler, Dict]]:
         # Returns None if an HPA should not be attached based on the config,
         # or the config is invalid.
 
@@ -587,7 +628,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             )
             return None
 
-        return V2beta2HorizontalPodAutoscaler(
+        hpa = V2beta2HorizontalPodAutoscaler(
             kind="HorizontalPodAutoscaler",
             metadata=V1ObjectMeta(
                 name=name, namespace=namespace, annotations=annotations
@@ -601,6 +642,17 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
                 ),
             ),
         )
+
+        # In k8s v1.18, HPA scaling policies can be set:
+        #   https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#support-for-configurable-scaling-behavior
+        # However, the python client library currently only supports v1.17, so
+        # we need to monkey-patch scaling policies until the library is updated
+        # v1.18.
+        scaling_policy = self.get_autoscaling_scaling_policy(max_replicas)
+        if scaling_policy:
+            hpa = kube_client.jsonify(hpa)  # this is a hack, see KubeClient class
+            hpa["spec"]["behavior"] = scaling_policy
+        return hpa
 
     def get_deployment_strategy_config(self) -> V1DeploymentStrategy:
         # get soa defined bounce_method
--- a/tests/test_kubernetes_tools.py
+++ b/tests/test_kubernetes_tools.py
@@ -241,6 +241,13 @@ def test_load_kubernetes_service_config():
 
 
 class TestKubernetesDeploymentConfig:
+    @pytest.fixture(autouse=True)
+    def mock_load_kube_config(self):
+        with mock.patch(
+            "paasta_tools.kubernetes_tools.kube_config.load_kube_config", autospec=True,
+        ) as m:
+            yield m
+
     def setup_method(self, method):
         mock_config_dict = KubernetesDeploymentConfigDict(
             bounce_method="crossover", instances=3,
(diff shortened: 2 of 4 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -486,12 +495,40 @@ def get_autoscaling_params(self) -> AutoscalingParamsDict:
             defaults=default_params,
         )
 
+    # TODO: move the default scaling policy to system paasta configs
+    # TODO: allow per-instance overrides in soa-configs, and merge this function
```
qui, line 499: I'm not sure if we necessarily need per-instance overrides, so let's not add it as a TODO?

qui:
Have you checked that the returned API object doesn't break anything? For example, last time I changed the HPA, the following things broke:
* setup_kubernetes_job exited non-zero when creating a new HPA (not updating an HPA)
* paasta status -v on instances with an HPA

paasta_tools/kubernetes_tools.py
```diff
@@ -486,12 +495,40 @@ def get_autoscaling_params(self) -> AutoscalingParamsDict:
...
+    def get_autoscaling_scaling_policy(self, max_replicas: int) -> Dict:
+        """Returns the k8s HPA scaling policy in raw JSON. Requires k8s v1.18
+        to work.
+        """
+        # policy in human terms: scale down at most every 60 seconds by 30% of
+        # current replicas or 5% of max replicas, whichever is higher.
+        return {
+            "scaleDown": {
+                "selectPolicy": "Max",
+                "stabilizationWindowSeconds": 300,
```
qui, line 510: How does the stabilization window interact with the period seconds on the policies? Would it be easier to reason about if they were the same?

tests/test_kubernetes_tools.py
```diff
@@ -254,6 +262,30 @@ def setup_method(self, method):
...
+    # TODO: remove once hpa scaling policy patch is removed
+    def patch_expected_autoscaling_spec(self, obj, max_replicas=3):
+        """Currently, HPA scaling policies are monkey-patched onto the HPA
+        OpenAPI objects. This function does the same thing so that any assertions
+        we do will pass.
+        """
+        if isinstance(obj, V2beta2HorizontalPodAutoscaler):
+            obj = KubeClient().serialize(obj)
+            obj["spec"]["behavior"] = {
+                "scaleDown": {
```
qui, line 274: Could you use the method defined in the code instead of rewriting it here?

paasta_tools/kubernetes_tools.py
```diff
@@ -486,12 +495,40 @@ def get_autoscaling_params(self) -> AutoscalingParamsDict:
...
 
+    # TODO: move the default scaling policy to system paasta configs
+    # TODO: allow per-instance overrides in soa-configs, and merge this function
+    #   with `get_autoscaling_params`
+    def get_autoscaling_scaling_policy(self, max_replicas: int) -> Dict:
+        """Returns the k8s HPA scaling policy in raw JSON. Requires k8s v1.18
+        to work.
+        """
+        # policy in human terms: scale down at most every 60 seconds by 30% of
+        # current replicas or 5% of max replicas, whichever is higher.
```
stug, line 506: Just out of curiosity, how did we come to this?

paasta_tools/kubernetes_tools.py
kawaiwanyelp: I've added more detailed comments to explain the difference. They serve two different purposes, and I don't think it makes sense to make them identical. The purposes of the two numbers are:

- `periodSeconds`: Let's say the HPA decides to scale down by a max of 4 pods. Setting `periodSeconds` of 60 seconds of tells the HPA to scale down by a max of 4 pods every 60 seconds. Not setting `periodSeconds` means the HPA will effectively default the `periodSeconds` to the interval at which the HPA runs, which is the HPA sync period. By default, this is 15 seconds.
- `stabilizationWindowSeconds`: Every sync period, the HPA will produce and record a recommended desired replica count. However, if the underlying metrics are fluctuating, the count can too. To fight this, a stabilization window can be set, which is the amount of time in the past the HPA will look for previous desired replica counts, and select the highest one. In this case, the HPA will select the highest desired replica count from the last 300 seconds. The idea is to smoothen downscaling against sometimes unreliable metrics.

paasta_tools/kubernetes_tools.py
kawaiwanyelp: The 30% I picked from the PaaSTA service autoscaler, which I believe had a scaling limit of 30% of current replicas. The 5% is an arbitrary low end I picked. If we wanted to simplify, we could just remove the "Pods" policy and just keep the "Percent" policy.

kawaiwanyelp:
@qui I used `setup_kubernetes_job` to test for creation and updates of an HPA, as well as used `paasta status -v` on the instance I ran setup on. In all cases, I didn't see any errors. I think that makes sense since the core HPA spec hasn't changed - a default scaling policy already exists. We are just modifying it.

paasta_tools/kubernetes_tools.py
```diff
@@ -404,13 +404,22 @@ def __init__(self) -> None:
...
-        self.request = kube_client.ApiClient().request
+
+        self.api_client = kube_client.ApiClient()
+        self.request = self.api_client.request
+        # This function is used by the k8s client to serialize OpenAPI objects
+        # into JSON before posting to the api. The JSON output can be used
+        # in place of OpenAPI objects in client function calls. This allows us
+        # to monkey-patch the JSON data with configs the api supports, but the
+        # Python client lib may not yet.
+        self.serialize = self.api_client.sanitize_for_serialization
```
stug, line 422: So below in `get_autoscaling_metric_spec` we are calling `kube_client.serialize` to get the json version of the HPA object and returning that.  Am I correct in thinking that we need this here so that it will accept that already-serialized HPA object...?

stug (approved):
I _think_ this looks roughly good, and I think we'll just have to see how scaling works once it's shipped.  I have one question about the monkey-patching, however.

paasta_tools/kubernetes_tools.py
stug: Or is this just for convenience so you can call `kube_client.serialize` instead of `kube_client.api_client.sanitize_for_serialization`?  If it's the latter I vote we not do this -- we're sort of doing a hacky thing, and I'd rather we showed that rather than making it look nicer.  (Also we're not really serializing -- that would result in a json string rather than a dict).

paasta_tools/kubernetes_tools.py
```diff
@@ -404,13 +404,22 @@ def __init__(self) -> None:
...
         self.deployments = kube_client.AppsV1Api()
         self.core = kube_client.CoreV1Api()
         self.policy = kube_client.PolicyV1beta1Api()
         self.apiextensions = kube_client.ApiextensionsV1beta1Api()
         self.custom = kube_client.CustomObjectsApi()
         self.autoscaling = kube_client.AutoscalingV2beta2Api()
-        self.request = kube_client.ApiClient().request
+
+        self.api_client = kube_client.ApiClient()
+        self.request = self.api_client.request
```
stug, line 416: Does this get used anywhere?

stug: this = `self.request`

paasta_tools/kubernetes_tools.py
kawaiwanyelp: It's a "make this hacky thing look nice" thing, but another reason I did this is because it makes it easier to document and reference how to do this in one place. I think it should be documented in some way because I suspect we'll have to do this from time to time because the Python client library is consistently behind the current k8s version, and we do do a fair amount of monkey-patching here and there.

If you have strong feelings about making it obvious a hacky thing is a hacky thing, then I'll remove teh `self.serialize` obfuscation.

paasta_tools/kubernetes_tools.py
kawaiwanyelp: Yes, in `kubernetes_tools.get_events_for_object`, to make a specific API call without using the OpenAPI-generated function call, because the latter doesn't support an option we want.

paasta_tools/kubernetes_tools.py
qui: I would be in favor of simplifying it. In the project meeting this week, we were talking about how HPA doesn't make it easy to understand the decisions that it makes, through its logging, so we'd rather have as many of the decisions visible through our metrics instead.

paasta_tools/kubernetes_tools.py
qui: This question was also coming from the perspective of making it as easy as possible to explain what the HPA is doing in a particular situation.

After trying to reason through both cases, I agree it doesn't really make it easier to explain if they're the same value, so that's fine.

paasta_tools/kubernetes_tools.py
stug: Hmm I guess that is fine, although I think it's not super likely to people will think to come find this comment when they just see `kube_client.serialize()` -- it looks pretty proper and official.  

Maybe you could either add a comment to the place where you use it or call it something more obvious like `convert_to_dict_representation` or something?  Your call

kawaiwanyelp: I changed to `.jsonify` and added a comment that it is a hack.

paasta_tools/kubernetes_tools.py
kawaiwanyelp: Done.

paasta_tools/kubernetes_tools.py
```diff
@@ -486,12 +495,44 @@ def get_autoscaling_params(self) -> AutoscalingParamsDict:
...
+        #   5. record the desired capacity, then pick the highest capacity from
+        #      the stabilization window (default: last 300 seconds) as the final
+        #      desired capacity.
+        #      - the idea is to stabilize scaling against (heavily) fluctuating
+        #        metrics
+        return {
+            "scaleDown": {
+                "stabilizationWindowSeconds": 300,
+                # the policy in a human-readable way: scale down every 60s by
+                # at most 30% of current replicas or 5% of max replicas.
```
qui, line 520: The comment above helps a lot. Don't forget to update the `5%` here.

qui (approved):

PR #2986, opened 2020-12-04
Read supported storage classes from system_paasta_config

lennrt:
## Description
This change makes `paasta_tools` read the set of supported storage classes from `system_paasta_config` rather than relying on hard-coded values. We can then manage these configs in Puppet which will be useful for allowing users to experiment with new StorageClasses (e.g. ones with the new `gp3` volume type).

## Testing Done
`make test` passes

## Release Strategy
`patch` release

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -134,6 +134,7 @@ from paasta_tools.utils import DeployWhitelist
 from paasta_tools.utils import DockerVolume
 from paasta_tools.utils import get_config_hash
 from paasta_tools.utils import get_git_sha_from_dockerurl
+from paasta_tools.utils import get_supported_storage_classes
 from paasta_tools.utils import load_service_instance_config
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import load_v2_deployments_json
@@ -168,7 +169,6 @@ DISCOVERY_ATTRIBUTES = {
     "pool",
     "hostname",
 }
-SUPPORTED_STORAGE_CLASSES = {"ebs", "ebs-slow", "ebs-retain"}
 
 GPU_RESOURCE_NAME = "nvidia.com/gpu"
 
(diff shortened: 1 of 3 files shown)


qui:
Looks good, but you have some failing tests and pre-commit formatting that need to be fixed.

paasta_tools/kubernetes_tools.py
```diff
@@ -1145,10 +1145,18 @@ def get_volume_claim_templates(self) -> Sequence[V1PersistentVolumeClaim]:
         ]
 
     def get_storage_class_name(self, volume: PersistentVolume) -> str:
-        storage_class_name = volume.get("storage_class_name", "ebs")
-        if storage_class_name not in SUPPORTED_STORAGE_CLASSES:
-            log.warning(f"storage class {storage_class_name} is not supported")
-            storage_class_name = "ebs"
+        try:
+            system_paasta_config = load_system_paasta_config()
```
qui, line 1149: This would be a little cleaner if you caught the PaastaNotConfiguredError here. Then supported_storage_classes could be an empty list in that case, and then you only have to do the default once.

qui (approved):

paasta_tools/kubernetes_tools.py
```diff
@@ -1150,12 +1150,11 @@ def get_storage_class_name(self, volume: PersistentVolume) -> str:
             supported_storage_classes = (
                 system_paasta_config.get_supported_storage_classes()
             )
-            storage_class_name = volume.get("storage_class_name", "ebs")
-            if storage_class_name not in supported_storage_classes:
-                log.warning(f"storage class {storage_class_name} is not supported")
-                storage_class_name = DEFAULT_STORAGE_CLASS_NAME
         except (PaastaNotConfiguredError):
             log.warning("No PaaSTA configuration was found, returning default value")
```
qui, line 1154: Don't forget to set `supported_storage_classes` to an empty list.

qui (approved):

PR #2987, opened 2020-12-07
PAASTA-17119 Replace kube_hpa_status_current_replicas uwsgi hpa metric

gvolco:
I'm replacing the metric **kube_hpa_status_current_replicas** with **kube_deployment_spec_replicas** on the signalflow code for the uwsgi HPA.

For details on this new metric please check PAASTA-17119.

I'm also moving the signalflow code to a paasta config variable, so that we can override it if needed.

### Tests:
- Installed a paasta_tools version to kubestage and confirmed that the signalflow code was updated on different HPA objects.
- This change didn't bounce any pod as the signalflow code lives on HPA objects (as an annotation).

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2544,7 +2544,7 @@ setpoint = {setpoint}
 moving_average_window = '{moving_average_window_seconds}s'
 filters = filter('paasta_service', '{paasta_service}') and filter('paasta_instance', '{paasta_instance}') and filter('paasta_cluster', '{paasta_cluster}')
 
-current_replicas = data('kube_hpa_status_current_replicas', filter=filters, extrapolation="last_value").sum(by=['paasta_cluster'])
+current_replicas = data('kube_deployment_spec_replicas', filter=filters, extrapolation="last_value").sum(by=['paasta_cluster'])
 load_per_instance = data('{signalfx_metric_name}', filter=filters, extrapolation="last_value", maxExtrapolations=10).below(1, clamp=True)
 
 desired_instances_at_each_point_in_time = (load_per_instance - offset).sum() / (setpoint - offset)


gvolco:
Force pushed to fix unit tests..

paasta_tools/kubernetes_tools.py
```diff
@@ -538,7 +518,10 @@ def get_autoscaling_metric_spec(
                 or load_system_paasta_config().get_hpa_always_uses_external_for_signalfx()
             ):
                 hpa_metric_name = self.namespace_external_metric_name(metrics_provider)
-                signalflow = LEGACY_AUTOSCALING_SIGNALFLOW.format(
+                legacy_autoscaling_signalflow = (
+                    load_system_paasta_config().get_legacy_autoscaling_signalflow()
+                )
+                signalflow = legacy_autoscaling_signalflow.format(
```
stug, line 524: Smart!  This will make responding to HPA incidents easier, I think.

paasta_tools/utils.py
```diff
@@ -2527,6 +2528,30 @@ def get_git_repo_config(self, repo_name: str) -> Dict:
...
+filters = filter('paasta_service', '{paasta_service}') and filter('paasta_instance', '{paasta_instance}') and filter('paasta_cluster', '{paasta_cluster}')
+
+current_replicas = data('kube_deployment_spec_replicas', filter=filters, extrapolation="last_value").sum(by=['paasta_cluster'])
+load_per_instance = data('{signalfx_metric_name}', filter=filters, extrapolation="last_value", maxExtrapolations=10).below(1, clamp=True)
+
+desired_instances_at_each_point_in_time = (load_per_instance - offset).sum() / (setpoint - offset)
+desired_instances = desired_instances_at_each_point_in_time.mean(over=moving_average_window)
+
+(desired_instances / current_replicas).above(0).publish()
+"""
```
stug, line 2550: Just to be sure -- the only difference here is that we're using `kube_deployment_spec_replicas` instead of `kube_hpa_status_current_replicas`, yeah?

stug (approved):
Pretty sure this is solid, and because the signalflow is now part of the system paasta config, if it breaks things we can just puppet out a change

qui (approved):

paasta_tools/utils.py
gvolco: yep, that's the only thing I changed on the query

gvolco:
Most of the changes in this PR where included in #2994, so I'm rebasing from master and replacing kube_hpa_status_current_replicas with kube_deployment_spec_replicas.

PR #2989, opened 2020-12-10
fix vault secret provider (s/do_renew/do_cert_renew)

piax93:
In https://github.com/Yelp/paasta/pull/2983 I broke the Vault secret provider by bumping the version of `vault-tools` because in the latest versions the `do_renew` function of the `cert_tools` module has been renamed to `do_cert_renew`. The function signature has remained the same so no other adjustment should be required.

### DIFF
--- a/paasta_tools/secret_providers/vault.py
+++ b/paasta_tools/secret_providers/vault.py
@@ -11,7 +11,7 @@ try:
     from vault_tools.paasta_secret import get_vault_client
     from vault_tools.gpg import TempGpgKeyring
     from vault_tools.paasta_secret import encrypt_secret
-    from vault_tools.cert_tools import do_renew
+    from vault_tools.cert_tools import do_cert_renew
     import hvac
 except ImportError:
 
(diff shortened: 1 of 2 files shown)


vkhromov (approved):

paasta_tools/secret_providers/vault.py
```diff
@@ -11,7 +11,7 @@
     from vault_tools.paasta_secret import get_vault_client
     from vault_tools.gpg import TempGpgKeyring
     from vault_tools.paasta_secret import encrypt_secret
-    from vault_tools.cert_tools import do_renew
+    from vault_tools.cert_tools import do_cert_renew
     import hvac
 except ImportError:
```
keymone, line 16: tests didn't fail because of this. do we really want to ignore import errors? if this is necessary lets add some constant that indicates whether modules were loaded or not and fail tests on the valuer of that constant.

keymone (changes requested):
lets make tests fail when imports fail

paasta_tools/secret_providers/vault.py
piax93: `vault-tools` is an internal library which will never be loaded in tests (which run publicly), so we have no real way of checking that.

keymone:
@piax93 during build we have an environment variable that tells you that the build is internal, you could enable the test depending on that variable. but it's not a blocker, so feel free to ship.

keymone (approved):

PR #2990, opened 2020-12-10
 Dockerfile and Makefile for building a sidecar image with uwsgi_exporter

EvanKrall:
PAASTA-17180

### DIFF
--- a/yelp_package/dockerfiles/uwsgi_exporter-sidecar/Dockerfile
+++ b/yelp_package/dockerfiles/uwsgi_exporter-sidecar/Dockerfile
@@ -1,0 +1,4 @@
+FROM docker-dev.yelpcorp.com/xenial_yelp
+
+COPY ./uwsgi_exporter/uwsgi_exporter /bin/uwsgi_exporter
+CMD /bin/uwsgi_exporter --stats.uri http://127.0.0.1:8889/ --collect.cores false --web.listen-address :9117
--- a/yelp_package/dockerfiles/uwsgi_exporter-sidecar/Makefile
+++ b/yelp_package/dockerfiles/uwsgi_exporter-sidecar/Makefile
@@ -1,9 +1,11 @@
-DOCKER_IMAGE ?= docker-paasta.yelpcorp.com:443/uwsgi_exporter-k8s-sidecar
+UWSGI_EXPORTER_TAG ?= v1.0.0
+
+DOCKER_IMAGE ?= docker-paasta.yelpcorp.com:443/uwsgi_exporter-k8s-sidecar:$(UWSGI_EXPORTER_TAG)
 
 all: docker_image push
 
 uwsgi_exporter:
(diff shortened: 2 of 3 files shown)


yelp_package/dockerfiles/uwsgi_exporter-sidecar/Makefile
```diff
@@ -0,0 +1,15 @@
+DOCKER_IMAGE ?= docker-paasta.yelpcorp.com:443/uwsgi_exporter-k8s-sidecar
+
+all: docker_image push
+
+uwsgi_exporter:
+	git clone https://github.com/timonwong/uwsgi_exporter
```
keymone, line 6: lets checkout specific commit hash

EvanKrall: good idea

yelp_package/dockerfiles/uwsgi_exporter-sidecar/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM docker-dev.yelpcorp.com/xenial_yelp
+
+COPY ./uwsgi_exporter/uwsgi_exporter /bin/uwsgi_exporter
+CMD /bin/uwsgi_exporter --stats.uri http://127.0.0.1:8889/ --collect.cores false --web.listen-address :9117
```
qui, line 4: Is that stats.uri the URI the exporter hits on the service to get its metrics? In that case, I thought the stats port could be set by the service - is there a place we're always overriding it to 8889?

yelp_package/dockerfiles/uwsgi_exporter-sidecar/Makefile
```diff
@@ -0,0 +1,15 @@
...
+	git clone --branch v1.0.0 https://github.com/timonwong/uwsgi_exporter
+
+uwsgi_exporter/uwsgi_exporter: uwsgi_exporter
+	make -C uwsgi_exporter
+
+docker_image: uwsgi_exporter/uwsgi_exporter
+	docker build -t $(DOCKER_IMAGE) .
+
+push:
+	sudo -i bash -c 'docker push $(DOCKER_IMAGE)'
```
qui, line 15: We've discussed versioning the docker images - do you want to add a tag for that here, or solve that issue later?

stug, line 15: I _think_ this might need `sudo -H` to get at the dockercfg in /root so that it can push to docker-paasta

EvanKrall: I think `-i` will end up with a similar effect. I'll change this to -H though since I don't think we need to be running a login shell.

```
$ sudo -i bash -c 'echo $HOME'
/root
```

PR #2991, opened 2020-12-10
Remove 0/0 replicasets before counting active SHAs, so that wait-for-…

EvanKrall:
…deployment doesn't wait for 0/0 replicasets to disappear. PAASTA-17186

### DIFF
--- a/paasta_tools/instance/kubernetes.py
+++ b/paasta_tools/instance/kubernetes.py
@@ -446,8 +446,12 @@ def kubernetes_status(
         kube_client=kube_client,
         namespace=job_config.get_kubernetes_namespace(),
     )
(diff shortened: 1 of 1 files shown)


paasta_tools/instance/kubernetes.py
```diff
@@ -446,8 +446,12 @@ def kubernetes_status(
         kube_client=kube_client,
         namespace=job_config.get_kubernetes_namespace(),
     )
+    # For the purpose of active_shas/app_count, don't count replicasets that are at 0/0.
+    actually_running_replicasets = [
```
qui, line 450: I think this list is the inverse of what you want to include in the next line? e.g. these are all the 0/0 replicasets you want to ignore.

PR #2992, opened 2020-12-11
Stricter schema checking for persistent_volumes, to catch containerPa…

EvanKrall:
…th instead of container_path

This catches the recent typo when run locally. I'm currently running this against locally against the rest of yelpsoa-configs to make sure I'm not going to block anybody's changes.


jfongatyelp (approved):
Thanks for catching this so quickly. Looks like we don't actually use any other fields in the code even if someone provides them, so restricting this seems like it'll also help avoid surprises for service owners.

PR #2998, opened 2020-12-18
Configure uwsgi_exporter sidecar container, gated behind a per-instance toggle.

EvanKrall:
for PAASTA-17092

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -724,6 +724,16 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         service_namespace_config: ServiceNamespaceConfig,
         hacheck_sidecar_volumes: Sequence[DockerVolume],
     ) -> Sequence[V1Container]:
+        return self.get_hacheck_sidecar_container(
+            system_paasta_config, service_namespace_config, hacheck_sidecar_volumes,
+        )
+
+    def get_hacheck_sidecar_container(
+        self,
+        system_paasta_config: SystemPaastaConfig,
+        service_namespace_config: ServiceNamespaceConfig,
+        hacheck_sidecar_volumes: Sequence[DockerVolume],
+    ) -> Sequence[V1Container]:
         registrations = " ".join(self.get_registrations())
         # s_m_j currently asserts that services are healthy in smartstack before
         # continuing a bounce. this readiness check lets us achieve the same thing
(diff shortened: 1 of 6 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -724,6 +712,16 @@ def get_sidecar_containers(
         service_namespace_config: ServiceNamespaceConfig,
         hacheck_sidecar_volumes: Sequence[DockerVolume],
     ) -> Sequence[V1Container]:
+        return self.get_hacheck_sidecar_container(
+            system_paasta_config, service_namespace_config, hacheck_sidecar_volumes,
+        ) + self.get_uwsgi_exporter_sidecar_container(system_paasta_config)
```
stug, line 717: nit, but I find returning 1-element lists a little unexpected

paasta_tools/utils.py
```diff
@@ -2556,6 +2558,16 @@ def get_legacy_autoscaling_signalflow(self) -> str:
...
 
+    def get_uwsgi_exporter_sidecar_image_url(self) -> str:
+        """Get the docker image URL for the uwsgi_exporter sidecar container"""
+        return self.config_dict.get(
+            "uwsgi_exporter_sidecar_image_url",
+            "docker-paasta.yelpcorp.com:443/uwsgi_exporter-k8s-sidecar:v1.0.0",
+        )
+
+    def default_should_run_uwsgi_exporter_sidecar(self) -> bool:
+        return self.config_dict.get("default_should_run_uwsgi_exporter_sidecar", False,)
```
stug, line 2569: nit: extra comma at the end?  Doesn't look like it changes things, though.

stug (approved):
Looks roughly sane.  I have small comments but only because I am a curmudgeon.

paasta_tools/kubernetes_tools.py
EvanKrall: Yeah, I guess that's weird. It just saves me from having to do something like 
```
hacheck_sidecar = self.get_hacheck_sidecar_container(...)
if hacheck_sidecar is not None:
    sidecars.append(hacheck_sidecar)
```
maybe there's a better one-liner for that.

paasta_tools/utils.py
EvanKrall: huh, surprised that Black let that stay.

qui (approved):

EvanKrall:
@stug I refactored to have those methods return Optional[V1Container] instead of Lists, and fixed the weird trailing comma.

stug (approved):

PR #3001, opened 2021-01-06
PAASTA-17176 Add rate limit param for s_k_j

gvolco:
This change adds a `--rate-limit` param to `setup_kubernetes_job` so that we can control how many updates are sent to the k8s API on every cronjob run.
The intention is to limit the damage that big bounces cause on our biggest clusters.

### Tests:
- Manually executed `setup_kubernetes_job` while removing some of the existing deployments on kubestage.


mattmb (approved):
Looks good to me

keymone (approved):

Rob-Johnson (approved):

PR #3002, opened 2021-01-06
fix docs for healthcheck_max_consecutive_failures

Rob-Johnson:
corrects the default value to match reality


atallahade:
@Rob-Johnson Is the default also 30s on mesos?

Rob-Johnson:
ah, thanks, I missed that one @atallahade

keymone (approved):

Rob-Johnson:
@keymone it sure would be nice to have these docs generated from the models somehow

PR #3003, opened 2021-01-06
Fix the documentation giving wrong defaults

atallahade:
The real default was taken from [here](https://github.com/Yelp/paasta/blob/master/paasta_tools/long_running_service_tools.py#L258).


atallahade:
Duplicates #3002

PR #3005, opened 2021-01-07
stop building on Trusty

Rob-Johnson:
there is only one host running trusty at Yelp now (and I think it can be
safely ignored).
we should stop building PaaSTA for Trusty now that we no longer have to support it.


Rob-Johnson:
this should unblock https://github.com/Yelp/paasta/pull/2350 and https://github.com/Yelp/paasta/pull/2960

EvanKrall (approved):

PR #3006, opened 2021-01-07
Add additional labels for Prometheus use

nemacysts:
deploy_group and docker_image are existing dimensions that we can't use
relabeling to include in metric scraping since they're only present in
the k8s container spec and Prometheus doesn't expose that.

most importantly, scrape_uwsgi_prometheus will be how we filter down the
set of pods that Prometheus should actually scrape - the idea is
currently that this label is only set when someone has opted into using
the uwsgi exporter, but in the future this label should be set if a
service has chosen to use uwsgi autoscaling instead of checking this
feature toggle

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1442,8 +1442,6 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             "paasta.yelp.com/service": self.get_service(),
             "paasta.yelp.com/instance": self.get_instance(),
             "paasta.yelp.com/git_sha": git_sha,
-            "paasta.yelp.com/deploy_group": self.get_deploy_group(),
-            "paasta.yelp.com/docker_image": self.get_docker_image(),
         }
 
         # Allow the Prometheus Operator's Pod Service Monitor for specified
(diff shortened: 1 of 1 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1237,6 +1237,8 @@ def get_kubernetes_metadata(self, git_sha: str) -> V1ObjectMeta:
                 "paasta.yelp.com/service": self.get_service(),
                 "paasta.yelp.com/instance": self.get_instance(),
                 "paasta.yelp.com/git_sha": git_sha,
+                "paasta.yelp.com/deploy_group": self.get_deploy_group(),
+                "paasta.yelp.com/docker_image": self.get_docker_image(),
```
nemacysts, line 1241: with these, we have the exact same set of service dimensions that /status/metrics provides (minus cluster, but I was gonna add that in the prometheus config)

EvanKrall:
What are the labels on the V1Deployment/V1StatefulSet for? I think we may only need the labels on the Pods. Just maintaining consistency?

EvanKrall (approved):

nemacysts:
> What are the labels on the V1Deployment/V1StatefulSet for? I think we may only need the labels on the Pods. Just maintaining consistency?

yup - I can leave them just on the pod if we don't care about these being consistent

EvanKrall (approved):

PR #3009, opened 2021-01-15
[WIP] Add support for k8s secret sync to namespaces

mattmb:
Putting this up early for some feedback, probably some tests need fixing and I need to test the script manually since we don't have an integration test. Mostly want opinions on `INSTANCE_TYPE_TO_K8S_NAMESPACE`, we don't seem to have this map anywhere so I'm relying on some conventions to work it out, kinda ugly, other ideas welcome.

To support paasta secrets in operator workloads we need to sync secrets
for services that run in other namespaces into those namespaces. This
adds mapping for workloads to k8s namespaces and changes the sync script
to list all services and find any instance types and sync to those
instance type's namespaces.

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -13,12 +13,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import argparse
+import importlib
 import json
 import logging
 import os
 import sys
+from collections import defaultdict
 from typing import Mapping
-from typing import Sequence
+from typing import Set
 
 from kubernetes.client.rest import ApiException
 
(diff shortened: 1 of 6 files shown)


paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -30,20 +32,35 @@
...
+        log.warning(
+            f"Cannot lookup k8s namespace no module called paasta_tools.{instance_type}_tools"
+        )
+        continue
+    INSTANCE_TYPE_TO_K8S_NAMESPACE[instance_type] = getattr(
+        instance_type_module, "KUBERNETES_NAMESPACE", "paasta"
+    )
+
+print(INSTANCE_TYPE_TO_K8S_NAMESPACE)
+
```
vkhromov, line 60: Remove `print()`.

vkhromov, line 60: It's interesting that we already have metainformation about namespaces in both `KUBERNETES_NAMESPACE` global variables for some things, and in `cr_id` functions for other things, see e.g. https://github.com/Yelp/paasta/blob/f39268ee424c6b7561eb532f3ba06b875b9cff55/paasta_tools/flink_tools.py#L152.  Can we maybe use `cr_id` everywhere instead of introducing new `KUBERNETES_NAMESPACE` variables?

It's also better to move `INSTANCE_TYPE_TO_K8S_NAMESPACE` outside of `paasta_secrets_sync.py`

mattmb: Interesting, I hadn't spotted that `cr_id` function. Good shout on location of `INSTANCE_TYPE_TO_K8S_NAMESPACE` too, I'll find somewhere more sensible

PR #3010, opened 2021-01-15
Make get_kubernetes_metadata use actual node pool

mattmb:
This fixes an assumption that pods use just a node selector. Some of the
cassandra Pods use node affinity and in practice a lot of things can
impact the node/pool that a Pod actually ends up on. This checks the Pod
to see what node it is scheduled on and gets the pool from the Nodes
labels (which I think is more reliable).

Note I tested this on kubestage but I don't know the impact if this
breaks something so please warn me if I need to be careful.

### DIFF
--- a/paasta_tools/contrib/get_running_task_allocation.py
+++ b/paasta_tools/contrib/get_running_task_allocation.py
@@ -134,7 +134,7 @@ def get_kubernetes_resource_request(
 
 
 def get_kubernetes_metadata(
-    pod: V1Pod,
+    kube_client: KubeClient, pod: V1Pod,
 ) -> Tuple[
     Optional[str],
     Optional[str],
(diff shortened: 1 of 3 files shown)


paasta_tools/contrib/get_running_task_allocation.py
```diff
@@ -143,14 +143,17 @@ def get_kubernetes_metadata(
     Optional[str],
     Optional[str],
 ]:
+    node = kubernetes_tools.get_pod_node(kube_client, pod, cache_nodes=True)
+    if node:
+        pool = node.metadata.labels.get("yelp.com/pool", "default")
+    else:
+        pool = "default"
```
qui, line 150: Maybe it would be worth pulling this out into a function?

I think this needs to handle the case where `labels` is None, just in case there are some rogue nodes. Also, what do you think also falling back to the node selector?

mattmb: Nice, yeah I think it would be cleaner as a separate function and good shout on the labels being optional. Not so sure it's worth having a fallback though, I'm happy to add it if you feel it's too risky without but it feels like extra complexity for small reassurance.

qui (approved):

PR #3012, opened 2021-01-20
Add support for auto-configuring cpu_burst_add (PAASTA-17172)

qui:
This makes `cpu_burst_add` a valid key in AutoConfigUpdater and fetches it from the Splunk CSV, but doesn't actually write cpu_burst_add in the rightsizer batch. This should be a no-op in terms of the configs updated in the daily cpu job.

Since we're adding suggested_cpu_burst_add to the existing cpu CSV, the `--csv-key` option will allow us to roll out the burst changes more slowly. My plan:
1. Release this change
2. Add `--csv-key cpus` to the existing Jenkins job for the cpu CSV.
3. Add cpu_burst_add to the rightsizer batch in a follow-up PR.
4. Add a new job with `--csv-key cpu_burst_add` that can be selectively applied to different clusters via the criteria filter (`-f`).
5. Once we're happy with the burst changes, we can eventually merge the two jobs back together.

I've tested the rightsizer batch manually to check that the `--csv-key` option works as expected.

### DIFF
--- a/paasta_tools/cli/schemas/autotuned_defaults/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/autotuned_defaults/kubernetes_schema.json
@@ -15,6 +15,11 @@
                     "minimum": 0,
                     "exclusiveMinimum": true
                 },
+                "cpu_burst_add": {
+                    "type": "number",
+                    "minimum": 0.0,
+                    "exclusiveMinimum": false
+                },
                 "disk": {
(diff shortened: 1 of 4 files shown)


stug (approved):
Sorry for the delay -- this looks roughly sane to me.

matfra:
Thanks for shipping this. I agree with the step by step rollout plan.

PR #3015, opened 2021-01-22
PAASTA-17231-LogSuffix-Fix

lanceleih:
### Description
Bug fix when running `paasta logs` for containers inside flink instances.

### Testing Done
`make test`

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -1333,7 +1333,6 @@ def paasta_logs(args: argparse.Namespace) -> int:
     if args.instances is None:
         instances = None
     else:
-
         instances = args.instances.split(",")
         # Check for instance inputs with suffixes (i.e. Flink instances)
         instances_without_suffixes = [x.split(".")[0] for x in instances]
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -1338,7 +1338,7 @@ def paasta_logs(args: argparse.Namespace) -> int:
         # Check for instance inputs with suffixes (i.e. Flink instances)
         instances_without_suffixes = [x.split(".")[0] for x in instances]
         args.instances = ",".join(instances_without_suffixes)
-        
+
         if verify_instances(args.instances, service, clusters):
             return 1
 
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -1333,6 +1333,9 @@ def paasta_logs(args: argparse.Namespace) -> int:
         instances = None
     else:
         instances = args.instances.split(",")
+        # Check for flint instance names only, not the containers inside them
+        check_flink_instances = [x.split(".")[0] for x in instances if x.split(".")]
+        args.instances = ",".join(check_flink_instances)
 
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/logs.py
```diff
@@ -1333,6 +1333,9 @@ def paasta_logs(args: argparse.Namespace) -> int:
         instances = None
     else:
         instances = args.instances.split(",")
+        # Check for flint instance names only, not the containers inside them
+        check_flink_instances = [x.split(".")[0] for x in instances if x.split(".")]
```
stug, line 1337: A couple thoughts:
* Although we know that these are probably flink instances, the log suffix behavior is something that other instance types could use, too.  I think it'd be better to call this something like `instances_without_suffixes` (which I know doesn't really roll off the tongue), reference PAASTA-15311 in the comment, and only talk about flink as an example use case, if at all
* if there is no period character, `split()` will return a 1-element list, so you don't actually need the `if` part of this -- you can just access `x.split(".")[0]`

tests/cli/test_cmds_logs.py
```diff
@@ -388,6 +388,19 @@ def test_extract_utc_timestamp_from_log_line_when_invalid_date_format():
...
+    service = "fake_service"
+    clusters = ["fake_cluster1"]
+    instance = "fake_instance.container"
+
+    with mock.patch(
+        "paasta_tools.cli.utils.verify_instances", autospec=True
+    ) as verify_instances_patch:
+        verify_instances_patch.return_value = True
+
+        assert verify_instances_patch(instance, service, clusters)
```
stug, line 401: This doesn't actually test anything -- you've replaced `verify_instances` with a mock object that will always return `True`, so you could run `assert verify_intances_patch("abc", 123, False)` and it'd still return True.

I think you should patch `verify_instances` and then call `paasta_logs` and assert that `verify_instances` was called with just the instance names without the suffxies.  Let me know if you want help doing that.

stug (changes requested):

tests/cli/test_cmds_logs.py
```diff
@@ -388,6 +388,23 @@ def test_extract_utc_timestamp_from_log_line_when_invalid_date_format():
     assert not logs.extract_utc_timestamp_from_log_line(line)
 
 
+def test_paasta_log_line_passes_filter_true_for_instances_with_suffixes():
```
stug, line 391: I think this name comes from the fact that you copied some of this from tests of the `logs.paasta_log_line_passes_filter` method.  We should rename this to explain it better.

paasta_tools/cli/cmds/logs.py
```diff
@@ -1332,7 +1333,11 @@ def paasta_logs(args: argparse.Namespace) -> int:
     if args.instances is None:
         instances = None
     else:
+
```
stug, line 1336: I think you added newlines inadvertently in a couple places.  I don't think they're really helping readability or anything, so probably worth removing them.

stug (approved):
looks good to me -- a couple tiny things that you can fix and ship.

tests/cli/test_cmds_logs.py
lanceleih: Renamed to `test_verify_instances_for_instances_with_suffixes()`

PR #3016, opened 2021-01-25
PAASTA-16376: Add a config option for kubernetes services to get a ro…

jfongatyelp:
…utable IP even if not in smartstack or used for prometheus scraping

This simply extends the work done in #3014 to allow for services to explicitly require a routable IP, rather than implied via smartstack or prometheus configurations.

Originally, *only* service instances which registered to smartstack would be assigned a routable IP.  In #3014 we also added the logic to give a routable IP to pods which needed to be exposed so prometheus could scrape their metrics.

There may be other usecases to expose a service (e.g. perhaps it needs to be accessible, but not through our service mesh, nor does it need to be scraped by prometheus), so it seems useful to give service owners the option to explicitly request a routable IP via their configuration when needed.  

See PAASTA-16376 for more info including pending usecases for this feature.  I've held off on adding to the documentation to verify the functionality internally first.

## Testing done
`make test` still passes

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -625,6 +625,9 @@
                     "type": "integer",
                     "minimum": 0
                 },
+                "routable_ip": {
+                    "type": "boolean"
+                },
                 "lifecycle": {
                     "type": "object",
                     "properties": {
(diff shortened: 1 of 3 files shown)


qui (approved):
Change looks good to me. Can you add a little context about why we need this here or in a ticket for future reference?

PR #3019, opened 2021-01-28
Shows replicaset table in nonverbose request

lanceleih:
Modified the API request to retrieve the replicaset data without having to use the verbose option when running `paasta status`.

**Verbose Option Deactivated**
image

**Verbose Option Activated**
image

### DIFF
--- a/paasta_tools/instance/kubernetes.py
+++ b/paasta_tools/instance/kubernetes.py
@@ -200,10 +200,12 @@ async def job_status(
     kstatus["app_id"] = app_id
     kstatus["pods"] = []
     kstatus["replicasets"] = []
-    num_tail_lines = calculate_tail_lines(verbose)
-    kstatus["pods"] = await asyncio.gather(
-        *[pod_info(pod, client, num_tail_lines) for pod in pod_list]
-    )
+
+    if verbose > 0:
+        num_tail_lines = calculate_tail_lines(verbose)
+        kstatus["pods"] = await asyncio.gather(
+            *[pod_info(pod, client, num_tail_lines) for pod in pod_list]
+        )
 
     for replicaset in replicaset_list:
         kstatus["replicasets"].append(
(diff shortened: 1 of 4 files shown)


paasta_tools/instance/kubernetes.py
```diff
@@ -200,29 +200,25 @@ async def job_status(
     kstatus["app_id"] = app_id
     kstatus["pods"] = []
     kstatus["replicasets"] = []
-    if verbose > 0:
-        num_tail_lines = calculate_tail_lines(verbose)
+    num_tail_lines = calculate_tail_lines(verbose)
+    kstatus["pods"] = await asyncio.gather(
+        *[pod_info(pod, client, num_tail_lines) for pod in pod_list]
+    )
```
stug, line 206: I think we don't want to show pods at the non-verbose level

stug (commented):
Can we also change the line about when the app was created?  Or maybe just straight up remove it since we have the replicaset timestamps now?

e.g. this:
```
App created: 2020-09-02 16:40:42 (4 months ago)
```

You can also do that as a follow-up if you'd rather.

paasta_tools/instance/kubernetes.py
lanceleih: Perhaps we can leave the App Created Timestamp in the verbose level?

tests/cli/test_cmds_status.py
```diff
@@ -1037,6 +1037,35 @@ def test_paasta_status_exception(system_paasta_config):
         )
 
 
+def test_format_kubernetes_replicaset_table_in_non_verbose(mock_kubernetes_status):
+    with mock.patch(
+        "paasta_tools.cli.cmds.status.format_kubernetes_replicaset_table", autospec=True
+    ) as mock_format_kubernetes_replicaset_table, mock.patch(
+        "paasta_tools.cli.cmds.status.bouncing_status_human", autospec=True
```
stug, line 1044: this tests that `paasta status` can form the replicasets if it receives them from the API in non-verbose mode, but you should also have a test of the changes to paasta_tools/instance/kubernetes.py so that you can show that the API in non-verbose mode it still returns replicasets

stug (commented):
Looks good but I think you need one more test.

(also sorry for the delay -- got caught up in an incident today)

stug (approved):

PR #3023, opened 2021-02-03
Extract some autoscaling values to named constants

nemacysts:
### Extract default autoscaling setpoint into a named constant

We'll need this constant to configure the Prometheus adapter, so let's
also make it a named constant so that we don't have multiple definitions
to keep in sync

###  Extract the default moving average window to a named constant

We use this in a couple places (and will use it in yet another soon), so
we might as well have a single place where it's defined.


paasta_tools/utils.py
```diff
@@ -118,6 +118,9 @@
 
 DEFAULT_SOA_CONFIGS_GIT_URL = "sysgit.yelpcorp.com"
 
+DEFAULT_AUTOSCALING_SETPOINT = 0.8
+DEFAULT_AUTOSCALING_MOVING_AVERAGE_WINDOW = 1800
```
EvanKrall, line 122: might make the most sense to have these in long_running_service_tools.py since that's where the superclass of MarathonServiceConfig / KubernetesJobConfig lives. I know Maksym is trying to make utils.py less of a kitchen-sink.

EvanKrall (changes requested):

paasta_tools/utils.py
nemacysts: sounds good!

EvanKrall (approved):

paasta_tools/long_running_service_tools.py
```diff
@@ -329,7 +332,7 @@ def get_autoscaling_params(self) -> AutoscalingParamsDict:
         default_params: AutoscalingParamsDict = {
             "metrics_provider": "mesos_cpu",
             "decision_policy": "proportional",
-            "setpoint": 0.8,
+            "setpoint": DEFAULT_AUTOSCALING_SETPOINT,
```
EvanKrall, line 335: this is where I would do the change I proposed in https://github.com/Yelp/paasta/pull/3027#discussion_r569883664, you may or may not want to do it in this commit

EvanKrall (approved):

PR #3024, opened 2021-02-03
Stop adding SFX custom metrics annotation

nemacysts:
This will cause issues once we add additional metrics to the HPA since
the SFX adapter will try to act on those metrics and, predictably, fail.

**NOTE: this is using https://github.com/Yelp/paasta/pull/3023 as a base to make it easier to break things up into multiple PRs - my plan is to merge the base PR and then merge this into master**

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -541,7 +541,6 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
                 )
             )
         elif metrics_provider in ("http", "uwsgi"):
-            annotations = {"signalfx.com.custom.metrics": ""}
             if (
                 autoscaling_params.get("forecast_policy") == "moving_average"
                 or "offset" in autoscaling_params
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -542,7 +542,6 @@ def get_autoscaling_metric_spec(
                 )
             )
         elif metrics_provider in ("http", "uwsgi"):
-            annotations = {"signalfx.com.custom.metrics": ""}
```
EvanKrall, line 544: You may want to go ahead remove the `else` block starting at line 582, since I think that functionality will be broken by this removal. (Or add this annotation within that block)

EvanKrall (changes requested):

paasta_tools/kubernetes_tools.py
nemacysts: hmm, I wonder if we should also remove the 
```
if (
    autoscaling_params.get("forecast_policy") == "moving_average"
    or "offset" in autoscaling_params
    or load_system_paasta_config().get_hpa_always_uses_external_for_signalfx()
)
``` 
check since that'll always be true anyway - but I guess we can clean that up when I fully remove http or when we cleanup once we're done here?

EvanKrall (approved):

PR #3026, opened 2021-02-03
Add skeleton for the prometheus adapter script

nemacysts:
The full script is more lines than anyone wants to review at once so
I've split it into two commits/PRs: one that sets up the skeleton/logic
flow and another that actually fill in all the stub methods

**NOTE: this is using #3025 as a base to make it easier to break things up into multiple PRs - my plan is to merge the base PR and then merge this into master**

### DIFF
--- a/paasta_tools/setup_prometheus_adapter_config.py
+++ b/paasta_tools/setup_prometheus_adapter_config.py
@@ -18,7 +18,6 @@ import argparse
 import logging
 import sys
 from pathlib import Path
-from typing import Any
 from typing import Dict
 from typing import List
 from typing import Optional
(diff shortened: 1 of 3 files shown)


EvanKrall (approved):

EvanKrall (commented):
actually wait a sec

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -0,0 +1,249 @@
...
+    Currently supports the following metrics providers:
+        * uwsgi
+    """
+    rules: List[PrometheusAdapterRule] = []
+    # we don't know ahead of time what services are autoscaled, so we need to figure out
+    # what services are running in our cluster and then check if they're autoscaled
+    for service_config_path in soa_dir.glob(f"*/kubernetes-{paasta_cluster}.yaml"):
+        service_name = str(service_config_path.relative_to(soa_dir).parent)
+        service_config = yaml.safe_load(service_config_path.read_text())
+        for instance, instance_config in service_config.items():
```
EvanKrall, line 142: I think you may want to use `utils.get_services_for_cluster` and PaastaServiceConfigLoader. This will do more processing than just loading the yaml (it'll create KubernetesDeploymentConfigs), but that's probably a good thing. It should do the `_` prefix filtering for you, and then you won't be reimplementing the interpretation of the autoscaling dict.

EvanKrall (changes requested):

paasta_tools/setup_prometheus_adapter_config.py
nemacysts: https://github.com/Yelp/paasta/pull/3026/commits/66caa8f747f8186b131849f13360a88898e7004b does this now

EvanKrall (approved):

PR #3030, opened 2021-02-09
Stop returning non-zero when no rules are generated

nemacysts:
This is a fail-safe to stop us from deleting all rules, but this is
currently tripping up our alerting on clusters that don't have
Prometheus-enabled HPAs

### DIFF
--- a/paasta_tools/setup_prometheus_adapter_config.py
+++ b/paasta_tools/setup_prometheus_adapter_config.py
@@ -375,7 +375,7 @@ def main() -> int:
 
     if not config["rules"]:
         log.error("Got empty rule configuration - refusing to continue.")
-        return 1
+        return 0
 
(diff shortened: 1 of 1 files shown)


jfongatyelp (approved):
How likely is it we could break generating config at all from soaconfigs? Seems this was the only thing that would catch that case but if that's pretty straightforward seems ok, and at least we can check for this case in logs 👍

nemacysts:
> How likely is it we could break generating config at all from soaconfigs? Seems this was the only thing that would catch that case but if that's pretty straightforward seems ok, and at least we can check for this case in logs 👍

probably not going to happen - this is mostly just to stop silly bugs/misuse from nuking all our config and breaking autoscaling for a bit (kinda like the way that our internal aws key rotation stuff won't do things automatically if going to delete everything)

PR #3031, opened 2021-02-09
Allow services to override the  uwsgi_exporter stats.uri parameter, t…

EvanKrall:
…o support services where uwsgi stats are on a nonstandard port.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -795,11 +795,18 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
     ) -> Optional[V1Container]:
 
         if self.should_run_uwsgi_exporter_sidecar(system_paasta_config):
+            stats_uri_env = V1EnvVar(
+                name="STATS_URI",
+                value=self.get_autoscaling_params().get(
+                    "uwsgi_stats_uri", "http://127.0.0.1:8889/"
+                ),
+            )
+
             return V1Container(
                 image=system_paasta_config.get_uwsgi_exporter_sidecar_image_url(),
                 resources=self.get_sidecar_resource_requirements("uwsgi_exporter"),
                 name=UWSGI_EXPORTER_POD_NAME,
-                env=self.get_kubernetes_environment(),
+                env=self.get_kubernetes_environment() + [stats_uri_env],
                 ports=[V1ContainerPort(container_port=9117)],
             )
 
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -887,7 +887,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             )
         return ret
 
-    def get_kubernetes_environment(self) -> Sequence[V1EnvVar]:
+    def get_kubernetes_environment(self) -> List[V1EnvVar]:
         kubernetes_env = [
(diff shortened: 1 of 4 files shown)


tests/test_kubernetes_tools.py
```diff
@@ -677,6 +677,37 @@ def test_get_uwsgi_exporter_sidecar_container_should_run(self):
...
+            }
+        )
+        if uwsgi_stats_uri is not None:
+            self.deployment.config_dict["autoscaling"][
+                "uwsgi_stats_uri"
+            ] = uwsgi_stats_uri
+
+        ret = self.deployment.get_uwsgi_exporter_sidecar_container(system_paasta_config)
+        expected_env_var = V1EnvVar(name="STATS_URI", value=expected_uri,)
+        print(ret)
```
nemacysts, line 708: i'm assuming this was a debug print :p

paasta_tools/kubernetes_tools.py
```diff
@@ -786,11 +786,18 @@ def get_uwsgi_exporter_sidecar_container(
     ) -> Optional[V1Container]:
 
         if self.should_run_uwsgi_exporter_sidecar(system_paasta_config):
+            stats_uri_env = V1EnvVar(
+                name="STATS_URI",
+                value=self.get_autoscaling_params().get(
+                    "uwsgi_stats_uri", "http://127.0.0.1:8889/"
+                ),
+            )
```
nemacysts, line 794: did we go for `STATS_URI` over `STATS_PORT` to not clobber anyone's existing `STATS_PORT` usages? 

it seems pretty unlikely that people will have a stats URI that's not localhost and served on something that's not http for uwsgi purposes - thoughts on just hardcoding the protocol/host and exposing something like `UWSGI_STATS_PORT` so that we hide all the other stuff from our users?

nemacysts (approved):
exposing just a port seems like it might be preferable from a config minimization standpoint - but I'm not super set on that if you disagree

tests/test_kubernetes_tools.py
EvanKrall: yep good catch

paasta_tools/kubernetes_tools.py
EvanKrall: I figure this is more flexible, in case we need it for some reason, but yeah maybe YAGNI applies here.

I'm 90% sure the environment variables for this container are independent of the environment in the service's container, so conflicts are not a problem.

tests/test_kubernetes_tools.py
EvanKrall: (I miss paasta_print and the pre-commit hook rejecting all regular prints - that always caught my debug prints)

paasta_tools/kubernetes_tools.py
nemacysts: I mostly just worry about having to add documentation to places that say "add a `uwsgi_stats_uri` key with value `http://127.0.0.1:${YOUR_CUSTOM_PORT}/` (make sure you don't forget the trailing `/`! )" and whatnot - there's a lot less our users can mess up if we just expose the port and there's less documentation for us to write as well :p 

that said: I'm also fine with exposing this - worst case we just do a mass `sed` in the future and switch everyone to `uwsgi_stats_port` if we want to punt on simplyding the config

nemacysts (approved):

PR #3035, opened 2021-02-11
PAASTA-17261-Paasta-Logs-Support-Tron-Instance-Names

lanceleih:
### Description
PAASTA-17231 validated instances with suffixes in a way where it only will validate the string before the suffix. This was a great fix for Flink instances for example.

`paasta logs -s stream_sql -i main.SUPERVISOR`

In this case, verify_instances will only validate `main`.

However, this fix excluded tron instance names which required the strings before and after the suffix to be validated. 

`paasta logs -s reminder -i compute-infrastructure-oncall-us.compute-infrastructure-oncall-us`

In this case, verify_instances will validate the entire string `compute-infrastructure-oncall-us.compute-infrastructure-oncall-us`.

This change allows for both instances with suffixes to be validated correctly.

### Tests
`make test` - PASSED

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -1334,9 +1334,6 @@ def paasta_logs(args: argparse.Namespace) -> int:
         instances = None
     else:
         instances = args.instances.split(",")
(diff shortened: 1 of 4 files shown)


stug (approved):
Looks good!

PR #3038, opened 2021-02-15
MESH-1070 Readiness check based on EDS yaml files

gvolco:
I'm modifying `are_services_up_in_pod()` so that backends for a given service are pulled from the EDS yaml files instead of the Envoy /clusters endpoint.

This is to mitigate a load issue on the Envoy admin endpoint, which is triggered when multiple pods are launched on a single node, firing a large number of readiness checks in parallel.

Note that I'm also dropping the check for `eds_health_status`. This param should be reported by the EDS, but `meshd` doesn't set it. As a consequence, Envoy always sees backends as `HEALTHY`.

### DIFF
--- a/paasta_tools/envoy_tools.py
+++ b/paasta_tools/envoy_tools.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import collections
+import os
 import socket
 from typing import AbstractSet
 from typing import Any
@@ -30,6 +31,7 @@ from typing import Tuple
 from typing import Union
 
 import requests
+import yaml
 from kubernetes.client import V1Pod
 from mypy_extensions import TypedDict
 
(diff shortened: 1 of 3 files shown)


paasta_tools/envoy_tools.py
```diff
@@ -47,38 +49,19 @@ class EnvoyBackend(TypedDict, total=False):
...
-        )
-
-        healthy_backends = [
-            backend
-            for backend in backends_per_registration.get(registration, [])
-            if backend[0]["address"] == pod_ip
-            and backend[0]["port_value"] == pod_port
-            and backend[0]["eds_health_status"] == "HEALTHY"
-        ]
+        backends_from_eds = get_backends_from_eds(registration, envoy_eds_path,)
```
Rob-Johnson, line 62: trailing comma?

paasta_tools/envoy_tools.py
```diff
@@ -120,6 +103,36 @@ def get_casper_endpoints(
     return frozenset(casper_endpoints)
 
 
+def get_backends_from_eds(service: str, envoy_eds_path: str) -> List[Tuple[str, int]]:
```
Rob-Johnson, line 106: I think it's probably more accurate to call the `service` param `namespace`?

paasta_tools/envoy_tools.py
```diff
@@ -120,6 +103,36 @@ def get_casper_endpoints(
...
+                       the requested service
+    """
+    backends = []
+    eds_file_for_service = f"{envoy_eds_path}/{service}/{service}.yaml"
+
+    if os.access(eds_file_for_service, os.R_OK):
+        with open(f"{envoy_eds_path}/{service}/{service}.yaml") as f:
+            eds_yaml = yaml.safe_load(f)
+            for resource in eds_yaml.get("resources", []):
+                endpoints = resource.get("endpoints", [])
```
Rob-Johnson, line 121: I don't think you need the default arg if you're checking for None below anyway (get will return None by default)

Rob-Johnson (approved):
minor nits, but lg2m.

can you get @qui's signoff too (since this is impacting service level stuff that we don't technically own anymore). Thanks!

Rob-Johnson:
@gvolco just thinking again, I think we should probably wrap this in a feature toggle that we can modify faster than we can downgrade paasta (and run the big bounce).

paasta_tools/envoy_tools.py
gvolco: yeah you are right

gvolco:
After some debate on #service-mesh we decided to wrap this in a feature toggle. I'm updating this PR to include it.

Rob-Johnson (approved):
lg2m!

PR #3042, opened 2021-02-23
Sort Prometheus rulesets for easier comparison

nemacysts:
Previously we were just doing strict dictionary equality, but this
breaks down when your values are lists and those lists aren't in the
same order.


EvanKrall (approved):
might be easier to make sure we create `rules` in a sorted order? I think all that's needed is to throw a `sorted()` in here:

https://github.com/Yelp/paasta/pull/3042/files#diff-9a59d2c71b041881064cae2c027b254613c0920c41a21366fb6e6ff67448184bR248

EvanKrall (approved):

PR #3044, opened 2021-02-26
Allow HPA scaledown policies to be overridden from soa configs

stug:
See PAASTA-17257

---

Because this is very Kubernetes-specific behavior, I've opted to try to expose the fields without modification (including leaving them camel cased).  I am also slightly worried that this is a bit brittle because we don't currently validate the autoscaling section of soa-configs, and it might be easy to get it wrong which could mess up setup_kubernetes_job...but I also don't expect this to be commonly-used behavior and we need it now so I kind of vote we just do it?

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -111,6 +111,7 @@ from kubernetes.client.rest import ApiException
 from mypy_extensions import TypedDict
 
 from paasta_tools.async_utils import async_timeout
+from paasta_tools.long_running_service_tools import AutoscalingParamsDict
 from paasta_tools.long_running_service_tools import host_passes_blacklist
 from paasta_tools.long_running_service_tools import host_passes_whitelist
 from paasta_tools.long_running_service_tools import InvalidHealthcheckMode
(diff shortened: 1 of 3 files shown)


thebostik (approved):
seems fine to solve the problem at hand. do we think this is still the solution in the medium/long term, that is, would this be replaced with or added on to what luisp is working on?

matfra:
I'm fine going with this temporary solution so we can validate than tuning stabilizationWindowSeconds works well in production for service that needs it. Maybe it'll end up not working as we expect and we'll end needing custom metrics/Prometheus integration but it's worth a shot.
We can add comments in the config saying "DO NOT TRY TO USE THIS AT HOME"

matfra (approved):

stug:
Yeah I definitely think this is a temporary solution until we get moving average support for cpu metrics from the prometheus metrics adapter.  I think it'd take some consideration to figure out a nice, paasta-y way to incorporate these settings into yelpsoa-configs, and that just doesn't seem worth it for what I am imagining is more of an emergency tool.

kawaiwanyelp (approved):
I agree that this is pretty brittle, especially without validation. This is fine as a temporary solution, but once we have a better alternative, I'd like to see this either deprecated or fixed "properly". The only thing I think we should add is additional comments to mark this as a quick n dirty fix.

PR #3045, opened 2021-02-27
Add --volume parameter to local-run. PAASTA-1262, PAASTA-16089

EvanKrall:
We suggest people use `paasta local-run` for ad-hoc batch jobs, but currently there's no way to add arbitrary volume mounts. This means our best advice for users who want to pass an input file to a batch job is to make a yelpsoa-configs change with an `extra_volumes` entry, which is pretty clunky.

This change adds a `--volume` parameter. (Note that `-v` already exists and means verbose.)

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -466,6 +466,17 @@ def add_subparser(subparsers):
         required=False,
         default=None,
     )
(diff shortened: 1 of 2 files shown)


stug (approved):
seems good to me, but as we were saying in slack maybe let's make sure this isn't a security issue first

PR #3046, opened 2021-03-03
Implement CPU-based moving average window autoscaling

nemacysts:
This additionally allows us to use `cpu` as a metrics provider instead of continuing to propagate `mesos_cpu` when we're not using Mesos anymore.

NOTE: we'll still continue to use the built-in CPU scaling that k8s provides for safety.

### DIFF
--- a/paasta_tools/setup_prometheus_adapter_config.py
+++ b/paasta_tools/setup_prometheus_adapter_config.py
@@ -292,9 +292,9 @@ def create_instance_cpu_scaling_rule(
         avg(
             irate(
                 container_cpu_usage_seconds_total{{
-                    namespace="paasta",
-                    container="{sanitized_instance_name}",
-                    paasta_cluster="{paasta_cluster}"
+                    namespace='paasta',
+                    container='{sanitized_instance_name}',
+                    paasta_cluster='{paasta_cluster}'
                 }}[1m]
             )
         ) by (pod, container)
@@ -303,13 +303,13 @@ def create_instance_cpu_scaling_rule(
     cpus_available = f"""
         sum(
             container_spec_cpu_quota{{
-                namespace="paasta",
-                container="{sanitized_instance_name}",
-                paasta_cluster="{paasta_cluster}"
+                namespace='paasta',
+                container='{sanitized_instance_name}',
+                paasta_cluster='{paasta_cluster}'
             }}
             / container_spec_cpu_period{{
-                namespace="paasta",
-                paasta_cluster="{paasta_cluster}"
+                namespace='paasta',
+                paasta_cluster='{paasta_cluster}'
             }}
         ) by (pod, container)
     """
@@ -321,10 +321,10 @@ def create_instance_cpu_scaling_rule(
     # {{deployment}}-{{10 character hex string}}) so that our query only considers the
     # service that we want to autoscale - without this we're only filtering by instance
     # name and these are very much not unique
+    # k8s:pod:info is an internal recording rule that joins kube_pod_info with
+    # kube_pod_status_phase
     pod_info_join = f"""
         on (pod) group_left(kube_deployment) label_replace(
-            # k8s:pod:info is an internal recording rule that joins kube_pod_info with
-            # kube_pod_status_phase
             k8s:pod:info{{
                 created_by_name=~'{deployment_name}.*',
                 created_by_kind='ReplicaSet',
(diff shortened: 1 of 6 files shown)


paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -185,9 +236,156 @@ def create_instance_uwsgi_scaling_rule(
...
+    series_query = f"""
+        kube_deployment_labels{{
+            deployment='{deployment_name}',
+            paasta_cluster='{paasta_cluster}',
+            namespace='paasta'
+        }}
+    """
+
+    cpu_usage = f"""
+        sum(
```
EvanKrall, line 288: Is this `sum()` here because we expect only one `container_cpu_usage_seconds_total` per (pod, container)? Should we `avg()` or `max()` or something instead, just in case we end up with two time series from two different targets?

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -185,9 +236,156 @@ def create_instance_uwsgi_scaling_rule(
...
+            (({cpu_usage}) / ({cpus_available})) * {pod_info_join}
+        ) by (kube_deployment)
+    """
+
+    moving_average_cpu_usage_per_deployment = f"""
+        avg_over_time(
+            ({average_cpu_usage_per_deployment})[{moving_average_window}s:]
+        )
+    """
+
```
EvanKrall, line 352: I think instead of `avg_over_time(avg(...) by (kube_deployment))` we want to structure this like the uwsgi query:

```
avg_over_time(
    sum(
        (({cpu_usage} / {cpus_available})) * {pod_info_join}
    )
) / {current_replicas}
```

That is: calculate the desired replicas based on instantaneous load, smooth that over time, and then divide by the non-smoothed current number of replicas.

Otherwise I think we'll see the oscillations that we fixed in https://github.com/Yelp/paasta/pull/2862

Imagine if we've been underscaled by 20% for ~30 minutes at 80 instances (where ideally we should have 100)

1. `avg_over_time()` would give `1.25*setpoint`, and so we'd scale up by 25% to 100 instances.
2. Current load would then be `1*setpoint`. `avg_over_time()` would still show average load of something like `(1 + 29*1.25) / 30 ~= 1.24`, so we'd scale up by 24% to 124 instances.
3. Current load would then be setpoint/1.24. `avg_over_time()` would give `(1/1.24 + 1 + 28*1.25) / 30  ~= 1.22`, so we'd scale up by 22% to 151 instances
4. Current load would then be setpoint/1.24/1.22. `avg_over_time()` would give `(1/1.24/1.22 + 1/1.24 + 1 + 27*1.25) / 30  ~= 1.20`, so we'd scale up by 20% to 181 instances.
5. and so on
6. Eventually avg_over_time() would drop below 1 and we'd start scaling down.
7. By the time we scale back down to 100, `avg_over_time()` is substantially below 1 (because we've been overscaled for a long time) and so we'll overshoot in the other direction

EvanKrall (changes requested):
Code structure looks fine; my only issue is the query.

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -125,8 +161,21 @@ def parse_args() -> argparse.Namespace:
...
 
 
+def _minify_promql(query: str) -> str:
+    """
+    Given a PromQL query, return the same query without comments and with some whitespace collapsed.
+
+    This is useful for allowing us to nicely format queries in code, but minimize the size of our
+    queries when they're actually sent to Prometheus by the adapter.
+    """
+    comments_removed = re.sub(r"#.*\n", "", query)
```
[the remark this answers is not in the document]
nemacysts: oh, I thought I had added a `\s*` to the beginning of this!

you're right though, there is a chance - lemme mess around with this a bit as I'm assuming I "temporarily" removed that 'cause I ran into an issue somewhere

paasta_tools/setup_prometheus_adapter_config.py
nemacysts: we should only have one series except for bounces - but avg probably makes more sense for handling this

paasta_tools/setup_prometheus_adapter_config.py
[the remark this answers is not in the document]
nemacysts: I thought about this a little more and decided to just stop trying to strip comments - I realized that I also wasn't going to correctly handle inline comments (i.e., comments on the same line as some promql). After refactoring the query into multiple variables this is also less necessary as we can just comment in Python right above any tricky parts.

EvanKrall (approved):

PR #3048, opened 2021-03-03
Make stdout/stderr default in paasta logs, and improve its help text

kawaiwanyelp:
## Description
I'm doing three things:

1. Make stdout/stderr the default components in paasta status. Most users only care about these components anyway, and it will mean the paasta logs by default will return faster, since we only need to look at app_output logs.
2. Remove Mesos/Marathon language from `paasta logs --help`, to avoid confusing users. The only exception is the `marathon` component, which I've listed as deprecated, since we're mainly on k8s anyway.
3. Organize the `paasta logs --help` text better so that more information is in the help text itself, from than the epilog. This should help users understand the components, and use the arg more.

## Testing
Tested by running `paasta logs` in my env. 

## Notes
what paasta logs looks like now: https://fluffy.yelpcorp.com/i/533mRdzpTDlqqS1Lh9Dj7sqljRXGdK1g.html

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -89,18 +89,29 @@ def add_subparser(subparsers) -> None:
         "--service",
         help="The name of the service you wish to inspect. Defaults to autodetect.",
     ).completer = lazy_choices_completer(list_services)
-    components_help = "A comma separated list of the components you want logs for."
     status_parser.add_argument(
-        "-C", "--components", help=components_help
+        "-c",
+        "--clusters",
+        help="The clusters to see relevant logs for. Defaults to all clusters to which this service is deployed.",
+    ).completer = completer_clusters
+    status_parser.add_argument(
+        "-i",
+        "--instances",
+        help="The instances to see relevant logs for. Defaults to all instances for this service.",
+    ).completer = completer_clusters
+    status_parser.add_argument(
+        "-C",
+        "--components",
+        type=lambda s: set(s.split(",")),
+        default=set(DEFAULT_COMPONENTS),
+        help=(
+            "A comma-separated list of the components you want logs for. "
+            "PaaSTA consists of 'components' such as builds and deployments, "
+            "for each of which we collect logs for per service. "
+            "See below for a list of components. "
+            "Defaults to %(default)s."
+        ),
     ).completer = lazy_choices_completer(LOG_COMPONENTS.keys)
-    cluster_help = "The clusters to see relevant logs for. Defaults to all clusters to which this service is deployed."
-    status_parser.add_argument(
-        "-c", "--clusters", help=cluster_help
-    ).completer = completer_clusters
-    instance_help = "The instances to see relevant logs for. Defaults to all instances for this service."
-    status_parser.add_argument(
-        "-i", "--instances", help=instance_help
-    ).completer = completer_clusters
     status_parser.add_argument(
         "-f",
         "-F",
@@ -132,7 +143,7 @@ def add_subparser(subparsers) -> None:
         dest="soa_dir",
         metavar="SOA_DIR",
         default=DEFAULT_SOA_DIR,
-        help="define a different soa config directory",
+        help=f"Define a different soa config directory. Defaults to %(default)s.",
     )
 
     status_parser.add_argument(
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/logs.py
```diff
@@ -69,11 +70,32 @@
...
+    """Custom argparse formatter that wraps each line in argument help text
+    individually, instead of wrapping the entire help text at once. This allows
+    us to maintain newlines where we want them.
+    """
+
+    def _split_lines(self, text, width):
+        wrapped_lines = []
+        for line in text.splitlines():
+            # add some indents if we see a list in help text
+            if line.startswith(TEXTWRAP_LIST_ITEM_PREFIX):
```
qui, line 89: This feels like we're blurring the boundaries of the formatter class. If you had to describe this feature (the extra list formatting) in the class docstring, does it make sense? At the very least, TEXTWRAP_LIST_ITEM_PREFIX should be a member of this class, and its length should be used the next line, at `line[1:]`, rather than assuming its length is 1.

I agree the component list is less readable in the help text without the list formatting/indenting. To me, though, it doesn't seem worth this complexity, and keeping the list of components in the epilog seems good enough.

kawaiwanyelp: I prefer how it looks, but I rolled this change back anyway.

paasta_tools/cli/cmds/logs.py
```diff
@@ -1338,12 +1373,7 @@ def paasta_logs(args: argparse.Namespace) -> int:
         if verify_instances(args.instances, service, clusters):
             return 1
 
-    if args.components is not None:
-        components = args.components.split(",")
-    else:
-        components = DEFAULT_COMPONENTS
-    components = set(components)
-
+    components = set(args.components)
```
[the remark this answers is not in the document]
kawaiwanyelp: Done.

paasta_tools/utils.py
```diff
@@ -1171,32 +1179,6 @@ def default(text: str) -> str:
                 "help": "Logs that tell you about task startup, failures, healthchecks, etc.",
             },
         ),
-        # I'm leaving these planned components here since they provide some hints
-        # about where we want to go. See PAASTA-78.
-        #
-        # But I'm commenting them out so they don't delude users into believing we
-        # can expose logs that we cannot actually expose. See PAASTA-927.
-        #
-        # ('app_request', {
```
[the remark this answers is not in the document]
kawaiwanyelp: I restored the commented out text.

paasta_tools/utils.py
```diff
@@ -1131,29 +1137,31 @@ def default(text: str) -> str:
...
             },
         ),
         (
             "app_output",
             {
                 "color": compose(PaastaColors.yellow, PaastaColors.bold),
-                "help": "Stderr and stdout of the actual process spawned by Mesos. "
-                "Convenience alias for both the stdout and stderr components",
+                "help": (
+                    "Stderr and stdout running containers. "
```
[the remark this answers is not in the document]
kawaiwanyelp: Done.

qui (approved):

paasta_tools/cli/cmds/logs.py
```diff
@@ -89,18 +89,29 @@ def add_subparser(subparsers) -> None:
...
+        "-C",
+        "--components",
+        type=lambda s: set(s.split(",")),
+        default=set(DEFAULT_COMPONENTS),
+        help=(
+            "A comma-separated list of the components you want logs for. "
+            "PaaSTA consists of 'components' such as builds and deployments, "
+            "for each of which we collect logs for per service. "
+            "See below for a list of components. "
+            f"Defaults to {DEFAULT_COMPONENTS}."
```
nemacysts, line 112: could also do `"Defaults to %(default)s"` to have argparse handle interpolating for you if you wanted to :)

nemacysts (approved):

paasta_tools/setup_kubernetes_job.py
```diff
@@ -48,11 +48,11 @@
 
 
 def parse_args() -> argparse.Namespace:
-    parser = argparse.ArgumentParser(description="Creates marathon jobs.")
+    parser = argparse.ArgumentParser(description="Creates Kubernetes jobs.")
```
stug, line 51: ha

stug (approved):
Looks roughly good to me!

paasta_tools/cli/cmds/logs.py
kawaiwanyelp: TIL

PR #3049, opened 2021-03-04
Make paasta status print what it can even if an error_message attribu…

EvanKrall:
…te exists. Make the error_message for autoscaling issues a little more descriptive. PAASTA-17298

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1132,11 +1132,6 @@ def print_kubernetes_status(
     kubernetes_status,
     verbose: int = 0,
 ) -> int:
-    error_message = kubernetes_status.error_message
-    if error_message:
-        output.append(error_message)
-        return 1
-
     bouncing_status = bouncing_status_human(
         kubernetes_status.app_count, kubernetes_status.bounce_method
     )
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -1132,11 +1132,6 @@ def print_kubernetes_status(
     kubernetes_status,
     verbose: int = 0,
 ) -> int:
-    error_message = kubernetes_status.error_message
```
nemacysts, line 1135: do we need to do anything on https://github.com/Yelp/paasta/blob/c12c315aacf8a1765c37f0432bbb0df23f2a6712/paasta_tools/cli/cmds/status.py#L1184 since https://github.com/Yelp/paasta/pull/3049/files#diff-92b6bc9b65efee17b175778ac635c201f717dec51361d8f16b7c17a85dbe579aR481 means that we won't ever set the autoscaling status? or does whatever turns kubernetes_status from a dict to the object that we're using here handle setting that to `None` for us?

nemacysts (approved):

paasta_tools/cli/cmds/status.py
EvanKrall: yeah, `kubernetes_status.autoscaling_status` will be None if it's not included in the response.

EvanKrall: (Except for the extra `error_message` attribute, I think the response basically looks the same as a normal response for a service that isn't autoscaled.)

PR #3052, opened 2021-03-08
Paasta 17286 new paasta status basic

stug:
This adds
* A url query param `?new` to the existing status API endpoint which returns data for the new version of paasta status for services
* A `--new` argument to paasta status which requests data from that endpoint and formats it
Example output:
```
service: example_happyhour
cluster: norcal-devc
    instance: main
    Git sha:    16053fab (desired)
    State: Running
    Running versions:
      16053fab - Started 2021-02-23 10:11:40 (13 days ago)
        Replica States: 3 Running
```

This sprinkles "v2" everywhere, which is a little annoying and will eventually need to be removed if we go through with this.

Other questions and things to note:
* I am trying to leave as much of the decision-making logic in the client so that if we want to tweak things we don't end up having to make multiple PRs
* A major remaining piece will be implementing the `get_replica_states` method, including providing it with the data it needs to do its job
* You should check out the `get_instance_state` method that I wrote to explain what state the instance as a whole is in.  I just tried to take a stab, but I imagine this will require some iteration.


paasta_tools/cli/cmds/status.py
```diff
@@ -1124,6 +1132,112 @@ def print_flink_status(
     return 0
 
 
+def print_kubernetes_status_v2(
+    cluster: str,
+    service: str,
+    instance: str,
+    output: List[str],
+    status,
```
nemacysts, line 1140: should we annotate `status` here, `replicasets` at https://github.com/Yelp/paasta/pull/3052/files?file-filters%5B%5D=.py#diff-36175af498b294330c337132851420b6d63c7469e72162fb6a4496fcd5cbaff2R1187 (and further below) and `pods` at https://github.com/Yelp/paasta/pull/3052/files?file-filters%5B%5D=.py#diff-36175af498b294330c337132851420b6d63c7469e72162fb6a4496fcd5cbaff2R1237? 

https://mypy.readthedocs.io/en/latest/common_issues.html?highlight=reveal_type#displaying-the-type-of-an-expression will probably help get the types of these

nemacysts: I would assume that `replicasets` is `V1ReplicaSet` and `pods` is `V1Pod` tho

paasta_tools/instance/kubernetes.py
```diff
@@ -407,6 +407,96 @@ def filter_actually_running_replicasets(
...
+        "replicas": replicaset.spec.replicas,
+        "ready_replicas": ready_replicas_from_replicaset(replicaset),
+        "create_timestamp": replicaset.metadata.creation_timestamp.timestamp(),
+        "git_sha": replicaset.metadata.labels.get("paasta.yelp.com/git_sha"),
+        "config_sha": replicaset.metadata.labels.get("paasta.yelp.com/config_sha"),
+        "pods": [get_pod_status(pod) for pod in pods],
+    }
+
+
+def get_pod_status(pod: V1Pod) -> Dict[str, Any]:
```
nemacysts, line 492: super-nit: should we make a `PodStatus` `TypedDict` so that we can replace this `Any`?

nemacysts: or a `NamedTuple`/`dataclass`, I guess :p

paasta_tools/instance/kubernetes.py
```diff
@@ -407,6 +407,96 @@ def filter_actually_running_replicasets(
...
+        kube_client=kube_client,
+        namespace=job_config.get_kubernetes_namespace(),
+    )
+    pods_by_replicaset = get_pods_by_replicaset(pod_list)
+    replicaset_list = kubernetes_tools.replicasets_for_service_instance(
+        service=job_config.service,
+        instance=job_config.instance,
+        kube_client=kube_client,
+        namespace=job_config.get_kubernetes_namespace(),
+    )
```
kawaiwanyelp, line 445: Even though it's not widely used, we can have statefulsets instead of replicasets here, so we should really be sorting pods by either statefulset or replicaset. 

You can maybe punt on this. However, keep in mind that since the current paasta status doesn't connect replicasets/statefulsets to pods, it's okay to leave out statefulsets. But, since we want to connect it all in a table, punting would probably mean not showing any table at all.

paasta_tools/cli/cmds/status.py
```diff
@@ -125,6 +125,13 @@ def add_subparser(subparsers,) -> None:
         default=DEFAULT_SOA_DIR,
         help="define a different soa config directory",
     )
+    status_parser.add_argument(
+        "--new",
```
kawaiwanyelp, line 129: nit: `--experimental` instead? `--new` feels like it applies officially released and supported.

paasta_tools/cli/cmds/status.py
```diff
@@ -1124,6 +1132,112 @@ def print_flink_status(
...
+    start_datetime = datetime.fromtimestamp(replicaset.create_timestamp)
+    humanized_start_time = humanize.naturaltime(start_datetime)
+    entry = [f"{version_name} - Started {start_datetime} ({humanized_start_time})"]
+    replica_states = get_replica_states(replicaset.pods)
+    entry.append(f"  Replica States: {replica_states}")
+    return entry
+
+
+def get_replica_states(pods) -> str:
+    return PaastaColors.green(f"{len(pods)} Running")
```
kawaiwanyelp, line 1238: nit: Add a todo comment here? It is slightly confusing to see an always green value, when we're really going to be adding many more.

paasta_tools/cli/cmds/status.py
```diff
@@ -1124,6 +1132,112 @@ def print_flink_status(
...
+    else:
+        return 0
+
+
+def get_instance_state(status) -> str:
+    num_replicasets = len(status.replicasets)
+    num_ready_replicas = sum(r.ready_replicas for r in status.replicasets)
+    if status.desired_state == "stop":
+        if num_replicasets == 0:
+            return PaastaColors.red("Stopped")
```
kawaiwanyelp, line 1160: All the states should organized in a class somewhere or put as global constants somewhere for consistency.

paasta_tools/cli/cmds/status.py
```diff
@@ -1124,6 +1132,112 @@ def print_flink_status(
...
+            git_shas = {r.git_sha for r in replicasets}
+            config_shas = {r.config_sha for r in replicasets}
+            bouncing_to = []
+            if len(git_shas) > 1:
+                bouncing_to.append(replicasets[0].git_sha[:8])
+            if len(config_shas) > 1:
+                bouncing_to.append(replicasets[0].config_sha)
+
+            bouncing_to_str = ", ".join(bouncing_to)
+            return PaastaColors.yellow(f"Bouncing to {bouncing_to_str}")
```
kawaiwanyelp, line 1182: The bouncing state could be: `Bouncing to abc123, configdef456`, which could be confusing. The git SHA on it's own would be intuitive, but the on it's own or with the git SHA seems confusing. Could we explain this more or not show it somehow? Maybe say "due to a SOA configs change"? Hmm, but it could due to a big bounce too...

kawaiwanyelp (changes requested):

qui (approved):

paasta_tools/api/views/instance.py
```diff
@@ -619,6 +619,7 @@ def instance_status(request):
     service = request.swagger_data.get("service")
     instance = request.swagger_data.get("instance")
     verbose = request.swagger_data.get("verbose") or 0
+    use_new = request.swagger_data.get("new") or False
```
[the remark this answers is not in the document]
stug: No they are subtly different:
```
>>> {"new": None}.get("new", False) is None
True
>>> {"new": None}.get("new") or False
False
```

paasta_tools/cli/cmds/status.py
stug: i guess id rather not given that i'd have to change this all over the place

paasta_tools/cli/cmds/status.py
stug: This is the data being returned by the Paasta API, so they're actually OpenAPI generated models, but I can look into annotating -- shouldn't be too hard

paasta_tools/instance/kubernetes.py
stug: hmm yeah good catch -- I will either fix this or file a follow-up

paasta_tools/cli/cmds/status.py
stug: This would be a bit of a pain to change at this point, so I am going to skip this unless you feel strongly

paasta_tools/cli/cmds/status.py
stug: yeah can do

paasta_tools/instance/kubernetes.py
```diff
@@ -407,6 +407,96 @@ def filter_actually_running_replicasets(
...
+        get_replicaset_status(
+            replicaset, pods_by_replicaset.get(replicaset.metadata.name),
+        )
+        for replicaset in actually_running_replicasets
+    ]
+    return status
+
+
+def get_pods_by_replicaset(pods: Sequence[V1Pod]) -> Dict[str, List[V1Pod]]:
+    # TODO: should we make sure that we get every pod...?
```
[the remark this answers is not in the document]
stug: yeah -- i think _probably_ we can just trust kubernetes?

paasta_tools/cli/cmds/status.py
```diff
@@ -1124,6 +1132,112 @@ def print_flink_status(
...
+        return 1
+    else:
+        return 0
+
+
+def get_instance_state(status) -> str:
+    num_replicasets = len(status.replicasets)
+    num_ready_replicas = sum(r.ready_replicas for r in status.replicasets)
+    if status.desired_state == "stop":
+        if num_replicasets == 0:
```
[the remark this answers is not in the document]
stug: hmm good call

paasta_tools/instance/kubernetes.py
stug: Yeah that would be nice but I guess I don't have time before I leave on PTO :-/

paasta_tools/cli/cmds/status.py
stug: yeah I am not sure.  I guess I would like to leave this as-is for now and we can revisit it

paasta_tools/instance/kubernetes.py
stug: Added a TODO

paasta_tools/cli/cmds/status.py
stug: actually I need to get out of here before PTO, so I left a TODO, sorry :-/

stug:
I've mostly left TODOs for this stuff, and I'm going to merge it so that other folks can work on it if they have time.  Otherwise I'll come back to this after PTO.

PR #3055, opened 2021-03-10
Add a preStop config for uwsgi-exporter

nemacysts:
We're currently dropping some metrics since this sidecar is getting
terminated before the main container.


EvanKrall (approved):
needs testing on kubestage I guess

PR #3058, opened 2021-03-11
Don't print bytes objects when tailing local-run container stdout/stderr

stug:
I think this probably has been messed up since the transition to python3?

See PAASTA-17039

---

I tested that this works with a local-run.  The `end=""` is needed so that we don't add extra newlines since the stdout/err already has newlines (I figure this is slightly better than `.strip()`?)


tzhu-yelp (approved):

EvanKrall (approved):

EvanKrall:
(maybe fix the PR title since the code doesn't seem to be decoding byte strings)

PR #3059, opened 2021-03-11
For services not in the mesh, set a readiness check that is the same …

stug:
…as their liveness check (which is configured by settings in soa configs)

Without this, services not in the mesh that set a healthcheck might get restarted for failing their liveness check, but Kubernetes will think they are ready for the purposes of bouncing.  This also means that we won't send replication alerts to the owning team.

See PAASTA-17303

---

I probably want to test this on kubestage before deploying it.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1089,6 +1089,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             ),
             name=self.get_sanitised_instance_name(),
             liveness_probe=self.get_liveness_probe(service_namespace_config),
+            readiness_probe=self.get_readiness_probe(service_namespace_config),
             ports=[V1ContainerPort(container_port=port) for port in ports],
             security_context=self.get_security_context(),
             volume_mounts=self.get_volume_mounts(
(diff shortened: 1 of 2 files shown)


kennydo (approved):
I'm not familiar with the details, but the description of the solution LGTM

stug:
FYI I am on PTO next week and haven't had a chance to test this.  I can take a look when I get back, or someone else on CI can push this the rest of the way if you need

paasta_tools/kubernetes_tools.py
```diff
@@ -1105,6 +1106,14 @@ def get_kubernetes_containers(
         )
         return containers
 
+    def get_readiness_probe(
+        self, service_namespace_config: ServiceNamespaceConfig
+    ) -> Optional[V1Probe]:
+        if service_namespace_config.is_in_smartstack():
+            return None
```
EvanKrall, line 1113: aside from causing a big bounce, how do we feel about always returning the liveness probe? I think this would most closely match setup_marathon_job's behavior, which considered both the service's healthcheck (service's liveness check) and whether it's in smartstack (hacheck container's readiness check).

EvanKrall (approved):

paasta_tools/kubernetes_tools.py
stug: I think I am roughly down with that.  That would mean that each service would need to respond to healthcheck 13 times a minute -- every 10s for both the liveness and readiness probes plus once per minutes from hacheck.  Do you think that is ok?

EvanKrall: I think our hacheck is configured to [cache for only 1 second](https://github.com/Yelp/paasta/blob/7eba1996de7a51a016536e2ee4b2f3c9de413c09/yelp_package/dockerfiles/hacheck-sidecar/hacheck.conf.yaml#L1).

PR #3062, opened 2021-03-12
Make paasta validate check that service mesh registrations are decomp…

stug:
…osable via decompose_job_id

See PAASTA-17314

---

Wasn't sure if we still needed warning logging in `get_registrations` (since validate should hopefully catch most of this and it obviously didn't help stop the issue I saw yesterday), but I figured why not.

### DIFF
--- a/paasta_tools/long_running_service_tools.py
+++ b/paasta_tools/long_running_service_tools.py
@@ -201,15 +201,13 @@ class LongRunningServiceConfig(InstanceConfig):
         return decompose_job_id(self.get_registrations()[0])[1]
 
     def get_registrations(self) -> List[str]:
+        for registration in self.get_invalid_registrations():
+            log.error(
+                "Provided registration {} for service "
+                "{} is invalid".format(registration, self.service)
+            )
+
         registrations = self.config_dict.get("registrations", [])
-        for registration in registrations:
-            try:
-                decompose_job_id(registration)
-            except InvalidJobNameError:
-                log.error(
-                    "Provided registration {} for service "
-                    "{} is invalid".format(registration, self.service)
-                )
 
         # Backwards compatibility with nerve_ns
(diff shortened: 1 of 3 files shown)


paasta_tools/long_running_service_tools.py
```diff
@@ -345,6 +353,19 @@ def get_autoscaling_params(self) -> AutoscalingParamsDict:
             defaults=default_params,
         )
 
+    def validate(self, params: Optional[List[str]] = None,) -> List[str]:
+        error_messages = super().validate()
```
EvanKrall, line 357: should you pass along `params` here?

paasta_tools/long_running_service_tools.py
```diff
@@ -220,6 +218,16 @@ def get_registrations(self) -> List[str]:
 
         return registrations or [compose_job_id(self.service, self.instance)]
 
+    def get_invalid_registrations(self) -> List[str]:
+        registrations = self.config_dict.get("registrations", [])
+        invalid_registrations: List[str] = []
+        for registration in registrations:
+            try:
+                decompose_job_id(registration)
```
EvanKrall, line 226: If you wanted, I think you could go further and check that the first element of the return value of `decompose_job_id` is a valid service name, though I'm not 100% sure that's a strict requirement right now.

EvanKrall (approved):

paasta_tools/long_running_service_tools.py
stug: Oh yeah, added that late after mypy complained about it.  I will fix.

paasta_tools/long_running_service_tools.py
stug: Yeah I guess I just want to ship this if that is ok

PR #3069, opened 2021-03-18
Gate signalfx HPA metrics behind a default-true toggle. PAASTA-17097

EvanKrall:
Before we delete this, let's make a toggle for it and try it on a few instances.


nemacysts (approved):
we should switching completely pretty soon right? otherwise, think its worth protecting against someone doing something silly and turning both providers off?

EvanKrall:
Yeah, I'm not sure exactly what would happen if you disabled both. I don't think the Paasta code would crash, but I dunno if the HPA would dislike having zero metrics defined.

PR #3072, opened 2021-03-22
PAASTA-17310- Replace Hostname w/ Pod name and Container ID

lanceleih:
### Description
As part of the changes to shorten log lines, we are replacing the host to the pod name and container ID. I've also removed 
the "[debug]" keyword in every log line as every log line is actually from debug. As it turns out, `debug` is the only level that we actually output.

For this reason, I've also removed `test_prettify_log_line_valid_json_requested_level_is_only_event`inside test_cmds_logs.py.

### Test
`make test` - PASSED
`tox -e general_tests` - PASSED


paasta_tools/cli/cmds/logs.py
```diff
@@ -462,15 +461,16 @@ def prettify_log_line(
...
                     {
                         "timestamp": prettify_timestamp(parsed_line["timestamp"]),
                         "component": prettify_component(parsed_line["component"]),
                         "cluster": "[%s]" % parsed_line["cluster"],
                         "instance": "[%s]" % parsed_line["instance"],
-                        "level": "%s" % pretty_level,
-                        "message": parsed_line["message"],
+                        "pod": "%s" % parsed_line["pod_name"],
+                        "container": "%s"
+                        % re.findall(r"\(.*?\)", parsed_line["message"])[0],
```
stug, line 473: This looks to me like it is replacing the `level` and `message` with `pod` and `container`.  Also for app_output log lines I think you have access to the container name in the `container_name` field, so you don't really need any regex shenanigans.

Most importantly, though, it looks to me like the hostname is included in the `message` field (based on the one service whose log lines I peeked at), meaning that we can't really get rid of it...so I am wondering if this ticket actually makes sense.

stug (changes requested):

paasta_tools/cli/cmds/logs.py
stug: I guess it probably still makes sense to get rid of level, though?

kawaiwanyelp: +1

kawaiwanyelp (changes requested):

paasta_tools/cli/cmds/logs.py
lanceleih: Only removing level and keeping message as long as pod and container.

Follow up actions: Enforce user to input only one of each (service, instance and cluster). By doing so, we will be able to remove those fields from the log lines, and thus saving space for both pod and container.

paasta_tools/cli/cmds/logs.py
```diff
@@ -461,18 +471,22 @@ def prettify_log_line(
                 }
             )
         else:
+            log_line = {
+                "timestamp": prettify_timestamp(parsed_line["timestamp"]),
+                "component": prettify_component(parsed_line["component"]),
+                "cluster": " [%s]" % parsed_line["cluster"],
```
kawaiwanyelp, line 477: nit: `"" if strip_cluster else f" [{parsed_line['cluster']}]"`? same for instance

(just inlining the if-else, and using f-strings, which is more py3)

paasta_tools/cli/cmds/logs.py
```diff
@@ -744,6 +758,9 @@ def callback(
             clusters=clusters, components=components, callback=callback
         )
 
+        strip_cluster = True if len(clusters) < 2 else False
```
kawaiwanyelp, line 761: nit: just `strip_cluster = len(clusters) < 3`? similar for `strip_instance`

paasta_tools/cli/cmds/logs.py
```diff
@@ -744,6 +758,9 @@ def callback(
...
+        strip_instance = True if instances is not None else False
```
kawaiwanyelp, line 762: I think we also want `len(instances) < 2` here? I don't think `instances` can be `None` here either because it's an `Iterable`, not an `Optional`.

paasta_tools/cli/cmds/logs.py
```diff
@@ -1346,11 +1392,19 @@ def paasta_logs(args: argparse.Namespace) -> int:
     else:
         clusters = args.clusters.split(",")
 
+        if len(clusters) > 1:
+            print("You can only filter one cluster.")
+            return 1
+
```
kawaiwanyelp, line 1398: Instead of constraining clusters and instances here, I think you should just change the arg parser to only allow one arg instead of being an narg/multi-arg.

kawaiwanyelp (changes requested):

paasta_tools/cli/cmds/logs.py
stug: Seems like we don't need this logic anyway since we are only allowing one cluster and one instance?

stug (commented):
I roughly agree with charle's suggestions but otherwise looks good to me I think

paasta_tools/cli/cmds/logs.py
lanceleih: @kawaiwanyelp `instances` can actually be `None` in this case. This happens when the user does want to specify an instance. 

See line 1401: `instances = None`

`clusters`, on the other hand, would never be `None` because:
1. If the user doesn't specify a `cluster`, we will list all the clusters in line 1392.
2. If the user does specify a `cluster`, there will be at least one because we are filtering specifically for that `cluster`.

This answers half of @stug 's question. In terms of the instance logic, I think it's plausible to remove it and just use instances itself as a boolean to check whether or not we need to strip the instance header. I guess I like to leave it in for readability.

paasta_tools/cli/cmds/logs.py
lanceleih: Done

paasta_tools/cli/cmds/logs.py
lanceleih: I wasn't using nargs for clusters and instances initially, but I've changed it to have that option which makes the arguments one list of one item for both clusters and instances.

Example for multiple clusters referenced:
`(py36-linux) lancelei@dev54-uswest1adevc:~/Repos/paasta$ paasta logs -s mtb_ranking -c norcal-devc,pnw-devc
Fetching 100 lines and applying filters. Try -n 1000 for more lines...
I don't know where scribe logs for norcal-devc,pnw-devc live?`

Example for multiple instances referenced:
`(py36-linux) lancelei@dev54-uswest1adevc:~/Repos/paasta$ paasta logs -s mtb_ranking -c norcal-devc -i long_timeout,main
mtb_ranking doesn't have any instances matching long_timeout,main on norcal-devc.
Did you mean any of these?
  long_timeout`

paasta_tools/cli/cmds/logs.py
```diff
@@ -1388,8 +1391,8 @@ def paasta_logs(args: argparse.Namespace) -> int:
             service=service,
             levels=levels,
             components=components,
-            clusters=clusters,
-            instances=instances,
+            clusters=cluster,
+            instances=instance,
```
stug, line 1395: It feels a little weird to me that this still works.  For clusters, I think the issue is that you are setting `nargs=1` for the `--cluster` argument, which causes it to return a 1-element list instead of a string.  I think that is roughly ok, but it'd be better if we just told it that `--cluster` was a `str` and then passed `clusters=[args.cluster]` or whatever here.  This isn't a huge deal, but would improve clarity a bit.

For instances, I think there are actually two weird things going on here that we should fix.  `instance` is now a `str` whereas previously we were passing a list of strings.  However, all the logs functions specify that `instances` is an `Iterable[str]`.  What they meant was something like `['instance1', 'instance2']`, but because strings are actually made up of shorter strings, it turns out that just a regular old string like `'instance1'` is actually also an `Iterable[str]`.  So this is passing its type check, but for counter intuitive reasons.

Even weirder, I was wondering how this was passing the `filter_fn` which checks that `parsed_line.get("instance") in instances` (see `paasta_app_output_passes_filter`).  Before, `instances` would have been a list of strings.  Now it is just a string.  But it all ends up working because the `in` is now checking whether the instance from the log line is a substring of the instance provided as an argument.  Very weird.

Anyway, I think we should do the same thing for instances -- pass `instances=[instance]` to these functions so that they behave the way we expect them to.

stug (approved):
I think this looks good, but I think we should change how we are passing the clusters and instances around a bit for clarity.

stug:
Also it looks like you now have some merge conflicts (probably with your other branch) that you'll need to fix.

paasta_tools/cli/cmds/logs.py
lanceleih: Removed `strip_cluster` and `strip_instance` altogether as it is unnecessary.

PR #3077, opened 2021-03-24
Make new paasta status for service instances work with StatefulSets i…

stug:
…n addition to ReplicaSets

---
Asking, like, everyone for feedback.

Basically I renamed the `replicasets` part of the api return to `versions` and added a `type` field, which will either have the value `StatefulSet` or `ReplicaSet`.  StatefulSets are a little bit tricky because when they are updated, there is no new object created.  That means that we can't figure out (or at least I couldn't figure out!) when a new version was created, and all we have to go on is which pods are tagged with with git and config sha.

Let me know your thoughts!


paasta_tools/instance/kubernetes.py
```diff
@@ -499,6 +518,42 @@ def get_pod_status(pod: V1Pod) -> Dict[str, Any]:
...
+    ready_pods = [pod for pod in pods if kubernetes_tools.is_pod_ready(pod)]
+    first_pod_created = min(
+        [pod.metadata.creation_timestamp.timestamp() for pod in pods]
+    )
+    return {
+        "name": None,
+        "type": "StatefulSet",
+        "replicas": len(pods),
+        "ready_replicas": len(ready_pods),
+        "create_timestamp": first_pod_created,
```
stug, line 550: A couple things I'd like feedback on here:
1.  it's a little misleading to give these version entries the type "StatefulSet" since they are actually subsets of the pods in the StatefulSet.  `"type": "StatefulSubset"`?  😛 
2. I'm using the min creation timestamp of the pods for a given version as the creation timestamp for the version...but that could easily be wrong with pod turnover.  But without this we can't really know which version is newer.

jfongatyelp: 1. Oh, I was just assuming type was just referencing what kind of k8s resource these pods belong to, not the type of this version 'object' itself, so it made perfect sense to stick as StatefulSet
2. Since our new status really only distinguishes the latest version as 'new' and the rest are all lumped up as 'old', it doesn't seem like it'd be too terrible a problem -- Do we know what the statefulset controller does during an update anyhow? If a older pod goes away before it is upgraded, does it get spun back up on the old version?

kawaiwanyelp: You could probably use `ControllerRevision`s, which contain update metadata for statefulsets.

kawaiwanyelp (commented):
looks good overall. I think you should use `ControllerRevision`s for figuring out when statefulsets were updated though.

paasta_tools/instance/kubernetes.py
stug: Yep you are right, and we even get multiple ControllerRevisions during a bounce.  I changed it to use this.

paasta_tools/instance/kubernetes.py
```diff
@@ -623,6 +654,47 @@ def get_pod_containers(pod: V1Pod) -> List[Dict[str, Any]]:
...
+    return versions
+
+
+def get_version_for_controller_revision(
+    cr: V1ControllerRevision, pods: Sequence[V1Pod], backends: Optional[Set[str]],
+) -> KubernetesVersionDict:
+    ready_pods = [pod for pod in pods if kubernetes_tools.is_pod_ready(pod)]
+    return {
+        "name": cr.metadata.name,
+        "type": "ControllerRevision",
```
kawaiwanyelp, line 688: nit: Should this be changed to `StatefulSet`, since that and `ReplicaSet` are the listed version types in the Swagger and OAPI spec?

kawaiwanyelp (approved):
lgtm, just one nit.

paasta_tools/instance/kubernetes.py
stug: oh yeah.  or maybe i should change them in the specs, honestly.  will fix before i ship

PR #3080, opened 2021-03-26
Shorten replication alert output, and add tip + runbook, and add OOM alert tip

kawaiwanyelp:
## Description
This PR does a number of things:

- Adds `--dry-run` for the oom and replication event checker
- Updates the tip for the oom checker
- Moves a lot of context from the output to the description of replication alert events. This is because the output was so long that it would prevent the tip and runbook from showing up in Slack alerts (400 char limit). The description was also not well-used. 
- Changes the default replication alert runbook to y/paastatools for help debugging services (we should probably create a dedicated runbook or section for this)

## Testing
`make test`
manual testing using `--dry-run` to see the sensu events
manual testing and seeing slack alert messages

### DIFF
--- a/paasta_tools/check_services_replication_tools.py
+++ b/paasta_tools/check_services_replication_tools.py
@@ -27,6 +27,7 @@ import a_sync
 from marathon import MarathonClient
 from marathon.models.task import MarathonTask
 from mypy_extensions import Arg
+from mypy_extensions import NamedArg
 
 from paasta_tools.kubernetes_tools import get_all_nodes
 from paasta_tools.kubernetes_tools import get_all_pods
@@ -59,6 +60,7 @@ CheckServiceReplication = Callable[
         Arg(InstanceConfig_T, "instance_config"),
         Arg(Sequence[Union[MarathonTask, V1Pod]], "all_tasks_or_pods"),
         Arg(Any, "replication_checker"),
+        NamedArg(bool, "dry_run"),
     ],
     Optional[bool],
 ]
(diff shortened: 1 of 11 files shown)


stug (approved):
I think this roughly makes sense to me.  Can you maybe post what the alerts look like after this change?

kawaiwanyelp:
Here's what the replication alert looks like.
Before:
Screen Shot 2021-03-31 at 1 45 31 PM
After:
Screen Shot 2021-03-31 at 1 45 14 PM

paasta_tools/monitoring_tools.py
```diff
@@ -55,6 +55,8 @@ def get_replication_for_instance(
     yelp_meteorite = None
 
 
+DEFAULT_REPLICATION_RUNBOOK = "y/paastatools"
```
qui, line 58: I agree y/paastatools isn't the best page for this. I think Section 3 of y/stuckbounce covers the same steps you would use for under-replication. Maybe you could pull that out into a separate page and link to it from both y/stuckbounce and the alerts.

qui (approved):

paasta_tools/monitoring_tools.py
kawaiwanyelp: Done. I created y/unhealthy-paasta-instances.

paasta_tools/check_services_replication_tools.py
```diff
@@ -150,13 +160,15 @@ def check_services_replication(
...
+    pct_under_replicated: float, cluster: str, scheduler: str, dry_run: bool = False,
 ) -> None:
-    meteorite_dims = {"paasta_cluster": cluster, "scheduler": scheduler}
-    gauge = yelp_meteorite.create_gauge(
-        "paasta.pct_services_under_replicated", meteorite_dims
-    )
-    gauge.set(pct_under_replicated)
+    metric_name = "paasta.pct_services_under_replicated"
+    if dry_run:
+        print(f"Would've sent value {metric_name} for metric '{pct_under_replicated}'")
```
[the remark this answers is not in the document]
lanceleih: +1 - Other than switching them, it looks good to me!

lanceleih (approved):

PR #3081, opened 2021-03-26
Remove swagger.json pre-commit check & fix pre-commit global exclude

jfongatyelp:
Unfortunately pre-commit's global `exclude` patterns are `AND` ed with any per-hook excludes, meaning that we would never have run the oapi-codegen hook against `paasta_tools/paastaapi` files that were modified.

I've removed the global exclude and added `paasta_tools/paastaapi` to other python-specific hooks' excludes.

In addition, as per @keymone 's [comment](https://github.com/Yelp/paasta/pull/3052#discussion_r591113114) in #3052, this PR removes the part of this pre-commit that checked that oapi.yaml and swagger.json were both modified, including renaming  `check_swagger_oapi_sync.sh` to `check_manual_oapi_changes.sh` and updating the description of the hook.

I verified that `pre-commit run --all-files` did not modify anything in paasta_tools/paastaapi after this change, and verified the `check_manual_oapi_changes` hook did capture a modification to one of the models e.g.


paasta_tools/contrib/check_manual_oapi_changes.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+set -uo pipefail
+
+diff_names=$(git diff origin/master --name-only)
+touched_schemas=$(
+    echo "$diff_names" |
+    grep 'paasta_tools/api/api_docs/oapi.yaml' |
+    wc -l)
```
keymone, line 8: this isn't used in the script

paasta_tools/contrib/check_manual_oapi_changes.sh
```diff
@@ -0,0 +1,24 @@
...
+
+schema_touched=$(echo "$diff_names" | grep 'paasta_tools/api/api_docs/oapi.yaml')
+api_touched=$(echo "$diff_names" | grep 'paasta_tools/paastapi')
+
+if [ ! -z "$(echo "$diff_names" | grep paasta_tools/paastaapi)" -a -z "$schema_touched" ]; then
```
keymone, line 13: `"$(echo "$diff_names" | grep paasta_tools/paastaapi)"` is the same as `"$api_touched"`

keymone (approved):
minor nits

PR #3089, opened 2021-03-31
If DEFAULT_USE_PROMETHEUS_UWSGI is true, run uwsgi_exporter when use_…

EvanKrall:
…prometheus is unspecified.

### DIFF
--- a/tests/test_kubernetes_tools.py
+++ b/tests/test_kubernetes_tools.py
@@ -767,18 +767,43 @@ class TestKubernetesDeploymentConfig:
             default_should_run_uwsgi_exporter_sidecar=mock.Mock(return_value=False)
         )
 
-        assert (
-            self.deployment.should_run_uwsgi_exporter_sidecar(
-                system_paasta_config_enabled
+        with mock.patch(
+            "paasta_tools.kubernetes_tools.DEFAULT_USE_PROMETHEUS_UWSGI",
+            autospec=False,
+            new=False,
+        ):
+            assert (
+                self.deployment.should_run_uwsgi_exporter_sidecar(
+                    system_paasta_config_enabled
+                )
+                is True
             )
-            is True
-        )
-        assert (
-            self.deployment.should_run_uwsgi_exporter_sidecar(
-                system_paasta_config_disabled
+            assert (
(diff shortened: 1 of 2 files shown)


jfongatyelp (approved):

tests/test_kubernetes_tools.py
```diff
@@ -767,18 +767,43 @@ def test_should_run_uwsgi_exporter_sidecar_defaults(self):
...
+                )
+                is False
+            )
+
+        # If the default for use_prometheus is True and config_dict doesn't specify use_prometheus, we should run
+        # uwsgi_exporter regardless of default_should_run_uwsgi_exporter_sidecar.
+        with mock.patch(
+            "paasta_tools.kubernetes_tools.DEFAULT_USE_PROMETHEUS_UWSGI",
+            autospec=True,
+            new=True,
```
nemacysts, line 793:
```suggestion
```

tests/test_kubernetes_tools.py
```diff
@@ -767,18 +767,43 @@ def test_should_run_uwsgi_exporter_sidecar_defaults(self):
             default_should_run_uwsgi_exporter_sidecar=mock.Mock(return_value=False)
         )
 
-        assert (
-            self.deployment.should_run_uwsgi_exporter_sidecar(
-                system_paasta_config_enabled
+        with mock.patch(
+            "paasta_tools.kubernetes_tools.DEFAULT_USE_PROMETHEUS_UWSGI",
+            autospec=True,
+            new=False,
```
nemacysts, line 773:
```suggestion
```

nemacysts (approved):

tests/test_kubernetes_tools.py
EvanKrall: ? These lines are important to the test. It's what's setting DEFAULT_USE_PROMETHEUS_UWSGI to True and False.

EvanKrall: (oh but I did forget to set autospec=False after the pre-commit complained about me not having an autospec line)

nemacysts: ah, I didn't read the error message better than the tests and suggested removing the wrong line :p

nemacysts (approved):

PR #3090, opened 2021-04-01
Don't fail if one of replication info source isn't available

vkhromov:
Request replication information from all available sources and fail only
if all of them are unavailable.


Rob-Johnson (approved):

paasta_tools/smartstack_tools.py
```diff
@@ -589,10 +589,18 @@ def get_replication_for_instance(
...
-                )
+                # Try to get information from all available hosts in the pool before giving up
+                hostnames = self.get_hostnames_in_pool(hosts, instance_pool)
+                for hostname in hostnames:
+                    try:
+                        replication_info[location] = self._get_replication_info(
+                            location, hostname, instance_config, provider
+                        )
+                        break
+                    except ConnectionError:
```
Rob-Johnson, line 600: maybe worth logging something here to mention that the connection has failed?

PR #3091, opened 2021-04-02
Add cpus_limit to paasta allocation

matfra:
Currently, cpu recommendations are not super accurate because we don't know how much CPU a task can really use before getting throttled ([cpus + cpu_burst_add](https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes_tools.py#L987) AKA cpus limit in k8s). We currently only keep track of the requested cpus.
We could get that info out of band in the hourly report we are using for autotuned_defaults by parsing the configuration directly but we would lose time accuracy. I am OK to do that if you have strong opinion against this change.

This PR adds cpus_limit to the allocation log which will enable us to track this accurately and ultimately make better recommendations.

I know this file is in contrib but I felt like it deserved some unit tests, so I added a test directory and one puny test.
A couple of things to note:
- I did not bother doing this for Mesos, should I ?
- I could have done the reverse calculation and logged the cpu_burst_add but I felt like maybe it's just better to log the actual limit

### DIFF
--- a/paasta_tools/contrib/get_running_task_allocation.py
+++ b/paasta_tools/contrib/get_running_task_allocation.py
@@ -126,14 +126,18 @@ def get_kubernetes_resource_request(
 ) -> Mapping[str, float]:
     if not resources:
         requests: Mapping[str, str] = {}
+        limits: Mapping[str, str] = {}
     else:
         requests = resources.requests or {}
+        limits = resources.limits or {}
 
-    parsed = kubernetes_tools.parse_container_resources(requests)
+    parsed_requests = kubernetes_tools.parse_container_resources(requests)
+    parsed_limits = kubernetes_tools.parse_container_resources(limits)
     return {
-        "cpus": parsed.cpus,
-        "mem": parsed.mem,
-        "disk": parsed.disk,
+        "cpus": parsed_requests.cpus,
+        "mem": parsed_requests.mem,
+        "disk": parsed_requests.disk,
+        "cpus_limit": parsed_limits.cpus,
     }
 
 
--- a/paasta_tools/contrib/get_running_task_allocation.py
+++ b/paasta_tools/contrib/get_running_task_allocation.py
@@ -2,8 +2,8 @@
 import argparse
 import time
 from typing import Any
-from typing import Iterable
 from typing import Dict
+from typing import Iterable
 from typing import Mapping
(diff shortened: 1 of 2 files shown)


matthewbentley (approved):
I think skipping mesos is fine

paasta_tools/contrib/get_running_task_allocation.py
```diff
@@ -122,14 +122,18 @@ def get_kubernetes_resource_request(
 ) -> Mapping[str, float]:
     if not resources:
         requests: Mapping[str, str] = {}
+        limits: Mapping[str, str] = {}
```
nemacysts, line 125: super-duper-nit/feel free to ignore me: iirc, Python convention is to use more concrete types (e.g., `Dict`) for return value types and leave more abstract types/interfaces (e.g., `Mapping`) for inputs

for example, imagine a function that returns a `collections.Counter`: telling the type-checker that we're returning a Mapping means that it can't help us later on if we actually want to use some of the convenience methods that are added onto a Counter since, as far as mypy knows, we've returned something that only supports basic dictionary functionality

we've been kinda inconsistent about this tho, so feel free to ignore me :)

qui (approved):
Looks good to me!

While you're at it, can you move these tests into your new file? https://github.com/Yelp/paasta/blob/ae45ddf16be8592723a87b8ce14c60e150c723ce/tests/test_kubernetes_tools.py#L66

qui:
I agree that actual limit is better, and that we don't need to do this for Mesos. For Mesos, if we wanted it, we would need to use some docker stats from fullerite - I don't think the Mesos task info would include it.

ningweiii:
Addressed comments on the PR and the tests are all passed now! This is ready for a second review. Would you mind looking at it when you are available? @stug

EvanKrall (approved):
Currently the get_kubernetes_resource_request function takes resources and returns a transformed version of resources.requests, which is pretty straightforward, but this makes it also return data from resources.limits, which is somewhat surprising.
I'd either adjust the function name or split it into two functions, and call the second one here https://github.com/Yelp/paasta/pull/3091/files#diff-b8fdff99352d449c3171425e2c2cdd4d41e60663155ffd7c729421b734a07f98R207

(but this is minor, feel free to ignore.)

PR #3092, opened 2021-04-05
New mesh status command

kawaiwanyelp:
## Description
This PR adds a new API route to get the mesh status for an instance, and adds a new CLI command for the route, `paasta mesh-status`. If the instance is configured for the mesh, the output is the same as the mesh info in `paasta status`, and has the same rules for verbosity. Otherwise, the API will return informative error output if the instance is not an instance type that can run on k8s (so, marathon instances are not supported for simplicity), or if the instance does not have a mesh namespace (i.e. it isn't meant to be discoverable). 

## Testing
`make test`, with new tests added
`make dev-api` locally to emulate norcal-devc, then I ran `paasta mesh-status` against the API:

```
$ paasta mesh-status -s video -c norcal-devc -v
service: video
cluster: norcal-devc
  instance: main
    Envoy:
      Service Name: video.main
      Backends:
        uswest1adevc - Healthy - in BackendType.ENVOY with (3/1) total backends UP in this namespace.
          Proxied through Casper: True
          Hostname:Port      Weight  Status
          XX.XX.XX.XX:8888   10      HEALTHY
          XX.XX.XX.XX:8888   10      HEALTHY
          XX.XX.XX.XX:8888   10      HEALTHY
        uswest1cdevc - Healthy - in BackendType.ENVOY with (3/1) total backends UP in this namespace.
          Proxied through Casper: True
          Hostname:Port     Weight  Status
          XX.XX.XX.XX:8888  10      HEALTHY
          XX.XX.XX.XX:8888  10      HEALTHY
          XX.XX.XX.XX:8888  10      HEALTHY
  instance: submit_for_processing_worker
    ERROR: Instance 'video.submit_for_processing_worker' is not configured for the mesh
  instance: activate_processed_video_worker
    ERROR: Instance 'video.activate_processed_video_worker' is not configured for the mesh
  instance: purge_casper_worker
    ERROR: Instance 'video.purge_casper_worker' is not configured for the mesh
  instance: video_export_worker
    ERROR: Instance 'video.video_export_worker' is not configured for the mesh
  instance: retry_video_processing.run
    ERROR: Getting mesh status for tron instances is not supported
```


paasta_tools/api/views/instance.py
```diff
@@ -845,3 +845,64 @@ def get_deployment_version(
...
+def instance_mesh(request):
+    service = request.swagger_data.get("service")
+    instance = request.swagger_data.get("instance")
+    verbose = request.swagger_data.get("verbose") or 0
+    include_smartstack = request.swagger_data.get("include_smartstack")
+    if include_smartstack is None:
+        include_smartstack = True
+    include_envoy = request.swagger_data.get("include_envoy")
+    if include_envoy is None:
+        include_envoy = True
```
nemacysts, line 862: should we use `request.swagger_data.get(..., True)` here?

paasta_tools/api/views/instance.py
```diff
@@ -845,3 +845,64 @@ def get_deployment_version(
...
     return actual_deployments[key][:8] if key in actual_deployments else None
+
+
+@view_config(
+    route_name="service.instance.mesh_status", request_method="GET", renderer="json",
+)
+def instance_mesh(request):
+    service = request.swagger_data.get("service")
+    instance = request.swagger_data.get("instance")
+    verbose = request.swagger_data.get("verbose") or 0
```
nemacysts, line 856: should we use `request.swagger_data.get("verbose", 0)` here?

paasta_tools/api/views/instance.py
```diff
@@ -845,3 +845,64 @@ def get_deployment_version(
...
+    include_smartstack = request.swagger_data.get("include_smartstack")
+    if include_smartstack is None:
+        include_smartstack = True
+    include_envoy = request.swagger_data.get("include_envoy")
+    if include_envoy is None:
+        include_envoy = True
+
+    instance_mesh: Dict[str, Any] = {}
```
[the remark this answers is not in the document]
nemacysts: alternatively, should we set the default values in the swagger spec?

paasta_tools/api/views/instance.py
```diff
@@ -845,3 +845,64 @@ def get_deployment_version(
...
+    instance_mesh["service"] = service
+    instance_mesh["instance"] = instance
+
+    try:
+        instance_type = validate_service_instance(
+            service, instance, settings.cluster, settings.soa_dir
+        )
+    except NoConfigurationForServiceError:
+        error_message = (
+            "Deployment key %s not found.  Try to execute the corresponding pipeline if it's a fresh instance"
```
nemacysts, line 874: nit: should we use an f-string for consistency?

paasta_tools/cli/cmds/mesh_status.py
```diff
@@ -0,0 +1,172 @@
...
+    return 0, output
+
+
+def report_mesh_status_for_cluster(
+    cluster: str,
+    service: str,
+    instances: List[str],
+    system_paasta_config: SystemPaastaConfig,
+    verbose: int = 0,
+) -> Tuple[int, Sequence[str]]:
```
nemacysts, line 124: feel free to ignore: imo, it's a tiny bit better to use concrete types for return values and only use abstract types for function parameters (e.g., robustness principle: be robust in what you accept and conservative in what you send)

paasta_tools/cli/cmds/mesh_status.py
```diff
@@ -0,0 +1,172 @@
...
+                        service=service,
+                        instances=list(instances.keys()),
+                        system_paasta_config=system_paasta_config,
+                        verbose=args.verbose,
+                    ),
+                )
+            )
+
+    with concurrent.futures.ThreadPoolExecutor(max_workers=20) as executor:
+        tasks = [executor.submit(t[0], **t[1]) for t in tasks]  # type: ignore
```
nemacysts, line 166:
```suggestion
        tasks = [executor.submit(func, **params) for func, params in tasks]  # type: ignore
```
might be a little more self-documenting, but feel free to ignore

nemacysts (commented):
feel free to ignore/defer my comments - they're mostly just small nits

paasta_tools/api/views/instance.py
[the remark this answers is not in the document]
kawaiwanyelp: ive added default values

paasta_tools/api/views/instance.py
kawaiwanyelp: added default values to api specs

paasta_tools/api/views/instance.py
kawaiwanyelp: added default values to api specs

paasta_tools/api/views/instance.py
```diff
@@ -845,3 +845,60 @@ def get_deployment_version(
...
+    instance_mesh["service"] = service
+    instance_mesh["instance"] = instance
+
+    try:
+        instance_type = validate_service_instance(
+            service, instance, settings.cluster, settings.soa_dir
+        )
+    except NoConfigurationForServiceError:
+        error_message = (
+            f"Deployment key {settings.cluster}.{instance} not found. "
```
qui, line 870: Is this the right error message? I think the above function validates if the instance name exists in soaconfigs.

paasta_tools/api/views/instance.py
```diff
@@ -845,3 +845,60 @@ def get_deployment_version(
...
+            include_envoy=include_envoy,
+        )
+    except Exception:
+        error_message = traceback.format_exc()
+        raise ApiFailure(error_message, 500)
+
+    if mesh_status:
+        instance_mesh.update(mesh_status)
+    else:
+        raise ApiFailure(
```
qui, line 901: kubernetes_mesh_status returns {} in a few different cases, I think we need to differentiate between them.

paasta_tools/cli/cmds/mesh_status.py
```diff
@@ -0,0 +1,172 @@
...
+        "--verbose",
+        action="count",
+        dest="verbose",
+        default=0,
+        help=(
+            "Prints the health of individual backends in the mesh, instead of "
+            "the overall health"
+        ),
+    )
+    mesh_status_parser.add_argument(
```
qui, line 52: Is this used?

paasta_tools/instance/kubernetes.py
```diff
@@ -867,3 +867,66 @@ def ready_replicas_from_replicaset(replicaset: V1ReplicaSet) -> int:
...
+    )
+
+    kmesh: Dict[str, Any] = {}
+    mesh_status_kwargs = dict(
+        service=service,
+        instance=job_config.get_nerve_namespace(),
+        job_config=job_config,
+        service_namespace_config=service_namespace_config,
+        pods=pod_list,
+        should_return_individual_backends=verbose > 0,
```
stug, line 920: I kind of vote that we make this always True -- since the idea is that it'd only be run when we need more info on a service instance's backends, seems like we will probably almost always want the full table.

stug (approved):
Roughly looks good to me, I just think we should probably output the whole table always (i.e. always do verbose mode)

paasta_tools/cli/cmds/mesh_status.py
```diff
@@ -0,0 +1,172 @@
...
+
+    return 1 if any(return_codes) else 0, output
+
+
+def paasta_mesh_status(args) -> int:
+    system_paasta_config = load_system_paasta_config()
+
+    return_codes = [0]
+    tasks = []
+    clusters_services_instances = apply_args_filters(args)
```
qui, line 149: Do you think it's worth the extra code to support multiple clusters and instances? Or could we require a cluster and instance, so we only need `paasta_mesh_status_on_api_endpoint`?

Imagining how this might be used, for debugging, I think users will likely only need a single service/instance.

paasta_tools/api/views/instance.py
kawaiwanyelp: It's the same error message for all the other times we validate the instance name in other routes.

paasta_tools/api/views/instance.py
```diff
@@ -845,3 +845,64 @@ def get_deployment_version(
 ) -> Optional[str]:
     key = ".".join((cluster, instance))
     return actual_deployments[key][:8] if key in actual_deployments else None
+
+
+@view_config(
+    route_name="service.instance.mesh_status", request_method="GET", renderer="json",
+)
+def instance_mesh(request):
```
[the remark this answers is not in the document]
kawaiwanyelp: done

paasta_tools/api/views/instance.py
```diff
@@ -845,3 +845,64 @@ def get_deployment_version(
...
+    service = request.swagger_data.get("service")
+    instance = request.swagger_data.get("instance")
+    verbose = request.swagger_data.get("verbose") or 0
+    include_smartstack = request.swagger_data.get("include_smartstack")
+    if include_smartstack is None:
+        include_smartstack = True
```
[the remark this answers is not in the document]
kawaiwanyelp: i added defaults in the api specs

paasta_tools/instance/kubernetes.py
kawaiwanyelp: i removed the verbose arg and made this always true

paasta_tools/cli/cmds/mesh_status.py
kawaiwanyelp: It is used implicitly. `apply_args_filters` requires it. I've added a comment to clarify.

paasta_tools/cli/cmds/mesh_status.py
kawaiwanyelp: In this case I think it's fine since the code to support multi-cluster/instance output is less than the code required to validate the service/instance/cluster. I guess it depends on the principle. I don't mind making the switch if you want it.

qui: OK, I just noticed we were duplicating a bunch of code from the status command to do this, and was wondering if we really needed to.

For validating a service/instance/cluster, I thought we had done something similar recently with paasta logs? Might be worth making that into another util function.

Anyway, I don't feel strongly about it - up to you about what you think would be easiest to maintain.

paasta_tools/api/views/instance.py
qui: Yeah, but I think it may have been incorrect that time, too... Can you double-check that the message describes the error correctly?

qui: I'm mostly concerned about making sure that the tip leads down the right path.

kawaiwanyelp: ive changed it to "No instance {service}.{instance} is configured to run in the {cluster} cluster", which should make more sense.

paasta_tools/cli/cmds/mesh_status.py
kawaiwanyelp: I changed it to be more like paasta logs.

qui (approved):

PR #3094, opened 2021-04-06
Set default for use_signalfx to False. PAASTA-17097.

EvanKrall:
I'll plan to ship this on Thursday, after I've shipped the change to disable use_signalfx for most services.


nemacysts (approved):
looks like we might need to fix a test?

PR #3095, opened 2021-04-08
Remove SignalFX-related HPA code. PAASTA-17097

EvanKrall:
I'll plan to ship this in a week or so, once all the services have been off signalfx for a while.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -98,7 +98,6 @@ from kubernetes.client import V1TCPSocketAction
 from kubernetes.client import V1Volume
 from kubernetes.client import V1VolumeMount
 from kubernetes.client import V2beta2CrossVersionObjectReference
-from kubernetes.client import V2beta2ExternalMetricSource
 from kubernetes.client import V2beta2HorizontalPodAutoscaler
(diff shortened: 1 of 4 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -621,44 +616,6 @@ def get_autoscaling_metric_spec(
                         ),
                     )
                 )
-
-            use_signalfx = autoscaling_params.get(
```
nemacysts, line 625: should we also clean up the use_prometheus (for uwsgi) bits above this?

nemacysts: if we want, we can probably also clean up the metric names that we use for the HPAs since we no longer need to suffix things with `-prom` for uniqueness - we'd just need to update the prometheus adapter config script

that said, we should also update the prometheus adapter config script if we decide to remove use_prometheus from uwsgi codepaths

nemacysts (approved):

paasta_tools/kubernetes_tools.py
EvanKrall: yeah, getting rid of `use_prometheus` support is probably a good call.

PR #3096, opened 2021-04-09
MONK-2327 Add pod management policy support to StatefuleSets

zhaoyanh1202:
### Summary
Adding `pod_management_policy` to StatefulSets spec to make it configurable, right now without this option we are using the default [OrderedReady](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#orderedready-pod-management) option, it would be helpful for monk to adopt the [Parallel](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#parallel-pod-management) policy when scaling up during an incident.

The config required in yelpsoa-config would be something like
```
pod_management_policy: Parallel
```

### Testing
`make test` passed

### Note
I reckon rolling out this change should not lead to sts restart existing pods, as the default config is `OrderedReady` but please let me know if there is anything I have missed or if there is any other concern about this change.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -236,6 +236,14 @@ For more information on selector operators, see the official Kubernetes
 documentation on `node affinities
 <https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity>`_.
 
+  * ``pod_management_policy``: An option for applications managed with `StatefulSets <https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/>`_ to determine if the pods are managed in parallel or in order.
+
+    The default value is `OrderedReady <https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#orderedready-pod-management>`_.
+    It can be set to `Parallel <https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#parallel-pod-management>`_. For example::
+
+      pod_management_policy: Parallel
+
+
 .. _mesos-placement-options:
 
 Mesos
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -649,6 +649,12 @@
                             ]
                         }
                     }
+                },
+                "pod_management_policy": {
+                    "enum": [
+                        "OrderedReady",
+                        "Parallel"
+                    ]
                 }
             }
         }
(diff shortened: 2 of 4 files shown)


geokous (approved):
lgtm!

paasta_tools/kubernetes_tools.py
```diff
@@ -1370,6 +1370,10 @@ def get_enable_envoy_readiness_check(
             "check_envoy", system_paasta_config.get_enable_envoy_readiness_check()
         )
 
+    def get_pod_management_policy(self) -> str:
```
rohangulati, line 1373: Do you think we can add the pod management policy variable to `KubernetesDeploymentConfigDict` [here](https://github.com/Yelp/paasta/blob/bdb878a1bee61e19193f85762374a81b9f6dede4/paasta_tools/kubernetes_tools.py#L286)

rohangulati:
LGTM. I will leave it to @Rob-Johnson for the final review

paasta_tools/kubernetes_tools.py
zhaoyanh1202: Thanks for pointing it out! Added.

docs/source/yelpsoa_configs.rst
```diff
@@ -236,6 +236,14 @@ For more information on selector operators, see the official Kubernetes
 documentation on `node affinities
 <https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#node-affinity>`_.
 
+  * ``pod_management_policy``: An option for applications managed with `StatefulSets <https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/>`_ to determine if the pods are managed in parallel or in order.
```
Rob-Johnson, line 239: you'll also need to update the schema for yelpsoa-configs here https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/schemas/kubernetes_schema.json

Rob-Johnson (changes requested):

rohangulati (approved):

docs/source/yelpsoa_configs.rst
zhaoyanh1202: thanks, just added 😄

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -649,6 +649,9 @@
                             ]
                         }
                     }
+                },
+                "pod_management_policy": {
+                    "type": "string"
```
Rob-Johnson, line 654: you probably want this as an enum to limit the values people can use

```
 "enum": [
     "Parallel",
     "OrderedReady",
]
```

Rob-Johnson (approved):

PR #3099, opened 2021-04-13
Paasta 17290 mfd endpoint

stug:
Because the v2 kubernetes status changes fields around, I'm making a `bounce_status` endpoint specifically for use by mark-for-deployment.  This endpoint has _just_ the fields that MFD looks at, so it should be pretty close to a drop-in replacement.  The one difference is that previously the status command returned the status for all instance types and ignored them for everything but kubernetes and marathon.  This new method just returns data for kubernetes instances and returns a 404 for non-kubernetes, which will require a small tweak to MFD.

Also I changed a method that was marked as async but didn't actually make any eternal calls.


paasta_tools/instance/kubernetes.py
```diff
@@ -424,6 +424,65 @@ def filter_actually_running_replicasets(
...
+    else:
+        replicasets = kubernetes_tools.replicasets_for_service_instance(
+            service=job_config.service,
+            instance=job_config.instance,
+            kube_client=kube_client,
+            namespace=job_config.get_kubernetes_namespace(),
+        )
+        version_objects = filter_actually_running_replicasets(replicasets)
+
+    active_shas = kubernetes_tools.get_active_shas_for_service([app, *version_objects],)
```
stug, line 480: Is it sufficient to look at just Deployment/Statefulset + Replicasets/ControllerRevisions?  I was hoping to avoid fetching pods just for this

paasta_tools/api/views/instance.py
```diff
@@ -815,6 +815,36 @@ def instance_delay(request):
...
+        error_message = (
+            "Deployment key %s not found.  Try to execute the corresponding pipeline if it's a fresh instance"
+            % ".".join([settings.cluster, instance])
+        )
+        raise ApiFailure(error_message, 404)
+    except Exception:
+        error_message = traceback.format_exc()
+        raise ApiFailure(error_message, 500)
+
+    if instance_type != "kubernetes":
```
kawaiwanyelp, line 838: I think this should be `not pik.can_handle(instance_type)` instead. This will eval to `False` for all instance types running on k8s, like flink, cassandracluster, etc. not just kubernetes, because a lot of operator instances do rely on m-f-d. It's just that when the bounce, the operator rolls the pods, not s-k-j.

paasta_tools/instance/kubernetes.py
```diff
@@ -424,6 +424,65 @@ def filter_actually_running_replicasets(
...
+        load_deployments=True,
+    )
+    expected_instance_count = job_config.get_instances()
+    status["expected_instance_count"] = expected_instance_count
+    desired_state = job_config.get_desired_state()
+    status["desired_state"] = desired_state
+
+    kube_client = settings.kubernetes_client
+    if kube_client is None:
+        return status
```
kawaiwanyelp, line 445: nit: Can we raise an exception (maybe `RuntimeError`?) here instead? This way, the caller gets the message that we can't access talk to k8s, instead of just getting an empty return value.

paasta_tools/instance/kubernetes.py
kawaiwanyelp: I think it is sufficient.

Getting individual pods is useful for figuring out the status of pods that are having problems, which users will do with paasta status to debug. I think it is okay to defer to users for this both to keep the m-f-d output lean (i.e. don't clutter it up with pod info), but also to maintain good performance because fetching pods can be very expensive. 

In any event, we don't need pods to figure out how many ready pods there are.

kawaiwanyelp (changes requested):
Just small changes needed

paasta_tools/instance/kubernetes.py
stug: I think the real way to do this is to control the HTTP response code -- let me see how easy that is

paasta_tools/api/views/instance.py
stug: I don't think they use `mark-for-deployment --watch` -- the current implementation only cares about kubernetes and marathon: https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/mark_for_deployment.py#L1198

paasta_tools/instance/kubernetes.py
stug: Ok I raised a `RuntimeError` and then added some logic to catch it and reraise as an `ApiFailure`

kawaiwanyelp (approved):
lgtm. feel free to change the 405 error code if something else fits better.

kawaiwanyelp (approved):
lgtm

PR #3100, opened 2021-04-13
Add release info to k8s task allocation

sergsyelp:
In order to have a better precision of data for autoscaling we need to be able to distinguish between different versions of the service. This change fetches {{config_sha}} and {{git_sha}} Docker labels and adds them to the
allocation report.

[An example of the updated script output](https://fluffy.yelpcorp.com/i/1dn48x5kNn30NHG6xQR0tlD60QHH6Frl.html)

### DIFF
--- a/paasta_tools/contrib/get_running_task_allocation.py
+++ b/paasta_tools/contrib/get_running_task_allocation.py
@@ -36,6 +36,8 @@ class TaskAllocationInfo(NamedTuple):
     pod_name: str
     pod_ip: str
     host_ip: str
+    git_sha: str
+    config_sha: str
     mesos_container_id: str  # Because Mesos task info does not have docker id
 
 
@@ -101,6 +103,8 @@ async def get_mesos_task_allocation_info() -> Iterable[TaskAllocationInfo]:
                 pod_name=None,
                 pod_ip=None,
                 host_ip=None,
+                git_sha=None,
+                config_sha=None,
                 mesos_container_id=mesos_container_id,
             )
         )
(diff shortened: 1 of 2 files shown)


paasta_tools/contrib/get_running_task_allocation.py
```diff
@@ -151,14 +155,18 @@ def get_kubernetes_metadata(
...
 ]:
     labels = pod.metadata.labels or {}
     pod_name = pod.metadata.name
     pod_ip = pod.status.pod_ip
     host_ip = pod.status.host_ip
     service = labels.get("paasta.yelp.com/service")
     instance = labels.get("paasta.yelp.com/instance")
-    return service, instance, pod_name, pod_ip, host_ip
+    git_sha = labels.get("paasta.yelp.com/git_sha")
+    config_sha = labels.get("paasta.yelp.com/config_sha")
```
ronin13, line 168: Quick note (and not sure if this takes care of that) but 'paasta.yelp.com/config_sha' is not set on all pods, specifically, it is not set on pods of cassandra_k8s.  (git_sha is always set though)


Would be good to check how config_sha is checked in rest of paasta code which takes care of optionality of this.

qui (approved):

paasta_tools/contrib/get_running_task_allocation.py
sergsyelp: @ronin13 I think we are ok with situation when git_sha and/or git_sha are not set. The idea behind this change is to increase granularity for task allocation data for services during restarts and we basically care about ones with git_sha set. If git_sha and/or git_sha labels do not exist we are just at a status quo.

matfra (approved):

PR #3104, opened 2021-04-15
Annotation values my be strings

davent:
Annotation values have to be strings. This patch has been tested on kube1-uswest1astagef:
```
paasta-tools:
  Installed: 0.98.00-yelp2
  Candidate: 0.98.00-yelp2
  Version table:
 *** 0.98.00-yelp2 100
        100 /var/lib/dpkg/status
        
davent@kube1-uswest1astagef:~$ sudo setup_kubernetes_job statsite.main
INFO:__main__:Updating statsite-main-619eca5e0c3e96c5a12299eff0ce7ab2e087ea52-configa4a38a09 because configs have changed.
INFO:__main__:Ensuring related API objects for statsite-main-619eca5e0c3e96c5a12299eff0ce7ab2e087ea52-configa4a38a09 are in sync
INFO:root:poddisruptionbudget statsite-main up to date
INFO:paasta_tools.kubernetes.application.controller_wrappers:No HPA required for statsite-main/name in paasta
```

Result applied as expected:
```
metadata:
  annotations:
    iam.amazonaws.com/role: ""
    paasta.yelp.com/prometheus_path: /metrics
    paasta.yelp.com/prometheus_port: "9102"
 ```

PR #3105, opened 2021-04-15
Use str instead of Any for strings

vkhromov:


nemacysts (approved):

TheNavigat:
@vkhromov Can you add more context, please? Why are we doing this?

vkhromov:
> @vkhromov Can you add more context, please? Why are we doing this?

To prevent bugs like https://github.com/Yelp/paasta/pull/3104

vkhromov:
Let's use https://github.com/Yelp/paasta/pull/3106 instead of this one.

PR #3106, opened 2021-04-15
Use TypedDict for podspec labels and annotations

vkhromov:


paasta_tools/kubernetes_tools.py
```diff
@@ -283,6 +283,36 @@ def _set_disrupted_pods(self: Any, disrupted_pods: Mapping[str, datetime]) -> No
 )
 
 
+KubePodAnnotations = TypedDict(
```
nemacysts, line 286: I was about to comment that we should use the class syntax, but then I realized that you probably tried that/realized more quickly than I did that some of our keys aren't valid Python identifiers :p

nemacysts (approved):

PR #3107, opened 2021-04-16
Update autoscaling docs to reflect k8s/HPA

EvanKrall:


docs/source/autoscaling.rst
```diff
@@ -2,55 +2,40 @@
...
 ======================
 
 Enabling autoscaling
 --------------------
 
-In order to use autoscaling, edit your ``marathon-*.yaml`` files in your soa
+In order to use autoscaling, edit your ``kubernetes-*.yaml`` files in your soa
 configs and add a ``min_instances`` and a ``max_instances`` attribute and
-remove the ``instances`` attribute from each instance you want to autoscale.
+(optionally) remove the ``instances`` attribute from each instance you want to autoscale.
```
nemacysts, line 16: is this actually optional these days? I thought that `instances` took precedence and would prevent autoscaling

nemacysts: ah, I see - we ignore `instances` if `max_instances` is set! this seems kinda confusing from a user perspective, maybe we should make the soaconfigs validator prohibit having instances + max_instances at some point so that its always obvious what will happen?

docs/source/autoscaling.rst
```diff
@@ -120,40 +96,20 @@ The currently available decicion policies are:
   :offset:
     Float between 0.0 and 1.0, representing expected baseline load for each container.
     Defaults to 0.0.
-  :forecast_policy:
-    See "Forecast policies" below.
   :good_enough_window:
```
nemacysts, line 99: should we mention that this isn't actually supported yet?
```suggestion
  :good_enough_window (not supported in Kubernetes):
```
or just go ahead and delete this section for now until we do?

EvanKrall: good catch, I'll reword this.

docs/source/autoscaling.rst
EvanKrall: Yeah, it's weird. It's been like this since we first introduced min_instances/max_instances. We needed to resolve the ambiguity one way or the other because the way we use yaml anchors doesn't make it easy to delete keys.

I've proposed fixing this by making `instances` into a Union of int and a TypedDict with `min` and `max` keys.

nemacysts (approved):

PR #3112, opened 2021-05-06
RFC: Optionally mounting in service specific keys to /etc/boto_cfg

Qmando:
This is mostly experimental at this point. What I'm hoping to achieve is that services can specify which keys they will need to use, and then those keys will be sourced out of a folder, put into a new folder, and have that mounted in as /etc/boto_cfg. This goes along with https://github.yelpcorp.com/sysgit/puppet/pull/3370/files.

We first expect /etc/boto_cfg_private to exist and have keys inside it (see other PR)

Then you add an entry into the service configuration for which boto_keys you will need

When the service is launched, if the boto_keys setting is found, it will attempt to create a new folder within boto_cfg_private for the service, link in the source keys that match the entries, and then add that as another volume.

I updated the volume deduplication code to dedupe based on containerPath only, such that the extra volume added above will trump the default /etc/boto_cfg one. (Note: I don't think it's valid to have two mounts with the same containerPath anyway, this should probably get shipped regardless)

Thus, the service will see /etc/boto_cfg as only the subset of keys it needs. If no boto_keys is specified, no extra volume is added, and the default boto_cfg mount will remain as is.

When the keys get updated, similar code in puppet (see linked review) will update the keys to their new values, and consequently the new keys will appear to the service. The reason this code exists in two places is to prevent race conditions (see linked review discussion).

This isn't super polished at this point, I'm mostly seeking feedback first. There's still some open questions as to whether we want to seriously pursue this. Any and all feedback is welcome.

To test this, I ran a paasta local-run with custom configs:
https://fluffy.yelpcorp.com/i/DPgHNlmPczdb7q1Gzpj5QJqwNR3WcQb4.html

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -98,6 +98,7 @@ DEPLOY_PIPELINE_NON_DEPLOY_STEPS = (
     "performance-check",
     "push-to-registry",
 )
+BOTO_SOURCE_DIR = "/etc/boto_cfg_private/"
 # Default values for _log
 ANY_CLUSTER = "N/A"
 ANY_INSTANCE = "N/A"
@@ -853,6 +854,33 @@ class InstanceConfig:
         <https://mesosphere.github.io/marathon/docs/native-docker.html>`_"""
         return self.config_dict.get("extra_volumes", [])
 
+    def get_boto_mount(self):
+        try:
+            instance_config = load_service_instance_config(
+                self.service,
+                self.instance,
+                self.get_instance_type(),
+                self.cluster,
+                self.soa_dir,
+            )
+        except paasta_tools.utils.NoConfigurationForServiceError:
+            return []
+        boto_keys = instance_config.get("boto_keys")
+        if not boto_keys:
+            return []
+        service_name = self.service + "." + self.instance
+        create_service_boto_folder(service_name, boto_keys)
+        boto_mount_path = BOTO_SOURCE_DIR + service_name
+        if os.path.exists(boto_mount_path):
+            return [
+                {
+                    "containerPath": "/etc/boto_cfg",
+                    "hostPath": boto_mount_path,
+                    "mode": "RO",
+                }
+            ]
+        return []
+
     def get_aws_ebs_volumes(self) -> List[AwsEbsVolume]:
         return self.config_dict.get("aws_ebs_volumes", [])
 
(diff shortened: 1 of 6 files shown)


paasta_tools/utils.py
```diff
@@ -853,6 +854,33 @@ def get_extra_volumes(self) -> List[DockerVolume]:
...
+                self.get_instance_type(),
+                self.cluster,
+                self.soa_dir,
+            )
+        except paasta_tools.utils.NoConfigurationForServiceError:
+            return []
+        boto_keys = instance_config.get("boto_keys")
+        if not boto_keys:
+            return []
+        service_name = self.service + "." + self.instance
```
stug, line 871: I think you can just do `self.job_id`

paasta_tools/utils.py
```diff
@@ -3090,6 +3122,23 @@ def get_running_mesos_docker_containers() -> List[Dict]:
     ]
 
 
+def create_service_boto_folder(service, keys):
+    service_folder = os.path.join(BOTO_SOURCE_DIR, service)
+    if not os.path.exists(service_folder):
```
stug, line 3127: mega-nit: can we call this a `directory` instead of a folder?

paasta_tools/utils.py
```diff
@@ -853,6 +854,33 @@ def get_extra_volumes(self) -> List[DockerVolume]:
...
+                self.soa_dir,
+            )
+        except paasta_tools.utils.NoConfigurationForServiceError:
+            return []
+        boto_keys = instance_config.get("boto_keys")
+        if not boto_keys:
+            return []
+        service_name = self.service + "." + self.instance
+        create_service_boto_folder(service_name, boto_keys)
+        boto_mount_path = BOTO_SOURCE_DIR + service_name
```
stug, line 873: Instead of having both this method and `create_service_boto_folder` build the `boto_mount_path`, it'd be nicer to just have `create_service_boto_folder` return the path to avoid the possibility that we change this in one place and forget to change it in the other

paasta_tools/utils.py
```diff
@@ -3090,6 +3122,23 @@ def get_running_mesos_docker_containers() -> List[Dict]:
...
+        for filename in os.listdir(BOTO_SOURCE_DIR):
+            if filename.split(".")[0] in keys:
+                os.link(
+                    os.path.join(BOTO_SOURCE_DIR, filename),
+                    os.path.join(BOTO_SOURCE_DIR, service, filename),
+                )
+    except OSError:
+        # Either BOTO_SOURCE_DIR is absent, or the keys inside already exist
+        # In either case, doing nothing is OK
+        pass
```
stug, line 3139: just to be clear, if there are multiple instances of a service running on a box, they will all share the same directory, yeah, meaning they'll all have access to each others' keys, yeah?

stug (commented):
I think this approach roughly makes sense to me, but I'd maybe be curious to hear what @EvanKrall or folks in London think, too.  I guess there is a question of how comfortable we are with assuming the existence of something like `boto_cfg_private` which is specific to yelp, but I am not too concerned.

Also I couldn't help myself and left you some nits I picked.

PR #3114, opened 2021-05-13
Stop using bintray

vkhromov:


.github/workflows/ci.yml
```diff
@@ -51,20 +51,3 @@ jobs:
         id: get_package_name
       - run: "echo -n ::set-output name=version::; dpkg-parsechangelog -l./debian/changelog | sed -n 's/^Version: //p'"
         id: get_version
-      - name: Upload deb package to Bintray
```
nemacysts, line 54: should we also remove the get_package_name and get_version steps? they were only used for uploading to bintray

nemacysts (approved):

.github/workflows/ci.yml
vkhromov: Let's better remove the whole `build_deps`

nemacysts: my only concern with that is that we won't know if we've broken building dpkgs externally (and in a way, internally since if `make itest_%` fails on GH Actions it could signal that it'll also fail in our jenkins environment) by removing building debs from our OSS testing

vkhromov: Oh, right, we don't have separate `itest_*` steps anymore...

nemacysts (approved):

PR #3118, opened 2021-05-21
Add autoscaling configuration validation

nemacysts:
Let's ensure that people don't try to set a (setpoint, offset)
pair that is invalid.

NOTE: I've deleted the existing autoscaling validation code as it was
only checking for an autoscaling configuration that we moved away from
(`horizontal_autoscaling`) - see #2582 for the initial implementation of
this and #2979 for its removal.

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -381,37 +381,38 @@ def validate_autoscaling_configs(service_path):
 
     :param service_path: Path to directory containing soa conf yaml files for service
     """
-    path = os.path.join(service_path, "*.yaml")
+    soa_dir, service = path_to_soa_dir_service(service_path)
     returncode = True
-    instances = {}
-    # Read and store all instance configuration in instances dict
-    for file_name in glob(path):
-        if os.path.islink(file_name):
-            continue
-        basename = os.path.basename(file_name)
-        if basename.startswith("kubernetes"):
-            cluster = basename[basename.rfind("kuernetes-") + 1 :]
-            instances[cluster] = get_config_file_dict(file_name)
-    # Validate autoscaling configurations for all instances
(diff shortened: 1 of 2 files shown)


nemacysts:
pics of what this looks like: https://fluffy.yelpcorp.com/i/q6Pmlf6gS1HwtBxWrX2fXDhN5x0wT2x4.html with a diff of https://fluffy.yelpcorp.com/i/qDQs81Rpvh4ZCxTBhBrkBpHHBmbc9g4x.html

lanceleih (approved):
LGTM

jvperrin (approved):
This seems reasonable to me, although I can't say I know too much about the code or methods used. The output looks great though!

paasta_tools/cli/cmds/validate.py
```diff
@@ -381,37 +381,38 @@ def validate_autoscaling_configs(service_path):
...
+            )
+
+            if (
+                instance_config.get_instance_type() == "kubernetes"
+                and instance_config.is_autoscaling_enabled()
+            ):
+                autoscaling_params = instance_config.get_autoscaling_params()
+                if autoscaling_params["metrics_provider"] in {"uwsgi", "http"}:
+                    setpoint = autoscaling_params["setpoint"]
+                    offset = autoscaling_params["offset"]
```
stug, line 406: I don't think you can guarantee that the offset key will be there from what I can tell?

stug (approved):
Looks good except I think that you might get KeyErrors for services with no offset?

paasta_tools/cli/cmds/validate.py
nemacysts: ah whoops, I was misremembering a change I had backed out where I provided defaults in `AutoscalingParamsDict` - you are indeed correct tho

stug (approved):

PR #3121, opened 2021-06-01
Scaffold out the k8s toggles for tronjobs

nemacysts:
We're handling the toggles in PaaSTA rather than Tron since the
cluster-level config isn't easily available when setting up an
ActionRun, so we will instead use setup_tron_namespace to configure Tron
to use either Mesos or Kubernetes.

Additionally, we'll need to add some translation layers for things like
Mesos constraints/Docker parameters, so we might as well centralize that
here rather than setup Mesos here and Kubernetes in Tron.

The general idea here is that we want our config sync to handle toggling the 
use of k8s - disabling k8s for an entire cluster will lead to this sync setting
`use_k8s` to false for all jobs and therefore set the executor back to `mesos`
for everything

### DIFF
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -169,6 +169,9 @@
             ],
             "additionalProperties": false,
             "properties": {
+                "use_k8s": {
+                    "type": "boolean"
+                },
                 "name": {
                     "$ref": "#definitions/name"
                 },
--- a/paasta_tools/setup_tron_namespace.py
+++ b/paasta_tools/setup_tron_namespace.py
@@ -24,6 +24,8 @@ import argparse
 import logging
 import sys
 
+import ruamel.yaml as yaml
+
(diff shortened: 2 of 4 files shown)


nemacysts:
We'll want to merge this after rolling out https://github.com/Yelp/Tron/pull/802 

testing: I tested this by running the linked Tron PR + this PR on `tronplayground`

lanceleih (approved):
I'm not too familiar with this script, but I looked through the whole PR and the logic looks correct to me.

jfongatyelp (approved):

jfongatyelp (approved):
LGTM, though we'll need to add the `use_k8s` flag to the schema used for `paasta validate` to allow this at the job level too.

jfongatyelp (approved):

nemacysts:
> LGTM, though we'll need to add the `use_k8s` flag to the schema used for `paasta validate` to allow this at the job level too.

done!

jfongatyelp (approved):

PR #3124, opened 2021-06-07
PAASTA-17337 - capturing instance misconfiguration exceptions for setup_kubernetes_job

cuza:
If one instance is misconfigured and hits an exception, it will cause the entire run to fail and will not progress to the other service instances that have yet to be setup.

### DIFF
--- a/paasta_tools/setup_kubernetes_job.py
+++ b/paasta_tools/setup_kubernetes_job.py
@@ -168,8 +168,7 @@ def setup_kube_deployments(
                 log.info(f"Ensuring related API objects for {app} are in sync")
                 app.update_related_api_objects(kube_client)
             except Exception:
-                log.error(f"Error while processing: {app}")
-                log.error(traceback.format_exc())
+                log.exception(f"Error while processing: {app}")
         if rate_limit > 0 and api_updates >= rate_limit:
             log.info(
                 f"Not doing any further updates as we reached the limit ({api_updates})"
--- a/paasta_tools/setup_kubernetes_job.py
+++ b/paasta_tools/setup_kubernetes_job.py
@@ -23,7 +23,6 @@ Command line options:
 import argparse
 import logging
 import sys
-import traceback
 from typing import Optional
 from typing import Sequence
 from typing import Tuple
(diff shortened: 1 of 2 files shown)


paasta_tools/setup_kubernetes_job.py
```diff
@@ -149,22 +150,26 @@ def setup_kube_deployments(
...
+                    app.update(kube_client)
+                    api_updates += 1
+                else:
+                    log.info(f"{app} is up-to-date!")
+
+                log.info(f"Ensuring related API objects for {app} are in sync")
+                app.update_related_api_objects(kube_client)
+            except Exception:
+                log.error(f"Error while processing: {app}")
+                log.error(traceback.format_exc())
```
nemacysts, line 172:
```suggestion
                log.exception(f"Error while processing: {app}")
```

will clean this up a bit and mean that we don't have to import/use `traceback`

cuza: Thanks for the suggestions

tests/test_setup_kubernetes_job.py
```diff
@@ -391,3 +391,31 @@ def test_setup_kube_deployments_rate_limit():
...
+    ), mock.patch(
+        "paasta_tools.setup_kubernetes_job.log", autospec=True
+    ) as mock_log_obj:
+        mock_client = mock.Mock()
+        mock_service_instances = ["fake.instance"]
+        fake_app = mock.Mock(create=mock.Mock())
+        fake_app.create = mock.Mock(side_effect=Exception("Kaboom!"))
+        fake_app.__str__ = mock.Mock(return_value="fake_app")
+        mock_create_application_object.return_value = (True, fake_app)
+
```
jfongatyelp, line 410: It looks like this test isn't actually verifying we'd 'skip' one with an exception and move onto the next since we're only testing 1 `mock_service_instances`. We *are* verifying we caught an exception, so that's great, but we should also do as the test is named and verify w/ multiple `mock_service_instances`, making sure we caught the exception on the first but can successfully move on to the 2nd. 

You should be able to configure multiple side effects for `mock_create_application_object`, making sure we actually called the 2nd app's `.create` after catching the exception on the first one.

jfongatyelp (approved):

PR #3126, opened 2021-06-08
Configure GHA to upload an sdist+wheel to public pypi

nemacysts:
nerve-tools uses paasta as a dependency, so without doing this we can't
use any new paasta features there.

we only stopped uploading sdists with the migration to GHA since we
wanted to focus on getting public CI back up and public PyPI uploads
could wait a bit.

lastly, I've also added back wheel uploads since it seems like we had
stopped doing that for a while in travis (and i'm assuming that wasn't
on purpose).

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -1,0 +1,46 @@
+---
+name: paasta-pypi
+
+on:
+  push:
+    tags:
+      - v*
+
+jobs:
+  tox:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        toxenv:
+          - py37-linux,docs,mypy,tests
+          - general_itests
+          - paasta_itests
+          - example_cluster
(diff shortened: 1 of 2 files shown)


.github/workflows/pypi.yml
```diff
@@ -0,0 +1,47 @@
...
+      - uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      # this will create a .tar.gz with all the code (i.e., an sdist) as well as
+      # a wheel (i.e., a "compiled" version of the code) in dist/
+      - run: python setup.py sdist bdist_wheel
+      # and finally, upload the above artifacts to public PyPI
+      - uses: pypa/gh-action-pypi-publish@v1.2.2
+        with:
+          password: ${{ secrets.pypi_password }}
```
danielpops, line 47: FYI this is added as a repo secret now

keymone (approved):
thanks!

PR #3130, opened 2021-06-10
Saving boto_cfg keys as secrets and mounting them into services

Qmando:
More info surrounding this is available in y/cep-1713

The idea is to config each service to allow a list of "boto keys", ie service user keys within boto_cfg folder. These values will be saved as a multi-value kubernetes secret for each instance. Those keys will then be mounted into the container to replace mounting /etc/boto_cfg.

To achieve this, I had to
- Add schema changes for `boto_keys` in service config
- Read these keys and create the multivalued secret in paasta_secret_sync
   - I needed to introduce two functions to create these, that don't use a secrets provider at all
- When the kubernetes app is created, the code will check for `boto_keys` and create a new volume mount at /etc/boto_cfg  with each file reproduced 

I've tested this in kubestage and verify that it works. The keys will also get updated without causing a service bounce when the files change.

I've also verified that if the secret doesn't exist in kubernetes, the pods will still launch as normal.

If you put in keys that don't exist, the secret sync will emit a warning, no k8s secret will be created, and the pods will work as normal.


paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -141,6 +145,74 @@ def sync_secrets(
     if not os.path.isdir(secret_dir):
         log.debug(f"No secrets dir for {service}")
         return True
+
+    # Update boto key secrets
+    service_dir = os.path.join(soa_dir, service)
+    kubernetes_config_file = f"kubernetes-{cluster}.yaml"
```
danielpops, line 151: nit: presumably there are abstractions elsewhere for getting a reference to the cluster yaml config file?

paasta_tools/kubernetes_tools.py
```diff
@@ -1732,7 +1769,7 @@ def sanitize_for_config_hash(
         calculation of config hash.
 
         :param config: complete_config hash to sanitize
```
danielpops, line 1771: nit: you "fixed" the "typo" in the line below, but missed this one :)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -141,6 +145,74 @@ def sync_secrets(
     if not os.path.isdir(secret_dir):
         log.debug(f"No secrets dir for {service}")
         return True
```
danielpops, line 147: This function will return early if you have no paasta secrets defined, which is not desirable

danielpops: in fact, maybe you should create a separate `sync_boto_secrets` function and call it from `sync_all_secrets`

paasta_tools/utils.py
```diff
@@ -3723,7 +3723,5 @@ def ldap_user_search(
 
 
 def _reorder_docker_volumes(volumes: List[DockerVolume]) -> List[DockerVolume]:
-    deduped = {
-        v["containerPath"].rstrip("/") + v["hostPath"].rstrip("/"): v for v in volumes
-    }.values()
+    deduped = {v["containerPath"].rstrip("/"): v for v in volumes}.values()
```
danielpops, line 3726: Can you comment on why this is safe to do and the expected behavior before vs after?

danielpops (commented):
Couple of minor nits, but this looks like it would do the trick. Deferring to paasta experts for the final say on design, entrypoint etc.

paasta_tools/kubernetes_tools.py
```diff
@@ -1183,16 +1183,40 @@ def get_pod_volumes(
...
+        )
+        volume = V1Volume(
+            name=f"secret-boto-key-{service_name}",
+            secret=V1SecretVolumeSource(
+                secret_name=f"paasta-boto-key-{service_name}",
+                default_mode=mode_to_int("0444"),
+                items=items,
+            ),
+        )
+        return volume
```
Qmando, line 1210: This should be an empty list if `boto_keys` was not found

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
mattmb: I think that extends to some of the below too. We have abstractions for reading the yelpsoa configs and properties within. https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes_tools.py#L444 is probably where you want to start.

mattmb (commented):
I'll leave others to do a more technical review but I have a couple of high level questions. Do we understand how rotation will work? When the key is rotated I think the code above will see the signature changing and update k8s, does that then cause a bounce of the service (I think paasta secrets mechanism is designed to do so because they are slow changing and env vars need a restart anyway). In theory I think k8s will update volume secrets but we probably don't want to bounce every service every n hours when the keys are rotated.

My other question is whether we are going to allow people to distribute new creds this way rather than via IAM rules (or do we just use this to improve current situation)? I'm happy for an approach like this in general but I'd like whatever we do to converge on Pod based IAM roles in the end.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -141,6 +145,74 @@ def sync_secrets(
...
+                continue
+            boto_keys = instance_config.get("boto_keys", [])
+            if not boto_keys:
+                continue
+            boto_keys.sort()
+            secret_data = {}
+            for key in boto_keys:
+                sanitised_key = key.replace(".", "-").replace("_", "--")
+                try:
+                    with open(f"/etc/boto_cfg/{key}") as f:
```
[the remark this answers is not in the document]
danielpops: Not yet, currently _all_ boto_cfgs go to every host. There is an in-progress project this quarter to only distribute keys where they actually need to be. We will keep that in mind to make sure the master has all the keys needed by any paasta service. Good call!

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -342,6 +342,12 @@
                         "type": "string"
                     }
                 },
+                "boto_keys": {
+                    "type": "array",
+                    "items": {
+                        "type": "string"
+                    }
+                },
```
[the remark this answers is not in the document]
danielpops: I'm really keen on the idea of changing "paasta local-run" to not run locally, and instead the workload gets launched on the kubernetes cluster, and then we automatically `kubectl exec` to give effectively the same experience for end-users. That would address this issue you're raising, as well as others where running locally is not a perfect representation of running on the cluster (such as assigning AWS Role to a pod via aws pod identity mechanism)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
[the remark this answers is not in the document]
jfongatyelp: I think there may still be 1 or 2 cases where some /etc/boto_cfg are deployed only on special pools, and therefore not the masters:

https://sourcegraph.yelpcorp.com/search?q=repo:sysgit/puppet+file:modules/profile_paasta.*+boto::distribute_users&patternType=regexp

paasta_tools/cli/schemas/kubernetes_schema.json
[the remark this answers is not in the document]
Qmando: For the time being, we'll basically be leaving keys deployed in dev if they are needed for paasta local-run. This is far from ideal, but hopefully dpopes' idea will obviate that.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -215,5 +232,84 @@ def sync_secrets(
...
+        cluster=cluster, instance_type_class=KubernetesDeploymentConfig
+    ):
+        instance = instance_config.instance
+        boto_keys = instance_config.config_dict.get("boto_keys", [])
+        if not boto_keys:
+            continue
+        boto_keys.sort()
+        secret_data = {}
+        for key in boto_keys:
+            for filetype in ["sh", "yaml", "json", "cfg"]:
```
nemacysts, line 255: should `["sh", "yaml", "json", "cfg"]` be a constant? we use it in two files in this PR

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -215,5 +232,84 @@ def sync_secrets(
...
+                this_key = key + "." + filetype
+                sanitised_key = this_key.replace(".", "-").replace("_", "--")
+                try:
+                    with open(f"/etc/boto_cfg_private/{this_key}") as f:
+                        secret_data[sanitised_key] = base64.b64encode(
```
nemacysts, line 260: this is just a k8s thing and the contents inside containers will automatically be decoded for us inside pods, right?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -215,5 +232,84 @@ def sync_secrets(
...
+                else:
+                    raise
+            create_kubernetes_secret_signature(
+                kube_client=kube_client,
+                secret=secret,
+                service=service,
+                secret_signature=signature,
+            )
+        elif signature != kubernetes_signature:
+            log.info(f"{secret} for {service} needs updating as signature changed")
```
nemacysts, line 296: this will cause a service bounce, right? 

do we rotate all keys at once or on a rolling basis? just worried about contant big bounces if this happens all at once :p

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -215,5 +232,84 @@ def sync_secrets(
...
+                            f.read().encode("utf-8")
+                        ).decode("utf-8")
+                except IOError:
+                    log.warning(
+                        f"Boto key {this_key} required for {service} could not be found."
+                    )
+        if not secret_data:
+            continue
+        secret = f"paasta-boto-key-{get_kubernetes_app_name(service, instance)}"
+        hashable_data = "".join([secret_data.get(key, "") for key in secret_data])
```
nemacysts, line 270: we should be able to use `secret_data[key]` here, right? I'm just trying to figure out in what cases we'd get to this line and potentially have a missing key in this dict

tests/test_kubernetes_tools.py
```diff
@@ -1216,6 +1216,36 @@ def test_get_volume_mounts(self):
...
+        )
+        with mock.patch.object(deployment, "get_boto_secret_hash", return_value="hash"):
+            volumes = deployment.get_boto_volume()
+        if config_dict:
+            assert (
+                volumes.secret.secret_name == "paasta-boto-key-my-service-my-instance"
+            )
+            assert len(volumes.secret.items) == len(config_dict["boto_keys"]) * 4
+            assert any(
+                [item.path == "scribereader.yaml" for item in volumes.secret.items]
```
nemacysts, line 1244: should this look at config_dict to figure out what to check for? technically we could break things such that we only use the path for the first item in the boto_keys list and this test would still pass :p

paasta_tools/kubernetes_tools.py
```diff
@@ -1308,6 +1339,35 @@ def get_volume_mounts(
                 for volume in secret_volumes
             ]
         )
+        if self.config_dict.get("boto_keys", []):
+            secret_hash = self.get_boto_secret_hash()
+            service_name = (
+                self.get_sanitised_service_name()
+                + "-"
+                + self.get_sanitised_instance_name()
+            )
```
nemacysts, line 1348: should this be 
```suggestion
            service_name = self.get_sanitised_deployment_name()
```
for consistency?

paasta_tools/kubernetes_tools.py
```diff
@@ -1308,6 +1339,35 @@ def get_volume_mounts(
...
+                        volume_mounts.remove(existing_mount)
+                        break
+                volume_mounts.append(mount)
+        return volume_mounts
+
+    def get_boto_secret_hash(self) -> str:
+        kube_client = KubeClient()
+        service_name = (
+            self.get_sanitised_service_name() + "-" + self.get_sanitised_instance_name()
+        )
```
nemacysts, line 1366: same here, should this be `self.get_sanitised_deployment_name()`?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -215,5 +232,84 @@ def sync_secrets(
     return True
 
 
+def sync_boto_secrets(
```
nemacysts, line 235: this should probably take a namespace param (just like `sync_secrets`) - otherwise we'll need to come back and add that in once we've added support for mounting secrets as files in task_proc/tron :)

paasta_tools/kubernetes_tools.py
```diff
@@ -2757,6 +2817,26 @@ def create_secret(
     )
 
 
+def create_plaintext_dict_secret(
+    kube_client: KubeClient, secret_name: str, secret_data: dict, service: str,
```
nemacysts, line 2821: this should also take a namespace param like `create_secret()`

paasta_tools/kubernetes_tools.py
```diff
@@ -2786,6 +2866,27 @@ def update_secret(
     )
 
 
+def update_plaintext_dict_secret(
+    kube_client: KubeClient, secret_name: str, secret_data: dict, service: str,
```
nemacysts, line 2870: and same here: this should also take a namespace param like `update_secret()`

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
Qmando: Yes! I can add some more concrete testing data to the PR to show this

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
Qmando: AFAIK it does not. I tested this by changing the value of a key while I had a shell in the service. Though, I exited the shell right after so there's a possibility it's still doing some delayed bounce.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
Qmando: Yeah, it's literally got `for key in secret_data` ...

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
Qmando: If I did miss this, are you thinking of a particular mechanism that causes a bounce?

nemacysts: I *think* that updating the secret content will cause a bounce once setup_kubernetes_job gets to the service in question - but I haven't confirmed that yet

paasta_tools/utils.py
danielpops: @Qmando still looking for your input here :)

danielpops (approved):

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
mattmb: https://github.com/Yelp/paasta/blob/c5f92d43a391fc1e12d51a6f3ef6fdbe29f3c30b/paasta_tools/kubernetes_tools.py#L1777 here's a code hint. Maybe we actually will ignore changes if it's just volume based?

mattmb: Generally speaking it would make sense if we just bounced for environment based secrets since I think file based secrets get live updated now in k8s? @Qmando have you thought much about the load when secrets rotate, do they happen all at once, are you satisfied that it's not a huge load on the k8s API? The other area where I'd suggest extra caution for you and reviewers is in the kubernetes_tools code since any case where we generate an invalid k8s object/error/exception could prevent deploying the service(s) and cause a DAR3

paasta_tools/utils.py
Qmando: I will add a variation of this as a comment in the code too:

The behavior before is to deduplicate based on containerPath+hostPath as we can clearly see. This means that if you have two mounts with the same container path, eg,`-v /etc/boto_cfg:/etc/boto_cfg -v /etc/foobar:/etc/boto_cfg`, it will end up keeping both of them and the container will fail to launch. This means that it's currently broken in preventing bad duplicate mounts.

The fix means that it now disregards hostPath when doing this, which is the correct behavior.

I don't remember exactly how this error manifests in paasta, but Docker complains to you with `Error response from daemon: Duplicate mount point: /etc/boto_cfg` if you were to do this.

Qmando: well actually I completely removed this change and did
```
                for existing_mount in volume_mounts:
                    if existing_mount.mount_path == "/etc/boto_cfg":
                        volume_mounts.remove(existing_mount)
                        break
```
instead, but I still think this change is more correct behavior.

danielpops (approved):

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
Qmando: > def sanitize_for_config_hash
Looks like this is explicitly looking at PAASTA_SECRET_* style secrets, and won't detect changes to the boto secret. But, if `boto_keys` in the config dict changes, it will bounce. This is the behavior we want.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -266,12 +268,17 @@ def sync_boto_secrets(
                     )
         if not secret_data:
             continue
+        # In order to prevent slamming the k8s API, add some artificial delay here
+        time.sleep(0.3)
         app_name = get_kubernetes_app_name(service, instance)
-        secret = f"paasta-boto-key-{app_name}"
-        hashable_data = "".join([secret_data.get(key, "") for key in secret_data])
+        secret = f"{namespace}-boto-key-{app_name}"
```
nemacysts, line 274: i think we actually wanna keep this as `paasta` since some namespaces can be extremely large (e.g., the nrtsearch namespace)

nemacysts (approved):

paasta_tools/kubernetes_tools.py
```diff
@@ -1265,16 +1266,47 @@ def get_pod_volumes(
                     ),
                 )
             )
+        boto_volume = self.get_boto_volume()
+        if boto_volume:
+            pod_volumes.append(boto_volume)
         return pod_volumes
 
+    def get_boto_volume(self):
```
nemacysts, line 1274:
```suggestion
    def get_boto_volume(self) -> Optional[V1Volume]:
```

nemacysts (approved):

PR #3133, opened 2021-06-23
Add registration labels to pods

TheNavigat:
Since rolling this out in one go will bounce all services, we need to roll it out carefully and slowly.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1570,6 +1570,12 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         if prometheus_shard:
             labels["paasta.yelp.com/prometheus_shard"] = prometheus_shard
 
+        if system_paasta_config.get_kubernetes_add_registration_labels():
+            # Allow Kubernetes Services to easily find
+            # pods belonging to a certain smartstack namespace
+            for registration in self.get_registrations():
+                labels[f"paasta.yelp.com/registrations/{registration}"] = "true"  # type: ignore
+
         # not all services use uwsgi autoscaling, so we label those that do in order to have
         # prometheus selectively discover/scrape them
         if self.should_run_uwsgi_exporter_sidecar(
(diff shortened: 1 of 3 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1570,6 +1570,11 @@ def get_pod_template_spec(
         if prometheus_shard:
             labels["paasta.yelp.com/prometheus_shard"] = prometheus_shard
 
+        # Allow Kubernetes Services to easily find
+        # pods belonging to a certain smartstack namespace
+        for registration in self.get_registrations():
+            labels[f"paasta.yelp.com/service/{registration}"] = "true" # type: ignore
```
Rob-Johnson, line 1576: I don't know whether we've thought about a hierarchy like this, but I'd be more inclined to continue using the `registrations` nomenclature (so `paasta.yelp.com/registrations/{registration}`).

@keymone can you weigh in here - I know you've thought about this a lot in the past

keymone: generally speaking we should keep label names as flat as possible, but in this case i think going one level deep is warranted. i do agree that `registrations` is better than `service`.

Rob-Johnson (approved):

Rob-Johnson:
actually, sorry @TheNavigat

TheNavigat:
Self-note: a good example of how this is done: https://github.com/Yelp/paasta/commit/ce29e367457a13303043d398cb50c0e7562185cd

Rob-Johnson (approved):

paasta_tools/kubernetes_tools.py
```diff
@@ -1570,6 +1570,12 @@ def get_pod_template_spec(
         if prometheus_shard:
             labels["paasta.yelp.com/prometheus_shard"] = prometheus_shard
 
+        if system_paasta_config.get_kubernetes_add_registration_labels():
+            # Allow Kubernetes Services to easily find
+            # pods belonging to a certain smartstack namespace
+            for registration in self.get_registrations():
+                labels[f"paasta.yelp.com/registrations/{registration}"] = "true"  # type: ignore
```
vkhromov, line 1577: Why the type error is ignored?

vkhromov (changes requested):

paasta_tools/kubernetes_tools.py
TheNavigat: https://github.com/python/mypy/issues/7178

PR #3135, opened 2021-06-28
Add istio sidecar injection option

Buffer0x7cd:
Allow paasta workloads to use Istio sidecar injection

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -301,6 +301,7 @@ KubePodLabels = TypedDict(
         "yelp.com/paasta_git_sha": str,
         "yelp.com/paasta_instance": str,
         "yelp.com/paasta_service": str,
+        "sidecar.istio.io/inject": str,
     },
     total=False,
 )
@@ -320,6 +321,7 @@ class KubernetesDeploymentConfigDict(LongRunningServiceConfigDict, total=False):
     prometheus_port: int
     routable_ip: bool
     pod_management_policy: str
+    sidecar_injection: bool
 
 
 def load_kubernetes_service_config_no_cache(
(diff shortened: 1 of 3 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1440,6 +1442,11 @@ def format_kubernetes_app(self) -> Union[V1Deployment, V1StatefulSet]:
     def get_kubernetes_service_account_name(self) -> Optional[str]:
         return self.config_dict.get("service_account_name", None)
 
+    def is_injection_eanbled(self):
```
nemacysts, line 1445: we probably want to add some types here and give this a more descriptive name - maybe something like
```suggestion
    def is_istio_injection_enabled(self) -> Literal["true", "false"]:
```

or
```suggestion
    def should_inject_istio_sidecar(self) -> Literal["true", "false"]:
```

paasta_tools/kubernetes_tools.py
```diff
@@ -320,6 +321,7 @@ class KubernetesDeploymentConfigDict(LongRunningServiceConfigDict, total=False):
     prometheus_port: int
     routable_ip: bool
     pod_management_policy: str
+    sidecar_injection: bool
```
nemacysts, line 324: should we have a more explicit name here just in case this flag hangs around for a while? wouldn't want us to have to inject another (unrelated) sidecar in the future and have this flag still around :p

paasta_tools/kubernetes_tools.py
Buffer0x7cd: `Literal` seem to be only available in python 3.8 and up. Ref https://docs.python.org/3/library/typing.html#typing.Literal

nemacysts: rip :(

`-> str` it is until we move to py3.8

Buffer0x7cd: closing this PR as this is mixed with another one

PR #3136, opened 2021-06-28
U/msingh/mesh 1144 add envoy sidecars option

Buffer0x7cd:
Add sidecars labels for paasta workload

previous references : https://github.com/Yelp/paasta/pull/3135

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -301,6 +301,7 @@ KubePodLabels = TypedDict(
         "yelp.com/paasta_git_sha": str,
         "yelp.com/paasta_instance": str,
         "yelp.com/paasta_service": str,
+        "sidecar.istio.io/inject": str,
     },
     total=False,
 )
@@ -1580,7 +1581,6 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             for registration in self.get_registrations():
                 labels[f"paasta.yelp.com/registrations/{registration}"] = "true"  # type: ignore
 
-        # Add istio sidecar injection label
         if self.is_istio_sidecar_injection_enabled():
             labels["sidecar.istio.io/inject"] = "true"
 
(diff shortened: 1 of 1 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1440,6 +1441,9 @@ def format_kubernetes_app(self) -> Union[V1Deployment, V1StatefulSet]:
     def get_kubernetes_service_account_name(self) -> Optional[str]:
         return self.config_dict.get("service_account_name", None)
 
+    def is_istio_sidecar_injection_enabled(self) -> bool:
+        return self.config_dict.get("is_istio_sidecar_injection_enabled")
```
nemacysts, line 1445: I think we want 
```suggestion
        return self.config_dict.get("is_istio_sidecar_injection_enabled", False)
```
otherwise we might get a `None` here

paasta_tools/kubernetes_tools.py
```diff
@@ -1576,6 +1580,11 @@ def get_pod_template_spec(
             for registration in self.get_registrations():
                 labels[f"paasta.yelp.com/registrations/{registration}"] = "true"  # type: ignore
 
+        # Add istio sidecar injection label
+        sidecar_injection_enabled = self.is_istio_sidecar_injection_enabled()
+        if sidecar_injection_enabled:
+            labels["sidecar.istio.io/inject"] = "true"
```
nemacysts, line 1586: we should probably be consistent and do something like
```suggestion
        if self.is_istio_sidecar_injection_enabled():
            labels["sidecar.istio.io/inject"] = "true"
```
 - especially since we're not using that value again later anyway

paasta_tools/kubernetes_tools.py
```diff
@@ -1576,6 +1580,10 @@ def get_pod_template_spec(
             for registration in self.get_registrations():
                 labels[f"paasta.yelp.com/registrations/{registration}"] = "true"  # type: ignore
 
+        # Add istio sidecar injection label
```
Rob-Johnson, line 1583: this comment doesn't add much, so either expand (explain what uses the label) or delete it.

Rob-Johnson (approved):
fix and ship

PR #3137, opened 2021-07-01
update schema for is_istio_sidecar_injection_enabled option

Buffer0x7cd:
Updated schema for istio sidecar option

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -655,6 +655,9 @@
                         "OrderedReady",
                         "Parallel"
                     ]
+                },
+                "is_istio_sidecar_injection_enabled": {
+                    "type": "boolean"
                 }
             }
         }


paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -655,6 +655,9 @@
                         "OrderedReady",
                         "Parallel"
                     ]
+                },
+                "is_istio_sidecar_injection_enabled": {
+                    "type": "boolean"
```
TheNavigat, line 660: Maybe add this?

`"default": false`

Otherwise lgtm

TheNavigat (approved):

paasta_tools/cli/schemas/kubernetes_schema.json
Buffer0x7cd: It's taking false as a default value here, https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes_tools.py#L1446 , So i think we can skip adding here

PR #3138, opened 2021-07-02
Use internal copies of ubuntu images for dockerfiles to avoid dockerh…

stug:
…ub rate limits

See PAASTA-17409

### DIFF
--- a/general_itests/fake_simple_service/Makefile
+++ b/general_itests/fake_simple_service/Makefile
@@ -13,9 +13,16 @@
 # limitations under the License.
 
 DOCKER_TAG ?= fake_simple_service-$(USER)-dev
-export DOCKER_TAG
+
+ifeq ($(findstring .yelpcorp.com,$(shell hostname -f)), .yelpcorp.com)
+	DOCKER_REGISTRY ?= docker-dev.yelpcorp.com/
+	PIP_INDEX_URL ?= https://pypi.yelpcorp.com/simple
+else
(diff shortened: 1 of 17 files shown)


nemacysts (commented):
i think we'll want to have the docker registry (or lack thereof) configured as an env var (as we do with PIP_INDEX_URL atm) - otherwise I'm pretty sure that the GHA itests will fail

stug:
> i think we'll want to have the docker registry (or lack thereof) configured as an env var (as we do with PIP_INDEX_URL atm) - otherwise I'm pretty sure that the GHA itests will fail

You are definitely right that they failed...not sure how to parametrize the registry, though...

general_itests/fake_simple_service/Makefile
```diff
@@ -15,7 +15,15 @@
...
+else
+	docker_registry = ""
+	pip_index_url = https://pypi.python.org/simple
+endif
+
 .PHONY: cook-image
 
 cook-image:
-	docker build -t "$$DOCKER_TAG" .
+	docker build --build-arg PIP_INDEX_URL=$(pip_index_url) --build-arg DOCKER_REGISTRY=$(docker_regsitry) -t "$$DOCKER_TAG" .
```
stug, line 29: this was the tricky bit -- something was wrong with the interpolation for the docker_registry.  it was using bash env variables, and I switched it to using variables in the Makefile itself.

nemacysts (approved):

PR #3141, opened 2021-07-09
Make config_utils more generic and reduce assumption of subdir

kedarv:
Right now, this class assumes that you're writing data to `AUTO_SOACONFIG_SUBDIR`. I'm building a new script in `contrib/` that relies on AutoConfigUpdater, and will be writing to the main service folder instead of the subdirectory. This PR removes some assumptions of `AUTO_SOACONFIG_SUBDIR`


kedarv:
I realized that `write_auto_config_data` is also making some small assumptions (eg. comment in yaml file) that need to be stripped out to make this class a bit more generic, will push an update to this branch next week (but a review of what I have so far would be great!)

paasta_tools/contrib/rightsizer_soaconfigs_update.py
```diff
@@ -181,11 +196,19 @@ def main(args):
...
+            updater.write_configs(
+                service,
+                extra_info,
+                existing_recommendations,
+                AUTO_SOACONFIG_SUBDIR,
+                HEADER_COMMENT,
+            )
 
         if args.push_to_remote:
             updater.commit_to_remote(extra_message=extra_message)
```
TheNavigat, line 214: Maybe add `AUTO_SOACONFIG_SUBDIR` here to retain the log verbosity?

TheNavigat (changes requested):

paasta_tools/config_utils.py
```diff
@@ -192,13 +200,13 @@ def validate(self):
             return_code = validate_auto_config_file(filepath) and return_code
         return return_code
 
-    def commit_to_remote(self, extra_message: str = ""):
-        if not self.validate():
+    def commit_to_remote(self, extra_message: str = "", validate: bool = True):
+        if validate and not self.validate():
```
stug, line 204: when would we not want to validate?

stug (approved):
Looks roughly sane to me

paasta_tools/config_utils.py
kedarv: `self.validate()` will fail if we pass a file that is unexpected (not supported) by the validator. This toggle just provides an escape hatch where we don't want to provide a validator (though ideally all files will have some validation of some sort)

PR #3142, opened 2021-07-12
adds script to automatically add deploy groups

kedarv:
This PR utilizes changes made in https://github.com/Yelp/paasta/pull/3141 to automatically mutate and push to configs. Given CLI arguments, this script will check if the service `args.service` has deploy group `args.deploy_group`, and if not, will modify `args.kube_file` and `deploy.yaml`, and insert data to create a new deploy group `args.deploy_group_prefix.deploy_group` with instance count `args.instance_count`. `args.non_relevant_steps` is provided to ignore miscellaneous deploy steps.

### DIFF
--- a/paasta_tools/contrib/deploy_group_update.py
+++ b/paasta_tools/contrib/deploy_group_update.py
@@ -115,7 +115,6 @@ def main(args):
             )
 
         if args.deploy_group not in deploy_groups:
-            # hardcode 1 instance for now
             kube_file[args.deploy_group] = {
                 "deploy_group": f"{args.deploy_group_prefix}.{args.deploy_group}",
                 "instances": args.instance_count,
--- a/paasta_tools/contrib/deploy_group_update.py
+++ b/paasta_tools/contrib/deploy_group_update.py
@@ -1,0 +1,144 @@
+import argparse
+import logging
+
+from paasta_tools.config_utils import AutoConfigUpdater
+from paasta_tools.utils import DEFAULT_SOA_CONFIGS_GIT_URL
+from paasta_tools.utils import format_git_url
+from paasta_tools.utils import load_system_paasta_config
+
+log = logging.getLogger(__name__)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="")
+    parser.add_argument(
+        "--git-remote",
+        help="Master git repo for soaconfigs",
+        default=None,
+        dest="git_remote",
+    )
+    parser.add_argument(
+        "--branch", help="Branch name to push to", required=True, dest="branch",
+    )
+    parser.add_argument(
+        "--local-dir",
+        help="Act on configs in the local directory rather than cloning the git_remote",
+        required=False,
+        default=None,
+        dest="local_dir",
+    )
+    parser.add_argument(
+        "-v",
+        "--verbose",
+        help="Logging verbosity",
+        action="store_true",
+        dest="verbose",
+    )
+    parser.add_argument(
+        "--source-id",
+        help="String to attribute the changes in the commit message.",
+        required=False,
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/deploy_group_update.py
```diff
@@ -0,0 +1,144 @@
...
+        )
+
+        # Ensure that deploy steps and groups agree before proceeding
+        if deploy_groups.keys() != pipeline_steps:
+            log.error(
+                f"deploy groups {deploy_groups.keys()} did not match deploy steps {pipeline_steps}. cannot proceed until these files are in agreement"
+            )
+
+        if args.deploy_group not in deploy_groups:
+            # hardcode 1 instance for now
```
connorworley, line 118: this is no longer the case right?

paasta_tools/contrib/deploy_group_update.py
```diff
@@ -0,0 +1,144 @@
...
+                    "step": f"{args.deploy_group_prefix}.{args.deploy_group}",
+                    "wait_for_deployment": True,
+                    "disabled": True,
+                }
+            )
+
+            updater.write_configs(args.service, "deploy", deploy)
+            updater.write_configs(args.service, args.kube_file, kube_file)
+
+            updater.commit_to_remote(validate=False)
```
connorworley, line 134: should we turn on validation?

connorworley (approved):
this looks good to me, just a couple of small comments.

we should do some trials to measure how long this takes to run.

paasta_tools/contrib/deploy_group_update.py
kedarv: removed!

paasta_tools/contrib/deploy_group_update.py
kedarv: https://github.com/Yelp/paasta/pull/3141#discussion_r668140607 explains why I added this, when we move away from our proof of concept yea it should be turned on and the appropriate validators should be added

paasta_tools/contrib/deploy_group_update.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+    return default_git_remote
+
+
+def main(args):
+    updater = AutoConfigUpdater(
+        config_source=args.source_id,
+        git_remote=args.git_remote or get_default_git_remote(),
+        branch=args.branch,
+        working_dir=args.local_dir or "/nail/tmp",
```
TheNavigat, line 99: nit: Maybe set `local_dir`'s default value to `/nail/tmp`?

paasta_tools/contrib/deploy_group_update.py
```diff
@@ -0,0 +1,143 @@
...
+        if args.deploy_group not in deploy_groups:
+            kube_file[args.deploy_group] = {
+                "deploy_group": f"{args.deploy_group_prefix}.{args.deploy_group}",
+                "instances": args.instance_count,
+            }
+            deploy["pipeline"].append(
+                {
+                    "step": f"{args.deploy_group_prefix}.{args.deploy_group}",
+                    "wait_for_deployment": True,
+                    "disabled": True,
```
TheNavigat, line 126: `disabled`? Huh?

TheNavigat (approved):

paasta_tools/contrib/deploy_group_update.py
kedarv: If we default it, then the condition a line below `do_clone=args.local_dir is None,` won't work properly

TheNavigat: Oh, right, my bad. Thanks for explaining :)

PR #3145, opened 2021-07-15
TRON-1625: Support adding paasta default_volumes to tron master k8s config

jfongatyelp:
This change adds support for updating/adding the `default_volumes` key into MASTER config, using the paasta system-level default volumes (i.e. /etc/paasta/volumes.json).

A manual dry-run of `setup_tron_namespace` shows it would generate the expected config: https://fluffy.yelpcorp.com/i/zqDppdXcs5zTqjzXBxX3Lv2DdBg3t7Pd.html

Note that this currently only _updates_ the `k8s_options` key if it already exists in MASTER config (i.e. we've already specified it manually in yelpsoa-configs ourselves), to allow us to gate this only to tron servers that have already been updated with k8s support.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -650,6 +650,15 @@ def format_master_config(master_config, default_volumes, dockercfg_location):
         }
     )
     master_config["mesos_options"] = mesos_options
+
+    k8s_options = master_config.get("k8s_options", {})
+    if k8s_options:
+        # Only add default volumes if we already have k8s_options
+        # TODO: Gate this on whether k8s on tron is enabled
+        k8s_options.update(
+            {"default_volumes": format_volumes(default_volumes),}
+        )
+        master_config["k8s_options"] = k8s_options
     return master_config
 
 
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -654,7 +654,6 @@ def format_master_config(master_config, default_volumes, dockercfg_location):
     k8s_options = master_config.get("k8s_options", {})
     if k8s_options:
         # Only add default volumes if we already have k8s_options
-        # TODO: Gate this on whether k8s on tron is enabled
         k8s_options.update(
             {"default_volumes": format_volumes(default_volumes),}
         )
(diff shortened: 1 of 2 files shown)


paasta_tools/tron_tools.py
```diff
@@ -650,6 +650,15 @@ def format_master_config(master_config, default_volumes, dockercfg_location):
...
     )
     master_config["mesos_options"] = mesos_options
+
+    k8s_options = master_config.get("k8s_options", {})
+    if k8s_options:
+        # Only add default volumes if we already have k8s_options
+        # TODO: Gate this on whether k8s on tron is enabled
+        k8s_options.update(
+            {"default_volumes": format_volumes(default_volumes),}
+        )
```
lanceleih, line 660: How come we also don't add dockercfg_location in this case?

nemacysts: afaik, we've configured k8s to know how to talk to our internal registry so we don't need to pass any additional config for that

paasta_tools/tron_tools.py
```diff
@@ -650,6 +650,15 @@ def format_master_config(master_config, default_volumes, dockercfg_location):
         }
     )
     master_config["mesos_options"] = mesos_options
+
+    k8s_options = master_config.get("k8s_options", {})
+    if k8s_options:
+        # Only add default volumes if we already have k8s_options
+        # TODO: Gate this on whether k8s on tron is enabled
```
nemacysts, line 657: i think that as long as k8s_options is there, we can go ahead and add these in

we can probably also just unconditionally (just like we do for mesos) do this once we've released a new tron version that won't barf on k8_options (which should be soon!)

nemacysts (approved):
i'd probably remove the `# TODO: Gate this on whether k8s on tron is enabled` line, but lgtm!

paasta_tools/tron_tools.py
jfongatyelp: +1 to the above -- The contents of dockercfg_location gets thrown into mesos task's `uris` field, so it basically just controls how we talk to various docker registries it seems, which our kubernetes clusters already know how to do.

nemacysts (approved):

PR #3150, opened 2021-07-19
PAASTA-17414 make paasta status --new the default

cuza:
After having a new and more helpful paasta status, it's hidden behind an obscure flag. This PR adds support for `paasta status --old` and for setting the `paasta_status_version` in `etc/paasta`

Note:  `paasta_status_version` needs to be set in `etc/paasta` to make the --new version the default one

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -131,13 +131,24 @@ def add_subparser(subparsers,) -> None:
         default=DEFAULT_SOA_DIR,
         help="define a different soa config directory",
     )
-    status_parser.add_argument(
+
+    version = status_parser.add_mutually_exclusive_group()
+
+    version.add_argument(
         "--new",
         dest="new",
         action="store_true",
         default=False,
         help="Use experimental new version of paasta status for services",
     )
+    version.add_argument(
+        "--old",
+        dest="old",
+        default=False,
+        action="store_true",
+        help="Use the old version of paasta status for services",
+    )
+
     add_instance_filter_arguments(status_parser)
     status_parser.set_defaults(command=paasta_status)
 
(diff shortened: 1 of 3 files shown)


EvanKrall (approved):

paasta_tools/cli/cmds/status.py
```diff
@@ -2241,6 +2253,20 @@ def status_marathon_job_human(
         )
 
 
+def _get_paasta_status_version(args, system_paasta_config) -> bool:
```
stug, line 2256: Since this returns a boolean, can we call it something that explains what the boolean is like "_use_new_paasta_status()"?

paasta_tools/cli/cmds/status.py
```diff
@@ -2241,6 +2253,20 @@ def status_marathon_job_human(
...
+        return True
+    elif args.old:
+        return False
+    else:
+        if system_paasta_config.get_paasta_status_version() == "old":
+            return False
+        elif system_paasta_config.get_paasta_status_version() == "new":
+            return True
+        else:
+            return True
```
stug, line 2267: nit: you can simplify this to `system_paasta_config.get_paasta_status_version() == "new"` maybe?  I guess that doesn't cover the case where it's neither "new" nor "old" but I don't think we're really going to encounter that (since we control it anyway)

stug (approved):
I have one small complaint about a function name, but otherwise looks good to me.

I considered suggesting that we add a test...but not sure how worthwhile that is.

paasta_tools/cli/cmds/status.py
cuza: Is like that exactly to cover the case where it's neither "new" nor "old" since I don't trust humans and they can have typos. The other thing is that all this code is probably gonna get deleted soon once we wipe all the `--old` code

PR #3151, opened 2021-07-19
Update paasta secret sync to support other namespaces

jfongatyelp:
This updates the paasta secret sync to optionally sync yelpsoa-config specified secrets into a namespace different than `paasta`, but leaves the default as `paasta`.

In addition to creating the resources in the new namespace, the naming convention changes from `paasta-` prefixed to `{namespace}-` prefixed.

I have added tests to verify leaving namespace unspecified results in reading from/writing to the `paasta` namespace.

Manual test runs against kubestage also work as expected:
```
$  KUBECONFIG=./etc_paasta_for_development/admin.conf PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_for_development/ python paasta_tools/kubernetes/bin/paasta_secrets_sync.py -n tron compute-infra-test-service -v
INFO:__main__:new-test-secret for compute-infra-test-service not found, creating

$  KUBECONFIG=./etc_paasta_for_development/admin.conf PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_for_development/ python paasta_tools/kubernetes/bin/paasta_secrets_sync.py -n tron compute-infra-test-service -v
INFO:__main__:new-test-secret for compute-infra-test-service up to date

$ kubectl-kubestage get cm -n tron
INFO: You are using an Okta authenticated kubectl wrapper
NAME                                                               DATA   AGE
istio-ca-root-cert                                                 1      3d20h
tron-secret-compute-infra-test-service-new-test-secret-signature   1      92s

 $ kubectl-kubestage get secret -n tron
INFO: You are using an Okta authenticated kubectl wrapper
NAME                                                     TYPE                                  DATA   AGE
default-token-9llsl                                      kubernetes.io/service-account-token   3      3d20h
tron-secret-compute-infra-test-service-new-test-secret   Opaque                                1      104s
```

I also verified this new version continues to work without specifying namespace on kubestage for noops, updates, and creating secrets.

The intent of this change is to sync paasta secrets that have tronjobs configured into th `tron` namespace to maintain parity with paasta secrets on mesos.

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -61,6 +61,13 @@ def parse_args() -> argparse.Namespace:
         help="define a different soa config directory",
     )
     parser.add_argument(
+        "-n",
+        "--namespace",
+        dest="namespace",
+        default="paasta",
+        help="destination namespace for secrets (Default: %(default)s)",
+    )
+    parser.add_argument(
         "-v", "--verbose", action="store_true", dest="verbose", default=False
     )
     args = parser.parse_args()
@@ -89,6 +96,7 @@ def main() -> None:
         secret_provider_name=secret_provider_name,
         vault_cluster_config=vault_cluster_config,
         soa_dir=args.soa_dir,
+        namespace=args.namespace,
     ) else sys.exit(1)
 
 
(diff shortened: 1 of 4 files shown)


nemacysts (approved):

jfongatyelp:
There was some question around whether https://github.com/Yelp/paasta/pull/3009/files should be shipped instead of this PR.

I will continue with pushing this change as I don't believe it blocks #3009 -- that is a larger change of moving the logic to determine which secrets to sync *into* paasta_secret_sync, rather than leaving it to be determined externally.

mattmb (approved):
Sounds good to me

PR #3152, opened 2021-07-21
Print a subset of paasta status every minute when bouncing

EvanKrall:
This prints the summary of pod statuses for each service/instance when bouncing.

This is based on https://github.com/Yelp/paasta/pull/3149, which makes this extra polling a bit easier. If we would rather make a smaller change, it's probably not a big deal to do this polling a different way.


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1405,6 +1490,7 @@ async def periodically_update_progressbar():
                 while True:
                     await asyncio.sleep(60)
                     bar.update(finished_instances)
+                    print()
```
cuza, line 1493: Did this print() stops the progress bar from acting out weird??

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -84,6 +85,9 @@
 DEFAULT_SLACK_CHANNEL = "#deploy"
 DEFAULT_STUCK_BOUNCE_RUNBOOK = "y/stuckbounce"
 
+TIME_BEFORE_FIRST_DIAGNOSIS = 300
+TIME_BETWEEN_DIAGNOSES = 60
```
stug, line 89: I am wondering if we should make these configurable either in the system_paasta_config or in the MFD args.  60s seems decent to me, but it's possible it'll increase the load on the api and we'll want to decrease?

stug (approved):

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: Yeah, that's probably wise. I should also probably make the polling interval in `check_if_instance_is_done` and the number of threads in the polling threadpool configurable.

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: It seemed to help, though not entirely.

PR #3153, opened 2021-07-21
Make wait-for-deployment warn pct time configurable

kawaiwanyelp:
### Description
Currently, we warn users when their deployment has reached 75% of the mfd timeout. This 75% is hardcoded, which isn't great. So, I've added an option to mfd to make the warn percentage configurable.

I've also lowered the default pct to 50% to leave more time for stuck bounce debugging. At 75% with a default 1hr/3600s timeout, we only get 15m of debugging time, which means by the time we've taken action, the deploy has already timed out, forcing users to restart Jenkins. 50% or 30m is better. If we find that some services naturally deploy slower, we can raise the warn pct for their mfd since it is now configurable.

### Testing
`make test` passes
manual test on `compute-infra-test-service` works. i ran the following command, and got a warning at 15s, after which it timed out at 30s, as expected:
```
paasta mark-for-deployment --commit <commit> --deploy-group dev-stage.everything -s compute-infra-test-service --wait-for-deployment -t 30 -w 50
```

### Notes
Assuming this is released, we'd need to make another change in our internal jenkins libraries so that `--warn` is configurable through `deploy.yaml`.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -75,6 +75,7 @@ from paasta_tools.utils import TimeoutError
 
 
 DEFAULT_DEPLOYMENT_TIMEOUT = 3600  # seconds
+DEFAULT_WARN_PERCENT = 50
 DEFAULT_AUTO_CERTIFY_DELAY = 600  # seconds
 DEFAULT_SLACK_CHANNEL = "#deploy"
 DEFAULT_STUCK_BOUNCE_RUNBOOK = "y/stuckbounce"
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -156,6 +157,18 @@ def add_subparser(subparsers):
...
+    list_parser.add_argument(
+        "-w",
+        "--warn",
+        dest="warn",
+        type=int,
+        default=DEFAULT_WARN_PERCENT,
+        help=(
+            "Percent of timeout to warn at if the deployment hasn't finished. "
+            "For example, --warn=75 will warn at 75%% of the timeout. "
+            "Defaults to %(default)s."
```
EvanKrall, line 169: took me a second to realize that the `s` here is part of argparse's format string

EvanKrall (approved):
you could make the diff a lot smaller by keeping it called `self.timeout_percentage_before_reminding` and setting a default value for it on `__init__`, but ¯\\\_(ツ)\_/¯

cuza (approved):

stug (approved):

PR #3160, opened 2021-08-03
Add logspout feature flag to all tron on k8s containers

nemacysts:
We don't want to turn this on for existing Mesos tronjobs since the logs
won't be used in that case.

We'll be using logspout to ship tronjob logs to our logging streams and
then displaying those in Tron rather than doing what we did with
Tron-on-Mesos and asking Mesos (or, in this case, K8s) to feed us logs
periodically.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -306,14 +306,8 @@ class TronActionConfig(InstanceConfig):
     def get_spark_cluster_manager(self):
         return self.config_dict.get("spark_cluster_manager", "mesos")
 
-    def get_env(self, use_k8s: bool = False) -> Dict[str, str]:
+    def get_env(self):
         env = super().get_env()
-        # for Tron-on-K8s, we want to ship tronjob output through logspout
-        # such that this output eventually makes it into our per-instance
-        # log streams automatically
-        if use_k8s:
-            env["TRON_ENABLE_LOGSPOUT"] = "1"
-
         if self.get_executor() == "spark":
             spark_config_dict = self.get_spark_config_dict()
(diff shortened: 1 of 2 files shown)


jfongatyelp (approved):
Cool, nice way to roll this out

paasta_tools/tron_tools.py
```diff
@@ -718,10 +718,15 @@ def format_tron_action_dict(action_config: TronActionConfig, use_k8s: bool = Fal
...
+
         result["secret_env"] = action_config.get_secret_env()
-        # For k8s, we do not want secret envvars to be duplicated in both `env` and `secret_env`
         all_env = action_config.get_env()
+        # For k8s, we do not want secret envvars to be duplicated in both `env` and `secret_env`
         result["env"] = {k: v for k, v in all_env.items() if not is_secret_ref(v)}
+        # for Tron-on-K8s, we want to ship tronjob output through logspout
+        # such that this output eventually makes it into our per-instance
+        # log streams automatically
+        result["env"]["TRON_ENABLE_LOGSPOUT"] = "1"
```
jfongatyelp, line 729: Sorry, thinking about this also -- technically this toggle controls whether we send to a per-instance stream_paasta_* stream vs per-service, logspout is already 'enabled' on tron.  

Seems like a very minor nit if this is going to be enabled everywhere eventually, so feel free to not block on this.

jfongatyelp (approved):

paasta_tools/tron_tools.py
nemacysts: `@marley what is temporary`

nah, you're right - I can rename this to `TRON_ENABLE_PER_INSTANCE_LOGSPOUT` just to be more explicit

PR #3165, opened 2021-08-09
Increase default mark-for-deployment timeout from 1 to 3 hours and de…

stug:
…crease default warning percentage so that we warn after about the same amount of time

See PAASTA-17425 for details

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -85,8 +85,8 @@ from paasta_tools.utils import RollbackTypes
 from paasta_tools.utils import TimeoutError
 
 
-DEFAULT_DEPLOYMENT_TIMEOUT = 3600  # seconds
-DEFAULT_WARN_PERCENT = 50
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -85,8 +85,8 @@
 from paasta_tools.utils import TimeoutError
 
 
-DEFAULT_DEPLOYMENT_TIMEOUT = 3600  # seconds
-DEFAULT_WARN_PERCENT = 50
+DEFAULT_DEPLOYMENT_TIMEOUT = 3 * 3600  # seconds
+DEFAULT_WARN_PERCENT = 17  # ~30min for default timeout
```
EvanKrall, line 89: this is going to warn pretty early for anybody who's set their timeout to something less than 3 hours. There's a whole bunch of `3600` and a handful of shorter timeouts: https://fluffy.yelpcorp.com/i/rbDKlxt1Qk052FM0KcKgGfG7zJRZ7qJm.html

Maybe delete all `timeout: 3600`s and manually set `warn` to something reasonable on anything with timeout < 3600?

EvanKrall (approved):

PR #3167, opened 2021-08-10
Fix cancellation of wait_for_deployment by replacing green_lights wit…

EvanKrall:
…h asyncio task cancellation. PAASTA-17443

In #3149, I refactored wait_for_deployment to run under asyncio rather than as a separate thread. The old thread-based approach used `threading.Event` (essentially a boolean you can wait on) objects that we called "green lights" to know when to die -- every iteration through certain loops, we would check those Events to see if they were still set, and if not, exited. This is how we stopped waiting for deployment of the new version when you click the rollback button, for example.

In my refactor, I didn't add any code that checked those events, so toggling them had no effect.

In this PR, I change all calls to `self.wait_for_deployment_green_lights[some_sha].clear()` to `self.stop_waiting_for_deployment(some_sha)`, which uses `asyncio.cancel` to kill the corresponding task object.

I also fixed some test bugs that only show up when you run `tests/cli/test_cmds_mark_for_deployment.py` by itself -- I think some sort of test pollution from earlier tests was making these pass when you ran the whole test suite.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -28,12 +28,9 @@ import socket
 import sys
 import time
 import traceback
-from collections import defaultdict
-from threading import Event
 from threading import Thread
 from typing import Any
 from typing import Collection
-from typing import DefaultDict
 from typing import Dict
 from typing import Iterator
 from typing import List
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1037,33 +1031,43 @@ def on_enter_deploy_cancelled(self) -> None:
         if self.deploy_group_is_set_to_notify("notify_after_abort"):
             self.ping_authors("Deploy cancelled")
 
-    def do_wait_for_deployment(self, target_commit: str) -> None:
+    def stop_waiting_for_deployment(self, target_commit: str) -> None:
+        try:
+            self.wait_for_deployment_tasks[target_commit].cancel()
+        except KeyError:
+            pass
```
kawaiwanyelp, line 1038: Theoretically, you could also run into an `asyncio.InvalidStateError` if you attempted to cancel on a non-pending task, so I'd catch that exc too. For good practice, I'd also remove the cancelled task from the tasks dict, so you'd never run into that case.

nit: I'd also await the task afterwards to allow it to finish gracefully cancelling, but right now that would just re-raise the `CancelledError` that killed that task in this control task, which would stop the entire thing, so it's not really necessary. I'd actually prefer that the commit's task just `pass`ed on a cancel, but that's just my preference. Feel free to ignore.

kawaiwanyelp (approved):
minor nit changes, but overall lgtm

PR #3171, opened 2021-08-11
Make m-f-d poll bounce status endpoint for progress

kawaiwanyelp:
### Description
This makes `paasta mark-for-deployment --wait-for-deployment` use the new bounce status endpoint for progress instead of the instance status endpoint. The case logic is basically the same, just a little simplified.

This PR also deprecates support for bouncing Marathon instances. If instances of that type are bounced, they will considered non-supported/unbounceable instances.

### Testing
`make test`
manual testing of `check_if_instance_is_done` on local `paasta api` on kubestage
e2e testing using `mark-for-deployment`, against kubestage as well

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -1421,143 +1421,100 @@ def check_if_instance_is_done(
             )
             return False
 
-    log.debug(
-        "Inspecting the deployment status of {}.{} on {}".format(
-            service, instance, cluster
-        )
-    )
+    inst_str = f"{service}.{instance} in {cluster}"
+    log.debug(f"Inspecting the deployment status of {inst_str}")
+
+    status = None
     try:
-        status = None
-        status = api.service.status_instance(
-            service=service,
-            instance=instance,
-            include_smartstack=False,
-            include_envoy=False,
-            include_mesos=False,
-        )
+        status = api.service.bounce_status_instance(service=service, instance=instance)
     except api.api_error as e:
-        if e.status == 404:
+        if e.status == 404:  # non-existent instance
             # TODO(PAASTA-17290): just print the error message so that we
             # can distinguish between sources of 404s
             log.warning(
                 "Can't get status for instance {}, service {} in "
                 "cluster {}. This is normally because it is a new "
-                "service that hasn't been deployed by PaaSTA yet".format(
+                "service that hasn't been deployed by PaaSTA yet.".format(
                     instance, service, cluster
                 )
             )
-        else:
+        else:  # 500 - error talking to api
             log.warning(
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1323,143 +1323,88 @@ def check_if_instance_is_done(
...
-            if git_sha not in active_git_shas:
-                print(f"  {service}.{instance} on {cluster} not yet running {git_sha}.")
-                return False
+    # Case: bounce not yet started
+    # at this pt, only one app/version is active, and a 'sha' is [git_sha, config_sha],
+    # so active_shas[0][0] is the active git sha (not config sha)
+    if not git_sha.startswith(status.active_shas[0][0]):
+        print(
+            f"  {inst_str} hasn't started bouncing to the desired version yet; "
+            f"it is still on: {status.active_shas[0][0]}"
```
EvanKrall, line 1376: I'd show all of the active shas here: `f"it is still running {', '.join({g for g, c in active_shas})}"`

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1323,143 +1323,88 @@ def check_if_instance_is_done(
...
-                service, instance, cluster
-            )
-        )
-        return True
-    elif (
-        long_running_status.expected_instance_count == 0
-        or long_running_status.desired_state == "stop"
-    ):
+
+    if not status:  # 204 - instance is not bounceable
```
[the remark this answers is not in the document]
kawaiwanyelp: I tested this by forcing a locally running paasta api to return a 204 immediately (though the bounce status endpoint does naturally return one), and status comes back as an empty string. It is likely because the way we construct the 204 response is create one using `Response()`, setting the status code to 204, then returning it immediately.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1323,143 +1323,100 @@ def check_if_instance_is_done(
...
-                "has {} out of {} required instances (out of a total of {})".format(
-                    service,
-                    instance,
-                    cluster,
-                    long_running_status.running_instance_count,
-                    required_instance_count,
-                    long_running_status.expected_instance_count,
-                )
+                f"  {inst_str} hasn't started bouncing to {short_git_sha}, because a previous "
+                f"bounces aren't done yet, with these other versions still running: {non_desired_shas}"
```
EvanKrall, line 1381: This implies that we have to wait for previous bounces to finish before starting to deploy the new version, which I'm pretty sure isn't true. Once we see `short_git_sha` in `active_shas`, we've started launching the new version.

EvanKrall (approved):
fix'n'ship

cuza (approved):

PR #3175, opened 2021-08-13
Stop returning no services if kubelet down

mattmb:
PAASTA-17446 for details. If the kubelet is down the services can still
be running so we don't want to tell nerve that there are no services
since that will cause it to deregister the services. I think that is
safe although it is very unsafe if we accidentally stopped all the
kubelets (and we have done that before). We also restart kubelets to
renew certs which triggers this case and deregisters services for a few
minutes before re-registering them when the kubelet starts again.
That creates some unnecessary churn in the mesh.


keymone (approved):
do we want to put in a few retries to give kubelet a chance to come back up?

mattmb:
I think that should be okay since configure_nerve runs every 5s it'll just get retried on the next run

PR #3176, opened 2021-08-13
Cleanup Tron's Completed Pods

lanceleih:
### Description
Once Tronjobs have finished running on Kubernetes, they are marked as PodCompleted, but they stay around and linger waiting to be terminated. The purpose of the script is to terminate the completed pods 10 minutes after they have completed.

### Test
`kubectl get pods -n tron` before command
- https://fluffy.yelpcorp.com/i/HTDBBnTv7C8JQLF8QSkj3654z4zSWGqQ.html
`paasta_cleanup_tron_completed_pods` running
- https://fluffy.yelpcorp.com/i/7XW1p7MLKPNjhz0D9tCvbLTtgLDZj0Pz.html
`kubectl get pods -n tron` after command
- https://fluffy.yelpcorp.com/i/6QW1vkGwZFjQ5HC0522P2dDRsCCrMGrq.html


paasta_tools/cleanup_tron_completed_pods.py
```diff
@@ -0,0 +1,63 @@
...
+                        namespace=pod.metadata.namespace,
+                        grace_period_seconds=0,
+                        propagation_policy="Background",
+                    )
+                    successes.append(pod_name)
+                except Exception as e:
+                    errors.append((pod_name, e))
+
+    if successes:
+        print(
```
nemacysts, line 44: might be a good idea to switch to use a python logger instead of print (and adding a `--verbose` argument)

paasta_tools/cleanup_tron_completed_pods.py
```diff
@@ -0,0 +1,63 @@
...
+from paasta_tools.kubernetes_tools import get_all_pods
+from paasta_tools.kubernetes_tools import get_pod_condition
+from paasta_tools.kubernetes_tools import is_pod_completed
+from paasta_tools.kubernetes_tools import KubeClient
+
+
+def main():
+    kube_client = KubeClient()
+    pods = get_all_pods(kube_client, "tron")
+    allowed_uptime_minutes = 10
```
nemacysts, line 15: it's probably a good idea to make this a generic script and take in a namespace and time-to-delete as an argument

paasta_tools/cleanup_tron_completed_pods.py
```diff
@@ -0,0 +1,63 @@
...
+    for pod in pods:
+        if is_pod_completed(pod):
+            time_finished = get_pod_condition(
+                pod, "ContainersReady"
+            ).last_transition_time
+            time_now = datetime.now(tzutc())
+            # convert total seconds since completion to minutes
+            completed_since = (
+                time_now - time_finished
+            ).total_seconds() / seconds_per_minute
```
nemacysts, line 28: thoughts on turning this into a function? then we'd have something like:
```Python
if (
    is_pod_completed(pod) 
    and _time_since_completion(pod) > time_to_delete
):
    ...
```

nemacysts (commented):
looking good!

paasta_tools/cleanup_tron_completed_pods.py
lanceleih: Added

paasta_tools/cleanup_tron_completed_pods.py
lanceleih: Ah yes, that's a good point. Adding logger and verbose 👍

paasta_tools/prune_completed_pods.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+def completed_since(pod: V1Pod, allowed_uptime_minutes: int) -> bool:
+    seconds_per_minute = 60
+    time_finished = get_pod_condition(pod, "ContainersReady").last_transition_time
+    time_now = datetime.now(tzutc())
+    # convert total seconds since completion to minutes
+    completed_since = (time_now - time_finished).total_seconds() / seconds_per_minute
+
+    return True if completed_since > allowed_uptime_minutes else False
```
nemacysts, line 46: super-nit:
```suggestion
    return completed_since > allowed_uptime_minutes
```

paasta_tools/prune_completed_pods.py
```diff
@@ -0,0 +1,125 @@
...
+def main():
+    args = parse_args()
+
+    if not args.namespace:
+        print("No namespace specified")
+        sys.exit(0)
+
+    if not args.minutes:
+        print("No minutes specified.")
+        sys.exit(0)
```
nemacysts, line 77: thoughts on making these required in argparse?

lanceleih: Make it so 👍

paasta_tools/prune_completed_pods.py
lanceleih: Done

paasta_tools/prune_completed_pods.py
lanceleih: Done

paasta_tools/prune_completed_pods.py
```diff
@@ -0,0 +1,134 @@
...
+    allowed_uptime_minutes = int(args.minutes)
+    completed_pods = []
+
+    setup_logging(args.verbose)
+
+    for pod in pods:
+        if is_pod_completed(pod) and completed_since(pod, allowed_uptime_minutes):
+            completed_pods.append(pod)
+
+    if len(completed_pods) < 1:
```
nemacysts, line 97:
```suggestion
    if not completed_pods:
```
might be a little more idiomatic - but feel free to ignore me

paasta_tools/prune_completed_pods.py
```diff
@@ -0,0 +1,134 @@
...
+        log.debug("No completed pods to terminate.")
+        sys.exit(1)
+
+    if args.dry_run:
+        log.debug(
+            "Dry run - {} would have terminated the following completed pods:\n ".format(
+                socket.gethostname()
```
nemacysts, line 104: should be fine to leave the hostname out of the log messages we emit - if this gets picked up by our logging aggregators those will inject the hostname

paasta_tools/prune_completed_pods.py
```diff
@@ -0,0 +1,134 @@
...
+
+def main():
+    args = parse_args()
+
+    kube_client = KubeClient()
+    pods = get_all_pods(kube_client, args.namespace)
+    allowed_uptime_minutes = int(args.minutes)
+    completed_pods = []
+
+    setup_logging(args.verbose)
```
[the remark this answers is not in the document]
lanceleih: Done.

paasta_tools/prune_completed_pods.py
lanceleih: Done 👍

paasta_tools/prune_completed_pods.py
lanceleih: Left it before since we were using print, but removing from script entirely since we're logging everything.

paasta_tools/prune_completed_pods.py
```diff
@@ -0,0 +1,130 @@
...
+        help="Namespace of the pods to terminate from",
+        required=True,
+    )
+    parser.add_argument(
+        "-m",
+        "--minutes",
+        default=0,
+        help="Minutes since the pods' completion. Terminates pods based on time since completion.",
+        required=True,
+    )
```
nemacysts, line 34:
```suggestion
    parser.add_argument(
        "-m",
        "--minutes",
        default=0,
        help="Minutes since the pods' completion. Terminates pods based on time since completion.",
        required=True,
        type=int,
    )
```

should allow us to get rid of the cast that we do on line 90

paasta_tools/prune_completed_pods.py
```diff
@@ -0,0 +1,130 @@
...
+    allowed_uptime_minutes = int(args.minutes)
+    completed_pods = []
+
+    for pod in pods:
+        if is_pod_completed(pod) and _completed_since(pod, allowed_uptime_minutes):
+            completed_pods.append(pod)
+
+    if not len(completed_pods):
+        log.debug("No completed pods to terminate.")
+        sys.exit(1)
```
nemacysts, line 99: hmm - thinking about this a little more, should we exit 0 here instead? it's possible that we have nothing to cleanup and we shouldn't treat that as an error case

paasta_tools/prune_completed_pods.py
```diff
@@ -0,0 +1,130 @@
...
+
+    if args.dry_run:
+        log.debug(
+            "Dry run would have terminated the following completed pods:\n "
+            + "\n ".join([pod.metadata.name for pod in completed_pods])
+        )
+        sys.exit(1)
```
nemacysts, line 106: probably also want to exit 0 here as well since there were no errors

nemacysts (commented):
should be good to go after this :)

jfongatyelp (approved):

paasta_tools/prune_completed_pods.py
lanceleih: Done

paasta_tools/prune_completed_pods.py
lanceleih: Done

paasta_tools/prune_completed_pods.py
lanceleih: Done

paasta_tools/prune_completed_pods.py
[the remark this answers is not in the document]
lanceleih: Removed

paasta_tools/prune_completed_pods.py
```diff
@@ -0,0 +1,130 @@
...
+    allowed_uptime_minutes = int(args.minutes)
+    completed_pods = []
+
+    for pod in pods:
+        if is_pod_completed(pod) and _completed_since(pod, allowed_uptime_minutes):
+            completed_pods.append(pod)
+
+    if not len(completed_pods):
+        log.debug("No completed pods to terminate.")
+        sys.exit(1)
```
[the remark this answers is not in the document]
lanceleih: Done

nemacysts (approved):

PR #3179, opened 2021-08-17
U/msingh/create k8s services

Buffer0x7cd:


paasta_tools/kubernetes_tools.py
```diff
@@ -1964,6 +1969,15 @@ def list_deployments(
...
 
+def list_kubernetes_services(kube_client: KubeClient,) -> Sequence[KubernetesService]:
+    services = kube_client.core.list_namespaced_service(namespace="paasta")
+
+    return [
+        KubernetesService(name=item.metadata.name, labels=item.metadata.labels)
+        for item in services.items
+    ]
+
+
```
keymone, line 1980: this and `KubernetesService` don't seem like they have enough meat to be separate entities, let's just use kubernetes entities directly

paasta_tools/setup_kubernetes_services.py
```diff
@@ -0,0 +1,183 @@
...
+import argparse
+import logging
+import sys
+from typing import Sequence
+
+from kubernetes.client import V1APIService
+from kubernetes.client import V1ObjectMeta
+from kubernetes.client import V1Service
+from kubernetes.client import V1ServicePort
+from kubernetes.client import V1ServiceSpec
```
keymone, line 32: it's going to become cumbersome to import every single entity separately, let's just `import kubernetes.client as k8s`?

keymone (changes requested):

keymone:
while we're at it, let's rename setup_kubernetes_services to setup_paasta_mesh or setup_istio_mesh?

paasta_tools/setup_istio_mesh.py
```diff
@@ -0,0 +1,235 @@
...
+    args = parser.parse_args()
+    return args
+
+
+def build_namespace_port_mapping(file_path: str) -> Mapping:
+    paasta_namespaces = {}
+    with open(file_path) as service_file:
+        smartstack_namespaces = json.load(service_file)
+        for namespace, endpoint in smartstack_namespaces.items():
+            paasta_namespaces[namespace] = endpoint["port"]
```
keymone, line 99: i don't think we need to bother with this, just return smartstack_namespaces directly, we'll know how to use them

paasta_tools/setup_istio_mesh.py
```diff
@@ -0,0 +1,235 @@
...
+    return paasta_namespaces
+
+
+def format_service_tuple(services: k8s.V1ServiceList) -> Sequence[KubernetesService]:
+    return [
+        KubernetesService(name=item.metadata.name, labels=item.metadata.labels)
+        for item in services.items
+    ]
```
keymone, line 107: this isn't used anywhere

paasta_tools/setup_istio_mesh.py
```diff
@@ -0,0 +1,235 @@
...
+    service_meta = k8s.V1ObjectMeta(name=UNIFIED_K8S_SVC_NAME)
+
+    # Add port 1337 for envoy unified listener.
+    # Clients can connect to this listenner and set x-yelp-svc header for routing
+    ports = [
+        k8s.V1ServicePort(
+            name=f"p{UNIFIED_SVC_PORT}",
+            port=UNIFIED_SVC_PORT,
+            protocol="TCP",
+            target_port=PAASTA_SVC_PORT,
```
keymone, line 139: we need appProtocol=http, otherwise istio is trying to guess protocol from service name

paasta_tools/setup_istio_mesh.py
```diff
@@ -0,0 +1,235 @@
...
+        )
+    ]
+
+    # Add smartstack ports for routing, Clients can connect to this
+    # Directly without need of setting x-yelp-svc header
+    for port in port_list:
+        port_spec = k8s.V1ServicePort(
+            name=f"p{port}", port=port, protocol="TCP", target_port=PAASTA_SVC_PORT
```
keymone, line 147: appProtocol=http here as well

keymone (changes requested):

paasta_tools/setup_istio_mesh.py
Buffer0x7cd: The reason to include the ports is that they will be needed to be set on the paasta-routing service, With this mapping we keeping the collection of ports that need to be exposed on paasta routing service

paasta_tools/setup_istio_mesh.py
Buffer0x7cd: For app protocol we need to upgrade the library , since appProtocol field is new. I will upgrade the library and add the fields in that commit

paasta_tools/setup_istio_mesh.py
keymone: sure, but `smartstack_namespaces` already contains ports so we don't need to rebuild the same hashmap

keymone (approved):
fix tests and ship so we can start working in parallel

paasta_tools/setup_istio_mesh.py
```diff
@@ -0,0 +1,245 @@
...
+
+        if service in existing_kube_services_names:
+            log.info(f"Service {service} alredy exists, skipping")
+            continue
+
+        log.info(f"Creating {service} because it does not exist yet.")
+
+        service_labels: KubeSvcLabels = {
+            "paasta.yelp.com/owner": "istio",
+            "paasta.yelp.com/unified_service": "false",
```
keymone, line 181: i don't see where we use this label, isn't matching by service name good enough? we can assume that all services created under paasta namespace are part of our mesh setup, nothing else should be creating services in here.

paasta_tools/setup_istio_mesh.py
```diff
@@ -0,0 +1,245 @@
...
+        }
+        service_meta = k8s.V1ObjectMeta(name=service, labels=service_labels)
+        service_spec = k8s.V1ServiceSpec(
+            selector={f"{PAATA_REGISTRATION_PREFIX}/{namespace}": "true"},
```
keymone, line 185: we should reuse/extend existing code in kubernetes_tools to prefix things correctly

paasta_tools/utils.py
```diff
@@ -2785,6 +2785,17 @@ def decompose_job_id(job_id: str, spacer: str = SPACER) -> Tuple[str, str, str,
...
 
+def validate_registration_name(service_instance: str) -> bool:
+    try:
+        service, instance, _, __ = decompose_job_id(service_instance)
+    except InvalidJobNameError:
+        log.error(
+            "Invalid service instance specified. Format is service%sinstance." % SPACER
+        )
+        return False
+    return True
```
keymone, line 2796: this was moved but now we're not using it anymore? if we don';t need it for this PR lets revert this change.

paasta_tools/setup_istio_mesh.py
```diff
@@ -0,0 +1,245 @@
...
+            name=f"p{port}",
+            port=port,
+            protocol="TCP",
+            target_port=PAASTA_SVC_PORT,
+            app_protocol="http",
+        )
+        ports.append(port_spec)
+
+    service_labels: KubeSvcLabels = {
+        "paasta.yelp.com/owner": "istio",
```
keymone, line 146: i assume this attribute will be used to garbage-collect services, but `owner` is too generic and vague for this purpose, let's use something specific like `managed_by=setup_istio_mesh.py` and put it in annotations to not pollute labels (we don't really need to index on that label as we're assuming that every service in paasta namespace is managed by this script)

keymone (changes requested):
probably last bunch of comments, i'll address them and push to this branch

keymone (approved):
lets ship and continue i with rest of the features in parallel

PR #3180, opened 2021-08-17
PAASTA-17447: Fix paasta logs to match correct instance name

jfongatyelp:
This is a minor fix for `paasta logs`.

Our various callbacks to apply filtering on a scribe stream expect `instances` to be an `Iterable[str]`, assuming it to be a list of valid instances we want to filter.  

However, `paasta_logs` arguments expect only a single instance to be passed, and this single instance name is passed as `instances`.  Because a string is still an `Iterable[str]`, this is accepted.

This unfortunately means any instance name that is a substring of the actual requested instance name will be included in log output, since we are validating `parsed_line.get("instance") in instances` (because e.g. `'autoscaler' in 'batch_autoscaler'` is `True`)

This corrects the issue by passing a list with a single element rather than passing the string.


nemacysts (approved):
thoughts on also updating the callback typing to take in a `Collection[str]` so that we don't accidentally do this again?

jfongatyelp:
> thoughts on also updating the callback typing to take in a `Collection[str]` so that we don't accidentally do this again?

😞  a str also appears to match a Collection of str:
```
>>> isinstance('abc', Collection)
True
```

nemacysts:
> > thoughts on also updating the callback typing to take in a `Collection[str]` so that we don't accidentally do this again?
> 
> 😞 a str also appears to match a Collection of str:
> 
> ```
> >>> isinstance('abc', Collection)
> True
> ```

rip - i guess we could type the callbacks as only taking a List[str] or a Union of the collection types that we're expecting instead then

jfongatyelp:
> i guess we could type the callbacks as only taking a List[str] or a Union of the collection types that we're expecting instead then

I've gone ahead and forced us to `List[str]` here since it seemed the only references to this code were in `paasta_logs` and we manually specify this as a list.

nemacysts (approved):

lanceleih (approved):

PR #3181, opened 2021-08-18
Fix typo in validation error message

nemacysts:
As-is, the error message is not particularly clear as to what is wrong.

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -861,7 +861,7 @@ class InstanceConfig:
             if deploy_group not in pipeline_deploy_groups:
                 return (
                     False,
-                    f"{self.service}.{self.instance} uses deploy_group {deploy_group}, but it is not deploy.yaml",
+                    f"{self.service}.{self.instance} uses deploy_group {deploy_group}, but {deploy_group} is not deployed to in deploy.yaml",
                 )  # noqa: E501
         return True, ""
 


jfongatyelp (approved):

paasta_tools/utils.py
```diff
@@ -861,7 +861,7 @@ def check_deploy_group(self) -> Tuple[bool, str]:
             if deploy_group not in pipeline_deploy_groups:
                 return (
                     False,
-                    f"{self.service}.{self.instance} uses deploy_group {deploy_group}, but it is not deploy.yaml",
+                    f"{self.service}.{self.instance} uses deploy_group {deploy_group}, but {deploy_group} is not deployed to in deploy.yaml",
```
jfongatyelp, line 864: Actually I missed this secondary typo: s/deployed to in/deployed in/ ?

This line is just cursed

nemacysts: this line is indeed cursed - I think this is correct (we're telling users that we're not deploying TO deploy group IN deploy.yaml), but it still doesn't read/flow very well

maybe `but {deploy_group} is not referenced in deploy.yaml`?

jfongatyelp: Ohhh i get it now, I can haz reading comprehension.  But agreed the second option is more obvious.

mattmb:
Just merging this rather than closing @nemacysts since it definitely reads better than the original line!

PR #3182, opened 2021-08-18
Ping authors during bounce when new pods are unhealthy.

EvanKrall:
PAASTA-17416

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -32,6 +32,7 @@ from collections import defaultdict
 from threading import Event
 from threading import Thread
 from typing import Any
+from typing import Callable
 from typing import Collection
 from typing import DefaultDict
 from typing import Dict
@@ -55,7 +56,9 @@ from paasta_tools import remote_git
 from paasta_tools.api import client
 from paasta_tools.cassandracluster_tools import CassandraClusterDeploymentConfig
 from paasta_tools.cli.cmds.push_to_registry import is_docker_image_already_in_registry
+from paasta_tools.cli.cmds.status import get_main_container
 from paasta_tools.cli.cmds.status import get_version_table_entry
+from paasta_tools.cli.cmds.status import recent_container_restart
 from paasta_tools.cli.utils import get_jenkins_build_output_url
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_deploy_groups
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1305,6 +1321,76 @@ def diagnose_why_instance_is_stuck(
...
+
+def maybe_ping_for_unhealthy_pods(
+    service: str,
+    instance: str,
+    cluster: str,
+    git_sha: str,
+    status: InstanceStatusKubernetesV2,
+    notify_fn: Optional[Callable[[str], None]] = None,
+) -> None:
+    if notify_fn and not already_pinged:
```
kawaiwanyelp, line 1346: nit: I think this check should be done by the caller, not inside the function.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1305,6 +1321,76 @@ def diagnose_why_instance_is_stuck(
             print(f"    {line}")
     print("")
 
+    if should_ping_for_unhealthy_pods:
+        maybe_ping_for_unhealthy_pods(
+            service, instance, cluster, git_sha, status, notify_fn
+        )
+
+
+already_pinged = False
```
kawaiwanyelp, line 1335: nit: I'm not sure I like this global state even though it's technically okay since this is a command. Ideally, the runner is a class with `should_ping` as instance state.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1305,6 +1321,76 @@ def diagnose_why_instance_is_stuck(
...
+        }.get(reason, f"restarted ({reason})")
+
+        tip = {
+            "Error": (
+                "This may indicate a bug in your code, a misconfiguration in yelpsoa-configs, or missing srv-configs. "
+                f"Take a look at the output of your unhealthy pods with `paasta status -s {service} -i {instance} -c {cluster} -vv` (more -v for more output.)"
+            ),
+            "CrashLoopBackOff": "This may indicate a bug in your code, a misconfiguration in yelpsoa-configs, or missing srv-configs.",
+            "OOMKilled": "This probably means your new version of code requires more memory than the old version. You may want to increase memory in yelpsoa-configs or roll back.",
+        }.get(reason, "")
```
kawaiwanyelp, line 1393: We might want to suggest asking for help for paasta owners, since we're otherwise not providing debugging help.

kawaiwanyelp (changes requested):
Still needs test, though logic looks mostly good.

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: Yeah this definitely feels a little bad to me, but I didn't feel like refactoring these functions into methods. I'll see how difficult this refactor would be.

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: I pulled the `notify_fn` check up to the caller.

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: I guess I could pull this into its own OnlyPingOnce class or something, but refactoring this so that the `already_pinged` variable lives on the MarkForDeploymentProcess instance seems not worth it. I think I'm okay with this small bit of global state.

PR #3183, opened 2021-08-18
Send node selectors and affinities to tron

nemacysts:
Just like we do for normal paasta services, we only expose a
`node_selectors` key and then translate that into the correct format
internally. However, unlike paasta services, we don't turn these into
the corresponding k8s representation since this is meant to be POSTed to
the Tron API.

I refactored some of the existing selector/affinity code for k8s
services in order to be able to reuse them as I didn't want to have to
duplicate those functions just so that we could use them for Tron.

### DIFF
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -97,6 +97,90 @@
                 "pool": {
                     "type": "string"
                 },
+                "node_selectors": {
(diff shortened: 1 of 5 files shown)


nemacysts:
This needs https://github.com/Yelp/Tron/pull/837 to be merged before we release this

jfongatyelp (approved):

lanceleih (approved):

PR #3185, opened 2021-08-19
Allow services to register in smartstack/envoy with a null proxy_port…

EvanKrall:
… (so they are only accessible via the envoy shared port.)

This would mean that kubernetes services that have `proxy_port: null` in smartstack.yaml get the `hacheck` container and a routable IP address, etc.

https://jira.yelpcorp.com/browse/PAASTA-17454

### DIFF
--- a/paasta_tools/long_running_service_tools.py
+++ b/paasta_tools/long_running_service_tools.py
@@ -118,10 +118,7 @@ class ServiceNamespaceConfig(dict):
         return self.get("discover", "region")
 
     def is_in_smartstack(self) -> bool:
-        if self.get("proxy_port") is not None:
-            return True
-        else:
-            return False
+        return "proxy_port" in self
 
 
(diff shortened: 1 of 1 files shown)


EvanKrall:
context: https://yelp.slack.com/archives/C01U4J35A6L/p1629326762003400

connorworley (approved):

kawaiwanyelp (approved):

paasta_tools/long_running_service_tools.py
```diff
@@ -118,7 +118,7 @@ def get_discover(self) -> str:
         return self.get("discover", "region")
 
     def is_in_smartstack(self) -> bool:
-        if self.get("proxy_port") is not None:
+        if "proxy_port" in self:
             return True
         else:
             return False
```
keymone, line 124: minor nit:

    return "proxy_port" in self

keymone (approved):

PR #3187, opened 2021-08-20
Increase default polling interval to 60 seconds

TheNavigat:
The 5 seconds interval massively increases the load on the etcd nodes. The change introduced here: https://github.com/Yelp/paasta/commit/de82504539b7f45cf2319da92fac400166ef1d70 reduces the interval from 60 to 5 seconds. This PR reverses that change to reduce the load on etcd during y-m and other service bounces.


nemacysts (approved):

mattmb:
We aren't actually 100% sure that this is the only thing overloading etcd, but timings from this mornings push line up so it's an easy thing to test. We might be able to reduce what paasta status retrieves from k8s and re-lower this time in the future

EvanKrall (approved):

PR #3190, opened 2021-08-31
U/msingh/mesh 1151 add service creation

Buffer0x7cd:
Fixed NoneType AttributeError while accessing annotations
Added setup_istio_mesh in setup.py and created link in paasta-tools.links

### DIFF
--- a/paasta_tools/setup_istio_mesh.py
+++ b/paasta_tools/setup_istio_mesh.py
@@ -103,7 +103,9 @@ def get_existing_kubernetes_service_names(kube_client: KubeClient) -> Set[str]:
     return {
         item.metadata.name
         for item in service_objects.items
-        if item.annotations.get(paasta_prefixed("managed_by")) == "setup_istio_mesh"
+        if item.metadata.annotations
+        if item.metadata.annotations.get(paasta_prefixed("managed_by"))
+        == "setup_istio_mesh"
     }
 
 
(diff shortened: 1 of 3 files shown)


paasta_tools/setup_istio_mesh.py
```diff
@@ -103,7 +103,9 @@ def get_existing_kubernetes_service_names(kube_client: KubeClient) -> Set[str]:
     return {
         item.metadata.name
         for item in service_objects.items
-        if item.annotations.get(paasta_prefixed("managed_by")) == "setup_istio_mesh"
+        if item.metadata.annotations is not None
```
Rob-Johnson, line 106: `if item.metadata.annotations` rather than just a `None` check?

paasta_tools/setup_istio_mesh.py
```diff
@@ -103,7 +103,9 @@ def get_existing_kubernetes_service_names(kube_client: KubeClient) -> Set[str]:
...
+        if item.metadata.annotations.get(paasta_prefixed("managed_by"))
```
Rob-Johnson, line 107: `and if`?

Buffer0x7cd: it's a nested condition not to avoid raising `NoneType` attribute error

paasta_tools/setup_istio_mesh.py
Buffer0x7cd: fixed

Rob-Johnson (approved):

PR #3191, opened 2021-08-31
Send PaaSTA Workload Contract labels to Tron

nemacysts:
As part of the PaaSTA Workload Contract, we should be setting these
labels on Pods that we launch so that our tooling (e.g., OPA) knows what
to do with them.


nemacysts:
We'll want to wait until https://github.com/Yelp/Tron/pull/841 is merged to merge this (or just break tronplayground temporarily :p)

lanceleih (approved):

PR #3192, opened 2021-09-01
If an SLO is failing when the deploy completes, explain why the deploy isn't doing anything. PAASTA-17453

EvanKrall:
https://jira.yelpcorp.com/browse/PAASTA-17453

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -1095,7 +1095,11 @@ class MarkForDeploymentProcess(SLOSlackDeploymentProcess):
         line = f"Deployment of {self.commit[:8]} for {self.deploy_group} complete"
         _log(service=self.service, component="deploy", line=line, level="event")
         self.send_manual_rollback_instructions()
-        if not (self.any_slo_failing() and self.auto_rollbacks_enabled()):
+        if self.any_slo_failing() and self.auto_rollbacks_enabled():
+            self.ping_authors(
+                "Because an SLO is currently failing, we will not automatically certify. Instead, we will wait indefinitely until you click one of the buttons above."
+            )
+        else:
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1095,7 +1095,11 @@ def on_enter_deployed(self) -> None:
         line = f"Deployment of {self.commit[:8]} for {self.deploy_group} complete"
         _log(service=self.service, component="deploy", line=line, level="event")
         self.send_manual_rollback_instructions()
-        if not (self.any_slo_failing() and self.auto_rollbacks_enabled()):
+        if self.any_slo_failing() and self.auto_rollbacks_enabled():
+            self.ping_authors(
+                "Because an SLO is currently failing, we will not automatically certify. Instead, we will wait indefinitely until you click one of the buttons above."
```
nemacysts, line 1100: is this actually indefinitely or do we have a timeout elsewhere that would kill this process? this might be the releng in me, but it seems a little wasteful to hold onto a CI executor forever (well, really until something in jenkins goes sideways or the cluster scales down and that agent gets caught in the fray and is terminated)

that said, if this is actually gonna wait forever, it's probably fine to ship this first and add a timeout later if adding a timeout is non-trivial here

nemacysts (approved):

cuza (approved):

paasta_tools/cli/cmds/mark_for_deployment.py
EvanKrall: There's a timeout (3 hours default but overridable in deploy.yaml). I guess I could mention it here.

nemacysts: ah cool - I wasn't sure if that timeout applied here or not. 

up to you if you want to mention it!

PR #3193, opened 2021-09-02
[setup_istio_mesh] walk soa_dir when loading namespaces in setup_istio_mesh

keymone:
services.yaml doesn't include namespaces without proxy_port,
those namespaces are still routable via header and sockets so
we need to set them up in istio

### DIFF
--- a/paasta_tools/setup_istio_mesh.py
+++ b/paasta_tools/setup_istio_mesh.py
@@ -197,7 +197,11 @@ def setup_kube_services(
         try:
             setup_unified_service(
                 kube_client=kube_client,
-                port_list=sorted(val["proxy_port"] for val in namespaces.values()),
+                port_list=sorted(
+                    val["proxy_port"]
+                    for val in namespaces.values()
+                    if val.get("proxy_port")
+                ),
             )
         except Exception as err:
             log.error(f"{err} while setting up unified service")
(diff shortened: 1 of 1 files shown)


paasta_tools/setup_istio_mesh.py
```diff
@@ -67,21 +69,35 @@ def parse_args() -> argparse.Namespace:
...
     return args
 
 
-def load_smartstack_namespaces(file_path: str) -> Mapping:
-    with open(file_path) as service_file:
-        return json.load(service_file)
+def load_smartstack_namespaces(soa_dir: str = DEFAULT_SOA_DIR) -> Mapping:
+    namespaces = {}
+
+    _, dirs, _ = next(os.walk(soa_dir))
```
Buffer0x7cd, line 87: Not important for now, but in future maybe want to check how long it takes us iterate over the whole yelpsoa-config since it's quite big, but for now this seems like a good enough choice.

Buffer0x7cd (approved):

paasta_tools/setup_istio_mesh.py
Rob-Johnson: yeah, I'm also concerned about this being slow (can you add a timer or something to get some understanding of how long this takes?)- can't we read these concurrently?

keymone: this line doesn't iterate entire tree, only first level of hierarchy. takes 150 milliseconds. it only opens single smartstack.yaml file per service and most often that file will already be in cache.

PR #3195, opened 2021-09-02
U/msingh/mesh 1151 garbage collect istio svc

Buffer0x7cd:
This PR adds support for removing k8s services that no longer exists in smartstack

### DIFF
--- a/paasta_tools/setup_istio_mesh.py
+++ b/paasta_tools/setup_istio_mesh.py
@@ -144,12 +144,12 @@ def setup_unified_service(kube_client: KubeClient, port_list: List) -> Iterator:
 def setup_paasta_namespace_services(
     kube_client: KubeClient,
     paasta_namespaces: AbstractSet,
-    existing_kube_services_names: Set[str] = set(),
+    existing_namespace_services: Set[str] = set(),
 ) -> Iterator:
     for namespace in paasta_namespaces:
         service = sanitise_kubernetes_service_name(namespace)
 
-        if service in existing_kube_services_names:
+        if service in existing_namespace_services:
             log.debug(f"Service {service} alredy exists, skipping")
             continue
 
@@ -169,13 +169,30 @@ def setup_paasta_namespace_services(
         )
 
 
-def setup_kube_services(
-    kube_client: KubeClient, soa_dir: str = DEFAULT_SOA_DIR,
+def cleanup_paasta_namespace_services(
+    kube_client: KubeClient,
+    paasta_namespaces: AbstractSet,
+    existing_namespace_services: Set[str],
 ) -> Iterator:
-    existing_kube_services_names = get_existing_kubernetes_service_names(kube_client)
+    declared_services = {
+        sanitise_kubernetes_service_name(ns) for ns in paasta_namespaces
+    }
+    for service in existing_namespace_services:
+        if service == UNIFIED_K8S_SVC_NAME or service in declared_services:
+            continue
+        log.info(
+            f"Garbage collecting {service} since there is no reference in services.yaml"
+        )
+        yield kube_client.core.delete_namespaced_service, (service, PAASTA_NAMESPACE)
+
+
+def process_kube_services(
+    kube_client: KubeClient, soa_dir: str = DEFAULT_SOA_DIR
+) -> Iterator:
+    existing_namespace_services = get_existing_kubernetes_service_names(kube_client)
     namespaces = load_smartstack_namespaces(soa_dir)
 
-    if UNIFIED_K8S_SVC_NAME not in existing_kube_services_names:
+    if UNIFIED_K8S_SVC_NAME not in existing_namespace_services:
         log.info(f"Creating {UNIFIED_K8S_SVC_NAME} because it does not exist yet.")
         yield from setup_unified_service(
             kube_client=kube_client,
@@ -187,7 +204,11 @@ def setup_kube_services(
         )
 
     yield from setup_paasta_namespace_services(
-        kube_client, namespaces.keys(), existing_kube_services_names
+        kube_client, namespaces.keys(), existing_namespace_services
+    )
+
+    yield from cleanup_paasta_namespace_services(
+        kube_client, namespaces.keys(), existing_namespace_services
     )
 
 
@@ -203,7 +224,9 @@ def main() -> None:
     ensure_namespace(kube_client, namespace="paasta")
     delay = 0 if args.rate_limit == 0 else 1.0 / float(args.rate_limit)
     success = True
-    for (fn, args) in setup_kube_services(kube_client, args.soa_dir):
+    for (fn, args) in process_kube_services(
+        kube_client=kube_client, soa_dir=args.soa_dir
+    ):
         time.sleep(delay)
         try:
             fn(*args)
--- a/tests/test_setup_istio_mesh.py
+++ b/tests/test_setup_istio_mesh.py
@@ -1,10 +1,13 @@
 import mock
 
 from paasta_tools.kubernetes_tools import registration_label
+from paasta_tools.setup_istio_mesh import cleanup_paasta_namespace_services
 from paasta_tools.setup_istio_mesh import sanitise_kubernetes_service_name
 from paasta_tools.setup_istio_mesh import setup_paasta_namespace_services
 from paasta_tools.setup_istio_mesh import setup_unified_service
+from paasta_tools.setup_istio_mesh import UNIFIED_K8S_SVC_NAME
 from paasta_tools.setup_istio_mesh import UNIFIED_SVC_PORT
+
 
 MOCK_PORT_NUMBER = 20508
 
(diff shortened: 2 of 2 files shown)


paasta_tools/setup_istio_mesh.py
```diff
@@ -172,6 +172,37 @@ def setup_paasta_namespace_services(
     return status
 
 
+def cleanup_kube_svc(
+    kube_client: KubeClient,
+    smartstack_namespaces: AbstractSet,
```
Rob-Johnson, line 177: why isn't this just `Set[str]`?

paasta_tools/setup_istio_mesh.py
```diff
@@ -172,6 +172,37 @@ def setup_paasta_namespace_services(
...
+    sanitized_smarstack_namespaces = {
+        sanitise_kubernetes_service_name(namespace)
+        for namespace in smartstack_namespaces
+    }
+
+    status = True
+    if not existing_svc_names:
+        return status
+
+    existing_svc_names = existing_svc_names.difference(sanitized_smarstack_namespaces)
```
Rob-Johnson, line 189: can you call this `difference` rather than overwriting the existing variable?

paasta_tools/setup_istio_mesh.py
```diff
@@ -172,6 +172,37 @@ def setup_paasta_namespace_services(
...
+
+    for svc in existing_svc_names:
+        if svc != UNIFIED_K8S_SVC_NAME:
+            try:
+                kube_client.core.delete_namespaced_service(
+                    name=svc, namespace=PAASTA_NAMESPACE
+                )
+                log.debug(
```
Rob-Johnson, line 197: I think it's probably useful to add this as a `log.info` since we'll want to see which svc objects are being deleted by the script.

please also add this above the delete call above, so we can see the intent even if it fails

paasta_tools/setup_istio_mesh.py
```diff
@@ -172,6 +172,37 @@ def setup_paasta_namespace_services(
...
+                    f"Garbage collecting {svc} since there is no reference in services.yaml"
+                )
+            except Exception as err:
+                log.warning(f"{err} while trying to grabage collect {svc}")
```
Rob-Johnson, line 201: please add more info here - log the exception thrown, etc.

also, s/grabage/garbage

paasta_tools/setup_istio_mesh.py
```diff
@@ -192,7 +223,7 @@ def setup_kube_services(
 
     return setup_paasta_namespace_services(
         kube_client, namespaces.keys(), existing_kube_services_names, rate_limit
-    )
+    ) and cleanup_kube_svc(kube_client, namespaces.keys(), existing_kube_services_names)
```
Rob-Johnson, line 226: rather than just returning `and` can you do something like

```
setup_services = etup_paasta_namespace_services(
        kube_client, namespaces.keys(), existing_kube_services_names, rate_limit
)
delete_services = cleanup_kube_svc(kube_client, namespaces.keys(), existing_kube_services_names)
if not setup_services:
   log.info("failed to setup kubernetes svc objects")
else:
   log.info("succesfully deleted kubernetes svc objects")

if not delete_services:
  ...
else:
  ...
```

that way it'll be clearer to the incident responder what's gone on when it has failed

Rob-Johnson: also, maybe you should move the call to `cleanup_kube_svc` to the `main` function, rather than overloading `setup_`

Buffer0x7cd: The idea behind returning `and` was that , we have already logged the error case in form of exceptions from calls to `setup_paasta_namespace_service` or `cleanup_kube_svc` ( example  https://github.com/Yelp/paasta/blob/340344c3765df36de69468b8bfc4a5d273b2c3bb/paasta_tools/setup_istio_mesh.py#L169 ). At this point we can fail with any non zero error code ( So that we can mark the cron as failure and trigger the desired results).

Buffer0x7cd: Here we are using `setup_kube_services` as an orchestrator , Since we needed information processed in earlier steps to get re-used ( like list of know namespaces  , getting svc etc). Maybe we need to rename it to something better ( like `process_services`, Still i think it would be preferable to keep the flow of `setup_kube_services`.

paasta_tools/setup_istio_mesh.py
Buffer0x7cd: https://github.com/Yelp/paasta/blob/340344c3765df36de69468b8bfc4a5d273b2c3bb/paasta_tools/setup_istio_mesh.py#L225 .keys() returns a `AbstractSet` object, which cause mypy tests to fail in case of `Set[str]`

Rob-Johnson: right - if you walk back to `load_smartstack_namespaces` you see we just return a `Mapping` rather than a more constrained type.

paasta_tools/setup_istio_mesh.py
Rob-Johnson: right, yeah - `setup_kube_services` is a bit of a misnomer. maybe `ensure_kube_services` or something, but it's minor. I'll drop this one

Rob-Johnson (approved):

paasta_tools/setup_istio_mesh.py
Buffer0x7cd: yeah, `load_smartstack_namespaces` returns a map of type 


```
{
  service_name: 
    {
      'port':  'port-number'
    }
}

```
To get the list of service names, we use `.keys()` method on the return value of `load_smartstack_namespaces` , which in turn returns an `AbstractSet object`

paasta_tools/setup_istio_mesh.py
```diff
@@ -172,7 +173,38 @@ def setup_paasta_namespace_services(
...
+    if not existing_svc_names:
+        return status
+
+    for svc in existing_svc_names.difference(sanitized_smarstack_namespaces):
+        if svc != UNIFIED_K8S_SVC_NAME:
+            try:
+                log.info(
+                    f"Garbage collecting {svc} since there is no reference in services.yaml"
+                )
+                kube_client.core.delete_namespaced_service(
```
keymone, line 212: this should count towards rate limit

paasta_tools/setup_istio_mesh.py
```diff
@@ -188,11 +220,12 @@ def setup_kube_services(
             )
         except Exception as err:
             log.error(f"{err} while setting up unified service")
+            traceback.print_exc()
             return False
 
     return setup_paasta_namespace_services(
         kube_client, namespaces.keys(), existing_kube_services_names, rate_limit
-    )
+    ) and cleanup_kube_svc(kube_client, namespaces.keys(), existing_kube_services_names)
```
keymone, line 247: let's not clump this all into one line, i'd rather `return True` in the end and call setup_ and cleanup_ on separate lines

keymone (approved):
address rate limit and ship

paasta_tools/setup_istio_mesh.py
Buffer0x7cd: Do you mean something like this ?
```
    setup_paasta_namespace_services_staus = setup_paasta_namespace_services(
        kube_client, namespaces.keys(), existing_kube_services_names, rate_limit
    )

    cleanup_kube_svc_status = cleanup_kube_svc(kube_client, namespaces.keys(), existing_kube_services_names)

    return setup_paasta_namespace_services_staus and cleanup_kube_svc_status
```

keymone: something like that

paasta_tools/setup_istio_mesh.py
```diff
@@ -240,13 +240,15 @@ def process_kube_services(
             traceback.print_exc()
             return False
 
-    setup_paasta_namespace_services_status = setup_paasta_namespace_services(
+    setup_paasta_namespace_services_staus = setup_paasta_namespace_services(
```
keymone, line 243: fix typo

paasta_tools/setup_istio_mesh.py
```diff
@@ -185,11 +187,52 @@ def setup_paasta_namespace_services(
...
+                f"Garbage collecting {svc} since there is no reference in services.yaml"
+            )
+            kube_client.core.delete_namespaced_service(
+                name=svc, namespace=PAASTA_NAMESPACE
+            )
+            api_updates += 1
+
+        except Exception as err:
+            log.warning(f"{err} while trying to garbage collect {svc}")
+            traceback.print_exc()
```
nemacysts, line 229: super-nit: this could also be written as 
```suggestion
        except Exception:
            log.exception(f"Error encountered while trying to garbage collect {svc}")
```
so that we don't need to manually print out the exception

paasta_tools/setup_istio_mesh.py
```diff
@@ -262,8 +279,10 @@ def main() -> None:
     kube_client = KubeClient()
     ensure_namespace(kube_client, namespace="paasta")
 
+    global RATE_LIMIT
+    RATE_LIMIT = args.rate_limit
```
nemacysts, line 286: just curious: how come we're using a global here rather than plumbing `args.rate_limit` through `process_kube_services` and then to all the other rate-limited functions?

paasta_tools/setup_istio_mesh.py
Buffer0x7cd: Thanks, this is quite nit

paasta_tools/setup_istio_mesh.py
Buffer0x7cd: Since rate_limit is a read-only variable and passing around rate_limit and api_updates to each function quickly became messy. Initially I thought about making rate_limit global but we have dropped that idea. https://github.com/Yelp/paasta/pull/3201 Have details on how we are planning to handle rate limiting

PR #3197, opened 2021-09-02
Monk 2524 add kubernetes status for monkrelayclusters

geokous:
Adding functionality to paasta to propagate the yelpsoa-configs values to the operator. Currently, this is a basic version that does not hold any information about kubernetes in the way the cassandra operator works

### DIFF
--- a/paasta_tools/monkrelaycluster_tools.py
+++ b/paasta_tools/monkrelaycluster_tools.py
@@ -57,7 +57,7 @@ class MonkRelayClusterDeploymentConfig(LongRunningServiceConfig):
     def get_instances(self, with_limit: bool = True) -> int:
         return self.config_dict.get("replicas", 1)
 
-    def validate(self, params: Optional[List[str]] = None,) -> List[str]:
+    def validate(self, params: List[str] = None,) -> List[str]:
         # Use InstanceConfig to validate shared config keys like cpus and mem
         # TODO: add mem back to this list once we fix PAASTA-15582 and
         # move to using the same units as flink/marathon etc.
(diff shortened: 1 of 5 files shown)


paasta_tools/utils.py
```diff
@@ -130,6 +130,7 @@
     "flink",
     "cassandracluster",
     "kafkacluster",
+    "monkrealycluster",
```
geokous, line 133: @zhaoyanh1202 mispell

paasta_tools/monkrelaycluster_tools.py
```diff
@@ -0,0 +1,143 @@
...
+            config_dict=config_dict,
+            branch_dict=branch_dict,
+        )
+
+    def get_instances(self, with_limit: bool = True) -> int:
+        return self.config_dict.get("replicas", 1)
+
+    def validate(
+        self,
+        params: List[str] = [
```
Rob-Johnson, line 62: you shouldn't use mutable values as defaults in python.

https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

Rob-Johnson (approved):
fix and ship

paasta_tools/monkrelaycluster_tools.py
```diff
@@ -0,0 +1,142 @@
...
+        soa_dir=soa_dir,
+    )
+
+
+# TODO: read this from CRD in service configs
+def cr_id(service: str, instance: str) -> Mapping[str, str]:
+    return dict(
+        group="yelp.com",
+        version="v1alpha1",
+        namespace="paasta-monkrelayclusters",
```
zhaoyanh1202, line 139: I don't understand this config very well, but should this be consistent with what [we defined in the yelpsoa-config](https://sourcegraph.yelpcorp.com/sysgit/yelpsoa-configs@9b66e9e7a98e5a97f9fa2049287de3c415fefff3/-/blob/monk-relay-operator/kubernetes-infrastage.yaml#L13:29)?

PR #3201, opened 2021-09-06
[setup_istio_mesh] rate limit arg limits number of write calls to k8s per second

keymone:


Buffer0x7cd (approved):

keymone:
@TheNavigat you may want to look at 5e7defc and b14fec5. i'm changing registration label names as they need to be limited in size.

paasta_tools/kubernetes_tools.py
```diff
@@ -499,6 +515,13 @@ def raw_selectors_to_requirements(
     return requirements
 
 
+def registration_label(namespace: str) -> str:
+    """ Returns namespace prefixed with regs.{paasta_prefix}/,
+        truncated to 63 characters with hash as suffix if needed.
+    """
+    return limit_size_with_hash(f"regs.{PAASTA_ATTRIBUTE_PREFIX}{namespace}")
```
TheNavigat, line 522: You should replace this with `return(f"regs.{PAASTA_ATTRIBUTE_PREFIX}{limit_size_with_hash(namespace)}")

The 63 character limit is on the suffix part of the label, not the entire length. The entire length can be up to 253 characters in total. Check https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set for that

TheNavigat (changes requested):

Rob-Johnson (approved):

PR #3202, opened 2021-09-07
Use k8s_namespace in PromQL filters

nemacysts:
Alertmanager on EKS has a conflict with this label (and we can't change
what label it looks at), so we'll need to use a different label name
here.

We'll need to add some relabeling configs to make this label exist
internally (while also having `namespace` in place) before shipping
this.

Once this is fully deployed, we'll probably get rid of the relabeling
config and go down to just having a single label with the namespace:
k8s_namespace


nemacysts:
we ended up not needing this thanks to some additional prometheus magic :)

PR #3203, opened 2021-09-09
Auto generate pod template for pod affinity with weights specified by the user [COREML-2557]

chelspea:
- Auto generate pod affinity with weights specified by the user
- Pod affinity in this case is preferred but not guaranteed, the higher number user specify the tighter bin packing will be
- Pod affinity dependent on spark.kubernetes.allocation.batch.size, the smaller spark.kubernetes.allocation.batch.size is the tighter bin packing will be
testing done with 
```paasta spark-run -a --cluster-manager kubernetes --cluster pnw-devc --pool stable_batch --cmd ' spark-submit /code/virtualenv_run/bin/spark_etl_runner.py --feature-config /code/batch/campaign_info_daily/job_configs/bidding_history_daily.yaml --env-config-path /nail/srv/configs/ad_bidding_transform.yaml --team-name ad_market_dynamics --notify-email gcoll+ad_market_dynamics+batch@yelp.com --start-date 2021-08-08 --end-date 2021-08-08 --publish-path s3a://yelp-sorted-logs-us-west-2/gcoll/ad_bidding_transform/ -v ' --spark-args ' spark.cores.max=60 spark.executor.memory=13g spark.executor.cores=2 spark.driver.memory=13g spark.driver.cores=2 spark.jars.ivy=/tmp/.ivy spark.executor.instances=30' --aws-profile prod --service ad_bidding_transform
```
All available nodes
Screenshot 2021-09-09 at 10 27 45
Nodes used in the job:
Screenshot 2021-09-09 at 10 27 34

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -6,6 +6,7 @@ import re
 import shlex
 import socket
 import sys
+import yaml
 from typing import Any
 from typing import Dict
 from typing import List
@@ -110,7 +111,13 @@ def add_subparser(subparsers):
         "--image",
         help="Use the provided image to start the Spark driver and executors.",
     )
-
+    list_parser.add_argument(
+        "-a",
+        "--autogen",
+        help="Auto generate pod template used in executors.",
+        action="store_true",
+        default=False,
+    )
     list_parser.add_argument(
         "--docker-registry",
         help="Docker registry to push the Spark image built.",
@@ -514,7 +521,7 @@ def run_docker_container(
     return 0
 
 
-def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
+def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]], autogen=False) -> str:
     """Use submitted batch name as default spark_run job name"""
     docker_cmds = (
         shlex.split(original_docker_cmd)
@@ -536,8 +543,39 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
 
     if spark_app_name is None:
         spark_app_name = "paasta_spark_run"
-
     spark_app_name += f"_{get_username()}"
+    if autogen:
+        document = """
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    spark: exec
+spec:
+  affinity:
+    podAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+      - weight: 95
+        podAffinityTerm:
+          labelSelector:
+            matchExpressions:
+            - key: spark
+              operator: In
+              values:
+              - exec
+          topologyKey: topology.kubernetes.io/hostname
+"""
+        parsed_pod_template = yaml.load(document)
+        print(parsed_pod_template)
+        parsed_pod_template['metadata']['labels']['spark'] += "-" + spark_app_name
+        parsed_pod_template['spec']['affinity'] \
+            ['podAffinity']['preferredDuringSchedulingIgnoredDuringExecution'][0] \
+            ['podAffinityTerm']['labelSelector']['matchExpressions'][0]['values'][0] \
+            = parsed_pod_template['metadata']['labels']['spark']
+        print(get_username())
+        f = open("/nail/home/"+get_username()+"/podTemplate.yaml", "w")
+        f.write(yaml.dump(parsed_pod_template))
+        f.close()
     return spark_app_name
 
 
@@ -808,7 +846,7 @@ def paasta_spark_run(args):
         return 1
 
     volumes = instance_config.get_volumes(system_paasta_config.get_volumes())
-    app_base_name = get_spark_app_name(args.cmd or instance_config.get_cmd())
+    app_base_name = get_spark_app_name(args.cmd or instance_config.get_cmd(), args.autogen)
     needs_docker_cfg = not args.build
     user_spark_opts = _parse_user_spark_args(args.spark_args)
     paasta_instance = get_smart_paasta_instance_name(args)
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -58,7 +58,26 @@ CLUSTER_MANAGERS = {CLUSTER_MANAGER_MESOS, CLUSTER_MANAGER_K8S}
 
 POD_TEMPLATE_PATH = "/nail/tmp/podTemplate.yaml"
 BATCH_SIZE = "2" # Smaller batch size reduces number of nodes initial request is spread across
-
+POD_TEMPLATE = """
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    spark: exec-{spark_app_name}
+spec:
+  affinity:
+    podAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+      - weight: 95
+        podAffinityTerm:
+          labelSelector:
+            matchExpressions:
+            - key: spark
+              operator: In
+              values:
+              - exec-{spark_app_name}
+          topologyKey: topology.kubernetes.io/hostname
+"""
 
 deprecated_opts = {
     "j": "spark.jars",
@@ -114,6 +133,7 @@ def add_subparser(subparsers):
         "--image",
         help="Use the provided image to start the Spark driver and executors.",
     )
+
     list_parser.add_argument(
         "-e",
         "--enable-k8s-autogen",
@@ -552,32 +572,15 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]], enable_k
 
     if spark_app_name is None:
         spark_app_name = "paasta_spark_run"
-    spark_app_name += f"_{get_username()}"
+
+    spark_app_name += "_{}".format(get_username())
     if enable_k8s_autogen:
-        document = f"""
-apiVersion: v1
-kind: Pod
-metadata:
-  labels:
-    spark: exec-{spark_app_name}
-spec:
-  affinity:
-    podAffinity:
-      preferredDuringSchedulingIgnoredDuringExecution:
-      - weight: 95
-        podAffinityTerm:
-          labelSelector:
-            matchExpressions:
-            - key: spark
-              operator: In
-              values:
-              - exec-{spark_app_name}
-          topologyKey: topology.kubernetes.io/hostname
-"""
+        document = POD_TEMPLATE.format(spark_app_name=spark_app_name)
         parsed_pod_template = yaml.load(document)
-        f = open("/nail/tmp/podTemplate.yaml", "w")
+        f = open(POD_TEMPLATE_PATH, "w")
         f.write(yaml.dump(parsed_pod_template))
         f.close()
+
     return spark_app_name
 
 
@@ -629,7 +632,6 @@ def configure_and_run_docker_container(
     volumes.append("/nail/home:/nail/home:rw")
     if args.enable_k8s_autogen:
         volumes.append(f"{POD_TEMPLATE_PATH}:{POD_TEMPLATE_PATH}:rw")
-
 
     environment = instance_config.get_env_dictionary()  # type: ignore
     spark_conf_str = create_spark_config_str(spark_conf, is_mrjob=args.mrjob)
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -847,6 +847,11 @@ def paasta_spark_run(args):
 
     volumes = instance_config.get_volumes(system_paasta_config.get_volumes())
     app_base_name = get_spark_app_name(args.cmd or instance_config.get_cmd(), args.autogen)
+
+    if args.autogen:
+        pod_template_path = "/nail/home/"+get_username()+"/podTemplate.yaml"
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -536,8 +542,37 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
 
     if spark_app_name is None:
         spark_app_name = "paasta_spark_run"
-
     spark_app_name += f"_{get_username()}"
+    if autogen != '-1':
+        document = f"""
```
gcollura, line 547: I would move this at the top of the module, as a constant. It seems also that you're programmatically changing some values inside the string, I would instead use `.format()` function

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -536,8 +542,37 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
...
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    spark: exec
+spec:
+  affinity:
+    podAffinity:
+      preferredDuringSchedulingIgnoredDuringExecution:
+      - weight: {autogen}
```
gcollura, line 557: Is the command line value passed to spark-run also the value for the weight? It doesn't seem to be part of the flag description. I would prefer using a boolean flag + number flag eg. `--enable-k8s-autogen --k8s-affinity-weight` or something along the line. If possible the weight should have a default value that's somewhat sensible as well.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -536,8 +542,37 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
...
+              - exec
+          topologyKey: topology.kubernetes.io/hostname
+"""
+        parsed_pod_template = yaml.load(document)
+        parsed_pod_template['metadata']['labels']['spark'] += "-" + spark_app_name
+        parsed_pod_template['spec']['affinity'] \
+            ['podAffinity']['preferredDuringSchedulingIgnoredDuringExecution'][0] \
+            ['podAffinityTerm']['labelSelector']['matchExpressions'][0]['values'][0] \
+            = parsed_pod_template['metadata']['labels']['spark']
+        f = open("/nail/home/"+get_username()+"/podTemplate.yaml", "w")
```
gcollura, line 573: Generating a file with a hardcoded path in the user home path might introduce issues with concurrent user runs or with docker containers that use a different `$USER` than a human, eg. service users or `nobody`.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -110,7 +111,12 @@ def add_subparser(subparsers):
         "--image",
         help="Use the provided image to start the Spark driver and executors.",
     )
-
+    list_parser.add_argument(
+        "-a",
+        "--autogen",
+        help="Auto generate pod template used in executors.",
+        default='-1',
```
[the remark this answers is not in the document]
gcollura: This comment is sort of superseded by the other comment below, so feel free to ignore it :)

paasta_tools/cli/cmds/spark_run.py
chelspea: yeah as it stands it's `--auto-gen 95` to supply both the weight and the automatically generate the file, because otherwise it's just a superlong argument someone to supply

chelspea: cool I'll change the default, and also put a better description on `--auto-gen`

paasta_tools/cli/cmds/spark_run.py
chelspea: Do wonder where would be the best to write to, need to be somewhere both paasta command can access and sensible within the container itself. Guess might need to do a switch case here because service user would have permission to write elsewhere but human only have permission in to `write in nail/home/get_username`

gcollura: `/nail/tmp` should always be accessible, from both services and humans and I think also always mounted as well.

gcollura: `/tmp` should also work? https://github.com/Yelp/service_configuration_lib/blob/master/service_configuration_lib/spark_config.py#L455

chelspea: good idea! it's always good to dig these things out you forget about them :P

chelspea: think /tmp had security restrictions y/docker-restrictions, so used nail/tmp

paasta_tools/cli/cmds/spark_run.py
chelspea: made this into a f string now

chelspea: also updated the description

paasta_tools/cli/cmds/spark_run.py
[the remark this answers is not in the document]
88manpreet: +1

paasta_tools/cli/cmds/spark_run.py
88manpreet: I think by default we would want to enable autogen (with some default optimal weight) as we would want to prefer pod affinity in all cases.
Makes migration simpler and less arguments to play around or learn in future.  Thoughts?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -110,7 +111,12 @@ def add_subparser(subparsers):
         "--image",
         help="Use the provided image to start the Spark driver and executors.",
     )
-
+    list_parser.add_argument(
+        "-a",
+        "--autogen",
```
88manpreet, line 116: Looks like autogen is an integer? If so, add the param type, `type=int`.
`autogen` seems to be used as a weight, which has range from (1, 100). [Ref.](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/).

Use `choice(1, 100)` too, in that case. Ref.: https://docs.python.org/3/library/argparse.html#choices

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -110,7 +111,12 @@ def add_subparser(subparsers):
...
+        help="Auto generate pod template with weighted pod affinity used in executors.",
```
88manpreet, line 117: Description needs to be more descriptive in terms of helping the tool-user to decide a decent value of autogen from the argument itself.
Example: What would happen if autogen is set to 1 vs autogen is set to 100?

88manpreet: I might be missing something here. 
From the description here: 
```The weight field in preferredDuringSchedulingIgnoredDuringExecution is in the range 1-100. For each node that meets all of the scheduling requirements (resource request, RequiredDuringScheduling affinity expressions, etc.), the scheduler will compute a sum by iterating through the elements of this field and adding "weight" to the sum if the node matches the corresponding MatchExpressions. This score is then combined with the scores of other priority functions for the node. The node(s) with the highest total score are the most preferred.
```, 
it seems like weight field is relevant to make the right choice b/w given set of nodes meeting the criteria of affinity.
AIU, the meeting criteria in our case is the `label` (or app name). And every pod would have same weight (autogen) value.
Would the value of weight matter in that case?

File: paasta_tools/cli/cmds/spark_run.py [hunk 7]
```diff
@@ -808,7 +841,17 @@ def paasta_spark_run(args):
...
+    app_base_name = get_spark_app_name(args.cmd or instance_config.get_cmd(), args.autogen)
+
+    if args.autogen != '-1':
+        pod_template_path = "/nail/tmp/podTemplate.yaml"
+        if args.spark_args:
+            args.spark_args += f" spark.kubernetes.executor.podTemplateFile={pod_template_path}"
+            args.spark_args += f" spark.kubernetes.allocation.batch.size=2"
+        else:
+            args.spark_args = f"spark.kubernetes.executor.podTemplateFile={pod_template_path}"
+            args.spark_args += f" spark.kubernetes.allocation.batch.size=2"
```
88manpreet, line 853: Reason for setting this to 2? 
Any values set to constants need to defined as constant at the top. 
This helps in easier debugging and understanding the code, to go through all the constants from the top.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -808,7 +841,17 @@ def paasta_spark_run(args):
         return 1
 
     volumes = instance_config.get_volumes(system_paasta_config.get_volumes())
-    app_base_name = get_spark_app_name(args.cmd or instance_config.get_cmd())
+    app_base_name = get_spark_app_name(args.cmd or instance_config.get_cmd(), args.autogen)
+
+    if args.autogen != '-1':
+        pod_template_path = "/nail/tmp/podTemplate.yaml"
+        if args.spark_args:
```
88manpreet, line 848: I don't think we should append to args.spark_args as these are not the user given args.
It's better to append the these key/values after args.spark_args are parsed. 
Line: https://github.com/Yelp/paasta/pull/3203/files#diff-30f8e9c33224158fab1e77efc4004f7d2344230ea7b4ffbf64b3fca21513cc00R856
`    user_spark_opts = _parse_user_spark_args(args.spark_args)`

88manpreet:
Please add the test results in some form of tabular format in the description, comparing results when autogen is disabled and with given different values of batch size and autogen. 
An example, I can think of is:

* Current stable-batch pool: 20 nodes
CI can help in scaling-up the cluster to 20 nodes: Example thread: https://yelp.slack.com/archives/CA53J9UJ0/p1630526717165800

Scenario-1: a) Job requiring 20 executors
Scenario-2: a) Job requiring 200 executors

||Scenario||Expected Result||Actual Result||
|Disabled autogen|1 executor each execturo|?|Time to launch all pods in running state|
|Enabled autogen (EA), batch size (BS) 2, weight (W) 1|?|?|?|
|EA, BS:10, W:2|?|?|?|
|EA, BS:2, W: 100|?|?|?|
...
other unique scenarios

* Current stable-batch pool: 2 nodes (Current state)
a) Job requiring: 200 executors
This helps us to observe how does scaling happens

paasta_tools/cli/cmds/spark_run.py
chelspea: you are right, in this sense it would only make sense if the user was running multiple application of the same label with different value then the difference would make sense, I don't mind setting it to a default value in this case

paasta_tools/cli/cmds/spark_run.py
chelspea: yeah as above with default value make sense

paasta_tools/cli/cmds/spark_run.py
chelspea: smaller batch size tend to make pod affinity more pronounced

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -537,7 +579,14 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
     if spark_app_name is None:
         spark_app_name = "paasta_spark_run"
 
-    spark_app_name += f"_{get_username()}"
+    spark_app_name += "_{}".format(get_username())
```
gcollura, line 582: In general spark_app_name should already be set and it should already have the username, at least for user-run batches.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -537,7 +579,14 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
...
+    if enable_k8s_autogen:
+        document = POD_TEMPLATE.format(spark_app_name=spark_app_name)
+        parsed_pod_template = yaml.load(document)
+        f = open(POD_TEMPLATE_PATH, "w")
+        f.write(yaml.dump(parsed_pod_template))
+        f.close()
```
gcollura, line 588: nit: in Python `with open(filename, ...):` is preferred, so that you don't need to write `f.close()` afterwards (the context takes care of it. In a more Pythonic way:
```python
with open(POD_TEMPLATE_PATH, 'w') as f:
    yaml.dump(parsed_pod_template, f)
```

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -587,6 +636,8 @@ def configure_and_run_docker_container(
 
     volumes.append("%s:rw" % args.work_dir)
     volumes.append("/nail/home:/nail/home:rw")
+    if args.enable_k8s_autogen:
+        volumes.append(f"{POD_TEMPLATE_PATH}:{POD_TEMPLATE_PATH}:rw")
```
gcollura, line 640: Should this be `:ro` instead? Not that it makes a big difference, feel free to drop.

gcollura (approved):
A couple of smaller observation, but it looks good to me. Thank you!

paasta_tools/cli/cmds/spark_run.py
chelspea: yeah, this one i'll just leave it :P

paasta_tools/cli/cmds/spark_run.py
88manpreet: Any downsides of keeping the batch size this low (2)? (Example on performance)

PR #3204, opened 2021-09-09
Remove docs on regex-matching endpoint timeouts for now

thejoycekung:
As actually _using_ regex-matching causes many errors, removing this from the docs so that people don't accidentally use it.


ningweiii (approved):
ship it!

PR #3208, opened 2021-09-17
Ensure paasta.yelp.com/instance length is <=63 chars for tron

nemacysts:
This is a hardcoded k8s limit and tron instance names can easily be >63
characters in length since they're {job name}.{action name} and these
are often pretty descriptive names.


lanceleih (approved):
lgtm

PR #3210, opened 2021-09-20
Compact Bin packing, fix for pr #3203 auto generate pod template for users [COREML-2557] 

chelspea:
-------------------------------------------------------------
1. What is compact bin packing:
Effectively asks the k8s scheduler to run the pods on the same node as much as possible, though not the affinity is not guaranteed in case of not enough resources 
2. How to use this feature
This feature can be enabled by using the `-e` flag, it generates a pod template automatically which is used as the template for executors at run time
3. Pod template per run
Pod template is generated uniquely (uid in file name) for each run in the tmp folder whereever the job is started to guarantee the availability of the pod template file, tmp folders in our case are cleaned periodically so the files accumulated over time shouldn't be an issue.
--------------------------------------------------------------
Test results see comments on COREML-2537
--------------------------------------------------------------


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -458,9 +488,12 @@ def get_spark_env(
     return spark_env
 
 
-def _parse_user_spark_args(spark_args: Optional[str]) -> Dict[str, str]:
+def _parse_user_spark_args(
+    spark_args: Optional[str], enable_k8s_autogen: bool = True
```
nemacysts, line 492: should this default to the same value in https://github.com/Yelp/paasta/pull/3210/files#diff-30f8e9c33224158fab1e77efc4004f7d2344230ea7b4ffbf64b3fca21513cc00R142

chelspea: Yep, it doesn't actually impact the running codepath and was only defaulted for tests, but I'll switch it to false 👍

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -111,6 +134,13 @@ def add_subparser(subparsers):
         help="Use the provided image to start the Spark driver and executors.",
     )
 
+    list_parser.add_argument(
+        "-e",
+        "--enable-k8s-autogen",
```
88manpreet, line 139: The help section variable name and description should cater towards requirements of the tool users.
In this case, most of the paasta spark-run users are spark-batch implementors or core-ml (onpointers).

I think the name `enable-k8s-autogen` may be overloaded with keywords which they are either not familiar with or don't care. Example: autogen is for auto-generation of pod template and k8s for kubernetes.

Our mission is to hide the complexity of these underneath infra components as much as possible.
One way I tend to think is the functional aspect of the flag. 
In this case, I think the flag caters to `enable` or `disable` preferred launching of executors on same or different nodes (bin-packing vs round-robin).

1. Maybe something like, `--enable-bin-packing` or something else is more relevant name?

2.  It seems like by default we want to enable `bin-packing`.  Maybe in that case it should be `--disable-bin-packing`

Thoughts?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -111,6 +134,13 @@ def add_subparser(subparsers):
...
+        help="Auto generate pod template with weighted pod affinity used in executors.",
```
88manpreet, line 140: Maybe change this to something more relevant in the language which spark-users are familiar with.
Something like, "Enabling/Disabling this flag would enable round-robin. That means, executors would be spread out across nodes. By default, executors are preferred to be launched on the same node."?

88manpreet:
I think you mentioned the performance improvement of 15-20% with pod-affinity enabled. Can you add the details here providing the below context:
* Command and configuration of the test case
* Clearly showing the run-time for with and without pod-affinity enabled
* With low and high-executor count (to compare the behavior difference of 2 test cases)

88manpreet (changes requested):
Please add the most common test-cases with results which reviewers expect to see the difference.
1. Important configuration parameters and environment used most relevant to this review
2. Link of spark-history job (for any other difference comparison), if desired by reviewer
3. Run-time with and without the flag
4. Since, this had a special case of race-condition, the test case with different users running the batch job on the same host would be useful to have, I think?

paasta_tools/cli/cmds/spark_run.py
gcollura: An additional thing to check during exec time is that we generate this configuration only with `--cluster-manager=kubernetes`. This should also be mentioned in the `help` message as well, that with Mesos this flag has no effect.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -473,6 +506,10 @@ def _parse_user_spark_args(spark_args: Optional[str]) -> Dict[str, str]:
             )
             sys.exit(1)
         user_spark_opts[fields[0]] = fields[1]
+
+    if enable_k8s_autogen:
+        user_spark_opts["spark.kubernetes.executor.podTemplateFile"] = POD_TEMPLATE_PATH
```
gcollura, line 511: Make `podTemplateFile` name unique. It doesn't matter if we don't delete the file afterwards, it's only a few bytes. @nemacysts do you know how often `/nail/tmp` gets cleared on prod batch boxes?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -538,6 +577,12 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
         spark_app_name = "paasta_spark_run"
 
     spark_app_name += f"_{get_username()}"
+    if enable_k8s_autogen:
+        document = POD_TEMPLATE.format(spark_app_name=spark_app_name)
+        parsed_pod_template = yaml.load(document)
+        with open(POD_TEMPLATE_PATH, "w") as f:
+            yaml.dump(parsed_pod_template, f)
```
gcollura, line 584: Just a nit: do we want to keep the `yaml.load/yaml.dump` parsing, instead of writing to file directly, just to verify that the format is YAML-compatible?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -538,6 +577,12 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
         spark_app_name = "paasta_spark_run"
 
     spark_app_name += f"_{get_username()}"
```
gcollura, line 579: I think this was missed in the past review, what happens if the user is batch or service or nobody?

paasta_tools/cli/cmds/spark_run.py
nemacysts: not off the top of my head - i can check what tmpwatches we have setup on them in puppet tho

paasta_tools/cli/cmds/spark_run.py
chelspea: I think this is the way historically been done, see https://github.com/Yelp/paasta/blame/8b87e0b13816c09b3d063b6d3271e6c7627fd264/paasta_tools/cli/cmds/spark_run.py#L540, guess if it's service name or nobody that would be attached to the job as well

paasta_tools/cli/cmds/spark_run.py
chelspea: Yeah I think i'll make the file name something like spark_pt_{uid}, and if we ever want to clean up after we can always set up a seperate cronjob for it on batch box

paasta_tools/cli/cmds/spark_run.py
chelspea: I agree with the naming being overloaded, however bin packing itself is already enabled by default and what we're doing is creating a tighter bin packing with pod affinity. Maybe something like `enable-compact-bin-packing`, I'd also like this to be defaulted to false so we can roll it out in the background.

paasta_tools/cli/cmds/spark_run.py
nemacysts: whoops sorry - forgot to respond: we do have a generic tmpwatch that runs daily, but it will only clean things up that are older than 10 days

paasta_tools/cli/cmds/spark_run.py
chelspea: yeah I think keep the laod/dump is good to make sure that the file format is good

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -807,10 +857,27 @@ def paasta_spark_run(args):
     if docker_image is None:
         return 1
 
+    pod_template_path = generate_pod_template_path()
+
+    if args.enable_compact_bin_packing and (
+        not os.access(POD_TEMPLATE_DIR, os.R_OK)
+        or args.cluster_manager != CLUSTER_MANAGER_K8S
+    ):
+        args.enable_compact_bin_packing = False
```
gcollura, line 866: Could you please log or throw a warning in case `enable_compact_bin_packing` is enabled but the other checks are not passing? Especially if `os.access` is False-y.

chelspea: Thanks 👍 I've separated this into it's own method `should_enable_compact_bin_packing` and added tests for the different cases.

gcollura (approved):

paasta_tools/cli/cmds/spark_run.py
88manpreet: `enable-copact-bin-packing` sounds great. 

> ...I'd also like this to be defaulted to false so we can roll it out in the background.
Not sure I follow this part. Wouldn't we want it to enable by `default`, if we are rolling it out in the background?

During the migration, would every migration need to specify `--enable-compact-bin-packing`, if they need to keep the behavior as status-quo?

88manpreet:
Thanks for recreating the review and addressing the comments. Just have one comment left. lgtm otherwise.

paasta_tools/cli/cmds/spark_run.py
chelspea: I guess either way the scheduling behavior is probably a bit different from mesos, but the -e would have to be specified yeah, we can do a pr to add '-e' in all-repo-fixers if we decide to enable that for all jobs

88manpreet: From functionality perspective, looks like the default value of `-e` doesn't matter. 

1. What is going to be the more preferred way of using this flag `-e` by default in `adhoc` or `deployment mode`?
2. Which is more closer behavior of the value of `-e` flag with status-quo? `True` or `False`.
3. What are the downsides of keeping `-e` flag `True` by default?

What it looks like to me so far is:

* For 1., it seems `-e` is going to be set to `True` for all batches. 
Why do we want to add the additional overhead for migration and all users in `adhoc` mode, to set `-e True`. 
It seems`users` would need to adapt to the new flags for paasta spark-run.

* For 2. It seems to `True`. Unless I am missed something here?

chelspea: I guess since we are adding `--cluster-manager kubernetes` on a case by case basis, we can add the `-e` flag in the meantime. The biggest downsides usually lies in the fact that the scheduler will need to compute the weight of each node and prioritise accordingly which can be an expensive when there are a lot of scheduling requests. I think as a feature it's safer to roll out in the background instead of making it the default.

88manpreet: I take back what I said above.

I think in short-term it's ok to rollout default as `False`. 
The benefit of this is, the already live-launched batches are at `less risk`. 

On more thought, I agree when we come around `repo-fixers`, with `--cluster-manager kubernetes`, it is not a big-deal to `-e True`.

However, in long-term, we would definitely, plan to get rid of both the flags (--cluster-manager and -e) and set the preferred default values.

88manpreet: I apologize for initially not thinking through from the perspective of `short-term` or `long-term` approach. 
Resolving this comment.

PR #3216, opened 2021-09-24
Cleanup failed Pods in prune_completed_pods.py

nemacysts:
The current version of this script doesn't capture Pods that end up in
Error or CreateContainerError states, and thus these end up sticking
around for a while.

There's no direct way to get what Pods are in this state, but we can use
the Failed phase as a proxy.

For now, the new arg I've added is optional until we've rolled this
version out everywhere and have updated all callsites to pass in
--error-minutes.

I'm planning on setting --error-minutes to a somewhat high value in case
we want to debug this using kubectl (so probably a couple hours?)

### DIFF
--- a/paasta_tools/prune_completed_pods.py
+++ b/paasta_tools/prune_completed_pods.py
@@ -33,6 +33,14 @@ def parse_args():
         type=int,
     )
     parser.add_argument(
+        "-e",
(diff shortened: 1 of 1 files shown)


lanceleih (approved):
LGTM

PR #3219, opened 2021-10-01
Add ability to read the default use_k8s value from disk

nemacysts:
As we progress through our migration, we want to ensure that any new
Tronjobs on clusters that have already been fully migrated run on k8s by
default.

This PR will be accompanied by an internal Puppet PR that will provide
the configuration that this is looking at.


lanceleih (approved):
Nice! Looks good to me.

PR #3222, opened 2021-10-07
Add cprofiling to paasta api

kawaiwanyelp:
### Description
Add `yelp_profiling` so that we can profile paasta api requests. By default, profiling doesn't happen, so to turn it on, we need to add system paasta configs.

### Testing
manual testing on devbox
i didnt write unit tests because nearly 100% of the code i wrote just uses `yelp_profiling` code directly.

### DIFF
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -31,6 +31,7 @@ import paasta_tools.api
 from paasta_tools import kubernetes_tools
 from paasta_tools import marathon_tools
 from paasta_tools.api import settings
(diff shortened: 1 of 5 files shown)


kawaiwanyelp:
I fixed an import bug with the cprofile tweens. The bug was that the `PaastaCProfileConfig` class used `CProfileConfig` as a super class, which comes from `yelp_profiling`. That means if `yelp_profiling` wasn't installed and the cprofile tween module was import, there would be an import error. This is what caused the itests and example_cluster tests to fail.

PR #3228, opened 2021-10-25
Record prometheus yelpsoaconfig timeouts metrics

ningweiii:
The script loops through all yelpsoa folders and parses smartstack.yaml to get either `endpoint_timeouts` or `timeout_server_ms` value. Then we convert timeouts values into prometheus format using `prometheus_client` lib and save into the textfile with name ending with `.prom` which will be scraped by node exporter.

- If `endpoint_timeouts` is defined, we record per endpoint timeouts value. 
- If  `timeout_server_ms` is given, we record `timeout_server_ms` value as default path `/`
- Otherwise, we take default 1000ms  timeout_server_ms 

The metrics format as below:
```
# HELP endpoint timeout value defined in yelpsoa-config.
# TYPE TYPE yelpsoaconfig_endpoint_timeouts_ms gauge
yelpsoaconfig_endpoint_timeouts_ms{path="/consumer_app/devices/braze/info",upstream="push_notifications.main.egress_cluster"} 10000.0
yelpsoaconfig_endpoint_timeouts_ms{path="default",upstream="mysql_read_security.main.egress_cluster"} 100.0
yelpsoaconfig_endpoint_timeouts_ms{path="default",upstream="mysql_snapshot_yelpwifirealtime.main.egress_cluster"} 1000.0

```
Output:
https://fluffy.yelpcorp.com/i/ZXqTVRl84Jvsllmnwv1NsJr8XD9PkhRV.html

### DIFF
--- a/paasta_tools/contrib/timeouts_metrics_prom.py
+++ b/paasta_tools/contrib/timeouts_metrics_prom.py
@@ -1,0 +1,55 @@
+#!/opt/venvs/paasta-tools/bin/python
+"""
+The goal is to walk through all smartstack.yaml in yelpsoa directories
+and convert timeout values to the prometheus format.
+
+For every upstream and path, if endpoint timeout is specified, we record
+per endpoint timeout value. If not and if timeout_server_ms is specified,
+we record timeout_server_ms value in path `default`. Otherwise, we record the default
+timeout_server_ms 1000.
+
+# HELP endpoint timeout value defined in yelpsoa-config.
+# TYPE TYPE yelpsoaconfig_endpoint_timeouts_ms gauge
+yelpsoaconfig_endpoint_timeouts_ms{path="/consumer_app/devices/braze/info",upstream="push_notifications.main.egress_cluster"} 10000.0
+yelpsoaconfig_endpoint_timeouts_ms{path="default",upstream="mysql_read_security.main.egress_cluster"} 100.0
+"""
+import os
+
+import yaml
+from prometheus_client import CollectorRegistry
+from prometheus_client import Gauge
+from prometheus_client import write_to_textfile
+
+SOA_DIR = "/nail/etc/services/"
+PROM_OUTPUT_FILE = "/nail/etc/services/.autotune_timeouts.prom"
+
+
+registry = CollectorRegistry()
+prom_metric = Gauge(
+    "yelpsoaconfig_endpoint_timeouts_ms",
+    "endpoint timeout value defined in yelpsoa-config",
+    ["path", "upstream"],
+    registry=registry,
+)
+for root, dirs, files in os.walk(SOA_DIR):
+    service = root.split("/")[-1]
+    for name in files:
+        if name == "smartstack.yaml":
+            with open(os.path.join(root, name)) as smartstack_file:
+                smartstack_yaml = yaml.safe_load(smartstack_file)
+                for instance_name, info in smartstack_yaml.items():
+                    upstream = service + "." + instance_name + ".egress_cluster"
+                    if "endpoint_timeouts" in info:
+                        for path, endpoint_timeouts in info[
+                            "endpoint_timeouts"
+                        ].items():
+                            prom_metric.labels(path, upstream).set(endpoint_timeouts)
+                    else:
+                        if "timeout_server_ms" in info:
+                            prom_metric.labels("default", upstream).set(
+                                info["timeout_server_ms"]
+                            )
+                        else:
+                            prom_metric.labels("default", upstream).set(1000)
+
+write_to_textfile(PROM_OUTPUT_FILE, registry)
--- a/paasta_tools/contrib/timeouts_metrics_prom.py
+++ b/paasta_tools/contrib/timeouts_metrics_prom.py
@@ -24,32 +24,37 @@ SOA_DIR = "/nail/etc/services/"
 PROM_OUTPUT_FILE = "/nail/etc/services/.autotune_timeouts.prom"
 
 
-registry = CollectorRegistry()
-prom_metric = Gauge(
-    "yelpsoaconfig_endpoint_timeouts_ms",
-    "endpoint timeout value defined in yelpsoa-config",
-    ["path", "upstream"],
-    registry=registry,
-)
-for root, dirs, files in os.walk(SOA_DIR):
-    service = root.split("/")[-1]
-    for name in files:
-        if name == "smartstack.yaml":
-            with open(os.path.join(root, name)) as smartstack_file:
-                smartstack_yaml = yaml.safe_load(smartstack_file)
-                for instance_name, info in smartstack_yaml.items():
-                    upstream = service + "." + instance_name + ".egress_cluster"
-                    if "endpoint_timeouts" in info:
-                        for path, endpoint_timeouts in info[
-                            "endpoint_timeouts"
-                        ].items():
-                            prom_metric.labels(path, upstream).set(endpoint_timeouts)
-                    else:
-                        if "timeout_server_ms" in info:
-                            prom_metric.labels("default", upstream).set(
-                                info["timeout_server_ms"]
-                            )
-                        else:
-                            prom_metric.labels("default", upstream).set(1000)
+def read_and_write_timeouts_metrics(root, service, prom_metric):
+    with open(os.path.join(root, "smartstack.yaml")) as smartstack_file:
+        smartstack_yaml = yaml.safe_load(smartstack_file)
+    for instance_name, info in smartstack_yaml.items():
+        upstream = service + "." + instance_name + ".egress_cluster"
+        if "endpoint_timeouts" in info:
+            for path, endpoint_timeouts in info["endpoint_timeouts"].items():
+                prom_metric.labels(path, upstream).set(endpoint_timeouts)
+        else:
+            if "timeout_server_ms" in info:
+                prom_metric.labels("/", upstream).set(info["timeout_server_ms"])
+            else:
+                prom_metric.labels("/", upstream).set(1000)
 
-write_to_textfile(PROM_OUTPUT_FILE, registry)
+
+if __name__ == "__main__":
+    registry = CollectorRegistry()
+    prom_metric = Gauge(
+        "yelpsoaconfig_endpoint_timeouts_ms",
+        "endpoint timeout value defined in yelpsoa-config",
+        ["path", "upstream"],
+        registry=registry,
+    )
+    # Walk through soa config dir and filter smartstack yaml
+    for root, dirs, files in os.walk(SOA_DIR):
+        service = root.split("/")[-1]
+        # Avoid confusion of the smartstacks.yaml under autotuned_defaults/ in the future
+        if "autotuned_defaults" == files:
+            continue
+        for f in files:
+            if f == "smartstack.yaml":
+                read_and_write_timeouts_metrics(root, service, prom_metric)
+
+    write_to_textfile(PROM_OUTPUT_FILE, registry)
--- a/paasta_tools/contrib/timeouts_metrics_prom.py
+++ b/paasta_tools/contrib/timeouts_metrics_prom.py
@@ -51,7 +51,7 @@ if __name__ == "__main__":
     for root, dirs, files in os.walk(SOA_DIR):
         service = root.split("/")[-1]
         # Avoid confusion of the smartstacks.yaml under autotuned_defaults/ in the future
-        if "autotuned_defaults" == files:
+        if "autotuned_defaults" == service:
             continue
         for f in files:
             if f == "smartstack.yaml":
--- a/paasta_tools/contrib/timeouts_metrics_prom.py
+++ b/paasta_tools/contrib/timeouts_metrics_prom.py
@@ -29,8 +29,8 @@ def read_and_write_timeouts_metrics(root, service, prom_metric):
         smartstack_yaml = yaml.safe_load(smartstack_file)
     for instance_name, info in smartstack_yaml.items():
         upstream = service + "." + instance_name + ".egress_cluster"
-        if "endpoint_timeout" in info:
-            for path, endpoint_timeout in info["endpoint_timeout"].items():
+        if "endpoint_timeouts" in info:
+            for path, endpoint_timeout in info["endpoint_timeouts"].items():
                 prom_metric.labels(path, upstream).set(endpoint_timeout)
         # always record default timeout
         default_timeout = info.get("timeout_server_ms", 1000)
--- a/paasta_tools/contrib/timeouts_metrics_prom.py
+++ b/paasta_tools/contrib/timeouts_metrics_prom.py
@@ -4,14 +4,14 @@ The goal is to walk through all smartstack.yaml in yelpsoa directories
 and convert timeout values to the prometheus format.
 
 For every upstream and path, if endpoint timeout is specified, we record
-per endpoint timeout value. If not and if timeout_server_ms is specified,
-we record timeout_server_ms value in path `default`. Otherwise, we record the default
+per endpoint timeout value. If timeout_server_ms is specified,
+we record timeout_server_ms value in path `/`. Otherwise, we record the default
 timeout_server_ms 1000.
 
 # HELP endpoint timeout value defined in yelpsoa-config.
 # TYPE TYPE yelpsoaconfig_endpoint_timeouts_ms gauge
 yelpsoaconfig_endpoint_timeouts_ms{path="/consumer_app/devices/braze/info",upstream="push_notifications.main.egress_cluster"} 10000.0
-yelpsoaconfig_endpoint_timeouts_ms{path="default",upstream="mysql_read_security.main.egress_cluster"} 100.0
+yelpsoaconfig_endpoint_timeouts_ms{path="/",upstream="mysql_read_security.main.egress_cluster"} 100.0
 """
 import os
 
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,55 @@
...
+import yaml
+from prometheus_client import CollectorRegistry
+from prometheus_client import Gauge
+from prometheus_client import write_to_textfile
+
+SOA_DIR = "/nail/etc/services/"
+PROM_OUTPUT_FILE = "/nail/etc/services/.autotune_timeouts.prom"
+
+
+registry = CollectorRegistry()
```
matthewbentley, line 27: Can you put the bulk of this file (starting here) into a `main()` method, and then call it in a `if __name__ == "__main__":` block?  It's generally best practice to put the entrypoint behind a "if name equals main" check to make testing easier (in the future if we want).

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,55 @@
...
+prom_metric = Gauge(
+    "yelpsoaconfig_endpoint_timeouts_ms",
+    "endpoint timeout value defined in yelpsoa-config",
+    ["path", "upstream"],
+    registry=registry,
+)
+for root, dirs, files in os.walk(SOA_DIR):
+    service = root.split("/")[-1]
+    for name in files:
+        if name == "smartstack.yaml":
```
matthewbentley, line 37: I'm not sure if this is worth addressing now, but this will get confused if/when we start autotuning anything in smartstack.yaml.  An easy workaround might be to just ignore if the "service" name is "autotuned_defaults"?

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,55 @@
...
+            with open(os.path.join(root, name)) as smartstack_file:
+                smartstack_yaml = yaml.safe_load(smartstack_file)
+                for instance_name, info in smartstack_yaml.items():
```
matthewbentley, line 40: This is getting to be a lot of indentation...can you split some part of this into a separate function to help readability a bit?

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,55 @@
...
+                    upstream = service + "." + instance_name + ".egress_cluster"
+                    if "endpoint_timeouts" in info:
+                        for path, endpoint_timeouts in info[
+                            "endpoint_timeouts"
+                        ].items():
+                            prom_metric.labels(path, upstream).set(endpoint_timeouts)
+                    else:
+                        if "timeout_server_ms" in info:
+                            prom_metric.labels("default", upstream).set(
```
matthewbentley, line 49: I think it would make sense to use `"/"` instead of "default" here (and below), since logically the "default" timeout is equivalent to a per endpoint timeout for "/"

matthewbentley (commented):
A few small comments, but overall this is in the right direction.

paasta_tools/contrib/timeouts_metrics_prom.py
ningweiii: Filtered out autotuned_defaults

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,60 @@
...
+    for instance_name, info in smartstack_yaml.items():
+        upstream = service + "." + instance_name + ".egress_cluster"
+        if "endpoint_timeouts" in info:
+            for path, endpoint_timeouts in info["endpoint_timeouts"].items():
+                prom_metric.labels(path, upstream).set(endpoint_timeouts)
+        else:
+            if "timeout_server_ms" in info:
+                prom_metric.labels("/", upstream).set(info["timeout_server_ms"])
+            else:
+                prom_metric.labels("/", upstream).set(1000)
```
matthewbentley, line 39: this section could be:
```python3
default_timeout = info.get("timeout_server_ms", 1000)
prom_metric.labels("/", upstream).set(default_timeout)
```

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,60 @@
...
+
+def read_and_write_timeouts_metrics(root, service, prom_metric):
+    with open(os.path.join(root, "smartstack.yaml")) as smartstack_file:
+        smartstack_yaml = yaml.safe_load(smartstack_file)
+    for instance_name, info in smartstack_yaml.items():
+        upstream = service + "." + instance_name + ".egress_cluster"
+        if "endpoint_timeouts" in info:
+            for path, endpoint_timeouts in info["endpoint_timeouts"].items():
+                prom_metric.labels(path, upstream).set(endpoint_timeouts)
+        else:
```
matthewbentley, line 35: I think we don't want this in an `else:` block: we want to record the default/base timeout whether or not there are overrides

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,60 @@
...
+PROM_OUTPUT_FILE = "/nail/etc/services/.autotune_timeouts.prom"
+
+
+def read_and_write_timeouts_metrics(root, service, prom_metric):
+    with open(os.path.join(root, "smartstack.yaml")) as smartstack_file:
+        smartstack_yaml = yaml.safe_load(smartstack_file)
+    for instance_name, info in smartstack_yaml.items():
+        upstream = service + "." + instance_name + ".egress_cluster"
+        if "endpoint_timeouts" in info:
+            for path, endpoint_timeouts in info["endpoint_timeouts"].items():
```
matthewbentley, line 33: nit: `endpoint_timeouts` probably shouldn't be plural here?

paasta_tools/contrib/timeouts_metrics_prom.py
ningweiii: Uh can you explain why we want to always record the default/base timeout? My vision was to get current per endpoint timeout in grafana, I didn't see why we would need default timeout.

paasta_tools/contrib/timeouts_metrics_prom.py
ningweiii: Updated! English 😭

paasta_tools/contrib/timeouts_metrics_prom.py
ningweiii: Updated! Thanks for the suggestion!

paasta_tools/contrib/timeouts_metrics_prom.py
matthewbentley: the "default" timeout will apply to any path that isn't matched by a per-endpoint timeout.  So if a service has endpoints "/action1" and "/action2", and only a per-endpoint timeout for "/action2", then the default timeout will apply for "/action1"

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,59 @@
...
+def read_and_write_timeouts_metrics(root, service, prom_metric):
+    with open(os.path.join(root, "smartstack.yaml")) as smartstack_file:
+        smartstack_yaml = yaml.safe_load(smartstack_file)
+    for instance_name, info in smartstack_yaml.items():
+        upstream = service + "." + instance_name + ".egress_cluster"
+        if "endpoint_timeout" in info:
+            for path, endpoint_timeout in info["endpoint_timeout"].items():
+                prom_metric.labels(path, upstream).set(endpoint_timeout)
+        else:
+            if "timeout_server_ms" in info:
```
matthewbentley, line 36: You also don't need this `if` statement: the `info.get` on the line below will return `info["timeout_server_ms"]` if it exists, otherwise it will return 1000

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,59 @@
...
+PROM_OUTPUT_FILE = "/nail/etc/services/.autotune_timeouts.prom"
+
+
+def read_and_write_timeouts_metrics(root, service, prom_metric):
+    with open(os.path.join(root, "smartstack.yaml")) as smartstack_file:
+        smartstack_yaml = yaml.safe_load(smartstack_file)
+    for instance_name, info in smartstack_yaml.items():
+        upstream = service + "." + instance_name + ".egress_cluster"
+        if "endpoint_timeout" in info:
+            for path, endpoint_timeout in info["endpoint_timeout"].items():
```
thejoycekung, line 33: hmmm I'm just thinking about how currently a lot of the endpoint timeouts that are set in configs are more generic because regex-matching wasn't available -- and so some of the endpoint timeouts in configs don't necessarily match what we use from pulling the swagger docs. 

like for example in configs if a timeout is set for `/bubbles/` (fake example) that timeout also applies to everything under `/bubbles/` too right? whereas in swagger docs, the list of endpoints we might actually be working with is `/bubbles/v1/blossom/` and `/bubbles/v1/buttercup` or something.

i don't know if _here_ while scraping the metrics is the best place to make sure the lists match though? that might be really hard to do. the easier thing to do might just be to set different dashboard variables on grafana (one set for the swagger endpoints, one set for these config endpoints)

paasta_tools/contrib/timeouts_metrics_prom.py
ningweiii: you are right, and actually that was my intention but for some reasons my brain was fuzzy and the code was not followed up 😢

ningweiii: fixed that!

paasta_tools/contrib/timeouts_metrics_prom.py
ningweiii: fixed that! thanks!

paasta_tools/contrib/timeouts_metrics_prom.py
thejoycekung: I'm not _entirely_ sure whether the regex field in grafana accepts using other variables (I know the query field does, and it allows regex-ing those variables too). I don't have a method for testing this what we have right now, but it's not mentioned in the docs [docs link here](https://grafana.com/docs/grafana/latest/variables/filter-variables-with-regex/)

That being said I also realised we'd have to use another set of dashboard variables either way because the metrics are being scraped to a different shard, so maybe this is a non-issue.

ningweiii: I will look into the regex regex-ed variable query using current metrics. Hm I don't know if we need `another set of dashboard variable`, does it mean the current variable (path, upstream) we scrape from  perel won't be able to used in query in compute-infra... I suspect it is the case, I will double check on it...

ningweiii: ok I tried and regex definitely works with `path=~".*$regex_path.*|/"` where `regex_path` will be the `yelpsoaconfig_endpoint_timeouts_ms`'s path.
I simulate `regex_path` query variable with regexed`\^(?<path>\S+)/\?\$` of `label_values(upstream_path_status:service_timings_seconds_bucket:rate30s_sum{upstream=~"$upstream"}, path)`
https://grafana.yelpcorp.com/d/Ug6GreD7z/timeouts-playground-copy?orgId=1&var-upstream=ad_delivery_java.main.egress_cluster&var-path=%5E%2Fv1%2Fquery_understanding%2F%3F$&var-cluster=pnw&var-regex_cluster=nova.*&var-regex_cluster_query=pnw&var-regex_path=%2Fv1%2Fquery_understanding

thejoycekung: Ning and I talked it out, and we figured out we'll have three dashboard variables:
- upstream (currently there)
- regex_path (pulls from these metrics, hidden to the user)
- path (pulls from swagger metrics, will be filtered based on regex_path)

Resolving this conversation now that we've got that sorted :)

matthewbentley (approved):
LGTM, assuming you've tested this manually once.  Also you should get a :ship: from CI onpoint before shipping this change (and they can also help bump the paasta version)

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def read_and_write_timeouts_metrics(root, service, prom_metric):
+    with open(os.path.join(root, "smartstack.yaml")) as smartstack_file:
+        smartstack_yaml = yaml.safe_load(smartstack_file)
+    for instance_name, info in smartstack_yaml.items():
+        upstream = service + "." + instance_name + ".egress_cluster"
+        if "endpoint_timeouts" in info:
+            for path, endpoint_timeout in info["endpoint_timeouts"].items():
+                prom_metric.labels(path, upstream).set(endpoint_timeout)
```
nemacysts, line 34: super-nit to get rid of one level of indentation
```suggestion
        for path, endpoint_timeout in info.get("endpoint_timeouts", {}).items():
            prom_metric.labels(path, upstream).set(endpoint_timeout)
```

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,58 @@
...
+        # always record default timeout
+        default_timeout = info.get("timeout_server_ms", 1000)
```
nemacysts, line 36: nit: it'd be nice to make this `1000` a named constant

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,58 @@
...
+    )
+    # Walk through soa config dir and filter smartstack yaml
+    for root, dirs, files in os.walk(SOA_DIR):
+        service = root.split("/")[-1]
+        # Avoid confusion of the smartstacks.yaml under autotuned_defaults/ in the future
+        if "autotuned_defaults" == service:
+            continue
+        for f in files:
+            if f == "smartstack.yaml":
+                read_and_write_timeouts_metrics(root, service, prom_metric)
```
nemacysts, line 56: nit: we know what we're looking for, so we can use a set and save ourselves some string comparisons (alternatively, if we're not particularly worried about runtime complexity we could just do `if "smartstack.yaml" in files:` instead)
```suggestion
        if "smartstack.yaml" in set(files):
            read_and_write_timeouts_metrics(root, service, prom_metric)
```

nemacysts (approved):
i mostly just have some small nits, none of which are blocking if you don't think they're worth the effort

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,58 @@
...
+
+For every upstream and path, if endpoint timeout is specified, we record
+per endpoint timeout value. If timeout_server_ms is specified,
+we record timeout_server_ms value in path `/`. Otherwise, we record the default
+timeout_server_ms 1000.
+
+# HELP endpoint timeout value defined in yelpsoa-config.
+# TYPE TYPE yelpsoaconfig_endpoint_timeouts_ms gauge
+yelpsoaconfig_endpoint_timeouts_ms{path="/consumer_app/devices/braze/info",upstream="push_notifications.main.egress_cluster"} 10000.0
+yelpsoaconfig_endpoint_timeouts_ms{path="/",upstream="mysql_read_security.main.egress_cluster"} 100.0
```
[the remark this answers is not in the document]
ningweiii: great suggestion! added `Example output`

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,58 @@
...
+"""
+import os
+
+import yaml
+from prometheus_client import CollectorRegistry
+from prometheus_client import Gauge
+from prometheus_client import write_to_textfile
+
+SOA_DIR = "/nail/etc/services/"
```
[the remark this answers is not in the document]
ningweiii: Updated to use DEFAULT_SOA_DIR

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,58 @@
...
+PROM_OUTPUT_FILE = "/nail/etc/services/.autotune_timeouts.prom"
```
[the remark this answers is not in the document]
ningweiii: Updated to use DEFAULT_SOA_DIR

paasta_tools/contrib/timeouts_metrics_prom.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def read_and_write_timeouts_metrics(root, service, prom_metric):
+    with open(os.path.join(root, "smartstack.yaml")) as smartstack_file:
+        smartstack_yaml = yaml.safe_load(smartstack_file)
+    for instance_name, info in smartstack_yaml.items():
+        upstream = service + "." + instance_name + ".egress_cluster"
```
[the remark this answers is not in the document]
ningweiii: updated!

paasta_tools/contrib/timeouts_metrics_prom.py
ningweiii: didn't update due to the readability for myself 😭

paasta_tools/contrib/timeouts_metrics_prom.py
ningweiii: updated to use DEFAULT_TIMEOUT

paasta_tools/contrib/timeouts_metrics_prom.py
ningweiii: didn't update that for readability for myself 😭

PR #3232, opened 2021-10-28
PAASTA-17482 Getting per instance streams for clusterman

cuza:
## Problem or Feature
Currently, paasta logs are aggregated by service, which makes filtering expensive, especially considering the fact that the most common filtering we do is per-instance.

## Solution
One huge step to improving paasta logs performance is to collect logs per instance into its own streams. This PR will release a prototype of paasta that reads per instance streams for the clusterman service


mattmb:

PR #3233, opened 2021-11-02
Republish docs on regex-matching endpoint timeouts

thejoycekung:
Reverts Yelp/paasta#3204

Now that we're fairly confident adding regex-matched timeouts won't break anything, I'm adding this again.


cuza (approved):

cuza:
Extra neat pick: We use the [jsonschema](https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/schemas/smartstack_schema.json) to validate SOA Configs. You could [add a valid regex pattern](https://json-schema.org/understanding-json-schema/reference/regular_expressions.html) for your endpoint timeout to the smartstack schema.

thejoycekung:
I'll add a pre-commit hook to SOA configs instead -- we tried building out a regex, but it got real messy real quickly. Thanks for the tip though Dave!

PR #3234, opened 2021-11-04
PAASTA-17500: Handle more timeouts in new kubernetes status

jfongatyelp:
This change uses asyncio.gather(return_exceptions=True) so we can leave existing try/except logic where they are when handling individual task results, rather than having to add it when we gather the tasks.

In addition, this catches the somewhat new exception where we are running into a timeout when fetching pods.  As the majority of the status won't be useful if we can't fetch pods, this also adds an error message useful for the user when we hit this problem.

In addition, because we use `pods_for_service_instance` only in status and `force_delete_pods`, where we typically _need_ the full results, I've bumped the timeout from the default 10s to 15s to reduce how many people hit this, while we investigate api slowness.

Finally this standardizes how we're patching some of our kubernetes_tools utilities in tests for `paasta_tools.instance.kubernetes`.

Verified this still runs successfully when not hitting timeouts, and validated the error behavior when setting timeout artificially low: https://fluffy.yelpcorp.com/i/TdhrrFbQm6PDT1HXxkfk4bzhdMm18c6W.html

### DIFF
--- a/paasta_tools/instance/kubernetes.py
+++ b/paasta_tools/instance/kubernetes.py
@@ -609,7 +609,7 @@ async def kubernetes_status_v2(
         )
         tasks.extend([pod_status_by_replicaset_task, versions_task])
 
-    await asyncio.gather(*tasks)
+    await asyncio.gather(*tasks, return_exceptions=True)
 
     desired_state = job_config.get_desired_state()
     status["app_name"] = job_config.get_sanitised_deployment_name()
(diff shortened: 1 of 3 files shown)


paasta_tools/instance/kubernetes.py
```diff
@@ -618,20 +618,33 @@ async def kubernetes_status_v2(
         job_config.get_instances() if desired_state != "stop" else 0
     )
     status["bounce_method"] = job_config.get_bounce_method()
-    status["versions"] = versions_task.result()
+
+    try:
+        pods_task.result()  # just verifies we have a valid result
```
stug, line 623: so, to clarify, this is so that the pods_task will return its exception here rather than deeper down in the stack where we don't have the error handling, yeah?

stug (approved):
Thanks a bunch for looking into this!

PR #3235, opened 2021-11-08
add piscina support

kedarv:


paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -288,6 +304,96 @@ def create_instance_uwsgi_scaling_rule(
...
+        (sum(
+            k8s:deployment:pods_status_ready{{{worker_filter_terms}}} >= 0
+            or
+            max_over_time(
+                k8s:deployment:pods_status_ready{{{worker_filter_terms}}}[{DEFAULT_EXTRAPOLATION_TIME}s]
+            )
+        ) by (kube_deployment))
+    """
+    load_per_instance = f"""
+        (piscina_queue_size{{{worker_filter_terms}}} / piscina_max_queue_size{{{worker_filter_terms}}}) by (kube_pod, kube_deployment)
```
kedarv, line 356: unsure if this calculation is correct

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -288,6 +304,96 @@ def create_instance_uwsgi_scaling_rule(
...
+    """
+    metrics_query = f"""
+        {desired_instances} / {current_replicas}
+    """
+
+    metric_name = f"{deployment_name}-uwsgi-prom"
+
+    return {
+        "name": {"as": metric_name},
+        "seriesQuery": f"piscina_queue_load{{{worker_filter_terms}}}",
```
kedarv, line 391: unsure if this name is correct

paasta_tools/kubernetes_tools.py
```diff
@@ -964,6 +960,20 @@ def should_run_uwsgi_exporter_sidecar(
...
+        if self.is_autoscaling_enabled():
+            autoscaling_params = self.get_autoscaling_params()
+            if autoscaling_params["metrics_provider"] == "piscina":
+                if autoscaling_params.get(
+                    "use_prometheus",
+                    DEFAULT_USE_PROMETHEUS_PISCINA
+                    or system_paasta_config.default_should_setup_piscina_prometheus_scraping(),
+                ):
+                    return True
+        return False
```
nemacysts, line 975: imo, we can skip adding awareness of `use_prometheus` here - this was meant as a toggle while we were transitioning from sfx to prometheus for uwsgi-based autoscaling. 

since this is something completely new, we can get away with just doing something like
```suggestion
    def should_setup_piscina_prometheus_scraping(self) -> bool:
        if self.is_autoscaling_enabled():
            autoscaling_params = self.get_autoscaling_params()
        
        return False
```

and uh, eventually cleanup `use_prometheus` for uwsgi :p (I think the only place we'd want to leave use_prometheus is for moving average cpu utilization scaling, which we just need to get back to at some point and debug)

paasta_tools/kubernetes_tools.py
```diff
@@ -1679,6 +1689,9 @@ def get_pod_template_spec(
             # character limit for k8s labels (63 chars)
             labels["paasta.yelp.com/deploy_group"] = self.get_deploy_group()
 
+        if piscina_autoscaling_configured:
+           labels["paasta.yelp.com/scrape_piscina_prometheus"] = "true"
```
nemacysts, line 1693: can we also add the label from https://github.com/Yelp/paasta/pull/3235/files#diff-f1b278fb2675abadd2f58f131c3f67ece5d7511b60fcad691255aeda6f490d17R1690?

this really should be a default label, but we forgot to promote this and I don't think we want to cause a big bounce for code that we're just rolling out for the first time

paasta_tools/kubernetes_tools.py
```diff
@@ -1679,6 +1689,9 @@ def get_pod_template_spec(
             # character limit for k8s labels (63 chars)
             labels["paasta.yelp.com/deploy_group"] = self.get_deploy_group()
 
+        if piscina_autoscaling_configured:
```
nemacysts, line 1692: super-nit: imo (and feel free to drop if you disagree) switching to `elif`
```suggestion
        elif piscina_autoscaling_configured:
```
would be a little clearer here: only one of these conditionals should ever be true

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -196,6 +196,22 @@ def should_create_uwsgi_scaling_rule(
...
+    Determines whether we should configure the prometheus adapter for a given service.
+    Returns a 2-tuple of (should_create, reason_to_skip)
+    """
+    if autoscaling_config["metrics_provider"] == "piscina":
+        if not autoscaling_config.get("use_prometheus", DEFAULT_USE_PROMETHEUS_UWSGI):
+            return False, "requested piscina autoscaling, but not using Prometheus"
+
+        return True, None
+
+    return False, "did not request piscina autoscaling"
```
nemacysts, line 212: same as earlier in this PR - we can probably skip teaching this new code about `use_prometheus`
```suggestion
    if autoscaling_config["metrics_provider"] == "piscina":
        return True, None
    return False, "did not request piscina autoscaling"
```

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -288,6 +304,96 @@ def create_instance_uwsgi_scaling_rule(
...
+    instance: str,
+    autoscaling_config: AutoscalingParamsDict,
+    paasta_cluster: str,
+) -> PrometheusAdapterRule:
+    """
+    Creates a Prometheus adapter rule config for a given service instance.
+    """
+    setpoint = autoscaling_config["setpoint"]
+    moving_average_window = autoscaling_config.get(
+        "moving_average_window_seconds", DEFAULT_UWSGI_AUTOSCALING_MOVING_AVERAGE_WINDOW
```
nemacysts, line 318: imo, we should probably have a separate constant for the default piscina moving average window (even if it ends up being the same as the uwsgi one to start)

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -288,6 +304,96 @@ def create_instance_uwsgi_scaling_rule(
...
+    )
+    # this should always be set, but we default to 0 for safety as the worst thing that would happen
+    # is that we take a couple more iterations than required to hit the desired setpoint
+    offset = autoscaling_config.get("offset", 0)
```
nemacysts, line 322: do we end up using a piscina worker/task queue slot/whatever the correct terminology is here to serve the metrics that prometheus scrapes? if not, we can get rid of `offset` (this is meant to correct for a uwsgi worker being utilized to serve metrics (e.g., if you have 4 uwsgi workers - ~1 worker will always be shown as busy even when you're 100% idle since hitting the metrics endpoint requires a worker to serve that data))

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -288,6 +304,96 @@ def create_instance_uwsgi_scaling_rule(
...
+        )
+    """
+    metrics_query = f"""
+        {desired_instances} / {current_replicas}
+    """
+
+    metric_name = f"{deployment_name}-uwsgi-prom"
+
+    return {
+        "name": {"as": metric_name},
```
nemacysts, line 390:
```suggestion
    return {
        "name": {"as": f"{deployment_name}-piscina-prom"},
```

paasta_tools/setup_prometheus_adapter_config.py
nemacysts: this needs to be a query that will return data for the resources that will be autoscaled - so 
```suggestion
        "seriesQuery": f"piscina_queue_size{{{worker_filter_terms}}}",
```
or
```suggestion
        "seriesQuery": f"piscina_max_queue_size{{{worker_filter_terms}}}",
```
are probably good enough options (in case you're curious: https://github.com/kubernetes-sigs/prometheus-adapter/blob/master/docs/config.md if you want to read up on what this config is being used for in the end)

paasta_tools/utils.py
```diff
@@ -2584,6 +2584,9 @@ def get_uwsgi_exporter_sidecar_image_url(self) -> str:
     def default_should_run_uwsgi_exporter_sidecar(self) -> bool:
         return self.config_dict.get("default_should_run_uwsgi_exporter_sidecar", False)
 
+    def default_should_setup_piscina_prometheus_scraping(self) -> bool:
```
nemacysts, line 2587: reminder to also remove this function once the code calling it has been cleaned up

paasta_tools/kubernetes_tools.py
```diff
@@ -179,6 +179,7 @@
 
 DEFAULT_USE_PROMETHEUS_CPU = False
 DEFAULT_USE_PROMETHEUS_UWSGI = True
+DEFAULT_USE_PROMETHEUS_PISCINA = True
```
nemacysts, line 182: (there's a comment below talking about use_prometheus - remove this constant once the usage below is cleaned up)

nemacysts (commented):
nice - looking good!

paasta_tools/cli/cmds/validate.py
```diff
@@ -401,7 +401,7 @@ def validate_autoscaling_configs(service_path):
                 and instance_config.is_autoscaling_enabled()
             ):
                 autoscaling_params = instance_config.get_autoscaling_params()
-                if autoscaling_params["metrics_provider"] in {"uwsgi", "http"}:
+                if autoscaling_params["metrics_provider"] in {"uwsgi", "http", "piscina"}:
```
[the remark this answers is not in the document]
nemacysts: I made PAASTA-17513 internally for this

paasta_tools/setup_prometheus_adapter_config.py
nemacysts: we can test this code out in infrastage before shipping to make sure that everything is gucci :)

paasta_tools/kubernetes_tools.py
kedarv: We still need ` if autoscaling_params["metrics_provider"] == "piscina":` though right?

nemacysts: uh, yea - I'm just bad at writing code in a textfield ;_;
```suggestion
    def should_setup_piscina_prometheus_scraping(
        self, system_paasta_config: SystemPaastaConfig,
    ) -> bool:
        if self.is_autoscaling_enabled():
            autoscaling_params = self.get_autoscaling_params()
            return autoscaling_params["metrics_provider"] == "piscina":
        return False
```
is what i meant

paasta_tools/setup_prometheus_adapter_config.py
kedarv: Oh no, we are just using the main process in Fastify to serve the metrics. I'll remove it.

nemacysts (approved):
nice! before we merge this we should do some science in infrastage
(run a service with the correct piscina setup in infrastage -> manually label the resulting Pod -> make a deb from this branch-> install it on the kube masters there -> make sure the promql behaves as we want it to now that prometheus will actually scrape this service)

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -288,6 +303,92 @@ def create_instance_uwsgi_scaling_rule(
...
+        (sum(
+            k8s:deployment:pods_status_ready{{{worker_filter_terms}}} >= 0
+            or
+            max_over_time(
+                k8s:deployment:pods_status_ready{{{worker_filter_terms}}}[{DEFAULT_EXTRAPOLATION_TIME}s]
+            )
+        ) by (kube_deployment))
+    """
+    load_per_instance = f"""
+        (piscina_queue_size{{{worker_filter_terms}}} / piscina_max_queue_size{{{worker_filter_terms}}}) by (kube_deployment)
```
nemacysts, line 353:
```suggestion
        (piscina_queue_size{{{worker_filter_terms}}} / piscina_max_queue_size{{{worker_filter_terms}}})
```

nemacysts (dismissed):

nemacysts (approved):

PR #3238, opened 2021-11-12
Require explicit specification of cluster, service, and instance to r…

stug:
…un `paasta stop`

This caused an incident when a user did not specify a cluster or instance when paasta stoping a service.

See PAASTA-17411

---

Note that the said to do this for all of start, stop, and restart, but I can find places where people are using things like deploy groups in start/restart commands in tron jobs, so I figured I'd let that be -- seems not especially problematic to start/restart more than you meant to, anyway.  Open to argument there, though.

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -318,5 +318,16 @@ def paasta_start(args):
     return paasta_start_or_stop(args, "start")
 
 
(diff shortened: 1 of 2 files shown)


EvanKrall (commented):
Should we clean up https://github.com/Yelp/paasta/pull/3238/files#diff-60b8e1141d0ec6abfd7adeec81b725f37c250705b0c2bc1864ff3a0847328a5eR199-R217 ?

EvanKrall (approved):

PR #3241, opened 2021-11-18
Set driver container memory and cpu limits

88manpreet:
Setting the docker container memory limit to 2x of the spark driver limit.
Cpu limit I am currently setting the same as the spark.driver.cores.
If needs to be higher kindly let me know.
I have added the manual test results in the ticket COREML-2743
I will test one of the heavy batches in production once the code is reviewed.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -658,15 +658,14 @@ def _calculate_docker_memory_limit(
     2. --spark-args or spark-submit: spark.driver.memory
     3. Default
     """
+    if memory_limit:
+        return memory_limit
+
     try:
-        if memory_limit:
-            docker_memory_limit_str = memory_limit
-            adjustment_factor = 1
-        else:
-            docker_memory_limit_str = spark_conf.get(
-                "spark.driver.memory", DEFAULT_DRIVER_MEMORY_BY_SPARK
-            )
-            adjustment_factor = DOCKER_RESOURCE_ADJUSTMENT_FACTOR
+        docker_memory_limit_str = spark_conf.get(
+            "spark.driver.memory", DEFAULT_DRIVER_MEMORY_BY_SPARK
+        )
+        adjustment_factor = DOCKER_RESOURCE_ADJUSTMENT_FACTOR
         match = re.match(r"([0-9]+)([a-z]*)", docker_memory_limit_str)
         memory_val = int(match[1]) * adjustment_factor
         memory_unit = match[2]
@@ -710,26 +709,9 @@ def configure_and_run_docker_container(
 
     # driver specific volumes
     volumes: List[str] = []
-    try:
 
-        docker_memory_limit_str = spark_conf.get(
-            "spark.driver.memory", DEFAULT_DRIVER_MEMORY_BY_SPARK
-        )
-        match = re.match(r"([0-9]+)([a-z]*)", docker_memory_limit_str)
-        memory_val = int(match[1]) * DOCKER_RESOURCE_ADJUSTMENT_FACTOR
-        memory_unit = match[2]
-        docker_memory_limit = f"{memory_val}{memory_unit}"
-    except Exception as e:
-        # This case shouldn't arise, but if for any reason it fails, we continue with default value
-        log.error(
-            f"Failed to parse docker memory limit. Error: {e}. Example values: 1g, 200m."
-            " Setting a default value {DEFAULT_MAX_DOCKER_MEMORY_LIMIT}"
-        )
-        docker_memory_limit = DEFAULT_MAX_DOCKER_MEMORY_LIMIT
-
-    docker_cpu_limit = str(
-        int(spark_conf.get("spark.driver.cores", DEFAULT_DRIVER_CORES_BY_SPARK))
-        * DOCKER_RESOURCE_ADJUSTMENT_FACTOR
+    docker_memory_limit = _calculate_docker_memory_limit(
+        spark_conf, args.docker_memory_limit
     )
     docker_cpu_limit = _calculate_docker_cpu_limit(spark_conf, args.docker_cpu_limit,)
 
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -61,7 +61,6 @@ DEFAULT_DRIVER_CORES_BY_SPARK = 1
 DEFAULT_DRIVER_MEMORY_BY_SPARK = "1g"
 # Extra room for memory overhead and for any other running inside container
 DOCKER_RESOURCE_ADJUSTMENT_FACTOR = 2
-DEFAULT_MAX_DOCKER_MEMORY_LIMIT = "64g"
 
 POD_TEMPLATE_DIR = "/nail/tmp"
 POD_TEMPLATE_PATH = "/nail/tmp/spark-pt-{file_uuid}.yaml"
@@ -675,10 +674,9 @@ def _calculate_docker_memory_limit(
     except Exception as e:
         # For any reason it fails, continue with default value
         print(
-            f"Failed to parse docker memory limit. Error: {e}. Example values: 1g, 200m."
-            " Setting a default value {DEFAULT_MAX_DOCKER_MEMORY_LIMIT}"
+            f"ERROR: Failed to parse docker memory limit. Error: {e}. Example values: 1g, 200m."
         )
-        docker_memory_limit = DEFAULT_MAX_DOCKER_MEMORY_LIMIT
+        raise
 
     return docker_memory_limit
 
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -58,10 +58,10 @@ CLUSTER_MANAGER_K8S = "kubernetes"
 CLUSTER_MANAGERS = {CLUSTER_MANAGER_MESOS, CLUSTER_MANAGER_K8S}
 # Reference: https://spark.apache.org/docs/latest/configuration.html#application-properties
 DEFAULT_DRIVER_CORES_BY_SPARK = 1
-DEFAULT_DRIVER_MEMORY_BY_SPARK = '1g'
+DEFAULT_DRIVER_MEMORY_BY_SPARK = "1g"
 # Extra room for memory overhead and for any other running inside containe
 DOCKER_RESOURCE_ADJUSTMENT_FACTOR = 2
-DEFAULT_MAX_DOCKER_MEMORY_LIMIT='64g'
+DEFAULT_MAX_DOCKER_MEMORY_LIMIT = "64g"
 
 POD_TEMPLATE_DIR = "/nail/tmp"
 POD_TEMPLATE_PATH = "/nail/tmp/spark-pt-{file_uuid}.yaml"
@@ -382,8 +382,19 @@ def should_enable_compact_bin_packing(enable_compact_bin_packing, cluster_manage
     return True
 
 
-def get_docker_run_cmd(container_name, volumes, env, docker_img, docker_cmd, nvidia, docker_memory_limit, docker_cpu_limit):
-    log.info(f"Setting docker memory limits as {docker_memory_limit} and docker cpus limit as {docker_cpu_limit}")
+def get_docker_run_cmd(
+    container_name,
+    volumes,
+    env,
+    docker_img,
+    docker_cmd,
+    nvidia,
+    docker_memory_limit,
+    docker_cpu_limit,
+):
+    log.info(
+        f"Setting docker memory and cpu limits as {docker_memory_limit}, {docker_cpu_limit} respectively"
+    )
     cmd = ["paasta_docker_wrapper", "run"]
     cmd.append(f"--memory={docker_memory_limit}")
     cmd.append(f"--cpus={docker_cpu_limit}")
@@ -566,7 +577,15 @@ def create_spark_config_str(spark_config_dict, is_mrjob):
 
 
 def run_docker_container(
-    container_name, volumes, environment, docker_img, docker_cmd, dry_run, nvidia, docker_memory_limit, docker_cpu_limit,
+    container_name,
+    volumes,
+    environment,
+    docker_img,
+    docker_cmd,
+    dry_run,
+    nvidia,
+    docker_memory_limit,
+    docker_cpu_limit,
 ) -> int:
 
     docker_run_args = dict(
@@ -617,6 +636,30 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
     return spark_app_name
 
 
+def _calculate_docker_memory_cpu_limit(spark_conf):
+    try:
+
+        docker_memory_limit_str = spark_conf.get(
+            "spark.driver.memory", DEFAULT_DRIVER_MEMORY_BY_SPARK
+        )
+        match = re.match(r"([0-9]+)([a-z]*)", docker_memory_limit_str)
+        memory_val = int(match[1]) * DOCKER_RESOURCE_ADJUSTMENT_FACTOR
+        memory_unit = match[2]
+        docker_memory_limit = f"{memory_val}{memory_unit}"
+    except Exception as e:
+        # For any reason it fails, continue with default value
+        log.error(
+            f"Failed to parse docker memory limit. Error: {e}. Example values: 1g, 200m."
+            " Setting a default value {DEFAULT_MAX_DOCKER_MEMORY_LIMIT}"
+        )
+        docker_memory_limit = DEFAULT_MAX_DOCKER_MEMORY_LIMIT
+
+    docker_cpu_limit = str(
+        int(spark_conf.get("spark.driver.cores", DEFAULT_DRIVER_CORES_BY_SPARK))
+    )
+    return docker_memory_limit, docker_cpu_limit
+
+
 def configure_and_run_docker_container(
     args: argparse.Namespace,
     docker_img: str,
@@ -632,21 +675,24 @@ def configure_and_run_docker_container(
     volumes: List[str] = []
     try:
 
-        docker_memory_limit_str = spark_conf.get("spark.driver.memory", DEFAULT_DRIVER_MEMORY_BY_SPARK)
+        docker_memory_limit_str = spark_conf.get(
+            "spark.driver.memory", DEFAULT_DRIVER_MEMORY_BY_SPARK
+        )
         match = re.match(r"([0-9]+)([a-z]*)", docker_memory_limit_str)
         memory_val = int(match[1]) * DOCKER_RESOURCE_ADJUSTMENT_FACTOR
         memory_unit = match[2]
-        docker_memory_limit =  f"{memory_val}{memory_unit}"
+        docker_memory_limit = f"{memory_val}{memory_unit}"
     except Exception as e:
         # This case shouldn't arise, but if for any reason it fails, we continue with default value
         log.error(
-            f"Failed to parse docker memory limit. Error: {e}. Example values: 1g, 200m." \
+            f"Failed to parse docker memory limit. Error: {e}. Example values: 1g, 200m."
             " Setting a default value {DEFAULT_MAX_DOCKER_MEMORY_LIMIT}"
         )
         docker_memory_limit = DEFAULT_MAX_DOCKER_MEMORY_LIMIT
 
-    docker_cpu_limit = str(int(spark_conf.get("spark.driver.cores", DEFAULT_DRIVER_CORES_BY_SPARK)) * \
-        DOCKER_RESOURCE_ADJUSTMENT_FACTOR
+    docker_cpu_limit = str(
+        int(spark_conf.get("spark.driver.cores", DEFAULT_DRIVER_CORES_BY_SPARK))
+        * DOCKER_RESOURCE_ADJUSTMENT_FACTOR
     )
 
     if cluster_manager == CLUSTER_MANAGER_MESOS:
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -606,6 +636,30 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
...
+        # For any reason it fails, continue with default value
+        log.error(
+            f"Failed to parse docker memory limit. Error: {e}. Example values: 1g, 200m."
+            " Setting a default value {DEFAULT_MAX_DOCKER_MEMORY_LIMIT}"
+        )
+        docker_memory_limit = DEFAULT_MAX_DOCKER_MEMORY_LIMIT
+
+    docker_cpu_limit = str(
+        int(spark_conf.get("spark.driver.cores", DEFAULT_DRIVER_CORES_BY_SPARK))
+    )
```
jfongatyelp, line 659: Do we ensure spark.driver.cores is convertible to an int anywhere before this, or should we catch exceptions here too like we do for memory?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -894,6 +954,16 @@ def paasta_spark_run(args):
...
+
+    # Get driver-memory and cores
+    sub_cmds = args.cmd.split(" ")  # spark.driver.memory=10g
+    for cmd in sub_cmds:
+        if cmd.startswith("spark.driver.memory") or cmd.startswith(
+            "spark.driver.cores"
+        ):
+            key, value = cmd.split("=")
+            user_spark_opts[key] = value
+
```
jfongatyelp, line 966: Is this config not always part of `--spark args '...'` in the cmd? If it is, could we not grab this out of the spark_args that we parsed just above this? It seems like [_parse_user_spark_args](https://github.com/Yelp/paasta/blob/012a17ea344e6c932a05607af105a03fdf2ed71c/paasta_tools/cli/cmds/spark_run.py#L545-L555) would just parse anything out of there and throw it into `user_spark_opts`?

jfongatyelp (approved):

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -56,6 +56,12 @@
 CLUSTER_MANAGER_MESOS = "mesos"
 CLUSTER_MANAGER_K8S = "kubernetes"
 CLUSTER_MANAGERS = {CLUSTER_MANAGER_MESOS, CLUSTER_MANAGER_K8S}
+# Reference: https://spark.apache.org/docs/latest/configuration.html#application-properties
+DEFAULT_DRIVER_CORES_BY_SPARK = 1
+DEFAULT_DRIVER_MEMORY_BY_SPARK = "1g"
```
gcollura, line 61: I'm actually worried in making this the default at the moment. There are a lot of batches that don't specify `spark.driver.cores` and/or `spark.driver.memory` that might fail if this change is pushed. What is the pushplan for this change?

paasta_tools/cli/cmds/spark_run.py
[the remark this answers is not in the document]
gcollura: Looks like @jfongatyelp beat me to this comment :)

paasta_tools/cli/cmds/spark_run.py
chelspea: Do have the same concern as above

88manpreet: These default values are the one provided by spark for the above configs. 
The behavior of the batch should be same irrespective of these configs not provided or provided with above values. 
Manual test results in the ticket.

tl;dr when batch doesn't provide these configs:
* Docker container memory limit as 2g
* Docker cpu limit same 1

My only concern is, if should have the core limit > the one set by spark driver.

paasta_tools/cli/cmds/spark_run.py
88manpreet: 1. If spark.driver.cores is convertible to int before this, it would still work, right?
2. I tested in manual test cases, and didn't see any issues. Ticket comments for results
3. I thought about catching the error here too, but this seems like to basic operation. Only cases I see it failing is spark.driver.cores is provided as (non-int) string or float (or a different type than int-str). 
Happy to catch it for extra caution.

paasta_tools/cli/cmds/spark_run.py
88manpreet: This was added to cover all edge cases.
If users doesn't provide driver.cores as part of `--spark-args` but provide as part of `spark-submit` command, it would fail.
Example: `spark-submit --conf spark.driver.cores=12g` would set container memory as `2g`. 
I caught this in detailed manual runs, details in the ticket comment.

Happy to revisit, if I still missed something :)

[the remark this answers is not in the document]
88manpreet: Check the response above :)

paasta_tools/cli/cmds/spark_run.py
jfongatyelp: One more note about this is that `docker run` will flat out fail to run if we specify more `--cpus` than the host has total -- with our batch boxes having only 8vcpu, we'd want to be a little careful about this; some configs in yelpsoa-configs already appear to specify spark.driver.cores=8, thankfully none more than 8.

Is there already documentation to tell folks not to use >8? If not, the user experience might be slightly unfriendly, where they only find out their tronjob's config is not valid when the job fails.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -56,6 +56,12 @@
...
+# Extra room for memory overhead and for any other running inside containe
```
nemacysts, line 62: super-nit
```suggestion
# Extra room for memory overhead and for any other running inside container
```

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -606,6 +636,30 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
     return spark_app_name
 
 
+def _calculate_docker_memory_cpu_limit(spark_conf):
```
nemacysts, line 639: nit: could we add types to any new functions we add? imo, it'd also be nice to separate this into two functions so that you don't need to look at the code to know what order memory/cpu are in

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -606,6 +636,30 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
...
+    try:
+
+        docker_memory_limit_str = spark_conf.get(
+            "spark.driver.memory", DEFAULT_DRIVER_MEMORY_BY_SPARK
+        )
+        match = re.match(r"([0-9]+)([a-z]*)", docker_memory_limit_str)
```
nemacysts, line 645: should the memory unit party of the regex be the subset of allowed units? e.g., 
```Python
        match = re.match(r"([0-9]+)(mb|gb)?", docker_memory_limit_str.lower())
```

or do we have a pre-commit hook or something that ensures that people don't make mistakes here?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -56,6 +56,12 @@
 CLUSTER_MANAGER_MESOS = "mesos"
 CLUSTER_MANAGER_K8S = "kubernetes"
 CLUSTER_MANAGERS = {CLUSTER_MANAGER_MESOS, CLUSTER_MANAGER_K8S}
+# Reference: https://spark.apache.org/docs/latest/configuration.html#application-properties
+DEFAULT_DRIVER_CORES_BY_SPARK = 1
+DEFAULT_DRIVER_MEMORY_BY_SPARK = "1g"
+# Extra room for memory overhead and for any other running inside containe
+DOCKER_RESOURCE_ADJUSTMENT_FACTOR = 2
```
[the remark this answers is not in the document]
88manpreet: There is driver.memory.overheadfactor which also uses driver memory (default 10%) and some users can have higher value. Accounting that into the calculation can make the code complicated.

Question: If the docker container memory limit is set to 4G (let's say), and the box has 40G storage. Does that mean only 10 docker containers will be allocated at max? Or 20 docker containers could be allocated too, but none of them would exceed 4G.

paasta_tools/cli/cmds/spark_run.py
88manpreet: If the units are wrong, the spark driver will fail and/or so does creating the container. 
I think adding this check will be an overkill here, and would make it harder to debug in future and needs to consider all possible edge cases (e.g. 1000 is valid value too just like 1000m).

paasta_tools/cli/cmds/spark_run.py
jfongatyelp: Ah, I didn't realize people could specify this in the spark-submit command instead, thanks for the explanation.

jfongatyelp: Could you add a comment indicating this for future us?

paasta_tools/cli/cmds/spark_run.py
88manpreet: Good Catch. I tested this max value case for driver-memory (in the ticket) where in `docker run` just set the value to max memory of the box by default. Strange to me, that it didn't do the same for `cpus' too.
I will test the same for the max value too.

What should be the max value you suggest? This is considering that many instance types are being used for batch and stable_batch boxes?
Is 8vcpu's standard and reliable long-term consistent value to be used to check for max allowed value?

jfongatyelp: 8 seems fine to leave as the max although it's literally the maximum available on these hosts. The batch and stable_batch boxes aren't involved with the driver, and the boxes that run our drivers are all only 1 instance type right now w/ only 8vcpu. 

So since we're counting on drivers moving to k8s soon (where we can actually schedule based on resources requested), it might be sufficient just to leave this and update documentation, maybe add a pre-commit to yelpsoa-configs.

88manpreet: Ok. I will leave it like that. 
Interesting in local test, when i tried setting 10 cores. It just let me and continued running.
`docker inspect 6c5e647b4632 | grep Nano
            "NanoCpus": 10000000000,`
Maybe it becomes more of the useless value when greater than max-cores as opposed to an issue. 
Which I think might be ok.

paasta_tools/cli/cmds/spark_run.py
88manpreet: Added.

paasta_tools/cli/cmds/spark_run.py
88manpreet: @gcollura , @chelspea let me know if I can resolve this comment?

paasta_tools/cli/cmds/spark_run.py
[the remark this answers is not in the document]
88manpreet: I make it configurable optionally.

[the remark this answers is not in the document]
nemacysts: for your question: since we're running these containers directly, it'd be the latter - there's nothing stopping Tron's SSH code from starting more docker containers (with a 4GB memory limit) than would actually fit onto a box. 

(e.g., it would happily try to start 100 containers with a memory limit of 4GB)

Setting this limit won't solve all the problems with running the drivers on bare-metal because of this, but it should significantly improve things when one driver starts hogging all the memory

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -894,6 +992,18 @@ def paasta_spark_run(args):
     user_spark_opts = _parse_user_spark_args(
         args.spark_args, pod_template_path, args.enable_compact_bin_packing
     )
+
+    # Get driver-memory and cores
+    sub_cmds = args.cmd.split(" ")  # spark.driver.memory=10g
```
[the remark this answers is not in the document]
nemacysts: this also seems a little brittle, but that's a problem for future-us I guess (something like providing native soaconfigs support for this would be nice)

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -606,6 +650,53 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
...
+    2. --spark-args or spark-submit: spark.driver.cores
+    3. Default
+    """
+    if cpu_limit:
+        docker_cpu_limit = cpu_limit
+    else:
+        docker_cpu_limit = str(
+            int(spark_conf.get("spark.driver.cores", DEFAULT_DRIVER_CORES_BY_SPARK))
+        )
+    return docker_cpu_limit
```
nemacysts, line 697: feel free to drop this suggestion:
```suggestion
    return cpu_limit if cpu_limit else spark_conf.get("spark.driver.cores", str(DEFAULT_DRIVER_CORES_BY_SPARK))
```

(from my understanding `spark.driver.cores` will always be a string, so we really just want to cast the default core value)

paasta_tools/cli/cmds/spark_run.py
88manpreet: Added

paasta_tools/cli/cmds/spark_run.py
jfongatyelp: ^This test was done on a host that had >10 cpu available; can you try again either specifying more cpu than your test host has or testing on one of the batch boxes?

my concern here is that for most `spark-run` tron jobs, the driver will be run on a host that has only 8 cpu available, so we either a) give users a bad experience where they can set up yelpsoa-configs with >8 spark.driver.cores for their job, then only find out it fails on the first scheduled run, or b) add some sort of safety here to just set this value to maximum for the host it's running on

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -606,6 +650,53 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
...
+) -> str:
+    """ In Order of preference:
+    1. Argument: --docker-memory-limit
+    2. --spark-args or spark-submit: spark.driver.memory
+    3. Default
+    """
+    try:
+        if memory_limit:
+            docker_memory_limit_str = memory_limit
+            adjustment_factor = 1
```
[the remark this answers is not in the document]
88manpreet: The reason I kept it is the same you mentioned later.
Making sure we apply the same validation rules to user inputs through command-line args or configs. 
I believe docker will validate anyways. I can simplify this.

paasta_tools/cli/cmds/spark_run.py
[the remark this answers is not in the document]
88manpreet: Makes sense. That's what I thought too. Thanks for confirming :)

paasta_tools/cli/cmds/spark_run.py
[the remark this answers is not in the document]
88manpreet: I had added the detailed comment at line #993-994. I will move it upwards. Hopefully, that is more clearer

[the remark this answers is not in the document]
88manpreet: > just double-checking: there's no values for any of these config options that could potentially have a space in them?
This is correct. In fact I could remove this too. There is no real use-case at Yelp being used in this.
I added it to take care of future possible edge cases.

tests/cli/test_cmds_spark_run.py
```diff
@@ -42,12 +44,21 @@ def test_get_docker_run_cmd(mock_getegid, mock_geteuid):
...
+        volumes,
+        env,
+        docker_img,
+        docker_cmd,
+        nvidia,
+        docker_memory_limit,
+        docker_cpu_limit,
     )
 
-    assert actual[5:] == [
```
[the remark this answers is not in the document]
88manpreet: Seems like an overkill check to me, as we are setting it for all possible edge cases.
This should get caught in unit-tests in case of any failure.

paasta_tools/cli/cmds/spark_run.py
88manpreet: I tried with > 10 cpus. It failed. I added more details in the help section for better user experience. This has been discussed offline with @jfongatyelp

paasta_tools/cli/cmds/spark_run.py
88manpreet: Fixed

paasta_tools/cli/cmds/spark_run.py
88manpreet: Added back.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -143,6 +149,20 @@ def add_subparser(subparsers):
...
+            "Set docker memory limit. Should be greater than driver memory. Defaults to 2x driver memory. Example: 2g, 500m, Max: 64g"
+        ),
+        default=False,
+    )
+    list_parser.add_argument(
+        "--docker-cpu-limit",
+        help=(
+            "Set docker cpus limit. Should be greater than driver cores. Defaults to 1x driver cores."
+        ),
+        default=False,
```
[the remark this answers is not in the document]
88manpreet: Added back.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -606,6 +650,53 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
...
+        log.error(
+            f"Failed to parse docker memory limit. Error: {e}. Example values: 1g, 200m."
+            " Setting a default value {DEFAULT_MAX_DOCKER_MEMORY_LIMIT}"
+        )
+        docker_memory_limit = DEFAULT_MAX_DOCKER_MEMORY_LIMIT
+
+    return docker_memory_limit
+
+
+def _calculate_docker_cpu_limit(spark_conf: Mapping[str, str], cpu_limit: str) -> str:
```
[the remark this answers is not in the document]
88manpreet: Added back.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -606,6 +650,53 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
     return spark_app_name
 
 
+def _calculate_docker_memory_limit(
+    spark_conf: Mapping[str, str], memory_limit: str
```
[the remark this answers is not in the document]
88manpreet: Added back.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -606,6 +652,52 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
...
+        memory_val = int(match[1]) * adjustment_factor
+        memory_unit = match[2]
+        docker_memory_limit = f"{memory_val}{memory_unit}"
+    except Exception as e:
+        # For any reason it fails, continue with default value
+        print(
+            f"Failed to parse docker memory limit. Error: {e}. Example values: 1g, 200m."
+            " Setting a default value {DEFAULT_MAX_DOCKER_MEMORY_LIMIT}"
+        )
+        docker_memory_limit = DEFAULT_MAX_DOCKER_MEMORY_LIMIT
```
nemacysts, line 681: thinking about this more: should we just fail here? we're already going to fail if people do something weird with the cpu limit too

paasta_tools/cli/cmds/spark_run.py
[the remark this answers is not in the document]
nemacysts: I was thinking of making this configurable through Puppet (using https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L1847) so that we don't need to release new paasta versions to tweak this factor - but this is fine for now

gcollura:
I don't have much to add, @nemacysts left great comments.
One thought though, Tron on Paasta has `cpus/mem` options, should we add a feature in Tron where cpu and mem for the driver can be set through tron config as well?

paasta_tools/cli/cmds/spark_run.py
88manpreet: Most users specify high memory limit as part of over-provisioning for successful run of the job.
Having it failed IMHO is not a smooth user-experience for both onpoint and end-user (delay, notice, reach onpoint, change memory limits, calculate the right value, knowing the max value).

Even if it fails, the user would need to specify the max driver limit of the container (or half of it). 
Having it as a warning is good enough to be aware of and take decisions if required. 
I would rather prefer to keep it like it this.

88manpreet:
> I don't have much to add, @nemacysts left great comments. One thought though, Tron on Paasta has `cpus/mem` options, should we add a feature in Tron where cpu and mem for the driver can be set through tron config as well?

I think if we can find the strong case for generic usage, it would be useful to have.
I think we would want to make our user experience less as less confusing with multiple parameters as possible.
Right now, they can do it with `--docker-<cpu|memory>-limit` (which is still configurable). But I doubt it is going to required much. The default values should suffice for 90% of the cases, I think. 

Still something good to keep in our mind, for future versions.

paasta_tools/cli/cmds/spark_run.py
nemacysts: this block would only get hit if users made a mistake setting up their config tho, right? e.g., if they set `10h` instead of `10g`, we/they would never know that their config is invalid and this would basically mean that we fall back to the status quo where a single spark job can take down a box

nemacysts:
> I don't have much to add, @nemacysts left great comments. One thought though, Tron on Paasta has `cpus/mem` options, should we add a feature in Tron where cpu and mem for the driver can be set through tron config as well?

@gcollura i was actually thinking of adding support for something like this when I start working on the driver-in-k8s stuff for y'all :)

paasta_tools/cli/cmds/spark_run.py
88manpreet: Sorry I misinterpreted your comment in lines with failing when user specifies > than the max allowed limit of the box.
Yes, you are right.
My original reasoning to not fail was, in the goal of not impacting batches doesn't fall into this regex and keep the job running as status-quo. As in WARN, Move-on, fix based on warning for next runs.

I guess I will change this to fail-hard.

nemacysts: would it be worth writing a quick script to check all the existing configs to see if there are any existing batches that have incorrect config?

nemacysts (approved):

gcollura (approved):

jfongatyelp (approved):

PR #3242, opened 2021-11-19
Add Pod Identity support for Tron Pods

nemacysts:
We'll want to merge this after https://github.com/Yelp/Tron/pull/852

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2965,7 +2965,10 @@ def create_or_find_service_account_name(
     iam_role: str, namespace: str = "paasta"
 ) -> str:
     kube_client = KubeClient()
-    sa_name = f"{namespace}--{_RE_NORMALIZE_IAM_ROLE.sub('-', iam_role)}"
+    # the service account is expected to always be prefixed with paasta- as using the actual namespace
+    # potentially wastes a lot of characters (e.g., paasta-nrtsearchservices) that could be used for
+    # the actual name
+    sa_name = f"paasta--{_RE_NORMALIZE_IAM_ROLE.sub('-', iam_role)}"
     if not any(
         sa.metadata.name == sa_name
         for sa in get_all_service_accounts(kube_client, namespace)
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2965,7 +2965,7 @@ def create_or_find_service_account_name(
     iam_role: str, namespace: str = "paasta"
 ) -> str:
     kube_client = KubeClient()
-    sa_name = "paasta--" + _RE_NORMALIZE_IAM_ROLE.sub("-", iam_role)
+    sa_name = f"{namespace}--{_RE_NORMALIZE_IAM_ROLE.sub('-', iam_role)}"
     if not any(
         sa.metadata.name == sa_name
         for sa in get_all_service_accounts(kube_client, namespace)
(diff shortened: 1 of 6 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -2965,7 +2965,7 @@ def create_or_find_service_account_name(
     iam_role: str, namespace: str = "paasta"
 ) -> str:
     kube_client = KubeClient()
-    sa_name = "paasta--" + _RE_NORMALIZE_IAM_ROLE.sub("-", iam_role)
+    sa_name = f"{namespace}--{_RE_NORMALIZE_IAM_ROLE.sub('-', iam_role)}"
```
nemacysts, line 2968: i'm assuming we don't want names starting with `paasta` in other namespaces

paasta_tools/tron_tools.py
```diff
@@ -775,6 +776,15 @@ def format_tron_action_dict(action_config: TronActionConfig, use_k8s: bool = Fal
         if action_config.get_team() is not None:
             result["labels"]["yelp.com/owner"] = action_config.get_team()
 
+        if (
+            action_config.get_iam_role_provider() == "aws"
+            and action_config.get_iam_role()
+        ):
+            result["annotations"]["iam.amazonaws.com/role"] = ""
```
nemacysts, line 783: do we actually need this if the service account is already associated with the role?

jfongatyelp (approved):

paasta_tools/kubernetes_tools.py
danielpops: idk, other things that feel "less paasta" than tron workloads are all prefixed with `paasta-`

```
$ kubectl-pnw-devc get namespaces
INFO: You are using an Okta authenticated kub
NAME                       STATUS   AGE
custom-metrics             Active   301d
default                    Active   2y91d
external-dns               Active   512d
ingress-nginx              Active   2y91d
jenkins                    Active   202d
jupyterhub                 Active   567d
kiam                       Active   2y91d
kube-node-lease            Active   2y91d
kube-public                Active   2y91d
kube-system                Active   2y91d
paasta                     Active   2y91d
paasta-cassandraclusters   Active   2y91d
paasta-flinks              Active   2y91d
paasta-kafkaclusters       Active   2y49d
paasta-monkrelays          Active   121d
paasta-nrtsearchservices   Active   570d
paasta-spark               Active   524d
tron                       Active   149d
```

danielpops: oh, i see that my point wasn't relevant to what you're saying here. presumably we've already decided to go with `tron` namespace for tron workloads, and you're saying that we should have service accounts within the namespace starting with the namespace name. makes sense to me.

danielpops (approved):

paasta_tools/kubernetes_tools.py
nemacysts: yup, everything that tron uses through task_proc exists in the tron namespace (secrets, pods, etc)

the names don't actually matter as long as they exist when task_proc tries to create a pod, but it offended my sensibilities to have dynamically created service accounts with names that referenced paasta in the tron namespace :p

vkhromov: nit: This should work, but I don't really like it tbh.
The `paasta-` prefix string was originally used to make sure that paasta-specific service account names don't clash with SA names added by other internal and external k8s components.  It was intended to be used in all namespace, not only in the `paasta` namespace (even that we use it for that namespace only atm).  In this sense that string it's isn't the namespace name but a prefix, even that they are the same atm.
The second reason is for namespaces named like `paasta-somethingsomething` such long prefix will waste additional characters from the SA name.

vkhromov (approved):
lgtm overall, see the comment

paasta_tools/kubernetes_tools.py
nemacysts: ah cool! i’ll revert this bit then and leave the prefix as paasta-

jfongatyelp (approved):

jfongatyelp (approved):

PR #3244, opened 2021-11-23
add more configuration options to shard script

kedarv:


paasta_tools/contrib/service_shard_update.py
```diff
@@ -118,7 +158,13 @@ def main(args):
                     if args.shard_name not in kube_file.keys():
                         kube_file[args.shard_name] = {
                             "deploy_group": f"{deploy_prefix}.{args.shard_name}",
-                            "instances": args.instance_count,
+                            "min_instances": args.min_instance_count,
+                            "max_instances": args.prod_max_instance_count
+                            if deploy_prefix == "prod"
+                            else args.non_prod_max_instance_count,
```
stug, line 164: yikes did the black autoformatting do this?  :-/

paasta_tools/contrib/service_shard_update.py
```diff
@@ -118,7 +158,13 @@ def main(args):
...
+                            "cpus": args.cpus,
+                            "mem": args.mem,
+                            "autoscaling": {"setpoint": args.setpoint,},
```
stug, line 167: since we are using these always, maybe we should make them required somehow?  What happens if they are not specified?

stug (commented):
Roughly makes sense to me, but I am worried about the case where not all arguments are specified

paasta_tools/contrib/service_shard_update.py
kedarv: oh it did

paasta_tools/contrib/service_shard_update.py
kedarv: I have the args with default values of 100, though in practice this tool is only called in automated workflows so I think it's okay to not have "more sane" defaults

stug: oh d'oh i missed that.

stug (approved):

kawaiwanyelp (approved):

PR #3245, opened 2021-12-07
Measure deploy duration for mark-for-deployment

kawaiwanyelp:
### Description
First step to setting up a deployment SLO is to measure deployment duration is to collect the relevant metrics. This PR will collect that metric whenever `mark-for-deployment --wait-for-deployment` is run. The following restrictions apply:

- Duration is not measured whenever `--wait-for-deployment` is not used, nor if m-f-d fails due to validation errors.
- By default, the metrics collection class is `NoMetrics` (i.e. noop); to turn on collection by `meteorite`, `metrics_provider` has to be provided in system paasta configs (currently we do not set it).

### Testing
- `make test`
- Manual testing using m-f-d on a test service, including with local system paasta configs with `metrics_provider` turned on

### Notes
Next step is to release it, then turn on metrics collection, first in stage, then everywhere else when we see that nothing crashes.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -68,6 +68,7 @@ from paasta_tools.deployment_utils import get_currently_deployed_sha
 from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
 from paasta_tools.long_running_service_tools import LongRunningServiceConfig
 from paasta_tools.marathon_tools import MarathonServiceConfig
+from paasta_tools.metrics import metrics_lib
 from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.paastaapi.models import InstanceStatusKubernetesV2
 from paasta_tools.paastaapi.models import KubernetesPodV2
@@ -403,7 +404,7 @@ def print_rollback_cmd(
         )
 
 
-def paasta_mark_for_deployment(args: argparse.Namespace) -> None:
+def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
     """Wrapping mark_for_deployment"""
     if args.verbose:
         log.setLevel(level=logging.DEBUG)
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -465,28 +466,51 @@ def paasta_mark_for_deployment(args: argparse.Namespace) -> None:
...
+    if args.block:  # only time if wait for deployment
+        metrics_factory = metrics_lib.get_metrics_interface
+    metrics = metrics_factory("paasta.mark_for_deployment")
+    deploy_timer = metrics.create_timer(
+        name="deploy_duration",
+        default_dimensions=dict(
+            paasta_service=service,
+            deploy_group=deploy_group,
+            old_version=old_git_sha,
+            new_version=commit,
```
nemacysts, line 479: should we skip metric emission if `old_git_sha == commit`? that'll mess with the timings otherwise

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -403,7 +404,7 @@ def print_rollback_cmd(
         )
 
 
-def paasta_mark_for_deployment(args: argparse.Namespace) -> None:
+def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
```
nemacysts, line 407: hmm, we should add the `py.typed` marker to sticht so that it would have actually caught this :p

paasta_tools/metrics/metrics_lib.py
```diff
@@ -24,13 +25,24 @@
 
 
 class TimerProtocol(Protocol):
+    def __enter__(self) -> "TimerProtocol":
```
nemacysts, line 28: just curious: what are these changes for? doesn't look like we're using any context managers in this diff

kawaiwanyelp: I was originally using `yelp_meteorite.create_timer` through the metrics_lib abstraction as a context manager, since meteorite allows it, but paasta's metrics_lib stuff doesn't support it so I just added the necessary stuff so that `NoMetrics` could also do it. However, as I commented in m-f-d, I ended up scrapping the context manager usage since on exceptions, meteorite's `create_timer` does not emit. Therefore, this code is optional but it's helpful so I didn't remove it. 

If you have strong feelings about it, I'll remove it.

nemacysts: nah, saves us some trouble in the future if we do ever use this in a place that won't potentially throw

paasta_tools/cli/cmds/mark_for_deployment.py
kawaiwanyelp: Good call.

nemacysts (approved):

PR #3248, opened 2021-12-09
Add validation for users overriding CPU request on CPU-scaled instances

nemacysts:
See PAASTA-17522 - but tl;dr: this can be a footgun (and especially so
for uwsgi-based services)


nemacysts:
I still have to write the wiki page, but we can review this before the page is real :p

nemacysts:
looks like some of the tests are failing since Azure's Ubuntu repos are having a bad time? It's safe to ignore those failures for now/this PR.

matthewbentley (approved):

EvanKrall (approved):

PR #3251, opened 2022-01-04
adds service_shard_remove script

kedarv:
The inverse of https://github.com/Yelp/paasta/blob/master/paasta_tools/contrib/service_shard_update.py (remove a shard)

### DIFF
--- a/paasta_tools/contrib/service_shard_remove.py
+++ b/paasta_tools/contrib/service_shard_remove.py
@@ -1,0 +1,141 @@
+import argparse
+import logging
+
+from paasta_tools.config_utils import AutoConfigUpdater
+from paasta_tools.utils import DEFAULT_SOA_CONFIGS_GIT_URL
+from paasta_tools.utils import format_git_url
+from paasta_tools.utils import load_system_paasta_config
+
+log = logging.getLogger(__name__)
+
+
+def parse_args():
+    parser = argparse.ArgumentParser(description="")
+    parser.add_argument(
+        "--git-remote",
+        help="Master git repo for soaconfigs",
+        default=None,
+        dest="git_remote",
+    )
+    parser.add_argument(
+        "--branch", help="Branch name to push to", required=True, dest="branch",
+    )
+    parser.add_argument(
+        "--local-dir",
+        help="Act on configs in the local directory rather than cloning the git_remote",
+        required=False,
+        default=None,
+        dest="local_dir",
+    )
+    parser.add_argument(
+        "-v",
+        "--verbose",
+        help="Logging verbosity",
+        action="store_true",
+        dest="verbose",
+    )
+    parser.add_argument(
+        "--source-id",
+        help="String to attribute the changes in the commit message.",
+        required=False,
+        default=None,
+        dest="source_id",
+    )
+    parser.add_argument(
+        "--service", help="Service to modify", required=True, dest="service",
+    )
+    parser.add_argument(
+        "--shard-name", help="Shard name to remove", required=True, dest="shard_name",
+    )
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


karmeleon:
okay, so it looks like we'll use this as part of a script that runs periodically to delete unused shards. what mechanism is going to add the shards back? is it just when a deploy happens?

kedarv:
> okay, so it looks like we'll use this as part of a script that runs periodically to delete unused shards. what mechanism is going to add the shards back? is it just when a deploy happens?

Yeah, I think that would happen separate of here, such as adding a blocklist of shards in the same mechanism that deletes unused shards.

paasta_tools/contrib/service_shard_remove.py
```diff
@@ -0,0 +1,141 @@
...
+    default_git_remote = format_git_url(
+        system_paasta_config.get_git_config()["git_user"],
+        repo_config.get("git_server", DEFAULT_SOA_CONFIGS_GIT_URL),
+        repo_config["repo_name"],
+    )
+    return default_git_remote
+
+
+SUPERREGION_K8S_MAPPING = {
+    "norcal-devc": "kubernetes-norcal-devc",
```
nemacysts, line 71: do we need this? the format will always be kubernetes-{x}

paasta_tools/contrib/service_shard_remove.py
```diff
@@ -0,0 +1,141 @@
...
+    "norcal-stageg": "stage",
+    "nova-prod": "prod",
+    "pnw-prod": "prod",
+}
+
+
+def main(args):
+    updater = AutoConfigUpdater(
+        config_source=args.source_id,
+        git_remote=args.git_remote or get_default_git_remote(),
```
nemacysts, line 90: should this be set as a default value when the argument parser is created?

paasta_tools/contrib/service_shard_remove.py
```diff
@@ -0,0 +1,141 @@
...
+        dest="source_id",
+    )
+    parser.add_argument(
+        "--service", help="Service to modify", required=True, dest="service",
+    )
+    parser.add_argument(
+        "--shard-name", help="Shard name to remove", required=True, dest="shard_name",
+    )
+    parser.add_argument(
+        "--superregion",
```
nemacysts, line 51: cluster might be a better name (e.g., i’m pretty sure infrastage and kubestage aren’t superregions)

nemacysts: this is more of a nit since y’all are probably never gonna run this in the test cluster tho

paasta_tools/contrib/service_shard_remove.py
```diff
@@ -0,0 +1,141 @@
...
+        nargs="+",
+        help="Superregion to remove shard from",
+        required=True,
+    )
+    return parser.parse_args()
+
+
+def get_default_git_remote():
```
nemacysts, line 59:
```suggestion
def get_default_git_remote() -> str:
```

nemacysts: could also add types to the other functions/constants, but not a blocker

paasta_tools/contrib/service_shard_remove.py
kedarv: Gotcha, I'm not super familiar with the naming of the hierarchy and how it relates to clusters, but the data source that we are using deals with superregions which is why I chose that here

paasta_tools/contrib/service_shard_remove.py
kedarv: oh right, good catch

karmeleon (approved):

PR #3252, opened 2022-01-05
Tronjobs can run paasta stop/start/restart

lanceleih:
We're planning to only allow certain users and groups to run `paasta stop/start/restart` in the future. Before implementing this new change, this fix will make sure that Tronjobs will still be allowed to run the command without it crashing because of the new authentication rules.

This is a temporary solution, until we inevitably remove Tronjob's ability to run `paasta stop/start/restart`.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -307,6 +307,16 @@ def mark_for_deployment(
 
 def can_user_deploy_service(deploy_info: Dict[str, Any], service: str) -> bool:
     deploy_username = get_username()
+
+    # Tronjobs can run paasta stop/start/restart
+    ssh_client_env = os.environ.get("SSH_CLIENT")
+    if ssh_client_env:
+        ssh_client = ssh_client_env.split()[0]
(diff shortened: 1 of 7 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -307,6 +307,16 @@ def mark_for_deployment(
...
+
+    # Tronjobs can run paasta stop/start/restart
+    ssh_client_env = os.environ.get("SSH_CLIENT")
+    if ssh_client_env:
+        ssh_client = ssh_client_env.split()[0]
+        hostname = socket.gethostbyaddr(ssh_client)[0]
+
+        if "tron" in hostname:
+            return True
+
```
jfongatyelp, line 319: Should we also restrict things running on tron to have `deploy_username` match the `batch` user too, not just the SSH_CLIENT hostname?

jfongatyelp (approved):

paasta_tools/cli/cmds/mark_for_deployment.py
lanceleih: Good point. Added check for batch user.

jfongatyelp (approved):

PR #3256, opened 2022-01-13
Skip k8s service account shenanigans during tronfig validation

nemacysts:
We don't actually want to talk to the k8s API during tronfig validation
- and much less to create service accounts.

We take advantage of the handy-dandy for_validation flag available in
TronActionConfig (which is set when `paasta validate` calls the tronfig
creation/validation functions) to do this skipping without having to
touch any other code.


nemacysts:
paasta validate without this change fails, with it it succeeds.

mks-m (approved):

PR #3257, opened 2022-01-14
Conditionally use ruamel in get_config_file_dict()

nemacysts:
Looks like ruamel doesn't handle multiple anchors at different levels of
nesting particularly well in some cases.

Since we only really need ruamel for parsing kubernetes config files which
are generally unlikely to do similar shenanigans, we can add a flag to
only use ruamel for these (and then furthermore, only for validating CPU
overrides).

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -22,6 +22,7 @@ from typing import Any
 from typing import Dict
 from typing import Optional
 
+import yaml
 from jsonschema import Draft4Validator
 from jsonschema import exceptions
 from jsonschema import FormatChecker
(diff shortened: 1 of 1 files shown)


nemacysts:
I also ran `paasta list | xargs -P5 -n1 --no-run-if-empty paasta validate -y /nail/home/luisp/pg/sysgit/yelpsoa-configs/ -s | rg '✗|Traceback' | fpb` and got https://fluffy.yelpcorp.com/i/FpZZpvHPzlXpHGSpzT8znTsHkWv2CMv9.html - which shows that there's no other weird failure cases

EvanKrall (approved):

PR #3258, opened 2022-01-18
[FLINK-3869] Fail explicitly for Flink service `restart`

marcos-sb:
Issuing 'paasta restart' of a flink service is not supported but we currently fail silently. This patch adds an intercept to notify the user about what happened and suggests a workaround.

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -41,7 +41,7 @@ from paasta_tools.utils import PaastaColors
 def add_subparser(subparsers):
     for command, lower, upper, cmd_func in [
         ("start", "start or restart", "Start or restart", paasta_start),
-        ("restart", "start or restart", "Start or restart", paasta_start),
+        ("restart", "start or restart", "Start or restart", paasta_restart),
         ("stop", "stop", "Stop", paasta_stop),
     ]:
         status_parser = subparsers.add_parser(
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -329,6 +329,28 @@ def paasta_start(args):
     return paasta_start_or_stop(args, "start")
 
 
+def paasta_restart(args):
```
AshishKhatkar, line 332: We currently have cases where there are services which have normal paasta instances and flink instance as well. It won't work for such services if user just did `paasta restart -s service -c pnw-devc` without instance for ex. The expectation from there side might be that atleast normal paasta instances will be restarted which was the case earlier. I will rather do `if not isinstance(service_config, FlinkDeploymentConfig)` call `paasta_start(args)` rather than exiting even if one instance is of Flink.
@Rob-Johnson what are your opinions on this?

AshishKhatkar (changes requested):

paasta_tools/cli/cmds/start_stop_restart.py
Rob-Johnson: initial thought is that if you just fail the flink instances and continue with the rest, the user might miss the log line/output that tells them the flink output fail.

you could present the user with a choice on whether to continue with the other instances to force the decision onto them and recognise that the flink instances won't be restarted.

Something like
```
paasta restart -s my-service -i flink_instance,batch_instance,web_instance

WARN: paasta restart is currently unsupported for Flink instances (my-service.flink_instance). To restart, please run:

paasta stop -s my-service -i flink_instance
paasta start -s myservice -i flink_instance

Would you like to restart the other instances (my-service.batch_instance, my-service.web_instance anyway)? (y/n)

y

... (default output for restart).
```

I think this would present pretty clear instruction for the user, and will force them to think about the choice and how to succeed with the restart.

For an example of how to present this kind of choice, see https://github.com/Yelp/paasta/blob/1ab8368b3eb9d32bfe8c7d2a63778c22cf87ba17/paasta_tools/cli/cmds/start_stop_restart.py#L172-L175

Rob-Johnson:
also cc'ing a few folks from CIPX for their input, since they own this @kawaiwanyelp @EvanKrall @jfongatyelp @nemacysts

AshishKhatkar (approved):
looks good to me. Let's wait for others if they have any comments.

Rob-Johnson:
please write a test, too.

Rob-Johnson (approved):

JoeMalt (approved):

AshishKhatkar (approved):

PR #3259, opened 2022-01-18
Drop xenial GHA builds + use internal source as much as possible

nemacysts:
I wasn't able to make the Mesos/Marathon Dockerfiles use our internal
bionic images, but these aren't long for this world anyway

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,7 +49,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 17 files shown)


nemacysts:
make itest_xenial itest_bionic itest k8s_itest seem good so far (itest and k8s_itest are still running locally tho)

nemacysts:
i'm gonna try to go with https://github.com/Yelp/paasta/pull/3260 and just hardcode all the xenial stuff to only work internally

PR #3260, opened 2022-01-19
Drop xenial GHA builds

nemacysts:
Our itest setup is annoyingly tangled and there's a lot of xenial
dockerfiles, so I'm dropping the itests from GHA until we have more time
to fix things correctly (likely by just removing all the mesos
code/itests)


nemacysts:
I can't get the itests to run on my devbox, so I tested this on a jenkins box and the itests have been running for a while now (seemingly successfully)

Buffer0x7cd (approved):

Rob-Johnson (approved):

PR #3262, opened 2022-01-21
Add support for syncing paasta secrets into multiple k8s namespaces

BeryJu:
Based on https://github.com/Yelp/paasta/pull/3009 and https://github.com/Yelp/paasta/pull/3151

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -40,13 +40,12 @@ from paasta_tools.kubernetes_tools import update_secret
 from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.secret_tools import get_secret_provider
 from paasta_tools.utils import DEFAULT_SOA_DIR
-from paasta_tools.utils import get_instance_type_to_k8s_namespace
 from paasta_tools.utils import get_service_instance_list
+from paasta_tools.utils import INSTANCE_TYPE_TO_K8S_NAMESPACE
 from paasta_tools.utils import INSTANCE_TYPES
 from paasta_tools.utils import load_system_paasta_config
 
 log = logging.getLogger(__name__)
-INSTANCE_TYPE_TO_K8S_NAMESPACE = get_instance_type_to_k8s_namespace()
 
 
 def parse_args() -> argparse.Namespace:
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -20,8 +20,10 @@ import logging
 import os
 import sys
 import time
+from collections import defaultdict
 from typing import Mapping
-from typing import Sequence
+from typing import Optional
+from typing import Set
 
 from kubernetes.client.rest import ApiException
 
@@ -38,9 +40,13 @@ from paasta_tools.kubernetes_tools import update_secret
 from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.secret_tools import get_secret_provider
 from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import get_instance_type_to_k8s_namespace
+from paasta_tools.utils import get_service_instance_list
+from paasta_tools.utils import INSTANCE_TYPES
 from paasta_tools.utils import load_system_paasta_config
 
 log = logging.getLogger(__name__)
+INSTANCE_TYPE_TO_K8S_NAMESPACE = get_instance_type_to_k8s_namespace()
 
 
 def parse_args() -> argparse.Namespace:
@@ -71,8 +77,7 @@ def parse_args() -> argparse.Namespace:
         "-n",
         "--namespace",
         dest="namespace",
-        default="paasta",
-        help="destination namespace for secrets (Default: %(default)s)",
+        help="Overwrite destination namespace for secrets",
     )
     parser.add_argument(
         "-v", "--verbose", action="store_true", dest="verbose", default=False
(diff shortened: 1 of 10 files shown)


paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -96,50 +100,66 @@ def main() -> None:
...
+            instances = get_service_instance_list(
+                service=service,
+                instance_type=instance_type,
+                cluster=cluster,
+                soa_dir=args.soa_dir,
+            )
+            if instances:
+                services_to_k8s_namespaces[service].add(
+                    get_instance_type_to_k8s_namespace()[instance_type]
+                )
```
nemacysts, line 115: should this be a function? one benefit we'd get from this being a function is that we could hide the defaultdict in the function and return a normal dict (so that we don't leave a trap for someone in the future when they thought they were working on a dict but were actually working on a defaultdict)

paasta_tools/utils.py
```diff
@@ -3806,3 +3807,33 @@ def _reorder_docker_volumes(volumes: List[DockerVolume]) -> List[DockerVolume]:
...
+            continue
+        # First attempt to use cr_id() to get the CR definition and namespace
+        cr_id = getattr(instance_type_module, "cr_id", default_cr_id)("", "")
+        if "namespace" in cr_id:
+            instance_type_to_k8s_namespace[instance_type] = cr_id["namespace"]
+        # Otherwise, for objects that don't have a cr_id(), fallback to `KUBERNETES_NAMESPACE`
+        instance_type_to_k8s_namespace[instance_type] = getattr(
+            instance_type_module, "KUBERNETES_NAMESPACE", "paasta"
+        )
+    return instance_type_to_k8s_namespace
```
nemacysts, line 3839: thoughts on having a simple dict of `{instance_type: namespace}`? 

might be a little clearer for supporting new instance types in the future - with this change, developers kinda need to either know that they have to set a KUBERNETES_NAMESPACE constant or define a `cr_id()` function whereas with a plain dict the mapping is all in one place (and they don't need to worry about what takes precedence)

it'd also make this code a little safer too - I personally always get a little worried by dynamic imports/getattrs since they circumvent mypy and are a little harder to follow as a human reading the code

(feel free to close this if you disagree though!)

tests/kubernetes/bin/test_paasta_secrets_sync.py
```diff
@@ -40,46 +43,49 @@ def test_main():
...
         "paasta_tools.kubernetes.bin.paasta_secrets_sync.PaastaServiceConfigLoader",
         autospec=True,
     ):
+        services_to_k8s_namespaces = defaultdict(set)
+        services_to_k8s_namespaces["foo"].add(
+            get_instance_type_to_k8s_namespace()["kubernetes"]
+        )
+        services_to_k8s_namespaces["bar"].add(
+            get_instance_type_to_k8s_namespace()["kubernetes"]
+        )
```
nemacysts, line 59: might be nice to construct this as a normal dict so that we make sure we're not depending on things that use this data to be expecting a defaultdict

paasta_tools/utils.py
BeryJu: I do mostly agree with you, I'd prefer something clearer aswell (I was thinking a decorator on the `cr_id` function that adds it), but just having a single dict is even easier. The only worry I have with the dict-approach is that people will forget to add their stuff there, but since there is a fallback to `paasta` as namespace I dont think it would cause any issues.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
BeryJu: I've refactored it into a function that returns a normal dict, not sure if I like either approach better

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -96,50 +111,76 @@ def main() -> None:
...
+                service=service,
+                instance_type=instance_type,
+                cluster=cluster,
+                soa_dir=soa_dir,
+            )
+            if instances:
+                services_to_k8s_namespaces[service].add(
+                    INSTANCE_TYPE_TO_K8S_NAMESPACE[instance_type]
+                )
+    return dict(services_to_k8s_namespaces)
```
vkhromov, line 145: Why
```
    return dict(services_to_k8s_namespaces)
```
and not just
```
    return services_to_k8s_namespaces
```
?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -96,50 +111,76 @@ def main() -> None:
...
         vault_cluster_config=vault_cluster_config,
         soa_dir=args.soa_dir,
-        namespace=args.namespace,
+        overwrite_namespace=args.namespace,
     ) else sys.exit(1)
 
 
+def get_services_to_k8s_namespaces(
+    service_list: List[str], cluster: str, soa_dir: str,
+) -> Mapping[str, Set]:
```
vkhromov, line 131: Here and everywhere below.
Isn't it `Mapping[str, Set[str]]`?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -96,50 +111,76 @@ def main() -> None:
...
+    services_to_k8s_namespaces: Mapping[str, set] = defaultdict(set)
```
vkhromov, line 132:
```
services_to_k8s_namespaces: Mapping[str, Set[str]]
```
?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -38,10 +41,23 @@
...
+INSTANCE_TYPE_TO_K8S_NAMESPACE = {
+    "marathon": "paasta",
+    "adhoc": "paasta",
+    "kubernetes": "paasta",
+    "tron": "tron",
+    "flink": "paasta",
+    "cassandracluster": "paasta-cassandraclusters",
+    "kafkacluster": "paasta",
+    "nrtsearchservice": "paasta",
+}
```
vkhromov, line 59: Could you please double-check that namespaces are correct for all instance types.
Specifically, why we use `paasta` namespace for `flink`/`kafkacluster`/`nrtsearchservice` instance type and not `paasta-flinks`/`paasta-kafkaclusters`/`paasta-nrtsearchservices` namespaces?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -38,10 +41,23 @@
 from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.secret_tools import get_secret_provider
 from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import get_service_instance_list
+from paasta_tools.utils import INSTANCE_TYPES
 from paasta_tools.utils import load_system_paasta_config
 
 log = logging.getLogger(__name__)
 
+INSTANCE_TYPE_TO_K8S_NAMESPACE = {
```
vkhromov, line 50: Please move `INSTANCE_TYPE_TO_K8S_NAMESPACE` into `paasta_tools.utils next` next to `INSTANCE_TYPES`.

vkhromov (commented):
Looks good, please see the comments.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
BeryJu: cast to dict since we initialise it with a `defaultdict`, and returning that might cause issues debugging

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
BeryJu: yeah true, I forgot that typing.Set is a generic since I normally don't use sets much

vkhromov (changes requested):
Before merging this PR please test that `paasta` still syncs the existing secrets as expected, i.e. they are in `paasta` namespace.
For that you could build a `.deb` package locally on the devbox with your PR included, install it on nodes in `kubestage` cluster, and check if the secrets there are still fine.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
BeryJu: didn't even notice that `INSTANCE_TYPES` was from there, definitely makes more sense there.

nemacysts (approved):

vkhromov (approved):

PR #3264, opened 2022-01-26
Update check_orphan_registration.py to print hostname

akshaysharma096:
This PR makes the following changes

- Remove logging of count of orphaned hosts and change it to the name of the orphaned hosts

Ticket - https://jira.yelpcorp.com/browse/COMPINFRA-925


TheNavigat:
It doesn't look like the output if this is going to be what we expect. Have you tested this? Can you add a sample of the output to the ticket's description?

PR #3265, opened 2022-01-28
cassandra status: fix error when status.nodes is None

siadat:
The value of status.nodes shouldn't be None normally, however, this PR ensures
that paasta status command will not break even if it is.

Verify it by running:
```
$ paasta status -s service -c cluster -i instance
service: service
cluster: cluster
    instance: instance
    Git sha:    xxxxxxxx (desired)
    Cassandra cluster:
        State: Running
        Nodes: No node status available
```

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1778,14 +1778,22 @@ def print_cassandra_status(
     else:
         state = PaastaColors.red(state)
 
+    nodes: List[Dict[str, Any]] = status.get("nodes") or []
     output.append(indent * tab + "State: " + state)
-    output.append(indent * tab + "Nodes: ")
+
+    if not nodes:
+        output.append(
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -1778,14 +1778,21 @@ def print_cassandra_status(
...
         state = PaastaColors.red(state)
 
+    nodes: List[Dict[str, Any]] = status.get("nodes") or []
     output.append(indent * tab + "State: " + state)
-    output.append(indent * tab + "Nodes: ")
+    if len(nodes) == 0:
+        output.append(
+            indent * tab + "Nodes: " + PaastaColors.red("No node status available")
+        )
+        return 1
```
ymuzammil, line 1787: Do we want to make it sound like an error?

Rob-Johnson (approved):

atallahade (approved):

ymuzammil (approved):

PR #3266, opened 2022-02-01
Add schema support for executor: spark

nemacysts:
We previously started work on running Spark drivers on Mesos directly
(rather than having them run on physical boxes), but this was never
used. I'm deleting the existing code so that we can start anew.

We're generally going to take the same approach here (i.e., have a
spark_args dictionary in soaconfigs that configures everything so
that users don't need to learn anything new/can use existing spark
knowledge + we don't have to stay on top of what Spark options exist to
add native support for them).

That said, I've added some schema checking for options that I've seen be
used somewhat frequently.

NOTE: this is completely unused at the moment, just splitting this into
its own PR to make future PRs smaller :p

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -79,7 +79,7 @@ VALID_MONITORING_KEYS = set(
         pkgutil.get_data("paasta_tools.cli", "schemas/tron_schema.json").decode()
     )["definitions"]["job"]["properties"]["monitoring"]["properties"].keys()
 )
-MESOS_EXECUTOR_NAMES = ("paasta", "spark")
+MESOS_EXECUTOR_NAMES = ("paasta",)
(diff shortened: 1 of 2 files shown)


paasta_tools/tron_tools.py
```diff
@@ -244,15 +244,6 @@ def get_docker_url(
...
 
-    def get_spark_paasta_cluster(self):
-        return self.config_dict.get("spark_paasta_cluster", self.get_cluster())
-
-    def get_spark_paasta_pool(self):
-        return self.config_dict.get("spark_paasta_pool", "batch")
-
-    def get_spark_cluster_manager(self):
-        return self.config_dict.get("spark_cluster_manager", "mesos")
-
```
nemacysts, line 255: turns out i missed deleting some stuff :p

chelspea (approved):

gcollura (approved):

PR #3267, opened 2022-02-03
Add soa-configs SHA to containers without messing up configs SHA use …

kawaiwanyelp:
…to diff for deployment updates

### Description
This PR makes instance configs pull in a soa-configs SHA from the new soa-configs metadata file, if there is one, and sets it as an env var in containers. This env var is explicitly removed so that it is not factored into hashing for the configs SHA to prevent services from re-deploying every time the soa-configs SHA changes (which is pretty often). 

### Testing
`make test`
local testing 

### Notes
This will need to be released in dev or stage first for safety to make sure we aren't redeploying service instances every time a soa-configs change is made.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -11,7 +11,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import base64
-import copy
 import hashlib
 import itertools
 import json
@@ -151,7 +150,6 @@ log = logging.getLogger(__name__)
 KUBE_CONFIG_PATH = "/etc/kubernetes/admin.conf"
 YELP_ATTRIBUTE_PREFIX = "yelp.com/"
 PAASTA_ATTRIBUTE_PREFIX = "paasta.yelp.com/"
-CONFIG_HASH_BLACKLIST = {"replicas"}
 KUBE_DEPLOY_STATEGY_MAP = {
     "crossover": "RollingUpdate",
     "downthenup": "Recreate",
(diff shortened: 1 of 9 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1855,20 +1853,26 @@ def sanitize_for_config_hash(
...
             service=self.get_service(), environment_variables=self.get_env()
         )
+
+        # remove data we dont want used to hash configs
+        # replica count
+        if ahash.get("spec", {}).get("replicas"):
+            del ahash["spec"]["replicas"]
+        # soa-configs SHA
+        try:
+            for container in ahash["spec"]["template"]["spec"]["containers"]:
```
nemacysts, line 1867: should we do the same type of `get()` that we do above here? if it was possible for `spec` to not exist in the previous conditional, it’s also possible here, right?

paasta_tools/kubernetes_tools.py
```diff
@@ -1855,20 +1853,26 @@ def sanitize_for_config_hash(
...
-            if key not in CONFIG_HASH_BLACKLIST
-        }
+        ahash = config.to_dict()  # deep convert to dict
         ahash["paasta_secrets"] = get_kubernetes_secret_hashes(
             service=self.get_service(), environment_variables=self.get_env()
         )
+
+        # remove data we dont want used to hash configs
+        # replica count
+        if ahash.get("spec", {}).get("replicas"):
```
nemacysts, line 1863: should this have an `is None` tacked on the end? we wouldn’t delete this key if there are 0 replicas otherwise (unless we never enter this function in that case)

paasta_tools/kubernetes_tools.py
kawaiwanyelp: I re-checked and it isn't possible for the keys to not exist because they are converted from k8s objects, but it is possible for the values to be `None` if they aren't explicitly. I've removed the `KeyError` exception, but left the `TypeError` catch just in case.

paasta_tools/marathon_tools.py
```diff
@@ -755,6 +755,8 @@ def sanitize_for_config_hash(
         ] = self.format_docker_parameters(
             with_labels=False, system_paasta_config=system_paasta_config
         )
+        if "PAASTA_SOA_CONFIGS_SHA" in ahash.get("env", {}):
```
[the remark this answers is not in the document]
kawaiwanyelp: I modified the Marathon stuff because yes tests were failing. However, I'd rather leave tests in until we formally delete everything for Marathon.

paasta_tools/kubernetes_tools.py
nemacysts: I wonder if it'd be too noisy to add a log line when we hit this case

nemacysts (approved):

jfongatyelp (approved):
I do think the new envvar will be confusing to most with deployments that aren't caused by actual changes to their service or configs, but seems like a good answer to this ticket. Could we add a comment somewhere and/or write a document explaining what PAASTA_SOA_CONFIGS_SHA really is (and maybe what it shouldn't be used for?)?

paasta_tools/kubernetes_tools.py
```diff
@@ -1855,20 +1853,26 @@ def sanitize_for_config_hash(
...
-            for key, value in config.to_dict().items()
-            if key not in CONFIG_HASH_BLACKLIST
-        }
-        spec = ahash["spec"]
-        ahash["spec"] = {
-            key: copy.deepcopy(value)
-            for key, value in spec.items()
-            if key not in CONFIG_HASH_BLACKLIST
-        }
+        ahash = config.to_dict()  # deep convert to dict
```
EvanKrall, line 1856: I think I see at least [one case](https://github.com/Yelp/paasta/blob/7eba1996de7a51a016536e2ee4b2f3c9de413c09/paasta_tools/paastaapi/model_utils.py#L1495) where model_to_dict returns a value directly without copying. (i.e. if the model has a list of lists, the outer list will be copied, but the inner lists won't be, so if someone modifies the return value of this, we might actually modify the model object)

It looks like we only ever call this in one place (also in kubernetes_tools.py), where we immediately pass the result to `get_config_hash`, which just converts to JSON and then md5sums the result. I think my concern is purely theoretical. It would make for some spooky failure modes, though.

EvanKrall (approved):
Too bad the generalized `CONFIG_HASH_BLACKLIST` feature isn't useful here.

PR #3270, opened 2022-02-08
Update utils.py to support parallel steps in deploy.yaml

darealMj:
This change proposes to allow deploy.yaml parsing of steps which are in a parallel block. Currently deploy.yamls which have parallel steps throw the following error:

```
File "/opt/venvs/paasta-tools/lib/python3.7/site-packages/paasta_tools/utils.py", line 3012, in 
    pipeline_steps = [step["step"] for step in get_pipeline_config(service, soa_dir)]
KeyError: 'step'
```

The current code change supports formats such as
-pipeline:
  -parallel:
    -step: stepname1
    -step: stepname2
  -step: stepname3
  
 It keeps backward compatibility for non parallelized steps. 
 Test non parallel steps output: https://pastebin.com/FKeKDir5
 Test parallel steps output: https://pastebin.com/hjqU10Fc

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -3013,7 +3013,15 @@ def get_pipeline_config(service: str, soa_dir: str = DEFAULT_SOA_DIR) -> List[Di
 def get_pipeline_deploy_groups(
     service: str, soa_dir: str = DEFAULT_SOA_DIR
 ) -> List[str]:
-    pipeline_steps = [step["step"] for step in get_pipeline_config(service, soa_dir)]
+    pipeline_steps = []
+    for step in get_pipeline_config(service, soa_dir):
+        if(step.get("parallel", False)):
+            for parallel_step in step.get("parallel"):
+                if(parallel_step.get("step", False)):
+                    pipeline_steps.append(parallel_step["step"])
+        else:
+            pipeline_steps.append(step["step"])
+            
     return [step for step in pipeline_steps if is_deploy_step(step)]
 
 


paasta_tools/utils.py
```diff
@@ -3013,7 +3013,15 @@ def get_pipeline_config(service: str, soa_dir: str = DEFAULT_SOA_DIR) -> List[Di
 def get_pipeline_deploy_groups(
     service: str, soa_dir: str = DEFAULT_SOA_DIR
 ) -> List[str]:
-    pipeline_steps = [step["step"] for step in get_pipeline_config(service, soa_dir)]
+    pipeline_steps = []
+    for step in get_pipeline_config(service, soa_dir):
+        if(step.get("parallel", False)):
```
nemacysts, line 3018: if you install/run the pre-commit hooks, it'll probably re-write this to 
```suggestion
        if step.get("parallel", False):
```

(and same below)

that said, should we rely on the default `get()` behavior of returning `None` if the key doesn't exist? doesn't seem like we explicitly need a boolean and None is falsey :)

paasta_tools/utils.py
```diff
@@ -3013,7 +3013,15 @@ def get_pipeline_config(service: str, soa_dir: str = DEFAULT_SOA_DIR) -> List[Di
...
+            for parallel_step in step.get("parallel"):
+                if(parallel_step.get("step", False)):
+                    pipeline_steps.append(parallel_step["step"])
```
nemacysts, line 3021: it might be worth adding a comment here about why we're doing this (and maybe what the config looks like for a `parallel` block? could either have a link here to y/deploy.yaml with how parallel works documented there or update https://paasta.readthedocs.io/en/latest/deploy_groups.html to have an example deploy.yaml with a parallel block)

darealMj: Got it. I may need to add it the paasta doc as the y/link wont be available for checking outside of our org. Thanks for the quick review @nemacysts

nemacysts: no worries about adding the ylink here - we already reference a bunch of ylinks :p (it is nice to have things documented in the open docs, though)

EvanKrall (approved):

paasta_tools/utils.py
```diff
@@ -3013,7 +3013,18 @@ def get_pipeline_config(service: str, soa_dir: str = DEFAULT_SOA_DIR) -> List[Di
 def get_pipeline_deploy_groups(
     service: str, soa_dir: str = DEFAULT_SOA_DIR
 ) -> List[str]:
-    pipeline_steps = [step["step"] for step in get_pipeline_config(service, soa_dir)]
+    pipeline_steps = []
+    for step in get_pipeline_config(service, soa_dir):
+        # added support for parallel steps in a deploy.yaml
+        # parallel steps would break previous functionality as steps arent
+        # expected to be nested in a parallel block
```
nemacysts, line 3020: thoughts on something like
```suggestion
        # steps can show up in one of two ways:
        # 1) inline (e.g., a "top-level" entry like `- step-name-here`)
        # 2) or, in a parallel block - e.g.:
        #    parallel:
        #        - step-name-here
        #        - another-step-name
```

nemacysts: oh, and could we add a test for `get_pipeline_deploy_groups()` that uses this new codepath?

PR #3272, opened 2022-02-10
PAASTA-17547: add kubestage reset script

BeryJu:
Moved https://github.yelpcorp.com/sysgit/puppet/pull/5691 to here since it doesn't make much sense shipping this script with puppet.

### DIFF
--- a/paasta_tools/contrib/reset_kubernetes_cluster.py
+++ b/paasta_tools/contrib/reset_kubernetes_cluster.py
@@ -1,0 +1,219 @@
+#!/usr/bin/python3.7
+import argparse
+from os import environ
+from random import choice
+from subprocess import run
+from textwrap import indent
+from typing import List
+from urllib.parse import urlparse
+
+import yaml
+
+ALLOWED_CLUSTERS = ["kubestage"]
+
+
+def inv_search(*filters: str) -> List[str]:
+    """Search inventory service and return matching hosts"""
+    _filters = [f"-F {item}" for item in filters]
+    return run(["inv", "find"] + _filters, capture_output=True).stdout.decode().split()
+
+
+class KubernetesResetter:
+    """Reset the cluster."""
+
+    dry: bool
+
+    master_node_names: List[str]
+    other_node_names: List[str]
+
+    etcd_prefix: str
+    etcd_host: str
+
+    cluster_name: str
+    cluster_ecosystem: str
+
+    def __init__(self, dry=True) -> None:
+        self.dry = dry
+        self.master_node_names = []
+        self.other_node_names = []
+        with open("/nail/etc/ecosystem", "r+") as _eco:
+            self.cluster_ecosystem = _eco.read().strip()
+        self.prepare_static()
+
+    def prepare_static(self):
+        """Fallback if the API server is unreachable/broken/etc
+        Does not have all features since we don't know which nodes exist"""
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/reset_kubernetes_cluster.py
```diff
@@ -0,0 +1,219 @@
...
+
+    cluster_name: str
+    cluster_ecosystem: str
+
+    def __init__(self, dry=True) -> None:
+        self.dry = dry
+        self.master_node_names = []
+        self.other_node_names = []
+        with open("/nail/etc/ecosystem", "r+") as _eco:
+            self.cluster_ecosystem = _eco.read().strip()
```
TheNavigat, line 40: Why are we reading the ecosystem from the file instead of having it read from stdin? That means that the file will only be executable on nodes in the same ecosystem. Why?

danielpops (commented):
My personal opinion is that this script doesn't really belong in the open source puppet project, at least not with the current implementation that is very specific to internal yelp tools and configurations, s3 buckets, etc (and yeah, I know that the Paasta codebase has some of that already, but I don't think we should keep adding more)

Specifically for the s3 bucket, I would like to strongly push back on hard-coding all these real bucket names in the code, as it is an unnecessary leak of information about our environment to the public.

paasta_tools/contrib/reset_kubernetes_cluster.py
```diff
@@ -0,0 +1,219 @@
...
+        )
+        print("> Running kube-init")
+        self.cmd(self.master_node_names, "sudo -H /nail/sys/bin/kube-init")
+        print("> Untainting master nodes...")
+        for host in self.master_node_names:
+            self.cmd(
+                [host],
+                (
+                    "sudo KUBECONFIG=/etc/kubernetes/admin.conf kubectl taint nodes "
+                    f"{host} node-role.kubernetes.io/master:NoSchedule-"
```
danielpops, line 181: Also including my comment from the internal PR:

```
I would like to avoid taking dependencies on the admin.conf (and therefore, the generic administrative client certificate) for pretty much anything. If humans are expected to use this, I'd prefer that you use your own okta federated credentials for authentication. If this is intended to be run as a service or script, then you need to create a new set of credentials for it, ideally scoped only to the actions and resources it needs access to (which would probably alleviate some of the other aforementioned concerns about accidentally running this on the wrong cluster)
```

paasta_tools/contrib/reset_kubernetes_cluster.py
BeryJu: The script needs access to the kubeadm config, so it has to be run on a kubernetes master, hence the master's ecosystem is assumed.

BeryJu:
@danielpops 
> My personal opinion is that this script doesn't really belong in the open source puppet project, at least not with the current implementation that is very specific to internal yelp tools and configurations, s3 buckets, etc (and yeah, I know that the Paasta codebase has some of that already, but I don't think we should keep adding more)
> 

I agree with you, I only put the script here since PE mentioned it doesn't really make sense in the puppet repo either, at this point I honestly don't know what the best place to put this would be.

> Specifically for the s3 bucket, I would like to strongly push back on hard-coding all these real bucket names in the code, as it is an unnecessary leak of information about our environment to the public.

I just had a quick look and sadly couldn't find any way to dynamically get the bucket name without hardcoding some parts of it; however we are also working on getting rid of the need for a bucket entirely.

PR #3273, opened 2022-02-10
Update deploy_groups

darealMj:
Addition of parallel steps documentation


mattmb:
Not sure why this wasn't merged earlier but afaict it is in use and working and the docs read sensibly to me

PR #3274, opened 2022-02-10
Only expose PAASTA_SOA_CONFIGS_SHA to k8s long-running service containers

kawaiwanyelp:
### Description
I previously made the `PAASTA_SOA_CONFIGS_SHA` env var available to what I thought was only long-running service containers, but because I made the change in `utils.py`, all containers configured through `paasta-tools` got it to. Because I only made sure the new env var doesn't affect config hashing for long-running services, Tron jobs were continuously reconfigured, causing an incident. 

To fix the issue, I've moved the env var setting to happen only for k8s long-running containers, and reverted changes to Marathon (which we no longer use) and `utils.py`.

### Testing
`make test`

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -107,6 +107,7 @@ from kubernetes.client import V2beta2ResourceMetricSource
 from kubernetes.client.models import V2beta2HorizontalPodAutoscalerStatus
 from kubernetes.client.rest import ApiException
 from mypy_extensions import TypedDict
+from service_configuration_lib import read_soa_metadata
 
 from paasta_tools.async_utils import async_timeout
 from paasta_tools.long_running_service_tools import AutoscalingParamsDict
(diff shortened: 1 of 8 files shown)


tests/test_tron_tools.py
```diff
@@ -187,6 +189,9 @@ def test_get_env(self, action_config, executor, monkeypatch):
             else:
                 assert not any([env.get("SPARK_OPTS"), env.get("CLUSTERMAN_RESOURCES")])
 
+        # ensure soa-configs SHA is not acquired to add as env var
+        assert mock_read_soa_metadata.call_count == 0
```
nemacysts, line 193: would the intent here be clearer without a comment if we did something like

```
assert "PAASTA_SOA_CONFIGS_SHA" not in env
```

nemacysts (approved):

tests/test_tron_tools.py
kawaiwanyelp: Yeah makes sense.

paasta_tools/kubernetes_tools.py
```diff
@@ -980,9 +981,21 @@ def should_setup_piscina_prometheus_scraping(self,) -> bool:
...
+        # this is deliberately set here to make sure it is only available for
+        # k8s long-running services. putting this in `InstanceConfig.get_env` will
+        # make it available for all workloads, which will cause big bounces and
+        # continuous reconfiguring every time soa-configs is updated unless the
+        # env var is deliberately excluded from config hashing for those workloads
+        # as well.
+        env["PAASTA_SOA_CONFIGS_SHA"] = read_soa_metadata(soa_dir=self.soa_dir).get(
+            "git_sha", ""
+        )
+
```
EvanKrall, line 995: I'd probably override `get_env` instead:

```
def get_env(self):
    env = super().get_env()
    env["PAASTA_SOA_CONFIGS_SHA"] = ...
    return env
```

kawaiwanyelp: Done.

EvanKrall (approved):

PR #3276, opened 2022-02-15
Begin scaffolding executor: spark tronjobs

nemacysts:
Spark jobs need two separate service accounts: one for the driver that
has access to the k8s api (to manage executor pods) and has the
necessary pod identity annotations (to access S3 and whatnot) + another
for the executors which just has pod identity annotations (since they
only need access to AWS resources - not k8s)

This PR does two things to this end:
* create_or_find_service_account_name can now additionally add a Role
  Binding given a Role name (which should already exist - it will NOT be
  created here)
* sends Tron an additional service account name for use by the driver
  that's created using the above feature

Additionally, since drivers will need to be accessible by executors
running on other boxes, I've toggled on routable ips for these pods

This won't be very usable until https://github.com/Yelp/Tron/pull/860
is deployed though.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -3114,25 +3114,35 @@ _RE_NORMALIZE_IAM_ROLE = re.compile(r"[^0-9a-zA-Z]+")
 
 
 def create_or_find_service_account_name(
-    iam_role: str, namespace: str = "paasta", k8s_role: Optional[str] = None
+    iam_role: str,
+    namespace: str = "paasta",
+    k8s_role: Optional[str] = None,
 ) -> str:
     kube_client = KubeClient()
     # the service account is expected to always be prefixed with paasta- as using the actual namespace
     # potentially wastes a lot of characters (e.g., paasta-nrtsearchservices) that could be used for
     # the actual name
-    sa_name = f"paasta--{_RE_NORMALIZE_IAM_ROLE.sub('-', iam_role)}"
-
-    # it's possible for an IAM role to be used for multiple purposes. Some usages may require a
-    # Kubernetes Role attached to the Service Account (e.g., Spark drivers may access S3 but also
-    # need to manage Spark executor Pods), while "normal" services/batches need a Service Account
-    # with only an IAM role attached.
-    # to support these two usecases, we'll suffix the name of a Service Account with the
-    # Kubernetes Role name to disambiguate between the two.
-    if k8s_role:
-        sa_name = f"{sa_name}--{k8s_role}"
+    if iam_role:  # this is either an empty string or a real role
+        # it's possible for an IAM role to be used for multiple purposes. Some usages may require a
+        # Kubernetes Role attached to the Service Account (e.g., Spark drivers may access S3 but also
+        # need to manage Spark executor Pods), while "normal" services/batches need a Service Account
+        # with only an IAM role attached.
+        # to support these two usecases, we'll suffix the name of a Service Account with the
+        # Kubernetes Role name to disambiguate between the two.
+        if k8s_role:
+            sa_name = f"paasta--{_RE_NORMALIZE_IAM_ROLE.sub('-', iam_role)}--{k8s_role}"
+        else:
+            sa_name = f"paasta--{_RE_NORMALIZE_IAM_ROLE.sub('-', iam_role)}"
+    # until Core ML migrates Spark to use Pod Identity, we need to support starting Spark drivers with a Service Account
+    # that only has k8s access
+    elif not iam_role and k8s_role:
+        sa_name = f"paasta--{k8s_role}"
+    # we should never get here in normal usage, but just in case we make a mistake in the future :)
+    else:
+        raise ValueError("Expected at least one of iam_role or k8s_role to be passed in!")
 
     if not any(
-        sa.metadata.name == sa_name
+        sa.metadata and sa.metadata.name == sa_name
         for sa in get_all_service_accounts(kube_client, namespace)
     ):
         sa = V1ServiceAccount(
@@ -3145,11 +3155,18 @@ def create_or_find_service_account_name(
         )
         kube_client.core.create_namespaced_service_account(namespace=namespace, body=sa)
 
-        # we're expecting that any Role dynamically associated with a Service Account already exist.
-        # at Yelp, this means that we have a version-controlled resource for the Role in Puppet.
-        if k8s_role:
-            # since the Role already exists, we just need to associate it with the Service Account through
-            # a Role Binding
+    # we're expecting that any Role dynamically associated with a Service Account already exists.
+    # at Yelp, this means that we have a version-controlled resource for the Role in Puppet.
+    # and since the Role already exists, we just need to associate it with the Service Account through
+    # a Role Binding
+    if k8s_role:
+        # that said, we still check that there's a RoleBinding every time this function is called so that
+        # we can self-heal if we somehow create a Service Account and then fail to create a Role Binding
+        # due to a transient issue
+        if not any(
+            rb.metadata and rb.metadata.name == sa_name
+            for rb in get_all_role_bindings(kube_client, namespace)
+        ):
             role_binding = V1RoleBinding(
                 metadata=V1ObjectMeta(name=sa_name, namespace=namespace,),
                 role_ref=V1RoleRef(
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -84,6 +84,8 @@ from kubernetes.client import V1PodTemplateSpec
 from kubernetes.client import V1Probe
 from kubernetes.client import V1ReplicaSet
 from kubernetes.client import V1ResourceRequirements
+from kubernetes.client import V1RoleBinding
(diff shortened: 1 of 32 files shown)


nemacysts:
with a config that looks like:
```
job_name_here:
  schedule: "daily 03:00:00"
  monitoring:
    check_that_every_day_has_a_successful_run: true
  actions:
    action_name_here:
      executor: spark
      pool: stable_batch
      aws_credentials: credential_name_here
      spark_args:
        spark.cores.max: 300
        spark.executor.memory: 32g
        spark.executor.instances: 150
        spark.driver.memory: 8g
        spark.driver.maxResultSize: 8g
        spark.sql.shuffle.partitions: 2400
      command: "spark-submit /code/SERVICE/base_batches/spark_etl_base_batch.py
                --feature-config=REMOVED
                --only-compute=REMOVED
                --scratch-path=s3a://REMOVED
                --start-date={shortdate-1}
                --end-date={shortdate-1}"
      retries: 2
```

a `paasta_setup_tron_namespace` dry-run returns: 
```
  job_name_here:
    actions:
      action_name_here:
        annotations:
          paasta.yelp.com/routable_ip: 'true'
        cap_add: []
        cap_drop:
        - SETPCAP
        - MKNOD
        - AUDIT_WRITE
        - CHOWN
        - NET_RAW
        - DAC_OVERRIDE
        - FOWNER
        - FSETID
        - KILL
        - SETGID
        - SETUID
        - NET_BIND_SERVICE
        - SYS_CHROOT
        - SETFCAP
        command: spark-submit /code/REMOVED/base_batches/spark_etl_base_batch.py --feature-config=REMOVED
          --only-compute=REMOVED --scratch-path=s3a://REMOVED --start-date={shortdate-1}
          --end-date={shortdate-1}
        cpus: 1
        disk: 1024
        docker_image: ''
        env:
          ENABLE_PER_INSTANCE_LOGSPOUT: '1'
          PAASTA_CLUSTER: REMOVED
          PAASTA_DEPLOY_GROUP: None
          PAASTA_DOCKER_IMAGE: ''
          PAASTA_GIT_SHA: ''
          PAASTA_INSTANCE: job_name_here.action_name_here
          PAASTA_INSTANCE_TYPE: spark
          PAASTA_MONITORING_TEAM: REMOVED
          PAASTA_RESOURCE_CPUS: '1'
          PAASTA_RESOURCE_DISK: '1024'
          PAASTA_RESOURCE_MEM: '4096'
          PAASTA_SERVICE: REMOVED
        executor: spark
        extra_volumes: []
        labels:
          paasta.yelp.com/cluster: REMOVED
          paasta.yelp.com/instance: job_name_here.action_name_here
          paasta.yelp.com/pool: stable_batch
          paasta.yelp.com/service: REMOVED
          yelp.com/owner: REMOVED
        mem: 4096
        node_selectors:
          yelp.com/pool: stable_batch
        retries: 2
        secret_env: {}```

note: the command doesn't include the spark args and the `PAASTA_RESOURCE_*` env vars are the default values for paasta instances right now

File: paasta_tools/kubernetes_tools.py [hunk 1]
```diff
@@ -3124,6 +3144,27 @@ def create_or_find_service_account_name(
...
+                subjects=[
+                    V1Subject(
+                        kind="ServiceAccount", namespace=namespace, name=sa_name,
+                    ),
+                ],
+            )
+            kube_client.rbac.create_namespaced_role_binding(
+                namespace=namespace, body=role_binding
+            )
+
```
jfongatyelp, line 3167: not a blocker, just thinking out loud -- Since this role binding is only created where we're also creating the service account, do you think there's anything we should do here to keep a close eye on whether we somehow create the service account successfully but fail to create the role binding? 

It seems like we could get into a place where the rolebinding is missing and the setup_tron_namespace job would never complain about it.

If a tronjob starts yelling about privs, the manual workaround/fix would be to delete the service account which seems pretty manageable to start though.

paasta_tools/tron_tools.py
```diff
@@ -617,7 +644,14 @@ def format_tron_action_dict(action_config: TronActionConfig, use_k8s: bool = Fal
...
         # log streams automatically
-        result["env"]["ENABLE_PER_INSTANCE_LOGSPOUT"] = "1"
+        # however, we're missing infrastructure in the superregion where we run spark jobs (and
+        # some normal tron jobs), so we take a slightly different approach here
+        if _use_suffixed_log_streams_k8s():
+            result["env"]["STREAM_SUFFIX_LOGSPOUT"] = (
+                "spark" if executor == "spark" else "tron"
+            )
+        else:
+            result["env"]["ENABLE_PER_INSTANCE_LOGSPOUT"] = "1"
```
jfongatyelp, line 654: Oh TIL this is already available ++

paasta_tools/tron_tools.py
```diff
@@ -651,10 +685,24 @@ def format_tron_action_dict(action_config: TronActionConfig, use_k8s: bool = Fal
             and action_config.get_iam_role()
             and not action_config.for_validation
         ):
+            # this service account will be used for normal Tron batches as well as for Spark executors
             result["service_account_name"] = create_or_find_service_account_name(
-                iam_role=action_config.get_iam_role(), namespace="tron"
+                iam_role=action_config.get_iam_role(),
+                namespace=EXECUTOR_TYPE_TO_NAMESPACE[executor],
```
jfongatyelp, line 691: Right now EXECUTOR_TYPE_TO_NAMESPACE maps `paasta` executors->`paasta` namespace, but here we see we used to create them in `tron` namespace, not `paasta`. 

Shoudl the `EXECUTOR_TYPE_TO_NAMESPACE` map get updated to return the tron namesapce instead?

jfongatyelp (approved):
I'm a little confused how the namespaces work right now, and one minor bug I think, but lgtm!

Since we're special-casing spark as its own executor on the spark side, is that where we'll set the right namespace shenanigans to go to `paasta-spark` instead of `tron`?

It might be more obvious if we just pass `namespace` in the generated tronfig instead of these hardcoded settings here and in the tron/taskproc namespace (wherever the `executor: spark` handling is added), but that should not be a blocker, just something i'd need to keep in mind when looking at this later!

paasta_tools/kubernetes_tools.py
nemacysts: good catch - lemme refactor this a bit so that we always check that both the Service Account and the RoleBinding exist

might as well have computers deal with this rather than have words in a runbook that you have to know exist :p

paasta_tools/tron_tools.py
nemacysts: the per-instance logspout stuff has been running for a while - its just the suffix stuff that's new (and I just need to actually deploy the new logspout version to enable it + i guess make tron look at it :p)

paasta_tools/tron_tools.py
nemacysts: doh, yup - i wasn't thinking too good here. lemme also add a test for this if its not too painful

nemacysts:
> I'm a little confused how the namespaces work right now, and one minor bug I think, but lgtm!
> 
> Since we're special-casing spark as its own executor on the spark side, is that where we'll set the right namespace shenanigans to go to `paasta-spark` instead of `tron`?
> 
> It might be more obvious if we just pass `namespace` in the generated tronfig instead of these hardcoded settings here and in the tron/taskproc namespace (wherever the `executor: spark` handling is added), but that should not be a blocker, just something i'd need to keep in mind when looking at this later!

@jfongatyelp we currently just hardcode the namespace on the tron side of things (since up until now we were only ever going to use the tron namespace AND the spark-specific stuff in tron hasn't been implemented yet), but I really like the idea of including the namespace as part of the payload.

lemme make a new ticket so that I can actually get this PR out first without blocking on tron :D

paasta_tools/tron_tools.py
nemacysts: actually, lemme write a test for this once i refactor to send the namespace to tron since that'll make this easier :p

jfongatyelp (approved):

chelspea:
just to check whether I understood the intention correctly, so for spark users they would have a service account (deployed in terraform?) on a team by team basis which would be used for the executors and then the service account for the driver is created on the fly?
one other question, driver needs to have a port exposed externally (or some sort of address) for accessing spark UI by normal browser wonder if it's already in or you might add that at some point?
But otherwise LGTM!!

chelspea (approved):

nemacysts:
> just to check whether I understood the intention correctly, so for spark users they would have a service account (deployed in terraform?) on a team by team basis which would be used for the executors and then the service account for the driver is created on the fly? one other question, driver needs to have a port exposed externally (or some sort of address) for accessing spark UI by normal browser wonder if it's already in or you might add that at some point? But otherwise LGTM!!

@chelspea we'll handle creating the service account with the code in this PR - the only thing that needs to exist already is the spark role (which we manage with puppet right now) and, once y'all update the yelpy spark boilerplate to be able to use pod identity, there'll be some terraform changes needed to allow the iam roles that people want to use to be usable for that purpose (but we can get to that later)

i am also planning to add the port stuff in a later pr :)

jfongatyelp (approved):

PR #3277, opened 2022-02-15
DREIMP-7929: add pod start time to Cassandra status writer

siadat:
Display node's StartTime, when this value is available. We want to make sure nothing breaks even when StartTime is not available. We have added a test for this scenario as well.

Context: DREIMP-7929

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1843,6 +1843,9 @@ def print_cassandra_status(
                 row["Initialized"] = "Yes" if details.get("initialized") else "No"
                 row["Drained"] = "Yes" if details.get("drained") else "No"
                 row["Draining"] = "Yes" if details.get("draining") else "No"
+                # Show absolute dates:
+                row["StartTime"] = node.get("startTime", "None")
+                row["InspectedAt"] = node.get("inspectTime", "None")
             if verbose > 1:
                 row["LocalHostID"] = details.get("localHostId")
                 row["Schema"] = details.get("schemaVersion")
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1790,31 +1790,31 @@ def print_cassandra_status(
     output.append(indent * tab + "Nodes:")
     indent += 1
     now = datetime.now(timezone.utc)
-    tableOkNodes = []
-    tableErrNodes = []
+    table_ok_nodes = []
+    table_err_nodes = []
 
     for node in nodes:
         ip = node.get("ip")
         err = node.get("error")
-        startAge = "None"
+        start_age = "None"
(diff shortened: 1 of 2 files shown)


siadat:
cc @ymuzammil

paasta_tools/cli/cmds/status.py
```diff
@@ -1796,24 +1796,35 @@ def print_cassandra_status(
     for node in nodes:
         ip = node.get("ip")
         err = node.get("error")
+        startAge = "None"
```
atallahade, line 1799: nit: we tend to use snake_case for variable names

siadat: Good catch. I converted them and pushed the change as a new commit.

atallahade (commented):
Would it make sense to show the absolute start date when using `-v` or `-vv`? I'm not sure if this is very useful, and you can always fetch that information elsewhere, but let me know what you think

siadat:
I guess it depends on how we are going to use the output. I can think of 2 advantages:

* Absolute time is more script-friendly
* Absolute time has more precision

We could experiment with it. I pushed a new commit with that change.

siadat:
@atallahade I added a a few tests for the -v and -vv modes. For convenience, you can also view them here:
* [output for -v](https://fluffy.yelpcorp.com/i/7QCx47rJgXsKH8M37ndj6l3t7Skcx1dD.html)
* [output for -vv](https://fluffy.yelpcorp.com/i/kqdbV2m7BDctk9ZHzQk0k0PgdkDJWHFt.html)

atallahade:
> @atallahade I added a few tests for the -v and -vv modes. The tests will be shorter and simpler after we refactor how we provide status information to Paasta. For convenience, you can also view them here:
> 
> * [output for -v](https://fluffy.yelpcorp.com/i/7QCx47rJgXsKH8M37ndj6l3t7Skcx1dD.html)
> * [output for -vv](https://fluffy.yelpcorp.com/i/kqdbV2m7BDctk9ZHzQk0k0PgdkDJWHFt.html)

I was interested in seeing the start time format, but in this example it shows `None`. Is this because you need the operator change to be deployed first? Is there another way to see how the start time would look like?

siadat:
@atallahade Here's a better output example for all verbosity levels (this is generated in the tests): [output](https://fluffy.yelpcorp.com/i/FbVR6kVkSBjlL7KpdMjMB0GXVj7NB8Z4.html)

(Btw, the failing Tox pipeline is unrelated to this PR, [another](https://github.com/Yelp/paasta/pull/3282) PR is failing at that stage too.)

siadat:
We need to wait for https://github.com/Yelp/paasta/pull/3281 to be merged and rebase for the pipelines to succeed.

atallahade (approved):

PR #3278, opened 2022-02-15
Update deploy_scheme to allow parallel keyword

darealMj:
An addition to https://github.com/Yelp/paasta/pull/3270 which will allow the keyword parallel to be seen as valid.

This is to facilitate #cid rolling out parallel steps to deploy.yamls.

This was tested on: https://github.yelpcorp.com/packages/py-gitolite/pull/51 (see test comment from Timmy)


dbgrigsby (approved):

anjaberens (approved):

tests/cli/test_cmds_check.py
```diff
@@ -285,6 +287,51 @@ def test_check_smartstack_check_is_ok_when_no_smartstack(mock_is_file_in_dir, ca
...
+                },
+                {
+                    'step': 'somesecurity.check',
+                },
+            ],
+            'step': 'anothercluster.something',
+            'wait_for_deployment': False,
+            'timeout': 3600,
+        }],
+    }
```
nemacysts, line 329: if we wanted to get real fancy, we could do something like this to make this test a little more modern (compared to the others in this file) by using pytest parametrization + mocking with a context manager:
(note - i wrote this in the github text field so the formatting is all messed up :p)
```suggestion
@pytest.mark.parametrize("config_dict,is_valid", (
        ({
        'pipeline': [{
             'parallel': [
                 {
                     'step': 'somecluster.something',
                 },
                 {
                     'step': 'somesecurity.check',
                },
             ],
             'step': 'anothercluster.something',
             'wait_for_deployment': False,
             'timeout': 3600,
          }],
    }, True),
    # invalid config - not expecting another key to have steps nested inside
    ({
        'pipeline': [{
             'not-parallel': [
                 {
                     'step': 'somecluster.something',
                 },
                 {
                     'step': 'somesecurity.check',
                },
             ],
             'step': 'anothercluster.something',
             'wait_for_deployment': False,
             'timeout': 3600,
          }],
    }, False),
  )
)
def test_check_parallel_schema(config_dict, is_valid):
    # Parallel steps is seen as valid
    with mock.patch("paasta_tools.cli.cmds.validate.get_config_file_dict", autospec=True, return_value=config_dict):
        assert validate_schema('my/fake-service','deploy') is is_valid
```
but this is just a suggestion - feel free to merge without doing this

tests/cli/test_cmds_check.py
```diff
@@ -285,6 +287,51 @@ def test_check_smartstack_check_is_ok_when_no_smartstack(mock_is_file_in_dir, ca
...
+
+    actual = validate_schema('my/fake-service','deploy') is True
+    assert actual is True
+
+@patch("paasta_tools.cli.cmds.validate.get_config_file_dict", autospec=True)
+def test_check_parallel_fails(mock_get_config_file_dict):
+    # Parallel steps is seen as valid
```
nemacysts, line 314: if we don't go for the suggestion below, this comment is inaccurate (looks like it wasn't changed when basing it off the test above?)

tests/cli/test_cmds_check.py
```diff
@@ -285,6 +287,51 @@ def test_check_smartstack_check_is_ok_when_no_smartstack(mock_is_file_in_dir, ca
...
+                    'step': 'somesecurity.check',
+                },
+            ],
+            'step': 'anothercluster.something',
+            'wait_for_deployment': False,
+            'timeout': 3600,
+        }],
+    }
+
+    actual = validate_schema('my/fake-service','deploy') is True
```
nemacysts, line 309:
```suggestion
    assert validate_schema('my/fake-service','deploy') is True
```
would be a little more idiomatic - especially since this is just one line and we're not using the value again to test for anything else (same thing in the failure-case test if we don't go for my proposed test change)

paasta_tools/cli/schemas/deploy_schema.json
```diff
@@ -125,8 +144,5 @@
             "type": "boolean"
         }
     },
-    "required": [
-        "pipeline"
-    ],
```
nemacysts, line 130: was this meant to be deleted? i think we want to enforce that there's no empty deploy.yaml files

darealMj: I was just here puzzling over why empty deploys started to pass. Yep it was an accident

tests/cli/test_cmds_check.py
darealMj: I actually like this fanciness. The only slight change is the return type for the failure would actually be None

tests/cli/test_cmds_check.py
darealMj: Yeah i like this fanciness. The only difference is the failing test should be None instead of False

dbgrigsby (approved):

tests/cli/test_cmds_check.py
```diff
@@ -285,6 +287,49 @@ def test_check_smartstack_check_is_ok_when_no_smartstack(mock_is_file_in_dir, ca
...
+            'wait_for_deployment': False,
+            'timeout': 3600,
+        }],
+    }
+
+    assert validate_schema('my/fake-service','deploy') is True
+
+@patch("paasta_tools.cli.cmds.validate.get_config_file_dict", autospec=True)
+def test_check_parallel_fails(mock_get_config_file_dict):
+    # Parallel steps is seen as invalid
```
nemacysts, line 313: super-nit: 
```suggestion
    # only "parallel" is allowed to have nested steps inside it
```

nemacysts (approved):

nemacysts (approved):

PR #3280, opened 2022-02-15
Use stdlib json to avoid ujson CVE

lsterk:
https://security.snyk.io/vuln/SNYK-PYTHON-UJSON-2359034 is an open CVE for the `ujson` python package. Although it looks like the library's maintainers are working on a PR for this, I don't suspect the value that `ujson` gives us justifies yet-another dependency that needs constant tending.

Using the `stdlib` is more common now than 6 years ago, when the built-in JSON library was slower; I suspect there are some performance gains we might be giving up, but avoiding libraries that rely on C-extensions also can simplify builds.


nemacysts (approved):
lgtm - i think we could easily replace ujson with orjson, but we can do that down the line

lsterk:
I'll wait for #3281 to get merged before releasing this (patch release), just to make sure the build is clean before I try to change anything.

PR #3281, opened 2022-02-15
Make paasta check-requirements clean

nemacysts:
We weren't actually running check-requirements in a way that blocked
things, so we had a decent chunk of unpinned requirements (both for dev
and real dependencies).


Makefile
```diff
@@ -74,7 +74,7 @@ clean:
 	-rm -rf .paasta
 
 yelpy: ## Installs the yelp-internal packages into the default tox environment
-	.tox/py37-linux/bin/pip-custom-platform install -i https://pypi.yelpcorp.com/simple -r yelp_package/extra_requirements_yelp.txt -r ./extra-linux-requirements.txt
```
nemacysts, line 77: pretty sure that `-r ./extra-linux-requirements.txt` is no longer needed - maybe an artifact of a bygone era when we maybe cared about running on not-linux (our laptops?)

requirements-minimal.txt
```diff
@@ -10,7 +10,6 @@ certifi
 choice >= 0.1
 cookiecutter >= 1.4.0
 croniter
-dataclasses >= 0.7; python_version < '3.7'
```
nemacysts, line 13: we use py37 - no need to conditionally include this backport :p

requirements.txt
```diff
@@ -3,7 +3,7 @@ addict==2.1.0
 aiohttp==3.7.4.post0
 argcomplete==0.8.1
 argparse==1.4.0
-arrow==0.12.1
+arrow==1.1.0
```
nemacysts, line 6: 0.12.1 was never used internally since we had things in our yelpy requirements that required newer versions

nemacysts (commented):
for all the new pins, I used what was downloaded on my devbox - `make test itest itest_bionic` all passed locally

nemacysts:
that said, we really should go one step further here and figure out a good way to make the yelpy requirements work with check-requirements

nemacysts:
hmm, looks like something might still be unpinned...

Makefile
```diff
@@ -74,7 +74,7 @@ clean:
...
+	.tox/py37-linux/bin/pip-custom-platform install -i https://pypi.yelpcorp.com/simple -r yelp_package/extra_requirements_yelp.txt
```
lsterk, line 77: It seems like the proper thing would be to put this in the tox file?

requirements-minimal.txt
lsterk: This wouldn't be true for package-based use cases; this file is used to populate the `install_requires` in `setup.py`, and there isn't a requirement that users of the `paasta-tools` package use py37 (e.g. `python_requires='>=3.7.0'`

yelp_package/extra_requirements_yelp.txt
```diff
@@ -1,22 +1,41 @@
 --index-url=https://pypi.yelpcorp.com/simple
+atomicfile==1.0
 cached-property==1.3.1
+cffi==1.15.0
+cffi==1.15.0
 clusterman_metrics==2.2.0
-crypto_lib==4.0.0
+crypto-lib==4.0.0
+cryptography==2.8
```
lsterk, line 8: is there a reason that we have to use such an old version? It might be worth ticketing this up, I'm actually surprised such an old version would work with newer libssl versions

Makefile
lsterk: (not that this is something required here, but seeing `.tox/py37-linux/bin/pip-custom-platform install` makes me cringe. It would be tricky to configure the alternative PyPI server without having a completely duplicated `tox` target)

requirements-minimal.txt
nemacysts: ugh, i forgot that we use paasta as a library in some places ;_; - lemme add this case explicitly in our setup.py since I don't want this to snowball into dropping support for a. python version that something might still be using internally (and that deps paasta)

yelp_package/extra_requirements_yelp.txt
nemacysts: i get the feeling that we just bump it whenever it causes an issue - lemme see what we can bump up to

Makefile
nemacysts: yea, i've been meaning to try to clean this repo's toxfile up when i have a sec :/

moving this to an optional step in our tox sounds good tho - i always get annoyed when i recreate my venv and forget to run this target :p

requirements-minimal.txt
nemacysts: actually, it looks like there's only one usecase at yelp that uses py36 and has paasta as a dep so i'm gonna go ahead and add a python_requires to gently nudge them :p

yelp_package/extra_requirements_yelp.txt
nemacysts: 3.0 works! (and is the highest we can bump to atm)

lsterk (approved):

ankit864 (approved):

Makefile
nemacysts: wow, having yelpy deps available broken a bunch of tests ;_;

EvanKrall (commented):
everything else seems fine if tests pass. The deleted tests are for dead functionality or never ran anyway.

tox.ini
```diff
@@ -19,11 +18,18 @@ deps =
...
+    # manually do so (through pip or make)
+    -pip-custom-platform install -i https://pypi.yelpcorp.com/simple -r yelp_package/extra_requirements_yelp.txt
 
 [testenv:tests]
 envdir = .tox/py37-linux/
 commands =
+    # these are only available at yelp so we optionally install them so that internal devs don't need to
+    # manually do so (through pip or make)
+    -pip-custom-platform install -i https://pypi.yelpcorp.com/simple -r yelp_package/extra_requirements_yelp.txt
+    check-requirements -vv
```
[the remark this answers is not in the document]
nemacysts: adding an additional env sounds like a good idea, lemme do that real fast

from a quick search it doesn't seem like we can really do any real logic for the default envlist, but i dunno how often people are running just a bare `tox`, so we can just pick an option imo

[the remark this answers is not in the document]
nemacysts: ok, I think https://github.com/Yelp/paasta/pull/3281/commits/ecf039a6d108e029e5aec1894be8e6ac99f256dd should be what we want?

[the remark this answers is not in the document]
nemacysts: (with https://github.com/Yelp/paasta/pull/3281/commits/7a2292fb0a66313b4ab94eff4eaf5efe898b0e76 on top of it, of course)

EvanKrall (approved):

PR #3282, opened 2022-02-16
[COREML-2697] spark-run defaults to kubernetes

gcollura:
## Changes

* make `kubernetes` the default option for `--cluster-manager`
* introduce `--disable-compact-bin-packing` which defaults to False
* make `--enable-compact-bin-packing` an always-True option which is basically a dummy option.
* update tests

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -145,7 +145,19 @@ def add_subparser(subparsers):
     list_parser.add_argument(
         "-e",
         "--enable-compact-bin-packing",
-        help="Enabling compact bin packing will try to ensure executors are scheduled on the same nodes. Requires --cluster-manager to be kubernetes.",
+        help=(
+            "Enabling compact bin packing will try to ensure executors are scheduled on the same nodes. Requires --cluster-manager to be kubernetes."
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -145,7 +145,19 @@ def add_subparser(subparsers):
     list_parser.add_argument(
         "-e",
         "--enable-compact-bin-packing",
-        help="Enabling compact bin packing will try to ensure executors are scheduled on the same nodes. Requires --cluster-manager to be kubernetes.",
+        help=(
+            "Enabling compact bin packing will try to ensure executors are scheduled on the same nodes. Requires --cluster-manager to be kubernetes."
+            " Always true by default, keep around for backward compability."
```
nemacysts, line 150: are we planning on removing this flag once it’s no longer referenced in soaconfigs?

nemacysts (approved):

paasta_tools/cli/cmds/spark_run.py
gcollura: Yes, at some point we're going to remove all `--enable-compact-bin-packing` from all of yelp-soa, but not today.

88manpreet (approved):

PR #3288, opened 2022-02-21
DREIMP-7988: Use properties list in Cassandra status writer

siadat:
This PR uses the new status "properties" (a list of key-values), as provided by the Cassandra Operator, to print Cassandra status. Here's an example of a properties list:
```python
  "properties": [
    {
      "verbosity": 0,
      "name": "IP",
      "value": "10.93.210.204",
      "type": "string",
    },
    {
      "verbosity": 0,
      "name": "Available",
      "value": True,
      "type": "bool",
    },
  ],
```

Which would be equivalent to this existing (to be deprecated) structure:
```python
  {
    "ip": "10.93.210.204",
    "available": True,
  }
```

While the old mechanism looks shorter, it is less flexible, and the information that it is neglecting is hard coded in Paastas status code (eg "ip" is displayed as "IP" and "startTime" can be parsed as a time value). Furthermore, adding, removing and modifying a new property requires coordination between the producer (operator) and the consumer (paasta). The new properties list will shrink the radius of change necessary whenever we want to add/remove/modify a property.

## Backward compatibility
This PR is backward compatible. It ensures that it uses the old status mechanism (a dict with predefined keys) when the new properties list is not available. We haven't removed the tests for the old mechanism. The old mechanism and its tests are marked as deprecated and can later be deleted once the new properties list is available for all clusters (DREIMP-7953).

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1789,137 +1789,197 @@ def print_cassandra_status(
 
     output.append(indent * tab + "Nodes:")
     indent += 1
-    now = datetime.now(timezone.utc)
-    table_ok_nodes = []
-    table_err_nodes = []
+    all_rows: List[CassandraNodeStatusRow] = []
 
     for node in nodes:
-        ip = node.get("ip")
-        err = node.get("error")
-        start_age = "None"
-        if node.get("startTime"):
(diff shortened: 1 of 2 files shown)


siadat:
@atallahade @ymuzammil Please review

atallahade (approved):

PR #3289, opened 2022-02-24
[COREML-2697] spark-run defaults to kubernetes

gcollura:
Well this was a bit embarassing..

_I can't believe I missed that line.._


nemacysts (approved):
so did all of us that reviewed the initial PR :p

PR #3290, opened 2022-03-01
PAASTA-17560: Update brutal deployment handling to background delete …

jfongatyelp:
…deployments and always force delete pods

With 'brutal' paasta deployments, we attempt to fully delete the k8s deployment and related pods, then recreate.

We used to delete the deployment w/ foreground deletion of dependent objects, wait 60s, and if the deployment still existed, we assumed this was due to pods still needing to be deleted.  After force-deleting pods, we assumed we were successful and continued with creation.  However, this is dependent on the garbage collector actually kicking in to clean up, which can lead to problems when GC doesn't occur fast enough.

This changes the behavior to delete the deployment w/ background cascading deletion which has the same effect (allowing garbage collection to delete dependent objects) but causes the deployment resource to be deleted immediately (https://v1-19.docs.kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#background-cascading-deletion), then forcibly delete pods ourselves.  This also adds a check to confirm the deployment is deleted prior to attempting to recreate it, to minimize the number of failed creations.

I'm not entirely sure how dangerous this change could be. the 1.19 changelog doesn't indicate any major changes in garbage collection or deletion propagation policies, so this change should work on both versions, and our version of the clientlib does support the `Background` propagation policy: https://github.com/kubernetes-client/python/blob/v18.20.0/kubernetes/docs/V1DeleteOptions.md

### DIFF
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -244,10 +244,10 @@ class DeploymentWrapper(Application):
                 )
             except ApiException as e:
                 if e.status == 404:
-                    # Deployment does not exist, nothing to delete but
-                    # we can consider this a success.
+                    # Pod(s) may have been deleted by GC before we got to it
+                    # We can consider this a success
                     self.logging.debug(
-                        "not deleting nonexistent deploy/{} from namespace/{}".format(
+                        "pods already deleted for {} from namespace/{}. Continuing.".format(
                             self.kube_deployment.service, self.item.metadata.namespace
                         )
                     )
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -2,6 +2,7 @@ import logging
 import threading
 from abc import ABC
 from abc import abstractmethod
+from time import sleep
 from typing import Optional
 from typing import Union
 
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -216,37 +217,31 @@ def create(self, kube_client: KubeClient) -> None:
...
-                if e.status == 404:
-                    # Deployment does not exist, nothing to delete but
-                    # we can consider this a success.
-                    self.logging.debug(
-                        "not deleting nonexistent deploy/{} from namespace/{}".format(
-                            self.kube_deployment.service, self.item.metadata.namespace
-                        )
+        # When deleting then immediately creating, we need to use Background deletion to ensure we can create the deployment immediately
+        self.deep_delete(kube_client, propagation_policy="Background")
+        # Also forcibly delete remaining pods
```
kawaiwanyelp, line 222: nit: We are losing the 60s wait on GC pod deletion, but that's okay since this fn is only called for brutal bounces. I would add a comment here, but not that impt.

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -216,37 +217,31 @@ def create(self, kube_client: KubeClient) -> None:
...
+                self.kube_deployment.instance,
+                self.item.metadata.namespace,
+                kube_client,
+            )
+        except ApiException as e:
+            if e.status == 404:
+                # Deployment does not exist, nothing to delete but
+                # we can consider this a success.
+                self.logging.debug(
+                    "not deleting nonexistent deploy/{} from namespace/{}".format(
```
kawaiwanyelp, line 236: This log line needs to change since the exception now only means that we failed to delete all pods, not that deployment deletion did not happen fast enough. That case is now in the check that you added below.

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -216,37 +217,31 @@ def create(self, kube_client: KubeClient) -> None:
         self.sync_horizontal_pod_autoscaler(kube_client)
 
     def deep_delete_and_create(self, kube_client: KubeClient) -> None:
```
kawaiwanyelp, line 219: Are there any tests that need updating?

kawaiwanyelp (changes requested):
I think this change is fine since it more closely follows the brutal bounce contract, but it might be worth digging around for k8s GC/controller logs to see what the hold up is for these background deletions. Any kind of resource deletion, even background, should be fast enough. 

I don't think the change is scary enough to warrant needing to test on stage, since this function is only called for brutal, and there are only a small number of instances that we have manual steps to remedy.

paasta_tools/kubernetes/application/controller_wrappers.py
jfongatyelp: Hm, this is a good point; we may have some services that use brutal but still need to have a <60s grace period. I originally thought this was strictly to let GC do its job, rather than allowing for graceful pod deletions.

However we've seen the problem is that GC may not kick in immediately after attempting the deployment deletion anyway -- should we instead try to delete pods, wait 60s, then force delete pods to be more explicit?

paasta_tools/kubernetes/application/controller_wrappers.py
jfongatyelp: I don't really understand this log line to begin with (or the comment in `force_delete_pods` indicating we needed to have deleted the deployment before hand) -- we are catching a 404 when attempting to delete pods by paasta.yelp.com/service and paasta.yelp.com/instance labels, rather than relating to the deployment object, so this would only throw a 404 if the pods were deleted.

However the effect is basically the same -- the pods probably got deleted before we got to this part, so we can consider it a success. I'll update to reflect this.

jfongatyelp: Oh, well...now we see why I didn't see API requests for deleting pods when setting up these brutal deployments; we're invoking force_delete_pods incorrectly here, sending the deployment name as the service name. I will fix this bug as well.

paasta_tools/kubernetes/application/controller_wrappers.py
jfongatyelp: Existing tests were unaffected, but I've added another to make sure we check this new logic

nemacysts (approved):

PR #3291, opened 2022-03-07
Fix pre-commit failures

nemacysts:
Not quite sure how these snuck by - are we not running pre-commit in
GHA?


darealMj:
++ nemacysts and thanks!

PR #3292, opened 2022-03-07
Run pre-commit as part of make test + readd general_itests

nemacysts:
We've had a couple pre-commit failures sneak in since `make test`
doesn't install/run hooks (and none of the toxenvs run in GHA do so
either)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
       matrix:
         toxenv:
           - py37-linux,docs,mypy,tests
+          - general_itests
     env:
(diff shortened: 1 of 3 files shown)


EvanKrall:
Are we not running `general_itests` on github actions?

nemacysts:
> Are we not running `general_itests` on github actions?

@EvanKrall nope, we stopped since deadsnakes dropping xenial broke our test container builds

EvanKrall:
iirc `general_itests` doesn't involve docker, it just runs `behave` from within the tox venv.

nemacysts:
nice, looks like I was over-zealous removing general_itests a while back ago

PR #3293, opened 2022-03-08
Add paasta_cluster to rollback events emitted to sfx

jfongatyelp:
When emitting events, we currently only indicate the deploy_group and service.  

However, SLOs are reported based on service and paasta_cluster, rather than deploy_group, so if we mapped the current events to the existing SLO charts, we'll include these events for some clusters where rollbacks did not actually occur. 

We already have the function `get_instance_configs_for_service_in_deploy_group_all_clusters` which will give us the specific instances by cluster in the requested deploy_group, used in wait_for_deployment when actually monitoring the affected instances.  This moves this function to be called when we initialize a MarkForDeploymentProcess and then uses this information when emitting events.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -611,6 +611,11 @@ class MarkForDeploymentProcess(SLOSlackDeploymentProcess):
         self.diagnosis_interval = diagnosis_interval
         self.time_before_first_diagnosis = time_before_first_diagnosis
         self.metrics_interface = metrics_interface
+        self.instance_configs_per_cluster: Dict[
+            str, List[LongRunningServiceConfig]
+        ] = get_instance_configs_for_service_in_deploy_group_all_clusters(
+            service, deploy_group, commit, soa_dir
+        )
 
         # Keep track of each wait_for_deployment task so we can cancel it.
         self.wait_for_deployment_tasks: Dict[str, asyncio.Task] = {}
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1229,26 +1235,24 @@ def log_slo_rollback(self) -> None:
...
-            name="rollback", dimensions=dimensions,
-        )
+        self._log_rollback(rollback_details)
+
+    def _log_rollback(self, rollback_details: Dict[str, str]) -> None:
+        base_dimensions = dict(rollback_details)
+        base_dimensions["paasta_service"] = self.service
+        # Emit one event per cluster to sfx
+        for cluster in self.instance_configs_per_cluster.keys():
+            dimensions = dict(base_dimensions)
```
lanceleih, line 1251: minor nit: is it necessary to explicitly declare it as dict()? Or is it more for readability?

jfongatyelp: I needed to copy the base_dimensions dict rather than modify it, so opted for `dict()` here. Is there a better pattern to follow for this?  `base_dimensions.copy()`?

lanceleih: Ah no no. That's correct. By adding dict(), it will make base_dimentions' own dict rather than a reference of rollback_details.

lanceleih (approved):

PR #3294, opened 2022-03-08
add special handling for pseudo-service _shared

BeryJu:
PAASTA-17581

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -121,6 +121,11 @@ def get_services_to_k8s_namespaces(
 ) -> Mapping[str, Set[str]]:
     services_to_k8s_namespaces: Mapping[str, Set[str]] = defaultdict(set)
     for service in service_list:
+        # Special handling for service `_shared`, since it doesn't actually exist
+        if service == "_shared":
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -121,6 +121,11 @@ def get_services_to_k8s_namespaces(
 ) -> Mapping[str, Set[str]]:
     services_to_k8s_namespaces: Mapping[str, Set[str]] = defaultdict(set)
     for service in service_list:
+        # Special handling for service `_shared`, since it doesn't actually exist
+        if service == "_shared":
+            services_to_k8s_namespaces[service].add(
+                INSTANCE_TYPE_TO_K8S_NAMESPACE["kubernetes"]
+            )
```
Rob-Johnson, line 128: maybe add a `break` after this?

Rob-Johnson (commented):
can you write a test so we don't accidentally see this regression again?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
BeryJu: it doesn't hurt if the other lookup runs too, but true

Rob-Johnson (approved):

PR #3297, opened 2022-03-08
Teach paasta cook-image --commit

nemacysts:
CID would like this so that Jenkins can skip itesting and just cook an
image if tests have already passed pre-merge.

Unless --commit is passed, paasta cook-image will behave as it always
has.


nattofriends (approved):
I like it

cuza (approved):

EvanKrall (approved):

PR #3299, opened 2022-03-08
Add hash limit function to boto_keys secret names

Qmando:
when the service and instance name are long enough, this can make the secret name hit k8s' 63 character limit, resulting in an error:
```
"reason":"FieldValueInvalid","message":"Invalid value: \"secret-boto-key-service-name-instance-name-blah-blah\": must be no more than 63 characters"
```

This should fix the issue by using the limit_size_with_hash function provided for exactly this purpose. I'm applying it to both the secret name and the volume name.

Tests pass, though I haven't done any additional testing.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1324,7 +1324,9 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             return None
         secret_name = limit_size_with_hash(f"paasta-boto-key-{service_name}")
         volume = V1Volume(
-            name=limit_size_with_hash(f"secret-boto-key-{service_name}"),
+            name=self.get_sanitised_volume_name(
+                f"secret-boto-key-{service_name}", length_limit=253
+            ),
             secret=V1SecretVolumeSource(
                 secret_name=secret_name, default_mode=mode_to_int("0444"), items=items,
             ),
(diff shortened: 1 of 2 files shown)


danielpops (approved):

paasta_tools/kubernetes_tools.py
```diff
@@ -1378,7 +1378,7 @@ def get_volume_mounts(
             if secret_hash:
                 mount = V1VolumeMount(
                     mount_path="/etc/boto_cfg",
-                    name=f"secret-boto-key-{service_name}",
+                    name=limit_size_with_hash(f"secret-boto-key-{service_name}"),
```
nemacysts, line 1381: I think for the volume names here (the V1Volume and the V1VolumeMount) we want to use [get_secret_volume_name](https://sourcegraph.yelpcorp.com/Yelp/paasta@f26daac26d5edd2091e2c75ff7f29b5c2e81a0f4/-/blob/paasta_tools/kubernetes_tools.py?subtree=true#L835:1) so that we ensure that we do all the other normalization stuff to comply with the restrictions on what volume names can look like

paasta_tools/kubernetes_tools.py
```diff
@@ -1322,9 +1322,11 @@ def get_boto_volume(self) -> Optional[V1Volume]:
         if not secret_hash:
             log.warning(f"Expected to find k8s secret {secret_name} for boto_cfg")
             return None
-        secret_name = f"paasta-boto-key-{service_name}"
+        secret_name = limit_size_with_hash(f"paasta-boto-key-{service_name}")
         volume = V1Volume(
-            name=f"secret-boto-key-{service_name}",
+            name=self.get_sanitised_volume_name(
```
nemacysts, line 1327: we could have also created a `get_boto_key_volume_name` function to be consistent with the other usages of `get_sanitised_volume_name`, but this is also fine imo

nemacysts (approved):

PR #3300, opened 2022-03-09
Add/Use cluster aliases for paasta spark-run

nemacysts:
We're in a bit of a weird situation with spark where we want to run
stuff in a separate sparl-specific cluster, but there's a lot of
assumptions in the paasta codebase where the cluster you're running
stuff on will be populated by {workload}-{cluster}.yaml files - and we
don't really want to embark on migrating all the existing configuration
(since that requires that we migrate/work on a bunch of other stuff like
Tron to "natively" support this as a separate cluster in the same way as
every other cluster works)

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -953,7 +953,9 @@ def paasta_spark_run(args):
         instance_config = get_instance_config(
             service=args.service,
             instance=args.instance,
-            cluster=args.cluster,
+            cluster=system_paasta_config.get_cluster_aliases().get(
+                args.cluster, args.cluster
+            ),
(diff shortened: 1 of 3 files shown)


nemacysts:
this still needs a corresponding puppet change to actually add in an alias for the new spark cluster (as well as certs to be deployed to batch boxes) - but https://fluffy.yelpcorp.com/i/VbKnD2SJR9xBwHRB9m3GbsgkvDJf0wzn.html is what I got with a "dumb" `cluster=DESIRED_CLUSTER if args.cluster == SPARK_CLUSTER else args.cluster` change (of which this is really just a productionized version of)

BeryJu (approved):

chelspea (approved):

PR #3302, opened 2022-03-10
fix flake8 by removing unused imports

bobtfish:
I expected `make test` to be enough to find things :)


BeryJu:
Should really merge https://github.com/Yelp/paasta/pull/3292

PR #3303, opened 2022-03-10
Fix additional pre-commit failures + fix oapi hook

nemacysts:
There were some other PRs merged in between the last pre-commit fix and
getting pre-commit running in GHA again

We were also running check_manual_oapi_changes once for every file whenever
we ran `pre-commit run --all-files`, which was making things unhappy when
the hook was run in parallel (since `make openapi-codegen` mucks with files)

### DIFF
--- a/paasta_tools/contrib/check_manual_oapi_changes.sh
+++ b/paasta_tools/contrib/check_manual_oapi_changes.sh
@@ -5,6 +5,9 @@ diff_names=$(git diff origin/master --name-only)
 schema_touched=$(echo "$diff_names" | grep 'paasta_tools/api/api_docs/oapi.yaml')
 api_touched=$(echo "$diff_names" | grep 'paasta_tools/paastaapi')
 
+# this case will only be hit as part of `make test` or someone running pre-commit on
+# all files manually since we've chosen to only run this hook on changes to the schema
+# file
 if [ ! -z "$api_touched" -a -z "$schema_touched" ]; then
     echo "paasta_tools/paastaapi must not be modified manually" >&2
     echo "Please revert your changes:" >&2
--- a/paasta_tools/contrib/check_manual_oapi_changes.sh
+++ b/paasta_tools/contrib/check_manual_oapi_changes.sh
@@ -1,18 +1,15 @@
 #!/bin/bash
 set -uo pipefail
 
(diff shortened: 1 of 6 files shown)


paasta_tools/contrib/check_manual_oapi_changes.sh
```diff
@@ -1,18 +1,15 @@
...
-        echo "paasta_tools/paastaapi must not be modified manually" >&2
-        echo "Please revert your changes:" >&2
-        echo -e "$api_touched" >&2
-        exit 1
-    fi
+if [ ! -z "$api_touched" -a -z "$schema_touched" ]; then
+    echo "paasta_tools/paastaapi must not be modified manually" >&2
+    echo "Please revert your changes:" >&2
+    echo -e "$api_touched" >&2
+    exit 1
```
jfongatyelp, line 12: It's somewhat counterintuitive to me that this case would only ever get hit during `--all-files` since the pre-commit only be invoked if $schema_touched is non-empty, so I might erroneously think we should delete this case later.

Maybe worth a quick comment?

jfongatyelp (approved):
TY!

paasta_tools/contrib/check_manual_oapi_changes.sh
nemacysts: i'm not the greatest at bash, but from my testing + how i read this: we're hitting this if the api is touched and the schema isn't (and then below we're checking if the schema was changed and the api wasn't) - but i'm probably wrong :p

jfongatyelp: Exactly, and so this means we're only ever going to hit this case when we run with `--all-files`, because we restrict this hook from being triggered _only_ when oapi.yaml was changed normally.  That's the part that is slightly counterintuitive to me.

nemacysts: ooooh, I see what you're saying

nemacysts: added a comment, lemme know if it looks good

jfongatyelp: Perfect thank you!

jfongatyelp (approved):

PR #3305, opened 2022-03-11
Allow issuetype key in tron monitoring configs

nemacysts:
We're deploying a version of tron that includes a pysensu-yelp version
that allows users to set the issuetype for any jira tickets that are
created - but the tron schema is strict and won't allow this to be set
without having a schema entry for the config key

### DIFF
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -431,6 +431,9 @@
                 "monitoring": {
                     "type": "object",
                     "properties": {
+                        "issuetype": {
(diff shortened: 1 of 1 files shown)


cuza (approved):

jfongatyelp (approved):

nemacysts:
I tested this with `paasta validate` locally (added an issuetype key -> paasta validate fails, added an issuetype key + this change -> paasta validate passes)

I did think about writing a test, but we don't seem to have really done that in the past here and have instead just focused on preventing "bad" keys so i skipped writing any :)

PR #3307, opened 2022-03-14
Limit size of service name labels for signatures and volume mount names

Qmando:
First fix attempt: https://github.com/Yelp/paasta/pull/3299/files

This change wasn't sufficient. We encountered an error when setting `boto_keys` for some services saying that the label attached to the ConfigMap object used to store secret signatures is too long:

(Service name changed slightly for the paste)
```
...HTTP response body: {\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"ConfigMap \\\"paasta-secret-servicename-this--one--just--has--a--really--long--instance--name-paasta-boto-key-servicename-this--one--just-yvib-signature\\\" is invalid: metadata.labels: Invalid value: \\\"servicename-this--one--just--has--a--really--long--instance--name\\\": must be no more than 63 characters\",\"reason\":\"Invalid\",\"details\":{\"name\":\"paasta-secret-servicename-this--one--just--has--a--really--long--instance--name-paasta-boto-key-servicename-this--one--just-yvib-signature\",\"kind\":\"ConfigMap\",\"causes\":[{\"reason\":\"FieldValueInvalid\",\"message\":\"Invalid value: \\\"servicename-this--one--just--has--a--really--long--instance--name\\\": must be no more than 63 characters\",\"field\":\"metadata.labels\"},{\"reason\":\"FieldValueInvalid\",\"message\":\"Invalid value: \\\"servicename-thi", ...
```

This happens because we were passing the full deployment name into `service` parameter of *_kubernetes_secret_hash instead of just the service name. 

After I made that change, I encountered another error that the volume name was also too long. I picked 253 previously because that's the limit on other K8s resource names. In this case, Volume names, it can only be 63.

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -321,7 +321,7 @@ def sync_boto_secrets(
         kubernetes_signature = get_kubernetes_secret_signature(
             kube_client=kube_client,
             secret=secret,
-            service=app_name,
+            service=service,
             namespace=namespace,
         )
         if not kubernetes_signature:
@@ -344,7 +344,7 @@ def sync_boto_secrets(
             create_kubernetes_secret_signature(
                 kube_client=kube_client,
                 secret=secret,
-                service=app_name,
+                service=service,
                 secret_signature=signature,
                 namespace=namespace,
             )
@@ -362,7 +362,7 @@ def sync_boto_secrets(
             update_kubernetes_secret_signature(
                 kube_client=kube_client,
                 secret=secret,
-                service=app_name,
+                service=service,
                 secret_signature=signature,
                 namespace=namespace,
             )
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2960,6 +2960,7 @@ def update_kubernetes_secret_signature(
     namespace: str = "paasta",
 ) -> None:
     service = sanitise_kubernetes_name(service)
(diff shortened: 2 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1325,7 +1325,7 @@ def get_boto_volume(self) -> Optional[V1Volume]:
         secret_name = limit_size_with_hash(f"paasta-boto-key-{service_name}")
         volume = V1Volume(
             name=self.get_sanitised_volume_name(
-                f"secret-boto-key-{service_name}", length_limit=253
+                f"secret-boto-key-{service_name}", length_limit=63
```
nemacysts, line 1328: might be nice to add a function like
```Python
def get_boto_secret_volume_name(self, service_name: str) -> str:
    return self.get_sanitised_volume_name(
        f"secret-boto-key--{service_name}", length_limit=63
    )
```
alongside all the other `get_*_volume_name` functions so that we only have one place to modify the volume name format and length

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -321,7 +321,7 @@ def sync_boto_secrets(
         kubernetes_signature = get_kubernetes_secret_signature(
             kube_client=kube_client,
             secret=secret,
-            service=app_name,
+            service=service,
```
nemacysts, line 324: whoops, dunno how i missed this in the original review

nemacysts (approved):

PR #3308, opened 2022-03-16
COMPINFRA-989: fix healthcheck_grace_period_seconds not being set for hacheck

BeryJu:
By default it'll continue to use 10s as delay for hacheck, but we can switch it (per cluster) with puppet like

```
  paasta_tools::config_file {'hacheck':
    config => {
      'hacheck_match_initial_delay' => true,
    }
  }
```

(probably in conjunction with a bunch of manual prep work)

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -896,13 +896,18 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             self.get_enable_nerve_readiness_check(system_paasta_config)
             or self.get_enable_envoy_readiness_check(system_paasta_config)
         ):
+            initial_delay = self.get_healthcheck_grace_period_seconds()
+            # COMPINFRA-989, this used to be hardcoded to always be 10 seconds
+            # and to not cause rolling updates on everything at once this is a config option for now
+            if not system_paasta_config.get_hacheck_match_initial_delay():
+                initial_delay = 10
             readiness_probe = V1Probe(
                 _exec=V1ExecAction(
                     command=self.get_readiness_check_script(system_paasta_config)
                     + [str(self.get_container_port())]
                     + self.get_registrations()
                 ),
-                initial_delay_seconds=10,
+                initial_delay_seconds=initial_delay,
                 period_seconds=10,
             )
         else:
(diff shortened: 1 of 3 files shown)


BeryJu:
> would it be worth just dealing with the big bounce? we can only toggle this setting per-cluster as-is, so we'd have a similar big bounce if we let released a new paasta version that always grabbed the correct delay
> 
> if we really want to avoid a big bounce, I think we'd probably want to add a toggle per-service or per-service-instance to slowly roll this out and then clean that up and make this a default once we're done rolling out

Yeah I considered adding it as service-level toggle but then it'd probably take even longer to rollout across everything

I'm currently trying to figure out if I can implement this in a way where it only use the new setting for new deployments and leaves it for existing deployments

nemacysts:
> I'm currently trying to figure out if I can implement this in a way where it only use the new setting for new deployments and leaves it for existing deployments

@BeryJu does https://sourcegraph.com/github.com/Yelp/paasta@aaf5b6c3925a068e609aea8b807f7681149cc9cf/-/blob/tests/test_kubernetes_tools.py?L3431:22 fail if we remove the toggle and unconditionally put the right value in? if not, then k8s should actually do this for us by default

BeryJu:
lemme fix the failing tests and then test with the global toggle switched

BeryJu:
actually I already set the global toggle to match the delay so it seems like `test_warning_big_bounce` is already catching it

BeryJu:
I tested this PR on kubestage with the defaults of `hacheck_match_initial_delay = True`, and it caused all services to be bounced, so we will probably have to do some manual migrating before deploying this

TheNavigat:
> I tested this PR on kubestage with the defaults of `hacheck_match_initial_delay = True`, and it caused all services to be bounced, so we will probably have to do some manual migrating before deploying this

Let's reverse the rollout procedure, then. Set the default value here to `False` to be extra safe, and using puppet we can configure it to `True` bit by bit.

TheNavigat (commented):
Redundant comment so that you can re-request review

TheNavigat (approved):
Please give testing on Kubestage another go, just to make sure that we have truly fixed causing everything to bounce.

BeryJu:
tested again on kubestage with a full itest deb and compared pods before & after, no pods were changed (besides the contract monitor which is always re-deployed)

PR #3309, opened 2022-03-16
Rewrite the Paasta Contract docs to make more sense to a 2022 audienc…

EvanKrall:
…e, not a 2015 audience.

This shouldn't change anything of substance about the actual contract (i.e. services that met the contract before should continue to meet the contract now.)

I also adjusted the format of the rST source to do one-sentence-per-line, which I prefer for git reasons.


docs/source/about/contract.rst
```diff
@@ -63,64 +91,86 @@ MUST EITHER:
...
-    E.g.:
-    DOCKER_TAG ?= <servicename>-dev-$(USER)
-    ...
-    build:
-    docker build -t $(DOCKER_TAG) .
+* The Makefile SHOULD respond to ``make cook-image`` by generating a Docker image tagged with ``$(DOCKER_TAG)`` for use by local-run.
 
-* SHOULD respond to make cook-image by generating a Docker image tagged with $(DOCKER_TAG) for use by local-run
-* cook-image MAY be added as a dependency of the itest target so that both Jenkins and local-run use the same image creation process::
+   * cook-image MAY be added as a dependency of the itest target so that both Jenkins and local-run use the same image creation process::
```
nemacysts, line 140: I wonder if we should change the `SHOULD` and `MAY` here with `MUST`

supporting cook-image and local-run building an image in different ways seems like something we'd want to avoid (and requiring cook-image would let CID do some nice optimizations in a "guaranteed" way

docs/source/about/contract.rst
```diff
@@ -63,64 +91,86 @@ MUST EITHER:
...
 
 Manual Steps (Not Recommended)
 ------------------------------
 
-This section describes how to operate the PaaSTA deploy system. If for some
-reason you don’t want to use Jenkins, you can pull the levers yourself. The
-PaaSTA CLI contains subcommands to help with this (these same subcommands are
-used by Jenkins)
+This section describes how to operate the PaaSTA deploy system.
+If for some reason you don’t want to use Jenkins, you can pull the levers yourself.
```
nemacysts, line 157: i feel like we should have stronger wording here about how unrecommended this is - but that's just me :p

nemacysts (approved):

docs/source/about/contract.rst
EvanKrall: Yeah, we can. That's just a more structural change than I wanted to do in this PR. I'd 🚢  a separate PR to `s/SHOULD|MAY/MUST/` on those two lines.

PR #3310, opened 2022-03-18
Delete setup_marathon_job, cleanup_marathon_jobs, deployd

EvanKrall:
... associated unit tests and itests, minor things that try to import deployd/SMJ.


jfongatyelp (approved):

Rob-Johnson (approved):

paasta_itests/steps/bounces_steps.py
```diff
@@ -11,327 +11,9 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import socket
```
nemacysts, line 14: should we just delete this file? i'm not sure that the code that the remaining test in this file touches is actually used by anything

nemacysts (approved):

paasta_itests/steps/bounces_steps.py
EvanKrall: paasta_native.feature uses the `a task has drained` step, and paasta_native is used by adhoc jobs. The draining functionality probably isn't used by adhoc jobs though.

PR #3312, opened 2022-03-18
Update scribereader to 0.15

geokous:
Update scribereader to 0.15 to get paasta logs working in uswest2-corpprod

### DIFF
--- a/yelp_package/extra_requirements_yelp.txt
+++ b/yelp_package/extra_requirements_yelp.txt
@@ -24,7 +24,7 @@ pyopenssl==19.0.0
 PySubnetTree==0.34
 python-jsonschema-objects==0.3.1
 retrying==1.3.3
-scribereader==0.10.0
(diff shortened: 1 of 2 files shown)


yelp_package/extra_requirements_yelp.txt
```diff
@@ -24,7 +24,7 @@ pyopenssl==19.0.0
 PySubnetTree==0.34
 python-jsonschema-objects==0.3.1
 retrying==1.3.3
-scribereader==0.10.0
+scribereader==0.15.0
```
nemacysts, line 27: we'll also want to bump python-dateutil>=2.8.1 in requirements.txt (seems like that lower bound was added to the scribereader setup.py about a year ago)

nemacysts (changes requested):

nemacysts (approved):

PR #3315, opened 2022-03-21
Add skeleton for generating spark driver env/cmd

nemacysts:
This just sets up some scaffolding for what we need to actually start a
driver - a follow-up PR will actually do all the spark_args dict parsing
and use the code from this PR to turn that into a command line


nemacysts:
depends on https://github.com/Yelp/paasta/pull/3314

jfongatyelp (approved):

PR #3319, opened 2022-03-25
Handle potentially empty but present started_at property of pod state

jfongatyelp:
This should fix the occasional 500s we get when a pod has recently started and has a container state with an empty `started_at` property.  
```
2022-03-25 12:26:53 | ERROR:paasta_tools.api.views.exception:Traceback (most recent call last):
2022-03-25 12:26:53 | AttributeError: 'NoneType' object has no attribute 'timestamp'
2022-03-25 12:26:53 | start_timestamp = this_state["started_at"].timestamp()
2022-03-25 12:26:53 | File "/opt/venvs/paasta-tools/lib/python3.7/site-packages/paasta_tools/instance/kubernetes.py", line 830, in get_pod_containers
2022-03-25 12:26:53 | "containers": containers_task.result(),
2022-03-25 12:26:53 | File "/opt/venvs/paasta-tools/lib/python3.7/site-packages/paasta_tools/instance/kubernetes.py", line 776, in get_pod_status
2022-03-25 12:26:53 | "pods": await asyncio.gather(*pod_status_tasks) if pod_status_tasks else [],
2022-03-25 12:26:53

<br class="Apple-interchange-newline">
```

We had correctly handled this when looking at `last_state` for the container, but not the current one. 

This also adds another test for `get_pod_containers` to check this exact issue sets `timestamp` correctly w/o throwing errors (None)

### DIFF
--- a/paasta_tools/instance/kubernetes.py
+++ b/paasta_tools/instance/kubernetes.py
@@ -826,7 +826,7 @@ async def get_pod_containers(
                     reason = this_state["reason"]
                 if "message" in this_state:
                     message = this_state["message"]
-                if "started_at" in this_state:
+                if this_state.get("started_at"):
                     start_timestamp = this_state["started_at"].timestamp()
 
         last_state_dict = cs.last_state.to_dict()
--- a/tests/instance/test_kubernetes.py
+++ b/tests/instance/test_kubernetes.py
@@ -817,7 +817,7 @@ async def test_get_pod_containers(mock_pod):
 
     with asynctest.patch(
         "paasta_tools.instance.kubernetes.get_tail_lines_for_kubernetes_container",
-        side_effect=[["current"], ["previous"]],
+        side_effect=[["current"], ["previous"], ["current"], ["previous"]],
         autospec=None,
     ), mock.patch(
         "paasta_tools.kubernetes_tools.recent_container_restart",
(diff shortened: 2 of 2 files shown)


paasta_tools/instance/kubernetes.py
```diff
@@ -826,7 +826,7 @@ async def get_pod_containers(
                     reason = this_state["reason"]
                 if "message" in this_state:
                     message = this_state["message"]
-                if "started_at" in this_state:
+                if this_state.get("started_at"):
                     start_timestamp = this_state["started_at"].timestamp()
```
kawaiwanyelp, line 830: If `started_at` is `None`, then there would be an exception when calling `.timestamp()` I think. You'll also need to check for the non-null case, but I'm not sure what you would retrieve here. Just set `this_state["started_at"]` to `None` maybe?

kawaiwanyelp (changes requested):

paasta_tools/instance/kubernetes.py
jfongatyelp: if started_at is None or empty or otherwise False-y, we won't get to the point where we call `this_state["started_at"].timestamp()`. That's exactly the situation we are trying to fix here.  

I figured we don't really need to distinguish from the state not having this info at all to having unuseful/blank info.

kawaiwanyelp (approved):

lanceleih (approved):
LGTM

PR #3324, opened 2022-04-01
use custom user-agent in k8s requests

BeryJu:
to better understand where k8s API requests are coming from, add a unique user-agent that we can filter in the logs

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -112,6 +112,7 @@ from kubernetes.client.rest import ApiException
 from mypy_extensions import TypedDict
 from service_configuration_lib import read_soa_metadata
 
+from paasta_tools import __version__
 from paasta_tools.async_utils import async_timeout
 from paasta_tools.long_running_service_tools import AutoscalingParamsDict
 from paasta_tools.long_running_service_tools import host_passes_blacklist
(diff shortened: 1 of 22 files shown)


nemacysts (approved):
++

i'm gonna make a ticket so that i can do something similar with tron+task_proc

BeryJu:
> I kinda think you should go even further, and make every caller of the `KubeApiClient` pass in a user agent.
> 
> that way we can see which specific script is making the calls, more than just `paasta_tools`.

Good point, added a basic version of that, that just uses `__file__` as identifier, should give us enough info for now but open for other ideas

paasta_tools/kubernetes_tools.py
```diff
@@ -1421,15 +1423,15 @@ def get_autoscaled_instances(self) -> Optional[int]:
         try:
             if self.get_persistent_volumes():
                 return (
-                    KubeClient()
+                    KubeClient(__file__)
```
Rob-Johnson, line 1426: this won't be that useful here, right? is it worth doing something fancy to get the name of the calling module?

BeryJu: yeah probably not, I thought about doing some stack inspection stuff like I did here https://github.com/goauthentik/authentik/blob/master/authentik/events/models.py#L201-L203, but in a lot of cases that would just return `__main__` since we call the scripts directly

Rob-Johnson: we could copy that and inspect __file__ though right?

BeryJu: but I think this is also only a band-aid until we have proper traces for paasta so probably dont have to worry about this too much

BeryJu: true, yeah that would be cleaner

paasta_tools/kubernetes_tools.py
```diff
@@ -445,16 +447,27 @@ def __init__(self) -> None:
...
-        self.rbac = kube_client.RbacAuthorizationV1Api()
+        if not component:
+            # If we don't get an explicit component set via constructor,
+            # try to find it by looking back in the stack, and getting `__file__` from
+            # the context calling this constructor
+            # Normally, `__module__` would make more sense, but since we have a lot of
+            # single scripts we directly call, that would be set to `__main__` most of the time.
+            current = currentframe()
+            parent = current.f_back
+            component = parent.f_globals["__file__"]
```
Rob-Johnson, line 458: is there any case where `__file__` can be missing from`globals`? probably more useful to do `.get('__file__', 'unknown')` instead?

Rob-Johnson (approved):

PR #3325, opened 2022-04-05
Catch timeout issues in bounce_status and present less-drastic error message

jfongatyelp:
We are occasionally seeing slowness grabbing responses from the k8s api, but eventually things do return.

This forces paasta_tools.api.views.instance to slightly modify the response body if we hit this particular issue such that mark-for-deployment/wait-for-deployment can identify this specific response & print a more friendly error to users.

I also considered throwing a 599 rather than a 500 for this issue to make things even more easily identifiable, however 599 is not exactly an HTTP standard, and this does still represent an error on the backend where we cannot return a valid repsonse to the user.  Let me know if it makes sense to go w/ an alternate return code here.

Previously during deploys where we wait-for-deployment, we would see only
`WARNING:paasta_tools.cli.cmds.mark_for_deployment:Error getting service status from PaaSTA API for kubestage: 500 Internal Server Error`.  This would lead to many people thinking something was wrong with their services/instances/the deploy, though it was a transient issue on the API side.

This will now print out 
`WARNING:paasta_tools.cli.cmds.mark_for_deployment:Timed out fetching service status from PaaSTA API for kubestage. Will retry on next poll interval.`

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -1503,9 +1503,9 @@ def check_if_instance_is_done(
                     instance, service, cluster
                 )
             )
-        elif e.body and "Temporary" in e.body:
+        elif e.status == 599:  # Temporary issue
             log.warning(
-                f"Timed out fetching service status from PaaSTA API for {cluster}. Will retry on next poll interval."
+                f"Temporary issue fetching service status from PaaSTA API for {cluster}. Will retry on next poll interval."
             )
         else:  # 500 - error talking to api
             log.warning(
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -1503,6 +1503,10 @@ def check_if_instance_is_done(
                     instance, service, cluster
                 )
             )
+        elif e.body and "Temporary" in e.body:
+            log.warning(
+                f"Timed out fetching service status from PaaSTA API for {cluster}. Will retry on next poll interval."
+            )
         else:  # 500 - error talking to api
             log.warning(
                 "Error getting service status from PaaSTA API for "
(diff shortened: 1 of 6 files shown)


tox.ini
```diff
@@ -28,6 +28,7 @@ commands =
 
 [testenv:dev-api]
 envdir = .tox/py37-linux/
+passenv = PAASTA_TEST_CLUSTER KUBECONFIG PAASTA_SYSTEM_CONFIG_DIR
```
jfongatyelp, line 31: Note I added this in as without it, I needed to actually export these in my environment, whereas the wiki instructs you to just declare them when you run `make dev-api` e.g. `PAASTA_SYSTEM_CONFIG_DIR=etc_paasta_for_deployment make dev-api`.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1499,6 +1499,10 @@ def check_if_instance_is_done(
                     instance, service, cluster
                 )
             )
+        elif e.body and "Temporary" in e.body:
```
nemacysts, line 1502: thought on subclassing ApiFailure so that we don't need to do a potentially brittle check for what's in the exception message and can instead check what the exception actually was?

nemacysts (approved):
i have a minor suggestion, but feel free to ignore if you disagree

paasta_tools/cli/cmds/mark_for_deployment.py
jfongatyelp: I couldn't quickly untangle the autogenerated oapi client-side handling of exceptions here, as they all come from a single auto-generated [ApiException](https://github.com/Yelp/paasta/blob/master/paasta_tools/paastaapi/exceptions.py#L101-L126) class.

I've gone with a slightly-less-brittle option of checking a specific non-standard error status code, rather than trying to check text.  Thoughts?

nemacysts (approved):

paasta_tools/cli/cmds/mark_for_deployment.py
nemacysts: rip

but i do prefer this over coupling the exception handling with the exception message

PR #3332, opened 2022-04-07
set progress_deadline_seconds if min_ready_seconds is more than 600

BeryJu:
While testing some stuff for COMPINFRA-989 we set the min_ready_seconds to 600 to test. Turns out that breaks paasta because we don't set progress_deadline_seconds (which needs to be more than min_ready_seconds), and progress_deadline_seconds defaults to 600s in kubernetes


paasta_tools/kubernetes_tools.py
```diff
@@ -1617,13 +1617,20 @@ def format_kubernetes_app(self) -> Union[V1Deployment, V1StatefulSet]:
                     ),
                 )
             else:
+                # Default to none for progress_deadline_seconds to not change
+                # existing deployments
+                progress_deadline_seconds = None
+                if self.get_min_task_uptime() >= 600:
```
nemacysts, line 1623: might be nice to have a named constant for 600 - then we can skip the comment explaining the magic number :)

mattmb:

PR #3333, opened 2022-04-07
Forward Spark ports to Tron/task_proc

nemacysts:
Right now, drivers will start and spin up executors - but these
executors will instantly crash since they can't communicate with the
driver.

This change sets things up so that drivers can be started in a way that
makes them reachable by the executors they start

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -876,9 +876,11 @@ def format_tron_action_dict(action_config: TronActionConfig, use_k8s: bool = Fal
             # spark, unlike normal batches, needs to expose  several ports for things like the spark
             # ui and for executor->driver communication
             result["ports"] = list(
-                _get_spark_ports(
-                    system_paasta_config=load_system_paasta_config()
-                ).values()
+                set(
+                    _get_spark_ports(
+                        system_paasta_config=load_system_paasta_config()
+                    ).values()
+                )
             )
 
     elif executor in MESOS_EXECUTOR_NAMES:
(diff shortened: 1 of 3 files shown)


paasta_tools/tron_tools.py
```diff
@@ -291,6 +304,10 @@ def _build_spark_config(self) -> Dict[str, str]:
             "spark.kubernetes.executor.label.paasta.yelp.com/cluster": self.get_cluster(),
             "spark.kubernetes.node.selector.yelp.com/pool": self.get_pool(),
             "spark.kubernetes.executor.label.yelp.com/pool": self.get_pool(),
+            # this relies on the PaaSTA workload contract being followed - $PAASTA_POD_IP
+            # will be, drumroll please, the routable IP for the Pod - this allows us to
+            # not need to worry about DNS
+            "spark.driver.host": "$PAASTA_POD_IP",
```
jfongatyelp, line 310: Cool, I was a little confused here since I didn't know _build_spark_config would just be used to generate the actual command being run in the container, and therefore envvar expansion would work.

Is that always going to be the case?

jfongatyelp (approved):

paasta_tools/tron_tools.py
nemacysts: it should always be the case for now - so far, i think the only thing in spark that can be configured with yaml is some 'advanced' k8s knobs like pod affinity (but if the driver is ever configured like that, then that'll be even simpler than this somewhat magic approach)

jfongatyelp (approved):

PR #3335, opened 2022-04-11
Set KubeDeployment replicas to None for autoscaled deployments

nemacysts:
We need a way to compare two KubeDeployment's that is not sensitive to
an autoscaler changing the amount of replicas present - otherwise, when
we go to compare the current and target deployment, we could oscillate
between what the autoscaler (HPA, in this case) has set a Deployments
replicas to and what setup_kubernetes_job saw prior to an HPA
scale-{up,down}

To accommplish this, we use the fact that we attach an annotation with
the autoscaling policy name to every autoscaled deployment and set the
replicas field of our internal KubeDeployment object to None for
anything that is autoscaled - which should be the same as leaving that
value unspecified (which will then let the HPA be the only thing
managing the replicas for a Deployment)

### DIFF
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -47,7 +47,13 @@ class Application(ABC):
             attr: item.metadata.labels[paasta_prefixed(attr)]
             for attr in ["service", "instance", "git_sha", "config_sha"]
         }
-        self.kube_deployment = KubeDeployment(replicas=item.spec.replicas, **attrs)
+
+        replicas = (
+            item.spec.replicas
+            if item.spec.template.metadata.annotations.get("autoscaling") is None
+            else None
+        )
+        self.kube_deployment = KubeDeployment(replicas=replicas, **attrs)
         self.item = item
         self.soa_config = None  # type: KubernetesDeploymentConfig
         self.logging = logging
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -215,7 +215,7 @@ class KubeDeployment(NamedTuple):
     instance: str
     git_sha: str
     config_sha: str
-    replicas: int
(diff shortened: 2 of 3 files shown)


nemacysts:
i tested this on kubestage and nothing seems to have broken, but I'm still trying to write a test for s_k_j that will allow us to prove that this works

jfongatyelp (approved):
>  we attach an annotation with
the autoscaling policy name to every autoscaled deployment

I unfortunately don't think think this is currently true for all HPAs, only uwsgi-scaled ones: https://github.com/Yelp/paasta/blob/134cf0986e09d097dff4fb554fad2dc3cbd30ad5/paasta_tools/kubernetes_tools.py#L1708-L1714 

But this will still prevent the churn for uwsgi scaled instances til we update the annotation for all autoscaled instances!

nemacysts:
> > we attach an annotation with
> > the autoscaling policy name to every autoscaled deployment
> 
> I unfortunately don't think think this is currently true for all HPAs, only uwsgi-scaled ones:
> 
> https://github.com/Yelp/paasta/blob/134cf0986e09d097dff4fb554fad2dc3cbd30ad5/paasta_tools/kubernetes_tools.py#L1708-L1714
> 
> But this will still prevent the churn for uwsgi scaled instances til we update the annotation for all autoscaled instances!

oh, that’s fun ;_;

mattmb (approved):
lgtm, ++ to shipping this first and then you could add an annotation/label in a followup PR to make it catch the other stuff. At least it will take some of the load off and validate that we've understood the bug :-)

jfongatyelp (approved):

nemacysts:
> lgtm, ++ to shipping this first and then you could add an annotation/label in a followup PR to make it catch the other stuff. At least it will take some of the load off and validate that we've understood the bug :-)

yea, i'll force-push back to the annotation version of this and do the label version (+ add some missing workload contract labels) in another pr

PR #3336, opened 2022-04-12
Add verbose flag to paasta validate that displays next 5 cron runs sc…

KaspariK:
Right now there isn't any real insight into the cron schedules when running paasta validate. This change introduces a verbose flag that will dump the next 5 runs for any valid cron schedule.

[Fluffy of the verbose output](https://fluffy.yelpcorp.com/i/2Kb3NkXTT9bfMrPlhQK80zjkp3HrmM56.html)

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -634,7 +634,9 @@ def validate_secrets(service_path):
     return return_value
 
 
-def paasta_validate_soa_configs(service, service_path, verbose=False):
+def paasta_validate_soa_configs(
+    service: str, service_path: str, verbose: bool = False
+) -> bool:
     """Analyze the service in service_path to determine if the conf files are valid
 
     :param service_path: Path to directory containing soa conf yaml files for service
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -21,6 +21,7 @@ from datetime import datetime
 from glob import glob
 from typing import Any
 from typing import Dict
+from typing import List
 from typing import Optional
 
 import yaml
@@ -46,7 +47,7 @@ from paasta_tools.secret_tools import get_secret_name_from_ref
 from paasta_tools.secret_tools import is_secret_ref
 from paasta_tools.secret_tools import is_shared_secret
 from paasta_tools.tron_tools import list_tron_clusters
-from paasta_tools.tron_tools import load_tron_service_config_no_cache
+from paasta_tools.tron_tools import load_tron_service_config
 from paasta_tools.tron_tools import validate_complete_config
 from paasta_tools.utils import get_service_instance_list
 from paasta_tools.utils import list_all_instances_for_service
@@ -267,7 +268,7 @@ def add_subparser(subparsers):
         "--verbose",
         action="store_true",
         required=False,
-        help="Display verbose output. This shows the next few cron runs scheduled.",
+        help="Toggle to display additional validation messages for humans.",
     )
     validate_parser.add_argument(
         "-y",
@@ -328,7 +329,7 @@ def path_to_soa_dir_service(service_path):
     return soa_dir, service
 
 
-def validate_tron(service_path, verbose=False):
+def validate_tron(service_path: str, verbose: bool = False) -> bool:
     soa_dir, service = path_to_soa_dir_service(service_path)
     returncode = True
 
@@ -336,15 +337,18 @@ def validate_tron(service_path, verbose=False):
         if not validate_tron_namespace(service, cluster, soa_dir):
             returncode = False
         elif verbose:
-            service_config = load_tron_service_config_no_cache(service, cluster)
+            # cron schedule has been validated and is safe to parse
+            service_config = load_tron_service_config(service, cluster)
             for config in service_config:
                 schedule = config.get_schedule()
-                num_runs = 5
 
                 if schedule.startswith("cron"):
-                    print(info_message(f"Next 5 cron runs for {config.get_name()}"))
-                    next_cron_runs = get_next_x_cron_runs(
-                        num_runs, schedule.replace("cron", ""), datetime.today()
+                    print(info_message(f"Upcoming runs for {config.get_name()}"))
+                    next_cron_runs = list_upcoming_runs(
+                        # most cron parsers won't understand our schedule tag, so we need to strip
+                        # that off before passing it to anything else
+                        cron_schedule=schedule.replace("cron", ""),
+                        starting_from=datetime.today(),
                     )
 
                     for run in next_cron_runs:
@@ -592,13 +596,11 @@ def check_secrets_for_instance(instance_config_dict, soa_dir, service_path, vaul
     return return_value
 
 
-def get_next_x_cron_runs(num_runs, schedule, start_datetime):
-    iter = croniter(schedule, start_datetime)
-    next_runs = []
-    for _ in range(num_runs):
-        next_runs.append(iter.get_next(datetime))
-
-    return next_runs
+def list_upcoming_runs(
+    cron_schedule: str, starting_from: datetime, num_runs: int = 5
+) -> List[str]:
+    iter = croniter(cron_schedule, starting_from)
+    return [iter.get_next(datetime) for _ in range(num_runs)]
 
 
 def validate_secrets(service_path):
@@ -676,7 +678,6 @@ def paasta_validate(args):
     """
     service_path = get_service_path(args.service, args.yelpsoa_config_root)
     service = args.service or guess_service_name()
-    verbose = args.verbose
 
-    if not paasta_validate_soa_configs(service, service_path, verbose):
+    if not paasta_validate_soa_configs(service, service_path, args.verbose):
         return 1
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -49,6 +49,7 @@ from paasta_tools.secret_tools import is_secret_ref
 from paasta_tools.secret_tools import is_shared_secret
 from paasta_tools.tron_tools import list_tron_clusters
 from paasta_tools.tron_tools import load_tron_service_config
+from paasta_tools.tron_tools import TronJobConfig
 from paasta_tools.tron_tools import validate_complete_config
 from paasta_tools.utils import get_service_instance_list
 from paasta_tools.utils import list_all_instances_for_service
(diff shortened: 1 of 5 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -567,6 +592,15 @@ def check_secrets_for_instance(instance_config_dict, soa_dir, service_path, vaul
     return return_value
 
 
+def get_next_x_cron_runs(num_runs, schedule, start_datetime):
```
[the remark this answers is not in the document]
nemacysts: additionally, what are your thoughts on renaming this function and updating the params a bit?
```suggestion
def list_upcoming_runs(cron_schedule: str, starting_from: datetime.datetime, num_runs: int = 5):
```

this name would be a bit less of a mouthful and clean up callsites (something something 5 runs should be enough for everybody something something :p) and calling it above is then just
```python
next_cron_runs = list_upcoming_runs(
    # most cron parsers won't understand our schedule tag, so we need to strip
    # that off before passing it to anything else
    cron_schedule=schedule.replace("cron", ""),
    starting_from=datetime.today()
)
```

paasta_tools/cli/utils.py
```diff
@@ -106,6 +106,13 @@ def x_mark():
     return PaastaColors.red("\u2717")
 
 
+def info_mark():
```
nemacysts, line 109:
```suggestion
def info_mark() -> str:
```

paasta_tools/cli/utils.py
```diff
@@ -124,6 +131,15 @@ def failure(msg, link):
     return "{} {} {}".format(x_mark(), msg, PaastaColors.blue(link))
 
 
+def info_message(msg):
```
nemacysts, line 134:
```suggestion
def info_message(msg: str) -> str:
```

paasta_tools/cli/utils.py
```diff
@@ -124,6 +131,15 @@ def failure(msg, link):
...
+    """Format a paasta info message.
+
+    :param msg: a string
+    :return: a beautiful string
+    """
+    return "{} {}".format(info_mark(), msg)
```
nemacysts, line 140:
```suggestion
    return f"{info_mark() {msg}"
```

paasta_tools/cli/cmds/validate.py
```diff
@@ -598,7 +632,7 @@ def validate_secrets(service_path):
     return return_value
 
 
-def paasta_validate_soa_configs(service, service_path):
+def paasta_validate_soa_configs(service, service_path, verbose=False):
```
nemacysts, line 635: while we're here, we can be good scouts and leave this a better place than we found it and add some types:
```suggestion
def paasta_validate_soa_configs(service: str, service_path: str, verbose: bool = False) -> bool:
```

paasta_tools/cli/cmds/validate.py
```diff
@@ -642,5 +676,7 @@ def paasta_validate(args):
     """
     service_path = get_service_path(args.service, args.yelpsoa_config_root)
     service = args.service or guess_service_name()
-    if not paasta_validate_soa_configs(service, service_path):
+    verbose = args.verbose
+
+    if not paasta_validate_soa_configs(service, service_path, verbose):
```
nemacysts, line 681: nit: i'd personally inline this - but if you think it's better to be consistent feel free to drop this comment

paasta_tools/cli/cmds/validate.py
```diff
@@ -317,13 +328,27 @@ def path_to_soa_dir_service(service_path):
     return soa_dir, service
 
 
-def validate_tron(service_path):
+def validate_tron(service_path, verbose=False):
```
nemacysts, line 331: while we're here, we can be good scouts and leave this a better place than we found it and add some types:
```suggestion
def validate_tron(service_path: str, verbose: bool = False) -> bool:
```

paasta_tools/cli/cmds/validate.py
```diff
@@ -317,13 +328,27 @@ def path_to_soa_dir_service(service_path):
...
     soa_dir, service = path_to_soa_dir_service(service_path)
     returncode = True
 
     for cluster in list_tron_clusters(service, soa_dir):
         if not validate_tron_namespace(service, cluster, soa_dir):
             returncode = False
+        elif verbose:
```
nemacysts, line 338: thoughts on adding a comment here to explain why this is in an else block? to someone quickly skimming this, this might look a little weird unless they realise that we're doing this to ensure that we're just parsing valid cron strings

paasta_tools/cli/cmds/validate.py
```diff
@@ -317,13 +328,27 @@ def path_to_soa_dir_service(service_path):
...
+            service_config = load_tron_service_config_no_cache(service, cluster)
```
nemacysts, line 339: we probably want to use the cached version of this (which iirc, is the same function name minus `_no_cache`

paasta_tools/cli/cmds/validate.py
```diff
@@ -258,6 +262,13 @@ def add_subparser(subparsers):
         required=False,
         help="Service that you want to validate. Like 'example_service'.",
     ).completer = lazy_choices_completer(list_services)
+    validate_parser.add_argument(
+        "-v",
+        "--verbose",
+        action="store_true",
+        required=False,
+        help="Display verbose output. This shows the next few cron runs scheduled.",
```
nemacysts, line 270: thoughts on something like
```suggestion
        help="Toggle to display additional validation messages for humans.",
```

if we ever add any other verbose output, we'd have to further update this help text if we leave this as-is

nemacysts (commented):
looking good!
 
i mostly just have minor suggestions/potential improvements :)

paasta_tools/cli/cmds/validate.py
```diff
@@ -317,13 +328,27 @@ def path_to_soa_dir_service(service_path):
...
+        elif verbose:
+            service_config = load_tron_service_config_no_cache(service, cluster)
+            for config in service_config:
+                schedule = config.get_schedule()
+                num_runs = 5
+
+                if schedule.startswith("cron"):
+                    print(info_message(f"Next 5 cron runs for {config.get_name()}"))
+                    next_cron_runs = get_next_x_cron_runs(
+                        num_runs, schedule.replace("cron", ""), datetime.today()
```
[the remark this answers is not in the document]
KaspariK: Added. You can see an example of a service with different timezones in the fluffy [here](https://fluffy.yelpcorp.com/i/Ztw9kq458vJc4zg9tHWhD29HMS8PlGJW.html).

paasta_tools/cli/cmds/validate.py
```diff
@@ -317,13 +330,33 @@ def path_to_soa_dir_service(service_path):
...
+                    print(info_message(f"Upcoming runs for {config.get_name()}:"))
+                    next_cron_runs = list_upcoming_runs(
+                        # most cron parsers won't understand our schedule tag, so we need to strip
+                        # that off before passing it to anything else
+                        cron_schedule=schedule.replace("cron", ""),
+                        starting_from=tz.localize(datetime.today()),
+                    )
+
+                    for run in next_cron_runs:
+                        print(f"{run}")
```
nemacysts, line 359: after looking at the fluffy of the output, what are your thoughts on adding a little whitespace here
```suggestion
                        print(f"'\t{run}")
```

paasta_tools/cli/utils.py
```diff
@@ -124,6 +131,15 @@ def failure(msg, link):
     return "{} {} {}".format(x_mark(), msg, PaastaColors.blue(link))
 
 
+def info_message(msg: str) -> str:
+    """Format a paasta info message.
+
+    :param msg: a string
+    :return: a beautiful string
```
nemacysts, line 138: probably fine to not add this since they don't buy us a ton (they're basically just repeating the type signature)

paasta_tools/cli/cmds/validate.py
```diff
@@ -317,13 +330,33 @@ def path_to_soa_dir_service(service_path):
...
     for cluster in list_tron_clusters(service, soa_dir):
         if not validate_tron_namespace(service, cluster, soa_dir):
             returncode = False
+        elif verbose:
+            # service config has been validated and cron schedules are safe to parse
+            service_config = load_tron_service_config(service, cluster)
+            for config in service_config:
+                schedule = config.get_schedule()
+
+                if isinstance(schedule, str) and schedule.startswith("cron"):
```
nemacysts, line 346: i think the last thing we'll want to handle is the dict form of a cron schedule (which i didn't know was a thing until today) - maybe we add a helper for figuring out if a schedule is cron-based and another for actually getting the cron expression?

paasta_tools/tron_tools.py
```diff
@@ -559,6 +559,19 @@ def get_node(self):
     def get_schedule(self):
         return self.config_dict.get("schedule")
 
+    def get_cron_expression(self) -> Optional[str]:
+        schedule = self.config_dict.get("schedule")
+        if (
+            isinstance(schedule, Dict)
```
nemacysts, line 565: does this work with `isinstance(schedule, dict)`? imo, this would be more familiar to folks if that does work

paasta_tools/tron_tools.py
```diff
@@ -559,6 +559,19 @@ def get_node(self):
...
+            and "type" in schedule
+            and schedule["type"] == "cron"
+        ):
+            return schedule["value"]
+        elif isinstance(schedule, str) and schedule.startswith("cron"):
+            return schedule
```
nemacysts, line 571: suggestion: thoughts on returning
```suggestion
            return schedule.replace("cron", "")
```
here so that we don't need to do any additional manipulation of this value in `print_upcoming_runs()`?

paasta_tools/cli/cmds/validate.py
```diff
@@ -317,17 +331,40 @@ def path_to_soa_dir_service(service_path):
...
+                if cron_expression:
+                    print_upcoming_runs(config, cron_expression)
 
     return returncode
 
 
+def print_upcoming_runs(config: TronJobConfig, cron_expression: str) -> None:
+    print(info_message(f"Upcoming runs for {config.get_name()}:"))
+
+    config_tz = config.get_time_zone() or "US/Pacific"
```
nemacysts, line 355: suggestion: we could extract `"US/Pacific"` to a `DEFAULT_TZ` constant in tron_tools.py

nemacysts (approved):
🥳

PR #3337, opened 2022-04-13
Add some missing workload labels + add autoscaled label

nemacysts:
this will cause a big bounce!

the annotation i'm using for https://github.com/Yelp/paasta/pull/3335 is only available for uwsgi-autoscaled instances, so this switches us to something more generic. and, while we're incurring a big bounce, we might as well also add in some missing paasta contract labels that we're missing

todo:
- still need to add owner, but i wanna talk to cic about what they'd like
there
- need to see what other objects they'd like these labels on (since some of the labels don't make sense for certain objects)

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1709,6 +1709,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
 
         # The HPAMetrics collector needs these annotations to tell it to pull
         # metrics from these pods
+        # TODO: see if we can remove this as we're no longer using sfx data to scale
         if metrics_provider == "uwsgi":
             annotations["autoscaling"] = metrics_provider
 
@@ -1791,6 +1792,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             "paasta.yelp.com/service": self.get_service(),
             "paasta.yelp.com/instance": self.get_instance(),
             "paasta.yelp.com/git_sha": git_sha,
+            "paasta.yelp.com/autoscaled": str(self.is_autoscaling_enabled()).lower(),
         }
 
         # Allow the Prometheus Operator's Pod Service Monitor for specified
@@ -2150,7 +2152,8 @@ def list_deployments(
             git_sha=item.metadata.labels.get("paasta.yelp.com/git_sha", ""),
             config_sha=item.metadata.labels["paasta.yelp.com/config_sha"],
             replicas=item.spec.replicas
-            if item.spec.template.metadata.annotations.get("autoscaling") is None
+            if item.metadata.labels.get(paasta_prefixed("autoscaled"), "false")
+            == "false"
             else None,
         )
         for item in deployments.items + stateful_sets.items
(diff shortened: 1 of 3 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -310,6 +310,10 @@ def _set_disrupted_pods(self: Any, disrupted_pods: Mapping[str, datetime]) -> No
         "paasta.yelp.com/scrape_uwsgi_prometheus": str,
         "paasta.yelp.com/scrape_piscina_prometheus": str,
         "paasta.yelp.com/service": str,
+        "paasta.yelp.com/pool": str,
```
Rob-Johnson, line 313: can you use the `paasta_prefixed` function here too?

paasta_tools/kubernetes_tools.py
```diff
@@ -1791,8 +1803,18 @@ def get_pod_template_spec(
...
+            "paasta.yelp.com/autoscaled": str(self.is_autoscaling_enabled()).lower(),
+            "paasta.yelp.com/cluster": self.get_cluster(),
+            "paasta.yelp.com/pool": self.get_pool(),
+            # ideally, we'd also have the docker image here for ease-of-use
+            # in Prometheus relabeling, but that information is over the
+            # character limit for k8s labels (63 chars)
+            "paasta.yelp.com/deploy_group": self.get_deploy_group(),
         }
 
+        if self.get_team():
```
Rob-Johnson, line 1815: is the label being absent better than something like `Unknown`?

paasta_tools/kubernetes_tools.py
```diff
@@ -1791,8 +1803,18 @@ def get_pod_template_spec(
             "paasta.yelp.com/service": self.get_service(),
             "paasta.yelp.com/instance": self.get_instance(),
             "paasta.yelp.com/git_sha": git_sha,
+            "paasta.yelp.com/autoscaled": str(self.is_autoscaling_enabled()).lower(),
+            "paasta.yelp.com/cluster": self.get_cluster(),
+            "paasta.yelp.com/pool": self.get_pool(),
+            # ideally, we'd also have the docker image here for ease-of-use
+            # in Prometheus relabeling, but that information is over the
+            # character limit for k8s labels (63 chars)
```
Rob-Johnson, line 1811: why not just shorten it? I think docker does 12 by default?

Rob-Johnson (approved):
some minor nits - would be good to just use the `paasta_prefixed` function everywhere and get rid of the `paasta.yelp.com` string.

paasta_tools/kubernetes_tools.py
nemacysts: can do!

paasta_tools/kubernetes_tools.py
nemacysts: hmm, i can check with dpopes tomorrow to see how OPA handles a value that doesn't map to a team

paasta_tools/kubernetes_tools.py
nemacysts: went ahead and used paasta_prefixed everywhere as you mentioned in your approval comment

paasta_tools/kubernetes_tools.py
nemacysts: OPA doesn't read this at all, although dpopes mentioned he's worried about confusing people with the `team` construct in OPA and this label (which can also be used in OPA if desired)

that said, i went ahead and made this a default label with a fallback value of Unknown

paasta_tools/kubernetes_tools.py
nemacysts: hmm, i'm a little wary of adding in a truncated form of this since that's not actually usable (e.g., if you try to docker pull with anything but the full tag docker will say there's nothing to pull)

PR #3340, opened 2022-04-21
Add autoscaled label

nemacysts:
The original approach here used an annotation that was only present for uwsgi-scaled services. To fix this, we'll add another label and use that instead in order to prevent any issues from our existing metrics collectors potentially getting confused by a new value for the existing annotation

### DIFF
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -50,7 +50,8 @@ class Application(ABC):
 
         replicas = (
             item.spec.replicas
-            if item.spec.template.metadata.annotations.get("autoscaling") is None
+            if item.metadata.labels.get(paasta_prefixed("autoscaled"), "false")
+            == "false"
             else None
         )
         self.kube_deployment = KubeDeployment(replicas=replicas, **attrs)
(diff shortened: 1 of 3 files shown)


nemacysts:
(https://github.com/Yelp/paasta/pull/3335 is the original approach)

Rob-Johnson (approved):

paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -50,7 +50,8 @@ def __init__(
 
         replicas = (
             item.spec.replicas
-            if item.spec.template.metadata.annotations.get("autoscaling") is None
+            if item.metadata.labels.get(paasta_prefixed("autoscaled"), "false")
```
mattmb, line 53: How come it's at a different level to the annotations? Is that because it's the Deployments labels rather than the Pod lables?

mattmb (approved):

paasta_tools/kubernetes/application/controller_wrappers.py
nemacysts: i'm actually not sure - i was going off off the other label accesses above...lemme run this in kubestage again real fast and see if labels are also available in the template spec or if its just annotations in there for whatever reason

nemacysts: huh, looks like i missed something when refactoring and didn't actually also set this label on the deployment (and just did the pods) - i'll add a follow-up commit that correctly refactors this

nemacysts: done - good catch! we should probably eventually add an itest/e2e test for s_k_j since my existing unit test was mocking things out in the expected format - but the actual code i pushed didn't match that format when run irl :p

jfongatyelp (approved):

ankit864 (approved):

PR #3341, opened 2022-04-21
Update to check all status versions before setting state to stopped

KaspariK:
Currently if an instance has been manually stopped with paasta stop, paasta status will show State: Stopping when instance count is 0. The state should show Stopped when the process is complete.

[Before](https://fluffy.yelpcorp.com/i/MD36h6KW5gqMbVDWPgzj6sl0341xgdJC.html)
[After](https://fluffy.yelpcorp.com/i/3TztB9xmgdCQQnw1MnfC19KPB94TrJtl.html)


nemacysts (approved):
nice! nothing like a one-line fix :)

PR #3344, opened 2022-04-26
Use workload owners for workload contract labels

nemacysts:
the workload contract was previously under-specified and did not mention
what the value for the "yelp.com/owner" label should be (i.e., should it
be the owner of the code that the labeled objects support or should it
be the owners of the infra itself)

We've gone ahead and decided that this should point to the infra (i.e.,
workload) owners.

Tron was the only thing in this repo currently setting this label
(adding these labels to PaaSTA itself is happening in another PR)


nemacysts:
This requires https://github.yelpcorp.com/sysgit/puppet/pull/6438 to be merged internally for these values to be set correctly

mattmb:

PR #3345, opened 2022-04-27
Support additional version information in push-to-registry

dpopacristobal:
We want push-to-registry to support the new docker image name:
`services-{service}:paasta-{git_sha}-{extrastuff}`

https://jira.yelpcorp.com/browse/COMPINFRA-1039

### DIFF
--- a/paasta_tools/generate_deployments_for_service.py
+++ b/paasta_tools/generate_deployments_for_service.py
@@ -202,12 +202,10 @@ def get_deploy_group_mappings(
     return mappings, v2_mappings
 
 
-def build_docker_image_name(
-    service: str, sha: str, version_metadata: str = None
-) -> str:
+def build_docker_image_name(service: str, sha: str, image_version: str = None) -> str:
     image_name = f"services-{service}:paasta-{sha}"
-    if version_metadata is not None:
-        image_name += f"-{version_metadata}"
+    if image_version is not None:
+        image_name += f"-{image_version}"
 
     return image_name
 
--- a/paasta_tools/generate_deployments_for_service.py
+++ b/paasta_tools/generate_deployments_for_service.py
@@ -46,6 +46,7 @@ import re
 from typing import Any
 from typing import Dict
 from typing import List
+from typing import Optional
 from typing import Tuple
 
 from mypy_extensions import TypedDict
@@ -202,7 +203,9 @@ def get_deploy_group_mappings(
     return mappings, v2_mappings
 
 
-def build_docker_image_name(service: str, sha: str, image_version: str = None) -> str:
+def build_docker_image_name(
+    service: str, sha: str, image_version: Optional[str] = None
+) -> str:
     image_name = f"services-{service}:paasta-{sha}"
     if image_version is not None:
         image_name += f"-{image_version}"
--- a/paasta_tools/generate_deployments_for_service.py
+++ b/paasta_tools/generate_deployments_for_service.py
@@ -202,8 +202,14 @@ def get_deploy_group_mappings(
     return mappings, v2_mappings
 
 
-def build_docker_image_name(service: str, sha: str) -> str:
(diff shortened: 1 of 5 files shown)


EvanKrall (approved):
Seems reasonable to me. I'm not sure how I feel about `--version-metadata`, but I'm also not sure what would be better. `--image-version`? `--extra-info`?

How much effort would it be to add mypy type annotations to the files you're touching here? (and set `disallow_untyped_defs = True` for this file in [mypy.ini](https://github.com/Yelp/paasta/blob/master/mypy.ini))

paasta_tools/utils.py
```diff
@@ -2865,13 +2865,17 @@ def build_docker_image_name(service: str) -> str:
...
-def build_docker_tag(service: str, upstream_git_commit: str) -> str:
+def build_docker_tag(
+    service: str, upstream_git_commit: str, image_version: str = None
+) -> str:
     """Builds the DOCKER_TAG string
 
     upstream_git_commit is the SHA that we're building. Usually this is the
     tip of origin/master.
     """
     tag = "{}:paasta-{}".format(build_docker_image_name(service), upstream_git_commit)
```
jfongatyelp, line 2876: Interesting; since you updated `build_docker_image_version` to optionally accept  `image_version`, why duplicate manually adding `-{image_version}` here?

paasta_tools/utils.py
```diff
@@ -3415,6 +3419,18 @@ def format_tag(tag: str) -> str:
     return "refs/tags/%s" % tag
 
 
+def build_image_identifier(
+    git_sha: str, sha_len: Optional[int] = None, image_version: Optional[str] = None
+) -> str:
```
jfongatyelp, line 3424: Seems useful! Which places were you planning on using this w/ a limited sha?

paasta_tools/generate_deployments_for_service.py
```diff
@@ -202,8 +202,12 @@ def get_deploy_group_mappings(
     return mappings, v2_mappings
 
 
-def build_docker_image_name(service: str, sha: str) -> str:
-    return f"services-{service}:paasta-{sha}"
+def build_docker_image_name(service: str, sha: str, image_version: str = None) -> str:
```
jfongatyelp, line 205: Should that be `image_version: Optional[str]` ?

paasta_tools/utils.py
dpopacristobal: I noticed that in places like [this](self.commit[:8]), `self.commit[:8]` is used to identify particular images, so was planning on using the limited sha there

paasta_tools/utils.py
dpopacristobal: That's a different `build_docker_image_name` that I haven't changed, confusingly there's two of them:
- [The one I changed](https://sourcegraph.yelpcorp.com/Yelp/paasta/-/blob/paasta_tools/generate_deployments_for_service.py?subtree=true#L205)
- [The other one](https://sourcegraph.yelpcorp.com/Yelp/paasta/-/blob/paasta_tools/utils.py?subtree=true#L2855)

Annoyingly, they both do slightly different things 😢

jfongatyelp: That is annoying, you're right! Thanks for clarifying.

paasta_tools/utils.py
```diff
@@ -2865,13 +2865,17 @@ def build_docker_image_name(service: str) -> str:
     return name
 
 
-def build_docker_tag(service: str, upstream_git_commit: str) -> str:
+def build_docker_tag(
+    service: str, upstream_git_commit: str, image_version: str = None
```
jfongatyelp, line 2869: This also needs to be `Optional[str]`?

jfongatyelp (approved):
LGTM, outside of the one last missed type annotation.

PR #3346, opened 2022-04-29
Support additional version information in generate-deployments-for-service

dpopacristobal:
https://jira.yelpcorp.com/browse/COMPINFRA-1041

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -334,6 +334,7 @@ class BranchDictV1(TypedDict, total=False):
 class BranchDictV2(TypedDict):
     git_sha: str
     docker_image: str
+    image_version: Optional[str]
     desired_state: str
     force_bounce: Optional[str]
 
@@ -3284,6 +3285,7 @@ class _DeploymentsJsonV2ControlsDict(TypedDict, total=False):
 class _DeploymentsJsonV2DeploymentsDict(TypedDict):
     docker_image: str
     git_sha: str
+    image_version: Optional[str]
 
 
 class DeploymentsJsonV2Dict(TypedDict):
@@ -3323,6 +3325,7 @@ class DeploymentsJsonV2:
         branch_dict: BranchDictV2 = {
             "docker_image": self.get_docker_image_for_deploy_group(deploy_group),
             "git_sha": self.get_git_sha_for_deploy_group(deploy_group),
+            "image_version": self.get_image_version_for_deploy_group(deploy_group),
             "desired_state": self.get_desired_state_for_branch(full_branch),
             "force_bounce": self.get_force_bounce_for_branch(full_branch),
         }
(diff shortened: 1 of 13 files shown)


paasta_tools/utils.py
```diff
@@ -3345,6 +3348,13 @@ def get_git_sha_for_deploy_group(self, deploy_group: str) -> str:
             e = f"{self.service} not deployed to {deploy_group}. Has mark-for-deployment been run?"
             raise NoDeploymentsAvailable(e)
 
+    def get_image_version_for_deploy_group(self, deploy_group: str) -> str:
+        try:
+            return self.config_dict["deployments"][deploy_group]["image_version"]
+        except KeyError:
+            e = f"{self.service} not deployed to {deploy_group}. Has mark-for-deployment been run?"
+            raise NoDeploymentsAvailable(e)
```
jfongatyelp, line 3356: Currently none of our config_dict['deployments'] have any `image_version` key, so we will always get this `KeyError`, but not because the `deploy_group` is missing, but because the `image_version` is missing. 

I think we might want to at least split this to two lines and `get('image_version')` so the KeyError + error message apply to the right conditions.

dpopacristobal: I've added a line to check that `image_version` is there, and otherwise return `None`. I think once these changes have propagated and `image_version` gets added to the deployments.json files, then we can remove that check (if I've understood it right). Is that okay?

jfongatyelp: Yup, once this runs once everywhere we should be able to remove it. This works fine, though I still think it is a strange edge case that would give the user the wrong error message if if deployments.json (or whatever alternate source of deployment info we're using) somehow lost keys we think 'always' exist.

paasta_tools/utils.py
```diff
@@ -3335,32 +3335,46 @@ def get_deploy_groups(self) -> Collection[str]:
...
-            return self.config_dict["deployments"][deploy_group]["docker_image"]
+            deploy_group_config = self.config_dict["deployments"][deploy_group]
         except KeyError:
             e = f"{self.service} not deployed to {deploy_group}. Has mark-for-deployment been run?"
             raise NoDeploymentsAvailable(e)
+        try:
+            return deploy_group_config[key]
+        except KeyError:
+            e = f"The configuration for service {self.service} in deploy group {deploy_group} does not contain '{key}' metadata."
+            raise KeyError(e)
```
jfongatyelp, line 3347: Oh, I was thinking something much more simple like just splitting `return self.config_dict["deployments"][deploy_group]["docker_image"]` to 
```
deploy_group_config = self.config_dict["deployments"][deploy_group]
return deploy_group_config["docker_image"]
```
so that at least the NoDeploymentsAvailable exception is raised at the right place, and we effectively return the exact same thing as before otherwise.

But, this is nice to catch the additional error with more yelpy information in the error message.

jfongatyelp (approved):
One additional manual test I did was ran generate_deployments_for_service to generate deployments.json with the new (null) image_version, then ensured setup_kubernetes_job using those newly generated deployments.json files didn't cause unexpected bounces.

So seems good to go!

PR #3347, opened 2022-05-02
Add initial auto-rollback file schema (+ validation)

nemacysts:
For the new (and hopefully improved!) auto-rollback system, we'll allow
users to specify any number of queries to predetermined metric sources
that will control whether or not their service should automatically
rollback during a deployment.

As a starting point, this PR adds the schema and some validation with
that schema against the files that users will provide us with.


lanceleih (approved):
LGTM

PR #3350, opened 2022-05-04
Support additional version information in KubernetesDeploymentConfig

dpopacristobal:
The relevant ticket for this is:
https://jira.yelpcorp.com/browse/COMPINFRA-1095

This is just to get some feedback on these changes, as this is built on top of the changes in #3346, which is not merged yet.

PR #3351, opened 2022-05-04
Support additional version information in paasta status API calls

dpopacristobal:
https://jira.yelpcorp.com/browse/COMPINFRA-1093

The PR where the labels used will be added is:
https://github.com/Yelp/paasta/pull/3350

### DIFF
--- a/tests/instance/test_kubernetes.py
+++ b/tests/instance/test_kubernetes.py
@@ -424,6 +424,65 @@ class TestKubernetesStatusV2:
             "pods": [mock.ANY],
         }
 
+    def test_statefulset_with_image_version(
+        self,
+        mock_controller_revisions_for_service_instance,
+        mock_LONG_RUNNING_INSTANCE_TYPE_HANDLERS,
+        mock_load_service_namespace_config,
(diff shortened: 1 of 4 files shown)


tests/instance/test_kubernetes.py
```diff
@@ -418,6 +419,7 @@ def test_statefulset(
             "ready_replicas": 1,
             "create_timestamp": datetime.datetime(2021, 4, 1).timestamp(),
             "git_sha": "aaa000",
+            "image_version": None,
```
EvanKrall, line 422: would be nice to see a test for the case where the label is set

EvanKrall (approved):

tests/instance/test_kubernetes.py
dpopacristobal: I've added a new one that's pretty much the same as `test_statefulset`, I wasn't sure if I should just set the image_version in `test_statefulset` or if we wanted a new test, so went for the latter, happy to change it though

jfongatyelp (approved):
Cool, running this locally looks good and I can see the new image_version info.

PR #3353, opened 2022-05-04
Correct hyperlink for Sensu

kepae:
sensuapp.org seems to be deprecated (under site maintenance). Meanwhile, sensu.io seems to be the current domain.


mattmb:
Belated thank you!

PR #3354, opened 2022-05-05
Create a script that sets up paasta-internal custom resource definitions

EvanKrall:
(not for operators run by other teams). COMPINFRA-1081


paasta_tools/setup_kubernetes_internal_crd.py
```diff
@@ -0,0 +1,159 @@
...
+                "kind": "DeployGroup",
+                "shortNames": ["dg"],
+            },
+            "validation": {
+                "openAPIV3Schema": {
+                    "type": "object",
+                    "properties": {
+                        "service": {"type": "string"},
+                        "deploy_group": {"type": "string"},
+                        "git_sha": {"type": "string"},
```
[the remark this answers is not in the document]
EvanKrall: oh yeah I suppose I should probably include that 😛

paasta_tools/setup_kubernetes_internal_crd.py
```diff
@@ -0,0 +1,159 @@
...
+        kind="CustomResourceDefinition",
+        metadata={
+            "name": "startstopcontrols.paasta.yelp.com",
+            "labels": {
+                paasta_prefixed("internal"): "true",
+            },
+        },
+        spec={
+            "group": "paasta.yelp.com",
+            "versions": [{"name": "v1", "served": True, "storage": True}],
```
jfongatyelp, line 81: I'm not familiar enough w/ k8s custom resources, does version matter to anything else about how we access the api?  I notice all operator workloads specify their crd's with `v1alpha1`.   Is there any significance to that vs `v1`?

jfongatyelp (approved):
it feels like we could add a little more validation later (e.g. we only allow start|stop for desired_state right?) but lgtm, once we add image_version

paasta_tools/setup_kubernetes_internal_crd.py
EvanKrall: I think it doesn't matter - this is for versioning of the CRDs themselves. We can start wherever we want and increment these as we see fit. Maybe v1beta1 is a better starting point though.

jfongatyelp (approved):

PR #3355, opened 2022-05-05
Add a toggle to exclude terminating pods when reporting local pods to nerve

jfongatyelp:
Currently configure_nerve uses [get_kubernetes_services_running_here_for_nerve](https://github.com/Yelp/nerve-tools/blob/master/src/nerve_tools/configure_nerve.py#L365-L368) to determine whether to report a local pod.

This caused an issue recently when a pod was stuck in `Terminating` for a long time, but its IP had already been reassigned to a different pod on the same host.  `get_kubernetes_running_services_for_nerve` reported the terminating pod, which then kept it in nerve configuration.  Because the new pod used the same healthcheck uri and returned a healthy result, the registration for the old pod continued to exist.

Currently our smartstack-registered pods' hacheck containers have a preStop that immediately marks the instance as down in the service mesh.  It seems correct to also stop reporting this instance to nerve at this time.

This PR adds a new paasta system toggle to exclude these terminating pods from being reported to nerve, but maintains the default of current behavior.

Note that there is no `Terminating` phase or state to check directly.  Pods will be in phase `Running` but include a `deletionTimestamp`. See [related docs](https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/)

This PR also changes the existing tests for `get_kubernetes_services_running_here` to make it clearer what we are testing, and which test pod is 'valid'.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2025,13 +2025,17 @@ def get_all_kubernetes_services_running_here() -> List[Tuple[str, str, int]]:
     return services
 
 
-def get_kubernetes_services_running_here() -> Sequence[KubernetesServiceRegistration]:
+def get_kubernetes_services_running_here(
+    exclude_terminating: bool = False,
+) -> Sequence[KubernetesServiceRegistration]:
     services = []
     pods = get_k8s_pods()
     for pod in pods["items"]:
-        if pod["status"]["phase"] != "Running" or "smartstack_registrations" not in pod[
-            "metadata"
-        ].get("annotations", {}):
+        if (
+            pod["status"]["phase"] != "Running"
+            or "smartstack_registrations" not in pod["metadata"].get("annotations", {})
+            or (exclude_terminating and pod["metadata"].get("deletionTimestamp"))
+        ):
             continue
         try:
             port = None
@@ -2058,7 +2062,7 @@ def get_kubernetes_services_running_here() -> Sequence[KubernetesServiceRegistra
 
 
 def get_kubernetes_services_running_here_for_nerve(
-    cluster: Optional[str], soa_dir: str
+    cluster: Optional[str], soa_dir: str, exclude_terminating_pods: bool = False
 ) -> Sequence[Tuple[str, ServiceNamespaceConfig]]:
     try:
         system_paasta_config = load_system_paasta_config()
@@ -2070,10 +2074,16 @@ def get_kubernetes_services_running_here_for_nerve(
         # these custom exceptions and return [].
         if not system_paasta_config.get_register_k8s_pods():
             return []
+        exclude_terminating = (
+            not system_paasta_config.get_nerve_register_k8s_terminating()
+        )
+
     except PaastaNotConfiguredError:
         log.warning("No PaaSTA config so skipping registering k8s pods in nerve")
         return []
-    kubernetes_services = get_kubernetes_services_running_here()
+    kubernetes_services = get_kubernetes_services_running_here(
+        exclude_terminating=exclude_terminating
+    )
     nerve_list = []
     for kubernetes_service in kubernetes_services:
         try:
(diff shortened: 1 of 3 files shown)


TheNavigat (approved):
Can't immediately see why the tests are failing. Otherwise this is some amazing work.

mattmb:
Interesting, recommend you get several opinions before you ship. This code is definitely risky, also remember when you change this you have to bump the paasta-tools version in nerve-tools to roll it out.

https://github.com/Yelp/paasta/blob/725e2c8d4d10ac0fab5751862db123fb6f4a81b3/paasta_tools/kubernetes_tools.py#L954 this is the part where we tell hacheck to serve 503s as soon as the Pod is set to terminate. So the idea is that causes nerve to react right away and remove the registrations whilst the main container is sleeping for a while before being terminated. One interesting point is that we do hadown specific registrations so maybe there's an edge case where nerve is talking to the "wrong" hacheck pod? Generally I share your thoughts that this might not actually change the behaviour. However, maybe it's part of the route to us getting rid of hacheck? @Buffer0x7cd and I have discussed in the past how we might do that but I can't remember the specifics? Now we have Envoy and don't do active healthchecks then maybe this is all we need to stop needing hacheck altogether? Overall, I think it would be sane to ship this as long as it was done slowly (I haven't read the actual code btw, just commenting on the general idea)

mattmb:
Btw, if this *is* all we need to do to get rid of hacheck then that would potentially be a good cost win, good complexity win, good reduction in containers for k8s to keep track of etc. One thing we'd have to check carefully is how many healthchecks nerve and the kubelet do send to services. Maybe the caching feature of hacheck is still beneficial if that number is high. But I would think it might be low enough that hacheck isn't worth it

Rob-Johnson:
maybe we should go all the way to replicating what the k8s endpoint controller does? To populate Envoy backends, Istio watches the contents of endpoint/endpoint slices, which are in turn managed by this controller.

https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/util/endpoint/controller_utils.go#L140

Buffer0x7cd:
+1 on @Rob-Johnson recommendation, I think it will be great if we can replicate the behaviour in the k8s control plane. That would bring this to parity with the expected default behaviour of other k8s components ( endpoint controller, cni etc)

jfongatyelp:
> maybe we should go all the way to replicating what the k8s endpoint controller does? To populate Envoy backends, Istio watches the contents of endpoint/endpoint slices, which are in turn managed by this controller.
> 
> https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/util/endpoint/controller_utils.go#L140

So it looks like this is excluding:
* Pods without PodIP
   * This is already implicitly handled since we will hit a `KeyError` trying to access `pod['spec']['podIP']`; did you want this to be added explicitly e.g. in the condition where we `continue` instead?
* When we want to exclude 'terminating' pods, pods with a deletionTimetstamp
   * This is what the current version of this PR does
* When pod restartPolicy is "never", pods that are in phase `Failed` or `Succeeded`
   * We actually seem to hard-code restartPolicy of `Always`
* When pod restartPolicy is "OnFailure", pods that are in phase `Succeeded`
   * essentially the same as above except the pod should theoretically never hit `Failed` phase anyway, given that if any container _does_ fail with an error it would get restarted and then the pod would be back to `Running`

So the biggest difference from what we do today and what they are doing seems to be in the treatment of phases.  But I think that our limitation to only `Running` phase makes sense -- the impact would be the same, plus it seems like their implementation might include `Pending` and `Unknown` pods which I think probably shouldn't be included in nerve either, given they are likely to not be responding to requests correctly at that point.

@Rob-Johnson @Buffer0x7cd would you mind letting me know what changes in particular you think we should add here?

Rob-Johnson:
this all makes sense, thanks for the explanation @jfongatyelp!

> TBH, it really feels like we should not include any pods that aren't ready to serve requests in nerve configs, but please let me know if you think otherwise.
yeah, I agree with this tbh, but it's something we're going to have to watch out for in istio.

Rob-Johnson (approved):

PR #3356, opened 2022-05-06
Status: print all available status objects

siadat:
The Cassandra status object contains data for both kubernetes and
cassandracluster. This change will make paasta status print all available
status objects. So, for Cassandra clusters, it will use the custom status
writer as well as the Kubernetes status writer, and for other objects the
output will not change.

The motivation for this change is that the Kubernetes writer provides
information that are not available in the custom Cassandra status writer.
So, we would ideally want to switch between them when necessary.

Context: [PAASTA-17617](https://jira.yelpcorp.com/browse/PAASTA-17617)

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -294,39 +294,46 @@ def paasta_status_on_api_endpoint(
     if status.git_sha != "":
         output.append("    Git sha:    %s (desired)" % status.git_sha)
 
-    instance_type = find_instance_type(status)
-
-    if system_paasta_config.get_enable_custom_cassandra_status_writer():
-        if status.get("cassandracluster") is not None:
-            instance_type = "cassandracluster"
-
-    if instance_type is not None:
-        # check the actual status value and call the corresponding status writer
-        service_status_value = getattr(status, instance_type)
-        writer_callable = INSTANCE_TYPE_WRITERS.get(instance_type)
-        return writer_callable(
-            cluster, service, instance, output, service_status_value, verbose
-        )
-    else:
-        print(
-            "Not implemented: Looks like %s is not a Marathon or Kubernetes instance"
-            % instance
+    instance_types = find_instance_types(status)
+    if not instance_types:
+        output.append(
+            PaastaColors.red(
+                f"{instance} is not currently supported by `paasta status` - "
+                f"unable to find status metadata in API response."
+            )
         )
         return 0
 
+    ret_code = 0
+    for instance_type in instance_types:
+        # check the actual status value and call the corresponding status writer
+        service_status_value = getattr(status, instance_type)
+        writer_callable = INSTANCE_TYPE_WRITERS.get(instance_type)
+        ret = writer_callable(
+            cluster, service, instance, output, service_status_value, verbose
+        )
+        if ret != 0:
+            output.append(
+                f"Status writer failed for {instance_type} with return value {ret}"
+            )
+            ret_code = ret
 
-def find_instance_type(status: Any) -> str:
+    return ret_code
+
+
+def find_instance_types(status: Any) -> List[str]:
     """
-    find_instance_type finds the instance type from the status api response it
-    iterates over all instance type registered in `INSTANCE_TYPE_WRITERS`
+    find_instance_types finds the instance types from the status api response.
+    It iterates over all instance type registered in `INSTANCE_TYPE_WRITERS`.
 
     :param status: paasta api status object
-    :return: the first matching instance type or else None
+    :return: the list of matching instance types
     """
+    types: List[str] = []
     for instance_type in INSTANCE_TYPE_WRITERS.keys():
         if status.get(instance_type) is not None:
-            return instance_type
-    return None
+            types.append(instance_type)
+    return types
 
 
 def print_adhoc_status(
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -189,6 +189,11 @@ def add_instance_filter_arguments(status_parser, verb: str = "inspect") -> None:
     status_parser.add_argument(
         "-r", "--registration", help=f"Only {verb} instances with this registration."
     )
+    status_parser.add_argument(
+        "-t",
+        "--instance-type",
+        help=f"Override the inferred instance type. By default, instance type is inferred from status object.",
```
mattmb, line 195: Could you please add a couple of examples of instance-type just so it's super easy for paasta users to understand what we mean by instance-type.

paasta_tools/cli/cmds/status.py
```diff
@@ -294,11 +300,20 @@ def paasta_status_on_api_endpoint(
...
+        if status.get(instance_type) is None:
+            output.append(
+                PaastaColors.red(
+                    f"Status contains no data for {instance_type} instance type.\n"
+                )
+            )
+            return 1
+    else:
+        instance_type = find_instance_type(status)
+        if system_paasta_config.get_enable_custom_cassandra_status_writer():
```
mattmb, line 313: do you still need this flag if the user can now chose this when they run paasta status?

siadat: That's correct, we won't need it anymore. I will remove it.

cuza (approved):

paasta_tools/cli/cmds/status.py
nemacysts: might even be able to use https://docs.python.org/3/library/argparse.html#choices (and point it at https://sourcegraph.yelpcorp.com/Yelp/paasta/-/blob/paasta_tools/instance/kubernetes.py?subtree=true#L51) so that users can't pass in invalid values

siadat: @mattmb @nemacysts 
Thank you for your suggestions! I simplified the implementation by replacing
```
--instance-type=kubernetes
```
with a simple boolean flag `--kubernetes` / `-k` for printing the status using the kubernetes writer when it is available. Overriding instance type was too confusing and was covering cases where it made no sense, but I hope the new flag self-descriptive :) Also, listed the CRs where this flag would be useful in the help message.

paasta_tools/cli/cmds/status.py
```diff
@@ -135,6 +138,19 @@ def add_subparser(
         default=DEFAULT_SOA_DIR,
         help="define a different soa config directory",
     )
+    status_parser.add_argument(
+        "-k",
+        "--kubernetes",
+        dest="kubernetes",
```
nemacysts, line 144: should this be a slightly more descriptive flag?
```suggestion
        "--force-kubernetes-status-writer",
        dest="force_kubernetes_status_writer",
```

`--kubernetes` seems a bit too generic imo

paasta_tools/cli/cmds/status.py
```diff
@@ -83,6 +84,8 @@
 from paasta_tools.utils import remove_ansi_escape_sequences
 from paasta_tools.utils import SystemPaastaConfig
 
+KUBERNETES_INSTANCE_TYPE = "kubernetes"
```
nemacysts, line 87: should this be defined in `paasta_tools.instance.kubernetes`? this file doesn't really seem like the best place for it

paasta_tools/cli/cmds/status.py
```diff
@@ -294,11 +311,20 @@ def paasta_status_on_api_endpoint(
...
+                PaastaColors.red(f"Status contains no data for {instance_type}.")
+            )
+            return 1
+    else:
+        instance_type = find_instance_type(status)
+        for instance_type_cr in INSTANCE_TYPES_K8S_CR:
+            if status.get(instance_type_cr) is not None:
+                # By default, the instance type is the CR instance type:
+                instance_type = instance_type_cr
+                break
```
nemacysts, line 327: how come we can't just use the instance type returned from `find_instance_type`?

paasta_tools/cli/cmds/status.py
siadat: The flag was removed entirely. We will print all statuses that are available.

paasta_tools/cli/cmds/status.py
siadat: This flag is remove. As you suggested, I changed the implementation to print all available statuses.

paasta_tools/instance/kubernetes.py
```diff
@@ -46,8 +46,6 @@
 
 
 INSTANCE_TYPES_CR = {"flink", "cassandracluster", "kafkacluster"}
-# TODO: remove "cassandracluster" from INSTANCE_TYPES_K8S after
-# print_cassandra_status proves to be stable.
```
siadat, line 50: I had added this comments previously, before we decided to print k8s status metadata as well. It i irrelevant now.

paasta_tools/cli/cmds/status.py
```diff
@@ -294,39 +294,44 @@ def paasta_status_on_api_endpoint(
...
-
-    if system_paasta_config.get_enable_custom_cassandra_status_writer():
-        if status.get("cassandracluster") is not None:
-            instance_type = "cassandracluster"
+    instance_types = find_instance_types(status)
+    if not instance_types:
+        print(
+            "Not implemented: Looks like %s is not a Marathon or Kubernetes instance"
+            % instance
+        )
```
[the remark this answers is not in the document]
siadat: Agreed, that's much better.

PR #3357, opened 2022-05-09
Build for Ubuntu Jammy

dbgrigsby:
## Problem or Feature
For HACK-3706
We'd like all packages Jenkins needs to be available on Ubuntu Jammy

## Solution
* Add Jammy CI configurations for github actions, jenkins
* Adjust debian/rules to use python3.8 instead of 3.7 for jammy, which the previous `focal` branch also did
* Copy paste the bionic dockerfile, adjust it to use Jammy, add / rename some packages to fit  what's available (python-tox -> tox, added sphinx-rtd-theme-common, python* -> python3.8*), make python commands use 3.8, add mesos-slave.secret file
* Commit `debian/paasta-tools.prerm.debhelper` file which was autogenerated when I ran the itest. Looks like this is automatically created whenever it's ran, and there are other auto-generated files in debian/, so it might as well be in version control if it's being used. 

## Verification
`make itest_jammy` passes


yelp_package/dockerfiles/jammy/Dockerfile
```diff
@@ -0,0 +1,59 @@
...
+        zsh > /dev/null \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN python3.8 -m pip install --upgrade pip==20.0.2
+RUN pip3.8 install virtualenv==16.0.0
+
+RUN cd /tmp && \
+    wget http://mirrors.edge.kernel.org/ubuntu/pool/universe/d/dh-virtualenv/dh-virtualenv_1.2.2-1.1_all.deb && \
+    gdebi -n dh-virtualenv*.deb && \
+    rm dh-virtualenv_*.deb
```
jvperrin, line 55: This probably isn't needed any more and can be merged back into the normal package installation stuff since the latest `dh-virtualenv` on jammy is `1.2.2` already: https://packages.ubuntu.com/jammy/dh-virtualenv

jvperrin (commented):
You probably could use 3.7 for jammy too if you want (it's on deadsnakes and we've imported it internally), although I do think switching to 3.8 at some point would be a good move.

yelp_package/dockerfiles/jammy/Dockerfile
dbgrigsby: Thanks for the comments @jvperrin 


I tried removing that highlighted bit from the dockerfile and got this error from `make itest_jammy`,  I re-added it and it went away.

```
dh: error: unable to load addon python-virtualenv: Can't locate Debian/Debhelper/Sequence/python_virtualenv.pm in @INC (you may need to install the Debian::Debhelper::Sequence::python_virtualenv module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.34.0 /usr/local/share/perl/5.34.0 /usr/lib/x86_64-linux-gnu/perl5/5.34 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.34 /usr/share/perl/5.34 /usr/local/lib/site_perl) at (eval 4) line 1.
BEGIN failed--compilation aborted at (eval 4) line 1.
```

jvperrin: Did you also add `dh-virtualenv` to the list of packages installed above this when this part was removed? If so then that's strange, maybe there's something different between the versions, but I'd be surprised for such a small version difference

dbgrigsby: I did not, just added it now and `make itest_jammy` passes, thanks!

jvperrin: Awesome, that's great, no problem!

jvperrin (approved):
Seems reasonable to me, ship it!

(probably should get a review from CI too though)

debian/rules
```diff
@@ -4,7 +4,7 @@
 PIP_INDEX_URL ?= https://pypi.yelpcorp.com/simple
 
 DISTRIB_CODENAME := $(shell sed -n 's|DISTRIB_CODENAME=\(.*\)|\1|p' </etc/lsb-release)
-PYTHON_VERSION := 3.7
+PYTHON_VERSION := $(if $(filter-out jammy,$(DISTRIB_CODENAME)),3.7,3.8)
```
nemacysts, line 7: imo, it’s probably better to stick to a single python version to reduce the cognitive overhead of having two different versions under which your code will run (and also, to minimize differences between the environment that paasta code runs under)

yelp_package/dockerfiles/jammy/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+ARG DOCKER_REGISTRY=docker-dev.yelpcorp.com/
+FROM ${DOCKER_REGISTRY}ubuntu:jammy
+
+ARG PIP_INDEX_URL=https://pypi.yelpcorp.com/simple
+ENV PIP_INDEX_URL=$PIP_INDEX_URL
+
+RUN sed -i 's/archive.ubuntu.com/us-east1.gce.archive.ubuntu.com/g' /etc/apt/sources.list
```
nemacysts, line 21: hmm, lemme do some digging tomorrow to figure out why we’re doing this in the first place - this seems a bit weird to be doing

yelp_package/dockerfiles/jammy/Dockerfile
```diff
@@ -0,0 +1,55 @@
...
+        python3.8-dev \
+        python3-pip \
+        python3.8-distutils \
+        tox \
+        wget \
+        sphinx-rtd-theme-common \
+        zsh > /dev/null \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN python3.8 -m pip install --upgrade pip==20.0.2
```
nemacysts, line 50: hmm, maybe i’ll try to modernize paasta's build setup for hackathon - it always struck me as a bit weird that we did this rather than use the normal requirements-bootstrap setup here

debian/rules
dbgrigsby: @nemacysts  Looks like taking it back to Python3.7 introduces an error, this might be why the original focal PR moved this to 3.8

From make itest_jammy on this branch, with 3.8 replaced with 3.7 in the Dockerfile and debian/control
```
Unpacking paasta-tools (0.127.2-yelp1) ...
dpkg: dependency problems prevent configuration of paasta-tools:
 paasta-tools depends on base-files (<< 11ubuntu) | python3.8; however:
  Version of base-files on system is 12ubuntu4.1.
  Package python3.8 is not installed.
  ```

yelp_package/dockerfiles/jammy/Dockerfile
nemacysts: ah yea, sorry - this was me just musing out loud, definitely not a blocker :)

debian/rules
dbgrigsby: Looks like vkhromov is not available in the near future, so we may not be able to ask them for help here. @nemacysts what do you think about shipping as-is, and deferring poking at debian/control at a later date? The hackathon ends tomorrow.

nemacysts (approved):

dbgrigsby (commented):
Your changes lgtm @nemacysts

PR #3359, opened 2022-05-17
Update git-tag format in generate-deployments-for-service

dpopacristobal:
Initially we wanted the git-tag format to be:
`paasta-{deploy_group}-{image_version}-{timestamp}-deploy`

However, this made it hard to parse when the deploy group included a `"-"`. As a result, we're instead opting for:
`paasta-{deploy_group}:{image_version}-{timestamp}-deploy`

This changes the format in the most recent changes. We'll use this format going forward in the upcoming changes.

### DIFF
--- a/paasta_tools/generate_deployments_for_service.py
+++ b/paasta_tools/generate_deployments_for_service.py
@@ -121,7 +121,7 @@ def get_latest_deployment_tag(
     most_recent_sha = None
     most_recent_image_version = None
     pattern = re.compile(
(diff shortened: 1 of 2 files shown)


jfongatyelp (commented):
Apologies, we reviewed the git tag spec and `:` is not an allowed character either. 
Instead of `:`, can we use `+` ti separate `deploy_group` and `image_version`?

jfongatyelp (approved):

EvanKrall (approved):

PR #3361, opened 2022-05-18
Show image-version in new paasta status

dpopacristobal:
This is for:
[COMPINFRA-1043](https://jira.yelpcorp.com/browse/COMPINFRA-1043)

The output should look like:
```
service: pii_logging_detection
cluster: pnw-prod
    instance: main
    Git sha:    409f8265 (desired)
    State: Running
    Running versions:
      409f8265 (image_version: none) - Started 2022-05-17 12:33:09 (20 hours ago)
        Replica States: 1 Healthy
    instance: pii_logging_detection_kafka_sampling.run
    Tron job: pii_logging_detection_kafka_sampling
      Dashboard: http://y/tron-pnw-prod#job/pii_logging_detection.pii_logging_detection_kafka_sampling
```

We could also make it so that it stays the same if image_version is unset to keep the status-quo? Happy to do it either way.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1302,11 +1302,11 @@ def get_version_table_entry(
     version_name = version.git_sha[:8]
     if show_config_sha or verbose > 1:
         version_name += f", {version.config_sha}"
+    if version.image_version is not None:
+        version_name += PaastaColors.blue(f" (image version: {version.image_version})")
     if version_name_suffix is not None:
         version_name += f" ({version_name_suffix})"
     version_name = PaastaColors.blue(version_name)
-    if version.image_version is not None:
-        version_name += PaastaColors.blue(f" (image version: {version.image_version})")
 
     start_datetime = datetime.fromtimestamp(version.create_timestamp)
     humanized_start_time = humanize.naturaltime(start_datetime)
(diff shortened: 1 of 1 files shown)


jfongatyelp:
> We could also make it so that it stays the same if image_version is unset to keep the status-quo? Happy to do it either way.
I think I lean towards keeping output the same for status-quo, but it's a pretty minor nit.

paasta_tools/cli/cmds/status.py
```diff
@@ -1305,6 +1305,8 @@ def get_version_table_entry(
     if version_name_suffix is not None:
         version_name += f" ({version_name_suffix})"
     version_name = PaastaColors.blue(version_name)
+    if version.image_version is not None:
+        version_name += PaastaColors.blue(f" (image version: {version.image_version})")
```
jfongatyelp, line 1309: I think I'd prefer this to come before `version_name_suffix` as folks will probably be used to searching for `(old)` vs `(new)` at the *end* of the line during bounces, and that would keep sha+image_version together as the complete 'version' concept we're trying to push elsewhere...  
We can play with how we present this info together as we get closer, but let's at least keep the ordering consistent w/ status quo.

This also means you could do this addition to `version_name` before we colorify things, and won't have to re-insert multiple color codes for the same color :)

jfongatyelp (approved):

PR #3364, opened 2022-05-18
COMPINFRA-1080: Update rollback to support new extra image metadata

jfongatyelp:
This adds a new DeploymentVersion type that contains just `sha` and `image_version`.  When converting this type to a string, it should only return the sha if image_version is None, but print the full version otherwise -- this should help keep user-presented output consistent w/ status quo until things actually start getting `image_version`.

Where rollback previously dealt with just listing past shas for deployment info, I've converted it to use this new class.

This required adding the `image_version` keyword to `paasta_tools.cli.cmds.mark_for_deployment.mark_for_deployment` but this leaves the actual mark-for-deployment changes to a separate task.

I confirmed that the command still works as expected w/o an `--image-version`, and the output is nearly identical with status quo:
old: https://fluffy.yelpcorp.com/i/8smJkb3vw0fH4FgJhFJXLBPs091zRzhC.html
new: https://fluffy.yelpcorp.com/i/f12z9GlRv9sRrJ0L37jl5lDgNPdn3NQr.html

### DIFF
--- a/paasta_tools/cli/cmds/rollback.py
+++ b/paasta_tools/cli/cmds/rollback.py
@@ -12,6 +12,13 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import argparse
+from typing import Collection
+from typing import Dict
+from typing import Generator
+from typing import Mapping
+from typing import Tuple
+
 from humanize import naturaltime
 
 from paasta_tools.cli.cmds.mark_for_deployment import can_user_deploy_service
@@ -23,11 +30,12 @@ from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_deploy_groups
 from paasta_tools.cli.utils import validate_full_git_sha
 from paasta_tools.cli.utils import validate_given_deploy_groups
-from paasta_tools.deployment_utils import get_currently_deployed_sha
+from paasta_tools.deployment_utils import get_currently_deployed_version
 from paasta_tools.remote_git import list_remote_refs
 from paasta_tools.utils import _log_audit
 from paasta_tools.utils import datetime_from_utc_to_local
 from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import DeploymentVersion
 from paasta_tools.utils import format_table
 from paasta_tools.utils import get_git_url
 from paasta_tools.utils import list_services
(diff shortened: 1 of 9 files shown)


paasta_tools/cli/cmds/rollback.py
```diff
@@ -193,36 +252,48 @@ def paasta_rollback(args):
         )
         return 1
 
-    git_shas = get_git_shas_for_service(service, deploy_groups, soa_dir)
+    versions = get_versions_for_service(service, deploy_groups, soa_dir)
     commit = args.commit
+    image_version = args.image_version
+    new_version = DeploymentVersion(sha=commit, image_version=image_version)
     if not commit:
```
dpopacristobal, line 259: Do we also want to check that `image_version` is set, given the line in the arg-parser docs:
`However if one is not provided, paasta rollback will instead output a list of valid versions to rollback to.`

jfongatyelp: Ah actually the arg-parser docs already state that it's only _required_ in certain circumstances in the sentence right before it: `If your service has enabled no-commit redeploys, both a commit and the extra metadata is required for paasta rollback to run`, and we have `required=False`.  

I'll add a TODO here for us to actually do the verification *if* we can detect the requested deploy group has no-commit deploys enabled, but think we can leave it w/o for now.

paasta_tools/cli/cmds/rollback.py
```diff
@@ -193,36 +252,48 @@ def paasta_rollback(args):
...
+            deploy_group=deploy_group,
+            commit=commit,
+            image_version=image_version,
         )
 
         # we could also gate this by the return code from m-f-d, but we probably care more about someone wanting to
         # rollback than we care about if the underlying machinery was successfully able to complete the request
-        if rolled_back_from != commit:
+
+        # TODO: Eventually we should include full version in rollback event, but currently leave things as status-quo
```
EvanKrall, line 292: are we not including the full version in the event? I see you changing both `rolled_back_from` and `rolled_back_to` to the new style (`version` rather than `sha`)

EvanKrall (approved):

paasta_tools/cli/cmds/rollback.py
jfongatyelp: Oh, good call --  I think in a previous iteration I conditionally set `rolled_back_to/from` to the commit or the version before I decided to just override DeployVersion's `__repr__`. I will remove this misleading comment.

PR #3365, opened 2022-05-20
Change active_shas to active_versions

dpopacristobal:
Instead of caring about `active_shas`, we want to look at `active_versions` which include the new `image_version` information. This adds another element to the tuple that used to represent `versions`; I wonder if we should just add something like `DeploymentVersion` to the API and just use that instead, happy to go with either!

### DIFF
--- a/tests/test_kubernetes_tools.py
+++ b/tests/test_kubernetes_tools.py
@@ -88,7 +88,7 @@ from paasta_tools.kubernetes_tools import ensure_namespace
 from paasta_tools.kubernetes_tools import filter_nodes_by_blacklist
 from paasta_tools.kubernetes_tools import filter_pods_by_service_instance
 from paasta_tools.kubernetes_tools import force_delete_pods
-from paasta_tools.kubernetes_tools import get_active_shas_for_service
+from paasta_tools.kubernetes_tools import get_active_versions_for_service
 from paasta_tools.kubernetes_tools import get_all_nodes
 from paasta_tools.kubernetes_tools import get_all_pods
 from paasta_tools.kubernetes_tools import get_annotations_for_kubernetes_service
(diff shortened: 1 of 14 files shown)


tests/test_kubernetes_tools.py
```diff
@@ -3144,10 +3145,22 @@ def test_get_active_shas_for_service():
                 }
             )
         ),
+        mock.Mock(
+            metadata=mock.Mock(
+                labels={
+                    "yelp.com/paasta_config_sha": "c123",
+                    "yelp.com/paasta_git_sha": "d456",
+                    "paasta.yelp.com/config_sha": "c123",
```
dpopacristobal, line 3153: What's the difference between the `yelp.com/` and the `paasta.yelp.com/` labels? I think we've only added the `image_version` to the paasta ones. Should I change that?

jfongatyelp: I believe we can safely just add `paasta.yelp.com` versions for any new labels that are specific to paasta workloads.  The `yelp.com/paasta_*` versions are there for legacy reasons, and there were some attempts in the past to move away from them afaik

jfongatyelp (commented):
I think this has the potential to be somewhat disruptive if we get it wrong.

I think a less disruptive road to maintain status quo until we migrate everything might be to just add `active_versions` to the api and only update mark-for-deployment to use active_versions instead of active_shas if it is passed extra image_version info, or if we can identify that this deploy_group has enabled no-commit redeploys.

EvanKrall:
+1 to leaving active_shas in the API -- paasta rollouts aren't atomic (prod is delayed by 2 hours normally, during which time mark-for-deployment would be looking for `active_versions` and not seeing it).

We'll either have to make the client side (mark-for-deployment/status) fall back to active_shas if active_versions isn't there, or split this into two (or three) PRs:

1. add `active_versions`, leaving `active_shas` in the API
2. Once the first PR is fully rolled out to all environments, another PR to change the client side to look for `active_versions`
3. (optionally) clean up `active_shas` (could also potentially merge this with the second PR)

dpopacristobal:
I've tried to keep `active_shas` as well as `active_versions` in the API so both of them are accessible. `mark_for_deployment` still uses `active_shas`, so hopefully that should work. Let me know if that's alright!

jfongatyelp (approved):
We actually probably could have skipped KuberentesStatus & MarathonStatus, apologies for not catching that. Either way no harm by updating them.

PR #3372, opened 2022-06-08
Add --disable-aws-credential-env-variables to spark run cli

jsleight:
This is needed for the Spark v3.2 stuff to prevent the weird race conditions with the temporary credentials provider.  I've already made changes to the other systems like service_configuration_lib to set creds in the spark configuration instead.

Is backwards compatible behavior by default so shouldn't impact any productions systems.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -353,6 +353,15 @@ def add_subparser(subparsers):
     )
 
     aws_group.add_argument(
+        "--disable-aws-credential-env-variables",
+        help="Do not put AWS credentials into environment variables.  Credentials "
+        "will still be read and set in the Spark configuration.  In Spark v3.2 "
+        "you want to set this argument.",
+        action="store_true",
+        default=False,
+    )
+
+    aws_group.add_argument(
         "--aws-region",
         help=f"Specify an aws region. If the region is not specified, we will"
         f"default to using {DEFAULT_AWS_REGION}.",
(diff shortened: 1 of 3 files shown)


88manpreet (approved):

jsleight:
Couple of nuances w.r.t. this flag:

- It only matters for the temporary credentials provider -- which is only for human runs using federated roles.  Production services with static key pairs don't need to set this, so no need for yelpsoa config changes.
- Even for temporary credentials provider this isn't strictly necessary.  All it does is eliminate a confusing non-fatal error which spark auto-recovers from.  But I know that warning/error would confuse people and this is the fix.
- Once spark-etl is upgraded most places we can make the --disable-aws-credential-env-variables True by default.  Note this is actually independent of spark versions, we just need services to expect the creds to come from the spark configs instead of env variables.  And then ultimately just make this PR the only behavior with no option to disable it.

88manpreet:
> Couple of nuances w.r.t. this flag:
> 
> * It only matters for the temporary credentials provider -- which is only for human runs using federated roles.  Production services with static key pairs don't need to set this, so no need for yelpsoa config changes.
> * Even for temporary credentials provider this isn't strictly necessary.  All it does is eliminate a confusing non-fatal error which spark auto-recovers from.  But I know that warning/error would confuse people and this is the fix.
> * Once spark-etl is upgraded most places we can make the --disable-aws-credential-env-variables True by default.  Note this is actually independent of spark versions, we just need services to expect the creds to come from the spark configs instead of env variables.  And then ultimately just make this PR the only behavior with no option to disable it.

That makes sense. Thanks for the details.

PR #3374, opened 2022-06-13
Update get_config_hash function to include boto_keys sig

Qmando:
I think that this should fix a race condition that happens in the follow case:

- Add boto_keys to a service config for the first time
- Paasta computes config hash and deploys
   - boto_keys secret is not yet synced and is not mounted
- paasta_secret_sync adds boto_keys secret and secret signature
- config hash remains unchanged and nothing happens

We'd like paasta to redeploy the service once the boto_keys secret is added. We don't want it to redeploy based no the value of the secret. Therefore I added a flag in the config hash to represent that the signature has some value, which should cause a redeployment at most once.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1966,6 +1966,13 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             service=self.get_service(), environment_variables=self.get_env()
         )
 
+        # Only record whether a boto_keys hash exists
+        # This allows the service to restart once when boto_keys secret is
+        # first added, not when the values change
+        if ahash.get("boto_keys"):
+            boto_secret_hash = self.get_boto_secret_hash()
+            ahash["boto_keys_sig"] = "present" if boto_secret_hash else ""
+
         # remove data we dont want used to hash configs
         # replica count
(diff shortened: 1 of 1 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1966,6 +1966,13 @@ def sanitize_for_config_hash(
             service=self.get_service(), environment_variables=self.get_env()
         )
 
+        # Only record whether a boto_keys hash exists
+        # This allows the service to restart once when boto_keys secret is
+        # first added, not when the values change
+        if ahash.get("boto_keys"):
+            boto_secret_hash = self.get_boto_secret_hash()
+            ahash["boto_keys_sig"] = "present" if boto_secret_hash else ""
```
nemacysts, line 1974: doesn't this end up being:
```suggestion
        if ahash.get("boto_keys") and self.get_boto_secret_hash()
            ahash["boto_keys_sig"] = "present"
```
? `get_boto_secret_hash` is typed as always returning a `str`, so i don't think we'll ever get anything but `"present"` if we're in this block

and just to confirm the behavior: we want to bounce a service when we go from having 0 -> N keys or N -> 0 keys - but not from N -> N ± 1 keys?

PR #3376, opened 2022-06-16
Update mark-for-deployment's wait-for-deployment components to use image_version

jfongatyelp:
I still have to create some tests and I'm sure I forgot something, but wanted to open this up for comment early!

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -637,7 +637,7 @@ class MarkForDeploymentProcess(SLOSlackDeploymentProcess):
         )
 
         # Keep track of each wait_for_deployment task so we can cancel it.
-        self.wait_for_deployment_tasks: Dict[str, asyncio.Task] = {}
+        self.wait_for_deployment_tasks: Dict[DeploymentVersion, asyncio.Task] = {}
 
         self.human_readable_status = "Waiting on mark-for-deployment to initialize..."
         self.progress = Progress()
@@ -1837,8 +1837,7 @@ def compose_timeout_message(
     timeout: float,
     deploy_group: str,
     service: str,
-    git_sha: str,
-    image_version: Optional[str] = None,
+    version: DeploymentVersion,
 ) -> str:
     paasta_status = []
     paasta_logs = []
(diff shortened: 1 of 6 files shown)


paasta_tools/cli/cmds/wait_for_deployment.py
```diff
@@ -81,6 +81,13 @@ def add_subparser(subparsers):
         required=True,
         type=validate_short_git_sha,
     )
+    list_parser.add_argument(
```
dpopacristobal, line 84: Do we also need to change `paasta_wait_for_deployment` in this file (and possibly the validation stuff)?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1480,7 +1529,7 @@ def check_if_instance_is_done(
     service: str,
     instance: str,
     cluster: str,
-    git_sha: str,
+    version: DeploymentVersion,
```
dpopacristobal, line 1532: Sometimes we're passing in `git_sha` and `image_version` separately, and other times together as `DeploymentVersion`, do we have a preference for one over the other in general?

dpopacristobal (commented):
Looks good to me! I'll make my changes on top of this branch and rebase later

paasta_tools/cli/cmds/mark_for_deployment.py
jfongatyelp: It's a good question; one I asked myself multiple times.  It might help keep things neater to prefer a DeploymentVersion and make sure it helps future devs conceptualize what a 'version' is better, without having to remember the set of params that make up that information.

In truth I had just been changing to deploymentversion in places where I didn't want to add `image_version` as a kwarg with default it to `None` since we have so many functions w/ only positional args (but later realized we just needed to make sure we passed `None` from its callers....)

paasta_tools/cli/cmds/wait_for_deployment.py
jfongatyelp: I was originally trying to target just mark_for_deployment.py with this PR to minimize changes, but I guess it's small enough I'll throw that in here too.

paasta_tools/utils.py
```diff
@@ -3498,6 +3499,41 @@ def format_tag(tag: str) -> str:
...
+        match = pattern.match(ref_name)
+        if match:
+            gd = match.groupdict()
+            dtime = gd["dtime"]
+            if most_recent_dtime is None or dtime > most_recent_dtime:
+                most_recent_dtime = dtime
+                most_recent_ref = ref_name
+                most_recent_sha = sha
+                most_recent_image_version = gd["image_version"]
+    return most_recent_ref, most_recent_sha, most_recent_image_version
```
dpopacristobal, line 3534: Do we want to return a DeploymentVersion here instead of returning the `sha `and the `image_version` separately?

dpopacristobal (commented):
All looks good to me!

dpopacristobal (approved):

PR #3378, opened 2022-06-21
Allow setting PreferredDuringSchedulingIgnoredDuringExecution anti-af…

EvanKrall:
…finity terms with anti_affinity_preferred key.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -102,6 +102,7 @@ from kubernetes.client import V1Subject
 from kubernetes.client import V1TCPSocketAction
 from kubernetes.client import V1Volume
 from kubernetes.client import V1VolumeMount
+from kubernetes.client import V1WeightedPodAffinityTerm
 from kubernetes.client import V2beta2CrossVersionObjectReference
 from kubernetes.client import V2beta2HorizontalPodAutoscaler
 from kubernetes.client import V2beta2HorizontalPodAutoscalerCondition
@@ -264,6 +265,10 @@ class KubeAffinityCondition(TypedDict, total=False):
     instance: str
 
 
+class KubeWeightedAffinityCondition(KubeAffinityCondition):
+    weight: int
+
+
 def _set_disrupted_pods(self: Any, disrupted_pods: Mapping[str, datetime]) -> None:
     """Private function used to patch the setter for V1beta1PodDisruptionBudgetStatus.
     Can be removed once https://github.com/kubernetes-client/python/issues/466 is resolved
@@ -336,6 +341,9 @@ class KubernetesDeploymentConfigDict(LongRunningServiceConfigDict, total=False):
     sidecar_resource_requirements: Dict[str, SidecarResourceRequirements]
     lifecycle: KubeLifecycleDict
     anti_affinity: Union[KubeAffinityCondition, List[KubeAffinityCondition]]
+    anti_affinity_preferred: Union[
+        KubeWeightedAffinityCondition, List[KubeWeightedAffinityCondition]
+    ]
     prometheus_shard: str
     prometheus_path: str
     prometheus_port: int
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1937,9 +1934,28 @@ def get_pod_anti_affinity(self) -> Optional[V1PodAntiAffinity]:
...
+        affinities with the "paasta.yelp.com" prefixed label selector
+        :return:
+        """
+
+        required_terms = self.get_pod_anti_affinity_terms(
+            config_key="anti_affinity",
+        )
+        preferred_terms = self.get_pod_anti_affinity_terms(
+            config_key="anti_affinity_preferred",
+        )
```
nemacysts, line 1951: i think ideally we'd have something that looked like:
```YAML
anti_affinity:
    preferred:
        - ...
    required:
        - ...
```

but we probably don't wanna deal with doing the migration to a new config format right now :p

paasta_tools/kubernetes_tools.py
```diff
@@ -1911,13 +1911,10 @@ def get_node_affinity(self) -> Optional[V1NodeAffinity]:
...
 
-    def get_pod_anti_affinity(self) -> Optional[V1PodAntiAffinity]:
-        """
-        Converts the given anti-affinity on service and instance to pod
-        affinities with the "paasta.yelp.com" prefixed label selector
-        :return:
-        """
-        conditions = self.config_dict.get("anti_affinity", [])
+    def get_pod_anti_affinity_terms(
+        self, config_key: str
```
nemacysts, line 1915: we can probably type config_key as a Literal, but that's probably not worth it atm

nemacysts (approved):

paasta_tools/kubernetes_tools.py
EvanKrall: easy enough, and it caught the fact that I never defined a type in `KubernetesDeploymentConfigDict` for `anti_affinity_preferred`.

paasta_tools/kubernetes_tools.py
```diff
@@ -1937,9 +1934,28 @@ def get_pod_anti_affinity(self) -> Optional[V1PodAntiAffinity]:
...
+            config_key="anti_affinity_preferred",
+        )
+
+        if required_terms is None and preferred_terms is None:
+            return None
 
         return V1PodAntiAffinity(
-            required_during_scheduling_ignored_during_execution=affinity_terms
+            required_during_scheduling_ignored_during_execution=required_terms,
+            preferred_during_scheduling_ignored_during_execution=preferred_terms,
```
EvanKrall, line 1958: d'oh, preferred_during_scheduling_ignored_during_execution needs to be a list of V1WeightedPodAffinityTerm, not V1PodAffinityTerm. I'm gonna have to un-DRY this.

nemacysts: heh, i guess that my assumption that there'd be a default weight and we could use the same class for both anti-affinities was wrong :p

nemacysts (approved):

EvanKrall:
@nemacysts how do you feel about this? Is this still worth shipping? COMPINFRA-1211

nemacysts:
> @nemacysts how do you feel about this? Is this still worth shipping? COMPINFRA-1211

@EvanKrall I think this has come up enough times that it's probably still worth having available - I do feel a little iffy about having people add this willy-nilly to their configs, but we can either use dyncow to require our approval OR just not publicize/document this too much :)

nemacysts (approved):

gmdfalk:
> > @nemacysts how do you feel about this? Is this still worth shipping? COMPINFRA-1211
> 
> @EvanKrall I think this has come up enough times that it's probably still worth having available - I do feel a little iffy about having people add this willy-nilly to their configs, but we can either use dyncow to require our approval OR just not publicize/document this too much :)

Agree, i think not advertising it will be okay and we could test the feature itself on paasta-contract-monitor.

When i release https://github.com/Yelp/paasta/pull/3512 next week, should i roll this one in as well?

PR #3381, opened 2022-06-24
Fix argparse for disabling temporary credential provider in spark cli

jsleight:
argparse struggles to parse into bool types.  `--auto-set-temporary-credentials-provider=False` resulted in `auto_set_temporary_credentials_provider="False"`, which of course equals True and meant that service_configuration_lib wouldn't disable the temporary credentials provider...

So making this be more standard "store_true" flow even if it is somewhat annoying due to the double negatives

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -362,11 +362,11 @@ def add_subparser(subparsers):
     )
 
     aws_group.add_argument(
-        "--auto-set-temporary-credentials-provider",
-        help="Automatically set the TemporaryCredentialsProvider if a session token "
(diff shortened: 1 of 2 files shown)


88manpreet (approved):
Interesting!

88manpreet:
Did the make test pass? 
I am getting this error. 
`        auto_set_temporary_credentials_provider = (
>           args.disable_temporary_credentials_provider is False
        )
E       AttributeError: 'Namespace' object has no attribute 'disable_temporary_credentials_provider'
`

jsleight:
yep all the tests passed.  If you're doing dev work in the repo then your test setup needs to explicitly set the disable_temporary_credentials_provider param in the Namespace object.  See the test file in the PR for an example

PR #3382, opened 2022-06-24
Add default timeout for all remaining spark jobs

88manpreet:
MLCOMPUTE-184
We have added optimal timeouts for 77% of batches. 
* Adding auto-timeout for remaining batches and for any new batches.
* Adding option to manually add timeout with argument

Will push it on Monday.

Tests:
* Local
* Manual dev runs

Monitoring:
* We will get notified on our slack notifications channel for any timed out jobs, if need manual adjustment


jsleight (approved):

jsleight:
I think it would be good to send an announcement to spark-users@ to inform of this too.  We obviously don't expect any issues but good to be cautious in case there is some job which takes longer than the default.

Buffer0x7cd:
Rolling back this PR due to issue with spark jobs
slack thread: https://yelp.slack.com/archives/CA05GTDB9/p1656582853745069

PR #3393, opened 2022-07-13
dar-1636: handle envoy admin timeout

BeryJu:


TheNavigat:
Hmmm, not sure if handling it that way is ideal

We should probably show the user that a particular host is not behaving nicely, instead of logging internally and just returning a potentially incomplete backends list

BeryJu:
> Hmmm, not sure if handling it that way is ideal
> 
> We should probably show the user that a particular host is not behaving nicely, instead of logging internally and just returning a potentially incomplete backends list

I dont think knowing that a particular host is bad is not really helpful for an enduser, hence I'm just kinda logging the error and moving on there. The error should be reported to us either by an active healthcheck checking envoy, or a splunk query that checks for this

BeryJu:
Well it wouldn't skip anything, it would just not crash if an envoy instance is timing out, and I dont think it makes sense to show that error to users as its not actionable for them, the only thing they can do is reach out to us, where we should have direct monitoring for this instead

paasta_tools/envoy_tools.py
```diff
@@ -217,17 +221,22 @@ def get_multiple_backends(
...
-    )
-
-    casper_endpoints = get_casper_endpoints(clusters_info)
-
     backends: DefaultDict[
         str, List[Tuple[EnvoyBackend, bool]]
     ] = collections.defaultdict(list)
+
+    try:
+        clusters_info = retrieve_envoy_clusters(
```
nemacysts, line 229: we talked a little about this in slack and i'm just dumping this here so that we don't forget:

perhaps we should randomize the box we use to check backends so that we don't completely break paasta status when the deterministic way we currently use always picks the same broken box (and maybe still do this change as well? bonus if we can figure out how to plumb this back to more user-facing code and have them know that the backend list is incomplete)

mattmb:

PR #3377, opened 2022-06-16
[WIP] Add flink endpoints to paasta api

AhmedSoliman7:

PR #3394, opened 2022-07-13
[FLINK-4114] Remove job-related output from paasta status

marcos-sb:
This is another version of https://github.com/Yelp/paasta/pull/3377. In this version we don't output any job-related metadata to avoid having to pull them from the Job Manager(s).


nemacysts (approved):

marcos-sb:
Merged the other version (#3377)

PR #3395, opened 2022-07-13
Enable spark.sql.adaptive.enabled to true for all spark jobs

88manpreet:
1. Enabled AQE: adaptive query execution by default.
2. Log Signalfx url and final command. There were not originally logged in tron web ui. 
Useful for debugging.

Test result added as a comment in ticket: [MLCOMPUTE-17](https://jira.yelpcorp.com/browse/MLCOMPUTE-17)

Rollout plan:
1. Live Experiment Stage: Roll this pr out
2. Monitoring: 1 week: Observe on internal channel for any unexpected behavior, impacted runtimes of spark batches
- Rollback, debug and fix any unprecedented issues
4. GA: Roll the change out on service_configuration_lib and rollback this change

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -863,7 +863,7 @@ def configure_and_run_docker_container(
             else:
                 raise
 
-    final_spark_submit_cmd_msg = f"Executing spark-submit command: {docker_cmd}"
+    final_spark_submit_cmd_msg = f"Final command: {docker_cmd}"
     print(PaastaColors.grey(final_spark_submit_cmd_msg))
     log.info(final_spark_submit_cmd_msg)
     return run_docker_container(
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -805,19 +805,26 @@ def configure_and_run_docker_container(
     )  # type:ignore
 
     webui_url = get_webui_url(spark_conf["spark.ui.port"])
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1108,6 +1118,13 @@ def paasta_spark_run(args):
         needs_docker_cfg=needs_docker_cfg,
         auto_set_temporary_credentials_provider=auto_set_temporary_credentials_provider,
     )
+    # Experimental: TODO: Move to service_configuration_lib once confirmed that there are no issues
+    # Enable AQE: Adaptive Query Execution
+    if "spark.sql.adaptive.enabled" not in spark_conf:
+        spark_conf["spark.sql.adaptive.enabled"] = "true"
```
jsleight, line 1124: ultimately I think we want this to go in spark-on-paasta right?  I want to avoid a situation where we have configs being specified from all these different places since that will make it hard to track things down later.

88manpreet: I think ultimately, I would prefer to have all configs being either set here (paasta/spark_run.py) or service_configuration_lib (where we set other dynamically calculated spark configs). 
spark-on-paasta poses the limitation to rollout configs to all batches easily.

As on now keeping it here, as it's easier to rollout and rollback. Once the experiment ok for a week, I will move this to `service_configuration_lib` where some of the other spark configs are set. 

Agreed on the part, that we want to avoid spark configs being spread out into different repositories.

jsleight (approved):

PR #3397, opened 2022-07-14
Remove mentions of norcal-prod

nemacysts:
This cluster no longer exists and mentions of it confuses folks.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -463,9 +463,9 @@ out of the load balancer, so it justifies having less sensitive thresholds.
 ``marathon-[clustername].yaml``
 -------------------------------
 
-e.g. ``marathon-norcal-prod.yaml``, ``marathon-mesosstage.yaml``. The
+e.g. ``marathon-pnw-prod.yaml``, ``marathon-mesosstage.yaml``. The
 clustername is usually the same as the ``superregion`` in which the cluster
-lives (``norcal-prod``), but not always (``mesosstage``). It MUST be all
+lives (``pnw-prod``), but not always (``mesosstage``). It MUST be all
 lowercase. (non alphanumeric lowercase characters are ignored)
 
 **Note:** All values in this file except the following will cause PaaSTA to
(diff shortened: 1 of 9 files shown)


paasta_tools/marathon_dashboard.py
```diff
@@ -91,7 +91,7 @@ def create_marathon_dashboard(
     dashboard_links = system_paasta_config.get_dashboard_links()
     marathon_links = dashboard_links.get(cluster, {}).get("Marathon RO")
 
-    # e.g. 'http://10.64.97.75:5052': 'http://marathon-norcal-prod.yelpcorp.com'
+    # e.g. 'http://10.64.97.75:5052': 'http://marathon-pnw-prod.yelpcorp.com'
```
jvperrin, line 94: IP is wrong here too, but I don't think it particularly matters

jfongatyelp (approved):

docs/source/yelpsoa_configs.rst
```diff
@@ -463,9 +463,9 @@ out of the load balancer, so it justifies having less sensitive thresholds.
 ``marathon-[clustername].yaml``
 -------------------------------
 
-e.g. ``marathon-norcal-prod.yaml``, ``marathon-mesosstage.yaml``. The
+e.g. ``marathon-pnw-prod.yaml``, ``marathon-mesosstage.yaml``. The
```
jvperrin, line 466: Is `mesosstage` still around?

jvperrin (approved):

paasta_tools/marathon_dashboard.py
nemacysts: you got me there :p

not sure how worth it is actually getting an ip that would be right if we were still running marathon tho (this file really should just be deleted alongside all the other marathon stuff)

docs/source/yelpsoa_configs.rst
nemacysts: nope! i half-debated re-writing this, but i think we're waiting to just delete all the mesos mentions in one swoops so i decided against investing too much time/effort into cleaning mentions of marathon up

(maybe once someone mentions they got confused by what mesosstage/marathon are after reading this :p)

PR #3400, opened 2022-07-15
Enable starting arbitrary # of paasta-api workers

nemacysts:
Until we make this an actual service, we can "scale up" by adding more
gunicorn workers.

### DIFF
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -254,7 +254,7 @@ def main(argv=None):
         os.path.join(sys.exec_prefix, "bin", "gunicorn"),
         "gunicorn",
         "-w",
-        "4",
+        str(load_system_paasta_config().get_paasta_api_number_workers()),
         "--bind",
         f":{args.port}",
         "--timeout",
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -1889,6 +1889,8 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     auto_config_instance_types_enabled: Dict[str, bool]
     auto_hostname_unique_size: int
     boost_regions: List[str]
+    cluster: str
+    cluster_aliases: Dict[str, str]
     cluster_autoscaler_max_decrease: float
     cluster_autoscaler_max_increase: float
     cluster_autoscaling_draining_enabled: bool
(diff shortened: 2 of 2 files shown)


paasta_tools/utils.py
```diff
@@ -1942,16 +1944,17 @@ class SystemPaastaConfigDict(TypedDict, total=False):
...
+    mark_for_deployment_default_diagnosis_interval: float
+    mark_for_deployment_default_polling_interval: float
+    mark_for_deployment_max_polling_threads: int
     mark_for_deployment_should_ping_for_unhealthy_pods: bool
     mesos_config: Dict
     metrics_provider: str
     monitoring_config: Dict
     nerve_readiness_check_script: List[str]
     nerve_register_k8s_terminating: bool
+    paasta_api_number_workers: int
```
nemacysts, line 1957: this is the only new change here - i just sorted what seems to have originally been a sorted list until we stopped caring :p

paasta_tools/api/api.py
```diff
@@ -254,7 +254,7 @@ def main(argv=None):
         os.path.join(sys.exec_prefix, "bin", "gunicorn"),
         "gunicorn",
         "-w",
-        "4",
+        str(load_system_paasta_config().get_paasta_api_number_workers()),
```
nemacysts, line 257: i'll see if i can get puppet to restart the systemd service whenever the config file that this is in gets updated - otherwise, we'd need to manually do that whenever we want to change this value

jfongatyelp: Oh I forgot we pass args to the api (like port, timeout) anyway, so if we made this an arg, we could control that in puppet anyway and maybe make the reloading a little less complicated.

But this works too

jfongatyelp (approved):

paasta_tools/api/api.py
nemacysts: oh, true...

that might be a better way to do things!

paasta_tools/utils.py
```diff
@@ -2686,6 +2687,9 @@ def get_cluster_aliases(self) -> Dict[str, str]:
     def get_hacheck_match_initial_delay(self) -> bool:
         return self.config_dict.get("hacheck_match_initial_delay", False)
 
+    def get_paasta_api_number_workers(self) -> int:
+        return self.config_dict.get("paasta_api_number_workers", 4)
+
```
EvanKrall, line 2692: Does this get called anywhere?

nemacysts: whoops - didn't save this before before commiting

jfongatyelp (approved):

PR #3403, opened 2022-07-20
Randomize host selected in mesh_status

nemacysts:
We sometimes see single hosts have envoy issues, and having these
cascade into paasta status breaking for the pool these hosts are in is
not a great experience.

To minimize the impact here, grab a random host from the target pool
rather than the "first".

### DIFF
--- a/paasta_tools/instance/kubernetes.py
+++ b/paasta_tools/instance/kubernetes.py
@@ -325,7 +325,7 @@ async def mesh_status(
 
     pods = await pods_task
     for location, hosts in node_hostname_by_location.items():
-        host = replication_checker.get_first_host_in_pool(hosts, instance_pool)
+        host = replication_checker.get_hostname_in_pool(hosts, instance_pool)
         if service_mesh == ServiceMesh.SMARTSTACK:
             mesh_status["locations"].append(
                 _build_smartstack_location_dict(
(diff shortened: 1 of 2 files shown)


EvanKrall (approved):

jfongatyelp (approved):

paasta_tools/instance/kubernetes.py
```diff
@@ -325,7 +325,7 @@ async def mesh_status(
 
     pods = await pods_task
     for location, hosts in node_hostname_by_location.items():
-        host = replication_checker.get_first_host_in_pool(hosts, instance_pool)
+        host = replication_checker.get_hostname_in_pool(hosts, instance_pool)
```
BeryJu, line 328: do we use `get_first_host_in_pool` for anything else that could cause issues in the future?

BeryJu (approved):

paasta_tools/instance/kubernetes.py
EvanKrall: https://sourcegraph.yelpcorp.com/search?q=get_first_host_in_pool&patternType=literal the only references are in paasta.

(though it looks like maybe there's a test that needs updating?)

nemacysts: there's just tests for the get_* functions - i thought about adding one for this, but it seems like i'd really just be checking that random.choice was called?

PR #3405, opened 2022-07-21
Reusing the same python process to make IntelliJ and VSCode debuggers happier

cuza:
Reusing the same python process to make IntelliJ and VSCode debuggers happier.


nemacysts (commented):
i'm not particularly a fan of having how the api is run in development be different from how it's actually run in prod, but this seems like a small enough difference (after we dedupe the args) that it'll be fine

nemacysts (approved):

bobokun (approved):

EmanElsaban (approved):

PR #3409, opened 2022-07-25
Bump timeout for readiness probe

TheNavigat:
This caused issues with the readiness probe where the timeout would kill the check too early


Buffer0x7cd (approved):

TheNavigat:
Closing in favour of reverting

PR #3410, opened 2022-07-25
MESH-1439: add SERVICE_REGISTRATIONS env to hacheck

BeryJu:
https://github.com/Yelp/hacheck/pull/42 and https://github.com/Yelp/nerve-tools/pull/94

(Yes this will cause a big bounce also, hopefully we can roll this out with 0.135.3)

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -966,6 +966,11 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         else:
             readiness_probe = None
 
+        hacheck_registrations_env = V1EnvVar(
+            name="SERVICE_REGISTRATIONS",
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -955,6 +955,11 @@ def get_hacheck_sidecar_container(
         else:
             readiness_probe = None
 
+        hacheck_registrations_env = V1EnvVar(
+            name="SERVICE_REGISTRATIONS",
+            value=" ".join(self.get_registrations()),
+        )
```
nemacysts, line 961: could always make this a function so that further below we could just do 
```Python
env=self.get_kubernetes_environment() + self.get_hacheck_env()
```, but that's pretty nit-picky and only slightly more aestheically pleasing :p

nemacysts (approved):

PR #3411, opened 2022-07-25
Revert "Wrap readiness check script with a timeout"

TheNavigat:
Reverts Yelp/paasta#3402


nemacysts (approved):
i guess we can try this again once we fix the hacheck autotune to not be so low :p

PR #3412, opened 2022-07-26
Fix regex to also assert timeout values of type float

88manpreet:
This is to fix the issue that if the users provide timeout as floating hours (instead of integer ones) example: 1.0h, it would still use the default timeout value as 12h.

Few more manual test runs: https://fluffy.yelpcorp.com/i/wFH1MlCVtZR5W82DWxMm5VBw9bHGgDJp.html

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -1003,7 +1003,7 @@ def _auto_add_timeout_for_job(cmd, timeout_job_runtime):
         return cmd
     try:
         timeout_present = re.match(
-            r"^.*timeout[\s]+[\d]*[m|h][\s]+spark-submit .*$", cmd
+            r"^.*timeout[\s]+[\d]+[\.]?[\d]*[m|h][\s]+spark-submit .*$", cmd
         )
         if not timeout_present:
             split_cmd = cmd.split("spark-submit")


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1003,7 +1003,7 @@ def _auto_add_timeout_for_job(cmd, timeout_job_runtime):
         return cmd
     try:
         timeout_present = re.match(
-            r"^.*timeout[\s]+[\d]*[m|h][\s]+spark-submit .*$", cmd
+            r"^.*timeout[\s]+[\d]+[\.]?[\d]*[m|h][\s]+spark-submit .*$", cmd
```
nlspoon, line 1006: @88manpreet not sure exactly how the timeout value is parsed later on but there are a few edge cases that might cause issues?
`timeout 0.2h` does match
`timeout .2h` does not match
`timeout 2.h` does match
Not sure if any of those will cause issues.

nlspoon (approved):

paasta_tools/cli/cmds/spark_run.py
88manpreet: Thanks for checking for further cases. 
It should be ok, though.

PR #3413, opened 2022-07-27
Wrap readiness probes in flock + timeout command, and make the wrappe…

EvanKrall:
…r configurable in /etc/paasta. PAASTA-17673

I'm up for debate about the timeout - maybe it should be equal to the configured timeout. The ability of liveness probe processes to outlive the timeout will break once we migrate off of `dockershim`, per the note on https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -917,6 +917,17 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             sidecars.append(uwsgi_exporter_container)
         return sidecars
 
+    def get_readiness_check_prefix(
+        self,
+        system_paasta_config: SystemPaastaConfig,
+        initial_delay: float,
+        period_seconds: float,
+    ) -> List[str]:
+        return [
+            x.format(initial_delay=initial_delay, period_seconds=period_seconds)
+            for x in system_paasta_config.get_readiness_check_prefix_template()
+        ]
+
     def get_hacheck_sidecar_container(
         self,
         system_paasta_config: SystemPaastaConfig,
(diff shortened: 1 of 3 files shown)


paasta_tools/utils.py
```diff
@@ -1958,6 +1958,7 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     pki_backend: str
     pod_defaults: Dict[str, Any]
     previous_marathon_servers: List[MarathonConfigDict]
+    readiness_check_prefix_template: str
```
EvanKrall, line 1961: whoops, this needs to be a list

paasta_tools/kubernetes_tools.py
```diff
@@ -936,14 +947,20 @@ def get_hacheck_sidecar_container(
...
+                        period_seconds=period_seconds,
+                    )
+                    + self.get_readiness_check_script(system_paasta_config)
                     + [str(self.get_container_port())]
                     + self.get_registrations()
                 ),
                 initial_delay_seconds=initial_delay,
-                period_seconds=10,
+                period_seconds=period_seconds,
             )
```
jfongatyelp, line 964: Should we also explicitly set our V1Probe's timeout_seconds=period_seconds (or anything >1) here, to better mimic our pre-1.18 behavior and give our first few readiness probes a better chance to succeed?

nemacysts: they should actually keep running regardless of the `timeout_seconds` value since we're still using dockershim - and i've noticed that once at least one of these succeeds, then future probes are way more likely to succeed/run faster

that said, we do probably still wanna figure out why these are slow at the start (which is probably 'cause we're giving this sidecar too little cpu)

jfongatyelp: hmm I thought from the 1.20 note in https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#configure-probes this implies the default timeout_seconds is now being used for the probe success, which is why readiness is failing more often (since it now needs to happen within the default 1s)?

nemacysts: ahh, ok - i misunderstood you! umm, i think either way will more or less have the same effect: once one of these actually passes the ones after should also pass

that said, we might want to do it just so that if we're still in a funky state by the time we stop using dickershim we run into less surprises :p

nemacysts (approved):
we'll be able to get rid of this flock/timeout stuff once we're off of dockershim, right? might be worth adding a comment about that so that we don't forget about this in the future

that said: we may still want to figure out what to do with our internal autotune system if these are being too slow in the first place 'cause of throttling

EvanKrall:
We'll definitely be able to get rid of the `timeout` part of it, and if our timeout is less than the interval, we should be able to get rid of flock too.

PR #3415, opened 2022-07-27
cli: have push-to-registry handle multiple tag conventions

piax93:
This change makes it so that when `--image-version` is specified, `push-to-registry` will push to the Docker registry both standard tag and the "versioned tag" for the built image, allowing easier backwards compatibility.

### DIFF
--- a/paasta_tools/cli/cmds/push_to_registry.py
+++ b/paasta_tools/cli/cmds/push_to_registry.py
@@ -75,7 +75,12 @@ def add_subparser(subparsers: argparse._SubParsersAction) -> None:
         type=str,
         required=False,
         default=None,
-        help="Extra version metadata to use when naming the remote image",
+        help=(
+            "Extra version metadata to use when naming the remote image."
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/push_to_registry.py
```diff
@@ -166,6 +169,57 @@ def paasta_push_to_registry(args: argparse.Namespace) -> int:
...
+    image_version_tag = build_docker_tag(service, commit, image_version)
+    returncode, _ = _run(
+        ["docker", "tag", image_version_tag, unversioned_tag],
+        timeout=30,
+        log=True,
+        component="build",
+        service=service,
+        loglevel="debug",
+    )
+    return returncode
```
jfongatyelp, line 220: I think this could lead to a little confusion down the line since the service's `make itest` or `cook-image` is really supposed to be handling all the tagging... but given we're talking about the exact same image, this seems like a clever way to minimize any jenkins-side logic!

jfongatyelp (approved):

PR #3418, opened 2022-07-29
[FLINK-4097] Get Flink cluster status via JM API

marcos-sb:
re: https://github.com/Yelp/paasta/pull/3399

From the linked PR:
_This PR exposes a new set of endpoints to talk to the Flink JM API to pull job-related information.
This change also prepares the paasta client to manage flink clusters once the new version of the flink operator is merged. This new version of the flink operator doesn't consume the JM API to attach flink job metadata to the CRD, and instead expects paasta to pull that information to display to the operator running paasta status._

This PR intends to reapply https://github.com/Yelp/paasta/pull/3399 but also including a client fallback not to use the new endpoints if the backend API does not include them -- re: 7512eb5bf559e72298f59fa2ac0c6e70f31067e6. Credit to @jfongatyelp for the suggestion.


marcos-sb:
The code for both paths (fallback and 'regular') has been duplicated (vs. factored) on purpose. This way, once every backend is on >= 0.138.0 we can very easily remove the fallback path.

nemacysts (approved):

nemacysts (approved):

PR #3421, opened 2022-08-02
CID-2427: Replacing raw 'node' with 'ynode'

conancain:
As part of K8s migration, we want to replace raw 'node' usage with 'ynode' so we can better control what node executors we use.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,7 +13,7 @@ commit = ''
 
 ircMsgResult(CHANNELS) {
     ystage('Test') {
-        ynode.forConfiguredHostType(ownerName: 'Yelp', repoName: 'paasta') {
+        ynode {
             ensureCleanWorkspace {
                 commit = clone(
                     PACKAGE_NAME,
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -32,7 +32,7 @@ ircMsgResult(CHANNELS) {
     )
 
     ystage('Upload to PyPi') {
-        ynode.forConfiguredHostType(ownerName: 'Yelp', repoName: 'paasta') {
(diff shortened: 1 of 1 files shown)


darealMj (approved):

Jenkinsfile
```diff
@@ -13,7 +13,7 @@ commit = ''
 
 ircMsgResult(CHANNELS) {
     ystage('Test') {
-        node {
+        ynode.forConfiguredHostType(ownerName: 'Yelp', repoName: 'paasta') {
```
nemacysts, line 16: is this going to do the right thing internally? there's no Yelp org in our GHE

conancain: it won't from what I see. basically this is used to check whether owner/repo name is in a mesos-opt-in list. In this case, we can just add "Yelp/paasta" to the list if we ever want to use mesos for it

nemacysts: should we just use a raw ynode? there's also no jenkins.yaml here so there's no host type to automatically pick up

conancain:
```suggestion
        ynode {
```

Jenkinsfile
```diff
@@ -32,7 +32,7 @@ ircMsgResult(CHANNELS) {
     )
 
     ystage('Upload to PyPi') {
-        node {
+        ynode.forConfiguredHostType(ownerName: 'Yelp', repoName: 'paasta') {
```
conancain, line 35:
```suggestion
        ynode {
```

nemacysts (approved):

nemacysts:
https://github.com/Yelp/paasta/pull/3427

PR #3422, opened 2022-08-03
PAASTA-17666: Fixes paasta validate to allow preceding comments

bobokun:
## Problem or Feature
Currently, paasta validate fails when entering any preceding comments in yelpsoa-configs.

## Solution
This PR will fix the failure and join together multiline comments.

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -507,16 +507,24 @@ def _get_comments_for_key(data: CommentedMap, key: Any) -> Optional[str]:
     # this is a little weird, but ruamel is returning a list that looks like:
     # [None, None, CommentToken(...), None] for some reason instead of just a
     # single string
-    raw_comments = [
-        comment.value for comment in data.ca.items.get(key, []) if comment is not None
-    ]
+    # Sometimes ruamel returns a recursive list of as well that looks like
+    # [None, None, [CommentToken(...),CommentToken(...),None], CommentToken(...), None]
+    def _flatten_comments(comments):
+        for comment in comments:
+            if comment is None:
+                continue
+            if isinstance(comment, list):
+                yield from _flatten_comments(comment)
+            else:
+                yield comment.value
+
+    raw_comments = [*_flatten_comments(data.ca.items.get(key, []))]
     if not raw_comments:
         # return None so that we don't return an empty string below if there really aren't
         # any comments
         return None
-    # there should really just be a single item in the list, but just in case...
+    # joining all comments together before returning them
     comment = "".join(raw_comments)
-
     return comment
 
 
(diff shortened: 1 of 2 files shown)


jfongatyelp (approved):

paasta_tools/cli/cmds/validate.py
```diff
@@ -507,16 +507,24 @@ def _get_comments_for_key(data: CommentedMap, key: Any) -> Optional[str]:
...
+            else:
+                yield comment.value
+
+    raw_comments = [*_flatten_comments(data.ca.items.get(key, []))]
     if not raw_comments:
         # return None so that we don't return an empty string below if there really aren't
         # any comments
         return None
-    # there should really just be a single item in the list, but just in case...
+    # joining all comments together before returning them
```
nemacysts, line 526: if we're going to return a potentially multi-line comment now, do we need to update the regex in https://github.com/Yelp/paasta/blob/458a683ee9a5de6af0277847bcfb210b791275b5/paasta_tools/cli/cmds/validate.py#L606?

bobokun: I was actually discussing with Jen on this as well. I could update the regex pattern or another alternative is to discard the comments above/before and only keep the one that is on the cpu line.

We also noticed a bug where we don't have any unit tests checking for cpu autotune override so although paasta validate was failing, it was passing the unit tests.

nemacysts (approved):
we probably wanna add the required comment to any instances that snuck by when the validation was broken

bobokun:
> we probably wanna add the required comment to any instances that snuck by when the validation was broken

Is that something we want to change in yelpsoa-configs repo manually one time and through a commit?
Do we also want to code PAASTA-17522 as the ticket in all the comments?

jfongatyelp (approved):

paasta_tools/cli/cmds/validate.py
```diff
@@ -507,16 +507,24 @@ def _get_comments_for_key(data: CommentedMap, key: Any) -> Optional[str]:
...
+    def _flatten_comments(comments):
+        for comment in comments:
+            if comment is None:
+                continue
+            if isinstance(comment, list):
+                yield from _flatten_comments(comment)
+            else:
+                yield comment.value
+
+    raw_comments = [*_flatten_comments(data.ca.items.get(key, []))]
```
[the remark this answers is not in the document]
bobokun: While doing further testing I've found ruamel to be very inconsistent with its return behaviour. For example when using the a base template such as below will return both the comments in line 3/4
```yaml
_base_template: &base_template
  registrations: ["experiment_allocation.main"]
  # this comment will be returned by raumel
  cpus: 1  # this comment will ALSO be returned by raumel

main:
  <<: *base_template
  deploy_group: prod.non_canary
...
```

however when not using a base template such as it will NOT return the comment in line 2 but will only return comment in line 3.
```yaml
main:
  # this comment will be ignored
  cpus: 1 #only this comment wil be returned by raumel
```

However in both examples any comments written below the cpus line will be returned.

nemacysts (approved):

PR #3425, opened 2022-08-04
[FLINK-4124] Small fix to display status info

marcos-sb:
Fix for the missing info that was reported earlier today on Slack.


nemacysts (approved):
whoops!

marcos-sb:
> whoops!

I was really surprised that something like this could fall through. I'm wondering if there's any linter for python that would complain on dead code

PR #3426, opened 2022-08-04
Use execlp instead of execlpe

88manpreet:
[MLCOMPUTE-436](https://jira.yelpcorp.com/browse/MLCOMPUTE-436)

Based on PR: https://github.com/Yelp/paasta/pull/3211 originally written by @benbariteau.
This should likely fix the docker daemon not found issue.

Definition: https://docs.python.org/3/library/os.html
```...For [execle()](https://docs.python.org/3/library/os.html#os.execle), [execlpe()](https://docs.python.org/3/library/os.html#os.execlpe), [execve()](https://docs.python.org/3/library/os.html#os.execve), and [execvpe()](https://docs.python.org/3/library/os.html#os.execvpe) (note that these all end in “e”), the env parameter must be a mapping which is used to define the environment variables for the new process (these are used instead of the current process’ environment); the functions [execl()](https://docs.python.org/3/library/os.html#os.execl), [execlp()](https://docs.python.org/3/library/os.html#os.execlp), [execv()](https://docs.python.org/3/library/os.html#os.execv), and [execvp()](https://docs.python.org/3/library/os.html#os.execvp) all cause the new process to inherit the environment of the current process....```

The reason I think this should work, AIU, is `execlpe` uses new given environment variable (currently None, I think).
`execlp` uses the previous process's env variables, which I think has DOCKER_HOST env variable.

Using the same logic as `local-run` to pass all the parent process and spark specific environment.

`Testing:`
a) Manual test accessing S3
b) Unit test

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -694,7 +694,7 @@ def run_docker_container(
         print(json.dumps(docker_run_cmd))
         return 0
 
-    os.execlpe("paasta_docker_wrapper", *docker_run_cmd)
+    os.execlp("paasta_docker_wrapper", *docker_run_cmd)
     return 0
 
 
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -4,7 +4,6 @@ import logging
 import os
 import re
 import shlex
-import shutil
 import socket
 import sys
 import uuid
@@ -473,8 +472,6 @@ def get_docker_run_cmd(
     cmd.append("--rm")
     cmd.append("--net=host")
 
-    sensitive_env = {}
-
     non_interactive_cmd = ["spark-submit", "history-server"]
     if not any(c in docker_cmd for c in non_interactive_cmd):
         cmd.append("--interactive=true")
@@ -486,7 +483,6 @@ def get_docker_run_cmd(
     for k, v in env.items():
         cmd.append("--env")
         if k in SENSITIVE_ENV:
-            sensitive_env[k] = v
             cmd.append(k)
         else:
             cmd.append(f"{k}={v}")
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: paasta_tools/cli/cmds/spark_run.py [hunk 1]
```diff
@@ -497,7 +494,6 @@ def get_docker_run_cmd(
         cmd.append("--volume=%s" % volume)
     cmd.append("%s" % docker_img)
     cmd.extend(("sh", "-c", docker_cmd))
-    cmd.append(sensitive_env)
```
nemacysts, line 500: how did this even work before?

nemacysts (approved):
lgtm, although it might be worth chatting with #security real fast to make sure this isn't going to do something silly with the sensitive env vars that the previous code was somehow protecting us from

paasta_tools/cli/cmds/spark_run.py
88manpreet: `execlpe` expected last argument as an environment as dictionary. Looks like the only main env variables which mattered the spawned process are the aws creds. This was appended as last element of the list `docker_run_cmd`

88manpreet:
> lgtm, although it might be worth chatting with #security real fast to make sure this isn't going to do something silly with the sensitive env vars that the previous code was somehow protecting us from

Earlier 
* the sensitive args were skipped being pass an env variable: `--env (nothing) --env SPARK_OPTS <xyz>` 
I kept this behavior the same.

* the sensitive args were pass as env as part of `execlp`
* I kept the behavior same.

paasta_tools/cli/cmds/spark_run.py
88manpreet: tl;dr: Sensitive env variables a.k.s aws creds were skipped to be added as part of the actual command (visible in logs). 
But passed as part of env variables.
I am still keeping the behavior of skipping adding them to `--env` the same.

PR #3427, opened 2022-08-04
Run internal tests on jenkins-k8s

nemacysts:
all the deleted tests are testing functionality we no longer run internally


tox.ini
```diff
@@ -65,28 +65,6 @@ commands =
     sphinx-apidoc -f -e -o docs/source/generated/ paasta_tools
     sphinx-build -b html -d docs/build/doctrees docs/source docs/build/html
 
-[testenv:paasta_itests]
```
jfongatyelp, line 68: There seems to be a [testenv:paasta_itests_inside_container] environment too that sets `changedir=paasta_itests`, we can probably clean up that one too?

jfongatyelp (approved):
TIL paasta_itests were just marathon things.

conancain (approved):

PR #3428, opened 2022-08-08
Revert "[FLINK-4124] Small fix to display status info"

mhaseebmlk:
Reverts Yelp/paasta#3425

This seems to be affecting the `check_job_status.py` check for cicd staleness. @jfongatyelp and @nemacysts suggested reverting these changes.


nemacysts:
@mhaseebmlk probably also want to revert https://github.com/Yelp/paasta/pull/3418 to go back to the old behavior

mhaseebmlk:
Created https://github.com/Yelp/paasta/pull/3429 instead of this

PR #3430, opened 2022-08-08
Print paasta status per instance and dont wait for retrieval of entire cluster's status

sidsatyelp:
https://jira.yelpcorp.com/browse/PAASTA-17657
Break up paasta status output

Currently paasta status waits for the output of an entire cluster to be retrieved before printing it. In this PR I have changed this behavior to print instances within a cluster as they are retrieved.


paasta_tools/cli/cmds/status.py
```diff
@@ -2643,10 +2655,9 @@ def paasta_status(args) -> int:
         tasks = [executor.submit(t[0], **t[1]) for t in tasks]  # type: ignore
         for future in concurrent.futures.as_completed(tasks):  # type: ignore
             return_code, output = future.result()
-            print("\n".join(output))
             return_codes.append(return_code)
 
-    return max(return_codes)
+    return max(return_codes), lock
```
EvanKrall, line 2660: Are you returning the lock object here just for testing? I think this function needs to return just an int because it's called by https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cli.py#L255 which is expecting an int.

tests/cli/test_cmds_status.py
```diff
@@ -864,7 +866,7 @@ def test_status_with_registration(
         service_instance=None,
         new=False,
     )
-    return_value = paasta_status(args)
+    return_value, lock = paasta_status(args)
```
EvanKrall, line 869: There are a few ways you could write this test (and the other test that follows the same pattern) without having `paasta_status` return the lock object:

# mocking

You could change these lines to:
```
with mock.patch("paasta_tools.cli.cmds.status.Lock", autospec=True) as mock_lock:
    return_value = paasta_status(args)
```
which would make it so when `paasta_status` calls `Lock()`, it's calling a MagicMock object instead (which would return another MagicMock.)

then in your `assert_called_once_with` call, you'd specify `lock=mock_lock()`.

You could also pre-define a fake lock object:

```
fake_lock = mock.MagicMock()
with mock.patch("paasta_tools.cli.cmds.status.Lock", autospec=True, return_value=fake_lock):
    return_value = paasta_status(args)
```
then later, you'd pass `lock=fake_lock` to `assert_called_once_with`.

# mock.ANY

You could just specify `lock=mock.ANY` on line 883, which would make assert_called_once_with not care about the specific value of the `lock` argument.

EvanKrall (approved):

PR #3434, opened 2022-08-10
Use apiextensions.k8s.io/v1 instead of v1beta1

TheNavigat:


jfongatyelp:
Resolved with https://github.com/Yelp/paasta/pull/3648 already

PR #3435, opened 2022-08-10
[FLINK-4097] Get Flink cluster status via JM API

marcos-sb:
This is a rework of #3418 that had to be reverted with #3429 because some automated sanity checks were failing in PROD environments.

This PR exposes a new set of endpoints to talk to the Flink JM API to pull Flink cluster status and Flink job-related information, instead of relying on the information stored in the Flink CRD.

This has been done as part of an initiative to remove the Flink Operator dependency on the Flink Job Manager, to be able to deploy Flink clusters on isolated environments.


paasta_tools/flink_tools.py
```diff
@@ -210,9 +211,14 @@ def _filter_for_endpoint(json_response: Any, endpoint: str) -> Mapping[str, Any]
 
 
 def _get_jm_rest_api_base_url(cr: Mapping[str, Any]) -> str:
```
marcos-sb, line 213: I'm happy to write a few test cases for this

marcos-sb:
Some of you have asked me out-of-band to explain how I tested this:
- Manually testing in all the environments where a Flink CRD is deployed if curling the REST API via the ingress works
  - URL template: `http://flink.k8s.{cluster}.paasta:{ingress-port}/{paasta-service-name}-{hashed-paasta-instance-name}`
- Outputting the generated URL to talk to the Job Manager on infrastage
- Wrote unit tests for the paasta client parts (mocked the API part)
- Manually tested the client fallback

paasta_tools/cli/cmds/status.py
```diff
@@ -1172,6 +1180,252 @@ def print_flink_status(
...
+    if not client:
+        output.append(
+            PaastaColors.red(
+                "paasta-api client unavailable - unable to get flink status"
+            )
+        )
+        return 1
+
+    api_version_minor = client.default.show_version().split(".")[1]
+    if api_version_minor <= "139":  # fallback
```
nemacysts, line 1408: will this need to change now?

paasta_tools/flink_tools.py
```diff
@@ -166,6 +187,64 @@ def _dashboard_get(cr_name: str, cluster: str, path: str) -> str:
...
+
+
+def _get_jm_rest_api_base_url(cr: Mapping[str, Any]) -> str:
+    metadata = cr["metadata"]
+    cluster = metadata["labels"][paasta_prefixed("cluster")]
+    base_url = get_flink_ingress_url_root(cluster)
+
+    # dashboard_url = http://flink-jobmanager-host:port/paasta-service-cr-name
+    dashboard_url = metadata["annotations"]["flink.yelp.com/dashboard_url"]
+    service_cr_name = dashboard_url[dashboard_url.rfind("/") + 1 :]
```
nemacysts, line 220: could this use https://docs.python.org/3.8/library/urllib.parse.html#urllib.parse.urlparse?

paasta_tools/flink_tools.py
```diff
@@ -166,6 +187,64 @@ def _dashboard_get(cr_name: str, cluster: str, path: str) -> str:
...
+
+    return base_url + service_cr_name
```
nemacysts, line 222: should this be a urljoin for safety?

paasta_tools/flink_tools.py
```diff
@@ -166,6 +187,64 @@ def _dashboard_get(cr_name: str, cluster: str, path: str) -> str:
...
+
+
+def curl_flink_endpoint(cr_id: Mapping[str, str], endpoint: str) -> Mapping[str, Any]:
+    try:
+        cr = get_cr(settings.kubernetes_client, cr_id)
+        if cr is None:
+            raise ValueError(f"failed to get CR for id: {cr_id}")
+        base_url = _get_jm_rest_api_base_url(cr)
+        url = f"{base_url}/{endpoint}"
```
nemacysts, line 231: same here: should this be a urljoin?

paasta_tools/flink_tools.py
```diff
@@ -166,6 +187,64 @@ def _dashboard_get(cr_name: str, cluster: str, path: str) -> str:
...
+        response = requests.get(url, timeout=FLINK_DASHBOARD_TIMEOUT_SECONDS)
+        if response.ok:
+            return _filter_for_endpoint(response.json(), endpoint)
+        return {}
```
nemacysts, line 235: will the error we presumably encountered if we hit this case be exposed in some other way?

nemacysts (commented):
looks generally fine to me, although it's hard to tell with the current diff what was pre-existing code and what was re-worked

paasta_tools/cli/cmds/status.py
marcos-sb: That's correct; I bumped it from `138`.
We're currently at `0.139.4`, and I intended to release a new `v0.140.0` with this changes, so any client talking to an API <= v139 should not expect the new endpoints to be in place.

paasta_tools/flink_tools.py
marcos-sb: yup, I'll rewrite it.

paasta_tools/flink_tools.py
marcos-sb: that's not a bad idea

paasta_tools/flink_tools.py
marcos-sb: I'm happy to change it too. We have test cases in place for this though, that would hopefully prevent any future breaking changes.

paasta_tools/flink_tools.py
marcos-sb: I agree it's a good idea to propagate any error client-side, not just an empty body. I'll look around and see what we do in other handlers to try to reproduce it.

paasta_tools/flink_tools.py
```diff
@@ -166,6 +189,71 @@ def _dashboard_get(cr_name: str, cluster: str, path: str) -> str:
...
+        }
+    return json_response
+
+
+def _get_jm_rest_api_base_url(cr: Mapping[str, Any]) -> str:
+    metadata = cr["metadata"]
+    cluster = metadata["labels"][paasta_prefixed("cluster")]
+    base_url = get_flink_ingress_url_root(cluster)
+
+    # dashboard_url = http://flink-jobmanager-host:port/paasta-service-cr-name
```
nemacysts, line 220: super-nit to prevent confusion from thinking this is debug code
```suggestion
    # this will look something like http://flink-jobmanager-host:port/paasta-service-cr-name
```

paasta_tools/flink_tools.py
```diff
@@ -166,6 +189,71 @@ def _dashboard_get(cr_name: str, cluster: str, path: str) -> str:
...
+    dashboard_url = urlparse(metadata["annotations"]["flink.yelp.com/dashboard_url"])
+    service_cr_name = dashboard_url.path
```
nemacysts, line 222: we could get fancy here and do something like:
```suggestion
    _, _, service_cr_name, *_ = urlparse(metadata["annotations"]["flink.yelp.com/dashboard_url"])
```

but that might be a little too clever to just save a single line so maybe ignore me here :p

paasta_tools/flink_tools.py
```diff
@@ -166,6 +189,71 @@ def _dashboard_get(cr_name: str, cluster: str, path: str) -> str:
...
+def curl_flink_endpoint(cr_id: Mapping[str, str], endpoint: str) -> Mapping[str, Any]:
+    try:
+        cr = get_cr(settings.kubernetes_client, cr_id)
+        if cr is None:
+            raise ValueError(f"failed to get CR for id: {cr_id}")
+        base_url = _get_jm_rest_api_base_url(cr)
+
+        # Closing 'base_url' with '/' to force urljoin to append 'endpoint' to the path.
+        # If not, urljoin replaces the 'base_url' path with 'endpoint'.
+        url = urljoin(base_url + "/", endpoint)
```
nemacysts, line 236: awww, i was misremembering how this worked ;_; - i could have sworn that urljoin would always append, but i guess not :p

(but at least this will make sure to dedupe `/` if base_url is ever updated to have a trailing `/` which is still nice)

nemacysts (approved):

paasta_tools/flink_tools.py
marcos-sb: aye!

paasta_tools/flink_tools.py
marcos-sb: I like fancy :P

paasta_tools/flink_tools.py
marcos-sb: > (but at least this will make sure to dedupe / if base_url is ever updated to have a trailing / which is still nice)

That's correct

marcos-sb:
@nemacysts Thanks for reviewing! I'll merge release this on Monday morning BST

PR #3436, opened 2022-08-11
Add validation to discourage large cpu_burst_add

nemacysts:
cpu_burst_add is added on to the `cpus` value (which we use for k8s
cpu requests to get a k8s cpu limit - and we've seen internal users
set excessively high values here (which can impact other services
running on the same node).

To discourage this without blocking legitimate usecases (/providing a
smooth transition to outright blocking this if possible in the future),
we allow users to bypass this validation by adding a special comment
(just like we do with the cpu-autoscaled-but-not-autotuned validation)

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -633,6 +633,12 @@ def validate_autoscaling_configs(service_path):
     return returncode
 
 
+def __is_templated(service: str, soa_dir: str, cluster: str, workload: str) -> bool:
+    return os.path.exists(
+        os.path.join(os.path.abspath(soa_dir), service, f"{workload}-{cluster}.in")
+    )
+
+
 def validate_min_max_instances(service_path):
     soa_dir, service = path_to_soa_dir_service(service_path)
     returncode = True
@@ -736,6 +742,10 @@ def validate_cpu_burst(service_path: str) -> bool:
 
     returncode = True
     for cluster in list_clusters(service, soa_dir):
+        if __is_templated(service, soa_dir, cluster, workload="kubernetes"):
+            # we should eventually make the python templates add the override comment
+            # to the correspoding YAML line, but until then we just opt these out of that validation
+            continue
         for instance in list_all_instances_for_service(
             service=service, clusters=[cluster], soa_dir=soa_dir
         ):
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -22,6 +22,7 @@ from functools import lru_cache
 from functools import partial
 from glob import glob
 from typing import Any
+from typing import Callable
 from typing import Dict
 from typing import List
 from typing import Optional
@@ -802,7 +803,7 @@ def paasta_validate_soa_configs(
     if not validate_service_name(service):
         return False
 
-    checks = [
+    checks: List[Callable[[str], bool]] = [
         validate_all_schemas,
         partial(validate_tron, verbose=verbose),
         validate_paasta_objects,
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -728,30 +802,18 @@ def paasta_validate_soa_configs(
...
+        partial(validate_tron, verbose=verbose),
+        validate_paasta_objects,
+        validate_unique_instance_names,
+        validate_autoscaling_configs,
+        validate_secrets,
+        validate_min_max_instances,
+        validate_cpu_burst,
+    ]
+
+    return all(check(service_path) for check in checks)
```
nemacysts, line 816: this is mostly me being cheeky - there's a bit of a behavior change here in that now we'll stop running checks as soon as one fails, but in a way that seems fine to me? e.g., if your files don't pass schema validation: we can't really assume that the rest of the checks will even have valid config files to look at

that said: i can undo this :p

paasta_tools/cli/cmds/validate.py
```diff
@@ -715,6 +727,68 @@ def validate_secrets(service_path):
...
+                    )
+                    is None
+                ):
+                    returncode = False
+                    print(
+                        failure(
+                            msg=f"Potentially excessive CPU burst (cpu_burst_add: {config[instance]['cpu_burst_add']} "
+                            f"higher than current threshold of {CPU_BURST_THRESHOLD} cores) detected in {cluster}: {service}.{instance}."
+                            " Please read the following link for next steps:",
+                            link="y/high-cpu-burst",
```
nemacysts, line 785: (still writing this)

paasta_tools/cli/cmds/validate.py
EvanKrall: I think it's fine to abort later checks if a failure in an early check would probably cause later checks to fail too, and the checks are sorted by priority/informativeness.

`validate_tron` seems pretty unrelated to `validate_paasta_objects` though.

You could restore the original, non-short-circuiting behavior by doing `all([...])` instead of `all(...)`.

paasta_tools/cli/cmds/validate.py
```diff
@@ -715,6 +728,68 @@ def validate_secrets(service_path):
...
+                        string=burst_comment,
+                    )
+                    is None
+                ):
+                    returncode = False
+                    print(
+                        failure(
+                            msg=f"Potentially excessive CPU burst (cpu_burst_add: {config[instance]['cpu_burst_add']} "
+                            f"higher than current threshold of {CPU_BURST_THRESHOLD} cores) detected in {cluster}: {service}.{instance}."
+                            " Please read the following link for next steps:",
```
EvanKrall, line 785: might be nice to print `OVERRIDE_CPU_BURST_ACK_PATTERN` in this message, so people who have seen this error before can quickly resolve it without reading docs

EvanKrall (approved):

paasta_tools/cli/cmds/validate.py
nemacysts: yea, I guess my train of thought here was that if you've seen this before and know what you're doing you'll be able to quickly scan the docs to get the right string OR you'll remember the previous occurrence of this and grab the string from there - imo, it's kinda nice to have a little bit of friction here to prevent folks from setting too high a value carelessly

paasta_tools/cli/cmds/validate.py
nemacysts: (ended up restoring the original behavior so that i didn't need to think about the optimal order here)

sidsatyelp (approved):

paasta_tools/cli/cmds/validate.py
```diff
@@ -218,6 +230,7 @@ def validate_service_name(service):
     return True
 
 
+@lru_cache()
```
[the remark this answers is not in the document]
nemacysts: we could - it doesn't really matter too much though: this code is only ever called in validate.py and validate.py only really serves `paasta validate` which runs for a single service and then exits.

PR #3440, opened 2022-08-16
[FLINK-4123] Remove the fallback for the paasta client

marcos-sb:
- The paasta client fallback handled the case where a paasta client was trying to consume an old paasta api (ie, without exposing the new endpoints the client was expecting). This case, produced 404s that canceled the request and ultimately broke some Sensu checks too

- This fallback is not required anymore as all environments are running a paasta api version >= 0.141.0


marcos-sb:
I'm not going to merge this until the new version of the flink operator had been successfully released first.

marcos-sb:
> wasn't there also a test for the fallback code?

Nope, there wasn't. I run it tests twice locally using comments to force a specific code path

marcos-sb:
> wasn't there also a test for the fallback code?

Yup, we use a hardcoded api version in one of the tests to force the non-fallback code path. I'll remove that too, thanks!

marcos-sb:
@nemacysts This is now ready. A quick review is appreciated but this is not high priority.

nemacysts (approved):
whoops - my bad, totally lost track of this pr!

PR #3442, opened 2022-08-19
[FLINK-4152] Use V1 apiextensions

marcos-sb:
Simple upgrade to apiextensions to v1.
This fixes our local testing environment for the flink operator: we were getting 400s (malformed request) while registering the CRD with local kind cluster.
~~I'm going to leave it as a draft until I have more clarity on the implications for other k8s services.~~

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -480,7 +480,7 @@ class KubeClient:
         self.deployments = kube_client.AppsV1Api(self.api_client)
         self.core = kube_client.CoreV1Api(self.api_client)
         self.policy = kube_client.PolicyV1beta1Api(self.api_client)
-        self.apiextensions = kube_client.ApiextensionsV1beta1Api(self.api_client)
+        self.apiextensions = kube_client.ApiextensionsV1Api(self.api_client)
         self.custom = kube_client.CustomObjectsApi(self.api_client)
         self.autoscaling = kube_client.AutoscalingV2beta2Api(self.api_client)
         self.rbac = kube_client.RbacAuthorizationV1Api(self.api_client)
--- a/paasta_tools/setup_kubernetes_internal_crd.py
+++ b/paasta_tools/setup_kubernetes_internal_crd.py
@@ -34,7 +34,7 @@ log = logging.getLogger(__name__)
 
 INTERNAL_CRDS = [
     V1beta1CustomResourceDefinition(
-        api_version="apiextensions.k8s.io/v1beta1",
+        api_version="apiextensions.k8s.io/v1",
         kind="CustomResourceDefinition",
         metadata={
             "name": "deploygroups.paasta.yelp.com",
@@ -44,7 +44,24 @@ INTERNAL_CRDS = [
         },
         spec={
             "group": "paasta.yelp.com",
-            "versions": [{"name": "v1beta1", "served": True, "storage": True}],
+            "versions": [
+                {
+                    "name": "v1beta1",
+                    "schema": {
+                        "openAPIV3Schema": {
+                            "type": "object",
+                            "properties": {
+                                "service": {"type": "string"},
+                                "deploy_group": {"type": "string"},
+                                "git_sha": {"type": "string"},
+                                "image_version": {"type": "string"},
+                            },
+                        },
+                    },
+                    "served": True,
+                    "storage": True,
+                },
+            ],
             "scope": "Namespaced",
             "names": {
                 "plural": "deploygroups",
@@ -52,21 +69,10 @@ INTERNAL_CRDS = [
                 "kind": "DeployGroup",
                 "shortNames": ["dg"],
             },
-            "validation": {
-                "openAPIV3Schema": {
-                    "type": "object",
-                    "properties": {
-                        "service": {"type": "string"},
-                        "deploy_group": {"type": "string"},
-                        "git_sha": {"type": "string"},
-                        "image_version": {"type": "string"},
-                    },
-                }
-            },
         },
     ),
     V1beta1CustomResourceDefinition(
-        api_version="apiextensions.k8s.io/v1beta1",
+        api_version="apiextensions.k8s.io/v1",
         kind="CustomResourceDefinition",
         metadata={
             "name": "startstopcontrols.paasta.yelp.com",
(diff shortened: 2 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -480,7 +480,7 @@ def __init__(self, component: Optional[str] = None) -> None:
         self.deployments = kube_client.AppsV1Api(self.api_client)
         self.core = kube_client.CoreV1Api(self.api_client)
         self.policy = kube_client.PolicyV1beta1Api(self.api_client)
-        self.apiextensions = kube_client.ApiextensionsV1beta1Api(self.api_client)
+        self.apiextensions = kube_client.ApiextensionsV1Api(self.api_client)
```
nemacysts, line 483: will this impact any code referencing the v1beta1 version? e.g., https://sourcegraph.yelpcorp.com/Yelp/paasta/-/blob/paasta_tools/setup_kubernetes_internal_crd.py?subtree=true#L37:22

marcos-sb: That's the intended 'secondary effect' of this patch. I have tagged some of you because there have been some issues with the latest k8s upgrade, and this might help or it might make things worse (re: COMPINFRA-1370).
For us, the upgrade broke the local flink operator cluster (specifically the part where we register the CRD).

marcos-sb:
hi @EvanKrall, can I get a review pls? :)

paasta_tools/kubernetes_tools.py
EvanKrall: Looks like we might be able to set `api_version="apiextensions.k8s.io/v1"` in setup_kubernetes_internal_crd.py. K8s v1.19 has both v1 and v1beta1 for apiextensions: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#-strong-api-groups-strong-

However, the schema for `validations` has been dropped since [v1beta1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#customresourcedefinitionspec-v1beta1-apiextensions-k8s-io) and [v1](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#customresourcedefinitionspec-v1-apiextensions-k8s-io) so it looks like we'll need to move the validation schema from `validations` to a `schema` in the `versions` entry: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#customresourcedefinitionversion-v1-apiextensions-k8s-io

marcos-sb: Uhm... the apiextensions.v1beta1 client doesn't seem very happy about handling requests for a v1 crd.

After editing `INTERNAL_CRDS` as you suggested, I got
```
DEBUG:paasta_tools.kubernetes_tools:{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"the API version in the data (apiextensions.k8s.io/v1) does not match the expected API version (apiextensions.k8s.io/v1beta1)","reason":"BadRequest","code":400}
```

I've also tried changing the CRD object version from V1beta1CustomResourceDefinition to V1CustomResourceDefinition, but it didn't make any difference.

Using `ApiextensionsV1Api` as `apiextensions` seems to register correctly `v1beta1` and `v1` versions of the CRD.

marcos-sb: Any pointers as to how to test that this change is safe?

EvanKrall: > After editing INTERNAL_CRDS as you suggested, I got
> ```
> DEBUG:paasta_tools.kubernetes_tools:{"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"the API version in the data (apiextensions.k8s.io/v1) does not match the expected API version (apiextensions.k8s.io/v1beta1)","reason":"BadRequest","code":400}```

This was with the `ApiextensionsV1Api` right?

> Using ApiextensionsV1Api as apiextensions seems to register correctly v1beta1 and v1 versions of the CRD.

If ApiextensionsV1Api is happy to submit v1beta1 CRDs, then I think you should be able to ship this branch and CI can update the internal CRDs later.

marcos-sb: > If ApiextensionsV1Api is happy to submit v1beta1 CRDs, then I think you should be able to ship this branch and CI can update the internal CRDs later.

The caveat here, I'm sorry I wasn't any clearer, is that we'd have to adapt the `spec` in the CRD object (even if we use the V1beta1 version of the object) to conform to the V1 API schema for requests to succeed.

```
--- a/paasta_tools/setup_kubernetes_internal_crd.py
+++ b/paasta_tools/setup_kubernetes_internal_crd.py
@@ -34,7 +34,7 @@ log = logging.getLogger(__name__)

 INTERNAL_CRDS = [
     V1beta1CustomResourceDefinition(
-        api_version="apiextensions.k8s.io/v1beta1",
+        api_version="apiextensions.k8s.io/v1",
         kind="CustomResourceDefinition",
         metadata={
             "name": "deploygroups.paasta.yelp.com",
@@ -44,7 +44,24 @@ INTERNAL_CRDS = [
         },
         spec={
             "group": "paasta.yelp.com",
-            "versions": [{"name": "v1beta1", "served": True, "storage": True}],
+            "versions": [
+                {
+                    "name": "v1",
+                    "schema": {
+                        "openAPIV3Schema": {
+                            "type": "object",
+                            "properties": {
+                                "service": {"type": "string"},
+                                "deploy_group": {"type": "string"},
+                                "git_sha": {"type": "string"},
+                                "image_version": {"type": "string"},
+                        },
+                    },
+                    "served": True,
+                    "storage": True,
+                },
+            ],
[...]

```

EvanKrall: That looks good, I'd 🚢 that diff. I think the second `v1beta1` (in spec.versions[0].name) is the version number of the CRD itself rather than the API, so that should remain at v1beta1.

It's a bummer that the v1 client doesn't support v1beta1 definitions, but I guess they expect that you'd just do this sort of refactor when you bump the client version. (Or instantiate a v1beta1 client in places where you want to use the v1beta1 api.)

marcos-sb: > That looks good, I'd 🚢 that diff. I think the second v1beta1 (in spec.versions[0].name) is the version number of the CRD itself rather than the API, so that should remain at v1beta1.

Looks that way, yeah. For completeness, I changed that value to `v1` (keeping the rest of the diff the same) and got 422s instead.

I'll push the changes you suggested, plus to the other 2 files we use to setup the CRs.

PR #3443, opened 2022-08-22
removed mesos_cpu from kubernetes_schema.json and other references

sidsatyelp:
[PAASTA-17210](https://jira.yelpcorp.com/browse/PAASTA-17210)

Rename `metrics_provider: mesos_cpu` to just `cpu`

This is part 2 of a 2 part change.
Part 1 was to rename `mesos_cpu` to just `cpu` in sysgit/yelpsoa-config ([link to PR](https://github.yelpcorp.com/sysgit/yelpsoa-configs/pull/20980)). Will need to merge that PR before this one. 

For now I have left both of these as a draft PR to get feedback from the team and additional tips on how to test this end to end..

How did I test this so far?

1. `make test` completes successfully.
2. `paasta validate` [completes successfully](https://fluffy.yelpcorp.com/i/hW10lN1D471vWwkdbZPCLLZmq39h6kQV.html)


paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -69,7 +69,7 @@
 DEFAULT_EXTRAPOLATION_PERIODS = 10
 DEFAULT_EXTRAPOLATION_TIME = DEFAULT_SCRAPE_PERIOD_S * DEFAULT_EXTRAPOLATION_PERIODS
 
-CPU_METRICS_PROVIDERS = ("mesos_cpu", "cpu")
+CPU_METRICS_PROVIDERS = "cpu"
```
nemacysts, line 72: this should remain a collection or the code that uses this should be refactored to use equality (rather than `in`)

nemacysts: (i'd probably switch to equality and either drop this constant or un-plural the constant name)

tests/test_setup_prometheus_adapter_config.py
```diff
@@ -21,15 +21,6 @@
...
     [
-        (
-            {
-                "metrics_provider": "mesos_cpu",
-                "decision_policy": "bespoke",
-                "moving_average_window_seconds": 123,
-                "setpoint": 0.653,
-            },
-            False,
-        ),
```
jfongatyelp, line 32: I see for the next test cases you're deleting, there are equivalents using `cpu` instead of `mesos_cpu`, but I don't see one here.  Is this alright that we no longer have coverage for cpu-autoscaling with this set of options?

jfongatyelp (approved):

tests/test_setup_prometheus_adapter_config.py
sidsatyelp: Good catch @jfongatyelp ! I have added the missing test case.

jfongatyelp (approved):

PR #3445, opened 2022-08-24
Evicted pods should cause replica state to be 'Not Running'

sidsatyelp:
Added a condition to set Evicted pod's replica state to be Not running.
All tests are passing.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1673,6 +1673,7 @@ def get_main_container(pod: KubernetesPodV2) -> Optional[KubernetesContainerV2]:
 def get_replica_state(pod: KubernetesPodV2) -> ReplicaState:
     phase = pod.phase
     state = ReplicaState.UNKNOWN
+    reason = pod.reason
     if phase is None or not pod.scheduled:
         state = ReplicaState.UNSCHEDULED
     elif pod.delete_timestamp:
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -1715,7 +1716,7 @@ def get_replica_state(pod: KubernetesPodV2) -> ReplicaState:
             state = ReplicaState.UNKNOWN
 
         # Catch
-    elif phase == "Failed":
+    elif phase == "Failed" or reason == "Evicted":
```
nemacysts, line 1719: hmm, interesting - evicted pods aren't in the `Failed` phase?

paasta_tools/cli/cmds/status.py
```diff
@@ -1715,7 +1716,7 @@ def get_replica_state(pod: KubernetesPodV2) -> ReplicaState:
...
         # e.g. pod.reason == evicted
```
nemacysts, line 1720: might be worth cleaning up these comments too - the current formatting is a bit confusing (and after adding this new conditional, i don't think they really add anything)

paasta_tools/cli/cmds/status.py
nemacysts: oh, I think I see what's happening based on the internal ticket: we're hitting https://github.com/Yelp/paasta/pull/3445/files#diff-36175af498b294330c337132851420b6d63c7469e72162fb6a4496fcd5cbaff2R1677, so I don't think that this change will actually fix the issue since we'll never get to this branch

sidsatyelp:
To test this I ran the below command to find pods that are currently evicted and ran the corresponding paasta status to ensure that the evicted pods are now showing as `Not Running` instead of `Unscheduled`

`kubectl-pnw-prod get pods -n paasta | grep Evicted`
https://fluffy.yelpcorp.com/i/lX4t82cTnGTXwKNqn9S88qL4K9GJzjmc.html

EvanKrall (approved):

PR #3446, opened 2022-08-24
Adds specifying vault token path PAASTA-17437

bobokun:
This PR will allow vault token path to be passed as a parameter in sync_secrets and default to /root/.vault_token if not defined to prevent any previous functions from breaking

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -43,6 +43,7 @@ from paasta_tools.kubernetes_tools import update_secret
 from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.secret_tools import get_secret_provider
 from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import DEFAULT_VAULT_TOKEN_FILE
 from paasta_tools.utils import get_service_instance_list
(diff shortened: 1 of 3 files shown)


paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -81,6 +82,13 @@ def parse_args() -> argparse.Namespace:
         dest="namespace",
         help="Overwrite destination namespace for secrets",
     )
+    parser.add_argument(
+        "-vtf",
```
nemacysts, line 86: (personal opinion): multi-char short options always look a little weird to me (they usually make me think of linux cli tools where you can chain multiple short options using 1 `-`) - thoughts on having the short option here be `-t` instead?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -196,14 +207,15 @@ def sync_secrets(
     vault_cluster_config: Mapping[str, str],
     soa_dir: str,
     namespace: str,
+    vault_token_file: str,
```
nemacysts, line 210: thoughts on having this look something like:
```suggestion
    vault_token_file: str = DEFAULT_VAULT_TOKEN_FILE,
```
this way, no tests need to be updated and the same defaults from using the cli version of this apply to anyone just importing `sync_secrets`

nemacysts (approved):
i left some comments, but they're non-blocking imo

nemacysts (approved):

PR #3450, opened 2022-08-31
Allow setting load balancer weight per instance in yelpsoa-configs

EvanKrall:
For PEREL-3176, this should let me run two sets of pods simultaneously in the same nerve namespace, with different sizes (cpus, number of workers) but also different weights.

This should let us do side-by-side comparisons of different sizes of containers, while keeping their relative traffic levels the same. (e.g. bigger pods would get more traffic, so that percent utilization is the same for both sizes.)

Corresponding configure_nerve PR: https://github.com/Yelp/nerve-tools/pull/96

### DIFF
--- a/paasta_tools/dump_locally_running_services.py
+++ b/paasta_tools/dump_locally_running_services.py
@@ -25,14 +25,16 @@ Command line options:
 import argparse
 import json
 import sys
+from typing import List
+from typing import Tuple
 
 from paasta_tools.kubernetes_tools import get_kubernetes_services_running_here_for_nerve
-from paasta_tools.marathon_tools import get_marathon_services_running_here_for_nerve
+from paasta_tools.long_running_service_tools import ServiceNamespaceConfig
 from paasta_tools.marathon_tools import get_puppet_services_running_here_for_nerve
 from paasta_tools.utils import DEFAULT_SOA_DIR
 
 
-def parse_args(argv):
+def parse_args(argv) -> argparse.Namespace:
     parser = argparse.ArgumentParser(
         description="Dumps information about locally running services."
     )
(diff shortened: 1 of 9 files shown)


paasta_tools/dump_locally_running_services.py
```diff
@@ -25,14 +25,16 @@
...
 
 from paasta_tools.kubernetes_tools import get_kubernetes_services_running_here_for_nerve
-from paasta_tools.marathon_tools import get_marathon_services_running_here_for_nerve
+from paasta_tools.long_running_service_tools import ServiceNamespaceConfig
 from paasta_tools.marathon_tools import get_puppet_services_running_here_for_nerve
 from paasta_tools.utils import DEFAULT_SOA_DIR
 
 
-def parse_args(argv):
+def parse_args(argv) -> argparse.Namespace:
```
nemacysts, line 37:
```suggestion
def parse_args(argv: Optional[Sequence[str]]) -> argparse.Namespace:
```

if you'd like to be extra thorough :p

nemacysts: (would be the same typing for `main()` too)

tests/test_kubernetes_tools.py
```diff
@@ -3682,7 +3693,7 @@ def test_warning_big_bounce_routable_pod():
             job_config.format_kubernetes_app().spec.template.metadata.labels[
                 "paasta.yelp.com/config_sha"
             ]
-            == "configf46d563a"
+            == "config1404b38f"
```
nemacysts, line 3696: fun times ahead :p

nemacysts (approved):

PR #3453, opened 2022-08-31
Have Github Actions create a Github release for release tags.

EvanKrall:
This should unblock me fixing the build in https://github.com/Yelp/nerve-tools/pull/96 by making paasta-tools debs available for the nerve-tools itests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,9 +77,8 @@ jobs:
           path: dist/
       - name: Release
         uses: softprops/action-gh-release@v1
-        # if: startsWith(github.ref, 'refs/tags/v')
+        if: startsWith(github.ref, 'refs/tags/v')
         with:
-          draft: true  # remove me, just testing
           files: |
             dist/paasta-tools_*.deb
           fail_on_unmatched_files: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,7 +66,7 @@ jobs:
     needs: build_debs
     steps:
       - uses: actions/checkout@v2
-      - run: mkdir -p dist/bionic dist/jammy
+      - run: mkdir -p dist/
       - uses: actions/download-artifact@v3
         with:
           name: deb-bionic
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,6 @@ jobs:
       - run: sudo apt-get install -yq devscripts
       - run: make itest_${{ matrix.dist }}
       - uses: actions/upload-artifact@v3
-
         with:
           name: deb-${{ matrix.dist }}
           path: dist/paasta-tools_*.deb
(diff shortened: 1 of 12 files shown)


EvanKrall:
https://github.com/Yelp/paasta/runs/8125620397?check_suite_focus=true shows this creating a release with the two debs; you can see the draft release at https://github.com/Yelp/paasta/releases/tag/untagged-c4156b5f5c0411bcbf70 (you may need to be logged in + a Paasta contributor to see this)

In 34803a4 I've made it only cut releases on tags, and not make them draft releases.

.github/workflows/ci.yml
```diff
@@ -56,3 +56,29 @@ jobs:
       - run: sudo apt-get update
       - run: sudo apt-get install -yq devscripts
       - run: make itest_${{ matrix.dist }}
+      - uses: actions/upload-artifact@v3
+
```
nemacysts, line 60: super-nit
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -56,3 +56,29 @@ jobs:
...
+        with:
+          name: deb-${{ matrix.dist }}
+          path: dist/paasta-tools_*.deb
+  cut_release:
+    runs-on: ubuntu-22.04
+    needs: build_debs
+    steps:
+      - uses: actions/checkout@v2
+      - run: mkdir -p dist/bionic dist/jammy
```
nemacysts, line 69: (non-blocking) hmm, i wonder if we can use a matrix for doing the ubuntu version stuff so that there's just a single place we need to update when we're supporting/dropping a given ubuntu version

yelp_package/dockerfiles/bionic/Dockerfile
```diff
@@ -18,7 +18,7 @@ FROM ${DOCKER_REGISTRY}ubuntu:bionic
 ARG PIP_INDEX_URL=https://pypi.yelpcorp.com/simple
 ENV PIP_INDEX_URL=$PIP_INDEX_URL
 
-RUN sed -i 's/archive.ubuntu.com/us-east1.gce.archive.ubuntu.com/g' /etc/apt/sources.list
+# RUN sed -i 's/archive.ubuntu.com/us-east1.gce.archive.ubuntu.com/g' /etc/apt/sources.list
```
nemacysts, line 21: nit: should we just delete this line and the commented mesosphere stuff in all the dockerfiles?

nemacysts (approved):

.github/workflows/ci.yml
EvanKrall: I don't know of a way to do this with softprops/action-gh-release (which uploads all release files during that one step). If I had a matrix for release, I'd end up with two release jobs, each trying to make their own release.

In a personal repo, I've used [actions/upload-release-artifact](https://github.com/actions/upload-release-asset) and [actions/create-release](https://github.com/actions/create-release) to create the release first, then upload the artifacts directly from the build steps, but create-release and upload-release-artifact are deprecated; their readmes point to softprops/action-gh-release.

PR #3459, opened 2022-09-07
Kkasp/compinfra 1367 metric state transitions

KaspariK:
Add transitions for metric rollbacks as part of COMPINFRA-1367.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -911,6 +911,18 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
                 "trigger": "rollback_slo_failure",
             }
             yield {
+                "source": self.rollforward_states,
+                "dest": "start_rollback",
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -910,6 +910,18 @@ def valid_transitions(self) -> Iterator[state_machine.TransitionDefinition]:
...
                 "trigger": "rollback_slo_failure",
             }
+            yield {
+                "source": self.rollforward_states,
+                "dest": "start_rollback",
+                "trigger": "rollback_metric_failure",
+                "before": self.log_metric_rollback,
+            }
+            # TODO COMPINFRA-1140 - Is this transition necessary? Internal transition with no before/after/prepare does...nothing?
+            yield {
```
KaspariK, line 920: Might be worth leaving this out for now and only adding if we notice missing functionality. I feel like the SLO one was added incorrectly, based on my understanding of Transitions

PR #3468, opened 2022-09-22
Have local PaaSTA API running on the local Kind Cluster - COMPINFRA-1467

EmanElsaban:
### Purpose of this PR

This PR introduces two make targets: ``generate_etc_paasta_playground`` and ``playground-api``. The ``playground-api`` creates the ``etc_paasta_playground`` directory and runs the paasta api locally. The ``generate_etc_paasta_playground``  make target allows the user to only create the etc_paasta_playground directory and set it up with the cluster specified. 

 In this PR, I also added ``Run setup k8s job`` , ``Generate deployments.json`` ,``paasta status playground`` and ``paasta API playground`` configurations. 

- ``Run setup k8s job``  :  used  to deploy the ``compute-infra-test-service`` on the kind cluster. 
- ``Generate deployments.json`` : used to generate the deployment.json file for ``compute-infra-test-service`` 
- ``paasta status playground`` :  used to run paasta status to connect to the paasta api locally.
- ``paasta API playground``  :  used to run the paasta api locally

### Testing

In testing, I tried the following steps to run paasta playground locally:

1. To generate the launch.json file with all the configurations mentioned, run ``make vscode_settings``
2. Create a Kind Cluster 
3. Run the make target ``generate_etc_paasta_playground`` to generate the ``etc_paasta_playground`` directory.
4. From `` Run and Debug`` section in VS Code, press on ``Generate deployments.json`` to generate deployments.json file if it doesn't exist.
5. From `` Run and Debug`` section in VS Code, press on ``Run setup k8s job`` to deploy ``compute-infra-test-service``  in the cluster if it's not deployed.
6. Finally, you can either run paasta api locally using the make target or from `` Run and Debug`` section in VS Code. To run  it using the make target do ``make playground-api`` and then run ``paasta status playground`` from ` Run and Debug`` section in VS Code to test the connection.


tox.ini
```diff
@@ -37,6 +37,22 @@ deps =
 commands =
     python -m paasta_tools.run-paasta-api-in-dev-mode
 
+[testenv:playground-api]
+envdir = .tox/py37-linux/
+passenv = PAASTA_TEST_CLUSTER KUBECONFIG PAASTA_SYSTEM_CONFIG_DIR PAASTA_API_SOA_DIR
+setenv =
+    KUBECONFIG = ./k8s_itests/kubeconfig
+    PAASTA_TEST_CLUSTER = kind-emanelsabban-k8s-test
```
bobokun, line 45: Maybe we should rename the test cluster to something more generic before merging this?

k8s_itests/deployments/paasta/fake_soa_config/compute-infra-test-service/deployments.json
```diff
@@ -1,20 +1,24 @@
 {
     "v1": {
-        "compute-infra-test-service:<%cluster%>.autoscaling": {
-            "docker_image": "services-compute-infra-test-service:paasta-a0ec9b362e29123022e5cebb6612aedc58a0889b",
+        "compute-infra-test-service:paasta-kind-emanelsabban-k8s-test.autoscaling": {
```
bobokun, line 3: should this be hardcoded?

tox.ini
bobokun, line 45: Maybe we should rename the test cluster to something more generic before merging this?

k8s_itests/deployments/paasta/fake_soa_config/compute-infra-test-service/deployments.json
EmanElsaban: Thanks! Will change it. This file is also automatically generated when running ``generate_deployments_for_service.py``. So, if I changed  it to  <%cluster%>, the user can replace it manually with their cluster name or run "Generate deployments.json" in launch.json to generate this file  - ``deployments.json`` - automatically.

paasta_tools/run-paasta-api-playground.py
```diff
@@ -0,0 +1,48 @@
...
+    print("-------------------------------------------------------")
+    cluster = os.environ.get("PAASTA_TEST_CLUSTER", "kind-emanelsabban-k8s-test")
+    config_path = "etc_paasta_playground"
+
+    # find unused ports
+    port = pick_random_port("paasta-dev-api")
+
+    # Generate api endpoints
+    api_endpoints = {"api_endpoints": {cluster: f"http://localhost:{port}"}}
+    api_endpoints_path = os.path.join(os.getcwd(), config_path, "api_endpoints.json")
```
nemacysts, line 24: might be good to also move away from getcwd here and use `__FILE__` or something so that this doesn't do weird things should someone call it with a different working dir than expected

paasta_tools/run-paasta-api-playground.py
```diff
@@ -0,0 +1,48 @@
...
+    os.chmod(api_endpoints_path, 0o777)
+    with open(api_endpoints_path, "w") as f:
+        json.dump(api_endpoints, f)
+
+    # export config path
+    os.environ["PAASTA_SYSTEM_CONFIG_DIR"] = config_path
+    os.execl(
+        ".tox/py37-linux/bin/python",
+        ".tox/py37-linux/bin/python",
```
nemacysts, line 33: is this meant to be duplicated?

paasta_tools/run-paasta-api-playground.py
```diff
@@ -0,0 +1,48 @@
...
+        "-m",
+        "paasta_tools.api.api",
+        *[
+            "-D",
+            "-c",
+            cluster,
+            "-d",
```
nemacysts, line 40: if we're going to exec for this, we should probably use the --long-form of these args for future us's sake

paasta_tools/contrib/ide_helper.py
```diff
@@ -144,6 +163,67 @@ def install_vscode_support() -> None:
...
+                    "KUBECONFIG": "./k8s_itests/kubeconfig",
+                    "PAASTA_SYSTEM_CONFIG_DIR": "./etc_paasta_playground/",
+                    "PAASTA_TEST_CLUSTER": "kind-emanelsabban-k8s-test",
+                    "PYDEVD_USE_CYTHON": "NO",
+                    "PYTHONUNBUFFERED": "1",
+                    "PAASTA_API_SOA_DIR": "./k8s_itests/deployments/paasta/fake_soa_config",
+                },
+            },
+            {
+                "name": "Run setup k8s job",
```
nemacysts, line 184: suggestion: might be nice to add `in playground` or something to the names of these

paasta_tools/contrib/create_etc_paasta_playground.py
```diff
@@ -0,0 +1,58 @@
...
+
+    # Add in volumes.json
+    hacheck_sidecar_volumes = {"volumes": [], "hacheck_sidecar_volumes": []}
+    volumes_path = os.path.join(os.getcwd(), config_path, "volumes.json")
+    os.chmod(volumes_path, 0o777)
+    with open(volumes_path, "w") as f:
+        json.dump(hacheck_sidecar_volumes, f)
+
+    # rename kubernetes-<%cluster%>.yaml with your cluster name
+    # check if the cluster is not already renamed
```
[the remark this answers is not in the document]
EmanElsaban: yep, will render files in this directory using ``render_template.py`` to ``soa_config_playground`` directory to preserve the templates and automate the process of naming this file

paasta_tools/run-paasta-api-playground.py
EmanElsaban: yep, I tried it before without having both and it won't work. It's also done like that in ``run-paasta-api-in-dev-mode.py`` file. This function takes os.execl(path, args) (path of the file to execute and then arguments)

paasta_tools/contrib/create_etc_paasta_playground.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def main():
+    cluster = os.environ.get("PAASTA_TEST_CLUSTER", "kind-emanelsabban-k8s-test")
+    config_path = "etc_paasta_playground"
+
+    # create an etc_paasta_playground directory if it doesn't exist
+    # and copy fake_etc_paasta content into etc_paasta_playground directory
+    if not os.path.isdir("etc_paasta_playground"):
+        os.mkdir(os.path.join(os.getcwd(), "etc_paasta_playground"))
```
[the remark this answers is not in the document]
EmanElsaban: it's meant to be the root of the repo. So, first I'm checking if this repo exists and if not then I create it and render the files in ``fake_etc_paasta``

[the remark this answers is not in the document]
EmanElsaban: I think we might not even need it since its root. So, it would be ``os.mkdir("etc_paasta_playground")``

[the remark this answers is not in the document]
nemacysts: will that work? afaik, `os.mkdir` requires an absolute path - not a relative one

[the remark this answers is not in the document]
EmanElsaban: yep, I thought the same too but it worked :p

paasta_tools/contrib/ide_helper.py
```diff
@@ -18,7 +18,19 @@ def merge_vscode_settings(file: str, setting_dict: Dict[str, Any]) -> None:
...
+                for i, dict_entry in enumerate(setting_dict[key]):
+                    if dict_entry not in settings[key]:
+                        if isinstance(settings[key], list):
+                            settings[key].append(dict_entry)
+                        else:
+                            temp_list = [settings[key]]
+                            temp_list.append(dict_entry)
+                            settings[key] = temp_list
+        else:
+            settings[key] = value
```
nemacysts, line 33: not a blocker, but it'd be nice to make this a `merge_settings` or `merge_dicts` function or something :p

paasta_tools/contrib/create_paasta_playground.py
```diff
@@ -0,0 +1,57 @@
+import json
+import os
+
+from render_template import render_values
+
+
+def main():
+    cluster = os.environ.get(
+        "PAASTA_TEST_CLUSTER", "kind-{USER}-k8s-test".format(**os.environ)
```
nemacysts, line 9: suggestion:
```suggestion
        "PAASTA_TEST_CLUSTER", "kind-{USER}-k8s-test".format(os.environ['USER'])
```
might be a clearer way of doing this

paasta_tools/contrib/create_paasta_playground.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+    config_path = "etc_paasta_playground"
+
+    # create an etc_paasta_playground directory if it doesn't exist
+    # and copy fake_etc_paasta content into etc_paasta_playground directory
+    if not os.path.isdir("etc_paasta_playground"):
+        os.mkdir("etc_paasta_playground")
+
+    render_values(
+        "./k8s_itests/deployments/paasta/fake_etc_paasta", "etc_paasta_playground", None
```
nemacysts, line 19: nit: we define `config_path` above, but don't use it here even though we're referring the same thing

paasta_tools/contrib/create_paasta_playground.py
```diff
@@ -0,0 +1,57 @@
...
+    volumes_path = os.path.join(config_path, "volumes.json")
+    with open(volumes_path, "w") as f:
+        json.dump(hacheck_sidecar_volumes, f)
+
+    # create a fake_soa_config directory if it doesn't exist
+    # and copy fake_soa_config content into soa_config_playground directory
+    if not os.path.isdir("soa_config_playground"):
+        os.mkdir("soa_config_playground")
+    render_values(
+        "./k8s_itests/deployments/paasta/fake_soa_config", "soa_config_playground", None
```
nemacysts, line 39: suggestion: we could use kwargs here to make this a little more self-documenting: 
```suggestion
        src="./k8s_itests/deployments/paasta/fake_soa_config", dst="soa_config_playground", values=None
```
since this function's name doesn't really give the reader a hint as to what these strings are for

nemacysts: that said, it would be nice to switch to an actual templating library like jinja2 or mako in the future rather than carrying this function forward :p

paasta_tools/contrib/create_paasta_playground.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    # Code below will rename kubernetes-<%cluster%>.yaml with your cluster name
+    # if the cluster is not already renamed
```
nemacysts, line 43: huh, should this be happening?

paasta_tools/contrib/ide_helper.py
```diff
@@ -99,6 +111,45 @@ def install_vscode_support() -> None:
                     "canary",
                 ],
             },
+            {
```
nemacysts, line 114: not necessary to do now: given how fast/quickly this is growing it might be nice to move these dicts to separate json files rather than inlining the data here

paasta_tools/contrib/ide_helper.py
```diff
@@ -156,6 +269,40 @@ def install_vscode_support() -> None:
         ]
     }
     merge_vscode_settings("extensions.json", recommended_extensions)
+
+    playground_task = {
```
nemacysts, line 273: just out of curiosity as a non-vscode user: what's the difference between this task thing and the other stuff above for running paasta stuff (e.g., the dict that has the `Generate deployments.json in playground` entry in it)?

paasta_tools/contrib/create_paasta_playground.py
nemacysts: also: we don't need to do this now but it'd be nice to separate the data we use for integration tests from the data we use to spin up a playground at some point

nemacysts (approved):

paasta_tools/contrib/ide_helper.py
EmanElsaban: The task will allow you to run right away the whole paasta playground if the user just cares about running it, however it won't allow us to debug it.  The configurations in launch.json including ``Generate deployments.json in playground`` will allow the user to debug. For example, lets say I want to debug the api and the client, then I will run the configuration for ``paasta API playground`` and then run the configuration for ``paasta status in playground``.  The tasks.json and launch.json are both VS Code features. This is why we have the make targets for running the API playground and generating the etc_paasta_playground ( and soa_config_playground).

paasta_tools/contrib/create_paasta_playground.py
EmanElsaban: Thanks for pointing it out! I can actually pass values.yaml file to the render_values function and then I won't need to do this nor 
``
    # Generate Cluster name
     clusters_names = {"clusters": [cluster], "cluster": cluster}
     clusters_path = os.path.join(config_path, "clusters.json")
     with open(clusters_path, "w") as f:
         json.dump(clusters_names, f)
``

paasta_tools/contrib/ide_helper.py
```diff
@@ -18,12 +18,28 @@ def merge_vscode_settings(file: str, setting_dict: Dict[str, Any]) -> None:
...
 
     # update settings with the default configurations we want each file to have
-    settings.update(setting_dict)
+    merge_dicts(setting_dict, settings)
 
     with open(json_path, mode="w") as settings_file:
         json.dump(settings, settings_file, indent=2)
 
 
+def merge_dicts(setting_dict: Dict[str, Any], settings: Dict[str, Any]) -> None:
```
nemacysts, line 27: super-nit: thoughts on this change?
```suggestion
def merge_dicts(merge_from: Dict[str, Any], merge_to: Dict[str, Any]) -> None:
```
would make it a little clearer as to how the merge is meant to work

nemacysts (approved):

paasta_tools/contrib/ide_helper.py
EmanElsaban: makes sense, thanks!

nemacysts (approved):

PR #3471, opened 2022-09-28
Adhoc spark runs use environment credentials provider.

jsleight:
Swap over to environment credentials provider from the temporary credentials provider.

This also allows us to drop several of the other options I had added which are no longer needed.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -377,23 +377,6 @@ def add_subparser(subparsers):
     )
 
     aws_group.add_argument(
-        "--disable-aws-credential-env-variables",
-        help="Do not put AWS credentials into environment variables.  Credentials "
-        "will still be read and set in the Spark configuration.  In Spark v3.2 "
-        "you want to set this argument.",
-        action="store_true",
-        default=False,
-    )
-
-    aws_group.add_argument(
-        "--disable-temporary-credentials-provider",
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -560,13 +543,12 @@ def get_spark_env(
...
-            spark_env["AWS_ACCESS_KEY_ID"] = access_key
-            spark_env["AWS_SECRET_ACCESS_KEY"] = secret_key
-            if session_token is not None:
-                spark_env["AWS_SESSION_TOKEN"] = session_token
+    access_key, secret_key, session_token = aws_creds
+    if access_key:
+        spark_env["AWS_ACCESS_KEY_ID"] = access_key
+        spark_env["AWS_SECRET_ACCESS_KEY"] = secret_key
+        if session_token is not None:
+            spark_env["AWS_SESSION_TOKEN"] = session_token
```
nemacysts, line 551: i'm looking forward to the drivers on k8s so that we can stop leaking these in the kubectl output for these pods :p

nemacysts (approved):

jsleight:
Noticed we were logging creds in the final command (which is happening in master too), so took that out.  Have also now done manual tests for:
- a spark 3.0 job using okta creds
- a spark 3.0 job using service creds
- a spark 3.2 job using okta creds
- a spark 3.2 job using service creds
and everything works 🎉 .  So I think this finally solves all the credential provider issues for good!!

PR #3472, opened 2022-09-30
Correctly display scribe selector

nemacysts:
Using -e will not always work (e.g., if you're calling `paasta logs` on something running in prod), so let's add some code to look at what the log_reader cluster_map says the scribe region/ecosystem is and correctly pick between -e and -r

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -637,6 +637,15 @@ class ScribeLogReader(LogReader):
             )
         self.cluster_map = cluster_map
 
+    def get_scribereader_selector(self, scribe_env: str) -> str:
+        # this is kinda silly, but until the scribereader cli becomes more ergonomic
+        # we'll need to do a little bit of string munging to let humans use scribereader
+        # in the same way we are (tl;dr: scribereader has sorta confusing behavior between
+        # what can be use for --ecosystem, --region, and --superregion and the fastest/least
+        # hacky thing to figure out which we wanna use is that any env with a - in it is a region
+        # and any without one is an ecosystem)
+        return "-e" if "-" in scribe_env else "-r"
+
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/logs.py
```diff
@@ -637,6 +637,15 @@ def __init__(self, cluster_map: Mapping[str, Any]) -> None:
...
         self.cluster_map = cluster_map
 
+    def get_scribereader_selector(self, scribe_env: str) -> str:
+        # this is kinda silly, but until the scribereader cli becomes more ergonomic
+        # we'll need to do a little bit of string munging to let humans use scribereader
+        # in the same way we are (tl;dr: scribereader has sorta confusing behavior between
+        # what can be use for --ecosystem, --region, and --superregion and the fastest/least
+        # hacky thing to figure out which we wanna use is that any env with a - in it is a region
+        # and any without one is an ecosystem)
+        return "-e" if "-" in scribe_env else "-r"
```
jfongatyelp, line 647: There might've been some shenanigans using `scribe_env_to_locations` since this checks if this is a region/superregion/ecosystem for us by looking it up in the `scribereader.PROD_REGIONS`/`scribereader.PROD_SUPERREGIONS` lists?

jfongatyelp (approved):
lol, simple but effective++

paasta_tools/cli/cmds/logs.py
nemacysts: oh, that would have been another good way to do this!

!choose use the simple yet effective way, do things in a more verbose manner so that i don't need to write a comment

mattmb:
Agreed, it's a bit hacky but still a better UX and we can always revisit. It looks safe enough that I'm just going to merge rather than ditching this PR.

PR #3473, opened 2022-10-01
COMPINFRA-1050, COMPINFRA-1043: fix paasta status output to show full version if provided

jfongatyelp:
Two issues fixed:
1. Stops using DeploymentsJsonV1 in `get_actual_deployments`. This was attempted once before but caused severe performance issues during bounces.  Because this now uses `PaastaServiceConfigLoader` those same performance issues have been mitigated, as we'll be loading the deployments.json only once per service, rather than clusters*instances times
In making this change, I found that the PaastaServiceConfigLoader comments were slightly incorrect about usage (`instance_type` vs `instance_type_class`), so fixed them
2. We had one reference to a top-level `git_sha` in our `InstanceStatus` type, so this adds an equivalent `version` field and sets it appropriately to the short representation of the DeploymentVersion.
Because this depends on a change to the API, the client side has a fallback to maintain status quo & use `git_sha` if `version` is not found yet.

### Verification
Tests all pass
1. Ran this version of the API against the same instances that were having problems bouncing with the last change, and confirmed API response for e.g. `v1/services/<service>/<instance>/status` returned in <30s, with similar results from calling the real API, which was not the case after https://github.com/Yelp/paasta/pull/3342.
2.  Testing with a modified API instance shows the correct output for both a non-no-commit instance and a no-commit instance: 
```
compute-infra-test-service.slow_termination in kubestage
    Version:    0ae13cf2 (desired)
    State: Running
    Running versions:
      0ae13cf2 - Started 2022-09-23 04:56:03 (7 days ago)
        Replica States: 1 Healthy

compute-infra-test-service.no_commits in kubestage
    Version:    DeploymentVersion(sha=0ae13cf2, image_version=20220929T181107) (desired)
    State: Running
    Running versions:
      0ae13cf2 (image_version: 20220929T181107) - Started 2022-09-29 18:18:03 (23 hours ago)
        Replica States: 1 Healthy 
```

And against current API w/o the `version` field in the response, still has slightly cosmetic problem of showing the wrong `Git sha` but includes the correct information under `Running versions`, and otherwise does not otherwise break the output:
```
compute-infra-test-service.slow_termination in kubestage
    Git sha:    0ae13cf2 (desired)
    State: Running
    Running versions:
      0ae13cf2 - Started 2022-09-23 04:56:03 (7 days ago)
        Replica States: 1 Healthy


compute-infra-test-service.no_commits in kubestage
    Git sha:    20220929 (desired)
    State: Running
    Running versions:
      0ae13cf2 (image_version: 20220929T181107) - Started 2022-09-29 18:18:03 (23 hours ago)
        Replica States: 1 Healthy
```

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -306,10 +306,10 @@ def paasta_status_on_api_endpoint(
         return 1
 
     if status.version and status.version != "":
-        output.append("    Version:    %s (desired)" % status.version)
+        output.append(f"    Version:    {status.version} (desired)")
     # TODO: Remove this when all clusters are returning status.version
     elif status.git_sha != "":
-        output.append("    Git sha:    %s (desired)" % status.git_sha)
+        output.append(f"    Git sha:    {status.git_sha} (desired)")
     instance_types = find_instance_types(status)
     if not instance_types:
         output.append(
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -72,6 +72,7 @@ from paasta_tools.marathon_tools import MarathonServiceConfig
 from paasta_tools.mesos_tools import format_tail_lines_for_mesos_task
 from paasta_tools.monitoring_tools import get_team
 from paasta_tools.monitoring_tools import list_teams
+from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.paastaapi.model.flink_job_details import FlinkJobDetails
 from paasta_tools.paastaapi.model.flink_jobs import FlinkJobs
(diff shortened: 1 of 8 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -301,9 +305,11 @@ def paasta_status_on_api_endpoint(
         output.append(str(e))
         return 1
 
-    if status.git_sha != "":
+    if status.version and status.version != "":
+        output.append("    Version:    %s (desired)" % status.version)
```
piax93, line 309: super nit, it'd be probably better to start using f-strings like in other places

piax93 (approved):

piax93 (approved):

PR #3474, opened 2022-10-07
Update bounce_margin_factor docs

nemacysts:
The defaults mentioned have not been accurate for a while.

### DIFF
--- a/docs/source/bouncing.rst
+++ b/docs/source/bouncing.rst
@@ -141,9 +141,9 @@ How to Select A Bounce Method
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 A service author can select a bounce method by setting ``bounce_method`` in
-the marathon configuration file. (e.g. ``marathon-SHARED.yaml``) This setting
+the kubernetes configuration file. (e.g. ``kubernetes-pnw-prod.yaml``) This setting
 is set per-instance. If not set, it will default to the ``crossover`` method.
-See the docs on the `marathon config <yelpsoa_configs.html#marathon-clustername-yaml>`_ file.
+See the docs on the `kubernetes config <yelpsoa_configs.html#kubernetes-clustername-yaml>`_ file.
 
 Additionally, a service author can configure how the bounce code determines
 which instances are healthy by setting ``bounce_health_params``. This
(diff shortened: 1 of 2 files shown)


docs/source/bouncing.rst
```diff
@@ -158,16 +158,16 @@ Valid options are:
...
+With the default setting (0.80) the ``crossover`` bounce will begin by draining and killing old copies of the code once new copies are healthy to replace them.
+For example, if ``instances: 10`` and ``bounce_margin_factor: 0.80`` (default), PaaSTA will make sure that at least 8 of the copies of a service are healthy/running before draining any additional copies.
 If 10 new copies of the service are up, then it will being draining the old 10 copies right away.
 If only one new copy of the service comes up, then the bounce will only drain one old copy as it gets replaced.
 
 The ``bounce_margin_factor`` adjusts how aggressive this procedure is.
-With the example of ``instances: 10`` and a ``bounce_margin_factor: 0.5`` (50%), then PaaSTA will preemptively being to drain and kill 5 copies of the old service to make room for the next 10 copies.
+With the example of ``instances: 10`` and a ``bounce_margin_factor: 0.8`` (80%), then PaaSTA will preemptively being to drain and kill 8 copies of the old service to make room for the next 10 copies.
```
EvanKrall, line 170: > then PaaSTA will preemptively being to drain and kill 8 copies of the old service

should be `2 copies` ((1-0.8) * 10)

EvanKrall (changes requested):
I'm pretty sure the default is still 1.0? https://github.com/Yelp/paasta/blob/bc5087f2d122539c34c49956082afa10a1d3519b/paasta_tools/long_running_service_tools.py#L369

nemacysts:
> I'm pretty sure the default is still 1.0?
> 
> https://github.com/Yelp/paasta/blob/bc5087f2d122539c34c49956082afa10a1d3519b/paasta_tools/long_running_service_tools.py#L369

@EvanKrall ah, looks like I got confused with how the logic works here since I was seeing PDBs that permitted 20% unavailable pods: if folks don't explicitly set a bounce_margin_factor in their configs it looks like we use system paasta config to configure the maxUnavailable value of their service's PDB in https://sourcegraph.yelpcorp.com/Yelp/paasta/-/blob/paasta_tools/kubernetes/application/controller_wrappers.py?L141

nemacysts:
hmm, this is going to be a little tricky to document here then - i'll have a think about what to do here since it seems a little iffy to tell folks to set b_m_f just for PDB purposes, but then also have that affect how their service bounces :/

docs/source/bouncing.rst
nemacysts: whoops, brain fart!

nemacysts:
closing this for now

PR #3475, opened 2022-10-08
Adding validation for when max_instances is missing in soaconfig

sidsatyelp:
Added a condition in kubernetes_schema.json to ensure that if min_instances is specified, max_instances should also be specified.
Added additional conditions in validate.py to block this while using the cli.
However we are now seeing validation failures for a lot of service soaconfigs, for example [yocobo](https://fluffy.yelpcorp.com/i/fTMWg1ScR8ZHcC0l7FT7pX0V0Jt0RT0p.html).

Will need some modifications before I can merge this one. Either update all soaconfig files with missing parameters or some other way to reduce the impact from this validation change.


paasta_tools/cli/cmds/validate.py
```diff
@@ -661,6 +661,18 @@ def validate_min_max_instances(service_path):
             if instance_config.get_instance_type() != "tron":
```
EvanKrall, line 661: I'm surprised this doesn't crash on adhoc instances right now - I don't think those would have `get_min_instances`/`get_max_instances`. I'd probably change this to `== "kubernetes"`.

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -780,6 +780,11 @@
                     "type": "integer",
                     "minimum": 1
                 }
+            },
+            "dependentRequired": {
+                "min_instances": [
+                    "max_instances"
+                ]
```
EvanKrall, line 787: do you also want the inverse relationship? i.e.

```
"max_instances": [
    "min_instances"
]
```

Also, if you've got this constraint specified in the schema, do you even need to write code that checks for the same thing?

paasta_tools/cli/cmds/validate.py
```diff
@@ -643,7 +643,7 @@ def validate_autoscaling_configs(service_path):
     return returncode
 
 
-def validate_min_max_instances(service_path):
+def validate_min_max_instances(service_path) -> bool:
```
EvanKrall, line 646: can you specify `service_path: str` as well?

paasta_tools/long_running_service_tools.py
```diff
@@ -314,6 +314,9 @@ def is_autoscaling_enabled(self) -> bool:
     def get_max_instances(self) -> Optional[int]:
         return self.config_dict.get("max_instances", None)
 
+    def get_instances_from_config(self) -> Optional[int]:
+        return self.config_dict.get("instances", None)
```
EvanKrall, line 318: Probably worth adding a docstring that explains why this method exists and how it differs from `get_instances`.

paasta_tools/cli/cmds/validate.py
```diff
@@ -658,7 +660,8 @@ def validate_min_max_instances(service_path):
                 load_deployments=False,
                 soa_dir=soa_dir,
             )
-            if instance_config.get_instance_type() != "tron":
+            instance_config = cast(LongRunningServiceConfig, instance_config)
```
nemacysts, line 663: is this required because typing the function signature made mypy type check this function?

nemacysts (approved):

paasta_tools/cli/schemas/kubernetes_schema.json
nemacysts: should be able to write a test that calls the validate function that uses the schema (there should. be some tests in https://github.com/Yelp/paasta/blob/master/tests/cli/test_cmds_validate.py that do this)

paasta_tools/cli/cmds/validate.py
sidsatyelp: I have updated the return type of `get_instance_config()` to `LongRunningServiceConfig`. The only concern here is that the LONG_RUNNING_INSTANCE_TYPE_HANDLERS does not implement `adhoc` and `tron` types. I don't see any references to `load_tron_instance_config`. Not sure how to handle the `adhoc` type and if that is still necessary?

sidsatyelp: ack

EvanKrall: tron and adhoc aren't in LONG_RUNNING_INSTANCE_TYPE_HANDLERS because they aren't long-running services. That is, they aren't meant to run indefinitely; tron jobs and adhoc jobs run until they complete. A long-running service like MarathonJobConfig/KubernetesServiceConfig is meant to run forever, and should get restarted when it exits. Long-running services are also the only ones that support running multiple copies, currently, which is why `get_min_instances()`/`get_max_instances()`/`get_instances()` are only defined on `LongRunningServiceConfig`, not on InstanceConfig.

I'd put the `cast()` inside the if statement, so it only gets cast if the instance type is `kubernetes`.

paasta_tools/cli/schemas/kubernetes_schema.json
sidsatyelp: Added tests for this

paasta_tools/cli/cmds/validate.py
```diff
@@ -658,7 +660,10 @@ def validate_min_max_instances(service_path):
                 load_deployments=False,
                 soa_dir=soa_dir,
             )
-            if instance_config.get_instance_type() != "tron":
+            # TODO: Remove this cast. This will require updating utils.LONG_RUNNING_INSTANCE_TYPE_HANDLERS to handle tron and adhoc type
+            # and change get_instance_config to return an object of type LongRunningServiceConfig
```
nemacysts, line 664: we'd probably want to do something slightly different since tron/adhoc aren't actually long-running so we don't want them to be of type `LongRunningServiceConfig` - i'd say it's probably fine to just leave this as a comment about how we're iterating over `InstanceConfig` objects and only certain instance types subclass `InstanceConfig` to provide autoscaling (and we don't partition these into long-running/not long-running at iteration time)

tests/cli/test_cmds_validate.py
```diff
@@ -192,6 +192,52 @@ def test_validate_min_max_instances_success(
...
+def test_validate_schema_min_max_instances_absent(mock_get_file_contents, capsys):
+    kubernetes_content = """
+invalid-max-only:
+  cpus: 0.1
+  max_instances: 3
+  mem: 250
+  disk: 512
+  cmd: virtualenv_run/bin/python adindexer/adindex_worker.py
+"""
+    mock_get_file_contents.return_value = kubernetes_content
```
nemacysts, line 205: we could probably store this as a dict and use `yaml.dump(kubernetes_content)` as the value of `mock_get_file_contents.return_value`, but that's pretty nit-picky so feel free to ignore this unless you like the dict form more :p

e.g,:
```suggestion
    kubernetes_content = {
        "invalid-max-only": {
            "cpus": 0.1
            "max_instances": 3
            "mem": 250
            "disk": 512
            "cmd": "virtualenv_run/bin/python adindexer/adindex_worker.py"
        }
    }
    mock_get_file_contents.return_value = yaml.dump(kubernetes_content)
```

nemacysts: (it would also be nice to make the mocking here use context managers as we usually do in newer tests/internally - but that's probably fine to ignore as well since every other test in this file is using the decorator approach to mocking)

nemacysts (commented):
the only real comment is the one about the TODO - looks good otherwise!

PR #3477, opened 2022-10-11
Secrets for teams feature in paasta secret cli

iiliescu2:
Checks wether a service has opted-in for team level access patterns to secrets in service configs.
If so, paasta secret decrypt cli changes logic to retrieve the secrets from k8s (using kubectl), instead of the default secret provider decryption. Otherwise, the logic remains unchanged.

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -238,10 +238,7 @@ def paasta_secret(args):
             )
             sys.exit(1)
 
-        try:
-            print(get_kubernetes_secret(args.secret_name, service, clusters[0]))
-        except Exception as e:
-            print(e)
+        print(get_kubernetes_secret(args.secret_name, service, clusters[0]))
         return
 
     secret_provider = _get_secret_provider_for_service(
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -238,7 +238,10 @@ def paasta_secret(args):
             )
             sys.exit(1)
 
-        print(get_kubernetes_secret(args.secret_name, service, clusters[0]))
+        try:
+            print(get_kubernetes_secret(args.secret_name, service, clusters[0]))
+        except Exception as e:
+            print(e)
         return
 
     secret_provider = _get_secret_provider_for_service(
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -17,14 +17,17 @@ import os
 import re
 import sys
 
+from service_configuration_lib import DEFAULT_SOA_DIR
+
 from paasta_tools.cli.utils import lazy_choices_completer
+from paasta_tools.kubernetes_tools import get_kubernetes_secret
 from paasta_tools.secret_tools import get_secret_provider
 from paasta_tools.secret_tools import SHARED_SECRET_SERVICE
 from paasta_tools.utils import _log_audit
+from paasta_tools.utils import is_secrets_for_teams_enabled
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import list_services
 from paasta_tools.utils import load_system_paasta_config
-
 
 SECRET_NAME_REGEX = r"([A-Za-z0-9_-]*)"
 
@@ -67,6 +70,13 @@ def add_subparser(subparsers):
         "services yaml files and should "
         "be unique per service.",
     )
+    secret_parser.add_argument(
+        "-y",
+        "--yelpsoa-config-root",
+        dest="yelpsoa_config_root",
+        help="A directory from which yelpsoa-configs should be read from",
+        default=DEFAULT_SOA_DIR,
+    )
 
     # Must choose valid service or act on a shared secret
     service_group = secret_parser.add_mutually_exclusive_group(required=True)
@@ -78,7 +88,6 @@ def add_subparser(subparsers):
         help="Act on a secret that can be shared by all services",
         action="store_true",
     )
-
     secret_parser.add_argument(
         "-c",
         "--clusters",
@@ -211,9 +220,36 @@ def paasta_secret(args):
             sys.exit(1)
     else:
         service = args.service
+
+    # Check if "decrypt" and "secrets_for_teams" first to avoid vault auth
+    if args.action == "decrypt":
+        secrets_for_teams_only = is_secrets_for_teams_enabled(
+            service, args.yelpsoa_config_root
+        )
+        if secrets_for_teams_only:
+            clusters = (
+                args.clusters.split(",")
+                if args.clusters
+                else list_clusters(service=service, soa_dir=os.getcwd())
+            )
+
+            if len(clusters) > 1 or not clusters:
+                print(
+                    "Can only decrypt for one specified cluster at a time!\nFor example,"
+                    " try '-c norcal-devc' to decrypt the secret for this service in norcal-devc."
+                )
+                sys.exit(1)
+
+            print(
+                get_kubernetes_secret(args.secret_name, service, clusters),
+                end="",
+            )
+            return
+
     secret_provider = _get_secret_provider_for_service(
         service, cluster_names=args.clusters
     )
+
     if args.action in ["add", "update"]:
         plaintext = get_plaintext_input(args)
         if not plaintext:
@@ -253,4 +289,5 @@ def decrypt_secret(secret_provider, secret_name):
             " to decrypt the secret for this service in norcal-devc."
         )
         sys.exit(1)
+
     return secret_provider.decrypt_secret(secret_name)
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -222,29 +222,24 @@ def paasta_secret(args):
         service = args.service
 
     # Check if "decrypt" and "secrets_for_teams" first to avoid vault auth
-    if args.action == "decrypt":
-        secrets_for_teams_only = is_secrets_for_teams_enabled(
-            service, args.yelpsoa_config_root
+    if args.action == "decrypt" and is_secrets_for_teams_enabled(
+        service, args.yelpsoa_config_root
+    ):
+        clusters = (
+            args.clusters.split(",")
+            if args.clusters
+            else list_clusters(service=service, soa_dir=os.getcwd())
         )
-        if secrets_for_teams_only:
-            clusters = (
-                args.clusters.split(",")
-                if args.clusters
-                else list_clusters(service=service, soa_dir=os.getcwd())
-            )
 
-            if len(clusters) > 1 or not clusters:
-                print(
-                    "Can only decrypt for one specified cluster at a time!\nFor example,"
-                    " try '-c norcal-devc' to decrypt the secret for this service in norcal-devc."
-                )
-                sys.exit(1)
-
+        if len(clusters) > 1 or not clusters:
             print(
-                get_kubernetes_secret(args.secret_name, service, clusters),
-                end="",
+                "Can only decrypt for one specified cluster at a time!\nFor example,"
+                " try '-c norcal-devc' to decrypt the secret for this service in norcal-devc."
             )
-            return
+            sys.exit(1)
+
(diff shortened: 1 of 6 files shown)


paasta_tools/utils.py
```diff
@@ -3091,6 +3092,13 @@ def get_pipeline_config(service: str, soa_dir: str = DEFAULT_SOA_DIR) -> List[Di
     return service_configuration.get("deploy", {}).get("pipeline", [])
 
 
+def is_secrets_for_teams_enabled(service: str, soa_dir: str = DEFAULT_SOA_DIR) -> bool:
+    service_yaml_contents = read_extra_service_information(service, "service", soa_dir)
+    if "secrets_for_owner_team" in service_yaml_contents:
+        return service_yaml_contents["secrets_for_owner_team"]
+    return False
```
piax93, line 3099: nit, this is effectively 
```suggestion
    return service_yaml_contents.get("secrets_for_owner_team", False)
```

paasta_tools/kubernetes_tools.py
```diff
@@ -3432,3 +3434,44 @@ def update_crds(
...
+
+
+def get_kubernetes_secret_name(service_name, secret_name, namespace="paasta"):
+    service = sanitise_kubernetes_name(service_name)
+    sanitised_secret = sanitise_kubernetes_name(secret_name)
+    name = f"{namespace}-secret-{service}-{sanitised_secret}"
+    return name
+
+
+def get_kubernetes_secret(secret_name, service_name, cluster):
```
piax93, line 3446: I'd make this method just accept a single cluster as a string, rather than expecting a list and then lookup index-0.

paasta_tools/cli/cmds/secret.py
```diff
@@ -211,9 +220,36 @@ def paasta_secret(args):
             sys.exit(1)
     else:
         service = args.service
+
+    # Check if "decrypt" and "secrets_for_teams" first to avoid vault auth
+    if args.action == "decrypt":
+        secrets_for_teams_only = is_secrets_for_teams_enabled(
+            service, args.yelpsoa_config_root
+        )
+        if secrets_for_teams_only:
```
piax93, line 229: uber mega nit, if you just do
```suggestion
    if args.action == "decrypt" and is_secrets_for_teams_enabled(
        service, args.yelpsoa_config_root
    ):
```
you save 1 level of indentation (and still avoid the file read if `action != "decrypt"`)

paasta_tools/cli/cmds/secret.py
```diff
@@ -211,9 +220,36 @@ def paasta_secret(args):
...
+            if len(clusters) > 1 or not clusters:
+                print(
+                    "Can only decrypt for one specified cluster at a time!\nFor example,"
+                    " try '-c norcal-devc' to decrypt the secret for this service in norcal-devc."
+                )
+                sys.exit(1)
+
+            print(
+                get_kubernetes_secret(args.secret_name, service, clusters),
+                end="",
```
piax93, line 245: I think it's alright the have the trailing newline, this is not really meant to be called from other scripts anyway. And avoids the issue for having the next bash prompt on the same line of the output.

piax93 (commented):
Apart from the few suggestions, looks pretty alright to me. You'll also want to get an opinion from someone in CI though.

paasta_tools/kubernetes_tools.py
```diff
@@ -3432,3 +3434,44 @@ def update_crds(
...
+        stdout=subprocess.PIPE,
+        env=env,
+    )
+
+    # Error running kubectl
+    if proc.returncode != 0:
+        process_errors = proc.stderr.decode("utf-8").strip()
+        if process_errors:
+            print(process_errors)
+        return "Could not get secret.\n"
```
piax93, line 3473: Sorry for not catching it in the previous review, this should probably just throw an exception.

paasta_tools/kubernetes_tools.py
```diff
@@ -3432,3 +3434,44 @@ def update_crds(
...
+
+    proc = subprocess.run(
+        [
+            "kubectl",
+            "get",
+            "secrets",
+            f"--cluster={cluster}",
+            f"-n=paasta",
+            f"{k8s_secret_name}",
+            f"--template={{{template_name}}}",
```
ramonatame, line 3461: Here we pass user input into a OS command, hence we are prone to command injection.  I would recommend to sanitize `cluster`, `k8s_secret_name` and `template_name` to avoid this vulnerability.
(I had a look at `sanitize_kubernetes_name` and what it does it doesn't help for this issue in particular.)

ramonatame (commented):
The rest of the PR looks good to me.

paasta_tools/kubernetes_tools.py
ramonatame: There are multiple sources explaining how to prevent command injection. Of all, my favorite is [PortSwigger](https://portswigger.net/web-security/os-command-injection) (which I find better than OWASP here and there).

ramonatame: It might not worth sanitizing it (other than best practice) if the following are true:
- this command would run on devbox
- the user already has access to the devbox
- the user has the same OS permissions as this service, so they wouldn't gain anything, other than potentially some repudiation

If the service has different permissions than devs (as in extra permissions), then it might worth implementing it.

I leave it up to you :)

iiliescu2: Added some sanitisation on alphanum chars and -, _ on all input.

piax93: This is an exec call with no shell involved though, so you can't really do command injection.

paasta_tools/cli/cmds/secret.py
```diff
@@ -238,7 +238,10 @@ def paasta_secret(args):
             )
             sys.exit(1)
 
-        print(get_kubernetes_secret(args.secret_name, service, clusters[0]))
+        try:
+            print(get_kubernetes_secret(args.secret_name, service, clusters[0]))
+        except Exception as e:
+            print(e)
         return
```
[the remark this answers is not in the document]
piax93: I'd be fine with letting the exception bubble up to the user without catching it, it's what the Vault provider does already anyway.

paasta_tools/kubernetes_tools.py
```diff
@@ -3444,6 +3444,12 @@ def get_kubernetes_secret_name(service_name, secret_name, namespace="paasta"):
 
 
 def get_kubernetes_secret(secret_name, service_name, cluster):
+    # sanitise input before cmd run
+    sanitise_exp = r"[^a-zA-Z0-9-_]"
```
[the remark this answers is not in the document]
iiliescu2: removed it

piax93 (approved):
LGTM

paasta_tools/cli/cmds/secret.py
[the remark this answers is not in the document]
iiliescu2: removed the try except then 👌

paasta_tools/kubernetes_tools.py
```diff
@@ -3432,3 +3434,44 @@ def update_crds(
...
+
+
+def get_kubernetes_secret(secret_name: str, service_name: str, cluster: str) -> str:
+    k8s_secret_name = get_kubernetes_secret_name(service_name, secret_name)
+    template_name = f"{{.data.{secret_name}}}"
+
+    env = os.environ.copy()
+    env["KUBECONFIG"] = DEFAULT_KUBECTL_CONFIG_PATH
+
+    proc = subprocess.run(
```
nemacysts, line 3455: can we the python kubernetes library instead of using subprocess?

iiliescu2: done.

paasta_tools/kubernetes_tools.py
```diff
@@ -3432,3 +3433,25 @@ def update_crds(
...
+) -> str:
+    service = sanitise_kubernetes_name(service_name)
+    sanitised_secret = sanitise_kubernetes_name(secret_name)
+    name = f"{namespace}-secret-{service}-{sanitised_secret}"
+    return name
+
+
+def get_kubernetes_secret(secret_name: str, service_name: str, cluster: str) -> str:
+    os.environ["KUBECONFIG"] = DEFAULT_KUBECTL_CONFIG_PATH
+    os.environ["KUBECONTEXT"] = cluster
```
nemacysts, line 3449: we probably want to do the kubeconfig/kubecontext stuff in cli/cmds/secret.py - otherwise, anyone trying to use this in paasta internals is going to potentially overwrite environment variables in undesirable ways

e.g., if someone where to call this in setup_kubernetes_job, our kube masters would switch from using the paasta admin creds to using the user creds here (and also need to start passing around cluster arguments)

iiliescu2: To double check I understand the suggestion, modifying the KubeClient constructor to this:
`class KubeClient:
    def __init__(self, component: Optional[str] = None, config_file: Optional[str] = None, context: Optional[str] = None) -> None:
        if config_file and context:
            kube_config.load_kube_config(
                config_file=config_file,
                context=context
            )
        else:
            kube_config.load_kube_config(
                config_file=os.environ.get("KUBECONFIG", KUBE_CONFIG_PATH),
                context=os.environ.get("KUBECONTEXT"),
            )
            ...`
and then calling it with the optional arguments, without setting environment variables:
`kube_client = KubeClient(config_file = DEFAULT_KUBECTL_CONFIG_PATH, context = cluster)`
should do the trick for us, without breaking other things that still take their configs from env vars, right?

BeryJu: yeah pretty much, allthough imo this is a bit easier to read:

```python
class KubeClient:
    def __init__(self, component: Optional[str] = None, config_file: Optional[str] = None, context: Optional[str] = None) -> None:
        if not config_file:
            config_file = os.environ.get("KUBECONFIG", KUBE_CONFIG_PATH)
        if not context:
            context = os.environ.get("KUBECONTEXT")
        kube_config.load_kube_config(
            config_file=config_file,
            context=context,
        )
        ...
```

paasta_tools/kubernetes_tools.py
```diff
@@ -158,6 +158,7 @@
 log = logging.getLogger(__name__)
 
 KUBE_CONFIG_PATH = "/etc/kubernetes/admin.conf"
+DEFAULT_KUBECTL_CONFIG_PATH = "/etc/kubernetes/paasta.conf"
```
BeryJu, line 161: nit: maybe rename this to something that makes it clearer why there's two kubeconfig files, like `KUBE_CONFIG_HUMAN_PATH` or `KUBE_CONFIG_USER_PATH`

BeryJu (approved):

BeryJu (approved):

paasta_tools/kubernetes_tools.py
```diff
@@ -3432,3 +3443,23 @@ def update_crds(
...
+    service = sanitise_kubernetes_name(service_name)
+    sanitised_secret = sanitise_kubernetes_name(secret_name)
+    name = f"{namespace}-secret-{service}-{sanitised_secret}"
+    return name
+
+
+def get_kubernetes_secret(secret_name: str, service_name: str, cluster: str) -> str:
+    k8s_secret_name = get_kubernetes_secret_name(service_name, secret_name)
+
+    kube_client = KubeClient(config_file=KUBE_CONFIG_USER_PATH, context=cluster)
```
nemacysts, line 3460: @iiliescu2 we probably want to refactor this a little further to look like the other helpers in this file: we should probably do something like [create_secret](https://sourcegraph.yelpcorp.com/Yelp/paasta@badf618668149abff341bc8d2383ff4f9625de42/-/blob/paasta_tools/kubernetes_tools.py?L3012) and pass in a KubeClient - as it is, this helper cannot be used outside of cli tooling since it specifies a default kubeconfig file to use.

passing in a kubeclient would also make it so that we don't need to take in a cluster param here (which would be more consistent with the rest of the helpers in this file)

iiliescu2: Missed it then, but I included the refactoring in: https://github.com/Yelp/paasta/pull/3482

PR #3480, opened 2022-10-12
FLINK-4163: Add more generic config loading for setup cr

AhmedSoliman7:
The changes attempt to customly load flink configurations when updating crs of kind Flink. 
Context: 
- Yelp internal flink operator looks for changes in CRs specs and apply the changes by reconciling the flink clusters. 
- In scope of https://jira.yelpcorp.com/browse/FLINK-4103 project, this is needed to load extra configuration for auto config files on updating Flink k8s CRs (if exists)
- This is similar to the approach taken when reading instances configs for paasta services with autotune (y/autotune)

Open questions: 
- Is this the right way to modify? Any feedback would be appreciated given the basic knowledge of paasta I have. 
- How to test this is completely working without rolling it out? 
- Why the [step](https://github.com/Yelp/paasta/actions/runs/3242377865/jobs/5315550295) is failing?

### DIFF
--- a/paasta_tools/setup_kubernetes_cr.py
+++ b/paasta_tools/setup_kubernetes_cr.py
@@ -47,8 +47,8 @@ from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import get_config_hash
 from paasta_tools.utils import get_git_sha_from_dockerurl
 from paasta_tools.utils import load_all_configs
-from paasta_tools.utils import load_all_flink_configs
 from paasta_tools.utils import load_system_paasta_config
+
 
 log = logging.getLogger(__name__)
 
@@ -168,20 +168,16 @@ def setup_all_custom_resources(
             continue
 
         # by convention, entries where key begins with _ are used as templates
-        raw_config_dicts = {}
-        if crd.kube_kind.singular == "flink":
-            raw_config_dicts = load_all_flink_configs(cluster=cluster, soa_dir=soa_dir)
-        else:
-            raw_config_dicts = load_all_configs(
-                cluster=cluster, file_prefix=crd.file_prefix, soa_dir=soa_dir
-            )
-        config_dicts = {}
-        for svc, raw_sdict in raw_config_dicts.items():
-            sdict = {inst: idict for inst, idict in raw_sdict.items() if inst[0] != "_"}
-            if sdict:
-                config_dicts[svc] = sdict
-        if not config_dicts:
-            continue
+        config_dicts = load_all_configs(
+            cluster=cluster, file_prefix=crd.file_prefix, soa_dir=soa_dir
+        )
+        # config_dicts = {}
+        # for svc, raw_sdict in raw_config_dicts.items():
+        #     sdict = {inst: idict for inst, idict in raw_sdict.items() if inst[0] != "_"}
+        #     if sdict:
+        #         config_dicts[svc] = sdict
+        # if not config_dicts:
+        #     continue
 
         ensure_namespace(
             kube_client=kube_client, namespace=f"paasta-{crd.kube_kind.plural}"
--- a/paasta_tools/setup_kubernetes_cr.py
+++ b/paasta_tools/setup_kubernetes_cr.py
@@ -168,16 +168,10 @@ def setup_all_custom_resources(
             continue
 
         # by convention, entries where key begins with _ are used as templates
+        # and will be filter out here
         config_dicts = load_all_configs(
             cluster=cluster, file_prefix=crd.file_prefix, soa_dir=soa_dir
         )
-        # config_dicts = {}
-        # for svc, raw_sdict in raw_config_dicts.items():
-        #     sdict = {inst: idict for inst, idict in raw_sdict.items() if inst[0] != "_"}
-        #     if sdict:
-        #         config_dicts[svc] = sdict
-        # if not config_dicts:
-        #     continue
 
         ensure_namespace(
             kube_client=kube_client, namespace=f"paasta-{crd.kube_kind.plural}"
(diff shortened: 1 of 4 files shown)


JoeMalt:
The logs for the `tox` step include this line:
```
ERROR:   mypy: commands failed
```
This could be caused by missing type annotations - Python obviously doesn't require them, but mypy can be configured to error out if they're not there.

AhmedSoliman7:
> The logs for the `tox` step include this line:
> 
> ```
> ERROR:   mypy: commands failed
> ```
> 
> This could be caused by missing type annotations - Python obviously doesn't require them, but mypy can be configured to error out if they're not there.

Can not really relate, my python experience is 👶🏻, trying to do more debugging. However this is not firing locally.

JoeMalt (approved):
This looks fine to me. I think it's worth waiting for a paasta expert though.

AhmedSoliman7:
@matthewbentley Can you please have a look? 
I am not able also to know why mypy is failing.

nemacysts:
@AhmedSoliman7 you should be able to run `tox -e mypy` to get mypy running locally (not sure why we don't have make test run mypy though...)

but the errors are:
```
paasta_tools/setup_kubernetes_cr.py:171: error: Need type annotation for 'raw_config_dicts' (hint: "raw_config_dicts: Dict[<type>, <type>] = ...")
paasta_tools/setup_kubernetes_cr.py:173: error: Incompatible types in assignment (expression has type "Mapping[str, Mapping[str, Any]]", variable has type "Dict[Any, Any]")
paasta_tools/setup_kubernetes_cr.py:175: error: Incompatible types in assignment (expression has type "Mapping[str, Mapping[str, Any]]", variable has type "Dict[Any, Any]")
```

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -167,9 +168,13 @@ def setup_all_custom_resources(
             continue
 
         # by convention, entries where key begins with _ are used as templates
-        raw_config_dicts = load_all_configs(
-            cluster=cluster, file_prefix=crd.file_prefix, soa_dir=soa_dir
-        )
+        raw_config_dicts = {}
+        if crd.kube_kind.singular == "flink":
+            raw_config_dicts = load_all_flink_configs(cluster=cluster, soa_dir=soa_dir)
```
nemacysts, line 173: could we do this in `load_all_configs` so that the loading logic is in a single place? should be able to use `file_prefix` to determine if we need to flink-specific logic or not

nemacysts: that said, i wonder if we could just have load_all_configs work like `load_service_instance_configs` and just automatically return the merged configs for everything (if not just call `load_service_instance_configs)? 

that way, future operators that want to be autotuned don't need to change any code :)

AhmedSoliman7: ++ 

I will work on this! Thanks for the suggestion

AhmedSoliman7:
@nemacysts Do you mean the current approach? Do you think it will work with flink services just fine? or do I need more modifications?

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -167,16 +168,16 @@ def setup_all_custom_resources(
...
-                config_dicts[svc] = sdict
-        if not config_dicts:
-            continue
+        # config_dicts = {}
+        # for svc, raw_sdict in raw_config_dicts.items():
+        #     sdict = {inst: idict for inst, idict in raw_sdict.items() if inst[0] != "_"}
+        #     if sdict:
+        #         config_dicts[svc] = sdict
+        # if not config_dicts:
+        #     continue
```
nemacysts, line 180: should we delete this commented out code?

paasta_tools/utils.py
```diff
@@ -4039,14 +4039,37 @@ def load_all_configs(
...
-            service
-        ] = service_configuration_lib.read_extra_service_information(
-            service, f"{file_prefix}-{cluster}", soa_dir=soa_dir
+        config_dicts[service] = load_service_instance_configs(
+            service, file_prefix, cluster, soa_dir
         )
     return config_dicts
 
 
+def load_all_flink_configs(
```
nemacysts, line 4048: doesn't seem like this is being called anymore: shall we go ahead and delete it?

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -167,16 +168,16 @@ def setup_all_custom_resources(
             continue
 
         # by convention, entries where key begins with _ are used as templates
-        raw_config_dicts = load_all_configs(
+        config_dicts = load_all_configs(
```
nemacysts, line 171: we'll need to also add flink to the `paasta_tools::public_config::auto_config_instance_types_enabled` list in puppet

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -167,16 +168,16 @@ def setup_all_custom_resources(
             continue
 
         # by convention, entries where key begins with _ are used as templates
```
nemacysts, line 170:
```suggestion
        # by convention, entries where key begins with _ are used as templates
        # and will be filter out here
```

nemacysts:
> Do you mean the current approach? Do you think it will work with flink services just fine? or do I need more modifications?

@AhmedSoliman7 yup, i was thinking of something like this :)

paasta_tools/setup_kubernetes_cr.py
AhmedSoliman7: Yes I would do that 

[DraftPR](https://github.yelpcorp.com/sysgit/puppet/pull/7734)

AhmedSoliman7:
@nemacysts Is there a way to test these changes without releasing?

nemacysts:
(there's a thread in #stream-processing in slack)

AhmedSoliman7:
Tested on infrastage and looks fine. 
Can this be pushed to devc only first for some time before releasing to prod?

AhmedSoliman7:
prerequisite PR is ready https://github.yelpcorp.com/sysgit/puppet/pull/7876

PR #3489, opened 2022-10-28
Fixing make go_entrypoint target

cuza:
Adding wget output file to fix make target.

### DIFF
--- a/yelp_package/Makefile
+++ b/yelp_package/Makefile
@@ -54,7 +54,7 @@ go_entrypoint:
 	rm -rf gopath && \
 	mkdir -p gopath && \
 	cd gopath && \
-	wget https://github.com/Yelp/paasta-tools-go/archive/v$(PTG_VERSION).zip && \
+	wget https://github.com/Yelp/paasta-tools-go/archive/v$(PTG_VERSION).zip -O v$(PTG_VERSION).zip && \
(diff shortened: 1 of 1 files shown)


bobokun (approved):

jfongatyelp (approved):
Seems fine to support new folks running this as themselves who may have a different .wgetrc that could impact what file this is saved to.

yelp_package/Makefile
```diff
@@ -54,7 +54,7 @@ go_entrypoint:
 	rm -rf gopath && \
 	mkdir -p gopath && \
 	cd gopath && \
-	wget https://github.com/Yelp/paasta-tools-go/archive/v$(PTG_VERSION).zip && \
+	wget https://github.com/Yelp/paasta-tools-go/archive/v$(PTG_VERSION).zip -O v$(PTG_VERSION).zip && \
```
nemacysts, line 57: do we need to set the filename? from reading the manpage, it seems like we should be able to just use -O (or preferably the --long-form of this option `--remote-name`) to ensure that the file is named the same as it was on the remote side

```
-O, --remote-name
              Write output to a local file named like the remote file we get. (Only the file part of the remote file is used, the path is cut off.
```

sidsatyelp (approved):

mattmb:

PR #3490, opened 2022-10-28
A python script that measures the time for get pod from a specific ns - PAASTA-17789

EmanElsaban:
### purpose of PR

The purpose of this script is to measure the time it takes to call the api to get the pods for a specific namespace.

### Testing

Tested the script manually

### DIFF
--- a/paasta_tools/monitoring/check_k8s_api_performance.py
+++ b/paasta_tools/monitoring/check_k8s_api_performance.py
@@ -1,0 +1,110 @@
+#!/usr/bin/env python
+# Copyright 2015-2016 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import argparse
+import time
+
+from kubernetes import client
+
+from paasta_tools.kubernetes_tools import KUBE_CONFIG_USER_PATH
+from paasta_tools.kubernetes_tools import KubeClient
+
+
+NAMESPACE = "paasta"
+LABEL_SELECTOR = None
+
+
+def parse_k8s_api_performance_options() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+
+    parser.add_argument(
+        "--kube-config-path",
+        "-k",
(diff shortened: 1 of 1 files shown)


paasta_tools/monitoring/check_k8s_api_performance.py
```diff
@@ -0,0 +1,129 @@
...
+    # By loading the auth and cluster information and storing it in client
+    config.load_kube_config(
+        config_file=kube_config_path,
+        context=context,
+    )
+
+    # Store the API object
+    v1 = client.CoreV1Api()
+
+    return v1
```
jfongatyelp, line 89: All of our interactions w/ k8s in paasta are abstracted using our `paasta_tools.kubernetes_tools.KubeClient` wrapper; i think it's probably worth reusing that rather than create another place where we have to define how to create our k8s client directly.

It may also help identify performance issues in case any of the boilerplate in KubeClient is adding latency (unlikely since it probably just boils down to exactly the calls you're doing here, but possible in the future we add some caching or reauth setup that interferes?)

EmanElsaban: Thank you, I will use the wrapper instead

EmanElsaban: In ``kubernetes_tools.py`` file where KubeClient is defined, I see 
```
KUBE_CONFIG_PATH = "/etc/kubernetes/admin.conf"
```
I think this path should be:
```
KUBE_CONFIG_PATH = "/etc/kubernetes/paasta.conf"
```
As the way it is rn doesn't work and I have to specify ``-k /etc/kubernetes/paasta.conf``. Should I also change this default path to ``/etc/kubernetes/paasta.conf`` ?

jfongatyelp: So, normally we are using `KubeClient` from restricted locations in order to do things to work with/create k8s resources for paasta (or operators), and those have currently been depending on the rights granted by `/etc/kubernetes/admin.conf`.  
So we'd need to make sure everything that currently uses KubeClient is updated to specifically use that path if we changed the default -- it'd fail with `/etc/kubernetes/paasta.conf` (and paasta.conf may only be put into place where users work, like devboxes). I think it's fine to leave as is currently for now.

paasta_tools/monitoring/check_k8s_api_performance.py
```diff
@@ -0,0 +1,109 @@
...
+def parse_k8s_api_performance_options() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
+
+    parser.add_argument(
+        "--kube-config-path",
+        "-k",
+        dest="kube_config_path",
+        default=None,
```
jfongatyelp, line 36: This should still probably default to the new KUBE_CONFIG_USER_CONF that was added in  https://github.com/Yelp/paasta/pull/3477/files (tl;dr this is the same as your original code, `/etc/kubernetes/paasta.conf`)

jfongatyelp (approved):

PR #3491, opened 2022-10-29
SLO Watcher Threads take precendence over Metric Watcher Threads

lanceleih:
For the time being, we're not able to run SLO Watcher Threads and Metric Watcher Threads at the same time. 

If there are no SLOs present, Metric Watcher Threads may run.

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -80,6 +80,7 @@ from paasta_tools.utils import _log_audit
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import DeploymentVersion
 from paasta_tools.utils import format_tag
+from paasta_tools.utils import get_files_of_type_in_dir
 from paasta_tools.utils import get_git_url
 from paasta_tools.utils import get_paasta_tag_from_deploy_group
 from paasta_tools.utils import get_username
@@ -361,6 +362,24 @@ def can_user_deploy_service(deploy_info: Dict[str, Any], service: str) -> bool:
             print(logline, file=sys.stderr)
             return False
     return True
+
+
+def can_run_metric_watcher_threads(
+    service: str,
+    soa_dir: str,
+) -> bool:
+    """
+    Cannot run slo and metric watcher threads together for now.
+    SLO Watcher Threads take precedence over metric watcher threads.
+    Metric Watcher Threads can run if there are no SLOs available.
+    """
+    slo_files = get_files_of_type_in_dir(
+        file_type="slo", service=service, soa_dir=soa_dir
+    )
+    rollback_files = get_files_of_type_in_dir(
+        file_type="rollback", service=service, soa_dir=soa_dir
+    )
+    return bool(not slo_files and rollback_files)
 
 
 def report_waiting_aborted(service: str, deploy_group: str) -> None:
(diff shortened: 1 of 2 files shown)


paasta_tools/utils.py
```diff
@@ -2995,6 +2996,27 @@ def get_hostname() -> str:
...
+    :return: a list
+    """
+    service = "**" if service is None else service
+    soa_dir = DEFAULT_SOA_DIR if soa_dir is None else soa_dir
+    file_type += "-*.yaml"
+    return [
+        file_path
+        for file_path in glob.glob(
+            os.path.join(soa_dir, service, file_type),
+            recursive=False,
```
nemacysts, line 3015: does this not need to be True for ** to recurse?

paasta_tools/utils.py
```diff
@@ -2995,6 +2996,27 @@ def get_hostname() -> str:
...
+) -> List[str]:
+    """Recursively search path if type of file exists.
+
+    :param file_name: a string of a type of a file (kubernetes, slo, etc.)
+    :param path: a string path
+    :return: a list
+    """
+    service = "**" if service is None else service
+    soa_dir = DEFAULT_SOA_DIR if soa_dir is None else soa_dir
+    file_type += "-*.yaml"
```
nemacysts, line 3010: should we make `file_type` required? right now this will error out if no `file_type` is provided upon trying to add a string to None

lanceleih: Ooh yes.

paasta_tools/utils.py
lanceleih: Ah right. This only captures slos written outside of a slo directory. Changed to True!

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -363,6 +364,21 @@ def can_user_deploy_service(deploy_info: Dict[str, Any], service: str) -> bool:
...
+    SLO Watcher Threads take precedence over metric watcher threads.
+    Metric Watcher Threads can run if there are no SLOs available.
+    """
+    slo_files = get_files_of_type_in_dir(
+        file_type="slo", service=service, soa_dir=soa_dir
+    )
+    rollback_files = get_files_of_type_in_dir(
+        file_type="rollback", service=service, soa_dir=soa_dir
+    )
+    return len(slo_files) < 1 and len(rollback_files) > 0
```
nemacysts, line 379:
```suggestion
    return not slo_files and rollback_files
```
might be a little more pythonic 

otherwise, the < 1 should probably be an == 0 😄

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -363,6 +365,24 @@ def can_user_deploy_service(deploy_info: Dict[str, Any], service: str) -> bool:
...
+    SLO Watcher Threads take precedence over metric watcher threads.
+    Metric Watcher Threads can run if there are no SLOs available.
+    """
+    slo_files = get_files_of_type_in_dir(
+        file_type="slo", service=service, soa_dir=soa_dir
+    )
+    rollback_files = get_files_of_type_in_dir(
+        file_type="rollback", service=service, soa_dir=soa_dir
+    )
+    return not slo_files and rollback_files
```
nemacysts, line 383: oh, sorry: i think we'll want this to be 
```suggestion
    return bool(not slo_files and rollback_files)
```

so that this will always return a boolean (and thus avoid the need to type this as returning a union)

lanceleih:
Tried using Literal for `get_files_of_type_in_dir`, but it's gotten a bit complicated. Might need to make it a separate PR. Set a TODO note for it anyway.

nemacysts (approved):

PR #3493, opened 2022-11-03
 Sync paasta spark secrets for paasta-spark namespace

88manpreet:
[MLCOMPUTE-516](https://jira.yelpcorp.com/browse/MLCOMPUTE-516)
Add support for syncing paasta secrets onto k8s clusters for `paasta-spark` namespace 

The issue was `tron` instance_type is mapped with `tron` namespace, which is mapped with `tron*` .yaml files.

Added description:
These changes also take care of many edge cases with existing way of choosing and restructuring the tricky and confusing  tron namespaces. 

Trick used is, `spark` instance_type maps to `tron*` .yaml files and `paasta-spark` namespace.

Testing: 
* Unit tests
* Manual tests: 
-- I have manually tested if I am able get get correct instance list for instance_type , `spark`
Result in the ticket. 

I can try to run this sync script manually if:
-- I can get permission from CI to run it and
-- I have authorization to run the same

Next step is to integrate this in `puppet`. 
Note:  For some reason the tests were not failing locally but fail here. 
Hopefully now they are fixed.

### DIFF
--- a/paasta_tools/list_tron_namespaces.py
+++ b/paasta_tools/list_tron_namespaces.py
@@ -16,6 +16,9 @@ import argparse
 from paasta_tools import tron_tools
 
 
+EXECUTOR_TYPES = ["paasta", "spark", "ssh_not_spark"]
+
+
 def parse_args():
     parser = argparse.ArgumentParser(
         description="Lists Tron namespaces for a cluster, excluding MASTER"
@@ -34,6 +37,14 @@ def parse_args():
         default=tron_tools.DEFAULT_SOA_DIR,
         help="Use a different soa config directory",
     )
+    parser.add_argument(
+        "-e",
+        "--executors",
+        default=EXECUTOR_TYPES,
+        nargs="+",
+        help="tron executor types to list namespaces for",
+        choices=EXECUTOR_TYPES,
+    )
     args = parser.parse_args()
     return args
 
@@ -41,7 +52,9 @@ def parse_args():
 def main():
     args = parse_args()
     namespaces = tron_tools.get_tron_namespaces(
-        cluster=args.cluster, soa_dir=args.soa_dir
+        cluster=args.cluster,
+        soa_dir=args.soa_dir,
+        tron_executors=args.executors,
     )
     print("\n".join(namespaces))
 
--- a/paasta_tools/list_tron_namespaces.py
+++ b/paasta_tools/list_tron_namespaces.py
@@ -16,9 +16,6 @@ import argparse
 from paasta_tools import tron_tools
 
 
-EXECUTOR_TYPES = ["paasta", "spark", "ssh_not_spark"]
-
-
 def parse_args():
     parser = argparse.ArgumentParser(
         description="Lists Tron namespaces for a cluster, excluding MASTER"
@@ -40,10 +37,10 @@ def parse_args():
     parser.add_argument(
         "-e",
         "--executors",
-        default=EXECUTOR_TYPES,
+        default=tron_tools.EXECUTOR_TYPES,
         nargs="+",
         help="tron executor types to list namespaces for",
-        choices=EXECUTOR_TYPES,
+        choices=tron_tools.EXECUTOR_TYPES,
     )
     args = parser.parse_args()
     return args
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -100,7 +100,7 @@ EXECUTOR_TYPE_TO_NAMESPACE = {
 }
 DEFAULT_TZ = "US/Pacific"
 clusterman_metrics, _ = get_clusterman_metrics()
-EXECUTOR_TYPES = ["paasta", "spark"]
+EXECUTOR_TYPES = ["paasta", "ssh"]
 
 
 class FieldSelectorConfig(TypedDict):
@@ -1190,16 +1190,7 @@ def validate_complete_config(
 
 def _is_valid_namespace(job: Any, tron_executors: List[str]) -> bool:
     for action_info in job.get("actions", {}).values():
-        if "paasta" in tron_executors and action_info.get("executor", "") in [
-            "paasta",
-            "",
-        ]:
-            return True
-        elif (
-            "spark" in tron_executors
-            and action_info.get("executor", "") == "ssh"
-            and "spark-run" in action_info.get("command", "")
-        ):
+        if action_info.get("executor", "paasta") in tron_executors:
             return True
     return False
 
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -1187,7 +1187,35 @@ def validate_complete_config(
     return []
 
 
-def get_tron_namespaces(cluster, soa_dir):
+def _is_valid_namespace(
+    job: {str: TronActionConfigDict}, tron_executors: List[str]
+) -> bool:
+    for _, action_info in job.get("actions", {}).items():
+        if "paasta" in tron_executors and action_info.get("executor", "") in [
+            "paasta",
+            "",
+        ]:
+            return True
+        elif (
+            "spark" in tron_executors
+            and action_info.get("executor", "") == "ssh"
+            and "spark-run" in action_info.get("command", "")
+        ):
+            return True
+        elif (
+            "ssh_not_spark" in tron_executors
+            and action_info.get("executor", "") == "ssh"
+            and "spark-run" not in action_info.get("command", "")
+        ):
+            return True
+    return False
+
+
+def get_tron_namespaces(
+    cluster: str,
+    soa_dir: str,
+    tron_executors: List[str] = ["paasta"],
+) -> List[str]:
     tron_config_file = f"tron-{cluster}.yaml"
     config_dirs = [
         _dir[0]
@@ -1195,7 +1223,21 @@ def get_tron_namespaces(cluster, soa_dir):
         if tron_config_file in _dir[2]
     ]
     namespaces = [os.path.split(config_dir)[1] for config_dir in config_dirs]
-    return namespaces
+    tron_namespaces = set()
+    for namespace in namespaces:
+        config = filter_templates_from_config(
+            read_extra_service_information(
+                namespace,
+                extra_info=f"tron-{cluster}",
+                soa_dir=soa_dir,
+                deepcopy=False,
+            )
+        )
+        for _, job in config.items():
+            if _is_valid_namespace(job, tron_executors):
+                tron_namespaces.add(namespace)
+                break
+    return list(tron_namespaces)
 
 
 def list_tron_clusters(service: str, soa_dir: str = DEFAULT_SOA_DIR) -> List[str]:
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -100,6 +100,7 @@ EXECUTOR_TYPE_TO_NAMESPACE = {
 }
 DEFAULT_TZ = "US/Pacific"
 clusterman_metrics, _ = get_clusterman_metrics()
+EXECUTOR_TYPES = ["paasta", "spark"]
 
 
 class FieldSelectorConfig(TypedDict):
@@ -1187,10 +1188,8 @@ def validate_complete_config(
     return []
 
 
-def _is_valid_namespace(
-    job: {str: TronActionConfigDict}, tron_executors: List[str]
-) -> bool:
-    for _, action_info in job.get("actions", {}).items():
+def _is_valid_namespace(job: Any, tron_executors: List[str]) -> bool:
+    for action_info in job.get("actions", {}).values():
         if "paasta" in tron_executors and action_info.get("executor", "") in [
             "paasta",
             "",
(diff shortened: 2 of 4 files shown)


paasta_tools/utils.py
```diff
@@ -135,6 +135,7 @@
     "kafkacluster",
     "monkrelays",
     "nrtsearchservice",
+    "spark",
```
nemacysts, line 138: we should probably add a comment here since spark isn't really an instance type in the normal paasta sense (i.e., we don't have spark-$CLUSTER.yaml files in soaconfigs)

paasta_tools/utils.py
```diff
@@ -3076,20 +3078,28 @@ def read_service_instance_names(
     service: str, instance_type: str, cluster: str, soa_dir: str
 ) -> Collection[Tuple[str, str]]:
     instance_list = []
-    conf_file = f"{instance_type}-{cluster}"
+    if instance_type == "spark":
```
nemacysts, line 3081: we should probably add a comment here as well since this is going to look pretty surprising otherwise

paasta_tools/utils.py
```diff
@@ -3076,20 +3078,28 @@ def read_service_instance_names(
...
+        conf_file = f"{instance_type}-{cluster}"
     config = service_configuration_lib.read_extra_service_information(
         service,
         conf_file,
         soa_dir=soa_dir,
         deepcopy=False,
     )
     config = filter_templates_from_config(config)
-    if instance_type == "tron":
+    if instance_type == "tron" or instance_type == "spark":
```
nemacysts, line 3092: nit:
```suggestion
    if instance_type in {"tron", "spark"}:
```

(or maybe even extract this set to `TRON_INSTANCE_TYPES`?)

paasta_tools/utils.py
```diff
@@ -3076,20 +3078,28 @@ def read_service_instance_names(
...
-            action_names = list(job.get("actions", {}).keys())
-            for name in action_names:
-                instance = f"{job_name}.{name}"
-                instance_list.append((service, instance))
+            for action, action_info in job.get("actions", {}).items():
+                instance = f"{job_name}.{action}"
+                if (
+                    action_info.get("executor", "") == "ssh"
+                    and instance_type == "spark"
+                ):
```
nemacysts, line 3099: maybe this should be a function called `_is_spark_action`? and perhaps we want to also check if the command is spark-related since there are some ancient tronjobs that are practically unowned and still using ssh executors for non-spark stuff

paasta_tools/utils.py
```diff
@@ -3076,20 +3078,28 @@ def read_service_instance_names(
...
+                    instance_list.append((service, instance))
+                elif instance_type == "tron":
```
nemacysts, line 3101: this could just be an `else`, right?

nemacysts (commented):
the general approach here looks fine to me (checking in with other folks on CI, though), but I'm curious as to what your thoughts would be on doing all this spark vs tron partitioning inside https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes/bin/paasta_secrets_sync.py#L132 instead?

that way, only the single place that actually needs this partitioning needs to be changed (and not a function called from a variety of places)

jfongatyelp (commented):
> doing all this spark vs tron partitioning inside https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes/bin/paasta_secrets_sync.py#L132 instead?

I'd lean towards this approach as well just because it *feels* like modifying `INSTANCE_TYPES` like this should require more support for said instance types, and we don't really do that for spark yet...they are primarily just a Tron instance type everywhere. 

That being said, I also see `INSTANCE_TYPES` and `read_service_instance_names` isn't that heavily used in a way that cares about the actual type we're looking for, outside of the secret sync (we might want to manually test paasta_tools/cli/cmds/check.py isn't affected by this change as the only other user of this list), so this shouldn't hurt too much if we want to go this route.   I'd just agree w/ Luis and please make it clear to folks reading that spark isn't a 'real' instance type (yet?)

paasta_tools/utils.py
88manpreet: I can add `else`. Reason I added `elif` to ignore other cases. Example, I don't want to add a case when instance_type is `spark` but executor is not `ssh` or the command is not `spark-submit. 
I think since the action is same both conditions I will make it a single statement.

tests/kubernetes/bin/test_paasta_secrets_sync.py
```diff
@@ -104,7 +104,7 @@ def test_sync_shared():
             pass
 
 
-@pytest.mark.parametrize("namespace", [None, "tron"])
+@pytest.mark.parametrize("namespace", [None, "tron", "spark"])
```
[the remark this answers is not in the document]
nemacysts: oh, i forgot that we weren't using the automatic mapping - that could actually make this somewhat simpler/have a smaller blast radius: we could just add a script (or modify paasta_list_tron_namespaces to take another parameter) that returns all the tron namespaces for spark jobs and then let the existing code do the rest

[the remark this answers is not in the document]
88manpreet: The premise of special casing in `utils.py` is parsing the config files to segregate non-spark vs spark tron jobs. 
I am not sure who can I do so, without either modifying the utils.py (reading file logic) or importing part of the code here.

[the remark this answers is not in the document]
88manpreet: I see. 
It seems to me `paasta_list_tron_namespaces` is actually returning all list of services having tron file. 
The idea is to modify the list_tron_namespaces to list services with for `paasta-spark` namespace. And use it as an input for the other script with namespace `paasta-spark`.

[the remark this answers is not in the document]
88manpreet: Modified. 
Why is it called `paasta_list_tron_namespaces`? It seem to be listing services which has tron jobs?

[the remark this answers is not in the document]
88manpreet: Got it clarified offline.

paasta_tools/tron_tools.py
```diff
@@ -1169,15 +1169,48 @@ def validate_complete_config(
     return []
 
 
-def get_tron_namespaces(cluster, soa_dir):
+def _is_action_spark(action_info):
```
nemacysts, line 1172: nit: we should be typing new functions
```suggestion
def _is_action_spark(action_info: InstanceConfigDict) -> bool:
```

paasta_tools/tron_tools.py
```diff
@@ -1169,15 +1169,48 @@ def validate_complete_config(
...
+    if action_info.get("executor", "") == "ssh" and "spark-submit" in action_info.get(
+        "command", ""
+    ):
+        return True
+    return False
+
+
+def get_tron_namespaces(cluster, soa_dir, k8s_namespace="tron"):
```
nemacysts, line 1180: nit: pretty sure this is what the current types look like (which i think the updated code will violate since the spark code will return `Set[str]`
```suggestion
def get_tron_namespaces(cluster: str, soa_dir: str, k8s_namespace="tron") -> List[str]:
```

tests/kubernetes/bin/test_paasta_secrets_sync.py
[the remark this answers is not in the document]
88manpreet: Addressed by moving logic to paasta_list_tron_namespaces.

paasta_tools/list_tron_namespaces.py
```diff
@@ -34,14 +34,24 @@ def parse_args():
         default=tron_tools.DEFAULT_SOA_DIR,
         help="Use a different soa config directory",
     )
+    parser.add_argument(
+        "-e",
+        "--executors",
+        default=["paasta"],
```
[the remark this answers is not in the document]
88manpreet: I am ok either way. Let me know.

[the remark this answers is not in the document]
jfongatyelp: I think just modifyign th edefault to match status quo (including all executor types) is the easiest route.

You'd have to work on timing both the paasta upgrade + puppet change at the same time on tron servers otherwise, which can be annoying.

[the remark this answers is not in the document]
88manpreet: I think the executors are set as `paasta, spark` which should cover all executor types right?
Why do we need to match the timing? 
Rollout: 
* Rollout this change as `paasta, spark` # work as status-quo
* Puppet: change `executors` to `paasta` only # better solution
* PUppet: add cron job for `executors` for `spark`

jfongatyelp (approved):

paasta_tools/list_tron_namespaces.py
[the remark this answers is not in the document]
88manpreet: I changed the default to 'paasta, spark' in my previous commit.
Seems like we are on same page :) 
Resolving this comment unless any other changes are needed.

paasta_tools/list_tron_namespaces.py
```diff
@@ -34,14 +34,24 @@ def parse_args():
         default=tron_tools.DEFAULT_SOA_DIR,
         help="Use a different soa config directory",
     )
+    parser.add_argument(
+        "-e",
+        "--executors",
+        default=list(tron_tools.EXECUTOR_TYPE_TO_NAMESPACE.keys()),
```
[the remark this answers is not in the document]
jfongatyelp: Yes, `list_tron_namespace` would have returned those that only had `executor: ssh` that may not be spark jobs previously, so we shouldn't disturb that status quo imho.

It might be true this argument has no practical effect (i.e. there are no real `executor: ssh` jobs right now that *aren't* `spark-run` commands), but unless we also enforce that (no ssh jobs except spark jobs) going forward we should probably not exclude them from being listed by the `list_tron_namespaces` command.

paasta_tools/tron_tools.py
```diff
@@ -1169,15 +1170,59 @@ def validate_complete_config(
...
+                    spark_namespaces.add(service)
+                    break
+                elif "paasta" in tron_executors and _is_action_regular_tron(
+                    action_info
+                ):
+                    spark_namespaces.add(service)
+                    break
+            # No need to iterate other jobs if at least one job or action qualified the service
+            if service in spark_namespaces:
+                break
```
nemacysts, line 1224: it might be a little cleaner to do something like iterate over all configs and create a dict of `{executor_type: [list/set, of, services, ...]}` and then just concatenate the requested ones for the return value

e.g.,
```
services_by_executor_type = defaultdict(set)
for service in namespaces:
        config = filter_templates_from_config(
            read_extra_service_information(
                service,
                extra_info=f"tron-{cluster}",
                soa_dir=soa_dir,
                deepcopy=False,
            )
        )

        for job_name, job in config.items():
            for action, action_info in job.get("actions", {}).items():
               
 services_by_executor_type[get_executor_type(action_info].add(service)

retval = set()
for executor_type in executor_types:
    retval |= services_by_executor_type[executor_type]
return retval
```

or something like that :p

paasta_tools/tron_tools.py
```diff
@@ -1169,15 +1170,70 @@ def validate_complete_config(
...
+                    break
+                elif "ssh_not_spark" in tron_executors and _is_action_ssh_not_spark(
+                    action_info
+                ):
+                    spark_namespaces.add(service)
+                    break
+            # No need to iterate other jobs if at least one job or action qualified the service
+            if service in spark_namespaces:
+                break
+    return list(spark_namespaces)
```
nemacysts, line 1236: seems like my original comment got removed by github somehow:

it might be a little cleaner to do something like iterate over all configs and create a dict of {executor_type: [list/set, of, services, ...]} and then just concatenate the requested ones for the return value

e.g.,
```
services_by_executor_type = defaultdict(set)
for service in namespaces:
        config = filter_templates_from_config(
            read_extra_service_information(
                service,
                extra_info=f"tron-{cluster}",
                soa_dir=soa_dir,
                deepcopy=False,
            )
        )

        for job_name, job in config.items():
            for action, action_info in job.get("actions", {}).items():
               
 services_by_executor_type[get_executor_type(action_info].add(service)

retval = set()
for executor_type in executor_types:
    retval |= services_by_executor_type[executor_type]
return retval
```
or something like that :p

nemacysts: where `get_executor_type` is something that returns one of [paasta, spark, ssh, whatever]

paasta_tools/list_tron_namespaces.py
```diff
@@ -34,14 +34,24 @@ def parse_args():
...
+        nargs="+",
+        help="tron executor types to list namespaces for",
+        choices=list(tron_tools.EXECUTOR_TYPE_TO_NAMESPACE.keys()),
```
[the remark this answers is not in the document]
88manpreet: I have modified this to include a 3rd value.

paasta_tools/tron_tools.py
88manpreet: I have simplified the code. It looks good to me. 
If you prefer a different way of implementation, please suggest commit suggestion, I would accept the suggestion.

paasta_tools/tron_tools.py
```diff
@@ -1180,15 +1180,57 @@ def validate_complete_config(
...
+        config = filter_templates_from_config(
+            read_extra_service_information(
+                namespace,
+                extra_info=f"tron-{cluster}",
+                soa_dir=soa_dir,
+                deepcopy=False,
+            )
+        )
+        for _, job in config.items():
+            if _is_valid_namespace(job, tron_executors):
```
nemacysts, line 1230: i think that once we just go towards what we talked about in slack, this can just be something like 
```python
...
all_namespaces = [os.path.split(config_dir)[1] for config_dir in config_dirs]
filtered_namespaces = set()

for job in load_tron_service_config(...):
    for action in job.get_actions():
        if action.get_executor() in executor_types:
            filtered_namespaces.add(job.service)

return list(filtered_namespaces)
```

PR #3494, opened 2022-11-07
Added support for namespace parameter in soaconfigs - PAASTA-17791

EmanElsaban:
### Purpose of this PR

Added a parameter ``namespace`` in ``KubernetesDeploymentConfigDict`` and replaced hardcoded ``paasta`` namespace in ``kubernetes_tools.py`` file with the new namespace parameter.

### Testing

Tested manually and using unit tests

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2235,13 +2235,8 @@ def ensure_namespace(kube_client: KubeClient, namespace: str) -> None:
 def list_deployments(
     kube_client: KubeClient,
     label_selector: str = "",
-    service_config: KubernetesDeploymentConfig = None,
+    namespace: str = "paasta",
 ) -> Sequence[KubeDeployment]:
-
-    if service_config:
-        namespace = service_config.get_namespace()
-    else:
-        namespace = "paasta"
 
     deployments = kube_client.deployments.list_namespaced_deployment(
         namespace=namespace, label_selector=label_selector
@@ -2512,12 +2507,8 @@ def pod_disruption_budget_for_service_instance(
     service: str,
     instance: str,
     max_unavailable: Union[str, int],
-    service_config: KubernetesDeploymentConfig = None,
+    namespace: str = "paasta",
 ) -> V1beta1PodDisruptionBudget:
-    if service_config:
-        namespace = service_config.get_namespace()
-    else:
-        namespace = "paasta"
     return V1beta1PodDisruptionBudget(
         metadata=V1ObjectMeta(
             name=get_kubernetes_app_name(service, instance),
@@ -2538,12 +2529,8 @@ def pod_disruption_budget_for_service_instance(
 def create_pod_disruption_budget(
     kube_client: KubeClient,
     pod_disruption_budget: V1beta1PodDisruptionBudget,
-    service_config: KubernetesDeploymentConfig = None,
+    namespace: str = "paasta",
 ) -> None:
-    if service_config:
-        namespace = service_config.get_namespace()
-    else:
-        namespace = "paasta"
     return kube_client.policy.create_namespaced_pod_disruption_budget(
         namespace=namespace, body=pod_disruption_budget
     )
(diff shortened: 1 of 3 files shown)


paasta_tools/delete_kubernetes_deployments.py
```diff
@@ -73,7 +73,10 @@ def main() -> None:
     for deployment_name in deployment_names:
         try:
             log.debug(f"Deleting {deployment_name}")
-            delete_deployment(kube_client=kube_client, deployment_name=deployment_name)
+            delete_deployment(
```
nemacysts, line 76: should we undo this change to make the diff smaller? doesn't look like we actually change the code in this file

paasta_tools/kubernetes_tools.py
```diff
@@ -2226,13 +2233,21 @@ def ensure_namespace(kube_client: KubeClient, namespace: str) -> None:
 
 
 def list_deployments(
-    kube_client: KubeClient, label_selector: str = ""
+    kube_client: KubeClient,
+    label_selector: str = "",
+    service_config: KubernetesDeploymentConfig = None,
```
nemacysts, line 2238: thoughts on unconditionally passing in `service_config`? would help us ensure that we're not missing any callers + get rid of all these if-elses to get the namespace in these helpers

nemacysts: that said...this is a little unfortunate: looks like all the operators subclass `LongRunningServiceConfig` directly and not `KubernetesDeploymentConfig` - not sure if we should just go ahead and merge `LongRunningServiceConfig` and `KubernetesDeploymentConfig` (since we only run long-running services on k8s or have the operators (flink, etc) subclass `KubernetesDeploymentConfig`

paasta_tools/delete_kubernetes_deployments.py
EmanElsaban: yep, I think this one was changed by the pre-commit hooks when I tried to add an argument then deleted it. Will fix it

EvanKrall (approved):

paasta_tools/kubernetes_tools.py
EmanElsaban: thanks, I removed ``service_config: KubernetesDeploymentConfig = None`` from functions in kubernetes_tools.py and passed ``namespace: str = "paasta"`` instead. I will add ``service_config.get_namespace()`` from the calling code (SKJ) in another PR since it will fall under changing the way skj works

EvanKrall (approved):

tests/test_kubernetes_tools.py
```diff
@@ -2816,8 +2816,19 @@ def test_max_unavailable(instances, bmf):
...
 
 def test_pod_disruption_budget_for_service_instance():
+    mock_kube_deploy_config = KubernetesDeploymentConfig(
+        service="my-service",
+        instance="my-instance",
+        cluster="cluster",
+        config_dict={},
+        branch_dict=None,
+    )
+    mock_namespace = mock_kube_deploy_config.get_namespace()
```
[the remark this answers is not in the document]
EmanElsaban: ohh I think these tests will fail if I set it to ``mock_namespace="fake_namespace"`` because for example in this test:
```
def test_create_pod_disruption_budget():
    mock_client = mock.Mock()
    mock_pdr = V1beta1PodDisruptionBudget()
    mock_namespace = "fake_namespace"
    create_pod_disruption_budget(mock_client, mock_pdr, mock_namespace)
    mock_client.policy.create_namespaced_pod_disruption_budget.assert_called_with(
        namespace="paasta", body=mock_pdr
    )
```
We check to see in ``assert_called_with`` if we are passing the namespace "paasta".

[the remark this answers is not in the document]
EmanElsaban: I don't mean to change the behaviour of the unit tests in this PR. I think we can also probably get away with not passing ``mock_namespace`` to the functions that has ``namespace: str = "paasta"`` since it's an optional argument. I wanted to be more clear so I passed the mock_namespace but we can also not pass it

EvanKrall (approved):

PR #3498, opened 2022-11-10
Fix TronActionConfig config_dict typing

nemacysts:
We've started stuffing some of the stuff that's Tron-specific in InstanceConfigDict - but this is a common base TypedDict that's used by all instances

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -323,9 +323,6 @@ class InstanceConfigDict(TypedDict, total=False):
     branch: str
     iam_role: str
     iam_role_provider: str
-    # the values for this dict can be anything since it's whatever
-    # spark accepts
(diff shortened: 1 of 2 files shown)


paasta_tools/utils.py
```diff
@@ -323,9 +323,6 @@ class InstanceConfigDict(TypedDict, total=False):
     branch: str
     iam_role: str
     iam_role_provider: str
-    # the values for this dict can be anything since it's whatever
-    # spark accepts
-    spark_args: Dict[str, Any]
```
nemacysts, line 328: we can probably move even more of the stuff in this dict out (e.g., aws_ebs_volumes) into more specific ones or just remove things entirely (cfs_period_us is probably unused), but that's for another day :p

jfongatyelp (approved):

PR #3499, opened 2022-11-11
Fix additional_namespaces arg

mattmb:
This was accidentally missing the -- so being treated as postitional

### DIFF
--- a/paasta_tools/contrib/get_running_task_allocation.py
+++ b/paasta_tools/contrib/get_running_task_allocation.py
@@ -281,7 +281,7 @@ def parse_args() -> argparse.Namespace:
         default="paasta",
     )
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/get_running_task_allocation.py
```diff
@@ -281,7 +281,7 @@ def parse_args() -> argparse.Namespace:
         default="paasta",
     )
     parser.add_argument(
-        "additional_namespaces",
+        "--additional_namespaces",
```
nemacysts, line 284: nit: i think i doubly messed up here and also forgot to make this dash-separated
```suggestion
        "--additional-namespaces",
```

nemacysts (approved):

PR #3501, opened 2022-11-16
Allow us to check service mesh orphans separately from collisions

jfongatyelp:
Because we now remove terminating pods from nerve config immediately, it seems like we're running into a race condition where this check can catch us between configure-nerve updating the config and nerve removing the affected registrations from zookeeper. This normally happens very quickly (within a second or two) in dev, but I suspect might take a little longer in prod due to the # of registrations per host (unconfirmed, but the alert behavior seems to support this).

This change to the script allows us to at least configure the 2 checks separately so we can continue to alert immediately if we see collisions, but alert only if orphans stick around after some configurable period of time (i.e. alert_after). 

Currently from searching splunk, the identified orphans come up only once in the check results and then go away, so this should help the volume of alerts right now.

### DIFF
--- a/paasta_tools/contrib/check_orphans.py
+++ b/paasta_tools/contrib/check_orphans.py
@@ -35,6 +35,7 @@ class ExitCode(Enum):
     OK = 0
     ORPHANS = 1
     COLLISIONS = 2
+    UNKNOWN = 3
 
 
 def get_zk_hosts(path: str) -> List[str]:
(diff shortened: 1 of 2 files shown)


paasta_tools/contrib/check_orphans.py
```diff
@@ -262,13 +274,34 @@ def main() -> ExitCode:
...
+        default=False,
+        action="store_true",
+        help="Skip checking collisions",
+    )
+    parser.add_argument(
+        "--no-check-orphans",
+        default=False,
+        action="store_true",
+        help="Skip checking orphans",
+    )
```
jfongatyelp, line 288: With apologies for the negatives, this is so we can keep the current invocation the same w/o affecting status quo

paasta_tools/contrib/check_orphans.py
```diff
@@ -212,43 +213,54 @@ def get_instance_data(
...
+        if orphans:
+            logger.warning("{} orphans found".format(len(orphans)))
+            print(dict(orphans_by_host))
+            return ExitCode.ORPHANS
+        else:
+            exit_log.append(
+                f"No orphans found out of {len(zk_instance_data)} service registrations seen."
+            )
+
+    logger.info("\n".join(exit_log))
```
nemacysts, line 262: just curious: why are we logging these at the same time rather than just as we hit those log statements?

nemacysts (approved):

paasta_tools/contrib/check_orphans.py
jfongatyelp: oh, that's a good call, ty! I'd organized this differently previously to only log+exit once but reverted everything else.

cuza (approved):

PR #3454, opened 2022-09-02
MLCOMPUTE-465 | add validations in paasta spark-run

CaptainSame:
Validations added for:
--pool
--cluster

Manual test runs:
- Valid pool
    - `paasta spark-run --service spark --instance adhoc --pool stable_batch --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --cluster-manager kubernetes --enable-compact-bin-packing --spark-args 'spark.logConf=true spark.driver.memory=1g spark.executor.memory=1g spark.executor.instances=4' -C 'timeout 10m spark-submit file:///work/integration_tests/s3.py'`
- Invalid pool
    - `paasta spark-run --service spark --instance adhoc --pool stable-batch --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --cluster-manager kubernetes --enable-compact-bin-packing --spark-args 'spark.logConf=true spark.driver.memory=1g spark.executor.memory=1g spark.executor.instances=4' -C 'timeout 10m spark-submit file:///work/integration_tests/s3.py'`
    - Result: `Invalid --pool value. List of valid pools for cluster pnw-devc: ['batch', 'stable_batch']`
- Valid cluster
    - `paasta spark-run --service spark --instance adhoc --cluster pnw-devc --pool batch --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --cluster-manager kubernetes --enable-compact-bin-packing --spark-args 'spark.logConf=true spark.driver.memory=1g spark.executor.memory=1g spark.executor.instances=4' -C 'timeout 10m spark-submit file:///work/integration_tests/s3.py'`
- Invalid cluster
    - `paasta spark-run --service spark --instance adhoc --cluster pnw-dev --pool batch --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --cluster-manager kubernetes --enable-compact-bin-packing --spark-args 'spark.logConf=true spark.driver.memory=1g spark.executor.memory=1g spark.executor.instances=4' -C 'timeout 10m spark-submit file:///work/integration_tests/s3.py'`
[...]

PR #3502, opened 2022-11-24
Cap docker memory limit

edingroot:
### Purpose of this PR

Cap the docker container memory limit based on node's memory size to avoid container process being killed by the OS due to OOM. This is also to fix spark driver OOM problem since the docker_wrapper is also called by `paasta spark-run`.   

### Testing

Tested by unit tests and manually:  

```bash
python3 -c 'from paasta_tools import docker_wrapper; docker_wrapper.main(["docker", "run", "-it", "--memory=100g", "ubuntu", "/bin/bash"])'
cat /sys/fs/cgroup/memory/memory.limit_in_bytes
```

image

### DIFF
--- a/paasta_tools/docker_wrapper.py
+++ b/paasta_tools/docker_wrapper.py
@@ -15,11 +15,14 @@ PIN_TO_NUMA_NODE which contains the physical CPU and memory to restrict the
 container to. If the system is NUMA enabled, docker will be called with the
 arguments cpuset-cpus and cpuset-mems.
 """
+import argparse
 import logging
 import os
 import re
 import socket
 import sys
+from typing import List
+from typing import Optional
 
 
 if "PATH" not in os.environ:
@@ -166,6 +169,37 @@ def add_argument(args, argument):
     return args
 
 
+def overwrite_argument(args: List[str], keys: List[str], new_value: str) -> List[str]:
+    """Overwrite the value of a cli argument"""
+    for idx, arg in enumerate(args):
+        for key in keys:
+            if not arg.startswith(key):
+                continue
+
+            raw_val = arg[len(key) :]
+
+            # e.g. "-k 123", "--kk 123"
+            if (
+                len(raw_val) == 0
+                and idx + 1 < len(args)
+                and not args[idx + 1].startswith("-")
+            ):
+                args[idx + 1] = new_value
+                return args
+
+            # e.g. "--kk=123"
+            if key.startswith("--") and raw_val.startswith("="):
+                args[idx] = f"{key}={new_value}"
+                return args
+
+            # e.g. "-k123"
+            if re.match(r"-[A-Za-z]", key):
+                args[idx] = f"{key}{new_value}"
+                return args
+
+    raise ValueError(f"no valid argument found with key: {key}")
+
+
 def get_cpumap():
     # Return a dict containing the core numbers per physical CPU
     core = 0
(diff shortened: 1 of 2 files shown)


mattmb:
I'm not sure this will affect executions on the actual k8s nodes btw, I think this wrapper was used by mesos to run containers but isn't used by k8s?

edingroot:
> I'm not sure this will affect executions on the actual k8s nodes btw, I think this wrapper was used by mesos to run containers but isn't used by k8s?

There are some Tron batch jobs calling `paasta spark-run` to run the spark driver on batch nodes within a container. I was trying to limit the memory based on the Tron node's memory size

nemacysts:
> > I'm not sure this will affect executions on the actual k8s nodes btw, I think this wrapper was used by mesos to run containers but isn't used by k8s?
> 
> There are some Tron batch jobs calling `paasta spark-run` to run the spark driver on batch nodes in a container. I was thinking to limit the memory based on the Tron node's memory size by this change?

@edingroot the batch boxes that run spark drivers are always going to be of a single instance type: perhaps we could just have something in SystemPaastaConfig (e.g., the puppeted config that we use to power things like https://github.com/Yelp/paasta/pull/3454/files) that sets the max allowed memory per-superregion and save ourselves quite a bit of code?

that said, I just wanna set some expectations here: this is not a silver bullet and OOMKills will still be possible - things running on the batch boxes are not using a resource-aware scheduler, so we could end up running 10 spark drivers with a max memory limit of 20% of the total memory on the box and end up overcommitted on memory (which would be fine until all 10 of these example drivers actually tried to use the memory they are limited to)

paasta_tools/docker_wrapper.py
```diff
@@ -336,6 +433,8 @@ def main(argv=None):
 
     if env_args.get("PIN_TO_NUMA_NODE"):
         argv = append_cpuset_args(argv, env_args)
+    else:
+        argv = cap_memory_arg(argv)
```
nemacysts, line 437: if we do go this route: this doesn't seem like the right place to add this

iirc: this wrapper is used by the vanishingly small amount of tasks that we still run on mesos + we don't really use the `PIN_TO_NUMA_NODE` feature, so this memory cap would be applied to everything still on mesos/using this wrapper

paasta_tools/docker_wrapper.py
```diff
@@ -166,6 +169,37 @@ def add_argument(args, argument):
     return args
 
 
+def overwrite_argument(args: List[str], keys: List[str], new_value: str) -> List[str]:
```
nemacysts, line 172: could we do this overwrite before we get to the point that we're working with docker args? we should ideally just be giving the docker wrapper an already pre-processed set of resource requirements

paasta_tools/docker_wrapper.py
```diff
@@ -295,6 +355,43 @@ def append_cpuset_args(argv, env_args):
     return argv
 
 
+def get_spark_memory_limit(node_memory_mbytes: int) -> int:
+    # by default, spark adds an overhead of 10% of the executor memory, with a
+    # minimum of 384mb
+    memory_overhead: int = max(384, int(0.1 * node_memory_mbytes))
```
nemacysts, line 361: if we go with this approach, we should refactor the numbers here into named constants

nemacysts (changes requested):
we should probably do a mix of what was mentioned in https://github.com/Yelp/paasta/pull/3502#issuecomment-1329219076 + moving the memory capping logic closer to the spark code

PR #3504, opened 2022-11-29
Cap spark driver memory and disk usage

edingroot:
### Purpose of this PR

1. Cap the docker container memory limit based on node's memory size to avoid OOM. As [mentioned](https://github.com/Yelp/paasta/pull/3502#issuecomment-1329219076) by @nemacysts, this is just a preliminary safety measure. 

2. Limit container disk usage to avoid the spark driver using too much storage space, the default value can also be overwritten by spark-run arguments.  

### Testing

Tested by unit tests added.


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -188,6 +191,14 @@ def add_subparser(subparsers):
         ),
         default=None,
     )
+    list_parser.add_argument(
+        "--docker-disk-limit",
+        help=(
+            "Set docker disk limit."
+            "Note: The default disk limit will be same as the memory limit if this argument is not provided."
```
nemacysts, line 198: just checking: could we move the default any lower? tron-used batch boxes don't really have particularly large disks (i just hopped onto a random one and there's only ~160GB of disk free) and the human-used batch boxes have enough folks running spark on them that the lower we can make the default, the better

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -711,6 +726,48 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
...
+                        "spark.driver.memoryOverhead",
+                        MIN_SPARK_DRIVER_MEMORY_OVERHEAD_MB,
+                    ),
+                )
+            )
+        memory_overhead_factor = float(
+            spark_conf.get("spark.driver.memoryOverheadFactor", memory_overhead_factor)
+        )
+    except Exception as e:
+        log.error(f"error calculating spark driver memory overhead: {e}")
```
nemacysts, line 764: just fyi: `log.exception` will automatically print out the traceback after the log message :)

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -711,6 +726,48 @@ def get_spark_app_name(original_docker_cmd: Union[Any, str, List[str]]) -> str:
     return spark_app_name
 
 
+def _convert_memory_string(value: int, input_unit: str, output_unit: str) -> float:
```
nemacysts, line 729: this could probably use a more descriptive function name or a docstring - as it is right now, one needs to read the function body to see what this is doing

tests/cli/test_cmds_spark_run.py
```diff
@@ -823,6 +835,130 @@ def test_configure_and_run_docker_driver_resource_limits(
...
+            )
+
+        assert retcode == 0
+        expected_capped_memory = DEFAULT_SPARK_DRIVER_NODE_MEMORY_MB - max(
+            MIN_SPARK_DRIVER_MEMORY_OVERHEAD_MB,
+            int(
+                DEFAULT_SPARK_DRIVER_OVERHEAD_FACTOR
+                * DEFAULT_SPARK_DRIVER_NODE_MEMORY_MB
+            ),
+        )
```
nemacysts, line 937: should this be a static value (or a value coming from the test params?)? this looks like the code in `_calculate_spark_driver_memory_overhead_mb` - so it seems a little off to re-write code that we're ostensibly testing in a test

nemacysts (commented):
might also be a good idea to add tests for the helper functions that were added in this pr - but that's up to y'all :p

paasta_tools/cli/cmds/spark_run.py
edingroot: @88manpreet Could you please have a look at this?

PR #3505, opened 2022-11-30
Added support for deploying/updating services in diff namespaces using SKJ

EmanElsaban:
### Purpose of PR

This PR adds support to deploying/updating services in different namespaces other than "paasta" using ``setup_kubernetes_job.py``


tests/test_setup_kubernetes_job.py
```diff
@@ -56,40 +116,103 @@ def test_main():
...
         ]
-        mock_service_instances = ["kuruptf_m"]
-        setup_kube_deployments(
-            kube_client=mock_client,
-            service_instances=mock_service_instances,
+        mock_load_kubernetes_service_config_no_cache.side_effect = (
+            NoDeploymentsAvailable
+        )
+        ret = get_kubernetes_deployment_config(
+            service_instances_with_valid_names=mock_get_service_instances_with_valid_names,
```
EvanKrall, line 166: Should you be calling `mock_get_service_instances_with_valid_names()` instead of passing the mock object directly? I could be misinterpreting how this test is supposed to work.

tests/test_setup_kubernetes_job.py
```diff
@@ -56,40 +116,103 @@ def test_main():
...
+        )
+        assert ret == [(True, None)]
+
+        mock_load_kubernetes_service_config_no_cache.side_effect = (
+            NoConfigurationForServiceError
         )
-        assert mock_create_application_object.call_count == 0
+
+        ret = get_kubernetes_deployment_config(
+            service_instances_with_valid_names=mock_get_service_instances_with_valid_names,
```
EvanKrall, line 177: also here

tests/test_setup_kubernetes_job.py
```diff
@@ -56,40 +116,103 @@ def test_main():
...
+            instance="instance",
+            cluster="fake_cluster",
+            soa_dir="nail/blah",
+            config_dict={},
+            branch_dict=None,
+        )
+        mock_load_kubernetes_service_config_no_cache.side_effect = None
+        mock_load_kubernetes_service_config_no_cache.return_value = mock_kube_deploy
+        ret = get_kubernetes_deployment_config(
+            service_instances_with_valid_names=mock_get_service_instances_with_valid_names,
```
EvanKrall, line 194: and here

paasta_tools/setup_kubernetes_job.py
```diff
@@ -102,19 +104,50 @@ def main() -> None:
...
+            soa_dir=soa_dir,
+            metrics_interface=deploy_metrics,
+            service_instance_configs_list=service_instance_configs_list,
+        )
+    else:
+        setup_kube_succeeded = False
+    sys.exit(0 if setup_kube_succeeded and service_instances_valid else 1)
+
+
+def get_service_instances_with_valid_names(service_instances: Sequence[str]) -> list:
```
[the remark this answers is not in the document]
EmanElsaban: sorry, forgot to update it

paasta_tools/setup_kubernetes_job.py
```diff
@@ -128,34 +161,60 @@ def validate_job_name(service_instance: str) -> bool:
...
+                % (service_instance[0], service_instance[1], cluster)
+            )
+            return [(True, None)]
+        except NoConfigurationForServiceError:
+            error_msg = (
+                f"Could not read kubernetes configuration file for %s.%s in cluster %s"
+                % (service_instance[0], service_instance[1], cluster)
+            )
+            log.error(error_msg)
+            return [(False, None)]
```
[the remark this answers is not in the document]
EmanElsaban: I meant to return 1-element list. The reason I meant to do this is I thought if we should exit right away if we have one instance that have NoDeploymentsAvailable or NoConfigurationForServiceError. But now that I'm thinking about it we shouldnt ignore other instances that could be processed especially if skj runs every 30s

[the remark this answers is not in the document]
EvanKrall: Yeah I think it's probably better to process the instances that we can: the `NoDeploymentsAvailable` situation will happen every time that someone defines a new service in yelpsoa-configs until their deploy pipeline succeeds for the first time - this can take a long time, and we don't want to block deployments of all other services in the mean time.

tests/test_setup_kubernetes_job.py
EmanElsaban: In this test, I wanted to test the different return values of ``get_kubernetes_deployment_config`` function, so first I tested the case where we have ``NoDeploymentsAvailable``. I wanted to see if it will return ``[(True, None)]``. Then I tested the case where we have ``NoConfigurationForServiceError`` and checked if the return value was ``[(False, None)]``. Lastly, I wanted to test if I passed a ``KubernetesDeploymentConfig`` would my return value be ``(True, app)``. I already assumed that ``mock_get_service_instances_with_valid_names()`` is valid and focused on the different return values I can get from this `get_kubernetes_deployment_config`` function

EmanElsaban: I tested ``get_service_instances_with_valid_names`` in this test:
```
def test_get_service_instances_with_valid_names():
    mock_service_instances = ["kurupt.f_m"]
    ret = get_service_instances_with_valid_names(
        service_instances=mock_service_instances
    )

    assert ret == [("kurupt", "f_m", None, None)]
```

EmanElsaban: Please let me know if this test makes sense or if there is a better way of testing ``test_get_kubernetes_deployment_config`` and ``test_get_service_instances_with_valid_names()``

tests/test_kubernetes_tools.py
```diff
@@ -2602,6 +2604,95 @@ def test_ensure_namespace():
...
+        deployments=mock.Mock(
+            list_deployment_for_all_namespaces=mock.Mock(return_value=mock_deployments),
+            list_stateful_set_for_all_namespaces=mock.Mock(
+                return_value=mock_stateful_sets
+            ),
+        )
+    )
+
+    assert list_all_matching_deployments(kube_client=mock_client) == []
+    mock_items = [
```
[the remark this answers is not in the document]
EmanElsaban: I also use this list when setting spec:
```
    type(mock_items[0]).spec = mock.Mock(**{"replicas": replicas})
    type(mock_items[1]).spec = mock.Mock(**{"replicas": replicas})
``

File: paasta_tools/kubernetes_tools.py [hunk 7]
```diff
@@ -2595,6 +2630,13 @@ def write_annotation_for_kubernetes_service(
         )
 
 
+def list_all_matching_deployments(kube_client: KubeClient) -> Sequence[KubeDeployment]:
```
EvanKrall, line 2633: Should we call this `list_all_paasta_deployments`? It's unclear what `matching` means. Might also be worth adding a docstring.

paasta_tools/setup_kubernetes_job.py
```diff
@@ -102,19 +104,50 @@ def main() -> None:
...
         cluster=args.cluster or load_system_paasta_config().get_cluster(),
-        rate_limit=args.rate_limit,
-        metrics_interface=deploy_metrics,
+        soa_dir=soa_dir,
     )
-    sys.exit(0 if setup_kube_succeeded else 1)
+    if ((False, None) in service_instance_configs_list) or (
+        len(service_instances_with_valid_names) != len(args.service_instance_list)
+    ):
+        service_instances_valid = False
```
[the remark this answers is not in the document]
EvanKrall: Oops, I'm realizing you already have error logging in `get_kubernetes_deployment_config` which makes this log.error line (and the one on line 117) redundant. Feel free to remove these. The errors from get_kubernetes_deployment_config are a little more informative since they include the problematic service/instance.

paasta_tools/setup_kubernetes_job.py
```diff
@@ -128,34 +169,60 @@ def validate_job_name(service_instance: str) -> bool:
...
+    service_instance_configs_list: List[Tuple[bool, KubernetesDeploymentConfig]],
     rate_limit: int = 0,
     soa_dir: str = DEFAULT_SOA_DIR,
     metrics_interface: metrics_lib.BaseMetrics = metrics_lib.NoMetrics("paasta"),
 ) -> bool:
-    if service_instances:
-        existing_kube_deployments = set(list_all_deployments(kube_client))
+
+    if service_instance_configs_list:
+        existing_kube_deployments = set(list_all_matching_deployments(kube_client))
```
EvanKrall, line 213: In the future we'll probably want to specify the set of namespaces that we care about, to reduce the number of k8s API queries we need to make; I think this is fine for now though.

paasta_tools/setup_kubernetes_job.py
```diff
@@ -200,40 +260,14 @@ def setup_kube_deployments(
...
+    return (False, None) not in applications
 
 
 def create_application_object(
-    kube_client: KubeClient,
-    service: str,
-    instance: str,
     cluster: str,
     soa_dir: str,
+    service_instance_config: KubernetesDeploymentConfig = None,
```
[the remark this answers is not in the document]
EvanKrall: I think you can just remove the default value `= None`, since it looks like all the call sites always specify `service_instance_config`.

tests/test_setup_kubernetes_job.py
EvanKrall: Oh, I think I understand why I was confused before. `mock_get_service_instances_with_valid_names` is defined twice in this test -- it gets defined as a mock object by `mock.patch(...) as mock_get_service_instances_with_valid_names` on lines 161-164, and then re-defined as a list on line 169-171. The list is what you're passing to the function under test. I hadn't noticed the re-definition before and assumed you were passing a mock object which was pretending to be a function, not a list.

Do you even need to mock.patch `get_service_instances_with_valid_names` in this test? I think `get_service_instances_with_valid_names` is only called by the `main` method in setup_kubernetes_job, which this test doesn't call. The test passes if I remove it:


```
krall@dev56-uswest1adevc:~/pg/paasta (u/emanelsabban/PAASTA-17775) $ git diff
diff --git a/tests/test_setup_kubernetes_job.py b/tests/test_setup_kubernetes_job.py
index 9dfceef5d..57b7d22c0 100644
--- a/tests/test_setup_kubernetes_job.py
+++ b/tests/test_setup_kubernetes_job.py
@@ -159,9 +159,6 @@ def test_main_invalid_job_name():

 def test_get_kubernetes_deployment_config():
     with mock.patch(
-        "paasta_tools.setup_kubernetes_job.get_service_instances_with_valid_names",
-        autospec=True,
-    ) as mock_get_service_instances_with_valid_names, mock.patch(
         "paasta_tools.setup_kubernetes_job.load_kubernetes_service_config_no_cache",
         autospec=True,
     ) as mock_load_kubernetes_service_config_no_cache:
krall@dev56-uswest1adevc:~/pg/paasta (u/emanelsabban/PAASTA-17775) $ make test
...
  tests-yelpy: commands succeeded
  congratulations :)
krall@dev56-uswest1adevc:~/pg/paasta (u/emanelsabban/PAASTA-17775) $
```

EvanKrall (approved):
fix and ship

tests/test_setup_kubernetes_job.py
EmanElsaban: yes, you are right , will remove the mock object

EvanKrall (approved):

PR #3507, opened 2022-12-05
MLCOMPUTE-216 | enable DRA by default in paasta spark-run

CaptainSame:
- Enable Spark DRA by default
- Remove mass enabling through executor range logic


88manpreet (approved):
Change looks good to me.
Let's wait till Thursday BST morning time to rollout the change.

PR #3509, opened 2022-12-07
PAASTA-17798: Support secret volumes for paasta local-run

danielpops:
## Problem
secret_volumes was not plumbed through in paasta local-run

## Solution
Add the necessary code to parse the yelpsoa-configs and pull secret contents from the secret provider, write a file to disk, and mount that file in at the specified location

## Testing Done
Manually tested on 3 different workloads that use secret_volumes, for both "team-based secrets" (kubernetes) and status quo (vault).

Unit tests forthcoming in a subsequent commit.

## Caveats
* The secret_volumes feature supports "mode" operation, but I am currently ignoring that. I'm not sure if this is important or not.
* It's a bit non-ideal to bring these secrets to the local filesystem. An improvement would be to change file ownership or mode, but there are implications of that (not guaranteed what the user in the container will run as, need sudo to change file ownership and mode despite being the owner of the file). A mitigation is to clear out the folder on each run, and another kinda not so great mitigation is to use a hidden folder and set random filenames so that casual observers can't look for known filenames. Open to suggestions for how to improve this. (Note that in the status quo, wherever this would be running, those secrets likely already exist on disk for $reasons)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -754,6 +768,35 @@ def run_docker_container(
         healthcheck_only or healthcheck
     ) and healthcheck_mode is not None
 
+    for container_mount_path, secret_content in secret_volumes.items():
+        # Create local temp file in a hidden folder in cwd
+        temp_secret_folder = os.path.join(os.getcwd(), ".secret_volumes/")
+        temp_secret_filename = os.path.join(temp_secret_folder + str(uuid.uuid4()))
```
nemacysts, line 774: just checking: was this meant to be
```suggestion
        temp_secret_filename = os.path.join(temp_secret_folder, str(uuid.uuid4()))
```

paasta_tools/kubernetes_tools.py
```diff
@@ -3526,3 +3531,60 @@ def get_kubernetes_secret_env_variables(
                 kube_client, secret_name, service_name
             )
     return decrypted_secrets
+
+
+def get_kubernetes_secret_volumes(
+    kube_client: KubeClient,
+    secret_volumes_config: Dict[str, Any],
```
nemacysts, line 3538: if i'm not mistaken, this is could be more tightly typed as `paasta_tools.utils.SecretVolume`

(same in `decrypt_secret_volumes()` and `decrypt_secret()` if this does need to be updated)

paasta_tools/kubernetes_tools.py
```diff
@@ -3502,14 +3502,19 @@ def get_kubernetes_secret(
...
 
     k8s_secret_name = get_kubernetes_secret_name(service_name, secret_name)
 
     secret_data = kube_client.core.read_namespaced_secret(
         name=k8s_secret_name, namespace=namespace
     ).data[secret_name]
-    secret = base64.b64decode(secret_data).decode("utf-8")
+    secret = base64.b64decode(secret_data)
+    # String secrets (e.g. yaml config files) need to be decoded
+    # Binary secrets (e.g. TLS Keystore or binary certificate files) cannot be decoded
```
nemacysts, line 3515: oh interesting: i guess no one has tried storing binary secrets yet and that's why this hasn't popped up before?

paasta_tools/kubernetes_tools.py
```diff
@@ -3526,3 +3531,60 @@ def get_kubernetes_secret_env_variables(
                 kube_client, secret_name, service_name
             )
     return decrypted_secrets
+
+
+def get_kubernetes_secret_volumes(
+    kube_client: KubeClient,
+    secret_volumes_config: Dict[str, Any],
+    service_name: str,
+) -> Dict[str, str]:
```
nemacysts, line 3540: should this be `Dict[str, Union[str, bytes]]` to account for binary file contents?

(same in `decrypt_secret_volumes()` and `decrypt_secret()` if this does need to be updated)

paasta_tools/kubernetes_tools.py
```diff
@@ -3502,14 +3502,19 @@ def get_kubernetes_secret(
     secret_name: str,
     service_name: str,
     namespace: str = "paasta",
+    decode: bool = True,
 ) -> str:
```
nemacysts, line 3506: is this now `-> Union[str, bytes]` since we're not always decoding?

paasta_tools/secret_tools.py
```diff
@@ -170,3 +170,94 @@ def decrypt_secret_environment_variables(
...
+def decrypt_secret_volumes(
+    secret_provider_name: str,
+    secret_volumes_config: Dict[str, Any],
+    soa_dir: str,
+    service_name: str,
+    cluster_name: str,
+    secret_provider_kwargs: Dict[str, Any],
+) -> Dict[str, str]:
+    secret_volumes = {}
+    # The config might look one of two ways:
```
nemacysts, line 184: it's a little unfortunate that the kubernetes_tools and secrets_tools functions are so similar, but i'm not really sure how to dedupe them without making the code more painful to work with :p

nemacysts (commented):
nice! just have some minor typing-related comments - but looks good otherwise

i can't come up with any better ideas for handling the leftover files on disk apart from maybe wrapping this all in a try-finally block that will cleanup the tempdir once the local-run exits - so if y'all are fine with this approach, i can't see why we wouldn't be :)

danielpops:
> i can't come up with any better ideas for handling the leftover files on disk apart from maybe wrapping this all in a try-finally block that will cleanup the tempdir once the local-run exits - so if y'all are fine with this approach, i can't see why we wouldn't be :)

@nemacysts I thought about that approach, but saw that there is at least some scenario in which paasta local-run transfers ownership of the process and exits early (let me know if I'm misinterpreting the code) https://github.com/Yelp/paasta/blob/a3eb188a89b19d03ab94f7417b262dee9edb8375/paasta_tools/cli/cmds/local_run.py#L786

nemacysts:
> @nemacysts I thought about that approach, but saw that there is at least some scenario in which paasta local-run transfers ownership of the process and exits early (let me know if I'm misinterpreting the code)

welp, that's unfortunate ;_;

paasta_tools/kubernetes_tools.py
danielpops: Yeah I think we just never had this scenario, since historically most paasta secrets end up as environment variable text

paasta_tools/kubernetes_tools.py
nemacysts: i'm honestly surprised that mypy didn't catch this - https://github.com/python/typeshed/blob/main/stdlib/base64.pyi#L30 says that `base64.b64decode` returns bytes, so i'm not sure what went wrong with the types here

paasta_tools/kubernetes_tools.py
danielpops: mmm maybe we're both mistaken 😅 

```
(Pdb) instance_config.get_secret_volumes()
[{'container_path': '/who/config/', 'secret_name': 'config_yaml_file', 'default_mode': '0444', 'items': [{'key': 'config_yaml_file', 'path': 'config.yaml'}]}]
(Pdb) type(instance_config.get_secret_volumes())
<class 'list'>
```

paasta_tools/kubernetes_tools.py
danielpops: oh, whoops! there are failing mypy tests that run in github actions, they just don't run on `make test` or any other Makefile target afaict

nemacysts: oh, 🤦🏻  - i've been meaning to fix that but doing so always gets back-burnered

danielpops: Fixed the mypy type hints

paasta_tools/kubernetes_tools.py
danielpops: Fixed the mypy type hints

paasta_tools/kubernetes_tools.py
danielpops: Fixed this accordingly to `List[Dict[str, Any]]`

nemacysts: `SecretVolume` is a TypedDict (which isn't really a real type - calling `type()` on it will show `dict`) and it looks like `get_secret_volumes()` is defined at https://sourcegraph.yelpcorp.com/Yelp/paasta/-/blob/paasta_tools/utils.py?L911 and is typed as returning List[SecretVolume]

nemacysts: (so this should be typed as `List[SecretVolume]` to match `get_secret_volumes()` - we generally wanna avoid using `Any` and using `TypedDict` on top of that makes mypy more powerful + gives IDEs/language servers more info)

danielpops: Fixed by making this `Sequence[SecretVolume]`

paasta_tools/secret_tools.py
danielpops: Same. It feels like this the concept of getting a secret provider can solve this, but it will likely take quite a bit of refactoring that I am hesitant to delve into at the moment. Or maybe something like a partial function? Something like
 
```
>>> def foo(a, b, c):
...  print("foo function")
...
>>> def bar(b, c, d, e):
...  print("bar function")
...
>>> def the_logic(secret_volumes, the_function, service_name):
...  for secret_volume in secret_volumes:
...   the_function(service_name, secret_volume["secret_name"])
...
>>> from functools import partial
>>> the_foo_partial = partial(foo, "a")
>>> the_bar_partial = partial(bar, "a", "b")
>>> the_logic([{"secret_name": "the_secret"}], the_foo_partial, "the_service")
foo function
>>> the_logic([{"secret_name": "the_secret"}], the_bar_partial, "the_service")
bar function
```

I think I prefer to save the exploration of such an improvement for another day.

nemacysts: yea, the painful solution i had in mind was using partial functions - definitely fine to leave this for another day :)

paasta_tools/kubernetes_tools.py
```diff
@@ -3502,22 +3502,26 @@ def get_kubernetes_secret(
...
+        return base64.b64decode(secret_data).decode("utf-8")
+    return base64.b64decode(secret_data)
 
 
 def get_kubernetes_secret_env_variables(
     kube_client: KubeClient,
     environment: Dict[str, str],
     service_name: str,
-) -> Dict[str, str]:
+) -> Dict[str, Union[str, bytes]]:
```
nemacysts, line 3524: hmm, we're in a bit of a pickle here - i'm pretty sure env vars must be strings

i think we're always going to get strings here (`get_kubernetes_secret()` will decode by default) so we're actually fine, but I don't think there's really a way to get mypy to follow the logic and know that `get_kubernetes_secret()` will always return a string if `decode` is true

not really sure what the best path forward is here, but my current thought is:
* `Dict[str, Union[str, bytes]]` -> Dict[str, str]
* explicitly call `get_kubernetes_secret()` with `decode=True` + maybe a comment 
* cast the call to `get_kubernetes_secret()` to `str` (to get mypy to be happy with the return value typing)

paasta_tools/kubernetes_tools.py
```diff
@@ -3526,3 +3530,60 @@ def get_kubernetes_secret_env_variables(
...
+            )
+            # Index by container path => the actual secret contents, to be used downstream to create local files and mount into the container
+            secret_volumes[
+                os.path.join(
+                    secret_volume["container_path"], secret_volume["secret_name"]
+                )
+            ] = secret_contents
+        else:
+            for item in secret_volume["items"]:
+                secret_contents = get_kubernetes_secret(
```
nemacysts, line 3580: does this call also need a `decode=False`? I just noticed that we're not decoding in the secret_tools version of this function in both cases

paasta_tools/kubernetes_tools.py
danielpops: Obliging these suggestions as stated

paasta_tools/secret_tools.py
danielpops: yeah, actually I think the "correct" solution here is to add a new `SecretProvider` for kubernetes, and have secret_tools just use the dynamically dispatched instance (kubernetes or vault). I might play with that in the future, but it's going to be a bunch of fallout from that (unit tests, etc) so I'm wary to sign up for it at the moment :)

danielpops:
That feeling when writing the unit tests took 3x as long as writing the code itself 😡

tests/cli/test_cmds_local_run.py
```diff
@@ -1959,3 +1960,352 @@ def test_missing_volumes_skipped(mock_exists):
...
+    return_value=("fake_healthcheck_mode", "fake_healthcheck_uri"),
+)
+@mock.patch("paasta_tools.cli.cmds.local_run.decrypt_secret_volumes", autospec=True)
+@mock.patch("paasta_tools.cli.cmds.local_run.shutil", autospec=True)
+@mock.patch(
+    "paasta_tools.cli.cmds.local_run.open",
+    new_callable=mock.mock_open(),
+    autospec=None,
+)
+@mock.patch("os.makedirs", autospec=True)
```
nemacysts, line 1986: imo, it's a little preferable to use the context manager form of mocking - but we take the decorator approach in enough places already that this is probably fine :)

tests/cli/test_cmds_local_run.py
```diff
@@ -1959,3 +1960,352 @@ def test_missing_volumes_skipped(mock_exists):
...
+    mock_shutil,
+    mock_decrypt_secret_volumes,
+    mock_get_healthcheck_for_instance,
+    mock_get_container_id,
+    mock_run,
+    mock_execlpe,
+    mock_get_docker_run_cmd,
+    mock_pick_random_port,
+):
+    mock_pick_random_port.return_value = 666
```
nemacysts, line 1999: should this and mock_decrypt_secret_volumes have their return values set at the decorator level to be consistent with how we mocked `paasta_tools.cli.cmds.local_run._run`?

nemacysts: (same would apply in the other tests if we take any action here)

tests/test_secret_tools.py
```diff
@@ -240,3 +243,115 @@ def test_decrypt_secret_environment_for_service(mock_get_secret_provider):
...
+    }
+
+    assert mock_secret_provider.decrypt_secret_raw.call_args_list == [
+        mock.call(
+            "the_secret_name1",
+        ),
+        mock.call(
+            "the_secret_name2",
+        ),
+    ]
```
nemacysts, line 286: hmm, do we need this test? seems like this just tells us that we called some function and passed its args in the right order - not that the expected behavior occurred :)

nemacysts (approved):
i have some comments about the tests, but i don't think they're blocking :)

tests/test_secret_tools.py
danielpops: Fair, but this test was largely based on the methodology in extant test above it for testing decrypt_secret_environment. 

It exercises the code paths I added, and confirms the logic that merges the `items` names with the container path is correct. 

Edit: Oh, is your point specifically about this assert or the test as a whole? I think you're right that the function parameters are not a very useful assertion, and I can remove them.

tests/cli/test_cmds_local_run.py
danielpops: Yeah, this is unfortunate, especially since I care zero about so many of those mocks for the scope of what I actually care to test in my test. It'd be nice if we had an abstraction like https://sourcegraph.yelpcorp.com/services/actionable_alerting_service/-/blob/tests/service_test.py?L23 to DRY this code up a bit, but I am going to respectfully decline / defer that here :)

tests/test_secret_tools.py
danielpops: Applying the same changes for the similar tests in `test_kubernetes_tools`

nemacysts: oh 🤦🏻 - english is hard: i just meant the call_args_list assert

paasta_tools/cli/cmds/local_run.py
```diff
@@ -754,6 +769,28 @@ def run_docker_container(
         healthcheck_only or healthcheck
     ) and healthcheck_mode is not None
 
+    for container_mount_path, secret_content in secret_volumes.items():
+        # Create local temp file in a hidden folder in cwd
```
nemacysts, line 773:
```suggestion
```

this is no longer accurate :)

nemacysts (approved):

jfongatyelp (approved):
Thank you!

nemacysts (approved):

danielpops:
ugh, since `/nail/tmp/` doesn't exist on the job runners, the fix to default to `/nail/tmp/` causes unit tests to break and the quick fix to set TMPDIR to /tmp/ in the unit tests didn't it. I'll need to get back into the mocking mindset and fix that before shipping

paasta_tools/cli/cmds/local_run.py
```diff
@@ -754,6 +769,29 @@ def run_docker_container(
         healthcheck_only or healthcheck
     ) and healthcheck_mode is not None
 
+    for container_mount_path, secret_content in secret_volumes.items():
+        temp_secret_folder = tempfile.TemporaryDirectory(
```
piax93, line 773: When having many volumes, I suspect there is the risk of this variable being garbage collected, and consequently the directory deleted.

danielpops: huh, yeah I had thought about this, based on the [documentation](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory):
> On completion of the context or destruction of the temporary directory object, the newly created temporary directory and all its contents are removed from the filesystem.

But I didn't observe that in practice in my local testing, but I hadn't considered the "when having many volumes" part.

danielpops: @mpiano++ great call here. 

```
  secret_volumes:
  - container_path: /who/config/
    secret_name: config_yaml_file
    default_mode: '0444'
    items:
    - key: config_yaml_file
      path: config.yaml
    - key: config_yaml_file
      path: config1.yaml
    - key: config_yaml_file
      path: config2.yaml
    - key: config_yaml_file
      path: config3.yaml
    - key: config_yaml_file
      path: config4.yaml
    - key: config_yaml_file
      path: config5.yaml
    - key: config_yaml_file
      path: config6.yaml
    - key: config_yaml_file
      path: config7.yaml
    - key: config_yaml_file
      path: config8.yaml
    - key: config_yaml_file
      path: config9.yaml
    - key: config_yaml_file
      path: config10.yaml
```

ends up with 
```
nobody@dev2-uswest2adevc:/who$ for i in $(ls /who/config/); do echo $i; cat /who/config/$i; done
config10.yaml
ldap:
    username: svc-aas@ad.yelpcorp.com
    password: REDACTED
    ca_cert: /etc/ssl/certs/ca-certificates.crt
    url: ldap://auth.ad.yelpcorp.com:389
    trace: 0
    retry_max: 10
    retry_delay: 2
    debug: 0
    timeout: 3
    base_dn: OU=Yelp,DC=ad,DC=yelpcorp,DC=com
    users_ou: OU=Users,OU=Yelp,DC=ad,DC=yelpcorp,DC=com
    disabled_ou: OU=DisabledUsers,OU=Yelp,DC=ad,DC=yelpcorp,DC=com

workday:
    username: ISU-Engineering
    password: "REDACTED"
config1.yaml
cat: /who/config/config1.yaml: Is a directory
config2.yaml
cat: /who/config/config2.yaml: Is a directory
config3.yaml
cat: /who/config/config3.yaml: Is a directory
config4.yaml
cat: /who/config/config4.yaml: Is a directory
config5.yaml
cat: /who/config/config5.yaml: Is a directory
config6.yaml
cat: /who/config/config6.yaml: Is a directory
config7.yaml
cat: /who/config/config7.yaml: Is a directory
config8.yaml
cat: /who/config/config8.yaml: Is a directory
config9.yaml
cat: /who/config/config9.yaml: Is a directory
config.yaml
cat: /who/config/config.yaml: Is a directory
```

Gonna switch back to just getting a temp directory and using os.makedirs (which I originally had, but thought tempfile.TemporaryDirectory was going to bet better for some reason)

danielpops:
@nemacysts FYI I'm going to merge this and will sync with you on cutting a release

PR #3510, opened 2022-12-07
Adding Yelp docker registry credentials to kind cluster

cuza:
This PR adds the default pool tag and the Yelp docker registry credentials to the local kind nodes


nemacysts (approved):
might be nice to try to see if we can just puppet this containerd file out to devboxes and then have the kind nodes mount that so that the file is always what we expect it to be (+ save some lines of python/bash code), but this seems fine for now if that's too painful/no one else thinks this is a good idea)

PR #3512, opened 2022-12-08
COMPINFRA-1833: add cluster label and environment variable to pods and containers for PaaSTA contract compliance

gmdfalk:
Addressing these two PaaSTA Workload contract violations:

1. Label [paasta.yelp.com/cluster](http://paasta.yelp.com/cluster) is missing on all Pods
2. Environment variable PAASTA_CLUSTER is missing on all sidecar containers (but not the main containers because it's included in the user_env)

Signed-off-by: Max Falk <gfalk@yelp.com>


ilkinmammadzada (approved):

paasta_tools/kubernetes_tools.py
```diff
@@ -1602,6 +1603,7 @@ def get_kubernetes_metadata(self, git_sha: str) -> V1ObjectMeta:
                 "paasta.yelp.com/service": self.get_service(),
                 "paasta.yelp.com/instance": self.get_instance(),
                 "paasta.yelp.com/git_sha": git_sha,
+                "paasta.yelp.com/cluster": self.cluster,
```
nemacysts, line 1606: we probably wanna use `paasta_prefixed` here

nemacysts (approved):

paasta_tools/kubernetes_tools.py
gmdfalk: Will do!

nemacysts:
@gmdfalk are you planning on merging this or shall i pick this up?

paasta_tools/kubernetes_tools.py
```diff
@@ -170,6 +170,7 @@
 UWSGI_EXPORTER_POD_NAME = "uwsgi--exporter"
 SIDECAR_CONTAINER_NAMES = [HACHECK_POD_NAME, UWSGI_EXPORTER_POD_NAME]
 KUBERNETES_NAMESPACE = "paasta"
+PAASTA_WORKLOAD_OWNER = "compute_infra_platform_experience"
```
nemacysts, line 175: it'd be nice to have this come from systempaastaconfig, but it's fine as-is (i.e., we can always enhance this later)

paasta_tools/kubernetes_tools.py
```diff
@@ -1103,7 +1105,10 @@ def get_container_env(self) -> Sequence[V1EnvVar]:
             V1EnvVar(name=name, value=value)
             for name, value in self.get_env().items()
             if name
-            not in list(secret_env_vars.keys()) + list(shared_secret_env_vars.keys())
+            # We drop PAASTA_CLUSTER here and get it instead via `get_kubernetes_environment()`
+            not in list(secret_env_vars.keys())
+            + list(shared_secret_env_vars.keys())
+            + ["PAASTA_CLUSTER"]
```
nemacysts, line 1111: thoughts on filtering out this env var at https://sourcegraph.yelpcorp.com/Yelp/paasta@691c69b75faa114fc654977b9f3401fbe7462d74/-/blob/paasta_tools/kubernetes_tools.py?L1076:5&subtree=true instead of here? it should have the same effect, but in a more centralized location

paasta_tools/kubernetes_tools.py
```diff
@@ -1173,6 +1178,16 @@ def get_kubernetes_environment(self) -> List[V1EnvVar]:
                     field_ref=V1ObjectFieldSelector(field_path="spec.nodeName")
                 ),
             ),
+            V1EnvVar(
+                name="PAASTA_CLUSTER",
+                value_from=V1EnvVarSource(
+                    field_ref=V1ObjectFieldSelector(
+                        field_path="metadata.labels['"
+                        + paasta_prefixed("cluster")
+                        + "']"
```
nemacysts, line 1187: i'm assuming we're not using an f-string here to get around having to deal with a mess of escaping quotes?

gmdfalk: yeah, i feel like it's somewhat readable this way at least.

paasta_tools/kubernetes_tools.py
nemacysts: I think we can punt on this for now (even if it takes a couple months) :)

nemacysts (approved):

PR #3513, opened 2022-12-08
PAASTA-17808: Add code coverage to paasta unit test targets

danielpops:
## Problem or feature
The repo doesn't have code coverage set up, making it difficult to know which code paths are covered by unit tests

## Solution
Add coverage dependencies and call pytest with coverage for `test` and `test-yelpy` Makefile targets

## Testing done
Ran unit tests

Note: the one test change is because of this change that happens when running with coverage vs not:

```
___________________________________ TestMarathonTools.test_get_marathon_client ____________________________________

self = <tests.test_marathon_tools.TestMarathonTools object at 0x7fb7f53040d0>

    def test_get_marathon_client(self):
        fake_url = ["nothing_for_me_to_do_but_dance"]
        fake_user = "the_boogie"
        fake_passwd = "is_for_real"
        with mock.patch(
            "paasta_tools.marathon_tools.MarathonClient", autospec=True
        ) as client_patch:
            marathon_tools.get_marathon_client(fake_url, fake_user, fake_passwd)
            client_patch.assert_called_once_with(
                fake_url, fake_user, fake_passwd, timeout=30, session=mock.ANY
            )
            args, kwargs = client_patch.call_args
            assert "User-Agent" in kwargs["session"].headers
>           assert "py.test" in kwargs["session"].headers["User-Agent"]
E           AssertionError: assert 'py.test' in 'test.py 0.152.2'

tests/test_marathon_tools.py:1671: AssertionError
```

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,23 @@
+[run]
+source =
+    .
+omit =
+    .tox/*
+    .paasta/*
+    tests/*
+    setup.py
+
+[report]
+
+exclude_lines =
+    # Have to re-enable the standard pragma
+    \#\s*pragma: no cover
+
+    # Don't complain if tests don't hit defensive assertion code:
+    ^\s*raise AssertionError\b
+    ^\s*raise NotImplementedError\b
+    ^\s*return NotImplemented\b
+    ^\s*raise$
+
+    # Don't complain if non-runnable code isn't run:
+    ^if __name__ == ['"]__main__['"]:$
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,6 @@ example_cluster/paasta/docker_registry.json
 general_itests/fake_etc_paasta/clusters.json
 pip-wheel-metadata
 debian/debhelper-build-stamp
+
+# Coverage artifacts
+.coverage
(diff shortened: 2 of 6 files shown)


.coveragerc
```diff
@@ -0,0 +1,23 @@
+[run]
```
sidsatyelp, line 1: I wonder if we want to set `branch = True` here?
https://coverage.readthedocs.io/en/6.5.0/branch.html#branch

nemacysts (approved):

.coveragerc
danielpops: @sidsatyelp branch coverage is a good idea. I originally had that here but it significantly increase the amount of information that gets reported so I elected to leave it out and consider adding it in the future. Having said that, I'm happy to add branch coverage right now if you prefer!

sidsatyelp: If that is the case then we could leave it out for now. When there is too much information reported, people tend to skip it! We can add it later if we see the need for it.

sidsatyelp (approved):

.coveragerc
danielpops: For posterity, coverage report [with --branch](https://fluffy.yelpcorp.com/i/Jd2WmXCL9jp97GcKGtl17hr48pqlXjDg.html) and [without --branch](https://fluffy.yelpcorp.com/i/9qcP35c3s0q9N5M2BJnFK7S2Dlm2Ln6f.html)

PR #3519, opened 2022-12-28
Fix mock usage

cklein:
Some tests used `assert mock_object.called_once_with()`. The return value of `called_once_with()` is a mock object which always evaluates to true which means it always passes the assertion.


nemacysts (approved):

cklein:
https://github.com/python/cpython/issues/100690

PR #3522, opened 2023-01-03
Adding group_lines.py functionality to paasta

cuza:
This PR adds `group_lines.py`(file from puppet) functionality into `paasta_list_kubernetes_service_instances`

### DIFF
--- a/paasta_tools/list_kubernetes_service_instances.py
+++ b/paasta_tools/list_kubernetes_service_instances.py
@@ -29,7 +29,7 @@ Command line options:
 import argparse
 import random
 import sys
-from typing import Sequence
+from typing import List
 
 from paasta_tools import kubernetes_tools
 from paasta_tools.utils import compose_job_id
@@ -113,8 +113,8 @@ def main():
     sys.exit(0)
 
 
-def group_lines(service_instances: Sequence[str], num_lines: int) -> None:
-    output = [[] for _ in range(num_lines)]
+def group_lines(service_instances: List[str], num_lines: int) -> None:
+    output: List[List[str]] = [[] for _ in range(num_lines)]
     curr_index = 0
     for instance in service_instances:
         output[curr_index].append(instance)
--- a/paasta_tools/list_kubernetes_service_instances.py
+++ b/paasta_tools/list_kubernetes_service_instances.py
@@ -27,6 +27,7 @@ Command line options:
 - -c <CLUSTER>, --cluster <CLUSTER>: Specify which cluster of services to read
 """
 import argparse
+import random
 import sys
 
 from paasta_tools import kubernetes_tools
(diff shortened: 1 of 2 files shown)


bobokun (approved):

paasta_tools/list_kubernetes_service_instances.py
```diff
@@ -88,9 +102,25 @@ def main():
         else:
             app_name = compose_job_id(name, instance)
         service_instances.append(app_name)
-    print("\n".join(service_instances))
+    if args.shuffle:
+        random.shuffle(service_instances)
+    if args.group_lines:
+        group_lines(service_instances, args.group_lines)
+    else:
+        print("\n".join(service_instances))
```
jfongatyelp, line 110: This doesn't seem to ever get around to printing the instances if shuffle or group_lines is specified since the print is in an `else` here

jfongatyelp: jk i have now read group_lines and see we're printing down there.

paasta_tools/list_kubernetes_service_instances.py
```diff
@@ -88,9 +102,25 @@ def main():
...
     sys.exit(0)
 
 
+def group_lines(service_instances, num_lines):
```
nemacysts, line 114:
```suggestion
def group_lines(service_instances: Sequence[str], num_lines: int) -> None:
```

paasta_tools/list_kubernetes_service_instances.py
```diff
@@ -70,6 +71,19 @@ def parse_args():
         default="kubernetes",
         help="Instance type to list, default %(default)s",
     )
+    parser.add_argument(
+        "--shuffle",
+        action="store_true",
+        help=("Shuffle the instances output."),
```
nemacysts, line 77:
```suggestion
        help="Shuffle the instances output.",
```

paasta_tools/list_kubernetes_service_instances.py
```diff
@@ -70,6 +71,19 @@ def parse_args():
...
+    )
+    parser.add_argument(
+        "--group-lines",
+        type=int,
+        dest="group_lines",
+        help=(
+            "Groups instances output into a desired number of lines with each instance separated by spaces"
```
nemacysts, line 84:
```suggestion
            "Groups instances output into a desired number of lines with each instance separated by a space"
```

paasta_tools/list_kubernetes_service_instances.py
nemacysts, line 110: suggestion:

```suggestion
    if args.shuffle:
        random.shuffle(service_instances)
        
    if args.group_lines:
        group_lines(service_instances, args.group_lines)
    else:
        print("\n".join(service_instances))
```

would make it clearer that these aren't mutually exclusive - without the whitespace, it's not immediately clear that one can hit both cases (thanks to the magic of human pattern matching :p)

nemacysts (approved):
some minor comments/nits - but none are particularly blocking

PR #3525, opened 2023-01-09
Enable explicitly setting service_account_name in tron actions

nemacysts:
There's internal users that would like to configure a Service Account
with specific permissions for use in their batches.

Right now, we can only dynamically create Service Accounts - and these
are generally only used for Pod Identity (or, they're mapped 1:1 with an
existing Role).

This PR enables users to explicitly set service_account_name in their
action configs and have that be plumbed all the way through to Tron

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -856,6 +856,14 @@ def format_tron_action_dict(action_config: TronActionConfig, use_k8s: bool = Fal
         "triggered_by": action_config.get_triggered_by(),
         "on_upstream_rerun": action_config.get_on_upstream_rerun(),
         "trigger_timeout": action_config.get_trigger_timeout(),
+        # outside of Spark usescases, we also allow users to specify an expected-to-exist Service Account name
+        # in the Tron namespace in case an action needs specific k8s permissions (e.g., a Jolt batch may need
+        # k8s permissions to list Jolt pods in the jolt namespace to do science™ to them).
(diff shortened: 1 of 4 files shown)


paasta_tools/tron_tools.py
```diff
@@ -856,6 +856,14 @@ def format_tron_action_dict(action_config: TronActionConfig, use_k8s: bool = Fal
         "triggered_by": action_config.get_triggered_by(),
         "on_upstream_rerun": action_config.get_on_upstream_rerun(),
         "trigger_timeout": action_config.get_trigger_timeout(),
+        # outside of Spark usescases, we also allow users to specify an expected-to-exist Service Account name
+        # in the Tron namespace in case an action needs specific k8s permissions (e.g., a Jolt batch may need
+        # k8s permissions to list Jolt pods in the jolt namespace to do science™ to them).
+        # if the provided Service Account does not exist, Tron should simply fail to create the Podspec and report
+        # a failure
+        # TODO: verify the above sentence
```
nemacysts, line 864:
```suggestion
```

jfongatyelp (approved):

nemacysts:
tested this on infrastage - everything seems to work fine. the only downside right now is that using an invalid service account name is only visible in the tron master logs (the action just goes into a failed state with no logs visible in tronweb)

this is probably fine since this is very much a power-user feature and i don't think we'll really run into this much

KaspariK (approved):

jfongatyelp (approved):

PR #3529, opened 2023-01-12
Enable setting a multiplier to uwsgi offset config

nemacysts:
This will allow us to slowly lower the impact of the offset parameter by decreasing this value from 1.0 -> 0 rather than completely removing it all at once (and potentially exposing problems hidden by our overscaling)

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -1995,6 +1995,7 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     spark_blockmanager_port: int
     skip_cpu_burst_validation: List[str]
     tron_default_pool_override: str
+    uwsgi_offset_multiplier: float
 
 
 def load_system_paasta_config(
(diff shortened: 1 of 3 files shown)


paasta_tools/utils.py
```diff
@@ -2734,6 +2734,15 @@ def get_tron_k8s_cluster_overrides(self) -> Dict[str, str]:
...
         return self.config_dict.get("tron_k8s_cluster_overrides", {})
 
+    def get_uwsgi_offset_multiplier(self) -> float:
+        """
+        Temporary configuration to allow us to slowly deprecate the usage of `offset` in uwsgi-based autoscaling
+        configurations without making a single massive change to how usage is calculated.
+
+        To be removed once PAASTA-17840 is done.
+        """
+        return self.config_dict.get("uwsgi_offset_multiplier", 1.0)
```
EvanKrall, line 2745: probably need to add this attribute to `SystemPaastaConfigDict`

EvanKrall (approved):

paasta_tools/utils.py
nemacysts: yup - just did that after running mypy locally (and between you loading this PR and reviewing it :p)

mtbentley (approved):
:ship:, provided the tests pass

PR #3530, opened 2023-01-17
removing hardcoded paasta ns in setup_prometheus_adapter_config file - PAASTA-17777

EmanElsaban:
### Feature

In this PR we remove hardcoded "paasta" namespace in ``setup_prometheus_adapter_config`` file and replace it with the namespace from soaconfigs. We also ensure that we call ``ensure_namespace`` for all namespaces not just "paasta"

### Testing

- Unit tests pass
- manually tested it on paasta playground and I see the adapter-config map created:
```
(py37-linux) emanelsabban@dev55-uswest1adevc:~/pg/paasta$ KUBECONFIG="./k8s_itests/kubeconfig" kubectl get configmap -n custom-metrics
NAME               DATA   AGE
adapter-config     1      22s
```
Describing configmap: https://fluffy.yelpcorp.com/i/hLq5wHQ8rXS4G0PM0mKzh0trzNr9Mh6k.html

### DIFF
--- a/paasta_tools/setup_prometheus_adapter_config.py
+++ b/paasta_tools/setup_prometheus_adapter_config.py
@@ -217,6 +217,7 @@ def create_instance_uwsgi_scaling_rule(
     instance: str,
     autoscaling_config: AutoscalingParamsDict,
     paasta_cluster: str,
+    namespace: str = "paasta",
 ) -> PrometheusAdapterRule:
     """
     Creates a Prometheus adapter rule config for a given service instance.
@@ -311,6 +312,7 @@ def create_instance_piscina_scaling_rule(
     instance: str,
     autoscaling_config: AutoscalingParamsDict,
     paasta_cluster: str,
+    namespace: str = "paasta",
 ) -> PrometheusAdapterRule:
     """
     Creates a Prometheus adapter rule config for a given service instance.
@@ -413,6 +415,7 @@ def create_instance_cpu_scaling_rule(
     instance: str,
     autoscaling_config: AutoscalingParamsDict,
     paasta_cluster: str,
+    namespace: str = "paasta",
 ) -> PrometheusAdapterRule:
     """
     Creates a Prometheus adapter rule config for a given service instance.
@@ -432,7 +435,7 @@ def create_instance_cpu_scaling_rule(
         kube_deployment_labels{{
             deployment='{deployment_name}',
             paasta_cluster='{paasta_cluster}',
-            namespace='paasta'
+            namespace='{namespace}'
         }}
     """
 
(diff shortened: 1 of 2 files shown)


paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -696,12 +702,15 @@ def create_prometheus_adapter_config(
             cluster=paasta_cluster,
             instance_type_class=KubernetesDeploymentConfig,
         ):
+            if not is_dry_run:
+                ensure_namespace(kube_client, namespace=instance_config.get_namespace())
```
nemacysts, line 706: out of curiosity: how are we planning on generally creating these namespaces?

i wonder if we should just assume that these namespaces will exist and let this script not worry about them - especially since this script doesn't actually require the instance namespace to exist

EmanElsaban: I thought of wherever we are calling ``ensure_namespace`` just make sure that we are calling it for all namespaces needed instead of just "paasta" namespace. So currently, we call ``ensure_namespace`` function in these files:
- setup_kubernetes_job : we took care of it in skj code by calling it for all service_instances
- setup_prometheus_adapter_config: this PR takes care of this
- setup_kubernetes_cr: we don't hardcode the namespaces here
- delete_kubernetes_deployments.py: we don't call this file anywhere
- setup_istio_mesh: I think this file is just used for testing in kubestage
I thought of this with the assumption that we already call ``ensure_namespace`` everywhere we need to have "paasta" namespace created. 
This way the service developer would need to just specify the namespace in yelp-soaconfigs and each script in paasta_tools would take care of creating the namespace if it doesn't exist. Not sure if this is a good idea or if there is a better way of approaching this

EvanKrall (approved):

paasta_tools/setup_prometheus_adapter_config.py
nemacysts: i think the `ensure_namespace(..., "paasta")` was a bad copy-paste on my part when i was first writing this script 😆 (we should really only need to ensure that the `custom-metrics` namespace exists for this script to work)

if the new per-service namespaces are generally going to be created elsewhere, then i think it's fine to not attempt to create them here (since there should be no impact if we generate an autoscaling rule for something that doesn't exist yet)

EmanElsaban: The new per-service namespaces should be created by skj since we call ``ensure_namespace`` there and since it runs every 30s then it should pick up fast if a namespace doesn't exist and create it. There could be a small gap in time between running skj and setup_prometheus_adapter file but I guess this gap is fine

EvanKrall (approved):

PR #3531, opened 2023-01-17
Adding zookeeper to paasta playground

cuza:
This PR adds a local docker zookeeper instance to the paasta playground

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,12 +136,19 @@ vscode_settings: .paasta/bin/activate .tox/py37-linux
 
 .PHONY: generate_paasta_playground
 generate_paasta_playground: .paasta/bin/activate .tox/py37-linux
-	.tox/py37-linux/bin/python  paasta_tools/contrib/create_paasta_playground.py
+	.tox/py37-linux/bin/python paasta_tools/contrib/create_paasta_playground.py
 	PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_playground/ .tox/py37-linux/bin/python -m paasta_tools.generate_deployments_for_service -d ./soa_config_playground -s compute-infra-test-service -v
 
 .PHONY: playground-api
 playground-api: .tox/py37-linux generate_paasta_playground
 	.paasta/bin/tox -i $(PIP_INDEX_URL) -e playground-api
+
+.PHONY: setup_kubernetes_job
+setup-kubernetes-job: k8s_fake_cluster generate_paasta_playground 
+	export KUBECONFIG="./k8s_itests/kubeconfig";\
+	export PAASTA_SYSTEM_CONFIG_DIR="./etc_paasta_playground/";\
+	export PAASTA_TEST_CLUSTER=kind-${USER}-k8s-test;\
+	.tox/py37-linux/bin/python -m paasta_tools.list_kubernetes_service_instances -d ./soa_config_playground --shuffle --group-lines 1 | xargs --no-run-if-empty .tox/py37-linux/bin/python -m paasta_tools.setup_kubernetes_job -d ./soa_config_playground -c kind-${USER}-k8s-test
 
 .PHONY: clean-playground
 clean-playground:
--- a/k8s_itests/Makefile
+++ b/k8s_itests/Makefile
@@ -21,18 +21,24 @@ endif
 
 all: clean itest
 
-fake_cluster: clean create_cluster
+fake_cluster: create_cluster
+ifeq (,$(wildcard ./.fake_cluster))
 	if [ "$(PAASTA_K8S_ENV)" = "YELP" ]; then \
 		./scripts/set-paasta-registry-credentials.sh $(CLUSTER); \
 	fi
+	touch .fake_cluster
+endif
 
 kind:
+ifeq (,$(wildcard ./kind))
 	./scripts/install-kind.sh
-
+endif
 
 create_cluster: kind
+ifeq (,$(wildcard ./.create_cluster))
 	./kind create cluster --name=$(CLUSTER) --config=$(KIND_CONFIG) --kubeconfig $(KUBECONFIG)
 	touch .create_cluster
+endif
 
 itest: create_cluster
 	../.paasta/bin/tox -e k8s_itests
--- a/paasta_tools/contrib/create_paasta_playground.py
+++ b/paasta_tools/contrib/create_paasta_playground.py
@@ -1,6 +1,9 @@
 import json
 import os
+import socket
 
+from paasta_tools.utils import get_docker_client
+from paasta_tools.cli.utils import pick_random_port
 from render_template import render_values
 
 
@@ -8,6 +11,13 @@ def main():
 
     config_path = "etc_paasta_playground"
     values_path = "./k8s_itests/deployments/paasta/values.yaml"
+    user = os.getenv('USER')
+    
+    # start a local copy of zookeeper on a random port
+    zookeeper_port = pick_random_port(f"{user}-paasta-zookeeper")
+    run_local_zookeeper(user, zookeeper_port)
+    os.environ['ZOOKEEPER_PORT'] = str(zookeeper_port) 
+    os.environ['HOST_IP'] = socket.gethostbyname(socket.gethostname())
 
     # create an etc_paasta_playground directory if it doesn't exist
     # and copy fake_etc_paasta content into etc_paasta_playground directory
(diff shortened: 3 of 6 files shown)


paasta_tools/contrib/create_paasta_playground.py
```diff
@@ -38,5 +49,29 @@ def main():
...
+    )
+    if zookeeper_container is None:
+        create_zookeeper_container(zookeeper_port, user)
+    elif zookeeper_container.get("Status") != "running":
+        client.remove_container(zookeeper_container, force=True)
+        create_zookeeper_container(zookeeper_port, user)
+
+
+def create_zookeeper_container(port, user):
+    os.system(
```
nemacysts, line 71: also, we should probably be a little more consistent and use subprocess.run here rather than os.system

k8s_itests/Makefile
```diff
@@ -21,18 +21,24 @@ endif
 
 all: clean itest
 
-fake_cluster: clean create_cluster
+fake_cluster: create_cluster
+ifeq (,$(wildcard ./.fake_cluster))
```
nemacysts, line 25: i *think* we might be able to just native make targets to do this (e.g., have a .fake_cluster target that does the work in this conditional - but we might want someone better at make to chime in here (e.g., @EvanKrall)

Makefile
```diff
@@ -136,13 +136,20 @@ vscode_settings: .paasta/bin/activate .tox/py37-linux
...
 
 .PHONY: playground-api
 playground-api: .tox/py37-linux generate_paasta_playground
 	.paasta/bin/tox -i $(PIP_INDEX_URL) -e playground-api
 
+.PHONY: setup_kubernetes_job
+setup-kubernetes-job: k8s_fake_cluster generate_paasta_playground
+	export KUBECONFIG="./k8s_itests/kubeconfig";\
+	export PAASTA_SYSTEM_CONFIG_DIR="./etc_paasta_playground/";\
+	export PAASTA_TEST_CLUSTER=kind-${USER}-k8s-test;\
```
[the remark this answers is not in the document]
cuza: it can't be done in multiple lines since each line is a subshell

paasta_tools/contrib/create_paasta_playground.py
```diff
@@ -38,5 +49,29 @@ def main():
...
+        f'docker run -d -p {port}:2181 -e "ALLOW_ANONYMOUS_LOGIN=yes" --name {user}-paasta-zookeeper zookeeper:3.5'
```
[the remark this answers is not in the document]
cuza: hmm good question

Makefile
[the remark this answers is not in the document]
nemacysts: :rip: - in that case, we probably either want to just be more consistent with quoting (or use single quotes where we don't need interpolation and double where we do

e.g., 
```suggestion
	export KUBECONFIG='./k8s_itests/kubeconfig'; \
	export PAASTA_SYSTEM_CONFIG_DIR='./etc_paasta_playground/'; \
	export PAASTA_TEST_CLUSTER="kind-${USER}-k8s-test"; \
```

or ```suggestion
	export KUBECONFIG=./k8s_itests/kubeconfig; \
	export PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_playground/; \
	export PAASTA_TEST_CLUSTER=kind-${USER}-k8s-test; \
```

File: Makefile [hunk 5]
```diff
@@ -134,15 +134,35 @@ swagger-validate:
...
+etc_paasta_playground soa_config_playground: .paasta/bin/activate .tox/py37-linux
+	.tox/py37-linux/bin/python paasta_tools/contrib/create_paasta_playground.py
+
+.PHONY: generate_deployments_for_service
+generate_deployments_for_service: | soa_config_playground .tox/py37-linux
+	export KUBECONFIG=./k8s_itests/kubeconfig;\
+	export PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_playground/;\
+	export PAASTA_TEST_CLUSTER=kind-${USER}-k8s-test;\
+	.tox/py37-linux/bin/python -m paasta_tools.list_kubernetes_service_instances -d ./soa_config_playground | cut -f1 -d"." | uniq | shuf | xargs -n 1 --no-run-if-empty \
+	.tox/py37-linux/bin/python -m paasta_tools.generate_deployments_for_service -d ./soa_config_playground -v -s
```
nemacysts, line 146: just curious: is this what we do internally? the `cut` seems a bit out of place (i.e., i get the feeling we'd have changed the output or added a flag to obiviate the need for this in our internal crons)

Makefile
```diff
@@ -134,15 +134,35 @@ swagger-validate:
...
 
 .PHONY: playground-api
-playground-api: .tox/py37-linux generate_paasta_playground
+playground-api: .tox/py37-linux | soa_config_playground
```
nemacysts, line 149: TIL about order-only prerequisites

Makefile
nemacysts: although, we do a `cut` internally for secret syncing, so maybe we do this internall for deployments as well...

docs/source/paasta_development.rst
```diff
@@ -61,7 +61,7 @@ Debugging PaaSTA playground components (in VS Code)
 
 Steps below outline running PaaSTA playground components with a debugger attached:
 
-1. Run ``make generate_paasta_playground`` to run the Make target that generates the ``etc_paasta_playground`` and ``soa_config_playground`` directories and ``deployments.json`` file.
+1. Run ``make generate_deployments_for_service`` to run the Make targets that generates the ``etc_paasta_playground`` and ``soa_config_playground`` directories and ``deployments.json`` files.
```
nemacysts, line 64: feel free to ignore: ```suggestion
1. Run ``make generate_deployments_for_service`` to run the Make targets that generate the ``etc_paasta_playground`` and ``soa_config_playground`` directories and the ``deployments.json`` file.
```

File: k8s_itests/scripts/setup.sh [hunk 8]
```diff
@@ -7,3 +7,9 @@ mkdir $PAASTA_CONFIG_DIR
 
 echo "Applying kubernetes resources"
 kubectl apply -Rf $KUBE_RESOURCE_DIR --context $KIND_CLUSTER
+
+
+
+ZOOKEEPER_PORT=$(shuf -i 2000-65000 -n 1)
```
nemacysts, line 13: hmm, this is tricky: I think we have some helpers internally for picking an ephemeral port so that we don't collide with any of the other zillion things that need a port on our boxes - not sure if that's been open-sourced or is small enough to vendor some of it in this repo

nemacysts: oh, we do have this available in this repo: [pick_random_port](https://github.com/Yelp/paasta/blob/2e1cafdc687f231eca07388f468bf4b00ebb95cd/paasta_tools/cli/utils.py#L1032) is what we likely want to do here

paasta_tools/contrib/create_paasta_playground.py
```diff
@@ -38,5 +50,56 @@ def main():
     )
 
 
+def run_local_zookeeper(user, zookeeper_port):
```
nemacysts, line 53:
```suggestion
def run_local_zookeeper(user: str, zookeeper_port: int) -> None:
```

paasta_tools/contrib/create_paasta_playground.py
```diff
@@ -38,5 +50,56 @@ def main():
...
+        None,
+    )
+    if zookeeper_container is None:
+        create_zookeeper_container(zookeeper_port, user)
+    elif zookeeper_container.get("Status") != "running":
+        client.remove_container(zookeeper_container, force=True)
+        create_zookeeper_container(zookeeper_port, user)
+
+
+def create_zookeeper_container(port, user):
```
nemacysts, line 71:
```suggestion
def create_zookeeper_container(port: int, user: str) -> None:
```

nemacysts: additionally, perhaps this should be `run_zookeeper_container`?

nemacysts: finally (and feel free to drop since i don't think anyone else would care): would be nice for consistency if the parameter ordering matched the `run_local_zookeeper` above (it's the same params in a different order :p)

paasta_tools/contrib/create_paasta_playground.py
nemacysts: additionally: maybe this could be `ensure_running_local_zookeeper`) so that we don't have a `create_` function that actually runs something :p

nemacysts (commented):
https://github.com/Yelp/paasta/pull/3531/files#diff-953c0435e333e4b50859f05713a9f741b79988eee8d2d5fecf6c27a00124d5b9R13 is my only "real" comment - feel free to ignore the rest

Makefile
cuza: this is what we do Internally copied/pasted here

Makefile
cuza: same

k8s_itests/scripts/setup.sh
cuza: this was committed by mistake 😭😭😭😭

Makefile
cuza: the cut is for getting only services since the output will have service.instance

nemacysts: we should probably add a new CLI at some point so that we can do this in a less funky way internally and here :p

k8s_itests/scripts/setup.sh
cuza: This is also what broke the test 😭😭😭😭😭

nemacysts (approved):

PR #3532, opened 2023-01-17
Add support for aws assume role in local-run

Qmando:
Seeking feedback on this proposed change to add more AWS functionality to local-run. The goal of this is to allow local-run to run within an AWS role context, to emulate what happens in a normal paasta context when pod identity (or kiam) is set.

This supports several scenarios:
- The service has an iam role configured, and you want to assume that role
- You want to specify a different role to assume
- You want to use whatever role aws-okta provides

Unless one of these new arguments is specified, behavior will be identical to before. 

If `--assume-pod-identity` is provided, it will check the relevant instance config for `iam-role`, and attempt to assume that role by using the aws-okta for an account matching the current runtimeenv. 

if `--use-okta-role` is provided, it will attempt to get a session from aws-okta and use those credentials.

if `--aws-role` is provided, it will attempt to assume into that role rather than one from the instance config.

Looking to get some feedback on this before I finish writing tests. 

Manual test / example behavior:
Attempt to use actual iam role (I don't have permissions) https://fluffy.yelpcorp.com/i/R6SrxqhtrXf8WXB8p5hZhLCv9kZMB0hK.html
Attempt to use my current role
https://fluffy.yelpcorp.com/i/sqZxqfhW4WpRJSj1vzCLwZFppN8wGhD4.html

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -774,7 +774,7 @@ def assume_aws_role(
         # The contents of "Credentials" key from assume_role is the same as from aws-okta
         cmd_output = assumed_role["Credentials"]
 
-    creds_dict = {
+    creds_dict: AWSSessionCreds = {
         "AWS_ACCESS_KEY_ID": cmd_output["AccessKeyId"],
         "AWS_SECRET_ACCESS_KEY": cmd_output["SecretAccessKey"],
         "AWS_SESSION_TOKEN": cmd_output["SessionToken"],
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -29,6 +29,7 @@ from urllib.parse import urlparse
 
 import requests
 from docker import errors
+from mypy_extensions import TypedDict
 
 from paasta_tools.adhoc_tools import get_default_interactive_config
 from paasta_tools.cli.cmds.check import makefile_responds_to
@@ -53,6 +54,7 @@ from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import get_docker_client
 from paasta_tools.utils import get_possible_launched_by_user_variable_from_env
 from paasta_tools.utils import get_username
+from paasta_tools.utils import InstanceConfig
 from paasta_tools.utils import is_secrets_for_teams_enabled
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import list_services
@@ -67,6 +69,12 @@ from paasta_tools.utils import timed_flock
 from paasta_tools.utils import Timeout
 from paasta_tools.utils import TimeoutError
 from paasta_tools.utils import validate_service_instance
+
+
+class AWSSessionCreds(TypedDict):
+    AWS_ACCESS_KEY_ID: str
+    AWS_SECRET_ACCESS_KEY: str
+    AWS_SESSION_TOKEN: str
 
 
 def parse_date(date_string):
@@ -465,7 +473,10 @@ def add_subparser(subparsers):
     )
     list_parser.add_argument(
         "--assume-role",
-        help="role ARN to assume before launching the service",
+        help=(
+            "role ARN to assume before launching the service. "
+            "Example format: arn:aws:iam::01234567890:role/rolename"
+        ),
         type=str,
         dest="assume_role",
         required=False,
@@ -679,15 +690,19 @@ def check_if_port_free(port):
 
 
 def assume_aws_role(
-    instance_config, service, assume_role, assume_pod_identity, aws_profile
-):
+    instance_config: InstanceConfig,
+    service: str,
+    assume_role: str,
+    assume_pod_identity: bool,
+    aws_profile: str,
+) -> AWSSessionCreds:
     """Runs AWS cli to assume into the correct role, then extract and return the ENV variables from that session"""
     pod_identity = instance_config.get_iam_role()
     if assume_role:
         pod_identity = assume_role
     if assume_pod_identity and not pod_identity:
         print(
-            "Error: --assume-pod-identity passed but no pod identity was found for this instance",
+            f"Error: --assume-pod-identity passed but no pod identity was found for this instance ({instance_config.instance})",
             file=sys.stderr,
         )
         sys.exit(1)
@@ -696,7 +711,7 @@ def assume_aws_role(
         with open("/nail/etc/ecosystem") as ecosystem_file:
             default_profile = ecosystem_file.read()
     profile = aws_profile or default_profile
-    if pod_identity and assume_pod_identity:
+    if pod_identity and (assume_pod_identity or assume_role):
         print(
             "Attempting to assume role {} using profile {}".format(
                 pod_identity, profile
@@ -713,43 +728,69 @@ def assume_aws_role(
                 get_username(),
             )
         else:
-            assume_role_cmd = "aws --profile {} sts assume-role --role-arn {} --role-session-name {}-{}"
-            assume_role_cmd = assume_role_cmd.format(
-                profile, pod_identity, service, get_username()
-            )
-        cmd = subprocess.run(assume_role_cmd, shell=True, stdout=subprocess.PIPE)
+            assume_role_cmd = [
+                "aws",
+                "--profile",
+                profile,
+                "sts",
+                "assume-role",
+                "--role-arn",
+                pod_identity,
+                "--role-session-name",
+                f"{service}-{get_username()}",
+            ]
+        cmd = subprocess.run(assume_role_cmd, stdout=subprocess.PIPE)
         if cmd.returncode != 0:
             print(
                 "Error assuming pod identity role. Remove --assume-pod-identity to run without pod identity role, or specify another role via --assume-role",
                 file=sys.stderr,
             )
             sys.exit(1)
+        cmd_output = json.loads(cmd.stdout.decode("utf-8"))
     elif aws_profile:
         print(f"Getting credentials for profile {aws_profile}")
         if os.getuid() == 0:
-            get_cred_process = "sudo -u {} HOME=/nail/home/{} aws configure get credential_process --profile {}"
-            get_cred_process = get_cred_process.format(
-                get_username(), get_username(), aws_profile
-            )
+            get_cred_process = [
+                "sudo",
+                "-u",
+                get_username(),
+                f"HOME=/nail/home/{get_username()}",
+                "aws",
+                "configure",
+                "get",
+                "credential_process",
+                "--profile",
+                aws_profile,
+            ]
         else:
-            get_cred_process = "aws configure get credential_process --profile {}"
-            get_cred_process = get_cred_process.format(aws_profile)
-        get_cred_process_cmd = subprocess.run(
-            get_cred_process, shell=True, stdout=subprocess.PIPE
-        )
+            get_cred_process = [
+                "aws",
+                "configure",
+                "get",
+                "credential_process",
+                "--profile",
+                aws_profile,
+            ]
+        get_cred_process_cmd = subprocess.run(get_cred_process, stdout=subprocess.PIPE)
         if get_cred_process_cmd.returncode != 0:
             print(
                 f"Error: no credential process found for profile {aws_profile}",
                 file=sys.stderr,
             )
             sys.exit(1)
-        cred_process = get_cred_process_cmd.stdout.decode("utf-8")
-        cred_process = "sudo -u {} HOME=/nail/home/{} " + cred_process
-        cred_process = cred_process.format(get_username(), get_username())
-        cred_process_cmd = subprocess.run(
-            cred_process, shell=True, stdout=subprocess.PIPE
-        )
+        cred_process = get_cred_process_cmd.stdout.decode("utf-8").split(" ")
+        if os.getuid() == 0:
+            cred_process = [
+                "sudio",
+                "-u",
+                get_username(),
+                f"HOME=/nail/home/{get_username()}",
+            ] + cred_process
+        cred_process_cmd = subprocess.run(cred_process, stdout=subprocess.PIPE)
         cmd_output = json.loads(cred_process_cmd.stdout)
+    else:
+        # Nothing to do here
+        return {}
 
     if "Credentials" in cmd_output:
         cmd_output = cmd_output["Credentials"]
@@ -854,10 +895,10 @@ def run_docker_container(
                 sys.exit(1)
         environment.update(secret_environment)
     if assume_role or assume_pod_identity or aws_profile:
-        creds_dict = assume_aws_role(
+        aws_creds = assume_aws_role(
             instance_config, service, assume_role, assume_pod_identity, aws_profile
         )
-        environment.update(creds_dict)
+        environment.update(aws_creds)
 
     local_run_environment = get_local_run_environment_vars(
         instance_config=instance_config, port0=chosen_port, framework=framework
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +678,89 @@ def check_if_port_free(port):
...
+            "Error: --assume-pod-identity passed but no pod identity was found for this instance",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    default_profile = ""
+    if not aws_profile:
+        with open("/nail/etc/ecosystem") as ecosystem_file:
+            default_profile = ecosystem_file.read()
+    profile = aws_profile or default_profile
+    if pod_identity and assume_pod_identity:
```
jfongatyelp, line 699: I don't see us ever setting `assume_pod_identity = True` if `assume_role` is set, but I do imagine we'd want to do this stuff if only `assume_role` is given?  So should this be `if pod_identity and (assume_pod_identity or assume_role)`?

paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +678,89 @@ def check_if_port_free(port):
...
+                pod_identity,
+                service,
+                get_username(),
+            )
+        else:
+            assume_role_cmd = "aws --profile {} sts assume-role --role-arn {} --role-session-name {}-{}"
+            assume_role_cmd = assume_role_cmd.format(
+                profile, pod_identity, service, get_username()
+            )
+        cmd = subprocess.run(assume_role_cmd, shell=True, stdout=subprocess.PIPE)
```
jfongatyelp, line 720: I might be reading too quickly, but don't see how we ever grab the creds from this and return it? Should we be doing 
```
cmd_output = json.loads(cmd.stdout)
```
like we do below for cred_process_cmd when we pull creds w/o assuming a role?

paasta_tools/cli/cmds/local_run.py
```diff
@@ -462,6 +463,30 @@ def add_subparser(subparsers):
         action="store_true",
         default=False,
     )
+    list_parser.add_argument(
+        "--assume-role",
+        help="role ARN to assume before launching the service",
```
jfongatyelp, line 468: Would help to show an example ARN (eg full `arn:aws:iam::...:role/...`) for what's expected, based on the param name `assume-role` my first instinct would be to just erroneously try to give the role name without something reminding me of the full ARN format.

jfongatyelp (commented):
This is a great bandaid to our local-run problems, thanks for this.

paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +678,89 @@ def check_if_port_free(port):
...
+        with open("/nail/etc/ecosystem") as ecosystem_file:
+            default_profile = ecosystem_file.read()
+    profile = aws_profile or default_profile
+    if pod_identity and assume_pod_identity:
+        print(
+            "Attempting to assume role {} using profile {}".format(
+                pod_identity, profile
+            )
+        )
+        if os.getuid() == 0:
```
nemacysts, line 705: just curious: when does this happen? is this due to the re-exec as root that we do in certain cases?

nemacysts (commented):
++++++

thank you!

paasta_tools/cli/cmds/local_run.py
Qmando: I was generally testing with `--pull` which will re-exec as root as you mention

paasta_tools/cli/cmds/local_run.py
danielpops: I feel that using `aws-okta` and letting the end-user choose the account / role / etc. interactively might be an easier interface for most than forcing folks to specify the full role-arn. Having said that, I haven't actually reviewed any of the code yet :)

paasta_tools/cli/cmds/local_run.py
```diff
@@ -462,6 +463,30 @@ def add_subparser(subparsers):
         action="store_true",
         default=False,
     )
+    list_parser.add_argument(
+        "--assume-role",
```
danielpops, line 467: if you want the ARN, i'd recommend changing this parameter to e.g. `--assume-role-arn` or `--role-arn`. `--assume-role` to me reads as a boolean :)

jfongatyelp: ++ agree, and would help remove confusion on just giving role name vs arn

paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +678,89 @@ def check_if_port_free(port):
...
+        print(
+            "Error: --assume-pod-identity passed but no pod identity was found for this instance",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    default_profile = ""
+    if not aws_profile:
+        with open("/nail/etc/ecosystem") as ecosystem_file:
+            default_profile = ecosystem_file.read()
+    profile = aws_profile or default_profile
```
danielpops, line 698: @qlo I think using the current ecosystem and letting the user choose a profile is likely to cause problems, since people might have different profiles set up in depending on their setups (or they may be running this from adhoc-prod, where they don't even have a `~/.aws/credentials` file set up. 

Mentioned in another comment, what do you think about using aws-okta, and choosing an account based on a static mapping from the clustername? e.g. for `-c pnw-prod` you would use `aws-okta -a prod`, for `-c norcal-devc`, you'd use `aws-okta -a dev` 

Then the user can be prompted to choose which role _they_ have access to assume, and then we can use those creds downstream to try and assume the pod identity role

```
$ aws-okta -a prod
Accessing aa-arbitrator socket to fetch Okta cookie (ctrl-c to interrupt) ...
Please choose the role you would like to assume:
[0]: arn:aws:iam::056083216413:role/security
[1]: arn:aws:iam::056083216413:role/security-powerusers
```

I guess the messaging is a little misleading, since it asks you "which role do you want to assume", and then we're going to do some implicit downstream role assumption that they may or may not have access to.

Qmando: This does seem like a better experience. Probably 99.99% of aws profiles here are going to be aws-okta anyway, which is probably the only argument for actually going through AWS cli itself.

paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +690,82 @@ def check_if_port_free(port):
...
+    if pod_identity and (assume_pod_identity or assume_role_arn):
+        print(
+            "Calling aws-okta to assume role {} using account {}".format(
+                pod_identity, aws_account
+            )
+        )
+    elif use_okta_role:
+        print(f"Calling aws-okta using account {aws_account}")
+    else:
+        # use_okta_role, assume_pod_identity, and assume_role are all empty. This shouldn't happen
```
nemacysts, line 721: should we print something here instead of silently returning nothing? or maybe error out in this case?

nemacysts: i'd probably err on the side of erroring out unless we want to break this function up into multiple functions and have the code that calls `assume_aws_role()` instead call the specific version that does what the cli args call for (e.g., `assume_okta_role()`, `assume_pod_identity_role()`, `assume_role_arn()`)

paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +690,82 @@ def check_if_port_free(port):
...
+    if assume_role_arn:
+        pod_identity = assume_role_arn
+    if assume_pod_identity and not pod_identity:
+        print(
+            f"Error: --assume-pod-identity passed but no pod identity was found for this instance ({instance_config.instance})",
+            file=sys.stderr,
+        )
+        sys.exit(1)
+    with open("/nail/etc/runtimeenv") as runtimeenv_file:
+        aws_account = runtimeenv_file.read()
```
nemacysts, line 711: hmm, i just realized that this won't work for `stage*` workloads - those live in the dev account but have their own runtimeenv (and at least my config for these reads `-a "dev"`

paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +690,82 @@ def check_if_port_free(port):
...
+                pod_identity, aws_account
+            )
+        )
+    elif use_okta_role:
+        print(f"Calling aws-okta using account {aws_account}")
+    else:
+        # use_okta_role, assume_pod_identity, and assume_role are all empty. This shouldn't happen
+        return {}
+
+    if os.getuid() == 0:
```
nemacysts, line 724: thoughts on commenting this a tiny bit for new folks looking at this in the future
```suggestion
    # local-run will sometimes run as root - make sure that we get the actual
    # users AWS credentials instead of looking for non-existent root AWS
    # credentials
    if os.getuid() == 0:
```

paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +690,82 @@ def check_if_port_free(port):
...
+            aws_secret_access_key=cmd_output["SecretAccessKey"],
+            aws_session_token=cmd_output["SessionToken"],
+        )
+        sts_client = boto_session.client("sts")
+        assumed_role = sts_client.assume_role(
+            RoleArn=pod_identity, RoleSessionName=f"{get_username()}-local-run"
+        )
+        cmd_output = assumed_role["Credentials"]
+
+    if "Credentials" in cmd_output:
```
nemacysts, line 759: might be worth commenting in which cases we're expecting the creds to not be top-level in the json and therefore need to "dig" into the returned output

paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +690,82 @@ def check_if_port_free(port):
...
+    """Runs AWS cli to assume into the correct role, then extract and return the ENV variables from that session"""
+    pod_identity = instance_config.get_iam_role()
+    if assume_role_arn:
+        pod_identity = assume_role_arn
+    if assume_pod_identity and not pod_identity:
+        print(
+            f"Error: --assume-pod-identity passed but no pod identity was found for this instance ({instance_config.instance})",
+            file=sys.stderr,
+        )
+        sys.exit(1)
```
danielpops, line 709: missing test coverage for this scenario

paasta_tools/cli/cmds/local_run.py
danielpops: also worth adding a unit test for this scenario

paasta_tools/cli/cmds/local_run.py
```diff
@@ -742,6 +858,16 @@ def run_docker_container(
...
         environment.update(secret_environment)
+    if assume_role_arn or assume_pod_identity or use_okta_role:
+        aws_creds = assume_aws_role(
+            instance_config,
+            service,
+            assume_role_arn,
+            assume_pod_identity,
+            use_okta_role,
+        )
+        environment.update(aws_creds)
```
danielpops, line 869: missing test coverage for this scenario

danielpops (approved):
lgtm, aside from the failing tests run on github actions

danielpops (approved):

paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +690,98 @@ def check_if_port_free(port):
...
+        )
+        # The contents of "Credentials" key from assume_role is the same as from aws-okta
+        cmd_output = assumed_role["Credentials"]
+
+    creds_dict = {
+        "AWS_ACCESS_KEY_ID": cmd_output["AccessKeyId"],
+        "AWS_SECRET_ACCESS_KEY": cmd_output["SecretAccessKey"],
+        "AWS_SESSION_TOKEN": cmd_output["SessionToken"],
+    }
+    return creds_dict
```
nemacysts, line 782: looks like mypy isn't happy with this for some reason?
I'm seeing: `paasta_tools/cli/cmds/local_run.py:782: error: Incompatible return value type (got "Dict[str, Any]", expected "AWSSessionCreds")` in the tests

i'm not quite sure why it's not able to infer things here (maybe our mypy is too old?) - but something like 
```suggestion
    creds_dict: AWSSessionCreds = {
        "AWS_ACCESS_KEY_ID": cmd_output["AccessKeyId"],
        "AWS_SECRET_ACCESS_KEY": cmd_output["SecretAccessKey"],
        "AWS_SESSION_TOKEN": cmd_output["SessionToken"],
    }
    return creds_dict
```
or 
```suggestion
    creds_dict = {
        "AWS_ACCESS_KEY_ID": cmd_output["AccessKeyId"],
        "AWS_SECRET_ACCESS_KEY": cmd_output["SecretAccessKey"],
        "AWS_SESSION_TOKEN": cmd_output["SessionToken"],
    }
    return cast(AWSSessionCreds, creds_dict)
```
*should* resolve that

paasta_tools/cli/cmds/local_run.py
```diff
@@ -653,6 +690,98 @@ def check_if_port_free(port):
...
+        )
+        sys.exit(1)
+    try:
+        with open("/nail/etc/runtimeenv") as runtimeenv_file:
+            aws_account = runtimeenv_file.read()
+            # Map runtimeenv in special cases to proper aws account name
+            if aws_account == "stage":
+                aws_account = "dev"
+            elif aws_account == "corp":
+                aws_account = "corpprod"
```
nemacysts, line 717: something like:
```Python
RUNTIMEENV_TO_ACCOUNT_OVERRIDES = {
    'stage': 'dev',
    'corp': 'corpprod',
}
...
def assume_aws_role(
    ...
):
     ...
     with open("/nail/etc/runtimeenv") as runtimeenv_file:
            aws_account = runtimeenv_file.read()
            # generally, every runtimeenv maps to an account with the same name
            # ...but there are exceptions to every rule.
            RUNTIMEENV_TO_ACCOUNT_OVERRIDES.get(aws_account, aws_account)
```
might be a little neater, but this is fine as-is if you prefer the existing code

tests/cli/test_cmds_local_run.py
```diff
@@ -1963,6 +1989,102 @@ def test_missing_volumes_skipped(mock_exists):
...
+    "paasta_tools.cli.cmds.local_run.open",
+    new_callable=mock.mock_open(),
+    autospec=None,
+)
+@mock.patch("os.makedirs", autospec=True)
+@mock.patch(
+    "paasta_tools.cli.cmds.local_run.assume_aws_role",
+    autospec=True,
+    return_value={"access_key": "abcdefg", "secret_key": "abcdefg"},
+)
```
nemacysts, line 2025: using the context manager form of mocks is a little more preferable than the decorator version (easier to scope the changes to just the code that needs the mocks active and easier to update the mocks used without needing to also update the function signature) - but we use this form enough already that it's likely not  worth updating this unless this comment somehow speaks to you :p

tests/cli/test_cmds_local_run.py
```diff
@@ -1963,6 +1989,102 @@ def test_missing_volumes_skipped(mock_exists):
...
+    mock_text_io_wrapper = mock.Mock(name="text_io_wrapper", autospec=True)
+    # Each file will try to be written up to twice (first non-binary then binary if non-binary fails)
+    # So we raise once, implying that the first file is binary and let the second write() succeed
+    # For the second file, we imply the file is non-binary and only need to mock success once
+    mock_text_io_wrapper.write.side_effect = [TypeError, mock.DEFAULT, mock.DEFAULT]
+
+    # Magic to make the context manager return the mock that we actually want
+    # Otherwise it just returns a new mock_open each time
+    mock_open.return_value = mock_open
+    mock_open.__enter__.return_value = mock_text_io_wrapper
```
nemacysts, line 2057: unfortunate that https://docs.python.org/3/library/unittest.mock.html#mock-open isn't flexible enough for this :p

nemacysts (approved):
lgtm after mypy issues are fixed - i left a couple other comments but feel free to drop 'em

tests/cli/test_cmds_local_run.py
Qmando: True, mostly just kept with the style of the existing tests.

PR #3535, opened 2023-01-19
Adding support to any ns that runs a paasta service in check_kubernetes_services_replication - PAASTA-17846

EmanElsaban:
### Feature

This PR adds support to arbitrary namespaces for paasta services.

### Testing 

Unit tests pass

**Note:** Added the elif  block because I realized that we call main() passing namespaces for ``check_cassandracluster_services_replication`` and ``check_flink_services_health``

### DIFF
--- a/paasta_tools/check_marathon_services_replication.py
+++ b/paasta_tools/check_marathon_services_replication.py
@@ -130,5 +130,4 @@ if __name__ == "__main__":
         instance_type_class=marathon_tools.MarathonServiceConfig,
         check_service_replication=check_service_replication,
         namespace=None,  # not relevant for mesos
-        mesos=True,
     )
--- a/paasta_tools/check_services_replication_tools.py
+++ b/paasta_tools/check_services_replication_tools.py
@@ -16,6 +16,7 @@ import logging
 import sys
 from typing import Any
 from typing import Callable
+from typing import Container
 from typing import List
 from typing import Optional
 from typing import Sequence
@@ -114,7 +115,7 @@ def parse_args() -> argparse.Namespace:
         help="prefix of the namespace to check services replication for"
         "Used only when service is kubernetes",
         dest="namespace_prefix",
-        default="paastasvc",
+        default="paastasvc-",
     )
     parser.add_argument(
         "--additional-namespaces",
@@ -196,7 +197,6 @@ def main(
     instance_type_class: Type[InstanceConfig_T],
     check_service_replication: CheckServiceReplication,
     namespace: str = None,
-    kubernetes_services: bool = False,
 ) -> None:
     args = parse_args()
     if args.verbose:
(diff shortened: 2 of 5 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -2707,6 +2707,12 @@ def get_all_pods(kube_client: KubeClient, namespace: str = "paasta") -> Sequence
     return kube_client.core.list_namespaced_pod(namespace=namespace).items
 
 
+def get_pods_in_all_namespaces(
+    kube_client: KubeClient, label_selector: str = "paasta.yelp.com/service"
+) -> Sequence[V1Pod]:
```
nemacysts, line 2712: normally I'd write some words here about how we should probably prefer more concrete types over ABCs, but we might as well be consistent with the rest of this file :p

paasta_tools/check_marathon_services_replication.py
```diff
@@ -130,5 +130,4 @@ def check_service_replication(
         instance_type_class=marathon_tools.MarathonServiceConfig,
         check_service_replication=check_service_replication,
         namespace=None,  # not relevant for mesos
-        mesos=True,
```
EvanKrall, line 133: We're probably able to delete this file entirely (though it probably makes sense to do it as a follow-up.)

EvanKrall (approved):

paasta_tools/check_services_replication_tools.py
```diff
@@ -190,14 +208,24 @@ def main(
...
 
-    if mesos:
-        tasks_or_pods, slaves = get_mesos_tasks_and_slaves(system_paasta_config)
-        replication_checker = MesosSmartstackEnvoyReplicationChecker(
-            mesos_slaves=slaves,
+    # throw an exception if kubernetes_services flag is false and namespace is not set
+    if kubernetes_services is False and namespace is None:
+        raise Exception("Please set namespace when calling the main function")
+
+    if kubernetes_services:
```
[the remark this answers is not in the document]
EmanElsaban: yep, ``kubernetes_services`` is a flag that is true when being called by ``check_kubernetes_services_replication`` and false otherwise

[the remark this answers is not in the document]
EmanElsaban: thank you, will go with the first solution of removing the flag and just checking if namespace is none since the second one will require us to change the cron job (I assume in puppet)

paasta_tools/check_services_replication_tools.py
```diff
@@ -249,10 +272,26 @@ def get_mesos_tasks_and_slaves(
 
 
 def get_kubernetes_pods_and_nodes(
-    namespace: str,
+    namespace_prefix: str = None,
+    namespace: str = None,
+    additional_namespaces: List[str] = None,
 ) -> Tuple[Sequence[V1Pod], Sequence[V1Node]]:
```
EvanKrall, line 278: I'd switch these return types from Sequence to List.

paasta_tools/check_services_replication_tools.py
```diff
@@ -249,10 +272,26 @@ def get_mesos_tasks_and_slaves(
 
 
 def get_kubernetes_pods_and_nodes(
-    namespace: str,
+    namespace_prefix: str = None,
+    namespace: str = None,
+    additional_namespaces: List[str] = None,
```
EvanKrall, line 277: I think these should be marked as Optional.

(Or, for `additional_namespaces`, switch to `Sequence[str]` and switch the default value to an empty tuple)

EvanKrall (changes requested):
Just noticed these type annotation issues.

https://yelpwiki.yelpcorp.com/display/~tzhu/Accept+general+types+and+return+specific+types

EvanKrall (approved):

PR #3536, opened 2023-01-23
[PAASTA-17844] Mount Vault KV secrets as k8s secret volumes

nurdann:
## TODO
- [x] Choose a key name for yelpsoa-config, i.e. `crypto_keys` 
- [x] Add schema changes to PaaSTA to test code fully https://github.com/Yelp/paasta/pull/3542
- [ ] (Non-blocking) Figure a way to distribute Vault keys to ecosystems based on `vault_cluster_map`
- [x] How to properly call Vault API; use `vault-tools`
- [x] Always pull arbitrary amount of key versions from Vault? That's an API call for each key version
- [x] How to map Vault key to V1Secret name (must match `[-._a-zA-Z0-9]+`)
- [x] Split test cases
- [ ] (Will address in a separate PR) Resolving ecosystem https://github.com/Yelp/paasta/pull/3536#discussion_r1145106912

## Workflow
Given Vault keys in yelpsoa-configs, e.g. `{"crypto_keys": {"decrypt": "foo"}}` maps to Vault key `private/foo`
1. Map yelpsoa-config keys to Vault keys under `private/` or `public/` paths
https://github.com/Yelp/paasta/blob/841051b60e3f15b36683aca42ae4e158d5745396/paasta_tools/kubernetes_tools.py#L466-L470
1. Fetch Vault keys using `kube_master` backend role 
https://github.com/Yelp/paasta/blob/841051b60e3f15b36683aca42ae4e158d5745396/paasta_tools/kubernetes/bin/paasta_secrets_sync.py#L344-L356
3. Create k8s secret, e.g. `private-foo` key under `paasta-crypto-key-` prefix at service level
https://github.com/Yelp/paasta/blob/841051b60e3f15b36683aca42ae4e158d5745396/paasta_tools/kubernetes/bin/paasta_secrets_sync.py#L376
4. Create k8s volume with secret: path is equivalent to Vault key, i.e. `/etc/crypto_keys/private/foo.json`
https://github.com/Yelp/paasta/blob/841051b60e3f15b36683aca42ae4e158d5745396/paasta_tools/kubernetes_tools.py#L1482-L1491

## Changes
- Add key signature as k8s configmap
- Add Vault key versions as JSON blob to k8s secrets
- [Optionally](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1SecretVolumeSource.md) mount k8s secrets, since warnings from `paasta_secrets_sync` should be enough to identify missing Vault keys that yelpsoa-configs expects
- Merge functionality for `boto_keys` and `crypto_keys` to update k8s secrets

## Questions
- Why is SecretProvider imported dynamically?https://github.com/Yelp/paasta/blob/8b51f18797c27f27140fc00fc1694b6600538c6f/paasta_tools/secret_tools.py#L76-L86 

### Notes

Expecting a yelpsoa-config with following format
```
crypto_keys:
  encrypt:
    - foo
  decrypt:
    - foo
```

Secret name is given by
```
             limit_size_with_hash(
                f"paasta-boto-key-{get_kubernetes_app_name(service, instance)}"
            )
```
which is same as `KubernetesDeploymentConfig.get_crypto_secret_name()`

Then it is mounted by referencing it 
```
        volume = V1Volume(
            name=self.get_crypto_secret_volume_name(service_name),
            secret=V1SecretVolumeSource(
                secret_name=self.get_crypto_secret_name(),
```

And the actual content of secret is specified in
```
                V1KeyToPath(
                    key=vault_key_to_V1Secret_data_key(crypto_key),
                    path=f"{crypto_key}.json",
```
where the `key` parameter must exist in k8s secret `.data` (str-to-str) mapping as a key.

## Tests
Tested on kubestage master manually, see y/how-to-test-mounting-k8s-secrets

After running on kubestage master
```
setup_kubernetes_job compute-infra-test-service.autoscaling_2
```
Succesfully mounted a key https://fluffy.yelpcorp.com/i/Hg3SVmrpfzT8vzBhQhn4gxXbLwL83Zl7.html


nemacysts (approved):

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -328,70 +436,87 @@ def sync_boto_secrets(
...
+    return True
+
+
+def update_k8s_secrets(
+    service: str,
+    secret: str,
+    secret_data: Mapping[str, str],
+    kube_client: KubeClient,
+    namespace: str,
+):
```
nemacysts, line 460: nit: 
```suggestion
) -> None:
```

paasta_tools/kubernetes_tools.py
```diff
@@ -453,6 +454,22 @@ def limit_size_with_hash(name: str, limit: int = 63, suffix: int = 4) -> str:
...
+def get_vault_key_secret_name(vault_key: str) -> str:
+    """
+    Vault path may contain `/` slashes which is invalid as secret name
+    V1Secret's data key must match regexp [a-zA-Z0-9._-]
+    Source: https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/V1Secret.md
+    """
+    return vault_key.replace("/", "-")
+
+
+def get_crypto_keys_from_config(crypto_keys_config: dict) -> List[str]:
```
nemacysts: suggestion: we should probably have a TypedDict for `crypto_keys_config`

paasta_tools/kubernetes_tools.py
```diff
@@ -1432,6 +1459,41 @@ def get_boto_volume(self) -> Optional[V1Volume]:
         )
         return volume
 
+    def get_crypto_volume(self) -> Optional[V1Volume]:
+        required_crypto_keys = get_crypto_keys_from_config(
+            self.config_dict.get("crypto_keys", {})
+        )
```
nemacysts, line 1465: thoughts on moving `get_crypto_keys_from_config` into this class so that this looks like:

```suggestion
        required_crypto_keys = self.get_crypto_keys_from_config()
```

and `get_crypto_keys_from_config` then looks like:
```python
def get_crypto_keys_from_config() -> List[str]:
    encryption_keys = self.config_dict.get("crypto_keys", {}).get("encrypt", [])
    decryption_keys = self.config_dict.get("crypto_keys", {}).get("decrypt", [])

    return [
        *(f"public/{key}" for key in encryption_keys),
        *(f"private/{key}" for key in decryption_keys),
    ]
```

it seems unlikely that we'll want to call get_crypto_keys_from_config outside the context of this class, so we might as well use the state that the class already holds rather than making a free function and having to split the dict traversal up into callee and caller portions

paasta_tools/kubernetes_tools.py
```diff
@@ -1501,6 +1579,18 @@ def get_boto_secret_hash(self) -> str:
             namespace=self.get_namespace(),
         )
 
+    def get_crypto_secret_hash(self) -> str:
```
nemacysts, line 1582: i'm surprised mypy isn't failing on this - `get_kubernetes_secret_signature` returns `Optional[str]`, not `str`

paasta_tools/kubernetes_tools.py
```diff
@@ -1487,6 +1549,22 @@ def get_volume_mounts(
                         volume_mounts.remove(existing_mount)
                         break
                 volume_mounts.append(mount)
+
+        if self.config_dict.get("crypto_keys", []):
+            if self.get_crypto_secret_hash():
+                mount = V1VolumeMount(
+                    mount_path="/etc/crypto_keys",
```
nemacysts, line 1556: should we move `"/etc/crypto_keys"` to a named constant? we do refer to it several times in this function already :)

tests/test_kubernetes_tools.py
```diff
@@ -1316,6 +1317,37 @@ def test_get_boto_volume(self, config_dict):
...
+            volumes = deployment.get_crypto_volume()
+        if config_dict:
+            (volumes.secret.secret_name == "paasta-crypto-key-my-service-my-instance")
+
+            assert len(volumes.secret.items) == len(config_dict["crypto_keys"])
+            assert set(get_crypto_keys_from_config(config_dict["crypto_keys"])) == {
+                item.path.rstrip(".json") for item in volumes.secret.items
+            }
+        else:
+            assert volumes is None
```
nemacysts, line 1349: should we split this test in two - i see that we followed this pattern for the boto tests, but it always makes me a little nervous to have branches in tests :)

tests/kubernetes/bin/test_paasta_secrets_sync.py
```diff
@@ -369,3 +371,138 @@ def test_sync_boto_secrets():
         assert mock_update_secret.called
         assert mock_create_kubernetes_secret_signature.called
         assert not mock_update_kubernetes_secret_signature.called
+
+
+def test_sync_crypto_secrets():
```
nemacysts, line 376: it'd be nice to split this and the other sync tests into multiple rather than testing every case in a single function body - but this is probably fine for now :p

paasta_tools/secret_providers/vault.py
```diff
@@ -159,3 +164,34 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
...
+
+    def get_vault_key_versions(
+        self,
+        key: str,
+        mountpoint: str = "keystore",
+    ) -> Iterable[Dict[str, Any]]:
+        """
+        Retrieve Vault key's all versions based on its metadata
+        """
+        client = self.clients[self.ecosystems[0]]
```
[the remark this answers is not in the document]
nemacysts: that said, should we do anything here to account for there not being a client for the target ecosystem? or will the keyerror be handled by the caller instead?

paasta_tools/secret_providers/vault.py
```diff
@@ -159,3 +164,34 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
...
+                yield {
+                    "keyname": key,
+                    "key": key_response["data"]["data"]["key"],
+                    "key_version": key_response["data"]["metadata"]["version"],
+                }
+        except hvac.exceptions.VaultError as e:
+            log.warning(
+                f"Could not fetch key versions for {key} on {self.ecosystems[0]} because of {type(e).__name__}"
+            )
+            pass
```
nemacysts, line 195: we can clean this up a little - the python logging system has a helper for automatically including traceback information alongside a log message, and we don't need a `pass` in a block that is otherwise non-empty
```suggestion
        except hvac.exceptions.VaultError:
            log.exception(
                f"Could not fetch key versions for {key} on {self.ecosystems[0]}"
            )
```

paasta_tools/secret_providers/vault.py
```diff
@@ -159,3 +164,34 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
...
         else:
             return None
+
+    def get_vault_key_versions(
+        self,
+        key: str,
+        mountpoint: str = "keystore",
+    ) -> Iterable[Dict[str, Any]]:
+        """
+        Retrieve Vault key's all versions based on its metadata
```
nemacysts, line 174:
```suggestion
        Retrieve all versions of a Vault key based on its metadata
```

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -305,6 +318,71 @@ def sync_secrets(
...
+        time.sleep(0.3)
+        update_k8s_secrets(
+            service=service,
+            # `kubernetes.client.V1SecretVolumeSource` expects the secret name below
+            secret=limit_size_with_hash(
+                f"paasta-crypto-key-{get_kubernetes_app_name(service, instance)}"
+            ),
+            secret_data=secret_data,
+            kube_client=kube_client,
+            namespace=namespace,
```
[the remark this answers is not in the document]
cuza: @nemacysts I would leave as is, get_namespace() is not making k8s api calls is just checking the `namespace` key in SOAConfigs and returning whatever is in there or "paasta" if it does not exists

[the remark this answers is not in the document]
nemacysts: yea - that's what I meant: should this be calling get_namespace() rather than passing a hard-coded one? :)

k8s_itests/deployments/paasta/fake_etc_paasta/logs.json
```diff
@@ -0,0 +1,10 @@
+{
+    "log_writer": {
+        "driver": "monk",
+        "options": {
+            "monk_host": "169.254.255.254",
+            "monk_port": 1473,
+            "monk_disable": true
+        }
+    }
+}
```
cuza, line 10: Maybe we might be better using the `null` log writer here

```suggestion
{
    "log_writer": {
        "driver": "null"
    }
}
```

Makefile
```diff
@@ -157,11 +157,16 @@ setup-kubernetes-job: k8s_fake_cluster generate_deployments_for_service
...
 .PHONY: paasta-secrets-sync
-paasta-secrets-sync: setup-kubernetes-job
+paasta-secrets-sync: setup-kubernetes-job .vault-token
 	export KUBECONFIG=./k8s_itests/kubeconfig;\
 	export PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_playground/;\
 	export PAASTA_TEST_CLUSTER=kind-${USER}-k8s-test;\
-	{ .tox/py37-linux/bin/python -m paasta_tools.list_kubernetes_service_instances -d ./soa_config_playground ; echo -n \ _shared; } | cut -f1 -d"." | uniq | shuf | xargs .tox/py37-linux/bin/python -m paasta_tools.kubernetes.bin.paasta_secrets_sync -v -d ./soa_config_playground
+	{ .tox/py37-linux/bin/python -m paasta_tools.list_kubernetes_service_instances -d ./soa_config_playground ; echo -n \ _shared; } | cut -f1 -d"." | uniq | shuf | xargs .tox/py37-linux/bin/python -m paasta_tools.kubernetes.bin.paasta_secrets_sync -v -d ./soa_config_playground -t ./.vault-token
+
+.vault-token:
```
cuza, line 166: We might add `rm .vault-token` to the `clean` target

paasta_tools/cli/cmds/secret.py
```diff
@@ -228,6 +228,7 @@ def _add_common_args(parser: argparse.ArgumentParser, allow_shared: bool = True)
         help="A directory from which yelpsoa-configs should be read from",
         default=DEFAULT_SOA_DIR,
     )
+    parser.add_argument("-t", "--vault_token_file", help="Vault token file")
```
nemacysts, line 231: doesn't this conflict with `--vault-token-file`?

paasta_tools/cli/cmds/secret.py
```diff
@@ -341,7 +342,10 @@ def is_service_folder(soa_dir, service_name):
 
 
 def _get_secret_provider_for_service(
-    service_name, cluster_names=None, soa_dir=None, secret_provider_extra_kwargs=None
+    service_name,
+    cluster_names=None,
+    soa_dir=None,
+    secret_provider_extra_kwargs=None,
 ):
```
nemacysts, line 349:
```suggestion
    service_name: str,
    cluster_names: Optional[str]=None,
    soa_dir: Optional[str]=None,
    secret_provider_extra_kwargs: Optional[Dict[str, str]=None,
) -> SecretProvider:
```
would be cool (although probably overkill) to have a non-total TypedDict for the kwargs we're passing here unless `get_secret_provider` and it's callees can handle any arbitrary kwarg that's passed in (in which case, nvm :p)

that said: looks like all we did here was reformat, so feel free to drop this comment

paasta_tools/kubernetes_tools.py
```diff
@@ -1416,22 +1441,62 @@ def get_boto_volume(self) -> Optional[V1Volume]:
...
+                        # key should exist in data section of k8s secret
+                        key=get_vault_key_secret_name(crypto_key),
+                        # path is equivalent to Vault key directory structure
+                        # e.g. private/foo will create /etc/crypto_keys/private/foo.json
+                        path=f"{crypto_key}.json",
+                        mode=mode_to_int("0444"),
+                    )
+                    for crypto_key in required_crypto_keys
+                ],
+                optional=True,
```
nemacysts, line 1496: just to confirm: we're expecting that users of this feature check that these secrets are available before using them, right? otherwise, we'd want to remove this optional flag so that we don't run the pod if we cannot use the volume

paasta_tools/kubernetes_tools.py
```diff
@@ -1487,20 +1552,50 @@ def get_volume_mounts(
...
+    def get_k8s_secret_name(self, name: str) -> str:
+        return limit_size_with_hash(
+            f"{self.get_namespace()}-{name}-{self.get_sanitised_deployment_name()}"
+        )
+
+    def get_boto_k8s_secret_name(self) -> str:
+        return self.get_k8s_secret_name("boto-key")
+
+    def get_crypto_k8s_secret_name(self) -> str:
+        return self.get_k8s_secret_name("crypto-key")
```
nemacysts, line 1582: ++!

paasta_tools/secret_providers/vault.py
```diff
@@ -159,3 +161,34 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
...
+            meta_response = client.secrets.kv.read_secret_metadata(
+                path=key, mount_point=mountpoint
+            )
+
+            for key_version in meta_response["data"]["versions"].keys():
+                key_response = client.secrets.kv.read_secret_version(
+                    path=key, version=key_version, mount_point=mountpoint
+                )
+                yield {
+                    "keyname": key,
```
nemacysts, line 184: should this have been named `key_name`?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -176,11 +173,11 @@ def get_services_to_k8s_namespaces(
...
     kube_client: KubeClient,
     cluster: str,
-    services_to_k8s_namespaces: Mapping[str, Set],
+    services_to_k8s_namespaces: Dict[str, Set],
     secret_provider_name: str,
-    vault_cluster_config: Mapping[str, str],
+    vault_cluster_config: Dict[str, str],
     soa_dir: str,
-    vault_token_file: str = DEFAULT_VAULT_TOKEN_FILE,
+    vault_token_file: Optional[str] = None,
```
nemacysts, line 180: just curious: why did we need to change the default value?

nemacysts: and will the callees of this function handle potentially getting a `None` value? I might be missing something, but I doesn't look like we've added any error handling for the `None` case

paasta_tools/secret_providers/__init__.py
```diff
@@ -49,4 +50,11 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
...
 class SecretProvider(BaseSecretProvider):
-    pass
+    # from .vault import CryptoKey # cannnot import of cyclical dependency
+    def get_vault_key_versions(
+        self, key_name: str, mount_point: str = "keystore"
+    ) -> Iterable[dict]:
+        """
+        Dummy attribute to satisfy `mypy` because the class is imported dynamically via __import__
+        """
+        raise NotImplementedError
```
nemacysts, line 60: should we `# type: ignore # class is imported dynamically and cannot be discovered by mypy` wherever mypy was failing rather than adding a dummy attribute?

nemacysts: (i'd say we should get rid of this entire interface/indirection since it seems unlikely for us to use anything but vault for the foreseeable future, but that's probably something for future us to discuss/do :p)

paasta_tools/secret_providers/__init__.py
```diff
@@ -49,4 +50,11 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
 
 
 class SecretProvider(BaseSecretProvider):
-    pass
+    # from .vault import CryptoKey # cannnot import of cyclical dependency
```
nemacysts, line 53: if we do stick with this dummy attribute, https://peps.python.org/pep-0484/#runtime-or-type-checking might help type this correctly

paasta_tools/secret_providers/__init__.py
danielpops: While I think this whole secretprovider thing needs to be re thought out, from basic object oriented principles, this base class shouldn't know about `vault` (or kubernetes, or aws ssm, or any other concrete implementation of SecretProvider). I would probably at least rename this to be `get_key_verisions` and any references accordingly.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nurdann: When testing on a playground, some commands were not propagating vault-token because it was being set as null. So default values actually serve no purpose if there are cases when `None` is being passed.
I had to add the following
https://github.com/Yelp/paasta/blob/b9d5953d8c9613ad23a06064494f185f5b2d1146/paasta_tools/cli/cmds/secret.py#L421-L424
to pass the token, and the default argument actually confused me as to why the default argument was not being used
https://github.com/Yelp/paasta/blob/b9d5953d8c9613ad23a06064494f185f5b2d1146/Makefile#L164

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -293,6 +304,99 @@ def sync_secrets(
...
+        cluster_names=[cluster],
+        secret_provider_kwargs={
+            "vault_cluster_config": vault_cluster_config,
+            "vault_auth_method": "token",
+            "vault_token_file": vault_token_file,
+        },
+    )
+
+    # use vault client directly
+    client = provider.clients[provider.ecosystems[0]]
```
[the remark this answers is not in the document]
nemacysts: not sure about your first question (but i'm assuming that as long as puppet has run on the masters, this *should* be able to give us a client)

for your second question:
that said: the way that we organize our clusters internally is generally by superregion (for yelpers: see y/habitat). while an ecosystem can have multiple superregions inside it (e.g., the devc ecosystem contains the norcal-devc and pnw-devc superregions), a superregion can only belong to a single ecosystem. furthermore, a k8s master will only manage a cluster in the same superregion as the master (which therefore means that a k8s master will not manage anything outside of the ecosystem it exists in)

(also, note that we do break this single-cluster-per-superregion trend in several places: see clusters like infrastage and kubestage)

paasta_tools/secret_providers/__init__.py
nemacysts: we could also move the typeddict into this file since if `get_key_versions` is meant to be part of the public interface, we probably want to actually define what that should look like :)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -245,63 +253,89 @@ def sync_secrets(
...
+            ).decode("utf-8")
+
+        if not secret_data:
+            continue
+
+        # In order to prevent slamming the k8s API, add some artificial delay here
+        time.sleep(0.3)
+        update_k8s_secret(
+            service=service,
+            # `kubernetes.client.V1SecretVolumeSource`'s `secret_name` must match `secret` below
```
nemacysts, line 331: editing this to include what you mentioned at https://github.com/Yelp/paasta/pull/3536#discussion_r1148101728 would make this comment clearer (imo)

e.g.,
```suggestion
            # the secret name here must match the secret name given in the secret volume config
```
(or even your comment as-is :))

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -335,73 +367,92 @@ def sync_boto_secrets(
...
-                        namespace=namespace,
-                    )
-                else:
-                    raise
-            create_kubernetes_secret_signature(
+    return True
+
+
+def get_dict_signature(data: Dict[str, str]) -> str:
+    return hashlib.sha1("".join(data.values()).encode("utf-8")).hexdigest()
```
nemacysts, line 388: hmm, thinking about this a little more: should we also include the keys in the signature generation?

e.g.,
```
{
    "super-secret-encryption-key": "jahmdflkjahsdf", 
    "super-secret-decryption-key": "abc",
}
```
is not the same secret data as 
```
    "super-secret-decryption-key": "jahmdflkjahsdf", 
    "super-secret-encryption-key": "abc",
```

also, might be nice to either add a docstring for this function or mark it as private (e.g., `_get_dict_signature`) since this isn't really a function we want folks to be importing as-is

paasta_tools/secret_providers/__init__.py
```diff
@@ -49,4 +50,11 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
...
+    def get_key_versions(
+        self, key_name: str, mount_point: str = "keystore"
```
nemacysts, line 55: are we planning on ever using this mountpoint arg? it feels a little weird to have this be in the interface given that this is so vault-specific

tests/kubernetes/bin/test_paasta_secrets_sync.py
nemacysts, line 1: thanks for refactoring the tests here - having multiple small tests rather than a large `test_sync_secrets` test is awesome :)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -208,6 +205,17 @@ def sync_all_secrets(
                 soa_dir=soa_dir,
             )
         )
+        results.append(
+            sync_crypto_secrets(
```
mattmb, line 209: I wonder if we should start pulling these out into separate jobs? If we keep adding things here it makes it harder to debug failures in my opinion. Also if this code has a bug and stalls/crashes then we might not be able to sync boto secrets or paasta secrets so it increases the reliability risk. If we had a separate cron job for each then we can run/monitor/alert on them separately.

paasta_tools/secret_providers/vault.py
nurdann: The `yield` below should capture any cases falling through

nemacysts: correct - so we should be able to remove the `pass` :)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nurdann: Let's change when making changes to `vault.py`

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -293,6 +306,98 @@ def sync_secrets(
...
+    config_loader = PaastaServiceConfigLoader(service=service, soa_dir=soa_dir)
+    for instance_config in config_loader.instance_configs(
+        cluster=cluster, instance_type_class=KubernetesDeploymentConfig
+    ):
+        instance = instance_config.instance
+        crypto_keys = instance_config.config_dict.get("crypto_keys", ["private/foo"])
+        if not crypto_keys:
+            continue
+
+        crypto_keys.sort()
```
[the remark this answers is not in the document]
nurdann: Vault returns key versions in order so we probably don't have to worry about it for crypto-keys. 
On hindsight, we will probably want to allow mounting specific version of Vault key only.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nurdann: How should keys and values be joined? Should we include some kind of separator otherwise there's still chance of overlapping signatures

paasta_tools/secret_providers/__init__.py
nurdann: Mountpoint is set on Vault backend so as long as it doesn't change, then we can hard code mountpoint. So it shouldn't be used mostly.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nemacysts: maybe something like this?
```Python
def _get_dict_signature(data: Dict[str, str]) -> str:
    unhashed_contents = (f"{key}:{value}" for key, value in data.items())
    return hashlib.sha1("|".join(unhashed_contents).encode("utf-8")).hexdigest()
```

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -335,73 +396,94 @@ def sync_boto_secrets(
...
-                    )
-                else:
-                    raise
-            create_kubernetes_secret_signature(
+    return True
+
+
+def _get_dict_signature(data: Dict[str, str]) -> str:
+    return hashlib.sha1(
+        "|".join(f"{key}f:{value}" for key, value in data.items()).encode("utf-8")
```
nemacysts, line 418:
```suggestion
        "|".join(f"{key}:{value}" for key, value in data.items()).encode("utf-8")
```

we only need the `f` outside the string :)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -176,20 +184,25 @@ def get_services_to_k8s_namespaces(
...
     soa_dir: str,
-    vault_token_file: str = DEFAULT_VAULT_TOKEN_FILE,
+    vault_token_file: str,
+    secret_type: Literal["all", "paasta-secret", "crypto-key", "boto-key"] = "all",
     overwrite_namespace: Optional[str] = None,
 ) -> bool:
     results = []
+
     for service, namespaces in services_to_k8s_namespaces.items():
+        sync_service_secrets: Dict[str, List[Callable]] = defaultdict(list)
```
nemacysts, line 198: suggestion: 

thoughts on refactoring this a bit to something like:
```suggestion
        sync_service_secrets: Dict[str, Callable] = {
            'paasta-secret': partial(...),
            'boto-key': partial(...),
            'crypto-key': partial(...),
        }
        
        if secret_type == "all":
            result = (
                all(sync() for sync in sync_service_secrets.values())
            )
        else:
            result = sync_service_secrets[secret_type]()
        
        return result
```

so that we don't need to duplicate the secret syncing type key names in both the dict and the `all` case (+ simplify the result logic a bit)?

nemacysts (approved):
I have one suggestion to simplify the secret syncing code - if no one else has reviewed by the time that's done (or dropped), we can probably go ahead and start rolling this out :)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nurdann: `paasta-secret` type still needs to have a list of `Callable` because of namespaces. So, I am inclined to keep the previous changes.

nemacysts: ah, missed that paasta secrets don't auto-discover the namespace from instance configs as the other ones do!

nurdann:
To account for namespace-per-service changes, I have
* Refactor creating (`create_secret`), updating (`update_secret`) and gettings (`get_secret`) kubernetes secrets and similarly for secret signatures, a.k.a configmaps (`create_secret_signature`, `update_secret_signature`, `get_secret_signature`)
   * k8s secret name has 253 character limit, and volume names have 63 character limit [[docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#:~:text=in%20RFC%201123.-,This%20means%20the%20name%20must%3A,start%20with%20an%20alphanumeric%20character)]
    * Looks like we never run into problem with paasta secrets over 63 characters, which was previously at 253 https://github.com/Yelp/paasta/blob/8c2c808928e449e0869e952de74ee7910085ee37/paasta_tools/kubernetes_tools.py#L915-L918
 * All secret names are derived from a single function https://github.com/Yelp/paasta/blob/8c2c808928e449e0869e952de74ee7910085ee37/paasta_tools/kubernetes_tools.py#L3725-L3738 
 * When does service or secret need to be sanitized? Because sometimes they are sanitized and sometimes not. For example, labels are not sanitized for pods
    https://github.com/Yelp/paasta/blob/8c2c808928e449e0869e952de74ee7910085ee37/paasta_tools/kubernetes_tools.py#L1807-L1810
   but are sanitized for secrets on master
   https://github.com/Yelp/paasta/blob/e1206c765f2c5d0ad841cf065726abbc579e2595/paasta_tools/kubernetes_tools.py#L3401-L3417
   For now I am opting not to sanitize them, and it's the responsibility of a callee to give correct values. I believe it decouples k8s api calls from getting correct secret or configmap names.

paasta_tools/kubernetes_tools.py
```diff
@@ -3680,28 +3722,72 @@ def update_crds(
...
+    key_name: str,
     namespace: str = "paasta",
     decode: bool = True,
 ) -> Union[str, bytes]:
-
-    k8s_secret_name = get_kubernetes_secret_name(service_name, secret_name, namespace)
-
     secret_data = kube_client.core.read_namespaced_secret(
-        name=k8s_secret_name, namespace=namespace
-    ).data[secret_name]
```
nemacysts, line 3704: isn't this a behavior change?

nemacysts: oh, I think I see - `secret_name` is coming from `get_paasta_secret_name` which should be doing what `get_kubernetes_secret_name` was doing before for `k8s_secret_name` and `key_name` is now what `secret_name` was before

this just "looked" wrong from the diff :p

nemacysts: I wonder if we should be a little more careful and add docstrings to anything dealing with secrets so that its clear how everything fits together without having to read the code?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -245,63 +283,95 @@ def sync_secrets(
...
-                        update_kubernetes_secret_signature(
-                            kube_client=kube_client,
-                            secret=secret,
-                            service=service,
-                            secret_signature=secret_signature,
-                            namespace=namespace,
-                        )
-                    else:
-                        log.info(f"{secret} for {service} in {namespace} up to date")
+                update_k8s_secret(
```
nemacysts, line 289: should this be renamed so its a little clearer without reading this function's body that it will create the secret if it doesn't exist?
```suggestion
                create_or_update_k8s_secret(
```

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -245,63 +283,95 @@ def sync_secrets(
...
+    soa_dir: str,
+    vault_token_file: str,
+) -> bool:
+    """
+    For each key-name in `crypto_key`,
+    1. Fetch all versions of the key-name from Vault superregion mapped from cluster, e.g. `kubestage` maps to `devc` Vault server.
+    2. Create K8s secret from JSON blob containing all key versions.
+    3. Create signatures as K8s configmap based on JSON blob hash.
+
+    So each instance of a service get the same key, thereby reducing requests to Vault API
```
nemacysts, line 325: is this still true? every instance has its own crypto_keys dict and we're calling `get_secret_provider(...).get_key_versions(...)` for every instance

paasta_tools/kubernetes_tools.py
```diff
@@ -1128,31 +1148,31 @@ def get_kubernetes_secret_env_vars(
...
             secret = get_secret_name_from_ref(v)
-            sanitised_secret = sanitise_kubernetes_name(secret)
             ret.append(
                 V1EnvVar(
                     name=k,
                     value_from=V1EnvVarSource(
                         secret_key_ref=V1SecretKeySelector(
-                            name=f"paasta-secret-{service}-{sanitised_secret}",
+                            name=get_paasta_secret_name(
+                                self.get_namespace(), self.get_service(), secret
```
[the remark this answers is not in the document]
nurdann: everything gets sanitized with the following function at the end https://github.com/Yelp/paasta/blob/98608ec01e6bbee9eda8abb0b7e73651112474f1/paasta_tools/kubernetes_tools.py#L3725-L3736

paasta_tools/kubernetes_tools.py
nurdann: Added some docstring to `create_secret()` let me know if it's sufficient. Or should we create a wiki?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nurdann: Worded it wrong: we are reducing API count for each instance/replica count of a service instance. We are fine making calls for each service instance, but we can probably cache results if soa-configs are same for instances.

paasta_tools/kubernetes_tools.py
```diff
@@ -3359,14 +3359,18 @@ def create_secret(
...
+    """
     kube_client.core.create_namespaced_secret(
         namespace=namespace,
         body=V1Secret(
             metadata=V1ObjectMeta(
                 name=secret_name,
                 labels={
-                    "yelp.com/paasta_service": service_name,
-                    "paasta.yelp.com/service": service_name,
+                    "yelp.com/paasta_service": sanitise_label_name(service_name),
```
nemacysts, line 3372: i think we only really want to limit the length here - there's no restrictions on characters inside the labels

paasta_tools/kubernetes_tools.py
```diff
@@ -3721,6 +3731,16 @@ def update_crds(
     return success
 
 
+def sanitise_label_name(service_name: str):
```
nemacysts, line 3734: technically, we're only using this on label values :)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -245,63 +283,96 @@ def sync_secrets(
...
+    soa_dir: str,
+    vault_token_file: str,
+) -> bool:
+    """
+    For each key-name in `crypto_key`,
+    1. Fetch all versions of the key-name from Vault superregion mapped from cluster, e.g. `kubestage` maps to `devc` Vault server.
+    2. Create K8s secret from JSON blob containing all key versions.
+    3. Create signatures as K8s configmap based on JSON blob hash.
+
+    So each instance count of a service instance get the same key, thereby reducing requests to Vault API
```
nemacysts, line 326:
```suggestion
    So each replica of a service instance gets the same key, thereby reducing requests to Vault API as we only talk to vault during secret syncing
```

paasta_tools/kubernetes_tools.py
nurdann: If I don't truncate I get
```
metadata.labels: Invalid value: \"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\": must be no more than 63 characters
```
from diff
```
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -3369,7 +3369,7 @@ def create_secret(
             metadata=V1ObjectMeta(
                 name=secret_name,
                 labels={
-                    "yelp.com/paasta_service": sanitise_label_name(service_name),
+                    "yelp.com/paasta_service": "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
```

nemacysts: we should still truncate - but https://github.com/Yelp/paasta/pull/3536/files#diff-f1b278fb2675abadd2f58f131c3f67ece5d7511b60fcad691255aeda6f490d17R3739 will do some character replacement which we likely don't want

paasta_tools/kubernetes_tools.py
nemacysts: I was thinking we could have docstrings on all of the functions we use for secret functionality so that it's clear how they all fit together + how they work (although, documenting in either our internal wiki or in the public paasta docs isn't a bad idea either as then we can add system diagrams and whatnot :))

that said, this isn't a blocker - but it'd definitely be nice to do (and in general, we should probably be writing more docstrings in paasta :p)

paasta_tools/kubernetes_tools.py
nemacysts: ah, sorry missed that this is the same as what we used to do in the past

paasta_tools/kubernetes_tools.py
```diff
@@ -3680,28 +3756,121 @@ def update_crds(
...
 
-def get_kubernetes_secret_name(
-    service_name: str, secret_name: str, namespace: str = "paasta"
+def sanitise_label_value(service_name: str) -> str:
+    """
+    :param service_name: service_name is sanitized and limited to 63 characters due to kubernetes restriction
+    :return: Sanitised at most 63-character label value
+    """
+    return limit_size_with_hash(
+        sanitise_kubernetes_name(service_name),
```
nemacysts, line 3765: final nit after our chat in slack:
```suggestion
def sanitise_label_value(value: str) -> str:
    """
    :param value: value is sanitized and limited to 63 characters due to kubernetes restriction
    :return: Sanitised at most 63-character label value
    """
    return limit_size_with_hash(
        sanitise_kubernetes_name(value),
```

nemacysts (commented):
one final comment and then we should be good to go!

paasta_tools/kubernetes_tools.py
nemacysts: (not all label values are service names)

nemacysts (approved):

PR #3542, opened 2023-01-30
Add `crypto_keys` field

nurdann:
Related to #3536. Choosing on `crypto_keys` field name.

### DIFF
--- a/paasta_tools/cli/schemas/adhoc_schema.json
+++ b/paasta_tools/cli/schemas/adhoc_schema.json
@@ -62,16 +62,24 @@
                     }
                 },
                 "crypto_keys": {
-                    "encrypt": {
-                        "type": "array",
-                        "items": {
-                            "type": "string"
-                        }
-                    },
-                    "decrypt": {
-                        "type": "array",
-                        "items": {
-                            "type": "string"
+                    "type": "object",
+                    "additionalProperties": false,
+                    "properties": {
+                        "encrypt": {
+                            "type": "array",
+                            "items": {
+                                "type": "string",
+                                "pattern": "^[a-zA-Z0-9_.-]+$"
+                            },
+                            "uniqueItems": true
+                        },
+                        "decrypt": {
+                            "type": "array",
+                            "items": {
+                                "type": "string",
+                                "pattern": "^[a-zA-Z0-9_.-]+$"
+                            },
+                            "uniqueItems": true
                         }
                     }
                 },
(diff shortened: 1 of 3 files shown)


nemacysts (commented):
this looks fine as-is - but it might be worth to hold off on merging any changes for this feature until the CEP is signed off (e.g., I just made a comment/suggestion about what how this dict could look)

paasta_tools/cli/schemas/tron_schema.json
```diff
@@ -207,6 +207,20 @@
...
+                            "type": "string"
+                        }
+                    },
+                    "decrypt": {
+                        "type": "array",
+                        "items": {
+                            "type": "string"
+                        }
+                    }
+                },
```
nemacysts, line 223: just fyi: we may need to update tron/task_processing to support crypto_keys

paasta_tools/cli/schemas/adhoc_schema.json
```diff
@@ -61,6 +61,20 @@
                         "type": "string"
                     }
                 },
+                "crypto_keys": {
+                    "encrypt": {
+                        "type": "array",
+                        "items": {
+                            "type": "string"
```
nemacysts, line 68: would it help to add a regex pattern to the schema for what the items should look like (e.g., the stuff in these arrays should match `[a-zA-Z]+`)?

nemacysts (approved):

paasta_tools/cli/schemas/tron_schema.json
nurdann: where would that need to happen?

nemacysts: for the payload that tron gets: probably somewhere around https://github.com/Yelp/paasta/blob/master/paasta_tools/tron_tools.py#L884

for processing in tron: probably something like https://github.com/Yelp/Tron/pull/822

and for task_processing: you'll likely want to pick up the work in https://github.com/Yelp/task_processing/pull/188/files


that said: if we're not expecting tronjobs to need this, it'll be a lot easier to just exclude this from the schema :p

nemacysts (approved):

paasta_tools/cli/schemas/tron_schema.json
nurdann: If the need arises, we can support direct authentication via AWS. So we don't need to make changes for tronjobs currently.

PR #3544, opened 2023-02-02
Clearing paasta status threads after catching KeyboardInterrupt

cuza:
This PR will clear threads after catching a `KeyboardInterrupt` to make sure that `paasta status` don't spookily 👻👻 take over people’s terminals

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -2414,6 +2414,7 @@ def paasta_status(args) -> int:
         except KeyboardInterrupt:
             executor._threads.clear()  # type: ignore
             concurrent.futures.thread._threads_queues.clear()  # type: ignore
+            raise KeyboardInterrupt
 
     return max(return_codes)
 
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -2412,6 +2412,9 @@ def paasta_status(args) -> int:
                 return_code, output = future.result()
                 return_codes.append(return_code)
         except KeyboardInterrupt:
+            # ideally we wouldn't need to reach into `ThreadPoolExecutor`
+            # internals, but so far this is the best way to stop all these
+            # threads until a public interface is added
             executor._threads.clear()  # type: ignore
             concurrent.futures.thread._threads_queues.clear()  # type: ignore
             raise KeyboardInterrupt
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -2412,8 +2412,8 @@ def paasta_status(args) -> int:
                 return_code, output = future.result()
                 return_codes.append(return_code)
         except KeyboardInterrupt:
-            executor._threads.clear()
-            concurrent.futures.thread._threads_queues.clear()
+            executor._threads.clear()  # type: ignore
+            concurrent.futures.thread._threads_queues.clear()  # type: ignore
 
     return max(return_codes)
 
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -2407,9 +2407,13 @@ def paasta_status(args) -> int:
...
-        for future in concurrent.futures.as_completed(tasks):  # type: ignore
-            return_code, output = future.result()
-            return_codes.append(return_code)
+        try:
+            for future in concurrent.futures.as_completed(tasks):  # type: ignore
+                return_code, output = future.result()
+                return_codes.append(return_code)
+        except KeyboardInterrupt:
+            executor._threads.clear()
+            concurrent.futures.thread._threads_queues.clear()
```
nemacysts, line 2416: should we re-raise here so that the return code isn't set by the `return max(return_codes)` below?

nemacysts, line 2416: unfortunate that we need to dig into internals here :/

we'll probably need to `# type: ignore` here + add a comment about why we had to resort to this

nemacysts (commented):
👍🏻 - thank you! 

this has annoyed me for a while but i haven't had a second to make a similar patch!

jvperrin:
Oh awesome, yeah I had this just happen to me recently and thought it was weird, so thanks for fixing it!

paasta_tools/cli/cmds/status.py
cuza: No clue I got exit code 1 (which is what I was expecting) when I tested.

cuza: There might be a cleaner way of doing this. But no clue....

nemacysts: I think this way is fine (i couldn't find anything better in my quick search) - but we should comment that we're reaching into python internals here 'cause there's no exposed interface to do this otherwise

nemacysts: my understanding is that we should be seeing 130 on a KeyboardInterrupt (e.g., that's what you get if you ^C something like `sleep 1000` and check the exit code)

cuza: For some reason in my head KeyboardInterrupt exit with code 1. Thanks for the catch!!!

paasta_tools/cli/cmds/status.py
```diff
@@ -2407,9 +2407,14 @@ def paasta_status(args) -> int:
...
-            return_code, output = future.result()
-            return_codes.append(return_code)
+        try:
+            for future in concurrent.futures.as_completed(tasks):  # type: ignore
+                return_code, output = future.result()
+                return_codes.append(return_code)
+        except KeyboardInterrupt:
+            executor._threads.clear()  # type: ignore
+            concurrent.futures.thread._threads_queues.clear()  # type: ignore
+            raise KeyboardInterrupt
```
nemacysts, line 2417: we could use a bare `raise` here as well, but this works just the same :)

paasta_tools/cli/cmds/status.py
```diff
@@ -2407,9 +2407,14 @@ def paasta_status(args) -> int:
...
         tasks = [executor.submit(t[0], **t[1]) for t in tasks]  # type: ignore
-        for future in concurrent.futures.as_completed(tasks):  # type: ignore
-            return_code, output = future.result()
-            return_codes.append(return_code)
+        try:
+            for future in concurrent.futures.as_completed(tasks):  # type: ignore
+                return_code, output = future.result()
+                return_codes.append(return_code)
+        except KeyboardInterrupt:
+            executor._threads.clear()  # type: ignore
```
nemacysts, line 2415: nit: giving an explanation for the `type: ignore`/private member access would probably be a good idea
```suggestion
            # ideally we wouldn't need to reach into `ThreadPoolExecutor`
            # internals, but so far this is the best way to stop all these
            # threads until a public interface is added
            executor._threads.clear()  # type: ignore
```

nemacysts (approved):

nemacysts (approved):

PR #3548, opened 2023-02-08
Ensure we pass ns to functions in kubernetes_tools.py - PAASTA-17857

EmanElsaban:
### Purpose
In this PR, we pass namespace to all the functions that requires it, otherwise it will be hardcode to "paasta" namespace.
These changes provide support for arbitrary namespaces in **paasta local-run** and **paasta_metastatus**.

### Testing
passes unit tests

Note: The functions below also must be changed by passing namespace  but I didn't change them since they fall under ``paasta_secrets_sync.py`` work:

- ``create_secret``
- ``create_plaintext_dict_secret``
- ``update_secret``
- ``update_plaintext_dict_secret``
- ``update_kubernetes_secret_signature``
- ``create_kubernetes_secret_signature``
- ``get_kubernetes_secret_signature``

[WIP functions] Still looking into these functions:

- ``list_all_deployments`` -- > in metastatus_lib.py (Done- addressed in second commit)
- ``get_all_pods_cached`` -- > in  metastatus_lib.py (Done- addressed in second commit)

[Ignored function] ``delete_deployment`` : it's being called in ``delete_kubernetes_deployments.py`` and we don't call this file anywhere. We should probably delete this file.

### DIFF
--- a/paasta_tools/paasta_metastatus.py
+++ b/paasta_tools/paasta_metastatus.py
@@ -32,6 +32,7 @@ from paasta_tools import __version__
 from paasta_tools.cli.utils import get_instance_config
 from paasta_tools.kubernetes_tools import is_kubernetes_available
 from paasta_tools.kubernetes_tools import KubeClient
+from paasta_tools.kubernetes_tools import load_kubernetes_service_config_no_cache
 from paasta_tools.marathon_tools import get_marathon_clients
 from paasta_tools.marathon_tools import get_marathon_servers
 from paasta_tools.marathon_tools import MarathonClient
@@ -95,6 +96,7 @@ def parse_args(argv):
     parser.add_argument(
         "-s",
         "--service",
+        dest="service_name",
         help=(
             "Show how many of a given service instance can be run on a cluster slave."
             "Note: This is only effective with -vvv and --instance must also be specified"
@@ -103,6 +105,7 @@ def parse_args(argv):
     parser.add_argument(
         "-i",
         "--instance",
+        dest="instance_name",
         help=(
             "Show how many of a given service instance can be run on a cluster slave."
             "Note: This is only effective with -vvv and --service must also be specified"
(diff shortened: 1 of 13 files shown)


EvanKrall (approved):

EmanElsaban:
For the latest commit, I'm only getting namespace for kubernetes type instances, not sure if this is a good solution. Otherwise, paasta local run will fail

paasta_tools/utils.py
```diff
@@ -418,6 +419,10 @@ def get_instance(self) -> str:
     def get_service(self) -> str:
         return self.service
 
+    def get_namespace(self) -> str:
```
EvanKrall, line 422: We should be able to delete the version from KubernetesDeploymentConfig now that this is here, since it appears to be identical.

EvanKrall (approved):

paasta_tools/paasta_metastatus.py
```diff
@@ -327,7 +333,11 @@ def print_output(argv: Optional[Sequence[str]] = None) -> None:
     args = parse_args(argv)
 
     system_paasta_config = load_system_paasta_config()
-
+    service_config_dict = load_kubernetes_service_config_no_cache(
```
EvanKrall, line 336: this should probably be using `load_kubernetes_service_config`, which will avoid re-parsing YAML if we happen to load the same config again within a short period of time.

paasta_tools/utils.py
```diff
@@ -3301,6 +3306,7 @@ def load_service_instance_config(
             f"Unable to load {instance_type} config for {service}.{instance} as instance name starts with '_'"
         )
     conf_file = f"{instance_type}-{cluster}"
+    print("instance type: ", instance_type)
```
EvanKrall, line 3309: can you clean up this debug print?

EmanElsaban: oppss, forgot to delete it

EvanKrall (approved):

PR #3549, opened 2023-02-09
Add toggle to disable default preStop hook

nemacysts:
Having a default preStop hook will interfere with upcoming plans to make some of our non-mesh workers gracefully shutdown - this should let us disable this hook per-cluster until we're done (after which, we'll just delete this code entirely) and give us an easy way to rollback without needing to pin paasta-tools back.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1326,7 +1326,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
     def get_kubernetes_container_termination_action(
         self,
         mesh_registered: bool,
-    ) -> V1Handler:
+    ) -> Optional[V1Handler]:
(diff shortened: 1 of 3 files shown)


mattmb:
Hmm, maybe this should just be gated behind whether a service is in the mesh or not? Basically if we have the hacheck sidecar then we should have this preStop sleep right? But if not then we shouldn't I think?

nemacysts:
> Hmm, maybe this should just be gated behind whether a service is in the mesh or not? Basically if we have the hacheck sidecar then we should have this preStop sleep right? But if not then we shouldn't I think?

@mattmb oh true - i dunno why I was thinking that the hacheck prestop was good enough! having the hacheck prestop isn't very helpful if the main container has disappeared :p

mattmb:

PR #3556, opened 2023-02-28
allow paasta secret decrypt to use token vault auth

calvinli:
This aligns `paasta secret decrypt` with `paasta local-run` and `paasta secret run` (#3541) both of which don't require a password to do the Vault authentication.

This enables `paasta secret decrypt` to be called (potentially repeatedly) from scripts without getting password prompts.

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -71,6 +71,7 @@ def add_decrypt_subparser(subparsers):
         "decrypt", help="decrypts a single paasta secret"
     )
     _add_common_args(secret_parser_decrypt)
+    _add_vault_auth_args(secret_parser_decrypt)
 
     secret_parser_decrypt.add_argument(
         "-n",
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/secret.py
```diff
@@ -337,7 +341,9 @@ def is_service_folder(soa_dir, service_name):
     return os.path.isfile(os.path.join(soa_dir, service_name, "service.yaml"))
 
 
-def _get_secret_provider_for_service(service_name, cluster_names=None, soa_dir=None):
+def _get_secret_provider_for_service(
+    service_name, cluster_names=None, soa_dir=None, secret_provider_extra_kwargs={}
+):
```
nemacysts, line 346: this is technically fine right now, but it might be safer to refactor this to 
```suggestion
def _get_secret_provider_for_service(
    service_name, cluster_names=None, soa_dir=None, secret_provider_extra_kwargs=None
):
    secret_provider_extra_kwargs = secret_provider_extra_kwargs or {}
```
so that folks don't get surprised in the future by the behavior of mutable default parameters

nemacysts (approved):
looks fine to me, but I'll defer to @danielpops as to whether this is fine security-wise

paasta_tools/cli/cmds/secret.py
calvinli: oh nice catch

calvinli:
happy to talk about the security implications more internally

PR #3557, opened 2023-03-02
Add support for pruning pending pods

edingroot:
## Description
Add an option `--pending-minutes` in `paasta_prune_completed_pods` script to cleanup pending pods older than a certain threshold.

## Test 
Test run in devbox with the following command:  
```bash
KUBECONFIG=/etc/kubernetes/paasta.conf KUBECONTEXT=pnw-devc paasta_prune_completed_pods --namespace paasta-spark -m 60 --error-minutes 60 --pending-minutes 1 --dry-run -v
```
image

### DIFF
--- a/paasta_tools/prune_completed_pods.py
+++ b/paasta_tools/prune_completed_pods.py
@@ -41,6 +41,14 @@ def parse_args():
         type=int,
     )
     parser.add_argument(
+        "-p",
+        "--pending-minutes",
+        help="Minutes since the pod was scheduled. Terminates pods whose phase is Pending based on time since scheduled. "
+        "Including pod status Pending/ContainerCreating/Terminating.",
+        required=False,
+        type=int,
+    )
+    parser.add_argument(
         "--dry-run",
         dest="dry_run",
         action="store_true",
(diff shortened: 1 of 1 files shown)


paasta_tools/prune_completed_pods.py
```diff
@@ -40,6 +40,14 @@ def parse_args():
         required=False,
         type=int,
     )
+    parser.add_argument(
+        "-p",
+        "--pending-minutes",
+        help="Minutes since the pod was scheduled. Terminates pods whose phase is Pending based on time since scheduled. "
```
nemacysts, line 46: nit: can we explicitly concatenate these two strings?

paasta_tools/prune_completed_pods.py
```diff
@@ -40,6 +40,14 @@ def parse_args():
...
+        "Including pod status Pending/ContainerCreating/Terminating.",
```
nemacysts, line 47: TIL that pods shown as `Terminating` in kubectl are actually in the Pending phase!

nemacysts (approved):

paasta_tools/prune_completed_pods.py
edingroot: Done.

paasta_tools/prune_completed_pods.py
edingroot: I also got confused between pod statuses and phases haha

88manpreet (approved):
Nice. 

We should have a similar change for `Running pods` > 36h for Jupyter pods.
I still see them on day to day basis.

nemacysts (approved):

edingroot:
> Nice.
> 
> We should have a similar change for `Running pods` > 36h for Jupyter pods. I still see them on day to day basis.

I was thinking for some running pods which their spark driver is still alive, won’t another executor pod be recreated immediately after we killed them?

PR #3558, opened 2023-03-03
Add options for sending slack notification

edingroot:
## Description
Add options in `paasta_prune_completed_pods` for sending slack notification.

## Test
Test dry-run in the devbox with the following command:  
```bash
KUBECONFIG=/etc/kubernetes/paasta.conf KUBECONTEXT=pnw-devc paasta_prune_completed_pods --namespace paasta-spark -m 60 --error-minutes 60 --pending-minutes 1 --slackme-webhook-id <HOOK_ID> --slack-channel "@chi" -v --dry-run
```

image

In a Kubernetes node:  
image

### DIFF
--- a/paasta_tools/prune_completed_pods.py
+++ b/paasta_tools/prune_completed_pods.py
@@ -1,9 +1,12 @@
 import argparse
+import json
 import logging
 import sys
 from datetime import datetime
+from typing import Any
 from typing import Sequence
 
+import requests
 from dateutil.tz import tzutc
 from kubernetes.client import V1Pod
 
@@ -11,6 +14,8 @@ from paasta_tools.kubernetes_tools import get_all_pods
 from paasta_tools.kubernetes_tools import get_pod_condition
 from paasta_tools.kubernetes_tools import is_pod_completed
 from paasta_tools.kubernetes_tools import KubeClient
+from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.utils import PaastaNotConfiguredError
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


nemacysts (changes requested):
could we have this emit metrics or log to scribe instead of talking to slack?

slack isn't really an observability platform and using `yelp_meteorite` to get this data into sfx/prometheus or using `clog` to get this data into a logging stream should be pretty simple and much more usable (and not require talking to the things outside our infra)

edingroot:
> could we have this emit metrics or log to scribe instead of talking to slack?
> 
> slack isn't really an observability platform and using `yelp_meteorite` to get this data into sfx/prometheus or using `clog` to get this data into a logging stream should be pretty simple and much more usable (and not require talking to the things outside our infra)

I agree we should definitely do this in a more reasonable way. We also have discussion internally of rebuilding spark alert & monitoring infra. However, as Manpreet mentioned in the puppet PR this is a high-priority change for saving cost. I think it would be better to create a ticket and do this later as an improvement?

nemacysts:
> I agree we should definitely do this in a more reasonable way. We also have discussion internally of rebuilding spark alert & monitoring infra. However, as Manpreet mentioned in the puppet PR this is a high-priority change for saving cost. I think we can create a ticket and do this later as an improvement?

@edingroot the actual high-priority change (cleaning up pods) was already merged - adding slack-based monitoring doesn't seem equally high-priority and doesn't seem particularly necessary (we know the script works - having the host reach out to slack just to tell us how many pods it cleaned up doesn't really give us much information). 

If we do still want to keep track of how many pods this is deleting, we'd prefer that the script log to stdout as that should still make it into our standard logging infra

edingroot:
> > I agree we should definitely do this in a more reasonable way. We also have discussion internally of rebuilding spark alert & monitoring infra. However, as Manpreet mentioned in the puppet PR this is a high-priority change for saving cost. I think we can create a ticket and do this later as an improvement?
> 
> @edingroot the actual high-priority change (cleaning up pods) was already merged - adding slack-based monitoring doesn't seem equally high-priority and doesn't seem particularly necessary (we know the script works - having the host reach out to slack just to tell us how many pods it cleaned up doesn't really give us much information).
> 
> If we do still want to keep track of how many pods this is deleting, we'd prefer that the script log to stdout as that should still make it into our standard logging infra

This make sense. I will make the puppet change first. Thank you!

nemacysts:
going to go ahead and close this as we're using our standard logging infra internally now :)

PR #3562, opened 2023-03-08
paasta secret run: subprocess.run -> os.execvpe

calvinli:
This change is to play nicely with [`pgctl`](https://github.com/Yelp/pgctl), which wants a single process (for ease of signal handling): https://pgctl.readthedocs.io/en/latest/user/quickstart.html#writing-playground-services

(`pgctl-poll-ready` also does some file descriptor juggling which doesn't work with `subprocess`)

There's no cleanup in `paasta_tools.cli.cli.main` beyond catching `KeyboardInterrupt` so this change should be transparent (we aren't cutting off any cleanup actions): https://github.com/Yelp/paasta/blob/d336d2a58089eab453f2ffc870ad399653494b47/paasta_tools/cli/cli.py#L257

and I've also been using this locally.

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -15,7 +15,6 @@
 import argparse
 import os
 import re
-import subprocess
 import sys
 
 from service_configuration_lib import DEFAULT_SOA_DIR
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/secret.py
```diff
@@ -482,13 +481,8 @@ def paasta_secret(args):
...
             new_environ["PS1"] = r"(paasta secret run) \$ "
 
-        command_to_run = subprocess.run(
-            args.cmd,
-            shell=False,
-            check=False,
-            env=new_environ,
-        )
-        sys.exit(command_to_run.returncode)
+        # Note: this never returns, it replaces the current process
```
nemacysts, line 484: imo, adding the pgctl context here might be worth doing - otherwise, someone might try to "clean" this up in the future and break things :)

nemacysts (approved):

PR #3563, opened 2023-03-09
Add --assume-pod-identity option to spark-run

Qmando:
Add a new parameter that will get a session based off a web identity token, and uses that session in a spark job. This is entirely just a passthrough parameter to a new feature of service_configuration_lib, added here: https://github.com/Yelp/service_configuration_lib/pull/105/files

`aws_creds` already supports session based credentials, so no other changes are needed.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -17,6 +17,7 @@ from typing import Union
 
 import yaml
 from boto3.exceptions import Boto3Error
+from botocore.exceptions import ClientError
 from service_configuration_lib.spark_config import get_aws_credentials
 from service_configuration_lib.spark_config import get_history_url
 from service_configuration_lib.spark_config import get_resources_requested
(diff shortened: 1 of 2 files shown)


danielpops (approved):

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -395,6 +395,14 @@ def add_subparser(subparsers):
         default=DEFAULT_AWS_REGION,
     )
 
+    aws_group.add_argument(
+        "--assume-pod-identity",
+        help="If running running in the context of a Kubernetes pod with pod identity set, "
```
nemacysts, line 400:
```suggestion
        help="If running in the context of a Kubernetes pod with pod identity set, "
```

nemacysts: afaik: this is functionally a noop right now - paasta spark-run will be called outside the context of a k8s pod (only jupyter will start the driver inside k8s)

PR #3566, opened 2023-03-16
Adding support for multiple namespaces in cleanup_k8s_jobs

cuza:
- this PR makes cleanup_k8s_jobs aware of paasta managed apps in all namespaces
- it creates a big bounce since we added some labels to every pod


paasta_tools/cleanup_kubernetes_jobs.py
```diff
@@ -95,44 +104,118 @@ def alert_state_change(application: Application, soa_dir: str) -> Generator:
...
+        if isinstance(application, DeploymentWrapper):
+            existing_app = application.get_existing_app(kube_client)
+            if (
+                application.kube_deployment.namespace == instance_config.get_namespace()
+                and (
+                    instance_config.config_dict.get("instances", None)
+                    == existing_app.status.ready_replicas
+                    or instance_config.config_dict.get("min_instances", None)
+                    <= existing_app.status.ready_replicas
+                )
```
EvanKrall, line 122: I think I'd use `instance_config.get_instances()` here instead of pulling from the config dicts -- that way you'd get the current HPA numbers.

paasta_tools/cleanup_kubernetes_jobs.py
```diff
@@ -95,44 +104,118 @@ def alert_state_change(application: Application, soa_dir: str) -> Generator:
         raise
 
 
+def instance_is_not_bouncing(
```
nemacysts, line 107: could this function check that only one deployment has a non-zero amount of expected replicas in order to simplify the logic? (perhaps I'm missing something - in which case, maybe we could add some comments here 'cause I'm not quite sure I'm reading this right in the first place :p)

paasta_tools/cleanup_kubernetes_jobs.py
```diff
@@ -95,44 +104,118 @@ def alert_state_change(application: Application, soa_dir: str) -> Generator:
...
+    :param kube_client:
+    :param soa_dir: The SOA config directory to read from
+    :return: list of applications to kill
+    """
+    log.info("Determining apps to be killed")
+
+    applications_to_kill: List[Application] = []
+    for instance_tuple, applications in applications_dict.items():
+        service = instance_tuple[0]
+        instance = instance_tuple[1]
```
nemacysts, line 158: in case you'd like to remove some indexing:
```suggestion
    for (service, instance), applications in applications_dict.items():
```
example of how this works:
```
>>> a = {(1,2): 3}
>>> for (x, y), z in a.items():
...     print(f"{x} {y} {z}")
...
1 2 3
```

paasta_tools/cleanup_kubernetes_jobs.py
```diff
@@ -95,44 +104,118 @@ def alert_state_change(application: Application, soa_dir: str) -> Generator:
...
+        instance_config = load_kubernetes_service_config(
+            cluster=cluster, service=service, instance=instance, soa_dir=soa_dir
+        )
+        if len(applications) == 1:
```
nemacysts, line 162: is it possible to have no applications? and do we want to do anything if that's possible?

paasta_tools/cleanup_kubernetes_jobs.py
```diff
@@ -95,44 +104,118 @@ def alert_state_change(application: Application, soa_dir: str) -> Generator:
...
     :param force: Force the cleanup if we are above the kill_threshold"""
     log.info("Creating KubeClient")
     kube_client = KubeClient()
 
     log.info("Loading running Kubernetes apps")
-    applications = list_namespaced_applications(
-        kube_client, "paasta", APPLICATION_TYPES
-    )
-
+    applications_dict = list_all_applications(kube_client, APPLICATION_TYPES)
```
nemacysts, line 203: could this still be called `applications`? future-us will thank us if this gets refactored to not return a dict :)

(and mypy and co will save us from any mistakes where we treat this as anything but a dict without the variable name referencing the type)

paasta_tools/cleanup_kubernetes_jobs.py
```diff
@@ -95,44 +104,118 @@ def alert_state_change(application: Application, soa_dir: str) -> Generator:
...
+        applications_dict, cluster, valid_services, kube_client, soa_dir
+    )
+
+    log.debug("Running apps: %s" % list(applications_dict))
     log.debug("Valid apps: %s" % valid_services)
     log.debug("Terminating: %s" % applications_to_kill)
     if applications_to_kill:
         above_kill_threshold = float(len(applications_to_kill)) / float(
-            len(applications)
+            len(list(applications_dict))
```
nemacysts, line 218: out of curiosity: why are we converting this to a list? `len()` on a dictionary *should* return the # of keys

paasta_tools/kubernetes/application/tools.py
```diff
@@ -31,45 +34,57 @@ def is_valid_application(deployment: V1Deployment):
...
-            namespace, **kwargs
-        ).items
-        if is_valid_application(deployment)
-    ]
+def list_paasta_managed_deployments(
+    kube_client: KubeClient, **kwargs
+) -> Dict[Tuple[str, str], List[Application]]:
+    deployments: Dict[Tuple[str, str], List[Application]] = {}
+    for deployment in kube_client.deployments.list_deployment_for_all_namespaces(
+        label_selector="paasta.yelp.com/managed", **kwargs
```
nemacysts, line 42: for consistency, it might be good to use the paasta_prefixed function to generate the label selector so that we always create labels the same way everywhere

paasta_tools/kubernetes/application/tools.py
```diff
@@ -31,45 +34,57 @@ def is_valid_application(deployment: V1Deployment):
...
-    return [
-        DeploymentWrapper(deployment)
-        for deployment in kube_client.deployments.list_namespaced_deployment(
-            namespace, **kwargs
-        ).items
-        if is_valid_application(deployment)
-    ]
+def list_paasta_managed_deployments(
+    kube_client: KubeClient, **kwargs
+) -> Dict[Tuple[str, str], List[Application]]:
```
nemacysts, line 39: note for future us: might be nice to create a namedtuple or dataclass for this tuple so that we never misuse it by getting the ordering wrong

paasta_tools/kubernetes/application/tools.py
```diff
@@ -31,45 +34,57 @@ def is_valid_application(deployment: V1Deployment):
...
     """
-    List all applications in the namespace of the types from application_types.
+    List all applications in the cluster of the types from application_types.
     Only applications with complete set of labels are included (See is_valid_application()).
     :param kube_client:
-    :param namespace:
+
     :param application_types:  types of applications
-    :return:
+    :return: A dictionary with (service, instance) as keys and a list of applications for each tuple
```
nemacysts, line 82: suggestion: 
```suggestion
    :return: A mapping from (service, instance) to application
```
is a little more concise (and doesn't repeat the retval type)

paasta_tools/kubernetes/application/tools.py
```diff
@@ -31,45 +34,57 @@ def is_valid_application(deployment: V1Deployment):
...
     """
-    apps = []  # type: ignore
+    apps: Dict[Tuple[str, str], List[Application]] = {}
     for application_type in application_types:
         if application_type == V1Deployment:
-            apps.extend(list_namespaced_deployments(kube_client, namespace))
+            apps = {**apps, **list_paasta_managed_deployments(kube_client)}
```
nemacysts, line 87: can't wait until we're on py3.9 and this can be `apps |= list_paasta_managed_deployments(kube_client)`

paasta_tools/cleanup_kubernetes_jobs.py
cuza: TIL

EvanKrall (changes requested):

paasta_tools/cleanup_kubernetes_jobs.py
cuza: It would be very weird to have no application, it means that we have a fresh service that setup_k8s_job have not created a corresponding k8s workload and is fine skipping delete of it

paasta_tools/kubernetes/application/tools.py
cuza: We do use this (service,instance) tuple a lot....

paasta_tools/kubernetes/application/tools.py
cuza: Can't wait to be able to get a switch/case in python!!!!!!

paasta_tools/cleanup_kubernetes_jobs.py
```diff
@@ -95,44 +104,118 @@ def alert_state_change(application: Application, soa_dir: str) -> Generator:
...
+            cluster=cluster, service=service, instance=instance, soa_dir=soa_dir
+        )
+        if len(applications) == 1:
+            application = applications[0]
+            if (
+                (service, instance) not in valid_services
+                or application.kube_deployment.namespace
+                != instance_config.get_namespace()
+            ):
+                applications_to_kill.append(application)
```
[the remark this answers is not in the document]
cuza: true I  didn't think ok a case of cleanup running before setup. we could unifi the conditions

EvanKrall (changes requested):

paasta_tools/cleanup_kubernetes_jobs.py
```diff
@@ -95,44 +103,107 @@ def alert_state_change(application: Application, soa_dir: str) -> Generator:
...
+    :param applications: a list of all deployments or stateful sets on the cluster that match the service
+     and instance of provided instance_config
+    """
+    for application in applications:
+        if isinstance(application, DeploymentWrapper):
+            existing_app = application.item
+            if existing_app.metadata.namespace == instance_config.get_namespace() and (
+                instance_config.config_dict.get("instances", None)
+                == existing_app.status.ready_replicas
+                or instance_config.get_instances() <= existing_app.status.ready_replicas
```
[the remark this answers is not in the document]
cuza: ooh nice TIL

EvanKrall (approved):

PR #3568, opened 2023-03-17
Support shared secrets in secrets for owner team

danielpops:
## Problem
The secrets for owner team implementation uses the wrong key for referencing shared secrets

## Solution
Copied [the same pattern used elsewhere](https://github.com/Yelp/paasta/blob/2e1cafdc687f231eca07388f468bf4b00ebb95cd/paasta_tools/kubernetes_tools.py#L2083) to reference the correct secret name when the secret is a shared one

## Testing
Confirmed this works with a service that uses shared secret and has `secrets_for_owner_team: true`
```
@@ -10,5 +10,6 @@ main:
+    PAASTA_SECRET_MY_SHARED_SECRET: SHARED_SECRET(test-shared)
```

```
$ paasta local-run --service the-service --cluster the-cluster -i main --interactive --pull
Re-executing paasta local-run --pull with sudo..
[...]
nobody@the-host:/code$ env | grep SHARED
PAASTA_SECRET_MY_SHARED_SECRET=SHARED_SECRET_TEST
nobody@the-host:/code$ exit
```

Without the fix in this PR:
```
$ paasta local-run --service the-service --cluster the-cluster -i main --interactive --pull
Re-executing paasta local-run --pull with sudo..
[...]
Failed to retrieve kubernetes secrets with ApiException: (404)
Reason: Not Found
HTTP response headers: HTTPHeaderDict({'Audit-Id': 'deadbeef-dead-beef-dead-beefdeadbeef', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'X-Kubernetes-Pf-Flowschema-Uid': 'deadbeef-dead-beef-dead-beefdeadbeef', 'X-Kubernetes-Pf-Prioritylevel-Uid': 'deadbeef-dead-beef-dead-beefdeadbeef', 'Date': 'Fri, 17 Mar 2023 19:29:57 GMT', 'Content-Length': '264'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"secrets \"paasta-secret-the--service-test-shared\" not found","reason":"NotFound","details":{"name":"paasta-secret-the--service-test-shared","kind":"secrets"},"code":404}

If you don't need the secrets for local-run, you can add --skip-secrets
```

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -3635,7 +3635,7 @@ def get_kubernetes_secret_env_variables(
             decrypted_secrets[k] = str(
                 get_kubernetes_secret(
                     kube_client,
-                    service_name,
+                    SHARED_SECRET_SERVICE if is_shared_secret(v) else service_name,
                     secret_name,
                     decode=True,
                     namespace=namespace,
--- a/tests/test_kubernetes_tools.py
+++ b/tests/test_kubernetes_tools.py
@@ -4243,11 +4243,12 @@ def test_get_kubernetes_secret_env_variables():
             "MY": "aaa",
             "SECRET_NAME1": "SECRET(SECRET_NAME1)",
             "SECRET_NAME2": "SECRET(SECRET_NAME2)",
+            "SHARED_SECRET1": "SHARED_SECRET(SHARED_SECRET1)",
         }
 
         mock_is_secret_ref.side_effect = lambda val: "SECRET" in val
-        mock_get_ref.side_effect = ["SECRET_NAME1", "SECRET_NAME2"]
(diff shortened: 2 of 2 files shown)


nemacysts (approved):
might be nice to add a test so that we don't accidentally regress in the future - but this is a pretty common pattern so i'd still ship without one :)

piax93 (approved):

danielpops:
> might be nice to add a test so that we don't accidentally regress in the future - but this is a pretty common pattern so i'd still ship without one :)

good call. Doesn't look like there is currently any test that explicitly captures secrets for teams w.r.t. environment variables, so i'll need to add one for this specific case

danielpops:
Test added, all checks passed.

PR #3569, opened 2023-03-21
Add AWS_SECURITY_TOKEN to local-run assume-role env vars

Qmando:
This change supports older boto(2) clients that use AWS_SECURITY_TOKEN. While this has been deprecated for some time, I don't see any reason why we should support it.

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -76,6 +76,7 @@ class AWSSessionCreds(TypedDict):
     AWS_ACCESS_KEY_ID: str
     AWS_SECRET_ACCESS_KEY: str
     AWS_SESSION_TOKEN: str
+    AWS_SECURITY_TOKEN: str
 
 
 def parse_date(date_string):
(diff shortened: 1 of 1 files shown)


danielpops (approved):

paasta_tools/cli/cmds/local_run.py
```diff
@@ -778,6 +778,7 @@ def assume_aws_role(
         "AWS_ACCESS_KEY_ID": cmd_output["AccessKeyId"],
         "AWS_SECRET_ACCESS_KEY": cmd_output["SecretAccessKey"],
         "AWS_SESSION_TOKEN": cmd_output["SessionToken"],
+        "AWS_SECURITY_TOKEN": cmd_output["SessionToken"],
```
nemacysts, line 781: we'll need to add this key to the AWSSessionCreds TypedDict

nemacysts (approved):
i'm fine with this in general - but on a more philosophical level: should this be a lever to get folks to modernize and move off of a long-deprecated library?

PR #3575, opened 2023-03-29
Split NOT_RUNNING state into different states for different underlying conditions.

EvanKrall:
I don't find it particularly helpful to have 4 different conditions end up with the same `Not Running` message, so I've distinguished between them. This should help with after-the-fact debugging in situations like https://jira.yelpcorp.com/browse/COMPINFRA-2270 where we just have `paasta status` / `paasta mark-for-deployment --wait-for-deployment` output.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1503,8 +1503,16 @@ def get_version_table_entry(
 class ReplicaState(Enum):
     # Order will be preserved in count summary
     RUNNING = "Healthy", PaastaColors.green
+
     UNREACHABLE = "Unreachable", PaastaColors.red
-    NOT_RUNNING = "Not Running", PaastaColors.red
+    EVICTED = "Evicted", PaastaColors.red
+    ALL_CONTAINERS_WAITING = "All Containers Waiting", PaastaColors.red
+    FAILED = "Failed", PaastaColors.red
+    MAIN_CONTAINER_NOT_RUNNING = "Main Container Not Running", PaastaColors.red
+    NO_CONTAINERS_YET = "No Containers Yet", PaastaColors.red
+    NOT_READY = "Not Ready", PaastaColors.red
+    SOME_CONTAINERS_WAITING = "Some Containers Waiting", PaastaColors.red
+
     WARNING = "Warning", PaastaColors.yellow
     UNSCHEDULED = "Unscheduled", PaastaColors.yellow
     STARTING = "Starting", PaastaColors.yellow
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -1465,7 +1473,7 @@ class ReplicaState(Enum):
     UNKNOWN = "Unknown", PaastaColors.yellow
 
     def is_unhealthy(self):
-        return self.name in ["UNREACHABLE", "NOT_RUNNING"]
+        return self.color == PaastaColors.red
```
nemacysts, line 1476: clever :p 

i'm a little worried we'll accidentally break this compared to checking if an element is in an `UNHEALTHY_REPLICA_STATES` set - but that seems kinda unlikely

paasta_tools/cli/cmds/status.py
```diff
@@ -1516,17 +1524,21 @@ def get_replica_state(pod: KubernetesPodV2) -> ReplicaState:
...
     elif phase is None or not pod.scheduled:
         state = ReplicaState.UNSCHEDULED
     elif pod.delete_timestamp:
         state = ReplicaState.TERMINATING
     elif phase == "Pending":
-        if not pod.containers or all([c.state == "waiting" for c in pod.containers]):
-            state = ReplicaState.STARTING
+        if not pod.containers:
+            state = ReplicaState.NO_CONTAINERS_YET
+        elif all([c.state == "waiting" for c in pod.containers]):
```
nemacysts, line 1538: just double-checking: the state is lowercase in the k8s api response?

nemacysts (approved):

paasta_tools/cli/cmds/status.py
EvanKrall: Interesting, [the docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-states) indicate it should be `Waiting`.

mattmb:
I'm not quite brave enough to just merge @EvanKrall @nemacysts but I won't close this one since it looks like a nice UX improvement and just needs someone to shepherd it out

paasta_tools/cli/cmds/status.py
nemacysts: @EvanKrall perhaps should we `c.state.lower()` just to be safe here?

or try to catch a pod in this state and see if the original code was right? :p

paasta_tools/cli/cmds/status.py
nemacysts: maybe we add a docstring in the ReplicaState enum to make it clear that the color is quite important?

nemacysts (approved):

PR #3576, opened 2023-03-29
Set bounce_margin_factor default to 0.95

EvanKrall:
In COMPINFRA-2270, several service owners are complaining about stuck deployments because of individual pods that never become healthy.

This change would make it so any service with 20-39 containers would ignore a single stuck pod; 40-59 would ignore 2, etc.

From what I can tell from the [code that manages PodDisruptionBudgets](https://github.com/Yelp/paasta/blob/02b165ac877700b28f98653f8ad2b30f0d73f339/paasta_tools/kubernetes/application/controller_wrappers.py#L147), this won't affect PDBs - anything that doesn't have `bounce_margin_factor` set gets the default PDB from SystemPaastaConfig.


nemacysts (approved):

nemacysts:
(might be good to add a description to this PR before merging)

tests/test_kubernetes_tools.py
```diff
@@ -4170,7 +4170,13 @@ def test_warning_big_bounce_routable_pod():
             job_config.format_kubernetes_app().spec.template.metadata.labels[
                 "paasta.yelp.com/config_sha"
             ]
+<<<<<<< HEAD
```
nemacysts, line 4173: looks like something funky happened here with the merge markers?

tests/test_kubernetes_tools.py
```diff
@@ -4170,7 +4170,13 @@ def test_warning_big_bounce_routable_pod():
...
             == "configa5abd828"
+||||||| parent of 0545b824e (Set bounce_margin_factor default to 0.95)
+            == "config295207ad"
+=======
+            == "config4217696e"
+>>>>>>> 0545b824e (Set bounce_margin_factor default to 0.95)
```
EvanKrall, line 4179: lol

nemacysts (approved):
(we should hold off on merging this until we finish the karpenter upgrade - so maybe tomorrow?)

nemacysts:
(should be safe to merge now, i think - if we pull master the one test failure should go away too)

nemacysts:
@EvanKrall i've merged master into this to get a green test run - probably good to merge after that imo :)

bobokun (approved):

bobokun (approved):

jfongatyelp (approved):

PR #3577, opened 2023-03-29
Have setup_kubernetes_job prioritize services whose git_shas are chan…

EvanKrall:
…ging, so that real bounces take priority over big bounces.


paasta_tools/setup_kubernetes_job.py
```diff
@@ -223,6 +230,31 @@ def setup_kube_deployments(
...
+                    app.kube_deployment.instance,
+                    app.kube_deployment.namespace,
+                ),
+                [],
+            ):
+                return 1
+            else:
+                return 0
+        else:
+            return 2
```
nemacysts, line 254: just to confirm that my brain isn't too broken: this is to have us handle "broken" apps after everything else so that we don't use any of our rate-limit on 'em?

nemacysts (approved):

paasta_tools/setup_kubernetes_job.py
```diff
@@ -213,6 +214,12 @@ def setup_kube_deployments(
             for deployment in existing_kube_deployments
         }
 
+        existing_git_shas: Dict[Tuple[str, str, str], List[str]] = {}
+        for deployment in existing_kube_deployments:
+            existing_git_shas.setdefault(
+                (deployment.service, deployment.instance, deployment.namespace), []
+            ).append(deployment.git_sha)
```
jfongatyelp, line 221: Thoughts on using DeploymentVersion (aka git sha+ image_version) instead of just git sha? We don't seem to have many using this right now but if we did, we'd probably want to prioritize those rebuilds over big bounces too?

paasta_tools/setup_kubernetes_job.py
EvanKrall: yep, exactly

EvanKrall: Though actually, the body of the for loop is wrapped in `if app:`, so broken apps don't count against the rate limit anyway. This is mostly here so attribute accesses on `app` don't fail.

I guess I should refactor this to use `if app:` instead of `if ok:`, to mirror the main for loop.

paasta_tools/setup_kubernetes_job.py
EvanKrall: Probably easy enough.

paasta_tools/kubernetes_tools.py
```diff
@@ -227,6 +227,10 @@ class KubeDeployment(NamedTuple):
     namespace: str
     replicas: Optional[int]
 
+    @property
+    def deployment_version(self):
+        return DeploymentVersion(self.git_sha, self.image_version)
+
```
jfongatyelp, line 233: <3

paasta_tools/setup_kubernetes_job.py
```diff
@@ -213,6 +214,12 @@ def setup_kube_deployments(
             for deployment in existing_kube_deployments
         }
 
+        existing_deployment_versions: Dict[Tuple[str, str, str], List[str]] = {}
```
jfongatyelp, line 217: Is the value actually a `List[str]` or `List[DeploymentVersion]`?

jfongatyelp (approved):

paasta_tools/setup_kubernetes_job.py
EvanKrall: I added comments to help make this more readable.

mattmb:
This also looks a bit too good/easy to close (conflict is minor). But I wouldn't want to just press merge since someone needs to validate everything is okay after rolling out.

jfongatyelp (approved):

paasta_tools/setup_kubernetes_job.py
```diff
@@ -254,6 +264,39 @@ def setup_kube_deployments(
         else (_, None)
         for _, service_instance in service_instance_configs_list
     ]
+
+    def sort_key(ok_app: Tuple[bool, Optional[Application]]) -> int:
+        """This will return 1 if the desired deployment_version matches an existing deployment_version, and 0
+        otherwise. This will cause applications that need a new deployment_version to be handled first.
```
nemacysts, line 270: on another read: the `0 otherwise` bit isn't 100% accurate since we return 2 for broken apps

nemacysts (approved):

paasta_tools/setup_kubernetes_job.py
nemacysts:
```suggestion
        """This will return 1 if the desired deployment_version matches an existing deployment_version, 2 if the deployment is unhealthy and 0
        otherwise. This will cause applications that need a new deployment_version to be handled first.
```

nemacysts: (probably want to make that change in your editor since i did this in GH and almost certainly have an overly long line)

nemacysts (approved):
small docstring nit, but lgtm!

PR #3579, opened 2023-03-30
Make resource minimums in k8s schemas inclusive

nemacysts:
We have a couple instances in our code where we assume that these limits are inclusive, but recent changes to metric collection have shown that this is not true :)

Rather than try to fix every location where we made this incorrect assumption - let's just make that assumption reality.

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -105,13 +105,13 @@
                 "mem": {
                     "type": "number",
                     "minimum": 32,
-                    "exclusiveMinimum": true,
+                    "exclusiveMinimum": false,
                     "default": 1024
                 },
                 "disk": {
                     "type": "number",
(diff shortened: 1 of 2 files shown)


bobokun (approved):

jfongatyelp (approved):

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -105,13 +105,13 @@
...
                     "minimum": 32,
-                    "exclusiveMinimum": true,
+                    "exclusiveMinimum": false,
                     "default": 1024
                 },
                 "disk": {
                     "type": "number",
                     "minimum": 0,
-                    "exclusiveMinimum": true,
+                    "exclusiveMinimum": false,
```
nemacysts, line 114:
```suggestion
                    "exclusiveMinimum": true,
```

EvanKrall (approved):

PR #3580, opened 2023-03-31
[paasta secret run]  pass -y through to get_instance_config

calvinli:
Fixes a bug found by @danielpops where `paasta secret run -y ...` would not reflect local changes made to an instance's environment variables (would always use the environment config from system yelpsoa-configs).


danielpops (approved):

nemacysts (approved):
classic :p

danielpops:
🚢

PR #3583, opened 2023-04-06
Make paasta secrets for teams decrypt available for multiple instance types and namespaces

iiliescu2:
k8s secrets decryption used to rely on the default paasta namespace. Adding some functionality to trace back from a secret name to the instance types where it is used, followed by k8s namespaces that they are stored under.
Keeping a fallback to default, in case the mapping fails.

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -20,9 +20,9 @@ import sys
 from service_configuration_lib import DEFAULT_SOA_DIR
 
 from paasta_tools.cli.utils import get_instance_config
+from paasta_tools.cli.utils import get_namespace_for_secret
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_instances
-from paasta_tools.cli.utils import get_namespace_for_secret
 from paasta_tools.kubernetes_tools import get_kubernetes_secret
 from paasta_tools.kubernetes_tools import KUBE_CONFIG_USER_PATH
 from paasta_tools.kubernetes_tools import KubeClient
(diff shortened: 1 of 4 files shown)


piax93 (commented):
Implementation looks good to me, but seems like there's a test failing.

paasta_tools/cli/utils.py
```diff
@@ -843,6 +847,36 @@ def get_instance_config(
     )
 
 
+def get_namespace_for_secret(
+    service: str, cluster: str, secret_name: str, soa_dir: str = DEFAULT_SOA_DIR
+) -> Set:
```
nemacysts, line 852: nit:
```suggestion
) -> Set[str]:
```

nemacysts (approved):

piax93 (approved):

paasta_tools/cli/cmds/secret.py
```diff
@@ -401,7 +402,19 @@ def paasta_secret(args):
             sys.exit(1)
 
         kube_client = KubeClient(config_file=KUBE_CONFIG_USER_PATH, context=clusters[0])
-        print(get_kubernetes_secret(kube_client, service, args.secret_name))
+
+        secret_to_k8s_mapping = get_namespace_for_secret(
+            service, clusters[0], args.secret_name, args.yelpsoa_config_root
+        )
+        if secret_to_k8s_mapping:
+            # assuming that if multiple namespaces, they should be the same
```
danielpops, line 410: My only small nit is that if one of the namespaces is in some experimental namespace or something, it may have an out of date or incorrect value. I don't think there's any change to be made here, just mentioning it because I tend to think of pathological things like this :)

danielpops (approved):

paasta_tools/cli/utils.py
```diff
@@ -843,6 +847,36 @@ def get_instance_config(
     )
 
 
+def get_namespace_for_secret(
```
nemacysts, line 850: thinking about this a little more: thoughts on renaming this to reflect that a collection will be returned?
```suggestion
def get_namespaces_for_secret(
```

nemacysts (approved):

paasta_tools/cli/cmds/secret.py
iiliescu2: Hmm the only other thing I can think of that we can do here is to create a priority_list = ["paasta", "tron", ..] to first check if these predefined non-experimental values are present in the secret_to_k8s_mapping and use them, otherwise fallback on .pop(). Do you think that would be useful in this case? If so, should it contain the set of values as they are now here: https://sourcegraph.yelpcorp.com/Yelp/paasta/-/blob/paasta_tools/utils.py?L140:1 ?

danielpops: Yeah that's a great idea. 🚢

paasta_tools/cli/utils.py
```diff
@@ -877,6 +877,32 @@ def get_namespace_for_secret(
...
+    if namespaces_count == 1:
+        return namespaces.pop()
+
+    prioritized_k8s_namespaces = [
+        "paasta",
+        "tron",
+        "paasta-flinks",
+        "paasta-cassandraclusters",
+        "paasta-kafkaclusters",
+        "paasta-nrtsearchservices",
```
piax93, line 895: It's not great to have these hardcoded here, it'd be better to load from some configuration.

Also know that services are eventually moving to individual namespaces prefixed with `paastasvc-`, so it'll need changing (PAASTA-17769).

piax93 (approved):

PR #3584, opened 2023-04-06
Allow spark-run to assume aws roles

Qmando:
This depends on the changes in https://github.com/Yelp/service_configuration_lib/pull/111. As such, tests are currently failing due to autospec on service_configuration_lib's get_aws_credentials function. This is NOT ready to ship until a new version of that library is deployed.

This just adds some new parameters that get passed through directly to the get_aws_credentials function.


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -395,6 +395,18 @@ def add_subparser(subparsers):
...
     )
 
+    aws_group.add_argument(
+        "--assume-aws-role",
+        help="Takes an ARN and will attempt to create a session",
+    )
+
+    aws_group.add_argument(
+        "--aws-role-duration",
+        help="Duration in seconds for the role if --assume-aws-role provided",
```
nemacysts, line 405: should we document that the max is 43200s/12h here?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -395,6 +395,18 @@ def add_subparser(subparsers):
         default=DEFAULT_AWS_REGION,
     )
 
+    aws_group.add_argument(
+        "--assume-aws-role",
+        help="Takes an ARN and will attempt to create a session",
```
nemacysts, line 400: thoughts on expanding on this a tad with something like
```suggestion
        help="Takes an AWS IAM role ARN and attempts to create a session",
```

nemacysts (approved):

PR #3585, opened 2023-04-10
Show status of apps in other namespaces, in case migration to a new namespace gets stuck.

EvanKrall:
example output with new API (which returns namespace for each version): https://fluffy.yelpcorp.com/i/JV5mGznpQ8kZZlSGr1J5r4VJq68gNc0K.html (see lines 31, 35, 39).

example output with old API (which doesn't include namespace, and doesn't include versions in the old namespace): https://fluffy.yelpcorp.com/i/r5zMQhHCfdssxNsgkFFJZFHbnsrwqq6R.html

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1365,10 +1365,10 @@ def get_versions_table(
     else:
         versions = sorted(versions, key=lambda x: x.create_timestamp, reverse=True)
         config_shas = {v.config_sha for v in versions}
-        if len(config_shas) > 1:
-            show_config_sha = True
-        else:
-            show_config_sha = False
+        show_config_sha = len(config_shas) > 1
+
+        namespaces = {v.namespace for v in versions}
+        show_namespace = len(namespaces) > 1
 
         table: List[str] = []
         table.extend(
@@ -1379,6 +1379,7 @@ def get_versions_table(
                 cluster,
                 version_name_suffix="new",
                 show_config_sha=show_config_sha,
+                show_namespace=show_namespace,
                 verbose=verbose,
             )
         )
@@ -1391,6 +1392,7 @@ def get_versions_table(
                     cluster,
                     version_name_suffix="old",
                     show_config_sha=show_config_sha,
+                    show_namespace=show_namespace,
                     verbose=verbose,
                 )
             )
@@ -1404,6 +1406,7 @@ def get_version_table_entry(
     cluster: str,
     version_name_suffix: str = None,
     show_config_sha: bool = False,
+    show_namespace: bool = False,
     verbose: int = 0,
 ) -> List[str]:
     version_name = version.git_sha[:8]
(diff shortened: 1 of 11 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -1365,10 +1365,10 @@ def get_versions_table(
...
         versions = sorted(versions, key=lambda x: x.create_timestamp, reverse=True)
         config_shas = {v.config_sha for v in versions}
-        if len(config_shas) > 1:
-            show_config_sha = True
-        else:
-            show_config_sha = False
+        show_config_sha = len(config_shas) > 1
+
+        namespaces = {v.namespace for v in versions}
+        show_namespace = len(namespaces) > 1
```
cuza, line 1371: do we just want to show namespaces during bounces?? We(CI) use kubectl a lot and now any instance might be on an arbitrary namespace

EvanKrall: `get_version_table_entry` always shows the namespace regardless of `show_namespace` if you do multiple -v's, e.g. `paasta status -vv`

paasta_tools/delete_kubernetes_deployments.py
```diff
@@ -61,8 +61,8 @@ def get_deployment_names_from_list(service_instance_list):
     return app_names
 
 
-def main() -> None:
-    args = parse_args()
+def main(args=None) -> None:
+    args = parse_args(args=args)
```
nemacysts, line 65: are the `args` changes just to make testing easier?

paasta_tools/instance/kubernetes.py
```diff
@@ -528,6 +529,45 @@ def bounce_status(
...
+        ret.extend(await coro)
+
+    return ret
+
+
+def find_all_relevant_namespaces(
+    service,
+    instance,
+    kube_client,
+    job_config,
```
nemacysts, line 560: nit: 
```suggestion
    service: str,
    instance: str,
    kube_client: kubernetes_tools.KubeClient,
    job_config: KubernetesDeploymentConfig,
```

(i'm guessing the type of `job_config` - might need to fix that one up)

paasta_tools/instance/kubernetes.py
```diff
@@ -707,15 +751,24 @@ async def get_versions_for_replicasets(
...
     pod_status_by_replicaset_task: "asyncio.Future[Mapping[str, Sequence[asyncio.Future[Dict[str, Any]]]]]",
 ) -> List[KubernetesVersionDict]:
-    replicaset_list = await kubernetes_tools.replicasets_for_service_instance(
-        service=service,
-        instance=instance,
-        kube_client=kube_client,
-        namespace=namespace,
-    )
+
+    replicaset_list: List[V1ReplicaSet] = []
```
nemacysts, line 758: normally i'd suggest that we should call this `replicasets` to not duplicate the type in both the typehint + the variable name, but this was already existing :p

paasta_tools/kubernetes_tools.py
```diff
@@ -2719,6 +2719,17 @@ def list_matching_deployments(
     )
 
 
+def list_matching_deployments_in_all_namespaces(
+    service: str,
+    instance: str,
+    kube_client: KubeClient,
+) -> Sequence[KubeDeployment]:
```
nemacysts, line 2726: should this be 
```suggestion
) -> Set[KubeDeployment]:
```
or are we sticking with the more abstract type to be consistent with the rest of this file?

EvanKrall: `list_deployments_in_all_namespaces` returns a Sequence so I just stuck with that. I could re-type `list_deployments_in_all_namespaces` to return a `List[KubeDeployment]`

nemacysts (approved):

PR #3586, opened 2023-04-10
Use properly formatted secret name in other namespaces + refactor to …

EvanKrall:
…use get_kubernetes_secret_name everywhere we need to format the secret name

This is what's wrong with clusterman.drainer in kubestage.


cuza (approved):

paasta_tools/kubernetes_tools.py
```diff
@@ -1398,7 +1398,11 @@ def get_pod_volumes(
                 V1Volume(
                     name=self.get_secret_volume_name(secret_volume),
                     secret=V1SecretVolumeSource(
-                        secret_name=f"paasta-secret-{service}-{sanitised_secret}",
+                        secret_name=get_kubernetes_secret_name(
+                            self.service,
```
nurdann, line 1402: should it be without `self`?

PR #3587, opened 2023-04-13
When determining which namespaces to query in the status API endpoint…

EvanKrall:
…, only look at namespaces which have the paasta.yelp.com/managed=true label.

This avoids a permissions issue introduced in #3585, because we don't give paasta-api permissions to list deployments/statefulsets in all namespaces.

Any namespace with the label `paasta.yelp.com/managed=true` should have been created by `ensure_namespace`, which will [make sure we have the appropriate role binding](https://github.com/Yelp/paasta/blob/a53c8982bb874d309347174c5c4ef5e4a73c62ee/paasta_tools/kubernetes_tools.py#L2320)

### DIFF
--- a/paasta_tools/instance/kubernetes.py
+++ b/paasta_tools/instance/kubernetes.py
@@ -38,6 +38,7 @@ from paasta_tools.instance.hpa_metrics_parser import HPAMetricsParser
 from paasta_tools.kubernetes_tools import get_pod_event_messages
 from paasta_tools.kubernetes_tools import get_tail_lines_for_kubernetes_container
 from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
+from paasta_tools.kubernetes_tools import paasta_prefixed
 from paasta_tools.long_running_service_tools import LongRunningServiceConfig
 from paasta_tools.long_running_service_tools import ServiceNamespaceConfig
 from paasta_tools.smartstack_tools import KubeSmartstackEnvoyReplicationChecker
@@ -563,7 +564,7 @@ def find_all_relevant_namespaces(
         deployment.namespace
         for deployment in kubernetes_tools.list_deployments_in_managed_namespaces(
             kube_client=kube_client,
-            label_selector=f"paasta.yelp.com/service={service},paasta.yelp.com/instance={instance}",
+            label_selector=f"{paasta_prefixed('service')}={service},{paasta_prefixed('instance')}={instance}",
         )
     }
 
--- a/paasta_tools/instance/kubernetes.py
+++ b/paasta_tools/instance/kubernetes.py
@@ -561,7 +561,7 @@ def find_all_relevant_namespaces(
 ) -> Set[str]:
     return {job_config.get_kubernetes_namespace()} | {
         deployment.namespace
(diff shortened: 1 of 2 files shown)


paasta_tools/instance/kubernetes.py
```diff
@@ -561,7 +561,7 @@ def find_all_relevant_namespaces(
 ) -> Set[str]:
     return {job_config.get_kubernetes_namespace()} | {
         deployment.namespace
-        for deployment in kubernetes_tools.list_deployments_in_all_namespaces(
+        for deployment in kubernetes_tools.list_deployments_in_managed_namespaces(
             kube_client=kube_client,
             label_selector=f"paasta.yelp.com/service={service},paasta.yelp.com/instance={instance}",
```
cuza, line 566: shouldn't we use `paasta_prefixed` from `paasta_tools.kubernetes_tools` here?

EvanKrall: yeah, probably

jfongatyelp (approved):

cuza (approved):

cuza (approved):

paasta_tools/kubernetes_tools.py
```diff
@@ -2405,6 +2413,21 @@ def list_deployments(
...
+    label_selector: str,
+) -> List[KubeDeployment]:
+    return [
+        d
+        for namespace in get_all_managed_namespaces(kube_client)
+        for d in list_deployments(
+            kube_client=kube_client,
+            label_selector=label_selector,
+            namespace=namespace,
+        )
```
jfongatyelp, line 2427: It occurs to me we might end up hitting paasta api 500s again if someone ever accidentally adds a new namespace with `paasta.yelp.com/managed=true` via flux (or otherwise) and forgets the paasta-api-service-per-namespace binding...
Do you think we could still do the `try: .. except 403: pass` option you mentioned previously?

paasta_tools/kubernetes_tools.py
```diff
@@ -2405,6 +2413,29 @@ def list_deployments(
...
+                    label_selector=label_selector,
+                    namespace=namespace,
+                )
+            )
+        except ApiException as exc:
+            log.error(
+                f"Error fetching deployments from namespace {namespace}: "
+                f"status: {exc.status}, reason: {exc.reason}."
+            )
+            pass
```
nemacysts, line 2435:
```suggestion
```

(super-nit)

paasta_tools/kubernetes_tools.py
```diff
@@ -2405,6 +2413,29 @@ def list_deployments(
...
+                    kube_client=kube_client,
+                    label_selector=label_selector,
+                    namespace=namespace,
+                )
+            )
+        except ApiException as exc:
+            log.error(
+                f"Error fetching deployments from namespace {namespace}: "
+                f"status: {exc.status}, reason: {exc.reason}."
+            )
```
nemacysts, line 2434: I assume we're not using log.exception here as we're generally not actually too interested in the full traceback here?

nemacysts (approved):

nemacysts (approved):

jfongatyelp (approved):

PR #3593, opened 2023-04-28
Revert "Revert "[PAASTA-17844] Mount Vault KV secrets as k8s secret v…

nurdann:
…olumes" (#3590)"

This reverts commit 05727ce83fe03658fb0662baf90a49bbe43bdc48.

Reverting on top of #3586

## Changes
- Do not attempt to create secret if signature not found for Paasta secret; answers KeyError messages
- Consolidate fetching PaaSTA secret names from a single function, i.e. `get_paasta_secret_name`

## TODO
Was not able to determine/reproduce the cause of config sha changes, so will need to test in an isolated environment stageg or stagef


cuza (approved):

nemacysts:
@nurdann can you continue testing with a locally built .deb on our test clusters? it's preferable to test that way over testing with code merged into master in case we need to make a paasta release (and we generally know that there appears to be a bug with how we're dealing with secret volumes)

PR #3596, opened 2023-05-01
Unrevert Mounting Vault keys as K8s secrets

nurdann:
Changes should be same as in https://github.com/Yelp/paasta/pull/3595 . Creating PR from fork, so I can test it locally and make changes if necessary

> Reverts Yelp/paasta#3594

> ## Changes on top of https://github.com/Yelp/paasta/pull/3536 
> - Added check so that no paasta secret is created if no signature found in soa config
> - Merged conflict changes from https://github.com/Yelp/paasta/pull/3586


paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -286,24 +286,26 @@ def sync_secrets(
...
-                            # If signatures does not match, it'll sys.exit(1)
-                            secret_provider.decrypt_secret_raw(secret)
-                        ).decode("utf-8")
-                    },
-                    secret_signature=secret_signature,
-                    kube_client=kube_client,
-                    namespace=namespace,
-                )
+
+                if secret_signature:
```
nemacysts, line 290: should all the other secret types have a similar check?

nurdann: No, because this is an expected signature that will be compared against k8s configmap signature, and boto/crypto-keys calculate their signature without calling API.

nurdann: What I meant is that when reading Paasta secret, you can't assume that secret for the ecosystem exists. While iterating boto/crypto-keys it should be there

paasta_tools/kubernetes_tools.py
```diff
@@ -1591,11 +1593,26 @@ def get_crypto_secret_name(self) -> str:
             self.get_namespace(), "crypto-key", self.get_service(), self.get_instance()
         )
 
-    def get_boto_secret_signature_name(self) -> str:
+    def get_boto_new_secret_signature_name(self) -> str:
+        """
+        Keeping in case we will want to canonize secret naming
+        """
```
nemacysts, line 1605: probably easier maintenance-wise to bring this back whenever do we want to go about doing this rather than leaving unused code in `master` :)

paasta_tools/kubernetes_tools.py
```diff
@@ -1584,8 +1584,14 @@ def get_volume_mounts(
...
 
     def get_boto_secret_name(self) -> str:
-        return _get_secret_name(
-            self.get_namespace(), "boto-key", self.get_service(), self.get_instance()
+        """
+        Namespace is ignored so that there are no bounces with existing boto_keys secrets
+        """
+        # return _get_secret_name(
+        #     self.get_namespace(), "boto-key", self.get_service(), self.get_instance()
+        # )
```
nemacysts, line 1592: same as with `get_boto_new_secret_signature_name`: we can grab this line from this PR/git history rather than leaving commented out code in `master`

paasta_tools/kubernetes_tools.py
```diff
@@ -1604,14 +1610,13 @@ def get_boto_new_secret_signature_name(self) -> str:
     def get_boto_secret_signature_name(self) -> str:
         """
         Keep the following signature naming convention so that bounces do not happen because boto_keys configmap signatures already exist, see PAASTA-17910
-        Most services reside in paasta namespace so we can pull it dynamically
 
-        Note: Since hashing is done only portion of secret, it may explode if service or namespace too long, so we should switch to get_boto_new_secret_signature_name as new namespaces are rolled out
+        Note: Since hashing is done only portion of secret, it may explode if service or instance names are too long, so we should switch to get_boto_new_secret_signature_name as new namespaces are rolled out
```
nemacysts, line 1614: imo, we probably want to move away from including the namespace in the secret name as it's redundant (and there's also currently no toggle for changing the current behavior by-namespace)

additionally, we could/should (in a future PR) update the return in this function to limit the signature name's size with `limit_size_with_hash` to avoid the scenario mentioned in the docstring :)

nemacysts:
```suggestion
        Note: Since hashing is done only portion of secret, it may explode if service or instance names are too long
```

paasta_tools/kubernetes_tools.py
nemacysts:
```suggestion
```

nemacysts:
@nurdann if you pull master the failing test should be fixed :)

nurdann:
Infrastage toggling between `0.173.0` and the PR's changes result in no bounce https://fluffy.yelpcorp.com/i/S63h9N17bSDZG1p9Pwz3HKmLHhKCNNJZ.html

nemacysts (approved):
from re-testing internally: looks good!

PR #3599, opened 2023-05-06
Add support for --use-eks in paasta spark-run

nemacysts:
`paasta spark-run` (when run with `--use-eks`) will require a different set of configuration: we'll be using a kubeconfig with the standard EKS setup rather than giving spark a URL to our masters + certs to auth with.

Technically can we do the env var shenanigans inside service-configuration-lib, but imo the added layer makes changing spark things more painful then necessary - and, most importantly, doesn't really use anything like PaastaSystemConfig or srv-configs in a standard way to support changing values without a PaaSTA or service-configuration-lib release.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -1068,8 +1068,7 @@ def _validate_pool(args, system_paasta_config):
 
 
 def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
-    system_paasta_config = load_system_paasta_config()
-    return system_paasta_config.get_kube_clusters().get(cluster, {}).get("server")
+    return load_system_paasta_config().system_paasta_config.get_kube_clusters().get(cluster, {}).get("server")
 
 
 def paasta_spark_run(args):
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -320,6 +320,13 @@ def add_subparser(subparsers):
         default=False,
     )
 
+    list_parser.add_argument(
+        "--use-eks",
+        help="Use the EKS version of the target cluster rather than the Yelp-managed target cluster",
+        action="store_true",
+        dest="use_eks",
+    )
+
     if clusterman_metrics:
         list_parser.add_argument(
             "--suppress-clusterman-metrics-errors",
@@ -567,6 +574,7 @@ def get_spark_env(
     spark_conf_str: str,
     aws_creds: Tuple[Optional[str], Optional[str], Optional[str]],
     ui_port: str,
+    system_paasta_config: SystemPaastaConfig,
 ) -> Dict[str, str]:
     """Create the env config dict to configure on the docker container"""
 
@@ -607,6 +615,9 @@ def get_spark_env(
         )
         spark_env["SPARK_DAEMON_CLASSPATH"] = "/opt/spark/extra_jars/*"
         spark_env["SPARK_NO_DAEMONIZE"] = "true"
+
+    if args.use_eks:
+        spark_env["KUBECONFIG"] = system_paasta_config.get_spark_kubeconfig()
 
     return spark_env
 
@@ -810,10 +821,21 @@ def configure_and_run_docker_container(
     if args.enable_compact_bin_packing:
         volumes.append(f"{pod_template_path}:{pod_template_path}:rw")
 
+    if args.use_eks:
+        volumes.append(
+            f"{system_paasta_config.get_spark_kubeconfig()}:{system_paasta_config.get_spark_kubeconfig()}:ro"
+        )
+
     environment = instance_config.get_env_dictionary()  # type: ignore
     spark_conf_str = create_spark_config_str(spark_conf, is_mrjob=args.mrjob)
     environment.update(
-        get_spark_env(args, spark_conf_str, aws_creds, spark_conf["spark.ui.port"])
+        get_spark_env(
+            args=args,
+            spark_conf_str=spark_conf_str,
+            aws_creds=aws_creds,
+            ui_port=spark_conf["spark.ui.port"],
+            system_paasta_config=system_paasta_config,
+        )
     )  # type:ignore
 
     webui_url = get_webui_url(spark_conf["spark.ui.port"])
@@ -1032,6 +1054,11 @@ def _validate_pool(args, system_paasta_config):
     return True
 
 
+def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
+    system_paasta_config = load_system_paasta_config()
+    return system_paasta_config.get_kube_clusters().get(cluster, {}).get("server")
+
+
 def paasta_spark_run(args):
     # argparse does not work as expected with both default and
     # type=validate_work_dir.
@@ -1161,6 +1188,10 @@ def paasta_spark_run(args):
         needs_docker_cfg=needs_docker_cfg,
         aws_region=args.aws_region,
         force_spark_resource_configs=args.force_spark_resource_configs,
+        use_eks=args.use_eks,
+        k8s_server_address=_get_k8s_url_for_cluster(args.cluster)
+        if args.use_eks
+        else None,
     )
     # TODO: Remove this after MLCOMPUTE-699 is complete
     if "spark.kubernetes.decommission.script" not in spark_conf:
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -1047,6 +1047,7 @@ def _auto_add_timeout_for_job(cmd, timeout_job_runtime):
 
 
 def _validate_pool(args, system_paasta_config):
+    return True
     if args.pool:
         valid_pools = system_paasta_config.get_cluster_pools().get(args.cluster, [])
         if not valid_pools:
@@ -1068,7 +1069,9 @@ def _validate_pool(args, system_paasta_config):
 
 
 def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
-    return load_system_paasta_config().system_paasta_config.get_kube_clusters().get(cluster, {}).get("server")
+    return (
+        load_system_paasta_config().get_kube_clusters().get(cluster, {}).get("server")
+    )
 
 
 def paasta_spark_run(args):
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -1047,7 +1047,6 @@ def _auto_add_timeout_for_job(cmd, timeout_job_runtime):
 
 
 def _validate_pool(args, system_paasta_config):
-    return True
     if args.pool:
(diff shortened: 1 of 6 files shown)


nemacysts:
this will require https://github.com/Yelp/service_configuration_lib/pull/115

requirements-minimal.txt
```diff
@@ -49,7 +49,7 @@ requests-cache >= 0.4.10
 retry
 ruamel.yaml
 sensu-plugin
-service-configuration-lib >= 2.16.4
+service-configuration-lib >= 2.17.0
```
nemacysts, line 52: this may change if someone else makes a release without https://github.com/Yelp/service_configuration_lib/pull/115

tox.ini
```diff
@@ -10,6 +10,7 @@ docker_compose_version = 1.26.2
 # running outside of Yelp.
 indexserver = https://pypi.yelpcorp.com/simple
 basepython = python3.7
+usedevelop = true
```
nemacysts, line 13: not necessary for this change, but this makes development a lot easier and i'm surprised we didn't already have this set

jfongatyelp (approved):
With most things changing here gated by the `--is-eks` flag, is the idea that we'll likely change the default of `--is-eks` to `true` and make sure existing aliases for the non-EKS clusters fetch the right EKS cluster?

bobokun (approved):

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -321,6 +321,13 @@ def add_subparser(subparsers):
         default=False,
     )
 
+    list_parser.add_argument(
+        "--use-eks",
+        help="Use the EKS version of the target cluster rather than the Yelp-managed target cluster",
```
88manpreet, line 326: [Nit, Optional] When I read this, it sounds to me that Spark Job uses EKS version of target k8s cluster instead of Yelp-managed target cluster. That isn't true, right? It's only the control plane/k8s masters which are the EKS version? 
"Use the EKS version of the target k8s master to connect to, rather than the Yelp-managed target cluster"

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1182,6 +1209,10 @@ def paasta_spark_run(args):
         needs_docker_cfg=needs_docker_cfg,
         aws_region=args.aws_region,
         force_spark_resource_configs=args.force_spark_resource_configs,
+        use_eks=args.use_eks,
+        k8s_server_address=_get_k8s_url_for_cluster(args.cluster)
+        if args.use_eks
```
88manpreet, line 1214: Nit: I would rather prefer this if condition earlier in the code as opposed to inside the function parameters.

Something like:
def _get_k8s_url_for_cluster(cluster: str, use_eks: bool) -> Optional[str]:
    if use_eks: return xyz
    else return None

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1045,6 +1067,11 @@ def _validate_pool(args, system_paasta_config):
     return True
 
 
+def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
+    system_paasta_config = load_system_paasta_config()
+    return system_paasta_config.get_kube_clusters().get(cluster, {}).get("server")
```
88manpreet, line 1072:
```suggestion
    return load_system_paasta_config().system_paasta_config.get_kube_clusters().get(cluster, {}).get("server")
```

tests/cli/test_cmds_spark_run.py
```diff
@@ -1374,6 +1405,8 @@ def test_paasta_spark_run_pyspark(
         needs_docker_cfg=False,
         aws_region="test-region",
         force_spark_resource_configs=False,
+        use_eks=False,
```
88manpreet, line 1408: I don't think we have test-case for use_eks=True, right?
If not, can you add one?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -619,6 +627,9 @@ def get_spark_env(
         spark_env["SPARK_DAEMON_CLASSPATH"] = "/opt/spark/extra_jars/*"
         spark_env["SPARK_NO_DAEMONIZE"] = "true"
 
+    if args.use_eks:
+        spark_env["KUBECONFIG"] = system_paasta_config.get_spark_kubeconfig()
```
88manpreet, line 631: hmm Interesting that this env variable wasn't set before.
What would this be this used for? I believe this file `/etc/kubernetes/spark.conf` is yet to be deployed via puppet?

88manpreet: Noticed in description: "we'll be using a kubeconfig with the standard EKS setup rather than giving spark a URL "
Slightly confused here: The major difference I notice here is providing a k8s_server_address for spark master configuration. Is this k8s_servcer_address internally somehow loaded from given KUGECONFIG?

88manpreet:
Description: Technically can we do the env var shenanigans inside service-configuration-lib, but imo the added layer makes changing spark things more painful then necessary ...

Agree with changing spark things being more painful.
Concern: Is this same set-up expected to be used for Jupyter notebooks? 
If so, is the plan to set KUBECONFIG in spark_tools separately? 
If not, server_configuration_lib acts as a common library for both spark_run and spark_tools (jupyter/folium notebooks).

88manpreet:
The code changes look overall ok to me. I also don't see any risk of breaking current infra.
However, as part of our standard protocol of releasing paasta spark changes, can you please add at least 1 test case with and without --use-eks?
`paasta   spark-run  -s spark      --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml   --cmd 'timeout 1h spark-submit  /work/integration_tests/s3.py'` would be good enough, IMO.

tox.ini
bobokun: This was causing issues in `k8s_itests` failing so I have removed this

paasta_tools/cli/cmds/spark_run.py
jfongatyelp: No, we will be creating a new cluster on EKS; nodes cannot be a part of multiple kubernetes cluster and as such, we will be creating a new cluster with the same expected spark pools.  However, we hope to be able to obscure/omit the EKS cluster name from the command when we get to the point where it can be made the default.

paasta_tools/cli/cmds/spark_run.py
jfongatyelp: > What would this be this used for? I believe this file /etc/kubernetes/spark.conf is yet to be deployed via puppet?

We are using this file to authenticate to the kubernetes API server, as we are no longer using the `spark.kubernetes.authenticate.clientKeyFile`/`spark.kubernetes.authenticate.clientCertFile` parameters to authenticate.  It appears we were never setting the KUBECONFIG envvar or using the puppet-deployed KUBECONFIG for the current, non-EKS clusters at all 😮 

This new kubeconfig file is already deployed by puppet, though is currently configured for our test EKS cluster "paasta-test" -- when we are able to spin up our productionized clusters, we will need to change puppet to target the correct cluster

> The major difference I notice here is providing a k8s_server_address for spark master configuration. Is this k8s_servcer_address internally somehow loaded from given KUGECONFIG?

For some reason, unfortuantely spark will not read the k8s_server_address from the specified KUBECONFIG, and so we need spark-run to still specify this in k8s_server_address as well.

nemacysts: we probably could have saved ourselves a bit of effort/code in the spark k8s migration by using the KUBECONFIG env var rather than hardcoding cert locations and whatnot - but otherwise, what jfong said!

88manpreet:
> With most things changing here gated by the `--is-eks` flag, is the idea that eventually we'll likely change the default of `--is-eks` to `true` and make sure existing aliases for the non-EKS clusters fetch the right EKS cluster?

Correct. Though we need to make sure that batches who are still not using spark 3.2 should set this to False explicitly.

paasta_tools/cli/cmds/spark_run.py
88manpreet: > No, we will be creating a new cluster on EKS; nodes cannot be a part of multiple kubernetes cluster and as such, we will be creating a new cluster with the same expected spark pools. However, we hope to be able to obscure/omit the EKS cluster name from the command when we get to the point where it can be made the default.

ack. 

IUC, the workflow is:
* Phase-1 now: k8s control plane is on EKS. 
* Phase-2: Spark pools will be on EKS (when is it expected).
The above explanation caters towards Phase-2, right?

paasta_tools/cli/cmds/spark_run.py
88manpreet: [Not a blocker, for understanding] Is the plan to also deploy this file also on batch and dev boxes? I am assuming this is used by spark driver (running on batch box) to authenticate.

88manpreet (approved):
lgtm

nemacysts:
> The code changes look overall ok to me. I also don't see any risk of breaking current infra. However, as part of our standard protocol of releasing paasta spark changes, can you please add at least 1 test case with and without --use-eks? `paasta spark-run -s spark --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --cmd 'timeout 1h spark-submit /work/integration_tests/s3.py'` would be good enough, IMO.

@88manpreet: 
# --use-eks
`paasta spark-run -s spark --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --cmd 'spark-submit --verbose /work/integration_tests/access_apollo_endpoint_executor.py' --spark-args "spark.executor.memory=2g" --use-eks --pool default --build --force-spark-resource-configs --cluster paasta-test` -> https://fluffy.yelpcorp.com/i/SbvctsNgV56wrDdWr6Z9QNtT4KBPCVPS.html
(the test cluster has pretty small boxes in the default pool and the automatic resource stuff was making these not fit + I'm building to use the latest spark-on-paasta-version)

# status-quo
`paasta spark-run -s spark --aws-credentials-yaml /etc/boto_cfg/mrjob.yaml --cmd 'spark-submit --verbose /work/integration_tests/access_apollo_endpoint_executor.py' --build` -> https://fluffy.yelpcorp.com/i/PP9K9LmQJPmJ5cv4bQDVV5hdrK60zRrp.html

nemacysts:
> With most things changing here gated by the `--is-eks` flag, is the idea that eventually we'll likely change the default of `--is-eks` to `true` and make sure existing aliases for the non-EKS clusters fetch the right EKS cluster?

@jfongatyelp yup - and then delete the --use-eks flag :)

paasta_tools/cli/cmds/spark_run.py
nemacysts: correct - we'll do the pool/cluster creation once we're ready to begin moving things (we can probably test this on our test cluster/pool with some real jobs once this is merged tho)

for now, the only folks use --use-eks will likely be us :)

bobokun (approved):

paasta_tools/cli/cmds/spark_run.py
nemacysts: yup - it should already be deployed on all batch and devboxes :)

PR #3598, opened 2023-05-05
Add support for vault token file for paasta {add,update} secret

danielpops:
## Problem
I found this desirable when adding lots of secrets in an environment so that don't have to re-auth for every secret. I basically just copied the pattern used for other APIs. Not sure if this was purposely omitted for add/update, so let me know if I'm missing any context on why this isn't desirable.

## Testing done
Made the unit tests pass, albeit kinda lazily by using mock.ANY. Happy to improve the tests if there are strong opinions.

PR #3588, opened 2023-04-13
Allow vault options override for paasta secret add and update

siadat:
## Problem or Feature

The add and update subcommands of `paasta secret` prompt for password each time. 

## Solution

Similar to the decrypt subcommand, allow overriding vault configs with the default values.

## Context

https://jira.yelpcorp.com/browse/DREIMP-9626

PR #3603, opened 2023-05-10
Change default directory to `paasta secret {add,update}`

nurdann:
There have been 3 PRs so far that wanted to add passing Vault token to the CLI `paasta secret`: https://github.com/Yelp/paasta/pull/3598 https://github.com/Yelp/paasta/pull/3588 but #3596  was merged

## Preserve existing behaviour
Since `vault_auth_method="token"` falls back to LDAP authentication, we don't have to change the default argument. However, because current directory is hard-coded the script must be run in the soa directory.

Was not sure if `DEFAULT_SOA_DIR` was safe to change
```diff
@@ -231,7 +231,7 @@ def _add_common_args(parser: argparse.ArgumentParser, allow_shared: bool = True)
         "--yelpsoa-config-root",
         dest="yelpsoa_config_root",
         help="A directory from which yelpsoa-configs should be read from",
-        default=DEFAULT_SOA_DIR,
+        default="./",
     )
```

The change should be safe because previously the argument was not passed at all. Previously, I passed the argument from argparse because it was more convenient to run from another directory.

**Update**: @nemacysts confirmed that the script is used only under dev environment and invoked by a user, so it's safe to change the defaults

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -20,8 +20,6 @@ from typing import Any
 from typing import Dict
 from typing import Optional
 
-from service_configuration_lib import DEFAULT_SOA_DIR
-
 from paasta_tools.cli.utils import get_instance_config
 from paasta_tools.cli.utils import get_namespaces_for_secret
 from paasta_tools.cli.utils import lazy_choices_completer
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/secret.py
```diff
@@ -444,7 +444,7 @@ def paasta_secret(args):
         secret_provider = _get_secret_provider_for_service(
             service,
             cluster_names=args.clusters,
-            soa_dir=args.yelpsoa_config_root,
+            soa_dir="./",
```
nemacysts, line 447: thoughts on having this be something like:
```suggestion
            # this will only be invoked on a devbox
            # and only in a context where we certainly
            # want to use the working directory rather
            # than whatever the actual soa_dir path is
            # configured as
            soa_dir=os.getcwd(),
```

nemacysts (approved):

PR #3604, opened 2023-05-10
Hard-code current directory only for `paasta secret {add,update}`

nurdann:
Changing default directory in argparse resulted in YM playground failing, so hard-coding location only for `add` and `update` commands which retain original behaviour.

Introduced by https://github.com/Yelp/paasta/pull/3603

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -20,6 +20,8 @@ from typing import Any
 from typing import Dict
 from typing import Optional
 
+from service_configuration_lib import DEFAULT_SOA_DIR
+
 from paasta_tools.cli.utils import get_instance_config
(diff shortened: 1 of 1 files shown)


nemacysts (approved):

paasta_tools/cli/cmds/secret.py
```diff
@@ -444,13 +441,21 @@ def paasta_secret(args):
         plaintext = get_plaintext_input(args)
         if not plaintext:
             print("Warning: Given plaintext is an empty string.")
+        print(args)
```
calvinli, line 444: left this in?

Qmando, line 444: Let's remove this before shipping

Qmando (approved):

PR #3605, opened 2023-05-11
Separate arguments for `paasta secret` (decrypt, run) vs (add, update)

nurdann:
Based of how default arguments are assumed across different environment, we can separate the arguments so that all commands retain original behaviour while allowing to override with command flags because `paasta secret {decrypt,run}` are run on a service instance, while `paasta secret {add,update}` are ran on a devbox.

## Changes
- Separate args with different default arguments, but should still be noop

## Notes
- For passing token, initially I thought I could use `VAULT_TOKEN_OVERRIDE` in conjunction with `--vault-auth-method token`, but that's not possible if `vault_auth_method="ldap"` hard-coded

### Tests
Tested locally `add`, `update` and `run`,  https://fluffy.yelpcorp.com/i/Pv94gRPCbtCjHcx88Szfqsh9Qfp4Vkg5.html

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -470,7 +470,7 @@ def paasta_secret(args):
         return
 
     if args.action in ["add", "update"]:
-        # this will only be invoked on a devbox only
+        # this should really only ever be invoked on a devbox
         plaintext = get_plaintext_input(args)
         if not plaintext:
             print("Warning: Given plaintext is an empty string.")
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -77,6 +77,7 @@ def add_decrypt_subparser(subparsers):
         "decrypt", help="decrypts a single paasta secret"
     )
     _add_common_args(secret_parser_decrypt)
+    _add_vault_auth_args_for_decrypt_and_run(secret_parser_decrypt)
 
     secret_parser_decrypt.add_argument(
         "-n",
@@ -119,6 +120,7 @@ def add_run_subparser(subparsers):
         conflict_handler="resolve",
     )
     _add_common_args(secret_parser_run, allow_shared=False)
+    _add_vault_auth_args_for_decrypt_and_run(secret_parser_run)
 
     secret_parser_run.add_argument(
         "-i",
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/secret.py
```diff
@@ -224,17 +236,37 @@ def _add_vault_auth_args(parser: argparse.ArgumentParser):
...
-def _add_common_args(parser: argparse.ArgumentParser, allow_shared: bool = True):
-    # available from any subcommand
+def _add_vault_auth_args_for_add_and_update(parser: argparse.ArgumentParser):
     parser.add_argument(
         "-y",
         "--yelpsoa-config-root",
         dest="yelpsoa_config_root",
         help="A directory from which yelpsoa-configs should be read from",
-        default=DEFAULT_SOA_DIR,
+        default=os.getcwd(),
```
nurdann, line 245: Any opinion on using this or "./"?

nemacysts: imo, getcwd() is a little nicer aesthetically (and makes it more obvious that this is what we want rather than someone seeing `./` and wondering if maybe there was a typo before reading the rest of the code)

paasta_tools/cli/cmds/secret.py
```diff
@@ -438,23 +470,17 @@ def paasta_secret(args):
         return
 
     if args.action in ["add", "update"]:
+        # this will only be invoked on a devbox only
```
nemacysts, line 473: nit:
```suggestion
        # this should really only ever be invoked on a devbox
```

(technically, my real nit is just that we say "only" twice here - so feel free to just remove the last "only")

paasta_tools/cli/cmds/secret.py
```diff
@@ -224,17 +236,37 @@ def _add_vault_auth_args(parser: argparse.ArgumentParser):
...
+        choices=["token", "ldap"],
+    )
+
+    parser.add_argument(
+        "--vault-token-file",
+        help="Override vault token file, defaults to %(default)s",
+        type=str,
+        dest="vault_token_file",
+        required=False,
+        default="~/.vault-token",
```
nemacysts, line 264: heh, the variety of different token locations sure is fun :p

nemacysts (approved):
i have a minor comment nit, but otherwise lgtm :)

(although, y'all (i.e., #security) are a lot more familiar with cmds/secret.py than i am :))

nurdann:
Closing since it does fix anything broken. But may useful, in future if we want to add support for separate default arguments.

PR #3606, opened 2023-05-13
Revert local-run locking from PAASTA-13417

nemacysts:
We've since upgraded the version of Docker we run and, according to https://github.com/moby/moby/pull/37781 - this has now been fixed.

Even without having found that PR, I'm pretty confident that this issue has been fixed as we pull an ungodly amount of simultaneous images in our kube nodes and have never seen issues like this :)


nemacysts:
(gotta fix some tests, but this should work)

nemacysts:
tests should be green now (at least, they are on my devbox)

jfongatyelp (approved):

nurdann (approved):

PR #3608, opened 2023-05-16
Use sha256 digest for paasta spark-run

edingroot:
## Description
Explicitly specify docker image digest instead of using image tag when launching a spark cluster to avoid inconsistent docker image between the driver and executors.

## Test
run locally in a service repo which has a cook-image makefile target.

```bash
# Auth
aws-okta -r core-ml -a dev -k -p dev

# Test 1: `--build`
paasta/.tox/py37-linux/bin/paasta spark-run --build --aws-profile=dev --cmd "spark-submit /work/integration_tests/s3.py"

# Test 2: `--image`
paasta/.tox/py37-linux/bin/paasta spark-run --image docker-dev.yelpcorp.com/paasta-spark-run-chi --aws-profile=dev --cmd "spark-submit /work/integration_tests/s3.py"

# Test 3: `--service`
paasta/.tox/py37-linux/bin/paasta spark-run --service spark --aws-profile=dev --cmd "spark-submit /work/integration_tests/s3.py"
```

### Before

Log output:  
image

In SparkUI:  
image

### After

Log output:  
image
image

In SparkUI:  
image

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -891,7 +891,7 @@ def _should_get_resource_requirements(docker_cmd: str, is_mrjob: bool) -> bool:
 def get_docker_cmd(
     args: argparse.Namespace, instance_config: InstanceConfig, spark_conf_str: str
 ) -> str:
-    original_docker_cmd = args.cmd or instance_config.get_cmd()
+    original_docker_cmd = str(args.cmd or instance_config.get_cmd())
 
     if args.mrjob:
         return original_docker_cmd + " " + spark_conf_str
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -1101,10 +1101,7 @@ def paasta_spark_run(args):
     docker_client = get_docker_client()
     image_details = docker_client.inspect_image(docker_image)
     if len(image_details["RepoDigests"]) < 1:
-        print(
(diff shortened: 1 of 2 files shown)


88manpreet:
Can we do few more test cases? Maybe add details in the ticket.
a) Result before this change. And after this change.
b) A test run for an actual spark job?

edingroot:
> Can we do few more test cases? Maybe add details in the ticket. a) Result before this change. And after this change. b) A test run for an actual spark job?

Updated in the PR description. Test by running spark-submit integration test script in the `spark` service.

88manpreet:
> lgtm. Might be better to push on Tuesday. Ping in the thread where the issue was seen. And let them know if they still see the issue.

The testing the description looks good enough to me to rollout this change.

cc'ing: @nemacysts for visibility in case you would  want to review.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -494,7 +495,9 @@ def get_docker_run_cmd(
     return cmd
 
 
-def get_docker_image(args, instance_config):
+def get_docker_image(
```
austinzh, line 498: It would be better if we get repo digest here.
Because we don't know what kind of url user might provide in the `args.image` or from from service config.
The principal here should be 
`The image we get | pull | build and run in driver MUST be exactly the same with the image we run in executor.`
For this to be true, we must use the repo digest for driver and for executor.

austinzh (approved):
LGTM

PR #3609, opened 2023-05-16
Postpone getting secret data until it is needed

nurdann:
Since secret data is not needed until actually it needs to be created, we can delay fetching them. Otherwise, Vault gets too many requests because cronjob runs every minute.

## Tests
Tested locally https://fluffy.yelpcorp.com/i/s2HZJKSjqgjXWv89t5PGhD5qXrcMkxWc.html

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -296,12 +296,14 @@ def sync_secrets(
                         secret_name=get_paasta_secret_name(
                             namespace, service, sanitise_kubernetes_name(secret)
                         ),
-                        secret_data={
-                            secret: base64.b64encode(
-                                # If signatures does not match, it'll sys.exit(1)
-                                secret_provider.decrypt_secret_raw(secret)
-                            ).decode("utf-8")
-                        },
+                        get_secret_data=(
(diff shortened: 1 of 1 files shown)


paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -369,7 +371,7 @@ def sync_crypto_secrets(
             # the secret name here must match the secret name given in the secret volume config,
             # i.e. `kubernetes.client.V1SecretVolumeSource`'s `secret_name` must match below
             secret_name=instance_config.get_crypto_secret_name(),
-            secret_data=secret_data,
+            get_secret_data=(lambda: secret_data),
```
nemacysts, line 374: we probably want a comment here to make it clear why we're doing this :)

nemacysts (approved):

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nurdann: Added param docstring, now I am having second thoughts that I need to postpone getting vault keys as well. Since `crypto_keys` is not widely used currently and that requires further changes, we can ship this to reduce Vault service load

nemacysts (approved):

PR #3611, opened 2023-05-22
updated apiextensions.k8s.io CRD references to V1

ajayOO8:
##Problem
There are couple of places where we refer to `V1beta1CustomResourceDefinition` which has been removed in k8s v1.22

I have update its references to `V1CustomResourceDefinition`

ref: https://kubernetes.io/docs/reference/using-api/deprecation-guide/#customresourcedefinition-v122

Tests:
https://fluffy.yelpcorp.com/i/x26KCgz9pTtZVM7678wH1bwBmnncV4M1.html

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -46,7 +46,6 @@ from kubernetes import config as kube_config
 from kubernetes.client import models
 from kubernetes.client import V1Affinity
 from kubernetes.client import V1AWSElasticBlockStoreVolumeSource
-from kubernetes.client import V1beta1CustomResourceDefinition
 from kubernetes.client import V1beta1PodDisruptionBudget
 from kubernetes.client import V1beta1PodDisruptionBudgetSpec
 from kubernetes.client import V1Capabilities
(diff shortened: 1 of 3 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -522,7 +522,7 @@ def __init__(
         self.deployments = kube_client.AppsV1Api(self.api_client)
         self.core = kube_client.CoreV1Api(self.api_client)
         self.policy = kube_client.PolicyV1beta1Api(self.api_client)
-        self.apiextensions = kube_client.ApiextensionsV1beta1Api(self.api_client)
+        self.apiextensions = kube_client.ApiextensionsV1Api(self.api_client)
```
jfongatyelp, line 525: Since we also use `update_crds` for setup_kubernetes_crd, and these come from yelpsoa-configs, there will be a short period where things break until the CRD definitions in yelpsoa are updated to also use V1 since they currently match what kube_client.apiextensions is using.

I think it would be a bit nicer if we could make this more flexible so paasta reads in what version CRD owners are using in their files.

But if we just want to support one version as you're doing here, we likely will need to pin paasta & roll out slowly, plus work with CRD owners to let them know as we update their configs in yelpsoa.

gmdfalk (approved):

ajayOO8:
This has already been merge through another PR

PR #3615, opened 2023-05-30
TRON-1636: Setup tron secret_volumes in setup_tron_namespace

Molaire:
Ticket: [TRON-1636](https://jira.yelpcorp.com/browse/TRON-1636)

This  pipes secret_volumes in tron. It was already set up in kubernetes, so most of the work here was adjusting the tests.

- Had to use a file-based approach to figure out if the secret is shared or not, as it's not explicit for secret_volumes.
- Added a TronSecretVolume type to allow passing secret_volume_name to Tron/task_processing, so they don't have to figure that out themselves.
- Tested with patched wheel of https://github.com/Yelp/task_processing/pull/188. So once this task_proc PR is merged, we should update requirements.txt

This was tested with 2 real tron jobs and patched paasta and tron on our `tron-infrastage` box.
Success: https://fluffy.yelpcorp.com/i/bsJZQ3TlDR9N6fx3dNjT6WQfT0NVGh03.html
Jobs config: https://sourcegraph.yelpcorp.com/sysgit/yelpsoa-configs/-/blob/compute-infra-test-service/tron-infrastage.yaml?L94-119

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -52,6 +52,7 @@ from paasta_tools.utils import NoConfigurationForServiceError
 from paasta_tools.utils import NoDeploymentsAvailable
 from paasta_tools.utils import time_cache
 from paasta_tools.utils import filter_templates_from_config
+from paasta_tools.utils import TronSecretVolume
 from paasta_tools.kubernetes_tools import (
     allowlist_denylist_to_requirements,
     create_or_find_service_account_name,
@@ -68,6 +69,7 @@ from paasta_tools.spark_tools import (
 )
 from paasta_tools.secret_tools import is_secret_ref
 from paasta_tools.secret_tools import is_shared_secret
+from paasta_tools.secret_tools import is_shared_secret_from_secret_name
 from paasta_tools.secret_tools import get_secret_name_from_ref
 from paasta_tools.kubernetes_tools import get_paasta_secret_name
 from paasta_tools.secret_tools import SHARED_SECRET_SERVICE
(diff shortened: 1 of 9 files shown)


tests/test_tron_tools.py
```diff
@@ -1430,7 +1535,7 @@ def test_create_complete_config_e2e(self, tmpdir):
         # that are not static, this will cause continuous reconfiguration, which
         # will add significant load to the Tron API, which happened in DAR-1461.
         # but if this is intended, just change the hash.
-        assert hasher.hexdigest() == "f740410f7ae2794f9924121c1115e15d"
+        assert hasher.hexdigest() == "35972651618a848ac6bf7947245dbaea"
```
danielpops, line 1538: Note that this may cause a big bounce with all tron jobs when deployed. Make sure you sync with #compute-infra before shipping these changes in case any special considerations are needed to accommodate that.

nemacysts: this *should* be fine - the main purpose of this test is too help us catch things like a timestamp (or something equally ever-changing) being added to the podspec as repeatedly reconfiguring every job is quite painful :)

paasta_tools/tron_tools.py
```diff
@@ -869,6 +869,7 @@ def format_tron_action_dict(action_config: TronActionConfig, use_k8s: bool = Fal
         "node": action_config.get_node(),
         "retries": action_config.get_retries(),
         "retries_delay": action_config.get_retries_delay(),
+        "secret_volumes": action_config.get_secret_volumes(),
```
nemacysts, line 872: lol - nice! always good when things just work out and we don't need to do anything special :)

nemacysts (approved):

tests/test_tron_tools.py
danielpops: luisp++ ah right there in the comment above :) thanks for the context!

nemacysts (approved):

paasta_tools/tron_tools.py
```diff
@@ -415,15 +415,17 @@ def get_job_name(self):
     def get_action_name(self):
         return self.action
 
-    def get_secret_volumes(self) -> List[TronSecretVolume]:
+    def get_secret_volumes(self) -> List[TronSecretVolume]:  # type: ignore
```
Molaire, line 418: TypedDict does not support isinstance and issubclass for a subclass of a defined typeddict, mypy is sad

nemacysts: rip, maybe we can unify TronSecretVolume and SecretVolume in the future (if we did it now we'd need another tron release, right?)

requirements.txt
```diff
@@ -104,7 +104,7 @@ sticht==1.1.18
 strict-rfc3339==0.7
 swagger-spec-validator==2.1.0
 syslogmp==0.2.2
-task-processing==0.1.3
+task-processing==0.12.1
```
nemacysts, line 107: do we actually use task_processing in paasta right now? 😮

nemacysts (approved):

requirements.txt
danielpops: It's "used" in the legacy `remote run` implementation

danielpops (approved):

PR #3617, opened 2023-06-06
Only sync secrets to namespaces where those secrets are actually used.

EvanKrall:


paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -140,34 +144,66 @@ def main() -> None:
...
+    {
+        "yelp-main": {
+            "paasta": {"secret1", "secret2"},
+            "paastasvc-yelp-main": {"secret1", "secret3"},
+            "paasta-flinks": None,
+        },
+        "_shared": {
+            "paasta": {"sharedsecret1"},
+            "paastasvc-yelp-main": {"sharedsecret1", "sharedsecret2"},
+            "paasta-flinks": None,
```
nemacysts, line 170: i wonder if this is a good opportunity to define `set()` as meaning there should be no secrets to allow us to truly sync things (i.e., lay the groundwork for deleting secrets that are no longer used rather than leaving expired keys in the k8s secret store forever)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -140,34 +144,66 @@ def main() -> None:
...
             if instance_type == "kubernetes":
-                config_loader = PaastaServiceConfigLoader(service, soa_dir)
-                for service_instance_config in config_loader.instance_configs(
-                    cluster=cluster, instance_type_class=KubernetesDeploymentConfig
-                ):
-                    services_to_k8s_namespaces[service].add(
-                        service_instance_config.get_namespace()
-                    )
+                continue  # handled above.
             else:
```
nemacysts, line 207: random idea: thoughts on moving removing the else to reduce a layer of indentation? the `continue` above means that it'd be safe to unconditionally do the logic inside the else now

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -176,17 +212,45 @@ def get_services_to_k8s_namespaces(
...
+                        services_to_k8s_namespaces_to_allowlist["_shared"][
+                            INSTANCE_TYPE_TO_K8S_NAMESPACE[instance_type]
+                        ] = None
+
+    return dict(services_to_k8s_namespaces_to_allowlist)
+
+
+def get_secrets_used_by_instance(
+    service_instance_config: KubernetesDeploymentConfig,
+) -> Tuple[Set[str], Set[str]]:
```
nemacysts, line 232: might be nice to turn this into a typeddict/namedtuple/dataclass so that we can handle other secret types more easily in the future

nemacysts: (although, I guess that maybe we just stuff any other secret types into `secrets_used` and not worry about env var vs non-env var secrets :p)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
EvanKrall: With the way the code is written, `set()` does mean nothing is allowed, which is why I used `None` for allow everything.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
EvanKrall: sure

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nemacysts: yea, or more completely: set() for a given service/namespace means we should delete all secrets that had previously been synced for that service in that namespace. 

a second step would then be to actually take the difference between the synced set and allowlist set and delete that per-service/namespace

(although, it's probably easier to do this all in one go at the end of syncing for shared namespaces like paasta-spark as you can union all the secret names that should exist and the delete the difference between the previously synced set and the union of allowlist sets)

cuza (approved):

PR #3619, opened 2023-06-09
returns image url in case docker pull fails

ruiqianyu:
This change avoids the error when `docker pull` fails before correct url can be passed.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -530,7 +530,7 @@ def get_docker_image(
 
     docker_url = ""
     if args.image:
-        docker_url = args.image
+        return args.image
     else:
         try:
             docker_url = instance_config.get_docker_url()


talalryz (approved):

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -530,7 +530,7 @@ def get_docker_image(
 
     docker_url = ""
     if args.image:
-        docker_url = args.image
+        return args.image
```
nemacysts, line 533: doesn't this mean we'll potentially use an image that doesn't exist?

talalryz: the change to check if the image can be pulled was included recently and has since caused build issues when using a `--image`. for now, we are just trying to roll back the change made, until the ml-compute team can figure out a fix

nemacysts: ah, I see!

PR #3620, opened 2023-06-12
Adding support for disabling ANSI colors

cklein:
Reading environment variable NO_COLOR to check if ANSI colors should be omitted, see https://no-color.org/


nemacysts (approved):
seems fine to me - lemme make sure no one else on CI has objections before merging

jfongatyelp (approved):

PR #3623, opened 2023-06-15
Pull paasta spark docker images without sudo in CI env

edingroot:
Sudo is needed for getting credentials when pulling images from paasta docker registry (docker-paasta.yelpcorp.com)  
However, in our CI env, we can't connect to docker from root, and paasta docker images can be pulled with normal user.

=> Pull paasta docker images without sudo in CI env.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -530,7 +530,7 @@ def get_docker_image(
 
     docker_url = ""
     if args.image:
-        return args.image
+        docker_url = args.image
     else:
         try:
             docker_url = instance_config.get_docker_url()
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -552,7 +552,11 @@ def get_docker_image(
         % docker_url,
         file=sys.stderr,
     )
-    retcode, _ = _run("sudo -H docker pull %s" % docker_url, stream=True, timeout=300)
+    # Need sudo for credentials when pulling images from paasta docker registry (docker-paasta.yelpcorp.com)
+    # However, in CI env, we can't connect to docker via root and we can pull with user `jenkins`
+    is_ci_env = "CI" in os.environ
+    cmd_prefix = "" if is_ci_env else "sudo -H "
+    retcode, _ = _run(f"{cmd_prefix}docker pull {docker_url}", stream=True, timeout=300)
```
nemacysts, line 559: I'm a little surprised that this is broken now - are people just now starting to use paasta spark-run in CI?

Additionally, even if folks are just starting to do this now: shouldn't they be building the image as part of their tests to ensure that they test what was actually pushed?

that said: this isn't a blocker as y'all own this file - I'm just not quite sure why this is required all of a sudden :)

edingroot: > are people just now starting to use paasta spark-run in CI?  

Yes, looks like we do run it in CI 🥲
Sourcegraph search:
* `r:ad_targeting_batch f:test lang:sh paasta spark-run`
* `r:spark-on-paasta paasta spark-run`


> shouldn't they be building the image as part of their tests to ensure that they test what was actually pushed?

Yes, but the tests do not build the image using `paasta spark-run --build`, they build and push image separately in the previous steps before running `paasta spark-run --image`.

PR #3624, opened 2023-06-16
Have kubernetes_remove_evicted_pods iterate one namespace at a time i…

EvanKrall:
…nstead of fetching all pods. This should decrease its peak memory usage.

I might try to make this use a field selector when listing pods, rather than doing the filtering in python.

### DIFF
--- a/paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
+++ b/paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
@@ -135,8 +135,12 @@ def evicted_pods_per_service(
     evicted_pods_aggregated: Dict[str, List[EvictedPod]] = defaultdict(list)
 
     for namespace in get_all_namespaces(client):
-        all_pods = get_all_pods(kube_client=client, namespace=namespace)
-        evicted_pods = get_evicted_pods(all_pods)
+        failed_pods = get_all_pods(
+            kube_client=client,
+            namespace=namespace,
+            field_selector="status.phase=Failed",
+        )
+        evicted_pods = get_evicted_pods(failed_pods)
         log.info(
             f"Pods in evicted state in {namespace}: {[pod.metadata.name for pod in evicted_pods]}"
         )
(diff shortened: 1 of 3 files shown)


EmanElsaban (approved):

EmanElsaban:
is the filter to only get namespaces prefixed with ``paastasvc-``

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
```diff
@@ -131,20 +132,25 @@ def remove_pods(
...
-    log.info(f"Pods in evicted state: {[pod.metadata.name for pod in evicted_pods]}")
     evicted_pods_aggregated: Dict[str, List[EvictedPod]] = defaultdict(list)
-    for pod in evicted_pods:
-        service = get_pod_service(pod)
-        if service:
-            evicted_pods_aggregated[service].append(
-                EvictedPod(
-                    pod.metadata.name, pod.metadata.namespace, pod.status.message
+
+    for namespace in get_all_namespaces(client):
```
nemacysts, line 137: should this be `get_all_managed_namespaces`

nemacysts (approved):
lgtm as a 1:1 port - that said, I don't think we can do a selector for evicted pods since their `phase` is Failed and afaik you can't select based on `reason`

paasta_tools/kubernetes/bin/kubernetes_remove_evicted_pods.py
EvanKrall: possibly. I stuck with get_all_namespaces because I wanted to do a 1:1 port for now (and this only removes pods with the `paasta.yelp.com/service` label anyway, afaict).

cuza (approved):

EvanKrall:
> lgtm as a 1:1 port - that said, I don't think we can do a selector for evicted pods since their `phase` is Failed and afaik you can't select based on `reason`

yeah, could at least filter by Failed -- then we wouldn't have to process all the non-Failed pods.

cuza (approved):

EvanKrall:
With the field selector:

```
2.25user 0.11system 0:04.60elapsed 51%CPU (0avgtext+0avgdata 96916maxresident)k
0inputs+0outputs (0major+36312minor)pagefaults 0swaps
```

Without this PR:
```
44.20user 1.62system 0:46.86elapsed 97%CPU (0avgtext+0avgdata 2504204maxresident)k
0inputs+0outputs (0major+674209minor)pagefaults 0swaps
```

so 26x less memory and 10x faster, lol

mattmb:

PR #3627, opened 2023-06-21
Update uwsgi_exporter dockerfile + build

nemacysts:
This change updates us to the latest uwsgi_exporter binary + builds a sidecar using an Ubuntu Jammy base image rather than an ancient Xenial base image.

Additionally, we now explicitly call the Make targets we're interested in (test and build) rather than the default (which goes into a very full-featured common Makefile that does a whole lot of things we don't necerssarily want to run)

### DIFF
--- a/yelp_package/dockerfiles/uwsgi_exporter-sidecar/Dockerfile
+++ b/yelp_package/dockerfiles/uwsgi_exporter-sidecar/Dockerfile
@@ -1,4 +1,4 @@
-FROM docker-dev.yelpcorp.com/xenial_yelp
+FROM docker-dev.yelpcorp.com/jammy_yelp
 
 COPY ./uwsgi_exporter/uwsgi_exporter /bin/uwsgi_exporter
 ENV STATS_PORT=8889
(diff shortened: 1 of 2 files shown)


nemacysts:
From the release notes, it looks like this is likely safe: https://github.com/timonwong/uwsgi_exporter/releases/tag/v1.1.0
https://github.com/timonwong/uwsgi_exporter/releases/tag/v1.2.0
https://github.com/timonwong/uwsgi_exporter/releases/tag/v1.3.0

but I can test this on one of our test clusters before merging this (unless we want to test post-merge, which should be safe since we pin the sidecar container image in SystemPaastaConfig).

I'm also planning on setting up a CI pipeline for this sidecar image so that we don't have to manually build+push to our internal registry.

nemacysts:
that said, if we don't want to try the upgrade, this also seems to work with 1.0.0

jvperrin (approved):
Awesome, thank you! 😍

nemacysts:
https://github.com/Yelp/paasta/pull/3627/files#diff-4da86e39a59f404e760d7ca22a0b7267e539408291bc69f2608a16c09f5409aaR28 adds a make target that i'll reference in an internal pr

jvperrin (approved):

EvanKrall (approved):

PR #3629, opened 2023-06-22
Introduce gunicorn_exporter-sidecar

ymilki:
I am adding autoscaling support for gunicorn services to paasta - allowing for existing uwsgi paasta services to switch to gunicorn.

This is a multi-part process:

- [x] Add `metrics_provider: gunicorn` to config validation (#3630)
- [ ] Add new statsd-exporter sidecar (named gunicorn_exporter-sidecar) (this PR)
- [ ] Duplicate uwsgi logic for gunicorn 

## In this PR

Add new statsd-exporter sidecar (named gunicorn-exporter-sidecar)

* Replicate the `uwsgi_exporter-sidecar` setup
  * listens on 9117 for the prometheus scraper
* Provides total worker and worker busy stats for prometheus consumption

## Notes

The `statsd-exporter` comes from https://github.com/prometheus/statsd_exporter
A gunicorn paasta service sends metrics via statsd to the exporter sidecar. Prometheus scrapes metrics from the sidecar.

PR #3630, opened 2023-06-22
Add `metrics_provider: gunicorn' to schema

ymilki:
I am adding autoscaling support for gunicorn services to paasta - allowing for existing uwsgi paasta services to switch to gunicorn.

This is a multi-part process:

- [ ] Add `metrics_provider: gunicorn` to config validation (this PR)
- [ ] Add new statsd-exporter sidecar (named gunicorn_exporter-sidecar) (#3629)
- [ ] Duplicate uwsgi logic for gunicorn 

## In this PR

Add `metrics_provider: gunicorn' to schema

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -197,6 +197,7 @@
                                 "uwsgi",
                                 "cpu",
                                 "piscina",
+                                "gunicorn",
                                 "arbitrary_promql"
                             ]
                         },
@@ -285,6 +286,7 @@
                                                 "uwsgi",
                                                 "cpu",
                                                 "piscina",
+                                                "gunicorn",
                                                 "if metrics_provider is arbitrary_promql, the prometheus_adapter_config parameter is required"
                                             ]
                                         }


paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -285,7 +286,8 @@
                                                 "uwsgi",
                                                 "cpu",
                                                 "piscina",
-                                                "if metrics_provider is arbitrary_promql, the prometheus_adapter_config parameter is required"
```
jvperrin, line 288: I think this probably needs to go at the end, from the comment above this? Not entirely sure on that though, just a guess

ymilki: Order shouldn't matter. `enum` is a set of fixed values used in validation.

http://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values
>  The enum keyword is used to restrict a value to a fixed set of values

But leaving it at the end makes more sense visually.

jvperrin: Ok, fair enough, I don't really know how this string helps with the error message anyways

jvperrin (approved):

PR #3637, opened 2023-06-29
Gunicorn autoscaling

ymilki:
I am adding autoscaling support for gunicorn services to paasta - allowing for existing uwsgi paasta services to switch to gunicorn.

This is a multi-part process:

- [x] Add `metrics_provider: gunicorn` to config validation (#3630 )
- [x] Add new statsd-exporter sidecar (named gunicorn_exporter-sidecar) (#3629)
- [ ] Duplicate uwsgi logic for gunicorn (this PR)

## In this PR

Duplicate uwsgi logic for gunicorn

* Update utilities and tools
* Add scaling rule
* Add gunicorn sidecar when requested
* Label pod appropriately

### Notes

I've duplicated most of uwsgi logic for gunicorn.

Changes from uwsgi:
* Dropped support for `offset` since that option is deprecated
* Assume prometheus will always be used.

## Testing

I've validated this branch in our staging environment with a gunicorn service. I was able to see the service scale up and down in response to changes in worker utilization.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -387,7 +387,7 @@ instance MAY have:
   * ``autoscaling``: See the `autoscaling docs <autoscaling.html>`_ for details
 
     * ``metrics_provider``: Which method the autoscaler will use to determine a service's utilization.
-      Should be ``cpu`` or ``uwsgi``.
+      Should be ``cpu``, ``uwsgi``, or ``gunicorn``.
 
     * ``decision_policy``: Which method the autoscaler will use to determine when to autoscale a service.
       Should be ``proportional`` or ``bespoke``.
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -149,6 +149,24 @@ Kubernetes
 
 These options are only applicable to tasks scheduled on Kubernetes.
 
+  * ``topology_spread_constraints``: A set of rules to spread Pods across topologies, for example to try spreading Pods evenly across both nodes and availability zones::
+
+      topology_spread_constraints:
+        - max_skew: 1
+          topology_key: "topology.kubernetes.io/zone"
+          when_unsatisfiable: "ScheduleAnyway"
+        - max_skew: 1
+          topology_key: "kubernetes.io/hostname"
+          when_unsatisfiable: "ScheduleAnyway"
+
+    These can be configured per cluster (or globally) and will be added to every Pod Spec template, using `paasta.yelp.com/service` and `paasta.yelp.com/instance` as selectors.
+
+    To avoid conflicts with the `deploy_whitelist` and `deploy_blacklist`, please only use `when_unsatisfiable: "ScheduleAnyway"` (at least until PAASTA-17951 is resolved).
+
+    For more information, see the official Kubernetes
+    documentation on `topology spread constraints
+    <https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/>`_.
+
   * ``node_selectors``: A map of labels a node is required to have for a task
     to be launched on said node. There are several ways to define a selector.
     The simplest is a key-value pair. For example, this selector restricts a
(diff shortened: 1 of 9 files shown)


paasta_tools/utils.py
```diff
@@ -2662,6 +2664,18 @@ def get_uwsgi_exporter_sidecar_image_url(self) -> str:
...
+        return self.config_dict.get(
+            "gunicorn_exporter_sidecar_image_url",
+            "docker-dev.yelpcorp.com/gunicorn_exporter-k8s-sidecar:v0.24.0-yelp0",
+        )
+
+    def default_should_run_gunicorn_exporter_sidecar(self) -> bool:
+        return self.config_dict.get(
+            "default_should_run_gunicorn_exporter_sidecar", False
+        )
+
```
nemacysts, line 2678:
```suggestion
```

looks like we're not using this function anywhere :)

tests/test_kubernetes_tools.py
```diff
@@ -1681,9 +1719,27 @@ def test_get_pod_template_spec(
...
+                instance="fm",
+                cluster="brentford",
+                config_dict=mock_config_dict,
+                branch_dict=None,
+            )
+            ret = deployment.get_pod_template_spec(
+                git_sha="aaaa123", system_paasta_config=mock_system_paasta_config
+            )
+        else:
+            ret = self.deployment.get_pod_template_spec(
```
nemacysts, line 1740: heh, took me a while to notice the difference between the two variable assignments here :p

paasta_tools/utils.py
```diff
@@ -1913,6 +1913,7 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     dashboard_links: Dict[str, Dict[str, str]]
     default_push_groups: List
     default_should_run_uwsgi_exporter_sidecar: bool
+    default_should_run_gunicorn_exporter_sidecar: bool
```
nemacysts, line 1916:
```suggestion
```

(unless i'm wrong below about this not being used)

docs/source/yelpsoa_configs.rst
```diff
@@ -369,7 +369,7 @@ instance MAY have:
   * ``autoscaling``: See the `autoscaling docs <autoscaling.html>`_ for details
 
     * ``metrics_provider``: Which method the autoscaler will use to determine a service's utilization.
-      Should be ``cpu`` or ``uwsgi``.
+      Should be ``cpu``, ``uwsgi``, or ``gunicorn`.
```
kaisen, line 372: missing a ` at the end

ymilki:
Rebased on master to fix conflicts.

* Removed `default_should_run_gunicorn_exporter_sidecar`
* Fixed rst typo

paasta_tools/kubernetes_tools.py
```diff
@@ -1095,6 +1106,42 @@ def should_run_uwsgi_exporter_sidecar(
...
 
+    def get_gunicorn_exporter_sidecar_container(
+        self,
+        system_paasta_config: SystemPaastaConfig,
+    ) -> Optional[V1Container]:
+
+        if self.should_run_gunicorn_exporter_sidecar():
+            return V1Container(
+                image=system_paasta_config.get_gunicorn_exporter_sidecar_image_url(),
+                resources=self.get_sidecar_resource_requirements("gunicorn_exporter"),
```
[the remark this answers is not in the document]
nemacysts: @ymilki good timing on the PR re-review request - we were just chatting about sidecar resources and we noticed that the uwsgi sidecar is definitely over-provisioned so it'd be good to make fix this from the start for a new sidecar :)

I have a feeling that this sidecar will behave similarly, so we can probably lower the cpu request for the gunicorn exporter to something a lot lower than 1 cpu (e.g., we can be conservative and go down to .5 and then slowly ramp down OR be a little more aggressive and go straight to .1 or .2) - and we'd also want to do something similar for the memory request here (even cutting this to 512mb would be nicer to our new memory-optimized default pool - assuming that the memory usage for this sidecar also matches the uwsgi one)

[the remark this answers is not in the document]
ymilki: @nemacysts   Did you want me to make that change here or only update the gunicorn sidecar by adding `sidecar_resource_requirements` in the configs?

tests/test_kubernetes_tools.py
ymilki: Made it more explicit as `autoscaled_deployment`

nemacysts (approved):

nemacysts (approved):

ymilki:
mypy/flake8 were failing for missing `V1Handler` because this branch was tested on the merge commit with master. In master, `kubernetes` was updated and the API has changed and `V1Handler` became `V1LifecyclerHandler`.

Merged master and fixed the reference. If the kubernetes update gets reverted, we'll have issues again.

PR #3655, opened 2023-07-19
Introduced new soa key for rotating database passwords

RobbySingh:
https://fluffy.yelpcorp.com/i/67hDPrWrG31CtcZRZf0RB17HqJg8XqXZ.html

Pasting in fluffy for security reasons

(updated Aug 9, 2023 @ 1:46pm)

### DIFF
--- a/k8s_itests/deployments/kind/cluster-devbox.yaml
+++ b/k8s_itests/deployments/kind/cluster-devbox.yaml
@@ -7,6 +7,14 @@ nodes:
   - role: worker
     labels:
       yelp.com/pool: default
+    # this mount is needed for services which rely on /nail/etc/{ecosystem,runtime}
+    extraMounts:
+      - hostPath: /nail/etc
+        containerPath: /nail/etc
   - role: worker
     labels:
       yelp.com/pool: default
+    # this mount is needed for services which rely on /nail/etc/{ecosystem,runtime}
+    extraMounts:
+      - hostPath: /nail/etc
+        containerPath: /nail/etc
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -486,6 +486,20 @@
                         }
                     }
                 },
+                "datastore_credentials": {
+                    "type": "object",
+                    "additionalProperties": true,
+                    "properties": {
+                        "mysql": {
+                            "type": "array",
+                            "items": {
+                                "type": "string",
+                                "pattern": "^[a-zA-Z][a-zA-Z0-9_.-]*$"
+                            },
+                            "uniqueItems": true
+                        }
+                    }
+                },
                 "extra_volumes": {
                     "type": "array",
                     "items": {
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -494,7 +494,7 @@
                             "type": "array",
                             "items": {
                                 "type": "string",
-                                "pattern": "^[a-zA-Z][a-zA-Z0-9_.-]*$"
+                                "pattern": "^[a-zA-Z0-9_-]+$"
                             },
                             "uniqueItems": true
                         }
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -488,7 +488,7 @@
                 },
                 "datastore_credentials": {
                     "type": "object",
-                    "additionalProperties": true,
+                    "additionalProperties": false,
                     "properties": {
                         "mysql": {
                             "type": "array",
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -449,8 +449,12 @@ def sync_datastore_credentials(
     vault_cluster_config: Dict[str, str],
     soa_dir: str,
     vault_token_file: str,
-    overwrite_namespace: Optional[str],
+    overwrite_namespace: Optional[str] = None,
 ) -> bool:
+    """
+    Map all the passwords requested for this service-instance to a single Kubernetes Secret store.
+    Volume mounts will then map the associated secrets to their associated mount paths.
+    """
     config_loader = PaastaServiceConfigLoader(service=service, soa_dir=soa_dir)
     system_paasta_config = load_system_paasta_config()
     datastore_credentials_vault_overrides = (
@@ -484,6 +488,7 @@ def sync_datastore_credentials(
                 },
             )
 
+        secret_data = {}
         for datastore, credentials in datastore_credentials.items():
             for credential in credentials:
                 vault_path = f"secrets/datastore/{datastore}/{credential}"
@@ -501,30 +506,19 @@ def sync_datastore_credentials(
                 # kubernetes expects data to be base64 encoded binary in utf-8 when put into secret maps
                 # may look like:
                 # {'master': {'passwd': '****', 'user': 'v-approle-mysql-serv-nVcYexH95A2'}, 'reporting': {'passwd': '****', 'user': 'v-approle-mysql-serv-GgCpRIh9Ut7'}, 'slave': {'passwd': '****', 'user': 'v-approle-mysql-serv-PzjPwqNMbqu'}
-                secret_data = {}
                 secret_data[vault_key_path] = base64.b64encode(
                     json.dumps(secrets).encode("utf-8")
                 ).decode("utf-8")
 
-                if not secret_data:
-                    continue
-
-                # delay between k8s updates
-                time.sleep(0.3)
-
-                create_or_update_k8s_secret(
-                    service=service,
-                    signature_name=instance_config.get_datastore_credentials_signature_name(
-                        datastore=datastore, credential=credential
-                    ),
-                    secret_name=instance_config.get_datastore_credentials_secret_name(
-                        datastore=datastore, credential=credential
-                    ),
-                    get_secret_data=(lambda: secret_data),
-                    secret_signature=_get_dict_signature(secret_data),
-                    kube_client=kube_client,
-                    namespace=namespace,
-                )
+        create_or_update_k8s_secret(
+            service=service,
+            signature_name=instance_config.get_datastore_credentials_signature_name(),
+            secret_name=instance_config.get_datastore_credentials_secret_name(),
+            get_secret_data=(lambda: secret_data),
+            secret_signature=_get_dict_signature(secret_data),
+            kube_client=kube_client,
+            namespace=namespace,
+        )
 
     return True
 
@@ -580,8 +574,6 @@ def sync_crypto_secrets(
         if not secret_data:
             continue
 
-        # In order to prevent slamming the k8s API, add some artificial delay here
-        time.sleep(0.3)
         create_or_update_k8s_secret(
             service=service,
             signature_name=instance_config.get_crypto_secret_signature_name(),
@@ -629,8 +621,6 @@ def sync_boto_secrets(
         if not secret_data:
             continue
 
-        # In order to prevent slamming the k8s API, add some artificial delay here
-        time.sleep(0.3)
         create_or_update_k8s_secret(
             service=service,
             signature_name=instance_config.get_boto_secret_signature_name(),
@@ -661,6 +651,9 @@ def create_or_update_k8s_secret(
     """
     :param get_secret_data: is a function to postpone fetching data in order to reduce service load, e.g. Vault API
     """
+    # In order to prevent slamming the k8s API, add some artificial delay here
+    time.sleep(0.3)
+
     kubernetes_signature = get_secret_signature(
         kube_client=kube_client,
         signature_name=signature_name,
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -27,6 +27,7 @@ from typing import Callable
 from typing import Dict
 from typing import Generator
 from typing import List
+from typing import Mapping
 from typing import Optional
 from typing import Set
 from typing import Tuple
@@ -110,7 +111,7 @@ def parse_args() -> argparse.Namespace:
         ],
         default="all",
         type=str,
-        help="Define which type of secret to add/update. Default is 'all'",
+        help="Define which type of secret to add/update. Default is 'all' (which does not include datastore-credentials)",
     )
     args = parser.parse_args()
     return args
@@ -118,7 +119,7 @@ def parse_args() -> argparse.Namespace:
 
 @contextlib.contextmanager
 def set_temporary_environment_variables(
-    **environ: Dict[str, str]
+    environ: Mapping[str, str]
 ) -> Generator[None, None, None]:
     """
     *Note the return value means "yields None, takes None, and when finished, returns None"*
@@ -471,9 +472,7 @@ def sync_datastore_credentials(
             else instance_config.get_namespace()
         )
         datastore_credentials = instance_config.get_datastore_credentials()
-        with set_temporary_environment_variables(
-            **datastore_credentials_vault_overrides
-        ):
+        with set_temporary_environment_variables(datastore_credentials_vault_overrides):
             # expects VAULT_ADDR_OVERRIDE, VAULT_CA_OVERRIDE, and VAULT_TOKEN_OVERRIDE to be set
             # in order to use a custom vault shard. overriden temporarily in this context
             provider = get_secret_provider(
@@ -489,27 +488,29 @@ def sync_datastore_credentials(
                 },
             )
 
-        secret_data = {}
-        for datastore, credentials in datastore_credentials.items():
-            for credential in credentials:
-                vault_path = f"secrets/datastore/{datastore}/{credential}"
-                secrets = provider.get_data_from_vault_path(vault_path)
-                if not secrets:
-                    # no secrets found at this path. skip syncing
-                    log.debug(
-                        f"Warning: no secrets found at requested path {vault_path}."
-                    )
-                    continue
+            secret_data = {}
+            for datastore, credentials in datastore_credentials.items():
+                # mypy loses type hints on '.items' and throws false positives. unfortunately have to type: ignore
+                # https://github.com/python/mypy/issues/7178
+                for credential in credentials:  # type: ignore
+                    vault_path = f"secrets/datastore/{datastore}/{credential}"
+                    secrets = provider.get_data_from_vault_path(vault_path)
+                    if not secrets:
+                        # no secrets found at this path. skip syncing
+                        log.debug(
+                            f"Warning: no secrets found at requested path {vault_path}."
+                        )
+                        continue
 
-                # decrypt and save in secret_data
-                vault_key_path = get_vault_key_secret_name(vault_path)
+                    # decrypt and save in secret_data
+                    vault_key_path = get_vault_key_secret_name(vault_path)
 
-                # kubernetes expects data to be base64 encoded binary in utf-8 when put into secret maps
-                # may look like:
-                # {'master': {'passwd': '****', 'user': 'v-approle-mysql-serv-nVcYexH95A2'}, 'reporting': {'passwd': '****', 'user': 'v-approle-mysql-serv-GgCpRIh9Ut7'}, 'slave': {'passwd': '****', 'user': 'v-approle-mysql-serv-PzjPwqNMbqu'}
-                secret_data[vault_key_path] = base64.b64encode(
-                    json.dumps(secrets).encode("utf-8")
-                ).decode("utf-8")
+                    # kubernetes expects data to be base64 encoded binary in utf-8 when put into secret maps
+                    # may look like:
+                    # {'master': {'passwd': '****', 'user': 'v-approle-mysql-serv-nVcYexH95A2'}, 'reporting': {'passwd': '****', 'user': 'v-approle-mysql-serv-GgCpRIh9Ut7'}, 'slave': {'passwd': '****', 'user': 'v-approle-mysql-serv-PzjPwqNMbqu'}
+                    secret_data[vault_key_path] = base64.b64encode(
+                        json.dumps(secrets).encode("utf-8")
+                    ).decode("utf-8")
 
         create_or_update_k8s_secret(
             service=service,
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -353,6 +353,7 @@ def sync_all_secrets(
                 vault_cluster_config=vault_cluster_config,
                 soa_dir=soa_dir,
                 vault_token_file=vault_token_file,
+                overwrite_namespace=overwrite_namespace,
             )
         )
 
@@ -448,6 +449,7 @@ def sync_datastore_credentials(
     vault_cluster_config: Dict[str, str],
     soa_dir: str,
     vault_token_file: str,
+    overwrite_namespace: Optional[str],
 ) -> bool:
     config_loader = PaastaServiceConfigLoader(service=service, soa_dir=soa_dir)
     system_paasta_config = load_system_paasta_config()
@@ -458,6 +460,11 @@ def sync_datastore_credentials(
     for instance_config in config_loader.instance_configs(
         cluster=cluster, instance_type_class=KubernetesDeploymentConfig
     ):
+        namespace = (
+            overwrite_namespace
+            if overwrite_namespace is not None
+            else instance_config.get_namespace()
+        )
         datastore_credentials = instance_config.get_datastore_credentials()
         with set_temporary_environment_variables(
             **datastore_credentials_vault_overrides
@@ -516,7 +523,7 @@ def sync_datastore_credentials(
                     get_secret_data=(lambda: secret_data),
                     secret_signature=_get_dict_signature(secret_data),
                     kube_client=kube_client,
-                    namespace=instance_config.get_namespace(),
+                    namespace=namespace,
                 )
 
     return True
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -25,6 +25,7 @@ from collections import defaultdict
 from functools import partial
 from typing import Callable
 from typing import Dict
+from typing import Generator
 from typing import List
 from typing import Optional
 from typing import Set
@@ -115,9 +116,14 @@ def parse_args() -> argparse.Namespace:
 
 
 @contextlib.contextmanager
-def set_env(**environ) -> None:
+def set_temporary_environment_variables(
+    **environ: Dict[str, str]
+) -> Generator[None, None, None]:
     """
-    :type environ: dict[str, unicode]
+    *Note the return value means "yields None, takes None, and when finished, returns None"*
+
+    Modifies the os.environ variable then yields this temporary state. Resets it when finished.
+
     :param environ: Environment variables to set
     """
     old_environ = dict(os.environ)  # ensure we're storing a copy
@@ -453,7 +459,9 @@ def sync_datastore_credentials(
         cluster=cluster, instance_type_class=KubernetesDeploymentConfig
     ):
         datastore_credentials = instance_config.get_datastore_credentials()
-        with set_env(**datastore_credentials_vault_overrides):
+        with set_temporary_environment_variables(
+            **datastore_credentials_vault_overrides
+        ):
             # expects VAULT_ADDR_OVERRIDE, VAULT_CA_OVERRIDE, and VAULT_TOKEN_OVERRIDE to be set
             # in order to use a custom vault shard. overriden temporarily in this context
             provider = get_secret_provider(
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -98,7 +98,13 @@ def parse_args() -> argparse.Namespace:
     )
     parser.add_argument(
         "--secret-type",
-        choices=["all", "paasta-secret", "boto-key", "crypto-key"],
+        choices=[
+            "all",
+            "paasta-secret",
+            "boto-key",
+            "crypto-key",
+            "database-credentials",
+        ],
         default="all",
         type=str,
         help="Define which type of secret to add/update. Default is 'all'",
@@ -259,7 +265,9 @@ def sync_all_secrets(
     vault_cluster_config: Dict[str, str],
     soa_dir: str,
     vault_token_file: str,
-    secret_type: Literal["all", "paasta-secret", "crypto-key", "boto-key"] = "all",
+    secret_type: Literal[
+        "all", "paasta-secret", "crypto-key", "boto-key", "database-credentials"
+    ] = "all",
     overwrite_namespace: Optional[str] = None,
 ) -> bool:
     results = []
@@ -313,12 +321,26 @@ def sync_all_secrets(
             )
         )
 
+        sync_service_secrets["database-credentials"].append(
+            partial(
+                sync_database_credentials,
+                kube_client=kube_client,
+                cluster=cluster,
+                service=service,
+                secret_provider_name=secret_provider_name,
+                vault_cluster_config=vault_cluster_config,
+                soa_dir=soa_dir,
+                vault_token_file=vault_token_file,
+            )
+        )
+
         if secret_type == "all":
             results.append(
                 all(sync() for sync in sync_service_secrets["paasta-secret"])
             )
             results.append(all(sync() for sync in sync_service_secrets["boto-key"]))
             results.append(all(sync() for sync in sync_service_secrets["crypto-key"]))
+            # note that since database-credentials are in a different vault, they're not synced as part of 'all'
         else:
             results.append(all(sync() for sync in sync_service_secrets[secret_type]))
 
@@ -392,6 +414,79 @@ def sync_secrets(
                         kube_client=kube_client,
                         namespace=namespace,
                     )
+
+    return True
+
+
+def sync_database_credentials(
+    kube_client: KubeClient,
+    cluster: str,
+    service: str,
+    secret_provider_name: str,
+    vault_cluster_config: Dict[str, str],
+    soa_dir: str,
+    vault_token_file: str,
+) -> bool:
+    config_loader = PaastaServiceConfigLoader(service=service, soa_dir=soa_dir)
+    for instance_config in config_loader.instance_configs(
+        cluster=cluster, instance_type_class=KubernetesDeploymentConfig
+    ):
+        # expects VAULT_ADDR_OVERRIDE, VAULT_CA_OVERRIDE, and VAULT_TOKEN_OVERRIDE to be set
+        # in order to use a custom vault shard
+        database_credentials = instance_config.get_database_credentials()
+
+        provider = get_secret_provider(
+            secret_provider_name=secret_provider_name,
+            soa_dir=soa_dir,
+            service_name=service,
+            cluster_names=[cluster],
+            # overridden by env variables but still needed here for spec validation
+            secret_provider_kwargs={
+                "vault_cluster_config": vault_cluster_config,
+                "vault_auth_method": "token",
+                "vault_token_file": vault_token_file,
+            },
+        )
+
+        for dstore in database_credentials.keys():
+            for credential in database_credentials[dstore]:
+                vault_path = f"secrets/database/{dstore}/{credential}"
+                secrets = provider.get_secrets_from_vault_path(vault_path)
+                if not secrets:
+                    # no secrets found at this path. skip syncing
+                    log.debug(
+                        f"Warning: no secrets found at requested path {vault_path}."
+                    )
+                    continue
+
+                # decrypt and save in secret_data
+                vault_key_path = get_vault_key_secret_name(vault_path)
+
+                # kubernetes expects data to be base64 encoded binary in utf-8 when put into secret maps
+                secret_data = {}
+                secret_data[vault_key_path] = base64.b64encode(
+                    json.dumps(secrets).encode("utf-8")
+                ).decode("utf-8")
+
+                if not secret_data:
+                    continue
+
+                # delay between k8s updates
+                time.sleep(0.3)
+
+                create_or_update_k8s_secret(
+                    service=service,
+                    signature_name=instance_config.get_database_credentials_signature_name(
+                        dstore=dstore, credential=credential
+                    ),
+                    secret_name=instance_config.get_database_credentials_secret_name(
+                        dstore=dstore, credential=credential
+                    ),
+                    get_secret_data=(lambda: secret_data),
+                    secret_signature=_get_dict_signature(secret_data),
+                    kube_client=kube_client,
+                    namespace=instance_config.get_namespace(),
+                )
 
     return True
 
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 import argparse
 import base64
+import contextlib
 import hashlib
 import json
 import logging
@@ -111,6 +112,21 @@ def parse_args() -> argparse.Namespace:
     )
     args = parser.parse_args()
     return args
+
+
+@contextlib.contextmanager
+def set_env(**environ) -> None:
+    """
+    :type environ: dict[str, unicode]
+    :param environ: Environment variables to set
+    """
+    old_environ = dict(os.environ)
+    os.environ.update(environ)
+    try:
+        yield
+    finally:
+        os.environ.clear()
+        os.environ.update(old_environ)
 
 
 def main() -> None:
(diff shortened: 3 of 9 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1516,8 +1516,63 @@ def get_pod_volumes(
...
+        )
+
+    def get_database_credentials_secrets_volumes(self) -> Optional[V1Volume]:
+        database_credentials = self.get_database_credentials()
+        if not database_credentials:
+            return None
+
+        result = []
+        for dstore in database_credentials.keys():
+            for credential in database_credentials[dstore]:
```
nemacysts, line 1554: nit: might be nice to use a bit of python magic here :)
```suggestion
        for dstore, credentials in database_credentials.items():
            for credential in credentials:
```

paasta_tools/kubernetes_tools.py
```diff
@@ -1516,8 +1516,63 @@ def get_pod_volumes(
...
+                v1_secret_source_name = self.get_database_credentials_secret_name(
+                    dstore=dstore, credential=credential
+                )
+
+                # map the entirety of this v1 secret to the volume name
+                result.append(
+                    V1Volume(
+                        name=volume_name,
+                        secret=V1SecretVolumeSource(
+                            secret_name=v1_secret_source_name,
```
nemacysts, line 1567: i'd suggest not including the version in the variable name as that's one more place to update should k8s have a new api version for secrets

```suggestion
                secret_name = self.get_database_credentials_secret_name(
                    dstore=dstore, credential=credential
                )

                # map the entirety of this secret to the volume name
                result.append(
                    V1Volume(
                        name=volume_name,
                        secret=V1SecretVolumeSource(
                            secret_name=secret_name,
```

paasta_tools/secret_providers/vault.py
```diff
@@ -155,6 +155,14 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
         else:
             return None
 
+    def get_secrets_from_vault_path(self, path=""):
+        data = self.clients[self.ecosystems[0]].read(path)
+        # returns None on empty path. return {} instead
+        # since we want to return the data['data'] key anyway
+        if data:
+            return data.get("data", {})
```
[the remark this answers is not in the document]
nemacysts: (from the crypto keys code, I assume that all vault responses will be inside a `data` dict and the values there will always be a dict as well, but just double-checking)

paasta_tools/secret_providers/vault.py
```diff
@@ -155,6 +155,14 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
         else:
             return None
 
+    def get_secrets_from_vault_path(self, path=""):
+        data = self.clients[self.ecosystems[0]].read(path)
+        # returns None on empty path. return {} instead
```
nemacysts, line 160: will we ever be passed in an empty path? and why is returning an empty dict preferable over None in this case?

paasta_tools/secret_providers/vault.py
```diff
@@ -155,6 +155,14 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
         else:
             return None
 
+    def get_secrets_from_vault_path(self, path=""):
```
nemacysts, line 158: i wonder if this should be `get_data_from_vault_path` just to ensure that folks don't try to use this directly (they still need to wrap this in a k8s secret) and future users of this might just have a single piece of data at a given path  (although, there's some existing vault method that mention `secret`, so maybe `get_secret_from_vault_path`?)

nemacysts: (and sidenote: this is missing type annotations)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -396,6 +418,79 @@ def sync_secrets(
...
+            # overridden by env variables but still needed here for spec validation
+            secret_provider_kwargs={
+                "vault_cluster_config": vault_cluster_config,
+                "vault_auth_method": "token",
+                "vault_token_file": vault_token_file,
+            },
+        )
+
+        for dstore in database_credentials.keys():
+            for credential in database_credentials[dstore]:
```
nemacysts, line 452:
```suggestion
        for dstore, credentials in database_credentials.items():
            for credential in credentials:
```

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -396,6 +418,79 @@ def sync_secrets(
...
+                    )
+                    continue
+
+                # decrypt and save in secret_data
+                vault_key_path = get_vault_key_secret_name(vault_path)
+
+                # kubernetes expects data to be base64 encoded binary in utf-8 when put into secret maps
+                secret_data = {}
+                secret_data[vault_key_path] = base64.b64encode(
+                    json.dumps(secrets).encode("utf-8")
```
nemacysts, line 468: might be good to have a comment for what `secrets` looks like

additionally, is there any way that this is something that's not json-serializable?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -396,6 +418,79 @@ def sync_secrets(
...
+    secret_provider_name: str,
+    vault_cluster_config: Dict[str, str],
+    soa_dir: str,
+    vault_token_file: str,
+) -> bool:
+    config_loader = PaastaServiceConfigLoader(service=service, soa_dir=soa_dir)
+    for instance_config in config_loader.instance_configs(
+        cluster=cluster, instance_type_class=KubernetesDeploymentConfig
+    ):
+        # expects VAULT_ADDR_OVERRIDE, VAULT_CA_OVERRIDE, and VAULT_TOKEN_OVERRIDE to be set
```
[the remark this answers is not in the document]
RobbySingh: ++ found a possible implementation: https://stackoverflow.com/questions/2059482/temporarily-modify-the-current-processs-environment. were you thinking we add this in as hardcoded values for now, or add them as keyword arguments to the paasta_secrets_sync script?

alternatively, I was thinking we do it this way. let me know if I'm mistaken:

```
# https://sourcegraph.yelpcorp.com/sysgit/puppet/-/blob/modules/profile_kube/manifests/management.pp?L169:178
command                => 'bash -o pipefail -c "[...] | xargs VAULT_ADDR_OVERRIDE=<> VAULT_CA_OVERRIDE=<> VAULT_TOKEN_OVERRIDE=<> paasta_secrets_sync --secret_type database-credentials; fi"', 
```

then the variable is only set during that single bash run

running it manually would require us to add `--secret_type database-credentials` as well since this vault shard has different data from the main vault shard for boto/crypto keys

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
RobbySingh: ++ will add a comment

for our use case, it'll always be JSON serializable, though it wouldn't hurt to add a check if you suggest

paasta_tools/kubernetes_tools.py
```diff
@@ -1516,8 +1516,63 @@ def get_pod_volumes(
...
+            # note an array is returned so we extend vs append
+            pod_volumes.extend(database_credentials_secrets_volumes)
+
         return pod_volumes
 
+    def get_database_credentials(self) -> List[str]:
+        database_credentials = self.config_dict.get("database_credentials", {})
+        return database_credentials
+
+    def get_database_credentials_secret_name(self, dstore="", credential="") -> str:
```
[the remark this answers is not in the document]
RobbySingh: ++ they should be required, will remove the defaults

we need the dstore/cred to make a unique secret_name. contention in naming might cause undesired effects

paasta_tools/secret_providers/vault.py
RobbySingh: ++ will make the change

paasta_tools/secret_providers/vault.py
RobbySingh: `None` would break the json serialization for the base64 dump me thinks. better to just return {} instead since if the secret is empty, it's not as catastrophic as if the secrets_sync script breaks. {} just gives us that serialization and still tells us it's empty at the same time. but open to suggestions

empty path shouldn't be passed for our use case but technically hvac.read doesn't care either given it's array of supported use cases https://hvac.readthedocs.io/en/stable/source/hvac_v1.html#hvac.v1.Client.read

what do you think is best?

paasta_tools/secret_providers/vault.py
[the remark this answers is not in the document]
RobbySingh: when I was checking the return result of .read, it actually returns something like this:

```
$ python
Python 3.7.13 (default, Apr 24 2022, 01:06:43)
[GCC 11.2.0] on linux
>>> import hvac
>>> client = hvac.Client()
>>> client.read("secrets/database/mysql/example_happyhour")
{'request_id': '3d5ecae5-24a3-7f7e-58d9-161d7533d157', 'lease_id': '', 'renewable': False, 'lease_duration': 2764800, 'data': {'master': {'passwd': '****', 'user': 'v-approle-mysql-serv-nVcYexH95A2'}, 'reporting': {'passwd': '****', 'user': 'v-approle-mysql-serv-GgCpRIh9Ut7'}, 'slave': {'passwd': '****', 'user': 'v-approle-mysql-serv-PzjPwqNMbqu'}}, 'wrap_info': None, 'warnings': None, 'auth': None}
```

so I guess the 'data' variable name isn't named well. might be better to call it 'entry', which has a key called 'data' as well as additional metadata

so for all use cases, this .read should correctly return only the data from this path

k8s_itests/deployments/kind/cluster-devbox.yaml
```diff
@@ -7,6 +7,14 @@ nodes:
   - role: worker
     labels:
       yelp.com/pool: default
+    # this mount is needed for services which rely on /nail/etc/{ecosystem,runtime}
```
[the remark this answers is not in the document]
nemacysts: nah, this is fine :)

paasta_tools/secret_providers/vault.py
nemacysts: imo, I'd return `None` -> document that this function will return `None` if the secret doesn't exist -> check for None wherever we call this function

afaict, we could technically have an empty secret, but the current code would mean there's no way to distinguish between an empty secret and a key that doesn't exist

RobbySingh: ++ makes sense to me, will make the change

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -396,6 +434,86 @@ def sync_secrets(
...
+                    "vault_auth_method": "token",
+                    "vault_token_file": vault_token_file,
+                },
+            )
+
+        for dstore, credentials in database_credentials.items():
+            for credential in credentials:
+                vault_path = f"secrets/database/{dstore}/{credential}"
+                secrets = provider.get_data_from_vault_path(vault_path)
+                if not secrets:
```
RobbySingh, line 476: @nemacysts if I understand correctly, this should be changed to "if secrets is None", thus allowing "secrets = {}" to be synced to K8s?

RobbySingh:
I ran a fresh make setup-kubernetes-job with the vault credentials set in etc_paasta_playground.vault.json under database_credentials_vault_env_overrides (just vault_addr_override and vault_token_override) and confirmed that the env worked as expected

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -313,12 +337,26 @@ def sync_all_secrets(
             )
         )
 
+        sync_service_secrets["database-credentials"].append(
```
mattmb, line 340: @RobbySingh @nemacysts I'm going to insist that we split this cron job up now. I think I mentioned it in the last addition (boto or crypto?). At a minimum let's put this new addition into a separate cron job. It will make them easier to debug, we can set different alerting/monitoring thresholds etc. And most importantly when one of these breaks or slows down it won't then impact the others.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -396,6 +434,86 @@ def sync_secrets(
...
+    database_credentials_vault_overrides = (
+        system_paasta_config.get_database_credentials_vault_overrides()
+    )
+
+    for instance_config in config_loader.instance_configs(
+        cluster=cluster, instance_type_class=KubernetesDeploymentConfig
+    ):
+        database_credentials = instance_config.get_database_credentials()
+        with set_env(**database_credentials_vault_overrides):
+            # expects VAULT_ADDR_OVERRIDE, VAULT_CA_OVERRIDE, and VAULT_TOKEN_OVERRIDE to be set
```
mattmb, line 457: This can go away too if you move it to its own cron job? Or that plus a small change to vault_tools to support non standard vault shards?

mattmb (changes requested):

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
RobbySingh: > move to its own cron job
definitely possible if we set the environment variables to the scope of the bash command, though @nemacysts advised this may enter some confusing territory when running manually and advised against it: https://github.com/Yelp/paasta/pull/3655#discussion_r1268381304

> a small change to vault_tools
unfortunately this wouldn't be small by my estimates. if you go through the vault_tools/paasta_secret.py file, you'll see the concept of ecosystem baked in everywhere and hardcoded paths and naming conventions. for fear of doing this incorrectly, we opted to use the overrides, which is a feature of the library anyway:
https://github.com/Yelp/paasta/pull/3655#discussion_r1272745254

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
RobbySingh: ++ this is something Luis mentioned when we were designing it. may need his help in how to refactor this as I'm not too sure how it should look or how this is used. I only know there's a puppet call to it on a CRON anyway: https://sourcegraph.yelpcorp.com/search?q=context:global+repo:puppet+paasta_secrets_sync&patternType=standard&sm=0

mattmb: Great, I'm sure he'll be happy to help but feel free to have a crack yourself, it should make your release workflow safer too.

High level it's going to be: duplicate this script and just keep your part and any boilerplate/argparse etc. Then if you search the name of this script in this repo you'll find refs to it that are needed in places like setup.py and some of the debian tooling to create an entrypoint in the package. Once you have that you should be able to build a deb and test the script. Then you can merge and release safely knowing that the existing stuff is unaffected. Then iterate etc. and when happy with the script you can duplicate the thing you found in puppet to create a new cron job

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
RobbySingh: ++ I'm sure vault_tools will need to be refactored at some point if more teams start requesting custom vault shards. I filed an onpoint ticket with PE-auto for something unrelated and highlighted the outdated docs

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nemacysts: we chatted internally about this and I have a PR for splitting up the cronjobs using the work that @nurdann introduced in their paasta secrets refactor a while back - so we should be fine here since this is simply adding another `--secret-type` option :)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nemacysts: @mattmb perhaps we could refactor this file a bit and move all the helper functions for each secret type to their own file? that would leave the existing file as just the driver/entrypoint for everything?

we might want to do that in a separate PR so that this diff doesn't become impossible to read tho :p

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nemacysts: depends on what y'all are expecting - if for the DRE case a secret not existing and a secret being empty (or I guess malformed if empty is not a valid value) are the same, then we can leave this line as-is :)

nemacysts (commented):
i still have to formulate some thoughts/suggestions re: tests - but for now i've left some other comments so that you're not blocked waiting for me to review

tests/kubernetes/bin/test_paasta_secrets_sync.py
```diff
@@ -827,6 +828,81 @@ def crypto_keys_patches():
...
+            mock_create_kubernetes_secret_signature,
+            mock_update_secret,
+            mock_update_kubernetes_secret_signature,
+            mock_config_loader,
+            mock_config_loader.return_value.instance_configs,
+            mock_get_secret_provider,
+        )
+
+
+def test_sync_database_secrets(database_credentials_patches):
```
nemacysts, line 870: we probably want to add a couple more tests here (e.g., all the other edge cases like no db credentials configured, db credentials configured but no update required, etc)

additionally, should this test be asserting more than just `sync_database_credentials` returns `True`?

tests/test_kubernetes_tools.py
```diff
@@ -1422,6 +1422,41 @@ def test_get_boto_volume(
...
+            soa_dir="/nail/blah",
+        )
+
+        volumes = deployment.get_database_credentials_secrets_volumes()
+        if config_dict:
+            assert volumes is not None
+            total_volumes = 0
+            for dstore in config_dict["database_credentials"]:
+                total_volumes += len(config_dict["database_credentials"][dstore])
+            assert len(volumes) == total_volumes
```
nemacysts, line 1456: can we add the expected amount of volumes to the fixture? that would allow us to remove this for-loop and simplify this check to 
```
assert len(volumes) == expected_num_volumes
```

ideally we'd also split this up into two tests so that we can get rid of the conditional in this test, but it looks like I missed asking for that refactor for boto_keys so it's not the end of the world if the conditional logic stays since it's not inconsistent with other tests (and worst case, an enterprising yelper can simply refactor all these cases in the future)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -396,6 +434,86 @@ def sync_secrets(
...
+    secret_provider_name: str,
+    vault_cluster_config: Dict[str, str],
+    soa_dir: str,
+    vault_token_file: str,
+) -> bool:
+    config_loader = PaastaServiceConfigLoader(service=service, soa_dir=soa_dir)
+    system_paasta_config = load_system_paasta_config()
+    database_credentials_vault_overrides = (
+        system_paasta_config.get_database_credentials_vault_overrides()
+    )
```
nemacysts, line 450: thinking about this more: can we set a default value for this? or should we add a check here that `database_credentials_vault_overrides` is not empty?

otherwise, what would happen if this gets called without the overrides?

paasta_tools/kubernetes_tools.py
```diff
@@ -1516,8 +1516,63 @@ def get_pod_volumes(
         if crypto_volume:
             pod_volumes.append(crypto_volume)
 
+        database_credentials_secrets_volumes = (
+            self.get_database_credentials_secrets_volumes()
+        )
+        if database_credentials_secrets_volumes:
+            # note an array is returned so we extend vs append
+            pod_volumes.extend(database_credentials_secrets_volumes)
```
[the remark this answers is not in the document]
RobbySingh: not saying this is correct, but @flopex and I figured it's best to keep the secrets in separate Kubernetes secret configs and map each individually. this also allows each secret to have it's own nested path (ie `/dstore/<mysql or cassandra>/<credential name>/<secret path>)

might be wrong, starting to forget the rationale but open to changing it if adding too many volumes is incorrect

if this just means we turn it into a for loop like in the secret_volumes, that is fair too. extend was just cleaner

paasta_tools/kubernetes_tools.py
```diff
@@ -1516,8 +1516,63 @@ def get_pod_volumes(
...
+
         return pod_volumes
 
+    def get_database_credentials(self) -> List[str]:
+        database_credentials = self.config_dict.get("database_credentials", {})
+        return database_credentials
```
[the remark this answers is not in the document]
RobbySingh: ++ good catch

paasta_tools/kubernetes_tools.py
```diff
@@ -1516,8 +1516,63 @@ def get_pod_volumes(
...
+
+    def get_database_credentials_secret_name(self, dstore, credential) -> str:
```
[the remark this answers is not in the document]
RobbySingh: I'm open to it! though we've unfortunately hardcoded it in multiple places, I'll have to refer to flopez on updating this

eg vault paths, yelp_conn code, terraform stuff?

paasta_tools/secret_providers/vault.py
```diff
@@ -155,6 +155,22 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
         else:
             return None
 
+    def get_data_from_vault_path(self, path) -> Optional[str]:
```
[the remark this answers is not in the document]
RobbySingh: ++

tests/kubernetes/bin/test_paasta_secrets_sync.py
RobbySingh: ++ good point, will add more cases

tests/test_kubernetes_tools.py
RobbySingh: ++ will simplify and split

enterprising refactorers are gems. if we find one, I'll name my firstborn after them

tests/kubernetes/bin/test_paasta_secrets_sync.py
RobbySingh: just went through the other sync_* methods - looks like the convention has been to just always return True? any other scenario is just a runtime error :shrug:

RobbySingh: guess I'll assert the number of times create_or_update_secret is called

paasta_tools/kubernetes_tools.py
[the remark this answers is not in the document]
nemacysts: afaik, we should still be able to give these their own paths using https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#project-secret-keys-to-specific-file-paths

if possible, it'd be nice to continue the 1 volumes -> N secrets that we do for most secrets (secret_volumes is an outlier here, but that's because every secret there can have a different mount path)

my understanding is that we should be able set the `mountPath` config to `/datastore` and then every secret that's mounted in `items[]` uses the `path` config to finish out the `/datastore/.../...` path

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nemacysts: might be worth testing the empty dict case in dev to make sure it won't cause chaos should this get deployed to prod without this config :)

(or crash early here if this is an empty dict)

paasta_tools/kubernetes_tools.py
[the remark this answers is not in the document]
nemacysts: i mentioned this in slack, but just scribing here as well: i was referring more to the variable name - but it might be more descriptive to also have the actual paths be something more verbose :)

paasta_tools/kubernetes_tools.py
```diff
@@ -1516,8 +1516,70 @@ def get_pod_volumes(
...
 
+        datastore_credentials_secrets_volumes = (
+            self.get_datastore_credentials_secrets_volumes()
+        )
+        for volume in datastore_credentials_secrets_volumes:
+            pod_volumes.append(volume)
+
         return pod_volumes
 
+    def get_datastore_credentials(self) -> Dict:
```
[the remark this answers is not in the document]
nemacysts: and re: usage: I'm assuming we'll also need to plumb through support for tron?

paasta_tools/secret_providers/vault.py
```diff
@@ -155,6 +155,22 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
         else:
             return None
 
+    def get_data_from_vault_path(self, path: str) -> Optional[dict]:
+        # clients.read returns None if not set
+        # if it is set, it returns an object with { **metadata, data: {} }
```
nemacysts, line 160: just double-checking for the benefit of anyone re-using this function in the future: would vault let us store a single string at a given path? e.g., should this be `Optional[Union[Dict, str]]` or something more permissive?

paasta_tools/utils.py
```diff
@@ -2562,6 +2562,13 @@ def get_topology_spread_constraints(self) -> List[Dict[str, Any]]:
         """List of TopologySpreadConstraints that will be applied to all Pods in the cluster"""
         return self.config_dict.get("topology_spread_constraints", [])
 
+    def get_datastore_credentials_vault_overrides(self) -> Optional[str]:
```
nemacysts, line 2565: I *think*
```suggestion
    def get_datastore_credentials_vault_overrides(self) -> Dict[str, str]:
```
might be the right type here? there's no way for this to return `None` and we generally expect a dict of strings to strings, right?

paasta_tools/kubernetes_tools.py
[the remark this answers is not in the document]
nemacysts: i.e., my concern isn't extend() vs append() - but having N volumes vs just 1 overall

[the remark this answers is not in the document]
RobbySingh: happy to refactor! let me know which is preferred

paasta_tools/kubernetes_tools.py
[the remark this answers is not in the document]
RobbySingh: ++ we changed everything to datastore where applicable. makes more sense!

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
RobbySingh: I tested it - empty dict doesn't crash but the script crashes anyway because it cannot connect to Vault. ie the provider calls are no-ops (`self.clients[self.ecosystems[0]].read` <-- becomes `None.read`, throws exception)

this is a good error though? given it's in a CRON, not like it'll crash PaaSTA services. or should it fail gracefully on no-connection to Vault?

paasta_tools/secret_providers/vault.py
RobbySingh: hmm let me see -

according to the Vault API docs, this should always return a dict of key value pairs (ie str:str): https://developer.hashicorp.com/vault/api-docs/secret/kv/kv-v1

though there's a chance this just returns an empty dict

can't see a scenario where it would return just [str], so I imagine this is just `Dict`

paasta_tools/utils.py
RobbySingh: not sure what Dict[str, str] implies - would this mean it's always non-empty? I get confused since we could also return an empty dict

paasta_tools/kubernetes_tools.py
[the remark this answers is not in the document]
RobbySingh: ahh good catch re: kubernetes_schema. didn't know that existed

dict would look like this:

```
datastore_credentials:
    mysql:
        - db1
        - db2
    cassandra:
        - db3
        - db4
```

so an optional key with nested keys with arrays of strings 🧠 🍳

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -396,6 +440,88 @@ def sync_secrets(
...
 
+def sync_datastore_credentials(
+    kube_client: KubeClient,
+    cluster: str,
+    service: str,
+    secret_provider_name: str,
+    vault_cluster_config: Dict[str, str],
+    soa_dir: str,
+    vault_token_file: str,
+) -> bool:
```
jfongatyelp, line 451: Apologies if I've missed this, but our other `sync_X` functions take in an optional `namespace` since the `paasta_secrets_sync` script can take it in as an argument ( eg https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes/bin/paasta_secrets_sync.py#L335 + https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes/bin/paasta_secrets_sync.py#L393).

Is there a specific reason database credentials won't need this/will _only_ ever sync to the service instance's namespace? 

That may become an issue when using db creds in e.g. tron jobs or spark jobs?

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
RobbySingh: asked them again: https://yelp.slack.com/archives/CCYF30A1K/p1691092164669369

think that should be sufficient for resolving this thread - feel free

paasta_tools/kubernetes_tools.py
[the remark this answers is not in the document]
nemacysts: in that case, a TypedDict like
class DatastoreCredentialsConfig(TypedDict, total=False):
    mysql: List[str]
    cassandra: List[str]
    ...

might be nice as a retval type to be more self-documenting :)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
RobbySingh: we eventually call `namespace=instance_config.get_namespace()` in the `create_or_update_k8s_secret(` call, which should reference the same I think? looks like sync_boto and sync_crypto do the same

paasta_tools/secret_providers/vault.py
nemacysts: gotcha, in that case `Optional[Dict[str, str]]` might be better than `Optional[dict]` :)

paasta_tools/utils.py
nemacysts: ah, sorry - Dict[str, str] just means that if the dict is not empty, the keys will always be strings and the key-values will also be strings (afaik, there's no way in mypy to type that a dict will never be empty)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nemacysts: oh, good catch!

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nemacysts: given that this script is doing somewhat non-standard things (i.e., not reading from our normal vault), i'd probably err towards the error coming from somewhere closer to the misconfiguration - otherwise, someone getting paged would need to follow the code all the way up to this overrides section and realize that something has gone wrong with the config

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
jfongatyelp: Unless we're overriding the `instance_config`'s namespace with the one provided as an argument to this script, I suspect that boto & crypto secrets are also never syncing to the requested namespace if one is provided (entirely possible, I think there is more work to be done to support boto_keys in tron for example, so we wouldn't have discovered this yet!).  We also pass a namespace rather than use the instance config's for spark secrets.

In retrospect, I do know that we currently have someone working on a known issue with https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes/bin/paasta_secrets_sync.py#L147C5-L147C48 in that it also does not honor the passed namespace, so for now, I think `instance_config.get_namespace()` is fine and should cover the default usecase, just like with boto/crypto.  You can follow https://jira.yelpcorp.com/browse/COMPINFRA-2860 for updates to that issue.

Just something to be aware about that we may be changing this soon and be warned if any other workloads in non-standard namespaces start to need these secrets!

RobbySingh: ah I see, I can imagine this affecting us as well for non-default-namespaced paasta services.

let me know if I should make that change as part of this PR or leave that to the COMPINFRA ticket's refactor

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -486,6 +486,20 @@
                         }
                     }
                 },
+                "datastore_credentials": {
```
RobbySingh, line 489: @nemacysts how does this look? I didn't include 'cassandra' since that hasnt been cepped yet. but additionalProperties: True should account for it ?

RobbySingh: also, all databases begin with an alphabetical character before going alphanumerical + underscores, so let me know if the pattern looks good

RobbySingh: spoke with @flopex who mentioned it should just be [a-zA-Z0-9_.-]+

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
jfongatyelp: I think one quick improvement that will help us prep for the future and will not have to change too much in the future is to have your `sync_datastore_credentials` accept a `namespace` and default it to the `override_namespace` arg provided in the `sync_all_secrets` function.  If it is none/empty, use `instance_config.get_namespace()` just as you do already.

RobbySingh: the var seems to be named `overwrite_namespace` - might be a phonetic typo of override? 😄

jfongatyelp: Yup, apologies, I meant `overwrite_namespace`! this one https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes/bin/paasta_secrets_sync.py#L263

RobbySingh: done! thanks for pointing this out

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -107,6 +114,21 @@ def parse_args() -> argparse.Namespace:
     return args
 
 
+@contextlib.contextmanager
+def set_env(**environ) -> None:
```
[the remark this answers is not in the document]
RobbySingh: updated typing, feel free to revisit

paasta_tools/kubernetes_tools.py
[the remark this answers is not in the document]
RobbySingh: opted to leave out cassandra for now as it was a proof of concept for the NoSQL team's benefit. the actual naming convention might change when they're ready to add it

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
RobbySingh: technically the error started at the `provider = get_secret_provider(` line since it doesn't do a test connection imo. maybe that's where it should go ?

otherwise, what I can do is something like:

```
 client = self.clients[self.ecosystems[0]]

 if client is None:
    return None <-- this happens during invalid connections since self.clients['devc'] is set to None
 result = client.read(...)
```

I'll lean towards the second approach but maybe long term, the providers have a "test connection" check?

RobbySingh: ++ added a check for client is not None. will leave this unresolved to give you a chance to assess (though these "outdated" code blocks are quite confusing in github)

paasta_tools/cli/schemas/kubernetes_schema.json
RobbySingh: ++ to strictness

>  are we envisioning that those need to do anything substantially different from the existing code? 
nope, will work seamlessly. they'll just have to add their associated key here to the schema. they *may* choose to do things differently though (eg vault paths), so I kept it largely open-ended as a proof of concept for NoSQL/other datastores but ++ to leaving it out then! ty

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -447,8 +574,6 @@ def sync_crypto_secrets(
         if not secret_data:
             continue
 
-        # In order to prevent slamming the k8s API, add some artificial delay here
-        time.sleep(0.3)
```
RobbySingh, line 451: @nemacysts note I refactored this out and added an associated time.sleep mock in the tests to consider it. didn't make sense to call it anywhere but the function. in the future, the function can be throttled using something similar to a <rate_limit> annotation eg https://pypi.org/project/ratelimit/

paasta_tools/secret_providers/vault.py
```diff
@@ -155,6 +155,29 @@ def get_secret_signature_from_data(self, data: Mapping[str, Any]) -> Optional[st
...
+    def get_data_from_vault_path(self, path: str) -> Optional[Dict[str, str]]:
+        # clients.read returns None if not set
+        # if it is set, it returns an object with { **metadata, data: {} }
+        # eg lease_id, request_id, etc. we only care about 'data' here
+
+        client = self.clients[self.ecosystems[0]]
+
+        # there is a chance client is None (ie if the connection is invalid)
+        if client is None:
+            return None
```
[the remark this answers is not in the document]
RobbySingh: ++ this was my thought too. likewise, I know this is yelpy code, but line 18 of this file could also have had a log.warning attached to it. it would be something like "warning, could not import yelp libraries. operating in standalone mode with no vault implementations"

thoughts?

will add an exception at this line though as requested

paasta_tools/kubernetes_tools.py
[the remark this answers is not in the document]
RobbySingh: ++ refactored as requested

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
RobbySingh: raised an exception on client is None in another thread. resolving this one

paasta_tools/cli/schemas/kubernetes_schema.json
RobbySingh: resolving - feel free to bring up again

paasta_tools/kubernetes_tools.py
```diff
@@ -1516,8 +1520,69 @@ def get_pod_volumes(
...
+        return limit_size_with_hash(
+            name=f"{prefix}-{service_name}-{instance_name}".replace("_", "--")
+        )
+
+    def get_datastore_credentials_secrets_volume(self) -> List[V1Volume]:
+        """
+        All credentials are stored in 1 Kubernetes Secret, which are mapped on an item->path
+        structure to /datastore/<datastore>/<credential>/<password file>.
+        """
+        datastore_credentials = self.get_datastore_credentials()
```
nemacysts, line 1559: just as below, I think we'll want to do what we do for boto/crypto secrets and ensure that the secret exists (by checking for the existence of the signature)

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nemacysts: ack - if we're not expecting the vault process that's manipulating this data to ever get us into this scenario, then that seems fine to me

(if we're not sure, it'd be nice to add the check to avoid a page on our end :))

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -396,6 +441,88 @@ def sync_secrets(
...
+    vault_token_file: str,
+    overwrite_namespace: Optional[str] = None,
+) -> bool:
+    """
+    Map all the passwords requested for this service-instance to a single Kubernetes Secret store.
+    Volume mounts will then map the associated secrets to their associated mount paths.
+    """
+    config_loader = PaastaServiceConfigLoader(service=service, soa_dir=soa_dir)
+    system_paasta_config = load_system_paasta_config()
+    datastore_credentials_vault_overrides = (
```
nemacysts, line 460: i was thinking that we'd add a check for
```
if not datastore_credentials_vault_overrides:
    raise Exception("Datastore credentials vault configuration is missing - refusing to continue.")
```
over checking in the secret provider - but I'll try to get my vault-tools refactor out and refactor this later :)

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -486,6 +486,20 @@
...
+                    "type": "object",
+                    "additionalProperties": true,
```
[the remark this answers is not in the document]
RobbySingh: ++ agreed

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
RobbySingh: maybe it should return false instead? odd how the sync scripts all return true and only true

I worry saying "overrides not existing is an error" - at which case, was it really an override or a requirement? without these overrides, it basically communicates with the default vault, which would have been the ideal situation but there are hard requirements on our vault infra for data storage

given this project is open source, maybe having it return False makes more sense?

paasta_tools/kubernetes_tools.py
```diff
@@ -1516,8 +1520,69 @@ def get_pod_volumes(
...
+            "datastore-credentials",
+            self.get_service(),
+            self.get_instance(),
+        )
+
+    def get_datastore_credentials_volume_name(self) -> str:
+        """
+        Volume names must abide to DNS mappings of 63 chars or less, so we limit it here and replace _ with --.
+        """
+        prefix = "datastore-credentials-volume"
```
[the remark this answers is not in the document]
RobbySingh: ++ agreed

paasta_tools/kubernetes_tools.py
```diff
@@ -1663,6 +1728,16 @@ def get_volume_mounts(
                         break
                 volume_mounts.append(mount)
 
+        datastore_credentials = self.config_dict.get("datastore_credentials", {})
+        if datastore_credentials:
+            volume_mounts.append(
```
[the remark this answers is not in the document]
RobbySingh: ++ good catch!

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
nemacysts: ah true, this would leak our internal constraints - we can definitely leave the code as-is tho :)

i think ideally we'd refactor the sync scripts a bit to make them not (as you've pointed out) unconditionally return true + handle errors in a more consistent way - but that's more work than you'd probably want to do now :p

RobbySingh: yep for sure, it would be ideal to refactor all of them. I'm always conflicted to fix up code smells in a feature branch. might make more sense to refactor them all as a group in a separate PR

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -99,7 +101,13 @@ def parse_args() -> argparse.Namespace:
...
+        choices=[
+            "all",
+            "paasta-secret",
+            "boto-key",
+            "crypto-key",
+            "datastore-credentials",
+        ],
         default="all",
         type=str,
         help="Define which type of secret to add/update. Default is 'all'",
```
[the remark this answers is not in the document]
RobbySingh: ++ fair point! will add that it. what I've gleaned from the conversations with @nemacysts and @mattmb is there may be a move away from using 'all' and separating the CRONs entirely! not sure if I paraphrased correctly though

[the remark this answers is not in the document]
jfongatyelp: That's quite true, since these are now split out, we should be able to remove the 'all' option entirely Soon:tm:, though don't know exactly when that will happen. Feel free to ignore me if we think we're near that point already.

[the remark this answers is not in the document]
nemacysts: I'm thinking of making a PR next week once the split cronjobs have been live for a couple days - but we've been using N different cronjobs since yesterday with no issue so far :)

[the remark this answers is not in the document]
jfongatyelp: In that case, resolving this comment 👍

tests/kubernetes/bin/test_paasta_secrets_sync.py
```diff
@@ -610,10 +620,7 @@ def boto_keys_patches():
     ) as mock_update_kubernetes_secret_signature, mock.patch(
         "paasta_tools.kubernetes.bin.paasta_secrets_sync.PaastaServiceConfigLoader",
         autospec=True,
-    ) as mock_config_loader, mock.patch(
```
RobbySingh, line 613: woops didnt mean to delete this

RobbySingh: actually yes I did. time.sleep is a fixture now

nemacysts (approved):

ilkinmammadzada (approved):

PR #3658, opened 2023-07-20
Fixing setup_kubernetes_cr exception

wilmer05:
Fixing setup_kubernetes_cr exception due to missing objects

[PAASTA-17925](https://jira.yelpcorp.com/browse/PAASTA-17925)

### DIFF
--- a/paasta_tools/setup_kubernetes_cr.py
+++ b/paasta_tools/setup_kubernetes_cr.py
@@ -155,6 +155,7 @@ def setup_all_custom_resources(
 ) -> bool:
 
     got_results = False
+    succeeded = False
     # We support two versions due to our upgrade to 1.22
     # this functions runs succefully when any of the two apiextensions
     # succeed to update the CRDs as the cluster could be in any version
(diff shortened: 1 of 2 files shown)


paasta_tools/setup_kubernetes_cr.py
```diff
@@ -162,12 +163,19 @@ def setup_all_custom_resources(
...
             ).items
-        }
+        except ApiException:
+            log.debug(
+                "Listing CRDs with apiextensions/v1beta1 not supported on this cluster, falling back to v1"
+            )
+            crds_list = []
+
+        cluster_crds = {crd.spec.names.kind for crd in crds_list}
+        succeeded = False
```
nemacysts, line 178: should we be setting the initial value for `succeeded` outside this loop?

otherwise, a cluster that supports v1 and not v1beta1 will always return False, right?

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -162,12 +163,19 @@ def setup_all_custom_resources(
...
-            for crd in apiextension.list_custom_resource_definition(
+
+        try:
+            crds_list = apiextension.list_custom_resource_definition(
                 label_selector=paasta_prefixed("service")
             ).items
-        }
+        except ApiException:
+            log.debug(
+                "Listing CRDs with apiextensions/v1beta1 not supported on this cluster, falling back to v1"
```
nemacysts, line 173: this is the other way around in this function, right? we try v1 first and then fall back to v1beta1?

paasta_tools/setup_kubernetes_cr.py
wilmer05: my mistake, yes you're right :)

paasta_tools/setup_kubernetes_cr.py
wilmer05: you're right, actually we'll try/catch both ubt I agree we fallback to v1beta1

nemacysts (approved):

PR #3663, opened 2023-07-28
COMPINFRA-2938: Support security-check in parallel steps

jfongatyelp:
Since we added `parallel` step support to `deploy.yaml`, `paasta check` has been complaining about pipelines that run the `security-check` in parallel with anything else (which is actually what we provide as a default nowadays!).

This updates `paasta check` to now also look at parallel steps to verify we have a security-check step, along with a test to verify.

### DIFF
--- a/paasta_tools/cli/cmds/check.py
+++ b/paasta_tools/cli/cmds/check.py
@@ -77,7 +77,13 @@ def deploy_check(service_path):
 
 def deploy_has_security_check(service, soa_dir):
     pipeline = get_pipeline_config(service=service, soa_dir=soa_dir)
-    steps = [step["step"] for step in pipeline]
+    steps = [step["step"] for step in pipeline if not step.get("parallel")]
+    steps += [
+        substep["step"]
+        for step in pipeline
+        if step.get("parallel")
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/check.py
```diff
@@ -77,7 +77,14 @@ def deploy_check(service_path):
...
     pipeline = get_pipeline_config(service=service, soa_dir=soa_dir)
-    steps = [step["step"] for step in pipeline]
+    steps = [step["step"] for step in pipeline if not step.get("parallel")]
+    steps += [
+        substep["step"]
+        for step in pipeline
+        if step.get("parallel")
+        for substep in step.get("parallel")
+    ]
+    print(steps)
```
nemacysts, line 87:
```suggestion
```

nemacysts (approved):

paasta_tools/cli/cmds/check.py
jfongatyelp: Oops hadn't pushed latest, thanks!

KaspariK (approved):

nemacysts (approved):

PR #3664, opened 2023-07-31
Don't hard-code legacy yelpcorp PyPI registry, remove pip-custom-platform

chriskuehl:
Internal ticket: CORESERV-12777

The general idea is to (1) remove pip-custom-platform, (2) stop hard-coding the PyPI registry URL in places where it's possible, and (3) hard-code the platformed registry at least if we have to.

We set the registry URL globally in `/etc/pip.conf` on Yelp hosts and inside Yelp's Docker images, so we should be able to get away with not specifying it in most locations. This also allows us to take advantage of the platform-specific registries which no longer need pip-custom-platform. Unfortunately it is not so easy to rely on this for many of the Docker images so I had to hard-code the platform-specific registry.

Same idea as https://github.com/Yelp/Tron/pull/921.

Passed locally on a yelpcorp host:

* [x] make test
* [x] make itest
* [ ] make k8s_itests
* [x] make itest_jammy

### DIFF
--- a/yelp_package/dockerfiles/xenial/Dockerfile
+++ b/yelp_package/dockerfiles/xenial/Dockerfile
@@ -15,7 +15,7 @@
 ARG DOCKER_REGISTRY=docker-dev.yelpcorp.com/
 FROM ${DOCKER_REGISTRY}xenial_pkgbuild
 
-ARG PIP_INDEX_URL=https://pypi.yelpcorp.com/simple
+ARG PIP_INDEX_URL=https://pypi.yelpcorp.com/xenial/simple
 ENV PIP_INDEX_URL=$PIP_INDEX_URL
 
 RUN echo "deb http://repos.mesosphere.com/ubuntu xenial main" > /etc/apt/sources.list.d/mesosphere.list && \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,6 @@ jobs:
           - py37-linux,docs,mypy,tests
           - general_itests
     env:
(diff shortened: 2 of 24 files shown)


yelp_package/dockerfiles/xenial/Dockerfile
```diff
@@ -15,7 +15,7 @@
 ARG DOCKER_REGISTRY=docker-dev.yelpcorp.com/
 FROM ${DOCKER_REGISTRY}xenial_pkgbuild
 
-ARG PIP_INDEX_URL=https://pypi.yelpcorp.com/simple
+ARG PIP_INDEX_URL=https://pypi.yelpcorp.com/xenial/simple
```
danielpops, line 18: are these still needed (here and elsewhere)?

danielpops (approved):

yelp_package/dockerfiles/xenial/Dockerfile
danielpops: Re-read the description and I see, this is needed and intentionally here (though I'm not 100% clear why)

chriskuehl: I think it may not be needed on this one since it uses our internal pkgbuild image, but for most of these it is still needed due to using the public Ubuntu base images. (Even though they all pull from our registry, they're pulling mirrored Ubuntu images as opposed to our internal modified Ubuntu images.)

In the near future (hopefully this quarter) we should be able to consolidate on the main root registry again (and without requiring pip-custom-platform) so we can do a followup round later to clean all of these up even more.

PR #3667, opened 2023-08-01
Add Filtering Options for Kubernetes Pod Selection

athy125:
The parse_k8s_api_performance_options() function can include new command-line arguments, namely --pod-status and --container-name, which allow users to filter pods by their status (e.g., Running, Pending, Succeeded, etc.) and by the name of their containers, respectively.

The get_pods_from_namespace_selector() function will be updated to support these filtering options. It constructs the appropriate field selectors to filter pods based on the provided criteria. The field selectors then can be used in the Kubernetes API request to retrieve pods matching the specified label selector and the additional filter criteria.


mattmb:

PR #3669, opened 2023-08-08
Adding shared sectret k8s namespaces

cuza:


EvanKrall (approved):

paasta_tools/utils.py
```diff
@@ -147,6 +147,10 @@
     "nrtsearchservice": "paasta-nrtsearchservices",
 }
 
+SHARED_SECRETS_K8S_NAMESPACES = {
+    "paasta-spark",
+}
+
```
jfongatyelp, line 153: We likely also need the cassandra namespace (See COMPINFRA-2934)

PR #3677, opened 2023-08-14
Added --docker-shm-size for spark-run

nlspn:
Added a pass through argument for docker run --shm-size to the spark-run command. Allows the user to set a value for `docker run --shm-size` via `paasta spark-run` using the `--docker-shm-size` option.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -174,7 +174,7 @@ def add_subparser(subparsers):
         "--docker-memory-limit",
         help=(
             "Set docker memory limit. Should be greater than driver memory. Defaults to 2x spark.driver.memory. Example: 2g, 500m, Max: 64g"
-            "Note: If memory limit provided is greater than associated with the batch instance, it will default to max memory of the box."
+            " Note: If memory limit provided is greater than associated with the batch instance, it will default to max memory of the box."
         ),
         default=None,
     )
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -185,6 +186,16 @@ def add_subparser(subparsers):
         ),
         default=None,
     )
+    list_parser.add_argument(
+        "--docker-shm-size",
+        help=(
+            "Set docker shared memory size limit. This is the same as setting docker run --shm-size and is mounted to /dev/shm in the container."
+            "Anything written to the shared memory mount point counts towards the docker memory limit. Therefore, this should be less than --docker-memory-limit"
```
88manpreet, line 193: What are the model training specific use-cases or category of use-cases, where users would want to set this option?
What are we lacking in status-quo that we require this option to be exposed? 
Any recommended values for users. E.g. 8g or 256m for most users.

nlspn: @88manpreet

- For model trainer it is probably mostly used to increase the size of the `/dev/shm` mount so that Tensorflow has enough shared memory for the NCCL protocol (for inter GPU communication and data sharing of the weights and gradient updates).
- We are unable to set the size of `/dev/shm/` in the docker container with the current version of spark-run.
- I have not been able to find in hard recommendations other than maybe >1g for NCCL applications. I don't know, its a fairly general parameter that could be used in other cases, its hard to give a blanket recommendation.

Are you wanting me to add additional info to the help string or are you just asking those questions to provide additional detail in the PR?

88manpreet: The general idea to get more context is:
* As much as possible users shouldn't be worrying knowing desired resources. And if they have to, there is some clear documentation of at least the new resource configs.
* Ideally I am not in favor of any new resource configs if possible. Here it seems like this is required for GPU training models. 
* More for my own learning on how this overlaps with executor memory or disk memory or relation with spark configs.
This can help us debug, build clear mental models or provide better recommendations.
Example: Is this part of the disk (EBS) memory or executor memory (Instance storage). Would it be a part of executor memory or separate?
 If this is separate, this can can impact how the memory usage is divided among the executors? 
> 1g is a good estimate for us.
How would the user determine that they would need to add more and by how much? Some metric? 

Additional info in help string is definitely helpful from a user point of view. Underneath magic details are helpful for both us and users :)

nlspn: @88manpreet Hows this for the help string? (I haven't push it yet):
```
"Set docker shared memory size limit for the driver's container. This is the same as setting docker run --shm-size and the shared memory is mounted to /dev/shm in the container."
"Anything written to the shared memory mount point counts towards the docker memory limit for the driver's container. Therefore, this should be less than --docker-memory-limit."
f"Defaults to {DEFAULT_DOCKER_SHM_SIZE}. Example: 8g, 256m"
f"Note: this option is mainly useful when training TensorFlow models locally, in the driver, with multiple GPUs using NCCL. The default value of {DEFAULT_DOCKER_SHM_SIZE} is typically not large
"enough for inter-gpu communication. A minimum value of >=1g is recommended for this use case but probably does not need to be greater than 1g per GPU used in the driver container."
"If you are observing low, average GPU utilization during epoch training (<65%-70%) try increasing this value; you may be resource constrained when GPUs sync training weights between mini-batches."
"This option also adds the --ulimit memlock=-1 to the docker run command since this is recommended for TensorFlow applications that use NCCL."
"Please refer to docker run documentation for more details on --shm-size and --ulimit memlock=-1."
```

88manpreet: The context looks good to me and is quite helpful. Thanks. 
I had few minor questions I left on slack channel. But ok to ship-it once added in help-doc.

88manpreet (approved):
Once the context is added as part of the help section.

88manpreet:
One point I realized, wouldn't this feature be needed in Jupyter Notebooks using GPU pools?

nlspn:
> One point I realized, wouldn't this feature be needed in Jupyter Notebooks using GPU pools?

This is only really meant for when a model is being trained in the driver. The typical work flow for Jupyterhub is to request a single gpu equipped executor with stage-level scheduling for the NN training stage. In that case, the shared memory size can be controlled with executor level spark configs passed to the spark session.

PR #3678, opened 2023-08-17
Creating an instance type class for EKS - PAASTA-17991

EmanElsaban:
I'm reusing the same config Dict ``KubernetesDeploymentConfigDict`` for ``EksDeploymentConfig`` as we want this class to be similar to ``KubernetesDeploymentConfig``

### Testing
unit tests for eks_tools.py


nemacysts (approved):
this'll be hard to test until we get further along in adding support for this instance type, but the scaffolding looks right to me :)

PR #3688, opened 2023-08-29
[COREJAVA-800] Implement active requests autoscaler

sclg-yelp:


paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -228,6 +231,119 @@ def should_create_piscina_scaling_rule(
...
+    instance: str,
+    autoscaling_config: AutoscalingParamsDict,
+    paasta_cluster: str,
+    namespace: str = "paasta",
+) -> PrometheusAdapterRule:
+    """
+    Creates a Prometheus adapter rule config for a given service instance.
+    """
+    desired_active_requests_per_replica = autoscaling_config[
+        "desired_active_requests_per_replica"
```
nemacysts, line 257: should there be a default for this? would also maybe make this more ergonomic to use (i.e., a dev wouldn't need to switch to this provider -> run validate -> be told they're missing config)

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -228,6 +231,119 @@ def should_create_piscina_scaling_rule(
...
+            0
+        )
+    """
+    total_load = f"""
+    (
+        sum(
+            {load_per_instance}
+        ) by (kube_deployment)
+        +
+        {missing_instances}
```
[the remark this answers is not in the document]
nemacysts: I'm not sure if we should add a TODO here to do something better, remove the missing instances addition, or figure out if we can do better now.

(the intent behind `missing_instances` is that every instance that's not contributing metrics is assumed to be so overloaded that it can't even report metics - i.e., at 100% utilization)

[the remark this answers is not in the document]
sclg-yelp: I see. In that case, we should probably do a `missing_instances * desired_active_requests_per_replica` to account for maximum load.

paasta_tools/setup_prometheus_adapter_config.py
sclg-yelp: I was thinking about this, but I wasn't sure what a sensible default would be other than 1 😕 We really can't say how many requests a pod can handle. I am sure a pod can handle atleast 1 😃

paasta_tools/cli/cmds/validate.py
```diff
@@ -580,6 +580,21 @@ def validate_autoscaling_configs(service_path):
...
+                    "active-requests",
+                }:
+                    desired_active_requests_per_replica = autoscaling_params.get(
+                        "desired_active_requests_per_replica", 0
+                    )
+                    if desired_active_requests_per_replica <= 0:
+                        returncode = False
+                        print(
+                            failure(
+                                msg="Autoscaling configuration is invalid: desired_active_requests_per_replica must be "
```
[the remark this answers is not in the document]
sclg-yelp: Thanks for pointing this out, will add.

paasta_tools/setup_prometheus_adapter_config.py
sclg-yelp: Also just to understand the flow, when will the checks in validate be run? We have added a condition to fail the configuration if the config is missing or less than zero.

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -228,6 +231,119 @@ def should_create_piscina_scaling_rule(
...
+    instance: str,
+    autoscaling_config: AutoscalingParamsDict,
+    paasta_cluster: str,
+    namespace: str = "paasta",
+) -> PrometheusAdapterRule:
+    """
+    Creates a Prometheus adapter rule config for a given service instance.
+    """
+    desired_active_requests_per_replica = autoscaling_config.get(
+        "desired_active_requests_per_replica", 1
```
nemacysts, line 257: this `1` should probably be a constant (that we also use in validate.py) - otherwise, we're liable to have different "defaults" (which is the case right now :p)

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -204,6 +204,9 @@
                         "decision_policy": {
                             "type": "string"
                         },
+                        "desired_active_requests_per_replica": {
```
nemacysts, line 207: we'll also want to update the list at line 196 with the new provider

paasta_tools/cli/cmds/validate.py
```diff
@@ -580,6 +583,22 @@ def validate_autoscaling_configs(service_path):
                 is False
             ):
                 autoscaling_params = instance_config.get_autoscaling_params()
+                if autoscaling_params["metrics_provider"] in {
+                    "active-requests",
+                }:
```
nemacysts, line 588: this is fine as-is, but technically
```suggestion
                if autoscaling_params["metrics_provider"] == "active-requests":
```
would be a little more accurate if we're thinking that there won't be another metrics provider with these same params :)

nemacysts (approved):
lgtm :)

nemacysts (approved):

nemacysts (approved):

PR #3694, opened 2023-09-06
PAASTA-18017: Read secret where instance service is defined, rather than service defined in CLI args.

Molaire:
Ticket: PAASTA-18017

This fixes the issue of secret usage when the "service: something" override is used.

Test simply confirms that service name while decrypting is based on the instance_config service name, not the service name given by the CLI args

tested on dev6:
test action (did not use private.py since there is no devc version): https://fluffy.yelpcorp.com/i/lK6F4r99Sm0sHmjWhHNS3BdC2tdBVdvv.html
result: https://fluffy.yelpcorp.com/i/XBRnjNSccFTMpr5kvrkr0g7zGmqgCLCw.html

### DIFF
--- a/tests/cli/test_cmds_local_run.py
+++ b/tests/cli/test_cmds_local_run.py
@@ -2081,9 +2081,6 @@ def test_run_docker_container_assume_aws_role(
     assert 0 == return_code
 
 
-fake_service_names = ["fake_service", "override_service_name"]
-
-
 @mock.patch(
     "paasta_tools.cli.cmds.local_run.pick_random_port",
     autospec=True,
@@ -2116,7 +2113,13 @@ fake_service_names = ["fake_service", "override_service_name"]
     autospec=None,
 )
 @mock.patch("os.makedirs", autospec=True)
-@mark.parametrize("override_service_name", fake_service_names)
+@mark.parametrize(
+    "original_service_name, override_service_name",
+    (
+        ("fake_service", "fake_service"),  # no service override
+        ("fake_service", "super_fake_service"),  # service override
+    ),
+)
 def test_run_docker_container_secret_volumes(
     mock_os_makedirs,
     mock_open,
@@ -2127,6 +2130,7 @@ def test_run_docker_container_secret_volumes(
     mock_execlpe,
     mock_get_docker_run_cmd,
     mock_pick_random_port,
+    original_service_name,
     override_service_name,
 ):
     mock_docker_client = mock.MagicMock(spec_set=docker.Client)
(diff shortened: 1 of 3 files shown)


.hypothesis/examples/0aba9ae8360068e8/eafddc3bf5741f53
nemacysts, line 1: we might be missing a rule in our gitignore - can you remove this file from your branch?

nemacysts (commented):
small nit re: tests - but not necessarily a blocker (but let's delete the binary file no matter what :p)

.hypothesis/examples/0aba9ae8360068e8/eafddc3bf5741f53
Molaire: oops, :ty:

tests/cli/test_cmds_local_run.py
```diff
@@ -2123,15 +2124,22 @@ def test_run_docker_container_secret_volumes(
...
         spec_set=docker.Client.remove_container
     )
+
     mock_service_manifest = mock.MagicMock(spec=MarathonServiceConfig)
     mock_service_manifest.cluster = "fake_cluster"
+    service = "fake_service"
+    different_service = "different_service"
+    mock_service_manifest.get_service = mock.MagicMock(
+        return_value=different_service if use_different_instance_service else service
+    )
```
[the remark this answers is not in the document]
Molaire: ack, I just didn't want to copy-paste the original service name so that's why I went with a condition

tests/cli/test_cmds_local_run.py
```diff
@@ -2081,6 +2081,9 @@ def test_run_docker_container_assume_aws_role(
     assert 0 == return_code
 
 
+fake_service_names = ["fake_service", "override_service_name"]
```
nemacysts, line 2084: i'd suggest inlining this just to be consistent with how we generally parametrize :)

nemacysts (approved):

tests/cli/test_cmds_local_run.py
Molaire: I can either 
a. parametrize inline and hardcode `service="fake_service"` within the test, 
b. add `service="fake_service"` to the start of the file and replace fake_service everywhere
c. keep it as-is

what's the best in your opinion?

nemacysts: maybe something like:
```python
@mark.parametrize("original_service_name, override_service_name", (
    ("fake_service", "fake_service"), # no service override
    ("fake_service", "super_fake_service"), # service override
))
...
mock_service_manifest.get_service = mock.MagicMock(
        return_value=override_service_name
    )
...
    return_code = run_docker_container(
        docker_client=mock_docker_client,
        service=original_service_name,
        ...
    )
...
assert decrypt_kwargs["service_name"] == override_service_name
```
would work?

Molaire: Thanks, didn't think of having both of em in the parameters :doh:

PR #3682, opened 2023-08-22
Bump service-configuration-lib to v2.18.6

edingroot:
## Description
To mainly include the following changes:
* Assign spark.app.id to a more intuitive name
* Pick spark ui port with preference
* Temporary adding `--add-id` argument for explicitly specifying spark.app.id when anything goes wrong

Full list: https://github.com/Yelp/service_configuration_lib/compare/v2.18.0...v2.18.6

## Test
```bash
. ./.tox/py38-linux/bin/activate
paasta spark-run --aws-credentials-yaml=/etc/boto_cfg/mrjob.yaml --cmd "spark-submit /work/integration_tests/s3.py"
```
Spark UI:  

image

With `--app-id` provided:  

image

PR #3695, opened 2023-09-06
MLCOMPUTE-746 | Remove the spark app-id option

edingroot:
We added the option for explicitly specifying Spark app id in #3682 as a safety measure for rolling out the new app id naming convention.

Now the new naming convention has been fully rolled out without any problem and no one is currently using the option. So I'm going to remove the option to cleanup the logic and avoid problems caused by user explicitly set the id.


88manpreet (approved):
Nice.

PR #3708, opened 2023-09-21
Ensure secret_volume volumes are not optional

nemacysts:
We noticed some logging from a newly added use of secret_volume that makes us think that the default for these is that V1SecretVolumeSources are optional - let's clear up any ambiguity and explicitly set a value for this

NOTE: I'm not quite sure if this will require a big bounce of anything using secret_volume - but I guess that's also less of a concern than a big bounce of *everything*


jfongatyelp (approved):

nemacysts:
(i'll merge this on monday)

EvanKrall (approved):

PR #3711, opened 2023-09-27
Run tox envs in parallel in GHA tests

nemacysts:
This will have two side effects:
1) it'll be easier to find what failed (rather than having to wade
   through an ocean of python setup logs)
2) hopefully make things faster since we'll do all these tests in
   parallel.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,10 @@ jobs:
       fail-fast: false
       matrix:
         toxenv:
-          - py38-linux,docs,mypy,tests
+          - py38-linux
+          - docs
+          - mypy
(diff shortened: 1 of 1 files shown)


jfongatyelp (approved):
+++++++

EvanKrall (approved):
I think I clumped these together to save overall build time on Travis (#1337) -- we were bottlenecked by the 4 simultaneous builders for the whole Yelp org on open source Travis, so reducing builder setup time was a win overall.

On GHA I think it should be fine to break these up?

nemacysts:
@EvanKrall yea - check out my recent thread in #opensource: we have plenty of headroom right now :)

PR #3713, opened 2023-09-28
Set DNSPolicy for spark executors

nemacysts:
This is required by the PaaSTA Workload contract and should lower the incidence of weird DNS-related issues.


nemacysts:
hmm, this doesn't seem to be getting applied for some reason

gmdfalk (approved):

nemacysts:
that said, @chelspea has a way of doing this with spark args - so I'll go ahead and close this :)

PR #3714, opened 2023-09-29
Validate iam_role and stop supporting kiam

jfongatyelp:
Since we removed kiam from our environments, we can stop supporting it and set the default of `iam_role_provider` to be `aws` now (similar to how this is already configured for tron)

In addition, that means iam_role must always be an AWS IAM role ARN, which has some very specific structure.

This applies the valid patterns per [aws's docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-names) and adds some minor tests to check a few different cases

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -629,7 +629,8 @@
                 },
                 "iam_role": {
                     "type": "string",
-                    "pattern": "^arn:aws:iam::[0-9]+:role/[a-zA-Z0-9+=,.@_-]+$"
+                    "pattern": "^arn:aws:iam::[0-9]+:role/[a-zA-Z0-9+=,.@_-]+$",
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -628,11 +628,11 @@
                     "type": "array"
                 },
                 "iam_role": {
-                    "type": "string"
+                    "type": "string",
+                    "pattern": "^arn:aws:iam::[0-9]+:role/[a-zA-Z0-9+=,.@_-]+$"
```
nemacysts, line 632: should we link to https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html#reference_iam-quotas-names above this (here and in the tron schema)?

i can't really see AWS changing this format at all, but might be worth it just in case :p

nemacysts (approved):

nemacysts (approved):

PR #3716, opened 2023-10-05
[COREJAVA-995] Refactor to pass instance_config to all scaling rules

sclg-yelp:
## Details
This is purely refactor PR to setup passing of instance_config to all scaling rules methods. 

The instance_config will be used later in `create_instance_active_requests_scaling_rule` to read registrations configured in yelpsoa configs. 

## Testing
Updated the unit tests.

## JIRA
COREJAVA-995

### DIFF
--- a/paasta_tools/setup_prometheus_adapter_config.py
+++ b/paasta_tools/setup_prometheus_adapter_config.py
@@ -61,6 +61,7 @@ from paasta_tools.long_running_service_tools import (
 from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import get_services_for_cluster
+from paasta_tools.utils import InstanceConfig_T
 from paasta_tools.utils import load_system_paasta_config
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -347,14 +348,15 @@ def create_instance_active_requests_scaling_rule(
...
+    instance_config: KubernetesDeploymentConfig,
     paasta_cluster: str,
-    namespace: str = "paasta",
 ) -> PrometheusAdapterRule:
     """
     Creates a Prometheus adapter rule config for a given service instance.
     """
+    autoscaling_config = instance_config.get_autoscaling_params()
+    instance = instance_config.instance
+    namespace = instance_config.get_namespace() or "paasta"
```
nemacysts, line 359: "paasta" is already the default (and get_namespace should always return a non-empty string)
```suggestion
    namespace = instance_config.get_namespace()
```

nemacysts: (we'll want to do this change here and in the other places we grab the namespace)

nemacysts (commented):
nice! just a minor comment re: how we're pulling out the namespace - but otherwise looks good :)

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -248,14 +248,15 @@ def should_create_active_requests_scaling_rule(
 
 def create_instance_active_requests_scaling_rule(
     service: str,
-    instance: str,
-    autoscaling_config: AutoscalingParamsDict,
+    instance_config: KubernetesDeploymentConfig,
     paasta_cluster: str,
-    namespace: str = "paasta",
```
sclg-yelp, line 254: @nemacysts thanks for the review. but as you can see here, the default value is actually set the method level but not at a level before. Because I have removed the default value, I was doing the 

```
instance_config.get_namespace() or "paasta"
```

paasta_tools/setup_prometheus_adapter_config.py
sclg-yelp: Done.

nemacysts (approved):

PR #3718, opened 2023-10-06
Sync paasta secrets for eks instances as well

nemacysts:
We could also teach this to only sync secrets for self-managed OR EKS clusters individually - but that seems like more effort (and there's no real harm to syncing these to both: in fact, it makes rolling back even easier since there's no need to wait for secrets to re-sync)

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -35,6 +35,7 @@ from typing import Tuple
 from kubernetes.client.rest import ApiException
 from typing_extensions import Literal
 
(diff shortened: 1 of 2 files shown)


nemacysts:
I still need to test this in kubestage - but I'm somewhat confident that this should work so I figured i'd put this up for review now since the tests are passing :)

jfongatyelp (approved):

nemacysts:
just tested this: 
```
INFO:__main__:Generating adapter config from soaconfigs.
INFO:__main__:Generated adapter config from soaconfigs.
INFO:__main__:Existing config matches soaconfigs - exiting.
```

nemacysts:
tested this on kubestage and it was a noop as expected

PR #3719, opened 2023-10-06
Sync autoscaling rules for EKS instances

nemacysts:
Same as with secrets, we could just do this for the set of things that are only on EKS OR self-managed k8s - but it doesn't seem worth the effort and will make it easier to flip back and forth between the two

(same as with the secrets PR - I need to run this on kubestage, but I'm pretty confident this is fine :))

### DIFF
--- a/paasta_tools/setup_prometheus_adapter_config.py
+++ b/paasta_tools/setup_prometheus_adapter_config.py
@@ -31,6 +31,7 @@ from kubernetes.client import V1ObjectMeta
 from kubernetes.client.rest import ApiException
 from mypy_extensions import TypedDict
 
+from paasta_tools.eks_tools import EksDeploymentConfig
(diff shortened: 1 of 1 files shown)


jfongatyelp (approved):
Do we expect paasta_cluster in the prometheus side to be the `eks-*` prefixed version for dual clusters?

nemacysts:
> Do we expect paasta_cluster in the prometheus side to be the `eks-*` prefixed version for dual clusters?

from what I'd been reading/chatting about internally with folks: on the prometheus side the paasta_cluster label should be the same for dual-clusters (but I can check in when folks are around tomorrow/in EU working hours)

nemacysts:
just tested this:
```
INFO:__main__:Generating adapter config from soaconfigs.
INFO:__main__:Generated adapter config from soaconfigs.
INFO:__main__:Existing config matches soaconfigs - exiting.```

PR #3720, opened 2023-10-07
Change the default k8s namespace for kubernetes services.

EvanKrall:
Incidentally, I spotted and fixed a bug in `paasta secret decrypt` where it wasn't passing the correct arguments to `get_secret` in certain situations.

Since `paasta` is no longer a reasonable default value for namespace, I deleted that default value for all the functions I could find that take namespace as an argument.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ test-not-yelpy: .paasta/bin/activate
 	.paasta/bin/tox -e tests
 
 quick-test: .tox/py38-linux
-	TZ=UTC .tox/py38-linux/bin/py.test --last-failed -x -- tests
+	TZ=UTC .tox/py38-linux/bin/py.test --failed-first -x --disable-warnings -- tests
 
 .tox/py38-linux: .paasta/bin/activate
 	.paasta/bin/tox
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -423,7 +423,8 @@ def paasta_secret(args):
                 get_secret(
                     kube_client,
                     get_paasta_secret_name(namespace, service, args.secret_name),
-                    namespace,
+                    key_name=args.secret_name,
+                    namespace=namespace,
                 )
             )
         # fallback to default in case mapping fails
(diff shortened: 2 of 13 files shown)


paasta_tools/cli/cmds/secret.py
```diff
@@ -423,7 +423,8 @@ def paasta_secret(args):
                 get_secret(
                     kube_client,
                     get_paasta_secret_name(namespace, service, args.secret_name),
-                    namespace,
+                    key_name=args.secret_name,
+                    namespace=namespace,
```
EvanKrall, line 427: @nurdann Can you take a look at the changes in this file and make sure they make sense? I was able to reproduce this bug locally by setting `secrets_for_owner_team: true` in my local yelpsoa-configs' `paasta-contract-monitor/service.yaml` and then running `paasta secret decrypt -s paasta-contract-monitor -n goss -c norcal-devc -y ~/pg/yelpsoa-configs/`.  This fails with 

```
  File "/opt/venvs/paasta-tools/bin/paasta", line 11, in <module>
    sys.exit(main())
  File "/opt/venvs/paasta-tools/lib/python3.8/site-packages/paasta_tools/cli/cli.py", line 256, in main
  File "/opt/venvs/paasta-tools/lib/python3.8/site-packages/paasta_tools/cli/cmds/secret.py", line 432, in paasta_secret
    get_secret(
  File "/opt/venvs/paasta-tools/lib/python3.8/site-packages/paasta_tools/kubernetes_tools.py", line 4157, in get_secret
KeyError: 'paasta'
```

By running `/nail/home/krall/pg/paasta/.tox/py38-linux/bin/paasta secret decrypt -s paasta-contract-monitor -n goss -c norcal-devc -y ~/pg/yelpsoa-configs/` (i.e. using this branch's version of secret.py), I get the value of the secret as desired.

Makefile
```diff
@@ -50,7 +50,7 @@ test-not-yelpy: .paasta/bin/activate
 	.paasta/bin/tox -e tests
 
 quick-test: .tox/py38-linux
-	TZ=UTC .tox/py38-linux/bin/py.test --last-failed -x -- tests
+	TZ=UTC .tox/py38-linux/bin/py.test --failed-first -x --disable-warnings -- tests
```
EvanKrall, line 53: I use `make quick-test` for iterative test fixing, and `--last-failed` would exit after successfully running one test that had previously failed. With `--failed-first`, if the previously-failing test succeeds, it will continue on to run all the other tests.

Without `--disable-warnings`, we have several pages of warnings (mostly from libraries we import) that mean you need to scroll up to see the traceback of the failed test.

nemacysts (approved):

PR #3722, opened 2023-10-11
Use docker_client.logs instead of docker_client.attach during paasta local-run

danielpops:
When running `paasta local-run` in unprivileged containers mode, `docker_client.attach` would cause paasta to crash with

```
Your service is now running! Tailing stdout and stderr:
Exception in thread Thread-1:
  File "/usr/lib/python3.8/http/client.py", line 556, in _get_chunk_left
    chunk_left = self._read_next_chunk_size()
  File "/usr/lib/python3.8/http/client.py", line 523, in _read_next_chunk_size
    return int(line, 16)
ValueError: invalid literal for int() with base 16: b''

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.8/http/client.py", line 588, in _readinto_chunked
    chunk_left = self._get_chunk_left()
  File "/usr/lib/python3.8/http/client.py", line 558, in _get_chunk_left
    raise IncompleteRead(b'')
http.client.IncompleteRead: IncompleteRead(0 bytes read)
```

This implies some kind of API incompatibility between the 2015 vintage of `docker-py`'s `docker_client.attach` with the set of parameters being passed to it and podman. By changing to `docker_client.logs` we maintain parity with the status quo log output and circumvent the crash.

FWIW, `docker_client.logs` is used elsewhere in `local-run` while doing healthchecks (added in https://github.com/Yelp/paasta/pull/2074 and there is some discussion about threads). TBH I'm not 100% clear why it was using `logs` there and `attach` here in the first place, or if we need to consider the same thread concerns addressed in the healthchecking code, since once we launch the paasta workload, we're just running indefinitely until the process finishes OR we terminate the process via ctrl+c, and both of those worked in my manual tests.

## Testing done
I'm going to assume that some unit tests will need updating, will circle back with those updates shortly. 

I manually tested the `paasta local-run` command in privileged docker mode without the patch and unprivileged docker mode with the patch and observed identical output.

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -1029,8 +1029,10 @@ def run_docker_container(
         running = docker_client.inspect_container(container_id)["State"]["Running"]
         if running:
             print("Your service is now running! Tailing stdout and stderr:")
-            for line in docker_client.attach(
-                container_id, stderr=True, stream=True, logs=True
+            for line in docker_client.logs(
+                container_id,
+                stderr=True,
+                stream=True,
             ):
                 # writing to sys.stdout.buffer lets us write the raw bytes we
                 # get from the docker client without having to convert them to


paasta_tools/cli/cmds/local_run.py
```diff
@@ -1029,8 +1029,10 @@ def _output_exit_code():
         running = docker_client.inspect_container(container_id)["State"]["Running"]
         if running:
             print("Your service is now running! Tailing stdout and stderr:")
-            for line in docker_client.attach(
-                container_id, stderr=True, stream=True, logs=True
+            for line in docker_client.logs(
```
nemacysts, line 1032: I think that the attach is on-purpose to allow for interactive control as it also hooks up stdin (and not just stdout/err) - but @EvanKrall can probably confirm or deny that

danielpops: I see. Do you know of any use cases where interactive control is necessary?

I tracked down the original introduction of `docker_client.attach` to https://github.com/Yelp/paasta/commit/4372cc0eeeaa023152060ea440470f63f587ba09 which says it's needed when running in `interactive` mode. I confirmed that the `--interactive` mode still accepts my stdin

```
$ paasta local-run --pull -s the-service -c norcal-devc -i main --interactive
Re-executing paasta local-run --pull with sudo..
Please wait while the image (the-docker-registry.com:443/services-the-service:paasta-7d55b80e45fe62a78320cee52ef0e117b8457549) is pulled (times out after 30m)...
104f856b02a5: Already exists
21492089cc22: Already exists
d43b75cbe280: Already exists
d684ff3ca408: Already exists
bbb52d385333: Already exists
eae051c2be5a: Already exists
061b02157a44: Already exists
f704c7d4592a: Already exists
e5371db25231: Already exists
f617b7f4c20b: Already exists
64612d9e18a5: Already exists
8a6ca6f7a590: Already exists
7b9a5ab1c685: Already exists
7007490126ef: Already exists
2b0f46eb4008: Pulling fs layer
2b0f46eb4008: Download complete
2b0f46eb4008: Download complete
the-docker-registry.com:443/services-the-service:paasta-7d55b80e45fe62a78320cee52ef0e117b8457549
Running docker command:
paasta_docker_wrapper run --env PAASTA_SERVICE --env PAASTA_INSTANCE --env PAASTA_CLUSTER --env PAASTA_DEPLOY_GROUP --env PAASTA_DOCKER_IMAGE --env PAASTA_RESOURCE_CPUS --env PAASTA_RESOURCE_MEM --env PAASTA_RESOURCE_DISK --env PAASTA_GIT_SHA --env PAASTA_MONITORING_TEAM --env PAASTA_INSTANCE_TYPE --env PAASTA_SECRET_MYSQL_SESSIONDEV_MASTER_VAR --env PAASTA_SECRET_MYSQL_SESSIONRDONLY_READ_VAR --env PAASTA_SECRET_MYSQL_SESSIONRPT_REPORTING_VAR --env PAASTA_SECRET_JWT_SIGNING_KEY --env STATS_PORT --env HOST --env PAASTA_LAUNCHED_BY --env PAASTA_HOST --env PAASTA_PORT --memory=637m --memory-swap=701m --cpu-period=100000 --cpu-quota=60000 --label=paasta_service=the-service --label=paasta_instance=main --init=true --cap-drop=SETPCAP --cap-drop=MKNOD --cap-drop=AUDIT_WRITE --cap-drop=CHOWN --cap-drop=NET_RAW --cap-drop=DAC_OVERRIDE --cap-drop=FOWNER --cap-drop=FSETID --cap-drop=KILL --cap-drop=SETGID --cap-drop=SETUID --cap-drop=NET_BIND_SERVICE --cap-drop=SYS_CHROOT --cap-drop=SETFCAP --publish=33247:8888 --name=paasta_local_run_dpopes_956029 --volume=/etc/boto_cfg:/etc/boto_cfg:ro --volume=/nail/bulkdata:/nail/bulkdata:ro --volume=/nail/etc/datacenter:/nail/etc/datacenter:ro --volume=/nail/etc/ecosystem:/nail/etc/ecosystem:ro --volume=/nail/etc/habitat:/nail/etc/habitat:ro --volume=/nail/etc/kafka_discovery/:/nail/etc/kafka_discovery/:ro --volume=/nail/etc/mrjob:/nail/etc/mrjob:ro --volume=/nail/etc/region:/nail/etc/region:ro --volume=/nail/etc/runtimeenv:/nail/etc/runtimeenv:ro --volume=/nail/etc/services:/nail/etc/services:ro --volume=/nail/etc/srv-configs:/nail/etc/srv-configs:ro --volume=/nail/etc/superregion:/nail/etc/superregion:ro --volume=/nail/etc/zookeeper_discovery/:/nail/etc/zookeeper_discovery/:ro --volume=/nail/opt/data_pipeline_schemas:/nail/opt/data_pipeline_schemas:ro --volume=/nail/srv:/nail/srv:ro --volume=/var/run/synapse/services/:/var/run/synapse/services/:ro --interactive=true --tty=true the-docker-registry.com:443/services-the-service:paasta-7d55b80e45fe62a78320cee52ef0e117b8457549 sh -c bash
nobody@dev5-uswest2adevc:/code$ ls
[...]
nobody@dev5-uswest2adevc:/code$ exit
exit

```

nemacysts: i...am not actually sure :blobsweats: - that said, if `stdin` is still accepted with this patch, then this seems pretty low-risk (as a rollback is pretty trivial)

danielpops: yeah, because when run with `--interactive` it goes through a different code path that `execlpe`'s https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/local_run.py#L973

danielpops: and if you try to do something weird like `paasta local-run --pull -s the-service -c norcal-devc -i main --cmd bash` you get 

```
ta-7d55b80e45fe62a78320cee52ef0e117b8457549 sh -c bash
  File "/nail/home/dpopes/github.com/Yelp/paasta2/.tox/py38-linux/bin/paasta", line 33, in <module>
    sys.exit(load_entry_point('paasta-tools', 'console_scripts', 'paasta')())
  File "/nail/home/dpopes/github.com/Yelp/paasta2/paasta_tools/cli/cli.py", line 256, in main
  File "/nail/home/dpopes/github.com/Yelp/paasta2/paasta_tools/cli/cmds/local_run.py", line 1314, in paasta_local_run
    return configure_and_run_docker_container(
  File "/nail/home/dpopes/github.com/Yelp/paasta2/paasta_tools/cli/cmds/local_run.py", line 1208, in configure_and_run_docker_container
    return run_docker_container(
  File "/nail/home/dpopes/github.com/Yelp/paasta2/paasta_tools/cli/cmds/local_run.py", line 996, in run_docker_container
    container_id = get_container_id(docker_client, container_name)
  File "/nail/home/dpopes/github.com/Yelp/paasta2/paasta_tools/cli/cmds/local_run.py", line 611, in get_container_id
    raise LostContainerException(
paasta_tools.cli.cmds.local_run.LostContainerException: Can't find the container I just launched so I can't do anything else.
Try docker 'ps --all | grep paasta_local_run_dpopes_278957' to see where it went.
Here were all the containers:
```
in both the status quo and with this patch

danielpops: oh interestingly that is different behavior in unpriv vs priv containers. cc @oholiab FYI

nemacysts (approved):
might be worth pinging krall in slack just to make sure we're not missing anything, but lgtm :)

piax93 (approved):

paasta_tools/cli/cmds/local_run.py
oholiab: @danielpops that's actually a race condition pre-existing in the paasta codebase, I was able to run the same command and have it return so I tested it with a `bash -c "echo lol"` and it triggered the error consistently, but with `bash - "sleep 1; echo lol"` it doesn't trigger the race. In that sense I think we've just found something that's more performant in the podman setup.

IMO this is a low priority edge case. I've not looked at the code for this but I would imagine it requires us to check for an *exited* container with that ID if there's not a running one and then return the exit code and logs for that.

oholiab (approved):
lgtm

EvanKrall (approved):
I don't think I have anything to add, looks like y'all already considered everything I would've mentioned.

PR #3726, opened 2023-10-23
Adding LimitRanges for paasta managed namespaces

cuza:
Adding LimitRanges for PaaSTA managed namespaces with `ensure_namespace()`

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2713,7 +2713,7 @@ def ensure_paasta_api_rolebinding(kube_client: KubeClient, namespace: str) -> No
 
 def ensure_paasta_namespace_limits(kube_client: KubeClient, namespace: str) -> None:
     limits = get_all_limit_ranges(kube_client, namespace=namespace)
-    limits_names = [item.metadata.name for item in limits]
+    limits_names = {item.metadata.name for item in limits}
     if "limit-mem-cpu-disk-per-container" not in limits_names:
         log.warning(
             f"Creating limit: limit-mem-cpu-disk-per-container on {namespace} namespace as it does not exist"
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -72,6 +72,9 @@ from kubernetes.client import V1HTTPGetAction
 from kubernetes.client import V1KeyToPath
 from kubernetes.client import V1LabelSelector
 from kubernetes.client import V1Lifecycle
+from kubernetes.client import V1LimitRange
+from kubernetes.client import V1LimitRangeItem
+from kubernetes.client import V1LimitRangeSpec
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -2676,14 +2679,15 @@ def ensure_namespace(kube_client: KubeClient, namespace: str) -> None:
         kube_client.core.create_namespace(body=paasta_namespace)
 
     ensure_paasta_api_rolebinding(kube_client, namespace)
+    ensure_paasta_namespace_limits(kube_client, namespace)
```
nemacysts, line 2682: should we move this into the `if namespace not in namespace_names:` once this has run on all our clusters? seems a bit wasteful to always be checking limit ranges when we know that they're guaranteed to exist (once this has run at least once)

paasta_tools/kubernetes_tools.py
```diff
@@ -2707,6 +2711,39 @@ def ensure_paasta_api_rolebinding(kube_client: KubeClient, namespace: str) -> No
         )
 
 
+def ensure_paasta_namespace_limits(kube_client: KubeClient, namespace: str) -> None:
+    limits = get_all_limit_ranges(kube_client, namespace=namespace)
+    limits_names = [item.metadata.name for item in limits]
```
nemacysts, line 2716: super-nit: this probably doesn't matter since there's probably only ever going to be a single limit range, but I personally like to make sure that anything i use `in` on is a set just in case :)

```suggestion
    limits_names = {item.metadata.name for item in limits}
```

nemacysts (approved):

paasta_tools/kubernetes_tools.py
cuza: nice suggestion but we currently have a bunch of existing namespaces without limits :(

cuza: nice suggestion but we currently have a bunch of existing namespaces without limits :(

EvanKrall: Maybe? We're inevitably gonna forget about this and then we'll have namespaces that are created by Flux which don't get these limits.

nemacysts: yea, exactly - once we've released this as-is all of these existing namespace will now have limits and we can clean this up to only set limit ranges on brand new namespaces

nemacysts: whoops, looks like I responded with a stale page and never saw your comment

@EvanKrall yea, I guess it depends on how we want to treat folks manually creating namespaces (and then reusing those namespaces for paasta: do we want them to have all of paasta's defaults as well or allow for them to do their own thing?

(this bit us earlier today in #cid - they're reusing a namespace for paasta and the limitrange we added affected pre-existing workloads)

PR #3727, opened 2023-10-23
Improvements to PaaSTA Playground

cuza:
- added paasta-cronjobs make target
  - setup-kubernetes-job
  - cleanup-kubernetes-job
  - paasta-secrets-sync
  - generate_deployments_for_service
- added VSCode debugpy as dev requirement
- made paasta playground not overrwrite existing paasta confs or yelpsoa-configs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,12 +158,32 @@ setup-kubernetes-job: k8s_fake_cluster generate_deployments_for_service
 	export PAASTA_TEST_CLUSTER=kind-${USER}-k8s-test;\
 	.tox/py38-linux/bin/python -m paasta_tools.list_kubernetes_service_instances -d ./soa_config_playground --shuffle --group-lines 1 | xargs --no-run-if-empty .tox/py38-linux/bin/python -m paasta_tools.setup_kubernetes_job -d ./soa_config_playground -c kind-${USER}-k8s-test
 
+.PHONY: cleanup-kubernetes-jobs
+cleanup-kubernetes-jobs:
+	export KUBECONFIG=./k8s_itests/kubeconfig;\
+	export PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_playground/;\
+	export PAASTA_TEST_CLUSTER=kind-${USER}-k8s-test;\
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -158,13 +158,33 @@ setup-kubernetes-job: k8s_fake_cluster generate_deployments_for_service
...
+- setup-kubernetes-job
+- cleanup-kubernetes-job
+- paasta-secrets-sync
+- generate_deployments_for_service
+endef
+export ANNOUNCE_CRONS_BODY
+.PHONY: paasta-crons
+make paasta-cronjobs:
+	@echo "$$ANNOUNCE_CRONS_BODY"
+	while true; do make paasta-secrets-sync && make cleanup-kubernetes-jobs; sleep 5; done
```
nemacysts, line 186: hmm, should we be a little nicer to our devboxes and run these every minute?

and just to verify that my eyes are reading this correctly: we're running this in the foreground so upon ^C we'll stop this loop?

requirements-dev-minimal.txt
```diff
@@ -1,6 +1,7 @@
 astroid
 asynctest
 coverage
+debugpy
```
nemacysts, line 4:
```suggestion
# VSCode debugging requirement
# See https://code.visualstudio.com/docs/python/debugging#_local-script-debugging
debugpy
```

might be helpful since there's no other reference to this library in this repo so an enterprising yelper trying to cleanup the unused dependencies for this repo might remove this unknowingly :)

nemacysts (approved):
i left some comments, but not are really blockers (although, i do feel a little uneasy about setting up an infinite loop that runs every 5s on code folks might be actively iterating on)

Makefile
cuza: This is running every 5s but also is kinda lightweight since is running just using the PaaSTA playground stuff, unless you made your playground super heavy with lots of test svcs it should be fine. I also tried to avoid wait as much as possible, I don't want to wait 1m to see what happened with my yelpsoa-config change when I'm doing some rapid testing

nemacysts (approved):

PR #3728, opened 2023-10-23
Respect a service's docker_registry for adhoc spark-runs

nemacysts:
There are some services where we want to ensure that a specialized docker registry is always used - even for adhoc development runs.

This change is loosely based on what we do in push-to-registry, where we read service.yaml to see if a specialized registry needs to be used.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -17,6 +17,7 @@ from typing import Union
 
 import yaml
 from boto3.exceptions import Boto3Error
+from service_configuration_lib import read_service_configuration
 from service_configuration_lib import spark_config
 from service_configuration_lib.spark_config import get_aws_credentials
 from service_configuration_lib.spark_config import get_grafana_url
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1059,7 +1068,12 @@ def build_and_push_docker_image(args: argparse.Namespace) -> Optional[str]:
     if cook_return != 0:
         return None
 
-    docker_url = f"{args.docker_registry}/{docker_tag}"
+    registry_uri = args.docker_registry or _get_adhoc_docker_registry(
+        service=args.service,
+        soa_dir=args.yelpsoa_config_root,
+    )
```
nemacysts, line 1074: the intent here is to allow --docker-registry to override whatever is in yelpsoa

nemacysts:
(I tested this locally with the -y flag and verified that by default docker-dev is still used)

nemacysts:
(and that --docker-registry takes precedence over service.yaml)

y-trobinso (approved):

jfongatyelp (approved):

88manpreet (approved):
lgtm.
Would be useful to test once pushed out.

PR #3729, opened 2023-10-23
making skj and ckj aware of downthenup bounces across namespaces

cuza:
making `setup_kubernetes_jobs.py` and `cleanup_kubernetes_jobs.py` aware of "_downthenup_" instance types bounces across namespaces

### DIFF
--- a/paasta_tools/cleanup_kubernetes_jobs.py
+++ b/paasta_tools/cleanup_kubernetes_jobs.py
@@ -124,7 +124,9 @@ def instance_is_not_bouncing(
         if isinstance(application, DeploymentWrapper):
             existing_app = application.item
             if (
-                existing_app.metadata.namespace == instance_config.get_namespace()
+                existing_app.metadata.namespace != instance_config.get_namespace()
+                and (instance_config.get_bounce_method() == "downthenup")
+                or existing_app.metadata.namespace == instance_config.get_namespace()
                 and (
                     instance_config.get_instances()
                     <= (existing_app.status.ready_replicas or 0)
--- a/paasta_tools/cleanup_kubernetes_jobs.py
+++ b/paasta_tools/cleanup_kubernetes_jobs.py
@@ -124,12 +124,16 @@ def instance_is_not_bouncing(
         if isinstance(application, DeploymentWrapper):
             existing_app = application.item
             if (
-                existing_app.metadata.namespace != instance_config.get_namespace()
-                and (instance_config.get_bounce_method() == "downthenup")
-                or existing_app.metadata.namespace == instance_config.get_namespace()
-                and (
-                    instance_config.get_instances()
-                    <= (existing_app.status.ready_replicas or 0)
+                (
+                    existing_app.metadata.namespace != instance_config.get_namespace()
+                    and (instance_config.get_bounce_method() == "downthenup")
+                )
+                or (
+                    existing_app.metadata.namespace == instance_config.get_namespace()
+                    and (
+                        instance_config.get_instances()
+                        <= (existing_app.status.ready_replicas or 0)
+                    )
                 )
             ) or instance_config.get_desired_state() == "stop":
                 return True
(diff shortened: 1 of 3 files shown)


paasta_tools/cleanup_kubernetes_jobs.py
```diff
@@ -124,7 +124,9 @@ def instance_is_not_bouncing(
         if isinstance(application, DeploymentWrapper):
             existing_app = application.item
             if (
-                existing_app.metadata.namespace == instance_config.get_namespace()
+                existing_app.metadata.namespace != instance_config.get_namespace()
+                and (instance_config.get_bounce_method() == "downthenup")
+                or existing_app.metadata.namespace == instance_config.get_namespace()
```
EvanKrall, line 129: I can never remember the precedence of `and`/`or` - can you add some parentheses here?

paasta_tools/setup_kubernetes_job.py
```diff
@@ -266,6 +268,17 @@ def setup_kube_deployments(
...
+                        exist_on_other_namespace = False
+                        for existing_app in existing_apps:
+                            if existing_app[:2] == (
+                                app.kube_deployment.service,
+                                app.kube_deployment.instance,
+                            ):
+                                exist_on_other_namespace = True
+                                break
+                        if exist_on_other_namespace:
+                            continue
```
EvanKrall, line 281: maybe this part of the code would be easier to read as:

```
if app.soa_config.get_bounce_method() == "downthenup":
    if any((app[:2] == (app.kube_deployment.service, app.kube_deployment.instance)) for app in existing_apps):
        # For downthenup, we don't want to create until cleanup_kubernetes_job has cleaned up the instance in the other namespace.
        continue
```

paasta_tools/setup_kubernetes_job.py
```diff
@@ -233,6 +233,8 @@ def setup_kube_deployments(
     eks: bool = False,
 ) -> bool:
 
+    existing_apps = set()
+    existing_kube_deployments = set()
```
EvanKrall, line 237: mypy is complaining about these not having type definitions.

paasta_tools/setup_kubernetes_job.py
```diff
@@ -233,6 +233,8 @@ def setup_kube_deployments(
...
     if service_instance_configs_list:
```
EvanKrall, line 238: Maybe we should flip this to:

```
if not service_instance_configs_list:
    return True

existing_kube_deployments = set(list_all_paasta_deployments(kube_client))
existing_apps = {
    (deployment.service, deployment.instance, deployment.namespace)
    for deployment in existing_kube_deployments
}
```

that way you don't need the empty `set()`s

paasta_tools/setup_kubernetes_job.py
cuza: oh Pylance was complaining about not being defined , maybe I forgot to re run mypy after changing that

paasta_tools/setup_kubernetes_job.py
cuza: I don't think we should call list_all_paasta_deployments() which calls the k8s API and gets all deployments on every namespace if `service_instance_configs_list` is None or empty. 

Type hinting the empty sets sound like a better solution to me, since the only reason why I moved the definition out is cuz it was created inside an if block but referenced later outside the if

cuza: maybe something like:

```
 existing_apps: Set[Tuple[str, str, str]] = set()
 existing_kube_deployments: Set[List[KubeDeployment]] = set()
```

paasta_tools/setup_kubernetes_job.py
cuza: :TY: black this went from 3 lines to a bunch......


```python
if app.soa_config.get_bounce_method() == "downthenup":
    if any(
        (
            app[:2]
            == (
                app.kube_deployment.service,
                app.kube_deployment.instance,
            )
        )
        for app in existing_apps
    ):
        # For downthenup, we don't want to create until cleanup_kubernetes_job has cleaned up the instance in the other namespace.
        continue
```

EvanKrall (approved):

PR #3730, opened 2023-10-23
Use the correct registry_uri to check if we need to sudo

jfongatyelp:
In #3728 we changed the `--docker_registry`'s default value from our `DEFAULT_SPARK_DOCKER_REGISTRY` and instead leave it undefined by default.

However, we still used `args.docker_registry` in our check whehter we needed sudo to push ot this registry, rather than using the generated `registry_uri`, which caused all runs with `--build` to try to sudo to push the registry, which doesn't work w/ the default registry.

This fixes to use our new registry_uri which should be either the requested docker_registry or the service's default, to get back to the normal behavior.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -1080,7 +1080,7 @@ def build_and_push_docker_image(args: argparse.Namespace) -> Optional[str]:
     if retcode != 0:
         return None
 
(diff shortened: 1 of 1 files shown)


y-trobinso (approved):

nemacysts (approved):

nemacysts:
i am dumb, sorry ;_;

PR #3731, opened 2023-10-26
RFC: Add additional custom namespaces to k8s schema

nemacysts:
I see two potential solutions for allowing folks to use arbitrary namespaces:
* we gatekeep them by needing us to add their namespace to this allowlist (which lets us make sure that they've correctly labeled their namespace as well as done everything else that we expect) OR,
* we get rid of the allowlist and allow folks to put whatever namespace they'd like and expect that they know what they're doing :)

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -864,7 +864,7 @@
                 },
                 "namespace": {
                     "type": "string",
-                    "pattern": "^(paasta|paastasvc-.*)$"
+                    "pattern": "^(paasta|jenkins|paastasvc-.*)$"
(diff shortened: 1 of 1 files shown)


EmanElsaban (approved):

EmanElsaban:
I think for now option 1 seems good because it's an exception but if we saw later more cases that require other ns then maybe we allow whatever ns?

nemacysts:
I would personally prefer the second option (getting rid of the allowlist) since it's less effort, but I'm not sure what everyone else thinks :)

EmanElsaban (approved):

nemacysts:
we'll go with https://github.com/Yelp/paasta/pull/3732 instead :)

PR #3735, opened 2023-10-30
better handle CLI params for capabilities

piax93:
This avoids having the same capability added and dropped at the same time. The Docker CLI just silently accepts this, and favours the capability being added, but Podman is more picky.
This also logs a warning message when Podman is used as containerizers and capabilities are being added, as likely the user won't have permission to set capabilities (don't know if we'd rather just throw an exception in that case instead).


grimmware (commented):
lgtm!

Only think I'd ask here is whether this code is going to cause issues when running *remotely*, i.e. someone using podman locally who is actually talking to paasta on a remote machine?

oholiab:
> lgtm!
> 
> Only think I'd ask here is whether this code is going to cause issues when running _remotely_, i.e. someone using podman locally who is actually talking to paasta on a remote machine?

Sorry, logged in with the wrong account, this was me.

danielpops (approved):

nemacysts:
> lgtm!
> 
> 
> 
> Only think I'd ask here is whether this code is going to cause issues when running *remotely*, i.e. someone using podman locally who is actually talking to paasta on a remote machine?

@oholiab remote-run hasn't worked since we transitioned off of mesos, so this shouldn't be a concern

nemacysts (approved):

PR #3738, opened 2023-11-04
Fix conditions for displaying disable auto rollbacks button

jfongatyelp:
It looks like we added a new condition with the intent that we would start to rollback if any SLO or new generic 'rollback metric' started to fail.

However, a valid transition's `conditions` must *all* be true to enable the transition.  As we don't yet have any services using metric rollbacks, a service would likely only ever be failing SLOs (`any_slo_failing`) but not ever failing rollback metrics (`any_metric_failing`). 

This replaces the two separate any_*_failing conditions with the new `any_rollback_condition_failing` method which applies the intended logic already of returning true if either any_slo_failing or any_metric_failing: https://github.com/Yelp/sticht/blob/master/sticht/rollbacks/base.py#L245-L246 

This was preventing the "Disable auto-rollbacks" button to be displayed as well, as the trigger's conditions must also `all()` succeed for the button to be included in displayed buttons: https://github.com/Yelp/sticht/blob/master/sticht/slack.py#L296-L308

This also adds a simple test to verify `get_available_buttons` returns the right things for this scenario, plus ~fixes~ deletes a test that I think was not actually testing what it claimed to be (`test_paasta_mark_for_deployment_when_verify_image_succeeds` was testing the same thing as `test_paasta_mark_for_deployment_when_verify_image_fails`!)

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -986,8 +986,7 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
                 "dest": None,  # Don't actually change state, just call the before function.
                 "trigger": "disable_auto_rollbacks_button_clicked",
                 "conditions": [
-                    self.any_slo_failing,
-                    self.any_metric_failing,
+                    self.any_rollback_condition_failing,
                     self.auto_rollbacks_enabled,
                 ],
                 "before": self.disable_auto_rollbacks,
--- a/tests/cli/test_cmds_mark_for_deployment.py
+++ b/tests/cli/test_cmds_mark_for_deployment.py
@@ -878,11 +878,16 @@ def test_MarkForDeployProcess_happy_path_skips_complete_if_no_auto_rollback(
 
 
 @patch(
+    "paasta_tools.cli.cmds.mark_for_deployment.get_instance_configs_for_service_in_deploy_group_all_clusters",
+    autospec=True,
+)
+@patch(
     "paasta_tools.cli.cmds.mark_for_deployment.MarkForDeploymentProcess.any_slo_failing",
     autospec=True,
 )
 def test_MarkForDeployProcess_get_available_buttons_failing_slos_show_disable_rollback(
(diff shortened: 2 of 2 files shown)


nemacysts (approved):

nemacysts (approved):

EvanKrall (approved):
I'm not sure I fully understand what `test_paasta_mark_for_deployment_when_verify_image_succeeds_no_privs` is supposed to do / what `test_paasta_mark_for_deployment_when_verify_image_succeeds` was supposed to do, but this generally looks fine.

jfongatyelp:
> I'm not sure I fully understand what `test_paasta_mark_for_deployment_when_verify_image_succeeds_no_privs` is supposed to do / what `test_paasta_mark_for_deployment_when_verify_image_succeeds` was supposed to do, but this generally looks fine.

I'm not sure what it was _supposed_ to do but previously it seemed like it was doing exactly what `test_paasta_mark_for_deployment_when_verify_image_fails` was doing, and now instead `test_paasta_mark_for_deployment_when_verify_image_succeeds_no_privs` ostensibly makes sure we get past the `verify_image` check in mark-for-deployment, but i short-circuit the rest of it by testing w/ a user who can no longer auth.

But now I realize, it does mean that this test is now a bit flawed in that if we ever reorder the checks in m-f-d so auth comes before validate_image, this test is no longer accurate... Is that a dealbreaker?

EvanKrall:
> But now I realize, it does mean that this test is now a bit flawed in that if we ever reorder the checks in m-f-d so auth comes before validate_image, this test is no longer accurate... Is that a dealbreaker?

Seems like the `mock_is_docker_image_already_in_registry.assert_called_with` should handle that situation by making the test fail.

Maybe expand on the comment in `# we just want to test verify_image logic`? Something like `# we just want to test the verify_image logic, so we make the next check (can_user_deploy_service) fail.`

Or maybe just delete the test entirely -- `test_paasta_mark_for_deployment_when_verify_image_fails` is probably sufficient here, given that the code in question is just `if not is_docker_image_already_in_registry(...): raise ValueError(...)`.

PR #3744, opened 2023-12-11
Add support for setting lower/upper-bounds for autotuned resources

nemacysts:
At the moment this only supports a specific subset of top-level resource keys:
* cpus
* mem
* disk

Supporting {min,max}_instances should be pretty trivial (but I'm not quite sure when we'd want to use that yet, so I left those out of the allowlist).

I'm explicitly not supporting these values for sidecars since: a) it seemed like it would complicate things a bit and, b) I'm not sure we have a usecase for this :p


paasta_tools/contrib/rightsizer_soaconfigs_update.py
```diff
@@ -276,7 +279,7 @@ def get_recommendations_by_service_file(
             result["cluster"],
         )  # e.g. (foo, marathon-norcal-stagef)
         instance_type = result["cluster"].split("-", 1)[0]
-        rec: Union[KubernetesRecommendation, CassandraRecommendation]
+        rec: Union[KubernetesRecommendation, CassandraRecommendation] = {}
```
nemacysts, line 282: not super necessary, but pyright was correctly calling this out as a potentially unbound variable below (and I figured this is innocuous enough to sneak in here :p)

paasta_tools/config_utils.py
```diff
@@ -259,3 +261,82 @@ def commit_to_remote(self, extra_message: str = ""):
...
+            )
+            for instance_name, recommendation in recommendations_by_instance.items():
+                log.debug(
+                    f"Merging recommendations for {instance_name} in {service}/{AUTO_SOACONFIG_SUBDIR}/{instance_type_cluster}.yaml..."
+                )
+                existing_recommendations.setdefault(instance_name, {})
+                existing_recommendations[instance_name].update(recommendation)
+                for limit_type, limits in (
+                    existing_configs[instance_name].get("autotune_limits", {}).items()
+                ):
```
EvanKrall, line 297: Would it make sense to move this whole for loop into its own function, maybe called `clamp_recommendations`? This function is a bit long.

paasta_tools/config_utils.py
```diff
@@ -259,3 +261,82 @@ def commit_to_remote(self, extra_message: str = ""):
             _push_to_remote(self.branch)
         else:
             log.info("No files changed, no push required.")
+
+    def merge_recommendations(
+        self, recs: Dict[Tuple[str, str], Dict[str, Any]]
+    ) -> List[Dict[Tuple[str, str], Dict[str, Any]]]:
```
EvanKrall, line 267: might be nice to have a docstring explaining what all these `str`s are.

```suggestion
    ) -> List[Dict[Tuple[str, str], Dict[str, Any]]]:
    """
    :param recs: Dictionary of (service, instance_type_cluster) -> recommendations. instance_type_cluster is e.g. "kubernetes-pnw-prod".
    :returns: a dictionary of the same format, with the previous recommendations merged in and autotune_limits applied.
    """
```

paasta_tools/config_utils.py
nemacysts: yup!

paasta_tools/config_utils.py
nemacysts: can do!

jfongatyelp: maybe tangential question but is there a technical reason we are going from a Dict of (service, instance_type_cluster) -> recommendations to a list of individual one-item {(service, instance_type_cluster): recs} dicts?

jfongatyelp (approved):
our first test for autotuned defaults logic!???

paasta_tools/config_utils.py
nemacysts: @jfongatyelp not really - it just fit my mental model better while I was writing this (and make functions as pure as possible)

I can try to keep the stricture the same if we'd like

jfongatyelp: ah thanks, i was just asking for my own education. Seems fine to leave, the typing makes this difference easy to see anyhow!

nemacysts: i've refactored this a bit since upon further inspection i just needed to think a tiny bit harder to make this work with the same types :p

EvanKrall (approved):

EvanKrall (approved):

PR #3746, opened 2023-12-13
Stop deleting brutal deployments

jfongatyelp:
We previously configured `brutal` deployments already to use a `RollingUpdate` strategy w/ `max_unavailable=100%`, which should immediately kill off all old replicas & spawn a new replicaset during a rolling restart. 

This sounds like exactly what we'd want a brutal bounce to do, and should be even faster than today's brutal bounces since we won't need to wait around for paasta to delete the object & check if it has fully been deleted plus delete all pods, then finally create it again. With these configs, Kubernetes should immediately delete all pods of the old version and spawn a new replicaset of the new.

This review updates it so setup_kubernetes_job will just update the brutal deployment in the same way as all other deployments, and removes the deletion code since brutal bounces were the only thing using it. 

A manual test of a deployment with a `RolingUpdate` strategy w/ `maxUnavailable=100%` with 100 replicas showed that a change to the pod template triggered a restart and all pods in the old replicaset were removed within 4s, and new replicaset and 100 new pods were created in the same time period. 

During a status quo brutal bounce, a deletion & creation took about 7s to create all 100 new pods, so the new method still seems as 'fast' a bounce method, so still fits our "fastest bounce possible" description in docs.

### DIFF
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -1,8 +1,6 @@
 import logging
-import threading
 from abc import ABC
 from abc import abstractmethod
-from time import sleep
 from typing import Optional
 from typing import Union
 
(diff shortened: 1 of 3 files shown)


nemacysts (approved):
nothing like deleting code!

I think I'm a tad confused now with the difference between downthenup and brutal (since they seem to do the same thing?) - but this lgtm regardless of the potential duplication

jfongatyelp:
> downthenup and brutal 

Yeah it seems like it's pretty much the same behavior, although different on the k8s side... `downthenup` uses the `Recreate` strategy, which per docs claim that it makes sure all running replicas are gone _before_ any new replicas come up.

With a `RollingUpdate` and `maxUnavailable=100%`, I don't think there's necessarily a _guarantee_ that all old pods are gone before starting up. In practice it seemed to delete all my existing pods pretty much immediately, but still a minor distinction that might be important to `downthenup` users.

ilkinmammadzada (approved):

EvanKrall:
Yeah, `brutal` means you don't care about any guarantees of the order or timing of deletions and creation. In the mesos/marathon world, we accomplished this by just deleting the old marathon app and creating the new one in the same run of setup_marathon_job.

PR #3747, opened 2023-12-13
Implement monitoring check which alerts when a service is at max_instances.…

EvanKrall:
… PAASTA-18103


paasta_tools/check_autoscaler_max_instances.py
```diff
@@ -0,0 +1,159 @@
...
+            if not job_config.get_autoscaling_metric_spec(
+                name=get_kubernetes_app_name(service, instance),
+                cluster=cluster,
+                kube_client=kube_client,
+                namespace=job_config.get_namespace(),
+            ):
+                # Not an instance that uses HPA, don't check.
+                # TODO: should we send status=0 here, in case someone disables autoscaling for their service / changes
+                # to bespoke autoscaler?
+                continue
```
nemacysts, line 69: seems reasonable to me - should maybe check with PE if sending "useless" events could be a problem tho (probs not!)

paasta_tools/check_autoscaler_max_instances.py
```diff
@@ -0,0 +1,159 @@
...
+                    f"HPA {job_config.get_sanitised_deployment_name()} not found."
+                )
+                continue
+
+            if (
+                autoscaling_status["min_instances"]
+                == autoscaling_status["max_instances"]
+            ):
+                status = 0
+                output = f"Not checking {service}.{instance} as min_instances == max_instances == {autoscaling_status['max_instances']}. This is probably a canary instance."
```
nemacysts, line 91: ah, would this be when folks use yaml anchors and have the canary reuse the main config?

nemacysts: (i would not expect any canaries to have an hpa)

paasta_tools/check_autoscaler_max_instances.py
```diff
@@ -0,0 +1,159 @@
...
+            elif (
+                autoscaling_status["desired_replicas"]
+                >= autoscaling_status["max_instances"]
+            ):
+                threshold = job_config.get_autoscaling_max_instances_alert_threshold()
+                setpoint = job_config.get_autoscaling_params()["setpoint"]
+                metric_threshold_target_ratio = threshold / setpoint
+
+                status = 3
```
nemacysts, line 100: do we have named constants for these anywhere? i can personally never remember what is CRITICAL and what is WARNING 😆

paasta_tools/check_autoscaler_max_instances.py
nemacysts: i do wonder if maybe there could be a case where someone overrode max_instances and autotune has raised min_instances to the current max_instance override

paasta_tools/check_autoscaler_max_instances.py
```diff
@@ -0,0 +1,159 @@
...
+            else:
+                status = 0
+                output = f"{service}.{instance} is below max_instances."
+
+            monitoring_overrides = job_config.get_monitoring()
+            monitoring_overrides.update(
+                {
+                    "page": False,  # TODO: remove this line once this alert has been deployed for a little while.
+                    "alert_after": "0m",
+                    "realert_every": -1,
```
nemacysts, line 121: what does a -1 mean here? always alert?

paasta_tools/check_autoscaler_max_instances.py
EvanKrall: oh, yeah, sensu uses the nagios return codes: https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/pluginapi.html

Looks like these are defined in pysensu_yelp: https://github.com/Yelp/pysensu-yelp/blob/master/pysensu_yelp/__init__.py#L123

paasta_tools/check_autoscaler_max_instances.py
EvanKrall: I think sensu sends the same notification again after the time duration specified by realert_every. I guess I can probably just delete this and let it be inherited from monitoring.yaml.

paasta_tools/check_autoscaler_max_instances.py
EvanKrall: > ah, would this be when folks use yaml anchors and have the canary reuse the main config?

Yeah, yelp-main's canaries use this pattern.

> i do wonder if maybe there could be a case where someone overrode max_instances and autotune has raised min_instances to the current max_instance override

Hmm, good point. There are a couple options:

- look for `canary` in the instance name
- look for other instances in the same cluster with the same `registrations`
- check to make sure min_/max_instances are both manually set, not set in autotune_defaults

nemacysts: yea, not sure what I'd pick if I was making this change - looking for `canary` in the name is maybe the option least likely to inadvertently be triggered?

paasta_tools/check_autoscaler_max_instances.py
```diff
@@ -0,0 +1,159 @@
...
+                        status = pysensu_yelp.Status.OK
+                        output = f"{service}.{instance}: Ratio of current value to target value ({current_value} / {target_value}) is below the ratio of max_instances_alert_threshold to setpoint ({threshold} / {setpoint})"
+            else:
+                status = pysensu_yelp.Status.OK
+                output = f"{service}.{instance} is below max_instances."
+
+            monitoring_overrides = job_config.get_monitoring()
+            monitoring_overrides.update(
+                {
+                    "page": False,  # TODO: remove this line once this alert has been deployed for a little while.
```
nemacysts, line 121: might be worth ticketing this (and adding a note in the ticket that we'll need to send an email letting folks know that there's a new paging alert being added)

nemacysts (approved):
i can see folks disabling autoscaling to "fix" this alert, so it might be nice to send an OK event in that case :)

(but feel free to ship like this imo)

PR #3750, opened 2023-12-14
Remove paasta performance-check

nemacysts:
This was never really utilized as intended and the necessary infra to run this was torn down long ago - let's delete this until someone has the desire and will to bring it back :)


nemacysts:
https://github.yelpcorp.com/sysgit/yelpsoa-configs/pull/38111 should probably be merged first

PR #3752, opened 2023-12-15
Set include_smartstack default to False since we no longer support it

jfongatyelp:
We have nothing actually running smartstack so attempting to get that information in a paasta API call will fail.

While I wasn't brave enough to outright delete this code yet (it _may_ be useful to have this prior art to show how to toggle one mesh provider in mesh_status vs another?), let's at least set the default to False to prevent accidental uses which can clog the API logs with errors.

### DIFF
--- a/paasta_tools/api/api_docs/oapi.yaml
+++ b/paasta_tools/api/api_docs/oapi.yaml
@@ -1831,7 +1831,7 @@ paths:
         required: false
         schema:
           type: boolean
-          default: true
+          default: false
       - description: Include Envoy information
         in: query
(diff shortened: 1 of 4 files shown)


nemacysts (approved):
imo, deleting the code outright would not be a bad idea :)

jfongatyelp:
> imo, deleting the code outright would not be a bad idea :)

Yeah, this is true...it will never ever work anymore. I was starting down that path and realized how silly all the `include_envoy` stuff when it is the only mesh choice so backed out.  But I'll try and do this in a separate PR!

PR #3753, opened 2023-12-15
Have 'paasta validate' check autotuned_defaults/ files against their own schemas.

EvanKrall:
Also delete the marathon-related schemas.

PR #3754, opened 2023-12-15
Optimize paasta validate

EvanKrall:
This is based on https://github.com/Yelp/paasta/pull/3753 but has an additional commit that speeds up paasta validate by about 3-4x by reducing the number of times we parse yaml or generate all the TronActionConfigs for an individual service.

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -738,7 +738,7 @@ class TronJobConfig:
     def get_expected_runtime(self):
         return self.config_dict.get("expected_runtime")
 
-    def _get_action_config(self, action_name, action_dict):
+    def _get_action_config(self, action_name, action_dict) -> TronActionConfig:
         action_service = action_dict.setdefault("service", self.get_service())
         action_deploy_group = action_dict.setdefault(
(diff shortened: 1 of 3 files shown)


paasta_tools/tron_tools.py
```diff
@@ -1093,23 +1093,41 @@ def load_tron_instance_config(
...
-                    return action
+    ):
+        if action.get_instance() == instance:
+            return action
     raise NoConfigurationForServiceError(
         f"No tron configuration found for {service} {instance}"
     )
 
 
+@time_cache(ttl=5)
```
nemacysts, line 1109: this could be a standard lru_cache, right? there's no long-running paasta code that would call this otherwise, is there?

(or does the paasta api call this for paasta status at some point?)

nemacysts (approved):
++ to getting rid of `clusters=[cluster]` :p

paasta_tools/tron_tools.py
EvanKrall: Hmm, good question. lru_cache could definitely end up causing problems for any long-running code. I'm just doing the same thing we do for `load_tron_service_config`

PR #3755, opened 2023-12-16
Support preferredDuringSchedulingIgnoredDuringExecution node affinity…

EvanKrall:
…. PAASTA-18145

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -432,6 +432,101 @@
                         }
                     }
                 },
+                "node_selectors_preferred": {
+                    "type": "array",
+                    "items": {
+                        "type": "object",
+                        "properties": {
+                            "weight": {
+                                "type": "integer"
+                            },
+                            "preferences": {
+                                "type": "object",
+                                "additionalProperties": false,
+                                "patternProperties": {
+                                    "^[a-zA-Z0-9]+[a-zA-Z0-9-_./]*[a-zA-Z0-9]+$": {
+                                        "anyOf": [
+                                            {
+                                                "type": "string"
+                                            },
+                                            {
+                                                "type": "array",
+                                                "items": {
+                                                    "type": "string"
+                                                },
+                                                "uniqueItems": true
+                                            },
+                                            {
+                                                "type": "array",
+                                                "items": {
+                                                    "anyOf": [
+                                                        {
+                                                            "type": "object",
+                                                            "additionalProperties": false,
+                                                            "properties": {
+                                                                "operator": {
+                                                                    "enum": [
+                                                                        "In",
+                                                                        "NotIn"
+                                                                    ]
+                                                                },
+                                                                "values": {
+                                                                    "type": "array",
+                                                                    "items": {
+                                                                        "type": "string"
+                                                                    },
+                                                                    "uniqueItems": true
+                                                                }
+                                                            },
+                                                            "required": [
+                                                                "operator",
+                                                                "values"
+                                                            ]
+                                                        },
+                                                        {
+                                                            "type": "object",
+                                                            "additionalProperties": false,
+                                                            "properties": {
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -432,6 +432,101 @@
...
+                        "type": "object",
+                        "properties": {
+                            "weight": {
+                                "type": "integer"
+                            },
+                            "preferences": {
+                                "type": "object",
+                                "additionalProperties": false,
+                                "patternProperties": {
+                                    "^[a-zA-Z0-9]+[a-zA-Z0-9-_./]*[a-zA-Z0-9]+$": {
```
nemacysts, line 447: just double-checking: this is the k8s label format regex, right?

if so, does jsonschema have a way for us to enforce how long this property could be (since labels can only be so many characters long)?

paasta_tools/kubernetes_tools.py
```diff
@@ -610,7 +659,7 @@ def raw_selectors_to_requirements(
                 raise ValueError(
                     f"Unknown k8s node affinity operator: {config['operator']}"
                 )
-            requirements.append((label, config["operator"], values))
+            requirements.append((label, cast(str, config["operator"]), values))
```
nemacysts, line 662: huh, surprised that mypy can't tell that a Literal string is also a str

paasta_tools/kubernetes_tools.py
```diff
@@ -2330,13 +2405,14 @@ def get_node_affinity(self) -> Optional[V1NodeAffinity]:
...
-            # changing it while the pod is running will not cause an eviction.
-            # this should be fine since if there are whitelist/blacklist config
-            # changes, we will bounce anyway.
-            required_during_scheduling_ignored_during_execution=selector,
+            required_during_scheduling_ignored_during_execution=V1NodeSelector(
+                node_selector_terms=[required_term]
+            ),
+            preferred_during_scheduling_ignored_during_execution=preferred_terms
+            if preferred_terms
+            else None,
```
nemacysts, line 2415: might be nice to do something like 
```python
preferred_terms = preferred_terms or None
return V1NodeAffinity(...)
```
to avoid the formatting here, but that's not a big deal

nemacysts (approved):

paasta_tools/cli/schemas/kubernetes_schema.json
EvanKrall: hm. Probably? Label lengths are complicated though, since I think there's an overall length limit as well as a limit on the sections between the dots. I can probably find or come up with a regex for this.

nemacysts: not too big of a deal if it's difficult to do now, we can always come back to this and figure out how to represent this in jsonchema (or express this in code in `paasta validate`)

paasta_tools/kubernetes_tools.py
EvanKrall: I had a really long slack thread in #python about this, I think I'll use the [suggestion](https://yelp.slack.com/archives/CA8G37W3Y/p1702924064196629?thread_ts=1702681402.140179&cid=CA8G37W3Y) to add a type annotation to the definition of `requirements` instead.

I think the `cast` here was just a complicated way of getting mypy to infer the correct type for `requirements` anyway.

paasta_tools/cli/schemas/kubernetes_schema.json
EvanKrall: I guess I could put an overall length limit of 253+1+63 on this, but I'm not sure how valuable that is compared to a more complicated regex.

I copied this from https://github.com/Yelp/paasta/pull/3755/files#diff-3d2bc1f8f532cc633bdeb9a38648967bf24a429729d63ebe60c1551072d07dfaR355 so if we fix this, we'll want to fix this in both places.

PR #3756, opened 2023-12-18
Always use docker registry server's digest information for pushed spark images

danielpops:
## Problem
Because of https://github.com/containers/podman/issues/14779, `docker push` has different behavior w.r.t. the local vs remote image digest value when run in privileged vs unprivileged mode. This was causing `paasta spark run` commands using `--build` to fail to allocate an executor node in unprivileged container mode because it was referring to an image digest that paasta couldn't find because it only exists on the devbox where `paasta spark run` is called.

## Solution
Parse the message from the remote docker registry when running `docker push` during `paasta spark run --build` so that we always use the server version of the image digest as the source of truth.

## Testing Done
Ran `paasta spark run` manually for both `--pull` and `--build` with this patch and confirmed the executor can be allocated and the jobs can complete.

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -41,7 +41,6 @@ from paasta_tools.spark_tools import get_webui_url
 from paasta_tools.spark_tools import inject_spark_conf_str
 from paasta_tools.utils import _run
 from paasta_tools.utils import DEFAULT_SOA_DIR
-from paasta_tools.utils import get_docker_client
 from paasta_tools.utils import get_possible_launched_by_user_variable_from_env
 from paasta_tools.utils import get_username
 from paasta_tools.utils import InstanceConfig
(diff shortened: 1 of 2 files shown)


piax93 (approved):
nice workaround

piax93 (commented):
I guess there are a couple of unit tests that need to be fixed though

danielpops:
> I guess there are a couple of unit tests that need to be fixed though

ah yeah I had some pre-commit shenanigans that was preventing me from running the tests locally, I will fix these ASAP

piax93 (approved):
🚢

chelspea (approved):

88manpreet (approved):
lgtm. I assume the plan to push this is after Jan 1st, 2024.

oholiab (approved):
Amazing <3

Let's get this merged ASAP now that we're in the new year so we can start doing more aggressive rollout to spark-heavy machines

edingroot (approved):

PR #3759, opened 2023-12-27
Improve caching in setup_kubernetes_job to reduce unnecessary work. 

EvanKrall:
PAASTA-18159

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -11,6 +11,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import base64
+import functools
 import hashlib
 import itertools
 import json
@@ -501,6 +502,19 @@ class InvalidKubernetesConfig(Exception):
 
 
 class KubeClient:
+    @functools.lru_cache()  # type: ignore
+    def __new__(
+        cls,
+        component: Optional[str] = None,
+        config_file: Optional[str] = None,
+        context: Optional[str] = None,
+    ) -> "KubeClient":
+        """By @lru_cache'ing this function, repeated instantiations of KubeClient with the same arguments will return the
+        exact same object. This makes it possible to effectively cache function calls that take a KubeClient as an
+        argument."""
+        return super().__new__(cls)
+
+    @functools.lru_cache()  # type: ignore
     def __init__(
         self,
         component: Optional[str] = None,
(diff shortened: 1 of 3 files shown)


paasta_tools/docker_wrapper.py
```diff
@@ -340,15 +340,14 @@ def main(argv=None):
     # Marathon sets MESOS_TASK_ID
     mesos_task_id = env_args.get("MESOS_TASK_ID")
 
-    hostname = socket.getfqdn()
+    fqdn = socket.getfqdn()
+    hostname = fqdn.partition(".")[0]
```
nemacysts, line 344: begone flaky tests :p

paasta_tools/kubernetes_tools.py
```diff
@@ -501,6 +502,16 @@ def __init__(self, exception: Exception, service: str, instance: str) -> None:
 
 
 class KubeClient:
+    @functools.lru_cache()  # type: ignore
```
nemacysts, line 505: think it's worth leaving a comment here re: the caching? it's probably going to be somewhat surprising for future yelpers when they stumble upon this otherwise

paasta_tools/kubernetes_tools.py
```diff
@@ -2647,6 +2658,7 @@ def force_delete_pods(
         )
 
 
+@time_cache(ttl=60)
 def get_all_namespaces(
```
nemacysts, line 2662: any chance that returning a set here makes a meaningful impact on speed now that we're calling this in ensure_namespace?

nemacysts (approved):

EvanKrall:
(I'm pulling the flaky test fixes out into their own PR, https://github.com/Yelp/paasta/pull/3760)

paasta_tools/kubernetes_tools.py
EvanKrall: The line that checks membership doesn't show up at all in my py-spy traces, so probably not.

tests/conftest.py
```diff
@@ -60,6 +61,12 @@ def mock_ktools_read_soa_metadata(mock_read_soa_metadata):
         yield mock_read_soa_metadata
 
 
+@pytest.fixture(autouse=True)
```
nemacysts, line 64: might be nice to use pytest-antilru eventually, but I think last time I tried to use it it broke a bunch of tests :p

nemacysts (approved):

cuza (approved):

PR #3760, opened 2023-12-27
Attempt to make local-run itests pass on Github Actions by setting ho…

EvanKrall:
…stname to the hostname, not the FQDN

Seems weird to be setting the hostname to an FQDN anyway.


nemacysts (approved):
++++++++

PR #3765, opened 2024-01-04
Fix debian link

nemacysts:
This is what I get for naming the file differently from the command - `paasta_habitat_fixer` does not work as bin/ does not contain a paasta_habitat_fixer.py


nemacysts:
```> which paasta_habitat_fixer                                                               /usr/bin/paasta_habitat_fixer```

as expected now :)

cuza (approved):

PR #3766, opened 2024-01-04
Guess the correct context (or allow override) in paasta_habitat_fixer

nemacysts:
Turns out that I didn't read the admin kubeconf closely enough and missed that the context names there are prefixed with kubernetes-admin@

### DIFF
--- a/paasta_tools/contrib/habitat_fixer.py
+++ b/paasta_tools/contrib/habitat_fixer.py
@@ -31,18 +31,13 @@ def parse_args() -> argparse.Namespace:
     parsed_args = parser.parse_args()
 
     if not parsed_args.context:
-        if parsed_args.kubeconfig == KUBE_CONFIG_PATH:
-            # if we're using the default on a box with an admin kubeconf, we'll
-            # need to ensure that we're using the right context name - which
-            # interestingly enough has a different name than usual even though
-            # these kubeconfigs only ever have a single entry :p
-            parsed_args.context = f"kubernetes-admin@{parsed_args.cluster}"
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/habitat_fixer.py
```diff
@@ -21,11 +21,31 @@ def parse_args() -> argparse.Namespace:
...
+            # need to ensure that we're using the right context name - which
+            # interestingly enough has a different name than usual even though
+            # these kubeconfigs only ever have a single entry :p
+            parsed_args.context = f"kubernetes-admin@{parsed_args.cluster}"
+        elif parsed_args.kubeconfig == KUBE_CONFIG_USER_PATH:
+            # otherwise, the user kubeconfig context names are just the cluster names
+            parsed_args.context = parsed_args.cluster
+        else:
+            print(
+                "WARNING: unable to guess context: will fallback to KUBECONTEXT env var if present"
```
jfongatyelp, line 45: Technically it would also fall back to admin.conf's current-context if we leave it `None` right? 
```
current-context: kubernetes-admin@pnw-devc
```

jfongatyelp (approved):

paasta_tools/contrib/habitat_fixer.py
nemacysts: true!

I can switch to that

jfongatyelp: shipit either way!

jfongatyelp (approved):

PR #3768, opened 2024-01-08
Load eks- and kubernetes- files in autotune merging

nemacysts:
It's possible for a service to be in one of several states:
* 100% on Yelp-managed k8s
* 100% on AWS-managed k8s
* X% on Yelp-managed k8s and 100-X% on AWS-managed k8s

Everything is fine in the first case, but for the other two we need to make sure that we load the eks- files as we cannot actually rely on the instance_type_cluster from the autotune CSV as we "fake" the instance type there to always be kubernetes

For now, we'll basically try to load both types and merge the resulting configuration to get the current configs (which we need in order to clamp autotune values if there are user-configured limits)

### DIFF
--- a/paasta_tools/config_utils.py
+++ b/paasta_tools/config_utils.py
@@ -337,8 +337,8 @@ class AutoConfigUpdater:
             # and eks- files for the existing configs, as there are services that at any given time will
             # only exist on one of these or may have a mix (and the csv file that we get fakes the current
             # cluster type)
-            # NOTE: if an instance appears in both files, the counterpart will always "win" - this 
-            # should only be possible while an instance is being migrated from one instance type to 
+            # NOTE: if an instance appears in both files, the counterpart will always "win" - this
+            # should only be possible while an instance is being migrated from one instance type to
             # another
             instance_type, _ = instance_type_cluster.split("-", maxsplit=1)
             existing_configs = {
--- a/paasta_tools/config_utils.py
+++ b/paasta_tools/config_utils.py
@@ -337,6 +337,9 @@ class AutoConfigUpdater:
             # and eks- files for the existing configs, as there are services that at any given time will
             # only exist on one of these or may have a mix (and the csv file that we get fakes the current
             # cluster type)
+            # NOTE: if an instance appears in both files, the counterpart will always "win" - this 
+            # should only be possible while an instance is being migrated from one instance type to 
+            # another
             instance_type, _ = instance_type_cluster.split("-", maxsplit=1)
             existing_configs = {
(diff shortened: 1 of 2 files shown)


paasta_tools/config_utils.py
```diff
@@ -28,6 +28,11 @@
     "cassandracluster",
 )
 
+# this could use a better name - but basically, this is for pairs of instance types
+# where you generally want to check both types (i.e.,g a status-quo and migration
+# instance type)
+INSTANCE_TYPE_COUNTERPARTS = {"eks": "kubernetes", "kubernetes": "eks"}
```
jfongatyelp, line 34: Just to make sure I'm following, we probably won't ever be using this first item until we start having the CSV spit out 'eks-*' instance_type_cluster results, right?

paasta_tools/config_utils.py
```diff
@@ -327,10 +332,25 @@ def merge_recommendations(
...
             )
-            existing_configs = self.get_existing_configs(
-                service=service,
-                file_name=instance_type_cluster,
-            )
+            # i'm so sorry.
+            # basically, we need to make sure that for every autotuned service, we load both kubernetes-
+            # and eks- files for the existing configs, as there are services that at any given time will
+            # only exist on one of these or may have a mix (and the csv file that we get fakes the current
+            # cluster type)
```
jfongatyelp, line 339: Do we think we need to add any extra note here that the eks-* file will 'win' in the current state since it's the 2nd dict, in case we have any instances with the exact same name appear in both files? (technically since this should only ever be during migration, they should have the exact same values, so maybe not terribly important to note? 🤞 )

jfongatyelp (approved):

jfongatyelp (approved):

paasta_tools/config_utils.py
nemacysts: yea - although I'm hoping that we never get to that point and simply s/eks-/kubernetes-/ all the soaconfigs :p

but, better safe than sorry in case we keep the eks files for a bit

paasta_tools/config_utils.py
nemacysts: yea, it should only be during migration - but i can add a note

nemacysts:
```suggestion
            # only exist on one of these or may have a mix (and the csv file that we get fakes the current
            # cluster type)
            # NOTE: if an instance appears in both files, the counterpart will always "win" - this 
            # should only be possible while an instance is being migrated from one instance type to 
            # another
```

jfongatyelp (approved):

PR #3769, opened 2024-01-10
Change the default branch name to push to

bonkin:
This PR proposes the change of the default value of the `--branch` argument to make it more inclusive.


nemacysts:
we'll need to make a couple things look up the default branch and push to that, so i'll close this for now since we'll likely do this with a different approach

PR #3771, opened 2024-01-11
Minor improvements to check_autoscaler_max_instances

EvanKrall:
Before we make this paging, I'd like to reduce the frequency of realerting, and make sure this shows up in docsearch if you search for `check_autoscaler_max_instances`.

### DIFF
--- a/paasta_tools/check_autoscaler_max_instances.py
+++ b/paasta_tools/check_autoscaler_max_instances.py
@@ -136,6 +136,7 @@ async def check_max_instances(
                 {
                     "page": False,  # TODO: remove this line once this alert has been deployed for a little while.
                     "runbook": "y/check-autoscaler-max-instances",
+                    "realert_every": 60,  # The check runs once a minute, so this would realert every hour.
                     "tip": (
                         "The autoscaler wants to scale up to handle additional load"
                         " because your service is overloaded, but cannot scale any"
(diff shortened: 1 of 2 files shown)


paasta_tools/check_autoscaler_max_instances.py
```diff
@@ -136,6 +136,7 @@ async def check_max_instances(
                 {
                     "page": False,  # TODO: remove this line once this alert has been deployed for a little while.
                     "runbook": "y/check-autoscaler-max-instances",
+                    "realert_every": 60,  # The check runs once a minute, so this would realert every hour.
```
nemacysts, line 139: is this the same behavior that the other paasta-native alerts have? e.g., the oom/under-replication pages?

nemacysts (approved):

paasta_tools/check_autoscaler_max_instances.py
EvanKrall: Most things don't specify, which means they get exponential backoff, I believe. `realert_every` is expressed in terms of number of times the event is sent to Sensu, rather than a time range, so exponential backoff would start out slower for things we check less frequently. For this, we send a Slack message at 1m, 2m, 4m, 8m, 16m, 32m, 64m, 128m, etc., meaning 8 alerts in the first ~2 hours, which is pretty noisy.

EvanKrall: check_oom_events has realert_every=1 and runs every 5 minutes.
check_kubernetes_service_replication doesn't seem to specify (so exponential backoff) and runs every minute.

nemacysts: gotcha - seems fine to me then :)

PR #3772, opened 2024-01-11
Remove mesos code from get_running_task_allocation

nemacysts:
There's no more mesos, we can get rid of this code :p


nemacysts:
we'll want to merge sysgit/puppet#12169 (internal) first :)

nemacysts:
this is happening in https://github.com/Yelp/paasta/pull/4088

PR #3773, opened 2024-01-11
Add placeholder value for non-existent boto_keys

Qmando:
Previously, using an invalid value here would cause the service to fail to launch. Now that we're migrating away from using AWS users, people mess this up more often, and it may be desirable to let the pod launch with placeholder values.

Placeholder values instead just ignoring them and not placing keys was much simpler to implement, and only very marginally less desirable IMO. 

We still get a log line that we can alert on when this happens.


nemacysts (approved):

nemacysts:
ty!

nurdann (approved):

PR #3775, opened 2024-01-12
Stop using uwsgi sidecars and supporting custom stats port

jfongatyelp:
We now have a centralized uwsgi-exporter set up which is responsible for scraping all uwsgi-stats-enabled services. This unfortunately means we need to standardize on stats port 8889 and stop supporting custom ports, but more importantly means we no longer need to run the uwsgi sidecar for these services.

This PR:
1. stops deploying the uwsgi sidecar at all
2. removes uwsgi_stats_port from the schema and stops using it entirely
3. updates uwsgi autoscaling docs to remove wording about the sidecar + highlight we don't support custom stats port

This *will* cause a bounce for all uwsgi-autoscaled instances due to the `scrape_uwsgi_prometheus` label removal, and will stop deploying the uwsgi sidecar in all environments.  However this should be safe as we no longer configure any prometheus to scrape the sidecar's port.

Note that I've kept the labelling for adding a routable_ip as well as deploy_group, as we still need prometheus to be able to scrape the metrics from the pod + apparently currently still use the deploy_group label in our relabeling configs for the metrics pulled via the centralized uwsgi-exporter.

I am also testing the deb in infrastage against the example_happyhour test service using uwsgi autoscaling currently

### DIFF
--- a/docs/source/autoscaling.rst
+++ b/docs/source/autoscaling.rst
@@ -65,20 +65,13 @@ The currently available metrics providers are:
   Measures the CPU usage of your service's container.
 
 :uwsgi:
-  With the ``uwsgi`` metrics provider, Paasta will configure your pods to run an additional container with the `uwsgi_exporter <https://github.com/timonwong/uwsgi_exporter>`_ image.
-  This sidecar will listen on port 9117, and will request metrics from your uWSGI master via its `stats server <http://uwsgi-docs.readthedocs.io/en/latest/StatsServer.html>`_.
-  The uwsgi_exporter container needs to know what port your uWSGI master's stats server is on - you can configure this with the ``uwsgi_stats_port`` key in the ``autoscaling`` dictionary.
-  ``uwsgi_exporter`` will translate the uWSGI stats into Prometheus format, which Prometheus will scrape.
+  With the ``uwsgi`` metrics provider, Paasta will configure your pods to be scraped from your uWSGI master via its `stats server <http://uwsgi-docs.readthedocs.io/en/latest/StatsServer.html>`_.
+  We currently only support uwsgi stats on port 8889, and Prometheus will attempt to scrape that port.
 
   .. note::
 
-    If you have configured your service to use a non-default stats port (8889), you need to explicity set ``uwsgi_stats_port`` in your autoscaling config with the same value to ensure that metrics are being exported.
+    If you have configured your service to use a non-default stats port (8889), PaaSTA will not scale your service correctly!
 
-  Extra parameters:
-
-  :uwsgi_stats_port:
-    the port that your uWSGI master process will respond to with stats.
-    Defaults to 8889.
 
 :gunicorn:
   With the ``gunicorn`` metrics provider, Paasta will configure your pods to run an additional container with the `statsd_exporter <https://github.com/prometheus/statsd_exporter>`_ image.
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -229,9 +229,6 @@
                         "use_resource_metrics": {
                             "type": "boolean"
                         },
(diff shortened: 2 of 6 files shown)


docs/source/autoscaling.rst
```diff
@@ -65,20 +65,13 @@ The currently available metrics providers are:
   Measures the CPU usage of your service's container.
 
 :uwsgi:
-  With the ``uwsgi`` metrics provider, Paasta will configure your pods to run an additional container with the `uwsgi_exporter <https://github.com/timonwong/uwsgi_exporter>`_ image.
-  This sidecar will listen on port 9117, and will request metrics from your uWSGI master via its `stats server <http://uwsgi-docs.readthedocs.io/en/latest/StatsServer.html>`_.
-  The uwsgi_exporter container needs to know what port your uWSGI master's stats server is on - you can configure this with the ``uwsgi_stats_port`` key in the ``autoscaling`` dictionary.
-  ``uwsgi_exporter`` will translate the uWSGI stats into Prometheus format, which Prometheus will scrape.
+  With the ``uwsgi`` metrics provider, Paasta will configure your pods to be scraped from your uWSGI master via its `stats server <http://uwsgi-docs.readthedocs.io/en/latest/StatsServer.html>`_.
+  We currently only support uwsgi stats on port 8889, and Prometheus will attempt to scrape that port.
```
nemacysts, line 69: +++++

paasta_tools/cli/schemas/kubernetes_schema.json
```diff
@@ -229,9 +229,6 @@
                         "use_resource_metrics": {
                             "type": "boolean"
                         },
-                        "uwsgi_stats_port": {
-                            "type": "integer"
-                        },
```
nemacysts, line 234: farewell footgun :p

paasta_tools/kubernetes_tools.py
```diff
@@ -1187,7 +1142,7 @@ def should_run_uwsgi_exporter_sidecar(
                 if autoscaling_params.get(
                     "use_prometheus",
                     DEFAULT_USE_PROMETHEUS_UWSGI
-                    or system_paasta_config.default_should_run_uwsgi_exporter_sidecar(),
+                    or system_paasta_config.default_should_use_uwsgi_exporter(),
```
nemacysts, line 1145: i think we can technically refactor this method to
```
return (
    self.is_autoscaling_enabled()
    and self.get_autoscaling_params()["metrics_provider"] == "uwsgi"
)```

since there's no other way to scale uwsgi services outside of prometheus these days

but this is probably fine for now if we'd rather make a minimal change!

File: paasta_tools/kubernetes_tools.py [hunk 4]
```diff
@@ -2321,22 +2276,19 @@ def get_pod_template_spec(
...
             # this should probably eventually be made into a default label,
-            # but for now we're fine with it being behind this feature toggle.
+            # but for now we're fine with it being behind these feature toggles.
             # ideally, we'd also have the docker image here for ease-of-use
             # in Prometheus relabeling, but that information is over the
             # character limit for k8s labels (63 chars)
             labels["paasta.yelp.com/deploy_group"] = self.get_deploy_group()
 
-        elif self.should_setup_piscina_prometheus_scraping():
+        if self.should_setup_piscina_prometheus_scraping():
```
nemacysts, line 2291: just curious: how come we want to change this to an `if`? 

(i don't think this will hurt, but it shouldn't be possible atm to configure something for both uwsgi and piscina scaling)

nemacysts (approved):
i see red, i ship :p

paasta_tools/kubernetes_tools.py
jfongatyelp: oops this is just me being bad at git; I had a previous commit where i'd removed the uwsgi section entirely so had to make this an if, but put it back once I found we were still using the deploy_group label.

PR #3776, opened 2024-01-12
Stop using uwsgi sidecars and supporting custom stats port (Again)

jfongatyelp:
https://github.com/Yelp/paasta/pull/3775 take 2, this time I won't accidentally merge to master before I'm ready. 🤦 

We now have a centralized uwsgi-exporter set up which is responsible for scraping all uwsgi-stats-enabled services. This unfortunately means we need to standardize on stats port 8889 and stop supporting custom ports, but more importantly means we no longer need to run the uwsgi sidecar for these services.

This PR:

- stops deploying the uwsgi sidecar at all
- removes uwsgi_stats_port from the schema and stops using it entirely
- updates uwsgi autoscaling docs to remove wording about the sidecar + highlight we don't support custom stats port

This _will_ cause a bounce for all uwsgi-autoscaled instances due to the `scrape_uwsgi_prometheus` label removal, and will stop deploying the uwsgi sidecar in all environments. However this should be safe as we no longer configure any prometheus to scrape the sidecar's port.

Note that I've kept the labelling for adding a routable_ip as well as deploy_group, as we still need prometheus to be able to scrape the metrics from the pod + apparently currently still use the deploy_group label in our relabeling configs for the metrics pulled via the centralized uwsgi-exporter.

I tested the deb in infrastage against the example_happyhour test service using uwsgi autoscaling, and it was deployed without uwsgi sidecars but was still able to autoscale based on load.

### DIFF
--- a/docs/source/autoscaling.rst
+++ b/docs/source/autoscaling.rst
@@ -65,20 +65,13 @@ The currently available metrics providers are:
   Measures the CPU usage of your service's container.
 
 :uwsgi:
-  With the ``uwsgi`` metrics provider, Paasta will configure your pods to run an additional container with the `uwsgi_exporter <https://github.com/timonwong/uwsgi_exporter>`_ image.
-  This sidecar will listen on port 9117, and will request metrics from your uWSGI master via its `stats server <http://uwsgi-docs.readthedocs.io/en/latest/StatsServer.html>`_.
-  The uwsgi_exporter container needs to know what port your uWSGI master's stats server is on - you can configure this with the ``uwsgi_stats_port`` key in the ``autoscaling`` dictionary.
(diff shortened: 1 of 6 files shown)


nemacysts (approved):

jvperrin (approved):
Nice, thanks!

I also suspect we could remove the gunicorn sidecar, but we don't have a centralized statsd exporter yet, so that'll probably have to be at another time.

jfongatyelp:
> Nice, thanks!
> 
> I also suspect we could remove the gunicorn sidecar, but we don't have a centralized statsd exporter yet, so that'll probably have to be at another time.

Good call, I'd forgotten that was going to affect the same pods as those using uwsgi autoscaling, but yup I"ll leave that for future-us.

PR #3779, opened 2024-01-16
Base64 encode the boto_key missing error placeholder text

Qmando:
This placeholder text was not previously base64 encoded, which resulted in a 400 error from the Kube api when trying to add it as a secret. This should resolve that issue by doing the encoding first.


nurdann (approved):

nemacysts (approved):
gosh, i always forget this is a thing that needs to be done

bobokun (approved):

PR #3777, opened 2024-01-16
Allow specifying multiple CRDs in crd-{CLUSTER}.yaml files

nemacysts:
Open-source operators may need to have N CRDs created/managed by PaaSTA, which doesn't mesh with our current 1 CRD per file approach.

PR #3780, opened 2024-01-16
Allow skipping cleanup of tagged CRDs

nemacysts:
Open-source operators may need to have N CRDs created, which is not currently suppored by PaaSTA due to our current 1 CRD per file approach.

For teach cases, we tend to manage CRDs with Flux - but PaaSTA will clean these CRDs up since they're in a PaaSTA-managed namespace, but there's no record of them in soaconfigs.


nemacysts:
We'll want to merge either this or https://github.com/Yelp/paasta/pull/3777

ilkinmammadzada (approved):

jfongatyelp (approved):

PR #3786, opened 2024-01-29
allow local-run to use currently set pod identity

Qmando:
We've encountered a couple of requests to allow the use of local-run within a kubernetes pod, and to have the pod identity be passed through from parent to child container. This change causes paasta local-run to detect a web identity token, assume into the relevant role, and launch the container with that session.

Manual test here: https://fluffy.yelpcorp.com/i/fR178F7KH9Sf7Cs1FT7s2vF6507RX6SG.html

The existing behavior will remain when using any of the aws related command line options, but this will be used by default if the env variables are present. You could override this behavior by running something like `env -u AWS_WEB_IDENTITY_TOKEN_FILE paasta local-run ...`, so I didn't see the need to add another argument to do that.

This resolves the internal jira ticket SEC-18598

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -734,13 +734,13 @@ def assume_aws_role(
         )
         boto_session = boto3.Session()
         credentials = boto_session.get_credentials()
-        creds_dict: AWSSessionCreds = {
+        assumed_creds_dict: AWSSessionCreds = {
             "AWS_ACCESS_KEY_ID": credentials.access_key,
             "AWS_SECRET_ACCESS_KEY": credentials.secret_key,
             "AWS_SESSION_TOKEN": credentials.token,
             "AWS_SECURITY_TOKEN": credentials.token,
         }
-        return creds_dict
+        return assumed_creds_dict
     else:
         # use_okta_role, assume_pod_identity, and assume_role are all empty, and there's no
         # pod identity (web identity token) in the env. This shouldn't happen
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -727,10 +727,26 @@ def assume_aws_role(
...
+        print(
+            f"Found Pod Identity token in env. Assuming into role {os.environ['AWS_ROLE_ARN']}."
+        )
+        boto_session = boto3.Session()
+        credentials = boto_session.get_credentials()
+        creds_dict: AWSSessionCreds = {
+            "AWS_ACCESS_KEY_ID": credentials.access_key,
+            "AWS_SECRET_ACCESS_KEY": credentials.secret_key,
+            "AWS_SESSION_TOKEN": credentials.token,
+            "AWS_SECURITY_TOKEN": credentials.token,
```
smrojas, line 741: this is for boto2 compatibility? I thought web identity didn't work with boto2?

smrojas (approved):
changes look good to me.

danielpops (approved):

nemacysts (approved):
looks like the public itests are failing due to some shenanigans with cryptography - i guess we're not pinning in our itest requirements somewhere?

that said, it doesn't seem related to your PR

paasta_tools/cli/cmds/local_run.py
Qmando: Yeah, it's probably time to get rid of this. Especially because, as you point out, boto2 doesn't support pod identity, so you wouldn't need it to work while emulating pod identity...

PR #3790, opened 2024-02-02
Expose AWS account for role assumption

nurdann:
## Problem

* There are use cases when a `paasta local-run` runs with a prod role in dev environment [[thread](https://yelp.slack.com/archives/CA4N6NW93/p1706909725006629?thread_ts=1696617970.548759&cid=CA4N6NW93)]
* @danielpops suggested something simlilar https://github.com/Yelp/paasta/pull/3532#discussion_r1081774925 . That is mapping cluster (`-c`) to the account.

## Solution

Expose the source AWS account as CLI-flag that is used for role assumption.

## Verification 

- Worked with `--assume-pod-identity` [[fluffy](https://fluffy.yelpcorp.com/i/mBMC27Lcm9tZpD54Lwg46Xdqz1hShR5t.html)]
- Worked with `--assume-role-arn` [[fluffy](https://fluffy.yelpcorp.com/i/XDxmSWVqmT6PKZ5fPscWVsRG3TQCfFGR.html)]
- Without `-a prod` running on dev you get AccessDenied as expected [[fluffy](https://fluffy.yelpcorp.com/i/sMVMlM6cNSfpzlC7JbG0SqCK87L7MNvv.html)]
- Using `-a prod` but omitting cluster (i.e. using current) will result in error as well as expected [[fluffy](https://fluffy.yelpcorp.com/i/BgMkFspnFRHMD7tDLLsJwCWXqCz8zqcM.html)]

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -712,8 +712,7 @@ def resolve_aws_account_from_runtimeenv(aws_account: str = None) -> str:
         "stage": "dev",
         "corp": "corpprod",
     }
-    aws_account = runtimeenv_to_account_overrides.get(runtimeenv, runtimeenv)
-    return aws_account
+    return runtimeenv_to_account_overrides.get(runtimeenv, runtimeenv)
 
 
 def assume_aws_role(
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -473,6 +473,12 @@ def add_subparser(subparsers):
         default=False,
     )
     list_parser.add_argument(
+        "--assume-role-aws-account",
+        "--aws-account",
+        "-a",
+        help="Specify AWS account from which to source credentials",
+    )
+    list_parser.add_argument(
         "--assume-role-arn",
         help=(
             "role ARN to assume before launching the service. "
@@ -688,12 +694,35 @@ def check_if_port_free(port):
     return True
 
 
+def resolve_aws_account_from_runtimeenv(aws_account: str = None) -> str:
+    if aws_account is not None:
+        return aws_account
+
+    try:
+        with open("/nail/etc/runtimeenv") as runtimeenv_file:
+            runtimeenv = runtimeenv_file.read()
+    except FileNotFoundError:
+        print(
+            "Unable to determine environment for AWS account name. Using 'dev'",
+            file=sys.stderr,
+        )
+        runtimeenv = "dev"
+
+    runtimeenv_to_account_overrides = {
+        "stage": "dev",
+        "corp": "corpprod",
+    }
+    aws_account = runtimeenv_to_account_overrides.get(runtimeenv, runtimeenv)
+    return aws_account
+
+
 def assume_aws_role(
     instance_config: InstanceConfig,
     service: str,
     assume_role_arn: str,
     assume_pod_identity: bool,
     use_okta_role: bool,
+    aws_account: str = None,
 ) -> AWSSessionCreds:
     """Runs AWS cli to assume into the correct role, then extract and return the ENV variables from that session"""
     pod_identity = instance_config.get_iam_role()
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -688,12 +694,35 @@ def check_if_port_free(port):
     return True
 
 
+def resolve_aws_account_from_runtimeenv(aws_account: str = None) -> str:
+    if aws_account is not None:
+        return aws_account
```
nemacysts, line 699: the typing here implies that aws_account will always be non-null

paasta_tools/cli/cmds/local_run.py
```diff
@@ -705,20 +734,9 @@ def assume_aws_role(
...
-            elif aws_account == "corp":
-                aws_account = "corpprod"
-    except FileNotFoundError:
-        print(
-            "Unable to determine environment for AWS account name. Using 'dev'",
-            file=sys.stderr,
-        )
-        aws_account = "dev"
+
+    aws_account = resolve_aws_account_from_runtimeenv(aws_account)
```
nemacysts, line 737: should this be something like:
```suggestion
    aws_account = aws_account or resolve_aws_account_from_runtimeenv(aws_account)
```
then we can remove the check for `if aws_account is not None` from `resolve_aws_account_from_runtimeenv()`

paasta_tools/cli/cmds/local_run.py
```diff
@@ -688,12 +694,35 @@ def check_if_port_free(port):
...
+    return aws_account
+
+
 def assume_aws_role(
     instance_config: InstanceConfig,
     service: str,
     assume_role_arn: str,
     assume_pod_identity: bool,
     use_okta_role: bool,
+    aws_account: str = None,
```
nemacysts, line 724:
```suggestion
    aws_account: Optional[str] = None,
```

danielpops (approved):
Can you make sure to add / update documentation for this new behavior? Can you also create a JIRA ticket with context for why this change is being made?

paasta_tools/cli/cmds/local_run.py
```diff
@@ -688,12 +694,34 @@ def check_if_port_free(port):
     return True
 
 
+def resolve_aws_account_from_runtimeenv(aws_account: str = None) -> str:
+    if aws_account is not None:
+        return aws_account
+
+    try:
```
jfongatyelp, line 701: If we want to do something closer to what @danielpops suggested in the original PR to also assume some right defaults if they also specify a `cluster`, you could use https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L2842-L2843 e.g. `system_paasta_config.get_kube_clusters().get(cluster, {}).get("aws_account_name", local_runtimeenv)` to grab the aws account associated with a particular cluster (stored in our /etc/paasta/kube_clusters.json file), otherwise fall back to the runtimeenv.

nurdann: For maximum compatibility, the AWS account is determined 
1. First by argument
2. If argument is omitted, then get from kube cluster mapping
3. Else get from runtime environment

nurdann: So new behaviour uses the AWS account of the kube cluster, unless it's overridden.

nurdann:
## Updated verification
- Assume from dev to prod cluster; needs `-a` and `-c` - fails as expected [[fluffy](https://fluffy.yelpcorp.com/i/0bMSBLgsqwxtk6QtGvHbqnxdZF8p9fzB.html)]
- Assume from prod to prod cluster; needs `-c` - succeeds [[fluffy](https://fluffy.yelpcorp.com/i/KBrL51DTz6n0cd2xPK5gl4VqTG3brf1V.html)]
- Omitted uses default configs [[fluffy](https://fluffy.yelpcorp.com/i/9V7F48sCW36MlKd6ZMBPgqdHzmSHFNLV.html)]

paasta_tools/cli/cmds/local_run.py
```diff
@@ -842,7 +856,7 @@ def run_docker_container(
     else:
         chosen_port = pick_random_port(service)
     environment = instance_config.get_env_dictionary()
-    secret_volumes = {}
+    secret_volumes = {}  # type: ignore
```
nurdann, line 859: Ignoring type here because it would requiring either changing how we write to file (as `str` first and  then as `bytes`) or changing signature of https://github.com/Yelp/paasta/blob/8133eaa2c062c918c2efd7ec32a78d0f5f796a2f/paasta_tools/secret_tools.py#L196

nemacysts (approved):
I'm not quite sure what suddenly changed to make mypy so unhappy in this file re: `secret_volumes` and the context manager variables - but outside of that this seems reasonable :)

PR #3797, opened 2024-02-07
FLINK-4925: Create flink eks instance type

AhmedSoliman7:
- Reusing the same config Dict `FlinkDeploymentConfigDict` for `FlinkEksDeploymentConfig`
- We will be using `flinkeks` file prefix for this instance type
- The new handlers are added to the mappings of prefix -> handlers

Context:

In order to migrate flink to eks and use paasta commands, we are doing some workaround and using a different file prefix for the crd `flinkeks` to create a special deployment config `FlinkEksDeploymentConfig` . This instance type will be used to educate paasta commands which paasta api to use (eks or not) 

This approach is temporary and we should revert back to the original file prefix after the migration is done.  

The assumption here is that if we have an additional (parallel) flink custom resource ([here](https://sourcegraph.yelpcorp.com/sysgit/puppet/-/blob/hieradata/common.yaml?L1608:30)) but with different file prefix (flinkeks-), `setup_kubernetes_cr.py` will pick those CRs and create the instances with the new `FlinkEksDeploymentConfig` type 

_There should be more follow up work around re-directing the paasta commands to the correct paasta api based on the flink instance type_

### DIFF
--- a/paasta_tools/cli/utils.py
+++ b/paasta_tools/cli/utils.py
@@ -44,8 +44,8 @@ from paasta_tools.adhoc_tools import load_adhoc_job_config
 from paasta_tools.cassandracluster_tools import load_cassandracluster_instance_config
 from paasta_tools.eks_tools import EksDeploymentConfig
 from paasta_tools.eks_tools import load_eks_service_config
-from paasta_tools.flink_eks_tools import load_flink_eks_instance_config
 from paasta_tools.flink_tools import load_flink_instance_config
+from paasta_tools.flinkeks_tools import load_flinkeks_instance_config
 from paasta_tools.kafkacluster_tools import load_kafkacluster_instance_config
 from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
 from paasta_tools.kubernetes_tools import load_kubernetes_service_config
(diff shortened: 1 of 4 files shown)


tests/test_flink_eks_tools.py
```diff
@@ -0,0 +1,79 @@
...
+from unittest.mock import MagicMock
+from unittest.mock import patch
+
+from paasta_tools.flink_eks_tools import DEFAULT_SOA_DIR
+from paasta_tools.flink_eks_tools import FlinkEksDeploymentConfig
+from paasta_tools.flink_eks_tools import load_flink_eks_instance_config
+from paasta_tools.utils import BranchDictV2
+
+
+class TestFlinkEksDeploymentConfig(unittest.TestCase):
```
nemacysts, line 11: it'd be preferable to use pytest instead of unittest - we don't really use the built-in unittest library except in a vanishingly small number of ancient tests :)

nemacysts: (should be pretty simple to convert this though!)

nemacysts (commented):
lgtm other than the `unittest` usage in the new test file :)

tests/test_flink_eks_tools.py
ajayOO8: updated

paasta_tools/flink_eks_tools.py
nemacysts, line 1: to be consistent, we probably want to name this `flinkeks_tools.py` as the filenames are generally named after the instance type

tests/test_flink_eks_tools.py
nemacysts, line 1: (same as above - if we rename flink_eks_tools to flinkeks_tools, this filename will also need to be updated)

nemacysts (commented):
one minor nit re: filenames that i didn't catch in the original review - but otherwise looks good!

paasta_tools/flink_eks_tools.py
ajayOO8: thanks. updated

tests/test_flink_eks_tools.py
ajayOO8: ty. renamed

jfongatyelp (approved):

nemacysts (approved):

PR #3798, opened 2024-02-07
MLCOMPUTE-872 | Support driver on k8s config format for paasta spark-run

edingroot:
## Description
To provide a way for users to test their Spark tronfig (driver on k8s) with the ad-hoc `paasta spark-run`.

This implementation just simply loads the yaml file and parse selected fields instead to using `tron_tools` to avoid tron-specific logic being included to the existing `paasta spark-run` and let user able to test simplified configs.

## Test

With tronfig`~/tmp/tron-spark-pnw-devc.yaml`
```
_role_template: &role_template
  iam_role: arn:aws:iam::<account_id>:role/spark_test_role
  iam_role_provider: aws

spark_driver_test_paasta_secrets:
  # every 30 mins for testing
  schedule: "cron */30 * * * *"
  deploy_group: dev.pnw
  enabled: false
  actions:
    bash:
      <<: *role_template
      executor: spark
      env:
        ENV1: env1_value
        SAMPLE_ENV: SECRET(classified_sample)
      pool: batch
      spark_args:
        spark.kubernetes.executor.secretKeyRef.SAMPLE_ENV: paasta-spark-secret-spark-classified--sample:classified_sample
      command: "
        /bin/bash
        "
```

In devbox:  
```
./.tox/py38-linux/bin/paasta spark-run --aws-profile=dev --tronfig ~/tmp/tron-spark-pnw-devc.yaml --job-id spark_driver_test_paasta_secrets.bash
```
image

In batch server:  
```
sudo ./.tox/py38-linux/bin/paasta spark-run --aws-role-duration=900 --tronfig ~/tmp/tron-spark-pnw-devc.yaml --job-id spark_driver_test_paasta_secrets.bash
```
image

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -1199,7 +1199,7 @@ def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
     )
 
 
-def parse_tronfig(tronfig_path: str, tronfig_target: str) -> Optional[Dict[str, str]]:
+def parse_tronfig(tronfig_path: str, tronfig_target: str) -> Optional[Dict[str, Any]]:
     splitted = tronfig_target.split(".")
     if len(splitted) != 2:
         return None
@@ -1289,7 +1289,7 @@ def update_args_from_tronfig(args: argparse.Namespace) -> Optional[Dict[str, str
 
 
 def paasta_spark_run(args: argparse.Namespace) -> int:
-    driver_envs_from_tronfig = dict()
+    driver_envs_from_tronfig: Dict[str, str] = dict()
     if args.tronfig is not None:
         if args.tronfig_target is None:
             print(
@@ -1297,9 +1297,10 @@ def paasta_spark_run(args: argparse.Namespace) -> int:
                 file=sys.stderr,
             )
             return False
-        driver_envs_from_tronfig = update_args_from_tronfig(args)
-        if driver_envs_from_tronfig is None:
+        result = update_args_from_tronfig(args)
+        if result is None:
             return False
+        driver_envs_from_tronfig = result
 
     # argparse does not work as expected with both default and
     # type=validate_work_dir.
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -885,7 +885,7 @@ def configure_and_run_docker_container(
     aws_creds: Tuple[Optional[str], Optional[str], Optional[str]],
     cluster_manager: str,
     pod_template_path: str,
-    extra_driver_envs: Dict[str, str],
+    extra_driver_envs: Dict[str, str] = dict(),
 ) -> int:
     docker_memory_limit = _calculate_docker_memory_limit(
         spark_conf, args.docker_memory_limit
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -340,14 +340,14 @@ def add_subparser(subparsers):
 
     list_parser.add_argument(
         "--tronfig",
-        help="Load the Tron config yaml. Use with --tronfig-target.",
+        help="Load the Tron config yaml. Use with --job-id.",
         type=str,
         default=None,
     )
 
     list_parser.add_argument(
-        "--tronfig-target",
-        help="Target <job_name>.<action_name> in the Tronfig to run. Use wuth --tronfig.",
+        "--job-id",
+        help="Tron job id <job_name>.<action_name> in the Tronfig to run. Use wuth --tronfig.",
         type=str,
         default=None,
     )
@@ -1199,8 +1199,8 @@ def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
     )
 
 
-def parse_tronfig(tronfig_path: str, tronfig_target: str) -> Optional[Dict[str, Any]]:
-    splitted = tronfig_target.split(".")
+def parse_tronfig(tronfig_path: str, job_id: str) -> Optional[Dict[str, Any]]:
+    splitted = job_id.split(".")
     if len(splitted) != 2:
         return None
     job_name, action_name = splitted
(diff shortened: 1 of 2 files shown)


88manpreet:
Can we make the argument format similar to `paasta local-run` for consistency purposes. I also think it would be reduce the steps for users as well.
The idea is to use `-y, --yelpsoa-configs-dir-path` option and the yaml file is being picked-up automatically with --service option.

88manpreet:
Suggestion: Change --tronfig-target  to --job-id? 
A more commonly used terminology for spark jobs?

edingroot:
> Can we make the argument format similar to `paasta local-run` for consistency purposes. I also think it would be reduce the steps for users as well. The idea is to use `-y, --yelpsoa-configs-dir-path` option and the yaml file is being picked-up automatically with --service option.

Currently, the configs for spark-run it's a bit tricky - it loads (1) PaaSTA-related configs from `adhoc-<cluster>.yaml` such as volume mounts, environment variables when starting the spark driver container, however the (2) spark-related configs are loaded from the `paasta spark-run` CLI arguments from the `tron-<cluster>.yaml` or the user input in devbox.

For (1) when already have `--yelpsoa-config-root`/`-y` argument for loading ad-hoc soa configs locally; for (2) the argument `--tronfig` added in this PR is to provide another option for users to testing using the tronfig instead of the current CLI arguments.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1181,7 +1199,107 @@ def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
...
+    # executor === spark
+    if action_dict.get("executor", "") != "spark":
+        print(
+            PaastaColors.red("Invalid Tronfig: executor should be 'spark'"),
+            file=sys.stderr,
+        )
+        return None
+
+    # iam_role / aws_profile
+    if "iam_role" in action_dict and action_dict.get("iam_role_provider", "") != "aws":
```
88manpreet, line 1245: TIL. Can iam_role_provider could be something else too at Yelp?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1181,7 +1199,107 @@ def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
...
+def paasta_spark_run(args: argparse.Namespace) -> int:
+    driver_envs_from_tronfig: Dict[str, str] = dict()
+    if args.tronfig is not None:
+        if args.job_id is None:
+            print(
+                PaastaColors.red("Missing --job-id when --tronfig is provided"),
+                file=sys.stderr,
+            )
+            return False
+        result = update_args_from_tronfig(args)
```
88manpreet, line 1298:
```suggestion
        driver_envs_from_ tronfig  = update_args_from_tronfig(args)
```

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1181,7 +1199,107 @@ def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
...
+        if result is None:
```
88manpreet, line 1299:
```suggestion
        if driver_envs_from_ is None:
```

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1181,7 +1199,107 @@ def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
...
+            return False
+        driver_envs_from_tronfig = result
```
88manpreet, line 1301:
```suggestion
```

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1181,7 +1199,107 @@ def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
...
+        return None
+
+    # Other args
+    fields_to_args = {
+        "pool": "pool",
+        "iam_role": "assume_aws_role",
+        "command": "cmd",
+        "spark_args": "spark_args",
+    }
+    for field_name, arg_name in fields_to_args.items():
```
88manpreet, line 1259: Can you add comment on why each one of the fields are being needed for special cases?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1181,7 +1199,107 @@ def _get_k8s_url_for_cluster(cluster: str) -> Optional[str]:
...
+        if field_name in action_dict:
+            value = action_dict[field_name]
+            if field_name == "spark_args":
+                value = " ".join([f"{k}={v}" for k, v in dict(value).items()])
+
+            arg_name_str = (f"--{arg_name.replace('_', '-')}").ljust(20, " ")
```
88manpreet, line 1265: Why do we need this?

88manpreet:
Overall the code looks good. One question:
It seems like this logic is implemented for only spark-on-k8s.
Why do we need to add logic to handle spark-args?
It seems like in driver-on-k8s, we don't expect --spark-args as one of the fields either in adhoc or batch mode.

paasta_tools/cli/cmds/spark_run.py
edingroot: No, there were other values available before but they are deprecated

paasta_tools/cli/cmds/spark_run.py
edingroot: To align padding in the printed message

edingroot:
> Overall the code looks good. One question: It seems like this logic is implemented for only spark-on-k8s. Why do we need to add logic to handle spark-args? It seems like in driver-on-k8s, we don't expect --spark-args as one of the fields either in adhoc or batch mode.

In driver-on-k8s, `--spark-args` are assigned by `job_name.action_name.spark_args` dict in the yaml file. The logic in this PR is to convert the values in yaml file to the CLI format for `paasta spark-run` to read. This is a temporary measure for users to test and should be removed after we have migrated `paasta spark-run` to driver-on-k8s, which could be a new command or a totally new implementation.

88manpreet (approved):

PR #3800, opened 2024-02-12
PAASTA-18005: Add default timeout to paasta-api

bobokun:
We've noticed that sometimes the paasta API can hang/get stuck and clients end up hanging indefinitely, this fix adds a default timeout to avoid this.

### DIFF
--- a/paasta_tools/api/client.py
+++ b/paasta_tools/api/client.py
@@ -60,6 +60,10 @@ def get_paasta_oapi_client_by_url(
     config.ssl_ca_cert = ssl_ca_cert
 
     client = paastaapi.ApiClient(configuration=config)
+    # PAASTA-18005: Adds default timeout to paastaapi client
+    client.rest_client.pool_manager.connection_pool_kw[
+        "timeout"
+    ] = load_system_paasta_config().get_api_client_timeout()
     return PaastaOApiClient(
         autoscaler=paastaapis.AutoscalerApi(client),
         default=paastaapis.DefaultApi(client),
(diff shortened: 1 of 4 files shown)


paasta_tools/utils.py
```diff
@@ -2246,6 +2247,13 @@ def get_auto_config_instance_type_aliases(self) -> Dict[str, str]:
         """
         return self.config_dict.get("auto_config_instance_type_aliases", {})
 
+    def get_api_client_timeout(self) -> int:
+        """
+        We've seen the Paasta API get hung up sometimes and the client not realizing this will sit idle forever.
+        This will be used to specify the default timeout
+        """
+        return self.config_dict.get("api_client_timeout", 120)
```
nemacysts, line 2255: it's worth double-checking that we don't have any timeouts that are under <120 in paasta status

nemacysts (approved):

paasta_tools/api/client.py
```diff
@@ -60,6 +60,10 @@ def get_paasta_oapi_client_by_url(
     config.ssl_ca_cert = ssl_ca_cert
 
     client = paastaapi.ApiClient(configuration=config)
+    # PAASTA-18005: Adds default timeout to paastaapi client
+    client.rest_client.pool_manager.connection_pool_kw[
+        "timeout"
+    ] = load_system_paasta_config().get_api_client_timeout()
```
nemacysts, line 66: we'll need to add a mock in the tests for this, fyi (that's why the tests are currently failing)

PR #3801, opened 2024-02-13
Fixes local tests and openapi-codegen

bobokun:
This fixes a change made with the unprivileged containers switch which causes openapi-codegen to fail. The full url is needed now to pull the image from the docker registry.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -115,8 +115,8 @@ k8s_clean: .paasta/bin/activate
 #   in paasta repo: java -jar ~/openapi-generator/modules/openapi-generator-cli/target/openapi-generator-cli.jar
 openapi-codegen:
 	rm -rf paasta_tools/paastaapi
-	docker run --rm -i --user `id -u`:`id -g` -v `pwd`:/src -w /src \
-		yelp/openapi-generator-cli:20201026 \
+	docker run --rm -i -v `pwd`:/src -w /src \
+		registry.hub.docker.com/yelp/openapi-generator-cli:20201026 \
 		generate \
 		-i paasta_tools/api/api_docs/oapi.yaml \
 		-g python-experimental \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -28,6 +28,9 @@ else
 	export INDEX_URL_BUILD_ARG ?= PIP_INDEX_URL
 endif
 
+# Use this variable for the docker run command
+DOCKER_REGISTRY_ARG=$(if [ "$(DOCKER_REGISTRY)" != "" ]; then echo "$(DOCKER_REGISTRY)"; fi)
```
jfongatyelp, line 32: Why not just reuse `$DOCKER_REGISTRY` in this case?

bobokun: I couldn't use it since the original `export DOCKER_REGISTRY ?= ""` would lead to `""yelp/openapi-generator-cli:20201026` and cause the github actions to fail

jfongatyelp: ! I think that indicates we've been setting `DOCKER_REGISTRY` incorrectly if so; i think the intent is not to have the explicit string `""`, but an empty string.

Makefile
```diff
@@ -24,7 +24,7 @@ endif
 ifeq ($(PAASTA_ENV),YELP)
 	export DOCKER_REGISTRY ?= docker-dev.yelpcorp.com/
 else
-	export DOCKER_REGISTRY ?= ""
+	export DOCKER_REGISTRY ?=
```
jfongatyelp, line 27: This could well have some unintentional consequences... imo you can probably ship this if tests pass and if it works on normal public builds (unless @nemacysts already recognizes a problem here) 👍 

If not, we may have to look a bit closer, but a quick fix would be to set it back to "" and just modify the codegen docker cmds as we have done elsewhere: https://github.com/Yelp/paasta/blob/master/yelp_package/Makefile#L38

Makefile
jfongatyelp: If we have to do this two-docker-registry arg thing, technically I would swap these variable names -- the one we're calling `DOCKER_REGISTRY` is actually being passed around as our itests/yelp_package dockerfiles' `DOCKER_REGISTRY` `ARG`s :D

Makefile
nemacysts: i *think* this would work? we could also probably sidestep this "" vs empty string issue by setting the `else` case to the full dockerhub url like the original commit did :)

nemacysts (approved):
might be nice to set `export DOCKER_REGISTRY ?= ""` to the full url (i.e,. include `registry.hub.docker.com/` here) just to be explicit, but this is also fine :)

nemacysts (approved):

PR #3802, opened 2024-02-13
Scale with ready pods count rather current replicas for uwsgi

nemacysts:
We had some surprising behavior recently where an autoscaled app was crashlooping and the HPA did not keep the desired scale as-is - we believe that this is due to a mismatch in how we calculate the value we send to k8s:

The existing PromQL divides by current replica to turn things into a utilization percentage such that when the HPA multiplies later on by the number of current replicas, we reach our target replicas.

However, the HPA will actually multiply by the number of ready replicas, leading to a mismatch.

### DIFF
--- a/paasta_tools/setup_prometheus_adapter_config.py
+++ b/paasta_tools/setup_prometheus_adapter_config.py
@@ -397,20 +397,6 @@ def create_instance_uwsgi_scaling_rule(
     worker_filter_terms = f"paasta_cluster='{paasta_cluster}',paasta_service='{service}',paasta_instance='{instance}'"
     replica_filter_terms = f"paasta_cluster='{paasta_cluster}',deployment='{deployment_name}',namespace='{namespace}'"
 
-    current_replicas = f"""
-        sum(
-            label_join(
-                (
-                    kube_deployment_spec_replicas{{{replica_filter_terms}}} >= 0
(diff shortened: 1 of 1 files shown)


paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -453,8 +450,14 @@ def create_instance_uwsgi_scaling_rule(
             )[{moving_average_window}s:]
         )
     """
+
+    # our Prometheus query is calculating a desired number of replicas, and then k8s wants that expressed as an average utilization
+    # so as long as we divide by the number that k8s ends up multiplying by, we should be able to convince k8s to run any arbitrary
+    # number of replicas.
+    # k8s happens to multiply by the # of ready pods - so we divide by that rather than by the amount of current replicas (which may
+    # include non-ready pods)
```
EvanKrall, line 458: might want to link to the k8s source line in question https://github.com/kubernetes/kubernetes/blob/7ec1a89a509906dad9fd6a4635d7bfc157b47790/pkg/controller/podautoscaler/replica_calculator.go#L278

nemacysts: done :)

EvanKrall (approved):

PR #3803, opened 2024-02-14
Speed up secret syncing

nemacysts:
It turns out that the slowness here was due to our hardcoded .3s sleep - we're
now syncing a lot more secrets (and generally doing more work) because of our
namespace sharding, so this sleep ends up taking quite a bit of time.

We generally haven't had to throttle much (outside of Pod creation, but that's
generally due to other constraints: e.g., cluster autoscaling and whatnot), so
it's probably fine to remove this sleep entirely for the time being.

That said, in the interest of safety, I've made this delay configurable - so if
this ends up causing issues after release (or in the far future), we can easily
tweak the delay.

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -14,7 +14,6 @@
 # limitations under the License.
 import argparse
 import base64
-import concurrent.futures
 import contextlib
 import hashlib
 import json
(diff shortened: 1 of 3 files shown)


nemacysts:
looks like the noop case is still unchanged, i'll profile a bit and see what happens

EmanElsaban (approved):

jfongatyelp (approved):

PR #3804, opened 2024-02-16
spark-run: properly set user under unpriv containers

piax93:
Solves the usual issue with file permissions when interacting with mounted user directories.

May create new issues for whatever gets run inside the container which I still need to verify.


oholiab (approved):

88manpreet:
Change lgtm. I would suggest to run some manual test cases (adhoc spark runs) before pushing this.
Seems like it's not blocking. Would suggest to push on Monday.

88manpreet (approved):

danielpops (approved):

PR #3806, opened 2024-02-20
Make some updates to our README

mattmb:
Mostly just correcting a few things and updating things that aren't true. I also tried to make it clearer that PaaSTA isn't intended for use outside Yelp.


ilkinmammadzada (approved):

gmdfalk (approved):
~~mesos~~

nemacysts (approved):

PR #3811, opened 2024-02-28
overriding fileprefix for eks clusters

ajayOO8:
currently EKS cluster will try to deploy both of the CRs(old cluster and EKS)
for eg {operator}-{cluster}.yaml and {operator}{eks}-cluster.yaml
overidding the fileprefix for eks clusters


nemacysts:
@ajayOO8 is this necessary? my understanding is that we've already migrated some flink workloads to eks :)

ajayOO8:
> is this necessary? my understanding is that we've already migrated some flink workloads to eks :)
This is not required now, we have update made this change in paasta system configs for CRDs

PR #3812, opened 2024-02-28
Fix PAASTA_PORT in local-run containers

jfongatyelp:
Looks like we've been setting `PAASTA_PORT` to the port used to bind docker to the container port.  This conflicts with what we specify in the [paasta contract](https://paasta.readthedocs.io/en/latest/about/contract.html#http-tcp-services) which says service owners should listen on `PAASTA_PORT` inside the container.

This switches to let the instance config's `get_env` set `PAASTA_PORT` which should match our expectations, especially as we are using container port in the docker `--publish` flag: https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/local_run.py#L560-L561 +
https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/local_run.py#L944

The switch to defer this work to the instance config's `get_env` also meant that for Kubernetes instances, we lose the `PAASTA_CLUSTER` envvar, so this also updates `get_local_run_environment_vars` to include it.

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -661,8 +661,11 @@ def get_local_run_environment_vars(instance_config, port0, framework):
         "PAASTA_DOCKER_IMAGE": docker_image,
         "PAASTA_LAUNCHED_BY": get_possible_launched_by_user_variable_from_env(),
         "PAASTA_HOST": hostname,
-        "PAASTA_PORT": str(instance_config.get_container_port()),
     }
+
+    if hasattr(instance_config, "get_container_port"):
+        env["PAASTA_PORT"] = str(instance_config.get_container_port())
+
     if framework == "marathon":
         fake_taskid = uuid.uuid4()
         env["MESOS_SANDBOX"] = "/mnt/mesos/sandbox"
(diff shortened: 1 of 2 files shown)


EvanKrall (commented):
Huh. It looks like PAASTA_PORT is set by `get_env` in LongRunningServiceConfig, but then local-run uses `get_env_dictionary` instead of `get_env`. I'm not sure why we have two methods that do the same thing (`InstanceConfig.get_env` just does `return self.get_env_dictionary(system_paasta_config=system_paasta_config)`).

Refactoring to make local-run use get_env might help prevent other differences between local-run and actual paasta clusters.

jfongatyelp:
> Huh. It looks like PAASTA_PORT is set by `get_env` in LongRunningServiceConfig, but then local-run uses `get_env_dictionary` instead of `get_env`. I'm not sure why we have two methods that do the same thing (`InstanceConfig.get_env` just does `return self.get_env_dictionary(system_paasta_config=system_paasta_config)`).
> 
> Refactoring to make local-run use get_env might help prevent other differences between local-run and actual paasta clusters.

This makes a ton of sense, thanks!
We had a little weirdness since KuberntesDeploymentConfig.get_env() explicitly `pop`s PAASTA_CLUSTER, but it seems like it should be no harm to just include that in `get_local_run_environment` instead. LMK if you think otherwise.

EvanKrall (approved):

nemacysts (approved):

PR #3814, opened 2024-03-07
Support building Debian package with unprivileged docker

edingroot:
## Description

We have rolled out unprivileged docker to dev boxes, which causes the following problems when building Debian packages:
1. Does not support resolving short image names
  => Error when running `docker pull "docker.io/yelp/paastatools_bionic_container"`.
2. Has a virtual in-container uid to system uid mapping.
  => The whole repository directory will be changed to a incorrect UID/GID, causing the build abort.

This PR fixes both issues during the build process.

## Test
`make -C yelp_package PAASTA_ENV=YELP package_bionic` in devbox.

### DIFF
--- a/yelp_package/Makefile
+++ b/yelp_package/Makefile
@@ -73,6 +73,11 @@ package_%: build_$$*_docker go_entrypoint
 
 	# noddebs to work around https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=897569
 	$(DOCKER_RUN) /bin/bash -c "DEB_BUILD_OPTIONS=noddebs dpkg-buildpackage -d && mv ../*.deb dist/"
+	# The in-house Podman deployment (unprivileged Docker) maps a virtual UID in the OS to root in the container,
+	# so we skip this chown statement to avoid broken permissions.
+ifeq ($(findstring podman,$(DOCKER_HOST)),)
+	$(DOCKER_RUN) chown -R $(UID):$(GID) /work
+endif
 	# then move them back
 	mv ../requirements.txt.old ../requirements.txt
 	mv ../debian/changelog.old ../debian/changelog
--- a/yelp_package/Makefile
+++ b/yelp_package/Makefile
@@ -19,7 +19,7 @@ SHELL=/bin/bash
 
 UID:=`id -u`
 GID:=`id -g`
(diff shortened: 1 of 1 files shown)


yelp_package/Makefile
```diff
@@ -73,7 +73,6 @@ package_%: build_$$*_docker go_entrypoint
 
 	# noddebs to work around https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=897569
 	$(DOCKER_RUN) /bin/bash -c "DEB_BUILD_OPTIONS=noddebs dpkg-buildpackage -d && mv ../*.deb dist/"
-	$(DOCKER_RUN) chown -R $(UID):$(GID) /work
```
piax93, line 76: This will still be needed in case this is built with actual docker, so I'd just suggest wrapping it in
```
ifeq ($(findstring podman,$(DOCKER_HOST)),)
...
endif
```

yelp_package/Makefile
```diff
@@ -73,7 +73,11 @@ package_%: build_$$*_docker go_entrypoint
...
+	# The in-house Podman deployment (unprivileged Docker) maps a virtual UID in the OS to root in the container,
+    # so we skip this chown statement to avoid broken permissions.
```
piax93, line 77: nit
```suggestion
	# The in-house Podman deployment (unprivileged Docker) maps a virtual UID in the OS to root in the container,
	# so we skip this chown statement to avoid broken permissions.
```

piax93 (approved):

yelp_package/Makefile
```diff
@@ -73,7 +73,11 @@ package_%: build_$$*_docker go_entrypoint
...
+	ifeq ($(findstring podman,$(DOCKER_HOST)),)
+		$(DOCKER_RUN) chown -R $(UID):$(GID) /work
+	endif
```
piax93, line 80: actually, looks like you'll want this not indented, otherwise `make` doesn't parse the if statement correctly.

```suggestion
ifeq ($(findstring podman,$(DOCKER_HOST)),)
	$(DOCKER_RUN) chown -R $(UID):$(GID) /work
endif
```

piax93 (changes requested):

yelp_package/Makefile
edingroot: Updated, thanks!

piax93 (approved):

PR #3816, opened 2024-03-12
stop monkey-patching the HPA object

drmorr0:
PaaSTA is on a modern-enough version of the Python kubernetes client that we don't need to monkey-patch the HPA dict anymore.

# Testing done
- make test

cc @jfongatyelp


jfongatyelp (approved):

nemacysts (approved):

drmorr0:
@jfongatyelp tested this on infrastage and confirmed the HPA still works, so I will merge now.

PR #3820, opened 2024-03-19
removed default eks var in kubernetes_cr.py

ajayOO8:
This will only add "paasta.yelp.com/eks" label when true, otherwise operators fails to reconcile existing stateful sets and other immutable k8s objects

### DIFF
--- a/paasta_tools/setup_kubernetes_cr.py
+++ b/paasta_tools/setup_kubernetes_cr.py
@@ -298,12 +298,13 @@ def format_custom_resource(
                 paasta_prefixed("service"): service,
                 paasta_prefixed("instance"): instance,
                 paasta_prefixed("cluster"): cluster,
-                "paasta.yelp.com/eks": str(is_eks),
             },
             "annotations": {},
         },
         "spec": instance_config,
     }
+    if is_eks:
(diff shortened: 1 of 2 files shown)


jfongatyelp (approved):

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -298,12 +298,13 @@ def format_custom_resource(
...
                 paasta_prefixed("instance"): instance,
                 paasta_prefixed("cluster"): cluster,
-                "paasta.yelp.com/eks": str(is_eks),
             },
             "annotations": {},
         },
         "spec": instance_config,
     }
+    if is_eks:
+        resource["metadata"]["labels"]["paasta.yelp.com/eks"] = str(is_eks)
```
nemacysts, line 307: should this be
```suggestion
        resource["metadata"]["labels"][paasta_prefixed("eks")] = str(is_eks)
```
to be consistent with the other paasta.yelp.com labels?

nemacysts (approved):

paasta_tools/setup_kubernetes_cr.py
ajayOO8: Thanks. updated the PR

PR #3821, opened 2024-03-19
migrate config from kubernetes to eks

ernestosilv:
Updated scripts to dynamically load `eks-*` for both `service_shard_update` and `service_shard_remove` if it exists in preparation for migration. Prior to the migration there should be no changes to the status quo since we'll still fetch `kubernetes-*` configs.

### DIFF
--- a/paasta_tools/contrib/service_shard_update.py
+++ b/paasta_tools/contrib/service_shard_update.py
@@ -130,9 +130,9 @@ def get_default_git_remote():
 
 
 DEPLOY_MAPPINGS = {
-    "dev": ["kubernetes-norcal-devc"],
-    "stage": ["kubernetes-norcal-stagef", "kubernetes-norcal-stageg"],
-    "prod": ["kubernetes-nova-prod", "kubernetes-pnw-prod"],
+    "dev": ["norcal-devc"],
+    "stage": ["norcal-stagef", "norcal-stageg"],
+    "prod": ["nova-prod", "pnw-prod"],
 }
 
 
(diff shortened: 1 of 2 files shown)


bobokun (approved):

paasta_tools/contrib/service_shard_update.py
```diff
@@ -172,7 +172,26 @@ def main(args):
 
             for deploy_prefix, config_paths in DEPLOY_MAPPINGS.items():
                 for config_path in config_paths:
-                    kube_file = updater.get_existing_configs(args.service, config_path)
+                    # Determine configuration suffix (PAASTA-18216)
+                    eks_config = updater.get_existing_configs(
```
nemacysts, line 176: i think my comment above re: both files being able to exist at a given time is maybe more pertinent here - but maybe we can simply temporarily disable the process that creates/updates shards as we do the big-bang migration so that we can always assume that for a given cluster only one of these files will exist?

nemacysts (approved):
seems fine if we're ok with some potential manual work at migration-time :)

panbenson (approved):

paasta_tools/contrib/service_shard_update.py
ernestosilv: Just curious, Do we have a timeline for the migration? We can probably go ahead and merged this changes and coordinate when we'd like to disable the process to create/modify those shards. Also do we have a rough time estimate how long we'd be disabling this process if we decide to go that route?

nemacysts: @ernestosilv we can do the migration in dev/stage whenever y'all are ready

as for how long we'd need to disable things - that depends on how safely we want to do the migration as we have a script that will do things in as safe a manner as possible:
* disables autoscaling in the "old" cluster + creates creates copies of all the instances (at the same # of containers) in the "new" cluster
* waits until instances are healthy in the new cluster, and deletes instances in the old cluster once the new instances are all healthy
* waits ~5min to allow autoscaling information to settle and then re-enables autoscaling in the new cluster

...but since these shards are (to my understanding) "optional", we could do this pretty quickly in a single PR by renaming the source file to eks-$CLUSTER.yaml and letting the new instances come up naturally 

but in either case: definitely under 1hr unless things were to go horribly wrong!

PR #3822, opened 2024-04-01
Add service/instance/managed labels to HPAs. PAASTA-18226

EvanKrall:
This should make the kube_horizontalpodautoscaler_labels metric from kube-state-metrics more useful.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -910,10 +910,18 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             max_replicas,
             autoscaling_params,
         )
+
+        labels = {
+            paasta_prefixed("service"): self.service,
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -910,10 +910,17 @@ def get_autoscaling_metric_spec(
             max_replicas,
             autoscaling_params,
         )
+
+        labels = {
```
nemacysts, line 914: might be nice to add the pool here too - then we can easily grab all the HPAs for a given pool

nemacysts (approved):

jfongatyelp (approved):

EvanKrall:
I added pool and refactored the tests to use mock.ANY for labels instead of hard-coding the labels, since these tests aren't meant to test the labels.

jfongatyelp (approved):

PR #3823, opened 2024-04-03
Add --memory option to paasta local-run to allow overriding of memory…

EvanKrall:
… limit set in yelpsoa-configs.

In my first attempt at this branch, I tried adding a `memory` argument to `run_docker_container`, and having 

```diff
```


jfongatyelp (dismissed):

nemacysts (commented):
would we also want to add similar versions of this for the other resource options?

(i also think that we didn't add this initially since we were expecting folks to use the -y flag to a local soaconfigs with the edited values)

PR #3829, opened 2024-04-15
optional min/max instance values instead of defaults

kedarv:
Our auto created shards are being instantiated with a minimum/maximum of the defaults described in the diff. Eventually we’d like to only define autotune cpu configs, so this PR allows us to not define any min/max if desired

### DIFF
--- a/paasta_tools/contrib/service_shard_update.py
+++ b/paasta_tools/contrib/service_shard_update.py
@@ -56,14 +56,12 @@ def parse_args():
         "--min-instance-count",
         help="If a deploy group is added, the min_instance count to create it with",
         required=False,
-        default=1,
         dest="min_instance_count",
     )
     parser.add_argument(
         "--prod-max-instance-count",
         help="If a deploy group is added, the prod max_instance count to create it with",
         required=False,
-        default=100,
         type=int,
         dest="prod_max_instance_count",
     )
@@ -71,7 +69,6 @@ def parse_args():
         "--non-prod-max-instance-count",
         help="If a deploy group is added, the non-prod max_instance count to create it with",
         required=False,
-        default=5,
         type=int,
         dest="non_prod_max_instance_count",
     )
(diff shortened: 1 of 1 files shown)


karmeleon (approved):
is the plan to add autotune max/min cpus next? i can take a crack at it too

kedarv:
> is the plan to add autotune max/min cpus next? i can take a crack at it too

I think that'd be a good idea, but not strictly necessary. this is mostly getting rid of the annoyance of capping shards at arbitrary default values

karmeleon (approved):

nemacysts:
@karmeleon could you add a PR/commit description with a little more context? i have a couple questions re: this change that are potentially answered with a little more info :)

kedarv:
> @karmeleon could you add a PR/commit description with a little more context? i have a couple questions re: this change that are potentially answered with a little more info :)

Updated!

paasta_tools/contrib/service_shard_update.py
```diff
@@ -194,14 +191,28 @@ def main(args):
 
                     instance_config = {
                         "deploy_group": f"{deploy_prefix}.{args.shard_name}",
-                        "min_instances": args.min_instance_count,
-                        "max_instances": args.prod_max_instance_count
-                        if deploy_prefix == "prod"
-                        else args.non_prod_max_instance_count,
```
nemacysts, line 200: ah cool, so one thing we should make sure to look at is what paasta will do if there's no min/max instances set, but autoscaling is configured - iirc, it might take until the first autotune run for min/max instances to be set (and therefore for the instance to actually be autoscaled)

if so, i think we need to update what paasta defaults to if there an autoscaling section set and no min/max (explicitly or in autotune)

nemacysts: @karmeleon looks like paasta will not setup an HPA (the thing that actually controls autoscaling) and will default to running a single instance

i *think* that once autotune runs, it will set a value for min/max_instance tho

that said: i think we can tweak paasta's behavior here

PR #3830, opened 2024-04-15
Add support for autotune limits in service_shard_update

karmeleon:
Added to PaaSTA at large in #3744. Also pulls in commit from #3829 to make min/max instance values optional and adds a new `--dry-run` param that skips committing and pushing changes to yelpsoa-configs.

Confirmed that the new kwargs do as expected, generating valid configs. If anyone knows a way to do this that doesn't involve a mess of nested if statements, I'm all ears lol


paasta_tools/contrib/service_shard_update.py
```diff
@@ -194,14 +240,28 @@ def main(args):
 
                     instance_config = {
                         "deploy_group": f"{deploy_prefix}.{args.shard_name}",
-                        "min_instances": args.min_instance_count,
-                        "max_instances": args.prod_max_instance_count
-                        if deploy_prefix == "prod"
-                        else args.non_prod_max_instance_count,
```
karmeleon, line 200: @nemacysts (from #3829) how would you suggest testing paasta's behavior with no max/min instances?

nemacysts: @karmeleon sorry, yesterday was a busy day!

i don't think we've setup the local paasta playground for autoscaling, but we should be able to create an instance in infrastage (or any test cluster) and see what happens

lemme do that real fast, actually :)

nemacysts: whoops - replied in the wrong place: https://github.com/Yelp/paasta/pull/3829#discussion_r1569091874

nemacysts: PAASTA-18283

karmeleon: thanks for checking! to clarify, we're totally okay with min/max instances both being at 1 for brand new shards, since they typically see traffic ramp up gradually enough for autotune to deal with it. do you think it's safe to ship this branch now, or should we wait for that ticket to be completed?

nemacysts: i'm trying to verify that autotune will do the "right" thing and setup min/max instances - but when i ran the rightsizer scripts locally it didn't seem to...that said, i'm not sure if it's cause i need to wait for splunk to have more data

nemacysts: whoops, sorry - forgot about this: i think i hallucinated that autotune would fixup the autoscaling config: my test instance has not had autotune add min/max instances correctly :(

nemacysts: (i.e., we'll either want to revert the {min,max}_instances changes here or tackle PAASTA-18283)

paasta_tools/contrib/service_shard_update.py
```diff
@@ -39,6 +39,13 @@ def parse_args():
         action="store_true",
         dest="verbose",
     )
+    parser.add_argument(
+        "-d",
+        "--dry-run",
+        help="Do not commit changes to git",
+        action="store_true",
+        dest="dry_run",
+    )
```
nemacysts, line 48: ++

(i'm also a fan of `--for-real` flags so that dry-run is the default, but that would be a little more annoying to rollout here)

paasta_tools/contrib/service_shard_update.py
```diff
@@ -214,6 +274,56 @@ def main(args):
...
+                        (
+                            args.autotune_min_cpus,
+                            args.autotune_max_cpus,
+                            args.autotune_min_mem,
+                            args.autotune_max_mem,
+                            args.autotune_min_disk,
+                            args.autotune_max_disk,
+                        )
+                    ):
+                        instance_config["autotune"] = {}
```
nemacysts, line 287:
```suggestion
                        instance_config["autotune_limits"] = {}
```
is the key we want instead

---

might also be nice to do something like:
```python
limit_config = instance_config["autotune_limits"]
...
limit_config["cpus"]["min"] = args.autotune_min_cpus
...
```
so that we don't need to repeat `instance_config["autotune_limits"]` so many times :)

nemacysts: i think maybe we could trim this code down by doing something like:
```
limit_config["cpus"] = {
    "min": args.autotune_min_cpus,
    "max": args.autotune_max_cpus
}
limit_config["mem"] = {
    "min": args.autotune_min_mem,
    "max": args.autotune_max_mem
}
limit_config["disk"] = {
    "min": args.autotune_min_disk,
    "max": args.autotune_max_disk
}```
and then adding a processing step that removes any min/max keys with `None` values - and any resource (e.g., cpus, mem, disk) keys with no subkeys

(that said, maybe the code to do that processing step wouldn't look too different and doing this would be a waste of time :p)

**Review by nemacysts — COMMENTED** [review]
generally lgtm - we'll probably want to back out the min/max instance changes + fix the autotune limits typo, but otherwise i don't see any concerns here

**Comment by karmeleon** [comment]
@nemacysts sorry for the 6 months of suspense! kinda forgot about this PR until we ran into some super anemic ssr shards with 0.1 cpus causing hella timeouts on prod lol

**Review by nemacysts — APPROVED** [review]
lgtm - unless you'd like to add tests here, i'm assuming that you ran this locally and got the expected output :)

(and in any case, we'd probably wanna wait until monday to release this anyway :p)

File: paasta_tools/contrib/service_shard_update.py [hunk 4]
```diff
@@ -217,6 +268,40 @@ def main(args):
...
+                            "max": args.autotune_max_mem,
+                        }
+                        limit_config["disk"] = {
+                            "min": args.autotune_min_disk,
+                            "max": args.autotune_max_disk,
+                        }
+
+                        # remove any None values to keep the config clean
+                        for resource in list(limit_config):
+                            for key in list(limit_config[resource]):
```
nemacysts, line 297: (these list()'s aren't necessary, but i'm assuming they're were helpful in some way while you were debugging and converting from a view to a list shouldn't really impact perf in a meaningful way for this script)

karmeleon: unfortunately they are necessary under py3 :(

```
  File "paasta_tools/contrib/service_shard_update.py", line 350, in <module>
    main(args)
  File "paasta_tools/contrib/service_shard_update.py", line 297, in main
    for key in limit_config[resource]:
RuntimeError: dictionary changed size during iteration
```

karmeleon:
yep, got the expected output :) will merge this next week

paasta_tools/contrib/service_shard_update.py
nemacysts: oh d'oh - i'm dumb: we're mutating the dict 

i blame friday-brain :p

PR #3835, opened 2024-04-19
update paasta validation code

drmorr0:
Update the paasta validate code in preparation for multi-metrics work.  This adds a few additional checks to the autoscaling validation to make sure that we're not using a parameter that isn't supported for a particular metrics-provider, and it refactors the autoscaling block into its own separate file.

In the second commit in the series, we push more of the validation (around number ranges) into the jsonschema itself instead of having to write custom code to validate numeric ranges.

## Testing done
- make test passes
- run on all services in yelpsoa-configs to make sure they all still pass validation
- confirmed that, e.g., adding `offset` into a field in yelpsoa-configs causes validation to fail


paasta_tools/cli/cmds/validate.py
```diff
@@ -187,17 +198,29 @@ def no_duplicate_instance_names_message(service, cluster):
...
+    autoscaling_path = "schemas/autoscaling_schema.json"
+    schema = pkgutil.get_data("paasta_tools.cli", schema_path).decode()
+    autoscaling_ref = pkgutil.get_data("paasta_tools.cli", autoscaling_path).decode()
+
+    base_schema = json.loads(schema)
+    autoscaling_schema = json.loads(autoscaling_ref)
+    store = {
+        "base": base_schema,
+        autoscaling_schema["$id"]: json.loads(autoscaling_ref),
+    }
```
nemacysts, line 216: what does this end up doing for schemas like tron's - where autoscaling doesn't exist?

paasta_tools/cli/cmds/validate.py
```diff
@@ -299,16 +322,11 @@ def validate_schema(file_path: str, file_type: str) -> bool:
...
     try:
-        schema = get_schema(file_type)
+        validator = get_schema_validator(file_type)
     except Exception as e:
         print(f"{SCHEMA_ERROR}: {file_type}, error: {e!r}")
         return False
 
-    if schema is None:
-        print(f"{SCHEMA_NOT_FOUND}: {file_path}")
-        return False
```
nemacysts, line 309: what happens if there's no schema now? will creating the validator thrown an exception which then gets caught by the catch-all handler above?

nemacysts: from the test, i assume that's what will happen/what we wanted to happen

paasta_tools/cli/cmds/validate.py
```diff
@@ -587,10 +694,9 @@ def validate_autoscaling_configs(service_path):
         for instance, instance_config in load_all_instance_configs_for_service(
             service=service, cluster=cluster, soa_dir=soa_dir
         ):
-
             if (
                 instance_config.get_instance_type() in K8S_TYPES
-                and instance_config.is_autoscaling_enabled()
+                and instance_config.is_autoscaling_enabled()  # type: ignore
```
nemacysts, line 699: just curious: what was the error here? i assume something to do with the weird way the instance config classes are setup inheritance-wise?

nemacysts (commented):
generally lgtm - just have a couple questions

paasta_tools/cli/cmds/validate.py
drmorr0: This is just setting up a cacheable "store" of schemas so that you can reference them later on.  If you don't actually reference anything (like in the tron case) then it does nothing.

I added a comment to this effect.

paasta_tools/cli/cmds/validate.py
drmorr0: `instance_config` is a generic `InstanceConfig` object, which has no `is_autoscaling_enabled` method.  The check above `get_instance_type() in K8S_TYPES` narrows it down to a KubernetesDeploymentConfig which does have an `is_autoscaling_enabled` method.  The alternate way to do this would be to assert that instance_config is either a `KubernetesDeploymentConfig` or an `EksDeploymentConfig` but..... that seems ugly.

I'll add a comment here.

drmorr0: Actually -- I can do a cast to LongRunningServiceConfig, which is slightly more explicit.

paasta_tools/cli/cmds/validate.py
```diff
@@ -601,91 +720,80 @@ def validate_autoscaling_configs(service_path):
...
-                    setpoint = autoscaling_params["setpoint"]
-                    if setpoint <= 0:
-                        returncode = False
-                        print(
-                            failure(
-                                msg="Autoscaling configuration is invalid: setpoint must be greater than zero",
-                                link="",
-                            )
-                        )
+                autoscaling_params = instance_config.get_autoscaling_params()  # type: ignore
```
nemacysts, line 723: it's too bad that mypy doesn't see the cast above and apply the cast here as well :p

paasta_tools/cli/cmds/validate.py
```diff
@@ -572,7 +598,96 @@ def __is_templated(service: str, soa_dir: str, cluster: str, workload: str) -> b
...
+        )
+    if len(cast(LongRunningServiceConfig, instance_config).get_registrations()) > 1:
+        raise AutoscalingValidationError(
+            "active-requests metrics provider doesn't support instances with multiple registrations"
+        )
+    # This is a bit incorrect, since the default autoscaling setpoint is 0.8,
+    # someone could theoretically bypass this by setting setpoint: 0.8 in their
+    # soaconfigs; but I think it's approximately fine for now
+    if (
+        autoscaling_params.get("setpoint", DEFAULT_AUTOSCALING_SETPOINT)
```
nemacysts, line 621: i was just thinking that it might be more future-proof to have an allowlist of `$provider: $valid_params` so that we don't have all these methods more or less checking similar fields - but that sounds like something we can refactor later :)

paasta_tools/cli/cmds/validate.py
```diff
@@ -601,91 +720,80 @@ def validate_autoscaling_configs(service_path):
...
+                        and not should_skip_cpu_override_validation
                     ):
-                        returncode = False
-                        print(
-                            failure(
-                                msg=f"CPU override detected for a CPU-autoscaled instance in {cluster}: {service}.{instance}. Please read "
-                                "the following link for next steps:",
-                                link="y/override-cpu-autotune",
+                        _validate_cpu_autoscaling_configs(autoscaling_params)
+                        # Do some extra validation below: we don't abstract that into the above function
```
nemacysts, line 750: i really should have tossed all this nonsense into a function when i initially wrote this :p

nemacysts (approved):

paasta_tools/cli/cmds/validate.py
drmorr0: Oh I can fix that.

paasta_tools/cli/cmds/validate.py
drmorr0: Oh that's probably a good idea.

drmorr0: Actually - I'm going to do that in #3837

PR #3837, opened 2024-04-23
multiple metrics support for paasta

drmorr0:
This change introduces the ability to add multiple metrics configs to an HPA via paasta.  We transform the existing AutoscalingParamsDict into a new version that looks like this:

```yaml
metrics_providers:
  - type: cpu
    setpoint: 0.8
  - type: active-requests
    desired_requests_per_replica: 10
scaledown_policies: ...
max_instances_alert_threshold: ...
```

In this PR, we support both the old and the new ways of doing things, and we transform the old autoscaling params dict into the new version when we read the YAML files.  After this PR is shipped, we will do a mass-refactor of yelpsoa into the new format and remove the "transformation" code so that only the new format is supported.  

The first commit shows the bulk of the changes, and the second commit shows additional validation code and tests that are performed.

## Testing done
- make test/tox -e mypy passes (with new tests added)
- paasta validate passes on all existing yelpsoa instances
- paasta validate passes on yelpsoa instances using the new format (just tested locally, there's none of these in production yet)
- setup_prometheus_adapter_config diffs in pnw-prod and nova-prod are empty
- confirmed that the before/after instance configs and the prometheus adapter configs for all of yelpsoa are identical (part of the testing in https://github.yelpcorp.com/sysgit/yelpsoa-configs/pull/43847)

## Open questions
- What to do if someone's using `active-requests` in check_autoscaler_max_instances?  ([link1](https://github.com/Yelp/paasta/pull/3837/files#diff-329d31e4ca35b5deeb89fe08cff4f72d9fc55c9b79ae786e415886f433e2830fR107), [link2](https://github.com/Yelp/paasta/pull/3837/files?diff=split&w=1#diff-719160feb410077ed31b8c62c719bf6d2495cbd251119458fb74ad4aea59942fR394))

cc @sclg-yelp

### DIFF
--- a/paasta_tools/autoscaling/autoscaling_service_lib.py
+++ b/paasta_tools/autoscaling/autoscaling_service_lib.py
@@ -91,7 +91,7 @@ ServiceAutoscalingInfo = namedtuple(
 )
 
 
-SERVICE_METRICS_PROVIDER_KEY = "metrics_provider"
+SERVICE_METRICS_PROVIDER_KEY = "metrics_providers"
 DECISION_POLICY_KEY = "decision_policy"
 
 AUTOSCALING_DELAY = 300
@@ -566,7 +566,7 @@ def get_new_instance_count(
     persist_data: bool,
 ):
     autoscaling_decision_policy = get_decision_policy(
-        autoscaling_params[DECISION_POLICY_KEY]
+        autoscaling_params[SERVICE_METRICS_PROVIDER_KEY][0][DECISION_POLICY_KEY]
     )
 
     zookeeper_path = compose_autoscaling_zookeeper_root(
@@ -605,7 +605,7 @@ def get_utilization(
     mesos_tasks,
 ):
     autoscaling_metrics_provider = get_service_metrics_provider(
-        autoscaling_params[SERVICE_METRICS_PROVIDER_KEY]
+        autoscaling_params[SERVICE_METRICS_PROVIDER_KEY][0]
     )
 
     return autoscaling_metrics_provider(
@@ -652,7 +652,7 @@ def autoscale_marathon_instance(
             )
             error = get_error_from_utilization(
                 utilization=utilization,
-                setpoint=autoscaling_params["setpoint"],
+                setpoint=autoscaling_params["metrics_providers"][0]["setpoint"],
                 current_instances=current_instances,
             )
             num_healthy_instances = len(marathon_tasks)
@@ -762,7 +762,7 @@ def _record_autoscaling_decision(
         "paasta_cluster": marathon_service_config.cluster,
         "paasta_instance": marathon_service_config.instance,
         "paasta_pool": marathon_service_config.get_pool(),
-        "decision_policy": autoscaling_params[DECISION_POLICY_KEY],  # type: ignore
+        "decision_policy": autoscaling_params[SERVICE_METRICS_PROVIDER_KEY][0][DECISION_POLICY_KEY],  # type: ignore
     }
     if yelp_meteorite:
         gauge = yelp_meteorite.create_gauge("paasta.service.instances", meteorite_dims)
@@ -802,7 +802,9 @@ def get_configs_of_services_to_scale(
             if (
                 instance_config.get_max_instances()
                 and instance_config.get_desired_state() == "start"
-                and instance_config.get_autoscaling_params()["decision_policy"]
+                and instance_config.get_autoscaling_params()["metrics_providers"][0][
+                    "decision_policy"
+                ]
                 != "bespoke"
             ):
                 configs.append(instance_config)
--- a/paasta_tools/check_autoscaler_max_instances.py
+++ b/paasta_tools/check_autoscaler_max_instances.py
@@ -99,20 +99,32 @@ async def check_max_instances(
                 autoscaling_status["desired_replicas"]
                 >= autoscaling_status["max_instances"]
             ):
-                threshold = job_config.get_autoscaling_max_instances_alert_threshold()
+
                 metrics_provider_configs = job_config.get_autoscaling_params()[
                     "metrics_providers"
                 ]
 
-                # TODO: this doesn't work for metrics_providers that don't use setpoint (e.g. active-requests)
-                max_setpoint = max(
-                    [provider["setpoint"] for provider in metrics_provider_configs]
-                )
-                metric_threshold_target_ratio = threshold / max_setpoint
-
                 status = pysensu_yelp.Status.UNKNOWN
                 output = "how are there no metrics for this thing?"
-                for metric in autoscaling_status["metrics"]:
+
+                # This makes an assumption that the metrics currently used by the HPA are exactly the same order (and
+                # length) as the list of metrics_providers dictionaries. This should generally be true, but between
+                # yelpsoa-configs being pushed and the HPA actually being updated it may not be true. This might cause
+                # spurious alerts, but hopefully the frequency is low. We can add some safeguards if it's a problem.
+                # (E.g. smarter matching between the status dicts and the config dicts, or bailing/not alerting if the
+                # lists aren't the same lengths.)
+                for metric, metrics_provider_config in zip(
+                    autoscaling_status["metrics"], metrics_provider_configs
+                ):
+
+                    setpoint = metrics_provider_config["setpoint"]
+                    threshold = metrics_provider_config.get(
+                        "max_instances_alert_threshold",
+                        setpoint,
+                    )
+
+                    metric_threshold_target_ratio = threshold / setpoint
+
                     current_value = suffixed_number_value(metric["current_value"])
                     target_value = suffixed_number_value(metric["target_value"])
 
@@ -123,7 +135,7 @@ async def check_max_instances(
                             " ratio of current value to target value"
                             f" ({current_value} / {target_value}) is greater than the"
                             " ratio of max_instances_alert_threshold to setpoint"
-                            f" ({threshold} / {max_setpoint})"
+                            f" ({threshold} / {setpoint})"
                         )
                     else:
                         status = pysensu_yelp.Status.OK
@@ -132,7 +144,7 @@ async def check_max_instances(
                             " ratio of current value to target value"
                             f" ({current_value} / {target_value}) is below the ratio of"
                             f" max_instances_alert_threshold to setpoint ({threshold} /"
-                            f" {max_setpoint})"
+                            f" {setpoint})"
                         )
             else:
                 status = pysensu_yelp.Status.OK
@@ -149,7 +161,7 @@ async def check_max_instances(
                         " because your service is overloaded, but cannot scale any"
                         " higher because of max_instances. You may want to bump"
                         " max_instances. To make this alert quieter, adjust"
-                        " autoscaling.max_instances_alert_threshold in yelpsoa-configs."
+                        " autoscaling.metrics_providers[n].max_instances_alert_threshold in yelpsoa-configs."
                     ),
                 }
             )
--- a/paasta_tools/check_autoscaler_max_instances.py
+++ b/paasta_tools/check_autoscaler_max_instances.py
@@ -100,8 +100,15 @@ async def check_max_instances(
                 >= autoscaling_status["max_instances"]
             ):
                 threshold = job_config.get_autoscaling_max_instances_alert_threshold()
-                setpoint = job_config.get_autoscaling_params()["setpoint"]
-                metric_threshold_target_ratio = threshold / setpoint
+                metrics_provider_configs = job_config.get_autoscaling_params()[
+                    "metrics_providers"
+                ]
+
+                # TODO: this doesn't work for metrics_providers that don't use setpoint (e.g. active-requests)
+                max_setpoint = max(
+                    [provider["setpoint"] for provider in metrics_provider_configs]
+                )
+                metric_threshold_target_ratio = threshold / max_setpoint
 
                 status = pysensu_yelp.Status.UNKNOWN
                 output = "how are there no metrics for this thing?"
@@ -116,7 +123,7 @@ async def check_max_instances(
                             " ratio of current value to target value"
                             f" ({current_value} / {target_value}) is greater than the"
                             " ratio of max_instances_alert_threshold to setpoint"
-                            f" ({threshold} / {setpoint})"
+                            f" ({threshold} / {max_setpoint})"
                         )
                     else:
                         status = pysensu_yelp.Status.OK
@@ -125,7 +132,7 @@ async def check_max_instances(
                             " ratio of current value to target value"
                             f" ({current_value} / {target_value}) is below the ratio of"
                             f" max_instances_alert_threshold to setpoint ({threshold} /"
-                            f" {setpoint})"
+                            f" {max_setpoint})"
                         )
             else:
                 status = pysensu_yelp.Status.OK
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -28,6 +28,7 @@ from typing import cast
 from typing import Dict
 from typing import List
 from typing import Optional
+from typing import Set
 from typing import Tuple
 from typing import Union
 
@@ -680,6 +681,7 @@ def validate_autoscaling_configs(service_path: str) -> bool:
     """
     soa_dir, service = path_to_soa_dir_service(service_path)
     returncode = True
+    link = ""
     skip_cpu_override_validation_list = (
         load_system_paasta_config().get_skip_cpu_override_validation_services()
     )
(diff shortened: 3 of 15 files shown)


paasta_tools/long_running_service_tools.py
```diff
@@ -345,20 +360,47 @@ def limit_instance_count(self, instances: int) -> int:
...
+            self.config_dict.get("autoscaling", AutoscalingParamsDict({}))
         )
+        if "metrics_providers" not in params or len(params["metrics_providers"]) == 0:
+            params["metrics_providers"] = [default_provider_params]
+        else:
+            for i in range(0, len(params["metrics_providers"])):
+                params["metrics_providers"][i] = deep_merge_dictionaries(
+                    overrides=params["metrics_providers"][i],
+                    defaults=default_provider_params,
+                )
```
nemacysts, line 379: non-blocking suggestion:
```suggestion
            params["metrics_providers"] = [
                deep_merge_dictionaries(
                    overrides=provider,
                    defaults=default_provider_params,
                )
                for provider in params["metrics_providers"]
            ]
```
i think this would do the same thing and get rid of the indexing - at the cost of folks needing to read a list-comp

(i'm personally somewhat averse to `for i in range(...)` and explicit indexing, but generally just for code-style reasons :p)

paasta_tools/long_running_service_tools.py
```diff
@@ -345,20 +360,47 @@ def limit_instance_count(self, instances: int) -> int:
...
+        return params
 
     def get_autoscaling_max_instances_alert_threshold(self) -> float:
         autoscaling_params = self.get_autoscaling_params()
         return autoscaling_params.get(
-            "max_instances_alert_threshold", autoscaling_params["setpoint"]
+            # TODO this default doesn't make sense for metrics providers that don't use setpoint
```
nemacysts, line 385: @EvanKrall what are your thoughts here?

i can see us resolving this in two ways:

1.  look at the autoscaling params and use the `desired_active_requests_per_replica` if present
2.  give up on `desired_active_requests_per_replica` and rename it to `setpoint` so that the code is neater" (at the cost of people sometimes getting confused as to what setpoint means :p)

paasta_tools/long_running_service_tools.py
```diff
@@ -345,20 +360,47 @@ def limit_instance_count(self, instances: int) -> int:
...
+            "max_instances_alert_threshold",
+            DEFAULT_AUTOSCALING_SETPOINT,
+        )
+
+    def get_autoscaling_metrics_provider(
+        self, provider_type: MetricsProviderType
+    ) -> Optional[MetricsProviderDict]:
+        autoscaling_params = self.get_autoscaling_params()
+        # We only allow one metric provider of each type, so we can bail early if we find a match
```
nemacysts, line 394: are we validating this somewhere? (i don't remember seeing anything in paasta validate for this, but maybe this was in the PR where you updated the schema?)

paasta_tools/long_running_service_tools.py
```diff
@@ -345,20 +360,47 @@ def limit_instance_count(self, instances: int) -> int:
...
+        for provider in autoscaling_params["metrics_providers"]:
+            if provider["type"] == provider_type:
+                return provider
+        return None
+
+    def should_use_metrics_provider(self, provider_type: MetricsProviderType) -> bool:
```
nemacysts, line 400: comment: i'd say we could lru_cache this - but it doesn't seem like this is getting called in a hot loop anywhere so probably fine

paasta_tools/kubernetes_tools.py
```diff
@@ -898,14 +877,47 @@ def get_autoscaling_metric_spec(
...
+
+        if self.get_desired_state() == "stop":
+            return None
+
+        if not self.is_autoscaling_enabled():
+            return None
+
+        autoscaling_params = self.get_autoscaling_params()
+        if autoscaling_params["metrics_providers"][0]["decision_policy"] == "bespoke":
+            return None
```
nemacysts, line 906: should we perhaps add some validation so that folks can't mix bespoke autoscaling with HPA-based autoscaling? i can see someone accidentally bypassing this check by adding an HPA-based provider above an existing bespoke provider

paasta_tools/kubernetes_tools.py
```diff
@@ -921,7 +933,7 @@ def get_autoscaling_metric_spec(
         hpa = V2beta2HorizontalPodAutoscaler(
             kind="HorizontalPodAutoscaler",
             metadata=V1ObjectMeta(
-                name=name, namespace=namespace, annotations=annotations, labels=labels
+                name=name, namespace=namespace, annotations=dict(), labels=labels
```
nemacysts, line 936: comment: heh, i was confused by this diff until i read the previous code and realized (as you did) that we never wrote to `annotations` 🤣

paasta_tools/long_running_service_tools.py
nemacysts: (and also ++ to this refactor)

paasta_tools/paasta_service_config_loader.py
```diff
@@ -179,6 +183,13 @@ def _create_service_config(
         """
 
         merged_config = self._get_merged_config(config)
+        if (
+            "autoscaling" in merged_config
+            and "metrics_providers" not in merged_config["autoscaling"]
+        ):
+            merged_config["autoscaling"] = _transform_autoscaling_params_dict(
+                copy.deepcopy(merged_config["autoscaling"])
+            )
```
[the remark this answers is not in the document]
drmorr0: Correct.

paasta_tools/long_running_service_tools.py
drmorr0: We are not validating this yet but we will be.  I need to add that code still.

paasta_tools/kubernetes_tools.py
drmorr0: Yea, good callout.  I'll add that to my list of things to validate.

paasta_tools/autoscaling/autoscaling_service_lib.py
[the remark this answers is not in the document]
drmorr0: I am also 75% certain that none of this is needed.  But if I don't make these changes then it will fail mypy checks.  I am 50% tempted to just submit a PR to delete this file entirely but it seems like one of those things that is maybe still called somewhere in some arcane way that's hard to suss out.

paasta_tools/cli/cmds/validate.py
```diff
@@ -712,75 +713,87 @@ def validate_autoscaling_configs(service_path: str) -> bool:
...
-                try:
-                    if autoscaling_params["metrics_provider"] == "active-requests":
-                        _validate_active_requests_autoscaling_configs(
-                            instance_config, autoscaling_params
-                        )
+                for metrics_provider in autoscaling_params["metrics_providers"]:
+                    try:
+                        if (
+                            metrics_provider["type"]
+                            == MetricsProviderType.ActiveRequests
```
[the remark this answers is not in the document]
drmorr0: TIL

honestly I never understood the difference between `==` and `is` in the first place.

paasta_tools/kubernetes_tools.py
```diff
@@ -783,51 +785,22 @@ def get_autoscaling_scaling_policy(
...
-        metrics = []
-        target = autoscaling_params["setpoint"]
-        annotations: Dict[str, str] = {}
+    def get_autoscaling_provider_spec(
+        self, name: str, namespace: str, provider: MetricsProviderDict
+    ) -> List[V2beta2MetricSpec]:
+        target = provider["setpoint"]
         # we will have both the SFX adapter and the Prometheus adapter serving metrics for some
         # HPAs and, since they may be looking at the same thing, we need to suffix one of these metric
         # identifiers because metric names inside an HPA must be unique.
```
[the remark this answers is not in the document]
drmorr0: I was wondering if I should delete this.

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -198,75 +196,71 @@ def _minify_promql(query: str) -> str:
...
-
-        return True, None
+def _uses_prometheus(provider: Optional[MetricsProviderDict]) -> bool:
+    if provider is None:
+        return False
 
-    return False, "did not request uwsgi autoscaling"
+    # Check provider-specific configs here
+    if provider["type"] == MetricsProviderType.Uwsgi:
+        return provider.get("use_prometheus", DEFAULT_USE_PROMETHEUS_UWSGI)
```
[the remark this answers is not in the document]
drmorr0: I was also wondering if this was still relevant or not.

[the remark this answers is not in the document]
nemacysts: yea, i've been meaning to come back to this at some point - but there hasn't been much of a demand for moving-average cpu autoscaling for a while and it keeps slipping further and further down my personal backlog :p

if it makes your work easier, i'd honestly ship removing all the prometheus cpu code :)

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -198,75 +196,71 @@ def _minify_promql(query: str) -> str:
...
+            service, instance_config, metrics_provider_config, paasta_cluster
+        )
+    if metrics_provider_config["type"] == MetricsProviderType.ActiveRequests:
+        return create_instance_active_requests_scaling_rule(
+            service, instance_config, metrics_provider_config, paasta_cluster
+        )
+    if metrics_provider_config["type"] == MetricsProviderType.ArbitraryPromQL:
+        return create_instance_arbitrary_promql_scaling_rule(
+            service, instance_config, metrics_provider_config, paasta_cluster
+        )
```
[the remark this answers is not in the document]
drmorr0: I thought about doing something like this, but I kinda prefer the explicitness of the ifs.... what I _really_ want is a case statement where I can match on the type, so that you never forget to update this in the future, but you need to update to a more recent version of Python for this.

paasta_tools/setup_prometheus_adapter_config.py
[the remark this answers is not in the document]
drmorr0: Ok it's gone: https://github.com/Yelp/paasta/pull/3838

paasta_tools/setup_prometheus_adapter_config.py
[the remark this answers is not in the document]
nemacysts: i *think* we should be unblocked relatively soon on bumping the minimum python version - but yea :p

paasta_tools/long_running_service_tools.py
drmorr0: Oh that's smart.  I don't know why I didn't do that.  I don't like explicit indexing either.

paasta_tools/long_running_service_tools.py
drmorr0: eh, YAGNI, probably

paasta_tools/long_running_service_tools.py
drmorr0: This code has been added.

paasta_tools/kubernetes_tools.py
drmorr0: Done

paasta_tools/cli/cmds/validate.py
```diff
@@ -627,30 +646,30 @@ def _validate_active_requests_autoscaling_configs(
...
         raise AutoscalingValidationError(
             "arbitrary promql metrics provider requires prometheus_adapter_config to be set"
         )
 
 
-def _validate_autoscaling_config(autoscaling_params: AutoscalingParamsDict):
-    metrics_provider = autoscaling_params["metrics_provider"]
-    for field in INVALID_AUTOSCALING_FIELDS[metrics_provider]:
-        if field in autoscaling_params:
+def _validate_autoscaling_config(metrics_provider_config: MetricsProviderDict):
```
nemacysts, line 667:
```suggestion
def _validate_autoscaling_config(metrics_provider_config: MetricsProviderDict) -> None:
```

i was debating not commenting here, but then i saw that you fixed the missing typing above :p

paasta_tools/cli/cmds/validate.py
```diff
@@ -692,70 +712,100 @@ def validate_autoscaling_configs(service_path: str) -> bool:
...
+                    try:
+                        # Generic validation of the config
+                        _validate_autoscaling_config(metrics_provider)
 
-                    elif autoscaling_params["metrics_provider"] == "arbitrary_promql":
-                        _validate_arbitrary_promql_autoscaling_configs(
-                            autoscaling_params
-                        )
+                        # Multi-metrics specific validation:
+                        # 1. Bespoke policies shouldn't also use CPU
```
nemacysts, line 724: i think it'd be worth documenting this with something a little more strongly worded lest someone "clean" this up by only checking for cpu autoscaling in conjunction with bespoke autoscaling

```suggestion
                        # 1. Bespoke policies cannot use multi-metric scaling
```

paasta_tools/long_running_service_tools.py
```diff
@@ -39,17 +40,39 @@
...
+
+def all_metrics_providers() -> List[str]:
+    return [
+        METRICS_PROVIDER_CPU,
+        METRICS_PROVIDER_UWSGI,
+        METRICS_PROVIDER_GUNICORN,
+        METRICS_PROVIDER_PISCINA,
+        METRICS_PROVIDER_ACTIVE_REQUESTS,
+        METRICS_PROVIDER_PROMQL,
+    ]
```
nemacysts, line 59: question: any reason to not make this a global constant?

paasta_tools/check_autoscaler_max_instances.py
```diff
@@ -100,8 +100,15 @@ async def check_max_instances(
...
-                metric_threshold_target_ratio = threshold / setpoint
+                metrics_provider_configs = job_config.get_autoscaling_params()[
+                    "metrics_providers"
+                ]
+
+                # TODO: this doesn't work for metrics_providers that don't use setpoint (e.g. active-requests)
+                max_setpoint = max(
+                    [provider["setpoint"] for provider in metrics_provider_configs]
+                )
+                metric_threshold_target_ratio = threshold / max_setpoint
```
EvanKrall, line 111: This feels weird to me so I've got a branch (based on this one) that makes max_instances_alert_threshold be per-metrics_provider: https://github.com/Yelp/paasta/pull/3850

paasta_tools/long_running_service_tools.py
EvanKrall: It is kind of odd that for active-requests we have some positive number 0-infinity, where for other providers `setpoint` is typically a number between 0-1 expressed as a ratio of some maximum resource utilization that's possible (100% worker util, 100% cpu, etc). We _could_ maybe change active-requests to have two parameters: `max_active_requests_per_replica` (which would be some larger number than `desired` is currently) and `setpoint`. But that's probably just adding complexity where it isn't necessary.

With my [branch](https://github.com/Yelp/paasta/pull/3850) we could have setpoint be a large number for active requests and 0-1 for other providers, and then max_instances_alert_threshold for each of those would also follow the same rules.

paasta_tools/cli/cmds/validate.py
drmorr0: lol I'm bad at saying things return nothing apparently

paasta_tools/long_running_service_tools.py
drmorr0: I've been too conditioned by golangci-lint

paasta_tools/check_autoscaler_max_instances.py
drmorr0: Cool, this makes more sense than what I was trying to do.  I misunderstood how this threshold was interpreted/applied.

paasta_tools/cli/cmds/validate.py
drmorr0: Whoops, yea, definitely brain-fart

nemacysts (approved):

paasta_tools/cli/cmds/validate.py
```diff
@@ -712,75 +713,87 @@ def validate_autoscaling_configs(service_path: str) -> bool:
...
-                            ),
-                            use_ruamel=True,
-                        )
-                        if config[instance].get("cpus") is None:
-                            # cpu autoscaled, but using autotuned values - can skip
-                            continue
+                        elif (
+                            metrics_provider["type"] == MetricsProviderType.CPU
+                            # to enable kew autoscaling we just set a decision policy of "bespoke", but
+                            # the metrics_provider is (confusingly) left as "cpu"
```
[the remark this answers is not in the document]
EvanKrall: yeah it really should be a different `metrics_provider`

EvanKrall (approved):

PR #3838, opened 2024-04-24
remove use_prometheus for cpu and uwsgi

drmorr0:
The use_prometheus flag for uwsgi is out of date because all the uwsgi stuff should be using prometheus, and the use_prometheus flag for cpu was an experiment that never went anywhere (or so sayeth @nemacysts).  So let's just delete them both.

## Testing done
- make tests/tox -e mypy
- setup_prometheus_adapter_config diff for nova-prod and pnw-prod is empty
- paasta validate (even using the stricter validation in https://github.com/Yelp/paasta/pull/3835) passes for all yelpsoa services

(following up after https://github.yelpcorp.com/sysgit/yelpsoa-configs/pull/43786)

COREJAVA-1250

cc @sclg-yelp

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -220,12 +220,6 @@
                         "moving_average_window_seconds": {
                             "type": "integer"
                         },
-                        "use_prometheus": {
-                            "type": "boolean"
-                        },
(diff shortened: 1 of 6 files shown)


nemacysts (approved):
++ - ty!

nemacysts:
(we'll also want to cleanup those soaconfigs before merging this)

PR #3842, opened 2024-04-26
Fix make test on Yelpy boxes

nemacysts:
Without this change, `make openapi-codegen` will fail on Yelpy boxes using unprivileged containers as moving the generated code from a temp directory to the final location will encounter permissions issues due to falling back to Docker instead of Podman and not passing user/group flags to Docker

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -6,7 +6,7 @@ docker_compose_version = 1.26.2
 
 [testenv]
 basepython = python3.8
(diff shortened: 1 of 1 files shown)


nemacysts:
@drmorr0 i think this is probably what you were running into recently :)

jfongatyelp (approved):

PR #3845, opened 2024-04-29
Set default executor pool to batch driver on k8s

chi-yelp:
To make it same as the default pool of `paasta spark-run`, so it won't cause problem when user did not specify `--pool` in the `paasta spark-run` command.


nemacysts (approved):
imo, it'd be nice to setup another SystemPaastaConfig option for this, but this works for now :)

PR #3849, opened 2024-04-30
PAASTA-18198: add pool_node_affinities to improve Karpenter AZ balance

gmdfalk:
This adds a new system config option, `pool_node_affinities`.
It's necessary to configure pods with a preselection of preferred 2-3 AZs per pool so that Karpenter does not try to consider all 6+ AZs in a region for scheduling decisions.

This option is a temporary solution and the proper fix will likely be to adopt something like the descheduler to achieve a better AZ balance (along with other benefits).
 
Signed-off-by: Max Falk <gfalk@yelp.com>

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2140,7 +2140,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         # need to check if there are node selectors/affinities. if there are none
         # and we create an empty affinity object, k8s will deselect all nodes.
         node_affinity = self.get_node_affinity(
-            system_paasta_config.get_pool_node_selectors()
+            system_paasta_config.get_pool_node_affinities()
         )
         if node_affinity is not None:
             pod_spec_kwargs["affinity"] = V1Affinity(node_affinity=node_affinity)
@@ -2286,7 +2286,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         return node_selectors
 
     def get_node_affinity(
-        self, pool_node_selectors: Dict[str, Dict[str, Any]] = None
+        self, pool_node_affinities: Dict[str, Dict[str, List[str]]] = None
     ) -> Optional[V1NodeAffinity]:
         """Converts deploy_whitelist and deploy_blacklist in node affinities.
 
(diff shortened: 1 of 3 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -2293,12 +2297,27 @@ def get_node_affinity(self) -> Optional[V1NodeAffinity]:
...
             )
         )
 
+        # PAASTA-18198: To improve AZ balance with Karpenter, we temporarily allow specifying zone affinities per pool
+        if pool_node_affinities:
+            current_pool_node_affinities = pool_node_affinities[self.get_pool()]
+            # If the service already has a node selector for a zone, we don't want to override it
+            if (
+                current_pool_node_affinities
+                and "topology.kubernetes.io/zone" not in node_selectors
```
ilkinmammadzada, line 2313: I'm wondering if `"topology.kubernetes.io/zone"` is the only label we should care about? 
Following labels include info about AZ:

- label_failure_domain_beta_kubernetes_io_zone
- label_karpenter_sh_nodepool
- label_topology_ebs_csi_aws_com_zone
- label_vpc_amazonaws_com_eni_config
- label_yelp_com_eni_config
- label_yelp_com_habitat

If the service owner adds one of them as node_selector our pod manifests will be inconsistent. Question is should we add them to the filter list?

paasta_tools/kubernetes_tools.py
```diff
@@ -681,6 +681,19 @@ def registration_label(namespace: str) -> str:
     return f"registrations.{PAASTA_ATTRIBUTE_PREFIX}{limited_namespace}"
 
 
+def contains_zone_label(node_selectors: Dict[str, NodeSelectorConfig]):
```
nemacysts, line 684: nit: this is missing a return type

paasta_tools/kubernetes_tools.py
```diff
@@ -2293,12 +2310,24 @@ def get_node_affinity(self) -> Optional[V1NodeAffinity]:
...
         requirements.extend(
             raw_selectors_to_requirements(
-                raw_selectors=self.config_dict.get("node_selectors", {}),
+                raw_selectors=node_selectors,
             )
         )
 
+        # PAASTA-18198: To improve AZ balance with Karpenter, we temporarily allow specifying zone affinities per pool
+        if pool_node_affinities:
+            current_pool_node_affinities = pool_node_affinities[self.get_pool()]
```
nemacysts, line 2322: is pool_node_affinities guaranteed to always contain the current pool?

paasta_tools/kubernetes_tools.py
```diff
@@ -681,6 +681,19 @@ def registration_label(namespace: str) -> str:
     return f"registrations.{PAASTA_ATTRIBUTE_PREFIX}{limited_namespace}"
 
 
+def contains_zone_label(node_selectors: Dict[str, NodeSelectorConfig]):
+    return any(
+        k in node_selectors
+        for k in (
+            "topology.kubernetes.io/zone",
```
ilkinmammadzada, line 688: creating a constant at the top of file would be nice. otherwise shipit

ilkinmammadzada (approved):

paasta_tools/kubernetes_tools.py
gmdfalk: thanks, added a check

paasta_tools/kubernetes_tools.py
gmdfalk: done but can't use `paasta_prefixed` in the constant

PR #3851, opened 2024-05-02
Add local-run role assumption usage docs

nurdann:
Let me know if there's another place it should go to.

### DIFF
--- a/docs/source/adhoc_instances.rst
+++ b/docs/source/adhoc_instances.rst
@@ -74,4 +74,4 @@ In the example below, specifying ``--cluster`` in the ``local-run`` command will
 
     paasta local-run --service <service-name> --pull --assume-pod-identity --instance <service-instance> --cluster pnw-prod --interactive
 
-Here, ``-I`` flag is used to get the interactive shell.
+Here, ``--interactive`` flag is used to get the interactive shell.
--- a/docs/source/adhoc_instances.rst
+++ b/docs/source/adhoc_instances.rst
@@ -65,9 +65,9 @@ defaults to an ipython repl. Also uses the canary version of the code::
 Assuming service role from another AWS account
 ----------------------------------------------
 
-If you need to run your instance with another account than your current environment's, e.g. use prod account from dev environment, you can specify ``--cluster`` and ``--assume-role-aws-account``.
+If you need to locally run your instance using a role from an AWS account that differs from your current environment's (e.g. you want to use a role from our production account in one of our dev environments), you will need to specify ``--cluster`` and ``--assume-role-aws-account``.
 
-In the example below, it's enough to specify ``--cluster`` so ``local-run`` will use configuration that are for ``pnw-prod`` and use prod account, no matter your current environment.
+In the example below, specifying ``--cluster`` in the ``local-run`` command will use instance configurations for ``pnw-prod`` and ``--assume-pod-identity`` will use the configured role from the prod account, no matter your current environment.
 
 
 .. code-block::sh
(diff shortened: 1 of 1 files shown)


docs/source/adhoc_instances.rst
```diff
@@ -61,3 +61,17 @@ defaults to an ipython repl. Also uses the canary version of the code::
     interactive:
       deploy_group: prod.canary
       mem: 10000
+
+Assuming service role from another AWS account
+----------------------------------------------
+
+If you need to run your instance with another account than your current environment's, e.g. use prod account from dev environment, you can specify ``--cluster`` and ``--assume-role-aws-account``.
+
+In the example below, it's enough to specify ``--cluster`` so ``local-run`` will use configuration that are for ``pnw-prod`` and use prod account, no matter your current environment.
```
nemacysts, line 70: optional wording suggestions:
```suggestion
If you need to locally run your instance using a role from an AWS account that differs from your current environment's (e.g. you want to use a role from our production account in one of our dev environments), you will need to specify ``--cluster`` and ``--assume-role-aws-account``.

In the example below, specifying ``--cluster`` in the ``local-run`` command will use instance configurations for ``pnw-prod`` and ``--assume-pod-identity`` will use the configured role from the prod account, no matter your current environment.
```

docs/source/adhoc_instances.rst
```diff
@@ -61,3 +61,17 @@ defaults to an ipython repl. Also uses the canary version of the code::
...
+
+
+.. code-block::sh
+
+    paasta local-run -s <service-name> --pull --assume-pod-identity -i <service-instance> --cluster pnw-prod -I
```
nemacysts, line 75: optional nit: imo, in documentation, we should try to use the long-form of options so that it's easier for folks to search for in the docs/in paasta --help
```suggestion
    paasta local-run --service <service-name> --pull --assume-pod-identity --instance <service-instance> --cluster pnw-prod --interactive
```

nemacysts (approved):
documentation-wise, this seems generally fine (at the very least, it's a good starting point :))

that said: do we really want to encourage this sort of thing by documenting this capability (or to go a little deeper: do we want this to be a capability that local-run provides)? this seems likely to become a footgun in the future.

additionally: if I'm not mistaken, this won't generally work for folks unless they've added their team's federated role to the list of roles that are allowed to assume the target service role

nurdann:
> additionally: if I'm not mistaken, this won't generally work for folks unless they've added their team's federated role to the list of roles that are allowed to assume the target service role

Yes, the federated team role must be able to assume the service role. 

> that said: do we really want to encourage this sort of thing by documenting this capability (or to go a little deeper: do we want this to be a capability that local-run provides)? this seems likely to become a footgun in the future.

This is a good feature to have because now you don't need to go to prod host just to run `paasta local-run` with AWS prod credentials.

docs/source/adhoc_instances.rst
```diff
@@ -61,3 +61,17 @@ defaults to an ipython repl. Also uses the canary version of the code::
...
+If you need to locally run your instance using a role from an AWS account that differs from your current environment's (e.g. you want to use a role from our production account in one of our dev environments), you will need to specify ``--cluster`` and ``--assume-role-aws-account``.
+
+In the example below, specifying ``--cluster`` in the ``local-run`` command will use instance configurations for ``pnw-prod`` and ``--assume-pod-identity`` will use the configured role from the prod account, no matter your current environment.
+
+
+.. code-block::sh
+
+    paasta local-run --service <service-name> --pull --assume-pod-identity --instance <service-instance> --cluster pnw-prod --interactive
+
+Here, ``-I`` flag is used to get the interactive shell.
```
nemacysts, line 77:
```suggestion
Here, ``--interactive`` flag is used to get the interactive shell.
```

whoops, we missed one flag update :)

PR #3852, opened 2024-05-02
Add a timer to some of our slowest cronjobs

jfongatyelp:
We currently have no way of knowing conclusively whether some of our cronjobs that don't emit start/stop lines are getting slower and approximations based on existing log lines seem to have always fallen flat in previous attempts, and relies on 100% success of our syslog-based pipeline. 

Instead, let's time this from a few of our slowest cronjobs itself and emit timings with some relevant info that I think will allow us to group by the things we need.  We can always adjust dimensions/names of metrics later if they don't end up being quite right, but I'd like to see if this is a worthwhile exercise at all first.

This adds a new helper `system_timer` in metrics_lib that gives us a timer to use in various system components of paasta that have to run periodically.
It uses a few base dimensions:
- `path`: the caller of system_timer, which will generally tell us which cronjob this is
- `host`: the hostname running this cronjob; probably not terribly useful but perhaps one host is slower than the other
- `parent_pid`: really only useful where we 'shard' our cronjobs by running multiple processes at once, spreading them across the full list of targets (e.g. service instances for s_k_j); had we just done 'pid' we might not be able to aggregate timings across a 'single' s_k_j 'run'/cron invocation, and instead only the individual calls for the subset of service instances passed to this particular run
- `pid`: not sure we'll need this but this would be the only distinguishing factor between each individual s_k_j call for a single cron run, for example

By collecting `parent_pid` (which seems to correctly get the bash pid started from the cronjob which then spawns parallel runs of setup_kubernetes_job), we can just grab relevant maximum/aggregate runtimes for each cron invocation, but by keeping individual s_k_j `pid` we could also see if there were outliers. 

When we stop timers, we also generally include the result, typically indicative of the exit code of the script. 

I've also added a few log lines that include info about the timers, _if_ we needed to with a log-based solution instead. 

I installed this version of paasta onto eksstage and so we can start graphing timings such as: https://grafana.yelpcorp.com/explore?schemaVersion=1&panes=%7B%2295q%22:%7B%22datasource%22:%22P627274A7C77417EA%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22avg(paasta_system_process_duration%7B%7D)%20by%20(path)%5Cn%5Cn%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22P627274A7C77417EA%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22from%22:%221715041055454%22,%22to%22:%221715042491591%22%7D%7D%7D&orgId=1

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
+++ b/paasta_tools/kubernetes/bin/paasta_secrets_sync.py
@@ -48,6 +48,7 @@ from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
 from paasta_tools.kubernetes_tools import sanitise_kubernetes_name
 from paasta_tools.kubernetes_tools import update_secret
 from paasta_tools.kubernetes_tools import update_secret_signature
+from paasta_tools.metrics import metrics_lib
 from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.secret_tools import get_secret_name_from_ref
 from paasta_tools.secret_tools import get_secret_provider
@@ -158,6 +159,14 @@ def main() -> None:
         cluster = args.cluster
     else:
         cluster = system_paasta_config.get_cluster()
+
+    timer = metrics_lib.system_timer(
+        dimensions=dict(
+            cluster=cluster,
+        )
+    )
+
+    timer.start()
     secret_provider_name = system_paasta_config.get_secret_provider_name()
     vault_cluster_config = system_paasta_config.get_vault_cluster_config()
     kube_client = KubeClient()
@@ -170,7 +179,7 @@ def main() -> None:
         )
     )
 
-    sys.exit(0) if sync_all_secrets(
+    result = sync_all_secrets(
         kube_client=kube_client,
         cluster=cluster,
         services_to_k8s_namespaces_to_allowlist=services_to_k8s_namespaces_to_allowlist,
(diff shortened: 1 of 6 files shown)


paasta_tools/setup_kubernetes_job.py
```diff
@@ -113,8 +115,28 @@ def main() -> None:
         logging.getLogger("kazoo").setLevel(logging.WARN)
         logging.basicConfig(level=logging.INFO)
 
+    # emit deploy events for updated jobs
     deploy_metrics = metrics_lib.get_metrics_interface("paasta")
 
+    # emit timing metrics for s_k_j
+    cluster = args.cluster or load_system_paasta_config().get_cluster()
+    system_metrics = metrics_lib.get_metrics_interface("paasta.setup_kubernetes_job")
```
jfongatyelp, line 123: If we start emitting this for more cronjobs, perhaps a more generic prefix is better here, then include the name of the script as a dimension or in the name of the timer below?

paasta_tools/setup_kubernetes_job.py
```diff
@@ -113,8 +113,19 @@ def main() -> None:
...
 
+    # emit timing metrics for s_k_j
+    cluster = args.cluster or load_system_paasta_config().get_cluster()
+    timer = metrics_lib.system_timer(
+        dimensions=dict(
+            cluster=cluster,
+            eks=args.eks,
+        ),
+    )
+    timer.start()
```
nemacysts, line 127: comment: i think we could technically do the same dimension update with a context manager - but it might be a bit more painful for not much gain :p

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -180,7 +189,12 @@ def main() -> None:
         vault_token_file=args.vault_token_file,
         overwrite_namespace=args.namespace,
         secret_type=args.secret_type,
-    ) else sys.exit(1)
+    )
+    timer.stop(tmp_dimensions={"result": result})
```
nemacysts, line 193: question: what will prometheus do with boolean dimensions? will it convert to 0/1? I ask because we seem to have a mix of result types: some are the integer exit code and others are boolean success states

nemacysts (approved):
i have some minor questions, but nothing blocking :)

nemacysts (approved):

PR #3857, opened 2024-05-09
MLCOMPUTE-1224 | add aws default region to prevent botocore.exceptions.NoRegionError

CaptainSame:
boto3 client complains with NoRegionError if there is no default region specified. https://fluffy.yelpcorp.com/i/zBXdRwX6FxszthrZLm6FVnWp8L4rjmzJ.html#L309-310
Replicating the behaviour in paasta spark-run, adding the AWS_DEFAULT_REGION env variable for jobs with Spark drivers on k8s


paasta_tools/tron_tools.py
```diff
@@ -442,6 +442,7 @@ def get_env(
         env = super().get_env(system_paasta_config=system_paasta_config)
 
         if self.get_executor() == "spark":
+            env["AWS_DEFAULT_REGION"] = DEFAULT_AWS_REGION
```
nemacysts, line 445: a comment about why this necessary would be useful for future yelpers :)

nemacysts (commented):
(it would also be useful for folks git blaming in their IDEs/editors if this commit had a description)

chi-yelp (approved):

PR #3861, opened 2024-05-10
Remove autoscale_all_services.py (which was for mesos/marathon), and code that was only called by that.

EvanKrall:


nemacysts (approved):

EvanKrall:
obsoleted by  #3869

PR #3862, opened 2024-05-10
Improve paasta rollback messaging

nemacysts:
Printing out the --help output when passing a short SHA is not helpful - and neither is not very loudly mentioning that changes *need* to be reverted in Git for a rollback to become permanent

### DIFF
--- a/paasta_tools/cli/utils.py
+++ b/paasta_tools/cli/utils.py
@@ -968,11 +968,11 @@ def validate_given_deploy_groups(
 
 
 def short_to_full_git_sha(short, refs):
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/utils.py
nemacysts, line 1: this is mostly me just wanting to be consistent with git/Git and sha/SHA :p

jimmingcheng (approved):

bobokun (approved):

EmanElsaban (approved):

PR #3864, opened 2024-05-13
minor docs improvements and updates

drmorr0:
1) Update the yelpsoa_configs.html autoscaling section
2) remove the DEFAULT_CPU_AUTOSCALING_MOVING_AVERAGE_WINDOW because we aren't using Prometheus for CPU scaling anymore (https://github.com/Yelp/paasta/pull/3838)

cc @sclg-yelp


docs/source/yelpsoa_configs.rst
```diff
@@ -386,18 +386,31 @@ instance MAY have:
...
+          and the service's utilization (as measured by your ``metrics_provider``) is above this value, you'll get an alert.
+          The default is the same as your ``setpoint``.
+
+        * ``moving_average_window_seconds``: A smoothing function to apply to the data received from your metrics
+          provider.
+
+        * ``prometheus_adapter_config``: **(advanced users only)** Custom prometheus configuration for the
+          ``arbitrary_promql`` metrics provider.
+
+    * ``scaledown_policies``: Custom configuration for the Kubernetes HPA controlling when the service will scale down
```
nemacysts, line 413: we could add a note here that the values for this look exactly like the documented HPA options at https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#configurable-scaling-behavior

(e.g., 
``` YAML
behavior:
  scaleDown:
    ...
```

the `...` would be exactly the same as what you'd write in a raw k8s manifest)

...but wording that better than i've done so here might require more effort than what you've already helpfully expended here so fine to skip imo :)

nemacysts (approved):

docs/source/yelpsoa_configs.rst
drmorr0: @nemacysts updated, WDYT?

drmorr0: wait this is not markdown FML

drmorr0: ok fixed now wdyt

nemacysts (approved):

dbgrigsby:
Woo congrats on v1.0.0!

PR #3865, opened 2024-05-13
Use vanilla pip

ymilki:
## Problem

`venv-update` is limited to `pip<=18.1` which does not support manylinux wheels. In order to continue supporting this repo, we need to update pip.

## Solution

* Drop `venv-update`
* Update bootstrap dependencies in Makefile and GHA
* Minor fixes for newer tox and venv behaviour

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -144,7 +144,7 @@ deps =
 commands =
     # TODO: upgrade behave if they ever take this reasonable PR
     pip install git+https://github.com/Yelp/behave@1.2.5-issue_533-fork
-    pylint -E paasta_tools/mesos/ --ignore master.py,task.py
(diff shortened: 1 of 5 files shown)


nemacysts (approved):

tox.ini
```diff
@@ -184,3 +186,6 @@ extend-ignore = E501,E203,W503
 
 [pep8]
 ignore = E265,E501
+
+[testenv:.tox]
+envdir = {toxworkdir}/{envname}
```
benbariteau, line 191: The top-level testev doesn't set envdir, so this is not required.

benbariteau (changes requested):

tox.ini
ymilki: reverted this change

benbariteau (approved):

PR #3867, opened 2024-05-14
Enable new vector-logs driver for paasta logs

yaroliakh:
Adding new "vector-logs" driver that implements `LogReader` `print_logs_by_time` method. 
This change requires changing paasta logs.json config to include `log_readers` containing at least one log_reader config with `components`, as shown below:
```
"log_readers": [
    {
      "driver": "scribereader",
      "components": ["deploy", "monitoring"],
      "options": {}
    },
    {
      "driver": "vector-logs",
      "components": ["stdout", "stderr"],
      "options": {}
      }
    }
  ]
```
The reason we need two readers (drivers) is to be able to read logs from two logging pipelines, as for now only `stdout` and `stderr` will be migrated to the new pipeline. This also means that the tooling won't be supporting mixing outputs for multiple components from different pipelines, e.g. "build" and "stdout".


yelp_package/extra_requirements_yelp.txt
```diff
@@ -23,18 +22,18 @@ pyopenssl==19.0.0
...
 slo-transcoder==3.3.0
 smmap2==2.0.3
-splunk-sdk==1.7.0
 srv-configs==1.1.0
 sticht[yelp_internal]==1.1.18
+tenacity==8.2.3
 vault-tools==0.9.2
 yelp-cgeom==1.3.1
-yelp-clog==5.2.3
+yelp-clog==7.1.1
```
jvperrin, line 34: Hopefully this is fine and doesn't impact anything else already using clog? I do know there's a wiki page named "Upgrading to yelp-clog version 7+" so that might be good to take a look at if you haven't seen it already about if there's anything else to modify as part of this upgrade

yaroliakh: overlooked that. thanks

yaroliakh: opened a separate PR to upgrade clog: https://github.com/Yelp/paasta/pull/3868

paasta_tools/cli/cmds/logs.py
```diff
@@ -291,7 +297,7 @@ def paasta_log_line_passes_filter(
     if not check_timestamp_in_range(timestamp, start_time, end_time):
         return False
     return (
-        parsed_line.get("level") in levels
```
nemacysts, line 294: question: is there any context behind this behavior change? this will now print messages that don't have a level field, right? (whereas they were previously filtered out)

paasta_tools/cli/cmds/logs.py
```diff
@@ -503,18 +509,46 @@ def outer(log_reader_class):
...
+def get_log_reader(components: Set[str]) -> "LogReader":
+    log_readers_config = load_system_paasta_config().get_log_readers()
+    # ideally we should use a single "driver" for all components, but in cases where more than one is used for different components,
+    # we should use the first one that supports all requested components
+    # otherwise signal an error and suggest to provide a different list of components
+    components_lists = []
+    for log_reader_config in log_readers_config:
+        supported_components = log_reader_config.get("components", [])
+        components_lists.append(supported_components)
+        if components <= set(supported_components):
```
nemacysts, line 531: imo, `issubset` is a little quicker to parse as a human (i had to look at the docs to remind myself what `<=` means for sets) - but i don't feel too strongly about this :)

paasta_tools/cli/cmds/logs.py
```diff
@@ -1182,6 +1216,75 @@ def cluster_to_scribe_env(self, cluster: str) -> str:
...
+
+    def __init__(self, cluster_map: Mapping[str, Any]) -> None:
+        super().__init__()
+
+        if S3LogsReader is None:
+            raise Exception("yelp_clog package must be available to use S3LogsReader")
+
+        self.cluster_map = cluster_map
+
+    def get_ecosystem_for_cluster(self, cluster: str) -> str:
```
nemacysts, line 1231:
```suggestion
    def get_ecosystem_for_cluster(self, cluster: str) -> Optional[str]:
```

nemacysts: (surprised that mypy didn't complain about this)

paasta_tools/cli/cmds/logs.py
```diff
@@ -1182,6 +1216,75 @@ def cluster_to_scribe_env(self, cluster: str) -> str:
...
+        return self.cluster_map.get(cluster, None)
+
+    def print_logs_by_time(
```
nemacysts, line 1234: nit: can we add some types to this function?

paasta_tools/utils.py
```diff
@@ -2367,6 +2369,19 @@ def get_log_reader(self) -> LogReaderConfig:
                 % self.directory
             )
 
+    def get_log_readers(self) -> List[LogReaderConfig]:
```
nemacysts, line 2372: just verifying: this will eventually be deleted in favor of having `get_log_reader()` return the vector reader?

paasta_tools/cli/cmds/logs.py
yaroliakh: in the current pipeline, all service logs have "debug" `level` added by the forwarder, which makes  me wonder if there is any usefulness of having it as part of the log event at all?
We haven't added "level" field in the new pipeline, therefore updating this filter condition to either accept new pipeline logs (without level field) or filter other logs if they have a different than default levels, which are defined here: https://github.com/Yelp/paasta/blob/9ed832dc727d46b649d3b7d0153e44a0c7e952bc/paasta_tools/cli/cmds/logs.py#L1422

paasta_tools/utils.py
yaroliakh: if we migrate all logs to a single pipeline, then yes. Until then we would use `get_log_readers()` i think

paasta_tools/cli/cmds/logs.py
yaroliakh: agree

yaroliakh:
@nemacysts, while preparing paasta configs update I've realized that we are missing per env rollout mechanism with new reader. We still might need to include some feature flag here

paasta_tools/utils.py
```diff
@@ -2319,6 +2321,25 @@ def get_log_reader(self) -> LogReaderConfig:
...
+            raise PaastaNotConfiguredError(
+                "Could not find log_readers in configuration directory: %s"
+                % self.directory
+            )
+
+    def get_log_readers_migration_status(self) -> Optional[Mapping[str, bool]]:
+        """
+        Get the log readers migration configuration out of global paasta config
+        """
+        return self.config_dict.get("log_readers_migration_status")
```
yaroliakh, line 2341: this would need a config item, like:
```
"log_readers_migration_status": {
    "cluster-1-dev": true,
    "cluster-2-prod": false
  },
```

yaroliakh: @nemacysts wdyt about having this while we are gradually migrating to the new pipeline?

nemacysts: i think something like `use_vector_logs` as a config name would be easier to reason about in the code/as a human - but in general seems like a good idea :)

yaroliakh: renamed a bit, to be less yelpy ))

nemacysts (commented):
i have one last question about how to handle the feature flags when multiple clusters are given to paasta logs, but lgtm otherwise

paasta_tools/cli/cmds/logs.py
```diff
@@ -1362,16 +1478,16 @@ def paasta_logs(args: argparse.Namespace) -> int:
 
     levels = [DEFAULT_LOGLEVEL, "debug"]
 
-    log.debug(f"Going to get logs for {service} on cluster {cluster}")
+    log.debug(f"Going to get logs for {service} on cluster {clusters}")
 
-    log_reader = get_log_reader()
```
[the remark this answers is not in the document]
yaroliakh: with the current state only a single cluster is supported, if i get it right, which is enforced by `nargs=1` (won't work with eg `-c cluster1 cluster2` parameter ) at https://github.com/Yelp/paasta/blob/56fac3eaacf701a171ec1f39ddaedf99103ddb6f/paasta_tools/cli/cmds/logs.py#L90-L94

and implicitly mentioned here: https://github.com/Yelp/paasta/blob/56fac3eaacf701a171ec1f39ddaedf99103ddb6f/paasta_tools/cli/cmds/logs.py#L1336-L1339 

This make things much easier tbh, otherwise with multiple clusters we'd need to implement multithreading in vector-logs driver to read logs from two backends in parallel, similar to what scribe has implemented in https://github.com/Yelp/paasta/blob/56fac3eaacf701a171ec1f39ddaedf99103ddb6f/paasta_tools/cli/cmds/logs.py#L599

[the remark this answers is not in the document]
nemacysts: ah, gotcha - I was led astray by the `cluster = args.cluster` -> `cluster` = args.cluster` rename :)

that said, it does say the default is all clusters, so i wonder how that works right now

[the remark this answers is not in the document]
nemacysts: ah, i think that help text is outdated :p

nemacysts (approved):
we should probably update the --cluster help-text, but that behavior hasn't changed with your PR so we can follow-up with that later :)

PR #3868, opened 2024-05-15
Bump yelp-clog and scribereader

yaroliakh:
Based on the 7+ version changelog:
> Deprecated support for logging bytes lines, emitting a [DeprecationWarning](https://docs.python.org/3/library/exceptions.html#DeprecationWarning) if bytes lines are logged.
A future version of yelp-clog will only support str lines. 
Removed support for the scribe protocol
Removed ScribeLogger, ScribeHandler, and ScribeMonkLogger classes.
Only Monk is supported as a backend for clog.
Removed the scribe readers
Removed NetCLogStreamReader and all other classes under the clog.readers module.
These classes are now available in the scribereader package.

none of the above seems to be used. 

Removing the only reference I could find to deprecated `clog.readers`. Everything else should be already using Monk backend and should work without changes.

Smoke test:
```python
➜  paasta git:(u/yaro/OBSPLAT-1830_clog_upgrade) ✗
 -> python
Python 3.8.13 (default, Apr 19 2022, 02:32:06)
[GCC 11.2.0] on linux
>>> from paasta_tools.utils import _log
>>> _log(service='statsite', line='test log 123', component='monitoring', level='debug', cluster='pnw-devc', instance='main')
[service statsite] test log 123
```

### DIFF
--- a/yelp_package/extra_requirements_yelp.txt
+++ b/yelp_package/extra_requirements_yelp.txt
@@ -1,40 +1,39 @@
-atomicfile==1.0
-cached-property==1.3.1
-cffi==1.15.0
-cffi==1.15.0
+atomicfile==1.0                     # vault-tools dependency
+cached-property==1.3.1              # slo-transcoder dependency
+cffi==1.15.0                        # vault-tools dependency
 clusterman_metrics==2.2.0
-crypto-lib==4.0.0
-cryptography==3.0
-dateglob==0.3
-geogrid==1.3.5
-gitdb2==2.0.3
-gitpython==2.1.9
-hvac==0.11.0
-inflection==0.3.1
-iso8601==0.1.14
-markdown==2.4
-monk==1.3.0
-mrjob==0.7.4
-named-decorator==0.1.4
-ndg-httpsclient==0.4.3
-pycparser==2.20
-pygpgme==0.3
-pyopenssl==19.0.0
-PySubnetTree==0.34
-python-jsonschema-objects==0.3.1
-retrying==1.3.3
+crypto-lib==4.0.0                   # vault-tools dependency
+cryptography==3.0                   # vault-tools dependency
+dateglob==0.3                       # scribereader dependency
+geogrid==1.3.5                      # scribereader dependency
+gitdb2==2.0.3                       # slo-transcoder dependency
+gitpython==2.1.9                    # slo-transcoder, vault-tools dependency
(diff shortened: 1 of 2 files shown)


gkousouris (approved):
lgtm!

nemacysts:
@yaroliakh did you also verify that `paasta logs` still works?

yelp_package/extra_requirements_yelp.txt
```diff
@@ -32,7 +32,7 @@ srv-configs==1.1.0
 sticht[yelp_internal]==1.1.18
 vault-tools==0.9.2
 yelp-cgeom==1.3.1
-yelp-clog==5.2.3
+yelp-clog==7.1.1
```
nemacysts, line 35: i think we'll want to add `tenacity` as well since that's now required by yelp-clog

(figuring this out was kinda annoying: i added the direct dependencies (i.e., things we import) into requirements-minimal.txt and copied the contents here to requirements.txt so that i could run `check-requirements`)

nemacysts: there's also some lines we can delete: https://fluffy.yelpcorp.com/i/wmSXXqXBvW1THmZPs5mfNGfF2cR9l69g.html is what i think we'd want as a final state

yaroliakh: So I've pinned deps and ran `check-requirements` for the following main dependencies:
```
yelp-clog
scribereader
clusterman-metrics
vault-tools
slo-transcoder
yelp-profiling
sticht[yelp_internal]==1.1.18
``` 
Not sure about deleting other deps, as I can see most of them being referenced in code

yaroliakh:
> @yaroliakh did you also verify that `paasta logs` still works?

seems to be working, as `paasta logs`  is using `scribereader` interfaces to fetch logs, which hasn't changed with newer version
https://fluffy.yelpcorp.com/i/N59Cl7k0T3sCqmwjXxGPPjkHCmSwWtnh.html

yelp_package/extra_requirements_yelp.txt
nemacysts: oh good catch - i somehow missed grepping for clusterman-metrics (which is maybe why my check-requirements said we could delete some deps)

nemacysts: also: big thank you for annotating these dependencies :)

nemacysts (approved):

PR #3869, opened 2024-05-15
Farewell Mesos

cuza:
Pretty self explanatory


paasta_tools/paasta_remote_run.py
nemacysts, line 1: i'm 90% sure that this no longer works post-mesos and can probably be removed

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -58,7 +59,6 @@
 from paasta_tools.long_running_service_tools import METRICS_PROVIDER_PISCINA
 from paasta_tools.long_running_service_tools import METRICS_PROVIDER_PROMQL
 from paasta_tools.long_running_service_tools import METRICS_PROVIDER_UWSGI
-from paasta_tools.long_running_service_tools import MetricsProviderDict
```
nemacysts, line 61: should the other autoscaling related imports from long_running_service_tools also be moved?

paasta_tools/utils.py
```diff
@@ -246,7 +245,7 @@ class InvalidInstanceConfig(Exception):
 
 Constraint = Sequence[str]
 
-# e.g. ['GROUP_BY', 'habitat', 2]. Marathon doesn't like that so we'll convert to Constraint later.
+# e.g. ['GROUP_BY', 'habitat', 2]. Tron doesn't like that so we'll convert to Constraint later.
 UnstringifiedConstraint = Sequence[Union[str, int, float]]
```
nemacysts, line 249: constraints are a mesos concept, so this should be deleteable

paasta_tools/utils.py
```diff
@@ -246,7 +245,7 @@ class InvalidInstanceConfig(Exception):
...
 
 SecurityConfigDict = Dict  # Todo: define me.
```
nemacysts, line 251: i bet this is only used for the firewall code, so if we delete that this can probably go too

paasta_tools/utils.py
```diff
@@ -1889,12 +1854,6 @@ class ResourcePoolSettings(TypedDict):
 PoolToResourcePoolSettingsDict = Dict[str, ResourcePoolSettings]
```
nemacysts, line 1854: this and the autoscaler stuff above also sound like old paasta autoscaler code that can be deleted

nemacysts: there's also some stuff below with Slave and whatnot in the name that is mesos-only

paasta_tools/utils.py
```diff
@@ -2006,7 +1965,6 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     log_reader: LogReaderConfig
     log_writer: LogWriterConfig
     maintenance_resource_reservation_enabled: bool
```
nemacysts, line 1967: this is mesos code as well

nemacysts: (and there's some other attributes with mesos in the name here)

paasta_tools/utils.py
```diff
@@ -3892,9 +3831,7 @@ def is_under_replicated(
 def deploy_blacklist_to_constraints(
```
nemacysts, line 3831: unless we have code that translates constraints into selectors/affinities, these are likely dead code

requirements.txt
nemacysts, line 1: i'm surprised that removing the marathon python library as a top-level dependency only removed 1 transitive dependencies :p

setup.py
```diff
@@ -63,14 +62,10 @@ def get_install_requires():
...
         "paasta_tools/kubernetes/bin/paasta_secrets_sync.py",
-        "paasta_tools/list_marathon_service_instances.py",
         "paasta_tools/log_task_lifecycle_events.py",
-        "paasta_tools/marathon_dashboard.py",
         "paasta_tools/monitoring/check_capacity.py",
-        "paasta_tools/monitoring/check_marathon_has_apps.py",
         "paasta_tools/monitoring/check_mesos_active_frameworks.py",
         "paasta_tools/monitoring/check_mesos_duplicate_frameworks.py",
-        "paasta_tools/monitoring/check_mesos_outdated_tasks.py",
         "paasta_tools/monitoring/check_mesos_quorum.py",
```
nemacysts, line 69: probably also deleteable :)

EvanKrall: Yeah I've got other PRs up for deleting a bunch of things we noticed while working on the merge conflicts for this one.

Didn't want to make this PR any bigger than it needed to be.

nemacysts (approved):
(i think all my comments can be handled in follow-up prs fwiw)

EvanKrall (approved):
let's ship this carefully obviously

PR #3875, opened 2024-05-17
Add noble package build

ymilki:
## Feature

Let's build for noble!

## Notes

Noble doesn't ship with python3.8 and we need to grab it from deadsnakes ppa. We'll want to update paasta to python3.10 soon...

### DIFF
--- a/yelp_package/dockerfiles/noble/mesos-slave-secret
+++ b/yelp_package/dockerfiles/noble/mesos-slave-secret
@@ -1,0 +1,4 @@
+{
+  "principal": "slave",
+  "secret": "secret1"
+}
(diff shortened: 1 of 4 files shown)


benbariteau (approved):

yelp_package/dockerfiles/noble/mesos-slave-secret
nemacysts, line 1: i thought we had deleted the other mesos-slave-secret files during our unhack, but i guess not

nemacysts:
note: i think you'll also need to update the internal jenkinsfile for this repo (in misc/jenkinsfiles)

PR #3876, opened 2024-05-17
Delete even more mesos code

EvanKrall:
This one isn't pure deletion, I moved a few things around.


general_itests/steps/deployments_json_steps.py
nemacysts, line 1: is this changing cause further in this pr we're switching to using just deploymentsv2v

if so,are we ready to use deploymentsv2? i vaguely remember there being some issues last time we tried to switch to using v2 for everything

paasta_tools/cli/cmds/local_run.py
```diff
@@ -79,6 +79,28 @@ class AWSSessionCreds(TypedDict):
     AWS_SECURITY_TOKEN: str
 
 
+def execute_in_container(docker_client, container_id, cmd, timeout):
```
nemacysts, line 82: i assume this is a move, but is it me or is `timeout` unused?

nemacysts: if we clean up the unused param, we can probably also
type this 
```suggestion
def execute_in_container(docker_client: docker.Client, container_id: str, cmd: str) -> Tuple[str, int]:
```

paasta_tools/cli/schemas/kubernetes_schema.json
nemacysts, line 1: just double-checking: there's no files in soaconfigs referencing these dead options?

paasta_tools/contrib/get_running_task_allocation.py
nemacysts, line 1: !8ball pull https://github.com/Yelp/paasta/pull/3772/files into this branch

paasta_tools/contrib/get_running_task_allocation.py
```diff
@@ -40,81 +36,9 @@ class TaskAllocationInfo(NamedTuple):
     host_ip: str
     git_sha: str
     config_sha: str
-    mesos_container_id: str  # Because Mesos task info does not have docker id
```
nemacysts, line 43: might want to double-check if any of the splunk queries that we have will break if this field isn't present

nemacysts: (it's probably safe since it's been null for ages, but you never know :p)

paasta_tools/firewall.py
nemacysts, line 1: with the tickets jfong found, i'm now like 99% certain we can cleanup soaconfigs and then delete this code entirely since it's not actually doing anything

paasta_tools/tron_tools.py
nemacysts, line 1: once we release the de-meso'd tron, we can probably also delete all the remaining mesos code in setup_tron_namespace :p

nemacysts (approved):

general_itests/steps/deployments_json_steps.py
EvanKrall: Vulture seems to think nothing is using deployments v1 (at least the mypy types). I'm not sure I believe that nothing is using the v1 part of the json, but I'm leaving the file format alone anyway.

PR #3878, opened 2024-05-21
Pin request in k8s_itests

yaroliakh:
Current `master` pipeline is broken due to a recent release of `requests`, which seems to be incompatible with python lib for `docker`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -88,6 +88,7 @@ whitelist_externals = bash
 # one day we'll use a fully pinned venv here...
 deps =
     urllib3<2.0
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -88,6 +88,7 @@ whitelist_externals = bash
 # one day we'll use a fully pinned venv here...
 deps =
     urllib3<2.0
+    requests<2.32.0	# tmp fix for broken python docker: https://github.com/psf/requests/issues/6707
```
nemacysts, line 91: i think we likely also want to pin requests in requirements-minimal.txt since i think there's other scripts that use dockerpy

nemacysts (approved):

PR #3883, opened 2024-05-28
paasta validate: allow duplicate instance names for kubernetes<->eks and nrtsearchservice(eks)

EvanKrall:
Currently we're routinely bypassing validation when doing migrations to EKS, which means that anybody else making yelpsoa-configs changes in the same directory in the meantime would have paasta validate fail.

This would keep validate from failing specifically for the pairs of instance types that we migrate to/from, while still making sure that we don't e.g. have a tron instance named the same as an eks instance.


jfongatyelp (commented):
I'm not all that convinced we want this; I do think we want to prevent people from accidentally creating duplicate-instance names between the two files, just not in the case of actually performing the migrations.

This improvement to `paasta validate` doesn't really help us change the migration behavior anyhow, since not only need to skip this validation, but also the PR approval step currently, otherwise it would require more than one person, multiple times, for each migration, which I think is unnecessary. 

If this makes nrtsearch migration vastly easier tho, seems like a fine option; I'd just recommend a few changes in that case: 
1. limiting the `duplicates_allowed` to the nrtsearch types
2. adding a TODO to link it to our cleanup ticket (COMPINFRA-3103) to make sure this is removed after migrations are finished.

jfongatyelp (approved):

nemacysts:
we decided to go another way with the nrtsearch migration, so we decided that we won't need this

PR #3884, opened 2024-05-28
Add support for VitessCluster CRs on EKS

VinaySagarGonabavi:
vitess-operator is migrated already, this is to prepare for the move of vitess-k8s service
https://jira.yelpcorp.com/browse/DREIMP-10779


VinaySagarGonabavi:
Not needed since we moved the setup to EKS with existing fileprefix

PR #3888, opened 2024-06-05
automatically mount service account tokens when needed

piax93:
I thought I would need to jump through many hoops, but this simple of a change should work.

* For new service instances, the list of volume gets updated dynamically on top of what is loaded from the config, so everything easy peasy there.
* For existing instances, I saw that the configuration hash is calculated on the whole pod spec, minus a couple of fields, but definitely including volumes, so if a service which was not required to do auth, and some point becomes required, the hash should mismatch and a new pod get deployed.

The only thing I'm not crazy about is hardcoding that "authenticating.yaml". Maybe I could place that in a field of the system config? Although that may be a way to add a step without much of a benefit. Very open to suggestions here.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2554,17 +2554,11 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         )
 
     def get_projected_sa_volumes(self) -> List[ProjectedSAVolume]:
-        config_volumes = super().get_projected_sa_volumes()
-        token_config = (
-            load_system_paasta_config().get_service_auth_token_volume_config()
+        return add_volumes_for_authenticating_services(
+            service_name=self.service,
+            config_volumes=super().get_projected_sa_volumes(),
+            soa_dir=self.soa_dir,
         )
-        if (
-            self.service in get_authenticating_services(self.soa_dir)
-            and token_config
-            and not any(volume == token_config for volume in config_volumes)
-        ):
-            config_volumes = [token_config, *config_volumes]
-        return config_volumes
 
 
 def get_kubernetes_secret_hashes(
@@ -4428,3 +4422,27 @@ def get_authenticating_services(soa_dir: str = DEFAULT_SOA_DIR) -> Set[str]:
     authenticating_services_conf_path = os.path.join(soa_dir, "authenticating.yaml")
     config = service_configuration_lib.read_yaml_file(authenticating_services_conf_path)
     return set(config.get("services", []))
+
+
+def add_volumes_for_authenticating_services(
+    service_name: str,
+    config_volumes: List[ProjectedSAVolume],
+    soa_dir: str = DEFAULT_SOA_DIR,
+) -> List[ProjectedSAVolume]:
+    """Add projected service account volume to the list of volumes if service
+    participates in authenticated traffic. In case of changes, a new list is returned,
+    no updates in-place.
+
+    :param str service_name: name of the service
+    :param List[ProjectedSAVolume] config_volumes: existing projected volumes from service config
+    :param str soa_dir: path to SOA configurations directory
+    :return: updated list of projected service account volumes
+    """
+    token_config = load_system_paasta_config().get_service_auth_token_volume_config()
+    if (
+        token_config
+        and service_name in get_authenticating_services(soa_dir)
+        and not any(volume == token_config for volume in config_volumes)
+    ):
+        config_volumes = [token_config, *config_volumes]
+    return config_volumes
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -21,6 +21,7 @@ import os
 import re
 from datetime import datetime
 from enum import Enum
+from functools import lru_cache
 from inspect import currentframe
 from pathlib import Path
 from typing import Any
@@ -2552,6 +2553,19 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             "topology_spread_constraints", default_pod_topology_spread_constraints
         )
 
+    def get_projected_sa_volumes(self) -> List[ProjectedSAVolume]:
+        config_volumes = super().get_projected_sa_volumes()
+        token_config = (
+            load_system_paasta_config().get_service_auth_token_volume_config()
+        )
+        if (
+            self.service in get_authenticating_services(self.soa_dir)
+            and token_config
+            and not any(volume == token_config for volume in config_volumes)
+        ):
+            config_volumes = [token_config, *config_volumes]
+        return config_volumes
+
 
 def get_kubernetes_secret_hashes(
     environment_variables: Mapping[str, str], service: str, namespace: str
@@ -4406,3 +4420,11 @@ def get_kubernetes_secret_volumes(
                 ] = secret_contents
 
     return secret_volumes
+
+
+@lru_cache()
+def get_authenticating_services(soa_dir: str = DEFAULT_SOA_DIR) -> Set[str]:
+    """Load list of services partecipating in authenticated traffic"""
+    authenticating_services_conf_path = os.path.join(soa_dir, "authenticating.yaml")
+    config = service_configuration_lib.read_yaml_file(authenticating_services_conf_path)
+    return set(config.get("services", []))
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -4424,7 +4424,7 @@ def get_kubernetes_secret_volumes(
 
 @lru_cache()
 def get_authenticating_services(soa_dir: str = DEFAULT_SOA_DIR) -> Set[str]:
-    """Load list of services partecipating in authenticated traffic"""
+    """Load list of services participating in authenticated traffic"""
     authenticating_services_conf_path = os.path.join(soa_dir, "authenticating.yaml")
     config = service_configuration_lib.read_yaml_file(authenticating_services_conf_path)
     return set(config.get("services", []))
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2017,6 +2017,7 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     sidecar_requirements_config: Dict[str, KubeContainerResourceRequest]
     eks_cluster_aliases: Dict[str, str]
     secret_sync_delay_seconds: float
+    service_auth_token_settings: ProjectedSAVolume
 
 
 def load_system_paasta_config(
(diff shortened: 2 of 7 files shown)


danielpops:
> For existing instances, I saw that the configuration hash is calculated on the whole pod spec, minus a couple of fields, but definitely including volumes, so if a service which was not required to do auth, and some point becomes required, the hash should mismatch and a new pod get deployed.

One thought that comes to mind is that now a change in ServiceA's yelpsoa-configs can cause a bounce for ServiceB, which I think is a necessary evil for this, but it's worth thinking about how we can make this clear to users since it likely wouldn't be intuitive (maybe jenkins posts this detail as a PR comment or something). I also could envision eventually unconditionally mounting the token for all services, regardless of whether they currently need to authenticate or not, which would alleviate this problem but could have other downsides (e.g. additional load on api server).

tests/test_kubernetes_tools.py
```diff
@@ -2773,6 +2781,25 @@ def test_get_persistent_volume_name(self):
...
+    @mock.patch(
+        "paasta_tools.kubernetes_tools.load_system_paasta_config", autospec=None
+    )
+    @mock.patch(
+        "paasta_tools.kubernetes_tools.get_authenticating_services", autospec=None
+    )
+    def test_get_projected_sa_volumes_token_automount(
+        self, mock_get_auth_services, mock_system_config
+    ):
+        mock_get_auth_services.return_value = {"service_a", "service_b", "kurupt"}
```
danielpops, line 2793: it's a little hard to follow why the test has these three services that are authenticating but the assertion is for only a single projected sa volume. Can you clarify the test setup / expectations with comments?

danielpops (approved):

danielpops:
> The only thing I'm not crazy about is hardcoding that "authenticating.yaml". Maybe I could place that in a field of the system config? Although that may be a way to add a step without much of a benefit. Very open to suggestions here.

No strong opinions here. Right now afaict this value is only in this repo and in puppet, which isn't too widespread and probably fine to be hard coded.

nemacysts:
> One thought that comes to mind is that now a change in ServiceA's yelpsoa-configs can cause a bounce for ServiceB, which I think is a necessary evil for this, but it's worth thinking about how we can make this clear to users since it likely wouldn't be intuitive (maybe jenkins posts this detail as a PR comment or something).

@danielpops i think as long as we're not bouncing every service when this happens, this isn't the end of the world: part of the paasta contract is that a paasta instance can be restarted at any time (see [here](https://paasta.readthedocs.io/en/latest/about/contract.html#bouncing): `PaaSTA reserves the right to cause a bounce (our term for when all of your service’s containers are replaced) on your service at any time.`

paasta_tools/kubernetes_tools.py
```diff
@@ -2552,6 +2553,19 @@ def get_topology_spread_constraints(
...
+    def get_projected_sa_volumes(self) -> List[ProjectedSAVolume]:
+        config_volumes = super().get_projected_sa_volumes()
+        token_config = (
+            load_system_paasta_config().get_service_auth_token_volume_config()
+        )
+        if (
+            self.service in get_authenticating_services(self.soa_dir)
+            and token_config
+            and not any(volume == token_config for volume in config_volumes)
+        ):
```
nemacysts, line 2565: when can this happen?

nemacysts:
@danielpops @piax93 i guess i should confirm: what's the worst case scenario for how many services would be bounced when updating the auth rules for 1 service? could we end up bouncing everything should there be a change that involves internalapi?

danielpopskandor:
@nemacysts a bounce will only happen when a service that previously was not referenced in any other service's authorization.yaml is added OR the opposite, a service previously in any other service's authorization.yaml has _all_ references removed. 

### Examples

## Service starts being referenced

security_happyhour/authorization.yaml starting state:
```
authorization:
  enabled: true
  rules:
  - name: service-level granularity
    description: Only security and itself can say hello to this service
    identity_groups:
      services:
        - yelp_main
```

anti_phishing is added to a rule:
```
authorization:
  enabled: true
  rules:
  - name: service-level granularity
    description: Only security and itself can say hello to this service
    identity_groups:
      services:
        - yelp_main
        - anti_phishing
```

When this ships it will cause anti_phishing to bounce

## Service no longer referenced

security_happyhour/authorization.yaml starting state:
```
authorization:
  enabled: true
  rules:
  - name: service-level granularity
    description: Only security and itself can say hello to this service
    identity_groups:
      services:
        - yelp_main
        - anti_phishing
```

yelp_main removed from authorization.yaml file and not referenced in any other authorization.yaml files:
```
authorization:
  enabled: true
  rules:
  - name: service-level granularity
    description: Only security and itself can say hello to this service
    identity_groups:
      services:
        - anti_phishing
```

When this ships it will cause yelp_main to bounce

nemacysts:
@danielpops thanks for the example - that makes things a lot clearer. i think in steady-state operation this probably wouldn't be too bad, but while we're rolling out this might cause a lot of churn (and some services like y-m are a bit noisy when restarted - which is made worse by the fact that it has many callers so it would presumably be in many authorization.yamls)

nemacysts:
(i've started a CI thread to see what folks think here)

danielpopskandor:
@nemacysts if yelp-main is in 10 authorization.yaml's and removed from one of them, a bounce wouldn't happen. The bounce would only happen when yelp-main is removed from _all_ of them, since it would no longer have the projected token volume mount causing a bounce. so from that perspective, the removal is a much less likely scenario that we'd encounter. 

The main problem is that the unsuspecting PR author and reviewer probably wouldn't realize that adding `yelp-main` to service-a's authorization.yaml _could_ cause a bounce for `yelp-main`

paasta_tools/kubernetes_tools.py
piax93: Do you mean the `any(...)` check? Just making sure we are not trying to mount 2 equal volumes if someone manually added the SA volume setting in a service instance config.

piax93:
@nemacysts @danielpops I think one easy solution for the issue is to keep a separate list of critical services which will always have the token mounted, so we can be sure they are not affect by someone referencing/removing them from an authorization config.

danielpops:
@nemacysts a bounce will only happen when a service that previously was not referenced in any other service's authorization.yaml is added OR the opposite, a service previously in any other service's authorization.yaml has _all_ references removed. 

### Examples

## Service starts being referenced

security_happyhour/authorization.yaml starting state:
```
authorization:
  enabled: true
  rules:
  - name: service-level granularity
    description: The description
    identity_groups:
      services:
        - yelp_main
```

anti_phishing is added to a rule:
```
authorization:
  enabled: true
  rules:
  - name: service-level granularity
    description: The description
    identity_groups:
      services:
        - yelp_main
        - anti_phishing
```

When this ships it will cause anti_phishing to bounce

## Service no longer referenced

security_happyhour/authorization.yaml starting state:
```
authorization:
  enabled: true
  rules:
  - name: service-level granularity
    description: The description
    identity_groups:
      services:
        - yelp_main
        - anti_phishing
```

yelp_main removed from authorization.yaml file and not referenced in any other authorization.yaml files:
```
authorization:
  enabled: true
  rules:
  - name: service-level granularity
    description: The description
    identity_groups:
      services:
        - anti_phishing
```

When this ships it will cause yelp_main to bounce

danielpopskandor (approved):

paasta_tools/kubernetes_tools.py
nemacysts: @piax93 oh, i just wasn't using all my neurons - i was trying to figure out how someone could write an authn.yaml that contains the token config... but authn.yaml and the projected_sa_volume config are two different things!

nemacysts (approved):

paasta_tools/kubernetes_tools.py
nemacysts: (i.e., you don't need to explicitly create a `projected_sa_volume` config for the authn stuff to work)

paasta_tools/kubernetes_tools.py
```diff
@@ -4406,3 +4414,35 @@ def get_kubernetes_secret_volumes(
...
+    :param str service_name: name of the service
+    :param List[ProjectedSAVolume] config_volumes: existing projected volumes from service config
+    :param str soa_dir: path to SOA configurations directory
+    :return: updated list of projected service account volumes
+    """
+    token_config = load_system_paasta_config().get_service_auth_token_volume_config()
+    if (
+        token_config
+        and service_name in get_authenticating_services(soa_dir)
+        and not any(volume == token_config for volume in config_volumes)
```
nemacysts, line 4445: i think it might be worth doing a more generic dedupe in the future - presumably having user-supplied duplicates is also something we'd like to avoid on top of preventing folks from colliding with the auto-mounted token-config volume
 
i.e., we probably wanna figure out a central place to dedupe all volume configs and handle all of them consistently :p

https://github.com/Yelp/paasta/blob/7e866929ff6748594dd96c1c6b98889e16160a6b/paasta_tools/kubernetes_tools.py#L1482 has us dedupe a couple different types of volume mounts, but not all of them ;_;

nemacysts (approved):

jfongatyelp (approved):
Appreciate that we'll be able to control this set of always-authenticating services in puppet/system configs. hopefully there aren't too many services that can't deal with 'surprise' bounces well.

paasta_tools/utils.py
```diff
@@ -2703,6 +2704,9 @@ def get_spark_kubeconfig(self) -> str:
     def get_kube_clusters(self) -> Dict:
         return self.config_dict.get("kube_clusters", {})
 
+    def get_service_auth_token_volume_config(self) -> ProjectedSAVolume:
+        return self.config_dict.get("service_auth_token_settings", {})
```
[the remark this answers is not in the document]
piax93: lol, apologies, I'm a terrible human being. I created this config first thinking that each JSON file is loaded in a different "namespace", but then realized that everything gets flattened out, so started adding words over words to the key.

paasta_tools/kubernetes_tools.py
```diff
@@ -2552,6 +2553,19 @@ def get_topology_spread_constraints(
             "topology_spread_constraints", default_pod_topology_spread_constraints
         )
 
+    def get_projected_sa_volumes(self) -> List[ProjectedSAVolume]:
+        config_volumes = super().get_projected_sa_volumes()
+        token_config = (
+            load_system_paasta_config().get_service_auth_token_volume_config()
```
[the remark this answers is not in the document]
piax93: The "PaaSTA Services" of CEP-2167 describes in more detail, but in short, yes. The only thing that we can actually control is the token audience, a that's going to be the same everywhere. Identity is then inferred from other metadata fields.

danielpops (approved):

PR #3889, opened 2024-06-05
Add support for generating vitess config from minimal yelpsoa configs

VinaySagarGonabavi:
*Using multiple commits to iterate on comments and making incremental changes, will squash into a single commit at the end*

Local testing:
Added a make target
```
+.PHONY: setup-kubernetes-cr
+setup-kubernetes-cr:  setup-kubernetes-job
+       export KUBECONFIG=./k8s_itests/kubeconfig;\
+       export PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_playground/;\
+       export PAASTA_TEST_CLUSTER=kind-${USER}-k8s-test;\
+       .tox/py38-linux/bin/python -m paasta_tools.setup_kubernetes_cr -d ./soa_config_playground -c kind-${USER}-k8s-test -s vitess-k8s
```
*Need to use kubeconfig context to apply CRD's in the expected namespace before vitess-k8s deployment can happen*
Test run output: https://fluffy.yelpcorp.com/i/zMjmrhtFwdqh1rz5mcvwPdRHb1XMNl13.html

Sample yelpsoa specification used for local testing:
```
_low_priority_resource_requirements: &low_priority_resource_requirements
  replicas: 1
  requests:
    cpu: 100m
    memory: 256Mi

main:
  deploy_group: "infrastage"
  cells:
  - "uswest1devc"
  zk_address: "10.40.10.198:2181,10.40.23.30:2181,10.40.11.46:2181"
  paasta_pool: "default"
  node_selectors:
    yelp.com/pool: ["default"]
  healthcheck_grace_period_seconds: 60
  healthcheck_mode: cmd
  healthcheck_cmd: 'true'
  vtgate_resources:
    <<: *low_priority_resource_requirements
  vtadmin_resources:
    <<: *low_priority_resource_requirements
  vtctld_resources:
    replicas: 1
    <<: *low_priority_resource_requirements
  keyspaces:
  - keyspace: "yelp_vitess_testing_environment"
    cluster: "vitess-testing-environment"
    vttablet_resources:
      <<: *low_priority_resource_requirements
  - keyspace: "yelp_aux"
    cluster: "aux"
    vttablet_resources:
      <<: *low_priority_resource_requirements
  - keyspace: "yelp"
    cluster: "main"
    vttablet_resources:
      <<: *low_priority_resource_requirements
```

Sample system paasta config added locally in paasta playground:
```
{
    "mysql_port_mappings": {
        "main": {
            "primary": 23129,
            "migration": 22212,
            "read": 23133,
            "reporting": 23137
        },
        "aux": {
            "primary": 23126,
            "migration": 22085,
            "read": 23131,
            "reporting": 23135
        },
        "service": {
            "primary": 23130,
            "migration": 22388,
            "read": 23134,
            "reporting": 23138
        },
        "nowaithost": {
            "primary": 23184,
            "migration": 22220,
            "read": 23186
        },
        "nowaitcad": {
            "primary": 23183,
            "migration": 22756,
            "read": 23185
        },
        "schematizer": {
            "primary": 23213,
            "migration": 22679,
            "read": 23214,
            "reporting": 23062
        },
        "mlflow": {
            "primary": 23955,
            "migration": 22209,
            "read": 23571,
            "reporting": 23095
        },
        "vitess-testing-environment": {
            "primary": 22281,
            "read": 22545,
            "reporting": 22541
        },
        "refresh-main": {
            "primary": 22747,
            "read": 22879
        },
        "refresh-aux": {
            "primary": 22505,
            "read": 22291
        },
        "refresh-service": {
            "primary": 22790,
            "read": 22687
        },
        "refresh-nowaithost": {
            "primary": 22351,
            "read": 22160
        },
        "refresh-nowaitcad": {
            "primary": 22419,
            "read": 22468
        },
        "refresh-schematizer": {
            "primary": 22840,
            "read": 22827
        },
        "sanitized-vitess": {
            "primary": 23824,
            "read": 23454,
            "reporting": 23161
        }
    },
    "vitess_images": {
        "vtctld_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
        "vtgate_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
        "vttablet_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
        "vtadmin_image": "docker-paasta.yelpcorp.com:443/vtadmin:v16.0.3"
    },
    "superregion_to_region_mapping": {
        "norcal": "uswest1",
        "pnw": "uswest2",
        "nova": "useast1",
        "kind-gonabavi-k8s-test": "uswest1-devc"
    },
    "vitess_tablet_types": [
        "primary",
        "migration"
    ],
    "vitess_tablet_pool_type_mapping": {
        "primary": "externalmaster",
        "migration": "externalreplica",
        "read": "externalreplica",
        "reporting": "externalreplica"
    },
    "vitess_throttling_config": {
        "migration": {
            "throttle_query_table": "migration_replication_delay",
            "throttle_metrics_threshold": "7200"
        },
        "read": {
            "throttle_query_table": "read_replication_delay",
            "throttle_metrics_threshold": "3"
        },
        "reporting": {
            "throttle_query_table": "reporting_replication_delay",
            "throttle_metrics_threshold": "7200"
        },
        "primary": {
            "throttle_query_table": "read_replication_delay",
            "throttle_metrics_threshold": "3"
        }
    }
}
```

Test coverage
```
paasta_tools/vitesscluster_tools.py                                               300     25    92%   221, 400-404, 509, 524-525, 530, 654, 664-680, 775-781, 842
```

### DIFF
--- a/paasta_tools/setup_kubernetes_cr.py
+++ b/paasta_tools/setup_kubernetes_cr.py
@@ -360,7 +360,6 @@ def reconcile_kubernetes_resource(
                 config = INSTANCE_TYPE_TO_CONFIG_LOADER[crd.file_prefix](
                     service=service,
                     instance=inst,
-                    instance_type=crd.file_prefix,
                     cluster=cluster,
                     soa_dir=DEFAULT_SOA_DIR,
                 )
--- a/paasta_tools/setup_kubernetes_cr.py
+++ b/paasta_tools/setup_kubernetes_cr.py
@@ -49,9 +49,13 @@ from paasta_tools.utils import get_config_hash
 from paasta_tools.utils import get_git_sha_from_dockerurl
 from paasta_tools.utils import load_all_configs
 from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.vitesscluster_tools import load_vitess_service_instance_configs
 
 
 log = logging.getLogger(__name__)
+
+
+INSTANCE_TYPE_TO_CONFIG_LOADER = {"vitesscluster": load_vitess_service_instance_configs}
 
 
 class StdoutKubeClient:
@@ -352,6 +356,14 @@ def reconcile_kubernetes_resource(
                 load_deployments=True,
                 soa_dir=DEFAULT_SOA_DIR,
             )
+            if crd.file_prefix in INSTANCE_TYPE_TO_CONFIG_LOADER:
+                config = INSTANCE_TYPE_TO_CONFIG_LOADER[crd.file_prefix](
+                    service=service,
+                    instance=inst,
+                    instance_type=crd.file_prefix,
+                    cluster=cluster,
+                    soa_dir=DEFAULT_SOA_DIR,
+                )
             git_sha = get_git_sha_from_dockerurl(soa_config.get_docker_url(), long=True)
             formatted_resource = format_custom_resource(
                 instance_config=config,
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2735,23 +2735,9 @@ class SystemPaastaConfig:
 
     def get_always_authenticating_services(self) -> List[str]:
         return self.config_dict.get("always_authenticating_services", [])
-    
+
     def get_mysql_port_mappings(self) -> Dict:
         return self.config_dict.get("mysql_port_mappings", {})
-    
-    def get_vitess_images(self) -> Dict:
-        return self.config_dict.get(
-            "vitess_images",
-            {
-                "vtctld_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
-                "vtgate_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
-                "vttablet_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
-                "vtadmin_image": "docker-paasta.yelpcorp.com:443/vtadmin:v16.0.3",
-            },
-        )
-
-    def get_superregion_to_region_mapping(self) -> Dict:
-        return self.config_dict.get("superregion_to_region_mapping", {})
 
     def get_vitess_images(self) -> Dict:
         return self.config_dict.get(
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2764,6 +2764,9 @@ class SystemPaastaConfig:
             },
         )
 
+    def get_superregion_to_region_mapping(self) -> Dict:
+        return self.config_dict.get("superregion_to_region_mapping", {})
+
 
 def _run(
     command: Union[str, List[str]],
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2025,6 +2025,9 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     mysql_port_mappings: Dict
     vitess_images: Dict
     superregion_to_region_mapping: Dict
+    vitess_tablet_types: List[str]
+    vitess_tablet_pool_type_mapping: Dict
+    vitess_throttling_config: Dict
 
 
 def load_system_paasta_config(
@@ -2752,6 +2755,35 @@ class SystemPaastaConfig:
 
     def get_superregion_to_region_mapping(self) -> Dict:
         return self.config_dict.get("superregion_to_region_mapping", {})
+
+    def get_vitess_tablet_types(self) -> List:
+        return self.config_dict.get("vitess_tablet_types", ["primary", "migration"])
+
+    def get_vitess_tablet_pool_type_mapping(self) -> Dict:
+        return self.config_dict.get("vitess_tablet_pool_type_mapping", {})
+
+    def get_vitess_throttling_config(self) -> Dict:
+        return self.config_dict.get(
+            "vitess_throttling_config",
+            {
+                "migration": {
+                    "throttle_query_table": "migration_replication_delay",
+                    "throttle_metrics_threshold": "7200",
+                },
+                "read": {
+                    "throttle_query_table": "read_replication_delay",
+                    "throttle_metrics_threshold": "3",
+                },
+                "reporting": {
+                    "throttle_query_table": "reporting_replication_delay",
+                    "throttle_metrics_threshold": "7200",
+                },
+                "primary": {
+                    "throttle_query_table": "read_replication_delay",
+                    "throttle_metrics_threshold": "3",
+                },
+            },
+        )
 
 
 def _run(
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2753,6 +2753,17 @@ class SystemPaastaConfig:
     def get_superregion_to_region_mapping(self) -> Dict:
         return self.config_dict.get("superregion_to_region_mapping", {})
 
+    def get_vitess_images(self) -> Dict:
+        return self.config_dict.get(
+            "vitess_images",
+            {
+                "vtctld_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
+                "vtgate_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
+                "vttablet_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
+                "vtadmin_image": "docker-dev.yelpcorp.com/vtadmin:v16.0.3",
+            },
+        )
+
 
 def _run(
     command: Union[str, List[str]],
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2760,7 +2760,7 @@ class SystemPaastaConfig:
                 "vtctld_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
                 "vtgate_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
                 "vttablet_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
-                "vtadmin_image": "docker-dev.yelpcorp.com/vtadmin:v16.0.3",
+                "vtadmin_image": "docker-paasta.yelpcorp.com:443/vtadmin:v16.0.3",
             },
         )
 
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2022,6 +2022,9 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     use_multiple_log_readers: Optional[List[str]]
     service_auth_token_settings: ProjectedSAVolume
     always_authenticating_services: List[str]
+    mysql_port_mappings: Dict
+    vitess_images: Dict
+    superregion_to_region_mapping: Dict
 
 
 def load_system_paasta_config(
@@ -2732,6 +2735,23 @@ class SystemPaastaConfig:
 
     def get_always_authenticating_services(self) -> List[str]:
         return self.config_dict.get("always_authenticating_services", [])
+    
+    def get_mysql_port_mappings(self) -> Dict:
+        return self.config_dict.get("mysql_port_mappings", {})
+    
+    def get_vitess_images(self) -> Dict:
+        return self.config_dict.get(
+            "vitess_images",
+            {
+                "vtctld_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
+                "vtgate_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
+                "vttablet_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
+                "vtadmin_image": "docker-paasta.yelpcorp.com:443/vtadmin:v16.0.3",
+            },
+        )
+
+    def get_superregion_to_region_mapping(self) -> Dict:
+        return self.config_dict.get("superregion_to_region_mapping", {})
 
 
 def _run(
--- a/paasta_tools/vitesscluster_tools.py
+++ b/paasta_tools/vitesscluster_tools.py
@@ -18,6 +18,7 @@ from paasta_tools.long_running_service_tools import load_service_namespace_confi
 from paasta_tools.utils import BranchDictV2
 from paasta_tools.utils import deep_merge_dictionaries
 from paasta_tools.utils import DEFAULT_SOA_DIR
+from paasta_tools.utils import get_git_sha_from_dockerurl
 from paasta_tools.utils import load_service_instance_config
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import load_v2_deployments_json
@@ -166,25 +167,13 @@ def get_affinity_spec(
     return spec
 
 
-def get_extra_labels(paasta_pool: str, paasta_cluster: str) -> Dict[str, str]:
-    """
-    get extra labels to adhere to paasta contract
-    """
-    extra_labels = {
-        "yelp.com/owner": "dre_mysql",
-        "paasta.yelp.com/cluster": paasta_cluster,
-        "paasta.yelp.com/pool": paasta_pool,
-    }
-    return extra_labels
-
-
 def get_cell_config(
     cell: str,
     paasta_pool: str,
-    paasta_cluster: str,
     region: str,
     vtgate_resources: Dict[str, str],
     env: dict,
+    labels: Dict[str, str],
 ) -> Dict[str, Collection[str]]:
     """
     get vtgate config
@@ -211,7 +200,7 @@ def get_cell_config(
                 "mysql_auth_vault_ttl": "60s",
             },
             "affinity": get_affinity_spec(paasta_pool),
-            "extraLabels": get_extra_labels(paasta_pool, paasta_cluster),
+            "extraLabels": labels,
             "extraEnv": updated_vtgate_extra_env,
             "replicas": replicas,
             "resources": {
@@ -226,10 +215,10 @@ def get_cell_config(
 def get_vitess_dashboard_config(
     cells: List[str],
     paasta_pool: str,
-    paasta_cluster: str,
     zk_address: str,
     vtctld_resources: Dict[str, str],
     env: dict,
+    labels: Dict[str, str],
 ) -> Dict[str, object]:
     """
     get vtctld config
@@ -247,7 +236,7 @@ def get_vitess_dashboard_config(
     config = {
         "cells": cells,
         "affinity": get_affinity_spec(paasta_pool),
-        "extraLabels": get_extra_labels(paasta_pool, paasta_cluster),
+        "extraLabels": labels,
         "extraEnv": updated_vtctld_extra_env,
         "extraFlags": VTCTLD_EXTRA_FLAGS,
         "replicas": replicas,
@@ -263,9 +252,9 @@ def get_vitess_dashboard_config(
 def get_vt_admin_config(
     cells: List[str],
     paasta_pool: str,
-    paasta_cluster: str,
     vtadmin_resources: Dict[str, str],
     env: dict,
+    labels: Dict[str, str],
 ) -> Dict[str, Union[Collection[object], int]]:
     """
     get vtadmin config
@@ -276,7 +265,7 @@ def get_vt_admin_config(
         "cells": cells,
         "apiAddresses": ["http://localhost:15000"],
         "affinity": get_affinity_spec(paasta_pool),
-        "extraLabels": get_extra_labels(paasta_pool, paasta_cluster),
+        "extraLabels": labels,
         "extraFlags": VTADMIN_EXTRA_FLAGS,
         "extraEnv": env,
         "replicas": replicas,
@@ -299,7 +288,6 @@ def get_tablet_pool_config(
     keyspace: str,
     port: str,
     paasta_pool: str,
-    paasta_cluster: str,
     zk_address: str,
     throttle_query_table: str,
     throttle_metrics_threshold: str,
@@ -307,6 +295,7 @@ def get_tablet_pool_config(
     region: str,
     vttablet_resources: Dict[str, str],
     env: dict,
+    labels: Dict[str, str],
 ) -> Dict[str, object]:
     """
     get vttablet config
@@ -353,7 +342,7 @@ def get_tablet_pool_config(
         "name": f"{db_name}_{tablet_type}",
         "type": type,
         "affinity": get_affinity_spec(paasta_pool),
-        "extraLabels": get_extra_labels(paasta_pool, paasta_cluster),
+        "extraLabels": labels,
         "extraEnv": updated_vttablet_extra_env,
         "extraVolumeMounts": [
             {
@@ -422,10 +411,10 @@ def get_keyspaces_config(
     cells: List[str],
     keyspaces: List[Dict[str, Any]],
     paasta_pool: str,
-    paasta_cluster: str,
     zk_address: str,
     region: str,
     env: dict,
+    labels: Dict[str, str],
 ) -> List[Dict[str, object]]:
     """
     get vitess keyspace config
@@ -473,7 +462,6 @@ def get_keyspaces_config(
                         keyspace,
                         port,
                         paasta_pool,
-                        paasta_cluster,
                         zk_address,
                         throttle_query_table,
                         throttle_metrics_threshold,
@@ -481,6 +469,7 @@ def get_keyspaces_config(
                         region,
                         vttablet_resources,
                         env,
+                        labels,
                     )
                     for cell in cells
                 ]
@@ -514,14 +503,12 @@ def get_keyspaces_config(
 class VitessDeploymentConfigDict(KubernetesDeploymentConfigDict, total=False):
     images: Dict[str, str]
     cells: List[str]
-    region: str
     zk_address: str
     vtgate_resources: Dict[str, str]
     vtadmin_resources: Dict[str, str]
     vtctld_resources: Dict[str, str]
     keyspaces: List[Dict[str, Any]]
-    updateStrategy: Dict[str, str]
-    paasta_cluster: str
+    update_strategy: Dict[str, str]
 
 
 class VitessDeploymentConfig(KubernetesDeploymentConfig):
@@ -561,6 +548,33 @@ class VitessDeploymentConfig(KubernetesDeploymentConfig):
             "vttablet": vitess_images["vttablet_image"],
         }
 
+    def get_env_variables(self) -> List[Dict[str, Any]]:
+        # get all K8s container env vars and format their keys to camel case
+        env = [env.to_dict() for env in self.get_container_env()]
+        formatted_env = [get_formatted_container_env(env_var) for env_var in env]
+        return formatted_env
+
+    def get_labels(self) -> Dict[str, str]:
+        # get default labels from parent class to adhere to paasta contract
+        docker_url = self.get_docker_url(
+            system_paasta_config=load_system_paasta_config()
+        )
+        git_sha = get_git_sha_from_dockerurl(docker_url)
+        return self.get_kubernetes_metadata(git_sha=git_sha).labels
+
+    def get_region(self) -> str:
+        superregion = self.get_cluster()
+        superregion_to_region_map = (
+            load_system_paasta_config().get_superregion_to_region_mapping()
+        )
+        for superregion_prefix in superregion_to_region_map:
+            if superregion.startswith(superregion_prefix):
+                return superregion.replace(
+                    superregion_prefix, superregion_to_region_map[superregion_prefix]
+                )
+
+        log.error(f"Invalid superregion {superregion}")
+
     def get_global_lock_server(self) -> Dict[str, Dict[str, str]]:
         zk_address = self.config_dict.get("zk_address")
         return {
@@ -571,75 +585,63 @@ class VitessDeploymentConfig(KubernetesDeploymentConfig):
             }
         }
 
-    def get_cells(self) -> List[str]:
+    def get_cells(self) -> List[Any]:
         paasta_pool = self.get_pool()
-        paasta_cluster = self.config_dict.get("paasta_cluster", self.get_cluster())
         cells = self.config_dict.get("cells")
-        region = self.config_dict.get("region")
+        region = self.get_region()
         vtgate_resources = self.config_dict.get("vtgate_resources")
-        env = [env.to_dict() for env in self.get_container_env()]
-        formatted_env = [get_formatted_container_env(env_var) for env_var in env]
+
+        formatted_env = self.get_env_variables()
+        labels = self.get_labels()
+
         return [  # type: ignore
             get_cell_config(
-                cell,
-                paasta_pool,
-                paasta_cluster,
-                region,
-                vtgate_resources,
-                formatted_env,
+                cell, paasta_pool, region, vtgate_resources, formatted_env, labels
             )
             for cell in cells
         ]
 
     def get_vitess_dashboard(self) -> Dict[str, object]:
         paasta_pool = self.get_pool()
-        paasta_cluster = self.config_dict.get("paasta_cluster", self.get_cluster())
         cells = self.config_dict.get("cells")
         zk_address = self.config_dict.get("zk_address")
         vtctld_resources = self.config_dict.get("vtctld_resources")
-        env = [env.to_dict() for env in self.get_container_env()]
-        formatted_env = [get_formatted_container_env(env_var) for env_var in env]
+
+        formatted_env = self.get_env_variables()
+        labels = self.get_labels()
+
         return get_vitess_dashboard_config(
-            cells,
-            paasta_pool,
-            paasta_cluster,
-            zk_address,
-            vtctld_resources,
-            formatted_env,
+            cells, paasta_pool, zk_address, vtctld_resources, formatted_env, labels
         )
 
     def get_vtadmin(self) -> Dict[str, Union[Collection[object], int]]:
         paasta_pool = self.get_pool()
-        paasta_cluster = self.config_dict.get("paasta_cluster", self.get_cluster())
         cells = self.config_dict.get("cells")
         vtadmin_resources = self.config_dict.get("vtadmin_resources")
-        env = [env.to_dict() for env in self.get_container_env()]
-        formatted_env = [get_formatted_container_env(env_var) for env_var in env]
+
+        formatted_env = self.get_env_variables()
+        labels = self.get_labels()
+
         return get_vt_admin_config(
-            cells, paasta_pool, paasta_cluster, vtadmin_resources, formatted_env
+            cells, paasta_pool, vtadmin_resources, formatted_env, labels
         )
 
     def get_keyspaces(self) -> List[Dict[str, object]]:
         paasta_pool = self.get_pool()
-        paasta_cluster = self.config_dict.get("paasta_cluster", self.get_cluster())
         cells = self.config_dict.get("cells")
         zk_address = self.config_dict.get("zk_address")
-        region = self.config_dict.get("region")
+        region = self.get_region()
         keyspaces = self.config_dict.get("keyspaces")
-        env = [env.to_dict() for env in self.get_container_env()]
-        formatted_env = [get_formatted_container_env(env_var) for env_var in env]
+
+        formatted_env = self.get_env_variables()
+        labels = self.get_labels()
+
         return get_keyspaces_config(
-            cells,
-            keyspaces,
-            paasta_pool,
-            paasta_cluster,
-            zk_address,
-            region,
-            formatted_env,
+            cells, keyspaces, paasta_pool, zk_address, region, formatted_env, labels
         )
 
     def get_update_strategy(self) -> Dict[str, str]:
-        return self.config_dict.get("updateStrategy", {"type": "Immediate"})
+        return self.config_dict.get("update_strategy", {"type": "Immediate"})
 
     def get_vitess_config(self) -> Dict[str, Any]:
         smartstack_config = load_service_namespace_config(
@@ -691,47 +693,10 @@ class VitessDeploymentConfig(KubernetesDeploymentConfig):
             return []
 
 
-def generate_vitess_instance_config(
-    instance_config: VitessDeploymentConfigDict,
-) -> VitessDeploymentConfigDict:
-    # Generate vitess instance config from yelpsoa config
-    zk_address = instance_config.get("zk_address")
-    cells = instance_config.get("cells")
-    keyspaces = instance_config.get("keyspaces")
-    region = instance_config.get("region")
-    vtgate_resources = instance_config.get("vtgate_resources")
-    vtadmin_resources = instance_config.get("vtadmin_resources")
-    vtctld_resources = instance_config.get("vtctld_resources")
-    deploy_group = instance_config.get("deploy_group")
-
-    # healthcheck cmd and mode are required to be in config_dict for LongRunningServiceConfig
-    healthcheck_cmd = instance_config.get("healthcheck_cmd")
-    healthcheck_mode = instance_config.get("healthcheck_mode")
-
-    vitess_images = load_system_paasta_config().get_vitess_images()
-
-    vitess_instance_config: dict = {
-        "zk_address": zk_address,
-        "cells": cells,
-        "region": region,
-        "vtgate_resources": vtgate_resources,
-        "vtadmin_resources": vtadmin_resources,
-        "vtctld_resources": vtctld_resources,
-        "keyspaces": keyspaces,
-        "images": vitess_images,
-        "updateStrategy": {"type": "Immediate"},
-        "healthcheck_cmd": healthcheck_cmd,
-        "healthcheck_mode": healthcheck_mode,
-        "deploy_group": deploy_group,
-    }
-    return VitessDeploymentConfigDict(vitess_instance_config)  # type: ignore
-
-
 def load_vitess_instance_config(
     service: str,
     instance: str,
     cluster: str,
-    instance_type: str = "vitesscluster",
     load_deployments: bool = True,
     soa_dir: str = DEFAULT_SOA_DIR,
 ) -> VitessDeploymentConfig:
@@ -740,13 +705,12 @@ def load_vitess_instance_config(
     )
     instance_config = VitessDeploymentConfigDict(
         load_service_instance_config(  # type: ignore
-            service, instance, instance_type, cluster, soa_dir=soa_dir
+            service, instance, "vitesscluster", cluster, soa_dir=soa_dir
         )
     )
-    vitess_instance_config = generate_vitess_instance_config(instance_config)
 
     general_config = deep_merge_dictionaries(
-        overrides=vitess_instance_config, defaults=general_config
+        overrides=instance_config, defaults=general_config
     )
 
     branch_dict: Optional[BranchDictV2] = None
@@ -779,12 +743,11 @@ def load_vitess_instance_config(
 def load_vitess_service_instance_configs(
     service: str,
     instance: str,
-    instance_type: str,
     cluster: str,
     soa_dir: str = DEFAULT_SOA_DIR,
 ) -> Dict[str, str]:
     vitess_service_instance_configs = load_vitess_instance_config(
-        service, instance, cluster, instance_type, soa_dir=soa_dir
+        service, instance, cluster, soa_dir=soa_dir
     ).get_vitess_config()
     return vitess_service_instance_configs
 
--- a/paasta_tools/vitesscluster_tools.py
+++ b/paasta_tools/vitesscluster_tools.py
@@ -5,10 +5,10 @@ from typing import Dict
 from typing import List
 from typing import Mapping
 from typing import Optional
-from typing import Sequence
 from typing import Union
 
 import service_configuration_lib
+from kubernetes.client import ApiClient
 
 from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
 from paasta_tools.kubernetes_tools import KubernetesDeploymentConfigDict
@@ -22,9 +22,6 @@ from paasta_tools.utils import get_git_sha_from_dockerurl
 from paasta_tools.utils import load_service_instance_config
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.utils import load_v2_deployments_json
-
-# from paasta_tools.long_running_service_tools import LongRunningServiceConfig
-# from paasta_tools.long_running_service_tools import LongRunningServiceConfigDict
 
 
 log = logging.getLogger(__name__)
@@ -126,54 +123,13 @@ def get_updated_environment_variables(env_vars: Dict[str, Any]) -> List[dict]:
     return updated_environment_variables
 
 
-def get_formatted_container_env(
-    env_var: Dict[str, Any],
-) -> Dict[str, Any]:
-    """
-    Helper function to recursively format K8s container environment variable dict which have keys in snake case and modify them to camel case
-    """
-    formatted_env_var = {}
-    for key in env_var.keys():
-        temp_key = key.split("_")
-        new_key = temp_key[0] + "".join(element.title() for element in temp_key[1:])
-        if isinstance(env_var[key], dict):
-            formatted_env_var[new_key] = get_formatted_container_env(env_var[key])
-        else:
-            formatted_env_var[new_key] = env_var[key]
-
-    return formatted_env_var
-
-
-def get_affinity_spec(
-    paasta_pool: str,
-) -> Dict[str, Dict[str, Dict[str, List[Dict[str, List[Dict[str, Sequence[str]]]]]]]]:
-    spec = {
-        "nodeAffinity": {
-            "requiredDuringSchedulingIgnoredDuringExecution": {
-                "nodeSelectorTerms": [
-                    {
-                        "matchExpressions": [
-                            {
-                                "key": "yelp.com/pool",
-                                "operator": "In",
-                                "values": [paasta_pool],
-                            }
-                        ]
-                    }
-                ]
-            }
-        }
-    }
-    return spec
-
-
 def get_cell_config(
     cell: str,
-    paasta_pool: str,
     region: str,
     vtgate_resources: Dict[str, str],
     env: dict,
     labels: Dict[str, str],
+    node_affinity: dict,
 ) -> Dict[str, Collection[str]]:
     """
     get vtgate config
@@ -199,7 +155,7 @@ def get_cell_config(
                 "mysql_auth_vault_tls_ca": f"/etc/vault/all_cas/acm-privateca-{region}.crt",
                 "mysql_auth_vault_ttl": "60s",
             },
-            "affinity": get_affinity_spec(paasta_pool),
+            "affinity": {"nodeAffinity": node_affinity},
             "extraLabels": labels,
             "extraEnv": updated_vtgate_extra_env,
             "replicas": replicas,
@@ -214,11 +170,11 @@ def get_cell_config(
 
 def get_vitess_dashboard_config(
     cells: List[str],
-    paasta_pool: str,
     zk_address: str,
     vtctld_resources: Dict[str, str],
     env: dict,
     labels: Dict[str, str],
+    node_affinity: dict,
 ) -> Dict[str, object]:
     """
     get vtctld config
@@ -235,7 +191,7 @@ def get_vitess_dashboard_config(
 
     config = {
         "cells": cells,
-        "affinity": get_affinity_spec(paasta_pool),
+        "affinity": {"nodeAffinity": node_affinity},
         "extraLabels": labels,
         "extraEnv": updated_vtctld_extra_env,
         "extraFlags": VTCTLD_EXTRA_FLAGS,
@@ -251,10 +207,10 @@ def get_vitess_dashboard_config(
 
 def get_vt_admin_config(
     cells: List[str],
-    paasta_pool: str,
     vtadmin_resources: Dict[str, str],
     env: dict,
     labels: Dict[str, str],
+    node_affinity: dict,
 ) -> Dict[str, Union[Collection[object], int]]:
     """
     get vtadmin config
@@ -264,7 +220,7 @@ def get_vt_admin_config(
     config = {
         "cells": cells,
         "apiAddresses": ["http://localhost:15000"],
-        "affinity": get_affinity_spec(paasta_pool),
+        "affinity": {"nodeAffinity": node_affinity},
         "extraLabels": labels,
         "extraFlags": VTADMIN_EXTRA_FLAGS,
         "extraEnv": env,
@@ -287,7 +243,6 @@ def get_tablet_pool_config(
     db_name: str,
     keyspace: str,
     port: str,
-    paasta_pool: str,
     zk_address: str,
     throttle_query_table: str,
     throttle_metrics_threshold: str,
@@ -296,6 +251,7 @@ def get_tablet_pool_config(
     vttablet_resources: Dict[str, str],
     env: dict,
     labels: Dict[str, str],
+    node_affinity: dict,
 ) -> Dict[str, object]:
     """
     get vttablet config
@@ -341,7 +297,7 @@ def get_tablet_pool_config(
         "cell": cell,
         "name": f"{db_name}_{tablet_type}",
         "type": type,
-        "affinity": get_affinity_spec(paasta_pool),
+        "affinity": {"nodeAffinity": node_affinity},
         "extraLabels": labels,
         "extraEnv": updated_vttablet_extra_env,
         "extraVolumeMounts": [
@@ -410,11 +366,11 @@ def get_tablet_pool_config(
 def get_keyspaces_config(
     cells: List[str],
     keyspaces: List[Dict[str, Any]],
-    paasta_pool: str,
     zk_address: str,
     region: str,
     env: dict,
     labels: Dict[str, str],
+    node_affinity: dict,
 ) -> List[Dict[str, object]]:
     """
     get vitess keyspace config
@@ -461,7 +417,6 @@ def get_keyspaces_config(
                         db_name,
                         keyspace,
                         port,
-                        paasta_pool,
                         zk_address,
                         throttle_query_table,
                         throttle_metrics_threshold,
@@ -470,6 +425,7 @@ def get_keyspaces_config(
                         vttablet_resources,
                         env,
                         labels,
+                        node_affinity,
                     )
                     for cell in cells
                 ]
@@ -550,9 +506,14 @@ class VitessDeploymentConfig(KubernetesDeploymentConfig):
 
     def get_env_variables(self) -> List[Dict[str, Any]]:
         # get all K8s container env vars and format their keys to camel case
-        env = [env.to_dict() for env in self.get_container_env()]
-        formatted_env = [get_formatted_container_env(env_var) for env_var in env]
-        return formatted_env
+
+        # Workaround from https://github.com/kubernetes-client/python/issues/390
+        api_client = ApiClient()
+        env = [
+            api_client.sanitize_for_serialization(env)
+            for env in self.get_container_env()
+        ]
+        return env
 
     def get_labels(self) -> Dict[str, str]:
         # get default labels from parent class to adhere to paasta contract
@@ -561,6 +522,21 @@ class VitessDeploymentConfig(KubernetesDeploymentConfig):
         )
         git_sha = get_git_sha_from_dockerurl(docker_url)
         return self.get_kubernetes_metadata(git_sha=git_sha).labels
+
+    def get_vitess_node_affinity(self) -> dict:
+        paasta_pool = self.get_pool()
+        self.config_dict.update(
+            {
+                "node_selectors": {
+                    "yelp.com/pool": [paasta_pool],
+                }
+            }
+        )
+
+        # Workaround from https://github.com/kubernetes-client/python/issues/390
+        api_client = ApiClient()
+        node_affinity = api_client.sanitize_for_serialization(self.get_node_affinity())
+        return node_affinity
 
     def get_region(self) -> str:
         superregion = self.get_cluster()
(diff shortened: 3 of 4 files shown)


paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+# Global variables
+TOPO_IMPLEMENTATION = "zk2"
+TOPO_GLOBAL_ROOT = "/vitess-paasta/global"
+SOURCE_DB_HOST = "169.254.255.254"
+TABLET_TYPES = ["primary", "migration"]
+WEB_PORT = "15000"
+GRPC_PORT = "15999"
+
+# Mapping of mysql cluster to target role envoy port
+MYSQL_CLUSTER_PORT_MAP = {
```
VinaySagarGonabavi, line 37: Will move this to srv-configs or some other internal repo

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+                "mysql_auth_vault_tls_ca": f"/etc/vault/all_cas/acm-privateca-{region}.crt",
+                "mysql_auth_vault_ttl": "60s",
+            },
+            "affinity": build_affinity_spec(paasta_pool),
+            "extraLabels": build_extra_labels(paasta_pool, paasta_cluster),
+            "extraEnv": build_extra_env(paasta_cluster),
+            "replicas": 1,
+            "resources": {
+                "requests": {
+                    "cpu": "100m",
```
VinaySagarGonabavi, line 319: These resource specifications cannot be static, will modify

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+    config = {
+        "cells": cells,
+        "affinity": build_affinity_spec(paasta_pool),
+        "extraLabels": build_extra_labels(paasta_pool, paasta_cluster),
+        "extraEnv": build_extra_env(paasta_cluster),
+        "extraFlags": VTCTLD_EXTRA_FLAGS,
+        "replicas": 1,
+        "resources": {
+            "requests": {
+                "cpu": "100m",
```
VinaySagarGonabavi, line 349: Same here, have to modify resource specifications based on service metadata in yelpsoa

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+        "apiAddresses": ["http://localhost:15000"],
+        "affinity": build_affinity_spec(paasta_pool),
+        "extraLabels": build_extra_labels(paasta_pool, paasta_cluster),
+        "extraFlags": VTADMIN_EXTRA_FLAGS,
+        "extraEnv": build_extra_env(paasta_cluster),
+        "replicas": 1,
+        "readOnly": False,
+        "apiResources": {
+            "requests": {
+                "cpu": "100m",
```
VinaySagarGonabavi, line 381: Same here, have to modify resource specifications based on service metadata in yelpsoa

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+            },
+            {"name": "vttablet-fake-credentials", "hostPath": {"path": "/dev/null"}},
+            {"name": "keyspace-fake-init-script", "hostPath": {"path": "/dev/null"}},
+        ],
+        "replicas": 1,
+        "vttablet": {
+            "extraFlags": vttablet_extra_flags,
+            "resources": {
+                "requests": {
+                    "cpu": "100m",
```
VinaySagarGonabavi, line 484: Same here, have to modify resource specifications based on service metadata in yelpsoa

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -49,11 +49,15 @@
...
 from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.vitesscluster_tools import load_vitess_service_instance_configs
 
 
 log = logging.getLogger(__name__)
 
 
+INSTANCE_TYPE_TO_CONFIG_LOADER = {"vitesscluster": load_vitess_service_instance_configs}
+
+
```
jfongatyelp, line 60: You shouldn't need to do this here; this script is already using the LONG_RUNNING_INSTANCE_TYPE_HANDLERS from here, which already has config for vitesscluster instance types: https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/utils.py#L697-L699

If the config loader listed there is incorrect, you should update it to your new one in that file rather than adding new logic here.

jfongatyelp (changes requested):

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+        if env["name"] == "KEYSPACE":
+            env["value"] = keyspace
+        if env["name"] == "VAULT_ADDR":
+            env["value"] = f"https://vault-dre.{region}.yelpcorp.com:8200"
+        if env["name"] == "VAULT_CACERT":
+            env["value"] = f"/etc/vault/all_cas/acm-privateca-{region}.crt"
+        config["extraEnv"].append(env)
+
+    # Add extra pod label to filter
+    config["extraLabels"]["tablet_type"] = f"{db_name}_{tablet_type}"
```
nemacysts, line 526: what's the longest db name and longest tablet type? labels are limited to 63 characters, so we'll likely need to truncate here (`limit_size_with_hash()` may be helpful for y'all here :))

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+            "user": "vt_app",
+            "credentialsSecret": {
+                "key": "/etc/credentials.yaml",
+                "volumeName": "vttablet-fake-credentials",
+            },
+        },
+        "dataVolumeClaimTemplate": {
+            "accessModes": ["ReadWriteOnce"],
+            "resources": {"requests": {"storage": "10Gi"}},
+            "storageClassName": "ebs-csi-gp3",
```
nemacysts, line 503: (we'll want to make sure that this storage class exists in all our clusters)

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+                "cpu": "100m",
+                "memory": "128Mi",
+            },
+            "limits": {"memory": "128Mi"},
+        },
+    }
+    return config
+
+
+def build_tablet_pool_config(
```
nemacysts, line 397: same as above, we'll want to add types here

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+        if env["name"] == "TOPOLOGY_FLAGS":
+            env[
+                "value"
+            ] = f"--topo_implementation {TOPO_IMPLEMENTATION} --topo_global_server_address ${zk_address} --topo_global_root {TOPO_GLOBAL_ROOT}"
+        config["extraEnv"].append(env)
+
+    return config
+
+
+def build_vt_admin_config(cells, paasta_pool, paasta_cluster):
```
nemacysts, line 366: same as above, we'll want to add types here

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+    for env in vtgate_extra_env:
+        if env["name"] == "VAULT_ADDR":
+            env["value"] = f"https://vault-dre.{region}.yelpcorp.com:8200"
+        if env["name"] == "VAULT_CACERT":
+            env["value"] = f"/etc/vault/all_cas/acm-privateca-{region}.crt"
+        config["gateway"]["extraEnv"].append(env)
+    return config
+
+
+def build_vitess_dashboard_config(cells, paasta_pool, paasta_cluster, zk_address):
```
nemacysts, line 336: same as above, we'll want to add types here

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+        {
+            "name": "PAASTA_POD_IP",
+            "valueFrom": {"fieldRef": {"fieldPath": "status.podIP"}},
+        },
+        {"name": "PAASTA_CLUSTER", "value": paasta_cluster},
+    ]
+    return extra_env
+
+
+def build_cell_config(cell, paasta_pool, paasta_cluster, region):
```
nemacysts, line 299: same as above, we'll want to add types here

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+    }
+    return spec
+
+
+def build_extra_labels(paasta_pool, paasta_cluster):
+    """
+    Build extra labels to adhere to paasta contract
+    """
+    extra_labels = {
+        "yelp.com/owner": "dre_mysql_working_hours",
```
nemacysts, line 278: fwiw: we don't page on these - it's just for ee-metrics use: so it'd probably be "better" to have this be `dre_mysql` (assuming y'all have a sensu team called dre_mysql)

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+            },
+        },
+    }
+    vtgate_extra_env = copy.deepcopy(VTGATE_EXTRA_ENV)
+    for env in vtgate_extra_env:
+        if env["name"] == "VAULT_ADDR":
+            env["value"] = f"https://vault-dre.{region}.yelpcorp.com:8200"
+        if env["name"] == "VAULT_CACERT":
+            env["value"] = f"/etc/vault/all_cas/acm-privateca-{region}.crt"
+        config["gateway"]["extraEnv"].append(env)
```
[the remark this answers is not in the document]
nemacysts: (and especially since we have to do this for loop in a couple other functions)

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+            env["value"] = f"/etc/vault/all_cas/acm-privateca-{region}.crt"
+        config["extraEnv"].append(env)
+
+    # Add extra pod label to filter
+    config["extraLabels"]["tablet_type"] = f"{db_name}_{tablet_type}"
+
+    return config
+
+
+def build_keyspaces_config(
```
nemacysts, line 531: same as above, we'll want to add types here

paasta_tools/vitesscluster_tools.py
```diff
@@ -72,21 +662,98 @@ def validate(
...
-) -> VitessDeploymentConfig:
+) -> VitessDeploymentConfigDict:
     general_config = service_configuration_lib.read_service_configuration(
         service, soa_dir=soa_dir
     )
     instance_config = load_service_instance_config(
-        service, instance, "vitesscluster", cluster, soa_dir=soa_dir
+        service, instance, instance_type, cluster, soa_dir=soa_dir
     )
+    vitess_instance_config = generate_vitess_instance_config(Dict(instance_config))
```
nemacysts, line 740: TIL that the typing.X classes can be used as normal constructors - that said, i don't think we need this if we update the parameter typing for this function to be `VitessDeploymentConfig`

paasta_tools/vitesscluster_tools.py
```diff
@@ -72,21 +662,98 @@ def validate(
...
+        "env": {
+            "OPERATOR_NAME": "vitess-operator",
+            "POD_NAME": "vitess-k8s",
+            "PS_OPERATOR_POD_NAME": "vitess-k8s",
+            "PS_OPERATOR_POD_NAMESPACE": "paasta-vitessclusters",
+            "WATCH_NAMESPACE": "paasta-vitessclusters",
+        },
+        "healthcheck_grace_period_seconds": 60,
+        "healthcheck_mode": "cmd",
+        "healthcheck_cmd": "true",
```
nemacysts, line 697: i assume we'll have a real healthcheck later?

paasta_tools/vitesscluster_tools.py
```diff
@@ -72,21 +662,98 @@ def validate(
...
+        "deploy_group": deploy_group,
+        "autoscaling": {"setpoint": 0.7},
+        "env": {
+            "OPERATOR_NAME": "vitess-operator",
+            "POD_NAME": "vitess-k8s",
+            "PS_OPERATOR_POD_NAME": "vitess-k8s",
+            "PS_OPERATOR_POD_NAMESPACE": "paasta-vitessclusters",
+            "WATCH_NAMESPACE": "paasta-vitessclusters",
+        },
+        "healthcheck_grace_period_seconds": 60,
```
nemacysts, line 695: might be good to make this configurable from the start - iirc, there was an incident for one of the other operator owners a while back where their healthchecks slowed down for some reason and it was a bit of a pain to make the healthcheck config configurable (or edit it to start with) during the incident)

paasta_tools/vitesscluster_tools.py
```diff
@@ -72,21 +662,98 @@ def validate(
...
+    region = instance_config.get("region")
+
+    vitess_instance_config = {
+        "namespace": "paasta-vitessclusters",
+        "cpus": cpus,
+        "mem": mem,
+        "min_instances": 1,
+        "max_instances": 1,
+        "deploy_group": deploy_group,
+        "autoscaling": {"setpoint": 0.7},
```
nemacysts, line 687: are we intending to autoscale this based on CPU usage? or will y'all add a custom autoscaling signal later?

paasta_tools/vitesscluster_tools.py
```diff
@@ -72,21 +662,98 @@ def validate(
...
 
-def load_vitess_instance_config(
+def generate_vitess_instance_config(
+    instance_config: Dict,
+) -> Dict:
+    # Generate the vitess instance config from the yelpsoa config
+
+    cpus = instance_config.get("cpus")
+    mem = instance_config.get("mem")
+    deploy_group = instance_config.get("deploy_group")
```
nemacysts, line 672: a couple of these can probably use the LongRunningServiceConfigDict getters rather than directly reading the config dict (which would also have the benefit of already setting defaults for these)

paasta_tools/vitesscluster_tools.py
```diff
@@ -72,21 +662,98 @@ def validate(
...
+    zk_address = instance_config.get("zk_address")
+    paasta_pool = instance_config.get("paasta_pool")
```
nemacysts, line 674: e.g., if this was get_pool()`, that would handle there not being a pool set and automatically set this to "default"

(or y'all could override get_pool for VitessDeploymentConfig to return the default pool you'd like to use)

paasta_tools/vitesscluster_tools.py
```diff
@@ -72,21 +662,98 @@ def validate(
...
+            for cell in cells
+        ],
+        "vitessDashboard": build_vitess_dashboard_config(
+            cells, paasta_pool, paasta_cluster, zk_address
+        ),
+        "vtadmin": build_vt_admin_config(cells, paasta_pool, paasta_cluster),
+        "keyspaces": build_keyspaces_config(
+            cells, keyspaces, paasta_pool, paasta_cluster, zk_address, region
+        ),
+        "updateStrategy": {"type": "Immediate"},
```
nemacysts, line 722: hmm, is this a statefulset update strategy? https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#update-strategies only documents OnDelete and RollingUpdate

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+    "enable-lag-throttler": "true",
+    "throttle_check_as_check_self": "true",
+    "throttle_metrics_query": "",
+    "throttle_metrics_threshold": "",
+    "db_charset": "utf8mb4",
+    "disable_active_reparents": "true",
+}
+
+
+def build_affinity_spec(paasta_pool):
```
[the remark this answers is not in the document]
nemacysts: and small naming note in case we're interested in being consistent with the paasta service code (and the other operators) - these types of functions are usually named get_X in the rest of paasta (e.g., get_node_affinity, get_extra_labels, etc)

nemacysts (commented):
might also be helpful to add an example of what we're expecting the soaconfigs config to look like :)

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: ++. This helps. Will move this as a config to puppet

paasta_tools/setup_kubernetes_cr.py
VinaySagarGonabavi: While this is right in some functionality, especially setup-kubernetes-job, when it comes to setup-kubernetes-cr reconciliation step, the config used to call kubectl client to create/update a resource https://github.com/Yelp/paasta/blob/master/paasta_tools/setup_kubernetes_cr.py#L357 uses config generated from https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L3264 which just loads configs from yelpsoa and not those generated in vitesscluster tooling. I kept running into this and have modified it this way.

Lmk if there is something I'm missing here or if this can be done differently

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
 from paasta_tools.utils import load_v2_deployments_json
 
 KUBERNETES_NAMESPACE = "paasta-vitessclusters"
+# Image variables
+IMAGE_TAG = "v16.0.3"
+VTCTLD_IMAGE = f"docker-paasta.yelpcorp.com:443/vitess_base:{IMAGE_TAG}"
+VT_GATE_IMAGE = f"docker-paasta.yelpcorp.com:443/vitess_base:{IMAGE_TAG}"
+VT_TABLET_IMAGE = f"docker-paasta.yelpcorp.com:443/vitess_base:{IMAGE_TAG}"
+VT_ADMIN_IMAGE = f"docker-dev.yelpcorp.com/vtadmin:{IMAGE_TAG}"
```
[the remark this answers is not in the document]
VinaySagarGonabavi: Yes, I meant to move the image tag as a metadata field in spec as well, so we can roll things out at a paasta cluster/superregion scope. I'll update this

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+        "valueFrom": {
+            "secretKeyRef": {
+                "name": "paasta-vitessclusters-secret-vitess-k8s-vault-vtgate-approle-secretid",
+                "key": "vault-vtgate-approle-secretid",
+            }
+        },
+    },
+    {
+        "name": "VAULT_CACERT",
+        "value": "/etc/vault/all_cas/acm-privateca-uswest1-devc.crt",
```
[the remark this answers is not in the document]
VinaySagarGonabavi: The global env variable has a default value, but later down, it's updated based on the region value passed in metadata config

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+    },
+]
+
+# Vault auth related variables
+VTGATE_EXTRA_ENV = [
+    {
+        "name": "VAULT_ADDR",
+        "value": "https://vault-dre.uswest1-devc.yelpcorp.com:8200",
```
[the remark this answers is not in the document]
VinaySagarGonabavi: The global env variable has a default value, but later down, it's updated based on the region value passed in metadata config

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+        },
+    },
+    {
+        "name": "VAULT_SECRETID",
+        "valueFrom": {
+            "secretKeyRef": {
+                "name": "paasta-vitessclusters-secret-vitess-k8s-vault-vtgate-approle-secretid",
+                "key": "vault-vtgate-approle-secretid",
+            }
+        },
```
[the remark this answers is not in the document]
VinaySagarGonabavi: Please correct me if I'm wrong, but we're not planning to use vault secrets as mount path on the pods (vtgate, vttablet), but instead are directly reading from vault for each auth operation. The vault path (currently named `secrets/vitess/vt-tablet/vttablet_credentials.json` or `secrets/vitess/vt-gate/vttablet_credentials.json`) with secrets will get new or updated secrets/credentials on frequent basis outside of the deployment. But, the role id and secret id wouldn't change, and are needed to be setup only during the initial deployment unless we update the secret id itself?

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
nemacysts: ah, i did notice that later on - that said, is it worth having a default that we'll always need to override? could we remove this and rely on the config generation to set the correct value?

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: > that said: for paasta services we tend to use the k8s client models (and then convert those to dicts if necessary) - that would probably be more type-safe (and safer in general) than raw dicts

ack. Will definitely add typed arguments to functions. Makes sense to use client models than raw dicts which we did in templates.py, which could be the majority of the rewrite but useful. ++

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+                "mysql_auth_server_impl": "vault",
+                "mysql_auth_vault_addr": f"https://vault-dre.{region}.yelpcorp.com:8200",
+                "mysql_auth_vault_path": "secrets/vitess/vt-gate/vttablet_credentials.json",
+                "mysql_auth_vault_tls_ca": f"/etc/vault/all_cas/acm-privateca-{region}.crt",
+                "mysql_auth_vault_ttl": "60s",
+            },
+            "affinity": build_affinity_spec(paasta_pool),
+            "extraLabels": build_extra_labels(paasta_pool, paasta_cluster),
+            "extraEnv": build_extra_env(paasta_cluster),
+            "replicas": 1,
```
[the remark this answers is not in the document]
VinaySagarGonabavi: Not always, some static field leftovers from templates.py have crept it. Will clean all static resource specifications in the next/updated commit

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
nemacysts: correct - but we generally want to avoid manually managed secrets: we don't currently do a full sync of secrets (as in, we don't delete unused data) , but in the future i'd like us to fully manage the secrets that we sync and ensure that k8s only has secrets that should actually exist

we've also discussed potentially sharding our EKS clusters, and having data like this that needs to be manually setup during initial deployments would complicate that since everyone involved would need to remember that there's a set of secrets that need to be manually created 

i think that as long as these secrets exist in the secrets/ directory, we'll be fine - barring a small window of time where it's possible for the secret sync to not have run on a fresh cluster/deployment, but this code has (which should be fine if i'm not mistaken: k8s should realize that the secrets eventually exist and everything should proceed)

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: These default extra env are still a dict, but yes, makes sense to have a common function to update or produce a new dict with overrides since it's done in building spec for multiple components differently, vtgate, vttablet, vtadmin etc.,

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Oh, that's helpful to know! There are definitely some longer service names, and the restriction on MySQL side is only on the username which we shorten it based on db name https://yelpwiki.yelpcorp.com/pages/viewpage.action?pageId=63881224#HowTo:CreateanewMySQLServiceDatabase-Guidelines. Will use the function to shorten the label name accordingly

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: True, I used this to get over the failing mypy tox target, but the ideal fix would be to return of type VitessDeploymentConfigDict

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Ah, another leftover from the PoC. We need to query the healthcheck endpoint of all components or vtctld at least to ensure we have a working cluster

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Honestly, we didn't hit the resource limits yet on any of these deployments and can use the default one to start with, but make it configurable in yelpsoa so we don't touch paasta_tools as often. But, yes, if needed in future, we might have to use a custom signal for autoscaling

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: It's not a statefulset update strategy but specific to VitessKeyspaces CRD understood by the operator which only has two options https://github.yelpcorp.com/misc/compute-infra-k8s/blob/master/clusters/eks-infrastage/90-vitess-operator/eks/40-crd-planetscale.com-vitesskeyspaces.yaml#L1074

nemacysts: ah, gotcha :)

paasta_tools/vitesscluster_tools.py
nemacysts: got it - there's some pitfalls with {min,max}_instances + instances, so we may not want to hardcode min/max instances if we're also looking to support `instances: X`

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/a8e1e2a9c43a88c68bfd4a83862cde0c2a94f1da. Resolving

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/6f56e5f7da82df3de44e035247fe865daf409ff4

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/6f56e5f7da82df3de44e035247fe865daf409ff4

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/6f56e5f7da82df3de44e035247fe865daf409ff4

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/6f56e5f7da82df3de44e035247fe865daf409ff4

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/commit/6f56e5f7da82df3de44e035247fe865daf409ff4

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/commit/6f56e5f7da82df3de44e035247fe865daf409ff4

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/0f12a638f8b31b69697a4d9e5ae42843a5adda32

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/0f12a638f8b31b69697a4d9e5ae42843a5adda32

paasta_tools/vitesscluster_tools.py
```diff
@@ -15,6 +17,594 @@
...
+                        ]
+                    }
+                ]
+            }
+        }
+    }
+    return spec
+
+
+def build_extra_labels(paasta_pool, paasta_cluster):
```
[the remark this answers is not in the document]
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/c7a6b7f7665b11eb4e812cba4a46457b851040cd

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/c7a6b7f7665b11eb4e812cba4a46457b851040cd

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/0f12a638f8b31b69697a4d9e5ae42843a5adda32

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/c7a6b7f7665b11eb4e812cba4a46457b851040cd

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/c7a6b7f7665b11eb4e812cba4a46457b851040cd

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/c7a6b7f7665b11eb4e812cba4a46457b851040cd

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/c7a6b7f7665b11eb4e812cba4a46457b851040cd

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/c7a6b7f7665b11eb4e812cba4a46457b851040cd

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,27 +1,544 @@
...
+    """
+    extra_labels = {
+        "yelp.com/owner": "dre_mysql_working_hours",
+        "paasta.yelp.com/cluster": paasta_cluster,
+        "paasta.yelp.com/pool": paasta_pool,
+    }
+    return extra_labels
+
+
+def get_extra_env(paasta_cluster: str) -> List[Dict[str, str]]:
```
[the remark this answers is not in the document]
nemacysts: (as-is: making a change to the base classes wouldn't update things for vitess - which probably isn't the end of the world, but it will make things more manual since it'd be easy to forget to update the vitess code)

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,27 +1,544 @@
...
+}
+
+
+def get_updated_environment_variables(original_list: List[Dict], update_dict: Dict):
+    for env in original_list:
+        if env["name"] in update_dict:
+            env["value"] = update_dict[env["name"]]
+            del update_dict[env["name"]]
+    original_list.extend([{"name": k, "value": v} for k, v in update_dict.items()])
+    return original_list
```
nemacysts: what are your thoughts on storing kv dicts for env vars and having a `better_name_here_to_k8s()` function that took a dict and turned it into a list of dicts? would save us some complexity in copying/modifying dicts :)

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
nemacysts: and i guess another question is: as-is, i don't think there'd be a way for folks to add in an `env` key for these instances to add custom env vars: is that fine (or is it just late and i'm mistaken :p)?

paasta_tools/vitesscluster_tools.py
```diff
@@ -48,8 +564,102 @@ def __init__(
...
+            "images", load_system_paasta_config().get_vitess_images()
+        )
+        return {
+            "vtctld": vitess_images["vtctld_image"],
+            "vtadmin": vitess_images["vtadmin_image"],
+            "vtgate": vitess_images["vtgate_image"],
+            "vttablet": vitess_images["vttablet_image"],
+        }
+
+    def get_globalLockserver(self) -> Dict[str, str]:
```
nemacysts:
```suggestion
    def get_lockserver(self) -> Dict[str, str]:
```
or 
```suggestion
    def get_global_lockserver(self) -> Dict[str, str]:
```
would be more idiomatic

paasta_tools/vitesscluster_tools.py
```diff
@@ -48,8 +564,102 @@ def __init__(
...
+        paasta_cluster = self.config_dict.get("paasta_cluster", self.get_cluster())
+        cells = self.config_dict.get("cells")
+        region = self.config_dict.get("region")
+        vtgate_resources = self.config_dict.get("vtgate_resources")
+        return [
+            get_cell_config(cell, paasta_pool, paasta_cluster, region, vtgate_resources)
+            for cell in cells
+        ]
+
+    def get_vitessDashboard(self) -> Dict[str, str]:
```
nemacysts:
```suggestion
    def get_dashboard(self) -> Dict[str, str]:
```
or 
```suggestion
    def get_vitess_dashboard(self) -> Dict[str, str]:
```

paasta_tools/vitesscluster_tools.py
```diff
@@ -48,8 +564,102 @@ def __init__(
...
+        cells = self.config_dict.get("cells")
+        zk_address = self.config_dict.get("zk_address")
+        region = self.config_dict.get("region")
+        keyspaces = self.config_dict.get("keyspaces")
+        return get_keyspaces_config(
+            cells, keyspaces, paasta_pool, paasta_cluster, zk_address, region
+        )
+
+    def get_updateStrategy(self) -> Dict[str, str]:
+        return self.config_dict.get("updateStrategy", {"type": "Immediate"})
```
nemacysts, line 639:
```suggestion
    def get_update_strategy(self) -> Dict[str, str]:
        return self.config_dict.get("update_strategy", {"type": "Immediate"})
```
(this will also require updating the key in our config dict)

paasta_tools/vitesscluster_tools.py
```diff
@@ -48,8 +564,102 @@ def __init__(
...
+            "deploy_group": self.get_deploy_group(),
+            "healthcheck_cmd": self.get_healthcheck_cmd(),
+            "healthcheck_mode": self.get_healthcheck_mode(smartstack_config),
+            "healthcheck_grace_period_seconds": self.get_healthcheck_grace_period_seconds(),
+            "env": {
+                "OPERATOR_NAME": "vitess-operator",
+                "POD_NAME": "vitess-k8s",
+                "PS_OPERATOR_POD_NAME": "vitess-k8s",
+                "PS_OPERATOR_POD_NAMESPACE": KUBERNETES_NAMESPACE,
+                "WATCH_NAMESPACE": KUBERNETES_NAMESPACE,
```
nemacysts, line 659: i think we might want to hardcode that `get_namespace` return `KUBERNETES_NAMESPACE` - otherwise, these values would be incorrect

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
nemacysts: (actually, i think if many of these functions became helpers in `VitessDeploymentConfig` we could maybe also cut down the number of params to some of these functions)

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,27 +1,544 @@
...
+    paasta_pool: str,
+    paasta_cluster: str,
+    region: str,
+    vtgate_resources: Dict[str, str],
+) -> Dict[str, str]:
+    """
+    get vtgate config
+    """
+    replicas = vtgate_resources.get("replicas", 1)
+    requests = vtgate_resources.get("requests", {"cpu": "100m", "memory": "256Mi"})
```
nemacysts, line 218: we'll also probably want to figure out disk limits for these containers at some point

nemacysts: note: this cpu/mem code is also probably something that we'd want to handle as part of VitessDeploymentConfig so that we can add in paasta defaults - .1cpus is a pretty tiny default

we'll also probably want a way to configure the requests per-pod too since there's a couple of these that come with one paasta instance

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Is there a way we can ensure this?

nemacysts (commented):
i still need to go through this some more, but i figured i'd post a review up to where i got to today

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: @flopex ^ Any thoughts on this?

paasta_tools/utils.py
```diff
@@ -2701,6 +2701,20 @@ def get_spark_kubeconfig(self) -> str:
...
+        return self.config_dict.get("mysql_port_mappings", {})
+
+    def get_vitess_images(self) -> Dict:
+        return self.config_dict.get(
+            "vitess_images",
+            {
+                "vtctld_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
+                "vtgate_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
+                "vttablet_image": "docker-paasta.yelpcorp.com:443/vitess_base:v16.0.3",
+                "vtadmin_image": "docker-dev.yelpcorp.com/vtadmin:v16.0.3",
```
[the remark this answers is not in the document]
VinaySagarGonabavi: https://github.yelpcorp.com/docker-images/vtadmin/pull/4

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,27 +1,550 @@
...
+    """
+    extra_labels = {
+        "yelp.com/owner": "dre_mysql",
+        "paasta.yelp.com/cluster": paasta_cluster,
+        "paasta.yelp.com/pool": paasta_pool,
+    }
+    return extra_labels
+
+
+def get_extra_env(paasta_cluster: str) -> List[object]:
```
nemacysts, line 193: just wanted to link to some old comments i made that seem to have gotten marked as outdated (maybe from force-pushing?) about some of these get_extra_* functions: https://github.com/Yelp/paasta/pull/3889#discussion_r1633899105

paasta_tools/vitesscluster_tools.py
nemacysts: we'll probably want to add it in our flux repo 

EDIT: looks like it's already there as a common import :)

paasta_tools/vitesscluster_tools.py
```diff
@@ -48,8 +570,102 @@ def __init__(
...
+            "images", load_system_paasta_config().get_vitess_images()
+        )
+        return {
+            "vtctld": vitess_images["vtctld_image"],
+            "vtadmin": vitess_images["vtadmin_image"],
+            "vtgate": vitess_images["vtgate_image"],
+            "vttablet": vitess_images["vttablet_image"],
+        }
+
+    def get_globalLockserver(self) -> Dict[str, Dict[str, str]]:
```
nemacysts, line 587: this and `get_vitessDashboard()` should be camel_cased :)

paasta_tools/vitesscluster_tools.py
nemacysts: oh, i just saw read the sample soaconfigs used for this - they are per-pod :)

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,27 +1,550 @@
...
+    images: Dict[str, str]
+    cells: List[str]
+    region: str
+    zk_address: str
+    vtgate_resources: Dict[str, str]
+    vtadmin_resources: Dict[str, str]
+    vtctld_resources: Dict[str, str]
+    keyspaces: List[Dict[str, Any]]
+    updateStrategy: Dict[str, str]
+    paasta_cluster: str
```
nemacysts, line 547: hmm, i just saw in the example config that we're passing in a paasta_cluster key: we shouldn't need to allow this to be user-specified (iirc, we'll already have the right value for that in `self.cluster`)

nemacysts: (and we can probably generate the correct region in code rather than have that be user-specified - that will stop folks from making mistakes :))

paasta_tools/vitesscluster_tools.py
```diff
@@ -48,8 +570,102 @@ def __init__(
...
+            }
+        }
+
+    def get_cells(self) -> List[str]:
+        paasta_pool = self.get_pool()
+        paasta_cluster = self.config_dict.get("paasta_cluster", self.get_cluster())
+        cells = self.config_dict.get("cells")
+        region = self.config_dict.get("region")
+        vtgate_resources = self.config_dict.get("vtgate_resources")
+        return [  # type: ignore
```
nemacysts, line 603: mypy is correct here - the return type for this function does not match what will actually be returned :)

paasta_tools/vitesscluster_tools.py
```diff
@@ -72,21 +688,62 @@ def validate(
             return []
 
 
+def generate_vitess_instance_config(
+    instance_config: VitessDeploymentConfigDict,
+) -> VitessDeploymentConfigDict:
```
nemacysts, line 693: on another pass of this: i'm a little confused by what we're doing here: we're taking a `VitessDeploymentConfigDict` and returning a `VitessDeploymentConfigDict`? is this necessary? can we not fully construct the original `VitessDeploymentConfigDict` when that gets created?

paasta_tools/vitesscluster_tools.py
```diff
@@ -72,21 +688,62 @@ def validate(
...
+        "deploy_group": deploy_group,
+    }
+    return VitessDeploymentConfigDict(vitess_instance_config)  # type: ignore
+
+
 def load_vitess_instance_config(
     service: str,
     instance: str,
     cluster: str,
+    instance_type: str = "vitesscluster",
```
nemacysts, line 731: fwiw: it should be fine to keep hardcoding vitesscluster - vitess won't need to migrate to EKS since it'll only ever exist on eks clusters so i don't see the instance type name changing in the foreseeable future :)

nemacysts:
some additional comments from looking at the sample soaconfigs:
```
  cpus: 0.1  # what is this configuring?
  mem: 128
  ...
  vtgate_resources:
    ...
    requests:
      cpu: 100m  # should we make cpu/memory match the usual soaconfigs names (and units?)
      memory: 128Mi  # and if we make them match/drop the limits section, we can probably make these top-level keys :)
    limits:  # imo, we shouldn't really allow setting separate limits (for most other workloads, the mem/disk requests = the limit, and there's a cpu_burst_add (or similar) option that's added onto the request to set the cpu limit)
      memory: 128Mi
```

paasta_tools/utils.py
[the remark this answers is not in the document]
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/6cbc15067cdfc3754ceaf751a8cbe8d8ee749b7b

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/6cbc15067cdfc3754ceaf751a8cbe8d8ee749b7b

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/6cbc15067cdfc3754ceaf751a8cbe8d8ee749b7b

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,27 +1,550 @@
...
+    "vreplication_tablet_type": "REPLICA",
+    "keep_logs": "72h",
+    "enable-lag-throttler": "true",
+    "throttle_check_as_check_self": "true",
+    "db_charset": "utf8mb4",
+    "disable_active_reparents": "true",
+}
+
+
+def get_updated_environment_variables(
```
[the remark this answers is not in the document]
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/6cbc15067cdfc3754ceaf751a8cbe8d8ee749b7b

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/6cbc15067cdfc3754ceaf751a8cbe8d8ee749b7b

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/6cbc15067cdfc3754ceaf751a8cbe8d8ee749b7b

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/6cbc15067cdfc3754ceaf751a8cbe8d8ee749b7b

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/48ebe609e02377336248dcbb9bfed29a0d6213ae

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/48ebe609e02377336248dcbb9bfed29a0d6213ae

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/48ebe609e02377336248dcbb9bfed29a0d6213ae

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/48ebe609e02377336248dcbb9bfed29a0d6213ae

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/48ebe609e02377336248dcbb9bfed29a0d6213ae

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,27 +1,550 @@
...
+        if env["name"] in update_dict:
+            env["value"] = update_dict[env["name"]]
+            del update_dict[env["name"]]
+    original_list.extend([{"name": k, "value": v} for k, v in update_dict.items()])
+    return original_list
+
+
+def get_affinity_spec(
+    paasta_pool: str,
+) -> Dict[str, Dict[str, Dict[str, List[Dict[str, List[Dict[str, Sequence[str]]]]]]]]:
```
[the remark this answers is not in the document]
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/7650c62e4a6ccf188e1f73ea83701444a417c24a

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,27 +1,544 @@
...
+def get_updated_environment_variables(original_list: List[Dict], update_dict: Dict):
+    for env in original_list:
+        if env["name"] in update_dict:
+            env["value"] = update_dict[env["name"]]
+            del update_dict[env["name"]]
+    original_list.extend([{"name": k, "value": v} for k, v in update_dict.items()])
+    return original_list
+
+
+def get_affinity_spec(paasta_pool: str) -> Dict[str, Dict[str, List[str]]]:
```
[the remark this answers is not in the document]
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/7650c62e4a6ccf188e1f73ea83701444a417c24a

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: This still needs fixing, has default to compute-infra when used from parent class 
```
'extraLabels': {'yelp.com/owner': 'compute_infra_platform_experience' ...
```

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/4bcdd469f1e5b47fd6ff85f18d3cb77bbcd44203

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: There are no specific disk requirements yet for these pods and we can start with the default value, but it can easily be added in pod resources, and it should just work?

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: @flopex to confirm, but upon discussion it seems like the secret id and role id would not be modified when we rotate the secrets in vault periodically. So, there is no relevant paasta secret sync to be done as the vttablet and vtgate pods directly communicate with vault based on role id and secret id set before the first deployment in an ecosystem

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: https://github.com/Yelp/paasta/pull/3889/commits/700065a7d9b82a3facfefb9e6a2f79e508b96dd7

VinaySagarGonabavi:
> some additional comments from looking at the sample soaconfigs:
> 
> ```
>   cpus: 0.1  # what is this configuring?
>   mem: 128
>   ...
>   vtgate_resources:
>     ...
>     requests:
>       cpu: 100m  # should we make cpu/memory match the usual soaconfigs names (and units?)
>       memory: 128Mi  # and if we make them match/drop the limits section, we can probably make these top-level keys :)
>     limits:  # imo, we shouldn't really allow setting separate limits (for most other workloads, the mem/disk requests = the limit, and there's a cpu_burst_add (or similar) option that's added onto the request to set the cpu limit)
>       memory: 128Mi
> ```

The top level cpu, mem, disk etc., don't mean much as we care mostly about the vitess pods? (please correct me if I'm wrong). The problem with using same resource specifications for all pods is that there are some keyspaces/dbs where we might have to configure more. We can have few different yaml templates for resource specifications to reuse for keyspaces and help minimize the yaml size

nemacysts:
> The top level cpu, mem, disk etc., don't mean much as we care mostly about the vitess pods

if the top-level resources aren't going to be used, we can probably add a schema that prevents them from being added since they're otherwise likely to confuse folks in the future if they happen to set a value there (that said - we probably need to figure out what to do with the get_cpu/get_mem/get_disk functions - i wonder what we do for some of the other multi-pod operators?)

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,31 +1,581 @@
...
+
+
+class KeyspaceConfigDict(TypedDict, total=False):
+    durabilityPolicy: str
+    turndownPolicy: str
+    partitionings: List[Dict[str, PartitioningValueDict]]
+    name: str
+
+
+def get_updated_environment_variables(env_vars: Dict[str, Any]) -> List[Dict[str, Any]]:
```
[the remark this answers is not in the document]
nemacysts: in the future, we can also probably type the return value as something like `List[Union[KVEnvVar, ValueFromEnvVar]]` (and if a future yelper really wanted to go for extra credit, the parameter could probably also be something like `Dict[str, Union[str, ValueFromDict]]`) - but this is fine for now (future yelper: please pick better names than my suggestions if you can :p)

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,31 +1,581 @@
...
+    node_affinity: dict,
+) -> CellConfigDict:
+    """
+    get vtgate config
+    """
+    try:
+        replicas = int(vtgate_resources.get("replicas"))
+    except ValueError:
+        log.error("Invalid replicas value for vtgate")
+        replicas = 1
```
nemacysts, line 212: it might be worth adding a schema at `paasta_tools/cli/schemas/` so that we can ensure that the right values are in place (and that they're in the correct format (i.e., ints) :))

paasta_tools/vitesscluster_tools.py
nemacysts: ah true, i think we have a default limitrange on all namespaces that should automatically set things up with a 1GB disk allocation

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,31 +1,581 @@
...
+    labels.update({"tablet_type": tablet_type_label})
+
+    try:
+        type = load_system_paasta_config().get_vitess_tablet_pool_type_mapping()[
+            tablet_type
+        ]
+    except KeyError:
+        log.error(
+            f"Tablet type {tablet_type} not found in system paasta config vitess_tablet_pool_type_mapping"
+        )
```
nemacysts, line 381: should we exit early/default to another type if we run into this case? or does this `type` not being found not prevent us from creating usable pods?

paasta_tools/vitesscluster_tools.py
nemacysts: otherwise, 
```suggestion
    try:
        replicas = int(vtgate_resources.get("replicas"))
    except ValueError:
        log.error("Invalid replicas value for vtgate - defaulting to a single replica.")
        replicas = 1
```
will make it a little clearer to anyone looking at logs that this is something that was recoverable

(and we'd want to do this for the other error messages)

nemacysts: that said, i'd still encourage adding a schema so that this isn't actually possible to trigger :)

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,31 +1,581 @@
...
+                "mountPath": "/etc/init_db.sql",
+                "name": "keyspace-fake-init-script",
+                "readOnly": True,
+            },
+        ],
+        extraVolumes=[
+            {"name": "vault-secrets", "hostPath": {"path": "/nail/etc/vault/all_cas"}},
+            {
+                "name": "acls",
+                "hostPath": {"path": "/nail/srv/configs/vitess_keyspace_acls"},
```
nemacysts, line 423: hmm, almost missed this - this is probably going to break in surprising ways later on: `/nail/srv/configs` is a symlink that will have its target change pretty frequently (every srv-configs deployment)

```
[I] dev208-uswest1adevc:~> exa -la /nail/srv/                                                                                                                                                                                                                  
lrwxrwxrwx 26 root 26 Aug  2022 configs -> /nail/etc/srv-configs/main

[I] dev208-uswest1adevc:~> exa -la /nail/etc/srv-configs/main                                                                                                                                                                                                   
lrwxrwxrwx 49 root 13 Jun 15:07 /nail/etc/srv-configs/main -> /nail/etc/srv-configs/.client/../.main-1718316455
```

nemacysts: what we normally do here is that we mount `/nail/srv` (which is not a symlink) and thus avoid losing track of files/dirs when the inodes go away :)

which brings me to another question: i wonder if it's worth using the code that we use for services to add in the puppet-configured default paasta volume mounts (which includes easy-to-get-wrong mounts like the srv-configs one)

paasta_tools/vitesscluster_tools.py
```diff
@@ -48,8 +597,153 @@ def __init__(
...
+        for superregion_prefix in superregion_to_region_map:
+            if superregion.startswith(superregion_prefix):
+                region = superregion.replace(
+                    superregion_prefix, superregion_to_region_map[superregion_prefix]
+                )
+        if region is None:
+            log.error(
+                f"Region not found for superregion {superregion}. Check superregion_to_region_mapping in system paasta config"
+            )
+        return region
```
nemacysts, line 666: is it safe to continue if this happens? we're saying get_region will return a str, but it's really returning `Optional[str]`

paasta_tools/vitesscluster_tools.py
```diff
@@ -48,8 +597,153 @@ def __init__(
...
+
+    def get_vitess_node_affinity(self) -> dict:
+        paasta_pool = self.get_pool()
+        self.config_dict.update(
+            {
+                "node_selectors": {
+                    "yelp.com/pool": [paasta_pool],
+                }
+            }
+        )
```
nemacysts, line 644: hmm, we probably don't want to be mutating config_dict here - what's the context behind this?

nemacysts: if this is for `get_node_affinties()` - it doesn't actually require that you manually set a pool node_selector as the general usage of these affinities is for things like grouping/spreading services (or adding metadata about instance types and whatnot)

nemacysts: we also tend to use node selectors instead of affinities for pool selection since affinities don't show up in the bare kubectl describe output (they're only displayed if you change the output format to json/yaml)

paasta_tools/vitesscluster_tools.py
```diff
@@ -48,8 +534,143 @@ def __init__(
...
+        return {
+            "vtctld": vitess_images["vtctld_image"],
+            "vtadmin": vitess_images["vtadmin_image"],
+            "vtgate": vitess_images["vtgate_image"],
+            "vttablet": vitess_images["vttablet_image"],
+        }
+
+    def get_env_variables(self) -> List[Dict[str, Any]]:
+        # get all K8s container env vars and format their keys to camel case
```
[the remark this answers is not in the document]
VinaySagarGonabavi: We're only formatting the environment variables from parent class, KubernetesDeploymentConfig which are still in V1EnvVar form and when converted to dict with to_dict() will use snake casing for keys as reported in https://github.com/kubernetes-client/python/issues/390 and other similar issues. The workaround provided is to use Apiclient method sanitize_for_serialization to convert them back to the format processable by operator

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: This attaches a type of vttablet created, which is either externalmaster or externalreplica in our case. We can definitely make this an exception with a default value

paasta_tools/vitesscluster_tools.py
```diff
@@ -48,8 +534,143 @@ def __init__(
...
+        superregion_to_region_map = (
+            load_system_paasta_config().get_superregion_to_region_mapping()
+        )
+        for superregion_prefix in superregion_to_region_map:
+            if superregion.startswith(superregion_prefix):
+                return superregion.replace(
+                    superregion_prefix, superregion_to_region_map[superregion_prefix]
+                )
+
+        log.error(f"Invalid superregion {superregion}")
```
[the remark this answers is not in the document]
VinaySagarGonabavi: Exit early in case of this exception https://github.com/Yelp/paasta/pull/3889/commits/2f6ecc76417760dbd1cb424c3b61dabd5c83ea09

paasta_tools/vitesscluster_tools.py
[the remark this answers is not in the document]
VinaySagarGonabavi: Made this better typing wise in https://github.com/Yelp/paasta/pull/3889/commits/2f6ecc76417760dbd1cb424c3b61dabd5c83ea09

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: I'm willing to add schema to ensure we have specifications in the right type and avoid unexpected values to be passed. I was thinking of adding the whole json schema from kuberenetes-schema.json and add the relevant vitess properties and their schemas. Would that be a fine compromise?

Also, avoided hitting this mypy error which made me add int type conversion in the first place with https://github.com/Yelp/paasta/pull/3889/commits/2f6ecc76417760dbd1cb424c3b61dabd5c83ea09

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/2f6ecc76417760dbd1cb424c3b61dabd5c83ea09

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/2f6ecc76417760dbd1cb424c3b61dabd5c83ea09. Lmk if this looks good now

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Fixed in https://github.com/Yelp/paasta/pull/3889/commits/2f6ecc76417760dbd1cb424c3b61dabd5c83ea09

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Added `node_selectors` in the service metadata instead of mutating config_dict here https://github.com/Yelp/paasta/pull/3889/commits/2f6ecc76417760dbd1cb424c3b61dabd5c83ea09

VinaySagarGonabavi:
@jfongatyelp , @matt-ullmer , @alexanderjulianmartinez and I have sync'd today to go over the pending concerns related to the PR and have detailed the discussion notes in https://yelpwiki.yelpcorp.com/x/epctF. I'll move forward on this once I have a sign off from CI. Thanks for joining and helping proceed on this :)

jfongatyelp (approved):

paasta_tools/vitesscluster_tools.py
```diff
@@ -1,31 +1,602 @@
...
+    "dba_pool_size": "4",
+    "vreplication_heartbeat_update_interval": "60",
+    "vreplication_tablet_type": "REPLICA",
+    "keep_logs": "72h",
+    "enable-lag-throttler": "true",
+    "throttle_check_as_check_self": "true",
+    "db_charset": "utf8mb4",
+    "disable_active_reparents": "true",
+}
+
```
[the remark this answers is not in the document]
VinaySagarGonabavi: ++. Will be needing to make changes to add support for paasta status/logs in the coming days. Will make this a config instead along with those changes

PR #3891, opened 2024-06-08
User can override pod template

austinzh:
In the existing design, we modify pod-template even if user already specify it in spark config.
This remove any flexibility of user want to have their own template instead of using the one in default config lib.


nemacysts:
closing this for now, feel free to re-open if y'all are still looking to get this merged!

PR #3892, opened 2024-06-10
MLCOMPUTE-1203 | Configure Spark driver pod memory and cores based on Spark args

CaptainSame:


nemacysts:
@CaptainSame is this for the new soaconfigs config style?

CaptainSame:
> @CaptainSame is this for the new soaconfigs config style? if so, we probably need to figure out how to reconcile this with the usual way of specifying resources

Yes there was a case where `cpus` and `mem` were not specified in the soa config but `spark.driver.memory` and `spark.driver.cores` were specified and the Spark driver container was started with default resources which were quite low.

Do you think it makes sense to override `cpus` and `mem` values with `spark.driver..` values if both are specified?
OR
We take the maximum of both sets of values if both are specified?
OR
Ask the users to keep both the values equal and we just read the `cpus` and `mem` in soaconfigs like the status quo?

chi-yelp:
In paasta spark-run, people can use spark configs (`spark.driver.cores` and `spark.driver.memory`) to specify driver resources needed. I think it's still important to support doing it in this way. Also I'm a little worried if we allow to use paasta config's `cpus` and `mem` for spark jobs, people might get confused for whether they are for Spark driver or executors

paasta_tools/spark_tools.py
```diff
@@ -247,3 +248,23 @@ def get_spark_driver_monitoring_labels(
...
+    memory_bytes = 0.0
+    if mem:
+        if mem[-1] in MEM_MULTIPLIER:
+            memory_bytes = float(mem[:-1]) * MEM_MULTIPLIER[mem[-1]]
+        else:
+            try:
+                memory_bytes = float(mem)
+            except ValueError:
+                print(f"Unable to parse memory value {mem}")
+    memory_unit = memory_bytes / MEM_MULTIPLIER[unit]
```
nemacysts, line 269: it's probably worth doing some validation in `paasta validate` / the schema to ensure that this is correctly specified: right now, if there's an issue with the value provided, this will return 0

(also, we probably want to verify that the provided `unit` is valid too - otherwise the division line will throw an exception when indexing into MEM_MULTIPLIER)

paasta_tools/tron_tools.py
```diff
@@ -282,6 +282,32 @@ def __init__(
...
+    def get_cpus(self) -> float:
+        # set Spark driver pod CPU if it is specified by Spark arguments
+        cpus = 0.0
+        if (
+            self.action_spark_config
+            and "spark.driver.cores" in self.action_spark_config
+        ):
+            cpus = float(self.action_spark_config["spark.driver.cores"])
+        # use the soa config otherwise
+        return cpus or super().get_cpus()
```
nemacysts, line 296:
```suggestion
    def get_cpus(self) -> float:
        if (
            self.get_executor() == "spark"
            and "spark.driver.cores" in self.action_spark_config
        ):
            return self.action_spark_config["spark.driver.cores"]
        # NOTE: we fallback to this if there's no spark.driver.cores config to
        # use the paasta default
        return super().get_cpus()
```

nemacysts: that said, maybe we want a different default CPU for spark drivers? the paasta default for both cpu and memory might be too small for drivers :)

nemacysts: (also, imo we should add `spark.driver.cores` to the spark_args schema and enforce that it's a float there so that we don't need to remember to cast it in the python code :))

paasta_tools/tron_tools.py
```diff
@@ -282,6 +282,32 @@ def __init__(
...
+        if (
+            self.action_spark_config
+            and "spark.driver.memory" in self.action_spark_config
+        ):
+            # need to set mem in MB based on tron schema
+            mem_mb = spark_tools.get_spark_memory_in_unit(
+                self.action_spark_config["spark.driver.memory"], "m"
+            )
+        # use the soa config otherwise
+        return mem_mb or super().get_mem()
```
nemacysts, line 310:
```suggestion
    def get_mem(self) -> float:
        if (
            self.get_executor() == "spark"
            and "spark.driver.memory" in self.action_spark_config
        ):
            return spark_tools.get_spark_memory_in_unit(
                self.action_spark_config["spark.driver.memory"], "m"
            )
        # NOTE: we fallback to this if there's no spark.driver.memory config to
        # use the paasta default
        return super().get_mem()
```

paasta_tools/spark_tools.py
nemacysts: (oh, i see - we hardcode the `unit` when we call this)

paasta_tools/spark_tools.py
```diff
@@ -247,3 +248,23 @@ def get_spark_driver_monitoring_labels(
         "spark.yelp.com/driver_ui_port": ui_port_str,
     }
     return labels
+
+
+def get_spark_memory_in_unit(mem: str, unit: str) -> float:
```
nemacysts, line 253:
```suggestion
def get_spark_memory_in_unit(mem: str, unit: Literal["k", "m", "g", "t") -> float:
```

paasta_tools/tron_tools.py
nemacysts: (we probably also want to do the same thing i mentioned above re: making the schema ensure that `spark.driver.memory` is a number :))

paasta_tools/tron_tools.py
```diff
@@ -912,6 +951,9 @@ def format_tron_action_dict(action_config: TronActionConfig):
         result["executor"] = EXECUTOR_NAME_TO_TRON_EXECUTOR_TYPE.get(
             executor, "kubernetes"
         )
+        if executor == "spark":
+            # inject spark configs to the original spark-submit command
+            action_config.action_spark_config = action_config.build_spark_config()
```
nemacysts, line 956: just curious: how come we had to move this up?

paasta_tools/tron_tools.py
nemacysts: and i think we might need to make sure this returns an int? i don't think we allow for fractional `mem` values

paasta_tools/tron_tools.py
CaptainSame: `get_env` was called above this, before the spark config had been built and set, resulting in incorrect values in env variables

paasta_tools/tron_tools.py
CaptainSame: I would rather handle enforce the default for `spark.driver.cores` and `spark.driver.memory` in service_configuration_lib than using another default value here.
Even if we enforce the float in schema, it passes through operations in service_configuration_lib and everything is returned as a string

paasta_tools/tron_tools.py
CaptainSame: `spark.driver.memory` doesn't have to be a number. We allow values allowed by Spark, which are JVM memory strings. I saw that the signature of the original `get_mem` method returns a float so I kept the same here
Should I change it to return int at both the places??

paasta_tools/spark_tools.py
CaptainSame: Changed it to return a default 2 GB memory

paasta_tools/tron_tools.py
```diff
@@ -605,6 +627,20 @@ def validate(self):
...
+            if "cpus" in self.config_dict:
+                error_msgs.append(
+                    f"{self.get_job_name()}.{self.get_action_name()} is a Spark job. `cpus` config is not allowed. "
+                    f"Please specify the driver cores using `spark.driver.cores`."
+                )
+            if "mem" in self.config_dict:
+                error_msgs.append(
+                    f"{self.get_job_name()}.{self.get_action_name()} is a Spark job. `mem` config is not allowed. "
+                    f"Please specify the driver memory using `spark.driver.memory`."
+                )
```
chi-yelp, line 643: [optional] we can can add checks in the [yelpsoa_configs' pre-commit hook](https://sourcegraph.yelpcorp.com/sysgit/yelpsoa-configs/-/blob/tron/validate_spark_tronfig), so users can notice the problem earlier before the next run

chi-yelp (approved):

paasta_tools/tron_tools.py
nemacysts: it's probably worth adding this to `paasta validate` now - it'd be essentially dropping this same code in there :)

paasta_tools/tron_tools.py
nemacysts: re: `spark.driver.memory`: i don't think that would work with the code as-is since we're hardcoding that the unit is `"m"`

that said, we can still add some schema/`paasta validate` validation so that we can ensure that we're not getting junk data from users 

re: changing the signature: we can probably leave this for another time - i see that flink is using floats for things like `mem: 1.5Gi`

paasta_tools/tron_tools.py
nemacysts: we need this to return the resources that are actually used tho (either user-specified or the defaults that are used) - this can call `service_configuration_lib` code if necessary :)

re: s_c_l: i think it's also worth slowly refactoring that code - it's sorta painful to work with when everything in that library is stringly-typed :)

paasta_tools/tron_tools.py
nemacysts: oh, re: `unit="m"` - i re-read the code with fresh eyes today and realized i'd been misreading it, the code is converting the code to mb :p

paasta_tools/tron_tools.py
nemacysts: oh actually, this is already being called by `paasta validate` :)

paasta_tools/spark_tools.py
```diff
@@ -247,3 +249,24 @@ def get_spark_driver_monitoring_labels(
...
+    """
+    Converts Spark memory to the desired unit.
+    mem is the same format as JVM memory strings: just number or number followed by 'k', 'm', 'g' or 't'.
+    unit can be 'k', 'm', 'g' or 't'.
+    Returns memory as a float converted to the desired unit.
+    """
+    memory_bytes = 0.0
+    if mem:
+        if mem[-1] in MEM_MULTIPLIER:
+            memory_bytes = float(mem[:-1]) * MEM_MULTIPLIER[mem[-1]]
```
nemacysts, line 264: if we don't add validation for the  user-provided value here now, then this needs to be wrapped in a try-except as well (since otherwise this will crash if a user puts in "lolk" or `1B` as a `spark.driver.memory` value)

paasta_tools/tron_tools.py
nemacysts: that said, all the pods i'm spot-checking only have integer `PAASTA_RESOURCE_MEM` values and the flink pods that are setting fractional `mem` values don't add that env var - so casting to an int here is probably a good idea to keep the data looking the same (mypy shouldn't complain about returning an int from a function typed to return a float from my testing)

PR #3893, opened 2024-06-13
Add a bulkdata flag

timmow:
If this is set to false, do not mount the /nail/bulkdata volume

Tested by
* following instructions on [this page](https://yelpwiki.yelpcorp.com/pages/viewpage.action?pageId=234312044)
* editing `etc_paasta_playground/volumes.json` and adding the bulkdata volume
* running `docker ps | grep kind` to find the worker node container ids
* using these ids, running `docker exec -t -i <id> bash` and creating the /nail/bulkdata directory on the worker nodes
* editing `config_playground/compute-infra-test-service/kubernetes-kind-tmower-k8s-test.yaml` and setting bulkdata: false

* verifying with `KUBECONFIG=./.paasta-kube-config kubectl get deployment -n paastasvc-compute-infra-test-service -oyaml` the /nail/bulkdata volume was not added

I'm not entirely sure why `black` has reformatted the `kubernetes_tools.py` file

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -736,9 +736,9 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             instance=self.instance,
             cluster=self.cluster,
             config_dict=self.config_dict.copy(),
-            branch_dict=self.branch_dict.copy()
-            if self.branch_dict is not None
-            else None,
+            branch_dict=(
+                self.branch_dict.copy() if self.branch_dict is not None else None
+            ),
             soa_dir=self.soa_dir,
         )
 
@@ -1960,7 +1960,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             supported_storage_classes = (
                 system_paasta_config.get_supported_storage_classes()
             )
-        except (PaastaNotConfiguredError):
+        except PaastaNotConfiguredError:
             log.warning("No PaaSTA configuration was found, returning default value")
             supported_storage_classes = []
         storage_class_name = volume.get("storage_class_name", "ebs")
@@ -2144,6 +2144,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         docker_volumes = self.get_volumes(
             system_volumes=system_paasta_config.get_volumes()
         )
+
         hacheck_sidecar_volumes = system_paasta_config.get_hacheck_sidecar_volumes()
         has_routable_ip = self.has_routable_ip(
             service_namespace_config, system_paasta_config
@@ -2406,11 +2407,11 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             preferred_terms = None
 
         return V1NodeAffinity(
-            required_during_scheduling_ignored_during_execution=V1NodeSelector(
-                node_selector_terms=[required_term]
-            )
-            if required_term
-            else None,
+            required_during_scheduling_ignored_during_execution=(
+                V1NodeSelector(node_selector_terms=[required_term])
+                if required_term
+                else None
+            ),
             preferred_during_scheduling_ignored_during_execution=preferred_terms,
         )
 
@@ -2701,7 +2702,7 @@ def get_kubernetes_services_running_here_for_nerve(
                     }
                 nerve_dict["weight"] = kubernetes_service.weight
                 nerve_list.append((registration, nerve_dict))
-        except (KeyError):
+        except KeyError:
             continue  # SOA configs got deleted for this app, it'll get cleaned up
 
     return nerve_list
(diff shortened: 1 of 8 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -2144,6 +2150,12 @@ def get_pod_template_spec(
         docker_volumes = self.get_volumes(
             system_volumes=system_paasta_config.get_volumes()
         )
+        if not self.has_bulkdata():
```
Rob-Johnson, line 2153: rather than excluding the volume from the default volumes, can you instead remove `/nail/bulkdata` from the default volumes, and then add when `has_bulkdata` is `True`? It just removes some of the surprise from what `default_volumes` really does

Rob-Johnson (commented):
please all add docs for the new yelpsoa-configs key here https://github.com/Yelp/paasta/blob/master/docs/source/yelpsoa_configs.rst#L290 - these docs aren't in a good state, but it'd be good to stop the rot if we can

paasta_tools/kubernetes_tools.py
nemacysts: i think we'd need to roll this out first before removing `/nail/bulkdata` from the default mounts - otherwise we'll potentially be removing the bulkdata mount from services that need it but are on a cluster where this paasta version has not yet rolled out

nemacysts (commented):
imo, something like `include_bulkdata`/`uses_bulkdata`/etc would be a more intuitive name for the config option - but i don't have particularly strong feelings about it if y'all prefer just `bulkdata`

paasta_tools/kubernetes_tools.py
timmow: I've changed this and added details in the tech spec for how we should roll this out

paasta_tools/utils.py
```diff
@@ -997,8 +998,24 @@ def get_net(self) -> str:
...
+        self,
+    ) -> bool:
+        return self.config_dict.get("uses_bulkdata", True)
+
     def get_volumes(self, system_volumes: Sequence[DockerVolume]) -> List[DockerVolume]:
         volumes = list(system_volumes) + list(self.get_extra_volumes())
+        # add a dictionary to docker_volumes with a hostPath of /nail/bulkdata
+        # if the service uses bulkdata
+        if self.has_bulkdata():
+            # the volume has a mode of RO
```
nemacysts, line 1011: nit:
```suggestion
        # we used to add bulkdata as a default mount - but as part of the 
        # effort to deprecate the entire system, we're swapping to an opt-in
        # model so that we can shrink the blast radius of any changes
        if self.has_bulkdata():
            # bulkdata is mounted RO as the data is produced by another 
            # system and we want to ensure that there are no inadvertent 
            # changes by misbehaved code
```
is likely a better set of comments since they add some context as to what's going on :)

paasta_tools/utils.py
```diff
@@ -997,8 +998,24 @@ def get_net(self) -> str:
...
+            volumes = volumes + [
+                {
+                    "hostPath": "/nail/bulkdata",
+                    "containerPath": "/nail/bulkdata",
+                    "mode": "RO",
+                }
+            ]
```
nemacysts, line 1018: nit:
```suggestion
            volumes.append(
                {
                    "hostPath": "/nail/bulkdata",
                    "containerPath": "/nail/bulkdata",
                    "mode": "RO",
                }
            )
```
is slightly preferable since it won't cause an entirely new list to be created (`append` will resize the existing list, whereas `+` results in a new list being created)

e.g.:
```
>>> test = []
>>> id(test)
139872597201776
>>> test = test + [{1:1}]
>>> id(test)
139872597231328
```
vs
```
>>> test = []
>>> id(test)
139872597201776
>>> test.append({1:1})
>>> id(test)
139872597201776
```

nemacysts (approved):
lgtm - just two minor nits (that are ignorable, but would be nice to handle)

Rob-Johnson (changes requested):
please add a test - realise that there is nothing there atm, but something is better than nothing.

paasta_tools/utils.py
```diff
@@ -997,8 +998,24 @@ def get_net(self) -> str:
         """
         return self.config_dict.get("net", "bridge")
 
+    def has_bulkdata(
+        self,
+    ) -> bool:
+        return self.config_dict.get("uses_bulkdata", True)
+
     def get_volumes(self, system_volumes: Sequence[DockerVolume]) -> List[DockerVolume]:
```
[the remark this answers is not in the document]
timmow: I've added a test that adding the uses_bulkdata flag adds the bulkdata volume. I've had to fix other failing tests by adding the uses_bulkdata flag set to False to prevent the bulkdata volume being added, so I think there is coverage of this method?

Rob-Johnson (approved):

nemacysts (approved):

PR #3902, opened 2024-06-26
Add paasta status for vitesscluster and mount /nail/etc/srv-configs to VTTablets

VinaySagarGonabavi:
Testing setup: 
- Apply the diff https://fluffy.yelpcorp.com/i/GZJrMCXcZfc5R2FTNsSggFCVcbXmZXwQ.html
- make setup-kubernetes-cr, apply CRDs from flux in the namespace, rerun make target and wait till operator and other pods are pending/created
- make playground-api to get the PaaSTA API server running with changes made in Makefile

Testing output:
Screenshot 2024-06-25 at 11 49 30 PM

VitessClusterStatus (https://github.com/planetscale/vitess-operator/blob/main/docs/api.md#vitessclusterstatus) and its fields used as a reference for TypedDicts added


matt-ullmer (approved):

jfongatyelp (approved):
As most changes look like they'll only impact vitess, seems safe from my perspective. 👍

paasta_tools/cli/cmds/status.py
```diff
@@ -1878,6 +1882,250 @@ def print_kafka_status(
     return 0
 
 
+class EtcdLockServerStatus(TypedDict, total=False):
+    observedGeneration: int
+    available: str
+    clientServiceName: str
```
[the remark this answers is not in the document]
VinaySagarGonabavi: We're using only zk implementation of lock server, but the status returns this field and hence created to accommodate it. This is expected to be unused and return an empty value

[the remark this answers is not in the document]
nemacysts: the typeddicts are non-total so we could always leave them out :)

paasta_tools/cli/cmds/status.py
```diff
@@ -1878,6 +1882,250 @@ def print_kafka_status(
...
+    vitess_status: Mapping[str, Any],
+    verbose: int = 0,
+) -> int:
+    tab = "    "
+    indent = 1
+
+    status: VitessClusterStatus = vitess_status.get("status")
+    if status is None:
+        output.append(
+            PaastaColors.red("indent * tab + Vitess cluster is not available yet")
```
nemacysts, line 1954: should this be something like: `PaastaColors.red(indent * tab + "Vitess cluster is not available yet")`?

paasta_tools/cli/cmds/status.py
```diff
@@ -1878,6 +1882,250 @@ def print_kafka_status(
...
+    output.append(indent * tab + "Vitess Cluster:")
+    indent += 1
+
+    output.append(
+        indent * tab
+        + "Observed Generation: "
+        + str(status.get("observedGeneration", 0))
+    )
+    output.append(
+        indent * tab + "Gateway Service Name: " + status.get("gatewayServiceName", "")
```
nemacysts, line 1967: not sure if this can happen, but might be nice to have a default like `UNKNOWN` or something to make it clearer that something is missing :)

paasta_tools/cli/cmds/status.py
```diff
@@ -1878,6 +1882,250 @@ def print_kafka_status(
...
+    )
+
+    output.append(indent * tab + "Cells:")
+    indent += 1
+    cells: Dict[str, VitessClusterCellStatus] = status.get("cells")
+    if not cells:
+        output.append(
+            indent * tab + "Cells: " + PaastaColors.red("No cell status available")
+        )
+        return 0
```
nemacysts, line 1977: just to confirm: for each of these sections, we can't proceed with the subsequent sections if the current one is unavailable?

paasta_tools/cli/cmds/status.py
```diff
@@ -1878,6 +1882,250 @@ def print_kafka_status(
...
+    gatewayServiceName: str
+    vitessDashboard: VitessDashboardStatus
+    cells: Dict[str, VitessClusterCellStatus]
+    keyspaces: Dict[str, VitessClusterKeyspaceStatus]
+    vtadmin: VTAdminStatus
+    orphanedCells: Dict[str, OrphanStatus]
+    orphanedKeyspaces: Dict[str, OrphanStatus]
+
+
+def print_vitess_status(
```
nemacysts, line 1940: might be a useful refactor in the future to break each of the sections into their own function so that `print_vitess_status()` isn't so long

nemacysts (commented):
sorry - was busy yesterday :)

PR #3903, opened 2024-06-27
Attempt #2 MLCOMPUTE-1203 | Configure Spark driver pod memory and cores based on Spark args

CaptainSame:
Changes from #3892 
+
DAR-2360 | pass `for_validation` flag correctly for paasta validate flows


paasta_tools/tron_tools.py
```diff
@@ -884,6 +922,12 @@ def format_tron_action_dict(action_config: TronActionConfig):
     :param action_config: TronActionConfig
     """
     executor = action_config.get_executor()
+    # not building the Spark config in TronActionConfig constructor since it needs a KUBECONFIG file and
+    # fails in yelpsoa configs repo validation while getting the action configs.
+    if executor == "spark":
+        # build the complete Spark configuration
+        # TODO: add conditional check for Spark specific commands spark-submit, pyspark etc ?
+        action_config.action_spark_config = action_config.build_spark_config()
```
nemacysts, line 930: can we instead use the `for_validation` flag? doing things this way will make interacting with these action configs in other scripts painful as folks will need to know that they need to manually call `build_spark_config()`

chi-yelp (approved):

chi-yelp:
Also verified with:
```
./.tox/py38-linux/bin/paasta validate -y <soa_configs_dir> -s spark
```

paasta_tools/cli/utils.py
```diff
@@ -968,6 +971,7 @@ def get_instance_configs_for_service(
                     cluster=cluster,
                     soa_dir=soa_dir,
                     load_deployments=False,
+                    for_validation=False,
```
nemacysts, line 974: we should be fine leaving this call alone - `False` is already the default

nemacysts: (we really just need to pass down `for_validation` where necessary)

nemacysts: (from chatting in gmeet: i think that this is mypy not understanding that this value has a default based on how we type the function signature)

nemacysts: we can fix this (and the load_deployments one) by using `DefaultNamedArg` instead of `NamedArg` - but we/I can do that in a follow-up to not extra changes to this PR

nemacysts:
lgtm - probably want someone from ML Compute to re-ship since i added a couple commits after working with @EvanKrall to refactor things a bit :)

PR #3904, opened 2024-06-27
Adding containerd compatability to oom_logger - COMPINFRA-3947

EmanElsaban:
Adds containerd support to oom_logger in paasta_tools. 

[Testing] 
- Echoed some log line to match the regex to test ``oom_regex_kubernetes`` and ``oom_regex_kubernetes_structured``. We see the log in the stream. This verifies that the Containerd Get request is doing the right thing and that we are getting the passed environment variables https://fluffy.yelpcorp.com/i/CMPWkJ7Fczg6ZTKwQJg8BZBqwTF4BkM9.html 
- looked at the error messages produced from spinning a cits instance in kubestage on containerd pool then exec in the container python and triggered an oom. The error messages: https://fluffy.yelpcorp.com/i/D8HgrgvMj9Zmt4h9qXmv5HzJVmkWdFSn.html added a regex for the message 
```
Jul  4 14:03:23 10-40-11-245-uswest1adevc kernel: : [ 7195.442928] oom-kill:constraint=CONSTRAINT_MEMCG,nodemask=(null),cpuset=cri-containerd-e216d2f1e6c625d363c71edb6b3cbab5a9e1b447641b61028d0b94b077adf27c.scope,mems_allowed=0,oom_memcg=/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod08768c36_163c_40e5_8e49_09cf42ff5046.slice,task_memcg=/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod08768c36_163c_40e5_8e49_09cf42ff5046.slice/cri-containerd-e216d2f1e6c625d363c71edb6b3cbab5a9e1b447641b61028d0b94b077adf27c.scope,task=python3,pid=485850,uid=33
``` 
which is called ``oom_regex_kubernetes_containerd_systemd_cgroup`` and tested it matched:
https://fluffy.yelpcorp.com/i/zTVqjL76X4tQDRG4wbpMdcX83BcxP7R8.html 
### NOTE
we need to make sure we merge the paasta pr first and have it rolled out everywhere before merging the [puppet pr](https://github.yelpcorp.com/sysgit/puppet/pull/12727) otherwise syslog wont send the logs as we have seen in testing where we had to first install the paasta debian package before running the d-branch to have syslog send the logs to scribereader


paasta_tools/monitoring/kill_orphaned_docker_containers.py
nemacysts, line 1: note: we'll want to remove any references of this from puppet if we haven't already

paasta_tools/oom_logger.py
```diff
@@ -76,6 +81,16 @@
 )
 
 
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description="paasta_oom_logger")
+    parser.add_argument(
+        "--containerd",
+        action="store_true",
+        help="Use containerd to inspect containers",
```
nemacysts, line 89: nit:
```suggestion
        help="Use containerd to inspect containers, otherwise use docker",
```

paasta_tools/oom_logger.py
```diff
@@ -136,11 +151,15 @@ def capture_oom_events_from_stdin():
                 break
 
 
-def get_container_env_as_dict(docker_inspect):
+def get_container_env_as_dict(is_cri_containerd: bool, container_inspect):
```
nemacysts, line 154: nit:
```suggestion
def get_container_env_as_dict(is_cri_containerd: bool, container_inspect: dict) -> Dict[str, str]:
```

comment: typing `container_inspect` is probably a PITA if the containerd/docker libraries don't have a good TypedDict/dataclass/etc that we can use - maybe once we're only using containerd we can have a non-total TypedDict with the bits that we do use tho :p

paasta_tools/oom_logger.py
```diff
@@ -209,18 +228,32 @@ def send_sfx_event(service, instance, cluster):
...
 
+def get_containerd_container(
+    is_cri_containerd: bool, container_id: str
+) -> containers_pb2.Container:
+    with grpc.insecure_channel("unix:///run/containerd/containerd.sock") as channel:
+        containersv1 = containers_pb2_grpc.ContainersStub(channel)
+        if is_cri_containerd:
+            namespace = "k8s.io"
+        else:
+            namespace = "moby"
```
nemacysts, line 239: hmm, could we use the containerd library for both cases? if so, we could remove the conditional in `main()` :)

otherwise, we can just assume that if this method is being called we want the k8s namespace

paasta_tools/monitoring/kill_orphaned_docker_containers.py
jfongatyelp: We'll also need to remove this from setup.py and debian/paasta-tools.links I believe.

That said, I don't think you'll need to do anything in puppet -- when I looked through for references to this script I found references to a script that had a similar name, but this specific paasta version was never actually referenced/used.

paasta_tools/oom_logger.py
```diff
@@ -76,6 +81,16 @@
...
+        "--containerd",
+        action="store_true",
+        help="Use containerd to inspect containers",
+    )
+    return parser.parse_args()
+
+
 def capture_oom_events_from_stdin():
     process_name_regex = re.compile(
         r"^\d+\s[a-zA-Z0-9\-]+\s.*\]\s(.+)\sinvoked\soom-killer:"
```
jfongatyelp, line 96: I think we'll need to do some experimentation w/ an actual oomkilled pod on nodes w/ containerd to verify what format the logs look like, and if we need to update any of these regexes.

paasta_tools/oom_logger.py
```diff
@@ -229,11 +262,18 @@ def main():
...
+            # then we're using containerd to inspect containers
+            container_info = get_containerd_container(args.containerd, container_id)
+            container_spec_raw = container_info.spec.value.decode("utf-8")
+            container_inspect = json.loads(container_spec_raw)
+        else:
+            # we're using docker to inspect containers
+            try:
+                container_inspect = client.inspect_container(resource_id=container_id)
+            except (APIError):
+                continue
```
jfongatyelp, line 275: We also probably want to mirror this try/except since containerd is likely to hit similar temporary issues (that said, it would probably be nice to log _something_ if we're dropping oom syslog messages and not processing them to paasta_oom_events here!)

paasta_tools/oom_logger.py
EmanElsaban: yea, I was thinking of also just assuming if we call this function then we are using containerd but I wasn't sure if we  would still need the moby ns

paasta_tools/oom_logger.py
```diff
@@ -136,11 +162,17 @@ def capture_oom_events_from_stdin():
...
     env_vars = {}
-    config = docker_inspect.get("Config")
-    if config is not None:
+    if is_cri_containerd:
+        config = container_inspect.get("process")
+        env = config.get("env", [])
+    else:
+        config = container_inspect.get("Config")
         env = config.get("Env", [])
+    if config is not None:
```
jfongatyelp, line 175: Technically this `if config is not None` should never be hit if we've got that far; if `config` was none we'd run into an exception when reaching into it while trying to get the `env` in each of the cases above.
Maybe this should be more like: 

```suggestion
    if is_cri_containerd:
        config = container_inspect.get("process")
        env_key = "env"
    else:
        config = container_inspect.get("Config")
        env_key = "Env"
    if config is not None:
        env = config.get(env_key, [])
        ...
```

jfongatyelp (approved):
python dependencies are pain 🫨

paasta_tools/oom_logger.py
```diff
@@ -115,6 +150,8 @@ def capture_oom_events_from_stdin():
         oom_regex_kubernetes,
         oom_regex_kubernetes_structured,
         oom_regex_kubernetes_systemd_cgroup,
+        oom_regex_kubernetes_containerd_systemd_cgroup,
+        oom_regex_kubernetes_containerd_systemd_cgroup_structured,
     ]
```
nemacysts, line 155: at some point we should probably audit all of these + remove unused ones + comment the remaining ones

from chatting in slack, it seems like we're not quite sure yet why sometimes we're only seeing the structured oomkill loglines and other times the unstructured one - but it does seem a bit weird that we can't use a single regex for containerd pods (or that in the past we needed 3 regexes for dockershim pods!)

paasta_tools/oom_logger.py
```diff
@@ -136,11 +173,18 @@ def capture_oom_events_from_stdin():
                 break
 
 
-def get_container_env_as_dict(docker_inspect):
+def get_container_env_as_dict(
+    is_cri_containerd: bool, container_inspect: dict
```
nemacysts, line 177: might be nice to have some TypedDicts for the container_inspect dict - but that'll be a little annoying since you'd need a Union of two TypedDicts and you'd need to add some casts in the the code below :p

tests/test_oom_logger.py
```diff
@@ -344,14 +345,21 @@ def test_send_sfx_event(mock_get_instance_config):
         assert mock_meteorite.create_counter.return_value.count.call_count == 1
 
 
+@patch("paasta_tools.oom_logger.argparse", autospec=True)
+def test_parse_args(mock_argparse):
+    assert parse_args() == mock_argparse.ArgumentParser.return_value.parse_args()
```
nemacysts, line 350: i think it's fine to leave this untested since we're not really checking for much here :)

tests/test_oom_logger.py
```diff
@@ -365,6 +373,7 @@ def test_main(
 ):
 
     mock_sys_stdin.readline.side_effect = sys_stdin
+    mock_parse_args.return_value.containerd = False
```
nemacysts, line 376: can we add some tests that exercise the new regexes/set containerd=True?

tox.ini
```diff
@@ -12,6 +12,7 @@ passenv = SSH_AUTH_SOCK PAASTA_ENV DOCKER_HOST CI
 setenv =
     TZ = UTC
 deps =
+    --only-binary=grpcio
```
nemacysts, line 15: i wonder if there's a way for us to set this in the requirements file so that we only need to specify this once - i was going to suggest adding a comment as to why this is here would be nice (i.e., we have wheels readily available internally but the same is not true in public pypi), but we'd need to copy-paste that quite a bit :p

tox.ini
```diff
@@ -62,7 +65,10 @@ commands =
 
 [testenv:tests-yelpy]
 envdir = .tox/py38-linux/
+setenv =
+    PIP_INDEX_URL = http://169.254.255.254:20641/simple/
```
nemacysts, line 69: we should probably refactor our env handling at some point so that this isn't required here, but that's probably not something we wanna tackle in this PR

tox.ini
nemacysts: oh nice, it is possible: https://pip.pypa.io/en/stable/reference/requirements-file-format/#global-options

if it doesn't break requirements-tools (i.e., check-requirements) to do this, that'd make this diff smaller (we wouldn't have to add this --only-binary option to all the toxenvs nor to the dockerfiles!) and it would let us add a comment explaining why we're singling out grpcio

nemacysts (commented):
we should add some tests for the new functionality, but otherwise looks good :)

tox.ini
nemacysts: (that said, also fine as a followup)

tests/test_oom_logger.py
```diff
@@ -178,6 +212,20 @@ def docker_inspect():
...
 
+@pytest.fixture
+def containerd_inspect():
+    return {
+        "process": {
+            "env": [
+                "PAASTA_SERVICE=fake_service",
+                "PAASTA_INSTANCE=fake_instance",
+                "PAASTA_RESOURCE_MEM=512",
+                "MESOS_CONTAINER_NAME=mesos-a04c14a6-83ea-4047-a802-92b850b1624e",
```
nemacysts, line 223: a more representative test would probably be to remove this env var
```suggestion
```
since we no longer actually set this :)

tests/test_oom_logger.py
```diff
@@ -193,6 +241,21 @@ def log_line():
...
+def log_line_containerd():
+    return LogLine(
+        timestamp=1720128512,
+        hostname="dev208-uswest1adevc",
+        container_id="e216d2f1e6c625d363c71edb6b3cbab5a9e1b447641b61028d0b94b077adf27c",
+        cluster="fake_cluster",
+        service="fake_service",
+        instance="fake_instance",
+        process_name="python3",
+        mesos_container_id="mesos-a04c14a6-83ea-4047-a802-92b850b1624e",
```
nemacysts, line 254: same here, but 
```suggestion
        mesos_container_id="mesos-null",
```
instead of deleting this line

tests/test_oom_logger.py
```diff
@@ -193,6 +241,21 @@ def log_line():
     )
 
 
+@pytest.fixture
+def log_line_containerd():
+    return LogLine(
+        timestamp=1720128512,
+        hostname="dev208-uswest1adevc",
```
nemacysts, line 248: just to make sure that there's nothing in this test that depends on where we run it - can we re-use the hostname from previous tests? i.e.,, 
```suggestion
        hostname="dev37-devc",
```

nemacysts (commented):
minor nits, but looks good otherwise!

nemacysts (approved):

PR #3908, opened 2024-06-29
Add NATS tailing for vector-logs reader

jvperrin:
I think this is fully ready to go! I've tested this locally with commands like `python paasta_tools/cli/cli.py logs -s example_happyhour -c pnw-devc -i main -f` or `python paasta_tools/cli/cli.py logs -s marley -c pnw-devc -i marley2 -f`

This does depend on some changes on the vector side of things (in `splunk/vector-logs`) in order to transform logs there so that the correct fields are present so that the JSON that NATS returns doesn't need to be parsed, changed a bit, and then re-encoded again. I won't merge this until that work is done and some more testing can be performed.

In order to test I've also been modifying https://github.com/Yelp/paasta/blob/4b465c299882784f01faf4f57ccb116ca7fa7fe6/paasta_tools/utils.py#L2348-L2352 to return `pnw-devc` and not just `kubestage` so that I can test locally without enabling this tailing for everyone actually using pnw-devc. Merging this should essentially be a noop for that reason, since this isn't enabled anywhere besides kubestage, and tailing will also only have infra set up in pnw-devc for the time being until it is rolled out to more locations.

### DIFF
--- a/paasta_tools/cli/cmds/logs.py
+++ b/paasta_tools/cli/cmds/logs.py
@@ -1171,16 +1171,16 @@ class VectorLogsReader(LogReader):
     SUPPORTS_TAILING = True
     SUPPORTS_TIME = True
 
-    # def __init__(self, cluster_map: Mapping[str, Any], nats_endpoint_map: Mapping[str, Any]) -> None:
-    def __init__(self, cluster_map: Mapping[str, Any]) -> None:
+    def __init__(
+        self, cluster_map: Mapping[str, Any], nats_endpoint_map: Mapping[str, Any]
+    ) -> None:
         super().__init__()
 
         if S3LogsReader is None:
             raise Exception("yelp_clog package must be available to use S3LogsReader")
 
         self.cluster_map = cluster_map
-        # self.nats_endpoint_map = nats_endpoint_map
-        self.nats_endpoint_map = {"pnw-devc": "nats-logs.infra.uswest2-devc.eks:4222"}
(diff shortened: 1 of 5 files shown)


yaroliakh (approved):

paasta_tools/cli/cmds/logs.py
```diff
@@ -1230,6 +1239,48 @@ def print_logs_by_time(
...
+        strip_headers: bool = False,
+    ) -> None:
+        stream_name = get_log_name_for_service(service, prefix="app_output")
+        endpoint = self.get_nats_endpoint_for_cluster(clusters[0])
+        if not endpoint:
+            raise NotImplementedError(
+                "Tailing logs is not supported in this cluster yet, sorry"
+            )
+
+        async def tail_logs_from_nats() -> None:
```
nemacysts, line 1260: this is a lot nicer than what the `scribereader` version of this does :p

nemacysts (approved):

paasta_tools/cli/cmds/logs.py
jvperrin: Thanks! Yeah the nats-py stuff is all async, so have to use it here, but it turned out quite clean which is great.

I'm not sure how this'll do concurrency-wise compared to the scribereader version, but I think that'll have to wait until prod to really see with something high-volume. From the NATS resources side of things at least it's been perfectly fine on load which is a good sign.

PR #3909, opened 2024-07-01
Speed up the default preStop hook to allow for graceful shutdown

nemacysts:
The default PaaSTA preStop hook sleeps for 30s - which is exactly the same as the default termination grace period, therefore preventing anything that gets this default hook from being able to gracefully shutdown.

We could (should?) look into simply not adding a hook for things that don't need one (i.e., not mesh-registed and no custom hooks), but in the meantime adding a silly message doesn't seem particularly harmful.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1472,7 +1472,10 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         if not command:
             return V1Handler(
                 _exec=V1ExecAction(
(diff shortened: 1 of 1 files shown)


nemacysts:
i haven't tested this in infrastage/kubestage yet - but the unit tests are passing on my devbox :p

jfongatyelp (approved):

nemacysts:
(i'll try to test this in infrastage tomorrow)

nemacysts:
(although the routable pod big bounce test not firing is a good sign :))

cuza (approved):

PR #3910, opened 2024-07-02
Revert "Attempt #2 MLCOMPUTE-1203 | Configure Spark driver pod memory and cores based on Spark args"

ankit864:
Reverts Yelp/paasta#3903


jfongatyelp (approved):
For context: this code assumed we had the spark kubeconfig everywhere but we only have it in the clusters where we have spark jobs, so everywhere else was failing setup_tron_namespace

PR #3912, opened 2024-07-03
Fix ensure_service_account breakages from the get_or_create_service_account refactor

nemacysts:
#3906 and #3903 were reverted due to issues with missing kubeconfigs on tron and ValueErrors on system nodes - these have both been fixed with 40ad655fb9c4f98887c4070446300f6a529f7ef9 (tron) and 30d5de84da9ad65caefa45168e96f8c7da6a2519 (system nodes)

PR #3913, opened 2024-07-03
U/krall/unrevert and patch

EvanKrall:


nemacysts:
we handled this in https://github.com/Yelp/paasta/pull/3912

PR #3918, opened 2024-07-10
adding sidecars for prometheus pushgateway and metric scrapper

alexanderjulianmartinez:
Adding k8s sidecars for Vitess paasta deployement Prometheus pushgateway and Vitess metric scrapper to run bash script. Additional information can be found in JIRA ticket below:
https://jira.yelpcorp.com/browse/DREIMP-10714

### DIFF
--- a/paasta_tools/vitesscluster_tools.py
+++ b/paasta_tools/vitesscluster_tools.py
@@ -201,6 +201,19 @@ class KeyspaceConfigDict(TypedDict, total=False):
     name: str
 
 
+class PrometheusPushGatewayConfigDict(TypedDict, total=False):
+    image: str
+    name: str
+    restartPolicy: str
+    command: List[str]
+
+
+class VitessMetricsConfigDict(TypedDict, total=False):
+    name: str
+    restartPolicy: str
+    command: List[str]
+
+
 def get_formatted_environment_variables(
     env_vars: Dict[str, Any]
 ) -> List[Union[KVEnvVar, KVEnvVarValueFrom]]:
@@ -591,6 +604,8 @@ class VitessDeploymentConfigDict(KubernetesDeploymentConfigDict, total=False):
     keyspaces: List[KeyspaceConfigDict]
     updateStrategy: Dict[str, str]
     globalLockserver: Dict[str, Dict[str, str]]
+    prometheusPushGateway: Dict[str, str]
+    vitessMetricScrapper: Dict[str, str]
 
 
 class VitessInstanceConfigDict(KubernetesDeploymentConfigDict, total=False):
@@ -756,6 +771,27 @@ class VitessDeploymentConfig(KubernetesDeploymentConfig):
     def get_update_strategy(self) -> Dict[str, str]:
         return {"type": "Immediate"}
 
+    def get_prometheus_pushgateway_config(self) -> PrometheusPushGatewayConfigDict:
+        config = PrometheusPushGatewayConfigDict(
+            name="prometheusPushGateway",
+            image="prometheus/pushgateway",
+            restartPolicy="OnFailure",
+            command=['--web.listen-address=":8000"'],
+        )
+        return config
+
+    def get_load_vitess_metrics_config(self) -> VitessMetricsConfigDict:
+        config = VitessMetricsConfigDict(
+            name="vitessMetricScrapper",
+            restartPolicy="OnFailure",
+            command=[
+                "sh",
+                "-c",
+                "puppet:///modules/profile_vitess/scrape-vitess-containers.sh",
+            ],
+        )
+        return config
+
     def get_vitess_config(self) -> VitessDeploymentConfigDict:
         vitess_config = VitessDeploymentConfigDict(
             namespace=self.get_namespace(),
@@ -766,6 +802,8 @@ class VitessDeploymentConfig(KubernetesDeploymentConfig):
             vtadmin=self.get_vtadmin(),
             keyspaces=self.get_keyspaces(),
             updateStrategy=self.get_update_strategy(),
+            prometheusPushGateway=self.get_prometheus_pushgateway_config(),
+            vitessMetricScrapper=self.get_load_vitess_metrics_config(),
         )
         return vitess_config
 
--- a/tests/test_vitesscluster_tools.py
+++ b/tests/test_vitesscluster_tools.py
@@ -610,6 +610,21 @@ VITESS_CONFIG = {
             "requests": {"cpu": "100m", "memory": "256Mi"},
         },
     },
+    "prometheusPushGateway": {
+        "name": "prometheusPushGateway",
+        "image": "prometheus/pushgateway",
+        "restartPolicy": "OnFailure",
+        "command": ['--web.listen-address=":8000"'],
+    },
+    "vitessMetricScrapper": {
+        "name": "vitessMetricScrapper",
+        "restartPolicy": "OnFailure",
+        "command": [
+            "sh",
+            "-c",
+            "puppet:///modules/profile_vitess/scrape-vitess-containers.sh",
+        ],
+    },
 }
 
 


paasta_tools/vitesscluster_tools.py
```diff
@@ -756,6 +771,27 @@ def get_keyspaces(self) -> List[KeyspaceConfigDict]:
...
+        return config
+
+    def get_load_vitess_metrics_config(self) -> VitessMetricsConfigDict:
+        config = VitessMetricsConfigDict(
+            name="vitessMetricScrapper",
+            restartPolicy="OnFailure",
+            command=[
+                "sh",
+                "-c",
+                "puppet:///modules/profile_vitess/scrape-vitess-containers.sh",
```
alexanderjulianmartinez, line 790: Wanted to make sure what's proper way of ensuring a specific bash script is mounted within a paasta pod? Will update this value based on feedback.

nemacysts:
@alexanderjulianmartinez this isn't particularly pressing, but could we replace the PR description with something more descriptive than a jira ticket link? it would aid future debugging if the commit description as seen in git has context rather than forcing a jump to jira :)

paasta_tools/vitesscluster_tools.py
nemacysts: this won't work as-is: paasta is pretty unaware of puppet. there's two ways of using a script in a paasta pod:
1. include it inside your containers
2. puppet the script onto kube nodes and mount the script into the pod

we don't really do per-pool puppet (for reasons™ - i.e., y/lasagna) - so i'd probably recommend including this script in your docker images

alexanderjulianmartinez:
> @alexanderjulianmartinez this isn't particularly pressing, but could we replace the PR description with something more descriptive than a jira ticket link? it would aid future debugging if the commit description as seen in git has context rather than forcing a jump to jira :)

Yes of course sorry updated the description to reflect what has been added.

paasta_tools/vitesscluster_tools.py
```diff
@@ -766,6 +802,8 @@ def get_vitess_config(self) -> VitessDeploymentConfigDict:
             vtadmin=self.get_vtadmin(),
             keyspaces=self.get_keyspaces(),
             updateStrategy=self.get_update_strategy(),
+            prometheusPushGateway=self.get_prometheus_pushgateway_config(),
+            vitessMetricScrapper=self.get_load_vitess_metrics_config(),
```
nemacysts, line 806: note: these sidecars should probably also have resource requests

nemacysts: question: will vitess understand this config? from looking at the vitess crd docs, it looks like a couple of their CRDs expose a sidecarContainers field that takes an array of k8s V1Container objects

paasta_tools/vitesscluster_tools.py
nemacysts: that said: i wonder if instead of porting over this pushgateway complexity we can instead have a simpler exporter that converts the vitess metrics into prometheus

(i.e., a single sidecar rather than two)

nemacysts: we previously used this pattern for uwsgi metrics before centralizing to a single centralized exporter :)

paasta_tools/vitesscluster_tools.py
alexanderjulianmartinez: Vitess already exposes the metrics via and endpoint so the sidecars are more for Prometheus/Grafana, and Vitess itself I don't think needs to know about the configs specifcally just need the sidecars running for push the metrics and filter them per tablet via the bash script.

nemacysts: correct, i'm suggesting that we do the prometheus translation with a single sidecar that we configure prometheus to scrape rather than have two sidecars (one for translation and another for actually getting the metrics into prometheus)

paasta_tools/vitesscluster_tools.py
nemacysts: to add a file to your image (which is what i'd recommend), you'd want to `COPY` the file from the repo that the image is built from into the image in the dockerfile

if we're mirroring an upstream image here, we'll need to do some additional work to create a dockerfile that starts from the mirrored image and adds any customizations

that said: i think this would be a whole lot simpler if we had a separate image just for something that would re-export the vitess metrics in a prometheus-scrapable format - e.g., we used to run github.com/timonwong/uwsgi_exporter as a sidecar for uwsgi-autoscaled instances to re-export the uwsgi metrics (which are not prometheus-scrapable) in a scrapable format

paasta_tools/vitesscluster_tools.py
```diff
@@ -756,6 +771,27 @@ def get_keyspaces(self) -> List[KeyspaceConfigDict]:
     def get_update_strategy(self) -> Dict[str, str]:
         return {"type": "Immediate"}
 
+    def get_prometheus_pushgateway_config(self) -> PrometheusPushGatewayConfigDict:
+        config = PrometheusPushGatewayConfigDict(
+            name="prometheusPushGateway",
+            image="prometheus/pushgateway",
```
nemacysts, line 777: just in case we do end up going with this: this would need to be an internally mirrored image on docker-paasta

nemacysts: i also don't think we'd want to run our own pushgateway here - we'd want to use the one that PE manages 

(although, you'd want to speak to them before using since pushgateway is not generally something that we encourage/support very much)

paasta_tools/vitesscluster_tools.py
alexanderjulianmartinez: Yeah that seems like a better idea, was mostly just trying to bring parity first to our puppet setup. If it's a requirement we can re-write the bash. Below is the puppet version which we run a cron job, hence why we push metrics through the gateway. Might also give better context of our solution initially at least:
https://sourcegraph.yelpcorp.com/sysgit/puppet@master/-/blob/modules/profile_vitess/manifests/init.pp?L168-186

paasta_tools/vitesscluster_tools.py
alexanderjulianmartinez: Ah do you mean this docker-images/pushgateway? https://sourcegraph.yelpcorp.com/docker-images/pushgateway

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: @alexanderjulianmartinez I think this might not work as is. The vttablet spec (and similarly for vtctld and other vitess components) has a field to include a `sidecarContainers` list (https://github.yelpcorp.com/misc/compute-infra-k8s/blob/master/clusters/eks-infrastage/90-vitess-operator/eks/40-crd-planetscale.com-vitesskeyspaces.yaml#L564) of type core v1 container (https://github.com/planetscale/vitess-operator/blob/f1fe438f1a753007f5cace547b198a6e3cfaf1e3/pkg/operator/vttablet/spec.go#L60)

Core V1 Container json spec fields: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/

We'd ideally want to include the image and container spec for prometheus_pushgateway in the sidecarContainers with an init command if needed

paasta_tools/vitesscluster_tools.py
nemacysts: (if PE is fine with using their pushgateway here or y'all are willing to manage your own pushgateway setup): probably the easiest way to do this would be to create an image just for this script and push that to docker-paasta alongside the other vitess images

that said: the script will need a significant bit of refactoring - i think the only reusable bits are the curls to get metrics and the curls to send the metrics to the pushgateway (all the docker stuff will not work inside the sidecar).

paasta_tools/vitesscluster_tools.py
nemacysts: if you meant the internally mirrored image: yea, we'd need to check if that's been pushed to docker-paasta

if you meant the PE-managed pushgateway: y/pushgateway :)

paasta_tools/vitesscluster_tools.py
alexanderjulianmartinez: Ah good catch almost didn't see those, yeah those should be replaceable with pod configs from vitess etc from the config. Will reach out to PE and work on creating a new script

VinaySagarGonabavi: @alexanderjulianmartinez IMO, we cannot use this script as is. We'd need to modify the hostname and port with prometheus_pushgateway sidecar pod name and port in https://sourcegraph.yelpcorp.com/sysgit/puppet@4edf3b7b030e04241c495cbe7e8075ad556a6b0f/-/blob/modules/profile_vitess/files/scrape-vitess-containers.sh

And add this shell script to the docker image, vitess_base used by vttablet, vtctld and vtgate https://sourcegraph.yelpcorp.com/docker-images/vitess_base. vtadmin has a separate docker image if we're scraping it as well

nemacysts: (unless the sidecar needs binaries and whatnot provided by the vitess base image, it might be easier to have a separate image just for the sidecar: that's what we did when we ran the uwsgi exporter sidecar)

nemacysts: (that said, it will be cheaper to run a single sidecar that is prometheus scrapable than to run 2 sidecars for this push workflow)

nemacysts:
oh, i just took a closer look at what the scrape script in puppet is doing (/what the vitess docs say) and we should be able to point prometheus at these pods directly using service discovery: /metrics should already be in the prometheus format from what i'm seeing)

alexanderjulianmartinez:
Sidecars not needed after adding metrics scraping through prometheus for paasta pods.

PR #3921, opened 2024-07-16
Ensure default Spark executor SA is created in Spark clusters

nemacysts:
This is mostly a disaster-recovery thing since this SA exists already - but to avoid any unpleasant times should we decide to start these clusters from scratch, it'd be best to ensure that this SA actually exists :)

(I was talking to @CaptainSame in Slack and he reminded me about this SA)

### DIFF
--- a/paasta_tools/setup_tron_namespace.py
+++ b/paasta_tools/setup_tron_namespace.py
@@ -97,6 +97,26 @@ def ensure_service_accounts(job_configs: List[TronJobConfig]) -> None:
                         namespace=spark_tools.SPARK_EXECUTOR_NAMESPACE,
                         kube_client=spark_kube_client,
                     )
+            elif (
+                action.get_executor() == "spark"
+                # NOTE: we only enter this block if there's no IAM role specified as otherwise the above block would do
+                # the same thing as calling ensure_service_account() with iam_role=get_spark_executor_iam_role()
+                and not action.get_iam_role()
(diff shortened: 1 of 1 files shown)


paasta_tools/setup_tron_namespace.py
```diff
@@ -97,6 +97,26 @@ def ensure_service_accounts(job_configs: List[TronJobConfig]) -> None:
                         namespace=spark_tools.SPARK_EXECUTOR_NAMESPACE,
                         kube_client=spark_kube_client,
                     )
+            elif (
```
CaptainSame, line 100: Do we need this block since there is always an iam_role returned by `action.get_iam_role()` if `action.get_executor() == "spark"`, so we will never hit this condition.
If the objective is to safeguard against `action.get_iam_role()` being None, then we can just take out the spark executor service account creation logic from this if else block and put it under `if action.get_executor()=="spark" and action.get_spark_executor_iam_role()`
What do you think?

nemacysts: ah, you're right: this should just be `action.get_executor() == "spark" and action.get_spark_executor_iam_role()`!

nemacysts: ah, i see - we can just combine the previous logic and this and call ensure_service_account twice: once with get_iam_role (for the driver) and then with get_spark_executor_iam_role for the executor

CaptainSame (approved):
LGTM

CaptainSame (approved):
LGTM

PR #3923, opened 2024-07-18
MLCOMPUTE-1389 | remove spark.app.id and generate static spark.app.name

CaptainSame:
- `spark.app.id` is generated fresh and added by service_configuration_lib with a random postfix because of which Tron keeps running update config flow every few minutes. This config should be removed and generated by yelp spark-submit wrapper or by Spark itself.
- `spark.app.name` is also generated fresh with a random timestamp and postfix which leads to the same problem. This config should be made static and potentially replaced by yelp spark-submit wrapper later

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -364,6 +364,16 @@ class TronActionConfig(InstanceConfig):
             ),
             user=spark_tools.SPARK_JOB_USER,
         )
+        # delete the dynamically generated spark.app.id to prevent frequent config updates in Tron.
+        # spark.app.id will be generated later by yelp spark-submit wrapper or Spark itself.
+        spark_conf.pop("spark.app.id", None)
+        # use a static spark.app.name to prevent frequent config updates in Tron.
+        # md5 and base64 will always generate the same encoding for a string.
+        # This spark.app.name might be overridden by yelp spark-submit wrapper.
+        if "spark.app.name" in spark_conf:
+            spark_conf["spark.app.name"] = limit_size_with_hash(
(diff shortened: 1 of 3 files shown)


paasta_tools/tron_tools.py
```diff
@@ -364,6 +364,16 @@ def build_spark_config(self) -> Dict[str, str]:
             ),
             user=spark_tools.SPARK_JOB_USER,
         )
+        # delete the dynamically generated spark.app.id to prevent frequent config updates in Tron.
+        # spark.app.id will be generated later by yelp spark-submit wrapper or Spark itself.
+        spark_conf.pop("spark.app.id", None)
+        # use a static spark.app.name to prevent frequent config updates in Tron.
+        # md5 and base64 will always generate the same encoding for a string.
+        # This spark.app.name might be overridden by yelp spark-submit wrapper.
```
nemacysts, line 372: from a sourcegraph search for `repo:^sysgit/yelpsoa-configs$ spark.app.name` it seems like some folks set short static app names, so we may want to follow this up with a follow-up PR that allows static/user-set app names to continue being set if that's something we want to support

nemacysts (approved):

paasta_tools/tron_tools.py
chi-yelp: It's also auto set by spark-etl jobs - sourcegraph: `repo:spark_etl$ spark.app.name`

chi-yelp (approved):

paasta_tools/tron_tools.py
CaptainSame: Fixed this

nemacysts (approved):

PR #3924, opened 2024-07-22
MLCOMPUTE-1208 | fix spark args regex

CaptainSame:
- Prevent validation failures

### DIFF
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -383,18 +383,18 @@
                         },
                         "spark.executor.memory": {
                             "type": "string",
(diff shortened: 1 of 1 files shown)


nemacysts (approved):
i had used `[kmg]` on purposes since using any units bigger or smaller than that seemed unlikely to ever work (e.g., i don't think we normally run boxes with a TB of RAM in spark pools and we certainly don't run with any that have a PB of RAM) - but if there's folks using values like these and they actually fit 🚢

chi-yelp:
Note that [b] is currently not supported when we parse these parameters, we could just remove it here since values with [b] unit generally don't make sense in spark.

https://github.com/Yelp/paasta/blob/cc5f1e698d17f4f2122cd8717030b54a36b4ba83/paasta_tools/spark_tools.py#L254

PR #3925, opened 2024-07-29
Color paasta rollback warnings yellow instead of red

nemacysts:
I've noticed a couple folks think that paasta rollback failed because of this coloring - let's avoid some confusion and color this text yellow instead (which is more accurate anyway since it's a warning)


karmeleon (approved):
<3

jfongatyelp (approved):

EvanKrall (approved):

PR #3926, opened 2024-08-05
MLCOMPUTE-1497 | include spark driver memory overhead in driver pod memory

CaptainSame:
- Spark driver pod memory needs to allocate memory for driver memory overhead config along with the driver memory config

### DIFF
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -445,7 +445,7 @@
                             "type": "boolean"
                         },
                         "spark.app.name": {
-                            "type": "boolean"
+                            "type": "string"
                         },
                         "spark.task.maxFailures": {
                             "type": "integer",
--- a/paasta_tools/spark_tools.py
+++ b/paasta_tools/spark_tools.py
@@ -6,7 +6,6 @@ from typing import Any
 from typing import cast
 from typing import Dict
 from typing import List
-from typing import Literal
 from typing import Mapping
 from typing import Set
 
--- a/paasta_tools/spark_tools.py
+++ b/paasta_tools/spark_tools.py
@@ -264,7 +264,21 @@ def get_spark_memory_in_unit(mem: str, unit: Literal["k", "m", "g", "t"]) -> flo
         try:
             memory_bytes = float(mem[:-1]) * MEM_MULTIPLIER[mem[-1]]
         except (ValueError, IndexError):
-            print(f"Unable to parse memory value {mem}. Defaulting to 2 GB.")
-            memory_bytes = 2147483648  # default to 2 GB
+            print(f"Unable to parse memory value {mem}. Defaulting to 0.")
+            memory_bytes = 0  # default to 0
     memory_unit = memory_bytes / MEM_MULTIPLIER[unit]
     return memory_unit
+
+
+def get_spark_driver_memory_overhead_in_mb(spark_config: Dict[str, str]) -> float:
+    """
+    Returns the Spark driver memory overhead in bytes.
+    """
+    driver_mem_overhead = "0"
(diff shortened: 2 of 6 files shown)


paasta_tools/spark_tools.py
```diff
@@ -264,7 +264,21 @@ def get_spark_memory_in_unit(mem: str, unit: Literal["k", "m", "g", "t"]) -> flo
...
+            memory_bytes = 0  # default to 0
     memory_unit = memory_bytes / MEM_MULTIPLIER[unit]
     return memory_unit
+
+
+def get_spark_driver_memory_overhead_in_mb(spark_config: Dict[str, str]) -> float:
+    """
+    Returns the Spark driver memory overhead in bytes.
+    """
+    driver_mem_overhead = "0"
```
chi-yelp, line 277: We'll need to also consider `spark.driver.memoryOverheadFactor`, it can be set by the user or from the default in Spark (0.1)

CaptainSame: We'll consider that in service_configuration_lib, which will spit out a final `spark.driver.memoryOverhead`. I don't want to add all that logic here

paasta_tools/spark_tools.py
```diff
@@ -264,7 +264,21 @@ def get_spark_memory_in_unit(mem: str, unit: Literal["k", "m", "g", "t"]) -> flo
         try:
             memory_bytes = float(mem[:-1]) * MEM_MULTIPLIER[mem[-1]]
         except (ValueError, IndexError):
-            print(f"Unable to parse memory value {mem}. Defaulting to 2 GB.")
-            memory_bytes = 2147483648  # default to 2 GB
+            print(f"Unable to parse memory value {mem}. Defaulting to 0.")
+            memory_bytes = 0  # default to 0
```
chi-yelp, line 268: Should we just throw an exception to catch the problem before the problem config is pushed / the job is deployed?

paasta_tools/tron_tools.py
```diff
@@ -302,15 +302,20 @@ def get_cpus(self) -> float:
...
-        if (
-            self.action_spark_config
-            and "spark.driver.memory" in self.action_spark_config
-        ):
-            return int(
-                spark_tools.get_spark_memory_in_unit(
-                    self.action_spark_config["spark.driver.memory"], "m"
+        if self.action_spark_config:
+            # default to 2GB if spark.driver.memory is not set
+            driver_mem_mb = 2048
```
chi-yelp, line 307: Let's make this as a constant or a configurable value?

paasta_tools/spark_tools.py
chi-yelp: Do you mean overwriting the `spark.driver.memoryOverhead` value after considering `spark.driver.memoryOverheadFactor` in service_configuration_lib? This doesn't sound better than reading both of them together to me

paasta_tools/spark_tools.py
chi-yelp, line 268: Perhaps returning `None` is better than 0?

paasta_tools/cli/schemas/tron_schema.json
```diff
@@ -445,7 +445,7 @@
                             "type": "boolean"
                         },
                         "spark.app.name": {
-                            "type": "boolean"
+                            "type": "string"
```
nemacysts, line 448: nice-to-have suggestion: might be worth adding some additional constraints here too re: min/max length since (iirc) this gets turned into a k8s label and those values have a max length (and if there's any characters that are not allowed, there's a field for adding a pattern that these strings need to match to pass validation)

nemacysts (approved):

chi-yelp (approved):

PR #3927, opened 2024-08-07
Allow setting a uses_bulkdata default

timmow:
In global paasta config, so we can set it on a per ecosystem/region level in puppet to enable us to roll this out without having to modify every instance in yelpsoa_configs to set uses_bulkdata: true

This is part of the [new proposed rollout
plan](https://docs.google.com/document/d/1RwcJ5JVi2G0Sc2IAXA6tze2CNj2NR8WHN3-rFLjxuOU/edit#bookmark=id.edii46zhxpmu)

Tested by creating a test cluster (from [these instructions](https://yelpwiki.yelpcorp.com/pages/viewpage.action?pageId=234312044)

```
make k8s_fake_cluster
make generate_deployments_for_service
```

saving kubeconfig to a file

```
kind get kubeconfig --name tmower-k8s-test > .paasta-kube-config 
```

Editing `k8s_itests/deployments/paasta/fake_etc_paasta/clusters.json` and setting `uses_bulkdata_default` to `false`

Then running `Run setup k8s job in playground` from the vscode run / debug menu

Then running

```
KUBECONFIG=./.paasta-kube-config kubectl get deployment -n paastasvc-compute-infra-test-service -oyaml | grep volume -i -C5
```

no /nail/bulkdata volume was added

Setting the uses_bulkdata_default value back to true, and rerunning setup k8s job in playground, the above command showed a bulkdata volume

I'm not sure if its sustainable to run `load_system_paasta_config()` every time `has_bulkdata` is called, or even if we move this call to when `InstanceConfig` is instantiated, but i'm not familiar enough with the code to know if theres a better way of calling `load_system_paasta_config()` once and passing it into `InstanceConfig`, or if we could add a `lru_cache` to `load_system_paasta_config` - so would appreciate help here

### DIFF
--- a/paasta_tools/frameworks/native_service_config.py
+++ b/paasta_tools/frameworks/native_service_config.py
@@ -171,7 +171,8 @@ class NativeServiceConfig(LongRunningServiceConfig):
         computed separately.
         """
         docker_volumes = self.get_volumes(
-            system_volumes=system_paasta_config.get_volumes()
+            system_volumes=system_paasta_config.get_volumes(),
+            uses_bulkdata_default=system_paasta_config.get_uses_bulkdata_default(),
         )
         task: TaskInfo = {
             "name": "",
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2142,7 +2142,8 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             service=self.service, namespace=self.get_nerve_namespace()
         )
         docker_volumes = self.get_volumes(
-            system_volumes=system_paasta_config.get_volumes()
+            system_volumes=system_paasta_config.get_volumes(),
+            uses_bulkdata_default=system_paasta_config.get_uses_bulkdata_default(),
         )
 
         hacheck_sidecar_volumes = system_paasta_config.get_hacheck_sidecar_volumes()
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -357,7 +357,10 @@ class TronActionConfig(InstanceConfig):
             paasta_service=self.get_service(),
             paasta_instance=self.get_instance(),
             docker_img=docker_img_url,
-            extra_volumes=self.get_volumes(system_paasta_config.get_volumes()),
+            extra_volumes=self.get_volumes(
+                system_paasta_config.get_volumes(),
+                uses_bulkdata_default=system_paasta_config.get_uses_bulkdata_default(),
+            ),
             use_eks=True,
             k8s_server_address=get_k8s_url_for_cluster(self.get_cluster()),
             force_spark_resource_configs=self.config_dict.get(
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -624,7 +624,8 @@ class InstanceConfig:
 
         Defaults to 1024 (1GiB) if no value is specified in the config.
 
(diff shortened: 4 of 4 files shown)


timmow:
running `tox -e tests` passes locally so unsure why this is failing here

timmow:
ahhh so tests pass on the devbox as the `/etc/paasta` directory exists, which it doesnt in the GHE test env, and the [default](https://sourcegraph.yelpcorp.com/-/editor?remote_url=git%40github.com%3AYelp%2Fpaasta.git&branch=u%2Ftmower%2Fallow-setting-uses_bulkdata-default-PERES-4995&file=paasta_tools%2Futils.py&editor=VSCode&version=2.2.17&start_row=2061&start_col=10&end_row=2061&end_col=10) is [`/etc/paasta`](https://sourcegraph.yelpcorp.com/-/editor?remote_url=git%40github.com%3AYelp%2Fpaasta.git&branch=u%2Ftmower%2Fallow-setting-uses_bulkdata-default-PERES-4995&file=paasta_tools%2Futils.py&editor=VSCode&version=2.2.17&start_row=92&start_col=0&end_row=92&end_col=32)

timmow:
@Yelp/paasta (notifying a team as per https://yelpwiki.yelpcorp.com/display/COMPUTE/Compute%2BInfra%2BOn-point#

EvanKrall (approved):

paasta_tools/frameworks/native_service_config.py
```diff
@@ -171,7 +171,10 @@ def base_task(
         computed separately.
         """
         docker_volumes = self.get_volumes(
-            system_volumes=system_paasta_config.get_volumes()
+            system_volumes=system_paasta_config.get_volumes(),
+            uses_bulkdata_default=system_paasta_config.config_dict.get(
+                "uses_bulkdata_default", True
+            ),
```
EvanKrall, line 177: this gets repeated a lot, and should probably be a method on SystemPaastaConfig.

I feel less strongly about this, but I think I'd also slightly prefer to see `get_volumes` just take the whole SystemPaastaConfig object as a parameter rather than all 3 callers having to pull two pieces of information from SPC.

timmow: > I feel less strongly about this, but I think I'd also slightly prefer to see get_volumes just take the whole SystemPaastaConfig object as a parameter rather than all 3 callers having to pull two pieces of information from SPC.

Interesting, I did go down this route, and found a lot of tests failed as they were not setting up the `SystemPaastaConfig` object, and by [law of demeter](https://en.wikipedia.org/wiki/Law_of_Demeter) was thinking perhaps the `InstanceConfig` object / `get_volumes` shouldn't need to know details about `SystemPaastaConfig`

Agree on making this a method on `SystemPaastaConfig` though

timmow: @EvanKrall updated with a method on SystemPaastaConfig

timmow: oh also worth noting - once we've rolled these changes out, theres no need for the `uses_bulkdata_default` param - we can hardcode it to `False`, so I can loop back and remove it

paasta_tools/frameworks/native_service_config.py
nemacysts, line 1: hopefully we remove this unused file in our next unhack :p

nemacysts (approved):

EvanKrall (approved):

PR #3932, opened 2024-08-16
Further calls to uses_bulkdata default

timmow:
Further to #3927 when discussing spark jobs with @chi [here](https://yelp.slack.com/archives/CA8BWU65D/p1723816981990909?thread_ts=1723740835.076679&cid=CA8BWU65D) I realised there were more calls to get_volumes on instance_config

I'm not sure how to test this

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -1209,7 +1209,10 @@ def configure_and_run_docker_container(
     if pull_image:
         docker_pull_image(docker_url)
 
-    for volume in instance_config.get_volumes(system_paasta_config.get_volumes()):
+    for volume in instance_config.get_volumes(
+        system_paasta_config.get_volumes(),
+        system_paasta_config.get_uses_bulkdata_default(),
+    ):
         if os.path.exists(volume["hostPath"]):
             volumes.append(
                 "{}:{}:{}".format(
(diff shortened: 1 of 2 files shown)


timmow:
@Yelp/paasta notifying as per y/ci-onpoint

nemacysts (approved):
for local-run, using your venv's paasta binary to run any `paasta local-run` command should work - for `paasta spark-run`, you'll need someone from ML Compute to give you an example command since that command is wholly owned by them

chi-yelp:
The following command runs the spark integration test by paasta binary in the venv:
```
.tox/py38-linux/bin/paasta spark-run --service spark --aws-profile=dev --cmd "spark-submit /code/integration_tests/s3.py"
```

and when the job is started, it'll print something like the following at the top which shows the volumes to be mounted to the Spark executors:
https://fluffy.yelpcorp.com/i/s4mpPG7BzMkb6rrBLVqCpDG5pvZlMGzQ.html#L64-113

chi-yelp (approved):

timmow:
tested local run, first tried running from within the paasta dir but realised i needed to be in my checked out peres_sandbox dir

Then ran

`../paasta/.tox/py38-linux/bin/paasta local-run -c norcal-devc -s peres_sandbox -i main`

But got an error as clog was not found - finally realised i needed to run

`source ../paasta/.tox/py38-linux/bin/activate` while in the peres_sandbox dir, and command ran fine

timmow:
the spark command ran as expected and included bulkdata

PR #3934, opened 2024-08-22
Add lifecycle hooks for cloudmap (de)registration operations on vtgate pods

VinaySagarGonabavi:
Added the scripts to the docker image in https://github.yelpcorp.com/docker-images/vitess_base/pull/21
AWS role and policy added for namespace pods in https://github.yelpcorp.com/misc/terraform-code/pull/24481

### DIFF
--- a/paasta_tools/vitesscluster_tools.py
+++ b/paasta_tools/vitesscluster_tools.py
@@ -545,7 +545,10 @@ def get_keyspaces_config(
         tablet_types = load_system_paasta_config().get_vitess_tablet_types()
         for tablet_type in tablet_types:
             ecosystem = region.split("-")[-1]
-            host = f"mysql-{cluster}-{tablet_type}.dre-{ecosystem}"
+            if tablet_type == "primary":
+                host = f"mysql-{cluster}.dre-{ecosystem}"
+            else:
+                host = f"mysql-{cluster}-{tablet_type}.dre-{ecosystem}"
 
             # We use migration_replication delay for migration tablets and read_replication_delay for everything else
             # Also throttling threshold for refresh and sanitized primaries is set at 30 seconds and everything else at 3 seconds
--- a/paasta_tools/vitesscluster_tools.py
+++ b/paasta_tools/vitesscluster_tools.py
@@ -134,6 +134,7 @@ class GatewayConfigDict(TypedDict, total=False):
     extraEnv: List[Union[KVEnvVar, KVEnvVarValueFrom]]
     extraFlags: Dict[str, str]
     extraLabels: Dict[str, str]
+    lifecycle: Dict[str, Dict[str, Dict[str, List[str]]]]
     replicas: int
     resources: Dict[str, Any]
     annotations: Mapping[str, Any]
@@ -236,6 +237,7 @@ def get_cell_config(
     labels: Dict[str, str],
     node_affinity: dict,
     annotations: Mapping[str, Any],
+    aws_region: str,
 ) -> CellConfigDict:
     """
     get vtgate config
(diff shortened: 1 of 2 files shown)


flopex (approved):

paasta_tools/vitesscluster_tools.py
```diff
@@ -712,6 +734,10 @@ def get_region(self) -> str:
             sys.exit(1)
         return region
 
+    def get_aws_region(self) -> str:
+        region = self.get_region()
+        return f"{region[:2]}-{region[2:6]}-{region[6:7]}"
```
nemacysts, line 739: this seems a tad brittle as this slicing won't work for all AWS regions - it's probably fine for where vitess will be deployed initially, but perhaps we can have the region identification happen inside the script itself?

paasta_tools/vitesscluster_tools.py
```diff
@@ -256,6 +258,26 @@ def get_cell_config(
     config = CellConfigDict(
         name=cell,
         gateway=GatewayConfigDict(
+            lifecycle={
+                "preStop": {
+                    "exec": {
+                        "command": [
+                            "/bin/sh",
+                            "-c",
+                            f"/cloudmap/scripts/deregister_from_cloudmap.sh vtgate-{cell} {aws_region}",
```
nemacysts, line 267: where is this script coming from?

nemacysts (commented):
are there any notes about y'all's discussion with CIC about potentially integrating vitess with our mesh? using hooks for service discovery registration/deregistration makes me a little nervous (e.g., what happens in cloudmap if a k8s node goes offline in a way where the prestop hooks for vitess pods don't run?)

paasta_tools/vitesscluster_tools.py
nemacysts: oh, whoops - this is mentioned in the PR description

VinaySagarGonabavi:
Reopening this as we moved back to a single VitessCluster setup in https://github.com/Yelp/paasta/pull/3947

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: > we can have the region identification happen inside the script itself?
Can you explain this? Or do you mean pass it from yelpsoa?

VinaySagarGonabavi:
> are there any notes about y'all's discussion with CIC about potentially integrating vitess with our mesh? using hooks for service discovery registration/deregistration makes me a little nervous (e.g., what happens in cloudmap if a k8s node goes offline in a way where the prestop hooks for vitess pods don't run?)

We had several rounds of discussion with CIC/Service Mesh on this which are noted in https://jira.yelpcorp.com/browse/DREIMP-10901 and the last discussion with CIPX is in https://yelp.slack.com/archives/C060C8L80LE/p1723597193236439?thread_ts=1723511315.755149&cid=C060C8L80LE

For the cases of when a pod goes away before it could deregister itself from cloudmap, we're planning to handle with an external monitoring check to query the exposed endpoint /debug/health from vtgate service on each pod ip registered in a cloudmap service. If either the pod is not responsive or the health check fails, it can be handled with an auto remediation step of deregistering the ip. It's not ideal but as Krall pointed out if the external monitoring job runs frequently enough, this case can be handled

paasta_tools/vitesscluster_tools.py
nemacysts: @VinaySagarGonabavi my bad, i meant that the register/deregister script could read /nail/etc/habitat and figure out the region from there - in general, it's probably better to have the script be somewhat independent of how paasta is configuring pods since it's easier/faster to roll out an image change for your service than to do anything else (e.g., you can avoid having to go through a paasta release if the script is called with no or minimal arguments)

nemacysts (approved):

PR #3938, opened 2024-08-27
DREIMP-10951: Autoscale VitessCell pods with HPA

siadat:
# Problem

We'd like to autoscale the vtgate deployments created by the vitess-operator for each VitessCell.

# Solution

Create an HPA for each VitessCell, where max_instances is set.

This PR should be merged after https://github.com/Yelp/paasta/pull/3940

# Context

- The VitessCell CRD should be changed as well: https://github.yelpcorp.com/misc/compute-infra-k8s/pull/1949
- Example autoscaled instance: https://github.yelpcorp.com/sysgit/yelpsoa-configs/pull/48782

http://y/DREIMP-10951

### DIFF
--- a/paasta_tools/vitesscell_tools.py
+++ b/paasta_tools/vitesscell_tools.py
@@ -187,19 +187,6 @@ class VitessCellConfig(VitessDeploymentConfig):
             "rootPath": TOPO_GLOBAL_ROOT,
         }
 
-    def update_related_api_objects(self, kube_client: KubeClient):
-        vitesscell_cr = get_cr(kube_client, cr_id(self.service, self.instance))
-        if not vitesscell_cr:
-            # Nothing to do, HPAs are deleted when its owner VitessCell is deleted.
-            return
-
-        uid = vitesscell_cr["metadata"]["uid"]
-        log.info(f"Reconciling HPA for {self.service} {self.instance}")
-        self.reconcile_vtgate_hpa(
-            kube_client,
-            owner_uid=uid,
-        )
-
     def get_desired_hpa(
         self,
         kube_client: KubeClient,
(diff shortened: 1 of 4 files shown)


paasta_tools/vitesscell_tools.py
```diff
@@ -115,6 +126,7 @@ def get_cell_config(
             },
             extraLabels=labels,
             replicas=replicas,
+            scale_selector=",".join([f"{k}={v}" for k, v in labels.items()]),
```
siadat, line 129: I could replace this with just service and instance labels, e.g. something like this:
```suggestion
            scale_selector=",".join([f"{k}={v}" for k, v in {
                paasta_prefixed("service"): self.service,
                paasta_prefixed("instance"): self.instance,
            }.items()]),
```
Let me know what you think.

VinaySagarGonabavi (approved):

nemacysts (commented):
imo, the autoscaling for vitess seems different enough that it's probably more maintainable to have a separate script that we use to configure HPAs rather than mix it into the existing code for long-running services

paasta_tools/vitesscell_tools.py
```diff
@@ -175,6 +183,129 @@ def get_global_lock_server(self) -> Dict[str, str]:
...
+            api_version="planetscale.com/v2", kind="VitessCell", name=name
+        )
+
+    def get_autoscaling_metric_spec(
+        self,
+        name: str,
+        cluster: str,
+        kube_client: KubeClient,
+        namespace: str,
+    ) -> Optional[V2beta2HorizontalPodAutoscaler]:
```
siadat, line 197: > imo, the autoscaling for vitess seems different enough that it's probably more maintainable to have a separate script that we use to configure HPAs rather than mix it into the existing code for long-running services

@nemacysts I added this method override to avoid using the existing one in KubernetesDeploymentConfig.

paasta_tools/vitesscell_tools.py
siadat: I'm leaving it like this for now. I think it is safe. We can change it later if necessary.

PR #3940, opened 2024-08-27
Revert "Revert "Allow to modify namespace when creating/updating/listing custom resources""

VinaySagarGonabavi:
This reverts commit c6e7bd5aca90d81a381b5dcc158e5364bb65a2e4.

Context:
Three types of custom resources, `VitessCluster`, `VitessCell`, and `VitessKeyspace` needs to be deployed in the same K8s namespace, `paasta-vitesscluster`. This change is a workaround to specify a namespace to perform k8s custom resource operations (create/update/list/delete)

What are the changes?
I added additional namespace argument to the core kubernetes_tools functions to make k8s api calls for custom objects. The default value when the argument is not passed is still set to the previous value.

Are these changes safe?
Since the default behavior of these functions don't change, and only is changed when explicitly passing a namespace value (only done in case of vitess custom resource operations), this change can be considered safe

### DIFF
--- a/paasta_tools/cleanup_kubernetes_cr.py
+++ b/paasta_tools/cleanup_kubernetes_cr.py
@@ -31,7 +31,7 @@ from paasta_tools.kubernetes_tools import KubeClient
 from paasta_tools.kubernetes_tools import list_custom_resources
 from paasta_tools.kubernetes_tools import load_custom_resource_definitions
 from paasta_tools.kubernetes_tools import paasta_prefixed
-from paasta_tools.setup_kubernetes_cr import INSTANCE_TYPE_TO_NAMESPACE_LOADER
+from paasta_tools.setup_kubernetes_cr import INSTANCE_TYPE_TO_NAMESPACE_MAPPING
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import load_all_configs
 from paasta_tools.utils import load_system_paasta_config
@@ -106,10 +106,11 @@ def cleanup_all_custom_resources(
         )
         if not config_dicts:
             continue
-        if crd.file_prefix in INSTANCE_TYPE_TO_NAMESPACE_LOADER:
-            namespace = INSTANCE_TYPE_TO_NAMESPACE_LOADER[crd.file_prefix]
-        else:
-            namespace = ""
+        if crd.file_prefix in INSTANCE_TYPE_TO_NAMESPACE_MAPPING:
+            # if there's no explicit mapping from instance type to a namespace,
+            # the None value here will serve as a sentinel later on for the "real"
+            # default value (which is based on the CR kind)
+            namespace = INSTANCE_TYPE_TO_NAMESPACE_MAPPING.get(crd.file_prefix, None)
         crs = list_custom_resources(
             kube_client=kube_client,
             kind=crd.kube_kind,
--- a/paasta_tools/cleanup_kubernetes_cr.py
+++ b/paasta_tools/cleanup_kubernetes_cr.py
@@ -31,6 +31,7 @@ from paasta_tools.kubernetes_tools import KubeClient
 from paasta_tools.kubernetes_tools import list_custom_resources
 from paasta_tools.kubernetes_tools import load_custom_resource_definitions
 from paasta_tools.kubernetes_tools import paasta_prefixed
+from paasta_tools.setup_kubernetes_cr import INSTANCE_TYPE_TO_NAMESPACE_LOADER
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import load_all_configs
 from paasta_tools.utils import load_system_paasta_config
@@ -105,11 +106,16 @@ def cleanup_all_custom_resources(
         )
         if not config_dicts:
             continue
+        if crd.file_prefix in INSTANCE_TYPE_TO_NAMESPACE_LOADER:
+            namespace = INSTANCE_TYPE_TO_NAMESPACE_LOADER[crd.file_prefix]
+        else:
+            namespace = ""
         crs = list_custom_resources(
             kube_client=kube_client,
             kind=crd.kube_kind,
             version=crd.version,
             group=crd.group,
+            namespace=namespace,
         )
         for cr in crs:
             service = config_dicts.get(cr.service)
--- a/paasta_tools/setup_kubernetes_cr.py
+++ b/paasta_tools/setup_kubernetes_cr.py
@@ -49,12 +49,18 @@ from paasta_tools.utils import get_config_hash
 from paasta_tools.utils import get_git_sha_from_dockerurl
 from paasta_tools.utils import load_all_configs
 from paasta_tools.utils import load_system_paasta_config
+from paasta_tools.vitesscell_tools import (
+    KUBERNETES_NAMESPACE as VITESSCELL_KUBERNETES_NAMESPACE,
+)
 from paasta_tools.vitesscell_tools import load_vitess_cell_instance_configs
-from paasta_tools.vitesscell_tools import VITESSCELL_KUBERNETES_NAMESPACE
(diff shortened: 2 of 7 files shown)


paasta_tools/cleanup_kubernetes_cr.py
```diff
@@ -105,11 +106,16 @@ def cleanup_all_custom_resources(
         )
         if not config_dicts:
             continue
+        if crd.file_prefix in INSTANCE_TYPE_TO_NAMESPACE_LOADER:
+            namespace = INSTANCE_TYPE_TO_NAMESPACE_LOADER[crd.file_prefix]
+        else:
+            namespace = ""
```
nemacysts, line 112: can the default be None to avoid confusion? i had to trace the code a bit to figure out if this was safe or not since i'm 90% sure that an empty string is not a valid namespace name

maybe something like
```suggestion
        # if there's no explicit mapping from instance type to a namespace,
        # the None value here will serve as a sentinel later on for the "real"
        # default value (which is based on the CR kind)
        namespace = INSTANCE_TYPE_TO_NAMESPACE_LOADER.get(crd.file_prefix, None)
```

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -64,6 +67,13 @@
 }
 
 
+INSTANCE_TYPE_TO_NAMESPACE_LOADER = {
```
nemacysts, line 70: this is maybe a bit of a misleading name since this isn't actually a loader/getter/etc - it's actually a mapping

paasta_tools/vitesskeyspace_tools.py
```diff
@@ -31,7 +31,7 @@
 log.addHandler(logging.NullHandler())
 
 
-KUBERNETES_NAMESPACE = "paasta-vitessclusters"
+VITESSKEYSPACE_KUBERNETES_NAMESPACE = "paasta-vitessclusters"
```
nemacysts, line 34: hmm, i think we only ever half-started using KUBERNETES_NAMESPACE for it's intended purpose - iirc, this was meant to be a standard constant so that we didn't need to keep a mapping from workload->namespace

nemacysts: since this diff isn't really doing anything special with the renamed constant, i'd probably leave the naming alone to be consistent with the rest of the instance types

nemacysts (commented):
i'm a little hesistant to go this route - i think that if we need to do manage multiple CRs from paasta, we'd probably want to make paasta multi-CR/CRD aware in a more complete way (e.g., revisit https://github.com/Yelp/paasta/pull/3778 or https://github.com/Yelp/paasta/pull/3777 and go from there)

paasta_tools/vitesskeyspace_tools.py
VinaySagarGonabavi: While this is correct, we'd have to import the whole .*_tools module instead of importing just the variable from the module since the variable name would be the same.
For example,

> from paasta_tools.vitesscell_tools import KUBERNETES_CONFIG
from paasta_tools.vitesskeyspace_tools import KUBERNETES_CONFIG

would create ambiguity for variable value

> from paasta_tools import vitesscell_tools

and referencing the variable with vitesscell_tools.KUBERNETES_CONFIG would eliminate the ambiguity but would require to import the module entirely.

Lmk your thoughts

VinaySagarGonabavi:
> i'm a little hesitant to go this route - i think that if we need to do manage multiple CRs from paasta, we'd probably want to make paasta multi-CR/CRD aware in a more complete way (e.g., revisit #3778 or #3777 and go from there)

IMO, the two PRs are trying to solve a different problem for CRD's added to yelpsoa while vitess crds are deployed with flux?
At least for this situation, the problem we're trying to solve is to have the same namespace for different CR's created using different file prefixes https://github.yelpcorp.com/sysgit/yelpsoa-configs/pull/48754 and https://github.yelpcorp.com/sysgit/puppet/pull/12896

paasta_tools/vitesskeyspace_tools.py
nemacysts: `from paasta_tools.vitesscell_tools import KUBERNETES_NAMESPACE as VITESS_CELL_NAMESPACE`, etc perhaps?

nemacysts:
> > i'm a little hesitant to go this route - i think that if we need to do manage multiple CRs from paasta, we'd probably want to make paasta multi-CR/CRD aware in a more complete way (e.g., revisit #3778 or #3777 and go from there)
> 
> IMO, the two PRs are trying to solve a different problem for CRD's added to yelpsoa while vitess crds are deployed with flux? At least for this situation, the problem we're trying to solve is to have the same namespace for different CR's created using different file prefixes https://github.yelpcorp.com/sysgit/yelpsoa-configs/pull/48754 and https://github.yelpcorp.com/sysgit/puppet/pull/12896

sorry, what i meant was that rather than hack around the problem of multiple CRs/CRDs for a workload how we've been doing it (e.g., managing some stuff in flux, having some vitess-specific CR code here, splitting vitess up like i just saw in https://github.com/Yelp/paasta/pull/3935, etc) we should build this support in in a more native way rather than essentially special-casing vitess

we'd been discussing the vitess setup a bit somewhat recently in CI and @jfongatyelp  had an interesting idea re: having something like a "wrapper" operator that creates the vitess CRs as well as supporting objects like HPAs and whatnot that then the real™ vitess operator handles as it would normally might be more maintainable for both CI and DRE. this is still kinda complex and will probably trip folks up at least once, but it might be better than adding all these special-case bits of logic for vitess if we don't/can't make paasta multi-CR/CRD-aware

nemacysts:
note: my comments in the original CEP about trying to figure out what the desired vitess soaconfigs would look like were to try to figure out what would need to be done to have paasta support running vitess as a native workload and attempt to frontload this discussion :)

VinaySagarGonabavi:
> > > i'm a little hesitant to go this route - i think that if we need to do manage multiple CRs from paasta, we'd probably want to make paasta multi-CR/CRD aware in a more complete way (e.g., revisit #3778 or #3777 and go from there)
> > 
> > 
> > IMO, the two PRs are trying to solve a different problem for CRD's added to yelpsoa while vitess crds are deployed with flux? At least for this situation, the problem we're trying to solve is to have the same namespace for different CR's created using different file prefixes https://github.yelpcorp.com/sysgit/yelpsoa-configs/pull/48754 and https://github.yelpcorp.com/sysgit/puppet/pull/12896
> 
> sorry, what i meant was that rather than hack around the problem of multiple CRs/CRDs for a workload how we've been doing it (e.g., managing some stuff in flux, having some vitess-specific CR code here, splitting vitess up like i just saw in #3935, etc) we should build this support in in a more native way rather than essentially special-casing vitess
> 
> we'd been discussing the vitess setup a bit somewhat recently in CI and @jfongatyelp had an interesting idea re: having something like a "wrapper" operator that creates the vitess CRs as well as supporting objects like HPAs and whatnot that then the real™ vitess operator handles as it would normally might be more maintainable for both CI and DRE. this is still kinda complex and will probably trip folks up at least once, but it might be better than adding all these special-case bits of logic for vitess if we don't/can't make paasta multi-CR/CRD-aware

Definitely agree that we're special casing for vitess in some of the paasta changes (file prefix per cr, custom instance config loading, and now namespaces). Definitely a good idea to have a wrapper to accommodate such things in the long run. This is a larger issue and happy to discuss more on this. But, at least the priority for now has been to deploy it right with as minimal changes as possible to the core paasta code.

VinaySagarGonabavi:
Discontinuing this change as the paasta playground testing results showed that a split down Vitess Deployment cannot coordinate the topology even with https://github.com/planetscale/vitess-operator/blob/c85f611722ae24f7ce68a30db0b39170ee182aef/docs/architecture/federation.md#topology-reconciliation

PR #3942, opened 2024-08-30
Fix the timestamps errors in the failed tests

EmanElsaban:
These tests were failing due to difference in timestamps. This PR fixes the timestamps to match the expected ones


nemacysts (approved):
i'm a little confused as to how these started failing in the first place since all the time stuff should be mocked 🤔

piax93 (approved):

wilmer05 (approved):

PR #3945, opened 2024-09-03
Autoscale vtgate replicas using HPA

siadat:
## Problem or Feature

Note: This PR will be unnecessary if the following one is merged: https://github.com/planetscale/vitess-operator/pull/598

We want to auto-scale vtgate using HPA. However, there are 2 problems:

1. VitessCells are created and reconciled by the vitess-operator. So, if we autoscale number of replicas in a cell, the value will be overwritten by the operator and copied from VitessCluster. Therefore, we need to do one of these things:
    - a. either create cells that are unmanaged by clusters (ie we create VitessCells, instead of having vitess-operator create them from VitessClusters),
    - b. or autoscale replicas in a VitessCluster in some other way
2. VitessCell (where vtgate is specified) is not scalable with HPA, because it does not define the /scale subresource.

## Solution

Therefore, this PR is doing the following:

1. Create a CR whose only purpose is to autoscale vtgate, call it ScalableVtgate
2. Create an HPA to auto-scale that ^
3. When reconciling the VitessCluster, read the number of replicas from ScalableVtgate

## Context

- This PR depends on https://github.yelpcorp.com/misc/compute-infra-k8s/pull/1973
- http://y/DREIMP-10951

### DIFF
--- a/paasta_tools/setup_kubernetes_cr.py
+++ b/paasta_tools/setup_kubernetes_cr.py
@@ -50,12 +50,20 @@ from paasta_tools.utils import get_git_sha_from_dockerurl
 from paasta_tools.utils import load_all_configs
 from paasta_tools.utils import load_system_paasta_config
 from paasta_tools.vitesscluster_tools import load_vitess_service_instance_configs
+from paasta_tools.vitesscluster_tools import (
+    update_related_api_objects as update_vitess_cluster_related_api_objects,
+)
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


VinaySagarGonabavi (approved):

nemacysts (commented):
we'll also probably want to add a schema for the vitessclusters files at some point (https://github.com/Yelp/paasta/tree/master/paasta_tools/cli/schemas)

PR #3948, opened 2024-09-05
tron: update schema for bulkdata + correctly pass extra volumes

nemacysts:
We collectively missed that get_volumes() gets called on the TronActionConfig and not the TronJobConfig. Furthermore, we missed that we weren't actually calling get_volumes() for normal Tron jobs - instead, we were only calling get_extra_volumes() and relying on the Tron `default_volumes` config to add the usual default volumes (which differs from how services work: we always call get_volumes() as we don't distinguish between user-provided volumes and default ones when generating the config there).

I've tested this on tron-infrastage (as well as ensured tests pass for this commit), so this should be safe.

### DIFF
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -231,6 +231,9 @@
                         }
                     }
                 },
+                "uses_bulkdata": {
+                    "type": "boolean"
+                },
                 "extra_volumes": {
                     "type": "array",
                     "items": {
@@ -660,9 +663,6 @@
                 },
                 "deploy_group": {
                     "type": "string"
-                },
-                "uses_bulkdata": {
-                    "type": "boolean"
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/schemas/tron_schema.json
```diff
@@ -231,6 +231,9 @@
                         }
                     }
                 },
+                "uses_bulkdata": {
+                    "type": "boolean"
+                },
```
nemacysts, line 236: oh, interesting - this was already at the right level - how did the schema not complain before?

nemacysts: hmm, or maybe this is just the GH diff view lying to me - `paasta validate` doesn't fail with the current version, but with my branch it does

nemacysts:
@88manpreet @CaptainSame just tagging y'all since i did touch some spark code - but it should work exactly the same as before (and the spark tests still pass) :)

paasta_tools/cli/schemas/tron_schema.json
jfongatyelp: I think it's the right indentation level but under the `job` schema definition rather than the `action` definition, so this PR does the right thing!

jfongatyelp (approved):

88manpreet (approved):

jfongatyelp (approved):

PR #3950, opened 2024-09-09
Mount srv-configs onto vtgate pods

VinaySagarGonabavi:
The cloudmap scripts expect srv-configs volume to be mounted on vtgate pods

### DIFF
--- a/paasta_tools/vitesscluster_tools.py
+++ b/paasta_tools/vitesscluster_tools.py
@@ -134,6 +134,8 @@ class GatewayConfigDict(TypedDict, total=False):
     extraEnv: List[Union[KVEnvVar, KVEnvVarValueFrom]]
     extraFlags: Dict[str, str]
     extraLabels: Dict[str, str]
+    extraVolumeMounts: List[Dict[str, Any]]
+    extraVolumes: List[Dict[str, Any]]
     lifecycle: Dict[str, Dict[str, Dict[str, List[str]]]]
     replicas: int
     resources: Dict[str, Any]
@@ -287,6 +289,28 @@ def get_cell_config(
                 "mysql_auth_vault_tls_ca": f"/etc/vault/all_cas/acm-privateca-{region}.crt",
                 "mysql_auth_vault_ttl": "60s",
             },
+            extraVolumeMounts=[
+                {
+                    "mountPath": "/nail/srv",
+                    "name": "srv-configs",
+                    "readOnly": True,
+                },
+                {
+                    "mountPath": "/nail/etc/srv-configs",
+                    "name": "etc-srv-configs",
+                    "readOnly": True,
+                },
+            ],
+            extraVolumes=[
+                {
+                    "name": "srv-configs",
+                    "hostPath": {"path": "/nail/srv"},
+                },
+                {
+                    "name": "etc-srv-configs",
+                    "hostPath": {"path": "/nail/etc/srv-configs"},
+                },
+            ],
             extraLabels=labels,
             replicas=replicas,
             resources={
--- a/tests/test_vitesscluster_tools.py
+++ b/tests/test_vitesscluster_tools.py
@@ -117,6 +117,21 @@ VITESS_CONFIG = {
                     "mysql_auth_vault_ttl": "60s",
                 },
                 "extraLabels": {"tablet_type": "fake_keyspaces_migration"},
+                "extraVolumeMounts": [
+                    {"mountPath": "/nail/srv", "name": "srv-configs", "readOnly": True},
+                    {
+                        "mountPath": "/nail/etc/srv-configs",
+                        "name": "etc-srv-configs",
+                        "readOnly": True,
+                    },
+                ],
+                "extraVolumes": [
+                    {"hostPath": {"path": "/nail/srv"}, "name": "srv-configs"},
+                    {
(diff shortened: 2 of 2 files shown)


nemacysts (commented):
i think we'll also need to setup the service account creation bits + make these pods use the pod identity service account for pod identity to work (and afaik, we still need the fs_group config) - i'm not sure that adding the iam annotation on your CRs will work (unless that's what flink is doing for pod identity)

flopex (approved):

VinaySagarGonabavi:
> i think we'll also need to setup the service account creation bits + make these pods use the paasta-created pod identity service account for pod identity to work (and afaik, we still need the fs_group config) - i'm not sure that adding the iam annotation on your CRs will work (unless that's what flink is doing for pod identity)

Added the service account creation bit, please take another look. Any ideas on how to add the service context? There is no PodSpec/PodTemplateSpec to be provided in the CRD https://github.yelpcorp.com/misc/compute-infra-k8s/blob/master/clusters/eks-infrastage/90-vitess-operator/eks/40-crd-planetscale.com-vitessclusters.yaml

matt-ullmer (approved):

paasta_tools/vitesscluster_tools.py
```diff
@@ -287,6 +289,28 @@ def get_cell_config(
                 "mysql_auth_vault_tls_ca": f"/etc/vault/all_cas/acm-privateca-{region}.crt",
                 "mysql_auth_vault_ttl": "60s",
             },
+            extraVolumeMounts=[
```
nemacysts, line 292: should we mount the default volumes defined in SystemPaastaConfig? or are we sure we only need/want srv-configs? (i.e., this isn't mounting the /nail/etc/{habitat,runtimeenv,superregion,etc} files)

paasta_tools/vitesscluster_tools.py
```diff
@@ -705,11 +718,24 @@ def get_annotations(self) -> Mapping[str, Any]:
...
+        if self.get_iam_role_provider() == "aws":
+            annotations["iam.amazonaws.com/role"] = ""
+            iam_role = self.get_iam_role()
+            kube_client = KubeClient()
+            if iam_role:
+                ensure_service_account(
+                    iam_role=iam_role,
+                    namespace=self.get_namespace(),
+                    kube_client=kube_client,
+                )
```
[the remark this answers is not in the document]
VinaySagarGonabavi: @nemacysts These changes are not included as we added annotations to the new service account created in https://github.yelpcorp.com/misc/compute-infra-k8s/pull/1972

paasta_tools/vitesscluster_tools.py
```diff
@@ -705,11 +718,24 @@ def get_annotations(self) -> Mapping[str, Any]:
...
             service_namespace_config, system_paasta_config
         )
-        annotations: Mapping[str, Any] = {
+        annotations: PodAnnotationsDict = {
             "smartstack_registrations": json.dumps(self.get_registrations()),
             "paasta.yelp.com/routable_ip": has_routable_ip,
         }
 
+        if self.get_iam_role_provider() == "aws":
+            annotations["iam.amazonaws.com/role"] = ""
```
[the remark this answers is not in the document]
VinaySagarGonabavi: @nemacysts These changes are not included as we added annotations to the new service account created in https://github.yelpcorp.com/misc/compute-infra-k8s/pull/1972

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: @nemacysts This is following the same pattern suggested earlier for vttablet volume mounts https://github.com/Yelp/paasta/blob/master/paasta_tools/vitesscluster_tools.py#L477-L516 . If there is an alternate/recommended way to do this?

For vtgates, we need the /nail/srv/configs/dre_service_discovery.yaml file as a part of lifecycle hook scripts

> i.e., this isn't mounting the /nail/etc/{habitat,runtimeenv,superregion,etc} files
Wouldn't the volumes mounted from the kube machine host paths be already having habit/runtimeenv/superregion specific srv-configs?

nemacysts: @VinaySagarGonabavi `system_paasta_config.get_volumes()`will read `/etc/paasta/volumes.json` and return some default mounts that are usually needed/helpful to have

> Wouldn't the volumes mounted from the kube machine host paths be already having habit/runtimeenv/superregion specific srv-configs?
yes, this is unrelated to srv-configs -  having these files inside your pods would allow for things like the register/deregister scripts to discover where a pod is running rather than having to configure those lifecycle hooks region parameters inside paasta :)

VinaySagarGonabavi: I'll update the script to find the region based on where it's running. The reliance on srv-configs is to find the cloudmap service id stored in srv-configs https://github.yelpcorp.com/sysgit/srv-configs/blob/master/ecosystem/devc/dre_service_discovery.yaml#L23

VinaySagarGonabavi: ++. Let me also add `system_paasta_config.get_volumes()` so we can find /nail/etc/region etc.,

VinaySagarGonabavi: @nemacysts https://github.com/Yelp/paasta/pull/3959

PR #3956, opened 2024-09-11
Attempt to batch config loading for tron deployments

nemacysts:
Right now we make at most 2N calls to the Tron API during config deployments: N to get the current configs and at most N if all services have changes.

To start, I'd like to reduce this to N by allowing GET /api/config to return all the configs so that the only requests needed are POSTs for changed configs. Since I'm a little worried about how tron-pnw-prod will fare with returning all the configs in a single request, using this new endpoint is behind a feature toggle. Hopefully this works out and we don't need to add pagination to the endpoint :)

Depending on how this goes, we can look into batching up the POSTs so that we can also do that in a single request (or at least <N requests if we want to chunk things a bit more).

NOTE: this requires https://github.com/Yelp/Tron/pull/996

### DIFF
--- a/paasta_tools/setup_tron_namespace.py
+++ b/paasta_tools/setup_tron_namespace.py
@@ -206,11 +206,23 @@ def main():
                         log.debug(f"Skipped {service}")
 
         except Exception:
-            log.exception(
-                f"Failed to create service account for {service} (will skip reconfiguring):"
-            )
+            if args.bulk_config_fetch:
+                # service account creation should be the only action that can throw if this flag is true,
+                # so we can safely assume that's what happened here in the log message
+                log.exception(
+                    f"Failed to create service account for {service} (will skip reconfiguring):"
+                )
+
+                # since service account creation failed, we want to skip reconfiguring this service
+                # as the new config will likely fail due to the missing service account - even though
+                # the rest of the config is valid
+                new_configs.pop(service, None)
+            else:
+                log.exception(f"Update for {service} failed:")
+
+            # NOTE: this happens for both ways of updating (bulk fetch and JIT fetch)
+            # since we need to print out what failed in either case
             failed.append(service)
-            new_configs.pop(service, None)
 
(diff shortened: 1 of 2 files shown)


paasta_tools/setup_tron_namespace.py
```diff
@@ -162,6 +170,7 @@ def main():
     k8s_enabled_for_cluster = (
         yaml.safe_load(master_config).get("k8s_options", {}).get("enabled", False)
     )
+    new_configs: Dict[str, str] = {}  # service -> new_config
```
jfongatyelp, line 173: til `create_complete_config` gives us back just a str; definitely not what i woulda expected by the name

paasta_tools/setup_tron_namespace.py
```diff
@@ -187,16 +197,31 @@ def main():
...
-                if client.update_namespace(service, new_config):
-                    updated.append(service)
-                    log.debug(f"Updated {service}")
-                else:
-                    skipped.append(service)
-                    log.debug(f"Skipped {service}")
         except Exception:
-            log.exception(f"Update for {service} failed:")
+            log.exception(
+                f"Failed to create service account for {service} (will skip reconfiguring):"
```
jfongatyelp, line 210: Should this change in error message + new_configs handling be moved to a gate on `args.bulk_config_fetch`, and the old error message/`failed` handling be used when we aren't using the flag?

jfongatyelp (approved):

paasta_tools/setup_tron_namespace.py
nemacysts: yea :(

paasta_tools/setup_tron_namespace.py
nemacysts: err, yea - that's a good catch!

nemacysts: updated!

jfongatyelp (approved):

PR #3958, opened 2024-09-11
PAASTA-18194: Stop querying all managed namespaces by default for paasta status

jfongatyelp:
Right now, for `paasta status` on any instance, we call `find_all_relevant_namespaces` to get the namespaces that have a deployment/sts for it, which in turn tries to `list_deployments` in _every_ managed namespace.

Since we should rarely need to look through _all_ managed namespaces for any running deployments (usually only when someone is moving their service), this PR just adds a new flag to go back to this behavior, but changes the default to search only the single configured namespace. 

Verified that calls to the API still work and `paasta status` honors the flag appropriately. 

The reduction is time is sorta shocking on norcal-devc:
Without all_namespaces param:
```
$ time curl "http://localhost:41893/v1/services/paasta-contract-monitor/main_eks/status?new=1 | jq '.'
...
real    0m0.355s
user    0m0.026s
sys     0m0.000s
```

With all_namespaces param (our current status quo):
```
$ time curl "http://localhost:41893/v1/services/paasta-contract-monitor/main_eks/status?new=1&all_namespaces=1" | jq '.'
...
real    0m5.254s
user    0m0.020s
sys     0m0.007s
```

And with cli:
```
$ time PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_for_development/ .tox/py38-linux/bin/python paasta_tools/cli/cli.py status -c norcal-devc -s paasta-contract-monitor -i main_eks


paasta-contract-monitor.main_eks in norcal-devc (EKS)
    Version:    5f01b83e (desired)
    State: Launching replicas
    Running versions:
      5f01b83e - Started 2024-09-02 03:16:04 (9 days ago)
        Replica States: 7 Healthy / 3 All Containers Waiting
          Unhealthy Replicas:
      ...

real    0m1.182s
user    0m0.698s
sys     0m0.180s

$ time PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_for_development/ .tox/py38-linux/bin/python paasta_tools/cli/cli.py status -c norcal-devc -s paasta-contract-monitor -i main_eks -A


paasta-contract-monitor.main_eks in norcal-devc (EKS)
    Version:    5f01b83e (desired)
    State: Launching replicas
    Running versions:
      5f01b83e - Started 2024-09-02 03:16:04 (9 days ago)
        Replica States: 7 Healthy / 3 All Containers Waiting
          Unhealthy Replicas:
         ...
         
real    0m7.130s
user    0m0.681s
sys     0m0.196s
```

Since the status endpoint accepts unknown parameters but just ignores them, the api+cli changes should be safe to deploy together. 

I have not added a systempaastaconfig to control the default for this, as I think it's pretty sure that we want to make the deafult search only the configured namespace for the requested instance.   Open to discussion if you think otherwise. 

There is also a minor fix to tox.ini to support some additional envvars that may be needed to set up your dev paasta api instances if you are testing w/ an EKS cluster.

### DIFF
--- a/paasta_tools/api/views/instance.py
+++ b/paasta_tools/api/views/instance.py
@@ -136,6 +136,7 @@ def instance_status(request):
     instance = request.swagger_data.get("instance")
     verbose = request.swagger_data.get("verbose") or 0
     use_new = request.swagger_data.get("new") or False
+    all_namespaces = request.swagger_data.get("all_namespaces") or False
     include_envoy = request.swagger_data.get("include_envoy")
     if include_envoy is None:
         include_envoy = True
@@ -199,6 +200,7 @@ def instance_status(request):
                     use_new=use_new,
                     instance_type=instance_type,
                     settings=settings,
+                    all_namespaces=all_namespaces,
                 )
             )
         elif instance_type == "tron":
--- a/paasta_tools/api/api_docs/oapi.yaml
+++ b/paasta_tools/api/api_docs/oapi.yaml
@@ -1480,6 +1480,12 @@ paths:
         required: false
         schema:
           type: boolean
+      - description: Search all namespaces for running copies
+        in: query
+        name: all_namespaces
+        required: false
+        schema:
+          type: boolean
       responses:
         "200":
           content:
--- a/paasta_tools/api/api_docs/swagger.json
+++ b/paasta_tools/api/api_docs/swagger.json
@@ -540,6 +540,13 @@
                         "name": "new",
                         "required": false,
                         "type": "boolean"
+                    },
+                    {
+                        "in": "query",
+                        "description": "Search all namespaces for running copies",
+                        "name": "all_namespaces",
+                        "required": false,
+                        "type": "boolean"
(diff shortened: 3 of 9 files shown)


paasta_tools/api/views/instance.py
```diff
@@ -136,6 +136,7 @@ def instance_status(request):
     instance = request.swagger_data.get("instance")
     verbose = request.swagger_data.get("verbose") or 0
     use_new = request.swagger_data.get("new") or False
+    all_namespaces = request.swagger_data.get("all_namespaces") or False
```
nemacysts, line 139: this could be 
```suggestion
    all_namespaces = request.swagger_data.get("all_namespaces", False)
```

but this is consistent with some of the other fields in this file so it's fine to keep as-is :)

nemacysts (approved):
i don't think folks are moving things to different namespaces outside of our completed namespace sharding migration, so the default you're setting here also seems like what we'd want to me :)

PR #3959, opened 2024-09-13
Add system volumes to vitess pods

VinaySagarGonabavi:
Additional changes needed locally to test
Modified `k8s_itests/deployments/paasta/fake_etc_paasta/volumes.json` to the devc version
Updated `paasta_tools/contrib/create_paasta_playground.py` to not modify the paasta config locally https://fluffy.yelpcorp.com/i/BhR02T8wwShMWjvK38JPptz9B2cX0c3g.html


matt-ullmer (approved):

paasta_tools/vitesscluster_tools.py
```diff
@@ -129,13 +130,58 @@ class ResourceConfigDict(TypedDict, total=False):
...
 
 
+class HostPathVolume(TypedDict, total=False):
+    path: str
+    type: str
+
+
+class KubernetesVolume(TypedDict, total=False):
+    name: str
+    hostPath: HostPathVolume
```
nemacysts, line 140: i'm a little surprised that we don't have typed dicts for this type of k8s volume shenanigans - but i guess that's maybe because we never store the volume info as a dict outside of representing the config as seen in soaconfigs

paasta_tools/vitesscluster_tools.py
```diff
@@ -775,16 +782,50 @@ def get_global_lock_server(self) -> Dict[str, Dict[str, str]]:
             }
         }
 
+    def get_kubernetes_volumes(self) -> List[KubernetesVolume]:
+        system_volumes = load_system_paasta_config().get_volumes()
+        docker_volumes = self.get_volumes(
```
nemacysts, line 787: we should hardcode passing in `uses_bulkdata_default=False` to all the self.get_volumes() calls so that we skip adding bulkdata mounts (that y'all don't need)

paasta_tools/vitesscluster_tools.py
```diff
@@ -843,6 +898,14 @@ def get_keyspaces(self) -> List[KeyspaceConfigDict]:
...
         node_affinity = self.get_vitess_node_affinity()
         annotations = self.get_annotations()
+        extra_volumes = self.get_kubernetes_volumes()
+        extra_volume_mounts = self.get_kubernetes_volume_mounts(extra_volumes)
+        api_client = ApiClient()
+        extra_volumes.extend(VTTABLET_EXTRA_VOLUMES)
+        for volume_mount in VTTABLET_EXTRA_VOLUME_MOUNTS:
+            extra_volume_mounts.append(
+                api_client.sanitize_for_serialization(volume_mount)
+            )
```
nemacysts, line 908: should we store this in the final serialized format instead of having to convert before using them?

VinaySagarGonabavi: This is the final format in which we expect it to be part of the spec, for reference in tests
```
{
                                            "mountPath": "/etc/vault/all_cas",
                                            "name": "vault-secrets",
                                            "readOnly": True,
                                        },
```

VinaySagarGonabavi: Oh, do you mean to store the global variable for `VTTABLET_EXTRA_VOLUME_MOUNTS` in a serialized format instead of List[V1VolumeMount]? I did that to avoid mypy errors when performing `extra_volume_mounts.append` operation

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Looking again at it, the get_volumes [method](https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L1003) of `InstanceConfig` has the uses_bulkdata_default argument but not SystemPaastaConfig [one](https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L2210)

paasta_tools/vitesscluster_tools.py
VinaySagarGonabavi: Yeah, looks like DockerVolume format is used in the parent classes, and I used this as required for Vitess manifest

matt-ullmer (approved):

PR #3963, opened 2024-09-18
Always use datetime.now(timezone.utc) in paasta status

nemacysts:
This will make it clear that we're always using UTC internally and avoids the footgun of using utcnow()

I've verified that the output for warming up pods is still the same in both this version of the code as well as in status quo - as well as the same for flink and kafka workloads

NOTE: there's still usages of utcnow or now() without an explicit timezone - but that's for another day

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -719,8 +719,11 @@ def should_job_info_be_shown(cluster_state):
 
 
 def get_pod_uptime(pod_deployed_timestamp: str):
-    pod_creation_time = datetime.strptime(pod_deployed_timestamp, "%Y-%m-%dT%H:%M:%SZ")
-    pod_uptime = datetime.utcnow() - pod_creation_time
+    # NOTE: the k8s API returns timestamps in UTC, so we make sure to always work in UTC
+    pod_creation_time = datetime.strptime(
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -1864,10 +1871,11 @@ def print_kafka_status(
             PaastaColors.green if broker["phase"] == "Running" else PaastaColors.red
         )
 
+        # NOTE: the k8s API returns timestamps in UTC, so we make sure to always work in UTC
         start_time = datetime.strptime(
             broker["deployed_timestamp"], "%Y-%m-%dT%H:%M:%SZ"
-        )
-        delta = datetime.utcnow() - start_time
+        ).replace(tzinfo=timezone.utc)
```
nemacysts, line 1877: this will add a +00:00 to broker started times - e.g., we'll go from 
```
Started
2024-09-18 01:11:36 (20 hours ago)
```

to

```
Started
2024-09-18 01:11:36+00:00 (20 hours ago)
```

but imo, this is a net positive since without looking at the parenthetical, it's entirely unclear what TZ is being used

jfongatyelp (approved):
Ty for all the breadcrumbs for future us so we don't make the same mistake again!

nleigh (approved):

PR #3964, opened 2024-09-19
Add CODEOWNERS for CI-owned files

nemacysts:
This is a first draft/stab at adding CODEOWNERS for CI-owned code in general across our OSS repositories - we can likely extend this in the future so that operator owners can CODEOWNERS their files to us + them :)

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -23,6 +23,9 @@ debian/rules @Yelp/paasta
 # this one is unfortunate since operators own the paasta status code for their
 # workloads, but atm all this is in one large file
 paasta_tools/cli/ @Yelp/paasta
+# exclude these files from being codeownered to us
+paasta_tools/cli/cmds/spark_run.py
+paasta_tools/cli/cmds/security_check.py
 
 
 # we don't own all the API code, but we do own most of it. if this turns into a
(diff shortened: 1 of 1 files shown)


CODEOWNERS
```diff
@@ -0,0 +1,96 @@
...
+
+## These contain entrypoints for PaaSTA or otherwise control entrypoint builds
+setup.py @Yelp/paasta
+debian/control @Yelp/paasta
+debian/paasta-tools.links @Yelp/paasta
+debian/paasta-tools.substvars @Yelp/paasta
+debian/rules @Yelp/paasta
+# this one is unfortunate since operators own the paasta status code for their
+# workloads, but atm all this is in one large file
+paasta_tools/cli/ @Yelp/paasta
```
jfongatyelp, line 25: Should we special case spark_run? security_check (is that still used or does that happen elsewhere now)?

jfongatyelp (approved):

CODEOWNERS
nemacysts: whoops, err - yea - i'm not quite sure how to exclude things without having N glob rules

...lemme see if a like that looks something like `paasta_tools/cli/spark_run.py` (i.e., no team) will work

nemacysts: looks like GH isn't complaining about the lines i added!

jfongatyelp (approved):

EmanElsaban (approved):

PR #3965, opened 2024-09-20
Upgrading k8s lib

wilmer05:
We need to update the kubernetes library to newer versions. 

Containerd gives problems when looking for container image names with the kubernetes client in the current version.

PAASTA-18345

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -605,12 +605,6 @@ class KubeClient:
         self.policy = kube_client.PolicyV1beta1Api(self.api_client)
         self.apiextensions = kube_client.ApiextensionsV1Api(self.api_client)
 
-        # We need to support apiextensions /v1 and /v1beta1 in order
-        # to make our upgrade to k8s 1.22 smooth, otherwise
-        # updating the CRDs make this script fail
-        # self.apiextensions_v1_beta1 = kube_client.ApiextensionsV1beta1Api(
-        #     self.api_client
-        # )
         self.custom = kube_client.CustomObjectsApi(self.api_client)
         self.autoscaling = kube_client.AutoscalingV2beta2Api(self.api_client)
         self.rbac = kube_client.RbacAuthorizationV1Api(self.api_client)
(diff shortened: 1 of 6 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -610,9 +608,9 @@ def __init__(
         # We need to support apiextensions /v1 and /v1beta1 in order
         # to make our upgrade to k8s 1.22 smooth, otherwise
         # updating the CRDs make this script fail
-        self.apiextensions_v1_beta1 = kube_client.ApiextensionsV1beta1Api(
-            self.api_client
-        )
+        # self.apiextensions_v1_beta1 = kube_client.ApiextensionsV1beta1Api(
```
ilkinmammadzada, line 611: was it dead code?

wilmer05: I removed it now

nemacysts: if we need this code again we can grab it from git history :)

paasta_tools/kubernetes_tools.py
```diff
@@ -4155,12 +4153,8 @@ def mode_to_int(mode: Optional[Union[str, int]]) -> Optional[int]:
...
 def update_crds(
     kube_client: KubeClient,
-    desired_crds: Collection[
-        Union[V1CustomResourceDefinition, V1beta1CustomResourceDefinition]
-    ],
-    existing_crds: Union[
-        V1CustomResourceDefinitionList, V1beta1CustomResourceDefinitionList
-    ],
+    desired_crds: Collection[Union[V1CustomResourceDefinition]],
+    existing_crds: Union[V1CustomResourceDefinitionList],
```
nemacysts, line 4151:
```suggestion
    desired_crds: Collection[V1CustomResourceDefinition],
    existing_crds: V1CustomResourceDefinitionList,
```
we don't need a union anymore since we're using a single type now :)

paasta_tools/setup_kubernetes_cr.py
```diff
@@ -175,7 +174,7 @@ def setup_all_custom_resources(
             ).items
         except ApiException:
             log.debug(
-                "Listing CRDs with apiextensions/v1 not supported on this cluster, falling back to v1beta1"
+                "Listing CRDs with apiextensions/v1 not supported on this cluster"
```
nemacysts, line 177: imo: i think we'd want to revert the changes to this file where we added the for loop rather than make this a for loop over a single version

nemacysts: (it would make the code more straightforward and while we may need to handle new apiextension versions in the future, imo it'd be better to only temporarily have this sort of multi-version code)

nemacysts (approved):

PR #3969, opened 2024-09-23
Update smartstack_schema.json to support minimum_error_rps for error alerting

andrewli611:
Update smartstack_schema.json to support minimum_error_rps for error alerting


thejoycekung (approved):

jfongatyelp (approved):
I'm guessing eventually these top-level keys & endpoint overrides could be in a definition so we don't have to repeat ourselves, but no need to block on that right now!

PR #3977, opened 2024-10-03
let enable_automated_redeploys default value be configurable

calvinli:
`paasta get-image-version` will output an empty string if the service has no deploy steps with `enable_automated_redeploys` enabled (for compatibility with pre-automated-redeploys behavior).

This introduces a new system config `enable_automated_redeploys_default`, itself defaulted to False, that controls whether `paasta get-image-version` treats a deploy step with no explicit `enable_automated_redeploys` as enabled or disabled.

(If set to True this doesn't actually cause automated redeploys to happen, that requires separate additional Jenkins changes.)

### Testing

I ran a random `paasta get-image-version -s <service> -c <commit>` for a service with no `enable_automated_redeploys` specified and confirmed that existing behavior continues (empty string with a stderr warning), and that you do get an image version if `enable_automated_redeploys_default` is set to True.


piax93 (approved):
++

jfongatyelp (approved):

PR #3979, opened 2024-10-10
MLCOMPUTE-2001 | Cleanup spark related logs from setup_tron_namespace

chi-yelp:
Includes the changes from https://github.com/Yelp/service_configuration_lib/pull/151, add option to mute some logs when calling functions from `setup_tron_namespace`, and fix some issues.


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -239,9 +239,9 @@ def add_subparser(subparsers):
             "default_pool"
         )
     except PaastaNotConfiguredError:
-        default_spark_cluster = "pnw-devc"
+        default_spark_cluster = "pnw-devc-spark"
         default_spark_pool = "batch"
-        valid_clusters = ["spark-pnw-prod", "pnw-devc"]
+        valid_clusters = ["pnw-devc-spark", "pnw-prod-spark"]
```
chi-yelp, line 244: To align the default values with the recent puppet changes for pool validation rules

paasta_tools/setup_tron_namespace.py
```diff
@@ -224,7 +224,7 @@ def main():
             # since we need to print out what failed in either case
             failed.append(service)
 
-    if args.bulk_config_fetch:
+    if args.dry_run and args.bulk_config_fetch:
```
chi-yelp, line 227: I guess we should also skip Tron API calls here in the dry run mode?

paasta_tools/tron_tools.py
```diff
@@ -380,16 +380,17 @@ def build_spark_config(self) -> Dict[str, str]:
...
                 f"Replacing spark.kubernetes.executor.podTemplateFile="
                 f"{spark_conf['spark.kubernetes.executor.podTemplateFile']} with "
                 f"spark.kubernetes.executor.podTemplateFile={spark_tools.SPARK_DNS_POD_TEMPLATE}"
             )
-            spark_conf[
-                "spark.kubernetes.executor.podTemplateFile"
-            ] = spark_tools.SPARK_DNS_POD_TEMPLATE
+        spark_conf[
+            "spark.kubernetes.executor.podTemplateFile"
+        ] = spark_tools.SPARK_DNS_POD_TEMPLATE
```
chi-yelp, line 393: Fix: we should set/overwrite the pod template file (for k8s dns mode, etc.) whether the pod template has already been set or not

paasta_tools/setup_tron_namespace.py
nemacysts: good catch!

paasta_tools/spark_tools.py
```diff
@@ -176,7 +176,9 @@ def inject_spark_conf_str(original_cmd: str, spark_conf_str: str) -> str:
     return original_cmd
 
 
-def auto_add_timeout_for_spark_job(cmd: str, timeout_job_runtime: str) -> str:
+def auto_add_timeout_for_spark_job(
+    cmd: str, timeout_job_runtime: str, silent: bool = False
+) -> str:
```
nemacysts, line 181: fwiw: for spark-drivers-on-k8s we should use the tron max_runtime config directly and get rid of this code entirely :)

nemacysts (approved):

paasta_tools/spark_tools.py
chi-yelp: Thanks for bringing this up, I think we can move this to be part of our custom spark-submit wrapper later, so it can be more easily monitored from the spark side and managed by spark configuration service & auto tuner, and also ensures the consistency of different types of spark deployments (tron, adhoc, jupyter).
(The current implementation of this also respects to the `max_runtime` tronfig from the caller side).

PR #3985, opened 2024-11-01
auth support for paasta APIs

piax93:
Support for an authn/z tween in the PaaSTA API (mostly inspired by previous work).

There's nothing too surprising here. Maybe the only unusual aspect is that I picked to make "dry-run" the default rather than the other way around. It makes things easier to roll out.

For how the codebase is structured loading the token transparently could be a bit of a pain, especially if we want to maintain a bunch of un-auth endpoints, so I just added the option to set the token header in the API client. From there in the CLI implementations that need auth we can do

```python
from paasta_tools.cli.utils import get_sso_service_auth_token
...
get_paasta_oapi_client(..., get_sso_service_auth_token())
```

### DIFF
--- a/paasta_tools/api/api.py
+++ b/paasta_tools/api/api.py
@@ -31,6 +31,7 @@ from wsgicors import CORS
 import paasta_tools.api
 from paasta_tools import kubernetes_tools
 from paasta_tools.api import settings
+from paasta_tools.api.tweens import auth
 from paasta_tools.api.tweens import profiling
 from paasta_tools.api.tweens import request_logger
 from paasta_tools.utils import load_system_paasta_config
@@ -79,6 +80,18 @@ def parse_paasta_api_args():
         default=4,
         help="Number of gunicorn workers to run",
     )
+    parser.add_argument(
+        "--auth-endpoint",
+        type=str,
+        default="",
+        help="External API authorization endpoint",
+    )
+    parser.add_argument(
+        "--auth-enforce",
+        action="store_true",
+        default=False,
+        help="Enforce API authorization",
+    )
     args = parser.parse_args()
     return args
 
@@ -105,6 +118,7 @@ def make_app(global_config=None):
 
     config.include("pyramid_swagger")
     config.include(request_logger)
+    config.include(auth)
 
     config.add_route(
         "flink.service.instance.jobs", "/v1/flink/{service}/{instance}/jobs"
@@ -256,6 +270,11 @@ def main(argv=None):
 
     if args.cluster:
         os.environ["PAASTA_API_CLUSTER"] = args.cluster
+
+    if args.auth_endpoint:
+        os.environ["PAASTA_API_AUTH_ENDPOINT"] = args.auth_endpoint
+        if args.auth_enforce:
+            os.environ["PAASTA_API_AUTH_ENFORCE"] = "1"
 
     gunicorn_args = [
         "gunicorn",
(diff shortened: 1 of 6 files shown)


danielpops (dismissed):

paasta_tools/api/api.py
nemacysts, line 1: just curious: if we had an endpoint that was protected by this, but was being called by automation - how would that work? `get_sso_service_auth_token()` makes it sound like there might be a 2fa challenge/sso login, which wouldn't really be possible for  something like jenkins

paasta_tools/api/tweens/auth.py
```diff
@@ -0,0 +1,124 @@
...
+                        "service": service,
+                    },
+                },
+                timeout=2,
+            ).json()
+        except Exception as e:
+            logger.exception(f"Issue communicating with auth endpoint: {e}")
+            return AuthorizationOutcome(False, "Auth backend error")
+
+        if "result" not in response or "allowed" not in response["result"]:
```
nemacysts, line 105: not a blocker: might be nice to have a TypedDict or something for the response type (or a link to somewhere that documents the format)

nemacysts: (until reading further ahead, i thought we were checking if the string allowed was in the result payload - but we're really checking that there's an allowed key in the response["result"] dictionary)

nemacysts: (i guess this could also be something like 
```Python
if response.get("result", {}).get("allowed") is None:
    ...
```
to make it clearer that we're expecting dicts)

tests/api/tweens/test_auth.py
```diff
@@ -0,0 +1,106 @@
...
+def mock_auth_tween():
+    with patch.dict(
+        os.environ,
+        {
+            "PAASTA_API_AUTH_ENFORCE": "1",
+            "PAASTA_API_AUTH_ENDPOINT": "http://localhost:31337",
+        },
+    ):
+        with patch("paasta_tools.api.tweens.auth.requests"):
+            yield auth.AuthTweenFactory(MagicMock(), MagicMock())
```
nemacysts, line 34: not a blocker, but i wonder if these magicmocks should have spec/spec_set set to their respective mockee values

tests/api/tweens/test_auth.py
```diff
@@ -0,0 +1,106 @@
...
+
+def test_call(mock_auth_tween):
+    mock_request = MagicMock(
+        path="/something",
+        method="post",
+        headers={"Authorization": "Bearer aaa.bbb.ccc"},
+        swagger_data={"service": "foobar"},
+    )
+    with patch.object(mock_auth_tween, "is_request_authorized") as mock_is_authorized:
+        mock_is_authorized.return_value = auth.AuthorizationOutcome(True, "Ok")
```
nemacysts, line 45: iirc, `patch.object` has a `return_value` kwarg we can do to set this where we create the mock - but not a blocker

paasta_tools/api/api.py
piax93: That's an open question, but the current plan is to just apply this to the new remote-run endpoints which are going to be implemented to run prod sandboxes, so anything else will remain open. That way we can collect telemetry and see if it's worth the effort to lock down more endpoints.

For the specific case of Jenkins, since we run that in k8s, if we decide to enforce authz it most likely going to be via a service account token mounted into the pods.

tests/api/tweens/test_auth.py
piax93: yep, I know, but I hate having long `with` statements so I usually opt to just update the mock object directly

nemacysts (approved):

PR #3986, opened 2024-11-02
work in progress - remote run

Qmando:
**Work in progress, not fully functional!**

Create a new API, /service/{service}/{instance}/remote_run

Currently accepts additional a username, whether it's interactive, and whether it should recreate the deployment

Launches a deployment with `-remote-run-{username}`. If interactive, replaces command with sleep. If deployment exists, maybe terminate it and relaunch. Returns the name of the pod and the namespace.

Created CLI cmd to interact with this api. Gives you a kubectl command to run to get a shell in the container.


paasta_tools/api/api_docs/oapi.yaml
```diff
@@ -1627,6 +1627,52 @@ paths:
...
+              type: object
+              properties:
+                interactive:
+                  type: bool
+                user:
+                  type: string
+                image:
+                  type: string
+              required:
+               - interactive
```
Qmando, line 1646: Still figuring out what parameters will be needed, so this doesn't match swagger right now

paasta_tools/cli/cmds/remote_run_2.py
```diff
@@ -0,0 +1,180 @@
...
+    print("Run the following command to enter your service pod")
+    print(exec_command)
+
+    return ret_code
+
+
+def remote_run(args) -> int:
+    """Run stuff, but remotely!"""
+    system_paasta_config = load_system_paasta_config(
+        "/nail/home/qlo/paasta_config/paasta/"
```
Qmando, line 176: I was using this to override the API endpoint

paasta_tools/paasta_remote_run_2.py
```diff
@@ -0,0 +1,101 @@
...
+    )
+
+
+def wait_until_deployment_gone(kube_client, namespace, deployment_name):
+    for retry in range(10):
+        pod = find_pod(kube_client, namespace, deployment_name, 1)
+        if not pod:
+            return
+        sleep(5)
+    raise Exception("Pod still exists!")
```
Qmando, line 64: This doesn't work. The request times out before the deployment terminates. I think we'd need to return and the client needs to wait and retry.

paasta_tools/cli/cmds/remote_run_2.py
```diff
@@ -0,0 +1,180 @@
...
+    pod_name, namespace = response["pod_name"], response["namespace"]
+    exec_command_tmpl = "kubectl{eks}-{cluster} exec -it -n {namespace} {pod} /bin/bash"
+    exec_command = exec_command_tmpl.format(
+        eks="-eks" if is_eks else "", cluster=cluster, namespace=namespace, pod=pod_name
+    )
+    print("Pod launched successfully:", pod_name)
+
+    # TODO figure out how to get this to work
+    # print('Attaching shell')
+    # cmd = subprocess.Popen(exec_command.split(' '))
```
Qmando, line 166: This results in `error: input/output error`

danielpops: `subprocess.run` is what I used in my POC and it seemed to work

paasta_tools/api/api.py
```diff
@@ -148,6 +148,10 @@ def make_app(global_config=None):
         "service.instance.tasks.task",
         "/v1/services/{service}/{instance}/tasks/{task_id}",
     )
+    config.add_route(
+        "service.instance.remote_run",
+        "/v1/services/{service}/{instance}/remote_run",
+    )
```
piax93, line 154: for how the authz design works, it'd be easier if this wasn't under the `/v1/services` umbrella

paasta_tools/cli/cmds/remote_run_2.py
piax93: Ideally it'd be nice to have a tty forwarded over a socket here, to keep k8s authz out of the way. Or maybe we can sort out a way to cut a short-lived token that can do the exec, and use that for the kubectl invocation. The second should definitely be much easier to implement.

piax93: Yeah, we can have a service account that can exec in all namespaces, and then create a new API endpoints that looks something like `/v1/remote_run/{service}/{instance}/token`, and return a token in the right namespace, with an expiration of just a few seconds. So it can be used for the exec then thrown away.

paasta_tools/paasta_remote_run_2.py
```diff
@@ -0,0 +1,101 @@
...
+    )
+    namespace = deployment.get_namespace()
+
+    # Set to interactive mode
+    if interactive:
+        deployment.config_dict["cmd"] = "sleep 604800"  # One week
+
+    # Create the app with a new name
+    formatted_application = deployment.format_kubernetes_app()
+    formatted_application.metadata.name += f"-remote-run-{user}"
```
piax93, line 33: there's the risk of the pod name being truncated, so we'll definitely also want to indicators of this being a remote-run in the pod's labels

paasta_tools/paasta_remote_run_2.py
```diff
@@ -0,0 +1,101 @@
...
+    # TODO Overriding the kube client config for now as the api has limited permissions
+    kube_client = KubeClient(config_file="/etc/kubernetes/admin.conf")
+
+    # TODO hardcoded for now
+    is_eks = False
+
+    # Load the service deployment settings
+    deployment = load_kubernetes_service_config_no_cache(
+        service, instance, cluster, DEFAULT_SOA_DIR
+    )
```
piax93, line 24: maybe we could run this as a "Job" rather than a deployment, so we can set a hard deadline for it directly in the template with `activeDeadlineSeconds`

https://kubernetes.io/docs/concepts/workloads/controllers/job/

paasta_tools/kubernetes_tools.py
```diff
@@ -2029,6 +2034,62 @@ def get_pod_management_policy(self) -> str:
...
+                api_version="batch/v1",
+                kind="Job",
+                metadata=self.get_kubernetes_metadata(git_sha),
+                spec=V1JobSpec(
+                    active_deadline_seconds=3600,
+                    selector=V1LabelSelector(
+                        match_labels={
+                            "paasta.yelp.com/service": self.get_service(),
+                            "paasta.yelp.com/instance": self.get_instance(),
+                        }
```
Qmando, line 2053: todo: This was copied over from the Deployment function below and is causing errors.

paasta_tools/kubernetes_tools.py
```diff
@@ -2029,6 +2034,62 @@ def get_pod_management_policy(self) -> str:
...
+        try:
+            docker_url = self.get_docker_url()
+            git_sha = get_git_sha_from_dockerurl(docker_url, long=True)
+            system_paasta_config = load_system_paasta_config()
+            complete_config = V1Job(
+                api_version="batch/v1",
+                kind="Job",
+                metadata=self.get_kubernetes_metadata(git_sha),
+                spec=V1JobSpec(
+                    active_deadline_seconds=3600,
```
piax93, line 2048: I guess we'll need to propagate this somehow from a parameter

paasta_tools/kubernetes_tools.py
piax93: mmm, yeah, I don't see why you would need a select at this stage

paasta_tools/kubernetes_tools.py
```diff
@@ -2029,6 +2034,62 @@ def get_pod_management_policy(self) -> str:
...
+            )
+            complete_config.metadata.labels["yelp.com/paasta_config_sha"] = config_hash
+            complete_config.metadata.labels["paasta.yelp.com/config_sha"] = config_hash
+
+            complete_config.spec.template.metadata.labels[
+                "yelp.com/paasta_config_sha"
+            ] = config_hash
+            complete_config.spec.template.metadata.labels[
+                "paasta.yelp.com/config_sha"
+            ] = config_hash
```
piax93, line 2087: this stuff is very related to deployment bounces, so I'd say we don't need it in this case

paasta_tools/kubernetes_tools.py
```diff
@@ -4354,6 +4425,20 @@ def get_kubernetes_secret_env_variables(
     return decrypted_secrets
 
 
+def create_temp_exec_token(kube_client: KubeClient, namespace: str, user: str):
+    """Create a short lived token for exec"""
+    audience = "remote-run-" + user
+    service_account = "default"
+    token_spec = V1TokenRequestSpec(
+        expiration_seconds=600, audiences=[audience]  # minimum allowed by k8s
```
piax93, line 4433: The audience will need to be the kubernetes one, otherwise it won't be usable for doing execs. You should be able to just leave it unset to achieve that.

paasta_tools/kubernetes_tools.py
```diff
@@ -4354,6 +4425,20 @@ def get_kubernetes_secret_env_variables(
     return decrypted_secrets
 
 
+def create_temp_exec_token(kube_client: KubeClient, namespace: str, user: str):
+    """Create a short lived token for exec"""
+    audience = "remote-run-" + user
+    service_account = "default"
```
piax93, line 4431: This will need to come from the main paasta config. We'll have to create a user that can just exec in to places.

paasta_tools/kubernetes_tools.py
```diff
@@ -4354,6 +4452,19 @@ def get_kubernetes_secret_env_variables(
     return decrypted_secrets
 
 
+def create_temp_exec_token(kube_client: KubeClient, namespace: str, user: str):
```
piax93, line 4455: The `user` parameter is not used here. Is the plan to include it in the service account name if we go for creating ephemeral service accounts?

piax93:
This has been fixed and refactored in more sensible chunks

PR #3991, opened 2024-12-04
Disallow having service account + iam_role

jfongatyelp:
Since we have to manage the service account (and its pod identity annotation) if someone specifies an `iam_role` for their instance, this is probably incompatible with specifying a service_account.

Technically nothing would break with both, we just wouldn't do anything with the iam_role they'd provided and it would be more difficult for them to troubleshoot iam issues. 

This will give a validation error like the following:
Screenshot 2024-12-04 at 11 27 25 AM

And for tron: 
Screenshot 2024-12-04 at 1 10 31 PM

### DIFF
--- a/paasta_tools/cli/schemas/kubernetes_schema.json
+++ b/paasta_tools/cli/schemas/kubernetes_schema.json
@@ -93,6 +93,14 @@
                             ]
                         }
                     ]
+                },
+                {
+                    "not": {
+                        "required": [
+                            "service_account_name",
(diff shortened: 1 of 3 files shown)


nemacysts (approved):
whoops, i think i brought this up once but i was a bad citizen and didn't fix or ticket this - thanks for being more diligent!

that said, the tron schema probably needs the same fix :)

nemacysts (approved):

PR #3995, opened 2024-12-16
Add uses bulkdata argument to paasta spark run

timmow:
This makes the change to paasta spark run so that
https://github.yelpcorp.com/sysgit/yelpsoa-configs/pull/52010 will work as expected

I'm not checking here if the /nail/bulkdata volume is specified in the spark config, e.g
`spark.kubernetes.executor.volumes.hostPath.0.mount.path=/nail/bulkdata`
- doing this and setting uses_bulkdata set to True would result in multiple docker volumes being set which would cause a failure.

This follows on from [this conversation in
slack](https://yelp.slack.com/archives/CA8BWU65D/p1729768030212919) and will allow us to complete [this
project](https://yelpwiki.yelpcorp.com/display/PRODENG/Project+Incredible+Bulk)

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -350,6 +350,13 @@ def add_subparser(subparsers):
         default=False,
     )
 
+    list_parser.add_argument(
+        "--uses-bulkdata",
+        help="Mount /nail/bulkdata in the container",
+        action="store_true",
+        default=False,
+    )
+
     aws_group = list_parser.add_argument_group(
         title="AWS credentials options",
         description="If --aws-credentials-yaml is specified, it overrides all "
@@ -784,6 +791,9 @@ def configure_and_run_docker_container(
         volumes = get_volumes_from_spark_k8s_configs(spark_conf)
     else:
         raise UnsupportedClusterManagerException(cluster_manager)
+
+    if args.uses_bulkdata:
+        volumes.append("/nail/bulkdata:/nail/bulkdata:ro")
 
     volumes.append("%s:rw" % args.work_dir)
     volumes.append("/nail/home:/nail/home:rw")
--- a/tests/cli/test_cmds_spark_run.py
+++ b/tests/cli/test_cmds_spark_run.py
@@ -429,6 +429,7 @@ class TestConfigureAndRunDockerContainer:
         "fake_dir",
     )
 
+    @pytest.mark.parametrize("uses_bulkdata", [True, False])
     @pytest.mark.parametrize(
         ["cluster_manager", "spark_args_volumes", "expected_volumes"],
         [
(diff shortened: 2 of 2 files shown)


nemacysts (commented):
lgtm (and i think it's perfectly fine to not check if there's a spark config for mounting `/nail/bulkdata` since it looks like that's not something anyone is currently doing - and i doubt any of our spark users would add such a mount un-prompted)

that said i'll let someone for ml compute ship since they own this file in its entirety :)

chi-yelp (dismissed):
Lgtm. Agree that as Luis mentioned no need to check if the user uses Spark options to mount bulkdata, and I also double checked that no one is doing that

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -350,6 +350,13 @@ def add_subparser(subparsers):
         default=False,
     )
 
+    list_parser.add_argument(
+        "--uses-bulkdata",
+        help="Mount /nail/bulkdata in the container",
+        action="store_true",
+        default=False,
```
SuperMatt, line 357: Should we not set the default to true for now, then roll out my change to add the flag everywhere, and then set the default to false?

chi-yelp: hmm I checked using the following command and it seems `/nail/bulkdata` isn't mounted:
```bash
paasta spark-run --aws-profile=dev --cmd bash
```

timmow: so volumes are found from the instance config [here](https://sourcegraph.yelpcorp.com/Yelp/paasta@8ec1f6ba0250ca120c076e4a078d398893e67ed3/-/blob/paasta_tools/cli/cmds/spark_run.py?L1208) which is then passed to spark conf [here](https://sourcegraph.yelpcorp.com/Yelp/paasta@8ec1f6ba0250ca120c076e4a078d398893e67ed3/-/blob/paasta_tools/cli/cmds/spark_run.py?L1245) and then later referenced [here](https://sourcegraph.yelpcorp.com/Yelp/paasta@8ec1f6ba0250ca120c076e4a078d398893e67ed3/-/blob/paasta_tools/cli/cmds/spark_run.py?L784)

Meaning it is important that I check if `/nail/bulkdata` is in the list of volumes before adding it. And the command @chi-yelp  shared is not mounting bulkdata because its not specifying a service / instance that uses bulkdata - @chi-yelp was that command run on my branch?

chi-yelp: Sorry for the late reply, yes I ran the command on the branch of this PR after creating the virtualenv by `make dev`.

The mount paths will be deduplicated [here](https://sourcegraph.yelpcorp.com/Yelp/service_configuration_lib@db10fa6d60d59082da8825fd82f3e25d6abcd4dd/-/blob/service_configuration_lib/spark_config.py?L227) by a dict, but I think it's also good for checking the input anyway

timmow:
closing in favor of #4005

PR #3997, opened 2024-12-20
PAASTA-18479: Add list-namespaces tool

jfongatyelp:
Our iam_role configs depend on role owners being able to fill out which k8s namespaces their role is going to be used in in `pod_identity_namespaces`, but paasta normally is figuring out that translation, and users don't usually need to care about this info.

This `list-namespaces` tool should just spit out the namespaces their service will be using, in list format to match what our iam_roles module needs.

I don't expect people will need to specify the instance/cluster filters at all since roles can be used across those lines, but this command leaves the hard work of filtering on those values to the already-existing `get_instance_configs_for_service` util. 

The tests already capture most of the list-namespaces specific logic we want, but testing this against some example services seems to also work as expected:
```
$ paasta list-namespaces -s ad_event
['paasta-spark', 'tron']

$ paasta list-namespaces -s ad_event -i ad_event.wait_for_ad_opportunity # not a spark action
['tron']

$ paasta list-namespaces -s ad_event -i ad_event.wait_for_ad_opportunity -c pnw-devc # Currently no errors thrown if filters do not match a real instance
[]
```

There are also 2 non-list-namespaces changes in here:
1. LongRunningServiceConfig used to throw a `NotImplementedError` for `get_namespace()`, but by me deleting that override, LongRunningServiceConfig.get_namespace() will use [InstanceConfig.get_namespace()](https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L452-L457) which uses [INSTANCE_TYPE_TO_K8S_NAMESPACE](https://github.com/Yelp/paasta/blob/master/paasta_tools/utils.py#L149-L159) to return the correct value.  I think the exception may not be necessary, as from a search of cases of us catching the `NotImplementedError`, we never expect to do $something if it was due to `get_namespaces()` throwing this error.
  * This _will_ actually stop throwing exceptions for anything trying to run `get_namespace` on a LongRunningServiceConfig subclass that does not define this method, such as Flink or NRTSearch.  However, I didn't find any code that expected to hit an exception when using get_namespace, as they mostly related to secret syncing, which special-case operator namespaces already. In fact, by allowing flink instances to actually `get_namespace`, we may be able to remove the special-case sync jobs later!
  * From original discussion when we started supporting arbitrary namespaces ([PR](https://github.com/Yelp/paasta/pull/3548#discussion_r1103221893)) on the `get_namespace` method, it seems like we generally want this to come from InstanceConfig anyhow, using default values if no override is specified, so this seems fine. 
  * PaaSTA services will still default to the correct default of paastasvc-* since we override get_namespace [here](https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes_tools.py#L2022-L2026)
2. I noticed some copypaasta in the `autoscale` cli cmd's help text, so I s/stop/autoscale/ where I found this issue.

### DIFF
--- a/paasta_tools/cli/cmds/list_namespaces.py
+++ b/paasta_tools/cli/cmds/list_namespaces.py
@@ -15,6 +15,7 @@
 from paasta_tools.cli.utils import get_instance_configs_for_service
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import validate_service_name
+from paasta_tools.spark_tools import SPARK_EXECUTOR_NAMESPACE
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import list_services
@@ -69,8 +70,14 @@ def paasta_list_namespaces(args):
         # We skip non-k8s instance types
         if instance.get_instance_type() in ("paasta-native", "adhoc"):
             continue
-        else:
-            namespaces.add(instance.get_namespace())
+        namespaces.add(instance.get_namespace())
+        # Tron instances are TronActionConfigs
+        if (
+            instance.get_instance_type() == "tron"
+            and instance.get_executor() == "spark"
+        ):
+            # We also need paasta-spark for spark executors
+            namespaces.add(SPARK_EXECUTOR_NAMESPACE)
 
     # Print in list format to be used in iam_roles
     print(list(namespaces))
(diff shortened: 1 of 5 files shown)


paasta_tools/cli/cmds/list_namespaces.py
```diff
@@ -0,0 +1,84 @@
...
+    )
+    for instance in instance_configs:
+        # We skip non-k8s instance types
+        if instance.get_instance_type() in ("paasta-native", "adhoc"):
+            continue
+        namespaces.add(instance.get_namespace())
+        # Tron instances are TronActionConfigs
+        if (
+            instance.get_instance_type() == "tron"
+            and instance.get_executor() == "spark"
```
nemacysts, line 77: i think it might be possible for batch-box-launched executors to use pod identity, but imo it's fine to leave them out u!less this becomes a problem (especially since things are actively migrating away from batch boxes)

nemacysts (approved):

PR #3999, opened 2024-12-23
Minor fixes for `paasta info`

jfongatyelp:
1. stop displaying `tcp://` urls for tcp services under `deployments` list; access via smartstack port is no longer supported thru paasta proxies
2. replace `paasta-$cluster.yelp` domains with the appropriate `proxy.$cluster.paasta` domains instead
3. make it clearer the urls provided for the 'deployments' are for testing purposes only, not for production/client access
4. Replace long-dead uchiwa dashboard link w/ y/sl2 dashboard instead
5. fix tests

### DIFF
--- a/paasta_tools/cli/cmds/info.py
+++ b/paasta_tools/cli/cmds/info.py
@@ -94,14 +94,9 @@ def get_deployments_strings(service: str, soa_dir: str) -> List[str]:
         )
         service_mode = service_config.get_mode()
         for cluster in deployments_to_clusters(deployments):
-            if service_mode == "tcp":
-                service_port = service_config.get("proxy_port")
+            if service_mode == "http" or service_mode == "https":
                 link = PaastaColors.cyan(
-                    "%s://paasta-%s.yelp:%d/" % (service_mode, cluster, service_port)
-                )
-            elif service_mode == "http" or service_mode == "https":
-                link = PaastaColors.cyan(
-                    f"{service_mode}://{service}.paasta-{cluster}.yelp/"
+                    f"{service_mode}://{service}.proxy.{cluster}.paasta/"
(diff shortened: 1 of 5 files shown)


paasta_tools/cli/cmds/info.py
```diff
@@ -94,14 +94,9 @@ def get_deployments_strings(service: str, soa_dir: str) -> List[str]:
         )
         service_mode = service_config.get_mode()
         for cluster in deployments_to_clusters(deployments):
-            if service_mode == "tcp":
-                service_port = service_config.get("proxy_port")
+            if service_mode == "http" or service_mode == "https":
```
nemacysts, line 97:
```suggestion
            if service_mode in {"http", "https"}:
```
or 

```
TESTABLE_SERVICE_MODES = {"http", "https"}
...
            if service_mode in TESTABLE_SERVICE_MODES:
```

would be a little less repetitive, but this is fine as-is too :)

paasta_tools/cli/cmds/info.py
```diff
@@ -111,8 +106,7 @@ def get_deployments_strings(service: str, soa_dir: str) -> List[str]:
 
 def get_dashboard_urls(service):
     output = [
-        " - %s (Sensu Alerts)"
-        % (PaastaColors.cyan("https://uchiwa.yelpcorp.com/#/events?q=%s" % service))
+        " - %s (service load y/sl2)" % (PaastaColors.cyan(f"http://y/{service}_load"))
```
nemacysts, line 109: once we can run on a newer python, we can use nested f-strings here :D

nemacysts (dismissed):

nemacysts (approved):

PR #4002, opened 2025-01-08
Support for HTTP2 mode & updated SmartStack schema

ankit864:
Support for HTTP2 mode & updated SmartStack schema.

### DIFF
--- a/paasta_tools/cli/schemas/smartstack_schema.json
+++ b/paasta_tools/cli/schemas/smartstack_schema.json
@@ -116,6 +116,7 @@
             "mode": {
                 "enum": [
                     "http",
+                    "http2",
                     "https",
                     "tcp"
                 ]
--- a/paasta_tools/long_running_service_tools.py
+++ b/paasta_tools/long_running_service_tools.py
@@ -110,7 +110,7 @@ class ServiceNamespaceConfig(dict):
                 return None
             else:
                 return "http"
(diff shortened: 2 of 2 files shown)


jfongatyelp (dismissed):

jfongatyelp:
There may be some other pieces that need updating, e.g. https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes_tools.py#L1374-L1389 this seems to only cycle thru all previously-valid service 'mode's :( 

You might need to actually run setup_kubernetes_job to find all those places tho

jfongatyelp (dismissed):

nemacysts:
related to ^ - probably also want to figure out what we wanna do with: 
* https://github.com/Yelp/paasta/blob/ff98b5c0b4baca76947294f03f41e89d2080bdc4/paasta_tools/cli/cmds/info.py#L39-L41
* https://github.com/Yelp/paasta/blob/ff98b5c0b4baca76947294f03f41e89d2080bdc4/paasta_tools/long_running_service_tools.py#L426-L427

otherwise, i think these 2 (+ the link @jfongatyelp provided) might be all? (that said, i agree with jfong that running skj with these changes would probably be the best/quickest way to see what else might be missing)

ankit864:
https://github.com/Yelp/paasta/pull/4004 sorry reopened different PR. 

I have fixed the remaining places where mode is required to check all value.

PR #4005, opened 2025-01-14
Add uses_bulkdata argument to paasta spark run instance_config

SuperMatt:
This makes the change to paasta spark run so that https://github.yelpcorp.com/sysgit/yelpsoa-configs/pull/52010 will work as expected.

This works by adding the uses_bulkdata key to the intsance config if the spark job has the key present and set to true.

I have added this arg to the tests so that they pass, and also created a test so that we can check all the different ways that uses_bulkdata can be set, either on paasta spark-run as an argument, or in the instance config.

See #3995 for more information about why we're doing this.

### DIFF
--- a/tests/cli/test_cmds_spark_run.py
+++ b/tests/cli/test_cmds_spark_run.py
@@ -1030,6 +1030,7 @@ def test_paasta_spark_run_bash(
         tronfig=None,
         job_id=None,
         use_web_identity=False,
+        uses_bulkdata=True,
     )
     mock_load_system_paasta_config_utils.return_value.get_kube_clusters.return_value = (
         {}
@@ -1156,6 +1157,7 @@ def test_paasta_spark_run(
         tronfig=None,
         job_id=None,
         use_web_identity=False,
+        uses_bulkdata=True,
     )
     mock_load_system_paasta_config_utils.return_value.get_kube_clusters.return_value = (
         {}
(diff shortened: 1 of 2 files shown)


tests/cli/test_cmds_spark_run.py
```diff
@@ -1358,6 +1361,89 @@ def test_paasta_spark_run_pyspark(
...
+
+    mock_get_instance_config.return_value.config_dict = {
+        "uses_bulkdata": instance_config_uses_bulkdata
+    }
+
+    spark_run.paasta_spark_run(args)
+
+    assert (
+        mock_get_instance_config.return_value.config_dict["uses_bulkdata"] == expected
+    )
```
nemacysts, line 1444: we've mocked this on L1436 - so this assertion isn't doing much atm (I guess right now this test is really just checking that `spark_run.paasta_spark_run(args)` doesn't crash if we remove this always-passing assertion?)

imo, deleting this test is fine (the new logic is pretty trivial) or updating the test to assert that the correct mounts are in place in the final command would be where we should go from here :)

SuperMatt:
Discussing with @nemacysts, we think that we don't need a test that the volume is present because we're testing that uses_bulkdata ends up on the instance config, but then we have another test already for the volume being added when uses_bulkdata is set to true, which can be seen [here](https://sourcegraph.yelpcorp.com/Yelp/paasta/-/blob/tests/test_utils.py?L2154). Adding a spark test for this volume being included would be redundent.

chi-yelp (approved):

PR #4007, opened 2025-01-25
PAASTA-18383: Update V2Beta2 HPA api references to V2 for 1.26

jfongatyelp:
As per https://kubernetes.io/docs/reference/using-api/deprecation-guide/#horizontalpodautoscaler-v126 the V2beta2 version of the HorizontalPodAutoscaler api is being deprecated in 1.26. 

This updates as per recommendations to v2, and as all of our clusters are >1.23 they should support it. 
Note that we did not use `targetAverageUtilization` that has been switched to `target.type=Utilization`+`target.averageUtilization`, as we'd already used the latter in our code.

This has been tested in paasta playground and infrastage and s_k_j was able to create/modify the appropriate HPA objects and autoscaling was successful. 

Confirmed with audit logs that paasta was sending the request to `autoscaling/v2` rather than `autoscaling/v2beta2` with the change.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -118,16 +118,16 @@ from kubernetes.client import V1Volume
 from kubernetes.client import V1VolumeMount
 from kubernetes.client import V1VolumeProjection
 from kubernetes.client import V1WeightedPodAffinityTerm
-from kubernetes.client import V2beta2CrossVersionObjectReference
-from kubernetes.client import V2beta2HorizontalPodAutoscaler
-from kubernetes.client import V2beta2HorizontalPodAutoscalerCondition
-from kubernetes.client import V2beta2HorizontalPodAutoscalerSpec
-from kubernetes.client import V2beta2MetricIdentifier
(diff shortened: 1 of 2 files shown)


wilmer05 (approved):

paasta_tools/kubernetes_tools.py
```diff
@@ -237,19 +237,19 @@
 # For detail, https://github.com/kubernetes-client/python/issues/553
 # This hack should be removed when the issue got fixed.
 # This is no better way to work around rn.
-class MonkeyPatchAutoScalingConditions(V2beta2HorizontalPodAutoscalerStatus):
+class MonkeyPatchAutoScalingConditions(V2HorizontalPodAutoscalerStatus):
```
nemacysts, line 240: i kinda wonder if we still need this - the linked GH issue fizzled out a while back,
but hopefully it's actually been fixed since we added this!

nemacysts (approved):

paasta_tools/kubernetes_tools.py
jfongatyelp: Good call, I will take a look at this and remove if we can!

PR #4009, opened 2025-02-24
Set value for uses_bulkdata_default to False

SuperMatt:
This changes the default value of uses_bulkdata to false, meaning that we can then remove the puppet hiera value which makes the default false that way. Everywhere that puppet is installed has a default value of False, so this shouldn't require a big bounce.

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -999,7 +999,9 @@ class InstanceConfig:
         return self.config_dict.get("net", "bridge")
 
     def get_volumes(
-        self, system_volumes: Sequence[DockerVolume], uses_bulkdata_default: bool = True
+        self,
+        system_volumes: Sequence[DockerVolume],
+        uses_bulkdata_default: bool = False,
     ) -> List[DockerVolume]:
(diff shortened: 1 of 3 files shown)


timmow (dismissed):

nemacysts (dismissed):
if we don't anticipate ever changing this default again, i'd say we could further clean this up and hardcode the default to false

timmow (approved):
@nemacysts the plan is to loop back and remove this once we've removed defaults from puppet in https://github.yelpcorp.com/sysgit/puppet/pull/13980 - but doing that without this change would cause a big bounce

nemacysts:
> the plan is to loop back and remove this once we've removed defaults from puppet in https://github.yelpcorp.com/sysgit/puppet/pull/13980 - but doing that without this change would cause a big bounce

@timmow sorry, I meant that we could get rid of `get_uses_bulkdata_default()` and replace any calls to that with `False` - then the puppet code can be removed without needing to come back to this file :)

nemacysts (approved):

PR #4010, opened 2025-02-25
MLCOMPUTE-2733 | Add paasta names to spark driver pod annotations

chi-yelp:
Add paasta names to spark driver pod annotations.
Follow up by: https://github.com/Yelp/service_configuration_lib/pull/153 which is for executor pod annotations.

### DIFF
--- a/paasta_tools/spark_tools.py
+++ b/paasta_tools/spark_tools.py
@@ -235,9 +235,18 @@ def get_spark_driver_monitoring_annotations(
     Returns Spark driver pod annotations - currently used for Prometheus metadata.
     """
     ui_port_str = str(spark_config.get("spark.ui.port", ""))
+    paasta_service_non_truncated = spark_config.get(
+        "spark.kubernetes.executor.annotation.paasta.yelp.com/service", ""
+    )
+    paasta_instance_non_truncated = spark_config.get(
+        "spark.kubernetes.executor.annotation.paasta.yelp.com/instance", ""
+    )
+
     annotations = {
         "prometheus.io/port": ui_port_str,
         "prometheus.io/path": "/metrics/prometheus",
+        "paasta.yelp.com/service": paasta_service_non_truncated,
+        "paasta.yelp.com/instance": paasta_instance_non_truncated,
     }
     return annotations
 
--- a/paasta_tools/spark_tools.py
+++ b/paasta_tools/spark_tools.py
@@ -265,3 +265,9 @@ def get_spark_driver_monitoring_labels(
         "spark.yelp.com/driver_ui_port": ui_port_str,
     }
     return labels
+
+
+def docker_volumes_to_mappings(
+    docker_volumes: List[DockerVolume],
+) -> List[Mapping[str, str]]:
+    return [{k: str(v) for k, v in volume.items()} for volume in docker_volumes]
--- a/paasta_tools/spark_tools.py
+++ b/paasta_tools/spark_tools.py
@@ -234,20 +234,23 @@ def get_spark_driver_monitoring_annotations(
     """
     Returns Spark driver pod annotations - currently used for Prometheus metadata.
     """
-    ui_port_str = str(spark_config.get("spark.ui.port", ""))
+    annotations: Dict[str, str] = dict()
+
+    ui_port_str = spark_config.get("spark.ui.port", "")
+    if ui_port_str != "":
+        annotations["prometheus.io/port"] = ui_port_str
+        annotations["prometheus.io/path"] = "/metrics/prometheus"
+
     paasta_service_non_truncated = spark_config.get(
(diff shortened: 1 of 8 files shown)


paasta_tools/spark_tools.py
```diff
@@ -235,9 +235,18 @@ def get_spark_driver_monitoring_annotations(
     Returns Spark driver pod annotations - currently used for Prometheus metadata.
     """
     ui_port_str = str(spark_config.get("spark.ui.port", ""))
+    paasta_service_non_truncated = spark_config.get(
+        "spark.kubernetes.executor.annotation.paasta.yelp.com/service", ""
+    )
+    paasta_instance_non_truncated = spark_config.get(
+        "spark.kubernetes.executor.annotation.paasta.yelp.com/instance", ""
+    )
```
nemacysts, line 243: i kinda wonder if we should return an empty dict if these aren't present (or error out?) - at least for the port missing, i think that would make sense since without a port these will be unscrapeable. but, having "" annotation values doesn't seem particularly useful (and shouldn't really happen anyway)

chi-yelp: These annotations will also be used by other things, such as OTel log processing, cost attribution (info from KSM), etc., so I think it might still make sense to populate anything that is available?

chi-yelp: Split it to two parts as discussed 👍 prometheus scraping and paasta service + instance

paasta_tools/spark_tools.py
```diff
@@ -234,11 +234,23 @@ def get_spark_driver_monitoring_annotations(
...
+
+    paasta_service_non_truncated = spark_config.get(
+        "spark.kubernetes.executor.annotation.paasta.yelp.com/service", ""
+    )
+    paasta_instance_non_truncated = spark_config.get(
+        "spark.kubernetes.executor.annotation.paasta.yelp.com/instance", ""
+    )
+    if paasta_service_non_truncated != "" and paasta_instance_non_truncated != "":
+        annotations["paasta.yelp.com/service"] = paasta_service_non_truncated
+        annotations["paasta.yelp.com/instance"] = paasta_instance_non_truncated
```
nemacysts, line 252:
```suggestion
    annotations: Dict[str, str] = {}

    ui_port_str = spark_config.get("spark.ui.port")
    if ui_port_str:
        annotations.update(
            {
                "prometheus.io/port": ui_port_str,
                "prometheus.io/path": "/metrics/prometheus",
            }
        )
           
    paasta_service_non_truncated = spark_config.get(
        "spark.kubernetes.executor.annotation.paasta.yelp.com/service"
    )
    paasta_instance_non_truncated = spark_config.get(
        "spark.kubernetes.executor.annotation.paasta.yelp.com/instance"
    )
    if paasta_service_non_truncated and paasta_instance_non_truncated:
        annotations.update(
            {
                "paasta.yelp.com/service": paasta_service_non_truncated,
                "paasta.yelp.com/instance": paasta_instance_non_truncated,
            }
        )
```
i'd maybe do something like this, but the existing code works too (mostly personal preference around not using "" over None and preferring dict.update())

paasta_tools/tron_tools.py
```diff
@@ -358,9 +358,11 @@ def build_spark_config(self) -> Dict[str, str]:
...
             paasta_instance=self.get_instance(),
             docker_img=f"{self.get_docker_registry()}/$PAASTA_DOCKER_IMAGE",
-            extra_volumes=self.get_volumes(
-                system_paasta_config.get_volumes(),
-                uses_bulkdata_default=system_paasta_config.get_uses_bulkdata_default(),
+            extra_volumes=spark_tools.docker_volumes_to_mappings(
+                self.get_volumes(
+                    system_paasta_config.get_volumes(),
+                    uses_bulkdata_default=system_paasta_config.get_uses_bulkdata_default(),
+                ),
```
nemacysts, line 365: was this change meant to be included in this diff?

chi-yelp: yeah it's because we updated service_configuration_lib (https://github.com/Yelp/service_configuration_lib/pull/152) and the change is for making the typing matches. (didn't use direct cast since it's more risky)

chi-yelp: Changed to direct casting. Do you mean to import the paasta repo later as a dependency of s_c_l? I checked that other repos which depend on s_c_l does not have the dependency to this repo (makes them more lightweight)

chi-yelp: Changed to direct casting. Do you mean to import the paasta repo later as a dependency of s_c_l? I checked that other repos which uses s_c_l do not have dependency to the paasta repo (makes them more lightweight)

nemacysts (approved):

PR #4011, opened 2025-02-26
Use C-accelerated safe_load for YAML

nemacysts:
PaaSTA is really just YAML all the way down, so using the pure Python
functions here leads to a considerable slowdown - this change alone
shaves ~20s off of setup_tron_namespace in one of our non-prod clusters

e.g., the below are from a "noop" (i.e., no tron config changes to
deploy) run of setup_tron_namespace:
status quo: https://fluffy.yelpcorp.com/i/2tSjKC6btSV1RR8t3Z9FdBJrQrhRBPTR.svg
this change: https://fluffy.yelpcorp.com/i/lMfLdcX1ZldCG8X1SNGQTB2Sg51fvldP.svg

(h/t to @Rob-Johnson for getting me to go down this rabbit-hole with his recent meshd change)


jfongatyelp (dismissed):

cuza (approved):

paasta_tools/paasta_yaml.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+def safe_load_yaml(stream):
+    """
+    yaml.safe_load() equivalent that will use a CSafeLoader if available.
+
+    This provides a significant speedup, but we allow falling back to the pure Python
+    codepath as we'd rather have things be slower than crash outright.
+    """
+    return yaml.load(stream, Loader=Loader)
```
EvanKrall, line 16: I kinda think we might just want to know immediately if CSafeLoader isn't available, rather than have silent performance degradations. We control what's in our own dh-virtualenv so we should be safe to assume that CSafeLoader is present. I guess technically people can import paasta-tools as a library, so that's not strictly true...

What if we do something jank here: check whether we're running from the venv in /opt/venvs/paasta-tools, and if so, assume CSafeloader is present. Otherwise, fall back to SafeLoader if CSafeLoader isn't present.

```
import sys

import yaml

if "paasta" in sys.prefix:  # try and catch both /opt/venvs/paasta-tools and ~/pg/paasta/.tox/py38-linux`
    from yaml import CSafeLoader as Loader
else:
   try:
        from yaml import CSafeLoader as Loader
    except ImportError:  # pragma: no cover
        from yaml import SafeLoader as Loader  # type: ignore
...
```

nemacysts: can do! I'll also refactor this a tad - i went to apply a similar change in Tron, but it looks like we're already doing this there (https://github.com/Yelp/Tron/blob/master/tron/yaml.py) and in a less obnoxiously named way

ilkinmammadzada (dismissed):

paasta_tools/paasta_yaml.py
nemacysts: (done)

nemacysts: hmm, actually - looks like the itests are failing after my refactor ;_;

nemacysts: i guess we can't be clever (and @jfongatyelp mentioned she'd run into the same issues in tron, so I was wrong with this being a solved problem) - i'll go back to this version and address your feedback there :p

nemacysts: okiedokes, this should actually be fine now

EvanKrall (approved):

PR #4014, opened 2025-03-03
move auth tween below pyramid_swagger

piax93:
The tween tries to access `swagger_data`, which won't be available before the swagger tween validated and loaded the request body, so it's currently failing.

I was undecided on whether to go this route, or trying to extract the "service" field from get/post data manually, but at the end I settled on moving the tween down the stack. This implementation does not include that many tweens, so I don't think not having auth right on top is going to be that impactful / dangerous.


nemacysts (approved):
i assume that we don't have a single place where we can control the tween ordering to add some comments about how we expect the tweens to be ordered

piax93:
> i assume that we don't have a single place where we can control the tween ordering to add some comments about how we expect the tweens to be ordered

@nemacysts each tween handles its positioning by itself, so not really.

PR #4015, opened 2025-03-03
Update pre-commit configuration to use Ruff for linting and formatting

cuza:
- Replaced pre-commit hooks `flake8`, `black`, `pyupgrade`, etc. with `ruff` for linting and formatting.


Rob-Johnson:
@cuza can I ask why? we use black for everything else at Yelp

cuza:
> @cuza can I ask why? we use black for everything else at Yelp.

@Rob-Johnson Great question, `ruff` is black/flake8 compatible, it's a lot faster than black/flake8, (like more than 10 times faster) and consolidate every python linter that we use for pre-commits hooks into one single tool.

lfdominguez:
> @cuza can I ask why? we use black for everything else at Yelp

`ruff` is converting into the new "standard" because include flake and black and additional rules. Far outperforms other linters, this give to you minor deploy time, so your DORA metrics because reduce the wait for linter step. Can check the format of the code too to maintain a well formated code, something really important in code bases like python. From my PoV moving to `ruff` is a win-win approach, great to see that companies so big like Yelp has people focus on this changes for their code bases.

general_itests/steps/validate_steps.py
```diff
@@ -28,10 +28,9 @@ def given_service(context, service_type):
 
 @when("we run paasta validate")
 def run_paasta_validate(context):
-    validate_cmd = (
-        "paasta validate "
-        "--yelpsoa-config-root %s "
-        "--service %s " % (context.soa_dir, context.service)
+    validate_cmd = "paasta validate --yelpsoa-config-root %s --service %s " % (
+        context.soa_dir,
+        context.service,
```
nemacysts: !8ball f-string while we're here

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -680,7 +678,7 @@ def get_spark_env(
             )
             sys.exit(1)
         spark_env["SPARK_HISTORY_OPTS"] = (
-            f"-D{args.spark_args} " f"-Dspark.history.ui.port={ui_port}"
```
nemacysts, line 683: lol

tests/test_generate_deployments_for_service.py
```diff
@@ -209,12 +213,15 @@ def test_get_deployments_dict():
...
-    }
+    assert (
+        generate_deployments_for_service.get_deployments_dict_from_deploy_group_mappings(
+            branch_mappings, v2_mappings
+        )
+        == {
+            "v1": branch_mappings,
+            "v2": mock.sentinel.v2_mappings,
+        }
+    )
```
nemacysts, line 224: i normally prefer the ruff formatting, but this is kinda lulzy :p

tests/kubernetes/bin/test_paasta_cleanup_stale_nodes.py
```diff
@@ -112,9 +112,7 @@ def test_main():
     ) as mock_terminate_nodes, mock.patch(
         "paasta_tools.kubernetes.bin.paasta_cleanup_stale_nodes.parse_args",
         autospec=True,
-    ) as mock_parse_args, mock.patch(
-        "boto3.client", autospec=True
-    ):
+    ) as mock_parse_args, mock.patch("boto3.client", autospec=True):
```
nemacysts, line 115: rip, the previous formatting makes for nicer diffs / i find it easier to read :p

paasta_tools/prune_completed_pods.py
```diff
@@ -150,8 +150,7 @@ def main():
                 )
         elif (
             # this is currently optional
-            allowed_pending_minues is not None
-            and pod.status.phase == "Pending"
+            allowed_pending_minues is not None and pod.status.phase == "Pending"
```
nemacysts, line 153: the previous formatting is probably easier to read, but it's not the end of the world

paasta_tools/cli/utils.py
```diff
@@ -184,7 +183,7 @@ class PaastaCheckMessages:
     DOCKERFILE_FOUND = success("Found Dockerfile")
 
     DOCKERFILE_MISSING = failure(
-        "Dockerfile not found. Create a Dockerfile and try again.\n  " "More info:",
+        "Dockerfile not found. Create a Dockerfile and try again.\n  More info:",
```
nemacysts: i think in both the old and new formatting it'd maybe be nicer to wrap these in parens and have them on separate lines, but that's maybe just me

nemacysts (dismissed):
seems like not too much churn now - i have some minor complaints, but i don't think they're blocking

tests/test_generate_deployments_for_service.py
cuza: I also suspect that black will produce the same output cuz iirc we were only blacking the diffs/new files not the whole project

paasta_tools/cli/utils.py
cuza: sadly ruff concats adjacent strings, I've tried a few options on how to make this and the only way to make it two lines is wrapping it on `# fmt: off .... # fmt: on` blocks :(

paasta_tools/prune_completed_pods.py
cuza: I don't think is the of the world but if you disagree I have no problem in wrapping it on `# fmt: off .... # fmt: on` blocks

paasta_tools/cli/utils.py
nemacysts: even if you wrap the strings in another layer of parens?

nemacysts: i guess it's not the end of the world, but it is kinda funny that both black and ruff like to do things that hurt readability :p

paasta_tools/prune_completed_pods.py
nemacysts: nah, this is fine - i prefer this over fmt off :p

nemacysts (approved):

PR #4016, opened 2025-03-04
Add support for usetting pod identity with spark-run

Qmando:
As described in an internal tech spec document. Support for using pod identity for manual spark-runs.

Pairs with https://github.com/Yelp/service_configuration_lib/pull/155

1. Check yelpsoa-configs for IAM roles assigned to a particular service when running spark-run. Use those to create a service account, and pass that to the spark conf builder
2. Allow you to override this, but it must be a valid role for at least one instance of the service
3. When this is used, don't send IAM credentials to the spark conf builder. They are still passed to `configure_and_run_docker_container`, where they will be used for the driver
4. If no aws role/profile/user is specified, use a default one (mrjob as placeholder. Subject to change)

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -12,6 +12,7 @@ from typing import Dict
 from typing import List
 from typing import Mapping
 from typing import Optional
+from typing import Set
 from typing import Tuple
 from typing import Union
 
@@ -31,6 +32,7 @@ from paasta_tools.cli.utils import get_service_auth_token
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import list_instances
 from paasta_tools.clusterman import get_clusterman_metrics
+from paasta_tools.kubernetes_tools import get_service_account_name
 from paasta_tools.spark_tools import auto_add_timeout_for_spark_job
 from paasta_tools.spark_tools import create_spark_config_str
 from paasta_tools.spark_tools import DEFAULT_SPARK_RUNTIME_TIMEOUT
@@ -38,6 +40,7 @@ from paasta_tools.spark_tools import DEFAULT_SPARK_SERVICE
 from paasta_tools.spark_tools import get_volumes_from_spark_k8s_configs
 from paasta_tools.spark_tools import get_webui_url
 from paasta_tools.spark_tools import inject_spark_conf_str
+from paasta_tools.tron_tools import load_tron_instance_configs
 from paasta_tools.utils import _run
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import filter_templates_from_config
@@ -419,6 +422,17 @@ def add_subparser(subparsers):
         default=False,
     )
 
+    aws_group.add_argument(
+        "--force-pod-identity",
+        help=(
+            "Normally the spark executor will use the pod identity defined "
+            "for the relevant instance in yelpsoa-configs. If the instance "
+            "isn't setup there yet, you can override the IAM role arn here."
+            " However, it must already be set for a different instance of "
+            "the service."
+        ),
+    )
+
     jupyter_group = list_parser.add_argument_group(
         title="Jupyter kernel culling options",
         description="Idle kernels will be culled by default. Idle "
@@ -625,6 +639,19 @@ def get_spark_env(
     spark_env["KUBECONFIG"] = system_paasta_config.get_spark_kubeconfig()
 
     return spark_env
+
+
+def get_all_iam_roles_for_service(
+    service: str,
+    cluster: str,
+) -> Set[str]:
+    tron_instance_configs = load_tron_instance_configs(service, cluster)
+    roles = set()
+    for action in tron_instance_configs:
+        role = action.get_iam_role()
+        if role:
+            roles.add(role)
+    return roles
 
 
 def _parse_user_spark_args(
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -627,6 +642,19 @@ def get_spark_env(
...
+    service: str,
+    cluster: str,
+) -> Set[str]:
+    tron_instance_configs = load_tron_instance_configs(service, cluster)
+    roles = set()
+    for action in tron_instance_configs:
+        role = action.get_iam_role()
+        if role:
+            roles.add(role)
+    return roles
```
nemacysts, line 655: could do a set comprehension like
```Python
    roles = {action.get_iam_role() for action in tron_instance_configs if action.get_iam_role()}
```

but this is good too :)

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1205,6 +1233,39 @@ def paasta_spark_run(args: argparse.Namespace) -> int:
...
+            service_account_name = get_service_account_name(args.force_pod_identity)
+        else:
+            service_account_name = get_service_account_name(iam_role)
+        if (
+            not args.aws_credentials_yaml
+            and not args.aws_profile
+            and not args.assume_aws_role
+        ):
+            args.aws_credentials_yaml = (
+                "/etc/boto_cfg/mrjob.yaml"  # Temporary until we get a better user
```
nemacysts, line 1265: might be worth adding this to SystemPaastaConfig so that this can be changed without a paasta release

tests/cli/test_cmds_spark_run.py
```diff
@@ -1644,3 +1655,179 @@ def test_get_aws_credentials_default_profile(mock_boto3_session, mock_use_aws_pr
...
+@mock.patch.object(spark_run, "get_aws_credentials", autospec=True)
+@mock.patch.object(spark_run, "get_docker_image", autospec=True)
+@mock.patch.object(spark_run, "get_spark_app_name", autospec=True)
+@mock.patch.object(spark_run, "auto_add_timeout_for_spark_job", autospec=True)
+@mock.patch.object(spark_run, "_parse_user_spark_args", autospec=True)
+@mock.patch(
+    "paasta_tools.cli.cmds.spark_run.spark_config.SparkConfBuilder", autospec=True
+)
+@mock.patch.object(spark_run, "configure_and_run_docker_container", autospec=True)
+@mock.patch.object(spark_run, "get_smart_paasta_instance_name", autospec=True)
```
nemacysts, line 1674: i usually prefer to use the context-manager form of mocks to avoid a large stack of decorators like this, but i think the decorator form is probably consistent with what already exists in this file

tests/cli/test_cmds_spark_run.py
```diff
@@ -1644,3 +1655,179 @@ def test_get_aws_credentials_default_profile(mock_boto3_session, mock_use_aws_pr
...
+        build=True,
+        image=None,
+        enable_compact_bin_packing=False,
+        disable_compact_bin_packing=False,
+        service="test-service",
+        instance="test-instance",
+        cluster="test-cluster",
+        pool="test-pool",
+        yelpsoa_config_root=custom_yelpsoa or spark_run.DEFAULT_SOA_DIR,
+        aws_credentials_yaml="/path/to/creds" if aws_creds_provided else None,
```
nemacysts, line 1730: i usually try to avoid branches in tests to keep them simple, but this (and the others) are probably fine

paasta_tools/cli/cmds/spark_run.py
nemacysts, line 1265:
```suggestion
    if iam_role or args.force_pod_identity:
        if args.force_pod_identity:
            if args.yelpsoa_config_root != DEFAULT_SOA_DIR:
                print(
                    "--force-pod-identity cannot be used with --yelpsoa-config-root",
                    file=sys.stderr,
                )
                return 1
                
            allowed_iam_roles = get_all_iam_roles_for_service(
                args.service, args.cluster
            )
            if args.force_pod_identity not in allowed_iam_roles:
                print(
                    f"{args.force_pod_identity} is not an allowed role for this service. "
                    f"Allowed roles are: {allowed_iam_roles}.",
                    file=sys.stderr,
                )
                return 1
                
            service_account_name = get_service_account_name(args.force_pod_identity)
        else:
            service_account_name = get_service_account_name(iam_role)
        if (
            not args.aws_credentials_yaml
            and not args.aws_profile
            and not args.assume_aws_role
        ):
            args.aws_credentials_yaml = (
                "/etc/boto_cfg/mrjob.yaml"  # Temporary until we get a better user
```

adding a little bit of whitespace in this section might be nice, but i also don't necessarily think it's necessary

nemacysts (dismissed):

nemacysts:
whoops, meant to add this note to my review: 

@Qmando: probably wanna wait for someone on ml-compute to ship this as well since they own this, but lgtm :)

paasta_tools/cli/cmds/spark_run.py
nurdann: Think it's not much different than getting config with a default value, so you are still specifying some user.

nurdann (approved):

chi-yelp:
> 1. Check yelpsoa-configs for IAM roles assigned to a particular service when running spark-run. Use those to create a service account, and pass that to the spark conf builder

Just to confirm - we're only referencing service accounts from the SOA configs, not creating them right?

chi-yelp:
> 1. Check yelpsoa-configs for IAM roles assigned to a particular service when running spark-run. Use those to create a service account, and pass that to the spark conf builder

Just to confirm - we're only getting the service accounts names here for executor pod identity, not creating them right?

chi-yelp (dismissed):

nurdann:
> > 1. Check yelpsoa-configs for IAM roles assigned to a particular service when running spark-run. Use those to create a service account, and pass that to the spark conf builder
> 
> Just to confirm - we're only getting the service accounts names here for executor pod identity, not creating them right?

Yes, even currently you can only use SA that already must exists on the namespace. This further filters it so that only IAM roles specified in yelpsoa-configs can be used.

paasta_tools/cli/cmds/spark_run.py
Qmando: But the current code isn't a default at all, so you can't change it without a release. I am adding it to the config

tests/cli/test_cmds_spark_run.py
Qmando: It's a little funky but I wanted to avoid copying the entire test several times. 🤷

nemacysts (dismissed):

chi-yelp (dismissed):

Qmando:
@nemacysts @chi-yelp 
I've put all the change in behavior behind a new arg, `--executor-pod-identity`. I was hoping to avoid having to do this, but I realized there will be cases where this change could break existing jobs. See https://yelp.slack.com/archives/C08920ZF0Q2/p1743024040622489 for more details.

This change makes this PR much safer to deploy, but it also means we'll need to be more thorough with manual testing. I hope to deprecate it in the future when everything is setup correctly.

nemacysts (approved):

nurdann (approved):

PR #4017, opened 2025-03-04
Fix timeout regex for spark submit command

timma-yelp:
### Problem
We check if there is already a timeout command in the user's spark command before auto adding it, but the check doesn't work with the case that an CLI argument is used, e.g. `timeout -v spark-submit`

### Solution
Fix the regex which checks if timeout is already present

### Testing
I test ran the `spark-submit` locally on dev-box with `timeout -v 1m`, as expected it timed out at `1m` 
command and output logs can be found [here](https://fluffy.yelpcorp.com/i/wbjZ7FrpPNGVCrg9TgXGnJcd2Hc48rVX.html)

### DIFF
--- a/paasta_tools/spark_tools.py
+++ b/paasta_tools/spark_tools.py
@@ -184,6 +184,9 @@ def auto_add_timeout_for_spark_job(
     if "spark-submit" not in cmd:
         return cmd
     try:
+        # Not supporting `=` for now, to support `=` we need to do lot of error checking,
+        # regex becomes a bit complex, use cases are very less, same command can be
+        # achieved without `=` also
         options_regex = "((-[a-zA-Z]+|\\-\\-[a-zA-Z\\-]+)(\\s+[a-zA-Z0-9\\-]+)?\\s*)*"
         duration_regex = r"[\d]+[\.]?[\d]*[m|h]"
         timeout_present = re.match(
--- a/paasta_tools/spark_tools.py
+++ b/paasta_tools/spark_tools.py
@@ -184,8 +184,10 @@ def auto_add_timeout_for_spark_job(
     if "spark-submit" not in cmd:
         return cmd
     try:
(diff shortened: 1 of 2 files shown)


chi-yelp:
Could you add an unit test and also add test run details (i.e. local paasta spark-run) in the PR description?

Example local spark-run cmd:
```
.tox/py38-linux/bin/paasta spark-run --aws-profile=dev --cmd "<timeout...> spark-submit /code/integration_tests/s3.py"
```

timma-yelp:
> Could you add an unit test and also add test run details (i.e. local paasta spark-run) in the PR description?
> 
> Example local spark-run cmd:
> 
> ```
> .tox/py38-linux/bin/paasta spark-run --aws-profile=dev --cmd "<timeout...> spark-submit /code/integration_tests/s3.py"
> ```

I looked for existing unit tests to update them but didn't find any, is it ok to add new one?

will update the description with local spark-run command

chi-yelp:
> > Could you add an unit test and also add test run details (i.e. local paasta spark-run) in the PR description?
> > Example local spark-run cmd:
> > ```
> > .tox/py38-linux/bin/paasta spark-run --aws-profile=dev --cmd "<timeout...> spark-submit /code/integration_tests/s3.py"
> > ```
> 
> I looked for existing unit tests to update them but didn't find any, is it ok to add new one?
> 
> will update the description with local spark-run command

Sure, feel free to create one, thanks!

paasta_tools/spark_tools.py
```diff
@@ -184,8 +184,10 @@ def auto_add_timeout_for_spark_job(
     if "spark-submit" not in cmd:
         return cmd
     try:
+        options_regex = "((-[a-zA-Z]+|\\-\\-[a-zA-Z\\-]+)(\\s+[a-zA-Z0-9\\-]+)?\\s*)*"
```
chi-yelp, line 187:
```suggestion
        options_regex = r"(--*[a-zA-Z]+(\s+\w+|=\w+)?\s*)*"
```
to cover `--kill-after=1m`

chi-yelp (approved):

PR #4018, opened 2025-03-12
remove vestigial remote-run implementation

piax93:
With the aim of soon going to integrate bit by bit the POC code implemented in https://github.com/Yelp/paasta/pull/3986, I'd like to start from ripping off the legacy remote-run code which has been dead since we moved away from Mesos. It technically also seems to support running stuff ssh-ing to adhoc instances, but I have not found any place where this would be used, as currently we prefer the pattern "Tron ssh executor + paasta local-run / spark-run".

There would be so many more crufty Mesos thing to get rid of, but I limited myself to `log_task_lifecycle_events`, since that was using some of the methods provided by the remote-run module, but it's absolutely not referenced anywhere in our environment. It's not even linked by the paasta-tools debian package.


paasta_tools/log_task_lifecycle_events.py
nemacysts, line 1: ty - we started going down the road of deleting a bunch of mesos code in a past unhack, but we ran out of time to completely excise things and haven't come back around to finish :)

paasta_tools/api/views/instance.py
nemacysts, line 1: i look forward to the day we finish removing the mesos code :dream:

nemacysts (approved):

PR #4019, opened 2025-03-13
add support for k8s jobs

piax93:
One first portion of https://github.com/Yelp/paasta/pull/3986, with a couple of minor improvements, as I didn't want to bunch everything together as it'd be a pretty large change to review.

Next I'll do the API endpoints, and I'll finish with the client side bits.

### DIFF
--- a/paasta_tools/kubernetes/application/controller_wrappers.py
+++ b/paasta_tools/kubernetes/application/controller_wrappers.py
@@ -6,6 +6,7 @@ from typing import Union
 
 from kubernetes.client import V1DeleteOptions
 from kubernetes.client import V1Deployment
+from kubernetes.client import V1Job
 from kubernetes.client import V1PodDisruptionBudget
 from kubernetes.client import V1StatefulSet
 from kubernetes.client.rest import ApiException
@@ -13,6 +14,7 @@ from kubernetes.client.rest import ApiException
 from paasta_tools.autoscaling.autoscaling_service_lib import autoscaling_is_paused
 from paasta_tools.eks_tools import load_eks_service_config_no_cache
 from paasta_tools.kubernetes_tools import create_deployment
+from paasta_tools.kubernetes_tools import create_job
 from paasta_tools.kubernetes_tools import create_pod_disruption_budget
 from paasta_tools.kubernetes_tools import create_stateful_set
 from paasta_tools.kubernetes_tools import ensure_service_account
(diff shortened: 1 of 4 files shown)


paasta_tools/kubernetes/application/controller_wrappers.py
```diff
@@ -411,14 +413,63 @@ def update(self, kube_client: KubeClient):
         )
 
 
+class JobWrapper(Application):
+    def __init__(
+        self,
+        item: V1Job,
+        logging=logging.getLogger(__name__),
+    ):
+        item.spec.replicas = None
```
nemacysts, line 422: i assume this is here so that we don't get an AttributeError in `sanitize_for_config_hash()`?

might be worth commenting this line so that an enterprising yelper doesn't try to be "helpful" and clean this up in the future :)

paasta_tools/kubernetes_tools.py
```diff
@@ -2044,6 +2047,47 @@ def get_pod_management_policy(self) -> str:
         """Get sts pod_management_policy from config, default to 'OrderedReady'"""
         return self.config_dict.get("pod_management_policy", "OrderedReady")
 
+    def format_kubernetes_job(
+        self,
+        job_label: str,
+        deadline_seconds: int = 3600,
```
nemacysts, line 2053: just curious: we'd talked about limiting the max uptime for the sandbox jobs - given that this doesn't do any limiting, I assume that we'll have the thresholds set in the new remote-run command? or do we want to add a max uptime check/clamp here?

paasta_tools/kubernetes_tools.py
```diff
@@ -2154,7 +2198,10 @@ def has_routable_ip(
         return "false"
 
     def get_pod_template_spec(
-        self, git_sha: str, system_paasta_config: SystemPaastaConfig
+        self,
+        git_sha: str,
+        system_paasta_config: SystemPaastaConfig,
+        restart: bool = True,
```
nemacysts, line 2204: non-blocking suggestion: `restart_on_failure` might be a slightly more descriptive name

nemacysts (commented):
something about how setup_kubernetes_job will finally actually be able to setup a kubernetes job cracks me up :p 

(generally lgtm, just a couple questions - but I can ship afterwards :))

paasta_tools/kubernetes/application/controller_wrappers.py
piax93: good point

paasta_tools/kubernetes_tools.py
piax93: yes, I was planning to clamp the value in the API view

nemacysts (approved):

PR #4020, opened 2025-03-13
Speed up check_kubernetes_service_replication

nemacysts:
At some point this ran in <1min, and we have >1 things assuming this is still running in <1min...but the current version of this script actually takes multiple minutes in most of our clusters.

I grabbed a couple flamegraphs of the current version of this (which I've since lost in my scrollback/browser history, but I can re-gen pretty easily) and noticed a couple obvious things:
* we spend a significant amount of time getting pods - let's parallelize that! Note: we do this with multiprocessing and not multithreading since there's a lot of serialization happening of k8s objects that would likely hold onto the GIL
* we were spending an obscene amount of time in filter_pods_by_service_instance() since we were calling it over and over again - let's group pods once in a smarter way and pass the grouping around instead :)
* this one is kinda ???: socket.gethostbyaddr() was pretty prominent in the flamegraphs - we don't actually use the hostnames in this check, so let's add a way to skip that hostname resolution

I also deleted the --additional-namespaces code since it's entirely unused now and it made things slightly cleaner.

EDIT: I regenerated the flamegraphs:
* status quo (457.90 secs, while py-spy'd): https://fluffy.yelpcorp.com/i/wDQ30vFtkF0N8wcDT2d360Rnkx8P1TRN.svg
* this PR (100.66 secs, while py-spy'd): https://fluffy.yelpcorp.com/i/SCt1JRkMJ9x4SWVXL4MQg1t7CFhS5ZWl.svg


nemacysts:
grr, i forgot that make test doesn't run mypy for some reason

nemacysts:
huh, what are our tests doing - mypy is showing me that this should have definitely broken some of the other checkers (e.g., the flink one)

EvanKrall (approved):
I'd prefer `grouped_pods` to be something more descriptive like `pods_by_service_instance` or something

nemacysts:
I also tested the flink replication checker on infrastage after mypy alerted me that I had broken it :)

nemacysts:
> I'd prefer `grouped_pods` to be something more descriptive like `pods_by_service_instance` or something

@EvanKrall done!

jfongatyelp (approved):

PR #4022, opened 2025-03-17
remote-run api implementation

piax93:
This is the largest portion of https://github.com/Yelp/paasta/pull/3986, implementing the new API for remote-run (look at the second commit, the first is mostly codegen stuff).

I deviated a bit from the original design poc code, as I saw that was waiting for the remote-run pods to become available as part of the `remote_run/.../start` endpoint, and that's not a great idea as it would just keep one API worker busy doing nothing. So I modified that endpoint to return as soon as the job resource is created, and then added another one to allow the CLI client to poll for updates until the pod becomes available.

So in summary the logic is the following:
* remote-run/start
  * load all the necessary config for a service instance, create a job resource, and launch it
  * return name of the job
* remote-run/poll
  * look if a pod for a remote-run job is ready
* remote-run/token
  * get temporary credentials to exec into the remote-run pod
* remote-run/stop
  * explicitly end the remote-run job (which will eventually be anyway since we set a deadline)

I also refused to add more stuff to the `kubernetes_tools` module as that already has **4500+ lines in it**, which is too much for me not to get triggered about, so I grouped all the new methods needed in a new module under `paasta_tools.kubernetes`


paasta_tools/kubernetes/remote_run.py
```diff
@@ -0,0 +1,373 @@
...
+    """Create service account to exec into remote-run pod
+
+    :param KubeClient kube_client: Kubernetes client
+    :param str namespace: pod namespace
+    :param str pod_name: pod name
+    :param str user: user requiring credentials
+    """
+    pod_name_hash = hashlib.sha1(pod_name.encode("utf-8")).hexdigest()[:12]
+    service_account_name = f"remote-run-{user}-{pod_name_hash}"
+    service_accounts = get_all_service_accounts(kube_client, namespace=namespace)
```
nemacysts, line 291: this could probably also use a label selector to pare down the SAs looked at

paasta_tools/kubernetes/remote_run.py
```diff
@@ -0,0 +1,373 @@
...
+) -> str:
+    """Create service account to exec into remote-run pod
+
+    :param KubeClient kube_client: Kubernetes client
+    :param str namespace: pod namespace
+    :param str pod_name: pod name
+    :param str user: user requiring credentials
+    """
+    pod_name_hash = hashlib.sha1(pod_name.encode("utf-8")).hexdigest()[:12]
+    service_account_name = f"remote-run-{user}-{pod_name_hash}"
```
nemacysts, line 290: i don't think we're really in danger of hitting the length limit for SA names, but might be good to future-proof this by wrapping thie with our truncation function just in case luispwithasuperduperreallylonghowlongisthisihopeitsover250somethingcharacters is ever hired :p

paasta_tools/kubernetes/remote_run.py
```diff
@@ -0,0 +1,373 @@
...
+    role: str,
+) -> None:
+    """Bind service account to role
+
+    :param KubeClient kube_client: Kubernetes client
+    :param str namespace: service account namespace
+    :param str service_account: service account name
+    :param str role: role name
+    """
+    role_binding = V1RoleBinding(
```
nemacysts, line 353: should we add some labels to this rolebinding so that we can more easily clean up later?

tests/kubernetes/test_remote_run.py
nemacysts, line 1: i think in new code we usually try to use the context manager form of mocking, but i don't think we have any firm rules on that :p

nemacysts (commented):
i assume that the code to cleanup the dynamically created objects will come in a future PR, but i kinda wonder if it'd be worth adding a `paasta.yelp.com/delete_after` label (or something of the sort) to make that code easier to write later :)

piax93:
> i assume that the code to cleanup the dynamically created objects will come in a future PR, but i kinda wonder if it'd be worth adding a paasta.yelp.com/delete_after label (or something of the sort) to make that code easier to write later :)

Yes, I'm looking into the clean up logic separately.

paasta_tools/kubernetes/remote_run.py
```diff
@@ -0,0 +1,373 @@
...
+    job: V1Job,
+    user: str,
+) -> str:
+    """Format name for remote run job
+
+    :param V1Job job: job definition
+    :param str user: the user requesting the remote-run
+    :return: job name
+    """
+    return f"remote-run-{user}-{job.metadata.name}"
```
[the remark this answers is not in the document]
piax93: this is a VERY good point, I'll need to do some DNS name math

paasta_tools/kubernetes/remote_run.py
```diff
@@ -0,0 +1,373 @@
...
+    :param str user: the user requesting the remote-run sandbox
+    :param bool interactive: whether it is expected to access the remote-run job interactively
+    :param bool recreate: whether to recreate remote-run job if existing
+    :param int max_duration: maximum allowed duration for the remote-ruh job
+    :return: outcome of the operation, and resulting Kubernetes pod information
+    """
+    kube_client = KubeClient(config_file="/etc/kubernetes/admin.conf")
+
+    # Load the service deployment settings
+    deployment_config = agnostic_load_service_config(service, instance, cluster)
```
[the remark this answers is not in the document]
piax93: Ok, let's cut that bridge then

paasta_tools/kubernetes/remote_run.py
piax93: I love DNS

tests/kubernetes/test_remote_run.py
piax93: But decorators are soooo much nicer. Chained `with`s are hard to read and to edit.

paasta_tools/kubernetes/remote_run.py
```diff
@@ -0,0 +1,373 @@
...
+    :param str service: service name
+    :param str instance: service instance
+    :param str cluster: paasta cluster
+    :param str user: the user requesting the remote-run sandbox
+    :param bool interactive: whether it is expected to access the remote-run job interactively
+    :param bool recreate: whether to recreate remote-run job if existing
+    :param int max_duration: maximum allowed duration for the remote-ruh job
+    :return: outcome of the operation, and resulting Kubernetes pod information
+    """
+    kube_client = KubeClient(config_file="/etc/kubernetes/admin.conf")
```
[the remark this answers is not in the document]
nemacysts: this would block running this code locally as this kubeconfig only exists on the paasta control plane and not devboxes - if this is temporary (and we'll give the paasta-api user the correct permissions later) then this is fine, but if we want something permanent we should update the paasta-api permissions

Qmando (commented):
LGTM other than minor swagger spec updates

paasta_tools/api/api_docs/oapi.yaml
```diff
@@ -1141,6 +1184,150 @@ paths:
...
+        required: true
+        schema:
+          type: string
+      - description: Job name
+        in: query
+        name: job_name
+        schema:
+          type: string
+        required: true
+      responses:
```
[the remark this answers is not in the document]
piax93: 204 is just in the response body, the HTTP code is always 200

paasta_tools/api/api_docs/oapi.yaml
```diff
@@ -1141,6 +1184,150 @@ paths:
...
+        "200":
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RemoteRunOutcome'
+          description: Successfully started remote-run sandbox
+        "404":
+          description: Service instance not found
+        "500":
+          description: Failure
```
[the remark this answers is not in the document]
piax93: It's not really useful, the swagger client won't handle them gracefully anyways.

nemacysts (approved):

PR #3994, opened 2024-12-16
Add support for iam_role in service_shard_update

panbenson:
create a `--iam-role` option to provide permissions when using `service_shard_update`

PR #4023, opened 2025-03-18
update iam-role to iam-role-mapping

panbenson:
follow up for https://github.com/Yelp/paasta/pull/3994

iam roles typically differ between environments so using same `--iam-role` may not always be valid.

update this variable to `--iam-role-mapping` so we can pass in a mapping from environment to role to account for this (e.g. `{"dev": "some_role", "prod": "another_role"}`)

### DIFF
--- a/paasta_tools/contrib/service_shard_update.py
+++ b/paasta_tools/contrib/service_shard_update.py
@@ -1,4 +1,5 @@
 import argparse
+import json
 import logging
 import sys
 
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/service_shard_update.py
```diff
@@ -165,11 +166,11 @@ def parse_args():
...
     parser.add_argument(
-        "--iam-role",
-        help="IAM role to use for the shard",
+        "--iam-role-mapping",
+        help="JSON string of deploy_prefix to IAM role mapping to use for the shard",
         required=False,
         type=str,
-        dest="iam_role",
+        dest="iam_role_mapping",
     )
```
nemacysts, line 174: you'll probably want to have both args exist temporarily - otherwise, there'll be a brief period of time where this is broken since either --iam-role or --iam-role-mapping won't exist in the deployed paasta package

nemacysts:
@panbenson is this something y'all are still interested in merging?

panbenson:
@nemacysts thanks for checking in! this has been superseded by https://github.com/Yelp/paasta/pull/4026 so this is no longer needed, closing this PR now

PR #4024, opened 2025-03-19
cleanup for remote-run resources

piax93:
Clean up logic for the (meant to be) ephemeral resources which are created in remote-run invocations (https://github.com/Yelp/paasta/pull/4022).


paasta_tools/kubernetes/bin/paasta_cleanup_remote_run_resources.py
```diff
@@ -0,0 +1,83 @@
...
+            get_remote_run_service_accounts,
+        ),
+        (kube_client.rbac.delete_namespaced_role, get_remote_run_roles),
+        (kube_client.rbac.delete_namespaced_role_binding, get_remote_run_role_bindings),
+    )
+    for delete_func, list_func in cleanup_actions:
+        for entity in list_func(kube_client, namespace):
+            if (
+                not entity.metadata.name.startswith("remote-run-")
+                or entity.metadata.creation_timestamp > age_limit
```
nemacysts, line 54: imo, we probably want to tag the resources with when they should be expired by - otherwise, we have to essentially always run this with `--max-age $REMOTE_RUN_JOB_MAX_DURATION` lest we delete an in-use SA/role/rolebinding

piax93: We actually want to always delete these quickly. They should just exist for the user to be able to exec into the pod, once the shell session is established they can get removed.

nemacysts: i guess there's not really an easy way for folks to exec back into the remote-run sandbox if their connection drops for whatever reason, so it does seem fine for this to happen :p

nemacysts (approved):

PR #4026, opened 2025-03-20
update service_shard_update to deploy on environment

panbenson:
follow up for https://github.com/Yelp/paasta/pull/3994

we should allow for deploys for certain environments, especially when using `--iam-role` because roles are often environment specific.
for backwards compatibility, this deploys to all environments when `--environment` is not passed in as an option.

### DIFF
--- a/paasta_tools/contrib/service_shard_update.py
+++ b/paasta_tools/contrib/service_shard_update.py
@@ -171,6 +171,13 @@ def parse_args():
         type=str,
         dest="iam_role",
     )
+    parser.add_argument(
+        "--environment",
+        help="Environment to deploy in, defaults to all environments if not specified",
+        required=False,
+        choices=DEPLOY_MAPPINGS.keys(),
+        dest="environment",
+    )
     return parser.parse_args()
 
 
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/service_shard_update.py
```diff
@@ -201,11 +208,17 @@ def main(args):
         working_dir=args.local_dir or "/nail/tmp",
         do_clone=args.local_dir is None,
     )
+    deploy_environments = DEPLOY_MAPPINGS
+    if args.environment is not None:
+        deploy_environments = {
+            k: v for k, v in DEPLOY_MAPPINGS.items() if k == args.environment
+        }
```
nemacysts, line 215: isn't this essentially
```suggestion
    deploy_environments = {args.environment: DEPLOY_MAPPINGS[args.environment]} if args.environment else DEPLOY_MAPPINGS
```
?

nemacysts (approved):
the `deploy_environments` could probably be simplified a bit, but i won't nitpick too much since this script it owned by your team :p

paasta_tools/contrib/service_shard_update.py
panbenson: ooo nice, that looks great!

nemacysts (approved):

PR #4033, opened 2025-03-26
First wave of fixes for remote-run

piax93:
* Describe 403 code for remote-run apis
    * Since OPA can deny remote-run requests, it'd be nice to have the right code returned from the API server rather than 500 cause the code is not in the spec (this was actually raise by @Qmando in #4022 but I turned the suggestion down, mistakenly, sorry)
* Add more output to remote-run cli
    * ... because it's way too silent now, and pod don't come online instantaneusly
* Fix wrong kwarg in paasta_cleanup_remote_run_resources
    * This is just linters failing me (and too much mocking in unit tests)
* Drop unneeded pod options for remote-run jobs
    * We do not want this to become a way for people to deploy services in prod, so in general these pods should not have routable IPs (it'll be a different story for toolbox containers as those will run sshd).
    * In my recent testing the job pod was failing to becomes ready due to hacheck. I think that may be due to some zookeeper locking which interfered with the existing replicas of the service, but at any rate, we don't want that, or any other sidecar, as this is effectively meant to be "local-run, but in a pod"


paasta_tools/kubernetes_tools.py
```diff
@@ -2214,8 +2225,10 @@ def get_pod_template_spec(
         )
 
         hacheck_sidecar_volumes = system_paasta_config.get_hacheck_sidecar_volumes()
-        has_routable_ip = self.has_routable_ip(
-            service_namespace_config, system_paasta_config
+        has_routable_ip = (
+            "false"
+            if force_no_routable_ip
+            else self.has_routable_ip(service_namespace_config, system_paasta_config)
```
nemacysts, line 2231: one day we'll update things so that the default is no routable ip #dream

nemacysts (approved):

PR #4037, opened 2025-04-01
Load credentials from file

nurdann:
## Problem
We want to allow reading IAM credentials from a root protected file. So that there's a single point of entry otherwise many other roles can fetch EKS token.

## Changes
- [x] Read root protected file
- [x] Re-run `spark-run` as sudo when given `--get-eks-token-via-iam-user`
- [x] Mount `/etc/kubernetes` directory - will make switching between configs easier
- [x] New paasta config key

### Mounting `/etc/kubernetes`
Currently, only a single file is mounted https://github.com/Yelp/paasta/blob/2e706a1879cbd012ad18f8e3e9324ffa62633cc8/paasta_tools/utils.py#L2736-L2737

Changing that requires changes to `/etc/paasta/spark-run.json` loaded from https://github.com/Yelp/paasta/blob/2e706a1879cbd012ad18f8e3e9324ffa62633cc8/paasta_tools/utils.py#L93-L94

populated via Puppet https://sourcegraph.yelpcorp.com/sysgit/puppet@4c6d259bc8e5cf7c48e75358a010741933591191/-/blob/modules/paasta_tools/manifests/public_config.pp?L501-508

## Verification (Updated Apr 10)
- [x] can `spark-submit` using `/etc/kubernetes/spark.conf` - got expected PATH_NOT_FOUND since S3 object doesn't exist [[cmd](https://fluffy.yelpcorp.com/i/8lKS163j1QVWV5gn95JXrLp658RkdSkZ.html)]
- [x] `spark-submit` with `/etc/kubernetes/spark.conf` fails to fetch credentials if EC2 metadata service is disabled as expected [[cmd](https://fluffy.yelpcorp.com/i/NgwPz6fXlcN3LnkHc4PQsTqHbVrH7SLj.html)]
- [x] can fetch EKS token via loaded ENV vars - got PATH_NOT_FOUND as well [[cmd](https://fluffy.yelpcorp.com/i/mpW1z1g6kgP0bTRxTz8CdtPNjgMPvQTt.html)]
    Config changes to come from https://github.yelpcorp.com/sysgit/puppet/pull/14234

### DIFF
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -640,9 +640,8 @@ def get_spark_env(
         spark_env["GET_EKS_TOKEN_AWS_SECRET_ACCESS_KEY"] = config["default"][
             "aws_secret_access_key"
         ]
-        spark_env["KUBECONFIG"] = "/etc/kubernetes/spark2.conf"
-    else:
-        spark_env["KUBECONFIG"] = system_paasta_config.get_spark_kubeconfig()
+
+    spark_env["KUBECONFIG"] = system_paasta_config.get_spark_kubeconfig()
 
     return spark_env
 
@@ -819,9 +818,8 @@ def configure_and_run_docker_container(
     if pod_template_path:
         volumes.append(f"{pod_template_path}:{pod_template_path}:rw")
 
-    volumes.append(
-        f"{system_paasta_config.get_spark_kubeconfig()}:{system_paasta_config.get_spark_kubeconfig()}:ro"
-    )
+    kubeconfig_dir = os.path.dirname(system_paasta_config.get_spark_kubeconfig())
+    volumes.append(f"{kubeconfig_dir}:{kubeconfig_dir}:ro")
 
     environment = instance_config.get_env_dictionary()  # type: ignore
     spark_conf_str = create_spark_config_str(spark_conf, is_mrjob=args.mrjob)
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -5,7 +5,6 @@ import os
 import re
 import shlex
 import socket
-import subprocess
 import sys
 from configparser import ConfigParser
 from typing import Any
@@ -787,31 +786,6 @@ def _calculate_docker_cpu_limit(
     )
 
 
-def create_volume_and_create_spark_conf(
-    spark_conf_volume,
-    spark_conf_name,
-):
-    subprocess.run(
-        ["docker", "volume", "create", spark_conf_volume],
-        check=True,  # raises error if non-0 return code
-        capture_output=True,  # no command output is shown without error
-    )
-    output = subprocess.check_output(
-        ["docker", "volume", "inspect", spark_conf_volume],
-    )
-
-    # write to volume without a container
-    volume_mountpoint = json.loads(output)[0]["Mountpoint"]
-    with open(os.path.join(volume_mountpoint, spark_conf_name), "w") as f:
-        f.write(get_spark_kubeconfig_template())
-
-
-def get_spark_kubeconfig_template():
-    return """
-<placeholder>
-    """
-
-
 def configure_and_run_docker_container(
     args: argparse.Namespace,
     docker_img: str,
@@ -845,20 +819,9 @@ def configure_and_run_docker_container(
     if pod_template_path:
         volumes.append(f"{pod_template_path}:{pod_template_path}:rw")
 
-    if args.get_eks_token_via_iam_user:
-        spark_kubeconfig = system_paasta_config.get_spark_kubeconfig()
-        # can re-use volume because KUBECONFIG does not change across users
-        # or this could be UUID but will result in many volumes - should probably be fine since it'll have a small file
-        spark_conf_volume = "nalma-vol"
-        create_volume_and_create_spark_conf(
-            spark_conf_volume=spark_conf_volume,
-            spark_conf_name=os.path.basename(spark_kubeconfig),
-        )
-        volumes.append(f"{spark_conf_volume}:{os.path.dirname(spark_kubeconfig)}:ro")
-    else:
-        volumes.append(
-            f"{system_paasta_config.get_spark_kubeconfig()}:{system_paasta_config.get_spark_kubeconfig()}:ro"
-        )
+    volumes.append(
+        f"{system_paasta_config.get_spark_kubeconfig()}:{system_paasta_config.get_spark_kubeconfig()}:ro"
+    )
 
     environment = instance_config.get_env_dictionary()  # type: ignore
     spark_conf_str = create_spark_config_str(spark_conf, is_mrjob=args.mrjob)
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -655,6 +655,10 @@ def get_spark_env(
         spark_env["SPARK_NO_DAEMONIZE"] = "true"
 
     if args.get_eks_token_via_iam_user:
+        # config looks as follows:
+        # [default]
+        # aws_access_key_id = ...
+        # aws_secret_access_key = ...
         with open(
             "/nail/etc/spark_driver_k8s_role_assumer/spark_driver_k8s_role_assumer.ini"
         ) as f:
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -85,6 +85,14 @@ DEPRECATED_OPTS = {
 
 SPARK_COMMANDS = {"pyspark", "spark-submit"}
 
+# config looks as follows:
+# [default]
+# aws_access_key_id = ...
+# aws_secret_access_key = ...
+SPARK_DRIVER_IAM_USER = (
+    "/nail/etc/spark_driver_k8s_role_assumer/spark_driver_k8s_role_assumer.ini"
+)
+
 log = logging.getLogger(__name__)
 
 
@@ -655,13 +663,6 @@ def get_spark_env(
         spark_env["SPARK_NO_DAEMONIZE"] = "true"
 
     if args.get_eks_token_via_iam_user:
-        # config looks as follows:
-        # [default]
-        # aws_access_key_id = ...
-        # aws_secret_access_key = ...
-        SPARK_DRIVER_IAM_USER = (
-            "/nail/etc/spark_driver_k8s_role_assumer/spark_driver_k8s_role_assumer.ini"
-        )
         with open(SPARK_DRIVER_IAM_USER) as f:
             config = ConfigParser()
             config.read_file(f)
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -664,6 +664,8 @@ def get_spark_env(
         ) as f:
             config = ConfigParser()
             config.read_file(f)
+
+        # these ENV variables are consumed by spark's KUBECONFIG command
         spark_env["GET_EKS_TOKEN_AWS_ACCESS_KEY_ID"] = config["default"][
             "aws_access_key_id"
         ]
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -659,9 +659,10 @@ def get_spark_env(
         # [default]
         # aws_access_key_id = ...
         # aws_secret_access_key = ...
-        with open(
+        SPARK_DRIVER_IAM_USER = (
             "/nail/etc/spark_driver_k8s_role_assumer/spark_driver_k8s_role_assumer.ini"
-        ) as f:
+        )
+        with open(SPARK_DRIVER_IAM_USER) as f:
             config = ConfigParser()
             config.read_file(f)
 
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -819,7 +819,9 @@ def configure_and_run_docker_container(
     if pod_template_path:
         volumes.append(f"{pod_template_path}:{pod_template_path}:rw")
 
-    volumes.append("/etc/kubernetes:/etc/kubernetes:ro")
+    volumes.append(
+        f"{system_paasta_config.get_spark_kubeconfig()}:{system_paasta_config.get_spark_kubeconfig()}:ro"
+    )
 
     environment = instance_config.get_env_dictionary()  # type: ignore
     spark_conf_str = create_spark_config_str(spark_conf, is_mrjob=args.mrjob)
--- a/paasta_tools/cli/cmds/spark_run.py
+++ b/paasta_tools/cli/cmds/spark_run.py
@@ -6,6 +6,7 @@ import re
 import shlex
 import socket
 import sys
+from configparser import ConfigParser
 from typing import Any
 from typing import cast
 from typing import Dict
@@ -150,6 +151,11 @@ def add_subparser(subparsers):
         "--force-use-eks",
         help=argparse.SUPPRESS,
         action=DeprecatedAction,
+    )
+    list_parser.add_argument(
+        "--get-eks-token-via-iam-user",
+        help="Use IAM user to get EKS token for long running spark-run jobs",
+        action="store_true",
     )
 
     group = list_parser.add_mutually_exclusive_group()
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1148,6 +1166,12 @@ def update_args_from_tronfig(args: argparse.Namespace) -> Optional[Dict[str, str
 
 
 def paasta_spark_run(args: argparse.Namespace) -> int:
+    if args.get_eks_token_via_iam_user and os.getuid() != 0:
+        print("Re-run proccess as root", file=sys.stderr)
```
Qmando, line 1170: Nit: update text to match other places in paasta, ie "Re-executing paasta spark-run with sudo.."

Qmando (changes requested):
Looks pretty good overall.

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -648,6 +654,19 @@ def get_spark_env(
...
+            config = ConfigParser()
+            config.read_file(f)
+        spark_env["GET_EKS_TOKEN_AWS_ACCESS_KEY_ID"] = config["default"][
+            "aws_access_key_id"
+        ]
+        spark_env["GET_EKS_TOKEN_AWS_SECRET_ACCESS_KEY"] = config["default"][
+            "aws_secret_access_key"
+        ]
+
     spark_env["KUBECONFIG"] = system_paasta_config.get_spark_kubeconfig()
```
chi-yelp, line 670: Shouldn't we use `spark2.conf` instead of the default `spark.conf` for `--get-eks-token-via-iam-user`?

nurdann: For now, you need to `export KUBECONFIG=...` to change it

chi-yelp (approved):

paasta_tools/cli/cmds/spark_run.py
nurdann: That's actually good point! Otherwise it makes it annoying to tweak command line

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -648,7 +654,22 @@ def get_spark_env(
         spark_env["SPARK_DAEMON_CLASSPATH"] = "/opt/spark/extra_jars/*"
         spark_env["SPARK_NO_DAEMONIZE"] = "true"
 
-    spark_env["KUBECONFIG"] = system_paasta_config.get_spark_kubeconfig()
+    if args.get_eks_token_via_iam_user:
+        with open(
+            "/nail/etc/spark_driver_k8s_role_assumer/spark_driver_k8s_role_assumer.ini"
```
nemacysts, line 659: probably worth making this configurable in SystemPaastaConfig

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -648,7 +654,22 @@ def get_spark_env(
...
+        ) as f:
+            config = ConfigParser()
+            config.read_file(f)
+        spark_env["GET_EKS_TOKEN_AWS_ACCESS_KEY_ID"] = config["default"][
```
nemacysts, line 663: this could probably use a comment re: what will consume these env vars

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -648,7 +654,22 @@ def get_spark_env(
...
+            "aws_access_key_id"
+        ]
+        spark_env["GET_EKS_TOKEN_AWS_SECRET_ACCESS_KEY"] = config["default"][
+            "aws_secret_access_key"
+        ]
+
+        spark_env["KUBECONFIG"] = system_paasta_config.get_spark2_kubeconfig()
```
nemacysts, line 670: how long do we anticipate the migration to this taking? i don't really love the spark2 naming filewise, but it'd be nice to avoid naming functions with spark2 :p

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -838,9 +859,8 @@ def configure_and_run_docker_container(
     if pod_template_path:
         volumes.append(f"{pod_template_path}:{pod_template_path}:rw")
 
-    volumes.append(
-        f"{system_paasta_config.get_spark_kubeconfig()}:{system_paasta_config.get_spark_kubeconfig()}:ro"
-    )
+    kubeconfig_dir = os.path.dirname(system_paasta_config.get_spark_kubeconfig())
+    volumes.append(f"{kubeconfig_dir}:{kubeconfig_dir}:ro")
```
nemacysts, line 863: seems like maybe a bit of a footgun to rely on the OG filepath for both cases? they both use the same directory for now so it's probably fine (and we're likely not going to change that, but y'know :p)

nemacysts: also, this could use a comment as to why we're mounting the directory (i.e., because the new kubeconfig relies on a helper script that will exist here)

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -1148,6 +1168,12 @@ def update_args_from_tronfig(args: argparse.Namespace) -> Optional[Dict[str, str
 
 
 def paasta_spark_run(args: argparse.Namespace) -> int:
+    if args.get_eks_token_via_iam_user and os.getuid() != 0:
+        print("Re-executing paasta spark-run with sudo..", file=sys.stderr)
+        # argv[0] is treated as command name, so prepending "sudo"
+        os.execvp("sudo", ["sudo"] + sys.argv)
```
nemacysts, line 1174: do we need the -H that we do in `paasta_local_run`?

paasta_tools/cli/cmds/spark_run.py
nemacysts: (and maybe documenting what the contents will look like?

paasta_tools/cli/cmds/spark_run.py
nurdann: Without separate config key, the cli-flag cannot switch to new kubeconfig unless new `KUBECONFIG` is exported. Once we no longer see file access to `spark.conf`, we can make the spark2 default flow

paasta_tools/cli/cmds/spark_run.py
nurdann: Home is inherited from running locally
```
$ sudo env | grep HOME
HOME=/nail/home/me
```

paasta_tools/cli/cmds/spark_run.py
nemacysts: when that's done, will we swap the names back to spark.conf /remove all the spark2 naming?

nemacysts: (i mostly don't want the `$FILENAME$N` naming to stick around for long since it's not particularly descriptive)

paasta_tools/cli/cmds/spark_run.py
nurdann: Don't think we are going to change the file. Once we can find a new way to launch drivers and we can entirely delete the file.

nemacysts: from a maintenance point-of-view, at the moment no one knows what this file looks like without having root on a box where it's deployed

having a hardcoded file will also make it harder to do things like potentially change the credentials used: but i guess if the plan is to get rid of this entirely then that's fine to skip. that said, this should probably be a named constant at the very least

paasta_tools/cli/cmds/spark_run.py
nemacysts: (-H changes that behavior :))

paasta_tools/cli/cmds/spark_run.py
nurdann: What do you mean by "constant"? Do you mean create a variable?

paasta_tools/cli/cmds/spark_run.py
nurdann: It doesn't matter for reading root owned file though

paasta_tools/cli/cmds/spark_run.py
nemacysts: (one other benefit of having this filename come from systempaastaconfig is that we're not leaking more internal filepaths beyond stuff that's documented already (e.g., `/nail/etc/{services,ecosystem,runtimeenv,etc}`)

paasta_tools/cli/cmds/spark_run.py
nurdann: Let me check how aws credentials are fetched

nemacysts: correct - we do this in local-run since that's needed to grab the creds to pull images from the registry. iiuc, `paasta spark-run` can also pull an image from the registry - will that still work when this re-execs in this way? 

i have a hunch that we have a similar `os.getuid() != 0` check in this file for pulling that might break since HOME will now be the calling user (i.e., a human) rather than the target user (root) and we'll already be uid=0

nurdann: User can still delete the file if the directory is user owned. But If we set home to root, then we cannot a user's profile

paasta_tools/cli/cmds/spark_run.py
nurdann: There's not much benefit to making systempaastaconfig because it's needs to be hard-coded in puppet anyway https://sourcegraph.yelpcorp.com/sysgit/puppet@d8983d66a129e40728717bf2eb6d345605c2028e/-/blob/modules/paasta_tools/manifests/public_config.pp?L501

nemacysts: correct: but one is hardcoded internally and in an easy-to-rollout location, and another is hardcoded publically in a repo that requires a longer deployment pipeline to release :)

paasta_tools/cli/cmds/spark_run.py
nurdann: So, the way docker command is constructed is different, no matter whether current uid is 0 or not, it will switch based on `DEFAULT_SPARK_DOCKER_REGISTRY` https://sourcegraph.yelpcorp.com/search?q=repo:%5EYelp/paasta%24+file:%5Epaasta_tools/cli/cmds/spark_run%5C.py%24+sudo&patternType=keyword&sm=0

paasta_tools/cli/cmds/spark_run.py
nurdann: If you think the effort is worth it, then I can pull the file name out of spark_run.py

nemacysts: probably fine as-is for now

that said, this named constant should now go at the top of this file alongside the other constants

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -648,7 +654,29 @@ def get_spark_env(
...
+        # aws_access_key_id = ...
+        # aws_secret_access_key = ...
+        SPARK_DRIVER_IAM_USER = (
+            "/nail/etc/spark_driver_k8s_role_assumer/spark_driver_k8s_role_assumer.ini"
+        )
+        with open(SPARK_DRIVER_IAM_USER) as f:
+            config = ConfigParser()
+            config.read_file(f)
+
+        # these ENV variables are consumed by spark's KUBECONFIG command
```
nemacysts, line 669:
```suggestion
        # these env variables are consumed by a script specified in the spark kubeconfig - and which will result in a tightly-scoped IAM identity being used for EKS cluster access
```

paasta_tools/cli/cmds/spark_run.py
nemacysts: and just for completeness: how long do we foresee this transition taking?

paasta_tools/cli/cmds/spark_run.py
```diff
@@ -838,9 +866,9 @@ def configure_and_run_docker_container(
     if pod_template_path:
         volumes.append(f"{pod_template_path}:{pod_template_path}:rw")
 
-    volumes.append(
-        f"{system_paasta_config.get_spark_kubeconfig()}:{system_paasta_config.get_spark_kubeconfig()}:ro"
-    )
+    # Mounting directory because there are multiple KUBECONFIGs due to migration
```
nemacysts, line 869: i'd argue that the more important bit is that there's a helper script we need included - otherwise we'd conditionally mount one file or the other :p
```suggestion
    # NOTE: we mount a directory here since the kubeconfig we're transitioning to requires a helper script that will co-exist in the same directory
```

nemacysts: (that said, as i mentioned in the other comment: it does seem somewhat brittle to depend on both kubeconfigs existing in the same directory - but probably fine if we're only going to have 2 kubeconfigs for a short period of time)

paasta_tools/cli/cmds/spark_run.py
nemacysts: hmm, we should probably try to make sure that we're not potentially leaving root-owned files around people's homedirs since that will otherwise generate onpoint load to get these cleaned up.

i'm assuming this is a problem because there's other parts of `spark-run` that will try to use a profile and create the file if not found?

tests/cli/test_cmds_spark_run.py
```diff
@@ -284,6 +288,40 @@ def test_get_spark_env(
...
+            "http://myrandomport",
+            SystemPaastaConfig(
+                SystemPaastaConfigDict(
+                    {"allowed_pools": {"test-cluster": ["test-pool", "fake-pool"]}}
+                ),
+                "fake_dir",
+            ),
+        )
+        iam_creds_file.assert_called_once_with(
+            "/nail/etc/spark_driver_k8s_role_assumer/spark_driver_k8s_role_assumer.ini"
```
nemacysts, line 317: probably want this to use the NAMED_CONSTANT created for this filepath

tests/cli/test_cmds_spark_run.py
```diff
@@ -284,6 +288,40 @@ def test_get_spark_env(
...
+    data = """
+[default]
+aws_access_key_id = <foo-id>
+aws_secret_access_key = <foo-secret>
+    """
+    with mock.patch(
+        "paasta_tools.cli.cmds.spark_run.open",
+        return_value=io.StringIO(data),
+        autospec=None,
+    ) as iam_creds_file:
```
nemacysts, line 301: i think we generally name mocks `mock_*` - so for consistency this is probably better off as `mock_open` or something similar

tests/cli/test_cmds_spark_run.py
```diff
@@ -1459,6 +1516,115 @@ def test_paasta_spark_run_uses_bulkdata(
...
+        uses_bulkdata=True,
+        get_eks_token_via_iam_user=True,
+    )
+    with mock.patch(
+        "paasta_tools.cli.cmds.spark_run.os.getuid",
+        return_value=0,
+        autospec=True,
+    ), mock.patch(
+        "paasta_tools.cli.cmds.spark_run.os.execvp", autospec=True
+    ) as exec_fn:
```
nemacysts, line 1623: (same as above re: mock naming)

paasta_tools/cli/cmds/spark_run.py
nurdann: Actually, I'm wrong. Then they would get error that profile doesn't exist
```
botocore.exceptions.ProfileNotFound: The config profile (devc) could not be found
```

paasta_tools/cli/cmds/spark_run.py
nurdann: Until we can get representative sample using the new kubeconfig, I want to be optimistic but I could see it staying in here until end of year.

chi-yelp (approved):

paasta_tools/cli/cmds/spark_run.py
chi-yelp: FYI - spark-run will [create](https://sourcegraph.yelpcorp.com/Yelp/service_configuration_lib@3e29dd2bb7d2c431902a6a60f8d5e1aed97f1b3f/-/blob/service_configuration_lib/spark_config.py?L1166-1168) a temporary pod template file under `/nail/tmp`, but I think it's fine to have a root-owned file under that path

Qmando (approved):

paasta_tools/cli/cmds/spark_run.py
nemacysts: oh my, if this is potentially sticking around that long we probably want some better naming: spark2 will be pretty meaningless to folks that aren't us several months into the future (unless they look through this PR/the tech spec/etc)

nurdann: Renamed variable

nemacysts (approved):

chi-yelp (approved):

PR #4039, opened 2025-04-07
improve cleanup of remote-run job resources

piax93:
`paasta remote-run` works fine now, but I noticed that unless people call `paasta remote-run stop` explicitly, the `job` resource would remain defined forever, even if the pod is terminated by the deadline being reached.

Tackling that on two different angles:
* Setting `ttl_seconds_after_finished` in the job spec should ensure that when the pod completes, the job gets cleaned up too by k8s itself.
* Adding jobs to the resources cleaned up by the cleanup script, but using the max duration as the age limit for that. Just in case k8s does not to the above correctly.

### DIFF
--- a/paasta_tools/kubernetes/bin/paasta_cleanup_remote_run_resources.py
+++ b/paasta_tools/kubernetes/bin/paasta_cleanup_remote_run_resources.py
@@ -47,7 +47,7 @@ def clean_namespace(
 
     :param KubeClient kube_client: kubernetes client
     :param str namepsace: kubernetes namespace
-    :param datetime creds_age_limit: expiration time for authentication resources
+    :param datetime auth_age_limit: expiration time for authentication resources
     :param datetime job_age_limit: expiration time for job resources
     :param bool dry_run: delete resources for real or not
     """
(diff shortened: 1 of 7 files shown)


paasta_tools/kubernetes/bin/paasta_cleanup_remote_run_resources.py
```diff
@@ -35,24 +37,44 @@
...
+    auth_age_limit: datetime,
+    job_age_limit: datetime,
+    dry_run: bool = False,
 ):
     """Clean ephemeral remote-run resource in a namespace
 
     :param KubeClient kube_client: kubernetes client
     :param str namepsace: kubernetes namespace
-    :param datetime age_limit: expiration time for resources
+    :param datetime creds_age_limit: expiration time for authentication resources
```
nemacysts, line 50:
```suggestion
    :param datetime auth_age_limit: expiration time for authentication resources
```

paasta_tools/kubernetes/bin/paasta_cleanup_remote_run_resources.py
```diff
@@ -96,9 +118,11 @@ def main():
     args = parse_args()
     logging.basicConfig(level=logging.DEBUG if args.verbose else logging.INFO)
     kube_client = KubeClient()
-    age_limit = datetime.now(tz=timezone.utc) - timedelta(seconds=args.max_age)
+    now = datetime.now(tz=timezone.utc)
+    age_limit = now - timedelta(seconds=args.max_age)
+    job_age_limit = now - timedelta(seconds=get_max_job_duration_limit())
     for namespace in get_all_managed_namespaces(kube_client):
-        clean_namespace(kube_client, namespace, age_limit, args.dry_run)
+        clean_namespace(kube_client, namespace, age_limit, job_age_limit, args.dry_run)
```
nemacysts, line 125: might be nice to use kwargs for the age limit params (at least) so that it's a little more self-documenting as to what `age_limit` is being used for

nemacysts (approved):

nemacysts (approved):

PR #4041, opened 2025-04-09
Added code to take spark driver pool from config

timma-yelp:
### Problem
Currently, the pool that the Spark driver of _driver on k8s tron jobs_ runs on is hardcoded to **_stable_** pool, to ensure the Spark driver not being interrupted due to spot interruptions.
 
However, in some cases, jobs need to pull images from the other docker registry which does not work in the stable pool. We need to make the driver pool configurable.

### Solution
Add the code to consider spark driver pool from config (if available) otherwise default to stable

### Testing
Unit tests

### DIFF
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -698,7 +698,7 @@ class TronActionConfig(InstanceConfig):
         which this function will read.
         """
         if self.get_executor() == "spark":
-            pool = load_system_paasta_config().get_spark_driver_default_pool_override()
+            pool = load_system_paasta_config().get_default_spark_driver_pool_override()
         else:
             pool = self.config_dict.get(
                 "pool", load_system_paasta_config().get_tron_default_pool_override()
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -697,13 +697,14 @@ class TronActionConfig(InstanceConfig):
         override this value. To control this, we have an optional config item that we'll puppet onto Tron masters
         which this function will read.
         """
-        return (
-            self.config_dict.get(
+        if self.get_executor() == "spark":
+            pool = load_system_paasta_config().get_spark_driver_default_pool_override()
+        else:
+            pool = self.config_dict.get(
                 "pool", load_system_paasta_config().get_tron_default_pool_override()
             )
-            if not self.get_executor() == "spark"
-            else spark_tools.SPARK_DRIVER_POOL
-        )
+
+        return pool
 
     def get_spark_executor_pool(self) -> str:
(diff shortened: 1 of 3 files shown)


paasta_tools/utils.py
```diff
@@ -2148,6 +2153,15 @@ def get_default_spark_iam_user(self) -> str:
             "default_spark_iam_user", "/etc/boto_cfg/mrjob.yaml"
         )
 
+    def get_spark_driver_default_pool_override(self) -> str:
+        """Get the spark driver's default pool override variable defined in this host's cluster config file.
```
nemacysts, line 2157:
```suggestion
        """
        If defined, fetches the override for what pool to run a Spark driver in. Otherwise, returns the default Spark driver pool.
```

I think we usually have the docstring start with a newline - also, default pool tends to mean something specific at yelp (i.e., a spot pool where disruption is allowed), so I tweaked the wording a bit to avoid that :)

paasta_tools/tron_tools.py
```diff
@@ -697,13 +697,14 @@ def get_pool(self) -> str:
         override this value. To control this, we have an optional config item that we'll puppet onto Tron masters
         which this function will read.
         """
-        return (
-            self.config_dict.get(
+        if self.get_executor() == "spark":
+            pool = load_system_paasta_config().get_spark_driver_default_pool_override()
```
nemacysts, line 701: it might be worth starting a discussion on how to allow for configuring the driver pool per-action since I assume eventually we'll want to also run GPU-using drivers through tron

paasta_tools/utils.py
```diff
@@ -2148,6 +2153,15 @@ def get_default_spark_iam_user(self) -> str:
...
+
+        :returns: The spark_driver_default_pool_override specified in the paasta configuration
+        """
+        return self.config_dict.get(
+            "spark_driver_default_pool_override", DEFAULT_SPARK_DRIVER_POOL
```
nemacysts, line 2162: i was racking my brain trying to find a good alternative name here to avoid `default_pool` appearing in the config option, but i couldn't really come up with anything better :p

other than maybe `spark_driver_pool_override` since this isn't otherwise currently configurable

timma-yelp: @nemacysts how about just `spark_driver_pool`?

nemacysts: works for me! i guess `default_spark_driver_pool_override` might also work and leave `spark_driver_pool` as an available unique option for a config value in the future to allow customizing the pool if so desired

nemacysts: (if we rename the option, we should also make sure to rename the function)

timma-yelp: all done

chi-yelp (approved):

nemacysts (approved):

PR #4043, opened 2025-04-09
COMPINFRA-4866: restrict alert_after values for tron/k8s

jfongatyelp:
This adds some extra validation for `alert_after` in tron and kubernetes/eks monitoring configs. 

Originally had hoped to create a monitoring.yaml schema + reuse definitions from within in each of these but we have some tooling that assumes each jsonschema here is self-contained so i'll need to detangle that before going that route.

For now, to prevent recurrences of these `1d` issues, this just adds individually to the separate schemas. 

new paasta validate would have caught this issue in both a tron-* and eks-* file: https://fluffy.yelpcorp.com/i/lXHsCt04BRWq9kBNjhKmq7QfBLcrfjZW.html

and am currently running validate against all services to ensure this doesn't catch anything else surprising that already exists. 
^ whoops, this caught that some folks are using a bare num of seconds for their alert_after, which both paasta + pysensu accept right now, so I've also adjusted the schema to accept either of these

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -1012,7 +1012,7 @@ Here is a list of options that PaaSTA will pass through:
 
  * ``alert_after``: Time string that represents how long a a check should be
    failing before an actual alert should be fired. Currently defaults to ``10m``
-   for the replication alert.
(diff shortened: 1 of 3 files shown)


nemacysts (approved):
nothing like a 0-code PR :p

nemacysts (approved):

PR #4044, opened 2025-04-10
publish a wheel

dimbleby:
This project currently publishes only a source distribution.

Publish both a source distribution and a wheel.


nemacysts (approved):
thanks! we don't actually use the pypi sdist (and once this is merged, the wheel) since we have internal builds for this - but i don't see why not :)

(and it's nice to move off of setup.py too)

PR #4045, opened 2025-04-11
remote-run for toolbox containers (alternative version)

piax93:
Functionally the same as https://github.com/Yelp/paasta/pull/4042, but using adhoc configs to provide base values for the deployment generation.


EvanKrall (dismissed):
I prefer the idea in this branch of keeping the container configuration in yelpsoa-configs, as long as that doesn't cause problems for you.

piax93:
All good to me. My strong opinions about using Puppet to manage configuration for this got weaker over time.

paasta_tools/cli/schemas/adhoc_schema.json
```diff
@@ -180,6 +183,15 @@
...
                     "type": "string"
+                },
+                "namespace": {
+                    "type": "string"
+                },
+                "container_port": {
+                    "type": "integer"
+                },
+                "docker_url": {
+                    "type": "string"
```
nemacysts, line 194: now that we've picked a direction: i'd been talking to @EvanKrall about this and he had a pretty nifty idea for getting multi-image support for a given service: we can use the DeploymentVersion code intended for no-commit-redeploys here since the various paasta deployment commands all take `--image-version $VERSION` as an arg - and we could have each toolkit container be in it's own deploy group and have a custom jenkins pipeline to build/tag/push the various images

that said: it's probably not too big a shift to go from docker_url to the image version approach, so i'm personally fine with trying the image version bits later if you'd like to start with this approach

paasta_tools/kubernetes/remote_run.py
```diff
@@ -141,53 +159,75 @@ def remote_run_start(
...
         return {"status": 404, "message": "No pod found"}
     if pod.status.phase == "Running":
-        return {
+        result: RemoteRunOutcome = {
             "status": 200,
             "message": "Pod ready",
             "pod_name": pod.metadata.name,
             "namespace": namespace,
         }
+        if is_toolbox:
```
nemacysts, line 198:
```suggestion
        # NOTE: we only do this for toolbox containers since we need to SSH in rather than exec in to maintain user identities
        # i.e., being `nobody` is fine in a normal remote-run, but in a toolkit, there's likely tooling that will require knowing the real user
        if is_toolbox:
```

nemacysts: (iirc, this is why the toolboxes are SSH-able - but please correct me if i'm wrong)

nemacysts: (mostly just wanna make sure no one tries to simplify things in the future down to always execing)

paasta_tools/kubernetes/remote_run.py
```diff
@@ -244,6 +284,45 @@ def remote_run_token(
...
+    :param str service: toolbox name
+    :param str cluster: target deployment cluster
+    :param str user: user requesting the toolbox
+    :return: deployment configuration
+    """
+    if not user.isalnum():
+        raise RemoteRunError(
+            f"Provided username contains non-alphanumeric characters: {user}"
+        )
+    adhoc_instance = service[len(TOOLBOX_MOCK_SERVICE) + 1 :]
```
nemacysts, line 301: i think i'm missing something here - what will `service` look like here? from the callsites of this function, it looks like we would have `instance` available to pass in here)

nemacysts: (i'm assuming that service as expected by this looks something like `prod-toolbox.instance` - but i'm also missing how that's getting setup)

paasta_tools/kubernetes/remote_run.py
```diff
@@ -244,6 +284,45 @@ def remote_run_token(
...
+    adhoc_deployment = load_adhoc_job_config(
+        TOOLBOX_MOCK_SERVICE,
+        adhoc_instance,
+        cluster,
+        load_deployments=False,
+    )
+    adhoc_deployment.config_dict.setdefault("extra_volumes", []).append(
```
nemacysts, line 308:
```suggestion
    # NOTE: we're explicitly dynamically mounting a single user's public keys as we want these pods to only be usable by said user
    adhoc_deployment.config_dict.setdefault("extra_volumes", []).append(
```

paasta_tools/kubernetes/remote_run.py
```diff
@@ -244,6 +284,45 @@ def remote_run_token(
...
+            "hostPath": f"/etc/authorized_keys.d/{user}.pub",
+            "mode": "RO",
+        },
+    )
+    adhoc_deployment.config_dict.setdefault("env", {})["SANDBOX_USER"] = user
+    adhoc_deployment.config_dict["routable_ip"] = True
+    return EksDeploymentConfig(
+        service=service,
+        cluster=cluster,
+        instance="main",
```
nemacysts, line 320: thoughts on having the `instance` here be the toolkit instance + the username? if i'm not mistaken, that would make it easier to filter/view who is using what with `kubectl get pods -L paasta.yelp.com/instance`

nemacysts: (unless that has other side-effects i'm not thinking of)

paasta_tools/utils.py
```diff
@@ -785,6 +787,8 @@ def get_docker_url(
         """Compose the docker url.
         :returns: '<registry_uri>/<docker_image>'
         """
+        if "docker_url" in self.config_dict:
```
nemacysts, line 790:
```suggestion
        # NOTE: we're explicitly only allowing this for adhoc instances to support remote-run toolkits - if you're looking at this to expand that support for non-remote-run cases: please chat with #paasta first
        if "docker_url" in self.config_dict:
```

paasta_tools/cli/schemas/adhoc_schema.json
nemacysts: (would also be nice since it would stop folks from being tempted to expand where else docker_url could be used :p)

nemacysts (commented):
just a couple questions + suggestions for comments for future us

paasta_tools/kubernetes/remote_run.py
piax93: The change I proposed in soa-configs shows that

paasta_tools/kubernetes/remote_run.py
piax93: There would be a bit much redundancy in the pod name. It'd basically container the user name twice. I think we can filter things decently enough already with the `job-name` label.

paasta_tools/kubernetes/remote_run.py
nemacysts: ah gotcha - i think this further confuses me then: shouldn't `service` here be something just like `prod-toolbox` then? or are we munging the service name+instance into a single string at some point and i'm just missing where we're doing that

afaict, this is called from remote_run_state which includes the instance name - but we're not passing that through to here

piax93: > are we munging the service name+instance into a single string

Users are specifying the munged name in full, we need that for authorization purposes. It's one of the reasons I had against using soa-configs vs system configurations, because I don't want to create N meta-services for N toolbox use cases.

paasta_tools/kubernetes/remote_run.py
```diff
@@ -244,6 +284,50 @@ def remote_run_token(
...
+    :param str user: user requesting the toolbox
+    :return: deployment configuration
+    """
+    if not user.isalnum():
+        raise RemoteRunError(
+            f"Provided username contains non-alphanumeric characters: {user}"
+        )
+    # NOTE: API authorization is enforced by service, and we want different rules
+    # for each toolbox, so clients send a combined service-instance string, and then
+    # we split it here to laod the correct instance settings.
```
nemacysts, line 303:
```suggestion
    # we split it here to load the correct instance settings.
```

nemacysts: (non-blocking tho :p)

nemacysts (approved):

PR #4046, opened 2025-04-14
Use 22.04 GHA runners + switch to deadsnakes

nemacysts:
GitHub is deprecating the 20.04 runners, but we still need to use py3.7, so we need to switch to grabbing Python from the deadsnakes PPA rather than using the GHA setup-pyhon action


nemacysts:
(let's see what the copilot review does :p)

PR #4047, opened 2025-04-14
Use 22.04 GHA runners (+ always use py3.8)

nemacysts:
GitHub is deprecating the 20.04 runners, and for some reason we're setting up py3.7 even though we're expecting 3.8.

Note: this required some packages bump: for some reason, `check-requirements` was reporting
that not all packages were installed -- even though the preceding tox loglines reported that
the "missing" packages were installed.

I tried to be pretty conservative with the bumps since we need to do a mass-upgrade at some
point in the future and it wasn't worth dealing with breaking changes and whatnot at the 
moment

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   tox:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


nemacysts:
oh, i vim'd too hard and didn't save my buffer with the 22.04 change 🤣

cuza (approved):

cuza:
oh wait this build is failing 😅

nemacysts:
> oh wait this build is failing 😅

@cuza yea, my latest commit (intended to speed up the deb build) doesn't seem to work for some reason - once i revert the build will go back to passing

PR #4048, opened 2025-04-15
Make the default pre-stop sleep delay for mesh-registered services configurable

EvanKrall:
This makes it possible for services to specify a longer (or shorter) time between `hadown` and pod termination, which allows us to gracefully drain services whose requests take longer than 30 seconds.


tests/test_kubernetes_tools.py
```diff
@@ -4486,7 +4486,7 @@ def test_warning_big_bounce_routable_pod():
             job_config.format_kubernetes_app().spec.template.metadata.labels[
                 "paasta.yelp.com/config_sha"
             ]
-            == "config45672f04"
+            == "config4ed19ece"
```
EvanKrall, line 4489: I configured adding a conditional to `get_termination_grace_period` to have it return None if there isn't an explicit override of `pre_stop_drain_seconds` in yelpsoa-configs, but decided I'd rather have a big bounce now than additional code complexity forever.

paasta_tools/kubernetes_tools.py
```diff
@@ -2614,9 +2625,22 @@ def sanitize_for_config_hash(
...
+        Defaults to a value long enough to allow the standard hadown draining to finish.
+        """
+
+        return self.get_lifecycle_dict().get(
+            "termination_grace_period_seconds",
+            (
+                self.get_hacheck_prestop_sleep_seconds() + 1
+                if service_namespace_config.is_in_smartstack()
+                else None
+            ),
```
EvanKrall, line 2643: I considered adding another conditional for whether `pre_stop_drain_seconds` is even specified in yelpsoa-configs (returning None if not) to avoid the big bounce for routable pods, but decided that would be too much complexity to avoid a small one-time pain.

I'm tempted to apply the same logic to the `is_in_smartstack` conditional, i.e. remove that conditional (and the extra argument) and just default to `self.get_hacheck_prestop_sleep_seconds() + 1` regardless. This would mean we'd change the default from `None` (which is really 30 seconds, since that's the default for k8s podSpec) to 32 seconds. Probably not a big deal in practice, and again just means we'd have a big bounce (of all pods, mesh-registered and not)

docs/source/yelpsoa_configs.rst
```diff
@@ -484,7 +484,15 @@ instance MAY have:
     * ``pre_stop_command``: The command to run in your container before stopping.  This could handle gracefully stopping or checkpointing your worker, for example.
       This can be a list of strings (command + arguments) or a single string (which gets turned into a single-element list by Paasta.)
 
-    * ``termination_grace_period_seconds``: the number of seconds to allow before forcibly killing your instance.  Note that the instance will be forcibly killed after this period, so your pre_stop_command should complete well within this time period!
+    * ``pre_stop_drain_seconds``: For services registered in the mesh, we apply a default ``pre_stop_command`` which starts draining the service from the mesh and waits this many seconds before stopping the container.
```
nemacysts, line 487: technically this will apply to everything, no?

(which i'd say is what we want anyway)

paasta_tools/kubernetes_tools.py
nemacysts: either way is fine imo

docs/source/yelpsoa_configs.rst
EvanKrall: Hmm. Do we actually want that for non-registered services? If we aren't triggering some sort of draining, I'm not sure there's any reason to delay the termination.

EvanKrall: I've updated so non-smartstack services get `sleep 0` by default (but if they specify `pre_stop_drain_seconds`, they'll get `sleep $that`).

This will cause a big bounce, but might speed up deployments for services that aren't in the mesh.

tox.ini
```diff
@@ -71,10 +71,11 @@ deps =
     --requirement={toxinidir}/requirements-dev.txt
     --requirement={toxinidir}/yelp_package/extra_requirements_yelp.txt
     --editable={toxinidir}
+whitelist_externals = bash
 commands =
     check-requirements -vv
     pre-commit install -f --install-hooks
-    pre-commit run --all-files
+    bash -c 'pre-commit run --all-files || (if [[ $HOSTNAME =~ dev.* ]]; then echo "Re-running pre-commit since this is a dev box"; pre-commit run --all-files; else false; fi)'
```
nemacysts, line 78: just curious: what is this for? is this to handle hooks that might want to re-run on the fixed output from the first run?

nemacysts (dismissed):

nemacysts:
NOTE: we'll probably wanna bundle this with https://github.com/Yelp/paasta/pull/4057 to not have two back-to-back big bounces

tox.ini
EvanKrall: Yeah, exactly. I'm tired of having to rerun it myself when e.g. black fixes things automatically. This will still fail if the hooks can't fix something.

paasta_tools/kubernetes_tools.py
```diff
@@ -1496,17 +1497,77 @@ def get_readiness_probe(
...
+        """The number of seconds to sleep between hadown and terminating the hacheck container. We want hacheck to be
+        up for slightly longer than the main container is, so we default to pre_stop_drain_seconds + 1.
+
+        It doesn't super matter if hacheck goes down before the main container -- if it's down, healthchecks will fail
+        and the service will be removed from smartstack, which is the same effect we get after running hadown.
+        """
+
+        # Everywhere this value is currently used (hacheck sidecar or gunicorn sidecar), we can pretty safely
+        # assume that the service is in smartstack.
+        return self.get_prestop_sleep_seconds(is_in_smartstack=True) + 1
```
EvanKrall, line 1520: honestly should probably just go back to the constant 31s. I'm pretty sure it doesn't matter if the hacheck container goes down -- we want the healthchecks to be failing anyway.

nemacysts (approved):

PR #4052, opened 2025-04-25
[rfc] add AWS app id to services env variables

piax93:
This is something that popped into Infrasec's conversations lately, as sometimes it's a bit hard to attribute specific AWS log events to service instances, [and since it's possible to inject identifiers into the User-Agent set by various AWS SDKs](https://docs.aws.amazon.com/sdkref/latest/guide/feature-appid.html), it'd be nice to get that by default from PaaSTA.

~I wouldn't want this to bump all services at the same time, nor to have to roll it out manually service-by-service with a feature flag, so I'd propose to temporarily exclude the env var form the config hash calculations, let some time pass so that services are redeployed naturally, and then clean up the exclusion. I care less about Tron deployments, since that'd be a single large update to job configs, but no disruption for stuff currently running.~

**RELEASING THIS WILL CAUSE ALL SERVICES TO BOUNCE** ([see conversation below](https://github.com/Yelp/paasta/pull/4052#discussion_r2060466965))

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1198,6 +1198,13 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             "git_sha", ""
         )
 
+        # Our workloads interact with AWS quite a lot, so it comes handy to
+        # propagate an "application ID" in the user-agent of API requests
+        # for debugging purposes (max length is 50 chars from AWS docs).
+        # TODO: temporarily setting this here rather than in `InstanceConfig`
+        # to only affect long running services.
+        env["AWS_SDK_UA_APP_ID"] = f"{self.service}.{self.instance}"[:50]
+
         # We drop PAASTA_CLUSTER here because it will be added via `get_kubernetes_environment()`
         env.pop("PAASTA_CLUSTER", None)
 
--- a/tests/test_kubernetes_tools.py
+++ b/tests/test_kubernetes_tools.py
@@ -794,6 +794,7 @@ class TestKubernetesDeploymentConfig:
             assert self.deployment.get_env() == {
                 "hello": "world",
                 "PAASTA_SOA_CONFIGS_SHA": "fake_soa_git_sha",
+                "AWS_SDK_UA_APP_ID": "kurupt.fm",
             }
 
     def test_get_container_env(self):
@@ -4440,7 +4441,7 @@ def test_warning_big_bounce_default_config():
             job_config.format_kubernetes_app().spec.template.metadata.labels[
                 "paasta.yelp.com/config_sha"
             ]
-            == "config1c7deb78"
+            == "config8031ca07"
         ), "If this fails, just change the constant in this test, but be aware that deploying this change will cause every service to bounce!"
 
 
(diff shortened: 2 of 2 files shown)


paasta_tools/utils.py
```diff
@@ -695,6 +695,10 @@ def get_env_dictionary(
         instance_type = self.get_instance_type()
         if instance_type:
             env["PAASTA_INSTANCE_TYPE"] = instance_type
+        # Our workloads interact with AWS quite a lot, so it comes handy to
+        # propagate an "application ID" in the user-agent of API requests
+        # for debugging purposes (max length is 50 chars from AWS docs).
+        env["AWS_SDK_UA_APP_ID"] = f"{self.service}--{self.instance}"[:50]
```
nemacysts, line 701: env var values don't have the same restrictions as k8s names, so we can save a character here by going from `--` to `.`

that said: there's a decent number of paasta service.instance pairs that share the same 50 char prefix - for the most part, it's just non-canary vs canary instance but https://fluffy.yelpcorp.com/i/cT7Bg7dx3WGdjmxSVvPNgt80kpflhrlg.html has a couple other collisions

not sure if this is fine for y'all's usecase or if we want to do something less collision-prone like what we do for label truncation in paasta (at the cost of needing to maintain a lookup table somewhere to actually use the APP_ID value during investigations)

nemacysts: (this is also probably significantly more likely to collide for tronjobs - i can generate a similar script for this if you'd like: i mostly didn't want to have to explain to an LLM how to parse tronfig dicts if that wasn't going to be useful :p)

paasta_tools/kubernetes_tools.py
```diff
@@ -2601,13 +2601,19 @@ def sanitize_for_config_hash(
...
                 container["env"] = [
                     e
                     for e in container["env"]
-                    if e.get("name", "") != "PAASTA_SOA_CONFIGS_SHA"
+                    if e.get("name", "")
+                    not in (
+                        "PAASTA_SOA_CONFIGS_SHA",
+                        # TODO: here temporarily to avoid a restart of all services
+                        # at the same time, remove once things have settled
+                        "AWS_SDK_UA_APP_ID",
```
nemacysts, line 2615: fwiw, i think we'll still big-bounce when this is removed from here since hashing the dict with and without this will result in different hashes, no?

paasta_tools/utils.py
piax93: We don't really need these to be accurate, they would just serve as a hint of where the requests would be coming from, so having them truncated is absolutely fine.

paasta_tools/kubernetes_tools.py
piax93: Ah, shoot, you are right, cause the job reconciling deployments just looks at the hash in the labels of the deployment, right? Rather than re-hashing the whole k8s config.

Welp, then I have no painless plan, if not just leave it excluded forever.

nemacysts: tron big bounces are a little more painful due to the config update endpoint being pretty slow, but hopefully my recent changes make that a little bit less painful

if anything, it'd be nice to roll this out for just paasta OR tron, and once things have settled, rollout for the other workload

piax93: I can adjust for that.

nemacysts (approved):
@piax93 is there any verification you'd want to do as this is rolling out or should we roll this out whenever is convenient for CI?

piax93:
@nemacysts Happy to go with "whenever is convenient for CI"

PR #4053, opened 2025-04-28
add auth support for Tron client

piax93:
I recently discovered that there is a duplicate implementation of a Tron API client in here, so I'm giving it [the same treatment of the main implementation](https://github.com/Yelp/Tron/blob/e4310a1cd81ae18de7008b44ca8480a40b75b071/tron/commands/client.py#L57-L60).

In this case, this client is not meant to be used by human, so we are going to use Vault identities for authentication.

PS: I had to refactor the auth methods to avoid a circular import.

### DIFF
--- a/paasta_tools/cli/authentication.py
+++ b/paasta_tools/cli/authentication.py
@@ -1,0 +1,77 @@
+# Copyright 2015-2016 Yelp Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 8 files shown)


nemacysts (approved):

nemacysts:
@piax93 i assume i can merge/release this whenever? looks like we're only setting `TRONCTL_API_AUTH` for the actual tronctl wrappers so this should be a noop if i were to merge now

nemacysts:
cc @Qmando since you might know the answer as well :)

piax93:
@nemacysts Yes, this is a noop right now

PR #4054, opened 2025-04-28
Enable labeling pods with AWS LB Controller readiness toggle

nemacysts:
As mentioned in the code comment below, the default way to enable this is per-namespace - that's somewhat annoying since we generally don't really edit namespace labels in PaaSTA post-creation (unless we're doing a one-time bulk edit): so let's instead see if we can configure the AWS LB Controller to only match on pods with this label (and remove the namespace selector)

NOTE: this pairs with https://github.yelpcorp.com/misc/compute-infra-k8s/pull/2505

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -510,6 +510,9 @@ instance MAY have:
 
     NOTE: it's possible to set only one of the bounds, if you only want to set a lower or upper bound (i.e., setting both min/max is not required).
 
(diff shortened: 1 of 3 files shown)


EvanKrall (dismissed):

EvanKrall (approved):
Could use a mention in yelpsoa_configs.rst

nemacysts:
> Could use a mention in yelpsoa_configs.rst

@EvanKrall done :)

EvanKrall (approved):

PR #4056, opened 2025-04-30
Take the max suggestion from autotune

nemacysts:
autotune currently does some aggregation for service instances where it takes into account the cpu (and if the cpu allocation changes, that's essentially treated as a new service instance) - this is likely done on purpose, but our current thinking is that this is hurting us more than it's helping us

however, changing the autotune query is a much bigger deal than changing the code that uses the results of that query, so we're thinking of simply taking the max from the results across these cpu aggregations - this is a little wasteful, but better than having values that are too low

### DIFF
--- a/paasta_tools/contrib/paasta_update_soa_memcpu.py
+++ b/paasta_tools/contrib/paasta_update_soa_memcpu.py
@@ -135,6 +135,11 @@ def cwd(path):
         os.chdir(pwd)
 
 
+def _force_str_to_int(value: str) -> int:
+    """Force convert strings into ints - even if they're technically floats."""
+    return int(float(value))
+
+
 def get_report_from_splunk(creds, app, filename, criteria_filter):
     """Expect a table containing at least the following fields:
     criteria (<service> kubernetes-<cluster_name> <instance>)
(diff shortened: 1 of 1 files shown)


paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -165,29 +170,83 @@ def get_report_from_splunk(creds, app, filename, criteria_filter):
...
+            "min_instances": ("suggested_min_instances", int),
+            "old_cpus": ("current_cpus", float),
+            # not quite sure why these are floats...they're ints in soaconfigs
+            "old_disk": ("current_disk", _force_str_to_int),
+            "old_mem": ("current_mem", _force_str_to_int),
+        }
+
+        for key, (source_key, conversion_function) in field_conversions.items():
+            value = d["result"].get(source_key)
+            serv[key] = conversion_function(value) if value is not None else None
```
jfongatyelp, line 211: I don't really know why we never had to convert these from strings before, but if we suspect this could be causing more autotune differences than we expect w/ this change, maybe we could just pass conversion_function as `max(v1, v2, key=conversion_function)` instead?

paasta_tools/contrib/paasta_update_soa_memcpu.py
```diff
@@ -165,29 +170,83 @@ def get_report_from_splunk(creds, app, filename, criteria_filter):
...
+        # this is ostensibly due to a theory that if you update resource allocation, existing
+        # autotune data is potentially invalidated - but in practice this ends up hampering
+        # autotune for services with highly variable resource allocation - e.g., we have some services
+        # that have their cpu allocation tweaked by +/-.1 cpu pretty frequently, but then min/max autotune
+        # is never updated.
+        if criteria in services_to_update:
+            for key in serv:
+                # we probably don't want to merge any other fields since they're going to be strings :p
+                if key not in field_conversions:
+                    continue
```
jfongatyelp, line 226: This might change results since we'd have been taking e.g. `*_reason` fields from only the latest row previously, but now we won't be overwriting those any longer even if we are overwtiting the values.  

I think it might add some complexity trying to make sure `_reason` fields stay with their respective values unfortunately tho; we'd probably need to consider a row as a distinct datapoint and figure out how to take the max of a datapoint + write all fields for that instead of going field by field?

jfongatyelp: oh, although I guess the `reason` is not really relevant here, and only in our splunk-based investigation dashboard, so i suppose this is moot

jfongatyelp: proving i can't read even further, i now see we don't even have reason fields in these `serv` dicts, apologies

jfongatyelp (dismissed):

paasta_tools/contrib/paasta_update_soa_memcpu.py
nemacysts: :o - clever!

i think that we end up converting later on, but this does sound like a less invasive change

KaspariK (approved):
Thank you for maybe saving me from my Splunk prison

paasta_tools/contrib/paasta_update_soa_memcpu.py
nemacysts: (i did this)

paasta_tools/contrib/paasta_update_soa_memcpu.py
nemacysts: yea, all of this is mighty confusing - i'm hoping that in autotune v2 we move this out of contrib/ and add a whole lot more comments and tests :p

ilkinmammadzada (approved):

PR #4057, opened 2025-05-01
move AWS_SDK_UA_APP_ID env var to parent instnace config

piax93:
... so that it'll affect Tron instances as well (follow up to https://github.com/Yelp/paasta/pull/4052).

I noticed that the effort of doing this in two steps is kinda wasted, as the `env` in a pod container spec is a list, and seems not to be sorted, so if the same key-value gets added but in different moments, it'll cause the config hash to be different.

A couple of possible options to address that:
* sort pod `env` by name, so we get one more big bounce, but it's stable from now on;
* `pop` and add back just this key in kube_tools (super ugly, but avoids a bounce);
* ignore the problem, and do another big bounce for this.


nemacysts (approved):
gah, not sorting strikes again!

I think another big bounce is fine since the couple of issues surfaced after the last one should be resolved (or we're aware of what to look for)

also: we should probably release this alongside https://github.com/Yelp/paasta/pull/4048 (cc @EvanKrall) since this is essentially a noop otherwise (and that change also causes a big bounce)

(i.e., CI should probably be the ones to merge this one so that we can minimize/coordinate the bounces)

EvanKrall (approved):

PR #4064, opened 2025-05-13
Check iam_role for existing ServiceAccounts

Qmando:
## Problem
When an `iam_role` changes underscores to dashes or vice versa, the normalization function will treat it as identical, and the ServiceAccount won't be updated.

## Solution
Actually verify the `iam_role` value matches when an existing ServiceAccount is found, and if it differs, patch it. 

## Verification
Unit tests only right now. Will update here if I perform additional testing.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -4325,21 +4325,42 @@ def ensure_service_account(
     kube_client: KubeClient,
     k8s_role: Optional[str] = None,
 ) -> None:
+    role_annotation = "eks.amazonaws.com/role-arn"
     sa_name = get_service_account_name(iam_role, k8s_role)
 
-    if not any(
-        sa.metadata and sa.metadata.name == sa_name
-        for sa in get_all_service_accounts(kube_client, namespace)
-    ):
+    existing_sa = None
+    for sa in get_all_service_accounts(kube_client, namespace):
+        if sa.metadata and sa.metadata.name == sa_name:
+            existing_sa = sa
+            break
+    else:
         sa = V1ServiceAccount(
             kind="ServiceAccount",
             metadata=V1ObjectMeta(
                 name=sa_name,
                 namespace=namespace,
-                annotations={"eks.amazonaws.com/role-arn": iam_role},
+                annotations={role_annotation: iam_role},
             ),
         )
         kube_client.core.create_namespaced_service_account(namespace=namespace, body=sa)
+    if existing_sa:
+        requires_patch = False
+        if (
+            sa.metadata.annotations
+            and sa.metadata.annotations.get(role_annotation, None) != iam_role
+        ):
+            sa.metadata.annotations[role_annotation] = iam_role
+            requires_patch = True
+        elif sa.metadata.annotations is None:
+            sa.metadata.annotations = {role_annotation: iam_role}
+            requires_patch = True
+        if requires_patch:
+            kube_client.core.patch_namespaced_service_account(
+                namespace=namespace, body=sa, name=sa.metadata.name
+            )
+            log.info(
+                f"Updated ServiceAccount {sa.metadata.name} iam_role to {iam_role}"
+            )
 
     # we're expecting that any Role dynamically associated with a Service Account already exists.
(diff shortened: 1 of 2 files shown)


piax93 (dismissed):

paasta_tools/kubernetes_tools.py
```diff
@@ -4325,21 +4325,42 @@ def ensure_service_account(
...
+        requires_patch = False
+        if (
+            sa.metadata.annotations
+            and sa.metadata.annotations.get(role_annotation, None) != iam_role
+        ):
+            sa.metadata.annotations[role_annotation] = iam_role
+            requires_patch = True
+        elif sa.metadata.annotations is None:
+            sa.metadata.annotations = {role_annotation: iam_role}
+            requires_patch = True
```
nemacysts, line 4356:
```suggestion
        if (
            not sa.metadata.annotations
            or sa.metadata.annotations.get(role_annotation, None) != iam_role
        ):
            # NOTE: we don't annotate SAs apart with anything other
            # than the pod identity role ARN, so this will remove
            # any annotations that folks may have manually added
            sa.metadata.annotations = {role_annotation: iam_role}
            requires_patch = True
```

should do more or less the same thing since we don't really expect any other annotations to be present :)

nemacysts (dismissed):

paasta_tools/kubernetes_tools.py
Qmando: Ack, thanks for confirming the extra logic isn't needed.

nemacysts (approved):
i have one more suggestion to clean things up, but otherwise lgtm

Qmando:
Hey @nemacysts can you clarify the last comment? Don't see another suggestion from you. Or was this related to adding a feature to delete unused SAs?

nemacysts:
> Hey @nemacysts can you clarify the last comment? Don't see another suggestion from you. Or was this related to adding a feature to delete unused SAs?

@Qmando uhhh, lemme re-read the PR: not quite sure happened to whatever comment I had actually meant to leave!

paasta_tools/kubernetes_tools.py
```diff
@@ -4325,21 +4325,42 @@ def ensure_service_account(
     kube_client: KubeClient,
     k8s_role: Optional[str] = None,
 ) -> None:
+    role_annotation = "eks.amazonaws.com/role-arn"
     sa_name = get_service_account_name(iam_role, k8s_role)
 
-    if not any(
-        sa.metadata and sa.metadata.name == sa_name
-        for sa in get_all_service_accounts(kube_client, namespace)
-    ):
```
nemacysts, line 4333: i don't quite know why we did things this way rather than getting a single SA with `read_namespaced_service_account()`, but this seems like something we can fix in another PR so as not to scope-creep this one

paasta_tools/kubernetes_tools.py
```diff
@@ -4325,21 +4325,42 @@ def ensure_service_account(
...
+            # any annotations that folks may have manually added
+            sa.metadata.annotations = {role_annotation: iam_role}
+            requires_patch = True
+        if requires_patch:
+            kube_client.core.patch_namespaced_service_account(
+                namespace=namespace, body=sa, name=sa.metadata.name
+            )
+            log.info(
+                f"Updated ServiceAccount {sa.metadata.name} iam_role to {iam_role}"
+            )
```
nemacysts, line 4363:
```suggestion
    if existing_sa:
        if (
            not sa.metadata.annotations
            or sa.metadata.annotations.get(role_annotation, None) != iam_role
        ):
            # NOTE: we don't annotate SAs apart with anything other
            # than the pod identity role ARN, so this will remove
            # any annotations that folks may have manually added
            sa.metadata.annotations = {role_annotation: iam_role}
            kube_client.core.patch_namespaced_service_account(
                namespace=namespace, body=sa, name=sa.metadata.name
            )
            log.info(
                f"Updated ServiceAccount {sa.metadata.name} iam_role to {iam_role}"
            )
```

non-blocking nit: it looks like we don't really need `requires_patch` - it's read immediately after being set and nothing happens in between, so we could merge these two block and get rid of the variable

nemacysts (dismissed):

jfongatyelp (approved):

PR #4065, opened 2025-05-14
Upgrade paasta to python 3.10

EmanElsaban:
Upgrade paasta to support python 3.10 and deleted this test: :tests/test_task_processing.py since we don't use task-processing in paasta.

All tests were passing locally using "make test"


requirements-dev-minimal.txt
```diff
@@ -1,3 +1,4 @@
+appdirs
```
nemacysts, line 1: just a note: the -minimal files should only contain libraries that we directly import - and it looks like this is most likely a transitive dependency

requirements-minimal.txt
nemacysts, line 1: note: i'm not quite sure why we ever bother with lower-bound pins in this file: we really should only ever have upper-bounds/explicit pins with comments - I don't think we ever really knowingly want to downgrade things and there's almost no way that we'll be disciplined enough to have truly accurate lower-bounds :P

nemacysts (commented):
(just a note, might wanna put this PR in draft mode)

requirements-minimal.txt
nemacysts, line 1: note: a bunch of these additions shouldn't be here - requirements-minimal is only for libraries we directly import

tests/test_kubernetes_tools.py
```diff
@@ -2925,7 +2923,12 @@ def test_get_volume_claim_templates(self):
             assert len(ret) == 2
 
     def test_get_storage_class_name_default(self):
-        pv = kubernetes_tools.PersistentVolume()
+        pv = kubernetes_tools.PersistentVolume(
+            mode="rw",
+            size=1000,
+            container_path="/tmp",
+            storage_class_name="ebs",
+        )
```
nemacysts, line 2931: i think this is changing the test behavior - what was the error here?

tox.ini
```diff
@@ -113,12 +113,14 @@ passenv =
     INDEX_URL_BUILD_ARG
 changedir=k8s_itests/
 commands =
-    # Build /etc/paasta used by docker-compose
-	{toxinidir}/k8s_itests/scripts/setup.sh
+    # Fail fast if Docker Compose V2 is not available
+    docker compose version
```
nemacysts, line 117: huh, this doesn't work on our devbox (at least, for me)

tox.ini
```diff
@@ -152,15 +154,17 @@ deps =
     behave==1.2.5
 commands =
     # TODO: upgrade behave if they ever take this reasonable PR
+    # starting from setuptools >= 58.0.0, use_2to3 wont be supported
+    # as behave includes use_2to3 in setup.py, it will fail to install
```
nemacysts, line 158: huh, how does this work now that we're on a newer setuptools already?

tox.ini
```diff
@@ -152,15 +154,17 @@ deps =
...
     pylint -E {toxinidir}/paasta_tools/mesos/ --ignore master.py,task.py
     behave {posargs}
 
 [testenv:mypy]
-basepython = python3.8
+basepython = python3.10
 deps =
     {[testenv]deps}
-    mypy==0.720
+    mypy==0.790
```
nemacysts, line 167: i dunno why we have a separate tox env for this and don't install mypy in requirements-dev - but that's a later us problem

paasta_tools/cli/cmds/local_run.py
```diff
@@ -977,12 +977,12 @@ def run_docker_container(
...
             # This is for text like config files
             with open(temp_secret_filename, "w") as f:
-                f.write(secret_content)
+                f.write(secret_content)  # type: ignore
         except TypeError:
             # If that fails, try to write it as bytes
             # This is for binary files like TLS keys
             with open(temp_secret_filename, "wb") as fb:
-                fb.write(secret_content)
+                fb.write(secret_content)  # type: ignore
```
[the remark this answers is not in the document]
EmanElsaban:
```
mypy run-test-pre: PYTHONHASHSEED='1607131619'
mypy run-test: commands[0] | mypy paasta_tools tests/frameworks tests/test_kubernetes_tools.py tests/test_long_running_service_tools.py tests/test_setup_kubernetes_job.py tests/test_setup_prometheus_adapter_config.py tests/test_secret_tools.py tests/secret_providers/test_secret_providers.py tests/secret_providers/test_vault.py tests/test_utils.py
paasta_tools/cli/cmds/local_run.py:980: error: Argument 1 to "write" of "IO" has incompatible type "Union[str, bytes]"; expected "str"
paasta_tools/cli/cmds/local_run.py:985: error: Argument 1 to "write" of "IO" has incompatible type "Union[str, bytes]"; expected "bytes"
Found 2 errors in 1 file (checked 279 source files)
```
seems that secret_content is ambigious. I see earlier in the code ``    secret_volumes = {}  # type: ignore``
which is what we get secret_content from.
```
for container_mount_path, secret_content in secret_volumes.items(
```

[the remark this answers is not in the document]
EmanElsaban: will add a comment 

```
# type: ignore  # secret_content can be str or bytes, and mypy cannot infer the correct type

```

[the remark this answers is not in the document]
EmanElsaban: we can probably also do something like this but i dont wanna play with the code's logic

```
if isinstance(secret_content, str):
    with open(temp_secret_filename, "w") as f:
        f.write(secret_content)
elif isinstance(secret_content, bytes):
    with open(temp_secret_filename, "wb") as fb:
        fb.write(secret_content)
else:
    raise TypeError("secret_content must be of type str or bytes")
```

requirements-bootstrap.txt
```diff
@@ -1,3 +1,3 @@
 pip==24.0
-setuptools==68.0.0
+setuptools==57.5.0
```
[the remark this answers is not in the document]
EmanElsaban: we're not doing that anymore :)

paasta_tools/instance/kubernetes.py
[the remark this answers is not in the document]
EmanElsaban: adding comments beside the ignores

tests/test_kubernetes_tools.py
EmanElsaban: mypy errors

```
tests/test_kubernetes_tools.py:2926: error: Keys ('mode', 'size', 'container_path', 'storage_class_name') missing for TypedDict "PersistentVolume"
Found 1 error in 1 file (checked 279 source files)
```

EmanElsaban: this is why I added these keys

EmanElsaban: will add an ignore comment

paasta_tools/cli/cmds/local_run.py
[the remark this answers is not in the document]
nemacysts: we should probably enable the `show_error_codes = true` mypy option and ignore the specific error here - the comment should maybe be something like ` # TODO: make this type-safe rather than rely on exceptions` - mypy is correctly inferring that this can be either type, but it can't tell that this is safe due to the exception handling (but also, we can do this without exceptions later :p)

tox.ini
EmanElsaban: hmm thats a good point

EmanElsaban: I'm checking the jenkins run tho and its passing: https://github.com/Yelp/paasta/actions/runs/15856531471/job/44703167701?pr=4065

EmanElsaban:
```
k8s_itests run-test: commands[2] | docker compose down
WARNING: test command found but not installed in testenv
  cmd: /usr/bin/docker
  env: /home/runner/work/paasta/paasta/.tox/k8s_itests
Maybe you forgot to specify a dependency? See also the allowlist_externals envconfig setting.
```

tests/test_kubernetes_tools.py
```diff
@@ -2925,7 +2923,7 @@ def test_get_volume_claim_templates(self):
             assert len(ret) == 2
 
     def test_get_storage_class_name_default(self):
-        pv = kubernetes_tools.PersistentVolume()
+        pv = kubernetes_tools.PersistentVolume()  # type: ignore  # PersistentVolume requires keys ('mode', 'size', 'container_path', 'storage_class_name') as per its TypedDict definition, but they are missing here; ignoring for test setup
```
nemacysts: this should probably be a comment to make this typeddict non-total

tox.ini
EmanElsaban: erroring locally, looking into it

paasta_tools/metrics/metastatus_lib.py
```diff
@@ -614,7 +614,7 @@ def group_slaves_by_key_func(
     """
     sorted_slaves: Sequence[_GenericNodeT]
     if sort_func is None:
-        sorted_slaves = sorted(slaves, key=key_func)
+        sorted_slaves = sorted(slaves, key=key_func)  # type: ignore
```
nemacysts: i'd say this should have a comment, but we'll be deleting this code soon :p

paasta_tools/instance/kubernetes.py
nemacysts, line 1: i'm not quite sure what's going on with our types here - there's definitely some weirdness going on here with Dict vs Mapping / List vs Sequence as well as with things that are completely wrong (Task[Sequence[Any]] vs Future[Dict[str, Any]])

imo, we should ticket fixing these for real and have that ticket be the comment rather than re-iterate the mypy output

.github/workflows/ci.yml
```diff
@@ -51,12 +51,12 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        dist: [bionic, jammy, noble]
```
nemacysts, line 54: note: we'll need to do some puppet shenanigans / send an email before we merge this

.github/workflows/ci.yml
nemacysts, line 1: i can't comment on the line since it's not included in the diff, but the `cut_release` target here still references bionic

paasta_tools/cli/cmds/local_run.py
```diff
@@ -977,12 +977,12 @@ def run_docker_container(
             # First try to write the file as a string
             # This is for text like config files
             with open(temp_secret_filename, "w") as f:
-                f.write(secret_content)
+                f.write(secret_content)  # type: ignore  # secret_content can be str or bytes, and mypy cannot infer the correct type
```
nemacysts:
```suggestion
                f.write(secret_content)  # type: ignore  # TODO: make this type-safe rather than rely on exceptions
```

paasta_tools/cli/cmds/local_run.py
```diff
@@ -977,12 +977,12 @@ def run_docker_container(
...
         except TypeError:
             # If that fails, try to write it as bytes
             # This is for binary files like TLS keys
             with open(temp_secret_filename, "wb") as fb:
-                fb.write(secret_content)
+                fb.write(secret_content)  # type: ignore  # secret_content can be str or bytes, and mypy cannot infer the correct type
```
nemacysts:
```suggestion
                fb.write(secret_content)  # type: ignore  # TODO: make this type-safe rather than rely on exceptions
```

paasta_tools/cli/cmds/local_run.py
nemacysts, line 1: that said, we can also continue to rely on exceptions and use `cast()` to silence mypy - but a TODO and swapping this to 
```python
write_mode = "wb" if isinstance(secret_content, bytes) else "w"
with open(temp_secret_filename, mode=write_mode) as f:
    f.write(secret_content)
```
later seems fine too

paasta_tools/instance/kubernetes.py
```diff
@@ -641,7 +641,7 @@ async def kubernetes_status_v2(
                 kube_client, job_config, job_config.get_kubernetes_namespace()
             )
         )
-        tasks.append(autoscaling_task)
+        tasks.append(autoscaling_task)  # type: ignore  # PAASTA-18698; ignoring due to type mismatch
```
nemacysts:
```suggestion
        tasks.append(autoscaling_task)  # type: ignore  # PAASTA-18698; ignoring due to unexpected type mismatch
```
might be a little more descriptive, but this is fine too

requirements.txt
```diff
@@ -119,11 +118,10 @@ urllib3==1.26.18
 utaw==0.2.0
 venusian==1.1.0
 webcolors==1.7
-WebOb==1.7.4
+WebOb==1.8.7 # required by pyramid==2.0.2
```
nemacysts: we can drop these comments :)

tox.ini
nemacysts: (also, we really should drop behave at some point)

paasta_tools/cli/cmds/local_run.py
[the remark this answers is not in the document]
EmanElsaban: sounds good, added show_error_codes and modified the comment

PR #4068, opened 2025-05-19
Enable temporary min-bound autoscaling overrides

nemacysts:
As it says on the tin: this allows folks to temporarily bump min_instances for an arbitrary amount of time - this is useful for scenarios like region failovers where our service autoscaler may have scaled down a service in one region, but we're preparing to shift traffic back and want to ensure we have the capacity to serve requests (or this can even be useful in the other direction where we're shifting traffic and want to pre-emptively scale up services)

NOTE: we still need to add a cronjob (or something else) to cleanup the configmap after an override has expired


paasta_tools/api/views/autoscaler.py
```diff
@@ -98,3 +113,111 @@ def update_autoscaler_count(request):
...
+
+    max_instances = instance_config.get_max_instances()
+    if max_instances is None:
+        raise ApiFailure(f"Autoscaling is not enabled for {service}.{instance}", 400)
+
+    if max_instances < min_instances_override:
+        raise ApiFailure(
+            f"min_instances ({min_instances_override}) cannot be greater than max_instances ({max_instances})",
+            400,
+        )
```
nemacysts, line 176: maybe we should also prevent setting min_instances below the soaconfigs-provided value as well?

jfongatyelp (dismissed):

jfongatyelp (approved):

PR #4069, opened 2025-05-19
Support None in `git_url` for toolboxes or other future pseudo-sevices

Qmando:
Allow setting the `git_url` of a service to `None` in order to have the generate deployment step skip it. Currently, if it tries to access a repo that doesn't exist, it will crash and the generate deployment script will stop completely. Returning these empty mapping objects is the same behavior as when a service "has no valid deploy groups" so it should be handled fine.

It will only be a `None` when it's explicitly set, as it will otherwise generate a default git url.

```
~/paasta (toolbox_generate_deployment) $ paasta validate -y /nail/home/qlo/yelpsoa-configs -s prod-toolbox
✓ Successfully validated schema: adhoc-pnw-prod.yaml
✓ Successfully validated schema: service.yaml
✓ All PaaSTA Instances for are valid for all clusters
✓ All prod-toolbox's instance names in cluster norcal-devc are unique
✓ All prod-toolbox's instance names in cluster norcal-stagef are unique
✓ All prod-toolbox's instance names in cluster norcal-stageg are unique
✓ All prod-toolbox's instance names in cluster nova-prod are unique
✓ All prod-toolbox's instance names in cluster pnw-corp are unique
✓ All prod-toolbox's instance names in cluster pnw-corpdev are unique
✓ All prod-toolbox's instance names in cluster pnw-devc are unique
✓ All prod-toolbox's instance names in cluster pnw-prod are unique
✓ No orphan secrets found

~/paasta (toolbox_generate_deployment) $ cat /nail/home/qlo/yelpsoa-configs/prod-toolbox/service.yaml
...
git_url: null
```

### DIFF
--- a/paasta_tools/cli/schemas/service_schema.json
+++ b/paasta_tools/cli/schemas/service_schema.json
@@ -10,7 +10,10 @@
             "type": "string"
         },
         "git_url": {
-            "type": "string",
+            "type": [
+                "string",
+                "null"
+            ],
             "pattern": "^git@github.yelpcorp.com:[-a-z]+/[-_a-z0-9]+(\\.git)?$",
             "$comment": "this is obviously very tied to how we name repos at Yelp"
         },
--- a/paasta_tools/generate_deployments_for_service.py
+++ b/paasta_tools/generate_deployments_for_service.py
@@ -130,6 +130,10 @@ def get_deploy_group_mappings(
     v2_mappings: V2_Mappings = {"deployments": {}, "controls": {}}
     git_url = get_git_url(service=service, soa_dir=soa_dir)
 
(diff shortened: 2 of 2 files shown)


piax93 (approved):
Seems a reasonable approach to me, but of course leaving the last word to CI

jfongatyelp (approved):

jfongatyelp:
I think this could cause (rarely used) paasta cli tools like `paasta check`/`paasta info` to give some failures as a side effect (if they aren't already broken with these pseudo-services), but seems prudent to allow this.

nemacysts:
a more native approach would probably be to use DeploymentVersions so that these can be deployed as a "normal" paasta service using mark-for-deployment, but that would require some more CI/CD-side work to enable (i.e., writing a custom jenkinsfile)

PR #4074, opened 2025-05-28
Add better error handling for 403s in remote-run

Qmando:
### Swagger spec
The swagger spec specified that for 403s in remote_run, there should be no data in the response. However, there are error messages that should be returned. Because of this, they got wrapped in another error:
```
The server has either erred or is incapable of performing the requested operation.

Response body should be empty: {\"reason\": \"User not owning service\"}
```

### Client error handling
The remote run code expects non-200s to be available in the client response objects, however, the current code will automatically raise ApiExceptions. So rather than "Error from PaaSTA APIs while...", you'd get a full client side traceback.

### Swagger spec
Add `reason` as a response type in swagger.json. I'm not sure if this needs to be reflected in oapi.yaml as well, as that doesn't seem to have any effect. I've also left every other API unchanged, even though some could generate 500s or 403s. These are the two that are most important. 

### Client error handling
Catch ApiExceptions and try parsing error body as best as possible. 

## Validation
```
paasta remote-run start -c cluster -s service -i main --interactive
Error from PaaSTA APIs while starting job: User not owning service

$ paasta remote-run stop -c cluster -s service -i main
Error from PaaSTA APIs while stopping job: User not owning service
```


piax93 (commented):
That's my bad, I assumed that swagger would treat responses without schema as "schemaless" and not bother to validate them, so we could do error responses anyway we liked. I guess I was wrong.

There's also [oapi.yaml](https://github.com/Yelp/paasta/blob/master/paasta_tools/api/api_docs/oapi.yaml), which is what's used to generate the API client which will need to be updated. Why do we keep two copies of pretty much the same thing? I have no idea, and it's very annoying.

nemacysts:
> There's also [oapi.yaml](https://github.com/Yelp/paasta/blob/master/paasta_tools/api/api_docs/oapi.yaml), which is what's used to generate the API client which will need to be updated. Why do we keep two copies of pretty much the same thing? I have no idea, and it's very annoying.

my understanding is that we meant to switch to oapi.yaml...but then never finished moving the final bit of the api to look at that file

nemacysts (dismissed):

piax93 (dismissed):

jvperrin (dismissed):
Nice

piax93 (approved):

nemacysts (approved):

PR #4076, opened 2025-05-30
Adding idempotent label to Tron pods - TRON-2154

EmanElsaban:
This PR adds a new field in soaconfigs for "idempotent", which will help us identify which Tron actions are idempotent vs which ones are not. It also creates a pod label with its value

### DIFF
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -91,6 +91,11 @@
                     },
                     "uniqueItems": true
                 },
+                "idempotent": {
+                    "type": "boolean",
+                    "default": false,
+                    "$comment": "This will be used to determine whether the action can be retried without side effects."
+                },
                 "extra_constraints": {
                     "type": "array",
                     "items": {
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -422,6 +422,9 @@ class TronActionConfig(InstanceConfig):
     def get_job_name(self):
         return self.job
 
+    def get_idempotent(self) -> bool:
+        return self.config_dict.get("idempotent", False)
+
     def get_action_name(self):
         return self.action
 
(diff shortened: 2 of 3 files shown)


paasta_tools/cli/schemas/tron_schema.json
```diff
@@ -91,6 +91,11 @@
                     },
                     "uniqueItems": true
                 },
+                "idempotent": {
+                    "type": "boolean",
+                    "default": false,
+                    "$comment": "This will be used to determine whether the action can be retried without side effects."
```
nemacysts, line 97: feel free to drop:
```suggestion
                    "$comment": If true, this action can be safely retried multiple times without unwanted side-effects."
```

paasta_tools/tron_tools.py
```diff
@@ -1046,6 +1051,9 @@ def format_tron_action_dict(action_config: TronActionConfig):
                 limit=63,
                 suffix=4,
             ),
+            "tron.yelp.com/idempotent-action": str(
```
nemacysts, line 1054: i'd personally prefer
```suggestion
            "tron.yelp.com/is-idempotent": str(
```
...but i also can't really justify why, so up to you :)

nemacysts (commented):
hmm, i'm not sure if we should document this flag in the paasta RTD or the tron one - https://paasta.readthedocs.io/en/latest/yelpsoa_configs.html#tron-clustername-yaml mentions that we only document paasta-specific stuff...but then we document some tron-only stuff anyway :p


that said: this generally looks good to me - although i'd prefer if more folks from the team would chime in on your proposal doc to make sure we're not just unilaterally picking what we want to do :p

paasta_tools/tron_tools.py
KaspariK: If you do this don't commit the suggestion, there is a typo in "idempontent" lol

nemacysts: i can't type 🤣

nemacysts: (fixed the suggestion)

PR #4077, opened 2025-06-06
Revert "Check iam_role for existing ServiceAccounts"

ilkinmammadzada:
Reverts Yelp/paasta#4064


piax93:
This is actually not the problematic part

PR #4078, opened 2025-06-09
Allow setting temporary max_instances overrides

nemacysts:
Just as setting temporary `min_instances` overrides is useful, so would being able to do the same for `max_instances`.

### DIFF
--- a/paasta_tools/api/api_docs/oapi.yaml
+++ b/paasta_tools/api/api_docs/oapi.yaml
@@ -269,12 +269,17 @@ components:
           type: integer
           description: Minimum number of instances to run
           minimum: 1
+        max_instances:
+          type: integer
+          description: Maximum number of instances to run
+          minimum: 1
         expire_after:
           format: float
           type: number
           description: Unix timestamp when this override is no longer valid
       required:
         - min_instances
+        - max_instances
         - expire_after
     InstanceBounceStatus:
       properties:
@@ -1494,7 +1499,12 @@ paths:
                     description: Instance name
                   min_instances:
                     type: integer
+                    nullable: true
                     description: Minimum number of instances to run
+                  max_instances:
+                    type: integer
+                    nullable: true
+                    description: Maximum number of instances to run
                   expire_after:
                     format: float
                     type: number
--- a/paasta_tools/api/views/autoscaler.py
+++ b/paasta_tools/api/views/autoscaler.py
@@ -156,19 +156,24 @@ def set_autoscaling_override(request):
 
     json_body = request.swagger_data.get("json_body", {})
     min_instances_override = json_body.get("min_instances")
+    max_instances_override = json_body.get("max_instances")
     expire_after = json_body.get("expire_after")
 
     if not isinstance(min_instances_override, int) or min_instances_override < 1:
         raise ApiFailure("min_instances must be a positive integer", 400)
 
+    if not isinstance(max_instances_override, int) or max_instances_override < 1:
+        raise ApiFailure("max_instances must be a positive integer", 400)
+
     if not expire_after:
         raise ApiFailure("expire_after is required", 400)
 
-    max_instances = instance_config.get_max_instances()
+    min_instances = min_instances_override or instance_config.get_min_instances()
+    max_instances = max_instances_override or instance_config.get_max_instances()
     if max_instances is None:
         raise ApiFailure(f"Autoscaling is not enabled for {service}.{instance}", 400)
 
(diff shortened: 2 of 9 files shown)


paasta_tools/api/views/autoscaler.py
```diff
@@ -156,19 +156,24 @@ def set_autoscaling_override(request):
 
     json_body = request.swagger_data.get("json_body", {})
     min_instances_override = json_body.get("min_instances")
+    max_instances_override = json_body.get("max_instances")
```
KaspariK, line 159: Nit, but should we add max_instances to the docstring?

paasta_tools/api/views/autoscaler.py
```diff
@@ -156,19 +156,24 @@ def set_autoscaling_override(request):
...
+    if not isinstance(max_instances_override, int) or max_instances_override < 1:
+        raise ApiFailure("max_instances must be a positive integer", 400)
+
     if not expire_after:
         raise ApiFailure("expire_after is required", 400)
 
-    max_instances = instance_config.get_max_instances()
+    min_instances = min_instances_override or instance_config.get_min_instances()
+    max_instances = max_instances_override or instance_config.get_max_instances()
     if max_instances is None:
```
KaspariK: Is this check dead given the previous validation?

paasta_tools/cli/cmds/autoscale.py
```diff
@@ -75,6 +75,14 @@ def add_subparser(subparsers):
         else autoscale_parser.error("Minimum instances must be >= 1"),
         default=None,
     )
+    override_group.add_argument(
```
KaspariK, line 78: Do we want a max max to avoid some edge case where a typo + some bad behaviour overloads a cluster? Or is that too paranoid?

paasta_tools/cli/cmds/autoscale.py
```diff
@@ -75,6 +75,14 @@ def add_subparser(subparsers):
...
+        "--set-max",
+        help="Set the maximum number of replicas (must be >= 1). Requires --for parameter.",
+        type=lambda x: int(x)
+        if int(x) >= 1
+        else autoscale_parser.error("Maximum instances must be >= 1"),
+        default=None,
+    )
     override_group.add_argument(
```
KaspariK, line 86: Super nit, but --for applies to both set-min and set-max now, right?

paasta_tools/api/views/autoscaler.py
KaspariK: This may or may not be based on my misunderstanding of how this falls back...

nemacysts: yes and no - max_instances can still be None here: but I think we need to move this around a bit so that we can't use this command to start autoscaling something that's not autoscaled (since most of the time we don't actually want cpu autoscaling)

...and i also just realized that maybe I should add an cleanup cronjob first or something - if you set a min_instances override N days in the past and then go to set a max_instances override later one, the min_instances override will kick in as well.

paasta_tools/cli/cmds/autoscale.py
nemacysts: we talked a little bit in our 1:1 - a max max is probably fine to obviate, but we might want some sort of similar concept for min (probably just ask for confirmation rather than hard-block things)

nemacysts: (i made a ticket for this)

paasta_tools/api/views/autoscaler.py
nemacysts: (cronjob was added and max_instances check moved around)

paasta_tools/api/api_docs/oapi.yaml
```diff
@@ -267,14 +267,19 @@ components:
...
-          minimum: 1
+        max_instances:
+          type: integer
+          nullable: true
+          description: Maximum number of instances to run
         expire_after:
           format: float
           type: number
           description: Unix timestamp when this override is no longer valid
       required:
```
KaspariK, line 280: This requirement is strange

KaspariK: nvm lol

KaspariK (approved):

paasta_tools/api/api_docs/oapi.yaml
nemacysts: yea, the override structure is a little weird here - maybe once we have a single api spec i can try to do some openapi shenanigans to better specify this

PR #4081, opened 2025-06-11
Add paasta_cleanup_expired_autoscaling_overrides

nemacysts:
As it says on the tin: we should probably be expiring things from this
configmap rather than leaving them around 5eva.


jfongatyelp (dismissed):

paasta_tools/cleanup_expired_autoscaling_overrides.py
```diff
@@ -0,0 +1,155 @@
...
+    )
+    args = parser.parse_args()
+    return args
+
+
+def setup_logging(verbose):
+    level = logging.DEBUG if verbose else logging.INFO
+    logging.basicConfig(level=level)
+    # silence kubernetes client metadata logs since debug ones are pretty noisy
+    # logging.getLogger("kubernetes.client.rest").setLevel(logging.ERROR)
```
nemacysts:
```suggestion
    logging.getLogger("kubernetes.client.rest").setLevel(logging.ERROR)
```

whoops, forgot to uncomment this

PR #3874, opened 2024-05-17
Delete more mesos-related code

EvanKrall:

PR #4088, opened 2025-06-25
Delete more Mesos code

EvanKrall:


paasta_tools/tron_tools.py
```diff
@@ -670,10 +668,7 @@ def validate(self):
         error_msgs.extend(super().validate())
         # Tron is a little special, because it can *not* have a deploy group
         # But only if an action is running via ssh and not via paasta
-        if (
-            self.get_deploy_group() is None
-            and self.get_executor() in MESOS_EXECUTOR_NAMES
-        ):
+        if self.get_deploy_group() is None and self.get_executor() != "ssh":
```
EvanKrall, line 671: I tweaked this manually. This will make this check stricter -- anything with executor: spark will now have this check.

tests/metrics/test_metastatus_lib.py
```diff
@@ -39,17 +38,7 @@ def test_fail_check_threshold():
...
-    fake_mesos_state = {
-        "slaves": [{"reserved_resources": {"maintenance": {"cpus": 1}}}]
-    }
-    total, used, available = metastatus_lib.get_mesos_cpu_status(
-        fake_metrics, fake_mesos_state
-    )
-    assert total == 3
-    assert used == 2
-    assert available == 1
+# test_get_mesos_cpu_status removed - function deleted with Mesos cleanup
```
EvanKrall: what the fuck

paasta_tools/cli/cmds/local_run.py
```diff
@@ -134,6 +133,24 @@ def perform_tcp_healthcheck(url, timeout):
         return (False, "%s (timeout %d seconds)" % (os.strerror(result), timeout))
 
 
+def execute_in_container(docker_client, container_id, command, timeout):
```
nemacysts: do we need to port any of the other logic that used to be in the `paasta_execute_docker_command` version?

(would also be nice to add some types to this since this isn't just a 1:1 move, but i won't block on that :p)

paasta_tools/contrib/get_running_task_allocation.py
```diff
@@ -261,9 +186,7 @@ def get_task_allocation_info(
     namespace: str,
     kube_client: Optional[KubeClient],
 ) -> Iterable[TaskAllocationInfo]:
-    if scheduler == "mesos":
-        return get_mesos_task_allocation_info()
-    elif scheduler == "kubernetes":
+    if scheduler == "kubernetes":
```
nemacysts: i think we can actually get rid of all the scheduler conditions: we don't pass --scheduler in puppet so we can clean this up without breaking anything

nemacysts: (but also fine to further cleanup later)

paasta_tools/metrics/metastatus_lib.py
nemacysts, line 1: shouldn't this entire file be deleteable? we deleted `paasta metastatus` back in https://github.com/Yelp/paasta/pull/3874

paasta_tools/cli/cmds/local_run.py
EvanKrall: silly robot, dunno why it rewrote this.

That said, the existing code seems to have a bug that I think I actually hit while using local-run the other day -- if something execs into the container before paasta does, it looks like it doesn't create an exec.

paasta_tools/contrib/get_running_task_allocation.py
```diff
@@ -248,36 +172,15 @@ def get_kubernetes_task_allocation_info(
                     host_ip=info.get("host_ip"),
                     git_sha=info.get("git_sha"),
                     config_sha=info.get("config_sha"),
-                    mesos_container_id=None,
```
nemacysts, line 251: note: might be worth checking that the consumers of this will handle this field not existing at all anymore (pretty sure only splunk reads this and that it will gracefully handle things tho)

paasta_tools/cli/cmds/local_run.py
EvanKrall: Ok, I've put back the original function and fixed the bug. Now if the loop that searches for an existing exec doesn't find the one we're looking for, we'll create a new one.

nemacysts (commented):
(looks like there's still some references to deleted files)

paasta_tools/contrib/get_running_task_allocation.py
EvanKrall: looks like the savedsearch might do exactly the opposite of what we want if the mesos_container_id is missing. It's currently checking specifically for "null".

PR #4092, opened 2025-07-03
Switch default spark user from mrjob to spark_driver

Qmando:
I've recently created this new IAM user to replace `mrjob`. It has the necessary permissions for most Spark jobs. I'm doing this so we can eventually delete the mrjob user.

Note that this is only used if a user does *not* specify aws profile or another credentials file, so this isn't likely to break anything, as few if any spark jobs are using this yet.


nurdann (approved):

Qmando:
Should be safe to release everywhere.

PR #4093, opened 2025-07-11
Enable conditional setting of unhealthyPodEvictionPolicy on PDBs

nemacysts:
This will ensure that pods in CrashLoopBackOff don't impact pool scaledowns (since otherwise, Karpenter will refuse to consolidate nodes due to PDBs)

This is not enabled by default due to our legacy clusters, but will be toggled at a per-cluster level through SystemPaastaConfig.

This PR is also somewhat spicy due to these legacy clusters, so we'll definitely need to do some testing to make sure that the upgraded k8s clientlib doesn't cause chaos there (or we'll have to take more drastic action).

That said, I've upgraded the k8s clientlib version to a somewhat more modern version to match the oldest non-legacy cluster we have and therefore had to make a number of changes (e.g., V1Subject was renamed to RbacV1Subject, etc)

TODO
---

- [ ] Fix HPA v2beta2 imports (how does this code work atm?!?!?!!
- [ ] Test this in an actual cluster (legacy and non-legacy)

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -50,6 +50,7 @@ from kubernetes import client as kube_client
 from kubernetes import config as kube_config
 from kubernetes.client import CoreV1Event
 from kubernetes.client import models
+from kubernetes.client import RbacV1Subject
 from kubernetes.client import V1Affinity
 from kubernetes.client import V1AWSElasticBlockStoreVolumeSource
 from kubernetes.client import V1Capabilities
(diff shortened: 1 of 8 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -3393,20 +3393,30 @@ def pod_disruption_budget_for_service_instance(
...
+    selector = V1LabelSelector(
+        match_labels={
+            "paasta.yelp.com/service": service,
+            "paasta.yelp.com/instance": instance,
+        }
+    )
+    if load_system_paasta_config().get_enable_unhealthy_pod_eviction():
+        spec = V1PodDisruptionBudgetSpec(
+            max_unavailable=max_unavailable,
+            unhealthy_pod_eviction_policy="AlwaysAllow",  # XXX: should this be configurable?
```
ilkinmammadzada, line 3405: I would always use `unhealthy_pod_eviction_policy` by reading value from config with default value ("IfHealthyBudget")

ilkinmammadzada: But.. I forgot about k8s clusters with older version. Do we pin old PaaSTA version there?

ajayOO8:
Forked this into a new PR https://github.com/Yelp/paasta/pull/4146 with some changes.
(resolving git conflict was a pain)
closing this one

PR #4095, opened 2025-07-16
remote-run: add option to pass command, and follow logs

piax93:
I initially thought to just pass the command to kubectl, but it's actually cleaner to propagate it in the job definition, as we can have the pod gracefully terminate once the execution is over, rather than relying on the Job deadline or the user terminating the remote-run explicitly. A bit more efficient.

Also realized that until now we pretty much assume non-interactive execution to happen blindly, which may not be great, so I added a flag to follow logs with kubectl. Don't know if we want to make that default behaviour.


paasta_tools/cli/cmds/remote_run.py
```diff
@@ -96,6 +99,7 @@ def paasta_remote_run_start(
                 recreate=args.recreate,
                 max_duration=args.max_duration,
                 toolbox=args.toolbox,
+                command=args.cmd or "",
```
nemacysts, line 102: `remote_run_start` is typed as accepting Optional[str] for this - do we need to have this `or ""`?

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -269,6 +285,13 @@ def add_subparser(subparsers: argparse._SubParsersAction) -> None:
         type=int,
         default=1800,
     )
+    start_parser.add_argument(
+        "-f",
+        "--follow",
+        help="Attach to and follow container output",
+        action="store_true",
+        default=False,
```
nemacysts, line 293: i guess if folks don't use this the logs would be available through our  usual container logging pipeline?

that said, imo it's probably more likely that folks wanna see the output of commands :p

nemacysts (approved):

paasta_tools/cli/cmds/remote_run.py
piax93: I'm afraid of what swagger API validation may do, since we live in this half-baked world where the client is generated from v3, in which I could set the property as nullable, while the API server uses v2 as the spec, where that's not possible.

piax93: (yes, this is a passive aggressive way of saying I'd really like only one spec existed)

paasta_tools/cli/cmds/remote_run.py
piax93: Correct. I added this because as a replacement for local-run, it makes sense to have an option to see what stdout/stderr say, rather than having to rely on a separate tool tailing service logs.

PR #4096, opened 2025-07-16
Remote run copy file

Qmando:
Support copying files from remote run jobs and also adding additional files if the job is already running.

Example: https://fluffy.yelpcorp.com/i/CXvf9mq5HdkZJ93bbm2fwsfqwMPFflbh.html
Toolbox example: https://fluffy.yelpcorp.com/i/46FPk43js5W0NrHBPbTqFbjC0K25gpp6.html


paasta_tools/cli/cmds/remote_run.py
```diff
@@ -71,6 +78,81 @@ def parse_error(body: str) -> str:
...
+    poll_response = client.remote_run.remote_run_poll(
+        service=args.service,
+        instance=args.instance,
+        job_name=job_name,
+        user=user,
+        toolbox=args.toolbox,
+    )
+    if poll_response.status != 200:
+        print(f"Unable to find running remote-run pod")
+        return 1
```
nemacysts: !8ball it's worth adding a while loop if the results is a 204 to support cases like `paasta remote-run start ...; paasta remote-run copy --to-pod ...`

that said, its fine to punt on this - just a thought :p

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -71,6 +78,81 @@ def parse_error(body: str) -> str:
...
+
+    poll_response = client.remote_run.remote_run_poll(
+        service=args.service,
+        instance=args.instance,
+        job_name=job_name,
+        user=user,
+        toolbox=args.toolbox,
+    )
+    if poll_response.status != 200:
+        print(f"Unable to find running remote-run pod")
```
nemacysts: non-blocking usability suggestion:
```suggestion
        print(f"Unable to find running remote-run pod: have you started one with `paasta remote-run start`?")
```

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -71,6 +78,81 @@ def parse_error(body: str) -> str:
...
+
+    if args.toolbox:
+        run_interactive_cli(exec_command)
+    else:
+        token_response = client.remote_run.remote_run_token(
+            args.service, args.instance, user
+        )
+        kubectl_wrapper = f"kubectl-eks-{args.cluster}"
+        if not shutil.which(kubectl_wrapper):
+            kubectl_wrapper = f"kubectl-{args.cluster}"
```
nemacysts: non-blocking: might be nice to turn this into a function at some point since this isn't the first place we have these lines :p

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -71,6 +78,81 @@ def parse_error(body: str) -> str:
...
+    )
+    if poll_response.status != 200:
+        print(f"Unable to find running remote-run pod")
+        return 1
+
+    if args.to_pod:
+        kubectl_cp_template = KUBECTL_CP_TO_CMD_TEMPLATE
+        exec_command = f"scp -A {args.copy_file_source} {poll_response.pod_address}:{args.copy_file_dest}"
+        if not args.copy_file_dest.startswith("/tmp") and not args.toolbox:
+            args.copy_file_dest = os.path.join("/tmp/", args.copy_file_dest)
```
nemacysts: i assume we're forcing /tmp/ so that folks can't be naughty and overwrite arbitrary files in the container - but is it possible that `copy_file_dest` is vulnerable to a path traversal attack? e.g., `../my_important_dir/my_important_file.py`?

nemacysts, line 129: am i getting confused or should this be above the `exec_command`?

nemacysts (commented):
(looks like there's a failing test btw)

paasta_tools/cli/cmds/remote_run.py
Qmando: It seems as though the file will be copied as the container user (nobody usually) and therefore does not have permissions to overwrite files that the user cannot themselves.
```
$ paasta remote-run copy --to-pod ... -i main input.txt ../input.txt
Error copying file from remote-run pod:
INFO: You are using an Okta authenticated kubectl wrapper
tar: input.txt: Cannot open: Permission denied
tar: Exiting with failure status due to previous errors
command terminated with exit code 2
```

Qmando: Added a comment here. It's a little misleading since `exec_command` is used for ssh/toolbox, and the command template is used for kube/nontoolbox copy. I lumped them together here just to save a couple extra if/else statements.

The exec_command is formatted already because we just directly execute it next. The kube one is a template because we need to fetch the token first before formatting it.

paasta_tools/cli/cmds/remote_run.py
Qmando: You can copy files as part of the start command, so I don't feel like it's really necessary 🤷

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -74,6 +81,85 @@ def parse_error(body: str) -> str:
...
+        print("Cannot get a paasta-api client")
+        return 1
+
+    # Create the config and extract the job name
+    user = get_username()
+    deployment_config = (
+        generate_toolbox_deployment(args.service, args.cluster, user)
+        if args.toolbox
+        else load_eks_service_config(args.service, args.instance, args.cluster)
+    )
```
piax93: this may need to be updated after https://github.com/Yelp/paasta/pull/4099

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -74,6 +81,85 @@ def parse_error(body: str) -> str:
...
+        if not args.copy_file_dest.startswith("/tmp") and not args.toolbox:
+            args.copy_file_dest = os.path.join("/tmp/", args.copy_file_dest)
+    else:
+        exec_command = f"scp -A {poll_response.pod_address}:{args.copy_file_source} {args.copy_file_dest}"
+        kubectl_cp_template = KUBECTL_CP_FROM_CMD_TEMPLATE
+
+    # Kubectl cp doesnt like directories as a destination
+    if os.path.isdir(args.copy_file_dest):
+        args.copy_file_dest = os.path.join(
+            args.copy_file_dest, os.path.split(args.copy_file_source)[1]
```
piax93: more readable
```suggestion
            args.copy_file_dest, os.path.basename(args.copy_file_source)
```

paasta_tools/cli/cmds/remote_run.py
```diff
@@ -74,6 +81,85 @@ def parse_error(body: str) -> str:
...
+        )
+
+    if args.toolbox:
+        run_interactive_cli(exec_command)
```
piax93: Does `scp` need a tty?

paasta_tools/cli/cmds/remote_run.py
piax93: +1

paasta_tools/cli/cmds/remote_run.py
Qmando: Nope

piax93 (dismissed):

PR #4099, opened 2025-07-21
Support adhoc instances in remote-run

Qmando:
People may want to run adhoc configured instances with remote-run as they often do with local-run. To do this, we can load the adhoc deployment configuration and put the config dict into a new EKS deployment object, similar to how toolbox containers function.

I tested this with a very basic adhoc config file that only specified memory and cpu. https://fluffy.yelpcorp.com/i/QcpxxNgXnTv4WXSfgDzr2qfSDNhHn5pK.html

### DIFF
--- a/paasta_tools/kubernetes/remote_run.py
+++ b/paasta_tools/kubernetes/remote_run.py
@@ -80,8 +80,13 @@ def _format_remote_run_job_name(
 
 
 def load_eks_or_adhoc_deployment_config(
-    service: str, instance: str, cluster: str, is_toolbox: bool = False, user: str = ""
+    service: str,
+    instance: str,
+    cluster: str,
+    is_toolbox: bool = False,
+    user: Optional[str] = None,
 ) -> EksDeploymentConfig:
+    assert user or not is_toolbox, "User required for toolbox deployment"
     try:
         deployment_config = (
             generate_toolbox_deployment(service, cluster, user)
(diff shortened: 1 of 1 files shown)


piax93 (dismissed):
looks reasonable

paasta_tools/kubernetes/remote_run.py
```diff
@@ -79,6 +79,29 @@ def _format_remote_run_job_name(
     return limit_size_with_hash(f"remote-run-{user}-{job.metadata.name}")
 
 
+def load_eks_or_adhoc_deployment_config(
+    service: str, instance: str, cluster: str, is_toolbox: bool = False, user: str = ""
```
nemacysts: is it actually valid to pass through a "" user to `generate_toolbox_deployment()`?

perhaps we should default to `None` and error if is_toolbox is `True`? or maybe get rid of is_toolbox and instead `s/user/toolbox_user` + default that to None (and essentially treat None as False and any other string as True?

nemacysts: or alternatively, i guess
```python
   deployment_config = (
        generate_toolbox_deployment(service, cluster, user)
        if is_toolbox
        else load_eks_or_adhoc_service_config(service, instance, cluster)
    )
```
would sidestep this :p

(although, i see why we did this - the current construct is repeated multiple times)

Qmando: True, None is probably slightly better, and I can throw in an `assert user or not is_toolbox`

PR #4101, opened 2025-08-01
remote-run: some minor fixes

piax93:
* When remote-running existing service instances we get lucky that setup-kubernetes-job pretty much in all cases previously set up a service account when there's an IAM role associated to the service instance. We don't have the same luxury with toolbox instances. I didn't go for a conditional logic though, since it does not hurt to double check that the service account we are trying to associate to the pod actually exists.
* When doing interactive remote-runs, the pod will often not have a listening processes running in it, so the liveness probe associated to long-running instances would eventually kill it, and we don't want that. Here again I didn't make anything conditional, since even when launching non-interactive instances, we don't care if they are alive or not, as they are not supposed to serve traffic, but we definitely don't want to have k8s making them flap.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -1461,6 +1461,7 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         service_namespace_config: ServiceNamespaceConfig,
         include_sidecars: bool = True,
         include_liveness_probe: bool = True,
+        include_readiness_probe: bool = True,
     ) -> Sequence[V1Container]:
         ports = [self.get_container_port()]
         # MONK-1130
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -1485,7 +1486,11 @@ def get_kubernetes_containers(
                 )
             ),
             name=self.get_sanitised_instance_name(),
-            liveness_probe=self.get_liveness_probe(service_namespace_config),
+            liveness_probe=(
+                self.get_liveness_probe(service_namespace_config)
+                if include_liveness_probe
+                else None
+            ),
             readiness_probe=self.get_readiness_probe(service_namespace_config),
```
nemacysts: we probably also don't want to include the readiness probe, right?

piax93: I have some pretty incredible tunnel vision. But yeah, probably.

paasta_tools/kubernetes_tools.py
```diff
@@ -1460,6 +1460,8 @@ def get_kubernetes_containers(
         secret_volumes: Sequence[SecretVolume],
         service_namespace_config: ServiceNamespaceConfig,
         include_sidecars: bool = True,
+        include_liveness_probe: bool = True,
+        include_readiness_probe: bool = True,
```
nemacysts, line 1464: i'd consider making this a single `include_probes` since it's very likely that we only ever want none or all of 'em - but this is a pretty easy refactor down the line so it's fine :)

nemacysts (approved):

paasta_tools/kubernetes_tools.py
piax93: awwwwwwww man, I was going to do that, but then thought "if I do this next review is going to be 'what if we just want to turn one off'", so I did it with separate params 🤣

PR #4105, opened 2025-08-08
Retry more intelligently when getting hosts in mesh_status

nemacysts:
We do currently have some request-level retries using a Session() if I'm reading things correctly, but that will retry on the same host - which generally won't be too helpful for hosts that are broken/dying/gone.

This is somewhat funky-looking, but without breaking this out into another function and using something like the `retry` library (or a similar artisanal decorator), this is probably good enough for now

I tested this on the current eksstage leader by installing a custom .deb -> injecting some code that would often ConnectTimeout -> curled the local API -> saw this retry :)

NOTE: tests were written with `claude` since testing this function is otherwise not much fun

### DIFF
--- a/tests/instance/test_kubernetes.py
+++ b/tests/instance/test_kubernetes.py
@@ -18,6 +18,7 @@ import a_sync
 import asynctest
 import mock
 import pytest
+import requests.exceptions
 
 import paasta_tools.instance.kubernetes as pik
 from paasta_tools import utils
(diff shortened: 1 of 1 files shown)


paasta_tools/instance/kubernetes.py
```diff
@@ -347,31 +351,49 @@ async def mesh_status(
...
-                    envoy_admin_endpoint_format=settings.system_paasta_config.get_envoy_admin_endpoint_format(),
-                    registration=registration,
-                    pods=pods,
-                    location=location,
-                    should_return_individual_backends=should_return_individual_backends,
-                )
-            )
+        max_retries = 3
+
+        for attempt in range(3):
```
jfongatyelp: extremely silly nit, maybe?
```suggestion
        for attempt in range(max_retries):
```

nemacysts: good catch - i dunno how i forgot to update this after introducing this variable lol

PR #4107, opened 2025-08-26
Document lack of boto_keys in local-run

nemacysts:
I initially just added this in the local-run --help output, but then I went
ahead and also updated the rST files.

As part of this, I noticed that `iam_role` wasn't documented, and added
an entry for that as well.

There's probably some additional keys that we should add here, but these
are the first two that came to mind.

I've also gone ahead and replaced the kubernetes-[clustername].yaml
heading with eks-

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -552,6 +552,7 @@ instance MAY have:
 
   * ``boto_keys``: A list of boto_cfg keys to be made available in the container's environment. Use of this option is discouraged in favor of using IAM roles, but it is still necessary for some legacy services.
     If specified, only the listed boto_cfg keys will be made available in the container's environment. Otherwise, the status-quo behavior will be used.
+    See `y/users-to-role <http://y/users-to-role>`_ for more information on how to switch to IAM roles.
     NOTE: this key is ignored for ``paasta local-run``.
 
 **Note**: Although many of these settings are inherited from ``smartstack.yaml``,


docs/source/yelpsoa_configs.rst
```diff
@@ -548,6 +548,12 @@ instance MAY have:
   * ``enable_aws_lb_readiness_gate``: A boolean indicating whether to enable injecting AWS Load Balancer readiness gates for this instance. Defaults to ``false``.
     This is only applicable to a very small subset of services at Yelp - please chat with #paasta before enabling this.
 
+  * ``iam_role``: The IAM role to be used by this service. If not specified, no IAM role will be used for Pod Identity.
+
+  * ``boto_keys``: A list of boto_cfg keys to be made available in the container's environment. Use of this option is discouraged in favor of using IAM roles, but it is still necessary for some legacy services.
+    If specified, only the listed boto_cfg keys will be made available in the container's environment. Otherwise, the status-quo behavior will be used.
```
EvanKrall, line 554: might be nice to mention what the status quo is

EvanKrall (dismissed):

docs/source/yelpsoa_configs.rst
nemacysts, line 554:
```suggestion
    If specified, only the listed boto_cfg keys will be made available in the container's environment. Otherwise, the status-quo behavior will be used.
    See `y/users-to-role <http://y/users-to-role>`_ for more information on how to switch to IAM roles.
```

nemacysts: (chatted in slack, just gonna link to the rb for migrating to iam roles for now)

EvanKrall (approved):

PR #4110, opened 2025-08-29
Validate smartstack.yaml in `paasta validate`

nemacysts:
We currently do this in an internal repo (`py-gitolite`), but I don't see why we don't just do this in `paasta validate` - especially since moving this check left means that folks are alerted to broken config earlier rather than when they attempt to integrate.

The smartstack_schema.json in this repo was behind the one in py-gitolite, so I've gone ahead and made them match.

The majority of the added code (outside the tests, which I had Claude Code regenerate for me) is more-or-less taken straight from py-gitolite with minimal changes.

I ran `paasta validate` over all soaconfigs and verified that there are no new failures when run with this code.

### DIFF
--- a/tests/cli/test_cmds_validate.py
+++ b/tests/cli/test_cmds_validate.py
@@ -1648,7 +1648,7 @@ def test_check_advertise_discover_rhs_less_general():
 
 def test_check_smartstack_proxied_through_no_proxied_through():
     smartstack_data = {}
-    _check_smartstack_proxied_through(smartstack_data)
+    _check_smartstack_proxied_through(smartstack_data, "/fake/soa/dir")
 
 
 def test_check_smartstack_proxied_through_valid():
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/schemas/smartstack_schema.json
```diff
@@ -117,6 +122,7 @@
             "mode": {
                 "enum": [
                     "http",
+                    "http2",
```
jfongatyelp: Do we need http2 here twice? (line 127)

nemacysts: huh, how did that happen

nemacysts:
```suggestion
```

paasta_tools/cli/cmds/validate.py
nemacysts, line 1: note: a bunch of this code comes straight from py-gitolite (thus the mix of string formatting methods and other fun shenanigans)

paasta_tools/cli/schemas/smartstack_schema.json
nemacysts, line 1: changes are copied over from the internal schema

PR #4111, opened 2025-08-29
Run `mypy` as part of tests

nemacysts:
it's always bothered me that these have to be run explicitly with `tox -e mypy` OR push to GH.

Additionally, I needed to upgrade mypy to use the "new" list[...]/dict[..., ...]/etc type syntax.

And finally: I deleted some Mesos-related code that the new mypy was flagging since I didn't want to deal with fixing dead code :)


EvanKrall (approved):

paasta_tools/metrics/metrics_lib.py
```diff
@@ -163,7 +172,7 @@ def set(self, value: Union[int, float]) -> None:
         log.debug(f"gauge {self.name} set to {value}")
 
 
-class Counter(GaugeProtocol):
+class Counter(CounterProtocol):
```
jfongatyelp, line 175: ! lol

jfongatyelp (approved):

tests/cli/test_cmds_status.py
```diff
@@ -2264,14 +2264,14 @@ def test_output(
...
+            "      Pods:",
+            "        Pod ID  Host deployed to  Deployed at what localtime      Health",
             f"        app_1   fake_host1        2019-07-12T13:31 ({mock_naturaltime.return_value})  {PaastaColors.green('Healthy')}",
             f"        app_2   fake_host2        2019-07-12T13:31 ({mock_naturaltime.return_value})  {PaastaColors.green('Healthy')}",
             f"        app_3   fake_host3        2019-07-12T13:31 ({mock_naturaltime.return_value})  {PaastaColors.red('Evicted')}",
             f"        {PaastaColors.grey('  Disk quota exceeded')}",
-            f"      ReplicaSets:",
-            f"        ReplicaSet Name  Ready / Desired  Created at what localtime       Service git SHA  Config hash",
+            "      ReplicaSets:",
+            "        ReplicaSet Name  Ready / Desired  Created at what localtime       Service git SHA  Config hash",
```
EvanKrall, line 2274: really wish black would be less strict about not using `f"` for strings without interpolations, so that alignment would be clearer

nemacysts: this is actually flake8, if i'm not mistaken

i think we can maybe add some targeted `# noqa: WHATEVERTHECODEIS` to these lines?

EvanKrall: if it has to go on every line, it might not be worth it ¯\_(ツ)_/¯

nemacysts: yea, doesn't look like there's a way to noqa a block/section of code :(

EvanKrall (approved):

jfongatyelp (approved):

PR #4115, opened 2025-09-12
 Remove gunicorn sidecar

ymilki:
We are no longer using the sidecar to provide metrics for prometheus.


jvperrin (dismissed):

paasta_tools/kubernetes_tools.py
```diff
@@ -1168,37 +1161,6 @@ def get_hacheck_sidecar_container(
...
         return None
 
-    def get_gunicorn_exporter_sidecar_container(
-        self,
-        system_paasta_config: SystemPaastaConfig,
-    ) -> Optional[V1Container]:
-
-        if self.should_use_metrics_provider(METRICS_PROVIDER_GUNICORN):
-            return V1Container(
-                image=system_paasta_config.get_gunicorn_exporter_sidecar_image_url(),
```
jtwang, line 1178: Do we need to delete this method defn too? It's in utils.py

jtwang (dismissed):

paasta_tools/kubernetes_tools.py
ymilki: Done!

PR #4116, opened 2025-09-18
Add pod labels for all configured metrics providers

ymilki:
## Problem

Each autoscaling metric may have some specific pod labels to identify the pod for the scraper. In the case of multiple metrics, we'll need to add all the pod labels. Currently, we only provide the labels from one metric.

## Solution

Change if/else logic to if/if.

## Validation

Add tests.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2489,11 +2489,11 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             # character limit for k8s labels (63 chars)
             labels["paasta.yelp.com/deploy_group"] = self.get_deploy_group()
 
-        elif self.should_use_metrics_provider(METRICS_PROVIDER_PISCINA):
+        if self.should_use_metrics_provider(METRICS_PROVIDER_PISCINA):
             labels["paasta.yelp.com/deploy_group"] = self.get_deploy_group()
             labels["paasta.yelp.com/scrape_piscina_prometheus"] = "true"
 
-        elif self.should_use_metrics_provider(METRICS_PROVIDER_GUNICORN):
+        if self.should_use_metrics_provider(METRICS_PROVIDER_GUNICORN):
             labels["paasta.yelp.com/deploy_group"] = self.get_deploy_group()
             labels["paasta.yelp.com/scrape_gunicorn_prometheus"] = "true"
 
--- a/tests/test_kubernetes_tools.py
+++ b/tests/test_kubernetes_tools.py
@@ -1897,6 +1897,218 @@ class TestKubernetesDeploymentConfig:
         assert ret == expected
 
     @mock.patch(
+        "paasta_tools.kubernetes_tools.load_service_namespace_config", autospec=True
+    )
+    @mock.patch(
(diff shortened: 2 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -2489,11 +2489,11 @@ def get_pod_template_spec(
             # character limit for k8s labels (63 chars)
             labels["paasta.yelp.com/deploy_group"] = self.get_deploy_group()
 
-        elif self.should_use_metrics_provider(METRICS_PROVIDER_PISCINA):
+        if self.should_use_metrics_provider(METRICS_PROVIDER_PISCINA):
             labels["paasta.yelp.com/deploy_group"] = self.get_deploy_group()
```
jfongatyelp, line 2459: We're now repeating this deploy_group label setting, but i guess no real harm :p

jfongatyelp (dismissed):

tests/test_kubernetes_tools.py
```diff
@@ -1896,6 +1899,228 @@ def test_get_pod_template_spec(
...
+        mock_get_pod_anti_affinity.return_value = None
+        mock_create_pod_topology_spread_constraints.return_value = []
+        mock_get_termination_grace_period.return_value = None
+
+        mock_system_paasta_config = mock.Mock()
+        mock_system_paasta_config.get_kubernetes_add_registration_labels.return_value = (
+            True
+        )
+        mock_system_paasta_config.get_topology_spread_constraints.return_value = []
+        mock_system_paasta_config.get_pod_defaults.return_value = dict(dns_policy="foo")
```
EvanKrall:
```suggestion
        mock_system_paasta_config.get_pod_defaults.return_value = {"dns_policy": "foo"}
```

jvperrin (dismissed):

EvanKrall (dismissed):

tests/test_kubernetes_tools.py
```diff
@@ -1896,6 +1899,228 @@ def test_get_pod_template_spec(
...
+        expected_labels: Dict[str, str],
+    ):
+        """Test get_pod_template_spec with multiple metrics providers in all combinations."""
+        mock_service_namespace_config = mock.Mock()
+        mock_load_service_namespace_config.return_value = mock_service_namespace_config
+        mock_service_namespace_config.is_in_smartstack.return_value = True
+        mock_get_node_affinity.return_value = None
+        mock_get_pod_anti_affinity.return_value = None
+        mock_create_pod_topology_spread_constraints.return_value = []
+        mock_get_termination_grace_period.return_value = None
```
nemacysts, line 2067: at some point i'll get an LLM to mass-migrate us from  decorator-based mocking + setting the mock return values inside the function to doing this with context-manager-based mocking - but there's other tests doing the exact same thing as this so doing the same here is fine for now :)

paasta_tools/kubernetes_tools.py
nemacysts: !8ball we should just add this as a default label

nemacysts (approved):

jfongatyelp (dismissed):

jfongatyelp (approved):

PR #4117, opened 2025-09-18
Update gunicorn autoscaling metric

ymilki:
## Feature

The gunicorn autoscaler functions the same way as the uwsgi autoscaler. The uwsgi_v2 (#4012) improves upon the calculations and we want to update the gunicorn autoscaler to follow.

## Solution

Update the gunicorn autoscaler policy and metric to follow that of the uwsgi_v2 autoscaler policy.

## Validation

Updated tests


ymilki:
We will abandon the gunicorn autoscaler in favor of #4120

PR #4118, opened 2025-09-23
Using PAASTA_PORT instead of hard-coded one in status

ilkinmammadzada:
Currently we are using hard-coded 8888 in paasta status.

The code now retrieves the port value directly from the container's environment variables, defaulting to "8888" if "PAASTA_PORT" is not set.

While it is possible to use container.ports.containerPort to determine the port, I find it easier and more consistent to use the "PAASTA_PORT" environment variable. This approach also aligns with how PaaSTA exposes the "container_port" from the yelpsoa config.

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -72,6 +72,7 @@ from paasta_tools.kubernetes_tools import format_tail_lines_for_kubernetes_pod
 from paasta_tools.kubernetes_tools import KubernetesDeploymentConfig
 from paasta_tools.kubernetes_tools import KubernetesDeployStatus
 from paasta_tools.kubernetes_tools import paasta_prefixed
+from paasta_tools.long_running_service_tools import DEFAULT_CONTAINER_PORT
 from paasta_tools.monitoring_tools import get_runbook
 from paasta_tools.monitoring_tools import get_team
 from paasta_tools.monitoring_tools import list_teams
@@ -1402,6 +1403,12 @@ def get_main_container(pod: KubernetesPodV2) -> Optional[KubernetesContainerV2]:
     )
 
 
+def get_container_env(container: Optional[KubernetesContainerV2]) -> Dict[str, str]:
+    if not container or not container.env:
+        return {}
+    return {env_var.name: env_var.value for env_var in container.env}
+
+
 def get_replica_state(pod: KubernetesPodV2) -> ReplicaState:
     phase = pod.phase
     state = ReplicaState.UNKNOWN
@@ -1474,9 +1481,14 @@ def create_replica_table(
     for state, pod in pods:
         start_datetime = datetime.fromtimestamp(pod.create_timestamp)
         humanized_start_time = humanize.naturaltime(start_datetime)
+        # Get port from the main container's PAASTA_PORT env var, fallback to DEFAULT_CONTAINER_PORT
+        main_container = get_main_container(pod)
+        container_env = get_container_env(main_container)
+        paasta_port = container_env.get("PAASTA_PORT", str(DEFAULT_CONTAINER_PORT))
+
         row = [
             pod.name,
-            f"{pod.ip}:8888" if pod.ip else "None",
+            f"{pod.ip}:{paasta_port}" if pod.ip else "None",
             pod.host or "None",
             humanized_start_time,
             state.formatted_message,
@@ -1491,7 +1503,6 @@ def create_replica_table(
                 )
             )
 
-        main_container = get_main_container(pod)
         if main_container:
             if main_container.timestamp:
                 timestamp = datetime.fromtimestamp(main_container.timestamp)


ajayOO8 (approved):

paasta_tools/cli/cmds/status.py
```diff
@@ -1474,9 +1474,17 @@ def create_replica_table(
     for state, pod in pods:
         start_datetime = datetime.fromtimestamp(pod.create_timestamp)
         humanized_start_time = humanize.naturaltime(start_datetime)
+        # Get port from first container's PAASTA_PORT env var, fallback to 8888
+        container = pod.containers[0]
```
giuli007: is it always the first container? is there any chance it might not be?

ilkinmammadzada: yup yup, it will be the first: https://github.com/Yelp/paasta/blob/346749bbd9b7a7e2bd4fa8c6518ed4a3bd0dc1ae/paasta_tools/kubernetes_tools.py#L1476

Rob-Johnson: @ilkinmammadzada hmm it feels iffy to depend on finding the 'right' container by an ordering.

Can we fetch this from the yelpsoa-configs for the instance instead?

ilkinmammadzada: I found something else: `get_main_container` function which does what I really need. What do you think?

ilkinmammadzada: I just pushed the commit @giuli007 @Rob-Johnson

paasta_tools/cli/cmds/status.py
```diff
@@ -1474,9 +1474,17 @@ def create_replica_table(
     for state, pod in pods:
         start_datetime = datetime.fromtimestamp(pod.create_timestamp)
         humanized_start_time = humanize.naturaltime(start_datetime)
+        # Get port from the main container's PAASTA_PORT env var, fallback to 8888
+        main_container = get_main_container(pod)
+        env_list = getattr(main_container, "env", [])
+        paasta_port = next(
```
giuli007: non-blocking observation: I find this line using `next` with a tuple comprehension, which would result in a single element at most, a bit inscrutable but probably just because I'm a bit rusty. no need to change it

giuli007 (dismissed):

paasta_tools/cli/cmds/status.py
ilkinmammadzada: I can convert the list to dict as an alternative:

```
env_list = getattr(main_container, "env", [])
env_dict = {env_var.name: env_var.value for env_var in env_list}
paasta_port = env_dict.get("PAASTA_PORT", "8888")
```

giuli007: yeah there are many other ways to do it. the simplest I thought of is
```python
paasta_port = 8888
for env_var in env_list:
    if env_var.name == "PAASTA_PORT":
        paasta_port = env_var.value
        break
```

but really I don't think it's strictly necessary to change it

ilkinmammadzada: Yeah I definitely agree that is the simplest option.

But this method (dict) avoids explicit iteration and makes the intent obvious: retrieve "PAASTA_PORT" if it exists, otherwise use "8888". While a for+if loop would also work, the dictionary approach is more direct for single, unique lookups and is considered more idiomatic I think

paasta_tools/cli/cmds/status.py
```diff
@@ -1474,9 +1474,15 @@ def create_replica_table(
     for state, pod in pods:
         start_datetime = datetime.fromtimestamp(pod.create_timestamp)
         humanized_start_time = humanize.naturaltime(start_datetime)
+        # Get port from the main container's PAASTA_PORT env var, fallback to 8888
+        main_container = get_main_container(pod)
+        env_list = getattr(main_container, "env", None) or []
```
nemacysts: do we need to use getattr here? shouldn't `env` always be present?

paasta_tools/cli/cmds/status.py
```diff
@@ -1474,9 +1474,15 @@ def create_replica_table(
...
+        env_dict = {env_var.name: env_var.value for env_var in env_list}
+        paasta_port = env_dict.get("PAASTA_PORT", "8888")
```
EvanKrall: nit: could use DEFAULT_CONTAINER_PORT from long_running_service_tools, but ¯\\\_(ツ)\_/¯

EvanKrall (approved):

paasta_tools/cli/cmds/status.py
```diff
@@ -1474,9 +1475,15 @@ def create_replica_table(
     for state, pod in pods:
         start_datetime = datetime.fromtimestamp(pod.create_timestamp)
         humanized_start_time = humanize.naturaltime(start_datetime)
+        # Get port from the main container's PAASTA_PORT env var, fallback to 8888
+        main_container = get_main_container(pod)
+        env_list = getattr(main_container, "env", None) or []
```
nemacysts: might be nice to extract this to a function 

e.g., 

```python
def get_container_env(container: V1Container) -> dict[str, str]:
    env_vars = container.env or []
    return {env_var.name: env_var.value for env_var in env_vars}
```

and then write something like:
```python
main_container = get_main_container(pod)
env = get_container_env(main_container)
paasta_port = env.get("PAASTA_PORT", str(DEFAULT_CONTAINER_PORT))
```

so that we don't have variables with types in their names :p

paasta_tools/cli/cmds/status.py
ilkinmammadzada: updated :)

paasta_tools/cli/cmds/status.py
```diff
@@ -1474,9 +1475,15 @@ def create_replica_table(
     for state, pod in pods:
         start_datetime = datetime.fromtimestamp(pod.create_timestamp)
         humanized_start_time = humanize.naturaltime(start_datetime)
+        # Get port from the main container's PAASTA_PORT env var, fallback to DEFAULT_CONTAINER_PORT
+        main_container = get_main_container(pod)
+        env_list = getattr(main_container, "env", None) or []
+        env_dict = {env_var.name: env_var.value for env_var in env_list}
+        paasta_port = env_dict.get("PAASTA_PORT", DEFAULT_CONTAINER_PORT)
```
nemacysts: nit if we don't go with my other suggestions:
```suggestion
        paasta_port = env_dict.get("PAASTA_PORT", str(DEFAULT_CONTAINER_PORT))
```
would be nice so that `paasta_port` isn't sometimes a string and sometimes an int :)

paasta_tools/cli/cmds/status.py
```diff
@@ -1402,6 +1403,10 @@ def get_main_container(pod: KubernetesPodV2) -> Optional[KubernetesContainerV2]:
     )
 
 
+def get_container_env(container) -> dict[str, str]:
```
nemacysts: i assume container is a V1Container from the k8s library
```suggestion
def get_container_env(container: V1Container) -> dict[str, str]:
```

paasta_tools/cli/cmds/status.py
```diff
@@ -1474,9 +1479,14 @@ def create_replica_table(
     for state, pod in pods:
         start_datetime = datetime.fromtimestamp(pod.create_timestamp)
         humanized_start_time = humanize.naturaltime(start_datetime)
+        # Get port from the main container's PAASTA_PORT env var, fallback to DEFAULT_CONTAINER_PORT
+        main_container = get_main_container(pod)
```
nemacysts, line 1485: oh, i also just noticed that the code below checks if main_container is truthy - do we need to do the same thing here?

jfongatyelp (approved):
My first instinct would have been to get this from yelpsoa-configs as well so we'd only have to grab it once rather than for each pod, buuuuuut this is a clever way to minimize changes (eg no need to add the info from the server side or add port info to KubernetesPodV2 spec, plus no need to coordinate server+cli releases), plus doing it this way will help the (admittedly hopefully short) period where a ysoa change to PAASTA_PORT may not have applied to the running pods just yet, so sgtm!

EvanKrall:
> My first instinct would have been to get this from yelpsoa-configs as well so we'd only have to grab it once rather than for each pod, buuuuuut this is a clever way to minimize changes (eg no need to add the info from the server side or add port info to KubernetesPodV2 spec, plus no need to coordinate server+cli releases), plus doing it this way will help the (admittedly hopefully short) period where a ysoa change to PAASTA_PORT may not have applied to the running pods just yet, so sgtm!

I had the same thought. I like this approach better than using `get_container_port` for both of those reasons.

EvanKrall (approved):

paasta_tools/cli/cmds/status.py
ilkinmammadzada: I did this check in `get_container_env`, wouldn't that be enough?

PR #4120, opened 2025-09-24
worker-load autoscaler

ymilki:
## Feature

The `worker-load` autoscaler functions equivalently to the `uwsgi-v2` autoscaler except it looks at the `worker_busy` metric (instead of `uwsgi_worker_busy`). This is meant to be an autoscaler that can apply to any service that depends on worker utilization - particularly targetting both uwsgi and gunicorn services.

## Changes

* Add autoscaling rule
* Add to `metrics_providers` schema
* Add `worker-load` autoscaler


paasta_tools/kubernetes_tools.py
```diff
@@ -2445,22 +2449,29 @@ def get_pod_template_spec(
...
+            for provider in metrics_providers_needing_deploy_group
+        ):
+            # Deploy group is needed for Prometheus relabeling properly
+            # This should probably eventually be made into a default label,
             # but for now we're fine with it being behind these feature toggles.
-            # ideally, we'd also have the docker image here for ease-of-use
+            # Ideally, we'd also have the docker image here for ease-of-use
             # in Prometheus relabeling, but that information is over the
             # character limit for k8s labels (63 chars)
             labels["paasta.yelp.com/deploy_group"] = self.get_deploy_group()
```
nemacysts, line 2469: hmm, maybe i should just make a PR to ship on monday to add this as a default label - it'll require a big bounce, but that's fine :p


(that said, no action from you - if this gets merged before i make my change i can just refactor)

paasta_tools/long_running_service_tools.py
```diff
@@ -49,6 +50,7 @@
 METRICS_PROVIDER_PISCINA = "piscina"
 METRICS_PROVIDER_ACTIVE_REQUESTS = "active-requests"
 METRICS_PROVIDER_PROMQL = "arbitrary_promql"
+METRICS_PROVIDER_WORKER_LOAD = "worker_load"
```
nemacysts: we should probably try to be more consistent with the names here and have this be `worker-load`

(i can make a pr to rename the promql one since nothing is using that yet)

ymilki: done

paasta_tools/kubernetes_tools.py
nemacysts: (https://github.com/Yelp/paasta/pull/4123 - i'll probably end up merging mine after this)

paasta_tools/cli/schemas/autoscaling_schema.json
```diff
@@ -10,8 +10,10 @@
                     "cpu",
                     "piscina",
                     "gunicorn",
+                    "active-requests",
                     "arbitrary-promql",
-                    "active-requests"
+                    "active-requests",
```
nemacysts:
```suggestion
```

nemacysts: (this is duplicated now)

PR #4124, opened 2025-09-29
adding env field to KubernetesContainerV2

ilkinmammadzada:


paasta_tools/api/api_docs/oapi.yaml
```diff
@@ -673,6 +673,12 @@ components:
         name:
           description: Name of the container
           type: string
+        env:
+          description: Environment variables set in the container
+          type: array
+          items:
+            $ref: '#/components/schemas/KubernetesContainerEnvVarV2'
+          nullable: true
```
nemacysts, line 681: we should probably return a dictionary here so that every client doesn't need to do the array->dict dance :)

paasta_tools/instance/kubernetes.py
```diff
@@ -935,6 +936,8 @@ async def get_pod_containers(
                 healthcheck = get_container_healthcheck(
                     pod.status.pod_ip, spec.liveness_probe
                 )
+            if spec.env:
+                env_list = [{"name": e.name, "value": e.value} for e in spec.env]
```
nemacysts:
```suggestion
            env = {e.name: e.value for e in (spec.env or {})}
```

is probably the code-golfiest way to flatten the array into a dict - but we'll probably want to end up using an actual for-loop here since i assume that we probably want to filter out env var values that aren't primitives (i.e., filter out secret refs and whatnot)? or is `e.value` for something like a secret ref just print the name of the secret?

paasta_tools/api/api_docs/oapi.yaml
```diff
@@ -731,6 +737,16 @@ components:
           $ref: '#/components/schemas/TaskTailLines'
           description: Stdout and stderr tail of the container
       type: object
+    KubernetesContainerEnvVarV2:
```
nemacysts: this is probably 
```suggestion
    KubernetesContainerV2EnvVar:
```
but if we swap to just returning an object (i.e., dict) - then i guess we end up not needing this def :p

paasta_tools/api/api_docs/oapi.yaml
ilkinmammadzada: Yeah we had discussed this in the meeting, but then I thought that returning dict in api call doesn't seem correct. Instead we should provide better info about object (here envvar). Also it will make it easy in future if we want to support complex envvars

nemacysts: this def is only used by the status endpoint (if we follow the the container definitions - we can see that there's nothing that will contain this definition that is used outside of the status endpoint)

imo, we likely won't ever want to return complex env vars from the instance status endpoint and at most would probably want to replace the value for complex vars to something like 
```python
{
  "PAASTA_SECRET_FOO": "redacted from <source: secret/...>", 
  ...
}
```

nemacysts: (and sidenote: changing the paasta api isn't too hard - we don't really provide api stability guarantees for most endpoints since we're not really expecting folks to directly use the API most of the time: so it's pretty simple to make sweeping changes since we control both the API and the consumers of the API)

ilkinmammadzada: Is there "dictionary" def in openapi? 

I know we can use "object" like:

```
type: object
additionalProperties:
  type: string
```
but from this definition it isn't really clear that what will api return. 

I would prefer to show what API will return which is better practice imho. 

I'm thinking about to remove list to dict conversion method and instead provide a way to get env variable value by key.

nemacysts: we could always add a description field :)

alternatively, maybe `patternProperties` would get closer to an in-between?

e.g.,
```
type: object
additionalProperties: false
patternProperties:
  "^[a-zA-Z0-9_-]+$":
    type: string
```
i.e., string key (note: my understanding is that from the swagger/oapi side of things, keys are always values) and string value

ilkinmammadzada: I just changed the client side code to handle list without needing dict. 

I agree we can use the options you provided but I'm not seeing the reason to not use standard practice. 
If you don't mind and don't have strong opinion I wanted to keep it clear in api side - using EnvVar array :)

nemacysts: i'm not a huge fan of an env var array here since it complicates client-side code, will be somewhat surprising to consumers, and we're almost certainly not gonna need the flexibility of an env var array

but: since paasta ownership is moving to y'all, i don't mind deferring to what y'all wanna do in cases where nothing will blow up either way :)

paasta_tools/cli/cmds/status.py
```diff
@@ -1403,10 +1403,12 @@ def get_main_container(pod: KubernetesPodV2) -> Optional[KubernetesContainerV2]:
...
-def get_container_env(container: Optional[KubernetesContainerV2]) -> Dict[str, str]:
-    if not container or not container.env:
-        return {}
-    return {env_var.name: env_var.value for env_var in container.env}
+def get_container_env(container: Optional[KubernetesContainerV2], env_name: str) -> str:
+    if container and container.env:
+        for env_var in container.env:
+            if env_var.name == env_name:
+                return str(env_var.value)
+    return ""
```
nemacysts:
```suggestion
def get_container_env(container: KubernetesContainerV2 | None, env_name: str) -> str | None:
    if container and container.env:
        for env_var in container.env:
            if env_var.name == env_name:
                return str(env_var.value)
    return None
```

we probably don't want to return `""` if we didn't find the env var since that's a valid env var value

that said: we could maybe also do something like:


```suggestion
def get_container_env(container: KubernetesContainerV2 | None, env_name: str, default=None: Any) -> str | Any:  # correctly typing this is left as an exercise for the reader
    if container and container.env:
        for env_var in container.env:
            if env_var.name == env_name:
                return str(env_var.value)
    return default
```

(NOTE: we can use the new union syntax since we're on a more modern python now - but we can also stick with the old syntax :p)

paasta_tools/cli/cmds/status.py
```diff
@@ -1403,10 +1403,12 @@ def get_main_container(pod: KubernetesPodV2) -> Optional[KubernetesContainerV2]:
...
 
-def get_container_env(container: Optional[KubernetesContainerV2]) -> Dict[str, str]:
-    if not container or not container.env:
-        return {}
-    return {env_var.name: env_var.value for env_var in container.env}
+def get_container_env(container: Optional[KubernetesContainerV2], env_name: str) -> str:
+    if container and container.env:
+        for env_var in container.env:
+            if env_var.name == env_name:
+                return str(env_var.value)
```
nemacysts: this should already be a string, no?

paasta_tools/cli/cmds/status.py
ilkinmammadzada: Why do we want to change Optional[KubernetesContainerV2] to KubernetesContainerV2 | None? 

I just tried to use the same logic as we used in other functions :)

ilkinmammadzada: > that's a valid env var value

good point! going to change

nemacysts: just to preempt a future pre-commit that updates the syntax everywhere :p 

(we don't need to, tho)

PR #4125, opened 2025-09-29
[worker-load autoscaler] Add missing validation config

ymilki:
## Problem

`paasta validate` is failing with the `worker-load` autoscaler because of a missing key

```
  File "/opt/venvs/paasta-tools/lib/python3.10/site-packages/paasta_tools/cli/cmds/validate.py", line 682, in _validate_autoscaling_config
    for field in INVALID_AUTOSCALING_FIELDS[metrics_provider_type]:
KeyError: 'worker-load'
```

## Solution

Duplicate the `uwsgi-v2` configs for invalid autoscaling fields


paasta_tools/cli/cmds/validate.py
```diff
@@ -163,6 +164,10 @@ class AutoscalingValidationError(SoaValidationError):
         "desired_active_requests_per_replica",
         "prometheus-adapter-config",
     },
+    METRICS_PROVIDER_WORKER_LOAD: {
```
nemacysts, line 167: we should probably refactor how this works since this is pretty confusing, but that's not on you

PR #4127, opened 2025-09-29
adding env field to KubernetesContainerV2

ilkinmammadzada:


paasta_tools/api/api_docs/oapi.yaml
```diff
@@ -673,6 +673,12 @@ components:
         name:
           description: Name of the container
           type: string
+        env:
+          description: Environment variables set in the container
+          type: array
+          items:
+            $ref: '#/components/schemas/KubernetesContainerV2EnvVar'
+          nullable: true
```
nemacysts, line 681: previous discussion: https://github.com/Yelp/paasta/pull/4124/files#r2388149263

paasta_tools/instance/kubernetes.py
```diff
@@ -935,6 +936,8 @@ async def get_pod_containers(
                 healthcheck = get_container_healthcheck(
                     pod.status.pod_ip, spec.liveness_probe
                 )
+            if spec.env:
+                env = [{"name": e.name, "value": e.value} for e in spec.env]
```
nemacysts: what do we get here if `e` has a Secret source?

ilkinmammadzada: I can ignore those cases with the following code, if it sounds okay:

env = [
                    {"name": e.name, "value": e.value}
                    for e in spec.env
                    if e.value_from is None
                ]

nemacysts: copying this over from slack: what if we just pass down the PAASTA_PORT value directly?

we could probably add a key in the status response for instance metadata to support future enhancements (e.g., I was chatting with krall and one thing that might be nice to show in the future would be the number of workers per-replica andwhatnot)

e.g., 
```
"metadata": {
    "paasta_port": 1234,
    "num_workers": 5678,
    ...
}
```

then we don't need to worry about filtering or transforming from a list[dict[str, str] and we get smaller response sizes to boot :)

PR #4278, opened 2026-03-27
Validate that clb_proxy points to a valid smartstack namespace. …

EvanKrall:
…MESH-2009

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -1216,15 +1216,27 @@ def _check_smartstack_valid_proxy(proxied_through: str, soa_dir: str) -> None:
 
 
 def _check_smartstack_proxied_through(
-    smartstack_data: dict[str, Any],
+    smartstack_data: dict[str, Any],  # XXX: we should use a TypedDict here
     soa_dir: str,
-) -> None:  # XXX: we should use a TypedDict here
+) -> None:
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -1227,6 +1227,18 @@ def _check_smartstack_proxied_through(
     _check_smartstack_valid_proxy(proxied_through, soa_dir)
 
 
+def _check_smartstack_clb_proxy(
+    smartstack_data: dict[str, Any],
+    soa_dir: str,
+) -> None:  # XXX: we should use a TypedDict here
```
```suggestion
def _check_smartstack_clb_proxy(
    smartstack_data: dict[str, Any],  # XXX: we should use a TypedDict here
    soa_dir: str,
) -> None:
```

nemacysts: (although, might be too long and might need to go above this line instead)

nemacysts (dismissed):
minor nit that you can ignore since i think it's also pretty clear from context where we'd want the TypedDict

(also, i'm assuming we ran `paasta validate` over all of the existing soaconfigs and it was gucci)

nemacysts (dismissed):
i also realized that we could probably update the mocks a bit in the new tests by mocking `get_config_file_dict` instead of `_check_smartstack_valid_proxy`, but up to you if you want to update the mocks (my bad for missing this in the original review)

nemacysts (approved):

ilkinmammadzada (approved):

PR #4288, opened 2026-04-20
Re-enable SLO-based autorollbacks

nemacysts:
This is a tiny bit silly :p

SLO-based autorollbacks have been silently broken since Yelp/paasta#4065 - python-jsonschema-objects==0.3.1 tries to import some types from `collections`, but some of those types were moved to `collections.abc` in 3.10.

However...as we do a blanket try-except with no logging - we totally missed this breakage as the only way to see this is to first notice that these SLO rollbacks are missing (usage of these is pretty spotty given some known issues) and then either hook up a debugger or point an LLM to mark-for-deployment to debug as this is otherwise a completely silent failure :p

I've gone ahead and bumped python-jsonschema-objects to a version compatible with 3.10 and added a somewhat silly test to ensure that slo_transcoder is loadable - this test is skipped if slo_transcoder is not installed (i.e., if you're running the tests outside of Yelp)

Once I bumped this package, I was able to see autorollbacks working for my test service as expected :D

### DIFF
--- a/tests/cli/test_cmds_mark_for_deployment.py
+++ b/tests/cli/test_cmds_mark_for_deployment.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import asyncio
+import importlib.metadata
 from unittest import mock
 from unittest.mock import ANY
 from unittest.mock import AsyncMock
@@ -19,6 +20,7 @@ from unittest.mock import MagicMock
 from unittest.mock import call
 from unittest.mock import patch
 
+import pytest
 from pytest import fixture
 from pytest import raises
 from slackclient import SlackClient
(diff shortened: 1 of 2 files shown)


jfongatyelp (approved):

giuli007 (approved):
🚢

giuli007:
> we do a blanket try-except with no logging 

non-blocking: anything we could/should improve regarding to this (as a TODO)?

ilkinmammadzada (approved):

nemacysts:
> > we do a blanket try-except with no logging
> 
> non-blocking: anything we could/should improve regarding to this (as a TODO)?

@giuli007 yea, this specific case is now handled with my silly new test - but we should probably audit sticht + paasta (and honestly, probably Tron as well) and make sure that all the places where we catch ImportError to allow non-Yelpers (or GHA if we're being real :p) to run our OSS code have similar tests OR log that expected functionality is being skipped

PR #4290, opened 2026-04-23
AIRFLOW-3: Add a new extra-namespaces option to paasta secrets to support secret syncing without adding paasta instances

jfongatyelp:
## Summary

  Adds support for syncing PaaSTA secrets to Kubernetes namespaces that have no PaaSTA instances (e.g. MWAA workloads). Previously the only way to get a secret into a namespace was to have a PaaSTA instance running there; now secrets can declare target namespaces directly in their JSON file.

  **Two new options:**

  - `paasta secret add/update --extra-namespaces <ns1,ns2>`: writes an `extra_namespaces` field into the secret's JSON file. Pass an empty string to clear it. Consistent with the existing `--clusters` comma-separated convention eg.
    - Note that this does sort of break conventions of only touching the json secret files thru vault_tools -- I think the more 'proper' path would be to have vault_tools support some sort of arbitrary metadata we could pass along, but for now, i think this achieves what we want
    - Just slightly clunky interface -- every time they use `paasta secret update` with `--extra-namespaces`, they need to list the *full* extra namespaces list -- it doesn't append. In addition omitting the arg doesn't delete existing extra-namespaces; user must explicitly call `paasta secret update `with --extra-namespaces ''` to unset existing ones.  This was the compromise to minimize complexity on the cli side, but happy to try something idfferent if you think it works better!
  - `paasta_secrets_sync --only-extra-namespaces`: instead of deriving target namespaces from instance configs, builds the namespace→allowlist map solely from each secret's `extra_namespaces` field. Intended for services with no PaaSTA instances. 
    - Intended invocation could be to pipe _every_ service (+_shared) to paasta_secrets_sync --only-extra-namespaces, rather than only those with particular instance types as with all our other paasta secret syncs, or if we wanted more efficiency, do some bash filtering to get only servicves w/ at least 1 secret file with an `extra_namespaces` field 🤷 

  **Implementation notes:**

  - `_update_extra_namespaces` patches the `extra_namespaces` field into an already-written secret JSON file. `vault_tools` owns the normal write path for the ciphertext; this function runs after `write_secret` returns. Uses `atomic_file_write` (write-to-temp-then-rename) consistent with `write_json_configuration_file` / `write_yaml_configuration_file` in `utils.py`, tho it wasn't originally intended to work on jsonsecret files.
  - `get_services_to_k8s_namespaces_from_extra_namespaces` returns the same `Dict[service, Dict[namespace, Optional[Set[str]]]]` shape as `get_services_to_k8s_namespaces_to_allowlist`, so `sync_all_secrets` consumes both paths unchanged. Each secret with `extra_namespaces` is added to the allowlist for each declared namespace; secrets without the field are skipped entirely.

Note that this feature would not only satisfy our AIRFLOW plans, but also help a handful of other usecases that currently have to add 'dummy' instances to get a secret synced somewhere - eg temporal secret_placeholder or cassandrar shared secrets\

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -209,6 +209,7 @@ def _add_and_update_args(parser: argparse.ArgumentParser):
     parser.add_argument(
         "--extra-namespaces",
         default=None,
+        type=lambda v: v.split(",") if v else [],
         help=(
             "A comma-separated list of additional Kubernetes namespaces to sync this secret to, "
             "beyond those derived from PaaSTA instance configs. "
@@ -401,12 +402,11 @@ def _get_secret_provider_for_service(
     )
 
 
-def _update_extra_namespaces(secret_path: str, extra_namespaces_arg: str) -> None:
+def _update_extra_namespaces(secret_path: str, namespaces: List[str]) -> None:
     """Update the extra_namespaces field in a secret JSON file in-place."""
     # TODO: Add metadata support to secret_provider impl (vault_tools) instead
     # vault_tools owns the normal write path for secret JSON files; this patches
     # only the extra_namespaces metadata field after the ciphertext has been written.
-    namespaces: List[str] = [n for n in extra_namespaces_arg.split(",") if n]
     with open(secret_path, "r") as f:
         data = json.load(f)
     if namespaces:
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -403,6 +403,7 @@ def _get_secret_provider_for_service(
 
 def _update_extra_namespaces(secret_path: str, extra_namespaces_arg: str) -> None:
     """Update the extra_namespaces field in a secret JSON file in-place."""
+    # TODO: Add metadata support to secret_provider impl (vault_tools) instead
     # vault_tools owns the normal write path for secret JSON files; this patches
     # only the extra_namespaces metadata field after the ciphertext has been written.
     namespaces: List[str] = [n for n in extra_namespaces_arg.split(",") if n]
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -13,11 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import argparse
+import json
 import os
 import re
 import sys
 from typing import Any
 from typing import Dict
+from typing import List
 from typing import Optional
 
 from service_configuration_lib import DEFAULT_SOA_DIR
@@ -36,6 +38,7 @@ from paasta_tools.secret_tools import SHARED_SECRET_SERVICE
 from paasta_tools.secret_tools import decrypt_secret_environment_variables
 from paasta_tools.secret_tools import get_secret_provider
 from paasta_tools.utils import _log_audit
+from paasta_tools.utils import atomic_file_write
 from paasta_tools.utils import is_secrets_for_teams_enabled
 from paasta_tools.utils import list_clusters
 from paasta_tools.utils import list_services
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/secret.py
```diff
@@ -470,6 +502,8 @@ def paasta_secret(args):
         secret_path = os.path.join(
             secret_provider.secret_dir, f"{args.secret_name}.json"
         )
+        if args.extra_namespaces is not None:
+            _update_extra_namespaces(secret_path, args.extra_namespaces)
```
nemacysts, line 506: should we fetch the existing list of extra namespaces in case someone is updating an existing secret? otherwise, when they call `paasta secret update` they need to re-add the extra namespaces to their cli args

paasta_tools/cli/cmds/secret.py
```diff
@@ -386,6 +401,23 @@ def _get_secret_provider_for_service(
     )
 
 
+def _update_extra_namespaces(secret_path: str, extra_namespaces_arg: str) -> None:
+    """Update the extra_namespaces field in a secret JSON file in-place."""
+    # TODO: Add metadata support to secret_provider impl (vault_tools) instead
+    # vault_tools owns the normal write path for secret JSON files; this patches
+    # only the extra_namespaces metadata field after the ciphertext has been written.
+    namespaces: List[str] = [n for n in extra_namespaces_arg.split(",") if n]
```
nemacysts: i think we could probably do some argparse shenanigans to have this be None or a list from the start

e.g., i think you can (ab)use `type` and so something like:
```
...
type=lambda v: v.split(','),
default=None,
...
```

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
```diff
@@ -298,6 +317,53 @@ def get_services_to_k8s_namespaces_to_allowlist(
...
+        with os.scandir(secret_dir) as secret_file_paths:
+            for secret_file_path in secret_file_paths:
+                if not secret_file_path.path.endswith(".json"):
+                    continue
+                with open(secret_file_path, "r") as f:
+                    data = json.load(f)
+                extra_namespaces = data.get("extra_namespaces", [])
+                if not extra_namespaces:
+                    continue
+                secret_name = secret_file_path.name[: -len(".json")]
```
nemacysts:
```suggestion
                secret_name = secret_file_path.name.removesuffix(".json")
```
might be nicer :p

nemacysts: (also, we don't need to do any shenanigans with sanitizing the file-derived name or anything?)

nemacysts (dismissed):
noice! just a couple comments/questions - but none are blocking (e.g., making `paasta secret update` work better can be another PR if you'd prefer to just get this out)

also, just curious: for the non-service secrets, will we essentially make a fake service directory in soaconfigs and start adding secrets there?

paasta_tools/cli/cmds/secret.py
jfongatyelp: Yeah that's a good idea--Originally I was thinking the same, but then that would require either supporting some syntax of expressing a removal in the arg (eg `--extra-namespaces newnstooadd,-oldnstoremove` ), or a separate arg (eg `--remove-extra-namespaces oldnstoremove`) that makes things a little more complicated? 

That said I also need to improve this workflow with an existing secret later because right now, if you just want to add some namespaces, this actually ends up going through the _entire_ `update` flow -- including having to provide the plaintext secret & re-encrypting again, which more than a bit annoying!

So yeah since I have more work to do here, I will try to merge this (to get at least extra namespace management possible) and focus on the 'update' stuff in a PR after.

jfongatyelp:
> also, just curious: for the non-service secrets, will we essentially make a fake service directory in soaconfigs and start adding secrets there?

Good question, I had actually just assumed any workflows would operate in the same sort of space as a normal service, but for these I feel like just treating them like a 'shared' secret might work well here? Shared secrets for all other usecases should have only been created if an instance used them, so any that got created only for these `extra_namespaces` should land only in those namespaces.

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
jfongatyelp: (ugh how did i miss this claude-ism, ty for pointing it out! That said apparently we already just do a replace(".json","") [elsewhere](https://github.com/Yelp/paasta/blob/master/paasta_tools/kubernetes/bin/paasta_secrets_sync.py#L463), which i'm glad hasn't caused an issue yet 😂 )

Good question, and one I didn't verify before, but it seems like we are good:
https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/secret.py#L471 

when adding/updating, we are directly using the secret_name as the filename too. There's some [checks](https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/cmds/secret.py#L47-L58) but no changes done to the passed-in arg

paasta_tools/cli/cmds/secret.py
jfongatyelp: Thanks, this is now a List everywhere but the args

paasta_tools/kubernetes/bin/paasta_secrets_sync.py
jfongatyelp: This should now be fixed

nemacysts (approved):

PR #4292, opened 2026-04-27
feat: remove smartstack annotation if not smartstack

sidtuladhar:
## Issue
Currently, even if a service isn't registering in smartstack, a "smartstack_registrations" annotation will still be assigned to pods. It would be cleaner to omit it when we don't need it.

## Changes
- Removed "smartstack_registration" depending on whether it's in the smartstack or not.

Note: Since smart_stack registration is now an optional schema, this will trigger a **big bounce**

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2343,10 +2343,16 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             if force_no_routable_ip
             else self.has_routable_ip(service_namespace_config, system_paasta_config)
         )
-        annotations: KubePodAnnotations = {
-            "smartstack_registrations": json.dumps(self.get_registrations()),
-            "paasta.yelp.com/routable_ip": has_routable_ip,
-        }
+        annotations: KubePodAnnotations = (
+            {
+                "smartstack_registrations": json.dumps(self.get_registrations()),
+                "paasta.yelp.com/routable_ip": has_routable_ip,
+            }
+            if service_namespace_config.is_in_smartstack()
+            else {
+                "paasta.yelp.com/routable_ip": has_routable_ip,
+            }
+        )
 
         # The HPAMetrics collector needs these annotations to tell it to pull
         # metrics from these pods
@@ -2853,10 +2859,8 @@ def get_kubernetes_services_running_here(
     services = []
     pods = get_k8s_pods()
     for pod in pods["items"]:
-        if (
-            pod["status"]["phase"] != "Running"
-            or "smartstack_registrations" not in pod["metadata"].get("annotations", {})
-            or (exclude_terminating and pod["metadata"].get("deletionTimestamp"))
+        if pod["status"]["phase"] != "Running" or (
+            exclude_terminating and pod["metadata"].get("deletionTimestamp")
         ):
             continue
         try:
@@ -2878,7 +2882,9 @@ def get_kubernetes_services_running_here(
                     port=port,
                     pod_ip=pod["status"]["podIP"],
                     registrations=json.loads(
-                        pod["metadata"]["annotations"]["smartstack_registrations"]
+                        pod["metadata"]["annotations"].get(
+                            "smartstack_registrations", "[]"
+                        )
                     ),
                     weight=weight,
                 )
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -2347,11 +2347,6 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             "paasta.yelp.com/routable_ip": has_routable_ip,
         }
 
-        if service_namespace_config.is_in_smartstack():
-            annotations["smartstack_registrations"] = json.dumps(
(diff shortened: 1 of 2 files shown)


paasta_tools/kubernetes_tools.py
```diff
@@ -2343,10 +2343,16 @@ def get_pod_template_spec(
...
+        annotations: KubePodAnnotations = (
+            {
+                "smartstack_registrations": json.dumps(self.get_registrations()),
+                "paasta.yelp.com/routable_ip": has_routable_ip,
+            }
+            if service_namespace_config.is_in_smartstack()
+            else {
+                "paasta.yelp.com/routable_ip": has_routable_ip,
+            }
+        )
```
nemacysts: i think i'd personally have this look like:

```suggestion
        annotations: KubePodAnnotations = {
            "paasta.yelp.com/routable_ip": has_routable_ip,
        }
        
        if service_namespace_config.is_in_smartstack():
            annotations["smartstack_registrations"] = json.dumps(self.get_registrations())
```

so that this looks a little less busy / is easier to scan - and also, to make it a little simpler to deal with if we ever add any additional "default" annotations

paasta_tools/kubernetes_tools.py
```diff
@@ -2853,10 +2859,8 @@ def get_kubernetes_services_running_here(
     services = []
     pods = get_k8s_pods()
     for pod in pods["items"]:
-        if (
-            pod["status"]["phase"] != "Running"
-            or "smartstack_registrations" not in pod["metadata"].get("annotations", {})
```
nemacysts, line 2858: i think we might wanna keep this conditional as-is based on the docstring from `get_all_kubernetes_services_running_here()` + some of the fields in `KubernetesServiceRegistration()` - i think our intent here is to only return mesh-registered services

(and removing the annotation check here would then result in returning additional services now, right?)

nemacysts (commented):
just left two small comments - otherwise looks pretty good!

paasta_tools/kubernetes_tools.py
sidtuladhar: i agree!

paasta_tools/kubernetes_tools.py
sidtuladhar: yes!

paasta_tools/kubernetes_tools.py
```diff
@@ -2878,7 +2882,9 @@ def get_kubernetes_services_running_here(
                     port=port,
                     pod_ip=pod["status"]["podIP"],
                     registrations=json.loads(
-                        pod["metadata"]["annotations"]["smartstack_registrations"]
+                        pod["metadata"]["annotations"].get(
+                            "smartstack_registrations", "[]"
+                        )
```
nemacysts: nit: i think that if we get to this line, pod["metadata"]["annotations"]["smartstack_registrations"] should be guaranteed to exist sice we're checking for `"smartstack_registrations" not in pod["metadata"].get("annotations", {})` further above

(if i'm not mistaken)

tests/test_kubernetes_tools.py
```diff
@@ -1901,11 +1899,19 @@ def test_get_pod_template_spec(
...
+                "smartstack_registrations": '["kurupt.fm"]',
+                "paasta.yelp.com/routable_ip": routable_ip,
+                "iam.amazonaws.com/role": "",
+            }
+        else:
+            expected_annotations = {
+                "paasta.yelp.com/routable_ip": routable_ip,
+                "iam.amazonaws.com/role": "",
+            }
+
```
nemacysts, line 1914: feel free to ignore, this was already silly before: i think ideally we'd have updated the pytest parametrization to have the expected annotations so that we don't need branches in our tests - but probably not really worth fixing now :p

(i.e., this is fine for now)

(we can maybe toss `claude` or something at our tests later and do some bulk-refactoring 'cause there's definitely some repeated antipatterns we've accrued over the years :p)

tests/test_kubernetes_tools.py
```diff
@@ -3565,8 +3571,8 @@ def test_get_kubernetes_services_running_here():
                             "paasta.yelp.com/instance": "fm",
                         },
                         "annotations": {
-                            "smartstack_registrations": "[]",
                             "iam.amazonaws.com/role": "",
+                            "smartstack_registrations": "[]",
```
nemacysts: thinking about this a little harder: after this change, will we ever have `"smartstack_registrations": "[]"`? if you're in our mesh, presumably there's a registration for you :p

nemacysts: (i.e., i think that maybe we have some mocks that don't reflect reality now?)

sidtuladhar: yeah it wouldn't be `[]` but some value but since these tests only check if some value exists in the tests my reasoning was '[]' or 'abcd' would be the same, especially since if now there's no `smartstack_registration` then it def doesnt exist. But better to be more explicit like `["kafka-operator.main"]`

nemacysts (approved):

ilkinmammadzada (approved):

PR #4293, opened 2026-04-27
fix: shell cli autocomplete

sidtuladhar:
## Issue
Currently the paasta CLI autocomplete is broken and we want to be able to tab complete services, clusters, etc.

## Changes
- Fix `list_func` doesn't take in kwargs
- Parser is built + provides autocomplete depending on the args

## Test
Works with `bash` and `zsh`

### DIFF
--- a/paasta_tools/cli/cli.py
+++ b/paasta_tools/cli/cli.py
@@ -218,9 +218,16 @@ def parse_args(argv):
     :return: an argparse.Namespace object mapping parameter names to the inputs
              from sys.argv
     """
+    # NOTE: in "real" use, these env vars are set by
+    #paasta_tabcomplete.sh (which we plop into 
+    # usr/share/bash-completion/completions/ as our completion script)
     if "_ARGCOMPLETE" in os.environ:
+        # NOTE: COMP_LINE will be the full command so far
         comp_line = os.environ.get("COMP_LINE", "")
         comp_words = comp_line.split()
+        
+        # ...but we want to grab the subcommand since the
+        # shell can handle tab-completing `paasta` :p
         if len(comp_words) >= 2 and comp_words[1] in PAASTA_SUBCOMMANDS:
             parser = get_argparser(commands=[comp_words[1]])
         else:
--- a/paasta_tools/cli/cli.py
+++ b/paasta_tools/cli/cli.py
@@ -226,6 +226,7 @@ def parse_args(argv):
         else:
             parser = get_argparser(commands=[])
         argcomplete.autocomplete(parser)
+        return parser.parse_args(argv), parser
 
     parser = get_argparser(commands=[])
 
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cli.py
```diff
@@ -218,14 +218,22 @@ def parse_args(argv):
     :return: an argparse.Namespace object mapping parameter names to the inputs
              from sys.argv
     """
+    if "_ARGCOMPLETE" in os.environ:
+        comp_line = os.environ.get("COMP_LINE", "")
+        comp_words = comp_line.split()
+        if len(comp_words) >= 2 and comp_words[1] in PAASTA_SUBCOMMANDS:
+            parser = get_argparser(commands=[comp_words[1]])
+        else:
+            parser = get_argparser(commands=[])
```
nemacysts, line 234: might be nice to add some comments here for future us, e.g.:

```suggestion
    # NOTE: in "real" use, these env vars are set by
    #paasta_tabcomplete.sh (which we plop into 
    # usr/share/bash-completion/completions/ as our completion script)
    if "_ARGCOMPLETE" in os.environ:
        # NOTE: COMP_LINE will be the full command so far
        comp_line = os.environ.get("COMP_LINE", "")
        comp_words = comp_line.split()
        
        # ...but we want to grab the subcommand since the
        # shell can handle tab-completing `paasta` :p
        if len(comp_words) >= 2 and comp_words[1] in PAASTA_SUBCOMMANDS:
            parser = get_argparser(commands=[comp_words[1]])
        else:
            parser = get_argparser(commands=[])
```

nemacysts: i think we could maybe also do some funky tuple unpacking to get rid of the len()/indexing - but at the cost of maybe needing to do some exception handling, so i think we can maybe just stick to adding some quick comments for now :p

paasta_tools/cli/utils.py
```diff
@@ -572,7 +572,7 @@ def run_on_master(
 
 def lazy_choices_completer(list_func):
     def inner(prefix, **kwargs):
-        options = list_func(**kwargs)
```
nemacysts, line 575: i think this is safe to remove atm since we're not using this, but was this causing any issues?

nemacysts (commented):
just a minor comment about comments - otherwise, looks good to me!

paasta_tools/cli/utils.py
sidtuladhar: yeah it was causing the tab complete to fail

sidtuladhar: for example, if i did
`paasta info --service <tab>`
then inner gets:
```
inner(prefix="", action=..., etc..)
```
but the `list_func` in this case is `list_services` and it doesn't take those kwargs so it fails (only take `soa_dir`).

nemacysts (approved):

PR #4294, opened 2026-04-27
PAASTA-17887 - improving PaaSTA deployment error handling

mbrankin-art:
### **Issue**
Improving error handling for PaaSTA. Can fail on deployment missing 'paasta.yelp.com/instance' label. 
### **Changes**
Updated 'list_all_paasta_deployments' to filter by both 'paasta.yelp.com/instance' and 'paasta.yelp.com/service' labels.
Updated the relevant unit test to verify the changes.

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -3588,8 +3588,9 @@ def write_annotation_for_kubernetes_service(
 
 
 def list_all_paasta_deployments(kube_client: KubeClient) -> Sequence[KubeDeployment]:
-    """Gets deployments in all namespaces by passing the service label selector"""
-    label_selectors = "paasta.yelp.com/service"
+    """Gets deployments in all namespaces and passes the 'managed=true' label selector to
+    ensure only PaaSTA-managed deployments are returned."""
+    label_selectors = "paasta.yelp.com/managed=true"
     return list_deployments_in_all_namespaces(
         kube_client=kube_client, label_selector=label_selectors
     )
--- a/tests/test_kubernetes_tools.py
+++ b/tests/test_kubernetes_tools.py
@@ -3880,6 +3880,12 @@ def test_list_all_paasta_deployments(addl_labels, replicas):
     )
 
     assert list_all_paasta_deployments(kube_client=mock_client) == []
+    mock_client.deployments.list_deployment_for_all_namespaces.assert_called_once_with(
+        label_selector="paasta.yelp.com/managed=true"
+    )
+    mock_client.deployments.list_stateful_set_for_all_namespaces.assert_called_once_with(
+        label_selector="paasta.yelp.com/managed=true"
+    )
     mock_items = [
         mock.Mock(
             metadata=mock.Mock(
@@ -3890,6 +3896,7 @@ def test_list_all_paasta_deployments(addl_labels, replicas):
                     "yelp.com/paasta_config_sha": "b12345",
                     "paasta.yelp.com/service": "kurupt",
                     "paasta.yelp.com/instance": "fm",
+                    "paasta.yelp.com/managed": "true",
                     "paasta.yelp.com/git_sha": "a12345",
                     "paasta.yelp.com/config_sha": "b12345",
                     **addl_labels,
@@ -3906,6 +3913,7 @@ def test_list_all_paasta_deployments(addl_labels, replicas):
                     "yelp.com/paasta_config_sha": "b12345",
                     "paasta.yelp.com/service": "kurupt",
                     "paasta.yelp.com/instance": "am",
+                    "paasta.yelp.com/managed": "true",
                     "paasta.yelp.com/git_sha": "a12345",
                     "paasta.yelp.com/config_sha": "b12345",
                     **addl_labels,
@@ -3914,6 +3922,12 @@ def test_list_all_paasta_deployments(addl_labels, replicas):
             )
         ),
     ]
+    mock_client.deployments.list_deployment_for_all_namespaces.assert_called_once_with(
+        label_selector="paasta.yelp.com/managed=true"
+    )
+    mock_client.deployments.list_stateful_set_for_all_namespaces.assert_called_once_with(
+        label_selector="paasta.yelp.com/managed=true"
+    )
 
     # Setting the number of replicas this way since spec
     # is a reserved argument for Mocks


paasta_tools/kubernetes_tools.py
```diff
@@ -3164,31 +3164,48 @@ def ensure_paasta_namespace_limits(kube_client: KubeClient, namespace: str) -> N
...
+        service = labels.get("paasta.yelp.com/service")
+        instance = labels.get("paasta.yelp.com/instance")
+
+        if not service or not instance:
+            raise ValueError(
+                f"Found object with incomplete PaaSTA labels: "
+                f"namespace={item.metadata.namespace} "
+                f"name={item.metadata.name} "
+                f"labels={labels}"
+            )
```
nemacysts: I wonder if we want to wrap this with a custom exception (e.g., `raise IncompletePaastaDeploymentMetadata()`) and then catch that further above so that we can still deploy everything else (and then at the end of setup_kubernetes_job, exit non-zero)

would give us some breathing room to fix things - otherwise, we're relying on the shuffling we do for service names to help us keep making forward progress until we fix whatever is broken :p)

paasta_tools/kubernetes_tools.py
```diff
@@ -3164,31 +3164,48 @@ def ensure_paasta_namespace_limits(kube_client: KubeClient, namespace: str) -> N
 def list_deployments_in_all_namespaces(
     kube_client: KubeClient, label_selector: str
 ) -> List[KubeDeployment]:
+    MANAGED_NAMESPACE_PREFIXES = ("paasta", "paastasvc")
```
nemacysts: we should probably move this to the top of the file with the other NAMED_CONSTANTS :)

paasta_tools/kubernetes_tools.py
```diff
@@ -3164,31 +3164,48 @@ def ensure_paasta_namespace_limits(kube_client: KubeClient, namespace: str) -> N
...
-            replicas=(
-                item.spec.replicas
-                if item.metadata.labels.get(paasta_prefixed("autoscaled"), "false")
-                == "false"
-                else None
-            ),
+
+    result = []
+    for item in deployments.items + stateful_sets.items:
+        if not item.metadata.namespace.startswith(MANAGED_NAMESPACE_PREFIXES):
```
nemacysts: I think we might want to do something slightly different here?

I think that doing something like
```
managed_namespaces = get_all_managed_namespaces(...)
...
for item in deployments.items + stateful_sets.items:
    if item.metadata.namespace not in managed_namespaces:
        continue
```

might be slightly preferable since not all namespaces that are paasta managed are guaranteed to start with `paasta` (we allow folks to override the namespace and some have picked pretty arbitrary names)


it also looks like we already have a [list_deployments_in_managed_namespaces()](https://github.com/Yelp/paasta/blob/807e7f4a97a15cd2e3dd8b5212783a11735902b4/paasta_tools/kubernetes_tools.py#L3228) too - so maybe we just move the existing callers of this to that other function?

nemacysts: (and maybe add the error-handling/logging to that other function?)

paasta_tools/kubernetes_tools.py
ilkinmammadzada: What about just writing a warning log line and keep iterate deployments?

nemacysts: imo, we're never gonna look at a warning line on its own, so we should essentially do both: write a warning line and continue iterating on the other deployments, but then exit with a non-zero exit code at the end so that we get alerted to whatever is broken and actually fix it

(that said, i think we could also probably have a log-based (or metric-based) ticketing alert if we don't want to get paged)

ilkinmammadzada: Nothing is broken in this case actually, is it? 
If we (probably namespace owner) want to get a ticket/page if any non-paasta workloads doesn't respect the contract, we can use gatekeeper. What do you think?

nemacysts: note: this is looking at paasta-managed namespaces - so the workloads in these namespaces should at least be partially paasta (i say partially since there's a couple funky cases where a namespace has both paasta and non-paasta pods :p)

that said, that sounds good to me - that gatekeeper shouldn't be too hard to write and should already have 0 violations, so we could put it in block mode pretty quickly :p

nemacysts: @mbrankin-art we should be able to simplify this a bit more now (get rid of the custom exception -> log whenever there's an issue in `list_deployments_in_managed_namespaces` + skip that deployment and continue iterating

and then we can make a new ticket for setting up a gatekeeper rule to catch deployments that are being skipped here :)

ilkinmammadzada (commented):
instead of checking only instance label, we can create a constant with all required labels (we may have one) and check here.

paasta_tools/kubernetes_tools.py
```diff
@@ -236,11 +236,14 @@
 AUTOSCALING_OVERRIDES_CONFIGMAP_NAME = "paasta-autoscaling-overrides"
 AUTOSCALING_OVERRIDES_CONFIGMAP_NAMESPACE = "paasta"
 
+REQUIRED_LABELS = ("service", "instance")
```
ilkinmammadzada: we have more required labels I think: https://github.com/Yelp/paasta/blob/32d7a902597cc770dfb0db6abd4b014bab108133/paasta_tools/kubernetes_tools.py#L3211-L3224
Probably we will need to add all of them to the list

paasta_tools/kubernetes_tools.py
```diff
@@ -3232,18 +3235,31 @@ def list_deployments_in_managed_namespaces(
     ret: List[KubeDeployment] = []
     for namespace in get_all_managed_namespaces(kube_client):
         try:
```
ilkinmammadzada, line 3236: As we discussed, client code "list_deployments_in_managed_namespaces" already have required label filters:
https://github.com/Yelp/paasta/blob/f8855e22c165be1d4a928eb0ebc8e4bc47b5893a/paasta_tools/instance/kubernetes.py#L707

Also the issue mentioned in ticket happened in `list_deployments_in_all_namespaces`.
We revisited the stack trace and found out we will need to change the direction a bit :)

ilkinmammadzada (dismissed):

paasta_tools/kubernetes_tools.py
```diff
@@ -3586,8 +3586,8 @@ def write_annotation_for_kubernetes_service(
 
 
 def list_all_paasta_deployments(kube_client: KubeClient) -> Sequence[KubeDeployment]:
-    """Gets deployments in all namespaces by passing the service label selector"""
-    label_selectors = "paasta.yelp.com/service"
+    """Gets deployments in all namespaces by passing the service and instance label selector"""
+    label_selectors = "paasta.yelp.com/service,paasta.yelp.com/instance"
```
ilkinmammadzada: I think we can use paasta.yelp.com/managed: true to make sure that deployments are managed by PaaSTA, how does it sound? @mbrankin-art @nemacysts

nemacysts: ++ 

i'm a fan of that idea! i honestly wouldn't mind tagging all the objects we create with a label like that to make it easier to exclude things that folks might be yoloing into their namespaces if they happen to have elevated privileges to do so :p

nemacysts (dismissed):
++

i think we can follow-up with the managed label in another PR (and maybe even add that managed label to more objects in said PR)

ilkinmammadzada:
> ++
> 
> i think we can follow-up with the managed label in another PR (and maybe even add that managed label to more objects in said PR)

We already have that label on our deployments, so we can use it directly here :)

ilkinmammadzada (approved):

nemacysts (approved):

PR #4296, opened 2026-05-01
PEREL-5334: Add timeout_server_threshold to smartstack schema

ahaswell:
## Summary

- Add `timeout_threshold_ratio` (number, 0-1) to smartstack monitoring schema
- Document the field in yelpsoa_configs.rst
- Remove dead fields: `latency_threshold_ms` (monitoring), `latency_threshold` (endpoints)

**Jira:** https://jira.yelpcorp.com/browse/PEREL-5334
**Exporter PR:** https://github.yelpcorp.com/docker-images/yelpsoa-configs-mirror/pull/37

## Plan

The perel shard's latency alert fires when p99 latency exceeds a threshold computed as **80% of `timeout_server_ms`**. This ratio is hardcoded in the yelpsoa-configs-mirror exporter - service owners have no way to tune it without setting an absolute ms value (a field that was never adopted).

Add a new `timeout_threshold_ratio` field (0-1 ratio) to `smartstack.yaml` under `monitoring`. The exporter multiplies this by `timeout_server_ms` to produce the metric value consumed by the alert.

### Example usage

```yaml
# smartstack.yaml
main:
  proxy_port: 20000
  timeout_server_ms: 60000
  monitoring:
    timeout_threshold_ratio: 0.5   # alert at 50% of 60s = 30s
```

If not specified, defaults to `0.8` (current behavior, unchanged).

### Scope of changes

| Repo | File | Change |
|------|------|--------|
| Yelp/paasta | `cli/schemas/smartstack_schema.json` | Add `timeout_threshold_ratio` (number, 0-1). Remove dead fields: `latency_threshold_ms` (monitoring), `latency_threshold` (endpoints). |
| Yelp/paasta | `docs/source/yelpsoa_configs.rst` | Document the new field in the Error Alerting section. |
| docker-images/yelpsoa-configs-mirror | `prometheus_exporter.py` | Read `timeout_threshold_ratio` from config (default 0.8 via `SvcMonitoringDefaults`). Compute `latency_threshold_ms = timeout_threshold_ratio * timeout_server_ms`. |
| docker-images/yelpsoa-configs-mirror | `tests/` | Update tests to exercise the new field. |

### What doesn't change

- Prometheus metric names (`yelpsoa_configs_service_latency_threshold_ms`, `..._endpoint`)
- Alerting rules on perel (already correct)
- Per-endpoint threshold behavior (inherits namespace value)
- `metrics.py` gauge definitions

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -1029,7 +1029,7 @@ These keys provide optional overrides for the default alerting behaviour.
           page_nonprod: true
           error_threshold_ratio: 0.02
           minimum_error_rps: 10
-          timeout_server_threshold: 0.8
+          timeout_threshold_ratio: 0.8
           default_endpoint_alerting: true
           endpoints:
             - name: GET /something
@@ -1052,7 +1052,7 @@ These keys provide optional overrides for the default alerting behaviour.
      environments. Defaults to **false**.
    - ``error_threshold_ratio``: Error threshold ratio (0-1) for errors under this namespace. Defaults to **0.01**.
    - ``minimum_error_rps``: Minimum error rate per second for errors under this namespace before an alert can be triggered, minimum is zero. Defaults to **5**.
-   - ``timeout_server_threshold``: Ratio (0-1) of ``timeout_server_ms`` at which to alert on p99 latency. Defaults to **0.8** (80%).
+   - ``timeout_threshold_ratio``: Ratio (0-1) of ``timeout_server_ms`` at which to alert on p99 latency. Defaults to **0.8** (80%).
    - ``default_endpoint_alerting``: Turn on alerts for all endpoints in this namespace. Defaults to **false**.
    - ``endpoint_error_threshold_ratio``: Error threshold ratio (0-1) for errors to any singular endpoint. Defaults to the namespace ``error_threshold_ratio`` if specified, or **0.01**.
    - ``endpoint_minimum_error_rps``: Minimum error rate per second for errors to any singular endpoint before an alert can be triggered for errors to any singular endpoint. Defaults to the namespace ``minimum_error_rps`` if specified, or **5**.
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -1029,6 +1029,7 @@ These keys provide optional overrides for the default alerting behaviour.
           page_nonprod: true
           error_threshold_ratio: 0.02
           minimum_error_rps: 10
+          timeout_server_threshold: 0.8
           default_endpoint_alerting: true
           endpoints:
             - name: GET /something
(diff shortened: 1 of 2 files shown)


thejoycekung (dismissed):

docs/source/yelpsoa_configs.rst
```diff
@@ -1051,6 +1052,7 @@ These keys provide optional overrides for the default alerting behaviour.
      environments. Defaults to **false**.
    - ``error_threshold_ratio``: Error threshold ratio (0-1) for errors under this namespace. Defaults to **0.01**.
    - ``minimum_error_rps``: Minimum error rate per second for errors under this namespace before an alert can be triggered, minimum is zero. Defaults to **5**.
+   - ``timeout_server_threshold``: Ratio (0-1) of ``timeout_server_ms`` at which to alert on p99 latency. Defaults to **0.8** (80%).
```
nemacysts: i assume that the p99 is a hardcoded percentile as well (and not configurable)?

nemacysts: that said, i think this name is potentially somewhat confusing (and inconsistent with the others)

...but i also don't really have any suggestions for a better name

(other than s/threshold/ratio to at least by consistent with the other params)

nemacysts: (although, i guess latency_threshold_ratio or latency_limit_ratio wouldn't be too bad -- but timeout_server_ratio would work for me :p)

ahaswell: I removed the use of latency per gtl requests to call it timeout as thats really whats being alerted on. I called it _threshold as its the % of the existing timeout_server_ms that the namespace owner wants to alert at. You can see its use in the linked pr for the soa-configs mirror

ahaswell: re p99 we are using it with otel:duration_milliseconds:byroute:p99:5m see the perel shard alerting rules

nemacysts: re: your  `_threshold` reasoning: is that not what all the existing `_ratio` params do as well? :) (% of x to alert on)

ahaswell: I can rename it timeout_server_ratio if thats preferred it seems more ambiguous to me. timeout_server_threshold_ratio seems clearer but verbose

ahaswell: also the failing tests dont seem related to my change

nemacysts: and re: p99s: what i'm actually getting at is: i'm not sure if we're digging ourselves into a confusing config hole if there are multiple levels of percentile-based alerts

ahaswell: p99 is fixed in the recording rule and isn't something service owners configure. The google SRE book uses p99 for latency SLOs so it seemed like a sensible default. the 5min window smooths out noise so it only fires on sustained slowness. the 80% default is arbitrary so we need to give owners a way to tune it for their service. i don't think there's a config hole since the percentile isn't something we're exposing

ahaswell: i can rename to timeout_threshold_ratio so its consistent with error_threshold_ratio from default error alerting, though it loses the obvious connection to timeout_server_ms

nemacysts: (for context: i am asking these questions because i don't have deep knowledge on how the default error alerting works (and what the plans for it are) and because historically our slo-based alerting has become too complex for our users to effectively configure)

re: the naming: i guess it's up to y'all since you'll get onpoint questions, but it does seem inconsistent to have this one param that is a ratio not include ratio in the name when all the others do :)

nemacysts (dismissed):
i (personally) think that it'd be nice to be consistent with the options here, but i also don't feel that strongly about it if y'all are fine with this as-is :)

docs/source/yelpsoa_configs.rst
ahaswell: agreed, consistency is better. renamed to timeout_threshold_ratio to match error_threshold_ratio. @nemacysts

nemacysts (approved):

PR #4299, opened 2026-05-05
PAASTA-18875: Making max surge configurable

mbrankin-art:
## Issue
`max_surge` is currently hard-coded at `"100%"` in `get_deployment_strategy_config()`. We want to make it configurable so services can control how aggressively they bounce.

## Changes
- Made bounce overprovision configurable at two levels:
  - **Per-service** — via `bounce_overprovision_factor` in EKS soaconfigs (float, 0.25-1). 0.25 is set as the hard floor to avoid any issues occurring when setting the value to 0%.
  - **System default** — via `default_bounce_overprovision_factor` in `SystemPaastaConfig` (defaults to `1.0`)
- Added `bounce_overprovision_factor` to the EKS schema (`eks_schema.json`) with number validation (`0 < value <= 1`)
- Updated documentation in `docs/source/bouncing.rst` and `docs/source/yelpsoa_configs.rst`
- Added unit tests for `get_bounce_overprovision_factor()` at both per-service and system levels
- Updated existing `test_get_deployment_strategy` and `test_warning_big_bounce_*` tests to mock the new method
- Uses 0-1 float values consistent with existing percent-based config like `bounce_margin_factor`

### DIFF
--- a/docs/source/bouncing.rst
+++ b/docs/source/bouncing.rst
@@ -179,3 +179,12 @@ Tron Bouncing
 
 In PaaSTA Tron jobs are simply configured to use new code or config **on the
 next execution of the job**. In progress jobs are not adjusted or killed.
+
+Controlling Surge During Bounces
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, PaaSTA attempts to start as many replicas as were previously running during a bounce at once.
+If your service (among other reasons) is large and/or lives in a small pool, you can tune how many replicas PaaSTA will
+essentially "overprovision" during a bounce with the ``bounce_overprovision_factor`` option
+
+ See the docs on `bounce_overprovision_factor <yelpsoa_configs.html#eks-[clustername].yaml>`_ file.
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -416,6 +416,13 @@ instance MAY have:
       This fraction of boxes must agree that the task is up for the bounce to treat a task as healthy.
       Defaults to 1.0 -- haproxy on all queried boxes must agree that the task is up.
 
+  * ``bounce_overprovision_factor``: controls the maximum number of extra instances allowed above the desired count during a bounce,
+    as a fraction from 0.25 to 1. 0.25 is set as the floor to ensure at least 25% surge capacity is allowed.
+    * **Type:** Number (0.25-1)
+    * **Default:** System default is set to ``1.0``
+
+    Example:: bounce_overprovision_factor : 0.25
+
   * ``bounce_margin_factor``: proportionally increase the number of old instances
     to be drained when the crossover bounce method is used.
     0 < bounce_margin_factor <= 1. Defaults to 1 (no influence).
--- a/paasta_tools/cli/schemas/eks_schema.json
+++ b/paasta_tools/cli/schemas/eks_schema.json
@@ -197,6 +197,13 @@
                 "deploy_group": {
                     "type": "string"
                 },
+                "bounce_overprovision_factor": {
+                    "type": "number",
+                    "minimum": 0.25,
+                    "maximum": 1,
+                    "exclusiveMinimum": false,
+                    "comment": "Fraction of extra instances allowed during a bounce. Defaults to 1.0. Floor set to 0.25"
+                },
                 "autoscaling": {
                     "$ref": "autoscaling_schema.json#autoscaling_params"
                 },
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -983,7 +983,10 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
         strategy_type = KUBE_DEPLOY_STATEGY_MAP[bounce_method]
 
         if strategy_type == "RollingUpdate":
-            max_surge = "100%"
+            # k8s expects an absolute number (an integer) or a integer percentage string (e.g., 55%)
+            # and since we don't allow folks to set an absolute number to surge by, we need to protect against someone being
+            # clever and trying to set a surge of something like ".1234" for 12.34% by rounding the input
+            max_surge = f"{int(round(self.get_bounce_overprovision_factor() * 100, ndigits=2))}%"
             if bounce_method == "crossover":
                 max_unavailable = "{}%".format(
                     int((1 - self.get_bounce_margin_factor()) * 100)
--- a/paasta_tools/long_running_service_tools.py
+++ b/paasta_tools/long_running_service_tools.py
@@ -95,6 +95,7 @@ class LongRunningServiceConfigDict(InstanceConfigDict, total=False):
     replication_threshold: int
     bounce_start_deadline: float
     bounce_margin_factor: float
+    bounce_overprovision_factor: float
     should_ping_for_unhealthy_pods: bool
     weight: int
     unhealthy_pod_eviction_policy: str
@@ -234,6 +235,12 @@ class LongRunningServiceConfig(InstanceConfig):
     # FIXME(jlynch|2016-08-02, PAASTA-4964): DEPRECATE nerve_ns and remove it
     def get_nerve_namespace(self) -> str:
         return decompose_job_id(self.get_registrations()[0])[1]
+
+    def get_bounce_overprovision_factor(self) -> float:
+        return self.config_dict.get(
+            "bounce_overprovision_factor",
+            load_system_paasta_config().get_bounce_overprovision_factor(),
+        )
 
     def get_registrations(self) -> List[str]:
         for registration in self.get_invalid_registrations():
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -1920,6 +1920,7 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     datastore_credentials_vault_env_overrides: Dict[str, str]
     default_push_groups: List
     default_should_use_uwsgi_exporter: bool
+    default_bounce_overprovision_factor: float
     deploy_blacklist: UnsafeDeployBlacklist
     deployd_metrics_provider: str
     deploy_whitelist: UnsafeDeployWhitelist
@@ -2166,6 +2167,9 @@ class SystemPaastaConfig:
         if hosts.startswith("zk://"):
             return hosts[len("zk://") :]
         return hosts
+
+    def get_bounce_overprovision_factor(self) -> float:
+        return self.config_dict.get("default_bounce_overprovision_factor", 1.0)
 
     def get_system_docker_registry(self) -> str:
         """Get the docker_registry defined in this host's cluster config file.
--- a/tests/test_long_running_service_tools.py
+++ b/tests/test_long_running_service_tools.py
@@ -17,6 +17,7 @@ from pytest import raises
 
 from paasta_tools import long_running_service_tools
 from paasta_tools.utils import InvalidInstanceConfig
+from paasta_tools.utils import SystemPaastaConfig
 
 
 class TestLongRunningServiceConfig:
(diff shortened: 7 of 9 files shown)


paasta_tools/long_running_service_tools.py
```diff
@@ -235,6 +235,13 @@ def get_drain_method_params(
     def get_nerve_namespace(self) -> str:
         return decompose_job_id(self.get_registrations()[0])[1]
 
+    # PAASTA-18875 - max surge configurable
```
nemacysts:
```suggestion
```

should be fine to not link to the ticket here :)

nemacysts: (and the same comment below)

paasta_tools/long_running_service_tools.py
```diff
@@ -235,6 +235,13 @@ def get_drain_method_params(
...
+    def get_max_surge(self) -> str:
+        system_paasta_config = load_system_paasta_config()
+        return self.config_dict.get(
+            "max_surge", system_paasta_config.get_default_max_surge()
```
nemacysts: we'll probably want to update the docs in https://github.com/Yelp/paasta/tree/master/docs/source to mention this new config option

and we might want to come up with a name that's more grokkable by our users without needing to read the docs / know what max_surge is in k8s

paasta_tools/utils.py
```diff
@@ -2165,6 +2166,10 @@ def get_zk_hosts(self) -> str:
             return hosts[len("zk://") :]
         return hosts
 
+    # PAASTA-18875: max_surge configurable
+    def get_default_max_surge(self) -> str:
+        return self.config_dict.get("default_max_surge", "100%")
```
nemacysts: hmm, i don't know if we have any SystemPaastaConfig atm that isn't just a pure string/number - i'm a little worried that someone will try to override this in puppet and forget the %

(and actually, same concern with `get_max_surge()` - although for that one we have the protection of the soaconfigs schema)

paasta_tools/long_running_service_tools.py
nemacysts: (and we'll also want to add this to the eks schema at https://github.com/Yelp/paasta/blob/master/paasta_tools/cli/schemas/eks_schema.json)

paasta_tools/long_running_service_tools.py
mbrankin-art: apologies! will remove this! :D

paasta_tools/long_running_service_tools.py
mbrankin-art: I have pushed a new version of the branch. Let me know if I have done the documentation in the correct place and if I have satisfied the other comments too :D!

docs/source/bouncing.rst
```diff
@@ -179,3 +179,11 @@ Tron Bouncing
 
 In PaaSTA Tron jobs are simply configured to use new code or config **on the
 next execution of the job**. In progress jobs are not adjusted or killed.
+
+Controlling Surge During Bounces
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, PaaSTA allows up to 100% extra instances during a bounce. This behaviour can be tuned using the ``max_surge``
+option in your service's config.
```
nemacysts:
```suggestion
By default, PaaSTA attempts to start as many replicas as were previously running during a bounce at once. If your service (among other reasons) is large and/or lives in a small pool, you can tune how many replicas PaaSTA will essentially "overprovision" during a bounce with the ``max_surge`` option
```

docs/source/bouncing.rst
```diff
@@ -179,3 +179,11 @@ Tron Bouncing
...
+
+See yelpsoa_configs for more details.
```
nemacysts: just fyi: we can link to the relevant section with something like https://github.com/Yelp/paasta/blob/f0db5744d5d16df3c93c72d09620612794ecfc57/docs/source/bouncing.rst?plain=1#L146 :D

docs/source/yelpsoa_configs.rst
```diff
@@ -416,6 +416,12 @@ instance MAY have:
       This fraction of boxes must agree that the task is up for the bounce to treat a task as healthy.
       Defaults to 1.0 -- haproxy on all queried boxes must agree that the task is up.
 
+  * ``bounce_surge``: controls the maximum number of extra instances allowed above the desired count during a bounce.
```
nemacysts: thoughts on something like `bounce_overprovision_factor`/`bounce_overprovision_percent`

(i kinda like this name too - just trying to think of what might be easier for feature devs to remember/use)

docs/source/yelpsoa_configs.rst
```diff
@@ -416,6 +416,12 @@ instance MAY have:
...
+    * **Type:** String (e.g. ``"100%"``, ``"50%"``)
+    * **Default:** System default is set to ``"100%"`` unless overridden.
```
nemacysts: i'm also thinking that maybe we want to be consistent with what we do for existing percent-based values and treat 'em as values from 0-1 so that we don't need to worry about converting strings to ints and can have some additional validation in the jsonschema

thoughts?

docs/source/yelpsoa_configs.rst
mbrankin-art: `bounce_overprovision_percent` sounds good to me!

docs/source/yelpsoa_configs.rst
mbrankin-art: sounds good to me, i'll refactor the code to be more inline with what we have :D

docs/source/bouncing.rst
```diff
@@ -179,3 +179,12 @@ Tron Bouncing
 
 In PaaSTA Tron jobs are simply configured to use new code or config **on the
 next execution of the job**. In progress jobs are not adjusted or killed.
+
+Controlling Surge During Bounces
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+By default, PaaSTA attempts to start as many replicas as were previously running during a bounce at once.
+If your service (among other reasons) is large and/or lives in a small pool, you can tune how many replicas PaaSTA will
+essentially "overprovision" during a bounce with the ``max_surge`` option
```
nemacysts:
```suggestion
essentially "overprovision" during a bounce with the ``bounce_overprovision_percent`` option
```

nemacysts: although, having this new name be right next to `bounce_margin_factor` makes me think we should go for `bounce_overprovision_factor` for consistency :p

paasta_tools/cli/schemas/eks_schema.json
```diff
@@ -197,6 +197,13 @@
                 "deploy_group": {
                     "type": "string"
                 },
+                "bounce_overprovision_percent": {
+                    "type": "number",
+                    "minimum": 0,
+                    "maximum": 1,
+                    "exclusiveMinimum": false,
```
nemacysts, line 204: just curious (i don't actually know the answer): what does k8s do if maxSurge is set to 0%?

paasta_tools/kubernetes_tools.py
```diff
@@ -983,7 +983,7 @@ def get_deployment_strategy_config(self) -> V1DeploymentStrategy:
         strategy_type = KUBE_DEPLOY_STATEGY_MAP[bounce_method]
 
         if strategy_type == "RollingUpdate":
-            max_surge = "100%"
+            max_surge = f"{round(self.get_bounce_overprovision_percent() * 100)}%"
```
nemacysts: another question that i also don't know the answer to already: would k8s handle values like 55.5% (e.g., if someone sets `.555`)? i.e,. do we need to round?

that said, i don't see folks doing anything but `0.XX` or `0.X` as overrides in general :p

nemacysts (commented):
noice! i just have one last comment re: naming (sorry!) and a couple questions 'cause i'm a curious person :p 

it'd be good to also get @ilkinmammadzada's review, but i think this is pretty much ready to be merged soon :D

paasta_tools/cli/schemas/eks_schema.json
mbrankin-art: I spoke to @ilkinmammadzada about this earlier on today and he had said if maxSurge was set to 0% it would not be ideal. Because of this we need a suitable floor. I was reading the k8s docs and saw that they list 'max surge' with a default of [25%](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-surge)  Is this too much or too little? what do you think?

paasta_tools/kubernetes_tools.py
mbrankin-art: looking at [docs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#max-surge) for max surge it seems the absolute number is calculated by rounding up the percentage.

so like you said im sure people are only doing 0.XX or 0.X so i suppose the edge case for rounding would be quite rare

nemacysts: coolio, from reading the docs it does seem like its something we'd want to prevent validation errors on the k8s side - so looks like the rounding is here to stay!

that said, we might wanna add a comment so that we don't confuse ourselves in the future when we inevitably forget this conversation :p

e.g., `k8s expects an absolute number (an integer) or a integer percentage string (e.g., 55%) - since we don't allow folks to set an absolute number to surge by, we need to protect against someone being clever and trying to set a surge of something like ".1234" for 12.34% by rounding the input :p`

nemacysts: that said: i think we need to set `ndigits=2` on the `round()` call - otherwise, `round(.55) == 1`) which isn't exactly what we want

paasta_tools/cli/schemas/eks_schema.json
nemacysts: 25%/.25 seems good as a minimum to me! it's pretty easy to update this value if we pick wrong, so we can go with this for now :D

paasta_tools/kubernetes_tools.py
mbrankin-art: i've recently pushed up new code to satisfy the new comments. lmk if it looks okay :D

docs/source/bouncing.rst
nemacysts:
```suggestion
essentially "overprovision" during a bounce with the ``bounce_overprovision_factor`` option
```

paasta_tools/cli/schemas/eks_schema.json
```diff
@@ -197,6 +197,13 @@
                 "deploy_group": {
                     "type": "string"
                 },
+                "bounce_overprovision_factor": {
+                    "type": "number",
+                    "minimum": 0,
```
nemacysts: @mbrankin-art did we want to set this to 
```suggestion
                    "minimum": 0.25,
```
?

paasta_tools/kubernetes_tools.py
```diff
@@ -983,7 +983,10 @@ def get_deployment_strategy_config(self) -> V1DeploymentStrategy:
         strategy_type = KUBE_DEPLOY_STATEGY_MAP[bounce_method]
 
         if strategy_type == "RollingUpdate":
-            max_surge = "100%"
+            """k8s expects an absolute number (an integer) or a integer percentage string (e.g., 55%)
+            since we don't allow folks to set an absolute number to surge by, we need to protect against someone being
+            clever and trying to set a surge of something like ".1234" for 12.34% by rounding the input"""
```
nemacysts: for consistency with our other multi-line comments, we should do something like

```suggestion
            # k8s expects an absolute number (an integer) or a integer percentage string (e.g., 55%)
            # and since we don't allow folks to set an absolute number to surge by, we need to protect against someone being
            # clever and trying to set a surge of something like ".1234" for 12.34% by rounding the input"""
```

paasta_tools/long_running_service_tools.py
```diff
@@ -235,6 +235,16 @@ def get_drain_method_params(
     def get_nerve_namespace(self) -> str:
         return decompose_job_id(self.get_registrations()[0])[1]
 
+    def get_bounce_overprovision_factor(self) -> float:
+        system_paasta_config = load_system_paasta_config()
+        return cast(
```
nemacysts: ah, just noticed the cast here - i think we can get rid of that by adding bounce_overprovision_factor to KubernetesDeploymentConfigDict in https://github.com/Yelp/paasta/blob/44136c17367c8f4913a78af3e72b1276233c0571/paasta_tools/kubernetes_tools.py#L438

paasta_tools/utils.py
```diff
@@ -2167,6 +2168,11 @@ def get_zk_hosts(self) -> str:
             return hosts[len("zk://") :]
         return hosts
 
+    def get_bounce_overprovision_factor(self) -> float:
+        return cast(
```
nemacysts: i think since this is in `SystemPaastaConfigDict` now, we don't need the cast anymore

tests/test_long_running_service_tools.py
```diff
@@ -472,3 +472,59 @@ def test_get_expected_instance_count_for_namespace():
             cluster="fake_cluster",
             instance_type_class=long_running_service_tools.LongRunningServiceConfig,
         )
+
+
+class TestGetBounceOverprovisionFactor:
+    def test_returns_value_from_config(self):
+        """Uses per-service config when set."""
+        mock_system_config = mock.Mock()
```
nemacysts: i think we might wanna set this to have a spec of SystemPaastaConfig so that `mock` can guard against us mocking things that aren't part of that class

nemacysts (commented):
i have some pretty minor comments after giving this another read-through - they're mostly consistency/cleanup things though, the overall approach is good :)

paasta_tools/cli/schemas/eks_schema.json
mbrankin-art: Yes, sorry missed this when refactoring my b!

nemacysts (approved):

ilkinmammadzada (approved):

PR #4303, opened 2026-05-06
Add SL2 Grafana dashboard link to paasta status output

sidtuladhar:
Shows a direct link to the service-load Grafana dashboard for each instance, pre-filtered by service, ecosystem, cluster, and instance.

Screenshot 2026-05-11 at 14 23 21

Kinda long but hopefully we can change it with y links at some point. 

Also, it accurately uses _cluster_ for the `SuperRegion` parameter

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -1031,9 +1031,9 @@ def print_kubernetes_status_v2(
     instance_state = get_instance_state(status)
     dashboard_url = get_sl2_dashboard(cluster, service, instance)
     output.append(
-        f"    {PaastaColors.yellow('y/sl2')}       {PaastaColors.blue(dashboard_url)}"
+        f"    {PaastaColors.yellow('y/sl2:')}      {PaastaColors.blue(dashboard_url)}"
     )
-    output.append(f"    State: {instance_state}")
+    output.append(f"    State:      {instance_state}")
     output.append("    Running versions:")
     if not verbose:
         output.append(
@@ -1537,7 +1537,7 @@ def print_kubernetes_status(
     )
     dashboard_url = get_sl2_dashboard(cluster, service, instance)
     output.append(
-        f"    {PaastaColors.yellow('y/sl2')}       {PaastaColors.blue(dashboard_url)}"
+        f"    {PaastaColors.yellow('y/sl2:')}      {PaastaColors.blue(dashboard_url)}"
     )
     output.append(f"    State:      {bouncing_status} - Desired state: {desired_state}")
 
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -327,18 +327,9 @@ def paasta_status_on_api_endpoint(
     is_eks: bool = False,
     all_namespaces: bool = False,
 ) -> int:
-    ecosystem = system_paasta_config.get_ecosystem_for_cluster(cluster)
-    dashboard_url = (
-        f"https://grafana.yelpcorp.com/d/d304815a-aea2-4de9-8e67-d37645967801/service-load"
-        f"?var-ServiceName={service}"
-        f"&var-Ecosystem={ecosystem}"
-        f"&var-SuperRegion={cluster}"
-        f"&var-Paasta_Instance={instance}"
-    )
     output = [
         "",
         f"\n{service}.{PaastaColors.cyan(instance)} in {cluster}{' (EKS)' if is_eks else ''}",
-        f"    Grafana:    {PaastaColors.blue(dashboard_url)}",
     ]
     client = get_paasta_oapi_client(
         cluster=get_paasta_oapi_api_clustername(cluster=cluster, is_eks=is_eks),
@@ -373,6 +364,7 @@ def paasta_status_on_api_endpoint(
     # TODO: Remove this when all clusters are returning status.version
     elif status.git_sha != "":
         output.append(f"    Git sha:    {status.git_sha} (desired)")
+
     instance_types = find_instance_types(status)
     if not instance_types:
         output.append(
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -327,9 +327,18 @@ def paasta_status_on_api_endpoint(
...
+        f"https://grafana.yelpcorp.com/d/d304815a-aea2-4de9-8e67-d37645967801/service-load"
+        f"?var-ServiceName={service}"
+        f"&var-Ecosystem={ecosystem}"
+        f"&var-SuperRegion={cluster}"
+        f"&var-Paasta_Instance={instance}"
+    )
     output = [
         "",
         f"\n{service}.{PaastaColors.cyan(instance)} in {cluster}{' (EKS)' if is_eks else ''}",
+        f"    Grafana:    {PaastaColors.blue(dashboard_url)}",
```
nemacysts: i think we only want to show this for now for normal paasta services (i.e., we'll need to look at the instance class, which means we might want to move this to happen in the paasta api instead since that should already have logic that keys off of the instance class/type and would let us do different things for type) - otherwise, the charts are gonna be slightly confusing

e.g., for a flink service, most of the graphs are going to be empty, but then there'll be a couple that are using different filters (e.g., the tracing/endpoint ones) that will show data unrelated to these instances

the sl2 dashboards are maybe still helpful for these folks since it'll still show data for things like resource utilization - but we can probably wait for these operator owners to add something similar for their instance types

(and we sorta need to point to a different dashboard anyway for tron :p)

nemacysts: i'm also fine with the label here being `Grafana`, but i do also wonder what'd be best from the perspective of our users: `Grafana`, `Dashboard`, `Load Dashboard`, `y/sl2`, etc :p

although, i guess given your perspective as a new-hire - that's probably enough evidence that a `Grafana` label will be understandable for our users :D

sidtuladhar: i actually like `y/sl2` bc grafana is kinda obvious from the url and for more info engineers can always go to` y/sl2`

sidtuladhar: actually SL2 is better for keep style the same

sidtuladhar: oh yeah that makes sense

paasta_tools/cli/cmds/status.py
```diff
@@ -373,6 +374,16 @@ def paasta_status_on_api_endpoint(
...
+    if any(t in ("kubernetes", "kubernetes_v2", "eks") for t in instance_types):
+        ecosystem = system_paasta_config.get_ecosystem_for_cluster(cluster)
+        dashboard_url = (
+            f"https://grafana.yelpcorp.com/d/d304815a-aea2-4de9-8e67-d37645967801/service-load"
+            f"?var-ServiceName={service}"
+            f"&var-Ecosystem={ecosystem}"
+            f"&var-SuperRegion={cluster}"
+            f"&var-Paasta_Instance={instance}"
+        )
+        output.append(f"    SL2:        {PaastaColors.blue(dashboard_url)}")
```
nemacysts: we could probably save ourselves having to maintain this allowlist if we push this down one step further and move this to the INSTANCE_TYPE_WRITERS functions

(which would then make it clearer to future us that if we wanted to have a similar thing for tronjobs (but pointing to y/tronjob-status), we'd make the change in the corresponding tron writer)

sidtuladhar:
I can move the dashboard_url creation to a seperate function if that's cleaner @nemacysts

paasta_tools/cli/cmds/status.py
```diff
@@ -126,6 +126,7 @@
             str,
             "cluster",  # noqa: F821  # flake8 false-positive, these are not var references
         ),
+        Arg(str, "ecosystem"),  # noqa: F821  # flake8 false-positive
```
nemacysts: imo, we can recalculate these in the writers as needed rather than always needing to plumb this through even if a writer doesn't need it

nemacysts: `system_paasta_config.get_ecosystem_for_cluster(cluster)` should be decently fast to run and if anything, we can always ttl_cache (if we call it outside of a CLI context) or lru_cache (if we only call it in a CLI context) it

paasta_tools/cli/cmds/status.py
```diff
@@ -1010,13 +1016,22 @@ def print_flinkeks_status(
...
     verbose: int = 0,
 ) -> int:
     instance_state = get_instance_state(status)
+    dashboard_url = (
+        f"https://grafana.yelpcorp.com/d/d304815a-aea2-4de9-8e67-d37645967801/service-load"
+        f"?var-ServiceName={service}"
+        f"&var-Ecosystem={ecosystem}"
+        f"&var-SuperRegion={cluster}"
+        f"&var-Paasta_Instance={instance}"
+    )
```
nemacysts: i'd probably wrap this in a function in case the dashboard gets moved around/renamed/etc - then we just have to update one place :p

nemacysts: (could also have the url come from SystemPaastaConfig, but if the URL is gonna change I think we'll have enough notice (and the existing URL will continue to work) that we can just change it in the one function :p)

paasta_tools/cli/cmds/status.py
```diff
@@ -1010,13 +1016,22 @@ def print_flinkeks_status(
...
+    output.append(f"    SL2:        {PaastaColors.blue(dashboard_url)}")
```
nemacysts: up to you: thoughts on `SL2:` vs `y/sl2:`?

as someone less blinded by years at yelp, which would you find more useful as a starting point to get more information? 

(or would either one be about the same to you as you'd just see a grafana link and assume its a dashboard and then read the url and see that it says service-load and use those as context clues :p)

sidtuladhar: yeah `y/sl2` is better when i think about it. I was thinking putting it in a color just to remind that it is a y link

Screenshot 2026-05-11 at 11 48 34

paasta_tools/cli/cmds/status.py
sidtuladhar: yessirr

nemacysts (approved):

PR #4304, opened 2026-05-06
Fix registry credentials on devboxes

cuza:
# Update the method for obtaining registry credentials.
- Moved to upstream kind
- Made kind build script less verbose
- Pinned kind and node version on Makefile
- Updated auth scripts to use containerd v2
- Removed the docker network option from the local cluster config since it does not exist upstream kind


k8s_itests/scripts/containerd_registry_setup.py
```diff
@@ -1,22 +1,34 @@
...
-    containerdcfg = tomli.load(containerdcfg_file)
 
-with open("/nail/etc/docker-registry-ro") as dockercfg_file:
-    dockercfg = json.load(dockercfg_file)
+def get_registry_auth() -> str:
+    """Get registry credentials using docker-credential-yelp-okta (interactive)."""
+    env = os.environ.copy()
+    env["PATH"] = "/usr/bin:/usr/local/bin:/bin:" + env.get("PATH", "")
+    result = subprocess.run(
+        ["docker-credential-yelp-okta", "get"],
```
nemacysts: do we need to set a PATH here? could we use the full path?

(although, this also looks like a pretty standard PATH too - shouldn't the default work?)

nemacysts (dismissed):

k8s_itests/scripts/containerd_registry_setup.py
cuza: I just deleted the script 🤣

cuza: I had that script cuz I was doing toml manipulation with containerd v1. For containerd v2 I just need to write the parts of the config I need:


```shell
podman exec "${node}" sh -c "
  mkdir -p /etc/containerd/certs.d/${REGISTRY} && \
  printf '[host.\"https://${REGISTRY}\"]\n  capabilities = [\"pull\", \"resolve\"]\n  [host.\"https://${REGISTRY}\".header]\n    Authorization = [\"Basic ${AUTH_TOKEN}\"]\n' \
    > /etc/containerd/certs.d/${REGISTRY}/hosts.toml
"
  ```

cuza: and to answer the original question I set PATH cuz the script was not able to find `docker-credential-yelp-okta` I was just going to call it with the full path but forgot to clean it up

nemacysts (approved):

piax93 (approved):

PR #4305, opened 2026-05-07
Fix local-run secrets retrieval for EKS clusters

danielpops:
## Problem or Feature 

The kube context for EKS clusters is `eks-{cluster}`, but `instance_config.cluster` only contains the base name (e.g. `pnw-devc`). This caused 404s when fetching secrets_for_teams secrets via local-run.

## Testing Done

- Tested with an eks workload (eks-pnw-devc)
- Tested with a tron workload (tron-pnw-devc)
- Both successfully retrieve secrets via `local-run --decrypt-secrets`

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -912,8 +912,12 @@ def run_docker_container(
         # if secrets_for_owner_team enabled in yelpsoa for service
         if is_secrets_for_teams_enabled(service, soa_dir):
             try:
+                cluster = instance_config.cluster
+                kube_context = (
+                    cluster if cluster.startswith("eks-") else f"eks-{cluster}"
+                )
                 kube_client = KubeClient(
-                    config_file=KUBE_CONFIG_USER_PATH, context=instance_config.cluster
+                    config_file=KUBE_CONFIG_USER_PATH, context=kube_context
                 )
                 secret_environment = get_kubernetes_secret_env_variables(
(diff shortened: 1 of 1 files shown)


paasta_tools/cli/cmds/local_run.py
```diff
@@ -912,8 +912,12 @@ def run_docker_container(
         # if secrets_for_owner_team enabled in yelpsoa for service
         if is_secrets_for_teams_enabled(service, soa_dir):
             try:
+                cluster = instance_config.cluster
+                kube_context = (
+                    cluster if cluster.startswith("eks-") else f"eks-{cluster}"
+                )
```
nemacysts, line 922: i'm not at my computer right now, but i think this might not work for eksstage or the spark clusters - i don't think that we unconditionally prefix the context names with eks (iirc, we only do that for clusters that have a non-eks pair)


that said, eksstage is only used by CI and the spark clusters only run spark-jobs (i.e., no one is local-running stuff running there), so this should be fine

(although, might be good to leave a breadcrumb comment here in case this changes)

nemacysts (dismissed):
heh, we'll have to make sure we don't re-break this when we tear down the non-eks clusters and cleanup all the dual-cluster code :p

paasta_tools/cli/cmds/local_run.py
danielpops: @nemacysts++ great call, I kinda figured that would be the case, and should have added a comment dpopes--

I'll add a comment

nemacysts (approved):

PR #4306, opened 2026-05-07
Add best-effort Jenkins link to paasta status -v output

sidtuladhar:
Prints a link to the service's Jenkins job page once per service when verbose mode is enabled.

Had to add `verbose` to some tests for the tests to pass

Screenshot 2026-05-08 at 16 14 39

### DIFF
--- a/paasta_tools/cli/cmds/status.py
+++ b/paasta_tools/cli/cmds/status.py
@@ -2147,6 +2147,14 @@ def apply_args_filters(
     return clusters_services_instances
 
 
+def print_jenkins_link(service: str, verbose: int) -> None:
+    if verbose >= 1:
+        jenkins_url = (
+            f"https://jenkins-paasta-k8s.yelpcorp.com/job/services-{service}-paasta/"
+        )
+        print(f"\nJenkins: {PaastaColors.blue(jenkins_url)}")
+
+
 def paasta_status(args) -> int:
     """Print the status of a Yelp service running on PaaSTA.
     :param args: argparse.Namespace obj created from sys.args by cli"""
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/status.py
```diff
@@ -2157,6 +2165,11 @@ def paasta_status(args) -> int:
     printer = CancellablePrinter()
     tasks = []
     clusters_services_instances = apply_args_filters(args)
+
+    services = set().union(*clusters_services_instances.values())
+    for svc in services:
+        print_jenkins_link(svc, args.verbose)
```
nemacysts: we should only ever have a single service here (you can only status a single service at a time)

this is technically fine (although i'd maybe drop the union and iterate directly over `clusters_services_instances.values()` (or, if there were potentially duplicates, drop the union and do something like `set(clusters_services_instances.values())`))

but we could probably do something like:
```
if verbose:
    [service] = clusters_services_instances.values() # or service, = clusters_services_instances.values() if that's too spooky :p
     print_jenkins_link(service)
```
to prevent any confusion in the future (or i guess add a comment somewhere around here :p)

i'd also probably move the verbosity check out of the `print_jenkins_link()` function since we can save parsing out the service if the verbosity isn't increased, and then we don't need the new function to recheck the verbosity (/ if we want to show the jenkins link elsewhere in the future, we might not want to tie that to verbosity at all)

paasta_tools/cli/cmds/status.py
```diff
@@ -2147,6 +2147,14 @@ def apply_args_filters(
     return clusters_services_instances
 
 
+def print_jenkins_link(service: str, verbose: int) -> None:
+    if verbose >= 1:
+        jenkins_url = (
+            f"https://jenkins-paasta-k8s.yelpcorp.com/job/services-{service}-paasta/"
+        )
+        print(f"\nJenkins: {PaastaColors.blue(jenkins_url)}")
```
nemacysts: thinking about this a little more: we might actually want to do something like link to `y/jenkins?filter=SERVICE`
and have the header be something like `Deployment Pipeline(s)` since we have some services with custom setups (e.g., y-m) that this link would be broken for (and others that have a standard pipeline + custom ones)

nemacysts (approved):

PR #4307, opened 2026-05-07
Fail mark-for-deployment when re-deploying same version without --wait-for-deployment

cuza:
Prevent deployment when attempting to redeploy the same version without the `--wait-for-deployment` flag.


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -490,11 +490,35 @@ def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
         service=service, deploy_group=deploy_group
     )
     if deployment_version == old_deployment_version:
-        print(
-            "Warning: The image asked to be deployed already matches what is set to be deployed:"
-        )
-        print(deployment_version)
-        print("Continuing anyway.")
+        if not args.block:
```
nemacysts, line 503: i think we'd want to flip this, no?
`args.block` is True with `--wait-for-deployment`, and we want to wait until the deploy group is healthy in that case rather than forging ahead?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -490,11 +490,35 @@ def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
...
+            print(
+                "Warning: The image asked to be deployed already matches "
+                f"what is set to be deployed in deploy group {deploy_group}:"
+            )
+            print(f"  {deployment_version}")
+            print("Checking if all instances are healthy before proceeding...")
```
nemacysts: i think we might also want to do something slightly different here - i think we probably want to then essentially pretend that we're doing a normal --wait-for-deployment bounce and poll until the deploy group is empty (and then timeout after whatever we have the usual timeout set to)

nemacysts: i.e., we want to treat an unhealthy deploy group as if it was previously on another version and wait until the "new" (really the same version, we're just re-polling again) version is healthy before continuing 

(and if --wait-for-deployment is not set, then don't do anything different: just yolo as usual)

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -487,14 +487,45 @@ def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
...
+                "Safe to proceed to the next deploy group."
+            )
+            return 0
+        else:
+            print(
+                "Error: Not all instances are healthy for this version. "
+                "A previous deploy may have failed or timed out. "
+                "Not safe to proceed to the next deploy group."
+            )
+            return 1
```
nemacysts: might be worth essentially doing what a normal m-f-d does nad run this logic in a loop until all the instances are healthy (or i guess some percentage are healthy - i think we only require `bounce_margin_factor` % of instances to be healthy to proceed?) rather than doing the check once and exiting

i think doing the logic in a loop would probably also allow us to remove the special-casing here since if everything is healthy, we'd excit that loop immediately and if not, we'd keep rechecking

paasta_tools/cli/cmds/mark_for_deployment.py
cuza: The condition is actually already doing what you're describing — `if not args.block` means "if --wait-for-deployment is NOT set, return early." When `args.block=True` (--wait-for-deployment IS set), we skip the early return and fall through to the `MarkForDeploymentProcess` state machine, which polls in a loop until all instances are healthy (respecting bounce_margin_factor) or times out.

So the two cases are:
- `--wait-for-deployment` not set → write the tag, return 0 (yolo)
- `--wait-for-deployment` set → fall through to the normal polling state machine, which does exactly what you describe: loops until healthy or timeout

paasta_tools/cli/cmds/mark_for_deployment.py
cuza: Yep — this is what the block=True path does. It falls through to MarkForDeploymentProcess which calls do_wait_for_deployment in a loop (checking bounce_margin_factor). If already healthy, it exits that loop immediately on the first poll. Auto-rollback is disabled for same-version (line 715: `self.deployment_version != self.old_deployment_version`), so it won't accidentally roll back to itself.

I've also force-pushed with a cleaner version that writes the tag on the block=False path too (needed for generate_deployments_for_service to work).

paasta_tools/cli/cmds/mark_for_deployment.py
nemacysts: oh yea, i'm dumb - i dunno why i thought that mark_for_deployment() was what kicked off the whole state machine stuff 🤣

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -499,11 +499,20 @@ def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
...
+                deploy_group=deploy_group,
+                commit=commit,
+                git_url=args.git_url,
+                image_version=args.image_version,
+            )
+            if ret != 0:
+                print("Failed to write deploy tag", file=sys.stderr)
+                return ret
+            return 0
+        print("Waiting for instances to be healthy (--wait-for-deployment).")
```
nemacysts:
```suggestion
        print("Waiting for instances to be healthy (--wait-for-deployment/wait_for_deployment: true is set).")
```

nemacysts: (just to be extra-clear)

tests/cli/test_cmds_check_deploy_health.py
```diff
@@ -0,0 +1,218 @@
...
+    autospec=True,
+)
+@patch(
+    "paasta_tools.cli.cmds.check_deploy_health.get_currently_deployed_version",
+    autospec=True,
+)
+@patch(
+    "paasta_tools.cli.cmds.check_deploy_health.validate_service_name",
+    autospec=True,
+)
```
nemacysts, line 25: i need to finish my auto-fix pre-commit branch, but can we swap these for context-manager patches? (that's what we normally do for new code since then you don't have to remember how the decorator order matches the param order :p)

...although, maybe this matter less now that llms write a lot of our tests :p

tests/cli/test_cmds_check_deploy_health.py
```diff
@@ -0,0 +1,218 @@
...
+def test_check_deploy_health_all_healthy(
+    mock_validate,
+    mock_get_version,
+    mock_get_instance_configs,
+    mock_check_done,
+):
+    mock_get_version.return_value = DeploymentVersion(sha="abc123", image_version="v1")
+    mock_instance_config = mock.Mock()
```
nemacysts: we probably want to have `spec=...` set here to ensure that we're mocking things that actually exist inisde an instance config class

nemacysts: !8ball this is another good pre-commit to write (ensure all Mock() have a spec set like we do for patches/autospec)

paasta_tools/cli/cmds/check_deploy_health.py
```diff
@@ -0,0 +1,112 @@
...
+    except NoSuchCluster:
+        print(
+            f"ERROR: Cluster for {service} not in paasta-api endpoints config",
+            file=sys.stderr,
+        )
+        return 2
+
+    instances = [
+        (cluster, ic) for cluster, ics in configs_by_cluster.items() for ic in ics
+    ]
```
nemacysts: i personally tend to break nested list comps into an actual for-loop since i find that they're probably easier for most people to read, but i don't have strong opinions (and outside of work, i kinda like nested list comps :p)

paasta_tools/cli/cmds/check_deploy_health.py
```diff
@@ -0,0 +1,112 @@
...
+    if not instances:
+        print(
+            f"ERROR: No instances found for {service}/{deploy_group}",
+            file=sys.stderr,
+        )
+        return 2
+
+    all_healthy = all(
+        check_if_instance_is_done(
```
nemacysts, line 95: just double-checking: this respects bounce_margin or whatever that's called, right? (i.e., we don't actually need all instances to be fully healthy, just a majority of 'em)

nemacysts (commented):
mostly just questions - i think this is basically exactly what we want otherwise!

paasta_tools/cli/cmds/mark_for_deployment.py
cuza: Yes — check_if_instance_is_done uses bounce_margin_factor (line 1975-1985). It considers an instance "done" when running_instance_count >= ceil(bounce_margin_factor * expected_instance_count).

paasta_tools/cli/cmds/mark_for_deployment.py
cuza: Done 👍

tests/cli/test_cmds_check_deploy_health.py
cuza: Added spec=LongRunningServiceConfig to the instance config mocks.

paasta_tools/cli/cmds/mark_for_deployment.py
cuza: No worries — the naming is confusing since mark_for_deployment() the function just writes the tag, while paasta_mark_for_deployment() is the full command with the state machine.

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -499,11 +499,20 @@ def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
         service=service, deploy_group=deploy_group, soa_dir=args.soa_dir
     )
     if deployment_version == old_deployment_version:
-        print(
-            "Warning: The image asked to be deployed already matches what is set to be deployed:"
-        )
-        print(deployment_version)
-        print("Continuing anyway.")
+        print(f"Warning: {deploy_group} already set to {deployment_version}")
+        if not args.block:
```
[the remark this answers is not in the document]
nemacysts: hmm, i can't remember if writing a deploy tag in this way will trigger a bounce - if it does, we probably do want to make sure we've done the can_user_deploy_service check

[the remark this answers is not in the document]
cuza: @nemacysts 
mark_for_deployment does call trigger_deploys(), but since it's force-pushing the same SHA to the same tag, deployd sees no version change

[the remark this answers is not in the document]
nemacysts: hmm, `on_enter_start_deploy()` calls `mark_for_deployment()` which calls `get_paasta_tag_from_deploy_group()` which adds the current timestamp to the tag - i'm admittedly a little rusty on the nitty gritty internals of how the tag handling works w.r.t. what is a noop and causes a bounce, but i'm a little worried that this will allow folks to bounce services that they shouldn't be able to bounce since the tag won't be exactly the same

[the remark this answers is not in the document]
nemacysts: @cuza just wanted to bump this question :p

PR #4310, opened 2026-05-07
Support secrets for teams in paasta secret run

danielpops:
## Summary

`paasta secret run` on a service with secrets_for_teams enabled didn't know about the Kubernetes-based secret path, and always used vault as the secret provider.

I also fixed a bug encountered while testing where `paasta secret run` always treated the service as a shared secret due to an argparse misconfiguration (store_false defaults to True)

## Testing Done

- [x] Verified paasta secret run fetches secrets from Kubernetes for a service with secrets_for_teams enabled
- [x] Verified the vault path still works for services without secrets_for_teams
- [x] Unit tests added for the new code path
- [x] Existing tests updated and passing

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -31,6 +31,7 @@ from paasta_tools.cli.utils import list_instances
 from paasta_tools.cli.utils import select_k8s_secret_namespace
 from paasta_tools.kubernetes_tools import KUBE_CONFIG_USER_PATH
 from paasta_tools.kubernetes_tools import KubeClient
+from paasta_tools.kubernetes_tools import get_kubernetes_secret_env_variables
 from paasta_tools.kubernetes_tools import get_paasta_secret_name
 from paasta_tools.kubernetes_tools import get_secret
 from paasta_tools.secret_providers import SecretProvider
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/secret.py
```diff
@@ -533,30 +530,42 @@ def paasta_secret(args):
...
+            instance=args.instance,
+            cluster=args.clusters,
             soa_dir=args.yelpsoa_config_root,
-            service_name=args.service,
-            cluster_name=args.clusters,
-            secret_provider_kwargs=secret_provider_kwargs,
         )
+        environment = instance_config.get_env()
+
+        if is_secrets_for_teams_enabled(service, args.yelpsoa_config_root):
```
nemacysts, line 541: might be nice to try to refactor these blocks into functions so that we don't have two very similar chunks of code here and in local-run, but i do see how that might be slightly funky given that local-run needs to also handle secret volumes and this doesn't -- so i won't push too hard on this :p

nemacysts (approved):

calvinli (approved):

paasta_tools/cli/cmds/secret.py
danielpops: I emotionally agree with you but I don't want to do that at the moment for the purpose of fixing this latent bug :). I promise to do it next time if I touch this code :)

PR #4311, opened 2026-05-08
Add more labels to paasta_secrets_sync timers, and support setting arbitrary dimensions via envvar

jfongatyelp:
I recently added Yet Another Cronjob for paasta secrets after #4290 , but realized our paasta_system_process_duration metrics didn't have enough information to distinguish between our standard paasta servcie secret sync, and only-namespaces syncs. 

This updates the paasta_secrets_sync system_timers to use some of our passed-in args as metric dimensions, but also adds support for callers to be able to set any arbitrary dimensions for a script using `system_timer` using a `PAASTA_METRICS_LABELS` envvar (e.g. they can set this in their cronjob(s), if they have any special labels they want to be consistent with others). 

I installed this version into eksstage and verified it would still operate as expected, and it emitted new metric labels as expected: [grafana](https://grafana.yelpcorp.com/explore?schemaVersion=1&panes=%7B%22a5k%22:%7B%22datasource%22:%22P627274A7C77417EA%22,%22queries%22:%5B%7B%22refId%22:%22A%22,%22expr%22:%22paasta_system_process_duration%7Bcluster%3D%5C%22eksstage%5C%22,%20host%3D%5C%2210-81-17-39-uswest2adevc%5C%22,%20cron_job_name%3D~%5C%22.%2A%5C%22%7D%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22P627274A7C77417EA%22%7D,%22editorMode%22:%22code%22,%22legendFormat%22:%22__auto%22%7D%5D,%22range%22:%7B%22to%22:%221778283811313%22,%22from%22:%221778283511313%22%7D%7D%7D)

### DIFF
--- a/tests/metrics/test_metrics_lib.py
+++ b/tests/metrics/test_metrics_lib.py
@@ -76,12 +76,7 @@ class TestParseMetricLabelsEnv(unittest.TestCase):
             self.assertEqual(_parse_metric_labels_env(), {})
 
     def test_envvar_absent(self):
-        env = {
-            k: v
-            for k, v in __import__("os").environ.items()
-            if k != "PAASTA_METRICS_LABELS"
-        }
-        with mock.patch.dict("os.environ", env, clear=True):
+        with mock.patch.dict("os.environ", {}, clear=True):
             self.assertEqual(_parse_metric_labels_env(), {})
 
     def test_single_label(self):
--- a/tests/metrics/test_metrics_lib.py
+++ b/tests/metrics/test_metrics_lib.py
@@ -4,6 +4,7 @@ from unittest import mock
 from py.test import raises
 
 from paasta_tools.metrics import metrics_lib
+from paasta_tools.metrics.metrics_lib import _parse_metric_labels_env
 
 
 class TestNoMetrics(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


tests/metrics/test_metrics_lib.py
```diff
@@ -67,3 +68,46 @@ def test_emit_event(self):
 
     def tearDown(self):
         metrics_lib.yelp_meteorite = None
+
+
+class TestParseMetricLabelsEnv(unittest.TestCase):
+    def test_empty_envvar(self):
+        with mock.patch.dict("os.environ", {"PAASTA_METRICS_LABELS": ""}):
+            self.assertEqual(_parse_metric_labels_env(), {})
```
nemacysts, line 76: !choose be consistent with `unittest`-style tests and have an llm rewrite these later, use plain ol' `assert X == Y` 

(i think either way is fine tbh - i wanna do a mass refactor of our tests at some point in the near future to make them more consistent (and to enforce that consistency :p)

tests/metrics/test_metrics_lib.py
```diff
@@ -67,3 +68,46 @@ def test_emit_event(self):
...
+
+    def test_envvar_absent(self):
+        env = {
+            k: v
+            for k, v in __import__("os").environ.items()
+            if k != "PAASTA_METRICS_LABELS"
+        }
```
nemacysts: i kinda think that we should maybe have a mock static env just in case someone has an env var that contains a secret - don't want them leaking those in slack when they ask for help debugging why this test is failing 

(not that it should ever really fail :p)

nemacysts (dismissed):
just some minor comments on the tests, but they're not particularly urgent

tests/metrics/test_metrics_lib.py
jfongatyelp: This is what i get for not reviewing the slop, apologies!  Good idea

tests/metrics/test_metrics_lib.py
jfongatyelp: oh til this is the _only_ test file in this repo using unititest.TestCase! Alright if i keep this as-is with tthe TestCase assertEquals format in that case?

nemacysts (approved):

PR #4312, opened 2026-05-11
feat: send Slack notification to authors after paasta rollback

sidtuladhar:
Pings the git authors whose commits are being rolled back via the service's configured Slack channel, improving visibility when a rollback occurs.

Screenshot 2026-05-11 at 15 02 55

### DIFF
--- a/paasta_tools/cli/cmds/rollback.py
+++ b/paasta_tools/cli/cmds/rollback.py
@@ -13,6 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import argparse
+from typing import Any
 from typing import Collection
 from typing import Dict
 from typing import Generator
@@ -22,6 +23,7 @@ from typing import Tuple
 from humanize import naturaltime
 
 from paasta_tools.cli.cmds.mark_for_deployment import can_user_deploy_service
+from paasta_tools.cli.cmds.mark_for_deployment import get_authors_to_be_notified
 from paasta_tools.cli.cmds.mark_for_deployment import get_deploy_info
 from paasta_tools.cli.cmds.mark_for_deployment import mark_for_deployment
 from paasta_tools.cli.utils import extract_tags
@@ -33,6 +35,7 @@ from paasta_tools.cli.utils import validate_given_deploy_groups
 from paasta_tools.deployment_utils import get_currently_deployed_version
 from paasta_tools.remote_git import create_rollback_tag
 from paasta_tools.remote_git import list_remote_refs
+from paasta_tools.slack import get_slack_client
 from paasta_tools.utils import DEFAULT_SOA_DIR
 from paasta_tools.utils import DeploymentVersion
 from paasta_tools.utils import PaastaColors
@@ -240,6 +243,34 @@ def list_previous_versions(
         )
 
 
+def notify_rollback_slack(
+    service: str,
+    deploy_group: str,
+    git_url: str,
+    rolled_back_from: DeploymentVersion,
+    new_version: DeploymentVersion,
+    deploy_info: Dict[str, Any],
+) -> None:
+    slack_client = get_slack_client()
+    channels = deploy_info.get("slack_channels", [])
+    if not channels:
+        return
+
+    authors = get_authors_to_be_notified(
+        git_url=git_url,
+        from_sha=new_version.sha,  # rollback target (old good SHA)
+        to_sha=rolled_back_from.sha,  # the bad SHA being rolled back
+        authors=None,
+    )
+
+    message = (
+        f":rewind: *Rollback* of `{service}` in `{deploy_group}`\n"
+        f"Rolled back from `{rolled_back_from.sha[:8]}` to `{new_version.sha[:8]}`\n"
+        f"{authors}"
+    )
+    slack_client.post(channels=[channels[0]], message=message)
+
+
 def paasta_rollback(args: argparse.Namespace) -> int:
     """Call mark_for_deployment with rollback parameters
     :param args: contains all the arguments passed onto the script: service,
@@ -334,6 +365,14 @@ def paasta_rollback(args: argparse.Namespace) -> int:
         # rollback than we care about if the underlying machinery was successfully able to complete the request
         if rolled_back_from != new_version:
             performed_rollback = True
+            notify_rollback_slack(
+                service=service,
+                deploy_group=deploy_group,
+                git_url=git_url,
+                rolled_back_from=rolled_back_from,
+                new_version=new_version,
+                deploy_info=deploy_info,
+            )
             audit_action_details = {
                 "rolled_back_from": str(rolled_back_from),
                 "rolled_back_to": str(new_version),
--- a/paasta_tools/cli/cmds/rollback.py
+++ b/paasta_tools/cli/cmds/rollback.py
@@ -23,7 +23,6 @@ from typing import Tuple
 from humanize import naturaltime
 
 from paasta_tools.cli.cmds.mark_for_deployment import can_user_deploy_service
-from paasta_tools.cli.cmds.mark_for_deployment import get_authors_to_be_notified
 from paasta_tools.cli.cmds.mark_for_deployment import get_deploy_info
 from paasta_tools.cli.cmds.mark_for_deployment import mark_for_deployment
 from paasta_tools.cli.utils import extract_tags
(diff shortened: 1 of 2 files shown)


sidtuladhar:
let me know if you'd like the message to be different @nemacysts

mbrankin-art (dismissed):

paasta_tools/cli/cmds/rollback.py
```diff
@@ -240,6 +243,34 @@ def format_timestamp(tstamp: str) -> str:
         )
 
 
+def notify_rollback_slack(
```
nemacysts, line 248: we should also probably tag/DM the person that initiated the rollback + add some words here to say that folks need to revert in Git as well or the rolled back change could be redployed

(I think we have some code internally in one of our jenkins-related repos that shows how to DM someone if they're not in a channel that they're tagged in a message in)

otherwise, looks good!

paasta_tools/cli/cmds/rollback.py
```diff
@@ -240,6 +243,34 @@ def format_timestamp(tstamp: str) -> str:
...
+        to_sha=rolled_back_from.sha,  # the bad SHA being rolled back
+        authors=None,
+    )
+
+    message = (
+        f":rewind: *Rollback* of `{service}` in `{deploy_group}`\n"
+        f"Rolled back from `{rolled_back_from.sha[:8]}` to `{new_version.sha[:8]}`\n"
+        f"{authors}"
+    )
+    slack_client.post(channels=[channels[0]], message=message)
```
nemacysts: we should probably post in all the channels in this list (most deploy.yaml files only have a single channel, but there's a couple with more than one)

paasta_tools/cli/cmds/rollback.py
sidtuladhar: yeah FYI 
 `# TODO: PAASTA-16927: support getting authors for services on GHE`
so we probably wont get authors in most cases?

sidtuladhar: maybe i could take a look into PAASTA-16927 before we merge this?

nemacysts: ah, in this case, we don't really want to necessarily ping the change author(s), but the person that is calling `paasta rollback` - we should be able to grab that with something like `getpass.getuser()` :)

paasta_tools/cli/cmds/rollback.py
```diff
@@ -240,6 +245,48 @@ def format_timestamp(tstamp: str) -> str:
...
+        return
+
+    authors = get_authors_to_be_notified(
+        git_url=git_url,
+        from_sha=new_version.sha,  # rollback target (old good SHA)
+        to_sha=rolled_back_from.sha,  # the bad SHA being rolled back
+        authors=None,
+    )
+
+    rollback_user = get_username()
```
nemacysts: oh nice, we already had a helper function for this :D

paasta_tools/cli/cmds/rollback.py
```diff
@@ -240,6 +245,48 @@ def format_timestamp(tstamp: str) -> str:
...
+    channels = deploy_info.get("slack_channels", [])
+    if not channels:
+        return
+
+    authors = get_authors_to_be_notified(
+        git_url=git_url,
+        from_sha=new_version.sha,  # rollback target (old good SHA)
+        to_sha=rolled_back_from.sha,  # the bad SHA being rolled back
+        authors=None,
+    )
```
nemacysts: i think we can remove getting the authors for now - i think we really care more about notifying the rollback user than anyone else 

(although, i do think that a comment saying we should support letting authors know would be nice - but for now we can let the rollback user let folks know that their changes have been rolled back/will be reverted :p)

paasta_tools/cli/cmds/rollback.py
```diff
@@ -240,6 +245,48 @@ def format_timestamp(tstamp: str) -> str:
...
+    deploy_group: str,
+    git_url: str,
+    rolled_back_from: DeploymentVersion,
+    new_version: DeploymentVersion,
+    deploy_info: Dict[str, Any],
+) -> None:
+    slack_client = get_slack_client()
+    channels = deploy_info.get("slack_channels", [])
+    if not channels:
+        return
```
nemacysts: i think that we always want to at least get to the DMing part - even if folks don't have a slack channel set

paasta_tools/cli/cmds/rollback.py
```diff
@@ -240,6 +244,41 @@ def format_timestamp(tstamp: str) -> str:
...
+    slack_client = get_slack_client()
+    channels = deploy_info.get("slack_channels", [])
+
+    authors = get_authors_to_be_notified(
+        git_url=git_url,
+        from_sha=new_version.sha,  # rollback target (old good SHA)
+        to_sha=rolled_back_from.sha,  # the bad SHA being rolled back
+        authors=None,
+    )
+
```
nemacysts: since this is somewhat of a noop right now, i'd probably move the comment below up here and then delete this bit
e.g., something like:

```suggestion
    # TODO(PAASTA-16927): we don't have a way of getting the authors for a range of commits from within paasta, but ideally we'd add the author list to the deploy channel's message (and potentially DM them to let them know their changes have been rolled back by @<rollback_user> and will likely be reverted soon)
```

(that's going to be a very long line, so if you accept the suggestion from GH you'll also want to pull the remote branch locally and break it into several lines :p)

paasta_tools/cli/cmds/rollback.py
```diff
@@ -240,6 +244,41 @@ def format_timestamp(tstamp: str) -> str:
...
+        f"{authors}\n"
+        f":warning: The rolled-back commits must also be reverted in Git, "
+        f"or they will be redeployed on the next push."
+    )
+
+    if channels:
+        slack_client.post(channels=channels, message=message)
+
+    # NICE TO HAVE: we DM all the authors but ticket below is a blocker
+    # TODO: PAASTA-16927: support getting authors for services on GHE has to be fixed first
```
nemacysts: i sorta allude to this above: but once we do have the ability to figure this out from paasta, i think we'd want this to essentially send 3 messages:
* the current message to the deploy slack channel
* the current DM to the rollback user
* a new message to authors letting them know that their changes are being rolled back (but not asking them to revert since presumably @<rollback_user> will do that :)

paasta_tools/cli/cmds/rollback.py
```diff
@@ -334,6 +373,14 @@ def paasta_rollback(args: argparse.Namespace) -> int:
         # rollback than we care about if the underlying machinery was successfully able to complete the request
         if rolled_back_from != new_version:
             performed_rollback = True
+            notify_rollback_slack(
```
nemacysts, line 373: sorry, i'm all over the place recently :p 

i think that we might want to either wrap this in a try-except that catches `SlackApiError` OR wrap the body of `notify_rollback_slack()` in a try-except (that also catches `SlackApiError`) so that we don't crash `paasta rollback` should slack be having any issues

nemacysts: (from reading paasta_tools/slack.py i was kinda hoping that using slackclient's api_call() method would skip any automatic exception raising, but from reading the slackclient code it doens't seem like that's the case (i.e., i traced a path that could get us to https://github.com/slackapi/python-slack-sdk/blob/v2/slack/web/slack_response.py#L194))

sidtuladhar: paasta uses v1 of the slack client so i dont think it has that exception

paasta_tools/cli/cmds/rollback.py
```diff
@@ -240,6 +243,37 @@ def format_timestamp(tstamp: str) -> str:
...
+            f":warning: The rolled-back commits must also be reverted in Git, "
+            f"or they will be redeployed on the next push."
+        )
+
+        if channels:
+            slack_client.post(channels=channels, message=message)
+
+        slack_client.post_single(channel=f"@{rollback_user}", message=message)
+    except Exception:
+        print("Warning: Failed to send rollback Slack notification")
```
nemacysts, line 276: we should at some point have a non-user visible log stream so that we can have data on when things like this happen - but that's for future us (this isn't the first time i've mentioned this in a PR :p - maybe i'll try this next unhack/hackathon :p)

paasta_tools/cli/cmds/rollback.py
nemacysts: ballin'!

wishful thinking on my part that we were using the latest version of the deprecated sdk :p

that said, v1 can throw a couple exceptions that i think might happen if Slack is having a bad time: e.g. ParseResponseError at https://github.com/slackapi/python-slack-sdk/blob/v1/slackclient/client.py#L191C19-L191C37 - but the exception handling code you added will catch that now!

paasta_tools/cli/cmds/rollback.py
```diff
@@ -240,6 +243,37 @@ def format_timestamp(tstamp: str) -> str:
...
+def notify_rollback_slack(
+    service: str,
+    deploy_group: str,
+    rolled_back_from: DeploymentVersion,
+    new_version: DeploymentVersion,
+    deploy_info: Dict[str, Any],
+) -> None:
+    try:
+        slack_client = get_slack_client()
+        channels = deploy_info.get("slack_channels", [])
```
nemacysts: not a blocker (/happy to ship adding this later), but might be nice to default to #deploy like we do in mark-for-deployment if there's no slack channel

nemacysts (dismissed):

paasta_tools/cli/cmds/rollback.py
sidtuladhar: i shall add it, its a small change

paasta_tools/cli/cmds/rollback.py
```diff
@@ -240,6 +245,39 @@ def format_timestamp(tstamp: str) -> str:
...
+            f":rewind: *Rollback* of `{service}` in `{deploy_group}` by <@{rollback_user}>\n"
+            f"Rolled back from `{rolled_back_from.sha[:8]}` to `{new_version.sha[:8]}`\n"
+            f":warning: The rolled-back commits must also be reverted in Git, "
+            f"or they will be redeployed on the next push."
+        )
+
+        if channels:
+            slack_client.post(channels=channels, message=message)
+        else:
+            slack_client.post(channels=[DEFAULT_SLACK_CHANNEL], message=message)
```
nemacysts:
```suggestion
        channels = deploy_info.get("slack_channels", [DEFAULT_SLACK_CHANNEL])

        # TODO(PAASTA-16927): we don't have a way of getting the authors for a range of commits
        # from within paasta, but ideally we'd add the author list to the deploy channel's message
        # (and potentially DM them to let them know their changes have been rolled back by @<rollback_user>
        # and will likely be reverted soon)
        rollback_user = get_username()
        message = (
            f":rewind: *Rollback* of `{service}` in `{deploy_group}` by <@{rollback_user}>\n"
            f"Rolled back from `{rolled_back_from.sha[:8]}` to `{new_version.sha[:8]}`\n"
            f":warning: The rolled-back commits must also be reverted in Git, "
            f"or they will be redeployed on the next push."
        )

        if channels:
            slack_client.post(channels=channels, message=message)
```

should work too :)

i.e., default to DEFAULT_SLACK_CHANNEL if no channel is set unless that's been set to a non-truthy value like an empty list (which we then treat as "don't send me slack messages")

nemacysts (commented):
one quick comment, but otherwise lgtm!

paasta_tools/cli/cmds/rollback.py
sidtuladhar: oh yeah, that is cleaner

nemacysts (approved):

PR #4313, opened 2026-05-12
STREAMAPP-15217: Support explicit pool field in Flink yelpsoa-configs

nleigh:
- Adds `pool` field to `FlinkDeploymentConfigDict`
 - Updates `get_pool()` to check for explicit `pool` value before falling back to spot-based logic
 - Enables targeting custom Karpenter pools like `flink-spot-ebs` directly from yelpsoa-configs

 ## Usage

 ```yaml
 # yelpsoa-configs: flinkeks-pnw-devc.yaml
 my_ebs_trial_instance:
   pool: flink-spot-ebs
   taskmanager:
     disk: 100Gi
 ```

 ## Context

 Part of the flink-spot-ebs experiment to trial EBS-backed instances for Flink RocksDB state.

 - Jira: STREAMAPP-15217
 - Infra PR: https://github.yelpcorp.com/misc/compute-infra-k8s/pull/3138
 - Operator PR: https://github.yelpcorp.com/services/flink-operator/pull/211

 ## Backwards compatible

 When `pool` is not set, behavior is unchanged (spot=False -> "flink", else -> "flink-spot").

 ## Test plan

 - [x] Added unit tests for explicit pool override
 - [ ] CI passes
 - [ ] Integration: deploy a test Flink job with `pool: flink-spot-ebs` on devc

### DIFF
--- a/paasta_tools/flink_tools.py
+++ b/paasta_tools/flink_tools.py
@@ -114,6 +114,7 @@ class FlinkInstanceDetails(TypedDict):
 class FlinkDeploymentConfigDict(LongRunningServiceConfigDict, total=False):
     taskmanager: TaskManagerConfig
     spot: bool
+    pool: str
     udf_plugin_name: str
     udf_plugin_version: str
 
@@ -166,19 +167,18 @@ class FlinkDeploymentConfig(LongRunningServiceConfig):
 
     def get_pool(self) -> Optional[str]:
         """
-        Parses flink_pool from a specific Flink Deployment instance's configuration data, using key 'spot'.
+        Returns the Karpenter pool name for this Flink instance.
 
-        Args:
-            flink_deployment_config_data: The FlinkDeploymentConfig for a specific Flink yelpsoa instance
-
-        Returns:
-            The flink pool string.
+        If an explicit 'pool' key is set, it takes precedence. Otherwise falls back
+        to the spot-based logic: spot=False -> "flink", else -> "flink-spot".
         """
+        pool = self.config_dict.get("pool", None)
+        if pool is not None:
+            return pool
         spot_config = self.config_dict.get("spot", None)
         if spot_config is False:
             return "flink"
         else:
-            # if not set or True, Flink instance defaults to use flink-spot pool
             return "flink-spot"
 
 
--- a/paasta_tools/flink_tools.py
+++ b/paasta_tools/flink_tools.py
@@ -166,7 +166,7 @@ class FlinkDeploymentConfig(LongRunningServiceConfig):
 
     def get_pool(self) -> Optional[str]:
         """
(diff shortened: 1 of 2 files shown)


paasta_tools/flink_tools.py
```diff
@@ -166,19 +167,18 @@ def get_replication_crit_percentage(self) -> int:
...
+        to the spot-based logic: spot=False -> "flink", else -> "flink-spot".
         """
+        pool = self.config_dict.get("pool", None)
+        if pool is not None:
+            return pool
         spot_config = self.config_dict.get("spot", None)
         if spot_config is False:
             return "flink"
         else:
-            # if not set or True, Flink instance defaults to use flink-spot pool
```
nemacysts, line 181: i'd personally leave this comment here (or `s/None/True` in `spot_config = self.config_dict.get("spot", None)` since I was real confused about how this worked until i checked what would happen in a python REPL and then came back here and read the code closer (+ saw the deleted comment) - but that's just me

tests/test_flink_tools.py
```diff
@@ -382,6 +382,30 @@ def test_non_bool_spot_value(self):
...
+        # When pool is explicitly set without spot, pool still takes precedence
+        config_dict = FlinkDeploymentConfigDict({"pool": "flink-spot-ebs"})
+        flink_deployment_config = FlinkDeploymentConfig(
+            service="test_service",
+            cluster="test_cluster",
+            instance="test_instance",
+            config_dict=config_dict,
+            branch_dict=None,
+        )
+        assert flink_deployment_config.get_pool() == "flink-spot-ebs"
```
nemacysts: non-blocking suggestion: this could be one test with `@pytest.mark.parametrize` - you could parametrize the FlinkDeploymentConfigDict and the expected pool

that said, this is also fine :)

nemacysts (commented):
just leaving a couple drive-by comments - since this isn't covered by our CODEOWNERs for this repo (on purpose), you don't really need someone from CI to ship :)

tub (dismissed):

paasta_tools/flink_tools.py
```diff
@@ -166,14 +166,14 @@ def get_replication_crit_percentage(self) -> int:
 
     def get_pool(self) -> Optional[str]:
         """
-        Parses flink_pool from a specific Flink Deployment instance's configuration data, using key 'spot'.
+        Returns the Karpenter pool name for this Flink instance.
```
nemacysts:
```suggestion
        Returns the pool a Flink instance is configured to use.
```

(or just removing Karpenter works too :p)

nemacysts: (the karpenter nodepool name is very different from the paasta one :p)

anohovsky (approved):

PR #4315, opened 2026-05-13
Emit additional deployment metrics

sidtuladhar:
- [x] How long does it take to rollback? 
- [x] How many times did rollback happen?
- [x] Add reason (rollout, rollback, automatic and etc..) label to `paasta_mark_for_deployment_deploy_duration` 
- [x] add wait_for_deployment (true, false) to `paasta_mark_for_deployment_deploy_duration` metric. 
- [x] add "paasta_instance" label to the metrics.  current one only knows about service and deploy_group.
- [x] add yelpy labels to metrics (ecosystem, runtimeenv, superregion, paasta_cluster and etc)

[Example of instance metrics](https://grafana.yelpcorp.com/explore?schemaVersion=1&panes=%7B%22qp5%22:%7B%22datasource%22:%22P627274A7C77417EA%22,%22queries%22:%5B%7B%22refId%22:%22B%22,%22expr%22:%22sum%20by%28paasta_instance,%20paasta_cluster,%20deploy_group,%20paasta_service,%20outcome,%20superregion,%20ecosystem%29%20%28paasta_mark_for_deployment_instance_deploy_duration%29%22,%22range%22:true,%22instant%22:true,%22datasource%22:%7B%22type%22:%22prometheus%22,%22uid%22:%22P627274A7C77417EA%22%7D,%22editorMode%22:%22builder%22,%22legendFormat%22:%22__auto%22,%22useBackend%22:false,%22disableTextWrap%22:false,%22fullMetaSearch%22:false,%22includeNullMetadata%22:false,%22exemplar%22:false%7D%5D,%22range%22:%7B%22from%22:%22now-1h%22,%22to%22:%22now%22%7D%7D%7D&orgId=1)

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -15,6 +15,7 @@
 """Contains methods used by the paasta client to mark a docker image for
 deployment to a cluster.instance.
 """
+
 import argparse
 import asyncio
 import concurrent
@@ -28,6 +29,7 @@ import socket
 import sys
 import time
 import traceback
+from enum import Enum
 from threading import Thread
 from typing import Any
 from typing import Callable
@@ -94,6 +96,13 @@ from paasta_tools.utils import get_username
 from paasta_tools.utils import ldap_user_search
 from paasta_tools.utils import list_services
 from paasta_tools.utils import load_system_paasta_config
+
+
+class DeployOutcome(Enum):
+    SUCCESS = "success"
+    TIMEOUT = "timeout"
+    CANCELLED = "cancelled"
+
 
 DEFAULT_DEPLOYMENT_TIMEOUT = 3 * 3600  # seconds
 DEFAULT_WARN_PERCENT = 17  # ~30min for default timeout
@@ -685,14 +694,6 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
         self.warn_pct = warn_pct
         self.mark_for_deployment_return_code = -1
         self.auto_certify_delay = auto_certify_delay
-        self.auto_abandon_delay = auto_abandon_delay
-        self.auto_rollback_delay = auto_rollback_delay
-        self.authors = authors
-        self.polling_interval = polling_interval
-        self.diagnosis_interval = diagnosis_interval
-        self.time_before_first_diagnosis = time_before_first_diagnosis
-        self.metrics_interface = metrics_interface
-        self.rollback_type: Optional[RollbackTypes] = None
         self.instance_configs_per_cluster: Dict[
             str, List[LongRunningServiceConfig]
         ] = get_instance_configs_for_service_in_deploy_group_all_clusters(
@@ -819,7 +820,7 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
 
     def schedule_paasta_status_reminder(self) -> None:
         def waiting_on_to_status(
-            waiting_on: Mapping[str, Collection[str]]
+            waiting_on: Mapping[str, Collection[str]],
         ) -> List[str]:
             if waiting_on is None:
                 return [
@@ -1147,7 +1148,7 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
         if self.block:
             thread = Thread(
                 target=self.do_wait_for_deployment,
-                args=(self.commit, self.image_version, None),
+                args=(self.commit, self.image_version),
                 daemon=True,
             )
             thread.start()
@@ -1201,6 +1202,8 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
         if self.block:
             thread = Thread(
                 target=self.do_wait_for_deployment,
+                # we pass in self.rollback_type to the thread instead of using self.rollback_type directly
+                # this avoids a race condition where self.rollback_type may change after the thread starts
                 args=(self.old_git_sha, self.old_image_version, self.rollback_type),
                 daemon=True,
             )
@@ -1871,7 +1874,7 @@ def _record_instance_duration(
     cluster: str,
     instance: str,
     elapsed: float,
-    outcome: str,
+    deploy_outcome: DeployOutcome,
     rollback_type: Optional[RollbackTypes] = None,
 ) -> None:
     if not metrics_interface:
@@ -1880,16 +1883,16 @@ def _record_instance_duration(
         cluster_info = kube_cluster.get(cluster, {})
         instance_timer = metrics_interface.create_timer(
             "instance_deploy_duration",
-            default_dimensions=dict(
-                paasta_service=service,
-                deploy_group=deploy_group,
-                paasta_cluster=cluster,
-                paasta_instance=instance,
-                superregion=cluster_info.get("superregion", "unknown"),
-                ecosystem=cluster_info.get("ecosystem", "unknown"),
-                outcome=outcome,
-                rollback_type=rollback_type.value if rollback_type else None,
-            ),
+            default_dimensions={
+                "paasta_service": service,
+                "deploy_group": deploy_group,
+                "paasta_cluster": cluster,
+                "paasta_instance": instance,
+                "superregion": cluster_info.get("superregion", "unknown"),
+                "ecosystem": cluster_info.get("ecosystem", "unknown"),
+                "deploy_outcome": deploy_outcome.value,
+                "rollback_type": rollback_type.value if rollback_type else "unknown",
+            },
         )
         instance_timer.record(elapsed)
     except Exception:
@@ -2015,7 +2018,7 @@ async def wait_for_deployment(
                         cluster,
                         instance,
                         elapsed,
-                        "success",
+                        DeployOutcome.SUCCESS,
                         rollback_type,
                     )
                     finished_instances += 1
@@ -2036,7 +2039,7 @@ async def wait_for_deployment(
                             cluster,
                             instance,
                             elapsed,
-                            "timeout",
+                            DeployOutcome.TIMEOUT,
                             rollback_type,
                         )
                 _log(
@@ -2064,7 +2067,7 @@ async def wait_for_deployment(
                             cluster,
                             instance,
                             elapsed,
-                            "cancelled",
+                            DeployOutcome.CANCELLED,
                             rollback_type,
                         )
                 # Wait for all the tasks to finish before closing out the ThreadPoolExecutor, to avoid RuntimeError('cannot schedule new futures after shutdown')
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -692,7 +692,7 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
         self.diagnosis_interval = diagnosis_interval
         self.time_before_first_diagnosis = time_before_first_diagnosis
         self.metrics_interface = metrics_interface
-        self.deploy_reason = "deploy"
+        self.rollback_type: Optional[RollbackTypes] = None
         self.instance_configs_per_cluster: Dict[
             str, List[LongRunningServiceConfig]
         ] = get_instance_configs_for_service_in_deploy_group_all_clusters(
@@ -1147,7 +1147,7 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
         if self.block:
             thread = Thread(
                 target=self.do_wait_for_deployment,
-                args=(self.commit, self.image_version, "deploy"),
+                args=(self.commit, self.image_version, None),
                 daemon=True,
             )
             thread.start()
@@ -1201,7 +1201,7 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
         if self.block:
             thread = Thread(
                 target=self.do_wait_for_deployment,
-                args=(self.old_git_sha, self.old_image_version, self.deploy_reason),
+                args=(self.old_git_sha, self.old_image_version, self.rollback_type),
                 daemon=True,
             )
             thread.start()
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -545,6 +545,7 @@ def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
             old_version=str(old_deployment_version),
             new_version=str(deployment_version),
             deploy_timeout=args.timeout,
+            wait_for_deployment=args.block,  # always True
```
nemacysts: oh, interesting: we don't emit any metrics if --wait-for-deployment isn't set or we're marking the same version atm


we might want to change part of that in the future (if not always actually emit metrics so that we can get accurate counts of m-f-d calls without having to look at the splunk metrics), but i think for now this should be fine - at least until we get to @cuza's change to not automatically skip a deploy group that is --wait-for-deployment and is already on the "new" version, but is crashlooping

but i'd maybe expand on this comment a lil' bit with something like 
```suggestion
            # atm, we'll only ever actually emit this with a
            # value of True, but this might change in the 
            # future as we update how paasta handles 
            # "redeploying" to an unhealthy deploy group
            # with --wait-for-deployment
            # (or even just to start getting metrics for the 
            # non-wait-for-deployment case :p)
            wait_for_deployment=args.block,
```

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1443,6 +1449,13 @@ def _log_rollback(self, rollback_details: Dict[str, str]) -> None:
                 name="rollback",
                 dimensions=dimensions,
             )
+        self.metrics_interface.create_counter(
+            "rollback_count",
+            default_dimensions=dict(
+                paasta_service=self.service,
+                deploy_group=self.deploy_group,
```
nemacysts: could we also add some of the dimensions from `rollback_details`?

that should have information around what type of rollback this was and whatnot

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1893,6 +1907,33 @@ async def wait_for_deployment(
...
+                    superregion=cluster_info.get("superregion", "unknown"),
+                    ecosystem=cluster_info.get("ecosystem", "unknown"),
+                    outcome=outcome,
+                ),
+            )
+            instance_timer.record(elapsed)
+        except Exception:
+            log.warning(
+                "Failed to record instance deploy duration metrics", exc_info=True
+            )
```
nemacysts: we tend to try to avoid nested functions, so if possible it'd be nice to move this out - i think we should be able to mostly pare this down to something like

```Python
# i changed the name a tad here since this is pretty single/special-purpose and probably want to discourage folks from importing this elsewhere :p
def _record_instance_deploy_duration(
    metrics_interface: TYPE_HERE,
    cluster: str,
    instance: str, 
    # i'm also fine with passing in the start time and doing the math in here
    # just seemed like a nicer signature to pass the duration directly :p
    elapsed_time: float,
    outcome: str,
) -> None:
    if not metrics_interface:
        return

    try:
        # XXX: maybe we want to add a TTL cache or something to
        # get_kube_clusters() so that we don't read this file a trillion times
        # for no reason?
        cluster_info = system_paasta_config.get_kube_clusters().get(cluster, {})
        instance_timer = metrics_interface.create_timer(
            "instance_deploy_duration",
            default_dimensions=dict(
                paasta_service=service,
                deploy_group=deploy_group,
                paasta_cluster=cluster,
                paasta_instance=instance,
                superregion=cluster_info.get("superregion", "unknown"),
                ecosystem=cluster_info.get("ecosystem", "unknown"),
                outcome=outcome,
            ),
        )
        instance_timer.record(elapsed)
    except Exception:
        log.warning(
            "Failed to record instance deploy duration metrics", exc_info=True
        )
```
and then call this as
```python
_record_instance_deploy_duration(metrics_interface, cluster, instance, "success")
```

nemacysts: i was also thinking maybe we could do this all as a decorator on `wait_until_instance_is_done()` with some context manager magic - but i think that's maybe best left as a possible future improvement :p

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -680,6 +692,7 @@ def __init__(
         self.diagnosis_interval = diagnosis_interval
         self.time_before_first_diagnosis = time_before_first_diagnosis
         self.metrics_interface = metrics_interface
+        self.deploy_reason = "deploy"
```
nemacysts: might be nice to have an enum for these so that we don't have the string "deploy" in various places

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1834,6 +1863,39 @@ def get_instance_configs_for_service_in_deploy_group_all_clusters(
...
+def _record_instance_duration(
+    metrics_interface: Optional[metrics_lib.BaseMetrics],
+    kube_cluster: Dict[str, Any],
+    service: str,
+    deploy_group: str,
+    cluster: str,
+    instance: str,
+    elapsed: float,
+    outcome: str,
+    reason: str = "deploy",
```
nemacysts: if we make this an enum of reasons, i think we'd also want to pass the enum as-is through to this function and then call <enum>.value inside this function to make the call sites a little cleaner)

paasta_tools/cli/cmds/mark_for_deployment.py
nemacysts: (that said, is my GH acting up or do we not read from self.deploy_reason anywhere?)

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1415,18 +1433,21 @@ def __build_rollback_audit_details(
         }
 
     def log_slo_rollback(self) -> None:
+        self.deploy_reason = RollbackTypes.AUTOMATIC_SLO_ROLLBACK.value
```
nemacysts: we might want a comment here (or in the constructor) about how we're using this :p

that said, i do also wonder if this is going to confuse us in the future - e.g., i think what we're gonna see in the metrics is that a mfd timer is going to take `sum(initial_deploy + potential_rollback)` and we won't ever emit timings for the initial deploy on its own

i do also somewhat worry about the ergonomics of this and us potentially using the wrong dimensions in places - perhaps we want this to be `self.rollback_reason` (and have that default to None)?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1443,6 +1464,14 @@ def _log_rollback(self, rollback_details: Dict[str, str]) -> None:
                 name="rollback",
                 dimensions=dimensions,
             )
+        self.metrics_interface.create_counter(
+            "rollback_count",
+            default_dimensions=dict(
+                paasta_service=self.service,
+                deploy_group=self.deploy_group,
+                rollback_type=rollback_details.get("rollback_type", "unknown"),
+            ),
```
nemacysts: oh, i dunno why a bunch of our older code does this - but you can also do

```suggestion
            default_dimensions={
                "paasta_service": self.service,
                "deploy_group": self.deploy_group,
                "rollback_type": rollback_details.get("rollback_type", "unknown"),
            },
```

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -545,6 +545,14 @@ def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
...
             new_version=str(deployment_version),
             deploy_timeout=args.timeout,
+            wait_for_deployment=args.block,
+            # atm, we'll only ever actually emit this with a
+            # value of True, but this might change in the
+            # future as we update how paasta handles
+            # "redeploying" to an unhealthy deploy group
+            # with --wait-for-deployment
+            # (or even just to start getting metrics for the
+            # non-wait-for-deployment case :p)
```
nemacysts: nit: the comment before the arg is probably a little easier to follow
```suggestion
            # atm, we'll only ever actually emit this with a
            # value of True, but this might change in the
            # future as we update how paasta handles
            # "redeploying" to an unhealthy deploy group
            # with --wait-for-deployment
            # (or even just to start getting metrics for the
            # non-wait-for-deployment case :p)
            wait_for_deployment=args.block,
```

paasta_tools/cli/cmds/mark_for_deployment.py
sidtuladhar: after thinking about this, we currently have:

1. MFD timer -> captures sum(full deploy + potential rollback) -> emits no details
2. Instance timer -> captures full deploy and potential rollback separately -> emits reason (deploy, slo_failure, metric_failure, user_rollback)
3. rollback counter -> increments on rollback per service -> emits rollback_type (user_rollback, cli_rollback, slo_failure, metric_failure)

sidtuladhar: having `rollback_type` instead of `reason` for instance timer would be clearer

paasta_tools/cli/cmds/mark_for_deployment.py
sidtuladhar: we call it through with a thread and pass an an arg so maybe thats why?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -582,7 +590,11 @@ def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
         ret = deploy_process.run()
         return ret
     finally:
-        deploy_timer.stop(tmp_dimensions={"exit_status": ret})
+        deploy_timer.stop(
+            tmp_dimensions={
+                "exit_status": ret,
```
nemacysts, line 604: oh, just noticed: do we need to set `reason` here as well?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1834,6 +1863,39 @@ def get_instance_configs_for_service_in_deploy_group_all_clusters(
...
+            "instance_deploy_duration",
+            default_dimensions=dict(
+                paasta_service=service,
+                deploy_group=deploy_group,
+                paasta_cluster=cluster,
+                paasta_instance=instance,
+                superregion=cluster_info.get("superregion", "unknown"),
+                ecosystem=cluster_info.get("ecosystem", "unknown"),
+                outcome=outcome,
+                rollback_type=rollback_type.value if rollback_type else None,
```
nemacysts:
```suggestion
                rollback_type=rollback_type.value if rollback_type else "unknown",
```
to match the rest?

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1834,6 +1863,39 @@ def get_instance_configs_for_service_in_deploy_group_all_clusters(
...
 
+def _record_instance_duration(
+    metrics_interface: Optional[metrics_lib.BaseMetrics],
+    kube_cluster: Dict[str, Any],
+    service: str,
+    deploy_group: str,
+    cluster: str,
+    instance: str,
+    elapsed: float,
+    outcome: str,
```
nemacysts: @sidtuladhar thoughts on making this an enum as well?

paasta_tools/cli/cmds/mark_for_deployment.py
nemacysts: actually - i think nvm: i think if we wanted the most accurate metrics we'd also need to end the og timer and start another time for the rollback should a rollback happen

...and that seems sorta annoying to do atm - and we can probably live with checking the exit status to determine if we're looking at the timer for a successful deploy or one that timedout/rolledback

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1134,7 +1147,7 @@ def on_enter_deploying(self) -> None:
         if self.block:
             thread = Thread(
                 target=self.do_wait_for_deployment,
-                args=(self.commit, self.image_version),
+                args=(self.commit, self.image_version, None),
```
nemacysts: just curious: do we need to pass None/self.rollback_type?

since these are all members of `MarkForDeploymentProcess`, they should have access to `self`/`self.rollback_type`, right? or is there some threading weirdness i'm missing

nemacysts: (/`do_wait_for_deployment` has a default for `rollback_type` - could we obviate the None here even if we still need the non-None case below?)

nemacysts: although, from looking at this closer: i think that we're passing this through since we might have more than one `do_wait_for_deployment` so we'd prefer to not have that read `self` for this?

...so maybe we can live with adding a comment in `on_enter_rolling_back()` as to why we're passing class members to the function rather than just reading them internally?

nemacysts (commented):
(sorry for all the comments - there's really just ~3: i was essentially just talking to myself here :p)

paasta_tools/cli/cmds/mark_for_deployment.py
sidtuladhar: yeah, that;s what i was thinking as well

paasta_tools/cli/cmds/mark_for_deployment.py
sidtuladhar: yep, good idea

paasta_tools/cli/cmds/mark_for_deployment.py
sidtuladhar: It makes sense to put this as `None` bc during a normal deploy with no rollback the `rollback_type` would be `None`

paasta_tools/cli/cmds/mark_for_deployment.py
sidtuladhar: we dont need to pass None since that's default, but yeah it would be good to add a comment

paasta_tools/cli/cmds/mark_for_deployment.py
nemacysts: oh true - not sure what i was thinking when i wrote this 🤣

nemacysts (approved):
:D

PR #4320, opened 2026-05-15
STREAMHAUS-1242 Validate Flink job monitoring team in paasta validate

anohovsky:
## Summary

- Adds \`validate_flink_monitoring_team()\` to \`paasta validate\`, checking that every job's \`monitoring.team\` in \`flink-*.yaml\` and \`flinkeks-*.yaml\` files is a valid Sensu team name
- Uses \`get_sensu_team_data()\` for validation — returns \`{}\` for unknown teams
- Wired into \`paasta_validate_soa_configs\` alongside existing checks
- Catches both missing \`team\` fields and invalid team names

## Motivation

Previously this check lived as a custom pre-commit hook in yelpsoa-configs using the paasta venv directly. Moving it to \`paasta validate\` is the right architectural fit since it's a static config check and \`paasta validate\` already runs on all changed flink configs via the \`run-paasta-validate.sh\` hook.

## Test plan

- [x] 5 unit tests added covering: no flink files, valid team, invalid team, missing team, missing monitoring block
- [x] Manually tested against all streamhouse-family services in yelpsoa-configs

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -1144,7 +1144,7 @@ def _check_proxy_port_in_use(service: str, namespace: str, port: int) -> bool:
 
 
 def _check_advertise_discover(
-    smartstack_data: dict[str, Any]
+    smartstack_data: dict[str, Any],
 ) -> None:  # XXX: we should use a TypedDict here
     """Need to ensure a few properties about smartstack files
     1) discover is a member of advertise
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -71,6 +71,7 @@ from paasta_tools.long_running_service_tools import METRICS_PROVIDER_UWSGI
 from paasta_tools.long_running_service_tools import METRICS_PROVIDER_UWSGI_V2
 from paasta_tools.long_running_service_tools import METRICS_PROVIDER_WORKER_LOAD
 from paasta_tools.long_running_service_tools import LongRunningServiceConfig
+from paasta_tools.monitoring_tools import get_sensu_team_data
 from paasta_tools.secret_tools import get_secret_name_from_ref
 from paasta_tools.secret_tools import is_secret_ref
 from paasta_tools.secret_tools import is_shared_secret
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -1272,6 +1273,95 @@ def validate_smartstack(service_path: str) -> bool:
     return True
 
 
+def validate_flink_monitoring_team(service_path: str) -> bool:
```
nemacysts, line 1276: normally i'd say that we should try to follow a format like what we do in the other validation hooks, i.e.(in pseudocode):
```
for cluster in list_clusters_for_service()):
   for instance, config in load_all_configs_for_service():
       if config.instance_type not in FLINK_TYPES:
           continue
       
       ... do stuff with config.GETTER_FUNCTIONS() here ...
```

but since we don't really have the flink schemas in paasta (and afaik, we don't have getters that work for the new config format), this is probably good enough for now :)

paasta_tools/cli/cmds/validate.py
```diff
@@ -1272,6 +1273,95 @@ def validate_smartstack(service_path: str) -> bool:
...
+                        continue
+                elif isinstance(monitoring_raw, dict):
+                    monitoring = monitoring_raw
+                else:
+                    continue
+
+                if not isinstance(monitoring, dict):
+                    continue
+
+                team = monitoring.get("team")
```
nemacysts, line 1339: @anohovsky just checking: does this mean that flink doesn't use monitoring.yaml for alerting defaults (i.e., setting a team and whatnot)?

nemacysts (commented):
just have one question re: how flink and monitoring.yaml currently interact - but otherwise this seems fine to me until the flink tools bits in paasta match the actual config and we can switch to doing something that looks more consistent with the other checks! 

(i xargs'd this over soaconfigs and saw no validation failures/tracebacks from this code)

paasta_tools/cli/cmds/validate.py
anohovsky: `monitoring.yaml` and job's config are now a part of flink yelpsoa config (for streamhouse services). There are no `monitoring.yaml` and job's configs in srv-configs repo for these services

anohovsky: example https://github.yelpcorp.com/sysgit/yelpsoa-configs/blob/master/streamhouse_queries/flinkeks-pnw-prod.yaml#L1072

nemacysts: @anohovsky yea, sorry - i meant that the usual standard is that a monitoring.yaml in a service directory (e.g., https://github.yelpcorp.com/sysgit/yelpsoa-configs/blob/master/streamhouse_queries/monitoring.yaml) serves as the default for any instance without an explicit `monitoring:` block (and instances with their own monitoring block essentially have that block overlaid on the default one so that you can, for example, just change a single parameter (e.g., default to paging and set a specific instance to not page)

if flink doesn't follow this behavior, that's slightly surprising (but i guess fine for now) and probably merits a comment in the code here otherwise folks in the future that haven't seen this thread will likely be confused :)

anohovsky: Hi Luis,

I think we’re referring to different monitoring configs 🙂

There are **job-level** and **instance-level** monitoring configs. I’m referring to the former, while I believe you’re referring to the latter.

The Data Processing team wants to define **job ownership** based on the monitoring.team value in the **job monitoring config**. That’s why I’m adding a pre-commit validation to ensure the team name is correct.

nemacysts: yea, i get that 😃 

i think that folks used to the paasta defaults would expect that the monitoring defaults would flow: monitoring.yaml (common) -> instance-level monitoring block (overlaid on/overrides monitoring.yaml) -> job-level monitoring block (overlaid on/overrides instance-level)

i'm suggesting a short comment that mentions that flink doesn't do this so that no one tries to be helpful in the future and tries to clean up code due to an incorrect understanding :p

nemacysts (approved):
(i can ship this for now, but i would encourage adding a small comment at some point that mentions that flink isn't applying the monitoring configs the same way that paasta does (i.e., as overrides) and that job-level monitoring configs are standalone)

anohovsky:
> (i can ship this for now, but i would encourage adding a small comment at some point that mentions that flink isn't applying the monitoring configs the same way that paasta does (i.e., as overrides) and that job-level monitoring configs are standalone)

@nemacysts thank you! I'll add a comment shortly

PR #4322, opened 2026-05-18
remove no-manylinux1 from dh_virtualenv build

benbariteau:
I'm not entirely sure why this was chosen to use, but manylinux is (for better or worse) the standard for python. Practically speaking, we need to use it for some of the dependencies of this project.


nemacysts (approved):
99% sure this was just to match what we were doing internally at the time

PR #4323, opened 2026-05-20
AIRFLOW-3: Add a new set-namespaces subcmd to paasta secret

jfongatyelp:
In #4290  I added an option to add `extra_namespaces` to a secret w/ `paasta secret`, but doing so with an existing secret could only be done with `paasta secret update` which forced users to go through the entire process of re-encrypting the secret. 

This PR instead just adds a new `set-namespaces` subcommand to `secret` so folks can manage *just* the extra_namespaces field w/o needing vault access etc. 

Invocation will look like:
```
$ paasta secret set-namespaces --namespaces temporal,mwaa -s paasta-contract-monitor -n goss
Before: (none)
After:  ['temporal', 'mwaa']
Apply this change to 'goss'?
(y/N)? y
Done. Commit and push /nail/home/jfong/work/yelpsoa-configs/paasta-contract-monitor/secrets/goss.json to apply the update.

$ git diff
diff --git a/paasta-contract-monitor/secrets/goss.json b/paasta-contract-monitor/secrets/goss.json
index 3db848fa687b..0de29363e587 100644
--- a/paasta-contract-monitor/secrets/goss.json
+++ b/paasta-contract-monitor/secrets/goss.json
@@ -48,6 +48,10 @@
         "yelpwifistage": {},
         "yrprod": {}
     },
+    "extra_namespaces": [
+        "temporal",
+        "mwaa"
+    ],
     "key_context": "paasta-contract-monitor",
...

$ paasta secret set-namespaces --namespaces '' -s paasta-contract-monitor -n goss
Before: ['temporal', 'mwaa']
After:  (none)
Apply this change to 'goss'?
(y/N)? y
Done. Commit and push /nail/home/jfong/work/yelpsoa-configs/paasta-contract-monitor/secrets/goss.json to apply the update.

$ git diff 
# no output; back to current version

$ paasta secret set-namespaces --namespaces temporal,mwaa -s paasta-contract-monitor -n goss -Y
Before: (none)
After:  ['temporal', 'mwaa']
Done. Commit and push /nail/home/jfong/work/yelpsoa-configs/paasta-contract-monitor/secrets/goss.json to apply the update.
```

Note that this does a full replace of namespaces with the given --extra-namespaces, it does not append, and the help text reflects that appropriately. For now, this seems fine? Lmk if you think otherwise. 

Because it does a full replace and because we don't have (and I didn't really think it merited it) a command to list current namespaces, the default option is to print out the old + new namespace lists and asks for confirmation.  This can be skipped with `-Y/--yes` (`-y` is reserved already for setting the yelpsoa-configs dir in a shared arg to `paasta secret`)

This PR does not change the support of adding `--extra-namespaces` at either secret creation or update that was already added in #4290

### DIFF
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -282,7 +282,7 @@ def _add_common_args(
         service_group.set_defaults(shared=False)
 
 
-def add_set_namespaces_subparser(subparsers):
+def add_set_namespaces_subparser(subparsers) -> None:
     parser = subparsers.add_parser(
         "set-namespaces",
         help="update the extra_namespaces field of an existing secret without changing its value",
@@ -295,6 +295,8 @@ def add_set_namespaces_subparser(subparsers):
             "changed JSON file to apply the update."
         ),
     )
+    # NOTE: this sub-command is entirely local, so we don't
+    # need to clutter the help-text with vault nonsense
     _add_common_args(parser, include_vault_args=False)
     parser.add_argument(
         "-n",
@@ -460,7 +462,7 @@ def _update_extra_namespaces(secret_path: str, namespaces: List[str]) -> None:
         f.write("\n")
 
 
-def paasta_secret_set_namespaces(args) -> None:
+def paasta_secret_set_namespaces(args: argparse.Namespace) -> None:
     service = SHARED_SECRET_SERVICE if args.shared else args.service
     soa_dir = os.getcwd()
     secret_path = os.path.join(soa_dir, service, "secrets", f"{args.secret_name}.json")
--- a/paasta_tools/cli/cmds/secret.py
+++ b/paasta_tools/cli/cmds/secret.py
@@ -241,7 +241,11 @@ def _add_vault_auth_args(parser: argparse.ArgumentParser):
     )
 
 
-def _add_common_args(parser: argparse.ArgumentParser, allow_shared: bool = True):
+def _add_common_args(
+    parser: argparse.ArgumentParser,
+    allow_shared: bool = True,
+    include_vault_args: bool = True,
+):
     # available from any subcommand
     parser.add_argument(
         "-y",
@@ -251,7 +255,8 @@ def _add_common_args(parser: argparse.ArgumentParser, allow_shared: bool = True)
         default=DEFAULT_SOA_DIR,
     )
 
-    _add_vault_auth_args(parser)
+    if include_vault_args:
+        _add_vault_auth_args(parser)
 
     if allow_shared:
         service_group = parser.add_mutually_exclusive_group(required=True)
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/secret.py
```diff
@@ -277,6 +282,45 @@ def _add_common_args(parser: argparse.ArgumentParser, allow_shared: bool = True)
...
+        description=(
+            "Replaces the extra_namespaces list on an existing PaaSTA secret JSON file "
+            "without touching the encrypted secret value. No Vault authentication required. "
+            "The supplied list completely replaces any previously set namespaces — include all "
+            "namespaces you want, not just new ones. "
+            "Run from the root of your yelpsoa-configs checkout, then commit and push the "
+            "changed JSON file to apply the update."
+        ),
+    )
+    _add_common_args(parser, include_vault_args=False)
```
nemacysts, line 300:
```suggestion
    # NOTE: this sub-command is entirely local, so we don't
    # need to clutter the help-text with vault nonsense
    _add_common_args(parser, include_vault_args=False)
```

paasta_tools/cli/cmds/secret.py
```diff
@@ -415,7 +460,43 @@ def _update_extra_namespaces(secret_path: str, namespaces: List[str]) -> None:
         f.write("\n")
 
 
+def paasta_secret_set_namespaces(args) -> None:
```
nemacysts:
```suggestion
def paasta_secret_set_namespaces(args: argparse.Namespace) -> None:
```
if we want bonus points

paasta_tools/cli/cmds/secret.py
```diff
@@ -277,6 +282,45 @@ def _add_common_args(parser: argparse.ArgumentParser, allow_shared: bool = True)
         service_group.set_defaults(shared=False)
 
 
+def add_set_namespaces_subparser(subparsers):
```
nemacysts: (if we wanted to type this, subparsers should be `argparse._SubParsersAction` - but it's always bothered me to add that as a type to a function signature due to the `_` :p )

nemacysts:
```suggestion
def add_set_namespaces_subparser(subparsers) -> None:
```
(but also, this seems fine if you're as bothered by the aesthetics here as much as i usually am :p)

paasta_tools/cli/cmds/secret.py
```diff
@@ -415,7 +460,43 @@ def _update_extra_namespaces(secret_path: str, namespaces: List[str]) -> None:
...
+    if not os.path.isfile(secret_path):
+        print(
+            f"Secret file not found: {secret_path}\n"
+            "Ensure you are in the root of your yelpsoa-configs checkout "
+            "and the secret already exists (use 'paasta secret add' to create it)."
+        )
+        sys.exit(1)
+
+    _update_extra_namespaces(secret_path, args.extra_namespaces)
+    _log_audit(
```
nemacysts, line 479: observe all the things!

nemacysts (dismissed):

paasta_tools/cli/cmds/secret.py
jfongatyelp: Sorry I keep forgetting to keep things typed! TY for continuing to remind us!

nemacysts (approved):

PR #4325, opened 2026-05-22
Improve `paasta rollback` output

nemacysts:
This generally does 4 things:
* Adds some much needed spacing between messages :p
* Makes it very explict that the rollback succeeded
    * + makes it clear that the rollback will happen in the background
* Provides a way to check if the rollback is done (we should maybe just do this automatically in the future)
* Makes the Git revert reminder match internal messaging

I will probably obviate the wait-for-deployment messaging by having `paasta rollback` automatically do that for you
...but for now i wanted to keep this to just small text tweaks :p

Examples:
Screenshot 2026-05-21 at 3 24 40 PM

Screenshot 2026-05-21 at 3 21 44 PM

### DIFF
--- a/paasta_tools/cli/cmds/rollback.py
+++ b/paasta_tools/cli/cmds/rollback.py
@@ -49,6 +49,11 @@ from paasta_tools.utils import parse_timestamp
 
 DEFAULT_SLACK_CHANNEL = "#deploy"
 
+REVERT_MESSAGE = (
+    "You must also revert the relevant commits in Git, or they will be redeployed on the next push.\n"
+    "See y/rollback-then-revert for more information."
+)
+
 
 def add_subparser(subparsers: argparse._SubParsersAction) -> None:
     list_parser = subparsers.add_parser(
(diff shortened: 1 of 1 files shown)


sidtuladhar (dismissed):

paasta_tools/cli/cmds/rollback.py
```diff
@@ -410,16 +415,17 @@ def paasta_rollback(args: argparse.Namespace) -> int:
             )
             returncode = 1
 
+    # let's add some visual separation between the m-f-d output (which we should maybe considering hiding?)
+    # and any additional output we'll add below
+    print()
     if performed_rollback:
+        print(PaastaColors.bold("\nSuccessfully rolled back!"))
```
EvanKrall: Maybe "Successfully started rollback" or "Started rollback" or "Successfully issued rollback"?

nemacysts:
```suggestion
        print(PaastaColors.bold("\nSuccessfully initiated rollback!"))
```

EvanKrall (approved):

PR #4326, opened 2026-05-26
add yelp.com/cost_owner label, default value specified in service.yaml with instance level overriding possible for paasta instances and tron jobs

sidsatyelp:
## Overview

Add a new Kubernetes pod label `yelp.com/cost_owner` to all pods managed by PaaSTA — both long-running EKS services and Tron jobs. The label value comes from soa-config files (`service.yaml` with instance/action-level overrides). 

This enables cost monitoring systems to attribute compute spend to the correct owning team at the pod level.

There will be a separate set of PRs in yelpsoa-configs that adds this metadata in for all services via recurring batch jobs. Teams will be informed that they can modify this at an instance level.

## Tech spec with all the details
https://docs.google.com/document/d/1jYWkNizn--1Hhmf_DANg9h9wmaZFa4c4SCMMdmu9ku4/edit?tab=t.np4y4uvd0upy#heading=h.kycyvvor8o1u

## Rollout plan

This is disabled by default and gated behind a puppet config. See further details [here](https://docs.google.com/document/d/1jYWkNizn--1Hhmf_DANg9h9wmaZFa4c4SCMMdmu9ku4/edit?tab=t.np4y4uvd0upy#heading=h.pi0efdafniks). 

## Discussion with CI (luisp) about design choices for more context
https://yelp.slack.com/archives/C0ARHDH4WSG/p1777047093823859

## Testing on paasta playground
I modified some of the playground files as seen in the below diff
https://fluffy.yelpcorp.com/i/pGLBqjjRNJwV81QtBLG90HhtfSR5qGSX.html

A couple of additional files to the playground were required
```
cat k8s_itests/deployments/paasta/fake_etc_paasta/cost_owner.json
{
    "enable_cost_owner_label": true
}

cat general_itests/fake_soa_configs_tron/fake_simple_service/service.yaml
---
cost_owner: data-eng
```

Then running the playground I was able to verify that the expected labels showed up on the service pods with the right inheritance

```
╰─ KUBECONFIG=./k8s_itests/kubeconfig kubectl get pods -n paastasvc-compute-infra-test-service -l "paasta.yelp.com/instance=autoscaling,paasta.yelp.com/config_sha=configbcca568a" -o custom-columns='NAME:.metadata.name,COST_OWNER:.metadata.labels.yelp\.com/cost_owner'                                   
NAME                                                      COST_OWNER
compute-infra-test-service-autoscaling-869586ff9f-5dxrt   compute-infrastructure-batch
compute-infra-test-service-autoscaling-869586ff9f-5hjvr   compute-infrastructure-batch
compute-infra-test-service-autoscaling-869586ff9f-5znw2   compute-infrastructure-batch
compute-infra-test-service-autoscaling-869586ff9f-82xnl   compute-infrastructure-batch
compute-infra-test-service-autoscaling-869586ff9f-97xjx   compute-infrastructure-batch
compute-infra-test-service-autoscaling-869586ff9f-bg686   compute-infrastructure-batch
compute-infra-test-service-autoscaling-869586ff9f-hfsxf   compute-infrastructure-batch
compute-infra-test-service-autoscaling-869586ff9f-jx85c   compute-infrastructure-batch
compute-infra-test-service-autoscaling-869586ff9f-p6ms8   compute-infrastructure-batch
compute-infra-test-service-autoscaling-869586ff9f-sx7jt   compute-infrastructure-batch


╰─ KUBECONFIG=./k8s_itests/kubeconfig kubectl get pods -n paastasvc-compute-infra-test-service -l "paasta.yelp.com/instance=cost-owner-override-test" -o custom-columns='NAME:.metadata.name,COST_OWNER:.metadata.labels.yelp\.com/cost_owner'                                                                
NAME                                                              COST_OWNER
compute-infra-test-service-cost-owner-override-test-75f7d5vvscr   compute-infrastructure-services

```

For tron testing it looks like we can't launch pods like paasta instances but the tronfigs are inheriting the right values now. [Full output](https://fluffy.yelpcorp.com/i/DC274MBf5LWVsV3SFwHzJlcdMsgbc6xR.html)

```
╰─ PAASTA_SYSTEM_CONFIG_DIR=./etc_paasta_playground/ .tox/py310-linux/bin/python -m paasta_tools.setup_tron_namespace --dry-run -a --soa-dir general_itests/fake_soa_configs_tron --cluster test-cluster 2>&1 | grep -A 2 "yelp.com/cost_owner"
          yelp.com/cost_owner: ml-team
          yelp.com/owner: compute_infra_platform_experience
        mem: 4096
--
          yelp.com/cost_owner: platform-eng
          yelp.com/owner: compute_infra_platform_experience
        mem: 4096
--
          yelp.com/cost_owner: data-eng
          yelp.com/owner: compute_infra_platform_experience
        mem: 4096
--
          yelp.com/cost_owner: data-eng
          yelp.com/owner: compute_infra_platform_experience
        mem: 4096
```

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -1276,7 +1276,7 @@ Various PaaSTA utilities look at the following keys from service.yaml
  * ``description``
  * ``external_link``
  * ``docker_registry`` This is optional. Set this to override the `system-wide docker registry <system_configs.html#configuration-options>`_, and specify an alternate docker registry for your service.
- * ``cost_owner`` A string identifying the team responsible for the cost of this service. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost attribution. Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be at most 63 characters. Can be overridden per-instance in EKS configs or per-job/action in Tron configs.
+ * ``cost_owner`` A string identifying the team responsible for the cost of this service. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost attribution. Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be at most 63 characters. Can be overridden per-instance in instance configs or per-job/action in Tron configs.
 
 Where does paasta_tools look for yelpsoa-configs?
 -------------------------------------------------------------
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -503,8 +503,9 @@ instance MAY have:
 
   * ``cost_owner``: A string identifying the team responsible for the cost of this
     instance. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost
-    attribution. Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be
-    at most 63 characters. If not specified, inherits from ``service.yaml``.
+    attribution. Should be in kebab-case and match a canonical owner name from the
+    Ownership service. Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$``
+    and be at most 63 characters. If not specified, inherits from ``service.yaml``.
 
   * ``replication_threshold``: An integer representing the percentage of instances that
     need to be available for monitoring purposes. If less than ``replication_threshold``
@@ -729,9 +730,10 @@ Each Tron **job** configuration MAY specify the following options:
 
   * ``cost_owner``: A string identifying the team responsible for the cost of this
     job. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost
-    attribution. Inherited by all actions unless overridden at the action level.
-    Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be at most
-    63 characters.
+    attribution. Should be in kebab-case and match a canonical owner name from the
+    Ownership service. Inherited by all actions unless overridden at the action level.
+    If not specified, inherits from ``service.yaml``. Must match the pattern
+    ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be at most 63 characters.
 
 Each Tron **action** of a job MAY specify the following:
 
@@ -1276,7 +1278,7 @@ Various PaaSTA utilities look at the following keys from service.yaml
  * ``description``
  * ``external_link``
  * ``docker_registry`` This is optional. Set this to override the `system-wide docker registry <system_configs.html#configuration-options>`_, and specify an alternate docker registry for your service.
- * ``cost_owner`` A string identifying the team responsible for the cost of this service. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost attribution. Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be at most 63 characters. Can be overridden per-instance in instance configs or per-job/action in Tron configs.
+ * ``cost_owner`` A string identifying the team responsible for the cost of this service. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost attribution. Should be in kebab-case and match a canonical owner name from the Ownership service. Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be at most 63 characters. Can be overridden per-instance in instance configs or per-job/action in Tron configs.
 
 Where does paasta_tools look for yelpsoa-configs?
 -------------------------------------------------------------
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -501,6 +501,11 @@ instance MAY have:
     to determine the order in which to build & deploy deploy groups. Defaults to
     ``clustername.instancename``. See the deploy group doc_ for more information.
 
+  * ``cost_owner``: A string identifying the team responsible for the cost of this
+    instance. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost
+    attribution. Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be
+    at most 63 characters. If not specified, inherits from ``service.yaml``.
+
   * ``replication_threshold``: An integer representing the percentage of instances that
     need to be available for monitoring purposes. If less than ``replication_threshold``
     percent instances of a service's backends are available, the monitoring
@@ -722,6 +727,12 @@ Each Tron **job** configuration MAY specify the following options:
 
   * ``monitoring``: See the `monitoring.yaml`_ section for details.
 
+  * ``cost_owner``: A string identifying the team responsible for the cost of this
+    job. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost
+    attribution. Inherited by all actions unless overridden at the action level.
+    Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be at most
+    63 characters.
+
 Each Tron **action** of a job MAY specify the following:
 
   * Anything in the `Common Settings`_.
@@ -739,6 +750,9 @@ Each Tron **action** of a job MAY specify the following:
     composed of mixed ``paasta`` and ``ssh`` actions.
 
   * ``deploy_group``: Same setting as the ``Job``, but on a per-action basis. Defaults
+    to the setting for the entire job.
+
+  * ``cost_owner``: Same setting as the ``Job``, but on a per-action basis. Defaults
     to the setting for the entire job.
 
   * ``command``: The command to run. If the action is configured with ``executor: paasta`` (default),
@@ -1262,6 +1276,7 @@ Various PaaSTA utilities look at the following keys from service.yaml
  * ``description``
  * ``external_link``
  * ``docker_registry`` This is optional. Set this to override the `system-wide docker registry <system_configs.html#configuration-options>`_, and specify an alternate docker registry for your service.
+ * ``cost_owner`` A string identifying the team responsible for the cost of this service. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost attribution. Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be at most 63 characters. Can be overridden per-instance in EKS configs or per-job/action in Tron configs.
 
 Where does paasta_tools look for yelpsoa-configs?
 -------------------------------------------------------------
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -28,7 +28,6 @@ from typing import cast
 
 from mypy_extensions import TypedDict
 from service_configuration_lib import read_extra_service_information
-from service_configuration_lib import read_service_configuration
 from service_configuration_lib import read_yaml_file
 from service_configuration_lib.spark_config import SparkConfBuilder
 from service_configuration_lib.spark_config import get_total_driver_memory_mb
@@ -77,6 +76,7 @@ from paasta_tools.utils import PoolsNotConfiguredError
 from paasta_tools.utils import ProjectedSAVolume
 from paasta_tools.utils import SystemPaastaConfig
 from paasta_tools.utils import TronSecretVolume
+from paasta_tools.utils import cached_read_service_configuration
 from paasta_tools.utils import filter_templates_from_config
 from paasta_tools.utils import get_k8s_url_for_cluster
 from paasta_tools.utils import load_system_paasta_config
@@ -732,9 +732,6 @@ class TronActionConfig(InstanceConfig):
         return projected_volumes if projected_volumes else None
 
 
-_cached_read_service_configuration = time_cache(ttl=5)(read_service_configuration)
-
-
 class TronJobConfig:
     """Represents a job in Tron, consisting of action(s) and job-level configuration values."""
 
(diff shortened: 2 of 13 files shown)


docs/source/yelpsoa_configs.rst
```diff
@@ -1262,6 +1276,7 @@ Various PaaSTA utilities look at the following keys from service.yaml
  * ``description``
  * ``external_link``
  * ``docker_registry`` This is optional. Set this to override the `system-wide docker registry <system_configs.html#configuration-options>`_, and specify an alternate docker registry for your service.
+ * ``cost_owner`` A string identifying the team responsible for the cost of this service. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost attribution. Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be at most 63 characters. Can be overridden per-instance in EKS configs or per-job/action in Tron configs.
```
nemacysts:
```suggestion
 * ``cost_owner`` A string identifying the team responsible for the cost of this service. Applied as a ``yelp.com/cost_owner`` Kubernetes pod label for cost attribution. Must match the pattern ``^[a-z0-9]([a-z0-9-]*[a-z0-9])?$`` and be at most 63 characters. Can be overridden per-instance in instance configs or per-job/action in Tron configs.
```

nemacysts: (since i assume you'll want to further extend this for cassandra/nrtsearch/etc)

Yannick-Bellerose (dismissed):
I’m not sure how Yelp-specific we want the code in this repo to be (I do see a lot of Yelp-specific configs in various places). However, if we decide to be more Yelp-specific, we should mention (without enforcing it) in the cost_owner description in docs/source/yelpsoa_configs.rst that it should be in kebab-case and match a canonical owner from Ownership.

nemacysts:
> I’m not sure how Yelp-specific we want the code in this repo to be (I do see a lot of Yelp-specific configs in various places). However, if we decide to be more Yelp-specific, we should mention (without enforcing it) in the cost_owner description in docs/source/yelpsoa_configs.rst that it should be in kebab-case and match a canonical owner from Ownership.

@Yannick-Bellerose we can be pretty Yelp-specific if necessary in the docs - this is more "developing in the open" than "developing something that can be used off-the-shelf elsewhere"

tests/cli/test_cmds_validate.py
```diff
@@ -1941,3 +1941,100 @@ def test_validate_flink_monitoring_team_missing_monitoring():
...
+)
+def test_cost_owner_schema_validation_eks(cost_owner, expected, instance_type, capsys):
+    instance_content = f"""
+test_instance:
+  cost_owner: {cost_owner}
+"""
+    with patch(
+        "paasta_tools.cli.cmds.validate.get_file_contents", autospec=True
+    ) as mock_get_file_contents:
+        mock_get_file_contents.return_value = instance_content
```
nemacysts: nit:
```suggestion
    with patch(
        "paasta_tools.cli.cmds.validate.get_file_contents",
        return_value=instance_content,
        autospec=True
    ) as mock_get_file_contents:
```
should be equivalent iirc

tests/cli/test_cmds_validate.py
```diff
@@ -1941,3 +1941,100 @@ def test_validate_flink_monitoring_team_missing_monitoring():
...
+        assert validate_schema("unused_service_path.yaml", instance_type) == expected
+        expected_output = SCHEMA_VALID if expected else SCHEMA_INVALID
+        output, _ = capsys.readouterr()
+        assert expected_output in output
```
nemacysts, line 1980: !8ball we can get away without checking the output with the combination of parametrized tests we have here

(i usually try to avoid testing the printed output since it's annoying to have to update tests if we change the messaging at some point - although in this case it should be fine)

tests/cli/test_cmds_validate.py
```diff
@@ -1941,3 +1941,100 @@ def test_validate_flink_monitoring_team_missing_monitoring():
...
+        ("compute-infra-batch", True),
+        # invalid: uppercase
+        ("Data-Eng", False),
+        # invalid: starts with hyphen
+        ("-data-eng", False),
+        # invalid: exceeds 63 characters
+        ("a" * 64, False),
+    ],
+)
+@patch("paasta_tools.cli.cmds.validate.get_file_contents", autospec=True)
```
nemacysts: can we swap this to use a context manager to be consistent with the other tests? (same below)

tests/cli/test_cmds_validate.py
nemacysts: (and same below)

tests/test_tron_tools.py
```diff
@@ -2287,3 +2287,222 @@ def test_list_tron_clusters(self, mock_glob):
...
         assert sorted(result) == ["dev-cluster2", "prod"]
+
+
+class TestCostOwnerLabel:
+    @pytest.fixture(autouse=True)
+    def mock_read_monitoring_config(self):
+        with mock.patch(
+            "paasta_tools.monitoring_tools.read_monitoring_config",
+            mock.Mock(return_value={"team": "default_team"}),
+            autospec=None,
```
nemacysts: probably worth a comment for why this is autospec=None :)

tests/test_tron_tools.py
```diff
@@ -2287,3 +2287,222 @@ def test_list_tron_clusters(self, mock_glob):
...
+                },
+            },
+            cluster="test-cluster",
+            load_deployments=False,
+            soa_dir="/mock/soa",
+        )
+        actions = job_config.get_actions()
+        assert actions[0].get_cost_owner() == "ml-team"
+
+    def test_cost_owner_label_in_format_tron_action_dict_when_enabled(self):
```
nemacysts: to make it easier to clean up the gate once this is fully rolled out, i'd probably update MOCK_SYSTEM_PAASTA_CONFIG to have the gate enabled in tests - then we can add the label check to existing tests and only have separate tests (that we'll later delete) for the disabled case

tests/test_tron_tools.py
nemacysts: (we should also probably have a better grouping for tests later since this autouse fixture technically already exists in this file - but that's a future-us problem :p)

tests/test_tron_tools.py
```diff
@@ -2287,3 +2287,222 @@ def test_list_tron_clusters(self, mock_glob):
...
+                    "enable_cost_owner_label": True,
+                    "enable_tron_tsc": True,
+                }
+            ),
+            "/mock/system/configs",
+        )
+        with mock.patch.object(
+            action_config,
+            "get_docker_registry",
+            return_value="docker-registry.com:400",
```
nemacysts, line 2421: is this necessary? seems to go against the value in the mocked system paasta config

docs/source/yelpsoa_configs.rst
```diff
@@ -722,6 +727,12 @@ Each Tron **job** configuration MAY specify the following options:
 
   * ``monitoring``: See the `monitoring.yaml`_ section for details.
 
+  * ``cost_owner``: A string identifying the team responsible for the cost of this
```
nemacysts, line 731: this this one also need a `If not specified, inherits from ``service.yaml``.` mention?

tests/cli/test_cmds_validate.py
```diff
@@ -1941,3 +1941,100 @@ def test_validate_flink_monitoring_team_missing_monitoring():
...
+def test_cost_owner_schema_validation_tron_job(
+    mock_get_file_contents, cost_owner, expected, capsys
+):
+    tron_content = f"""
+test_job:
+  node: batch_box
```
nemacysts: this is going to be a slightly confusing test since nothing that runs on the batch boxes (mostly spark at this point) will have a pod created

also: only spark pods with `executor: spark` will have k8s labels applied to them - and you'll also probably wanna chat with the ML Compute folks to have them pass through the cost owner bits to the executor pods - otherwise, you're just gonna get the label on the spark driver pod

paasta_tools/utils.py
```diff
@@ -956,6 +957,9 @@ def get_role(self) -> Optional[str]:
         """Which mesos role of nodes this job should run on."""
         return self.config_dict.get("role")
 
+    def get_cost_owner(self) -> Optional[str]:
+        return self.config_dict.get("cost_owner")
```
nemacysts, line 965: atm we're not inheriting from service.yaml, right? or am i missing something?

nemacysts (commented):
my only real comments are in the docs + a question about where we're doing the service.yaml inheritance 

my test comments are mostly nits - although, it will probably be easier to write the tests now so that we can just delete the disabled-case ones later rather than having to update existing tests once the toggle is gone (on top of deleting the disabled-case tests)

tests/cli/test_cmds_validate.py
sidsatyelp: Good point. We have a separate milestone to deal with the various operators. I will look into the spark bits then. For now we are only doing pure paasta tron instances

paasta_tools/utils.py
sidsatyelp: I have fixed the inheritance bit now. See the PR description for the test outputs using paasta playground

tests/test_tron_tools.py
```diff
@@ -2287,3 +2288,247 @@ def test_list_tron_clusters(self, mock_glob):
         ]
         result = tron_tools.list_tron_clusters("foo")
         assert sorted(result) == ["dev-cluster2", "prod"]
+
+
+class TestCostOwnerLabel:
+    # autospec=None is used here because we're passing a pre-configured Mock as the
+    # new value rather than letting patch create one with autospec constraints.
```
nemacysts: is this necessary though? 

read_monitoring_config returns a bare dict, so we could also do something like
```
with mock.patch(
    "paasta_tools.monitoring_tools.read_monitoring_config",
    return_value={"team": "default_team"},
    autospec=True,
):
```
if i'm not mistaken

paasta_tools/tron_tools.py
```diff
@@ -828,6 +829,14 @@ def _get_action_config(self, action_name, action_dict) -> TronActionConfig:
         action_deploy_group = action_dict.setdefault(
             "deploy_group", self.get_deploy_group()
         )
+        job_cost_owner = self.config_dict.get("cost_owner")
+        if not job_cost_owner and self.get_service():
+            service_config = read_service_configuration(
+                self.get_service(), soa_dir=self.soa_dir
+            )
+            job_cost_owner = service_config.get("cost_owner")
```
nemacysts: should this be extracted to `TronJobConfig.get_cost_owner()` to match the pattern everywhere else?

paasta_tools/tron_tools.py
```diff
@@ -828,6 +829,14 @@ def _get_action_config(self, action_name, action_dict) -> TronActionConfig:
         action_deploy_group = action_dict.setdefault(
             "deploy_group", self.get_deploy_group()
         )
+        job_cost_owner = self.config_dict.get("cost_owner")
+        if not job_cost_owner and self.get_service():
+            service_config = read_service_configuration(
```
nemacysts: we'll probably also want to do something like https://github.com/Yelp/paasta/blob/e38598b50bb5a23947ce183069aaee538aa6a2c4/paasta_tools/monitoring_tools.py#L155-L159 btw - no use in doing all the file I/O for every job if service.yaml isn't gonna change :p

tests/test_tron_tools.py
```diff
@@ -2287,3 +2288,247 @@ def test_list_tron_clusters(self, mock_glob):
...
+    @pytest.fixture(autouse=True)
+    def mock_read_monitoring_config(self):
+        with mock.patch(
+            "paasta_tools.monitoring_tools.read_monitoring_config",
+            mock.Mock(return_value={"team": "default_team"}),
+            autospec=None,
+        ) as f:
+            yield f
```
nemacysts:
```suggestion
            yield
```

pretty sure we can also yield nothing since this is an autotuse fixture and nothing is actually gonna use the reference to this mock

paasta_tools/tron_tools.py
sidsatyelp: Good call. Fixed it.

paasta_tools/tron_tools.py
sidsatyelp: done

paasta_tools/tron_tools.py
```diff
@@ -823,11 +827,23 @@ def get_cluster(self):
     def get_expected_runtime(self):
         return self.config_dict.get("expected_runtime")
 
+    def get_cost_owner(self) -> Optional[str]:
+        cost_owner = self.config_dict.get("cost_owner")
+        if not cost_owner and self.get_service():
+            service_config = _cached_read_service_configuration(
```
lanceleih: Thanks for changing it to cache :) It would have been fine to read from read_service_configuration, but it's better not having to read the same file over and over again, esp. if there are many actions in a tronjob.

paasta_tools/tron_tools.py
```diff
@@ -731,6 +732,9 @@ def get_projected_sa_volumes(self) -> Optional[List[ProjectedSAVolume]]:
         return projected_volumes if projected_volumes else None
 
 
+_cached_read_service_configuration = time_cache(ttl=5)(read_service_configuration)
```
nemacysts: probably want a comment here for why we're doing this fwiw

(and we can probably move this to utils or something in another PR (unless you'd like the extra credit for refactoring this + the other use of this pattern :p))

nemacysts (dismissed):

lanceleih (dismissed):
lgtm! But I would wait on the paasta folks for the official seal of approval 👍 

One thing I would flag is for the sync job later. We should have a warning in the PR description about how merging the PR could cause a bounce.

paasta_tools/tron_tools.py
nemacysts: hard to tell without running this, but i think without the cache we'd probably see setup_tron_namespace get a bit slower :p

paasta_tools/tron_tools.py
sidsatyelp: Will add a comment for now. I can refactor in a separate PR to keep it small

nemacysts:
> One thing I would flag is for the sync job later. We should have a warning in the PR description about how merging the PR could cause a bounce.

imo, should be fine - the paasta contract makes it clear that paasta can restart you at any time :)

paasta_tools/tron_tools.py
sidsatyelp: Nvm. Did the refactor for this and the monitoring_tools use cases.

nemacysts (approved):

PR #4328, opened 2026-05-28
Automatically rollback after persistent crashlooping

nemacysts:
If we're crashing consistently, something has probably gone wrong and we don't want to keep trying to deploy in the interest of not wasting compute on broken code :)

(and also, so that we don't leave a timebomb after the deployment times out if no one is watching: should this happen, the next deployment would skip the broken deploy group and impact prod)

To enable this, I'm ;iggybacking off of our existing code for diagnosing stuck/slow bounces: after `time_before_first_diagnosis` seconds, we start checking pod restart counts. If all pods have:
1. `restart_count >= min_restarts_for_crashloop_rollback` (new config option, default: 2), AND
2. `restart_count > baseline` (i.e., at least one new restart since we started watching)

...then we immediately trigger the autorollback sequence

The baseline comparison prevents false positives from pods that crashed at startup but recovered — we only roll back if crashes are happening during the diagnosis loop

Once triggered, the only way to exit the autorollback state is to click the Slack buttons in the deploy thread: it's not exactly obvious when/if a Pod has recovered for things like OOMKills as those might take a couple minutes but still be prevalent enough to prevent a deployment from completing (and also: if your code crashed N times already during this deployment, it's broken :p)

NOTE: The tests are all `claude`-authored (which is great 'cause mocking some of the paasta api bits would have been kinda annoying by hand atm :p)

that said, I've also done some manual testing in the paasta playground to ensure that the tests aren't providing me a false sense of security:

(note: `cb9e0385` is a "good" SHA and an actual commit in the compute-infra-test-service repo, and `c1413ea9` only exists on my devbox and is an image that always runs `exit 1`)

With crashloop rollback enabled in deploy.yaml and default diagnosis parameters (5min delay, 60s interval) when marking c1413ea9:

```
Would update the slack thread with: Marked `c1413ea9` for prod.main.
Would update the slack thread with: Will automatically roll back in 30 seconds, (at 14:15:42)! Click "Disable auto rollbacks :close_eyes_monkey:" to cancel this!
Would update the slack thread with: Time's up, will now automatically roll back.
Would update the slack thread with: Marked `cb9e0385` for prod.main.
```

Total time from deploy to rollback: ~6.5 minutes (300s first diagnosis delay + 60s second probe + 30s countdown).

With crashloop rollback disabled, diagnosis parameters tuned down so that i didn't have to wait so long, and the m-f-d wrapped with a `timeout 90`: none of the rollback messages showed up and the command was killed after 90 seconds, but i still saw the diagnosis output show up 

NOTE2: I've refactored this quite a bit from the initial version after a conversation with rofe made me realize there were some edge cases I hadn't considered earlier (spurred on by a question about whether or not this would catch OOMKills and rollback on 'em) :p
...which also made me go ahead and add deploy.yaml configs/toggles for things since this is more opinionated now

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -29,6 +29,7 @@ import socket
 import sys
 import time
 import traceback
+from collections import defaultdict
 from enum import Enum
 from threading import Thread
 from typing import Any
@@ -109,6 +110,7 @@ DEFAULT_WARN_PERCENT = 17  # ~30min for default timeout
 DEFAULT_AUTO_CERTIFY_DELAY = 600  # seconds
 DEFAULT_SLACK_CHANNEL = "#deploy"
 DEFAULT_STUCK_BOUNCE_RUNBOOK = "y/stuckbounce"
+MIN_CONSECUTIVE_CRASHES_FOR_ROLLBACK = 2
 
 
 log = logging.getLogger(__name__)
@@ -496,7 +498,7 @@ def paasta_mark_for_deployment(args: argparse.Namespace) -> int:
     deployment_version = DeploymentVersion(commit, args.image_version)
 
     old_deployment_version = get_currently_deployed_version(
-        service=service, deploy_group=deploy_group
+        service=service, deploy_group=deploy_group, soa_dir=args.soa_dir
     )
     if deployment_version == old_deployment_version:
         print(
@@ -708,8 +710,25 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
             service, deploy_group, soa_dir
         )
 
+        self.crashloop_auto_rollback_enabled = (
+            load_system_paasta_config().get_enable_crashloop_auto_rollback()
+        )
+
         # Keep track of each wait_for_deployment task so we can cancel it.
         self.wait_for_deployment_tasks: Dict[DeploymentVersion, asyncio.Task] = {}
+
+        # this is maybe slightly funky, but we keep track of what instances are crashlooping
+        # (even though we rollback at the deploy group granularity)
+        # and we consider an instance crashlooping if it's been seeing crashing N times
+        # in a row (at the time of writing, N=2).
+        # we do this with a per-instance counter rather than just relying on the state machine
+        # since (atm, at least) we don't have data on how long diagnosis can take - i.e., for
+        # our large monoliths, will the PaaSTA API always return a response fast enough for us
+        # to actually have probed for failures N times?
+        # NOTE: the key here is (cluster, instance_name)
+        self._crashloop_checks_by_instance: Dict[Tuple[str, str], int] = defaultdict(
+            int
+        )
 
         self.human_readable_status = "Waiting on mark-for-deployment to initialize..."
         self.progress = Progress()
@@ -988,6 +1007,17 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
                 "before": self.log_metric_rollback,
             }
             yield {
+                "source": self.rollforward_states,
+                "dest": "start_rollback",
+                "trigger": "rollback_crashloop_failure",
+                "before": self.log_crashloop_rollback,
+            }
+            yield {
+                "source": self.rollback_states,
+                "dest": None,
+                "trigger": "rollback_crashloop_failure",
+            }
+            yield {
                 "source": self.rollback_states,
                 "dest": "start_deploy",
                 "trigger": "forward_button_clicked",
@@ -1074,6 +1104,23 @@ class MarkForDeploymentProcess(RollbackSlackDeploymentProcess):
             "trigger": "metrics_stopped_failing",
             "before": functools.partial(
                 self.cancel_auto_rollback_countdown, "rollback_metric_failure"
+            ),
+        }
+        yield {
+            "source": "*",
+            "dest": None,
+            "trigger": "crashloops_started_failing",
+            "unless": [self.already_rolling_back],
+            "before": functools.partial(
+                self.start_auto_rollback_countdown, "rollback_crashloop_failure"
+            ),
+        }
+        yield {
+            "source": "*",
+            "dest": None,
+            "trigger": "crashloops_stopped_failing",
+            "before": functools.partial(
+                self.cancel_auto_rollback_countdown, "rollback_crashloop_failure"
             ),
         }
         yield {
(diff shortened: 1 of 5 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1550,12 +1642,15 @@ async def periodically_diagnose_instance(
...
     notify_fn: Optional[Callable[[str], None]] = None,
+    crashloop_fn: Optional[Callable[[str, str, bool], None]] = None,
 ) -> None:
     await asyncio.sleep(time_before_first_diagnosis)
     loop = asyncio.get_running_loop()
     while True:
         try:
-            await loop.run_in_executor(
+            # XXX: might be worth extracting the status call so that we don't have to return a value from here
+            # to prevent making multiple potentially expensive status calls
```
nemacysts, line 1654: for context: on y-m, the per-instance status endpoint seems to take 2-10+ seconds per-instance :(

nemacysts: (but i also didn't want to refactor too much at once)

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -708,9 +710,26 @@ def __init__(
...
         self.wait_for_deployment_tasks: Dict[DeploymentVersion, asyncio.Task] = {}
 
+        # this is maybe slightly funky, but we keep track of what instances are crashlooping
+        # (even though we rollback at the deploy group granularity)
+        # and we consider an instance crashlooping if it's been seeing crashing N times
+        # in a row (at the time of writing, N=2).
+        # we do this with a per-instance counter rather than just relying on the state machine
+        # since (atm, at least) we don't have data on how long diagnosis can take - i.e., for
+        # our large monoliths, will the PaaSTA API always return a response fast enough for us
+        # to actually have probed for failures N times?
```
nemacysts: answer: atm, this is gonna be pretty painful for y-m given the # of instances and how long it takes to get a single instances status

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -708,9 +710,26 @@ def __init__(
             service, deploy_group, soa_dir
         )
 
+        self.crashloop_auto_rollback_enabled = (
+            load_system_paasta_config().get_enable_crashloop_auto_rollback()
+        )
```
nemacysts: assuming nothing blows up, this should be deleted pretty quickly since this really only makes sense for jenkins

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1457,6 +1515,38 @@ def log_metric_rollback(self) -> None:
...
+    def log_crashloop_rollback(self) -> None:
+        self.rollback_type = RollbackTypes.AUTOMATIC_CRASHLOOP_ROLLBACK
+        rollback_details = self.__build_rollback_audit_details(
+            RollbackTypes.AUTOMATIC_CRASHLOOP_ROLLBACK
+        )
+        self._log_rollback(rollback_details)
+
+    def any_crashloop_failing(self) -> bool:
+        return any(
+            # XXX: make this configureable in deploy.yaml?
```
cuza:
```suggestion
            # XXX: make this configurable in deploy.yaml?
```

mbrankin-art (dismissed):

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -109,6 +110,7 @@ class DeployOutcome(Enum):
 DEFAULT_AUTO_CERTIFY_DELAY = 600  # seconds
 DEFAULT_SLACK_CHANNEL = "#deploy"
 DEFAULT_STUCK_BOUNCE_RUNBOOK = "y/stuckbounce"
+MIN_CONSECUTIVE_CRASHES_FOR_ROLLBACK = 2
```
ilkinmammadzada: Can we make this configurable per cluster for now?

nemacysts: yea, i can do that!

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1667,6 +1783,21 @@ def should_ping_for_pod(pod: KubernetesPodV2) -> bool:
...
+    status: InstanceStatusKubernetesV2,
+) -> bool:
+    pods = [
+        pod
+        for v in status.kubernetes_v2.versions
+        if v.git_sha == version.sha and v.image_version == version.image_version
+        for pod in v.pods
+    ]
+    return len(pods) > 0 and all(
+        recent_container_restart(get_main_container(pod)) for pod in pods
```
nemacysts: i just realized that this looks at a pretty wide range of time (last 900s - i'll need to pass a smaller `time_window` to `recent_container_restart()` as otherwise we're gonna potentially rollback unnecessarily

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -1433,6 +1484,12 @@ def deploy_group_is_set_to_notify(self, notify_type: str) -> bool:
             self.deploy_info, self.deploy_group, notify_type
         )
 
+    def _get_deploy_group_config(self, key: str, default: Any) -> Any:
+        for step in self.deploy_info.get("pipeline", []):
+            if step.get("step", "") == self.deploy_group:
+                return step.get(key, default)
+        return default
```
nemacysts, line 1491: this is kind of unfortunate - we sidestep this for the current autorollbacks feature by passing this down from jenkins - but that seems like a pain to do since now i'd have to couple things to our jenkinsfile_stdlib code

paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -708,9 +708,27 @@ def __init__(
...
+            system_paasta_config.get_enable_crashloop_auto_rollback(),
+        )
+        self.min_restarts_for_crashloop_rollback = self._get_deploy_group_config(
+            "min_restarts_for_crashloop_rollback",
+            system_paasta_config.get_min_restarts_for_crashloop_rollback(),
+        )
+        self.crashloop_rollback_percentage_threshold = self._get_deploy_group_config(
+            "crashloop_rollback_percentage_threshold",
+            system_paasta_config.get_crashloop_rollback_percentage_threshold(),
+        )
```
nemacysts, line 723: i.e., use SystemPaastaConfig unless there's a value overriding that in deploy.yaml

ilkinmammadzada (approved):

PR #4329, opened 2026-06-02
feat: use raw metrics over recording rule rollout

sidtuladhar:
[PAASTA-18886](https://jira.yelpcorp.com/browse/PAASTA-18886)

The worker-load HPA formula uses a `missing_instances` term (pods with no worker_busy data, assumed at 100% load) to account for pods that haven't reported metrics yet. 

When pods disappear, `worker_busy` drops immediately but the recording rule `k8s:deployment:pods_status_ready` goes stale so pods start to appear missing and are assumed 100% load, causing the HPA to scale up. 

This PR switches to raw kube-state-metric metrics so both sides reflect current state, which prevents any staleness and the HPA doesn't create phantom pods.


paasta_tools/utils.py
```diff
@@ -2820,6 +2821,9 @@ def get_unhealthy_pod_eviction_policy(self) -> str:
         """
         return self.config_dict.get("unhealthy_pod_eviction_policy", "IfHealthyBudget")
 
+    def get_use_raw_metric(self) -> bool:
```
nemacysts: (same as ilkin's comment in the puppet PR - would be good to include "hpa" in the name of this somewhere :))

tests/test_setup_prometheus_adapter_config.py
```diff
@@ -170,7 +170,14 @@ def test_create_instance_uwsgi_v2_scaling_rule() -> None:
     assert rule["name"]["as"].endswith("-uwsgi-v2-prom")
 
 
-def test_create_instance_worker_load_scaling_rule() -> None:
+@mock.patch(
+    "paasta_tools.setup_prometheus_adapter_config.load_system_paasta_config",
+    autospec=True,
+    return_value=MOCK_SYSTEM_PAASTA_CONFIG,
+)
```
nemacysts: we'll want to do this as a context manager rather than as a decorator to be consistent (we usually only do the decorator form in old files) - the decorator form is slightly more painful to use as a human since it makes you think about the order of the decorator + the order of the arguments in the function signature :p

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -589,26 +590,41 @@ def create_instance_worker_load_scaling_rule(
...
+        """
     load_per_instance = f"""
         avg(
             worker_busy{{{worker_filter_terms}}}
         ) by (kube_pod, kube_deployment)
     """
     missing_instances = f"""
         clamp_min(
-            {ready_pods} - count({load_per_instance}) by (kube_deployment),
+        {ready_pods} - count({load_per_instance}) by (kube_deployment),
```
nemacysts: (the whitespace here was nice for readability)

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -589,18 +590,42 @@ def create_instance_worker_load_scaling_rule(
...
+    use_raw_metric = paasta_system_config.get_use_raw_metric_for_hpa()
+
+    if use_raw_metric:
+        ksm_filter_terms = f"paasta_cluster='{paasta_cluster}'"
+        raw_ready_pods = f"""
+            sum by (kube_deployment) (
+                label_replace(label_replace(label_replace(
+                    kube_pod_status_ready{{condition="true",{ksm_filter_terms}}}
+                    * on (pod, paasta_cluster, namespace) group_left (label_paasta_yelp_com_instance, label_paasta_yelp_com_service)
+                    kube_pod_labels{{label_paasta_yelp_com_service="{service}",label_paasta_yelp_com_instance="{instance}",{ksm_filter_terms}}}
```
ilkinmammadzada: Can we create a variable for `label_paasta_yelp_com_service="{service}",label_paasta_yelp_com_instance="{instance}",{ksm_filter_terms}` similar to other usages

ilkinmammadzada (approved):

sidtuladhar:
🥳

PR #4330, opened 2026-06-10
Support paasta restart for Flink instances

tub:
## Summary
- Removes the "paasta restart is currently unsupported for Flink instances" error
- Sends `desired_state="restart"` via the existing `instance_set_state` API
- The flink-operator handles the full stop→start cycle atomically (no client polling needed)
- Adds 'restart' desired state to paasta status output

Refactors the flink paasta status output to be less verbose:
- Moves long URLs to hyperlinks
- Removes redundant "flink" everywhere
- Updates unit tests to be less brittle, checking for specific functionality rather than exact verbatim output

image

## Depends on
- [x] Internal PR for flink operator-side restart support

## Test plan
- [x] Existing unit tests pass (no behavior change for non-Flink instances)
- [x] Manual test: `paasta restart -s <flink-service> -i <instance> -c <cluster>` triggers operator restart

### DIFF
--- a/paasta_tools/cli/cmds/start_stop_restart.py
+++ b/paasta_tools/cli/cmds/start_stop_restart.py
@@ -181,6 +181,11 @@ def print_flink_message(desired_state):
             "'Stop' will put Flink cluster in stopping mode, it may"
             "take some time before shutdown is completed."
         )
+    elif desired_state == "restart":
+        print(
+            "'Restart' will tell the Flink operator to stop and start the cluster. "
+            "This may take some time to complete."
+        )
 
 
 def confirm_to_continue(cluster_service_instances, desired_state):
@@ -304,10 +309,15 @@ def paasta_start_or_stop(args, desired_state):
                         print_kubernetes_message(desired_state)
                         kubernetes_message_printed = True
 
+                    # For git-tag-based services, "restart" is just "start" with
+                    # a new force_bounce timestamp — deployd only knows start/stop.
+                    git_desired_state = (
+                        "start" if desired_state == "restart" else desired_state
+                    )
                     issue_state_change_for_service(
                         service_config=service_config,
                         force_bounce=force_bounce,
-                        desired_state=desired_state,
+                        desired_state=git_desired_state,
                     )
 
     return_val = 0
@@ -340,7 +350,7 @@ def paasta_start_or_stop(args, desired_state):
                 )
             except client.api_error as exc:
                 print(exc.reason)
-                return exc.status
+                return 1
 
             return_val = 0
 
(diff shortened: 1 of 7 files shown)


paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -382,35 +382,40 @@ def paasta_restart(args):
...
-        if not affected_non_flinks:
-            return 1
+            try:
+                client.service.instance_set_state(
+                    service=service,
+                    instance=instance,
+                    desired_state="restart",
+                )
+            except client.api_error as exc:
+                print(exc.reason)
```
nemacysts: is this an intentional print or a leftover debugging one?

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -382,35 +382,40 @@ def paasta_restart(args):
...
+                return exc.status
```
nemacysts: would this return an HTTP status code?

nemacysts: (which i think is fine, but my understanding is that the actual reported range is up to ~256 and anything outside that gets wrapped)

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -382,35 +382,40 @@ def paasta_restart(args):
...
-        )
-        print("To restart, please run:", end="\n\n")
-        for flink in affected_flinks:
-            print(
-                f"paasta stop -s {flink.service} -i {flink.instance} -c {flink.cluster}"
-            )
-            print(
-                f"paasta start -s {flink.service} -i {flink.instance} -c {flink.cluster}",
-                end="\n\n",
+        system_paasta_config = load_system_paasta_config()
```
nemacysts: should all this be in `paasta_start_or_stop()` instead?

paasta_tools/cli/cmds/status.py
```diff
@@ -822,12 +817,15 @@ def _print_flink_status_from_job_manager(
...
     labels = metadata.get("labels", {})
+    annotations = metadata.get("annotations", {})
     config_sha = labels.get(paasta_prefixed("config_sha"), "").removeprefix("config")
+    desired_state = annotations.get(paasta_prefixed("desired_state"))
     output.append(f"    Config SHA: {config_sha}")
 
     # Initialize all vars upfront; populated only when cluster is running
     dashboard_url = None
     overview = None
+    instance_details: Optional[flink_tools.FlinkInstanceDetails] = None
```
nemacysts: for consistency, we should probably `from paasta_tools.flink_tool import FlinkInstanceDetails` - i think we mostly just use `flink_tools.X` for functions since those might collide with other workload functions

tests/test_flink_tools.py
```diff
@@ -703,6 +702,30 @@ def test_taskmanagers_slots_line(self):
         assert "1 taskmanagers" in slots_line
         assert "2/8 slots available" in slots_line
 
+    def test_restart_desired_state_shown(self):
+        from paasta_tools.utils import PaastaColors
```
nemacysts: hmm, i wonder why claude didn't see the guidance in our CLAUDE.md to not do inline imports :(

paasta_tools/utils.py
```diff
@@ -1200,6 +1200,13 @@ def grey(text: str) -> str:
     def default(text: str) -> str:
         return PaastaColors.color_text(PaastaColors.DEFAULT, text)
 
+    @staticmethod
+    def terminal_link(url: str, label: str) -> str:
+        """Format an OSC 8 terminal hyperlink (clickable in supported terminals)."""
+        if os.getenv("NO_COLOR", "0") == "1":
+            return f"{label} ({url})"
+        return f"\033]8;;{url}\033\\{label}\033]8;;\033\\"
```
nemacysts, line 1208: fancy!

paasta_tools/flink_tools.py
```diff
@@ -503,35 +507,54 @@ def format_flink_udf_info(config: FlinkDeploymentConfig) -> List[str]:
...
     return output
 
 
 def format_flink_config_links(service: str, ecosystem: str) -> List[str]:
-    """Format configuration repository links."""
+    """Format configuration repository links.
+
+    Deprecated: use format_flink_instance_metadata which now includes config links.
+    Kept for backwards compatibility.
+    """
```
nemacysts: we control all this code and don't really generally support folks using paasta as a library - so we don't need to worry about backwards compatibility :)

nleigh:
@tub Thanks for refactoring, lgtm (after addressing feedback from Luis)

nleigh (dismissed):

tub:
@nemacysts I've addressed the comments, thanks for the feedback, it's much cleaner now. 
It has the nice side-effect of re-using the multiple service/instance check too.

Working locally:

```
➜ .tox/bin/paasta restart -s acorn -c pnw-devc
You are about to restart the following instances:
Either --instances or --clusters not specified. Asking for confirmation.
cluster = pnw-devc, instance = main
Are you sure you want to restart these 1 instances?
(y/N)? y
'Restart' will tell the Flink operator to stop and start the cluster. This may take some time to complete.
      
➜ .tox/py310-linux/bin/paasta status  -s acorn -c pnw-devc

acorn.main in pnw-devc (EKS)
    Version:    3825bbf8 (desired)
    Config SHA: e6d31473
    State: Stoppingsupervisor (desired: restart)
    Pods: 5 running, 0 evicted, 0 other, 5 total
    Jobs:
      Job Name   State       Started
```

paasta_tools/cli/cmds/start_stop_restart.py
```diff
@@ -304,10 +309,15 @@ def paasta_start_or_stop(args, desired_state):
                         print_kubernetes_message(desired_state)
                         kubernetes_message_printed = True
 
+                    # For git-tag-based services, "restart" is just "start" with
+                    # a new force_bounce timestamp — deployd only knows start/stop.
```
nemacysts: deployd hasn't been a thing for a while :)

(and technically, all PaaSTA services are git-tag based)

i'd maybe update this to something like:

```suggestion
                    # "restart" is just "start" with
                    # a new force_bounce timestamp — paasta only knows start/stop and will restart since force_bounce has changed
```

nemacysts (commented):
(one last comment re: a comment)

paasta_tools/cli/cmds/start_stop_restart.py
tub: Urgh, sorry, I kept correcting the agent but deployd kept creeping in & giving me flashbacks to 2017.

nemacysts (approved):

PR #4334, opened 2026-06-16
Flink: add y/ links and remove dead is_eks branching

tub:
## Summary
- Replace verbose GitHub/Grafana/Sourcegraph URLs in Flink status output with shorter `y/` redirect links that work in all terminals
- Remove the `is_eks` parameter from `get_flink_ingress_url_root` and its callers — all Flink clusters run on EKS now, so the k8s/eks URL branching was dead code
- Fix uncaught exceptions in `curl_flink_endpoint` that caused intermittent 500s from the PaaSTA API when fetching Flink config/overview/jobs

## Bug fix details
`curl_flink_endpoint` caught `ApiException` from `paasta_tools.paastaapi.exceptions`, but `get_cr()` raises `kubernetes.client.rest.ApiException` — a completely different class. When the Kubernetes API had transient errors (503, timeouts, rate limiting), the exception propagated unhandled through the Pyramid view, causing a generic HTML 500 response.

Also added a `KeyError` handler for cases where the Flink CR exists but is missing the expected `flink.yelp.com/dashboard_url` annotation or `paasta.yelp.com/cluster` label.

## Test plan
- [x] `make test` passes (pre-commit, mypy, pytest)
- [x] Verify `paasta status` for a Flink instance shows correct y/ links
- [x] Confirm Flink health checks still resolve the ingress URL correctly
- [x] New tests verify KubeApiException and missing annotation are caught

----

`paasta status`:
image

`paasta status -v`:
image

### DIFF
--- a/paasta_tools/flink_tools.py
+++ b/paasta_tools/flink_tools.py
@@ -248,15 +248,12 @@ def cr_id(service: str, instance: str) -> Mapping[str, str]:
     )
 
 
-def get_flink_ingress_url_root(cluster: str, is_eks: bool) -> str:
-    if is_eks:
-        return f"http://flink.eks.{cluster}.paasta:{FLINK_INGRESS_PORT}/"
-    else:
-        return f"http://flink.k8s.{cluster}.paasta:{FLINK_INGRESS_PORT}/"
+def get_flink_ingress_url_root(cluster: str) -> str:
+    return f"http://flink.eks.{cluster}.paasta:{FLINK_INGRESS_PORT}/"
 
 
-def _dashboard_get(cr_name: str, cluster: str, path: str, is_eks: bool) -> str:
-    root = get_flink_ingress_url_root(cluster, is_eks)
+def _dashboard_get(cr_name: str, cluster: str, path: str) -> str:
+    root = get_flink_ingress_url_root(cluster)
     url = f"{root}{cr_name}/{path}"
     response = requests.get(url, timeout=FLINK_DASHBOARD_TIMEOUT_SECONDS)
     response.raise_for_status()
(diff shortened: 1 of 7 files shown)


nleigh (dismissed):

paasta_tools/flink_tools.py
```diff
@@ -512,25 +509,21 @@ def format_flink_instance_metadata(
     """Format verbose instance metadata (repo links, pool, owner, runbook, configs)."""
     output: List[str] = []
     output.append("    Links:")
-    git_url = f"https://github.yelpcorp.com/services/{service}"
-    sg_url = f"https://sourcegraph.yelpcorp.com/services/{service}"
+    sg_url = f"y/service-sg/{service}"
     output.append(
-        f"      Repo:       {PaastaColors.terminal_link(git_url, 'y/github')}"
-        f" | {PaastaColors.terminal_link(sg_url, 'sourcegraph')}"
+        f"      Code:       {PaastaColors.terminal_link('http://'+sg_url, sg_url)}"
```
nemacysts: should we set the protocol when we set the variable? the string concatenation otherwise looks a little funky here (and if we want, maybe we f-string this?)

nemacysts: (and same below)

tests/test_flink_tools.py
```diff
@@ -291,6 +291,29 @@ def test_curl_flink_endpoint_get_job_checkpoints(
     }
 
 
+@mock.patch("paasta_tools.flink_tools.get_cr", autospec=True)
+def test_curl_flink_endpoint_kube_api_exception(mock_get_cr):
+    from kubernetes.client.rest import ApiException as KubeApiException
```
nemacysts: this should be a top-level import

paasta_tools/flink_tools.py
tub: Those were the only bits I did by hand 😆

nemacysts (approved):

PR #4335, opened 2026-06-16
Add prometheus-adapter rule generation for nrtsearch GPU autoscaling

vim345:
## Summary

Teaches `setup_prometheus_adapter_config.py` to read nrtsearch CRD manifests and generate prometheus-adapter rules for GPU-based autoscaling.

  ## Context

The nrtsearch-operator manages its own HPA (not through PaaSTA's autoscaling framework), so the existing `get_rules_for_service_instance` path doesn't apply. However, the operator's HPA still needs the prometheus-adapter to serve its custom metric.

 This adds a new function `get_rules_for_nrtsearch_gpu()` that:
  1. Reads `{soa_dir}/nrtsearch/nrtsearchserviceeks-{cluster}.yaml`
  2. Finds any serverSet with `targetGpuUtilization > 0`
  3. Generates an adapter rule that joins `DCGM_FI_DEV_GPU_UTIL` with `kube_pod_labels` to associate GPU utilization with nrtsearch pods
  4. Divides by the target so the HPA sees 1.0 when at target utilization

  The generated metric name follows the existing convention: `{deployment-name}-arbitrary_promql-prom`.

  ## Test plan

  - [x] New unit tests: rule generation, missing file handling, no-gpu-config skip

### DIFF
--- a/paasta_tools/setup_prometheus_adapter_config.py
+++ b/paasta_tools/setup_prometheus_adapter_config.py
@@ -289,7 +289,9 @@ def create_instance_scaling_rule(
 
 def create_instance_active_requests_scaling_rule(
     service: str,
-    instance_config: KubernetesDeploymentConfig,
+    instance_config: Union[
+        KubernetesDeploymentConfig, NrtsearchServiceEksDeploymentConfig
+    ],
     metrics_provider_config: MetricsProviderDict,
     paasta_cluster: str,
     metric_name: str,
@@ -399,7 +401,9 @@ def create_instance_active_requests_scaling_rule(
 
 def create_instance_uwsgi_scaling_rule(
     service: str,
-    instance_config: KubernetesDeploymentConfig,
+    instance_config: Union[
+        KubernetesDeploymentConfig, NrtsearchServiceEksDeploymentConfig
+    ],
     metrics_provider_config: MetricsProviderDict,
     paasta_cluster: str,
     metric_name: str,
@@ -501,7 +505,9 @@ def create_instance_uwsgi_scaling_rule(
 
 def create_instance_uwsgi_v2_scaling_rule(
     service: str,
-    instance_config: KubernetesDeploymentConfig,
+    instance_config: Union[
+        KubernetesDeploymentConfig, NrtsearchServiceEksDeploymentConfig
+    ],
     metrics_provider_config: MetricsProviderDict,
     paasta_cluster: str,
     metric_name: str,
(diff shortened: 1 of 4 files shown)


paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -1010,6 +1108,9 @@ def create_prometheus_adapter_config(
                     )
                 )
 
+    # Generate rules for nrtsearch GPU autoscaling from CRD manifests
+    rules.extend(get_rules_for_nrtsearch_gpu(paasta_cluster, soa_dir))
```
nemacysts: i think we can probably do this without having to add a special case function for nrtsearch

we'd want to update `K8S_INSTANCE_TYPE_CLASSES` so that https://github.com/Yelp/paasta/blob/85e76d744dee3715ac3281bd5391ec566757ff2e/paasta_tools/setup_prometheus_adapter_config.py#L971 covers nrtsearch configs

then, we probably need to figure out some typing shenanigans for `instance_config` in `get_rules_for_service_instance()`, but then as long as we implement/have implemented all the `instance_config.FOO()` methods we're calling here, we might not need to do much else re: making this work for nrtsearch?

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -964,6 +964,104 @@ def get_rules_for_service_instance(
...
+            if not autoscaling:
+                continue
+            target_gpu_utilization = autoscaling.get("targetGpuUtilization", 0)
+            if target_gpu_utilization <= 0:
+                continue
+
+            is_primary = server_set.get("primary", False)
+            instance_name = "primary" if is_primary else "replica"
+            deployment_name = f"{cluster_name}-{instance_name}-dep"
+            namespace = "paastasvc-nrtsearch"
```
nemacysts: aren't these going to be in `paasta-nrtsearchservices`?

paasta_tools/setup_prometheus_adapter_config.py
nemacysts: especially since serverSets should only have 2 items and only the non-primary is autoscalable, right? ('cause then we can essentially have any of the methods that look up autoscaling data just do that for the non-primary :D)

vim345: When `series_query` is provided in the `MetricsProviderDict`, the `create_instance_arbitrary_promql_scaling_rule` path skips the `label_replace()` wrapping and uses the raw `metrics_query/series_query` directly. Is that sufficient for the prometheus-adapter to correctly associate the metric with the target deployment, or do we still need the `label_replace` for deployment/namespace labels in the metrics_query?

PR #4338, opened 2026-06-24
STREAMAPP-15217: Add requireSSD to FlinkDeploymentConfigDict

nleigh:
## Summary

Documents the `requireSSD` Flink config key in the typed config dict. When `True`, the Flink instance's pods are pinned to instance-store (local NVMe) nodes; when `False`/unset, they may run on instance-store or EBS-backed nodes, letting Karpenter cost-optimize across both within a shared pool.

Part of the "one pool + `requireSSD`" capacity-diversification design (STREAMAPP-15217).

Jira: https://jira.yelpcorp.com/browse/STREAMAPP-15217

## Why this is small

No runtime logic is needed in paasta-tools: `setup_kubernetes_cr.format_custom_resource` sets `"spec": instance_config`, passing the yelpsoa config dict straight through as the Flink CR spec. So `requireSSD` reaches the flink-operator unchanged, where it is translated into a node affinity on `karpenter.k8s.aws/instance-local-nvme`. This change just makes the key type-visible and documented in `FlinkDeploymentConfigDict`.

Note: Flink configs have no JSON schema under `cli/schemas/` and `yelpsoa_configs.rst` has no Flink section, so the typed dict comment is the in-repo home for this field's docs.

## Dependency

Pairs with **flink-operator [#219](https://github.yelpcorp.com/services/flink-operator/pull/219)**, which implements the actual affinity. Until that lands, the key is a no-op.

## Test plan

- [x] `mypy paasta_tools/flink_tools.py` clean
- [x] pre-commit (black/flake8/isort/...) pass
- [x] `pytest tests/test_flink_tools.py` — 51 passed


paasta_tools/flink_tools.py
```diff
@@ -116,6 +116,12 @@ class FlinkInstanceDetails(TypedDict):
 class FlinkDeploymentConfigDict(LongRunningServiceConfigDict, total=False):
     taskmanager: TaskManagerConfig
     spot: bool
+    # When True, pin this Flink's pods to instance-store (local NVMe) nodes.
+    # When False/unset, pods may run on instance-store or EBS-backed nodes,
+    # letting Karpenter cost-optimize across both within a shared pool. Passed
+    # through to the Flink CR spec and enforced by the flink-operator as a node
+    # affinity on karpenter.k8s.aws/instance-local-nvme; see STREAMAPP-15217.
```
nleigh:
```suggestion
```

paasta_tools/flink_tools.py
```diff
@@ -116,6 +116,7 @@ class FlinkInstanceDetails(TypedDict):
 class FlinkDeploymentConfigDict(LongRunningServiceConfigDict, total=False):
     taskmanager: TaskManagerConfig
     spot: bool
+    requireSSD: bool
```
nemacysts:
```suggestion
    require_ssd: bool
```

sorry, was just yolo-typing in slack - this should be snake_case for consistency :p

nemacysts (commented):
(lgtm other than the casing suggestion - i'm also assuming we don't need to plumb this through anywhere else and that we just ship the entire config dict to the operator)

paasta_tools/flink_tools.py
nleigh: @nemacysts updated to snake_case here and in the other PR's

nemacysts (approved):

PR #4339, opened 2026-06-24
Handle malformed Flink job details in status output

0dunay0:
## Summary
- skip Flink job detail entries that do not contain required display fields (`name`, `start_time`) before table width and grouping logic
- prevent `paasta status` from crashing when job-details responses deserialize to empty objects
- add a regression test covering a malformed job detail (`{}`) alongside a valid job

## Problem
`paasta status` for Flink can crash with:

```py
ApiAttributeError: FlinkJobDetails has no attribute 'name'
```

This happens when the jobs list includes a stale job id and the job-details endpoint returns an error payload (e.g. 404 wrapper) that deserializes to `{}`.

## Repro
```bash
paasta status -c pnw-prod -s datapipe_s3_flink_processor -i bunsen
```

## Testing
- runtime sanity check of `format_flink_jobs_table([{}, valid_job], ...)` confirms no crash and valid row output
- added unit test: `test_skips_jobs_missing_required_display_fields`

### DIFF
--- a/tests/test_flink_tools.py
+++ b/tests/test_flink_tools.py
@@ -821,6 +821,16 @@ class TestFormatFlinkJobsTable:
         output_text = "\n".join(result)
         assert "None" not in output_text
 
+    def test_skips_jobs_missing_required_display_fields(self, mock_terminal_size):
+        mock_terminal_size.return_value = mock.Mock(columns=120)
+        malformed_job = {}
+        valid_job = self._make_job({"jid": "good123", "name": "beam.main.good_job"})
+        result = flink_tools.format_flink_jobs_table(
+            [malformed_job, valid_job], "http://dashboard", 0
+        )
+        output_text = "\n".join(result)
+        assert "good_job" in output_text
+
 
 class TestFormatFlinkUdfInfo:
     def _make_config(self, extra_config: dict) -> FlinkDeploymentConfig:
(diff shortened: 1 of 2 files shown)


tests/test_flink_tools.py
```diff
@@ -821,6 +821,16 @@ def test_no_dashboard_url(self, mock_terminal_size):
...
 
+    def test_skips_jobs_missing_required_display_fields(self, mock_terminal_size):
+        mock_terminal_size.return_value = mock.Mock(columns=120)
+        malformed_job = {}
+        valid_job = self._make_job({"jid": "good123", "name": "beam.main.good_job"})
+        result = flink_tools.format_flink_jobs_table(
+            [malformed_job, valid_job], "http://dashboard", 0
+        )
+        output_text = "\n".join(result)
+        assert "good_job" in output_text
```
abareghi, line 835: It's good to assert malformed_job is filter out, maybe by asserting result size == 1 
Also maflformed_job var name is a bit confusing. It's empty job and not malformed. maybe either rename it or better to have two vars: 1- empty_job 2- malfolmed_job which doesn't have required fields, but has other fields

0dunay0: Fixed

abareghi (approved):

PR #4341, opened 2026-06-30
feat: shared prometheus adapter rules

sidtuladhar:
[COMPINFRA-5816](https://jira.yelpcorp.com/browse/COMPINFRA-5816)

k8s-prometheus-adapter configmap is very close to the limit! configmap size is around 750kb+! 

we reduce the size by instead of having a query per instance, we can have find the share queries among HPAs since a lot of them have the same query.

  - Before: 634 rules, one per HPA, each with the service/instance baked into the PromQL -> 770kb
  - After: 51 rules, one per unique (autoscaling_type, moving_average_window) -> 45kb


tests/test_setup_prometheus_adapter_config.py
```diff
@@ -390,6 +410,199 @@ def test_get_rules_for_service_instance(
...
+    moving_average_window = 20120302
+    rule = create_shared_gunicorn_scaling_rule(paasta_cluster, moving_average_window)
+
+    assert LABEL_MATCHERS in rule["metricsQuery"]
+    assert str(moving_average_window) in rule["metricsQuery"]
+    assert paasta_cluster in rule["seriesQuery"]
+    assert "gunicorn_worker_busy" in rule["seriesQuery"]
+    assert rule["name"]["as"] == f"gunicorn-prom-{moving_average_window}"
+    assert "paasta_service=" not in rule["metricsQuery"]
+    assert "paasta_instance=" not in rule["metricsQuery"]
```
nemacysts: might be worth using `pytest.mark.parametrize` with something like 
```
@pytest.mark.parametrize(
    "create_func,expected_series_metric,expected_name_prefix",
    [
        (create_shared_worker_load_scaling_rule, "worker_busy", "worker-load-prom"),
        (create_shared_uwsgi_v2_scaling_rule, "uwsgi_worker_busy", "uwsgi-v2-prom"),
        (create_shared_uwsgi_scaling_rule, "uwsgi_worker_busy", "uwsgi-prom"),
        (create_shared_piscina_scaling_rule, "piscina_pool_utilization", "piscina-prom"),
        (create_shared_gunicorn_scaling_rule, "gunicorn_worker_busy", "gunicorn-prom"),
    ]
)
def test_create_shared_scaling_rule(rule_func, expected_series_metric, expected_name_prefix) -> None:
    paasta_cluster = "test_cluster"
    moving_average_window = 20120302

    rule = rule_func(paasta_cluster, moving_average_window)

    assert LABEL_MATCHERS in rule["metricsQuery"]
    assert str(moving_average_window) in rule["metricsQuery"]
    assert paasta_cluster in rule["seriesQuery"]
    assert expected_series_metric in rule["seriesQuery"]
    assert rule["name"]["as"] == f"{expected_name_prefix}-{moving_average_window}"
    # if service/instance are provided, we're not compressing the amount of output :p
    assert "paasta_service=" not in rule["metricsQuery"]
    assert "paasta_instance=" not in rule["metricsQuery"]
```
since all the test bodies are essentially the same thing :)

tests/test_setup_prometheus_adapter_config.py
```diff
@@ -390,6 +410,199 @@ def test_get_rules_for_service_instance(
...
+)
+@mock.patch(
+    "paasta_tools.setup_prometheus_adapter_config.get_services_for_cluster",
+    autospec=True,
+    return_value=[("svc_a", "inst1"), ("svc_b", "inst1")],
+)
+@mock.patch(
+    "paasta_tools.setup_prometheus_adapter_config.PaastaServiceConfigLoader",
+    autospec=True,
+)
```
nemacysts: for new tests we normally try to stick to using `with mock.patch(...) as ...` pattern since it's a little easier to work with as a human since you don't need to remember how the decorator ordering corresponds to the function arg ordering :)

tests/test_setup_prometheus_adapter_config.py
```diff
@@ -390,6 +410,199 @@ def test_get_rules_for_service_instance(
...
+                {
+                    "type": provider_type,
+                    "setpoint": 0.5,
+                    "moving_average_window_seconds": window,
+                }
+                if x == provider_type
+                else None
+            )
+        ),
+    )
```
nemacysts: i was gonna comment on the mocks in this PR not setting `spec=...`, but it looks like i missed that in prior PRs that added tests and we're missing a spec on a ton of mocks here 

it would be helpful if you could have `claude` (or your agent of choice) try to add a spec to the newly added mocks in this PR - but worst case we can always fix this up later

paasta_tools/kubernetes_tools.py
```diff
@@ -823,13 +834,41 @@ def get_autoscaling_scaling_policy(
...
     ) -> Optional[V2MetricSpec]:
         target = provider["setpoint"]
         prometheus_hpa_metric_name = self.namespace_custom_prometheus_metric_name(
             provider["type"]
         )
+        if use_shared_rules and provider["type"] in TEMPLATEABLE_PROVIDERS:
+            window = provider.get(
+                "moving_average_window_seconds",
+                DEFAULT_WORKER_LOAD_AUTOSCALING_MOVING_AVERAGE_WINDOW,
+            )
```
nemacysts, line 856: should this use DEFAULT_MOVING_AVERAGE_WINDOW_BY_PROVIDER?

(/ should we move that from setup_prometheus_adapter_config.py to long_running_service_tools.py while we're doing that? otherwise it looks a little weird to import from setup_prometheus_adapter_config :p)

paasta_tools/kubernetes_tools.py
```diff
@@ -823,13 +834,41 @@ def get_autoscaling_scaling_policy(
     def namespace_custom_prometheus_metric_name(self, metric_name: str) -> str:
         return f"{self.get_sanitised_deployment_name()}-{metric_name}-prom"
 
+    @staticmethod
```
nemacysts: imo, this could just be a free function (i.e., defined outside of a class) rather than a staticmethod since i don't think we'd ever really call this as a staticmethod with something like `KubernetesDeploymentConfig.shared_prometheus_metric_name(...)`

paasta_tools/kubernetes_tools.py
```diff
@@ -823,13 +834,41 @@ def get_autoscaling_scaling_policy(
...
+    def shared_prometheus_metric_name(
```
nemacysts: the plan is to eventually remove use_shared_rules once we've enabled this everywhere, right?

if so, maybe we can name this with the future in mind - e.g., something like `get_prometheus_metric_name()`?

i'm also thinking that when we're doing the cleanup in the future, we could probably also have this function call `namespace_custom_prometheus_metric_name()`, but i think that might be a funkier refactor since we only wanna do that for non-arbitrary-promql case :p

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -901,10 +905,350 @@ def create_instance_arbitrary_promql_scaling_rule(
...
+    METRICS_PROVIDER_PISCINA: DEFAULT_PISCINA_AUTOSCALING_MOVING_AVERAGE_WINDOW,
+    METRICS_PROVIDER_GUNICORN: DEFAULT_GUNICORN_AUTOSCALING_MOVING_AVERAGE_WINDOW,
+}
+
+
+def create_shared_worker_load_scaling_rule(
+    paasta_cluster: str,
+    moving_average_window: int,
+) -> PrometheusAdapterRule:
+    metric_name = f"{METRICS_PROVIDER_WORKER_LOAD}-prom-{moving_average_window}"
```
nemacysts: hmm, maybe this should actually use the getter from kubernetes_tools so that there's a single place that defines the metric name format?

paasta_tools/setup_prometheus_adapter_config.py
```diff
@@ -901,10 +905,350 @@ def create_instance_arbitrary_promql_scaling_rule(
...
+DEFAULT_MOVING_AVERAGE_WINDOW_BY_PROVIDER = {
+    METRICS_PROVIDER_WORKER_LOAD: DEFAULT_WORKER_LOAD_AUTOSCALING_MOVING_AVERAGE_WINDOW,
+    METRICS_PROVIDER_UWSGI_V2: DEFAULT_UWSGI_AUTOSCALING_MOVING_AVERAGE_WINDOW,
+    METRICS_PROVIDER_UWSGI: DEFAULT_UWSGI_AUTOSCALING_MOVING_AVERAGE_WINDOW,
+    METRICS_PROVIDER_PISCINA: DEFAULT_PISCINA_AUTOSCALING_MOVING_AVERAGE_WINDOW,
+    METRICS_PROVIDER_GUNICORN: DEFAULT_GUNICORN_AUTOSCALING_MOVING_AVERAGE_WINDOW,
+}
+
+
+def create_shared_worker_load_scaling_rule(
```
nemacysts, line 912: i don't really love duplicating the functions, but the only  other alternatives i can think of aren't much better (e.g., push some branches into the existing `create_*()` functions or add some additional indirection through getters and whatnot)

...so i guess this stays :p

tests/kubernetes/application/test_controller_wrapper.py
```diff
@@ -28,6 +28,16 @@ def mock_load_system_paasta_config():
...
 
 
+@pytest.fixture(autouse=True)
+def mock_kubernetes_tools_load_system_paasta_config():
+    with mock.patch(
+        "paasta_tools.kubernetes_tools.load_system_paasta_config",
+        autospec=True,
+    ) as m:
+        m.return_value.get_use_prometheus_adapter_shared_rules.return_value = False
+        yield m
```
nemacysts, line 44: note: we normally try to not use autouse fixtures unless it's something that's 100% necessary for all tests

that said, i do sorta think that we do generally want to be better about autousing SystemPaastaConfig fixtures since it's a huge PITA to have tests work locally 'cause we actually have SPC available on a devbox and not on Github Actions :p

...so maybe we we want to do here is to mock the config that load_system_paasta_config() reads and then have all the values that we want to use for tests set in the mock (e.g., something kinda like what we to in test_tron_tools.py with MOCK_SYSTEM_PAASTA_CONFIG) rather than patching individual functions?

nemacysts (commented):
generally lgtm, mostly just some suggestions/me talking out loud

tests/test_setup_prometheus_adapter_config.py
sidtuladhar: oh yeah that clean

paasta_tools/kubernetes_tools.py
sidtuladhar: yeah removing shared rules after we rollout everywhere

paasta_tools/setup_prometheus_adapter_config.py
sidtuladhar: oh yeah!

paasta_tools/kubernetes_tools.py
sidtuladhar: good catch lol

tests/test_setup_prometheus_adapter_config.py
```diff
@@ -410,127 +412,82 @@ def test_get_rules_for_service_instance(
...
-    ]
-    config = create_prometheus_adapter_config("test_cluster", mock.Mock())
-    rules = config["rules"]
+    with mock.patch(
+        "paasta_tools.setup_prometheus_adapter_config.load_system_paasta_config",
+        autospec=True,
+        return_value=mock.Mock(
+            spec=SystemPaastaConfig,
+            get_use_prometheus_adapter_shared_rules=lambda: True,
+        ),
```
nemacysts, line 477: do we still need this mock if we use autouse mock like tests/test_kubernetes_tools.py in this file, and then replace with a non-autouse one just for the tests that don't use the default?

nemacysts (approved):
i left some comments re: the tests, but the test are somewhat temporary since this toggle won't be around long - so not really a blocker imo

tests/test_setup_prometheus_adapter_config.py
sidtuladhar: yeah ill delete this stuff in a different PR once we remove the toggle

PR #4343, opened 2026-07-01
DREIMP-11556 Add cassandracluster/cassandraclustereks counterpart mapping

siadat:
## Problem

After the non-EKS `cassandracluster-*.yaml` files were removed from yelpsoa-configs (TOOL-589), the autotune rightsizer silently skipped all Cassandra recommendations because
  `INSTANCE_TYPE_COUNTERPARTS` had no mapping for `cassandracluster` ↔ `cassandraclustereks`.

`cassandra-paasta-autotune` Jenkins job did not produce changes and reported "No files changed"

## Summary

Add the missing counterpart pair so `merge_recommendations` can find instances in the EKS config files.

### DIFF
--- a/paasta_tools/config_utils.py
+++ b/paasta_tools/config_utils.py
@@ -30,7 +30,12 @@ KNOWN_CONFIG_TYPES = (
 # this could use a better name - but basically, this is for pairs of instance types
 # where you generally want to check both types (i.e.,g a status-quo and migration
 # instance type)
-INSTANCE_TYPE_COUNTERPARTS = {"eks": "kubernetes", "kubernetes": "eks"}
+INSTANCE_TYPE_COUNTERPARTS = {
+    "eks": "kubernetes",
+    "kubernetes": "eks",
+    "cassandracluster": "cassandraclustereks",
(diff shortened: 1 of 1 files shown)


paasta_tools/config_utils.py
```diff
@@ -30,7 +30,12 @@
 # this could use a better name - but basically, this is for pairs of instance types
 # where you generally want to check both types (i.e.,g a status-quo and migration
 # instance type)
-INSTANCE_TYPE_COUNTERPARTS = {"eks": "kubernetes", "kubernetes": "eks"}
+INSTANCE_TYPE_COUNTERPARTS = {
+    "eks": "kubernetes",
+    "kubernetes": "eks",
+    "cassandracluster": "cassandraclustereks",
```
ilkinmammadzada, line 36: Do we really need this one? shouldn't `"cassandraclustereks": "cassandracluster",` enough?

nemacysts: i think technically we do just need one - but might as well be consistent and have both for now (and then we can clean up all of this once we've fully shut off the non-eks clusters)

nemacysts (approved):

PR #4345, opened 2026-07-09
Increase default disk size in paasta from 1024 MiB to 4096 MiB

CharlesPeacock:
follow up to https://github.com/Yelp/Tron/pull/1103 which bumped the Tron default disk size, maintaining consistency between the two repos. Previously the 1024MiB default caused issues for certain jobs, this change aims to mitigate that.

I also updated test assertions and hashes of YAML configs to reflect the new defaults.

### DIFF
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -741,8 +741,7 @@ Each Tron **action** of a job MAY specify the following:
 
   * Anything from :ref:`General Placement Options <general-placement-options>`
 
-  * ``disk``: Same as Common Settings, but defaults to 4096 MB for Tron Actions.
-     Spark Actions still use the Spark driver disk default when disk is omitted.
+  * ``disk``: Same as `Common Settings`, but defaults to 4096 MB for non-Spark Actions (Spark Actions use a default of 5120 MB instead).
 
   * ``service``: Uses a docker image from different service. When ``service`` is set
     for an action, that setting takes precedence over what is set for the job.
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -741,8 +741,7 @@ Each Tron **action** of a job MAY specify the following:
 
   * Anything from :ref:`General Placement Options <general-placement-options>`
 
-  * ``disk``: Same as `Common Settings`, but defaults to 4096 MB for Tron Actions.
-     Spark Actions still use the Spark driver disk default when disk is omitted.
+  * ``disk``: Same as `Common Settings`, but defaults to 4096 MB for non-Spark Actions (Spark Actions use a default of 5120 MB instead).
 
   * ``service``: Uses a docker image from different service. When ``service`` is set
     for an action, that setting takes precedence over what is set for the job.
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -741,6 +741,9 @@ Each Tron **action** of a job MAY specify the following:
 
   * Anything from :ref:`General Placement Options <general-placement-options>`
 
+  * ``disk``: Same as Common Settings, but defaults to 4096 MB for Tron Actions.
+     Spark Actions still use the Spark driver disk default when disk is omitted.
+
   * ``service``: Uses a docker image from different service. When ``service`` is set
     for an action, that setting takes precedence over what is set for the job.
 
--- a/docs/source/yelpsoa_configs.rst
+++ b/docs/source/yelpsoa_configs.rst
@@ -741,7 +741,7 @@ Each Tron **action** of a job MAY specify the following:
 
   * Anything from :ref:`General Placement Options <general-placement-options>`
 
-  * ``disk``: Same as Common Settings, but defaults to 4096 MB for Tron Actions.
+  * ``disk``: Same as `Common Settings`, but defaults to 4096 MB for Tron Actions.
      Spark Actions still use the Spark driver disk default when disk is omitted.
 
   * ``service``: Uses a docker image from different service. When ``service`` is set
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -77,7 +77,9 @@
                 "disk": {
                     "type": "number",
                     "minimum": 0,
-                    "exclusiveMinimum": true
+                    "default": 4096,
+                    "exclusiveMinimum": true,
+                    "$comment": "Setting default to 4096MiB to non-spark Tron actions"
                 },
                 "constraints": {
                     "type": "array",
--- a/paasta_tools/cli/schemas/tron_schema.json
+++ b/paasta_tools/cli/schemas/tron_schema.json
@@ -78,8 +78,7 @@
                     "type": "number",
                     "minimum": 0,
                     "default": 4096,
-                    "exclusiveMinimum": true,
-                    "$comment": "Setting default to 4096MiB to non-spark Tron actions"
+                    "exclusiveMinimum": true
                 },
                 "constraints": {
                     "type": "array",
--- a/paasta_tools/tron_tools.py
+++ b/paasta_tools/tron_tools.py
@@ -312,7 +312,7 @@ class TronActionConfig(InstanceConfig):
         return super().get_mem()
 
     def get_disk(self, default: float = 4096) -> float:
-        # Make default disk size for Spark pod 4gb because 1gb was too low 
+        # Make default disk size for Spark pod 4gb because 1gb was too low
         if self.action_spark_config and "disk" not in self.config_dict:
             return spark_tools.SPARK_DRIVER_DEFAULT_DISK_MB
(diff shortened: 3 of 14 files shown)


paasta_tools/tron_tools.py
```diff
@@ -311,8 +311,8 @@ def get_mem(self) -> float:
         # we fall back to this default if there's no Spark config
         return super().get_mem()
 
-    def get_disk(self, default: float = 1024) -> float:
-        # increase default threshold for Spark driver pod memory because 1G is too low
+    def get_disk(self, default: float = 4096) -> float:
+        # Make default disk size for Spark pod 4gb because 1gb was too low
```
KaspariK: Changing the default to 4096 makes sense for Tron, but I don't think this comment update is quite right. This branch is only for Spark actions, and Spark still uses `SPARK_DRIVER_DEFAULT_DISK_MB` when `disk` is omitted, which is currently 5120. The 4096 default applies to non-Spark Tron actions.

paasta_tools/utils.py
```diff
@@ -634,12 +634,12 @@ def use_docker_disk_quota(
     def get_docker_init(self) -> Iterable[DockerParameter]:
         return [{"key": "init", "value": "true"}]
 
-    def get_disk(self, default: float = 1024) -> float:
+    def get_disk(self, default: float = 4096) -> float:
```
KaspariK: We should avoid changing this shared default. `InstanceConfig.get_disk()` is inherited by non-Tron configs too (including Kubernetes service pods), so this changes the default disk for regular PaaSTA pods as well. Our intended scope is Tron, so I think this should stay at 1024 and the new default should live in `TronActionConfig`.

paasta_tools/tron_tools.py
KaspariK: We also need to pass the Tron default through here. Otherwise `default=4096` on this is ignored for non-Spark Tron actions, because calling `super().get_disk()` uses the base `InstanceConfig` default instead.

paasta_tools/tron_tools.py
```diff
@@ -311,8 +311,8 @@ def get_mem(self) -> float:
         # we fall back to this default if there's no Spark config
         return super().get_mem()
 
-    def get_disk(self, default: float = 1024) -> float:
-        # increase default threshold for Spark driver pod memory because 1G is too low
+    def get_disk(self, default: float = 4096) -> float:
```
KaspariK, line 314: Since this changes the public default for Tron action `disk`, we should update the Tron schema/docs to match. I think `docs/source/yelpsoa_configs.rst` and `paasta_tools/cli/schemas/tron_schema.json`.

For `docs/source/yelpsoa_configs.rst` we don't have a Tron-specific line for `disk` today. I think adding a note/bullet in the Tron Action section saying Tron Actions override the common disk default. Maybe like:
> `disk`: Same as `Common Settings`, but defaults to 4096 MB for Tron Actions. Spark Actions default to the Spark driver disk default when `disk` is omitted.

For `paasta_tools/cli/schemas/tron_schema.json` we don't seem to set a default, but I think adding one is reasonable.

tests/test_tron_tools.py
```diff
@@ -2095,7 +2095,7 @@ def test_create_complete_config_e2e(self, tmpdir):
         # that are not static, this will cause continuous reconfiguration, which
         # will add significant load to the Tron API, which happened in DAR-1461.
         # but if this is intended, just change the hash.
-        assert hasher.hexdigest() == "5cf9985b6227ce18dc87376b687046f7"
+        assert hasher.hexdigest() == "eb3b5612be2c275ae768a7ff6225d005"
```
KaspariK, line 2098: Just fyi, this hash update is expected. 

`format_tron_action_dict()` emits `disk` for `executor: paasta`/k8s Tron actions, so actions that omit `disk` will now generate `disk: 4096` instead of `disk: 1024`. Actions with an explicit `disk` keep their configured value, Spark actions without `disk` still use `SPARK_DRIVER_DEFAULT_DISK_MB` (and `executor: ssh` actions don't emit these, I think?).
  
This hash changing is not a concern. The Kubernetes service pod hash changes are the ones that would indicate the scope leaked beyond Tron.

paasta_tools/tron_tools.py
CharlesPeacock: Thanks for the  clarification, I have corrected this. Sorry for the misunderstanding

tests/test_tron_tools.py
CharlesPeacock: Thank you, that is great to know. I have added the correct hashes after corrections.

paasta_tools/tron_tools.py
```diff
@@ -311,12 +311,12 @@ def get_mem(self) -> float:
...
+        # Make default disk size 4gb for non-spark Tron actions
         if self.action_spark_config and "disk" not in self.config_dict:
             return spark_tools.SPARK_DRIVER_DEFAULT_DISK_MB
         # we fall back to this default if there's no Spark config
-        return super().get_disk()
+        return super().get_disk(default=4096)
```
KaspariK: This will ignore the default parameter. This should be:
```python
return super().get_disk(default=default)
```
Otherwise `get_disk(default=...)` callers will still always get 4096 as the fallback here.

paasta_tools/tron_tools.py
```diff
@@ -311,12 +311,12 @@ def get_mem(self) -> float:
         # we fall back to this default if there's no Spark config
         return super().get_mem()
 
-    def get_disk(self, default: float = 1024) -> float:
-        # increase default threshold for Spark driver pod memory because 1G is too low
+    def get_disk(self, default: float = 4096) -> float:
+        # Make default disk size 4gb for non-spark Tron actions
```
KaspariK: Nit: I don’t think the original comment was especially good/helpful, but this changes its intent. My read is that the comment was meant to explain the Spark-specific branch:
```python
if self.action_spark_config and "disk" not in self.config_dict:
    return spark_tools.SPARK_DRIVER_DEFAULT_DISK_MB
```
The new wording reads more like a docstring for get_disk() overall. I’d either remove it altogether or make it branch-specific, e.g. `# Spark driver pods use a separate default when disk is omitted.`

docs/source/yelpsoa_configs.rst
```diff
@@ -741,6 +741,9 @@ Each Tron **action** of a job MAY specify the following:
 
   * Anything from :ref:`General Placement Options <general-placement-options>`
 
+  * ``disk``: Same as Common Settings, but defaults to 4096 MB for Tron Actions.
```
KaspariK: Nit: just for consistency, can we link `Common Settings`. Like how it's done on line 740.

docs/source/yelpsoa_configs.rst
```diff
@@ -741,6 +741,9 @@ Each Tron **action** of a job MAY specify the following:
 
   * Anything from :ref:`General Placement Options <general-placement-options>`
 
+  * ``disk``: Same as `Common Settings`, but defaults to 4096 MB for Tron Actions.
+     Spark Actions still use the Spark driver disk default when disk is omitted.
```
nemacysts:
```suggestion
  * ``disk``: Same as `Common Settings`, but defaults to 4096 MB for non-Spark Actions (Spark Actions use a default of 5120 MB instead).
```
(just to have it all in one place)

paasta_tools/tron_tools.py
nemacysts: yea, the schema defaults aren't super important - i think at most those are just used by IDEs that understand the schema to show better intellisense-style context :p

paasta_tools/cli/schemas/tron_schema.json
```diff
@@ -77,7 +77,9 @@
                 "disk": {
                     "type": "number",
                     "minimum": 0,
-                    "exclusiveMinimum": true
+                    "default": 4096,
+                    "exclusiveMinimum": true,
+                    "$comment": "Setting default to 4096MiB to non-spark Tron actions"
```
nemacysts: imo, we can skip the comment here :)

nemacysts (commented):
feel free to ignore me :)

paasta_tools/cli/schemas/tron_schema.json
KaspariK: Funny thing I found while reviewing last week is that our version of json schema doesn't even support comments, but we have them elsewhere and unknown keys are ignored so it's mostly inconsequential.

edit: I do agree with your comment though, just trivia dumping

CharlesPeacock: Thanks for the help guys! I have removed the comment.

CharlesPeacock: This is weird. I deleted the comment [here](https://github.com/Yelp/paasta/commit/2847ad7795755463280e76c64e41a075d62f1621)

But the PR does not show the update. Maybe some eventual consistency stuff? I may have messed up.

Edit: Fixed, should be good now.

KaspariK (approved):

nemacysts (approved):

PR #4347, opened 2026-07-10
remove secrets_for_teams

piax93:
This was a feature powered by our OPA authorization layer, which is dead since the migration to EKS, so worth removing support.

**EDIT:** since apparently there are some people that still relied on the feature, using it wrong though, I introduced a CLI argument that can be used to trigger the same codepath that will query k8s APIs for secrets.

### DIFF
--- a/paasta_tools/cli/cmds/local_run.py
+++ b/paasta_tools/cli/cmds/local_run.py
@@ -68,7 +68,6 @@ from paasta_tools.utils import _run
 from paasta_tools.utils import get_docker_client
 from paasta_tools.utils import get_possible_launched_by_user_variable_from_env
 from paasta_tools.utils import get_username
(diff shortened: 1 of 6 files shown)


danielpopskandor (commented):
<wrong account>

danielpops (commented):
FYI there is at least one team that relies on this functionality to run their paasta services locally. I don't think it's OK to remove this functionality without providing an alternative

nemacysts (approved):
i assume we'll send some comms out to let the few folks that use this that they'll need to pass some new args?

piax93:
> i assume we'll send some comms out to let the few folks that use this that they'll need to pass some new args?

Yep, I have a grand total of 2 DMs to send. Manageable :P

PR #4348, opened 2026-07-10
PEREL-5643: support Gi/Mi strings in autotune_limits mem/disk clamping

matt-ullmer:
## Summary

- Add `_resource_value_to_mib()` to normalize resource values (int/float MiB or K8s strings like `"4Gi"`/`"512Mi"`) to a common unit before comparing
- Update `_clamp_recommendations()` to use it so Cassandra clusters can set mem/disk floors via `autotune_limits` without TypeError on string comparisons
- Fix pre-existing log bug: max-clamping branch was logging `min_value` instead of `max_value`

Cassandra autotune recommendations use Gi strings for mem/disk. Direct `<`/`>` comparisons against integer limits would raise TypeError. This is the paasta-side change; a follow-up yelpsoa-configs PR will update the `cassandra_k8s/validate.py` schema from `integer` to `string` pattern once this is deployed.

Jira: https://jira.yelpcorp.com/browse/PEREL-5643

## Test plan

- [ ] `pytest tests/test_config_utils.py` — all 37 tests pass including new parametrized `test_resource_value_to_mib` cases and `test_auto_config_updater_merge_recommendations_limits_gi_strings`
- [ ] pre-commit clean on changed files

### DIFF
--- a/paasta_tools/config_utils.py
+++ b/paasta_tools/config_utils.py
@@ -9,6 +9,7 @@ from typing import List
 from typing import Optional
 from typing import Set
 from typing import Tuple
+from typing import Union
 
 import ruamel.yaml as yaml
 
@@ -21,7 +22,7 @@ log = logging.getLogger(__name__)
 _RESOURCE_UNIT_TO_MIB = {"Gi": 1024.0, "Mi": 1.0, "Ki": 1.0 / 1024}
 
 
-def _resource_value_to_mib(value: Any) -> float:
+def _resource_value_to_mib(value: Union[int, float, str]) -> float:
     if isinstance(value, (int, float)):
         return float(value)
     if isinstance(value, str):
@@ -313,18 +314,15 @@ class AutoConfigUpdater:
 
             # otherwise, we can do some pretty rote clamping of resource values
             elif unclamped_resource_value is not None:
-                if min_value and _resource_value_to_mib(
-                    unclamped_resource_value
-                ) < _resource_value_to_mib(min_value):
+                unclamped_mib = _resource_value_to_mib(unclamped_resource_value)
+                if min_value and unclamped_mib < _resource_value_to_mib(min_value):
                     log.debug(
                         f"{limit_type} autotune config under configured limit ({min_value}), using autotune limit lower bound."
                     )
                     clamped_recomendation[limit_type] = min_value
-                if max_value and _resource_value_to_mib(
-                    unclamped_resource_value
-                ) > _resource_value_to_mib(max_value):
+                if max_value and unclamped_mib > _resource_value_to_mib(max_value):
                     log.debug(
-                        f"{limit_type} autotune config over configured limit ({min_value}), using autotune limit upper bound."
+                        f"{limit_type} autotune config over configured limit ({max_value}), using autotune limit upper bound."
                     )
                     clamped_recomendation[limit_type] = max_value
             else:
--- a/paasta_tools/config_utils.py
+++ b/paasta_tools/config_utils.py
@@ -19,14 +19,13 @@ from paasta_tools.utils import DEFAULT_SOA_DIR
 log = logging.getLogger(__name__)
 
 _RESOURCE_UNIT_TO_MIB = {"Gi": 1024.0, "Mi": 1.0, "Ki": 1.0 / 1024}
+_SUFFIX_RESOURCE_TYPES = {"mem", "disk"}
 
 
-def _resource_value_to_mib(value: int | float | str) -> float:
-    """Normalize a resource value to MiB for comparison. For numeric types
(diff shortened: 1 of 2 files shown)


paasta_tools/config_utils.py
```diff
@@ -18,6 +19,19 @@
 
 log = logging.getLogger(__name__)
 
+_RESOURCE_UNIT_TO_MIB = {"Gi": 1024.0, "Mi": 1.0, "Ki": 1.0 / 1024}
+
+
+def _resource_value_to_mib(value: Union[int, float, str]) -> float:
```
nemacysts: non-blocking: since we're running on a python that supports pep-604
```suggestion
def _resource_value_to_mib(value: int | float | str) -> float:
```
might be a little more readable :p

paasta_tools/config_utils.py
```diff
@@ -18,6 +19,19 @@
...
+    if isinstance(value, (int, float)):
+        return float(value)
+    if isinstance(value, str):
+        for suffix, factor in _RESOURCE_UNIT_TO_MIB.items():
+            if value.endswith(suffix):
+                return float(value[: -len(suffix)]) * factor
```
nemacysts: non-blocking suggestion if you wanna get rid of the indexing
```suggestion
                return float(value.removesuffix(suffix)) * factor
```

paasta_tools/config_utils.py
```diff
@@ -300,14 +314,15 @@ def _clamp_recommendations(
...
+                if min_value and unclamped_mib < _resource_value_to_mib(min_value):
                     log.debug(
                         f"{limit_type} autotune config under configured limit ({min_value}), using autotune limit lower bound."
                     )
                     clamped_recomendation[limit_type] = min_value
-                if max_value and unclamped_resource_value > max_value:
+                if max_value and unclamped_mib > _resource_value_to_mib(max_value):
                     log.debug(
-                        f"{limit_type} autotune config over configured limit ({min_value}), using autotune limit upper bound."
+                        f"{limit_type} autotune config over configured limit ({max_value}), using autotune limit upper bound."
```
nemacysts, line 325: whoops, thank you :p

paasta_tools/config_utils.py
```diff
@@ -300,14 +314,15 @@ def _clamp_recommendations(
 
             # otherwise, we can do some pretty rote clamping of resource values
             elif unclamped_resource_value is not None:
-                if min_value and unclamped_resource_value < min_value:
+                unclamped_mib = _resource_value_to_mib(unclamped_resource_value)
+                if min_value and unclamped_mib < _resource_value_to_mib(min_value):
```
nemacysts: hmmm, the naming here is gonna be a bit confusing - this is used for all 3 resource types (cpu, mem, disk), but only applies to mem/disk

...not sure if we wanna branch of the type or if we want to just have a comment here explaining the weirdness

nemacysts: @matt-ullmer i think this is the only outstanding comment now - i'm not really sure what's best though :p 

i think another potential option might be something like
```python
for limit_type, limits in config.get("autotune_limits", {}).items():
            log.debug(f"Processing {limit_type} autotune limits...")
            min_value = _normalized_value(limits.get("min"), limit_type)
            max_value = _normalized_value(limits.get("max"), limit_type)
            unclamped_resource_value = _normalized_value(clamped_recomendation.get(limit_type), limit_type)
```

where `_normalized_value` is something like:
```python
NORMALIZABLE_VALUE_TYPES = {"mem", "disk", ...}

...

def _normalized_value(value: int | float | str, value_type: str) -> float: # XXX: maybe an enum/Literal/etc for value_type would work?
    """Normalize a resource value to MiB for comparison. For numeric types
    (e.g. cpus as float), this is a passthrough cast to float."""

    if isinstance(value, (int, float)):
        return float(value)
    elif value_type in NORMALIZABLE_VALUE_TYPES and isinstance(value, str):
        for suffix, factor in _RESOURCE_UNIT_TO_MIB.items():
            if value.endswith(suffix):
                return float(value.removesuffix(suffix)) * factor

    raise ValueError(f"Cannot parse {value_type} resource value: {value!r}")
```

(naming yolo'd - i am not tied to any of the names here and didn't think too hard about them)

matt-ullmer: Good idea — went with something close to your sketch. Pushed `2a1b013`: renamed to `_normalize_resource_value(value, resource_type)` with a `_SUFFIX_RESOURCE_TYPES = {"mem", "disk"}` guard. Numeric values pass through for any type, suffix parsing only fires for mem/disk, and strings for cpu now raise ValueError. Added a test for the cpu-string case too.

nemacysts: do we still want to remove the `unclamped_mib` naming since not all of these will have an mib unit value?

PR #4350, opened 2026-07-14
fix: non worker load autoscaling label selector

sidtuladhar:
Fix shared adapter rules for uwsgi/piscina/gunicorn: `<<.LabelMatchers>>` includes `kube_namespace` which doesn't exist on `k8s:deployment:pods_status_ready` or `kube_deployment_spec_replicas`, causing empty query results. Use `<<index .LabelValuesByName ...>>` to select only labels each metric actually has.


nemacysts:
@sidtuladhar could you add a quick description for this PR? :D

(i.e., i know that this is essentially the same issue from https://github.com/Yelp/paasta/commit/505b3b6bebaa4e2019d080d7bf5c86940c18cccd - but for future us it'd be nice to have a description :p)

nemacysts (approved):

PR #4352, opened 2026-07-20
PAASTA-18940: Create getter for exclusion list for monitoring.yaml files in paasta validate

jall0h:
Add exclusion list getter in SystemPaastaConfig for check monitoring.yaml files for paasta validate command
[PAASTA-18940](https://jira.yelpcorp.com/browse/PAASTA-18940)

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -2021,6 +2021,7 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     spark_blockmanager_port: int
     skip_cpu_burst_validation: List[str]
     skip_unique_instance_name_validation: List[str]
+    skip_check_monitoring_file_exists: List[str]
     tron_default_pool_override: str
     spark_kubeconfig: str
     spark_iam_user_kubeconfig: str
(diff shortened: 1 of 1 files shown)


paasta_tools/utils.py
```diff
@@ -2705,6 +2706,9 @@ def get_skip_cpu_override_validation_services(self) -> List[str]:
     def get_skip_cpu_burst_validation_services(self) -> List[str]:
         return self.config_dict.get("skip_cpu_burst_validation", [])
 
+    def get_skip_check_monitoring_file_exists(self) -> List[str]:
+        return self.config_dict.get("skip_check_monitoring_file_exists", [])
```
nemacysts: i'd maybe name this `skip_monitoring_file_existence_validation` e.g.,

```suggestion
    def get_skip_monitoring_file_existence_validation_services(self) -> List[str]:
        return self.config_dict.get("skip_monitoring_file_existence_validation", [])
```
so that it reads a little more clearly

nemacysts (commented):
if you want, we can also start using this in this PR as well - should be safe since we have a default value in the getter :)

nemacysts (dismissed):
whoops, not sure how i missed re-reviewing this!

nemacysts (approved):

PR #4355, opened 2026-07-23
creating validation checks for single replica

mbrankin-art:
Creating validation checks to check for single replica instances

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -956,26 +956,24 @@ def warn_single_replica_instances(service_path: str) -> bool:
 
             if replicas == 1:
                 config_file_path = instance_config.get_config_path()
-                config = get_config_file_dict(config_file_path, use_ruamel=True)
                 config_flat = _get_config_flattened(config_file_path)
 
-                if config[instance].get(replica_key) is None:
-                    # relying on implicit default of 1 — no explicit acknowledgment possible
+                if config_flat[instance].get(replica_key) is None:
                     returncode = False
                     print(
                         failure(
                             msg=f"Instance {instance} on cluster {cluster} has only 1 replica (implicit default). "
-                            f"Set instances explicitly or increase to more than 1.",
+                            f"Set `instances` explicitly if not autoscaled or set `min_instances` greater than 1 if autoscaled.",
                             link="y/override-single-replica",
                         )
                     )
                     continue
 
                 comment = _get_comments_for_key(
-                    data=config[instance],
+                    data=config_flat[instance],
                     key=replica_key,
-                    full_config=config,
-                    key_value=config[instance].get(replica_key),
+                    full_config=config_flat,
+                    key_value=config_flat[instance].get(replica_key),
                     full_config_flattened=config_flat,
                 )
                 if (
@@ -990,6 +988,7 @@ def warn_single_replica_instances(service_path: str) -> bool:
                     print(
                         failure(
                             msg=f"Instance {instance} on cluster {cluster} has only 1 replica. "
+                            f"Single replicas are a SPOF and we do not encourage it. "
                             f"Add a comment to acknowledge: # override-single-replica (PROJ-1234)",
                             link="y/override-single-replica",
                         )
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -141,6 +141,10 @@ OVERRIDE_CPU_AUTOTUNE_ACK_PATTERN = r"#\s*override-cpu-setting\s+\(.+[A-Z]+-[0-9
 # but we don't have a $ anchor in case users want to add an additional
 # comment
 OVERRIDE_CPU_BURST_ACK_PATTERN = r"#\s*override-cpu-burst\s+\(.+[A-Z]+-[0-9]+.+\)"
+# we expect a comment that looks like # override-single-replica (PROJ-1234)
+OVERRIDE_SINGLE_REPLICA_ACK_PATTERN = (
+    r"#\s*override-single-replica\s+\(.+[A-Z]+-[0-9]+.+\)"
+)
 # for now, double the autotune cap to give people the benefit of the doubt
 # if we see that people are still misusing this configuration, we can lower
 # this to the autotune cap (i.e., 1)
@@ -853,11 +857,7 @@ def validate_autoscaling_configs(service_path: str) -> bool:
 
                             # we need access to the comments, so we need to read the config with ruamel to be able
                             # to actually get them in a "nice" automated fashion
-                            config_file_path = os.path.join(
-                                soa_dir,
-                                service,
-                                f"{instance_config.get_instance_type()}-{cluster}.yaml",
-                            )
+                            config_file_path = instance_config.get_config_path()
                             config = get_config_file_dict(
                                 config_file_path,
                                 use_ruamel=True,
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +937,35 @@ def validate_min_max_instances(service_path):
...
+            else:
+                replicas = instance_config.get_instances()
+
+            if replicas is not None and replicas <= 1:
+                returncode = False
+                print(
+                    failure(
+                        f"Instance {instance} on cluster {cluster} has only {replicas} replica(s). "
+                        f"Set the instances field to a value greater than 1.",
+                        "",
```
nemacysts: (we'll probably want a y/ link to a wiki page that explains why this is important in more detail to pass as the link arg to `failure()`)

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +937,35 @@ def validate_min_max_instances(service_path):
...
+            service=service, cluster=cluster, soa_dir=soa_dir
+        ):
+            if not isinstance(instance_config, LongRunningServiceConfig):
+                continue
+            if instance_config.is_autoscaling_enabled():
+                replicas = instance_config.get_min_instances()
+            else:
+                replicas = instance_config.get_instances()
+
+            if replicas is not None and replicas <= 1:
```
nemacysts: i *think* we just want `== 1`: it's "allowed" to have 0 replicas if you want to turn off a service instance without deleting all the config

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +937,35 @@ def validate_min_max_instances(service_path):
...
+    for cluster in list_clusters(service, soa_dir):
+        for instance, instance_config in load_all_instance_configs_for_service(
+            service=service, cluster=cluster, soa_dir=soa_dir
+        ):
+            if not isinstance(instance_config, LongRunningServiceConfig):
+                continue
+            if instance_config.is_autoscaling_enabled():
+                replicas = instance_config.get_min_instances()
+            else:
+                replicas = instance_config.get_instances()
```
nemacysts, line 954: would we want different messages for these? right now, the failure message for `min_instances: 1` would tell folks to set `instances: NOT_ONE`, which isn't exactly what we want them to do

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +937,35 @@ def validate_min_max_instances(service_path):
     return returncode
 
 
+def validate_single_replica(service_path: str) -> bool:
```
nemacysts: !8ball `warn_single_replica_instances` or `validate_no_single_replica_instances` would be a clearer name

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +937,35 @@ def validate_min_max_instances(service_path):
...
+    soa_dir, service = path_to_soa_dir_service(service_path)
+    returncode = True
+
+    # We need to validate to see if the config has more than 1 single replica instance.
+    # If it does, we need to print a warning message.
```
nemacysts: i think we'll also want a way to opt-out of this - at least initially, there's probably going to be a decent number of instances failing this check that we'll want to grandfather in

paasta_tools/cli/cmds/validate.py
mbrankin-art: makes sense to me! will change this now

paasta_tools/cli/cmds/validate.py
mbrankin-art: gotcha will do this

paasta_tools/cli/cmds/validate.py
mbrankin-art: understood will reflect this

paasta_tools/cli/cmds/validate.py
mbrankin-art: i went for `warn_single_replica_instances` in the end

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +941,71 @@ def validate_min_max_instances(service_path):
...
+            if not isinstance(instance_config, LongRunningServiceConfig):
+                continue
+            if instance_config.is_autoscaling_enabled():
+                replicas = instance_config.get_min_instances()
+                replica_key = "min_instances"
+            else:
+                replicas = instance_config.get_instances()
+                replica_key = "instances"
+
+            if replicas is not None and replicas == 1:
```
nemacysts: should this be just `if replicas == 1`? seems like the null check is redundant, right?

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +941,71 @@ def validate_min_max_instances(service_path):
...
+                config_file_path = os.path.join(
+                    soa_dir,
+                    service,
+                    f"{instance_config.get_instance_type()}-{cluster}.yaml",
+                )
+                config = get_config_file_dict(config_file_path, use_ruamel=True)
```
nemacysts: thoughts on just loading the config flattened once here rather than loading unflattened here and then potentially re-loading flattened later?

paasta_tools/cli/cmds/validate.py
nemacysts: i'm also half-debating whether or not we should extract this to an internal helper (e.g., `def _has_min_replica_optout(...)`) since i was initially wondering why we were re-loading the config dict rather than using the getters :p 

(that said, i wouldn't be surprised if i was silly and didn't do this (or add a comment) for the other instances of a comment-based-optout that i've added :p)

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +941,71 @@ def validate_min_max_instances(service_path):
...
+            else:
+                replicas = instance_config.get_instances()
+                replica_key = "instances"
+
+            if replicas is not None and replicas == 1:
+                config_file_path = os.path.join(
+                    soa_dir,
+                    service,
+                    f"{instance_config.get_instance_type()}-{cluster}.yaml",
+                )
```
nemacysts: hmm, i kinda wonder if it's worth adding a quick `def get_config_path()` to `InstanceConfig` - the class knows what the soa_dir, service, cluster, and instance type are :p 

(that said, this is optional :p)

paasta_tools/cli/cmds/validate.py
nemacysts: actually, i guess we can't refactor to something like that since we're also doing the validation at the same time (and the logic is somewhat conditional)

...so maybe just a comment? :p

paasta_tools/utils.py
```diff
@@ -4144,7 +4153,7 @@ def timeout(
     if use_signals:
 
         def decorate(
-            func: Callable[..., _TimeoutFuncRetType]
+            func: Callable[..., _TimeoutFuncRetType],
```
mbrankin-art: black formatting?

paasta_tools/utils.py
```diff
@@ -459,6 +459,15 @@ def __get_interpolation_facts(self) -> Dict[str, str]:
             "service": self.service,
         }
 
+    def get_config_path(self) -> str:
+        instance_type = self.get_instance_type()
+        assert instance_type is not None
```
nemacysts, line 464: i'm surprised we default this to None :p

nemacysts: (for future us to maybe refactor, i guess :p)

paasta_tools/cli/cmds/validate.py
nemacysts: (i.e., drop `get_config_file_dict()` and just use `_get_config_flattened()`

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +937,67 @@ def validate_min_max_instances(service_path):
...
+                config = get_config_file_dict(config_file_path, use_ruamel=True)
+                config_flat = _get_config_flattened(config_file_path)
+
+                if config[instance].get(replica_key) is None:
+                    # relying on implicit default of 1 — no explicit acknowledgment possible
+                    returncode = False
+                    print(
+                        failure(
+                            msg=f"Instance {instance} on cluster {cluster} has only 1 replica (implicit default). "
+                            f"Set instances explicitly or increase to more than 1.",
```
```suggestion
                            f"Set `instances` explicitly if not autoscaled set `min_instances` greater than 1 if autoscaled.",
```

maybe? i was re-reading this a bit and i'm trying to see this from the perspective of a service owner

(that said, i think we can also just see if people have feedback once this is live)

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +937,67 @@ def validate_min_max_instances(service_path):
...
+                        pattern=OVERRIDE_SINGLE_REPLICA_ACK_PATTERN,
+                        string=comment,
+                    )
+                    is None
+                ):
+                    returncode = False
+                    print(
+                        failure(
+                            msg=f"Instance {instance} on cluster {cluster} has only 1 replica. "
+                            f"Add a comment to acknowledge: # override-single-replica (PROJ-1234)",
```
nemacysts, line 992: we might want also want to be a little more explicit about why this check exists here and have some sort of a mention about this being unsafe (otherwise, it reads as we're just randomly adding friction :p)

nemacysts (commented):
i have some wording comments (one that i think we could punt on if we'd like and another that we should maybe do now) + a quick suggestion re: the config file dict loading

but otherwise i think this pretty much good to go

paasta_tools/utils.py
mbrankin-art: what should we default it to?

paasta_tools/cli/cmds/validate.py
mbrankin-art: updated in the new version. lmk what you think!

paasta_tools/cli/cmds/validate.py
mbrankin-art: updated in the new version. lmk what you think!

paasta_tools/utils.py
nemacysts: i think we'd probably wanna look into this in the future and see if we even need a default or if we should just crash if something has subclassed InstanceConfig and not configured things correctly

(i.e., this is fine for now - it's just kinda weird that we did things this way when we first wrote get_instance_type())

nemacysts (approved):

ilkinmammadzada (approved):

PR #4357, opened 2026-07-28
Add a `verify-image-exists` sub-command

nemacysts:
This will be helpful for some model training adhoc jobs that get images built in a somewhat roundabout way.

Without this, someone has to be watching the build and waiting for it to complete before triggering their adhoc job.
However, with this: they can prefix their adhoc launch command with this and not have to manually do anything :)


cuza (dismissed):
LGTM since we're just doing a cli wrapper around `is_docker_image_already_in_registry()`

nemacysts:
@cuza need another shipit - on my N+1 look at the tests i noticed i had left in some vibecoded slop tests :p

cuza (approved):

PR #4358, opened 2026-07-29
feat: New config to configure CPU limits per cluster per pool

sidtuladhar:
[PAASTA-18929](https://jira.yelpcorp.com/browse/PAASTA-18929)

We want to prevent large CPU usage services on the stable pool. This is a generic validation that allows configuring cpu limits per cluster per pool. The configuration will be managed through karpenter

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -956,7 +956,7 @@ def validate_pool_limits(service_path: str) -> bool:
             cpu = instance_config.get_cpus()
             pool = instance_config.get_pool()
 
-            pool_limits = pool_limits_for_cluster.get(pool, {})
+            pool_limits: Dict[str, Any] = pool_limits_for_cluster.get(pool, {})
             max_cpus_for_pool = pool_limits.get("max_cpus", float("inf"))
             recommended_pool = pool_limits.get("recommended_pool")
 
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -938,13 +938,15 @@ def validate_min_max_instances(service_path):
 
 
 def validate_pool_limits(service_path: str) -> bool:
-    # We need to validate pool.
-    # If it does, we need to print a warning message.
+    """
+    Validate that services in specific pools won't exceed per-node capacities.
+    For the most part, this isn't normally an issue - but there are several pools where
+    folks tend to want to run extra-large workloads that won't fit (e.g., large single-node batches).
+    """
     soa_dir, service = path_to_soa_dir_service(service_path)
     returncode = True
 
     for cluster in list_clusters(service, soa_dir):
-
         pool_limits_for_cluster = (
             load_system_paasta_config().get_pool_limits().get(cluster, {})
         )
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -937,6 +937,30 @@ def validate_min_max_instances(service_path):
     return returncode
 
 
+def validate_stable_pool_workloads(service_path: str) -> bool:
+    soa_dir, service = path_to_soa_dir_service(service_path)
+    returncode = True
+
+    for cluster in list_clusters(service, soa_dir):
+        for instance, instance_config in load_all_instance_configs_for_service(
+            service=service, cluster=cluster, soa_dir=soa_dir
+        ):
+            cpu = instance_config.get_cpus()
+            pool = instance_config.get_pool()
+
+            if pool == "stable" and cpu >= 16:
+                returncode = False
+                print(
+                    failure(
+                        f"""Instance {instance} on cluster {cluster} has {cpu} CPUs and is in the stable pool.
+                        The stable pool should not have workloads with more than 15 CPUs.
+                        If you need to run a workload with more than 15 CPUs, consider using the stable-giant pool.""",
+                        "",
+                    )
+                )
+    return returncode
+
+
 def check_secrets_for_instance(
     instance_config_dict: InstanceConfigDict, soa_dir: str, service: str, vault_env: str
 ) -> bool:
@@ -1411,6 +1435,7 @@ def paasta_validate_soa_configs(
         validate_autoscaling_configs,
         validate_secrets,
         validate_min_max_instances,
+        validate_stable_pool_workloads,
         validate_cpu_burst,
         validate_smartstack,
         validate_flink_monitoring_team,
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -82,6 +82,7 @@ from paasta_tools.tron_tools import load_tron_service_config
 from paasta_tools.tron_tools import validate_complete_config
 from paasta_tools.utils import InstanceConfig
 from paasta_tools.utils import InstanceConfigDict
+from paasta_tools.utils import PoolLimits
 from paasta_tools.utils import get_service_instance_list
 from paasta_tools.utils import list_all_instances_for_service
 from paasta_tools.utils import list_clusters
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +937,32 @@ def validate_min_max_instances(service_path):
     return returncode
 
 
+def validate_stable_pool_workloads(service_path: str) -> bool:
```
nemacysts: i think we might be able to make this generic? 

i'm thinking something like:
```Python
def validate_pool_workload_sizing(service_path: str) -> bool:
    soa_dir, service = path_to_soa_dir_service(service_path)
    ok = True

        pool_limits_for_cluster = load_system_paasta_config().get_pool_limits().get(cluster, {})
        for instance, instance_config in load_all_instance_configs_for_service(
            service=service, cluster=cluster, soa_dir=soa_dir
        ):
            cpu = instance_config.get_cpus()
            pool = instance_config.get_pool()
            
            pool_limits = pool_limits_for_cluster.get(pool, {})
            max_cpus_for_pool = pool_limits.get("max_cpus", float("inf"))
            suggested_pool = pool_limits.get("suggested_pool")

            if pool in pool_limits and cpu >= max_cpus_for_pool:
                ok = False
                print(
                    failure(
                        f"""{service}.{instance} in {cluster} has {cpu} CPUs, which exceeds the limit of {max_cpus_for_pool} for the {pool} pool.
                        If you need to run a workload with this many CPUs, consider using the {suggested_pool} pool instead.""",
                        "",
                    )
                )
    return ok
```

where in system_paasta_config we have something like:
```json
{
  "cool-cluster": {
    "cool-pool": {
      "max_cpus": 16,
      "suggested_pool": "cooler-pool"
    },
    "uncool-pool": {
      "max_cpus": 160,
      "suggested_pool": "uncooler-pool"
    }
  },
  "uncool-cluster": {
    "cool-pool": {
      "max_cpus": 61,
      "suggested_pool": "cooler-pool"
    },
    "uncool-pool": {
      "max_cpus": 16,
      "suggested_pool": "uncooler-pool"
    }
  }
}
```

that way, we don't need to make a paasta release if we change the instance limits for a pool (or do anything special if we have different limits for different clusters) - plus, we can then extend this more easily later to also limit things like memory usage as well (which we may also want to consider doing)

that said, it would also be nice if we could figure out a way to do this a little more automatically (e.g., something that can look at our karpenter config)

sidtuladhar: oh that's really flexible!

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +937,39 @@ def validate_min_max_instances(service_path):
     return returncode
 
 
+def validate_pool_limits(service_path: str) -> bool:
+    # We need to validate pool.
+    # If it does, we need to print a warning message.
```
nemacysts: i'd maybe make this a docstring that looks something like

```suggestion
    """
    Validate that services in specific pools won't exceed per-node capacities.
    
    For the most part, this isn't normally an issue - but there are several pools where folks tend to want to run extra-large workloads that won't fit (e.g., large single-node batches).
```

nemacysts: (i wrote that in the GH UI, so probably needs some line breaks :p)

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +937,39 @@ def validate_min_max_instances(service_path):
...
+    soa_dir, service = path_to_soa_dir_service(service_path)
+    returncode = True
+
+    for cluster in list_clusters(service, soa_dir):
+
```
```suggestion
```
(just to be consistent)

paasta_tools/utils.py
```diff
@@ -1987,6 +1987,7 @@ class SystemPaastaConfigDict(TypedDict, total=False):
     pki_backend: str
     pod_defaults: Dict[str, Any]
     pool_node_affinities: Dict[str, Dict[str, List[str]]]
+    pool_limits: Dict[str, Dict[str, Dict[str, Any]]]
```
nemacysts: non-blocking suggestion: might be nice to have a comment above this with what we expect this dict to look like (or as a docstring for get_pool_limits() - `Dict[str, Dict[str, Dict[str, Any]]]` is helpful, but it doesn't really tell folks what it will look like in reality without reading the `paasta validate` code that uses it)

nemacysts: /i think we could maybe make the inner dict a typed dict?

for now, we're expecting the innermost dict to look like
```Python
class PoolLimits(TypedDict):
  max_cpus: float
  recommended_pool: str
```

so:
```Python
...

class PoolLimits(TypedDict):
    max_cpus: float
    recommended_pool: str

...

class SystemPaastaConfigDict(TypedDict, total=False):
    ...
    # i.e., {cluster: {pool: LIMIT_CONFIG}}
    pool_limits: Dict[str, Dict[str, PoolLimits]]
    ...

...
```

nemacysts (dismissed):
i have some minor suggestions, but they're non-blocking if you'd prefer to do them in another PR (or disregard them)

paasta_tools/utils.py
sidtuladhar: yeah this is more explicit which is better. Also, adding a comment to show what it looks like is good since whenever I'm reading code I'm always like what does this return and what does it look like

paasta_tools/cli/cmds/validate.py
sidtuladhar: yeah i was gonna change the comment since it sounded a bit caveman-like lol

paasta_tools/utils.py
nemacysts: haha, same!

hopefully some of my old (internal) ownership service code no longer exists: that had some gnarly dict types that were utterly incomprehensible without an example comment 🤣

paasta_tools/utils.py
```diff
@@ -2545,6 +2552,10 @@ def get_pool_node_affinities(self) -> Dict[str, Dict[str, List[str]]]:
         """Node selectors that will be applied to all Pods in a pool"""
         return self.config_dict.get("pool_node_affinities", {})
 
+    def get_pool_limits(self) -> Dict[str, Dict[str, Any]]:
```
nemacysts:
```suggestion
    def get_pool_limits(self) -> Dict[str, Dict[str, PoolLimits]]:
```
right?

EvanKrall (commented):
My main concern is that some jobs that are currently below the threshold might be bumped over the threshold by autotune, which I think would cause the autotune jenkins job to fail. Maybe we need to make the autotune job clamp at max_cpus or change the pool automatically?

The other approach would be to have Paasta automatically map `pool: stable` to `stable` / `stable-giant` behind the scenes. (Or keep all the boxes in the stable pool as `pool: stable` but have some sort of sub-pool label or something.)

sidtuladhar:
If we have a clamp, it could lead to the service having to throttle. As for the automatic pool change, we could:

`scale up -> move to recommended pool -> scale down -> stay in recommended pool`

but i would be concerned if we have a week where traffic is much higher, and we get a bunch of jobs moving to the recommended pool and staying there forever.

`scale up -> move to recommended pool -> scale down -> move back down to original pool`

in this case we would just have to additional logic, and maybe there are consequences to moving services between pools  (like we would be need to enforce stable and stable-giant are exactly the same config besides machines)

nemacysts:
> My main concern is that some jobs that are currently below the threshold might be bumped over the threshold by autotune, which I think would cause the autotune jenkins job to fail. Maybe we need to make the autotune job clamp at max_cpus or change the pool automatically?

@EvanKrall i think we mostly want this for the `tron` pool (where pods aren't autotuned) since most stuff in the stable pool is actually pretty small (basically the only things above 4 would be non-pyramid services) - so we should be fine with this a a quick stopgap to prevent folks from bumping into any sharp edges with the new tron pool limits

> The other approach would be to have Paasta automatically map `pool: stable` to `stable` / `stable-giant` behind the scenes. (Or keep all the boxes in the stable pool as `pool: stable` but have some sort of sub-pool label or something.)

that said, i really like this idea! maybe we could merge this -> follow-up with the background mapping idea?

the mapping idea could probably actually end up re-using the same config format :p

sidtuladhar:
mo labels mo solutions

nemacysts (dismissed):

ilkinmammadzada (dismissed):

EvanKrall (dismissed):
Good point about the Tron pool being mostly un-autotuned.

sidtuladhar:
merge conflict

sidtuladhar:
another merge conflct :sadcat:

mbrankin-art (approved):

PR #4359, opened 2026-07-30
Scaffold basic alertmanager autorollback configs

nemacysts:
There'll very likely be more than these in the near future, but just
adding a couple simple ones to get started

### DIFF
--- a/paasta_tools/utils.py
+++ b/paasta_tools/utils.py
@@ -189,6 +189,7 @@ CAPS_DROP = [
 class RollbackTypes(Enum):
     AUTOMATIC_SLO_ROLLBACK = "automatic_slo_rollback"
     AUTOMATIC_METRIC_ROLLBACK = "automatic_metric_rollback"
+    AUTOMATIC_ALERTMANAGER_ROLLBACK = "automatic_alertmanager_rollback"
     AUTOMATIC_CRASHLOOP_ROLLBACK = "automatic_crashloop_rollback"
     USER_INITIATED_ROLLBACK = "user_initiated_rollback"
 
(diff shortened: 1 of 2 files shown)


paasta_tools/utils.py
```diff
@@ -2673,6 +2677,15 @@ def get_mark_for_deployment_should_ping_for_unhealthy_pods(self) -> bool:
...
         )
 
+    def get_alertmanager_url(self) -> Optional[str]:
+        return self.config_dict.get("alertmanager_url", None)
+
+    def get_enable_alertmanager_rollback(self) -> bool:
+        return self.config_dict.get("enable_alertmanager_rollback", False)
+
+    def get_alertmanager_poll_interval_s(self) -> int:
+        return self.config_dict.get("alertmanager_poll_interval_s", 30)
```
ilkinmammadzada, line 2687: Should we avoid lower numbers here?

paasta_tools/utils.py
```diff
@@ -2673,6 +2677,15 @@ def get_mark_for_deployment_should_ping_for_unhealthy_pods(self) -> bool:
             "mark_for_deployment_should_ping_for_unhealthy_pods", True
         )
 
+    def get_alertmanager_url(self) -> Optional[str]:
+        return self.config_dict.get("alertmanager_url", None)
```
ilkinmammadzada, line 2681: was wondering if we need to validate URL here?

paasta_tools/utils.py
nemacysts: yea, the schema will enforce a 30s minimum for the user-controllable part of this (in deploy.yaml) - i think we could also probably bump this to 1min or something as well

nemacysts: (although, maybe i should rename this: this is really just the default poll interval)

paasta_tools/utils.py
nemacysts: we'll really just set this once in SystemPaastaConfig and never touch it again - this was mostly to try to not keep exposing internal urls in public github :p 

(we have a single alertmanager instance we'd want to use for this)

ilkinmammadzada (approved):
Ahh right. both are system config. shipit

PR #4362, opened 2026-07-31
Flesh out filtering for custom alerts based on deploy group

nemacysts:
We can't just filter by service name since an instance in a
cluster/deploy group with no changes might be encountering issues, and
those are (likely) unrelated to our current deployment


nemacysts:
err, not sure how i made 2 PRs here

PR #4363, opened 2026-07-31
Build deploy group to AlertManager filter mapping

nemacysts:
We can't just filter by service name since an instance in a
cluster/deploy group with no changes might be encountering issues, and
those are (likely) unrelated to our current deployment


ilkinmammadzada (dismissed):

sidtuladhar (approved):
LGTM

PR #4366, opened 2026-08-03
Add rollback filtering for Default PaaSTA Error Alerting

nemacysts:
As mentioned in the CEP, these (pre-existing) alerts have a different
set of labels, so we can't filter in quite the same way as we do for the
"custom" alerts.

However, as (also) mentioned in the CEP, we can still construct a filter
that will match these alerts given the data we have available.

This hopefully isn't too complex (and hopefully my comments are useful),
but it is slightly funky :p - although, imo, mostly due to needing to
figure out the "real" region for a cluster + needing to handle
multi-registration instances :p

### DIFF
--- a/paasta_tools/cli/cmds/mark_for_deployment.py
+++ b/paasta_tools/cli/cmds/mark_for_deployment.py
@@ -376,6 +376,53 @@ def can_user_deploy_service(deploy_info: Dict[str, Any], service: str) -> bool:
     return True
 
 
+def _build_default_error_alert_filter(
+    service: str,
+    instance_configs_per_cluster: Dict[str, List[LongRunningServiceConfig]],
+) -> List[str]:
+    """Build a filter for Default PaaSTA Error Alerts.
+
+    These alerts use a composite `instance` label: {region}.{service}.{server_namespace}
+    (with an optional .{endpoint} suffix for per-endpoint alerts).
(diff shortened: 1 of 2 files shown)


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -376,6 +376,53 @@ def can_user_deploy_service(deploy_info: Dict[str, Any], service: str) -> bool:
...
 
+def _build_default_error_alert_filter(
+    service: str,
+    instance_configs_per_cluster: Dict[str, List[LongRunningServiceConfig]],
+) -> List[str]:
+    """Build a filter for Default PaaSTA Error Alerts.
+
+    These alerts use a composite `instance` label: {region}.{service}.{server_namespace}
+    (with an optional .{endpoint} suffix for per-endpoint alerts).
+    NOTE: server_namespace above is the same as the PaaSTA mesh registration
```
ilkinmammadzada, line 387: `server_namespace` - I guess we'll rollback all instances if they use "broken" server_namespace, right?

nemacysts: only if they're in the deploy group that's being deployed to :D

ilkinmammadzada (approved):

PR #4368, opened 2026-08-05
Correctly handle setpoint for Piscina autoscaling

nemacysts:
We didn't notice that the Piscina HPA was setting the target to 1 - which, in the shared rules code, means that we're essentially setting the setpoint to 100% utilization.

We should follow this up with moving the 2 remaining services on `type: uwsgi` autoscaling to `type: worker-load` and then deleting the uwsgi/and gunicorn autoscaling code (since those are superseded by worker-load)

NOTE: I think that we've also had broken autoscaling for the 2 type: uwsgi services for this time, but interestingly enough, no one seems to have noticed :p

### DIFF
--- a/paasta_tools/kubernetes_tools.py
+++ b/paasta_tools/kubernetes_tools.py
@@ -881,8 +881,26 @@ class KubernetesDeploymentConfig(LongRunningServiceConfig):
             METRICS_PROVIDER_UWSGI,
             METRICS_PROVIDER_PISCINA,
             METRICS_PROVIDER_GUNICORN,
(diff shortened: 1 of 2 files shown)


KaspariK (approved):
I'm not the expert but fwiw this does seem right

EvanKrall (approved):

ilkinmammadzada (approved):

sidtuladhar (approved):
LGTM, thanks for catching this @nemacysts!

sidtuladhar:
I'm going to merge this and test in dev/stage. Hope that's ok @nemacysts

PR #4369, opened 2026-08-06
Disabling single replica check

mbrankin-art:
Had an onpoint where it was stopping people from merging because instance hadn't been set for non prod services
https://yelp.slack.com/archives/CA05GTDB9/p1786031902645339


nemacysts (approved):
ah, i should have asked if this passed on all existing files first :p 

imo, fully reverting is maybe a little more preferable - but this is fine too!

mbrankin-art:
> ah, i should have asked if this passed on all existing files first :p
> 
> imo, fully reverting is maybe a little more preferable - but this is fine too!

i assume this would be safe until i fix it asap or is it preferred to revert?

PR #4371, opened 2026-08-07
creating single replica commit hook

mbrankin-art:
1. Creating a commit check that will look for single replica instances. This is done in order to avoid an SPOF and to allow for recycling. 
3. The commit hook will only target services running in the production environment. 
5. If the instance is set implicitly (default) then an info message will be shown but passes the check. 
7. Skips canary services.
8. The validation check can be skipped using the `# override-single-replica (PROJ-1234)` comment in the respective service config files.

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -45,7 +45,6 @@ from jsonschema import ValidationError
 from jsonschema import exceptions
 from mypy_extensions import TypedDict
 from ruamel.yaml import YAML
-from ruamel.yaml import SafeConstructor
 from ruamel.yaml.comments import CommentedMap
 
 from paasta_tools import yaml_tools as yaml
@@ -58,6 +57,7 @@ from paasta_tools.cli.utils import guess_service_name
 from paasta_tools.cli.utils import info_message
 from paasta_tools.cli.utils import lazy_choices_completer
 from paasta_tools.cli.utils import success
+from paasta_tools.eks_tools import EksDeploymentConfig
 from paasta_tools.kubernetes_tools import sanitise_kubernetes_name
 from paasta_tools.long_running_service_tools import DEFAULT_AUTOSCALING_SETPOINT
 from paasta_tools.long_running_service_tools import DEFAULT_PROMQL_AUTOSCALING_SETPOINT
@@ -72,6 +72,7 @@ from paasta_tools.long_running_service_tools import METRICS_PROVIDER_UWSGI_V2
 from paasta_tools.long_running_service_tools import METRICS_PROVIDER_WORKER_LOAD
 from paasta_tools.long_running_service_tools import LongRunningServiceConfig
 from paasta_tools.monitoring_tools import get_sensu_team_data
+from paasta_tools.paasta_service_config_loader import PaastaServiceConfigLoader
 from paasta_tools.secret_tools import get_secret_name_from_ref
 from paasta_tools.secret_tools import is_secret_ref
 from paasta_tools.secret_tools import is_shared_secret
@@ -138,9 +139,11 @@ SCHEMA_TYPES = {
 # comment
 OVERRIDE_CPU_AUTOTUNE_ACK_PATTERN = r"#\s*override-cpu-setting\s+\(.+[A-Z]+-[0-9]+.+\)"
 
-# we expect a comment that looks like # override-cpu-burst PROJ-1234
-# but we don't have a $ anchor in case users want to add an additional
-# comment
+# we expect a comment that looks like # override-single-replica (PROJ-1234)
+OVERRIDE_SINGLE_REPLICA_ACK_PATTERN = (
+    r"#\s*override-single-replica\s+\(.+[A-Z]+-[0-9]+.+\)"
+)
+
 OVERRIDE_CPU_BURST_ACK_PATTERN = r"#\s*override-cpu-burst\s+\(.+[A-Z]+-[0-9]+.+\)"
 # for now, double the autotune cap to give people the benefit of the doubt
 # if we see that people are still misusing this configuration, we can lower
@@ -148,6 +151,11 @@ OVERRIDE_CPU_BURST_ACK_PATTERN = r"#\s*override-cpu-burst\s+\(.+[A-Z]+-[0-9]+.+\
 CPU_BURST_THRESHOLD = 2
 
 K8S_TYPES = {"eks", "kubernetes"}
+
+
+SINGLE_REPLICA_VALIDATED_ECOSYSTEMS = {
+    "prod",
+}
 
 _PROMQL_ONLY_FIELDS = {"metrics_query", "series_query", "target_type", "resources"}
 
@@ -379,9 +387,8 @@ def get_config_file_dict(file_path: str, use_ruamel: bool = False) -> Dict[Any, 
                 # sets disk: 100 -> an instance uses that template and overwrites
                 # it with disk: 1000)
                 ruamel_loader.allow_duplicate_keys = True
-                # Note: we do NOT use flatten_mapping here because it breaks simple
-                # override patterns. For nested merge patterns, we load a flattened
-                # version separately for comment detection only.
+                # Note: we do NOT use flatten_mapping here because it breaks
+                # comment detection on merge-inherited keys.
                 return ruamel_loader.load(config_file)
             else:
                 return yaml.safe_load(config_file)
@@ -652,27 +659,13 @@ def validate_unique_instance_names(service_path):
     return check_passed
 
 
-def _get_config_flattened(file_path: str) -> CommentedMap:
-    """Load config with flatten_mapping enabled (for nested merge pattern comment detection)"""
-    config_file = get_file_contents(file_path)
-    ruamel_loader = YAML(typ="rt")
-    ruamel_loader.allow_duplicate_keys = True
-    ruamel_loader.Constructor.flatten_mapping = SafeConstructor.flatten_mapping
-    return ruamel_loader.load(config_file)
-
-
 def _get_comments_for_key(
     data: CommentedMap,
     key: Any,
-    full_config: Optional[Dict[Any, Any]] = None,
-    key_value: Any = None,
-    full_config_flattened: Optional[Dict[Any, Any]] = None,
 ) -> Optional[str]:
     # this is a little weird, but ruamel is returning a list that looks like:
     # [None, None, CommentToken(...), None] for some reason instead of just a
     # single string
-    # Sometimes ruamel returns a recursive list of CommentTokens as well that looks like
-    # [None, None, [CommentToken(...),CommentToken(...),None], CommentToken(...), None]
     def _flatten_comments(comments):
         for comment in comments:
             if comment is None:
@@ -683,38 +676,31 @@ def _get_comments_for_key(
                 yield comment.value
 
     raw_comments = [*_flatten_comments(data.ca.items.get(key, []))]
-    if not raw_comments:
-        # If we didn't find a comment in the instance itself, check if this key
-        # might be inherited from a template. Look for ANY other instance/template
-        # in the config that has the same key with the same value and a comment.
-        if full_config is not None and key_value is not None:
-            for config_key, config_value in full_config.items():
-                if isinstance(config_value, CommentedMap):
-                    if config_value.get(key) == key_value:
-                        other_comments = [
-                            *_flatten_comments(config_value.ca.items.get(key, []))
-                        ]
-                        if other_comments:
-                            return "".join(other_comments)
+    if raw_comments:
+        return "".join(raw_comments)
 
-        # If still not found and we have a flattened config, check there
-        # (flattened config is needed for nested merges)
-        if full_config_flattened is not None:
-            for config_key, config_value in full_config_flattened.items():
-                if isinstance(config_value, CommentedMap):
-                    if config_value.get(key) == key_value:
-                        flattened_comments = [
-                            *_flatten_comments(config_value.ca.items.get(key, []))
-                        ]
-                        if flattened_comments:
-                            return "".join(flattened_comments)
-
-        # return None so that we don't return an empty string below if there really aren't
-        # any comments
+    # If the instance explicitly sets the key, the comment must be on that
+    # line — don't follow the merge chain.
+    if key in getattr(data, "_ok", set()):
         return None
-    # joining all comments together before returning them
-    comment = "".join(raw_comments)
-    return comment
+
+    # Follow the <<: *anchor merge chain to find comments on inherited keys
+    visited: Set[int] = set()
+    sources = list(getattr(data, "merge", None) or [])
+    while sources:
+        _, src = sources.pop(0)
+        src_id = id(src)
+        if src_id in visited:
+            continue
+        visited.add(src_id)
+        if not isinstance(src, CommentedMap):
+            continue
+        src_comments = [*_flatten_comments(src.ca.items.get(key, []))]
+        if src_comments:
+            return "".join(src_comments)
+        sources.extend(getattr(src, "merge", None) or [])
+
+    return None
 
 
 def __is_templated(service: str, soa_dir: str, cluster: str, workload: str) -> bool:
@@ -854,11 +840,7 @@ def validate_autoscaling_configs(service_path: str) -> bool:
 
                             # we need access to the comments, so we need to read the config with ruamel to be able
                             # to actually get them in a "nice" automated fashion
-                            config_file_path = os.path.join(
-                                soa_dir,
-                                service,
-                                f"{instance_config.get_instance_type()}-{cluster}.yaml",
-                            )
+                            config_file_path = instance_config.get_config_path()
                             config = get_config_file_dict(
                                 config_file_path,
                                 use_ruamel=True,
@@ -875,15 +857,9 @@ def validate_autoscaling_configs(service_path: str) -> bool:
                                 # cpu autoscaled, but using autotuned values - can skip
                                 continue
 
-                            # Load flattened config for comment detection (handles nested merges)
-                            config_flattened = _get_config_flattened(config_file_path)
-
                             cpu_comment = _get_comments_for_key(
                                 data=config[instance],
                                 key="cpus",
-                                full_config=config,
-                                key_value=config[instance].get("cpus"),
-                                full_config_flattened=config_flattened,
                             )
                             # we could probably have a separate error message if there's a comment that doesn't match
                             # the ack pattern, but that seems like overkill - especially for something that could cause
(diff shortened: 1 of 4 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +935,77 @@ def validate_min_max_instances(service_path):
     return returncode
 
 
+def check_for_single_replica(instance_config: InstanceConfig) -> bool:
+    return (
+        isinstance(instance_config, LongRunningServiceConfig)
+        and instance_config.get_instance_type() in K8S_TYPES
+        and not instance_config.get_deploy_group().endswith(".canary")
```
nemacysts: just scribing from slack: imo, having an instance name allowlist will probably be a better route - deploy groups can be named pretty arbitrarily (although, so can instances :p) and the allowlist for canary instances is probably going to be smaller than an equivalent allowlist for deploy groups

paasta_tools/cli/cmds/validate.py
```diff
@@ -937,6 +935,77 @@ def validate_min_max_instances(service_path):
...
+
+
+def validate_single_replica_instances(service_path: str) -> bool:
+    soa_dir, service = path_to_soa_dir_service(service_path)
+    returncode = True
+    system_paasta_config = load_system_paasta_config()
+
+    for cluster in list_clusters(service, soa_dir):
+        ecosystem = system_paasta_config.get_ecosystem_for_cluster(cluster)
+        if ecosystem != "prod":
```
nemacysts: might be nice to make this a set or something,  that way this isn't as hardcoded :p 

e.g., 
```Python
SINGLE_REPLICA_VALIDATED_ECOSYSTEMS = {
    # for now, we'll just enforce this in prod
    # since folks tend to tolerate downtime in dev
    "prod",
}

File: paasta_tools/cli/cmds/validate.py [hunk 3]
```diff
@@ -937,6 +935,77 @@ def validate_min_max_instances(service_path):
...
+            continue
+
+        for instance, instance_config in load_all_instance_configs_for_service(
+            service=service, cluster=cluster, soa_dir=soa_dir
+        ):
+            if not check_for_single_replica(instance_config):
+                continue
+            assert isinstance(instance_config, LongRunningServiceConfig)
```
nemacysts: oh actually, since we're really just looking at the eks-* files: maybe we should use PaastaServiceConfigLoader so that we can drop the `assert`? (and pre-filter out all the stuff we don't care about :p)

e.g., https://github.com/Yelp/paasta/blob/f6614757816377317898ede68d90476fcd9db1cc/paasta_tools/setup_prometheus_adapter_config.py#L1458-L1465

paasta_tools/cli/cmds/validate.py
mbrankin-art: gotcha will do this!

paasta_tools/cli/cmds/validate.py
mbrankin-art: so i ended up using the reg and allow list check - reasons is because some services didn't have the same registration e.g. x_canary was falling through. with both checks we are catching both of these edge cases

thoughts? @nemacysts @EvanKrall

paasta_tools/cli/cmds/validate.py
```diff
@@ -1004,6 +1009,107 @@ def check_secrets_for_instance(
...
+
+
+def validate_single_replica_instances(service_path: str) -> bool:
+    soa_dir, service = path_to_soa_dir_service(service_path)
+    returncode = True
+    system_paasta_config = load_system_paasta_config()
+
+    for cluster in list_clusters(service, soa_dir):
+        ecosystem = system_paasta_config.get_ecosystem_for_cluster(cluster)
+        if ecosystem != "prod":
```
```suggestion
        if ecosystem not in SINGLE_REPLICA_VALIDATED_ECOSYSTEMS:
```
(looks like we added the constant, but didn't commit using it :p)

paasta_tools/cli/cmds/validate.py
```diff
@@ -1004,6 +1009,107 @@ def check_secrets_for_instance(
...
+                other_replicas = config.get_instances()
+
+            if instance_config.is_autoscaling_enabled():
+                current_replicas = instance_config.get_max_instances()
+            else:
+                current_replicas = instance_config.get_instances()
+            if other_replicas is not None and other_replicas > 1:
+                if current_replicas is not None and current_replicas == 1:
+                    return True
+    return False
```
nemacysts, line 1017: hmm, i wonder if we should pre-process things a bit before we get in here?

we're doing a lot of repeated work since we're essentially doing:
```
for _ in all_configs:
    for __ in all_configs:
        ...
```

when i think we could maybe pass down a dict formatted such that we can go directly to what other instances share the same registration?

nemacysts: (might also be good to add a docstring or some comments here - this is probably pretty clear to krall, you, and me...but future us and anyone else not in that slack convo might be a little confused as to what we're doing here :p)

tests/cli/test_cmds_validate.py
```diff
@@ -234,6 +246,114 @@ def test_validate_min_max_instances_success(
...
+    mock_get_file_contents.return_value = f"""
+---
+fake_instance1:
+  instances: {instances} {comment}
+"""
+
+    assert validate_single_replica_instances("fake-service-path") is expected
+
+
+class TestIsCanary:
```
nemacysts: !choose remove test class to be consistent, start using these more in general

paasta_tools/cli/cmds/validate.py
```diff
@@ -149,6 +151,13 @@ class AutoscalingValidationError(SoaValidationError):
 
 K8S_TYPES = {"eks", "kubernetes"}
 
+
+SINGLE_REPLICA_VALIDATED_ECOSYSTEMS = {
+    "prod",
+}
+
+CANARY_INSTANCE_NAME_ALLOWLIST = {"canary"}
```
nemacysts: we probably won't expand the checked ecosystems anytime soon, but we might add more canary allowlists if our registration check isn't bulletproof - so this might be better off in SystemPaastaConfig

...but also, maybe we can just wait until we need to add another entry and move things then :p

paasta_tools/cli/cmds/validate.py
```diff
@@ -1004,6 +1009,107 @@ def check_secrets_for_instance(
...
+                config_flat = _get_config_flattened(config_file_path)
+
+                if config_flat[instance].get(replica_key) is None:
+                    print(
+                        info_message(
+                            f"Instance {instance} on cluster {cluster} has only 1 replica (implicit default). "
+                            f"We recommend setting `instances` / `min instances` to a value greater than 1 to avoid SPOF. "
+                        )
+                    )
+                    continue
```
nemacysts, line 1075: sorry, my brain is mush: can you remind me why we're only warning for this case? i vaguely remember us talking about this in a call or in slack, but i'm blanking :p

paasta_tools/cli/cmds/validate.py
mbrankin-art: so funny , will change that!

paasta_tools/cli/cmds/validate.py
mbrankin-art: yeah 100% i'll annotate it better to make it a bit clearer for everyone!

paasta_tools/cli/cmds/validate.py
mbrankin-art: so i used an info message here because some services may not have `instances` or `min_instances` set and just defaulting on 1 so created an info message that would pick this up and recommend setting it explicitly

paasta_tools/cli/cmds/validate.py
mbrankin-art: https://github.yelpcorp.com/sysgit/puppet/pull/16844 - added it to SystemPaastaConfig for future-proofing

paasta_tools/cli/cmds/validate.py
nemacysts: ah gotcha - thinking a little harder about this, i think we either need to have this load the autotuned configs as well or just skip some of these for now (and add a TODO :p): most autoscaled stuff will ideally not have min_instances set and rely on the autotuned values, which we set to a >1 value as a floor

nemacysts: and i guess: do we not want to validate the default `instances: 1` case the same as an explicit `instances: 1` config? or is the idea that we'll eventually set the default value for a missing `instances` to something >1?

paasta_tools/cli/cmds/validate.py
```diff
@@ -1004,6 +1007,123 @@ def check_secrets_for_instance(
...
+        for registration in instance_config.get_registrations():
+            if registration not in registration_dict:
+                registration_dict[registration] = []
+            registration_dict[registration].append(instance_config)
+    return registration_dict
+
+
+def is_canary(
+    instance_config: LongRunningServiceConfig,
+    reg_lookup: Dict[str, List[LongRunningServiceConfig]],
```
nemacysts: thoughts on a rename like 
```python
    registration_to_instances: Dict[str, List[LongRunningServiceConfig]],
```
?

nemacysts: and/or a similar name for `registration_lookup()`?

paasta_tools/utils.py
```diff
@@ -2732,6 +2742,9 @@ def get_skip_cpu_burst_validation_services(self) -> List[str]:
     def get_skip_unique_instance_name_validation_services(self) -> List[str]:
         return self.config_dict.get("skip_unique_instance_name_validation", [])
 
+    def get_canary_instance_name_allowlist(self) -> List[str]:
+        return self.config_dict.get("canary_instance_name_allowlist", ["canary"])
```
nemacysts: !choose `common_canary_instance_names`, make the name more specific to the check where it'll be used

nemacysts: (this is a little too generic of a name atm, i think)

paasta_tools/cli/cmds/validate.py
mbrankin-art: `registration_to_instances` sounds good to me!

paasta_tools/cli/cmds/validate.py
mbrankin-art: > do we not want to validate the default instances: 1 case the same as an explicit instances: 1 config? or is the idea that we'll eventually set the default value for a missing instances to something >1?

with regards to this i imagined the idea was that services with `instances` or `min_instances` would eventaully set the default value for `instances` / `min_instances` to something > 1

paasta_tools/utils.py
mbrankin-art: will rename it to be a bit more specific!

paasta_tools/cli/cmds/validate.py
```diff
@@ -1004,6 +1007,123 @@ def check_secrets_for_instance(
...
+        if ecosystem not in SINGLE_REPLICA_VALIDATED_ECOSYSTEMS:
+            continue
+
+        all_configs: List[LongRunningServiceConfig] = [
+            instance_config
+            for _, instance_config in load_all_instance_configs_for_service(
+                service=service, cluster=cluster, soa_dir=soa_dir
+            )
+            if isinstance(instance_config, LongRunningServiceConfig)
+        ]
```
nemacysts: hmm, i did a last-second check and it looks like the operator configs subclass this, so wouldn't this also continue checking flink/cassandra/etc?

i was gonna suggest that we refactor this in another PR to use PaastaServiceConfigLoader to get rid of the isinstance/not load all configs - but i think we might actually wanna do that now (or `s/LongRunningServiceConfig/EksDeploymentConfig`)

paasta_tools/cli/cmds/validate.py
```diff
@@ -1004,6 +1007,123 @@ def check_secrets_for_instance(
...
+        reg_lookup = instance_registration_lookup(all_configs)
+        canary_allowlist = system_paasta_config.get_common_canary_instance_names()
+
+        for instance_config in all_configs:
+            if is_canary(instance_config, reg_lookup):
+                continue
+            instance = instance_config.get_instance()
+            if any(pattern in instance for pattern in canary_allowlist):
+                continue
```
nemacysts: not sure if it matters, but would it maybe be faster to check the allowlist first?

paasta_tools/cli/cmds/validate.py
```diff
@@ -1004,6 +1007,123 @@ def check_secrets_for_instance(
...
+
+            if replicas == 1:
+                config_file_path = instance_config.get_config_path()
+                config_flat = _get_config_flattened(config_file_path)
+
+                if config_flat[instance].get(replica_key) is None:
+                    print(
+                        """Info message is used in case a service hasn't explicitly set the number of replicas,
+                        and is set to the default of 1. We don't want to fail validation but we do want to warn them
+                        that they should set it explicitly.""",
```
nemacysts: should this be a normal comment rather than a print?

nemacysts: (and maybe add something about what you mentioned below re: "we're hoping update the default values to be >1 in the future"?)

nemacysts (commented):
i think the only blockers are:
* the LongRunningServiceConfig comment
* removing the info message print

otherwise, looks good to me

mbrankin-art:
@nemacysts 
> oh gotcha - thinking a little harder about this, i think we either need to have this load the autotuned configs as well or just skip some of these for now (and add a TODO :p): most autoscaled stuff will ideally not have min_instances set and rely on the autotuned values, which we set to a >1 value as a floor

do you think it's worth adding the autotuned config to this pr? as it does make sense to add it

paasta_tools/cli/cmds/validate.py
mbrankin-art: Yeah sorry i went ham on the doc string - changed it back locally just need to push it up

paasta_tools/cli/cmds/validate.py
mbrankin-art: yeah makes sense to do it before the loop

nemacysts:
> do you think it's worth adding the autotuned config to this pr? as it does make sense to add it

@mbrankin-art uhh, i guess it depends on how much more work it is :p

if it's not too hard, i think it's fine to do it here too - but i think it's also fine to follow-up with a separate PR to expand this

mbrankin-art:
> > do you think it's worth adding the autotuned config to this pr? as it does make sense to add it
> 
> @mbrankin-art uhh, i guess it depends on how much more work it is :p
> 
> if it's not too hard, i think it's fine to do it here too - but i think it's also fine to follow-up with a separate PR to expand this

okay i'll sort the other things out and timebox it to see!!

paasta_tools/utils.py
```diff
@@ -2732,6 +2742,9 @@ def get_skip_cpu_burst_validation_services(self) -> List[str]:
     def get_skip_unique_instance_name_validation_services(self) -> List[str]:
         return self.config_dict.get("skip_unique_instance_name_validation", [])
 
+    def get_common_canary_instance_names(self) -> List[str]:
+        return self.config_dict.get("canary_instance_name_allowlist", ["canary"])
```
nemacysts: (i'd maybe also have this match the getter name (`common_canary_instance_names`) just for consistency)

paasta_tools/cli/cmds/validate.py
```diff
@@ -1004,6 +1010,135 @@ def check_secrets_for_instance(
...
+            continue
+
+        all_configs: List[EksDeploymentConfig] = list(
+            config_loader.instance_configs(
+                cluster=cluster, instance_type_class=EksDeploymentConfig
+            )
+        )
+        reg_lookup = instance_registration_lookup(all_configs)
+        canary_allowlist = system_paasta_config.get_common_canary_instance_names()
+        auto_config = load_service_instance_auto_configs(
```
nemacysts: i think that we don't actually need to do this? PSCL (PaastaServiceConfigLoader) seems to already call `load_service_instance_configs()` (see https://github.com/Yelp/paasta/blob/b329a04315c978818f58c0e7740d3ff1249dc065/paasta_tools/paasta_service_config_loader.py#L134) which will merge the explicit and autotuned configs for us :D

nemacysts: i.e., we can remove the
```Python
            # autotuned values set min_instances to a >1 floor, so if the autotuned
            # config indicates this instance isn't single-replica, skip it
            instance_auto_config = auto_config.get(instance, {})
            if (
                instance_auto_config.get("max_instances", 0) > 1
                or instance_auto_config.get("min_instances", 0) > 1
            ):
                continue
```
block

nemacysts (commented):
just some small comments, i think we're almost done!

paasta_tools/cli/cmds/validate.py
mbrankin-art: gotcha will remove this in latest commit!

paasta_tools/cli/cmds/validate.py
```diff
@@ -1004,6 +1009,124 @@ def check_secrets_for_instance(
...
+        for instance_config in all_configs:
+            instance = instance_config.get_instance()
+            if any(pattern in instance for pattern in canary_allowlist):
+                continue
+            if is_canary(instance_config, reg_lookup):
+                continue
+
+            if instance_config.is_autoscaling_enabled():
+                max_instance_count = instance_config.get_max_instances()
+                if max_instance_count is not None and max_instance_count > 1:
```
nemacysts, line 1052: technically this will never be None since is_autoscaling_enabled() will only return true if this isn't None, but i wouldn't worry about this for now (we can always fix this in another PR so that you don't have to wait for me to review again :p)

paasta_tools/cli/cmds/validate.py
```diff
@@ -1004,6 +1009,124 @@ def check_secrets_for_instance(
...
+                        string=comment,
+                    )
+                    is None
+                ):
+                    returncode = False
+                    print(
+                        failure(
+                            msg=f"Instance {instance} on cluster {cluster} has only 1 instance / min_instance. "
+                            f"\nSingle replicas are a SPOF and we do not encourage it. "
+                            f"\nAdd a comment to acknowledge: # override-single-replica (PROJ-1234)",
```
```suggestion
                            f"\nAdd a comment like the following to acknowledge: # override-single-replica (PROJ-1234)",
```

nemacysts (dismissed):

nemacysts (dismissed):
we'll still want to wait to merge this until we've added the override comment to all the existing failures + sent some comms - but looks good otherwise!

paasta_tools/cli/cmds/validate.py
```diff
@@ -683,38 +673,26 @@ def _flatten_comments(comments):
...
+            continue
+        visited.add(src_id)
+        if not isinstance(src, CommentedMap):
+            continue
+        src_comments = [*_flatten_comments(src.ca.items.get(key, []))]
+        if src_comments:
+            return "".join(src_comments)
+        sources.extend(getattr(src, "merge", None) or [])
+
+    return None
```
nemacysts, line 705: @mbrankin-art would we follow the chain with something like this?

```
some_random_worker_A: &cool_template
    instances: 1 # override-single-replica



some_random_worker_B: &other_template
   <<: *cool_template
   instances: 2
   mem: 12345

some_random_worker_C:
   <<: *other_template
   instances: 1
```

(in this case, i'd expect A and B to validate and C to fail)

paasta_tools/cli/cmds/validate.py
```diff
@@ -652,27 +659,10 @@ def validate_unique_instance_names(service_path):
...
     key: Any,
-    full_config: Optional[Dict[Any, Any]] = None,
-    key_value: Any = None,
-    full_config_flattened: Optional[Dict[Any, Any]] = None,
 ) -> Optional[str]:
-    # this is a little weird, but ruamel is returning a list that looks like:
-    # [None, None, CommentToken(...), None] for some reason instead of just a
-    # single string
-    # Sometimes ruamel returns a recursive list of CommentTokens as well that looks like
-    # [None, None, [CommentToken(...),CommentToken(...),None], CommentToken(...), None]
```
nemacysts, line 675: lol, why did i repeat this twice

anyway: we probably wanna keep the first 
```
    # this is a little weird, but ruamel is returning a list that looks like:
    # [None, None, CommentToken(...), None] for some reason instead of just a
    # single string
```
comment 'cause this is definitely a little funky

paasta_tools/cli/cmds/validate.py
mbrankin-art: @nemacysts yes we would C to fail since it has explicitly stated an `instances` value and has not added the override comment - do you agree that its the correct behaviour?

paasta_tools/cli/cmds/validate.py
mbrankin-art: gotcha will edit this

paasta_tools/cli/cmds/validate.py
nemacysts: yup yup - just wanted to make sure that was the behavior :p

paasta_tools/cli/cmds/validate.py
nemacysts, line 670: (it was fine to remove this part of rthe comment since it's basically the same thing at the bit above)

nemacysts (dismissed):
nice!

sidtuladhar (approved):

PR #4372, opened 2026-08-14
Accept OCI manifests for checking image existence

tzhu-yelp:
# Problem

Image verification check incorrectly claims image is not already there, because the newly pushed image was pushed as OCI rather than Docker.

Good thing we don't actually care about that, so just need to advertise that we are OK with getting OCI format stuff for manifests

# Solution

Advertise that in headers

# Testing

Nothing yet


piax93 (approved):
Ah yes, the classic

nemacysts (approved):

nemacysts:
@tzhu-yelp i tested this with `paasta verify-image-exists` (which calls this code) and it seems to work :)

PR #4373, opened 2026-08-14
Refactor the cpu limit check

sidtuladhar:
Refactor the cpu limit check:

- remove recommended pool
- require comment next to `cpu` in yelpsoa to override

### DIFF
--- a/paasta_tools/cli/cmds/validate.py
+++ b/paasta_tools/cli/cmds/validate.py
@@ -943,6 +943,8 @@ def validate_pool_limits(service_path: str) -> bool:
     Validate that services in specific pools won't exceed per-node capacities.
     For the most part, this isn't normally an issue - but there are several pools where
     folks tend to want to run extra-large workloads that won't fit (e.g., large single-node batches).
+
+    Users can override this check by adding any comment next to `cpus` in their yelpsoa config.
     """
     soa_dir, service = path_to_soa_dir_service(service_path)
     returncode = True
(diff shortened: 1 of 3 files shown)


paasta_tools/cli/cmds/validate.py
```diff
@@ -958,18 +960,46 @@ def validate_pool_limits(service_path: str) -> bool:
...
-                print(
-                    failure(
-                        f"""{service}.{instance} in {cluster} has {cpu} CPUs, which exceeds the limit of {pool_limits['max_cpus']} for the {pool} pool.
-                        If you need to run a workload with this many CPUs, consider using the {pool_limits['recommended_pool']} pool instead.""",
-                        "",
-                    )
+            if __is_templated(
+                service, soa_dir, cluster, workload="kubernetes"
+            ) or __is_templated(service, soa_dir, cluster, workload="eks"):
+                continue
```
nemacysts: hmm, do we not copy comments from the .in files to the .yaml files?

nemacysts: (we can probably also drop the `workload="kubernetes"` branch - i don't see any .in files for kubernetes-*)

paasta_tools/cli/cmds/validate.py
```diff
@@ -958,18 +960,46 @@ def validate_pool_limits(service_path: str) -> bool:
...
+
+            config_file_path = os.path.join(
+                soa_dir,
+                service,
+                f"{instance_config.get_instance_type()}-{cluster}.yaml",
+            )
```
nemacysts, line 980: @mbrankin-art has a nice helper for this in his validation PR, but if this gets merged first, we can always switch this to the new helper later :p

paasta_tools/cli/cmds/validate.py
```diff
@@ -958,18 +960,49 @@ def validate_pool_limits(service_path: str) -> bool:
...
-            if cpu >= pool_limits["max_cpus"]:
-                returncode = False
-                print(
-                    failure(
-                        f"""{service}.{instance} in {cluster} has {cpu} CPUs, which exceeds the limit of {pool_limits['max_cpus']} for the {pool} pool.
-                        If you need to run a workload with this many CPUs, consider using the {pool_limits['recommended_pool']} pool instead.""",
-                        "",
-                    )
+            if instance_config.get_instance_type() not in ("kubernetes", "eks"):
+                continue
```
nemacysts, line 971: i think similar to mbrankin's PR: we might want to use PaastaServiceConfigLoader so that we can quickly filter and just load the configs we care about

paasta_tools/cli/cmds/validate.py
```diff
@@ -958,18 +960,49 @@ def validate_pool_limits(service_path: str) -> bool:
...
+                full_config_flattened=config_flattened,
+            )
+
+            if cpu_comment is not None and cpu_comment.strip():
+                continue
+
+            returncode = False
+            print(
+                failure(
+                    f"{service}.{instance} in {cluster} has {cpu} CPUs, which exceeds the limit of {pool_limits['max_cpus']} for the {pool} pool."
```
```suggestion
                    f"{service}.{instance} in {cluster} has {cpu} CPUs, which exceeds the recommended limit of {pool_limits['max_cpus']} for the {pool} pool."
```

nemacysts: and hmm, maybe we want to also include what the true max is so that folks don't try to do something like `cpus: 512` during an incident to give something a ton of resources and get confused as to why stuff isn't running :p

paasta_tools/cli/cmds/validate.py
```diff
@@ -958,18 +960,49 @@ def validate_pool_limits(service_path: str) -> bool:
...
+
+            cpu_comment = _get_comments_for_key(
+                data=config[instance],
+                key="cpus",
+                full_config=config,
+                key_value=config[instance].get("cpus"),
+                full_config_flattened=config_flattened,
+            )
+
+            if cpu_comment is not None and cpu_comment.strip():
```
nemacysts: we might want something like the existing `override-$FOO` comment markers just to make folks be extra explicit

paasta_tools/cli/cmds/validate.py
sidtuladhar: my concern for adding a true max is that we'd always need to have it updated with whatever the max we've set in the karpenter nodepool. In your experience, do engineers just use an unrealistic big number during DARs for mitigation? In most cases, if they do double the CPU + override comment should go to the fallback pool ig

nemacysts: they likely won't do something super crazy like cpus: 1000 - but they also won't really know if the amount they're adding an override comment for is schedulable or not

nemacysts: i.e., i think we sorta need to check two things: 
* if they're using something that will fall into the fallback nodepool (and that they then have an override)
* that they're not using something that won't even fit in the fallback nodepool

i think we could also probably script updating the SystemPaastaConfig here from our flux repo? i have a similar type of script for updating some of our default TSC and whatnot :)

PR #4378, opened 2026-08-20
Add resolute .deb to releases

nemacysts:
@vulpine says I'll be their new best friend if I do this

real description: Yelp/nerve-tools tests currently grab paasta-tools debs from our releases for GHA builds. this is fine, but what is not fine is that we are only uploading (and testing against) a jammy deb - which doesn't quite work out so well if you're trying to add a resolute build to nerve-tools

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,10 @@ jobs:
         with:
           name: deb-jammy
           path: dist/
+      - uses: actions/download-artifact@v4
+        with:
+          name: deb-resolute
+          path: dist/
       - name: Release
         uses: softprops/action-gh-release@v1
         if: startsWith(github.ref, 'refs/tags/v')


nemacysts:
@vulpine this will only do this for new releases - i think i can backfill an old release if necessary with some CLI nonsense tho!

vulpine (approved):

vulpine:
> @vulpine this will only do this for new releases - i think i can backfill an old release if necessary with some CLI nonsense tho!

Thanks so much - if you're able to backfill 1.60.X that should be perfect (since I know nerve-tools passes its tests with 1.60 Jammy.)

nemacysts:
> Thanks so much - if you're able to backfill 1.60.X that should be perfect (since I know nerve-tools passes its tests with 1.60 Jammy.)

@vulpine err, upon thinking about this a little more - this is probably gonna be slightly painful to do without potentially leaking internal build nonsense: i think i'd need to hack up the makefile to build paasta without including any internal packages, since i can't just set the Make flag to use public dependencies due to our external pypi blocks and whatnot :(

nemacysts:
that said, i think this should kick in with 1.60.3 (which i just released now)

PR #4384, opened 2026-09-02
Build arm64 paasta-tools debs in the release pipeline - COMPINFRA-6038

EvanKrall:
COMPINFRA-6038

Have Github Actions build jammy/arm64 debs as well. (I'm open to the idea of building arm debs on noble+resolute as well).


jvperrin (commented):
Building for resolute sounds good to me, but no need to for noble (I don't even see any references to it in this file except on line 57?)

nemacysts (approved):
we can probably punt on resolute/etc arm debs for now

PR #4387, opened 2026-09-07
Upgrade paasta to python 3.11

jall0h:
PaaSTA python upgrade from 3.10 to 3.11 alognside other requirements upgrades and test fixes.

- upgrade paasta to python from 3.10 to 3.11 
- upgrade `black` version from 22.3.0 to 22.10.0.
- upgrade `multidict` from 4.7.6 to 6.0.5
- upgrade `yarl` from 1.1.1 to 1.9.3
- upgrade `pygpgme` from 0.3 to 0.6
- fix tests that failed due to `InvalidSpecError` in` test_kubernetes_tools, test_cmds_local_run, test_cmds_mark_for_deployment and test_cmds_verify_image_exists`

PAASTA-18819

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -63,7 +63,7 @@ ply==3.4
 progressbar2==4.3.2
 prometheus-client==0.7.1
 protobuf==3.15.0
-py==1.5.2
+py==1.11.0
 pyasn1==0.4.2
 pyasn1-modules==0.2.1
 pyformance==0.4
--- a/requirements.txt
+++ b/requirements.txt
@@ -49,7 +49,7 @@ manhole==1.5.0
 MarkupSafe==1.1.1
 monotonic==1.4
 msgpack-python==0.5.6
-multidict==4.7.6
+multidict==6.0.5
 mypy-extensions==1.0.0
 nats-py==2.8.0
 networkx==2.4
@@ -115,6 +115,6 @@ WebOb==1.8.7
 websocket-client==0.44.0
 ws4py==0.5.1
 wsgicors==0.7.0
-yarl==1.1.1
+yarl==1.9.3
 zope.deprecation==4.2.0
 zope.interface==7.2
--- a/tests/cli/test_cmds_local_run.py
+++ b/tests/cli/test_cmds_local_run.py
@@ -249,7 +249,7 @@ def test_perform_http_healthcheck_failure_with_multiple_content_type(mock_http_c
 
 
 @mock.patch("paasta_tools.cli.cmds.local_run.perform_http_healthcheck", autospec=True)
-@mock.patch("time.sleep", autospec=True)
+@mock.patch("time.sleep", autospec=None)
 def test_run_healthcheck_http_success(mock_sleep, mock_perform_http_healthcheck):
     mock_docker_client = mock.MagicMock(spec_set=docker.APIClient)
     fake_container_id = "fake_container_id"
@@ -265,7 +265,7 @@ def test_run_healthcheck_http_success(mock_sleep, mock_perform_http_healthcheck)
 
 
 @mock.patch("paasta_tools.cli.cmds.local_run.perform_http_healthcheck", autospec=True)
-@mock.patch("time.sleep", autospec=True)
+@mock.patch("time.sleep", autospec=None)
 def test_run_healthcheck_http_fails(mock_sleep, mock_perform_http_healthcheck):
     mock_docker_client = mock.MagicMock(spec_set=docker.APIClient)
     fake_container_id = "fake_container_id"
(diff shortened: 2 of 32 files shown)


tests/cli/test_cmds_local_run.py
```diff
@@ -249,7 +249,7 @@ def test_perform_http_healthcheck_failure_with_multiple_content_type(mock_http_c
 
 
 @mock.patch("paasta_tools.cli.cmds.local_run.perform_http_healthcheck", autospec=True)
-@mock.patch("time.sleep", autospec=True)
```
nemacysts, line 252: what was the error we were seeing here? autospec=None turns off spec enforcement, so it's usually a last resort :p

tests/cli/test_cmds_mark_for_deployment.py
```diff
@@ -125,7 +125,7 @@ def mock_periodically_update_slack():
 @patch("paasta_tools.cli.cmds.mark_for_deployment._log_audit", autospec=True)
 @patch("paasta_tools.remote_git.create_remote_refs", autospec=True)
 @patch(
-    "paasta_tools.cli.cmds.mark_for_deployment.load_system_paasta_config", autospec=True
+    "paasta_tools.cli.cmds.mark_for_deployment.load_system_paasta_config", autospec=None
```
nemacysts, line 128: (same comment as my time.sleep mock note above)

requirements.txt
```diff
@@ -49,7 +49,7 @@ manhole==1.5.0
 MarkupSafe==1.1.1
 monotonic==1.4
 msgpack-python==0.5.6
-multidict==4.7.6
+multidict==6.0.5
```
nemacysts, line 52: ignorable: we could probably go to 6.8.0 while we're making major version bumps, but this is fine too (and from looking at the changelogs for this package, it looks like the reason for the major bumps was just dropping support for older versions of Python)

yelp_package/extra_requirements_yelp.txt
```diff
@@ -19,7 +19,7 @@ named-decorator==0.1.4              # yelp-profiling dependency
 ndg-httpsclient==0.4.3              # vault-tools dependency
 okta-auth==1.3.2
 pycparser==2.20                     # vault-tools dependency
-pygpgme==0.3                        # vault-tools dependency
+pygpgme==0.6                        # vault-tools dependency
```
nemacysts, line 22: might be good to double-check that there's no breaking changes between 0.3 and 0.6

from some recent shenanigans, i think this is pretty safe - but would be good to verify :D

nemacysts (commented):
ty for splitting this into multiple commits - made it super easy to review! :D

my only real concern is the autospec=None bits, fwiw

tests/cli/test_cmds_local_run.py
nemacysts: if i'm not mistaken, what's happening here is that we're likely mocking something that was mocked elsewhere (which is now an error)

jall0h:
```
E       unittest.mock.InvalidSpecError: Cannot spec a Mock object. [object=<MagicMock name='load_system_paasta_config' spec='function' id='125045998520912'>]

/usr/lib/python3.11/unittest/mock.py:508: InvalidSpecError
```

> "This error typically occurs in Python 3.11 when you try to use spec or autospec  with an object that's already a Mock . This is due to stricter validation added in Python 3.11."

from the upgrade wiki page FAQ, 

the fix suggested for these was to set autospec=None

nemacysts: ah yea, in this case what we'd want to do is fix our mocks so that we're not nesting them :D

jall0h:
@nemacysts Autospec issues were being caused by `mock_load_system_paasta_config` being patched as well as being an `autouse` fixture. In order to fix autospec error issues with `mock_load_system_paasta_config` for failing tests the following approaches were taken :

1. if it was being patched but not used in the function, it was removed from both the patch decorator and as a function parameter
2. if it was being patch and had a return_value parameter being set (in the decorator), it was removed from the patch decorator but kept as a function parameter with the return value being set in the function body.

(just a though may be useful to set up a ticket to clean up some functions that use `mock_load_system_paasta_config` as sometimes its not being used but has mock values set up in test function bodies despite autouse)

tests/cli/test_cmds_mark_for_deployment.py
```diff
@@ -158,11 +153,8 @@ def test_mark_for_deployment_happy(
 @patch("paasta_tools.cli.cmds.mark_for_deployment._log", autospec=True)
 @patch("paasta_tools.cli.cmds.mark_for_deployment._log_audit", autospec=True)
 @patch("paasta_tools.remote_git.create_remote_refs", autospec=True)
-@patch(
-    "paasta_tools.cli.cmds.mark_for_deployment.load_system_paasta_config", autospec=None
-)
 def test_mark_for_deployment_sad(
-    mock_load_system_paasta_config, mock_create_remote_refs, mock__log_audit, mock__log
+    mock_create_remote_refs, mock__log_audit, mock__logmock_load_system_paasta_config
```
nemacysts: i think this was meant to be `mock__log`?

nemacysts (commented):
one minor comment, but otherwise lgtm!

tests/cli/test_cmds_mark_for_deployment.py
jall0h: oops fixed in commit [d479551](https://github.com/Yelp/paasta/pull/4387/commits/d4795511bfe7c29ed629328a772be1ff12423018)

nemacysts (approved):

PR #4389, opened 2026-09-14
PAASTA-18976: Update alerting region for norcal-devc cluster

jall0h:
### Summary
Update region , where prometheus-otel shard is located, for norcal-devc cluster. 
There is one prometheus-otel shard for tracing for both norcal and pnw. It is located in uswest2.
PAASTA-18976


paasta_tools/cli/cmds/mark_for_deployment.py
```diff
@@ -404,6 +404,10 @@ def _build_default_error_alert_filter(
                 "skipping default error alert filter for this cluster"
             )
             continue
+        if (
+            cluster == "norcal-devc"
+        ):  # PAASTA-18976 - uswest1-devc does not have a prometheus-otel shard
+            region = "uswest2-devc"
```
nemacysts: can we do this without hardcoding these here?

maybe we can have some SystemPaastaConfig that maps cluster regions to the regions of the corresponding otel shard?

(or, if we only currently run these in a single region per-runtimeenv, maybe we map that way?)

cuza (approved):

